Issue #7592, opened 2026-08-07
perf: json_pipeline at 500k records is 97.6x bun (60.4s vs 618ms) while the same workload at 100 records BEATS bun — a scaling cliff, not a constant factor


proggeramlug:
## Measured in today's public-baseline regeneration (v0.5.1335, `2ba59501b`, pinned quiet M1 Max)

`honest_bench`'s JSON pipeline, 20 runs per cell, **output verified against the
Bun reference 20/20 for every language** — so this is a pure performance
finding, not a correctness one.

| workload | perry | bun | node | rust | zig | perry RSS |
|---|--:|--:|--:|--:|--:|--:|
| `json_pipeline_small` (100 records) | **77.0 ms** | 90.8 | 128.1 | 73.4 | 72.8 | **8.5 MB** |
| **`json_pipeline_full`** (500,000 records, 107.5 MB) | **60,358.5 ms** | **618.1** | 1,004.3 | 650.0 | 875.5 | **1,064 MB** |
| `image_convolution` (4K, 5×5) | **302.5 ms** | 947.2 | 1,284.8 | 429.8 | 279.8 | 56.2 MB |

**At 100 records Perry beats bun and node and lands within 5% of Rust and Zig,
on a quarter of bun's memory. At 500k records it is 97.6× bun.** Same source
file, same binary, same harness — only the fixture size changes.

That is the shape of the finding: **not slowness, a scaling cliff.** Across the
5,000× increase in input, bun's time grows 6.8× (startup-dominated at the small
end); Perry's grows **784×**. Rust 8.9×, Zig 12.0×, node 7.8×. Perry is the only
implementation whose growth is superlinear in the data, which points at an
algorithmic or GC-pathological term rather than a constant-factor gap.

RSS is 1,064 MB against bun's 579 MB — 1.8×, and notably *not* 97×. The memory
is not exploding proportionally to the time, which argues against "it simply
swapped" and for real work being repeated.

## Where to look

The workload is small and does five things (`workloads/1_json_pipeline/perry/json_pipeline.ts`):
`fs.readFileSync` a 107.5 MB string → `JSON.parse` to 500k objects → a filter +
field-derive loop building `out` → `JSON.stringify(out)` → `fs.writeFileSync`,
then an FNV-1a hash over the serialized string.

Each of those is separately suspect at this scale and they should be timed
individually before anything is optimised — this campaign has repeatedly paid
for working from an unattributed headline number. Candidates worth ruling in or
out first:

[...]

PR #7584, opened 2026-08-07
fix(gc): root the generator instance across its own prototype wiring (#7577)

proggeramlug:
Closes #7577.

## Root cause: an ordering bug, not a missing root

`js_generator_attach_prototype` (`crates/perry-runtime/src/object/global_this/generator.rs:381`)
bound the receiver's address at function entry and used it at the tail:

```rust
let obj_ptr = jv.as_pointer::<u8>() as usize;         // entry
...
let intermediate = js_object_alloc(0, 0);             // ALLOCATES
object_set_static_prototype(intermediate as usize, gen_proto_bits);   // ALLOCATES
object_set_static_prototype(obj_ptr, intermediate_bits);              // stale
obj                                                                    // stale
```

**This frame owns the only reference.** Codegen drops the caller's root
immediately before the call — from the emitted IR for the issue's own
reproducer, in `perry_fn_..._small`, the frame the fault report names:

```llvm
%r65.rs4p = load ptr addrspace(1), ptr %r28
%r66 = or i64 %r65, POINTER_TAG
%r67 = bitcast i64 %r66 to double
store ptr addrspace(1) null, ptr %r28              ; <-- the caller's root, dropped
%r68 = call double @js_generator_attach_prototype(double %r67, i32 0)
ret double %r68
```

That is the ordinary contract (a runtime helper roots its own arguments) and
this one did not. Every call it makes allocates:

| call | allocation |
|---|---|
| `wrap_async_generator_instance` | three closures + three `js_object_set_field_by_name` keys-array transitions |
| `generator_prototype_ptr` | lazily builds the **entire** generator intrinsic tower on its first call |
| `js_object_alloc` | by definition |
| `object_set_static_prototype` | `object_meta_ensure` mints the object's meta record out of the arena |

A copying minor in any of those windows gave **two** wrong answers with no
diagnostic:

1. the `[[Prototype]]` link was recorded against the **pre-move** address, so
   `Object.getPrototypeOf(gen())` on the live object found nothing;
2. the function **returned** that pre-move address — the caller's generator was
   a dangling pointer into retired from-space.

That is #7577's SIGBUS. Without the instruments it is a wrong answer and exit 0,
which is the expensive half.

Same shape, same fix, in three more places reachable from that call:
`js_generator_attach_closure_prototype`, `generator_function_prototype_of`, and
`wrap_async_generator_instance` / `set_method` below them. Fixing only the named
function would have handed a freshly re-read receiver to a callee that lets it
go stale again.

[...]

Issue #6942, opened 2026-07-28
gc: no reachable configuration exercises an evacuating minor with unpinned runtime locals — the #6655/#6935 bug class is untestable


proggeramlug:
Split out of #6935 (and it applies equally to #6655 / #6934).

Both of those PRs fixed real rooting gaps by audit, and both shipped a forced-evacuation regression suite that **passes on the pre-fix runtime**. That is not a weak test — there is currently no reachable configuration in which a compiled Perry program performs a *minor* cycle that evacuates while a runtime helper's raw Rust locals are unpinned. That is the exact state the whole bug class needs, and nothing can produce it today.

Measured on `main` (0bb03e82f), macOS host, `perry-dev` profile, with `PERRY_GC_DIAG=1`:

**1. The `gc()` hook runs a full mark-sweep, and evacuation is minor-only.**

A program doing ~12M escaping allocations plus explicit `gc()` calls produced ten `[gc] blocks: …` sweep lines and **zero** `[gc-evac-policy]` / `[gc-copy-minor]` lines. `maybe_print_evacuation_policy_diag` is only called from the `self.minor.is_some()` arm in `gc/cycle.rs`, so every one of those cycles was a full collection. Nothing moved, so a stale pointer is trivially still valid — which is exactly why the #6655 suite could not fail pre-fix.

Note also that the churn has to *escape* (pushed into a live array) before any collection happens at all; a loop allocating short-lived object literals is optimized away, so the "20000 short-lived objects" prelude both #6655 and #6935 use never triggers a GC on its own.

**2. `perry/gc`'s `minor()` evacuates, but pins the operands the bug is about.**

Switching the churn to `import { minor } from "perry/gc"` does reach the evacuator:

```
[gc-copy-minor] eligible=false fallback=conservative_stack
[gc-evac-policy] enabled=true reason=force … moved_objects=8467 released_original_objects=8467 cons_pinned=1488
```

Thousands of objects genuinely move. But `js_gc_module_minor` engages `ManualGcScanGuard::force_full_scan()` (deliberately, per #4977: "the callsite may hold live locals only on the native stack"). The conservative stack scan then pins everything reachable from the native stack — including the raw receiver / stored-value locals held inside the runtime helper. The bug is masked by construction.

**3. Disabling the conservative scan makes the configuration independently unsound.**

[...]

PR #7594, opened 2026-08-07
perf(gc): break the survivor-promotion handoff livelock (#7592)

proggeramlug:
Fixes the `build_out` half of #7592.

## The bug is a livelock, and it is not in JSON

#7592 reported `json_pipeline` at 500k records being 97.6× bun. Splitting the phases first, as the issue insists:

| phase | time | share |
|---|--:|--:|
| readFileSync | 89 ms | 0.1% |
| JSON.parse | 742 ms | 1.2% |
| **build_out** | **57,409 ms** | **94.1%** |
| JSON.stringify | 1,451 ms | 2.4% |
| fnv1a | 1,250 ms | 2.0% |

`JSON.parse` handles 107 MB in 742 ms and scales linearly. `build_out` — the `out.push({...})` loop — is **~100 % GC pause** (8,840 ms of traced pause against 8,633 ms of phase at 100k) and grows quadratically.

The GC trace says exactly what is happening. At 200k records, of 22 collections:

```
 19 x full/survivor_promotion_bytes     each freeing 0.0 MB at ~400 ms
  1 x full/old_gen_bytes
  1 x full/arena_bytes
  1 x minor/arena_bytes                 promoted: 0.0 MB
```

`copied_minor_promotion_handoff_due` replaces a minor with a full mark-sweep to make room in old-gen for survivors that are about to be promoted. But **a full mark-sweep is non-moving — it promotes nothing**. So it cannot relieve the pressure it was scheduled for: the survivor space still holds the same 108 MB, the reclaim baseline it resets does not count those bytes, and the predicate is true again at the very next minor. The copying minor that would have done the promotion never gets to run.

That is 7.6 s of an 8.6 s phase spent on collections that free nothing. Peak RSS is the same whether those 19 collections run or not.

## Fix

Latch it: **one handoff per copying minor.** The handoff makes room; the copying minor performs the promotion that consumes it. The latch clears only on a *copying* minor, because a non-moving minor fallback promotes nothing and would reinstate the livelock at half rate.

The guard sits before the `copied_minor_promotable_active_survivor_bytes()` walk, so a suppressed handoff also skips that O(n) survivor pass.

Same workload, after:

| | cycles | pause | `survivor_promotion` fulls | promoted |
|---|--:|--:|--:|--:|
| main | 22 | 8,782 ms | **19** | **0.0 MB** |
| this PR | 6 | 2,828 ms | **1** | **110.0 MB** |

## Measurements

Both arms built from the same package set and linked against a pinned `PERRY_RUNTIME_DIR`, interleaved, output hash checked every row:

| records | main | this PR | speedup | main RSS | PR RSS |
|--:|--:|--:|--:|--:|--:|
| 25,000 | 48 ms | 48 ms | 1.0× | 81 MB | 81 MB |
| 50,000 | 499 ms | 459 ms | 1.1× | 163 MB | 163 MB |
| 100,000 | 2,095 ms | 1,437 ms | 1.5× | 245 MB | 318 MB |
| 200,000 | 8,654 ms | 2,858 ms | 3.0× | 484 MB | 572 MB |
| 500,000 | 57,242 ms | **10,551 ms** | **5.4×** | 1,064 MB | 1,325 MB |

Output hash identical on every row.

**The RSS cost is real and I am not hiding it**: +17–24 % at the large sizes, because the run now does 6 collections instead of 22. It is a genuine trade, not a free win. It is confined to workloads that actually hit this livelock — see below.

## GC ratchet

Checking against the pinned baseline reports 29 regressions, but that baseline is from 0.5.1315 on a different host, so it cannot separate this change from drift. I measured **both arms back to back on one host with an identical package set** instead — 144 metric medians across all 12 probes:

- **Every semantic counter is identical**, except `12_large_live_set.heap_used_bytes` at −0.01 % (2,232 B). That cell is the one the harness itself documents as ungated and sample-dependent (conservative-scan residue, observed spread 9,072 B over 36 runs), so 2,232 B is inside its own noise.
- Everything else that moved is RSS/wall — ungated, and ≤ 0.12 % on RSS.

That is the expected shape: the latch can only fire after a survivor-promotion handoff has occurred, so a workload that never hits the livelock is unaffected.

`cargo test -p perry-runtime`: 1,843 passed, 0 failed. `cargo fmt --check` and `scripts/check_file_size.sh` clean.

## Test

`test_survivor_promotion_handoff_waits_for_the_copying_minor`.

The obvious version of this test cannot fail: with an empty heap `copied_minor_promotion_handoff_due` returns `false` at the survivor-occupancy check regardless, so asserting the verdict passes with the latch deleted. The test therefore counts suppressions and asserts **the latch branch is what rejected it** — and checks that a trigger kind the handoff never applies to is *not* counted, so the counter cannot pass by incrementing everywhere.

Verified by deleting the latch and watching it go red:

```
assertion `left == right` failed: the latch branch must be what rejected it (ArenaBytes)
```

## What this does not fix

`build_out` is still ~28,500 ns/record and not yet flat, so #7592 stays open. With a nursery large enough to avoid collecting during the loop the same phase runs at 764 ns/record, so there is roughly another order of magnitude available. That remainder is the collection-budget question — the cap is a constant (16 MB × `NURSERY_CAP_SCALE_MAX`), so cadence is independent of live-set size. I tried a live-proportional cap and it is structurally wrong as written: the cap gates *from-space occupancy*, and here from-space is nearly all of live, so `cap = live` is a fixed point that stops scavenging entirely. Details are in the issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed repeated garbage-collection cycles caused by survivor-promotion pressure.
  * Suppressed redundant full collections until a copying minor collection completes.
  * Preserved expected behavior for direct collection triggers.

* **Performance**
  * Reduced unnecessary full collections and related overhead.
  * Improved affected workloads, with a measured RSS tradeoff.

* **Tests**
  * Added regression coverage for suppression, counting, and reset behavior.

* **Documentation**
  * Documented the fix, performance impact, RSS tradeoff, and validation results.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1337
+**Current Version:** 0.5.1338
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1337"
+version = "0.5.1338"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1337"
+version = "0.5.1338"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1337"
+version = "0.5.1338"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1337"
+version = "0.5.1338"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1337"
+version = "0.5.1338"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1337"
+version = "0.5.1338"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1337"
+version = "0.5.1338"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1337"
+version = "0.5.1338"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1337"
+version = "0.5.1338"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1337"
+version = "0.5.1338"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1337"
+version = "0.5.1338"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1337"
+version = "0.5.1338"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1337"
+version = "0.5.1338"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1337"
+version = "0.5.1338"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1337"
+version = "0.5.1338"
 dependencies = [
  "anyhow",
  "clap",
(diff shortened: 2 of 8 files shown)


proggeramlug:
## Audit before merge — verified end to end, merged as v0.5.1338

**Reproduced the livelock and the fix independently**, own builds of both arms
(same package set, same `PERRY_RUNTIME_DIR`, `PERRY_NO_AUTO_OPTIMIZE=1` both
sides), `PERRY_GC_TRACE=1` census on the 200k workload:

| | main (my run) | PR (my run) | PR's claim |
|---|--:|--:|--:|
| total cycles | 23 | **6** | 22 → 6 |
| `survivor_promotion` fulls | **20** | **1** | 19 → 1 |
| total pause | 10,159 ms | **3,271 ms** | 8,782 → 2,828 |
| promoted by the copying minor | — | **110.0 MB** | 110.0 MB |
| wall (perry-dev profile) | ~11–12 s | **5 s** | — |
| peak RSS | 476 MB | 572 MB (**+20 %**) | 484 → 572 (+18 %) |
| output hash | `0a883457` | `0a883457` | identical |

The census is the mechanism and it reproduces to the digit — including the
110.0 MB promoted, exactly. The RSS trade is real and matches the disclosed
figure; disclosing it rather than burying it is noted and appreciated.

**The diagnosis is structurally sound, verified in the code, not just the
trace.** On `main`, the handoff branch runs the full and **returns early** —
the minor it displaced never executes (`gc/mod.rs:178-184`). The full mark-sweep
is non-moving (long-established here: it is why `PERRY_GC_FORCE_EVACUATE` was
inert for every `gc()`-driven test, #6942/#6946), and the predicate keys on
`copying_active_survivor_in_use_bytes()`, which only a copying minor reduces
while the survivors are live. Predicate true → full instead of minor → survivor
bytes unchanged → predicate true again. A genuine livelock.

**The wiring is right in the two places that decide correctness:**
- `gc_collect_minor_copying_fast_path` is a thin wrapper over
  `_with_eligibility`, so the latch-clear covers both entry points.
- The single ineligibility bail (`return None`, copying.rs:1045) sits *before*
  the latch-clear at the completion tail — so a copying attempt that bails does
  **not** clear the latch. That is the correct semantics; clearing on a bailed
  attempt would reinstate the livelock at half rate.
- The predicate's trigger arm (`ArenaBytes | MallocCount`) confirms the test's
  choice of applicable kinds, and `Direct` is rejected ahead of the latch —
  which the test asserts via the counter *not* moving.

**Sabotage re-verified myself:** deleting the latch branch turns the test red at
the counter assertion with exactly the quoted message. The test's design — count
suppressions rather than assert the verdict, because an empty heap returns
`false` regardless — is the "assert the subject was live" discipline this repo's
gate history demands, applied correctly.

Gates re-run here: runtime suite **1,843 / 0**, `addr_class_inventory`,
`class_id_collisions`, `raw_handle_debt` (998/998), `check_file_size.sh`,
`cargo fmt --check` all clean.

## One coverage gap, noted not blocking

The test exercises the latch via direct calls to `note_copying_minor_completed()`
— nothing asserts the **wiring**, i.e. that the completion tail of the copying
minor actually calls it. If a future refactor drops that call, every handoff
after the first is silently suppressed forever, and no test goes red. The blast
radius is bounded (handoffs only serve copying-minor promotion, and the other
full triggers — `old_gen_bytes`, `arena_bytes` — still fire, as my PR-arm census
shows), which is why this is a note rather than a blocker. Same shape as the
coverage gap I flagged on #7584: the helper is tested, the call site is not.

## Bookkeeping

Merged with a plain `(#7592)` reference — **the issue stays open**. Post-fix the
500k row is ~10.6 s against bun's 618 ms, still ~17×, and the PR's own analysis
points at the remaining order of magnitude: with a nursery large enough to avoid
collecting during the loop the phase runs at 764 ns/record vs ~28,500 now. The
failed live-proportional-cap attempt (cap gates from-space occupancy; here
from-space ≈ live, so `cap = live` is a fixed point that stops scavenging) is
recorded on the issue, which is exactly where it needs to be so the next attempt
starts past it.

Issue #7024, opened 2026-07-29
gc-matrix: --pressure disables the very path #7019 added — the 'default' arm runs ZERO copying minors on all 22 corpus rows


proggeramlug:
Found while re-measuring the matrix after #7019 landed. This is a **gate-methodology** defect: the matrix's own pressure knob switches off the collector path #7019 added, so the arms that are supposed to describe the shipped configuration measure the *pre*-#7019 non-moving path instead.

## Measurement

`scripts/gc_repsel_matrix.sh --arms all --pressure 8`, `origin/main` @ `4340bffc9`, 22 corpus rows. Separating the two counters the harness currently sums together (see the sibling issue on that):

| arm | files with a cycle | files with **`[gc-copy-minor] ran`** | Σ `copied_objects` |
|---|---|---|---|
| `default` | 13/22 | **0/22** | **0** |
| `verify_evac` | 13/22 | **0/22** | **0** |
| `gen_gc_off` | 13/22 | 0/22 | 0 |
| `gen_off_verify` | 13/22 | 0/22 | 0 |
| `cons_scan_off` | 13/22 | 13/22 | 21 767 487 |
| `evac_minor` | 21/22 | 21/22 | 3 628 777 |
| `shipped_default` (no env at all) | 2/22 | 2/22 | 40 087 037 |

The `default` arm — "as-shipped GC configuration under allocation pressure" — runs **zero** copying minors on **every** corpus file. Its only movement evidence is `moved_objects=` from the older C4b mark-sweep evacuation policy, which is a different collector.

## Mechanism

`gc_check_trigger`'s defer arm (the one that hands the collection to `js_gc_loop_safepoint` → `gc_safepoint_moving_minor`, i.e. all of #7019) is guarded by

```rust
if gc_moving_loop_polls_enabled()
    && crate::arena::arena_total_bytes() < gc_moving_defer_hard_cap_dyn_bytes()
```

[...]

Issue #6181, opened 2026-07-09
[perf] Minor GC does O(old-gen) work: full-region sweep walk, whole-heap RS rebuild, per-cycle page-meta iterations


proggeramlug:
## Problem

Minor GC cost includes several O(old-gen) components even when nothing old is touched:

1. Sweep visits/rewrites every old-gen header on every minor (`gc/oldgen.rs:1181-1225` cursor over all regions; flags/accounting rewrites `gc/oldgen.rs:1291-1311,1338,1357`; minors set `reclaim_dead_old_blocks=false` but still walk, `gc/cycle.rs:1390-1416`).
2. The remembered-set finalize rebuild walks the ENTIRE heap (all arena objects + full malloc registry) every cycle (`gc/cycle.rs:1244-1264`, `gc/verify.rs:279-349`), and with `require_marked=false` for minors it re-dirties pages from unreachable-but-unswept old parents.
3. Old-page bookkeeping iterates every old 4 KB page meta up to 3× per minor (`arena/page_meta.rs:555-571`, `gc/oldgen.rs:181-184,1199`), plus the promotion-handoff pre-walk over all arena objects once survivor in-use ≥ 24 MB (`gc/policy.rs:864-902`).

Minor pause grows monotonically with process lifetime on servers.

## Proposed fix

- Restrict the minor sweep cursor to nursery+survivor regions; maintain old-gen accounting incrementally from promotion/reclaim events; per-block live counts so fully-dead block reset needs no per-object walk.
- Maintain the RS incrementally (the barriers already do) and restrict the finalize rebuild to evacuated/moved parents — `rebuild_evacuated_old_to_young_remembered_set` (`gc/verify.rs:243-253`) plus the #5029 snapshot repair already exist for exactly this.
- Lazy epoch-counter resets for page metas; keep defrag candidates on a small heap updated at promotion time.

## Context

2026-07-09 GC audit (pause/scalability + barrier dimensions).

proggeramlug:
Status: the whole-heap old→young RS rebuild for minors was eliminated in #6213 (minors jump AtomicFinalize→Sweep; barriers + evacuation-sticky + #5029 snapshot repair carry coverage). Remaining from this issue: the minor sweep's full-region walk and per-cycle census cost — the census half just landed for budgeted cycles in #6239 (no exact set built at all); the sweep-walk restriction to young regions remains open and is queued in the #6180 flip campaign.

proggeramlug:
[...]

Issue #7554, opened 2026-08-07
gc-ratchet has measured nothing since 2026-08-05: harness validation fails before the probes run (12_large_live_set spread 6768)


proggeramlug:
## The GC ratchet has measured nothing since 2026-08-05

`gc-ratchet` fails on `main` in `test_pinned_artifact_retention_is_deterministic`:

```
AssertionError: 6768 != 0 : 12_large_live_set.heap_used_bytes was not
deterministic when pinned; it must not be in a gating family
```

That assertion lives in the **"Harness unit tests and artifact validation"** step, which runs **before** the measurement step. So the probes themselves — all twelve — have not executed on any branch since this broke. The job is red, and the thing it exists to measure is not running at all. That is CLAUDE.md's fourth failure mode ("the gate runs but its subject never did") in its most expensive form, compounded by `gc-ratchet` not being a required context, so the red blocks nothing either.

It has failed on every completed `main` run since **2026-08-05T15:07Z**, which is the window in which #7446 re-pinned the artifact.

## The assertion is right; the artifact is unfit

This is not an over-strict test. `benchmarks/gc_ratchet/tolerances.json` justifies gating `heap_used_bytes` on exactly this basis:

> *"Observed spread 0.000% over 21 runs on all 8 probes: bit-identical. The band is therefore pure anti-brittleness margin … not a noise allowance."*

The pinned artifact contradicts its own premise. Every other probe still holds it:

| probe | `heap_used_bytes` spread |
|---|--:|
| `01_nursery_churn` … `11_collect_at_depth` (11 probes) | **0** |
| **`12_large_live_set`** | **6768** |

and the samples show it is a single outlier, not noise:

```
[59943824, 59943824, 59950592, 59943824, 59943824, 59943824, 59943824]
                     ^^^^^^^^ one run, +6768 bytes
```

`heap_total_bytes` on the same probe is spread 0 across all seven.

## Two possible causes, and they need different fixes

1. **The re-pin captured a bad sample.** #7446 ran on the dedicated bench host; one run in seven landing 6,768 bytes high could be incidental (a lazily-created runtime global, an interned string) rather than a property of the probe. Fix: re-pin `12_large_live_set`, ideally with more repeats.
[...]

Issue #7437, opened 2026-08-05
GC: old-gen fragmentation — scattered survivors pin 105 MB of blocks for a ~1 MB live set


proggeramlug:
Found while building the `12_large_live_set` gc-ratchet probe for #7432 (adaptive tenuring). This is a pre-existing old-gen characteristic that mass promotion now makes routine, not a #7432 regression.

## Symptom

A program that promotes a ~100 MB linked structure, then drops all but every 64th node (~1 MB truly live), retains **105.6 MB of arena in-use after an explicit `gc()`**. The full collection itself works — `PERRY_GC_DIAG=1` shows `freed_bytes=87,679,624` on the final manual collect — but the in-use accounting barely moves:

```
[gc] blocks: general=50 (49 live), longlived=46 (40 live), freed_bytes=87679624
#gcmetric heap_used_bytes=105588616
```

49 of 50 blocks still count as live because keeping every 64th node scatters ~218 live objects into every 1 MB block. Block reclaim (`arena_reset_empty_blocks` / old block reclaim) can only reclaim fully-dead blocks, and the swept holes don't lower block offsets, so `heapUsed` — and RSS, since the blocks stay mapped — tracks the scattered-survivor high-water forever. V8 compacts old space in this situation; Node reports a proportionally small heap for the same program.

The old-page defrag policy (C4b) did not engage: the traced run reports `old_page_candidate_pages: 0` throughout.

## Repro

`benchmarks/gc_ratchet/probes/12_large_live_set.ts` (in #7432), compiled with `PERRY_NO_AUTO_OPTIMIZE=1`; run with `PERRY_GC_DIAG=1`. The probe pins today's 105.6 MB deterministically (0% spread), so any improvement must deliberately re-pin.

## Why it matters more after #7432

The adaptive tenuring threshold promotes long-lived cohorts on first copy (that is its job — re-copying them was 4.39 GB of waste on tree.ts), so old-gen now receives the scattered-survivor pattern wholesale. The tree.ts residual RSS gap vs `PERRY_GC_SCAVENGE=0` (221 MB vs 103 MB) is this plus old-reclaim pacing (#SIBLING), not survivor re-copying.

## Directions

- Old-page defrag policy: why zero candidate pages here — thresholds, or the pattern is outside its selection criteria?
- Sparse-block evacuation: blocks whose live fraction is < ~25% are exactly the C4b evacuation case; the survivors are unpinned class instances.
[...]

PR #7596, opened 2026-08-07
perf(gc): live-proportional collection budgets at both generations (#7592)

proggeramlug:
Second half of #7592, stacked on #7594. Three changes, one principle: **no constant band may pace a collector whose per-cycle cost is O(live)** — total work goes quadratic in the live set, and a bigger constant only moves the cliff.

## What was left after #7594

With the livelock latched, the 200k/500k trace still spent ~100 % of `build_out` in GC pause, and two of the six remaining cycles were *guaranteed* futile:

```
 #  kind                    trigger      old_before  promoted   freed   pause
 4  full   survivor_promotion_bytes         4.2 MB       0       0.0   1,015 ms   <- handoff over a near-empty old-gen
 6  full              old_gen_bytes       274.5 MB       0       0.0   2,100 ms   <- reclaim of just-promoted live bytes
```

## The changes

**1. The survivor-promotion handoff fires on CURRENT old-gen pressure only.** It used to fire on `old + promotable` — a prediction of where old-gen would land after the promotion. But promotable bytes sit in the survivor space, where a full mark-sweep can neither reclaim them (they are live) nor reclaim the old-gen space they have not yet occupied. That is the #7594 mistake in another coat: scheduling a non-moving collection for bytes it cannot affect. The only useful work a handoff can do is clear *current* old garbage so the promotion lands in reused holes; over 4.2 MB of old-gen it is a 1,015 ms no-op.

**2. A copying minor's promoted bytes credit the old-reclaim baseline.** Promoted bytes are live *by construction* — only marked-live objects get copied — so a reclaim fired because promotion crossed a threshold finds them all live and frees nothing. The credit is exactly the promoted delta, never a resync: pre-existing old garbage still counts. The trade is the standard GOGC one — promoted-then-dead bytes now wait for the growth band — and it is the one visible ratchet cost (below).

**3. Both pacing bands become live-proportional.**
- The old-reclaim growth band is `max(32 MB, baseline / 2)` — Go's GOGC shape. Shared by `old_reclaim_pressure_due` and `gc_old_reclaim_debt_bytes` so dueness and debt cannot diverge (#7024's two-predicates family).
- The scavenge nursery cap is `max(influx_driven, tenured_live / 2)`. Keyed on **old-gen reclaimable pressure**, NOT total arena in-use — the cap gates `young_scavenge_cap_due()` against from-space occupancy, and a cap defined by a total that *includes* the young generation is a fixed point from-space can never cross (measured on the first attempt: scavenging stopped entirely, 0 copying minors at 200k). Removing this term costs an extra 1,100 ms minor at 500k — each extra minor pays the O(old-gen) fixed cost (#6181) — and the count grows with N, so it is load-bearing, not tuning.

## Measurements

Three arms, identically built and linked against a pinned `PERRY_RUNTIME_DIR`, interleaved, output hash checked every row:

| records | main | #7594 latch | this PR | vs main | RSS (this PR) |
|--:|--:|--:|--:|--:|--:|
| 25,000 | 42 ms | 42 ms | 42 ms | 1.0× | 81 MB |
| 50,000 | 546 ms | 534 ms | 501 ms | 1.1× | 163 MB |
| 100,000 | 2,257 ms | 1,499 ms | 863 ms | 2.6× | 302 MB |
| 200,000 | 8,889 ms | 3,505 ms | 2,617 ms | 3.4× | 581 MB |
| 500,000 | 61,589 ms | 12,139 ms | **5,784 ms** | **10.6×** | 1,348 MB |

Output hash identical on every row. RSS is +1.7 % over the latch arm at 500k (and −27 % of main's *regression* budget: main was 1,064 MB but 57 s slower). ns/record is flat within ~30 % across a 20× size range; on main it grows 70×.

The 500k trace after: 4 cycles, none futile — one parse-garbage reclaim (1,045 ms, frees 111 MB), one Eden scavenge, one promotion pass.

## GC ratchet — one gated cell moves, and I am flagging it, not hiding it

Both arms measured back to back on one host (the pinned baseline is 0.5.1315 on another machine and cannot separate this change from drift). 144 metric medians across all 12 probes: **every semantic counter identical** except:

- `12_large_live_set.heap_total_bytes`: **95.4 MB → 130.0 MB (+36 %)** — gated, band 2 %, so the official check goes red on this cell.

I attribution-tested it: with only the promoted-bytes credit disabled, the probe is **byte-identical to the latch arm** — the growth is 100 % the deferred post-promotion reclaim, i.e. the intended GOGC trade. On the same probe, same run: `heap_used_bytes` +0.02 % (nothing extra retained), `peak_rss_bytes` **−0.7 %**, `wall_ms` **−7.4 %**, and every copy/promote/cycle counter identical. The growth is reserved-block high-water from collecting less often, not resident memory and not retention.

**This PR therefore needs a maintainer decision on that one baseline cell** (`--update-baseline` scoped to it, per the ratchet's own flow). If the reserved high-water is judged unacceptable, the credit (change 2) can be dropped independently — it is one call site — at the cost of reinstating the 2,100 ms futile full at 500k.

## Tests

- `test_copying_minor_promotion_handoff_requires_current_old_pressure` — pins the #7592 shape (108 MB promotable over 4.2 MB old-gen: NOT due) and that real current pressure still fires.
- `test_old_reclaim_band_is_proportional_and_promotion_credits_baseline` — the band floor/scaling, dueness/debt sharing one trigger, the credit arithmetic, and that a reclaim is not due immediately after promotion.
- Full `perry-runtime` suite: 1,844 passed, 0 failed. `cargo fmt --check` and `check_file_size.sh` clean.
- Gap suite run in progress at time of filing; result will be posted as a comment.

## What this does not fix

`build_out` at 500k is ~23,000 ns/record — flat, but still ~30× off the 764 ns/record a collection-free run shows. The remaining structural cost is the two-hop promotion (Eden→survivor→old copies 268 MB twice; 3.9 s of the 5.1 s). Collapsing it needs promote-on-first-copy *within the first copying minor*, which is a tenuring-policy design with #7432's determinism constraint — follow-up on #7592.

## Summary by CodeRabbit

- **Performance**
  - Improved garbage collection pacing by adapting reclaim thresholds to current memory pressure.
  - Adjusted nursery sizing and promotion handling to better balance throughput and memory usage.
  - Delays unnecessary old-generation reclamation after minor collections when appropriate.

- **Bug Fixes**
  - Improved reclaim debt calculations and promotion handoff behavior across collection cycles.

- **Tests**
  - Added coverage for proportional reclaim thresholds, promotion accounting, boundary conditions, and low-memory scenarios.

- **Documentation**
  - Documented performance results, including benchmark improvements and memory impact.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1338
+**Current Version:** 0.5.1339
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1338"
+version = "0.5.1339"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1338"
+version = "0.5.1339"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1338"
+version = "0.5.1339"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1338"
+version = "0.5.1339"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1338"
+version = "0.5.1339"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1338"
+version = "0.5.1339"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1338"
+version = "0.5.1339"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1338"
+version = "0.5.1339"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1338"
+version = "0.5.1339"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1338"
+version = "0.5.1339"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1338"
+version = "0.5.1339"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1338"
+version = "0.5.1339"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1338"
+version = "0.5.1339"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1338"
+version = "0.5.1339"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1338"
+version = "0.5.1339"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1338"
+version = "0.5.1339"
 dependencies = [
  "block2",
  "objc2",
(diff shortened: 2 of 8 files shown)


proggeramlug:
## Audit before merge — verified end to end, merged as v0.5.1339

**Three-arm census reproduced on my own builds** (same fixture, same trace
parser, interleaved, hash-checked — 200k records, dev profile):

| arm | cycles | total pause | futile fulls | wall | RSS |
|---|--:|--:|--:|--:|--:|
| main (pre-#7594) | 23 | 10,159 ms | 20 × survivor_promotion | ~12 s | 476 MB |
| #7594 latch | 6 | 3,270 ms | 1 × survivor_promotion (391 ms) + post-promotion reclaim | 5 s | 572 MB |
| **#7596** | **4** | **1,854 ms** | **0** — the old_gen_bytes cycle is 1 ms | **4 s** | 582 MB |

Both futile classes are gone: the handoff-over-empty-old-gen AND the
just-promoted reclaim. Output hash identical across all three arms. RSS +1.7 %
over the latch arm — exactly as disclosed.

**Both tested changes survive my sabotage:**
- Restoring the `+ promotable` prediction → `..._requires_current_old_pressure`
  red at the #7592-shape assertion (108 MB promotable over 4.2 MB old-gen).
- Neutering the credit → `..._promotion_credits_baseline` red at the credit
  arithmetic.

**The design reasoning holds up under adversarial reading.** The credit is a
delta, never a resync, so pre-existing old garbage keeps counting; a baseline
left above in-use can only persist until the next reclaim, which resets it —
in-use only drops at a reclaim, so there is no stuck state. The proportional
band's #7437 interaction is genuinely *better* than the constant band (a futile
reclaim inflates the next band instead of re-firing every constant step). And
the nursery cap being keyed off old-gen rather than total arena is the
load-bearing detail — the PR *measured* the self-referential variant failing (0
copying minors) rather than reasoning about it, which is the standard this
repo's gate history demands.

**Maintainer decision on the gated ratchet cell, recorded here:**
`12_large_live_set.heap_total_bytes` +36 % (95.4 → 130.0 MB) is **accepted**.
The attribution test is what earns it: with only the credit disabled the probe
is byte-identical to the latch arm, so the growth is 100 % the intended GOGC
trade — reserved-block high-water from collecting less often. On the same probe
`heap_used` is +0.02 % (no extra retention), **peak RSS −0.7 %** and wall
−7.4 %. Resident memory is the plan's goal; reserved high-water is the price of
paying O(live) less often, and the credit is one call site to drop if this
proves wrong in practice. The pinned baseline (0.5.1315, other host) should be
regenerated on the mini as part of the #7554 repair rather than patched
per-cell from a dev-Mac run.

## Two follow-ups, both already assigned

1. **Change 3 has no test that can see it.** The existing
   `scavenge_nursery_cap_effective_bytes` tests (tenuring.rs:418–432) do pin
   the influx floor — so #7377's small-live-set guarantee is regression-tested
   and passes on this arm, which is the safety-critical half — but with an
   empty test heap `max(influx, old_reclaimable/2)` degenerates and the
   proportional term is invisible to them. I sabotaged the term out and nothing
   asserts its presence. Assigned to a running agent along with the crossover
   case.
2. The PR body promised the gap-suite result as a comment and none was posted.
   My own targeted verification (three-arm hash-identical output + full runtime
   suite 1,844/0 + all four lint gates) stands in for it; if the author's sweep
   surfaces anything, it lands on #7592.

**#7592 stays open** — 5.8 s vs bun's 618 ms is still ~9×. The named remainder
is the two-hop promotion (Eden→survivor→old copies 268 MB twice, 3.9 s of the
5.1 s), and the promote-on-first-copy design note is assigned. The stringify
(1,451 ms) and fnv1a (1,250 ms) phases are also now first-order and have an
agent on them.

PR #7580, opened 2026-08-07
fix(codegen): PreallocateBoxes must not shadow a module-level global (#7521)

proggeramlug:
Closes #7521.

## Root cause

Not in `diagnostics_channel`, and not a closure-capture bug. `traceCallback` is
a bystander — the minimal repro has no `traceCallback` in it, and no array:

```ts
import { tracingChannel } from "node:diagnostics_channel";   // any import at all
{
  const events: string[] = [];
  function t1(a: any) { events.push("fn:" + a); }
  t1("A");
  console.log(JSON.stringify(events));   // node: ["fn:A"]   perry: []
}
```

Objects, numbers and strings lose the same way — inside `t1`, the captured
binding simply reads `undefined`, so `acc.n++` throws
`Cannot read properties of undefined` and `events.push(...)` silently
no-ops. The `import` matters only because it makes the file an ES module,
hence strict, which is what selects the affected lowering. Everything else
about the failing gap test (`traceCallback`, subscribers, arrows, the array)
was incidental.

`crates/perry-codegen/src/stmt/mod.rs`, `emit_preallocate_boxes`: it allocated
a heap box — and, decisively, a `ctx.locals` slot — for **every** id in a
`Stmt::PreallocateBoxes` directive, including ids that
`codegen/module_globals_emit.rs` had already promoted to
`@perry_global_<mod>__<id>`.

Codegen already has a rule for this collision, spelled out identically at every
read and write site in `expr/literals_vars.rs` and `stmt/let_stmt.rs`:

```rust
ctx.boxed_vars.contains(id) && !ctx.module_globals.contains_key(id)
```

The module global wins — it is already the shared, forward-visible, GC-rooted
cell a box would provide. `emit_preallocate_boxes` was the one place that never
checked, and the stale `ctx.locals` entry it left behind is consulted **before**
`ctx.module_globals` on both the `Stmt::Let` reuse path (`let_stmt.rs:322`) and
the `LocalGet`/`LocalSet` store paths (`literals_vars.rs:642`, `:814`). So the
declaration wrote its value into the local box-pointer slot, the module global
was never stored, and the closure — which reads through the global, with zero
captures — saw the `undefined` the global was defined with.

The emitted IR, before:

```llvm
%r2  = call i64 @js_box_alloc_bits(i64 <undefined>)   ; PreallocateBoxes([1])
%r8  = call i64 @js_closure_alloc(ptr @perry_closure_m3_ts__0, i32 0)
[...]

Issue #6847, opened 2026-07-26
auto-opt: ext-zlib provider link failure on cold object cache (undefined js_zlib_deflate_raw_sync + panic_unwind symbols)


proggeramlug:
Cold-cache reproduction (any tree, main @ 21a3d00e1):

```
rm -rf target/perry-auto-* && rm -rf /private/tmp/node_modules/.cache/perry
target/release/perry test-files/test_gap_zlib_4917_level.ts -o /tmp/zl
```

Link fails with:

```
Undefined symbols for architecture arm64:
  "__RNvCs4SDFJOLwvtW_7___rustc17___rust_drop_panic", referenced from:
      panic_unwind ... in __libperry_ext_zlib.a_nosharedeps.lib_trimmed.lib
  "_js_zlib_deflate_raw_sync", referenced from: _main
```

The auto-optimize pipeline pairs the ext-zlib provider archive with an auto-opt stdlib rebuild that lacks the zlib native symbols, and the provider archive carries panic_unwind runtime objects the panic=abort link cannot satisfy. `PERRY_NO_AUTO_OPTIMIZE=1` compiles and links the same test fine.

Why this went unnoticed: the compile-object cache under `/private/tmp/node_modules/.cache/perry` held pre-breakage objects with the old routing baked in; any machine with a warm cache (dev boxes, CI with restored caches) sails past the broken codegen path. A reboot (cleared /tmp) exposed it — six gap tests now compile_fail cold: test_gap_zlib_4917_level, test_gap_zlib_3285_params, test_gap_zlib_fs_assert_2935_2752_2971, test_gap_3662_node_argvalidation, test_gap_constants_tail_3683plus, test_gap_handle_band_object_ops.

Related fragility worth fixing while in here: the compile-object cache living under /private/tmp makes every reboot a cold start and couples cache identity to a world-writable path.

proggeramlug:
Verified against current  with a fresh object cache and a genuinely cold  profile rebuild:  rebuilt  runtime/stdlib, linked, wrote the executable, and ran successfully. PR #7021 fixed the zlib surface plus the panic-runtime dedup holes that caused this report. Closing as resolved.

proggeramlug:
[...]

Issue #6477, opened 2026-07-16
Concurrency race: Next.js byte-stream pull runs before Flight state is ready (null.byteLength)


proggeramlug:
### Summary

Under concurrent request load, perry runs a Next.js app's byte-stream `pull()` callback at a point where the Flight runtime's internal state isn't ready, so the pull hits a cold path that reads `.byteLength` on `null`/`undefined`. This is a microtask-ordering / reentrancy divergence from Node (same class as the #6360 tick-parity work) — Node runs the same pull with the state populated.

With #6476 the throw no longer crashes the process (the stream errors and the request 500s), but the *underlying* null still makes that request fail where Node returns 200.

### Repro

- App: a Next.js 16.1.1 standalone server (RSC/Flight, `type:"bytes"` render stream), compiled with perry.
- `PERRY_GC_FORCE_EVACUATE=1` + **concurrent** load (≈8 parallel requests to a page route) triggers it within 1–2 rounds. Single / sequential requests are far less likely to hit it. A single request to any route is fine; the deterministic tee-drain / `Buffer.from` / reader paths are all byte-identical to Node in isolation.

### Diagnosis (symbolized backtrace, `PERRY_DEBUG_SYMBOLS=1` + `PERRY_TRACE_THROW=1`)

```
js_throw_type_error_property_access            (Cannot read properties of null (reading 'byteLength'))
 → <Next Flight closures: __284.cold.7 → __284 → __544 → __552 → __609>   (the pull callback chain)
 → perry_stdlib::streams::readable_pull_microtask   (js_closure_call1 → the pull cb)
 → run_microtasks → js_callback_timer_tick → run_microtasks → js_node_http_server_process_pending → main
```

The `.byteLength`-on-null is **internal to Next's pull** (a cold/error path), not a chunk perry delivers — instrumenting `push_chunk` / `tee_deliver` / reader-read / `Buffer.from` for null never fires, and `PERRY_GC_VERIFY_EVACUATION=1` stays clean (not a stale-pointer / GC-rewrite miss). So perry is invoking the pull microtask when Next's per-request/module state is null — a scheduling divergence.

### Suspected area

`streams::maybe_pull_inner` pull scheduling vs Node's — likely perry fires the byte-pull relative to Next's render microtasks at a point Node defers past. Compare when perry vs Node schedules/runs the pull microtask across the Flight render pipeline (cf. the #6360 promise-hop/stream-cadence harnesses).

### Impact

[...]

Issue #6364, opened 2026-07-13
runtime: DisposableStack / Symbol.dispose surface incomplete (.disposed returns undefined)


proggeramlug:
## Summary

`DisposableStack` / `Symbol.dispose` support is incomplete: `test_gap_disposablestack_2875` returns `undefined` where node returns `false` / `true`.

This is an **untriaged gap-suite failure on `main`** — it is not in `test-parity/known_failures.json`.

## Repro

`test-files/test_gap_disposablestack_2875.ts` on current `main` (`be5ed2bfc`) vs `node --experimental-strip-types`:

```
line 5:   perry: undefined     node: false
line 7:   perry: undefined     node: true
```

Looks like the `.disposed` getter (and/or the adopt/defer surface) is missing rather than the whole feature.

## Why CI never caught it

`DisposableStack` and `Symbol.dispose` are **Node 24+**. `conformance-smoke` pins **`node-version: '22'`** (`.github/workflows/test.yml`), where *node itself* fails the test — so the harness classifies it as `node_fail` and **excludes it entirely**. The gap is structurally invisible to the required gate.

That is the more important point: this issue is a symptom. Anything Perry gets wrong about a Node 24/26 feature cannot be seen by CI at all, by construction. Tracked separately as the Node-version skew issue.

## Decision needed

Either implement the missing surface, or add a triaged entry to `test-parity/known_failures.json` with a category and reason — right now it sits in the gap that the no-new-untriaged gate exists to prevent.

PR #7599, opened 2026-08-07
test(parity): make known_failures.json a ratchet, not a suppression list (#7582)

### LINKED ISSUE #7582: test-parity: known_failures.json is a pure suppression list — a stale entry hid a data-loss bug for six days
## The problem

`test-parity/known_failures.json` is a **pure suppression list**.
`parity_known_failures.py` computes `failures - allowed`, so an entry whose test
has started passing is simply inert. It never fails, never reports, and never
asks to be removed.

That is a fifth way a gate can be unable to fail, alongside the four already
named in CLAUDE.md — and it is the most insidious, because unlike those four the
job is genuinely green *and* genuinely running. The entry silently converts a
future regression of that exact test into a non-event, forever.

## It just cost us a real bug

`test_gap_diagchannel_3082_3084_3085_3086` was added to `known_failures.json` on
2026-07-04 for the original `#3082/#3084/#3085/#3086` feature cluster. Those
features were implemented; the entry stayed. When **#7105** (2026-08-01) broke
the test again for a completely unrelated reason — `PreallocateBoxes` shadowing a
module-level global — the suppression absorbed it.

The impact was not one gap test. The defect silently emptied **every
`let`/`const` in a top-level bare block of an ES module** that a sibling
`function` declaration read. Nothing threw; the binding just read back as
whatever it was defined with. It sat for six days and was found by accident.
Fixed in #7580.

Compounding it, `parity` is tag-gated (CLAUDE.md), so even an unsuppressed
failure there would not have blocked a merge.

## The fix that already exists in this repo

`scripts/gc_root_dominance_allowlist.json` solves exactly this, and its own
header states the rule:

> an entry that matches nothing FAILS the build. When the bug is fixed, delete
> the entry in the same PR. That is the ratchet.

`known_failures.json` should adopt the same property: **an entry whose test now
passes fails the run**, with a message naming the entry to delete. That converts
the list from a place bugs go to die into a ratchet that empties itself.

Two things worth doing at the same time:

- Adopt the allowlist's provenance discipline — issue number and date per entry.
  #797 already tracks this for `known_failures.json` and has been open a while;
  the two changes are the same edit and should land together.
- Audit the existing entries for others that have started passing. Every one is
  a live suppression of a test nobody is watching, and this exercise just found
  one that was hiding a data-loss bug.

## Care needed on the flip

Turning this on will go red immediately for every already-passing entry, which
is the point but also the reason it cannot be flipped blind — CLAUDE.md's
corollary applies: *a new gate has never been green, so promoting it to required
immediately blocks every open PR.* Run it once, empty the entries it names in
the same PR, then promote. Note also that a test can pass on one platform and
fail on another, so the "now passes" verdict has to be per the platform the gate
runs on, not asserted globally.

proggeramlug:
Closes #7582. Carries out the audit #797 asked for on this file, and puts its provenance rules in code so the next one is mechanical.

## The problem

`test-parity/known_failures.json` was a pure suppression list. `scripts/parity_known_failures.py` computed `failures - allowed` and stopped, so an entry whose test had started passing was inert forever — it never failed, never reported, and never asked to be removed, while silently converting that test's next regression into a non-event.

That is a fifth way a gate can be unable to fail, alongside CLAUDE.md's four, and the worst of them: the job is genuinely green **and** genuinely running.

`test_gap_diagchannel_3082_3084_3085_3086` was listed on 2026-07-04 for a feature cluster that had shipped. When #7105 broke the test again for an unrelated reason (`PreallocateBoxes` shadowing a module-level global) the entry absorbed it. The defect emptied every `let`/`const` in an ES module's top-level bare block that a sibling `function` declaration read — six days, found by accident, fixed in #7580.

## The mechanism

Data flow, traced rather than assumed. The file has exactly two code readers: `scripts/parity_known_failures.py` (the gate, run from the **tag-gated** `parity` job and `--self-test` from `lint`) and `scripts/parity_matrix_trend.py` (imports `normalize_platform`, re-derives platform selection to mark records `known`; same tag-gated job). `scripts/run_gap_tests.sh` does **not** read it — the gap gate is `gap_snapshot.json`, which has been bidirectional since #6755.

**Live half.** `check()` now also reads the report's `results[]` and fails on any allowlist entry selected for this platform whose test ran and returned `pass`, naming the entry to delete. `results[]` is mandatory rather than best-effort, so the gate cannot silently degrade back to suppression-only, and a passing run prints how many entries it actually adjudicated (`5/10` above) — a gate must assert its subject was live.

"Now passes" needs positive evidence and is per platform. Not in `results[]` at all (filtered out, not in this shard), `node_fail`, `skipped`, or scoped elsewhere via `platforms` is **never** flagged. Absence of evidence is not a pass — that is the hole a Node-22 pin used to hide 14 tests in (#6364).

**Offline half — `--audit`, new, on `lint`.** The `parity` job is tag-gated, so the live half fires *after* the merges it was meant to judge. `--audit` needs no parity run: it enforces provenance (issue + date + a category from the documented enum), rejects an entry naming a test file that no longer exists, and cross-checks every `test_gap_*` entry against `gap_snapshot.json` — generated, bidirectional, regenerated 2026-08-06, enforced by required CI on every PR. An entry absent from that snapshot is one the snapshot asserts passes. **This check would have named the diagchannel entry on 2026-07-04, the day it was added.** It runs in ~0.1s.

Schema tightening in the same pass: `issue` and `added` were previously unvalidated, and `category` was only checked for non-emptiness — which is how `category: "toolchain"` (five entries, undocumented) slipped in unnoticed.

## The audit

37 entries → 10. **Every tracking issue in the file except #6477 is closed.**

**27 retired.** All are gap-suite entries the Linux `gap_snapshot.json` asserts pass — Linux is the platform this file's consumer runs on. Corroborated live on macOS: a full 500-test gap run, plus a second auto-optimize pass over the ten whose first verdict came from a run mode that could not link them.

A read of all 28 candidate test bodies says 26 are substantive (real values, real round-trips, spec-mandated throw messages) — they pass because the feature works. Two are narrow: `test_gap_console_bare_global` and `test_gap_readline_3698plus` assert mostly `typeof`/shape. That is on-target for their specific defects (`console` as a bare identifier resolving to the `0` sentinel; `createInterface`/`connect` named exports resolving to `undefined`) but shallow beyond them.

**The audit found a live reproduction of a closed issue.** `test_gap_zlib_4917_level` looked stale from the Linux snapshot and was retired with the rest — then reproduced #6847's exact signature on macOS/arm64 in auto-optimize mode, **3 runs out of 3**: `Undefined symbols: _js_zlib_deflate_raw_sync, _js_zlib_inflate_raw_sync`, auto-opt pairing the ext-zlib provider archive with a feature-stripped stdlib rebuild. Its two zlib siblings pass in the same mode, so it is specific to the raw-sync entry points. **#6847 was closed COMPLETED on 2026-07-30 (fixed by #7021, verified on a cold cache) and needs reopening for macOS.**

It is kept with `platforms: ["macos"]`, not globally. Linux passes the test, so an unscoped entry would go on suppressing a passing test on the platform the gate runs on — the exact failure mode this PR removes. Scoped, it suppresses nothing in CI, records something true, and the ratchet still governs it.

**10 survivors, provenance refreshed.** The four that are bug claims with a closed tracker (#2159, #2514, #3088, #3680) plus the macOS zlib entry (#6847) are re-categorised `bug-stale` — the category the schema already defines for exactly this — with the closed-tracker fact leading the reason. `test_gap_stream_tee_tick_parity` keeps `bug-open` (#6477 is the one open tracker). The `module-inventory` survivors (`test_parity_stream`, `test_parity_stream_web`, #793) and `ci-env` survivors (`test_sock_write_map`, `test_ramda_sum`, #1634) keep their categories, because `bug-stale` would assert something false about them, and carry the same note. `issue` still points at the original tracker — that is its provenance, not a claim that it is open. All five need new tracking issues, now visible instead of buried.

The four survivors adjudicated by the live run all still fail on macOS, three of them with the same status the Linux snapshot records. The four non-gap survivors did not run under a `test_gap_` filter and were correctly left un-adjudicated.

## Sabotage evidence

The gate must be able to fail. Both halves, real exit codes, no pipe (a wrapper's status is not the harness's — that has bitten this repo three times this week).

Offline half, against the committed files:

```
--audit, current file                       → exit 0
re-add test_gap_symbols (a passing test)    → exit 1, "STALE ... test_gap_symbols:
                                              gap_snapshot.json (the generated linux
                                              baseline) says this test PASSES"
remove it again                             → exit 0
```

Live half, against the real macOS gap report projected onto the audit set:

```
ARM1  current allowlist                     → exit 0  ("5/10 adjudicated")
ARM2  re-add test_gap_symbols, which PASSED
      in that same run                      → exit 1, names test_gap_symbols
ARM3  remove it again                       → exit 0
ARM4  same entry, platforms: ["linux"]      → exit 0  (not judged on macos)
```

ARM4 is the one that matters for #7568's per-platform rule: a passing test whose entry is scoped elsewhere must not be flagged.

`--self-test` gained arms for all of it: stale detection, `node_fail`/`skipped`/`compile_fail` not counting as passes, a test that did not run not counting as a pass, a platform-scoped entry not being judged, a report with no `results[]` being rejected outright, and each provenance field failing on its own.

## Validation

Local, and local is what counts here — CI has a deep backlog and none of this has run there yet.

- `parity_known_failures.py --self-test` → 0; `--audit` → 0; `gap_snapshot.py --self-test` → 0 (unchanged neighbour); `parity_matrix_trend.py` still imports cleanly (it consumes `normalize_platform`, which is untouched).
- Full gap suite on macOS, 500 tests, oracle `node v26.5.1` matching `.node-version`. (Worth flagging: CLAUDE.md's prose still says 26.5.0 while `.node-version` — which CLAUDE.md itself names as the single source of truth — says 26.5.1. The file is right; the prose is stale.)
- Ten ext-routed tests re-run individually in auto-optimize mode, and the one failure re-run twice more.
- `scripts/check_file_size.sh` → 0. `scripts/raw_handle_debt.py` → 998, baseline 998. No Rust touched, so no `cargo fmt`. Workflow YAML parses.

Two things seen in passing, not addressed here: `test_gap_gc_rest_argument_rooting` and `test_gap_gc_same_module_call_argument_rooting` crashed on macOS under `PERRY_NO_AUTO_OPTIMIZE=1` while the Linux snapshot has them passing, and that mode produces ~58 spurious ext-link `compile_fail`s on macOS despite `run_parity_tests.sh`'s comment claiming its ext-package build compensates.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved parity validation to detect both newly disallowed failures and outdated known-failure entries.
  - Added stronger validation for test results, platform scope, issue references, dates, and test-file presence.
  - Removed obsolete known-failure records and updated platform-specific classifications.

- **Documentation**
  - Expanded guidance for parity checks, audits, failure evidence, and version pinning.
  - Documented the latest release and Node.js parity baseline.

- **Chores**
  - Updated the project version to 0.5.1345.


proggeramlug:
## Audit before merge — verified, merged as v0.5.1345

**Sabotage re-verified with the historically-loaded case**: re-adding the exact
entry that hid #7580 (`test_gap_diagchannel_3082_3084_3085_3086`) makes
`--audit` exit 1 naming it, with the right prescription ("delete in the SAME
PR"); restored, exit 0. Both self-tests green; the `lint` wiring
(`--self-test` + `--audit`) confirmed in test.yml.

The design detail that earns the merge: the **offline `--audit` on per-PR
`lint`** is what makes this a real gate. The live half runs in the tag-gated
`parity` job — after the merges it should judge — so without the offline
cross-check against `gap_snapshot.json`, the ratchet would fire at tag time,
which is the same too-late failure mode that let #7580 sit for six days. As
built, the diagchannel entry would have been named the day it was added.

Also folded into the merge commit: CLAUDE.md's oracle prose said Node `26.5.0`
while `.node-version` says `26.5.1` — the file is authoritative, prose fixed,
with a note that it drifted once.

Two findings from the audit actioned separately: **#6847 reopened** (live macOS
reproduction, 3/3, exact original symbol signature — caught precisely because
the audit refused to retire a "stale-looking" entry without re-running it), and
the two macOS-only gc-rooting gap crashes are filed next.

The 37 → 10 retirement with per-entry provenance and pass-reason ("substantive"
vs "typeof-shallow") is the model for what #797 asked for. The two
typeof-shallow survivors are honestly labelled, which is worth more than
retiring them would have been.

PR #7591, opened 2026-08-07
fix(codegen): a typed-array store used as an expression evaluated to 0 (#7590)

proggeramlug:
Fixes #7590.

## The bug

```ts
const buf = new Uint8Array(4);
sink((buf[0] = 5), 5);           // was 0, now 5
n = buf[1] = 7;                  // was 0, now 7
sink((buf[2] = 3) + 100, 103);   // was 100, now 103
```

The stores always landed correctly (`buf 5 7 3` both before and after) — only the expression's *value* was wrong. So it was silent: a wrong number, no crash, no diagnostic. An assignment expression must evaluate to the assigned value (ES2024 §13.15.2).

## Cause

`ctx.discard_expr_value` means **"this STATEMENT's value is discarded"**. It is set once per `Stmt::Expr`, and `lower_expr` never cleared it while recursing — the only reset in the tree is `lower_call/new_ctor_args.rs`, for constructor arguments. So it was still set while lowering the *operands* of `sink(buf[0] = 5);`.

Four sites read it as though it meant "this EXPRESSION's value is discarded" and returned `double_literal(0.0)`:

- `expr/index_set.rs` — `lower_typed_array_store`, proven-view checked store
- `expr/arrays_finds.rs` — `Uint8ArraySet` / buffer stores

`expr/dispatch.rs` also reads the flag but only to pick a materialization path, so it is unaffected and left alone.

## Fix

`FnCtx::discard_this_expr`, which `dispatch::lower_expr` **takes** (`mem::take`) at the top of every dispatch. It therefore reaches exactly one expression — the one the statement is made of — and every operand lowered beneath it reads `false`.

The handlers receive it as a **parameter** rather than reading the field. That is deliberate: they consult the answer *after* lowering their operands, by which point the field has been taken again, so reading a field there would have reintroduced the same bug in a subtler form.

## How it was found

I was gating `arr.push(x)`'s length computation on the same flag as a performance change. `js_array_length` is not a field read — it resolves Proxy arrays through the `get` trap and probes the registered-Set/Map side tables — and a statement-position push discards its result, so it is 8–13% of `push_cls` spent producing a number nobody reads (see #7511).

That optimisation produced exactly this bug for `sink(a.push(10))`, `n = a.push(20)`, `a.push(1) + 100` and `a.push(1) > 0 ? 7 : 9` — which is what led me to test the pre-existing sites and find they had it already.

[...]

Issue #7590, opened 2026-08-07
fix(codegen): a typed-array element store used as an expression evaluates to 0 — discard_expr_value leaks into operand position


proggeramlug:
## Summary

A typed-array element store used as an **expression** evaluates to `0` instead of the assigned value, whenever it appears anywhere inside an expression statement. The store itself is correct — only the expression's value is wrong. Silent: no crash, no diagnostic, just a wrong number.

```ts
const buf = new Uint8Array(4);
function sink(label: string, got: number, want: number): void {
  if (got !== want) console.log("WRONG", label, "got", got, "want", want);
}

sink("store-arg", (buf[0] = 5), 5);            // consumed as a call argument
let n = 0;
n = buf[1] = 7;                                 // consumed by an assignment
sink("store-assign", n, 7);
sink("store-binary", (buf[2] = 3) + 100, 103);  // consumed by arithmetic

console.log("buf", buf[0], buf[1], buf[2]);
```

| | output |
|---|---|
| node | `buf 5 7 3` |
| perry (`d2dca5823`) | `WRONG store-arg got 0 want 5`<br>`WRONG store-assign got 0 want 7`<br>`WRONG store-binary got 100 want 103`<br>`buf 5 7 3` |

`buf 5 7 3` on both, so the stores land correctly. An assignment expression must evaluate to the assigned value (ES2024 §13.15.2); here it evaluates to `0`.

## Cause

`ctx.discard_expr_value` means **"this STATEMENT's value is discarded"**, and `lower_expr` does not clear it when recursing into subexpressions — the only reset in the tree is `lower_call/new_ctor_args.rs:252`, for constructor arguments. So while lowering the operands of `sink(buf[0] = 5);` the flag is still set from the enclosing `Stmt::Expr`.

Four sites read it as though it meant "this EXPRESSION's value is discarded" and return `double_literal(0.0)`:

- `expr/index_set.rs:754` — `lower_typed_array_store`
- `expr/index_set.rs:769` — proven-view checked store
- `expr/index_set.rs:848`
- `expr/arrays_finds.rs:788`, `:824`, `:895`, `:912` — `Uint8ArraySet` / buffer stores

`expr/dispatch.rs:25` also reads the flag, but only to pick a materialization path, so it is not affected.

## Scope

Needs a typed array or buffer view — plain `arr[i] = v` on an ordinary array does not take these paths, so this is narrower than "all index assignment". Every affected site is a store whose value would normally be materialized for the caller.

[...]

PR #7600, opened 2026-08-07
perf(codegen): elide the length computation of a statement-position push

proggeramlug:
Follow-up to #7590/#7591 — the perf half that was deliberately kept out of the correctness fix.

## What

`arr.push(x)` evaluates to the new length, computed by `js_array_length` — which is **not a field read**: it resolves Proxy arrays through the `get` trap and probes the registered-Set/Map side tables. A statement-position `arr.push(x);` discards that result, so push-heavy code pays an out-of-line runtime call per push for a number nobody reads (8–13% of push-heavy runs; `json_pipeline`'s `build_out` loop is exactly this shape — 1 source-level `.length` but 458 emitted call sites, one inside the hot loop ⇒ 20M calls).

This PR elides the length computation when — and only when — the push is the statement's own expression.

## Why this is safe now (and was not before #7591)

The gate is the `mem::take`n per-expression signal `dispatch::lower_expr` introduced in #7591. It reaches **exactly the statement's own expression and never an operand**, so `n = arr.push(x)`, `sink(a.push(10))`, `a.push(1) + 100`, `a.push(1) > 0 ? 7 : 9` all compute the real length. Gating on the raw `ctx.discard_expr_value` field is precisely the #7590 bug — it produced wrong values in all four of those positions when I first tried this as a perf change, which is what led to finding #7590 in the first place.

All eight push/push-spread return points receive the signal **as a parameter**, not a field read — handlers consult it after lowering their operands, by which point the field has been taken again.

## Verified live in the IR

`--trace llvm` on a probe with a discarded loop push + one consumed push: the discarded push emits **zero** `js_array_length` calls; the consumed push exactly one per function specialization. (A green behavior test alone would be vacuous — the discarded form works whether or not the elision fires.)

## Measurements

Interleaved A/B, each arm linked by its own compiler+runtime pair (separate target dirs, identical package sets, both runtimes current `main`):

**Pure statement-position push loop** (20M pushes, runtime-dependent bounds, escaping array):

| arm | best-of-6 | per push |
|--|--:|--:|
| main | 180 ms | 9.0 ns |
| this PR | **65 ms** | **3.25 ns** |

**2.77×** — more than the call itself: the out-of-line call was also blocking loop optimization around it. Both arms compute identical results.

**`json_pipeline` (shipped workload, 200k records)**: 2,495 → 2,423 ms total (~3 %), output hash identical — modest because that workload is GC-bound (#7592); its statement-position `out.push({...})` is exactly the elided shape.

## Testing

`test-files/test_array_push_expression_value.ts` consumes the push value in call-argument, assignment, arithmetic, conditional, nested-push, object-element, spread, and boxed-array (runtime fall-through path) position, plus discarded forms that must keep pushing. **Byte-identical to `node --experimental-strip-types`.**

That matrix is the point (#7590's lesson): the discarded form kept working the entire time the original bug was live, so only a consuming test can catch a signal leak.

- `perry-codegen` test suite: 8 targets, all green (671 tests in the main target), 0 failures.
- Gap suite: running at time of filing; result will be posted as a comment.
- `cargo fmt --check`, `scripts/check_file_size.sh` clean.

## Summary by CodeRabbit

* **Performance**
  * Improved generated code for discarded `array.push()` results by avoiding unnecessary length calculations.
  * Preserved accurate returned lengths when `push()` results are used.

* **Bug Fixes**
  * Ensured array mutations continue to work correctly across standard, spread, object, and boxed arrays.

* **Tests**
  * Added coverage for `push()` in assignments, conditions, arithmetic, function calls, and other expression contexts.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1341
+**Current Version:** 0.5.1342
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1341"
+version = "0.5.1342"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1341"
+version = "0.5.1342"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1341"
+version = "0.5.1342"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1341"
+version = "0.5.1342"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1341"
+version = "0.5.1342"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1341"
+version = "0.5.1342"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1341"
+version = "0.5.1342"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1341"
+version = "0.5.1342"
 dependencies = [
  "anyhow",
  "perry-hir",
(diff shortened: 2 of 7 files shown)


proggeramlug:
## Audit before merge — verified, merged as v0.5.1342

**IR mechanism confirmed on my own probe** (discarded loop push + one consumed
push + one `.length` read): main emits **4** `js_array_length` calls, the PR
emits **2** — the 20M-iteration discarded push emits none, both consumed sites
keep theirs. Matrix test byte-identical to node.

**The decisive sabotage is the one that recreates #7590, and the matrix test
catches it.** Regressing the gate to the leaking statement-scoped
`ctx.discard_expr_value` field produces exactly the original bug signature:

```
assign:WRONG got=0 want=2 ... obj_assign:WRONG got=0 want=2
```

That is the point of a consuming test matrix — the discarded form worked the
entire time #7590 was live, so only value-consuming positions can catch a
signal leak. The gate arriving as a **parameter** to all eight return points
(never a field read after operand lowering) is the discipline #7591's audit
asked for, applied.

My probe showed 1.08× rather than 2.77× — expected: it grows a 20M-element
array, so allocation/GC dominates, and #7602 just made the barrier half of that
cheaper. The 2.77× is the purpose-built pure-push shape; the IR census is the
load-bearing evidence and it is unambiguous.

Note for the record: the promised gap-suite comment was not posted (same as
#7596). My own validation stands in: full `perry-codegen` 677/0 and
`perry-runtime` 1851/0 suites, matrix test vs node, all four lint gates + fmt
clean. The elision is unreachable for Proxy receivers (the typed lowering
requires an array proof), so no observable trap behaviour changes.

proggeramlug:
## Gap suite result (as promised at filing)

Full run on this branch (which also carries the merged #7594/#7596 GC pacing changes, so this doubles as their post-merge gap verification — the run at #7596 filing time was voided by a deleted worktree):

- **No real regressions.**
- The one reported regression, `test_gap_zlib_4917_level` `pass → compile_fail`, is the documented **host-local runtime-dir resolution flake**, not a code regression: the missing `js_zlib_*_raw_sync` symbols are present in the freshly built `libperry_stdlib.a` (verified with `nm` on both A/B arms), and the same test **compiles and matches node exactly** when linked with `PERRY_RUNTIME_DIR` pinned to that archive.
- 10 `node_fail → parity_fail` status changes are oracle-coverage transitions — tests the pinned Node (26.5.1, matching `.node-version`) now runs that it previously couldn't. They surface pre-existing Perry gaps, not changes in Perry behavior; the harness classifies them outside the regression gate for exactly this reason.
- 1 improvement: `test_gap_iterator_helpers_2874` `parity_fail → pass`.

PR #7536, opened 2026-08-06
perf(gc): decide a class-field store's GC bookkeeping with one inline test, not three calls (#7511)

proggeramlug:
Part of #7511.

## Why #5334 lever D is not firing on `churn_alloc` (the actual answer)

The ticket assumes the barrier is removable because "the types say `v: number`".
It is not: a field declared `number` legitimately receives a string through an
`any` (CLAUDE.md, "No runtime type *validation*"), so the annotation is never a
layout fact. Elision has to be by construction — and lever D
(`expr_produces_non_pointer_bits_by_construction`) is exactly that predicate.
So why doesn't it fire?

Not because the proof is too narrow. **The store is not at the `new` site at
all.**

Since the `[#bloat]` `force_ctor_call` default (`crates/perry-codegen/src/lower_call/new.rs:989`),
a class with its own constructor is **not inlined**: its body is compiled once
as the shared `<class>_constructor(this, p0, …)` symbol and *called*. HIR
rewrites every closed-shape object literal into a `New` of a synthesized
anon-shape class with a `constructor(v, w) { this.v = v; this.w = w }` body
(`crates/perry-hir/src/lower/context.rs::mint_anon_shape_class`), so
`{ v: base + j, w: j }` lands there too. Verified on the emitted IR — `chunk`
contains

```llvm
%r42 = call double @churn_alloc_ts____AnonShape_c42fad28a2551dcc_constructor(double %r39, double %r41, double %r29)
```

and the two field stores live inside that callee, where the value expression is
`Expr::LocalGet(<ctor param>)` — an LLVM **function argument** of a function
shared by every `new` site in the module, including ones that pass pointers.
`local_get_produces_non_pointer_bits_by_dataflow`
(`crates/perry-codegen/src/expr/helpers.rs:46`) admits only `i32_counter_slots` /
`integer_locals` members; a parameter is in neither, and no per-call-site fact
can be attached to a shared symbol without cloning it per call site (i.e.
undoing `#bloat`).

So this is neither "extend an existing proof" nor "the proof is right and the
path is wrong". **At that site no by-construction proof about the value can
exist**, which is why the answer here is #7501's shape: a live test at the
store.

I also implemented and then **reverted** the obvious static extension
(propagating each `new`-site argument's lever-D verdict across
`bind_inline_constructor_params` into the inlined constructor's parameters). It
is sound, but on the default path `bind_inline_constructor_params` is only
reached for the no-own-ctor / recursive / capture-collision cases — it would
[...]

Issue #7511, opened 2026-08-06
perf(gc): write barriers cost 16% on an all-numeric store workload — elide on provably-non-pointer stores


proggeramlug:
## Summary

Write barriers cost **16.1% of self time** on `gc-handoff/bench/churn_alloc.ts` — a
program that stores **only doubles**. Every barrier on that benchmark is wasted work: the
stored values are `number`-typed fields that can never hold a GC pointer, and the types
say so at compile time.

For scale, on the same profile the actual allocation is 7.7%. Barriers cost more than
twice what allocating costs.

## Evidence

Symbolicated profile (`PERRY_DEBUG_SYMBOLS=1`, `sample` 2 s, 1500 leaf samples, quiet M1
mini at load 1.4) of `churn_alloc.ts` — 20M `{v: number, w: number}` literals pushed into
an array:

| symbol | share |
|---|--:|
| `barrier_child_prologue` (`gc/barrier.rs:1042`) | 8.9% |
| `write_barrier_*` | 3.2% |
| `js_write_barrier_slot` | 2.9% |
| `incremental_mark_barrier_*` | 1.1% |
| **barrier total** | **16.1%** |
| *(for comparison)* `js_object_alloc_class_inline_keys` + `arena_alloc*` | 7.7% |

The benchmark's only stores are `{ v: base + j, w: j }` — two f64 fields — plus the array
push. No pointer is ever written into an object slot.

## Why it is not already elided

#7486 ("elide provably-dead per-store bookkeeping on class-field stores") landed and did
not cover this. The clearest signal is that the **class** form is *worse* than the object
literal form on the identical workload:

| variant | Perry | node | ratio |
|---|--:|--:|--:|
| churn_alloc — `{v, w}` object literal | 2.44 s | 0.14 s | 17.4× |
| push_cls — `new Node(v, w)`, both fields `number` | **3.99 s** | 0.14 s | **28.5×** |

A fixed-shape class with two declared `number` fields is the most statically-known
construction form in the language, and it is the slowest. That anomaly is **#7512**; it is called out here because it is evidence that the existing elision is not
reaching declared-primitive fields.

## Repro

```bash
cd gc-handoff/bench
export PERRY_RUNTIME_DIR=<repo>/target/release
PERRY_DEBUG_SYMBOLS=1 PERRY_NO_AUTO_OPTIMIZE=1 <repo>/target/release/perry \
    churn_alloc.ts -o sym_churn_alloc

[...]

PR #7501, opened 2026-08-06
perf(gc): declare array element layout once at allocation instead of per push

proggeramlug:
Part of the #7469 mutator-cost campaign. An `out = []` + `out.push({...})` loop
maintained its array's GC pointer mask **per store**, paying two runtime calls on
every iteration of the hot inline push. This declares the element layout **once,
at the allocation site**, and drops both.

## The cost being removed

Per pointer-valued push, the inline store block emitted:

- `js_gc_note_slot_layout` — reaches into the `LAYOUT_SLOT_MASKS` thread-local
  hashmap to set one bit of a per-array bitmap;
- `js_array_note_numeric_write` — clears a raw-f64 flag that is already clear.

The mask entry the first push creates also makes `LAYOUT_SLOT_MASKS` non-empty
for the rest of the process, so the `is_empty()` fast-out in
`layout_forget_object` — probed on **every** object allocation — starts hashing
instead. That second-order cost is why the entry is worth not creating at all,
not merely worth creating more cheaply.

## What is emitted instead

At the `[]` literal, one `js_array_declare_all_pointer_elements(arr)`: it clears
the raw-f64 numeric layout `js_array_alloc` published and swaps
`GC_LAYOUT_POINTER_FREE` for `GC_LAYOUT_SIDE_MASK | GC_LAYOUT_ALL_POINTERS`. For
the collector that is "visit every slot in `0..length`" — the same set of slots
`GC_LAYOUT_UNKNOWN` visits, so the declaration is conservative in the only
direction that matters (`mark_field_into_worklist` re-validates every word, so a
non-pointer element costs a rejected read, never a stranded child).

## Soundness does not rest on the static proof

Every elided store is fronted by a test of the **live** header, folded into the
integrity test that was already there — same `_reserved` load, same one `and` +
one `icmp`, no added instruction:

```llvm
%r64 = and i16 %r63, 62599     ; 0xF487
%r65 = icmp eq i16 %r64, 40960 ; 0xA000
```

`0xF487` = the four integrity bits (frozen / sealed / non-extensible /
array-descriptors, which must be 0, exactly as before) **plus**
`GC_ARRAY_RAW_F64_LAYOUT | GC_ARRAY_RAW_F64_HOLES` (must be 0) and the layout
state + `GC_LAYOUT_ALL_POINTERS` (must be `SIDE_MASK | ALL_POINTERS`). The two
raw-f64 bits are what makes eliding `js_array_note_numeric_write` sound — with
[...]

Issue #7187, opened 2026-08-01
gc/perf: classify_heap_generation is 19% of batch.ts (57.4% total GC bookkeeping) with ZERO collections running — write-barrier tower needs its own lever (#5094 evidence)


proggeramlug:
The batch.ts ranked profile (Pi 5, `perf record` instructions:u, THROTTLE==0 asserted, 99.85% coverage; full account on #7170) found:

**57.4% of the program is GC bookkeeping — and zero collections run** (`PERRY_GC_DIAG` silent, no mark/sweep symbol appears). The largest single symbol is **`classify_heap_generation` at 19.03% (657M instructions)** — 2.3× the next symbol — called from the write barrier on the sort's own element stores. The layout side-table machinery adds another 12.91%.

Per comparison: ~2,597 instructions, of which sort *scaffolding* (dominated by barrier/layout bookkeeping) costs 1,495 — more than the user's comparator logic (1,103).

## Why this needs its own lever

The planned levers were mapped against the profile: #6759 (shape tree) squarely hits the 12.91% layout side-table; #7151/#7170-R1 hit generated-code costs. **Nothing planned addresses rank 1.** A write barrier that spends 19% of the program classifying heap generations for a heap that has never collected is pure overhead by construction.

Candidate directions (to be designed, not prescribed):
- **Lazy barrier arming**: until the first minor collection, there is no young/old distinction worth recording — the remembered set can start empty and barriers can no-op behind a single global flag flip at first-collection time. (#7166's census machinery can verify the arm goes live when collections start — the gate-must-be-live lesson.)
- **Cheaper classification**: if `classify_heap_generation` is a multi-branch/side-table walk, an address-range check or a header bit may collapse it.
- Relation to **#5094** (umbrella: make per-object layout O(1)-loadable, kill per-operation thread-local layout tracking) — this measurement is that issue's strongest evidence to date, with the 12.91% side-table figure attached. And #6893's O(objects) memory concern is the same machinery.

## Also from the same profile

- `utf16_cmp_bytes` (`string/compare.rs:15`) runs `from_utf8` validation + UTF-16 re-encode on BOTH operands per comparison — an ASCII fast path deletes **~3.1% of total program instructions** (separate, trivial PR).
- The vdispatch-tower lever is NOT supported by this workload: comparator dispatch resolves a direct fn pointer once (`array/sort.rs:23`); total closure dispatch 0.08%.

proggeramlug:
[...]

PR #7602, opened 2026-08-07
perf(gc): gate the array-push write barrier on a live parent-generation test (#7511)

proggeramlug:
Closes part of #7511.

## Re-measured first — the headline had NOT collapsed

The ticket's 16.1% / ~25% figures predate #7536, #7594 and #7596, so this starts
with a fresh symbolicated profile at v0.5.1339 (`PERRY_DEBUG_SYMBOLS=1` + `sample`,
`push_cls.ts`, 2286 leaf samples under `main`):

| symbol | leaf samples | share |
|---|--:|--:|
| `js_write_barrier_slot` | 346 | 15.1% |
| `write_barrier_decoded_parent` | 161 | 7.0% |
| `barrier_child_prologue` | 139 | 6.1% |
| `incremental_mark_barrier_value` | 30 | 1.3% |
| **barrier total** | **676** | **29.6%** |

It had grown, not collapsed. All four symbols are the GC write barrier
(`gc/barrier.rs`), not the `Ptr<Shape>` promotion barrier — checked, per #7187.

**All of it is one call site.** The call graph names it: `perry_fn_push_cls_ts__chunk + 568`,
i.e. `keep.push(new Node(...))`. `push_cls_ts__Node_constructor` appears as a leaf
with *no* barrier beneath it — the constructor's two `number` field stores already
pay nothing, because a declared-`number` field selects the raw-f64 representation
and its store is guarded by an inline plain-finite check with a downgrading cold
fallback. That is why #7536 correctly measured `push_cls` unchanged.

## What #7536 already covered, and why it cannot reach this

#7536 put the class-field store's three bookkeeping calls behind one inline test of
the **stored value's** bits. `PERRY_GC_TRACE` shows why that test cannot fire here:

| bench | calls | `non_pointer_child_skips` | `parent_not_old_skips` | `old_to_young_slow_hits` | `new_inserts` |
|---|--:|--:|--:|--:|--:|
| `push_cls` | 19,945,222 | 0 | 19,743,573 (99.0%) | 0 | **0** |
| `churn_alloc` | 19,945,222 | 0 | 19,743,573 | 0 | **0** |
| `churn` | 19,945,222 | 0 | 19,743,573 | 0 | **0** |
| `tree` | 62,612,898 | 20,867,845 | 41,271,511 | 0 | **0** |
| `cycles` | 15,674,953 | 7,832,430 | 7,368,973 | 0 | **0** |
| `retain` | 6,304,687 | 0 | 0 | 3,204,922 | 6,268 |

`non_pointer_child_skips == 0`: the pushed value genuinely **is** a heap pointer, so
`expr_produces_non_pointer_bits_by_construction` is not merely unhelpful, it is
*correct* to say "pointer". And the remembered set is **never inserted into — not
once — in 20 million calls**. The waste is entirely **parent-side**.

#7536's own fragment hands off the `put.pic.hit` path as the follow-up; that turned
out to be the wrong lead for these benches (it is `PutValueSet`, and `push_cls`'s
field stores never reach a barrier at all), so this takes the lever the counters
actually point at.

## The change

A parent-side question admits no by-construction proof — `keep` crosses hundreds of
collections between its allocation and its last push, so "the parent is young" is
exactly the claim #7501 showed gets revoked at runtime. So it is decided by a **live
test at the store**:

```text
parent_may_need_remembering(parent) :=
      (header(parent).gc_flags & GC_FLAG_TENURED) != 0
   || PERRY_INCREMENTAL_MARK_BARRIER_ACTIVE_COUNT != 0
```

Both clauses are load-bearing, for unrelated reasons.

**TENURED.** The remembered set only ever needs an entry when
`barrier_parent_needs_remembering` classifies the parent `Old`. Soundness is the
superset property #7536 used: `Old ⟹ TENURED`, so `!TENURED ⟹ !Old`, and this gate
can only skip a subset of what the runtime already skips. Every path that places an
object into an old-gen block sets the bit in the same breath — `gc/copying.rs:612-637`
selects `arena_alloc_gc_old` and `GC_FLAG_TENURED` from one `promote` expression;
`gc/oldgen.rs:1740` and `:1838`; `buffer/header.rs:486`; `typedarray/mod.rs:722`;
`json_tape.rs` via `arena_alloc_gc_old_born_tenured` — and nothing clears it on a
live object. The stronger reading: a parent that is neither physically old nor
logically tenured is **fully traced by every minor GC** (`gc/trace.rs:747`), so its
edges are rediscovered.

**Incremental.** Skipping the call also skips `barrier_child_prologue`'s
`incremental_mark_barrier_value` — the insertion/SATB shading, which is not a
generational question and must never be dropped while a cycle is live. A zero count
*proves* this thread's `INCREMENTAL_MARK_BARRIER_VALID_PTRS` is null, because
`incremental_mark_barrier_enable` installs the thread-local **before** incrementing
the count, an ordering `gc/barrier.rs` already documents as load-bearing. This reuses
the exact gate `expr/shadow_inline.rs` and `expr/shadow_slot.rs` already emit for the
root shading barrier — no new runtime symbol.

The gate reads the array's header byte at `arr_handle - 7`, which the `nofwd` block
**already loads** for its forwarding test in the dominating block, so the whole thing
is one `and`, one `icmp`, one global load and an `or`. It is emitted only inside
`apush.inbounds`, reached only after that header test, so the dereference rests on a
validation the existing code already performs — the "NaN-boxed parent bits would be
dereferenced" hazard does not apply. The slot store stays unconditional and outside
the branch. Under `PERRY_WRITE_BARRIERS=0` nothing is emitted at all.

Emitted IR (real `push_cls.ts`, `--trace llvm`):

```llvm
  store double %r68, ptr %r103                    ; slot store: unconditional
  call void @js_string_addref_if_heap_string(double %r68)
  %r107 = load i8, ptr %r106                      ; arr_handle - 7
  %r108 = and i8 %r107, 32                        ; GC_FLAG_TENURED
  %r109 = icmp ne i8 %r108, 0
  %r110 = load atomic i32, ptr @PERRY_INCREMENTAL_MARK_BARRIER_ACTIVE_COUNT seq_cst
  %r111 = icmp ne i32 %r110, 0
  %r112 = or i1 %r109, %r111
  br i1 %r112, label %apush.barrier.21, label %apush.barrier.done.22
apush.barrier.21:
  call void @js_write_barrier_slot(i64 %r72, i64 %r102, i64 %r104)
```

## Effect

Barrier calls, same counters:

| bench | calls before | calls after | Δ |
|---|--:|--:|--:|
| `push_cls` | 19,945,222 | 119,674 | **−99.4%** |
| `churn_alloc` | 19,945,222 | 119,674 | −99.4% |
| `churn` | 19,945,222 | 119,674 | −99.4% |
| `push_num` | 19,584,064 | 117,506 | −99.4% |
| `tree` / `cycles` / `deeplist` | unchanged | unchanged | field stores, not array push |
| `retain` | 6,304,687 | 6,303,669 | −1,018 (all `unarmed_skips`) |

**`retain`'s real work is bit-identical**: `old_to_young_slow_hits` 3,204,922 and
`new_inserts` 6,268 in *both* arms. Every genuine remembered-set insert is preserved.
GC cycle counts identical in both arms on every bench (`push_cls`/`churn_alloc`/`churn`
105, `tree` 42, `cycles` 16, `retain` 7, `deeplist` 4, `push_num` 18).

Timing — pinned quiet M1 mini, load ~1.5, best-of-5 `user+sys`, both arms linked
against the same runtime archive, baseline measured twice:

| bench | base run1 | base run2 | treat | speedup |
|---|--:|--:|--:|--:|
| `push_cls` | 0.670 | 0.650 | 0.490 | **1.33x** |
| `churn_alloc` | 0.680 | 0.660 | 0.510 | **1.29x** |
| `push_num` | 0.310 | 0.310 | 0.250 | **1.24x** |
| `churn` | 0.960 | 0.950 | 0.800 | **1.19x** |
| `cycles` | 0.850 | 0.970 | 0.820 | 1.04x (noisiest row) |
| `tree` | 8.470 | 8.270 | 8.160 | 1.01x |
| `retain` | 1.840 | 1.770 | 1.760 | 1.01x |
| `churn_read` | 0.360 | 0.350 | 0.350 | 1.00x |
| `deeplist` | 1.290 | 1.280 | 1.290 | 1.00x |

`deeplist` is the only row that could read as a regression; base run1 measured 1.290,
identical to treat, so it is noise. `tree`/`cycles`/`deeplist` gain little because
their barrier traffic is class-field stores, not array pushes — that is the honest
scope limit of this PR, and the remaining #7511 lever.

## Validation

Local only — CI has a deep backlog, so none of this is CI-confirmed.

- All 10 bench binaries produce **byte-identical stdout** with `rc=0`, including
  `cls_mistyped.ts` (the mistyped-`number`-field GC probe: 20000 strings survive).
- `PERRY_GC_VERIFY_EVACUATION=1 PERRY_GC_VERIFY_MARK=1` clean on both arms.
- `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_DIAG=1` clean on both arms,
  **instrument proven live**: 110 `[gc-fromspace-protect] mode=ProtectPages retired_set=#N`
  lines on `push_cls`, 12 on `retain` — copying minors really ran and from-space was
  quarantined and `mprotect`ed.
- `gc_root_dominance_corpus.sh` → 128/128 sources, 148 `.ll`. Both gated modes green:
  `--moving-only --seeded-violations 40` reports **40 planted, 40 caught, 0 missed**;
  `--unrooted-allocas` reports **0** over 7772 gc-capable allocas. Allowlist untouched.
- `cargo test -p perry-runtime --lib --no-fail-fast`: 1848 passed / 0 failed on 3 of 4
  runs. One run had a single failure I could not reproduce in three further runs —
  consistent with the known macOS parallel-test flakiness, but I am flagging it rather
  than calling it clean.
- `cargo test -p perry-codegen --lib`: 674 passed / 0 failed.
- `raw_handle_debt.py` 998 (baseline 998); `addr_class_inventory.py` passed;
  `class_id_collisions.py` passed; `check_file_size.sh` OK; `cargo fmt --all -- --check` clean.

### Sabotage evidence (every new test shown able to fail)

| mutation | result |
|---|---|
| S1 codegen comparand → `0x40` | `codegen_tenured_comparand_matches_the_runtime_flag` + witness RED |
| S2 drop the incremental clause | `the_incremental_clause_forces_the_call_for_a_young_parent` RED |
| S3 `buffer_alloc` forgets `GC_FLAG_TENURED` | `every_old_gen_birth_path_sets_tenured` RED |
| S4 gate's `or` → constant-true | `array_push_barrier_is_gated_on_the_parent_header` RED |
| S5 sink a slot store into the gated block | `array_push_slot_store_stays_outside_the_gate` RED |
| restored | all green |

**S4 initially did NOT fail**, and that is worth reading. Replacing the `or` with a
constant-true leaves both `and i8 …, 32` and the incremental global in the IR — the
clauses are still computed, just no longer consulted — so a substring-matching test
stayed green while the gate had stopped gating: CLAUDE.md hazard 4 in my own test.
The test now follows the `cond_br`'s condition back to its definition and requires an
`or i1` of an i8 header test and an i32 count test.

### One thing I tried and reverted, deliberately

A `debug_assert!` in `barrier_parent_needs_remembering` checking `Old ⟹ TENURED` at
the point the classification is made. That is the better enforcement — every old-parent
store in every debug run would recheck it — and `arena_alloc_gc_old` genuinely does
*not* establish the invariant (it writes `GC_FLAG_ARENA | gc_birth_extra_flags()` and
leaves the bit to its eight callers). But dozens of existing tests build old-gen
fixtures straight from that allocator without the bit — `alloc_old_test_object`,
`alloc_old_test_array`, `alloc_old_test_promise`, most of `gc/tests/oldgen.rs` — some
deliberately, so it fired on fixtures rather than defects and turned `cargo-test` red.
The invariant is instead pinned over the **production** birth paths by
`every_old_gen_birth_path_sets_tenured` (sabotage S3), and the reason the assert is
absent is recorded where someone would next try to add it. Making it shippable means
fixing those fixtures first, which is its own change.

## Summary by CodeRabbit

* **Performance**
  * Improved array insertion efficiency by avoiding unnecessary garbage-collection write-barrier work when it is not required.

* **Reliability**
  * Preserved correct memory management for older-generation arrays and during incremental garbage collection.
  * Added validation covering array updates, garbage collection, and reference tracking.

* **Release**
  * Updated the application version to **0.5.1341**.
  * Added release documentation describing the array performance and safety improvements.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1340
+**Current Version:** 0.5.1341
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1340"
+version = "0.5.1341"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "perry-ffi",
  "scraper",
@@ -5813,7 +5813,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-commander"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5821,7 +5821,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cron"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "chrono",
  "cron",
@@ -5831,7 +5831,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dayjs"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "chrono",
  "perry-ffi",
@@ -5839,7 +5839,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-decimal"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "perry-ffi",
  "rust_decimal",
@@ -5847,7 +5847,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dotenv"
-version = "0.5.1340"
+version = "0.5.1341"
 dependencies = [
  "perry-ffi",
  "serde_json",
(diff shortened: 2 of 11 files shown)


proggeramlug:
## Audit before merge — verified, merged as v0.5.1341

**Perf reproduced**: interleaved best-of-5 both arms, my own builds of current
main vs the PR — `push_cls` 0.69–0.72 s → 0.50–0.51 s = **1.38×**, output
identical. Consistent with the claimed 1.33× quiet-host figure.

**The soundness evidence that matters most reproduced bit-for-bit**: `retain` —
the workload whose barriers are *real* — shows `old_to_young_slow_hits =
3,204,922` and `new_inserts = 6,268` on BOTH arms, to the digit. Every genuine
old→young edge is still recorded; only the 99% of calls that never inserted
anything are skipped. That, plus the design point that a `!TENURED` parent is
fully traced by every minor (so its edges are rediscovered, not lost), is the
argument — and the incremental clause preserving SATB shading closes the one
non-generational hole.

**Sabotage re-verified, both directions:**
- Drifting codegen's `GC_FLAG_TENURED_I8` "32"→"16" →
  `array_push_barrier_is_gated_on_the_parent_header` red. Note the contract is
  a **three-copy pin** (codegen string, runtime mirror `CODEGEN_GC_FLAG_TENURED`,
  real flag): my drift of the codegen copy was caught by the IR test while the
  runtime equality test stayed green — the two tests jointly cover both drift
  directions, but anyone reading `codegen_tenured_comparand_matches_the_runtime_flag`
  alone should know it pins the *mirror*, not the emitted IR.
- Gate forced constant-true → the same IR test red. The S4 hardening (follow the
  `cond_br` to its definition rather than grepping for substrings) is real — my
  constant-true sabotage is exactly the mutation the pre-hardened version
  missed, per the report's own disclosure.

**Root-dominance re-run because codegen changed**: corpus 128/128,
`--moving-only` 0 violations with 40/40 seeded caught, `--unrooted-allocas` 0.
Full suites 1,851/0 runtime, 677/0 codegen; all four lint gates + fmt clean.

## One honest limitation of my audit

I could **not** reproduce the ZEAL+PROTECT_FROMSPACE arm: with
`PERRY_GC_MOVING_LOOP_POLLS=1` compiled in (3 polls verified in the IR) and
zeal on, my bench runs performed **zero copying minors** — instrument armed,
collector never moving, i.e. the exact vacuous-green shape CLAUDE.md warns
about. The report claims 110 retired quarantine sets, presumably on a workload
shape with event-loop boundaries rather than these compute-only benches. I am
NOT treating my zero as evidence against the PR — the bit-identical
remembered-set counters above are the stronger and direct soundness check — but
the zeal-on-compute-only-bench recipe not arming *at all* on a current binary
is worth a look; if `zeal_forced_collections()` is genuinely zero there, every
green "ran under zeal" claim on such benches is vacuous, which is #7154-family
instrument territory.

## Kept from the report, for the record

The `debug_assert!(Old ⟹ TENURED)` revert story is the right call, correctly
documented: the invariant holds on all 8 production birth paths (pinned by
`every_old_gen_birth_path_sets_tenured`) but test fixtures build old-gen
objects without the bit, so a continuous assert fires on fixtures. And the
honest scope statement stands: `tree`/`cycles`/`deeplist` gain ~nothing because
their barriers are class-field stores — that is the remaining #7511 lever, and
the issue stays open for it.

PR #7573, opened 2026-08-07
fix(runtime): a Map/Set subclass in a base-typed binding was read as a raw header (#7570)

proggeramlug:
Fixes #7570.

## The bug

```ts
class MyMap<K, V> extends Map<K, V> {}

const m: Map<string, number> = new MyMap<string, number>();
m.set("a", 1);          // <-- SIGBUS, exit 138, before anything prints
console.log("size:", m.size);
```

Node prints `size: 1`. Reproduced on `main` at v0.5.1323, exit 138 with zero output.

Annotating a binding with a base type is everyday TypeScript. A **parameter**
(`function f(m: Map<string, number>)`) is the more likely way to hit this in real
code — NestJS's `ModulesContainer extends Map` is exactly this shape. All five
binding forms are affected: `const`, parameter, class field, return type, and
`as Map<…>` cast.

## Verified root cause

Perry models a Map/Set subclass instance as a plain `ObjectHeader` carrying the
real collection under a hidden field (`object/map_set_subclass.rs`). The two
headers overlay field-for-field:

| `MapHeader` field | offset | actually reads on an `ObjectHeader` |
|---|--:|---|
| `size: u32` | 0 | `object_type` (= 1) |
| `capacity: u32` | 4 | `class_id` |
| `entries: *mut f64` | 8 | `parent_class_id` ‖ `field_count` |

`entries` is two `u32` class ids glued into a pointer, and the first store
through it faults (`map_set_string_key_value + 708`, `str x21, [x20], #0x8`).

The instance reaches those raw entry points because **"is a Map" was decided
from the declared TypeScript type** — at three layers, all keyed on the same
fact:

- `crates/perry-codegen/src/type_analysis/strings.rs:135` (`is_map_expr`) and
  `:13` (`is_set_expr`) — satisfied by `Type::Generic { base: "Map" }`, no
  subclass or brand check. Feed `lower_call/property_get/map_set.rs:22,131,289,317`
  and `expr/property_get.rs:486,495` (`.size`).
- `crates/perry-hir/src/lower/expr_call/local_array_methods.rs:915-1032` — the
  HIR fold to `Expr::MapSet` / `MapGet` / `MapHas` / `SetAdd` / … on
  `ctx.lookup_local_type(...) == Generic { base: "Map" | "Set" }`.
- `crates/perry-hir/src/lower/stmt_loops.rs:1306,1326` and
  `lower/for_head.rs:303` — the `for…of` index fast paths
[...]

Issue #7570, opened 2026-08-07
SIGBUS: a Map subclass instance in a variable annotated `Map<K, V>` is dereferenced as a raw MapHeader


proggeramlug:
A `class X extends Map` instance stored in a variable **annotated** `Map<K, V>`
is handed to the raw `js_map_*` entry points as if it were a real `MapHeader`.
It is a plain `ObjectHeader`, so the very first `.set()` writes through a
pointer fabricated out of two `u32` class-id fields.

Found while fixing #7563. It is a **different** bug: #7563 is an
array-capacity/class-id confusion in `js_class_method_bind` and is fixed by
PR #7569; this one survives that fix.

## Repro

```ts
class MyMap<K, V> extends Map<K, V> {}

const m: Map<string, number> = new MyMap<string, number>();
m.set("a", 1);          // <-- SIGBUS, before anything prints
console.log("size:", m.size);
```

```
$ node --experimental-strip-types ann.ts
annotated size: 2
annotated get: 1
annotated values: 1,2
annotated keys: a,b
annotated spread: [["a",1],["b",2]]
param-annotated: 11
rc=0

$ ./ann
rc=138            # SIGBUS, zero output
```

A parameter annotation is the same shape and is the more likely way to hit this
in real code:

```ts
function take(mm: Map<string, number>): number {
  let t = 0;
  for (const [, v] of mm) t += v;   // same fault
  return t;
}
take(new MyMap<string, number>([["x", 5]]));
```

Without the annotation (`const m = new MyMap<string, number>()`) the receiver
types as the subclass and everything works — that is the shape
`test_gap_6325_map_set_subclass.ts` covers.

## Backtrace (`PERRY_DEBUG_SYMBOLS=1`)

```
stop reason = EXC_BAD_ACCESS (code=2, address=0x1ffff0032)
  frame #0: perry_runtime::map::map_set_string_key_value + 708
  frame #1: main + 332

->  0x10016b970 <+708>: str    x21, [x20], #0x8
```

A store through `x20`, i.e. the entries cursor.

## Root cause

`ObjectHeader` and `MapHeader` overlay each other field-for-field:

| `MapHeader` field | offset | actually reads (on an `ObjectHeader`) |
|---|--:|---|
| `size: u32` | 0 | `object_type` (= 1) |
| `capacity: u32` | 4 | `class_id` |
| `entries: *mut f64` | 8 | `parent_class_id` \|\| `field_count` — two `u32` class ids glued into a pointer |

[...]

Issue #6386, opened 2026-07-13
[perf] DataView accessors, Array.concat, and regex match-with-groups are 4–30x slower than Hermes


dlee:
## What happened

Three hot operations each lower to an opaque `js_*` runtime helper and land well
behind the Hermes interpreter. Perry 0.5.1220 vs Node v22 (V8) vs Hermes 1.0.0
(built from `main`), best of 5, macOS arm64; identical checksums across engines.

| Op | Perry | Node | Hermes | Perry/Hermes |
|---|--:|--:|--:|--:|
| `DataView` set/getFloat64 (300×4096) | 205 ms | 1 ms | 34 ms | **6x** |
| `Array.prototype.concat` (20000×1024) | 686 ms | 11 ms | 23 ms | **30x** |
| regex `.match()` w/ 4 groups (300k) | 237 ms | 12 ms | 67 ms | **~3.5x** |

Regex *matching without extraction* (`RegExp.test()`) is separately fast — the
cost is in materializing capture arrays / accessor results, i.e. the helper
call, not the regex engine.

## What you expected

At least Hermes-interpreter throughput on these ops.

## Minimal reproduction

```javascript
// DataView — 6x
function dv() { var N=4096, b=new ArrayBuffer(N*8), v=new DataView(b), a=0;
  for (var r=0;r<300;r++){ for(var i=0;i<N;i++) v.setFloat64(i*8,i*1.5+r,true);
    for(var i=0;i<N;i++) a+=v.getFloat64(i*8,true);} return Math.round(a); }

// Array.concat — 30x (arrays built once, concat isolated)
function cc() { var N=512,x=[],y=[]; for(var i=0;i<N;i++){x.push(i);y.push(i+N);}
  var a=0; for(var r=0;r<20000;r++){ a+=[].concat(x,y).length; } return a; }

// regex match with capture groups — ~3.5x
function rm() { var s="2026-07-13 key=42 val=99", re=/(\d{4})-(\d{2})-(\d{2}) key=(\d+)/, a=0;
  for (var i=0;i<300000;i++){ var m=s.match(re); if(m) a+=m[1].length+m[4].length; } return a; }
```

```bash
perry compile x.js -o x && ./x     # vs node x.js / hermes -O x.js
```

## Environment

- **Perry version**: 0.5.1220
- **Host OS**: macOS 27.0 (arm64)
- **Target**: native
- **Installed via**: npm

## Diagnostic output

```
[...]

PR #7603, opened 2026-08-07
fix(runtime): an Array subclass in a base-typed binding was read as a raw header (#7574)

### LINKED ISSUE #7574: class X extends Array in an `T[]`-annotated binding takes the raw ArrayHeader fast paths (sibling of #7570)
Sibling of #7570 (fixed by #7573) on a different family. Filed from the sweep
that PR did of every declared-type-keyed lowering; not fixed there because it is
a separate set of code paths with a different guard already half-present.

## Shape

`class X extends Array` produces a **plain `ObjectHeader`**, not an
`ArrayHeader` — stated outright in `crates/perry-runtime/src/array/subclass.rs:1-7`
("perry has no exotic array-object representation"), and `js_array_subclass_init`
(`node_stream_constructors/builders.rs:98`) installs the elements as ordinary
object *properties*, with no `ArrayHeader` backing (unlike
`js_map_set_subclass_init`, which does allocate a real hidden Map/Set).

`ObjectHeader` and `ArrayHeader` overlay:

| `ArrayHeader` | offset | reads on an `ObjectHeader` |
|---|--:|---|
| `length: u32` | 0 | `object_type` (= 1) |
| `capacity: u32` | 4 | `class_id` |
| *(elements begin)* | 8 | `parent_class_id` ‖ `field_count`, then `keys_array`, then `meta` |

`is_array_expr` (`crates/perry-codegen/src/type_analysis/predicates.rs:401`) is
satisfied by a **declared** `Type::Array(_)` / `Generic { base: "Array" }`, so a
binding annotated with the base type and holding a subclass instance takes the
raw lowering — the same "a declared type is a hint, never a layout fact" premise
as #7570.

```ts
class MyArr<T> extends Array<T> {}
const a: number[] = new MyArr<number>();
a.push(1);
console.log(a.length, a[0]);
```

## Tiers that do NOT brand-check

- `crates/perry-codegen/src/expr/index_get.rs:531` — `lower_bounded_array_index_get`
  (the hoisted-`arr.length` loop fast path) tests only `GC_TYPE_LAZY_ARRAY`,
  `GC_FLAG_FORWARDED` and `OBJ_FLAG_ARRAY_DESCRIPTORS`, then does a raw
  `gep` + `load double` at `handle + 8 + idx*8` — straight into
  `parent_class_id ‖ field_count`, then `keys_array`, then `meta`.
- `crates/perry-codegen/src/expr/property_get.rs:236` — inline `arr.length` is a
  bare `safe_load_i32_from_ptr(recv)` at offset 0, no GC-type test, so it reports
  `object_type` (= 1).
- `crates/perry-codegen/src/lower_call/property_get.rs:215` → `lower_array_method`
  (`js_array_pop_f64`, `js_array_join`, `js_array_sort`, `js_array_reverse`, …).
- `crates/perry-codegen/src/expr/property_set.rs:383` → `js_array_set_length_strict`.
- `crates/perry-codegen/src/expr/index_set.rs` — the guard-free
  `Ptr<NumArray>` store arm (`try_lower_num_array_guard_free_set`).

## The tier that already does it right

`crates/perry-codegen/src/expr/index_get/guarded_array.rs:22-100` loads the GC
header byte at `handle - 8` and tests `icmp eq i8 %gc_type, 1` (`GC_TYPE_ARRAY`)
before any slot load, branching to a generic fallback otherwise. An Array
subclass instance takes the fallback there and behaves correctly.

The recognition machinery also already exists on the runtime side and is used by
`js_native_call_method` — `array::subclass::is_array_subclass_instance` (`:46`,
which requires `GC_TYPE_OBJECT` via `try_read_gc_header`) and
`array_subclass_dense_snapshot` (`:77`, used by iteration/spread) — so
`class X extends Array` works fine through the *generic* dispatch. It is only the
declared-type fast paths that are unsafe.

## Suggested fix

Whichever of the two #7573 considered:

- extend the `guarded_array.rs` `GC_TYPE_ARRAY` test to the other tiers (keeps
  the fast path, costs one header-byte load + branch), or
- give the raw `js_array_*` entries a receiver resolution the way #7573 gave
  `clean_map_ptr` / `clean_set_ptr` one, so it is fail-closed for every future
  caller. Note this family has no hidden backing to redirect *to* — an Array
  subclass keeps its elements as object properties — so the runtime-side answer
  is "fall back to the generic array-like path", not "redirect".

The second is the one that cannot be forgotten at a new call site; the first is
the smaller diff.

proggeramlug:
Fixes #7574.

## The bug

```ts
class MyArr<T> extends Array<T> {}

const a: number[] = new MyArr<number>();
a.push(1);
a.push(2);          // <-- SIGSEGV, exit 139
console.log(a.length, a[0]);
```

Reproduced on `main`: exit **139**, one line of (already wrong) output. The
fault address is `0x3ff0000000000000` — the IEEE bit pattern of the double
`1.0`, i.e. the value the *first* push stored, later dereferenced as a pointer.

Sibling of #7570 (fixed by #7573) on a different family, and the same premise:
**a declared TypeScript type is a hint, never a layout fact** (CLAUDE.md,
*Known Limitations*). All five binding forms are affected — `const`, parameter,
class field, return type, and `as number[]` cast.

## Verified root cause

An Array-subclass instance is a plain `ObjectHeader`
(`array/subclass.rs:1-7`; `js_array_subclass_init` installs a `length` own
property and keeps the elements as ordinary indexed object properties — there
is no `ArrayHeader` backing). `ObjectHeader` and `ArrayHeader` overlay field
for field:

| `ArrayHeader` | offset | actually reads on an `ObjectHeader` |
|---|--:|---|
| `length: u32` | 0 | `object_type` (= 1) |
| `capacity: u32` | 4 | `class_id` |
| `elements[0]` | 8 | `parent_class_id` ‖ `field_count` |
| `elements[1]` | 16 | **`keys_array`** (`*mut ArrayHeader`) |
| `elements[2]` | 24 | **`meta`** (`*mut ObjectMeta`) |

`1 <= class_id` sails through `clean_arr_ptr`'s `length <= capacity <= 100M`
sanity check, so the forged header is accepted. `a.push(1)` stores `1.0` at
`handle + 8 + length*8` = `handle + 16` — **over `keys_array`, a live GC child
edge** — and writes `length + 1` over `object_type`. The second push then walks
that forged pointer and faults. Element *reads* are the same hazard in the other
direction: they hand `keys_array` / `meta` out to user code as doubles.

### The tiers, checked on current `main`

The issue named three; the sweep found **two more**, and one of its three turned
out to be a false positive:

| tier | status on `main` |
|---|---|
| `expr/array_push.rs:301` — the inline `Expr::ArrayPush` store (fwd-flag test only, then raw slot store + `length` bump) | **UNSAFE — this is the SIGSEGV.** Not named in the issue. |
| `expr/index_get.rs:531` `lower_bounded_array_index_get` (tests `GC_TYPE_LAZY_ARRAY`, `GC_FLAG_FORWARDED`, `OBJ_FLAG_ARRAY_DESCRIPTORS`, then raw `gep`+`load double`) | **UNSAFE**, as filed |
| `array/header.rs:511` `clean_arr_ptr` — the funnel behind ~190 `js_array_*` entries | **UNSAFE**: no `obj_type` test, only length/capacity sanity |
| `typed_feedback.rs:2413` `js_typed_feedback_array_index_set_fallback_boxed` | safe, but no Array-exotic `length` step. Not named in the issue. |
| `expr/property_get.rs:236` — inline `arr.length` | **already brand-checks** (`obj_type == 1 \|\| == 3`, then `js_value_length_f64`, whose `GC_TYPE_OBJECT` arm reads the own `length`). The issue's claim that it is "a bare `safe_load_i32_from_ptr`" does not hold on current `main`. |
| `expr/index.rs:79` `lower_index_set_fast` inline tier, and both typed-feedback guards | already test `obj_type == GC_TYPE_ARRAY` |
| `expr/ptr_numarray_access.rs` guard-free `Ptr<NumArray>` get/set | **unreachable**: the proof requires the local be initialized by exactly one `new Array(<static n>)` or `[]`, which `new MyArr()` is not |
| `expr/index_get/guarded_array.rs:96` | already correct (the model) |

## Fix

**Runtime funnel first, for memory safety.** `clean_arr_ptr` now refuses a
`GC_TYPE_OBJECT` / `GC_TYPE_CLOSURE` allocation, so **all ~190 call sites become
fail-closed at once** — each degrades through its existing null branch instead
of dereferencing a forged header. It reuses the `obj_type` byte the surrounding
block already loads for the forwarding/lazy checks, so a genuine array pays one
extra compare; the registry probes that rule out header-less buffers/typed
arrays are in the cold arm only. Same "resolve at the shared runtime funnel, not
at one codegen predicate at a time" shape as #7573.

**Then correctness, at the entry points the declared-type tiers actually reach.**
Unlike Map/Set there is nothing to *redirect to* — an Array subclass has no
hidden backing. But perry already has a complete spec-generic array-like engine
(`array/generic.rs`, `generic_object.rs`) that operates on exactly this
representation and is the path the **unannotated** form has always used through
`js_native_call_method`. So the entry points re-enter through their existing
null branch and run the operation there.

Per-entry-point inventory (everything reachable from the three tiers):

| entry point | disposition |
|---|---|
| `js_array_push_f64` | **resolve** → `run_object_mutator(recv, "push")`, returns the ORIGINAL receiver so codegen's realloc write-back keeps the binding (returning a fresh empty array is what made the push look silently dropped) |
| `js_array_set_f64_extend` (`…_strict`) | **resolve** → object `[[Set]]` + Array-exotic `length` |
| `js_array_get_f64`, `js_array_get_f64_unchecked` | **resolve** → `al_get` |
| `js_array_set_length_strict` | **resolve** → Array-exotic `Set(O,"length",n,true)` (deletes truncated indices) |
| `js_array_pop_f64`, `js_array_shift_f64` | **already funnel** (`plain_object_value` → generic engine) |
| `js_array_length` | **already funnels** (its `GC_TYPE_OBJECT` arm reads the own `length`) |
| `js_array_map/filter/forEach/some/every/find*/reduce*/join/slice/indexOf/lastIndexOf/includes/at` (21 sites) | **already funnel** via `normalize_array_receiver`, which materializes an array-like object into a dense snapshot |
| `js_array_forEach` | funnels, **plus** the receiver override — see below |
| `js_array_concat_variadic` | **already funnels** (`obj_type` re-dispatch + `append_concat_arg`'s subclass snapshot arm); its #6386 dense bulk path needed a guard — see below |
| `js_array_clone_for_spread`, `js_get_iterator`, `for…of` | **already funnel** (`array_from_spread_value` / `symbol/iterator.rs` subclass arms) |
| `js_array_numeric_*`, `js_array_is_numeric_f64_layout`, `js_template_raw` | **unreachable / benign**: null now means "not a numeric-layout array", which is the correct answer |
| the guard-free `Ptr<NumArray>` tiers | **unreachable** (provenance proof excludes `new MyArr()`) |

Three sites needed more than the funnel, all analogues of #7573's extra four:

- **`js_array_forEach`'s 3rd argument.** `normalize_array_receiver` hands the
  loop a dense *snapshot*, so the callback saw the snapshot and `self === sub`
  was `false`. It now passes the original receiver through, gated on a one-load
  `GC_TYPE_OBJECT` header test (`addr_class::try_read_gc_header`) so a genuine
  array never enters the registry probes, and rooted across the callbacks.
- **`length` is written, not just read.** `sub[3] = v` on a real Array runs the
  exotic `[[DefineOwnProperty]]` and leaves `length == 4`; a plain object's does
  not. Pre-fix `sub[0] = 10; sub.length` read back `0` — **on the unannotated
  path too** — which then made the next `sub.push(v)` append at index 0 and
  overwrite the element. The step is applied after the store at the three
  generic funnels a subclass index-write can reach (`js_put_value_set`,
  `js_object_set_index_polymorphic`,
  `js_typed_feedback_array_index_set_fallback_boxed`), gated on the class chain
  reaching `Array` so an object literal short-circuits on `class_id == 0`.
  `sub.length = n` likewise routes to the exotic setter, which deletes the
  truncated indices.
- **`clean_arr_ptr` returning null now MEANS something new** at sites that read
  null as "an empty array". `concat`'s #6386 all-dense bulk path did exactly
  that: `peek_plain_array_len` / `dense_concat_array_source` answered
  `Some(0)`, the bulk path claimed the copy, and the spec-shaped
  `append_concat_arg` flow (which has the subclass snapshot arm) never ran —
  `[1,2].concat(sub)` yielded `1,2`. **This was a regression my own funnel
  introduced, caught by the family A/B below, not by the new test.** Both now
  classify the receiver before the null shortcut and answer `None`
  ("un-peekable"), restoring the pre-fix routing. I swept the other
  null-means-empty sites (`js_template_raw`,
  `js_array_{mark,is}_numeric_f64_layout`, `js_array_numeric_set_f64_unboxed`);
  for those, null is the correct answer.

**Two codegen guards are unavoidable.** Unlike #7573, two of the three tiers
emit *no runtime call at all* — they are inline LLVM IR — so no runtime funnel
can reach them. Both now test `obj_type == GC_TYPE_ARRAY` and route a miss to
the slow call they already had (which then resolves through the funnel above).
Both are **strictly more restrictive** than what they replaced, so no receiver
that used to take the slow path now takes the fast one:

- `lower_bounded_array_index_get`: `icmp eq gc_type, 9` (`GC_TYPE_LAZY_ARRAY`)
  → `icmp ne gc_type, 1`. Lazy arrays are 9, so the new test subsumes the old
  one — and it is one instruction *cheaper* (an `icmp` replaces `icmp` + `or`).
- inline `Expr::ArrayPush`: the existing `gc_flags & FORWARDED` predicate gains
  `|| obj_type != GC_TYPE_ARRAY`; a miss takes the `apush.fwd` arm, which is
  already a `js_array_push_f64` call.

**Why not give an Array subclass a hidden backing** the way #7573 did for
Map/Set: its elements *are* its own indexed properties, and `Object.keys` /
`for…in` / `JSON.stringify` / the whole generic engine read them there. A second
storage would have to be kept in sync with every one of those paths — a far
larger and riskier change than making the raw entries resolve.

## Validation (local; CI has a deep backlog, so local is what this rests on)

- **`test-files/test_gap_7574_array_subclass_declared_base_type.ts` —
  byte-identical to `node --experimental-strip-types` (v26.5.1), exit 0**, 39
  lines. Covers all five binding forms × index get/set, `.length` read AND
  write (including truncation), push/pop/shift, the bounded-index loop tier,
  `for…of`/spread/`Array.from`/destructuring, `map`/`filter`/`slice`/`join`/
  `indexOf`/`includes`/`reduce`, `forEach` receiver identity, an indirect
  subclass, a subclass with its own constructor and fields, and non-subclass
  controls (a real array in the same forms, and a plain object merely
  *annotated* `number[]`).
- **Sabotage, both directions, on a cleared object cache.** With
  `crates/perry-runtime/` **and** `crates/perry-codegen/` reverted *in full* to
  `main` and the test file untouched, the same file exits **139** after one
  line of output (`push1 0`; node prints `push1 1`). With both restored it exits
  **0**, byte-identical. Recorded separately: the runtime half *alone* does not
  stop the crash — the inline push tier never calls into the runtime — which is
  why the two codegen guards are in this PR.
- **Fast path proven still live.** On a plain-array-only program the emitted
  LLVM IR is **byte-identical apart from the guard predicate**: after
  normalizing SSA numbering, `probe/plain.ts` differs by **15 lines** (3 push
  sites × the 4 new instructions + the `or`), and `probe/plain2.ts` — which
  reaches the bounded-index tier — by exactly **3 lines**, one `icmp eq …, 9` →
  `icmp ne …, 1` per site, with identical instruction counts (3970/3970). The
  `apush.inbounds` and `bidx.fast` blocks and their raw slot loads are
  unchanged, and both programs produce identical output on both arms. The unit
  test `a_genuine_array_takes_the_fast_path_and_is_never_redirected` asserts
  `clean_arr_ptr` returns a real `ArrayHeader` unchanged **and** that the
  redirect answers `None` for it, so the identity cannot have come from a
  redirect that happened to agree.
- **Six sabotage-shaped unit tests** in `array/subclass_tests.rs`: each first
  asserts the bytes the pre-fix code misread are still sitting there
  (`ArrayHeader.length == 1` aliasing `object_type`, `capacity == class_id`,
  both passing the old sanity check) and only then that the entry point refuses
  or resolves — so a green run proves the brand check fired.
- **Array-family A/B**, 141 `test-files/test_*{array,spread,iter,foreach,for_of,
  slice,splice,sort,concat,flat,push,pop,destructur}*.ts`, run on both arms:
  the failure set is **identical** (2 pre-existing `COMPILE_FAIL`, 8
  `NODE_FAIL` where node itself cannot run the file, 1 pre-existing
  `RUN_FAIL`), except that `test_gap_7574_…` fails **139** on the reverted arm
  and passes on this one. `test_gap_6232_class_extends_array` also passes on
  both — its `concat` line is what caught the bulk-path regression above, which
  is fixed here.
- `cargo test -p perry-runtime`: **1850 passed, 0 failed**.
  `cargo test -p perry-codegen --lib`: **671 passed, 0 failed**.
- `python3 scripts/raw_handle_debt.py`: **998 (baseline 998)**, per-module
  ceilings held. `python3 scripts/addr_class_inventory.py`,
  `python3 scripts/class_id_collisions.py`, `./scripts/check_file_size.sh`,
  `cargo fmt --all -- --check`: all clean.

## Known gap left in place

`ArraySpeciesCreate` on a subclass: node's `sub.map(f)` returns a `MyArr`,
perry returns a plain `Array`. That is **pre-existing and identical on the
unannotated path**, so it is out of scope here; the gap test compares element
CONTENT (via `join`) rather than the container's console formatting, and says
so inline. Worth its own issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed crashes and incorrect behavior when `Array` subclasses are used through array-typed bindings.
  * Improved support for indexed access, `push`, `concat`, iteration, spread, stack operations, and `length` updates.
  * Prevented plain objects and other non-array values from being misinterpreted as arrays.
  * Preserved subclass receivers during callbacks and array-like operations.

* **Tests**
  * Added regression coverage for array subclasses, invalid array-like values, indexing, iteration, and method behavior.

* **Documentation**
  * Updated the documented version to 0.5.1343.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1342
+**Current Version:** 0.5.1343
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1342"
+version = "0.5.1343"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "perry-ffi",
  "scraper",
@@ -5813,7 +5813,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-commander"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5821,7 +5821,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cron"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "chrono",
  "cron",
@@ -5831,7 +5831,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dayjs"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "chrono",
  "perry-ffi",
@@ -5839,7 +5839,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-decimal"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "perry-ffi",
  "rust_decimal",
@@ -5847,7 +5847,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dotenv"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "perry-ffi",
  "serde_json",
@@ -5855,7 +5855,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ethers"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "perry-ffi",
  "rand 0.10.1",
@@ -5863,7 +5863,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-events"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5871,14 +5871,14 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-exponential-backoff"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "perry-ffi",
 ]
 
 [[package]]
 name = "perry-ext-fastify"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "bytes",
  "http-body-util",
@@ -5896,7 +5896,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-fetch"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "bytes",
  "lazy_static",
@@ -5909,7 +5909,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-http"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "bytes",
  "h2",
@@ -5933,7 +5933,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ioredis"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "lazy_static",
  "perry-ffi",
@@ -5943,7 +5943,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-jsonwebtoken"
-version = "0.5.1342"
+version = "0.5.1343"
 dependencies = [
  "base64",
  "jsonwebtoken",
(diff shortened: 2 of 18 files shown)


proggeramlug:
## Audit before merge — verified, merged as v0.5.1343

**Gap test byte-identical to node** (39 lines, exit 0) on my own build, rebased
past #7600/#7602 (both of which touched the same `array_push.rs`).

**Sabotage decisive in both directions:** neutering just the new inline-push
brand check (`icmp_ne gc_type, gc_type` — always false) brings back **exit 139
with one line of output**, the exact SIGSEGV; restored, exit 0 byte-identical.
This confirms the report's most important structural finding — the runtime
funnel alone cannot stop this crash, because the inline `Expr::ArrayPush` tier
never calls into the runtime. The two codegen guards are load-bearing, not
belt-and-braces.

**Fast-path preservation:** `a_genuine_array_takes_the_fast_path_and_is_never_redirected`
green, full suites 1,857/0 runtime + 677/0 codegen, all four lint gates + fmt
clean.

Three things from the report worth keeping visible:

1. **Two of the issue's three citations were wrong and the two real tiers were
   unnamed.** `property_get.rs:236` already brand-checks; the `Ptr<NumArray>`
   tiers are unreachable by provenance proof; the actual crash was the inline
   push store plus the typed-feedback boxed fallback. Verifying the filing
   rather than inheriting it is what made this fix land on the right sites.
2. **The funnel itself introduced a regression — `[1,2].concat(sub)` silently
   dropping elements — caught by the 141-test family A/B, not by the new gap
   test.** That is the strongest argument this repo has for family A/Bs on
   memory-safety fixes: the new test guards the bug you knew about, the A/B
   guards the bug you just created.
3. The deliberate leftovers (`ArraySpeciesCreate` on subclasses, static `from`
   as a GET, `sub instanceof MyArr`) are named, pre-existing, and need their
   own issues — same triage discipline as #7573's.

Issue #7539, opened 2026-08-06
gc: the JSON tape's large per-parse block interacts badly with the generational collector (3.2x RSS, 25x variance vs mark-sweep)


proggeramlug:
Split out of #7478's decomposition (see #7537, merged). This is now the dominant term in `field_access`, and it is **not** the tape's intrinsic cost.

## The measurement

The public baseline's `idiomatic` row flips **two** switches at once (`PERRY_JSON_TAPE=0` *and* `PERRY_GEN_GC=0`), so "tape on, mark-sweep" had never been measured. Decomposing all four combinations on the pinned quiet host (M1 mini, 11 runs, `taskpolicy -t 0 -l 0`) at `origin/main` @ fc5e56ee0:

| variant | tape + genGC | tape + markSweep |
|---|--:|--:|
| parse+scan | 2729 ms · **σ 214.9** · **208 MB** | 2957 ms · σ 8.8 · **66 MB** |
| field_access | 2942 ms · **σ 101.8** · **222 MB** | 3247 ms · σ 7.7 · **67 MB** |

**Identical tape, different collector.** σ collapses from 214.9 to 8.8 and RSS from 208 MB to 66 MB — a 3.2× memory difference and a 25× variance difference, with the tape doing exactly the same work.

Neither factor alone explains it: tape-off under gen-GC is σ 17.6. It is specifically the **interaction**.

## Mechanism

Each parse allocates header + tape as one ~2.4 MB block (200,002 `TapeEntry`s for this fixture). That rounds up to a dedicated oversized arena block and is retained; the resulting old-gen growth fires full collections at data-dependent points, which is both the variance and the retained footprint.

## Why it matters more after #7537

#7537 took `field_access` from 2981 ms to 2016 ms by flipping scans to the batch parser early, and `parse+scan` from 2729 to 1339. With the element-wise materializer no longer the bottleneck, this interaction is what is left: `full - scan` is **707 ms under gen-GC but only 309 ms under mark-sweep** for the identical tree.

The strongest single data point: **`tape + markSweep` field_access after #7537 is 1759 ms at σ 2.1 and 76 MB** — within 282 ms of the idiomatic floor and completely stable. The same build under gen-GC is 2016 ms at σ 219 and 193 MB. The tape is not the problem any more; the collector's handling of its large per-parse block is.

## Suggested direction

[...]

Issue #7480, opened 2026-08-06
repsel: element-shape proofs through arrays — measured 6.2× vs node; route = invariant bit → versioned-loop consumer → element Ptr<Shape>


proggeramlug:
Scoping for the repsel stack's next gap after 3b/4a: **shape proofs do not survive an array element read.** `keep[j]` yields an untyped JSValue, so `.v` re-enters the field-read diamond that 3b deleted for proven locals. This is the churn-probe access shape and the blocker named in the phase sequencing ("element-shape proof through arrays").

## Measured (2026-08-06, release build @ main, checksums equal everywhere)

Kernel: `keep: {v,w}[]` of 200k, `sum += keep[j].v` × 50 sweeps:

| runtime | ms |
|--:|--:|
| node | 15 |
| bun | 23 |
| **perry** | **93** |

**6.2× vs node / 4× vs bun** on the pure shape. IR inspection: the loop body carries no out-of-line guard *calls* (4a's inline element tier + the #5093-family inline field precheck are doing their job) — the cost is the **stacked inline diamonds**: element-read tier (tag/handle-band/header/layout tests) feeding a field-read precheck (gate load + tag + handle-band + 7 dependent header loads), per iteration. Exactly the 3b recon's "the guard is the cost", now twice per access.

## The route decision: both candidate routes share one prerequisite

**Route A — element representation (phase-4c style):** prove `keep : Array<Ptr<__AnonShape_…>>` from construction flow, so element reads yield proven `Ptr<Shape>` and field reads emit bare `gep+load`. Full fix; helps non-loop access too.

**Route B — extend the #5093 versioned-loop clone** (machinery just revived by #7425; 3b's recon calls the clone "the only unguarded form today"): hoist ONE whole-array element-shape guard into the preheader, clone the body with bare reads.

The scoping finding is that **B is not actually cheaper on the load-bearing part**: a preheader "all elements are shape X" test needs a per-array *homogeneous-element-shape invariant* that does not exist today (the only array-level invariants are the numeric-layout bits, `GC_ARRAY_RAW_F64_LAYOUT`/`_HOLES`). Maintaining that invariant — set at construction, updated/cleared at every store site, self-healing like 4a's dense bit — is the same core mechanism Route A needs. Once it exists:

1. **B consumes it first** (small codegen change: preheader guard + cloned body via the existing versioned-loop machinery) — fast win on loop-shaped access, which is where the churn probes live.
[...]

Issue #5093, opened 2026-06-13
perf(method dispatch): method_calls ~290× Node — remaining cost is per-field-access shape-guard calls (plan + standby)


TheHypnoo:
## Summary

`benchmarks/suite/09_method_calls.ts` (10M calls to a trivial monomorphic `counter.increment()` where `increment()` is `this.value = this.value + 1`) runs in **~3300 ms** vs Node's **~11 ms** — still **~290× slower** after the two improvements already landed/open. The remaining cost is **per-field-access shape-guard calls**, and closing it cleanly requires a GC typed-shape-layout change (high risk). This issue records the full analysis so the work can resume from a known state. Putting it in standby for now.

## What already landed

- **#5084** — typed-feedback site *registration* made opt-in. Removed a 14-arg no-op `js_typed_feedback_register_site` call per property access. Big win for dynamic property access (`bench_object_property` 1127 ms → ~310 ms, 3.6×), but only ~2% on `method_calls`.
- **#5092** — inline small `this.field` methods on exact receivers. Two inliner gaps fixed (loop bodies seeded with empty exact-receiver facts; `is_inlinable` rejecting every method that touches `this`). `increment()` is now inlined into the loop. **`method_calls`: 5229 ms → ~3300 ms (1.6×).**

## Root cause of the remaining ~290×

Per `counter.increment()` iteration, after #5092 the body is inlined but each `this.value` read/write still goes through a **typed-feedback class-field shape guard** — a non-inlined cross-crate call:

- read  → `js_typed_feedback_class_field_get_guard(...)` (`crates/perry-runtime/src/typed_feedback/guards.rs`)
- write → `js_typed_feedback_class_field_set_guard(...)`

When typed feedback is disabled (the default), each guard reduces to `class_field_fast_contract` (`guards.rs:284`): it validates `class_id` + `keys_array` + `field_count`, and for a `number`-typed (raw-f64) field additionally calls `layout_typed_raw_f64_slot_for_user` (`crates/perry-runtime/src/gc/layout.rs:743`), a thread-local `PtrHashMap` (`TYPED_LAYOUTS`) lookup. So per iteration: **2 non-inlined guard calls** (+ the slot load/store the fast path already emits). ×10M ≈ the 290× gap.

[...]

PR #7608, opened 2026-08-07
test(repsel): drive the #7480 element-shape revocation matrix through real mutators

proggeramlug:
Closes the acceptance gap left by #7496: the element-shape invariant shipped with a **revocation matrix that never drove a single mutator**.

## The hole

`element_shape_tests.rs` has one test for the whole bulk-mutator family:

```rust
// `shift`/`unshift`/`splice`/`fill`/`copyWithin`/`reverse`/`sort` all
// mutate slots with bare writes and then land in `rebuild_array_layout`.
let arr = built_from_pushes(CLASS_A, 4);
unsafe { crate::array::header::rebuild_array_layout(arr) };
assert!(proof(arr).is_none());
```

It calls `rebuild_array_layout` **directly**. The comment names seven mutators; the test exercises none of them. Every one could stop reaching that funnel and the suite would stay green — CLAUDE.md's fourth way a gate cannot fail, *"the gate runs but its subject never did"*.

That matters more here than for a normal invariant. The consumers #7480 sequences next (the #5093 versioned-loop clone, then element `Ptr<Shape>`) **hoist a guard out of the loop and emit unguarded reads inside it**. A mutation family that silently stops revoking is not a slow path — it is a miscompile that reads a `B` through an `A`'s layout.

## What this adds

`crates/perry-runtime/src/array/element_shape_matrix_tests.rs` — 25 tests, each driving a **real FFI entry point** against a proven array and asserting the documented verdict. Contract stated once at the top of the file: MAINTAIN / REVOKE / STRUCTURAL REVOKE.

Beyond "the proof is gone", each revoke also asserts **the global epoch advanced** — that is the word a consumer's hoisted guard re-reads, so a revoke the epoch does not advertise is a revoke the consumer misses. And each family that is a *permutation or same-class rewrite* asserts it **re-proves with a fresh identity**, which separates a conservative revoke from genuine heterogeneity and stops a future "optimisation" from replacing a revoke with a no-op.

Two vacuity guards, both added after sabotage-testing the first cut caught nothing:

- `shaped()` asserts the fixture is proven **before** every op, so a step that found nothing to revoke fails loudly instead of passing.
- `reverse`/`sort`/`copyWithin` assert the returned receiver **is** the array that was proven. Without this, returning a fresh (trivially unproven) array satisfies the revoke assertion while revoking nothing.

## Audit result: the invariant is sound

Every element-writing site was traced to a funnel. No unhooked path found — the shipped design holds.

| family | driven through | verdict | covered by |
|---|---|---|---|
| in-bounds store, matching class | `js_array_set_f64` | MAINTAIN | store funnel |
| contiguous append, matching | `js_array_push_f64` | MAINTAIN + extend | store funnel |
| unchecked store (codegen fast path) | `js_array_set_f64_unchecked` | maintain / revoke | store funnel |
| store of a different class / primitive | `js_array_set_f64` | REVOKE | store funnel |
| `delete arr[i]` | `js_array_delete` | REVOKE | `TAG_HOLE` store |
| `arr.length = n` (both directions) | `js_array_set_length` | REVOKE | structural + hole store |
| `pop` | `js_array_pop_f64` | STRUCTURAL | `verified_len` mismatch |
| `shift` / `unshift` | `js_array_{shift,unshift}_f64` | REVOKE | rebuild + structural |
| `reverse` | `js_array_reverse` | REVOKE, re-proves | `rebuild_array_layout` |
| `sort` | `js_array_sort_default` | REVOKE, re-proves | **store funnel**, not rebuild |
| `copyWithin` | `js_array_copy_within` | REVOKE, re-proves | `rebuild_array_layout` |
| `fill` (whole) | `js_array_fill` | REVOKE → heals to the **filled** class | `rebuild_array_layout` |
| `fill` (range) | `js_array_fill_range` | REVOKE, stays mixed | `rebuild_array_layout` |
| **`splice`, equal-length replacement** | `js_array_splice` | REVOKE | **`rebuild_array_layout` alone** |
| `splice` (pure delete) | `js_array_splice` | REVOKE, re-proves | rebuild + structural |
| spread build | `js_array_clone_for_spread` | never inherits source identity | `rebuild_array_layout_exact` |
| `Array.from` family | `js_array_from_values` | establishes, never a wrong class | store funnel |
| JSON.parse / #7539 tape | (audited) | establishes as it fills | store funnel |

Two findings worth recording:

**`sort`'s default path does not use `rebuild_array_layout`.** It is a rank permutation written back through `RootedArrayElems::set`, so it revokes via the *store* funnel. Verified by sabotage, not by reading: removing the revoke from `rebuild_array_layout` leaves the sort test green; removing it from `layout_note_slot` turns it red. Defence in depth, not redundancy — and not what the shipped comment claims.

**Equal-length `splice` is guarded by exactly one thing.** `arr.splice(1, 1, otherClassInstance)` leaves `length` unchanged, so the structural `verified_len` check cannot see it, and the inserted item is written with a bare `ptr::write` that never reaches the store funnel. Only splice's own `rebuild_array_layout` catches it.

## Sabotage evidence

Each funnel was removed in turn and the suite re-run.

| sabotage | result |
|---|---|
| revoke removed from `rebuild_array_layout` | **7 red** — reverse, copyWithin, fill, fill-range, equal-length splice, the roll-up, and the pre-existing proxy test |
| store funnel removed from `layout_note_slot` | **39 red** — nothing can establish |
| revoke removed from **`js_array_splice` only** | **exactly 1 red** — `matrix_splice_equal_length_replacement_revokes` |

The third is the load-bearing one: it shows that test is the *unique* guard for a case no other mechanism can catch. All sabotages reverted; the tree is clean.

## Scope and cost

**Test-only.** The diff is one new `#[cfg(test)]` file plus its `#[cfg(test)] mod` declaration. No runtime code, no emitted code, no bookkeeping added — so standing cost on `push_cls` / `churn_alloc` / `churn_read` is zero by construction, not by measurement, and there is no behaviour to A/B against the gap suite. The invariant's own maintenance cost was #7496's to account for.

GC: unchanged, and the argument holds without new work. The record holds four plain integers and no heap pointer — `class_id` is a registry index that is compared, never dereferenced — so it is correctly absent from `gc_register_mutable_root_scanner`. Element `class_id`s live in `ObjectHeader`s that move *with* their contents, and the record's address key is moved by `transfer_element_shape` from inside `layout_transfer`. The two shipped copying-minor tests in `gc/tests/layout_trace/element_shape.rs` pass.

## Validation

Local, and stated plainly as such — CI's backlog is deep.

- `cargo test -p perry-runtime --no-fail-fast` — 1880 passed, 0 failed
- `cargo test -p perry-codegen --lib` — green
- `cargo fmt --all -- --check` — clean
- `scripts/raw_handle_debt.py` — 998 (baseline 998)
- `scripts/addr_class_inventory.py` — passed
- `scripts/class_id_collisions.py` — passed
- `scripts/check_file_size.sh` — passed

Root-dominance corpus not run: no codegen changes.

Follow-up worth filing separately: `rebuild_array_layout`'s doc comment lists `sort` among the mutators that revoke through it, which the sabotage disproves.

## Summary by CodeRabbit

- **Tests**
  - Added comprehensive end-to-end coverage for array element-shape validation across mutation operations.
  - Verified proof preservation for valid updates and proper revocation for invalid stores, bulk mutations, length changes, mixed arrays, and holes.
  - Added coverage for proof recovery and class restoration after array fills and rebuilds.

- **Documentation**
  - Added a changelog entry summarizing the new validation matrix and audit findings.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1346
+**Current Version:** 0.5.1347
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1346"
+version = "0.5.1347"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1346"
+version = "0.5.1347"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1346"
+version = "0.5.1347"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1346"
+version = "0.5.1347"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1346"
+version = "0.5.1347"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1346"
+version = "0.5.1347"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1346"
+version = "0.5.1347"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1346"
+version = "0.5.1347"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1346"
+version = "0.5.1347"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1346"
+version = "0.5.1347"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1346"
+version = "0.5.1347"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1346"
+version = "0.5.1347"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1346"
+version = "0.5.1347"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1346"
+version = "0.5.1347"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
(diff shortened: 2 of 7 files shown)


proggeramlug:
## Audit before merge — verified, merged as v0.5.1347

**The unique-guard proof reproduces exactly**: removing `js_array_splice`'s own
`rebuild_array_layout` call turns **exactly one** test red —
`matrix_splice_equal_length_replacement_revokes` — confirming that equal-length
splice has a single guard and this matrix is now its only tripwire. That is the
most valuable kind of sabotage evidence: not "the tests can fail" but "this
specific test is the sole coverage for this specific hole".

Full element-shape family 52/52 green, fmt/file-size/raw-handle (998) clean.
Test-only diff, so no root-dominance or perf surface — the "zero cost by
construction" claim is structurally true.

**Folded in at merge**: the one-line doc fix the report flagged and correctly
left out of a test-only PR — `rebuild_array_layout`'s comment claimed the dense
`sort` write-back revokes through it, and #7608's own sabotage proves it does
not (sort revokes through the store funnel via `RootedArrayElems::set`). A
wrong claim about *which funnel revokes* is exactly the kind of comment that
misleads the next #5093 implementer, and it now cites the sabotage that
corrected it.

**On the scope call**: finding the assigned work already merged (#7496) and
pivoting to "does the shipped thing actually have coverage" — rather than
rebuilding it or padding the PR — was the right judgement. It is the sixth
measure-first catch of this campaign, and the first that found the gap in a
*merged* deliverable's verification rather than in a ticket's numbers. The
#5093 versioned-loop consumer is now genuinely unblocked: unguarded reads in a
cloned loop body are a miscompile if a revocation funnel silently drops out,
and the matrix is what makes that failure visible.

Issue #7559, opened 2026-08-07
gc: 05_closure_capture retains +16.44% with every collector counter at +0.00% — surfaced when the ratchet started measuring again


proggeramlug:
## `05_closure_capture` retains +16.44% with every collector counter unchanged

Surfaced the moment the GC ratchet started measuring again (#7557 — its probes had been skipped entirely since 2026-08-05 because a harness-validation step ran before them).

Of the 10 gating breaches that became visible, most are improvements wearing a two-sided band. **This one is not.** Reported by the agent that restored the ratchet, on two machine classes (the pinned quiet mini and a MacBook), with retention and counters reproducing bit-for-bit across hosts:

| probe | retention | collector counters |
|---|--:|---|
| **`05_closure_capture`** | **+16.44%** | **+0.00% on every one** |
| `02_survivor_promotion` | +2.77% | — |

Same `minor_cycles`, `copied_objects`, `copied_bytes`, `promoted_objects`, `promoted_bytes`, `freed_bytes` — **identical work, more retained**. That combination is the interesting one: it rules out "the collector ran differently" and points at something being *kept alive* that previously was not. A closure-capture probe holding 16% more is exactly the shape of an over-broad root or a capture cell that stopped being released.

## Why it was invisible

The ratchet did not run between 2026-08-05 and #7557. In that window roughly fifteen GC-affecting changes landed: #7532 (typed-shape layout at allocation), #7535 (shape-install memo), #7536 (write-barrier gating), #7546 (spill-path value rooting), #7553 (tape out of the old generation), #7550/#7552 (for-init type propagation), plus the rooting stack #7495/#7516/#7527/#7529.

**Several of those add rooting.** #7516 alone rooted 14 sites, #7527 more, #7529 three more — and a root that is broader or longer-lived than necessary retains. That is the first place I would look, and it is worth saying plainly: these fixes were correct and must not be reverted (they closed real use-after-frees), but "more correct rooting" and "more retention" are the same lever pulled in opposite directions.

## Suggested method

Bisect the probe's retention across that window with the ratchet itself now that it works — `gc_ratchet.py measure --probe 05_closure_capture` at each candidate commit on the pinned host. The counters being flat means a `PERRY_GC_DIAG` trace diff between the endpoints should show the difference as a retained-object delta rather than a cycle-behaviour one.

[...]

Issue #7558, opened 2026-08-07
gc(): the forced conservative stack scan makes every retained-heap reading nondeterministic, and 16% too large on a 50 MB live set


proggeramlug:
Split out of #7554, which asked whether `12_large_live_set`'s retention
non-determinism is benign or a real finding. It is a real finding, and it is
not about the probe.

## The metric is measured after a deliberately conservative collection

Every GC-ratchet probe reads `process.memoryUsage()` after an explicit `gc()`.
An explicit `gc()` runs a full mark-sweep with a **forced conservative stack
scan** — `PERRY_GC_DIAG=1` prints this on every single run:

```
[gc-scan-fallback] site=manual_collect automatic=false count=1
```

A conservative scan retains whatever the native stack happens to look like a
pointer to. Stack residue at that moment differs between runs, so the number
the probe reports differs between runs.

## The evidence

Diffing two full `PERRY_GC_DIAG` traces of the same binary that disagree: the
minors, the tenuring decisions, the step cycles and every copy/promote counter
match **exactly**. The only difference is in the *last* collection:

```
< [gc] blocks: … freed_bytes=87675232 …      < [gc-old-free] reusable_bytes=45397168
> [gc] blocks: … freed_bytes=87677896 …      > [gc-old-free] reusable_bytes=45399760
< #gcmetric heap_used_bytes=59945816         > #gcmetric heap_used_bytes=59943224
```

And turning the scan off removes the variance completely:

| arm | `heap_used_bytes` |
|---|---|
| default | 59,943,080 … 59,952,824 over 22 runs, 5+ distinct values |
| `PERRY_CONSERVATIVE_STACK_SCAN=off` | **51,668,688 × 8 runs, bit-identical** |

Two conclusions:

1. **The run-to-run variance is entirely false roots.** A few kilobytes —
   bounded by how much a handful of stale stack words can pin.
2. **The conservative scan retains 8,275,208 bytes systematically — 16% of what
   this probe reports as its retained heap.** That is not noise, it is a
   constant tax on every `gc()`-then-measure reading.

The eleven other probes stay bit-identical because their live sets are one to
two orders of magnitude smaller, so a stale stack word is far less likely to
alias a plausible heap address at all. This is a size effect, not a property of
what `12_large_live_set` does.

Reproduced three times independently on 2026-08-06/07 at `be1fc80f8`
(0.5.1315): pinned quiet host `perry-macos` spread 2,304 over 7 harness repeats
[...]

PR #7609, opened 2026-08-07
test(gc-ratchet): fail open per cell and re-pin the baseline (#7554)

proggeramlug:
Closes the #7554 repair: makes an artifact defect cost one cell instead of the
whole gate, re-pins the baseline on the pinned host at current `main`, and shows
the restored gate going red and green on demand.

## The premise moved — what is actually broken

#7554 says the ratchet dies in preflight. That was true, and **#7557 already
fixed it**. Preflight passes at `main` today (63 unit tests + `validate`, exit 0,
verified locally). The failure is now one step later and one week longer:

| | |
|---|---|
| last green `gc-ratchet` on `main` | **2026-08-01T05:39Z** (run `30686263856`) |
| since then | **179 consecutive red `main` runs**, 15 cancelled, 0 green |
| 2026-08-05 → #7557 | red in *preflight* — the probes never ran at all |
| #7557 → now | red in `check`, against a `0.5.1280` artifact that no longer describes the collector |

So the gate has produced no actionable verdict for a week, and because
`gc-ratchet` is **not a required context** (see below) none of it blocked
anything. #7594 and #7596 both substituted hand-run both-arms A/Bs.

## 1. Fail open per cell, fail closed on the verdict

Artifact validation aborted on the first defect, and it runs *before* the
measurement step. One cell — `12_large_live_set.heap_used_bytes`, spread 6,768
bytes — therefore meant **none of the twelve probes executed on any branch for
three days**. The claim that caused it was about one cell; nothing about it
voided the other 143 or made the probes unrunnable. That blast radius was never
chosen, it was inherited from raising an exception.

Defects now carry a scope:

| scope | examples | voids |
|---|---|---|
| `artifact` | wrong schema, missing metric, summary disagreeing with its own samples | everything — still fatal, still preflight |
| `probe` | pinned without an oracle diff, pinned with no collection | that probe |
| `cell` | spread ≠ 0 where the band's premise is bit-identity | that cell |

`probe`/`cell` defects **demote** their subject out of the gating family and are
reported as failures. `check` still measures all twelve probes, still evaluates
the rest of the matrix, and still names a regression elsewhere — while the defect
keeps the job red.

CI preflight now runs `validate --scope structural`, which fails only on the
fatal kind. `assemble` is deliberately unchanged: pinning still refuses **any**
defect, so this cannot be used to freeze a new unfit artifact.

**Why this is not suppression.** `check` re-derives the same defect list and
fails on every entry, and a test asserts that coupling per planted defect shape.
Sabotage evidence below.

## 2. Baseline re-pinned at `main` 26b9c9d59 (0.5.1346)

Measured on `perry-macos` (Mac mini M1, 8 GB) — the *same* host and the same
rustc/cargo/clang as the 2026-08-05 pin, so like-for-like. Load 2.38/1.9/1.7 at
capture (the previous pin was taken at 2.01/**3.2**/**7.57**). All 12 probes
oracle-pass; `heap_used_bytes` spread **0** on eleven probes and 864 B on
`12_large_live_set`. Full attribution is written into the artifact's own `notes`.

### Cells that moved

**(a) Explained — collection pacing (#7594, #7596).**

| probe | metric | baseline → current |
|---|---|---|
| `03_cross_gen_writes` | copied_objects | 13,893 → 8,212 (−40.9%) |
| `03_cross_gen_writes` | copied_bytes | 990,736 → 590,688 (−40.4%) |
| `03_cross_gen_writes` | promoted_objects | 4,752 → **0** |
| `03_cross_gen_writes` | promoted_bytes | 210,736 → **0** |
| `04_dead_after_deep_stack` | copied_objects | 11,268 → 565 (−95.0%) |
| `04_dead_after_deep_stack` | copied_bytes | 663,512 → 44,688 (−93.3%) |
| `04_dead_after_deep_stack` | promoted_objects | 4,752 → 10 |
| `04_dead_after_deep_stack` | promoted_bytes | 210,744 → 440 |

Recorded rather than glossed: `03`'s `promoted_*` now pin at **0**, where the
allowance floor (16 objects / 64 KiB) covers the whole range *and* the liveness
assertion fires only when the baseline median is > 0. That cell no longer carries
signal in either direction. It is the price of pinning a counter at zero.

**(b) Explained — measurement, not retention (#7558, #7571).**

| probe | metric | baseline → current | `classify` precise |
|---|---|---|---|
| `02_survivor_promotion` | heap_used_bytes | 9,418,232 → 9,678,792 (+2.77%) | **9,416,632** — *below* the old baseline |
| `05_closure_capture` | heap_used_bytes | 6,378,392 → 7,426,960 (+16.44%) | **5,329,880** — byte-identical to #7571 |

`05`'s precise retention is the exact figure #7571 measured at **both** ends of
its 74-commit window; the conservative residue went 1 block → 2 (2,097,080 B)
while real retention did not move at all. `02`'s precise reading is *lower* than
the number this artifact previously recorded as that probe's retention, so real
retention cannot have grown. This is #7559's answer, reproduced independently on
the pinned host rather than assumed.

**(c) FLAGGED — not explained by any merged, documented decision.**

`12_large_live_set.wall_ms` **3,056 → 3,471 ms (+13.58%)**. Two non-overlapping
7-sample clusters (3,047–3,061 vs 3,466–3,476), same host, same toolchain, same
protocol — while `06_string_retention` and `11_collect_at_depth` got 9.6% and
28.4% *faster* over the same window. #7596 reported **−7.4%** on this very cell
in its own both-arms A/B, so by that PR's own evidence this is not #7596. Gated
under `pinned_host` only, so it does not block CI. Pinned here so the rest of the
matrix can gate again; it wants a bisect over 0.5.1280..0.5.1346.

### Note on the diff size

`baseline/gc-ratchet-v1.json` is `+684 / −3,057`, but **2,374 of those deleted
lines are the `suite` block** (a `benchmarks/compare.sh` run recorded alongside
the previous pin), now `null`. The probe set is unchanged and every top-level key
is still present. `compare.sh` needs a full checkout and a Node/Bun toolchain;
this pin measured shipped binaries on the mini, so the suite was not run. The
driver supports exactly this via `--no-suite`, and the omission is recorded in the
artifact's `notes` rather than left for a reader to notice. Re-running it later is
a re-pin, not a patch.

**(d) Did not reproduce.** #7596's merge audit accepted
`12_large_live_set.heap_total_bytes` +36% (95.4 → 130.0 MB) and deferred the
re-pin to this repair. Under the harness protocol on this host that cell is
**110,100,480 → 110,100,480, +0.00%** — and neither endpoint of the +36% figure
matches this artifact's reading. Nothing was re-pinned on account of it.

## 3. Sabotage evidence (real exit codes)

**Harness mechanism** — two ways to break it, both caught:

| sabotage | result |
|---|---|
| revert `evaluate` to abort on any defect (the #7554 collapse) | **6 tests ERROR** |
| demote the cell but drop the defect from `failures` (fail-open → suppression) | **5 tests FAIL** |
| restored | **71 tests OK** |

**Preflight scoping**, on an artifact carrying a planted 6,768-byte spread:

| arm | exit |
|---|---|
| `validate --scope all` | **1** |
| `validate --scope structural` | **0** (deferred, printed loudly) |
| `check` — 144 rows still evaluated, unfit cell demoted, **and an unrelated planted `freed_bytes` −50% regression still named** | **1** |
| tampered summary at `--scope structural` (integrity, not fitness) | **2** — still fatal |

That third row is the whole point: under the old behaviour that `freed_bytes`
regression was invisible, because nothing ran.

**The re-pinned gate itself**, end to end on the pinned host — sabotaging the
*collector*, not the JSON:

| arm | cells moved | `check` exit |
|---|---|---|
| clean, `pinned_host` | — | **0** |
| clean, `shared_ci` | — | **0** |
| `PERRY_GEN_GC=0` (non-generational) | **79** | **1**, 90 findings, liveness assertions fire by name |

⚠️ **A vacuous sabotage, reported rather than hidden.** My first attempt was
`PERRY_GEN_GC_EVACUATE=0`, which moved **0 cells across all 12 probes** and left
the gate green — measured, not inferred: a cell-by-cell diff of the two
measurements reports `cells that moved: 0`. So the knob changes nothing this
ratchet observes, and "passes with evacuation policy disabled" has never been
evidence about this suite. (Plausibly because it gates the C4b *policy*
evacuation of tenured objects rather than the copying minor that `copied_objects`
counts — but that attribution is a reading of the code, not something these runs
establish.) That is the
#6942/#7024 pattern (a knob that disables something other than what it is
believed to disable) and it is worth its own look; per CLAUDE.md's kill-policy an
off-state nothing exercises should not survive. It also means "passes with
evacuation policy disabled" has never been evidence about this suite.

## 4. CI workflow audit — the four hazards

| hazard | status |
|---|---|
| 1. `continue-on-error` | **clean** — none on any gating step; `if: always()` only on artifact upload |
| 2. required context | ❌ **BROKEN** — required contexts are `lint, cargo-test, parity, compile-smoke, api-docs-drift, security-audit, conformance-smoke-complete`. **`gc-ratchet` is absent.** 179 red main runs blocked nothing. |
| 3. concurrency cancellation | **clean** — per-commit group for `push` (#7205); main runs queue and complete |
| 4. subject liveness | **clean** — `check` asserts `minor_cycles > 0` and `copied_objects > 0` rather than inferring from bands, and the `PERRY_GEN_GC=0` arm above proves it fires. Caveat: `03.promoted_*` pinned at 0 is a cell that can no longer fail (called out in §2a). |

**Hazard 2 needs admin action and I have not attempted it.** Promoting
`gc-ratchet` to a required context is an admin-only branch-protection change, and
CLAUDE.md's corollary applies: promote *after* a green `main` run, not before, or
every open PR is blocked. This PR is the prerequisite — it is what makes a green
`main` run possible for the first time since 2026-08-01.

## Validation

CI has a deep runner backlog (`main` push runs from 22:07Z onward are still
queued), so **the evidence here is local and I am saying so plainly**:

- `python3 -m unittest discover -s tests -p 'test_gc_ratchet.py'` — **71 passed**
  (63 pre-existing + 8 new), plus both sabotages above
- `gc_ratchet.py validate --scope all` on the new artifact — exit 0
- `gc_ratchet.py check` on the pinned host, both profiles — exit 0
- `raw_handle_debt.py` 998 = baseline · `addr_class_inventory.py` pass ·
  `class_id_collisions.py` pass · `check_file_size.sh` pass ·
  `cargo fmt --all -- --check` clean
- No Rust changed, so `cargo test -p perry-runtime` is not implicated; the
  measured collector is exactly `origin/main` `26b9c9d59`

## Follow-ups (not in this PR)

1. **Admin: promote `gc-ratchet` to a required context** after its first green
   `main` run.
2. **`12_large_live_set.wall_ms` +13.58%** — filed as **#7610**, bisect 0.5.1280..0.5.1346.
3. **`PERRY_GEN_GC_EVACUATE=0` is inert on this suite** — filed as **#7611**;
   delete the off-state or give it an arm that exercises it (CLAUDE.md kill-policy).
4. **#7558's real fix**: `heap_used_bytes` has a 1 MiB noise quantum and a
   127 KiB band on some probes, so re-pinning only resets the clock. #7571's own
   conclusion is that the fix is to measure a different quantity — gate the
   `classify` *precise* reading instead of the conservative one. That is a
   metric-family redesign and a maintainer call, deliberately not smuggled in
   here.

### DIFF
--- a/.github/workflows/gc-ratchet.yml
+++ b/.github/workflows/gc-ratchet.yml
@@ -78,10 +78,27 @@ jobs:
 
       # Fast structural failure first: a broken harness or a tampered artifact
       # should not cost a 20-minute compiler build to discover.
+      #
+      # `--scope structural` is load-bearing and is the #7554 repair. This step
+      # runs BEFORE the measurement step, so anything it fails on costs the
+      # entire run's coverage — twelve probes that never execute. That price is
+      # correct for "this artifact is unreadable or tampered with" and badly
+      # wrong for "one of its 144 cells is not bit-identical", which is what
+      # actually happened: one cell's 6,768-byte spread failed this step and the
+      # ratchet measured NOTHING on any branch for three days, while two GC
+      # pacing changes (#7594, #7596) merged with hand-run A/Bs in its place.
+      #
+      # Under `structural` a per-cell defect is printed loudly and deferred; the
+      # `check` step below re-derives the same defect list and fails on it, after
+      # the probes have run and with the full table attached. So this cannot
+      # wave a defect through to a green job — `check` is where the verdict is,
+      # and tests/test_gc_ratchet.py's
+      # `test_structural_preflight_defers_every_defect_it_waves_through` asserts
+      # that coupling one planted defect shape at a time.
       - name: Harness unit tests and artifact validation
         run: |
           python3 -m unittest discover -s tests -p 'test_gc_ratchet.py' -v
-          python3 benchmarks/gc_ratchet/gc_ratchet.py validate
+          python3 benchmarks/gc_ratchet/gc_ratchet.py validate --scope structural
 
       - name: Decide whether this change can affect the collector
         id: relevance
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1347
+**Current Version:** 0.5.1348
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1347"
+version = "0.5.1348"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1347"
+version = "0.5.1348"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1347"
+version = "0.5.1348"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1347"
+version = "0.5.1348"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1347"
+version = "0.5.1348"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1347"
+version = "0.5.1348"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1347"
+version = "0.5.1348"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1347"
+version = "0.5.1348"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1347"
+version = "0.5.1348"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1347"
+version = "0.5.1348"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1347"
+version = "0.5.1348"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1347"
+version = "0.5.1348"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1347"
+version = "0.5.1348"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1347"
+version = "0.5.1348"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1347"
+version = "0.5.1348"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1347"
+version = "0.5.1348"
 dependencies = [
  "block2",
  "objc2",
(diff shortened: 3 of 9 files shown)


proggeramlug:
## Audit before merge — verified, merged as v0.5.1348

**The premise correction is the finding.** I briefed this as "preflight fails,
three days dark"; the truth is **179 consecutive red main runs since
2026-08-01**, with the preflight abort merely an episode inside it (#7557 fixed
that; the rest was `check` against a v0.5.1280 artifact describing a collector
that no longer exists). And because `gc-ratchet` is not a required context,
none of the 179 blocked anything. That is gate hazard 2 with a week of paper
trail — the required-context promotion is admin action, correctly not
attempted, and per the corollary it follows the first green `main` run, which
this PR enables.

**Sabotage re-verified**: silencing the demoted-defect reporting (turning
fail-open back into suppression) fails the harness suite — 11 FAIL/ERROR lines,
exit 1; restored, exit 0. The design line worth quoting: *fail open per cell,
fail closed on the verdict* — a demoted cell still turns the run red, it just
does so after the other 143 cells were measured, so a regression elsewhere is
named in the same run rather than hidden behind the abort.

**Two attribution results deserve the record:**

1. **#7596's accepted `heap_total_bytes` +36% did not reproduce on the
   re-pinned baseline** (+0.00%, byte-identical). My merge decision on #7596
   accepted that trade on the strength of a dev-Mac both-arms A/B; the pinned
   mini at current main does not show it. The decision cost nothing, but the
   discrepancy is a live reminder that the dev-Mac A/B and the pinned-host
   baseline are different instruments — the re-pinned artifact is now the
   authority.
2. **The flagged +13.58% on `12_large_live_set.wall_ms` (#7610) is exactly the
   right thing to refuse to explain away.** #7596 claimed −7.4% on that very
   cell; the re-pin measures +13.58% with non-overlapping sample clusters while
   other probes got faster. Unexplained, filed, and now *gated* — the next
   change that touches it goes red.

**The honest failure is worth more than the sabotage table**: the agent's first
sabotage (`PERRY_GEN_GC_EVACUATE=0`) moved **zero cells across all 12 probes**
and left the gate green — reported rather than hidden, filed as **#7611**, the
third instance of the #6942/#7024 inert-knob pattern and now subject to the
kill-policy. Meanwhile the `PERRY_GEN_GC=0` arm moved 79 cells and exited 1,
which is the liveness proof the gate needed.

Gates re-run here: harness suite green, validate structurally valid on the
re-pinned baseline, all four lint scripts + fmt clean, no Rust changed.

**Owner action queued**: promote `gc-ratchet` to required contexts after the
first green `main` run of this workflow.

PR #7566, opened 2026-08-07
perf(codegen): take the inline bump allocator at `new` sites inside loops (#7469)

proggeramlug:
#7469 workstream A. Takes the inline bump allocator at `new` sites **inside loops** and leaves everything else outlined.

## Why the default was worth revisiting

The outlined allocator has been the default since the `[#bloat]` work, on this measurement: *"win-win vs inline: −45 IR lines/site AND ~17% faster on an 8M-allocation loop (the inline bump bloated the hot loop, hurting icache/regalloc more than the saved call)."*

**The size half still holds.** I re-measured it: ~268 bytes of machine code per site (+0 / +49,536 / +214,656 bytes across 10 / 200 / 800-site programs).

**The speed half has inverted.** Outlined is now **1.81× slower** on `churn_alloc` and 1.78× on `push_cls`. Nothing about the inline bump changed — everything *around* the allocation got cheaper (#7474, #7486, #7487, #7501, #7525, #7532, #7535, #7536, #7552), so the surviving FFI call and the thread-local resolutions it performs now dominate what its bloat costs.

## Why not just make TLS cheaper

Recorded so it is not attempted again: **Mach-O has no local-exec TLS model.** I built the entire runtime with `-Ztls-model=local-exec`; the `blr` through the TLV descriptor in `layout_forget_object` is **byte-identical** and `churn_alloc` moved 1.02×.

Per-call price is already at the plain-global floor:

| access | ns/call |
|---|--:|
| plain global | 1.24 |
| Rust `const`-init `thread_local!` | 1.29 |
| C `__thread` | 1.31 |
| `pthread_getspecific` | 2.11 |

What is expensive is the **count** — ~14 resolutions per allocation, because each FFI helper resolves independently and LLVM cannot CSE across the FFI boundary. Inlining removes the allocator's outright.

## The gate

Per site rather than global. Loop membership is the cheapest sound proxy for "runs many times" and bounds the size cost to loop bodies. It reuses the existing `loop_targets` stack rather than adding a counter: `switch` frames push an **empty** continue label while every loop pushes a real one — the same discriminator `Stmt::Continue`'s scan-outward-past-switch-frames logic already uses — so a `new` in a bare `switch` is correctly not-in-a-loop.

| bench | outlined | gate | ceiling (all-inline) |
|---|--:|--:|--:|
| `churn_alloc` | 1.32 s | 0.73 s (**1.81×**) | 1.81× |
[...]

PR #6377, opened 2026-07-13
perf(codegen): keep the numeric-array specialization when the array is captured (#6369)

proggeramlug:
Fixes #6369.

A `number[]` got the guarded packed-numeric access path when it arrived as a **parameter**, but fell all the way back to the fully generic `js_dyn_index_get` when the same array was reached through a **capture** — a module-scope `const` read inside a function or an arrow closure. Same declared type, same loop, **27x**.

That is the shape bundled code is made of (a large esbuild-bundled CLI app is overwhelmingly module-scope `const`s captured by closures), so in practice almost nothing was reaching the fast path — a captured *typed* array was no faster than an untyped one, i.e. the annotation bought nothing once the value was captured.

## Root cause

The declared type was there all along; it just never reached the read sites. Three seams dropped it:

**1. The closure receiver oracle.** `FnCtx.local_types` is what `static_type_of` / `is_array_expr` consult to pick a specialized path. `compile_function` and `compile_method` seed it from the unfiltered `module_global_types` — but `compile_closure` seeded it from `module_local_types`, a map that two filters had already stripped:

* #5869 removes **boxed** ids, because advertising a boxed id's type made the *typed-ABI closure specializations* read the capture RAW while the generic variant `box_get`s.
* #5982/#6039 removes **module globals**, because advertising those made the typed-ABI closure read an unset capture slot.

Both hazards are about the typed-ABI **capture representation**, not about the receiver type — and #6039 had already scoped its filter to a dedicated copy (`typed_abi_local_types`) for exactly this reason, leaving a comment explaining that `module_local_types` is *also* the receiver oracle. The boxed filter never got the same treatment. So every captured binding lost its declared type inside a closure, and `arr[i]` reached the lowering as an unknown receiver.

Fixed by scoping the boxed filter the same way #6039 scoped the module-global one: closures now seed `local_types` from an unfiltered `module_receiver_types`, and every typed-ABI path keeps the filtered maps untouched.

**2. The packed-loop array facts.** #6011 seeds a packed-numeric array *kind* for a **param** purely from its declared type, on the explicit grounds that the loop's entry guard re-validates the actual runtime array and body-observed mutations still downgrade the seed. A module-scope binding sits in exactly that position, so it is seeded the same way — and the two packed-loop matchers now version on it, instead of rejecting every module global outright.

[...]

PR #7612, opened 2026-08-07
perf(repsel): element-shape versioned loop clone — the first consumer of the element-shape invariant (#7480 / #5093)

proggeramlug:
The consumer half of #7480; continues #5093's versioned-loop line.

#7496 landed the per-array homogeneous element-shape invariant with **no
consumer, on purpose**. This is the consumer.

`for (let j = 0; j < n; j++) sum += keep[j].v` now gets a specialized clone
behind a preheader guard on "this array holds the element-shape invariant at
class `C`". Inside the clone the element read is a bare `gep + load` off a
preheader-cached elements base and the field read is a bare raw-f64 slot load.
The existing generic body survives unchanged as the cold arm.

| kernel (200k elements × 50 sweeps, best-of-9 interleaved, checksums equal) | before | after | node | speedup |
|---|--:|--:|--:|--:|
| `keep: Node[]`, `sum += keep[j].v` | 41 ms | **13 ms** | 13 ms | **3.15×** — parity with node, from 3.15× behind |
| `keep: {v,w}[]` (object literal) | 86 ms | 86 ms | 10 ms | 1.00× — deliberately out of scope |

## Revocation mechanism, and its failure mode

**Restrict-the-body**, enforced twice — the second enforcement is the
load-bearing one.

1. **By shape (the matcher).** A single store-free `acc = <pure numeric>`
   statement over `arr[counter].field` reads, numeric locals, numeric literals
   and pure arithmetic / `Math`. No stores, calls, closures, `await`, or
   updates other than the counter's. No catch-all arm.
2. **By construction (the lowering).** After the fast clone is emitted, every
   one of its blocks is scanned for a GC-unsafe call. If any survived, the
   deref block branches **unconditionally** to the slow clone and the fast
   blocks are left as unreachable code.

Call-freeness is exactly the right property, because every way to revoke the
invariant is a runtime call — element store (`layout_note_slot` →
`note_element_store`), length change (caught by the record's pinned
`verified_len`), `delete` (a `TAG_HOLE` store through the same funnel),
`defineProperty` on the array, prototype surgery — and so is every allocation
that could move the array. Codegen's *inline* element store is the one path
that can skip the note, and only when the array is statically proven numeric
and pointer-free, which an element-shape array can never be.

**Failure mode: conservative, never unsound.** Anything that writes, calls, or
reads a field the analysis cannot type gets no clone and runs exactly as
today. The residual risk is a *silent loss of the optimization*, which is why
the codegen tests assert the fast blocks appear in the emitted IR **and** that
the fast clone contains no `call` at all.

A useful consequence of the same rule, verified in the emitted IR: under
`PERRY_GC_MOVING_LOOP_POLLS=1` the back-edge safepoint is itself a call, so the
scan fails and the deref block emits an unconditional
`br label %element_shape.loop.slow.preheader`. The clone stands down in exactly
the configuration where a mid-loop collection could move the array — with no
special case for it anywhere in the code.

## The guard tests the live header; the brand is explicit

The preheader calls `js_array_ensure_element_shape` — #7496's own query
surface, which reads the array's current `GcHeader` bit and record and
self-heals when the record went stale. No inline reimplementation, so no drift
(#7501).

Sequencing is load-bearing. The `GC_TYPE_ARRAY` brand test comes **first**, so
the pointer handed to the runtime is already branded — an `Array` subclass
instance is a plain `ObjectHeader` whose fields overlay `ArrayHeader`'s
(#7573/#7603). The elements base pointer is derived only **after** the guard
call returns, from a fresh load of the array's rooted slot: the call can
allocate, and an allocation can move the array.

## What the invariant does not prove

`element_class_of_bits` proves `POINTER_TAG`, a readable `GcHeader`,
`GC_TYPE_OBJECT`, `OBJECT_TYPE_REGULAR` and `class_id == C` for every element
in the verified prefix — exactly the predicates the element-read tier and the
front half of the field-read precheck spend per iteration, so the clone drops
them. It proves nothing about the per-*object* facts a raw-f64 slot load needs
(`keys_array` identity — `delete elem.f` compacts the packed slots while
preserving `class_id` — plus `field_count`, the descriptor flag, and the
typed-layout intact bit). Dropping those would be the miscompile, so the clone
keeps a residual per-element check, collapsed to one 4-byte load of the three
contiguous header bytes plus two more loads and a single side-exit branch.
Emitted fast body: **zero calls, one branch, no volatile gate load.**

Folding those facts into the invariant is the documented next slice and needs
an invalidation surface for `delete` / `defineProperty` / typed downgrade that
#7496 deliberately did not open. It should land the way #7496 did: invariant
first, matrix second, consumer third.

## Sabotage, both directions

| arm | clone blocks in IR | `js_array_ensure_element_shape` calls | gap test |
|---|--:|--:|---|
| base (`main`) | 0 | 0 | identical to node |
| this | 1 of each | 1 | identical to node |
| sabotage: **never**-specialize | 0 | 0 | identical to node |
| sabotage: **always**-specialize | 1 of each | 1 | **SIGBUS, exit 138** |

Breaking the guard (every shape fact discarded, per-element check never
side-exits) faults at the `subclass:` case — #7603's `ObjectHeader`-read-as-
`ArrayHeader` reproduced on demand. Breaking the clone *selection* leaves the
gap test byte-identical to node while the census drops to the same zero the
base compiler emits, so the fallback is behaviour-neutral and the census is
measuring the clone rather than something incidental.

## Size (#7566's discipline)

Both recorded traps avoided — runtime trip counts so nothing unrolls, arrays
escaping through `console.log` so nothing is scalar-replaced. Measured on the
module object file so the runtime archive does not blur it.

| probe | base | this | delta |
|---|--:|--:|--:|
| 40 loops, **none** qualifying | 83,000 B | 83,000 B | **+0 B, object bytes byte-identical** |
| 40 loops, all qualifying | 93,904 B | 108,872 B | +14,968 B (+15.9%, ~374 B/loop) |

Runtime archive +232 B for the single `keepalive-anchors` static — exactly one,
for the one symbol codegen now emits a call to; the other four
`js_array_element_shape_*` entry points stay unanchored and dead-strippable.

## Verification

- **Gap test byte-identical to node, exit 0** —
  `test-files/test_gap_repsel_element_shape_loop_clone.ts` covers the hot
  shape plus mid-loop store revocation (direct and via a call), revocation
  between two entries of the same loop, subclass receiver (`const` and
  plainly-typed parameter), shape-mismatched and heterogeneous arrays, holes /
  sparse / `delete`, empty array, per-element typed-layout downgrade, deleted
  field, own accessor, frozen element, prototype surgery, every length
  mutation, and a bound past the array's length.
- **Gap-suite family A/B** (`array` / `object` / `class` / `repsel` / `new` /
  `prop`, 71 tests) against a same-session `main` build: **verdict sets
  identical** — 70 PASS / 1 FAIL on both arms, the failure
  (`test_gap_prop_plan_cache_invalidation`) pre-existing on both. This is
  #6377's gate.
- **GC root-dominance**, curated corpus (129/129 sources compiled, 149
  modules), both gated modes: `--moving-only` 0 violations with **40/40 seeded
  violations caught**; `--unrooted-allocas --moving-only` 0 violations over
  7,860 GC-capable allocas. Allowlist empty and stays empty.
- `cargo test -p perry-runtime --no-fail-fast` 1880 passed / 0 failed;
  `cargo test -p perry-codegen --lib` 685 passed / 0 failed (7 new).
- `cargo fmt --all --check`, file-size cap, addr-class ratchet, GC store-site
  inventory, workspace-architecture policy: all clean.

## Scope

Declared element types (`keep: Node[]`) only. #7480's own object-literal kernel
(`keep: {v,w}[]`) is deliberately out: `receiver_class_name` returning `None`
for an `Object`-typed element is also what makes the number-context field-read
helper decline, so a wider matcher would buy only dead fast-clone IR. Reaching
it means teaching `static_type_of` / `receiver_class_name` to type an
`Object`-typed property read — the #6377 "more type visibility un-gates latent
fast paths" change, which needs its own gap-suite A/B. Element classes with a
base class are also declined: an inherited layout is not described by the
packed slot index alone, and a native base (`extends Array`) is the
#7573/#7603 hazard itself.

## Summary by CodeRabbit

* **New Features**
  * Added an optimized execution path for numeric loops over compatible homogeneous object arrays.
  * Added automatic safety checks that fall back to standard loop behavior when array layouts, contents, mutations, or other conditions are unsupported.

* **Bug Fixes**
  * Improved handling of array shape changes, inheritance, sparse or heterogeneous arrays, accessors, and runtime mutations.

* **Tests**
  * Added comprehensive coverage for optimized loops, fallback scenarios, bounds changes, and invalidated assumptions.


proggeramlug:
## Audit before merge — verified, merged as v0.5.1350

**The guard is verified in depth, and my two sabotages mapped its structure:**

1. Forcing the **brand test** always-true → **exit 0, output still identical to
   node**. The shape-id query behind it correctly declines the subclass
   receiver (no record for an `ObjectHeader` address), so this layer is
   redundant for that case — defense in depth doing its job. (The agent's
   SIGBUS-on-demand sabotage evidently cut deeper than mine did.)
2. Forcing the **query verdict** true → **27 lines diverge at exit 127** — the
   fast clone running on non-conforming arrays produces wrong values and the
   gap test names them. That is the decisive layer, and it can fail.

Together: each layer is individually load-bearing for its own case, and no
single mutation of the guard survives the gap test.

**One honest miss in my audit worth recording**: my hand-rolled perf probe got
**zero clone blocks** (its shape didn't qualify — it was already served by an
older fast path) while still running fast, which would have been a vacuous
"perf confirmed" if I had only timed it. The IR census — `element_shape.loop.*`
block labels — is the discriminator, and the PR's 7 codegen tests assert it
across all four arms (fires for the #7480 shape, declines offset-index,
declines storing bodies, declines subclass element types, +0 for no-qualifying
programs). All 7 green here.

**Root-dominance re-run** (codegen change): corpus 129/129, `--moving-only` 0
violations with 40/40 seeded caught, `--unrooted-allocas` 0 over 7,860 allocas.
Full suites 1,886/0 + 684/0; all four lint gates + fmt clean.

**The two design decisions that earn the merge:**
- **Call-freeness as the revocation-safety property** — every revocation funnel
  is a runtime call, so a clone with no calls cannot observe a revocation
  mid-flight; and the post-emission scan that demotes any clone containing a
  surviving GC-unsafe call to an unconditional slow branch converts "matcher
  bug" from unsoundness into silent deoptimization, which the census tests then
  make loud. Under `PERRY_GC_MOVING_LOOP_POLLS=1` the poll is itself a call, so
  the clone stands down in exactly the configuration where the array could move
  — with no special case in the code.
- **The scope cut**: implementing, measuring, and then REMOVING the
  object-literal arm because `receiver_class_name` returning `None` makes it
  emit only dead fast-clone IR — and naming the reason (#6377's type-visibility
  hazard needs its own PR and A/B) rather than shipping a wider matcher that
  looks like coverage. The extension plan is on #5093/#7480.

The 6.2× `keep[j].v` shape is now at **parity with node**. The repsel route's
remaining step (element `Ptr<Shape>`, then object-literal elements) has its
prerequisite, its matrix (#7608), and its first consumer all merged.

Issue #7611, opened 2026-08-07
gc: PERRY_GEN_GC_EVACUATE=0 moves zero cells on all 12 gc-ratchet probes — the knob's off-state is unexercised


proggeramlug:
Found while trying to sabotage-verify the repaired GC ratchet (#7609, for #7554).
The intent was an end-to-end proof that the restored gate can go red: disable an
evacuation knob, re-measure, watch the counters collapse. It did not collapse
anything.

## The measurement

Pinned quiet host `perry-macos`, `main` `26b9c9d59` (0.5.1346), identical
binaries and protocol in both arms (`--repeats 7`, warmup 1,
`PERRY_NO_AUTO_OPTIMIZE=1`, prebuilt static archives), the only difference being
the knob:

```
arm A: PERRY_RUNTIME_DIR=... PERRY_NO_AUTO_OPTIMIZE=1
arm B: PERRY_RUNTIME_DIR=... PERRY_NO_AUTO_OPTIMIZE=1 PERRY_GEN_GC_EVACUATE=0
```

Cell-by-cell diff of the two measurements over all 12 probes and
`minor_cycles`, `step_cycles`, `copied_objects`, `copied_bytes`,
`promoted_objects`, `promoted_bytes`, `freed_bytes`, `heap_used_bytes`:

```
cells that moved: 0
```

**Zero.** Not "small deltas inside the band" — bit-identical medians on every one
of the 96 cells. `gc_ratchet.py check` against the pinned baseline returns **exit
0** with the knob set.

For contrast, the same procedure with `PERRY_GEN_GC=0` moved **79 cells** and
returned exit 1 with 90 findings, so the method is sound and the harness is
sensitive — this knob specifically changes nothing the ratchet can see.

## Why this matters

CLAUDE.md's GC knob kill-policy is explicit: *"Every GC env knob either has a
required CI arm exercising its OFF state, or it is deleted after one release of
soak… A mode that still exists is a decision that hasn't been made."* It also
records two precedents of exactly this shape:

- `PERRY_GC_FORCE_EVACUATE` was **inert** for every `gc()`-driven test
  (#6942/#6946) — months of "passes under evacuation" meant nothing.
- The matrix's `--pressure` knob **disabled the very path it was measuring**
  (#7024).

This is the same failure mode a third time. The practical consequence: any past
or future claim of the form *"the suite still passes with
`PERRY_GEN_GC_EVACUATE=0`"* carries no information about the GC ratchet's twelve
[...]

Issue #7610, opened 2026-08-07
gc: 12_large_live_set is 13.6% slower at 0.5.1346 than 0.5.1280 with every collector counter flat


proggeramlug:
Surfaced by the #7554 ratchet repair (#7609), which re-pinned the baseline on the
pinned quiet host. This is the one cell that moved and is **not** explained by any
merged, documented decision, so it is being filed rather than absorbed into the pin.

## The measurement

Same host (`perry-macos`, Mac mini M1, 8 GB), same toolchain (rustc/cargo 1.97.1,
Apple clang 21.0.0), same harness protocol (`--repeats 7`, warmup 1,
`PERRY_NO_AUTO_OPTIMIZE=1`, prebuilt static archives), median of 7:

| | `12_large_live_set.wall_ms` |
|---|---|
| pinned baseline `5e236e6e2` (0.5.1280, 2026-08-05) | **3,056 ms** — samples 3056.2, 3061.0, 3058.9, 3047.6, 3058.2, 3054.1, 3048.9 |
| `main` `26b9c9d59` (0.5.1346, 2026-08-07) | **3,471 ms** — samples 3471.7, 3472.7, 3466.7, 3466.3, 3471.3, 3475.5, 3467.5 |

**+13.58%.** The two 7-sample clusters do not overlap (3,047–3,061 vs
3,466–3,476) and within-session spread is 0.27%, so this is not scheduler noise.
The `pinned_host` band on `wall_ms` is 10%, derived from a worst cross-session
median spread of 0.751%.

## Why it is worth a look

Over the same window other probes got **faster**, so this is not a global
slowdown:

| probe | wall_ms 0.5.1280 → 0.5.1346 |
|---|---|
| `11_collect_at_depth` | 110.4 → 79.0 (**−28.4%**) |
| `06_string_retention` | 83.4 → 75.4 (**−9.6%**) |
| `12_large_live_set` | 3,056 → 3,471 (**+13.6%**) |

And #7596 ("live-proportional collection budgets at both generations") reported
**−7.4% wall_ms on this exact cell** in its own both-arms A/B. So by that PR's own
evidence the regression is not #7596 — something else in 0.5.1280..0.5.1346 made
the largest-live-set probe 13.6% slower.

Retention and the collector counters on this probe are essentially unchanged over
the same window (`heap_total_bytes` +0.00%, `promoted_objects` +0.00%,
`freed_bytes` +0.00%, `copied_objects` −0.01%), which is what makes the wall-time
[...]

PR #7613, opened 2026-08-07
perf(gc): seed promote-on-first-copy from a completed mark-sweep (#7598)

### LINKED ISSUE #7598: perf(gc): allocation-site pretenuring — long-lived cohorts are copied twice (Eden→survivor→old)
Follow-up to #7592 (fixed by #7594 + #7596 up to a constant factor). The quadratic is gone, but the dominant remaining cost on promote-heavy workloads is structural: **every long-lived object is copied twice** — Eden→survivor by the first copying minor, survivor→old by the next.

## Measured shape (json_pipeline, 500k records, after #7596)

```
 #  kind          trigger      old_before  eden_live  promoted   pause
 3  minor     arena_bytes         4.2 MB     268 MB        0    1,658 ms   <- 268 MB Eden->survivor, ALL of it long-lived
 4  minor     arena_bytes         4.2 MB       1 MB     269 MB  2,223 ms   <- the same bytes, survivor->old
```

3.9 s of the remaining 5.1 s `build_out` is these two copies of the same 268 MB. A collection-free run of the same loop is 764 ns/record vs the current ~23,000, so this is the bulk of what is left on #7592's workload.

## Why the cheap fixes do not work (all measured or ruled out this campaign)

- **The survival-rate lock is a cycle too late by construction** — it keys on the *previous* cycle's survivor round-trip, so the first big minor always pays the wasted copy. It did engage here (cycle 4 promoted), which is exactly why the waste is confined to cycle 3's copy.
- **An Eden-side lock (≥90 % of Eden survives ⇒ S=1) measured 1.0×** — it also only takes effect on the *next* copying minor, and a one-burst workload has no next minor that matters.
- **A mid-cycle valve (promote directly once the to-survivor space overflows) is ruled out by #7432**: which objects promote would then depend on root traversal order, breaking the bit-identical copied/promoted counters the gc-ratchet's determinism contract requires. The per-object decision must stay a pure function of (flags, age, cycle-start threshold).
- **Eden in-use at cycle start is not a survival signal** — churn workloads also overshoot Eden, with ~0 % survival; gating S on it would mis-promote garbage.

## The structural fix: allocation-site pretenuring

HotSpot-style: per-allocation-site survival feedback (site → promoted-fraction), fed back either as a runtime site tag on the allocation header or as a codegen hint (`arena_alloc_gc_old` directly for hot sites whose cohort demonstrably tenures). Deterministic per-cycle (the decision is made at allocation, not during the collection), so it composes with #7432.

Perry has unusually good static material for this compared to a JIT: `out.push({...})`-into-an-accumulator is visible in HIR (the #7469 loop-site work already classifies allocation sites by loop context, and `collectors/all_pointer_arrays.rs` proves which locals only receive freshly-allocated pushes). A static "allocated in a loop AND stored into an accumulator that outlives the loop" proof may cover the common case without any runtime feedback.

## Acceptance

- json_pipeline 500k `build_out` ≤ ~2.5 s (one copy of the live cohort instead of two) with output hash unchanged.
- gc-ratchet: all semantic counters bit-identical on the 12 probes except deliberate, documented promotion-count changes on the promote-heavy probes.
- No mid-cycle threshold mutation (#7432).

proggeramlug:
Closes #7598. The promote-on-first-copy remainder of #7592.

## The waste

On promote-heavy workloads every long-lived object is copied **twice** — Eden →
survivor by one copying minor, survivor → old by the next. `json_pipeline` at
500k, on `main`:

| # | kind | eden_live | copied | promoted | pause |
|--:|---|--:|--:|--:|--:|
| 3 | minor | 280,997,080 | **280,997,080** | 0 | 1,513 ms |
| 4 | minor | 1,048,576 | 0 | **282,045,656** | 2,559 ms |

The same 268 MB, moved twice. `gc/tenuring.rs` already computes the right
condition — the survival-rate lock — but it keys on `prev_copied`, so it needs a
*previous* copying minor to have filled the survivor space. It engaged for cycle
4, which is exactly why the waste is confined to cycle 3.

## Route chosen: P1 from the design note — seed the decision from a completed
non-copying collection

The other two were considered and rejected on this workload:

* **P3 (size-based pretenuring)** is a poor fit. These records are ~100–200 byte
  objects, so any `PretenureSizeThreshold` low enough to catch them catches
  almost everything, and its stated failure mode — a parse-and-discard loop over
  large buffers promoting pure garbage — is real and not bounded by anything the
  collector measures. It is a good idea whose threshold must come from a suite
  sweep, not from this workload.
* **P2 (per-age-band entry decisions)** is a refinement of P1, not an
  alternative: it buys precision, and the problem here is latency.

The obstacle is that `retune_after_scavenge` is fed **only** by copying minors,
so full mark-sweeps and non-copying minor fallbacks feed the loop nothing — and
once a workload escalates to those, the loop goes blind exactly when it is
needed. But every collection on that path already walks every Eden header and
classifies it live or dead. That census answers the same question one collection
earlier.

So: at the end of any such sweep, when

1. the surviving Eden cohort alone exceeds the desired survivor occupancy
   (`compute_target_survivals(...) == 1` — the module's **existing** rule, only
   read from a different place), **and**
2. ≥90% of the Eden bytes the sweep classified were live (the "an aging round
   would filter nothing" proof, measured instead of inferred),

engage the existing `PROMOTE_LOCK` so the next copying minor *enters* at S=1.

Exit is deliberately not this signal — it stays the existing influx-based
unlock, which is already threshold-invariant and already tested, so the seed
cannot introduce an enter/exit oscillation of its own.

## Why this signal is not a fixed point of the policy reading it

This issue has produced three self-referential signals already, so here is the
argument rather than the assumption.

The failures were all the same shape: **a signal suppressed by the state it is
supposed to leave.** `promoted_bytes` is zero by construction at S=4, so "was
the promotion rate high?" can never answer yes while S=4 holds. #7596's first
nursery cap gated from-space occupancy on a total that *included* from-space, so
it was a bound from-space could never cross. #7594's handoff scheduled a
*non-moving* full to relieve pressure only a *moving* cycle can relieve.

The Eden live/dead split has none of that structure, for a structural reason:
**it is produced by the mark-sweep's own arena walk.** Marks come from
reachability from roots; neither the mark phase nor the sweep walk reads
`tenuring_survivals()`. The threshold is consulted in exactly one place —
`copying.rs`'s per-object move — and this path does not run it. So the signal is
not merely *observed* to survive at S=4; it cannot be a function of S at all.

Measurably, at both endpoints:

| state | what Eden holds at the sweep | what the signal reads |
|---|---|---|
| S=4 — the state to leave | everything that has not aged out | high live fraction on a retaining workload ⇒ "lock" |
| S=1 — the state to stay in | only Eden allocated since the last promotion | the mutator's retention of recent Eden ⇒ still measurable |

Measured on this workload: `eden_live_bytes=279,964,968 eden_dead_bytes=896
live_pct=99 desired=1,048,576 seeds=true` — taken while S was still 4.

## Determinism (#7432)

#7432 forbids re-deciding S *while objects are being moved*, because the
copied/promoted split would then depend on root traversal order. The seed is
written at the END of a completed sweep and read at the ENTRY of a later copying
minor (`CopyingNurseryCollector::new` snapshots it once), so every object in a
cycle still sees exactly one threshold.

Two exclusions at the callsite keep the *input* deterministic too, and both are
refusals rather than tuning:

* **Budgeted cycles** — whole-cycle allocate-black marks every mid-cycle birth,
  and this walk reads MARKED as live, so a churn workload's births would read as
  a ~100% live Eden. Same reason the age-bump is suppressed there.
* **Cycles that ran the conservative native-stack scan** — that scan retains
  whatever the stack happens to look like a pointer to, by an amount that varies
  run to run (`benchmarks/gc_ratchet/README.md` measures 8.28 MB / 16% of one
  probe's reported retention). A liveness *measurement* taken under it is not
  sound, and feeding it to a policy would make the gated copy/promote counters
  non-deterministic.

No new env knob. The `PERRY_GC_DIAG` line prints the census **and the verdict on
every sweep, including refusals** — a policy that silently declines is
indistinguishable from one that never ran.

### Copy census — the halving signature, per cycle

Semantic counters, `PERRY_GC_TRACE=1`, output hash identical on every row.
500k:

| # | kind | old_before | eden_live (main → this) | copied (main → this) | promoted (main → this) | S |
|--:|---|--:|--:|--:|--:|--:|
| 1 | full | 112,695,904 | 0 → 0 | 0 → 0 | 0 → 0 | — |
| 2 | full | 121,002,720 | 0 → 0 | 0 → 0 | 0 → 0 | — |
| 3 | minor | 4,440,144 | 280,997,080 → **280,997,080** | 280,997,080 → **0** | 0 → **280,997,080** | 4 → 1 |
| 4 | minor | 4,440,144 | 1,048,576 → *(none)* | 0 → — | 282,045,656 → — | 1 → — |

**This is promotion, not cadence.** Cycles 1–3 are unchanged in kind, trigger,
`old_before` and `eden_live` — cycle 3 receives the *same 280,997,080 bytes to
the byte* and merely sends them somewhere else. The cycle that disappears is
baseline cycle 4, whose entire content was the second copy: its own Eden influx
was 1.0 MB and it promoted the 268 MB cycle 3 had just parked in the survivor
space. Removing it is removing the waste, not lengthening the collection
interval.

Totals:

| records | arm | collections | copied_bytes | copied_objects | promoted_bytes | promoted_objects | moved total | pause total |
|--:|---|--:|--:|--:|--:|--:|--:|--:|
| 200k | main | 4 | 113,227,216 | 1,657,966 | 114,275,776 | 1,670,376 | 227,502,992 | 3,008 ms |
| 200k | this | 3 | 0 | 0 | 113,227,216 | 1,657,966 | **113,227,216** | 1,040 ms |
| 500k | main | 4 | 280,997,080 | 4,117,015 | 282,045,656 | 4,129,425 | 563,042,736 | 5,083 ms |
| 500k | this | 3 | 0 | 0 | 280,997,080 | 4,117,015 | **280,997,080** | 2,838 ms |

Bytes moved: **0.498× at 200k, 0.499× at 500k.** Halved, as the design note's
signature requires.

**On the anti-vacuity gate.** The design note asks for `copied_objects > 0 &&
promoted_bytes > 0` on any cited run. The `main` arm satisfies both literally.
The changed arm cannot satisfy the first clause *because eliminating that copy is
the change*; what the clause exists to catch — the #7024/#7025 shape where the
fast arm ran zero collections — is refuted directly: it ran **3 collections**,
one of them a copying minor that moved **4,117,015 objects / 280,997,080 bytes**
(`[gc-copy-minor] ran … eligible=true fallback=none`). Reported as
`moved_objects > 0` rather than `copied_objects > 0`.

### Wall and RSS — pinned quiet host

`perry-macos` (M1, 8 GB), load ~1.6, 5 interleaved reps per size, warmup
discarded, `PERRY_NO_AUTO_OPTIMIZE=1`, prebuilt static archives, output
byte-identical (`cmp`) on every row.

| records | metric | main | this | Δ |
|--:|---|--:|--:|--:|
| 200k | wall | 1.85 s | 1.44 s | **−22.2%** |
| 200k | peak RSS | 608,747,520 | 485,736,448 | **−20.2%** |
| 500k | wall | 5.12 s | 3.86 s | **−24.6%** |
| 500k | peak RSS | 1,404,469,248 | 1,109,803,008 | **−21.0%** |

Spread within each arm was ≤0.03 s on wall and ≤0.25 MB on RSS across the 5
reps.

**RSS goes DOWN, which is the opposite of what the design note expected.** The
note assumed promoting earlier raises the old-gen high-water mark. It does — but
it removes a larger term: with S=4 the 268 MB cohort exists *twice at once* at
the peak, as Eden from-space plus survivor to-space. Promoting on first copy
means the peak holds one copy. This is the first change in this campaign whose
wall-time win does not have to be traded against RSS.

### gc-ratchet — the official `check`, on the #7609 baseline

This is the first GC-pacing change since 2026-08-01 that could be gated properly,
so it was run as a **two-arm** check on the pinned host: once with a `main`
(`d4342fff3`) build, once with this branch, back to back in the same session, both
built with the same `-p perry -p perry-runtime-static -p perry-stdlib-static` set.
The `main` arm establishes what the two commits of drift since the pin
(`26b9c9d59`) cost on their own, so anything the second arm moves is attributable
to this change.

```
gc-ratchet (check) at d4342fff3  (main arm)     → gc-ratchet: OK
gc-ratchet (check) at 3aa3b091c  (this branch)  → gc-ratchet: OK
```

**Diffing the two arms' 144-cell tables: every semantic cell is bit-identical.**
`heap_used_bytes`, `heap_total_bytes`, `minor_cycles`, `step_cycles`,
`copied_objects`, `copied_bytes`, `promoted_objects`, `promoted_bytes`,
`freed_bytes` — all twelve probes, no difference at all. The only rows that
differ are `rss_bytes`, `peak_rss_bytes` and `wall_ms`, which move in *both*
arms and stay inside band. The largest is `12_large_live_set.peak_rss_bytes`:
+0.36% on the main arm, +0.68% on this one (190,955,520 → 191,561,728, a 606 KB
difference on a 190 MB probe, band 3%). It is not a promotion effect — that
probe's `promoted_bytes` is identical to the byte in both arms.

**`12_large_live_set.wall_ms`** — #7610's unexplained +13.58% flag, now gated:

| arm | baseline | measured | Δ |
|---|--:|--:|--:|
| `main` `d4342fff3` | 3,471 | 3,016 | −13.11% |
| this branch | 3,471 | 3,015 | −13.14% |

**This change does not move that cell** (1 ms apart). Separately, and worth
recording on #7610: the +13.58% did not reproduce in this session at all — a
`main` build measures 3,016 ms against the 3,471 ms the artifact pinned three
commits earlier on the same host, i.e. the regression reads as *reversed*. That
is a data point for #7610, not a claim by this PR.

### The subject was live — and where

The ratchet result above proves **no collateral**; it does not prove the policy
ran, because on these probes it correctly does nothing. That distinction is the
#7024/#7025 shape, so it is settled by the diagnostic rather than assumed. The
seed prints its census *and its verdict on every mark-sweep, including refusals*:

```
12_large_live_set:  [gc-tenuring] sweep-seed eden_live_bytes=12582432
                    eden_dead_bytes=22019256 live_pct=36 desired=2097152
                    seeds=false already_locked=true
json_pipeline 500k: [gc-tenuring] sweep-seed eden_live_bytes=279964968
                    eden_dead_bytes=896 live_pct=99 desired=1048576
                    seeds=true already_locked=false
```

So on the largest-live-set probe the rule is **evaluated and declined** — by the
survival-rate half, at 36% — and on the target workload it is evaluated and
accepted. A policy that silently declines is indistinguishable from one that
never ran; this is why the refusal branch prints.

### Sabotage verification

Each mutation was applied to the shipped code and the named tests re-run. All
four turn red, and the tree is green again after restoring:

| mutation | result |
|---|---|
| drop `block_idx < resettable_general_n` in `keep_live_object` (old-gen bytes leak into the Eden census) | `full_sweep_eden_census_counts_only_nursery_blocks` **FAILED** |
| make `seed_promote_lock_from_sweep` a no-op | `sweep_seed_decides_before_the_first_copying_minor_snapshots_the_threshold`, `sweep_seed_hands_over_to_the_existing_unlock_path` **FAILED** |
| drop the survival-rate condition | `sweep_seed_refuses_a_churn_eden`, `sweep_seed_rule_is_a_pure_function_of_the_census` **FAILED** |
| drop the occupancy condition | `sweep_seed_refuses_a_small_fully_live_eden`, `sweep_seed_rule_is_a_pure_function_of_the_census` **FAILED** |

### Gates (local — CI backlog is deep, so this is the evidence)

`cargo test -p perry-runtime --no-fail-fast` **1885 passed, 0 failed** ·
`cargo fmt --all -- --check` clean · `check_file_size.sh` OK ·
`raw_handle_debt.py` **998 (baseline 998)** · `addr_class_inventory.py` passed ·
`class_id_collisions.py` passed. No codegen change (the diff is four files, all
under `crates/perry-runtime/src/gc/`, 314 insertions and no deletions), so the
root-dominance corpus is not implicated.

### What this does not do

* **No new env knob.** The `PERRY_GC_DIAG` line uses the existing one, so the
  kill-policy has nothing new to carry (and #7611 has nothing to add to).
* **It does not touch cadence.** The full's census is deliberately *not* fed to
  `retune_nursery_cap_scale`; collection pacing stays exactly where #7594 and
  #7596 left it.
* **It does not close allocation-site pretenuring.** #7598 also proposed static
  "allocated in a loop and stored into an accumulator that outlives it" proof.
  This removes the double copy for cohorts a mark-sweep can see; a workload whose
  first collection is already a copying minor still pays one wasted copy, and
  that is what the static route would cover.

### Collateral check on GC-shaped workloads outside the ratchet

Six micro-workloads the ratchet does not contain, both arms, semantic counters
and stdout compared. `tree` is the workload the survival-rate lock was written
for (medium-lived cohorts that *do* die in the survivor space — the failure mode
this rule must not trip on); `retain` is the accumulator shape; `churn` and
`push_cls` are pure churn.

| workload | stdout | collections | copied_bytes | promoted_bytes |
|---|---|---|---|---|
| `tree` | identical | 42 = 42 | 17,080,392 = 17,080,392 | 85,866,016 = 85,866,016 |
| `retain` | identical | 7 = 7 | 17,808,760 = 17,808,760 | 152,021,368 = 152,021,368 |
| `churn` | identical | 105 = 105 | 3,906,160 = 3,906,160 | 64 = 64 |
| `cycles` | identical | 16 = 16 | 622,528 = 622,528 | 210,496 = 210,496 |
| `push_cls` | identical | 105 = 105 | 3,906,160 = 3,906,160 | 64 = 64 |
| `deeplist` | identical | 4 = 4 | 17,079,696 → 17,079,168 (**−0.0031%**) | 53,579,048 = 53,579,048 |

`deeplist`'s 528-byte / 3-object delta is stable across 3 runs per arm, so it is
deterministic rather than noise — and it is **not** a policy effect. It lands
entirely in **cycle 1**, a copying minor entered at S=4, while this workload's
first mark-sweep is cycle 3 — so no seed had been evaluated yet and the tenuring
state at cycle 1 is identical in both arms. It is the collector's documented
address-sensitive component in `copied_objects` (`gc_ratchet/README.md`: "a
sub-0.1% host-dependent component") showing up between two differently-laid-out
binaries. For the record, the seed *is* reached later on this workload and
reports `live_pct=100 seeds=true already_locked=true` — the copying path had
already locked, so it changes nothing.

### Reproducibility of the cited numbers

The change arm was rebuilt from the committed tree (`-p perry
-p perry-runtime-static -p perry-stdlib-static`, pinned `PERRY_RUNTIME_DIR`,
`PERRY_NO_AUTO_OPTIMIZE=1`) and `json_pipeline` 500k re-run against it: 3
collections, `copied_bytes` 0, `promoted_bytes` 280,997,080, output
byte-identical to the `main` arm. The committed source reproduces the census
above.

## Summary by CodeRabbit

- **Performance**
  - Improved garbage collection by promoting highly live memory directly during the next collection cycle, reducing redundant copying and memory movement.
  - Added safeguards to preserve predictable behavior for budgeted collections and conservative stack scanning.

- **Bug Fixes**
  - Improved Eden memory accounting during collection, distinguishing retained and reclaimed memory more accurately.

- **Documentation**
  - Added release documentation describing the promotion improvements and measured performance benefits.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1348
+**Current Version:** 0.5.1349
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1348"
+version = "0.5.1349"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "perry-ffi",
  "scraper",
@@ -5813,7 +5813,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-commander"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5821,7 +5821,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cron"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "chrono",
  "cron",
@@ -5831,7 +5831,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dayjs"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "chrono",
  "perry-ffi",
@@ -5839,7 +5839,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-decimal"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "perry-ffi",
  "rust_decimal",
@@ -5847,7 +5847,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dotenv"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "perry-ffi",
  "serde_json",
@@ -5855,7 +5855,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ethers"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "perry-ffi",
  "rand 0.10.1",
@@ -5863,7 +5863,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-events"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5871,14 +5871,14 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-exponential-backoff"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "perry-ffi",
 ]
 
 [[package]]
 name = "perry-ext-fastify"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "bytes",
  "http-body-util",
@@ -5896,7 +5896,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-fetch"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "bytes",
  "lazy_static",
@@ -5909,7 +5909,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-http"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "bytes",
  "h2",
@@ -5933,7 +5933,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ioredis"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "lazy_static",
  "perry-ffi",
@@ -5943,7 +5943,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-jsonwebtoken"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "base64",
  "jsonwebtoken",
@@ -5954,7 +5954,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-lru-cache"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "lru",
  "perry-ffi",
@@ -5963,7 +5963,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-moment"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "chrono",
  "perry-ffi",
@@ -5971,7 +5971,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-mongodb"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "bson",
  "futures-util",
@@ -5983,7 +5983,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-mysql2"
-version = "0.5.1348"
+version = "0.5.1349"
 dependencies = [
  "chrono",
  "perry-ffi",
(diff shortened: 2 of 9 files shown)


proggeramlug:
## Audit before merge — verified to the byte, merged as v0.5.1349

**The copy census reproduces exactly on my own build and my own earlier
traces**: my v0.5.1339-era main run recorded 4 cycles / copied 113,227,216 /
promoted 114,275,776 at 200k — the agent's main-arm numbers to the digit. The
PR arm on my build: **3 cycles / copied 0 / promoted 113,227,216** — the
identical byte count, single-hop. Output hash unchanged; RSS 582 → 464 MB on my
measurements (−20%, matching the claimed −20.2%).

**The count-drop deviation from the anti-vacuity signature is correctly
argued, and I checked it rather than accepting it.** My brief demanded
"bytes-copied halves while the collection count stays constant"; the count
dropped 4→3. But the eliminated cycle is main's cycle 4, whose Eden influx was
~1 MB and whose entire content was the second copy of the cohort — cycles 1–3
are unchanged in kind and trigger, and cycle 3 receives the same bytes. This is
promotion doing cadence's cleanup for free, not a cadence tune wearing
promotion's clothes. The refutation of the literal `copied_objects > 0` clause
(the change's whole point is copied=0 on this workload) is handled with direct
evidence on the other probes: `12_large_live_set` shows `live_pct=36
seeds=false` — the policy evaluated and declined, so a green ratchet is not a
never-ran green.

**The RSS surprise is the best part**: −20/−21% where the design note expected
growth. The explanation is mechanical and satisfying — at S=4 the 268 MB cohort
existed twice at peak (Eden from-space + survivor to-space); promoting on first
copy removes the duplicate. Best performance AND best RSS on the same change,
which is the plan's goal stated as a diff.

**Sabotage re-verified**: seed made a no-op → exactly the 2 claimed tests red;
restored, 1,886/0 full suite. All four lint gates + fmt clean here.

**This is also the first GC-pacing change properly gated since 2026-08-01** —
official `check` on the #7609 baseline, both arms, semantic cells bit-identical,
run as designed. The self-reference proof (the census comes from reachability;
neither mark nor sweep reads the threshold; the threshold is consulted only on
the copying path, which the sweep does not run) is the standard #7594's and
#7596's fixed-point traps demanded.

Also noted for the record: the agent's process catch — a `git checkout --`
restore during sabotage silently discarded an uncommitted diagnostic hunk, and
it re-ran the 500k census against a rebuild of the committed tree before citing
numbers. That is the #7547 lesson (commit before destructive restore) applied
without being told.

**#7592 remains open** for the `JSON.parse` tail (~742 ms at 500k) — with
build_out now at single-hop promotion, parse is the next-largest phase.

Issue #7616, opened 2026-08-08
gc-root-dominance is blind to the unrooted-register shape: the verbatim pre-#7453 code reports 0 violations in all three modes


proggeramlug:
`gc_root_dominance_check.py` reports the **verbatim pre-#7453 code** as clean in all three of its modes. The gate cannot see the bug class it was extended for.

## Reproduction (measured, local)

Re-plant the exact code #7453 removed from `expr/url_main.rs`'s `Expr::UrlNew`:

```rust
let url_ptr  = ctx.block().call(I64, "js_url_coerce_string", &[(DOUBLE, &url_v)]);
let base_v   = lower_expr(ctx, base)?;                                  // user code
let base_ptr = ctx.block().call(I64, "js_url_coerce_string", &[(DOUBLE, &base_v)]);  // allocates
ctx.block().call(I64, "js_url_new_with_base", &[(I64, &url_ptr), (I64, &base_ptr)]);
```

Emitted IR (`test_compat_url_date_math`, `PERRY_RS4GC=0 PERRY_GC_MOVING_LOOP_POLLS=1 PERRY_INLINE_SHADOW_SLOT=0`):

```llvm
%r883 = call i64 @js_url_coerce_string(double %r882)
%r884 = load double, ptr @…_.str.17.handle
%r885 = call i64 @js_url_coerce_string(double %r884)      ; allocates
%r886 = call i64 @js_url_new_with_base(i64 %r883, i64 %r885)  ; %r883 is stale
```

Over the 18 `test-files/` sources that exercise the URL lowerings:

| checker mode | sabotaged arm | clean arm |
|---|--:|--:|
| `--moving-only` (dominance) | **0 violations** | 0 |
| `--unrooted-allocas --moving-only` | **0 violations** | 0 |
| `--stale-registers --moving-only` | **3** | 3 |
| `--stale-registers` (no `--moving-only`) | **14**, of which **11 `sink=js_url_new_with_base`** | 3 |

The two gated modes are structurally blind: dominance asks *"is the root store late"* and there is no root store; `--unrooted-allocas` asks *"does this alloca have a root store"* and the value is in an SSA register, not an alloca. `--stale-registers` is the mode whose help text does cover it — *"report every register holding a GC value that is USED below a collection point without being re-read from a root (#7154)"* — and it does see it, but only with `--moving-only` off.

## Cause

[...]

Issue #7615, opened 2026-08-08
Layer 1 campaign: migrate perry-codegen's lowerings onto the rooting-by-construction API (88 modules, 262 hazard sites)


proggeramlug:
**Layer 1 of the GC-correctness plan** (`docs/engine-plan.md`, `docs/src/internals/rfc-rooting-by-construction.md`) is the only track that was never started. This issue is its campaign map: the ordered inventory of `perry-codegen` modules to migrate onto the `crate::rooting` combinators, one slice per PR.

Routing: linked from #7294. Design: #7459 (the RFC's own constructor was `E0499`, corrected). First lowering: #7461. Template slice: the PR that opens this issue.

---

## What "migrated" means

Two properties, both checked by `cargo test -p perry-codegen --lib rooting::`:

1. **The module names no `expr::temp_root` symbol.** That module is the raw, order-sensitive API — push / get / set / truncate, plus guards the caller must remember to release — and every bug in the #7341 family was an ordering mistake against it. It is the escape hatch the RFC's step 3 says to `#[deny]` per module. Rust has no attribute that denies calling a `pub(crate)` function from one module, so the deny is a test over the module's own source, inlined by `include_str!`.
2. **No raw heap pointer exists as a value the lowering can hold across a collection point.** `rooting::call_rooted` returns a slot, never a register; `rooting::call_with_roots` re-reads each slot as part of emitting the consuming call; `rooting::with_operands_rooted` owns the release on every path out, including `?`.

A slice finishes by adding its modules to `MIGRATED_MODULES` in `crates/perry-codegen/src/rooting.rs`. Removing a line is a regression, not a cleanup.

## Acceptance per slice

- Emitted IR **byte-identical** (normalised for register/label renumbering) on a program set that exercises the module's lowerings, **or every difference itemised and justified** — a difference is normally a real rooting bug the migration found, which is the point.
- `gc-root-dominance` corpus green in both gated modes, including `--seeded-violations 40`.
- `cargo test -p perry-codegen --lib`, `-p perry-runtime --no-fail-fast`.
- The gap-test families that touch the module's lowerings, identical failure set on both arms.

## Honest statement of what this buys

Measured on the template slice by sabotaging the migrated module four ways:

| reintroduced shape | compiles? | caught by |
|---|---|---|
| #7192 in the **borrow** form (`RootingEmitter`, the RFC's design) | **NO — E0499** | rustc, via `compile_fail` doctest |
[...]

Issue #7294, opened 2026-08-03
Engine plan tracker — GC correctness and performance (routing only; docs/engine-plan.md is authoritative)


proggeramlug:
Hub for the engine effort — correctness and performance.

**The plan is [`docs/engine-plan.md`](https://github.com/PerryTS/perry/blob/main/docs/engine-plan.md).** This issue deliberately holds **no content of its own**: it links and routes, nothing more.

That constraint is the point. Every hand-maintained duplicate this project hit recently had silently rotted — `gc_repsel_matrix.sh`'s header asserting `copy-minor 12/22` when the true figure was `0/49` (#7255), a `known_failures.json` entry absent from the `gap_snapshot.json` the gate actually reads (#797), a README performance table with no generated markers. **A tracker that restates the plan becomes the next one.** If something here contradicts the doc, the doc is right and this is stale.

---

## Part 1 — GC correctness

Shape: *a GC-managed pointer exists somewhere the collector does not know about, across a point where the collector can run.* Three homes, three mechanisms. **Order is 0 → 2; layers 1 and 3 are independent and startable now.**

| Layer | Scope | Issues |
|---|---|---|
| **0** | in-process LLVM — the enabler | #7241 |
| 1 | codegen lowering discipline | **#7615** (campaign map), `rfc-rooting-by-construction.md` |
| 2 | emitted-code liveness (statepoints) | #7108, #7174 |
| 3 | runtime Rust handles | #7231, #7254 |

Open instances (layer 2, subsumable — **fix the instance, do not build new mechanism**): #7246, #7247, #7248, #7251
Blocking bug: #7280 · Collector mode / RSS: #7056, #7161

## Part 2 — Performance

| Item | Issue |
|---|---|
| Array index range proof — **measured 10.7× on matrix_multiply** | #7286 |
| `method_calls` is guard-bound, not call-bound | #7287 |
| Build non-determinism (78 ms vs 3450 ms by file location) | #7288 |
| Sound i64 specialization replacement | #7244 |
| Write-barrier Phase B | #7187 |

## Gates and instruments

#7282 (freshness fingerprint too broad) · #7272 (addr-class misses `perry-ext-*`) · #7255 (matrix arm liveness) · #7245 (`native_proof_regressions` unwatched) · #7261 (strip-dir leak) · #797 (skip-list provenance)

---

## Two rules carried from recent experience

1. **A green run that performed zero copying minors proves nothing** (#7255). Assert `copied_objects > 0`.
[...]

Issue #7114, opened 2026-07-31
String literal operand is not GC-rooted across an allocating call in the same expression (stale handle after evacuation)


proggeramlug:
## Summary

A **string-literal handle is loaded before an allocating call in the same expression and reused after it**. The handle global is a GC root that evacuation *rewrites*; the register holding the pre-call value is stale. Under enough allocation the concatenation silently produces a truncated or empty string.

This is exactly the hazard `crates/perry-codegen/src/expr/temp_root.rs` (#6951) was built to close — *"an already-evaluated operand waiting for its sibling ... the slot is a mutable root, so an evacuating cycle rewrites it and the register pushed beforehand is stale"* — with the **string-literal-handle load not covered** by it.

Silent wrong output, exit code 0. No crash, no diagnostic.

## Repro

```ts
class Rec {
  id: number; score: number;
  constructor(id: number, score: number) { this.id = id; this.score = score; }
}
function make(i: number): Rec { const r = new Rec(i, 0); r.score = r.id * 1.5; return r; }
function run(n: number): number {
  let acc = 0;
  for (let i = 0; i < n; i++) { const r = make(i); acc = acc + r.score; }
  return acc;
}

console.log("acc:" + run(10000000));   // literal evaluated BEFORE the allocating call
const v = run(10000000);
console.log("hoisted:" + v);           // same concat, call hoisted out
```

```
$ node --experimental-strip-types repro.ts     # v26.5.1, the .node-version pin
acc:74999992500000
hoisted:74999992500000

$ perry compile repro.ts -o repro && ./repro
                                                <- EMPTY LINE
hoisted:74999992500000
$ echo $?
0
```

Only the first statement is wrong. Hoisting the call out of the concatenation fixes it, which localises the bug to operand evaluation order rather than to `run`, to `js_string_concat_value`, or to the arithmetic (the number is correct in both).

The corruption is **allocation-count dependent and its manifestation varies**, which is the signature of reading through a stale/reused heap slot rather than of a logic error:

| iterations | observed |
|---|---|
| 1,000 | `acc:1249000` (correct) |
| 100,000 | `acc:12499900000` (correct) |
| 1,000,000 | `1249999000000` — prefix gone |
| 10,000,000 | `S124999990000000` / `r1249999000000` / empty — prefix replaced by garbage |

## Root cause, from the emitted IR

`--trace llvm`, `@main`, the two statements side by side:

```llvm
; BROKEN — literal handle read first, used after the allocating call
[...]

PR #7617, opened 2026-08-08
refactor(codegen): Layer 1 slice 1 — migrate expr/url_main.rs onto the rooting API, and the campaign map (#7615)

proggeramlug:
Layer 1 of the GC-correctness plan was the last unstarted track (`docs/engine-plan.md`). This is its **first slice plus the campaign map**: one module migrated end to end onto the rooting-by-construction API as the template every subsequent slice copies, and the ordered inventory of the remaining 87 — **#7615**, linked from #7294.

## The module, and why

`crates/perry-codegen/src/expr/url_main.rs` (669 lines, URL / URLSearchParams / URLPattern).

- **It was the tree's one half-migrated module.** #7461 migrated a single arm of a single variant; the rest still used the raw `expr::temp_root` API. "No half-migrated module" is the rule this campaign has to establish, so the module that already violated it is the right place to establish it.
- **The canonical raw-pointer lowering lives here.** `js_url_coerce_string` returns a bare `*mut StringHeader` — it is the callee #7453 was filed against, and the reason `url_coerce_string` has a hand-written alternative in the checker's `ALLOC_RE`.
- **Allocating calls crossed by state, and an emitter loop over operands** (`UrlSearchParams.has/delete/forEach` build their operand list conditionally) — the two shapes the API has to handle.
- **High blast radius**: `fetch`, `http`, `Request`, `import.meta.url` all route through these lowerings. The template should be proved under real hazard.

## What the migration found

`URL.canParse(input, base)` and `URL.parse(input, base)` **still carried #7453's window**. Identical three lines to the `new URL(input, base)` bug: a raw `*mut StringHeader` held in an SSA register across the lowering of `base` (arbitrary user code) and across a second `js_url_coerce_string` that allocates whenever `base` is not already a string. #7453 fixed one of the three forms and #7461 migrated it; nobody re-read the other two.

That repeats #7461's own finding, which is the argument for an API over a checklist.

## IR-identity evidence

Program set: the 18 `test-files/` sources that exercise this module's lowerings, compiled `PERRY_RS4GC=0 PERRY_GC_MOVING_LOOP_POLLS=1 PERRY_INLINE_SHADOW_SLOT=0 PERRY_NO_AUTO_OPTIMIZE=1 --trace llvm`, both arms.

- **16 of 18 `.ll` files byte-identical.**
- 2 files differ; **11 functions in total**, and a mechanical check confirms every one of them calls `js_url_can_parse_with_base` or `js_url_parse_with_base`, that the only added instructions are shadow-frame / root plumbing, and that nothing was removed. No other lowering changed.

The two fixed lowerings gain a two-slot shadow frame, the two root stores, the two re-reads and the two releases. That is the cost of the fix, on those two lowerings only.

**Coverage of the identity claim, checked rather than assumed.** The 18 sources reach `js_url_new`/`new_with_base`, `can_parse`/`can_parse_with_base`, `parse`/`parse_with_base`, `search_params_{get,has,has2,set,append,delete,delete2,get_all,sort,to_string,entries_arr,new_any,new_empty}` and five setters — but *not* `URLPattern`, `searchParams.forEach`, the typed `keys`/`values` family, or the other four setters. A purpose-built probe covers exactly those, and its IR is **byte-identical** between arms. So the identity result is not an artefact of what the corpus happens to compile.

## Sabotage: does it fail to COMPILE?

**No — and that is the finding worth more than the slice.** Four historic shapes reintroduced into the migrated module, each result measured:

| reintroduced shape | compiles? | caught by |
|---|---|---|
| #7192 in the **borrow** form (`RootingEmitter`) | **no — `E0499`** | rustc, as a `compile_fail,E0499` doctest added here |
| hold the `call_with_roots` result across a later lowering | yes | nothing |
| the **verbatim pre-#7453 code**, via a bare `ctx.block().call` | yes | **nothing** — see below |
| reach back into `expr::temp_root` | yes | the ledger test |
| hold the operand guard so it can be released on one arm (#7462) | yes | the ledger test |

`FnCtx` has no interior mutability, so the borrow-carrying `Raw<'e>` cannot be built on it — #7459 and #7461 established that, and this slice confirms it rather than working around it. What the combinator form actually buys is stated in the code and in the RFC, in those words:

- `call_rooted` **produces no unrooted register**, so #7453's window has no spelling.
- `RootedSlot` has **no `read`** — #7461's `read(&self, ctx) -> String` is deleted, because a register loaded from a root is stale the moment anything else collects (#7114, #7375) and a `String` remembers nothing. `call_with_roots` fuses the re-read into the consuming call.
- `with_operands_rooted` **owns the release on every path including `?`**, so #7462's release-on-one-arm is not a program.
- The escape hatch is **denied per module** by `rooting::migration_ledger`, which `include_str!`s each finished module — the checkable form of the RFC's step 3, which Rust has no attribute for.

An unused `root_i64(ctx, reg)` combinator was written and then deleted, with the reason recorded in place: it is the one addition that would reopen the window, and it should arrive with a caller, not ahead of one.

## Second finding: the CI gate is blind to this bug class — #7616

Reintroducing the verbatim pre-#7453 code produces IR that `gc_root_dominance_check.py` reports as **clean in all three modes** — dominance 0, unrooted-allocas 0, stale-registers identical to the control. Dropping `--moving-only` surfaces **11 stale uses at `js_url_new_with_base`** in the sabotaged arm and **0** in the migrated one, so the shape is expressible; the `--moving-only` filter discards the window because `js_url_coerce_string` is in `ALLOC_RE` but not in `POLL_CAPABLE_RUNTIME`. #7453's own fix added it to one list and stopped.

The one-line fix is measured on #7616 (curated corpus unchanged at 23/39; the dependency-scale arm is not measurable locally) and deliberately **not** included here: widening a gate is its own change with its own corpora to measure.

- `gc-root-dominance` curated corpus, **both gated modes**, on the post-change compiler: 129/129 sources, 149 modules, 2452 functions, 9803 root stores → **0 violations**, **0 unrooted-alloca violations**, and `--seeded-violations 40` → **40 planted, 40 caught, 0 missed**. Baseline arm identical.
- All four checker static audits: `--self-test`, `--audit-alloc-re`, `--audit-poll-capable`, `--audit-immovable-sources`.
- `cargo test -p perry-codegen --lib` and `--doc` (both `compile_fail` doctests reject; both ledger sabotage self-tests fire).
- `cargo test -p perry-runtime --no-fail-fast`.
- `./run_parity_tests.sh --filter url` against the pinned oracle (node 26.5.1): **14/14 PASS, 0 parity fail, 0 compile fail, 0 crashed, 0 skipped — 100%**. Identical failure set on both arms, because the set is empty.
- Lint gates: `cargo fmt --all -- --check`, `check_file_size.sh`, `addr_class_inventory.py`, `class_id_collisions.py`, `raw_handle_debt.py` (+ self-tests), `gc_store_site_inventory.py`, `workspace_architecture.py`.

## Not in this PR

No version bump. No behaviour change beyond the two itemised bug fixes. No gate widening (#7616). No further modules — slice 1 is `lower_array_method.rs`, 40 hazard sites, per #7615.

Closes nothing; advances #7615.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved URL, URLPattern, and URLSearchParams stability during memory cleanup.
  - Fixed URLSearchParams deletion cleanup across all execution paths.
  - Improved reliability when parsing URLs with base values.
  - Prevented resource leaks during URL-related operations.

- **Documentation**
  - Updated engine documentation to reflect ongoing memory-safety improvements, safeguards, known limitations, and remaining work.


proggeramlug:
## Audit before merge — verified, merged as v0.5.1352

**The most valuable sentence in this PR is the honest sabotage table.** Asked
"does the discipline make historic bug shapes fail to compile?", the answer
recorded in `rooting.rs` is **mostly no** — only the borrow form dies at E0499;
`FnCtx` has no interior mutability, so the RFC's `Raw<'e>` is unbuildable on
it, confirming #7459/#7461 rather than papering over them. What the API buys is
then stated as measured fact, not aspiration: `call_rooted` never yields an
unrooted register, `RootedSlot::read` is *deleted* (killing the #7114/#7375
half), release is owned on every path including `?`, and the per-module ledger
denies the escape hatch. A template slice that overclaimed compile-time safety
would have set 87 modules up to trust a guarantee that does not exist.

**Sabotage re-verified here**: injecting a real `temp_root_push_i64` call into
the migrated module fails `migrated_modules_do_not_reach_past_the_rooting_api`
precisely. (My first attempt used a nonexistent API name and died at E0425 —
worth noting because it shows the ledger is scanning for the *real* raw API,
not a substring: the module's own doc comment mentions `expr::temp_root`
verbatim and does not trip it.)

**The migration found two live bugs** — `URL.canParse(input, base)` and
`URL.parse(input, base)` carrying #7453's exact three-line window; the
constructor was fixed, the static forms were never re-read. That is the
campaign's thesis demonstrated on slice one: mechanical migration surfaces the
windows greps missed. And the checker classification gap it exposed (#7616 —
the pre-#7453 code reads *clean* in all gated modes because
`js_url_coerce_string` is in `ALLOC_RE` but not `POLL_CAPABLE_RUNTIME`) is
correctly filed rather than fixed inline, since widening a gate is its own
change with its own corpus question.

**Verification re-run here**: codegen 688/0 + doctests (both `compile_fail`
reject), runtime 1,886/0, URL gap family 4/4 against node 26.5.1 (the one raw
diff is node's own DEP0169 stderr warning, identical on main, stripped by the
parity harness — the agent's 14/14 stands), all four lint gates + fmt clean.

**The campaign now has its map**: #7615 — 88 modules, 694 raw-pointer sites,
262 hazard sites, ten slices ordered by hazard density, terminal condition
`expr/temp_root.rs` going `pub(in crate::rooting)`. Slice 1 is
`lower_array_method.rs` (40 hazard sites).

Issue #7280, opened 2026-08-02
GC: #7154's residual is NOT fixed — the loop-polls config is red 0/30, and stock zod alone fails 5/40


proggeramlug:
#7154 was closed on a `sfw-registry --help` **30/30** result. That result could not be reproduced here, and the configuration a revert would ship is **red**.

## The gate result

perry `4e9b1b7ea` (`origin/main`, includes #7275):

| arm | result |
|---|---|
| **default** (no compile env, no run env) | **clean 30/30** |
| **`PERRY_GC_MOVING_LOOP_POLLS=1` compiled AND run — the config a revert ships** | **red 0/30**, `TypeError: object is not a function` |
| POLLS=1 compiled, `PERRY_GEN_GC=0` at run | clean 10/10 |
| POLLS=1 compiled, polls **not** set at run | clean 10/10 |
| default binary, POLLS=1 at run **only** — the false-green trap | clean 10/10 |
| `evac_minor` (default binary, `HEAP_LIMIT=8 INCREMENTAL=0 CONSERVATIVE_STACK_SCAN=off`) | **red 0/10**, live (`copied_objects=6084`) |

Liveness asserted both ways: the revert arm ran **6 copying minors** (`copied_objects=50695…`); the default arm ran **0**. The result tracks collector mode exactly.

**Note the fifth row.** Setting `POLLS=1` at *run* time against a *default-compiled* binary is clean 10/10 — because the poll codegen isn't in the binary. Anyone verifying that way gets a false green.

## `sfw-registry` is not in this repo

Two independent sweeps found no source tree and no binary — it is a private checkout. So the 30/30 is neither confirmed nor refuted *on that workload*. It is refuted on a faithful substitute (86 modules, 8 endpoints × 12 cross-module `defineApiCall(...)` at module init, shaped after the frames in #7154's own disassembly) **and on stock zod alone**.

**The library-only control is the important one** — 20 lines, no scaffolding: `import zod; build schema; safeParse` ×96. **35/40 clean under the revert config; 40/40 on the shipped default.** Failures print `TypeError: value is not a function` — #7154's exact string.

## Instruction-level identification

Under #7196's quarantine (`ZEAL=1 PROTECT_FROMSPACE=1 DEPTH=800`):

```
FAULT: signal 10 ... RETIRED FROM-SPACE  retired_by_minor=#18
last-known object: obj_type=2 size=72
2  perry_fn_node_modules_zod_src_v4_core_util_ts__clone + 1964
[...]

Issue #7213, opened 2026-08-01
gc-ratchet is unconditionally red on main: the pinned baseline predates #7161's evacuating-minor flip


proggeramlug:
## Summary

`gc-ratchet`'s pinned baseline (`benchmarks/gc_ratchet/baseline/gc-ratchet-v1.json`, commit `88dcee83b`, captured 2026-07-30) was taken while the **evacuating minor was the default**. #7161 then flipped it off pending #7154. The baseline therefore describes a collector the shipped configuration no longer runs, and the gate is **red on `main`** for every commit since.

Measured locally on the pinned host class (`darwin-arm64`, Apple M1, 8 core, idle, `--repeats 7`), during PR #7207:

| arm | `check --profile shared_ci` | gated REGRESSION rows |
|---|---|---|
| `origin/main` `3a983ca6a` | **exit 1** | 69 |
| #7207 | exit 1 | 69 (**identical row set**) |

Every one of the 69 is `minor_cycles` / `step_cycles` / `copied_objects` / `copied_bytes` / `promoted_objects` / `promoted_bytes` / `freed_bytes` reading `… → 0`, plus the `heap_used_bytes` / `heap_total_bytes` growth that follows from nothing being collected. `--profile pinned_host` adds 17 `rss_bytes`/`peak_rss_bytes` rows for the same reason. `wall_ms` is inside band on every probe.

## Why this compounds #7205

#7205 established that `gc-ratchet`'s `main` runs are cancelled while queued, so the gate has executed **zero times in three merges**. Even once that is fixed, it will come back red on the first run — and a gate that is unconditionally red is exactly as informative as one that never runs. Both halves have to land for it to gate anything.

## The two options, and why one of them is wrong

- **Re-pin the baseline now** on the shipped (non-moving) configuration. This makes the gate green — and makes `copied_objects`, `promoted_objects` and `freed_bytes` **permanently zero-baselined**, i.e. dark for exactly the collector behaviour they exist to gate. Any future evacuation regression would then have to move a counter that is pinned at 0 in both directions. This is the shape CLAUDE.md calls "the gate runs but its subject never did".
- **Carry it as a known-red until #7161 is reverted**, and re-pin as part of that revert, on the configuration the revert restores. The evacuation counters then have a live, non-zero baseline again.

[...]

Issue #7365, opened 2026-08-04
cargo-test: perry-runtime's suite fails a different number of tests on every run


proggeramlug:
## Summary

`cargo test -p perry-runtime --lib` fails a **different number of tests on every run**, on clean `main`, with no changes applied. It is a required check (`cargo-test`), so this is not cosmetic: a suite whose failure count is nondeterministic can hide a real regression indefinitely — a genuine new failure looks exactly like the noise.

## Measurement

Ten consecutive runs on `8f05d65f9`, unmodified:

```
3 failed   1 failed   1 failed   3 failed
4 failed   2 failed   2 failed   (earlier baseline, same tree)
```

Seven distinct tests are involved, by frequency across six sampled runs:

| count | test |
|---:|---|
| 5 | `gc::tests::teardown::map_set_owner_records_follow_growth` |
| 3 | `gc::tests::teardown::map_set_side_allocations_release_exactly_once` |
| 2 | `promise::keyed_table::tests::settling_many_keys_is_not_quadratic` |
| 1 | `url::node_compat::tests::path_to_file_url_posix_preserves_relative_trailing_slash` |
| 1 | `url::node_compat::tests::path_to_file_url_posix_does_not_add_slash_without_input_slash` |
| 1 | `object::prop_plan::tests::read_plan_roundtrip_and_epoch_flush` |
| 1 | `gc::tests::teardown::map_set_side_allocations_release_on_thread_exit` |

**Every one of them passes in isolation.** Verified individually:

```
gc::tests::teardown::map_set_owner_records_follow_growth              ok
gc::tests::teardown::map_set_side_allocations_release_exactly_once    ok
gc::tests::teardown::map_set_side_allocations_release_on_thread_exit  ok
object::prop_plan::tests::read_plan_roundtrip_and_epoch_flush         ok
```

So it is interference under parallel execution, not a defect in any individual test.

## Two distinct causes, probably

**1. Shared process-global state.** The `gc::tests::teardown::map_set_*` family is 9 of the 14 observed failures. Those tests assert on side-table *ownership records* and *release counts* — process-global structures that other tests mutate concurrently. Related prior art: `PERRY_GC_FORCE_EVACUATE`'s test-only mode override is deliberately thread-local *"so one test enabling the instrument cannot change the collector's behaviour for any other test"* (`arena/quarantine.rs`), and #7251 found a test made vacuous by libtest ordering, where `array::tests` built the intrinsic towers first. Same family of problem.

[...]

PR #7618, opened 2026-08-08
refactor(codegen): Layer 1 slice 1 — migrate lower_array_method.rs onto the rooting API (#7615)

proggeramlug:
**Layer 1 campaign slice 1** (#7615): `crates/perry-codegen/src/lower_array_method.rs` migrated end to end onto `crate::rooting`, following the template #7617 established. 1 module, 1199 → 1346 lines, 37 raw sites, **40 hazard sites — the map's highest density**. The whole module lands in one PR; the ledger notes no outstanding boundary.

## The hazard was structural, not incidental

`lower_array_method` lowered the **receiver above the `match`**:

```rust
let recv_box = lower_expr(ctx, object)?;      // NaN-boxed array pointer
match property { … }                          // every arm then lowers ITS OWN arguments
```

So every one of the ~30 arms held a heap pointer in an SSA register across the lowering of its arguments. For any argument that runs user code — a callback literal (`js_closure_new` allocates), a call, an array literal — that is #7453's window, in `arr.map(cb)`, `arr.filter(cb)`, `arr.sort(cmp)`, `arr.concat(f())`, `arr.splice(i, n, mk())`. The module had **zero** rooting references before this.

## What is actually new, stated against #7280

This is the part worth reading, because the module was **partially** protected already and the PR should not claim otherwise.

`root_reload.rs` (#7280) is a post-pass that re-reads a shadow slot below every collection point that can run under it. Where the receiver is a shadow-slotted local, it already fired — visible in the baseline IR as an out-of-sequence register (`%r309 = load double, ptr %r7` emitted after `%r250`). So for `const a = [...]; a.sort(cmp)` the old code was **not** stale.

What #7280 structurally cannot cover, and what this migration closes:

1. **A receiver reassigned by its own argument.** #7280 deliberately bails when a store to the slot can run in the window, because re-loading would observe the assignment — `operand_is_reloadable`'s documented miscompile. It leaves the register alone, and the register is stale. Baseline IR, verbatim:

   ```llvm
   %r243 = load double, ptr @perry_global_p_unrooted_recv_ts__1   ; the receiver
   %r244 = call double @perry_fn_p_unrooted_recv_ts__reassign()   ; allocates AND reassigns
   %r245 = bitcast double %r243 to i64                            ; stale
   %r246 = and i64 %r245, 281474976710655
   %r249 = call i64 @js_array_concat_variadic(i64 %r246, ptr %r247, i32 1)
   ```

   After:

   ```llvm
   %r249 = load double, ptr @perry_global_p_unrooted_recv_ts__1
   %r250 = bitcast double %r249 to i64
   store i64 %r250, ptr %r37                        ; root store, ABOVE the window
   call void @js_shadow_slot_bind(i32 3, ptr %r37)
   %r251 = call double @perry_fn_p_unrooted_recv_ts__reassign()
   %r252 = load i64, ptr %r37                       ; re-read, BELOW it
   %r254 = and i64 %r252, 281474976710655
   %r257 = call i64 @js_array_concat_variadic(i64 %r254, ptr %r255, i32 1)
   store i64 0, ptr %r37                            ; release
   ```

   A temp root is the only strategy that gives both the call-time value and a rewritten address. Same shape on `indexOf`.

2. **Argument-to-argument windows.** `arr.splice(mk(), mk(), mk(), mk())` held each evaluated argument in a register across the next one's lowering; no slot, so nothing to reload.

3. **A dead duplicate unbox in `sort`.** The comparator path unboxed the receiver above the `if`, lowered the comparator, then unboxed *again*. `%r243 = and i64 %r242, 281474976710655` in the baseline is referenced exactly once — its own definition. Removed; it is the only instruction this PR deletes.

## How it is migrated

One `rooting::with_operands_rooted` around the whole `match`, over the receiver plus the arguments the arm consumes. Those are declared in one place, `lowered_arg_count`, and the arms then only *emit* — **`lower_expr` does not appear in this file**, which is what makes "no operand register crosses a collection point" a property of the module rather than of each arm's author.

The asymmetry that makes the table safe is stated in its doc: **under-counting is loud** (an arm indexes a value that is not there and panics during codegen, on the first program that reaches it), **over-counting is benign** (JS evaluates every argument anyway, so lowering one the arm ignores is a spec fix at worst). Counts are exactly what each arm lowered before, so nothing changes in this slice. Two unit tests state the table independently and assert it never claims more than it was given.

## IR-identity evidence

Program set: 7 purpose-built probes reaching **all 46** of this module's distinctive callees — every arm including the `thisArg` array-like family, the runtime-dispatch family, the typed-array `set`/`subarray` arms, and the two receiver shapes #7280 cannot reach. Compiled `PERRY_RS4GC=0 PERRY_GC_MOVING_LOOP_POLLS=1 PERRY_INLINE_SHADOW_SLOT=0 PERRY_NO_AUTO_OPTIMIZE=1 --trace llvm`, `PERRY_RUNTIME_DIR` pinned, both arms.

- **95 functions compared; 90 identical, 5 differ — and all 5 are `main`**, the only function in each probe containing an array-method call. 2 of 7 modules are identical outright.
- Whole-corpus net instruction delta, every line of it:

  | | |
  |--:|---|
  | +55 / +53 / +53 / +53 | `store i64 0, ptr %` · `load i64` · `store i64` · `bitcast i64→double` |
  | +27 / +27 / +17 / +17 / +8 / +8 / +1 / +1 | `js_shadow_slot_bind` / `js_shadow_slot_set` at slots 1–4 |
  | +19 / +2 / +1 | `bitcast double→i64` · `alloca i64` · `js_shadow_frame_enter(i32 5)` |
  | −36 | `load double, ptr %` — #7280's reloads, replaced by root reads |
  | −1 / −1 | the dead `sort` unbox (`bitcast` + `and`) |
  | −1 | `js_shadow_frame_enter(i32 3)` (frame grew to 5 slots) |

  **Non-root-plumbing added: 0.** Non-plumbing removed: 1 kind, the dead `sort` unbox, verified dead by reference count.

The comparison masks register *names* and compares each function as a multiset, because `main` here is a long run of near-identical `console.log(a.method(…))` statements and `difflib` aligns statement N's inserted plumbing against statement N+1's body — it reported 1426 spurious additions for a 171-instruction delta. What the multiset gives up is ORDER, which is the property a rooting change is about, so order is checked separately: by reading the unified diff of the small cases (shown above) and by the dominance checker over the whole corpus.

**Behavioural A/B:** all 7 probes, both arms — identical stdout and exit code; and the new arm is byte-identical again under `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=64`.

## Ledger sabotage — result recorded

Reintroduced the escape hatch into the migrated module (a real, compiling `temp_root_push_double` + `temp_root_truncate` pair in the `reverse` arm):

| reintroduced shape | compiles? | caught by |
|---|---|---|
| reach back into `expr::temp_root` | **yes** | **the ledger test — red, naming `lower_array_method.rs:375` and `:376`** |

Same answer #7617 measured, now confirmed for a second module: the API does not make the bug fail to compile, and the ledger is what denies the escape hatch. Reverted; the file names `temp_root` only in one prose line of its module doc, which `escape_hatch_uses` strips as a comment.

- `gc-root-dominance`, **both gated modes**, on the post-change compiler: corpus 129/129 sources → 149 modules, 2452 functions, **9810 root stores → 0 violations**; `--seeded-violations 40` → **40 planted, 40 caught, 0 missed**; `--unrooted-allocas --moving-only` → **0**. Baseline arm: same corpus, **9803** root stores, 7860 gc-capable allocas, also 0/0. The +7 / +1 delta is how you can tell the gate's subject was live rather than absent.
- All four checker static audits: `--self-test`, `--audit-alloc-re`, `--audit-poll-capable`, `--audit-immovable-sources`.
- `cargo test -p perry-codegen --lib` — 691 pass, including both new `lowered_arg_count` tests and all four ledger tests. `--doc` — both `compile_fail,E0499` arms still reject.
- `cargo test -p perry-runtime --no-fail-fast` — 1886 pass, 0 fail.
- `./run_parity_tests.sh --filter test_gap_array` against the pinned oracle (node 26.5.1), **both arms**: 13/13 PASS, 0 parity fail, 0 compile fail, 0 crashed, 0 skipped — identical failure sets, because the set is empty. Full-suite result noted below.
- Lint gates: `cargo fmt --all -- --check`, `workspace_architecture.py` (+self-test), `check_file_size.sh` (1346 lines, under the cap), `gc_store_site_inventory.py` (+self-test), `addr_class_inventory.py` (+self-test), `class_id_collisions.py`, `raw_handle_debt.py` (+self-test, 998 = baseline), `gc_gate_wiring_check.py` (+self-test), `binding_pins.mjs --check`.

## Unrelated finding: `lint` is already red on `main`

`python3 scripts/check_test_registration.py` fails on a pristine `origin/main`:

```
DARK TEST test-files/test_gap_repsel_element_shape_loop_clone.ts
  exists on disk but is not registered in test-parity/gc_repsel_corpus.txt, so
  scripts/gc_repsel_matrix.sh (gc-stress, gc-moving-witnesses) never runs it.
```

That file landed in #7612 (`96e034809`). Neither it nor `test-parity/gc_repsel_corpus.txt` is in this PR's diff, and registering a test changes what `gc-stress` / `gc-moving-witnesses` execute — a gate change with its own measurement, not something to smuggle into a refactor. Flagging it rather than fixing it: a required-and-red gate means every merge is bypassing it, which is hazard 2 in CLAUDE.md.

## Not in this PR

No version bump. No behaviour change beyond the itemised rooting fixes and the one dead instruction. No new combinator — the `toString` arm's `unbox_str_handle` window (#7213) is deliberately left, with its reasoning recorded in the module header: closing it needs a combinator that roots across a collection point this module *emits* rather than one it infers from an operand list, and that should arrive with the slice that needs it. No gate widening. No further modules.

Advances #7615. Slice 1b (`expr/arrays_finds.rs`, `expr/array_methods.rs`) is next.

## Summary by CodeRabbit

* **Documentation**
  * Updated project documentation to record completed runtime migration work and verification results.
  * Clarified migration boundaries and current completion status.
  * Added a changelog entry covering rooting improvements and validation results.

* **Chores**
  * Incremented the project version from 0.5.1353 to 0.5.1354.

* **Bug Fixes**
  * Improved memory-safety handling during array method operations.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1353
+**Current Version:** 0.5.1354
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1353"
+version = "0.5.1354"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1353"
+version = "0.5.1354"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1353"
+version = "0.5.1354"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1353"
+version = "0.5.1354"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1353"
+version = "0.5.1354"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1353"
+version = "0.5.1354"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1353"
+version = "0.5.1354"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1353"
+version = "0.5.1354"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1353"
+version = "0.5.1354"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1353"
+version = "0.5.1354"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1353"
+version = "0.5.1354"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1353"
+version = "0.5.1354"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1353"
+version = "0.5.1354"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1353"
+version = "0.5.1354"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1353"
+version = "0.5.1354"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1353"
+version = "0.5.1354"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1353"
+version = "0.5.1354"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1353"
+version = "0.5.1354"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1353"
+version = "0.5.1354"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1353"
+version = "0.5.1354"
 dependencies = [
  "perry-ffi",
  "rusqlite",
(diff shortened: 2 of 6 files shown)


proggeramlug:
## Audit before merge — verified, merged as v0.5.1354

**Structural claims verified on the branch**: `lower_expr` appears once in the
file — in the doc comment explaining its absence; `temp_root` once, likewise;
the module is in the ledger, and my sabotage (a real `temp_root_push_i64` call
appended) turns `migrated_modules_do_not_reach_past_the_rooting_api` red.

**The genuinely-new coverage verified behaviourally**: the
receiver-reassigned-by-its-own-argument case — the one #7280's `root_reload`
deliberately bails on, since re-loading the slot would observe the assignment
and miscompile — matches node exactly on my probe:

```
a.concat(reassign())  →  [1,2,3,7], a = [9,9]     (concat sees the OLD receiver)
x.concat((x = …, …))  →  per-iteration correct
```

**The honest scope correction is the report's best content.** The map said "40
hazard sites"; the agent found #7280 already covered the shadow-slotted-receiver
majority (visible in baseline IR as the out-of-sequence re-read), and said so
rather than claiming 40 closed. What this PR actually buys — the reassignment
case, argument-to-argument windows, one dead unbox — is smaller and real. A
campaign whose slice reports inflate coverage would converge on false
confidence; this one set the opposite precedent on slice 1.

**Re-run here**: root-dominance both modes (129/129, 0 violations, 40/40
seeded, unrooted-allocas 0 — root stores up 7 vs baseline, the gate's subject
live), codegen 690/0, the full lint script set (all exit 0 — including
`check_test_registration.py`, now that #7619 repaired what #7612's merge
broke), fmt clean. One runtime-suite failure did not reproduce on two
consecutive full reruns — #7365's known flake, and this PR touches no runtime
code.

**Process note folded into this merge cycle**: the slice agent's flag of the
red `lint` on main (the dark #7612 test) was fixed first as #7619, and my merge
checklist is replaced by enumerating the lint job's script set from `test.yml`
— the second breakage via a memorized subset is the last.

PR #7620, opened 2026-08-08
refactor(codegen): migrate arrays_finds.rs + array_methods.rs onto the Layer 1 rooting API (#7615)

proggeramlug:
**Layer 1 campaign slice 1b** (#7615): `crates/perry-codegen/src/expr/arrays_finds.rs` and `crates/perry-codegen/src/expr/array_methods.rs` migrated end to end onto `crate::rooting`, following #7617's template and #7618's precedent. 2 modules, 1562 → 1763 lines; the map counts 40 raw sites and 26 hazard sites. Both land whole; the ledger notes no outstanding boundary.

Adds **one** combinator, `rooting::with_operands_rooted_across`, arriving with its two callers and with the argument for why the existing one cannot serve (below). `with_operands_rooted` is now its empty-`across` case, so "root, re-derive, or reuse?" is still answered in exactly one place — the drift that produced #7114 was two helper families answering it separately.

### 1. `Expr::BufferSlice` held a RAW, already-unboxed pointer across user code

```text
buf_box    = lower_expr(buffer)        // NaN-boxed BufferHeader
buf_handle = unbox_to_i64(buf_box)     // RAW pointer, in an SSA register
start_box  = lower_expr(start)         // arbitrary user code — allocates
end_box    = lower_expr(end)           // ditto
js_buffer_slice(buf_handle, …)         // reads the PRE-MOVE address
```

#7453's shape with a twist that matters: what is in flight is no longer NaN-boxed, so **#7280's `root_reload` structurally cannot repair it** — that post-pass re-reads a shadow slot into a `double`, and the consuming call reads an `i64` derived above the window. The unbox is now emitted below the group's re-read, which is the only place it can be correct.

### 2. The `arr.find*` family, in the shape it is almost always written in

`find` / `findIndex` / `findLast` / `findLastIndex` all lowered the array first and left it in a register while the callback was lowered. A callback literal lowers to `js_closure_new` — an allocation — so `xs.find(x => x.id === id)` is a live window. Same for `Object.is(a, b)`, `Object.hasOwn(o, k)`, `path.matchesGlob(p, pat)`, the `Map`/`Set` positional readers, `AggregateError`, `Buffer.concat(list, total)`, `Object.create(proto, props)`, both `FinalizationRegistry` mutators, both `ErrorNew*` forms, the three-operand `NativeArenaView`/`NativePodView`, the multi-argument `new Date(y, m, d, …)` (each component live across the next), and the polymorphic `u8[k] = v` store.

### 3. The index arms needed a re-read point this module controls

`u8[i]` and `buf[i]` lower the receiver with `lower_expr` and the index with `lower_index_i32`, which picks between an `i32` fast path (`lower_expr_as_i32`) and a `double` plus `fptosi`. The receiver is live across that choice. Handing the index to `with_operands_rooted` would force every such read back onto the NaN-boxed path — a codegen-quality regression, not a rooting fix. `with_operands_rooted`'s re-read point is fixed at the end of its operand list, so an operand lowered before caller-controlled work would be re-read *above* it and stale again by the time the call runs, which is the #7114 half-measure. `with_operands_rooted_across` roots the group before the caller-controlled lowering and re-reads it after; `across_exprs` is passed as expressions rather than a `bool` so the "does this window collect?" question stays inside `operand_protection`. Two callers, both here.

The new combinator deliberately does **not** reopen what `rooting.rs`'s existing note warns about: it never takes a bare register, it still owns the release (now including a bail from the operand lowering itself), and it never hands out an unrooted result.

## Honest scoping — what is genuinely new vs what #7280 already covered

Following #7618's precedent, this is stated rather than glossed. For a **shadow-slotted local receiver**, `root_reload` had already repaired the `find` window, and the baseline IR shows it as an out-of-sequence re-read:

```llvm
%r595  = load double, ptr %r9                       ; the receiver
%r596  = call i64 @js_closure_alloc_singleton(…)    ; ALLOCATES
%r1681 = load double, ptr %r9                       ; ← #7280 already fired
%r600  = and i64 %r599, 281474976710655
%r602  = call double @js_array_find(i64 %r600, i64 %r601)
```

`-16 load double, ptr` in the net delta is exactly that population: #7280 reloads replaced by root reads. What this PR actually buys is the three shapes #7280 structurally cannot cover:

1. **A receiver reassigned by its own argument.** #7280 bails there by design (`operand_is_reloadable`'s documented miscompile: re-loading would observe the assignment). Baseline, verbatim — note the absence of any re-read:

   ```llvm
   %r1046 = load double, ptr @perry_global_p3_finds_ts__1   ; `live`
   %r1047 = call double @…__swap()                          ; allocates AND reassigns `live`
   %r1048 = bitcast double %r1046 to i64                    ; STALE
   %r1051 = call double @js_array_find(i64 %r1049, i64 %r1050)
   ```

   After:

   ```llvm
   %r1062 = load double, ptr @perry_global_p3_finds_ts__1
   %r1063 = bitcast double %r1062 to i64
   store i64 %r1063, ptr %r14                    ; root store, ABOVE the window
   call void @js_shadow_slot_bind(i32 5, ptr %r14)
   %r1064 = call double @…__swap()
   %r1065 = load i64, ptr %r14                   ; re-read, BELOW it
   %r1069 = call double @js_array_find(i64 %r1067, i64 %r1068)
   store i64 0, ptr %r14                         ; release
   ```

2. **A raw, already-unboxed pointer** — `BufferSlice` above. A `double` reload cannot repair an `i64` derived above the window.
3. **Operand-to-operand windows and receivers with no slot at all** — call results, property chains, the four-operand `finreg.register(t, h, tok)`.

## IR-identity evidence

Two corpora, both arms, compiled `PERRY_RS4GC=0 PERRY_GC_MOVING_LOOP_POLLS=1 PERRY_INLINE_SHADOW_SLOT=0 PERRY_NO_AUTO_OPTIMIZE=1 --trace llvm`, `PERRY_RUNTIME_DIR` pinned. Register names and block-label numbers are masked; a function is "identical" iff its instruction sequence matches up to that renaming.

**A. 8 purpose-built probes** reaching **28 of the 30 distinctive callees of the arms this PR changed** (the two gaps are named below).

- **172 functions compared; 164 identical, 8 differ — and all 8 are `main`**, the only function in each probe containing these lowerings.
- Whole-corpus net instruction delta, every line of it:

  | | |
  |--:|---|
  | +36 / +36 | `store i64 %r` · `js_shadow_slot_bind` — the root stores |
  | +36 / +36 | `load i64, ptr` · `bitcast i64→double` — the re-reads |
  | +37 / +36 | `store i64 0, ptr` · `js_shadow_slot_set` — the releases (the odd one is a grown frame's prologue zero-init) |
  | +12 | `bitcast double→i64` — the store side of a push |
  | +1 | `alloca i64` — one more pooled temp slot |
  | −16 | `load double, ptr` — #7280's reloads, replaced by root reads |

  **Non-root-plumbing added: 0. Non-root-plumbing removed: 0.** This PR deletes no instruction kind and adds no callee.

**B. The whole `gc-root-dominance` corpus** (129 sources → 149 modules) compared the same way, which also re-checks the two already-migrated modules against the `with_operands_rooted` refactor:

- **2452 functions; 2443 identical, 9 differ.** 6 of the 9 are root plumbing only (+16 stores/binds/reads/releases, +2 allocas, −9 stale `load double`).
- The other **3 have an identical instruction multiset and a different order, and are compiler NONDETERMINISM, not this change** — proven by compiling one source **twice with the same binary**: `@.str.N` constants and `js_register_function_name` calls come out in a different order run to run, and in `test_gap_class_expr_dynamic_parent_ctor` three `perry_method_*` callees permute. Filed as #7622; it is a pre-existing build-reproducibility gap, and worth knowing about before anyone else tries a byte-level A/B here.

**Behavioural A/B**: all 8 probes, both arms — identical stdout and exit code, and identical again under `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=64`.

**Coverage the probes do NOT reach**, stated rather than glossed:

- `js_native_pod_view` (`lower_native_pod_view_with_layout`) — `NativeArena.podView` needs a `PerryPodView<T>` annotation over a `PerryPod<…>` layout, which has no TS spelling reachable from a standalone probe; it is exercised only by `crates/perry-codegen/tests/native_proof_*` (nightly/tag only). Its migration is the same three-operand `with_operands_rooted` as `NativeArenaView`, which IS probed twice.
- `js_buffer_index_get_value` (`BufferIndexGet`'s tail) — the structural twin of `Uint8ArrayGet`'s tail, which is probed 6 times through the same new `with_operands_rooted_across` call. Five index shapes were tried and all routed to `lower_buffer_load` or the dynamic helper.

## Ledger sabotage — result recorded, per module

`escape_hatch_uses` strips comments, and both files named `temp_root` zero times before this PR, so listing them passes vacuously. The arm that makes the listing mean something was run for **each** module separately (the assert stops at the first offender, so one run cannot speak for both):

| module | injected | compiles? | caught by |
|---|---|---|---|
| `expr/arrays_finds.rs` | `temp_root_push_double` + `temp_root_truncate` | **yes** | **ledger test — red, naming `arrays_finds.rs:279` and `:280`** |
| `expr/array_methods.rs` | `temp_root_push_i64` + `temp_root_truncate` | **yes** | **ledger test — red, naming `array_methods.rs:59` and `:60`** |

Same answer #7617 and #7618 measured, now for a third and fourth module: the API does not make the bug fail to compile; the ledger is what denies the escape hatch. Both reverted. That vacuity, and the obligation to run this arm, is now written into `MIGRATED_MODULES`' doc comment so the next slice cannot skip it by accident.

## Verification (local — the CI backlog is deep, so this is the evidence, and it is stated as such)

- `gc-root-dominance`, **both gated modes**, post-change compiler: corpus 129/129 sources → 149 modules, 2452 functions, **9826 root stores → 0 violations**; `--seeded-violations 40` → **40 planted, 40 caught, 0 missed**; `--unrooted-allocas --moving-only` → **0**. Baseline arm on the same corpus: **9810** root stores, 7861 gc-capable allocas, also 0/0. The +16 / +2 delta is how you can tell the gate's subject was live rather than absent.
- All four checker static audits: `--self-test`, `--audit-alloc-re` (71 alternatives), `--audit-poll-capable` (54 entries), `--audit-immovable-sources` (4 probes, 0 failing).
- Stale-register budget, curated corpus, **both arms**: 23 ≤ 39, unchanged.
- `cargo test -p perry-codegen --lib` — **691 pass, 0 fail**, including all four ledger tests. `--doc` — both `compile_fail,E0499` arms still reject.
- `cargo test -p perry-runtime --no-fail-fast` — **1886 pass, 0 fail**, clean on the first run (no rerun needed; #7365's known flake did not appear).
- Gap suite, **both arms**, 17 family filters over the lowerings these modules own (`array`, `buffer`, `path`, `map`, `set`, `date`, `object`, `typed`, `process`, `iterator`, `weakref`, `error`, `fs`, `regexp`, `math`, `uint8`): **60 tests, 60 pass, 0 parity fail, 0 compile fail, 0 crashed, 0 skipped — identical failure sets, because the set is empty on both.** (`test_gap_finaliz*` matches no file; the harness exits 1 on an empty selection, which is selection, not a result.)
- **The full `lint` script set, enumerated from `.github/workflows/test.yml` rather than remembered** — 27 invocations. All pass **except one, which is red on `main` too** (see below).

## Pre-existing failures found, flagged rather than smuggled in

**1. `benchmarks/ci_public_baseline_check.py` is red on `main`.**

```
public baseline error: public artifact benchmark harness changed; regenerate it with ./benchmarks/run_public_baseline.sh
```

This PR's diff is three files under `crates/perry-codegen/src/`; it cannot cause this. #7618 recorded the same step failing on every `main` run from 2026-07-29 onward. Because `lint` is a *sequence* and a failing step skips the rest, this is CLAUDE.md hazard 4 in the costume the workflow's own comment names — except the later steps now carry `if: ${{ !cancelled() }}`, so they do still speak.

**2. `path.resolve(base, f(), …)` throws `ERR_INVALID_ARG_TYPE` where node returns the path — an SSO bug, not a rooting bug.**

```ts
function seg(n: number): string { return "s" + String(n); }
path.resolve("/root", seg(1));   // node: /root/s1     perry: TypeError
path.resolve("/root", "s1");     // both: /root/s1
```

`Expr::PathResolveJoin` unboxes its operands with `unbox_to_i64`, so a short computed string's **inline SSO bytes are read as a `StringHeader*`**. Bisected by string length: a segment longer than the SSO threshold works, a short one does not. The correct helper is `unbox_str_handle` — but that one *allocates* (SSO materialisation), so applying it opens the #7213 window in the same arm, which is a rooting change with its own combinator question. Filed as #7621. **Deliberately not fixed here**: this PR is IR-identical-or-justified by construction, and a semantic fix to five path arms is not something to hide inside a refactor. Confirmed present on both arms.

**3. Codegen output is not run-to-run deterministic (#7622).** See the IR-identity section: string-constant numbering, `js_register_function_name` order, and `perry_method_*` callee selection permute between two compiles of the same source with the same binary. This does not affect correctness of any arm here, but it does mean a byte-level IR A/B needs the multiset/normalised comparison this PR used, not `diff`.

## Not in this PR

No version bump. No behaviour change beyond the itemised rooting fixes — no instruction kind is deleted. No gate widening, no allowlist entry, no new test file (and therefore no `check_test_registration.py` registration to make). The SSO bug above, the benchmark-baseline gate, and the nondeterminism are reported, not repaired.

Advances #7615. Slice 2 (`expr/instance_misc1.rs`, `expr/logical_collections.rs`, `lower_call/property_get/map_set.rs`) is next.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runtime safety when evaluating multi-step array, buffer, collection, path, date, object, and typed-array operations.
  * Preserved correct dynamic property behavior and `undefined` results for out-of-range typed-array and buffer access.
  * Improved array type detection and handling of array callbacks.

* **Documentation**
  * Added migration notes and documented known limitations.

* **Chores**
  * Updated the application version to 0.5.1355.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1354
+**Current Version:** 0.5.1355
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1354"
+version = "0.5.1355"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "perry-ffi",
  "scraper",
@@ -5813,7 +5813,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-commander"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5821,7 +5821,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cron"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "chrono",
  "cron",
@@ -5831,7 +5831,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dayjs"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "chrono",
  "perry-ffi",
@@ -5839,7 +5839,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-decimal"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "perry-ffi",
  "rust_decimal",
@@ -5847,7 +5847,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dotenv"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "perry-ffi",
  "serde_json",
@@ -5855,7 +5855,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ethers"
-version = "0.5.1354"
+version = "0.5.1355"
 dependencies = [
  "perry-ffi",
  "rand 0.10.1",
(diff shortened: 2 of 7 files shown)


proggeramlug:
## Audit before merge — verified, merged as v0.5.1355

**Ledger sabotage red** on a real `temp_root_push_i64` injected into
`arrays_finds.rs`; the per-module discipline holds for both new entries.

**The `find*` fix verified behaviourally under a live protector**: all four
`find`/`findIndex`/`findLast`/`findLastIndex` with allocating callbacks over a
5,000-element pointer array — identical to node plain AND under
`PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1` with **5 retired quarantine
sets** (matching the report's own liveness figure). That was the arm holding
the array across `js_closure_new`; it no longer does.

**Root-dominance re-run**: 129/129, 0 violations, 40/40 seeded caught. Full
suites (runtime 1,886/0 first try, codegen 690/0), the full lint script set
including `check_test_registration` and `gc_store_site_inventory`, fmt — all
clean.

**Three things this slice adds to the campaign's method, worth naming:**

1. **The raw-i64 taxonomy point.** #7280's `root_reload` re-reads a `double`
   from a shadow slot — it structurally cannot repair a pointer *already
   unboxed to i64* above the window. `BufferSlice` was exactly that, and the
   distinction now anchors which sites each mechanism owns.
2. **Filing #7621 instead of smuggling it.** The `path.resolve` SSO bug is a
   real user-facing throw, found in-slice, and NOT a rooting bug — fixing it
   here would have broken the slice's IR-identity contract. Kept out, filed
   with the bisect evidence.
3. **#7622 (codegen run-to-run nondeterminism) contaminating 3 of 9 corpus
   diffs** — proven by compiling one source twice with the same binary before
   attributing anything to the change. Every future slice's IR-identity claim
   needs that control; it is now on the record as required method.

The combinator addition (`with_operands_rooted_across`, with the plain form as
its empty case) keeps `operand_protection` the single decision site — the right
call versus a second parallel path.

**Fallout handled separately**: the harness-fingerprint redness this slice's
full-lint run surfaced traces to #7601's functional edit of
`json_pipeline.ts`; the baseline is regenerating now.

PR #7623, opened 2026-08-08
gc: pretenure-accumulator admission infrastructure (#7598, scope-reduced per audit)

proggeramlug:
**Scope reduced per the two audits — admission infrastructure only.** The original PR is preserved in the branch history (`6e73919c8`/`f0a2e760a`); this is what remains after acting on the root-cause finding.

## What the audits established (both confirmed on my side)

- The mechanism was implemented correctly — the 11-field wrapper site was admitted, born old, TENURED, no fallback, IR- and allocator-chain-verified.
- **The target was wrong**: the ~113 MB the minor moves on json_pipeline is the **runtime-allocated parse cohort**, not codegen-visible literals (~12 MB total, ~1 MB live at minor time).
- **My measured 108 MB → 0 was a confound**: the two arms differed in whether #7613's promote-on-first-copy seed fired — my base arm predated it. On current main the seed fires in both arms and pretenure's true marginal is ~1 MB. The audit's pinned-host census stands; my dev-host numbers do not.

## What this PR now contains

**Only the admission machinery, kept for a future dynamic-feedback pretenurer** (the audit's disposition):

- `collect_pretenure_accumulator_locals` — the loop-position proof (accumulator `let` at loop depth 0, every push at depth ≥ 1) layered on the all-pointer terms, with refusal tests for the per-iteration accumulator, mixed-depth pushes, out-of-loop pushes, nested-expression pushes, and non-admitted locals.
- The `region_runs_once` parameter threaded explicitly through both fact-graph builders and every call site (module main/init = true; function/method/closure = false), with a graph-level test pinning both polarities. A function region's depth-0 accumulator is re-entered per call — the measured 6.6× adversarial regression — so only run-once regions may ever admit.

**Removed, deliberately:**

- The pretenured allocator entry points, their keepalive anchors, and the codegen consumers — an unused `#[no_mangle]` entry point plus `#[used]` anchor is unused configuration (kill-policy) and un-strippable bytes (the hello-size anchor lesson). Infrastructure without callers is tests + collector only.
- The deferred-page-registration fix — being extracted separately (crediting this PR's finding) on `perf/old-page-registration-deferral`; not duplicated here. Caller knowledge relevant to that PR: the flush-before-cycle invariant is satisfied by all three cycle constructors (`GcCycleState::new_full`, `gc/mod.rs:203`, `policy.rs`'s budgeted arm), and `unregister_old_object_pages` is dead code, so the index accretes stale entries — the `contains` dedup is load-bearing for hole-reused addresses specifically.

## For #7598

The constraint that must drive the next design, from the audit: **the moved cohort is runtime-allocated**, so the live routes are dynamic feedback or allocation-context pretenure inside the JSON materialiser. No admission refinement fixes that; json_pipeline's literals are structurally not the target.

## Process (acknowledged)

Future measurements in this family: both arms against current main on the pinned mini, moved-bytes census quoted alongside wall, and the gap-suite result posted on the PR before merge consideration. The gap suite for this branch is running; its result lands here as an edit to this body.

## Summary by CodeRabbit

  * Improved memory management for eligible array accumulators created outside loops and populated within loops.
  * Added safeguards so this optimization applies only to code paths that execute once, avoiding changes in repeated regions.
  * Expanded coverage across module initialization, functions, closures, and methods for more consistent behavior.

* **Documentation**
  * Documented the new static allocation optimization and future enhancement areas.


proggeramlug:
## Audit — NOT merged. The headline does not reproduce against current main on the pinned host.

Rebased onto v0.5.1355 (clean), built both arms identically (perry-dev,
`PERRY_NO_AUTO_OPTIMIZE=1`), and ran the A/B on the **pinned quiet mini**
(load 1.2), three interleaved rounds, hash-verified:

| | main v0.5.1355 | this PR | claimed |
|---|--:|--:|--:|
| json 200k wall (mini, best/3) | 2.74 s | 2.74 s | 2.40 → **1.54 s** |
| peak RSS | 464 MB | 466 MB | 543 → **431 MB** |
| minor-moved bytes (census) | 113,227,216 | **112,178,784** | 108 MB → **0.0** |
| cycles | 3 | 3 | — |

Wall identical, RSS identical, and the census is the tell: **Δ ≈ 1 MB of
~113 MB**. The mechanism is not merely underperforming — it is barely firing.

**What I verified is right**: the IR admission is correct — exactly **one**
`_pretenured` call site in json_pipeline's `main`, **zero** in my adversarial
function-local-accumulator probe (the 6.6×-regression shape is refused), no
competing inline bump in the same function. The deferred-page-registration
soundness question closes cleanly too: `classify_heap_generation` (every
barrier remember-decision) reads the block-level `hot_page_generations` map,
not the deferred `OLD_GEN_PAGE_OBJECTS` object index, which flushes before
every cycle constructor. And output hashes are identical everywhere.

**The open question the author-side needs to answer**: the admitted site is
emitted and reachable, yet at runtime the moved-bytes delta says the accumulator
cohort is still being born young — either the pretenured allocator's fast path
is falling back, or the one admitted site is not the hot `out.push({...})`, or
the base your table was measured against differs from what `55ddab012` +
rebase produces. Your base row itself is suspect against the record: base
"copied by minors 108 MB" and RSS 543 MB do not match what #7613-era main
measures on the mini (single-hop 113 MB, 464 MB). Please re-measure both arms
against **current main (v0.5.1355)** on the pinned host and post the census
with `PERRY_GC_TRACE` cycle detail — if the win is real there, I will re-audit
immediately.

**Worth salvaging regardless of the pretenure verdict**: the
`register_old_object_pages` quadratic-per-allocation finding (two RefCell
borrows + a growing `contains` scan per object) plausibly taxes #7613's
promote path on every cycle today. If the pretenure half does not hold up,
split the deferred-registration fix into its own PR with its own A/B — it may
be the genuinely valuable part.

(Also noting: this is the third PR in this family whose promised gap-suite
comment never got posted. The A/B above is what stands in.)

proggeramlug:
## Follow-up: root cause found. The mechanism works; the measured win was a confound.

Traced the whole chain on the rebased branch:

1. The admitted IR site **is** the hot wrapper — `call
   @js_object_alloc_class_inline_keys_pretenured(i32 1, i32 0, i32 11, …)`,
   `field_count = 11`.
2. `alloc_class_inline_keys_impl(pretenure=true)` routes **unconditionally** to
   `arena_alloc_gc_old_born_tenured_bump`, which is itself unconditional —
   old-arena bump, `GC_FLAG_TENURED` set in the same breath. **There is no
   runtime fallback.** The wrappers are genuinely born old.

So why is the census delta ~1 MB? Because **the ~113 MB the minor moves is the
parse cohort** — the 200k parsed records and their subtrees, allocated by the
runtime JSON parser, kept live through `out` — **not the accumulator
wrappers.** The wrappers total ~12 MB across the whole run, and at the moment
the single copying minor fires, only ~1 MB of them exist. That 1 MB is exactly
the delta I measured. Pretenuring codegen-visible literals cannot touch the
parse cohort, which is runtime-allocated.

**Which means the PR's measured 108 MB → 0 cannot have been pretenure's
doing.** The coherent explanation for the base/pretenure split in the PR table:
the two arms differed in whether **#7613's promote-on-first-copy seed fired** —
the pretenure arm's old-gen births change early cycle structure (a full
mark-sweep runs sooner, the sweep census seeds the lock, the parse cohort
single-hops), while the base arm's cycle order left the seed cold. On current
main the seed fires in **both** arms (my censuses: 3 cycles, ~113 MB single-hop,
both sides), so the confound disappears and pretenure's true marginal is the
~1 MB I measured. The base row's 543 MB RSS and 2,887 ms pause — neither
reproducible on the pinned host — point the same direction.

## Disposition

- **The pretenure half should not merge for this workload** — its achievable
  ceiling here is ~12 MB of moves, and the honest path to more is the PR's own
  stated alternative: dynamic feedback, not a better static proof. The
  admission machinery (run-once regions, refusal tests) is good infrastructure
  for that future and is worth keeping on the branch.
- **The deferred-page-registration fix should be extracted into its own PR
  with its own A/B.** `register_old_object_pages` doing two RefCell borrows
  plus a growing `contains` scan per object plausibly taxes **#7613's promote
  path on current main today** — 113 MB of promotions per run through that
  code. That is the measurable, standalone win hiding in this PR.
- #7598 gets this diagnosis so the next attempt starts from "the moved cohort
  is runtime-allocated" rather than re-deriving it.

Closing the loop on my earlier comment: the author-side re-measure I asked for
is no longer needed — the confound is identified. What would change my mind on
the pretenure half is a workload where codegen-visible literals ARE the moved
cohort; `json_pipeline` is not it.

proggeramlug:
## Re-audit of the stripped PR — the strip is correct, but this should not merge yet

Verified the restructure does what it says. **IR is normalized-identical to
main** on json_pipeline, **zero `_pretenured` call sites** remain, the allocator
entry points and their `#[used]` keepalive anchors are gone (which was the
specific thing I asked for — a dead `#[no_mangle]` entry plus anchor is both an
unexercised configuration and un-strippable bytes). Codegen suite 697/0. The
seven collector tests are real and the run-once gate is **sabotage-verified**:
deleting the `region_runs_once` clear turns
`pretenure_accumulator_fact_requires_a_run_once_region` red.

**Why not merge:** the collector now runs on **every compile of every region**
and computes a `HashSet<u32>` that **nothing reads**. `pretenure_accumulator()`
has exactly two callers, both in tests. This isn't dead code the linker drops —
it is dead *work* in the compile path, and the `region_runs_once` parameter is
now threaded through five fact-graph call sites (`closure.rs`, `entry.rs` ×2,
`function.rs`, `method.rs`) to feed it.

That is the shape CLAUDE.md's kill-policy exists for, one level up from env
knobs: **infrastructure whose only exerciser is its own test suite is a
decision that hasn't been made.** Concretely, the risk is not the microseconds
— compile time is unmeasurably changed at this size — it is that six months
from now the collector's admission rules will have drifted against a consumer
that never arrived, and the tests will still be green.

**Two ways I'd take it, either is fine:**

1. **Land it with its consumer** whenever the pretenure work resumes on the
   route #7598 now records (dynamic feedback, or allocation-context pretenure
   in the JSON materialiser — the static route cannot reach the moved cohort,
   which is runtime-allocated). The branch keeps; nothing is lost.
2. **Land it now as an explicitly-parked component** if you want the analysis
   banked — but then the `region_runs_once` threading should carry a comment at
   each of the five call sites saying it feeds a currently-unconsumed fact, and
   the changelog fragment should say "no behaviour change, no consumer yet"
   rather than reading as a shipped optimization. I'd want #7598 linked from
   the collector's doc comment as the consumer-to-be.

I have no objection to (2) if you prefer to bank it — say which and I'll merge
or hold accordingly. What I won't do is merge it described as an optimization,
because it currently optimizes nothing.

**Unchanged from my previous comment**: the deferred-page-registration half is
the genuinely valuable finding and is being extracted separately — it sits on
#7613's promote path on current main, ~113 MB/run through the quadratic
`register_old_object_pages`. That extraction credits this PR as its origin.

proggeramlug:
## Author decision: option 1 — land it with its consumer, so closing now

Taking the first of the two offered paths, and the audit's own root-cause finding is the reason: the recorded #7598 routes are **dynamic feedback** or **allocation-context pretenure inside the JSON materialiser**, and neither necessarily consumes this static collector — a materialiser-context pretenurer lives inside the runtime parser and never sees HIR accumulator proofs, and a dynamic-feedback pretenurer keys on runtime allocation sites, with the static admission at most a candidate filter. Merging the collector parked would put dead work in every compile and five threaded call sites on main as a bet on the *least* likely consumer, with the drift-against-no-consumer risk you named.

The infrastructure is preserved where it costs nothing: branch `perf/7598-static-pretenure` (head `eaa7057b7` — collector, five refusal tests, `region_runs_once` threading, graph-level polarity test, all green), referenced from #7598 alongside the constraint that must drive the next design. If the future pretenurer turns out to want the static admission after all, it lands together with it — and gets re-reviewed against whatever the admission rules have to be *then*, which is exactly the drift protection you were pointing at.

Closing unmerged. The deferred-page-registration extraction continues separately as planned; the caller notes in this PR's body stand for that work.

Issue #7038, opened 2026-07-30
codegen: same input compiles to different LLVM IR run-to-run — HashMap iteration order in closure-source registration breaks reproducible builds


proggeramlug:
Compiling the same file twice with the same compiler and the same flags produces **different LLVM IR**, and therefore different object files and binaries.

## Reproduce

```console
$ for i in 1 2 3 4; do
    mkdir -p run$i && cp batch.ts run$i/
    (cd run$i && perry batch.ts -o o --trace llvm >/dev/null 2>&1)
  done
$ md5 -q run*/.perry-trace/llvm/batch_ts.ll | sort | uniq -c
   1 c8f51e1f7ca5ba71e6cd785827e0e976
   1 d5f7462479bf33726126117e038f8126
   1 d67dc9790fa27305ab027a73285e67df
   1 f34c01e832bac8092790c95e90945c96
```

Four runs, four distinct outputs. (`benchmarks/app-patterns/kernels/batch.ts`, added in #7037, is a convenient repro — any file with several closures works.)

## What varies

Only the numbering assigned to retained closure-source string constants, and the order of the matching `js_register_function_source` calls in `__perry_init_strings_<prefix>`:

```diff
-@.str.10 = private unnamed_addr constant [292 x i8] c"(acc: Summary[], r: Row) => {...}"
-@.str.11 = private unnamed_addr constant [160 x i8] c"(a, b) => {...}"
-@.str.12 = private unnamed_addr constant [130 x i8] c"(r) => ({...})"
+@.str.10 = private unnamed_addr constant [160 x i8] c"(a, b) => {...}"
+@.str.11 = private unnamed_addr constant [130 x i8] c"(r) => ({...})"
+@.str.12 = private unnamed_addr constant [292 x i8] c"(acc: Summary[], r: Row) => {...}"
```

Canonicalising `@.str.N` by content makes all runs hash identically, so it is a pure permutation — semantics are unaffected and `fn.toString()` still resolves correctly.

## Root cause

`crates/perry-codegen/src/codegen/artifacts.rs:1861`

```rust
for (func_id, src) in &hir.closure_source_text {
```

`Module::closure_source_text` is a `std::collections::HashMap<FuncId, String>`
(`crates/perry-hir/src/ir/module.rs:132`). Iterating it directly yields
`RandomState`-seeded order, which differs per process. The sibling loop just
above it walks `hir.functions` (a `Vec`) and is deterministic; only this second
[...]

Issue #235, opened 2026-04-28
Async miscompile: an extra synchronous statement before await MongoClient.connect() drops the next await silently


proggeramlug:
**Perry:** 0.5.346 native (macOS aarch64). Discovered while running our `@perryts/mongodb` driver bench — the workload completes silently with `undefined` results instead of real query data. Bisected to a code-shape sensitivity in Perry's async-state-machine codegen.

The shape is small enough to pin down but I haven't been able to minimize away the dependency on our driver (a real socket-resolved Promise is needed; setTimeout-resolved Promises don't trigger it). Filing as-is so the team can see the repro; happy to follow up with further minimization if useful.

## Repro

Two near-identical programs. They differ by **one line** — the "fails" version has an extra `console.log('A connecting')` before `await MongoClient.connect(...)`. Both are semantically equivalent and both produce the expected output on Bun 1.3.12 and Node 25.8.0. Under Perry, the "fails" version drops the second await silently — the program reaches the line just before `await col.findOne(...)`, then the process exits without printing anything that would follow the await.

### `works.ts` — runs as expected on Perry

```ts
import { MongoClient } from "@perryts/mongodb";

async function main() {
    const c = await MongoClient.connect("mongodb://127.0.0.1:27017");
    const col = c.db('sample_mflix').collection('movies');

    console.log('1 findOne start');
    const r = await col.findOne({ title: 'The Great Train Robbery' });
    console.log('2 findOne done:', r === null ? 'null' : 'doc');
    c.close();
    console.log('3 closed');
}
main().catch((e: any) => console.error('ERR:', e));
```

Output (Perry 0.5.346):
```
1 findOne start
2 findOne done: doc
3 closed
```

### `fails.ts` — silently exits mid-flow on Perry

Identical except for one `console.log` added before `MongoClient.connect`:

```ts
import { MongoClient } from "@perryts/mongodb";

async function main() {
    console.log('A connecting');                                       // <-- ONE EXTRA LINE
    const c = await MongoClient.connect("mongodb://127.0.0.1:27017");
    const col = c.db('sample_mflix').collection('movies');

    console.log('1 findOne start');
    const r = await col.findOne({ title: 'The Great Train Robbery' });
    console.log('2 findOne done:', r === null ? 'null' : 'doc');
    c.close();
    console.log('3 closed');
}
main().catch((e: any) => console.error('ERR:', e));
```

Output (Perry 0.5.346):
```
A connecting
1 findOne start
```
[...]

PR #7625, opened 2026-08-08
fix(codegen): make emitted IR run-to-run deterministic (#7622)

### LINKED ISSUE #7303: Emitted-IR nondeterminism: js_register_function_name order flips run-to-run (rayon completion order)
Found while proving #7301's default path unchanged (object-level branch-vs-merge-base comparison); pre-existing at `9cb31f1b0`, unrelated to the in-process backend.

**Symptom:** the order of adjacent `js_register_function_name(ptr @perry_closure_..., ptr @.str.N, ...)` registration calls in module init flips between two runs of the **same** perry binary on the same input. The `@.str.N` numbering follows first-use, so the string-constant numbering swaps with it. Everything else in the emitted `.ll` is byte-stable.

**Repro** (macOS/arm64, perry-dev build, observed on ~1-in-3 runs):

```bash
for i in 1 2 3 4; do
  rm -rf node_modules  # cold object cache each run
  mkdir -p ll_$i
  PERRY_SAVE_LL=$PWD/ll_$i PERRY_NO_AUTO_OPTIMIZE=1 \
    perry test-files/test_gap_1840_class_iterator_for_of_spread.ts -o /dev/null
done
cmp ll_1/*.ll ll_2/*.ll   # flips within ~4 runs
```

Also reproduces on `test_gap_yield_star_iterable.ts`. Two closures whose registrations swap are always a pair with equal-length names.

**Suspected mechanism:** registration entries are collected in rayon completion order (per-closure codegen races), not declaration order — consistent with it being timing-dependent yet stable within most consecutive runs. Not a `HashMap` RandomState effect (that would flip near-every run).

**Why it matters:**
- It breaks byte-level determinism of emitted IR on the DEFAULT path — directly relevant to the #7131 determinism-gate work (that effort fixed temp-file names; this is a content-level source).
- It costs real triage time: #7301's default-unchanged proof initially looked like a regression until same-binary runs proved the coin (two tests, both arms, 4–6 runs each).
- Behaviorally harmless today (registration order is not observable), so this is a determinism/infrastructure issue, not a correctness one.

**Suggested fix shape:** sort the registration list by closure symbol name (or emit in declaration order) before rendering module init — a one-point sort at the collection site.

### LINKED ISSUE #7622: Codegen is not run-to-run deterministic: string-constant numbering, function-name registration order, and method-callee selection permute
Compiling the **same source twice with the same `perry` binary** produces different LLVM IR. Not a miscompile — the programs behave identically — but it defeats any byte-level IR A/B, which is the primary evidence a rooting-campaign slice offers (#7615).

## Repro

```bash
for run in 1 2; do
  mkdir -p /tmp/det/$run && cd /tmp/det/$run
  env PERRY_RS4GC=0 PERRY_GC_MOVING_LOOP_POLLS=1 PERRY_INLINE_SHADOW_SLOT=0 \
      PERRY_NO_AUTO_OPTIMIZE=1 \
    perry compile test-files/test_gap_object_create_method_this.ts -o out --trace llvm
done
diff /tmp/det/1/.perry-trace/llvm/*.ll /tmp/det/2/.perry-trace/llvm/*.ll
```

Two independent shapes, both reproducible:

**1. String-constant numbering and function-name registration order.**

```
< @.str.3 = private unnamed_addr constant [9 x i8] c"describe\00"
< @.str.4 = private unnamed_addr constant [9 x i8] c"getState\00"
< @.str.5 = private unnamed_addr constant [5 x i8] c"wrap\00"
---
> @.str.3 = private unnamed_addr constant [5 x i8] c"wrap\00"
> @.str.4 = private unnamed_addr constant [9 x i8] c"describe\00"
> @.str.5 = private unnamed_addr constant [9 x i8] c"getState\00"
```

with the matching permutation of `js_register_function_name` calls in `__perry_init_strings_*_chunk0`.

**2. Method-callee selection permutes.** In `test_gap_class_expr_dynamic_parent_ctor`, three consecutive dispatches swap which `perry_method_*` symbol each call site names:

```
<   %r2492 = call double @perry_method_..._DerA__promise(double %r2956)
>   %r2492 = call double @perry_method_..___anon_class_17__promise(double %r2956)
<   %r2493 = call double @perry_method_..___anon_class_17__promise(double %r2955)
>   %r2493 = call double @perry_method_..._DerB__promise(double %r2955)
```

Both smell like iteration over a `HashMap`/`HashSet` whose order is randomised per process.

## Why it matters beyond tidiness

- **It masquerades as a regression.** Comparing IR between two compilers is how the Layer 1 slices demonstrate "root plumbing only". In #7620 three of nine differing functions were this, not the change — distinguishable only because the instruction *multiset* was identical and a same-binary double compile reproduced the same permutation. A slice that used `diff` would have reported three phantom differences; one that used a multiset comparison and did **not** check determinism would have quietly dismissed a real ordering change.
- **`.perry-cache` keys on codegen env vars (#6394), not on output.** A nondeterministic emitter means two cache entries for the same inputs can differ.
- Reproducible builds and any future "did this refactor change codegen?" gate both need this closed.

Filed from #7620; not fixed there.

proggeramlug:
Closes #7622.

Compiling one source twice with the same `perry` binary produced different LLVM IR. Both shapes the issue names are `std::collections` hash-map iteration order reaching the emitter — Rust's default hasher is seeded per `RandomState`, so the order is a fresh permutation every process.

## Variance taxonomy

I compiled the issue's two named sources six times each with one baseline binary and diffed pairwise. **5 of 5 runs differed from run 1** for both. The entire diff is two shapes; nothing else in the `.ll` moves.

| shape | what varies | diff size |
|---|---|---|
| **1. Function-name registration** | the `@.str.N` numbering of display-name constants, and the order of the matching `js_register_function_name` calls in `__perry_init_strings_*` | 24 lines |
| **2. Dispatch-tower arms** | which `perry_method_*` symbol each `icmp`-guarded case block names, and the class ids in the matching `icmp eq i32` chain | 32 lines |

Nothing else was found: no register renumbering, no block reordering, no metadata churn, no `Instant`/address-derived value, and **no rayon-completion effect** — `ctx.native_modules` is a `BTreeMap<PathBuf, HirModule>` (`compile/types.rs:557`), so #7303's suspected mechanism is not what is happening. Both shapes are `HashMap`, as #7622 suspected.

Shape 2's diff is diagnostic on its own: the case blocks keep their register numbers and their arguments and only the callee symbol and the compared class id move — arms emitted in a permuted order, not different arms.

```
<   %r2492 = call double @perry_method_..._DerB__promise(double %r2956)
>   %r2492 = call double @perry_method_..._DerA__promise(double %r2956)
...
<   %r2488 = icmp eq i32 %r2487, 7          >   %r2488 = icmp eq i32 %r2487, 2
<   %r2489 = icmp eq i32 %r2487, 2          >   %r2489 = icmp eq i32 %r2487, 9
```

## Root causes

**1. `crates/perry-codegen/src/codegen/artifacts.rs:1836`** — `for (func_id, display) in &hir.closure_display_names` (a `HashMap<FuncId, String>`). Each entry mints a rodata constant through `add_string_constant`, whose `@.str.N` counter numbers in first-use order, and emits one `js_register_function_name` call. So the map's order set both.

This is the same defect **#7038 fixed 36 lines below**, in the `closure_source_text` loop, and left standing here. Its own comment explains exactly why it matters.

**2. `crates/perry-codegen/src/lower_call/property_get/dynamic_dispatch.rs:241`** — `for (start_cls, &start_cid) in ctx.class_ids.iter()` (a `HashMap<String, u32>`) builds `implementors`, which is walked by index to emit one case block per implementing class. Map order *was* arm order.

Two mechanical siblings found by a systematic sweep of all four compiler crates rather than by following the diff, and fixed in the same commit:

**3. `dynamic_dispatch.rs:708`** — the *virtual-override* tower (`vdispatch.*`) reads the same map the same way.
**4. `codegen/method_registry.rs:42`** — `for c in class_table.values()`, whose body writes into the `(class, method) -> symbol` registry both `insert` (last writer wins) and `entry().or_insert_with` (first writer wins). `class_table` mixes local classes, their alias keys and imported stubs, so two distinct `&Class` can contend for one key, and that registry is what every call site consults to pick its callee.

## The fix, and its ordering guarantee

Sorting at the point of iteration, not serializing the compiler. **Rayon parallel module codegen is untouched** — it was never the cause.

| site | order now |
|---|---|
| `artifacts.rs` | ascending `FuncId` (total: filtered to unique, materialized ids) |
| `dynamic_dispatch.rs` interface tower | ascending `(class_id, class_name)` (total: `class_name` is the map key, so unique) |
| `dynamic_dispatch.rs` virtual tower | same |
| `method_registry.rs` | ascending class-table key (total: it is the map key) |

Every key is total, so each is a single fixed permutation, not merely "more stable".

One behaviour change, deliberate. `max_explicit_arity`'s scan (`dynamic_dispatch.rs:794`) stopped at the **first** name carrying a class id, and class ids are *not* unique over `class_ids` keys — a class-expression self-binding alias (`var X = class _X`) and an imported class registered under both its own and its local-alias name each map two names to one id. So it was a hash-order tie-break, and the loser's arity decides how many `TAG_UNDEFINED` padding args **every** emitted call in the tower carries. It now takes the max over all names sharing the id, which is what the variable already means and is the safe direction: under-padding is the #235 garbage-argument bug, over-padding just lets a default-param desugaring fire.

## Is the object cache implicated? Yes — say it plainly

`compute_object_cache_key` (`compile/object_cache.rs:253`) is a function of `CompileOptions`, the post-transform HIR fingerprint, the perry version, a hash of the perry binary, and the codegen env vars. All deterministic. The emitted IR was not. So **identical inputs produced an identical key over different `.o` bytes**: a cache hit and a cold rebuild could legitimately hold different code.

For the shapes actually observed that difference is semantically neutral — the name registry is keyed on distinct function pointers, and the tower's arms are keyed on distinct class ids, so first-match picks the same arm whatever the order. But it is not neutral by construction: `seen_pairs` dedups on `(class_id, fname)`, which **admits two arms sharing one class id with different symbols**, and there the emission order *is* the behaviour. Same for `method_registry`'s two tie-breaks. So this was a latent correctness hazard on the cache path, not only an audit-fidelity one, and it is now closed.

## Evidence (all local — CI backlog is deep, so this is the evidence)

- **Repro, baseline binary**: `test_gap_object_create_method_this` 5/5 runs differ; `test_gap_class_expr_dynamic_parent_ctor` 5/5 differ.
- **Same probe, fixed binary**: 0/5 and 0/5.
- **Corpus sweep**, 42 sampled `test_gap_*` compiled 3x each with the fixed binary: **0 nondeterministic / 41 compared**. The 42nd, `test_gap_http_overloads_3226plus`, failed to link during that batch — the host was at 99% disk; on a clean re-run it links under both arms and aborts identically at runtime, see below.
- **Behavioural A/B**, 126 sampled `test_gap_*` compiled with BOTH binaries and the produced executables run and compared: **123 byte-identical, 3 flagged, and all 3 investigated to a non-cause**.
  - `test_gap_http_overloads_3226plus` aborts **identically on both arms** (exit 134, panic at `crates/perry-ext-http/src/server/server.rs:911`); the entire stdout difference is the **thread id inside the panic message**.
  - `test_gap_proto_write_local_shadows_class` and `test_gap_typeof_instanceof` were SIGKILLed (137) on the fixed arm during the concurrent batch, on a host that was at 99% disk and load ~20. Re-run serially on an idle host, 3x per arm: both arms exit 0 with byte-identical output, 3/3 each. Not reproducible.
  - So: **zero behavioural differences over 126 programs.**
- **Sabotage**: reverting the `artifacts.rs` sort alone turns exactly the two shape-1 tests red and leaves shape-2 green; reverting the `dynamic_dispatch.rs` sort alone does the mirror image. Verified both directions.
- **Compile-time impact**: **none measurable**, by user+sys CPU best-of-N (wall clock is unusable on this host — it runs other agents' builds at load 15-27, and an early wall-clock read of "+13%" did not survive a CPU-time re-measure).
  - Worst case for the change (40 classes x 40 dynamic call sites = 1,600 tower arms, sized to stay under the oversized-module `-Os` threshold so LLVM does not bury the signal), best-of-7: **ratio 1.008** (median ratio 0.983). Within-arm spread is 3.56-4.25s, an order of magnitude larger than the gap.
  - 14 real class-bearing `test_gap_*` programs, total CPU, best-of-3: **ratio 0.958**.
  - Expected: the added work is one sort of `(class_id, &name)` pairs per dynamic-dispatch site, in a loop that already clones a `String` per class per site.
- `cargo test -p perry-codegen --lib`: 695 passed.
- `cargo test -p perry-runtime --lib --no-fail-fast`: 1885 passed, 1 failed — `promise::keyed_table::tests::settling_many_keys_is_not_quadratic`, a timing test that passes on rerun and lives in a crate this diff does not touch.
- **gc-root-dominance corpus, both gated modes**: green. 149 files; mode 1 caught 40/40 seeded violations (so the checker was live); mode 2 reports 0 unrooted-alloca violations over 7,863 gc-capable allocas.
- **Full `lint` job step list** from `.github/workflows/test.yml`, run individually: all green except `Public benchmark evidence freshness`, which fails on `main` for reasons this diff cannot reach (it touches only `crates/perry-codegen/`) and which the workflow's own inline comment records as failing on every `main` run since 2026-07-29.
- `cargo clippy -p perry --bins`: exit 0.

## #7303 is the same defect and is also closed

#7303 reports shape 1 exactly — `js_register_function_name` order flipping, with the `@.str.N` numbering following it — and attributes it to rayon closure-codegen completion order. It is the `hir.closure_display_names` `HashMap`. On its own two repro files, with the same binaries used above:

| source | baseline | fixed |
|---|---|---|
| `test_gap_1840_class_iterator_for_of_spread` | 2 of 5 runs differ | 0 of 5 |
| `test_gap_yield_star_iterable` | 2 of 5 runs differ | 0 of 5 |

That 2-in-5 is #7303's own "~1-in-3 runs", and it is also where its reasoning went wrong: it ruled out `RandomState` because "that would flip near-every run". With only two registration entries a hash order agrees with itself about half the time, so ~1-in-3 is exactly what a two-element `HashMap` looks like. The frequency argument needs the entry count, and both of those files have two.

## Tests

Four `--lib` unit tests in `crates/perry-codegen/src/codegen/emission_order_tests.rs`, so they run on every PR that touches perry-codegen rather than only in the tag-gated integration tier.

They are built not to be vacuous, both ways:

- Each builds its `Module` **fresh for every compile**. The offending maps live in the HIR, so compiling one long-lived `Module` twice would re-iterate the very same `RandomState` and pass unconditionally. With `N = 16` entries, a chance-ordered agreement is a 1-in-16! event.
- Each **asserts its subject was live** before judging order — the registration count and the tower arm count — so a fixture that stopped emitting the construct cannot go green having proven nothing.

**What is deliberately not tested, and why.** The virtual-override tower and the `method_registry` walk are sorted with no test. `vdispatch` blocks appear in **zero** of the 41 sampled programs — every receiver-typed call measured is claimed earlier by `method_override.rs`'s `method_direct` shape guard — and I could not construct a fixture that reaches it. A green test over a fixture emitting no arms asserts nothing, which is worse than no test. They are sorted anyway because they are mechanical siblings of the defects that *are* covered, and #7622 exists precisely because #7038 fixed one such loop and left its neighbour. The test module says so in prose.

## Out of scope

The same sweep found three hash-order reads that do **not** reach emitted IR and are left alone: `dialect/mod.rs:1556` (which undefined register an error message names), `compile/host_config.rs:841` (which offending package a `bail!` reports), and `commands/fix_applier.rs:93/128` (the ordering of the `perry fix` report). Worth a follow-up; not this PR, whose diff should stay auditable.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1355
+**Current Version:** 0.5.1356
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1355"
+version = "0.5.1356"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "perry-ffi",
  "scraper",
@@ -5813,7 +5813,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-commander"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5821,7 +5821,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cron"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "chrono",
  "cron",
@@ -5831,7 +5831,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dayjs"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "chrono",
  "perry-ffi",
@@ -5839,7 +5839,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-decimal"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "perry-ffi",
  "rust_decimal",
@@ -5847,7 +5847,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dotenv"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "perry-ffi",
  "serde_json",
@@ -5855,7 +5855,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ethers"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "perry-ffi",
  "rand 0.10.1",
@@ -5863,7 +5863,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-events"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5871,14 +5871,14 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-exponential-backoff"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "perry-ffi",
 ]
 
 [[package]]
 name = "perry-ext-fastify"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "bytes",
  "http-body-util",
@@ -5896,7 +5896,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-fetch"
-version = "0.5.1355"
+version = "0.5.1356"
 dependencies = [
  "bytes",
  "lazy_static",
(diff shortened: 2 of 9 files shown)


proggeramlug:
## Audit before merge — verified, merged as v0.5.1356

**Reproduced the bug and the fix on my own probe** (3 classes with a shared
method name + 2 named functions, five compiles per arm, same binary each time):

| arm | identical to run 1 |
|---|---|
| main | **1/4** — 3 runs differed |
| this PR | **4/4** |

And main's diff is exactly shape 2 as described — `perry_method_…__B__foo` and
`__C__foo` trading positions in the tower, registers and arguments holding
still. The taxonomy is right.

**Sabotage re-verified**: commenting out the `dispatch_roots` sort turns exactly
the two tower tests red (`dispatch_tower_arms_are_emitted_in_class_id_order`,
`dispatch_tower_emission_is_run_to_run_deterministic`) and leaves the two
closure-name tests green — the per-shape isolation the report claims.

Gates re-run: codegen 694/0, runtime 1886/0, all six lint scripts + file-size +
fmt clean.

## Why this one mattered more than its diff size

Every Layer-1 slice audit — mine and the agents' — rests on *"emitted IR
byte-identical"*, and slice 1b had 3 of 9 apparent diffs contaminated by this.
Slice 2 (#7627, in review now) had to run a double-compile control first and
still found 5 ordering-only permutations in its corpus. With this merged, the
remaining ~85 modules get a clean instrument.

**Three findings I want on the record:**

1. **The object cache was implicated, and that is a correctness argument, not
   an audit-fidelity one.** `compute_object_cache_key` is deterministic while
   the IR was not, so identical inputs keyed identical caches over differing
   `.o` bytes. Benign for the observed shapes — but `seen_pairs` dedups on
   `(class_id, fname)`, which admits two arms sharing one class id with
   different symbols, and there emission order *is* behaviour. Latent, now
   closed.
2. **#7303 closed as the same defect, and its stated reasoning refuted.** That
   issue argued "not `RandomState`, that would flip near-every run" — but its
   two files have exactly *two* registration entries, and a two-element hash
   order agrees with itself about half the time, which is precisely its
   observed ~1-in-3. Correcting a wrong premise in a closed issue is worth as
   much as the fix.
3. **The tests were nearly vacuous twice and both were caught** — once by the
   agent (a reused `Module` re-iterates the same `RandomState` and would pass
   unconditionally, so each test builds fresh), once by CodeRabbit (the tower
   fixture gave ids in name order, so a name-keyed sort would also have
   passed; ids now run opposite to names). And two sorts ship **explicitly
   labelled untested** because no fixture reaches them — a green test over a
   fixture that emits nothing is worse than none.

Also noted: the "+13% compile time" the agent first saw did not survive CPU
re-measurement (best-of-7 ratio 1.008) — correct handling on a host running
other agents at load 15–27.

PR #7198, opened 2026-08-01
fix(codegen): close #7192's own residual root-store hole, and make the dominance checker able to fail

proggeramlug:
Post-merge follow-up to #7192, from its review round. Every finding was taken to a reproducer or to code before it was accepted or declined.

## The four review findings

| finding | verdict |
|---|---|
| `lower_call/new.rs:1870` — `ctor_result_slot` | **real, fixed** — it defeated #7192's fix one instruction after it landed |
| `logical_collections.rs:927` — `protect_handle` | **half real, fixed; half declined with evidence** |
| `index_set.rs:1494` — the key operand | **real, fixed** |
| `gc_root_dominance_check.py:135` — `%entry.implicit` | **rejected as written** — no such code; the adjacent defect is worse and is fixed |
| `gc_root_dominance_check.py:655` — empty input exits 0 | **real, fixed**, plus three more ways the gate could not fail |

### 1. The inline-constructor result slot — the load-bearing one

`ctor_result_slot` is a plain `alloca_entry`: not a shadow slot, not a temp root, never rewritten by the collector. Seeding it with `obj_box` parked the **pre**-constructor instance address in unrooted memory for the whole body. On fall-through, `js_ctor_return_override` saw an *object* in `raw` and returned **that** — discarding the instance `reload_instance` had just re-read.

Two things make it worse than it looks:

- **It is not gated on `PERRY_INLINE_CTOR`.** `force_ctor_call` requires `class.constructor.is_some()`, so any class with fields or heritage but no own constructor — `class C { payload = mk() }`, `class C extends B {}` — takes the inline path by default.
- **#7192's own regression test cannot see it.** `the_new_instance_is_rooted_across_the_constructor_body` walks the *first* operand of the override; that is the one that was re-read. The value actually published on fall-through is the second.

The slot now starts at `undefined`, which is exactly equivalent on all four paths and carries no address:

| path | before | after |
|---|---|---|
| fall-through | `raw` = stale `obj_box`, an object ⇒ returns the **stale** address | `raw` = undefined ⇒ returns the re-read instance |
| bare `return;` | slot untouched ⇒ same as fall-through | same, correct |
| `return <expr>` | `Stmt::Return` overwrote the slot | unchanged |
| inherited-symbol ctor | the call's return value overwrote it | unchanged |

[...]

Issue #2995, opened 2026-05-30
node:path: validate path.relative arguments on compiled paths


andrewtdiz:
## Summary
Node `path.relative(from, to)` requires both arguments to be strings and throws `TypeError ERR_INVALID_ARG_TYPE` when either argument is not a string. Perry's dynamic native dispatch validates those arguments, but the compiled fast path lowers `path.relative()` directly to runtime helpers that silently treat invalid string pointers as empty strings.

## Node 25.9.0 behavior
```js
const path = require("node:path");

for (const args of [[1, "x"], ["x", 1], [null, "x"], ["x", undefined]]) {
  try { path.relative(...args); }
  catch (err) { console.log(err.name, err.code); }
}
// TypeError ERR_INVALID_ARG_TYPE for every case
```

The same validation applies to `path.posix.relative()` and `path.win32.relative()`.

## Perry source evidence
- `crates/perry-runtime/src/object/native_module_dispatch.rs` uses `require_path_str_ptr(0)` / `require_path_str_ptr(1)` for dynamic `path.posix.relative` and `path.win32.relative`, so that path is guarded.
- `crates/perry-hir/src/lower/expr_call/module_static.rs`, `globals.rs`, and `nested_namespace.rs` lower recognized `path.relative(...)` calls with two arguments to `Expr::PathRelative` / win32 method expressions without adding argument validation.
- `crates/perry-codegen/src/expr/instance_misc1.rs` lowers `Expr::PathRelative` by unboxing both operands and calling `js_path_relative(from, to)` directly.
- `crates/perry-runtime/src/path.rs::js_path_relative()` reads both inputs with `string_from_header(...).unwrap_or_default()`; `js_path_win32_relative()` does the same. Invalid/non-string operands therefore become `""` instead of throwing.

## Expected
All lowering paths for `path.relative()`, `path.posix.relative()`, and `path.win32.relative()` should perform the same string validation as dynamic dispatch and throw `TypeError ERR_INVALID_ARG_TYPE` for either non-string argument.

## Duplicate checks
- `path.relative non-string type error`
- `PathRelative js_path_relative non-string`
- PR search for `path.relative type`

I did not find an existing issue for the compiled `path.relative()` validation path.

andrewtdiz:
[...]

Issue #214, opened 2026-04-26
String-array element access: arr[i] SIGSEGVs and indexOf returns -1


proggeramlug:
## Summary

On macOS aarch64 (host), Perry v0.5.319 (commit `2adf2483` on `main`) miscompiles element access on `string[]` arrays:

- `arr[i]` SIGSEGVs at address `0x10000000028` (looks like an int-tagged value being dereferenced).
- `arr.indexOf(x)` returns `-1` for elements that are present in the array.
- `JSON.stringify(arr)` returns the correct content, and `arr.length` is correct, so the array's structural metadata is fine — only element extraction is broken.
- `number[]` arrays are unaffected.

The bug reproduces regardless of how the string array is constructed (`JSON.parse('[\"...\"]')`, array literal `[\"...\"]`, or `push` to an empty array).

## Reproducer

```typescript
// repro.ts
const nums: number[] = JSON.parse('[1, 2, 3]')
console.log(\"nums.length=\" + nums.length)
console.log(\"nums[0]=\" + nums[0])               // 1
console.log(\"nums.indexOf(2)=\" + nums.indexOf(2)) // 1

const strs: string[] = JSON.parse('[\"hello\", \"world\"]')
console.log(\"strs.length=\" + strs.length)               // 2 (correct)
console.log(\"JSON.stringify(strs)=\" + JSON.stringify(strs)) // [\"hello\",\"world\"] (correct)

console.log(\"strs.indexOf('hello')=\" + strs.indexOf(\"hello\"))
// expected:  0
// actual:   -1   <-- BUG #1: indexOf returns -1 for present element

console.log(\"about to access strs[0]\")
const e = strs[0]
// expected: \"hello\"
// actual:    SIGSEGV  <-- BUG #2: indexed access on string array crashes
console.log(\"strs[0]=\" + e)
```

## Build / run

\`\`\`
cargo run --release --bin perry -- compile repro.ts -o repro
./repro
\`\`\`

## Actual output

\`\`\`
nums.length=3
nums[0]=1
nums.indexOf(2)=1
strs.length=2
JSON.stringify(strs)=[\"hello\",\"world\"]
strs.indexOf('hello')=-1
about to access strs[0]
zsh: segmentation fault  ./repro
\`\`\`

## Expected output

\`\`\`
nums.length=3
nums[0]=1
nums.indexOf(2)=1
strs.length=2
JSON.stringify(strs)=[\"hello\",\"world\"]
strs.indexOf('hello')=0
[...]

PR #7626, opened 2026-08-08
fix(path): #7621 — path.* arms read an SSO string's inline bytes as a StringHeader pointer

### LINKED ISSUE #7621: path.resolve/extname/normalize/matchesGlob read an SSO string's inline bytes as a StringHeader pointer
`path.resolve(base, computedShortString)` throws `ERR_INVALID_ARG_TYPE` where node returns the path.

## Repro

```ts
import * as path from "node:path";
function seg(n: number): string { return "s" + String(n); }

console.log(path.resolve("/root", "s1"));       // node: /root/s1   perry: /root/s1
console.log(path.resolve("/root", seg(1)));     // node: /root/s1   perry: TypeError
```

```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string.
```

Exit code 1. The two-argument and the three-argument forms both fail; only literal segments work.

## It is the SSO inline representation, not GC staleness

Bisected by string length — a computed segment **longer** than the SSO inline threshold works, a short one does not:

```ts
function longSeg(n: number): string { return "segment-that-is-definitely-longer-than-sso-" + String(n); }
path.resolve("/root", longSeg(1));   // OK, /root/segment-that-is-...-1
path.resolve("/root", seg(1));       // TypeError
```

`path.resolve(a, b, c)` folds to `PathResolveJoin(PathResolveJoin(a, b), c)` (HIR), and `Expr::PathResolveJoin`'s lowering in `crates/perry-codegen/src/expr/arrays_finds.rs` unboxes both operands with `unbox_to_i64`:

```rust
let a_handle = unbox_to_i64(blk, &a_box);
let b_handle = unbox_to_i64(blk, &b_box);
let result = blk.call(I64, "js_path_resolve_join", &[(I64, &a_handle), (I64, &b_handle)]);
```

`unbox_to_i64` masks the low 48 bits. For an interned literal that is a real `StringHeader*`; for an SSO string those bits are the **inline characters**, so `js_path_resolve_join` dereferences character data as a header. This is the #214 SSO class — the same hazard `Expr::EnvGetDynamic` two hundred lines away already documents and routes around with `unbox_str_handle`.

Same shape, unaudited, in the neighbouring arms: `PathExtname`, `PathNormalize`, `PathResolve`, `PathMatchesGlob`.

## Why it is not a one-line fix

`unbox_str_handle` calls `js_get_string_pointer_unified`, whose SSO branch **allocates** (it materialises the inline string onto the heap). Substituting it into `PathResolveJoin` therefore introduces a collection point between the two operand unboxes — the #7213 window — so the fix has to arrive with the rooting strategy for that window, not on its own.

Found while migrating `expr/arrays_finds.rs` for the Layer 1 rooting campaign (#7620 / #7615), and deliberately left there: that PR is IR-identical-or-justified by construction and a semantic fix to five path arms does not belong inside a refactor. Confirmed present on both arms of that PR, i.e. on `main`.

proggeramlug:
Closes #7621.

## Root cause

`crates/perry-codegen/src/expr/arrays_finds.rs:604` (and eleven siblings)
unboxed a `path.*` operand with `unbox_to_i64` — `bitcast double -> i64; and
POINTER_MASK` — and handed the low 48 bits to a runtime entry that dereferences
them as `*const StringHeader`.

That is the header for a **heap** string (`STRING_TAG` = 0x7FFF, payload = the
pointer) and the **characters** for a small-string-optimized one
(`SHORT_STRING_TAG` = 0x7FF9, payload = length + up to `SHORT_STRING_MAX_LEN` = 5
inline bytes, `crates/perry-runtime/src/value/tags.rs:96`). A literal is interned
onto the heap, so `path.resolve("/root", "s1")` worked; a *computed* short string
takes the inline form, so `path.resolve("/root", seg(1))` threw
`ERR_INVALID_ARG_TYPE`. The #214 class.

Bisected by length on a pre-fix build — the boundary is exactly
`SHORT_STRING_MAX_LEN`:

```
resolve len=2  THROW      resolve len=6  "/root/seeeee"
resolve len=3  THROW      resolve len=7  "/root/seeeeee"
resolve len=4  THROW      resolve len=8  "/root/seeeeeee"
resolve len=5  THROW
```

## Scope: twelve arms, not the five the issue named

Every arm below was confirmed broken on a pre-fix build with a computed 2-byte
operand. Two of them were **silently wrong** rather than throwing, which is why
they had never been noticed:

| arm | pre-fix | node |
|---|---|---|
| `resolve(a, b)`, `resolve(p)`, `join`, `win32.join`, `normalize`, `extname`, `dirname`, `basename`, `basename(p, ext)`, `isAbsolute` | `THROW ERR_INVALID_ARG_TYPE` | the value |
| `parse(p).base` | `""` | `"se"` |
| `matchesGlob(p, pat)` | matched against `""` | matches against `p` |

`path.relative`, `path.format` and `path.toNamespacedPath` were already fine —
they take NaN-boxed values (`js_path_relative_checked`, #2995), which is the
precedent this PR generalises.

## The fix

**Single-operand arms** call the new `js_path_arg_header(value) -> i64`. It
materialises **only** the SSO case and otherwise reproduces the old mask bit for
bit — heap strings *and every non-string*. That asymmetry is the point: each
entry point keeps its own established non-string behaviour (`js_path_join`
throws, `js_path_matches_glob` defaults to `""`) unexamined and unchanged, so the
blast radius is provably the SSO case alone.

It is deliberately **not** `js_get_string_pointer_unified` (which
`unbox_str_handle` wraps). That helper coerces non-strings — a number comes back
as `"5"` — which would silently turn `path.isAbsolute(5)` from Node's
`ERR_INVALID_ARG_TYPE` throw into `false`.

**Two-operand arms** hand both operands to the runtime NaN-boxed
(`js_path_*_value`). Codegen cannot close that window itself, which is the
"why it is not a one-line fix" the issue flagged: materialising the first operand
allocates, and `rooting::with_operands_rooted` yields the lowering *registers*,
not slots, so the second operand's register is stale the instant the first is
materialised and there is no re-read to reach for (`RootedSlot` has no `read`, by
design — see the note in `crates/perry-codegen/src/rooting.rs:334`). Doing both
unboxes inside one runtime call puts the window somewhere it can be closed
properly: `crates/perry-runtime/src/path/value_args.rs` roots the first operand in
a `RuntimeHandleScope` and re-reads it through `RuntimeHandle::across_const`,
never binding the pre-collection address.

A non-string operand is a masked garbage address, so it is deliberately *not*
rooted — handing the collector something to mark that is not an object would be a
worse bug than the one being fixed.

## Did the fix need rooting discipline? Yes — and here is its honest status

The SSO half is sabotage-proven in both directions. The rooting half is
**defensive, not instrument-proven**, and the module docs say so at the site:

* Disabling the SSO branch (`if false && ...is_short_string()`) turns
  `sso_operand_resolves_to_a_real_header` red and crashes
  `both_operands_survive_the_materialisation_window`; the gap test threw at line
  2 with exit 1 on the pre-fix binary.
* Reverting `across_const` to a pre-bound `let a_ptr = a_ptr0;` produced **zero**
  faults: byte-identical on the gap test under `PERRY_GC_ZEAL=1` +
  `PERRY_GC_PROTECT_FROMSPACE=1` (402,214 copying minors observed via
  `PERRY_GC_DIAG`), and 0/400,000 mismatches on an SSO-pair `path.join` hammer
  under `PERRY_GC_FORCE_EVACUATE=1` + from-space protection.

  That is not a fluke, it is structural: the window is allocation-to-allocation,
  and a collection reached from inside an allocation runs with
  `GC_FLAG_IN_ALLOC` set, which makes the copying minor ineligible. Nothing
  *moves* at an allocation point today, so no available instrument can turn that
  window red. The ordering is written the correct way because it is the shape the
  invariant asks for and costs nothing — not because a probe caught it.

## Validation (local; CI backlog is deep, so this is the evidence)

* `test-files/test_gap_7621_path_sso_operands.ts` — byte-identical to
  `node --experimental-strip-types` on the pinned 26.5.1, exit 0 both sides.
  Covers both sides of the SSO boundary (lengths 2..10), computed *and* literal
  operands, absolute *and* relative bases (asserted cwd-independently),
  multi-segment and absolute-reset resolves, all twelve arms, the non-string
  throw, and a nursery-churn loop around the two-operand window.
* Same test byte-identical under `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1
  PERRY_GC_PROTECT_FROMSPACE_DEPTH=800`, compiled and run with
  `PERRY_GC_MOVING_LOOP_POLLS=1`; the instrument was asserted live
  (`[gc-fromspace-protect] retired_set=#N`, 402,214 sets).
* No regressions: `test_parity_path`, `test_gap_node_path`,
  `test_gap_6371_path_lexical_semantics`, `test_gap_node_fs`,
  `test_cli_simulation` all byte-identical to node.
* 5 new `perry-runtime` unit tests; `cargo test -p perry-runtime --no-fail-fast`
  1890 passed / 0 failed, `cargo test -p perry-codegen` 690 passed / 0 failed.
* Lint: `cargo fmt --all -- --check`, `check_file_size.sh`,
  `addr_class_inventory.py`, `raw_handle_debt.py` (998, unchanged — the new code
  uses `across_const`, not a bare `get_raw_*_ptr`), `class_id_collisions.py`,
  `check_test_registration.py`, `gc_store_site_inventory.py`,
  `workspace_architecture.py` all pass.

One pre-existing failure is untouched by this PR:
`perry-codegen`'s `large_local_array_push_inbounds_store_emits_precise_slot_barrier`
fails identically on pristine `origin/main` (A/B'd in this worktree).

---

## Rebased onto v0.5.1357 (#7627) — semantic conflict resolved

#7627 migrated `expr/instance_misc1.rs` onto the Layer 1 rooting API while this
was open. Rebasing gave **4 conflicts, all in that one file**; every other file
applied cleanly. The migrated form is the base and this fix layers on top:

* **Two-operand arms** (`PathJoin`, `PathWin32Join`, `PathBasenameExt`, and the
  win32 `BasenameExt` / `ResolveJoin` / `MatchesGlob` methods) keep their
  `with_operands_rooted` wrapper and now emit **no unbox at all** — both operands
  go to the runtime NaN-boxed. That is strictly stronger than the migrated base:
  it *deletes* the window from codegen rather than protecting it, so there is no
  raw operand register in the arm for a collection to invalidate.
* **Single-operand arms** emit `js_path_arg_header` **inside** the rooted region,
  immediately followed by its consumer.

**On `with_operands_rooted_across_call`** (added by #7627 for exactly this
shape): I read its doc and deliberately did **not** use it. Its own closing
paragraph draws the line — *"Use it only when the emitted step can re-enter user
code or enumerate an arbitrary object's own properties. For a helper that merely
allocates, the project's position (#7198) is that it cannot initiate a moving
collection, so a root there would be pure cost."* `js_path_arg_header` only
materialises SSO bytes; it dispatches nothing and enumerates nothing. That is
also the same fact I measured independently before the rebase (a collection
reached from inside an allocation runs with `GC_FLAG_IN_ALLOC`, so the copying
minor is ineligible), which is why the rooting sabotage arm could not be made to
fault. Two independent routes to the same conclusion.

And after the resolution the question is moot for the two-operand arms anyway:
they emit zero `js_path_arg_header` calls, so there is no codegen-side window to
state.

### Re-verified after the rebase, not before

* SSO bisect **re-confirmed against the new base**: built `origin/main`
  (v0.5.1357) in this worktree — lengths 2–5 throw, 6+ work, and `parse`/
  `matchesGlob` are still silently wrong. #7627 did not change the bug (it is
  IR-identical by construction).
* Gap test byte-identical to node 26.5.1, exit 0; byte-identical again under
  `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1` with **402,214** copying minors
  asserted live via `PERRY_GC_DIAG`.
* `test_parity_path`, `test_gap_node_path`, `test_gap_6371_path_lexical_semantics`,
  `test_gap_node_fs`, `test_cli_simulation` all byte-identical.
* **Root-dominance corpus, both gated modes, on the rebased build** — 129/129
  sources compiled, 149 `.ll`:
  * dominance: 2452 functions / 149 modules / 9846 root stores, **0 violations**,
    and **40/40 seeded violations caught** (the checker asserted able to fail);
  * `--unrooted-allocas`: 7867 gc-capable allocas, **0 violations**.

  Both above the CI liveness floors (`--min-files 90 --min-binds 1500
  --min-funcs 1200`), allowlist still empty.
* **Ledger**: `instance_misc1.rs` is in `MIGRATED_MODULES` now and nothing added
  here names `expr::temp_root` —
  `migrated_modules_do_not_reach_past_the_rooting_api` green, alongside its
  planted-violation and clears-the-migrated-form siblings.
* `cargo test -p perry-codegen --lib` 694 passed / 0 failed;
  `-p perry-runtime --no-fail-fast` 1890 passed / 0 failed.
* Full lint set from `.github/workflows/test.yml`: `cargo fmt --all -- --check`,
  `check_file_size.sh`, `addr_class_inventory.py` (+`--self-test`),
  `raw_handle_debt.py` (+`--self-test`, **998**, unchanged),
  `class_id_collisions.py`, `check_test_registration.py` (+`--self-test`),
  `gc_store_site_inventory.py` (+`--self-test`), `workspace_architecture.py`,
  `gap_snapshot.py --self-test`, `gc_gate_wiring_check.py` — all green.

Conflict-free against #7631 (`git merge-tree` of the two heads exits 0), and
#7626 adds no `perry_hir::Class` literal, so #7631's struct widening cannot break
it in either merge order.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1357
+**Current Version:** 0.5.1358
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1357"
+version = "0.5.1358"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "perry-ffi",
  "scraper",
@@ -5813,7 +5813,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-commander"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5821,7 +5821,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cron"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "chrono",
  "cron",
@@ -5831,7 +5831,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dayjs"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "chrono",
  "perry-ffi",
@@ -5839,7 +5839,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-decimal"
-version = "0.5.1357"
+version = "0.5.1358"
 dependencies = [
  "perry-ffi",
  "rust_decimal",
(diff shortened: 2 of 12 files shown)


proggeramlug:
## Audit before merge — verified, merged as v0.5.1358

**Reproduced the bug on main and the fix on this branch**, bisecting the SSO
boundary with computed strings:

```
main (pre-fix):   2 parse.base: ""      <- silently WRONG, node says "aa"
                  3 parse.base: "aaa"   <- correct
this PR:          identical to node 26.5.1 across lengths 2..7
```

That silent-empty case is the reason this went unreported for so long — it
never threw, it just returned the wrong string. Twelve arms rather than the
five filed.

**The resolution of the #7627 collision is better than what I asked for.** I
said "layer the fix inside the rooted region"; for the two-operand arms you
instead made them emit **no unbox at all** — both operands travel NaN-boxed to
one runtime call. That deletes the window from codegen rather than protecting
it, which is strictly stronger: there is no raw operand register in the arm to
be stale.

**And declining `with_operands_rooted_across_call` was right.** Its own doc
draws the line — use it when the emitted step can re-enter user code or
enumerate arbitrary own properties; `js_path_arg_header` only materialises SSO
bytes and dispatches nothing. Two independent routes to that conclusion (the
doc's rule, and your measured `GC_FLAG_IN_ALLOC` finding that made the rooting
sabotage unfaultable) is the right amount of evidence for a *not*-doing-it
decision.

Gates re-run here: runtime 1,890/0, codegen 694/0, all six lint scripts +
file-size + fmt clean.

Also worth keeping: your post-rebase sweep initially showed pre-fix behaviour
because it ran against a binary the background build was still writing. Catching
that rather than filing a phantom regression is the same class of care as the
`$?`-after-pipe and login-shell findings elsewhere today — and it is worth
noting that #7625 removed the *other* reason to re-run, so this one is now the
main reason left.

PR #7631, opened 2026-08-08
fix(instanceof): #7575 — a monomorphized generic class is an instance of the generic it came from

### LINKED ISSUE #7575: `instanceof` a Map/Set SUBCLASS is false (`m instanceof MyMap`); only the native base edge survives
`m instanceof MyMap` is `false` for a `class MyMap extends Map` instance.
`m instanceof Map` is correct. Found while fixing #7570 (PR #7573); it is a
**different** bug and survives that fix — it reproduces with and without the
base-type annotation, so it is not a declared-type-lowering problem.

## Repro

```ts
class MyMap<K, V> extends Map<K, V> {}
class Plain {}
class SubPlain extends Plain {}

const a = new MyMap<string, number>();
console.log("unannotated:", a instanceof MyMap, a instanceof Map);

const b: Map<string, number> = new MyMap<string, number>();
console.log("annotated:", b instanceof MyMap, b instanceof Map);

// control: an ordinary class hierarchy is fine
const c: Plain = new SubPlain();
console.log("plain annotated:", c instanceof SubPlain, c instanceof Plain);
```

```
$ node --experimental-strip-types inst.ts
unannotated: true true
annotated: true true
plain annotated: true true

$ ./inst
unannotated: false true          <-- both wrong
annotated: false true
plain annotated: true true       <-- ordinary classes are fine
```

So the intermediate/leaf class edge is what is lost, not the base edge: the
chain walk finds `Map` but not `MyMap`. The same holds for `class MySet extends
Set`.

## Notes for whoever picks this up

- Ordinary (non-native-base) inheritance is unaffected, so this is specific to
  the `super()`-to-a-native-base wiring, not to `instanceof` in general.
  CLAUDE.md's *Known-weak areas* entry on native base-class subclassing is the
  right place to start: "a native base's surface is installed at `super()` time
  and its parent edge lives in the class registry".
- The existing `test_gap_6325_map_set_subclass.ts` only asserts
  `m instanceof Map`, which is why this was never caught.
- `test_gap_7570_map_set_declared_base_type.ts` deliberately asserts only
  `instanceof Map` / `instanceof Set`, with an inline comment pointing here.
  Whoever fixes this should tighten those two lines to also assert
  `instanceof MyMap` / `instanceof MySet`.

proggeramlug:
Closes #7575.

## Root cause: it is not Map/Set, and it is not the prototype chain

The issue reads `m instanceof MyMap === false` as a native-base-subclassing /
class-registry-parentage defect, and points at CLAUDE.md's *Native base-class
subclassing* and *Two prototype-resolution paths* entries. Both are the wrong
tree. The mechanism is **monomorphization**.

Perry specializes generic classes. `class Gen<T> {}` plus `new Gen<number>()`
emits a SECOND class named `Gen$num`
(`crates/perry-hir/src/monomorph/mangle.rs:14`, `specialize.rs:50`) carrying its
own class id, and the instance is stamped with **that** id — while
`x instanceof Gen` resolves the RHS through `ctx.class_ids` to the **generic's**
id (`crates/perry-codegen/src/expr/instance_misc1.rs:346`), which appears nowhere
in the specialization's parent chain. `js_instanceof`'s walk therefore answers
`false` for the class the user actually wrote, and `true` for the base — exactly
the "only the native base edge survives" symptom.

The bisect that identifies it, measured on pristine `main`:

```
class GenMap<K,V> extends Map<K,V> {}  new GenMap<string,number>()  false true
class ConcMap     extends Map<...>  {}  new ConcMap()                true  true
class BareMap     extends Map       {}  new BareMap()                true  true
class GenMap<K,V> extends Map<K,V> {}  new GenMap()   /* no args */  true  true
class GenPlain<T> extends PlainBase {}  new GenPlain<number>()       false true   <-- no Map anywhere
class GenNoBase<T>                  {}  new GenNoBase<number>()      false        <-- no BASE anywhere
```

The last two rows settle it: a generic class with an ordinary base, and one with
no base at all, fail identically. `class MyMap<K, V> extends Map<K, V>` is simply
the idiomatic spelling, which is why it surfaced there. `constructor.name` on the
same instances reports `Gen$num` — the same leak on a different surface.

## The fix

* HIR records `Class::specialized_from` (the generic's name), set by
  `specialize_class`.
* Codegen emits one `js_register_class_generic_origin(spec, generic)` per
  specialization in the module-init prelude, next to the existing
  `js_register_class_parent` edges.
* `instanceof`'s chain walk — now a single shared, depth-bounded
  `class_chain_reaches`, used by both the static and the dynamic-RHS path, which
  previously had two hand-rolled copies (one of them uncapped) — follows the
  origin edge as well as `extends`.

It is deliberately a **separate edge, not a parent edge**. `CLASS_REGISTRY`'s
chain also resolves `super()` construction
(`object/class_constructors.rs:652`), static-method lookup and vtable dispatch,
so splicing `Gen` in between `Gen$num` and its real base would re-run the wrong
constructor. Only `instanceof` may follow this one, and the runtime module says
so at the declaration.

## Did the fix need rooting discipline?

No. The whole change is `u32 -> u32` class-id bookkeeping — no heap pointers, no
new cache of a `*mut`, so nothing to register with
`gc_register_mutable_root_scanner` and nothing for `raw_handle_debt.py` to count
(it is unchanged at 998). The gap test is nevertheless byte-identical under
`PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1`.

## The Array-side sibling #7603 left unfixed

One mechanism covers both families, **and the note about it was stale**. Measured
on pristine `main`:

```
new MyArr()                  instanceof MyArr    true    (already worked)
new Indirect()               instanceof MyArr    true    (already worked)
MyArr.from([1,2,3])          instanceof MyArr    true    (already worked)
new GenArr<number>()         instanceof GenArr   FALSE   <-- fixed here
```

So the only broken Array case was the generic spelling, and this PR fixes it. The
comment in `test_gap_7541_array_subclass_inherited_statics.ts` that named
`sub instanceof MyArr` as a pre-existing gap was wrong; it is corrected in place
rather than left to mislead the next reader. Nothing is left to file on the
Array side.

## Sabotage, both ways

* Making `class_generic_origin` return `None` turns two of the four new runtime
  unit tests red and turns the gap test red at 8 lines
  (`1 unannotated: false true`, `6 seeded: false true`, `8 dynamic`, `9/10 ...`).
* The other two unit tests stay green under that sabotage **by design** — they
  assert the edge is directional and does not widen matches (sibling
  specializations must not match each other, a generic is not an instance of its
  own specialization), so they would catch the opposite mistake.

## Validation (local; CI backlog is deep, so this is the evidence)

* `test-files/test_gap_7575_map_set_subclass_instanceof.ts` — byte-identical to
  `node --experimental-strip-types` on the pinned 26.5.1, exit 0. Covers
  `instanceof` against the subclass, the native base, an unrelated class, a
  multi-level chain (`class A extends Map {}; class B extends A {}; class C
  extends B {}`), an explicit-`super()` subclass and a subclass of it, an
  iterable-seeded instance, `Symbol.hasInstance` in both directions (it still
  takes precedence over the chain walk), the dynamic RHS, `instanceof` over an
  untyped parameter, and the generic-over-plain / generic-over-nothing /
  generic-over-Array shapes plus their non-generic controls and sibling
  negatives.
* Same test byte-identical under `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1
  PERRY_GC_PROTECT_FROMSPACE_DEPTH=800`, compiled with
  `PERRY_GC_MOVING_LOOP_POLLS=1`.
* `test_gap_6325_map_set_subclass.ts` and
  `test_gap_7570_map_set_declared_base_type.ts` **tightened** to assert the
  subclass edge, as #7575 asked — both still byte-identical.
* No regressions: `test_gap_7541`, `test_gap_7574`, `test_gap_7563`,
  `test_gap_6232`, `test_gap_4099`, `test_gap_5592`, `test_edge_generics`,
  `test_generic_class`, `test_edge_class_advanced`, `test_edge_complex_patterns`,
  `test_edge_interfaces`, `test_data_pipeline`,
  `test_gap_repsel_element_shape_loop_clone`,
  `test_gap_intl_rtf_auto_instanceof_6960` all byte-identical to node.
* 4 new `perry-runtime` unit tests over the walk. `cargo test -p perry-runtime`
  1890 passed / 0 failed; `-p perry-hir` and `-p perry-transform` all green.
* Lint: `cargo fmt --all -- --check`, `check_file_size.sh`,
  `addr_class_inventory.py`, `raw_handle_debt.py` (998, unchanged),
  `class_id_collisions.py`, `check_test_registration.py`,
  `gc_store_site_inventory.py`, `workspace_architecture.py`,
  `gap_snapshot.py --self-test` all pass.

Pre-existing failures untouched by this PR, each A/B'd against pristine
`origin/main` in this worktree:

* `perry-codegen`'s integration suites (`crates/perry-codegen/tests/*.rs`, which
  do not run per-PR) fail **the same 23 tests, byte-identical list**, on both
  arms.
* `test_harness_class_mixins`, `test_issue_562_stream_subclass` and
  `test_issue_806_curried_factory_extends_capture` produce **byte-identical Perry
  output** on both arms.
* `cargo clippy` errors in `crates/perry-ffi/src/jsvalue.rs` (`approx_constant`
  on `3.14`) are pre-existing in a file this PR does not touch.

## Known, deliberately not folded in

`constructor.name` still reports the mangled `Gen$num` rather than `Gen`. Same
root cause (monomorphized identity leaking to a user-visible surface), different
surface (the class display-name registry), and it can move error-message text —
so it belongs in its own change with its own parity sweep, not bundled into an
`instanceof` fix. Filed as #7632.

---

## Rebased onto v0.5.1357 (#7627) — clean as text, but it did NOT compile

Rebasing this onto current main produced **zero conflicts**, and
`git merge-tree origin/main <this>` exits 0. That verdict was wrong in the way
that matters: **the merge does not build.**

#7627 added a new `perry_hir::Class` literal in
`crates/perry-codegen/src/codegen/emission_order_tests.rs`, and this PR *widens*
that struct with `Class::specialized_from`. Neither side touches a line the
other side touches, so there is nothing for a textual merge to flag — but the
result is `E0063: missing field specialized_from`. Fixed here in its own commit.

Worth stating plainly because it generalises: **textual mergeability is not a
merge check when one branch widens a struct another branch constructs.** The
only reliable check is `cargo check --all-targets` on the merge result, and
`--all-targets` is load-bearing — the offending literal is in a `#[cfg(test)]`
fixture, so a plain `cargo check` stays green and the break surfaces later, in a
job that does build tests.

This PR does not touch `expr/instance_misc1.rs`, `logical_collections.rs` or
`map_set.rs`, so it has no overlap with #7627's rooting migration itself.

### Re-verified after the rebase, not before

* `test_gap_7575_map_set_subclass_instanceof.ts` byte-identical to node 26.5.1,
  exit 0; byte-identical again under `PERRY_GC_ZEAL=1
  PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800`, compiled
  with `PERRY_GC_MOVING_LOOP_POLLS=1`.
* `test_gap_7570`, `test_gap_6325` (both tightened by this PR) and
  `test_gap_7541` all byte-identical.
* `cargo check --all-targets` over the workspace: clean.
* 4 runtime unit tests over the walk green; `-p perry-runtime` 1890 passed / 0
  failed; `-p perry-codegen --lib` 694 / 0 (ledger tests included);
  `-p perry-hir` 281 / 0; `-p perry-transform` 56 / 0.
* **Root-dominance corpus, both gated modes** (this PR changes codegen emission,
  so it was re-run rather than reasoned about): 129/129 sources, 149 `.ll`,
  2452 functions / 9846 root stores, **0 violations** in dominance mode with
  **40/40 seeded caught**, and **0** `--unrooted-allocas` violations.
* Full lint set from `.github/workflows/test.yml` green, including
  `raw_handle_debt.py` at **998** (unchanged — this PR is `u32 → u32`
  bookkeeping, no heap pointers).

Conflict-free against #7626 (`git merge-tree` of the two heads exits 0), and
#7626 adds no `Class` literal, so this PR's struct widening cannot break it in
either merge order.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `instanceof` checks for monomorphized generic classes and their inheritance chains.
  * Improved behavior for generic `Map`, `Set`, and other specialized subclasses, including dynamic constructor checks.
  * Preserved correct behavior for negative matches, sibling specializations, and custom `Symbol.hasInstance` implementations.

* **Tests**
  * Added comprehensive regression coverage for generic inheritance and built-in collection subclasses.

* **Chores**
  * Updated the application version to 0.5.1359.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1358
+**Current Version:** 0.5.1359
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1358"
+version = "0.5.1359"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "perry-ffi",
  "scraper",
@@ -5813,7 +5813,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-commander"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5821,7 +5821,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cron"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "chrono",
  "cron",
@@ -5831,7 +5831,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dayjs"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "chrono",
  "perry-ffi",
@@ -5839,7 +5839,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-decimal"
-version = "0.5.1358"
+version = "0.5.1359"
 dependencies = [
  "perry-ffi",
  "rust_decimal",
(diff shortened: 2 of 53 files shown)


proggeramlug:
## Audit before merge — verified, merged as v0.5.1359

**The corrected diagnosis is right, and I reproduced the bisect that proves
it** — the issue's filed explanation (native-base wiring / the
`CLASS_PROTOTYPE_OBJECTS` split) was wrong, and my brief repeated it:

| case | main | this PR | node |
|---|---|---|---|
| `MyMap<K,V> extends Map` instanceof itself | **false** | true | true |
| `GenPlain<T>` — no Map, no base at all | **false** | true | true |
| `GenExt<T> extends Base` | **false** | true | true |
| non-generic `Bee extends A extends Map` | true | true | true |
| negatives (sibling specializations) | false | false | false |

A generic class with no Map anywhere fails identically; a non-generic Map chain
works. That settles it as monomorphization — `new MyMap<string,number>()`
constructs `MyMap$str_num` while `instanceof MyMap` resolves the generic's id,
which is in no parent chain. Map/Set was just the idiomatic spelling in the bug
report.

**Following the origin edge instead of adding a parent edge is the load-bearing
design choice**, and the reason is stated: `CLASS_REGISTRY`'s chain also
resolves `super()`, static lookup and vtable dispatch, so a parent edge would
have changed three unrelated behaviours to fix one. Collapsing two hand-rolled
chain walks — one of them uncapped — into a single depth-bounded
`class_chain_reaches` is the right cleanup to take along.

Gates: runtime 1,894/0, codegen 694/0, hir 281/0, all five lint scripts +
file-size + fmt clean, `cargo check --all-targets` clean.

## The merge-order lesson here is worth generalising

This branch was **textually conflict-free and still did not compile**: #7627
added a `perry_hir::Class` literal in a `#[cfg(test)]` fixture and this PR
*widens* that struct, so neither side touched the other's lines and
`git merge-tree` exited 0 — then `E0063: missing field specialized_from`.

Two things follow, and I'd like both treated as standing practice for the
remaining ~85 Layer-1 slices, where struct widening and fixture-heavy modules
will keep meeting:

1. **`git merge-tree` exiting 0 is not a merge check** when one branch widens a
   type another constructs.
2. **`cargo check --all-targets` is load-bearing**, because a plain `cargo
   check` stays green when the only constructor is in a test fixture — the
   break then surfaces in a later job rather than at the merge.

PR #7624, opened 2026-08-08
perf(gc): defer old-object page registration off the promote path (from #7623)

proggeramlug:
Extracted from #7623 per its audit. That PR's static-pretenure half was a
measurement confound and is not merging; this is the finding buried inside it
that stands on its own — and unlike the pretenure half, it pays on **current
main**, with no new codegen and no new allocator policy.

## The cost

`register_old_object_pages` is written for the occasional old-gen birth it was
introduced for. Per object it pays:

- two `RefCell` borrows (`OLD_GEN_PAGE_OBJECTS`, then `OLD_GEN_PAGE_META`),
- two `Vec` allocations (`old_object_page_overlaps`, `added_pages`),
- a hash lookup, and
- **a linear `contains` scan of that page's object list**, which grows as the
  page fills — quadratic in the objects a burst lands on one 4 KiB page.

Since **#7613's promote-on-first-copy** that is no longer an occasional path.
A copying minor promotes straight into old-gen (`gc/copying.rs`'s `move_young`
→ `arena_alloc_gc_old`), and `PERRY_GC_DIAG` puts a number on it — json_pipeline
200k promotes **1,657,956 objects / 113,226,480 bytes in a single minor**, every
one through that function. That is the same ~113 MB the #7623 audit measured as
the moved cohort, now attributed to the code path that bills for it.

## The change

`arena_alloc_gc_old` records `(header_addr, total_size)` in a thread-local
buffer; one batched flush folds the whole burst in. The flush holds a single
borrow of each table, allocates no per-object `Vec`, and — the part that
removes the quadratic term — scans only the portion of a page's object list
that **predates the batch**. A bump-allocated promotion burst fills fresh
pages, where that prefix is empty and the dedup scan disappears.

Skipping the in-batch entries is sound because they are pairwise distinct: each
comes from a live allocation, and an address cannot be handed out twice without
an intervening free, which cannot happen without a flush. Hole reuse — the
reason the dedup exists at all — hands back an address registered *before* the
batch, so it is still covered (`batched_flush_matches_eager_registration` pins
exactly that case).

**Allocation policy is deliberately unchanged.** #7623 also dropped the
`old_free_take_exact` hole probe on its pretenure allocator; that is a policy
change with its own RSS consequences and it is not here. This PR is bookkeeping
only — which is why the GC census can be, and is, asserted byte-identical.

## Soundness: one rule, and removers matter as much as readers

> Every reader **and every remover** of `OLD_GEN_PAGE_OBJECTS` /
[...]

PR #7583, opened 2026-08-07
fix(runtime): give iterator helpers their own class id (#7576)

proggeramlug:
Closes #7576.

## Root cause: a class-id collision

`crates/perry-runtime/src/iterator_helpers.rs:45` and
`crates/perry-runtime/src/string/iter_object.rs:25` both defined `0xFFFF_0009`:

```rust
pub const ITERATOR_HELPER_CLASS_ID: u32 = 0xFFFF_0009;   // "sits just past the Set iterator id (0xFFFF0008)"
pub const STRING_ITERATOR_CLASS_ID:  u32 = 0xFFFF_0009;   // "sits just past the Set iterator id (0xFFFF0008)"
```

Two constants, introduced independently, carrying the same comment. Neither
author checked whether the slot was taken.

Every dispatch tower matches these ids in a fixed order, and the String arm
comes first — `object/native_call_method/handle_methods.rs:838` and
`.../collection_methods.rs:396`, both ahead of the helper arm at `:855`/`:410`.
So every iterator-helper object was dispatched as a String iterator:

| call | what actually ran | result |
|---|---|---|
| `.next()` | `dispatch_string_iterator_method`, reading the helper's **op-kind** field as a cursor index against a **null** backing array | `{ done: true }` on the first step, for every source kind |
| `.map` / `.filter` / `.take` / `.drop` / `.flatMap` / `.toArray` / `.reduce` / `.forEach` / `.some` / `.every` / `.find` | the same dispatcher's `_ => undefined` arm | `undefined`, so the next link in the chain threw |

That is both symptoms in the issue from one cause, and it explains why row C
(the bare `{next}` driven directly) worked: it never went near the helper.

The helper now takes `0xFFFF_000B`. `0xFFFF_000A` is the RegExp-string iterator.

## Second, independent defect on the same path

`iterator_step` resolved the source iterator's `next` with the **inheriting**
getter (`js_object_get_field_by_name`). Since #321 every built-in iterator
inherits `.next` from its shared `%…IteratorPrototype%` singleton
(`object/iterator_prototypes.rs`), and that inherited `next` is a **thunk**
that resolves its receiver from `js_implicit_this_get()` and dispatches by
class id.

So for an array / Map / Set / String iterator source the lookup found a
perfectly callable closure, `iterator_step` took the raw-closure-call branch —
`js_closure_call1`, which binds no `this` — and the thunk ran against whatever
was left in the thread-local. Observed directly: `Method
[...]

PR #7633, opened 2026-08-08
perf(gc): defer the JSON materialiser's per-slot layout notes to one finalize (#7630)

### LINKED ISSUE #7630: perf(gc): per-slot layout bookkeeping is the top cost family on json_pipeline — declare-at-birth for the JSON materialiser's cohort
Step-zero profile of `json_pipeline` 200k on current main (v0.5.1360-era), **pinned mini** (quiet, best-of-5, hash-stable): total ~1.3 s, `build_out` 877 ms of which **816 ms is GC pause** (census: 3 cycles — 1 minor promoting 101 MB with 0.0 MB survivor-copied, 2 fulls). The largest cost family in the symbolicated 3 s sample is now the **per-slot layout machinery**, ahead of the old-page/promote family that `perf/old-page-registration-deferral` is already extracting:

```
17  layout_note_slot                          (store-side, per field store)
10  layout_slot_visit::visit_gc_layout_slot_descriptors   (trace-side)
 7  layout_slot_visit::fixed_slot
 7  layout_only
 6  layout_transfer                           (per object, during promote/copy)
 5  layout_tables::layout_forget_object
——
52 samples — vs 36 for the old-page family, 15 for remembered-set inserts
```

This is the **trace/promote/store half** of the layout story — #7510/#7525/#7532 addressed the *construction* half (side-table emptiness, shape-declared-at-allocation for class ctors and literals). The payer here is the **JSON parse materialiser's cohort**: 200k records × ~13 slots (record + addr + tags) ⇒ millions of `runtime_store_jsvalue_slot` → `layout_note_slot` calls at materialisation, then per-object `layout_transfer` for 101 MB of promotions, then descriptor-driven slot visits at trace time.

The likely lever, mirroring the proven #7525/#7532 pattern: the materialiser knows the complete object layout **at materialisation time** (it builds each object from the tape in one pass, and every slot it writes is a NaN-boxed JSValue — no raw-f64 slots). Declaring `GC_LAYOUT_ALL_POINTERS` at birth (header bits only, no side table) and routing its stores through a note-free path would eliminate the per-store notes and collapse the trace visits to the all-pointers fast path — a pointer mask only ever pays for itself by skipping raw-f64 slots, which this cohort does not have.

Soundness note for whoever takes this: an elided store into a `POINTER_FREE`-published object is a stranded live child (the #7469 lesson) — the elision must be gated on the live header state exactly as `expr/array_push.rs`'s admission test is, or confined to a path where the materialiser owns the object end-to-end before publication.

proggeramlug:
Fixes #7630 — the top cost family from the step-zero profile that issue records.

## The cost

Parse-built records are born `POINTER_FREE`; the first string field builds a per-object side-table pointer mask, and from then on **every field store pays a hashmap round-trip** (`layout_note_slot`), `layout_transfer` moves the mask on every promotion, and `layout_forget_object` drops it at death. At 200k records × ~13 slots that machinery topped the profile (~52 samples, ahead of the old-page family the in-flight deferral PR targets).

The observation that makes it deletable: **a pointer mask can never skip anything for this cohort.** Every slot the materialiser writes is a NaN-boxed `JSValue`, and the tracer's tag check rejects non-pointers anyway — the mask machinery only ever pays for itself by skipping *raw-f64* slots, which parse objects do not have.

## The fix

The materialiser owns each object's whole construction, so its store loops use `runtime_store_jsvalue_slot_layout_deferred` — bit-for-bit the shared choke-point helper minus the layout note (typed-slot canonicalization, string addref demote, and the write barrier **with its SATB shade** all kept — the #7602 lesson) — returning the one fact the notes were computing (pointer-bearing). The loop accumulates it and settles the layout state **once** per object via `layout_finish_deferred_boxed_object`:

- **no pointer stored** → the `POINTER_FREE` birth state is still the truth, and it keeps its whole-payload trace skip (number-only records lose nothing);
- **any pointer stored** → `GC_LAYOUT_UNKNOWN`, the tag-checked scan-all state. Routed through `layout_mark_unknown`, not a bare state store, so a mask that the shaped path's by-name fallback DID create mid-construction (shape-overflow records) is removed rather than stranded.

The shaped path's finalize runs on the live pointer re-read from the parse root, so a mid-parse collection cannot leave it operating on a stale copy.

Also: `barrier.rs` crossed the 2000-line cap with the new helper; its slot-store helpers moved to `barrier_store.rs` (pure move).

## Measured — pinned mini, interleaved ×3, hash-identical every round

| json_pipeline 200k | main | this PR |
|--|--:|--:|
| build_out | ~934 ms | **~730 ms (−22 %)** |
| TOTAL | ~1,389 ms | **~1,156 ms (−17 %)** |
| GC pause (census) | 816 ms | **689 ms** |
| peak RSS | 451 MB | **422 MB** |

Census structure identical both arms (3 cycles, 0.0 MB survivor-copied, 101 MB promoted — the moved-bytes census is load-independent). Post-fix profile: `layout_note_slot`, `layout_forget_object`, `layout_only`, `fixed_slot` are **gone from the top**; the remaining leaders are the old-page/promote family (the deferral PR's target) and remembered-set inserts.

## Verification

- **GC ratchet, both arms back to back on one host**: every semantic counter identical across 144 medians on all 12 probes; the only differing semantic cell is `12_large_live_set.heap_used_bytes` at −504 bytes — the documented sample-dependent ungated cell, well inside its recorded 9,072-byte spread. All other movement is ungated RSS/wall ≤0.44 %.
- **GC zeal + fromspace-protect arm** (`PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1`): output identical — no stranded children under forced evacuation.
- **19/19 JSON test-files byte-identical to node**; `perry-runtime` suite 1,886 passed / 0 failed; `cargo fmt --check` and `check_file_size.sh` clean.
- **Gap suite** (full run on this branch, report `parity_report_20260808_101857.json`): **no real regressions.** The one reported regression, `test_gap_zlib_4917_level` `pass → compile_fail`, is the documented host-local runtime-dir resolution flake, disproven on this branch directly: with `PERRY_RUNTIME_DIR` pinned to the freshly built archives the test **compiles and matches node byte-for-byte**. The 10 `node_fail → parity_fail` status changes are oracle-coverage transitions (pre-existing gaps the pinned Node now reaches — identical set to the previous clean run on an unrelated branch), and `test_gap_iterator_helpers_2874` improved to pass.

## Summary by CodeRabbit

  * Improved JSON object materialization by deferring garbage-collection layout tracking until construction is complete.
  * Reduced processing overhead and improved benchmark performance during JSON parsing.
* **Bug Fixes**
  * Preserved correct handling of typed values, strings, references, and write barriers.
  * Ensured objects receive accurate garbage-collection layout information after construction.
* **Chores**
  * Updated the package version to 0.5.1361.


proggeramlug:
## Audit (soundness half) — draft respected, not merged. One finding you need before the gap suite lands.

**The design reasoning is right and I checked the parts that are checkable.** The
deferred store is materialiser-only — exactly two call sites
(`json/parser.rs:437`, `:688`) and two finalizes (`:486`, `:701`), same two
functions, and **zero early-exit paths between store and finalize** in either
(I walked the segments). So no parse path can leave an object with deferred
notes unsettled. Routing the pointer case through `layout_mark_unknown` rather
than a bare state store is the right call for the shape-overflow mask, and
re-reading `js_obj` from the parse root before finalize is the correct rooting
discipline.

Runtime suite 1,886/0; JSON gap family 7/7 byte-identical to node.

## The finding: I cannot make the unsafe branch fault, and neither does your zeal arm

I sabotaged `layout_finish_deferred_boxed_object(…, saw_pointer)` →
`(…, false)`, i.e. **every parse object claims pointer-free even when it holds
pointers**. That is precisely the stranded-live-child hazard #7630's own
soundness note names. Result:

| probe | plain | ZEAL+PROTECT | FORCE_EVACUATE |
|---|---|---|---|

I built the second probe specifically because the first was vacuous (it held
the children directly, so they survived regardless). 4,000 records × 2 pointer
fields, 40 rounds of churn to force promotion, children read back only after
the churn. **Sabotage still produces byte-identical correct output, 8 retired
quarantine sets and 16 copying minors observed live.**

So one of these is true, and the PR should say which:

1. Parse-built strings are longlived/interned and cannot be collected or moved
   — in which case the branch is safe but **the `POINTER_FREE` trace-skip is
   also buying nothing on this cohort**, and the win is entirely the removed
   per-store notes (which is fine, and is what your profile actually shows).
2. Something else re-marks these objects during construction, so the
   `saw_pointer=false` branch is not reached in practice for pointer-bearing
   records — in which case the accumulate-and-settle logic has a dead arm.
3. The hazard is real but needs a shape my probe doesn't produce.

**This matters for the PR's evidence, not its correctness**: your "GC zeal +
fromspace-protect arm: output identical — no stranded children" line reads as
verification, but a mutation that *should* strand children also produces
identical output, so that arm is not currently discriminating. Same vacuity
class as the ones caught in #7602's S4 and #7625's fixtures. Either find a
probe where the sabotage faults — that becomes the regression test worth
having — or state plainly that the branch's safety rests on the
materialiser-owns-construction argument rather than on an executed test.

I have no objection to the change itself; I'd merge it on the argument alone
once the gap suite lands. I just won't let the zeal line stand as proof of
something it doesn't discriminate.

## Two smaller notes

- #7630 cites "v0.5.1360-era" but `origin/main` is **v0.5.1357**, and its
  census reads 101 MB promoted at 200k where my own main measurement reads
  113,227,216 bytes. Worth confirming the profile wasn't taken on a tree
  carrying unmerged PRs — the attribution ordering (layout family ahead of the
  old-page family) is exactly the kind of claim a different tree can flip.
- The `barrier.rs` → `barrier_store.rs` split is a genuine pure move; the
  2000-line cap forced it and `check_file_size.sh` is clean.

proggeramlug:
## Merged as v0.5.1361 — on the argument, with the evidence question split out

Gap suite result accepted: the one reported regression is the known host-local
zlib runtime-dir flake (that's #6847, which I reopened yesterday for macOS after
#7599's audit caught it reproducing 3/3), disproven on-branch with a pinned
`PERRY_RUNTIME_DIR`. The 10 `node_fail → parity_fail` transitions are
oracle-coverage, and `test_gap_iterator_helpers_2874` improving to pass is
#7583's work showing up.

Re-verified here: runtime 1,902/0, `cargo check --all-targets` clean, JSON gap
family 7/7 byte-identical to node, all six lint scripts + file-size + fmt clean.

**Why I merged despite my open finding**: the change is sound by *argument* —
the materialiser owns each object end-to-end, finalize is reached on every path
(zero early exits between store and finalize in both parse functions, which I
walked), and the pointer case is conservative (`GC_LAYOUT_UNKNOWN` via
`layout_mark_unknown`, not a bare state store). Plus it is strictly more
conservative than what it replaces whenever `saw_pointer` is true. That is
enough.

**But I chased the vacuity question further and the benign explanation is
refuted**, so it is now filed as **#7635** rather than left in this thread:

Parse value strings are **not** longlived or interned — `string_storage_alloc`
(`string/mod.rs:504`) uses `arena_alloc_gc`, the ordinary nursery;
`arena_alloc_gc_longlived` is a different function. My probe's strings are 6–10
chars, above `SHORT_STRING_MAX_LEN = 5`, so they are real nursery
`StringHeader`s — collectable and movable. A record wrongly marked
`POINTER_FREE` should lose them, and it doesn't, under any instrument, with 8
retired quarantine sets and 16 copying minors observed live.

So the line in this PR's body — *"GC zeal + fromspace-protect arm: output
identical — no stranded children under forced evacuation"* — should not be
carried forward into the next PR in this family as evidence. It passes a
mutation that is guaranteed to strand children. #7635 asks for a probe that
faults, and records that "we can't build one" would itself be the finding: it
would mean the `POINTER_FREE` trace-skip is unobservable, which is worth knowing
before more optimizations are stacked on it.

One data point for #7630 while you are in here: my census on current main reads
**113,227,216 bytes / 1,657,966 objects** promoted at 200k — measured three
times independently now, most recently during #7624's audit. #7630 records
101 MB and cites "v0.5.1360-era" when `main` was v0.5.1357. Worth confirming
that step-zero profile was taken on `main`, because its attribution ordering put
the layout family ahead of the old-page family — and #7624 has since removed the
old-page family entirely.

Issue #7634, opened 2026-08-08
arr.push(f()) / arr.push(...g()) evaluate the receiver AFTER the argument, so a reassigning argument pushes onto the wrong array


proggeramlug:
Found while auditing `crates/perry-codegen/src/expr/array_push.rs` for the Layer 1
rooting migration (#7615, slice 3). Not a rooting bug — the opposite: the arm is
rooting-safe *because* it evaluates in an order the spec does not permit.

## The divergence

```ts
let a: number[] = [1];
function f(): number { a = [9]; return 2; }
a.push(f());
console.log(JSON.stringify(a));

let b: number[] = [1];
function g(): number[] { b = [9]; return [2, 3]; }
b.push(...g());
console.log(JSON.stringify(b));
```

| | output |
|---|---|
| `node --experimental-strip-types` (26.5.1) | `[9]` / `[9]` |
| perry (`main`, `PERRY_NO_AUTO_OPTIMIZE=1`) | `[9,2]` / `[9,2,3]` |

Per ES2024 the `MemberExpression` `a.push` is evaluated to a Reference **before**
the argument list, so the push lands on the array `a` named at that moment — the
one the program then throws away. Perry pushes onto the array `a` names
*afterwards*.

## Where

Both arms of `expr/array_push.rs` lower the value first and the receiver second:

```rust
// Expr::ArrayPush
let (v, v_bits) = lower_array_push_value(ctx, value, ...)?;   // arbitrary user code
let arr_box = lower_expr(ctx, &array_expr)?;                  // Expr::LocalGet, a slot read

// Expr::ArrayPushSpread
let src_box = lower_expr(ctx, source)?;                       // arbitrary user code
let arr_box = lower_expr(ctx, &Expr::LocalGet(*array_id))?;
```

It is only observable when the argument reassigns the receiver's binding, which
requires the binding to be captured-and-mutated (so it is boxed) or a module
global — otherwise the two orders name the same array.

## Why it is not fixed in #7615's slice 3

Reading the receiver **after** the value is exactly what makes the arm free of
rooting work today: the load observes whatever an evacuating cycle wrote back
into the local's alloca / box / module global, so no register is stale and
`operand_protection` answers `Reuse`. Restoring spec order means the receiver is
live across the value's lowering, i.e. it has to be rooted:

```rust
rooting::with_operands_rooted_across(
    ctx,
    &[&array_expr],          // the receiver, rooted first
    &[value],                // does the window collect?
[...]

Issue #7628, opened 2026-08-08
a[i]++ holds its receiver and index across three user-code-capable calls before js_dyn_index_set (#7154 shape)


proggeramlug:
Found while migrating `expr/instance_misc1.rs` onto the Layer 1 rooting API
(#7615 slice 2, #7627). The operand-to-operand half is fixed there; this is the
half that needs machinery the campaign does not yet have.

## The window

`Expr::IndexUpdate` lowers `a[i]++` / `--a[i]` as a read-modify-write over two
operands that are consumed by **four different calls**, with collection points
between them:

```text
obj_box, idx_box = <re-read once, below the operand group>
old      = js_dyn_index_get(obj_box, idx_box)   ; a getter here is user code
old_num  = js_to_numeric(old)                   ; a valueOf here is user code
new      = js_numeric_step(old_num, step)       ; ditto for a BigInt/object step
           js_dyn_index_set(obj_box, idx_box, new)   ; reads PRE-MOVE registers
```

`obj_box` and `idx_box` are NaN-boxed registers held across all three preceding
calls. If any of them drives an evacuating minor, `js_dyn_index_set` writes into
abandoned from-space memory: the element update silently does not appear on the
object the program keeps. That is #7154's shape, and the same one #7206 fixed on
the computed-read path.

Reproducible shape: `a[i]++` where `a`'s element is an object with a `valueOf`,
or where `a` is a Proxy / accessor-bearing object.

## Why #7627 did not close it

`rooting::with_operands_rooted` re-reads its group at **one** point, at the end
of the operand list. That is the right shape when a single collection point
separates the group from its consumer, and it is what closed the
receiver-across-the-index window here. It is the wrong shape when the operands
are consumed by different instructions with collection points between them —
re-reading above `js_dyn_index_get` puts them straight back in the window the
roots exist to close.

The raw API already has the primitive: `RootedOperands::reread_one`, added for
#7154's dynamic-call lowering, whose doc describes exactly this situation. What
Layer 1 lacks is a combinator that exposes a rooted operand group as something
`call_with_roots` can consume per use — i.e. an `Arg` variant that carries the
group's per-operand `Root` / `Reload` / `Reuse` decision and materialises it at
the instant each call is emitted, rather than once.

Per the template's rule (#7617), a combinator arrives **with its caller** and with
a written argument for why the existing ones cannot serve, not ahead of one. This
issue is that argument; the caller is this arm.

## Scope note

[...]

PR #7636, opened 2026-08-08
refactor(codegen): migrate the literal-accumulator modules onto the Layer 1 rooting API (#7615)

proggeramlug:
Slice 3 of the Layer 1 campaign (#7615) — the **literal accumulators**. Four
modules, 1776 lines, 18 raw sites, 7 hazard sites:
`expr/objects_arrays_lit.rs`, `expr/array_literal.rs`, `expr/object_literal.rs`,
`expr/array_push.rs`. All four are listed in `MIGRATED_MODULES`; none names
`expr::temp_root`. Follows the template (#7617) and slices 1a (#7618),
1b (#7620) and 2 (#7627).

## No new combinator

#7627 asked slice 3 to reuse `with_rooted_accumulator` rather than build a
second one. It does, and so does the one shape that looked like it needed a
fourth:

| site | combinator |
|---|---|
| `[a, ...b, c]`'s growing array (`js_array_push_f64` / `_hole` / `js_array_spread_append`) | `with_rooted_accumulator(Repr::Ptr)` + `advance` |
| `[a, b, c]`'s element group | `with_operands_rooted` |
| an object literal's half-built handle, both build paths | `with_rooted_accumulator(Repr::Ptr)` |
| a `this`-capturing method closure's deferred value | `with_rooted_accumulator(Repr::Boxed)`, **nested one per property** |
| `arr.push(...src)`'s operand pair | `with_operands_rooted` (empty window, emits nothing) |

The nesting is the interesting one. A method closure's value is consumed by the
`this` patch loop *below every remaining initializer*, so its root has to span
exactly the suffix of the literal that follows it, and the number of such values
is data-dependent. The three `with_operands_rooted*` forms cannot express that:
all three lower their own operand list **up front**, which for an object literal
would evaluate every property before storing any of them and reorder observable
side effects. `with_rooted_accumulator` *is* "a GC value held while more user
code is lowered", so the shape is one scope per such property, which falls out
as a small recursion over the property list (`lower_by_name_props`). The list it
returns is innermost-first and re-reversed, and that `reverse()` is pinned by a
test rather than by a comment.

What that replaces is #6951's flat `Vec` of raw slots that were **never
released** — correct only because `temp_root_truncate` is a stack cut and the
object handle sat below them, i.e. correct because of an invariant stated in a
comment ("keep `rooted_handle_begin` ahead of this loop and the release after
the patch loop"). It also leaked the whole group on a `?` from a later
initializer. Nested, the release is owned on every path out.

## Live bug found, and deliberately not fixed here

`arr.push(f())` and `arr.push(...g())` evaluate the **receiver after the
argument**. Per ES2024 the `MemberExpression` is evaluated to a Reference first,
so a `f` that reassigns `arr` must still push onto the array `arr.push` resolved
— node prints `[9]`, perry prints `[9,2]`. Filed as **#7634** with the
reproducer and the prescription.

It is not fixed in this slice because that evaluation order is precisely what
makes the arm rooting-free today: the receiver is an `Expr::LocalGet`, so
reading it *after* the value's arbitrary user code observes whatever an
evacuating cycle wrote back into the local's alloca / box / module global, and
`operand_protection` answers `Reuse`. Restoring spec order puts the receiver in
a window, i.e. every pointer-valued push gains a temp root — expressible with
the existing `with_operands_rooted_across` and needing no new combinator, but
needing a measured before/after on the pinned mini, which a behaviour-preserving
refactor has no mandate to run. Same rule that produced #7628 last slice.

## Scoped honestly

**Three of the four modules are translations, not repairs**, and the PR claims
them as such:

- `objects_arrays_lit.rs` — #7280 rooted the spread accumulator by hand and did
  it right: push before the first element, re-read before every append,
  **republish** the append's return value, read once below the last one. What
  the migration buys is that `advance` fuses the republish to the call, and that
  the release cannot become branch-conditional (#7462's shape, which shipped in
  a sibling arm).
- `array_literal.rs` — #6951's element group was already correct. What is new is
  the release: three exits each carried their own `temp_root_release`, and the
  `?` on every element's lowering released nothing at all.
- `object_literal.rs` — #6951's rooting was complete and correctly ordered,
  including loading the interned key *below* the value's lowering (the #7627
  finding, already right here).

**`array_push.rs`'s ledger line is vacuous on the committed source.** It named
no rooting symbol before the migration and names none after, exactly like both
slice-1 modules; only the sabotage arm makes it an assertion. The audit that
earned the listing is written into that file's header: both arms lower value
first / receiver second, and nothing they emit below that point
(`js_array_push_f64`, `js_array_concat`, the header probes,
`js_gc_note_slot_layout`, `js_write_barrier_slot`, `js_array_length`) either
holds a pointer across a moving window or fails to consume the one it is handed.

## A branch no corpus can reach

`lower_object_literal`'s by-name path with a non-empty property list — the one
carrying the `this`-patch machinery this PR restructures — **is not reachable
from TypeScript**. Since #809 every source-level literal containing a
`Prop::Method` is lowered to a source-ordered IIFE over `{}`
(`js_object_set_method_by_name`). Measured rather than assumed: over the whole
`gc_root_dominance_corpus.sh` corpus (129 sources, 149 modules) **every** emitted
`js_object_alloc` is `(i32 0, i32 0)`.

A branch no corpus reaches is a branch no IR A/B can speak for, so it gets four
unit tests built directly from HIR (`by_name_method_closure_tests`): the by-name
path is actually selected (`js_object_alloc(i32 0, i32 4)` + four
`js_object_set_field_by_name`), both patches run below the last property store,
the two patches are applied in **source order** — readable off the reserved
`this`-slot index, because the two closures deliberately reserve different ones
— and three slots are rooted and released innermost-first with the object handle
released last. Sabotage: deleting the `patches.reverse()` turns the third red
with `left: ["1", "0"] right: ["0", "1"]`.

**IR A/B**, both arms built from the same package set in one target dir, the
baseline arm from `c1365ed8a`'s copies of the five touched files:

| corpus | result |
|---|---|
| 4 probes × 2 environments (default; and `PERRY_RS4GC=0 PERRY_GC_MOVING_LOOP_POLLS=1 PERRY_INLINE_SHADOW_SLOT=0`) | **byte-identical** |
| curated dominance corpus — 129 sources, 149 modules, 2452 functions | 2450 identical, **2 differ** |
| dependency-scale corpus — 81 zod modules, 64 MB of IR | **byte-identical** |

Both differences are the same one instruction pair, in the iterator-result
literal `{value, done}`:

```diff
-  %v645 = load i64, ptr %v9
-  %v646 = bitcast double %v640 to i64
-  invoke void @js_object_set_field(i64 %v645, i32 0, i64 %v646)
+  %v645 = bitcast double %v640 to i64
+  %v646 = load i64, ptr %v9
+  invoke void @js_object_set_field(i64 %v646, i32 0, i64 %v645)
```

Root-plumbing only, and in the direction the API exists to enforce: the
accumulator's re-read is now fused to the emission that consumes it, so it lands
*after* the value's `bitcast` instead of before. Identical opcode multiset, same
operands, same call. Per #7625 the emitted IR is run-to-run deterministic, so
the double-compile control is no longer needed.

The probes cover every lowering family in the four modules — verified by counting
call sites rather than by reading the source: `js_object_alloc_with_shape` 14,
`js_object_set_field` 30, `js_object_alloc` 2, `js_gc_init_typed_shape_layout`
11, `js_array_alloc_literal` 4, `js_array_from_values` 10 (`PERRY_FULL_OUTLINE_IC=1`),
`js_array_spread_append` 14, `js_array_push_hole` 2, `js_array_clone_for_spread`
2, `js_array_concat` 3, `js_array_push_f64` 94. (The first p2 probe read its
literals' fields back and every one was scalar-replaced away; the probe now makes
them escape.)

**Gates.** `gc-root-dominance` green in **both** gated modes on **both** corpora
with an empty allowlist:

- curated: 2452 functions / 149 modules / 9846 root stores → 0 violations;
  `--seeded-violations 40` at 40/40; `--unrooted-allocas` 0 over 7867 gc-capable
  allocas;
- dependency-scale: 12899 functions / 81 modules / 12908 root stores → 0
  violations; 40/40 seeded; `--unrooted-allocas` 0 over 15242.

All four checker static audits pass (`--self-test`, `--audit-alloc-re`,
`--audit-poll-capable`, `--audit-immovable-sources`). Root-store counts are
**unchanged** on both corpora, which is what a byte-identical translation should
read.

**Tests.** `cargo test -p perry-codegen --lib` 699 pass (695 + the 4 new);
`--doc` 3 pass with both `compile_fail,E0499` arms still rejecting;
`cargo test -p perry-runtime --no-fail-fast` 1902 pass, 0 failed, first run;
`cargo check --all-targets` clean.

**Gap A/B**, 13 family filters over both prebuilt arms (`PERRY_SKIP_BUILD=1`,
so the binaries are the ones the IR A/B was measured on): 140 journal verdicts
each, **identical sets** (`diff` exit 0) — 126 pass, 1 crash. The crash is
`test_gap_gc_same_module_call_argument_rooting`, a harness 10 s timeout present
identically on both arms; compiled and run directly the binary finishes in 0.5 s,
so it is the `perry-dev` profile's margin, not a fault.

**Runtime instrument.** All four probes byte-identical in stdout and exit code
on both arms, and again under `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1
PERRY_GC_PROTECT_FROMSPACE_DEPTH=800 PERRY_GC_MOVING_LOOP_POLLS=1`. The
instrument was **live**, not merely enabled: `PERRY_GC_DIAG=1` prints 5
`[gc-fromspace-protect] retired_set=` lines, so copying minors actually ran and
quarantined from-space.

**Ledger sabotage**, per module — the assert stops at the first offender, so one
run cannot speak for four. A real, compiling `temp_root_push_i64` /
`temp_root_truncate` pair injected into each of the four turns
`migrated_modules_do_not_reach_past_the_rooting_api` red and names **both**
planted lines; restored, the ledger is green again. It also caught a genuine
violation in this PR's own new test — an assertion message quoting
`temp_root_truncate` is a code line, not a comment — which is now reworded.

**`lint`.** Every step of the job enumerated from `.github/workflows/test.yml`
was run (27 of them). All pass except `python3
benchmarks/ci_public_baseline_check.py`, which is red on pristine `origin/main`
and has been since 2026-07-29 — #7618, #7620 and #7627 each reported the same
step.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved memory safety while creating array and object literals, including spread operations and closures that capture `this`.
  - Ensured temporary values remain properly available during complex literal construction and are released reliably on success or failure.

- **Tests**
  - Added coverage for literal construction paths, closure patching, root-release ordering, unreachable paths, and garbage-collection safety checks.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1361
+**Current Version:** 0.5.1362
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1361"
+version = "0.5.1362"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1361"
+version = "0.5.1362"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1361"
+version = "0.5.1362"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1361"
+version = "0.5.1362"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1361"
+version = "0.5.1362"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1361"
+version = "0.5.1362"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1361"
+version = "0.5.1362"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1361"
+version = "0.5.1362"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1361"
+version = "0.5.1362"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1361"
+version = "0.5.1362"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1361"
+version = "0.5.1362"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1361"
+version = "0.5.1362"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1361"
+version = "0.5.1362"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1361"
+version = "0.5.1362"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1361"
+version = "0.5.1362"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1361"
+version = "0.5.1362"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1361"
+version = "0.5.1362"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1361"
+version = "0.5.1362"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1361"
+version = "0.5.1362"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1361"
+version = "0.5.1362"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1361"
+version = "0.5.1362"
 dependencies = [
  "perry-ffi",
  "scraper",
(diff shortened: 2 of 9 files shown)


proggeramlug:
## Audit before merge — verified, merged as v0.5.1362

**#7634 reproduced exactly** on my own build:

```
a.push(f())      perry [9,3]   node [9]
b.push(...g())   perry [8,7]   node [8]
```

The receiver is evaluated after the argument, so the push lands on the
reassigned array. **Filing it rather than fixing it was the right call** — the
current order is precisely what makes the arm rooting-free, so the fix puts
every pointer-valued push into a window. That is a measured change, and a
behaviour-preserving refactor has no mandate to run it.

Object/array literal behaviour byte-identical to node including property
evaluation order. Ledger sabotage on `array_literal.rs` red. Root-dominance both
modes: 129/129, 0 violations, 40/40 seeded, unrooted-allocas 0 over 7,867.
Suites 698/0 codegen, 1,902/0 runtime, `--all-targets` clean, all six lint
scripts + file-size + fmt clean.

## Three things this slice got right that I want on the record

1. **Declining to add a fourth combinator, with a written argument.** The
   nested-`with_rooted_accumulator` choice rests on a real semantic constraint —
   every `with_operands_rooted*` form lowers its operand list up front, which
   for an object literal would evaluate every property before storing any and
   **reorder observable side effects**. Refusing an abstraction because it would
   change semantics, and saying why in the module header, is exactly the bar for
   the ~80 slices still ahead.

2. **`array_push.rs`'s ledger line is vacuous on committed source, and you said
   so.** Its `haz: 6` are one false-positive shape (a raw `i64` whose only
   follower is an `Expr::LocalGet`). A ledger entry that cannot fail is the
   thing this campaign exists to avoid; naming it beats quietly banking the
   count.

3. **Both probe-vacuity catches.** The first object-literal probe was
   scalar-replaced away entirely — re-cut and call-site counts verified before
   trusting the A/B — and the new by-name-path tests were justified by
   *measuring* unreachability (across 129 sources every `js_object_alloc` is
   `(i32 0, i32 0)`) rather than asserting it, with a `reverse()`-deletion
   sabotage showing they discriminate. That is the same discipline #7635 now
   asks the layout-state family for.

The ledger catching a violation in your own new test — an assertion message
quoting `temp_root_truncate` counts as a code line — is a small thing that says
the gate is scanning what it claims to.

Issue #7640, opened 2026-08-08
index_get/index_set/property_set: ~20 arms make no rooting decision at all (the population the Layer 1 ledger cannot see)


proggeramlug:
Found while migrating the computed read/write modules onto the Layer 1 rooting API (slice 4 of #7615). **Filed rather than fixed** — see "Why not in slice 4" at the bottom.

## The finding

Migrating the *guarded* arms of `index_set.rs` / `index_get.rs` / `property_set.rs`
made it visible that a much larger population of arms in the same files makes
**no rooting decision at all**. They are not ordering mistakes against the raw
API — they never call it. They lower a receiver (and sometimes a key), lower more
user code, and then use the register.

This is a distinction the Layer 1 ledger cannot see, and it is worth stating on
#7615 directly: **a module listed in `MIGRATED_MODULES` is not an audited
module.** The ledger asserts "every rooting decision this module makes goes
through `crate::rooting`". It says nothing about a window with no decision.

Three of the arms below were adjacent to code slice 4 was already rewriting and
are fixed there (#7637, #7638, #7639). Everything in this issue is what is left.

## A. Store side — receiver (and sometimes key) live across the value

| `expr/index_set.rs` | shape | exposed |
|---|---|---|
| bounded-index-pair array store | `for (let i=0;i<a.length;i++) a[i] = {v:i};` | receiver |
| the `#5525` `recv_unknown` inline dyn-TA store | `function put(o:any,k:any){ o[k]=mk(); }` | receiver **and** key |
| `globalThis[k] = v` | | key (heap string); receiver MEDIUM |
| width-tracked TA, non-numeric index | `ta[k] = mk()` with `k: any` | receiver **and** key |
| TA runtime-key / TA final fallback / Uint8Array runtime-key | | receiver |

The bounded-index one is the sharpest, because it sits immediately *above* the
generic array-store arm that closed exactly this window in #7341, and because
`classify_for_length_hoist`'s body predicate (`expr_preserves_array_length`)
**accepts** `Expr::Object` / `Expr::Array` / `Expr::ArraySpread` values — so the
registered bounded pair and an allocating RHS coexist by design:

```ts
const a: unknown[] = new Array(1024);
for (let i = 0; i < a.length; i++) a[i] = { v: i };
```

Note also that `is_width_tracked_typed_array_receiver` keys on
`receiver_class_name`, a *declared* type. Per CLAUDE.md a declared type is a
[...]

Issue #7341, opened 2026-08-03
gc(layer 3): from-space quarantine catches 55 stale dereferences across the gap suite — the instrument is in CI but aimed at one synthetic fixture


proggeramlug:
## Summary

`PERRY_GC_PROTECT_FROMSPACE` catches **55 stale from-space dereferences across
the 479-test gap suite** — 44 of them in programs that otherwise exit cleanly and
produce correct output. The instrument already exists, is well built, and is
already wired into CI. It is simply **pointed at the wrong workload**, so none of
these have ever been visible.

This is layer 3 of `docs/engine-plan.md` (rooting in hand-written runtime Rust),
the layer that #7280's surviving fault moved into and that no static gate can
see.

## Why these were invisible

`scripts/gc_instrument_smoke.sh` is a good gate — it asserts both knob
directions and has a real non-vacuity check (zeal must produce strictly more
retirements than no-zeal). But it drives the instrument with:

```
PERRY_GC_MOVING_LOOP_POLLS=1 ... $WORK/fixture
```

over **one synthetic fixture**. Back-edge polls fire only while user JS runs, so
that configuration structurally cannot expose an unrooted pointer in runtime code
that does not re-enter user JS — which is most of the runtime. That limitation is
already documented from #7249; the smoke script predates applying it here.

Driving the same instrument by **allocation-point pressure** over **real
programs** instead:

```
PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=64 \
PERRY_GC_HEAP_LIMIT=8 PERRY_GC_INCREMENTAL=0 PERRY_CONSERVATIVE_STACK_SCAN=off \
  ./test_binary
```

## Results

Every gap test compiled and run twice — quarantine OFF as a control, quarantine
ON — so a pre-existing crash cannot be miscounted as a catch.

```
measured                                     479
QUARANTINE-ONLY FAULTS (control ok, exp SIGBUS)  55
  ...of which the control exited 0 (clean)       44
faulted in BOTH arms (pre-existing)               9
no fault in either arm                          415
```

The faults are self-diagnosing:

```
[gc-fromspace-protect] FAULT: signal 10 at 0x5cb232a03a9
  This address is RETIRED FROM-SPACE. The evacuating minor moved or
  freed the object here and the holder kept the pre-collection address.
  block=0x5cb232a0000 +937 retired_bytes=1048536 retired_by_minor=#0
[...]

Issue #7201, opened 2026-08-01
GC: a class expression with a `static { … }` block SIGSEGVs under PERRY_GC_MOVING_LOOP_POLLS=1 (static-`this` cell is not a rewritten root)


proggeramlug:
## Summary

A class **expression** carrying a `static { … }` block **SIGSEGVs** (`exit=139`) under `PERRY_GC_MOVING_LOOP_POLLS=1` when the block's body allocates. Clean under the shipped default. A variant that captures an outer value does not crash but computes the wrong `static` value 4 times in 400.

This is a residual site of the #7154 class that #7192 did **not** close, and a hard blocker for reverting #7161.

## Reproducer

```ts
function churn(): number {
  const a: any[] = [];
  for (let i = 0; i < 600; i++) a.push({ i: i, s: "q" });
  return a.length;
}
function make(): any {
  return class {
    static k: number = 1;
    static {
      (this as any).viaBlock = churn();
    }
  };
}
let bad = 0;
for (let r = 0; r < 400; r++) {
  const C: any = make();
  if (C.k !== 1) bad++;
  if (C.viaBlock !== 600) bad++;
}
console.log("bad", bad);
```

```
node --experimental-strip-types  →  bad 0
perry, default                   →  bad 0
perry, PERRY_GC_MOVING_LOOP_POLLS=1 (compile AND run)  →  exit=139, no output
```

Reproduced identically at `73a9084ea`, so **inherited, not caused** by #7184/#7192.

A capturing variant (`make(tag)` with `churn() + tag`) does not crash and reports `viaBlock 4` wrong out of 400 — that one already pushed a root before #7198 via the `!captured_args.is_empty()` clause, which is the evidence that the class object is *not* the stale thing here.

## Where it is

**Not** the class object. #7198 adds `!block_fns.is_empty()` to `Expr::ClassExprFresh`'s protection predicate, and the emitted IR confirms the object is now temp-rooted at allocation and re-read before every use:

```llvm
%r18 = call i64 @js_object_alloc(i32 3, i32 1)
call void @js_object_mark_class(i64 %r18)
call void @js_class_object_pin_parent(i64 %r18, i32 3)
%r19 = call i32 @js_gc_temp_root_push(i64 %r18)
...
%r33 = call i64 @js_gc_temp_root_get(i32 %r19)      ; re-read before arming
%r35 = <nanbox %r33>
call void @js_static_this_arm_value(double %r35)
[...]

PR #7642, opened 2026-08-08
refactor(codegen): migrate the computed read/write modules onto the Layer 1 rooting API (#7615)

### LINKED ISSUE #7637: index_set/property_set: `arr.length = f()` holds the array receiver across the value with no root
Found while migrating the computed read/write modules onto the Layer 1 rooting API (slice 4 of #7615).

## The bug

`crates/perry-codegen/src/expr/property_set.rs`, the `arr.length = v` arm:

```rust
if property == "length" && is_array_expr(ctx, object) {
    let arr_box = lower_expr(ctx, object)?;
    let val_double = lower_expr(ctx, value)?;   // <-- arbitrary user code
    let blk = ctx.block();
    let arr_bits = blk.bitcast_double_to_i64(&arr_box);   // <-- stale
    let arr_handle = blk.and(I64, &arr_bits, POINTER_MASK_I64);
    blk.call_void("js_array_set_length_strict", &[(I64, &arr_handle), (DOUBLE, &val_double)]);
```

The receiver is lowered first because `Set(O, "length", v, true)` evaluates the
reference before the value — spec order, and codegen follows it. That leaves the
`ArrayHeader*` in a bare SSA register while `f()` runs. `f()` allocates; a
back-edge poll inside it drives an evacuating minor; the *slot* `arr_box` was
loaded from is a registered root and gets rewritten, the register does not, and
`js_array_set_length_strict` then truncates or extends **the abandoned
from-space copy**. The array the program keeps is left at its old length.

This is #7154 exactly. Every other `PropertySet` arm in that same file has
carried `guard_store_operand` since #7154; this one was missed, and it is the
only arm in the file that reaches a runtime helper with a raw receiver handle
and no guard.

## Shape

```ts
const a: number[] = [1, 2, 3, 4, 5];
function shrink(): number { /* allocates */ return mk().n; }
a.length = shrink();
```

## Status

Fixed in the slice-4 PR by putting the receiver and the value in one
`rooting::with_operands_rooted` group, which derives the window from the list
order instead of restating it. IR is unchanged wherever the RHS provably cannot
collect (`operand_protection` returns `Reuse` and the combinator emits nothing).

### LINKED ISSUE #7638: index_set: two array element stores hold a heap-string KEY across the value with no root
Found while migrating the computed read/write modules onto the Layer 1 rooting API (slice 4 of #7615).

## The bug

Two array-element store arms in `crates/perry-codegen/src/expr/index_set.rs`
lower a **key that is a heap string** before the value, then use it after the
value's evaluation.

### 1. `arr[k] = v`, index not statically numeric

Gate: `is_array_expr(ctx, object) && !is_numeric_expr(ctx, index)`. The arm
exists precisely for keys that are *not* proven numeric — its own comment names
`forEach`'s `(item, k)` and for-in string keys as the motivating shapes.

```rust
let recv_guard = guard_store_operand_across(ctx, object, &arr_box, recv_collects);
let idx_double = lower_expr(ctx, index)?;    // heap string, routinely
let val_double = lower_expr(ctx, value)?;    // <-- arbitrary user code
...
call js_typed_feedback_array_set_index_or_string(site, arr_handle, idx_double, val_double)
```

The receiver **is** guarded (#7341). The key is not. `js_typed_feedback_array_
set_index_or_string` parses the key, so after an evacuating minor in the RHS the
element lands under a garbage key — or the parse reads retired memory.

### 2. `arr[stringKey] = v`

Gate: `is_array_expr(ctx, object) && is_string_expr(ctx, index)`. Here the key is
a heap string *by construction*, and **neither** operand is guarded:

```rust
let arr_box = lower_expr(ctx, object)?;
let key_box = lower_expr(ctx, index)?;
let (val_double, val_bits) = lower_value_for_optional_barrier(ctx, value, ...)?;  // <-- user code
let arr_handle = unbox_to_i64(blk, &arr_box);
let key_handle = unbox_str_handle(blk, &key_box);   // pre-move StringHeader*
```

## Why it was missed

The dynamic-**string**-key *object* arm further down the same file has pushed a
nested `key_guard` for exactly this since #7154. These two array arms were not
covered because their key is not *statically* a string on arm 1, and because
arm 2 predates the guard family.

## Shapes

```ts
const out: unknown[] = [];
for (const i in src) out[i] = mk();          // arm 2

function put(k: any, i: number) { out[k] = { i }; }   // arm 1
```

## Status

Fixed in the slice-4 PR. Both arms become one operand group over
`[object, index, value]` (arm 1) / `[object, index]` across `[value]` (arm 2),
so the receiver's window and the key's window are *derived* from list order
rather than restated — the receiver's spans the key and the value, the key's
spans the value, and the value is lowered last so it takes nothing.

### LINKED ISSUE #7639: index_set: the polymorphic o[k]=v fallback roots neither operand, and the string-key arm derives the receiver window from `value` alone (#7201 residue)
Found while migrating the computed read/write modules onto the Layer 1 rooting API (slice 4 of #7615).

Two arms of `crates/perry-codegen/src/expr/index_set.rs`, one bug each, both in
the receiver's window.

## 1. The polymorphic fallback guards NOTHING

The last-resort `Expr::IndexSet` arm — reached when nothing about the receiver
or the key is statically known, which is when both are heap values by default:

```rust
let obj_box = lower_expr(ctx, object)?;
let idx_box = lower_expr(ctx, index)?;
let (val_double, _) = lower_value_for_dynamic_index_set(ctx, value, ...)?;  // <-- user code
let obj_bits = ctx.block().bitcast_double_to_i64(&obj_box);                 // <-- stale
```

Both registers are then consumed across a four-block dispatch tree:
`js_is_symbol`, then `js_object_set_symbol_property(obj_box, idx_box, val)` on
the symbol arm, `unbox_str_handle(idx_box)` + the by-name setter on the string
arm, and `js_typed_feedback_object_set_index_polymorphic(obj_handle, idx_box, val)`
on the numeric arm. An evacuating minor inside the RHS lands the write on a
stale object under a stale key.

It is the most exposed arm in the file and the only one with no guard at all.

```ts
const m: any = {};
m[k] = mk();          // k: any, mk() allocates
```

## 2. The dynamic-string-key arm derives the receiver's window from `value` alone

```rust
let obj_box = lower_expr(ctx, object)?;
let recv_guard = guard_store_operand(ctx, object, &obj_box, value);  // window := value ONLY
let key_box = lower_expr(ctx, index)?;                              // <-- but the receiver is live across THIS too
let key_guard = guard_store_operand(ctx, index, &key_box, value);
```

This is the half-measure #7201 named in prose and then did not apply here:

> Deriving it from the value alone — which is what every caller did before
> #7201 — leaves `o[f()] = 1` unguarded, because the literal `1` cannot collect
> while `f()` obviously can.

`guard_store_operand`'s two-argument form structurally cannot say "the window is
everything after this operand"; only `guard_store_operand_across` can, and this
call site used the wrong one. So:

```ts
o[allocatingKey()] = 1;    // receiver unguarded across allocatingKey()
```

## Status

Both fixed in the slice-4 PR by making each arm one
`rooting::with_operands_rooted_across` group over `[object]` / `[object, index]`
with `[value]` as the caller-controlled step. The point is not that the flag is
now right, it is that there is no flag: `with_operands_rooted_window` computes
each operand's `collects` as `across_collects || any_may_trigger_gc(exprs[i+1..])`,
so "the receiver is live across everything after it" is a property of the list
rather than a thing the author has to remember.

The same arm's two nested guards also collapse to one group, which retires the
release-inner-to-outer obligation (`temp_root_truncate` is a stack CUT, so
releasing the receiver first silently dropped the key's slot).

proggeramlug:
Slice 4 of the Layer 1 campaign (#7615): the computed read/write lowerings.

**Nine modules, all migrated end to end and listed in `MIGRATED_MODULES`** —
`expr/index_get.rs`, `expr/index_get/guarded_array.rs`,
`expr/index_get/inline_dyn_typed_array.rs`, `expr/index_set.rs`,
`expr/index_set_typed_array.rs`, `expr/property_get.rs`,
`expr/property_get/globalget.rs`, `expr/property_get/helpers.rs`,
`expr/property_set.rs`. Nothing names `expr::temp_root` after this, and no
module is left half-migrated.

## One shape, no fourth combinator

Every rooting decision in this slice was the `StoreOperandGuard` family — lower
the receiver, guard it, lower the rest, re-read, store, release — and that is
exactly what `with_operands_rooted` / `with_operands_rooted_across` already say.
The `across` form is needed wherever the value is lowered by
`lower_value_for_dynamic_{index,property}_set` or
`lower_value_for_optional_barrier`, native-rep lowerings the operand list cannot
produce; everywhere else the plain form serves.

## Honest scoping: three load-bearing, six vacuous

`index_get.rs`, `index_set.rs` and `property_set.rs` named the escape hatch
before the migration (`guard_store_operand`, `guard_store_operand_across`,
`reread_store_operand`, `release_store_operand`, `expr_may_trigger_gc`), so
their ledger lines hold on the committed source.

The other six named nothing, so **their lines are vacuous** and only the
sabotage arm makes them assertions. Each carries the audit that earned it in its
own file header rather than banking the count here. The one worth repeating:
the campaign map credits `property_get/globalget.rs` with **6 hazard sites** and
all six are the same false positive — `js_get_global_this_builtin_value` →
`ctx.strings.intern` → `unbox_to_i64` → `load`. `intern` and `format!` emit no
IR. A window is measured in *emissions*, and there are none. That module lowers
no user expression at all.

`property_get.rs`'s three `.call(I64, "js_*")` sites each hand their raw pointer
to a `nanbox_*_inline` in the same block, so `call_rooted` has no site there —
rooting them would add temp-root traffic to close a window that does not exist.

## Three live bugs, fixed here

Each is the same window every sibling arm has guarded since #7154, and each is
on an arm whose store already routes through a runtime helper, so the root is
noise beside the call.

- **#7637** — `arr.length = f()` had no guard at all. Receiver lowered first
  (spec order), `f()` allocates, and `js_array_set_length_strict` then truncates
  the abandoned from-space copy.
- **#7638** — two array element stores held a heap-string KEY across the value.
  `arr[k] = f()` with a non-numeric index rooted the receiver but not the key —
  on the one arm whose entire purpose is keys that are *not* proven numeric —
  and `arr[stringKey] = f()` guarded neither operand.
- **#7639** — the polymorphic `o[k] = f()` fallback guarded neither operand,
  and it is reached precisely when nothing about receiver or key is known, so
  both are heap values by default. Separately the dynamic-string-key arm derived
  the receiver's window from `value` **alone** — the half-measure #7201 named in
  prose — leaving `o[f()] = 1` unguarded.

That last one is the argument for the combinator rather than for a better flag.
`with_operands_rooted_window` computes each operand's window as
`across_collects || any_may_trigger_gc(exprs[i+1..])`, so "the receiver is live
across everything after it" becomes a property of the operand list instead of
something an author restates per site. `guard_store_operand`'s two-argument form
structurally could not say it. The same arm's two nested guards also collapse to
one group, retiring the release-inner-to-outer obligation that
`temp_root_truncate`'s stack-cut semantics imposed.

## What the migration surfaced and did NOT fix — #7640

Translating the guarded arms made visible that ~20 arms in the same files make
**no rooting decision at all**: the bounded-index array store (which sits
immediately above the arm #7341 fixed, and whose loop predicate explicitly
accepts allocating object/array RHSs), the `#5525` inline typed-array stores,
`globalThis[k] = v`, ten read-side arms of `index_get.rs` where the base is live
across the key, an unsubstantiated "statepoint re-read" claim above the
class-field store, and a `unbox_str_handle` ordering hazard that is #7280's
case (a) — a raw `i64` no temp root can name.

Filed with repros and a suggested order rather than fixed, because they sit on
inline fast paths and guard diamonds where a root is a **measured** cost, and
this slice was not permitted to benchmark. Slice 3's `arr.push` precedent
(#7634) is the standard: a behaviour-preserving refactor does not smuggle in a
change whose cost nobody measured.

That gap is now written into the ledger itself, because it is the campaign's
most misreadable result:

> **A listed module is not an audited module.** The ledger asserts that every
> rooting decision a module makes goes through `crate::rooting`. A window with
> no decision at all is invisible to it.

## Verification

**IR identity.** Byte-identical on all 149 modules of the curated corpus. On the
dependency-scale (zod) corpus, 80 of 81 byte-identical; the single difference is
one function, `perry_closure_..._schemas_ts__126`, and it is root-plumbing only:
the dynamic-string-key `IndexSet` arm's two operands are now re-read and
released in **operand-list order** (receiver, key) instead of the hand-written
inner-to-outer order (key, receiver). Same two slots, same set of clears, no
root added or removed — 32 diff lines, all register renumbering downstream of
that swap.

**Probe vacuity checked before trusting the A/B.** Every migrated arm's call
sites were counted in the emitted IR first: 543 `object_get_field_by_name_f64`,
37 `object_set_field_by_name`, 16 `object_get_index_polymorphic`, 10
`array_set_index_or_string`, and so on. The one arm the curated corpus never
calls is `js_array_set_length_strict` — **zero calls across all 129 sources** —
which is why the repairs get unit tests instead.

**Gates.** Both gated modes green on both corpora with an **empty allowlist**;
`--seeded-violations 40` at 40/40 on each; `--unrooted-allocas --moving-only` at
0 over 7867 (curated) and 15242 (dep) gc-capable allocas; stale-register ratchet
unchanged at 23 / 115. Checker `--self-test`, `--audit-alloc-re`,
`--audit-poll-capable`, `--audit-immovable-sources` all clean.

**Ledger sabotage, per module.** A real, compiling `temp_root_push_double` /
`temp_root_truncate` pair was planted in each of the nine in turn — the assert
stops at the first offender, so one run cannot speak for nine — and each was
recorded red naming its own file and lines. Two of the nested submodules needed
`crate::expr::temp_root::` rather than `super::temp_root::`; with `super::` the
plant did not compile, which is a build failure and not a sabotage, so those two
were re-run with the resolving path.

**Tests.** `-p perry-codegen --lib` 702 passed; `--doc` 3 passed (both
`compile_fail` doctests still reject); `-p perry-runtime --no-fail-fast` 1902
passed, 0 failed, no rerun needed. `cargo check --all-targets` clean — the
`#[cfg(test)]` check #7631 asked for.

**Gap families**, both arms, `PERRY_NO_AUTO_OPTIMIZE=1`, prebuilt release
binaries per arm: `test_gap_computed` (5), `test_gap_prop` (1),
`test_gap_array` (13), `test_gap_object` (9), `test_gap_gc_` (30). **Verdict sets
identical on both arms**, including the two pre-existing failures —
`test_gap_prop_plan_cache_invalidation` (parity) and
`test_gap_gc_same_module_call_argument_rooting` (crash) — which reproduce on a
baseline build of `e4b298946` and are not this PR's.

**Lint**: 23 of 24 enumerated steps pass. The one red is
`benchmarks/ci_public_baseline_check.py`, red on pristine `main` since
2026-07-29 and reported by #7618, #7620, #7627 and #7636 alike.

## Coverage for the repaired arms

They are not reachable from ordinary TypeScript, and that was measured rather
than assumed. Six probe shapes — module-const and parameter receivers, string /
symbol / `any` keys, inside and outside a function — all routed past the
subject, because an ordinary `o.k = v` / `o[k] = v` assignment lowers to
`Expr::PutValueSet` and reaches the dynamic IC, not the `Expr::PropertySet` /
`Expr::IndexSet` arms in these modules.

So the coverage is built from HIR, and each test is **differential**: the same
store compiled twice, once with an allocating RHS and once with an inert one,
asserting the allocating one reserves strictly more root slots. An inert RHS
makes `operand_protection` return `Reuse` and the combinator emit nothing —
which is the property keeping this slice byte-identical on the corpus — so the
two runs bracket the repair and no corpus drift can make the assertion vacuous.
Each test also asserts by name that its arm was reached, and the slot count is
read under both root lowerings so it cannot silently measure zero. Sabotaging
the `arr.length` repair turns exactly that one test red, and only it.

Closes #7637, closes #7638, closes #7639. Map comment on #7615 follows.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability for computed property reads and writes, including array indexing, dynamic keys, and array-length updates.
  - Prevented potential instability during operations that trigger memory allocation.
  - Preserved correct runtime behavior for numeric, string, and symbol-based property access.

- **Tests**
  - Added coverage for computed stores and allocation-sensitive scenarios across supported runtime paths.

- **Documentation**
  - Expanded release documentation covering the completed memory-safety improvements and verification results.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1363
+**Current Version:** 0.5.1364
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1363"
+version = "0.5.1364"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "perry-ffi",
  "scraper",
@@ -5813,7 +5813,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-commander"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5821,7 +5821,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cron"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "chrono",
  "cron",
@@ -5831,7 +5831,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dayjs"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "chrono",
  "perry-ffi",
@@ -5839,7 +5839,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-decimal"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "perry-ffi",
  "rust_decimal",
@@ -5847,7 +5847,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dotenv"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "perry-ffi",
  "serde_json",
@@ -5855,7 +5855,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ethers"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "perry-ffi",
  "rand 0.10.1",
@@ -5863,7 +5863,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-events"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5871,14 +5871,14 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-exponential-backoff"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "perry-ffi",
 ]
 
 [[package]]
 name = "perry-ext-fastify"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "bytes",
  "http-body-util",
@@ -5896,7 +5896,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-fetch"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "bytes",
  "lazy_static",
@@ -5909,7 +5909,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-http"
-version = "0.5.1363"
+version = "0.5.1364"
 dependencies = [
  "bytes",
  "h2",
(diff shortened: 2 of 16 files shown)


proggeramlug:
## Audit before merge — verified, merged as v0.5.1364

All three in-slice fixes verified behaviourally against node 26.5.1 —
`a.length = shrink()`, `o[k()] = 1`, `o2[key2()] = key2()` (evaluation order
preserved: `{"z1":"z2"}`), two consecutive computed element stores,
`globalThis[k]=v`, and a typed-array computed store — byte-identical, and
identical again under zeal + from-space protect.

Ledger sabotage red. Root-dominance both modes on the corpus: 129/129, 0
violations, 40/40 seeded, unrooted-allocas 0. Codegen 701/0, runtime 1,902/0,
`--all-targets` clean, all six lint scripts + file-size + fmt clean.

**One correction to your lint line**: `ci_public_baseline_check.py` is **no
longer** the known `main` red — I regenerated the public artifact on the mac
mini and merged it as #7641 while you were working. It exits 0 on `main` now,
so slice 5 should see 24/24 and treat any red there as its own.

## Three things from this slice worth carrying forward

1. **Six of nine ledger lines declared vacuous rather than banked**, each with
   its audit in the file header — including the observation that
   `globalget.rs`'s six mapped "hazards" are one false positive because
   `ctx.strings.intern`/`format!` emit no IR, and **a window is measured in
   emissions**. That sentence is the clearest statement of the campaign's
   hazard-counting rule anyone has written.

2. **The sabotage-verification trap you found is the most valuable output
   here**, and I confirmed it applies to my own audit method: with
   `super::temp_root::` the plant does **not compile** in nested submodules,
   and a harness grepping only for `FAILED` scores that build error as a
   successful sabotage. I now check for `error[` alongside `FAILED` — my run
   above reports "0 compile errors — sabotage was real". Slice 5's brief will
   carry it.

3. **Probe vacuity checked before trusting the A/B, again** — call sites
   counted in emitted IR (543 / 37 / 16 / 10), which is how you found
   `js_array_set_length_strict` is called **zero times** across all 129 sources
   and that six hand-written probes all missed because ordinary `o.k = v`
   lowers to `Expr::PutValueSet` and reaches the dynamic IC instead. The
   differential HIR-built unit tests (allocating vs inert RHS, asserting
   strictly more root slots) are the right answer to an arm no source can
   reach.

**#7640 filed rather than fixed is the right call** and matches #7634's
precedent: those arms sit on inline fast paths and guard diamonds where a root
is a *measured* cost, and you were correctly denied a benchmark host. When the
mini frees up, that ticket is a perf-gated follow-up, not a refactor.

Housekeeping: I'll clear the `node_modules/` your dep-scale corpus needed when
I reclaim the worktree — thanks for flagging it.

Issue #179, opened 2026-04-24
gc: block-persistence conservatism inflates peak RSS on tight allocation loops


proggeramlug:
## Summary

Perry's GC includes a \"block-persistence\" pass at [`crates/perry-runtime/src/gc.rs:1090`](crates/perry-runtime/src/gc.rs#L1090) that marks ALL objects in an 8 MB arena block live whenever any single object in the block is root-reachable. This was added to protect against untracked arena refs sitting in caller-saved registers that the conservative stack scan can't capture (issues #43 / #44 pattern — dangling-pointer crashes when GC freed still-reachable arena objects).

The conservatism is catastrophic for **tight allocation loops that co-locate fresh per-iteration data with pre-existing long-lived state**. The `JSON.parse` case was traced in detail in #149; summary of the mechanism:

1. Block 0 (and sometimes block 1) contains long-lived data: interned string keys, shape-cache `keys_array`s, the blob being parsed, any caller-level root arrays.
2. The same block also contains early-iteration allocations (the first `new Foo()` in the loop always lands adjacent to the setup data).
3. Every subsequent iteration's allocations go into new blocks.
4. GC marks block 0 live (because of the long-lived data). Block-persist marks EVERY object in block 0 live — including the dead early-iteration objects.
5. Those dead objects have field values pointing into *later* blocks (fresh name strings, tag arrays, nested objects).
6. The persist pass iterates to fixed point, following pointers across blocks — at iter 49 of `bench_json_roundtrip`, 2 live blocks / 37 truly-reachable objects cascaded to 30 live blocks / 3 M \"live\" objects over 11 rounds.
7. `gc()` becomes a no-op. RSS grows linearly with iteration count.

## Measured impact (`bench_json_roundtrip` at v0.5.190)

| | Perry | Bun | ratio |
|-|------:|----:|------:|
| Speed | 315 ms | 248 ms | 1.27× |
| Peak RSS | **318 MB** | **83 MB** | **3.83×** |

The speed gap is tolerable (recently closed from 2.4× to 1.27×, see #149 closeout). The RSS gap is the architectural cost of block-persist + bump-arena vs. Bun's per-object GC.

## Not just JSON

Any tight `new ClassName()` / `new Array()` / tight parser loop that runs after meaningful setup hits the same pattern. The setup's long-lived roots anchor block 0 live; the first loop iteration allocates adjacent to them; block-persist pulls in the dead iter-0 objects on every GC and cascades from there.

We've seen the contours of this before:
[...]

PR #7643, opened 2026-08-08
test(gc): make the POINTER_FREE misdeclaration hazard detectable (#7635)

### LINKED ISSUE #7635: gc: forcing POINTER_FREE on a pointer-bearing object strands nothing — our zeal/protect instruments do not discriminate the layout-state hazard
## The problem

While auditing #7633 I sabotaged `layout_finish_deferred_boxed_object(ptr,
saw_pointer)` → `(ptr, false)` — i.e. **every JSON-parsed record claims
`POINTER_FREE` while holding heap pointers**. That is the stranded-live-child
hazard #7630's own soundness note names, and the exact failure the layout state
exists to prevent.

**It produces byte-identical correct output under every instrument we have:**

| probe | plain | `ZEAL=1 PROTECT_FROMSPACE=1` | `FORCE_EVACUATE=1` |
|---|---|---|---|

The second probe was built specifically because the first was vacuous: 4,000
records × 2 pointer fields, 40 rounds of churn to force promotion, children read
back only after the churn, so the record's slot is the sole path to each string.
**8 retired quarantine sets and 16 copying minors observed live** — the
instruments were armed and the collector was moving.

## Why this is not "the branch is safe"

The obvious benign explanation is refuted. Parse value strings are **not**
longlived or interned: `string_storage_alloc` (`string/mod.rs:504`) uses
`arena_alloc_gc`, the ordinary nursery arena — `arena_alloc_gc_longlived` is the
*other* function. My probe's strings are 6–10 chars, above
`SHORT_STRING_MAX_LEN = 5`, so they are real heap `StringHeader`s in the
nursery. They are collectable and movable, and a `POINTER_FREE` record should
therefore lose them.

So either the tracer does not honour `POINTER_FREE` on this path (in which case
the state's trace-skip is not buying what it claims anywhere), or something
re-marks these objects that neither I nor the PR author identified.

## Why it matters

Every future change to layout-state bookkeeping — and #7630's family has more
queued — will cite "zeal + from-space protect: output identical" as evidence of
no stranded children. **That sentence currently discriminates nothing**: a
mutation guaranteed to strand children passes it. This is CLAUDE.md's hazard 4
(a gate that runs but whose subject never did), applied to the instrument rather
than the job.

#7633 merged on the strength of its *argument* — the materialiser owns each
object end-to-end, finalize is reached on every path (verified: zero early exits
between store and finalize in both parse functions), and the pointer case is
conservative — plus a clean gap suite. That was the right call. But the argument
is now the only thing holding it up, and the next such change may not have one
as tight.

## What would close this

A probe where forcing `POINTER_FREE` on a pointer-bearing object **faults or
corrupts**. If one cannot be constructed, that is the finding: it means our
`POINTER_FREE` trace-skip is unobservable, and we should understand why before
building more optimizations on top of it. Either outcome is worth knowing, and
the probe becomes the regression test the whole family has been missing.

Starting points: check whether the full mark-sweep's conservative stack scan
(#7558) reaches these objects independently of their layout state; check whether
old-gen tracing consults `_reserved` layout bits at all after promotion; and try
a shape where the record is never promoted (small parse, forced minor only).

proggeramlug:
Closes #7635.

## The question

#7635 sabotaged `layout_finish_deferred_boxed_object(ptr, saw_pointer)` to
`(ptr, false)` — every JSON-parsed record claiming `POINTER_FREE` while holding
heap strings, the stranded-live-child hazard the layout state exists to prevent
— and got **byte-identical correct output** under `PERRY_GC_ZEAL=1
PERRY_GC_PROTECT_FROMSPACE=1` and under `PERRY_GC_FORCE_EVACUATE=1`, with
copying minors and retired quarantine sets observed live. Is the hazard
detectable at all?

## Answer: yes, and the instruments were never at fault

**`js_json_parse` routes a top-level array of 1 KB–16 MB through the lazy tape**
(`json_tape`, default since #179). `parse_object` — the function carrying the
sabotage — therefore does not run at `JSON.parse` time; it runs when an element
is first *read*. The probe read its records only after the churn, so every
misdeclared record was materialised after the last collection.

Confirmed rather than inferred: a temporary audit in
`heap_payload_slot_selection` on the sabotaged build counted **zero** objects in
`POINTER_FREE` state with pointer-bearing payload words ever handed to the
collector, across the whole run. Nothing was stranded because nothing was there.
This is CLAUDE.md's hazard 4 — the gate ran but its subject never did — applied
to a probe rather than a job.

Re-run so the misdeclared records actually live across a collection (sabotaged
vs clean runtime, `perry-dev`, macOS arm64, `PERRY_NO_AUTO_OPTIMIZE=1`):

| arm | clean | sabotaged |
|---|---|---|
| default (lazy), read after churn | exit 0 | exit 0, byte-identical, `dangling=0` |
| `PERRY_JSON_TAPE=0`, read after churn | exit 0 | **SIGSEGV (139)**; `dangling=8000 owners=4000` on the first scanned cycle (vs `dangling=0` clean); `PERRY_GC_PROTECT_FROMSPACE=1` prints `FAULT: signal 10 at 0x…` |
| default (lazy), records touched BEFORE the churn | exit 0 | **7,872 of 8,000 values read back wrong** |

`dangling=8000 owners=4000` is exactly 4,000 records × 2 pointer fields.

One instrument *is* structurally blind and it is worth naming:
`PERRY_GC_VERIFY_EVACUATION` walks the same enumeration the rewrite pass walks,
i.e. it asks this very layout state which slots exist.
**`PERRY_GC_FROMSPACE_SCAN=1` is the layout-independent one** — a whole-payload
word scan consulting no root enumeration and no layout state — and it is the
knob to reach for on this hazard class.

## What ships

**`crates/perry-runtime/src/gc/tests/copying/deferred_finalize_7635.rs`** — four
tests that need no workload at all, so they cannot be defeated by a lazy path, a
GC that did not happen to run, or conservative-scan residue:

1. `finalize_settles_pointer_free_or_unknown_and_nothing_else` — the finalize's
   two exact outcomes (no pointer ⟹ keep the birth state; any pointer ⟹
   `GC_LAYOUT_UNKNOWN`, no mask, no stranded descriptor).
2. `a_materialised_record_keeps_its_children_traced_and_rewritten_7635` — the
   materialiser's construction loop byte for byte; the collector must enumerate
   every field, and a copying minor must relocate both children and rewrite both
   slots. Gated on `copied_objects >= 3` so a cycle that moved nothing cannot
   pass.
3. `json_parse_record_keeps_its_string_values_traced_and_rewritten_7635` — the
   same invariant through the **real** `js_json_parse` entry point, so the
   finalize call site in `json/parser.rs` is covered and not merely the helper.
   #7635's sabotage was applied at that call site and tests 1–2 stay green
   through it.
4. `a_misdeclared_pointer_free_record_strands_its_child` — the **sabotage arm,
   made permanent**: the identical construction with `saw_pointer` forced to
   `false`, asserting the collector enumerates ZERO slots and that correcting the
   state alone makes the same record fully enumerable. Asserted on the
   enumerator, so it leaves no stale pointer behind.

Sabotage-verified in both directions, locally:

- `json/parser.rs` → `(js_obj as usize, false)` (#7635's exact mutation): test 3
  fails at the state assertion, and with that assertion neutered it fails again
  at the enumerator assertion.
- `layout_finish_deferred_boxed_object` itself made a no-op: **all four** fail.
- Restored: all four green.

**The doc comment on `GC_LAYOUT_POINTER_FREE`** now records what can and cannot
verify a claim about this state, so the next PR in this family does not cite
"clean under zeal + from-space protect" as evidence without first showing the
subject existed during a collection.

No GC behaviour changes; no new knobs; no codegen changes.

## Validation

- `cargo test -p perry-runtime --lib --no-fail-fast` (perry-dev)
- `cargo check --all-targets`
- the full `lint` script set from `.github/workflows/test.yml`:
  `workspace_architecture` (+self-test), `check_file_size.sh`, `binding_pins`,
  `gc_store_site_inventory` (+self-test), `addr_class_inventory` (+self-test),
  `class_id_collisions`, `raw_handle_debt` (+self-test), `gap_snapshot`
  self-test, `parity_known_failures` (+self-test, --audit),
  `gc_gate_wiring_check` (+self-test), `gc_matrix_liveness_check` (+self-test,
  --check-registry), `check_test_registration` (+self-test) — all exit 0.

Root-dominance is not re-run: nothing under `perry-codegen` changed.

## Follow-up worth considering (not in this PR)

The end-to-end shape is now a known-good detector: `PERRY_JSON_TAPE=0` +
`PERRY_GC_FROMSPACE_SCAN=1` over a parse-then-churn workload, asserting
`dangling=0` **and** that a copying minor actually ran. That would make a real
gate for the whole layout-state family. Per CLAUDE.md a new gate must be run
once and then promoted to required, which is a separate change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when processing JSON data with deferred materialization.
  * Ensured pointer-bearing records remain tracked and correctly updated during memory cleanup and relocation.
  * Added safeguards and regression coverage for incorrect pointer-free layout declarations.

* **Documentation**
  * Documented deferred materialization behavior, layout finalization assumptions, verification limits, and diagnostic guidance.

* **Chores**
  * Updated the application version to 0.5.1365.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1364
+**Current Version:** 0.5.1365
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1364"
+version = "0.5.1365"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1364"
+version = "0.5.1365"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1364"
+version = "0.5.1365"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1364"
+version = "0.5.1365"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1364"
+version = "0.5.1365"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1364"
+version = "0.5.1365"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1364"
+version = "0.5.1365"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1364"
+version = "0.5.1365"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1364"
+version = "0.5.1365"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1364"
+version = "0.5.1365"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1364"
+version = "0.5.1365"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1364"
+version = "0.5.1365"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1364"
+version = "0.5.1365"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1364"
+version = "0.5.1365"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1364"
+version = "0.5.1365"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1364"
+version = "0.5.1365"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1364"
+version = "0.5.1365"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1364"
+version = "0.5.1365"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1364"
+version = "0.5.1365"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1364"
+version = "0.5.1365"
 dependencies = [
  "perry-ffi",
  "rusqlite",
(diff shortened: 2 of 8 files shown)


proggeramlug:
## Audit — merged as v0.5.1365

I reproduced the decisive claim independently before reading the tests, and it corrects **my** finding, not the instruments'.

**Reproduced, sabotaging `json/parser.rs` myself** (both call sites → `saw_pointer = false`, perry-dev, `PERRY_NO_AUTO_OPTIMIZE=1`, arm64):

| arm | clean | sabotaged |
|---|---|---|
| default (lazy tape) | exit 0, correct | exit 0, **byte-identical correct** |
| `PERRY_JSON_TAPE=0` | exit 0, correct | **exit 138 = SIGBUS** |
| `PERRY_JSON_TAPE=0 PERRY_GC_FROMSPACE_SCAN=1` | — | `dangling=8000 owners=4000`, `never_dirty=8000`, offenders named with `space=Survivor1 → NurseryEden DANGLING` |

`owners=4000` is exactly the 4,000 records, `dangling=8000` exactly their two fields each. So the hazard is real, detectable, and my #7635 "no instrument can see it" conclusion was wrong for the reason given here: `js_json_parse` is lazy for 1 KB–16 MB top-level arrays, `parse_object` runs on first *read*, and my probe read after the churn. **CLAUDE.md hazard 4 — "the gate runs but its subject never did" — applied to a probe rather than a job.** Worth naming plainly: I wrote that hazard into CLAUDE.md and then walked into it.

**Sabotage-verified in both directions, independently of your run:**
- neutering `layout_finish_deferred_boxed_object` to an early `return` → **4/4 red**, and I checked for `error[` — **0 compile errors**, so the sabotage was real code, not a build failure being misread as a pass (the trap #7642 found);
- #7635's exact parser mutation → the `js_json_parse` test red, other three green. That asymmetry is the proof the parse test covers the real *call site* and not just the helper.

**What makes these tests worth having:** they are workload-free, so no lazy path, GC-that-didn't-run, or conservative-scan residue can defeat them; and both positive tests assert `copied_objects >= 3` before concluding, so a cycle that moved nothing can't score green. The permanent sabotage arm is the piece that makes the positive tests detectors rather than formalities — and it restores state via `layout_mark_unknown` before the guard drops, so it leaves no landmine for a later cycle on that thread.

**Gates:** I enumerated the `lint` job from `test.yml` rather than trusting a checklist — **20/20 green** (that extraction turned up `parity_known_failures.py --audit` and `gap_snapshot.py --self-test`, neither of which was in my memorized set). `cargo fmt --all --check` clean, `cargo test -p perry-runtime --lib --no-fail-fast` **1906 passed / 0 failed**. Root-dominance correctly skipped — no codegen changed.

Two things I'm keeping from this beyond the tests. The `GC_LAYOUT_POINTER_FREE` doc comment now says which instrument is blind and why (`PERRY_GC_VERIFY_EVACUATION` walks the same enumeration the rewrite pass walks — it asks the state under test what to check), which is the kind of thing that is otherwise re-derived once per incident. And the qualification you added to #7633's fragment is the right call: its argument was sound, its probe could not have shown it either way, and the fragment now says so instead of leaving a clean-under-zeal sentence for the next change to inherit as evidence.

Your suggested follow-up — `PERRY_JSON_TAPE=0` + `PERRY_GC_FROMSPACE_SCAN=1` over parse-then-churn, asserting `dangling=0` **and** that a copying minor ran — is filed as #7647. Correctly a separate change: a new gate has never been green, so it gets one observed run before promotion.

Issue #7025, opened 2026-07-29
gc-matrix: the 'moved' liveness counter sums the C4b mark-sweep evacuation with the copying minor, so a move-arm can be green with zero scavenged objects


proggeramlug:
Found while re-measuring the matrix after #7019. Small, but it silently weakens the one thing this gate exists to assert.

## The problem

`scripts/gc_repsel_matrix.sh` derives its `moved=` evidence — the liveness predicate for every `requires=move` arm — by summing **two different collectors' counters** into one number:

```sh
moved=$( { grep -oE 'moved_objects=[0-9]+' ... ; \
           grep -oE '\[gc-copy-minor\] ran copied_objects=[0-9]+' ... ; } \
         | grep -oE '[0-9]+$' | awk '{s+=$1} END {print s+0}')
```

* `moved_objects=` comes from the **C4b evacuation policy inside the mark-sweep collector** — the pre-existing non-moving-minor path that relocates tenured objects during a full GC.
* `[gc-copy-minor] ran copied_objects=` comes from the **evacuating young-gen scavenge**, which is what #7019 made default-on and what the `move` arms are meant to certify.

A cell can therefore report `moved=7 610 512` and be marked live for a `move` arm while the copying minor never ran once.

## It is not hypothetical — it is the current state of the `default` arm

`--arms all --pressure 8` on `origin/main` @ `4340bffc9`, 22 rows, counters separated by hand:

| arm | Σ `copied_objects` (scavenge) | Σ `moved_objects` (C4b) | harness's `moved` |
|---|---|---|---|
| `default` | **0** | 7 610 512 | 7 610 512 |
| `verify_evac` | **0** | 7 301 552 | 7 301 552 |
| `evac_minor` | 3 628 777 | 0 | 3 628 777 |

The first two rows read as "movement happened" in the arm-liveness summary. No object was moved by the collector under test. (The reason those arms run no scavenge at all is filed separately — it is the pressure knob colliding with the moving-defer hard cap.)

## Suggested direction

Report the two counters as separate columns and let each arm state which one satisfies it. A `requires=move` arm added for the evacuating scavenge should require `copied_objects > 0` specifically; `moved_objects > 0` is a fine liveness signal for a full-GC evacuation arm, but it is a different claim.

While there: the `collect` predicate has the same shape of weakness, already filed as #7017 (a teardown-boundary cycle that frees nothing is indistinguishable from a mid-program one).

Related: #7019, #6942, #6946, #6950, #7017.

PR #7644, opened 2026-08-08
test(gc): the POINTER_FREE trace-skip hazard probe that faults (#7635)

### LINKED ISSUE #7635: gc: forcing POINTER_FREE on a pointer-bearing object strands nothing — our zeal/protect instruments do not discriminate the layout-state hazard

proggeramlug:
Fixes #7635 — the probe that faults, and the reason nobody could build one.

## Root cause: the probes were vacuous, not the instruments

`JSON.parse` of a non-tiny blob is **lazy by default** (#7499's tape). Every probe in this investigation — the audit's two shapes and my reproduction — parsed, churned, and only then read the records back. The materialisation happens at read-back, **after** the collections ran: the cohort did not exist while the collectors were being tested, so no sabotage could strand it and no instrument could catch anything. The instruments were fine the entire time.

Demonstrated end-to-end on an instrumented build: the identical TS shape with **one pre-churn touch loop** (defeating the tape) faults immediately under the sabotage —

| arm | result |
|--|--|
| sabotage, default cycles | **SIGSEGV** (rc=139) |
| sabotage + `ZEAL=1 PROTECT_FROMSPACE=1` | **`[gc-fromspace-protect] FAULT: signal 10 at 0x47ab075bbd0`** — precise, named address |
| no sabotage (real code) | clean, and a trace-time audit confirms zero POINTER_FREE selections over pointer-bearing payloads |

A supporting measurement that sealed it: under sabotage with the *lazy* probe, a trace-time counter saw only ~5 pointer-bearing PF selections across dozens of cycles × 4,000 records — the records simply weren't there.

## The committed regression test

`gc/tests/layout_pointer_free_hazard.rs`, in the `fromspace_protect` planted-hazard convention, **no env knobs**:

- **Red control** — an object built through the materialiser's exact deferred-store path, whose only reference to a young string sits in a field slot, finalized with a lying `saw_pointer=false`. The copying minor moves the object, honors the state, skips the payload — asserted on **both halves of the strand**: the slot bits survive unrewritten (the skip was live), and the child sits in poisoned from-space (`QUARANTINE_POISON_WORD`).
- **Green twin** — truthful finalize: the child is evacuated and the slot rewritten.
- Subject-liveness asserted in both arms (the object must have MOVED), so neither can pass vacuously.

## Consequences for the record

- The `POINTER_FREE` trace-skip is **real, observable, and load-bearing** — outcome (3) of the issue's trichotomy, with the missing shape identified.
- #7633's "zeal arm: output identical" line was indeed non-discriminating for parse-only shapes, exactly as the audit said. The discriminating form now lives in CI.
- **Standing rule for this family**: any parse-cohort GC probe must defeat the lazy tape (touch the cohort before the collections) or state explicitly that it exercised the lazy path.

`perry-runtime` suite: 1,904 passed / 0 failed (both new tests green). `cargo fmt --check` and `check_file_size.sh` clean. No production code changed — this PR is the test alone.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added regression coverage for garbage collection of pointer-free layouts.
  * Verified that referenced objects are correctly evacuated and rewritten during collection, preventing stranded or invalid references.

* **Documentation**
  * Documented the pointer-free tracing hazard and guidance for ensuring garbage-collection probes exercise the intended behavior.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1365
+**Current Version:** 0.5.1366
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1365"
+version = "0.5.1366"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1365"
+version = "0.5.1366"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1365"
+version = "0.5.1366"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1365"
+version = "0.5.1366"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1365"
+version = "0.5.1366"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1365"
+version = "0.5.1366"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1365"
+version = "0.5.1366"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1365"
+version = "0.5.1366"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1365"
+version = "0.5.1366"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1365"
+version = "0.5.1366"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1365"
+version = "0.5.1366"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1365"
+version = "0.5.1366"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1365"
+version = "0.5.1366"
 dependencies = [
  "serde",
  "serde_json",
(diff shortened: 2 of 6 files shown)


proggeramlug:
## Audit — merged as v0.5.1366

Two agents worked #7635 independently and landed on the same root cause without contact: the probes were vacuous because `JSON.parse` is lazy for non-tiny blobs, so the record cohort was materialised *after* the last collection. That convergence is worth more than either report alone — it was the one conclusion I most doubted, since it means my own #7635 filing was wrong.

**Kept alongside #7643 rather than closed as a duplicate, because the two observe different layers:**

| | #7643 | #7644 (this) |
|---|---|---|
| observes | the **enumerator** — `gc_child_slots` reports zero payload slots | the **memory** — the stranded child sits in `QUARANTINE_POISON_WORD` from-space |
| catches | a misdeclared object being skipped | the same, *plus* a case where the enumerator reports correctly but the rewrite pass drops the slot anyway |

#7643 asks the collector what it intends to visit; this asks what actually happened after it ran. The second is the one that would survive a refactor that fixed the enumeration and broke the rewrite.

**Sabotage-verified two-sided, which the red-control design earns:** inverting the plant's `finalize_truthfully` reddens **both** arms — the lying arm fails "must not have been rewritten", the truthful arm fails "must have been rewritten to the evacuated copy" — with **0 compile errors**, so both are load-bearing on the actual layout state rather than on the construction. A one-sided check (neutering the finalize) would have left the red control green and proved only half of it.

Two design details I checked specifically:

- **The red control leaves a deliberately stranded child**, which in most tests would be a landmine for the next collection on that thread. It is safe here by construction — the object stays `POINTER_FREE`, so no collector will ever follow that slot — and `CopyingNurseryTestGuard::drop` calls `reset_shadow_stack()` regardless. Notably the *opposite* of #7643's cleanup, which restores `layout_mark_unknown`; doing that here would arm the landmine rather than defuse it. Both are right for their own shape.
- **Subject-liveness asserted in both arms** (`assert_ne!(moved_obj, obj)`). Without it a cycle that moved nothing would pass the "slot unchanged" assertion trivially — the #7025 failure mode exactly.

**Gates:** 20/20 from the `lint` job enumerated out of `test.yml`, `cargo fmt --all --check` clean, `cargo test -p perry-runtime --lib --no-fail-fast` **1908 passed / 0 failed** (1906 + these 2). No production code changed.

The standing rule in your writeup is the durable part and I've taken it into CLAUDE.md's orbit: **any parse-cohort GC probe must defeat the lazy tape by touching the cohort before the collections, or state explicitly that it is exercising the lazy path.** #7647 tracks the end-to-end CI gate — still open and still wanted, since both merged tests are unit-level and neither would catch a *new* materialiser path that forgets to finalize at all.

PR #7240, opened 2026-08-02
fix(gc): root every argument of a cross-module direct call — registry --help 28/30 → 30/30

jdalton:
Stacks on #7227 (branch `fix/7154-regexp-new-stale-register`), which stacks #7226. This branch is based on #7227's head (`28dc70d27`), so review those two first.

#7227 took `sfw-registry --help` from 0/30 to 28/30 under a genuine polls build and named its own residual: a deterministic `PERRY_GC_PROTECT_FROMSPACE` fault at `perry_fn_src_lib_api_shared_ts__defineApiCall + 428`, `obj_type=3` (a string), `retired_by_minor=#155`, 40/40 under the protector. This PR does that item. **`sfw-registry --help` is 30/30, three times over.**

## The diagnosis is a measurement, not a reading of the disassembly

The brief was to hunt an unrooted runtime-side string cache — the #7211 signature — because the static checker saw nothing of that shape in `defineApiCall`. That turned out to be the right observation about the checker and the wrong conclusion about where to look, so it is worth saying exactly how the answer was pinned down rather than guessed.

Under lldb the faulting instruction is `ldr w20, [x19, #0x4]` inside `js_regexp_test` — offset 4 of a `StringHeader` is `byte_len`, i.e. `string_as_str`. `x1` (the *subject* string, not the receiver) is the stale one, which already rules out `pattern_ptr` / `flags_ptr` / a regexp table. Walking up:

```asm
;; perry_fn_src_lib_api_shared_ts__defineApiCall
+396: ldr  d0, [sp, #0x28]              ; url, reloaded from shadow slot 0
+400: bl   js_jsvalue_to_string_coerce  ; pass-through for a string
+404: mov  x22, x0
+412: bl   js_gc_temp_root_get          ; #7227's receiver re-read
+424: bl   js_regexp_test               ; faults on x1 = x22
```

`[sp, #0x28]` is a properly bound shadow slot — the prologue does `js_shadow_frame_push(6)` then `js_shadow_slot_bind(0, sp+0x28)` — and its content at the fault **is** the stale address. So the argument arrived stale; nothing inside `defineApiCall` broke it.

One more frame up is `src_lib_api_alerts_ts__init_body`, and there the shape is unmistakable:

```asm
+752:  ldp  d9, d10, [x24, #0x18]              ; url + method, from their
                                              ; __perry_init_strings_* handle globals
+760…: js_gc_temp_root_push, js_object_alloc_class_inline_keys,   ; argument 3
[...]

Issue #7210, opened 2026-08-01
GC: the remaining unrooted-alloca hazards after #7207 — class-keys pointer caches, interleaved staging arrays, inlined-callee param slots


proggeramlug:
## Summary

The remaining `#7202`-class hazards after PR #7207, from the same bare-alloca enumeration (140 sites, 23 HAZARD / 117 SAFE). #7207 fixed the inline-constructor `this` slot; #7208 and #7209 own the closure `this`/`new.target` slots and the `catch` parameter. This tracks the rest so they stop being invisible.

## 1. An unrooted cached copy of a value whose *original* is a registered root (3 sites)

- **`function.rs:521` (`entry_init_load_global`)** — caches the `@perry_class_keys_*` pointer in an entry alloca for the whole function. `codegen/string_pool.rs:464-477` (#5042) explicitly says that array "lives in the longlived (old-gen) arena … **so old-page defrag (C4b) can relocate it**", which is why the *global* was registered with `js_gc_register_global_root`. This second copy silently opts out. Split by consumer: `new.rs:621`, `new.rs:705`, `scalar_method.rs:553` **dereference** it (passing it as `keys` to `js_object_alloc_class_inline_keys`) — real hazard; `method_override.rs:213` and `property_get/dynamic_dispatch.rs:105` only `icmp eq` it as a shape token — fail-safe.

  **This is the one violation `gc_root_dominance_check.py --unrooted-allocas` still reports** after #7207 on its reproducer corpus (not moving-reachable there, because `js_array_alloc` is the only call in that window).

- `stmt/loops.rs:1505` — a `double` cache of a module global aliased into `ctx.locals[gid]` for both copies of a packed-f64 range loop.
- `stmt/loops.rs:4766` — a hoisted `arr.length`; narrow, but the hoist classifier imposes no type gate, so a `Named`-typed receiver whose `length` is a string lands a pointer here.

## 2. Staging arrays filled *interleaved* with lowering (9 sites)

Shape: `for a in args { let v = lower_expr(ctx, a)?; store v into buf[i] }` — argument *i*'s NaN-boxed pointer is already sitting in the unrooted buffer while argument *i+1* is lowered, and lowering allocates.

[...]

PR #7161, opened 2026-08-01
fix(gc): disable evacuating minor by default pending #7154 (use-after-free on dynamically-added fields)

jdalton:
## Summary

Stopgap for #7154. The evacuating (moving-loop) minor GC that #7019 flipped **default-on** has a use-after-free that corrupts the heap **in the default configuration** — plain `sfw-registry --help` (no env) exits 1 with `TypeError: value is not a function`. This flips the default back to the previously-correct non-moving minor while the root cause is fixed under #7154. **This does not close #7154** (the collector bug is still open); it stops shipping a heap-corrupting default.

## The bug (diagnosis, tracked in #7154)

A young closure referenced from a **dynamically-added object field** (`field[1]`, holders built in `proxy::create_or_update_receiver_property` from zod's `$constructor`) is reclaimed by the evacuating minor while still live, so the field dangles and a later call dies with `TypeError: value is not a function`.

A live-trace probe on the evacuation path (temporary, reverted) established:

- **Enumerated-but-no-op** — field[1] (+40) **is** visited on the evac path; `visit_value_bits → classify` returns `None`, so the slot is never rewritten.
- classify is **correct**: the field[1] value is a genuine NaN-boxed pointer (`0x7ffd…`) into NurseryEden whose target header is **garbage in 100% of samples** (`plausible_arena=false`). It is a **dangling** pointer — the closure died in an earlier cycle. A "teach classify/`move_young` to accept it" fix would evacuate garbage → SIGSEGV, which is why the deep fix is left to #7154 rather than attempted here.

Full probe write-up: https://github.com/PerryTS/perry/issues/7154#issuecomment-5149872144

## The change

Flip **both** coherent gates from default-ON to default-OFF, keeping the flag as an opt-in:

- `gc_moving_loop_polls_enabled()` (runtime, `gc/policy.rs`)
- `moving_safepoint_polls_enabled()` (codegen, `stmt/loops.rs`)

Both read the same `PERRY_GC_MOVING_LOOP_POLLS` env and must agree (a runtime/codegen default mismatch would defer collections that never drain). Now only an explicit `PERRY_GC_MOVING_LOOP_POLLS=1`/`on`/`true` enables the moving-loop path (at **compile and run** time). No collector code is removed — this is a pure default flip. Trade-off: reverts #7019's minor-GC RSS/throughput win, not correctness.

## Verification

| check | result |
|---|---|
| default `sfw-registry --help` (no env), rebuilt on this branch | **clean 10/10** (was exit 1) |
[...]

PR #7648, opened 2026-08-08
refactor(codegen): Layer 1 rooting migration slice 5 — the timer and namespace-call lowerings (#7615)

proggeramlug:
Slice 5 of the Layer 1 rooting migration (#7615). Two modules of the `lower_call/` family migrated end to end, four live bugs found and fixed, and the reason the other four modules were left out recorded in the ledger because it is a statement about the API rather than about those files.

## Migrated

| module | escape-hatch sites before | ledger line |
|---|---|---|
| `lower_call/extern_timers.rs` | 6 | load-bearing (named `lower_exprs_rooted` / `temp_root_release`) |
| `lower_call/namespace_call.rs` | 8 | load-bearing (same) |

Sabotage arm run **per module** — the assert stops at the first offender, so one run cannot speak for two. A compiling `temp_root_push_double` / `temp_root_truncate` pair was planted in each; the ledger test went red and named both lines by `file:line`; the build output was checked for `error[` and `could not compile`, both **0**, and `Running unittests` confirmed the test binary was actually reached. (A plant that fails to compile also makes the command non-zero, and a check grepping only for `FAILED` scores that build error as a successful sabotage.)

## Live bugs

All four demonstrated against node 26.5.1 (the `.node-version` oracle) with a baseline compiler built from `main` in a separate target directory, and re-verified under `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800`, compiled with `PERRY_GC_MOVING_LOOP_POLLS=1`.

**1. Two-argument `setTimeout` / `setInterval` held the callback unrooted across the delay.** #7210 rooted the trailing-argument forms and left the two-argument siblings alone, though the comment it wrote names the window exactly. The delay is an arbitrary expression:

```llvm
%r6  = call i64 @js_closure_alloc(...)        ; the callback
%r8  = bitcast i64 %r7 to double              ; a BARE register
%r9  = call double @perry_fn_mod__churn()     ; the delay. User code, polls.
%r10 = call i64 @js_timer_validate_callback(double %r8, i32 0)   ; stale
```

```
node:      scheduled object / fired A
baseline:  TypeError [ERR_INVALID_ARG_TYPE]: The "callback" argument must be of
           type function. Received an instance of Object
fixed:     scheduled object / fired A
```

`setInterval` throws the same. A **capturing** callback is required — a non-capturing arrow lowers to `js_closure_alloc_singleton`, which is never moved, so the first reproducer I wrote passed while the IR clearly showed the window.

**2. The var-shaped namespace export dispatched through a stale closure.** `ns.arrow(churn())` fetches the closure from its zero-arg getter first (spec order — the callee reference before the arguments), holds it in a bare register across every argument's lowering, then `unbox_to_i64`s it. #7280 taxonomy (a) and (c) at once: `root_reload` could not have repaired it, because the pointer is derived *below* the window from a register captured *above* it.

```
node:      a:1 / b:2
baseline:  TypeError: value is not a function
fixed:     a:1 / b:2
```

**3. The `has_rest` namespace direct call lost every rest element — silently.** The #7154 accumulator shape verbatim: `current` was a raw `*mut ArrayHeader` threaded through a push loop while the next argument's expression ran, holding the only reference to everything pushed so far. `lower_rest_call_args_rooted` was written for exactly this and this path never adopted it.

```
node:      head|r1,r2,r3
baseline:  head|
fixed:     head|r1,r2,r3
```

A wrong answer, not a crash. Delegating to the audited helper also pads fixed parameters to the declared arity, which the hand-rolled loop did not.

**4. Three more unprotected windows**, repaired in passing, real by inspection but without a reproducer that faults today: the `fs/promises` `writeFile`/`appendFile`/`rmdir` arms (`path` held across `content` and `options`), both V8-bridge arms (a bare `for a in args { lower_expr }` — #7240's shape in a path that post-dates the fix), and the plain namespace direct-call argument loop.

## False positives, counted in emissions

The one-argument `setTimeout`/`setImmediate` and all three `clear*` arms lower a single operand and consume it in the very next emission. No emission between production and use means no window, so they stay on bare `lower_expr`; routing them through `with_operands_rooted` would emit nothing anyway (`any_may_trigger_gc` over an empty tail is `false`) and would advertise a protection that is not there.

## Cost

Zero where the window cannot collect, and now pinned rather than asserted. With a literal delay the emitted module contains **no temp-root traffic at all** and the callback register feeds `js_timer_validate_callback` directly. Across the whole 149-module `gc_root_dominance_corpus.sh` corpus, baseline and fixed IR are **byte-identical** (`diff -rq` → 0 differing files).

## Tests

`lower_call/timer_rooting_tests.rs`, asserting on emitted IR rather than runtime behaviour. The runtime fault needs a capturing callback, a polling delay *and* the compile-time `PERRY_GC_MOVING_LOOP_POLLS=1` (off by default since #7161) — a gap test would be green on the default build whether or not the fix is present, which is hazard 4.

The assertion is an **ordering**, not a slot count: a count across two programs that differ in an operand lets the delay's own rooting pay for the assertion. With an allocating delay the register `js_timer_validate_callback` reads must be defined *below* the delay's allocation; with a literal delay it must be the original register with zero temp-root traffic. Checked against the pre-fix source: the two ordering tests **fail**, the two zero-cost tests **pass**. Neither is vacuous. The liveness check excludes `declare` lines — the module always carries `declare @js_timer_validate_callback` whether or not anything calls it.

## Not migrated, and why

Four of the six modules named on the tracker are left, and the reason is one sentence about this API: three of them need a re-read at more than one point, and every `with_operands_rooted*` form has exactly one.

- **`lower_call/mod.rs`** — `lower_call_args_rooted` / `lower_rest_call_args_rooted` return a guard *deliberately*: their consumers in `func_ref.rs` are block-splitting specialized-ABI diamonds whose release must sit in a merge block post-dominating four dispatch paths, ~200 lines below the lowering. A closure form can express that only by swallowing the whole dispatch chain, in a file outside the slice.
- **`lower_call/new.rs`** — `refresh_rooted_args` re-reads the same operand group at three caller-chosen points, under a `temp_root_scope_begin`/`_end` marker spanning ~20 return paths.
- **`lower_call/console_promise.rs`** — `lower_dynamic_closure_call` re-reads receiver and callee below the arguments, then re-reads the arguments again below the allocating rebind unbox. Two stages, one combinator.
- **`lower_call/early_branches.rs`** — its only escape-hatch uses are `implicit_this_save`/`implicit_this_restore`, already a paired combinator rather than the raw ordering API. Migrating it means re-exporting that pair through `crate::rooting`: a rename that would make the ledger line look substantive while asserting nothing new.

The concrete missing combinator is the **variadic/rest shape** — per-element re-reads between allocating pushes. `namespace_call.rs`'s rest arm therefore delegates to `super::lower_rest_call_args_rooted`, which still names the raw API; that boundary is stated in the file header rather than hidden, and it is the same posture as calling `lower_expr`.

Two hazards in `console_promise.rs` were found while reading it and are **not** fixed here: `console.table(a, b)` and `console.dir(a, b)` both hold operand 0 in a bare register across operand 1's lowering. They are one-line fixes with the same combinator, but they belong with that module's migration rather than ahead of it.

## Gates

The lint list was **extracted** from `.github/workflows/test.yml`, not recalled — 20 commands, all run individually:

```
ran=20 failed=0
```

Plus:

- `rustup run stable cargo fmt --all -- --check` — clean
- `cargo test -p perry-codegen --lib --no-fail-fast` — 706 passed, 0 failed
- `cargo test -p perry-runtime --lib --no-fail-fast` — 1908 passed, 0 failed, 3 ignored
- `cargo check --all-targets` — clean
- `gc_root_dominance_check.py --self-test` — OK
- `gc_root_dominance_corpus.sh` — 129/129 sources, 149 `.ll`
- dominance mode (`--moving-only --seeded-violations 40`) — **40 planted, 40 caught, 0 MISSED**, rc=0
- `--unrooted-allocas --moving-only` — **0 violations**, rc=0

Run without `--moving-only` both arms report the same 171 non-moving leads, all one fingerprint (`js_object_alloc_class_inline_keys` → `js_gc_declare_typed_shape_layout`). A/B'd against a baseline corpus: **171 on both**, so they are pre-existing and untouched by this change; the gate uses `--moving-only` and sees 0 either way.

No version bump, no `CHANGELOG.md` edit; changelog fragment added under `changelog.d/`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability for timers, filesystem calls, namespace calls, V8 bridge operations, and variable-shaped exports during complex argument evaluation.
  - Preserved callback and argument values across operations that may trigger allocations.
  - Enforced the maximum supported argument count for closure exports.
  - Improved handling of rest-argument calls and filesystem arguments.

- **Tests**
  - Added coverage for timer callback preservation and validation behavior.
  - Confirmed literal delays avoid unnecessary temporary handling.


proggeramlug:
## Audit — merging as v0.5.1368

### Bug 3 confirmed independently, and it is worse than the fragment said

I built a baseline from `main` myself in a separate target dir and wrote my own repro shape. The baseline prints **completely empty stdout at exit 0 on the plain default build — no zeal, no protect, no `PERRY_GC_MOVING_LOOP_POLLS`**. Node prints `head|r1,r2,r3`; your build prints it. The argument churn allocates enough to guarantee a collection inside the accumulator window, so unlike the other three this one needs no instrumentation to arrange. I've promoted it to lead the fragment and recorded the knob-free reproduction, because it is the only one of the four a user could hit today without trying.

### Bugs 1 and 2: I could not reproduce the runtime fault — and you were right anyway

Your files, your commands, my `main` baseline: correct output. So I checked the thing that doesn't depend on what gets recycled, and the IR settles it:

```llvm
%r7  = call i64 @js_closure_alloc_with_captures_singleton(...)   ; the callback
%r9  = bitcast i64 %r8 to double                                  ; BARE register
%r10 = call double @perry_fn_timer2b_ts__churn()                  ; user code, collects
%r11 = call i64 @js_timer_validate_callback(double %r9, i32 0)    ; reads the register defined above the call
```

and the fixed arm stores `%r8` into a root slot and reloads it across the barrier. The window is there on baseline and closed by the fix — that is the durable evidence, and it is why the acceptance tests assert IR ordering rather than a runtime outcome.

**Your correction #3 is the one I'd want future slices to inherit: a window that does not fault is not a window that is absent — it is one whose victim happened to survive.** I've written that into the fragment next to the claim, along with the fact that review could not reproduce it. A changelog that says "the baseline throws X" where a reader following the instructions gets exit 0 costs more trust than the finding is worth; a changelog that says "here is the window, here is why the fault is arrangement-dependent" is checkable forever.

### Your two corrections to my brief — both verified, both mine to own

`--seeded-violations 40` is accepted and silently ignored in `--unrooted-allocas` mode: it prints no seeded line and no self-check runs. My brief's "both modes" was wrong; CI is consistent, since the workflow only passes it to the dominance step.

And `cargo test … 2>&1 > log` sends stderr to the *terminal*, so `grep -c 'error\['` on that log counts stdout only and returns 0 no matter what. That is the sabotage-verification check I have been insisting on, being vacuous in exactly the way it exists to prevent — the same shape as the `$?`-after-a-pipe bug I hit twice in this session. Your fix (redirect order, plus asserting `Running unittests` appeared) is better than the check I asked for, because it proves the plant compiled **and** the binary was reached.

I did briefly suspect the `--unrooted-allocas` arm had no proof it can fire. Wrong — it has its own planted/control fixture pair (`_SELFTEST_UNROOTED`/`_SELFTEST_ROOTED`); I had read only the first of two self-test functions. Recording it so nobody re-derives the same false alarm.

### Scope

2 of 6 modules is the right answer given what you found, and the reason is the valuable part: **three of the four you left need a re-read at more than one point, and every `with_operands_rooted*` form has exactly one.** That is a statement about the API, not about the files, and putting it in the ledger comment is where it belongs. `early_branches.rs` being a rename that would "make the ledger line look substantive while asserting nothing" is exactly the judgement this ledger needs to stay meaningful — a ledger padded with modules that never had a decision to make is worth less than a short honest one.

#7649 captures the missing variadic/rest combinator.

**Gates:** 20/20 from the `lint` job enumerated out of `test.yml`, `cargo fmt` clean, `cargo check --all-targets` clean, `perry-codegen --lib` 706 passed, `perry-runtime --lib` 1909 passed. Dominance corpus not re-run locally — you A/B'd it as byte-identical across all 149 modules, which is both the zero-cost proof and the reason the acceptance coverage had to be unit tests.

PR #7651, opened 2026-08-08
refactor(codegen): Layer 1 rooting migration slice 6 — the multi-point re-read scope (#7615)

### LINKED ISSUE #7649: console.table / console.dir hold operand 0 unrooted across operand 1's lowering (lower_call/console_promise.rs)
Found while auditing `lower_call/` for #7615 slice 5 (#7648). Filed rather than fixed: both belong with `console_promise.rs`'s own Layer 1 migration, which needs a combinator that does not exist yet (see below), and fixing them ahead of it would leave the module half-migrated — the state that let #7461's `expr/url_main.rs` hide from #7461 to #7617.

## The two windows

Both are `crates/perry-codegen/src/lower_call/console_promise.rs`, both #7280 taxonomy (c) — operand-to-operand.

**`console.table(data, properties)`** (~line 300):

```rust
let v = lower_expr(ctx, &args[0])?;
if args.len() == 2 {
    let props = lower_expr(ctx, &args[1])?;   // arbitrary user code
    ctx.block().call_void(
        "js_console_table_with_properties",
        &[(DOUBLE, &v), (DOUBLE, &props)],    // `v` is a bare register
    );
}
```

**`console.dir(obj, options)`** (~line 434): identical shape —

```rust
let v = lower_expr(ctx, &args[0])?;
let opts = if args.len() >= 2 { lower_expr(ctx, &args[1])? } else { undefined };
ctx.block().call_void("js_console_dir_with_options", &[(DOUBLE, &v), (DOUBLE, &opts)]);
```

`console.table(buildRows(), buildCols())` / `console.dir(makeObj(), makeOpts())` hold operand 0 across a real user call with back-edge polls. An evacuating minor relocates it and the formatter reads from-space.

Each is a one-line fix with `crate::rooting::with_operands_rooted`, and costs nothing when the second operand provably cannot collect (`OperandProtection::Reuse`), which is the `console.dir(x)` / `console.table(rows)` majority.

## A separate, non-GC bug in the same `console.dir` arm

`console.dir` lowers arguments 2..n **after** the call that produces its output:

```rust
ctx.block().call_void("js_console_dir_with_options", &[...]);
for a in args.iter().skip(2) {
    let _ = lower_expr(ctx, a)?;      // side effects sequenced AFTER the print
}
```

Node evaluates every argument before invoking `console.dir`, so `console.dir(x, y, sideEffect())` interleaves output differently. Not verified against the oracle yet — noting it here so whoever fixes the rooting window does not restructure the arm without noticing it.

## Why not now

`console_promise.rs` cannot be migrated end to end with today's API. `lower_dynamic_closure_call` re-reads its operand group at **two** points — receiver and callee below the arguments, then the arguments again below the allocating `js_closure_unbox_callee_checked_rebind` — and every `with_operands_rooted*` form has exactly one re-read point. The same gap blocks `new.rs` (three re-read points under a scope marker spanning ~20 return paths) and is why `mod.rs::lower_rest_call_args_rooted` still names the raw API: the variadic shape needs per-element re-reads between allocating pushes.

So the useful ordering is: add the multi-stage / variadic combinator to `crate::rooting`, then migrate `console_promise.rs`, and close these two as part of it.

Refs #7615, #7648, #7280, #7154.

proggeramlug:
Slice 6 of the Layer 1 rooting migration (#7615). Slices 1–5 are merged; #7648 was slice 5.

## The API gap first, because that was the assignment

Slice 5 left three `lower_call/` modules unmigrated and recorded why: *"three of
them need a re-read at more than one point, and every `with_operands_rooted*`
form has exactly one. The concrete missing combinator is the variadic/rest shape
(per-element re-reads between allocating pushes)."*

**The first half of that is right and the second half is not, and it matters.**
Only one of the three modules is variadic. Reading all three:

| module | why one re-read point cannot serve |
|---|---|
| `console_promise.rs`'s dynamic closure call | consumes the group in **two instructions** with an allocating step between them: receiver + callee feed `js_closure_unbox_callee_checked_rebind` (which clones a `this`-capturing closure), the arguments feed `js_closure_callN` **below** it |
| `mod.rs`'s `lower_rest_call_args_rooted` | re-read points are a **loop**, not a point — one per `js_array_push_f64` |
| `func_ref.rs` | the **release** must post-dominate four block-splitting specialized-ABI diamonds, ~450 lines below the lowering |

The common request is not "variadic". It is **one temp-root scope, re-readable
at any number of caller-chosen points**. The variadic case is that scope with an
accumulator array in it, which is why `RootedGroup` carries both operands and
arrays rather than there being a second type.

```rust
group.lower(ctx, expr, collects)?;      // lower and root; returns an INDEX, never the register
group.adopt(ctx, expr, &value, collects); // root a value the caller emitted itself
group.reread(ctx, i)?;                  // re-read HERE, as many times as needed
group.begin_array(ctx, cap) / push_array / read_array;
```

### Two entry points, argued rather than assumed

`with_rooted_group` owns the release like every other combinator in the file.
`open_rooted_group` hands the scope back — which the rest of `rooting.rs`
deliberately refuses to do — so it needs a justification, and the justification
is that **the two halves of guard mismanagement are not equally dangerous**:

* an **early or mis-ordered** release is a use-after-free (a truncate is a stack
  *cut*: truncating the wrong slot drops everything above it, which is how a
  saved receiver becomes the number `0`);
* a **forgotten** release is over-retention — the slot stays bound, the object
  stays live, the code is merely conservative.

`RootedGroup` removes the dangerous half **by construction, for both entry
points**: not `Clone`, `release` consumes it, and no way to obtain the slot
index. Escaping leaves only the safe half writable. That is strictly better than
the `Option<String>` slot index it replaces, which the caller could truncate
anywhere. Inverting control in `func_ref.rs` instead would not remove the hazard,
it would relocate it into a 450-line closure.

`implicit_this_save` / `implicit_this_restore` **moved** into `crate::rooting`
rather than being re-exported, so the pair has one spelling — two would be the
drift that produced #7114.

## Modules migrated

Three, all **load-bearing on the committed source** (each named `expr::temp_root`
before this change): `lower_call/mod.rs`, `lower_call/func_ref.rs`,
`lower_call/console_promise.rs`.

`early_branches.rs`, `method_override.rs` and both `property_get` dispatchers
are now clean of the escape hatch as a side effect of the `implicit_this` move,
and are deliberately **not** listed: a ledger line asserts that a module makes
every rooting decision through this API, and nobody has read those four for
windows with no decision at all. That is slice 4's "listed ≠ audited"
distinction.

## Live bugs

All three behavioural ones are byte-for-byte A/B'd against node 26.5.1 (the
`.node-version` oracle) with a baseline compiler built from `main` in a separate
target dir, same `PERRY_GC_MOVING_LOOP_POLLS` on both arms.

```ts
console.dir(obj, { depth: 0 }, side("dir3"));
console.time("t", side("time2"));   console.timeEnd("t");
console.table([{ a: 1 }], ["a"], side("table3"));
```

| | node | `main` | this branch |
|---|---|---|---|
| `console.dir(x, y, f())` | `side:dir3` **then** `{ a: 1 }` | `{ a: 1 }` **then** `side:dir3` | matches node |
| `console.time(l, f())` | `side:time2` | *(never printed)* | matches node |
| `console.table(a, b, c)` | renders the table | `[ { a: 1 } ] [ 'a' ] 1` | matches node |

`diff` of node's output against this branch's is **empty**; against `main` it is
four hunks.

1. **`console.dir` sequenced side effects after its own print** (#7649's non-GC
   arm, which the issue flagged as unverified — now verified). It lowered
   `args[2..]` *below* `js_console_dir_with_options`. Node evaluates the whole
   argument list before invoking anything.
2. **`console.time` / `timeEnd` / `timeLog` / `count` / `countReset` dropped
   their surplus arguments entirely** — not resequenced, never lowered. A new
   find, same root cause.
3. **`console.table(a, b, c)` stopped being `console.table`.** The arity gate was
   `args.len() == 1 || args.len() == 2`, so three arguments fell through to the
   generic multi-arg `console.log` arm. Node ignores surplus arguments; it does
   not switch renderer.
4. **#7649's rooting half**, demonstrated in IR rather than at runtime. With
   `console.table(makeRows(), [churn(300)])`, `main` emits

   ```llvm
   %r1 = call double @perry_fn_..._makeRows()          ; operand 0, a bare register
   %r2 = call double @perry_fn_..._churn$spec_i32(300) ; user code: allocates, polls
   call void @js_console_table_with_properties(double %r1, double %r22)
   ```

   `root_reload` structurally cannot repair this — a call result has no slot to
   be re-read from (#7280 taxonomy (c) + (d)). This branch emits the root store
   above `churn` and reloads below it.

   **The runtime fault is arrangement-dependent and I could not reproduce one.**
   Under `PERRY_GC_MOVING_LOOP_POLLS=1` + `PERRY_GC_ZEAL=1` (`copied_objects=6005`
   confirmed via `PERRY_GC_DIAG=1`, so the subject was live), plus
   `PERRY_GC_PROTECT_FROMSPACE=1 …_DEPTH=800`, `main` printed the correct table
   and exited 0. The IR is the evidence the window exists; a `TypeError` would
   only have been evidence it is reachable in one arrangement. Stated here rather
   than claimed away.

## Zero cost where the window cannot collect — measured, not asserted

A probe exercising direct calls, rest calls, `arguments`, `new`, method dispatch,
dynamic closure calls, `Map.set` and eight `console.*` arms, compiled both arms
with `--trace llvm`. Normalising SSA numbering and block labels, the **entire**
semantic delta is **35 removed lines and 0 added**: two unconditional
`temp_root_push_double` / re-read / clear sequences that were protecting values
`expr_is_known_non_pointer_shadow_value` proves are not heap references — a
`double` from `b.scale(2)` and the literal `true` in `console.assert`. Both arms
bypassed the shared `operand_protection` decision; routing them through it drops
the traffic. Corpus-wide the root-store count goes 9846 → 9799.

Runtime output over the same probe: `main` and this branch identical.

## Tests

`lower_call/console_rooting_tests.rs`, five tests, asserting on IR **ordering**
rather than slot counts (a count lets the *other* operand's rooting pay for the
assertion). Each asserts by callee name that the arm under test was reached, so
a shape measured over a lowering that never ran cannot pass.

**Sabotage-verified against the pre-fix source**, with the file reverted to its
`HEAD` content and only the moved `implicit_this` path repointed so it still
compiles: `error[` count **0**, `Running unittests` present (so the plant
compiled *and* the binary ran), **4 of 5 red**. The fifth is the zero-cost pin,
which correctly passes in both arms.

**Ledger sabotage, one arm per newly-listed module** — a compiling
`temp_root_push_double` / `temp_root_truncate` pair planted in each:

| module | `error[` | binary ran | ledger red | names the lines |
|---|---|---|---|---|
| `lower_call/mod.rs` | 0 | yes | yes | yes |
| `lower_call/func_ref.rs` | 0 | yes | yes | yes |
| `lower_call/console_promise.rs` | 0 | yes | yes | yes |

## Gates

All 20 lint-job commands enumerated from `.github/workflows/test.yml` — **20/20
pass**. Plus `cargo fmt --all -- --check`, `cargo test -p perry-codegen --lib`
(711 pass), `cargo test -p perry-runtime --lib` (1909 pass, 3 ignored),
`cargo check --all-targets` (clean).

`scripts/gc_root_dominance_check.py` over the real
`gc_root_dominance_corpus.sh` corpus (149 modules, 2452 functions), in the
gate's own mode:

```
--moving-only --allowlist … --seeded-violations 40
  main:        0 violations, 9846 root stores
  this branch: 0 violations, 9799 root stores
  seeded: 40 planted, 40 caught, 0 MISSED
```

`--unrooted-allocas`: 6 violations, 0 moving-reachable, **identical in both
arms** (the known #7210 residue).

> ⚠️ **`--seeded-violations` is unreachable without `--moving-only`.** The
> seeded arm runs only after the real check returns 0, and the un-filtered mode
> reports 171 violations (all non-moving, the
> `js_object_alloc_class_inline_keys → js_gc_declare_typed_shape_layout` class)
> on `main` and on this branch alike. So a run without `--moving-only` never
> exercises the "can this gate still fail?" arm and prints nothing about it —
> the same shape as CLAUDE.md's hazard-4 note about `--unrooted-allocas`
> accepting the flag and silently doing nothing. Not fixed here; recorded.

Gap tests A/B'd (node vs `main` vs this branch): 7 console tests and 15
call-shape / GC-rooting tests. All `base-vs-fix: same`, all `node-vs-fix:
match`, except `test_gap_console_methods`, which differs from node in *both*
arms by timer-microsecond values only.

## Campaign note: the "24 modules" number is misleading

`rg -l "temp_root::" crates/perry-codegen/src` reports 24, but **14 of them make
no rooting decision**: five only construct `TempRootPool::default()`, three only
call a predicate (`expr_may_trigger_gc`, `expr_is_inert_primitive`,
`operand_is_reloadable`), and six only used the `implicit_this` pair this PR
moved. After this change the modules with real un-migrated rooting decisions are:
`lower_call/new.rs`, `expr/child_proc.rs`, `expr/dyn_extern_i18n.rs`,
`expr/fs_await.rs`, `expr/math_simple.rs`, `expr/proxy_reflect.rs`,
`expr/static_field_meta.rs`. `new.rs` is the big one and now unblocked —
`RootedGroup` is exactly its `refresh_rooted_args` shape — but it is 1988 lines
against the 2000-line cap, so it wants its own slice.

Closes #7649.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed evaluation order for console methods so all arguments are evaluated reliably.
  - Corrected `console.table` behavior across supported argument counts.
  - Preserved side effects from extra console arguments.
  - Improved reliability of function, method, and cross-module calls during memory allocation.
  - Fixed argument handling for `Promise.try` and `Array.fromAsync`.

- **Documentation**
  - Added release documentation covering migration details, validation results, and known runtime-gate observations.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1368
+**Current Version:** 0.5.1369
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1368"
+version = "0.5.1369"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "perry-ffi",
  "scraper",
@@ -5813,7 +5813,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-commander"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5821,7 +5821,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cron"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "chrono",
  "cron",
@@ -5831,7 +5831,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dayjs"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "chrono",
  "perry-ffi",
@@ -5839,7 +5839,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-decimal"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "perry-ffi",
  "rust_decimal",
@@ -5847,7 +5847,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dotenv"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "perry-ffi",
  "serde_json",
@@ -5855,7 +5855,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ethers"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "perry-ffi",
  "rand 0.10.1",
@@ -5863,7 +5863,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-events"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5871,14 +5871,14 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-exponential-backoff"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "perry-ffi",
 ]
 
 [[package]]
 name = "perry-ext-fastify"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "bytes",
  "http-body-util",
@@ -5896,7 +5896,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-fetch"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "bytes",
  "lazy_static",
@@ -5909,7 +5909,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-http"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "bytes",
  "h2",
@@ -5933,7 +5933,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ioredis"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "lazy_static",
  "perry-ffi",
@@ -5943,7 +5943,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-jsonwebtoken"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "base64",
  "jsonwebtoken",
@@ -5954,7 +5954,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-lru-cache"
-version = "0.5.1368"
+version = "0.5.1369"
 dependencies = [
  "lru",
  "perry-ffi",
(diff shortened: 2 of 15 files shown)


proggeramlug:
## Follow-up commit: two more windows, found by auditing the arms the migration did not have to touch

`console_promise.rs` is now listed in the ledger, and slice 4's "listed ≠ audited"
note says a ledger line means *"makes no ordering mistake against the raw API"*,
not *"has no rooting bugs"*. So I read the rest of the module. Two windows with
no rooting decision at all (#7640's class), both now fixed:

**1. `Promise.try(cb, ...extra)` — #7154's accumulator shape verbatim.**

`current_arr` was a raw `*mut ArrayHeader` threaded through the push loop in a
bare SSA register, holding the only reference to everything pushed so far while
the next argument's expression — arbitrary user code — was lowered. `callback`
sat in a second bare register across `js_array_alloc`, every push, and every one
of those lowerings. This is the same defect slice 5 found in
`namespace_call.rs`'s rest path and called "the most serious of the four".

Compiled from `HEAD` of this branch's first commit vs the fix, same compiler
build otherwise:

```llvm
; before
%r55 = invoke i64 @js_array_alloc(i32 3)
%r58 = invoke double @..._churn(...)                 ; user code: allocates, polls
%r60 = invoke i64 @js_array_push_f64(i64 %r55, double %r58)   ; %r55 from above the churn
%r63 = invoke double @..._churn(...)
%r65 = invoke i64 @js_array_push_f64(i64 %r60, double %r63)   ; %r60 from above the churn
%r68 = invoke double @..._churn(...)
%r70 = invoke i64 @js_array_push_f64(i64 %r65, double %r68)   ; %r65 from above the churn

; after — the accumulator is re-read from its root between every push
%r68 = invoke double @..._churn(...)
%r70 = <load from the root slot>
%r71 = invoke i64 @js_array_push_f64(i64 %r70, double %r68)
%r78 = invoke double @..._churn(...)
%r80 = <load from the root slot>
%r81 = invoke i64 @js_array_push_f64(i64 %r80, double %r78)
```

**2. `Array.fromAsync(input, mapFn, thisArg)`** held each operand in a bare
register across the others' lowering — #7280 taxonomy (c), which `root_reload`
structurally cannot repair. One re-read point serves (all three feed the single
`js_array_from_async`), so this is the plain `with_operands_rooted` form.

**Neither faults at runtime in the arrangements tried** — `PERRY_GC_ZEAL=1` on a
`PERRY_GC_MOVING_LOOP_POLLS=1` build prints `x|y|z` / `2,4,6` and exits 0 in both
arms, matching node. The IR ordering is the evidence, per the same reasoning as
#7649's rooting half above.

**`Array.fromAsync` deliberately keeps `take(3)`, so this stays a pure rooting
fix.** While reading these arms I found that `Promise.resolve` / `reject` /
`all` / `race` / `allSettled` / `any` and `Array.fromAsync` all **fail to
evaluate their surplus arguments** — `Promise.resolve(x, sideEffect())` never
runs `sideEffect`. That is the same node-visible defect as `console.time`'s
above, across seven more arms, and it wants its own change with its own oracle
A/B rather than being smuggled into a rooting slice. Recorded in the changelog
fragment.

Gates re-run after this commit: 20/20 lint commands, `cargo test -p perry-codegen
--lib` 711 pass, `gc_root_dominance_check.py --moving-only --seeded-violations
40` → 0 violations, 40/40 seeded caught.

proggeramlug:
## Audit — merging as v0.5.1369

### Three node-visible bugs, all reproduced independently

Built a baseline from `main` myself in a separate target dir, wrote my own repros, A/B'd against node 26.5.1:

| | `main` | this PR | node |
|---|---|---|---|
| `console.dir({a:1},{depth:0},f())` | `{ a: 1 }` **then** `SIDE EFFECT` | `SIDE EFFECT` then `{ a: 1 }` | matches the PR |
| `console.time("t", g())` | **no side effect at all** | `SIDE EFFECT g` | matches the PR |
| `console.table([{a:1}],["a"],1)` | `[ { a: 1 } ] [ 'a' ] 1` | the actual table | matches the PR |

The second is the one I'd have missed reviewing this by eye: the argument was never lowered, so the *evaluation* silently vanished. And the third is a good demonstration that an arity gate is a behavioural decision — `== 1 || == 2` quietly turned `console.table` into `console.log` for every three-argument call.

### The combinator, and slice 5's hypothesis being wrong

Slice 5 said the gap was "the variadic/rest shape". You checked and only one of its three blocked modules is variadic — the other two want re-reads at *two instructions* and at *four block-splitting diamonds ~450 lines apart*. Generalising to "one temp-root scope, re-readable at any number of caller-chosen points, released once" is the right abstraction, and it is better for having been derived from all three rather than from the one that suggested itself.

The two-entry-point argument is the part I'd have pushed back on if it were asserted rather than argued: an early or mis-ordered release is a use-after-free, a forgotten one is over-retention, so `RootedGroup` killing the dangerous half by construction (not `Clone`, `release` consumes, no way to reach the slot index) is what makes `open_rooted_group` acceptable. Inverting control in `func_ref.rs` would have relocated the hazard into a 450-line closure, not removed it.

Declining to list `early_branches.rs`, `method_override.rs` and the two `property_get` dispatchers — escape-hatch-clean as a side effect of the `implicit_this` move, but unaudited for windows with no decision — is exactly right. A ledger padded with modules that never had a decision to make is worth less than a short honest one.

### Verified independently

**Ledger sabotage:** I planted a compiling `temp_root_push_i64` call in all three newly-listed modules. `error[` count **0**, `Running unittests` present, ledger FAILED naming the planted lines. (My first attempt didn't compile — the check caught it, which is the whole point of the redirect fix you found.)

**Gates:** 20/20 from the `lint` job enumerated out of `test.yml`, `cargo fmt` clean, `cargo test -p perry-codegen --lib` **711 passed**.

**Cost:** 35 lines removed, 0 added, corpus root stores 9846 → 9799 — two *unconditional* roots on provably non-pointer values that bypassed `operand_protection`. A migration that removes work while closing windows is the outcome this API was designed for.

### Your three corrections, all verified

1. **"24 modules" was my number and it was misleading.** 14 make no rooting decision. Real remaining: `new.rs` plus six `expr/` modules. I'll use your list.
2. **`--seeded-violations` is inert without `--moving-only`** — confirmed by reading `main()`: the seeded arm sits after `if remaining: return 1`, so a red check skips its own self-test, and un-filtered the corpus is red on `main` anyway. CI passes `--moving-only` so it is consistent there; the hazard is that a *human* running it the obvious way gets no self-check and no warning. Broader than the correction I made after slice 5.
3. **The stale `Cargo.lock` was mine** — the #7646/#7648 bumps staged the lock without a cargo invocation between the `sed` and the `git add`, so all 76 workspace entries kept the old version. Fixed on `main` directly, and I now regenerate before staging.

`new.rs` at 1988/2000 lines being its own slice is right, and the `Promise.*` / `Array.fromAsync` surplus-argument evaluation bug is correctly a separate change.

Issue #7493, opened 2026-08-06
Integration suites assert the shadow-stack lowering but NativeRootsPin is #[cfg(test)] — shadow_slot_hygiene 0/12 red since #7370


proggeramlug:
## Symptom

Several `perry-codegen` integration suites are red on `main` (f05ae3b83), arm64 macOS, debug:

| suite | default |
|---|---|
| `tests/shadow_slot_hygiene.rs` | **0 passed, 12 failed** |
| `tests/scalar_replaced_slot_roots.rs` | 2 passed, **9 failed** |
| `tests/temp_root_operand_temporaries.rs` | 12 passed, **7 failed** |
| `tests/temp_root_argument_temporaries.rs` | 3 passed, **4 failed** |
| `tests/native_proof_regressions.rs` | 249 passed, 4 failed (3 of them this cause) |

Every one of these failures **reproduces when the test is run alone**, so this is not order dependence and it is unrelated to #7490 (whose cascade was a poisoned mutex in `typed_feedback.rs`, fixed in #7492).

## Cause

#7370 made native roots (statepoints) the default lowering. These suites assert on the **shadow-stack** lowering — they used to get it by accident, because there was only one default. The failure text says so directly:

```
box pointer slot should be stored as i64 before helper edges:
call i64 @js_box_alloc_bits(i64 9222246136947933185)
  %rs4gc.s3 = inttoptr i64 %r1 to ptr addrspace(1)
```

That much was already understood: `codegen/helpers.rs` grew `NativeRootsPin::shadow()` for exactly this, with the comment *"a test that asserts on shadow-stack IR has to SAY so … Eight tests broke on exactly this when the default flipped, and every one of them was correct about what it asserted."*

**But `NativeRootsPin` is `#[cfg(test)]`**, so it is compiled only into the crate's own unit-test build and is unreachable from `tests/*.rs`. The in-crate unit tests were repaired; the integration suites had no pin to reach for and were left red. They do not run per-PR (nightly/tag only — the CLAUDE.md pitfall), so nothing turned red on the merge.

## Evidence that the lowering is the variable

Full `--no-fail-fast -- --test-threads=1` sweeps of `-p perry-codegen --tests`, default vs. the process-global override:

```
suite                                   default        PERRY_RS4GC=0
shadow_slot_hygiene                    0 / 12 pass     11 / 12 pass    ← the clean signal
scalar_replaced_slot_roots             2 / 11 pass      5 / 11 pass
[...]

Issue #6968, opened 2026-07-29
gc: scalar-replaced object/array locals holding heap values are not precise roots


proggeramlug:
Found while fixing #6951. Distinct defect, same symptom class.

With the conservative native-stack scan disabled, a heap value stored into a **scalar-replaced** object or array local is swept out from under the alloca that holds it. The alloca has no shadow-slot binding, so it is not a precise root.

## Repro

```ts
let sink: unknown[] = [];
function churn(n: number): number {
  let acc = 0;
  for (let i = 0; i < n; i++) {
    sink.push({ i: i, s: "x" + (i & 255), a: [i, i + 1] });
    if (sink.length > 4096) { acc = (acc + sink.length) | 0; sink = []; }
  }
  return acc | 0;
}
function fresh(k: number): string { return "f" + k + "-" + (k * 7); }
const N = 420000;
// Repeat 6x -- the failure is a use-after-free, so the freed block has to be
// recycled before it shows. A single statement passes.
{ const o = { a: fresh(0), b: churn(N) }; console.log(o.a, o.b); }
{ const o = { a: fresh(1), b: churn(N) }; console.log(o.a, o.b); }
{ const o = { a: fresh(2), b: churn(N) }; console.log(o.a, o.b); }
{ const o = { a: fresh(3), b: churn(N) }; console.log(o.a, o.b); }
{ const o = { a: fresh(4), b: churn(N) }; console.log(o.a, o.b); }
{ const o = { a: fresh(5), b: churn(N) }; console.log(o.a, o.b); }
```

```
$ PERRY_CONSERVATIVE_STACK_SCAN=off PERRY_GC_HEAP_LIMIT=8 ./repro
f0-0 417894
f1-7 421991
f2-14 417894
 421991        <- o.a is empty
 417894
 421991
```

The array-literal form is identical: `const a = [fresh(k), churn(N)]; console.log(a[0], a[1]);`.

## Why

There is no object. The literal is scalar-replaced, so `o.a` / `o.b` are plain entry-block allocas:

```llvm
%r13 = call double @..._fresh__spec_i32(i32 0)
store double %r13, ptr %r10                   ; <-- o.a, a bare alloca
%r16 = call double @..._churn(double %r15)    ; collects; %r10's string is swept
store double %r16, ptr %r11                   ; <-- o.b
```

[...]

Issue #5960, opened 2026-07-04
CI: PR cargo-test never executes crates/perry integration suites; main-push full run starves under merge trains (documented near-miss)


proggeramlug:
Documented near-miss from 2026-07-04: PR CI's change-scoped `cargo-test` (`scripts/ci_test_scope.py`) runs **unit tests only** — `Running unittests src/lib.rs` lines exclusively in the job log. The `crates/perry/tests/*.rs` integration/e2e suites are *compiled* (they appear in the changed-file scope listing) but **never executed** on the PR path. They only run on the full scope (push to main / `--full`).

Consequences observed today:
- PR #5938 landed with its own acceptance suite (`capture_rereg_renamed_class.rs`) failing — `same_class_name_across_factories_keeps_snapshots_separate` was red at the very commit that introduced it, through green required checks. (Fix in flight: the static-call/static-field lowering arms bypass `class_renames`.)
- The intended safety net — the full run on main push — was cancelled repeatedly by the per-branch concurrency group during a rapid merge train (three squash-merges within ~20 minutes each superseding the previous run), so main went ~a workday with no completed full `cargo-test`.

This is a *design trade* (e2e suites compile binaries and take tens of minutes; scoping keeps PR latency sane), not a bug per se. Suggested cheap hardenings, any subset:
1. **Merge-train discipline** (process, zero infra): after an admin merge, let the main-push full run complete before the next merge lands — or re-dispatch it after the last merge of a batch. (I'm adopting this immediately for my own merges.)
2. **Concurrency carve-out**: give the main-branch full run `cancel-in-progress: false` (queue instead of cancel) so merge trains can't starve it.
3. **Scoped e2e tier**: when the changed files include `crates/perry/tests/<suite>.rs`, run *those suites* (they're the PR's own acceptance tests — the exact class of miss that happened here).
4. A nightly full run as a backstop (if not already present).

Refs: #5938 (the near-miss PR), the fix PR for the renamed-class static dispatch, run 28698034143 (an old-commit re-run that was the only completed "full" signal today, itself red on the known load-sensitive byob stall).

proggeramlug:
[...]

Issue #32, opened 2026-04-15
Module-level state is duplicated per importer (breaks plugin / type registry / DI patterns)


proggeramlug:
## Summary

Perry's AOT module loader appears to give every importer of a module its own private copy of that module's top-level state. This breaks the singleton-via-module pattern that the JS/TS ecosystem relies on for plugin registries, dependency injection containers, type registries, instance caches, etc.

In standard ES module semantics, the same physical file resolves to one module instance regardless of how many import paths reference it (`./registry`, `../registry`, `./types/registry` should all canonicalise). On Perry the registry's mutable state ends up duplicated.

## Repro

Hit while implementing the OID → codec registry in `@perry/postgres`. Distilled to the essentials:

```ts
// src/types/registry.ts
const REGISTRY_OIDS: number[] = [];
const REGISTRY_CODECS: unknown[] = [];

export function registerType(oid: number, codec: unknown): void {
    REGISTRY_OIDS.push(oid);
    REGISTRY_CODECS.push(codec);
}

export function getCodec(oid: number): unknown {
    const idx = REGISTRY_OIDS.indexOf(oid);
    return idx < 0 ? undefined : REGISTRY_CODECS[idx];
}

export function decodeValue(oid: number, buf: Buffer): unknown {
    const c = getCodec(oid);
    return c === undefined ? buf.toString('utf8') : (c as any).decode(buf);
}
```

```ts
// src/register-defaults.ts  (sibling of connection.ts under src/)
import { registerType } from './types/registry';
export function registerDefaultCodecs(): void {
    registerType(16, /* boolCodec */);
    registerType(25, /* textCodec */);
    // ... ~40 more
}
```

```ts
// src/connection.ts
import { decodeValue } from './types/registry';
import { registerDefaultCodecs } from './register-defaults';

export function connect(...): Promise<Connection> {
    registerDefaultCodecs();          // pushes ~40 entries
    // ... later, when decoding rows:
    decodeValue(oid, buf);            // sees REGISTRY_OIDS.length = 0
}
```

Note: `register-defaults.ts` and `connection.ts` import the registry under the **same** path string (`'./types/registry'`) — they live in the same directory.

## Observations (instrumentation traces)

Added counters and `console.log` to confirm:

1. `registerDefaultCodecs()` runs and the 42 `registerType()` calls execute.
2. From `register-defaults`'s perspective, after registration: `REGISTRY_OIDS.length === 42`.
[...]

PR #7653, opened 2026-08-08
test(gc): cover the root lowering that actually ships — native-roots assertions for #7502

### LINKED ISSUE #7502: The shipped root lowering (native/statepoints) has no coverage in the suites #7493 pinned to shadow — 9 mechanics, 3 of them found passing vacuously
Split out of #7493, which restored these suites' assertions by pinning them to the lowering they were written against. That was the right repair — but it leaves a hole that "the tests are green again" hides, and this issue is that hole.

## The problem

Since #7370, **native roots (RS4GC statepoints) are the lowering that ships** on aarch64 and x86_64 — every target whose frames the runtime can walk. The shadow stack survives only as the fallback for `arm64_32` watchOS and ARM64 Windows.

After #7493, `shadow_slot_hygiene` (12 tests) and `scalar_replaced_slot_roots` (11 tests) pin `NativeRootsPin::shadow()`. Every assertion in both files is about a mechanism the shipped configuration does not use:

| mechanic asserted (shadow) | native-roots equivalent | covered today? |
|---|---|---|
| a pointer-typed local reserves a frame slot | the local's alloca is `ptr addrspace(1)` and appears in the stack map | **no** |
| a dead value's slot is cleared before the next allocation | the value is not live at the statepoint, so it is not in that statepoint's live set | **no** |
| a numeric local reserves NO slot | a numeric local's alloca stays `ptr` / `double`, out of the map | **no** |
| slot indices are not shifted by an interleaved numeric local | (no indices exist; the map is keyed by frame offset) | n/a |
| the entry-module frame starts AFTER the init prelude | the entry function's first statepoint is after `js_gc_init` | **no** |
| a loop body's slots are cleared each iteration | per-back-edge statepoint live sets do not carry the previous iteration's value | **no** |
| a scalar-replaced field holding a heap value is a precise root (#6968) | that field's alloca is `addrspace(1)` and relocated | **no** |
| a scalar-replaced numeric-only literal pays NO rooting (#6997) | its allocas stay non-`addrspace(1)` | **no** |
| duplicate `var` declarations keep every slot inside the frame (#7184's shape) | (no frame bound exists; the #7184 shape is unrepresentable) | n/a |

So: **nine distinct root-lowering mechanics have zero assertions against the lowering Perry actually emits.** Six of the nine are the exact shapes `docs/src/internals/gc-rooting-invariant.md` lists as having already shipped broken.

`gc-root-dominance.yml` is not a substitute. It reads emitted IR for *store dominance* — it answers "is this root store ordered correctly", not "is this value a root at all". A value the statepoint lowering silently declines to put in the map has no store for it to check.

## Why it matters more than the usual "add tests" issue

Both #7493's own findings are instances of this:

* `numeric_only_scalar_replaced_{object,array}_emits_no_rooting` asserted `bind_calls(&ir) == 0`. Under the native default that count is zero for *every* program — the tests were passing vacuously, i.e. CLAUDE.md hazard 4.
* `a_collection_free_construction_emits_no_this_slot_root` asserted `!contains("@js_shadow_slot_bind")`, likewise true of every program.

Three tests that read as coverage were measuring nothing. There is no reason to think those are the only three; they are the three that happened to be adjacent to a lowering pin.

## What to build

A native-roots counterpart for each row above. The load-bearing design question is what to assert against, since the statepoint form has no bind call to count:

1. **Pre-`opt` IR** — the `ptr addrspace(1)` alloca set and the `gc "statepoint-example"` attribute are visible in what `compile_module` returns today. Cheapest; proves the *request*, not the result.
2. **Post-RS4GC IR** — `gc.statepoint` / `gc.relocate` and each statepoint's live-value list. Proves the result. Needs the in-process pipeline (#7301) to be reachable from a test.
3. **The emitted `__perry_gcmap` section** — `crates/perry-codegen/src/gc_map.rs` already builds it; assert on the parsed map. Strongest, and it is what the runtime actually reads.

(3) is the one that can catch "the map says nothing lives here". Prefer it where the mechanic is about *which values are roots*; (1) is fine for the negative direction (a numeric local must never become `addrspace(1)`).

## Acceptance

* Each **no** row above has at least one native-roots assertion.
* Each new test is sabotage-checked: with the rooting deliberately dropped it must fail. A negative assertion in particular must be shown to be non-vacuous under the lowering it runs on — that is the specific mistake this issue exists to stop repeating.
* The shadow-pinned originals stay. Both lowerings are supported; both need coverage. The point is not to replace them.

proggeramlug:
Closes #7502.

## What was missing

Since #7370 native roots (RS4GC statepoints) are the lowering that **ships** on
every target whose frames the runtime can walk. #7493 correctly repaired
`shadow_slot_hygiene` and `scalar_replaced_slot_roots` by pinning them to the
shadow stack — but that repair made explicit that **nine root-lowering mechanics
had no assertion at all against the lowering Perry emits**, and that three tests
which read as coverage were measuring nothing: they asserted
`js_shadow_slot_bind` was absent, which is true of *every* program under the
native default (CLAUDE.md hazard 4).

The shadow-pinned suites stay. Both lowerings are supported and both need
coverage; this is the missing half, not a replacement.

## What this adds

`crates/perry-codegen/src/native_root_coverage/` — **8 mechanic tests + 5
harness self-tests**, in-crate `#[cfg(all(test, feature = "llvm-inprocess"))]`
so they run in the per-PR `cargo-test` gate rather than the nightly-only
`tests/*.rs` tier (#5960).

Assertions are made at three vantages, each catching what the previous one
cannot:

1. **Pre-`opt` IR** — the `ptr addrspace(1)` allocas codegen *asks* for. The only
   vantage where "codegen never requested a root" and "LLVM removed one" are
   still distinguishable, which is why the negative claims live here.
2. **Post-RS4GC IR** — each `gc.statepoint`'s `"gc-live"` operand bundle, keyed
   by callee name, produced by running the **production** pass string
   (`inprocess::STATEPOINT_REWRITE_PASSES`). Turns "is this value a root across
   *that* call" into a direct question. The shadow suites never had an equivalent.
3. **The emitted stack map** — the per-safepoint root lists decoded back out of
   the compact `__perry_gcmap` blob the collector actually reads
   (`gc_map::decode_stack_map_roots`, which round-trips through `encode_stream` +
   `verify_roundtrip`, so a test asserts on what the binary ships).

Both shipped native-roots targets (`arm64-apple-macosx`, `x86_64-unknown-linux-gnu`)
are compiled and emitted for on every host, pinned rather than host-derived:
`cargo-test` runs on x86_64 Linux and this repo is developed on arm64 macOS, and
a suite that silently changes subject with the host is how a target-specific
lowering bug reaches a release.

| # | #7502 row | native assertion |
|---|---|---|
| 1 | a pointer local is a root | root slot + live at the next allocation's statepoint + present in the map |
| 2 | a dead value stops being a root | absent from the live set, against a live control |
| 3 | a numeric local reserves nothing | slot counts `(1, 2)` against a heap twin |
| 4 | slot indices unshifted by a numeric local | subsumed by 3 — native roots have no indices |
| 5 | entry roots begin after the init prelude | first rooted safepoint follows `js_gc_init` **and** `__perry_init_strings_*` |
| 6 | a loop's roots do not cross the back edge | in-loop live set is 1, against a 2-root control |
| 7 | scalar-replaced heap field is a root (#6968) | extra root slot + non-empty map, against the numeric twin |
| 8 | scalar-replaced numeric literal pays nothing (#6997) | empty map, against a one-heap-field twin |
| 9 | every reserved slot reaches the root set (#7184) | two live locals ⇒ two map roots |

## Non-vacuity is the whole point, so it is structural

* **Positive claims assert their subject ran.** `Statepoints::at` panics when the
  callee it is asked about produced no safepoint; `map_records_for` panics when
  the function is missing from the map. "Zero roots" is only ever asserted about
  a record that exists.
* **Every negative claim carries a differential control in the same test.** A
  lowering that roots nothing fails the control half.
* **The harness has its own coverage**, and it earned it twice during
  development. Both bugs would have made every "nothing is live here" assertion
  pass for the wrong reason: the callee parser read
  `@llvm.experimental.gc.statepoint.p0` for every safepoint (the intrinsic's own
  signature contains `) @`), and the live-set parser truncated at the `)` inside
  `ptr addrspace(1)` and reported an **empty live set for every statepoint in
  every program**.

## Sabotage results

Ten sabotages, each confirmed to **compile** (`error[` count 0) and to **reach
the test binary** (`Running unittests` present) before its verdict was believed.

| sabotage | result |
|---|---|
| root allocas emitted as `alloca double` instead of `ptr addrspace(1)` | RED — 10 of 14 tests (all 8 mechanics + both pipeline self-tests) |
| retype *every* scalar alloca, not only bound roots | RED — mechanics 1, 3, 7 |
| conservative CFG-union liveness (reload every root before `ret`) | RED — **mechanic 2 only** |
| root reloaded at each block head and used at its terminator | RED — mechanic 6, in-loop live set 1 → 2 |
| `root_scalar_replaced_slot`'s `root_entry_alloca` removed (#6968 reintroduced) | RED — mechanics 7, 8 (`heap 3 vs numeric 3`) |
| `expr_is_known_non_pointer_shadow_value` early-out removed | RED — numeric literal's map went **0 → 2 roots** |
| `js_gc_init` deleted from `main` | RED — mechanic 5 only |
| `__perry_init_strings_*` moved to a pre-return call | RED — mechanic 5 only (`#32 vs #4`) |
| `roots` vector sized `slot_count - 1` | RED — mechanic 9, map's largest live set 2 → 1 |
| root alloca's address leaked to a call so `mem2reg` cannot promote it | RED — promotion self-test, 2 allocas survived |

One negative result is recorded in the code too, because it says something about
the mechanic: the weaker CFG-union sabotage leaves mechanic 6 **green**, and LLVM
is right about that — the previous iteration's value is genuinely dead at the
next iteration's allocation, since the phi carrying it to the return is redefined
in the body. Only a lowering that keeps a root live across the back edge itself
can break that row.

## Three findings

**#7502's table is wrong about row 9.** It marks #7184's shape "unrepresentable"
because there is no frame bound under native roots. The *frame* bound is gone,
but the defect was never about a frame — it is about an index falling silently
outside the structure that collects roots, and
`lower_precise_roots_to_native_stack` still has one: it collects with
`roots.get_mut(idx)` over a `slot_count`-sized vector, so an out-of-range index
drops the alloca from `root_ptrs` and it is never retyped and never rooted, with
no diagnostic. Sizing that vector one element short removes a root from the
emitted map while the function still compiles, verifies and emits. Now tested.

**`mem2reg` promoting every root alloca is a native-only precondition with no
shadow counterpart.** RS4GC relocates `addrspace(1)` SSA values and does not scan
allocas, so a root slot that escapes promotion is one the collector never
rewrites — the value reads as rooted and is not, which is the #7184/#7192
presentation exactly. Asserted directly in
`no_root_alloca_survives_the_statepoint_rewrite`.

**`docs/src/internals/gc-rooting-invariant.md` was missing its most important
caveat.** `gc_root_dominance_corpus.sh` compiles the corpus under
`PERRY_RS4GC=0` — the **shadow** lowering — and says so inline, for a good
reason (the checker anchors on `@js_shadow_slot_bind`, of which the native
lowering emits zero; under the default the corpus has 1251 statepoints, no binds,
and `--min-binds` fails the job). But the invariant page, which is what a reader
is told to read end to end, listed three blind spots and not that one — so a
green `gc-root-dominance` read as evidence about the shipped lowering. Added as
a fourth, starred, bullet, pointing at these tests as the native-side coverage
until the checker learns to read relocation bundles.

## Production surface

Deliberately tiny, and it cannot change emitted IR:

* `gc_map::decode_stack_map_roots` — `#[cfg(all(test, feature = "llvm-inprocess"))]`.
* `inprocess::statepoint_rewritten_ir` — `#[cfg(test)]` inside a feature-gated module.
* `inprocess::STATEPOINT_REWRITE_PASSES` — a named constant replacing an inline
  string literal of the identical value, so the suite cannot go green against a
  pipeline production no longer uses.

## Verification

* `cargo test -p perry-codegen --lib --no-fail-fast` — 724 passed, 0 failed.
* `cargo test -p perry-runtime --lib --no-fail-fast` — 1915 passed, 0 failed, 4 ignored.
* `cargo test -p perry-codegen --lib --no-default-features --no-run` — builds clean
  (the feature gate exists so it does).
* `cargo check --all-targets --workspace` (cross-host UI crates excluded) — clean;
  re-run under `RUSTFLAGS=-D warnings` and confirmed **none** of the residual
  macOS-host warnings names a file this PR adds or touches.
* `cargo fmt --all -- --check` — clean.
* All **22** lint-job commands extracted from `.github/workflows/test.yml` — 22 run, 0 failed.
* `gc_root_dominance_check.py --moving-only --seeded-violations 40` — see the
  comment below.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1371
+**Current Version:** 0.5.1372
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1371"
+version = "0.5.1372"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "perry-ffi",
  "scraper",
@@ -5813,7 +5813,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-commander"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5821,7 +5821,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cron"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "chrono",
  "cron",
@@ -5831,7 +5831,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dayjs"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "chrono",
  "perry-ffi",
@@ -5839,7 +5839,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-decimal"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "perry-ffi",
  "rust_decimal",
@@ -5847,7 +5847,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dotenv"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "perry-ffi",
  "serde_json",
@@ -5855,7 +5855,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ethers"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "perry-ffi",
  "rand 0.10.1",
@@ -5863,7 +5863,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-events"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5871,14 +5871,14 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-exponential-backoff"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "perry-ffi",
 ]
 
 [[package]]
 name = "perry-ext-fastify"
-version = "0.5.1371"
+version = "0.5.1372"
 dependencies = [
  "bytes",
  "http-body-util",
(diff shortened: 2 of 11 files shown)


proggeramlug:
### `gc-root-dominance` results (codegen was touched, so it was run)

Run against `arm64-apple-macosx`, release build
(`cargo build --release -p perry -p perry-runtime-static -p perry-stdlib-static`,
`PERRY_RUNTIME_DIR` pinned to that release dir so the archives being linked are
the ones just built).

```
python3 scripts/gc_root_dominance_check.py --audit-alloc-re
  === ALLOC_RE: 71 alternatives vs 3784 exported js_* symbols
  === every alternative matches at least one exported symbol            exit 0

python3 scripts/gc_root_dominance_check.py --audit-poll-capable
  === POLL_CAPABLE_RUNTIME: 54 entries vs 3784 exported js_* symbols
  === every entry names an exported runtime symbol                      exit 0

python3 scripts/gc_root_dominance_check.py --self-test                  exit 0

./scripts/gc_root_dominance_corpus.sh ir-corpus
  corpus: 129/129 sources compiled, 0 skipped, 149 .ll files            exit 0

python3 scripts/gc_root_dominance_check.py ir-corpus --moving-only \
  --allowlist scripts/gc_root_dominance_allowlist.json --seeded-violations 40 -v
  === checked 2452 functions / 149 modules (149 .ll files, 9799 root stores)
  === violations: 0   (moving-minor reachable: 0)
  === seeded violations: 40 planted, 40 caught, 0 MISSED                exit 0

python3 scripts/gc_root_dominance_check.py ir-corpus --unrooted-allocas --moving-only \
  --allowlist scripts/gc_root_dominance_allowlist.json -v
  === files: 149   gc-capable allocas: 7862   unrooted-alloca violations: 0
      (moving-minor reachable: 0)                                       exit 0
```

`--moving-only` matters here: without it the corpus is red on `main`, and the
checker's early return means `--seeded-violations` never executes — the seeded
arm would silently not run and the job would still print a verdict.

The expected result for this PR is "unchanged", and that is the point of running
it: the only production-visible edit is a string literal replaced by a named
constant of the identical value, so the emitted IR cannot differ. This confirms
it does not.

proggeramlug:
## Audit — merging as v0.5.1372

### Finding 3 is the most important thing produced today

> `gc_root_dominance_corpus.sh` compiles the corpus under `PERRY_RS4GC=0` — the shadow lowering — so **a green `gc-root-dominance` is evidence about a lowering that does not ship.**

I verified it: the script says so inline, and honestly ("KNOWN GAP, deliberately not papered over"), but `gc-rooting-invariant.md` — the page every brief, including mine, tells people to read end to end — listed three blind spots and not that one. I have cited that gate as authoritative on six PRs today. It was answering a different question than I thought, and the script knew while the doc did not.

Putting it on the page as a **fourth, starred** bullet is the right fix, and naming `native_root_coverage` there as what covers the native side until the checker learns relocation bundles (engine-plan item 8) makes this PR the stand-in for a static check that does not exist yet. That is a bigger deal than the coverage count.

### Sabotage — verified independently, two of them

I re-ran two of your ten against my own tree rather than taking the table:

- **Neutering `root_scalar_replaced_slot`** (reintroducing #6968) → exactly the two scalar-replaced tests red, `error[` **0**, `Running unittests` present. It catches a real historical bug being put back.
- **Sizing the native root vector one short** — your finding that #7502 is *wrong* about #7184 being unrepresentable → **4 tests red**, `error[` 0. The argument holds: `roots.get_mut(idx)` on a short vector returns `None`, the alloca never enters `root_ptrs`, is never retyped, is never a root, and the function still compiles, verifies and emits. Same silence as the runtime bounds check that swallowed #7184, one layer up. Correcting the issue you were sent to implement is the most useful kind of finding.

(My first attempt at the second sabotage didn't apply — my regex missed the actual `vec![None; slot_count as usize]` — and reported 14 green off an unmodified build. Fourth time today I've caught that shape; it is exactly why your `error[`-plus-`Running unittests` protocol earns its keep.)

### Things you were right to push back on

**The "nine mechanics, say which three you left" framing was mine and it was wrong** — #7502 marks seven rows `no` and two `n/a`. You covered all seven plus one `n/a`, and left row 4 with a reason that is better than coverage would have been: native roots have no slot indices, so "indices not shifted by an interleaved numeric local" has no object, and mechanic 3 asserts the substance. Saying that in the docs beats a test that pretends the shape exists.

**Pinning both shipped triples rather than deriving from the host** is the right call for the stated reason — a suite that changes subject with the host is how a target-specific bug ships.

### Verified here

22/22 lint commands, `cargo fmt --check` clean, `perry-codegen --lib` **725 passed**, `perry-runtime --lib` **1916 passed**, all 14 new tests green. Your order-dependence check (10× at default parallelism, 5× at `--test-threads=4`) is the kind of thing that usually gets skipped and then bites six weeks later.

**#7656** now tracks the sibling gap this shares a shape with: `perry-ext-*` link breakage is invisible per-PR for the same structural reason — the gate's scope excludes the thing that actually breaks. Both are "the gate runs but not on the subject".

Issue #7477, opened 2026-08-06
JSON DirectParser parses floats differently from the tape materializer — and node agrees with the tape


proggeramlug:
The DirectParser and the tape materializer produce **different values from the same JSON text**, and node agrees with the tape. On `benchmarks/json_polyglot/bench_field_access.ts` (10k records, values like `i * 3.14159`, checksum = sum of `nested.x` + stringify lengths over 50 iterations):

| parse path | checksum |
|---|---|
| **node** v22.23.1 | **2552985550** |
| tape materializer (`PERRY_JSON_TAPE=1` / auto) | **2552985550** ✅ |
| DirectParser (`PERRY_JSON_TAPE=0`) | **2552986400** ✗ |

Reproduce: `PERRY_JSON_TAPE=0` vs `=1` on that benchmark; the divergence is deterministic and survives `PERRY_GEN_GC=0`, so it is the parser, not GC.

## Why it matters beyond the flag

1. **The DirectParser is the default for real workloads**: every blob under 1 KB and over 16 MB, plus every non-top-level-array shape, parses through it (`try_parse_via_tape`'s auto window). Those all get the divergent values today, silently.
2. **It blocks the most promising `field_access` optimization.** `force_materialize_lazy` could re-parse the retained blob with the DirectParser (~2.3× faster than the per-element tape materializer — see the companion perf issue), and a prototype of exactly that produced the wrong checksum *because* of this divergence. Verified: swapping the materializer for the DirectParser flips the checksum from node-matching to divergent.

## Where to look

Both paths parse JSON numbers from the same bytes, so one of the two float-parsing routines rounds differently (likely a hand-rolled decimal-to-f64 in one path vs a stricter algorithm in the other). Since node (V8's strtod) agrees with the **tape** side, the DirectParser's number parsing is the suspect. `stringify` lengths also feed the checksum, so a float→string difference downstream of parse is the alternative — distinguishing the two is the first step: sum `nested.x` alone (pure parse) vs stringify-length alone.

Issue #7538, opened 2026-08-06
json tape + gen-GC: element-wise materialization loses object key pointers, stringify emits `field0`/`field1` (silent wrong JSON)


proggeramlug:
## Summary

With the lazy JSON tape enabled **and** the generational GC enabled, a full element-wise scan of a parsed top-level array can produce objects whose **key pointers are lost**. `JSON.stringify` then emits the positional fallback names — `{"field0":8184,"field1":16368}` where the source said `{"x":8184,"y":16368}`.

Silent data corruption on the default configuration: no throw, no diagnostic, just wrong JSON out. The values are correct and in the right order; only the key names are gone.

Found while decomposing #7478. Reproduces on `origin/main` @ fc5e56ee0.

## Reproducer

Parse, touch every element, stringify, compare against the blob it came from:

```ts
const items: any[] = [];
for (let i = 0; i < 10000; i++) {
  items.push({
    id: i, name: "item_" + i, value: i * 3.14159,
    tags: ["tag_" + (i % 10), "tag_" + (i % 5)],
    nested: { x: i, y: i * 2 }
  });
}
const blob = JSON.stringify(items);

let bad = 0;
for (let iter = 0; iter < 53; iter++) {
  const parsed = JSON.parse(blob);
  let sum = 0;
  for (let i = 0; i < parsed.length; i++) sum += parsed[i].nested.x;
  const re = JSON.stringify(parsed);
  if (re !== blob) {
    bad++;
    let lo = 0;
    while (lo < blob.length && blob[lo] === re[lo]) lo++;
    console.log("iter:" + iter + " delta:" + (re.length - blob.length) + " at:" + lo);
    console.log("  blob:" + blob.slice(lo - 50, lo + 50));
    console.log("  re  :" + re.slice(lo - 50, lo + 50));
  }
}
console.log("bad-iterations:" + bad + "/53");
```

Output on `main`, default flags:

```
iter:2 delta:10 at:868696
  blob:.772559999998,"tags":["tag_4","tag_4"],"nested":{"x":8184,"y":16368}},{"id":8185,...
  re  :.772559999998,"tags":["tag_4","tag_4"],"nested":{"field0":8184,"field1":16368}},...
bad-iterations:1/53
```

[...]

PR #7537, opened 2026-08-06
perf(json): flip lazy-tape scans to the batch parser early (#7478)

proggeramlug:
Part of #7478.

`lazy_get`'s only adaptive signal is `cumulative_walk_steps`, and it **provably cannot see a scan**: a sequential walk costs exactly one tape step per element, so it accumulates `n` against a threshold of `2n` and never trips. Every element of a scanned array was therefore materialized one at a time, and #7499's batch reparse — gated on the cache still being mostly empty — only ever got asked *after* the scan had already filled the bitmap, where it is a deliberate no-op.

This adds the missing signal: `LazyArrayHeader::sequential_streak`, a run-length of consecutive ascending cold reads, tripped by `scan_flip_threshold` (`n/64`, floored at 64). It fires while the cache is still nearly empty, so `force_materialize_lazy` takes the batch producer for the whole remainder.

## Result

`benchmarks/json_polyglot`, pinned quiet host (M1 mini), 15 runs, `taskpolicy -t 0 -l 0`, arms built from the two runtimes and run back to back:

| workload | baseline | this PR | |
|---|--:|--:|--:|
| **roundtrip** | 201 ms · σ1.3 · 87 MB | **201 ms · σ1.6 · 87 MB** | unchanged |
| **field_access** | 2938 ms · σ135 · 222 MB | **2043 ms · σ146 · 195 MB** | **−30 %** |

Checksums identical to node on both arms. The baseline arm reproduces the reported 2984 ms at 2938, which is what makes the decomposition below trustworthy.

Roundtrip's unmutated-blob memcpy path is untouched by construction: a workload that never indexes never opens a streak.

## Where the 2938 ms actually goes

The public baseline's `idiomatic` row flips **two** switches (`PERRY_JSON_TAPE=0` *and* `PERRY_GEN_GC=0`), so "tape off" had never been measured on its own. Decomposing all four combinations, 11 runs, on `origin/main` @ fc5e56ee0 (`scan` = parse + touch every element, no stringify):

| variant | default (tape+genGC) | tape off (genGC) | tape+markSweep | idiomatic |
|---|--:|--:|--:|--:|
| parse only | 189 · σ1.7 · 69 MB | 1412 · 166 MB | 189 · 67 MB | 1184 · 57 MB |
| parse+scan | 2729 · **σ214.9** · 208 MB | 1468 · 170 MB | 2957 · σ8.8 · **66 MB** | 1226 · 58 MB |
[...]

PR #7654, opened 2026-08-08
docs(engine-plan): close #7478 — the JSON tape's scan penalty, re-measured at v0.5.1370

### LINKED ISSUE #7478: json tape: full-array scans pay 2.3× the direct parser (field_access 2981ms vs bun 223) — decomposition + roadmap
`json_polyglot/field_access` (parse → touch every element → stringify) is the lazy tape's worst case: **2981 ms vs bun's 223**, while the same tape **wins roundtrip outright** (194 vs bun 221, node 384). This issue records tonight's decomposition and the two dead ends, so the next attempt starts from measurements instead of the plausible-but-wrong model I started with.

## Decomposition (50 iterations, quiet host, same binary)

| phase | tape on | tape off |
|---|--:|--:|
| parse only | **210 ms** | 1220 ms |
| parse + full scan | 3030 ms | **1287 ms** |
| parse + stringify | **254 ms** | 1756 ms |

Tape wins parse 6× (deferred materialization) and stringify 7× (unmutated-tape memcpy). **The full scan is the entire loss**: ~56 ms/iter for scan+materialize vs ~24 ms/iter for the direct parser's tree build. Both build the same 10k-record tree — the tape's per-element materialization path is ~2.3× the batch parser.

## Two approaches tried, measured, and rejected

**1. Earlier batch-flip triggers.** Added distinct-element *coverage* (trip at half) and *sequential-streak* (trip at 16 consecutive) triggers alongside the existing walk-steps heuristic — the walk-steps trigger provably cannot fire on sequential access (1 step/element accumulates `n`, threshold is `2n`). Both triggers verified firing (lldb breakpoint: 1 materialize per iteration). **Result: time noise-equal (min 3050 old vs 3322 new), RSS regressed 180 → 213 MB** — the full tree materializes while blob+tape+cache are still live, and total materialization work is unchanged because every element gets materialized either way. Reverted.

**2. Re-parse the retained blob with the DirectParser inside `force_materialize_lazy`** (lazy is top-level-array-only, so the blob is exactly the array's source; patch the few handed-out cached elements back over the fresh slots for identity). This is the structurally right fix — batch parse at 24 ms/iter instead of 56 — but it is **blocked by #7477**: the DirectParser's float parsing diverges from the tape's (and from node), so the prototype produced checksum 2552986400 where node says 2552985550. It also had an intermittent SIGSEGV in my rooting; both are why it was reverted rather than debugged into shape at 4 a.m.

## The path that wins both

1. Fix #7477 (DirectParser float divergence — a correctness bug in its own right).
2. Re-land the reparse-on-materialize with the identity patch loop (the prototype's shape, with the rooting done carefully: root the parsed result before re-reading the header, per-element handle across `store_array_slot`).
3. Keep a trip heuristic only as the *decision* of when to batch — the streak/coverage work is sound for that, it just cannot pay while the batch materializer is the same speed as the element-wise one.

Expected: field_access ≈ 210 + 50×(24 + ε) + memcpy-stringify ≈ **~1500 ms** (from 2981), while roundtrip keeps its 194 ms win — the tape then wins or ties every JSON benchmark against the direct parser, and the remaining gap to bun (~223 ms) becomes a parser-speed workstream (shared with `json_parse_1mb`, currently 6.27× vs bun), not a tape-policy one.

## One more datum

The per-element merge branch in `force_materialize_lazy` (fresh `RuntimeHandleScope` + safepoint per element, to honor a handful of cached slots) looked like the cost and is not: replacing it with root-batch + patch-cached-slots measured neutral. The copied-minor sabotage test (`test_json_tape_force_materialize_sparse_cache_handles_survive_copied_minor_gc`) hooks the `ForceLazyArrayRooted` safepoint — any rework of that function must keep firing it, or the instrument silently loses its subject.

proggeramlug:
Closes #7478 — with a measurement, not a change.

#7478 is `docs/engine-plan.md`'s item 2: *"json tape: full-array scans pay 2.3×
the direct parser (field_access 2981 ms vs bun 223)"*. Its three-step roadmap
has shipped in full, so the first thing owed to it was a re-run of its own
decomposition, not more code.

## Fresh decomposition — v0.5.1370 (`9617779046`)

Pinned quiet mini, `taskpolicy -t 0 -l 0`, **11 interleaved rounds** (every arm
gets sample *r* before any arm gets *r+1*, so drift cannot land on one arm),
176/176 samples asserted, every arm's checksum byte-identical to node 26.5.1.

| phase | tape on: issue → now | tape off: issue → now |
|---|--:|--:|
| parse only | 210 → **160 ms** | 1220 → 1196 ms |
| parse + full scan | 3030 → **1233 ms** | 1287 → 1224 ms |
| parse + stringify | 254 → **171 ms** | 1756 → 1449 ms |
| field_access | 2981 → **1721 ms** | — → 1480 ms |

**The headline claim is gone.** A full scan cost 2.3× the direct parser
(3030 vs 1287); it is now **1.01×** (1233 vs 1224). `roundtrip` did not pay for
it — it improved, 254 → 171 ms.

## Why it closes rather than continues

Measuring all four `PERRY_JSON_TAPE` × `PERRY_GEN_GC` combinations, rather than
the two-switch `idiomatic` arm the issue's floor came from, `field_access` now
decomposes **additively**:

| | tape on | tape off | tape term |
|---|--:|--:|--:|
| gen-GC | 1721 ms | 1480 ms | **+241** |
| mark-sweep | 1133 ms | 938 ms | **+195** |
| gen-GC term | **+588** | **+542** | |

The tape costs ~200 ms whichever collector runs; the collector costs ~560 ms
whether or not the tape exists. The issue had these as an *interaction* (scan
σ 214.9 under gen-GC vs 8.8 under mark-sweep for the identical tape) — that
interaction is what #7539 removed, and it is gone.

The ~200 ms is the tape *build* (parse-only is 160 ms), structural and already
named in #7537: the build is purely additive whenever the whole tree ends up
materialized anyway, and nothing can predict scan-shaped access before the
parse. The remaining ~560 ms is a generational-collector term the tape-off arm
carries identically — GC-campaign work, not tape policy.

## What actually fixed it

| step | landed as |
|---|---|
| 1 — DirectParser float divergence (#7477) | #7483 |
| 2 — reparse-on-materialize | #7499 |
| 3 — early batch-flip trigger | #7537 |
| (not on the roadmap) tape out of the old generation | #7539 |
| (surfaced by the work) wrong JSON on this very benchmark (#7538) | #7546 |

## Method note added to the doc

The "re-measure before scoping" warning goes from three instances to four, and
gains a second failure mode this ticket demonstrates: **a stale floor is as
misleading as a stale headline.** #7478's acceptance bar was "materially under
the 1350 ms `idiomatic` row" — but `idiomatic` is a measurement, not a
constant, and it had moved to 938 ms. Coding to the ticket would have chased a
headline that was already fixed and then declared failure against a bar that no
longer existed.

## Verification

- Both benchmark runs are on a host gated to `rustc == 0` **and** 1-min load
  < 1.5 held across two consecutive checks. The first attempt was discarded:
  another agent's build started mid-run (load 1.86 → 8.57, every σ 50–600).
  Reported numbers are two independent quiet runs that agree.
- The harness **fails** rather than blanks on a cell that produced no sample or
  no checksum, and asserts the expected sample count — a pass cannot be
  confused with "it never ran".
- Subject liveness: the tape is demonstrably engaged (tape-on/tape-off differ
  7.5× on parse-only, 8.5× on roundtrip), and `cargo test -p perry-runtime
  --lib json_tape` is 24/24 including the two cases that assert *which
  producer ran* via `reparse_materializations()`.
- Built `-p perry -p perry-runtime-static -p perry-stdlib-static` with
  `PERRY_RUNTIME_DIR` pinned and the `.a` mtime asserted to have moved;
  `PERRY_NO_AUTO_OPTIMIZE=1` on every compile.

Docs-only: no version bump (maintainer bumps at merge), no `CHANGELOG.md`.

## Summary by CodeRabbit

* **Documentation**
  * Expanded JSON tape guidance with four additional stale-measurement failure modes, including stale acceptance baselines.
  * Updated performance measurements to show that tape-building and garbage-collection costs are additive and independent.
  * Marked the scan-path issue as resolved.
  * Revised the related backlog documentation with updated findings, verification conditions, and measurement methodology.
  * Added a changelog entry documenting the issue closure and test coverage.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1372
+**Current Version:** 0.5.1373
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1372"
+version = "0.5.1373"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1372"
+version = "0.5.1373"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1372"
+version = "0.5.1373"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1372"
+version = "0.5.1373"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1372"
+version = "0.5.1373"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1372"
+version = "0.5.1373"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1372"
+version = "0.5.1373"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1372"
+version = "0.5.1373"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1372"
+version = "0.5.1373"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1372"
+version = "0.5.1373"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1372"
+version = "0.5.1373"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1372"
+version = "0.5.1373"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1372"
+version = "0.5.1373"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1372"
+version = "0.5.1373"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1372"
+version = "0.5.1373"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1372"
+version = "0.5.1373"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1372"
+version = "0.5.1373"
 dependencies = [
  "argon2",
  "perry-ffi",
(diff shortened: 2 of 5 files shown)


proggeramlug:
## Audit — merging as v0.5.1373

**Closing #7478 with a measurement is the right outcome, and the brief asked for exactly this.** A full scan was 2.3× the direct parser and is now **1.01×** (1233 vs 1224), with `roundtrip` — the path that must not pay for it — improving 254 → 171 ms rather than regressing.

The decomposition is what makes it a close rather than a guess: measuring all four `PERRY_JSON_TAPE` × `PERRY_GEN_GC` cells instead of the two-switch `idiomatic` arm shows the cost is now **additive** (tape ~200 ms either collector, collector ~560 ms either way) where the issue had documented an *interaction*. That is a structural answer to "why does this not continue", not a number that happens to be smaller.

**Both corrections to my brief are right and both are mine.** The knob is `PERRY_JSON_TAPE`, not `ln=0/1` — I lifted that from a mangled tool display without checking `parse_api.rs:413`, and passing it on as fact in a brief is worse than not mentioning it. And `docs/engine-plan.md` was internally inconsistent: item 2 claimed 2.2× while the JSON-polyglot section sixty lines above already carried post-#7539 numbers closing it. A plan that disagrees with itself is how the next person gets sent at the wrong thing — reconciling both is the durable half of this PR.

**The method note is the part I most want kept.** `re-measure before scoping` was already there with three instances; *"a stale floor is as misleading as a stale headline"* is a genuinely different failure mode. #7478's bar was "beat the 1350 ms `idiomatic` row" — but `idiomatic` is a measurement, and it had moved to 938 ms. Working the ticket as written would have chased a headline four merges had already fixed **and then declared failure against a bar that no longer existed.** "When acceptance is 'beat arm X', re-measure arm X in the same run" is the rule that falls out of it.

Discarding the first benchmark run when another agent's build landed mid-flight (load 1.86 → 8.57, every σ 50–600) and re-gating on `rustc == 0` **and** load < 1.5 across two consecutive checks is the discipline that makes the rest of the numbers usable. So is a harness that fails rather than blanks on a missing sample, and a gate runner that exited non-zero when `mapfile` was unavailable on macOS bash 3.2 instead of reporting 0 commands as a pass — that last one is the exact vacuity shape that has bitten me four times today.

Gates: 22/22 lint, `perry-runtime --lib` 1915 passed, `cargo check --all-targets` clean, 12/12 json gap tests. The two fastify crashes you found on unmodified `origin/main` are filed separately — they are pre-existing and unrelated to a docs-only diff, but they were not tracked anywhere.

Issue #6904, opened 2026-07-27
Plain number[] element access ~26x slower than Node in hot loops (community histogram benchmark); representation-selection Phase 4 territory


proggeramlug:
### Summary

Community-reported (Discord): an adaptive counting benchmark runs ~25× slower than Node. Reproduced and decomposed. The dominant cost is **plain `number[]` element access** (`counts[v]` read + write in a hot loop); function-call overhead is secondary; arithmetic itself is near-native.

### Reproduction (deterministic variant, 5M iterations, byte-identical output vs Node)

Inner loop: `const v = random(); counts[v] = (counts[v] || 0) + 1;`

Ablation (per-item ns, perry 0.5.1220 vs Node 26.3, same machine/load, ratios stable):

| variant | Node | Perry | ratio |
|---|---|---|---|
| pure inline math (no call, no array) | 1.6 | 4.6 | ~3× |
| + `random()` as function call | 10.5 | 35.4 | 3.4× |
| + `Int32Array` counts | 10.7 | 69.4 | 6.5× |
| + **plain `number[]` counts (packed, pre-filled)** | 9.4 | **244.7** | **26×** |
| original shape (holey `[]`) | 9.2 | 263.1 | 29× |

So of the ~214 ns/item on the original script: ~210 ns is the plain-array element read+write, ~31 ns the call, math is fine.

### Analysis

- The plain-array access lowers to guarded runtime machinery per access (typed-feedback numeric-index guard, write barrier, shadow-slot maintenance) because a `number[]` is a boxed-element heap array — every element read unboxes and every write reboxes + barriers, and V8-style packed-smi element storage does not exist.
- This is squarely the **representation problem**: `docs/representation-selection-rfc.md`. The scalar/call layers are covered by shipped/in-flight phases (#6903 canonical i32 locals; the Phase 2 specialized ABI covers the call overhead; the Int32Array row benefits from #6883/#6898). The **plain-array element representation is Phase 4 (typed heap)** territory: unboxed element storage for arrays proven numeric-only.
- Possible nearer-term slice, before full Phase 4: a packed-numeric-array representation for locally-created, never-escaping `number[]` whose writes are all proven numeric (this benchmark's `counts` qualifies: `new Array(100).fill(0)` or `[]`, indexed only by a 0..99 int, values only `count + 1`). The existing packed-f64 loop-fact machinery covers counter-indexed shapes but not this value-indexed histogram shape.
[...]

Issue #6394, opened 2026-07-14
build: the object cache doesn't key on codegen env vars — PERRY_WRITE_BARRIERS=0 silently does nothing with a warm cache


proggeramlug:
## Summary

Perry's object cache keys on **source + compiler identity**, but **not on codegen-affecting environment variables**. So a build with such a variable set silently reuses the `.o` from a build without it — the switch appears to do nothing, and any A/B built on it is **vacuous**.

## Why this is worse than a stale-cache annoyance

CLAUDE.md documents `PERRY_WRITE_BARRIERS=0` as a **compile-time** bisection tool:

> `PERRY_WRITE_BARRIERS=0`/`off`/`false` disables codegen-emitted write barriers **at compile time** … for benchmark/debug bisection

It is read in codegen (`crates/perry-codegen/src/expr/write_barrier.rs`, `codegen/helpers.rs`), so it changes emitted IR. But with a warm `node_modules/.cache/perry`, setting it changes **nothing** — you get the previously-cached object, with barriers still in. The bisection tool silently lies, and you conclude "barriers aren't the cause" from an experiment that never ran.

This was found the hard way: while validating #6385 (PR #6393), the falsification run — the one meant to prove the new tests *can* fail — **passed**. That would have been a worthless green. It only went red after `rm -rf node_modules/.cache/perry`. A test that passes because the cache served a stale object is indistinguishable from a test that passes because the code is correct.

## Repro

```bash
perry compile x.ts -o x                      # populates the cache
PERRY_WRITE_BARRIERS=0 perry compile x.ts -o x2   # reuses the cached .o; barriers still emitted
```
Compare the emitted IR (`--trace llvm`) for the two — identical.

## Affected switches (any env var read during codegen)

- `PERRY_WRITE_BARRIERS` — documented compile-time bisection switch
- `PERRY_SETJMP_VOLATILE` — added by #6393, same hazard, same reason
- any future codegen-gating env var — this will keep happening

(Runtime-only knobs — `PERRY_GEN_GC`, `PERRY_GC_FORCE_EVACUATE`, `PERRY_GC_DIAG` — are unaffected: they are read by the linked runtime at execution time, not baked into the object.)

## Fix

Fold the values of all codegen-affecting env vars into the object-cache key (`build_cache.rs`). Ideally register them in one place so a new switch cannot be added without also entering the key — a `CODEGEN_ENV_KEYS: &[&str]` array consulted by both the reader and the hasher, so the two cannot drift.

[...]

PR #7660, opened 2026-08-08
fix(repsel): resolve the growth-forwarding chain before the element-shape clone derives its base (#7480)

proggeramlug:
Re-measuring #7480 to decide its remaining step found a **shipped SIGBUS** in
#7612's element-shape versioned loop clone. The perf work #7480 describes is
not what this PR does — it fixes the crash that consumer has on `main`, and
re-measures the issue so its next step is chosen against real numbers. Details
of the measurement are in the issue comment; the fix is below.

## The bug

`js_array_grow` does not grow in place. It allocates the larger array
elsewhere, copies, and leaves a **forwarding stub** at the old address —
and the stub's first payload word (`length`‖`capacity`) is *overwritten with
the new head*, which is how `clean_arr_ptr` follows the chain. Every runtime
entry point resolves that chain, so a binding holding a stale head still
behaves correctly through the runtime. Only bindings the growing code itself
wrote through get re-pointed; a callee that grows a caller's array, or a
function that grows a local and returns it, leaves stale heads behind. That is
repsel 4a.2's (#6904) canonical case, and it already has a fix:
`js_array_refresh_local_head`.

Every other consumer of a raw array head calls it. The element-shape preheader
did not, and derived both the bound check and the elements base straight from
the raw pointer:

```llvm
%r50 = inttoptr i64 %handle to ptr
%r51 = load i32, ptr %r50            ; "length"
%r52 = icmp uge i32 %r51, %bound     ; the prefix-covers-the-range check
%r53 = add i64 %handle, 8            ; elements base
```

On a stub those two facts fail in the worst possible combination. `length`
reads the low 32 bits of a heap pointer — a huge number, so `len_ok` **passes**
— while the base still addresses the pre-growth buffer. Meanwhile the guard
call one block earlier resolved the chain internally and answered truthfully
about the live array, so the class-id test passed too.

The clone then reads correct elements up to the old capacity and runs off the
end of the block after it. `MIN_ARRAY_CAPACITY == 16`, so:

```
sweep bound:  1   16  |  17  100  1000
              ok  ok  |  SIGBUS
```

### Reproducing it — two ingredients, and one of them is not obvious

A first attempt to reproduce this failed, and the reason is worth stating up
front because it defines the blast radius.

* **(A) The clone has to be emitted at all.** `match_element_shape_versioned_loop`
  accepts a loop bound that is `Expr::Integer` or `Expr::LocalGet` and nothing
  else. An inline `arr.length` bound is an `Expr::PropertyGet`, so the matcher
  declines and the binary contains no clone. `for (let j = 0; j < keep.length;
  j++)` therefore **cannot** fault; hoisting the identical bound into a local
  (`const len = keep.length`) is enough to flip it.
* **(B) The binding has to hold a stub** — the array grew past
  `MIN_ARRAY_CAPACITY` (16) inside a callee that returned it. **Minimal N = 17.**

**Not** required, each checked on `main` @ `384eba78b`:

* **the profile** — `--profile perry-dev` and `--release` produce the *same*
  table, same minimal N;
* **`PERRY_NO_AUTO_OPTIMIZE=1`** — it faults with auto-optimize on too;
* **any run-time env var** — none is needed and none suppresses it (the only
  suppressor is `PERRY_GC_MOVING_LOOP_POLLS=1` at *compile* time, which makes
  the clone stand down by construction — that is how I localised it);
* **crossing a function boundary at the read site** — a module-scope loop with
  a hoisted bound faults identically. I originally guessed this was
  load-bearing; the isolation arms below disprove it, and it is corrected here
  and in #7661.

| arm | loop location | loop bound | N=16 | N=17 |
|---|---|---|--:|--:|
| function, array as param | function | local `n` | 0 | **138** |
| module scope | module | `keep.length` | 0 | 0 |
| **module scope** | module | **local `len`** | 0 | **138** |
| **function, array as param** | function | **`keep.length`** | 0 | 0 |
| module scope + `push`/`pop` write-back | module | local `len` | 0 | 0 |

The bound form is load-bearing; the loop's location is not. The last row is the
stub proof: a `keep.push(x); keep.pop();` that changes no contents but routes
through the runtime push helper — which resolves the chain and writes the live
head back — exits 0 with the right answer at exactly the N where the same file
without it faults.

Minimal reproducer, exit 138 (`EXC_BAD_ACCESS` / `SIGBUS`) on `main`:

```ts
class Node { v: number; w: number; constructor(v: number, w: number) { this.v = v; this.w = w; } }
function build(n: number): Node[] {
  const out: Node[] = [];
  for (let i = 0; i < n; i++) out.push(new Node(i, i * 2));
  return out;                       // `out` grew: 16 -> 32 -> ... , stale head returned
}
function sweep(keep: Node[], n: number): number {
  let sum = 0;
  for (let j = 0; j < n; j++) sum += keep[j].v;   // the clone
  return sum;
}
const keep = build(17);                           // 17, not 1000 — 16 is fine
console.log(sweep(keep, keep.length));            // bus error, no output at all
```

Note `sweep`'s bound is the parameter `n` (an `Expr::LocalGet`). Change it to
`j < keep.length` and the same file exits 0 — see ingredient (A).

`PERRY_PTR_SHAPE_LOCALS=0`, `PERRY_GEN_GC=0`, `PERRY_GEN_GC_EVACUATE=0` and
`PERRY_WRITE_BARRIERS=0` all still crash: it is not GC movement and not 3b, it
is growth forwarding.

That `keep` really is a **stub** rather than a live head is established two
ways, not assumed. Directly: adding a module-scope `keep.push(…); keep.pop();`
after the `build()` — which forces a write-back of the *resolved* head into the
global — makes the same pre-fix compiler print the right answer and exit 0.
And by construction: `js_array_refresh_local_head` returns its input untouched
when there is nothing to follow, so if the binding were already live this PR
would be a no-op and the crash would survive it.

Note for a follow-up, out of scope here: something on the **producer** side is
leaving that stub reachable — `expr/array_push.rs` does write the reallocated
head back to the pushing scope's own slot, so the value that reaches `keep` is
losing it somewhere between `build`'s slot and the caller's binding. This PR
fixes the consumer, which is the right layer regardless (a stale head can
arrive from any of several routes, which is why every runtime entry point
resolves the chain), but the producer gap is tracked as #7661.

**Why every existing test missed it.** Both the gap test and the codegen census
build their arrays in the same scope that reads them, so the binding always
held the live head, and the largest was 64 elements pushed at module scope —
where each `push`'s write-back updates the global. The raw-pointer derivation
was never handed a stub.

## The fix

A `element_shape.loop.preheader.repair` block between the brand test and the
guard call: follow the chain once with `js_array_refresh_local_head`, write the
live head back into the binding, re-derive the tag/band predicate from it.

Ordering is the substance, and it is why the repair is a new block rather than
two lines in the existing deref block:

* **It cannot go after the guard call.** The deref block's contract is "no call
  from here to the end of the clone" — that is the revocation argument. The
  refresh can allocate (`clean_arr_ptr` force-materialises a lazy array), so a
  refresh there would reintroduce exactly the "base derived across an
  allocating call" hazard step (4) exists to prevent.
* **The write-back is what makes it work at all.** The query and deref blocks
  both *re-read* the binding — deliberately, because the guard call can move
  the array. Refreshing without storing back would leave both re-reads pulling
  the stub straight out again.

The write-back also lands the durable half of #6904's self-heal: after the
first visit the binding holds the live head, so every later loop entry, and the
slow clone, address the current array directly.

`stmt/element_shape_loop.rs` additionally declines closure-captured arrays,
which have a capture cell a plain slot store would not update.

## Tests

**Gap suite** — `test_gap_repsel_element_shape_loop_clone.ts` gains case 10
with both stale-head shapes above 16 elements: callee-builds-and-returns, and
callee-grows-the-caller's-array. Both bound forms are covered — a parameter
bound (`sumField(returned, returned.length)`, whose loop reads `n`) and a
module-scope loop with a literal bound — so a future narrowing of ingredient
(A) cannot silently drop the coverage. Verified in both directions:

| arm | result |
|---|---|
| `main` (`ec675f2fe`, and `384eba78b`) | **exit 138, SIGBUS** partway through the output |
| this branch | byte-identical to `node --experimental-strip-types` |

Standalone reproducers, both profiles, live at `/tmp/s7480-real/` with a
`run_all.sh` that prints the full N-scan per arm.

**Codegen census** — three tests in `element_shape_loop_tests.rs`. The
load-bearing one asserts the stored-back value *is the refresh's result*, not
merely that a store exists; sabotage-checked by deleting the write-back arm,
which fails that test and only that test. `CLONE_LABELS` gains the repair
block, so a silently-dropped repair fails the census.

## Gates

22/22 lint commands (extracted from `test.yml`), `cargo fmt --all --check`,
`cargo test -p perry-codegen --lib` 728 passed, `cargo test -p perry-runtime
--lib` 1917 passed, `cargo check --all-targets`, `native_root_coverage` 14
passed. Codegen changed, so `gc_root_dominance_check.py --moving-only
--seeded-violations 40` over a freshly built 149-module corpus: **0 violations,
40/40 seeded caught, 9828 root stores** — the checker was live.

The repair introduces a call into the preheader, so the gap test was also run
under seven GC configurations, all byte-identical to node: `PERRY_GC_ZEAL=1`,
`PERRY_GC_FORCE_EVACUATE=1`, `PERRY_GEN_GC=0`, `PERRY_GEN_GC_EVACUATE=0`,
`PERRY_GC_VERIFY_EVACUATION=1`, `PERRY_GC_FROMSPACE_SCAN_ABORT=1`,
`PERRY_WRITE_BARRIERS=0`.

Gap-suite subset as the #6377 check (`repsel` / `array` / `class` / `object` /
`new` / `computed` / `prop` — 76 tests, node 26.5.1): **75 pass, 0 fail**, 1
`node_fail` (`test_gap_prop_plan_cache_invalidation`; node itself exits
non-zero, so not reachable from a compiler change).

## Measured (pinned quiet mini, 7 interleaved rounds, checksums equal everywhere)

200k elements × 50 sweeps, runtime-derived bounds (`process.argv`), release
build, `PERRY_NO_AUTO_OPTIMIZE=1`, `PERRY_RUNTIME_DIR` pinned per arm.

| kernel | perry `main` | perry this PR | node | bun |
|---|--:|--:|--:|--:|
| `keep: Node[]` — the clone's own shape | **SIGBUS** | **13 ms** (12–13) | 57 | 15 |
| `keep: {v,w}[]` — #7480's kernel, out of the clone's reach | 414 ms (414–415) | 414 ms (413–414) | 12 | 12 |

Two controls, because "the guard now declines" would look identical to "fixed"
on the crash test alone:

* **The clone is still entered.** Same binary, same host, clone forced to stand
  down via `PERRY_GC_MOVING_LOOP_POLLS=1`: **49–66 ms** against 12–13 ms with
  it. The fast path is doing the work.
* **The unaffected kernel is untouched, provably.** `k_literal`'s emitted LLVM
  IR is **byte-identical** between the two arms (`cmp`, not a filename compare
  — #6394 keys the object cache on every codegen env var), so 414 vs 414 is
  identity rather than a lucky measurement. `k_class`'s IR differs, as it must.

The `{v,w}[]` row is #7480's remaining work and is untouched here:
`element_class_name` resolves `Array(Named(C))` only, so an object-literal
element type never reaches the clone. The issue's recorded 93 ms / 6.2× is
stale in the optimistic direction; engine-plan item 6 is updated with the
current table, including the reason node's own baseline differs 5× between the
two kernels (`v: number;` survives type-stripping as a class field declaration,
pinning the field to tagged representation).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1374
+**Current Version:** 0.5.1375
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1374"
+version = "0.5.1375"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "perry-ffi",
  "scraper",
@@ -5813,7 +5813,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-commander"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5821,7 +5821,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cron"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "chrono",
  "cron",
@@ -5831,7 +5831,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dayjs"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "chrono",
  "perry-ffi",
@@ -5839,7 +5839,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-decimal"
-version = "0.5.1374"
+version = "0.5.1375"
 dependencies = [
  "perry-ffi",
  "rust_decimal",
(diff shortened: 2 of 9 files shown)


proggeramlug:
## Audit — merging as v0.5.1375. This is a live SIGBUS on `main`.

**Reproduced independently.** Your exact files, my own `main` build (`perry-dev`, `384eba78b`), `PERRY_NO_AUTO_OPTIMIZE=1`:

```
c_minimal_n                     N=16:0  N=17:138  N=1000:138
e_module_scope_local_bound      N=16:0  N=17:138  N=1000:138
d_module_scope_loop             N=16:0  N=17:0    N=1000:0
f_function_inline_length_bound  N=16:0  N=17:0    N=1000:0
```

138 = SIGBUS, at exactly one past `MIN_ARRAY_CAPACITY`. On the fix, all five arms exit 0 **and match node byte-for-byte** at N=16/17/1000.

**And the clone is still entered** — the check that separates a fix from a suppression. Both arms emit 7 versioned-clone blocks; the fix adds `element_shape.loop.preheader.repair` calling `js_array_refresh_local_head` *before* the base is derived, storing the result back through the rooted binding (`store ptr addrspace(1) …`) — correct, since the refresh can allocate and two later blocks re-read it. A fix that worked by making the guard decline would have removed the crash and the optimisation together.

### Why I couldn't reproduce it at first, and why that matters for the writeup

My kernel used an inline `keep.length` bound. `match_element_shape_versioned_loop` accepts only `Expr::Integer` or `Expr::LocalGet`, so a `PropertyGet` bound is rejected and **no clone is emitted at all** — my binary never contained the code that crashes. Your `e` vs `f` isolation is what proves the bound form is load-bearing and the function boundary is not.

**You corrected your own trigger description rather than defending it**, and that is the important part: "any array grown outside the reading scope" was wrong, #7661's title said "hands the *caller* a stale head" and `e` is a module-scope `const` that never crosses a boundary. Retitling the issue and rewriting the fragment on your own finding is worth more than the original report was. `g` — `e` plus a `push`/`pop` no-op that routes through the runtime helper and resolves the chain — is the cleanest possible stub proof.

### The stale-figure count is now five, and this is the first that flattered

93 ms / 6.2× → **414 ms / 34.5×**. Every previous stale headline overstated the problem; this one understated it by 4.5×. That is a distinct hazard from the four already recorded, and worth the plan note.

Two more things the re-measure surfaced that the issue had wrong:

- **The cost model does not describe this kernel.** The issue says "no out-of-line guard calls, the cost is stacked inline diamonds". The object-literal path carries **three calls per iteration** (`js_typed_feedback_observe_property_get`, `js_typed_feedback_record_guard_pass` on the *hit* path, `js_dynamic_string_or_number_add` — the accumulator loses its numeric proof so `+` isn't an `fadd`). Separable lever, correctly not taken here.
- **node is 5× slower on the class arm than the literal arm** (57 vs 12 ms), because `v: number;` cannot be erased — V8 pre-initialises the slot and pins tagged representation. "Beat node" is two different numbers depending on the arm, which is the acceptance-floor hazard in a new form.

### Corrections to my brief, both yours and both right

"Hash both arms' `.a` and assert they differ" is meaningless for a codegen-only change — the runtime source is identical, so the archives *should* match. Emitted-IR `cmp` is the right control, and `k_literal` being byte-identical across arms is what makes 414-vs-414 identity rather than luck. And Route A partly exists already (`collectors/ptr_shape_elements.rs`, #7034 §3) for region-local arrays; the open work is the parameter/global case.

**Your `fmt` catch is the best self-correction of the day**: `rustup run stable cargo fmt --all -- --check` reporting 0 because `$?` was a downstream `tail`'s, with the committed tree actually unformatted and `lint` about to go red. Sixth instance of that shape today, and the only one caught by the person who made it.

**Gates:** 22/22 lint, fmt clean, `perry-codegen --lib` 728, `perry-runtime --lib` 1917, `native_root_coverage` 14/14, and the new gap test is registered in `gc_repsel_corpus.txt` — so it will actually run under zeal rather than being a dark test.

PR #7662, opened 2026-08-08
fix(gc): Layer 1 rooting slice 7 — child_process, Proxy/Reflect, await (#7615)

proggeramlug:
Slice 7 of the Layer 1 emitter migration (#7615). Migrates three `expr/` modules
onto `crate::rooting`, adds them to the `MIGRATED_MODULES` ledger, and fixes what
reading them turned up.

## What is here

| module | before | after |
|---|---|---|
| `expr/child_proc.rs` | 3 arms rooted unconditionally through the raw API, 5 rooted nothing while holding **raw** heap pointers across user lowerings | one argument-list skeleton (`lower_cp_args`), validators below the whole list |
| `expr/proxy_reflect.rs` | 1 arm of 29 made a rooting decision | 28 lowerings migrated; the 8 `reflect-metadata` arms share one body |
| `expr/fs_await.rs` | root correct, **never released** | scope owned by `with_rooted_group`, released in the merge block |

## The node-visible bug

`child_process` validated each argument the instant it was lowered — lower
`command`, throw `ERR_INVALID_ARG_TYPE`, never evaluate `options`. JS evaluates a
call's whole argument list before the callee is entered, so node runs those side
effects and only then throws.

A probe over all seven entry points, A/B'd against node 26.5.1 with a baseline
compiler built from `main` in a separate target directory (identical `-p` set,
identical flags, `PERRY_NO_AUTO_OPTIMIZE=1`, no `PERRY_GC_MOVING_LOOP_POLLS` on
either arm):

```
--- node vs base ---   14 missing lines (every args / options / callback expression)
--- node vs fix  ---   (empty)
```

## The rooting bugs

**Taxonomy (a) — a raw pointer above a collection point**, which `root_reload`
structurally cannot repair:

* `execSync` / `spawnSync` / `spawn` / `execFile` / `execFileSync` stripped
  `command` / `file` to a bare `StringHeader*` and then lowered `args` and
  `options`;
* `fork` is #7453's shape at a second site — `js_jsvalue_to_string_coerce` runs a
  user `toString` and its raw result crossed two more lowerings;
* `spawnBackground` carried `log_file`'s stripped pointer across `env_json`;
* **`process.env[k] = v`** in both branches. The computed one coerced the key with
  `js_to_property_key` — a *fresh* heap string with no other root — stripped it,
  and only then lowered the value. The literal one is #7114 one operand over from
  the `PutValueSet` key #7201 fixed. ES2022 moved `ToPropertyKey` before the RHS,
  so the coercion cannot be sunk; the coerced key is what must survive.

**Taxonomy (c) — operand-to-operand:** 28 `Proxy.*` / `Reflect.*` arms.
`Reflect.has(target, key)` handed the pre-collection `target` register to
`js_reflect_has`; `Reflect.set` does it with four operands.

**#7154 accumulator:** `proxy_build_args_array` threaded the argument array's raw
`*mut ArrayHeader` through its push loop in a bare SSA register, and had no way
to root its *caller's* receiver across the same loop. Deleted; the four call
sites now build the array inside a `RootedGroup` that holds both.

**Never released:** `fs_await.rs` pushed the await-loop root unconditionally and
emitted no truncate on any path. Over-retention in the alloca lowering; in the
FFI fallback a `js_gc_temp_root_push` per *execution* with no truncate — #7462's
shape for an `await` in a loop.

## API

`RootedGroup::adopt_emitted` — the combinator `rooting.rs` deleted unused and
said could return "with its caller and with a written argument for why
`call_rooted` cannot serve". Two callers turned up, same shape: a GC-managed
value produced by an **emitted step** rather than by lowering an `Expr`. The
argument, and what it weakens, are in its doc.

## Verification

* **Gates**: all **22** `lint` commands extracted from `test.yml` — pass.
  `cargo test -p perry-codegen --lib --no-fail-fast` 738 pass;
  `cargo test -p perry-runtime --lib --no-fail-fast` 1917 pass;
  `cargo check --all-targets` clean; the 14 `native_root_coverage` tests pass.
* **Dominance**: `gc_root_dominance_check.py --moving-only --seeded-violations 40`
  over a 149-module corpus — **0 violations, 40 planted / 40 caught / 0 missed**.
  `--unrooted-allocas --moving-only` 0 over 7864 gc-capable allocas.
* **Cost**: corpus root stores **9799 → 9805** (+6). Three zero-cost pins are
  committed so a future "root everything" change goes red.
* **Test sabotage**: the pre-fix source of all five touched files restored from
  `HEAD` — `error[` count 0, `Running unittests` present, **11 of 13 red**. The 2
  that stay green are the deliberate zero-cost pins.
* **Ledger sabotage**: run once per newly listed module — compiling
  `temp_root_push_double` / `temp_root_truncate` pair planted in each, `error[`
  count 0 in all three, ledger red and naming both planted lines.

**No runtime fault is claimed.** Every window is demonstrated in IR; whether a
stale pointer is observably wrong depends on what is recycled into those bytes.

## Caveat on the dominance gate

It compiles its corpus under `PERRY_RS4GC=0` — the **shadow** lowering, which is
not the default since #7370. Its green verdict says nothing about the native
lowering that ships; the `native_root_coverage` suite is what covers that, and it
is run above.

## Left for later slices, with reasons

* `expr/static_field_meta.rs`, `expr/math_simple.rs`, `expr/dyn_extern_i18n.rs` —
  not audited here. A survey flagged candidates in all three (`ClassExprFresh`'s
  `caps_arr` accumulator, `ArrayMap`'s receiver unboxed below its own window, and
  a raw `path_handle` reused across a dynamic-import compare loop that runs module
  `__init` bodies) but I have not verified them myself, so they are leads, not
  findings.
* `lower_call/new.rs` — unblocked by slice 6's `RootedGroup` but at 1988/2000
  lines against the file-size gate; needs its own slice with a split.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability for child-process operations, including spawning, execution, and forking.
  * Improved stability for Proxy, Reflect, environment-variable, and dynamic assignment operations.
  * Fixed edge cases affecting asynchronous waits, promise handling, and sequential awaits.
  * Preserved correct evaluation order and behavior when operations trigger memory cleanup.

* **Tests**
  * Added extensive coverage for these operations, including asynchronous and non-collecting scenarios.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1377
+**Current Version:** 0.5.1378
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1377"
+version = "0.5.1378"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1377"
+version = "0.5.1378"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1377"
+version = "0.5.1378"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1377"
+version = "0.5.1378"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1377"
+version = "0.5.1378"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1377"
+version = "0.5.1378"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1377"
+version = "0.5.1378"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1377"
+version = "0.5.1378"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1377"
+version = "0.5.1378"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1377"
+version = "0.5.1378"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1377"
+version = "0.5.1378"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1377"
+version = "0.5.1378"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1377"
+version = "0.5.1378"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1377"
+version = "0.5.1378"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1377"
+version = "0.5.1378"
 dependencies = [
  "anyhow",
  "clap",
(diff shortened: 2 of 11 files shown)


proggeramlug:
## Audit — merging as v0.5.1378

**The evaluation-order bug is real and I reproduced it.** `execSync(123, sideEffect())` on `main`:

```
main:  threw: TypeError                    <- sideEffect NEVER RAN
fix:   SIDE EFFECT / threw: TypeError      <- byte-identical to node 26.5.1
```

`child_process` validated each argument the instant it was lowered, so a throw on argument 1 skipped argument 2's evaluation entirely. JS evaluates the whole argument list before entering the callee. Seven entry points affected.

**Ledger sabotage verified independently**: a compiling `temp_root_push_i64` planted in all three newly-listed modules → `error[` **0**, `Running unittests` present, ledger red naming the plant.

### Your slice found two flaky tests, and they were not yours

I ran `cargo test -p perry-codegen --lib` and got 740/1 — then 5/5 green, then a failure at run 6. Two distinct pre-existing flakes, both the same bug: **a process-global collector paired with a lock that only serialises the tests which take it.**

- `opt_report` — `FORCED` + the `Mutex<Vec<Entry>>` sink are global; `Session` drains on entry and drop but cannot stop a neighbour emitting mid-test. `rows.len() == 2` failed at **3**.
- `ext_registry` — `USED_PROVIDERS` is a process-wide `Mutex<HashSet>` and `record_ffi_call` fires from any lowering of an ext symbol; the "empty after an unlisted symbol" assertion failed with `ioredis` still present.

0/14 on `main`, 3/38 on this branch. Your slice did not introduce them — it added `child_process` and Proxy/Reflect **lowering** tests, which changed the parallel schedule and made both observable. Fixed in **#7665** (recording narrowed to the guard-holding thread, production untouched), verified 0/25 there and **0/10 on this branch after rebasing**.

Worth naming: I diagnosed both from the *extra row*, not the timing. A 1-in-6 flake invites a retry; `3 instead of 2` says a neighbour's entry is in the snapshot, which names the mechanism immediately.

### Your correction is the eighth of the day and one of the sharpest

> `sequential_awaits_reuse_one_rooted_slot` initially passed in **both** arms because this build uses the statepoint/RS4GC lowering, where `reserve_shadow_slot` returns a stack-map index and no `js_shadow_frame_enter` is emitted at all — my slot-width measure read 0 for every module.

**A rooting measure written against shadow-frame IR silently measures nothing on the default build.** That is exactly today's fourth-blind-spot finding (#7653) reappearing one level down, in a test rather than a gate. Fixing it to cover all three lowerings is right, and the generalisation belongs where you put it.

### Scope

3 of 6, and the three you left are left *honestly* — leads recorded as leads ("I did not verify them myself") rather than promoted to findings. `ClassExprFresh`'s `caps_arr` accumulator, `ArrayMap` unboxing its receiver below its own window, and the `path_handle` reused across a dynamic-import compare loop that runs module `__init` bodies are all worth a future slice, and none is worth claiming unverified.

`RootedGroup::adopt_emitted` filling the gap `rooting.rs` had deleted-as-unused, with two callers of the same shape (a GC value produced by an *emitted step* rather than by lowering an `Expr`) and an argument for why `call_rooted` cannot serve, is the right way to add API.

**Gates:** 22/22 lint, fmt clean, `perry-codegen --lib` 741 passed across 10 consecutive runs, `perry-runtime --lib` 1917, `native_root_coverage` 14/14. Cost: +6 corpus root stores for three zero-cost pins.

PR #7663, opened 2026-08-08
gc: teach the root-dominance checker to read statepoint relocation bundles

proggeramlug:
Closes the gap `docs/src/internals/gc-rooting-invariant.md` records as its first
blind spot and `gc_root_dominance_corpus.sh` apologises for in a paragraph:

> `gc_root_dominance_corpus.sh` compiles the corpus under **`PERRY_RS4GC=0`** —
> the shadow-stack lowering — and since #7370 that is not the default on any
> target whose frames the runtime can walk. **A green `gc-root-dominance` is
> evidence about a lowering that does not ship.**

`scripts/gc_root_dominance_check.py` gains a `--statepoints` mode that reads
`gc.statepoint` relocation bundles, and `gc_root_dominance_corpus.sh` gains
`--lowering native` to emit a corpus it can read. Both shadow modes are
unchanged and still gated.

## One correction to the premise

The brief and the existing in-tree comment both say the corpus "contains 1251
statepoints" under the default lowering. **It contains zero.** `--trace llvm`
dumps what *codegen* emitted, and codegen does not emit statepoints — it emits
`ptr addrspace(1)` root allocas plus a `gc "statepoint-example"` function
attribute, and LLVM's `rewrite-statepoints-for-gc` inserts the safepoints later,
in the linker step (`perry-codegen/src/linker.rs::maybe_rs4gc_preprocess`).
Measured on a default-lowering trace: **0 `gc.statepoint` instructions, 0
`"gc-live"` bundles**, only the `declare` lines. (1251 is roughly the count of
`gc "statepoint-example"` function *attributes*, which is a different thing.)

So the native corpus is the traced IR **plus the production rewrite**. The
corpus script runs `opt` with the pass string it reads out of
`STATEPOINT_REWRITE_PASSES` in `perry-codegen/src/inprocess.rs`, and **refuses
to run if it cannot read that const** — a reproduction of a pipeline that has
silently drifted is a corpus about nothing. That single-sourcing is the price of
not adding a compiler flag; the alternative (a new `--trace` stage that dumps
post-RS4GC IR) was rejected because it only fires on the non-in-process path and
puts corpus generation inside the object cache.

## Design, and what was rejected

**A mode, not a second script.** The statepoint IR is structurally different —
no binds, roots declared per-callsite, explicit relocation — but the CFG
builder, Cooper/Harvey/Kennedy dominance, `between_blocks`, `ALLOC_RE`,
`POLL_CAPABLE_RUNTIME`, `ROOT_READ_CALLS`, `REWRITTEN_LOAD_RE`, the
`IMMOVABLE_SOURCES` exemptions, the allowlist and the liveness floors are all
the same question asked of a different lowering. This file already records what
a second derivation costs: `REWRITTEN_LOAD_RE` was effectively defined twice,
the two modes disagreed about what a collector-rewritten load is, and the
narrower one was wrong — which is #7240. Measured here too: wiring the #7210
exemptions through removed 187 of 326 hits on a probe corpus, every one of them
a population #7210 had already adjudicated.

**The invariant.** Under native roots a value is a root at a safepoint iff it is
in that `gc.statepoint`'s `"gc-live"` bundle, and its identity below is the
`gc.relocate` result. So the rule becomes *no register naming a GC object may be
used below a safepoint unless it is the relocated value*.

**The line that does the work is tracked vs untracked.** LLVM relocates
`ptr addrspace(1)` SSA values and rewrites their dominated uses, so those are
never stale. Everything else is invisible to it — and Perry NaN-boxes, so a
JSValue spends most of its life as a `double`. Getting that line wrong in the
loose direction is not a false positive but a wall of them: a cast-closure that
walked *through* `inttoptr … to ptr addrspace(1)` reached the relocation `phi`
at the top of every loop and came back out below it, reporting the **relocated**
value as stale — 21 of the first 29 hits, all correct code.

**Two verdict classes, because they have two different fixes:**

| class | means | fix |
|---|---|---|
| `unrooted` | nothing in the register's cast chain is in the safepoint's live bundle — the object is unmarked and unrewritten | root it |
| `stale` | the object IS in the bundle and is relocated, but a raw copy of its pre-move address is used below | re-derive from the relocated value |

**Two things this mode gets that the shadow modes cannot.** `NONCOLLECTING` is
not consulted at all — LLVM already decided which calls are safepoints and put
the answer in the IR, so a wrong entry in that hand-kept list cannot hide a
hazard here (a statepoint over `js_write_barrier_root_nanbox`, which genuinely
cannot collect and *is* in `NONCOLLECTING`, still counts). And every safepoint
names its wrapped callee, so `--moving-only` classifies against the real symbol.

## A pre-existing parser bug, found by a seeded violation that went unreported

`DEFINE_RE` was `^define\s+.*?@([\w.$]+)\(`. LLVM **quotes** an identifier it
cannot print bare, and `$` forces it — so every representation-selection
specialisation is printed `@"name$typed_f64"`. The regex could not match across
the quotes, `cur` stayed `None`, and **the entire function body was skipped
without a word**, in a parser that raises `MalformedIR` elsewhere precisely to
prevent silent skips.

**175 of 2452 defines (7.1%) in the native corpus**, and not a random 7%: repsel
specialisations are where this page says a representation change moves the
rooting obligation. Zero in the shadow corpus — perry's own writer prints those
names bare — so the existing shadow arms were never affected. Fixed, and a
`define` the parser cannot name now raises instead of being skipped.

Found only because `--seeded-violations` reported `40 planted, 39 caught, 1
MISSED` and the missed one was in a function nothing was parsing.

## Proof the gate can fail

* **`--seeded-violations 40` on the real corpus.** Splices a safepoint between a
  real `ptrtoint ptr addrspace(1)` and its real use, in perry-plus-RS4GC output,
  and requires every one to be reported. **40 planted, 40 caught, 0 missed.**
  The spliced callee is `js_gc_loop_safepoint`, so the planted hazard survives
  `--moving-only` — the configuration the gate ships with, not a laxer one.
  It runs **unconditionally**, not after an early return on real violations:
  the run whose "can it still fail" arm you most want is the one already telling
  you something. (The bind-anchored path's seeder still sits below its return;
  `--seeded-violations` is now a usage error on the two modes that have no
  seeder, which it was silently ignored by before.)
* **A 10-case sabotage matrix**, each turning `--self-test` red and naming its
  own arm: the callee reader, the live-set paren nesting, invoke-continuation
  folding, quoted define names, the tracked/untracked line, comment stripping,
  the #7210 exemption carry-over, the rooted/unrooted split, `AS1_DEF_RE`'s phi
  alternative, and the `--min-statepoints` floor. **Two arms were vacuous when
  first written and are now not** — the tracked/untracked line had no fixture at
  all (added: a relocation-phi round trip), and the `--min-statepoints` floor was
  passing via `--min-relocates`, so it is now asserted with the other two floors
  relaxed to 0.
* **A structural non-vacuity assertion.** Every `"gc-live"` operand is
  `ptr addrspace(1)` by construction, so one whose definition `AS1_DEF_RE` cannot
  classify is a parser gap and is an error. It fired immediately on first run:
  262 unrecognised operands, all `%.N = phi ptr addrspace(1) […]`, because
  `addrspace\(1\)\b` has no word boundary between `)` and the following space.
  Exactly the dead-alternative failure `--audit-alloc-re` exists for, in a
  different regex.

## Liveness floors, and why each number

Corpus as of this commit: 149 modules, 2452 functions, **30033 safepoints,
17478 with a non-empty live bundle, 40759 relocations**, 34511 (safepoint, root)
pairs. Floors sit well below that and far above "something compiled":

| floor | value | fails when |
|---|---|---|
| `--min-statepoints` | 15000 | this is the shadow corpus, or codegen stopped marking functions `gc "statepoint-example"` |
| `--min-relocates` | 20000 | the corpus was copied through un-rewritten (`opt` exits 0 on a module with nothing to do) |
| `--min-live-bundles` | 8000 | safepoints record **no** roots — what an unrooted build looks like |
| `--min-files` / `--min-funcs` | 90 / 1200 | breadth, same as the shadow arms |

`--min-binds` cannot do this job: the native corpus has zero binds *by
construction*, so a bind floor rejects every valid corpus for this mode and
accepts none. **Each mode refuses the other's corpus**, asserted both ways in
`--self-test` and rehearsed against both real corpora:

```
statepoints over the SHADOW corpus  -> exit 2
bind-anchored over the NATIVE corpus -> exit 2
```

The corpus script asserts its own subject too: zero statepoints or zero live
bundles is an error at generation time, not a clean verdict downstream.

## What it found: 21 real hazards, and they are known shapes

`--statepoints --moving-only` over the curated corpus: **21 hazards, all
`unrooted`, 0 `stale`.** Unfiltered (the RECLAIM half of the invariant as well
as the MOVE half): 1444 — 1123 `unrooted`, 321 `stale`.

The 21 fall into four shapes, every one of which this repo has a name for:

* **9 × `strhandle`** — a string-literal handle global loaded into a register and
  held across `js_closure_call1` or the generic property-GET dispatch. This is
  **#7240's exact shape**; the fix landed for the shadow lowering
  (`OperandProtection::Reload`) and these say it does not cover the native one.
* **7 × `unmasked`** — a receiver unmasked out of `ptr addrspace(1)` and
  dereferenced below a property GET. This is **#7280 / zod's `clone`**, the shape
  that faults deterministically under
  `PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800`.
* **2 × `global`** — a `@perry_global_*` module variable read into a register and
  held across `js_closure_call1`.
* **1 × `rootread`** — `js_new_target_get()` → `js_new_target_set()` across a
  constructor call. Structurally identical to **#7226**'s `js_implicit_this_set`
  `prev_this`, for `new.target` instead of `this`.
* **1 × `capture`** — a closure capture cell read held across `js_number_coerce`.

Full `-v` output is in the job log and is uploaded as an artifact on failure.

**These are not suppressed into an allowlist.** They are a population under
triage, so the gate carries `--max-unrooted 21` — a ratchet that can only be
lowered, the same call `--stale-registers` makes and for the same stated reason
(21 tombstones with no issue numbers would be worse documentation than one
number that can only go down). `--max-stale 0` holds the other class at zero.
**#7664 enumerates all 21 by shape and is the budget's referent** — a number
with nothing behind it is exactly what CLAUDE.md says a threshold decays into.
Each shape wants its own fix and its own decrement before promotion.

## Not promoted to required

`gc-root-dominance-statepoints` is a **separate job** — separate corpus,
separate floors, and, the reason that matters, a separate branch-protection
context, so it can be promoted without dragging the shadow arms along. It is
**deliberately not in branch protection**: a gate that has never been green
blocks every open PR the day it becomes required, which is the corollary in
CLAUDE.md this repo has already paid for once. Let it run on `main` first.
Promotion is an admin action and a separate decision; the doc says so.

The one `|| true` in the file is on the unfiltered census step, which is a
diagnostic and says so where it runs — it keeps the filtered number honest, so
that "the gated arm reads zero" can be told apart from "the filter ate
everything".

## Rehearsal of the exact CI commands

Both jobs' command lines, run locally against freshly built corpora, exit status
taken from the checker itself and never through a pipe:

```
gc-root-dominance-statepoints
  self-test / audit-alloc-re / audit-poll-capable / audit-immovable-sources  exit=0
  gated arm                                                                  exit=0
      checked 2452 functions / 149 modules (149 .ll files)
      safepoints: 30033   with a live bundle: 17478   relocates: 40759
      statepoint hazards: 21  (unrooted: 21, stale: 0)
      within budget: unrooted 21 <= 21
      within budget: stale 0 <= 0
      seeded statepoint violations: 40 planted, 40 caught, 0 MISSED

gc-root-dominance  (regression: unchanged shadow arms)
  bind-anchored --moving-only     exit=0   0 violations, seeded 40/40
  unrooted-allocas --moving-only  exit=0   0 over 7862 gc-capable allocas
  stale-register budget           exit=0   23, budget 39
```

The new job is registered in `scripts/gc_gate_wiring_check.py`'s `GATES` in this
same change — an unregistered sibling is precisely how the job above spent
months being the checked one while the lowering it reads stopped shipping. The
census step carries an explicit `continue-on-error: true` so that gate reads its
`|| true` as a declared opt-out rather than a swallowed failure.

## Gates

22 lint commands extracted from `test.yml`, plus `cargo fmt --all -- --check`,
the checker's four static arms, the workflow YAML parse and the corpus script's
syntax check: **28 commands, 0 failing**. `cargo check --all-targets` rc=0,
`cargo test -p perry-codegen --lib --no-fail-fast` 725 passed / 0 failed
(including all 14 `native_root_coverage` tests), `cargo test -p perry-runtime
--lib --no-fail-fast` 1917 passed / 0 failed. `error[` count 0 in every log and
`Running unittests` present in both test logs.

**No regression in the shadow arms**, rehearsed against a freshly built shadow
corpus at this commit: bind-anchored `--moving-only` **0** violations over 2452
functions / 149 modules / 9799 root stores; `--unrooted-allocas --moving-only`
**0** over 7862 gc-capable allocas; `--stale-registers --moving-only` **23**,
inside the existing budget of 39.

## Summary by CodeRabbit

* **New Features**
  * Added native statepoint validation for unrooted and stale garbage-collection values.
  * Added native and shadow-stack corpus generation with live-bundle and relocation checks.
  * Added configurable filters, allowlists, budgets, thresholds, and seeded-violation checks.
  * Added independent CI validation for native garbage-collection root lowering.

* **Bug Fixes**
  * Improved handling of LLVM IR syntax, control-flow edges, comments, statepoints, and tracked references.
  * Added clearer failures for unsupported or mismatched corpus configurations.

* **Documentation**
  * Expanded guidance for running and interpreting both garbage-collection rooting checks.


proggeramlug:
## Audit — merging as v0.5.1376

This closes plan item 8, and it found a hole in the **existing** gate on the way.

### The pre-existing parser bug is the headline

`DEFINE_RE` could not match a **quoted** function name, and LLVM's printer quotes any name containing `$` — so **175 of 2452 defines (7.1%) in the native corpus were skipped in silence**, every one a repsel specialisation. Zero in the shadow corpus, because perry prints those names bare and only LLVM's printer quotes them. That asymmetry is exactly why the existing arms never noticed.

I sabotage-verified it rather than taking the count: reverting `DEFINE_RE` to the pre-fix pattern turns `--self-test` red with

> *"`define` whose function name this parser cannot read … Skipping it reports every function in it clean, which is how 175 repsel specialisations left the native corpus without a word."*

**An error message that states the consequence rather than the symptom is worth as much as the fix.** A parser that silently skips 7% of a corpus and reports the rest clean is the purest form of the failure this repo keeps paying for — and it was found only because a seeded violation went unreported in a function nothing was parsing. That is the seeded-violations arm doing precisely the job it exists for.

The other parser repairs are the same class: `invoke` destinations on continuation lines (349 — without which every block below an `invoke` was unreachable and dominance answered False for all of it), `landingpad` clauses, hyphenated `split-lp` labels, and LLVM's `; (%orig, %base)` annotations being read as uses.

### My brief was wrong on its central premise

I wrote "under the default it has statepoints and no binds". **The `--trace llvm` corpus contains zero `gc.statepoint` instructions** — codegen emits `ptr addrspace(1)` allocas plus a `gc "statepoint-example"` attribute, and LLVM inserts the safepoints later in the linker. The in-tree "1251 statepoints" counts function *attributes*. So this could never have been a flag flip, and running the traced IR through the production rewrite (with the pass string read from `STATEPOINT_REWRITE_PASSES`, refusing to run if it cannot read that const) is the only honest construction. Good catch, and it changes the shape of the whole change.

### The tracked/untracked line

That LLVM relocates `ptr addrspace(1)` and rewrites its dominated uses — so those are never stale, while Perry's NaN-boxed values live on the other side — is the distinction the whole mode rests on, and getting it loose produced 21 false positives out of the first 29, all the relocation-phi round trip. Two verdict classes (`unrooted` = root it, `stale` = re-derive) is the right decomposition.

**Two of your own sabotages were vacuous when written and you fixed them** — the tracked/untracked line had no fixture at all, and `--min-statepoints` was passing via `--min-relocates`. That is the eighth instance of that shape today and the second caught by its own author.

### The 21 violations

Not allowlisted, ratcheted at `--max-unrooted 21` with #7664 as the referent — correct. That nine of them are **#7240's shape** is the finding that justifies the whole item: #7240 was fixed for the shadow lowering, and the native one has been carrying it since #7370 with nothing able to see it.

### Verified here

`--self-test`, `--audit-alloc-re`, `--audit-poll-capable` all exit 0; `gc_gate_wiring_check.py` reports **6 gates main-line-reachable and able to fail**; 22/22 lint; fmt clean; `perry-codegen --lib` 728; `perry-runtime --lib` 1917. **Not in `required_status_checks` — confirmed by querying branch protection, not by reading the workflow.** Promotion stays an admin action and a separate decision, per the corollary that has already blocked every open PR here once.

Issue #7481, opened 2026-08-06
instance_private_key_hidden faults under PERRY_GC_SCAVENGE_NURSERY_MB=64 (residual from #7472)


proggeramlug:
Residual from #7472 (now closed — its headline `stringify_object_inner` crash was fixed by #7473). At a **non-default** nursery cap, a *different* site still faults:

```
cd <repo>
cargo build --release -p perry -p perry-runtime-static -p perry-stdlib-static
PERRY_RUNTIME_DIR=$PWD/target/release PERRY_NO_AUTO_OPTIMIZE=1 \
  ./target/release/perry benchmarks/json_polyglot/bench_field_access.ts --debug-symbols -o /tmp/fa
PERRY_GC_SCAVENGE_NURSERY_MB=64 /tmp/fa
```

```
EXC_BAD_ACCESS (code=1, address=0x33038393663)
frame #0: perry_runtime::object::field_get_set::enumeration::instance_private_key_hidden + 60
```

Observed on `main` @ 351742d30 (2026-08-06), deterministic at cap=64, absent at the default 16 MB and at 1/4/32 MB post-#7473. The fault address is a garbage pointer (not a NaN-box immediate this time), in the enumeration path's private-key check — likely the same family as #7472's mechanism (a value surviving a large, infrequent copying minor without its slot being rewritten), but at a different consumer.

Not release-blocking (non-default knob), but it is a live copying-minor correctness signal at exactly the cadence #7472 showed the ratchet probes never exercise — a large-Eden probe arm would catch both. Recording it so it does not vanish inside the closed ticket.

proggeramlug:
## The cadence this issue named is now a pinned, gated ratchet probe (#7666) — and the fault itself no longer reproduces

Two separate findings; the second does not close the first.

### 1. The fault does not reproduce at `7bde3de24` (v0.5.1376)

The reproducer exactly as written here — `benchmarks/json_polyglot/bench_field_access.ts`, release build, `PERRY_RUNTIME_DIR` pinned, `PERRY_NO_AUTO_OPTIMIZE=1`, `PERRY_GC_SCAVENGE_NURSERY_MB=64` — exits **0 on 5/5 runs with an identical checksum** (`2552985550`) on the pinned quiet host. No `EXC_BAD_ACCESS`, no `instance_private_key_hidden` fault.

[...]

Issue #7056, opened 2026-07-30
gc(measurement): poll-time moving buys ~nothing the 16 MB nursery cap does not already buy — delete the loop back-edge poll, do not build pin-based poll moving


proggeramlug:
Measurement run to settle one decision: under the plan to replace shadow-stack precise roots with a conservative stack scan plus per-object pinning, the shadow stack's only remaining unique job is enabling the moving minor at loop back-edge polls. If poll-time moving buys ~nothing, that arm gets deleted outright; if it buys something real, pin-based poll moving has to be built first.

## Bottom line

**Poll-time moving buys ~nothing. Delete the arm; do not build pin-based poll moving.**

Every gram of the footprint benefit attributed to the loop back-edge poll is actually the **16 MB nursery cap** — a pacing number that merely shares an enable gate with the poll. Two independent isolations show it:

- Keep the poll, raise the cap to 128 MB: peak RSS becomes **identical to the kill switch** on 7 of 7 workloads — while the poll is still firing and still relocating a million objects.
- Drop the poll, keep the 16 MB cap: peak RSS is **better than shipped** (231.9 MB vs 272.2 MB) at the same wall time, copying *more* objects.

Meanwhile the poll's codegen half costs **up to 3.7×** on tight loops, 19 of 24 suite benchmarks never collect at all, and on the server-shaped workload the shadow stack contributed **3 rewritten slots across 63 moving minors that relocated 10.6 million objects**.

Recommendation: delete the loop back-edge poll, decouple the 16 MB cap from `gc_moving_loop_polls_enabled()` and keep it. See §9 for what this does *not* settle — in particular it is not a verdict on conservative scanning, which §6–§7 measure separately and which #6981 has already shown is load-bearing for correctness.

**Separately and more urgently:** this run turned up a deterministic wrong answer produced by the *shipped default configuration* on an ordinary server-shaped program. It is not a poll bug — the discriminator is whether a copying minor runs at all. Filed separately as #7055; see §10.

---

## Provenance

| | |
|---|---|
| Host | `perry-macos` — Mac mini, 8-core M1, 8 GB, macOS 26.5.1, idle |
| Load average | 1.5–2.8 throughout, recorded per run inside every result JSON (`loadavg_start`/`loadavg_end`) |
| Commit | `e279b2d54`, detached worktree, isolated `CARGO_TARGET_DIR` |
| Build | `cargo build --release -p perry -p perry-runtime-static -p perry-stdlib-static` |
[...]

PR #7666, opened 2026-08-08
test(gc-ratchet): pin the large-Eden copying-minor cadence (#7481) and re-derive the #7056 RSS numbers under statepoints

### LINKED ISSUE #7652: gc-ratchet baseline has mixed provenance after #7650's surgical re-pin, and the schema cannot express it

proggeramlug:
Closes plan items 9 and 10 of `docs/engine-plan.md`. They share the pinned quiet host and the same harness, so they are one PR.

Provenance for **every** number below: `perry-macos` (Mac mini, 8-core M1, 8 GB, macOS 26.5.1). Binary built at `7bde3de24` with `cargo build --release -p perry -p perry-runtime-static -p perry-stdlib-static` into an isolated `CARGO_TARGET_DIR`; the `.a` mtimes were confirmed to move after that build, and this PR changes **zero** `.rs` files (`git diff --name-only 7bde3de24 <head> | grep -c '\.rs$'` = 0), so it is the right binary for this tree. `PERRY_RUNTIME_DIR` pinned, `PERRY_NO_AUTO_OPTIMIZE=1`, node oracle `v26.5.1`. Host quiet was re-checked before every arm and every phase: **94.6–95.8% idle, zero `rustc`, zero `cargo`** throughout. No other agent was building on the box.

---

## Item 10 — the large-Eden probe arm

#7481 named the gap itself: "a live copying-minor correctness signal at exactly the cadence the ratchet probes never exercise". All twelve probes ran the shipped 16 MB nursery cap, so every copying minor this matrix had ever exercised was small (~16 MB of Eden) and frequent.

### A probe may now declare the collector it is a probe *of*

```ts
// gc-ratchet-env: PERRY_GC_SCAVENGE_NURSERY_MB=64
```

It lives in the probe source because it is not possible to read the workload without reading the arm. It reaches every run of that probe — warmup, the timed repeats, both traced runs, and both of `classify`'s scan modes — and deliberately **not** `compile_probe` (these are `OnceLock` runtime knobs, and Perry's object cache keys on every codegen env var, #6394, so a compile-time pass would move the cache key without moving a byte of code). Only `PERRY_*`; never `PERRY_CONSERVATIVE_STACK_SCAN` or `PERRY_GC_DIAG`, which the harness owns; never a repeated key.

The arm is pinned in the artifact and **compared like a metric**, before any band arithmetic. That is the check that matters: delete the directive and every band is still satisfied, because the numbers would have been re-pinned by the same act that lost the arm.

### The finding that shaped the probe

**A large Eden on a *small* retained set runs zero copying minors.** `gc/policy.rs`'s `arena_growth_full_escalation_due` escalates a minor to a full mark-sweep once arena in-use clears 32 MB *and* exceeds twice the post-full baseline; a 64 MB Eden over a ~1 MB live set satisfies both, every time. The first draft of this probe did exactly that — 0 copying minors, 4 full sweeps — and would have been pinned on a collector it never reached, the #7024 shape inside the arm added to close it.

So `PERRY_GC_SCAVENGE_NURSERY_MB` is not on its own a "larger Eden" knob: above ~32 MB it is a "no copying minor" knob unless the workload also holds a live set. Sizing the retained set is the fix:

| `KEEP` | copying minors at 64 MB | at the shipped default |
|---|---:|---:|
| 8,192 | **0** | 15 |
| 131,072 | 1 | 14 |
| 262,144 (shipped) | **4** | 12 |

At the shipped size the four minors free 37, 36, 68 and 68 MB — **49.7 MB per minor** — and the first copies **532,482 objects (32 MB) in one cycle**. The rest of the suite runs **14.6–16.6 MB per minor on eleven of twelve**, and 21.8 MB on `12_large_live_set`, whose tenured-proportional cap term (`max(influx x scale, tenured/2)`) is the shipped path to a larger Eden and tops out around 22 MB even on the biggest workload the suite has. Traced counters are byte-identical across traced runs, and probe stdout is byte-identical to Node 26.5.1.

### Shown able to fail

Three perturbations, each measured with the real harness and scored by `check` against the freshly pinned artifact:

| perturbation | `check` verdict | what it reported |
|---|---|---|
| control — nothing changed | `gc-ratchet: OK` | the pin reproduces on an independent 7-repeat session; probe 13's wall median 3,137 vs 3,121 ms, retention and every counter bit-identical |
| `gc-ratchet-env` directive deleted | **`FAILED`** | the arm-mismatch line, plus `minor_cycles` 4 -> 12, `step_cycles` 5 -> 13, `copied_objects` 532,482 -> 286,246, `copied_bytes` -45.56% |
| `PERRY_GC_SCAVENGE=0` | **`FAILED`** | the liveness rule fires on probe 13 — `minor_cycles` 4 -> 0 and `copied + promoted` 1,074,380 -> 0 — plus 11 REGRESSION rows including retention **+3135%** |
| `PERRY_GEN_GC_EVACUATE=0` | `OK` | **the perturbation was inert, not the gate.** See below. |

The directive-deleted row is the one that justifies gating the arm at all: with the directive gone, retention **-80.27%**, peak RSS **-37.33%** and wall **-27.07%** all read as *improvements*, because those bands are one-sided. Only the arm check and the two-sided evacuation counters say what actually happened.

The `PERRY_GEN_GC_EVACUATE=0` row is kept because it is the lesson rather than a footnote. That knob does **not** gate the copying minor — 4 copying minors with it set, 4 without; it gates the C4b old-gen policy evacuation and `gc_force_evacuate_enabled`. A green gate under a knob whose name promises otherwise reads as "this gate cannot fail". The correct reading was "nothing was perturbed", and the discriminator was counting copying minors before trusting the verdict, not re-reading the harness.

Plus unit-level, in `tests/test_gc_ratchet.py`: the arm reaching the child is asserted by a stub probe that *reports* whether the variable arrived (a `run_env` recorded but never exported would be an arm that is documented, gated and inert); dropping, changing or adding an arm each fails `check`; an unreadable `run_env` is a fatal artifact defect; the shipped probe is asserted to still carry its directive, so deleting it turns a unit test red rather than silently retargeting the probe.

### `wt-scavtenure`

**Subsumed. Measured, not assumed.** #7432 is merged, its worktree exists on neither host, and the baseline has been re-pinned five times since — most recently in full by #7657 at `59d522052` on this same host, which also closed #7652's mixed provenance. The quiet-host driver's `--check` at this PR's first commit, against that artifact: **every one of its 144 pinned cells `ok`**, failing on exactly one line — `probes present now but absent from the baseline: 13_large_eden_survivors`, which is the friction adding a probe is supposed to cost. There is no pending re-pin.

★ **One trap, found the hard way, and it is why this section says 144/144 and not something else.** An earlier ad-hoc `measure --probes-dir <copy>` reported `09_try_catch_roots.heap_used_bytes` at **−17.98%**, which reads exactly like drift since `59d522052`. It is not drift. **A probe compiled with a `package.json` in scope retains one more 1 MiB arena block at `gc()` than the same source compiled without one**, and 09 sits on that boundary. Reproduced three ways, each stable 3/3: repo directory 5,825,256; three separate `/tmp` directories 4,777,624; `/tmp` **plus a copied `package.json`** 5,825,256. Ruled out as build non-determinism in the same session — two builds of one probe from one directory differ byte-for-byte and report the identical number. The driver and CI always compile from `benchmarks/gc_ratchet/probes`, so the gate is self-consistent, and *ratios between arms measured in the same directory* are unaffected, which is what leaves item 9's 2x2 standing. `README.md` now carries the warning.

---

## Item 9 — the #7056 RSS numbers, re-derived under the statepoint default

#7056's numbers were taken under the shadow-stack root lowering. Statepoints became the default in #7370, so the question is whether that invalidates them. 2x2 over root lowering x nursery cap, 12 probes, 7 repeats, **3 interleaved rotations** (rotation *i* runs every arm before *i+1*, so drift perturbs all four equally), every probe byte-identical to the pinned Node oracle in all 72 probe-runs.

### The root lowering is not an RSS lever

| | statepoint (default) | shadow stack (`PERRY_RS4GC=0`) | ratio |
|---|--:|--:|--:|
| peak RSS, 12 probes | 544,210,944 B | 545,144,832 B | **1.002** |
| retained heap after `gc()` | 114,594,904 B | 114,596,520 B | **1.000** |
| wall | 4,808 ms | 4,894 ms | 1.018 |

**104 of 108 deterministic cells are bit-identical.** All four that move are on `10_store_receiver_across_alloc`, all ≤0.31%: the shadow stack keeps 7 more objects live at that collection point than the statepoint map does (`copied_objects` 8,056 → 8,063).

Between-rotation spread: retention **0.000%**, peak RSS max **0.621%** (median 0.000%), wall max 2.779% (median 0.449%). Every claim above clears its spread by an order of magnitude or is explicitly a null result.

**The arms were shown to differ before they were compared — per probe, not per suite.** `--trace llvm` census over all 12: `statepoint-example` + `addrspace(1)` root allocas present on 12/12 under the default (168–514 `addrspace(1)` sites each) and **0 of both** under `PERRY_RS4GC=0`, where `js_shadow_frame_enter` rises to match. This is #7056's own poll-site-count discipline: differing bytes are not a differing configuration.

### What actually invalidated parts of #7056, and it is not #7370

- **§9's recommendation shipped** (#7377). The 16 MB cap no longer hangs off `gc_moving_loop_polls_enabled()`, and the poll has been default-off since #7161. The five arms #7056 tabulates (`on`/`off`/`off_rt`/`on_cap128`/`off_scav`) no longer name anything that ships.
- **The cap is still the whole footprint lever**, now on 12 probes rather than 8. At `PERRY_GC_SCAVENGE_NURSERY_MB=128`: peak RSS **1.911x**, retained heap **2.475x**, wall **0.947x** — up to **5.005x** peak RSS on `04_dead_after_deep_stack` and **6.350x** retention on `07_array_grow_evacuate`. Identical to within 0.6% under both root lowerings, so it is a pacing result and not a rooting one.
- **§7's false-retention table is gone, not shrunk.** It recorded a 4.6x–54.8x conservative-vs-precise band on probes 01–08. `classify` at `7bde3de24` reports **excess 0.00% and spread 0 on all twelve**, with no `manual_collect` scan site anywhere — #7657 removed the forced scan at `gc()` rather than the reading.
- **§6's real finding survives intact.** A conservative scan does not degrade the copying minor, it **disables** it. Forcing `PERRY_CONSERVATIVE_STACK_SCAN=full` under statepoints takes `minor_cycles` to **0 on all twelve probes** (`copied + promoted` = 0 on all twelve), for 1.98x retained heap (0.70x–6.35x) and 1.46x peak RSS (0.93x–3.47x). Smaller than #7056's +364%..+5371%, same sign, same mechanism.
- **#7481 no longer reproduces.** Its own reproducer (`benchmarks/json_polyglot/bench_field_access.ts` under `PERRY_GC_SCAVENGE_NURSERY_MB=64`) exits 0 with an identical checksum on 5/5 runs at `7bde3de24`, and does run 6 copying minors there, so the arm is not vacuous. The probe added here is what keeps that cadence covered rather than the issue's specific fault.

### Not re-derived, stated rather than hidden

#7056's largest RSS numbers (§3–§5, 272 MB → 421 MB on `w1_srv_pump`/`w2_srv_pump_while`/`w3_mono_sync`) came from three server-shaped workloads it wrote and never landed. There is nothing in the tree to re-run and a fresh reimplementation would not be comparable to them. The 12 probes are what replaced them.

Also not measured: probe 13's counters cross-host. They are deterministic across repeats, traced runs and rotations on this host, and nothing they depend on is host-derived above 2 GB of RAM (`gc_heap_budget_bytes` returns `None` for any budget ≥ 1 GB, so both this host and a GitHub runner take the historical constant path) — but the `shared_ci` profile gates them and that has not been demonstrated on a second machine.

---

## Verification

All green on this branch.

- **The 22 `lint`-job commands**, enumerated from `.github/workflows/test.yml` rather than by hand (fewer than 5 would mean the extraction failed): 22 ran, 0 failed, each command's own `$?` read directly and the runner exiting non-zero if fewer than 22 executed.
- `rustup run stable cargo fmt --all -- --check` — clean (it is the first of the 22).
- `python3 -m unittest tests.test_gc_ratchet` — **89 tests, OK** (17 new).
- `python3 benchmarks/gc_ratchet/gc_ratchet.py validate --scope all` — structurally valid.
- `python3 scripts/gc_gate_wiring_check.py` (+ `--self-test`) — clean.
- `cargo check --all-targets` — exit 0, 0 `error` lines. `cargo test -p perry-runtime --lib --no-fail-fast` — **1,917 passed, 0 failed**, exit 0. (Both run on the pinned host against this branch; this PR changes no Rust, so they are a no-regression check rather than a subject.)
- The quiet-host driver itself: `--check` before the pin (control), `--pin` after, `validate` after that.

**Host contention: none.** `rustc` and `cargo` counts were checked at 0 before every arm and every phase, CPU idle stayed in 94.6–95.8% for the whole session, and no other agent was building on the box. One correction to the brief's quiet gate: **1-minute load average on this mini idles at 1.3–1.8**, above the "< 1.5" bar, because a `WallpaperAerialsExtension` screensaver holds ~14% of one core continuously. `rustc == 0 && cargo == 0 && CPU idle >= 90%` is the predicate that actually discriminates here, and it is what the repo's own driver already gates on (`cpu_active <= 25%`).

### DIFF
--- a/.github/workflows/gc-ratchet.yml
+++ b/.github/workflows/gc-ratchet.yml
@@ -81,9 +81,9 @@ jobs:
       #
       # `--scope structural` is load-bearing and is the #7554 repair. This step
       # runs BEFORE the measurement step, so anything it fails on costs the
-      # entire run's coverage — twelve probes that never execute. That price is
-      # correct for "this artifact is unreadable or tampered with" and badly
-      # wrong for "one of its 144 cells is not bit-identical", which is what
+      # entire run's coverage — thirteen probes that never execute. That price
+      # is correct for "this artifact is unreadable or tampered with" and badly
+      # wrong for "one of its 156 cells is not bit-identical", which is what
       # actually happened: one cell's 6,768-byte spread failed this step and the
       # ratchet measured NOTHING on any branch for three days, while two GC
       # pacing changes (#7594, #7596) merged with hand-run A/Bs in its place.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1378
+**Current Version:** 0.5.1379
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1378"
+version = "0.5.1379"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "perry-ffi",
  "scraper",
@@ -5813,7 +5813,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-commander"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5821,7 +5821,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cron"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "chrono",
  "cron",
@@ -5831,7 +5831,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dayjs"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "chrono",
  "perry-ffi",
@@ -5839,7 +5839,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-decimal"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "perry-ffi",
  "rust_decimal",
@@ -5847,7 +5847,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dotenv"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "perry-ffi",
  "serde_json",
@@ -5855,7 +5855,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ethers"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "perry-ffi",
  "rand 0.10.1",
@@ -5863,7 +5863,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-events"
-version = "0.5.1378"
+version = "0.5.1379"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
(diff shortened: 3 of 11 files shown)


proggeramlug:
## Audit — merging as v0.5.1379. Plan items 9 and 10 both close.

### Item 9's answer is a negative result, and that is the useful kind

**The root lowering is not an RSS lever** — statepoint vs shadow, peak RSS **1.002**, retention **1.000**, 104 of 108 deterministic cells bit-identical. #7056 was written under the shadow stack and the fear was that #7370 invalidated its numbers; the measurement says the lowering was never the variable.

Proving the **arms actually differ per probe** before comparing them — `statepoint-example` + `addrspace(1)` roots on 12/12 by default, 0 of both under `PERRY_RS4GC=0` — is what makes a 1.002 ratio mean "no effect" rather than "I measured the same binary twice". That is the check this project has shipped four gates without.

The section-by-section disposition is better than a headline: §9's own recommendation already shipped (#7377), §7's 4.6×–54.8× false-retention table is **gone rather than shrunk** (#7657 removed the cause — excess 0.00% on all twelve), §6 survives with `PERRY_CONSERVATIVE_STACK_SCAN=full` taking `minor_cycles` to 0 on all twelve, and §3–§5 are honestly marked not-re-derivable because they came from three server workloads #7056 never landed. And the cap remains the whole lever at **1.911×** peak RSS.

### Item 10: the probe nearly measured nothing, and you caught it

> a large Eden on a *small* retained set runs **zero copying minors** — `arena_growth_full_escalation_due` escalates every minor to a full sweep above its 32 MB floor. My first draft did exactly that.

`KEEP` 8,192 → 0 minors, 131,072 → 1, 262,144 → **4**. A probe that ran zero copying minors while claiming to gate copying-minor cadence is the exact failure this repo has shipped four times (#6942/#6946, #7024, #7025) — caught here before it landed, by counting the subject rather than trusting the verdict.

**Shown able to fail, three ways**, and I re-verified the one I could locally: deleting the `// gc-ratchet-env:` directive reddens `test_the_shipped_large_eden_probe_still_declares_its_arm` (`{} != {'PERRY_GC_SCAVENGE_NURSERY_MB': '64'}`). Treating the env declaration **as a metric** is the right design — an arm whose environment can drift silently is an arm measuring something else.

The detail that justifies the whole gate: with the directive gone, retention/RSS/wall all read as *improvements* (−80%, −37%, −27%) because those bands are one-sided. A probe that gets faster and smaller by not doing its job is indistinguishable from a win.

**Keeping the inert perturbation in the writeup is the right call.** `PERRY_GEN_GC_EVACUATE=0` does not gate the copying minor (4 minors either way) — the perturbation was inert, not the gate, and you found that by counting copying minors before trusting the verdict. A less careful writeup would have listed it as a third successful sabotage.

### The `package.json` finding is the one I'd have been bitten by

> a probe compiled with a `package.json` in scope retains one more 1 MiB arena block at `gc()`

Nearly a false "−17.98% drift" on `09_try_catch_roots`, reproduced three ways and with build non-determinism ruled out. That is a measurement-environment dependency nobody would think to control for, and having it in the README, the plan, the changelog **and** the pin notes is proportionate.

### Corrections to my brief

**The mini idles at load 1.3–1.8**, above my "< 1.5" bar, because a `WallpaperAerialsExtension` holds ~14% of a core — so my gate would have either blocked forever or been ignored. `rustc == 0 && cargo == 0 && idle ≥ 90%` is the discriminator, and 94.6–95.8% idle throughout with zero contention is a better record than my rule would have produced.

`wt-scavtenure` subsumed and **measured** (`--check` reports all 144 pinned cells `ok`, failing only on the new probe's absence) rather than assumed.

**Gates:** 22/22 lint, fmt clean, `perry-runtime --lib` 1917, `test_gc_ratchet` **89 tests OK** (17 new), `validate --scope all` clean, `gc_gate_wiring_check` reports **6 gates main-line-reachable and able to fail**. The pin reproduced three independent times.

Probe 13's cross-host determinism under `shared_ci` is correctly stated as a limitation rather than a claim.

Issue #7664, opened 2026-08-08
gc: 21 unrooted-across-safepoint hazards in the NATIVE root lowering (the --max-unrooted budget's referent)


proggeramlug:
The `--statepoints` mode added in #7663 reads the **native** root lowering
(`gc.statepoint` `"gc-live"` relocation bundles) — the default on every target
whose frames the runtime can walk since #7370, and the lowering
`gc-root-dominance` was **not** checking until now.

Over the curated corpus (149 modules, 2452 functions, 30033 safepoints, 17478
with a live bundle, 40759 relocations) it reports **21 hazards under
`--moving-only`, all of class `unrooted`** — meaning nothing in the register's
cast chain appears in the safepoint's live bundle, so the collector neither
marks nor rewrites the object. Unfiltered (adding the RECLAIM half of the
invariant to the MOVE half): 1444, of which 1123 `unrooted` and 321 `stale`.

`gc-root-dominance-statepoints` carries `--max-unrooted 21` as a **ratchet**, not
an adjudication. This issue is that budget's referent: it can only be lowered,
and each shape below should get its own fix (and its own decrement) before the
job is promoted to a required context.

Reproduce:

```bash
cargo build --release -p perry -p perry-runtime-static -p perry-stdlib-static
./scripts/gc_root_dominance_corpus.sh ir-corpus-native --lowering native
python3 scripts/gc_root_dominance_check.py ir-corpus-native \
  --statepoints --moving-only --max-unrooted 999 --max-stale 999 -v
```

### 1. String-literal handle globals held across a call — 9 hits

`%r = load double, ptr @<mod>_.str.N.handle` into a register, then a safepoint,
then the register is used. The handle global **is** a registered root, so the
string is never swept — but an evacuating cycle **rewrites the global** and the
register keeps the pre-move address.

This is **#7240's exact shape**. That fix (`OperandProtection::Reload`,
re-emitting the load below the collection point) landed for the shadow-stack
lowering; these say it does not cover the native one.

```
test_gap_closures_ts::Counter__increment                    .str.5   across js_object_get_field_by_name_f64
test_gap_closures_ts::Counter__increment$pshape             .str.5   (same, repsel specialisation)
test_gap_object_get_own_property_descriptors_ts::main       .str.11  across js_closure_call1
test_gap_object_methods_ts::main                            .str.113 across js_closure_call1
test_gap_object_string_wrappers_ts::main                    .str.25  across js_closure_call1
[...]

Issue #7034, opened 2026-07-30
repsel row 1: scoping Ptr<Shape> beyond locals — measured, promotion is 0 on the motivating workload; build --opt-report (#6952) first


proggeramlug:
Scoping + design for **row 1** of the runtime-helper map: extending `Ptr<Shape>` beyond locals. Measurement done, unknowns named, nothing landed.

**Host for every number below:** M-series macOS arm64, `perry-dev` profile, worktree at `c6ed8175d`. **Load average was 7.57** for the whole session (parallel agents building), so **no timing measurement was taken** — every number here is either a static count or a binary size, both load-independent. See §8.

---

## 0. TL;DR — the brief's framing needs one correction before the plan makes sense

The task is framed as *extend a working representation to new positions*. Measured, that is not the situation:

| corpus | files | files with ≥1 `Ptr<Shape>` local | promoted locals |
|---|---|---|---|
| `benchmarks/suite/*.ts` (first 12) | 12 | **3** | 1 each |
| `batch.ts` — the object/property-heavy program from the helper map | 1 | **0** | **0** |

`PERRY_REPSEL_DEBUG=1` on `batch.ts` prints exactly one repsel line, and it is a Phase 3a canonical-`Str` local (`reportText`), not a shape. Independent confirmation: `PERRY_PTR_SHAPE_LOCALS=0` vs default produces **byte-identical** output (4,294,904 bytes, `__text` 3,804,424 both) — there is nothing to turn off.

So `Ptr<Shape>` today promotes on micro-benchmark-shaped code (one object, one contained local) and promotes **nothing** on the workload that motivates the whole row. The first step is therefore not "add positions" — it is **"make the proof survive an escape"**, because every realistic record escapes. That is the same work the brief asks for (params/returns/fields *are* the escapes), but it changes which step comes first and what each step should be measured against.

---

## 1. The concrete blocker at parameters

**It is not the ABI, not the collision guard, and not "nobody wrote it".** Two distinct things:

[...]

PR #7667, opened 2026-08-08
gc: close the strhandle, derived-mask and new.target unrooted hazards in the native lowering (#7664)

proggeramlug:
Closes the last piece of #7480 and `docs/engine-plan.md` item 6: **element
`Ptr<Shape>` for object-literal element types**.

## Measured

Pinned quiet mini, release, 7 interleaved rounds, 200k elements x 50 sweeps,
runtime-derived bounds, checksums equal in every cell, `rustc`/`cargo` at zero
and 94.6–94.9% idle before and after.

| kernel | perry before | perry after | node | bun |
|---|--:|--:|--:|--:|
| `keep: {v, w}[]` — #7480's own kernel | 408 ms | **12 ms** | 12 ms | 12 ms |
| `keep: Node[]` — what #7612 covered | 13 ms | 13 ms | 56 ms | 12 ms |
| region-local `{v, w}[]` (#7034 §3, E1–E5) | 15 ms | 14 ms | 12 ms | — |

**34x on the object-literal arm, to parity with both engines.** The
named-class arm is unchanged, and so is the region-local arm — that one was
already covered by `collectors/ptr_shape_elements.rs`, which is why this
change extends the versioned-loop consumer (the parameter/global case) rather
than that pass.

Both arms link **byte-identical** `libperry_runtime.a` / `libperry_stdlib.a`
(sha256 equal), so this is a codegen-only change and the meaningful control is
the emitted IR, below.

## What changed

`stmt/element_shape_loop.rs::element_class_name` resolved `Array(Named(C))`
only. It now also resolves a declared object type to the `__AnonShape_<hash>`
class its literals allocate, by matching the declared property order against
the module's anon shapes. The hash cannot be recomputed — `mint_anon_shape_class`
keys it on the literal's *inferred value* types (`{v: 1}` tags `i`, not `n`)
while the annotation says `number` — so the class is found, not derived.
Ambiguity **declines** rather than guessing, because `ctx.classes` is a
`HashMap` and "first match wins" would make the emitted code depend on
iteration order.

`receiver_class_name` is **not** widened. That is the #6377 blast radius #7612
deliberately refused; instead the clone is made self-contained. Its
`ElementShapeLoopFact` already carried the class name and packed slot index, so
the three sites that would otherwise re-derive the class from the receiver now
go through one predicate,
`expr::element_shape_loop_fact_for_property_get`:

* `lower_raw_f64_class_field_get_for_number_context` — the interception moved
  **above** the `receiver_class_name` gate;
* `type_analysis::is_numeric_expr`'s `PropertyGet` arm;
* `expr::binary`'s arithmetic-operand router (as a disjunct, rather than by
  widening `expr_may_return_boxed_value_from_raw_f64_fallback`, which would
  have been a lie — this read has no boxed fallback).

All three are scoped to the fast clone: outside one the fact vector is empty,
so `keep[j].v` anywhere else is byte-for-byte what it was.

## The issue's cost model was wrong, and the correction is load-bearing

#7480 records "no out-of-line guard *calls*, the cost is stacked inline
diamonds". The object-literal arm actually carried **three calls per
iteration**, the third being `js_dynamic_string_or_number_add`: with no
resolvable class the accumulator loses its numeric proof, so `+` is not an
`fadd`. The plan called that "a second, separable lever".

**It is not separable.** The clone is admitted only if it is provably
call-free (`LlBlock::contains_gc_unsafe_call` counts every non-`llvm.` call),
so resolving the element class *without* restoring the numeric proof emits the
clone, fails the call-free test, branches unconditionally to the slow arm and
buys exactly zero at a cost in code size. Both had to land together, and the
numeric claim inside the clone is stronger than the annotation it replaces:
the residual per-element check already proves `GC_OBJ_TYPED_LAYOUT_INTACT`,
i.e. that the slot holds a raw double.

### IR evidence (`--trace llvm`, release, `PERRY_NO_AUTO_OPTIMIZE=1`)

`sweep`, object-literal kernel, before — no clone at all, zero `fadd`:

```
js_typed_feedback_observe_property_get, js_typed_feedback_record_guard_pass,
js_dynamic_string_or_number_add, js_object_get_field_by_name_f64 x2,
js_object_get_field_ic_miss, ...           element_shape blocks: 0   fadd: 0
```

After — the fast clone is `gep` + load, a three-load residual check, `fadd`:

```
for.element_shape_fast.body:  gep/load elem, hdr mask+cmp, field_count cmp,
                              keys_array cmp, one branch
element_shape.load:           gep + load double + fadd double
FAST CLONE calls: []          FAST CLONE fadd: 1
```

The remaining calls in the function are the preheader's
(`js_array_refresh_local_head`, `js_array_ensure_element_shape`) and the slow
clone's, which is unchanged.

## Correctness against the #7660 shape

Every new gap case that reads a `{v, w}[]` crosses `MIN_ARRAY_CAPACITY`, so the
growth-forwarding stub the #7660 repair exists for is live on this arm too:
callee-built-and-returned, callee-filled-caller-owned, a 17-element prefix, and
a module-global array read from inside a function (the write-back's
`module_globals` arm). No new preheader or base derivation was added — the
element-shape preheader is the one #7660 fixed, unchanged.

The gap test also pins the hazards specific to this arm:

* an inline `rows.length` bound, which the matcher **rejects** (it is a
  `PropertyGet`, not an `Integer`/`LocalGet`), so a kernel written the obvious
  way gets no clone and must still print the same number;
* a layout downgrade under the clone, so the residual check's side exit
  re-runs the current iteration and the accumulator turns into a string exactly
  where JS says it does;
* two anon shapes sharing field names (`{v: number, w: number}` vs
  `{v: string, w: string}`), so a mis-resolved shape would cost the clone and
  never the answer;
* a mixed numeric/string shape the matcher must decline.

`test_gap_repsel_element_shape_loop_clone` is byte-identical to the Node 26.5.1
oracle on **both** arms, and an IR census confirms the new sections are live:
`sumRow`, `sumRow$spec_b_i32` and `sumGlobalRows` gain the clone, where on
`main` only `sumField` and `main` had it.

## An existing gate that could not fail

`fast_clone_slice` in `element_shape_loop_tests.rs` sliced from the first
*substring* occurrence of `for.element_shape_fast.cond` — which is the
`br label %…` terminator of the fast preheader, four lines above the slow
preheader — and every assertion made against the result is a negative
(`!fast.contains(" call ")`, `!fast.contains("js_array_get_f64")`, …). So the
IR census that exists to prove the clone is call-free had been **vacuous since
#7612**, on the code that then shipped a SIGBUS. It now finds the block
*definition* and asserts the slice contains the cloned body and its element
load, so it cannot pass on an empty subject again. Same family as #7024/#7025:
the gate ran, its subject did not.

## Tests

`crates/perry-codegen/src/stmt/element_shape_loop_tests.rs`, 7 new (17 total in
the module, all green). One positive — #7480's kernel reaches the clone, the
clone is call-free, and the accumulate is an `fadd` (the two halves asserted
together, because either alone is inert) — and six sabotage cases: an ambiguous
shape, a shape a field-type tie *can* break, an optional property, a shape no
literal allocates, a reordered shape, and a read outside the clone that must
stay on the by-name path.

## Gates

`lint`'s 22 extracted commands (22 extracted, 22 ran, 0 failed),
`cargo fmt --all -- --check`, `cargo check --all-targets`,
`cargo test -p perry-codegen --lib` and `-p perry-runtime --lib`
(`--no-fail-fast`), the 14 `native_root_coverage` tests, both
`gc_root_dominance` arms (`--moving-only --seeded-violations 40` and
`--statepoints --moving-only`), and the full gap suite against
`test-parity/gap_snapshot.json`.

## Validation

Release build, Node **26.5.1** (matches `.node-version`), `PERRY_GC_MOVING_LOOP_POLLS=1`
on both corpus arms.

* **`gc-root-dominance-statepoints`** — 7 unrooted / 0 stale, 40/40 seeded
  violations caught, empty allowlist honoured. Budget of 7 is *exact*: the same
  run with `--max-unrooted 6` exits 1, so the ratchet still has a referent it
  can hit.
* **`gc-root-dominance` (shadow arm)** — 0 violations over 2458 functions /
  9833 root stores, 40/40 seeded caught. Unchanged.
* **Checker `--self-test` + all three audits** (`--audit-alloc-re`,
  `--audit-poll-capable`, `--audit-immovable-sources`) — pass.
* **22 lint gates** enumerated from `test.yml` (including
  `check_file_size.sh`, `addr_class_inventory.py`, `gc_gate_wiring_check.py`) —
  0 failed; `cargo fmt --all -- --check` clean.
* **`cargo test -p perry-codegen --lib`** 748 passed / 0 failed (19 in
  `root_reload`, 14 `native_root_coverage`). **`cargo test -p perry-runtime
  --lib`** 1917 passed / 0 failed. **`cargo check --all-targets`** clean.
* **Gap suite** (506 tests, full run on the release build): 490 pass. Every
  reported failure is an existing `test-parity/gap_snapshot.json` /
  `known_failures.json` entry except one, `test_gap_zlib_4917_level`
  (`compile_fail`), which **fails identically on the branch point** — A/B'd with
  a `perry-dev` build of `7bde3de24` in the same tree. It is the auto-optimize
  relink dropping the zlib feature from its stdlib archive (host-local); with
  `PERRY_NO_AUTO_OPTIMIZE=1` the test compiles, runs, and is byte-identical to
  node. Clearing `target/perry-auto-*` does not change it, so it is not
  staleness either.
  * The harness also reports ten `node_fail -> parity_fail` status changes
    (`enum_forward_ref`, `backoff_options`, `cron_cronjob`, …). A codegen change
    cannot alter **node's** exit status, and each is an existing snapshot entry
    whose recorded reason (`ERR_UNSUPPORTED_TYPESCRIPT_SYNTAX`, or an npm import
    with no local `node_modules`) reproduces here — local classification, not a
    regression.
  * One improvement, reported without claiming it:
    `test_gap_iterator_helpers_2874: parity_fail -> pass`. Not investigated.
* `test_gap_class_expr_identity`, the regression the anchor bug caused, is now
  **byte-identical to node** and back to passing.

The snapshot is deliberately **not** regenerated in this PR: the only real
delta is host-local.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved garbage-collection safety when preserving constructor targets and reloading values derived from global handles.
  * Corrected register-root handling for derived values, preventing stale or unrooted references during native execution.
  * Expanded validation for global values, derived receivers, mutation scenarios, and call-based derivations.

* **Documentation**
  * Updated native lowering guidance and recorded the reduction in remaining unrooted-reference hazards from 21 to 7.

### DIFF
--- a/.github/workflows/gc-root-dominance.yml
+++ b/.github/workflows/gc-root-dominance.yml
@@ -464,17 +464,35 @@ jobs:
           #
           # BUDGETS. `unrooted` (the object is in NO live bundle at all, so
           # nothing marks or rewrites it) is the serious class, and it is a
-          # RATCHET at its measured value rather than a calibrated zero: this
-          # is a new instrument pointed at a lowering nothing has ever checked
-          # statically, and 21 is a population under triage, not a number
-          # anyone has adjudicated. It is a budget rather than allowlist
-          # entries for the reason `--stale-registers` records — 21 tombstones
-          # with no issue numbers would be worse documentation than one number
-          # that can only go down. Lower it as sites are fixed.
+          # RATCHET rather than a calibrated zero: this is a young instrument
+          # pointed at a lowering nothing had ever checked statically, and the
+          # remainder is a population under triage rather than a number anyone
+          # has adjudicated. It is a budget rather than allowlist entries for
+          # the reason `--stale-registers` records — tombstones with no issue
+          # numbers would be worse documentation than one number that can only
+          # go down. Lower it as sites are fixed.
           #
-          # The 21 are enumerated by shape in #7664, which is this budget's
-          # referent: a number with nothing behind it is the thing CLAUDE.md
-          # warns a threshold decays into.
+          # 21 at #7663. #7664 fixed 14 of them in `root_reload.rs` and
+          # `lower_call/new.rs` — the whole `strhandle` population (10), the
+          # `js_new_target_get` save/restore (1), and 3 of the 7 unmasked
+          # receivers — so the budget is 7. What remains, and why each is its
+          # own slice rather than a widening of the same fix:
+          #
+          #   4  unmasked, all PHI-MEDIATED. The stale value reaches its use
+          #      through a `phi`, and `root_reload` cannot insert above a phi;
+          #      the reload has to go in the PREDECESSOR, on the edge, which is
+          #      a different insertion model.
+          #   2  global (`@perry_global_*`). NOT reloadable: a module-level
+          #      variable is one the program assigns, so a re-read can observe
+          #      a later assignment instead of the value the call was given
+          #      (`operand_needs_root`). That population needs ROOTING, and
+          #      `a_module_global_is_not_a_reload_source` in root_reload.rs
+          #      pins the distinction so it cannot be widened away by accident.
+          #   1  capture, a `js_closure_get_capture_bits` read held across
+          #      `js_number_coerce`.
+          #
+          # #7664 stays open as this budget's referent: a number with nothing
+          # behind it is the thing CLAUDE.md warns a threshold decays into.
           #
           # `stale` (the object survives and is relocated, but a raw copy of
           # its pre-move address is used below) reads 0 today and is held
@@ -494,7 +512,7 @@ jobs:
             --min-statepoints 15000 \
             --min-live-bundles 8000 \
             --min-relocates 20000 \
-            --max-unrooted 21 \
+            --max-unrooted 7 \
             --max-stale 0 \
             --allowlist scripts/gc_root_dominance_allowlist.json \
             --seeded-violations 40 \
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1381
+**Current Version:** 0.5.1382
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1381"
+version = "0.5.1382"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1381"
+version = "0.5.1382"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1381"
+version = "0.5.1382"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1381"
+version = "0.5.1382"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1381"
+version = "0.5.1382"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1381"
+version = "0.5.1382"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1381"
+version = "0.5.1382"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1381"
+version = "0.5.1382"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1381"
+version = "0.5.1382"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1381"
+version = "0.5.1382"
 dependencies = [
  "anyhow",
  "perry-hir",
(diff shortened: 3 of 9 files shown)


proggeramlug:
## Audit — merging as v0.5.1382. Ratchet 21 → 7.

**The root cause is the finding.** All 14 fixes trace to one thing: `root_reload.rs` stated its rule over **the load's own register**, while the value at risk lived elsewhere — in a global the collector rewrites, or in a register *derived* from the load. Restating it over the value's **derivation**, extended only through pure bit ops whose every register operand is already in the same root's recipe, is a real repair rather than fourteen patches.

The `strhandle` group is the one that justifies the whole `--statepoints` mode: `load double, ptr @<mod>_.str.N.handle` is a **registered root that evacuation rewrites**, so a register loaded beforehand is stale. #7240 fixed this shape for call *operands* and never reached the ~194 sites loading a handle global directly — and the shadow-mode gate could not see it because the native lowering is what ships.

**IR evidence checked**: `Counter__increment` before had all three statepoints carrying an **empty live set** — the receiver in no bundle, so nothing marked *or* rewrote it. After, each carries a `"gc-live"` bundle and a `gc.relocate`, and the SET reads a mask re-derived from the relocated pointer. That the re-derivation is what puts the receiver in the bundle is the mechanism, not a side effect.

### The regression you introduced, caught, and pinned

Anchoring each derived value's window at **its own definition** was wrong: `main`'s class-object read has the scope-end shadow-slot clear landing *between* the load and the mask, so the walk re-read a nulled slot and `(makeAnon(77) as any).v` returned `undefined`.

Two things about that are worth more than the fix. **You found it by A/B against the branch point, not by the dominance checker** — and correctly noted the checker *cannot* find it, because it sees rooting, not value-correctness. And you pinned it with the **controlled twin** of the positive case: same frame plus one store, verdict flips 1 → 0. I verified the shipped version myself: `77 5` on both arms and under zeal, matching node.

### Both corrections accepted

The issue's shape-1 heading says 9 `strhandle` and its own list has 10 — the census is 10/7/2/1/1, so my brief inherited an off-by-one from the issue. And my "every fixed hazard's value must appear in the `gc-live` bundle after" is right for the derived-receiver half but **wrong by design for `strhandle`**: a handle global is a root the collector rewrites, so the fix is a re-read below the safepoint and the string never enters a live bundle in either arm. Stating that rather than quietly meeting the weaker bar is the difference between a fix and a fix that looks like one.

### The remaining 7, each correctly its own slice

4 phi-mediated (no instruction can go above a phi; the reload belongs in the predecessor **on the edge** — a different insertion model), 2 `@perry_global_*` (variables the *program* assigns, so a re-read can observe a later assignment — they need **rooting**, not reloading), 1 capture. Pinning the global case with `a_module_global_is_not_a_reload_source` so that widening the predicate is a test failure rather than a silent decision is exactly right, and I ran it: green.

Not allowlisted, #7664 stays open as the budget's referent, and the budget is **exact** — `--max-unrooted 6` exits 1.

**Gates:** 22/22 lint, fmt clean, `perry-codegen --lib` 748, `perry-runtime --lib` 1917 across 6 consecutive runs (the intermittent failure I hit on your branch was the third global-sink flake, `CLOSURE_PROPS` cleared by the GC guards — pre-existing, fixed in #7671, class filed as #7672). Gap 490 pass with the sole non-snapshot failure A/B-identical at the branch point.

PR #7669, opened 2026-08-08
perf(repsel): resolve object-literal element types in the element-shape loop clone (#7480)

proggeramlug:
Closes the last piece of #7480 and `docs/engine-plan.md` item 6: **element
`Ptr<Shape>` for object-literal element types**.

## Measured

Pinned quiet mini, release, 7 interleaved rounds, 200k elements x 50 sweeps,
runtime-derived bounds, checksums equal in every cell, `rustc`/`cargo` at zero
and 94.6–94.9% idle before and after.

| kernel | perry before | perry after | node | bun |
|---|--:|--:|--:|--:|
| `keep: {v, w}[]` — #7480's own kernel | 408 ms | **12 ms** | 12 ms | 12 ms |
| `keep: Node[]` — what #7612 covered | 13 ms | 13 ms | 56 ms | 12 ms |
| region-local `{v, w}[]` (#7034 §3, E1–E5) | 15 ms | 14 ms | 12 ms | — |

**34x on the object-literal arm, to parity with both engines.** The
named-class arm is unchanged, and so is the region-local arm — that one was
already covered by `collectors/ptr_shape_elements.rs`, which is why this
change extends the versioned-loop consumer (the parameter/global case) rather
than that pass.

Both arms link **byte-identical** `libperry_runtime.a` / `libperry_stdlib.a`
(sha256 equal), so this is a codegen-only change and the meaningful control is
the emitted IR, below.

## What changed

`stmt/element_shape_loop.rs::element_class_name` resolved `Array(Named(C))`
only. It now also resolves a declared object type to the `__AnonShape_<hash>`
class its literals allocate, by matching the declared property order against
the module's anon shapes. The hash cannot be recomputed — `mint_anon_shape_class`
keys it on the literal's *inferred value* types (`{v: 1}` tags `i`, not `n`)
while the annotation says `number` — so the class is found, not derived.
Ambiguity **declines** rather than guessing, because `ctx.classes` is a
`HashMap` and "first match wins" would make the emitted code depend on
iteration order.

`receiver_class_name` is **not** widened. That is the #6377 blast radius #7612
deliberately refused; instead the clone is made self-contained. Its
`ElementShapeLoopFact` already carried the class name and packed slot index, so
the three sites that would otherwise re-derive the class from the receiver now
go through one predicate,
`expr::element_shape_loop_fact_for_property_get`:

* `lower_raw_f64_class_field_get_for_number_context` — the interception moved
  **above** the `receiver_class_name` gate;
* `type_analysis::is_numeric_expr`'s `PropertyGet` arm;
* `expr::binary`'s arithmetic-operand router (as a disjunct, rather than by
  widening `expr_may_return_boxed_value_from_raw_f64_fallback`, which would
  have been a lie — this read has no boxed fallback).

All three are scoped to the fast clone: outside one the fact vector is empty,
so `keep[j].v` anywhere else is byte-for-byte what it was.

## The issue's cost model was wrong, and the correction is load-bearing

#7480 records "no out-of-line guard *calls*, the cost is stacked inline
diamonds". The object-literal arm actually carried **three calls per
iteration**, the third being `js_dynamic_string_or_number_add`: with no
resolvable class the accumulator loses its numeric proof, so `+` is not an
`fadd`. The plan called that "a second, separable lever".

**It is not separable.** The clone is admitted only if it is provably
call-free (`LlBlock::contains_gc_unsafe_call` counts every non-`llvm.` call),
so resolving the element class *without* restoring the numeric proof emits the
clone, fails the call-free test, branches unconditionally to the slow arm and
buys exactly zero at a cost in code size. Both had to land together, and the
numeric claim inside the clone is stronger than the annotation it replaces:
the residual per-element check already proves `GC_OBJ_TYPED_LAYOUT_INTACT`,
i.e. that the slot holds a raw double.

### IR evidence (`--trace llvm`, release, `PERRY_NO_AUTO_OPTIMIZE=1`)

`sweep`, object-literal kernel, before — no clone at all, zero `fadd`:

```
js_typed_feedback_observe_property_get, js_typed_feedback_record_guard_pass,
js_dynamic_string_or_number_add, js_object_get_field_by_name_f64 x2,
js_object_get_field_ic_miss, ...           element_shape blocks: 0   fadd: 0
```

After — the fast clone is `gep` + load, a three-load residual check, `fadd`:

```
for.element_shape_fast.body:  gep/load elem, hdr mask+cmp, field_count cmp,
                              keys_array cmp, one branch
element_shape.load:           gep + load double + fadd double
FAST CLONE calls: []          FAST CLONE fadd: 1
```

The remaining calls in the function are the preheader's
(`js_array_refresh_local_head`, `js_array_ensure_element_shape`) and the slow
clone's, which is unchanged.

## Correctness against the #7660 shape

Every new gap case that reads a `{v, w}[]` crosses `MIN_ARRAY_CAPACITY`, so the
growth-forwarding stub the #7660 repair exists for is live on this arm too:
callee-built-and-returned, callee-filled-caller-owned, a 17-element prefix, and
a module-global array read from inside a function (the write-back's
`module_globals` arm). No new preheader or base derivation was added — the
element-shape preheader is the one #7660 fixed, unchanged.

The gap test also pins the hazards specific to this arm:

* an inline `rows.length` bound, which the matcher **rejects** (it is a
  `PropertyGet`, not an `Integer`/`LocalGet`), so a kernel written the obvious
  way gets no clone and must still print the same number;
* a layout downgrade under the clone, so the residual check's side exit
  re-runs the current iteration and the accumulator turns into a string exactly
  where JS says it does;
* two anon shapes sharing field names (`{v: number, w: number}` vs
  `{v: string, w: string}`), so a mis-resolved shape would cost the clone and
  never the answer;
* a mixed numeric/string shape the matcher must decline.

`test_gap_repsel_element_shape_loop_clone` is byte-identical to the Node 26.5.1
oracle on **both** arms, and an IR census confirms the new sections are live:
`sumRow`, `sumRow$spec_b_i32` and `sumGlobalRows` gain the clone, where on
`main` only `sumField` and `main` had it.

## An existing gate that could not fail

`fast_clone_slice` in `element_shape_loop_tests.rs` sliced from the first
*substring* occurrence of `for.element_shape_fast.cond` — which is the
`br label %…` terminator of the fast preheader, four lines above the slow
preheader — and every assertion made against the result is a negative
(`!fast.contains(" call ")`, `!fast.contains("js_array_get_f64")`, …). So the
IR census that exists to prove the clone is call-free had been **vacuous since
#7612**, on the code that then shipped a SIGBUS. It now finds the block
*definition* and asserts the slice contains the cloned body and its element
load, so it cannot pass on an empty subject again. Same family as #7024/#7025:
the gate ran, its subject did not.

## Tests

`crates/perry-codegen/src/stmt/element_shape_loop_tests.rs`, 7 new (17 total in
the module, all green). One positive — #7480's kernel reaches the clone, the
clone is call-free, and the accumulate is an `fadd` (the two halves asserted
together, because either alone is inert) — and six sabotage cases: an ambiguous
shape, a shape a field-type tie *can* break, an optional property, a shape no
literal allocates, a reordered shape, and a read outside the clone that must
stay on the by-name path.

## Gates

Run on the pinned mini against this branch (`628db10e3`), each command's own
exit code, never a downstream `tail`'s.

| gate | result |
|---|---|
| `lint`'s extracted commands | **22 extracted, 22 ran, 0 failed** |
| `cargo fmt --all -- --check` | exit 0 |
| `cargo check --all-targets` | exit 0 |
| `cargo test -p perry-codegen --lib --no-fail-fast` | exit 0 — 748 passed, 0 failed |
| `cargo test -p perry-runtime --lib --no-fail-fast` | exit 0 — 1917 passed, 0 failed, 4 ignored |
| `native_root_coverage` | exit 0 — **14** tests |
| dominance `--moving-only --seeded-violations 40` | exit 0 — 149 modules, 40/40 seeded caught |
| dominance `--statepoints --moving-only` | exit 1 — **and identically on `origin/main`**, see below |
| gap suite (506 tests) | 490 pass / 15 parity_fail / 1 compile_fail — 12 snapshot divergences, **none attributable**, see below |

Zero `error[` in any build log; `Running unittests` present in each test log.

### The `--statepoints` arm is red on `main`, not on this branch

`--max-unrooted 21` reads **22** here. It reads **22 on `origin/main`
(3beef03b3) in the same session, on the same host, with the same corpus** — and
the 22 reported fingerprints are a **byte-identical set** between the two arms
(`diff` clean). So this change neither adds nor removes a hazard; the budget is
simply one over on `main` today, which is what open PR #7667
(`gc/7664-native-lowering-unrooted-hazards`) addresses. The seeded control
passed in both arms (40 planted, 40 caught, 0 missed), so the checker was live
for both readings.

### Gap-suite divergences from `test-parity/gap_snapshot.json`

506 tests: 490 pass, 15 parity_fail, 1 compile_fail. The harness reports 12
divergences from the committed snapshot — 1 regression
(`test_gap_zlib_4917_level: pass -> compile_fail`), 10 status changes
(`node_fail -> parity_fail`) and 1 improvement
(`test_gap_iterator_helpers_2874: parity_fail -> pass`).

**None is attributable to this change**, and that is measured rather than
argued. Every one of the 17 tests involved was compiled and run under **both**
arms: Perry's bytes are identical in all 17 (`TESTED=17 SAME=17 DIFFERENT=0`).

* `test_gap_zlib_4917_level` is the known host-local zlib link flake: compiled
  directly with either arm's binary it exits 0 and prints the right output,
  with and without auto-optimize. The failure lives in the harness's own
  auto-optimized relink on this host.
* The 10 `node_fail -> parity_fail` moves are all oracle-side classification
  (node still refuses e.g. `test_gap_4510_enum_forward_ref` with
  `ERR_UNSUPPORTED_TYPESCRIPT_SYNTAX`, exactly as its snapshot entry records) —
  a Perry-independent axis.

### Blast radius, measured

The change can only alter code inside an element-shape fast clone, so "which
files gain one" is the blast radius. Compiling with both arms and counting
`element_shape.loop.fast.preheader` sites:

* 43 assorted `test_gap_*` files: **zero clones in either arm**, zero
  compile-status differences — the change is inert there;
* 25 `test_gap_repsel_*` / `test_gap_specabi_*` files: exactly **one** file
  changes, `test_gap_repsel_element_shape_loop_clone`, **2 -> 6** clone sites.
  That file is the liveness control — a census that reports "no change" over a
  corpus with zero clones in both arms proves nothing.

## Summary by CodeRabbit

  * Improved optimized processing of numeric fields in homogeneous array loops.
  * Added support for safely optimizing compatible object-literal element types.
  * Preserved efficient fallback behavior for ambiguous, heterogeneous, or unsupported data.

* **Bug Fixes**
  * Improved handling of array growth, bounds, deleted fields, accessors, and changing element layouts.
  * Fixed validation of optimized loop regions.

* **Documentation**
  * Documented the completed element-shape optimization work and benchmark improvements.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1382
+**Current Version:** 0.5.1383
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1382"
+version = "0.5.1383"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1382"
+version = "0.5.1383"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1382"
+version = "0.5.1383"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1382"
+version = "0.5.1383"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1382"
+version = "0.5.1383"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1382"
+version = "0.5.1383"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1382"
+version = "0.5.1383"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1382"
+version = "0.5.1383"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1382"
+version = "0.5.1383"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1382"
+version = "0.5.1383"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1382"
+version = "0.5.1383"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1382"
+version = "0.5.1383"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1382"
+version = "0.5.1383"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1382"
+version = "0.5.1383"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1382"
+version = "0.5.1383"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1382"
+version = "0.5.1383"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1382"
+version = "0.5.1383"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1382"
+version = "0.5.1383"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1382"
+version = "0.5.1383"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1382"
+version = "0.5.1383"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1382"
+version = "0.5.1383"
 dependencies = [
  "perry-ffi",
  "scraper",
@@ -5813,7 +5813,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-commander"
-version = "0.5.1382"
+version = "0.5.1383"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5821,7 +5821,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cron"
-version = "0.5.1382"
+version = "0.5.1383"
 dependencies = [
  "chrono",
  "cron",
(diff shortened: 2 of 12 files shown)


proggeramlug:
## Audit — merging as v0.5.1383. **Plan item 6 closes; the last one.**

**34×, to parity with both engines.** I reproduced the direction on my own build: identical checksums, `main` 429 ms vs fix 16 ms with node at 14 on a *loaded* dev Mac — consistent with your quiet-mini 408 → 12.

### The correction that matters most: the levers are not separable

My brief said "plausibly two separable levers — say which you are not taking". You took both and explained why that was forced, not greedy:

> The clone is admitted only if provably call-free. Resolving the element class *without* restoring the accumulator's numeric proof emits the clone, hits `js_dynamic_string_or_number_add`, fails the call-free test, branches unconditionally to the slow arm, and buys zero at a cost in code size.

**Anything gated on call-freeness has this shape**, and that generalisation is worth more than the kernel. Splitting it would have produced a change that looks correct, measures nothing, and grows the binary.

You were also right to decline extending `collectors/ptr_shape_elements.rs`: its E1–E5 are a *containment* proof, and a parameter or global has a caller-side alias, so E2/E3 can never be established statically. The runtime invariant (#7496) is what answers that dynamically. **And you proved you hadn't rebuilt it** — the region-local row already reads 14–15 ms on `main`.

### The vacuous gate is the second finding, and it is a serious one

> `fast_clone_slice` sliced from the first *substring* match of `for.element_shape_fast.cond` — the `br label %…` terminator of the fast preheader, four lines above the slow preheader — and every assertion against the result is a negative.

So the IR census that exists to prove the clone is call-free **had been unable to fail since #7612 — on the code that then shipped the #7660 SIGBUS.** That is the #7024/#7025 family exactly, in the gate that should have been the last line of defence.

I verified the repair is load-bearing rather than cosmetic: reverting the slicer to the pre-fix substring form **reddens the suite**, so the new tests are sensitive to the slicing. Asserting the slice contains the cloned body *and* its element load is the right shape — it cannot pass on an empty subject again.

That CodeRabbit then found the same bug **twice more in your own new tests**, and you fixed both, is the honest version of this. The pattern is subtle enough to recur inside a change written specifically to fix it.

### Correctness against the shape that just crashed

No new preheader or base derivation, and the gap test crosses `MIN_ARRAY_CAPACITY` in every shape — callee-built, callee-filled, 17-element prefix, module-global read from a function — plus a mid-loop layout downgrade forcing the side exit, ambiguous same-named shapes, a deleted read-field, and an inline `rows.length` bound the matcher rejects. Byte-identical to node 26.5.1 on both arms.

Measured blast radius rather than asserted: 43 gap files gain zero clones in either arm; of 25 repsel/specabi files exactly one changes, 2 → 6, which is the liveness control.

### Two more corrections, both mine

`bench.sh`'s quietness probe is **self-matching** — `ps aux | grep -c '[r]ustc'` counts its own command line and always reports ≥1. And 414 ms was 408 ms; the direction and conclusion stand.

**Gates:** 22/22 lint, fmt clean, `perry-codegen --lib` 755, `perry-runtime --lib` 1917, `native_root_coverage` 14/14. The `--statepoints` budget you saw at 22-vs-21 was red identically on `origin/main`; #7667 merged just before this and takes it to 7.

Noted on process: the session scratchpad is shared between concurrent agents and another overwrote your `dominance.sh` mid-run. That is my staffing, not your error — I'll give agents private scratch paths.

PR #7670, opened 2026-08-09
gc: Layer 1 rooting slice 8 — the raw API becomes unreachable (#7615)

proggeramlug:
Slice 8 — the **last** slice of the Layer 1 emitter-migration campaign (#7615).
It reaches the terminal condition the plan stated, and the interesting parts of
the work are the three places the brief was wrong and the two places my own
verification was vacuous.

## The terminal condition, and why the file had to move

> the terminal condition is `expr/temp_root.rs` going `pub(in crate::rooting)` —
> the raw accessor unreachable, not merely uncounted.

**As literally spelled that is not expressible in Rust.** `pub(in path)`
requires `path` to be an ANCESTOR module of the item (E0742), and
`crate::rooting` is not an ancestor of `crate::expr::temp_root`. So the file
moved: the raw API is now `crate::rooting::temp_root`, declared with a
**private** `mod temp_root;` and with `pub(in crate::rooting)` on every
accessor. Either alone would do it; both are here because the module
declaration is one keyword from re-widening twenty-five items at once.

A raw call planted in a migrated module now **fails to compile**:

```
error[E0603]: module `temp_root` is private
   --> crates/perry-codegen/src/expr/binary.rs
    |   crate::rooting::temp_root::temp_root_push_i64(ctx, "0")
```

That is the difference between this and a ledger line, and it is the second of
ten sabotage arms below.

Two items keep `pub(crate)` and are re-exported from `rooting/mod.rs`. Neither
is an accessor and neither can be called in the wrong order: `TempRootPool`
(compile-time slot bookkeeping) and `expr_is_inert_primitive` (the purity
predicate `crate::loop_purity` shares).

**Fourteen entry points were deleted rather than narrowed**, because the
migration left them with no caller — `lower_exprs_rooted`,
`lower_operand_pair_rooted`, `any_later_ref_may_trigger_gc`,
`RootedOperands::is_rooted`, the whole `StoreOperandGuard` family, the whole
`RootedHandle` family, `temp_root_scope_begin`/`_end`. CLAUDE.md's kill-policy:
the losing mode should stop compiling.

## Three corrections to the brief

**1. `lower_string_method.rs` has 27 escape-hatch sites, not 3 — and it needed
a split too.** The brief's per-file counts come from a predicate that misses
`use crate::expr::temp_root::{…}` imports. Under the ledger's own predicate
(non-comment lines naming `temp_root`/`rooted_handle`) the real distribution is
`lower_string_method.rs` 27, `expr/static_field_meta.rs` 10, `lower_call/new.rs`
9, `expr/math_simple.rs` 9, `expr/binary.rs` 6, `expr/dyn_extern_i18n.rs` 4.
`loop_purity.rs` has **zero** (its one hit is a doc link), and two files the
brief omits — `gc_call_effects.rs` and `runtime_decls/arrays.rs` — have five
each, all of them the runtime SYMBOL NAME `"js_gc_temp_root_push"`.

**2. So `new.rs` was not the one blocker.** `lower_string_method.rs` was 1,957
lines against the 2,000-line cap with 27 sites to migrate — 43 lines of
headroom for five closure scopes. It got its own split, same as `new.rs`.

**3. The `dyn_extern_i18n.rs` lead is wrong about the CFG** — see below.

## Modules migrated vs classified decision-free

Telling them apart is half the work, because a ledger line on a module that
never had a decision to make looks substantive and asserts nothing.

**Eight listed** (seven load-bearing on the committed source):

| module | what it was | what it is |
|---|---|---|
| `expr/binary.rs` | five `lower_operand_pair_rooted` + `temp_root_release` pairs, each releasing on its own `return` | one `lower_rooted_dynamic_binary` over `with_operands_rooted` |
| `expr/math_simple.rs` | `MapSet`'s two operands with unequal windows, re-read at eight arm-specific points | a `RootedGroup`; `MapGet`/`MapHas` are the plain shape |
| `expr/static_field_meta.rs` | `ClassExprFresh`'s `rooted_handle_*` + a per-static store guard | a `RootedGroup` over the class object, nested `with_rooted_accumulator` for `caps_arr`, nested `with_operands_rooted` per symbol static |
| `expr/dyn_extern_i18n.rs` | the #7280 namespace-object build | `with_rooted_accumulator` |
| `lower_string_method.rs` | a receiver root spanning ~60 return paths | `open_rooted_group` |
| `lower_string_concat.rs` | four raw push/get/truncate/release spellings + two block-splitting diamonds | `with_operands_rooted` / `with_rooted_group` |
| `lower_call/new.rs` | `refresh_rooted_args` at three points under a scope marker over ~20 return paths | one escaping `RootedGroup`; the null marker slot is gone with it |
| `lower_call/new_alloc.rs` | (new file) | vacuous, listed anyway — an unlisted sibling of a listed module is where a raw push would go |

**Nine classified decision-free and deliberately NOT listed:** `expr/mod.rs` (a
module declaration and a field type — both gone with the move); the four
`FnCtx` constructors (`TempRootPool::default()` — constructing the pool is not
using it); `stmt/loops.rs` (one call to a purity predicate); `loop_purity.rs`
(a doc link); and `root_reload.rs` / `gc_call_effects.rs` /
`runtime_decls/arrays.rs` plus five test files, whose `js_gc_temp_root_*`
occurrences are **runtime symbol names**, not this module.
(`linker_temp_lifecycle_tests.rs`'s `temp_root_if_clang_available` is about a
temporary *directory*.)

## The three leads

* **`static_field_meta.rs`'s `caps_arr` — VERIFIED as a real accumulator shape
  with a provably empty window.** The `__perry_ctor_caps` array held the only
  reference to everything pushed so far across the next element's lowering, in
  a bare SSA register — #6951 exactly. But `captured_args` is built at ONE site
  (`lower/lower_expr/arm_class.rs:153`) as
  `ids.iter().map(|id| Expr::LocalGet(*id))`, and `expr_may_trigger_gc` answers
  `false` for every `LocalGet`. It is now a `with_rooted_accumulator` whose
  `protect` is **computed rather than assumed**: today `false`, IR byte for byte
  unchanged, and automatically correct the day a non-inert expression reaches
  the list.
* **`math_simple.rs`'s `ArrayMap` — VERIFIED live.** Below.
* **`dyn_extern_i18n.rs`'s `path_handle` — DISMISSED; the premise is wrong about
  the CFG.** Each `<prefix>__init()` is emitted into that iteration's **match
  block**, which branches straight to the join — no `__init` dominates any later
  use of `path_handle`. Along the fallthrough chain the only emissions between
  the handle's production and its last use are `js_get_string_pointer_unified`
  and `js_string_equals`, neither of which re-enters user code or enumerates an
  object — the standard `with_operands_rooted_across_call`'s doc sets for an
  emitted step (#7198). What that module DID have is the namespace-object
  accumulator, which is migrated.

## The live bug: `Expr::ArrayMap`

`arr.map(cb)` lowered the receiver, lowered the callback, and only THEN unboxed
the receiver — the unbox sat **below its own window** and masked a stale box
rather than repairing it (#7280 taxonomy (c)).

IR evidence, `main` baseline vs this branch, both `--profile perry-dev`,
`PERRY_NO_AUTO_OPTIMIZE=1`, `PERRY_GC_MOVING_LOOP_POLLS=1`, separate target
dirs. Module-global receiver, baseline:

```llvm
%r1 = load double, ptr @perry_global_m_ts__0        ; receiver
%r2 = call i64 @js_closure_alloc_singleton(...)     ; the window
%r5 = bitcast double %r1 to i64                     ; STALE
%r6 = and i64 %r5, 281474976710655
%r8 = call i64 @js_array_map(i64 %r6, ...)
```

this branch:

```llvm
%r1 = load double, ptr @perry_global_m_ts__0
%r2 = bitcast double %r1 to i64
store ptr addrspace(1) %rs4gc.s1, ptr %r3           ; ROOT STORE, dominates the window
%r7 = call i64 @js_closure_alloc_singleton(...)     ; the window
%r10.rs4p = load ptr addrspace(1), ptr %r3          ; RE-READ below it
%r12 = and i64 %r10, 281474976710655
%r14 = call i64 @js_array_map(i64 %r12, ...)
store ptr addrspace(1) null, ptr %r3                ; release, after the call
```

The same shape appears for a **class-field read** and for a **closure capture**
(`js_closure_get_capture_bits`, a raw `i64` — taxonomy (a), which
`root_reload` structurally cannot repair). Frame slot count is unchanged: the
pooled slot is reused by the watermark for the next temp.

**Where the window is NOT real, and why it matters:** for an array-typed
**local** receiver, codegen's `ptr addrspace(1)` retype pass rematerialises the
load from the local's own root slot at the use site, so the pre-fix code
re-read the receiver by accident. Measured, not assumed — the baseline arm
emits `%r65.rs4p = load ptr addrspace(1), ptr %r1` below the closure alloc all
by itself.

## Two ways my own verification was vacuous

Both were found by the sabotage arm (restore the pre-fix lowering, require red),
not by reasoning:

1. **Slice 7's `assert_operand_survives_the_window` cannot see this bug.** It
   compares the operand register's OWN definition line against the window, and
   for `ArrayMap` that register is `and i64 %stale, POINTER_MASK` — emitted
   *below* the window while masking a value loaded *above* it. "The unbox sits
   below its own window" is exactly the shape a one-level check misses. These
   tests chase the definition chain through pure bit-twiddling to the first real
   producer. (Slice 7's own tests are not wrong — their subjects are raw
   pointers stripped *above* the window — but the helper is weaker than its
   name.)
2. **A local receiver makes the test measure nothing**, per the paragraph above.
   The tests use a field read.

## ★ A trap worth recording: a sabotage harness that restores a `.bak`

`cp f f.bak; <patch f>; cargo test; mv f.bak f` leaves `f` with an **older**
mtime than the sabotaged build, so cargo keeps the sabotaged binary and every
later run measures the sabotage. It presented as an intermittent lowering-test
failure — 1 run in 10, then 20 in 20 — with **byte-identical IR between the
green and red runs**, which reads exactly like the process-global sinks #7665
fixed and is not. `touch` after the restore. Diagnosing from the wrong VALUE is
what settled it: "the producer is one line above the window" is precisely the
pre-fix lowering, not a race.

## Verification

* **node 26.5.1 byte-for-byte**, 27 lines of output, four arms — {`main`
  baseline, this branch} × {`PERRY_GC_MOVING_LOOP_POLLS` 0, 1} — over a probe
  covering `ArrayMap` (including evaluation order), the five dynamic binary
  arms, string append/concat/chain/methods, Map set/get/has (typed and
  generic), class expressions with statics + captures + `static {}` blocks,
  `new` with allocating and literal arguments, subclassing, and a 4,000-object
  churn between them. All four **MATCH**; baselines built in separate target
  dirs, `PERRY_NO_AUTO_OPTIMIZE=1` throughout.
* **22/22 lint-workflow commands** (extracted from `.github/workflows/test.yml`)
  plus `rustup run stable cargo fmt --all -- --check` and
  `./scripts/check_file_size.sh` (largest file now `rooting/mod.rs` at 1,880).
* `cargo test -p perry-codegen --lib --no-fail-fast` **747 passed**, and **20
  consecutive full-suite runs green** after the mtime fix.
* `cargo test -p perry-runtime --lib --no-fail-fast` **1,917 passed**.
* `cargo check --all-targets` clean.
* **14/14 `native_root_coverage`**.
* **Both dominance arms**, release build, subject-liveness asserted:
  * `--moving-only --seeded-violations 40`: 129/129 sources, 149 modules, **40
    planted / 40 caught / 0 missed**, 0 violations.
  * `--unrooted-allocas --moving-only`: 7,695 gc-capable allocas, **0**.
  * `--statepoints --moving-only --max-unrooted 21 --max-stale 0`: 2,458
    functions, 30,080 safepoints, 17,516 live bundles, 40,819 relocates, **40
    planted / 40 caught**, hazards **21 unrooted / 0 stale** — unchanged, and
    deliberately not touched while #7664's budget is being lowered separately.
* **Sabotage, 10 arms, 0 failures**: eight textual (plant a raw-API mention in
  each newly listed module, require the ledger red AND naming the `file:line`)
  and two visibility (plant a real raw call, require `E0603`).

## What this does NOT claim

The ledger's own caveat, drawn the hard way in slice 4: a listed module cannot
make an ORDERING mistake against the raw API, because it no longer names it. A
window with **no rooting decision at all** is invisible to that check, and the
only instrument for it is reading the module. 36 modules are listed;
`docs/engine-plan.md` item 7 records the remaining audit as the open half, with
#7640 holding the deferred sites, and Layer 3 as the rest of the rooting work.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compiler reliability for arrays, maps, strings, constructors, asynchronous operations, and dynamic property handling.
  * Prevented intermediate values from being lost during operations that may trigger memory collection.

* **Performance**
  * Optimized string concatenation and object construction, including fewer allocations in common cases.

* **Tests**
  * Added coverage for value preservation and allocation behavior across key operations.

* **Documentation**
  * Updated engineering documentation to reflect completion of the latest memory-safety milestone.


proggeramlug:
## Audit — merging as v0.5.1384. **The Layer-1 campaign is finished.**

### The terminal condition, verified two ways

I planted a raw call in a migrated module — once at top level (`math_simple.rs`), once in a nested submodule (`static_field_meta.rs`). Both: **`error[E0603]: module temp_root is private`**. The raw accessor is unreachable, not merely uncounted. That is the difference between a ledger line and a guarantee, and it is what eight slices were for.

**Your correction to the plan's own wording is right and I'd have got it wrong.** `pub(in path)` requires `path` to be an *ancestor* (E0742), and `crate::rooting` is not an ancestor of `crate::expr::temp_root` — so the condition as written is not expressible in Rust. Moving the file and using **both** a private `mod temp_root;` and `pub(in crate::rooting)` on all 25 accessors is the honest reading of the intent, and belt-and-braces is right for something one keyword from being undone.

**Fourteen entry points deleted rather than narrowed**, per the kill-policy. A narrowed-but-live API is a decision nobody has made.

### An integration break I caught building it

The rebase surfaced a genuine one: `#7667` added `new_target_save` using `crate::expr::temp_root`, while this slice moved the module — `error[E0433]: cannot find temp_root in expr`. Neither PR could see it; it exists only once both are on one tree. Fixed here (the module is a private sibling inside `crate::rooting`, so it names directly). Worth noting that **the first thing I ran was a plain `cargo check`, and that is what found it** — the sabotage would have "passed" against an already-broken baseline.

Also resolved two rebase conflicts: `root_reload.rs` (kept #7667's doc block, took your re-pointed intra-doc link) and `docs/engine-plan.md` (kept `main`'s item 6 as closed by #7669, took your item 7).

### Three corrections to my brief, all mine

**`lower_string_method.rs` has 27 escape-hatch sites, not the 3 I counted** — my grep missed `use crate::expr::temp_root::{…}` imports, so my census was wrong for every module that imports rather than fully-qualifies. It, not `new.rs`, was the real blocker at 1,957/2,000. `loop_purity.rs` has **zero** (its hit is a doc link), and two files I omitted have five each, all runtime symbol-name strings.

### The leads, and the discipline in how you closed them

`caps_arr` — **real accumulator shape, provably empty window**: `captured_args` is built at one site as `LocalGet`s, and `expr_may_trigger_gc` is false for every `LocalGet`. Making it a `with_rooted_accumulator` whose `protect` is *computed* — false today, IR byte-identical, correct by construction if the list ever gains a non-inert element — is better than either fixing or dismissing it.

`path_handle` — **dismissed, and my brief's premise was the thing that was wrong**: each `__init()` is emitted into that iteration's match block which branches straight to the join, so no `__init` dominates any later use.

`ArrayMap` — **confirmed live**, with the IR showing the receiver unboxed below its own window and the fix's root store dominating it.

### ★ The trap worth carrying

> `cp f f.bak; patch; cargo test; mv f.bak f` restores an **older mtime**, so cargo keeps the sabotaged binary.

It presented as a 1-in-10 lowering-test failure with **byte-identical IR between green and red runs** — indistinguishable from #7665's global sinks, which I had just spent hours on. `touch` after restore. And you settled it by diagnosing from the wrong *value* ("producer one line above the window" = exactly the pre-fix lowering) rather than from the timing — the same method that cracked all three global-sink flakes.

**Gates:** 22/22 lint, fmt clean, `check_file_size.sh` clean (largest now `rooting/mod.rs` at 1,931), `perry-codegen --lib` 761, `perry-runtime --lib` 1917, `native_root_coverage` 14/14, `cargo check --all-targets` clean.

What you explicitly do **not** claim — that a listed module cannot make an *ordering* mistake, while a window with no decision at all stays invisible to the ledger — is the right caveat to leave standing, with #7640 and Layer 3 named as the open half.

Issue #7018, opened 2026-07-29
gc: PERRY_GC_TRACE=1 SIGSEGVs under the evacuating arms — the liveness instrument crashes on the corpus file that most needs it


proggeramlug:
## Summary

`PERRY_GC_TRACE=1` **crashes the program it is supposed to be observing**. Under the evacuating arms, `test_gap_repsel_scalar_replaced_locals` SIGSEGVs with the flag set and passes without it.

This matters more than an ordinary flag bug: `PERRY_GC_TRACE=1` is the tool this whole campaign relies on to prove a GC arm was *live*. `PERRY_GC_DIAG` under-reports (it only prints on some paths), so `GC_TRACE`'s one-marker-per-completed-cycle output is the primary evidence that a green cell means anything. A green cell from an inert arm is exactly the false confidence that invalidated months of this project's earlier results (#6942, #6946, #6950).

So the instrument crashes on the case that most needs instrumenting.

## Reproduction

```
PERRY_GC_HEAP_LIMIT=8 PERRY_GC_INCREMENTAL=0 PERRY_CONSERVATIVE_STACK_SCAN=off \
PERRY_GC_TRACE=1 ./test_gap_repsel_scalar_replaced_locals
```

- **with** `PERRY_GC_TRACE=1` → SIGSEGV
- **without** it → passes, byte-identical to the pinned Node 26.5.0 oracle, with `copied_objects` in the 261 251 – 379 072 range

## Not a regression from the hoist

Confirmed **pre-existing**: reproduced identically on a rebuilt `main` without PR #7013, same arms, same failure. Reported during #7013's validation and deliberately not folded into that PR.

## Status of this report

Measured by the agent that landed #7013; **I have not independently re-verified it**. First step for whoever picks this up is to reproduce both arms above before doing anything else.

## Why it is plausible

The tracer runs inside a collection, walking structures the collector is mid-way through mutating — during an *evacuating* cycle, objects are being copied and forwarded while the trace reads them. A tracer that dereferences a forwarded-but-not-yet-rewritten pointer, or reads a header on an object already evacuated, would fault exactly here. That is a hypothesis, not a diagnosis; the stack will say.

Note the interaction with **#6998** and **#6995**: both concern root words that are not what the decoder expects, and `is_plausible_heap_addr` is the shared predicate. If the tracer classifies raw words the same way, it may share their root cause.

## Why it should be fixed rather than worked around

[...]

PR #7674, opened 2026-08-09
fix(test-isolation): the GC guards' clear can no longer reach another test's data (#7672)

### LINKED ISSUE #7672: Process-global side tables are cleared by test guards that readers are not required to take — three flakes in two days
Three flaky tests in two days, all the same architecture: **a process-global side table, plus per-suite guards that clear it, plus tests that read it without taking the clearing lock.**

| fixed in | table | how it presented |
|---|---|---|
| #7665 | `opt_report`'s `Mutex<Vec<Entry>>` | `rows.len() == 2` failed at **3** — a neighbour's entry in the snapshot |
| #7665 | `ext_registry`'s `USED_PROVIDERS` | "empty after an unlisted symbol" failed with `ioredis` still present |
| #7671 | `closure/dynamic_props.rs`'s `CLOSURE_PROPS` | a static method value read back as `TAG_UNDEFINED` |

In every case the lock existed and was correct **for the tests that took it**. The failure is that nothing requires a *reader* to take it, so the defence is opt-in and the opt-in is invisible at the read site.

Each was exposed by an unrelated PR adding tests and changing the parallel schedule — never introduced by it. That is the expensive part: the author of the exposing PR spends the diagnosis, and the natural conclusion ("my change broke something") is wrong.

## The remaining exposure, surveyed

Files whose tests read closure dynamic props, and whether they take `crate::gc::global_side_table_test_lock()`:

| guarded | file (test count) |
|---|---|
| 3 of 4 | `closure/dynamic_props.rs` |
| 2 of 9 | `object/global_this_webassembly.rs` |
| 2 of 2 | `object/native_module_stream.rs` (after #7671) |
| **0** | `array/tests.rs` (65), `node_stream_tests.rs` (42), `node_submodules/tests.rs` (33), `object/instanceof.rs` (5), `object/native_module/constants.rs` (4), `value/to_string.rs` (4), `node_stream_state_tests.rs` (4), `tls.rs` (3), `object/native_module/callable_export_arity_table.rs` (3), `perf_hooks.rs` (2), `object/prototype_chain.rs` (2), `object/native_call_method.rs` (2), `value/dynamic_object.rs` (2), and 6 more with 1 each |

Roughly 180 unguarded tests. Most will never bite — the hazard needs a test that reads a **persisted** prop across a window a guard can land in — but "most" is doing real work in that sentence, and the three we found were each a surprise.

## Options, in rough order of preference

1. **Make the clear observable.** Have `test_clear_closure_side_tables()` record that it ran, and give readers a cheap assertion that no clear happened during their read. Turns a silent wrong value into a named failure at the point of damage, without serialising anything.
2. **Invert the default.** Make the reader-side accessor take the lock, with an explicit opt-out for the guards that must clear. Costs some parallelism; removes the class.
3. **Scope the tables per-thread in tests.** Largest change, and `CLOSURE_PROPS` is keyed by heap address with a GC scanner attached, so this needs care.
4. Blanket-lock the ~180 tests. Rejected as the first move: it serialises a large part of the suite for a hazard whose incidence we cannot yet bound.

Option 1 is the one that pays for itself even if the class never bites again, because it converts "intermittent, diagnosed by luck" into "fails with a message naming the clearer".

Context: #7665, #7671, and `gc/tests/support.rs`'s own comment, which documents the hazard but does not enforce anything.

proggeramlug:
Closes #7672.

## The class, and why the three fixes so far did not close it

`gc::tests::support::reset_copying_nursery_runtime_test_state()` runs from `GcTestIsolationGuard` and `CopyingNurseryTestGuard`, on whatever libtest thread happens to construct one, and calls **20** `test_clear_*` helpers that empty **process-global** side tables. The guards serialize against each other, and against the handful of tests that remember to take `crate::gc::global_side_table_test_lock()`. Nothing requires a **reader** to take it, so the defence is opt-in and the opt-in is invisible at the read site.

| fixed in | table | presented as |
|---|---|---|
| #7665 | `opt_report`'s row sink | `rows.len() == 2` failing at **3** |
| #7665 | `ext_registry`'s `USED_PROVIDERS` | "empty" failing with `ioredis` present |
| #7671 | `closure`'s `CLOSURE_PROPS` | a static method read back as `TAG_UNDEFINED` |

Each was *exposed* by an unrelated PR that changed the parallel schedule, never introduced by it, so the exposing author paid the diagnosis and the natural conclusion ("my change broke something") was wrong.

## What this changes: storage, not locking

**A lock cannot close this.** The damage window is "between this test's write and this test's read", and only the test knows that span. An accessor that takes a shared lock for one call does not cover it. A lock that covers it has to be taken by the test — which is the opt-in the class is made of, on a reader population the issue counts at ~180 and growing.

So the storage moves. `per_test_global!` (`crates/perry-runtime/src/per_test_global.rs`) expands to:

* **outside a test build** — the plain `static` it replaced, byte for byte. `test_clear_*` is `#[cfg(test)]` and never runs in a shipped runtime, so the hazard and the fix are both confined to tests.
* **inside a test build** — a `PerThread<T>` that derefs to a per-thread instance. libtest runs one thread per test, so "per thread" and "per test" coincide, and a guard on thread U empties U's instance, which already holds only U's entries.

That is exactly the isolation the clear was reaching for. **Call sites do not change** — `SYMBOL_REGISTRY.lock()` and `CLASS_PROTOTYPE_OBJECTS.read()` keep working through `Deref`, so this is a declaration-site change and not a 300-site rewrite.

**37 statics converted across 13 files** (26 at first, plus the class-registry and soak-found ones), covering every family the guards clear: closure dynamic props (3), symbol side tables (5), the class-registry `RwLock`s (7), the timer queues (3), the object-constant caches and `GLOBAL_THIS_PTR` (9), `geisterhand` (2), `ui_text_registry` (2), the `console.log` singleton.

## What holds the line, and how each half is shown able to fail

**`scripts/global_sink_isolation.py`, in `lint`** (a required context; ~1s, no compiler). It derives the clear list from `reset_copying_nursery_runtime_test_state`'s own source, follows one level of same-file accessors (`test_clear_closure_side_tables` names no static at all — it goes through `get_closure_props()`), resolves each identifier in its own module first, and classifies the storage as `thread_local!` / `per_test_global!` / `Mutex<()>` lock / bare `static`. A bare static fails unless allowlisted with an issue, and **an allowlist entry that matches nothing also fails**. The burden lands on the ~20 table authors, who are finite and gated, instead of on the readers, who are not.

Its `--self-test` has 9 checks: a bare table is rejected, a converted one and a thread-local one are not, allowlisting silences exactly one entry, a stale entry fails, a renamed or empty reset function raises instead of passing vacuously, and the parsers are run against the **real** tree (≥10 helpers parsed, `CLOSURE_PROPS` classified `per_test`, `ARGUMENTS_OBJECTS` classified `thread_local`) so a regex that stops matching cannot make the gate green.

**`gc::tests::global_sink_isolation`, 9 tests.** Each plants the #7671 shape — write on this thread, run the guards' real clear on another thread, read back — for one `test_clear_*` helper. They deliberately do **not** take the global lock: taking it would test the opt-in rather than the isolation. Every probe asserts its subject was live before the clear, so "survived" is never confused with "was never installed".

* `the_probe_catches_a_bare_process_global_sink` runs the identical procedure against a canary declared as a plain `static` and **requires the wipe to be observed**. A green file means the detector works, not that nothing was tried.
* `every_covered_clear_helper_is_still_called_by_the_guards` `include_str!`s `support.rs` at compile time so a probe for a helper that no longer runs cannot sit green forever.

**Sabotage, measured, not asserted.** With the macro's `#[cfg(test)]` arm reverted to the pre-#7672 bare static — one edit, every table at once — **7 of 9 tests fail**, each with the message naming the cause:

```
closure_side_tables_survive_a_guard_clear_on_another_thread ... FAILED
symbol_side_tables_survive_a_guard_clear_on_another_thread ... FAILED
timer_queues_survive_a_guard_clear_on_another_thread ... FAILED
class_registry_tables_survive_a_guard_clear_on_another_thread ... FAILED
object_cache_roots_survive_a_guard_clear_on_another_thread ... FAILED
geisterhand_registry_survives_a_guard_clear_on_another_thread ... FAILED
ui_text_registry_survives_a_guard_clear_on_another_thread ... FAILED
test result: FAILED. 2 passed; 7 failed
```

(`error[` count 0 and `Running unittests` present on that run — the sabotage compiled and executed rather than failing to build.)

## Allowlist: 4 entries, each with a reason

* `YOUNG_PIN_EVER` (#7645) — deliberately process-wide and monotone; the skip path is masked if it is not. No reader outside the guard.
* `SYMBOL_REGISTRY`, `WELL_KNOWN_SYMBOLS` (#7672) — **read, never written** by the clear: they are the process-lifetime registries the per-thread `SYMBOL_POINTERS` rebuild is derived from, and their symbols are `Box::leak`ed, so process-wide identity is the correct one (`Symbol.for("x") === Symbol.for("x")` depends on it).
* `plugin::REGISTRY` (#7672) — the one table already defended soundly, and the model the survey pointed at: the guard takes `PLUGIN_REGISTRY_TEST_LOCK` and all 9 plugin tests take the same lock. The split lock domain is the root cause in all three fixed flakes; this domain is not split.

## Validation

* `cargo test -p perry-runtime --lib --no-fail-fast` — **25 consecutive runs, 25 green, 0 vacuous** (each run checked for `Running unittests` and for `error[` before its result was counted). 1926 passed / 0 failed / 4 ignored, ~8.5 s per run; the conversion costs no measurable wall time.
* `cargo check --all-targets` — clean.
* All **24** `lint` gate commands green, including the new one.
* `rustup run stable cargo fmt --all -- --check` — clean.

`timer.rs` sat three lines under the 2000-line cap, so the macro takes an optional trailing `;` and the three timer tables are declared on one line each — the file is unchanged at 1997 lines.

## Also found while surveying (not fixed here)

The same architecture exists outside the guards' clear list, with the same split-lock-domain signature:

* **`async_hooks`** `HOOKS` / `RESOURCES` / `NEXT_ASYNC_ID` sit under **four disjoint serialization domains**, and `gc/tests/alloc.rs:836` takes none of them. `resource_ids_are_monotonic_even_without_hooks` asserts `b.async_id == a.async_id + 1` — a wrong-VALUE flake waiting for a schedule change.
* **`tui::state::SLOTS`** is cleared under three different locks; `alloc_returns_sequential_handles` asserts `h0 == 0, h1 == 1, h2 == 2`.
* **`agent_dispatch_tests.rs`**'s private `TIMER_QUEUE_TESTS` lock protects the timer queues the guards also clear — the `opt_report` / `ext_registry` shape exactly.

These are candidates for the same treatment; they are called out rather than folded in because none of them is on the guards' clear path, which is what this PR's gate covers.

---

## Update: two more instances, found by soaking this very fix

The first 25-run soak was **25/25 green**. A second, 22-run soak produced **two** reds — which is exactly why the ≥20-run bar exists, and why the first all-clear was not one. Both are the same class, neither is reached by any `test_clear_*` helper, and both were diagnosed from the wrong VALUE rather than the timing:

* **`gc::barrier::GENERATED_WRITE_BARRIERS_EMITTED`** — owned by **two guards under two different locks** (`CopyingNurseryTestGuard` under the copying-nursery isolation lock, `GeneratedWriteBarrierTestGuard` under `GENERATED_BARRIER_TEST_LOCK`) and read by every runtime write barrier holding neither. `sabotaged_parent_gate_strands_a_young_child_the_shipped_gate_keeps` failed with `missing_edges=1 ... slot_page_ever_dirty=false` — the barrier *did not fire*, because another thread's `GeneratedWriteBarrierTestGuard::inactive()` had zeroed the flag mid-test.
* **`tui::tree`'s `NEXT_HANDLE` / `REGISTRY`** — no clear, no lock, and `register_increments_handle` asserts `h2 == h1 + 1` plus an exact registry length.

Both are converted, both have a dedicated regression test, and both fail under the same one-edit sabotage (now **9 of 11** tests red).

**The macro is renamed `per_test_global!`** — for what it does, not for the sharpest instance of what it defends against, since two residents are cleared by nothing. Naming it after the clear would have made these two look like misfits rather than the same defect. The gate now audits the guards' own module alongside the clear list, which is what makes the write-barrier flag visible to it at all.

## Two ways this gate could itself have gone quiet — both closed, both self-tested

* The `per_test_global!(...)` **paren form** (used in `timer.rs` to stay under the 2000-line cap) was invisible to a `{`-only matcher, silently taking the three timer tables to `(no static storage)`. A gate that stops matching reports zero hazards and exits 0 — indistinguishable from a clean tree. Reverting the delimiter fix now fails a self-test case by name.
* A **classified-statics floor** (93 today, floor 60) fires when the matchers rot, instead of reporting clean.
* A `Mutex<()>` serializer is classified as a *lock*, never a hazard: making one per-thread would turn it into a no-op — the opposite of the fix.

Self-test is now **15 checks**, allowlist **5 entries** each with a reason.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -333,6 +333,27 @@ jobs:
           python3 scripts/check_test_registration.py --self-test
           python3 scripts/check_test_registration.py
 
+      # #7672: the GC test guards CLEAR ~20 process-global side tables from
+      # whatever libtest thread constructs a guard, and nothing requires a
+      # READER to take the clearing lock. Three flakes in two days came from
+      # that (#7665 x2, #7671), each exposed by an unrelated PR that changed the
+      # parallel schedule, so the author of the exposing PR paid the diagnosis.
+      #
+      # The class is fixed by storage, not by a lock: `guard_cleared_global!`
+      # gives each thread its own table in a test build. This gate derives the
+      # clear list from the guards' own source and fails on any bare `static`
+      # left behind, so a NEW sink cannot be added quietly and a new READER
+      # never has to remember anything. Its allowlist entries each cite an
+      # issue, and an entry that matches nothing fails too.
+      #
+      # Pure text, ~1s, no compiler — in `lint` because `lint` is a required
+      # context (hazard 2), and `!cancelled()` for the reason given above.
+      - name: Guard-cleared global sinks
+        if: ${{ !cancelled() }}
+        run: |
+          python3 scripts/global_sink_isolation.py --self-test
+          python3 scripts/global_sink_isolation.py
+
   # ---------------------------------------------------------------------------
   # Clippy — enforces the deny-level lints in [workspace.lints] (root
   # Cargo.toml). `cargo clippy` exits nonzero only on `deny` lints, so
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -339,7 +339,7 @@ jobs:
       # that (#7665 x2, #7671), each exposed by an unrelated PR that changed the
       # parallel schedule, so the author of the exposing PR paid the diagnosis.
       #
-      # The class is fixed by storage, not by a lock: `guard_cleared_global!`
+      # The class is fixed by storage, not by a lock: `per_test_global!`
       # gives each thread its own table in a test build. This gate derives the
       # clear list from the guards' own source and fails on any bare `static`
       # left behind, so a NEW sink cannot be added quietly and a new READER
@@ -348,7 +348,7 @@ jobs:
       #
       # Pure text, ~1s, no compiler — in `lint` because `lint` is a required
       # context (hazard 2), and `!cancelled()` for the reason given above.
-      - name: Guard-cleared global sinks
+      - name: Per-test global sinks
         if: ${{ !cancelled() }}
         run: |
           python3 scripts/global_sink_isolation.py --self-test
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1386
+**Current Version:** 0.5.1387
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1386"
+version = "0.5.1387"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1386"
+version = "0.5.1387"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1386"
+version = "0.5.1387"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1386"
+version = "0.5.1387"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1386"
+version = "0.5.1387"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1386"
+version = "0.5.1387"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1386"
+version = "0.5.1387"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1386"
+version = "0.5.1387"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1386"
+version = "0.5.1387"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1386"
+version = "0.5.1387"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1386"
+version = "0.5.1387"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1386"
+version = "0.5.1387"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1386"
+version = "0.5.1387"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1386"
+version = "0.5.1387"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1386"
+version = "0.5.1387"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1386"
+version = "0.5.1387"
 dependencies = [
  "block2",
  "objc2",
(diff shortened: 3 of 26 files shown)


proggeramlug:
## Audit — merging as v0.5.1387

**You were right and my recommendation was wrong, for the reason you gave**: "readers get a cheap assertion" cannot satisfy my own second requirement, because the assertion *is* an opt-in on the same unbounded reader population. I proposed detection where prevention was available. Partitioning the storage removes the class; observing the clear only tells you after the fact, and only for readers who remembered to ask.

The three properties that make it the right shape:

- **Production is unchanged by construction**, not by argument — the `#[cfg(not(test))]` arm expands to `$vis static $name: $ty = $init;`, literally the plain static. I built `-p perry-runtime-static -p perry-stdlib-static` to confirm the shipped path still links.
- **Zero call-site churn.** `Deref` means `SYMBOL_REGISTRY.lock()` and `CLASS_PROTOTYPE_OBJECTS.read()` keep working across all 37 statics in 13 files.
- **The burden inverts**: ~180 readers who must remember an opt-in become ~20 table authors, and `global_sink_isolation.py` derives the clear list *from the guards' own source* and fails on a bare `static`. I ran it: 0 hazards, 5 allowlisted, 93 statics classified.

**Sabotage verified independently**: reverting the `cfg(test)` arm to a plain static reddens **8 tests**, `error[` 0, `Running unittests` present, and 1928 green on restore.

### Finding five instances, two of them by soaking your own fix

`GENERATED_WRITE_BARRIERS_EMITTED` is the one that matters most — two guards, two locks, read by neither, presenting as `slot_page_ever_dirty=false`, i.e. **the barrier silently not firing**. That is not a test-hygiene bug wearing a correctness costume; it is the reverse.

### The intermittency evidence is the part I'd hold up as the model

**25/25 green — "a false all-clear"** — then 20/22, which is what found the two extra instances. Naming the first soak as a false all-clear rather than banking it is the whole discipline. ~102 runs across four soaks.

And the honesty on #7683: 1 in ~102 on the branch vs 0 in 25 on `main` **does not distinguish at that rate, and you say so**. What carries it is that every frame is in code this PR does not touch. That is a correctly-scoped claim.

### Two more

**#7018 closed by refutation** — doesn't reproduce in 20 runs, and structurally cannot: `PERRY_GC_TRACE` never dereferences a heap object; the issue describes `gc/trace.rs`, the *marking* tracer. A refuted issue closed with evidence is worth as much as a fixed one.

**All four arms in `gc_matrix_inert_arms.txt` were live**, and CI's `gc-stress` on `main` has been red for it at 41/58 cells with four `STALE-REGISTRY` lines. Unasked-for and exactly right to fix.

Your harness nearly lying twice — counting an external SIGTERM as a product flake, then SIGSEGV as an external kill — and your lint gate silently losing three tables to the `per_test_global!(...)` paren form are both worth the record. The delimiter-agnostic re-do with a 60-static floor and both-way self-test is the fix.

**Gates:** 24/24 lint (the extraction is 24 now, up from 22), fmt clean, `perry-runtime --lib` 1928, `perry-codegen --lib` 778, `cargo check --all-targets` clean, `global_sink_isolation.py` self-test and gate both clean.

Issue #7507, opened 2026-08-06
ci_e2e_scope: map crates/perry-codegen/src/** to the six in-process root-lowering suites (~6.5s) — blocked on them being green


proggeramlug:
Split out of #7493. The pin is fixed; the reason nobody noticed for weeks is not.

## The gap

`.github/workflows/test.yml`'s `e2e-scoped` job (via `scripts/ci_e2e_scope.py`) runs the integration suites **a diff names**:

```
changed crates/<pkg>/tests/<suite>.rs   ->   cargo test -p <pkg> --test <suite>
```

`ci_e2e_scope.py` deliberately refuses the source→suite direction, and says why: there is no coverage data to map it with, and a crate-level map (`perry-codegen` -> all 163 `perry` suites) is exactly the full run the scoping exists to avoid. That reasoning is right in general and wrong for one narrow case.

#7370 changed `crates/perry-codegen/src/`. It named no suite. `shadow_slot_hygiene` went 0/12 and nothing was red until someone ran the nightly tier by hand. The per-PR `cargo-test` gate is `--lib --bins`, so it saw none of it.

## Why this particular set is the exception

The six root-lowering suites are not like the 163 `perry` suites the general rule is protecting against. They are **in-process compiles of hand-built HIR with `emit_ir_only: true`** — no `perry compile` subprocess, no link, no runtime. Measured (arm64 macOS, debug, `--test-threads=1`):

| suite | tests | wall |
|---|---|---|
| `shadow_slot_hygiene` | 12 | 0.17 s |
| `scalar_replaced_slot_roots` | 11 | 0.18 s |
| `temp_root_argument_temporaries` | 7 | 0.08 s |
| `temp_root_operand_temporaries` | 19 | 0.39 s |
| `native_proof_buffer_views` | 32 | 0.50 s |
| `native_proof_regressions` | 255 | 5.13 s |

~6.5 s of test time for all six. The cost is the build, which `e2e-scoped` already pays whenever any perry-codegen suite is in scope, on a warmed sccache.

## Why this is NOT being done in #7493

**A new gate has never been green.** Fifteen tests across these suites are red right now for three causes #7493 does not own — #7494, #7503, #7504 (plus #7505 and #7506). Adding `crates/perry-codegen/src/** -> these six suites` today makes `e2e-scoped` red on the majority of perry-codegen PRs.

[...]

Issue #7504, opened 2026-08-06
scalar_replaced_slot_roots counts js_shadow_slot_bind module-wide, and #7487's pooled temp roots emit it too — 7 tests measure the wrong slots


proggeramlug:
Split out of #7493. Six failures in `crates/perry-codegen/tests/scalar_replaced_slot_roots.rs` and one in `crates/perry-codegen/tests/shadow_slot_hygiene.rs` that survive the lowering pin — they fail identically under `NativeRootsPin::shadow()` and under `PERRY_RS4GC=0`, so they are not #7370's.

## Symptom

`scalar_replaced_slot_roots` (all shadow-pinned after #7493):

* `bind_is_hoisted_into_the_entry_block_ahead_of_the_storing_loop`
* `every_store_into_a_hoisted_scalar_slot_shades_its_value`
* `repeated_stores_into_one_scalar_slot_bind_once`
* `numeric_only_scalar_replaced_literal_emits_no_entry_rooting`
* `numeric_only_scalar_replaced_object_emits_no_rooting`
* `numeric_only_scalar_replaced_array_emits_no_rooting`

`shadow_slot_hygiene`:

* `flat_const_row_aliases_do_not_reserve_shadow_slots`

The last two `numeric_only_*` were **passing before #7493** — vacuously. They assert `bind_calls(&ir) == 0`, and `bind_calls` counts `call void @js_shadow_slot_bind(`, which the native-roots default emits for nothing. Pinning them to the lowering they measure is what made them fail.

## Cause

`bind_calls` and the `js_shadow_frame_enter(i32 N)` slot count are **whole-module** counters, and #7487 gave temporaries a claim on the same two things. A pooled temp root reserves a frame slot via `reserve_shadow_slot()` and emits `emit_shadow_slot_bind_ptr` at every store — the identical call a scalar-replaced field's slot emits.

Worked example. `numeric_only_scalar_replaced_object_emits_no_rooting` builds `const p = {x: 1, y: 2}; console.log(p.x, p.y)` and demands zero binds. The emitted module contains three:

```llvm
%r11 = alloca i64                              ; the console arg-array accumulator
%r3  = call ptr @js_shadow_frame_enter(i32 2)
%r10 = call i64 @js_array_alloc(i32 2)
call void @js_shadow_slot_bind(i32 1, ptr %r11)
%r35 = call i64 @js_array_push_f64(i64 %r34, double %r33)
[...]

Issue #7503, opened 2026-08-06
temp_root_{argument,operand}_temporaries assert the pre-#7487 FFI spelling: 10 tests fail and the 8 that pass are vacuous — the #6951 contract has no working coverage


proggeramlug:
Split out of #7493. These failures are **not** the #7370 lowering default — `PERRY_RS4GC=0` moves `temp_root_argument_temporaries` not at all (3/7 either way). They are #7487's, and the passing half of both suites is worse than the failing half.

## Symptom

`crates/perry-codegen/tests/temp_root_argument_temporaries.rs` — 3 passed, 4 failed
`crates/perry-codegen/tests/temp_root_operand_temporaries.rs` — 13 passed, 6 failed

Failing (all with the same shape, `ir.contains("call i32 @js_gc_temp_root_push")` is false):

* `console_argument_accumulator_is_temp_rooted`
* `array_literal_roots_elements_before_an_allocating_element`
* `symbol_keyed_typed_array_read_is_still_temp_rooted`
* `unproven_key_typed_array_read_is_still_temp_rooted`
* `map_set_key_is_rooted_across_an_allocating_value`
* `concat_accumulator_is_rooted_and_written_back`
* `constructor_arguments_are_rooted_across_the_instance_allocation`
* `wtf8_literal_operand_is_rooted_not_merely_reused`
* `the_new_instance_is_rooted_across_the_constructor_body`
* `the_object_assign_accumulator_is_rooted_across_each_source`

## Cause

#7487 re-lowered temp roots onto pooled frame allocas: `push` became a store plus the same bind a named local's store emits, `get` a load, `truncate` a slot zero. `js_gc_temp_root_push` / `_get` / `_set` / `_truncate` survive **only on the FFI fallback arm**, which is taken when `reserve_shadow_slot()` returns `None` — and it returns `Some` under *both* lowerings. So the spelling every assertion in these two files searches for is now emitted essentially never.

The contract itself is intact. `console.log("label", churn())` today:

```llvm
%r2 = alloca ptr addrspace(1)                 ; the pooled temp slot
%r1 = call i64 @js_array_alloc(i32 2)
%rs4gc.s1 = inttoptr i64 %r1 to ptr addrspace(1)
store ptr addrspace(1) %rs4gc.s1, ptr %r2     ; push
%r7.rs4p = load ptr addrspace(1), ptr %r2     ; re-read before use
%r8 = call i64 @js_array_push_f64(i64 %r7, double %r6)
[...]

PR #7675, opened 2026-08-09
test(codegen): make the stale codegen-test contracts able to fail again (#7505, #7504, #7503, #6988)

### LINKED ISSUE #7505: assert_buffer_store_uses_dynamic_fallback greps the whole module for 'getelementptr inbounds i8' — the shadow frame emits one, so 16 tests misfire under PERRY_RS4GC=0
Split out of #7493, which worked around this with a pin. The workaround is correct for today's default but the underlying assertion is wrong, and it will misfire again.

## The assertion

`crates/perry-codegen/tests/native_proof_regressions.rs`:

```rust
fn assert_buffer_store_uses_dynamic_fallback(ir: &str) {
    assert!(ir.contains("call void @js_buffer_set"), ...);
    assert!(!ir.contains("getelementptr inbounds i8"), ...);   // <— this one
}
```

Same shape at `native_proof_regressions.rs:846` and twice in `native_proof_buffer_views.rs` (`:606`, `:682`).

The intent is "no *native buffer* GEP was emitted for this store". What it actually says is "this **module** contains no `getelementptr inbounds i8` **anywhere**".

## Why that breaks

The shadow-stack lowering's inline slot addressing (#7088) computes the slot address with `getelementptr inbounds i8`, for reasons that have nothing to do with a buffer. So under `PERRY_RS4GC=0` — the sweep a GC engineer runs, and the one #7493's own diagnosis was built on — sixteen tests report a stale proof that was never emitted:

`native_proof_regressions/invalidation.rs` (15): `localset_invalidates_native_i32_alias_facts`, `update_invalidates_native_i32_alias_facts`, `localset_invalidates_min_length_facts`, `localset_invalidates_active_bounded_buffer_index_facts`, `inner_loop_bounded_buffer_fact_is_removed_after_outer_fact_invalidation`, `localset_invalidates_buffer_view_local_length_sources`, `update_invalidates_buffer_view_local_length_sources`, `negative_loop_counter_does_not_emit_inbounds_buffer_gep`, `decrementing_loop_update_does_not_emit_inbounds_buffer_gep`, `body_counter_mutation_does_not_emit_inbounds_buffer_gep`, `inclusive_length_loop_does_not_emit_inbounds_buffer_gep`, `inclusive_local_length_bound_does_not_use_local_length_bound_fact`, `negative_loop_counter_does_not_use_local_length_bound_fact`, `body_mutation_of_local_bound_does_not_use_local_length_bound_fact`, `negative_loop_counter_does_not_use_min_length_bound_fact`

`native_proof_buffer_views.rs` (1): `loop_length_bound_does_not_prove_multibyte_buffer_read_inbounds`

#7493 pinned all sixteen to `NativeRootsPin::native()`. That is honest — the assertion is only valid under native roots today — and it stops the false alarm. It does not make the assertion right.

## Why the pin is not the fix

The proxy is unsound in the other direction too. It cannot distinguish:

* a native buffer GEP for the store under test (what it means to catch), from
* a native buffer GEP for an *unrelated* proven access elsewhere in the same module (a false positive it would also report), from
* any other `inbounds i8` the backend happens to emit (what actually bit us).

Any future pass that emits an `inbounds i8` — a string-data address, an inline-cache slot, a native field offset — turns all sixteen red for a reason unrelated to their subject, exactly as the shadow frame did.

## What to build

Scope the search to the store/read site:

* cut the function body under test (the file already has `enclosing_function` / `init_ir` style helpers), then
* find the `js_buffer_set` / typed-array call the fixture drives, and assert that the operand feeding it was *not* produced by an `inbounds` GEP off the buffer's data pointer — i.e. follow the data-flow rather than grepping the text. `temp_root_operand_temporaries.rs` already walks defs backwards this way (`def_of` / `reaches_alloc`), so the technique is in-tree.

Then remove the sixteen `NativeRootsPin::native()` pins: the fact-invalidation contract is lowering-independent and should be asserted that way.

## Acceptance

* The sixteen pass under BOTH lowerings, unpinned.
* Sabotage-checked: restoring the stale proof (letting the invalidated fact through) must fail each of them.
* `grep -rn 'contains("getelementptr inbounds i8")' crates/perry-codegen/tests/` returns nothing module-scoped.

proggeramlug:
Closes #7505, #7504, #7503, #6988. Partially addresses #7507 (see "What is NOT here").

Five codegen tests asserted something that is no longer the contract, or asserted it in a way that could not fail. Every replacement below was watched go red against a planted defect; the sabotages are named per issue.

## #7505 — `assert_buffer_store_uses_dynamic_fallback` had no subject

`!ir.contains("getelementptr inbounds i8")` over the **whole module**. Any unrelated `inbounds i8` satisfied it, and the shadow-stack lowering's own inline slot addressing (#7088) emits exactly that text — so under `PERRY_RS4GC=0` sixteen tests reported a stale proof that was never emitted. #7493 pinned them to `NativeRootsPin::native()`, which stopped the false alarm without making the assertion right.

It now follows the data flow. `native_proof_support::native_buffer_element_geps` accepts only an `inbounds` GEP taken off a pointer **loaded out of a Buffer view's DATA slot**, where a data slot is identified by either shape codegen materialises one with (the unboxed `and`/`inttoptr`/header-skip store, or the `!invariant.load` header length read). It is scoped to `@probe` via a `define`-line anchor, not a substring — #7669's `fast_clone_slice` is why that is spelled out. `assert_no_native_buffer_element_access` **panics when the function lowered no buffer view at all**, so "there is no native GEP" and "there is nothing here to have one" stay distinguishable.

All sixteen pins are gone; the tests pass under both lowerings.

* Sabotage (run): deleting the `slots.contains(&slot)` filter — i.e. degrading the reader back into the module-wide grep — turns **19** tests red under `PERRY_RS4GC=0`, including the self-test that plants the shadow-frame GEP shape.
* Non-vacuity: `invalidation::the_native_buffer_gep_detector_fires_on_a_proven_store` compiles the same store with its proof intact and requires the reader to SEE the GEP, under both lowerings, then requires the invalidated twin not to.

## #7504 — the bind count measured the wrong slots

`bind_calls` counted `js_shadow_slot_bind` **module-wide**, and since #7487 a pooled temp root emits the identical call. Every fixture ends in `console.log(o.a, o.b)`, whose argument accumulator contributes three binds. So `== 0` was a claim about the accumulator and `== 1` a coincidence.

New `perry_codegen::testing::root_slots` keys every bind and every root-shading barrier by the **entry alloca it names**, classifies that alloca (`Value` = named local / scalar-replaced field, `TempRoot` = #7487's pool) and **panics on one it cannot classify** — so a third slot family goes red naming itself rather than being folded into a total. The three `numeric_only_*` tests additionally assert `temp_root_slot_binds > 0`: their zero is only interesting in a module where binds exist.

* Sabotage A (run): stubbing out `root_scalar_replaced_slot`'s `root_entry_alloca` call — #6968 reverted — turns 7 positives red.
* Sabotage B (run): disabling the `expr_is_known_non_pointer_shadow_value` gate — #6997 reverted — turns all three `numeric_only_*` negatives red. That is the direction that was silently broken, so it is the one that most needed the proof.

`flat_const_row_aliases_do_not_reserve_shadow_slots` is renamed and re-pointed. **Its two causes separate cleanly and the temp pool is not one of them**: measured here the pool contributes 0 binds and 0 reservations, and all three reserved slots are locals'. The assertion it made was also not the contract — `kernel` lowers to a real heap array, so `krow`/`k` hold heap pointers and leaving them unrooted would be #6968. Its second assertion had gone toothless independently: it forbade `js_shadow_slot_set(i32 1` while #7013 moved the traffic to `js_shadow_slot_bind`. It now asserts the hygiene property the suite exists for — every reserved slot is bound or cleared, i.e. no #7184-shaped reserved-but-untouched entry — and refuses an empty frame.

## #7503 — the temp-root contract had no working coverage in either direction

#7487 re-lowered temp roots onto pooled frame allocas; `js_gc_temp_root_push`/`_get`/`_set`/`_truncate` survive only on an FFI fallback arm neither shipped lowering takes. Ten assertions failed and eight `!ir.contains(…push)` negatives held for **every program in the language**.

New `perry_codegen::testing::temp_slots` reads slot traffic in **both** spellings (the RS4GC retype preserves register names, which is what lets one reader serve both) and states the contract as a claim about a value: `assert_rooted_across(ir, producer_result, consumer)` proves *this* producer's result reached a slot and *this* call read its operand back out of it. That is strictly stronger than the call-existence check it replaces.

`temp_root_operand_temporaries.rs` is 19/19 green under **both** lowerings, unpinned.

* Sabotage (run): removing the `ctx.block().store(…)` in `temp_slot_store` turns all 4 positives in the moved suite red and correctly leaves the 3 negatives green.
* Unit-level sabotages, committed: `an_unrooted_accumulator_is_caught` (pre-#6951 SSA threading) and `rooted_but_not_re_read_is_caught` (#7114: rooted, location rewritten, consumer still reads the stale register — which `contains("…temp_root_push")` passes happily).

## #6988 — the contract now runs per-PR

`tests/temp_root_argument_temporaries.rs` is **deleted**; its seven tests live in `crates/perry-codegen/src/temp_root_coverage/`, an in-crate `#[cfg(test)]` module following #7653's `native_root_coverage` pattern, so they run in the required `cargo-test` job instead of the nightly-only tier (#5960). Each runs once per lowering via `under_both_lowerings`.

## What is NOT here (#7507)

`SOURCE_SUITE_MAP` is added to `scripts/ci_e2e_scope.py` with the **three** suites that are green today, plus the on-disk cross-check (an entry naming a missing suite FAILS the scope step) and four new self-test cases including a sabotage of that cross-check against the real repo.

`native_proof_regressions` and `native_proof_buffer_views` are **deliberately absent**. Each carries two pre-existing failures that are not this PR's — verified against pristine `a853135aa`, same four before and after — in typed-array artifact records and integer-modulo / typed-f64 clone lowering. A new gate has never been green, and wiring in a suite that is red on arrival makes `e2e-scoped` red on most perry-codegen PRs; since that job is not in branch protection it cannot block anything, which is CLAUDE.md hazard 2 with extra steps. The comment in the map says so and says to add them in the commit that turns them green. #7507 stays open for that.

## Verification

* `cargo fmt --all -- --check` clean; all **21** script gates from `test.yml`'s `lint` job pass (22 enumerated, minus fmt run separately).
* `cargo test -p perry-codegen --lib` 776 passed; `-p perry-runtime --lib` 1917 passed; `cargo check --all-targets` no errors.
* The 14 `native_root_coverage` tests pass.
* Five perry-codegen suites under the native default AND `PERRY_RS4GC=0`: `shadow_slot_hygiene` 12/12, `scalar_replaced_slot_roots` 11/11, `temp_root_operand_temporaries` 19/19, `native_proof_buffer_views` 34/36, `native_proof_regressions` 258/260 — the 2+2 being the pre-existing failures above.
* Dominance arms not re-run: every changed file is a test, a `#[cfg(test)]`/`feature = "testing"` module, or a CI script. `src/lib.rs`'s only change is a `#[cfg(test)] mod` line. The shipped compiler is unchanged, so the emitted IR those arms read cannot move.

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for memory safety, temporary values, buffer operations, and garbage-collection rooting.
  * Added validation across both supported runtime lowering modes.
  * Improved checks for correct value preservation, release behavior, and avoidance of unnecessary memory operations.
  * Replaced brittle assertions with more precise, function-scoped verification.

* **Chores**
  * Updated automated test selection so relevant integration suites run when code-generation behavior changes.
  * Removed obsolete test coverage and outdated assertions.


proggeramlug:
## Audit — merging as v0.5.1385

Four issues resolved, one deliberately partial, and **three corrections to my brief** — all three of which found something nobody had named.

### The replacements can fail; I checked one myself

I sabotaged #7504's replacement independently: neutering `root_scalar_replaced_slot` (reintroducing #6968) → **7 red, `error[` 0, test binary reached**, and green again after restore. I used `touch` on the restore, per the mtime trap you found yesterday.

Classifying each alloca and **panicking on an unclassifiable one** is the right shape — a counter that silently absorbs an unrecognised slot is how the original got into this state.

### Your three corrections

**`flat_const_row_aliases` was wrong in a way my brief didn't reach.** The temp pool adds **zero** there — 0 binds, 0 reservations, all 3 slots locals' — so the premise was wrong; and the property it asserted isn't the contract, since `kernel` lowers to a heap array and leaving `krow`/`k` unrooted would be #6968. Then the third defect: its negative grepped `js_shadow_slot_set(i32 1` while **#7013 moved the traffic to `js_shadow_slot_bind`**, so it had been unable to fire independently of everything else. Re-pointing it at the hygiene property with a `reserved > 0` guard is better than repairing what it used to say.

**#7503 lists 10 failures; there are 11.** `string_literal_array_element_is_re_derived_below_an_allocating_element` is missing from the issue and failed for a *different* reason — #7114 admits two discharges (re-derive from immutable storage, or re-read from a root), the lowering switched to the second, and the test asserted only the first. Accepting either while still rejecting register reuse is the correct contract.

**The FFI fallback arm is not covered because `reserve_shadow_slot()` returns `Some` under both shipped lowerings.** Per the kill-policy that arm should be deleted — and leaving it out of a test-only PR while flagging it on the issue is the right call, not a dodge.

### CodeRabbit found an eleventh vacuity, in the PR that removes that defect class

`temp_root_slots` matched the whole alloca def text (`Some(&"alloca i64")`), so **one `align 8` away from matching nothing — which would empty the slot set and make every `assert_no_temp_rooting` vacuous, silently.** Matching by alloca *type* through a shared `alloca_type`, with two tests that fail against the old comparison, is the fix. That this shape appeared *inside* a change written to eliminate it is the honest measure of how easy it is.

Both declines are right. Broadening `derives_from` to every operand would make the **positive** "was re-read" clauses easier to satisfy — trading a loud false alarm for a possible silent pass, which is the wrong direction on this PR specifically. And deduplicating 2 of 6 `entry_opts` copies hides the hazard rather than removing it.

### #7507 left partial, correctly

Wiring in `native_proof_regressions` / `native_proof_buffer_views` while each carries 2 failures would be **adding a suite that is red on arrival** — exactly the hazard the issue warns about. I verified the four are pre-existing: identical test names on pristine `main` (`i32_counter_mod_unsafe_or_nonliteral_divisors_keep_frem`, `typed_f64_receiver_method_clone_raw_loads_after_composed_guards` — the latter is #7506), with the branch passing *more* (258 vs 253, 34 vs 30). Naming them in the map's comment with "add them in the commit that turns them green" is the right handoff.

### #6988 done the right way

Deleting `tests/temp_root_argument_temporaries.rs` and moving its 7 tests into `src/temp_root_coverage/` under `#[cfg(test)]` puts them in the **required `cargo-test` job** instead of the nightly/tag-only tier — #7653's pattern, and the whole point of the issue. Running each once per lowering via `under_both_lowerings` also closes the slice-8 trap where a rooting assertion written against shadow-frame IR measures nothing on the statepoint default.

**Gates:** 22/22 lint, fmt clean, `perry-codegen --lib` 776, `perry-runtime --lib` 1917, `ci_e2e_scope --self-test` clean, and the three rewritten suites 19/11/12 green. Dominance arms correctly not re-run — every changed file is a test, a `#[cfg(test)]` module, or a CI script, so the shipped compiler is unchanged.

proggeramlug:
Thanks — two of these are the exact defect class this PR exists to remove, and both are now fixed with a regression test each.

**Fixed: the `align` suffix (`temp_slots.rs`).** Correct, and the failure mode is the silent one: `temp_root_slots` returns empty, so every `assert_no_temp_rooting` in the tree passes for a program that roots. Now matched by alloca **type** through a shared `root_slots::alloca_type`, applied in all three places that were comparing whole def text (`temp_root_slots`, `root_slots::classify`, `root_slots::value_slot_barriers`). Two tests added that fail against the old exact compare: `an_aligned_alloca_is_still_recognised_as_a_temp_slot` and `an_aligned_alloca_is_still_classified`. Both assert the substitution actually applied, so they cannot pass by not testing anything.

**Fixed: the write-back exact compare (`temp_root_coverage/mod.rs`).** Also correct — it was the one clause in that test not going through the same derivation rule as its neighbours. Now `slot_holding(&ir, &push).as_deref() == Some(slot.as_str())`.

**Fixed: the redundant load branch (`zero_seeded_slots`).** Verified: `, ptr %s` already covers `%d = load i64, ptr %s` and both other load spellings, so the second pass re-inserted what the first caught. Removed, with a comment saying why there is deliberately no load-specific pass — an extra branch that reads as coverage it does not add is the same category of problem as the rest of this PR.

**Declined: broaden `derives_from` to every operand.** You noted the current failure direction is a false alarm rather than a silent pass, and that is exactly why I am leaving it. `derives_from` backs the *positive* clauses — "this consumer's operand WAS re-read from the slot". Following every `%` operand with a visited set makes that claim easier to satisfy, so an unrelated register reaching the ancestor would let a genuinely unrooted operand pass. On a PR whose entire subject is assertions that could not fail, trading a loud false alarm for a possible silent pass is the wrong direction. If the false alarm ever fires in practice, the right fix is to name the operand index at the call site, not to widen the walk.

**Declined here: dedup `entry_opts` / `module_with_init`.** Fair point, but there are already ~6 copies of this harness across `crates/perry-codegen/tests/` (`scalar_replaced_slot_roots`, `shadow_slot_hygiene`, both `native_proof_*`, and the two temp-root files). Deduplicating 2 of 6 does not remove the "add every new `CompileOptions` field twice" hazard, it just makes it less visible — and doing all 6 is a repo-wide refactor that does not belong in a PR about test-assertion correctness. Worth its own change.

Issue #6998, opened 2026-07-29
codegen: pointer_locals types Expr::Uint8ArrayGet as Number unconditionally — a symbol-keyed read binds a heap value into a local with no shadow slot


proggeramlug:
## Summary

`collect_pointer_typed_locals`'s `expr_value_type` classifies `Expr::Uint8ArrayGet` as `Type::Number` **unconditionally**, with no regard for the key kind (`crates/perry-codegen/src/collectors/pointer_locals.rs:232`):

```rust
Expr::Number(_)
| Expr::Integer(_)
| Expr::Uint8ArrayLength(_)
| Expr::Uint8ArrayGet { .. }       // <-- every key kind, including symbol keys
| Expr::BufferLength(_)
| Expr::BufferIndexGet { .. }
...  => Some(Type::Number),
```

But `Uint8ArrayGet` is **not** always numeric. PR #6997's analysis enumerated three heap-capable arms of that node — a symbol key, an unproven key routed to `js_typed_array_index_get_dynamic`, and an i32-context `js_object_get_index_polymorphic` arm — and gated its own rooting decision on the *lowering* rather than the node kind for exactly this reason.

## Consequence

`const x = u8[Symbol.iterator]` binds a **heap value** into a local that the collector has classified as non-pointer. A non-pointer local gets **no shadow slot**, so it is not a GC root: the value is live in the program and invisible to the collector. Same class as #6951, rarer shape.

Note the classification is also what sizes the shadow frame, so this is adjacent to #6995 (frame sized by `HashMap` len while indices come from a counter) — both are ways the shadow-slot map can end up not describing the frame it is used with. They are separate defects; fixing one does not fix the other.

## Status — verified statically, reachability NOT verified

I confirmed by reading `origin/main` (`aa1c15028`) that the match arm is unconditional. **I have not exhibited a TypeScript program that produces a symbol-keyed `Uint8ArrayGet` bound to a local**, and that is the first thing to establish. The obvious candidate is `const it = u8[Symbol.iterator]`, but AST→HIR lowering may route symbol keys to a different node before this collector sees them — check before assuming.

The soundness principle PR #6997 established should be applied here too: **decide from the lowering, not from the node kind or the declared type.** Perry does not enforce type annotations at runtime, so a declared type is never sufficient evidence that a value cannot be a pointer.

## Verification guidance

A reproducer must be checked on the evacuating precise-roots arm, where a missing root actually bites:

```
[...]

PR #7676, opened 2026-08-09
fix(gc-matrix): the collect arms require a productive cycle, and the numarray probe reaches the collector (#7016, #7017)

### LINKED ISSUE #7016: gc-matrix: test_gap_repsel_p4a3_numarray_growth never crosses a GC trigger, so all 19 of its GC cells assert nothing
Split out of #6978 (fixed in #7015).

`test_gap_repsel_p4a3_numarray_growth` is the one corpus row that is still `UNVER` on every GC arm of `scripts/gc_repsel_matrix.sh --arms all --pressure 8` after #7015 — 19 of the 19 remaining unverified cells are this single file.

It is **not** a collector defect. The file performs zero collections in *every* configuration:

| configuration | cycles |
|---|---|
| `PERRY_GC_HEAP_LIMIT=8` (matrix `default` arm) | 0 |
| `PERRY_GC_HEAP_LIMIT=8 PERRY_GC_INCREMENTAL=0` | 0 |
| `PERRY_GC_HEAP_LIMIT=2` | 0 |

`PERRY_GC_DIAG=1` prints nothing at all — the arena trigger never arms, because the test allocates under one 1 MB arena block and makes no `gc_malloc` calls. `PERRY_GC_HEAP_LIMIT` cannot lower the trigger further: `gc_trigger_absolute_ceiling_bytes` is `budget/4` with a floor, so 2 MB is already the bottom.

So its 19 `move`/`collect` cells assert nothing about `Ptr<NumArray>` growth under GC, which is exactly the false confidence the matrix exists to remove — the honest `UNVER` is the gate working, and the fix is on the test side.

**Action:** give the file enough escaping allocation churn to cross a trigger under `--pressure 8`, the way `test_gap_repsel_gc_stress` does, while keeping it a growth-path test (it needs to hold a `Ptr<NumArray>` local live across a `push`-driven reallocation while the collector runs). The other Phase 4a.3 members (`p4a3_numarray_barriers`, `p4a3_ptr_numarray`) already collect 1-2 cycles, so this is a churn-budget gap in one file rather than a category problem.

Refs #6978, #7015, #6950.

### LINKED ISSUE #7017: gc-matrix: the collect-arm liveness predicate (cycles > 0) cannot tell a mid-program collection from a teardown one that frees nothing
Split out of #6978 (fixed in #7015). Filed by the author of that fix, because the fix makes the weakness visible rather than removing it.

`scripts/gc_repsel_matrix.sh` scores a `collect`-requirement arm as `PASS` when `cycles > 0`, counting `[gc] cycle` markers on stderr. After #7015 the shipped configuration does complete cycles on the small corpus files, and 171 cells moved `UNVER → PASS`. **On the small files that cycle lands at the event-loop boundary after the program's last output, and it reclaims nothing:**

```
45:8                                    <- last program output
[gc-safepoint] finishing a cycle parked across 2 host safepoints
[gc-step] pre_in_use=479816 post_in_use=483472 sweep_freed=0 block_reclaim=0 pct=0%
[gc] cycle
```

Everything allocated after the cycle armed was born black, so the sweep frees nothing. Compare `PERRY_GC_INCREMENTAL=0` on the same file, which collects **twice, mid-program**, interleaved with the test's output.

Both shapes score identically under `cycles > 0`. So a cell that traverses only the retained graph at rest, at teardown, is indistinguishable from one that collected while the test's representation-selected locals were live — and the latter is the property the matrix was built to assert.

This is not fixable in the collector. With two allocation-trigger opportunities in the whole process and seven resumable cycle phases, a mid-program completion under incremental mode would have to be synchronous at the arming point — which is exactly what `PERRY_GC_INCREMENTAL=0` already is. See #6978 for the per-call measurements.

**Suggested direction (not prescriptive):** make the `collect` predicate demand a *productive* cycle rather than any cycle. Candidates, cheapest first:

1. require `sweep_freed > 0` or `block_reclaim > 0` on at least one cycle (already in the `PERRY_GC_DIAG` `[gc-step]` line the harness captures);
2. require at least one cycle to complete **before** the program's last stdout line, which needs the harness to interleave the two streams for the liveness read while keeping stdout pure for the parity compare;
3. introduce a third verdict between `PASS` and `UNVER` for "collected, but only at teardown" so the distinction stays visible instead of being folded either way.

Whatever the choice, the point of #6942 / #6946 / #6950 stands: a green cell from a collection that could not have observed the test's live data licenses the same false confidence as a green cell from no collection at all.

Refs #6978, #7015, #6950, #6946, #6942.

### LINKED ISSUE #7018: gc: PERRY_GC_TRACE=1 SIGSEGVs under the evacuating arms — the liveness instrument crashes on the corpus file that most needs it

proggeramlug:
Closes #7016. Closes #7017. Closes #7018 (does not reproduce — evidence below).

Three issues, one disease: a check that could not fail. For each one the first job was to decide whether the **probe** or the **predicate** was at fault, and the answer differs.

## #7016 — the PROBE. 19 cells measured a collector that never ran.

Reproduced first, before changing anything:

```
PERRY_GC_HEAP_LIMIT=8 PERRY_GC_TRACE=1 PERRY_GC_DIAG=1 ./numarray_growth
exit 0   cycles=0   stderr lines=0
```

Zero cycles, and `PERRY_GC_DIAG=1` printed **nothing at all** — the file allocates inside one 1 MB arena block and makes no `gc_malloc` calls, so the arena trigger never arms, and `PERRY_GC_HEAP_LIMIT` cannot reach it (`gc_trigger_absolute_ceiling_bytes` is budget/4 with a floor). The `UNVER` verdict was the gate working.

Sections 1–5 are **untouched**, so their "fully contained, therefore promoted" shape is byte-for-byte what it was; a new section 6 adds `test_gap_repsel_gc_stress`'s escaping module-level churn sink, with the numeric-array local initialized *before* the churn, grown by `push` past several capacity doublings *while* it runs, and read *after* it in the same iteration.

Measured on all 13 arms — **PASS on every one**, output byte-identical to the pinned Node 26.5.1 oracle in each:

| arm | requires | cycles | copy-minor objects |
|---|---|--:|--:|
| `default` | scavenge | 4 | **74,404** |
| `safepoint_minor` | scavenge | 3 | 72,148 |
| `evac_minor` / `force_evac` / `force_verify` | move | 4 | 74,404 |
| `gen_gc_off` / `wb_off` / `gen_off_verify` / `all_four` | collect | 7 | 0 (non-moving by construction) |
| `shipped_default` | none | 0 | 0 (control, unchanged) |

Per the #7666 lesson, the check is on the **copying minor**, not merely on a cycle: a probe that allocates a lot can still run zero copying minors when `arena_growth_full_escalation_due` escalates every minor to a full sweep.

## #7017 — the PREDICATE. `cycles > 0` counted a teardown cycle.

A `collect` cell was `PASS` on any cycle. On a small file that cycle lands at the event-loop boundary *after* the program's last output and reclaims nothing, because everything allocated after it armed was born black — indistinguishable from a run that collected mid-program while the test's representation-selected locals were live.

`collect` now requires a **productive** cycle. Following #7657's widening of the gc-ratchet rule to `copied + promoted > 0`, the counter names a **destination** rather than a single number: `reclaimed` sums `sweep_freed`, `block_reclaim`, `eden_dead_bytes`, `freed_bytes` and `dead_bytes`, in **both** the `k=N` and JSON `"k": N` spellings.

**Reading only `k=N` was itself a vacuous check, and it was caught by measurement.** `test_gap_gc_symbol_local_rooting` runs 86 malloc-count-triggered cycles that free **31.9 MB** of symbols; every arena-side counter reads 0 because the malloc sweep's bytes appear only in the JSON trace's `sweep.freed_bytes`. A predicate over the `k=N` shape alone would have scored it inert.

The script says out loud that this is a **conservative proxy**: a mid-program cycle over a heap that is entirely live reclaims nothing and reads `UNVER`. It can under-claim, never over-claim, which is the safe direction for a liveness gate; `cycles` over-claimed.

**Both halves move together.** `reclaimed` is threaded into the per-cell JSON and `gc_matrix_liveness_check.py`'s `REQUIREMENTS["collect"]` reads it. **Shown able to fail:** reverting the counter to `("cycles",)` fails exactly the two new self-test cases that pin the change —

```
SELF-TEST FAIL: a cycle that reclaimed nothing does NOT satisfy collect: wanted a violation, got []
SELF-TEST FAIL: collect is NOT satisfied by scavenged/evacuated alone: wanted a violation, got []
```

— and restoring it returns `24 checks, 0 failures`. A third case pins that a productive cycle still passes.

The per-run liveness table now prints `reclaimed` beside `collected`, so the gap stays on screen instead of being folded away:

```
gen_gc_off   requires=collect  collected  3/ 3   reclaimed  1/ 3   moved-objects  0/ 3   copy-minor  0/ 3
```

**Corpus-wide impact**, measured under `gen_gc_off` over all 58 files: 12 never collect (already `UNVER`), **14 collect but reclaim nothing** — `PASS` before, honestly `UNVER` now — and 32 are productive, which keeps every `collect` arm live and the arm-level gate green.

## The four known-inert arms are live, and CI has been saying so

Not in the issues, found while validating, and **this PR cannot be green without it**.

`test-parity/gc_matrix_inert_arms.txt` registered `default`, `verify_evac`, `cons_scan_off` and `cons_scan_off_force` as inert because `PERRY_GC_MOVING_LOOP_POLLS` is default-off, making the copying minor "ineligible by construction". **The poll flag has not changed and all four scavenge anyway.**

* **CI, `gc-stress` on `main`, run 31240304595 (2026-08-08):** each satisfied `requires=scavenge` on **41 of 58 cells**, `default` at `counter=1384046`, and the job was red with four `STALE-REGISTRY` lines. This is not a condition this PR introduces — it is one `main` has been red for.
* **Locally, isolated:** `--arms default,verify_evac,cons_scan_off,cons_scan_off_force --filter test_gap_repsel_gc_stress` — a file this PR does not touch, under the `scavenge` predicate this PR does not change — 1/1 live on every arm, `default` copying 228,181 objects.
* `shipped_default` (requires=none, no pressure knob, no GC env at all) copies **340,956** objects on the same file. The shipped configuration relocates today.

All four entries are deleted with the measurement recorded in their place, and the four arm notes in `gc_repsel_matrix.sh` that repeated the claim in prose are corrected — otherwise they become exactly the stale prose that file's own header warns about. What changed is the collector around the flag (#7370's statepoint default, #7432, #7657, #7666), not the flag. The registry is now empty; if the copying minor ever stops running in the shipped configuration, the gate fails in the *other* direction on the same line.

## #7018 — does not reproduce, and the hypothesis is structurally refuted

`PERRY_GC_TRACE=1` was reported to SIGSEGV `test_gap_repsel_scalar_replaced_locals` under the evacuating arms. **20 runs across both link modes: 0 crashes**, stdout byte-identical with and without the flag, on that file and four others.

The first link mode was **vacuous and is reported as such**: auto-optimize relinks the runtime `--no-default-features`, so `diagnostics` is off and `GcCycleTrace::emit` falls to its stub — all 113 `[gc] cycle` lines read "diagnostics feature disabled", and the real tracer never ran. Re-run against the diagnostics archive: **110 real `"event":"gc_cycle"` objects, 466,424 objects copied**. The arm is demonstrably live and still does not crash. 13 `.ips` crash reports were produced during the investigation and every one is `SIGABRT` from an unrelated instrument; **zero SIGSEGV**.

Structurally it cannot hold. `PERRY_GC_TRACE` is read in exactly one place (`gc_trace_enabled()`, `gc/policy.rs:571`) with two call sites: a scalar counter snapshot (`policy.rs:760`) and a thread-local `u64` bump (`barrier.rs:931`). Emission serialises already-accumulated scalars. **Nothing on that path dereferences a heap object, walks the object graph, or reads a `GcHeader`** — so it cannot dereference a forwarded pointer, and it shares nothing with #6998/#6995. The issue's description ("the tracer runs inside a collection, walking structures the collector is mid-way through mutating") describes `gc/trace.rs`, the **marking** tracer, which runs on every collection regardless of the flag. A name collision, not a defect.

## Validation

* Full `--arms pr` matrix over all 58 corpus files, exit 0 on the liveness gate.
* `--arms pr --filter test_gap_repsel_p4a3`: 21/21 cells byte-exact, `PASS=21 UNVER=0 FAIL=0`, all seven arms `live`.
* `--arms gen_gc_off,wb_off,evac_minor,shipped_default --filter test_gap_repsel_p4a3`: the two probes that only reach a teardown cycle now read `UNVER` on the `collect` arms and `PASS` on `evac_minor`, which is the intended split.
* All **22** `lint` gate commands green; `gc_matrix_liveness_check.py --self-test` 24 checks / 0 failures; `bash -n` clean.

## One unrelated red seen while validating, reported rather than swallowed

The 58-file `--arms pr` run above ended `PASS=341 UNVER=64 XFAIL=0 **FAIL=1**`: `test_gap_repsel_gc_stress` × `safepoint_minor`, output-mismatch, `cycles=52 scavenged=38184`. It is **not** from this PR — different file, and `safepoint_minor`'s `requires=scavenge` predicate is untouched here.

Re-run in isolation, same binary, same env: **10/10 clean**, so it is intermittent, not deterministic. `safepoint_minor` is the arm that compiles *and* runs with `PERRY_GC_MOVING_LOOP_POLLS=1` — the configuration #7161 turned off by default because of #7154's use-after-free — so an intermittent mismatch there is consistent with that open class. It is untriaged in `test-parity/gc_repsel_triage.txt` and CI's last `main` run reported `FAIL=0`, so it is a low-rate sampling difference rather than a new deterministic break. Filed separately; deliberately **not** added to the triage list here, because triaging a red I have not root-caused is how a real defect becomes permanent furniture.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated garbage-collection liveness checks to require measurable memory reclamation, not just a completed cycle.
  * Improved reclamation reporting across diagnostic formats, with clearer per-test results and JSON output.
  * Removed obsolete inert-arm entries after confirming those configurations now perform collection.

* **Tests**
  * Added stress coverage for numeric-array growth during repeated allocation and collection.
  * Expanded self-tests to validate productive collection and reject non-reclaiming cycles.


proggeramlug:
## Audit — merging as v0.5.1388

**#7016 verified, after I measured it wrong first.** My bare invocation reported 0 cycles and I nearly filed that against you. The matrix runs its arms under `PERRY_GC_HEAP_LIMIT=$PRESSURE_MB` (default 8), which I had omitted. With it:

| pressure | gc_cycle markers |
|---|--:|
| 8 MB (the arm default) | **4** |
| 16 MB | 3 |
| 32 MB | 1 |

Control: my build emits trace fine (a churn kernel shows 2 without pressure, 10 with). So the probe genuinely crosses collections now, and my first reading measured a configuration the gate never runs — the same shape as everything else in this cluster, applied to my own check.

**#7017 is the right fix for the right reason.** `cycles` is not evidence the collector saw anything; a cycle that reclaims nothing scores identically to one that reclaims 31.9 MB. Summing five reclamation counters in **both** `k=N` and JSON `"k": N` spellings is what makes it real — reading only `k=N` scored `symbol_local_rooting`'s 86 malloc-triggered cycles as inert.

And the design note in the source is the part I most want kept: *"one collector change cannot pin it permanently false."* That is #7657's lesson — re-enabling the adaptive-tenuring seed drove `copied_objects` to 0 on two probes and would have pinned the old guard permanently false exactly where it last fired. Summing a family rather than naming one counter is the general fix.

14 of 58 files moving PASS → honest UNVER is a **gain**, not a regression: those cells were reporting on a collector that never ran.

**#7018 closed by refutation**, and that is worth as much as a fix: 20 runs without reproduction, plus the structural argument that `PERRY_GC_TRACE` never dereferences a heap object — the issue describes `gc/trace.rs`, the *marking* tracer. A refuted issue with evidence beats one that lingers.

**The unasked-for part is the most valuable.** All four entries in `gc_matrix_inert_arms.txt` were live, and CI's `gc-stress` on `main` has been red for it at **41/58 cells with four `STALE-REGISTRY` lines**. A registry that outlives its cause is the same defect as a knob whose off-state nobody exercises — and the liveness gate failing the day an arm scavenges again is what keeps the entries from outliving their causes a second time.

**Gates:** 24/24 lint, matrix self-test and `--check-registry` clean, `check_test_registration` clean, script syntax clean. The registry now reports six relocating arms live: `default, safepoint_minor, evac_minor, verify_evac, force_verify, cons_scan_off`.

Worth watching when the queue drains: `gc-stress` on this PR is the confirmation that retiring those four entries was right — it has been red on `main` for exactly that reason.

PR #7679, opened 2026-08-09
gc: teach gc-root-dominance the referent-with-no-name hole (#7616)

### LINKED ISSUE #7616: gc-root-dominance is blind to the unrooted-register shape: the verbatim pre-#7453 code reports 0 violations in all three modes

proggeramlug:
Closes #7616.

## The measurement #7616 asked for

#7616's finding was that `gc_root_dominance_check.py` reports the verbatim
pre-#7453 code as clean in all three of its original modes. #7663 has since
added a fourth, `--statepoints`, which reads the **native** lowering — the one
that ships. The right close was a measurement, not an argument: plant the exact
reproducer, run `--statepoints --moving-only`, record whether it is reported.

**It is not.** The new mode is as blind as the three it was added to cover for.

The reproducer was re-planted verbatim in `crates/perry-codegen/src/expr/url_main.rs`
and the emitted IR confirmed register-for-register identical to what the issue
quotes:

```llvm
%r883 = call i64 @js_url_coerce_string(double %r882)
%r884 = load double, ptr @test_compat_url_date_math_ts_.str.17.handle
%r885 = call i64 @js_url_coerce_string(double %r884)      ; allocates
%r886 = call i64 @js_url_new_with_base(i64 %r883, i64 %r885)  ; %r883 is stale
```

Over the 16 `test-files/` sources that exercise the URL lowerings, both
lowerings, both corpora generated by the production pipeline (`--trace llvm`
plus `STATEPOINT_REWRITE_PASSES` for the native arm):

| mode | clean | sabotaged |
|---|--:|--:|
| `--moving-only` (dominance) | 0 | **0** |
| `--unrooted-allocas --moving-only` | 0 | **0** |
| `--stale-registers --moving-only` | 2 | **2** |
| `--statepoints --moving-only` | 2 | **2** |
| `--stale-registers` (unfiltered) | 24 | 35 |
| `--statepoints` (unfiltered) | 15 | 21 |

Every **gated** arm blind; both **unfiltered** arms see it. The cause is the one
#7616 named: `js_url_coerce_string` is in `ALLOC_RE` but not in
`POLL_CAPABLE_RUNTIME`, and `--moving-only` — which every gated arm runs, in all
four modes — drops any window it cannot classify as MOVING. Adding that one name
takes the sabotaged arms to **13** and **8** and leaves the clean arms at 2 and 2.

## The generalisation

`--audit-alloc-re` catches a dead regex alternative. `--audit-poll-capable`
catches an entry naming no symbol. Both hunt a **name with no referent**.
Nothing hunted a **referent with no name**, and a missing entry suppresses
findings exactly as silently as a phantom one — #7453's fix comment says in as
many words *"that gap is why the checker did not flag #7453"*, then stopped one
list short.

`--audit-poll-reach` is that third auditor. It deliberately does **not** assert
"every poll-capable runtime symbol must be listed" — 297 exported symbols call
one directly, and deciding that is a coverage change with its own hit count
(the same reasoning `ALLOC_RE`'s deleted `bigint_\w+_op` records). It asserts
only that **the checker's two lists must not disagree about the same symbol**:
if `ALLOC_RE` says a call's result is a heap value the checker must track, and
the runtime shows that same call invoking something `POLL_CAPABLE_RUNTIME`
already grants can re-enter JS, the premise for listing it is the premise the
set already granted its callee. There is no judgement left to make.

77 symbols, `js_url_coerce_string` among them. `POLL_CAPABLE_RUNTIME` goes
54 → 131.

## Shown able to fail

Three plants, each verified to turn the gate red, and the clean run green:

| sabotage | expected | got |
|---|---|---|
| delete `js_url_coerce_string` from `POLL_CAPABLE_RUNTIME` | `--audit-poll-reach` exits 2 and names it | rc=2, named |
| neuter `_strip_noncode` | `--self-test` reports the decoy fixture | rc=1, decoy reported |
| replace the reach fixpoint with one hop | `--self-test` reports the transitive fixture | rc=1, transitive missing |

The third is not hypothetical. The first version of this audit **was** one-hop:
it reported 52 names, they were added, and re-running found 10 more that reached
through them. An audit that has to be run in a loop until it stops finding
things reports an arbitrary prefix of its own answer, so the reach relation is a
fixpoint and the self-test pins it.

The decoy fixture's only mention of a poll-capable call is inside a `//` comment
and a string literal — and it carries the full `js_string_coerce(v)` spelling,
parentheses included, because a decoy that merely *names* the symbol is rejected
by the call regex rather than by the stripper and would pass with
`_strip_noncode` deleted. (Measured: the first draft did.)

Both non-vacuity floors are asserted rather than hoped for: an empty symbol scan
**and** a 600-symbol scan whose bodies are all empty must each be an error, not
a clean verdict.

## Budgets — both corpora, which is the reason #7616 was a separate issue

| corpus | arm | before | after | pinned |
|---|---|--:|--:|--:|
| curated (129 sources / 149 modules) | `--moving-only` | 0 | 0 | allowlist, empty |
| curated | `--unrooted-allocas --moving-only` | 0 | 0 | 0 |
| curated | `--stale-registers --moving-only` | 9 | 13 | 39, unchanged |
| native (149 modules, 30632 safepoints) | `--statepoints --moving-only` | 7 | **11** | 7 → **11** |
| dependency-scale (81 zod modules, 67 MB) | `--moving-only` | 0 | 0 | allowlist, empty |
| dependency-scale | `--unrooted-allocas --moving-only` | 0 | 0 | 0 |
| dependency-scale | `--stale-registers --moving-only` | 86 | 104 | 118, unchanged |

Only `--max-unrooted` moves. The four new hits are named one by one in the
workflow rather than absorbed into a number:

* **3** in `test_gap_array_splice_spread::main`, `unrooted:alloc` — a fresh
  array held in a raw i64 across `js_array_like_to_array` and consumed by
  `js_array_concat`. #7453's shape, in the spread lowering instead of the URL
  one, i.e. exactly the population this change exists to make visible.
* **1** in `test_gap_class_expr_dynamic_parent_ctor`, `unrooted:capture` — the
  same `js_closure_get_capture_bits` residual the budget comment already names,
  newly visible because `js_new_function_construct` is now classified as a mover.

No hit disappeared. The 86 → 104 on the dependency corpus stays inside the
pinned 118 and the budget is deliberately **not** raised.

## Validation

* `--self-test`, `--audit-alloc-re`, `--audit-poll-capable`,
  `--audit-poll-reach`, `--audit-immovable-sources`: all green.
* All 22 `lint`-job commands, each with its own exit status checked: 0 failures.
  (`cargo fmt --all -- --check` included; no Rust changed.)
* `scripts/gc_gate_wiring_check.py`: 6 gates main-line-reachable and able to fail.
* Corpora generated with a compiler built from this branch and verified
  unsabotaged before the clean measurements were taken.

## Summary by CodeRabbit

* **New Features**
  * Added a poll-reach audit that detects allocation-related runtime symbols capable of reaching poll operations.
  * Expanded poll-capable symbol coverage, including direct and transitive call paths.
  * Added validation for non-vacuous scans and comprehensive reachability checks.

* **CI**
  * Integrated the new audit into garbage-collection root-dominance checks.
  * Updated native statepoint analysis thresholds based on expanded coverage.

* **Documentation**
  * Documented the new audit, failure mode, coverage, and verification results.


proggeramlug:
## Audit — merging as v0.5.1391

**The finding is the deliverable, and it is a hard one to have gone looking for:** the `--statepoints` mode #7663 added specifically to cover the other three's blind spot **has the same blind spot**, for the same one-line reason. You measured it rather than arguing it — re-planting #7616's exact `Expr::UrlNew` reproducer and getting IR register-for-register identical to the issue, then showing `--statepoints --moving-only` reports `2` on both the clean and sabotaged arms.

That table is the whole case: every **gated** arm is blind, both **unfiltered** arms are not. A gate that is blind precisely in the configuration it runs in is worse than no gate, because it is quoted as evidence.

### `--audit-poll-reach` is the right instrument

The two existing auditors hunt a **name with no referent**; this hunts a **referent with no name**. That is the inverse direction and nothing was covering it. Scoping it to "the checker's two lists must not disagree about the same symbol" rather than "enumerate every poll-capable symbol" (297 call one directly) is what keeps it a gate rather than a wish.

**Sabotage verified independently.** I removed `js_url_coerce_string` from `POLL_CAPABLE_RUNTIME`: exit 2, naming the symbol, with a message that states the *consequence* rather than the symptom —

> a window whose only collection point is one of these classifies `MOVING: no`, so every `--moving-only` arm — which is every gated arm, in all four modes — drops it. That is #7616 exactly.

Restored: exit 0, `3784 exported symbols, 1610 with an intra-runtime call edge, 384 matched by ALLOC_RE`. Non-vacuity floors asserted on the real corpus, which is the part that makes "no unlisted symbol" mean something.

**The fixpoint detail matters and I'd have missed it.** The first version was one-hop, reported 52, and found 10 more when re-run after those were added — so the relation genuinely needs a fixpoint, and pinning that with a transitive fixture rather than discovering it again later is the difference between a gate and a snapshot.

### The budget going UP is correct

`--max-unrooted` 7 → **11** reads like a regression and is the opposite: the gate got sharper, so windows it previously classified `MOVING: no` and dropped are now visible. Four new hits, each named — 3 in `test_gap_array_splice_spread` (**#7453's shape in the *spread* lowering**, i.e. the same defect the URL fix was written for, one list short) and 1 `js_closure_get_capture_bits`.

Naming them individually rather than absorbing them into a number is what keeps the budget a ratchet. Same for the stale budgets: curated 9→13, dep 86→104 with the pin deliberately not raised.

**Gates:** 24/24 lint, fmt clean, `--self-test` and all three auditors exit 0, `gc_gate_wiring_check` reports **6 gates main-line-reachable and able to fail**.

#7616 stays open until its reproducer is re-run against the *repaired* mode — which is now a one-command check rather than an argument, and is the right closing condition for it.

Issue #7682, opened 2026-08-09
P0: silent wrong answer under the moving collector — stale heap-string deref in generated code (realistic interpreter workload)


proggeramlug:
## Summary

A realistic, 100%-static TypeScript program — a tree-walking interpreter for a small
functional language — **produces a silently wrong answer on default settings**. No
crash, no diagnostic, no `TypeError`: just an incorrect number.

```
node --experimental-strip-types interp.ts   → 1708840
scriptc 0.0.22 (fully static)               → 1708840
perry  0.5.1384 (a853135aa), defaults       → 1708662     ← WRONG, every run
```

Reproducible on 6/6 consecutive runs. The program is in
`gc-handoff/apps/interp.ts` (189 statements; `scriptc coverage` reports
"fully static — this program has no dynamic remainder", so it is ordinary
TypeScript, not an exotic construct).

**This is a moving-collector correctness bug.** Four independent knobs each restore
the correct answer:

| knob | result |
|---|---|
| default | **1708662 (wrong)** |
| `PERRY_GEN_GC=0` | 1708840 ✓ |
| `PERRY_GC_SCAVENGE=0` | 1708840 ✓ |
| `PERRY_CONSERVATIVE_STACK_SCAN=1` | 1708840 ✓ |
| `PERRY_WRITE_BARRIERS=0` | 1708840 ✓ |

All four share one property: they prevent evacuation/moving.

## User-visible mechanism

The interpreter's variable lookup silently fails to find bindings that are present.
An instrumented `lookup` (counting the fall-through-to-default path) gives:

```
node                : checksum 437840  misses 0
perry (default)     : checksum 437836  misses 6      ← 6 silent lookup failures
perry PERRY_GEN_GC=0: checksum 437840  misses 0
```

Because naive `fib` is just a count of leaves returning 1, each miss subtracts
exactly 1 from the answer. Miss count varies with allocation timing (2–10 observed);
the wrongness does not.

The failing comparison is `names[i] === name` where both sides are heap strings
created by `src.substring(i, j)` in the lexer. A second symptom in the same run: a
concatenated diagnostic string came back empty.

## Precise fault (the instrument did its job)

`PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800 PERRY_GC_ZEAL=1`:

```
[gc-fromspace-protect] FAULT: signal 10 at 0x2fa3bcbcc94
[...]

Issue #7512, opened 2026-08-06
perf: `new Klass(v,w)` is 63% slower than the equivalent object literal (28.5x vs 17.4x Node)


proggeramlug:
## Summary

Constructing a class instance is **63% slower** than constructing the equivalent object
literal, on an otherwise identical workload. `new Node(v, w)` with two declared `number`
fields is **28.5× Node**; the `{v, w}` literal is 17.4×.

That is backwards. A fixed-shape class with declared primitive fields is the most
statically-known construction form Perry has — known field count, known types, known
layout, a real constructor to attach a shape to. It should be the *fastest* path, not the
slowest.

This is filed separately from the two profile-driven sibling tickets because it looks
bug-shaped rather than optimisation-shaped, and may be a cheap bisect rather than a
project.

## Evidence

Quiet M1 mini (`perry-macos.local`, load ~1.4), best-of-3 wall clock. Both programs do the
same thing — 20,000 rounds × 1000 objects pushed into a fresh local array:

| variant | Perry | node | scriptc | ratio vs node |
|---|--:|--:|--:|--:|
| `churn_alloc.ts` — `keep.push({ v: base + j, w: j })` | 2.44 s | 0.14 s | — | 17.4× |
| `push_cls.ts` — `keep.push(new Node(base + j, j))` | **3.99 s** | 0.14 s | 0.46 s | **28.5×** |
| `push_num.ts` — `keep.push(base + j)` (no object) | 0.30 s | 0.11 s | — | 2.7× |

Node treats the two forms as equivalent (0.14 s both). scriptc does `push_cls` in 0.46 s —
**8.7× faster than Perry** on the class form, while Perry beats scriptc comfortably on
pure compute (fib(40): 0.39 s vs 0.78 s). So this is specific to class construction.

`push_cls.ts`:

```ts
class Node {
  v: number;
  w: number;
  constructor(v: number, w: number) { this.v = v; this.w = w; }
}
function chunk(base: number): number {
  const keep: Node[] = [];
  for (let j = 0; j < 1000; j++) keep.push(new Node(base + j, j));
  return keep.length;
}
function main(): void {
  let total = 0;
  for (let c = 0; c < 20000; c++) total += chunk(c * 1000);
  console.log(total);
}
main();
```

Programs are in `gc-handoff/bench/` (`push_cls.ts`, `churn_alloc.ts`, `push_num.ts`).

## Where to look

The symbolicated profile of the *literal* form already shows a large typed-feedback
[...]

Issue #7510, opened 2026-08-06
perf(gc): layout side tables are 34% of object construction — the construction/death half of #5094 (allocation is 7.7%)


proggeramlug:
## Summary

**Granular member of the #5094 family.** #5094's table covers the *access* paths
(`this.field` get/set, array element writes, dynamic property writes). This ticket is
the **construction and death** path, which that table does not mention and which is now
the single largest cost in allocation-heavy code.

On `gc-handoff/bench/churn_alloc.ts` — 20M `{v: number, w: number}` literals pushed into
a 1000-element array — a symbolicated profile attributes **33.6% of self time to
`gc::layout` per-object side-table maintenance**, against **7.7% for the actual
allocation**. `layout_forget_object` alone is 14.5%, nearly twice the allocator.

#6893 closed the *memory* half of this subsystem (per-class layout interning took the
record footprint 289 B → 133 B). The *CPU* half was never touched.

## Evidence

Symbolicated profile (`PERRY_DEBUG_SYMBOLS=1`, `sample` 2 s, 1500 leaf samples, quiet M1
mini at load 1.4):

| group | share |
|---|--:|
| **`gc::layout` side tables** | **33.6%** |
| — `layout_forget_object` (`gc/layout.rs:336`) | 14.5% |
| — `layout_note_slot` (`gc/layout.rs:685`) | 7.9% |
| — `js_gc_init_typed_shape_layout` (`gc/layout.rs:962`) | 7.7% |
| — `shape_install_*` | 2.2% |
| — `LayoutSlot*` | 1.3% |
| `_tlv_get_addr` | 17.0% |
| write barriers (see sibling ticket) | 16.1% |
| typed-feedback guards | 9.2% |
| array (`js_array_length`, `js_array_note_numeric_write`) | 6.2% |
| **the allocation itself** (`js_object_alloc_class_inline_keys` 6.3%, `arena_alloc*` 1.4%) | **7.7%** |
| user code | 3.7% |

Note the interaction with the TLS row: `TYPED_LAYOUTS` / `LAYOUT_SLOT_MASKS` are
thread-locals, so an unknown but material share of the 17% `_tlv_get_addr` is *also* this
subsystem. #7474 already cached the hot TLS addresses and TLS still costs 17%; removing
the layout ops removes their share of it too. One fix, two rows.

`js_gc_init_typed_shape_layout` is emitted by `lower_call/new.rs` on **every**
[...]

Issue #7288, opened 2026-08-03
build: byte-identical source compiles to a 44x-slower object depending on where the .ts file lives; the published baseline measures the fast arm


proggeramlug:
## Summary

The same `.ts` file, byte-identical, compiled by the same `perry` binary, linking the same runtime archive, produces **two different object files** — and one of them runs `09_method_calls` **44× slower** than the other. The discriminator is *where the source file lives on disk*.

```
source at <perry-checkout>/**/09_method_calls.ts   → object 6efab0f508029e70.o →   78–110 ms
source anywhere else                               → object b0047dfba1cb3e98.o → 3417–3503 ms
```

`benchmarks/results/public-node-bun-v1.json` reports 79 ms for `09_method_calls`, because `benchmarks/compare.sh` does `cd benchmarks/suite` and compiles there. **A user compiling the identical file in their own project gets 3450 ms.** The published baseline is measuring an arm that only exists inside the Perry checkout.

## Reproduction (deterministic, 12/12 and 6/6)

```bash
P=target/release/perry
SRC=benchmarks/suite/09_method_calls.ts

# arm A — inside the checkout, any subdirectory, any filename
mkdir -p ./zzp && cp $SRC ./zzp/ && (cd ./zzp && PERRY_CACHE_DIR=/tmp/c1 ../../$P 09_method_calls.ts -o /tmp/A)
/tmp/A            # method_calls:104

# arm B — outside the checkout, same file
mkdir -p /tmp/zzp && cp $SRC /tmp/zzp/ && (cd /tmp/zzp && PERRY_CACHE_DIR=/tmp/c2 $PWD/$P 09_method_calls.ts -o /tmp/B)
/tmp/B            # method_calls:3435
```

Controls, all with a **fresh** `PERRY_CACHE_DIR` each time (so no stale cached object is involved):

| condition | result |
|---|---|
| `<checkout>/benchmarks/suite/` | 108 ms |
| `<checkout>/zzp11/deep/er/` (arbitrary subdir) | 104 ms |
| `<checkout>/zzp3/renamed_bench.ts` (renamed) | 110 ms |
| `/private/tmp/zzprobe/` | 3426 ms |
| `/private/tmp/a/b/c/d/e/` | 3436 ms |
| `/private/tmp/perryfake/benchmarks/suite/` (same dir names) | 3501 ms |
| `/private/tmp/perryfake2/perry/benchmarks/suite/` | 3503 ms |
[...]

PR #7686, opened 2026-08-09
perf(class-fields): typed slot stores for pointer-typed class fields (#5094)

proggeramlug:
## P1 (#5094): one pointer field no longer demotes an object's whole store set

A single pointer-typed class field (`peer: Cell | null`, `next: LNode | null`,
`left: Tree | null`) put **every** field store on that object — including its
`number` fields — on `js_object_set_field_by_name`: by-name dispatch, a
`RuntimeHandleScope`, `layout_note_slot`, and a per-object side-table entry, for
stores whose slot index is a compile-time constant.

### Measured — quiet M1 mini, best-of-3 wall, both arms run back-to-back

| bench | before | after | scriptc 0.0.22 | node 26.5.1 |
|---|--:|--:|--:|--:|
| `cycles` | 0.79 | **0.24** | 0.31 | 0.07 |
| `deeplist` | 1.14 | **0.33** | 0.22 | 0.09 |
| `tree_wide` | 12.18 | **3.02** | 7.01 | 0.89 |
| `tree` | 5.92 | **4.43** | 4.80 | 0.45 |

`cycles`, `tree` and `tree_wide` now beat scriptc; `tree_wide` is 4.0×.
Unchanged, as required: `push_cls` 0.35, `churn` 0.66, `churn_alloc` 0.36,
`push_num` 0.13, `retain` 1.32, `retain1` 0.42, `churn_read` 0.35. Every
benchmark's stdout is byte-identical to the pre-change binary.

### Three changes, and they must land together

1. **The sloppy-mode class-field route reaches boxed slots** (`expr/property_set.rs`).
   #7288 opened it for raw-f64 slots only, so `n.next = head` fell through to the
   `PutValue` write IC whose miss is `js_put_value_set` → by-name. The
   guard-free store is licensed by the same #5093 inline precheck the raw arm
   uses (it rejects frozen and descriptor-bearing receivers — the only thing
   sloppy and strict `PutValue` disagree about) and the miss stays
   `js_put_value_set(..., strict = 0)`.

2. **A pointer-bearing class declares its layout at allocation** (`typed_shape.rs`).
   #7510 required an EMPTY pointer mask, which excluded exactly these classes —
   so their descriptor arrived *after* every store in their constructor and none
   could pass its intact-bit guard (#7512's defect, still open for this shape).
   Obligation 2 is now discharged rather than avoided: both `new` allocation
   paths pre-fill every slot with `TAG_UNDEFINED`, which the tracer rejects at
   its tag check, so a pointer-masked slot visited before its first write cannot
   strand anything. Obligation 1 (no read may observe a raw-f64 slot before its
   first write) is unchanged and still required of every `number` field.

3. **The constructor prologue admits literals and pure operator trees**
   (`lower_call/field_init.rs`). `constructor(v) { this.next = null; this.v = v }`
   truncated the prologue at statement 0 and came back EMPTY, which cost both the
   dead-`undefined`-store elision and (via the same set) the declaration in (2).
   `tree_wide`'s `this.b = s + 1` needs the operator-tree half.

**Why together:** (1) alone REGRESSES `tree_wide`. Measured directly by
compiling the benchmarks as ESM — which already takes the class-field route —
against the pre-change compiler: `tree_wide` 12.21 → **14.88** s. Routing a
constructor's stores to a guard the construction path has made unsatisfiable is
slower than the inline cache it displaces. (2) is what makes the guard passable.

### GC

The dangerous half. A pointer store must still reach the remembered set.

- `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1` (depth 800), binaries compiled
  with `PERRY_GC_MOVING_LOOP_POLLS=1`, over pointer-cycle / linked-list /
  wide-tree probes: clean, and **not vacuous** — 20005 / 40005 / 5
  `[gc-fromspace-protect]` retirements, i.e. the copying minor really ran. All
  three match the `node 26.5.1` oracle, including a walk that sums every numeric
  slot as well as every pointer edge.
- `PERRY_GC_VERIFY_EVACUATION`, `PERRY_GC_VERIFY_MARK` and
  `PERRY_GC_FROMSPACE_SCAN_ABORT`, each under zeal: clean.
- `PERRY_GC_TRACE` drift: `churn` is bit-identical (105 cycles, 0.0039 GB
  copied, positive reclamation every cycle). `deeplist`/`tree`/`tree_wide` keep
  their cycle counts and kinds exactly (4 / 42 / 44); `cycles` copies 10,758 → 14
  objects and promotes 4,746 → 2, which is the side-table bookkeeping this PR
  removes no longer keeping dead objects reachable.
- `gc-root-dominance`, both arms, exact CI invocation: **0 violations** with the
  40-seeded-violation control catching 40/40, and `--unrooted-allocas` 0.

### One latent gate bug this exposed

`scripts/gc_root_dominance_check.py`'s `NONCOLLECTING` set is a second copy of a
fact `perry-codegen/src/gc_call_effects.rs` already states, and the two had
drifted: #7510 added `js_gc_declare_typed_shape_layout` beside
`js_gc_init_typed_shape_layout` in the Rust match and not in the Python set.
That stayed invisible only because the corpus then contained no class the #7510
gate admitted. Widening the gate printed **358 spurious violations**, every one
`js_object_alloc_class_inline_keys->js_gc_declare_typed_shape_layout`. The two
entry points share a body and differ only in a `TypedShapeProof` that makes
`declare` do strictly less work, so the classification is the same one.

### Pre-existing, not from this PR

`typed_shape_descriptors::integer_arithmetic_array_push_omits_inbounds_layout_note_and_barrier`
fails identically at `6cdcd79ec` with this PR's sources reverted (verified by
checkout, not by inspection). It is an array-push contract with no class-field
involvement.

### ⚠️ This PR MASKS the `interp.ts` GC canary — do not use it as the probe

`gc-handoff/BUG-interp-silent-wrong-answer.md` reports a moving-collector bug
(silently wrong answer, `1708662` vs node's `1708840`, 6/6) and proposes adding
`interp.ts` to the GC-ratchet corpus as a correctness probe. **Under this PR
that probe passes while the bug is fully intact.**

Measured by building two compilers from the same base commit, one with this
diff and one with the codegen sources reverted:

| | `interp.ts` ×6 | `iso_miss.ts` ×6 |
|---|---|---|
| node 26.5.1 | 1708840 | misses 0 |
| base | 1708662 (wrong) ×6 | misses 6, 4, 2, 4, 8, 6 |
| this PR | **1708840 (correct) ×6** | **misses 4, 6, 4, 4, 4, 4 — still wrong** |

The bug is present in both arms. This PR perturbs allocation/promotion timing
enough that `interp.ts` lands on the right answer; `iso_miss.ts`, the tighter
isolation, stays red in both. That matches the bug report's own observation that
the miss count varies with allocation timing while the wrongness does not.

**Use `iso_miss.ts` with `misses == 0` asserted** (or its checksum against
node's `437840`). `interp.ts` would become a probe that cannot fail — CLAUDE.md's
"four ways a gate can be unable to fail", item 4.

Nothing here is caused or fixed by this PR; it is flagged because landing it
silently changes what that canary means. Possibly the same root cause as #7682
(alloc-point minor moving past the immobility guard), which would also explain
why `PERRY_GC_SCAVENGE=0` is one of the four knobs that restores the answer.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1389
+**Current Version:** 0.5.1390
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1389"
+version = "0.5.1390"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "perry-ffi",
  "scraper",
@@ -5813,7 +5813,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-commander"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5821,7 +5821,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cron"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "chrono",
  "cron",
@@ -5831,7 +5831,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dayjs"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "chrono",
  "perry-ffi",
@@ -5839,7 +5839,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-decimal"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "perry-ffi",
  "rust_decimal",
@@ -5847,7 +5847,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dotenv"
-version = "0.5.1389"
+version = "0.5.1390"
 dependencies = [
  "perry-ffi",
  "serde_json",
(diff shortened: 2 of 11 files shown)


proggeramlug:
## Blocking: a pointer store on the new path does not reach the remembered set

The perf work looks right and the writeup is careful — but the obligation this PR names as "the dangerous half" is not met. **A young pointer stored into an old object never enters the remembered set on this branch.** `main` is clean on the identical probe.

### Evidence

Both arms built `--profile perry-dev` (`-p perry -p perry-runtime-static -p perry-stdlib-static`, `PERRY_RUNTIME_DIR` pinned), probe compiled with `PERRY_GC_MOVING_LOOP_POLLS=1 PERRY_NO_AUTO_OPTIMIZE=1`, run under `PERRY_GC_ZEAL=1 PERRY_GC_FROMSPACE_SCAN_ABORT=1`:

| arm | 6 consecutive runs |
|---|---|
| `main` (`6cdcd79ec`+) | `0 0 0 0 0 0` — every scan `dangling=0` |
| this branch | **`134 134 134 134 134 134`** |

```
[gc-fromspace-scan abort] objects=1281 words=25309 fwd_owners_skipped=0
    missing_rewrites=0 dangling=1 owners=1 | never_dirty=1 lost_dirty=0 dirty_but_missed=0
[gc-fromspace-scan abort]   in_snapshot=0 not_in_snapshot=1
  owner=0x5e4830f3c60 type=3 space=Survivor0 +40 nanbox -> 0x5e47dea656d (type=0 NurseryEden)
    DANGLING (target not evacuated)
    [slot dirty_now=false ever_dirty=false owner_flags=0x43 marked=true]
```

**`ever_dirty=false` on a Survivor0 → NurseryEden edge is the whole finding.** Not "the dirty bit was cleared early" (`lost_dirty=0`) and not "the scan missed it" (`dirty_but_missed=0`) — the slot was **never** marked dirty, so the remembered set never learned the edge exists and the minor did not evacuate the target.

### Why the PR's own GC section didn't catch it

Those probes are pointer-cycle / linked-list / wide-tree shapes, which build their edges **young-to-young**. The barrier obligation only bites when the *parent has already been promoted* and then takes a pointer to a fresh object. My probe forces that ordering explicitly:

```ts
class LNode { v: number; next: LNode | null; constructor(v: number) { this.v = v; this.next = null; } }
class Cell  { n: number; peer: Cell  | null; constructor(n: number) { this.n = n; this.peer = null; } }

function build(n: number): LNode {
  let head = new LNode(0);
  for (let i = 1; i < n; i++) { const nd = new LNode(i); nd.next = head; head = nd; }
  return head;
}

// 1. allocate the parents, 2. churn hard so they get promoted,
// 3. THEN store a fresh young pointer into them.
const olds: Cell[] = [];
for (let i = 0; i < 400; i++) olds.push(new Cell(i));
let churn: any[] = [];
for (let i = 0; i < 200000; i++) { churn.push({ a: i, b: "s" + (i & 255) }); if (churn.length > 1000) churn = []; }
for (let i = 0; i < 400; i++) olds[i]!.peer = new Cell(i + 10000);   // OLD <- YOUNG

let psum = 0; for (let i = 0; i < 400; i++) psum += olds[i]!.peer!.n + olds[i]!.n;
const head = build(30000);
let walk = 0, nodes = 0; let cur: LNode | null = head;
while (cur) { walk += cur.v; nodes++; cur = cur.next; }
console.log(psum, walk, nodes);
```

Both arms print `4159600 449985000 30000`, matching node 26.5.1 — **the wrong answer is not observable at the source level here**, which is the usual shape: the collector had a live edge it did not know about, and this run happened not to reuse the bytes.

It is also invisible without zeal — `PERRY_GC_FROMSPACE_SCAN_ABORT=1` alone gives `0 0 0` on both arms. The fault needs a collection to land inside the window.

### Where I'd look

Change **1** — the sloppy-mode class-field route reaching boxed slots. #7288 opened that route for **raw-f64** slots, which need no barrier; the boxed arm does. If the guard-free constant-slot store inherited the raw arm's barrier-free emission, this is exactly what you would see: correct value, correct dispatch, no dirty bit.

Worth checking against #7602's rule while you're there — the array-push barrier gates on the parent's `GC_FLAG_TENURED` **or** an active incremental mark, and "the parent is young so no barrier" was explicitly measured and rejected as unsound.

### Everything else I checked, which is clean

- `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1 …_DEPTH=800`: exit 0 on both arms, and **not vacuous** — 18 `[gc-fromspace-protect]` / `[gc-copy-minor]` lines, so the copying minor really ran on my probe too.
- `PERRY_GC_VERIFY_EVACUATION` and `PERRY_GC_VERIFY_MARK` under zeal: exit 0. Consistent with the diagnosis rather than against it — `VERIFY_EVACUATION` walks the same enumeration the rewrite pass walks, so it is blind to an edge the remembered set never recorded (this is why `FROMSPACE_SCAN` is the layout-independent instrument, see `GC_LAYOUT_POINTER_FREE`'s doc comment).
- Output byte-identical to node on both arms.

### The gate-drift finding is worth keeping regardless

`NONCOLLECTING` in `gc_root_dominance_check.py` being a second copy of `gc_call_effects.rs`'s fact, drifted since #7510 and invisible only because the corpus contained no class the gate admitted — that is a real latent bug and the 358 spurious violations are the proof. Whatever happens to the rest, that half should land.

Happy to re-audit as soon as the barrier is in; the reproducer above is self-contained.

proggeramlug:
## Withdrawing my block — and the cause was not in this PR

My earlier review reported a deterministic `FROMSPACE_SCAN_ABORT` on this branch (`134` ×6) against a clean `main`, with `dangling=1 … never_dirty=1` on a `Survivor0 → NurseryEden` edge. **That reproduction was real, and my diagnosis of it was wrong.**

Re-tested on current `main`, both arms rebuilt:

| arm | 6 runs of my probe under `ZEAL=1 FROMSPACE_SCAN_ABORT=1` |
|---|---|
| `main` | `0 0 0 0 0 0` |
| this branch | `0 0 0 0 0 0` |

Not vacuous — the copying minor still runs on both (10 and 14 `[gc-fromspace-protect]`/`[gc-copy-minor]` lines), and both print `4159600 449985000 30000`, matching node.

**What changed is `main`, not this branch: #7687 landed.** That was #7682 — an allocation-point collection that *moved* things while running at a program point neither root lowering can describe. My probe's dangling edge was that bug, and this PR is what made it reachable on that shape: moving the store off the by-name path onto a register-held fast path is exactly what turns an imprecise alloc point from theoretical into observable. The interaction was real; the defect was upstream.

**And the barrier is present.** I read the emitted IR rather than inferring from the runtime result. `class_field_sloppy_set.boxed_fast` stores the value, then tag-checks it (`lshr 48` against `0x7FFD` / `0x7FFF` / `0x7FFA`, plus the raw-pointer range test) and branches to `class_field_set.gc_bookkeeping` when it is pointer-like. Both arms emit exactly one `js_write_barrier_slot` and three `js_gc_note_slot_layout` sites. My "the guard-free store inherited the raw arm's barrier-free emission" was a guess, and it was wrong.

I should have read the IR before writing the review. The runtime abort was strong evidence *something* was wrong, and it was — but attributing it to a missing barrier rather than to an upstream moving collection cost you a round trip.

## The audit, now that it stands

**Verified on current main:** 24/24 lint, `cargo fmt --check` clean, `perry-runtime --lib` **1930**, `perry-codegen --lib` **778**, `native_root_coverage` 14/14.

**Pre-existing failure confirmed as claimed**, by checkout rather than inspection: `typed_shape_descriptors::integer_arithmetic_array_push_omits_inbounds_layout_note_and_barrier` fails identically on `main` at `f3e14a61e` — same test, same 17-passed/1-failed count. Not from this PR.

**The gate-drift finding stands on its own merits.** `NONCOLLECTING` in `gc_root_dominance_check.py` being a second copy of a fact `gc_call_effects.rs` already states, drifted since #7510 and invisible only because the corpus then contained no class the gate admitted — that is the same defect class as everything else this week, and the 358 spurious violations are the proof it was load-bearing.

**The three changes landing together is right**, and the reason is worth keeping: #7510 required an *empty* pointer mask, which excluded exactly these classes, so their descriptor arrived after every constructor store and none could pass its intact-bit guard. Discharging obligation 2 by pre-filling every slot with `TAG_UNDEFINED` — which the tracer rejects at its tag check — is a real discharge rather than an avoidance.

`cycles` 0.79 → 0.24, `tree_wide` 12.18 → 3.02 (4.0×), three benchmarks now ahead of scriptc, and the seven that must not move all unchanged with byte-identical stdout.

Issue #6987, opened 2026-07-29
gc: policy gate docs state "default OFF" while the code defaults ON (gc_incremental_enabled, gc_safepoint_moving_minor) — this caused #6972 to dismiss a real concern


proggeramlug:
## Summary

Two GC policy gates document themselves as **default OFF** while their implementations default **ON**.
This is not a cosmetic docs nit: it caused a real correctness concern to be dismissed on a merged PR,
and it was independently tripped over by two separate investigations within hours of each other.

### 1. `gc_incremental_enabled` — `gc/policy.rs:291`

The doc header says:

```rust
/// Phase 4 of the moving-GC project: gate the INCREMENTAL old-gen collector (the
/// budgeted stepper). **EXPERIMENTAL — default OFF.** ...
pub(crate) fn gc_incremental_enabled() -> bool {
    ...
        // DEFAULT ON (#6180 flip): ordinary allocation pressure is collected
```

The header and the body comment eight lines later state opposite defaults. The body is correct — the
`#6180` flip made it ON, and the header was never updated.

### 2. `gc_safepoint_moving_minor` — `gc/policy.rs:1396`

```rust
/// gated by `gc_moving_safepoint_enabled` (default off).
```

`gc_moving_safepoint_enabled` (`policy.rs:279`) is `!matches!(env, "0"|"off"|"false")` — default **ON**,
and its own doc correctly calls it "**Perry's default GC**". Only this reference at :1396 is stale.

(`policy.rs:320`, the `PERRY_GC_MOVING_LOOP_POLLS` line, appears accurate — that one really is opt-in.)

## Why this is worth an issue rather than a drive-by edit

**It changed a merge decision.** PR #6972 dismissed the production-reachability of #6970 on the grounds
that `gc_check_trigger` "forces the conservative scan on both automatic arms". That reasoning depended
on incremental mode being off. With `gc_incremental_enabled` actually default ON, the guarded
direct-minor arm is largely bypassed and budgeted cycles skip the conservative subphase *structurally*,
independent of any guard (`gc/cycle.rs:568-573`, classifier mode).

Measured with `PERRY_GC_TRACE=1` and **no environment variables set at all**: 3 completed cycles,
**2 of them with `conservative_root_count = 0`**. Production genuinely completes precise-roots-only
cycles today. The conclusion in #6972 was reached honestly from the documented behaviour and was wrong
because the documentation was wrong.

For the record, the follow-up investigation could not make #6970's reproducer fail in that
[...]

PR #7687, opened 2026-08-09
fix(gc): stop the collector moving objects at register-imprecise allocation points (#7682)

### LINKED ISSUE #7682: P0: silent wrong answer under the moving collector — stale heap-string deref in generated code (realistic interpreter workload)

proggeramlug:
Fixes #7682.

## The bug

A 189-statement tree-walking interpreter — ordinary TypeScript, no exotic
construct — returned a **silently wrong number** on default settings, every run:
`1708662` where Node and a fully static build both give `1708840`. No crash, no
`TypeError`, no diagnostic.

```
node --experimental-strip-types interp.ts   → 1708840
perry, before                               → 1708662     ← wrong, 6/6 runs
perry, after                                → 1708840      3/3, and with auto-optimize
```

## Root cause

`gc_check_trigger`'s nursery-churn arm collects from inside `arena_cell_alloc`,
i.e. at whatever half-finished expression happened to need a fresh arena block.
**That program point is described by neither root lowering**: the shadow stack
names only values codegen has already stored to a slot, and RS4GC relocates only
what it can type as `ptr addrspace(1)`, which a NaN-boxed `double` operand in an
SSA register is not.

The arm therefore took `ManualGcScanGuard::force_full_scan()`. Its job there is
not retention but **immobility** — a conservative native-stack scan makes the
copying minor ineligible (`CopiedMinorFallbackReason::ConservativeStack`), so the
non-moving in-place minor runs and nothing relocates.

`PERRY_GC_SCAVENGE` gated that guard off. **The guard is now unconditional.**

### Why it was conditional, and why both halves of the reason were false

The flag's doc comment said:

> Phase-1 de-risking flag (OFF by default) … NOT sound as a production default
> yet — the alloc point can be register-imprecise — so it stays behind this flag
> for measurement … only.

Eight lines below it, the body said `ON BY DEFAULT (#7056)`. That is the #6987
shape CLAUDE.md warns about, and this time the stale half was the one carrying
the soundness argument.

The body's own claim — "enabling this also defers alloc-point collections to a
precise safepoint" — was false too: that deferral is gated on
`gc_moving_loop_polls_enabled()`, **OFF by default since #7161**. In the shipped
configuration the two flags disagree, the deferral is dead code, and the
alloc-point minor ran right there with neither a scan nor a safepoint.

### The failure, end to end

`evalNode` lowers `{ names: [n.name], … }` by reading `n.name` into a register,
then inline-bump-allocating the one-element array. The bump overflows its block,
`js_inline_arena_slow_alloc` → `arena_cell_alloc` → `gc_check_trigger` runs an
*evacuating* minor, and the string moves. Control returns to the merge block —
which LLVM tail-merged across the fast and slow paths — and stores the pre-move
address into the new array. `lookup` then compares `names[i] === name`, a live
string against a moved one, falls through to its default, and naive `fib`, which
is just a count of leaves returning `1`, comes back short by exactly the number
of missed lookups.

Localised with `PERRY_GC_PROTECT_FROMSPACE=1 DEPTH=800 PERRY_GC_ZEAL=1`, then
disassembled: the reported frame was the *caller* (the faulting callee is a
frameless leaf), and the faulting instruction is `js_string_addref_if_heap_string`
reading `refcount` at `user_ptr + 12` on the value the tail-merged block just
stored.

## Why every existing gate was green

| gate | verdict | why it cannot see this |
|---|---|---|
| `PERRY_GC_VERIFY_MARK` | OK | marking is correct; the wrong holder is a register, so there is nothing in the heap to find |
| `PERRY_GC_FROMSPACE_SCAN` | no live offender | every owner it reports is `marked=false`, i.e. already dead — same reason |
| `gc_root_dominance_check.py` | 0 violations / 380 root stores | the value is never bound to a slot, so there is no store whose dominance it could question |
| the `test_gap_gc_*` probe corpus | green | every probe holds its subject across an **explicit** churn call; none holds one across the allocation of the literal being built |

## Tests

- **`the_alloc_point_nursery_minor_retains_native_stack_values_under_shipped_pacing`**
  drives the arm with a value reachable only from a live native-stack word and
  asserts it survives, that a collection actually ran, and that the census
  counted the forced scan.
- **`the_alloc_point_plant_dies_when_the_scan_is_pinned_off`** is its sabotage
  control: identical plant, scan pinned off, plant must DIE. Without it, "the
  malloc sweep never ran" and "the guard held" are the same green.
- **`policy::force_shipped_default_gc_pacing`** pins polls OFF + scavenge ON.
  That combination had no test guard — `force_legacy_gc_pacing` pins both off,
  `force_moving_gc_pacing` pins both on — so every test in the crate declared a
  pacing mode in which the two flags *agreed*, and the interaction that broke is
  exactly the one where they disagree.
- **`test-files/test_gap_gc_alloc_point_no_move.ts`** is the interpreter itself,
  diffed byte-for-byte against Node, registered in
  `test-parity/gc_repsel_corpus.txt` so `gc-moving-witnesses` runs it.

**Verified non-vacuous both ways.** The gap probe prints `1708662` against the
pre-fix runtime under the harness's own `PERRY_NO_AUTO_OPTIMIZE=1`; and with the
one-line guard reverted, the new unit test fails.

## Cost, stated plainly

Alloc-point nursery collections are non-moving again — what they were before
#7056. Copying minors keep running at the precise safepoints
(`gc_safepoint_moving_minor`), where the root set is real. `PERRY_GC_SCAVENGE`
keeps its other job (routing nursery-churn triggers to the direct minor rather
than the budgeted non-moving stepper) and is now documented as the pacing knob
it is.

**The `gc-ratchet` baseline pins the old, evacuating behaviour and will report a
breach on the evacuation counters.** That artifact is regenerated only when a
change is deliberately accepted, on the pinned quiet host; this is such a change.
Measured delta is in the comments below.

The way to get the evacuation back *soundly* is to make the alloc point precise,
or to let the copying minor run with conservative roots as **pinning** sources
rather than refusing outright (`CopiedMinorEligibility::evaluate_with_stack_decision`
returns `ConservativeStack` unconditionally today). Both are their own change
with their own proof obligation, and neither belongs in a P0 correctness fix.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed allocation-point garbage collections so live values in imprecise registers are retained safely.
  - Ensured these collections remain non-moving while preserving moving collections at precise safepoints.
  - Clarified that garbage-collection pacing settings control timing, not collection safety or object movement.

- **Tests**
  - Added regression coverage for native-stack values, allocation-point collections, and shipped pacing defaults.
  - Added interpreter-based tests covering repeated allocations, closures, recursion, and cyclic environments.

- **Documentation**
  - Updated garbage-collection guidance to explain collection pacing and movement behavior.


proggeramlug:
## Audit — merging as v0.5.1389. **P0 confirmed and fixed.**

Reproduced independently, both arms built `--profile perry-dev` with `PERRY_RUNTIME_DIR` pinned, `PERRY_NO_AUTO_OPTIMIZE=1`:

| arm | 6 consecutive runs |
|---|---|
| `main` | `1708662` ×6 — **wrong, deterministic** |
| this branch | `1708840` ×6 |
| node 26.5.1 | `1708840` |

A silently wrong number on default settings, from ordinary TypeScript. No crash, no diagnostic. That is the worst failure shape this project has.

### The root cause is the one that matters most

> That program point is described by **neither** root lowering: the shadow stack names only values codegen has already stored to a slot, and RS4GC relocates only what it can type as `ptr addrspace(1)`, which a NaN-boxed `double` operand in an SSA register is not.

So the guard's job at the alloc point is not retention but **immobility** — the conservative scan makes the copying minor ineligible, and that is the entire mechanism keeping a register-held value valid across a collection nobody can describe. Gating it on a pacing flag traded a cost problem for a soundness problem, which is exactly what #7148 forbids.

**And both halves of the reason it was conditional were false.** The doc comment said "OFF by default … NOT sound as a production default"; eight lines below, the body said "ON BY DEFAULT (#7056)". That is the #6987 shape CLAUDE.md warns about — *"a merge decision was made on the wrong one"* — and this time the stale half was the one carrying the soundness argument. The body's fallback claim (that enabling it also defers to a precise safepoint) was gated on `gc_moving_loop_polls_enabled()`, **off by default since #7161**, so in the shipped configuration the deferral was dead code and the two flags disagreed.

### Sabotage verified independently

Restoring the `(!gc_scavenge_enabled()).then(…)` gate on the **nursery-churn** arm reddens `the_alloc_point_nursery_minor_retains_native_stack_values_under_shipped_pacing` while its control `the_alloc_point_plant_dies_when_the_scan_is_pinned_off` stays green — `error[` 0, test binary reached. (My first attempt re-gated the *old-reclaim* arm by mistake and reddened a different test; that sabotage proved nothing and I redid it.)

The control is the part I want to single out: **"the malloc sweep never ran" and "the guard held" are the same green** without it. That is the distinction four of this repo's gates have failed to make.

`force_shipped_default_gc_pacing` is the other real finding — every test in the crate declared a pacing mode in which the two flags *agreed*, so the one interaction that breaks had no guard at all.

### Why every existing gate was green

The table in the description is correct and worth keeping: `VERIFY_MARK` sees correct marking; `FROMSPACE_SCAN` reports only `marked=false` owners; the dominance checker has no store whose dominance it could question, because the value is never bound to a slot; and every `test_gap_gc_*` probe holds its subject across an **explicit** churn call rather than across the allocation of the literal being built. Four instruments, one blind spot, and the new probe is shaped to sit in it.

**Gates:** 24/24 lint, fmt clean, `perry-runtime --lib` **1930**, `perry-codegen --lib` 778, and the gap test is registered in `gc_repsel_corpus.txt` so `gc-moving-witnesses` actually runs it rather than it being a dark test.

### The cost, and the immediate follow-up

Alloc-point nursery collections are non-moving again — what they were before #7056 — with copying minors still running at the precise safepoints where the root set is real. That is the right trade for a correctness bug of this shape.

**This PR does not carry the ratchet re-pin, so `gc-ratchet` will breach on the evacuation counters the moment it lands.** I am merging anyway because a deterministic wrong answer on default settings outranks a red measurement gate — but the re-pin is now the top of my queue, on the pinned quiet mini, with the delta recorded as a deliberate acceptance. Anyone reading a red `gc-ratchet` before that lands should attribute it here.

The two sound routes you name for getting the evacuation back — make the alloc point precise, or let the copying minor treat conservative roots as **pinning** sources rather than refusing outright (`evaluate_with_stack_decision` returns `ConservativeStack` unconditionally today) — are the right framing, and correctly out of scope for a P0 fix.

proggeramlug:
## Measured cost, and why it is recoverable

Pinned quiet host (load 1.4), 5 repeats, the interpreter probe. Same binary in
every arm except **D**, which is a separate compile because the poll flag is
compile-time as well as run-time.

| arm | wall | peak RSS | answer |
|---|---|---|---|
| **A** fixed default — non-moving at the alloc point | 9.06 s | 58.2 MB | **1708840 ✓** |
| **B** `PERRY_CONSERVATIVE_STACK_SCAN=off` — the pre-fix code path | 4.78 s | 88.4 MB | 1708839 ✗ |
| **C** fixed + `PERRY_GC_SCAVENGE=0` | 6.88 s | 102.8 MB | 1708840 ✓ |
| **D** fixed + `PERRY_GC_MOVING_LOOP_POLLS=1` | 4.85 s | 88.4 MB | **1708840 ✓** |

**B is a one-env-var reproducer on the fixed binary** — `PERRY_CONSERVATIVE_STACK_SCAN=off`
beats `force_full_scan` (asserted by `conservative_scan_env_off_still_beats_a_forced_scan`),
so it re-enters the exact pre-fix path: 37 copying minors, wrong answer, same
compiler, same runtime. That is the cleanest A/B available and it is what the
table is built on.

Two things worth reading off it:

1. **This is not a memory regression.** A retains **34% less** than the unsound
   arm it replaces. The nursery cap plus a non-moving minor keeps a lower
   high-water mark here than evacuation does. The trade is time, not footprint.
2. **The time is fully recoverable, soundly.** Arm D runs the *same 37 copying
   minors* as the unsound arm, at the same RSS, within noise of its wall time,
   and gets the right answer — because a back-edge poll puts the copying minor
   at a **declared safepoint**, where the root set is real.
   `PERRY_GC_MOVING_LOOP_POLLS` is off by default only as #7161's stopgap for
   #7154, whose class now has its own static gate (`gc-root-dominance.yml`) with
   an empty allowlist. Re-examining that stopgap is the obvious follow-up, and
   this change is a precondition for it rather than an obstacle.

## The collector was already printing this

`PERRY_GC_DIAG=1`, default env, pre-fix:

```
[gc-copy-minor] ran copied_objects=6149 … trigger=ArenaBytes declared_safepoint=false
```

37 of those, and zero `nursery_churn_slack_valve` scan fallbacks. After the fix
it inverts exactly: 0 copying minors, 376 valve fires. `declared_safepoint=false`
on a `[gc-copy-minor] ran` line is the bug in one field.

## It also restores a claim already checked in

`test-parity/gc_matrix_inert_arms.txt` registers the `default` matrix arm as:

> Shipped config + pressure. … It still COLLECTS; **it cannot scavenge**.

`main` contradicted that. After this change the registry entry is true again.

## Gates

- `gc-native-roots`' evacuation arms set `PERRY_CONSERVATIVE_STACK_SCAN=off`, which
  beats the guard, so they are unaffected — `gc_evacuation_liveness_assert.py` still
  sees its copying minors.
- `gc-moving-witnesses` runs the `loop_polls` arm, which defers to a safepoint and
  never reaches the guard; the new probe relocates there like the rest of the family.
- **`gc-ratchet` will breach** on the evacuation counters (`copied_objects`,
  `freed_bytes`) — that artifact pins the old behaviour and needs regenerating on
  the pinned host as a deliberate acceptance. It is not a required context.

## Known sharp edge, not changed here

`PERRY_CONSERVATIVE_STACK_SCAN=0`/`off` still opts back into the unsound
configuration — that is arm B. It is documented as a bisection escape hatch and a
test pins the precedence, so narrowing it is a separate decision; but it is worth
knowing that this hatch now buys wrong answers, not merely less retention.

proggeramlug:
## Update: the fix now costs nothing

The soundness fix alone was **not shippable**, and chasing that produced two more
changes. Final state on the pinned quiet host, `iso_FIB`, 5 repeats:

| | wall | peak RSS | answer |
|---|---|---|---|
| `main` before this PR | 4.78 s | 88.4 MB | 437839 ✗ |
| **this PR** | **4.80 s** | **88.4 MB** | **437840 ✓** |
| (1) alone — the guard | 9.06 s | 58.2 MB | ✓ |
| (1)+(2), polls off (`PERRY_GC_MOVING_LOOP_POLLS=0`) | 5.20 s | 248.7 MB | ✓ |

Same speed, same footprint, right answer. 37 copying minors — the same count the
unsound path ran — and now **all 37 report `declared_safepoint=true`**, where
before every one said `false`. Zero alloc-point valve fires.

### (2) The nursery cap applies only when the minor can evacuate

Fixing (1) alone took `test_gap_gc_index_get_receiver_rooting` from **0.66 s to
6.6 s**. That is a livelock, not a trade: the cap's basis is
`copying_from_space_in_use_bytes()`, which a **non-moving** minor never reduces —
it sweeps in place and from-space stays occupied — so a capped trigger firing a
non-moving minor is due again on the very next block. One whole-arena collection
per 1 MB allocated. Confirmed with no rebuild at all, via the tuning dial:
`PERRY_GC_SCAVENGE_NURSERY_MB=4096` takes that test to **0.13 s**. Same shape as
#7592.

#7056's own 2x2 already said the cap and the evacuating minor "ship together,
because either alone is a bad trade". That was advisory; it is now load-bearing.

### (3) Moving-loop back-edge polls default-ON (#7161's stopgap retired)

Both conditions #7161 named for putting this back are met:

* **Its correctness reason is closed.** Its own title is "pending #7154"; #7154
  closed 2026-08-01, and that class now has a static gate whose allowlist is
  empty.
* **Its codegen-quality reason is discharged by its own stated condition** —
  *"until the poll is emitted only in loops that actually ALLOCATE"*. It already
  is: `emit_gc_loop_safepoint` consults `loop_purity::loop_may_allocate`. Only
  the doc comment above the flag never got updated.

And after (1), leaving it off was the *more* dangerous state. Nursery pressure
has exactly two precise collection points — this poll and the microtask-pump
boundary — and a compute-only program reaches neither with polls off. "Polls
off" never meant "collect later, precisely"; it meant "never collect precisely
at all".

**The flip is less of a leap than it looks.** `gc-moving-witnesses` runs
`gc_repsel_matrix.sh --arms loop_polls --filter test_gap_gc_` on every PR, and
`loop_polls` **is** the configuration this makes default. The whole 56-file
`test_gap_gc_*` corpus has therefore been gated in exactly this mode all along,
with UNVER rejected as hard as FAIL. The matrix header's warning that these arms
were red on 14 of 20 files when they first moved is about a state that job has
been holding ever since.

The four `#7161` entries in `test-parity/gc_matrix_inert_arms.txt` named this
flip as their deletion condition and are deleted. One of them — `default | It
still COLLECTS; it cannot scavenge` — was **false on main for the whole time it
was listed**: the shipped default *was* scavenging, at the allocation point. The
liveness gate checks that a listed arm has not started biting again; nothing
checked whether the reason given for listing it was still true.

### Two tests changed, and neither was waved through

`generator_attach_prototype`'s pair (#7577) failed on their own **live-subject**
assertion — *"subject not live — no copying minor moved the receiver during the
call, so this test proved nothing"* — not on a wrong address. They inject their
collection at an allocation point, which (1) makes unconditionally non-moving, so
nothing relocated and they correctly refused to pass. They now pin the
conservative scan off across the injected window (`AllocPointRelocationGuard`,
restored on drop including on a panicking assert) — the documented lever that
beats `force_full_scan`, and the one the matrix's `%E%` arms use. What they
assert is unchanged: **a runtime helper must not bind a receiver's address across
its own allocation.** (1) removes one route to that hazard in the shipped
default; it does not make the helper correct, and
`PERRY_CONSERVATIVE_STACK_SCAN=off` is a supported configuration where the route
is still open.

`test_effective_arena_trigger_respects_armed_values` mirrors the cap gate, so it
follows it.

### Local verification

| check | result |
|---|---|
| interpreter / `iso_FIB` / gap probe, default env | correct, 3/3 each, byte-equal to Node |
| `[gc-copy-minor] declared_safepoint` | 37/37 `true` (was 0/37) |
| from-space protector, `DEPTH=800`, default env | **clean**, instrument live (38 retired page-sets) |
| `gc_root_dominance_check.py` under polls-ON codegen | 0 violations / 372 root stores |
| new alloc-point unit test with the guard reverted | **FAILS**, as it must |
| gap probe against the pre-fix runtime | `341590` vs oracle `341768` — non-vacuous |

### Residual risk, stated

The rest of the gap corpus (479 tests) under polls-ON is **not** gated per-PR —
`parity` is tag-gated — and I could not run it locally (the box is at 100% disk).
`gc-moving-witnesses` covers the `test_gap_gc_*` slice, which is the slice this
class lives in, but a broader sweep before merge would be worth someone's disk.

`gc-ratchet` pins the pre-change evacuation accounting and needs regenerating on
the pinned host.

Issue #7577, opened 2026-08-07
js_generator_attach_prototype derefs retired from-space memory (SIGBUS under GC zeal)


proggeramlug:
Constructing a generator in a hot loop takes a **bus error on retired
from-space memory** under the #7154 rooting instruments. The faulting frame is
`js_generator_attach_prototype`, i.e. generator construction — not stepping,
not the result object.

Found while working #7564. **Not caused by it** — #7564 touches no generator
code (`git diff --stat` covers only `array/iter_object.rs`, `buffer/iter.rs`,
`collection_iter_object.rs`, `iterator_helpers.rs`, `string/iter_object.rs`,
`string/mod.rs`, `iter_result.rs`, `gc/mod.rs`). It surfaced only *after*
#7564, because before that the process died earlier at
`iterator_helpers::make_iter_result` and never reached this site.

## Repro

```ts
function churn(i: number): string {
    return "pad-" + i + "-" + (i * 7919) + "-" + (i % 13);
}
function* small(n: number): Generator<number, void, undefined> {
    for (let k = 0; k < n; k++) yield k + n;
}
let acc = 0;
for (let i = 0; i < 4000; i++) {
    const g = small(4);          // ← construction is what faults
    let gs = g.next();
    while (!gs.done) { acc += gs.value as number; gs = g.next(); }
    churn(i);
}
console.log(acc);
```

Compile with `PERRY_GC_MOVING_LOOP_POLLS=1` (loop polls are off by default
since #7161, so a default build never collects inside the loop), then:

```
PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1 \
PERRY_GC_PROTECT_FROMSPACE_DEPTH=800 PERRY_GC_DIAG=1 ./repro
```

## Fault

```
[gc-fromspace-protect] mode=ProtectPages retired_set=#0 blocks=17 sets_held=1/800
                       bytes_protected=17825792
[gc-fromspace-protect] FAULT: signal 10 at 0x2cc0000fea0
  This address is RETIRED FROM-SPACE. The evacuating minor moved or
  freed the object here and the holder kept the pre-collection address.
  block=0x2cbfff10000 +1048224 retired_bytes=1048512 retired_by_minor=#1
  last-known object: user_ptr=0x2cc0000fea8 obj_type=2 size=72
2  js_generator_attach_prototype + 248
3  js_generator_attach_prototype + 248
4  perry_fn_..._small + 668
5  main + 7840
```

[...]

PR #7690, opened 2026-08-09
fix(gc): restore evacuation at precise safepoints — the pacing half of #7682

proggeramlug:
Follow-up to #7687, which landed **only the first of three changes**.

#7687's guard is correct — an allocation-point collection may not MOVE anything,
because that program point is described by neither root lowering. On its own it
is also **not shippable**, and `main` is in that state right now:

| | wall | peak RSS | answer |
|---|---|---|---|
| before #7687 | 4.78 s | 88.4 MB | 437839 ✗ |
| **main today (#7687 alone)** | **9.06 s** | 58.2 MB | 437840 ✓ |
| **this PR** | **4.80 s** | **88.4 MB** | **437840 ✓** |

`test_gap_gc_index_get_receiver_rooting` went **0.66 s → 6.6 s** on main. Gap
tests near the harness's 10 s `PERRY_RUN_TIMEOUT` can now be recorded as
`crash` — that is how #7687's own probe first failed, before it was resized.

## 1. The nursery cap applies only when the minor can evacuate

The cap's basis is `copying_from_space_in_use_bytes()`, which a **non-moving**
minor never reduces — it sweeps in place and from-space stays occupied. A capped
trigger firing a non-moving minor is therefore due again on the very next block:
one whole-arena collection per 1 MB allocated. Confirmed with no rebuild at all,
via the tuning dial — `PERRY_GC_SCAVENGE_NURSERY_MB=4096` takes that test to
**0.13 s**. Same shape as #7592.

#7056's own 2x2 already said the cap and the evacuating minor "ship together,
because either alone is a bad trade". That was advisory; `nursery_cap_active`
makes it load-bearing.

## 2. Back-edge polls default-ON (#7161's stopgap retired)

Both conditions #7161 named are met:

* **Correctness** — its own title is "pending #7154"; #7154 closed 2026-08-01,
  and that class now has a static gate whose allowlist is empty.
* **Codegen quality** — the objection was *"the poll emits a call at every loop
  back-edge … **until** the poll is emitted only in loops that actually
  ALLOCATE"*. It already is: `emit_gc_loop_safepoint` consults
  `loop_purity::loop_may_allocate`. Only the doc comment above the flag never
  got updated.

And after #7687, leaving it off is the *more* dangerous state. Nursery pressure
has exactly two precise collection points — this poll and the microtask-pump
boundary — and a compute-only program reaches neither with polls off. "Polls
off" never meant "collect later, precisely"; it meant "never collect precisely
at all".

**The flip is less of a leap than it looks:** `gc-moving-witnesses` runs
`gc_repsel_matrix.sh --arms loop_polls --filter test_gap_gc_` on every PR, and
`loop_polls` **is** the configuration this makes default. That 56-file corpus
has been gated in exactly this mode all along, with UNVER rejected as hard as
FAIL.

It also makes an assertion already on `main` true again:
`test-parity/gc_matrix_inert_arms.txt` no longer lists `default` as known-inert,
i.e. `main` currently claims the shipped default relocates while polls are off.

## 3. The two #7577 witnesses

`generator_attach_prototype`'s pair inject their collection at an allocation
point, which after #7687 neither moves nor — with polls on — happens there at
all. Both failed on their own **live-subject** assertion ("subject not live"),
correctly refusing to pass while proving nothing. They now pin
`force_shipped_default_gc_pacing()` plus a scan override, and assert exactly what
they asserted before: **a runtime helper must not bind a receiver's address
across its own allocation.**

It has to be that guard and not `force_legacy_gc_pacing()`, which also turns
scavenge off — and scavenge is the disjunct that routes nursery pressure to the
direct arm at all (`registered_root_scanners_block_budgeted_gc()` reduces to
"any COPY-ONLY scanner" under `gc_incremental_enabled()`, and that registry holds
only a mutable one). With scavenge off the trigger goes to the budgeted stepper,
which is non-moving by construction — a third route to the identical message.
All three are recorded in the file.

## Verification

| check | result |
|---|---|
| `cargo test -p perry-runtime --lib` | **1930 passed, 0 failed** |
| interpreter / `iso_FIB` / gap probe, default env | correct, byte-equal to Node |
| `[gc-copy-minor] declared_safepoint` | 37/37 `true` (was 0/37) |
| from-space protector, `DEPTH=800` | clean, instrument live (38 retired page-sets) |
| `gc_root_dominance_check.py` under polls-ON codegen | 0 violations / 372 root stores |

`gc-ratchet` pins the pre-change evacuation accounting and needs regenerating on
the pinned host.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored moving garbage-collection checks at loop back edges by default.
  * Nursery limits now apply only when collection can relocate nursery objects.
  * Explicitly disabling moving-loop checks remains supported through configuration.
  * Improved garbage-collection behavior for object relocation and allocation-point scenarios.

* **Documentation**
  * Clarified garbage-collection configuration defaults, trigger behavior, safety considerations, and supported collection paths.
  * Added benchmark and validation details to the changelog.


proggeramlug:
## The `default` matrix arm: green, and demonstrably relocating

`gc_repsel_matrix.sh --arms default --filter test_gap_gc_` on this branch, over
the full 56-file registered corpus (31 cells after the filter):

```
summary: PASS=29 UNVER=2 XFAIL=0 FAIL=0
byte-exact vs node 26.5.1: 31/31 cells   (the parity property)

arm liveness across the corpus (cells where the arm actually bit):
  default   requires=scavenge   collected 30/31   reclaimed 29/31
                                moved-objects 29/31   copy-minor 29/31

liveness gate (#7255): every arm must have exercised what it claims
  default   requires=scavenge   live 29/31   counter=269199   live
```

This is the check the matrix header warns about: *"AND WHEN THESE ARMS FIRST
MOVED, THEY WERE RED"* — 14 of 20 corpus files, back when the shipped default
first started relocating. It is green now, and the arm is **not** vacuous:
269,199 objects actually moved, on 29 of 31 cells, with every cell byte-exact
against the pinned oracle.

`gc_index_get_receiver_rooting` — the file that runs at **6.6 s on `main`
today** and **0.66 s before #7687** — passes here.

The two `UNVER` cells (`gc_staging_args_rooting`, `gc_interval_args_rooting`)
are "output matched but the arm was inert on this file": those two programs
never drove a copying minor. That is a per-cell coverage note, not a failure —
the `#7255` liveness gate is about an arm going inert *corpus-wide*, and this
one bit on 29 of 31.

proggeramlug:
## Parts 1 and 3 are right. Part 2 silently reverts #7669's 34×.

First, confirming the premise, because it is mine to own: **#7687 alone is a severe regression and I shipped it.** On this host `test_gap_gc_index_get_receiver_rooting` is **6.94 s on `main` vs 0.76 s on this branch — 9.1×**, both printing the correct `bad 0`. The nursery-cap argument is convincing on its own terms: the cap's basis is `copying_from_space_in_use_bytes()`, which a **non-moving** minor never reduces, so a capped trigger firing a non-moving minor is due again on the very next block. That is a real defect and `nursery_cap_active` is the right shape for it.

And the safepoint number is the strongest thing in the PR:

| arm | collections at a **declared** safepoint |
|---|---|
| `main` (polls off) | **0** |
| this branch | **38** |

"Polls off" really does mean "never collect precisely at all", and after #7687 that is the more dangerous state.

## The blocker

**Three codegen tests fail on this branch and pass on `main` (778/0):**

```
element_shape_versioned_loop_fires_for_the_7480_access_shape          FAILED
element_shape_versioned_loop_resolves_an_object_literal_element_type  FAILED
the_repair_does_not_put_a_call_inside_the_fast_clone                  FAILED
```

Measured consequence on #7480's kernel (200k × 50, identical checksums):

| arm | wall |
|---|--:|
| `main` | **17 ms** |
| this branch | **114 ms** |
| node 26.5.1 | 15 ms |

**6.7×.** The clone blocks are still emitted — 56 on both arms — but the fast arm now carries **4 `js_gc_loop_safepoint` calls**, so `LlBlock::contains_gc_unsafe_call` declines it and the deref block branches unconditionally to the slow clone. #7669 landed four hours ago and this returns that kernel to roughly its pre-#7669 state.

## Why the obvious repair is unsound, and what I think is right

The tempting fix is to teach the admission that a back-edge poll is not a "GC-unsafe call" for its purposes. **That is unsound**, and `element_shape_loop.rs`'s own module docs say why:

> call-free is precisely "no funnel that can revoke the invariant, and **no allocation that can move the array**, runs while the clone does"

A poll can collect, a collection can move the array, and the fast clone holds a base derived in the preheader. That is the #7660 growth-forwarding SIGBUS shape with a different trigger — and #7660 was three days ago.

**The sound repair is to emit the poll per-arm rather than per-loop.** `emit_gc_loop_safepoint` already consults `loop_purity::loop_may_allocate`, but that decision is taken for the *loop*, before the versioned split. After the split the two arms have different answers by construction: the fast clone is **provably call-free**, so it cannot allocate, so it needs no poll; the slow clone may allocate and should keep one. That preserves both this PR's precision argument and #7669's win, and it needs no new soundness claim — it applies the existing predicate at the granularity where it is actually true.

If that turns out to be more than a follow-up, **splitting is fine**: parts 1 and 3 stand alone and fix the regression I shipped. Worth measuring whether the cap fix alone recovers most of the 9.1× — that is one build, and if it does, part 2 can land on its own schedule with the per-arm poll.

## Not yours

`promise::keyed_table::tests::settling_many_keys_is_not_quadratic` fails **identically on `main`** (1929 passed / 1 failed on both). Pre-existing; don't chase it.

## Everything else I checked

`test_gap_gc_alloc_point_no_move` → `1708840` on both arms, matching node — #7687's correctness fix is intact here. My own old→young barrier probe prints `4159600 449985000 30000` on both, and `FROMSPACE_SCAN_ABORT` and `PROTECT_FROMSPACE` each exit 0 on this branch (the zeal+abort combination times out rather than faulting, which is expected: polls-ON under zeal collects at every back-edge). 24/24 lint, `cargo fmt --check` clean, the two re-pinned `generator_attach_prototype` tests green.

Re-pinning those two on `force_shipped_default_gc_pacing()` rather than `force_legacy_gc_pacing()` is right for the reason you give — legacy also turns scavenge off, and scavenge is the disjunct that routes nursery pressure to the direct arm at all.

proggeramlug:
## Measured: part 2 is a net negative on both benchmarks — the cap fix does all the work

I built the "parts 1+3 only" configuration (this branch, polls forced off at both compile and run time, so the cap fix and the doc/test changes are live and the default flip is not):

| configuration | `index_get_receiver_rooting` | #7480 kernel |
|---|--:|--:|
| `main` (#7687 alone) | 6.94 s | 17 ms |
| this branch, **polls ON** | 0.76 s | **126 ms** |
| this branch, **polls OFF** | **0.32 s** | **19 ms** |

Two things fall out, and the first one I did not expect:

1. **The polls flip makes the regression test it targets 2.4× SLOWER** — 0.32 s → 0.76 s. The 9.1× recovery is entirely the nursery-cap fix. Part 2 is not paying for itself even on its own benchmark.
2. On #7480's kernel it costs **6.6×** (19 ms → 126 ms), for the reason in my previous comment: the back-edge poll is a call, `contains_gc_unsafe_call` declines the fast clone, and the deref block branches unconditionally to the slow arm.

So the split is not a scheduling compromise — **parts 1 and 3 are strictly better than the whole PR on both measurements**, and they fix the regression I shipped with #7687. Checksums identical in every cell (`bad 0`, `999995000000`).

## This does not weaken the argument for polls

The precision case stands and I still find it convincing: 0 declared safepoints with polls off, 38 with them on, and after #7687 "polls off" genuinely means "never collect precisely at all". That is a real hole and it should be closed.

But it should be closed **with the per-arm poll emission**, not before it. `emit_gc_loop_safepoint` already consults `loop_purity::loop_may_allocate`; applying that predicate after the versioned split rather than before it gives the fast clone (provably call-free, therefore provably non-allocating) no poll and the slow clone one. Then the precision win and #7669's win coexist, and the numbers above suggest the poll's cost on ordinary loops is worth re-measuring on its own once it is not being paid by the clone.

**Suggested shape:** land parts 1 + 3 now as the #7687 regression fix — they need no new soundness argument and they are the better configuration today. Take part 2 as its own PR with the per-arm emission, where its cost can be measured against a fast clone that still fires.

Happy to re-audit either half as soon as it is up.

proggeramlug:
## Review adjustments pushed (`6a67940`)

Thanks — the Major finding was right, and it is a variant of the defect this
branch exists to fix.

**`force_shipped_default_gc_pacing()` pins polls OFF, which stopped being the
shipped default in the same PR that introduced the guard.** Three tests — one of
them already merged in #7687 — were claiming to assert the default while
asserting the kill switch. A name that quietly became false, a few lines from
the code that falsified it.

* Renamed to **`force_alloc_point_minor_pacing()`** and documented as the
  `PERRY_GC_MOVING_LOOP_POLLS=0` configuration it selects, with the history kept
  (it *was* the default; it is where #7682 was found).
* The three tests renamed to name their mode. Their assertions stand: `=0` is
  supported, and "the collection moved elsewhere by default" is not a reason to
  let the alloc-point minor relocate when a user turns that route off.

**On the second half — "add a witness with polls enabled that relocates at a
precise loop safepoint" — I did not, deliberately.** Under the default there is
no collection inside the callee to relocate anything: the trigger defers and the
minor runs after the helper has returned. That test would report *"subject not
live"* forever — a gate that cannot fail, which is the failure mode CLAUDE.md is
most insistent about.

Instead, `the_shipped_default_defers_the_trigger_out_of_the_callees_window`
asserts the routing that **closed** the window: no collection at the allocation
point, and `GC_SAFEPOINT_PENDING` set — the latter doubling as the live-subject
check, so it cannot pass on an unarmed trigger. "The window is closed" becomes a
measurement instead of an argument in a comment.

**The other two:**

* `gc-ratchet` — the fragment now scopes it explicitly as out of scope for this
  PR, names the re-pin command, and states that a breach on the evacuation
  family is expected until a maintainer runs it on the pinned host.
* rooting-invariant doc — `2f0fe92` fixed line 27; **lines 53-54 still said the
  poll is off by default**. Now: coverage exists only where an allocating loop
  emits a poll (`loop_purity::loop_may_allocate`), and `=0` removes it.

`cargo test -p perry-runtime --lib`: **1931 passed, 0 failed**, including all
three generator witnesses.

proggeramlug:
## Merging, reduced to the nursery-cap fix. I made the change rather than asking for it.

The newest commits (`6a679408b` and before) address the pacing guard, the generator tests and the docs — but not the element-shape interaction, and the three tests are still red on the branch, including `the_repair_does_not_put_a_call_inside_the_fast_clone`, which is named for exactly this. So I reverted part 2 here and left parts 1 and 3.

**Final measurement, all on one build:**

| configuration | `index_get_receiver_rooting` | #7480 kernel |
|---|--:|--:|
| `main` (#7687 alone) | 6.94 s | 17 ms |
| this PR **with** the flip | 0.76 s | **126 ms** |
| **as merged** | **0.11 s** | 17 ms |

Item 6 is genuinely untouched, not just "close": five runs each, fix `16 19 17 31 20` against main `18 22 21 31 16`, identical checksums. And **0.11 s is better than either arm I measured before** — the nursery-cap fix does all of the recovery and then some.

`test_gap_gc_alloc_point_no_move` still prints `1708840`, so #7687's correctness fix is intact.

### What I changed and why

Reverted the default on **both** sides (`moving_safepoint_polls_enabled` in codegen, `moving_loop_polls_enabled_from_env` in the runtime — they must agree or codegen emits polls nothing consumes), and removed `test_moving_loop_minor_on_by_default_7682`, which pinned the flip.

Kept everything else: the `nursery_cap_active` gate, the trigger-arithmetic test, the two re-pinned `generator_attach_prototype` tests, and the doc corrections. Also resolved a `CLAUDE.md` conflict in `main`'s favour — its zeal row is current post-#7684 (which **deleted** `PERRY_GEN_GC_EVACUATE`) and post-#7604 (which added the exit-70 verdict); the branch's copy still described the deleted knob as able to veto forced evacuation.

### The precision argument is not withdrawn

0 declared safepoints with polls off against 38 with them on is real, and after #7687 "polls off" does mean "never collect precisely at all". **That should land — with per-arm emission.** `emit_gc_loop_safepoint` already consults `loop_purity::loop_may_allocate`; applying that predicate *after* the versioned split gives the fast clone (provably call-free, therefore non-allocating) no poll and the slow clone one. Then both wins coexist.

The shortcut of teaching the admission to ignore polls is **unsound**, and `element_shape_loop.rs` says why in its own words: call-freeness is "no funnel that can revoke the invariant, and **no allocation that can move the array**, runs while the clone does". The fast clone holds a preheader-derived base — that is #7660's SIGBUS shape with a different trigger.

I have recorded all of this in the changelog fragment so the follow-up starts from the measurement rather than from the argument.

**Gates:** 24/24 lint, `cargo fmt --check` clean, `perry-codegen --lib` **778**, `perry-runtime --lib` **1935**, `cargo check --all-targets` clean. The `settling_many_keys_is_not_quadratic` timing flake did not recur and is pre-existing on `main` either way.

Issue #446, opened 2026-05-04
perry-codegen: `import type { Class }` silently no-ops method calls on parameters of that class type


proggeramlug:
## What happened

When a class is imported into a consumer module via TypeScript `import type { Foo }`, perry codegen drops the method-dispatch path for parameters typed `Foo`. At runtime, `obj.method(...)` calls on those parameters **silently no-op** — the body never runs, no `TypeError` is thrown, the call appears to succeed. Direct field reads/writes on the same object still work, because perry resolves those structurally.

This is a *latent* bug that bites whenever a refactor that splits a large class file uses `import type` for class types in a sibling module. In the wild it caused the `@codehz/ecs` library's command-buffer execution path to silently drop every `world.set(...)` — basic demos that bun runs cleanly produce wrong values or crash with downstream archetype errors when compiled with perry.

## Minimal reproduction

Four files. Type-only import of the class is the only trigger.

```typescript
// issue_a.ts — defines the class
import { bumpBodyRan } from "./issue_counter.ts";

export class Inner {
  readonly adds = new Map<number, string>();

  setAdd(k: number, v: string): void {
    bumpBodyRan();
    console.log("[Inner.setAdd] body ran");
    this.adds.set(k, v);
  }
}
```

```typescript
// issue_counter.ts — independent counter, proves the body runs (or doesn't)
let n = 0;
export function bumpBodyRan(): void { n++; }
export function getBodyRan(): number { return n; }
```

```typescript
// issue_consumer.ts — consumer with TYPE-ONLY import (the bug trigger)
import type { Inner } from "./issue_a.ts";

export function consume(inner: Inner): void {
  console.log("[consume] typeof inner.setAdd =", typeof (inner as any).setAdd);
  console.log("[consume] adds.size before  =", inner.adds.size);
  inner.setAdd(1, "hello");
  console.log("[consume] adds.size after   =", inner.adds.size);
}
```

```typescript
// issue_main.ts — driver
import { Inner } from "./issue_a.ts";
import { consume } from "./issue_consumer.ts";
import { getBodyRan } from "./issue_counter.ts";

const inner = new Inner();
consume(inner);
console.log("Inner.setAdd body ran:", getBodyRan(), "time(s)");
console.log("inner.adds.get(1)    :", inner.adds.get(1));
```

## Expected output (bun)

```
[consume] typeof inner.setAdd = function
[consume] adds.size before  = 0
[Inner.setAdd] body ran
[consume] adds.size after   = 1
[...]

PR #7691, opened 2026-08-09
fix(class-refs): static method extracted as a value must not resolve the same-named instance method (#7689)

### LINKED ISSUE #7689: npm: `marked` compiles natively but throws at runtime — "Cannot read properties of undefined (reading 'pedantic')"
## Summary

Perry compiles the `marked` npm package natively ("2 native, 0 JavaScript") but the
result is functionally broken: any parse throws

```
TypeError: Cannot read properties of undefined (reading 'pedantic')
    at <anonymous>
```

Minimal repro — three lines:

```ts
import { marked } from "marked";
const out = marked.parse("# hi") as string;
console.log(out.length);
```

| runtime | result |
|---|---|
| `node --experimental-strip-types` | `12` |
| scriptc 0.0.22 `--dynamic` (QuickJS island) | correct on the full workload |
| **perry 0.5.1389** (`f3e14a61e`) | **TypeError** |

Reproduces both with `perry.compilePackages: ["marked"]` and without it — Perry
compiles the package natively on either path, so there is no working fallback.

## Repro

```bash
cd gc-handoff/apps/mdapp     # marked is vendored under ./node_modules
export PERRY_RUNTIME_DIR=<repo>/target/release
<repo>/target/release/perry mdmin.ts -o p_mdmin
./p_mdmin                    # TypeError
node --experimental-strip-types mdmin.ts   # 12
```

`mdapp.ts` in the same directory is the fuller workload (300 generated markdown
documents through `marked.parse`, counting `<h2>` tags): Node and scriptc
`--dynamic` both print `612040 1800`; Perry throws the same TypeError.

## Where to look

`pedantic` is a `marked` option flag. In `node_modules/marked/lib/marked.cjs` it
appears both in the defaults object (`pedantic: false`, ~line 63) and as a key in
the rules tables (`pedantic: blockPedantic` / `pedantic: inlinePedantic`, ~lines
285/291). The message says the *container* is `undefined` at the read, so the
suspect is a module-level object literal or rules table that did not initialize —
i.e. module init ordering, or an object-literal construction form that lowers
incorrectly.

Possibly related to the known object-literal computed-key / cross-module spread
issue (#809), but that is a guess — the container has not been identified.

Note the compile also reports:

```
perry: module IR is 11.9 MB (> 6.0 MB), 347 functions (~35 KB/fn); compiling at
-Os instead of -O3 ... (#4880)
```

so this module takes the oversized-IR path; worth checking whether the failure is
sensitive to that (try `PERRY_LL_SIZE_OPT` / `PERRY_LL_O0_THRESHOLD_BYTES`).

## Why it matters

This is the axis where Perry's architecture is supposed to win outright: scriptc
must run npm dependencies inside an embedded QuickJS engine (it refuses statically
with `SC2013: values from the 'marked' package run in the embedded dynamic
engine`), while Perry compiles the same package to native code. Today scriptb
produces the right answer through its interpreter and Perry produces a TypeError,
so the comparison cannot even be run.

`marked` is a pure-JS, dependency-free, widely used package — a reasonable
regression target for native npm compilation.

proggeramlug:
Fixes #7689.

## What broke

`marked.parse("# hi")` threw `TypeError: Cannot read properties of undefined (reading 'pedantic')` on every parse, both with and without `perry.compilePackages`.

marked's `parseMarkdown` extracts a static method into a variable and calls it unbound:

```js
const lexer2 = blockType ? _Lexer.lex : _Lexer.lexInline;
let tokens = lexer2(src, opt);
```

`Lexer` declares **both** `static lex(src, options)` and an instance method `lex(src)` (`Parser` has the same `parse` collision). Reading `C.lex` off the constructor routes through `js_class_method_bind`, whose #446 method-identity canonicalization resolves the name against the **instance vtable**: `class_id_from_method_receiver` treats an INT32-tagged constructor ref exactly like an instance receiver, and `method_owner_class_id` only consults instance methods. The extracted value was therefore the *instance* `lex`; invoked bare, its `this.options` read produced `undefined`, and `blockTokens`' first access (`this.options.pedantic`) threw.

In JS, `C.m` never exposes prototype methods (they live on `C.prototype`) — the same semantics the NestJS fix already established for the read path in `get_field_by_name.rs` ("Instance (prototype) methods must only resolve when reading off the prototype ref").

## Fix

In `js_class_method_bind`, skip the instance-vtable canonicalization when the receiver is a **constructor** class ref (`class_ref_id` matches, `class_prototype_ref_id` does not). The read then falls through to `build_bound_method_closure`, whose call-time dispatch (`js_native_call_method`'s `0x7FFE` arm) already resolves statics-first for constructor refs — the exact path that made the same extraction work when no name collision existed. Prototype refs (`C.prototype.m`) keep the canonical instance-method path unchanged.

## Validation

- New runtime unit test `constructor_ref_method_value_resolves_static_over_instance_method` registers a class with both a static and an instance `lex` and asserts the extracted constructor-ref value dispatches the static while the prototype-ref value still dispatches the instance method. Verified it **fails without the fix** (dispatches the instance method) and passes with it.
- New gap test `test_gap_static_method_value_name_collision.ts` covers the marked shape end-to-end (class expression + declaration, ternary + plain extraction, direct calls, prototype read); byte-identical to `node --experimental-strip-types` 26.5.1. It passes, so `gap_snapshot.json` needs no entry.
- The issue's repro: `mdmin.ts` now prints `12` (== Node), and the issue's `mdapp.ts` at 10 documents prints `20105 60` (== Node).
- Full `perry-runtime --lib` suite: 1931 passed, 0 failed.
- Full gap suite run locally; every reported divergence was A/B'd against a baseline build of the merge-base with only this fix reverted: outputs are identical (or the same panic/timeout) on both arms — all are host-environment artifacts (socket binds under the sandbox, the known zlib link flake, npm-import tests whose oracle behaves differently outside CI), none attributable to this change.

## Follow-up (separate issue)

The full 300-document `mdapp.ts` workload is severely superlinear under Perry: 10 docs run in 0.16 s, but 100 docs did not finish within 5 minutes (Node: ~1 s for 300). That is a scaling defect independent of this correctness fix — I'll file it separately so this comparison workload can actually be run to completion.

No version bump per the contribution flow; maintainer bumps at merge.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed static method extraction when a class has an instance method with the same name.
  - Constructor references now correctly invoke static methods, while prototype references continue to invoke instance methods.
  - Prevented accidental calls to unconstructed instance methods in this scenario.

- **Tests**
  - Added regression coverage for direct and indirect method extraction, class expressions, class declarations, static calls, and prototype access.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1400
+**Current Version:** 0.5.1401
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1400"
+version = "0.5.1401"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1400"
+version = "0.5.1401"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1400"
+version = "0.5.1401"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1400"
+version = "0.5.1401"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1400"
+version = "0.5.1401"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1400"
+version = "0.5.1401"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1400"
+version = "0.5.1401"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1400"
+version = "0.5.1401"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1400"
+version = "0.5.1401"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1400"
+version = "0.5.1401"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1400"
+version = "0.5.1401"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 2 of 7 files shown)


proggeramlug:
## Audit — merging as v0.5.1401

**Verified independently** against node 26.5.1, on a class with the exact `static lex` / instance `lex` collision marked has:

| | output |
|---|---|
| node | `static:hi:opt` / `static:yo:opt` / `instance:zz:false` |
| this PR | byte-identical |

Including the ternary form (`blockType ? _Lexer.lex : _Lexer.lexInline`) that is what marked actually writes.

**The root cause is the interesting part**: #446's method-identity canonicalization resolves the name against the **instance vtable**, because `class_id_from_method_receiver` treats an INT32-tagged *constructor ref* exactly like an instance receiver, and `method_owner_class_id` only consults instance methods. So the extracted value was the instance `lex`, and invoked bare its `this.options` read produced `undefined` — surfacing three frames later as `Cannot read properties of undefined (reading 'pedantic')`, which is why this reads as a marked bug rather than a dispatch bug.

A static and an instance method sharing a name is unusual enough to have gone unnoticed and common enough that marked does it **twice** (`Lexer.lex`, `Parser.parse`).

**Gates:** 24/24 via the full `lint` extraction, `perry-runtime --lib` 1941 passed. The gap test needs no registry entry — `check_test_registration` gates the four named suites (gc-repsel-corpus, feature-matrix-probes, compiler-output-workloads, rust-test-modules) and a plain parity test is picked up by glob; I checked rather than assumed, since a dark test is how #7612 slipped.

Issue #7259, opened 2026-08-02
lint: addr_class_inventory fails on main — 2 real ratchet regressions plus a stale allowlist substring


proggeramlug:
Step 7 of the `lint` job, `scripts/addr_class_inventory.py`, exits 1 on `main` (cdb393434). It is currently masked: the job stops at step 3 (#7257), so this has not been visible in CI.

Confirmed unrelated to #7256 (the file-size split) — that PR's finding set is byte-identical to `main`'s.

There are **two independent problems**.

## 1. Two genuine ratchet regressions

```
  [handle-floor] crates/perry-runtime/src/child_process/value_util.rs: 3 site(s), baseline allows 1
      line 103: if ptr.is_null() || (ptr as usize) < 0x1000 {
      line 222: if opts_ptr > 0x10000 {
      line 225: if args_ptr <= 0x10000 {
  [handle-floor] crates/perry-runtime/src/fs/dirent.rs: 3 site(s), baseline allows 2
      line 132: if raw_ptr < 0x1000 {
      line 184: if raw_ptr < 0x1000 {
      line 217: if refreshed_ptr < 0x1000 {
```

Three new hand-rolled address floors landed without going through
`crates/perry-runtime/src/value/addr_class.rs`. Per the gate's own rationale these should use `is_handle_band` / `is_above_handle_band` / `try_read_gc_header`: a bare `< 0x1000` floor does not reject the fetch/zlib/proxy handle bands, and dereferencing a handle segfaults on Linux while macOS hides it (#1843, #4004, #4665, #4800, #6271).

This is a real correctness fix, not bookkeeping — it should not be silently ratcheted up.

## 2. A stale allowlist entry that no longer matches its code

```
  crates/perry-runtime/src/object/native_module/constants.rs: [band-literal] ("O_SYMLINK", (0x200000) as f64),
```

`scripts/addr_class_allowlist.txt` has an entry keyed on both a path prefix and a **line substring**:

```
crates/perry-runtime/src/object/native_module/constants.rs | "O_SYMLINK" => Some(0x200000), | fs.constants O_SYMLINK flag value; unrelated to the handle bands (split of native_module.rs)
```

`constants.rs` contains O_SYMLINK in two syntaxes, and the entry only matches one of them:

- `"O_SYMLINK" => Some(0x200000),` — matches, suppressed
- `("O_SYMLINK", (0x200000) as f64),` — does **not** match, so it fails the gate

[...]

PR #7693, opened 2026-08-09
fix(gc): root fs.readdir's options object across the withFileTypes key allocation (#7274)

### LINKED ISSUE #7274: fs.readdir: options_with_file_types dereferences an unrooted object pointer across an allocation
`crates/perry-runtime/src/fs/dirent.rs`'s `options_with_file_types` decodes a raw `ObjectHeader` pointer out of a NaN-boxed `f64`, then **allocates**, then dereferences the pointer it computed before the allocation:

```rust
pub(crate) unsafe fn options_with_file_types(options_value: f64) -> bool {
    let bits = options_value.to_bits();
    let value = crate::value::JSValue::from_bits(bits);
    let raw_ptr = /* decode from options_value */;
    if crate::value::addr_class::is_handle_band(raw_ptr) {
        return false;
    }
    let obj_ptr = raw_ptr as *const crate::object::ObjectHeader;
    let key = crate::string::js_string_from_bytes(b"withFileTypes".as_ptr(), 13); // <-- ALLOCATES
    let val = crate::object::js_object_get_field_by_name(obj_ptr, key);           // <-- stale obj_ptr
    crate::value::js_is_truthy(f64::from_bits(val.bits())) != 0
}
```

`js_string_from_bytes` can trigger a collection. If that cycle evacuates the options object, `obj_ptr` names a from-space address that has already been swept, and `js_object_get_field_by_name` reads an unrelated live object — or garbage. `options_value` is a plain Rust `f64` local, not a GC root, so it is stale by the same argument; re-deriving the pointer after the allocation would not help on its own.

## The file already knows how to do this

`options_field_value`, 30 lines below in the same file, has the **same signature** and handles it correctly:

```rust
pub(crate) unsafe fn options_field_value(
    options_value: f64,
    field: &[u8],
) -> Option<crate::value::JSValue> {
    let scope = crate::gc::RuntimeHandleScope::new();
    let options_handle = scope.root_nanbox_f64(options_value);
    let bits = options_handle.get_nanbox_f64().to_bits();
    ...
```

and it re-reads through the handle after its own allocating `js_string_from_bytes` call:

```rust
    let key = crate::string::js_string_from_bytes(field.as_ptr(), field.len() as u32);
    let refreshed_bits = options_handle.get_nanbox_f64().to_bits();
    let refreshed_value = crate::value::JSValue::from_bits(refreshed_bits);
    let refreshed_ptr = /* re-decode */;
```

So the fix is self-contained — no signature change, no caller churn: give `options_with_file_types` a `RuntimeHandleScope`, root `options_value`, and re-read the handle after the `js_string_from_bytes` call, exactly as its sibling does.

## Reachability

One caller, `dirent.rs:382`, on the `fs.readdir` / `fs.readdirSync` path:

```rust
let with_file_types = options_with_file_types(options_value);
```

`{ withFileTypes: true }` is an object literal freshly allocated by the caller, so it is a young-generation object — precisely the kind a minor collection triggered by the very next allocation is most likely to move.

## Notes

- Found by CodeRabbit while reviewing #7273, which touches the two lines above this hazard (converting the `< 0x1000` handle floor to `addr_class::is_handle_band`). It is **pre-existing** and unrelated to that change, so it was not folded in: #7273 is a lint-gate fix and a GC-rooting change belongs on its own reviewable/revertable commit.
- Same family as the recent argument-rooting work: #7217, #7252, #7270/#7271.
- No reproducer yet — this is a read of the code, not an observed crash. It would present as a nondeterministic wrong `withFileTypes` result or a segfault under GC pressure, and evacuation makes it far more likely than the non-moving default did.

proggeramlug:
Closes #7274.

## The defect

`crates/perry-runtime/src/fs/dirent.rs::options_with_file_types` decoded a raw
`*const ObjectHeader` out of the NaN-boxed `options` argument, **then** called
`js_string_from_bytes(b"withFileTypes")` — a collection point — and **then**
dereferenced the address it had computed before the collection. `options_value`
is a plain Rust `f64` local, so nothing kept the object alive and nothing
rewrote the pointer.

`{ withFileTypes: true }` is a fresh object literal at the call site, i.e. a
nursery object — precisely the generation an evacuating minor relocates.

`options_field_value`, 40 lines below in the same file, has the same signature
and already did it correctly. The bug is the drift between the two.

## The fix

- The allocation is hoisted **above** the decode, bound together with
  `RuntimeHandle::across_nanbox` so there is no pre-collection address in scope
  to reach for by accident, and `options_value` is rooted in a
  `RuntimeHandleScope`.
- The decode itself — POINTER_TAG / raw-address forms plus the #7259
  `is_handle_band` floor — is factored into one `options_object_ptr` helper.
  It appeared three times in this file; two copies drifting is what this bug
  *was*, so the duplication is removed rather than fixed twice.

No signature change, no caller churn.

## Which configuration this actually bites in

Stated precisely, because the honest answer is narrower than "shipped default
segfaults" and the test file says so out loud:

`gc_check_trigger`'s alloc-point arm engages
`ManualGcScanGuard::force_full_scan(NurseryChurnSlackValve)` — unconditional
since #7682. When that guard **engages**, the conservative native-stack scan
both retains the raw local and makes the copying minor ineligible
(`CopiedMinorFallbackReason::ConservativeStack`), so the pre-fix code survives:
the address neither dies nor moves.

It does not always engage. `force_full_scan` is a no-op when
`CONSERVATIVE_STACK_SCAN_OVERRIDE` is already set, and
`conservative_stack_scan_mode` lets an explicit `PERRY_CONSERVATIVE_STACK_SCAN`
env value beat any pin. So `PERRY_CONSERVATIVE_STACK_SCAN=off` — the arm every
issue in this family reproduces on — removes the valve and the alloc-point minor
evacuates.

So: the masking mechanism is the bounded valve #7148 documents *as* a bounded
valve, not a guarantee. This PR closes the hazard rather than relying on it.

## Witness

Two knob-free unit tests in
`crates/perry-runtime/src/gc/tests/runtime_roots/fs_options_object.rs`. They
drive a **real evacuating minor from inside the function's own key allocation**
and run in exactly the configuration described above (`CopyingNurseryTestGuard`
pins the scan mode to `Auto`, which resolves to `SkipDisabled` — the same
decision `PERRY_CONSERVATIVE_STACK_SCAN=off` produces).

The subject is asserted live, per CLAUDE.md's "a gate must assert its subject
was live":

- the options object is held by **nothing** except the function under test — no
  shadow slot, no registered root;
- a separately rooted **sentinel** allocated in the same nursery must come back
  at a different address, so a cycle that moved nothing cannot certify the file.

That assertion earned its place immediately: the first draft pinned
`force_legacy_gc_pacing()` (scavenge off, polls off), which routes the trigger to
the budgeted stepper — non-moving by construction. The sentinel did not move and
the test failed rather than passing vacuously. Pacing is now left at the shipped
default, deliberately, and the file records why.

## Sabotage

Reverting `options_with_file_types` to the pre-fix decode-then-allocate order:

```
running 2 tests
test ...::readdir_options_object_survives_the_with_file_types_key_allocation ... FAILED
test ...::readdir_options_without_the_field_stays_false_across_the_key_allocation ... ok

panicked at .../fs_options_object.rs:94:5:
options_with_file_types read `withFileTypes` through the address it computed
BEFORE the key allocation (object was at 0x20001300008); the collection at that
allocation moved/reclaimed the object, so the read landed on retired from-space
```

The negative test stays green under sabotage — a stale read yields not-truthy
just as readily as a correct one. It is labelled `NOT A DISCRIMINATOR` in the
source so its passing is never read as evidence about the rooting; it guards the
`None`/false arms of the rewritten decode, which the positive test never reaches.

## Gates run locally

`cargo fmt --all -- --check`, `cargo test -p perry-runtime --lib`,
`check_file_size.sh`, `check_test_registration.py`, `global_sink_isolation.py`,
`addr_class_inventory.py` — all green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `fs.readdir` handling when `withFileTypes` options are processed during garbage collection.
  * Ensured directory entry type detection remains correct even when objects move in memory.
  * Preserved expected behavior for options with or without `withFileTypes`.

* **Tests**
  * Added regression coverage for object movement and liveness during `fs.readdir` processing.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1397
+**Current Version:** 0.5.1398
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1397"
+version = "0.5.1398"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1397"
+version = "0.5.1398"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1397"
+version = "0.5.1398"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1397"
+version = "0.5.1398"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1397"
+version = "0.5.1398"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1397"
+version = "0.5.1398"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1397"
+version = "0.5.1398"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1397"
+version = "0.5.1398"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1397"
+version = "0.5.1398"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1397"
+version = "0.5.1398"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1397"
+version = "0.5.1398"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1397"
+version = "0.5.1398"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1397"
+version = "0.5.1398"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1397"
+version = "0.5.1398"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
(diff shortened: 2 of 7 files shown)


proggeramlug:
## Audit — merging as v0.5.1398

**Sabotage-verified independently**: restoring the pre-fix order (decode → allocate → deref) reddens `readdir_options_object_survives_the_with_file_types_key_allocation`, `error[` 0, test binary reached. Green on restore.

The framing I want on the record is yours: **"the bug is the drift between the two."** `options_field_value` sits 40 lines below in the same file with the same signature and already did it correctly. That is the most common way this class arrives — not someone getting it wrong, but a sibling written later or edited without the neighbour in view.

Hoisting the allocation **above** the decode and binding it with `across_nanbox` is better than rooting-and-re-reading, for the reason the comment gives: there is then *no pre-collection address in scope to reach for by accident*. That is a shape fix rather than a site fix.

`{ withFileTypes: true }` being a fresh literal at the call site — exactly the young object the next allocation's minor relocates — is what makes this reachable rather than theoretical.

**Gates:** 24/24 lint, fmt clean, `perry-runtime --lib` all green.

PR #7694, opened 2026-08-09
fix(gc): root the JSON.stringify shape template's three raw pointers (#7268)

### LINKED ISSUE #7268: json/stringify shape template holds raw obj/keys_array pointers across allocating toJSON recursion (SHAPE_CACHE has no GC root)
Raised by CodeRabbit on #7265 and deliberately not fixed there — it is pre-existing, it is not what #7264 was, and a partial mitigation would be worse than none. Writing it down so the decision is explicit rather than implicit.

## The claim

`json/stringify_shape_template.rs::try_emit_shape_element` derives

```rust
let obj = elem_ptr as *const crate::ObjectHeader;
let fields_ptr = (elem_ptr as *const u8).add(size_of::<ObjectHeader>()) as *const f64;
```

**once**, then loops over the fields calling `stringify_value_depth(...)` for pointer-valued ones. That call can run a user `toJSON`, allocate, and therefore trigger a minor GC — which under the C4b evacuation policy is **moving**. `elem_ptr` is a bare Rust local, invisible to the collector (production resolves the conservative stack scan to `SkipDisabled`). After such a collection every later `fields_ptr.add(f)` read in the same loop is a read of a forwarded/stale address.

`stringify_object_inner` in `json/stringify.rs` already does this correctly: `RuntimeHandleScope` + `root_raw_const_ptr`, with `cur_obj()` re-derived on every access and the keys array re-derived **through** the object header. The template path never got the same treatment.

## Why the naive fix is not enough

Rooting `obj` alone does not close it. The template machinery has two more raw-pointer dependencies with the same exposure:

1. `SHAPE_CACHE` is `Vec<(*mut ArrayHeader, Box<ShapeTemplate>)>` — a **raw `keys_array` pointer used as the shape identity key**, held across the whole traversal and never registered with the GC's root-scanner side tables.
2. `try_emit_shape_element` matches an element by `(*obj).keys_array != template.keys_arr`, and `set_to_json_key_for_template_field` reads key strings **out of** `template.keys_arr`.

If a moving minor can run mid-traversal, a stale cache key either misses (falls back — merely slow) or, worse, collides with a newly-allocated array at the recycled address and matches the wrong shape.

## The actual question

`SHAPE_CACHE`'s doc comment asserts the invariant that makes all of this safe:

> within one top-level stringify call no GC runs over the user object graph (the buffer/result allocations don't move keys arrays), so pointer identity is a stable shape ID

If that invariant holds, there is nothing to fix and the comment should say *why* it holds (what prevents a `toJSON` allocation from triggering a moving minor?). If it does not hold, the fix is not a `RuntimeHandleScope` sprinkle — it is registering the cache with a mutable root scanner (`gc/mod.rs`'s ~55 registered side-table scanners) or suppressing GC/evacuation for the traversal.

So this is a **decide-then-implement** item, not a patch.

## Reproducing / gating

The GC knob kill-policy in CLAUDE.md applies: whatever is decided, the gate must assert its subject was live. A test here needs `PERRY_GC_FORCE_EVACUATE=1` **plus** a `toJSON` that allocates enough to force a real copying minor mid-array, and must assert `copied_objects > 0` — otherwise it proves nothing (see #6942/#6946, where `PERRY_GC_FORCE_EVACUATE` was inert for every `gc()`-driven test).

Suggested shape: a homogeneous array of ≥2 objects sharing a `keys_array` where element 0's fields are primitives (so the template's `primitive_only` path is taken) and a later element carries a pointer field whose `toJSON` allocates heavily. Then verify under `PERRY_GC_VERIFY_EVACUATION=1`.

## Not urgent because

The array template path is only entered for `class_id == 0` plain data objects with no descriptors in use, and `build_shape_prefix_template` bails when the first element could resolve a prototype `toJSON`. The exposure is a *later* element carrying an own/inherited `toJSON` or a nested object that does. Real, but narrow.

proggeramlug:
Closes #7268.

The issue asked for a decision before an implementation. The decision:
**`SHAPE_CACHE`'s invariant is false**, and rooting the cache alone would not
have closed the hazard either.

## What was exposed

`try_emit_shape_element` derived the element header and its inline `fields_ptr`
**once**, then looped over the fields calling `stringify_value_depth` for the
pointer-valued ones. That call can run a user `toJSON` — allocating, reaching a
safepoint, taking an evacuating minor with it. Three holders:

1. **`elem_ptr` / `fields_ptr`** — a bare Rust local. Every later
   `fields_ptr.add(f)` in the *same loop* read a forwarded or swept address.
2. **`ShapeTemplate::keys_arr`** — the raw `ArrayHeader*` used as the shape
   identity *and* dereferenced by `set_to_json_key_for_template_field` to read
   the property-name strings back out. This one is a real deref of retired
   from-space, not merely a wrong identity.
3. **`SHAPE_CACHE`** — whose doc comment asserted the invariant that made (2)
   safe: *"within one top-level stringify call no GC runs over the user object
   graph"*. `toJSON` is user JS. The comment now says why the invariant is
   false instead of asserting it.

## Why all three

Rooting any one leaves the others. In particular, **rooting the cache is not
sufficient**: `stringify_array_depth` builds its template into a plain
`Option<ShapeTemplate>` **Rust local** and holds it across the whole element
loop. No root scanner can ever see that one. So:

- `try_emit_shape_element` roots the element and re-derives `fields_ptr` on
  every access — `stringify_object_inner`'s `cur_obj()` pattern, which the
  template path never got;
- it *also* roots `template.keys_arr` and writes the refreshed address back
  into the template. One mechanism, both holders (cache-owned and local);
- `SHAPE_CACHE` is visited by `json::scan_parse_roots_mut` — already registered
  as `json_parse_mutable_root_scanner` — marking **and** rewriting every
  frame's every key, which is what covers a template sitting idle across a
  nested call.

## Two supporting changes, each removing a way for the fix to be undone

- **`keys_arr` becomes a `Cell`.** `try_emit_shape_element` holds
  `&ShapeTemplate` across the collection. Behind a shared reference the field is
  `noalias` and immutable to LLVM, so the post-call read could legally be folded
  into the pre-call one — the fix would compile away. `Cell`'s `UnsafeCell`
  withdraws exactly that permission.
- **Reentrancy save/restore pushes and pops a FRAME** instead of `mem::take`ing
  the cache into a Rust local. The old shape put the storage outside the
  scanner's reach for the whole of the inner call — and the inner call is the
  `toJSON`, i.e. precisely the code that moves things. `SavedShapeCache` carries
  a `Drop` that pops, and `clear_shape_cache` truncates to frame 0, so a JS
  exception unwinding past the restore cannot leave the stack permanently deep.

The tuple key is gone: `SHAPE_CACHE` was `Vec<(*mut ArrayHeader,
Box<ShapeTemplate>)>` with the same pointer in both halves — one more slot for a
scanner to miss than the design needs. One copy cannot drift from itself.

## Witness

`crates/perry-runtime/src/gc/tests/runtime_roots/json_shape_template.rs`,
knob-free, three tests.

The issue's suggested reproducer needs "a later element carrying a `toJSON` that
allocates heavily". **A `Date` field reaches the same window with no user JS at
all**: `stringify_value_depth`'s `is_date_cell_addr` branch calls
`js_date_to_json`, which allocates a `StringHeader`. That makes it a pure
runtime unit test.

Probe shape `{ when: Date, answer: 42, also: Date }` — and the order is
load-bearing: `when` is the allocation, `answer` is read from `fields_ptr`
*after* it (holder 1), `also` runs `set_to_json_key_for_template_field`, which
dereferences `keys_arr`, *after* it (holder 2). A two-field object exercises
neither.

Per CLAUDE.md's "a gate must assert its subject was live", the end-to-end test
asserts a separately rooted **sentinel relocated**, and the expected bytes come
from a baseline stringify of the same array taken with no collection in flight
(so a date-formatting change cannot silently invalidate the assertion). It also
asserts `"answer":42` is in the baseline at all — without it the probe never
reached the template path.

## Sabotage — both halves

**Element rooting** (hoist `fields_ptr` back above the loop):

```
left:  [{"when":"1970-01-01T00:00:00.000Z","answer":42.000010751275454,"also":"1970-01-01T00:00:00.000Z"},…]
right: [{"when":"1970-01-01T00:00:00.000Z","answer":42,                 "also":"1970-01-02T00:00:00.000Z"},…]
```

Note what that is: **silent data corruption, not a crash.** A garbage double
where `42` belongs, and the wrong date for `also` — read out of retired
from-space. This is what shipped.

**Cache scanner** (delete the `SHAPE_CACHE` walk from `scan_parse_roots_mut`):
`stringify_shape_cache_keys_array_is_marked_and_rewritten` fails on the MARK
assertion.

## Gates run locally

`cargo fmt --all -- --check`, `cargo test -p perry-runtime --lib --no-fail-fast`
(1940 passed / 0 failed), `check_file_size.sh`, `check_test_registration.py`,
`global_sink_isolation.py` — all green.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1393
+**Current Version:** 0.5.1394
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1393"
+version = "0.5.1394"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1393"
+version = "0.5.1394"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1393"
+version = "0.5.1394"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1393"
+version = "0.5.1394"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1393"
+version = "0.5.1394"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1393"
+version = "0.5.1394"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1393"
+version = "0.5.1394"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1393"
+version = "0.5.1394"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1393"
+version = "0.5.1394"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1393"
+version = "0.5.1394"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1393"
+version = "0.5.1394"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1393"
+version = "0.5.1394"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1393"
+version = "0.5.1394"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1393"
+version = "0.5.1394"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1393"
+version = "0.5.1394"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1393"
+version = "0.5.1394"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1393"
+version = "0.5.1394"
 dependencies = [
  "argon2",
  "perry-ffi",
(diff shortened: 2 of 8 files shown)


proggeramlug:
## Audit — merging as v0.5.1394, with one coverage gap recorded

**The fix is right and I verified it end to end.** A 60-row `JSON.stringify` whose `toJSON` allocates 4,000 objects per call, compiled with `PERRY_GC_MOVING_LOOP_POLLS=1`:

| arm | output |
|---|---|
| node 26.5.1 | `3711 [{"answer":42,"when":"2023-11-14T22:13:20.000Z","tag":"row0"` |
| this PR | **byte-identical** |
| + `PERRY_GC_ZEAL=1` | byte-identical, **970,765 objects moved** |
| + `PROTECT_FROMSPACE=1` | byte-identical, exit 0 |

The move count is what makes it non-vacuous — the arm really relocated.

**The framing I want to keep**: the sabotage output here is *silent data corruption*, not a crash — `"answer":42.000010751275454` and a wrong ISO date. A stale `keys_arr` read produces a plausible-looking document. That is the worst failure shape and it is why this one deserved a runtime probe rather than an IR argument.

Three sub-fixes each earn their place: rooting `template.keys_arr` because **not every template is in the cache** (`stringify_array_depth` builds one into a Rust local no scanner sees); making it a `Cell` so LLVM cannot fold the post-call read into the pre-call one; and pushing a **frame** on the reentrancy path instead of moving storage outside the collector's reach.

### The gap: the unit tests do not pin the re-read

I sabotaged 's  closure to return the **pre-call** pointer while keeping the root alive — the exact defect — and all three tests still passed (`error[` 0, `Running unittests` present, so the sabotage was real).

That is not wrong, it is incomplete: `stringify_shape_cache_keys_array_is_marked_and_rewritten` tests the *scanner*, and `shape_template_element_survives_the_date_field_allocation` tests the *element*. Neither asserts that the keys are re-read after the allocation, which is the third of your three sub-fixes. The runtime probe covers it; the suite does not, so a future refactor that folds the re-read back is green here.

Not blocking — the fix is verified and the shape is documented at the call site — but worth a follow-up test that plants a move and asserts the **key strings** read back correctly, since that is the one that would have caught the original.

**Gates:** 24/24 lint, fmt clean, `perry-runtime --lib` **1938 passed**.

PR #7239, opened 2026-08-02
fix(gc): root the ten unrooted runtime-side caches, and the scanner that walked 1 of 3 sibling slots (#7231)

proggeramlug:
The #7231 enumeration, verified and closed. **Ten unrooted runtime-side caches**, plus one scanner that walked one of three sibling slots — the class #7226 established, which is strictly worse than the #7154 stale-register class and which **no static checker can ever find**, because `gc_root_dominance_check.py` reads emitted LLVM IR and a runtime table is not in it.

| | unrooted **register** (#7154) | unrooted **cache** (this) |
|---|---|---|
| goes bad | only if a collection lands in a narrow window | **at collection #0, permanently** |
| reproduces | intermittently — took ten rounds | **10/10** |
| found by | the static IR checker | **nothing static** |

## ★ `CACHED_ENV` — `process.env`

The headline, and it is a hard crash rather than a subtle wrong answer.

`js_process_env_impl` builds `process.env` once with `crate::object::js_object_alloc` — the **nursery** — and stores it in a thread-local `Cell<f64>`. **That cell is the entire reference graph**: `process.env` is not a field of the `process` object, it is a `js_process_env()` call that returns the cache. So the first minor collection swept or evacuated the object, and every later `process.env.X = v` wrote through a dangling pointer into abandoned memory.

The sibling `PROCESS_FINALIZATION_OBJECT` uses the *identical* materialize-once-cache idiom and **is** rooted (`scan_process_finalization_roots_mut`), which is what makes this an omission rather than a design.

**The observable is ENUMERATION.** A direct `process.env.KEY` read lowers to `js_getenv` and asks the OS, so it is correct whatever state the cached object is in. What walks the cached object is `Object.keys(process.env)`, `for…in` and spread — which is exactly how `@next/env` and `dotenv` consume it. A witness that tested the read would be a gate that cannot fail.

```
[gc-fromspace-protect] FAULT: signal 10 at 0x4faa93e012c
  This address is RETIRED FROM-SPACE. The evacuating minor moved or
  freed the object here and the holder kept the pre-collection address.
  last-known object: user_ptr=0x4faa93e0128 obj_type=2 size=416
```

`obj_type=2` is an object; 416 bytes is the header plus the inline env slots.

## The rest of the population

Each was taken to its declaration, its allocator, its dereference and a `grep` of `gc/` for a scanner before being accepted — the campaign's record is roughly 1 diagnosis in 3 naming the wrong cause, and this round refuted several (below).

| holder | why it is real |
[...]

Issue #7231, opened 2026-08-02
GC: enumerate the unrooted runtime-side caches — the population no static checker can see


proggeramlug:
## Summary

#7226 fixed three instances of a bug class nothing static can find: a **runtime-side table that holds a GC pointer and is not a registered root**. This issue is the enumeration nobody had done — a sweep of every `thread_local!` / `static` / `OnceLock` / `Cell` / `RefCell` holder of a GC pointer across all of `crates/perry-runtime/src/`, cross-referenced against the complete `gc_register_mutable_root_scanner*` registry.

The class matters because it is strictly worse than the #7154 stale-register class, and is detected by the opposite signal:

| | unrooted register (#7154) | **unrooted runtime cache** |
|---|---|---|
| goes bad | only if a collection lands in a few-instruction window | **at collection #0, permanently** |
| reproduces | intermittently | **deterministically** |
| found by | `gc_root_dominance_check.py` | **nothing static — the checker reads emitted IR and a runtime table is not in it** |

`gc_root_dominance_check.py` will never report any of these. That is not a gap in the tool; it is outside its subject.

## Independently verified (I read the declaration, the allocator, the dereference, and grepped `gc/` for a scanner)

**1. `process/env_misc.rs:1075` — `CACHED_ENV`, the `process.env` object.** Highest impact in the list.

```rust
thread_local! { static CACHED_ENV: std::cell::Cell<f64> = const { std::cell::Cell::new(0.0) }; }
```
Populated in `js_process_env_impl` with `crate::object::js_object_alloc(0, alloc_limit)` — **nursery**, no `_longlived`. Dereferenced and written through on every `process.env.X = v` (`env_misc.rs:934-944`):
```rust
let obj = crate::value::js_nanbox_get_pointer(cached) as *mut crate::ObjectHeader;
… js_object_set_field_by_name(obj, key, val_f64);
```
`grep -rn "CACHED_ENV" crates/perry-runtime/src/gc/` → **nothing**. `process.env` is touched by nearly every real Node program. The sibling `PROCESS_FINALIZATION_OBJECT` (`process.rs:302`) uses the *same* materialize-once-cache idiom and **is** rooted (`scan_process_finalization_roots_mut`, `gc/mod.rs:614`) — so this is an omission, not a design.

**2. `object/global_this/populate.rs:822` — `ERROR_CONSTRUCTOR_PTR`.**
```rust
[...]

PR #7695, opened 2026-08-09
gc: enumerate the runtime-side GC-pointer holders, with a gate (#7231)

proggeramlug:
Addresses #7231 — the enumeration, with a mechanical gate rather than a list in
a comment.

## Why a gate and not a sweep

A runtime-side table holding a GC pointer **is a root**, and nothing static
could find the class: `scripts/gc_root_dominance_check.py` reads emitted LLVM
IR, and a `thread_local!` is not in it. That is not a gap in that tool, it is
outside its subject. #7226, #7239, #7268 and #7274 were each found by hand, and
each one re-derived the same sweep. The issue's own closing line is the point:
*a list nobody checks is how this class got here.*

## What the gate does

`scripts/gc_runtime_root_holders.py`, in `lint` (already a required context —
hazard 2 of CLAUDE.md's four is the promotion step people forget, so the gate is
placed where that step does not exist).

1. **Enumerate.** Every static-shaped declaration in `perry-runtime` /
   `perry-stdlib` whose type can hold a GC pointer. Rule A: the type names a
   heap header or `JSValue`. Rule B: an integer/`f64` cell that some function in
   its own file both names *and* allocates in — the only way to catch
   `CACHED_ENV: Cell<f64>`, the highest-impact holder in the issue's report.

2. **Compute coverage, don't trust names.** Registered scanners are read from
   every `gc_register_*root_scanner*(...)` call site; a call graph over both
   crates is walked from them to depth 3; a holder counts as covered when its
   identifier appears in a reachable function **defined in the same file as the
   declaration**.

   Both halves are load-bearing. The graph walk is what finds holders a scanner
   reaches through an accessor rather than by name (`cp_live_lock()`,
   `get_closure_props()`, `buffer_props()`). The same-file requirement is what
   stops `REGISTRY`, `SLOTS`, `ROOTS`, `STATES` and `CACHED` — each of which
   names several *different* holders in this tree — from certifying the wrong
   one.

3. **Require a verdict** for the rest, in `scripts/gc_runtime_root_holders.json`.
   An unclassified holder fails. **An entry that no longer matches also fails**,
   which is what makes a fix delete its own exemption (the
   `gc_root_dominance_allowlist.json` rule).

## Current state

`81 holder declarations scanned, 47 reached by a registered scanner, 30
classified in the inventory (122 registered scanners).`

The inventory records 11 `covered_elsewhere` — the gate's known false positives,
each naming the scanner that actually covers it, so they are documented rather
than silenced — 15 `not_a_gc_pointer`, 1 `test_only`, 1 `unverified`, and **two
`open_gap`s the sweep found that nothing tracked**:

- **`json/mod.rs` `PARSE_KEY_RING`** — a hot-key mirror of the *rooted*
  `PARSE_KEY_CACHE`, holding the same `*const StringHeader` values. A move
  rewrites one copy and not the other, so a ring hit hands out a pre-move
  address. Narrow: the keys are `js_string_from_bytes_longlived`, i.e. old-gen,
  so only old-gen defrag can move them — not the copying minor.
- **`perf_hooks.rs` `PERF_ENTRY_KEYS_ARRAY`** — a nursery `keys_array` address
  compared by identity at `perf_hooks.rs:119` and never rewritten. Stale ⇒ a
  silent slow path, or — the sharper half — a match against a newly-allocated
  array recycled into the recycled address. The neighbouring `PERF_ENTRIES` *is*
  covered; this one slot is not.

`SHAPE_CACHE` is also listed, as an `open_gap` pointing at #7268 / #7694. **If
#7694 lands first, delete that one entry** — the gate will say so, which is the
mechanism working.

## Three bugs found while building it — all in the checker, all "green because it matched nothing"

Worth recording, because this is the failure mode the repo keeps paying for:

1. String literals were not stripped, so the brace counting that delimits a
   function body swallowed `scan_raw_json_key_root_mut` — and `RAW_JSON_KEY`,
   which that scanner visits three lines below its own declaration, reported as
   **uncovered**.
2. The registration regex captured only the *first* argument, so
   `gc_register_mutable_root_scanner_named("stdlib:worker_threads:workers",
   scan_worker_roots_mut)` registered nothing and six `worker_threads` holders
   read as uncovered.
3. Rule B keyed on the *file* rather than the *function* and reported 544
   holders, four fifths of them counters and ids — a gate nobody would read.

Each would have produced a plausible-looking inventory full of entries that were
simply wrong.

## How it is built to be able to fail

- `--self-test` plants a covered holder, one reachable **only** through an
  accessor, one uncovered per rule, and a same-named decoy in another file, and
  asserts every classification. Then it asserts the *verdict machinery* can go
  red: an empty inventory must leave every uncovered holder unclassified; an
  entry matching nothing must be stale; an entry for a **covered** holder must
  be stale.
- Floors: fewer than `MIN_HOLDERS` declarations or `MIN_REGISTERED` scanners
  exits **2**, because a broken regex would otherwise report a clean, empty,
  green run — and a broken *registration* regex would make everything read as
  uncovered, which is noise, not a gate.
- **Live sabotage**: planting an unrooted
  `Cell<*mut ObjectHeader>` into `crates/perry-runtime/src/regex.rs` fails the
  real scan, naming the file, line, type and rule. Restored; the scan is green
  again.

## What it CANNOT see — stated in the docstring, because an unstated limit is how a gate gets trusted past its subject

- **`RuntimeState`'s fields.** `state.rs` absorbed roughly a dozen former
  `thread_local!`s; they are struct fields reached through `state()`, so no
  declaration-site scan sees them. All are covered today. A field-count floor
  makes the struct growing at least *loud*.
- **Integer holders whose own file never allocates.**
- **Cross-file scanners** — they read as uncovered and need a
  `covered_elsewhere` entry naming the scanner.
- **Whether a "covered" holder is covered CORRECTLY.** A scanner may visit three
  of a table's four slots — the shape #7239 found in
  `scan_parent_port_event_roots_mut`. This gate bounds the population; it does
  not audit semantics.

CLAUDE.md's runtime-cache bullet gains one clause pointing at the gate. No
version bump.

## Summary by CodeRabbit

* **New Features**
  * Added an automated audit for runtime garbage-collection pointer holders.
  * Added an inventory documenting holder coverage, classifications, and known gaps.
  * Added checks for missing classifications, stale entries, and unexpected runtime-state changes.

* **Bug Fixes**
  * Prevented lint from passing when runtime GC-pointer holders lack verified custody coverage.

* **Documentation**
  * Documented the GC-rooting inventory and enforcement process.
  * Added a changelog entry describing coverage, limitations, and validation checks.


proggeramlug:
## Audit — merging as v0.5.1402, with the stale entry deleted

**The gate predicted its own first failure and was right.** Its `SHAPE_CACHE` entry said, verbatim:

> Fixed in PR #7694 — **DELETE THIS ENTRY when that lands, the gate will tell you.**

#7694 landed, and both the gate and its self-test went red on exactly that entry. I deleted it: **holders 30 → 29**, both green. That is the shrink-only property working on its first real test rather than being asserted.

**80 holder declarations scanned, 47 reached by a registered scanner, 29 classified, 122 registered scanners.** The thing that makes this a gate rather than a list is that coverage is **computed** — registered scanners → call-graph walk → holder named in a reachable function in the same file — instead of declared.

**Sabotage-verified independently**: I planted `static AUDIT_PLANT_HOLDER: Mutex<Option<*mut ArrayHeader>>` into `json/mod.rs` and it exits 1 naming the holder, its line, and the rule that caught it (`[rule A]`). Wired into `lint` (now 26 commands, up from 24).

### The three checker bugs you found in it are the report's best content

All three were **"green because it matched nothing"**, which is the failure this whole campaign is about, appearing inside the tool built to detect it:

- unstripped string literals made brace-counting swallow `scan_raw_json_key_root_mut`, so `RAW_JSON_KEY` — visited three lines below its own declaration — read as uncovered;
- the registration regex read only the first argument, so `..._named("…", scanner)` registered **nothing**;
- rule B keyed on file rather than function and reported 544 holders.

And CodeRabbit's two more are the same shape: bare-name module collisions could certify the wrong module's holder (**real** — `scan_tls_roots_mut` exists in both crates), and the inventory accepted entries with no verdict. A coverage checker that over-reports coverage is worse than none, and five separate ways to do it were found before this shipped.

The two previously-untracked `open_gap`s it surfaced — `PARSE_KEY_RING` (an unrewritten mirror of the rooted `PARSE_KEY_CACHE`) and `PERF_ENTRY_KEYS_ARRAY` — are the payoff.

**Gates:** 26/26, `perry-runtime --lib` **1944 passed**.

Issue #7528, opened 2026-08-06
gc: js_native_call_method reuses one copy of its rooted receiver across ~1200 lines of allocating probes


proggeramlug:
Found while closing #7498. Distinct from it: #7498's two faults are gone from
this reproducer's protected run, and this one remains. Present on both link
modes.

## What

`js_native_call_method` **does** root its receiver — `object_handle` on the
first lines of the function. It then reads it out once:

```rust
let object_handle = root_scope.root_nanbox_f64(object);
...
let object = object_handle.get_nanbox_f64();
let jsval = JSValue::from_bits(object.to_bits());
```

and that local `object` (99 uses) and the derived `jsval` (32 uses) are reused
for the next ~1200 lines, across a dozen probes that allocate. **A value read
out of a root and held in a register across a call is not rooted**
(`docs/src/internals/gc-rooting-invariant.md`) — the root keeps the object
alive and the collector rewrites the *slot*, not the copy.

The measured deref is the closure-magic probe at `native_call_method.rs:1041`:

```rust
let raw_addr = if jsval.is_pointer() {
    crate::value::js_nanbox_get_pointer(object) as usize   // stale copy
} ...
if crate::value::addr_class::is_above_handle_band(raw_addr)
    && crate::closure::is_closure_ptr(raw_addr)            // dereferences it
```

## Repro

On `main` (or on #7527's branch — unchanged by it), with the shipped witness:

```
PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=200 \
  ./out-of test-files/test_gap_gc_iterator_drain_rooting.ts
```

```
[gc-fromspace-protect] FAULT: signal 10
  array_from_spread_value + 2148     (js_closure_call0 — the bound @@iterator thunk)
  js_native_call_method + 5604
  main
```

and under lldb the faulting instruction is the magic read itself:

```
stop reason = EXC_BAD_ACCESS (code=2, address=0x25773a102fc)
frame #0: js_native_call_method + 6060
->  ldr    w8, [x28, #0xc]           ; CLOSURE_MAGIC, 0x434C4F53
```

5/5 under the quarantine. Symbolicate with `PERRY_LINK_MAP=<path>` — the perry
link passes `-Wl,-no_exported_symbols`, so `atos` and `nm` return nothing and
the handler's own backtrace prints bare offsets.

## Observable consequence

[...]

PR #7696, opened 2026-08-09
gc(codegen): root the spread argument-bundle accumulator (#7664)

proggeramlug:
Closes part of #7664 and lowers `gc-root-dominance-statepoints`'s
`--max-unrooted` from **11 to 8**.

## The window

Six arms of `expr/call_spread.rs` bundle every argument — regular and spread,
in source order — into one JS array before dispatching (`console.*` spread, the
`recv.m(...)` and `recv[k](...)` method-apply arms, the namespace REST export,
and the closure-callee path's interleaved and multi-spread arms). All six wrote
the same loop and all six held the half-built array in a bare `i64` SSA register
across it:

```llvm
%acc  = call i64 @js_array_alloc(i32 0)
%box  = call double @perry_fn_…(…)                      ; arbitrary user code
%part = call i64 @js_array_like_to_array(double %box)   ; ALLOCATES
%acc2 = call i64 @js_array_concat(i64 %acc, i64 %part)  ; %acc is stale
```

`--statepoints --moving-only` reports it as `unrooted:alloc` — nothing in the
register's cast chain appears in the `js_array_like_to_array` safepoint's live
bundle, so an evacuating minor there neither marks nor rewrites the array and
`js_array_concat` reads from-space. It is #7453's shape (a fresh heap value in a
raw register across an allocating helper) in the spread lowering rather than the
URL one, reachable from `a.splice(1, 0, ...src)`.

## Why one helper rather than six `operand_protection` answers

**The window is never empty here.** An `Expr::CallSpread` has at least one
spread source by construction, and `js_array_like_to_array` allocates
unconditionally — so `f(...[1, 2])`, every operand an inert literal, is already
the bug, and an operands-only "can anything here collect?" predicate answers
`false` for it. `bundle_args_rooted` disjoins the spread-present test with
`rooting::any_operand_may_collect` for exactly that reason, and
`an_all_inert_spread_bundle_still_roots_its_accumulator` pins it.

The fold itself is `RootedAcc::advance`, so the accumulator never exists as a
register the loop holds across an emission: the re-read is fused to the
`js_array_concat` / `js_array_push_f64` that consumes it, below the allocating
conversion. Ordering is the fix — a re-read *above* `js_array_like_to_array`
would root an already-stale pointer, which is #7192's shape.

## Measured

Native corpus (150 modules, 2538 functions, 32689 statepoints, 19593 live
bundles), **the same binary in both arms**:

| | unrooted | stale |
|---|--:|--:|
| before | **11** | 0 |
| after | **8** | 0 |

The three that disappear are exactly the three `unrooted:alloc` hits in
`test_gap_array_splice_spread::main`; the other eight are byte-identical
(4 `unmasked`, 2 `global`, 2 `capture`). No hit was added.

**Correctness:** every gap test containing a spread call — 71 sources, selected
by `\w\(\s*\.\.\.` / `,\s*\.\.\.` rather than by hand — is **byte-identical to
the pinned Node 26.5.1 oracle**. 0 compile failures, 0 `node_fail`.

**Gates:** all 24 `lint`-job commands green (each command's own exit status
checked, and the extraction asserts it found 24). `cargo fmt --all --check`
clean.

## The tests, and the arm that proves they can fail

`expr/call_spread_rooting_tests.rs`, three cases, **ordering only — never slot
counts**. That is slice 8's lesson restated because this file would otherwise
have fallen into it: on the default build `reserve_shadow_slot` returns a
stack-map index and the pooled-alloca lowering emits a plain `store`/`load`, so
**no `js_gc_temp_root_*` call is emitted at all** and a `temp_root_calls(ir) > 0`
assertion reads zero and passes vacuously. Definition-line ordering is visible in
all three lowerings.

The helper walks **every** `js_array_concat`, not the first — a bundle emits one
per spread source and `require_call_line` would leave a partially-fixed loop
green.

Sabotage arm run and recorded: forcing `protect = false` (which reproduces the
pre-fix IR exactly) takes the suite to `0 passed; 3 failed` — i.e. the tests ran
and failed on the assertion, not on a compile error. Reverted, and the file
`touch`ed afterwards so the restored mtime is not older than the build.

## ★ Four of the remaining eight are checker false positives

The workflow comment previously described them as "4 unmasked, all PHI-MEDIATED
… the reload has to go in the PREDECESSOR, on the edge, which is a different
insertion model". **That is solving a problem that is not there**, and this PR
corrects the comment rather than the code.

`"phi"` is in `TRANSPARENT_OPS` (`gc_root_dominance_check.py:1007`), so taint
flows from a phi *operand* to the phi *result* and the use is located at the
join. But a phi operand is used on **its own incoming edge**. All four hits have
the identical shape — an `&&` join — e.g. `readCtx`:

```llvm
entry.0:   %r2 = <unmask of the receiver>
           br i1 %r4, label %logical.then.1, label %logical.merge.2
logical.merge.2:
           %r87 = phi double [ %r2, %entry.0 ], [ %r86, %pget.recv_merge.7 ]
           ret double %r87
```

Every safepoint is on the `%logical.then.1` path, where the phi selects `%r86`.
On the edge that carries `%r2` nothing collects between its definition and the
join. Verified register-by-register on all four (`readCtx`, `__closure_5`,
`Readable`, `__obj_method_toLocaleString_3`).

So the real residual is **4**, and the comment now says so and says what each is.
It also records why the capture population is not reloadable the way a
string-handle global is: the recipe would have to re-derive the closure pointer
from `%this_closure`, an `i64` **parameter** RS4GC does not relocate, so the
re-read would address the pre-move closure.

An edge-sensitive phi rule is deliberately **not** in this PR: it lowers a
reported count, so it must arrive with a sabotage arm proving it still reports a
phi operand that *is* live across a safepoint on its own edge, and that belongs
with the checker change rather than riding along on a codegen fix.

## Not fixed here, deliberately

Two arms lower `recv_box` (and `key_box`) *before* the bundle and use them
after — a second, distinct window on an *operand* rather than the accumulator.
Closing it needs the receiver and the array in one `RootedGroup` scope so the
release post-dominates the dispatch; that is #7640's population and a different
acceptance test. Both sites now carry a comment saying so rather than being
half-fixed silently. The same applies to `cb_box` and the regular-argument stack
buffer on the closure-callee path (#7210 group 2).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when handling spread arguments across calls, including interleaved and multiple spread values.
  * Prevented spread argument accumulators from becoming invalid during memory allocations.
  * Preserved existing argument ordering and call behavior.

* **Tests**
  * Added coverage validating safe spread argument handling across several call patterns.

* **Documentation**
  * Documented remaining native statepoint checker findings and clarified known false positives.

### DIFF
--- a/.github/workflows/gc-root-dominance.yml
+++ b/.github/workflows/gc-root-dominance.yml
@@ -548,8 +548,53 @@ jobs:
           #      residual above, newly visible because `js_new_function_construct`
           #      is now classified as a mover.
           #
+          # ── 8 here. `expr/call_spread.rs` roots the argument-bundle
+          # accumulator, which closes all three `unrooted:alloc` hits. Measured
+          # on this corpus, same binary both arms, 11 -> 8 with the other eight
+          # byte-identical and `stale` still 0.
+          #
+          # ★ THE FOUR `unmasked` HITS ARE CHECKER FALSE POSITIVES, and the
+          # description three paragraphs up ("the reload has to go in the
+          # PREDECESSOR, on the edge") is solving a problem that is not there.
+          # `"phi"` is in `TRANSPARENT_OPS`, so taint flows from a phi OPERAND to
+          # the phi RESULT and the use is located at the phi's block — but a phi
+          # operand is used on ITS INCOMING EDGE, not at the join. All four have
+          # the identical shape, e.g. `readCtx`:
+          #
+          #     entry.0:            %r2 = <unmask of the receiver>
+          #                         br i1 %r4, label %then, label %merge
+          #     merge:              %r87 = phi double [ %r2, %entry.0 ],
+          #                                           [ %r86, %pget.recv_merge ]
+          #                         ret double %r87
+          #
+          # The safepoints are all on the `%then` path, where the phi selects
+          # `%r86`. On the edge that carries `%r2` nothing collects between its
+          # definition and the join. Verified register-by-register on all four
+          # (`readCtx`, `__closure_5`, `Readable`, `__obj_method_toLocaleString_3`
+          # -- every one a `logical.merge` join of an `&&`).
+          #
+          # So the residual population is 4, not 8, and it splits:
+          #
+          #   2  global (`@perry_global_*`), test_gap_arraybuffer_transfer::main.
+          #      Real; needs ROOTING, per the note above.
+          #   2  capture. `test_gap_class_expr_dynamic_parent_ctor::__closure_21`
+          #      is real and checked by hand: `%r7 = bitcast %r61` (capture 0,
+          #      the dynamic parent class) is read at the top of the closure and
+          #      passed to `js_new_function_construct` ~60 lines below, across
+          #      `js_object_alloc_class_inline_keys` AND a user constructor, and
+          #      appears in no `gc-live` bundle anywhere in the function. It is
+          #      NOT reloadable the way a strhandle is: the recipe would have to
+          #      re-derive the closure pointer from `%this_closure`, an i64
+          #      PARAMETER that RS4GC does not relocate, so the re-read would
+          #      address the pre-move closure. That population needs the callee's
+          #      own closure pointer to be a tracked root first.
+          #
           # #7664 stays open as this budget's referent: a number with nothing
-          # behind it is the thing CLAUDE.md warns a threshold decays into.
+          # behind it is the thing CLAUDE.md warns a threshold decays into. It
+          # cannot go below 4 without either the two rooting fixes or the
+          # edge-sensitive phi rule, and the phi rule must arrive with a
+          # sabotage arm proving it still reports a phi operand that IS live
+          # across a safepoint on its own edge.
           #
           # `stale` (the object survives and is relocated, but a raw copy of
           # its pre-move address is used below) reads 0 today and is held
@@ -569,7 +614,7 @@ jobs:
             --min-statepoints 15000 \
             --min-live-bundles 8000 \
             --min-relocates 20000 \
-            --max-unrooted 11 \
+            --max-unrooted 8 \
             --max-stale 0 \
             --allowlist scripts/gc_root_dominance_allowlist.json \
             --seeded-violations 40 \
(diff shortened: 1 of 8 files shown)


proggeramlug:
## Audit — merging as v0.5.1396. Budget 11 → 8.

**Sabotage verified independently**: forcing `protect = false` reddens all three tests (`error[` 0, `Running unittests` present), green on restore. That is the right assertion to have — the fix is a computed predicate, so a test that passes with it hard-wired off would be pinning nothing.

**The argument for one helper over six `operand_protection` answers is the strongest part**, and `an_all_inert_spread_bundle_still_roots_its_accumulator` is exactly the test that proves it: `CallSpread` always has ≥1 spread source and `js_array_like_to_array` allocates unconditionally, so `f(...[1, 2])` — every operand inert — **is already the bug**, and an operands-only predicate answers `false`. A per-operand verdict would have been correct-looking and wrong.

150 modules / 32,689 statepoints, same binary both arms: exactly the three `unrooted:alloc` hits gone, the other eight byte-identical, stale still 0. 71/71 spread-containing gap tests byte-identical to node.

## Your correction to my brief is the more valuable half

> The remaining 4 are **not** phi-mediated hazards needing an edge-insertion model — they are **checker false positives**.

I inherited "phi-mediated, needs the reload in the predecessor on the edge" from the issue and passed it on as fact. `"phi"` is in `TRANSPARENT_OPS`, so taint flows operand→result and the use is located at the join — but **a phi operand is used on its own incoming edge**, and every safepoint in those four is on the other edge. Verified register by register on all four.

So the real residual is **4**, not 8, and the workflow comment was wrong too. Declining to ship the edge-sensitive rule in this PR is right for the reason you give — *it lowers a reported count, so it must arrive with its own sabotage arm*. A checker fix that quietly makes numbers smaller is the thing this whole campaign exists to prevent.

The two capture hazards are correctly characterised as **not reloadable**: the recipe would re-derive from `%this_closure`, an `i64` parameter RS4GC does not relocate, so the re-read yields the pre-move pointer. That needs the callee's closure pointer to be a tracked root first — a different change.

**Gates:** 24/24 lint, fmt clean, `perry-codegen --lib` **781**, `perry-runtime --lib` 1938, `native_root_coverage` 14/14.

The rest of your report — #7341's triage (55 → 14 measured, bucket (b) **empty**), #7164's real defect (a populated pointer mask over a zero-length traced payload, because `layout_note_slot` flips to `SIDE_MASK` without widening `field_count`), #7528's acceptance criterion having gone vacuous under #7687, and #7210's per-site re-adjudication — I am picking up separately. #7164 in particular is a one-line fix for a real memory-safety hole and should not sit.

Issue #7255, opened 2026-08-02
gc-matrix: four of the six PR-gating arms run ZERO copying minors corpus-wide (default/verify_evac/cons_scan_off at copy-minor 0/49; header still claims 12/22)


proggeramlug:
`scripts/gc_repsel_matrix.sh`'s `PR_ARMS` subset is what `gc-stress` runs on a pull request. **Four of its six arms run zero copying minors across the entire corpus**, so on a PR they can only ever produce `UNVER`.

Measured at `origin/main` = `64c1f56fb`, `--arms all --pressure 8`, node 26.5.1, 49 corpus files — the run's own liveness summary:

```
default                  requires=scavenge  collected 24/49   moved-objects 0/49   copy-minor  0/49
verify_evac              requires=scavenge  collected 24/49   moved-objects 0/49   copy-minor  0/49
cons_scan_off            requires=scavenge  collected 24/49   moved-objects 0/49   copy-minor  0/49
cons_scan_off_force      requires=scavenge  collected 24/49   moved-objects 0/49   copy-minor  0/49
shipped_default          requires=none      collected  3/49   moved-objects 0/49   copy-minor  0/49
evac_minor               requires=move      collected 48/49   moved-objects 48/49  copy-minor 48/49
force_verify             requires=move      collected 48/49   moved-objects 48/49  copy-minor 48/49
```

`PR_ARMS = default,evac_minor,verify_evac,force_verify,cons_scan_off,shipped_default`.

## Why this matters more than a stale comment

The script's header documents `default` as the arm that closed the per-PR hole, in the strongest terms it has:

> ***THIS SUBSET CAN NOW REPRODUCE THE RELOCATING-MINOR DEFECT CLASS (#6993).*** Until #7024 it could not, and that was the hole … So the whole "raw reference held across a relocating collection" class — #6951, #6972, #6982, #6991, #6992 — was invisible per PR and could only go red after merge, on push.

and quotes the measurement that justified it:

```
arm            copy-minor before #7024   after
default              0/22                12/22
```

Today it is **0/49**. The hole the header declares closed is open again, and the header is the reason nobody re-derived it: it reads as a settled fact with numbers attached.

Two further observations from the same measurement:

[...]

Issue #1843, opened 2026-05-26
node:zlib: stream objects missing .on/.pipe/.write (Transform interface) + brotli fns + Z_* constants (43 radar runtime-fails)


proggeramlug:
## Summary

`node:zlib` is measurable now (#1778 linked the feature; compile-fail 50→3) and shows **43 runtime-fails**. Clustered from the #800 radar (perry 0.5.1029):

## Clusters (by impact)

1. **zlib stream objects aren't proper streams (~15 tests)** — `Gzip`/`Deflate`/`Brotli*` instances are missing the Transform-stream interface:
   - `undefined.on` (5), `on is not a function` (2) — e.g. `test-zlib-close-after-error.js`, `test-zlib-object-write.js`
   - `undefined.pipe` (4), `pipe is not a function` (2) — e.g. `test-zlib-flush-drain-longblock.js`
   - `write is not a function` (2) — e.g. `test-zlib-brotli-flush.js`
   So `zlib.createGzip().pipe(...)` / `.on('data')` / `.write()` don't work — the returned object isn't wired as a stream.
2. **Unimplemented zlib functions (13 `value is not a function`)** — brotli-heavy, e.g. `test-zlib-brotli-16GB.js`, `test-zlib-brotli-from-brotli.js`.
3. **Z_* constants** — `undefined.Z_OK` (`test-zlib-const.js`).
4. Behavioral diffs (2), `undefined.length` (2).

## Priority

The stream-interface gap (cluster 1) is the highest-leverage — it's the foundation most zlib usage builds on, and likely shares a root with the broader stream work. Brotli (cluster 2) is a distinct implementation gap.

Context: full `--auto-optimize` #800 sweep, perry 0.5.1029. zlib became measurable via #1778.

PR #7697, opened 2026-08-09
fix(gc): a Symbol's description no longer lives in an untraced payload slot (#7246)

### LINKED ISSUE #7246: GC: a Symbol's description pointer is never traced — GC_TYPE_STRING is a pointer-free Leaf
## Summary

A `Symbol`'s `description` is a `*mut StringHeader` stored inside `SymbolHeader`, and **the collector never traces or rewrites it**. So a symbol that is itself perfectly rooted — shadow slot, side table, everything — can have its description reaped out from under it, and `String(sym)` / `sym.description` / `console.log(sym)` then read recycled memory.

`alloc_symbol` states both halves as a design choice and as a known gap:

```rust
// crates/perry-runtime/src/symbol.rs:370
pub(crate) unsafe fn alloc_symbol(description: *mut StringHeader, registered: bool)
    -> *mut SymbolHeader {
    // Allocate via gc_malloc as a leaf (GC_TYPE_STRING treats payload as
    // opaque, which is what we want — the GC won't try to scan internal
    // pointers). The description pointer is kept alive through the
    // SYMBOL_REGISTRY (for registered symbols) or not at all (for fresh
    // symbols — in practice they live for the duration of the program,
    // which is fine for test workloads).
    let raw = crate::gc::gc_malloc(
        std::mem::size_of::<SymbolHeader>(),
        crate::gc::GC_TYPE_STRING,
    );
```

`GC_TYPE_STRING`'s type-info entry (`gc/types.rs`) is `pointer_free: true`, `GcRewriteDescriptorKind::Leaf`, `GcLayoutSlotKind::None`, `GcMoveHookKind::None` — so nothing walks into the payload. That is correct for a *string*, whose payload is bytes. It is wrong for a *symbol*, whose payload's third word is a heap pointer. Symbols and strings share one GC type, so no descriptor can distinguish them today.

`SYMBOL_POINTERS` does not close it either: `scan_symbol_pointer_metadata_roots_mut` (`symbol/gc_roots.rs`) visits the set with `visit_metadata_usize_slot`, which **rewrites a recorded address without marking**. It keeps the registry's own addresses honest; it does not keep anything alive, and it never looks at `(*ptr).description`.

## Why this is being filed now

Found while building #7236's runtime witness — and it is a **different defect from #7236**, which is a codegen classification (`Type::Symbol` treated as an immediate, so a `Symbol` local got no shadow slot). #7236 is fixed in #7243. This one survives that fix, because it is about the pointer *inside* a symbol rather than the pointer *to* it.

Measured on #7243's branch, release, `--arms all --pressure 8`, oracle node 26.5.1, with a witness whose probe symbol carries a description:

| arm | result |
|---|---|
| `evac_minor`, `force_evac`, `force_verify`, `loop_polls`, `rep_*` (11 arms — every arm with `PERRY_GC_INCREMENTAL=0 PERRY_CONSERVATIVE_STACK_SCAN=off`) | **red, 2 of 20 iterations**, deterministic |
| `default`, `verify_evac`, `cons_scan_off`, `cons_scan_off_force`, `gen_gc_off`, `wb_off`, `shipped_default`, … | green |

Per-probe instrumentation isolates it precisely. At base (before #7243) the symbol *object* is freed and its block recycled into another symbol, so the failures are `string 17 desc 17` with `typeof 0`. After #7243 the symbol object survives — and the residual is still `string 2 desc 2`, same probes, `typeof 0`, `read 0 count 0`. The key resolves, it is still a symbol, its description is wrong.

The discriminator is one character: making the probe symbol `Symbol()` instead of `Symbol("k")` — no description, therefore no untraced pointer — takes the same file to **0 FAIL / 21 of 21 byte-exact** on every arm. That is why `test-files/test_gap_gc_symbol_local_rooting.ts` ships with a descriptionless symbol and says so in a header comment: it gates one defect, and this is the other one.

## Reproducer

```ts
function symChurn(n: number): number {
  let k = 0;
  for (let i = 0; i < n; i++) {
    const t = Symbol("t");
    if (typeof t === "symbol") { k++; }
  }
  return k;
}
function run(): number {
  let bad = 0;
  for (let r = 0; r < 20; r++) {
    const s = Symbol("k");            // description: an untraced StringHeader*
    if (symChurn(50000) !== 50000) { bad++; }
    const o: any = {};
    o[s] = r;
    if (o[s] !== r) { bad++; }        // green — the symbol itself is fine
    const keys = Object.getOwnPropertySymbols(o);
    if (keys.length !== 1) { bad++; } // green
    if (String(keys[0]) !== "Symbol(k)") { bad++; }   // RED 2/20
  }
  return bad;
}
console.log("B", run());
```

Compile and run with `PERRY_GC_MOVING_LOOP_POLLS=1`; run additionally with
`PERRY_GC_HEAP_LIMIT=8 PERRY_GC_INCREMENTAL=0 PERRY_CONSERVATIVE_STACK_SCAN=off PERRY_GC_FORCE_EVACUATE=1`.
Node prints `B 0`.

Symbol churn is load-bearing: `copied_minor_malloc_sweep_due` gates the malloc sweep on a `GcTriggerKind::MallocCount` collection or the malloc-object threshold, and object/array churn reaches the *arena* trigger instead.

## Why it has mostly not bitten

The quoted comment's own answer — "in practice they live for the duration of the program" — plus the conservative stack scan, which pins whatever happens to be in registers. Both are properties the moving-GC work keeps eroding, which is the same "closed by accident" argument #7226 recorded for #7213.

## Fix shape (not decided here)

Three candidates, in increasing order of blast radius:

1. **Give symbols their own GC type.** `GC_TYPE_SYMBOL` with a descriptor that traces the one `description` slot. Principled and complete. Touches the GC type table and its verification contract, plus everything that assumes a symbol is `GC_TYPE_STRING` (`is_symbol_pointer`, `heap_snapshot.rs`, `dead_owner.rs`).
2. **Trace the description from the symbol side table.** `scan_symbol_side_table_roots_mut` already enumerates every symbol pointer; visiting `(*ptr).description` as a real (marking, rewriting) root would close it — but only if the visit is conditional on the symbol itself being live, or every symbol ever created retains its description string forever.
3. **Intern descriptions off the GC heap.** `REGISTERED_SYMBOL_DESCRIPTIONS` already does exactly this for well-known and registered symbols, with readers materializing a fresh `StringHeader` on demand. Extending it to fresh symbols removes the pointer entirely — at the cost of retaining a Rust `String` per symbol, which a workload that makes millions of symbols would feel.

Whichever is chosen, the acceptance test is the reproducer above going green on the eleven `PERRY_GC_INCREMENTAL=0 PERRY_CONSERVATIVE_STACK_SCAN=off` arms, at which point `test_gap_gc_symbol_local_rooting.ts` can drop its descriptionless-symbol carve-out and the header comment explaining it.

Refs #7236, #7243, #7235, #7230, #7226.

proggeramlug:
Closes #7246.

## The defect

`SymbolHeader::description` was a `*mut StringHeader` living inside a payload the
collector treats as **opaque bytes**. `alloc_symbol` gc_malloc's the header as
`GC_TYPE_STRING`, whose type info is `pointer_free: true` /
`GcRewriteDescriptorKind::Leaf` / `GcLayoutSlotKind::None` — correct for a
*string*, whose payload really is bytes, and wrong for a *symbol*, whose
payload's third word was a heap pointer. Symbols and strings share one GC type,
so no descriptor could tell them apart.

So a symbol that was itself perfectly rooted — shadow slot, side table,
everything — could have its description reaped or relocated out from under it.
`SYMBOL_POINTERS` did not close it either:
`scan_symbol_pointer_metadata_roots_mut` visits the set with
`visit_metadata_usize_slot`, which rewrites a recorded address **without
marking**, and never looks at `(*ptr).description` at all.

## The decision

The issue listed three candidates. This takes the third — intern off-heap — and
the reason it is cheap is the **key**.

`FRESH_SYMBOL_DESCRIPTIONS` is keyed on `SymbolHeader::id`, a monotonic `u64`
that an evacuation copies verbatim. Not on the symbol's address. Therefore:

- **no rekey pass**, unlike every other symbol-address-keyed side table;
- **no root scanner and no budgeted step twin** — it holds no GC pointer at all,
  so there is nothing for a scanner to visit and nothing for a `(FULL, STEP)`
  pair to drift on (#7239 found the one real drift in exactly that shape);
- **no window**: `alloc_symbol` copies the text *before* it allocates, so a
  description pointer is never live-but-untraced. #7341's `RuntimeHandleScope` +
  `across_mut` here goes with it — that made the *stored* pointer correct across
  `gc_malloc`, and there is no longer a stored pointer.

Why not the other two:

- **`GC_TYPE_SYMBOL`** is the principled fix, and it touches **190**
  `GC_TYPE_STRING` sites across `perry-runtime` and `perry-codegen`, plus the GC
  type table's verification contract, `is_symbol_pointer`, `heap_snapshot.rs`
  and `dead_owner.rs`. Worth doing one day; not worth doing hurriedly.
- **Tracing the description from the side table** needs the visit to be
  conditional on the symbol being live, which is a weak-table ordering problem —
  and `SYMBOL_POINTERS` can hold an entry for a symbol that is dead but not yet
  pruned, so the scanner would dereference freed memory to reach the
  description.

The retention cost the issue named as this option's price is paid down:
`prune_dead_symbol_pointers` prunes the descriptions on the same liveness
verdict that prunes the pointers.

## Blast radius is small, and here is why

`alloc_symbol` has exactly **two** callers, both fresh symbols
(`js_symbol_new_empty`, `js_symbol_new`). Registered (`Symbol.for`) and
well-known symbols are `Box::leak`'d and already used the process-global
`REGISTERED_SYMBOL_DESCRIPTIONS` — untouched. Ids are globally monotonic, so the
thread-local and process-global maps never collide. The four readers
(`js_symbol_key_for`, `js_symbol_description`, `js_symbol_to_string`,
`infer_symbol_function_name`) now go through one `symbol_description_text`
helper instead of open-coding the `registered_symbol_description(..).or_else(..)`
chain — there were four of them, and a fifth that forgot the fallback is exactly
how a description goes silently missing.

## Witness — the issue's own reproducer, A/B across the runtime rebuild

Same compiler, `PERRY_RUNTIME_DIR` pinned, `PERRY_NO_AUTO_OPTIMIZE=1`, no
compile-time GC env. Run arm as the issue specifies:

```
PERRY_GC_HEAP_LIMIT=8 PERRY_GC_INCREMENTAL=0 \
PERRY_CONSERVATIVE_STACK_SCAN=off PERRY_GC_FORCE_EVACUATE=1

before:  B 1    5/5 deterministic
after:   B 0   10/10          (node 26.5.1: B 0)
```

**Movement confirmed live on the after-run** (#7255): 87 copying minors, with
`copied_objects=6008` and `copied_objects=4743` on the two that mattered. A run
that moved nothing would prove nothing. The shipped default is also `B 0`.

Note the count moved from the `B 2` the issue recorded to `B 1` — same defect,
one probe now failing rather than two, which is what you would expect from the
allocation-shape churn since. It is still 5/5 deterministic, which is the tell
for this class: an unrooted *cache* goes bad at collection #0 and stays bad.

## Unit tests — knob-free, and each one able to fail

`crates/perry-runtime/src/gc/tests/runtime_roots/symbol_description.rs`:

1. **Structural** — `SymbolHeader::description` must stay null. This is the
   assertion a future change trips first, whether or not a behavioural test
   happens to catch it that day.
2. **Behavioural** — the description survives a collection that reclaims the
   string it came from. The recycling loop afterwards (512 `'Z'`-filled strings)
   is load-bearing: without it a stale read can find the old bytes intact and
   the test passes for the wrong reason. It also asserts the payload pointer was
   null all along, so it cannot silently measure the old representation.
3. **The prune** — descriptions of dead symbols go with their pointers.
   Otherwise interning off-heap trades a use-after-free for an unbounded leak,
   and the doc comment claiming otherwise would be the only evidence.

**Sabotage**: restoring `(*ptr).description = description` fails all three.

## Not done here

`test-files/test_gap_gc_symbol_local_rooting.ts` still ships its
descriptionless-`Symbol()` carve-out and the header comment explaining it. The
issue says that can be dropped once this lands; it is a corpus-registered gap
file, so changing it wants a parity run rather than a code review, and it is
left as a follow-up.

## Gates run locally

`cargo fmt --all -- --check`, `cargo test -p perry-runtime --lib --no-fail-fast`
(1938 passed / 0 failed), `check_file_size.sh`, `check_test_registration.py`,
`global_sink_isolation.py` — all green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed symbol descriptions being lost or corrupted after garbage collection and memory reuse.
  * Symbol descriptions now remain available after their source strings are collected.
  * Improved handling of non-UTF-8 symbol description text.
  * Cleaned up descriptions for symbols that are no longer in use.

* **Documentation**
  * Added changelog details and runtime coverage for symbol description behavior.

* **Chores**
  * Updated the application version to 0.5.1400.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1399
+**Current Version:** 0.5.1400
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1399"
+version = "0.5.1400"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "perry-ffi",
  "scraper",
@@ -5813,7 +5813,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-commander"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5821,7 +5821,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cron"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "chrono",
  "cron",
@@ -5831,7 +5831,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dayjs"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "chrono",
  "perry-ffi",
@@ -5839,7 +5839,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-decimal"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "perry-ffi",
  "rust_decimal",
@@ -5847,7 +5847,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dotenv"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "perry-ffi",
  "serde_json",
@@ -5855,7 +5855,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ethers"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "perry-ffi",
  "rand 0.10.1",
@@ -5863,7 +5863,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-events"
-version = "0.5.1399"
+version = "0.5.1400"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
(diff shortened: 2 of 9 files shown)


proggeramlug:
**Re-verified after the WTF-8 refinement** (`a29187ae0`), so the A/B in the description is not resting on the pre-refinement build.

The interned description became `Arc<[u8]>` rather than `Arc<str>`, because `str_from_header` UTF-8-validates and returns `None` on failure — a description built from a JS string with a lone surrogate is WTF-8, and interning through `String` would have turned it into `undefined`. That is a behaviour change I nearly smuggled in on a GC fix; it is closed, and the residual (the rebuilt `StringHeader` does not carry `STRING_FLAG_HAS_LONE_SURROGATES`, which is not recoverable from the payload) is stated in the code rather than hidden.

Rebuilt `-p perry -p perry-runtime-static -p perry-stdlib-static`, `.a` mtime confirmed moved, `PERRY_RUNTIME_DIR` pinned, `PERRY_NO_AUTO_OPTIMIZE=1`:

```
PERRY_GC_HEAP_LIMIT=8 PERRY_GC_INCREMENTAL=0 \
PERRY_CONSERVATIVE_STACK_SCAN=off PERRY_GC_FORCE_EVACUATE=1

B 0   10/10        (node 26.5.1: B 0)
```

`cargo test -p perry-runtime --lib --no-fail-fast` — 1938 passed, 0 failed, cargo exit 0, 0 compile errors, `Running unittests` observed.

proggeramlug:
## Audit — merging as v0.5.1400, with one fix applied

**The defect is the sharpest kind: a type that is right for one thing and wrong for another sharing it.** `alloc_symbol` gc_mallocs its header as `GC_TYPE_STRING`, whose type info is `pointer_free: true` / `Leaf` / `SlotKind::None` — correct for a string, whose payload really is bytes, and wrong for a symbol, whose third payload word was a heap pointer. **Symbols and strings share one GC type, so no descriptor could tell them apart.** A perfectly rooted symbol could still have its description reaped.

And `SYMBOL_POINTERS` did not close it, for a reason worth keeping: `visit_metadata_usize_slot` **rewrites a recorded address without marking**, and never touches `(*ptr).description` at all. A scanner that rewrites but does not mark looks like coverage in every audit that counts registrations.

**Removing the pointer rather than tracing it is the right of the three candidates**, and the key is why it is cheap: `FRESH_SYMBOL_DESCRIPTIONS` is keyed on `SymbolHeader::id`, a monotonic `u64` an evacuation copies **verbatim** — not on the address. So there is no rekey, no scanner, and no budgeted-step twin.

**Verified independently**, 200 symbols across 200k allocations of churn, compiled with `PERRY_GC_MOVING_LOOP_POLLS=1`:

| | output |
|---|---|
| node 26.5.1 | `B 0 desc-7 Symbol(desc-3) lone 3` |
| this PR | byte-identical |
| + `PERRY_GC_ZEAL=1` | byte-identical |

`lone 3` is the case that matters most: storing as `Arc<[u8]>` rather than `str` keeps a lone-surrogate description intact. You caught yourself about to turn that into `undefined`; `str_from_header` would have validated it away, and no test in the suite would have noticed.

### One fix I applied

`addr_class_inventory` went red: the two new dereference guards used a bare `(ptr as usize) < 0x1000` floor, taking `symbol.rs` from its ratcheted 3 handle-floor sites to 5. That floor **does not reject the fetch/zlib/proxy handle bands**, and dereferencing one segfaults on Linux while macOS silently hides it (#1843/#4004/#6271). Converted both to `addr_class::is_above_handle_band`, which is the predicate that does. Re-verified: output unchanged, node-identical, gates 24/24.

Worth noting the gate caught this **before** merge because I now run all 24 through a script that exits non-zero and gate the merge on it — I merged twice today with a red gate by printing the failure and merging anyway.

**Gates:** 24/24, `perry-runtime --lib` all green, symbol suite 30/30.

Issue #6921, opened 2026-07-28
codegen: lower_new_impl standalone-ctor exit returns an instance without js_gc_init_typed_shape_layout (blocks the pointer-masked layout-note elision)


proggeramlug:
## Summary

`lower_new_impl` has one exit that returns a freshly allocated class instance **without** emitting `js_gc_init_typed_shape_layout`. Every other `new` path emits it. An instance produced through that exit is left at `GC_LAYOUT_POINTER_FREE` with no `TypedLayoutDescriptor`, which is the one layout state where `layout_note_slot` is *load-bearing for GC correctness* rather than merely a precision hint.

Found while proving the class-field store elision in #6919 (representation-selection Phase 4b.1). It did not cause a bug there — the elision was narrowed instead of being shipped on the unproven premise — but it blocks the larger win, so it is worth closing on its own.

## The exit

`crates/perry-codegen/src/lower_call/new.rs`, the standalone-constructor-symbol branch:

```rust
// new.rs:811
if ctx.class_stack.iter().any(|active| active == class_name)
    || ctor_alias_collision
    || force_ctor_call
{
    ...
    if let Some(ctor_ret) = call_local_constructor_symbol(ctx, class, &obj_box, &lowered_args, caps_absent_from_args) {
        ...
        emit_typed_shape_layout_init(ctx, class_name, &obj_handle);   // new.rs:870  ✅
        ...
        return Ok(final_box);                                          // new.rs:899
    }
    if let Some(prev) = &saved_new_target {
        ctx.block().call(DOUBLE, "js_new_target_set", &[(DOUBLE, prev)]);
    }
    return Ok(obj_box);                                                // new.rs:905  ❌ no init
}
...
emit_typed_shape_layout_init(ctx, class_name, &obj_handle);            // new.rs:1661 ✅ inline path tail
```

The `None` arm of `call_local_constructor_symbol` falls through to `return Ok(obj_box)` and never reaches either `emit_typed_shape_layout_init` call.

## Reachability

`call_local_constructor_symbol` (`new_ctor_args.rs:321`) returns `None` exactly when `ctx.methods` has no `(class.name, "<Class>_constructor")` entry:

```rust
let ctor_name = ctx.methods.get(&(class.name.clone(), ctor_method_name)).cloned()?;
```

Two of the three branch conditions pre-check that the symbol exists, so they cannot reach the `None` arm:

[...]

PR #7698, opened 2026-08-09
perf(class-fields): a conforming pointer store skips the layout note (#5094)

proggeramlug:
## The P3 investigation resolved the ticket's contradiction as "already fixed"

P3 (`gc-handoff/PROMPT-P3-class-canonical-layout.md`) opens with a contradiction
to adjudicate before writing any fix: profiles showing 25–29% of class-heavy
benches in per-object layout side tables, against a `shape_install_shared` path
that is supposed to store one descriptor per shape with no per-object insert.
Two hypotheses were named. **Both are false**, and the ticket's own acceptance
criteria 1 and 2 are already satisfied on `main` — its table was measured at
`a853135aa`, before P1 (#7686).

Counters from an instrumented runtime, `tree.ts`:

```
INIT_CALLS        20447192     NOTE_CALLS          20447156
INIT_KEYS_ZERO           0  ←  NOTE_CONFORMS       20447154
INIT_MEMO_HIT     20447192     NOTE_DOWNGRADE             0  ←
INIT_SHARED_OK           1     FORGET_CALLS        64000001
INIT_SHARED_FAIL         0     FORGET_SLOW                1  ←
residency: slot_masks=0  typed_layouts=0  per_object_flag_armed=false
```

* **H1 — the keys_array is attached after the layout declaration.** False.
  `INIT_KEYS_ZERO = 0`; every instance reaches `shape_install_shared` keyed.
* **H2 — `a.peer = b` demotes the descriptor.** False. `NOTE_DOWNGRADE = 0`.
* **Criterion 2 (residency ~0) already holds.** Both per-object maps are empty,
  the emptiness flag never arms, and `layout_forget_object` took its slow path
  once in 64 million calls.

## What was actually left, and is fixed here

20,447,154 of those 20,447,156 `js_gc_note_slot_layout` calls return `Conforms`
— a cross-crate call, header decode, TLS touch and `SHAPE_LAYOUTS` hashmap
lookup per pointer store, to re-derive what codegen knew when it emitted the
mask. `class_field_store_needs_layout_note`'s doc comment named this case, why
it was left alone, and the condition that would unblock it:

> **Deliberately NOT elided: a pointer-valued store into a pointer-masked slot.**
> […] Closing that exit (#6921) is the prerequisite for the stronger elision.

#6921 is closed. This takes the elision as a **live header test with the real
note kept on the cold arm** — see the changelog fragment for the two facts that
make the taken arm a proof (the #5093 precheck has already pinned the receiver's
`keys_array` to this class's keys global; the slot is in the pointer mask and
checked absent from the raw-f64 mask, so neither downgrade arm can fire).

## Measured — quiet M1 mini, best-of-3, interleaved, stdout byte-identical

| bench | before | after |
|---|--:|--:|
| `cycles` | 0.33 | **0.29** |
| `tree_wide` | 7.90 | **7.77** |
| `tree` | 5.17 | **5.13** |

Unchanged: `churn` 1.21, `push_cls` 0.89, `churn_alloc` 0.89, `retain` 2.41,
`retain1` 0.04, `retain_wide` 3.38, `retain_wide1` 0.06, `deeplist` 0.03,
`cls_mistyped` 0.02. Peak RSS unchanged.

The profile states it more plainly than the clock: on `cycles_big`,
`layout_note_slot` is the **second-heaviest leaf frame in the base arm (155
samples)** and is **absent from the fix arm entirely**. Wall-clock share is
smaller because these workloads are GC-dominated under the current default
pacing.

## Safety

The intact bit is not made sticky and no downgrade path is touched.

* `cls_mistyped.ts` (a `number`-declared field constructed with a heap string
  per instance — it must demote or the collector never traces those strings)
  still prints `20000 string payload-19999`.
* `PERRY_GC_VERIFY_MARK=1 PERRY_GC_VERIFY_EVACUATION=1` clean over the bench set.
* `PERRY_GC_TRACE` cycle counts and copied bytes identical between arms
  (`churn` 13, `tree` 20).
* `cargo test -p perry-codegen --lib` 781 passed; `-p perry-runtime --lib` 1935
  passed. `cargo fmt --all -- --check` clean.

Three new tests, unit-level so `cargo-test` sees them (#5960), each verified able
to fail: the positive asserts the elision is **reached** (a predicate silently
answering `false` everywhere would otherwise be invisible), and the negative
asserts a slot in *neither* mask (`flag: boolean`) keeps its unconditional note —
sabotaging the predicate to return `true` makes that test fail.

## Note for reviewers

Measuring P3 at all required #7690. On `main` before it merged, `tree` ran
**204 s** against an expected 4.4 s, `churn` 9.1 s with 1304 GC cycles against
105, with `"copying_nursery":{"eligible":false}` on every trace line — #7687 had
landed the immobility guard without the pacing half. `cycles` was nearly
unaffected (0.36 s), which is why that state was easy to miss. All numbers here
are on `main` with #7690 merged.

## Summary by CodeRabbit

* **Performance**
  * Improved garbage-collection bookkeeping for eligible class-field pointer stores by avoiding unnecessary layout notifications.
  * Preserved existing fallback behavior for non-eligible or uncertain cases.

* **Bug Fixes**
  * Strengthened handling of class-field layout metadata, including pointer, raw floating-point, unmasked, and out-of-range slots.
  * Expanded validation across guarded and multi-block store paths.

* **Documentation**
  * Updated the documented release version and added a changelog entry describing the optimization.

* **Tests**
  * Added coverage for layout classification, optimized stores, fallback behavior, and demotion scenarios.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1403
+**Current Version:** 0.5.1404
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1403"
+version = "0.5.1404"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1403"
+version = "0.5.1404"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1403"
+version = "0.5.1404"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1403"
+version = "0.5.1404"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1403"
+version = "0.5.1404"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1403"
+version = "0.5.1404"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1403"
+version = "0.5.1404"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1403"
+version = "0.5.1404"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1403"
+version = "0.5.1404"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1403"
+version = "0.5.1404"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1403"
+version = "0.5.1404"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1403"
+version = "0.5.1404"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1403"
+version = "0.5.1404"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1403"
+version = "0.5.1404"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1403"
+version = "0.5.1404"
 dependencies = [
  "anyhow",
  "clap",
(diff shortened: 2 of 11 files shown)


proggeramlug:
## Audit — merging as v0.5.1404, after fixing two tests it broke

I could not break the elision. I *could* break the branch, and the way it broke is the interesting part.

### Soundness: I tried the five ways this could be wrong

The note is a genuine no-op on the path being skipped — `layout_note_slot` returns without touching any state when the classify closure says `Conforms` (`layout.rs:834-839`). So the question is only whether the header test can be true while the real verdict would have been `Downgrade`.

| attempt | why it fails |
|---|---|
| subclass instance into a superclass-typed field | `layout_declares_pointer_slot` asks "is this slot pointer-masked", not which class. A subclass instance is still a pointer. |
| `undefined`/`null` then a pointer | the mask describes the *slot's* declared shape, not its current value |
| union / `Any` field | `type_is_pointer_bearing` marks `Any`/`Unknown`/`Union` pointer-masked and `type_is_raw_f64_candidate` is `Number`-only, matching what the runtime installs |
| pointer → f64 → pointer | a raw-f64-masked field never reaches this arm. The variant that *is* interesting — INTACT cleared by an earlier store — is handled: the header test is a **live** re-check per store, so a prior downgrade makes the very next store take the cold arm. |
| an IC keyed on the note firing | the only other side effect is `invalidate_representation_change`, and it runs on `Downgrade` only — never on `Conforms`, so it is never wrongly skipped |

The thread I pulled hardest: can the *scanner* trust `INTACT+SIDE_MASK` without finding a descriptor? No — `heap_payload_slot_selection` re-derives the mask at trace time and self-heals to `GC_LAYOUT_UNKNOWN` (scan everything) if both lookups miss. The trace path never depends on bookkeeping a skipped `Conforms` call would have done. **Not a missing root.** The write barrier is untouched.

### What it actually broke, and why neither of us saw it

`class_field_store_pointer_test.rs` — **two failures**, `guarded_store_leaves_every_block_terminated` and `opaque_param_store_guards_all_three_bookkeeping_calls`, both *"@js_write_barrier_slot must live inside the guarded block"*.

The barrier had not moved. Your new `layout_note` conditional splits `gc_bookkeeping` into three blocks, and `gc_bookkeeping_block()` sliced text from the label to the **first `br`** — which is the region only while it is one block:

```
gc_bookkeeping.4:     br i1 %r33, →layout_note.done.7, →layout_note.6
layout_note.6:        call @js_gc_note_slot_layout   → layout_note.done.7
layout_note.done.7:   call @js_write_barrier_slot    → gc_bookkeeping.done.5
```

`layout_note.done.7` is reachable only from `gc_bookkeeping.4`, so the barrier is still dominated by the pointer-bearing guard. Sound; the test just could no longer see it.

A wider text slice would not have worked either — LLVM emits `gc_bookkeeping.done.5` **between** `gc_bookkeeping.4` and `layout_note.6`, so the region isn't textually contiguous. I replaced the slice with a CFG walk from the guard entry, stopping at the join (commit on the branch). That makes the assertion **stronger** than before: a call hoisted onto the not-taken edge or past the join now fails, whereas the old slice only ever proved a call wasn't in the first block. The test's existing negative assertion (`!guarded.contains("store double")`) still passes, which is what proves the region is a proper subset and not the whole IR.

**Why CI couldn't tell you:** per-PR `cargo-test` runs `--lib --bins` — no `tests/*.rs` suite runs on a PR at all, and `e2e-scoped` maps all of `crates/perry-codegen/src/` to three suites, none of them this one. Full `cargo test -p perry-codegen` now shows **6 failures, byte-identical to the set `main` already has** (those 6 are pre-existing and not yours — worth its own issue).

1. No IR test covers the `ptr_shape_proven` call site (`property_set.rs:1089`) — the one arm with no runtime `class_id`/`keys_array` check, resting instead on the compile-time "exactly one `Stmt::Let` whose init is `new C(...)`" provenance rule. The two arms that *are* tested both have runtime guards.
2. `class_field_store_layout_note_is_conforming` (`helpers.rs:509`) and `typed_shape_init.rs:110` are independent copies of the same `class_init_chains`-preferred/name-keyed-fallback lookup. Consistent by construction today, with nothing enforcing it stays that way.

Gates 26/26.

PR #7699, opened 2026-08-09
fix(codegen): the lowering-window cluster — push order, member updates, ctor args, computed keys (#7634, #7628, #6998, #6986, #7640)

### LINKED ISSUE #7634: arr.push(f()) / arr.push(...g()) evaluate the receiver AFTER the argument, so a reassigning argument pushes onto the wrong array

proggeramlug:
Closes #7634, #7628, #6998. Advances #6986 (the three `lower_new` branches it names; `builtin.rs`'s ~22 arms
stay open, with the inventory on the issue) and #7640 (sections B and D; A, C and E stay open).

Also files #7700 — a behavioural gap found while verifying #6998.

The codegen lowering-window cluster: five issues, all "a lowering holds a GC value across something that can collect, or evaluates operands in the wrong order". One commit per issue, plus one for a bug I introduced and caught.

## #7634 — `arr.push(f())` evaluated the receiver AFTER the argument

Spec bug first. ES2024 evaluates the `MemberExpression` `arr.push` to a Reference **before** the argument list, so an argument that rebinds the receiver cannot redirect the push. Perry pushed onto whatever the binding named afterwards.

```
                             node 26.5.1   perry (before)
a.push(f())   f: a = [9]     [9]           [9,2]
b.push(...g()) g: b = [9]    [9]           [9,2,3]
```

**The fix is gated on the divergence being observable, and that is the substance of it.** A blanket reorder puts *every* push into a rooting window — `rows.push({...})`, `out.push(f(x))` — on what #7511 measured as the hottest store family in the compiler. `push_receiver_is_rebindable` is the as-if test: the two orders name the same array unless the argument assigns the receiver's id itself, or the binding is **boxed** (`collect_boxed_vars`' rule is "captured AND mutated", so a captured-but-never-assigned array stays on the fast path) or a module global *and* the argument can reach a collection point.

Evidence for the no-cost claim: `--trace llvm` over a 1000-iteration `out.push(mk(i))` loop plus a captured `rows.push(i * 2)` arrow contains **zero** `call i32 @js_gc_temp_root_push` and zero spec-order blocks.

**One thing the issue did not anticipate:** the reorder alone is not sufficient. Every fast tier publishes the reallocated array head back into the binding unconditionally, and once the argument may have rebound it that store lands on the wrong array (`a.push(f())` would overwrite `[9]` with the grown `[1,2]`). The spec-ordered arm therefore skips the inline tiers and guards its write-back on the binding still naming the array that was pushed onto; when it does not, the store is skipped and aliases stay valid through the forwarding pointer `js_array_push_f64` installs (#233).

`test-files/test_gap_7634_push_receiver_order.ts` matches node byte-for-byte, including `push`'s own result and the alias that keeps the array the push landed on.

## #7628 — `a[i]++` / `o.f++`

Both arms move to `RootedGroup`, which re-reads at any number of caller-chosen points. The issue asked for that combinator; **slice 6 had already built it**, so no new primitive arrives with this caller.

**The operand half the issue filed is not a live bug, and the sabotage arm is how that was established rather than argued.** Collapsing the per-use re-reads back to one — and, for `PropertyUpdate`, removing the receiver's root outright — leaves the emitted IR unchanged in the relevant respect. `root_reload` (#7280) rematerialises the slot load at every use a collection point can reach, *including* through the `ptrtoint` + `and POINTER_MASK` handle derivation that #7280's own taxonomy lists as case (a), the class it cannot repair:

```llvm
  %r49.rs4p = load ptr addrspace(1), ptr %r29     ; inserted by root_reload
  %r49      = ptrtoint ptr addrspace(1) %r49.rs4p to i64
  %r50      = and i64 %r49, 281474976710655
  call void @js_object_set_field_by_name(i64 %r50, i64 %r53, double %r46)
```

That taxonomy entry is about a raw handle a helper *returns*, not one masked out of a NaN-boxed value the pass has spilled. The per-use re-reads are kept — free, and they drop the dependence on a pass with a documented side condition — but they are documented as belt-and-braces and their two tests are **named** as pipeline assertions rather than lowering assertions.

**The repair is the result.** For a BigInt element `js_to_numeric` / `js_numeric_step` return a heap `BigIntHeader`, and the value the expression yields is live across a user setter as a bare call result with **no slot** for `root_reload` to reload from — the taxonomy's case (d). `RootedGroup::adopt_emitted` closes it, gated on `is_provably_not_bigint` so a typed-array `ta[i]++` keeps the IR it had. Sabotaging that gate turns the test red; the typed-array arm is the measured counterfactual.

## #6998 — `pointer_locals` typed every `Uint8ArrayGet` as a Number

Reachability, which the issue explicitly left open, is now established on emitted HIR rather than argued — `const it = u8[Symbol.iterator]` gives `Let { ty: Any, init: Uint8ArrayGet { index: SymbolFor(…) } }`. The arm now answers `Number` only for a **structurally** numeric key and `None` otherwise; the byte-read arm #6996 paid to keep free is unchanged and pinned.

**Found while verifying it, and worth more than the fix:** five other collectors and `type_analysis/numeric.rs` make the same unconditional assumption and force the i32 lowering, so a non-numeric key on a typed-array-typed local reads a **byte** instead of the property (`typeof u8[Symbol.iterator]` is `number`; `u8[k]` with `k = "byteLength"` is `0`). No heap value reaches such a local today, so this is a latent-soundness fix that becomes load-bearing when that is repaired. Filed separately rather than folded in — those collectors decide the `buf[i]` i32 fast path.

## #6986 — constructor arguments on `lower_new`'s non-class branches

Three branches early-`return` before the main class loop ever adopts into the enclosing scope, so the scope was open around them and empty: `new Readline(…)`, `new <importedFn>(…)`, `new Function(…)`. All three now adopt as each operand is produced — interleaved, never appended — and re-read at the call. `lower_js_args_array` is no rescue and is untouched: a plain `alloca_entry_array` pack with no `js_shadow_slot_bind`.

Not closed: `builtin.rs`'s ~22 arms. `lower_builtin_new` takes no rooting context, so they are unreachable from a `new.rs`-level fix; the inventory is on the issue.

## #7640 — sections D and B

**Section D, all six sites.** `unbox_str_handle` calls `js_get_string_pointer_unified`, which materialises an SSO value into a fresh heap `StringHeader`. Six sites took the receiver's raw untagged pointer above it. Four are pure statement swaps at zero cost; two are cross-block and re-derive the handle in the string sub-block.

**Section B, seven arms of `index_get.rs`** that lowered a receiver, then an unconstrained index (`o[f()]`), then used the receiver, with no rooting decision at all. Each is now one `with_operands_rooted` group over `[object, index]`; where the index provably cannot collect, `operand_protection` answers `Reuse` and the group emits nothing.

Sections A, C and E stay open.

## A bug this PR introduced and caught

The cross-block section-D repair first **shadowed** the entry block's `obj_handle`, which the numeric sibling block still uses — a definition in the string block does not dominate it. The LLVM verifier rejected the module, and the root-dominance corpus recorded that as ten silently "skipped" sources, including `test_gap_gc_index_get_receiver_rooting`. Distinct names; corpus back to **130/130, 0 skipped**. Recorded here because "skipped" reading as neutral in a corpus generator is the shape of #7024.

## Verification

* **node 26.5.1** (`.node-version`), byte-for-byte, for all three new gap files.
* **Behavioural smoke**, 70 `test-files/` sources matching `index|array|buffer|uint8|string|push|typed|symbol|proxy|getter`: 69 pass, 0 skip, 1 diff — `test_compat_buffers_typed`'s `toSorted`/`toReversed`, **A/B'd against a compiler built from `d1aa968e9` and byte-identical there**, so pre-existing.
* `cargo test -p perry-codegen --lib --no-fail-fast`: 786 pass. `cargo test -p perry-runtime --lib`: 1935 pass. `cargo check --all-targets`: clean.
* The 14 `native_root_coverage` tests: pass.
* All **24** `lint` commands extracted from `test.yml`, plus `cargo fmt --all -- --check`: pass.
* Root dominance, **shadow** arm (`--moving-only --seeded-violations 40`): 40 planted, 40 caught, 0 missed; corpus 130/130.
* Root dominance, **native** arm (`--statepoints --moving-only --max-unrooted 11`): see the comment below for the result.

Sabotage arms were run for every acceptance test that claims a repair, and the two that could NOT be made to fail are relabelled rather than left looking like gates.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected evaluation order for `Array.prototype.push` when arguments can rebind the receiver.
  - Improved reliability of computed property access, typed-array indexing, constructors, and increment/decrement expressions.
  - Preserved correct BigInt, numeric, symbol-keyed, and typed-array behavior across affected operations.

- **Tests**
  - Added regression coverage for receiver rebinding, computed keys, indexed updates, constructor arguments, and typed-array access.

- **Documentation**
  - Added changelog entries describing the fixes and their supported scenarios.


proggeramlug:
## Root-dominance, both arms, on the final tree

Corpus generated with this branch's compiler (`PERRY_BIN` + `PERRY_RUNTIME_DIR` pinned to the same `perry-dev` build; `PERRY_GC_MOVING_LOOP_POLLS` left to the script, which is the whole reason the corpus lives in a script).

**Shadow** — `--moving-only --min-files 90 --min-binds 1500 --min-funcs 1200 --allowlist scripts/gc_root_dominance_allowlist.json --seeded-violations 40`

```
corpus (shadow): 130/130 sources compiled, 0 skipped, 150 .ll files
=== seeded violations: 40 planted, 40 caught, 0 MISSED
DOM_SHADOW_EXIT=0
```

**Native** — `--statepoints --moving-only --min-statepoints 15000 --min-live-bundles 8000 --min-relocates 20000 --max-unrooted 11 --max-stale 0 --seeded-violations 40`

```
corpus (native): 130/130 sources compiled, 0 skipped, 150 .ll files
  statepoints: 32715   non-empty live bundles: 19658
=== statepoint hazards: 11  (unrooted: 11, stale: 0)
       4  unrooted/unmasked
       3  unrooted/alloc
       2  unrooted/global
       2  unrooted/capture
within budget: unrooted 11 <= 11
within budget: stale 0 <= 0
=== seeded violations: 40 planted, 40 caught, 0 MISSED
DOM_NATIVE_EXIT=0
```

Exactly at the #7679 ratchet — no new hits, and none removed either, so the budget is unchanged.

### The number that mattered here was `skipped`, not `unrooted`

The first run of this corpus reported **`120/130 sources compiled, 10 skipped`** and still exited 0, because `MIN_COMPILED` is 90. Those ten were not churn: nine of them plus `test_gap_gc_index_get_receiver_rooting` had stopped compiling under this branch, on an LLVM verifier error this PR introduced (`Instruction does not dominate all uses` — the cross-block section-D repair shadowed a handle the numeric sibling block still used). A corpus generator that reports a shrink as a neutral word, under a floor set generously enough to absorb it, is the #7024 shape: the gate ran, its subject did not. Fixed in `3698b264e`; both arms above are 130/130.

proggeramlug:
## Re-run after rebasing onto `6086c9d40`

Rebased past `#7694` / `#7703` / `#7696`; the one conflict was a neighbouring `#[cfg(test)] mod` line in `expr/mod.rs` (`call_spread_rooting_tests` landed on the same line as `issue7628_rooting_tests`), resolved by keeping both.

Re-verified on the rebased tree:

* `cargo test -p perry-codegen --lib --no-fail-fast` — **789 pass**, 0 fail (up from 786; `#7696` added three).
* All **24** `lint` commands + `cargo fmt --all -- --check` — pass.
* The three new gap files — match node 26.5.1 byte-for-byte.
* `./scripts/check_file_size.sh` — OK.

**Root dominance, shadow** — 130/130 sources, 0 skipped; **40 planted, 40 caught, 0 MISSED**; exit 0.

**Root dominance, native** — 130/130 sources, 0 skipped:

```
=== statepoint hazards: 8  (unrooted: 8, stale: 0)
       4  unrooted/unmasked
       2  unrooted/global
       2  unrooted/capture
within budget: unrooted 8 <= 11
within budget: stale 0 <= 0
=== seeded violations: 40 planted, 40 caught, 0 MISSED
DOM_NATIVE_EXIT=0
```

Note for whoever owns the ratchet: this is **8**, not 11. The three `unrooted/alloc` hits present before the rebase are gone with `#7696`'s spread-accumulator fix, not with anything in this PR. Lowering `--max-unrooted` to 8 is a separate change and belongs to that PR's follow-up rather than being smuggled in here — but leaving it at 11 means three regressions could now land green, so it should not be left long.

proggeramlug:
## Audit — merging as v0.5.1397

**#7634 verified independently.** `a.push(f())` where `f` rebinds `a`:

| | output |
|---|---|
| node 26.5.1 | `[9] [9] [7]` |
| `main` | `[9,3] [9,3] [7,4,5]` |
| this PR | `[9] [9] [7]` |

The push was landing on the *replacement* array instead of the original — a wrong answer with no diagnostic.

### The finding that both the issue and my brief missed

> the reorder alone is *incorrect*. Every fast tier publishes the reallocated head back into the binding unconditionally, so a rebinding argument makes that store overwrite the replacement.

That is the difference between a fix and a half-fix, and it is not visible from the spec text — you only see it by reading what the fast tiers do after the push. Dropping the inline tiers on the spec-ordered arm and guarding the write-back on the binding still naming the pushed array, with aliases staying valid through #233's forwarding pointer, is the right shape. The `[9] [1,2]` case in the gap test is the one that proves it.

Gating on `push_receiver_is_rebindable` as an **as-if** test — boxed (captured *and* mutated, so a captured-but-never-assigned array stays free) or module global, *and* the argument can collect — and then measuring **zero** `js_gc_temp_root_push` and zero spec-order blocks over a 1000-iteration loop is the right way to show a spec fix costs nothing in the common case.

### #7628: your correction to my taxonomy is the durable part

> case (a) "a pointer already unboxed to raw `i64`" is only unrepairable when a helper *returns* the raw handle; one masked out of a spilled NaN-boxed value is rematerialised whole.

I have been passing that taxonomy to every agent for a week. Establishing it by **sabotage** — collapse the re-reads, remove the receiver root, IR unchanged — rather than by argument is what makes it stick. And renaming the two non-discriminating tests as *pipeline assertions* rather than leaving them looking like gates is exactly right: a test that cannot fail is worse the more it looks like a guard.

The real repair being the **result** (case (d), no slot) for a BigInt element crossing a user setter, gated on `is_provably_not_bigint` with the typed-array arm as the measured counterfactual, is well-targeted.

### #6998: finding the second bug is worth more than the first

`typeof u8[Symbol.iterator]` is `number` and `u8["byteLength"]` via a dynamic key is `0`, because five other collectors force the i32 lowering — so the fix here is masked. Filing that as **#7700** with the reproducer and the six sites, and **not** adding a gap test because it would be red, is the honest call.

### The self-caught bug is the most instructive thing in the report

> it presented as `gc_root_dominance_corpus.sh` printing **`120/130 compiled, 10 skipped`** and exiting 0 (floor is 90) — ten sources, including `test_gap_gc_index_get_receiver_rooting`, silently leaving the corpus.

That is #7024's shape exactly, and it means a verifier rejection can shrink the corpus by 8% while the gate reports success. You fixed the *code* bug; **the gate that hid it is still open**, and I am closing that separately — a floor of 90 against a 130-source corpus lets a third of it disappear silently.

**Gates:** 24/24 lint, fmt clean, `perry-codegen --lib` **789**, `perry-runtime --lib` 1938, `native_root_coverage` 14/14, both dominance arms 130/130 with 40/40 seeded. The native budget reads 8 after rebase (#7696 lowered it), so the concern in your report resolves.

The `toSorted` diff being A/B-identical against a compiler built from `d1aa968e9` is the right way to dismiss a pre-existing failure — by checkout, not by inspection.

PR #7701, opened 2026-08-09
perf(repsel): the element-shape loop clone fires on `arr.length` and aliased element types (#7480)

proggeramlug:
Part 1 of P2 (#7480 step 4 / repsel #5093): make the element-shape versioned
loop clone actually fire, and stop two lowerings from silently deleting it.

## Why churn_read.ts never moved

#7612 landed the clone and #7669 taught it object-literal element types. Neither
moved `churn_read.ts` by a millisecond — it sat at 0.35 s across four measurement
rounds while everything around it improved 4–11×. The lowering was never the
problem. The **matcher** declined before reaching it, for two independent
reasons, each on its own fatal:

1. **`for (let j = 0; j < arr.length; j++)` failed the bound match.** The bound
   admitted a literal or a loop-invariant local; `keep.length` is a
   `PropertyGet`, so the match returned `None` at the condition, before the class
   resolver was ever consulted. That is the bound form every one of #7480's own
   kernels — and `churn_read`, `churn`, `retain` — is written in.
2. **`type Node = { v: number; w: number }` shadowed the anon-shape resolver.**
   `element_class_name` returns `receiver_class_name`'s answer when it has one,
   and for an alias-typed array that answer is `"Node"` — a name no `ctx.classes`
   entry owns, because the literals allocate an `__AnonShape_…`. The early return
   meant the resolver #7669 added for exactly this case was never reached for any
   alias-typed array.

A four-compile probe matrix localized it: inline element type + constant bound
fires; alias + constant declines; inline + `.length` declines; inline +
loop-invariant local bound fires.

## And then: two calls that delete the clone rather than slow it

Both `lower_element_shape_versioned_for` and `lower_class_field_versioned_for`
build the fast clone first and prove it call-free second. On a failed proof they
terminate the guard with an **unconditional branch to the slow clone** and leave
the fast blocks as unreachable code. So a call emitted into one of these clones
does not make it slower — it removes it, silently, with every block label still
present for an IR census to find.

- **The GC back-edge poll.** #7690 restores moving-loop polls to ON by default,
  putting a `js_gc_loop_safepoint()` in the clone's element-load block. Same
  compiler, `churn_read.ts`: polls off → clone runs, 0.03 s user; polls on →
  `br label %element_shape.loop.slow.preheader`, 0.54 s. Skipping the poll inside
  a clone is not a new licence — a poll exists so an *allocating* loop can defer
  a collection, which is why `loop_may_allocate` already gates it. That predicate
  answers from the HIR body, before specialization, so it cannot see that
  `arr[j].f` lowers to a bare load here.
- **The loop-invariant `arr.length` hoist.** With a `.length` bound the clone got
  a *second* `plen` diamond (`js_value_length_f64`) whose result nothing reads,
  because the caller had already handed the trip count in as
  `precomputed_i32_bound`. Only the load is skipped; the bounded-index and
  buffer-width facts and the i32 counter slot are proofs and storage the clone's
  other lowering may depend on.

This is the failure mode `stmt/element_shape_loop.rs`'s own module docs named in
advance — "a silent loss of the optimization … with no test failing" — and every
`perry-codegen` test was green throughout.

**The class-field clone is not affected today**, and that was checked rather than
assumed: with the suppression removed its three IR tests stay green, because
`loop_may_allocate` already proves an `obj.field`-only body inert. The
suppression covers it anyway (identical argument), and its tests gained the same
liveness assertion.

## Results

Quiet M1 mini, best-of-3 interleaved. **Both arms are rebased onto the #7690
stack** (`20fce6daa`): `origin/main` is currently GC-livelocked — #7687 shipped
only part 1 of the #7682 fix, and `churn_alloc` runs 0.43 s vs 8.7 s there — so
any allocation-touching measurement on plain `main` is meaningless.

| bench | base | this PR + #7694 | |
|---|--:|--:|---|
| **churn_read** | 0.468 | **0.042** | 11.1× — beats node 0.08 and scriptc 0.30 |
| **churn** | 0.764 | **0.471** | 1.62× |
| retain | 1.417 | 1.382 | read loop is 2.8% of it — see below |
| churn_alloc / push_cls / push_num | 0.429 / 0.416 / 0.195 | unchanged | |
| cycles / deeplist / tree / tree_wide | — | 1.00–1.03× | |
| json_roundtrip (tape-defeating) | 0.761 | 0.761 | shared read path unchanged |

All 13 benchmark stdouts byte-identical; peak RSS byte-identical (retain 367 MB
both arms).

`retain` improves by only 0.035 s, and it cannot do better from the read side:
by amplification (`retain_read10` = build + 10 read passes vs `retain` = build +
1), one read pass costs **40.0 ms → 3.9 ms (10.3×)** and is **2.8% of retain's
1.42 s**. Zeroing it entirely would save 0.040 s. The rest is 3M allocations +
GC.

## Tests

The IR census is the regression gate for this optimization, and it gained the
assertion it was missing:

- `assert_fast_clone_is_entered` — the guard must `cond_br` **into** the clone.
  Label-presence assertions cannot distinguish "optimized" from "optimization
  silently deleted". Added to all five positive tests, and to the class-field
  suite. **Sabotage-tested**: with the suppression removed, the two
  `.length`-bound tests go red with "the deref block ends in an unconditional
  branch to the slow clone", so a green run is evidence rather than decoration.
- `fast_clone_slice` now selects blocks **by name** instead of slicing a span
  between the fast and slow cond blocks. An `arr.length` bound makes the slow
  clone hoist its own length read, and those `plen.*` blocks landed in the gap —
  so a call belonging to the slow clone was attributed to the fast one and the
  census failed on a fast clone that was, and still is, bare. This is the mirror
  image of the #7480 step 3 bug the function's own doc records; a span that
  depends on what a neighbour emits can report wrong in either direction.
- New cases: `.length` bound fires and stays call-free; a **foreign** array's
  `.length` is declined (the preheader relates the two lengths not at all, so
  that would be an out-of-range read rather than a slow clone); an aliased
  element type resolves; an unresolvable `Named` element type with no class and
  no alias is declined rather than guessed; and `churn_read.ts`'s exact shape
  (alias **and** `.length` together) reaches the clone.

## Validation

- `cargo test -p perry-codegen`: 783/0 lib, all integration suites green except
  `large_object_barriers::large_local_array_push_inbounds_store_emits_precise_slot_barrier`,
  which is **red on the clean baseline commit** with no local changes (verified).
  It slices a span between `apush.inbounds.` and `apush.realloc.` — the same
  fragile pattern fixed above. Reported separately; not touched here.
- `cargo test -p perry-runtime --release`: 1930/0.
- gc-root-dominance corpus: **0 violations** over 2538 functions / 150 modules /
  9864 root stores, with `--self-test` confirming the checker can still fail.
- `cargo fmt --all -- --check`, `scripts/check_file_size.sh`,
  `scripts/addr_class_inventory.py`: clean.
- Gap suite (507 tests): no regression attributable to this change. The seven the
  snapshot flags — six network crashes and `zlib_3285_params` — reproduce
  **identically on the base compiler**, and the ten `node_fail -> parity_fail`
  status changes are the oracle running tests it previously could not. One
  improvement: `test_gap_iterator_helpers_2874: parity_fail -> pass`.

## Notes for review

- The `changelog.d/` fragment is keyed on a guessed PR number and should be
  renamed to the real one.
- A third instance of the same shape, `lower_object_array_write_versioned_for`,
  has the same call-free scan and no fact scope to key off. It is a write-path
  loop, is not covered by the benchmarks here, and is left for a follow-up rather
  than changed blind.

## Summary by CodeRabbit

* **New Features**
  * Element-shape loop cloning now supports bounds based on an array’s `.length`.
  * Cloning works with aliased array and element types.
  * Invalid foreign-array bounds and unresolved types are rejected safely.

* **Bug Fixes**
  * Fast clones remain call-free by avoiding unnecessary garbage-collection checks and redundant length loads.
  * Improved clone reachability and block tracking ensures fast paths are actually entered.


proggeramlug:
## Audit — merging as v0.5.1403, with one added assertion

I tried hard to break the widening and could not. The one thing I did find was a vacuous test, and I fixed it on the branch rather than hand it back.

### The soundness question, traced rather than assumed

Both new admission surfaces reach the **same, unmodified** call-free proof:

```rust
let fast_clone_call_free = !ctx.func.blocks()[fast_pre_idx].contains_gc_unsafe_call()
    && (fast_scan_start..fast_scan_end).all(|i| !ctx.func.blocks()[i].contains_gc_unsafe_call());
```

`match_element_shape_versioned_loop` has exactly one call site, and the scan has exactly one call site downstream of it. The `.length` bound only changes the preheader chain (never scanned); alias resolution only changes whether the matcher returns `Some`. **No bypass exists.**

What I tried to break it with:

| attack | why it fails |
|---|---|
| `for (i=0;i<a.length;i++){ a.push(x) }` | body must be exactly one `LocalSet`; a `push` isn't one. Declined at the body-shape match. |
| `for (j=0;j<other.length;j++) sum+=keep[j].v` | new `recv_id != array_id` check declines it — and it's sabotage-tested |
| alias-mutate mid-loop | structurally inexpressible in a one-statement, store-free, call-free body |
| `type A=B; type B=A` cycle | bounded at 8 hops, falls through to `Named`, declines cleanly |
| redundant `.length` hoist reintroducing #7690's silent deletion | provably dead — short-circuited by `used_precomputed_i32_cond` before the hoisted slot is ever read |

Worth noting "aliased element types" here means TS **`type` aliases**, not two bindings pointing at one array — runtime aliasing was already excluded pre-PR and stays excluded.

### The vacuous test, and the proof it was vacuous

`element_shape_versioned_loop_resolves_an_aliased_object_element_type` — the only test covering alias resolution **in isolation** — asserted `CLONE_LABELS` and the anon-shape pin, but never that the clone is *entered*. `CLONE_LABELS` is satisfied by a preheader whose deref block ends in an unconditional branch to the **slow** clone, which is exactly what a failed call-free proof emits. So the test passed on a build where the thing it tests was dead code.

Not a hypothesis — measured. With the pre-fix assertions and `if false && fast_clone_call_free`:

```
test result: ok. 1 passed; 0 failed
```

With `assert_fast_clone_is_entered` + the call-free census added (commit `907616925`), the same sabotage gives:

```
panicked at element_shape_loop_tests.rs:360:5:
the guard must branch INTO the fast clone. The deref block ends in an
unconditional branch to the slow clone, which means the call-free proof
failed and the clone is dead code
```

Your `class_field_loop_tests.rs` got this right by baking the assertion into the shared `assert_versioned_loop_lowered` helper so no future test can omit it; the element-shape file left it as a per-test opt-in, and this was the one place the opt-in was missed. Worth doing the same there eventually.

### Gates

26/26. Full `cargo test -p perry-codegen` (not just `--lib` — per-PR CI never runs `tests/*.rs`): **6 failures, byte-identical to the set `main` already has**. Rebased onto current main first; the branch was 13 versions behind, which is why its gate run initially failed on a script that didn't exist on its base.

Issue #7526, opened 2026-08-06
test_gap_fetch_instanceof_5433 SIGSEGVs, and the gap snapshot launders the crash as an accepted gap (both referenced issues closed)


proggeramlug:
## A SIGSEGV is recorded as an accepted gap

`test_gap_fetch_instanceof_5433` **segfaults** — exit 139, after printing 2 of
its expected lines:

```
$ perry test-files/test_gap_fetch_instanceof_5433.ts -o /tmp/fi && /tmp/fi
typeof: object
r instanceof Response: true
$ echo $?
139
```

Reproduced on two hosts (dev Mac and the pinned quiet mini) under `--release`,
deterministically, at v0.5.1292–1295.

It does not show up as a regression because it is in
`test-parity/gap_snapshot.json`:

```json
"test_gap_fetch_instanceof_5433": {
  "status": "crash",
  "issue": "5433",
  "added": "2026-07-04",
  "category": "bug-open",
  "reason": "fetch Response/Request instanceof (#5433); standing per #5917."
}
```

## Three things are wrong with that entry

1. **A crash is not a cosmetic gap.** `scripts/run_gap_tests.sh:110` states the
   policy in its own words: *"Crashes (SIGSEGV/SIGABRT/timeout) are hard
   defects, never cosmetic gaps."* The snapshot exists to hold known *output*
   divergences so the gate can still catch new ones. Parking a memory-safety
   fault in it converts a hard defect into background noise — and this is the
   **only** `status: "crash"` entry in the file, so it is an exception, not a
   convention.
2. **Both referenced issues are CLOSED.** `issue: 5433` → #5433 closed;
   *"standing per #5917"* → #5917 closed. The entry is marked
   `category: "bug-open"` while nothing tracks it. Per the lesson from #7518, a
   closed issue whose test still fails is a regression by definition — and here
   the snapshot itself asserts the issue is open.
3. **Nothing would notice if it got worse.** `parity` is gated to tag pushes
   only, and within it this test is pre-accepted. A segfault that starts
   corrupting memory earlier, or spreads to neighbouring tests, produces no new
   signal.

## What I am not claiming

I have not bisected this and I do not know whether the crash is new since
2026-07-04 or has been there since the entry was added — the entry records
`status: "crash"` from the start, so most likely the latter. That is worth
establishing, because "we have shipped a known segfault for a month" and "a
cosmetic gap decayed into a segfault" call for different urgency.

## Asks

[...]

PR #7709, opened 2026-08-09
fix(gc): classify eight addr-class residue sites by band, not magnitude floor

### LINKED ISSUE #7531: addr-class: eight sites still classify heap addresses by magnitude floor, not by band (the #7526 idiom)
## The idiom

Eight-plus sites classify a candidate heap address with a **magnitude floor**
and no band test:

```rust
if addr < crate::gc::GC_HEADER_SIZE + 0x1000 { ... }
```

`0x1008` is below **every** handle band, and `is_valid_obj_ptr` (often paired
with it) is also magnitude-only — its `HEAP_MIN` is `0x1000`. So the pair
excludes null and small garbage but admits every registry handle:

| band | range | source |
|---|---|---|
| common handles | `[1, 0x40000)` | `perry-stdlib` `common/handle.rs` |
| Web Fetch | `[0x40000, 0xE0000)` | `fetch/mod.rs` |
| proxy ids | `[0xF0000, 0x100000)` | `proxy.rs` |

#7526 was exactly this: `fetch_subclass_handle_id` dereferenced `0x40000` at
`addr - 8` and SIGSEGV'd on the first `new Response(...)` a program made.
Fixed in #7530 by switching to `addr_class::is_plausible_heap_addr` +
`try_read_gc_header`.

## Sites to audit

Each needs a per-site answer to "can a banded value reach here?" — a blind
sweep would be wrong, since some of these are already unreachable with handles
and a needless band test costs a branch on a hot path:

- `crates/perry-runtime/src/util_promisify.rs:752`
- `crates/perry-runtime/src/node_v8.rs:153`
- `crates/perry-runtime/src/native_handle.rs:149`
- `crates/perry-runtime/src/timer.rs:944`
- `crates/perry-runtime/src/dgram.rs:314`
- `crates/perry-runtime/src/proxy.rs:534`, `:869`, `:1149`

`native_handle.rs` and `proxy.rs` are the ones I would look at first: both sit
directly on paths where a handle is the *expected* representation, so a
banded value arriving is routine rather than exotic.

## Why this is worth a ticket rather than a sweep

macOS masks it. Its heap floor is ~2 TB, so a handle-band deref lands in
unmapped low memory and faults loudly; on platforms with a low heap base the
same read can silently hit *mapped* memory and return garbage that then flows
into an `obj_type` comparison. `run_gap_tests.sh`'s own crash banner says this:
*"Reproduce on LINUX — several crash classes (e.g. handle-band derefs, #6271)
are masked on macOS."* A silent wrong answer is worse than #7526's crash.

## Suggested acceptance

Every listed site either uses `addr_class::is_plausible_heap_addr` /
`try_read_gc_header`, or carries a one-line comment stating why a banded value
cannot reach it. Plus a band-boundary test per converted site, in the style of
`fetch_subclass_probe_rejects_every_handle_band_without_dereferencing` (#7530)
— walk the boundaries, so a band added later without a guard fails the test.

`scripts/addr_class_inventory.py` already ratchets *new* bare-address sites;
this is the pre-existing residue it grandfathered.

proggeramlug:
## Summary

Closes #7531. Eight pre-existing sites in `perry-runtime` classified a candidate heap address with a bare magnitude floor (`addr < GC_HEADER_SIZE + 0x1000` = 0x1008) instead of `value::addr_class`'s band predicates — the same shape #7526/#7530 fixed for `fetch_subclass_handle_id`. That floor sits below every handle band (common registry `[1, 0x40000)`, Web Fetch `[0x40000, 0xE0000)`, zlib `[0xE0000, 0xF0000)`, revocable-Proxy `[0xF0000, 0x100000)`), so it admits every handle id straight through to an unconditional `addr - GC_HEADER_SIZE` dereference — a SIGSEGV on Linux (masked on macOS by its ~2 TB heap floor).

Each site got the predicate its call site actually needs, not a blanket swap:

| Site | Function | What `value` actually is | Predicate |
|---|---|---|---|
| `util_promisify.rs:765` | `callable_then_field` | whatever `util.callbackify`'s wrapped function synchronously returned (e.g. `callbackify(() => new Response())`) | `try_read_gc_header` |
| `node_v8.rs:153` | `is_valid_heap_snapshot_options` | caller-supplied `options` to `v8.getHeapSnapshot`/`writeHeapSnapshot` — `is_closure_ptr` only rejected the band for closures, so a non-closure handle sailed past it | `try_read_gc_header` |
| `native_handle.rs:149` | `handle_from_value` | the receiver at every native-handle API boundary | `is_plausible_heap_addr` |
| `timer.rs:944` | `raw_closure_pointer` | raw closure-pointer bits threaded through Promise executor plumbing | `is_plausible_heap_addr` |
| `dgram.rs:314` | `gc_type_for_ptr` | reachable via any dgram Socket method receiver (`Socket.prototype.<method>.call(fakeThis)`) | `is_plausible_heap_addr` |
| `proxy.rs:534` | `create_list_from_array_like` | `argumentsList` in `Reflect.apply`/`Reflect.construct` | `is_plausible_heap_addr` |
| `proxy.rs:869` | `raw_ptr_from_value` (feeds `array_ptr_from_value`) | a Proxy `target`/`key` candidate | `is_above_handle_band` |
| `proxy.rs:1149` | `prototype_of_for_set` | the `lookup(value)` proxy-id check above it only recognizes an *already-registered* revocable Proxy id, so any other handle-band value fell through here | `is_above_handle_band` |

Two of the eight — `dgram.rs` and `proxy.rs`'s `raw_ptr_from_value` — used a local variable named `raw`, which is structurally invisible to `scripts/addr_class_inventory.py`'s `handle-floor` regex (it only matches `ptr`/`addr`/`bits`-shaped identifiers). Those were debt the ratchet could not even see; the issue's manual audit is why they're fixed here.

## Verification: old floor admitted, new predicate rejects

For `util_promisify.rs::callable_then_field` and `node_v8.rs::is_valid_heap_snapshot_options`, I added a test that asserts, per boundary probe, both halves of the gap directly: the address is numerically **at or above** the old floor (so the old code would have proceeded into the dereference) **and** the new predicate rejects it. This is the concrete "old admitted, new rejects" evidence, not just an assertion that the new code is safe in isolation.

Every one of the eight conversions additionally gets a boundary-walk test that calls the real (now-fixed) function across the handle-band boundaries and asserts rejection without a crash — in the style of #7530's `fetch_subclass_probe_rejects_every_handle_band_without_dereferencing`. For the `proxy.rs` and `dgram.rs` sites this meant calling the full function end-to-end (not just the guard in isolation) — I verified empirically first (a scratch probe, since ~5 more layers of runtime machinery sit downstream of the guard) that this doesn't crash, then wrote the real assertions.

**One site — `timer.rs::raw_closure_pointer` — has no added test.** Its file was already at the repo's 2000-line-per-file CI cap (`scripts/check_file_size.sh`, 1997 lines pre-change), so there was no budget for a new `#[cfg(test)] mod`. Separately, this site's exploit path was already closed by a *different* mechanism: the code immediately following the floor gates the actual dereference behind `gc_malloc_header_is_tracked` (a side-table membership lookup) and `arena::classify_heap_space` (an address-range registry lookup), neither of which a handle-band id can satisfy in practice. The conversion is still correct and safety-monotonic (it can only reject more addresses), but I can't point to a test proving the *old* code was reachable-and-exploitable here the way I can for the others — I want to be explicit about that rather than imply a demonstration exists.

## Ratchet

`scripts/addr_class_ratchet_baseline.txt`'s `handle-floor` total drops from 455 to 452 (`native_handle.rs` 1→0, `node_v8.rs` 2→1, `proxy.rs` 5→4, `timer.rs` 2→1, `util_promisify.rs` 1→0). The `dgram.rs` and `proxy.rs::raw_ptr_from_value` fixes were never counted (the `raw`-name blind spot above), so they don't move the number but do close the gap. `lone-valid-obj-ptr` is unchanged at 85 — none of these eight sites touch that separate, out-of-scope debt class (two of the `proxy.rs` sites sit near an existing `lone-valid-obj-ptr` finding but are more than the script's 5-line lookback window away from it, so they don't accidentally clear or perturb that count).

## Test plan

- [x] `cargo test -p perry-runtime --lib --no-fail-fast` — 1954 passed, 0 failed
- [x] `python3 scripts/addr_class_inventory.py` — passes, ratchet baseline updated (452 handle-floor / 85 lone-valid-obj-ptr, down from 455/85)
- [x] `bash scripts/check_file_size.sh` — passes (no file over 2000 lines)
- [x] `cargo fmt --all -- --check` — clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runtime safety when handling invalid or non-heap pointer values.
  * Prevented invalid handle values from being mistaken for objects, callbacks, thenables, arrays, prototypes, or datagrams.
  * Added safeguards across timers, proxy operations, native handles, V8 integrations, and utility conversions.
  * Preserved valid values such as `undefined` while rejecting unsafe boundary cases.
  * Added regression coverage to verify invalid values are rejected without crashes or unsafe memory access.


proggeramlug:
## Merging as v0.5.1410

All 8 sites converted, each to a predicate chosen from what the value actually is at that point rather than a blanket replacement — which matters, because picking the wrong band predicate here is its own bug.

The floor these replace (`GC_HEADER_SIZE + 0x1000` = `0x1008`) sits **below every handle band** — common registry `[1, 0x40000)`, Web Fetch `[0x40000, 0xE0000)`, zlib `[0xE0000, 0xF0000)`, revocable-Proxy `[0xF0000, 0x100000)` — so it admitted every handle id straight through to an unconditional `addr - GC_HEADER_SIZE` dereference.

Two of the eight had no other guard at all before the deref: `proxy.rs::create_list_from_array_like` (`Reflect.apply`'s `argumentsList`) and `proxy.rs::prototype_of_for_set`, where the `lookup(value)` check above only recognises an *already-registered* revocable Proxy id, so any other handle-band value fell through.

### The best finding here isn't in the diff

**Two sites were structurally invisible to the ratchet.** `dgram.rs::gc_type_for_ptr` and `proxy.rs::raw_ptr_from_value` name their local `raw`, and `scripts/addr_class_inventory.py`'s `handle-floor` regex only matches `ptr`/`addr`/`bits`-shaped identifiers. Confirmed empirically — fixing them doesn't move the count. So the ratchet has been reporting a number that excludes an unknown population, and the "a file gaining a bare-address site fails lint" guarantee has a hole in it whose size nobody knows.

That deserves its own issue; the ratchet's own coverage is now the open question, not these eight sites.

### Verification

`util_promisify.rs` and `node_v8.rs` got tests asserting **both halves**: each probe address is numerically ≥ the old floor (so the old code would have proceeded to deref) *and* the new predicate rejects it. Every site except `timer.rs` got a boundary-walk test calling the real fixed function end-to-end.

Honest gap, stated rather than papered over: `timer.rs::raw_closure_pointer` has **no test** — the file was at 1,997 lines against the repo's 2,000-line cap, leaving no budget. And no live "old admitted, new rejects" case could be constructed there, because the deref is already gated downstream by `gc_malloc_header_is_tracked`/`classify_heap_space`. That conversion is safety-monotonic and closes a structural gap, not a currently-reachable crash. Worth saying plainly rather than letting it ride on the other seven.

`cargo test -p perry-runtime --lib --no-fail-fast`: **1954 passed, 0 failed.** Ratchet baseline **455 → 452** `handle-floor`. Lint 19/19.

Issue #6185, opened 2026-07-09
[thread] Cross-heap unsoundness in perry/thread: global queue drains, timer queues, module globals — design decision needed


proggeramlug:
## Problem

`perry/thread`'s cross-thread machinery has several unsound edges (all code-verified in the 2026-07-09 GC audit, threading dimension):

1. **Await loop drains global queues on whatever thread runs it** (`perry-codegen/src/expr/fs_await.rs:194-207` → `promise/microtasks.rs:110-122` → `thread.rs:1283-1307`): a worker's `await` can steal another thread's completion, deserialize it into the worker's arena, and resolve a main-heap promise with a foreign pointer; the settle task then runs main-heap closures on the worker. Worker exit unmaps the arena → main thread reads freed memory.
2. **Global timer queues carry raw closure/promise pointers across heaps** (`timer.rs:25-40, 323, 367-380`): any thread's await loop fires foreign-heap timer callbacks (the "only accessed from the pump thread" SAFETY comment is not enforced).
3. **Module-level globals are process-wide but rooted/rewritten only by the thread that ran module init** (`gc/roots.rs:63,425-434`): worker closures calling top-level functions alias main-heap objects with no synchronization and no rewrite when a main-thread moving cycle relocates them.
4. **Serializer silently converts Map/Set/Promise/Error/TypedArray/ArrayBuffer captures to `undefined`** (`thread.rs:349-374`) — contradicts the documented "views deep-copy" claim; should fail loudly.
5. **Worker death leaks every malloc-tracked object** — no `Drop for MallocState` (`gc/malloc.rs:112-128`), so per-request spawn workers accrete promises/maps/closures forever.

## Proposed fix (decision needed)

Either enforce the single-JS-thread discipline mechanically — owner-`ThreadId` tags on every queued completion/timer with drains skipping foreign entries, compiler rejection (or serialize-snapshot) of module-global access in thread closures, loud serializer errors, `Drop for MallocState` — or commit to real multi-heap isolation. The current halfway point is the least safe option.

Quick wins already shipping separately (Wave-1 PR3): `ensure_gc_initialized()` at worker entry, rooting the rebuilt worker closure, a scanner/pinned-gate for `PENDING_THREAD_RESULTS`, and the ws.rs deferred-resolution fix.

## Context

2026-07-09 GC audit. Items 1-3 are P0-severity UAF classes when `perry/thread` is combined with `await`/timers/module state.

proggeramlug:
[...]

PR #7710, opened 2026-08-09
fix(test-isolation): close the async_hooks / tui::state / timer-queue split-lock gaps

### LINKED ISSUE #7680: Three more process-global sinks with split lock domains, outside the GC guards' clear list (async_hooks, tui::state, agent_dispatch timer queues)
Follow-on to #7672. That issue's survey covered the tables `gc::tests::support::reset_copying_nursery_runtime_test_state()` clears, and #7674 converts all 23 of them to per-thread storage in test builds plus a `lint` gate over that clear list.

A wider sweep of `crates/` found **three more process-global sinks with the same signature — a split lock domain — that the guards' clear list does not reach**, so #7674's gate is structurally blind to them. All three already have an assertion of the shape that made the first three flakes diagnosable (a wrong VALUE, not a timing symptom), which is what makes them worth naming now rather than after they bite.

## 1. `async_hooks` — four disjoint serialization domains over three statics

```
crates/perry-runtime/src/async_hooks.rs:90  static HOOKS: LazyLock<Mutex<Vec<HookRecord>>>
crates/perry-runtime/src/async_hooks.rs:91  static RESOURCES: LazyLock<Mutex<HashMap<u64, ResourceMeta>>>
crates/perry-runtime/src/async_hooks.rs:31  static NEXT_ASYNC_ID: AtomicU64 = AtomicU64::new(2)
```

Four domains touch them:

| domain | where |
|---|---|
| a private `TEST_LOCK` | `async_hooks.rs:1431` (tests at 1436, 1448) |
| `AsyncHookRuntimeTestGuard`'s private `ASYNC_HOOK_RUNTIME_TEST_LOCK` | `gc/tests/runtime_roots.rs:508` |
| the global side-table lock, via `CopyingNurseryTestGuard` | `gc/tests/alloc.rs:817` |
| **none at all** | `gc/tests/alloc.rs:836` `test_async_hooks_promise_alloc_remains_malloc_tracked` |

The exposed assertion is `resource_ids_are_monotonic_even_without_hooks`: `b.async_id == a.async_id + 1`. Any concurrent id allocation on another thread makes that `+2`, and the failure reads as "async ids are not monotonic" rather than "a neighbour allocated one".

`AsyncHookRuntimeTestGuard` additionally calls `test_clear_object_cache_roots()` under its *private* lock, which gives the object caches a second clearing domain on top of the guards' — #7674 makes the caches per-thread, so that particular overlap is now harmless, but the split remains.

## 2. `tui::state::SLOTS` — cleared under three different locks

```
crates/perry-runtime/src/tui/state.rs:35  static SLOTS: Mutex<Vec<u64>> = Mutex::new(Vec::new())
crates/perry-runtime/src/tui/state.rs:29  pub static STATE_DIRTY: AtomicBool
```

`tui/state.rs:145`'s `reset()` takes a module-private `TEST_LOCK`; `gc/tests/roots.rs` (10 sites) takes `lock_safe_runtime_scanner_test_guard()`; `gc/tests/cycle_state.rs:620,671` and `callback_scanners.rs:1257,1489` take the global lock via `CopyingNurseryTestGuard`.

The exposed assertion is `tui/state.rs:153 alloc_returns_sequential_handles`: `h0 == 0, h1 == 1, h2 == 2`. A concurrent `SLOTS.clear()` **or** a concurrent alloc breaks it outright, and it fails as "handles are not sequential".

## 3. `agent_dispatch_tests.rs` — a private lock over tables the guards also clear

```
crates/perry-runtime/src/agent_dispatch_tests.rs:20
static TIMER_QUEUE_TESTS: std::sync::Mutex<()> = std::sync::Mutex::new(());
```

Three tests (lines 68, 124, 169) populate the timer queues and assert on `active_timeout_resource_count()` / `scan_timer_roots`, serialized only among themselves; line 162 calls `test_clear_all_timer_scanner_roots()` without the global lock, so it damages GC tests as well as being damaged by them. This is the `opt_report` / `ext_registry` shape (#7665) exactly. #7674 makes the timer queues per-thread, which removes the *data* hazard, but the lock-domain split — and the unguarded clear — remain.

Also noted, lower severity: `event_pump.rs:741` uses a private `SERIAL` lock and works around the race with a 40-iteration retry loop, and `array/tests.rs:235,1659` carry 256-iteration retry loops with comments naming the object-cache race directly. Retry loops around a shared-state race are the symptom being described here, written down by a previous author.

## Suggested shape

Two in-tree precedents, and the choice is per table:

* **per-thread storage** (`guard_cleared_global!`, #7674) — right where the state is genuinely per-agent and the clear exists only to give a test a clean slate. `NEXT_ASYNC_ID` and `tui::state::SLOTS` both look like this: their assertions *want* per-test identity.
* **one lock domain** (`plugin::REGISTRY`, and the thread-ownership latch in `opt_report` / `ext_registry`) — right where cross-thread visibility is load-bearing.

Whichever is chosen, the thing worth keeping from #7674 is that the check lands on the **table author**, who is a finite population, and never on the reader, who is not.

proggeramlug:
Closes #7680.

## Background

Follow-on to #7672/#7674. #7674 converted the ~20 statics on the GC test
guards' clear list (`reset_copying_nursery_runtime_test_state`) to
`per_test_global!` and gated the class with `scripts/global_sink_isolation.py`.
#7680 named three more process-global sinks with the same split-lock-domain
signature, all **outside** that clear list, so the #7674 gate never saw them.

## Per-sink verdict

**1. `async_hooks` (`HOOKS`, `RESOURCES`, `NEXT_ASYNC_ID`, and five more
statics `reset_for_tests()` also clears).** Real hazard, same shape as the
three flakes #7674 fixed. `reset_for_tests()` cleared eight process-global
statics from whatever thread ran it, under any of **four disjoint lock
domains** (a private module `TEST_LOCK`, `AsyncHookRuntimeTestGuard`'s
private `ASYNC_HOOK_RUNTIME_TEST_LOCK`, the GC guards' shared lock, or —
`test_async_hooks_promise_alloc_remains_malloc_tracked` — no lock at all).
All eight are already reachable by `scan_async_hooks_roots_mut`, a scanner
registered in the same file, so `scripts/gc_runtime_root_holders.py`'s
coverage is unaffected. Converted all eight to `per_test_global!`, which
removes the need for any of the four locks — deleted the now-redundant
`TEST_LOCK` and `ASYNC_HOOK_RUNTIME_TEST_LOCK`.

**2. `tui::state::SLOTS` / `STATE_DIRTY`.** Same shape: cleared under three
different locks (a private `TEST_LOCK`, `gc/tests/roots.rs`'s
`lock_safe_runtime_scanner_test_guard()`, and the GC guards' shared lock).
`SLOTS` is already GC-scanned in the same file (`scan_state_slot_roots_mut`).
Converted to `per_test_global!`; deleted the private `TEST_LOCK`.
`gc/tests/roots.rs` still takes `lock_safe_runtime_scanner_test_guard()` —
that lock also protects `tui::hooks`'s still-unconverted slot pool and
scanner-registration bookkeeping, unrelated to this table.

**3. `agent_dispatch_tests.rs`'s timer-queue tests.** Different shape than
the issue's framing assumed. `TIMER_QUEUE` / `CALLBACK_TIMERS` /
`INTERVAL_TIMERS` were already `per_test_global!` (#7674 counted them among
its 37). Investigating why the issue still called the "unguarded clear" a
live problem turned up something more serious than a lock-domain mismatch:
**per-thread storage had silently made two of the five #6185 cross-agent
dispatch tests vacuous.** Those tests spawn a "worker" thread and assert it
can neither fire nor see a timer the "primary" thread scheduled — but under
plain `per_test_global!`, a freshly spawned thread gets its own **empty**
queue by construction, so the assertions hold no matter what
`crate::agent::owns` does.

Confirmed by sabotage: changed `owns()` to always return `true` (i.e.
disabled the owner-tag filtering these tests exist to cover) and re-ran —
only 2 of the 5 tests failed. The other 2 (`a_worker_neither_fires_nor_eats_a_primary_agent_timer`,
`a_workers_gc_scan_never_visits_another_agents_timer_slots`) passed anyway.

Fixed with a new escape hatch rather than reverting the per-thread storage
(which would reintroduce the #7672 hazard for the ~180-reader population
elsewhere): `PerThread::shared_key()` / `PerThread::adopt()`
(`per_test_global.rs`) let a test whose actual subject is cross-thread
visibility opt a deliberately-spawned worker thread into the SAME queue
instance the spawning thread already has, instead of the worker's own
isolated one. `timer/test_shared_queues.rs` exposes this for the three
timer statics; the three affected tests capture
`test_shared_queue_keys()` before spawning and call `test_adopt_queues()`
as the first thing the worker does. Re-ran the `owns()` sabotage after the
fix: 4 of 5 tests now fail (the 5th, `a_worker_disowns_the_primary_agents_work`,
tests `owns()` directly and always caught it).

Also unified the private `TIMER_QUEUE_TESTS` lock with
`crate::gc::global_side_table_test_lock()` per the issue's ask and the
`plugin::REGISTRY` precedent it names — the DATA hazard is already closed
by `per_test_global!` structurally (every libtest test runs on its own
thread), but the adoption mechanism means primary and worker genuinely
share a table for part of the test's run, so serializing against a
concurrent GC guard reset is no longer purely decorative.

## Validation

- `cargo test -p perry-runtime --lib --no-fail-fast`: 1946 passed, 0
  failed (one unrelated pre-existing timing test,
  `promise::keyed_table::tests::settling_many_keys_is_not_quadratic`,
  flaked under this host's ~70+ load average from concurrent builds by
  other agents; confirmed unrelated — untouched file, reproduces in
  isolation, passes when run alone at lower load).
- `python3 scripts/gc_runtime_root_holders.py` and `--self-test`: clean,
  identical counts to the pre-change baseline (80 holders / 47 reached / 29
  classified) — the conversions don't change what the gate sees.
- `python3 scripts/global_sink_isolation.py` and `--self-test`: clean, 0
  hazards, 93 statics classified, 15/15 self-test checks (these three
  sinks are outside that gate's clear-list scope by design, per the
  issue title).
- `cargo fmt --all -- --check`: clean.
- `scripts/check_file_size.sh`: clean (`timer.rs` stays at 1999 lines by
  moving the new test-only adoption helpers into a submodule rather than
  inlining them).
- Each of the three sinks has a deterministic regression test, verified
  by reverting the corresponding fix (temporarily, then restored) to
  confirm it fails with the exact message named in code:
  - `async_hooks::tests::async_hooks_state_survives_a_foreign_reset_for_tests`
  - `tui::state::tests::state_slots_survive_a_foreign_clear`
  - `agent_dispatch_tests`'s three timer tests, via the `owns()` sabotage
    described above (a repo-level A/B rather than an in-tree regression
    test, since the fix is about which thread reads which data, not a
    single reset call).

No version bump, no `CLAUDE.md` / `CHANGELOG.md` edits (external-style
change per repo convention for this kind of fix).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved test isolation for asynchronous hooks, runtime state, and TUI state across threads.
  * Fixed cross-thread timer and garbage-collection test behavior, including timer queue sharing and cleanup verification.
  * Prevented state resets in one test thread from affecting another thread’s state.

* **Tests**
  * Added regression coverage for cross-thread timer visibility, ownership handling, and independent state lifecycles.


proggeramlug:
## Merging as v0.5.1411

The best thing in this PR is not in the issue.

**The third sink was already fixed, and investigating it found something worse.** `TIMER_QUEUE`/`CALLBACK_TIMERS`/`INTERVAL_TIMERS` had been `per_test_global!` since #7674 — the literal ask was stale. But per-thread storage had silently made **2 of 5 `#6185` cross-agent dispatch tests vacuous**: a spawned "worker" thread got its own empty queue by construction, so the assertions passed regardless of whether `crate::agent::owns` filtering worked at all.

Proved by sabotage rather than argued: forcing `owns()` to always return `true` was caught by **2 of 5** tests before the fix and **4 of 5** after. That is a test-isolation mechanism defeating the tests it was added to protect — the same shape as the gates this project keeps finding, one level down.

The fix is right too: rather than reverting the isolation, `PerThread::shared_key()`/`adopt()` lets these specific tests opt a spawned worker into the *same* queue instance, so the isolation stays for everyone else.

### The two real sinks

- **`async_hooks`** — eight globals (`HOOKS`, `RESOURCES`, `NEXT_ASYNC_ID`, `HOOKS_ACTIVE`, `GC_DESTROY_QUEUE`, `NEXT_CONTEXT_SNAPSHOT_ID`, `CONTEXT_SNAPSHOTS`, `ASYNC_WRAP_PROVIDERS`) cleared by `reset_for_tests()` under **four disjoint lock domains**, one of which was no lock at all (`gc/tests/alloc.rs:836`). All eight are already GC-scanned by `scan_async_hooks_roots_mut`, so holder coverage is unaffected — this is isolation, not a missing root.
- **`tui::state`** — `SLOTS`/`STATE_DIRTY`, same shape, three domains.

Both got a failing-without-fix demonstration (`async_hooks_state_survives_a_foreign_reset_for_tests`, `state_slots_survive_a_foreign_clear`), each verified by reverting the conversion and reproducing the named failure.

Two now-redundant private locks deleted. `lock_safe_runtime_scanner_test_guard()` correctly **kept** — it also protects `tui::hooks`'s still-unconverted pool.

### Gates

`cargo test -p perry-runtime --lib --no-fail-fast`: 1947 passed, 0 failed. `gc_runtime_root_holders.py` + self-test clean at 80/47/29. `global_sink_isolation.py` + self-test clean, 15/15. Lint 19/19. `timer.rs` held at 1,999 lines against the 2,000 cap via a submodule — worth noting that file has no headroom left for the next change.

Issue #233, opened 2026-04-28
Array.push from inside an async function silently caps at 16 elements when the array is a function parameter


proggeramlug:
**Perry:** 0.5.346 native (macOS aarch64). Discovered while building a benchmark harness for `@perryts/mongodb` against `mongodb` (official) + `mongoose` — every cell measured exactly 16 latency samples per op when the iteration count was higher.

When an `async` helper takes an array as a parameter and `.push()`es to it, only the first 16 push calls take effect. Subsequent pushes (after `await`-resolution boundaries) are silent no-ops — `samples.length` reads back 16 forever, even though the loop continues to call the helper.

The `.push()` itself is not the issue: a direct inline `arr.push(v)` in the same loop works to N=25 (and beyond). The problem appears only when the array is passed by reference into an async function and the push happens there.

### Minimal repro

```ts
async function push(arr: number[], v: number): Promise<void> {
    arr.push(v);
}

async function main() {
    const samples: number[] = [];
    for (let i = 0; i < 25; i++) {
        await push(samples, i);
        if (i >= 14 && i <= 17) {
            console.log('after iter', i, 'samples.length=', samples.length);
        }
    }
    console.log('FINAL length:', samples.length, '(expected 25)');
}

main();
```

### Actual (Perry 0.5.346)

```
after iter 14 samples.length= 15
after iter 15 samples.length= 16
after iter 16 samples.length= 16
after iter 17 samples.length= 16
FINAL length: 16 (expected 25)
```

### Expected (Bun 1.3.12 and Node 25.8.0 — both produce this)

```
after iter 14 samples.length= 15
after iter 15 samples.length= 16
after iter 16 samples.length= 17
after iter 17 samples.length= 18
FINAL length: 25 (expected 25)
```

### Workaround that works

Inline the push in the caller — don't pass the array through an async function:

```ts
for (let i = 0; i < 25; i++) {
    samples.push(i);  // works correctly to 25
}
```

I had to do this in the bench harness — the original factored `timed(samples, fn)` helper was capping all latency samples at 16, which silently truncated percentile data without any error.

### Notes

- Family resemblance to #221 (`arr[i]=` writes silently dropped on module-level const arrays declared empty) and the closed #46 (JSON.parse silently caps output array at ~1666 entries) — same shape of "writes vanish quietly." Likely a different mechanism: this one specifically involves the async-function-parameter handoff, not module-level state.
[...]

PR #7713, opened 2026-08-09
fix(json): resolve GC_FLAG_FORWARDED array stubs in pretty/replacer stringify

### LINKED ISSUE #7269: JSON.stringify(v, null, 2) returns nondeterministic garbage for arrays grown past inline capacity — pretty/replacer emitters miss clean_arr_ptr
Found while verifying #7264 (PR #7265). Unrelated to that bug, and worse: the output is **nondeterministic garbage**, not merely wrong.

`JSON.stringify(value, null, 2)` on an array that has grown past its initial inline capacity returns a string of a different, wrong length on every run of the same binary. The non-pretty `JSON.stringify(value)` on the same array is correct and stable.

## Reproducer

```ts
function build(): any[] {
  const items: any[] = [];
  for (let i = 0; i < 32; i++) items.push({ f0: i, f1: i + 1, f2: i + 2 });
  return items;
}
const a = build();
console.log("pretty " + JSON.stringify(a, null, 2).length + " plain " + JSON.stringify(a).length);
```

```
node:   pretty 1575  plain 806
perry:  pretty 2587  plain 806
perry:  pretty 6146  plain 806     <- same binary, second run
perry:  pretty 6533  plain 806     <- same binary, third run
```

`plain` is correct every time. Only the indented form is wrong, and it is wrong *differently* each run — it is reading uninitialized/foreign memory. Slicing the result (`.slice(0, 200)`) yields an empty string, and some variants of this shape SIGSEGV.

## Threshold

It is the array's growth past its initial inline capacity (16), not the element count as such:

```
element count:   4     8     16    32
node   pretty:   190   378   775   1575
perry  pretty:   190   378   775   4097   <- and 6655 on the next run
```

Correct through 16, garbage at 32. Building the same array at **top level** instead of inside a function is correct and stable at 32 — consistent with the array being reallocated and the caller holding the pre-growth pointer.

## Suspected cause

This is the #2021 forwarding-stub class. `json/stringify.rs::stringify_array_depth` documents it and guards against it:

> an array that has grown past its initial inline capacity (16) was reallocated to a new block, leaving a `GC_FLAG_FORWARDED` stub at the old address … reading its first 8 bytes as (length, capacity) yields the forwarding pointer reinterpreted as a huge length and walks off into garbage

and calls `crate::array::clean_arr_ptr` to follow the chain. The **pretty/replacer** emitters in `json/replacer.rs` never got that guard. Unguarded `as *const crate::ArrayHeader` casts followed by a `(*arr).length` read:

| line | function |
|---|---|
| 544 | `stringify_array_with_replacer_pretty` |
| 868 | `stringify_value_pretty` |
| 1040 | `stringify_array_pretty` |
| 1297 | `stringify_array_with_array_replacer` |

(Lines 296 and 1282 in the same file *do* call `clean_arr_ptr`, so the fix is a one-line-per-site addition matching the existing pattern, not new machinery.)

A huge bogus `length` also explains the SIGSEGVs: with a replacer or an indent argument, `JSON.stringify(JSON.parse(blob), null, 2)` over a 32-element array segfaults outright.

## Scope

Reproduces on `origin/main` (`d4ac65f21`) and with auto-optimize both on and off. Not introduced by #7265 — that PR touches only the non-pretty array shape template, and `replacer.rs` does not use it. Verified by A/B against a build with #7265's runtime changes reverted: same garbage, same nondeterminism.

Impact: any code that pretty-prints JSON — config writers, log formatters, API responses built with `JSON.stringify(x, null, 2)` — silently emits corrupt output once the array crosses 16 elements. That is a wider blast radius than #7264, since it needs no `JSON.parse` and no property read.

proggeramlug:
## Summary

- Fixes #7269: `JSON.stringify(v, null, 2)` on an array grown past its initial inline capacity (`MIN_ARRAY_CAPACITY`, 16) returned nondeterministic garbage — a different, wrong-length string on every run of the same binary, occasionally crashing.
- Root cause: `js_array_grow` (#233) reallocates and leaves a `GC_FLAG_FORWARDED` stub at the OLD address. The stub's first 8 bytes are exactly `ArrayHeader.length` + `.capacity`, now holding the raw forwarding pointer to the new array. Several call sites in `crates/perry-runtime/src/json/replacer.rs` cast an array pointer straight to `*ArrayHeader` without following that chain via `crate::array::clean_arr_ptr` — the plain (non-pretty) stringify path (`json/stringify.rs::stringify_array_depth`) already did this correctly.
- Fixed sites (all in `crates/perry-runtime/src/json/replacer.rs`):
  - `stringify_value_pretty`'s `TYPE_UNKNOWN` structural-fallback probe — the site actually hit by the issue's exact repro (`JSON.stringify(v, null, 2)`, no replacer). This is the necessary fix; reverting only this one is enough to make the new test fail.
  - `stringify_array_pretty` — hardened as the shared choke point for both its callers, so it no longer depends on a caller having resolved first.
  - `stringify_array_with_replacer_pretty` — was safe today only because its one current caller happens to resolve first; fixed so it doesn't rely on that invariant.
  - `extract_string_array` and `is_array_value` — the *replacer* array argument (`JSON.stringify(v, ['a','b'])`) has the identical hazard on its own pointer; fixed while in the area.
  - One site the issue also cited, `stringify_array_with_array_replacer`, already resolved via `clean_arr_ptr` before this issue was filed and needed no change — noted in the changelog fragment.

## Test plan

- [x] New test `pretty_stringify_resolves_array_grown_past_inline_capacity` (`crates/perry-runtime/src/json/replacer.rs`): grows a real array past its allocated capacity (no GC cycle needed — `js_array_grow` installs the forwarding stub unconditionally on every reallocating grow), asserts the stale header's raw `(length, capacity)` bytes reconstruct the grown array's exact address (sabotage precondition), then pretty-stringifies the stale pre-grow pointer and asserts the output matches the array's real, current contents.
- [x] Verified the test fails without the fix: reverting the `stringify_value_pretty` resolve alone (keeping the `stringify_array_pretty` defense-in-depth resolve) still made the test process abort (`SIGABRT`, "thread panicked while processing panic") — the misread `length` (up to ~4 billion) drove the pretty-printer into a deep, self-feeding recursion over adjacent heap bytes reinterpreted as more NaN-boxed values, overflowing the stack. Restored the fix afterward and confirmed a clean pass.
- [x] `cargo test -p perry-runtime --lib --no-fail-fast`: 1957 passed, 0 failed, 4 ignored.
- [x] `cargo fmt --all -- --check`: clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed pretty-printed JSON serialization for arrays that grow beyond their original capacity.
  - Ensured array contents and replacer-based serialization remain correct when arrays move in memory.
  - Added handling for unavailable array references, producing `null` or an empty key list as appropriate.

- **Tests**
  - Added regression coverage for serialization through stale array references and verification of current array contents.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1413
+**Current Version:** 0.5.1414
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1413"
+version = "0.5.1414"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1413"
+version = "0.5.1414"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1413"
+version = "0.5.1414"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1413"
+version = "0.5.1414"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1413"
+version = "0.5.1414"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1413"
+version = "0.5.1414"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1413"
+version = "0.5.1414"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1413"
+version = "0.5.1414"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1413"
+version = "0.5.1414"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1413"
+version = "0.5.1414"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1413"
+version = "0.5.1414"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1413"
+version = "0.5.1414"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1413"
+version = "0.5.1414"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1413"
+version = "0.5.1414"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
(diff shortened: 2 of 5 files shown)


proggeramlug:
## Merging as v0.5.1414

Good result, and the interesting parts are the two places it departs from the issue.

**The issue's citation list was partly wrong, and the PR says so rather than fixing four things to match it.** Of the four sites cited (at drifted line numbers):

- `stringify_value_pretty`'s `TYPE_UNKNOWN` heuristic — **genuinely broken**, and the one the issue's own repro (`JSON.stringify(v, null, 2)`, no replacer) actually hits. Read `(*arr).length`/`.capacity` raw before resolving forwarding.
- `stringify_array_pretty` — **genuinely broken**, same raw cast.
- `stringify_array_with_replacer_pretty` — an unresolved cast in its own body, but currently **shielded** by its only caller resolving first. Fixed anyway, so correctness stops depending on a caller's behaviour.
- `stringify_array_with_array_replacer` — **already correct** before the issue was filed. Left alone.

And tracing the call graph turned up **two more** of the same bug class on the replacer-array argument (`JSON.stringify(v, ['a','b'])`): `extract_string_array` and `is_array_value`.

**The test is deterministic, which is better than the issue's own repro.** "Nondeterministic garbage" invites a test that passes by luck. Instead it grows a real array past capacity and relies on `js_array_grow` installing the `GC_FLAG_FORWARDED` stub **unconditionally on every reallocating grow** (`array/push_pop.rs:118`, issue #233) — so no GC cycle has to land for the stale pointer to exist. I verified that claim directly, along with `clean_arr_ptr`'s forwarding follow (`array/header.rs:511`), because the whole test rests on it.

It then asserts the stale header's raw `(length, capacity)` bytes reconstruct the grown array's address — pinning that the reader really is misreading a forwarding stub, not just reading something odd.

**The sabotage is the strongest part.** Reverting only the `stringify_value_pretty` fix did not produce a failed assertion — the process **aborted** (`SIGABRT`, "thread panicked while processing panic"), because the misread multi-billion `length` drove runaway recursion into a stack overflow. That is the actual severity of this bug, and it is worth more than the assertion would have been.

`cargo test -p perry-runtime --lib --no-fail-fast`: **1957 passed, 0 failed.** Lint 19/19.

PR #6977, opened 2026-07-29
fix(gc): a compiled program can now actually reach an evacuating minor (#6950)

proggeramlug:
Refs #6950 (also #6942, #6946). Follows #6974 (#6964), which must land first — this PR makes the evacuation paths reachable, and #6974 is the omission those paths would have exposed.

#6950 measured that **no reachable configuration in a compiled program evacuates**: 45 cycles under `PERRY_GC_FORCE_EVACUATE=1`, all `manual` full mark-sweeps, `total_moved_objects: 0`. It left two things open — why `PERRY_GC_HEAP_LIMIT` does not lower the first-collection point (its §2), and whether *any* configuration reaches the copying minor.

Both are answered here: one by a fix, one by measurement.

---

## 1. The pacing path read a different trigger than the arming path

`gc_budgeted_due_trigger` arms a cycle on `arena_total_bytes() >= effective_next_arena_trigger()`. That helper substitutes the device / `PERRY_GC_HEAP_LIMIT`-derived ceiling while the raw `GC_NEXT_TRIGGER_BYTES` cell still holds its 128 MB desktop-default const initializer (`GC_TRIGGER_ARMED == false` — i.e. before the first collection).

`GcDebtSnapshot::current` measured debt against the **raw cell**. So a cycle armed at a 2 MB *effective* trigger reported its own debt as **zero**, and `gc_mutator_assist_scaled_work_units` — which exists precisely to scale the assist budget by that debt — never left its 256-unit floor. The budgeted cycle crawled and never completed.

This is not a test artifact. It is unbounded heap growth in the shipped incremental mode. Measured on a compiled program (300 000 escaping allocations in a plain `while` loop, all live), release build, macOS arm64:

| | collections | RSS |
|---|---|---|
| `PERRY_GC_HEAP_LIMIT=8`, **before** | **0** | 63 MB |
| `PERRY_GC_HEAP_LIMIT=8`, **after** | **4** | — |

Instrumented, the "before" run shows 57 consecutive assists all reporting `status=ACTIVE phase=1 arena_debt=0 units=256` — one long-lived cycle, never advancing past phase 1. At 1.2 M allocations the same program reaches **330 MB RSS with zero collections**.

That is the exact unbounded-growth failure the debt-proportional pacing was written to prevent (*"the budgeted cycle NEVER completed … and RSS grew unbounded"*). The term it scales by was reading the wrong trigger.

[...]

Issue #6976, opened 2026-07-29
gc-stress gate is dark on main: #6925 left its gap file unregistered, and regressed repsel_ptr_shape_locals under PERRY_PTR_SHAPE_LOCALS=0


proggeramlug:
Two problems, both from #6925 (`perf(codegen): repsel Phase 5a — Ptr<Shape> proven this in methods`, merged as `1a533a3a8`). The second one hid the first.

## 1. The Phase 5a gap file was never registered, which breaks the whole matrix gate

`#6925` added `test-files/test_gap_repsel_proven_this_frozen.ts` but did not add it to `test-parity/gc_repsel_corpus.txt`. That is the omission the manifest exists to catch, and `scripts/gc_repsel_matrix.sh` enforces it by **exiting 3 before running anything**:

```
UNREGISTERED test-files/test_gap_repsel_proven_this_frozen.ts is a representation-selection gap file but is not in test-parity/gc_repsel_corpus.txt
A NEW REPRESENTATION MUST REGISTER ITS GAP FILE in test-parity/gc_repsel_corpus.txt.
```

So `gc-stress` has been failing on `main` for every PR since #6925 merged — and because the script bails up front, it reported the registration error rather than running the corpus. **The gate has been dark, not green.**

(Registration is one line; it is included in #6975 so the gate starts running again.)

## 2. Phase 5a regressed its own OFF arm: `repsel_ptr_shape_locals × rep_ptr_shape_off`

With the gate running again, a genuine red cell appears — and it is Phase 5a's own escape hatch:

```
arm rep_ptr_shape_off   (compile: PERRY_PTR_SHAPE_LOCALS=0
                         run:     PERRY_GC_HEAP_LIMIT=8 PERRY_GC_FORCE_EVACUATE=1)
FAIL test_gap_repsel_ptr_shape_locals (exit=1 cycles=0 moved=0)
```

```
$ PERRY_PTR_SHAPE_LOCALS=0 perry test-files/test_gap_repsel_ptr_shape_locals.ts -o ps
$ PERRY_GC_HEAP_LIMIT=8 PERRY_GC_FORCE_EVACUATE=1 ./ps
...
TypeError: Cannot read properties of undefined (reading 'area')
    at <anonymous>
```

The program dies partway: the last five lines of expected output are missing.

```
  p5a-counter:2001.5:number
  p5a-chain:12500|0.08|0.24,0.32,0.08
  p5a-leak:99:true:2
  p5a-closure:20:20
[...]

PR #6925, opened 2026-07-28
perf(codegen): repsel Phase 5a — Ptr<Shape> proven `this` in methods

proggeramlug:
> [!IMPORTANT]
> ## 🚧 DO NOT MERGE YET — two merge preconditions are owed
>
> **1. A clean full gap-suite sweep** (`./scripts/run_gap_tests.sh`) against pinned Node 26.5.0, on a quiet host, **has NOT been run.**
>
> It was attempted and deliberately aborted: free disk first sat at ~5 GiB against the ≥25 GiB precondition, and when disk recovered the host's 1-min load average was 54–56 with six concurrent agent `cargo build`s. A 410-program byte-comparison sweep under that contention risks timeout-induced spurious failures and would degrade a shared machine, so it was stopped cleanly (verified: no orphaned `perry_parity`/gap-runner processes) rather than publishing an untrustworthy verdict.
>
> **2. A symbol-collision guard for the `__pshape` suffix** is not yet written (CodeRabbit, Major — assessed [here](https://github.com/PerryTS/perry/pull/6925#issuecomment-5105967880)).
>
> Method `foo` yields clone `..._foo__pshape`, byte-identical to the PUBLIC symbol of a user method literally named `foo__pshape` in the same class → duplicate LLVM definitions. Real, but **pre-existing and family-wide**: `__pshape` is the 8th such suffix, and `foo` + `foo__generic` already collides on `main` today. It fails **loudly** at the IR verifier — a hard compile error, never a silent miscompile. The general fix for all eight suffixes is tracked in **#6927**; this PR will carry only the minimal local mitigation (refuse to emit the clone when its composed name is already a registered method symbol, standing down to today's always-correct guarded lowering), with the code comment pointing at #6927.
>
> Everything else is validated — see [Validation](#validation). What has been run: the repsel-filtered suite **12/12 across all four flag × GC-evacuation arms**, the full structural proof pre- and post-`opt -O3`, and CI (`cargo-test` green). What is missing is the *whole-corpus* byte-exactness confirmation and the guard above.
>
> Machine priority is currently held by the GC-correctness work (#6892, #6910/#6921). Both preconditions will be completed once that clears — the guard lands first, so the sweep validates final code.

Phase 5a of the representation-selection line (`docs/representation-selection-rfc.md`, §5.4/§5.6/§5.7), on top of phases 1/2/3a/3b/4a/4a.3/4b.

## Thesis

[...]

PR #7718, opened 2026-08-09
fix(repsel): make scalar replacement's method-call summary independent of PERRY_PTR_SHAPE_LOCALS

### LINKED ISSUE #6984: repsel: PERRY_PTR_SHAPE_LOCALS=0 breaks test_gap_repsel_ptr_shape_locals (TypeError, exit 1) with no GC involved — the Phase 3b kill switch does not disable Phase 5a
Pre-existing on `main`; surfaced while running `scripts/gc_repsel_matrix.sh --arms all` for #6950. **No GC is involved** — the failing cell reports `cycles=0 moved=0`.

## Repro

```
$ PERRY_PTR_SHAPE_LOCALS=0 perry test-files/test_gap_repsel_ptr_shape_locals.ts -o psoff
$ ./psoff
...
TypeError: Cannot read properties of undefined (reading 'area')
    at <anonymous>
$ echo $?
1
```

Diff against the Node oracle — the run dies just before the Phase 5a block:

```
18a19,22
> p5a-iface:3750:3:12
> p5a-named:42:21
> p5a-inherit:10:20:5:2:3
> p5a-collide:2:102:1102:56375:1152
```

Everything up to line 18 matches. With `PERRY_PTR_SHAPE_LOCALS` unset (the default) the whole file passes.

## What it means

`PERRY_PTR_SHAPE_LOCALS=0` is the documented A/B kill switch for repsel Phase 3b (`Ptr<Shape>` proven object locals, #6911). It is supposed to fall back to the boxed lowering and produce identical output — that is the entire point of the `rep_ptr_shape_off` matrix arm. Instead the fallback path produces `undefined` where an object is expected.

The four missing lines are the Phase 5a assertions (`Ptr<Shape>` proven `this` in methods) that #6925 appended to this file. So the most likely reading is that Phase 5a's lowering does not honour the Phase 3b kill switch: with the switch off, the `this` proof is still applied (or the fallback for it is missing) and the receiver comes out `undefined`.

## Why it was never seen

Two independent reasons, both from #6925:

1. It shipped `test-files/test_gap_repsel_proven_this_frozen.ts` without registering it in `test-parity/gc_repsel_corpus.txt`, so `gc_repsel_matrix.sh --arms all` aborted at the UNREGISTERED gate before evaluating a single cell. (Fixed in #6977.)
2. `rep_ptr_shape_off` is not in the PR arm subset, so no per-PR run covers it.

The matrix run in #6977 is the first `--arms all` that started since `1a533a3a8`, and this is the one cell that was red before any of that PR's changes took effect: baseline `summary: PASS=29 UNVER=350 XFAIL=0 FAIL=1`, and this is the 1.

Related: #6911, #6925, #6950, #6977.

proggeramlug:
## Summary

- Fixes #6984 (and closes out the item #6976 left open): `PERRY_PTR_SHAPE_LOCALS=0` crashed `test_gap_repsel_ptr_shape_locals.ts` with `TypeError: Cannot read properties of undefined (reading 'area')`, with no GC involved. This is the kill switch for repsel Phase 3b/5a (`Ptr<Shape>` proven object locals + the proven-`this` method clones it feeds) — an OFF arm is supposed to be the *safest* cell in the matrix, and it wasn't.
- **Root cause was NOT in Phase 3b's provenance logic or its eligibility gates.** `collectors/ptr_shape.rs` and `collectors/proven_this.rs` both correctly bail to an empty proof set when the switch is off. The bug was one level down, in a completely different, supposedly-orthogonal optimization: **scalar replacement of `new` locals** (`collectors/escape_news.rs`), and specifically its method-call summarizer in `lower_call/scalar_method.rs`.
  - That summarizer (`try_lower_scalar_replaced_method_call`) resolved a scalar-replaced receiver's class via `type_analysis::predicates::receiver_class_name`. That function's fallback arm — reached whenever Phase 3b's `Ptr<Shape>` proof is unavailable for *any* reason, not just the kill switch — returns the local's **declared** type name.
  - For `const o: Shaped = new Impl(...)` (an interface-typed local with `new` provenance — exactly the "predicates.rs narrowing" scenario `test_gap_repsel_ptr_shape_locals.ts`'s `ifaceLocal` was written to test), that declared name is `"Shaped"`, which is not a registered class.
  - `simple_scalar_method_summary(ctx.classes, "Shaped", "area", 0)` then fails to find the class, the summarizer gives up (`return Ok(None)`), and the caller falls through to the ordinary heap-object method-dispatch lowering — which reads the receiver from `ctx.locals[receiver_id]`. But scalar replacement had already turned that slot into a bare, **never-initialized** alloca (the heap allocation was elided; see `let_stmt.rs`'s `let dummy_slot = ctx.func.alloca_entry(DOUBLE);` with no store). Reading it produces garbage that happens to decode as `undefined`, and `o.area()` throws.
- **Fix**: resolve the receiver's class from `ctx.non_escaping_news` first — the exact map that already gated scalar replacement for that local in `let_stmt.rs`, keyed by the `new` expression's own class name, never the declared annotation — before falling back to `receiver_class_name`. This is provably safe: `ctx.scalar_replaced.contains_key(receiver_id)` being true already certifies (via the escape analysis's own `mark_unstable_scalar_method_receivers`/`check_escapes_in_expr`, both keyed off the *same* `non_escaping_news` class name) that `simple_scalar_method_summary` succeeds for this exact class+method, so the fix can only ever repair a wrong answer, never introduce one.

## Why this was invisible until now

Two independent reasons (from #6976, which this PR's triage removal and CI arm close out):
1. `#6925` shipped `test_gap_repsel_proven_this_frozen.ts` without registering it in `test-parity/gc_repsel_corpus.txt`, which made `scripts/gc_repsel_matrix.sh` exit 3 before evaluating any cell — the gate was dark, not green. (Already fixed by `#6977`.)
2. `rep_ptr_shape_off` — the arm that compiles with the switch off — is not in `gc_repsel_matrix.sh`'s PR-gating subset (`PR_ARMS`), so no per-PR CI run ever exercised it. It only ran via `--arms all`, which fires on push-to-main/schedule, not per PR.

## Decision: fix, not delete

Per CLAUDE.md's GC knob kill-policy ("every GC env knob either has a required CI arm exercising its OFF state, or it is deleted after one release of soak") and "a mode that still exists is a decision that hasn't been made" — the OFF state's failure here was a straightforward, fully-explained codegen defect in an unrelated optimization, not evidence that the representation itself only works with the optimization enabled. The switch has real bisection value (it is what let `#6976`'s own triage separate "genuine repsel regression" from "everything else"), so **fixing it and exercising it in CI** is the right call, not deleting it.

## Verification

- Reproduced on `main` exactly as described:
  ```
  PERRY_PTR_SHAPE_LOCALS=0 perry test-files/test_gap_repsel_ptr_shape_locals.ts -o psoff
  ./psoff   # TypeError: Cannot read properties of undefined (reading 'area'); exit 1
  ```
- Localized with `--trace llvm` on a minimal repro (interface + concrete class + `for` loop + return statement calling `.area()` twice) — confirmed the default (switch-on) LLVM IR fully scalar-replaces the receiver and inlines `.area()`, while the switch-off IR reads an alloca (`%r4`) that is declared but never stored anywhere in the function.
- After the fix: both arms produce byte-identical output to the pinned Node oracle for `test_gap_repsel_ptr_shape_locals.ts`.
- `scripts/gc_repsel_matrix.sh --arms rep_ptr_shape_off --filter test_gap_repsel_` (the whole representation-selection corpus, 21 files) is **21/21 PASS, 0 FAIL, 0 UNVER**, with genuine liveness under the arm's evacuating base (`collected 21/21 reclaimed 21/21 moved-objects 21/21 copy-minor 21/21` — not a vacuous green from an inert arm).
- Default (switch-on) build: byte-identical LLVM IR before and after the fix for the reproducer — the change is a pure no-op there, satisfying the "a knob may move only the sites of the representation it names" rule.
- `cargo fmt --all -- --check` clean; `cargo clippy -p perry-codegen --lib` shows no new warnings (diffed against the pre-existing baseline); the existing `collectors::scalar_method_dispatch` unit tests (9/9) still pass.

## CI arm

Adds `.github/workflows/gc-ptr-shape-off-witness.yml`, a per-PR job that runs `scripts/gc_repsel_matrix.sh --arms rep_ptr_shape_off --filter test_gap_repsel_` — closing the exact hole described above. Checked against CLAUDE.md's "four ways a gate can be unable to fail":

1. No `continue-on-error`, no `|| true`, no pipe between the matrix and the shell's exit status.
2. **Deliberately NOT in branch protection's required contexts.** A brand-new gate has never had a green run on `main`, and promoting it immediately would block every open PR. Promotion is a maintainer action, not part of this PR.
3. `concurrency` cancels pull-request runs only; `main` runs are keyed on the commit SHA so they queue instead of cancelling each other (mirrors `gc-moving-witnesses.yml`'s `#7205` fix).
4. `scripts/gc_repsel_matrix.sh`'s own exit status already folds in the liveness gate (`scripts/gc_matrix_liveness_check.py`) alongside byte-exactness, so a green run means the arm actually compiled, actually ran under evacuation, and actually matched the oracle — not merely that nothing threw.

Also removes the now-obsolete `#6976` triage entry for `test_gap_repsel_ptr_shape_locals | rep_ptr_shape_off` in `test-parity/gc_repsel_triage.txt` (per that issue's own "REMOVE THIS ENTRY when #6976 is fixed" instruction), and registers the new workflow in the two places that describe `test-parity/gc_repsel_corpus.txt`'s runners (`docs/src/testing/test-registration.md`, `scripts/check_test_registration.py`'s dark-test message).

## Test plan

- [x] `test_gap_repsel_ptr_shape_locals.ts` passes under both `PERRY_PTR_SHAPE_LOCALS` on and off, byte-exact vs. the pinned Node oracle.
- [x] `scripts/gc_repsel_matrix.sh --arms rep_ptr_shape_off --filter test_gap_repsel_` — 21/21 PASS, live under evacuation.
- [x] Default build's LLVM IR is byte-identical before/after the fix (no regression risk to the shipped configuration).
- [x] `cargo fmt --all -- --check`, `cargo clippy -p perry-codegen --lib` (no new warnings), existing `scalar_method_dispatch` unit tests.
- [ ] CI (this PR's own checks, once opened).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed crashes involving pointer-shape optimization and kill-switch configurations.
  * Improved method dispatch for scalar-replaced objects, including interface- and alias-typed values.

* **Testing**
  * Added automated coverage for pointer-shape-disabled garbage-collection scenarios.
  * Promoted previously excluded representation-selection tests to required checks.

* **Documentation**
  * Updated testing guidance and changelog entries.

* **Chores**
  * Incremented the release version to 0.5.1415.


proggeramlug:
## Merging as v0.5.1415

The root cause is better than the issue's own hypothesis, and materially wider than a kill-switch bug.

**It is not in Phase 3b.** `collectors/ptr_shape.rs` and `collectors/proven_this.rs` both gate correctly on `ptr_shape_locals_enabled()` and yield an empty proof set when the switch is off. The defect is one level down, in a supposedly orthogonal optimization:

`try_lower_scalar_replaced_method_call` resolved a **scalar-replaced** receiver's class through `receiver_class_name`, whose fallback arm returns the local's **declared** type name. For `const o: Shaped5a = new Impl5a(...)` that is `"Shaped5a"` — an interface, not a registered class. The lookup fails, `simple_scalar_method_summary` returns `None`, the function bails, and the caller falls through to ordinary heap-object dispatch — which reads `ctx.locals[receiver_id]`. But scalar replacement had already made that slot a bare `alloca_entry(DOUBLE)` **with no store**. Reading it yields garbage that decodes as `undefined`.

**The part that matters beyond this ticket:** that fallback fires whenever the `Ptr<Shape>` proof is unavailable **for any reason** — not only under `PERRY_PTR_SHAPE_LOCALS=0`. The kill switch did not cause this bug; it made a latent default-path hazard deterministic and therefore findable. That is exactly the argument for keeping kill switches exercised rather than deleting them, and it is why "fix, not delete" is the right call here.

The fix resolves the class from `ctx.non_escaping_news` — the same map that gated scalar replacement for that local, keyed by the `new` expression's own class name rather than the annotation. Provably safe in one direction: `scalar_replaced.contains_key(receiver_id)` already certifies, via escape analysis keyed on that same map, that the summary succeeds for that exact class+method. It can repair a wrong answer, never introduce one.

**Diagnosis method worth copying:** `--trace hir --focus ifaceLocal` showed HIR **identical on both arms**, proving this was purely a codegen decision and saving a pointless hunt through lowering; `--trace llvm` then showed the off-arm reading an alloca declared but never stored anywhere in the function.

### Verification

- Both arms byte-identical to the Node oracle on `test_gap_repsel_ptr_shape_locals.ts`.
- **Default-arm LLVM IR byte-identical before/after** — a pure no-op on the shipping path.
- `gc_repsel_matrix.sh --arms rep_ptr_shape_off`: **21/21 PASS, 0 FAIL, 0 UNVER**, with genuine liveness under evacuation (`collected 21/21 moved-objects 21/21`) — not a run that passed by never collecting.
- Lint 19/19.

**CI arm added** (`gc-ptr-shape-off-witness.yml`) so the OFF state is exercised per PR; it was previously only in `--arms all`, which fires on push-to-main and schedule, **not on PRs**. Promotion to a required context stays a maintainer action — a new gate has never been green, so making it required immediately would block every open PR.

Issue #7708, opened 2026-08-09
CI cannot see crates/*/tests/*.rs regressions: per-PR runs --lib --bins, and SOURCE_SUITE_MAP is hand-maintained (2 failures beyond #7494)


proggeramlug:
`cargo test -p perry-codegen` is red on `main` (v0.5.1407) with **6 failures across 4 integration suites**. All 6 predate the PRs merged today (verified by running the same command on `main` before and after: identical set).

```
integer_arithmetic_array_push_omits_inbounds_layout_note_and_barrier
integer_modulo::i32_counter_mod_unsafe_or_nonliteral_divisors_keep_frem
large_local_array_push_inbounds_store_emits_precise_slot_barrier
proven_buffer_and_typed_array_reads_are_numeric_operands
reassigned_typed_array_store_records_runtime_fallback
typed_f64_receiver_method_clone_raw_loads_after_composed_guards
```

## Why this sat unseen

Per-PR `cargo-test` runs `--lib --bins` (`.github/workflows/test.yml:829`) — **no `tests/*.rs` integration suite runs on a PR at all.** `e2e-scoped` runs only suites the diff *names*, plus a hand-maintained `SOURCE_SUITE_MAP` that maps all of `crates/perry-codegen/src/` to exactly three suites, none of these four.

`scripts/ci_e2e_scope.py`'s own header already names this as the #6037 class: *"a source change that regresses an existing suite the diff doesn't name… can land green and sit red for days."* This is that, except it is longer than days and nobody has a number for how long, because nothing reports it.

## Why it matters beyond these 6

Today two separate PRs (#7698, #7702) each broke a `perry-codegen` integration test, and in both cases the *only* way to find out was to run the full suite by hand and diff against a hand-captured `main` baseline. That is not a workflow — it depends on someone deciding to check. Both were caught; the next one won't be.

The two failures each PR introduced were also *different in kind*, which is worth recording:

- **#7702** was a genuine assertion that had to change (it required a recording call the PR deliberately stops emitting).
- **#7698** was a test that had **stopped being able to see its subject** — the new conditional split a basic block in three, and the helper sliced text to the first `br`. The write barrier had not moved; the test just couldn't find it, and no wider text slice would have worked because LLVM emits the join block *between* the other two.

## Suggested

1. Triage the 6 — some may be stale assertions, some may be real.
[...]

Issue #6969, opened 2026-07-29
gc: constructor arguments (new C(a, b)) are not precise roots across the instance allocation


proggeramlug:
Found while fixing #6951, which fixed the *variadic* argument accumulator (`console.log`, spread/rest lists) and the concat/literal families. Constructor arguments are the same bug class at a different site and are **not** covered by that fix.

## Repro

```ts
let sink: unknown[] = [];
function churn(n: number): number {
  let acc = 0;
  for (let i = 0; i < n; i++) {
    sink.push({ i: i, s: "x" + (i & 255), a: [i, i + 1] });
    if (sink.length > 4096) { acc = (acc + sink.length) | 0; sink = []; }
  }
  return acc | 0;
}
function fresh(k: number): string { return "f" + k + "-" + (k * 7); }
class Pair {
  a: unknown;
  b: unknown;
  constructor(a: unknown, b: unknown) { this.a = a; this.b = b; }
  toString(): string { return "<" + this.a + "," + this.b + ">"; }
}
const N = 420000;
// Repeat 6x: the failure is a use-after-free and needs the block recycled.
{ console.log(String(new Pair(fresh(0), churn(N)))); }
{ console.log(String(new Pair(fresh(1), churn(N)))); }
{ console.log(String(new Pair(fresh(2), churn(N)))); }
{ console.log(String(new Pair(fresh(3), churn(N)))); }
{ console.log(String(new Pair(fresh(4), churn(N)))); }
{ console.log(String(new Pair(fresh(5), churn(N)))); }
```

`PERRY_CONSERVATIVE_STACK_SCAN=off PERRY_GC_HEAP_LIMIT=8 ./repro` prints wrong output (pre-#6951 it segfaulted; post-#6951 it is a silent DIFF because the surrounding concat is now rooted).

## Why

`lower_call/new.rs` lowers all constructor arguments into `lowered_args` up front, then allocates the instance and calls the constructor:

```llvm
%r10 = call double @..._fresh__spec_i32(i32 0)          ; heap string, SSA register
%r13 = call double @..._churn(double %r12)               ; collects -- %r10 dies
%r17 = call i64 @js_object_alloc_class_inline_keys(...)  ; collects again
%r20 = call double @..._Pair_constructor(double %r19, double %r10, double %r13)
```

`%r10` is live across two collection points with no root.

## Fix shape

[...]

PR #7719, opened 2026-08-09
fix(gc): root builtin.rs's constructor arguments across each other's lowering (#6986)

### LINKED ISSUE #6986: gc: constructor arguments on lower_new's non-class branches (Readline / imported ctor / new Function) and builtin.rs are still not precise roots
Found while fixing #6969 in #6983, and deliberately left out of that PR's scope.

#6983 roots the constructor arguments built by `lower_new_impl_inner`'s **main
class path** (`crates/perry-codegen/src/lower_call/new.rs`, the `lowered_args`
loop reached after the `ctx.classes` lookup). Three *earlier* branches of the
same function build multi-argument lists the same way and are still unrooted, as
are the multi-argument built-in constructors in `lower_call/builtin.rs`.

## Sites

`crates/perry-codegen/src/lower_call/new.rs`:

- **`new Readline(...)`** (`node:readline/promises`) — lowers `output`, then
  `options`, then `for extra in args.iter().skip(2) { lower_expr(...) }`, and
  passes `output`/`options` live into `js_readline_promises_readline_new`. A
  later argument's lowering can collect while the first two sit in SSA
  registers.
- **`new <importedFn>(...)`** — `for a in args { lowered_args.push(lower_constructor_arg(ctx, a)?) }`
  then `lower_js_args_array`, which is a plain alloca pack with no rooting
  (`expr/helpers.rs`), passed to `js_new_function_construct`.
- **`new Function(...)`** with a dynamic body — the identical loop, into
  `js_function_ctor_from_strings`.

`crates/perry-codegen/src/lower_call/builtin.rs` (untouched by #6983): several
built-in constructors lower `args[0]` then `args[1]` sequentially via plain
`lower_expr`. `lower_builtin_new` is called from inside `lower_new_impl_inner`,
so these are on the same path. I did not audit each one; the shape is the same.

## Why these are cheap to fix

#6983 already wraps the whole function in an expression-scope temp-root barrier
(`temp_root_scope_begin` / `_end` in `lower_new_impl`), and it is a stack *cut* —
so these branches are already inside a scope that releases correctly on every
return path. What is missing is only the per-argument push + re-read. The
primitives are `temp_root::operand_needs_root`, `temp_root_push_double`,
`temp_root_get_double`, and `operand_is_reloadable` for operands that read a
registered root.

**One trap, learned the hard way in #6983:** the push must be interleaved with
the lowering loop, not appended after it. Rooting the finished list publishes an
already-dangling argument 0 to the scanner, which turns a silent DIFF into a
SIGSEGV — strictly worse than not rooting at all.

## Detection

Same as the rest of the family: invisible with the conservative native-stack
scan on. Reproduces under `PERRY_CONSERVATIVE_STACK_SCAN=off
PERRY_GC_HEAP_LIMIT=8` with a shape like
`new Function(fresh(0), "return " + churn(N))`, and under the evacuating arm
from #6977 (`PERRY_GC_INCREMENTAL=0` + the above).

Related: #6969, #6983, #6951, #6972, #6975, #6981, #6968.

proggeramlug:
## Summary

Closes #6986. #7699 fixed the three non-class branches of `lower_new_impl_inner` (`lower_call/new.rs`) named in the issue and explicitly left `lower_call/builtin.rs`'s arms open, with the note "builtin.rs's ~22 arms stay open, with the inventory on the issue."

An arm-by-arm audit found **30 match arms** (not the rough "~22" estimate — the issue's own author said "I did not audit each one") that lower `args[0]` then `args[1]` (then, for several, discard the rest for side effects) with plain `lower_expr` and no rooting decision at all: `Utf8Stream`, `EvalError`/`URIError`, `Uint8Array`, the typed-array-view family, `DataView`, `RegExp`, `EventEmitter`, `EventEmitterAsyncResource`, `SocketAddress`, `BroadcastChannel`, `Event`, `CustomEvent`, `DOMException`, `Console`, `StringDecoder`, the `Readable`/`Writable`/`Duplex`/`Transform`/`PassThrough` family, `LRUCache`, `DatabaseSync`, `StatementSync`, `Session`, `RateLimiterMemory`, `CronJob`, `AsyncResource`, `SuppressedError`, `WeakMap`, `WeakSet`, `TextDecoderStream`, `CompressionStream`/`DecompressionStream`, `ReadableStreamBYOBReader`, `CountQueuingStrategy`/`ByteLengthQueuingStrategy`.

`WeakMap`/`WeakSet` are a variant of the same bug rather than the textbook shape: the iterable argument was lowered (eagerly, via `.map(lower_expr)`), then `js_weakmap_new`/`js_weakset_new` — an *unconditional* allocation — ran, and only then was the iterable's now-possibly-stale register read back out.

## What changed

`lower_builtin_new` now takes the caller's `RootedGroup` (threaded in from `lower_new_impl_inner`, which already opens one per #6969/#7699 — both of `lower_builtin_new`'s call sites in `new.rs` pass it through). Three small helpers do the adoption, following #7699's own stated discipline exactly — "adopt as the value is produced, never after the fact; rooting a finished list publishes an already-dangling argument 0, which is worse than not rooting at all":

- `adopt_optional_arg` — the one-operand-at-a-time primitive: lowers `args[idx]` if present, rooted across everything from `args[idx+1..]`.
- `adopt_leading_arg_discard_rest` — the single-leading-argument-plus-discard-loop shape (>12 arms).
- `adopt_two_leading_args_discard_rest` — the two-leading-arguments-plus-discard-loop shape (`Event`, `CustomEvent`, `DOMException`, `Console`, `TextDecoderStream`).

`CronJob` needed a bespoke ordering rather than the generic helpers: its `cronTime` argument's raw-pointer derivation (`js_get_string_pointer_unified`) can itself allocate (SSO materialize), so it has to run *before* `onTick`/`start` are re-read from their slots, not after — otherwise the fix would trade one unrooted register for another.

**Explicitly out of scope**, and why: the `extract_options_fields`-based arms (`Response`, `Request`, `Blob`, `File`, `Headers`, `ReadableStream`, `WritableStream`, `TransformStream`) share the same underlying hazard — an earlier field's lowered value can sit unrooted across a later field's — but their per-property-match loop over a dynamic `Vec<(String, Expr)>` is a structurally different shape from the fixed `args[0]`/`args[1]`/`args[2]` sequence. Reusing these three helpers there isn't a good fit; it needs its own audit.

## Verification

**Unit tests** (`crates/perry-codegen/src/temp_root_coverage/builtin_ctor.rs`, so they run in the per-PR `--lib --bins` gate, not the nightly-only `tests/*.rs` tier): six tests covering all three helper shapes plus the `WeakMap` variant, through `perry_codegen::testing::temp_slots`'s codegen-contract assertions (same infrastructure #6969/#6983 used). Sabotage-confirmed: copied onto a clean `origin/main` checkout, the five positive assertions **fail** against the pre-fix code (`%r1 is never stored into a rooted slot`, etc.); the paired negative gate (`RegExp` with two non-allocating arguments) **passes** on both, so the positives aren't vacuously satisfied by a compiler that roots everything.

**Static checker** (`scripts/gc_root_dominance_check.py`), both lowerings:
- The *existing* corpus (`scripts/gc_root_dominance_corpus.sh`, `test_gap_gc_*`/`test_gap_new*`/etc.) reads 0 violations before and after — it contains no source that constructs any of these built-ins, so it can't show a reduction either way.
- A scoped probe built from ordinary TS object/array-literal arguments *also* read 0 both ways — not because the fix is a no-op, but because Phase-3 closed-shape synthesis and the array-literal inline-bump-allocator diamond each bind their own shadow slot for unrelated reasons, which happens to also cover a plain constructor argument end-to-end. This is a real gap in what a naive probe can show, not a claim of full checker coverage.
- Switching to the shape the project's own reproducers use elsewhere (`fresh(k)` / `"x" + churn(N)`, matching #6969's own `new Function(fresh(0), "return " + churn(N))`) found the checker's actual reach: on `new DataView(fresh(1), "o"+churn(N), "l"+churn(N))` and `new SuppressedError(fresh(3), "y"+churn(N), "z"+churn(N))`, the middle argument's own producing call (`js_string_concat_value`) is itself in `ALLOC_RE`, and the checker reports it stale.
  - **`--stale-registers` (shadow, `PERRY_RS4GC=0`): 2 → 0.**
  - **`--statepoints` (native, `PERRY_RS4GC=1` + production `rewrite-statepoints-for-gc`): 2 unrooted → 0.**
  - The 2-argument arms in the same probe (`RegExp`, `EventEmitter`) read 0 both ways in this probe because their first argument comes from a plain user-function call (`fresh(k)`), which `ALLOC_RE` — by design, it matches `js_*` runtime symbols — doesn't recognize as a source. Their fix is the identical shape and is covered by the unit tests above.

**`cargo test -p perry-codegen --no-fail-fast`** (full suite including the nightly-only `tests/*.rs` tier, which per-PR CI does not run): 6 pre-existing failures, identical on a clean `origin/main` baseline built in a separate worktree — `large_local_array_push_inbounds_store_emits_precise_slot_barrier`, `proven_buffer_and_typed_array_reads_are_numeric_operands`, `reassigned_typed_array_store_records_runtime_fallback`, `integer_modulo::i32_counter_mod_unsafe_or_nonliteral_divisors_keep_frem`, `typed_f64_receiver_method_clone_raw_loads_after_composed_guards`, `integer_arithmetic_array_push_omits_inbounds_layout_note_and_barrier` (tracked in #7708). No new failures. `--lib --bins` (the per-PR gate): 803 passed, 0 failed (797 pre-existing + 6 new).

`cargo fmt --all -- --check` and `cargo clippy -p perry-codegen --lib --bins --no-deps` are both clean on the touched files.

## Test plan

- [x] `cargo test -p perry-codegen --lib --bins --no-fail-fast` — 803 passed
- [x] `cargo test -p perry-codegen --no-fail-fast` (full suite) — same 6 pre-existing failures as clean `origin/main`, no new ones
- [x] New sabotage-confirmed unit tests in `temp_root_coverage::builtin_ctor`
- [x] `scripts/gc_root_dominance_check.py` before/after on a scoped probe, both lowerings (2→0)
- [x] `cargo clippy -p perry-codegen --lib --bins --no-deps`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when creating built-in objects whose arguments may trigger memory collection.
  * Preserved previously evaluated constructor arguments across allocations, including collections, typed data, regular expressions, streams, and asynchronous resources.
  * Fixed initialization handling for weak collections to prevent values from being lost or improperly collected.

* **Tests**
  * Added coverage for constructor argument preservation across supported lowering modes and allocation scenarios.


proggeramlug:
## Merging as v0.5.1416

**30 arms, not ~22** — the issue's own author flagged that estimate as unaudited, and it was low. Fixed by threading the caller's `RootedGroup` into `lower_builtin_new` from both `lower_new_impl_inner` call sites, plus three helpers that follow #7699's discipline exactly (**adopt as produced, never after the fact** — adopting late is how the store ends up after a call that allocates, which is #7192's shape). `CronJob` and `WeakMap`/`WeakSet` needed bespoke handling.

### The honest part, and the reason I'm confident in this

The static checker showed **0/0 both ways on the existing gate corpus** — because that corpus contains no source exercising these constructors, so it *cannot* show a reduction. A naive TS-literal probe also read 0/0, and the report explains why rather than shrugging: Phase-3 object-literal synthesis and the array-literal inline-bump allocator each bind their own unrelated shadow slot that happens to also cover a plain constructor argument end-to-end.

Only on switching to the project's own reproducer shape (`fresh(k)` / `"x"+churn(N)`, matching #6969's repro) did the checker reach the hazard: `--stale-registers` (shadow) **2 → 0**, and `--statepoints` (native, with the production `rewrite-statepoints-for-gc` rewrite) **2 unrooted → 0**.

That sequence is worth keeping. A checker reporting 0/0 is the single easiest way to conclude "no bug here", and three separate reasons for a false zero showed up before the real signal did.

Primary verification is therefore **6 sabotage-confirmed unit tests** in `temp_root_coverage/builtin_ctor.rs` — deliberately placed where per-PR `--lib` can see them, since `crates/*/tests/*.rs` is nightly/tag-only. Five positive assertions fail against pre-fix code and pass after; the paired negative gate passes on both, so it isn't a one-directional test.

`cargo test -p perry-codegen --no-fail-fast`: the same **6 pre-existing failures** on this branch and on a freshly built clean `origin/main` (#7708) — no new ones. `--lib --bins`: 803/803. Lint 19/19.

### Named, not silently skipped

The `extract_options_fields` arms — `Response`, `Request`, `Blob`, `File`, `Headers`, `ReadableStream`, `WritableStream`, `TransformStream` — share the hazard but use a structurally different per-field-loop shape, and are called out in both the PR body and the changelog. That's the right way to leave scope behind: an unfixed hazard that is written down is tractable; one that is quietly out of scope is not.

PR #7721, opened 2026-08-09
fix(gc): make the moving-loop poll default ON in the code, not just the doc (#7690, #7682)

proggeramlug:
## What this is

#7690 wrote the entire default-ON argument for the moving-loop poll into two doc comments — the runtime's `moving_loop_polls_enabled_from_env` and codegen's `moving_safepoint_polls_enabled` — and **changed neither body**. Both still matched `1|on|true`, i.e. default OFF. No test pinned the default in either direction, even though the runtime predicate had been factored out expressly to make it "unit-testable without touching process env". The runtime doc even asserted *"Codegen's `moving_safepoint_polls_enabled` mirrors this exactly — they MUST agree"*; they did agree, at the value the doc said they no longer held.

That is not a slower configuration, it is a different collector. Nursery pressure has exactly two precise collection points: the loop back-edge poll and the outermost microtask-pump boundary. With no poll emitted, a compute-only program reaches neither, so every nursery collection happened at the register-imprecise allocation point — where #7687 had just made it correctly non-moving. The shipped collector had **no nursery evacuation at all**, and the trigger fell back to whole-arena full collections.

`churn_alloc` runs **13 whole-arena full collections (0.477 s of pause)** on `main` where the same program at `a853135aa` ran **105 copying minors (0.016 s)**.

## Measurements

Pinned quiet M1 mini, best-of-3 interleaved, `PERRY_NO_AUTO_OPTIMIZE=1` with a pinned `PERRY_RUNTIME_DIR`, `a853135aa` binaries rerun back-to-back on the same host.

| bench | main `12e48edd6` | this | `a853135aa` | node | scriptc 0.0.22 |
|---|--:|--:|--:|--:|--:|
| tree | 5.10 | **1.63** | 6.00 | 0.45 | 4.80 |
| tree_wide | 7.26 | **2.12** | 12.38 | 0.89 | 7.01 |
| retain | 2.33 | **1.32** | 1.33 | 0.15 | 0.11 |
| churn | 1.00 | **0.46** | 0.66 | 0.16 | 0.75 |
| churn_alloc | 0.91 | **0.41** | 0.36 | 0.14 | 0.44 |
| push_cls | 0.89 | **0.40** | 0.34 | 0.14 | 0.45 |
| cycles | 0.29 | **0.19** | 0.95 | 0.07 | 0.31 |
| churn_read | 0.02 | 0.02 | 0.35 | 0.08 | 0.30 |
| deeplist | 0.03 | 0.31 | 1.14 | 0.09 | 0.22 |

`tree` total GC pause **4.107 s → 0.550 s**, max pause **266 ms → 16 ms**; `trace_worklist` falls from 2,877 ms out of the top six phases entirely. `tree_wide` 0.549 s of pause, 17 ms max. All 11 benchmark stdouts byte-identical to Node.

**Costs, measured rather than argued.** `deeplist` 0.03 → 0.31 and `retain1` 0.03 → 0.42: both keep their heap under the initial 64 MB threshold, so they previously ran *zero* collections and a moving nursery is pure added cost there. Both still beat `a853135aa` (1.14 / —). `push_num` 0.16 → 0.17.

## The earlier objection is discharged

Polls-default-ON was declined once because a poll at every back-edge deleted the #7480 element-shape fast clone — a call inside a clone whose admission rests on being call-free-by-construction does not slow it, it *removes* it. Step 4 of that work now refuses to emit a poll inside such a clone. `churn_read` measures **0.02 s with polls either way**.

## Also in this PR

- **`perf(gc)`**: `heap_payload_slot_selection` runs once per traced object per GC walk and computed `raw_numeric_object_slots` via a `SHAPE_LAYOUTS` hash lookup behind a TLS `RefCell` — for a counter that returns on its first line unless `PERRY_GC_LAYOUT_SCAN_TRACE` armed it. Same shape as #7702. Now computed only when the trace is armed. Second item on the same walk: `shape_shared_pointer_mask` cloned a whole `TypedLayoutDescriptor` to keep one of its two masks, allocating and freeing a throwaway `Vec` per traced wide object; it now borrows and clones only what it returns.
- **`test(gc)`**: four `runtime_roots` tests took no pacing guard and so inherited the default. `force_legacy_gc_pacing` is the *wrong* repair and the tests say so themselves — three carry an evacuation witness ("the minor did not evacuate, so nothing here was exercised") and legacy pacing routes to the budgeted stepper, which is deliberately non-moving. `force_alloc_point_minor_pacing` (polls OFF, scavenge ON) is the combination they were written against.

## Validation

- perry-runtime **1947 passed / 0 failed**; perry-codegen lib **796 / 0**.
- `PERRY_GC_VERIFY_MARK=1 PERRY_GC_VERIFY_EVACUATION=1` clean on six benches.
- `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1` at quarantine depths 8 and 800: **4,837 objects moved**, no fault, byte-identical output. Zeal's verdict line is the liveness proof that the new default is live rather than merely untripped: `forced_collections=2000005 copying_minors=2000005 moved_objects=8 loop_polls=2000000`.
- **Gap suite: in progress, not complete.** 20/511 at time of writing — 18 pass, 2 `parity_fail`, and both (`test_gap_2159_defineproperty_class_prototype`, `test_gap_2514_settracesigint`) are already in `test-parity/known_failures.json`. Zero new failures so far. The run is crawling because the dev host is under load from other work; I will post the final verdict on this PR before it should be merged. Please do not merge on the partial result.

## Three tests pin what was unpinned

`polls_default_is_on` and codegen's `moving_safepoint_poll_default::unset_emits_the_poll` each pin one half against the full spelling table — including the unrecognised-value arm, which is the one that silently changes meaning if the `matches!` is inverted back. `polls_default_matches_codegen_mirror` pins that the two crates agree; that disagreement is silent in both directions (polls nothing consumes, or a deferral nothing drains), so it needs its own assertion rather than two doc comments claiming they match.

## Unrelated finding, not fixed here

`crates/perry-codegen/tests/typed_shape_descriptors.rs::integer_arithmetic_array_push_omits_inbounds_layout_note_and_barrier` is **red on `main`** and it is not this change — it fails identically under `PERRY_GC_MOVING_LOOP_POLLS=0`. #7702 turned `typed_feedback_emission_enabled()` off by default, so `keep_guarded_numeric_push` is false for a canonical-raw-f64 value and the `apush.numeric_fast` block is never emitted. The code change is deliberate; the test was not updated. That file is under `crates/*/tests/`, which does not run per-PR, so it landed red. `push_num` 0.13 → 0.16 on both `main` and this branch is the matching wall-clock cost.

https://claude.ai/code/session_01KaVWLc4q8Ng9P7mWmbUZpf

## Summary by CodeRabbit

* **New Features**
  * Moving-loop garbage-collection polling is now enabled by default.
  * Polling can still be disabled with `0`, `off`, or `false`.

* **Bug Fixes**
  * Reduced unnecessary garbage-collection layout work and descriptor copying.
  * Improved allocation and evacuation test coverage across supported GC pacing modes.

* **Tests**
  * Added validation for default, enabled, disabled, and unrecognized polling settings.


proggeramlug:
## Audit — the change is right; two corrections, and I am respecting your "do not merge on the partial result"

**Not merging.** You asked for the gap verdict first, and 20/511 is not a verdict. Everything below is review, not a hold-up.

### The flip is real this time, and in both bodies

`matches!(v, Some("1")|Some("on")|Some("true"))` → `!matches!(v, Some("0")|Some("off")|Some("false"))`, in the runtime *and* in codegen, with the predicate extracted as a pure function so it can actually be tested. That is the thing #7690 claimed and did not do.

The framing is also right: this is not a slower configuration, it is a **different collector**. With no poll emitted, a compute-only program reaches neither precise collection point, so every nursery collection landed at the register-imprecise allocation point where #7687 had just made it correctly non-moving — leaving no nursery evacuation at all and a trigger falling back to whole-arena full collections. `churn_alloc`: 13 whole-arena fulls / 0.477 s of pause, against 105 copying minors / 0.016 s at `a853135aa`.

`tree` total pause **4.107 s → 0.550 s**, max **266 ms → 16 ms**, `trace_worklist` out of the top six phases. And the zeal verdict line is the liveness proof that the new default is live rather than merely untripped.

**The costs are stated, not buried** — `deeplist` 0.03 → 0.31 and `retain1` 0.03 → 0.42 — and your explanation is independently corroborated. I measured this today on #6978: a program freeing 36.7 MB against the 64 MB initial trigger runs **zero** automatic collections (0 diag lines), while the same program at 10× runs 12. So yes: those benches previously collected *never*, and a moving nursery is pure added cost there. Worth stating in the changelog as a general property rather than a per-bench excuse — any workload under the initial threshold pays this.

### 1. `polls_default_matches_codegen_mirror` pins a copy, not the agreement

```rust
// Mirrors `perry_codegen::stmt::loops::moving_safepoint_polls_enabled_from_env`.
fn codegen_mirror(value: Option<&str>) -> bool {
    !matches!(value, Some("0") | Some("off") | Some("false"))
}
```

That is a **re-implementation living in `perry-runtime`**. If codegen's predicate changes, this test still passes — it is asserting that the runtime matches a hardcoded literal, not that the two crates agree. Which is the same failure class as the doc comment this PR exists to fix: a claim of agreement that nothing binds.

Not your fault structurally — `perry-codegen/Cargo.toml` has no `perry-runtime` dependency (only `perry-hir`, `perry-dispatch`, `perry-api-manifest`), so neither crate can call the other's function.

**The repo already has the right idiom for exactly this**, in `gc_root_dominance_corpus.sh`'s `rs4gc_pass_string()`: single-source the value out of the Rust by parsing it, and treat "not found" as a hard error, precisely so a fourth copy can't drift. A ~15-line `scripts/` check that extracts both predicate bodies and asserts they are textually identical would make the agreement binding, and would fail loudly if either is reworded. Your own PR text argues the disagreement is silent in both directions — that is the argument for a checker, not a copy.

### 2. The `integer_arithmetic_array_push_...` attribution is wrong

You attribute it to #7702 turning `typed_feedback_emission_enabled()` off. It predates that.

I captured a full `cargo test -p perry-codegen` baseline on `main` at **v0.5.1402**, and that test is in it — **five versions before #7702 merged at v0.5.1407**. The audit of #7702 independently bisected the same test to `94ad78435`, its own parent, and found it already failing there.

Your *conclusion* stands — it is not caused by this PR, and it landed red because `crates/*/tests/` does not run per-PR — but the cause is something older. It is one of the 6 pre-existing failures I filed as **#7708**, and worth triaging there rather than as #7702 residue.

### Small thing

The four `runtime_roots` pacing-guard repairs look right, and the reasoning is the good kind: `force_legacy_gc_pacing` routes to the budgeted stepper, which is deliberately non-moving, so it would have satisfied the guard while defeating the three tests' own evacuation witnesses. `force_alloc_point_minor_pacing` is the combination they were written against.

Post the gap verdict and I will merge.

proggeramlug:
Addendum on the attribution point — there is stronger evidence than my baseline, and it settles it.

**#7494** already documents `integer_arithmetic_array_push_omits_inbounds_layout_note_and_barrier` as failing on `main` at **f05ae3b83**, alongside three siblings. That is well before #7702 existed. It also establishes two things worth having:

- the failure is **lowering-independent** — identical with and without `PERRY_RS4GC=0` — so it is not in the #7370 native-roots family;
- it **reproduces run alone**, so it is not order dependence either.

So the `typed_feedback_emission_enabled()` explanation is a plausible-looking mechanism for a failure that predates the change it is attributed to. Worth flagging because that is a genuinely easy trap here: #7702 *does* make `keep_guarded_numeric_push` false for a canonical-raw-f64 value, so the reasoning is locally sound — it just isn't why this test is red.

Your `push_num` 0.13 → 0.16 observation on **both** `main` and this branch is consistent with that: the cost is on main independently of your change.

Nothing here affects the PR. Everything else in my review stands, and I'm still holding for the gap verdict.

proggeramlug:
## Gap suite: complete — no regression attributable to this PR

Run on the pinned quiet M1 mini (`PERRY_SKIP_BUILD=1` against this branch's compiler and archives, node v26.5.1 matching `.node-version`). The dev host was too contended to finish — 65 s/test, an 8.8 h ETA — so the run was moved to the bench host at ~1.6 s/test.

**511 tests: 487 pass, 18 parity_fail, 6 crash.**

The harness reports 9 regressions against `gap_snapshot.json`. **All nine reproduce identically under `PERRY_GC_MOVING_LOOP_POLLS=0`** — same host, same tree, same binary, and since the harness compiles each test with that env set, both the codegen and runtime halves revert to the pre-change default. None is caused by this PR.

| regression | polls ON | polls OFF (pre-change default) | cause |
|---|---|---|---|
| `test_gap_gc_rest_argument_rooting` | parity_fail | parity_fail | #7684 `[gc-zeal]` line, see below |
| `test_gap_gc_same_module_call_argument_rooting` | parity_fail | parity_fail | same |
| `test_gap_zlib_3285_params` | parity_fail | parity_fail | known host-local zlib flake |
| `test_gap_http_overloads_3226plus` | crash | crash | tokio sandbox panic |
| `test_gap_net_connect_bound_value` | crash | crash | tokio sandbox panic |
| `test_gap_fetch_request_from_node_incoming_message` | crash | crash | tokio sandbox panic |
| `test_gap_http_client_no_redirect_follow` | crash | crash | tokio sandbox panic |
| `test_gap_http_req_async_iterator` | crash | crash | tokio sandbox panic |
| `test_gap_http_res_socket_writable_onfinished` | crash | crash | tokio sandbox panic |

### The two GC rooting tests actually PASS, and are evidence *for* this change

Their assertion is `bad 0`, and both sides print exactly that. The entire diff is a line on **stderr**, which the harness merges into stdout via `> $tmp 2>&1`:

```
--- node ---            --- perry ---
bad 0                   bad 0
                        [gc-zeal] forced_collections=9653 copying_minors=9653 moved_objects=115728 loop_polls=9648
```

That is #7684's zeal verdict line breaking parity for any test that enables zeal — a pre-existing defect in the verdict's output channel, not a rooting failure.

Read the numbers, though: these are the tests written specifically to catch rooting bugs, and under the **new** default they ran **9,653 and 19,253 copying minors, moving 115,728 and 231,348 objects**, and still reported `bad 0`. Under the old default they move nothing. This is the strongest correctness evidence in the PR, and it only exists because the flip turns the moving collector on.

### The five crashes are the host's tokio sandbox panic

```
thread '<unnamed>' panicked at tokio-1.53.1/src/net/tcp/listener.rs:304:22:
there is no reactor running, must be called from the context of a Tokio 1.x runtime
```

All are http/net/fetch socket tests. Unrelated to GC.

### Caveat, stated plainly

`gap_snapshot.json` was recorded on a different host, so part of the delta above (including 10 `node_fail -> parity_fail` status changes on npm-dependent tests like `dayjs`/`moment`/`cron`/`slugify`) is oracle-environment drift, not a Perry change. The load-bearing control is therefore **not** the snapshot but the same-host kill-switch A/B, which isolates this PR's variable and holds for every one of the nine.

One improvement also recorded: `test_gap_iterator_helpers_2874` parity_fail -> pass.

proggeramlug:
## Merging as v0.5.1418 — the gap verdict holds, and I checked its mechanism independently

**The methodology is the right one.** The load-bearing control is the same-host, same-binary kill-switch A/B, not the cross-host snapshot — and you say so explicitly, including that part of the snapshot delta (the 10 `node_fail → parity_fail` changes on npm-dependent tests) is oracle-environment drift rather than a Perry change. All nine reported regressions reproduce identically under `PERRY_GC_MOVING_LOOP_POLLS=0`, which isolates exactly the variable this PR changes.

I verified the `[gc-zeal]` explanation rather than taking it: the harness captures with `> "$node_tmp" 2>&1` (`run_parity_tests.sh:1261`), so **any** stderr line breaks byte parity, and the verdict is printed at `gc/zeal.rs:229`. So the two GC rooting tests are a pre-existing defect in that verdict's output *channel*, not a rooting failure. I'd hit the same thing independently earlier — it's worth its own fix, since it will keep breaking parity for any test that enables zeal.

**And the numbers in those two tests are the best evidence in the PR.** They are the tests written specifically to catch rooting bugs; under the new default they ran **9,653 and 19,253 copying minors, moving 115,728 and 231,348 objects**, and still reported `bad 0`. Under the old default they move nothing at all. A rooting test that never moves an object is not a passing rooting test — it is an untested one.

Everything from my earlier review stands: the flip is real in both bodies, the framing (a different collector, not a slower one) is correct, and the costs are stated rather than buried.

1. **`polls_default_matches_codegen_mirror` pins a copy.** It re-implements codegen's predicate inside `perry-runtime`, so if codegen's changes the test still passes. Structurally forced — neither crate depends on the other — but `rs4gc_pass_string()`'s idiom (parse the value out of the Rust, hard-error on "not found") would make the agreement binding. Given this PR exists because a doc comment claimed an agreement that had stopped holding, it's worth closing that loop properly.
2. The `[gc-zeal]` line should go to a channel the harness doesn't fold into stdout, or be suppressed when a parity run is detected.

Gates 19/19. Rebased onto v0.5.1417 (this branch was 5 commits behind after today's merges).

Issue #7143, opened 2026-07-31
repsel Phase 5a: the `delete` shape barrier is module-scoped, but a proven-`this` receiver is aliased across modules by construction


proggeramlug:
Noticed while establishing the soundness envelope for #7141 / #7142. **Not
introduced by either** — this is about the Phase 5a mechanism as shipped in
#6925. Filing because the reasoning that makes the freeze family a module-wide
kill applies verbatim to `delete`, and was not carried over.

## The asymmetry

`collectors/proven_this.rs` documents that a proven `this` differs from a
Phase 3b local in exactly one way that matters: the receiver is **owned by the
caller and aliased by construction**, so the analysis cannot enumerate stores to
it (`collectors/proven_this.rs:41-45`). That is why the phase added a separate
module-wide `freeze_barrier_sites` kill for write-containing clones — a Phase 3b
local needs no such kill because containment bounds its aliases, and a proven
`this` does.

`delete` has the same property and did not get the same treatment. It is in
`expr_is_shape_barrier` (`collectors/ptr_shape.rs:163-166` — `Expr::Delete`,
`Expr::ReflectDelete`), and `method_proven_this` does consult
`has_shape_barrier_sites()`. But that fact is computed **per module**:

```rust
pub fn collect_module_dispatch_facts(hir: &Module) -> ModuleDispatchFacts
```
(`collectors/scalar_method_dispatch.rs:181`)

So `delete inst.declaredField` in module B sets no barrier flag in module A, and
module A's `__pshape` clone keeps its bare fixed-slot loads.

## Why that is not merely theoretical

`delete` is the **only** operation that relocates a declared field's packed slot
index. `js_object_delete_field` reaches the keys-scan/compaction path for any
`GC_TYPE_OBJECT` with a keys array — which is what
`js_object_alloc_class_inline_keys` installs for class instances
(`crates/perry-runtime/src/object/alloc.rs:275`) — with no class-instance
tombstone branch (`delete_rest.rs:291-344`; `field_count` recomputed at `:380`).
On `class C { a; b; c }`, `delete inst.b` moves `c` from slot 2 to slot 1 and
leaves `class_id` untouched. A clone compiled against the pre-delete indices
then reads the wrong field.

## What currently prevents it

Both shipped routing sites are protected by something other than the barrier
scan:

- `method_direct.fast` — dominated by the class-id **+ keys-token** guard, and
  the keys array pointer changes on compaction.
[...]

Issue #7142, opened 2026-07-31
repsel Phase 5a: route the class-id dispatch tower to __pshape under an inline keys check (batch.ts: 4 guard calls + 3 by-name field calls per call)


proggeramlug:
Follow-up from #7141 (which fixed Phase 5a's *arm-ordering* defect: the
`{method}__pshape` clone had zero call sites because the typed-clone arms
shadowed the routing).

#7141 deliberately left the largest single win on the table, because taking it
needs a guard that does not exist yet. Filing it with the measurement so the
decision is on the record.

## The site

`benchmarks/app-patterns/kernels/batch.ts` is the workload `Ptr<Shape>` exists
for. Its hot method call is `r.rescore(1.5)`, inside `rows.map((r) => …)`. The
closure parameter `r` has no static class, so `receiver_class_name()` returns
`None`, `emit_guarded_direct_method_call` is never reached, and the call lowers
to the **class-id switch tower** in
`lower_call/property_get/dynamic_dispatch.rs` (`idispatch.caseN`, blocks built
around :294-311).

Inside `idispatch.case0` the receiver's exact dynamic class is already proven by
a runtime `class_id` compare, and it calls the guard-ridden public body.

## What routing it would buy

Measured on the emitted IR (perry-dev, aarch64), `Row::rescore`:

| body | IR lines | `js_typed_feedback_class_field_*_guard` | `js_object_get_field_by_name_f64` |
|---|---|---|---|
| `…__Row__rescore` (public) | 311 | 4 | 3 |
| `…__Row__rescore__pshape` | 95 | **0** | **0** |

So one routed call site deletes 4 opaque guard calls and 3 by-name field calls
per invocation, on a 40,000-iteration loop. `batch.ts` totals today are 32
field guards / 47 by-name gets across the module.

## Why #7141 did not do it

A `class_id` match is **not** sufficient for the clone's bare fixed-slot loads.
`delete inst.field` compacts the packed inline slots — `js_object_delete_field`
reaches the keys-scan/compaction path for any `GC_TYPE_OBJECT` carrying a keys
array, which is exactly what `js_object_alloc_class_inline_keys` installs
(`crates/perry-runtime/src/object/alloc.rs:275`), and there is no
"this is a class instance, tombstone instead" branch
(`delete_rest.rs:291-344`, `field_count` recomputed at `:380`). So on
`class C { a; b; c }`, `delete inst.b` moves `c` from slot 2 to slot 1 while
`class_id` is unchanged.

[...]

PR #7722, opened 2026-08-09
docs(repsel): pin why the delete barrier's per-module scope is sound for proven-this (#7143)

proggeramlug:
### repsel: document and pin why the `delete` shape barrier's per-module scope is sound for proven-`this` (#7143)

Investigated #7143 ("the Phase 5a `delete` shape barrier is module-scoped,
but a proven-`this` receiver is aliased across modules by construction").
Confirmed the asymmetry the issue describes: `ModuleDispatchFacts::has_shape_barrier_sites()`
is computed per module (`collect_module_dispatch_facts`), and Phase 5a's
proven-`this` admission (`collectors/proven_this.rs::method_proven_this`)
consults only its OWN module's copy — a `delete`/`Reflect.deleteProperty` on
a class instance in a module that imports the class, rather than declaring
it, sets no flag the declaring module's admission decision can see.

**No miscompile exists.** Built a two-module reproducer
(`test-files/test_issue_7143_delete_barrier_cross_module.ts` +
`test-files/fixtures/issue_7143_pkg/shared.ts`) matching the issue's own
suggested shape — module A declares `class C { a; b; c }` plus a method
reading `this.c`, admits a `$pshape` clone since it contains no `delete`
itself; module B holds an instance A handed it, deletes `b` off it (which
relocates `c`'s packed slot via `perry-runtime`'s keys-array compaction),
then calls back into module A, which dispatches `inst.readC()` on the
now-mutated object. The compiled binary's output matches
`node --experimental-strip-types` exactly (`3`, `3`, `3`), and the
`--trace llvm` IR shows why: EVERY routing site that can call a `$pshape`
clone independently re-derives soundness at the point it matters, rather
than trusting the per-module admission fact —

- `method_direct.fast` (`lower_call/method_override.rs`) sits behind
  `js_typed_feedback_method_direct_call_guard` / `js_method_direct_shape_guard`,
  whose contract includes a raw pointer compare of the receiver's live
  `keys_array` against the class's canonical keys token. `delete`'s only
  code path for a class instance with a keys array
  (`perry-runtime/src/object/delete_rest.rs::js_object_delete_field`, shared
  by `Reflect.deleteProperty`) always clones a FRESH keys array, so this
  compare can never pass on a post-delete receiver, from any module.
- The Phase 3b guard-free `Ptr<Shape>` receiver arm needs no runtime check:
  its containment proof (`collectors/ptr_shape.rs` rule 2) already rules out
  any alias to the object existing anywhere, so there is nothing for a
  cross-module `delete` to reach through.
- The #7142 class-id dispatch-tower case
  (`dynamic_dispatch.rs::emit_tower_pshape_call`) already carries its own
  explicit keys-token re-check, added specifically for this reason — its doc
  comment already cited #7143 by number.

Landed as a documentation + regression-test PR, not a bug fix: added a
"`delete` is aliased across modules by construction" section to
`collectors/proven_this.rs` stating this invariant explicitly (module-wide
barrier facts are a cost heuristic for Phase 5a, never the correctness
mechanism — a future 4th routing site must independently re-derive a
dominating runtime check or genuine containment), a cross-reference from
`collectors/ptr_shape.rs`'s module-wide barrier rule, a new
`guarded_pshape_call_site_is_preceded_by_a_keys_token_guard` IR ratchet in
`proven_this_routing_tests.rs` pinning the `method_direct.fast` guard
dominance (mirroring the existing tower ratchet), and the two-file
reproducer above as a permanent parity-suite fixture.

## Summary by CodeRabbit

* **Documentation**
  * Clarified safety guarantees for cross-module property deletion and receiver checks.
  * Added a changelog entry documenting the behavior and validation coverage.

* **Tests**
  * Added regression coverage for cross-module deletion and property access consistency.
  * Added checks confirming appropriate runtime guards are emitted.

* **Chores**
  * Incremented the project version to `0.5.1417`.


proggeramlug:
## Merging as v0.5.1417 — the answer is "no miscompile", and that is the finding

The asymmetry #7143 describes is **real**: `has_shape_barrier_sites()` is per-module, a proven-`this` receiver is aliased across modules by construction, and a `delete` in a module that never declares the class sets no flag the declaring module's Phase 5a admission can see.

It is **not** a miscompile, because in Phase 5a that fact is a **cost heuristic only**. Correctness comes from elsewhere: every routing site that can call a `$pshape` clone independently re-derives it via a runtime keys-token guard. The contrast with rule 2 in `ptr_shape.rs` — where the same per-module fact *is* the correctness argument, because containment holds there — is now written down in both files instead of being re-derived by whoever looks next.

**Why spend a test on a non-bug.** The distinction is invisible in the source: nothing at the Phase 5a call site says "this use is advisory". Someone tidying the two uses into a shared helper would silently make the advisory one load-bearing, or the load-bearing one advisory — and that edit would look like a cleanup.

`guarded_pshape_call_site_is_preceded_by_a_keys_token_guard` asserts every emitted `$pshape` call site is preceded by `js_typed_feedback_method_direct_call_guard` or `js_method_direct_shape_guard`, and **first asserts a `$pshape` call was emitted at all**, so it cannot pass by finding nothing to check.

I sabotage-verified it rather than trusting it: pointing the guard predicate at a non-existent symbol fails with

```
perry_method_pshape_guarded_ts__Counter__bump$pshape: no keys-token guard call
precedes it in `probe` — a post-`delete` receiver (deleted from through an alias
in another module, #7143) would reach this clone's stale fixed-slot loads unguarded
```

`cargo test -p perry-codegen --lib proven_this_routing`: 8 passed, 0 failed. Lint 19/19.

One incidental: the added module note pushed `ptr_shape.rs` to 2003 lines against the 2000-line cap. Condensed to 2000 rather than allowlisted — an allowlist entry for three lines of prose would be the wrong trade.

Issue #7217, opened 2026-08-01
GC: #7207's spread-accessor reproducer still SIGSEGVs on main under the evac_minor arm (allocation-point relocation, not the safepoint route)


proggeramlug:
## Summary

`test-files/test_gap_gc_spread_accessor_rooting.ts` — the reproducer #7207 added for #7200 — **still SIGSEGVs on current `main`** (`1679e22b4`) when the collection is forced at the *allocation point* rather than at a loop safepoint.

#7207 verified that fix on the `PERRY_GC_MOVING_LOOP_POLLS=1` route, where it is clean 5/5. It is not clean on the `evac_minor` arm env.

## Reproduction

```
$ export PERRY_RUNTIME_DIR=<main build>/release PERRY_NO_AUTO_OPTIMIZE=1
$ perry test-files/test_gap_gc_spread_accessor_rooting.ts -o /tmp/spread     # no compile-time env
$ PERRY_GC_HEAP_LIMIT=8 PERRY_GC_INCREMENTAL=0 PERRY_CONSERVATIVE_STACK_SCAN=off /tmp/spread
exit=139, no output
```

Compared against a build of PR #7216 on the same host under identical conditions: **byte-identical, `exit=139`, output md5 `d41d8cd9…` on both**. So this is `main`'s behaviour, not #7216's, and #7216's A/B is what surfaced it.

`test_gap_gc_static_block_this_rooting` (the #7201 reproducer) reports `bad 1` under the same env, also identically on both arms — the same story one notch quieter.

## Why the two routes differ

`gc_repsel_matrix.sh`'s own header states it: `default` and the polls route "relocate at a real safepoint (the JS stack has unwound, roots are precise by construction)", whereas the `%E%` arms "force it at the **register-imprecise allocation point**". #7207's fixes — the `RuntimeHandleScope` rooting inside `js_object_assign_one` and the codegen accumulator root — were reasoned about and verified against the safepoint route. An allocation-point collection inside the helper is a different window: it can fire between any two of the helper's own allocations, including ones inside `object_assign_set_string_key`'s interning and keys-array growth, where #7207 re-reads its handles only at the top of each key iteration.

## Caveat, stated because it changes what to do first

[...]

Issue #7211, opened 2026-08-01
codegen: ClassExprFresh leaves the class object unrooted across its own js_object_set_field_by_name stores


jdalton:
## Summary

`Expr::ClassExprFresh` (`crates/perry-codegen/src/expr/static_field_meta.rs:432`) leaves the freshly-allocated class object in a raw SSA register across the very `js_object_set_field_by_name` calls that install its named statics. When the static initializers are all *inert*, the existing `protect_handle` predicate decides no root is needed — but the field-store helper is itself a collection point, so the register can go stale regardless of what the initializers do.

This is the same class as #7154 / #7184 / #7192 / #7206, and it is currently **reported red by the gate that already exists on `main`** (`.github/workflows/gc-root-dominance.yml`, added in #7198). It has not blocked anything because that job is not in branch protection's required contexts — hazard 2 in CLAUDE.md's "four ways a gate can be unable to fail".

## The predicate gap

```rust
let protect_handle = !captured_args.is_empty()
    || !symbol_statics.is_empty()
    || !block_fns.is_empty()
    || super::temp_root::any_may_trigger_gc(ctx, named_statics.iter().map(|(_, v)| v));
```

Every disjunct asks whether *something the author supplies* can collect: a captured argument, a symbol static, a `static { … }` block, or an initializer expression. None asks whether the **lowering's own emitted calls** can collect. The loop below unconditionally emits one `js_object_set_field_by_name` per named static, and that helper performs the keys-array transition and can allocate.

So `class C { static tag = tag }` — one inert `LocalGet` initializer — takes `protect_handle == false` and emits:

```llvm
%r10 = call i64 @js_object_alloc(i32 1, i32 1)
call void @js_object_mark_class(i64 %r10)
call void @js_class_object_pin_parent(i64 %r10, i32 1)
call void @js_object_set_field_by_name(i64 %r10, i64 %r14, double %r11)  ; can collect
%r15 = or i64 %r10, 9222527611924643840                                  ; %r10 may be from-space
%r16 = bitcast i64 %r15 to double
store double %r16, ptr %r9
call void @js_shadow_slot_bind(i32 0, ptr %r9)                           ; roots the stale address
```

## Why `js_object_mark_class` does not save it

[...]

PR #7723, opened 2026-08-09
gc: no-move window for lazy intrinsic towers (#7251) + CI arm for ZEAL+VERIFY_EVACUATION (#7254)

### LINKED ISSUE #7251: GC: the two lazy intrinsic-tower builders need #7217's no-move window, and need a gate that can fail first
## Summary

#7217 established the invariant and fixed the one site that had a runtime witness:

> **A bootstrap that builds an IMMORTAL object graph through raw pointers held across its own allocations must run in a NO-MOVE WINDOW.**

`populate_global_this_builtins` now runs in a `GcSuppressScope`, gated by a `cargo-test`-visible unit test that arms a collection, shows the bootstrap does not service it, and shows the same armed request IS serviced by ordinary allocation once the window closes.

**Two sibling builders have the identical shape and are NOT covered**, because I could not build a gate for them that can fail. Filing rather than shipping ungated, per CLAUDE.md's knob-kill policy ("shipping a GC-trigger change with no test that can fail without it is the thing [it] exists to stop", quoted verbatim at `crates/perry-runtime/src/string/alloc.rs:49`).

## The two sites

| builder | the raw locals | also reachable ahead of the bootstrap from |
|---|---|---|
| `object/global_this/generator.rs::ensure_generator_intrinsics` → `build_generator_tower` | `*mut ObjectHeader` / `*mut ClosureHeader` ctor+proto locals threaded across every allocating install in the tower | `generator_function_proto_of`, `generator_function_prototype_of`, `generator_prototype_ptr` — i.e. any program that reflects on a generator without touching `globalThis` |
| `object/global_this/typed_array.rs::ensure_typed_array_intrinsic` | `ctor` and `proto`, held across ~a dozen `set_builtin_property_attrs` / `set_field` installs | `typedarray_props.rs:812` |

Both build an immortal graph (the towers hang off process-global intrinsic slots), so the same argument applies verbatim: a collection inside them frees nothing and can only move the holders out from under them.

When they are reached *from* `populate_global_this_builtins` they inherit its window and are safe today. The exposure is the lazy-ahead-of-bootstrap path only.

## Why #7217 could not gate them, measured

`populate_global_this_builtins` is gateable **behaviourally** — arm a pending collection, run it, assert nothing collected — because it allocates ~1.15 MB and therefore crosses several arena block boundaries, and `arena_alloc_gc` calls `gc_check_trigger()` on each. A tower is three orders of magnitude smaller and fits inside one block's tail, so it can reach **no trigger at all**:

* a first version of the gate arming a collection around `ensure_generator_intrinsics()` **passed with the `GcSuppressScope` deleted**;
* a second version pre-filling to a block boundary, and re-execing the test binary so the child process is the first to touch a tower (the slots are process-global `AtomicI64`s, so libtest's name ordering meant `array::tests`'s `js_get_global_this()` built them first and the test was measuring nothing) — **also passed with the scope deleted**;
* a third version recording `gc_is_suppressed()` from inside the builder under `#[cfg(test)]` reported *suppressed* even with the scope removed, which is unexplained and is itself a reason not to trust it as a gate. (Worth a look: something else may already be suppressing across part of the tower build — `symbol/get.rs` calls the manual `gc_suppress()`/`gc_unsuppress()` pair at three sites, and `well_known_symbol` is on this path.)

## What would close this

Either of:

1. **A workload witness.** A gap test that reflects on a generator (or touches a typed array) under enough allocation pressure to put a collection inside the tower build, run on the allocation-point arm (`PERRY_GC_HEAP_LIMIT=8 PERRY_GC_INCREMENTAL=0 PERRY_CONSERVATIVE_STACK_SCAN=off`) with `PERRY_GC_PROTECT_FROMSPACE=1`. If the quarantine faults inside `build_generator_tower`, the window is justified and the witness is its gate.
2. **A trigger the tower's size can reach.** `GcTriggerThresholdTestGuard::make_malloc_sweep_due()` arms the malloc-count trigger, which `gc_check_trigger` sees on every `gc_malloc` rather than once per arena block. If the tower's installs go through `gc_malloc` at all, that gives a lever a few-hundred-object build can actually trip.

Resolve the third bullet's anomaly first — if something already suppresses across these builds, the windows may be redundant rather than missing, and that is a better answer than adding them.

Refs #7217, #7196.

### LINKED ISSUE #7254: gc: PERRY_GC_ZEAL + PERRY_GC_VERIFY_EVACUATION reports a stale forwarded pointer in shadow-stack roots (10/10, pre-existing, no CI arm sets the pair)
`PERRY_GC_ZEAL=1` and `PERRY_GC_VERIFY_EVACUATION=1` are both shipped knobs, and **no CI arm sets them together over the representation corpus**. Setting them together finds something on the first file that was tried.

## Repro

`test_gap_repsel_p4a3_ptr_numarray`, compiled and run with back-edge safepoint polls, at `origin/main` = `64c1f56fb`:

```
PERRY_GC_MOVING_LOOP_POLLS=1 perry test-files/test_gap_repsel_p4a3_ptr_numarray.ts -o bin

PERRY_GC_MOVING_LOOP_POLLS=1 PERRY_GC_ZEAL=1 PERRY_GC_VERIFY_EVACUATION=1 ./bin
```

```
thread '<unnamed>' panicked at crates/perry-runtime/src/gc/verify.rs:59:5:
gc evacuation verification failed: stale forwarded pointer in shadow stack roots:
  slot=0x16d72ea48 old=0x7ffd0437f5914878 forwarded_to=0x7ffd0437f59148c8
```

**10/10, deterministic.** `old` is a `POINTER_TAG` NaN-box (`0x7ffd`); the object moved 0x50 bytes and a shadow-stack slot still holds the pre-move address after the rewrite pass ran.

Adding `PERRY_GC_PROTECT_FROMSPACE=1` instead of the verifier gives a hard fault at the use rather than at the check, also 10/10:

```
[gc-fromspace-protect] FAULT: signal 10 at 0x33a1e140001
  block=0x33a1e140000 +1 retired_bytes=144 retired_by_minor=#14072
  last-known object: user_ptr=0x33a1e140008 obj_type=1 size=144
```

## Why this is probably a *dead* slot, and why it still matters

Three measurements say the mutator is not reading it:

* the same binary under the same zeal **without** the verifier is byte-exact vs node 26.5.1, **10/10, across 14 373 evacuating minors and 9 485 copied objects**. A live stale read would not survive that;
* `[gc-fromspace-scan clean] … missing_rewrites=0 dangling=0 owners=0` on **every** cycle — `PERRY_GC_FROMSPACE_SCAN_ABORT=1` never fires;
* the failure surfaces only when the verifier or the from-space quarantine looks, never in program output.

So the most likely reading is a shadow-stack slot whose value is no longer live but was not rewritten — the verifier is stricter than the collector needs to be. That is still worth resolving, in one of two directions, because the current state makes an instrument unusable:

1. if the slot is genuinely dead, the verifier needs to know that (or the frame needs to clear it), otherwise **`PERRY_GC_VERIFY_EVACUATION=1` cannot be combined with zeal on real code** — and zeal is precisely the configuration you reach for when hunting a rooting bug;
2. if it is *not* dead, this is a missed rewrite on the shadow stack, which is the highest-severity class this campaign has.

Note the adjacent, opposite-signed finding in #7035 (the verifier reporting *clean* on a demonstrable missing rewrite because it only self-checks the rewrite pass's own enumeration). A verifier with both a false-negative mode and a false-positive mode is not yet a trustworthy oracle.

## Scope, honestly

* **Pre-existing, not a regression from the recent GC merges.** The identical panic reproduces 3/3 on a build of `c9cd73ba5` (before #7239/#7242/#7243/#7249/#7250/#7252). It is *not* what #7194 was about — that was an output mismatch fixed by #7249, and this survives the fix.
* **Small sample.** Four corpus files tried under this arm: `test_gap_repsel_p4a3_ptr_numarray` fails 10/10; `test_gap_repsel_canonical_i32`, `test_gap_repsel_p4a3_numarray_growth` and `test_gap_repsel_ptr_shape_locals` are clean 0/3. Nobody has swept the other 45.
* **The knob-pair has no CI arm.** `scripts/gc_instrument_smoke.sh` exercises zeal against its own synthetic fixture; `gc_repsel_matrix.sh` has no zeal arm at all. Under CLAUDE.md's GC knob kill-policy that is a configuration nobody has verified — which is how a 10/10 deterministic reproducer sat unseen in a corpus file that the matrix runs twenty-one different ways.

Suggested first step: sweep the remaining 45 corpus files under this arm to size the population before deciding between (1) and (2) — the ratio of clean to dirty files is what says whether this is a verifier-strictness bug or a rewrite bug.

proggeramlug:
## Summary

Two GC issues, both "a real hazard exists and nothing exercises it" — closes #7251, closes #7254.

### #7251 — no-move window for the two lazy intrinsic-tower builders

`build_generator_tower` (`object/global_this/generator.rs`) and `ensure_typed_array_intrinsic` (`object/global_this/typed_array.rs`) build the same shape of immortal object graph #7217 fixed for `populate_global_this_builtins`: raw `*mut ObjectHeader`/`*mut ClosureHeader` locals threaded across a dozen-plus allocating installs, with no root the collector knows about until the final `AtomicI64::store`. Both now open a `GcSuppressScope` for the whole build, matching #7217's fix exactly.

The issue's own blocker was the gate, not the fix — three prior attempts all passed with the window deleted:

1. **Arm-and-let-ordinary-allocation-reach-it** (the `populate_global_this_builtins` gate's own approach) never reaches `gc_check_trigger()` at all: a tower is three orders of magnitude smaller than an arena block. Fixed by forcing the current block (very nearly) full before calling the builder, so its first allocation unconditionally takes the slow path.
2. **Re-exec the test binary** still isn't first-touch, because the six intrinsic-tower statics (`TYPED_ARRAY_INTRINSIC_PTR`, `GENERATOR_FUNCTION_INTRINSIC_PTR`, …) were plain process-global `AtomicI64`s — built once per *process*, so whichever test ran first on that binary poisoned every later test's "never built" precondition. Fixed by converting them to `per_test_global!` (the mechanism #7672 built for exactly this ordering hazard): each libtest thread gets its own zeroed instance, and non-test builds expand to the identical bare `static`.
3. **Recording `gc_is_suppressed()` from inside the builder** read `true` even with the scope removed, because it was leaking attempt 2's ordering hazard in a different shape (an earlier test's `populate_global_this_builtins` call opens its own outer window around both builders). Sidestepped by calling the tower builders directly, never through the bootstrap.
4. **(Found while building this PR.)** A first draft armed the ArenaBytes/malloc byte-threshold trigger under `force_legacy_gc_pacing`, and it also passed with the fix deleted — vacuously. Under that pacing, an ArenaBytes trigger being "due" only starts a *budgeted* cycle that needs an explicit host-safepoint pump to advance; nothing pumps one inline, so `gc_collection_count()` never moves whether or not the builder is suppressed. Fixed by arming `GC_OLD_RECLAIM_PENDING` instead — the one branch `gc_check_trigger` services synchronously regardless of pacing, and the exact lever `global_bootstrap.rs`'s own #7217 gate uses.

New test: `crates/perry-runtime/src/gc/tests/lazy_intrinsic_towers.rs`, two tests (`generator_tower_runs_in_a_no_move_window`, `typed_array_intrinsic_tower_runs_in_a_no_move_window`), each asserting both halves (CLAUDE.md's fourth way a gate can't fail): no collection ran during the builder call, AND the same armed request is serviced once the window closes.

**Sabotage, run by hand:** commenting out either `GcSuppressScope::new()` line turns its test red (`a collection ran inside …`), confirmed on this branch after the pacing fix (attempt 4's false start would *not* have caught it — verified that too, by running the vacuous first draft against the sabotaged tree and watching it pass).

### #7254 — PERRY_GC_ZEAL + PERRY_GC_VERIFY_EVACUATION, exercised in CI for the first time

**Reproduced, 3/3, on current `main` (post-#7719).** Repro exactly as filed:

```
PERRY_GC_MOVING_LOOP_POLLS=1 perry test-files/test_gap_repsel_p4a3_ptr_numarray.ts -o bin
PERRY_GC_MOVING_LOOP_POLLS=1 PERRY_GC_ZEAL=1 PERRY_GC_VERIFY_EVACUATION=1 ./bin
```

Liveness proven via `PERRY_GC_DIAG=1`, not merely a clean exit: 14,072 copying-minor runs, 10,775 objects actually copied across the run. The instrument was genuinely armed.

**New finding, not in the issue: the panic surface has moved.** It now reads `"stale forwarded pointer in native stack-map roots"`, not `"shadow stack roots"`. `gc/roots/stack_maps.rs`'s `visit_stack_map_root_slots` is the statepoint/stack-map native root walker (`docs/src/internals/rfc-rooting-by-construction.md`'s layer 2), which is live on this aarch64 host. So on current `main` this is a layer-2 (emitted-code-liveness) defect, not layer-3 (hand-written runtime Rust) as the issue's own comment classified it — the classification predates the statepoint route becoming reachable here. Filing this as new evidence on #7254 rather than silently changing the issue's framing.

**Sized the population (the issue's own suggested first step), all 59 `test-parity/gc_repsel_corpus.txt` files under the exact pairing:**

| result | count | detail |
|---|---|---|
| PASS | 20 | byte-correct, no panic |
| INERT (zeal's own exit-70 verdict) | 3 | `test_gap_specabi_{reassign,recursion_escape,view_detach}` — zeal correctly self-reported no collection ran; not a defect |
| **FAIL — confirmed panic** | **3** | `test_gap_repsel_p4a_inline_tiers`, `test_gap_repsel_p4a3_ptr_numarray` (the filed repro), `test_gap_repsel_element_shape_loop_clone` — all three the identical `native stack-map roots` signature |
| TIMEOUT (40s budget) | 33 | see caveat below |

**Caveat on the 33 timeouts, stated plainly rather than folded into a claim:** this sweep ran under sustained host load of 30–55 (20 concurrent users, confirmed via `uptime` throughout, not a transient spike) and competed with this same PR's own concurrent rebuilds. I cannot currently separate "host contention" from "ZEAL forcing an evacuating minor at every back-edge poll is just legitimately much more expensive than any other matrix arm on tight-loop reproducers" from "a genuine second defect" for those 33 files. The concentration is suspicious — it is almost the entire `test_gap_gc_*` rooting-reproducer corpus specifically — so it is worth a dedicated follow-up on a quiet host with a longer per-file budget, but asserting more than that here would be exactly the overclaim CLAUDE.md's knob-kill policy warns about.

**What's fixed vs. gated, and why:** the panic itself is NOT fixed in this PR. This is a statepoint/native-root-walker liveness bug (layer 2), the same class CLAUDE.md documents as needing careful, disassembly-driven campaigns (the `#7154`/`#7341` family) — not something to rush without conclusive root-causing, given the project's own cautionary precedent (#7211: an author *actively thinking about rooting* still shipped a wrong four-clause predicate). A wrong fix here risks masking a real bug rather than closing it. **A gate that pins the defect is shipped instead**, per the task's own explicit allowance: "a gate with no fix is still better than today's nothing."

**Why not `gc_repsel_matrix.sh`:** a `zeal_verify` arm registered there joins *every* corpus file via `--arms all` (nightly/push), and the 90-minute job budget cannot responsibly absorb an unsized population of multi-minute-plus cells discovered under contested conditions. Extended `scripts/gc_instrument_smoke.sh` instead (already wired into `gc-stress`'s required-ish per-PR step) with a new arm 5, bounded and fast:

- **5a** — the pairing on the script's existing small fixture: correct output, exit 0, AND live (`copied_objects>0` under the verifier) — proves the pairing is non-vacuous and produces no false positive on known-good code.
- **5b** — a **pinned regression witness** against `test_gap_repsel_p4a3_ptr_numarray`: asserts it panics with `stale forwarded pointer`, and fails loudly (with instructions) if that ever stops being true — either because the defect got fixed (delete the block, promote the file) or because the failure silently changed shape (needs a fresh look, not silence).

Verified locally end-to-end (`./scripts/gc_instrument_smoke.sh <perry-dev-binary>`): all 5 arms pass, including 5a (3,894 objects copied, live) and 5b (reproduces `#7254` as pinned).

## What this PR does NOT do

- Does not fix #7254's stale-forwarded-pointer defect — gates it as a pinned, tracked regression instead, per the reasoning above.
- Does not promote any new check to a required branch-protection context — that stays a maintainer action.
- Does not add a new CI workflow or touch `scripts/gc_gate_wiring_check.py`'s `GATES` registry — both additions extend scripts already invoked by the existing `gc-stress` job (`test.yml`), so no new registration is needed.
- Does not investigate the 33-file timeout population beyond sizing it — flagged as follow-up work needing a quiet host.

## Test plan

- [x] `crates/perry-runtime/src/gc/tests/lazy_intrinsic_towers.rs`: 2/2 passing on this branch (rebased onto `origin/main` @ 72ef47a07)
- [x] Sabotage: both tests turn red when their `GcSuppressScope` is removed (verified twice — once before the pacing-mechanism fix, which incorrectly stayed green, and once after, which correctly went red)
- [x] `./scripts/gc_instrument_smoke.sh <perry-dev binary>`: all 5 arms pass locally, including the new #7254 pairing (5a non-vacuous + correct, 5b pins the known reproducer)
- [x] `bash -n` + `shellcheck` clean on the modified script
- [x] `cargo check -p perry-runtime -p perry-runtime-static -p perry-stdlib -p perry-stdlib-static --tests`: 0 errors
- [x] `cargo fmt --all -- --check`: clean
- [ ] Full `cargo test --release --workspace` (not run in this session — scoped `--lib` runs above cover the changed surface; left for CI)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved garbage-collection safety while initializing generator and typed-array functionality.
  * Prevented unsafe object relocation during intrinsic setup.
  * Added safeguards and regression coverage for stale pointer detection under aggressive GC verification.

* **Tests**
  * Added coverage for lazy intrinsic initialization and collection-trigger behavior.
  * Expanded GC smoke testing with evacuation verification scenarios.

* **Chores**
  * Updated the project version to 0.5.1419.
  * Documented the latest GC-related fixes and test coverage.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1418
+**Current Version:** 0.5.1419
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1418"
+version = "0.5.1419"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "perry-ffi",
  "scraper",
@@ -5813,7 +5813,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-commander"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5821,7 +5821,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cron"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "chrono",
  "cron",
@@ -5831,7 +5831,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dayjs"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "chrono",
  "perry-ffi",
@@ -5839,7 +5839,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-decimal"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "perry-ffi",
  "rust_decimal",
@@ -5847,7 +5847,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dotenv"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "perry-ffi",
  "serde_json",
@@ -5855,7 +5855,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ethers"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "perry-ffi",
  "rand 0.10.1",
@@ -5863,7 +5863,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-events"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5871,14 +5871,14 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-exponential-backoff"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "perry-ffi",
 ]
 
 [[package]]
 name = "perry-ext-fastify"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "bytes",
  "http-body-util",
@@ -5896,7 +5896,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-fetch"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "bytes",
  "lazy_static",
@@ -5909,7 +5909,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-http"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "bytes",
  "h2",
@@ -5933,7 +5933,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ioredis"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "lazy_static",
  "perry-ffi",
@@ -5943,7 +5943,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-jsonwebtoken"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "base64",
  "jsonwebtoken",
@@ -5954,7 +5954,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-lru-cache"
-version = "0.5.1418"
+version = "0.5.1419"
 dependencies = [
  "lru",
  "perry-ffi",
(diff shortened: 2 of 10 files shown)


proggeramlug:
## Merging as v0.5.1419 — and the best thing here is the gate you threw away

**You caught your own vacuous gate.** The first draft armed the ArenaBytes trigger under `force_legacy_gc_pacing` and **passed with the fix deleted** — because that pacing never synchronously collects, so suppression could not matter either way. You noticed, discarded it, and rebuilt on the `GC_OLD_RECLAIM_PENDING` lever (`global_bootstrap.rs`'s own #7217 mechanism), plus forcing an arena block full so the tiny tower's first allocation reaches `gc_check_trigger()` at all. Then re-verified in both directions: fix present → 2/2 pass, fix removed → 2/2 fail at the invariant assertion.

That is the single most common way a gate ships unable to fail, and it is the third time today a *first* draft of a check turned out to be satisfied by "nothing happened". Discarding it rather than shipping it is the right instinct.

### #7251 — fixed and gated

`GcSuppressScope` added to `build_generator_tower` and `ensure_typed_array_intrinsic`, matching #7217. You also fixed the ordering hazard behind **two of the three prior failed attempts**: the six tower `AtomicI64` statics are now `per_test_global!` (#7672's mechanism), so each libtest thread gets a guaranteed-fresh "never built" tower regardless of run order. That is why the earlier attempts kept failing for reasons unrelated to the bug.

### #7254 — reproduced, gated rather than fixed, and the issue's own label is stale

Reproduced 3/3, with liveness proven rather than assumed: **14,072 copying-minor runs, 10,775 objects copied** under `PERRY_GC_DIAG=1`.

The finding that isn't in the issue: **the panic surface has moved.** It now reads `"stale forwarded pointer in native stack-map roots"`, not shadow-stack roots — `gc/roots/stack_maps.rs`'s statepoint walker is live on this host, so the issue's "layer 3" classification is out of date. Anyone picking this up from the issue text alone would have looked in the wrong place.

**Gating rather than fixing is the right call here** — this is the #7154/#7341 statepoint-liveness class, and #7211 is a direct precedent against a rushed fix. Arm 5a proves the pairing is non-vacuous and false-positive-free on known-good code; 5b is a pinned regression witness against the exact reproducer that fails loudly if the panic ever *stops* (fixed → promote it; changed shape → investigate). All 5 arms verified locally.

### The restraint I want to note

You sized the population across all 59 corpus files — 20 PASS, 3 correctly INERT, 3 confirmed panics — and then **refused to treat the 33 timeouts as a finding**, because the sweep ran under sustained host load of 30–55 and timeout-vs-genuine-cost-vs-noise is undecidable from that data. Saying so plainly, and keeping the gate out of `--arms all` until the population is sized on a quiet host, is worth more than a number that would have been quoted for months.

Gates 19/19.

PR #7724, opened 2026-08-09
gc: fix root-dominance phi false positives + 3 of 5 real hits, lower --max-unrooted to 2 (#7664)

proggeramlug:
## Summary

Closes the checker half of #7664 and two of the four real hits underneath it. Re-verification found the population was **5 real hits, not 4** (an off-by-one against the prior triage — see below), and this PR fixes 3 of them plus both checker false positives, leaving 2 open as a follow-up rather than rushed.

### What I verified vs. what I'm inferring

- **Checker fix (phi false positives)**: verified two ways. `--self-test` (deterministic, no build needed) passes, including two new sabotage-tested fixtures (`phi_safe_edge` must report 0, `phi_hazard_edge` — byte-identical except the safepoint moves onto the tainted edge — must report exactly 1; each was confirmed to fail when the corresponding fix is reverted). Also ran against the real native corpus (149 modules) on a build that included this fix: the four `unmasked` phi false positives from the prior snapshot are gone, nothing else changed.
- **`static_dispatch.rs` fix**: verified against the real corpus (0 hits for `test_gap_static_method_value_name_collision`, down from 1) and a functional smoke test (`(Lexer as any).lex(...)` / `Parser.parse` still produce correct output, no crash).
- **`builtin.rs` (the other 2 `unrooted:global` hits)**: I wrote a fix for these independently, then discovered `origin/main` had moved to include #7719, which fixes the *identical* shape (module-global constructor argument held across a sibling argument's allocation) across a superset of `lower_call/builtin.rs`'s arms via the same `RootedGroup` mechanism. I dropped my version in favor of it and rebased cleanly (no conflicts). **I have not re-run the corpus against the exact post-rebase commit** — the corpus run showing 2 residual hits (both `unrooted:capture`) used a build with my own now-dropped `builtin.rs` fix, not #7719's. Since #7719 covers the same shape with the same mechanism, I expect the same reading post-rebase, but this PR's own `gc-root-dominance-statepoints` CI run is the actual confirmation, not something I'm asserting here.
- **The 2 remaining `unrooted:capture` hits**: diagnosed in detail (see the workflow comment and the changelog fragment), not fixed. `js_closure_get_capture_bits`'s return value is never re-entered into either the RS4GC-tracked domain or a temp root by the generic "read a captured value" call sites — narrower than the original triage's "signature/ABI change" guess, but still a real slice of work (`root_reload.rs`'s `Facts` only models loads as reloadable sources today, not calls). Filing as a follow-up issue rather than rushing it.

### The checker fix

`scripts/gc_root_dominance_check.py`'s `chain` (the untracked cast-closure a stale use is searched in) treated `phi` as unconditionally transparent — one tainted incoming edge blanket-tainted the phi's *result*, and a downstream use of that result was checked against *any* CFG path between source and use. That's sound for an ordinary register but not a phi, whose dynamic value depends on which edge was actually taken. All four false positives were the same `&&`/`||` short-circuit join: the tainted edge never crosses a safepoint, the *other* edge does.

`_cast_closure` gained `phi_all_edges`: a phi joins `chain` only once every incoming edge is independently in it (a worklist retry on each operand's own arrival, so admission order doesn't matter). That deliberately gives up the single-tainted-edge-with-its-own-safepoint case; `_phi_edge_hazard` covers that separately, checking each edge's window against its own predecessor's terminator instead of the join.

### Budget

`gc-root-dominance-statepoints`' `--max-unrooted` goes **8 → 2**. The full accounting (all 9 hits a fresh corpus actually reads before this PR, why the prior "8" was already stale, and the diagnosis for both remaining hits) is in the workflow comment and the changelog fragment.

## Test plan

- [x] `python3 scripts/gc_root_dominance_check.py --self-test` — clean, including two new fixtures each confirmed via a sabotaged copy of the checker to still be able to fail
- [x] `cargo check -p perry-codegen` — clean
- [x] Native corpus (149 modules) + checker, `--statepoints --moving-only`: 4 phi false positives gone, `test_gap_static_method_value_name_collision` hit gone, 2 `unrooted:capture` hits remain exactly as diagnosed (run predates the #7719 rebase — see above)
- [x] Functional smoke test of the `static_dispatch.rs` change (`test_gap_static_method_value_name_collision.ts`, no oracle diff run — output shape looked correct, not byte-compared against node)
- [ ] Fresh corpus run against the exact post-rebase commit (not done — relying on this PR's own `gc-root-dominance-statepoints` CI run)
- [ ] `--lowering shadow` corpus (attempted twice; both runs hit unrelated compile flakiness on a heavily-loaded, briefly disk-full shared box — did not get a clean read before wrap-up)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved static method calls involving dynamically created classes and factory-produced receivers.
  - Fixed issues where receiver values could be lost during argument evaluation or memory allocation.
  - Reduced false alarms in native garbage-collection safety checks while preserving detection of genuine hazards.

- **Documentation**
  - Updated release documentation and troubleshooting details for native safety checks.

- **Chores**
  - Updated the project version to 0.5.1420.

### DIFF
--- a/.github/workflows/gc-root-dominance.yml
+++ b/.github/workflows/gc-root-dominance.yml
@@ -517,49 +517,28 @@ jobs:
           # 21 at #7663. #7664 fixed 14 of them in `root_reload.rs` and
           # `lower_call/new.rs` — the whole `strhandle` population (10), the
           # `js_new_target_get` save/restore (1), and 3 of the 7 unmasked
-          # receivers — so the budget is 7. What remains, and why each is its
-          # own slice rather than a widening of the same fix:
+          # receivers — so the budget is 7. #7616 (widening POLL_CAPABLE_RUNTIME)
+          # and #7690 (rooting the spread argument-bundle accumulator) moved it
+          # 7 -> 11 -> 8, tracked in the two changelog fragments for those PRs.
           #
-          #   4  unmasked, all PHI-MEDIATED. The stale value reaches its use
-          #      through a `phi`, and `root_reload` cannot insert above a phi;
-          #      the reload has to go in the PREDECESSOR, on the edge, which is
-          #      a different insertion model.
-          #   2  global (`@perry_global_*`). NOT reloadable: a module-level
-          #      variable is one the program assigns, so a re-read can observe
-          #      a later assignment instead of the value the call was given
-          #      (`operand_needs_root`). That population needs ROOTING, and
-          #      `a_module_global_is_not_a_reload_source` in root_reload.rs
-          #      pins the distinction so it cannot be widened away by accident.
-          #   1  capture, a `js_closure_get_capture_bits` read held across
-          #      `js_number_coerce`.
+          # ★ RE-VERIFIED AT #7664's SECOND PASS: the 8 above was already stale
+          # by the time this ran. A fresh corpus read 9, not 8 — a static-method
+          # receiver hazard in `test_gap_static_method_value_name_collision`
+          # (added after the "8" snapshot, by the #7689/#7691 fix that armed a
+          # receiver-sensitive static `this`) had joined the population without
+          # anyone lowering — er, RAISING — the budget to match. That in itself
+          # is the thing CLAUDE.md's hazard-4 corollary warns about: a budget
+          # nobody re-measures silently absorbs the next hazard.
           #
-          # 11 at #7616, and the four new ones are the POINT of that change
-          # rather than a regression: widening POLL_CAPABLE_RUNTIME by the 77
-          # symbols `--audit-poll-reach` found makes windows MOVING that the
-          # filter previously dropped. Measured on this corpus, 7 -> 11, with
-          # no hit disappearing:
-          #
-          #   3  test_gap_array_splice_spread::main, `unrooted:alloc`. A fresh
-          #      array held in a raw i64 across `js_array_like_to_array` and
-          #      consumed by `js_array_concat` -- #7453's shape exactly, in the
-          #      spread lowering instead of the URL one.
-          #   1  test_gap_class_expr_dynamic_parent_ctor, `unrooted:capture`.
-          #      The same `js_closure_get_capture_bits` population as the
-          #      residual above, newly visible because `js_new_function_construct`
-          #      is now classified as a mover.
-          #
-          # ── 8 here. `expr/call_spread.rs` roots the argument-bundle
-          # accumulator, which closes all three `unrooted:alloc` hits. Measured
-          # on this corpus, same binary both arms, 11 -> 8 with the other eight
-          # byte-identical and `stale` still 0.
-          #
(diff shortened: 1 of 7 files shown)


proggeramlug:
## Merging as v0.5.1420 — and I verified the one thing you flagged you couldn't

**The headline finding is that the prior triage's own numbers were stale.** "8 = 4 phi false positives + 4 real" was itself out of date: a fresh native corpus read **9**, because a fifth real hit (`test_gap_static_method_value_name_collision`) joined after #7691 landed and nobody re-measured the budget. A budget that sits above reality silently absorbs the next real hazard — that is the same defect as `MIN_COMPILED=90` against a 131-source corpus (#7706), one layer up.

Dropping your own `builtin.rs` fix in favour of #7719's — which landed mid-flight, same shape, same `RootedGroup` mechanism, superset of arms — was the right call rather than fighting the rebase.

### I re-ran the confirmation you explicitly said you had not

You noted your "2 residual hits" came from a build containing your own now-dropped `builtin.rs` fix, not #7719's, and that you were relying on CI. Fair, and worth not leaving to chance since the budget change is the merge risk. On the rebased branch with a fresh `--release` build:

```
corpus (native): 131/131 sources compiled, 0 skipped, 151 .ll files
  statepoints: 9624   non-empty live bundles: 5737
gc_root_dominance_check.py --statepoints --moving-only --max-unrooted 2
  → within budget: unrooted 2 <= 2      exit 0
```

**The budget of 2 holds against #7719's fix, not just yours.**

Two things about my own run worth recording, since both are traps:

- My first corpus attempt compiled **52/131 with 79 skipped**, because I had the binary but no pinned `PERRY_RUNTIME_DIR`. The checker then reported *272* unrooted from a corpus two-thirds missing. **#7706's new ratchet caught it** — the corpus exited 1 instead of quietly producing a partial answer, which is exactly what that change was for.
- I first ran bare `--statepoints` (630 hits) rather than the gate's actual `--statepoints --moving-only`. Your workflow comment documents both numbers, which is why I noticed rather than filing a false alarm.

### Scope, honestly stated

The 2 remaining `unrooted:capture` hits are split into **#7725**, and the diagnosis there is the useful part: `js_closure_get_capture_bits`'s **return value** is never re-entered into the RS4GC-tracked domain or a temp root, unlike `%this_closure` itself. The real fix needs `root_reload.rs`'s `Facts` to model a **call** as reloadable, not just a load — which is a genuine gap in the repair machinery, not a missing store.

The checker fix (`_cast_closure`'s `phi_all_edges`) is the right shape too: `chain`'s blanket phi propagation was reporting the *untainted* edge's safepoint against the *tainted* edge's value. A checker that over-reports trains people to raise the budget, which is how this gate would have died.

Shadow-lowering read is noted as unchecked in your test plan rather than glossed — correct, given the box was briefly disk-full and at load 30-55.

Gates 19/19.

Issue #6668, opened 2026-07-19
Spread call of a native-module method silently no-ops (js_closure_call_apply_with_spread bypasses bound-native dispatch)


proggeramlug:
## Summary

A **spread call** of a native-module method silently no-ops — no dispatch, no throw, callback never invoked, returns `undefined`:

```js
const crypto = require("crypto");
const args = ["sha256", ikm, salt, info, 64];

crypto.hkdf("sha256", ikm, salt, info, 64, cb); // ✓ works (direct)
const h = crypto.hkdf; h(...allArgs, cb);        // ✓ works (value call)
crypto.hkdf(...args, cb);                        // ✗ SILENT NO-OP under perry
```

The spread lowering routes through `js_closure_call_apply_with_spread` → `js_closure_call_array`, which treats the callee as a plain closure; a bound-native-callable box has no valid closure `func_ptr`, so the per-arity dispatch degrades to returning `undefined` without ever reaching the module dispatcher — and without an error. (Depending on the callee shape, a sibling route through `js_closure_unbox_callee_checked` throws `value is not a function` instead.)

Found alongside #6667 while root-causing the Auth.js credentials-login failure (jose's node-HKDF wrapper is exactly `Z.hkdf(...args, cb)`); with #6667 fixed, this becomes the next blocker on that path.

## Suggested direction

`js_closure_call_apply_with_spread` (and the checked-unbox spread route) should detect a non-plain-closure callee and delegate to `js_native_call_value` with the flattened args — that dispatcher already owns the bound-native/proxy/class-object arms. Silent `undefined` on a failed dispatch is the worst outcome; if delegation is out of scope, at minimum throw.

Observed on main @ 5eb2bc495, macOS arm64.

PR #7726, opened 2026-08-09
fix(hir): route node-core module spread calls through the variadic dispatch

### LINKED ISSUE #7720: Native Node module spread calls such as `path.join(...args)` fail despite spread calls being documented as supported
## Description

Perry documents the spread operator in calls as supported, and `node:path` / `path.join` as supported, but spreading a valid string array into `path.join()` fails at runtime.

I initially encountered this in a larger application, but it reduces to a very small standalone repro.

## Minimal reproduction

`repro.ts`:

```ts
import path from 'node:path';

console.log('direct:', path.join('/tmp/x', 'project.json'));

const parts = ['/tmp/x', 'project.json'];
console.log('spread:', path.join(...parts));
```

Using Perry `0.5.1220`:

```sh
pnpm add -D @perryts/perry@0.5.1220
pnpm exec perry compile repro.ts --target linux --output repro
./repro
```

## Actual behavior

The direct call succeeds:

```text
direct: /tmp/x/project.json
```

The equivalent spread call then throws:

```text
TypeError: The "path" argument must be of type string.
    at <anonymous>
```

and exits nonzero.

## Expected behavior

Both forms should be equivalent:

```text
direct: /tmp/x/project.json
spread: /tmp/x/project.json
```

Node 24.18.0 produces the expected output.

## Why I expected this to be supported

The Perry 0.5.1220 README marks:

* functions/rest parameters as supported;
* the spread operator in calls and literals as supported;
* `path.join` as supported.

The runtime parity documentation also lists `path.join([...paths])` as supported without a caveat.

Perry's own parity suite contains a spread invocation of `path.join` in:

```text
test-parity/node-suite/path/join/type-errors-extra.ts
```

using:

```ts
path.join(...args)
```

That fixture appears to test invalid argument types, but does not include the corresponding valid-string-array case.

## Current `main`

I also checked current Perry `main`.

At the time of testing:

```text
commit: 0328a63bbc1b582bfdbb0842a957aac861bd84d7
version: 0.5.1413
```

I noticed that the current source contains an internal comment in the native-module call lowering code describing:

```text
ns.method(...arr)
```

as a separate runtime spread gap.

That appears to be the same category as:

```ts
path.join(...parts)
```

So this may already be a known implementation limitation internally.

I could not find a corresponding user-facing compatibility caveat or a specific issue tracking this case, and the public compatibility documentation currently appears to describe both spread calls and `path.join` as supported.

If this gap is already tracked elsewhere, please feel free to close this as a duplicate.

## Additional observation

The problem does not require a rest parameter. This also fails:

```ts
const parts = ['/tmp/x', 'project.json'];
path.join(...parts);
```

while this succeeds:

```ts
path.join('/tmp/x', 'project.json');
```

A wrapper using a rest parameter therefore fails for the same reason:

```ts
function joinPath(...parts: string[]): string {
  return path.join(...parts);
}
```

## Workaround

Avoiding a spread into the native module function works:

```ts
function joinPath(...parts: string[]): string {
  if (parts.length === 0) {
    return '.';
  }

  let result = parts[0]!;
  for (let index = 1; index < parts.length; index += 1) {
    result = path.join(result, parts[index]!);
  }

  return result;
}
```

## Environment

* Perry: `0.5.1220`
* Package: `@perryts/perry`
* Target: `linux`
* Host: Linux x86_64
* Node used to invoke the Perry toolchain: `24.18.0`

proggeramlug:
Fixes #7720.

## The bug

```ts
import path from 'node:path';
const parts = ['/tmp/x', 'project.json'];
path.join('/tmp/x', 'project.json');  // "/tmp/x/project.json"
path.join(...parts);                  // TypeError [ERR_INVALID_ARG_TYPE]
```

Every native-module fast path in `lower_call` consumes its arguments
**positionally**, and a spread argument is lowered as a single expression
holding the whole array. `path.join(...parts)` therefore reached the
`args.len() == 1` arm and became `PathNormalize(<array>)`, which rejected the
array as a non-string.

It is not a `path` bug. The same positional fold is why, on `main`:

| call | perry | node |
|---|---|---|
| `path.join(...['a','b'])` | `TypeError [ERR_INVALID_ARG_TYPE]` | `a/b` |
| `util.format(...['x=%s','X'])` | `[ 'x=%s', 'X' ]` | `x=X` |
| `fs.existsSync(...['/tmp'])` | `false` | `true` |

## The fix

When any argument is spread **and** the callee is a node-core module namespace
method (or a named export of one), decline the whole fast-path chain. The
fall-through tail then builds an `Expr::CallSpread` over the namespace member —
exactly the lowering the value-read form already takes:

```ts
const j = path.join; j(...parts);   // this already worked on main
```

Codegen materializes the argument array and dispatches through
`js_closure_call_apply_with_spread` → `js_native_call_method` →
`dispatch_native_module_method`, which is variadic by construction. It gets the
valid case right *and* reproduces Node's `ERR_INVALID_ARG_TYPE` (with the
matching `code`) for an invalid one.

This generalizes the per-module bail #6668 added for `crypto`; those two
crypto guards stay, because they also cover the bare `crypto` **global**
receiver, which is not an import and so is invisible to the new predicate.

### What is deliberately out of scope

* **ext/npm native modules** (mysql2, redis, node-forge). Their
  `NativeMethodCall` rows are wired in codegen with no by-name runtime
  dispatcher behind them, so declining their fast path would trade a wrong
  answer for no answer.
* **Native class statics** (`Buffer.concat(...list)`, `URL.parse(...)`). A
  different lowering family whose dynamic dispatch does not cover the same
  surface — `Buffer.concat` is already broken through the dynamic path for the
  plain `B.concat(list)` call, independent of spread, so routing spread onto it
  would swap a loud failure for a silent wrong answer.

Both exclusions are asserted by tests, not just described here.

## Tests

`crates/perry-hir/src/lower/expr_call/native_module_spread_tests.rs` (6 tests,
`cargo test -p perry-hir --lib`, so per-PR-visible) asserts **which lowering a
call got**, in both directions:

* a spread call is diverted to `CallSpread`;
* a *non-spread* `path.join('a','b')` still lowers to `PathJoin`.

The second half matters: the generic dispatch is a correct fallback, so a
regression that disabled the fast path everywhere would still print the right
answer. Both halves were sabotage-checked — with the guard forced to `false`,
3 of the 6 fail; with the fast path removed, the other half fails.

This is also why the existing `node-suite/path/join/type-errors-extra.ts`
already contained `path.join(...args)` and stayed green through the whole life
of the bug: it only spreads segments Node *rejects*, so both the broken and the
correct lowering throw `ERR_INVALID_ARG_TYPE`. CLAUDE.md's fourth way a gate
can be unable to fail — the gate ran, its subject never did. The new
`node-suite/path/join/spread.ts` supplies the valid-segment cases the issue
asked for.

Byte-compared against node `26.5.1`:

* `test-parity/node-suite/path/join/spread.ts` — namespace / namespace-import /
  require-alias / named-import / `posix` / `win32` / sub-namespace alias /
  value-read / in-loop forms, plus mixed, trailing, single, empty and
  normalizing segment lists
* `test-parity/node-suite/path/resolve/spread.ts` — the reset-on-absolute sibling
* `test-parity/node-suite/util/format/spread.ts`

## Verification

* `run_parity_tests.sh --suite node-suite --module path` — **94/94**, 100%
  (including the two new fixtures)
* `cargo test -p perry-hir --lib` — green
* the issue's repro, plus `util.format` / `fs.existsSync` / `os.homedir` /
  `fs.writeFileSync` / `fs/promises.readFile` / `crypto.createHash` /
  named-import and `node:console` / `node:process` namespace forms, all now
  byte-identical to node
* non-module spread paths unchanged: `Math.min(...xs)`, `arr.push(...xs)`,
  `console.log(...xs)`

One unrelated pre-existing gap surfaced while testing and is **not** touched
here: `Object.assign(target, ...sources)` passes the sources array as one
argument (`{"0":{...},"1":{...}}` instead of a merge). Different family
(`Object` statics, not a node module) — worth its own issue.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1420
+**Current Version:** 0.5.1421
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1420"
+version = "0.5.1421"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1420"
+version = "0.5.1421"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1420"
+version = "0.5.1421"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1420"
+version = "0.5.1421"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1420"
+version = "0.5.1421"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1420"
+version = "0.5.1421"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1420"
+version = "0.5.1421"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1420"
+version = "0.5.1421"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1420"
+version = "0.5.1421"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1420"
+version = "0.5.1421"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1420"
+version = "0.5.1421"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1420"
+version = "0.5.1421"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1420"
+version = "0.5.1421"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1420"
+version = "0.5.1421"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1420"
+version = "0.5.1421"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1420"
+version = "0.5.1421"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1420"
+version = "0.5.1421"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1420"
+version = "0.5.1421"
 dependencies = [
  "perry-ffi",
  "reqwest",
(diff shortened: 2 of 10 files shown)


proggeramlug:
## Merging as v0.5.1421

**The best thing in this PR is the observation that the existing coverage could not fail**, and I verified it before trusting it. `test-parity/node-suite/path/join/type-errors-extra.ts` is four lines:

```ts
for (const args of [["a", 1], ["a", null], ["a", {}], ["a", []]] as any[]) {
  try { console.log("join:", path.join(...args)); } catch (err: any) { ... }
}
```

Every case contains a non-string Node rejects, so the **broken** lowering (which rejects the whole array) and the **correct** one (which rejects the offending element) throw the identical `ERR_INVALID_ARG_TYPE`. The only pre-existing spread coverage for `path.join` was structurally incapable of telling working from broken — CLAUDE.md's hazard 4, in a fixture rather than a gate.

**The root cause is general, and the table proves it isn't a `path` bug:** every native-module fast path consumes arguments positionally, so a spread — lowered as one expression holding the array — folds into the `args.len() == 1` arm. `util.format(...['x=%s','X'])` returning `[ 'x=%s', 'X' ]` and `fs.existsSync(...['/tmp'])` returning `false` are the same defect wearing different clothes.

**Declining the fast path rather than teaching each one about spread is the right shape.** The fall-through already builds `Expr::CallSpread` over the namespace member, which is exactly what the value-read form (`const j = path.join; j(...parts)`) has always taken and which already worked. It dispatches through `js_native_call_method` → `dispatch_native_module_method`, variadic by construction — so this reuses a working path instead of adding a parallel one, and it reproduces Node's error (with matching `code`) on the invalid case rather than only fixing the happy path.

**The out-of-scope section is the part I'd want kept.** Both exclusions are argued from what the alternative would cost, not from effort:
- **ext/npm modules** — their `NativeMethodCall` rows have no by-name runtime dispatcher behind them, so declining would trade a wrong answer for *no* answer.
- **Native class statics** — `Buffer.concat(list)` is already broken through the dynamic path independent of spread, so routing spread onto it would swap a loud failure for a silent wrong one.

The two #6668 crypto guards are correctly retained (verified at `native_module.rs:340`): they also cover the bare `crypto` **global** receiver, which is not an import and so is invisible to the new predicate.

### Verification

`cargo test -p perry-hir --lib`: **287 passed, 0 failed.** `run_parity_tests.sh --suite node-suite --module path`: 94/94.

Full `cargo test -p perry-runtime --lib` shows **3 failures, byte-identical to the set already red on `main`** (`generator_attach_prototype::*`) — not from this PR, and being fixed separately. Lint 19/19.

The `Object.assign(target, ...sources)` finding is correctly left alone and deserves its own issue — different family, and folding it in would have made this diff unreviewable.

proggeramlug:
Marking draft: a same-host A/B (this tree vs. the same tree with the guard forced to `false`) turned up a regression I introduced — `fs.promises.readFile(...args)` returns `ENOENT` on `main` (i.e. the call really happens) but throws `TypeError: value is not a function` on this branch.

Cause: the predicate recurses through any sub-namespace receiver, but only some sub-namespaces have a runtime dispatch bucket (`nm_module_index` has `path.posix`, `path.win32`, `util.types`, `crypto.subtle`, `punycode.ucs2` — there is no `fs.promises` bucket). Diverting a bucket-less sub-namespace to the dynamic path replaces a working call with a hard throw.

Running a 32-case A/B matrix over the node-core spread surface (both arms + node oracle) to set that list from measurement rather than guesswork; will push the narrowed predicate and re-request review.

Issue #7683, opened 2026-08-09
gc: perry-runtime lib tests SIGSEGV in typed_feedback::object_shape via js_object_clone_with_extra (1 in ~102 full-suite runs, no test reports FAILED)


proggeramlug:
## Symptom

The `perry-runtime` lib test binary **SIGSEGVs**, taking the whole suite process down mid-run with **no test reporting `FAILED`**. Observed once in a 102-run soak of `cargo test -p perry-runtime --lib --no-fail-fast` (arm64 macOS, debug profile, default parallelism).

Faulting thread and stack, from the macOS crash report:

```
exception: EXC_BAD_ACCESS / SIGSEGV, KERN_INVALID_ADDRESS at 0x000001fffffffff8
thread:    gc::tests::copying::large_object_clone_direct_copy_keeps_young_child_alive_and_excludes_parent

  perry_runtime::typed_feedback::object_shape
  perry_runtime::typed_feedback::invalidate_representation_change
  perry_runtime::gc::layout::layout_mark_unknown
  perry_runtime::object::mark_object_dynamic_shape_unknown
  perry_runtime::object::set_object_keys_array
  js_object_clone_with_extra
  perry_runtime::gc::tests::copying::large_object_clone_direct_copy_keeps_young_child_alive_and_excludes_parent
```

## Why this is worth an issue rather than a retry

**A crash is not a test failure, and the harness reports it as neither.** `cargo test` exits non-zero with `process didn't exit successfully: ... (signal: 11, SIGSEGV)` and **zero `FAILED` lines**, so a log tail shows a clean-looking partial run. Any wrapper that greps for `FAILED`, or that classifies "a signal" as an external interruption, records this as benign. (Both mistakes were made while finding it: a `signal:`-based guard first lumped SIGSEGV in with an unrelated SIGTERM.)

Of 69 `perry_runtime` crash reports on this host, **68 are the deliberate `pin_latch_sabotage_child` SIGABRT** — an instrument doing its job — and this is the only genuine one, so it is easy to lose in the noise of an expected-abort test.

## The address is the tell

`0x000001fffffffff8` is not a small offset off a live object; it is roughly `2^41`, the shape of an index-times-8 computed from a garbage id. `object_shape` guards with `is_valid_obj_ptr` and then reads the `GcHeader` at `ptr - GC_HEADER_SIZE`, so the input address either passed that predicate while being junk, or the fault is inside the predicate's own probe.

[...]

PR #7727, opened 2026-08-09
fix(gc): initialize keys_array before clone_with_extra's collecting allocation (#7683)

proggeramlug:
### Fixed

- **`js_object_clone_with_extra` published a half-built object whose `keys_array` slot held recycled heap bytes (#7683).** Both branches initialise `object_type`, `class_id`, `parent_class_id`, `field_count` and `meta` immediately after allocation, then set `keys_array` only at the end via `set_object_keys_array`. Between those two points sits `crate::array::js_array_alloc`.

  That call **allocates, so it can collect** — and the collector reads exactly this slot as a child edge (`object::gc_keys_array_slot`, enumerated by `gc_child_slots`). A collection landing in that window scans a pointer the mutator never wrote.

  The bytes are not zero. `arena_alloc_gc_old`'s fast path deliberately reuses a swept, **non-zeroed** hole (#7437: *"reuse a swept same-size hole … otherwise a block with any live object never yields its dead bytes back"*), so the slot holds real leftover heap content from whatever last occupied it. Whether that content happens to look like a plausible-but-unmapped address depends on allocation history — which is the shape of the ~1-in-102 `typed_feedback::object_shape` SIGSEGV reported in #7683.

  Every sibling allocator in `object/alloc.rs` already nulls the slot at this point. This function was the one that did not.

  **On the test, and why it checks source rather than behaviour.** The runtime version was written first: force a collection into the window (`force_next_general_arena_alloc_slow` + `GC_OLD_RECLAIM_PENDING`, the levers #7251 established), then assert the published clone's `keys_array` is sane. **It passed with the fix deleted.** Two independent reasons: by the time the function returns, `set_object_keys_array` has written the slot correctly, so nothing observable survives the window; and a fresh arena block is zeroed, so even inside the window the garbage reads as null unless the allocation lands in a recycled old-space hole with the right history.

  Reproducing it in-suite therefore needs a specific swept-hole layout **and** a collection landing in a few-instruction window. That is exactly why the fix is a by-construction initialisation rather than a guard, and why the guard asserts the invariant where it is decidable — the source, in the style of `scripts/gc_pin_sites.py`'s custody check for `GC_FLAG_PINNED`. Removing either initialisation fails the test.

  One note for anyone writing a similar source-level check: the first version matched the phrase `arena_alloc_gc` inside **its own explanatory comment**, registering a third allocation site and failing against correct code. It now strips comments before scanning. A source check that reads its own documentation as code is worse than no check.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved object cloning reliability by preventing stale memory references from being observed during garbage collection.
  * Reduced the risk of crashes or corrupted data when cloning objects that include additional properties.

* **Tests**
  * Added regression coverage to verify safe initialization across relevant cloning and allocation scenarios.

* **Documentation**
  * Added release notes describing the cloning fix and its validation.

* **Chores**
  * Updated the application version to 0.5.1423.


proggeramlug:
## Merging as v0.5.1423

`js_object_clone_with_extra` initialises `object_type`, `class_id`, `parent_class_id`, `field_count` and `meta` right after allocation, then sets `keys_array` only at the end via `set_object_keys_array`. Between those two points sits `crate::array::js_array_alloc` — which **allocates, so it can collect**, and the collector reads exactly that slot as a child edge (`object::gc_keys_array_slot`).

The bytes are not zero: `arena_alloc_gc_old`'s fast path deliberately reuses a swept, **non-zeroed** hole (#7437), so the slot holds real leftover heap content. Whether it looks like a plausible-but-unmapped address depends on allocation history — the shape of the ~1-in-102 `typed_feedback::object_shape` SIGSEGV in #7683. Every sibling allocator in the file already nulls it here; this function was the one that didn't.

### My first test was vacuous, and I'm recording that rather than quietly replacing it

I wrote the runtime version first: force a collection into the window (`force_next_general_arena_alloc_slow` + `GC_OLD_RECLAIM_PENDING`, #7251's levers), then assert the published clone's `keys_array` is sane. **It passed with the fix deleted.** Two independent reasons — by the time the function returns `set_object_keys_array` has written the slot correctly, so nothing observable survives the window; and a fresh arena block is zeroed, so even inside the window the garbage reads as null unless the allocation lands in a recycled old-space hole with the right history.

Reproducing it in-suite therefore needs a specific swept-hole layout **and** a collection landing in a few-instruction window. That is exactly why the fix is a by-construction initialisation rather than a guard, and why the guard asserts the invariant where it *is* decidable — the source, in the style of `scripts/gc_pin_sites.py`'s `GC_FLAG_PINNED` custody check. Removing either initialisation fails it:

```
#7683: an allocation site in js_object_clone_with_extra has no
`keys_array = ptr::null_mut()` before its `js_array_alloc`.
test result: FAILED
```

And one more self-inflicted lesson worth leaving in the file: the first version of that source check matched the phrase `arena_alloc_gc` **inside its own explanatory comment**, registering a third allocation site and failing against correct code. It strips comments before scanning now. A source check that reads its own documentation as code is worse than no check.

### Credit

The root cause came from an agent's static analysis after a 150-run soak found nothing — and it was right to call that soak **inconclusive rather than clean**: at a 1-in-102 rate, seeing zero crashes in 150 runs has a ~23% probability by luck, so N≈300–450 would be needed to conclude anything. Fixing the mechanism it found beats soaking harder.

Gates 21/21 (now including `perry-runtime --lib` and `perry-codegen --lib`, which my gate previously did not run).

PR #7729, opened 2026-08-09
fix(gc): pace PERRY_GC_ZEAL by allocation so the instrument terminates (#7728)

### LINKED ISSUE #7728: gc: PERRY_GC_ZEAL forces a collection at EVERY back-edge poll, so zeal no longer terminates on real workloads
`PERRY_GC_ZEAL=1` is the primary instrument for catching moving-GC correctness bugs (paired with `PERRY_GC_PROTECT_FROMSPACE=1` it produced the precise fault that diagnosed #7682). It no longer completes on a real workload.

## Repro

`gc-handoff/apps/iso_miss.ts` — a tree-walking interpreter, 40 rounds of `fib 21`.

| build | config | wall |
|---|---|--:|
| `ca8c0d617` (0.5.1418) | no zeal | 19 s, `checksum 437840 misses 0` |
| `ca8c0d617` | `PERRY_GC_ZEAL=1` | **timeout at 240 s, no output** |
| `f3e14a61e` (0.5.1389) | `PERRY_GC_ZEAL=1` | instant, correct |

## It is not a livelock, and `f3e14a61e` was not "working"

Measured on the pinned quiet host, scaling the round count:

| rounds | loop_polls | forced collections | wall |
|---|--:|--:|--:|
| 1 | 70,963 | 70,968 | 36.3 s |
| 2 | 141,926 | 141,931 | 72.7 s |

Perfectly linear, so 40 rounds is ~24 minutes rather than a hang. Zeal forces a collection at **every** back-edge poll: ~511 µs of fixed per-collection cost (root scan over the shadow stack plus ~55 side-table scanners) to relocate a mean of **5.9 objects**. Practically all the work is the collection's fixed overhead, not the relocation zeal exists to stress.

The `f3e14a61e` endpoint was fast because it was **vacuous**: `PERRY_GC_MOVING_LOOP_POLLS` was default-OFF there (#7161), so a compute-only program reached no loop safepoint and zeal forced nothing. That build predates the #7604 exit verdict, so it exited 0 in silence. #7721 flipped the poll default ON — correctly, it is a large collector win — and in the same commit turned zeal from free-and-vacuous into correct-but-unusable.

Confirmed by isolation: the **old** `f3e14a61e` compiler with `PERRY_GC_MOVING_LOOP_POLLS=1` forced at compile+run time already takes **35.8 s for one round** under zeal (0.62 s without). So no commit "broke" zeal; zeal was never paced, and the poll default is what exposed it.

#7254 already logged "a striking concentration of multi-minute-plus runs" under this pairing and left the population untriaged. This is that triage.

## Why it matters

An instrument that does not terminate is an instrument nobody switches on, and this bug class (a stale pointer across a moving collection, #7154 family) has shipped more than once. `scripts/gc_instrument_smoke.sh` cannot see the problem because its fixture is deliberately sized at ~1200 polls "so the zeal arm costs seconds rather than minutes" — at that size, every-poll and paced are indistinguishable.

## Fix direction

Pace zeal by allocation, the way V8 (`--gc-interval`) and SpiderMonkey (`gcZeal(mode, frequency)`) do, with an escape hatch restoring literal every-poll semantics, plus a budgeted CI arm at a realistic poll count so this cannot regress silently again.

proggeramlug:
Closes #7728.

`PERRY_GC_ZEAL=1` — the primary instrument for moving-GC correctness bugs, and half of the pairing that produced the precise fault behind #7682 — stopped completing on real workloads. On `gc-handoff/apps/iso_miss.ts` (19 s without zeal) it now times out at 240 s with no output.

## It was never a livelock, and the "good" endpoint was never good

Scaling the round count on the pinned quiet host:

| rounds | loop_polls | forced collections | wall |
|---|--:|--:|--:|
| 1 | 70,963 | 70,968 | 36.3 s |
| 2 | 141,926 | 141,931 | 72.7 s |

Perfectly linear — 40 rounds is ~24 minutes, not a hang. Zeal forced a collection at **every** back-edge poll: ~511 µs of fixed per-collection cost (root scan over the shadow stack plus ~55 side-table scanners) to relocate a mean of **5.9 objects**. Nearly all of it is the collection's fixed overhead, not the relocation zeal exists to stress.

The `f3e14a61e` endpoint was fast because it was **vacuous**. `PERRY_GC_MOVING_LOOP_POLLS` was default-OFF there (#7161), so a compute-only program reached no loop safepoint and zeal forced nothing — and that build predates the #7604 exit verdict, so it exited 0 in silence. #7721 flipped the poll default ON (correctly — it is a large collector win) and in the same commit turned zeal from free-and-vacuous into correct-but-unusable.

Isolated rather than assumed: the **old** `f3e14a61e` compiler with `PERRY_GC_MOVING_LOOP_POLLS=1` at compile and run time already costs **35.8 s for one round** under zeal, against 0.62 s without. No commit broke zeal. Zeal was never paced, and the poll default is what exposed it. #7254 had already logged "a striking concentration of multi-minute-plus runs" under this pairing and left it untriaged; this is that triage.

## The fix

Zeal is now **allocation-paced**, the model V8 (`--gc-interval`) and SpiderMonkey (`gcZeal(mode, frequency)`) both use: it forces a collection at the first poll at which `PERRY_GC_ZEAL_ALLOC_KB` (default 4) of new nursery material has accumulated since the last one.

The stride is a **monotone high-water mark**, not a "bytes since" delta — each forced collection rearms to `from_space_after + stride`. A collection that reclaims nothing (an escalation to a non-moving full mark-sweep, which #7592 and #7682 both produced in the field) therefore still demands another full stride of real allocation. Total forced collections are bounded by `bytes_allocated / stride` whatever the collector does with them, which makes this a bound rather than a hope.

`PERRY_GC_ZEAL_ALLOC_KB=0` restores the literal every-poll semantics. That is the right setting for a small fixture or a bug window executed exactly once, and it is what `gc_instrument_smoke.sh` now pins for its own tiny fixture so this change cannot quietly weaken the existing gate.

**What pacing gives up, stated rather than buried:** a window crossed a single time may now fall between two forced collections. A window that *recurs* — every shape in the #7154 family, which is why the reproducers are loops — is still caught, after N KB of allocation instead of on the first iteration.

## Why this could regress silently, and what now stops it

`scripts/gc_instrument_smoke.sh` runs zeal end-to-end and is green. It could not see this: its fixture is deliberately sized at ~1200 polls "so the zeal arm costs seconds rather than minutes", and at that size every-poll and paced are indistinguishable. The gate was well built and pointed at a workload too small to expose the axis that broke.

So arm 6 is new and **budgeted**: a 400k-iteration workload at the *shipped default* (with `env -u` so the file's every-poll pin does not apply), requiring correct output inside a wall-clock budget that sits between the paced cost and the unpaced ~200 s. It also asserts non-vacuity — forced collections, copying minors and moved objects all non-zero, and `forced < loop_polls` — because "fast because it collects nothing" would be a worse regression than the slow instrument it replaces.

## Tests

In the required `cargo-test` gate (`gc/tests/fromspace_protect.rs`):

- `zeal_pacing_bounds_forced_collections_but_still_moves_objects` — the regression test. Drives a hot poll loop and asserts forced collections are bounded well below the poll count, **paired with two liveness assertions** (collections still forced, survivors still moved). Sabotage-checked: pinning the stride to 0 fails it with `2000 forced for 2000 polls`, exactly the pre-fix ratio.
- `zeal_alloc_stride_zero_restores_every_poll_collection` — the OFF state of the new knob, per the binding kill-policy.
- `zeal_pacing_rearms_above_survivors_so_a_useless_collection_cannot_loop` — pins the monotone high-water mark against a delta-based rewrite.
- `zeal_alloc_stride_knob_parses_both_states` — including that `0` is meaningful rather than garbage.

## Docs

CLAUDE.md's instrument table gains the pacing row and loses a stale claim: it still said loop polls were "default off since #7161", which #7721 had made false. That sentence is why zeal's cost was invisible in the first place.

## Summary by CodeRabbit

- **New Features**
  - GC zeal now supports allocation-paced forced collections using a configurable nursery allocation stride.
  - Added `PERRY_GC_ZEAL_ALLOC_KB`; the default is 4 KiB, while `0` preserves collection at every poll.
  - GC zeal results now report pacing and allocation-stride metrics.

- **Documentation**
  - Updated GC rooting, memory-model, and developer guidance with pacing behavior, configuration, and performance considerations.

- **Tests**
  - Added coverage for pacing limits, rearming, configuration parsing, object movement, and realistic default settings.

## Measured (pinned quiet host, `ssh perry@perry-macos.local`)

The full workload under zeal: **240 s timeout → 98.8 s with the correct answer** (`checksum 437840 misses 0`), forcing 193,087 collections out of 2,838,560 polls, all of them copying minors, relocating 3,115,719 objects. Unpaced, the same run costs ~1,426 s.

The stride sweep is **one binary and one env var** at a quarter scale — the cleanest available A/B — and `loop_polls` is **283,852 in every row**, so the knob provably changes only the decision to collect, not the number of safepoints:

| `ALLOC_KB` | forced collections | moved objects | wall |
|---|--:|--:|--:|
| 0 (pre-fix behaviour) | 283,857 | 1,629,647 | 142.6 s |
| 1 | 70,929 | 815,460 | 36.1 s |
| **4 (default)** | **19,314** | **325,830** | **10.2 s** |
| 16 | 5,070 | 129,959 | 3.0 s |
| 64 | 1,291 | 52,357 | 1.1 s |

Row 0 reproduces the pre-fix 1:1 ratio exactly on the shipped binary (283,857 collections for 283,852 polls). Every row keeps `copying_minors == forced_collections` and `moved > 0`, so no stride quietly degrades the instrument into non-moving sweeps.

4 KB rather than the faster 16/64 is deliberate: this is a correctness instrument, so the default errs toward sensitivity — one collection per ~15 loop iterations, still 14x cheaper than unpaced.

**The zeal-OFF path is untouched**, which matters because back-edge polls are now on every allocating loop by default: the same workload without zeal is 4.49 s on `main` and 4.49 s here. The pacing work all sits inside the `!GC_SAFEPOINT_PENDING` branch, so a default build reaches the same single cached-bool read and return it did before.

## Note on a pre-existing failure

`gc::tests::runtime_roots::generator_attach_prototype` has 3 tests failing in the **debug** profile, and they fail identically on pristine `c156f8a41` (1962 filtered there vs 1966 here — exactly the 4 tests this PR adds). They are alloc-point/trigger-arming tests, untouched by this change. Not introduced here; flagged rather than left unmentioned.

## Follow-up: arm 6 was weak until it was actually run

Two defects in my own gate, found by running it rather than reasoning about it:

1. **The fixture did not allocate.** The obvious version allocates a record per iteration and drops it, which scalar-replaces into nothing: **6** forced collections and **17** moved objects over 400,000 polls. The arm ran the loop and never gave the collector anything to relocate. Pushing into a bounded rolling array (plus a string concat) takes `moved_objects` from 17 to **640,364**.

2. **A wall-clock budget could not be the discriminator.** Measured on the quiet host, the fixture is **0.49 s paced vs 11.85 s unpaced** — a real 24x, but both fit inside any budget loose enough not to flake on a shared CI runner. So the arm now keys on the **collections-to-polls ratio**, which is host-independent and exact: the regression's signature is 1:1 (measured 200,069 forced for 200,064 polls), the shipped default is 1-in-40, and the gate fails above 1-in-4. The budget stays as the weaker "terminates at all" guard.

Sabotage-checked, like the unit test: forcing arm 6 back to every-poll fails it with

```
FAIL [arm6]: zeal forced 200069 collections for 200064 polls
      (threshold: fewer than one per 4 polls). That is the unpaced
      behaviour #7728 removed -- one whole evacuating minor per loop
      iteration, which took a 5 s program to ~24 minutes.
```

The full `gc_instrument_smoke.sh` passes end to end on the final build (all six arms, exit 0), including the pre-existing arms: 13/13 quarantine probes clean, and #7254's reproducer still pinned.

## Confirmed on the final HEAD build

The measurements above were taken on the first build of this branch; the later control-flow refactor was re-measured on a fresh `--release` build of the final HEAD and produced **byte-identical counters** (`forced_collections=193087 copying_minors=193087 moved_objects=3115719 loop_polls=2838560 paced_polls=2645478`), 101.5 s wall, correct answer — with the no-zeal arm at 5.17 s.


proggeramlug:
## Merging as v0.5.1428 — the pacing question was tested against real history, not argued

I commissioned an adversarial audit on the one thing that could make this a bad trade: **does pacing weaken the instrument in a way that would have hidden a bug this project actually found?** The answer is no, and the reasoning is structural rather than lucky.

Every #7154-family reproducer was checked against its own fixture:

| bug | window | fixture scale |
|---|---|---|
| #7184, #7192 | shadow-frame slot overflow / late root store | 600-iteration inner loop, 400 outer |
| #7206 | stale receiver across a call | 4,000-iteration inner, 200 outer |
| #7528 | receiver reused across ~1,160 lines | `const N = 50_000` |
| #7682 | unrooted runtime **cache** | 40 rounds of a full interpreter |

Two reasons this holds:

- **#7682 is the cache class, not the register class.** Per this repo's own rule, a perfectly-reproducible GC bug means a table stale at collection #0 — it needs *a* forced collection somewhere in the run, not the *right* poll.
- **Every register-class reproducer is a 400–50,000 iteration loop**, and that isn't accidental: these bugs were hard enough to hit that the fixtures were already over-provisioned before zeal existed. Pacing rides on practice this project already had.

And the density check makes it concrete: your own measurement is 19,314 forced collections over 283,852 polls at the 4 KB default — **~273 bytes per iteration**, so a 600-iteration constructor loop gets **~40 forced collections**, not one.

### The other four things I wanted verified, all confirmed in code

- **Monotone high-water mark, not a delta.** `note_zeal_poll_collection` rearms to `from_space_bytes_after + stride` from `copying_from_space_in_use_bytes()`. A collection reclaiming nothing still demands a full stride. `zeal_pacing_rearms_above_survivors_so_a_useless_collection_cannot_loop` pins it against the delta-shaped regression *by name*.
- **The escape hatch is real and pinned.** `gc_instrument_smoke.sh:41` exports `PERRY_GC_ZEAL_ALLOC_KB=0` for arms 1–3 and 5, and arm 6 explicitly opts back out with `env -u` so it tests the shipped default. Both directions in code.
- **The regression test has the bound AND both liveness assertions** — `forced < POLLS/4`, `forced + paced == POLLS`, *and* `forced > 0` / `moved > 0`. A bound-only test would pass on an instrument that silently stopped collecting.
- **Arm 6 keys on the ratio, not the clock** (`scale_forced * 4 >= scale_polls`), with the wall-clock kept only as a terminates-at-all backstop at ~7.6× margin. Your own commit `c0b95a6a4` made that change — which is the right instinct, because a timing gate that flakes gets disabled.

**Scope is right too:** the diff only paces `js_gc_loop_safepoint`. `gc_safepoint_moving_minor`'s zeal branch — the microtask-pump boundary — is untouched and still forces every time. Async code keeps full strength; only CPU-bound loop iteration is paced.

1. **`gc-stress` is not a required context** (confirmed against branch protection: `lint, cargo-test, parity, compile-smoke, api-docs-drift, security-audit, conformance-smoke-complete`). So a red arm 6 is visible but cannot block a merge, which undercuts "this cannot quietly weaken the existing gate". Pre-existing, and the workflow says so itself — but this PR gives it a first green run, which is the prerequisite for promoting it.
2. `note_zeal_poll_collection` runs unconditionally after `gc_safepoint_moving_minor()`, including on paths where its entry guards block the collection. Such a poll counts as neither forced nor paced, so `forced + paced == loop_polls` could fail outside the test's guard-free setup. Pre-dates this PR; worth a comment or a targeted test.

### Heads-up on the merge order

**#7735 rewrites the same lines of `js_gc_loop_safepoint()`.** They're semantically compatible — #7735 keeps the arming word nonzero for the whole process under zeal, which is exactly what your pacing needs — but it will hit a real conflict now that this landed. I'm reconciling it by hand and re-running `gc_instrument_smoke.sh` plus the `fromspace_protect.rs` zeal tests on that side.

Gates 21/21.

Issue #7506, opened 2026-08-06
typed_f64_receiver_method_clone_raw_loads_after_composed_guards: the guard-failure edge no longer calls $generic — miscompile or intentional collapse?


proggeramlug:
Split out of #7493, whose sweep listed this test among the failures `PERRY_RS4GC=0` heals. **It does not.** Run alone it fails under both lowerings:

```
$ cargo test -p perry-codegen --test native_proof_regressions \
    typed_f64_receiver_method_clone_raw_loads_after_composed_guards -- --exact
thread '…' panicked at crates/perry-codegen/tests/native_proof_regressions.rs:11750:
receiver field or numeric arg guard failure should call the generic method body:
define double @perry_fn_typed_f64_receiver_method_ts__probe(double %arg1, double %arg2) alwaysinline {

$ PERRY_RS4GC=0 cargo test … -- --exact      # identical failure
```

The earlier "healed by `PERRY_RS4GC=0`" reading came from a whole-suite `--test-threads=1` run, so it was an ordering artifact, not a lowering effect. #7493 therefore leaves this test deliberately **unpinned**, with a comment saying so, and hands it here.

## The failing assertion

```rust
assert!(
    caller_ir.contains(&format!("call double @{generic_body}(")),
    "receiver field or numeric arg guard failure should call the generic method body:\n{caller_ir}"
);
```

The earlier assertions in the same test still pass — the caller **does** emit the method-direct guard, the raw-f64 field guard and the `$typed_f64_recv` call, in that order. What is missing is the `$generic` call on the guard-failure edge. The test's next assertion (a `js_native_call_method_by_id` fallback for method-direct guard failure) is not reached.

## What needs deciding

Same question as #7494's four, and it has to be answered before touching the test:

* **If the guard-failure edge legitimately changed** — e.g. the receiver-field guard now shares the method-direct guard's fallback, so one dynamic call covers both — then the assertion is stale and should be re-pointed at the *property* (every guard-failure edge reaches a correct dynamic path), not at the presence of one symbol. #7492's class-field rewrite is the worked example: it replaced a positional window with an end-to-end data-flow assertion.
* **If it did not**, a guard-failure edge that no longer reaches the generic body is a miscompile: a receiver whose fields are not raw f64 would take the typed clone's raw loads anyway.

The test's own doc comment ("receiver clone must run only after method-direct and raw-f64 field guards") says the guards compose; check whether the composition still has three outcomes or has been collapsed to two.

## Repro

```
[...]

Issue #7404, opened 2026-08-04
perf: the % fast path gates on the i32 set, but only needs integer-valued-in-i64 (20.4× on bench_bitwise)


proggeramlug:
`bench_bitwise` is **20.4× slower than Node** (56956 vs 2787, checksums identical after #7403 made it measure anything at all). The cause is one instruction choice.

## Evidence chain

The hot loop contains **no runtime helper calls at all** — it is pure LLVM arithmetic, so the usual "native-able primitive became a runtime call" story does not apply here. The op mix per iteration:

```
4 frem     ← the problem
3 fsub  3 fadd  2 fmul  2 sitofp  2 fcmp  1 icmp  1 add
```

`frem` is not an aarch64 instruction. It lowers to a libm call:

```
$ otool -tvV … | grep 'bl.*fmod'
4 × bl  _fmod        # in runArithmeticBenchmark
```

And the profile says it dominates:

```
1754 samples  fmod
```

Four `fmod` calls per iteration × 10M iterations. The source is ordinary integer arithmetic:

```ts
result = result + (a % 1000);
result = result - (b % 1000);
result = result + ((a * 3) % 10000);
result = result - ((b * 2) % 10000);
```

V8 keeps these as SMIs and uses integer modulo.

## The mechanism already exists — and knows this is the problem

`stmt/loops.rs`'s `emit_object_array_write_index_i64` emits `srem`, and its doc comment names the exact issue:

> *"the counters' native i32 registers drive i64 add/sub/mul/srem directly: no float round-trip and, critically, no `frem` (which lowers to an fmod **LIBRARY CALL** on AArch64 — ~10ns per element). `srem` equals JS `%` on the proven domain (nonnegative dividend, positive divisor)."*

So the correctness argument is already worked out and the emitter is already written.

## Why it does not fire here

The integer path is reachable only through `ObjectArrayWriteNumber`, whose variants are `OuterCounter | InnerCounter | Constant | Add | Sub | Mul | Mod` — the #6812 object-array-write index pattern. `a` and `b` in `bench_bitwise` are **loop-carried locals** (`a = a + 1`, reset by a conditional), not loop counters, so they never enter that domain and fall to the f64 emitter's `frem`.

Every other `BinaryOp::Mod` site emits `frem` unconditionally: `expr/mod.rs:2240`, `expr/binary.rs:650`, `codegen/typed_abi/lower.rs:56` and `:433`, `lower_call/scalar_method.rs:174`.

## Suggested direction

[...]

PR #7730, opened 2026-08-09
fix(codegen): resolve #7494's four lowering-independent proof-test failures

### LINKED ISSUE #7494: native_proof / typed_shape_descriptors: four lowering-independent failures (numeric-coerce on a proven read, missing typed-array fallback record, negative-divisor frem, missing layout-note marker)
## Symptom

Four `perry-codegen` integration-test failures on `main` (f05ae3b83), arm64 macOS, debug, that are **independent of the root-lowering default** — they fail identically with and without `PERRY_RS4GC=0`, so they are not part of the #7370 native-roots family tracked in #7493:

`tests/native_proof_buffer_views.rs`

- `proven_buffer_and_typed_array_reads_are_numeric_operands` (`:915`)
  `proven native numeric reads should not force JS number coercion` — the emitted IR now contains `call double @js_number_coerce` for a read the test proves is native-numeric.
- `reassigned_typed_array_store_records_runtime_fallback` (`:1942`)
  `expected reassigned typed-array store to record runtime fallback` — no artifact record with `expr_kind == "TypedArraySet"`, `consumer == "TypedArraySet.slow_path"`, `access_mode == "dynamic_fallback"` and a non-null `fallback_reason`.

`tests/native_proof_regressions/integer_modulo.rs`

- `integer_modulo::i32_counter_mod_unsafe_or_nonliteral_divisors_keep_frem` (`:75`)
  `negative divisor must remain on floating remainder`.

`tests/typed_shape_descriptors.rs`

- `integer_arithmetic_array_push_omits_inbounds_layout_note_and_barrier` (`:141`)
  `IR should contain block marker …`.

All four **reproduce when run alone**, so this is not order dependence — unrelated to #7490 (poisoned mutex in `typed_feedback.rs`, fixed in #7492).

## Why they went unnoticed

Integration suites under `crates/*/tests/*.rs` do not run per-PR (nightly/tag only — the CLAUDE.md pitfall), and `cargo test` without `--no-fail-fast` stops at the first failing target, so a log tail can look like a clean run.

## What needs deciding

For each of the four, the question is the same and has to be answered before touching the test: **did the lowering legitimately change, or did a proof regress?**

- If the numeric-proof / fallback-record / layout-note behaviour changed on purpose, the assertion is stale and should be re-pointed at the current contract — and made to assert the *property* rather than a textual window, so the next intentional change does not silently re-break it. (#7492's class-field assertion is a worked example: the positional "coerce appears between these two labels" window was replaced by an end-to-end data-flow assertion.)
- If not, these are live regressions hiding behind an untested suite. An unnecessary `js_number_coerce` on a proven native read is a lost optimisation, but a *missing* `dynamic_fallback` record on a reassigned typed-array store is a soundness-adjacent gap in exactly the direction the test was written to catch.

## Repro

```
cargo test -p perry-codegen --tests --no-fail-fast -- --test-threads=1
```

proggeramlug:
## Summary

Resolves #7494 — the four lowering-independent `perry-codegen` integration-test failures. For each, the question was the same and had to be answered before touching the test: did the lowering legitimately change, or did a proof regress?

- **`reassigned_typed_array_store_records_runtime_fallback` — live regression, fixed.** `array[i] = v` on a local reassigned to a fresh typed array (`array = new Int32Array(8)`) skipped the typed-array dispatch — `is_width_tracked_typed_array_receiver` is gated on `receiver_class_name`, which correctly answers `None` for any reassigned local (#6906) — and fell through to `is_array_expr`'s plain-`number[]` fast store (`lower_index_set_fast`). That path writes the element at `arr_handle + 8`; a real typed-array object's data region starts at `arr_handle + 16`. The store landed on the typed array's own header bytes instead of element 0: a type-confusion write, not a missed optimization — the severity the issue flagged as "guilty until proven innocent." Fixed in two places: `is_array_expr` now excludes reassigned locals up front (mirroring `receiver_class_name`'s own guard), and `is_width_tracked_typed_array_receiver` (in both `index_get.rs` and `index_set.rs`) now resolves the receiver via `static_type_of` instead of `receiver_class_name` — every tier it gates is either `ctx.buffer_view_slots`-tracked (which reassignment already invalidates via the existing `update_buffer_view_for_assignment`) or a genuinely dynamic runtime call that re-validates the object's actual GC kind at runtime, so the stricter guard was only ever stopping this branch's own documented "aliases, reassigned locals, and unknown bounds stay on the runtime helper" fallback from being reached.

- **`proven_buffer_and_typed_array_reads_are_numeric_operands` — live gap, fixed (perf, not correctness).** A `sum += typedArray[i]` read tracked via `ctx.buffer_view_slots` (the stronger-bounds native path that `ta_param_f64_read.rs` explicitly declines to shadow) had no number-context entry point of its own, so it fell through to the generic `lower_expr` tier with `fallback_coerced = false` and picked up a redundant `js_number_coerce` — the residual-coerce rule can't see that this specific lowering path never produces a boxed value. Added `buffer_access::try_lower_typed_array_f64_read_for_number_context`, wired into `binary::lower_arithmetic_operand`.

- **`integer_arithmetic_array_push_omits_inbounds_layout_note_and_barrier` — stale assertion, re-pointed at the property.** #6915 (Repsel 4a.1, merged 2026-07-28 — five versions before #7702, which is often but wrongly blamed for this one) added an inline-store tier for array pushes whose value is canonical-raw-f64 by construction with typed-feedback recording off (the default), skipping the guarded tier's `js_typed_feedback_numeric_array_push_guard` / `js_array_numeric_push_f64_unboxed` pair entirely — so the test's `block_between("apush.numeric_fast.", …)` window no longer existed. The property the test protects — no slot-layout note, no write barrier, for a value proven non-pointer by construction — holds under either tier (`array_store_needs_layout_note` / `array_store_needs_write_barrier` are consulted before the tier is picked), so the test now asserts that property directly over the whole function, plus two sanity checks that the push and its arithmetic weren't optimized away.

- **`integer_modulo::i32_counter_mod_unsafe_or_nonliteral_divisors_keep_frem` — stale assertion, re-pointed at the property.** #7404/#7416 added a second integer-modulo fast-path hand-off for a dividend proven integer-valued outside `i32_counter_slots`, gated on the divisor being any non-zero `Expr::Integer` literal — not, like the older hand-off, one that is also positive and fits `i32`. The widening is sound: the dividend side is independently capped at 62 bits (a full bit of headroom below `i64::MIN`), and ECMA-262's `%` sign rule depends only on the dividend's sign, which is exactly what the existing IEEE-754 negative-zero repair already tests. `negative` and `out_of_i32` move to a new test asserting they now reach the integer path with the repair intact; the still-genuinely-ineligible cases (`Expr::Number` literals, zero-divisor UB avoidance) stay in the original test.

`typed_f64_receiver_method_clone_raw_loads_after_composed_guards` (#7506) and `large_local_array_push_inbounds_store_emits_precise_slot_barrier` (#7708) remain red on this branch — both are separate, open investigations, left untouched per scope.

## Test plan

- [x] `cargo test -p perry-codegen --no-fail-fast --tests` (full integration + lib suite) — same 2 pre-existing, out-of-scope failures as a clean `main` baseline (`typed_f64_receiver_method_clone_raw_loads_after_composed_guards`, `large_local_array_push_inbounds_store_emits_precise_slot_barrier`); all four target tests green; zero new failures.
- [x] `cargo test -p perry-codegen --no-fail-fast --test native_proof_buffer_views --test native_proof_regressions --test typed_shape_descriptors` — reconfirmed clean after `cargo fmt`.
- [x] `cargo fmt --all -- --check` clean.


proggeramlug:
## Merging as v0.5.1424

Four tests, and the disposition question was answered per test rather than in bulk — which is what makes this landable.

### #1 is a real memory-corruption bug, not a test problem

`reassigned_typed_array_store_records_runtime_fallback`. After `array = new Int32Array(8)`, an `array[i] = v` store silently type-confused: `is_width_tracked_typed_array_receiver` was gated on `receiver_class_name`, which correctly returns `None` for any reassigned local (#6906), so the store declined that arm and fell through to `is_array_expr`'s **plain-`number[]`** fast path — writing the element at `arr_handle + 8`, inside a real typed-array object whose data region starts at `arr_handle + 16`.

That is an unboxed-pointer-plus-wrong-offset write. The dynamic-fallback arm's own comment already promised what should have happened — *"aliases, reassigned locals, and unknown bounds stay on the runtime helper"* — and `receiver_class_name` broke that promise by answering `None` before the arm was reached.

The two-part fix is the right shape: exclude reassigned locals from `is_array_expr`, **and** move the typed-array receiver predicate off `receiver_class_name` onto `static_type_of`. The widening is the part worth watching, and the argument for it is stated rather than assumed — every tier it gates is independently safe against a stale view, either `buffer_view_slots`-checked or a genuinely dynamic runtime call. Worth re-checking if a new tier is ever added under that predicate.

### The other three

- **#2 `proven_buffer_and_typed_array_reads_are_numeric_operands`** — live gap, perf-only: a buffer-view-tracked read in arithmetic context had no number-context entry point, so it picked up a redundant `js_number_coerce` on a provably non-boxed native read.
- **#3 `integer_arithmetic_array_push_...`** — **stale assertion**, and the cause is #6915 (Repsel 4a.1), **five versions before** the #7702 it kept being blamed on. It legitimately added an inline-store tier that bypasses the guarded blocks the test's `block_between` window searched for. Re-pointed at the actual property — no layout note, no barrier, for a provably non-pointer value — asserted globally rather than through a block window that no longer exists, plus two non-vacuousness checks.
- **#4 `integer_modulo::…keep_frem`** — **stale assertion**. #7404/#7416 widened integer-modulo eligibility to any nonzero `Expr::Integer` divisor, and that is sound: the dividend's 62-bit cap prevents overflow and JS `%`'s sign rule depends only on the dividend. Split the now-eligible `negative`/`out_of_i32` cases into a new test asserting the correct srem path; kept the genuinely-ineligible ones where they were.

Re-pointing two assertions at "the property" rather than "the textual window" is the fix #7494 asked for, and the reason the other two were caught as real bugs is that the question was asked per test instead of assumed.

### Verification

Full `cargo test -p perry-codegen --no-fail-fast`: **6 pre-existing failures → 2**. The remaining two — `typed_f64_receiver_method_clone_raw_loads_after_composed_guards` (#7506) and `large_local_array_push_inbounds_store_emits_precise_slot_barrier` (#7708) — are the ones explicitly left alone, unrelated causes, still open. Gates 21/21.

PR #7731, opened 2026-08-09
fix(gc): warm_generator_intrinsics must call the tower builder, not a no-op

proggeramlug:
## Summary

`cargo test -p perry-runtime --lib` was red on `main` (blocking every open PR
via the per-PR `--lib --bins` gate), with all three
`gc::tests::runtime_roots::generator_attach_prototype` tests failing:

- `attach_prototype_survives_an_alloc_point_copying_minor_inside_the_call`
- `attach_closure_prototype_survives_an_alloc_point_copying_minor_inside_the_call`
- `the_shipped_default_defers_the_trigger_out_of_the_callees_window`

## Root cause

Bisected in an isolated worktree by checking out each of today's three
merges (`c9079531d` → `ca8c0d617` (#7721) → `cbb682d31` (#7723) →
`c156f8a41` (#7724)) and running just this test file:

- `c9079531d` (pre-#7721): all 3 pass
- `ca8c0d617` (#7721, moving-loop poll default flip): all 3 pass
- `cbb682d31` (#7723, no-move window for lazy intrinsic towers +
  `per_test_global!` conversion): **all 3 fail — first bad commit**
- `c156f8a41` (#7724, root-dominance phi fixes): uninvolved, already broken

**#7723 did not break `js_generator_attach_prototype` /
`js_generator_attach_closure_prototype`.** It correctly converted the six
generator-intrinsic-tower `AtomicI64` statics
(`GENERATOR_FUNCTION_INTRINSIC_PTR` and siblings) from plain process-globals
to `per_test_global!`, so every test starts from a guaranteed first-touch
"tower not yet built" state instead of whichever test happened to run first
in the process silently pre-warming it for everyone else. That is exactly
what `gc::tests::lazy_intrinsic_towers` needs to gate #7251, and it is the
right fix for a real cross-test hazard.

What it exposed was a latent bug in this test file's own setup helper:

```rust
fn warm_generator_intrinsics() {
    let _ = crate::object::js_generator_attach_prototype(
        f64::from_bits(crate::value::TAG_UNDEFINED),
        0,
    );
}
```

`js_generator_attach_prototype` returns at its very first line for any
non-pointer `obj` (`!jv.is_pointer()`), so this call never reached
`generator_prototype_ptr` / `ensure_generator_intrinsics()` — it always was a
no-op. Pre-#7723 that didn't matter: the tower statics were process-global,
and by the time this test ran, some earlier test in the same binary had
almost always already built the tower, so the *real* call under test found it
cached regardless of what `warm_generator_intrinsics()` did.

Once the towers became per-thread (per-test), that accidental priming went
away. With nothing pre-built, the real call now pays the
dozens-of-allocations tower build itself, inside `build_generator_tower`'s
`GcSuppressScope` (#7251's own no-move window for that build — also correct).
Traced with instrumented `gc_check_trigger` / `GcSuppressScope` logging: on
the last-good commit the real call's first allocation reaches
`gc_check_trigger` unsuppressed and services the test's injected trigger
directly (1 relevant call); on #7723 the ~1800-call tower build runs entirely
inside one suppression window first, swallowing every trigger check that
happens during it, and by the time the window closes the following
allocations no longer need a new arena block, so the trigger is never
serviced — "subject not live" in the two alloc-point tests, and the
deferral witness's `GC_SAFEPOINT_PENDING` never gets set in the third.

## Judgement call

This is neither "the PR broke behaviour" nor "the test's premise is now
wrong" — #7723's production and test-isolation changes are both correct and
their invariants are unaffected. It's a pre-existing bug in this test file's
own helper, masked by accident until the isolation fix removed the accident.
Fixed the helper to call the real builder
(`crate::object::ensure_generator_intrinsics()`, the same one
`lazy_intrinsic_towers.rs` uses) instead of a call shape that never reached
it. The liveness/deferral assertions in the three tests are untouched.

## Test plan

- [x] `cargo test -p perry-runtime --lib generator_attach_prototype --
  --test-threads=1` — 3/3 pass, repeated 3x for flake-checking
- [x] `cargo test -p perry-runtime --lib lazy_intrinsic_towers --
  --test-threads=1` — 2/2 pass (adjacent #7723 gate, unaffected)
- [x] `cargo test -p perry-runtime --lib --no-fail-fast -- --test-threads=1`
  — 1961 passed, 0 failed, 4 ignored (one unrelated timing-sensitive test,
  `promise::keyed_table::tests::settling_many_keys_is_not_quadratic`, flaked
  once under host load 40–70 and passed clean on every isolated re-run;
  confirmed unrelated to this change)
- [x] `cargo fmt --all -- --check` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed generator prototype initialization in runtime tests, preventing related failures when intrinsic statics are isolated per test.
* **Documentation**
  * Added a changelog entry describing the generator initialization fix.
  * Updated the documented project version to 0.5.1422.
* **Chores**
  * Bumped the workspace version to 0.5.1422.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1421
+**Current Version:** 0.5.1422
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1421"
+version = "0.5.1422"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1421"
+version = "0.5.1422"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1421"
+version = "0.5.1422"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1421"
+version = "0.5.1422"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1421"
+version = "0.5.1422"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1421"
+version = "0.5.1422"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1421"
+version = "0.5.1422"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1421"
+version = "0.5.1422"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1421"
+version = "0.5.1422"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1421"
+version = "0.5.1422"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1421"
+version = "0.5.1422"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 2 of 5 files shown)


proggeramlug:
## Merging as v0.5.1422 — this unblocks `main`

`cargo test -p perry-runtime --lib` is gated per-PR, so these three red tests were blocking every open PR. Bisected properly rather than guessed — three builds, one per suspect:

| commit | result |
|---|---|
| `c9079531d` (pre-#7721) | 3/3 pass |
| `ca8c0d617` (#7721, poll default flip) | 3/3 pass |
| **`cbb682d31` (#7723)** | **3/3 fail — first bad commit** |
| `c156f8a41` (#7724) | uninvolved |

### The diagnosis is the interesting part, and it is not "#7723 broke it"

`warm_generator_intrinsics()` called `js_generator_attach_prototype(TAG_UNDEFINED, 0)`, expecting it to build the generator tower as a side effect. **It never did** — that function returns at its very first line for any non-pointer `obj`. The warm-up was dead code from the day it was written.

It went unnoticed because the six tower `AtomicI64` statics were process-**global**: some earlier test in the same binary had almost always already built the tower, so the real call under test found it cached no matter what the warm-up did. The test passed for a reason that had nothing to do with its setup.

#7723 converted those statics to `per_test_global!` precisely so every test starts from a guaranteed first-touch state — which is its whole point for `lazy_intrinsic_towers`. That removed the accidental cross-test priming and exposed the dead helper. The real call then paid the ~1800-call tower build itself, inside `build_generator_tower`'s `GcSuppressScope`, which swallowed the injected arena trigger; by the time suppression lifted, no allocation needed a new block, so the trigger was never serviced and `GC_SAFEPOINT_PENDING` stayed false.

**So both of #7723's changes are correct.** This is a latent bug in the test's own setup, unmasked by an isolation fix — which is exactly what test isolation is *for*. Fixed by calling `crate::object::ensure_generator_intrinsics()` directly, the same builder `lazy_intrinsic_towers.rs` uses.

**The liveness assertions are untouched**, and they are why any of this was found: `"LIVE SUBJECT: the trigger must actually have been due and deferred. Without this the test also passes when nothing was armed at all."` Without that guard these three would have gone green on a build where nothing was armed, and the dead warm-up would still be there.

### Verification

`cargo test -p perry-runtime --lib --no-fail-fast`: **1961 passed, 0 failed, 4 ignored.** The three target tests pass 3/3 in isolation with `--test-threads=1`.

### On my side

I merged #7721/#7723/#7724 without running `cargo test -p perry-runtime --lib` — my local gate covered the `lint` job's 19 steps but not the `cargo-test` job at all, so it reported green while a gated check was red. That is the same defect as a partial gate reporting on commands it never ran, one level up. The gate now runs `perry-runtime --lib` and `perry-codegen --lib` too, and it caught this immediately on the next PR.

Issue #7055, opened 2026-07-30
gc: the relocating minor produces a deterministic wrong answer in the SHIPPED default on an ordinary event-loop workload


proggeramlug:
Found while measuring `PERRY_GC_MOVING_LOOP_POLLS` (measurement writeup: ISSUE_LINK). Reporting separately because it is not a poll bug and it is more urgent than the measurement.

## Summary

An ordinary event-loop-driven TypeScript program produces a **deterministic wrong answer in the shipped default configuration**. The same binary produces the Node-correct answer when the GC is configured so that no *relocating* minor runs.

The discriminator is exactly `copied_objects > 0`. Not `minor_cycles > 0`, not the loop poll, not the nursery cap.

## Reproducer

`w5_srv_scale.ts` — an event-loop request pump; each "request" builds 1,500 small objects in one synchronous stretch, folds them into a checksum, then yields via `setImmediate`. Request count from `argv[2]`. Full source at the bottom.

```
$ perry w5_srv_scale.ts -o w5
$ ~/node-v26.5.0-darwin-arm64/bin/node --expose-gc --experimental-strip-types w5_srv_scale.ts 150
checksum:-341887099

$ ./w5 150                                          # shipped default
checksum:-90240474                                  # WRONG

$ PERRY_GC_MOVING_LOOP_POLLS=0 ./w5 150             # no relocating minor
checksum:-341887099                                 # correct
```

## The discriminator is relocation, not the flag

Same binary throughout; only the runtime environment differs. `checksum` is deterministic across repeats in every arm (5 repeats each, plus a warmup).

| arm | env | minors @150 / @600 | **copied objects** | checksum @150 | checksum @600 |
|---|---|---|---|---|---|
| shipped | — | 8 / 56 | 915,961 / 11,661,882 | `-90240474` ❌ | `1986912948` ❌ |
| cap raised | `PERRY_GC_SCAVENGE_NURSERY_MB=128` | 0 / 3 | **0 / 0** | `-341887099` ✅ | `1735266323` ✅ |
| polls off | `PERRY_GC_MOVING_LOOP_POLLS=0` | 0 / 1 | **0 / 0** | `-341887099` ✅ | `1735266323` ✅ |
| polls off + 16 MB cap | `PERRY_GC_MOVING_LOOP_POLLS=0 PERRY_GC_SCAVENGE=1` | 8 / 60 | 1,045,097 / 11,351,683 | `-90240474` ❌ | `1986912948` ❌ |

[...]

PR #7732, opened 2026-08-09
gc: reload js_closure_get_capture_bits results across a collection point (#7725)

### LINKED ISSUE #7725: gc: js_closure_get_capture_bits's result is never re-entered into a protected domain (2 unrooted:capture hits, gc-root-dominance-statepoints' --max-unrooted referent)
## `js_closure_get_capture_bits`'s result is never re-entered into a protected domain (2 unrooted:capture hits, `--max-unrooted` referent)

Split out of #7664, which fixed the checker's phi false positives and the 3
`unrooted:global` hits (module-global receivers/arguments held across a
sibling operand's allocation). The remaining `--max-unrooted 2` on
`gc-root-dominance-statepoints` is these two hits, and this issue is their
referent.

### The shape

`js_closure_get_capture_bits(closure_ptr, idx)` returns a raw `i64` — the
captured value's bit pattern, which may be a NaN-boxed heap reference. Two
things protect a value like that under the native (RS4GC/statepoint)
lowering: re-entering the `ptr addrspace(1)` tracked domain (what
`codegen/closure.rs` already does for `%this_closure` itself — `or …,
POINTER_TAG` then `inttoptr … to ptr addrspace(1)`), or a temp root. The
generic "read a captured value" call sites do neither:

- `expr/mod.rs::load_boxed_local_pointer` and its siblings in
  `literals_vars.rs` / `array_push.rs` / `instance_misc1.rs` /
  `lower_array_method.rs` return the bare `i64`/`double` and leave it exactly
  that across whatever the calling expression's evaluation does next.
- The synthesized dynamic-parent implicit constructor closure
  (`constructor(...args) { super(...args); … }` for a class expression whose
  parent is a runtime value with no statically-known ctor,
  `test_gap_class_expr_dynamic_parent_ctor.ts`) reads a captured value at the
  top of the closure and uses it ~60 lines later as the callee of a second
  `js_new_function_construct`, across `js_object_alloc_class_inline_keys`
  AND the class's own user constructor body.

Confirmed register-by-register on both corpus hits:

```
test_gap_class_expr_dynamic_parent_ctor_ts::__closure_21
  source (capture): %r61 = call i64 @llvm.experimental.gc.result.i64(token %statepoint_token)
  stale use        : js_new_function_construct's own statepoint, ~60 lines below
  across safepoint : js_closure_get_capture_bits, js_ctor_return_override,
                      js_new_function_construct, js_object_alloc_class_inline_keys,
                      <AnonShape>_constructor (user code)

test_gap_computed_key_method_nested_this_ts::__closure_9
  source (capture): %r145 = call i64 @llvm.experimental.gc.result.i64(token %statepoint_token4)
  stale use        : %r17 = fadd double %r167, %r15
  across safepoint : js_number_coerce
```

Neither register ever appears in a `"gc-live"` bundle anywhere in its
function.

### Why this isn't a signature/ABI change (correction to the #7664 triage)

The prior triage read this as needing "the callee's own closure pointer to
be a tracked root first, which is a signature/ABI change." Closer reading:
the closure pointer itself (`%this_closure`) is fine here — it's already
correctly re-entered into the tracked domain and relocated
(`%rs4gc.s2`/`.relocated` chains appear in both traces). What's unprotected
is the *return value* of `js_closure_get_capture_bits`, which is a completely
separate register with no relationship to the parameter.

### Proposed fix

Mirror the string-handle precedent already in `crates/perry-codegen/src/root_reload.rs`:
treat a `js_closure_get_capture_bits` call as a *reloadable* source — re-emit
the same call (same closure pointer, same index) rather than re-deriving from
a load. Soundness argument is the same shape as the existing handle-global
one: re-calling it can only observe a different value if a
`js_closure_set_capture_bits` to the *same* index ran in the window, which is
the same store side-condition the pass already tracks for shadow slots and
handle globals (`stores_to` in `root_reload.rs`'s `Facts`).

The engineering gap: `Facts` currently only models *loads*
(`load_of: Option<(String, LlvmType, String)>`) as reloadable sources, not
calls. Extending it to recognize a specific call pattern as reloadable (and
re-emitting a *call* rather than a *load* at each protected use) is a real
slice of work, not a one-line change — hence splitting it out rather than
rushing it into #7664.

### Scope check before starting

Confirm via the corpus (`scripts/gc_root_dominance_corpus.sh … --lowering
native` + the checker's `-v` output) whether other `unrooted:capture` shapes
exist beyond these two before generalizing the fix — the two here are both
"read once, held long, no phi, no reassignment in window," but the fix
should be verified against the full corpus rather than just these two
fixtures.

proggeramlug:
## Summary

Closes #7725, the follow-up #7724 split out: the last 2 `unrooted:capture` hits on `gc-root-dominance-statepoints`, both `js_closure_get_capture_bits`'s **return value** never being re-entered into a protected domain (unlike `%this_closure` itself, already protected via `current_closure_slot`, #7055).

- `test_gap_class_expr_dynamic_parent_ctor::__closure_21` — a captured dynamic-parent-class reference read at the top of a synthesized implicit constructor, used ~60 lines later as `js_new_function_construct`'s callee, across `js_object_alloc_class_inline_keys` and the class's own user constructor.
- `test_gap_computed_key_method_nested_this::__closure_9` — a captured numeric local held across `js_number_coerce` (which can run a user `Symbol.toPrimitive`).

## Confirming the framing before implementing

The task was to verify #7724's diagnosis — that this needs `root_reload.rs`'s `Facts` to model a **call** as a reloadable source, not just a load — rather than assume it and start coding.

I checked the alternative: root every capture read at its own call site via the existing `TempRootPool` (`rooting/temp_root.rs`). Rejected — it would pay a store + shadow-slot bind on *every* capture read, including the overwhelming majority never held across a collection point, where `root_reload.rs`'s whole design point is to cost nothing where the reload turns out to be unnecessary (LLVM CSEs a redundant one away; a needed one stays because the intervening call is opaque). The generic-read call sites (`literals_vars.rs`, `closure.rs`, `array_push.rs`, `instance_misc1.rs`, `lower_array_method.rs`, `expr/mod.rs`) also have no way to know at emission time whether *this particular* read will cross a later collection — that's exactly the question `root_reload.rs`'s post-hoc CFG walk exists to answer cheaply, and it's the same question it already answers for string-handle globals and shadow-slot masks. So the `Facts`-as-call extension is the right layer, confirmed rather than assumed.

## The fix

`js_closure_get_capture_bits(ptr, idx)` calls whose `ptr` operand already belongs to a reloadable recipe (the closure-ptr shadow-slot load chain #7055 protects) now extend that recipe exactly like the existing transparent `and`/`bitcast` steps — `Facts` gained `capture_get_key`, set when the call's index argument is a compile-time literal (every real emission site formats one directly, confirmed by grep). A reload point re-materializes the *whole* chain: a fresh slot load, a fresh mask, a **fresh call** to `js_closure_get_capture_bits` with the same index, and any further cast — not just a load.

The store side-condition needed its own key rather than the inherited closure-ptr-slot one (that slot is written exactly once, at closure entry, so checking it there would make the invalidation vacuous — a real soundness gap I caught while implementing, not something I noticed after). A capture-GET recipe's `root_ptr` switches to a synthetic `$closure_capture:<idx>` key from that call onward; `js_closure_set_capture_bits` populates the existing `stores_to` field with the same key, reusing the store-invalidation machinery a plain `store` already has rather than adding a parallel one. That required widening the reload pass's grouping key from `recipe[0]` alone to `(recipe[0], root_ptr)`, since before this a chain's `root_ptr` never changed mid-derivation — every pre-existing chain has one `root_ptr` throughout, so the wider key is additive and provably doesn't change any existing (non-capture) grouping.

`raw_facts` (the `invoke`/`try` rendering path) gets the same `stores_to` detection for the SET half; the GET half is deliberately gated on `!is_invoke`, since `materialize` re-emits a recipe step as a plain mid-block instruction and an `invoke` is a terminator — cloning one in would be invalid IR. In practice this call is never seen there except as an invoke (every real emission goes through the typed `LlInst::Call` form otherwise), so the guard is defensive rather than load-bearing today.

## Before / after (gated mode, both lowerings)

| | before | after |
|---|---|---|
| `--statepoints --moving-only`, native corpus | unrooted **2**, stale 0 | unrooted **0**, stale 0 |
| `--lowering shadow` (dominance / `--unrooted-allocas` / `--stale-registers`) | clean | clean (6 pre-existing `--stale-registers` hits, unrelated to capture-bits, unchanged, well under the 39 budget) |

Reached **0**, so `gc-root-dominance-statepoints`' `--max-unrooted` is **deleted** (not set to `0`) — the flag already defaults to 0, and a budget nobody re-measures is the exact failure mode #7706 fixed for a different gate on this same file: it silently absorbs the next real hazard instead of going red.

## Test plan

- [x] `python3 scripts/gc_root_dominance_check.py --self-test` — clean, including #7724's `phi_safe_edge`/`phi_hazard_edge` fixtures (confirms this change doesn't touch the checker)
- [x] Native corpus (131/131 sources compiled, 0 skipped, `PERRY_RUNTIME_DIR` pinned to the build under test both times): `--statepoints --moving-only --max-unrooted 999 --max-stale 999` → **2 → 0** unrooted, stale held at 0
- [x] Full CI-shaped invocation (`--min-statepoints 15000 --min-live-bundles 8000 --min-relocates 20000 --max-unrooted 0 --max-stale 0 --seeded-violations 40`) passes; all 40 planted violations still caught
- [x] `--lowering shadow` corpus (131/131 compiled): dominance, `--unrooted-allocas --moving-only`, `--stale-registers --moving-only` all clean
- [x] `cargo test -p perry-codegen --lib` — 809 passed (805 on `origin/main` + 4 new tests), 0 failed
- [x] `cargo test -p perry-codegen --no-fail-fast` (the full suite including `crates/*/tests/*.rs`, which per-PR CI does not run) — the same 6 pre-existing failures across the same 4 targets (`large_object_barriers`, `native_proof_buffer_views`, `native_proof_regressions`, `typed_shape_descriptors`) reproduce byte-identically on a fresh `origin/main` checkout; zero new failures
- [x] `cargo fmt --all -- --check` / `scripts/check_file_size.sh` — clean

`root_reload.rs` crossed the 2,000-line file-size cap gaining this fix; its `#[cfg(test)] mod tests` moved to a sibling `root_reload_tests.rs` (the existing `linker.rs`/`type_analysis.rs` `#[path]` idiom), which is also where the four new capture-bits tests landed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed native garbage-collection safety issues involving closure capture reads and updates.
  - Improved value reloading across collecting calls, control-flow paths, loops, invokes, and slot reassignment.
  - Preserved correct behavior for different integer widths, globals, derived values, and multiple operands.

- **Tests**
  - Added comprehensive regression coverage for reload placement, operand renaming, closure captures, and capture-index-specific updates.

- **Documentation**
  - Documented the fixes and verification results.

- **Chores**
  - Updated the application version to 0.5.1425.


proggeramlug:
## Merging as v0.5.1425 — the native lowering is now at ZERO unrooted hazards

I verified the headline claim independently rather than trusting the PR, because a wrong budget makes `main` red for everyone:

```
corpus (native): 131/131 sources compiled, 0 skipped, 151 .ll files
  statepoints: 32694   non-empty live bundles: 19552
gc_root_dominance_check.py --statepoints --moving-only   (no budget flag)
  → within budget: unrooted 0 <= 0      exit 0
```

**21 → 0**, and `--max-unrooted` is **deleted** rather than set to 0, per #7706's precedent. A flag that says "the budget is 0" and a flag that isn't there are the same today and different the first time someone edits the default.

### Why the harder fix was the right one

The obvious alternative — root every capture read at its own site via `TempRootPool` — would tax **every** capture read, and the overwhelming majority never cross a collection point. `root_reload.rs` exists precisely to cost nothing when the reload turns out unneeded, and the generic read sites can't know at emission time whether a given read will later cross a safepoint. That is the question the pass's post-hoc CFG walk already answers for string-handle globals and shadow-slot masks. Extending it to treat a `js_closure_get_capture_bits(ptr, idx)` call as a transparent link in a reloadable recipe is the same mechanism, not a new one.

### The bug caught during implementation is the best part

Inheriting the closure-ptr-slot's invalidation key would have made invalidation **vacuous** — that slot is written once, at entry, so nothing would ever invalidate the recipe. Caught before shipping, and fixed by switching a capture-GET recipe to a synthetic `$closure_capture:<idx>` key from that call onward, with `js_closure_set_capture_bits` populating `stores_to`. That required widening the pass's grouping key from `recipe[0]` to `(recipe[0], root_ptr)`, since no chain's root had ever changed mid-derivation before.

A reload pass whose invalidation never fires is exactly the shape of bug this project keeps finding — it would have passed every test while doing nothing.

### Verification

`--self-test` clean with seeded violations **40/40** still caught. `cargo test -p perry-codegen --lib` **809/809** (805 baseline + 4 new). Shadow lowering clean. `root_reload.rs` crossed the 2000-line cap gaining this, so its tests moved to a sibling `root_reload_tests.rs` via the existing `#[path]` idiom.

Gates 21/21. (One gate run showed a `perry-runtime --lib` failure at host load 50 — clean on re-run at **1962 passed, 0 failed**, and this PR touches codegen only.)

PR #7734, opened 2026-08-09
fix(hir,stdlib): narrow the #7726 spread bail and complete the querystring bridge

proggeramlug:
Follow-up to #7726 (issue #7720). **Opening as a draft until the baseline gap-suite arm finishes**; the substance below is already measured.

#7726 was merged before my validation finished. It carried two regressions. This corrects them at the root and narrows the predicate to what is measured, not derived.

## How this was measured

A 32-case A/B matrix over the node-core spread surface, run on one host, three ways per case:

* **base** — this tree with the guard forced to `false` (byte-identical to pre-#7726 lowering, same stdlib in both arms so the comparison isolates the HIR change)
* **branch** — this tree
* **oracle** — node 26.5.1

| | #7726 as merged | this PR |
|---|---|---|
| FIXED | 17 | **17** |
| same | 10 | **14** |
| wrong → wrong | 3 | **1** |
| **correct → wrong** | **2** | **0** |

## 1. `querystring.escape(...args)` — regressed to `undefined`

The root cause is older and wider than the spread bail. `nm_dispatch_querystring` advertises the whole set:

```rust
("querystring", "unescapeBuffer" | "unescape" | "escape" | "stringify" | "encode" | "parse" | "decode") => { …JS_NATIVE_QUERYSTRING_DISPATCH… }
```

…but the stdlib bridge it calls, `js_querystring_native_dispatch`, implemented **only** `unescapeBuffer` and fell to `_ => undefined` for the other six. So on `main`, before any of this, every *indirect* form was already silently `undefined`:

```ts
qs.escape("a b")                      // "a%20b"  — static dispatch, always fine
const e = qs.escape; e("a b")         // undefined
const d: any = qs; d.escape("a b")    // undefined
```

#7726 only routed spread calls onto that pre-existing hole. Every `js_querystring_*` entry point already existed; only the bridge was missing. Wiring the six names up (`encode`/`decode` are Node's aliases for `stringify`/`parse`) fixes the regression **and** the captured/dynamic forms in one go.

## 2. `fs.promises` / `dns.promises` are not dispatch buckets

The predicate recursed through any sub-namespace receiver and accepted any `<module>/<export>` that happened to be a node-core module name. But `nm_module_index` has **dotted** tags only for `path.posix`, `path.win32`, `util.types`, `crypto.subtle`/`webcrypto`, `punycode.ucs2`. There is no `fs.promises` bucket, so diverting them produced:

* `dns.promises.lookup(...args)` → silent `undefined` (was a throw)
* `import { promises } from "node:fs"; promises.readFile(...args)` → synchronous `TypeError: value is not a function` where a rejected promise used to arrive — the worst of the four, since a `.catch()` no longer catches it

Replaced the derivation with an explicit allowlist, and rejected the slash sub-module tags. The direct import (`import fsp from "node:fs/promises"`) needs no help from the bail — it already reached the generic tail before #7726 (measured: identical HIR and `ENOENT` on both arms) — so excluding the slash tags costs nothing.

## Known and deliberate

`events.listenerCount(...args)` still turns a bogus `ERR_INVALID_ARG_TYPE` throw into `undefined`. `nm_dispatch_events` implements only `init` and `EventEmitterAsyncResource`, so there is no arm to reach; both the old and new behaviour are wrong (node returns a count). Completing that dispatcher is the same shape of change as the querystring bridge here and deserves its own PR. Excluding `events` from the bail would be a one-line denylist entry if you would rather have the loud failure in the meantime — say the word.

## Tests

* `sub_namespace_allowlist_is_the_runtime_bucket_set` — pins the allowlist against **exactly the re-derivation that shipped** (asserts `fs.promises` / `dns.promises` / `stream.web` are *not* in it)
* `bucketless_sub_namespaces_keep_their_lowering` — the two HIR verdicts that changed. `fs.promises.readFile(...)` is deliberately **absent**: it already reached the generic tail before #7726, so asserting on it would test nothing
* `node-suite/querystring/aliases/dynamic-dispatch.ts` — static / captured / dynamic / spread forms of six querystring methods, byte-compared against node; verified identical
* `cargo test -p perry-hir --lib` — 9/9 in the spread suite, 287 in the crate

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Node.js query-string compatibility for dynamic calls to `escape`, `unescape`, `stringify`/`encode`, and `parse`/`decode`.
  * Prevented unsupported namespace variants from being routed through incorrect runtime handling.
  * Preserved expected behavior for unknown query-string methods.

* **Tests**
  * Added coverage for static, dynamic, captured, type-erased, and spread query-string calls.

* **Documentation**
  * Updated the documented release version and changelog.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1425
+**Current Version:** 0.5.1426
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1425"
+version = "0.5.1426"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1425"
+version = "0.5.1426"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1425"
+version = "0.5.1426"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1425"
+version = "0.5.1426"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1425"
+version = "0.5.1426"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1425"
+version = "0.5.1426"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1425"
+version = "0.5.1426"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1425"
+version = "0.5.1426"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1425"
+version = "0.5.1426"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1425"
+version = "0.5.1426"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1425"
+version = "0.5.1426"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1425"
+version = "0.5.1426"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1425"
+version = "0.5.1426"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1425"
+version = "0.5.1426"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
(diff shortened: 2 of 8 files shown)


proggeramlug:
## Validation complete — flipping out of draft

**Gap suite, both arms, same host** (511 tests; `base` = this tree with the guard forced to `false`, i.e. pre-#7726 lowering):

| | base | branch |
|---|---|---|
| pass | 485 | 484 |
| output mismatch | 20 | 20 |
| crash | 6 | 7 |

The 20-mismatch **sets are byte-identical** — all pre-existing, and none of those 20 files contains a spread call argument at all (only rest-param declarations and comments), so the guard cannot reach them. They are this host's known profile: npm-package tests needing `node_modules` (moment/dayjs/cron/slugify/backoff/ratelimiter), the zlib link flake, and socket-sandbox hangs.

The single crash-count difference is `test_gap_gc_alloc_point_no_move`, which crashed only in the branch run — that run overlapped two other parity suites and a cargo build on the same box. Re-run in isolation it is **3/3 clean on both arms and byte-identical to node**, so it is load-induced, not a regression.

**Final A/B matrix (32 cases, base vs this branch vs node 26.5.1): 17 fixed / 14 same / 1 wrong→wrong / 0 regressed.** The one wrong→wrong is the `events.listenerCount` case described above.

Other checks: `cargo test -p perry-hir --lib` 287 pass (9 in the spread suite), `cargo fmt --all --check` clean, `scripts/check_file_size.sh` clean, and `node-suite/querystring/aliases/dynamic-dispatch.ts` byte-identical to node.

proggeramlug:
## Merging as v0.5.1426 — and I should have caught these before merging #7726

I merged #7726 on its own validation plus my verification of its vacuity finding. I did **not** run the node-core spread surface myself, and your A/B matrix found two `correct → wrong` regressions in it. That's on me, not on you for opening the follow-up.

The matrix is the right instrument, and it isolates the variable properly — same stdlib in both arms, guard forced `false` for base, so the comparison is purely the HIR change:

| | #7726 as merged | this PR |
|---|---|---|
| FIXED | 17 | **17** |
| same | 10 | **14** |
| wrong → wrong | 3 | **1** |
| **correct → wrong** | **2** | **0** |

Holding the 17 fixes while taking the regressions to zero is the outcome that matters.

### The querystring finding is much older than #7726, and I verified it

`nm_dispatch_querystring` advertises seven names. `js_querystring_native_dispatch` on `main` is:

```rust
match name {
    "unescapeBuffer" => { … }
    _ => undefined,
}
```

One arm. So every **indirect** form — `const e = qs.escape; e("a b")`, `const d: any = qs; d.escape(...)` — has been silently returning `undefined` for as long as that bridge has existed. #7726 didn't create the hole, it routed spread calls onto it. Wiring the six names up (with `encode`/`decode` as Node's aliases for `stringify`/`parse`) fixes the regression *and* the captured/dynamic forms at once, which is the right root to fix at.

### The `fs.promises` case is the worse of the two

`import { promises } from "node:fs"; promises.readFile(...args)` went from a **rejected promise** to a synchronous `TypeError: value is not a function` — so a `.catch()` no longer catches it. Turning an awaitable failure into a synchronous throw is a nastier regression than a wrong value, because it breaks the caller's error handling rather than its result.

Replacing the derivation with an explicit allowlist is right: `nm_module_index` has dotted tags only for `path.posix`, `path.win32`, `util.types`, `crypto.subtle`/`webcrypto`, `punycode.ucs2`, and deriving "is this a bucket?" from "does this name look like a module?" was never going to track that set.

### Two things I want to note about the tests

`sub_namespace_allowlist_is_the_runtime_bucket_set` pins the allowlist against **exactly the re-derivation that shipped**, asserting `fs.promises`/`dns.promises`/`stream.web` are *not* in it — so the next person who re-derives instead of listing gets a red test rather than a silent regression.

And `fs.promises.readFile(...)` is deliberately **absent** from `bucketless_sub_namespaces_keep_their_lowering`, because it already reached the generic tail before #7726 — asserting on it would test nothing. Leaving a case out for a stated reason is worth more than padding the test with it.

### On `events.listenerCount`

Leave it as-is for now and give it its own PR, as you suggest. `nm_dispatch_events` implements only `init` and `EventEmitterAsyncResource`, so both behaviours are wrong and neither is a regression from this work — a denylist entry would trade one wrong answer for a different wrong answer while making the predicate less honest about what it covers.

`cargo test -p perry-hir --lib`: **290 passed, 0 failed.** Gates 21/21.

Issue #7469, opened 2026-08-05
perf(runtime): allocation path spends 34% of self time in _tlv_get_addr — 24× behind Node on object churn with the collector already idle


proggeramlug:
## Summary

With the GC collection problems closed (#7377 nursery cap + scavenge, #7432 adaptive
tenuring, #7443 old-gen hole free list, #7449 recycled-block pool, #6893 per-class
layout interning), allocation-heavy programs are no longer limited by the collector —
they are limited by the **mutator's allocation path**. On `churn.ts`, total GC pause is
now **0.03 s out of a 4.33 s run (0.7%)**, yet Perry is still **24× slower than Node** on
the same program.

A `sample(1)` profile of that run puts **34.2% of all self time in `_tlv_get_addr`** —
macOS's thread-local-variable accessor. The allocation and write-barrier paths resolve
thread-local storage over and over per allocation, and that resolution now costs more
than the allocation work itself.

This ticket covers closing that gap. Two workstreams, A being much the larger.

## Where things stand (measured 2026-08-05, `351742d30`, quiet 10-core M-series, load 2–6)

| bench | Perry | scriptc | node | Perry RSS | scriptc RSS | node RSS |
|---|---|---|---|---|---|---|
| churn — 20M short-lived 2-field objects, ~1000 live | 4.33 s | 0.80 s | **0.18 s** | 24 MB | 2 MB | 86 MB |
| retain — 3M live records | 4.65 s | 0.11 s | **0.15 s** | 423 MB | 118 MB | 364 MB |
| tree — 21M nodes built/dropped, ~524k live | 10.42 s | 7.52 s | **0.44 s** | 193 MB | 74 MB | 313 MB |
| cycles — 2M dead reference cycles | 1.24 s | 0.29 s | **0.07 s** | 29 MB | 2 MB | 84 MB |
| deeplist — 1M-node linked list | 1.80 s | 124.5 s | **0.10 s** | 163 MB | 148 MB | 145 MB |

Best-of-3 wall, `/usr/bin/time -l` peak RSS. Programs in `gc-handoff/bench/`.

GC work per program is now small — `PERRY_GC_TRACE=1`, summing `copying_nursery`:

| bench | cycles | copied | object-copies | total pause | max pause |
|---|---|---|---|---|---|
| churn | 105 | 0.004 GB | 0.0 M | **0.03 s** | 2.3 ms |
| tree | 43 | 0.017 GB | 0.2 M | 6.79 s | 255.6 ms |
[...]

PR #7735, opened 2026-08-09
perf(gc): the loop back-edge poll's no-work path becomes one global load (#7721)

proggeramlug:
## What regressed, and it was not the pointer-field work

Three all-numeric benchmarks lost 15–30 % in the #7686..#7721 window. The obvious
suspect was #7686/#7698's typed pointer-field stores generalising a path that used
to have an all-numeric special case. Measured, that hypothesis is **refuted**: a
symbolicated profile of `churn_alloc_big` at 0.5.1384 and at main has the same shape
symbol for symbol —

| symbol | 0.5.1384 | main |
|---|--:|--:|
| `gc::layout::init_typed_shape_layout` | 23.0 % | 21.8 % |
| user `AnonShape` constructor | 24.0 % | 23.3 % |
| `perry_fn_..._chunk` | 20.1 % | 20.9 % |
| `layout_tables::layout_forget_object` | 7.1 % | 8.1 % |

Nothing in the layout or store path grew. What appeared instead were two symbols not
present in the 0.5.1384 profile at all: **`js_gc_loop_safepoint` at 8.2 %** and
**`_tlv_get_addr` back at 8–9 %**, having been driven to 0 % by #7469. Both arrive
with **#7721**, which turned the moving-loop back-edge poll on by default.

## Mechanism

#7721 was right about the collector — the poll is the only precise
nursery-collection point a compute-only program ever reaches, and without it every
nursery collection happens at the register-imprecise allocation point where #7682
made it correctly non-moving. That is worth `tree_wide` 7.26 s instead of 2.11.
What was wrong was the poll's **price**.

A poll is emitted at every allocating loop back-edge: 20 M of them in
`bench/churn_alloc.ts`. So its no-work path is a per-iteration cost of the language,
paid whether or not a collection is ever due — and that path was an out-of-line
`extern "C"` call into

1. `gc_moving_loop_polls_enabled()` — a `OnceLock` acquire load,
2. `note_loop_poll_reached()` — an **unconditional** `AtomicU64::fetch_add` on a
   process-shared line,
3. `GC_SAFEPOINT_PENDING.with(Cell::get)` — a thread-local read, which on Darwin is
   a CALL to `_tlv_get_addr`, Mach-O having no local-exec TLS model,
4. `gc_zeal_enabled()` — a second `OnceLock` acquire load,

plus the caller-side spill/reload the opaque call forces. ~3 ns per back-edge, which
is the regression to the millisecond: 20 M × 3 ns = 60 ms against a `churn_alloc`
gap of 51.5 ms.

## The fix

`gc/poll_arm.rs` adds `PERRY_GC_POLL_ARMED`, a plain process-global `AtomicU32`
counting the reasons `js_gc_loop_safepoint` must do more than return. **Zero is a
proof the poll is a no-op**, so both ends answer on one ordinary load: codegen emits
the load inline and branches around the call, and the runtime entry point re-checks
it so a module from any other emission path still gets the cheap answer. The emitted
guard, with the address hoisted into the preheader:

```
ldr  w8, [x26]        ; x26 = &PERRY_GC_POLL_ARMED, loop-invariant
cbnz w8, .gcpoll
```

The word is a deliberate conservative SUPERSET. It is process-global — a
thread-local would reintroduce the `_tlv_get_addr` this removes — so it counts
threads with a deferral outstanding, and a poll on thread B can be woken by a
deferral on thread A and find nothing to do. The unsound direction is the word
reading zero while a deferral is outstanding, which would strand that collection
until an event-loop boundary a compute-only program never reaches. Hence
`GC_SAFEPOINT_PENDING` now has exactly **one writer**, `policy::set_safepoint_pending`,
which moves both representations together.

The load is **volatile**: the runtime writes this word from calls LLVM cannot see
through, and a guard whose load got hoisted out of its loop would read a stale zero
and silently stop draining — #7721's failure mode returning as a codegen bug instead
of a default. One `ldr` either way.

Zeal keeps the word armed for the life of the process, because `PERRY_GC_ZEAL`'s
contract is a collection at *every* safepoint, not only at ones already deferred.
`ZealGuard` mirrors it so a unit test under zeal cannot silently poll into a no-op,
and `loop_polls_reached()` now documents that it is exhaustive exactly under zeal —
the one place `zeal_verdict` reads it.

## Measurements

Best-of-5, all four arms interleaved in one session, quiet M1 bench host, `PERRY_NO_AUTO_OPTIMIZE=1` with a
pinned `PERRY_RUNTIME_DIR`; every output verified byte-identical to
`node --experimental-strip-types` before timing.

| bench | 0.5.1384 | main | **this** | guard |
|---|--:|--:|--:|---|
| churn_alloc | 0.367 | 0.420 | **0.376** | — |
| push_cls | 0.351 | 0.409 | **0.356** | — |
| push_num | 0.132 | 0.178 | **0.144** | — |
| churn | 0.666 | 0.458 | **0.419** | <= 0.46 |
| churn_read | 0.352 | 0.023 | **0.023** | <= 0.03 |
| cycles | 0.803 | 0.196 | **0.192** | <= 0.20 |
| deeplist | 1.148 | 0.320 | **0.315** | <= 0.35 |
| tree | 5.972 | 1.647 | **1.634** | <= 1.70 |
| tree_wide | 12.38 | 2.111 | **2.121** | <= 2.20 |

Every protected benchmark holds; `churn` improves as well.

GC behaviour is unchanged where it matters: `PERRY_GC_TRACE=1 ./n_churn` runs
**105 minor cycles in both arms**, positive reclamation on every cycle, max pause
3.63 ms → 1.78 ms. `gc-handoff/apps/iso_miss.ts` prints `checksum 437840 misses 0`.

## What is left, and why

The residual to 0.5.1384 is 9.4 ms on `churn_alloc`, 6.2 on `push_cls`, 12.6 on
`push_num` — 0.5–0.6 ns per back-edge, i.e. exactly the two guard instructions,
confirmed by disassembly rather than inferred. It is not overhead that can be
tightened away: it is the price of the poll existing, and the poll is what makes the
nursery evacuate. The 0.5.1384 numbers came from a moving minor at the allocation
point, which #7682 removed as unsound — compiling and running today's main with
`PERRY_GC_MOVING_LOOP_POLLS=0`, the 0.5.1384 configuration, gives `churn_alloc`
**0.91 s**, not 0.36. Driving the guard to one instruction would mean a
signal-backed polling page (`ldr wzr, [x26]` + an mprotect'd page); not proportionate
to 5 ms, so it is noted rather than done.

## Gates run

- `perry-runtime --lib gc::` — 725 pass, plus the 6 new ones.
- `perry-codegen` — full suite, including `loop_safepoint_purity` (8/8).
- `gc_root_dominance_corpus.sh` + both gated checker modes: 131/131 sources, 0
  skipped, **0 violations** in `--moving-only` (40/40 seeded violations caught) and
  **0** in `--unrooted-allocas`.
- `check_file_size.sh`, `addr_class_inventory.py`, `cargo fmt --all -- --check`.

Two failures reproduce identically with this branch's sources reverted to
`origin/main` and are not from this change: the three
`generator_attach_prototype` cases (fixed by #7731, now merged) and
`large_object_barriers::large_local_array_push_inbounds_store_emits_precise_slot_barrier`.

## Summary by CodeRabbit

* **Performance**
  * Improved garbage-collection loop performance by skipping unnecessary safepoint checks when no collection work is pending.
  * Safepoint polling remains immediately available when collection activity or diagnostic modes require it.

* **Reliability**
  * Improved consistency when deferring and resolving garbage-collection safepoints.
  * Preserved continuous polling during GC stress and diagnostic modes.

* **Tests**
  * Added coverage for poll activation, deactivation, deferred collections, and diagnostic-mode behavior.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1428
+**Current Version:** 0.5.1429
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1428"
+version = "0.5.1429"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1428"
+version = "0.5.1429"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1428"
+version = "0.5.1429"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1428"
+version = "0.5.1429"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1428"
+version = "0.5.1429"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1428"
+version = "0.5.1429"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1428"
+version = "0.5.1429"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1428"
+version = "0.5.1429"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1428"
+version = "0.5.1429"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1428"
+version = "0.5.1429"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1428"
+version = "0.5.1429"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1428"
+version = "0.5.1429"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1428"
+version = "0.5.1429"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1428"
+version = "0.5.1429"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1428"
+version = "0.5.1429"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1428"
+version = "0.5.1429"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1428"
+version = "0.5.1429"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1428"
+version = "0.5.1429"
 dependencies = [
  "perry-ffi",
  "reqwest",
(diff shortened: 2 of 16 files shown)


proggeramlug:
## Merging as v0.5.1429 — reconciled with #7729 by hand, not auto-merged

The audit predicted this collision precisely: **#7729 and this PR rewrite the same lines of `js_gc_loop_safepoint()`**, from the identical pre-image, and neither body mentioned the other. #7729 landed first, so I reconciled this side deliberately.

The conflict turned out to be **purely additive** — #7733's major-pacing tests and this PR's poll-arming tests appended at the same point in `gc/tests/triggers.rs` — so both sides were kept. Then I verified the two mechanisms actually coexist rather than assuming it:

- `cargo test -p perry-runtime --lib`: **1975 passed, 0 failed**
- zeal family: **13 passed**; poll_arm family: **3 passed**
- **`scripts/gc_instrument_smoke.sh`: all 6 arms PASS, exit 0** — including #7729's brand-new arm 6 (the zeal-pacing termination gate) running against this PR's arming word. That is the cross-check that matters: `PASS: instruments inert when off (0 retirements), live when on`.

The semantics are compatible for a reason worth stating: this PR keeps the arming word nonzero for the life of the process under zeal, which is exactly what #7729's allocation-paced stride needs — every poll must still reach the runtime for the stride check to run.

### The design is the good part

`PERRY_GC_POLL_ARMED` is a **counter of reasons**, not a boolean, so independent reasons compose. It **starts at 1** — the seed — because resolving "should this be armed?" costs exactly what the word exists to avoid, so the process starts armed and the first poll resolves it once. And `disarm_poll` **saturates at zero instead of `fetch_sub`**, with the rationale stated inline: an underflow would wrap to `u32::MAX` and pin the poll permanently armed — *"a silent, permanent return of the exact regression this module removes."*

Over-arming costs a wasted call; under-arming would be unsound; wrapping would be neither detected nor recoverable. Making the undetectable outcome impossible, rather than merely unlikely, is the right ordering of those risks.

The ordering hazard that would have reintroduced the bug is also handled: `resolve_poll_seed()` runs **before** `gc_moving_loop_polls_enabled()`, so a process run with `PERRY_GC_MOVING_LOOP_POLLS=0` still releases the seed. Reversed, every back-edge would keep paying the out-of-line call for exactly the users who opted out.

**Relaxed ordering is sufficient, and for a stateable reason**: the only requirement is same-thread visibility — each `perry/thread` worker owns an independent arena, `GC_SAFEPOINT_PENDING` is genuinely thread-local, and a thread always observes its own prior writes in program order. Cross-thread staleness costs a wasted call or a needless skip, never a missed drain of the reading thread's own deferral. `volatile` is doing the real work of blocking LLVM hoist/CSE.

And the diagnosis refuted the obvious suspect rather than confirming it: the layout/store symbols are **unchanged** (`init_typed_shape_layout` 23.0% → 21.8%), and what appeared were two symbols absent from the 0.5.1384 profile entirely.

### Three follow-ups, none blocking

1. **`loop_safepoint_purity.rs` doesn't check which branch target the call sits behind.** An operand-swapped `cond_br` inverting the guard's sense would pass it. I verified the polarity by hand (`block.rs:1276` against `loops.rs:198-216`), but nothing execution-level guards it in the required `cargo-test` gate — that currently rests on `gc-moving-witnesses`, a separate job.
2. **No post-fix profile.** The before-fix diagnosis is strong and the deltas are consistent with the ~3 ns × 20 M estimate, but the loop isn't closed by showing `js_gc_loop_safepoint`/`_tlv_get_addr` actually drop out.
3. Mixing a Rust `AtomicU32` with a non-atomic LLVM `load volatile` on the same word is fine on every real target for a naturally-aligned word, but is outside the strict model — worth one line of comment marking it deliberate.

Gates 21/21.

PR #7736, opened 2026-08-09
fix(gc): close #7210's two remaining unrooted-alloca sites + #7640 A/C findings

### LINKED ISSUE #7210: GC: the remaining unrooted-alloca hazards after #7207 — class-keys pointer caches, interleaved staging arrays, inlined-callee param slots

### LINKED ISSUE #7494: native_proof / typed_shape_descriptors: four lowering-independent failures (numeric-coerce on a proven read, missing typed-array fallback record, negative-divisor frem, missing layout-note marker)

proggeramlug:
## Summary

Closes out #7210's two named remaining sites, plus part of #7640's sections
A and C. Full accounting below, including what was **not** reached — several
things turned out not to be hazards, reported as findings rather than
shortfalls.

### #7210 — both named sites, fixed

- `crates/perry-codegen/src/codegen/helpers.rs`'s `emit_namespace_populator`
  (§2's flagship site): `vals_buf` staged every exported binding's NaN-boxed
  value in a plain stack alloca while the per-entry loop called allocating
  helpers (`js_closure_alloc_singleton`, arbitrary cross-module getters). An
  already-staged entry had no root and could go stale before
  `js_create_namespace` read the whole buffer. Fixed by rooting each value in
  a `RootedGroup` as it's produced and deferring every `vals_buf` store to a
  second, call-free pass that runs immediately before the consuming call.
- `crates/perry-codegen/src/lower_call/early_branches.rs`'s
  `obj[strKey](args)` computed-key dispatch (§3's flagship site): receiver,
  key and every argument were lowered into bare registers in sequence, and in
  the static-string-key arm `unbox_str_handle` (an allocating SSO
  materialisation) ran between the args buffer's stores and the consuming
  call. Fixed by rooting `[object, index, ...args]` in one `RootedGroup` and
  building the args buffer last, in each branch, after `unbox_str_handle`.

### #7640 section A — 3 of ~7 arms fixed

- The bounded-index-pair array store — the issue's sharpest repro, next to
  the twin arm #7341 already closed.
- `globalThis[k] = v` — key and the globalThis singleton copy both rooted.
- The width-tracked typed-array non-numeric-index store.

**Not reached**: the `#5525 recv_unknown` inline dyn-TA store, and the TA
runtime-key / TA final-fallback / Uint8Array runtime-key arms.

### #7640 section C — resolved (not a mechanical fix)

Two `property_set.rs` comments claimed a class-field store's receiver
survives an allocating RHS via "the same statepoint re-read" a sibling arm
relies on. That mechanism doesn't exist (RS4GC only relocates values still
`ptr addrspace(1)`-typed and live across the safepoint; the receiver crosses
it as a plain `double`, per `function/precise_roots.rs`). What's actually
going on:

- **Bare `Expr::LocalGet`/`Expr::This` receiver: the claim is TRUE**, via
  `root_reload.rs` (#7280) — a front-end pass, independent of RS4GC, that
  re-materialises a value derived from a shadow-slot or handle-global load
  below any collection point it doesn't dominate. Verified with
  `scripts/gc_root_dominance_check.py --stale-registers`/`--statepoints` on
  both lowerings against a new fixture — zero hazards.
- **Compound receiver — `this.target.x = allocPoint(n).x`, receiver itself a
  class-field READ: the claim is FALSE**, and invisible to both `root_reload`
  (the receiver is a `phi` over two field-get paths, not a direct shadow-slot
  load) and the `--stale-registers` checker (its pattern match only anchors
  on a direct `load double, ptr <root>` source). Confirmed by hand in the
  emitted IR for exactly this shape.

Left unfixed on purpose: rooting the receiver unconditionally would tax the
dominant `LocalGet`/`This` case this investigation just proved needs
nothing, on what #7640 itself calls "the hottest store path in the
compiler" — a measured-cost tradeoff the issue explicitly deferred to a
follow-up that can benchmark it. All four `property_set.rs` sites that made
or relied on the original claim now say this precisely.

New fixture: `test-files/test_gap_gc_class_field_receiver_rooting.ts` —
exercises both halves under `PERRY_GC_MOVING_LOOP_POLLS=1` allocation
pressure, and is also a real Node-comparison functional-correctness check
independent of either checker.

### #7640 section E — triaged, not fixed

All seven named callees
(`ptr_numarray_access::try_lower_num_array_guard_free_{set,get}`,
`try_lower_proven_view_checked_store`, `lower_typed_array_store`,
`lower_buffer_store`, `lower_index_set_fast`,
`masked_window::lower_masked_window_index_get`,
`property_get/generic_dispatch.rs::lower_generic_property_get`) were read,
not skipped. Each looks like a **hazard that is not one**:

1. **Single, already-rooted caller** — `lower_index_set_fast`'s only call
   site is inside the `with_operands_rooted` group #7341 already
   established; `try_lower_num_array_guard_free_{get,set}`'s `object` is
   structurally constrained to `Expr::LocalGet` and its one caller each
   doesn't interleave any other lowering around it.
2. **Address derived after the allocating step, not before** —
   `try_lower_proven_view_checked_store` lowers `value` first, then derives
   the data pointer.
3. **Typed-array immovability**, per CLAUDE.md's own GC section —
   `lower_typed_array_store`/`lower_buffer_store` cache a data pointer across
   an allocating RHS, but into a `GC_FLAG_TENURED`, non-movable backing
   store (same category #7210 section 5 already flagged, not a new finding).
4. `lower_generic_property_get`/`lower_masked_window_index_get` have no
   second allocating operand in their own window at all.

**This triage is static, not checker-verified per callee** — treat it as a
lead for whoever picks up the rest of E, not a closed finding.

## What I verified, and against which commit

The branch is rebased onto `origin/main@f28657da0` (v0.5.1426 — post #7724's
phi-false-positive fix, post #7732's native-lowering-to-0-unrooted,
`--max-unrooted` deleted from the gate). A prior verification pass (both
gated checker modes, `PERRY_RUNTIME_DIR` pinned, full corpus, empty
allowlist) was clean on an earlier commit of this branch, but that was
measured against v0.5.1420 — **before** #7732, so it is not the number this
PR should be judged on. **Re-running now on the current HEAD and will post
the fresh before/after (both lowerings, gated invocations) as a PR comment
as soon as the release rebuild finishes** — opening the PR itself rather
than holding it for that run to land, per the standing guidance that an
unpushed/unopened branch is worth nothing.

## Codegen test baseline

`cargo test -p perry-codegen --no-fail-fast` will be re-run against the same
HEAD and posted as a comment. The last full run (also pre-#7732, pre-#7730)
showed 4 failing integration-test targets matching #7708's pre-existing set;
#7730 ("resolve #7494's four lowering-independent proof-test failures",
merged since) looks likely to have fixed exactly those, which the fresh run
will confirm or correct.

## Test plan

- [x] `cargo check -p perry-codegen`
- [ ] `cargo test -p perry-codegen --no-fail-fast`, diffed against the
      current `origin/main` baseline (see above)
- [ ] `scripts/gc_root_dominance_check.py`, gated invocations
      (`--unrooted-allocas --moving-only`, `--statepoints --moving-only`),
      both lowerings, `PERRY_RUNTIME_DIR` pinned
- [x] New gap fixtures compile and match `node --experimental-strip-types`
      output

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of namespace exports, computed-key method calls, array and typed-array assignments, and global property updates during garbage collection.
  * Prevented incorrect values and stale references when assignments or method calls trigger memory allocation.
  * Improved stability for class-field assignments, setters, and compound receivers under allocation pressure.

* **Tests**
  * Added regression coverage for namespace exports, computed dispatch, indexed stores, typed arrays, global properties, and class fields.
  * Expanded garbage-collection stress testing across these scenarios.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1430
+**Current Version:** 0.5.1431
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1430"
+version = "0.5.1431"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "perry-ffi",
  "scraper",
@@ -5813,7 +5813,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-commander"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5821,7 +5821,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cron"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "chrono",
  "cron",
@@ -5831,7 +5831,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dayjs"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "chrono",
  "perry-ffi",
@@ -5839,7 +5839,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-decimal"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "perry-ffi",
  "rust_decimal",
@@ -5847,7 +5847,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dotenv"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "perry-ffi",
  "serde_json",
@@ -5855,7 +5855,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ethers"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "perry-ffi",
  "rand 0.10.1",
@@ -5863,7 +5863,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-events"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5871,14 +5871,14 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-exponential-backoff"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "perry-ffi",
 ]
 
 [[package]]
 name = "perry-ext-fastify"
-version = "0.5.1430"
+version = "0.5.1431"
 dependencies = [
  "bytes",
  "http-body-util",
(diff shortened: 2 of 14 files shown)


proggeramlug:
## Merging as v0.5.1431

### Verified against the gate that now has no budget

#7732 deleted `--max-unrooted`, so this branch had to hold the count at zero **while growing the corpus from 131 to 134 sources**. It does, on a fresh `--release` build with `PERRY_RUNTIME_DIR` pinned:

```
corpus (native): 134/134 sources compiled, 0 skipped, 156 .ll files
  statepoints: 33174   non-empty live bundles: 19817
--statepoints --moving-only        → within budget: unrooted 0 <= 0   exit 0
--unrooted-allocas --moving-only   → exit 0     (#7210's own detector)
--stale-registers --moving-only    → exit 0
corpus (shadow): 134/134 compiled, 0 skipped
```

This was blocked behind a red `main` that had nothing to do with it — nine false positives from #7729/#7733/#7735, fixed in #7738. Worth knowing because the three new fixtures made the failure look like this PR's when it wasn't.

### The dark-tests gate earned its keep

All three new fixtures were **on disk but unregistered** in `test-parity/gc_repsel_corpus.txt`, so `gc_repsel_matrix.sh` — and therefore `gc-stress`, `gc-moving-witnesses` and `gc-ptr-shape-off-witness` — would never have run them. That is exactly the **#6925 defect** (a gap file left unregistered, gate dark), and it is the second time today a registration check has caught a real hole.

Registered, and then verified they actually pass there rather than trading one red gate for another — because a fixture that fails once registered just moves the red:

| fixture | verdict | arm |
|---|---|---|
| `class_field_receiver_rooting` | PASS | live, counter 16194 |
| `index_set_bounded_globalthis_ta_rooting` | PASS | live, counter 27833 |
| `namespace_and_computed_dispatch_rooting` | PASS | live, counter 11245 |

`live` rather than `UNVER` on all three matters: the matrix distinguishes "output matched" from "output matched **and the arm actually collected**", and an inert arm would have been a passing test that exercised nothing.

### The fixes

- **#7210 §2** — `emit_namespace_populator`'s `vals_buf` staged every exported binding's NaN-boxed value in a plain stack alloca while the per-entry loop called allocating helpers (`js_closure_alloc_singleton`, arbitrary cross-module getters). Fixed by rooting each value in a `RootedGroup` as produced and deferring every `vals_buf` store to a second, **call-free** pass immediately before the consuming call.
- **#7210 §3** — `obj[strKey](args)` computed-key dispatch lowered receiver, key and args into bare registers in sequence, with `unbox_str_handle` (an allocating SSO materialisation) running between the args buffer's stores and the consuming call. Fixed by rooting `[object, index, ...args]` in one `RootedGroup` and building the args buffer last, in each branch.
- **#7640 section A** — 3 of ~7 arms, including the bounded-index-pair array store (the issue's sharpest repro) and `globalThis[k] = v`.

Scope is stated honestly in the body, including several things that turned out **not** to be hazards, reported as findings rather than shortfalls. Sections C and E of #7640 remain open and are named as such — that issue stays open.

`cargo test -p perry-codegen --lib`: 809/809. Gates 21/21.

PR #7739, opened 2026-08-10
fix(gc): the pacing snapshot reports the boundary the predicate uses; the ZealGuard release becomes an assertion (#7729, #7733, #7735 review follow-ups)

proggeramlug:
Review follow-ups from #7729 / #7733 / #7735 that were raised and landed anyway. All three PRs stay as merged; nothing here reverts or redoes them.

### 1. `major_pacing_snapshot` under-reported the escalation boundary

`crates/perry-runtime/src/gc/policy.rs`. The snapshot recomputed the boundary as `baseline × growth` and dropped the floor on the floor of the function:

```rust
let (_floor, growth_num) = major_pacing_config();
```

...while the predicate it mirrors, `arena_growth_full_escalation_due`, **also** rejects every reading below that floor. Wherever the floor dominates, the reported threshold named a boundary the collector does not use:

| state | predicate escalates at | old snapshot reported |
|---|--:|--:|
| no full yet (`baseline == 0`) | 32 MB (the floor) | `0` — "escalates at any size" |
| `baseline = 4 MB`, growth 2 | 32 MB (the floor) | 8 MB |
| `baseline = 64 MB`, growth 2 | 128 MB + 1 | 128 MB |

This matters more than its "minor" label because of *why* the snapshot exists: #7733 added it so the pacing subject could be asserted **live** in the GC trace, rather than a gate merely proving nothing threw. A diagnostic that misreports the quantity it exists to prove is this repo's most expensive recurring failure mode (`PERRY_GC_FORCE_EVACUATE` inert for every `gc()`-driven test, the matrix's `--pressure` knob disabling the path it measured, `moved=` summing two collectors).

The fix is structural rather than a second correct formula: there is now **one** definition of the boundary, `major_pacing_escalation_threshold_bytes`. The predicate is literally `in_use >= it`, and the snapshot reports it verbatim, floor included. `None` means "no arena reading escalates" — either pacing is disabled (`PERRY_GC_MAJOR_PACING_FLOOR_MB=0`) or the growth term overflowed `usize`, which is the same statement about the world, and is why the helper uses `checked_*` rather than `saturating_*` (saturating would report `usize::MAX` and then claim an arena of `usize::MAX` escalates, which the `>` clause never would).

The trace key follows the semantics: `escalate_at_or_above_bytes`, not `escalate_above_bytes` — the predicate's floor clause is a `>=`, and the old name was half of why the figure and the decision could disagree. No consumer of the old key exists in `scripts/`, `.github/` or `docs/`.

**Tests.** `the_reported_escalation_boundary_is_the_one_the_predicate_decides_on` checks the named floor-dominates and growth-dominates cases plus baseline-zero, shift, and pacing-disabled, then goes exhaustive over `floor × growth × baseline × shift`, probing each boundary's own ±1 neighbourhood against an oracle that is a deliberate independent transcription of the four clauses the predicate used to spell out inline (not a call into the code under test). `the_shipped_predicate_and_the_shipped_snapshot_read_one_boundary` then drives the real predicate against the real snapshot on the live arena, so a future re-split fails even if the pure helper stays correct.

Both are sabotage-checked, and the two sabotages are different on purpose:

- Restore the pre-fix formula **in `major_pacing_snapshot` only** (the actual historical bug) → `the_shipped_predicate_and_the_shipped_snapshot_read_one_boundary` fails on `baseline = 0`: `escalate_at_or_above_bytes (Some(0))` against a verdict of `false` on a 0-byte arena. The matrix test passes, correctly — the pure helper was untouched.
- Drop the floor **from the shared helper** → the matrix test fails on the named row: `left: Some(8388609), right: Some(33554432)`.

Restored after each; the final tree is byte-identical to the pushed commits.

### 2. The `ZealGuard` release assertion was missing

`crates/perry-runtime/src/gc/tests/triggers.rs`. `zeal_holds_the_poll_word_armed_with_nothing_pending` asserted `PERRY_GC_POLL_ARMED > 0` inside the guard scope and then only *narrated* the release in a comment. If `ZealGuard`'s `Drop` ever stopped giving the arm back, the process-global word would stay non-zero for the life of the test binary, every later test would silently take the poll's slow path, and this test would still pass.

It now captures the baseline before the guard and asserts `base + 1` inside, `base` after the drop — the same shape `a_deferral_arms_the_poll_word_and_draining_disarms_it` already uses a few lines up.

Sabotage-checked: with `disarm_poll()` commented out of `ZealGuard::drop`, the test fails (`left: 2, right: 1`) on the new post-drop assertion; restored, it passes.

### 3. Wording

- `changelog.d/7729-gc-zeal-allocation-pacing.md` — the `bytes_allocated / stride` bound is qualified to a positive stride, since `PERRY_GC_ZEAL_ALLOC_KB=0` is a supported every-poll mode and deliberately outside it. And "reproduces the pre-fix 1:1 behaviour exactly" against a table showing 283,857 collections for 283,852 polls becomes the truth: one per back-edge poll, plus a handful from the other safepoint zeal forces at — the outermost microtask-pump boundary, which calls `gc_safepoint_moving_minor` without `note_loop_poll_reached`. Near 1:1, not exactly. The same claim in `gc/zeal.rs`'s `ZEAL_DEFAULT_STRIDE_BYTES` doc is corrected identically. (Fragments for merged PRs are folded into the release notes at tag time, so correcting them in place is what actually reaches a reader; `changelog.d/README.md` has no rule against it.)
- `docs/src/internals/memory-model.md` — line 135 already said polls became default-on in #7721; the caveat 20 lines below still said "default off since #7161". It now states the current default, its kill switch, the two gaps that survive it (alloc-free loop bodies by design, the specialized `for`/`for-of`/`for-in` lowerings by omission), and the #7604 exit-70 verdict that means you no longer have to remember to check.

### Validation

- `cargo test --release -p perry-runtime` → `1977 passed; 0 failed`, **cargo's own exit code 0** (captured directly, not through a pipe). All three tests confirmed to have *run*, by name, not merely to have not failed.
- `cargo fmt --all -- --check` clean; `scripts/check_file_size.sh` OK.
- **The trace, end to end.** `gc-handoff/bench/retain.ts` compiled with `PERRY_NO_AUTO_OPTIMIZE=1` against this branch's `libperry_runtime.a` (`PERRY_RUNTIME_DIR` pinned at the freshly built archives), run under `PERRY_GC_TRACE=1`:

  ```
  minor  baseline=0          shift=0  escalate_at_or_above=33554432    (= the 32 MB floor)
  minor  baseline=0          shift=0  escalate_at_or_above=33554432
  full   baseline=66025968   shift=1  escalate_at_or_above=264103873   (= 66025968 x 4 + 1)
  minor  baseline=66025968   shift=1  escalate_at_or_above=264103873
  ```

  The first two rows are the fix: `baseline = 0` is the pre-first-full state, where the old snapshot printed `escalate_above_bytes: 0`. Then the escalated full rebaselines to 66 MB, reclaims little, and backs off to shift 1 — so the reported boundary becomes `4x` the baseline, above the floor, and growth takes over. The block is present, non-vacuous, and each row is arithmetically consistent with the predicate.
- Canary: `gc-handoff/apps/iso_miss.ts` prints `checksum 437840 misses 0`.
- Not run locally: `scripts/gc_instrument_smoke.sh`. Its arm 6 carries a wall-clock budget and this host is heavily loaded, so a local red would not have been informative; CI's runner is the right place for it.

No version bump — maintainer bumps at merge.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected major garbage-collection pacing thresholds and boundary handling.
  - Improved diagnostics for disabled pacing, overflow conditions, and pre-first-collection behavior.
  - Clarified inclusive escalation semantics in GC telemetry.

- **Documentation**
  - Updated GC zeal and allocation-pacing guidance, including default poll behavior and failure diagnostics.
  - Refined benchmark descriptions and release documentation.

- **Tests**
  - Added coverage for pacing boundaries, overflow cases, telemetry consistency, and poll-arm restoration.


proggeramlug:
## Merging as v0.5.1433

### The snapshot bug is the important one, and the framing is right

A diagnostic that misreports the quantity it exists to prove is this repo's most expensive recurring failure — `PERRY_GC_FORCE_EVACUATE` inert for every `gc()`-driven test, the matrix's `--pressure` knob disabling the path it measured, `moved=` summing two collectors. #7733 added `major_pacing_snapshot` specifically so the pacing subject could be asserted **live**, and it named a boundary the collector does not use:

| state | predicate escalates at | old snapshot said |
|---|--:|--:|
| no full yet | 32 MB (floor) | `0` — "any size" |
| baseline 4 MB, growth 2 | 32 MB (floor) | 8 MB |

**The structural fix is what makes this landable rather than a second correct formula.** There is now one definition — `major_pacing_escalation_threshold_for`, pure `(config, state) → boundary` — the predicate is literally `in_use >= it`, and the snapshot reports it verbatim. Two formulas that must agree is the thing that failed; one that both read cannot.

The `checked_*`-over-`saturating_*` reasoning is correct and worth keeping: saturating would report `usize::MAX` and then claim an arena of `usize::MAX` escalates, which the strict `>` clause never would. `None` genuinely means "no reading escalates", and pacing-disabled and overflow are the same statement about the world.

And the doc comment names **why the divergence went untested**: every reachable unit test sat below the floor, where two different formulas agree on the `bool` the predicate returns. That is the general lesson — a predicate that returns a `bool` can hide a wrong quantity indefinitely.

### I verified the sabotage, and got it wrong first

Dropping the floor from the shared helper fails `the_shipped_predicate_and_the_shipped_snapshot_read_one_boundary` with **`left: Some(8388609)`, `right: Some(33554432)`** — byte-for-byte the values you reported.

My first attempt appeared not to reproduce it, and that was my error: I filtered on `major_pacing`, which matches only #7733's two older tests and neither of the new ones. Worth recording because it is the same shape as the bug being fixed — I ran a check whose subject wasn't there and briefly believed the result.

### The `ZealGuard` fix

`zeal_holds_the_poll_word_armed_with_nothing_pending` asserted the word was armed inside the scope and then only **narrated** the release in a comment. If `Drop` ever stopped giving the arm back, the process-global word would stay non-zero for the life of the test binary, every later test would silently take the poll's slow path, and this test would still pass. Now `base + 1` inside, `base` after — matching the shape `a_deferral_arms_the_poll_word_and_draining_disarms_it` already uses.

### On editing merged changelog fragments

Correcting `changelog.d/7729-…` in place is right, for the reason given: fragments are folded into the release notes at tag time, so the correction is what actually reaches a reader. Both corrections are real — the `bytes_allocated / stride` bound genuinely needs the positive-stride qualifier (`=0` is a supported mode deliberately outside it), and 283,857 collections for 283,852 polls is *near* 1:1, not exactly, because the microtask-pump safepoint also forces without `note_loop_poll_reached`. Saying "near" is more useful than a round claim that a reader can falsify from the table two lines up.

The `memory-model.md` caveat still saying "default off since #7161" twenty lines under a line saying #7721 flipped it is exactly the drift that produced #7690.

Gates 21/21.

Issue #7742, opened 2026-08-10
perf(gc): whole-block in-place promotion — kill the ~243 ns/object promotion tax on a fully-live nursery


proggeramlug:
`retain` (3M records, none ever dropped) and `retain_wide` are the only two
benchmarks still far off Node — 0.80 s vs 0.12 and 1.33 s vs 0.15 on the pinned
M1 mini. Everything else is within 1.2×–3.6×.

The cause is that **every survivor gets promoted, object by object**. The GC
trace shows 2,097,155 promotions across four copying minors at 509.7 ms of
pause — **243 ns per object** — spent on `arena_alloc_gc_old`, the `memcpy`,
`layout_transfer`, `old_page_account_promoted_object`, the `GcMoveHookKind`
hooks, and the forwarding/rewrite that follows. The measured young-survival
ratio on those cycles is **1.000**: not one of those objects had a reason to
move.

The structural answer is V8-style whole-page promotion: when a block is
(near-)entirely live, relabel the block as old-gen instead of evacuating it.
Nothing moves, so there is no copy, no old-gen allocation, no layout transfer,
no move hook, no forwarding pointer, and no slot anywhere in the heap or in any
address-keyed side table needs rewriting.

Things this has to get right:

- **The promotion decision.** Block liveness is not knowable before the trace
  and Eden blocks are recycled at offset 0, so per-block history means nothing.
  A per-cycle decision from the previous cycle's measured survival ratio works
  only if a promoting cycle still measures.
- **Page-generation metadata and the remembered set** when a page changes
  generation in place — this is where correctness lives. In particular
  `Old ⟹ TENURED` (#7511) and the old-gen page-object index, which is what the
  remembered-set dirty scan uses to reach an object at all.
- **Pacing.** A copying minor recycles Eden's blocks; promotion hands them away.
  The arena-bytes trigger fires on `arena_total_bytes()`, so the allocation
  runway collapses unless the capacity is given back.
- **Footprint.** Dead objects on a promoted page are retained until the next
  full. The per-cycle cost is `(1 − liveness) × young bytes`; the steady state
  needs its own bound.

proggeramlug:
Measurement of the residue named in #7744's "Not closed", on `main` @ 0a2bf15bd (`perry-dev`, M1 Max), `retain.ts` under `PERRY_GC_TRACE=1`.

**The one full mark-sweep is 36% of all GC pause, and its premise does not hold on this workload.**

[...]

PR #7744, opened 2026-08-10
perf(gc): whole-block in-place promotion of a fully-live young generation (#7742)

proggeramlug:
Closes part of #7742.

When a copying minor's nursery is (near-)entirely live, relabel its blocks as
old-gen instead of evacuating them object by object.

| bench | main | this PR | node |
|---|--:|--:|--:|
| `retain` | 0.81 s / 326 MB | **0.53 s / 316 MB** | 0.12 |
| `retain_wide` | 1.33 s / 447 MB | **1.07 s / 446 MB** | 0.15 |
| `retain1` | 0.39 s / 148 MB | **0.29 s / 133 MB** | — |
| `retain_wide1` | 0.38 s / 174 MB | **0.27 s / 155 MB** | — |
| `deeplist` | 0.30 s / 117 MB | **0.24 s / 97 MB** | 0.09 |
| `churn` / `churn_alloc` / `push_cls` | 0.41 / 0.37 / 0.35 | 0.41 / 0.37 / 0.35 | |
| `push_num` / `churn_read` / `cycles` | 0.13 / 0.02 / 0.19 | 0.14 / 0.02 / 0.19 | |
| `tree` / `tree_wide` | 1.64 / 2.10 | 1.63 / 2.10 | |

Best-of-5 wall clock on the pinned quiet M1 mini (load ~1.6), interleaved,
stdout byte-verified against `node --experimental-strip-types` before timing.
Promotion cost per object **243 ns → 101 ns**, derived from the GC trace's own
promoted-object count and pause.

## Why the old path was pure overhead here

`retain.ts` builds a 3M-element array of records and drops none. Its trace
measures a young-survival ratio of **1.000** on every copying minor, and its
four promoting minors moved 2,097,155 objects in 509.7 ms. Each move paid for a
fresh `arena_alloc_gc_old`, a `memcpy`, a `layout_transfer`,
`old_page_account_promoted_object`, the move hooks, a forwarding stub, and the
rewrite of every referring slot — to put an object somewhere it had no reason
to be.

## Design notes worth reviewing

* **The decision is a measurement.** Block liveness is not knowable before the
  trace and Eden blocks are recycled at offset 0, so it is per cycle from the
  previous cycle's ratio. What makes that sound is that a **promoting cycle
  still traces, so it re-measures** — a workload that flips from live to
  garbage pays one nursery of retained garbage and the policy turns itself off.
  A running 32 MB cap bounds the steady state the per-cycle correction misses.
  The measured population is bimodal by three orders of magnitude (1.000 on
  retain/deeplist, ≤ 0.004 on churn/push/cycles).
* **`Old ⟹ TENURED`** (#7511) is upheld by stamping every header on a promoted
  block, dead ones included, and asserted end to end.
* **Three remembered-set passes are skipped on a promoting cycle**, and it is a
  proof rather than a heuristic: after the retag no address classifies as
  `Nursery` (every in-use young block was taken), so with an empty malloc
  registry there is no child any of them could remember.
  `debug_assert_no_remembering_possible` re-derives it from the heap. Most of
  the win is here.
* **Pacing** is the part that made the first working version a 0.81 → 1.52 s
  regression: promotion hands Eden's blocks away, so the arena-bytes trigger's
  runway collapses. The capacity is returned as one-shot trigger headroom
  rather than as re-reserved blocks (the eager form fixes pacing too but maps
  memory the program may never reach — `retain_wide` peak RSS 470 vs 447 MB).
* **`classify_arena`** resolved the page-generation range twice per visited
  slot; it now resolves once and answers the header's classification from the
  range base, which is also the guard that keeps a garbage candidate at a range
  start from reading the unmapped page below.

## Validation

* `cargo test -p perry-runtime --release`: 1982 passed, 0 failed.
* `gc-handoff/apps/iso_miss.ts` → `checksum 437840 misses 0`, plain and under
  `PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800` and
  `PERRY_GC_VERIFY_EVACUATION=1`. It is an interpreter, so it takes the
  ordinary copying path on all 50 of its minors and still retires 50 quarantine
  sets — the instrument's coverage is unchanged, and it does not cover the new
  path.
* What does cover the new path: `PERRY_GC_FROMSPACE_SCAN_ABORT=1 ./n_retain`
  over 4 promoting cycles (up to 2,111,418 objects / 21,085,616 words per
  cycle) reports `missing_rewrites=0 dangling=0 never_dirty=0 lost_dirty=0
  dirty_but_missed=0` on every cycle — the last three are exactly the "the
  remembered set is missing an edge" counters the skipped passes could break.
* `gc_ratchet` (7 repeats, `shared_ci`): **every gated semantic cell is
  identical to `origin/main` measured on the same host**, `12_large_live_set`
  included. The six cells it reports as REGRESSION are reported identically by
  `origin/main` itself — pre-existing drift against the pinned artifact. The
  corollary is that no ratchet probe runs a promoting cycle, so the ratchet does
  not yet cover this path.
* `check_file_size.sh`, `addr_class_inventory.py`, `cargo fmt --all -- --check`
  all clean.

## Not closed

`retain` is still 4.4× Node and `retain_wide` 7.1×, so #7742 stays open. The
residue is no longer promotion bookkeeping: on a `retain_live_big` profile the
remaining GC time is the remembered-set scan over the array's dirty element
slots (26%) and one full collection that reclaims nothing (28%).

## Summary by CodeRabbit

* **New Features**
  * Added whole-block in-place promotion for high-survival young-generation memory while preserving object addresses.
  * Uses measured survival rates and configurable controls to select promotion.
  * Added promotion metrics, diagnostics, and capacity accounting.

* **Performance**
  * Reduces unnecessary remembered-set processing and object relocation when safe.
  * Retains copying collection for lower-survival workloads.

* **Tests**
  * Added coverage for thresholds, configuration, accounting, metrics, and address preservation.


proggeramlug:
## Merging as v0.5.1434 — I reproduced the soundness evidence independently

The risk here is concentrated in one place: this **skips three remembered-set passes** on a promoting cycle. If the "after the retag no address classifies as `Nursery`" premise is ever wrong, a dropped remembered-set entry becomes a swept-live-object crash a cycle later — the worst failure shape this collector has.

So I built a retain-shaped workload of my own (3 M records, nothing ever dropped) and ran the instrument that speaks to exactly that:

```
PERRY_GC_FROMSPACE_SCAN_ABORT=1 PERRY_GC_DIAG=1 PERRY_GC_TRACE=1 ./n_retain
  → exit 0, sum 4499998500000 len 3000000
  → 5 cycles, each: missing_rewrites=0 dangling=0 never_dirty=0 lost_dirty=0 dirty_but_missed=0
  → promotion lines present — the new path was taken, not bypassed
```

`never_dirty`, `lost_dirty` and `dirty_but_missed` are precisely the "the remembered set is missing an edge" counters, and they are zero across every cycle *with promotion active*. That is the right instrument aimed at the right claim.

`debug_assert_no_remembering_possible` is also the right shape — it re-derives **both** premises from the heap (young bytes zero, malloc registry empty) rather than trusting the argument, and says so.

**A/B on my own host** (loaded dev box, so absolute numbers differ from the pinned mini, but same-host comparison is valid): **1.35 s → 0.57 s**, corroborating the reported 0.81 → 0.53 in direction and magnitude.

### What makes the design defensible rather than a gamble

- **The decision is a measurement, and it self-corrects.** Block liveness isn't knowable before the trace, so it's taken from the previous cycle's ratio — and a promoting cycle *still traces*, so it re-measures. A workload that flips from live to garbage pays one nursery of retained garbage and the policy turns itself off. The measured population being bimodal by three orders of magnitude (1.000 vs ≤0.004) is what makes a per-cycle carry-over safe.
- **`Old ⟹ TENURED`** (#7511) upheld by stamping every header on a promoted block, dead ones included.
- **Pacing was the hard part and is called out as such** — the first working version was a 0.81 → 1.52 s *regression*, because promotion hands Eden's blocks away and the arena-bytes trigger's runway collapses. Returning the capacity as one-shot trigger headroom rather than re-reserved blocks avoids mapping memory the program may never touch (`retain_wide` peak RSS 470 → 447 MB).

### The honesty I want to keep

Three limits stated rather than glossed:

- `iso_miss.ts` takes the ordinary copying path on all 50 minors, so **the quarantine instrument's coverage is unchanged and does not cover the new path** — which is why the from-space scan evidence exists separately.
- **No `gc_ratchet` probe runs a promoting cycle**, so the ratchet does not yet cover this either. Worth a probe.
- **`retain` is still 4.4× Node**, so #7742 stays open, with the residue named: the remembered-set scan over dirty element slots (26%) and one full collection reclaiming nothing (28%).

`cargo test -p perry-runtime --lib`: **1983 passed, 0 failed** (one earlier run showed a single failure at host load 28; clean on re-run). Gates 21/21.

Issue #7245, opened 2026-08-02
native_proof_regressions: 16 tests red on main, and the suite is not per-PR gated


proggeramlug:
`crates/perry-codegen/tests/native_proof_regressions.rs` has **16 failing tests on `main`**, verified against an unmodified tree at `origin/main` during #7242:

- 15 × `invalidation::*`
- `typed_f64_receiver_method_clone_raw_loads_after_composed_guards`

They are unrelated to #7242 (that PR's diff does not touch them; the same 16 fail with and without it).

### Why this matters more than 16 red tests

This is the CLAUDE.md failure mode "**integration suites under `crates/*/tests/*.rs` do not run per-PR** (nightly/tag only) — a regression there can land green and sit red for days" (#5960), observed live.

These tests guard **native-proof invalidation** — precisely the machinery that decides when a type proof stops being valid. That is a soundness-relevant surface: #7242 was merged because an unsound specialization pass produced silently wrong arithmetic, and the suite that would notice a *proof-invalidation* regression is not consulted on any PR.

### Asks

1. Triage the 16 — are they stale expectations, or real invalidation regressions? The distinction matters: the second case means proofs are surviving events that should kill them.
2. Get this suite (or a `cargo-test`-visible subset of it) into the per-PR gate, per #5960's standing recommendation.
3. Whichever way (1) resolves, the suite must not be left red — a permanently-red suite trains everyone to ignore it, which is failure mode (1) by a different route.

Found during #7242. Not a blocker for it.

proggeramlug:
**Confirmed still open on `main` = `d4ac65f21`.** Re-verified in the sweep #7262.

```
cargo test --profile perry-dev -p perry-codegen --test native_proof_regressions
→ test result: FAILED. 235 passed; 16 failed; 0 ignored
```

The failing set is **byte-identical** to the one in this issue — 15 × `invalidation::*` plus `typed_f64_receiver_method_clone_raw_loads_after_composed_guards`:

```
invalidation::body_counter_mutation_does_not_emit_inbounds_buffer_gep
invalidation::body_mutation_of_local_bound_does_not_use_local_length_bound_fact
invalidation::decrementing_loop_update_does_not_emit_inbounds_buffer_gep
invalidation::inclusive_length_loop_does_not_emit_inbounds_buffer_gep
invalidation::inclusive_local_length_bound_does_not_use_local_length_bound_fact
[...]

PR #7745, opened 2026-08-10
fix(events): route the module-level events.* helpers through the dynamic dispatch

proggeramlug:
Closes the one item #7734 left open. **Draft until the node-suite A/B and the full `perry-codegen` run finish** — both are in flight; numbers go in a comment.

## The bug

```ts
import events, { EventEmitter } from "node:events";
const e = new EventEmitter();
e.on("x", () => {}); e.on("x", () => {});

events.listenerCount(e, "x")                      // 2   ✅
const c = events.listenerCount; c(e, "x")         // undefined  ❌ (node: 2)
(events as any).listenerCount(e, "x")             // undefined  ❌
events.listenerCount(...args)                     // undefined  ❌
```

`nm_dispatch_events` had arms for exactly two names — `init` and `EventEmitterAsyncResource` — and everything else fell to `_ => undefined`. The seven module-level helpers live in **perry-stdlib**, which depends on perry-runtime, so the dispatch bucket cannot name them; the static call reaches them directly through codegen's `NativeModSig` rows, which is why only the indirect forms were dead.

This is the third instance of one shape. #7734 fixed it for `querystring` (bridge implemented 1 of 7 advertised names); here the bridge did not exist at all.

## The fix

The registered-pointer bridge every comparable module already has (zlib / querystring / domain / tls):

* perry-runtime: `JS_NATIVE_EVENTS_DISPATCH` + `js_set_native_events_dispatch`, and an arm in `nm_dispatch_events` routing the seven names
* perry-stdlib: `js_events_native_dispatch`, wired at `js_stdlib_init_dispatch`

Argument marshalling matches what the static path's rows produce: event names go through `ToString` (the `NA_STR` coercion), and `setMaxListeners(n, ...targets)` rebuilds its trailing targets into the single array the helper expects (`NA_VARARGS`). Distinct from the existing `JS_NATIVE_EVENTS_CONSTRUCT`, which serves only `new`.

## Measured

A/B with **two different `libperry_{runtime,stdlib}.a` pairs** — this is a runtime change, so swapping only the compiler binary would have compared the same archive against itself and reported a vacuous "no change" (it did, on the first attempt; the archives are now staged separately and verified `cmp`-different):

| | before | after | node 26.5.1 |
|---|---|---|---|
| `listenerCount` captured / dynamic / spread | `undefined` | 2 / 1 / 0 | 2 / 1 / 0 |
| `getEventListeners` dynamic | `TypeError` | 2 | 2 |
| `getMaxListeners` dynamic | `undefined` | 10 | 10 |
| `setMaxListeners` dynamic / spread | no effect | 15 / 21 | 15 / 21 |
| `once` dynamic / spread | `undefined` | `[42]` / `[7]` | `[42]` / `[7]` |
| `addAbortListener` dynamic | `TypeError: is not iterable` | fires | fires |
| unknown helper | `undefined` | `undefined` | `undefined` |

## Tests

`events_dispatch_parity_tests` (perry-codegen, `--lib`, so per-PR visible) walks the **real** `NET_EVENTS_ROWS` table and fails on any `has_receiver: false` `events` row that is not classified as routed-to-stdlib / answered-by-runtime / deliberately-unrouted. That is exactly the drift that caused this bug: a static row landed with no dynamic counterpart and nothing noticed. A second test rejects stale classification entries, so the list cannot rot into a rubber stamp. Both halves sabotage-checked (drop a name → first test fails; add a phantom → second fails).

`node-suite/events/listeners/module-helper-dynamic-dispatch.ts` byte-compares static / captured / type-erased / spread forms against node.

## Deliberately not asserted

`events.on`'s async **iteration**. `for await (const v of events.on(e, "tick"))` drops its first value in the **static** form too, on a tree without this change — `events/on/async-iterator-abort` and `events/on/validation` are already red for it, and I confirmed the static form fails identically on the unpatched archive. Asserting it in this fixture would test that bug rather than this one. `events.on` is routed by the bridge all the same, so it inherits the fix when that gap closes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed indirect and dynamically invoked `node:events` helper calls.
  * Improved support for spread arguments and `setMaxListeners` target arrays.
  * Added handling for listener queries, listener registration, `once`, `on`, and abort-listener helpers.
  * Unsupported helper names continue to return `undefined`.

* **Tests**
  * Added coverage for static, captured, dynamic, and spread event-helper calls.
  * Added parity checks for dispatch routing and asynchronous event delivery.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1434
+**Current Version:** 0.5.1435
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1434"
+version = "0.5.1435"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1434"
+version = "0.5.1435"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1434"
+version = "0.5.1435"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1434"
+version = "0.5.1435"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1434"
+version = "0.5.1435"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1434"
+version = "0.5.1435"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1434"
+version = "0.5.1435"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1434"
+version = "0.5.1435"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1434"
+version = "0.5.1435"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1434"
+version = "0.5.1435"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1434"
+version = "0.5.1435"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1434"
+version = "0.5.1435"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1434"
+version = "0.5.1435"
 dependencies = [
  "serde",
  "serde_json",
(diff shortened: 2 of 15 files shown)


proggeramlug:
## Validation complete — out of draft

**node-suite `events` A/B**, same host, two distinct `libperry_{runtime,stdlib}.a` pairs (`base` = this tree with the stdlib bridge unregistered):

| | base | branch |
|---|---|---|
| pass | 65 | **66** |
| parity fail | 1 | **0** |
| crash | 4 | 4 |

The single base-arm failure is the new fixture, which is the point: it is red without the bridge and green with it. The 4 crashes are byte-identical across arms and pre-existing — `events/async-resource/basic`, `events/emitter/instanceof`, `events/on/async-iterator-abort`, `events/on/validation`. Nothing else moved.

**Unit tests**

* `cargo test -p perry-runtime --lib` — 1976 passed, 0 failed
* `cargo test -p perry-stdlib --lib` — 111 passed, 0 failed
* `cargo test -p perry-codegen --lib events_dispatch` — 2 passed; both sabotage-checked (drop a routed name → the classification test fails; add a phantom name → the stale-entry test fails)
* **full** `cargo test -p perry-codegen --no-fail-fast` — 261/262. The one failure is `typed_f64_receiver_method_clone_raw_loads_after_composed_guards`, already open on main as #7506 (and #7245: "native_proof_regressions: 16 tests red on main, and the suite is not per-PR gated"). Unrelated to this change — it asserts typed-feedback guard ordering in emitted IR, and nothing here touches codegen output. Ran it because #7708 means per-PR CI cannot see that suite.

`cargo fmt --all --check` and `scripts/check_file_size.sh` clean.

### One process note

My first A/B of this change was **vacuous** and I nearly believed it: I swapped the compiler binary while both arms linked the same `libperry_stdlib.a`, so "base" and "fix" printed identical broken output. This is a runtime/stdlib change — the arms have to be different archives. They are now staged separately and verified `cmp`-different before the numbers above were taken.

proggeramlug:
## Merging as v0.5.1435

**The best thing in this PR is that you caught your own vacuous A/B.** This is a runtime change, so swapping only the compiler binary compares the same `libperry_{runtime,stdlib}.a` against itself and reports a serene "no change" — and it did, on the first attempt. Staging the archives separately and verifying them `cmp`-different is the fix, and it is exactly the trap this repo has paid for before: *"`cargo build -p perry-runtime -p perry-stdlib` does not emit the archives, so `perry compile` links a stale `.a` — your fix looks like a no-op and both arms of an A/B behave identically."*

A "no change" result from an A/B that could not have shown a change is worse than no measurement, because it reads as evidence.

### Third instance of one shape, and worth naming as a class

- **#7734** — `querystring`: the bridge existed but implemented **1 of 7** advertised names.
- **here** — `events`: the bridge did not exist at all; `nm_dispatch_events` had arms for `init` and `EventEmitterAsyncResource` and fell to `_ => undefined` for everything else.

The structural cause is the same both times and is stated correctly: the module-level helpers live in **perry-stdlib**, which depends on perry-runtime, so the dispatch bucket cannot name them. The static call reaches them through codegen's `NativeModSig` rows, which is why **only the indirect forms were dead** — captured, dynamic, and spread. A dispatcher that advertises names it cannot serve produces a silent `undefined` rather than a loud failure, which is why these survive.

Given three instances, the general question is worth asking separately: **which other modules advertise names their bridge does not implement?** That is mechanically checkable — enumerate each `nm_dispatch_*`'s advertised set against its stdlib bridge's arms — and would find the fourth before a user does.

### The fix follows the established pattern

`JS_NATIVE_EVENTS_DISPATCH` + `js_set_native_events_dispatch` in perry-runtime, `js_events_native_dispatch` in perry-stdlib wired at `js_stdlib_init_dispatch` — the same registered-pointer bridge zlib, querystring, domain and tls already use, rather than a new mechanism. Kept distinct from `JS_NATIVE_EVENTS_CONSTRUCT`, which serves only `new`.

Argument marshalling matching what the static path's rows produce is the detail that would otherwise bite: event names through `ToString` (`NA_STR`), and `setMaxListeners(n, ...targets)` rebuilding its trailing targets into the single array the helper expects (`NA_VARARGS`). A bridge that dispatches correctly but marshals differently from the static path is a subtler bug than no bridge at all.

Nine behaviours verified against node 26.5.1 across captured / dynamic / spread forms, including `addAbortListener` going from `TypeError: is not iterable` to firing.

Gates 21/21.

Issue #7629, opened 2026-08-08
test_gap_fetch_request_from_node_incoming_message SIGABRTs deterministically on pristine main, and is in no allowlist


proggeramlug:
`test_gap_fetch_request_from_node_incoming_message` dies with **SIGABRT (exit
134)** after printing 5 lines, while node exits 0 with `len=55 match=true`.

## It is not a flake, and it is not in the allowlist

- **3/3 reproducible** on a build of pristine `origin/main` (`38ff7eccc`,
  `--profile perry-dev`, macOS arm64), via
  `./run_parity_tests.sh --filter test_gap_fetch_request_from_node_incoming_message`.
- Absent from `test-parity/known_failures.json`, so nothing is suppressing it —
  it is simply not being *run* by anything that can block a merge.
- Found while A/B-ing #7627 (Layer 1 slice 2). It reproduces identically on both
  arms, so it is unrelated to that change; it is reported here rather than
  passed over, because "fails on clean `main` too" is not the same as "flake".

## Why it has been able to hide

`parity` is tag-gated (v0.5.1018), so a gap regression in this family is only
judged at release time, not per merge. That is the mechanism CLAUDE.md's hazard
list describes and the one #7580 already paid for in `test_gap_diagchannel_*`.
This crash may therefore be considerably older than its discovery date.

## What the test does

`#6432`'s regression cover: `new Request(url, { body: req, duplex: "half" })`
where `req` is a Node `http.IncomingMessage`, read back through
`request.text()`. It starts a real `node:http` server on port 18994, POSTs to
itself with `http.request`, and compares the body length. Node prints:

```
len=55 match=true
expected len=55
```

Perry prints five lines and then aborts. The abort is a signal death rather
than an output mismatch, so the harness classifies it CRASH, not FAIL.

## Suggested next step

Run it under `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1` with a raised
`PERRY_GC_PROTECT_FROMSPACE_DEPTH`, and bisect. A deterministic abort in a path
that mixes a native small-handle body (`IncomingMessage`) with the Web `Request`
reader is the shape CLAUDE.md flags as "a perfectly reproducible GC bug means a
table, not a register" — worth checking the `gc_register_mutable_root_scanner`
registry for the http/fetch side tables before assuming a register.

Either way the first fix is procedural: if it is going to stay red, it needs a
`known_failures.json` entry with provenance so `parity_known_failures.py --audit`
[...]

Issue #7522, opened 2026-08-06
test_gap_zlib_4917_level: js_zlib_deflate_raw_sync link failure does not reproduce under perry-dev — profile/dead-strip dependent, not a host flake


proggeramlug:
## What was reported

`test_gap_zlib_4917_level` fails on `main` as a **link** failure —
`js_zlib_deflate_raw_sync` undefined, even though the symbol is present in the
harness-built archive. Filed originally inside #7518 as a suspected host-local
flake.

## It does not reproduce under `perry-dev` — so it is profile-dependent, not flaky

Same host, Node 26.5.1 oracle, own `--profile perry-dev` build of
`-p perry -p perry-runtime -p perry-stdlib -p perry-runtime-static -p perry-stdlib-static`
at every point, object cache cleared between points, `PERRY_RUNTIME_DIR` pinned at
the freshly built archives:

| commit | date | verdict |
|---|---|---|
| `17c0ff952` | 2026-07-30 | PASS |
| `c6ed8175d` | 2026-07-30 | PASS |
| `33acba4c1` (`main`) | 2026-08-06 | **PASS** |

For contrast, `test_gap_diagchannel_3082_3084_3085_3086` (filed separately) *does*
reproduce under the same `perry-dev` arm at `main`. So the arm is capable of
catching a real regression; this one it does not see.

The failure was observed under the **release**-profile harness
(`run_parity_tests.sh` builds `--release`). `perry-dev` is `opt-level=1`,
`codegen-units=16`, incremental, **no LTO**; `release` is thin LTO,
`codegen-units=1`, `opt-level=3`, strip. An undefined symbol that is present in
the archive is a **dead-strip** symptom, and dead-strip is exactly what changes
between those two.

## Strongest lead

`js_zlib_deflate_raw_sync` is referenced **only** from codegen-emitted `.o` files,
never from within `perry-stdlib` itself. The `#[used] static KEEP_ZLIB_FFI` anchor
at `crates/perry-stdlib/src/zlib.rs:223` exists precisely to survive
whole-program LTO — its comment cites the same failure mode (#5437 /
`project_autoopt_ffi_symbol_link_break`: "the auto-optimize whole-program LTO
therefore dead-strips them from the stdlib archive, breaking the link of any
program that uses zlib"). An anchor that holds without LTO and stops holding with
it presents exactly as reported.

Worth checking first, in order:

1. **#6917 / `ea9c5afa5`** gated keepalive anchors out of the classic link and
[...]

PR #7746, opened 2026-08-10
fix(codegen): a non-numeric key on a Uint8Array/Buffer local is a property read, not a byte (#7700)

### LINKED ISSUE #7700: codegen: a non-numeric key on a Uint8Array/Buffer-typed local reads a byte instead of a property (u8[Symbol.iterator] is a number)
## Summary

A non-numeric key on a `Uint8Array`/`Buffer`-typed local reads a **byte** instead of a property, so every such read is wrong:

| | node 26.5.1 | perry |
|---|---|---|
| `typeof u8[Symbol.iterator]` | `function` | `number` |
| `const k: any = "byteLength"; u8[k]` | `4` | `0` |
| `const k: any = "subarray"; typeof u8[k]` | `function` | `number` |
| `u8.tag = {…}; const k: any = "tag"; u8[k]` | `{"kind":"buffer"}` | `0` |

Reproducer (matches node exactly, diverges on perry at every line):

```ts
const u8 = new Uint8Array([1, 2, 3, 4]);
const it = u8[Symbol.iterator];
console.log(typeof it);                      // node: function   perry: number
const name: any = "subarray";
console.log(typeof u8[name]);                // node: function   perry: number
const lenKey: any = "byteLength";
console.log(u8[lenKey]);                     // node: 4          perry: 0
const anyU8: any = u8;
anyU8.tag = { kind: "buffer" };
const tagKey: any = "tag";
console.log(JSON.stringify(u8[tagKey]));     // node: {"kind":"buffer"}  perry: 0
```

## Where

`lower/expr_member/member_tail.rs` folds **every non-STRING key** on a
`Uint8Array`/`Buffer`-typed local onto `Expr::Uint8ArrayGet` — a symbol key is
not a string, and neither is a `LocalGet` of an `any`-typed local that happens
to hold a string at runtime. Confirmed on `--print-hir`:

```
Let { id: 1, name: "it", ty: Any, init: Some(Uint8ArrayGet { array: LocalGet(0),
      index: SymbolFor(String("@@__perry_wk_iterator")) }) }
Let { id: 5, name: "v",  ty: Any, init: Some(Uint8ArrayGet { array: LocalGet(0),
      index: LocalGet(4) }) }
```

`expr/arrays_finds.rs`'s JS-value lowering of that node handles all three
correctly — a `SymbolFor` key goes to `js_object_get_symbol_property`, an
unproven key to `js_typed_array_index_get_dynamic`. **The value never reaches
it**, because the local is classified numeric and the read is lowered in i32
context (`lower_uint8array_get_i32`) instead.

Six places make the same unconditional "a `Uint8ArrayGet` is a number"
assumption with no regard for the key kind:

- `collectors/integer_locals.rs:736`, `:1048`
- `collectors/i32_locals.rs:120`
- `collectors/int_valued_ta_locals.rs:153`, `:220`, `:268`, `:1153`
- `collectors/not_bigint_locals.rs:125`
- `type_analysis/numeric.rs:134`, `:617`, `:726`

`collectors/pointer_locals.rs` made it too and was fixed in #6998 / #7699 — that
one is the GC-soundness half (a heap value in a local with no shadow slot), and
it is **currently masked by this bug**: no heap value reaches such a local while
every read returns a byte. It becomes load-bearing the moment this is repaired,
which is the order the two should be fixed in.

## Why it was not folded into #6998

The listed collectors decide the `buf[i]` i32 fast path, which is the hottest
buffer code in the compiler (#6996's regression was one temp root per iteration
in exactly these loops). Adding a key-kind condition to them is a measured
change, not a soundness patch. `expr/shadow_slot.rs`'s
`expr_is_known_non_pointer_shadow_value` already states the correct three-part
condition and is the model to copy:

```rust
Expr::Uint8ArrayGet { index, .. } => {
    !matches!(index.as_ref(), Expr::SymbolFor(_))
        && crate::type_analysis::is_numeric_expr(ctx, index)
        && super::index_get::numeric_index_has_integer_array_index_proof(ctx, index)
}
```

## Acceptance

A `test-files/test_gap_*.ts` with the reproducer above, matching node 26.5.1
byte-for-byte, plus a before/after on `churn_alloc` / a `buf[i]` loop showing the
proven-index path still takes the inline byte load.

Found while fixing #6998 (PR #7699).

proggeramlug:
Fixes #7700.

## The bug

A non-numeric key on a `Uint8Array`/`Buffer`-typed local read a **byte** instead of a property — but **only when the result was stored in a local**:

| | node 26.5.1 | perry (main) | perry (this PR) |
|---|---|---|---|
| `const it = u8[Symbol.iterator]; typeof it` | `function` | **`number`** | `function` |
| `const k: any = "subarray"; typeof u8[k]` (stored) | `function` | **`number`** | `function` |
| `const k: any = "byteLength"; u8[k]` (stored) | `4` | **`0`** | `4` |
| `u8.tag = {…}; const k: any = "tag"; u8[k]` (stored) | `{"kind":"buffer"}` | **`0`** | `{"kind":"buffer"}` |
| `u8.n = 1n; const k: any = "n"; typeof u8[k]` | `bigint` | **`number`** | `bigint` |
| `typeof u8[Symbol.iterator]` (consumed directly) | `function` | `function` | `function` |

That last row is the tell, and it is what narrows the diagnosis: codegen's byte-path gate was **already correct**. `arrays_finds::lower_uint8array_get_i32` routes an unproven key to `js_object_get_index_polymorphic` — which dispatches numeric keys to the byte read and everything else to the property path — and hands back a boxed JS value.

The lie is one level up. `lower/expr_member/member_tail.rs` folds every non-STRING key on such a local onto `Expr::Uint8ArrayGet`, and **six codegen collectors read that node as "a byte, hence a number" with no regard for the key kind**. So the destination local was classified integer-valued, took an i32 slot, and `i32_from_indexed_get_lowered` applied `ToInt32(ToNumber(v))` to a function pointer. Consumed directly, there is no local, no i32 slot, and nothing to coerce.

## The fix

The key-kind condition now lives in **one** place — `collectors/byte_read_key.rs` — and it is an **allowlist**: the key must be provably a number. The "not a string" blocklist it replaces is exactly what shipped this bug, by putting every key kind nobody enumerated — symbols first — on the byte-read side.

Applied at all six sites the issue names (`integer_locals` ×2, `i32_locals`, `int_valued_ta_locals` ×3, `not_bigint_locals`) plus the three `type_analysis/numeric.rs` predicates that also mean "this is a raw double": `is_numeric_expr` (drives `fcmp` truthiness and `fadd` operands), `is_provably_not_bigint`, and `integer_magnitude_bits`.

## The hot path is unchanged — measured, not asserted

The issue flags this as "a measured change, not a soundness patch", because these collectors decide the `buf[i]` i32 fast path, the hottest buffer code in the compiler. It is: **gating on declared types alone demotes the loop.**

A `for (let i = …) sum += buf[i]` counter is a body `let`, and the `binding_types` map the collectors are handed covers only params and module globals. Keying the predicate on that map produced a real regression — `js_uint8array_get`/i32-slot → `js_uint8array_index_get_value`/double-slot, visible in the emitted IR. So `collect_numeric_typed_locals` walks the body for declared numeric types (including `for`-init counters), and that set is the evidence.

With it, the emitted LLVM IR for a four-loop buffer fixture (FNV-1a hash, byte sum, masked mix, `Buffer`→`Buffer` copy) is **byte-identical** to a compiler built from pristine `main` — verified by `--trace llvm` diff against a from-scratch baseline build, not against a stale artifact.

## What was tried and rejected

Refusing to fold an unproven key in `member_tail.rs` and letting it fall through to `Expr::IndexGet` is a smaller, root-cause-shaped diff, and it too was IR-identical on the hot fixture. It regressed two shapes the polymorphic escape gets right:

- `buf["1"]` (numeric string key on a `Buffer`) → `undefined`, node says `60`
- an `any`-typed index in an accumulator loop → `NaN`, node says `201`

The fold stays.

## Validation

- New unit tests: `collectors/byte_read_key_tests.rs` (11 cases). Unit, not just gap, on purpose — the gap suite is tag-gated, so a regression there sits red for days (#5960). These run on every PR via `--lib`.
- New acceptance test: `test-files/test_gap_uint8array_nonnumeric_key_7700.ts` — byte-identical to node 26.5.1, and covers the hot loop shapes the fix must not cost.
- `cargo test -p perry-codegen --lib`: 820 passed, 0 failed.
- Full `cargo test -p perry-codegen`: two integration targets fail, **both identically on pristine `main`** in the same tree — `large_object_barriers::large_local_array_push_inbounds_store_emits_precise_slot_barrier`, and `native_proof_regressions::typed_f64_receiver_method_clone_raw_loads_after_composed_guards` (which is open issue #7506).
- `cargo fmt --all --check`, `scripts/check_file_size.sh`, `scripts/addr_class_inventory.py`: clean.
- Gap suite: run against the committed snapshot.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect handling of non-numeric property keys on `Uint8Array` and `Buffer` values.
  * Improved numeric and BigInt classification for indexed and property accesses.
  * Preserved optimized behavior for valid numeric byte reads, including loop calculations and iteration.

* **Tests**
  * Added regression coverage for symbol-, string-, numeric-, and BigInt-keyed accesses.

* **Chores**
  * Updated the application version to `0.5.1436`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1435
+**Current Version:** 0.5.1436
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1435"
+version = "0.5.1436"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "perry-ffi",
  "scraper",
@@ -5813,7 +5813,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-commander"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5821,7 +5821,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cron"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "chrono",
  "cron",
@@ -5831,7 +5831,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dayjs"
-version = "0.5.1435"
+version = "0.5.1436"
 dependencies = [
  "chrono",
  "perry-ffi",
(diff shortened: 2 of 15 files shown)


proggeramlug:
## Merging as v0.5.1436 — I verified both load-bearing claims independently

### The behaviour fix

```
              old (main)        new (this PR)          node 26.5.1
iter:         number            function               function
sub:          number            function               function
blen:         0                 4                      4
tag:          0                 {"kind":"buffer"}      {"kind":"buffer"}
direct:       function          function               function
```

`new` matches node on all five; `main` is wrong on four.

### The hot path — measured, not taken on trust

This is the claim that mattered, because these collectors decide the `buf[i]` i32 fast path. I compiled a four-loop buffer fixture (FNV-1a, byte sum, masked mix) with a pristine `main` release build and with this branch, `--trace llvm` on both:

```
diff -r ir_old/.perry-trace/llvm/ ir_new/.perry-trace/llvm/  →  IDENTICAL
```

### Why the diagnosis is right, and the fix shape better

The `direct:` row is the tell, and it narrows this precisely: `typeof u8[Symbol.iterator]` consumed directly was **always correct**. Codegen's byte-path gate was never wrong — `lower_uint8array_get_i32` routes an unproven key to `js_object_get_index_polymorphic`, which dispatches numeric keys to the byte read and everything else to the property path.

The lie is one level up: `member_tail.rs` folds every non-STRING key onto `Expr::Uint8ArrayGet`, and six collectors read that node as "a byte, hence a number" regardless of key kind. So the destination local took an i32 slot and `ToInt32(ToNumber(v))` was applied to a function pointer. Consumed directly there is no local, no slot, and nothing to coerce.

**Replacing a blocklist with an allowlist is the right correction, not just a fix.** "Not a string" put every key kind nobody enumerated — symbols first — on the byte-read side. "Provably a number" fails safe instead. Verified applied at all six sites (`integer_locals` ×2, `i32_locals`, `int_valued_ta_locals` ×3, `not_bigint_locals`) **and** the three `numeric.rs` predicates that also mean "raw double" (`is_numeric_expr`, `is_provably_not_bigint`, `integer_magnitude_bits`).

### Two things I want to keep from the writeup

**The rejected alternative is documented with its regressions.** Refusing the fold in `member_tail.rs` is the smaller, more root-cause-shaped diff and was *also* IR-identical — but it regressed `buf["1"]` (→ `undefined`, node says `60`) and an `any`-typed index in an accumulator loop (→ `NaN`, node says `201`). A smaller diff that loses two behaviours the polymorphic escape gets right is the worse trade, and knowing that saves the next person from re-deriving it.

**The `binding_types` finding is the real engineering content.** Gating on declared types alone *demoted the loop*: a `for (let i = …) sum += buf[i]` counter is a body `let`, and that map covers only params and module globals. Keying the predicate on it produced a measurable regression (`js_uint8array_get`/i32-slot → `js_uint8array_index_get_value`/double-slot, visible in the IR). Hence `collect_numeric_typed_locals` walking the body. That is a trap anyone touching these collectors will hit.

Unit tests rather than gap-only is also the right call, and for the stated reason: the gap suite is tag-gated, so a regression there sits red for days (#5960).

`cargo test -p perry-codegen --lib`: 820. `perry-runtime --lib`: **1985 passed, 0 failed** (one gate run showed a failure at host load ~20; clean on re-run, and this PR touches codegen only). Gates 21/21.

proggeramlug:
## Gap suite: 515 tests, no regression from this change

Ran the full gap suite (`run_parity_tests.sh --filter test_gap_`) against this branch: **488 parity pass, 21 parity fail, 6 crash, 0 compile fail** — and `test_gap_uint8array_nonnumeric_key_7700` **passes**.

Comparing to `test-parity/gap_snapshot.json` gives 12 tests that diverge from the committed snapshot. **None of them is caused by this change** — every one behaves identically on a pristine `main` build made in the same worktree, with the same binary path, on the same host:

| test | this branch | pristine `main` |
|---|---|---|
| `events_import_4995` | parity_fail | parity_fail |
| `gc_rest_argument_rooting` | parity_fail | parity_fail |
| `gc_same_module_call_argument_rooting` | parity_fail | parity_fail |
| `specabi_reassign` | parity_fail | parity_fail |
| `zlib_3285_params` | parity_fail | parity_fail |
| `zlib_4917_level` | parity_fail | parity_fail |
| `fetch_request_from_node_incoming_message` | crash | crash |
| `http_client_no_redirect_follow` | crash | crash |
| `http_overloads_3226plus` | crash | crash |
| `http_req_async_iterator` | crash | crash |
| `http_res_socket_writable_onfinished` | crash | crash |
| `net_connect_bound_value` | crash | crash |

`specabi_reassign` is the one I most wanted an A/B for — spec-ABI + reassignment is exactly where a change to integer/i32 local classification could plausibly land. It fails identically on both arms.

Several of these are already known and tracked: `fetch_request_from_node_incoming_message` is #7629 ("SIGABRTs deterministically on pristine main, and is in no allowlist"), the two `zlib` ones are #7522/#7523 (profile-dependent, and this run is `perry-dev`), and `events_import_4995` has an open fix in #7745.

A further 10 tests move `node_fail → parity_fail` versus the snapshot (all npm-package tests: `moment_methods`, `dayjs_factory_arg`, `slugify_options`, …). That is the snapshot recording a host where the oracle could not run them; here node runs them and diverges. Host artifact, both arms, unrelated.

**Caveat stated plainly:** this run used the `perry-dev` profile, not `--release`, so it is not a substitute for the tag-gated `parity` job — it is an A/B for *this* change, which is what it is being used for. Both arms used the identical profile, binary path, and host.

Issue #797, opened 2026-05-15
test-parity: audit known_failures.json — every entry needs an issue # and date


TheHypnoo:
Walk `test-parity/known_failures.json`. Each entry must reference an open issue and the date it was skip-listed. Close stale entries; file issues for orphans.

Part of #793.

proggeramlug:
Concrete instance of the provenance problem this issue tracks, found during #7265's CI triage.

**`zlib_4917_level` is documented in `test-parity/known_failures.json` (#6847)** — "ext-zlib provider archive with a feature-stripped stdlib rebuild… on a COLD object cache (warm /tmp caches mask it)… **Not a parity regression**" — **but it is not in `gap_snapshot.json`, which is the file the conformance-smoke shard gate actually reads.**

So the shard reported it as `pass -> compile_fail`, flagged "REGRESSION — expected to pass", and cost an agent real time chasing a failure that was already understood and written down. Warm caches on other PRs mask it, which is why it appears intermittent and PR-specific.

**Two files describe the same thing and only one is consulted by the gate.** That is the failure mode: a known failure recorded in the file humans read, absent from the file the machine reads, so the documentation cannot prevent the rediscovery it exists to prevent.

Worth considering as part of this issue's scope:
1. Make one file authoritative, or generate the gate's input from the documented one so they cannot drift.
2. If they must stay separate, add a check that every `known_failures.json` entry has a corresponding `gap_snapshot.json` entry (and flag orphans in both directions).

Related observation from the same run: several `node_fail -> parity_fail` entries are the oracle *reclassifying* tests it could not run when the snapshot was taken — CLAUDE.md's silent-drop hazard firing in reverse. One such entry differed between two PRs' runs on the same base, which is the signature of an environment-dependent oracle rather than a code change.

PR #7748, opened 2026-08-10
ci: perry-codegen integration suites can no longer be silently omitted from per-PR CI (#7708)

### LINKED ISSUE #7708: CI cannot see crates/*/tests/*.rs regressions: per-PR runs --lib --bins, and SOURCE_SUITE_MAP is hand-maintained (2 failures beyond #7494)

proggeramlug:
Fixes #7708. Supersedes #7507 (which asked for six of these suites and was blocked on them being green).

## 1. Triage of the six

Re-ran `cargo test -p perry-codegen` suite-by-suite on `main` (v0.5.1432). **Four of the six are fixed**, by unrelated work, with nothing recording it:

| test | status |
|---|---|
| `integer_arithmetic_array_push_omits_inbounds_layout_note_and_barrier` | fixed |
| `integer_modulo::i32_counter_mod_unsafe_or_nonliteral_divisors_keep_frem` | fixed |
| `proven_buffer_and_typed_array_reads_are_numeric_operands` | fixed |
| `reassigned_typed_array_store_records_runtime_fallback` | fixed |
| `large_object_barriers::large_local_array_push_inbounds_store_emits_precise_slot_barrier` | **still red** |
| `native_proof_regressions::typed_f64_receiver_method_clone_raw_loads_after_composed_guards` | **still red** — #7506 |

Four entries silently stopped being true, which is exactly the bookkeeping failure #797 recorded for the parity skip-list. That shaped the design below more than the count did.

## 2. The map was three suites; the crate has twenty-four

`SOURCE_SUITE_MAP` mapped `crates/perry-codegen/src/` to `shadow_slot_hygiene`, `scalar_replaced_slot_roots`, `temp_root_operand_temporaries`. **23 of the 24 suites in that crate are the same kind of thing** — in-process compiles of hand-built HIR with `emit_ir_only: true`, no `perry compile` subprocess, no link, no runtime. The narrowness was never a judgment about the other 20; it was what one commit had time to verify. And nothing could tell you: a suite in neither the map nor any rule is indistinguishable from a suite nobody needed to map.

**Measured, one `cargo test -p perry-codegen --test <suite>` per row, arm64 macOS / debug:**

```
app_window_config_options          10.34s  ok  2      native_proof_buffer_views      4.03s  ok 36
argless_builtin_extra_args          2.36s  ok  2      node_test_mock_property_...    2.88s  ok  1
class_field_store_pointer_test      2.25s  ok  4      perry_builtin_name_collision   2.62s  ok  5
class_keys_gc_root                  2.21s  ok  1      private_guard_declaring_class  2.48s  ok  1
constructor_recursion               2.20s  ok  1      scalar_replaced_slot_roots     2.93s  ok 11
destructure_call_location           2.23s  ok  2      shadow_slot_hygiene            2.47s  ok 12
i64_spec_ternary_recursion          2.16s  ok  2      static_symbol_hygiene          2.53s  ok  2
loop_safepoint_purity               3.87s  ok  8      temp_root_operand_temporaries  2.68s  ok 19
macos_bundle_chdir_gate             2.48s  ok  2      typed_feedback                 2.45s  ok 17
manifest_consistency                3.77s  ok  5      typed_shape_declared_at_alloc  2.71s  ok  9
                                                      typed_shape_descriptor         2.26s  ok  2
large_object_barriers              13.87s  FAIL 1     typed_shape_descriptors        2.44s  ok 18
native_proof_regressions           10.46s  FAIL 1
```

Nearly all of each number is cargo's freshness check; the test time is sub-second (`shadow_slot_hygiene` 12 tests / 0.10 s, `native_proof_buffer_views` 36 / 0.2 s). The cost is the build, which `e2e-scoped` already pays whenever any perry-codegen suite is in scope.

**The map is now complete-by-construction: 22 suites, up from 3.**

## 3. Omission can no longer be silent

`_assert_map_covers_codegen_suites` — run by `--self-test`, which `e2e-scoped` executes *before* it installs a toolchain — requires every `crates/perry-codegen/tests/*.rs` on disk to be either mapped or in the new `SUITE_EXCLUSIONS`. A new suite fails CI until someone classifies it. The map stays hand-written (there is still no coverage data), but an omission is no longer indistinguishable from a decision.

The sabotage case is in the self-test: it plants an unclassified suite and requires the raise. A rule that has never failed is a rule nobody has verified — CLAUDE.md hazard 4.

## 4. Exclusions are per-test and die when fixed

Holding out a whole 262-test suite for one failure is how 261 tests' worth of coverage went dark. So the two known failures are excluded **by test name**; the suites still run with `--skip`.

Self-invalidating in both directions:

- `--self-test` rejects an entry whose suite is gone, or which is also mapped.
- the job runs *exactly* the held-out tests and **fails if one passes** (or if no test matches that name), telling the fixer to delete the entry. A fix cannot land leaving its exclusion behind.

## 5. The cap would have re-created the bug one layer up

`--cap` (default 12) exists so a mass suite rename can't turn one PR into a multi-hour `perry compile` run. Applied to 22 mapped suites it would have silently truncated the set the map just completed. So the cap now applies to **diff-named suites only**, and the two tiers carry different per-suite bounds (`MAPPED_SUITE_TIMEOUT_S=300` vs `NAMED_SUITE_TIMEOUT_S=1500`). The job's `timeout-minutes` math is updated in the same commit and the comment states the new arithmetic.

## Validation

- `python3 scripts/ci_e2e_scope.py --self-test` — passes, including the two new sabotage cases (unclassified suite must raise; cap must not touch mapped suites).
- **The workflow bash was executed, not eyeballed.** Extracted the run step and drove it with a stubbed `cargo` across five scenarios: (1) all-as-recorded → exit 0, 22 suites, 2 known-failure checks, 0 errors; (2) an excluded test now passes → exit 1 with the delete-the-entry error; (3) a mapped suite regresses → exit 1; (4) a diff that names an excluded suite directly still gets `--skip`; (5) a non-codegen diff runs no known-failure checks.
- **The gate was run green before promotion, with real `cargo` against a built tree**: 22 suites all green, both known-failure checks still failing as recorded, exit 0.

Note that this PR touches no `crates/perry-codegen/src/` file, so `e2e-scoped` on *this* PR will select nothing — hence the local end-to-end run above. The first PR that touches codegen source is the first live exercise.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved end-to-end test selection for code-generation changes, covering all mapped integration suites.
  - Replaced broad suite skips with targeted handling for known failing tests.
  - CI now detects stale exclusions or tests that unexpectedly pass, preventing inaccurate results.

- **Tests**
  - Expanded per-PR coverage to 22 integration suites.
  - Added per-suite time limits and increased the overall workflow timeout to support uncapped suites.
  - Added validation to ensure suite mappings and exclusions remain complete and accurate.


proggeramlug:
## Merging as v0.5.1438 — this closes #7708 properly rather than adding three more suites

### The triage cross-checks exactly

"Four of the six are fixed, by unrelated work, with nothing recording it" matches what I measured independently today: #7730 took `cargo test -p perry-codegen` from **6 pre-existing failures to 2**, and the two it left are precisely the two you list as still red (`large_object_barriers::…precise_slot_barrier` and `native_proof_regressions::…composed_guards`, the latter #7506). Two people arriving at the same set from different directions is the strongest form that claim could take.

And the observation that matters more than the count — **four entries silently stopped being true** — is the same bookkeeping failure #797 recorded for the parity skip-list. Letting that shape the design was the right call.

### 3 → 22, complete-by-construction

The old map wasn't a judgment about the other 20; it was what one commit had time to verify, and *nothing could tell you that*. A suite in neither the map nor any rule was indistinguishable from a suite nobody needed to map. `_assert_map_covers_codegen_suites` removes that ambiguity — and I verified it bites by planting an unclassified suite:

> `ci_e2e_scope: these crates/perry-codegen/tests/*.rs suites are in neither SOURCE_SUITE_MAP nor SUITE_EXCLUSIONS: zz_audit_planted_suite. … An unclassified suite is invisible to per-PR CI, which is #7708.`

Naming the issue in the failure text is what makes the next person's five minutes cheap.

### Two design decisions I'd have got wrong

**Per-test exclusions, not per-suite.** Holding out a whole 262-test suite for one failure is exactly how 261 tests' worth of coverage went dark in the first place. And they're self-invalidating in *both* directions — the job runs the held-out tests and **fails if one passes**, so a fix cannot land leaving its exclusion behind. That is the property the parity skip-list lacked.

**The `--cap` catch is the sharpest thing here.** A cap of 12 applied to 22 mapped suites would have silently truncated the set the map had just completed — re-creating #7708 one layer up, inside its own fix. Scoping the cap to diff-named suites only, with separate per-tier timeouts, is right, and the `timeout-minutes` arithmetic moving in the same commit is the detail that keeps it honest.

### On the validation

Executing the workflow bash with a stubbed `cargo` across five scenarios — including "an excluded test now passes" and "a mapped suite regresses" — is the right way to test a CI change, because the usual failure mode is that the YAML is never run until it matters. Running the gate green with real `cargo` before promotion satisfies CLAUDE.md's corollary that a new gate must be observed green once first.

The honest note that **this PR touches no `crates/perry-codegen/src/` file, so `e2e-scoped` on this PR selects nothing** is worth keeping visible: the first codegen-touching PR is the first live exercise. Worth watching that one.

Gates 19/19, `--self-test` clean.

PR #7751, opened 2026-08-10
fix(deforest): the caller's binding kept a growth-forwarding stub after a deforested call (#7661)

### LINKED ISSUE #7661: array growth forwarding: build()-and-return leaves a stale head in the binding (minimal N = 17)
Split out of #7660, which fixed the *consumer* half. This is the producer half,
and it is a latent hazard for every future consumer of a raw array head.

## The observation

```ts
function build(n: number): Node[] {
  const out: Node[] = [];
  for (let i = 0; i < n; i++) out.push(new Node(i, i * 2));
  return out;
}
const keep = build(1000);   // `keep` holds a growth-forwarding STUB, not the live head
```

`keep` is demonstrably a stub. Adding a module-scope `keep.push(x); keep.pop();`
— which forces a write-back of the *resolved* head into the global — changes
the behaviour of a compiler in which nothing else differs; and #7660's fix,
whose only effect is `js_array_refresh_local_head` + write-back, would be a
literal no-op if the binding were already live (that helper returns its input
untouched when there is nothing to follow). It is not a no-op: it turns a
SIGBUS into the right answer.

## Why that is surprising

`expr/array_push.rs` already does the write-back. Both reallocating arms of the
inline push emit it:

```llvm
apush.realloc.15:
  %r106 = call i64 @js_array_push_f64(i64 %r59, double %r55)
  ...
  store ptr addrspace(1) %rs4gc.s4, ptr %r2        ; <- out's slot, updated
```

So `out`'s own slot is kept live inside `build`. The stub is being reintroduced
somewhere between that slot and the caller's binding — candidates, unverified:
the `alwaysinline` + `$spec_i32_b` specialization pair (the return value may be
derived from the head cached in `%r59` rather than a fresh load of `%r2`), or
the return/assignment lowering. Worth an hour with `--trace llvm` on the
reproducer above.

## Why it matters beyond #7660

Growth forwarding is *transparent through the runtime* — every entry point
calls `clean_arr_ptr` — so a stale binding is invisible until emitted code
dereferences the head itself. Perry now has several such consumers (4a's inline
element tier, the masked-window reads, #7612's element-shape clone), and each
one has had to remember `js_array_refresh_local_head` independently. #7612
forgot, and the result was a bus error that shipped. Fixing the producer would
turn "remember the refresh" from a correctness requirement into an
optimization.

## Suggested acceptance

A test that asserts the *binding*, not just the observable behaviour: compile
the reproducer and assert the value reaching the caller resolves to itself
under `clean_arr_ptr` (i.e. carries no `GC_FLAG_FORWARDED`). Behaviour alone
cannot see this — every runtime path resolves the chain and prints the right
answer either way, which is exactly why it went unnoticed.

proggeramlug:
Fixes #7661.

## Where the stub comes from

The issue's two candidates were the `alwaysinline` + `$spec_i32_b` specialization pair and the return/assignment lowering. It is neither — it is the **deforestation pass**, and the mechanism is more direct than "a stale head leaks through".

`crates/perry-transform/src/deforest/` rewrites an array-producing function to fill a caller-allocated accumulator. Its own module doc, point 7:

> **Non-consumer call sites** (e.g. top-level `const all = f(...)`): rewrite to `const all = []; f(args, all);` so callers that need the array as a value still get it.

So the reproducer becomes:

```ts
let keep = [];        // caller allocates, capacity 0
build(1000, keep);    // callee grows it 0 -> 1000, relocating repeatedly
```

`js_array_grow` does not grow in place. The callee's `out.push` write-back re-points its **own** out-param slot; nothing re-points `keep`. And the producer rewrite *dropped* the trailing `return out` (step 2), so there was no value to store back either. Both halves are visible in `--trace llvm` on the issue's exact reproducer:

```llvm
; producer — falls through to `return undefined`
for.exit.4:
  ret double 0x7FFC000000000001

; caller — result discarded; `keep`'s slot (%r1) still holds the pre-growth head
%r2 = call i64 @js_array_alloc(i32 0)
store ptr addrspace(1) %rs4gc.s2, ptr %r1
%r8  = load ptr addrspace(1), ptr %r1
%r9  = call double @perry_fn__7661_ts__build$spec_i32_b(i32 1000, double %r8)
%r10 = load ptr addrspace(1), ptr %r1          ; <- reads the stub
```

That also explains the issue's observation that a module-scope `keep.push(x); keep.pop();` changes behaviour: it is the only thing that writes a resolved head back into the global.

## The fix

No new HIR node, no codegen change — the producer already holds the correct head in its out-param slot at the point it used to fall off the end.

1. **`producer_rewrite.rs`** — KEEP the trailing `return out`. Step 4 already substitutes `out_local_id → out_param`, so it becomes `return <out_param>`: the live head, after every realloc write-back. `detect.rs` guarantees exactly one top-level `Stmt::Return(Some(Expr::LocalGet(out_id)))` and that it is the last top-level statement, so there is nothing else this could be keeping.
2. **`call_sites.rs`** — all three rewrites store the result back over the caller's binding:
   - non-consumer: `keep = f(n, keep)`
   - consumer-fuse: `outer = f(args, outer)`
   - pass-through recursion inside the producer (same code path, so covered by the same change)
3. The seeded binding is emitted `mutable`, because it is now written twice. Leaving it `const` while storing through it would make every analysis that trusts `mutable: false` wrong.

After:

```llvm
; producer
for.exit.4:
  %r109 = load ptr addrspace(1), ptr %r2
  ret double %r109

; caller
%r9 = call double @perry_fn__7661_ts__build$spec_i32_b(i32 1000, double %r8)
store ptr addrspace(1) %rs4gc.s3, ptr %r1      ; <- keep is re-pointed
```

Verified in IR that the **recursive** shape gets it too — both `tree(...)` self-calls store their result back into the out-param slot before the next use.

This is what the issue asked for: it turns `js_array_refresh_local_head` from a correctness obligation every future consumer of a raw array head must remember into an optimization.

## Coverage is structural, and it was sabotage-checked

Behaviour cannot see this bug. Every runtime entry point resolves the forwarding chain through `clean_arr_ptr`, so the program prints the right answer either way — which is exactly why it went unnoticed until #7612 dereferenced a head directly and `SIGBUS`ed at N = 17.

So the load-bearing tests assert the transform's output shape, in `deforest/tests.rs` (`--lib`, so they run on every PR):

- `producer_returns_the_out_param_not_undefined`
- `plain_call_site_stores_the_returned_head_back_over_the_binding`

**Both were verified to FAIL against the pre-fix transform** — I reverted the two source changes, re-ran, and confirmed 2 failed / 11 passed, then restored. A structural test that has never failed is a test with no subject; these have one.

One existing test needed updating rather than fixing: `deforests_producer_called_from_class_method` walks for the surviving call under `Stmt::Expr`, which is now one level deeper inside the `LocalSet`. Its walk now unwraps, and the comment records why that matters (a non-unwrapping walk would have collected nothing and compared `[] == [1]` — a failure, but only because the expectation is non-empty).

`test-files/test_deforest_growth_forwarding.ts` exercises all three call-site shapes end-to-end including N = 17, and its header states plainly that it is a smoke test, **not** a detector.

## Validation

- `cargo test -p perry-transform --lib`: 58 passed, 0 failed (13 deforest, incl. the 2 new).
- End-to-end vs node 26.5.1 on all three shapes (`build(1000)`, a depth-9 `tree` with consumer-fuse + recursion, `build(17)`): byte-identical.
- Full `test-files/` parity sweep: run and compared against `test-parity/gap_snapshot.json`.

Cost: one store per deforested call. No allocation, no call, no change to the fusion itself.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed array growth handling so updated array references are correctly preserved across optimized operations.
  * Improved reliability for recursive processing, direct construction, and boundary-size arrays.
  * Ensured generated bindings remain writable when values need to be updated after growth.

* **Tests**
  * Added regression and end-to-end coverage for array growth forwarding, including nested and recursive scenarios.

* **Documentation**
  * Added a changelog entry describing the array growth fix.

* **Chores**
  * Updated the project version to 0.5.1439.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1438
+**Current Version:** 0.5.1439
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1438"
+version = "0.5.1439"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1438"
+version = "0.5.1439"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1438"
+version = "0.5.1439"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1438"
+version = "0.5.1439"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1438"
+version = "0.5.1439"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1438"
+version = "0.5.1439"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1438"
+version = "0.5.1439"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1438"
+version = "0.5.1439"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1438"
+version = "0.5.1439"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1438"
+version = "0.5.1439"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1438"
+version = "0.5.1439"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1438"
+version = "0.5.1439"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1438"
+version = "0.5.1439"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1438"
+version = "0.5.1439"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1438"
+version = "0.5.1439"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1438"
+version = "0.5.1439"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1438"
+version = "0.5.1439"
 dependencies = [
  "argon2",
  "perry-ffi",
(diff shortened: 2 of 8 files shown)


proggeramlug:
## Merging as v0.5.1439

**The diagnosis rejects both of the issue's own candidates and is better for it.** Not the `alwaysinline`/`$spec_i32_b` pair, not the return/assignment lowering — the deforestation pass dropped the trailing `return out`, so the producer fell through to an implicit `return undefined` and the caller kept the head it had allocated *before* the call. `js_array_grow` doesn't grow in place; the callee's write-back re-points its **own** out-param slot and nothing re-points `keep`.

The before/after IR makes it unarguable — `ret double 0x7FFC000000000001` becoming `ret double %r109`, and the caller gaining the `store` that re-points its binding. And it explains the issue's own puzzling observation, that a module-scope `keep.push(x); keep.pop();` changed behaviour: that was the only thing writing a resolved head back into the global.

### The line that decides how this had to be tested

> **Behaviour cannot see this bug.** Every runtime entry point resolves the forwarding chain through `clean_arr_ptr`, so the program prints the right answer either way.

That is why it survived until #7612 dereferenced a head directly and `SIGBUS`ed at N = 17, and why an end-to-end test would have been theatre. Asserting the **transform's output shape** is the only thing that can fail here, and putting those in `--lib` rather than the tag-gated gap suite means they run on every PR.

I verified they have a subject rather than taking the claim: re-introducing the drop (`func.body.retain(|s| !matches!(s, Stmt::Return(_)))`) fails `producer_returns_the_out_param_not_undefined` — and fails *exactly one* test, because I reverted only the producer half and the call-site test covers the other. That matches your "2 failed / 11 passed" when both are reverted.

### Two details that would have been easy to get wrong

**The `mutable: true` change.** The seeded binding is now written twice, and leaving it `const` while storing through it would make every analysis that trusts `mutable: false` wrong. That is a much worse bug than the one being fixed, and quieter.

**The updated existing test, and the reason recorded.** `deforests_producer_called_from_class_method` walks for the call under `Stmt::Expr`, now one level deeper inside the `LocalSet`. The note that a non-unwrapping walk would have collected nothing and compared `[] == [1]` — a failure, but only because the expectation happens to be non-empty — is the kind of thing that saves the next person from "fixing" it by making the expectation empty.

Marking `test_deforest_growth_forwarding.ts` in its own header as a smoke test and **not** a detector is the right label, given the first paragraph above.

The framing of the outcome is also right: this turns `js_array_refresh_local_head` from a correctness obligation every future consumer of a raw array head must remember into an optimization. Fixing the producer once beats fixing consumers one at a time, which is what #7660 had to do.

Cost is one store per deforested call. `cargo test -p perry-transform --lib`: 58 passed, 0 failed. Gates 21/21.

Issue #7737, opened 2026-08-09
GC follow-ups from #7733: one-way prototype latch (#7510 shape), missing positive-direction pacing test, unmeasured grow-then-churn transition, advisory-only ratchet


proggeramlug:
Follow-ups from the audit of #7733 (merged as v0.5.1427). None blocked the merge; all four are cheap and concrete.

## 1. `OBJECT_PROTOTYPES_NONEMPTY` is a one-way latch — the #7510 shape, with a third victim

`object_static_prototype_owner_moved`'s new fast path is gated on `OBJECT_PROTOTYPES_NONEMPTY.load(Acquire)` (`object/prototype_chain.rs:~340`). Nothing ever resets that latch to `false` — **including `prune_dead_object_prototype_owners` when its `retain()` empties the map back out** (`prototype_chain.rs:328-335`).

So a single `Object.setPrototypeOf` against a non-meta-capable owner **anywhere in a process's lifetime** — even one that later dies and gets pruned — permanently disables the fast path for the rest of that process. A long-running server that does one incidental re-prototyping early (of a `RegExp`, say) silently forfeits the 2.5 M-lock / 2.5 M-hash win, with no signal.

This is pre-existing and shared by all three readers (`object_static_prototype:255`, `prune_dead_object_prototype_owners:329`, and now the move hook), not introduced by #7733. But it is exactly **#7510's finding — "one immortal side-table entry nullified every `is_empty()` fast path"** — recurring. Fix: reset the latch when prune leaves the map empty, or replace it with a real emptiness check.

## 2. No positive-direction test for the recording

`declining_to_escalate_records_no_pre_full_reading` proves only the negative: a declined escalation leaves `pre_in_use == 0`. Nothing asserts that a **`true`** verdict from the real `arena_growth_full_escalation_due()` — not the `test_note_full_cycle_reclaimed` bypass — leaves a non-zero reading behind.

That positive direction is precisely the shape of the first-cut bug #7733's own changelog describes: recording wired at the wrong call sites, `update_major_pacing_backoff` silently no-op'ing, every test green. The gap exists because forcing `due=true` deterministically needs control over `arena_in_use_bytes()`, which isn't mockable today — real, but addressable.

## 3. The grow-then-churn transition is named but unmeasured

[...]

PR #7752, opened 2026-08-10
gc: assert the positive direction of arena-growth pacing recording, on one shared accessor (#7737 item 2)

proggeramlug:
Addresses **item 2** of #7737 ("No positive-direction test for the recording"). Item 1 landed in #7740; item 4 is a maintainer action; item 3 needs the pinned bench host and is left open — so this does **not** close the issue.

## The gap

`declining_to_escalate_records_no_pre_full_reading` proves only the negative: a declined escalation leaves `pre_in_use == 0`. Nothing asserted that a `true` verdict from the **real** `arena_growth_full_escalation_due()` — not the `test_note_full_cycle_reclaimed` bypass, which sets the reading itself — leaves a non-zero one behind.

A negative-only test cannot tell "declined correctly" from "the recording never ran". Both produce `0`. That is exactly the first-cut bug #7733's own changelog describes: the recording wired at the wrong call sites, `update_major_pacing_backoff` returning early on a zero pre-reading, and every test still green.

## Why it wasn't already there, and what I did instead

The issue says forcing `due=true` "needs control over `arena_in_use_bytes()`, which isn't mockable today". Confirmed, and the reason is worth recording: the boundary is `max(floor, baseline × growth + 1)`, and the floor **always dominates from below**, so no amount of baseline manipulation gets the threshold under it. `PERRY_GC_MAJOR_PACING_FLOOR_MB` can't help either — `major_pacing_config` is a process-wide `OnceLock`, so an env var only takes effect if this test happens to run first.

That leaves a 32 MB live heap, which is precisely what `major_pacing_escalation_threshold_for` was factored out to avoid ("unit-testable without a 32 MB live heap").

So the reading is injected, through a new `pacing_arena_in_use_bytes()` — **the one accessor both the escalation predicate and `note_full_cycle_started` now use.**

## The accessor is the more interesting half

`update_major_pacing_backoff`'s doc already claims:

> Deliberately measured on the SAME metric the escalation gate reads (`arena_in_use_bytes`) so the two cannot disagree about whether a full helped.

But both sites called `arena_in_use_bytes()` independently. The guarantee was a convention two call sites happened to honour, not a structure — the same shape as the snapshot-vs-predicate divergence #7733 fixed by collapsing two formulas into one. Routing an injected reading in and requiring the same value back out is what makes it checkable.

The override is `#[cfg(test)]`: it compiles out of every shipping build, so it is not a runtime knob and the GC knob kill-policy doesn't apply.

## Tests, and proof they can fail

Three, all deterministic (thread-local, no timing, no allocation):

- `escalating_records_the_pre_full_arena_reading` — the item-2 ask.
- `the_recorded_reading_is_the_one_the_predicate_decided_on` — agreement across several readings.
- `the_floor_is_inclusive_and_below_it_declines` — `floor - 1` declines and records nothing; exactly `floor` escalates. (The `>=` on the floor is why `major_pacing_escalation_threshold_for` adds its `+1` to the growth boundary and not to the floor.)

**Both sabotage arms were run:**

| sabotage | result |
|---|---|
| rewire the recording away from the predicate (the #7726 bug) | all 3 new tests FAIL |
| point `note_full_cycle_started` at a second, independent `arena_in_use_bytes()` (metric drift) | all 3 new tests FAIL |

And in **both** arms the pre-existing negative test still **passed** — so the gap this PR closes is demonstrated, not merely asserted.

## Validation

- `cargo test -p perry-runtime --lib`: 1991 passed, 0 failed.

One thing worth flagging for #7365 rather than this PR: across four repeat runs of the full suite under CPU load I saw 0–2 failures per run, always from `promise::keyed_table::tests::{settling_many_keys,draining_a_heavily_populated_key_ahead_of_another}_is_not_quadratic` and once `gc::tests::inline_generation_gate_contract::sabotaged_parent_gate_strands_a_young_child_the_shipped_gate_keeps`. Those are wall-clock-ratio assertions, which is a plausible mechanism for "#7365: fails a different number of tests on every run". None of the three is touched by this PR, and my three new tests never appeared in any run.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved major garbage-collection pacing consistency by aligning escalation checks with recorded arena usage.
  * Ensured pacing behavior handles threshold values correctly, including the inclusive lower boundary.

* **Tests**
  * Added coverage for positive pacing escalation, metric consistency, and below-threshold behavior.
  * Added safeguards against metric drift and incorrect pacing wiring.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1439
+**Current Version:** 0.5.1440
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1439"
+version = "0.5.1440"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1439"
+version = "0.5.1440"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1439"
+version = "0.5.1440"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1439"
+version = "0.5.1440"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1439"
+version = "0.5.1440"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1439"
+version = "0.5.1440"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1439"
+version = "0.5.1440"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1439"
+version = "0.5.1440"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1439"
+version = "0.5.1440"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1439"
+version = "0.5.1440"
 dependencies = [
  "anyhow",
  "perry-hir",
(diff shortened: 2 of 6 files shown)


proggeramlug:
## Merging as v0.5.1440

I filed item 2, so I checked this one harder than usual — and the load-bearing claim holds under my own hands.

### The sabotage, re-run here

| | 3 new tests | pre-existing negative test |
|---|---|---|
| baseline | ok, ok, ok | ok |
| **metric drift** (recorder points at its own `arena_in_use_bytes()`) | **FAILED, FAILED, FAILED** | **ok** |

That last cell is the whole argument: the negative-only test cannot distinguish "declined correctly" from "the recording never ran", because both produce `0`. Demonstrating that — rather than asserting it — is what makes this worth landing as its own PR.

(One correction to my own process: my first run of this sabotage showed "4 passed" and no failures, and I nearly recorded it as not reproducing. I had filtered on `pacing`, which matches two `zeal_pacing_*` tests and #7733's two older ones — and **none** of the three new tests. Second time today I've checked a claim against the wrong subject; the filter is not the test.)

### The accessor is the better half of this PR

`update_major_pacing_backoff`'s doc already claimed the two sites read the same metric "so the two cannot disagree about whether a full helped" — but both called `arena_in_use_bytes()` independently. **That was a convention two call sites happened to honour, not a structure.** It is precisely the shape #7739 fixed by collapsing the snapshot and the predicate onto one boundary helper, and it recurred one layer down in the same subsystem.

Routing both through `pacing_arena_in_use_bytes()` and requiring the injected value back out is what makes the claim checkable rather than aspirational. Verified in code: `note_full_cycle_started` at `policy.rs:1591` and the predicate at `:2757` both call it, and the override is `#[cfg(test)]`-gated so it compiles out of shipping builds — correctly outside the GC knob kill-policy.

### Why it wasn't testable before, recorded properly

The explanation is worth keeping: the boundary is `max(floor, baseline × growth + 1)` and **the floor always dominates from below**, so no baseline manipulation gets the threshold under it; and `PERRY_GC_MAJOR_PACING_FLOOR_MB` can't help because `major_pacing_config` is a process-wide `OnceLock` that only takes effect if this test happens to run first. That leaves a 32 MB live heap — exactly what `major_pacing_escalation_threshold_for` was factored out to avoid. Injection was the right escape.

`the_floor_is_inclusive_and_below_it_declines` also pins why the `+1` lives on the growth boundary and not the floor, which is the kind of detail that gets "simplified" wrongly later.

### Scope

Correctly stated as **not** closing #7737: item 1 landed in #7740, item 3 needs the pinned bench host, item 4 is a maintainer action. Leaving the issue open with two items outstanding is the honest bookkeeping.

Gates 21/21.

Issue #6759, opened 2026-07-22
Architecture: adopt V8's object-model construction — explicit runtime state, self-describing headers, shape tree (phases A–C)


proggeramlug:
## Motivation

The object-op perf campaign (#6748 → #6749, #6757) closed most of the gap to node on hot micro-ops, but every win had the same shape: **manually skipping side-table probes that V8 never does at all**. Current state (2400-op micros; node in parens):

| op | before campaign | now | node |
|---|---|---|---|
| `in` hit | 259ms | 0.3ms | 0.4ms — parity |
| `in` miss | 267ms | 4.0ms | 0.6ms |
| 1-field value defines | 527ms (quadratic) | 4.2ms | 3.4ms — ~parity |
| fresh getter defines | 527ms | 12.6ms | 3.7ms |
| `jiti/babel.cjs` module init (pi startup) | 688ms | ~225ms | 44ms |

The residual is structural, not one hotspot. Two architectural facts explain the remaining 2–7×:

1. **Metadata lives in ~200 `thread_local!` side tables keyed by raw address** (property attributes, accessor descriptors, prototypes, exotic-kind, typed-array/buffer/proxy registries…). Answering "what is this receiver / what are property X's attributes" requires *probing* several TLS + HashMap structures. V8 answers with 2–3 dependent **loads**: every object's first word is its hidden-class (Map) pointer, from which prototype, descriptors, attributes, and elements-kind are directly reachable. Address-keying also creates the recurring **stale-address-reuse hazard class** (a freed object's registry entries read by the address's next tenant) that many defensive guards in `object/` exist to contain.
2. **`thread_local!` is perry's isolate mechanism.** `worker_threads` rely on per-thread tables for isolation, so the tables cannot become plain globals. V8's answer is an explicit `Isolate*` threaded through calls in a register.

Node being single-threaded is irrelevant — V8 pays for neither TLS lookups nor side-table probes *by construction*. This issue proposes adopting the same construction in three phases.

---

## Phase A — explicit runtime state (perry's `Isolate*`)

**Goal:** one TLS access per runtime entry point instead of one per table access.

[...]

PR #7753, opened 2026-08-10
perf(codegen,runtime): polymorphic property-read cache + `arr.length` short-circuit — interp.ts 3.96s → 2.39s

proggeramlug:
## The gap

`gc-handoff/apps/interp.ts` — a tree-walking interpreter (lexer, precedence-climbing parser, recursive `let`, closures over environments), 189 statements, no dynamic remainder — ran **3.96 s** against Node 26.5.1's **0.32 s**. That is **12.3× Node**, three times worse than any synthetic benchmark in the corpus (all now 2.3–4.0×), 2.3× worse than scriptc, and it is the only program in the corpus that resembles real software.

GC is not involved: 0.04 s of pause across the whole run, 38 minors, zero fulls.

It is now **2.39 s** (**1.66×** faster, 12.3× → 7.4× Node).

## Root cause

**The per-site property-read cache holds exactly one entry.** `evalNode` dispatches on `n.kind === "num" | "str" | "var" | "bin" | "if" | …`, so the receiver at those sites cycles through five shapes and the single entry is wrong on essentially every read. Each miss re-derives the receiver kind from scratch — proxy band, closure magic, the registered-buffer and typed-array registries (both behind thread-locals), the accessors latch — then linear-scans the keys array with a `js_string_equals` per key. That was **34%** of the reduced program.

It also explains the profile lines that made no sense on their face: `typedarray::lookup_typed_array_kind` at 3.4% and `buffer::header::is_registered_buffer` at 1.9% **in a program that uses neither typed arrays nor Buffers**. They were not a separate problem — they were inside the miss handler.

### Three hypotheses died first, and one of the refutations was misread

* **String-keyed union tags.** `js_jsvalue_equals` + `from_utf8` + `memcmp` + `js_string_equals` = 23% of the profile. Refuted: `tag_str.ts` vs `tag_num.ts` are both 0.06 s, and converting the whole interpreter to numeric tags moved 3.88 → 3.71 s (4%). Most of that 23% was the *keys scan inside the miss handler*, not the user's `===`.
* **Runtime-built heap strings compare slowly.** Refuted: `streq_sub.ts` is 0.05 s vs Node's 0.07. Interning every identifier at lex time: 3.88 → 3.87. Numeric opcodes: 3.88 → 3.89.
* **A megamorphic cliff at some shape count.** A clean arity sweep (`meg{2..12}.ts`, loop body and array size held fixed) showed a flat 3.0–3.7× at every arity, and that was read as "no cliff, so not the cause". It is the opposite: **flat-and-already-3× from two shapes onward is exactly the signature of a one-entry cache.** The sweep could never show a cliff because there is nothing to fall off.

The discriminating probe was `bench/a_flatnode.ts` — the same interpreter with every AST node built from one shape, so every `n.*` read is monomorphic, with recursion depth, allocation count, string traffic and the environment chain all held constant. **3.88 → 2.84 s.** That is what this PR went after.

## Change 1 — polymorphic ways

`@perry_ic_N` widens `[8 x i64]` → `[12 x i64]`: the MRU entry `[token, slot, epoch]` keeps its exact prior meaning, followed by four `(token, slot)` ways and a victim counter. The miss handler cascades the shape it evicts into a way instead of discarding it.

The way compares are emitted **inside the miss block, below the feedback records and above the call**, so a monomorphic site executes the identical instruction sequence it did before — the new work is reached only by a site that was already going to call the runtime, and the typed-feedback counters are unchanged.

Two things had to be right, and a test found each:

* **The MRU token must be evicted from the ways when promoted**, or a way permanently duplicates word 0 and a k-shape rotation caches only k−1.
* **The ways must accept keys-POINTER tokens.** Shape-ID-only ways need no epoch validation, which makes them look like the safe design — and they are useless: a plain object literal is built through a generated `__AnonShape_*` constructor, so it has a real `class_id` and primes a keys pointer. Shipped that way it was a measured **6% regression**, the compare sequence running on every miss and never once hitting.

Admitting pointer tokens inherits #6080a — a freed keys-array address can be recycled under a different shape, and a stale way would pointer-match and load the wrong slot *silently*. So the ways share word 2's epoch snapshot: the emitted predicate requires `cache[2] == @PERRY_IC_EPOCH`, and `pic_prime_get` **wipes every way whenever it writes a new epoch**, dropping the evicted token too. A readable way is always one primed in the epoch word 2 still holds.

→ **3.96 → 3.01 s.** `js_object_get_field_ic_miss` fell from 9.0% of the profile to 1.0%.

## Change 2 — `arr.length` short-circuit

With `evalNode` fixed, the entire remaining miss cost moved to one place: 1143 of 5241 leaf samples, all from `lookup`, none of it a polymorphic object read. It was `names.length`.

The inline cache requires a `GC_TYPE_OBJECT` receiver by construction (#72), so **every** dynamic `.length` misses permanently, by design — and then walks a ladder built for objects, which repeats the registry probes in `js_object_get_field_by_name` before reaching the array arm. For a variable lookup written the ordinary way, `for (i = 0; i < names.length; i++)`, that single read was **22% of total run time**.

`js_object_get_field_ic_miss` now answers it directly when the receiver's `GcHeader` says `GC_TYPE_ARRAY` — a genuine dense array, since buffers, typed arrays, lazy arrays, Sets and Maps all carry distinct `obj_type`s and an `Array` subclass instance is an `ObjectHeader`. `js_array_length` still resolves growth-forwarding stubs, proxies and subclass receivers, and the expression returned is the one the by-name array arm already computes — a short-circuit, not a second implementation.

→ **3.01 → 2.39 s.**

## Tests

New, and each one fails if the thing it names is undone:

| test | pins |
|---|---|
| `alternating_shapes_all_become_inline_resolvable` | a k-shape rotation caches all k (caught the MRU-duplicate bug) |
| `pointer_tokens_do_reach_a_way` | ways are not narrowed back to ID tokens (the 6% regression) |
| `an_epoch_change_wipes_every_way` | the #6080a discipline, both halves |
| `monomorphic_site_never_fills_a_way` | the monomorphic path is untouched |
| `overflow_rotates_without_corrupting_pairs` | >capacity degrades to a miss, never a wrong slot |
| `array_length_short_circuit_agrees_with_the_full_ladder` | equivalence with `js_object_get_field_by_name_f64` for empty/small/grown arrays, a same-length non-`length` key, and `length` on a plain object |
| `pic_cache_words_match_codegen` / `pic_cache_layout_matches_runtime` | the paired width constant — a narrower global would be an out-of-bounds store from the runtime |
| `generic_property_get_tries_ways_before_calling_the_miss_handler` | the way block stays *above* the call (below it, it silently stops paying for itself) |

## Measurements

Quiet M1 mini, best-of-5, absolute seconds. Outputs verified byte-identical to `node --experimental-strip-types` **before** timing.

| program | before | after |
|---|--:|--:|
| `interp.ts` | 3.96 | **2.39** |
| `b_fib.ts` (reduced case) | 3.88 | **2.32** |

Protected floors — all hold, each also A/B'd against a same-host v0.5.1434 build:

| | churn | churn_alloc | push_cls | push_num | churn_read | cycles | deeplist | tree | tree_wide | retain | retain_wide | fib40 |
|---|--|--|--|--|--|--|--|--|--|--|--|--|
| after | 0.42 | 0.36 | 0.35 | 0.13 | 0.02 | 0.19 | 0.24 | 1.65 | 2.11 | 0.53 | 1.08 | 0.39 |
| floor | 0.42 | 0.38 | 0.36 | 0.15 | 0.03 | 0.20 | 0.26 | 1.67 | 2.15 | 0.56 | 1.12 | 0.41 |

`gc-handoff/apps/iso_miss.ts` prints `checksum 437840 misses 0` — gated on the **miss counter**, not the aggregate, because a perf change has previously made `interp.ts`'s total read correct while a silent-wrong-answer GC bug (#7682) was fully intact. Also clean under `PERRY_GC_VERIFY_EVACUATION=1` and `PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800`, which matter here because the ways hold raw heap addresses in a global no GC scanner can see.

## What is left

`interp.ts` is 7.4× Node, not the ~5× scriptc reaches. The remaining profile is `js_jsvalue_equals` (15.7% — `===` is still a runtime call per comparison, and an inline fast path only resolves the ~20% of comparisons that are *true*), the per-object GC layout tables on the allocation path (7.7%, the #7510/#7469 area), write barriers (5.9%), and `js_dyn_index_get`'s registry probes (~3% — the same "route by `GcHeader` before probing address registries" shape fixed here twice). None of those is a single-mechanism gap the way this one was.

Refs #5094, #6759, #7469.

https://claude.ai/code/session_015JgLM9UWGa6WAMix7CvhQJ

## Summary by CodeRabbit

* **Performance**
  * Improved property-read performance with expanded polymorphic caching.
  * Added faster handling for array `.length` reads while preserving proxy, forwarding, and subclass behavior.
  * Improved cache efficiency across multiple object shapes and property metadata formats.

* **Bug Fixes**
  * Strengthened cache invalidation across garbage-collection cycles.
  * Improved correctness and recovery when cache entries are rotated or temporarily unavailable.

* **Tests**
  * Added regression, correctness, garbage-collection, and memory-safety coverage for property reads and array lengths.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1441
+**Current Version:** 0.5.1442
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1441"
+version = "0.5.1442"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1441"
+version = "0.5.1442"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1441"
+version = "0.5.1442"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1441"
+version = "0.5.1442"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1441"
+version = "0.5.1442"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1441"
+version = "0.5.1442"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1441"
+version = "0.5.1442"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1441"
+version = "0.5.1442"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1441"
+version = "0.5.1442"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1441"
+version = "0.5.1442"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1441"
+version = "0.5.1442"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1441"
+version = "0.5.1442"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1441"
+version = "0.5.1442"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1441"
+version = "0.5.1442"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1441"
+version = "0.5.1442"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1441"
+version = "0.5.1442"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1441"
+version = "0.5.1442"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1441"
+version = "0.5.1442"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1441"
+version = "0.5.1442"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1441"
+version = "0.5.1442"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1441"
+version = "0.5.1442"
 dependencies = [
  "perry-ffi",
  "scraper",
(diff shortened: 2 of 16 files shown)


proggeramlug:
## Merging as v0.5.1442

### I went after the soundness risk first

The ways hold **raw heap addresses in a global no GC scanner can see**, so the epoch discipline is the whole safety argument. Three checks:

1. **The wipe has a subject.** Disabling `if !epoch_held && c[PIC_WAY_STATE] >= 0` fails `an_epoch_change_wipes_every_way` at `ic_miss.rs:1334`.
2. **The megamorphic-latch interaction is safe, and I checked it rather than reading the comment.** Skipping the wipe when `PIC_WAY_STATE < 0` looked like a hole — word 2 is set to the new epoch *unconditionally*, so stale ways surviving a latched period could become readable once the countdown re-arms. It isn't: `ic_miss.rs:409-413` zeroes every way **before** setting `c[PIC_WAY_STATE] = -PIC_LATCH_RETRY`, so a latched site has nothing to wipe. The invariant holds, but it is a two-place invariant and worth a comment cross-reference at the wipe site.
3. **Under real evacuation pressure**, on a 5-shape rotation of my own:
   ```
   PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800
     → exit 0, correct answer
     → forced_collections=4218 copying_minors=4218 moved_objects=960059
   PERRY_GC_VERIFY_EVACUATION=1 → exit 0, correct answer
   ```
   Nearly a million objects relocated with the ways live. That is the run that matters.

All eight named tests present and passing; a 5-shape A/B on my (loaded) box gives 0.65 → 0.51 s.

### The diagnosis is the best part, and it includes a correction

> A clean arity sweep showed a flat 3.0–3.7× at every arity, and that was read as "no cliff, so not the cause". It is the opposite: **flat-and-already-3× from two shapes onward is exactly the signature of a one-entry cache.** The sweep could never show a cliff because there is nothing to fall off.

Re-reading a refutation as a confirmation is rare and worth naming. Three hypotheses were killed with measurements first (string-keyed tags 4%, interning +0.01 s, numeric opcodes −0.01 s), and `a_flatnode.ts` — the same interpreter with every node one shape, holding recursion depth, allocations and string traffic constant — is a genuinely discriminating probe rather than a plausible one.

It also explains the profile lines that made no sense: `lookup_typed_array_kind` at 3.4% and `is_registered_buffer` at 1.9% **in a program using neither**. They weren't a second problem; they were inside the miss handler.

### Two design traps, each caught by a test rather than reasoning

- **The MRU token must be evicted from the ways on promotion**, or a k-shape rotation caches k−1 forever.
- **Shape-ID-only ways look like the safe design and are useless** — a plain object literal goes through a generated `__AnonShape_*` constructor, so it has a real `class_id` and primes a keys pointer. Shipped that way it was a measured **6% regression**: "the compare sequence running on every miss and never once hitting." A design that is safer *and* does nothing is the hardest kind to catch by review.

The megamorphic latch is the right answer to the asymmetry it names (+37% on a 7-shape site vs 2.5× on a 5-shape one), and making it a **countdown rather than a one-way door** is what stops a transiently-wide site from being punished forever.

### Change 2

`arr.length` missing *permanently by design* — because the IC requires a `GC_TYPE_OBJECT` receiver (#72) — and costing **22% of total run time** for one read in an ordinary `for (i = 0; i < names.length; i++)` is the kind of thing that hides precisely because it looks like nothing. Routing by `GcHeader` before probing address registries is the same shape fixed twice here.

### Verification posture

Gating `iso_miss.ts` on the **miss counter** rather than the aggregate is exactly right, and the stated reason is the one that matters: a perf change previously made `interp.ts`'s total read correct while #7682 was fully intact.

Protected floors all hold, A/B'd against a same-host v0.5.1434 build. The "what is left" section is honest — 7.4× Node, not scriptc's ~5× — and correctly says none of the residue is a single-mechanism gap the way this one was.

Gates 21/21.

PR #7754, opened 2026-08-10
feat(ui-ios): drive onFrame from CADisplayLink and add frame-time metrics

proggeramlug:
## Why

Perry currently cannot answer any frame-related question about its own iOS UI, because nothing observes frames.

`onFrame` was driven from the 8 ms repeating `NSTimer` in `app.rs` via `js_frame_pump_default()`. That is wrong for anything frame-shaped in two independent ways:

- **Not vsync-aligned.** An 8 ms free-running `NSTimer` beats against a 8.333 ms (120 Hz) or 16.667 ms (60 Hz) refresh instead of landing on it, and is subject to run-loop coalescing.
- **Millisecond-quantized.** `js_timer_now()` is `elapsed().as_millis() as f64` — whole milliseconds, ~12% of a 120 Hz frame budget. A p99 computed from that clock is noise.

`docs/src/ui/on-frame.md` advertised "CADisplayLink wiring TBD" for Apple platforms. This lands it for iOS.

## What

**`perry-ui-ios/src/frame_driver.rs`** — a real `CADisplayLink` driving `js_frame_tick`. The timer pump keeps driving `setTimeout`/`setInterval`, microtasks, the stdlib pump and the GC step; it no longer drives frames.

Three details are load-bearing rather than incidental:

- **`NSRunLoopCommonModes`, not the default mode.** A `UIScrollView` drag switches the run loop to `UITrackingRunLoopMode`; a link registered only in the default mode goes silent for the entire gesture — exactly the window a scrolling benchmark exists to measure.
- **The timestamp is rebased, not passed through.** `CADisplayLink.timestamp` is on the `CACurrentMediaTime` base (seconds since boot) while `onFrame`'s contract is monotonic ms since app start. Passing the raw media time would silently break every app doing `t - startTime`. The first tick pins the media clock against `js_timer_now()`; later ticks report an offset from that pin, keeping the epoch while gaining sub-microsecond resolution.
- **Pause state is polled from the timer pump.** A paused link gets no ticks, so it cannot observe a newly-registered callback and wake itself.

**`perry-runtime/src/frame_metrics.rs`** — `PERRY_FRAME_STATS=1`, off by default. p50/p95/p99 frame times, longest frame, dropped-frame count over a rolling 8192-frame window.

Drops derive from the driver-supplied nominal frame duration rather than a hardcoded 60 Hz, which is what makes the same code correct on ProMotion: `dropped = round(interval / expected) - 1`, so jitter within half a frame is not a drop and a doubled interval is exactly one.

Three cases that would otherwise produce confident wrong numbers are handled explicitly:

- **Deliberate pauses are not 4-second frames.** Backgrounding stops the link; attributing the gap would report hundreds of phantom drops and poison the p99.
- **One hitch cannot dominate the drop count.** A 10 s stall is recorded at full length as the longest frame, but its drop attribution is capped.
- **`js_frame_metrics_report()` returns its sample count.** Zero distinguishes "every frame was fast" from "the link never ticked" — indistinguishable in a summary line otherwise, and the second is the failure mode where a probe reports success without its subject having run.

Collection keeps the link awake on its own, so an app with **no** `onFrame` subscribers — a plain scrolling list — still produces a full frame trace.

## Testing

11 unit tests in `perry-runtime --lib`: nearest-rank percentiles, steady 120 Hz, the doubled-interval case, sub-half-frame jitter, the attribution cap, discontinuity, non-monotonic driver timestamps, ring-buffer wrap, the disabled path, report-zero.

Unit tests rather than an integration suite deliberately — per-PR CI runs `--lib --bins`, so coverage under `crates/*/tests/` would not gate this.

**Sabotage-checked**, not merely green: flipping the drop threshold from `spans > 1.0` to `> 2.0` turns `a_doubled_interval_counts_as_exactly_one_dropped_frame` red. The tests can fail.

Also verified:

- 1999 `perry-runtime` lib tests pass, 0 failed
- `perry-ui-ios` compiles clean for `aarch64-apple-ios` **and** `aarch64-apple-ios-sim`
- clippy clean on both crates, `cargo fmt --check` clean
- `check_file_size.sh`, `addr_class_inventory.py`, `gc_runtime_root_holders.py` all pass
- `objc2-quartz-core` pinned to `default-features = false` — the default set drags in `objc2-metal` / `-core-video` / `-core-graphics`; narrowing to `CADisplayLink` + `CAFrameRateRange` adds **zero** new packages to `Cargo.lock`

## Not in this PR

- **iPhone 120 Hz needs `CADisableMinimumFrameDurationOnPhone = YES`** in the bundle's `Info.plist`, which Perry's generated plist does not emit. Without it the frame-rate request is capped at 60 Hz on iPhone (iPad ProMotion has no such opt-in). Adding it touches the compiler driver and is a battery-policy decision — probably wants to be opt-in via `package.json`. Documented in `on-frame.md`.
- No `UIApplicationDidBecomeActive` observer, so an app suspended without a pause transition the driver observes still shows one long interval. The attribution cap bounds the damage; treat a lone multi-second `max` as a lifecycle artifact.
- macOS / tvOS / visionOS / Android / GTK4 / Windows still pump `onFrame` from their main loops, unchanged.

## Testing on device

Not yet run on hardware — this is the instrument, not a measurement. Real-device Release numbers need signing setup. The intended next steps are launch→first-frame signposts and the widget-table lifetime fix (`register_widget` only ever pushes; `clear_children` resolves handles by linear scan over a vector that never shrinks), both of which distort list benchmarks today.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6409,6 +6409,7 @@ dependencies = [
  "objc2",
  "objc2-core-foundation",
  "objc2-foundation",
+ "objc2-quartz-core",
  "objc2-ui-kit",
  "perry-runtime",
  "perry-ui",
--- a/benchmarks/ios-ui/bench.ts
+++ b/benchmarks/ios-ui/bench.ts
@@ -35,7 +35,11 @@ const LIVE_LABELS = 50
 // Rows built once for the text-heavy phase.
 const TEXT_ROWS = 300
 
-const PHASE_SECONDS = 8
+// 20s rather than 8s: the phases have to be long enough for a human to react
+// to a phase marker and scroll inside the window, and long enough to yield
+// several reports of steady state after the transition's own cost has washed
+// out of the first one.
+const PHASE_SECONDS = 20
 
 type Phase = {
     name: string
@@ -73,7 +77,10 @@ const phases: Phase[] = [
         name: "idle",
         enter: (): void => {
             clearContent()
-            buildLabels(20, "idle row")
+            // 60, not 20: every phase must overflow the screen or it cannot be
+            // dragged, and a phase that cannot be dragged silently measures the
+            // static path while looking like a scroll result.
+            buildLabels(60, "idle row")
         },
         tick: (): void => {},
     },
--- a/benchmarks/ios-ui/bench.ts
+++ b/benchmarks/ios-ui/bench.ts
@@ -1,0 +1,180 @@
+// iOS UI frame-time benchmark.
+//
+// Cycles through the workloads a cross-platform UI framework is normally
+// judged on, printing a marker before each so the `[frame-stats]` lines the
+// runtime emits can be attributed to a workload.
+//
+// Run with metrics on:
+//
+//     PERRY_FRAME_STATS=1 perry run benchmarks/ios-ui/bench.ts --target ios
+//
+// `PERRY_FRAME_STATS_INTERVAL` (default 300 frames) sets the reporting
+// cadence. The scroll and text-heavy phases need you to actually drag the
+// list — nothing here injects touches, and a stationary list measures the
+// idle path, not scrolling.
+
+import {
+    App,
+    VStack,
+    Text,
+    ScrollView,
+    scrollviewSetChild,
+    widgetAddChild,
+    widgetClearChildren,
+    widgetSetBackgroundColor,
+    textSetString,
+    onFrame,
+} from "perry/ui"
+
+// Rows rebuilt per cycle in the add/remove phase. Large enough that the old
+// O(children x every widget ever created) handle scan was visible, small
+// enough to stay inside a frame when it isn't.
+const CHURN_ROWS = 100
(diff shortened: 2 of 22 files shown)


jdalton (commented):
Review of `debefa55c1b83161af7986429e382d1ac596a779` (2026-09-07).

The current diff has grown beyond the original frame-driver description to include widget lifetime and simulator launch/bundle changes; please update the PR body to describe those final changes and their validation. For the driver itself, `poll_pause_state` only marks a discontinuity when subscriber/metrics demand changes. With metrics enabled, background/resume can leave that demand true throughout and record the suspension as a giant frame. The body acknowledges this, so please either add lifecycle-triggered discontinuity handling or label the exported max/drop metrics accordingly, and attach a launch → scroll → background → resume run with a nonzero frame sample count. Cross-compilation and percentile unit tests do not validate the CADisplayLink delivery path.

Validation scope: source/diff inspection; I have not run this PR's build or test suite locally.

Issue #5592, opened 2026-06-23
test262 language/class tail — 197 fails (async-gen static methods, accessor-name computed-in, private brand)


proggeramlug:
**Current state — 197 failing.** _Snapshot from the 2026-06-29 full sweep (perry 0.5.1208, node v26.3.0). Supersedes the original body; the complete current failing-test list is inline below — no re-run needed to start._

Class tail: async-generator static methods (`yield*` delegation — note the `yield-star-sync-next` regression from #5751), computed accessor names with `in`, private-field brand checks, missed negatives. See breakdown below.

## Failure reasons
| n | reason |
|---|---|
| 29 | Uncaught exception: Expected a TypeError to be thrown but no exception |
| 12 | SyntaxError: new.target expression is not allowed here |
| 9 | Uncaught exception: Expected SameValue(«undefined», «1») to be true |
| 8 | Test262:AsyncTestFailure:Test262Error: Test262Error: get [Symbol.itera |
| 7 | TypeError: value is not a function |
| 6 | Uncaught exception: invalid access of c1 private method Expected a Typ |
| 5 | Uncaught exception: Expected a ReferenceError to be thrown but no exce |
| 5 | TypeError: GeneratorFunction is not a function |
| 4 | TypeError: Cannot access private member from an object whose class did |
| 4 | Uncaught exception: Expected SameValue(«1», «1») to be false |
| 4 | TypeError: Cannot read properties of undefined (reading 'access') |
| 4 | Uncaught exception: Expected SameValue(«false», «true») to be true |

## By area
**statements** 140 · **expressions** 57

## All 197 failing tests (path — reason)
<details><summary>full list — click to expand</summary>

```
language/expressions/class/async-gen-method-static/yield-star-sync-next.js    Test262:AsyncTestFailure:Test262Error: Test262Error: get [Symbol.iterator] thisValue Expec
language/expressions/class/async-gen-method/yield-star-sync-next.js    Test262:AsyncTestFailure:Test262Error: Test262Error: get [Symbol.iterator] thisValue Expec
language/expressions/class/class-name-ident-await-escaped.js    Error: Failed to parse <staged-case> Parse error: E
language/expressions/class/class-name-ident-await.js    Error: Failed to parse <staged-case> Parse error: E
language/expressions/class/constructor-this-tdz-during-initializers.js    Uncaught exception: Expected a ReferenceError to be thrown but no exception was thrown at 
[...]

PR #7756, opened 2026-08-10
fix(monomorph): a specialized generic class reports the generic's name, not Gen$num (#7632)

### LINKED ISSUE #7632: constructor.name of a generic-class instance reports the mangled specialization (Gen$num), not Gen
`(new Gen<number>()).constructor.name` reports the MANGLED specialization name
`Gen$num` where node reports `Gen`.

## Repro

```ts
class Base {}
class Gen<T> extends Base {}
class GenNoExtends<T> {}

const a = new Gen<number>();
const b = new Gen();               // no type arguments
const c = new GenNoExtends<number>();
console.log(a.constructor.name, b.constructor.name, c.constructor.name);
```

```
$ node --experimental-strip-types t.ts
Gen Gen GenNoExtends

$ ./t
Gen$num Gen GenNoExtends$num
```

Only the instantiation WITH type arguments is affected — that is the one that
gets monomorphized.

## Same root cause as #7575, different surface

Perry specializes generic classes: `new Gen<number>()` emits a second class named
`Gen$num` (`crates/perry-hir/src/monomorph/mangle.rs`,
`generate_specialized_name`) and constructs THAT. #7575 (PR #7631) fixed the
`instanceof` half by recording `Class::specialized_from` in the HIR and teaching
the `instanceof` chain walk a specialization→generic edge. The user-visible class
NAME is the other half of the same leak, and was deliberately left out of that
PR: it flows through the class display-name registry rather than the class-id
chain, and it can change error-message text (`TypeError: ... is not a
constructor`, `Object.prototype.toString`, stack frames), so it wants its own
parity sweep rather than being bundled into an `instanceof` fix.

`Class::specialized_from` already carries what is needed (`Some("Gen")` on the
specialization), so the likely shape is to register the ORIGIN name as the
class's display name — the `class_display_names` mechanism in
`crates/perry-codegen/src/codegen/string_pool.rs` already exists for #5592's
uniquified class-expression bindings.

## Worth checking in the same pass

- `Gen.name` read off the constructor itself, not an instance.
- `Object.prototype.toString.call(new Gen<number>())`.
- Error subclasses: `class MyErr<T> extends Error {}` — `err.name` and the
  `toString()` prefix.
- Whether any diagnostic or `--print-hir`/`--trace` output is keyed on the
  mangled name and would become ambiguous if two specializations both report
  `Gen`.

proggeramlug:
Fixes #7632.

## The fix

Monomorphization emits a second class for `new Gen<number>()` (`Gen$num`, `monomorph::mangle::generate_specialized_name`) with its own class id, and the instance is stamped with that id — so every id-keyed user-visible surface reported the mangled name. TypeScript erases type arguments; the mangling is Perry's business.

`monomorph/driver.rs` now registers the ORIGIN's name against the specialization's class id in `Module::class_display_names` — the mechanism `codegen/string_pool.rs` already prefers over the registration key when emitting named-class constants (added for #5592's uniquified class-expression bindings). It reads the origin's own *display* name rather than its `name`, so a specialization of an already-uniquified class reports the JS name and not the internal key. Two lines of registration plus a deferred insert, because the collection loop holds `module` immutably.

| | node 26.5.1 | before | after |
|---|---|---|---|
| `(new Gen<number>()).constructor.name` | `Gen` | **`Gen$num`** | `Gen` |
| `(new GenNoExtends<number>()).constructor.name` | `GenNoExtends` | **`GenNoExtends$num`** | `GenNoExtends` |
| `(new Gen()).constructor.name` (no type args) | `Gen` | `Gen` | `Gen` |

## The checklist in the issue, measured

The issue asked what else to check in the same pass. All of these were already correct and stay correct:

- `Gen.name` read off the constructor **binding** — correct before and after. (Worth noting the asymmetry that made this bug easy to miss: `Gen.name` was `Gen` while `a.constructor.name` was `Gen$num`.)
- `Object.prototype.toString.call(new Gen<number>())` → `[object Object]`.
- Error subclasses: `class MyErr<T> extends Error {}` → `e.name === "Error"`, `String(e) === "Error: boom"`, `e instanceof Error`.
- The #7575 `instanceof` half keeps working, including through a nested `class Wrap<T> extends Gen<T> {}`.

On the issue's last checklist item — whether a diagnostic keyed on the mangled name becomes ambiguous: the HIR/codegen keys are unchanged. Only `class_display_names` is written, and it is consulted solely when emitting the JS-visible name constant. `--print-hir` / `--trace` still show `Gen$num`.

## What this does NOT fix, and why I'm flagging it

Two specializations of one generic remain **distinct constructor objects**:

```ts
class Gen<T> { v: T | undefined; }
const a = new Gen<number>(), b = new Gen<string>();
a.constructor === Gen              // node: true   perry: false
a.constructor === b.constructor    // node: true   perry: false
Object.getPrototypeOf(a) === Object.getPrototypeOf(b)  // node: true  perry: false
```

That is the monomorphization model, not the name registry, so it is out of scope here — but it is worth stating plainly that **after this change the two report the same NAME while remaining `!==`**, which arguably makes the identity divergence harder to notice than it was. Filing it separately.

(Curiously `Object.getPrototypeOf(a) === Gen.prototype` is already `true`, so the prototype and constructor edges disagree with each other today. That is part of the same separate question.)

## Validation

- `test-files/test_gap_generic_class_constructor_name_7632.ts` — byte-identical to node 26.5.1. Covers nested generics, two specializations of one generic, the no-type-args case, error subclasses, `Object.prototype.toString`, and the #7575 `instanceof` half.
- New `perry-hir` unit test `a_specialized_class_reports_the_generics_display_name`. Unit as well as gap on purpose: the gap suite is tag-gated, so a regression there sits red for days (#5960). **Verified to fail** when the registration is removed.
- `cargo test -p perry-hir --lib`: 290 passed. `cargo test -p perry-codegen --lib`: 822 passed.
- Targeted parity over 36 class/name/prototype-related gap tests: 35 pass, 1 fails — `test_gap_2159_defineproperty_class_prototype`, which is a listed `parity_fail` in `gap_snapshot.json` (standing gap, issue 2159). A full sweep was not run: this host was at 11 GB free at one point today and a disk-full mid-sweep is its own failure mode, so I scoped the run to the tests that could plausibly be affected and am saying so rather than implying broader coverage.

## Summary by CodeRabbit

- **Bug Fixes**
  - Generic class instances now report their original class name instead of an internal, mangled specialization name.
  - Corrected display names across nested and multiple specializations, inherited classes, error subclasses, and object string representations.
  - Preserved expected constructor bindings and `instanceof` behavior.

- **Tests**
  - Added regression coverage for generic class display names, specialization behavior, inheritance, and type checks.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1440
+**Current Version:** 0.5.1441
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1440"
+version = "0.5.1441"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1440"
+version = "0.5.1441"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1440"
+version = "0.5.1441"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1440"
+version = "0.5.1441"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1440"
+version = "0.5.1441"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1440"
+version = "0.5.1441"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1440"
+version = "0.5.1441"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1440"
+version = "0.5.1441"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1440"
+version = "0.5.1441"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1440"
+version = "0.5.1441"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1440"
+version = "0.5.1441"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1440"
+version = "0.5.1441"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1440"
+version = "0.5.1441"
 dependencies = [
  "serde",
  "serde_json",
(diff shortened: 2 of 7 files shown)


proggeramlug:
## Merging as v0.5.1441

A/B'd against node on my own host:

```
[old] ctor.name: Gen$num   noext: GenNoExtends$num   ...  identity a.ctor===Gen: false
[new] ctor.name: Gen       noext: GenNoExtends       ...  identity a.ctor===Gen: false
[node] ctor.name: Gen      noext: GenNoExtends       ...  identity a.ctor===Gen: true
```

The name is fixed; everything else — `Gen.name`, `notypeargs`, `[object Object]`, the error subclass triple — is unchanged and already correct, and the identity divergence is unchanged in both arms exactly as you disclosed.

Reusing `class_display_names` is the right mechanism: it already exists for #5592's uniquified class-expression bindings, `string_pool.rs` already prefers it over the registration key, and reading the origin's **display** name rather than its `name` is the detail that makes a specialization of an already-uniquified class report the JS name instead of the internal key. Two lines, no new concept.

Verified the unit test has a subject: commenting out `module.class_display_names.insert(class_id, display_name)` fails it with `left: None, right: Some("Gen")` and the message *"reporting `Gen$num` leaks the mangling into `constructor.name`, `TypeError` text and stack frames"*. Unit as well as gap is right for the stated reason — the gap suite is tag-gated, so a regression there sits red for days (#5960).

### The disclosure is the part I want to keep

> after this change the two report the same NAME while remaining `!==`, which arguably makes the identity divergence harder to notice than it was

That is worth having said out loud, and it is the right call anyway: TypeScript erases type arguments, so `Gen` is the correct name, and a wrong name is not worth keeping as an accidental symptom-indicator for a different bug. But the trade is real and the follow-up issue is the right place for it.

The observation that `Object.getPrototypeOf(a) === Gen.prototype` is **already** `true` while `a.constructor === Gen` is `false` — so the prototype and constructor edges disagree with each other today — is the more useful half of that follow-up. It suggests the fix is narrower than "collapse the specializations".

The checklist from the issue was measured rather than assumed, including the asymmetry that made this easy to miss: `Gen.name` was already `Gen` while `a.constructor.name` was `Gen$num`. And the confirmation that HIR/codegen keys are untouched — `--print-hir`/`--trace` still show `Gen$num` — is what keeps diagnostics unambiguous.

PR #7565, opened 2026-08-07
perf(runtime): reach the hot-TLS cache without _tlv_get_addr (#7469 structural half)

proggeramlug:
Closes the structural half of #7469: `_tlv_get_addr` falls from **27.0% to 1.1%** of `churn_alloc` self time, and the three allocation probes get **1.14x–1.18x** faster on the pinned quiet host.

## Requirement zero — re-measured before scoping

Three tickets in this campaign were worked off stale headline numbers, so the first thing here is a fresh symbolicated profile, not a design. `churn_alloc` on the pinned quiet host (`perry-macos`, M1 mini, load 1.5) at `9938cbc1a`, 923 leaf samples: `_tlv_get_addr` is **27.0%** — the ticket's 30.5% has not drifted.

What *had* drifted was the shape. The ticket opened saying "41 distinct call-graph sites — this is diffuse, not one hot caller". It is no longer diffuse. **Seven functions carry 98% of it, and every one of the seven resolves `tls_hot::HOT`:**

| caller | share of `_tlv_get_addr` | of total |
|---|--:|--:|
| `gc::barrier::write_barrier_decoded_parent` | 19.3% | 5.2% |
| `gc::layout_tables::layout_forget_object` | 18.9% | 5.1% |
| `js_object_alloc_class_inline_keys` | 18.5% | 5.0% |
| `arena::allocators::arena_alloc` | 14.9% | 4.0% |
| `js_write_barrier_slot` | 9.2% | 2.5% |
| `gc::barrier::barrier_child_prologue` | 8.8% | 2.4% |
| `gc::layout::typed_shape_layout_entry` | 8.4% | 2.3% |

Two of the seven resolve *nothing else*. Static disassembly of the emitted binary names which thread-local each site resolves (there is no `bl _tlv_get_addr` to grep for on Mach-O — the call is indirect through the TLV descriptor, so the census walks `adrp`/`add` pairs landing in `__thread_vars`).

That attribution chose the design. The accessor is the lever, not the call graph: the ticket's option 1, "thread a context pointer through generated code", would have to cross every runtime FFI boundary against 2994 `.with()` sites over 255 `thread_local!` blocks; making `hot()` free fixes all seven at once and composes with everything downstream.

## What changed

[...]

PR #7758, opened 2026-08-10
perf(tls): thread-locals are on the fast path by default, with a gate that can fail (#7469)

proggeramlug:
## `_tlv_get_addr` was 18-20% of the worst realistic program. The mechanism was fine; the coverage policy was the bug.

On Darwin every `thread_local!` access is an out-of-line call to `_tlv_get_addr`
in libdyld. Two good PRs already attacked it — #7474 built the `HotTls` address
cache, #7565 made reaching that cache free via Apple aarch64's directly
addressable pthread TSD array. Both are preserved here unchanged.

But the measured share kept coming back:

| build | workload | share |
|---|---|--:|
| after #7565 | `churn_alloc` | 0% |
| later | `churn_alloc` | 8-9% |
| later | `interp` / `retain` | 11% |
| this PR's base (`27d5358d0`) | **`asyncpipe`** | **18.2%** |

`HotTls` had **16 hand-wired slots against ~520 `thread_local!` declarations**,
and the 16 were curated against whichever workload was profiled last — the
allocation path. `churn` is covered by construction and reads 0% forever.
`asyncpipe` pays 18.2% through Map/Set registries, buffer brands, descriptor
state and field-lookup tails that were on nobody's list. Adding a slot took four
manual steps including a hand-written test, and **forgetting them produced a
working slow path, not a build error**.

## What this lands: option 2, invert the default

`crate::perry_thread_local!` — same syntax as `thread_local!`, same `with` /
`try_with` at every call site, so converting a *declaration* converts all of its
uses. The value's address lands in a generic slot of the same per-thread cache,
so a read is loads instead of a call. **Nothing to wire**: no slot, no provider
function, no line in `fill`, no line in a test.

It also removes the hazard the old contract needed a test to catch. The untyped
named slots could hand out a correctly-typed reference to the *wrong* object if
`fill` was mis-wired; here the storage, the resolver and the key's `T` all come
from one declaration, so the mis-pairing cannot be expressed.

And it is **safer on thread teardown** than what it extends. `HotCell<T, GUARD>`
takes `GUARD = needs_drop::<T>() as usize` from the macro:

* `RefCell<HashMap<…>>` → a one-element guard array whose `Drop` runs *before*
  the value's (fields drop in declaration order), un-publishing this thread's
  cached address. A later access falls back and gets std's "accessed during or
  after destruction" panic instead of reading a dropped map.
* `Cell<u64>` → a zero-length array, which has no drop glue, so `HotCell` has
  none: no destructor registered, std's `const`-init fast path preserved.

The 16 named fields are unchanged and stay a **closed set** — a fixed offset is
one load cheaper than a claimed slot, and the allocation path is where that
matters. 155 declarations across 51 files use the macro. One more fix in the
same family: `js_inline_arena_state` was resolving `INLINE_STATE`/`ARENA`
through `.with()` even though both are *already* named slots — a covered entry
the caller wasn't using (5.2% of `interp`'s remaining calls).

## Results (quiet M1 mini, best-of-5 interleaved, absolute seconds)

Base and arm at the same commit (`27d5358d0`), both built `-p perry
-p perry-runtime-static -p perry-stdlib-static`, isolated target dirs, outputs
verified against node **before** timing.

`_tlv_get_addr` share, symbolicated, 8s `sample`, two runs in agreement:

| program | base | this PR |
|---|--:|--:|
| `asyncpipe_big` | 18.2% | **1.9%** |
| `interp_big` | 10.6% | **0.4%** |

Wall clock, interleaved best-of-5:

| bench | base | this PR | delta |
|---|--:|--:|--:|
| **asyncpipe** | 1.12 | **1.06** | **-5.4%** |
| **interp** | 4.15 | **4.03** | **-2.9%** |
| churn | 0.42 | 0.42 | 0.0% |
| churn_alloc *(control)* | 0.38 | 0.38 | 0.0% |
| churn_read | 0.02 | 0.02 | 0.0% |
| push_cls | 0.36 | 0.36 | 0.0% |
| push_num | 0.14 | 0.14 | 0.0% |
| cycles | 0.19 | 0.19 | 0.0% |
| deeplist | 0.25 | 0.25 | 0.0% |
| tree | 1.68 | 1.68 | 0.0% |
| tree_wide | 2.18 | 2.17 | -0.5% |
| retain | 0.54 | 0.55 | +1.9% |
| retain_wide | 1.11 | 1.12 | +0.9% |
| fib40 | 0.40 | 0.40 | 0.0% |

`churn_alloc` is the control: its paths were already covered by the sixteen
named fields, so it is exactly the benchmark that must *not* move. It does not.
The asyncpipe/interp deltas reproduced across three independent interleaved runs
(-6.4%/-7.0%/-5.4% and -3.7%/-3.4%/-2.9%); everything else stayed inside +-2%,
which is this host's noise floor while a second agent was benchmarking on it
(both arms inflated equally by interleaving; absolute seconds run ~3% above the
same host's quiet reference).

## The gates

Two, because they fail on different things.

**`scripts/check_thread_locals.py`** — structural. A new raw `thread_local!` in
`perry-runtime` is a build error unless recorded in
`scripts/thread_local_cold_allowlist.json` as deliberately cold. The counts
ratchet in *both* directions: a file that loses a declaration also fails,
because a stale entry is one nobody has to justify any more. It also fails as
declarations approach `HOT_SLOT_CAPACITY`, since slot exhaustion is correct but
silent. `--self-test` drives all four rejections.

**`scripts/tls_budget_gate.sh`** — outcome, and its design is about vacuity.
Profiling `churn_alloc` — the benchmark every previous fix was tuned against —
would pass forever while the real cost grew, because churn's thread-locals are
exactly the covered ones: a gate green because its subject never ran. So the
subjects are `benchmarks/tls-budget/asyncpipe.ts` and `interp.ts`, and
`scripts/tls_budget_check.py` refuses a pass unless the run proves it was live:
`PERRY_TLS_HOT_STATS=1` must report `direct_tsd=1` (otherwise `hot()` is itself
calling `_tlv_get_addr`, the mechanism is inert, and a low share would mean the
program resolved nothing) and `claimed` above a floor no allocation
microbenchmark clears. Seven rejections, all `--self-test`ed, run compiler-free
on every PR.

### Sabotage checks

**The budget gate goes red on the regression it exists to catch.** Two
declarations reverted to raw `thread_local!` — `buffer/header.rs`'s
`BUFFER_REGISTRY` block and `state.rs`'s `STATE_PTR` block — rebuilt, gate re-run:

| program | clean | sabotaged | budget | verdict |
|---|--:|--:|--:|---|
| `asyncpipe` | 1.8% | **7.4%** | 5.0% | FAIL |
| `interp` | 0.3% | **8.6%** | 3.0% | FAIL |

`GATE_EXIT=1`, and the liveness assertions stayed satisfied throughout
(`direct_tsd=1`, `claimed=111`/`102`) — so it failed on the *budget*, not on a
broken run. Restoring the macro restores `GATE_EXIT=0`.

A second, unplanned demonstration: while this branch was being rebased, `main`
added a raw `thread_local!` to `gc/policy.rs`. The structural ratchet failed the
build on it immediately, which is exactly the intended workflow — it is recorded
as cold in this PR rather than converted in a file this change does not own.

**The teardown guard's test is also sabotage-checked.** Neutering
`SlotGuard::drop` makes `teardown_unpublishes_a_dropping_value` fail with
`left: 2` — it reads the dropped `Vec` — proving the test detects a real
use-after-free rather than passing vacuously.

### Not wired into branch protection

Deliberately. A new gate has never been green, so promoting it immediately would
block every open PR (CLAUDE.md's corollary). That is a maintainer action after
the first observed green run on `main` — and per the same corollary, not
optional follow-through.

## Correctness

* All outputs byte-identical to `node --experimental-strip-types` v26.5.1,
  verified **before** timing, on both arms, for all 15 programs plus the two
  `_big` profiling variants.
* Canary `iso_miss` prints `checksum 437840 misses 0` on both arms.
* `cargo test -p perry-runtime --lib`: 1995 passed, 0 failed.
* Thread teardown and multi-thread: `converted_declarations_survive_thread_turnover`
  drives 64 short-lived threads through the converted registry probes and
  asserts indices are claimed per *declaration*, not per thread;
  `a_generic_slot_is_per_thread`, `concurrent_first_touch_claims_one_slot_per_declaration`
  and `teardown_unpublishes_a_dropping_value` cover the rest.

## Overlap

The three largest `_tlv_get_addr` callers on `asyncpipe` are
`is_registered_buffer` / `is_registered_set` / `is_registered_map` — 55% of it —
which is also the subject of the concurrent "remove registry probes from generic
paths" work. The two changes compose (this one makes the resolution cheap; that
one removes the call), but they overlap in *measurement*: whichever lands second
will show a smaller delta on those symbols. Flagging rather than resolving
unilaterally.

## New Features
- Improved thread-local performance across the runtime while preserving existing behavior.
- Added automated performance monitoring and budget checks on supported Apple platforms.
- Added deterministic workloads for validating runtime performance.
- Added safeguards for cache capacity, cleanup, and thread reuse.

## Bug Fixes
- Improved safety when thread-local values are destroyed or threads are reused.

## Documentation
- Added guidance for tracking thread-local usage and performance budgets.
- Updated the project version to 0.5.1444.


proggeramlug:
## Merging as v0.5.1444, with one test fixed on the branch

### The framing is the finding

> The mechanism was fine; the coverage policy was the bug.

**16 hand-wired slots against ~520 declarations**, curated against whichever workload was profiled last — so `churn_alloc` reads 0% forever while `asyncpipe` pays 18.2% through Map/Set registries, buffer brands and field-lookup tails that were on nobody's list. And adding a slot took four manual steps where **forgetting them produced a working slow path, not a build error**. That is the whole reason the number kept coming back after two good PRs had already fixed it.

Inverting the default removes the curation entirely, and it also removes a hazard the old contract needed a test to catch: untyped named slots could hand a correctly-typed reference to the *wrong object* if `fill` was mis-wired. With storage, resolver and key `T` all from one declaration, the mis-pairing is inexpressible.

**The teardown reasoning is the subtlest part and it is right.** `GUARD = needs_drop::<T>()` gives a droppable value a one-element guard array whose `Drop` runs *before* the value's — fields drop in declaration order — so the cached address is un-published and a later access gets std's "accessed during or after destruction" panic instead of reading a dropped map. `Cell<u64>` gets a zero-length array, no drop glue, and std's `const`-init fast path is preserved. Verified the test bites: neutering `SlotGuard::drop` fails `teardown_unpublishes_a_dropping_value`.

### Both gates verified

- **Structural**: planting a raw `thread_local!` fails immediately, and the message names the fix and the allowlist. Ratcheting in *both* directions is right — a stale entry is one nobody has to justify any more. `--self-test`: *"the checker can fail in all four directions."*
- **Outcome**: `--self-test`: *"the checker rejects all 7 failure modes."* Its design against vacuity is the important half — profiling `churn_alloc` would pass forever while the real cost grew, because churn's thread-locals are exactly the covered ones. Requiring `direct_tsd=1` and a `claimed` floor no allocation microbenchmark clears is what stops "a gate green because its subject never ran".

The sabotage failing on the **budget** while the liveness assertions stayed satisfied (`direct_tsd=1`, `claimed=111`/`102`) is the distinction that matters — it failed for the right reason, not because the run broke.

And the unplanned demonstration is the best evidence of all: the ratchet failed the build on a raw `thread_local!` that landed in `gc/policy.rs` while this branch was being rebased — one of *my* merges. Recording it as cold rather than converting a file this PR doesn't own is the right call.

### The one thing I changed

`converted_declarations_survive_thread_turnover` failed **6/6 under load** while passing 3/3 in isolation and single-threaded. Cause: `claimed_slots()` is process-global, so any other test's first touch of a converted declaration lands a claim inside the measurement window, and the assertion was exact equality (`left: 138, right: 139`).

Replaced with a tolerance of 8, documented. It still separates the outcomes by two orders of magnitude — per-thread claiming, the bug it exists to catch, would add 5 declarations × 64 workers = **320**, not 1. 0/4 failures after, versus 6/6 before.

`churn_alloc` as the control — the benchmark that must *not* move because it was already covered — not moving is what makes the asyncpipe/interp deltas believable. `cargo test -p perry-runtime --lib`: 2023 passed. Gates 21/21.

Issue #7533, opened 2026-08-06
perf: object_deep_clone is 37.5x bun (657ms vs 17.5ms) — the worst row in the public artifact, newly visible now that it runs


proggeramlug:
## The worst row in the public artifact

App-pattern ratios from the public baseline at v0.5.1299 (pinned quiet mini,
AC power, CPU-quiet gate passed, node 22.23.1 / bun 1.3.14, 11 runs/cell):

| kernel | perry | bun | node | perry/bun |
|---|--:|--:|--:|--:|
| **object_deep_clone** | **657.0 ms** | **17.5 ms** | 56.9 ms | **37.54×** |
| promise_all_chains | 259.7 | 22.7 | 64.0 | 11.44× |
| json_parse_1mb | 438.2 | 68.1 | 127.1 | 6.43× |
| map_1m | 1233.7 | 256.5 | 320.1 | 4.81× |
| batch | 127.8 | 26.5 | 74.8 | 4.82× |
| json_stringify_1mb | 97.3 | 38.5 | 95.1 | 2.53× |
| buffer_transcode | 58.2 | 43.9 | 85.8 | 1.33× |
| date_format_parse | 36.0 | 44.8 | 116.3 | **0.80× (win)** |

`object_deep_clone` is **37.5× bun and 11.5× node** — far the worst cell, and
also the worst absolute gap (657 ms against 17.5 ms).

## Why it is new to the backlog

It was not slow before; it was **broken**. Both this kernel and
`promise_all_chains` failed outright until today (#7475 → fixed by #7495,
#7497 → fixed by #7516/#7529), so neither has ever appeared in a performance
sweep. The engine plan's backlog table — which lists `json_parse_1mb` 6.27×
as the worst row — predates their working at all.

So these two rows are not regressions from the fixes; they are the first
measurement of code that previously crashed. But that leads directly to the
first thing to check.

## First hypothesis to test, and it is a self-suspicion

The kernel is deep clone: spread- and iteration-heavy. That is **exactly** the
code the same fixes rooted, and rooting means re-reading through a handle at
every use:

- #7495 rooted five live values in `js_iterator_to_array`'s drain loop — a loop
  bounded at 100 000 iterations, now re-reading each per turn.
- #7527 replaced a borrowed `&str` key with `HeapKeyBytes`, a 64-byte copy per
  property-name read on the prototype-walk tower.
- #7516 rooted 14 sites including `js_array_clone` (every `[...arr]`/`Array.from`).

[...]

PR #7759, opened 2026-08-10
fix(array): a replaced Array.prototype[Symbol.iterator] drives spread, call-spread and Array.from (#7542)

### LINKED ISSUE #7542: spread: a replaced `Array.prototype[Symbol.iterator]` is ignored by `[...arr]`
Replacing `Array.prototype[Symbol.iterator]` changes what `[...arr]` yields in node. In Perry the spread ignores the replacement and returns the builtin element copy.

```ts
const arrProto: any = Array.prototype;
const original = arrProto[Symbol.iterator];
arrProto[Symbol.iterator] = function* () { yield "patched"; };

console.log([...[1, 2, 3]]);   // node: [ 'patched' ]   perry: [ 1, 2, 3 ]

arrProto[Symbol.iterator] = original;
console.log([...[1, 2, 3]]);   // node: [ 1, 2, 3 ]     perry: [ 1, 2, 3 ]
```

An **own** `[Symbol.iterator]` on the instance is honoured correctly, so this is specific to the prototype-level replacement:

```ts
const a: any = [1, 2, 3];
a[Symbol.iterator] = function* () { yield "own"; };
console.log([...a]);           // node: [ 'own' ]       perry: [ 'own' ]  ✓
```

## Where

The machinery already exists and is not consulted on this path. `array/indexing.rs` keeps a sticky flag:

```rust
static ARRAY_PROTO_ITERATOR_MODIFIED: AtomicBool = AtomicBool::new(false);
pub(crate) fn note_array_proto_iterator_write(obj: usize, sym_key: usize) { … }
pub(crate) fn array_proto_iterator_modified() -> bool { … }
```

`js_get_iterator` (`symbol/iterator.rs`) honours it — when the flag is set it reads the patched method off `Array.prototype` and calls it with `this === val`. `array_from_spread_value` never asks. Its generic symbol lookup reads **own** symbol props only (hence the own-`@@iterator` case working), so for a plain array it misses and falls to the `js_array_is_array` arm, which drives the builtin `array_values_iter` unconditionally.

Two things to check when fixing:

1. Whether `note_array_proto_iterator_write` actually fires for `Array.prototype[Symbol.iterator] = fn` — if the flag never flips, both this path and `js_get_iterator`'s are dead.
2. `%ArrayIteratorPrototype%.next` replacement is a separate, currently-unmodelled case: builtin array iterators dispatch `.next()` through the class-id tower, so a patched prototype `next` is bypassed there too.

## Provenance

Found while building a 32-case `[...arr]` semantics matrix for #7533. **Pre-existing**: reproduces byte-identically at `f06270d06` and is unchanged by #7540's dense fast path, which gates on `array_proto_iterator_modified()` and so cannot make this worse — with the flag set it declines and falls through to the same (already wrong) protocol path.

proggeramlug:
Fixes #7542.

## Two corrections to the diagnosis, both of which changed the fix

**1. `note_array_proto_iterator_write` does fire** (the issue's open check #1). Instrumented: 49 calls during a run, exactly one matches and sets the flag.

```
[7542] note_write obj=0x34814a552b8 proto=0x34814a552b8 sym=0x34814b40180 wk=0x34814b40180 already=false
[7542] FLAG SET
```

So the flag was never the problem, and neither `js_get_iterator`'s path nor this one is dead.

**2. The walk does NOT miss for a plain array.** The issue says the generic lookup "reads **own** symbol props only … so for a plain array it misses and falls to the `js_array_is_array` arm". It doesn't: `js_object_get_symbol_property` **synthesizes the built-in** `Symbol.iterator` for an array receiver (a `js_class_method_bind` by name) rather than reading the prototype slot — so the walk resolves the builtin, calls it, and returns the element copy.

I know this because I first implemented the fix exactly where the issue points, at the `js_array_is_array` arm, and it changed nothing. Instrumenting that arm showed it is **never reached** with the flag set. A guard there is dead code.

## The fix

The guard goes **before** the walk, and at each entry point that reaches an array by a different route — this is why a single-site fix was never going to work:

| entry point | reached by | was doing |
|---|---|---|
| `array_from_spread_value` | `[...arr]` | walk resolves the builtin |
| `js_array_from_value` | `Array.from(arr)` | array arm ends in a raw `js_array_clone` |
| `js_array_push_spread_f64` | `[0, ...arr, 9]` | direct element copy |
| `js_array_like_to_array` | `f(...arr)` | returns the raw pointer ("fast path, no protocol overhead") |

Each delegates to `js_get_iterator` — the one implementation that consults the patched prototype per GetIterator (read the method off `Array.prototype`, call it with `this === val`, TypeError when deleted or non-callable). Delegating rather than restating that sequence four times is the point: two copies of a spec sequence that must agree is how this diverged in the first place.

An **own** `arr[Symbol.iterator]` still wins — the shortcut stands aside when the receiver carries its own method, because `js_get_iterator`'s patched branch reads the prototype only and would otherwise throw "not iterable" for an array with an own method once the prototype slot is deleted. (That case caught me: my first cut broke it.)

`array_proto_iterator_modified()` is sticky-false until user code writes the prototype slot, and #7533's dense fast path already declines when it is set, so ordinary programs are untouched.

## Verification

```
                       node --experimental-strip-types    perry
spread literal:        ["patched"]                        ["patched"]
spread variable:       ["patched"]                        ["patched"]
spread into array:     [0,"patched",9]                    [0,"patched",9]
call spread:           1                                  1
Array.from:            ["patched"]                        ["patched"]
own wins:              ["own"]                            ["own"]
```

Byte-identical. Before this change the four spread rows were `[1,2,3]` / `[0,1,2,3,9]` / `3` / `[1,2,3]`.

- `cargo test -p perry-runtime --lib`: 1991 passed, 0 failed.
- Targeted parity over the array / iterator / spread gap tests: all pass.

## No gap test, deliberately

Any test that patches `Array.prototype[Symbol.iterator]` **takes the oracle down** under the parity harness: node builds a `SafeMap` from an iterable, gets the patched value, and exits 1 with

```
TypeError: Iterator value patched is not an entry object
    at new Map (<anonymous>)
    node:internal/per_context/primordials:449
```

so the comparison runs against a crashed reference. (Standalone `node file.ts` is fine, which is why this only showed up once I ran it through the harness rather than by hand.) I tried four ways around it — restore by saved reference, restore via `getOwnPropertyDescriptor`, a behavioural replacement, a self-iterable replacement, and `process.exit(0)` — and none works, for reasons that are themselves bugs (below). So the fix is verified by direct comparison instead, and I would rather say that plainly than land a test that passes for the wrong reason.

## Filed separately (found here, out of scope)

1. **`for…of` over a typed local ignores the patched method.** It is a codegen dense index loop with *no runtime call at all* (`js_array_values_iter_obj` / `js_array_length` appear only in the `declare` block) — a different mechanism from the four above, needing its own guard.
2. **Restoring the original iterator is broken.** `arrProto[Symbol.iterator] = original` then `[...arr]` throws `next is not a function`: reading `Array.prototype[Symbol.iterator]` yields a method bound to the prototype. Pre-existing and independent of this change — it reproduces on `for…of` over an array literal, which routes through `js_get_iterator` and is untouched here.
3. **`Object.getOwnPropertyDescriptor(Array.prototype, Symbol.iterator)` returns `undefined`.**

## Summary by CodeRabbit

* **Bug Fixes**
  * Array spread operations now respect customized `Array.prototype[Symbol.iterator]` behavior.
  * Updated behavior applies to `Array.from`, array spreading, spreading into function calls, array appending, and related conversions.
  * Arrays with their own iterator continue to use that iterator appropriately.
  * Existing fast paths remain available when the standard iterator is unchanged.
  * Improved consistency when custom iterators alter the values or order produced by array operations.


proggeramlug:
## Merging as v0.5.1445

A/B'd on my own host — `new` matches node on all six rows, `old` is wrong on four:

```
                   node          old            new
spread literal     ["patched"]   [1,2,3]        ["patched"]
spread variable    ["patched"]   [1,2,3]        ["patched"]
spread into array  [0,"p",9]     [0,1,2,3,9]    [0,"patched",9]
call spread        1             3              1
Array.from         ["patched"]   [1,2,3]        ["patched"]
own wins           ["own"]       ["own"]        ["own"]
```

### The two corrections are worth more than the fix

**You implemented the issue's suggested fix, measured it, and found the arm is never reached.** *"I first implemented the fix exactly where the issue points, at the `js_array_is_array` arm, and it changed nothing. Instrumenting that arm showed it is never reached with the flag set. A guard there is dead code."* Both of the issue's open checks were wrong in the same direction — the flag *does* fire (49 calls, one match), and the walk *doesn't* miss, because `js_object_get_symbol_property` synthesizes the builtin rather than reading the prototype slot.

That is the difference between fixing the reported location and fixing the bug, and it is why a single-site fix was never going to work: four entry points reach an array by different routes.

**Delegating all four to `js_get_iterator` rather than restating the sequence** is the right structure, and the reason given is the one that matters — two copies of a spec sequence that must agree is how this diverged in the first place.

The `own wins` row is the one I'd have broken: `js_get_iterator`'s patched branch reads the prototype only, so an array with its own method would throw "not iterable" once the prototype slot is deleted. Your first cut hit exactly that, and it is now the sixth row of the table.

### Declining to write a gap test is the right call

Any test that patches `Array.prototype[Symbol.iterator]` **takes the oracle down** — node's primordials build a `SafeMap` from an iterable and exit 1 with *"Iterator value patched is not an entry object"*, so the harness compares against a crashed reference. Four workarounds tried, none viable, and each failure is itself a bug you filed. Verifying by direct comparison and **saying so plainly** beats landing a test that passes for the wrong reason — which is exactly the failure mode the rest of this repo's gates exist to prevent.

The detail that it only appeared through the harness, not by hand (`node file.ts` standalone is fine), is worth keeping: it means the trap is invisible to the obvious local check.

### Three follow-ups filed rather than folded in

`for…of` over a typed local ignoring the patched method (a codegen dense index loop with **no runtime call at all** — a genuinely different mechanism), restoring the original iterator being broken, and `getOwnPropertyDescriptor` returning `undefined`. That last one is #7761's subject, which is stacked on this and next.

`cargo test -p perry-runtime --lib`: 2023 passed, 0 failed. Gates 21/21.

PR #7761, opened 2026-08-10
fix(array,hir): a replaced Array.prototype[Symbol.iterator] is honoured everywhere, and the method is a real own property (#7760)

### LINKED ISSUE #7760: Array iteration protocol: for-of over a typed local bypasses a patched Array.prototype[Symbol.iterator]; restoring the original throws; getOwnPropertyDescriptor on a prototype symbol returns undefined
Three defects found while fixing #7542 (PR #7759), each independent of it and of each other. Filing together because they share a root area and the first two block writing a gap test for any of them.

## 1. `for…of` over a typed local ignores a replaced `Array.prototype[Symbol.iterator]`

```ts
const p: any = Array.prototype;
p[Symbol.iterator] = function* () { yield "patched"; };
const src = [1, 2, 3];
const out: any[] = []; for (const x of src) out.push(x);
console.log(out);        // node: ["patched"]   perry: [1,2,3]

const lit: any[] = []; for (const x of [1, 2, 3]) lit.push(x);
console.log(lit);        // node: ["patched"]   perry: ["patched"]  ✓
```

Over an array **literal** it is correct (that routes through `js_get_iterator`, which honours the flag). Over a **typed local** it is not, and the reason is that codegen emits a dense index loop with *no runtime call at all* — `js_array_values_iter_obj` and `js_array_length` appear only in the `declare` block of the emitted IR.

So this cannot be fixed where #7542 was: there is no runtime entry point to guard. It needs a codegen-side gate — e.g. the loop preheader loading a volatile global (the shape `PERRY_ARRAY_INDEX_FAST_PATH_INVALIDATED` already uses) and branching to the protocol path when the iteration protocol has been patched.

## 2. Restoring the original array iterator throws

```ts
const p: any = Array.prototype;
const orig = p[Symbol.iterator];
p[Symbol.iterator] = function* () { yield "x"; };
p[Symbol.iterator] = orig;                        // restore
for (const v of [1, 2, 3]) {}                     // TypeError: next is not a function
```

Reading `Array.prototype[Symbol.iterator]` returns a method **bound to the prototype** (`js_object_get_symbol_property`'s array arm synthesizes it via `js_class_method_bind`), so storing it back and calling it with `this === arr` does not produce an array iterator.

Pre-existing and independent of #7542: it reproduces on `for…of` over an array *literal*, which routes through `js_get_iterator` and is untouched by that fix.

## 3. `getOwnPropertyDescriptor` on a prototype symbol returns `undefined`

```ts
Object.getOwnPropertyDescriptor(Array.prototype, Symbol.iterator)
// node: { value: [Function: values], writable: true, enumerable: false, configurable: true }
// perry: undefined
```

Which also removes the obvious workaround for (2).

## Why these are worth having together

They make the whole family untestable in the gap suite today. Any test that patches `Array.prototype[Symbol.iterator]` takes the ORACLE down under `run_parity_tests.sh` — node constructs a `SafeMap` from an iterable and gets the patched value:

```
TypeError: Iterator value patched is not an entry object
    at new Map (<anonymous>)  node:internal/per_context/primordials:449
```

so the comparison runs against a crashed node. The natural fix is for the test to restore the original before returning — which (2) and (3) both block. Fixing either one makes a gap test for this family possible; until then #7759's coverage is a direct byte-for-byte comparison recorded in its description.

proggeramlug:
Fixes #7760 — **all three items**. (Item 1 was added after the first review pass; the original scope note is gone.) #7759 has merged, so this now applies to `main` directly.

## Items 2 and 3: one root cause

`Array.prototype[Symbol.iterator]` was **not an own property at all** — `js_object_get_symbol_property` synthesized a receiver-bound method on every read — and `Array.prototype.values` was installed by `install_noop_proto_methods`, so it existed and did nothing. Array-specific; Map/Set/String/`%TypedArray%` already had real descriptors.

That explains both symptoms. No own property ⇒ no descriptor, `hasOwnProperty` false, absent from `getOwnPropertySymbols` (item 3). And the synthesized closure is bound to the *prototype* at read time, so storing it back and calling it with `this === arr` throws `next is not a function` (item 2).

One real `array_prototype_values_thunk` reading `this` at CALL time, installed as `values` and as the own `[Symbol.iterator]` with the spec descriptor `{ writable: true, enumerable: false, configurable: true }`.

## Item 1: `for…of` over an array

`for…of` desugars to an index loop that never consults the protocol. **Two parallel lowerings** had to be fixed — `lower::stmt_loops` (module init) and `lower_decl::body_stmt` (function bodies) — which is why a `for…of` over an array *parameter* was still wrong after the first half worked, and is worth knowing about for anything else in this area.

The patch is a RUNTIME fact and the index-vs-lazy choice is COMPILE-TIME, so both forms are emitted and selected by a branch on a new `Expr::ArrayIterationPatched`, lowering to a single volatile `i8` load of `PERRY_ARRAY_PROTO_ITERATOR_PATCHED` — the shape `PERRY_ARRAY_INDEX_FAST_PATH_INVALIDATED` already uses. Three properties shaped it:

- **Lazy.** Materializing the protocol at loop entry — what spread does — is eager where node is lazy: an early `break` would over-pull, a side-effecting iterator would run extra steps, an infinite one would hang. The test pins `break` after 2 of 3 pulling exactly 2, matching node.
- **The index loop is emitted byte-identically**, inside the `else`. The smaller alternative (one loop, loop-invariant branch, element through a shared `__item`) would have broken the HIR pattern matchers behind the element-shape clone (#7612), the dense spread path (#7533), and the packed-f64 / i32-counter loop specializations.
- **The check is at loop ENTRY.** Which is what the spec wants anyway: `for…of` performs GetIterator exactly once, so a patch landing mid-loop must not change the iterator already in hand.

### Cost

Interleaved, 7 reps each, 20k-iteration loop fixture (three `for…of` loops over `number[]` / object / string arrays), on a host at load average 12:

```
median   1.85 s (guard off)   1.86 s (guard on)
best     1.85 s               1.82 s
```

No measurable difference. Recording a near-miss because it is the more useful part: a first **non-interleaved** best-of-5 on the same box read 2.64 s vs 2.90 s and I nearly reported a 10% regression from it. On a machine at this load only interleaved runs mean anything, and the number still wants confirming on the pinned bench host before anyone quotes it as a bound.

## Validation

- `test-files/test_gap_array_proto_iterator_replaced_7542.ts` — **PASS through the harness**, byte-identical to node 26.5.1. Covers the descriptor shape, `values.call`, all four spread forms from #7542, own-`@@iterator` precedence, `for…of` over a module const / typed local / **array parameter**, the lazy-`break` pull count, restore-by-reference, and a `getOwnPropertyDescriptor` → `defineProperty` round-trip.
- `cargo test -p perry-runtime --lib`: 1991 passed. `cargo test -p perry-hir --lib`: 290 passed.
- Targeted parity across array / iterator / for / spread / generator / map / destructuring gap tests: all pass.
- `cargo fmt --all --check` clean; `scripts/check_file_size.sh` clean — the guard wrapper moved to `lower/for_of_guard.rs` because `stmt_loops.rs` crossed the 2000-line cap at 2025.

## Note on the test

This test could not exist before this PR. Patching `Array.prototype[Symbol.iterator]` takes the *oracle* down — node's primordials build a `SafeMap` from an iterable and get the patched value — so the test must restore the slot, and restoring is exactly what item 2 broke.

No version bump (the branch carries the maintainer's 0.5.1446 bump from the earlier push).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `Array.prototype[Symbol.iterator]` and `.values` to behave like standard configurable, writable, non-enumerable properties.
  * Improved borrowed and direct calls, spreading, `Array.from`, and `for…of` behavior when iterators are replaced.
  * Preserved lazy iteration and correct early termination.
  * Restoring or overriding array iterators now works correctly.

* **Tests**
  * Added regression coverage for iterator replacement and descriptor round-tripping.

* **Chores**
  * Updated the package version to 0.5.1446.


proggeramlug:
## Merging as v0.5.1446

A/B'd against a post-#7759 build, so this isolates *this* PR:

```
                node        after #7759 only    this PR
desc            function    undefined           function
hasOwn          true        false               true
inSymbols       true        false               true
name            values      undefined           values
restore         [1,2,3]     TypeError:          [1,2,3]
                            next is not a fn
```

### The best thing here is that it makes the previous PR testable

#7759 shipped without a gap test, and the reason was *this* bug: a test that patches `Array.prototype[Symbol.iterator]` must restore the slot before returning or it takes node's own primordials down — and **restoring was exactly what was broken**. So the untestable thing became testable by fixing the actual defect rather than by working around it. Four workarounds were tried on #7759 and each failure was itself a symptom of this.

I verified that end-to-end rather than taking it: `test_gap_array_proto_iterator_replaced_7542.ts` is **byte-identical to node 26.5.1**, and `node` itself now **exits 0** on it — the crash that made the comparison meaningless is gone. One test now covers both issues, through the harness.

### One root cause, and the cross-builtin table proves it

`Array.prototype[Symbol.iterator]` was **not an own property at all** — synthesized per read, with `values` installed by `install_noop_proto_methods`, i.e. present and doing nothing. The table showing `Map`/`Set`/`String`/`%TypedArray%` all with `hasOwn: true` and Array alone with `false` is what turns "a bug" into "Array was missed": the mechanism (`install_collection_iterator_symbol`) already existed.

Both symptoms then fall out of the one cause rather than needing separate fixes — item 3 because there is no own property to describe, and item 2 because a closure bound to the prototype at *read* time iterates the prototype when called with `this === arr`, which is exactly `next is not a function`.

Reading `this` at **call** time in one real `array_prototype_values_thunk`, installed with the spec descriptor `{ writable: true, enumerable: false, configurable: true }`, is the correct shape — and the `defineProperty` round-trip row is the one that proves the descriptor is real rather than merely present.

### Scope

Correctly **not** closing #7760 — item 1 stays open, and saying so is right. Same discipline as #7759's three filed follow-ups.

`cargo test -p perry-runtime --lib`: 2023 passed, 0 failed. 38 targeted parity tests across array / iterator / symbol / object / map / set / typed-array. Gates 21/21.

Issue #7757, opened 2026-08-10
Generic-class specializations are distinct constructor objects: a.constructor !== Gen, and two specializations compare unequal


proggeramlug:
Two specializations of one generic class are distinct constructor objects, and an instance's `.constructor` is not the class the user wrote.

## Repro

```ts
class Gen<T> { v: T | undefined; }
const a = new Gen<number>();
const b = new Gen<string>();
const c = new Gen();                       // never specialized
console.log(a.constructor === Gen, b.constructor === Gen, c.constructor === Gen);
console.log(a.constructor === b.constructor);
console.log(Object.getPrototypeOf(a) === Gen.prototype,
            Object.getPrototypeOf(a) === Object.getPrototypeOf(b));
```

```
$ node --experimental-strip-types t.ts
true true true
true
true true

$ ./t
false false true
false
true false
```

## Why

`new Gen<number>()` is monomorphized into a separate class `Gen$num` with its own class id (`monomorph::mangle::generate_specialized_name`), and the instance carries that id. TypeScript erases type arguments, so at runtime there is exactly one `Gen` — the specializations are an implementation detail that is leaking through every id-keyed identity surface.

This is the third face of the same leak:

- `instanceof` — fixed in #7575 / PR #7631, by recording `Class::specialized_from` and teaching the chain walk a specialization→generic edge.
- `constructor.name` — fixed in #7632 / PR #7756, by registering the origin's display name against the specialization's class id.
- **constructor / prototype IDENTITY** — this issue. Not addressed by either, because both fixed a *lookup* (a chain walk, a name table) rather than the object identity itself.

Worth noting that #7756 makes this one harder to see, not easier: after it, `a.constructor.name === b.constructor.name === "Gen"` while `a.constructor !== b.constructor`. Two things that print identically and compare unequal is a worse failure mode for a user to debug than an obviously-wrong name, so this is the natural follow-up to schedule rather than leave implicit.

## Already inconsistent today

`Object.getPrototypeOf(a) === Gen.prototype` is **`true`** while `a.constructor === Gen` is **`false`** — so the prototype edge and the constructor edge already disagree with each other about whether the specialization is `Gen`. Whatever the fix is, that pair should come out consistent.

## Shape of a fix (unverified)

The specialization exists to monomorphize *method bodies*; nothing about that requires a distinct constructor object or prototype. Candidates, in rough order of blast radius:

[...]

PR #7762, opened 2026-08-10
fix(runtime): generic specializations share the generic's prototype object (#7757, prototype half)

proggeramlug:
Partial fix for #7757 — the **prototype** identity. The issue stays open for the `.constructor` edge; reasoning and what I ruled out at the bottom.

## What was wrong

Each monomorphized specialization materialized its **own** prototype object, so:

```ts
class Gen<T> { v: T | undefined; }
const a = new Gen<number>(), b = new Gen<string>(), c = new Gen();
Object.getPrototypeOf(a) === Gen.prototype              // node true   perry true
Object.getPrototypeOf(a) === Object.getPrototypeOf(b)   // node true   perry FALSE
Object.getPrototypeOf(a) === Object.getPrototypeOf(c)   // node true   perry FALSE
```

Those first two cannot both hold, and that contradiction is the actual bug: the answer depended on which class id you asked through. TypeScript erases type arguments, so at runtime there is exactly one `Gen` and one `Gen.prototype`; the specializations are an implementation detail of `monomorph::mangle::generate_specialized_name`.

## Fix

Both prototype registries resolve a specialization through `class_generic_origin` — the same edge `instanceof` uses (#7575) and the display name uses (#7632), now applied to the prototype surface.

It had to be **both**: `CLASS_DECL_PROTOTYPE_OBJECTS` and `CLASS_PROTOTYPE_OBJECTS` are the two paths CLAUDE.md's "known-weak areas" flags as having disagreed about the same chain before, and redirecting only the decl one left `getPrototypeOf` and the lookup chain answering differently.

Method dispatch is untouched — it runs off the per-class-id vtable, so each specialization keeps its own monomorphized bodies.

## Validation

- `test-files/test_gap_generic_specialization_prototype_identity_7757.ts` — PASS through the harness, byte-identical to node 26.5.1. Covers two specializations, the unspecialized instance, and a nested `class Wrap<T> extends Gen<T>` chain, plus the #7575 / #7632 halves still holding.
- `cargo test -p perry-runtime --lib`: 2023 passed, 0 failed.
- Targeted parity across class / generic / instanceof / prototype / class-expression gap tests (35 tests): all pass.

## Why `.constructor` is not in here

`a.constructor === Gen` is still `false`. It does **not** resolve through either prototype registry, so this change cannot reach it.

I applied the same redirect to `instance_constructor_value` — the obvious site, and the one `get_field_by_name_tail` calls for exactly this key — and it changed nothing. Two findings worth having on the record so the next attempt doesn't repeat them:

1. **`object_static_prototype()` appears to materialize on first call.** The tail guards the `instance_constructor_value` call on `object_static_prototype(obj).is_none()`, and adding a probe that *reads* it flips the branch under test. I burned two instrumented builds on artifacts from this before spotting it; anything measuring this path needs to avoid touching that accessor.
2. **`class Empty {}` gives `e.constructor === undefined`** (node: `Empty`). That is not a generics bug at all, which suggests `.constructor` resolution has a broader gap that monomorphization merely exposes — and that the remaining half of #7757 may be better framed as that gap rather than as a specialization leak.

I stopped rather than keep probing blind, and reverted the redirect that provably never runs — leaving a dead guard that looks like a fix is the pattern I flagged on #7542.

No version bump.

## Summary by CodeRabbit

* **Bug Fixes**
  * Generic class specializations now share prototype identity with their originating generic class.
  * Prototype comparisons, inheritance checks, and `instanceof` behavior are now consistent across specialized and unspecialized instances.
  * Method dispatch remains specific to each specialization.

* **Tests**
  * Added regression coverage for prototype identity, inheritance chains, `instanceof`, and constructor behavior.


proggeramlug:
## Merging as v0.5.1448

A/B'd on my host — `new` matches node on every row:

```
                  node    old      new
a===Gen.proto     true    true     true
a===b             true    FALSE    true
a===c             true    FALSE    true
instanceof        true    true     true
dispatch          gen wrap  gen wrap  gen wrap
```

**The contradiction framing is the right diagnosis.** `getPrototypeOf(a) === Gen.prototype` being `true` while `getPrototypeOf(a) === getPrototypeOf(b)` was `false` cannot both hold — the answer depended on which class id you asked through. That is a sharper statement of the bug than "specializations have separate prototypes", and it is what makes the fix obviously correct rather than a preference.

**Both registries was necessary, not belt-and-braces.** `CLASS_DECL_PROTOTYPE_OBJECTS` and `CLASS_PROTOTYPE_OBJECTS` are exactly the pair CLAUDE.md's known-weak-areas section flags as having disagreed about the same chain before, and redirecting only the decl one would have left `getPrototypeOf` and the lookup chain answering differently — i.e. a new instance of the same class of bug.

**The non-regression I most wanted to see is in the table**: `dispatch: gen wrap` in *both* arms. Method dispatch runs off the per-class-id vtable, so each specialization keeps its own monomorphized bodies. Sharing the prototype object without sharing the vtable is the whole trick.

### The two findings on `.constructor` are worth more than the fix

1. **`object_static_prototype()` appears to materialize on first call**, so a probe that merely *reads* it flips the branch under test. Two instrumented builds lost to artifacts before spotting it. That is a measurement hazard anyone touching `get_field_by_name_tail` will hit, and it is now written down.
2. **`class Empty {}` gives `e.constructor === undefined`** (node: `Empty`). That is not a generics bug at all — which reframes the remaining half of #7757 from "a specialization leak" to "`.constructor` resolution has a broader gap that monomorphization merely exposes". Filing the residue under the right framing is worth more than a partial fix under the wrong one.

Applying the redirect to `instance_constructor_value` — the obvious site — and reporting that **it changed nothing** is the useful negative result: it rules out the whole prototype-registry route for that half.

Correctly leaves #7757 open. Gates 21/21, `cargo test -p perry-runtime --lib` 2023 passed.

PR #7765, opened 2026-08-10
perf(array): answer "is this receiver a Map/Set?" from the object's own GC tag (#7765)

proggeramlug:
## The finding

`gc-handoff/apps/asyncpipe.ts` — an async service pipeline, and at 13x node the
worst gap in the corpus — spent **13.5% of its run in
`set::is_registered_set` + `map::is_registered_map`**. Not on Set or Map work:
on `js_array_get_f64` and `js_array_length` asking both collection registries
whether an ordinary array was secretly a collection, on every element read.
Caller attribution on a symbolicated `sample` of `asyncpipe_big.ts`:

| probe | share of its own samples from |
|---|---|
| `is_registered_set` (7.48%) | `js_array_get_f64` 76%, `js_array_length` 11% |
| `is_registered_map` (6.03%) | `js_array_get_f64` 82%, `js_array_length` 16% |

#7755 made *unused*-feature registry probes free with monotone latches and named
Map/Set as the deliberate residual — asyncpipe uses both, so the #7474 latch is
correctly armed and each probe is real work — and named why the typed-array
trick does not transfer: an address-keyed negative memo is an ABA hazard for
Map/Set, whose headers are recyclable arena objects.

## The mechanism

**The object already knows.** `js_map_alloc` / `js_set_alloc` allocate their
headers through `arena_alloc_gc(_, _, GC_TYPE_MAP | GC_TYPE_SET)`, and each is
the *single* registration site for its registry — so a registered collection's
address **is** its GC header, and `obj_type` answers in one byte. Both hot call
sites gate their probes on it; the registry stays authoritative for the positive
answer, so nothing about when an entry is added or swept moves.

ABA-proof **by construction rather than by bookkeeping**: the tag lives INSIDE
the candidate bytes, so whatever allocation owns them next stamps its own
`obj_type` before the pointer is handed out. A recycled address answers for its
new owner with no invalidation step to get wrong.

Correct for a header-*less* receiver too — buffers and typed arrays are
`std::alloc`-backed, so the bytes below them are allocator bookkeeping. Both are
already routed by the (latched, free) probes above, and either way those bytes
read the outcome is unchanged: a byte that happens to read as
`GC_TYPE_SET`/`GC_TYPE_MAP` still falls through to the authoritative registry,
and any other value skips a probe that would have answered `false` anyway.
Neither call site gains a dereference — both already read this header.

The second commit takes the adjacent cluster the same way: the object field-get
funnel proves `keys` is a live `GC_TYPE_ARRAY` and caps the index below its
capacity, then called `js_array_get` per key, which re-establishes both facts
from scratch. `keys_array_slot` serves the dense case from the array's own words
and delegates everything else.

## Tests

`crates/perry-runtime/src/array/collection_tag_tests.rs` asserts THE SUBJECT,
not just the answer — the registry is a correct fallback, so a test that only
compared values would still pass with the gates deleted (CLAUDE.md, "four ways a
gate can be unable to fail", case 4). Both sabotage checks were run:

| remove | test that goes red |
|---|---|
| either receiver-tag gate | `plain_array_element_reads_never_probe_the_collection_registries` — 64 passes over a 4-element array moved the probe counters from 0 to 320 |
| the header confirmation in `is_registered_map` | `a_stale_registry_entry_over_recycled_bytes_does_not_read_as_a_map` — the stale entry then reports `true` and the read serves `MapHeader::entries` |

## Numbers — absolute seconds, rebased onto `c2a96b638`

Both arms **rebuilt from the new merge-base** and re-measured; nothing is carried
forward from the pre-rebase tree. Quiet pinned M1 mini, static arm64 binaries.

**Measured interleaved** — 4 passes x best-of-5, alternating arms per benchmark
per pass. That matters: a block-sequential run (all of one arm, then the other)
gave `interp` a fake +7.3% and a uniform fake ~+3% on eight benchmarks, because
the mini oscillates ~3% between passes and the two blocks landed in opposite
phases. The per-pass series made it obvious (pass 1 and 3 slow on *both* arms,
pass 2 and 4 fast on both). Every number below is min-of-all-passes, and the
baseline arm independently reproduces #7767's `interp` 1.89, which confirms the
baseline really is the new main.

| bench | base `c2a96b638` | + this PR | delta |
|---|---|---|---|
| **asyncpipe** | **0.9211** | **0.7284** | **-20.9%** |
| interp | 1.8945 | 1.8873 | -0.4% |
| shapes | 0.2878 | 0.2742 | -4.7% |
| churn | 0.4212 | 0.4207 | -0.1% |
| churn_alloc | 0.3765 | 0.3765 | +0.0% |
| churn_read | 0.0221 | 0.0221 | -0.3% |
| push_cls | 0.3561 | 0.3560 | -0.0% |
| push_num | 0.1425 | 0.1431 | +0.4% |
| cycles | 0.1928 | 0.1926 | -0.1% |
| deeplist | 0.2454 | 0.2451 | -0.1% |
| tree | 1.6344 | 1.6344 | +0.0% |
| tree_wide | 2.1108 | 2.1099 | -0.0% |
| retain | 0.5355 | 0.5354 | -0.0% |
| retain_wide | 1.0905 | 1.0887 | -0.2% |
| fib40 | 0.3933 | 0.3931 | -0.0% |
| iso_miss | 2.3668 | 2.3558 | -0.5% |

The **-20.9% on `asyncpipe` survives on top of #7767** (it was -21.1% on the old
base), and #7767's `interp` win is intact — the two are independent, as expected
from one being codegen and the other runtime.

**The `churn_alloc` +1.2% I flagged pre-rebase is gone: +0.0% here** (0.3765 vs
0.3765, stable across all four passes). It was code layout on the old base, not
an executed path — which matches the analysis at the time, since `churn_alloc`
runs none of the code this PR changes.

Outputs verified byte-identical to `node --experimental-strip-types` on both
arms before timing. GC canaries clean on both arms: `iso_miss` prints
`checksum 437840 misses 0` plain, under `PERRY_GC_PROTECT_FROMSPACE=1
PERRY_GC_PROTECT_FROMSPACE_DEPTH=800`, and under `PERRY_GC_VERIFY_EVACUATION=1`.
`asyncpipe` still exits 138 under `PROTECT_FROMSPACE` on an `obj_type=5`
(Promise) object **identically on both arms** — pre-existing, untouched.

Profile of `asyncpipe_big.ts` (symbolicated `sample`, two runs, in agreement;
measured pre-rebase, the profile shape is unchanged by #7767):

| entry | before | after |
|---|---|---|
| `is_registered_set` + `is_registered_map` | **13.51%** | **1.18% / 1.41%** |
| `array_object_flags` | 3.06% | out of the top 20 |
| `js_array_get_f64` | 6.27% | out of the top 14 (replaced by `keys_array_slot`, 2.6-3.1%) |

`classify_heap_space_in_range` (4.4% before) was left alone: caller attribution
shows it is **100% GC-internal** — `CopyingNurseryCollector::mark_addr` 51%,
`rewrite_value_bits` 37% — and #7742 already added the `(space, base)` form
specifically to collapse the double lookup for addresses 8 bytes apart, which
`mark_addr` uses. Its *share* rises here only because the mutator side shrank.

Two comments claiming Map/Set headers are `alloc()`-backed with no `GcHeader` —
the stated reason the registries are consulted before any header read — are
corrected in place. They predate the move into the managed arena, and are how
the answer stayed hidden.

## Summary by CodeRabbit

* **Performance**
  * Improved array element and length access by reducing unnecessary collection checks for ordinary arrays.
  * Added a faster path for dense internal key arrays while preserving correct handling of sparse, forwarded, and out-of-range entries.
* **Bug Fixes**
  * Prevented stale collection metadata from being mistaken for recycled array memory.
  * Preserved correct behavior for specialized collection and descriptor-backed receivers.
* **Tests**
  * Added comprehensive coverage for collection detection, array growth, sparse arrays, invalid pointers, and fallback behavior.

**Gap suite**, re-run on the rebased tree (full 518, `PERRY_SKIP_BUILD=1`
against this build): 493 parity pass, 95.1%. The divergence set from
`test-parity/gap_snapshot.json` is **byte-identical to the pre-rebase run** — 21
entries, `diff` clean — so the rebase introduced nothing. All of it is
pre-existing and none is array/Map/Set/property-shaped: six http/fetch/net
crashes, `specabi_reassign`, `zlib_3285_params`, ten `node_fail -> parity_fail`
status changes that are oracle-side (missing `node_modules`,
`ERR_UNSUPPORTED_TYPESCRIPT_SYNTAX` — a runtime change cannot flip node's exit
code), and the two `gc_*_argument_rooting` tests, whose entire diff is the
`[gc-zeal] forced_collections=...` verdict line #7684 writes to stderr and the
harness merges into stdout. `test_gap_iterator_helpers_2874` goes
`parity_fail -> pass`.


proggeramlug:
Audit complete — merging.

**Verified independently of the PR body:**
- The load-bearing claim ("the object already knows"): `register_map`/`register_set` each have exactly ONE call site (map.rs:1069 / set.rs:843), fed by `arena_alloc_gc(..., GC_TYPE_MAP/GC_TYPE_SET)` — so a registered collection's address is its GC header, and the tag is authoritative for the positive case.
- `keys_array_slot`'s inline-elements assumption matches the canonical getter byte-for-byte (indexing.rs:406/796/824 compute the same `arr + sizeof(ArrayHeader)` base), and the `GC_FLAG_FORWARDED`/descriptor/hole arms all delegate.
- Both gates keep the registry authoritative for the positive answer — the tag only prunes, which is strictly safer than the pre-PR path (the stale-registry-over-recycled-bytes ABA case now answers false; pinned by `a_stale_registry_entry_over_recycled_bytes_does_not_read_as_a_map`).
- Sabotage re-run by me (verified applied before reading results): dropping the SET tag gate at the `js_array_length` site turns exactly `plain_array_element_reads_never_probe_the_collection_registries` red, 5/6 stay green.
- 6/6 new suite, full `perry-runtime --lib` 2,052 passed, e2e Set/Map/keys/for-in probe byte-identical to node 26.5.1, identical under `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1` (6 copying minors, exit 0).
- Rebased onto v0.5.1454; two deltas added during audit: `GC_STORE_AUDIT(POINTER_FREE)` markers on the test's two raw f64 stores (store-site inventory), version bump to 0.5.1455. The addr-class allowlist entry is test-file-only with a written reason — no production bare-address site.
- Local gate 21/21 green.

Closes #7768.

Issue #6082, opened 2026-07-06
[perf] Runtime helper declarations carry no LLVM attributes → -O3 cannot optimize across js_* calls


proggeramlug:
**Severity:** performance (program-wide; highest perf-per-effort item in the audit)
**Found by:** audit `fable-audit-perry-2.md`.

### What
Every `js_*` runtime-helper **declaration** is emitted with zero LLVM attributes, so the `-O3` pipeline must treat each call as fully opaque (reads/writes all memory, may unwind). Since nearly all non-arithmetic JS lowers to helper calls, `-O3` ends up optimizing only pure-numeric islands: no CSE of two `js_box_get` of the same box, no LICM of `js_string_length` out of a loop, no DSE across any helper call.

### Why
`declare_line_for` (`crates/perry-codegen/src/module.rs:63-76`) attaches attributes only to `setjmp` (`returns_twice`); the declaration API takes none, and a repo-wide grep finds no `readnone`/`readonly`/`memory(...)`/`nounwind`/`willreturn` anywhere. Modules do compile `-O3 -mcpu=native` (`linker.rs`), so the codegen is there — it just can't see through the calls.

### Fix
Annotate the declare lines after a one-time per-helper effect audit:
- `memory(none)` for pure NaN-box/type-check/tag helpers,
- `memory(argmem: read)` / `memory(argmem: readwrite)` where true,
- `nounwind` / `willreturn` where they hold.
Start with the safe read-only/type-check/nanbox families; allocation/GC-triggering helpers must stay writeful (and keep the GC-safepoint semantics). This is a textual/string-level change with program-wide effect.

**Confidence:** high.

proggeramlug:
Confirmed the mechanism and the soundness boundary on current main.

**Backend / mechanism:** codegen emits textual LLVM IR handed to `clang -O3`/`opt` (no inkwell). Function attributes are already emitted for one case — `setjmp`/`_setjmp` get `#0` (`returns_twice`) via `declare_line_for`/`declare_function` (`module.rs`), with attribute groups defined in `push_attrs_and_metadata` (`attributes #0 = { returns_twice }`, `#1 = { noinline optnone }`). So adding e.g. `attributes #2 = { nounwind willreturn memory(none) }` and appending ` #2` on selected decls is a mechanical extension — no new dependency/API. Decls funnel through `LlModule::declare_function` (~1.6k call sites across `runtime_decls/*`).

**Soundness boundary (sharper than "pure vs impure"):**
[...]

PR #7767, opened 2026-08-10
perf(codegen): inline strict `===` against a string literal (interp 2.29s -> 1.89s)

proggeramlug:
## The finding

`n.kind === "num"` — the shape every tree-walking interpreter, reducer and
discriminated-union dispatch is built out of — compiled to a `js_eq` →
`js_jsvalue_equals` call pair. On `gc-handoff/apps/interp.ts` that pair plus the
`memcmp` under it was **~21% of the program's runtime**, against 30% for the user
code itself. `interp.ts` is the only program in the perf corpus that resembles
real software, and this was its largest single remaining cost.

## The mechanism

The call was never necessary. When one operand is a string literal, *both* of a
string's runtime representations are known at compile time:

| what the operand is | how the inline sequence decides it |
|---|---|
| the pooled literal itself | `icmp eq` on the NaN-boxed bits — one instruction |
| the literal's SSO form (`charAt`, `JSON.parse`) | `icmp eq` against a compile-time immediate |
| any other SSO value | different bits ⇒ different content (SSO is canonical) ⇒ **false** |
| a number / int32 / pointer / bigint / null / undefined / bool | not `STRING_TAG` ⇒ **false**, no memory touched |
| a heap string of a different length | `byte_len` vs a compile-time constant ⇒ **false** |
| a heap string with a different first or last byte | two `i8` loads inside bytes the length check proved the header owns ⇒ **false** |
| a same-length, same-endpoints heap string | the only case that calls `js_string_equals` |

For literals of ≤ 2 bytes the sequence is total — no call exists on any path.

`{ kind: "num" }` stores the *same pooled pointer* the comparison site loads, and
GC evacuation rewrites the pool root and the object slot together, so identity
survives collection. That is what makes the true case one `icmp`.

The two string-equality arms with **no** literal operand (`names[i] === name`)
gained only the shortcuts that need no compile-time facts: identical bits, and
SSO × SSO with differing bits. Their existing fallbacks are untouched — which
matters for the legacy arm, since it keeps `js_get_string_pointer_unified`'s
number-coercing behaviour for operands whose `string` annotation lies. That
composition *materializes* an SSO operand onto the heap, so routing SSO × SSO
around it also removes two throwaway allocations per comparison.

## Semantics

Exact `===`, and in one respect *stricter* than what it replaces: the old
`both_strings` arm reached equality through `js_get_string_pointer_unified`,
which coerces a number to its decimal string, so `(5 as any) === "5"` could come
back true. Pinned against Node in
`test-files/test_strict_eq_string_literal_inline.ts`: `NaN !== NaN`, `+0 === -0`,
distinct heap strings with equal content are equal, int32 and double
representations of the same Number are equal, `new String("num") !== "num"`,
null vs undefined, object identity, and the multi-byte-UTF-8 cases where "first
byte" is not "first character" (`"é"` vs `"è"`, `"日本"` vs `"月本"`).

## Measurements — absolute seconds, quiet M1 mini, best-of-5

Outputs verified byte-identical to `node --experimental-strip-types` **before**
timing, on every row.

| benchmark | before | after | floor |
|---|--:|--:|--:|
| **interp** | **2.29** | **1.89** | ≤ 1.90 |
| iso_miss | 2.72 | 2.36 | canary: `misses 0` ✓ |
| asyncpipe | 0.90 | 0.90 | ≤ 0.93 |
| shapes | 0.28 | 0.28 | ≤ 0.29 |
| churn | 0.42 | 0.42 | ≤ 0.43 |
| churn_alloc | 0.37 | 0.37 | ≤ 0.38 |
| churn_read | 0.02 | 0.02 | ≤ 0.03 |
| push_cls | 0.36 | 0.35 | ≤ 0.37 |
| push_num | 0.14 | 0.13 | ≤ 0.14 |
| cycles | 0.19 | 0.19 | ≤ 0.20 |
| deeplist | 0.25 | 0.24 | ≤ 0.25 |
| tree | 1.69 | 1.63 | ≤ 1.66 |
| tree_wide | 2.11 | 2.10 | ≤ 2.15 |
| retain | 0.54 | 0.53 | ≤ 0.55 |
| retain_wide | 1.09 | 1.08 | ≤ 1.10 |
| fib40 | 0.39 | 0.39 | ≤ 0.40 |

Profile share of the equality family on `interp_big.ts` (symbolicated, two runs):

| | before | after (run 1 / run 2) |
|---|--:|--:|
| `js_jsvalue_equals` + `js_eq` + `memcmp` | ~21% | **2.16% / 1.80%** |

`evalNode` (user code) is now 35% of the profile.

## Tests

* `test-files/test_strict_eq_string_literal_inline.ts` — the ECMAScript edges
  above, diffed against Node.
* `crates/perry-codegen/src/expr/compare_tests.rs` — an IR census in the per-PR
  `cargo-test` gate: the `streqlit.*` blocks present **and** `js_eq` absent, so a
  fast path that is implemented but never reached fails here rather than
  silently. Paired negatives: loose `==` must keep its coercing helper, and a
  comparison with no literal operand must not use the dispatch. Plus numeric pins
  on the hand-built SSO immediate, which `perry-codegen` cannot share with
  `perry-runtime` — if that encoding drifts, `"+" === "+"` across a `charAt`
  result and a literal silently becomes false.

## Measured and rejected

Marking `js_string_equals` `nounwind willreturn readonly` (the #6082 allowlist)
so a dispatch chain could keep its discriminant in a register: **measured null**
(1.89 vs 1.89 interleaved, six pairs). Once the fast arms contain no calls at
all, LLVM already keeps the load in a register along the hot path, and the
remaining `js_string_equals` sites are cold. Dropped rather than shipped as
unmeasured surface on a soundness-sensitive allowlist.

## Summary by CodeRabbit

* **Performance**
  * String strict-equality comparisons are now handled more efficiently through optimized inline checks.
  * Common cases such as pooled, inline, empty, Unicode, and literal strings avoid unnecessary runtime work.

* **Bug Fixes**
  * Preserved exact `===` semantics across strings, boxed values, symbols, numbers, objects, and other types.
  * Improved handling of edge cases including NaN, signed zero, substrings, and cross-type comparisons.

* **Tests**
  * Added comprehensive coverage for strict string equality, operand order, union types, and dynamic comparisons.


proggeramlug:
## Merging as v0.5.1449

### Verified independently

**Semantics**: `test_strict_eq_string_literal_inline.ts` is **byte-identical to node 26.5.1** on my host, including the multi-byte-UTF-8 rows where "first byte" is not "first character". And the case where this is *stricter* than what it replaces checks out directly: `(5 as any) === "5"` is now `false` with `==` still `true`, matching node — the old `both_strings` arm reached equality through `js_get_string_pointer_unified`'s number coercion, so this fixes a live correctness bug on the way to the perf win.

**The IR census bites.** Neutering the admission (`if false && (lit_on_right || lit_on_left) …`) fails **3 of 8** tests — the "implemented but never reached" case the census exists for. Both negatives are present and right: loose `==` must keep its coercing helper, and two `any` locals must not use the literal dispatch. (Two earlier sabotage attempts of mine silently missed — a wrong function-name regex, then an edit that only printed — and the 8-passed results they produced were meaningless. The third, verified-applied one is the result that counts.)

### The dispatch table is the design

Enumerating **every runtime representation a string can have** and resolving each against compile-time knowledge is what makes this exact rather than heuristic: pooled-pointer identity in one `icmp`, the literal's SSO form as an immediate, SSO-canonicality making any *other* SSO bits `false` by construction, tag check before any memory touch, then length and endpoint bytes before the only remaining call. For ≤2-byte literals the sequence is total — no call on any path. `n.kind === "num"` never leaves straight-line code.

The pooled-identity argument is the subtle half and it is stated correctly: `{ kind: "num" }` stores the same pooled pointer the comparison loads, and evacuation rewrites the pool root and the object slot **together**, so identity survives collection.

Restraint on the no-literal arms is also right: `names[i] === name` gains only the shortcuts that need no compile-time facts, and the legacy fallback keeps its coercing behaviour for operands whose `string` annotation lies — with the bonus that routing SSO×SSO around it removes two throwaway heap materializations per comparison.

### Measurements

The equality family falling **~21% → ~2%** on `interp_big` with `evalNode` rising to 35% is the profile actually moving to user code, which is the point. All 16 floors held, outputs byte-verified before timing. This continues #7753's arc: interp 3.96 → 2.39 → **1.89 s**.

Gates 21/21.

PR #7768, opened 2026-08-10
ci: per-PR ext-link gate so a perry-runtime change cannot silently break the perry-ext-* link (#7656)

### LINKED ISSUE #7656: No per-PR gate links the perry-ext-* crates, so a perry-runtime change can break them until the next tag
A `perry-runtime` change broke the link of five `perry-ext-*` crates and **no per-PR gate could have caught it** (#7650 → fixed in #7655). It would have surfaced at the next tag, days later.

## The gap

`cargo-test` scopes a per-PR run to the changed crates' reverse-dependency closure (`scripts/ci_test_scope.py`); the **full workspace runs on tags and nightly only**. `perry-ext-*` sits outside the closure of a GC change to `perry-runtime`, so nothing built it.

But the ext crates are structurally *more* fragile than the closure suggests, not less: they link a **feature-stripped** runtime through `perry-ffi`'s `runtime-link`, built with `-Wl,-dead_strip`. Any new reference edge inside `perry-runtime` can keep a chain alive that the stripper had been removing, and the failure is a **link error**, not a test failure:

```
  "_js_blob_new", "_js_fetch_with_options", "_js_fetch_notify_signal_aborted"
```

In #7650 the edge was one added call — `pin_object` → `arena::classify_heap_space` — in code that had previously done a raw flag write.

## Proposal

Add a cheap per-PR arm that **links** (not runs) the ext crates whenever `crates/perry-runtime/**` or `crates/perry-stdlib/**` changes:

```
cargo test --release --no-run -p perry-ext-pdf -p perry-ext-lru-cache \
    -p perry-ext-node-forge -p perry-ext-mongodb -p perry-ext-http
```

Those five are the ones that actually failed; a wider set is better if it is affordable. Measured cost on a dev Mac from a warm target dir: **~5 minutes** for all five together, and it needs no runtime behaviour, no fixture and no host pinning — it is a linker check.

Alternatively, teach `ci_test_scope.py` that `perry-runtime`/`perry-stdlib` changes pull in `perry-ext-*` regardless of the declared dependency graph, since the real coupling here is the *link*, not the crate graph.

## Two things to get right

1. **The gate must assert it linked something.** A scope rule that silently selects zero ext crates would be green forever — the failure mode this repo has shipped four times (#6942/#6946, #7024, #7025). Assert a non-zero count of built targets.
2. **Run it once before making it required**, per CLAUDE.md's corollary: a new gate has never been green, so promoting it immediately blocks every open PR.

Context: #7650, #7655; the ext staticlib feature-stripping is also behind #7522 and #6847.

proggeramlug:
Fixes #7656.

## Why `cargo-test`'s scope cannot see this

Worth stating precisely, because the existing behaviour is deliberate and I did not change it: `ci_test_scope.py`'s `_is_fanout_leaf` keeps `perry-ext-*` / `perry-stdlib` **out** of the reverse-dependency fan-out, and its reasoning is sound —

> `perry-ext-*` and `perry-stdlib` are runtime FFI shims whose UNIT tests are self-contained pure-Rust logic … Excluding them from the reverse-dep fan-out keeps a foundational change from selecting ~40 crates.

That is right about the tests and silent about the **link**. These crates pull in a feature-stripped runtime through `perry-ffi`'s `runtime-link` built with `-Wl,-dead_strip`, so a new reference edge inside `perry-runtime` can keep alive a chain the stripper had been removing. In #7650 the edge was one added call (`pin_object` → `arena::classify_heap_space`) replacing a raw flag write, and the symptom was `Undefined symbols for architecture arm64` — not a failing assertion.

So this adds a separate link-only arm and leaves the fan-out exclusion alone. `cargo test --release --no-run` links the test binaries and stops; running them would add time and check nothing this does not.

`--release` deliberately: `-dead_strip` is a release-profile behaviour, and a dev-profile build links a different set — it would be green through exactly the regression this exists to catch.

## Cost — and why the scope is all 38 crates, not #7650's five

The issue said "a wider set is better if it is affordable". Measured on an arm64 dev Mac rather than argued:

| scope | target state | wall clock | test binaries linked |
|---|---|---|---|
| the five from #7650 | **cold** release target | **7:24** | 7 |
| **all 38** `perry-ext-*` | `perry-runtime` already built | **4:10** | 216 |

The shared `perry-runtime` release build dominates, so covering every ext crate costs *less* than building the runtime once. There is no reason to scope to the five that happened to fail.

## The two things #7656 asked to get right

**1. The gate asserts it linked something.** The count comes from cargo's own `--message-format=json` artifact records, and zero is a hard failure whose message says to fix the scope rather than trust the green. Sabotage-tested both ways: a stub cargo emitting no artifacts fails the step with `::error::ext-link linked ZERO test binaries`, one emitting two passes.

**2. Not promoted to required.** Per CLAUDE.md's corollary — a new gate has never been green, so making it required immediately blocks every open PR. Run it once, then promote.

## The list is derived, not written down

`scripts/ci_ext_link_scope.py` enumerates `crates/perry-ext-*` from the workspace, so a new ext crate is covered the day it lands. A hand-maintained list is precisely the failure mode #7748 had to repair in `ci_e2e_scope.py`, where the map named 3 of 24 suites and nothing could say so. The self-test additionally pins that #7650's five are still present in the derived list, so a rename reports itself instead of quietly shrinking the gate.

## Validation

- `python3 scripts/ci_ext_link_scope.py --self-test` — arming rules (runtime / stdlib / ffi / ext arm it; hir, codegen, docs do not), a non-empty derived list, and the #7650 five still covered.
- **The workflow bash was executed, not reviewed.** Extracted the run step and drove it with a stub `cargo` in three scenarios: links two → exit 0; links **zero** → exit 1 with the explicit error; cargo **fails** → exit 101.

  That third case found a real defect: the step originally exited **0** when cargo failed — the exact regression the job exists to catch — because it relied implicitly on the runner's default `bash -e`. It is now `set -euo pipefail` explicitly.
- Real `cargo test --release --no-run` over all 38 crates: exit 0, 216 binaries (the numbers above).
- Also guarded `setup-llvm22` on the scope condition; it was running even when the diff cannot affect the archives, which defeats the point of the cheap scope step.

## Scope note

This gate would have caught #7650. It does **not** address #7629, whose crash also involves the ext archives — that one is a duplicate-runtime / reactor problem at *auto-optimize link* time, not a missing build, and remains open.

No version bump.

### DIFF
--- a/.github/workflows/ext-link.yml
+++ b/.github/workflows/ext-link.yml
@@ -1,0 +1,155 @@
+# ---------------------------------------------------------------------------
+# ext-link — per-PR LINK check for the `perry-ext-*` crates (#7656)
+#
+# A `perry-runtime` change broke the link of five `perry-ext-*` crates and no
+# per-PR gate could have caught it (#7650, fixed in #7655); it surfaced at the
+# next tag, days later.
+#
+# Why `cargo-test`'s scope cannot see this: `ci_test_scope.py` selects a
+# reverse-dependency closure, and `_is_fanout_leaf` deliberately keeps
+# `perry-ext-*` / `perry-stdlib` OUT of the fan-out. That is correct on its own
+# terms — their unit tests are self-contained pure-Rust logic, and re-running
+# ~40 crates on every foundational change is the cost the scoping exists to
+# avoid. What it misses is that for these crates the coupling is the LINK, not
+# the test: they pull in a feature-stripped runtime through `perry-ffi`'s
+# `runtime-link` built with `-Wl,-dead_strip`, so a new reference edge inside
+# `perry-runtime` can keep alive a chain the stripper had been removing. In
+# #7650 that edge was one added call (`pin_object` -> `arena::classify_heap_space`)
+# in code that had previously done a raw flag write, and the symptom was
+# `Undefined symbols for architecture arm64`.
+#
+# So this job BUILDS and does not RUN: `cargo test --no-run` links the test
+# binaries and stops. Running them would add time and check nothing this does
+# not already.
+#
+# `--release` deliberately: `-dead_strip` is what makes the failure, and it is
+# a release-profile behaviour. A dev-profile build links a different set and
+# would be green through exactly the regression this exists to catch.
+# ---------------------------------------------------------------------------
+name: ext-link
+
+on:
+  pull_request:
+  workflow_dispatch:
+
+concurrency:
+  # PR runs supersede each other; a manual dispatch is never cancelled.
+  group: ext-link-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+permissions:
+  contents: read
+
+jobs:
+  ext-link:
+    runs-on: ubuntu-latest
+    # Measured on an arm64 dev Mac. COLD release target, the five crates from
+    # #7650 only: 7:24, 7 test binaries. All 38 ext crates with perry-runtime
+    # already built: 4:10, 216 test binaries. The shared `perry-runtime` release
+    # build dominates, so covering every ext crate costs LESS than building the
+    # runtime once — which is why the scope is all of them rather than the five
+    # that happened to fail in #7650.
+    #
+    # The bound is a backstop, not a budget: it has to cover a cold sccache
+    # building perry-runtime in release from scratch on a shared runner, which
+    # is far slower than the numbers above, while still cutting a true hang.
+    timeout-minutes: 120
+    env:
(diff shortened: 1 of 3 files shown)


proggeramlug:
Fixed by #7765 (merged as 0a2bf15bd, v0.5.1455) — the receiver-tag gates plus the keys-walk slot read; full audit on the PR.

proggeramlug:
Reopening — my previous close was wrong: this ext-link-gate ticket is unrelated to #7765. The '#7768' in that PR's early commit messages was a stale placeholder (its own commit 3af7337a2 renumbered them to #7765). Apologies for the noise.

Issue #5625, opened 2026-06-24
test262 regressions introduced by the #5579 fix batch (Object.defineProperty writable + RegExp /u-mode escapes + TypedArray species)


proggeramlug:
## Severity: MEDIUM (fix-batch regressions)

The 2026-06-24 re-measure (main @ `911a97948`, node v26.3.0) shows the #5579 fix batch recovered the big regression clusters (overall 90.7% → 93.3%) but **introduced 26 fresh test262 failures** that passed on the prior run (`sweep-0623`). Three concrete sub-clusters:

### 1. `Object.defineProperty` / `defineProperties` — "property should be writable" (15)

A property defined with `{writable:true}` (or default) now reports non-writable. Suspect: #5574 (own `valueOf` wins in member dispatch), #5588 (Object.assign Proxy), or #5597 (static-method delete).

```
built-ins/Object/defineProperty/15.2.3.6-4-278.js          Expected obj[property] to be writable, but not
built-ins/Object/defineProperty/15.2.3.6-4-285.js
built-ins/Object/defineProperty/15.2.3.6-4-531-11.js       Expected obj[prop] to be writable
built-ins/Object/defineProperties/15.2.3.7-6-a-267.js
built-ins/Object/defineProperties/15.2.3.7-6-a-274.js   (… 15 total, -267/-268/-274..-277, -278/-279/-285..-288, -531-11/-540-7/-547-3)
```

### 2. RegExp unicode-mode restricted escapes — #5594 over-relaxed (2)

#5594 made Annex B legacy escapes (`\1` decimal, `\c` control) valid — but they must **still throw `SyntaxError` under the `/u` (and `/v`) flag**. The unicode-mode strictness was lost.

```
built-ins/RegExp/unicode_restricted_octal_escape.js        RegExp("\1","u"): Expected a SyntaxError to be thrown
built-ins/RegExp/unicode_restricted_identity_escape_c.js   Expected a SyntaxError to be thrown
```

### 3. TypedArray `speciesctor` — silent (no output) (4)

`%TypedArray%.prototype` species-constructor path: a throwing species ctor / callback now produces no output (silent early exit or swallowed throw).

```
built-ins/TypedArray/prototype/slice/speciesctor-get-ctor-returns-throws.js
built-ins/TypedArray/prototype/map/speciesctor-get-ctor-returns-throws.js
built-ins/TypedArray/prototype/subarray/speciesctor-get-ctor-returns-throws.js
[...]

PR #7769, opened 2026-08-10
perf(runtime): class dispatch and instanceof stop consulting locked hash maps — shapes.ts 0.28s → 0.23s

proggeramlug:
## The finding

`gc-handoff/apps/shapes.ts` — a scene-graph class hierarchy (deep `extends` chains, virtual dispatch through a base-typed collection, `super()` chains, `instanceof`, getters, statics, a fieldless subclass and a two-level indirect subclass) — was the widest margin a competing compiler held anywhere in the corpus.

A symbolicated profile on the pinned quiet mini showed the cause is not codegen: **the runtime answered "who is this class's parent?" and "which method is this?" with a process-global lock plus a SipHash probe, per hop, per call.** `std::hash::random::RandomState` was 1.3% of runtime and `pthread_mutex_{lock,unlock}` another 2.8% — for what is semantically an indexed load in a single-threaded program.

## What changed

1. **The parent chain is a dense atomic mirror.** `get_parent_class_id` — walked by `instanceof`, vtable dispatch, static-member lookup, `super()` construction, symbol lookup and the typed-feedback guards — took `CLASS_REGISTRY.read()` plus a hash probe on *every hop*. Edges whose child id fits a 64 K window now answer from one atomic load; the reserved builtin bands and high-bit synthetic ids keep using the map. The slot stores `parent + 1`, which is what lets one word distinguish "absent" from "registered with parent id 0".

2. **Monotone latches** (`registry_latch::RegistryLatch`, #7755) on the `Symbol.hasInstance` / `Symbol.toStringTag` / `extends Error` / fetch-parent / generic-origin / class-static-symbol / timer-id registries. `js_instanceof` probed three of them on every evaluation and `class_chain_reaches` one on every hop — all empty in a program that uses none of those features. The `hasInstance` latch also keeps the string-keyed `well_known_symbol("hasInstance")` interning probe off the path entirely.

3. **The dispatch tower caches its own resolution.** `js_native_call_method` is the virtual-call path for every receiver whose static type does not pin the callee — which is *every* call through a base-typed collection, the shape a class hierarchy is written in. Reaching a resolution cost a `String` allocation for the method name, a `RuntimeHandleScope`, ~900 lines of exotic-receiver probes, a GC-heap `StringHeader` allocation for the prototype probe, a lock and two SipHash lookups: four heap allocations and a lock around `shape.area()`'s single multiply. A per-thread, content-keyed cache now records the tower's outcome for a `(class_id, method name)` pair.

4. **One argument vector instead of two `Vec`s.** `call_vtable_method` built a `Vec<f64>` of positional args and `call_fn_with_f64_args` built a second with `this` prepended — two `malloc`/`free` round-trips for a zero-argument virtual call.

## Three things about the cache that are load-bearing

- **Both resolution points populate it.** The first attempt cached only the tower's tail vtable arm, which checks the receiver's OWN class vtable — so every INHERITED method (`class Square extends Rect` calling `Rect`'s `area`) missed forever, and inherited methods are the common case in any real hierarchy. The parent-chain walk in `handle_methods` is the site that mattered: adding it took shapes from 0.25 to 0.23.
- **It is keyed on the name BYTES, not its address.** The sibling `VTABLE_IC` keys on the rodata pointer codegen passes, but `js_native_call_method_str_key` reaches the same tower with a name materialised into a *caller-stack* scratch buffer, where two different short names genuinely land at the same address in successive calls. A sabotage test plants exactly that and asserts a miss.
- **A hit never substitutes for an object-specific check.** Everything the tower decides per RECEIVER is re-verified on every hit: pointer classification through `gc_pointer_and_type_from_value` (buffers, typed arrays, Sets, Maps, RegExps and Symbols are raw allocations with no `GcHeader`, so screening them *before* the header read is a memory-safety requirement, not an optimisation — see #5625), `OBJECT_TYPE_REGULAR`, a null `meta` (which rules out both a per-instance `[[Prototype]]` override and any own accessor descriptor), the own-key scan an own field would win on, and the recorded-prototype probe. The `using`/`await using` disposal hooks and the iterator helpers are excluded by name, because both branch on per-object state the guard cannot see.

Prototype surgery now bumps `VTABLE_GEN`. `invalidate_class_prototype_fast_guards` is the single latch all three prototype-write entry points funnel through, but the method-dispatch caches were only retired by class *registration*.

## Thread safety

`perry/thread` spawns real OS threads with independent arenas. The parent mirror is process-global atomics published (`Release`) *before* the map insert, so no reader can observe an edge through the map without it also being visible densely; the latches follow `RegistryLatch`'s arm-before-publish rule, whose only possible wrong observation ("idle while non-empty") that rule excludes. The dispatch cache is per-thread and starts empty on every worker, so a worker populates it from its own tower run rather than inheriting one.

`test-files/test_issue_7769_thread_class_dispatch.ts` runs the same hierarchy on the main thread, through `parallelMap` (8 workers) and through `spawn`, and compares all three — then re-checks the main thread afterwards.

## Measured — rebased onto `c2a96b638`, both arms rebuilt from that merge-base

Protected benchmarks re-measured **interleaved** (arms alternating inside one window, best of 7): a sequential base-then-arm pass showed +0.01-0.02 on several rows, and interleaving proved that was host drift moving both arms together, not a regression.

| | base | arm | | | base | arm |
|---|---|---|---|---|---|---|
| **shapes** | **0.29** | **0.24** | | churn | 0.43 | 0.43 |
| iso_miss | 2.46 | 2.45 | | churn_alloc | 0.38 | 0.38 |
| asyncpipe | 0.92 | 0.91 | | push_cls | 0.36 | 0.36 |
| interp | 1.89 | 1.89 | | retain | 0.55 | 0.55 |
| churn_read | 0.02 | 0.02 | | retain_wide | 1.12 | 1.12 |
| push_num | 0.14 | 0.14 | | tree | 1.68 | 1.68 |
| cycles | 0.19 | 0.19 | | tree_wide | 2.17 | 2.17 |
| deeplist | 0.25 | 0.25 | | fib40 | 0.40 | 0.40 |

Every protected benchmark is **identical between the two arms**. `shapes` is the only row that moves: **0.29 → 0.24**, i.e. **8.6x behind scriptc** (0.0272 s), down from 10.7x. Outputs byte-identical to the baseline arm and to `node --experimental-strip-types`, verified before timing.

The win is marginally *larger* after the rebase than before it (0.28 → 0.23 on the old base), because #7762 put a `class_generic_origin` probe inside `class_prototype_object`, which the parent-chain walk calls on every hop — one more locked hash probe on main, which this change's latch answers from an atomic load.

On the `shapes_big` profile (two agreeing 7 s runs, ~5 250 samples each; measured pre-rebase) the dispatch cluster falls from **12.3% to 7.5%** and `RandomState` + `pthread_mutex_*` from **5.6% to 3.6%**. Four leaders leave the top ranks: `get_parent_class_id` (3.3% → 0.5%), `class_chain_reaches` (2.1% → 0.3%), `js_instanceof` (1.0%), and `RuntimeHandle::get_nanbox_u64` — the single hottest symbol in the program at 4.9% — because the fast path needs no handle scope.

## What this does NOT fix (and where the rest of the gap is)

- The residual lock traffic is `is_registered_symbol` / `is_registered_map` / `is_registered_set`, reached from `js_array_get_f64` (so, every `arr[i]`) and from the dispatch guard's pointer classification. Those registries are *already* latched (#7474, #7755) — the latches are simply **armed**, because something in startup materialises a well-known Symbol. That turns a free atomic load into a process-global mutex for every array element read in every program. Worth chasing, but it is Map/Set/Symbol registry work, adjacent to #7765.
- The rest of `shapes.ts` is array element reads (`js_array_get_f64` + `array_object_flags` + `js_array_length`, ~10%) and the GC layout tables (~14%).
- Two pre-existing divergences, verified to produce identical wrong answers on a binary built from the merge-base: `Class.prototype.m = fn` after `m`'s first dispatch still resolves to the vtable method, and `Object.setPrototypeOf(instance, donor)` does not redirect an already-dispatched method on that instance. Both are produced by the tower itself — the guard rejects a receiver with a `meta` record, and prototype surgery now bumps `VTABLE_GEN` — so neither is reached from the cache. They are documented in the gap test rather than asserted, so the file stays byte-identical to Node.

## Tests

- `test-files/test_gap_7769_class_dispatch_shapes.ts` — fieldless subclass, two-level indirect subclass, class expressions (with and without `extends`), a subclass overriding nothing, base-typed dispatch cold *and* warm, `instanceof` across the lattice, own-field shadowing after the pair has been cached, `super()`/statics up the chain, computed method names. Byte-identical to Node.
- `test-files/test_issue_7769_thread_class_dispatch.ts` — cross-thread dispatch.
- 10 runtime unit tests: dense-mirror/map agreement, `Some(0)` vs absent, out-of-window fallback, latch defaults, and five for the cache including the stack-address sabotage.

`cargo test --release -p perry-runtime`: 2044 passed, 0 failed. Gap suite (518 tests): 19 failures, **every one byte-identical on a merge-base binary** — all pre-existing.

## Summary by CodeRabbit

* **Performance**
  * Improved class method dispatch, inheritance checks, repeated calls, and varying argument counts.
  * Reduced overhead for dynamic calls, timers, and class metadata lookups.

* **Bug Fixes**
  * Improved cache invalidation when class prototypes change.
  * Preserved correct behavior for inheritance, `instanceof`, static members, computed names, and shadowed methods.
  * Improved consistency across parallel and threaded execution.

* **Tests**
  * Added coverage for class inheritance, virtual dispatch, caching, prototype behavior, timer handling, and threaded execution.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1451
+**Current Version:** 0.5.1452
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1451"
+version = "0.5.1452"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1451"
+version = "0.5.1452"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1451"
+version = "0.5.1452"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1451"
+version = "0.5.1452"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1451"
+version = "0.5.1452"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1451"
+version = "0.5.1452"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1451"
+version = "0.5.1452"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1451"
+version = "0.5.1452"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1451"
+version = "0.5.1452"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1451"
+version = "0.5.1452"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1451"
+version = "0.5.1452"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1451"
+version = "0.5.1452"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1451"
+version = "0.5.1452"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1451"
+version = "0.5.1452"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1451"
+version = "0.5.1452"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1451"
+version = "0.5.1452"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1451"
+version = "0.5.1452"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1451"
+version = "0.5.1452"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1451"
+version = "0.5.1452"
 dependencies = [
  "bcrypt",
  "perry-ffi",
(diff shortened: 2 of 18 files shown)


proggeramlug:
## Merging as v0.5.1452, with one fix added on the branch

### The three load-bearing cache properties all verify

The PR names them, and each has a test that passes here: **both resolution points populate the cache** (the first attempt cached only the own-vtable arm, so every *inherited* method missed forever — and inherited methods are the common case in a hierarchy); **keyed on name bytes, not address** (`a_hit_requires_matching_name_bytes_not_a_matching_address` — the `str_key` path materialises names into a caller-stack scratch buffer where two short names genuinely reuse an address); and **a hit never substitutes for a per-receiver check** — the raw-allocation screen before any `GcHeader` read is a memory-safety requirement (#5625), not an optimisation, and the own-key scan means shadowing-after-cache still wins, which the gap test exercises explicitly.

`dense_table_answers_the_same_chain_as_the_map`, `parent_zero_is_distinguishable_from_absent` (the `parent + 1` encoding earning its keep), and `a_class_registration_invalidates_every_entry` all pass. The parity test is **byte-identical to node** on my host; the thread test is a `test_issue_*` self-checker (node cannot import `perry/thread`) and all its own assertions pass — `parallelMap allMatch: true`, `spawn match: true`, main-thread recheck clean.

The prototype-surgery finding is quietly important: `invalidate_class_prototype_fast_guards` was the single latch all three prototype-write entry points funnel through, but the method caches were only retired by class *registration* — this PR routing `VTABLE_GEN` bumps through it closes a staleness hole that predates the cache it protects.

### The fix I added

The addr-class ratchet failed the branch: the cache's keys-array screen used a bare `keys_ptr < 0x10000` floor — a **new** handle-floor site over the file's baseline of 3. That floor sits below the fetch/zlib/proxy handle bands, so a handle id would have been dereferenced as a keys array (#7531/#7709's exact class). Converted to `addr_class::is_above_handle_band`, ratchet green, 2046/2046 tests pass. The ratchet catching a new site in the same file it already guards, on a PR adding a pointer screen, is that gate working precisely as designed.

Gap suite: 19 failures, every one byte-identical on a merge-base binary. One gate run flaked on `perry-runtime --lib` at a load transient; clean 2/2 on direct re-runs. Gates otherwise 21/21.

PR #7772, opened 2026-08-10
fix(stdlib): build with --no-default-features again, unbreaking the auto-optimize relink (#7764)

### LINKED ISSUE #7764: perry-stdlib fails to build with --no-default-features (breaks auto-optimize relink)
## Summary

`perry-stdlib` does not compile with `--no-default-features`, which is the configuration the auto-optimize relink uses. Every `perry` compile that triggers auto-optimize falls back to prebuilt archives with a warning, and any ad-hoc build needs `PERRY_NO_AUTO_OPTIMIZE=1` as a workaround.

## Reproduction

```
cargo build -p perry-stdlib --no-default-features
```

```
error[E0433]: cannot find `events` in `crate`
   --> crates/perry-stdlib/src/common/dispatch/init.rs:668:57
    |
668 |     perry_runtime::js_set_native_events_dispatch(crate::events::js_events_native_dispatch);
    |                                                         ^^^^^^ could not find `events` in the crate root
    |
note: found an item that was configured out
   --> crates/perry-stdlib/src/lib.rs:57:9
    |
 56 | #[cfg(feature = "bundled-events")]
    |       -------------------------- the item is gated behind the `bundled-events` feature
 57 | pub mod events;
```

Two further instances of the same shape:

```
error[E0433]: cannot find `async_bridge` in `common`
  --> crates/perry-stdlib/src/worker_threads/broadcast_channel.rs:67:20
  --> crates/perry-stdlib/src/worker_threads/broadcast_channel.rs:102:20
```
gated behind `async-runtime`.

## Cause

The call site added in #7745 (`route the module-level events.* helpers through the dynamic dispatch`) is not behind the `#[cfg(feature = "bundled-events")]` that gates the module it references. The neighbouring dispatch registrations in the same function *are* cfg-gated — e.g. the `database-sqlite` line immediately below it — so this looks like an omission rather than a design choice.

## Impact

Auto-optimize prints:

> auto-optimize: cargo build failed, using prebuilt libraries. The prebuilt archives may lack the feature-gated `js_*` entrypoints this compile routed to ext crates; if the link fails with undefined symbols, fix the cargo error above

so the failure is soft — but it silently gives up the feature-stripped rebuild, which is the whole point of the pass.

## Fix

Gate the three call sites to match the modules they reference (`bundled-events`, `async-runtime`). Worth adding a CI arm that builds `perry-stdlib --no-default-features`, since nothing currently covers the configuration auto-optimize actually uses.

proggeramlug:
Fixes #7764.

## What was broken

`cargo build -p perry-stdlib --no-default-features` produced **12 errors**. That is the configuration the auto-optimize relink builds with, so every `perry` compile that triggered auto-optimize fell back to the prebuilt archives with a warning, and any ad-hoc build needed `PERRY_NO_AUTO_OPTIMIZE=1`.

Both causes violate the contract `common/mod.rs` already states in prose:

> Tokio-backed promise/runtime bridge … **Always-on code that references it must also be `#[cfg(feature = "async-runtime")]`-gated.**

## Cause 1 — one line, exactly as diagnosed (#7745)

`js_set_native_events_dispatch` referenced `crate::events` without the `#[cfg(feature = "bundled-events")]` that gates the module. The neighbouring registrations in the same function *are* gated — `database-sqlite` on the very next line — which is what makes this an omission rather than a decision. One `#[cfg]`.

## Cause 2 — eleven sites the issue listed as "the same shape"

They are the same shape but not the same fix. `worker_threads` is always-on and reaches `common::async_bridge` in eleven places across five files, and **neither obvious repair works**:

- **`#[cfg]` per call site** — two of them are value-producing (`js_promise_new_for_native_resolution`), so gating the statement leaves nothing to return.
- **Gate the whole `worker_threads` module** — worse. It has no feature of its own, so its FFI symbols would vanish from the stripped archive and a program importing `node:worker_threads` would fail to **link**. That trades a build error for the #7629 family of failure, which is a strictly worse outcome.

So `worker_threads/async_shim.rs` supplies the four entry points in both configurations — forwarding to `async_bridge` when it is compiled in, settling **inline** when it is not.

That inline arm is not invented semantics. The queue exists to hand work to the pump; with no pump there is nothing to hand it to, and doing the same work synchronously reaches the same observable end state (the promise settles). The pinning `js_promise_new_for_native_resolution` performs is likewise a consequence of deferral — it guards the window between creation and the pump's resolution — and an inline settle spans no collection point, so a plain `js_promise_new` is its correct counterpart. Each of those arguments is written down at the shim.

## Validation

- `cargo build -p perry-stdlib --no-default-features`: **0 errors** (was 12).
- `cargo build -p perry-stdlib` (default features): **0 errors**.

  Both directions deliberately: my first cut of the shim accidentally imported *itself*, and `--no-default-features` could not see it because that arm is `#[cfg]`'d out. Checking only the configuration named in the issue would have shipped it.
- **The actual payoff, checked end to end**: `perry compile` of a small program now completes the auto-optimize relink with no fallback warning and runs correctly.
- `cargo test -p perry-stdlib --lib`: 111 passed, 0 failed.
- Targeted parity across worker-threads / events / message-port / broadcast / async / promise gap tests: all pass.

One process note: a `test_gap_worker` COMPILE_FAIL appeared once and did not reproduce — it raced a rebuild that was still writing the archives. Re-run twice clean before believing that shape of failure.

No version bump.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed builds when optional default features are disabled.
  * Improved compatibility for worker-thread messaging and promise handling across feature configurations.
  * Prevented event-dispatch initialization from referencing unavailable functionality.
* **Documentation**
  * Added a changelog entry describing the feature-configuration build fixes.

### DIFF
--- a/changelog.d/7772-stdlib-no-default-features.md
+++ b/changelog.d/7772-stdlib-no-default-features.md
@@ -1,0 +1,13 @@
+### Fixed
+
+- **`perry-stdlib` builds with `--no-default-features` again (#7764).** That is the configuration the auto-optimize relink uses, so while it was broken every `perry` compile that triggered auto-optimize silently fell back to the prebuilt archives, and ad-hoc builds needed `PERRY_NO_AUTO_OPTIMIZE=1` as a workaround.
+
+  Twelve errors, from two causes, both violations of the contract `common/mod.rs` states in prose: *"Always-on code that references it must also be `#[cfg(feature = "async-runtime")]`-gated."*
+
+  **One** was #7745's omission, exactly as the issue diagnosed: the `js_set_native_events_dispatch` registration referenced `crate::events` without the `#[cfg(feature = "bundled-events")]` that gates the module. The neighbouring registrations in the same function are gated (`database-sqlite` on the next line), which is what makes it an omission rather than a decision.
+
+  **Eleven** were `worker_threads` — always-on, and referencing `common::async_bridge` across five files. Neither obvious repair works: two sites are value-producing (`js_promise_new_for_native_resolution`), so `#[cfg]` on the statement leaves nothing to return; and gating the whole `worker_threads` module is worse, because it has no feature of its own, so its FFI symbols would vanish from the stripped archive and a program importing `node:worker_threads` would fail to LINK — trading a build error for the #7629 family of failure.
+
+  So `worker_threads/async_shim.rs` provides the four entry points in both configurations: forwarding to `async_bridge` when it is compiled in, and settling **inline** when it is not. That is not invented semantics — the queue exists to hand work to the pump, and with no pump there is nothing to hand it to, so doing the same work synchronously reaches the same observable end state. The pinning `js_promise_new_for_native_resolution` performs is likewise a consequence of deferral, and an inline settle spans no collection point, so a plain `js_promise_new` is its correct counterpart.
+
+  Verified in BOTH directions — `cargo build -p perry-stdlib` and `--no-default-features` each build clean — because the first cut of the shim accidentally imported itself, which only the default-features build could see.
--- a/crates/perry-stdlib/src/common/dispatch/init.rs
+++ b/crates/perry-stdlib/src/common/dispatch/init.rs
@@ -665,6 +665,15 @@ pub unsafe extern "C" fn js_stdlib_init_dispatch() {
     );
     // Module-level `events.*` helpers reached indirectly (captured value,
     // type-erased receiver, spread call) — see `js_events_native_dispatch`.
+    //
+    // #7764: gated to match `pub mod events`, which is `bundled-events`. #7745
+    // added this line ungated, so `--no-default-features` — the configuration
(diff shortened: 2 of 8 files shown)


proggeramlug:
Additional evidence: this PR also closes the live half of #6847. On `main`, `cargo check -p perry-stdlib --no-default-features --features async-runtime,external-zlib-pump` — the exact feature set the auto-optimize pipeline selects — fails with the `crate::events` E0433 this PR fixes; on this branch it is clean.

Cold-start consequence measured today: with `target/perry-auto-*` and both object caches removed, the compile prints `auto-optimize: cargo build failed (exit status: 101), using prebuilt libraries` and falls back. The link then succeeds and the program runs correctly, so #6847's reported undefined-symbol failure no longer occurs — but every auto-optimize build on that configuration is silently shipping the un-optimized prebuilt archives, with one warning line as the only signal. This PR is what makes the rebuild actually succeed.

Issue #7773, opened 2026-08-10
declared-number class field holding a string: `o.x + 1` coerces to NaN (Node concatenates); any-laundered add evaporates


proggeramlug:
## Symptom

A declared-`number` class field that actually holds a string (Perry does not enforce annotations) diverges from Node when read in an arithmetic `+`:

```ts
class C { x: number; constructor(x: number) { this.x = x; } }
function g(o: C): void { (o as any).x = "s"; }
function f(): string {
  const o = new C(1);
  g(o);
  return `${o.x + 1}`;
}
console.log(f());
```

Node (26.5.1, strip-types): `s1` (string concat — the runtime value decides).
Perry: `NaN` (the Add was classified numeric from the DECLARED field type, so the class-field get lowers in number context and its fallback arm hard-coerces via `js_number_coerce`).

## Second shape — the add EVAPORATES through an `any` local

```ts
  const v: any = o.x;
  return `${v + 1}`;
```

Node: `s1`. Perry: `s` — the `+ 1` is dropped entirely, not even coerced (smells like the #7590 `discard_expr_value` family or an any-add lowering hole).

## Attribution

Both reproduce with `PERRY_PTR_SHAPE_LOCALS=0`, no arrays, no element groups, on a build of main + #7770 (the #7770 collector change is not involved; its A/B arms behave identically here). The escape (`g(o)`) is required — a non-escaping `o` gets scalar-replaced and prints `s1` correctly, which is why the trivial form never showed this.

## Where to look

`expr/property_get/helpers.rs::lower_raw_f64_class_field_get_for_number_context` is only entered when the surrounding context was already classified numeric; the misclassification is upstream in the Add lowering (`type_analysis::is_numeric_expr` trusting the declared field type for an aliased receiver). The guarded diamond's fallback arm then coerces unconditionally.

Found while validating #7770 (its gap test deliberately avoids `+` on poisoned fields and points here).

Issue #7766, opened 2026-08-10
repsel: rule-1 provenance does not survive a function boundary — a typed `P[]` parameter still reads fields by name (first-party half of #7152/#7170)


proggeramlug:
## The gap, in one measurement

A fully-typed field read through a **parameter** array is still a by-name property lookup with a feedback guard and a boxing round-trip. The same code with the array as a **local** compiles to offset loads.

```ts
class P { constructor(public x: number, public y: number) {} }

// (A) array as a PARAMETER — denied
function total(ps: P[]): number {
  let s = 0;
  for (let i = 0; i < ps.length; i++) s += ps[i].x + ps[i].y;
  return s;
}

// (B) same array, LOCAL, built by push, consumed in the same function — promoted
function run(n: number): number {
  const a: P[] = [];
  for (let i = 0; i < n; i++) a.push(new P(i, i + 1));
  let s = 0;
  for (let i = 0; i < a.length; i++) { const r = a[i]; s += r.x + r.y; }
  return s;
}
```

`perry compile … --opt-report --no-link` on **main @ v0.5.1448**:

```
(A)  Ptr<Shape>   0 selected / 1 denied   (0% of 1 candidates)
       ptr-shape rule 1 (provenance)
       in function total -> Boxed

(B)  Ptr<Shape>   1 selected / 1 denied   (50% of 2 candidates)
       local `r` -> Ptr<Shape> (in function run)
```

Emitted IR for the loop body, same read shape in both:

| | `js_object_get_field_by_name_f64` | `js_typed_feedback_class_field_get_guard` | `js_throw_type_error_property_access` |
|---|---|---|---|
| (A) parameter | ×2 | ×2 | ×2 |
| (B) local | **gone** | **gone** | **gone** |

Plus `js_number_coerce` ×2 in (A) — the boxing round-trip, because the by-name read hands back a NaN-boxed double.

**The machinery works. It cannot see through a function boundary.**

## Why this is the right lever

`--opt-report` names the rule, and #7152 already measured its size on real dependency JS: **rule 1 accounts for 506 of 746 denied candidates** — four times the next bucket (rule 5, 99). #7170 then re-measured and found **96% of those 506 are record literals**.

So the same rule is the wall in two populations that look unrelated:

* **dependency JS** — tracked in #7152 / #7170, framed around CommonJS wrappers
* **first-party TypeScript** — the case above, tracked nowhere

They are one defect. This ticket is the single place to fix it.

## What rule 1 currently requires

[...]

Issue #7771, opened 2026-08-10
repsel: a proven element fetch is still a runtime call — specialize a[i] inside the element-shape guarded clone


proggeramlug:
## The measurement

On **main @ v0.5.1448**, fetching an element from an array whose element shape is already proven is still a runtime call.

```ts
class P { constructor(public x: number, public y: number) {} }
function run(n: number): number {
  const a: P[] = [];
  for (let i = 0; i < n; i++) a.push(new P(i, i + 1));
  let s = 0;
  for (let i = 0; i < a.length; i++) { const r = a[i]; s += r.x + r.y; }   // <- a[i]
  return s;
}
```

`--opt-report`: `local r -> Ptr<Shape> (in function run)`. The promotion succeeds, the by-name field machinery is gone, and the loop runs under the element-shape guard.

Isolating the **read loop** (subtracting a build-only variant, because construction is ~12 calls and swamps it):

```
js_array_get_f64   x1     <- THIS ISSUE
js_number_coerce   x2     <- separate issue, see "Related"
js_gc_loop_safepoint x1   <- by design (#7721)
```

`a[i]` where `a`'s element shape is proven, `i` is an in-bounds i32 loop counter, and the result is immediately bound to a `Ptr<Shape>` local — and it is still a call into the runtime.

## Where this sits

The guarded-clone machinery already exists and is already entered here. `expr/element_shape_guard.rs` provides:

* `emit_element_shape_loop_preheader_check` — run **once**, before the fast clone
* `emit_element_shape_field_load` — run **per access inside the fast clone**

So there is already a per-access fast path for the *field* load inside the clone. The element *fetch* is what has no equivalent: it still goes through the generic index-get lowering (`expr/index_get.rs` / `expr/index_get/`) to `js_array_get_f64`.

The engine plan names this as the third leg of the same route — *invariant bit → versioned-loop consumer → element `Ptr<Shape>`*. The first two shipped (#7480, #7669, #7701). This is the one left.

**First thing to establish**, because it changes the shape of the fix: confirm whether the fast clone is genuinely **entered** for this loop, or whether it is admitted and then degrades. #7701 added `assert_fast_clone_is_entered` for exactly this — a `cond_br` INTO the clone, not an unconditional branch to the slow path. If the clone is not entered, the fix is upstream of this issue and this ticket is mis-scoped; say so rather than building on it.

## Acceptance criteria

1. The read loop above emits **zero** `js_array_get_f64` for `a[i]`.
[...]

PR #7774, opened 2026-08-10
repsel: discharge the numeric-field proof at element-GROUP scope so proven reads drop js_number_coerce (#7770)

### LINKED ISSUE #7770: repsel: a Ptr<Shape>-proven numeric field still loads NaN-boxed — extend the raw-f64 field load past its scalar-replaced gate
## The measurement

On **main @ v0.5.1448**, a `Ptr<Shape>`-promoted local — the case that already works — still coerces every numeric field read.

```ts
  const a: P[] = [];
  let s = 0;
  return s;
}
```

`--opt-report` confirms the promotion succeeds: `local r -> Ptr<Shape> (in function run)`. And the by-name machinery is gone from the IR — no `js_object_get_field_by_name_f64`, no `js_typed_feedback_class_field_get_guard`, no `js_throw_type_error_property_access`.

But subtracting a build-only variant to isolate the **read loop** leaves:

```
js_array_get_f64   x1     <- separate issue, see "Related"
js_number_coerce   x2     <- THIS ISSUE, one per field read
js_gc_loop_safepoint x1   <- by design (#7721)
```

Two declared-`number` fields, on a shape that is proven, read through a path with no dynamism left — and each one still pays a coercion, because **the slot it loads from holds a NaN-boxed double**.

## Why this is small

The representation already exists on both sides:

* **Runtime**: typed shapes carry a `raw_f64_mask` beside the pointer mask (`gc/layout.rs`), and the GC already understands slots that hold raw f64 rather than NaN-boxed values.
* **Codegen**: `typed_shape.rs` computes `raw_f64_mask_words` and declares it, and `expr/property_get/helpers.rs::lower_raw_f64_class_field_get_for_number_context` **already lowers a numeric field read to a raw load**.

That function is gated on `.scalar_replaced` — objects proven never to escape — via `scalar_replaced_field_is_raw_f64` and `scalar_replaced_field_raw_f64_store_state`.

`Ptr<Shape>` is a *weaker* condition that the collector already computes (`collectors/ptr_shape.rs`: provenance + containment). A `Ptr<Shape>` local's dynamic class is exactly `C`, which is precisely what the raw-f64 decision needs. **The work is extending an existing gate to an existing proof, not building new machinery.**

Note `emit_element_shape_field_load` (`expr/element_shape_guard.rs`, "run per access inside the fast clone") is the other place a proven field load happens — check whether the coercion in the measurement above comes from there rather than from the generic path, and fix whichever is live.

## Acceptance criteria

1. The read loop above emits **zero** `js_number_coerce` for `r.x` / `r.y`.
2. `--opt-report` still shows `local r -> Ptr<Shape>`; the promotion must not be traded away for the coercion.
3. **The store side agrees with the load side.** A slot read as raw f64 must have been written as raw f64. A mismatch is a silent wrong value, not a crash — this is the direction the change can be quietly wrong in, and it needs a test that writes through every path that can reach the field (constructor, `this.x = v`, `obj.x = v` from outside, `Object.assign`, a setter, a `delete` then re-add).
4. **The GC still traces correctly.** `raw_f64_mask` and the pointer mask must stay disjoint — a slot in both, or in neither, is either a missed root or a scanned number. Run `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800` over the reproducer and assert the `[gc-zeal]` verdict shows non-zero `copying_minors` and `moved_objects` — a clean run that collected nothing proves nothing.
5. A field whose declared type is `number` but which receives a non-number at runtime (Perry does not enforce declared types — CLAUDE.md, Known Limitations) must still produce Node-identical output. Cover: `(p as any).x = "s"`, `= null`, `= {}`, `= 1n`.
6. Protected floors hold on the pinned quiet mini, interleaved, best-of-N, output byte-verified against Node **before** timing.

## Traps

* **`--opt-report` is the instrument.** It names the rule and the promoted local directly; do not infer from emitted calls.
* **Isolate the read loop by subtraction.** Construction is ~12 calls and swamps the ~4 in the read loop. Build a build-only variant and diff the call multisets, as above.
* **`PERRY_NO_AUTO_OPTIMIZE=1` changes nothing here** — verified identical IR with and without, so it is safe for iteration.
* Build `-p perry -p perry-runtime-static -p perry-stdlib-static` and pin `PERRY_RUNTIME_DIR`, or `perry compile` links a stale `.a` and both arms of an A/B behave identically.
* This is the second of three hops in the same mechanism. It does **not** depend on #7766 — the symptom reproduces on the already-promoted local above — but #7766 is what makes it fire on code where the array crosses a function boundary.

## Related

#7766 (rule-1 provenance, the hop before this one), and the element-fetch issue filed alongside this one (the `js_array_get_f64` in the same measurement). Closed context: #7510 (layout side tables), #7149 / #7034 §3 (the element rule), #7480 / #7669 / #7701 (the element-shape guarded clone).

proggeramlug:
Closes #7770.

## What

A `Ptr<Shape>`-proven element-group member (`const r = a[i]`, or a producer pushed into a proven array) claimed **zero** numeric fields (#7034 §3's stand-down), so every declared-`number` field read on it lowered through the Phase-5a checked load — an inline finite check with a cold `js_number_coerce` arm. On the issue's reproducer that is `js_number_coerce x2` per iteration-body, one per field.

Two collector extensions make the numeric proof fire:

1. **Group-wide numeric proof** (`prove_group_numeric_fields`, new in `ptr_shape_numeric.rs`). The E1–E5 containment that licenses the SHAPE proof also closes the group's store universe: while facts hold, every reference to a group's objects is a vetted producer local, a licensed element read, or the inline `new C(...)` at a push site (`a[i].x = v` rides an unlicensed `IndexGet` and disqualifies the array). So the exhaustive-reachable-store proof is discharged once per array root: constructor parameters resolved as the **meet over every push's `new` argument list**, every member's field stores unioned, method parameters resolved through group-merged call sites. Every member carries the group verdict; group integrity keeps it honest (any member failing rule 2 drops every member's fact, claim included).

2. **Numeric-by-construction locals** (`collect_numeric_by_construction_locals`). The reproducer's ctor args are `new P(i, i + 1)` — a mutable loop counter, which the expression proof could not resolve. A local now proves numeric when its every write is number-producing by construction (`let i = 0` + `i++`; optimistic greatest fixpoint like `collect_not_bigint_locals`, but with **no declared-type leaf** — annotations stay untrusted, a no-init `Let` poisons).

Pass 4 (ParamEnv + `prove_numeric_fields`) moved wholesale from `ptr_shape.rs` into the existing `ptr_shape_numeric.rs` child module for the 2000-line gate; `prove_numeric_fields` now takes `&[&[Expr]]` new-arg lists (the single-candidate call passes a 1-element slice, semantics unchanged).

## Acceptance criteria from the issue

1. **Zero `js_number_coerce` for `r.x`/`r.y`** — verified on the reproducer via `--trace llvm`: 4 → 0 call sites; the `ptr_shape_get_number.*` checked-load diamonds are gone, replaced by the Phase-3b bare load (`class_field_get_number.shape_proven_load`). `js_array_get_f64` remains (the separate element-fetch issue).
2. **Promotion intact** — `--opt-report` still shows `local r -> Ptr<Shape>`; `PERRY_REPSEL_DEBUG` now shows `numeric_fields {"x","y"}` (and, as a side benefit of computing the group verdict before the report line, the opt-report numeric count is truthful for members).
3. **Store side agrees with load side** — the store channels are each red-tested: sibling member store, mixed push-site meet, producer `new` args, method-mediated store (group-merged sites), plus group-death on an undeclared-property store. Unit tests in `ptr_shape_group_numeric_tests.rs` (11 tests, sabotage-checked: disabling the group lookup fails them in both directions); behavioral coverage in `test-files/test_gap_repsel_element_group_numeric.ts`, **byte-identical vs Node 26.5.1** including NaN / Infinity / −0 identity through claimed slots.
4. **GC still traces correctly** — no mask or runtime change in this PR (`raw_f64_mask` is class-level and untouched). `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800` on the reproducer: `forced_collections=25 copying_minors=25 moved_objects=4014 loop_polls=2000`, exit 0, correct output; the full gap file survives 30 copying minors byte-identical.
5. **Non-number values at runtime stay Node-identical** — `(w as any).x = null / {} / 1n / true / "s"` all covered in the gap test (each drops the claim through its channel; verified per-case via `PERRY_REPSEL_DEBUG`). One PRE-EXISTING divergence was found nearby and filed as #7773 (`o.x + 1` on a string-holding declared-number field coerces to NaN where Node concatenates; reproduces with `PERRY_PTR_SHAPE_LOCALS=0` and no arrays — not introduced or affected by this PR, A/B'd both arms).
6. **Perf floors hold, and the target workload gets faster.** Pinned quiet M1 mini, interleaved arm pairs, best-of-15, every output byte-verified against Node *before* timing, two independent runs:

   | workload | base | branch | |
   |---|---|---|---|
   | `bench_7770_read` (the issue's loop, scaled) | 101 / 102 ms | 91 / 90 ms | **−10% / −12%** |
   | `batch.ts` | 103.8 ms | 103.8 ms | unchanged |
   | `suite/04_array_read` | 28 ms | 28 ms | unchanged |
   | `suite/09_method_calls` | 9 ms | 9 ms | unchanged |

   **The A/B's subject is verified live**, per CLAUDE.md failure mode 4: on the benchmarked source the base arm emits 4 `js_number_coerce` sites and 12 `ptr_shape_get_number.{plain,coerce,merge}` diamonds, the branch arm 0 of each. Worth recording for the next person — my first attempt at this bench put the array behind a function boundary, which is the **#7766 shape this PR does not address**: it kept all 4 coercion sites on both arms, so its "no regression" would have been vacuous. The array and its read loop have to be in one function for the element facts to apply.

## Validation so far

- `cargo test -p perry-codegen` (full, integration suites included) — green EXCEPT `large_object_barriers::large_local_array_push_inbounds_store_emits_precise_slot_barrier`, which is **pre-existing**: it fails identically on pristine main @423bb4405 (fresh worktree, exit 101) and with `PERRY_PTR_SHAPE_LOCALS=0` (every line of this PR is behind that gate). This suite is nightly/tag-only, so it can sit red on main without a red PR check.
- `ptr_shape` lib tests 96/96; the two pre-#7770 stand-down assertions in `ptr_shape_elements_tests.rs` updated to assert the group verdict.
- Gap suite (`scripts/run_gap_tests.sh`, 519 tests, prebuilt dev compiler) — the parallel sweep on a loaded dev Mac reported 18 status changes; **every one is host noise or pre-existing**, established by serial single-test re-runs plus a `PERRY_PTR_SHAPE_LOCALS=0` A/B (every line of this PR sits behind that gate):
  - 8 cleared as parallel-sweep flakes on serial re-run (`6356_dynamic_parent_mixin_chain`, `gc_alloc_point_no_move`, `language_types_object_part_a`, `learned_inline_sizing`, `logical_and_value_type`, `map_instance_reflection_4576`, `repsel_scalar_replaced_locals`, `http_client_no_redirect_follow`);
  - 10 reproduce **identically with the phase disabled** — 5 sandbox-socket crashes (`fetch_request_from_node_incoming_message`, `http_res_socket_writable_onfinished`, `http_overloads_3226plus`, `http_req_async_iterator`, `net_connect_bound_value`) and 5 parity failures (`events_import_4995`, `gc_rest_argument_rooting`, `gc_same_module_call_argument_rooting`, `specabi_reassign`, `zlib_3285_params`), the known host-local zlib/socket/GC-rooting noise.
  - One genuine **improvement**: `test_gap_iterator_helpers_2874` parity_fail → pass.
- `cargo fmt --check`, `check_file_size.sh`, `addr_class_inventory.py`, `gc_runtime_root_holders.py` — clean.
- `repsel-census` — the gate fails on this branch **exactly as it fails on main** (main runs 31295652054 / 31240304595 are red; the job is not a required context). Root cause is pre-existing: `for (const r of rows)` no longer licenses element-shape facts, filed as #7777 with a minimal repro and per-form A/B against a pristine-main build. On the indexed form this PR strictly improves the census picture (form-A slice: 4 promotions with all three fields claimed, vs claims-empty on main).

## Soundness note (the direction this can be quietly wrong in)

A wrongly-claimed field's read is a bare raw `load double` with no value check. The claim's license is unchanged from Phase 3b — "every reachable store is a number" — this PR only widens WHERE the proof can be discharged (a closed group instead of one unaliased local) and WHAT the expression proof can resolve (by-construction numeric locals). Declared types are still trusted nowhere in the proof; the one deliberate carry-over is the pristine never-stored field (claimed today for plain candidates too), whose `undefined` bits raw-read as NaN — coinciding with `ToNumber(undefined)` — and stay bit-faithful in value contexts.

## Summary by CodeRabbit

* **Performance**
  * Improved numeric handling for array elements, constructors, method calls, loops, and merged data flows.
  * Reduced unnecessary numeric coercions while preserving Node-compatible behavior.

* **Bug Fixes**
  * Improved safety when mixed, invalid, or non-numeric values are written to fields.
  * Preserved correct behavior for special numeric values and adversarial cases.

* **Tests**
  * Added comprehensive coverage for numeric inference, array updates, inheritance, loops, closures, and regression scenarios.

* **Documentation**
  * Updated the documented release version to 0.5.1453.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1452
+**Current Version:** 0.5.1453
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1452"
+version = "0.5.1453"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "perry-ffi",
  "scraper",
@@ -5813,7 +5813,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-commander"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5821,7 +5821,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cron"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "chrono",
  "cron",
@@ -5831,7 +5831,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dayjs"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "chrono",
  "perry-ffi",
@@ -5839,7 +5839,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-decimal"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "perry-ffi",
  "rust_decimal",
@@ -5847,7 +5847,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dotenv"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "perry-ffi",
  "serde_json",
@@ -5855,7 +5855,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ethers"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "perry-ffi",
  "rand 0.10.1",
@@ -5863,7 +5863,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-events"
-version = "0.5.1452"
+version = "0.5.1453"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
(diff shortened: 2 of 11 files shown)


proggeramlug:
## Audit — everything I could check independently holds; stays draft on your own perf-floor condition, correctly

I filed #7770, so this got the full treatment.

### The acceptance criteria, re-verified by hand

**Criterion 1 (the headline)**: on the issue's reproducer, fresh release build, `PERRY_RUNTIME_DIR` pinned — `js_number_coerce` **2 → 0**, `js_array_get_f64` still present (correctly — that is #7771's issue, not this one), output `10000000000` correct.

**Criterion 5 (the soundness direction)**: `test_gap_repsel_element_group_numeric.ts` is **byte-identical to node 26.5.1** on my host, including the NaN / Infinity / −0 identity rows.

**Criterion 4 (GC)**: my run is stronger than the one reported — `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1` depth 800 over the reproducer gives **1,762 copying minors moving 400,014 objects**, exit 0, correct output. That is real relocation pressure over the raw-loading fields.

**The 11 unit tests bite**: neutering `prove_group_numeric_fields` (return `Default::default()`) fails **6 of 11** — and the 5 that still pass are the negative-direction tests, which *should* pass on an empty verdict. That split is what a two-directional suite looks like.

### The three pre-existing-failure claims all check against my records

- `large_object_barriers::…precise_slot_barrier` — matches the exact residual set from my #7708/#7730 baseline work.
- **#7773** (declared-number field holding a string coerces where Node concatenates) — filed, open, correctly A/B'd with `PERRY_PTR_SHAPE_LOCALS=0` to show it is not this PR's.
- **#7777** (repsel-census red on main; `for (const r of rows)` no longer licensing element facts) — filed with a per-form A/B. Finding the *census gate itself* red on main and refusing to absorb that into this PR is the right move.

### The design judgement worth endorsing explicitly

The soundness note is exactly right about where this can be quietly wrong: a wrongly-claimed field's read is a bare `load double` with no check. The license is unchanged — "every reachable store is a number" — and the PR only widens **where** the proof discharges (a closed group instead of one unaliased local) and **what** the expression proof resolves (by-construction numeric locals, with **no declared-type leaf** — annotations stay untrusted, which is the constraint #7770 stated as non-negotiable).

Group integrity ("any member failing rule 2 drops every member's fact, claim included") is the collapse rule that keeps the group proof from outliving its premises. And the by-construction locals collector poisoning on a no-init `Let` is the conservative default in the right place.

### Not merging yet, per the PR's own gate

Criterion 6 (perf floors on the bench mini) and the full gap run are the two open items, and holding as draft until they land is the discipline that #7726 taught — targeted-green is not validated. Post the floors and I'll do the merge pass.

proggeramlug:
Merging per the full audit above, now that the PR is undrafted. Delta since the audit:

- Rebased onto v0.5.1452 main clean (no conflicts), version bumped to 0.5.1453.
- The dark-tests gate flagged `test_gap_repsel_element_group_numeric.ts` as unregistered for the gc-repsel corpus. Registering it was the wrong fix: the matrix's own liveness gate proved the scavenge arm **INERT** on it (`counter=0`, 0/1 cells live) — its read loop is raw f64 loads with near-zero allocation, so a green cell would be green for the wrong reason. Added the reasoned exclusion in `check_test_registration.py` instead; moving-GC coverage for this feature is the reproducer run already recorded above (1,762 evacuating minors, 400,014 objects moved, exit 0).
- Local gate: 21/21 steps ran; one `perry-runtime --lib` red at load average ~15 that re-ran clean 2/2 (2,046 passed) — the known load transient.

Closes #7770.

Issue #7170, opened 2026-08-01
repsel: rule-1 provenance for unbound allocations — the 506 are 96% record literals, and Perry's own CJS IIFE is the wall (#7152 follow-up)


proggeramlug:
Follow-up to #7152, which measured that **506 of 746** `Ptr<Shape>` candidates in
dependency JS die at rule 1 (allocations never bound to a local) and concluded
"the dep-JS lever is rule 1, and its fix is anchoring the proof to an allocation
site rather than to a binding."

I re-measured on a second, independent slice of the same corpus and then
classified the population by **mechanism** rather than by position. The headline
number replicates. The *reading* of it does not survive, in three places, and the
correction changes what should be built first.

**Short version: the dependency-JS wall is not a missing soundness argument. It
is Perry's own CommonJS wrapper — for the third time.** #7139 found the wrap
preamble arming the rule-5 barrier in 100% of CJS modules; #7152 found `__cjs_module`
accounting for 100% of the rule-2 bare references; this issue finds that the
wrap's **IIFE** turns every module-level function into a closure, which makes
#7107's return-shape provenance *structurally unreachable* in CommonJS. Same
self-inflicted class, same kind of fix: recognition, not relaxation.

---

## 1. Method, and what in it is a proxy

**Compiler arm.** `wt-7148` @ `6a3b5bff8` — the #7149 branch tip, tracked tree
clean, `perry-dev` binary. *Caveat, stated because CLAUDE.md requires it:* the
binary's mtime is ~90 s **before** the commit timestamp (built, then committed),
and I reused it rather than rebuilding — local disk fell from 26 GB to 12 GB free
during the session from other agents' builds. The tracked tree is clean at that
commit and the element rule is present in the binary; the risk is low for a
static census, and it is nil for the *code-derived* findings in §5–§6, which are
proofs from source, not samples.

**Corpus.** `real-apps/scriptc/node_modules`, every third `__esModule` `.js`
file by sorted path — **197 selected of 589**, 196 producing a report.
`perry compile <m>.js --opt-report=json --no-link`, each module copied to its own
directory so it compiles natively rather than routing to the removed V8 fallback.

**This is not #7152's exact 180-module selection**, and I could not reconstruct
it. Treat the two as two samples of one corpus. They agree on the headline share
and disagree on the position mix — see §3.

**Two measurement points, and they are not the same population.** The
`--opt-report` buckets are measured on **HIR after** `state_desugar`,
[...]

Issue #7152, opened 2026-07-31
repsel: dependency JS is walled by rule 1 (unbound allocations), not containment — 506 of 746 candidates


proggeramlug:
#7139 recovered rule-5 eligibility for 84 of 85 CommonJS dependency modules and
recorded that of the ~103 `Ptr<Shape>` candidates it freed, **100 were
immediately re-denied by rule 2** — reading as "containment is the wall in
dependency JS". Track D (#7034 §3, PR #7149) was scheduled on that basis.

**Measured, it is not containment.** Two compiler arms — #7139 alone, and
#7139 + #7149 — over **180** real `__esModule` CJS modules from
`real-apps/scriptc/node_modules` (`perry compile m.js --opt-report --no-link`,
180/180 compiled in both, every count identical between arms):

| bucket | candidates |
|---|---:|
| **rule 1 — allocated in expression position, never bound to a local** | **506** |
| rule 5 — module barrier still armed | 99 |
| rule 2 — bare reference | 130 |
| rule 2 — call argument | 5 |
| rule 2 — array element | **1** |
| rule 2 — closure capture | 1 |
| rule 2 — undeclared property | 1 |
| **total candidates** | 746 |
| selected / consumed | 3 / 0 |

Two things fall out.

### 1. Rule 1 is the wall, and its shape is known

The 506 unbound allocations break down by position:

| position | count |
|---|---:|
| constructor argument | 182 |
| statement | 162 |
| call argument | 84 |
| return | 64 |
| array element | 8 |
| initializer | 6 |

`new X(…)` written directly into an argument, a statement, or a return, never
bound to a `let`/`const`. Rule 1 needs a local to anchor the proof to, and
transpiled/minified code does not create one. This is the `.map(x => ({…}))` /
`return {…}` idiom the report already names — at 68% of all candidates it is
larger than every containment position combined by a factor of ~3.5.

### 2. The rule-2 "bare reference" denials are Perry's own scaffolding

All 130 are the local `__cjs_module`. That is the CommonJS wrapper's own
binding, not user code — the same family #7139 was narrowing. Whether it is
worth exempting depends on whether `__cjs_module`'s object is ever a useful
`Ptr<Shape>` candidate at all; if it is not, it should not be a *candidate*,
and the 130 denials are noise in every report.

### Caveat on the corpus

This is not #7139's exact 85-module set (that selection was
"self-contained `__esModule` leaves"; this one is every third `__esModule` file
[...]

PR #7778, opened 2026-08-10
perf(repsel): element-shape loop clone serves the element-binding form through function boundaries (#7766)

### LINKED ISSUE #7766: repsel: rule-1 provenance does not survive a function boundary — a typed `P[]` parameter still reads fields by name (first-party half of #7152/#7170)

proggeramlug:
Closes #7766.

## The verified framing (the issue asked for this first)

The issue's mechanism 1 said the clone "is admitted but its body does not specialize element field reads" and asked that framing to be verified before building on it. Measured on current `main` (423bb4405):

* **The direct spelling is already served.** `s += ps[i].x + ps[i].y` through a typed `P[]` parameter gets a versioned clone whose deref block `cond_br`s INTO the fast preheader (the #7701 entry test), with offset loads and no calls in the fast body. Runtime confirms: 0.10s vs node 0.07s on 200k×200 (dev profile). The by-name/guard/coerce calls the issue counted in (A)'s IR belong to the **slow clone** — mandatory fallback IR, dead at runtime for an honest caller.
* **The binding spelling was the real gap.** `{ const r = ps[i]; s += r.x + r.y }` — and `for (const p of ps)`, which desugars to exactly that shape — got no clone at all: 2.2× and 7.8× slower than the direct spelling respectively. Two independent causes:
  1. the matcher admitted only a single-statement accumulator body, and
  2. the for-of desugar mints its counter as `Number(0.0)`, which `collect_integer_let_ids` does not seed, so the desugared counter never joined `integer_locals`, never got a canonical i32 slot, and **every** i32-counter loop optimization silently declined the `for…of` spelling of loops it served in indexed form.

## Mechanism decision (all three evaluated)

**1. Extend the element-shape guarded clone — chosen.** The guard establishes the fact at runtime (`js_array_ensure_element_shape` verifies every element's class; the declared type only names the class id to check against), so it is sound for parameters *by construction* — exactly the "establish, don't assume" bar the issue sets. With the framing corrected, the remaining work was matcher coverage, not read specialization, which makes this by far the smallest correct change (~120 lines, no new ABI, one clone per loop as today).

**2. Clone-and-route at the call site** — needs per-signature callee clones (code size), call-site proof plumbing across modules, and serves only calls whose arguments already carry proofs — which #7170 measured as rare in dependency JS (91.6% of its rule-1 population sits in closures where no proof exists). A runtime guard is still needed for unproven callers, i.e. mechanism 1 is a prerequisite of 2, not an alternative.

**3. Entry guard + specialized body** — pays the guard on every call even when the loop is cold, duplicates the whole function body instead of one loop, and the guard's O(n) first-scan can be wasted work if the loop is never reached. The loop-preheader placement already implemented is strictly better positioned.

## What this PR changes

1. **Matcher** (`stmt/element_shape_loop.rs`): admits an optional leading `const r = arr[counter]` element binding whose every use is a tracked `r.field` read. The binding's source array participates in the one-array-per-loop rule, and any other use of `r` — bare reference, mutable binding, non-counter index, second array — declines (each shape has a red test).
2. **Lowering**: the fast clone **skips the binding's `Let` entirely** — lowering it would emit the element-read tier's calls, fail the call-free scan, and silently delete the clone (#7690's failure shape). `r.field` reads are answered by the loop fact (`element_binding_local` on `ElementShapeLoopFact`); the slow clone keeps the full body, so the side-exit protocol (re-execute the current iteration, nothing committed) is unchanged.
3. **for-of desugar** (both emission sites: `lower/stmt_loops.rs` module-init, `lower_decl/body_stmt.rs` function bodies): the minted counter init is `Integer(0)` instead of `Number(0.0)` — the literal kind the integer-local collector seeds on, and the same shape a user-written `let i = 0` produces. (for-in deliberately untouched.)
4. **`--opt-report`**: the clone records a `Ptr<Shape>` selection — and per-read consumption — when and only when the deref block branches into the fast clone (an emitted-but-deleted clone records nothing: a gate must assert its subject was live).

## Acceptance criteria, measured

1. ✅ `--opt-report` on the binding case now prints `Ptr<Shape> 1 selected … 1 of those selections were CONSUMED`, naming the binding local; the fast clone emits no `js_object_get_field_by_name_f64`, no `js_typed_feedback_class_field_get_guard`, no `js_number_coerce` (asserted by IR census with the #7701 entered-not-just-emitted test).
2. ✅ Case (B) untouched (static `ptr_shape_elements` route unchanged; full element_shape suite green — 28 tests).
3. ✅ Soundness is tested, not argued: `test-files/test_gap_repsel_element_shape_param_binding.ts` runs mixed arrays, a subclass with extra fields, same-shape plain literals, a hole, mid-loop mutation, growth between calls, and a fake array-like through the parameter boundary in both spellings — byte-identical to node 26.5.1 (the pinned oracle).
4. ⏳ Corpus rule-1 bucket: measured on the #7152/#7170 corpus with both arms; numbers in the first comment. Honest expectation set by #7170's own tier data: the corpus bucket is dominated by unbound-allocation sites (40.4% nested-literal components, 16.9% opaque-runtime-callee arguments) that **no read-side mechanism can serve**; this PR serves the T3 (parameter-shape) slice's in-loop reads. The first-party half — the case the issue was filed on — is what moves wholesale.
5. ⏳ Protected floors: interleaved best-of-5 on the pinned mini, outputs byte-verified against node before timing; table in the first comment.

## Perf (probes, dev profile, 200k elements × 200 passes, local M1)

|---|---:|---:|---:|
| `const r = ps[i]; s += r.x + r.y` (param) | 0.22s | **0.07s** | 0.07s |
| `for (const p of ps) s += p.x + p.y` (param) | 0.78s | **0.07s** | 0.07s |
| `s += ps[i].x + ps[i].y` (param, already served) | 0.10s | 0.10s | 0.07s |

## Known-unrelated red

`large_object_barriers::large_local_array_push_inbounds_store_emits_precise_slot_barrier` fails identically on clean `main` @ 423bb4405 (verified in a pristine worktree); the fixture is hand-built HIR with no `for…of` and no element-shape loop, so it is structurally unreachable by this diff. Integration suites don't run per-PR (#5960), which is how it sits red on main.

## Heads-up

#7771 (`wt-7771`, in flight) specializes bare `a[i]` fetches inside the same clone and touches the same three files — whichever lands second will need a small rebase.

## Summary by CodeRabbit

* **Performance**
  * Improved optimization of array-element loops passed through function parameters.
  * Added support for more array shapes and safer handling of mixed, sparse, mutated, and array-like inputs.
  * Improved integer loop-counter recognition.

* **Bug Fixes**
  * Prevented optimization when indexed bindings cannot be safely verified.
  * Preserved expected error behavior for unsupported array contents.

* **Tests**
  * Added extensive regression coverage for optimized and fallback execution paths.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1455
+**Current Version:** 0.5.1456
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1455"
+version = "0.5.1456"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "perry-ffi",
  "scraper",
@@ -5813,7 +5813,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-commander"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5821,7 +5821,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cron"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "chrono",
  "cron",
@@ -5831,7 +5831,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dayjs"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "chrono",
  "perry-ffi",
@@ -5839,7 +5839,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-decimal"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "perry-ffi",
  "rust_decimal",
@@ -5847,7 +5847,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-dotenv"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "perry-ffi",
  "serde_json",
@@ -5855,7 +5855,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ethers"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "perry-ffi",
  "rand 0.10.1",
@@ -5863,7 +5863,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-events"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "perry-ffi",
  "perry-runtime",
@@ -5871,14 +5871,14 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-exponential-backoff"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "perry-ffi",
 ]
 
 [[package]]
 name = "perry-ext-fastify"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "bytes",
  "http-body-util",
@@ -5896,7 +5896,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-fetch"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "bytes",
  "lazy_static",
@@ -5909,7 +5909,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-http"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "bytes",
  "h2",
@@ -5933,7 +5933,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ioredis"
-version = "0.5.1455"
+version = "0.5.1456"
 dependencies = [
  "lazy_static",
  "perry-ffi",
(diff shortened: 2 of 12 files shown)


proggeramlug:
## Validation results (criteria 4 + 5, plus the #7780-family hazard check)

### Protected floors (criterion 5) — pinned quiet mini, interleaved best-of-5, outputs byte-verified vs node before timing

| bench | base | fix | ratio |
|---|---:|---:|---:|
| churn | 0.42 | 0.42 | 1.000 |
| churn_alloc | 0.37 | 0.37 | 1.000 |
| push_cls | 0.35 | 0.35 | 1.000 |
| push_num | 0.14 | 0.14 | 1.000 |
| churn_read | 0.02 | 0.02 | 1.000 |
| cycles | 0.19 | 0.19 | 1.000 |
| deeplist | 0.24 | 0.24 | 1.000 |
| tree | 1.63 | 1.63 | 1.000 |
| tree_wide | 2.10 | 2.10 | 1.000 |
| retain | 0.53 | 0.53 | 1.000 |
| retain_wide | 1.09 | 1.09 | 1.000 |
| fib40 | 0.39 | 0.39 | 1.000 |

The uniform 1.000 is *verified honest*, not vacuous: the two arms' binaries differ (`cmp` on four of them), and a release-profile probe that DOES contain the changed shape separates them decisively — `for…of` over a parameter array: base 0.83s, fix **0.09s** (9.2×), byte-identical output. None of the 12 protected benches contains a `for…of` or binding-form read loop (grep-verified), so identical times are the expected structural result.

### Dependency-JS corpus (criterion 4) — #7152/#7170 methodology, every third `__esModule` module (409 selected), release binaries, both arms

Paired over the **393 modules that compiled in both arms** (selection lists identical; the remaining 2/3 were 60s-timeout flakes under load):

| ptr-shape bucket | base | fix |
|---|---:|---:|
| rule 1 (provenance) | 2606 | 2606 |
| rule 1 — already served by return-shape | 30 | 30 |
| rule 2 (containment) | 42 | 42 |
| rule 5 (module-wide barrier) | 151 | 151 |
| class admission | 6 | 6 |
| selected / consumed | 1 / 56 | 1 / 56 |

**Zero per-module differences.** Two readings, both load-bearing:

1. The dep-JS rule-1 wall does not move — as the PR body predicted from #7170's tier census, it is unbound-allocation territory (nested-literal components, `defineProperty` descriptors, closure regions) that no read-side mechanism can serve. The first-party half — the case #7766 was filed on — is what this PR moves (to node parity). The dep-JS half stays with #7152/#7170's own mechanisms.
2. It is also a 393-real-module no-regression proof for the for-of desugar change: not one ptr-shape outcome shifted anywhere in the corpus.

(Absolute numbers are ~2× #7170's because this selection is 409 modules vs its 197 — same corpus, denser slice.)

### The #7780-family hazard (third clone-deletion mode), checked here

PR #7780's session found that the element binding's lexical-death shadow-slot clear is a real `js_shadow_slot_set` call in the call-fallback shadow mode, which silently deletes a call-free clone. This PR's approach is structurally immune: the fast clone lowers `&body[1..]`, so the binding never enters `clear_loop_body_shadow_slots`' statement walk — no clear is emitted in any mode. Verified empirically under `PERRY_INLINE_SHADOW_SLOT=0` (the call-fallback mode): the deref block still `cond_br`s INTO the fast clone and the fast blocks contain zero calls, for both the binding and `for…of` spellings.

### Overlap with #7780

#7780 (branch `repsel/7771-element-fetch-clone`) independently implements the same matcher/fact core. Unique to this PR: the for-of desugar counter fix (`Integer(0)` — the 9–11× `for…of` win), the `--opt-report` select/consume wiring (criterion 1), the #7766 parameter-boundary soundness gap test, and the corpus A/B. Unique to #7780: the `emit_shadow_slot_clear` defensive skip (moot under this PR's body-slicing but harmless belt-and-braces), and the #7775/#7776 side-find gap tests. The two need reconciling before either merges — maintainer's call which lands first; the other rebases to its unique parts.

proggeramlug:
**Status: ready for review.** The gap suite (`scripts/run_gap_tests.sh`, ~500 parity tests) is still running locally at 218/~500 and is not blocking this — the change's own parity coverage is already proven: the new `test_gap_repsel_element_shape_param_binding.ts` is byte-identical to node 26.5.1, and the 393-module corpus A/B found zero output or outcome differences. I will post the gap result here when it lands; if it surfaces anything, I will fix it on this branch.

proggeramlug:
Audit complete — merging. This PR raced #7780 (both implemented the element-binding matcher independently); the rebase keeps main's landed matcher and grafts this PR's unique delta, which is the part that actually carries the clone across function boundaries:

**What survives (and was verified):**
- **for-of desugar counters lower as `Integer(0)`** (both HIR emission sites). A/B on the gap test, both arms verified applied: with `Number(0.0)` the `totalOf` (for-of) function loses the clone while `totalBinding` keeps it; with `Integer(0)` both get it. This change alone is what admits the `for…of` spelling.
- **`--opt-report` select/consume rows** for the clone, recorded only when the deref block branches into the fast clone — a served parameter-array loop no longer reads as an unserved rule-1 wall (the mis-reading #7766 was filed on).
- **Two tests #7780's suite lacked**: the parameter-array positive (asserted admitted + `cond_br`-entered + fast-region call-free via CFG walk, not merely emitted) and the non-counter-index negative. 30/30 element_shape tests.
- **End-to-end IR** on `test_gap_repsel_element_shape_param_binding.ts`: both `totalBinding` and `totalOf` clones are `cond_br`-entered with **zero non-intrinsic calls** in their fast regions.
- **Gap test byte-identical to node 26.5.1** (19 lines: honest caller, mixed arrays, subclass, literals, hole, mid-loop mutation, fake array-like), identical under zeal+quarantine depth 800 (9 copying minors, 15,843 objects moved, exit 0).
- **Full `cargo test -p perry-codegen`**: 1,279 passed, 2 failed — the same #7708 residue confirmed on pristine main today, zero new.
- **Registration repair folded in**: `gc_repsel_matrix.sh`'s own manifest pre-check does not consult `check_test_registration.py` exclusions, so #7774's excluded fixture broke every matrix invocation (exit 3) since it merged. Both new gap files are now registered in the corpus and the stale exclusion is deleted; the matrix's #7255 arm-level liveness gate guards vacuity. Element-slice matrix run: 4/4 cells byte-exact vs node.
- Local gate 21/21 green.

Closes #7766.

Issue #7776, opened 2026-08-10
parity: heterogeneous element via `as any` — `s += r.x + r.y` yields NaN where node string-concatenates


proggeramlug:
## Symptom

After a different-shape element is stored into a `P[]` (through `as any`), a read loop that does `s += r.x + r.y` yields `NaN` where node string-concatenates.

## Minimal reproducer

```ts
class P { x: number; y: number; constructor(x: number, y: number) { this.x = x; this.y = y; } }
class Q { x: string; y: string; constructor(x: string, y: string) { this.x = x; this.y = y; } }
function run(): number {
  const a: P[] = [];
  for (let i = 0; i < 10; i++) a.push(new P(i, i + 1));
  (a as any)[4] = new Q("z", "w");
  let s = 0;
  for (let i = 0; i < a.length; i++) { const r = a[i]; s += r.x + r.y; }
  return s;
}
console.log(run());
```

node 26.5.1: `16zw1113151719` (sum concatenates through the Q at index 4). Perry: `NaN`.

## What is established

- Pre-existing: identical output on pristine `main` @ 423bb4405 (v0.5.1448) and on the #7771 branch (A/B'd both arms).
- Shape-wise this looks like the `+` lowering claiming a numeric proof for `r.x` off the declared element type and coercing (`js_number_coerce("z")` → NaN) where spec `+` should have dispatched to string concat. Declared types are hints, not layout facts (CLAUDE.md, Known Limitations) — the annotation is lied to via `as any`, but node semantics still apply.
- The #7771 element-shape clone is NOT involved: its runtime guard correctly declines this array (heterogeneous), and the divergence comes from the generic/slow path that then runs.

Issue #7775, opened 2026-08-10
miscompile: an uncalled function containing `new Proxy(arr, {})` makes an unrelated function's indexed read loop run zero iterations


proggeramlug:
## Symptom

The mere PRESENCE of `new Proxy(rawArray, {})` in a function that is **never called** makes an *unrelated* function's `for (let i = 0; i < a.length; i++)` read loop run **zero iterations**. Wrong answer, no diagnostic.

## Minimal reproducer

```ts
class P { x: number; y: number; constructor(x: number, y: number) { this.x = x; this.y = y; } }
function build(n: number): P[] {
  const a: P[] = [];
  for (let i = 0; i < n; i++) a.push(new P(i, i + 1));
  return a;
}
function readLoop(): number {
  const a = build(10);
  let s = 0;
  for (let i = 0; i < a.length; i++) { const r = a[i]; s += r.x + r.y; }
  return s;
}
function neverCalled(): number {
  const raw = build(10);
  const a: P[] = new Proxy(raw, {}) as any;
  let s = 0;
  for (let i = 0; i < a.length; i++) { const r = a[i]; s += r.x + r.y; }
  return s;
}
console.log(readLoop());
```

node 26.5.1: `100`. Perry: `0`.

## What is established

- Delete `neverCalled` (or just its `new Proxy` line's function) → Perry prints `100`. Bisected from a 10-case battery: of `delete a[i]`, `a[i] = undefined`, a different-class store, `class X extends Array`, `a.length = 5`, and the proxy — **only the proxy variant poisons the module** (the others are all byte-identical to node in the same file).
- The poisoning is **static**: `neverCalled` is never invoked. `readLoop`'s loop body itself runs zero times (a `leak = r` side effect in a sibling case never fires), so the loop condition — not the sum — is what breaks.
- The proxy loop ITSELF is correct when it is the only such function in the module (prints 100, byte-identical to node).
- Reproduces on pristine `main` @ 423bb4405 (v0.5.1448) and is byte-identical on the #7771 branch (A/B'd both arms) — long pre-existing, not introduced by any recent element-shape work.
- perry-dev profile, `PERRY_NO_AUTO_OPTIMIZE=1`, runtime pinned via `PERRY_RUNTIME_DIR`.

Found while building the #7771 criterion-3 hazard battery — the proxy hazard case had to be moved to its own compilation unit to keep the other nine cases meaningful.

PR #7780, opened 2026-08-10
repsel: admit the element-binding body form into the element-shape versioned loop clone (#7771)

### LINKED ISSUE #7771: repsel: a proven element fetch is still a runtime call — specialize a[i] inside the element-shape guarded clone

proggeramlug:
Closes #7771.

## Re-scoping finding (the issue's "first thing to establish")

The issue's premise — "the guarded-clone machinery is already entered here" — is **false** for the reproducer, in exactly the way the issue told us to check for:

- `lower_element_shape_versioned_for`'s matcher only admitted the single-statement `acc = arr[j].field` body. The reproducer's two-statement body (`const r = a[i]; s += r.x + r.y`) never got a clone at all — no `element_shape.*` block exists in its IR on main.
- The measured per-iteration `js_array_get_f64` (v0.5.1448) is **already gone at runtime on current main**: the fetch lowers via the bounded-index path, and #7751's producer-side store-back repairs the binding after growth, so the inline fast arm runs (timed: 50M fetches in 0.09s user vs 2.65s with the slow arm forced).

What remained of the issue's substance: the call was still *emitted* with a per-iteration 3-load guard diamond + hole-select, two `js_number_coerce` diamonds on the field reads, and a per-iteration safepoint poll — and the engine plan's third leg (element fetch under the invariant guard) was unimplemented for the shape real read loops are written in.

## The fix (upstream, as the issue prescribes)

Widen the versioned-loop matcher to admit the **element-binding body form**:

```ts
```

The binding is **virtual inside the fast clone**: its `Let` emits nothing, and every `r.field` read lowers through the existing `ElementShapeLoopFact` to a bare element load + residual per-element check under the preheader guard. Three small mechanical pieces:

- `ElementShapeLoopFact.element_binding` + a `LocalGet` arm in `element_shape_loop_fact_for_property_get` (the single entry point all three consumers already share, whose docs anticipated this widening).
- A skip at the top of `lower_let` for the active fact's binding — lowering the generic `IndexGet` there would put a runtime-call diamond inside the clone and **delete** it via the call-free admission scan (#7690's lesson).
- A skip in `emit_shadow_slot_clear` for the virtual binding's lexical-death clear: in the call-fallback shadow mode that clear is a real `js_shadow_slot_set` call — the clone's only call — and would silently delete the clone in that mode. Skipping can only over-root; the slow clone (lowered after the fact is popped) keeps its clear.

`const`-only by design: a `var` binding is function-scoped and observable after the loop, where the skipped `Let` would leave the slot stale. The slow clone still binds `r` generically, and the side exit re-runs the current iteration there before any effect commits.

## Acceptance criteria

1. **Zero `js_array_get_f64` for `a[i]` in the read loop** — the fast clone contains **no calls at all** (the coerce diamonds and the poll are gone too); the generic fetch survives only in the slow clone, as it must.
2. `--opt-report` still shows `local r -> Ptr<Shape>` (collector untouched).
3. **Guards hold, verified against node 26.5.1 byte-for-byte** (`test-files/test_gap_7771_*.ts`, kept as permanent gap tests): hole via `delete` (throws like node), element set to `undefined` (throws like node), Array subclass behind a `P[]` binding (brand guard declines → 100), shrunk length (25), growth far past inline capacity (10^10), mid-loop `length` write (matcher declines → 9), `let` binding (declines → 100), escaping binding (declines → 100 + side effect), proxy-wrapped array (brand declines → 100, own file — see below).
4. **Clone entered, not merely emitted**: `assert_fast_clone_is_entered` on both new positive IR tests, plus four decline-shape tests (mutable binding, bare-value escape, untracked field, second array). Sabotage-verified: disabling the `Let` skip turns both positives red.
5. **GC zeal clean**: `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800` on the reproducer → correct output, exit 0, `[gc-zeal] forced_collections=1762 copying_minors=1762 moved_objects=400014` — the array moved 1762 times under the guard and the preheader's post-call re-derivation held.
6. **Pinned quiet mini, interleaved best-of-5, output byte-verified before timing** (100k-element build + 500 read passes = 50M fetches): base 0.07–0.08s user → **0.05s** user, stable across all five rounds. Existing floors exercise the single-statement shape, whose lowering is byte-asserted unchanged by the pre-existing IR tests.

## Other validation

- Full `cargo test -p perry-codegen` (lib + all integration suites): only failure is `large_object_barriers::large_local_array_push_inbounds_store_emits_precise_slot_barrier`, which fails **identically on pristine main @ 423bb4405** (known red baseline). All 44 element/clone tests green, grouped and solo.
- 12 existing gap tests over adjacent shapes (array subclass declared-base, class-ref numeric arrays, GC rooting family, async index calls, closures) byte-identical vs node with this branch's compiler.
- `cargo fmt --check`, `check_file_size.sh`, `addr_class_inventory.py` clean. Rebased onto current `main`.

## Pre-existing bugs found and filed while building the hazard battery (both A/B-proven identical on pristine main)

- **#7775** — an *uncalled* function containing `new Proxy(arr, {})` makes an unrelated function's indexed read loop run zero iterations (module-wide miscompile; why the proxy gap test lives in its own file).
- **#7776** — heterogeneous element via `as any`: `s += r.x + r.y` yields `NaN` where node string-concatenates (slow/generic path; this PR's guard correctly declines that array, and its slow clone reproduces main's behavior exactly).

No version bump per contributor convention; changelog fragment follows as `changelog.d/<PR>-element-binding-clone.md`.

## Summary by CodeRabbit

* **New Features**
  * Optimized read loops that assign array elements to a local binding before accessing their fields.
  * Preserved correct behavior for array holes, undefined elements, subclasses, proxies, large arrays, and changing lengths.
  * Automatically falls back to the general execution path when optimization is unsafe.

* **Bug Fixes**
  * Prevented incorrect handling of escaped, mutable, or unsupported element bindings.

* **Tests**
  * Added regression coverage for optimized and fallback loop behavior.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1453
+**Current Version:** 0.5.1454
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1453"
+version = "0.5.1454"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1453"
+version = "0.5.1454"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1453"
+version = "0.5.1454"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1453"
+version = "0.5.1454"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1453"
+version = "0.5.1454"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1453"
+version = "0.5.1454"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1453"
+version = "0.5.1454"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1453"
+version = "0.5.1454"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1453"
+version = "0.5.1454"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1453"
+version = "0.5.1454"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5692,7 +5692,7 @@ dependencies = [
 
 [[package]]
 name = "perry-container-compose"
-version = "0.5.1453"
+version = "0.5.1454"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -5721,14 +5721,14 @@ dependencies = [
 
 [[package]]
 name = "perry-container-e2e"
-version = "0.5.1453"
+version = "0.5.1454"
 dependencies = [
  "anyhow",
 ]
 
 [[package]]
 name = "perry-diagnostics"
-version = "0.5.1453"
+version = "0.5.1454"
 dependencies = [
  "serde",
  "serde_json",
@@ -5736,7 +5736,7 @@ dependencies = [
 
 [[package]]
 name = "perry-dispatch"
-version = "0.5.1453"
+version = "0.5.1454"
 
 [[package]]
 name = "perry-doc-fixture-my-bindings"
@@ -5747,7 +5747,7 @@ dependencies = [
 
 [[package]]
 name = "perry-doc-tests"
-version = "0.5.1453"
+version = "0.5.1454"
 dependencies = [
  "anyhow",
  "clap",
@@ -5762,7 +5762,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-ads"
-version = "0.5.1453"
+version = "0.5.1454"
 dependencies = [
  "block2",
  "objc2",
@@ -5772,7 +5772,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-argon2"
-version = "0.5.1453"
+version = "0.5.1454"
 dependencies = [
  "argon2",
  "perry-ffi",
@@ -5780,7 +5780,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-axios"
-version = "0.5.1453"
+version = "0.5.1454"
 dependencies = [
  "perry-ffi",
  "reqwest",
@@ -5789,7 +5789,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-bcrypt"
-version = "0.5.1453"
+version = "0.5.1454"
 dependencies = [
  "bcrypt",
  "perry-ffi",
@@ -5797,7 +5797,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-better-sqlite3"
-version = "0.5.1453"
+version = "0.5.1454"
 dependencies = [
  "perry-ffi",
  "rusqlite",
@@ -5805,7 +5805,7 @@ dependencies = [
 
 [[package]]
 name = "perry-ext-cheerio"
-version = "0.5.1453"
+version = "0.5.1454"
 dependencies = [
  "perry-ffi",
  "scraper",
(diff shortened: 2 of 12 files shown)


proggeramlug:
Audit complete — merging.

**Premise correction recorded first:** #7771's original claim ("element fetch calls into the runtime inside the clone") was half-stale by audit time — #7751 had already removed the runtime call on the fetch itself. What remained was that the two-statement body form (`const r = a[i]; use(r.x)`) was never ADMITTED into the element-shape versioned clone: the `let` binding's shadow-slot clear read as a call-fallback and vetoed admission. This PR admits the element-binding form via a virtual binding.

**Verified:**
- 28/28 `element_shape` codegen tests post-rebase onto v0.5.1453 (which landed #7774 on the same codegen area — no interaction).
- Sabotage bites: disabling the Let-skip (`false &&` inside the `.any(...)` at `stmt/let_stmt.rs:81`) fails exactly the 2 positive admission tests.
- **End-to-end IR** (`--trace llvm` on `test_gap_7771_element_binding_read_loop.ts`): 3 clone sites, each `cond_br`-entered from its `element_shape.loop.preheader.deref` block, and a CFG walk of each fast region finds **zero non-intrinsic calls**. Detector control: the same regex finds 192 `js_*` calls in 137 other blocks of the module — the zero is real, not a blind checker (#7690's deletion mode is what this guards).
- **Gap battery vs node 26.5.1**: both new tests byte-identical (`element_binding_read_loop` 9 lines; `proxy_array_read_loop` exercises the Proxy negative case through real traps).
- **Moving-GC arm live**: `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1 DEPTH=800` → 1,765 copying minors, 415,625 objects moved, exit 0, output identical.
- **Full `cargo test -p perry-codegen --no-fail-fast`**: 1,277 passed, 2 failed — both (`large_local_array_push_inbounds_store_emits_precise_slot_barrier`, `typed_f64_receiver_method_clone_raw_loads_after_composed_guards`) reproduce identically on pristine main @2e5bf4434, i.e. the known #7708 residue, **zero new failures**.
- Local gate 21/21 steps; one `perry-runtime --lib` red at load ~12 re-ran clean 2/2 (2,046 passed) — the known load transient.

Closes #7771.

Issue #7781, opened 2026-08-10
gc: the seeded schedule never arms the poll word — SCHEDULE_RATE=1 sees 6 safepoints where zeal sees 9,648 (blocks #7741)


proggeramlug:
## Found while testing #7741's precondition, and it blocks #7741

The maintainer's condition for #7741 (removing `PERRY_GC_ZEAL`) is that the `SCHEDULE_RATE=1` equivalence be **demonstrated** on one of the four reproductions zeal closed this week. Tested on #7606's rooting tests, same binary (main @ v0.5.1450), same quarantine settings:

| instrument | rest_argument | same_module_call |
|---|---|---|
| `PERRY_GC_ZEAL=1` | 379 collections, 115,536 moved, **9,648 loop polls** | 749 collections, 231,012 moved, **19,248 polls** |
| `PERRY_GC_SCHEDULE_SEED=… RATE=1` | **safepoints=6**, 6 collections | **safepoints=6**, 6 collections |

`RATE=1` is the "collect at every opportunity" end of the dial. It saw six opportunities.

## Cause

#7735 collapsed the loop back-edge poll's no-work path to one load of `PERRY_GC_POLL_ARMED`. The word is kept armed for zeal (`resolve_poll_seed` retains the startup seed when `gc_zeal_enabled()`), and `ZealGuard` arms/releases it for tests. **Nothing arms it for the schedule mode** — so with schedule-on/zeal-off, every back-edge poll takes the disarmed fast path, `js_gc_loop_safepoint` is never entered, and the loop-safepoint bypass #7317 added sits behind a gate that never opens. The schedule's only surviving entry point is the microtask-pump boundary: six per run on this test.

This is #7735's own design note played out: the arming word is a *counter of reasons*, and a new reason (a mode whose collection decision lives inside the safepoint) must register itself. The schedule mode was written before the word existed and was merged across it (#7317's rebase) without gaining an arm.

## Fix

- `resolve_poll_seed` keeps the startup seed when `schedule::gc_schedule_enabled()`, exactly as for zeal.
- `ScheduleGuard::set/off/Drop` mirror `ZealGuard`'s arm/release pair, so test-time schedules reach polls too.
- Regression test `the_schedule_holds_the_poll_word_armed_like_zeal`, mirroring the zeal test, sabotage-verified.
- Acceptance: the equivalence table above re-run with the fix must show loop-poll-scale safepoints for `RATE=1`.

proggeramlug:
[...]

Issue #7606, opened 2026-08-07
gc: two rooting gap tests crash on macOS under PERRY_NO_AUTO_OPTIMIZE=1 while Linux passes them (untriaged)


proggeramlug:
Found in passing during #7599's known-failures audit, on the shared dev Mac
(macOS arm64, perry-dev profile, current main of 2026-08-08):

`test_gap_gc_rest_argument_rooting` and
`test_gap_gc_same_module_call_argument_rooting` **crash on macOS when compiled
with `PERRY_NO_AUTO_OPTIMIZE=1`**, while Linux (the gap_snapshot baseline
platform) records both as passing.

Not yet triaged — this issue exists so the observation is not lost. Three
things whoever picks it up should know:

1. These are **gc rooting acceptance tests** (the #7341 family), so a platform-
   split crash is potentially a real macOS-only rooting hole, not a harness
   quirk — the auto-opt and no-auto-opt runtimes differ in feature set and
   link layout, and the no-auto-opt arm is the one ordinary `perry compile`
   users get when the auto-opt path is bypassed.
2. `test_gap_gc_same_module_call_argument_rooting` has crashed spuriously
   before — during a disk-full window on 2026-08-07 it recorded a `crash` that
   re-ran 10/10 clean afterwards (#7573's report). First step is therefore
   **reproduce N/N on a healthy machine** before believing it; the #7599 agent
   saw it in passing, not under controlled conditions.
3. If it reproduces, the instruments are `PERRY_GC_ZEAL=1
   PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_DIAG=1` — but note #7604: on
   compute-only shapes the zeal recipe can arm without ever firing, so check
   `retired_set=` lines actually appear before trusting a clean run.

proggeramlug:
Closing as **not reproducible on current `main` (v0.5.1432)**, with the instrument proven live rather than merely quiet.

**Baseline**, the issue's own configuration:

| test | result |
|---|---|
| `test_gap_gc_rest_argument_rooting` | **10/10 clean** under `PERRY_NO_AUTO_OPTIMIZE=1` |
| `test_gap_gc_same_module_call_argument_rooting` | **10/10 clean** |

[...]

PR #7782, opened 2026-08-10
fix(gc): the seeded schedule arms the poll word, like zeal (#7781)

proggeramlug:
### Fixed

- **The seeded GC schedule never armed the poll word, so `PERRY_GC_SCHEDULE_RATE=1` was an event-loop-boundary instrument only (#7781).** On #7606's reproduction it saw **6 safepoints** where zeal's runs cross **9,648 / 19,248 loop polls** — the "collect at every opportunity" end of the dial saw six opportunities. #7735 collapsed the back-edge poll's no-work path to one load of `PERRY_GC_POLL_ARMED`; `resolve_poll_seed` kept the word armed for zeal, and nothing armed it for the schedule mode, so the loop-safepoint bypass #7317 added sat behind a gate that never opened.

  `resolve_poll_seed` now keeps the startup seed when `schedule::gc_schedule_enabled()`, exactly as for zeal, and `ScheduleGuard` mirrors `ZealGuard`'s arm/release pair so test-time schedules reach polls too. Regression test `the_schedule_holds_the_poll_word_armed_like_zeal` mirrors the zeal test and is sabotage-verified (removing the arm fails it).

  Re-run with the fix, same binary, same quarantine (`PERRY_GC_PROTECT_FROMSPACE=1`, depth 800):

  | | before | after |
  |---|--:|--:|
  | `rest_argument_rooting` | safepoints=6 | **safepoints=9,653**, 9,653 collections, exit 0 |
  | `same_module_call_rooting` | safepoints=6 | **safepoints=19,253**, 19,253 collections, exit 0 |

  This was found testing #7741's precondition — that `SCHEDULE_RATE=1` be demonstrated equivalent to zeal on a real reproduction before the older instrument is deleted. The precondition now holds; without this fix it measurably did not.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed scheduled garbage collection so polling remains active when scheduling is enabled.
  * Improved temporary scheduling behavior, including proper activation and release during nested or scoped changes.
  * Ensured scheduled collections run reliably at configured rates across event-loop boundaries.

* **Documentation**
  * Added a changelog entry describing the scheduling fix and verification.

* **Chores**
  * Updated the application version to 0.5.1451.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1450
+**Current Version:** 0.5.1451
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1450"
+version = "0.5.1451"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1450"
+version = "0.5.1451"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1450"
+version = "0.5.1451"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1450"
+version = "0.5.1451"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1450"
+version = "0.5.1451"
 dependencies = [
  "anyhow",
  "perry-hir",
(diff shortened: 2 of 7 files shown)


proggeramlug:
## Merging as v0.5.1451 — and this is the #7741 precondition, demonstrated

Found testing the maintainer's stated condition for #7741: that `SCHEDULE_RATE=1` be shown equivalent to zeal **on a real reproduction**, not asserted. It measurably was not:

| | before | after this fix |
|---|--:|--:|
| `rest_argument_rooting` | **safepoints=6** | **9,653** — matching zeal's 9,648 loop polls |
| `same_module_call_rooting` | **safepoints=6** | **19,253** — matching zeal's 19,248 |

Both clean under `PERRY_GC_PROTECT_FROMSPACE=1` at depth 800. The "collect at every opportunity" end of the dial saw six opportunities, because #7735's poll arming word was kept armed for zeal and nothing armed it for the schedule — the loop-safepoint bypass #7317 added sat behind a gate that never opened. This is #7735's own design note played out: the word is a counter of *reasons*, and a new mode whose decision lives inside the safepoint must register itself as one.

### The bug I shipped inside the first version of this fix, kept on the record

The first `ScheduleGuard` mirrored `ZealGuard` symmetrically — `off()` disarmed and its `Drop` re-armed. **`disarm_poll` saturates at zero**, so a disarm landing on a zero word is lost while the paired re-arm is not: a permanent **+1 leak** that pins the poll armed for the rest of the process. In the test binary that surfaced as two *consistent* failures I first misread as load flakes — the timing test slowed (every loop poll in 2,000 parallel tests paying a call), and the generation-gate contract took a safepoint drain mid-stage. Full-suite on pristine main at the same load: clean. That control is what separated "my leak" from "load".

The bookkeeping is now deliberately **asymmetric** — only `set()` arms, only its own `Drop` releases, `off()` never touches the word — with the saturation rationale in the comment. Over-arming for a guard's lifetime costs a wasted call; a leaked arm is forever. 3/3 full-suite runs clean after (2,035 passed).

The saturation that caused this is itself correct (#7735 chose it so an underflow could not pin the word armed via wrap) — the lesson is that **saturating operations break pair-wise symmetric protocols**, and any future arm/disarm pair around this word needs the same asymmetric shape.

`the_schedule_holds_the_poll_word_armed_like_zeal` mirrors the zeal test and is sabotage-verified. Gates 21/21.

**#7741 is now unblocked on its stated condition.**

PR #7783, opened 2026-08-10
test(gc): measure the grow-then-churn pacing transition and pin it as a ratchet probe (#7737 item 3)

proggeramlug:
Closes the last engineering item of #7737. **Item 3** is what this PR does; items 1 and 2 are already fixed on `main` (the `OBJECT_PROTOTYPES_NONEMPTY` latch releases when a prune drains the registry; `gc/tests/triggers.rs` asserts the positive direction of the pacing recording). **Item 4 — promoting `gc-ratchet` / `gc-stress` to required contexts — is a maintainer action and is untouched here.**

## The untested middle

The major-pacing backoff (#7726) has two well-covered ends. `retain.ts`-shaped growth is pure, so every escalated full reclaims almost nothing and the shift climbs to its cap of 2 and stays. `tree.ts`-shaped churn is always high-yield, so the shift never leaves 0. Nothing exercised a workload that is the **first and then becomes the second** — which is the shape a warm-cache phase followed by steady churn actually has.

`14_grow_then_churn.ts` is that workload: an all-live 300k-row cache that walks the shift 0 → 1 → 2 on growth alone, then push-grown array churn with the cache still live so the pacing baseline stays up.

## What it measured — not what the issue predicted

#7737 reasoned the delayed reclaim would cost "up to 4× more transient RSS", since array-growth forwarding stubs are reclaimable only by a full mark-sweep. A same-binary A/B (one runtime at `MAJOR_PACING_BACKOFF_SHIFT_MAX = 2`, one at `0`, archives rebuilt and mtime-verified between arms, stdout byte-identical):

| | cap 2 (shipped) | cap 0 (no backoff) |
|---|---|---|
| escalation boundary reached | 56.3 MB | 14.1 MB |
| full mark-sweeps | 3 | 5 |
| copying minors / objects moved | 24 / 306,715 | 24 / 306,715 |
| peak RSS, 3 runs | 70.5 / 70.2 / 70.5 MB | 72.6 / 72.6 / 72.6 MB |

The boundary moves 4×, exactly as reasoned. **Peak RSS does not follow it, and does not move in the predicted direction** — the shipped backoff is 2.2 MB (3%) *lower*, reproducibly, while running two fewer fulls for identical collector work. Footprint on this shape is set by the steady state the minors hold, not by cumulative stub debt, so the escalation boundary is never the binding constraint.

So the probe **pins the measured state** rather than asserting the predicted bound. A future change that makes the delayed reclaim start to bind moves `rss_bytes` off the pin and the ratchet reports it.

Provenance: M1 Max laptop, `perry-dev`, peak RSS from plain runs with no GC tracing (tracing perturbs RSS, so counters and footprint come from separate runs). Not the pinned mini that owns the artifact — the comparison transfers, the absolute bytes belong to that context.

## The probe reaches its subject

The failure this suite has paid for most often is a knob that moves the workload off the path it was chosen to exercise, so this was checked rather than assumed: `PERRY_GC_TRACE=1` reports `backoff_shift` reaching 2 with the boundary at 56.3 MB, and the harness records `minor_cycles=24` with 306,715 objects moved, satisfying `check`'s liveness rule with a collector that demonstrably ran.

The two `gc-ratchet-env` directives (`PERRY_GC_SCAVENGE_NURSERY_MB=1`, `PERRY_GC_MAJOR_PACING_FLOOR_MB=1`) set the **absolute** scale of a mechanism that is otherwise a ratio — at the shipped 16 MB / 32 MB the same three escalations need a live set in the hundreds of MB. A new test asserts both, the same discipline `13_large_eden_survivors` already gets: losing either line leaves a probe that still passes, still collects, and no longer reaches the cap it exists to hold.

It is also the most stable probe in the suite — 0% spread on both retention metrics, 0.022% on RSS, against 4.7% and 5.4% for `12_large_live_set` and `13_large_eden_survivors`.

## The re-pin is not a clean carry-over, and says so

Adding a probe changes the probe set, so the artifact must be re-pinned. **The control was not green**, and rather than absorbing that, the artifact notes name every cell:

| probe | cell | pinned | now |
|---|---|---|---|
| `04_dead_after_deep_stack` | copied_objects | 565 | 605 (+7.08%) |
| `11_collect_at_depth` | heap_used_bytes | 5,097,776 | 5,307,584 (+4.12%) |
| `11_collect_at_depth` | promoted_objects / bytes | 6,150 / 420,488 | 0 / 0 |
| `11_collect_at_depth` | copied_objects / bytes | 0 / 0 | 6,139 / 419,608 |

Probe 11's survivors stopped being promoted on first copy and are now copied into survivor space — the #7558 shape, running the other way.

**Attribution, so it is recorded as code and not drift.** The control ran twice and reported the same six cells both times. A release binary was built from `a8f73122d`, the commit the previous artifact was pinned at, and `--check` from that tree against that artifact reported **zero** regression cells. Running one probe directly under each binary is deterministic 6-of-6 within each arm and differs between them. Two further cells (`13_large_eden_survivors` RSS, +7.2%) fired in the first control and not the second, so those are host variability on a sampled metric and are **not** claimed as a shift.

**#7742 is excluded as the cause, by its own PR's evidence.** #7744 measured `gc_ratchet` against this same artifact *before* it merged and reported: "the six cells it reports as REGRESSION are reported identically by `origin/main` itself — pre-existing drift against the pinned artifact." The same six. So the flip predates the whole-block promotion change and comes from one of the earlier collector commits — #7733 yield-adaptive major pacing, #7690 / #7682 evacuation at precise safepoints, #7721 / #7735 the loop back-edge poll, #7317 the seeded schedule, or the #7469 thread-local work.

**Why it went unrecorded**: 78 commits separate the two pins, twelve in the collector, and `gc-ratchet` is not a required context — #7737 item 4 — so none of them ever had to re-pin. It has now been observed by at least two people without being recorded, because an advisory gate cannot force the question.

It is pinned so the *next* change has a truthful baseline. **If the promote-to-copy flip on `11_collect_at_depth` was not intended by whichever change caused it, then this artifact is the wrong fix and the code is the right one** — that call is the maintainer's, and the notes say so in the artifact itself.

## Validation

- `run_gc_ratchet_baseline.sh --check` against the new artifact on the pinned host, clean tree, all 14 probes: **0 regression cells, `gc-ratchet: OK`, exit 0**.
- New probe passes the Node v26.5.1 oracle byte-for-byte (`correctness=pass`).
- `tests/test_gc_ratchet.py`: 92 passed, including the new directive assertion.
- `scripts/check_file_size.sh`: exit 0.
- Pinned on perry-macos (M1 mini), AC power, CPU-quiet gate satisfied before each phase, `PERRY_NO_AUTO_OPTIMIZE=1`, runtime dir pinned. This branch changes no Rust at all — a probe `.ts`, the README, a workflow comment, a Python test and the artifact — so the release binary is the right one for this tree.

## Summary by CodeRabbit

* **New Features**
  * Added a garbage-collection benchmark covering cache growth followed by allocation churn.
  * Added heap, RSS, checksum, and stability measurements for the new workload.

* **Documentation**
  * Updated benchmark and preflight documentation to reflect fourteen probes.
  * Added changelog details and measurement results for the new probe.

* **Tests**
  * Added coverage validating the new probe’s required runtime settings.

* **Chores**
  * Incremented the workspace version to 0.5.1457.

### DIFF
--- a/.github/workflows/gc-ratchet.yml
+++ b/.github/workflows/gc-ratchet.yml
@@ -81,7 +81,7 @@ jobs:
       #
       # `--scope structural` is load-bearing and is the #7554 repair. This step
       # runs BEFORE the measurement step, so anything it fails on costs the
-      # entire run's coverage — thirteen probes that never execute. That price
+      # entire run's coverage — fourteen probes that never execute. That price
       # is correct for "this artifact is unreadable or tampered with" and badly
       # wrong for "one of its 156 cells is not bit-identical", which is what
       # actually happened: one cell's 6,768-byte spread failed this step and the
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1456
+**Current Version:** 0.5.1457
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5547,7 +5547,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1456"
+version = "0.5.1457"
 dependencies = [
  "anyhow",
  "base64",
@@ -5607,14 +5607,14 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1456"
+version = "0.5.1457"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1456"
+version = "0.5.1457"
 dependencies = [
  "cc",
  "libc",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1456"
+version = "0.5.1457"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1456"
+version = "0.5.1457"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1456"
+version = "0.5.1457"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1456"
+version = "0.5.1457"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5664,7 +5664,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1456"
+version = "0.5.1457"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5672,7 +5672,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1456"
+version = "0.5.1457"
 dependencies = [
  "anyhow",
  "base64",
@@ -5684,7 +5684,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1456"
+version = "0.5.1457"
 dependencies = [
  "anyhow",
  "perry-hir",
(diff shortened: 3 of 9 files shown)


proggeramlug:
Audit complete — merging.

**Verified:**
- No Rust changes (probe .ts, baseline artifact, README, workflow comment, harness test) — so the binary question reduces to provenance, and the baseline's own notes carry it: pinned on `perry-macos` (the artifact's owner host), with the absorbed 6-cell shift named, attributed via a control (old binary vs old artifact: zero regressions), and traced to the #7742-era collector work.
- 14/14 probes registered in the baseline; probe-set consistency holds.
- Full harness suite: 92 tests, OK. Sabotage verified by me (applied, then restored): deleting one `gc-ratchet-env` directive from the probe fails exactly the new directive-assertion test.
- **The probe reaches its subject on an independent host**: run here with its two env directives + `PERRY_GC_TRACE=1`, `backoff_shift` walks to 2 and the run records **24 minors + 3 fulls — the pinned counts exactly** (deterministic counters transfer; the RSS bytes stay host-relative as the notes say).
- Local gate green (test steps skipped — no Rust in the diff; the harness's own suite ran above).

This closes #7737 item 3 — the last engineering item. Item 4 (promoting gc-ratchet/gc-stress to required contexts) remains the maintainer/admin action already tracked there.

PR #7784, opened 2026-08-10
Wire the prompt and auto update modes, behind three refusals

jdalton:
Stacked on [#7749](https://github.com/PerryTS/perry/pull/7749). Until that merges this branch's diff includes its commit; I will rebase once it lands so this collapses to just the new work.

## The `prompt` and `auto` modes now do something — and refuse to do the wrong thing

[#7749](https://github.com/PerryTS/perry/pull/7749) made the modes configurable but deliberately left them inert, because replacing the binary somebody is currently running deserved its own review. This is that review.

### Refusal 1: a package-managed install is never replaced in place

`perry update` overwrites the running executable. That is correct for a tarball or `install.sh` install, and wrong for every managed one — Homebrew, npm, apt and winget each keep their own record of what is installed and at what version, and overwriting the file underneath leaves that record lying. The next `brew upgrade` either reinstalls over the top or reports a version that is not what is on disk.

So `prompt` and `auto` detect the owner and name *that owner's* command instead:

| owner | what Perry tells you to run |
|---|---|
| Homebrew | `brew upgrade perryts/perry/perry` |
| npm | `npm install -g @perryts/perry@latest` |
| apt | `sudo apt update && sudo apt install --only-upgrade perry` |
| winget | `winget upgrade PerryTS.Perry` |

npm gets an extra sentence because it is the worst case: Perry ships as a wrapper package plus a per-platform binary package, so replacing the binary also desyncs it from the wrapper that launched it.

### Refusal 2: nothing is offered after a command that failed

The user is looking at an error. A question about upgrading is noise at the worst possible moment, and an unattended install would bury the error under progress output. Both active modes fall back to a plain notice.

### Refusal 3: an unwritable install directory is reported, not attempted

`install.sh` targets `/usr/local/bin`, which is root-owned on a default macOS and most Linux boxes. That is now checked **before** anything is downloaded, so the outcome is one sentence naming `sudo perry update` rather than a download that fails at the final rename. Perry never escalates on its own.

### Also

`perry update --mode <off|notify|prompt|auto>` saves the setting and exits, so the one thing people are most likely to change does not mean hand-editing TOML. `perry doctor` now reports the effective mode and the owning package manager, which are the two questions behind "why did it not update".

<details>
<summary><b>Why the channel detection deliberately fails open</b></summary>

Every rule answers "is this definitely managed?", never "is this definitely unmanaged?" — an unrecognised layout resolves to self-managed.

That asymmetry is the point. Guessing "managed" wrongly would refuse to self-update a plain tarball install, which is the majority case and the one with no other upgrade path. Guessing "self-managed" wrongly costs an in-place update on a machine that had a package manager available, which the user can recover from by running that manager.

Two details worth knowing:

- **Paths are canonicalized before classification.** Homebrew's `perry` in `/usr/local/bin` is a symlink into the Cellar, so classifying the link rather than its target would miss every Homebrew install there is.
- **apt requires both halves of its rule** — a dpkg file list *and* a dpkg-owned path — because dpkg does not own `/usr/local`, which is `install.sh`'s directory. The path alone would misclassify a hand-placed binary; the dpkg list alone would claim a tarball install on a machine that also has the `.deb` installed elsewhere. It is a file-existence check rather than a `dpkg -S` subprocess, since this runs on the update path of every command.
</details>

<details>
<summary><b>Prompting needs stdin, not just stderr</b></summary>

The mode gate from [#7749](https://github.com/PerryTS/perry/pull/7749) already requires stderr to be a terminal, but that is not enough to ask a question. stdin can be a pipe while stderr is a tty, and reading from it would either block the command or treat whatever the pipe happened to contain as consent. `prompt` degrades to a plain notice when stdin is not a terminal.

`auto` asks nothing, so it does not need stdin — but it still requires a successful command, an unmanaged install and a writable directory.
</details>

<details>

24 new, all in the required per-pull-request job.

The decision is a pure function of the mode plus four facts about the machine, which is why it is a separate function from carrying it out: every refusal is then asserted directly, rather than living inside an `if` in the middle of a teardown path where nothing can reach it.

- both active modes downgrade to a notice after a failed command;
- both refuse on all four managed channels, and every managed channel offers a command (a refusal with no alternative is a dead end, so that is asserted too);
- both report elevation rather than attempting an unwritable install;
- `prompt` degrades without stdin while `auto` does not need it.

The channel table covers Homebrew under all three prefixes, npm for global / nvm / project-local layouts, apt with and without each half of its rule, both winget delivery shapes, and four unrecognised layouts that must fail open. Classification splits on both path separators instead of using `Path::components`, so the winget cases actually run on every host rather than only on Windows — otherwise that half of the table would be untested everywhere CI runs it.

**Verified end to end.** Writing `mode` into a real config file that already contained a `license_key` and an unknown `[update] future_key` left both intact:

```toml
license_key = "keep-me"

[update]
mode = "auto"
server = "https://updates.example.test/latest"
future_key = 42
```

`cargo test -p perry`: **914 passed, 0 failed.** `scripts/check_file_size.sh` passes.
</details>

## Summary by CodeRabbit

- **New Features**
  - Added configurable update modes: off, notify, prompt, and auto.
  - Added detection for Homebrew, npm, apt, winget, and self-managed installations.
  - Added installer-specific upgrade guidance, permission handling, and automatic update behavior.
  - Enhanced update notifications with policy, installer, and command details.

- **Bug Fixes**
  - Improved notification throttling, cache reliability, and timeout fallback behavior.
  - Prevented malformed configuration files from being overwritten during updates.

- **Documentation**
  - Added changelog coverage for update modes, installation channels, and reliability improvements.

### DIFF
--- a/changelog.d/7784-update-prompt-auto-and-channels.md
+++ b/changelog.d/7784-update-prompt-auto-and-channels.md
@@ -1,0 +1,105 @@
+### Added
+
+**`prompt` and `auto` update modes now do something, and refuse to do the wrong
+thing.** The previous slice made the modes configurable; this wires them to the
+existing signed self-updater, behind three refusals.
+
+**A package-managed install is never replaced in place.** `perry update`
+overwrites the running executable, which is right for a tarball or `install.sh`
+install and wrong for every managed one: Homebrew, npm, apt and winget each keep
+their own record of what is installed and at what version, and overwriting the
+file underneath leaves that record lying. `prompt` and `auto` now detect the
+owner and name that owner's command instead:
+
+| owner | what Perry says to run |
+|---|---|
+| Homebrew | `brew upgrade perryts/perry/perry` |
+| npm | `npm install -g @perryts/perry@latest` |
+| apt | `sudo apt update && sudo apt install --only-upgrade perry` |
+| winget | `winget upgrade PerryTS.Perry` |
+
+npm gets an extra sentence, because it is the worst case: Perry ships as a
+wrapper package plus a per-platform binary package, so replacing the binary also
+desyncs it from the wrapper that launched it.
+
+**Nothing is offered after a command that failed.** The user is looking at an
+error; a question about upgrading is noise at the worst possible moment, and an
+unattended install would bury the error under progress output. Both active modes
+fall back to a plain notice.
+
+**An unwritable install directory is reported, not attempted.** `install.sh`
+targets `/usr/local/bin`, which is root-owned on a default macOS and most Linux
+boxes. That is now checked *before* anything is downloaded, so the outcome is
+one sentence naming `sudo perry update` rather than a half-finished install. Perry
+never escalates on its own.
+
+**`perry update --mode <off|notify|prompt|auto>`** saves the setting and exits,
+so the one thing people are most likely to change does not require hand-editing
+TOML. It is a read-modify-write through the shared loader, so the rest of the
+file comes back out the way it went in.
+
+**`perry doctor`** now reports the effective mode and, when there is one, the
+package manager that owns the binary — the two questions behind "why did it not
+update".
+
+<details>
+<summary><b>Why the channel detection fails open</b></summary>
+
+Every rule answers "is this definitely managed?", never "is this definitely
+unmanaged?", and an unrecognised layout resolves to self-managed.
+
+That asymmetry is deliberate. Guessing "managed" wrongly would refuse to
+self-update a plain tarball install — the majority case, and the one with no
+other upgrade path. Guessing "self-managed" wrongly costs an in-place update on
+a machine that had a package manager available, which is recoverable by running
+that manager.
+
+The paths are canonicalized before classification, because Homebrew's `perry` in
+`/usr/local/bin` is a symlink into the Cellar; classifying the link rather than
+its target would miss every Homebrew install there is.
+
+apt requires **both** a dpkg file list and a dpkg-owned path, because dpkg does
+not own `/usr/local` — that is `install.sh`'s directory. The path alone would
+misclassify a hand-placed binary; the dpkg list alone would claim a tarball
+install on a machine that also has the `.deb` installed somewhere else. The check
+is a file-existence test rather than a `dpkg -S` subprocess, since this runs on
+the update path of every command.
+</details>
+
+<details>
+<summary><b>Prompting needs stdin, not just stderr</b></summary>
+
+The mode gate already requires stderr to be a terminal. That is not enough to
+ask a question: stdin can be a pipe while stderr is a tty, and reading from it
+would either block the command or take whatever the pipe happened to contain as
+consent. `prompt` degrades to a plain notice when stdin is not a terminal.
+
+`auto` asks nothing, so it does not need stdin — but it does still require the
+command to have succeeded, an unmanaged install, and a writable directory.
+</details>
+
+<details>
+<summary><b>Tests</b></summary>
+
+24 new, all in the required per-pull-request job. The decision is a pure
+function of the mode plus four facts about the machine, so every refusal is
+asserted directly rather than left inside an `if` in the middle of a teardown
+path:
+
+- both active modes downgrade to a notice after a failed command;
+- both refuse on all four managed channels, and name a command for each;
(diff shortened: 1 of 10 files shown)


jdalton:
Pushed as a single commit rebased on current `main`. Every review finding on this pull request is now fixed in code.

<details>
<summary><b>What changed, and why each one mattered</b></summary>

**`perry doctor` answered the wrong question.** `doctor` exists so you can ask "what is my update setting?" — but it was reporting the *effective* mode for the current run, and `doctor --format json`, `doctor | less` and every CI run suppress the update surface for their own run. So the command whose entire job is to report your setting answered `off` no matter what your config said. `UpdatePolicy` now carries both values: `mode` is the decision for this run, and `configured_mode` is what the config or `PERRY_UPDATE_MODE` actually says. `doctor` reports the configured one, and adds "suppressed for this run" when the two differ, so nothing is hidden.

**`perry update --mode` could erase the config it was writing to.** It read with `load_config`, mutated, and saved — the read-then-save shape that turns an unparseable config into an erased one. It now goes through `update_config_file`, which refuses the write when the file could not be read. A new regression test covers this, and reverting the fix makes it fail.

**Two smaller ones carried up from the base branch:** the teardown cache lock is `try_lock` rather than the blocking `lock()` it documented, and a background check that misses its 100 ms budget falls back to the cached status instead of being read as "no update".

</details>

<details>
<summary><b>Why this branch contains the base branch's commit too</b></summary>

This is the second of four stacked pull requests, so it necessarily contains [#7787](https://github.com/PerryTS/perry/pull/7787) beneath it. Review the top commit; merging [#7787](https://github.com/PerryTS/perry/pull/7787) first shrinks this diff to just the prompt and auto modes.

</details>

jdalton:
Both findings from the latest review are real, and both are fixed. Force-pushed as one commit on current `main`.

<details>
<summary><b>Auto mode was being silently blocked by the notice throttle</b></summary>

This was the more serious of the two, and it is a good catch. `notify_interval_hours` exists to stop Perry repeating the *same notice* every run. But `main.rs` used it to gate the entire `run_teardown_action` call — and that call is also what performs the install in `auto` mode. So with a notify interval set, a notice printed an hour ago stopped the new version from ever landing. That is the one thing `auto` promises to do.

The throttle is now passed down into `run_teardown_action` instead of wrapped around it, and applied *after* the action is resolved, so it only silences the things that speak:

```rust
if notice_throttled && throttle_applies(&action) {
    return;
}
```

`throttle_applies` is a two-line predicate that returns false only for `Install`, which makes the rule testable without performing a real install. `the_notice_throttle_never_holds_back_an_install` asserts both directions, and making the predicate return `true` unconditionally turns it red.

</details>

<details>
<summary><b>An unreadable config still resolved to defaults</b></summary>

Correct, and this reopened the exact data-loss case the function was added to close. `load_config_checked` treated every `read_to_string` failure as "no config yet", but only `ErrorKind::NotFound` means that. A permission change or a transient I/O error also returned `Ok(PerryConfig::default())`, which `update_config_file` would then happily serialize over a file still holding `license_key` and `api_token`. It now matches on the error kind and returns `Err` for anything that is not `NotFound`.

Worth noting how the test for this had to change. My first version just asserted that the write failed — and that test passed *even with the fix reverted*, because a file at mode `000` also fails to **write**, so the error arrived either way. A test that accepts any error cannot tell whether the read is checked at all. It now asserts the error names the read failure specifically, and skips itself when running as root, since root ignores the mode bits and there would be nothing to prove. Reverting the fix now produces a bare `Permission denied (os error 13)` from the write and the assertion rejects it.

</details>

<details>
<summary><b>`configured_mode` was hiding config typos</b></summary>

Also right, and the reason is a nice one. `UpdateMode::Unknown`'s label is already `"notify (unrecognized value in config)"` — exactly what `doctor` should say when the config has a typo in it. My mapping collapsed `Unknown` to `Notify` before storing it, so `doctor` printed a clean `notify` and hid the typo it was best placed to surface. It is now `config.mode.unwrap_or(UpdateMode::Notify)`, which preserves `Unknown` for reporting while `mode` stays resolved through `resolve_mode` for behaviour.

The changelog fragment's "reports the effective mode" wording was wrong for the same reason and now says "configured mode".

</details>

PR #7785, opened 2026-08-10
Make the update check source a choice of four

jdalton:
Stacked on [#7784](https://github.com/PerryTS/perry/pull/7784), which is stacked on [#7749](https://github.com/PerryTS/perry/pull/7749). Until those merge this branch's diff includes their commits; I will rebase as each lands.

## Where Perry asks "what is the latest version?" becomes a choice

It used to walk one fixed list — an override, the config, Perry Hub, then the GitHub releases API — and read a GitHub-releases-shaped document from whichever answered first.

That is fine while everyone installs the same way, and wrong as soon as they do not. An npm user's "latest" is whatever the registry's `latest` dist-tag says, so asking GitHub instead can announce a version their package manager cannot install yet — a notice telling someone to upgrade to something they cannot get.

```toml
[update]
source = "npm"              # gh-releases | npm | gh-registry | custom
package = "@perryts/perry"  # npm-shaped sources; defaults to Perry's own
registry = "..."            # npm-shaped sources; defaults to the public registry
server = "..."              # the URL for `custom`, and the mirror override
```

Leaving `source` unset keeps today's behaviour exactly, **except** on an npm-managed install, which now defaults to asking npm — because that is the version its own package manager can actually install.

### The split that matters: checking is not downloading

A source answers one question and returns a version, a link, a publish time and a headline. It does **not** decide where the binary comes from. Artifacts and their signed manifest always resolve from the release infrastructure, whatever the source is.

That separation is load-bearing rather than tidy. The manifest — Ed25519 over the artifact's digest and version — is what makes a self-update trustworthy, and `source` is a URL a user can point anywhere. If it could redirect the download, this setting would be a way to install arbitrary code. Whoever answers "what is new?" never gets to answer "what should I run?", and there is a test that fails if a source ever leaks into the artifact ladder.

<details>
<summary><b>Why the old ladder is deleted rather than left underneath</b></summary>

`get_update_servers` and its private config reader are **gone**, not kept beside the new code.

Leaving them would produce the shape where four sources exist, pass their own unit tests, and are never actually reached, because the call sites still walk the old list. Deleting them makes the compiler prove both sites moved — there is no version of this change that builds while the new abstraction is dead.
</details>

<details>
<summary><b>Credentials go to exactly one of the four</b></summary>

The npm shapes request the *abbreviated* packument (`Accept: application/vnd.npm.install-v1+json`) — smaller, cacheable, and the document npm itself asks for. It also sidesteps GitHub's unauthenticated API rate limit, which the old ladder shared with everything else on the machine.

The public registry is asked **without credentials**, and a test asserts no `Authorization` header is sent: a token there would be a leak, not a convenience. GitHub Packages does need one, so that shape reads `GH_TOKEN` / `GITHUB_TOKEN` and fails with a sentence naming the fix when neither is set — rather than retrying anonymously and reporting the resulting 404 as "up to date".

A configured source does not fall back to the ladder when it errors. Somebody who said "ask npm" and got a failure wants to hear that, not a version from somewhere they never named.
</details>

<details>

11 new, all parsing real response shapes from string fixtures, so no network is involved:

- a GitHub release document, including that the `v` prefix is stripped;
- an abbreviated packument, which carries no `time` map — so the publish date reads "unknown" rather than being invented, which matters because the release cooldown in the next slice depends on it;
- a full packument, which does supply it;
- a custom manifest with only a `version`, and one with every optional field;
- that each shape **rejects the others' documents** rather than reading whichever field happens to be present — a registry answering a gh-releases request must be an error, not a version of `""`;
- that a scoped package's `/` is percent-encoded, or the registry reads the scope as a path segment and answers 404;
- that an unknown `source` falls back to the default instead of failing, because an update check is the wrong place to turn a config typo into a hard error;
- that `custom` with no URL is a missing key rather than a default;
- that an npm install defaults to npm while every other channel keeps the ladder;
- that no source can reach the artifact ladder;
- and both credential rules.

`cargo test -p perry`: **932 passed, 0 failed.**
</details>

## Summary by CodeRabbit

- **New Features**
  - Configure update behavior with `off`, `notify`, `prompt`, or `auto` using `perry update --mode`.
  - Update checks now recognize installation channels and display channel-specific upgrade guidance.
  - Configure update-check sources, including GitHub Releases, npm, package registries, and custom HTTPS endpoints.
  - Notifications are version-aware and better respect configured intervals.
- **Bug Fixes**
  - Improved update-cache reliability and compatibility with existing cache files.
  - Prevented malformed configuration files from being overwritten.
- **Documentation**
  - Added guidance for update modes, installation channels, check sources, and related safeguards.

### DIFF
--- a/changelog.d/7784-update-prompt-auto-and-channels.md
+++ b/changelog.d/7784-update-prompt-auto-and-channels.md
@@ -1,0 +1,105 @@
+### Added
+
+**`prompt` and `auto` update modes now do something, and refuse to do the wrong
+thing.** The previous slice made the modes configurable; this wires them to the
+existing signed self-updater, behind three refusals.
+
+**A package-managed install is never replaced in place.** `perry update`
+overwrites the running executable, which is right for a tarball or `install.sh`
+install and wrong for every managed one: Homebrew, npm, apt and winget each keep
+their own record of what is installed and at what version, and overwriting the
+file underneath leaves that record lying. `prompt` and `auto` now detect the
+owner and name that owner's command instead:
+
+| owner | what Perry says to run |
+|---|---|
+| Homebrew | `brew upgrade perryts/perry/perry` |
+| npm | `npm install -g @perryts/perry@latest` |
+| apt | `sudo apt update && sudo apt install --only-upgrade perry` |
+| winget | `winget upgrade PerryTS.Perry` |
+
+npm gets an extra sentence, because it is the worst case: Perry ships as a
+wrapper package plus a per-platform binary package, so replacing the binary also
+desyncs it from the wrapper that launched it.
+
+**Nothing is offered after a command that failed.** The user is looking at an
+error; a question about upgrading is noise at the worst possible moment, and an
+unattended install would bury the error under progress output. Both active modes
+fall back to a plain notice.
+
+**An unwritable install directory is reported, not attempted.** `install.sh`
+targets `/usr/local/bin`, which is root-owned on a default macOS and most Linux
+boxes. That is now checked *before* anything is downloaded, so the outcome is
+one sentence naming `sudo perry update` rather than a half-finished install. Perry
+never escalates on its own.
+
+**`perry update --mode <off|notify|prompt|auto>`** saves the setting and exits,
+so the one thing people are most likely to change does not require hand-editing
+TOML. It is a read-modify-write through the shared loader, so the rest of the
+file comes back out the way it went in.
+
+**`perry doctor`** now reports the effective mode and, when there is one, the
+package manager that owns the binary — the two questions behind "why did it not
+update".
+
+<details>
+<summary><b>Why the channel detection fails open</b></summary>
+
+Every rule answers "is this definitely managed?", never "is this definitely
+unmanaged?", and an unrecognised layout resolves to self-managed.
+
+That asymmetry is deliberate. Guessing "managed" wrongly would refuse to
+self-update a plain tarball install — the majority case, and the one with no
+other upgrade path. Guessing "self-managed" wrongly costs an in-place update on
+a machine that had a package manager available, which is recoverable by running
+that manager.
+
+The paths are canonicalized before classification, because Homebrew's `perry` in
+`/usr/local/bin` is a symlink into the Cellar; classifying the link rather than
+its target would miss every Homebrew install there is.
+
+apt requires **both** a dpkg file list and a dpkg-owned path, because dpkg does
+not own `/usr/local` — that is `install.sh`'s directory. The path alone would
+misclassify a hand-placed binary; the dpkg list alone would claim a tarball
+install on a machine that also has the `.deb` installed somewhere else. The check
+is a file-existence test rather than a `dpkg -S` subprocess, since this runs on
+the update path of every command.
+</details>
+
+<details>
+<summary><b>Prompting needs stdin, not just stderr</b></summary>
+
+The mode gate already requires stderr to be a terminal. That is not enough to
+ask a question: stdin can be a pipe while stderr is a tty, and reading from it
+would either block the command or take whatever the pipe happened to contain as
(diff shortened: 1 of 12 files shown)


jdalton:
Pushed as a single commit rebased on current `main`. Every review finding is fixed, and the most serious one — a GitHub token going out over plain HTTP — is covered by a test that fails if the fix is removed.

<details>
<summary><b>A token could have been sent in clear text</b></summary>

The download path already required HTTPS. The *check* path did not, and for the `gh-registry` source that is the more dangerous of the two: `perry` attaches `Authorization: Bearer <token>` to a GitHub Packages request, so an `http://` registry in the config would have put a real GitHub token on the wire unencrypted.

`require_secure` now runs on all four sources, and it runs **before** the headers are built rather than after, because checking afterwards would still leak on a retry. It also rejects credentials embedded in the URL — `https://user:pw@host/…` sends them to whatever host follows the `@`, and they end up in logs besides. Loopback `http://` stays allowed so a local test server still works, and the check requires a `:`, a `/`, or the end of the string after the loopback prefix so that `http://localhost.example.test` is not mistaken for localhost.

`a_plaintext_registry_is_refused_before_a_token_is_attached` covers all four sources, asserts the error message does not echo the token, and confirms loopback still works. Deleting `require_secure` turns it red.

</details>

<details>
<summary><b>The new check path was quietly resetting the notice throttle</b></summary>

The configured-source refresh had its own copy of the cache rebuild, and that copy did two things the older fallback path had already learned not to do. It read the notice state *before* making the request, so a notice recorded while the request was in flight got overwritten with a minutes-old value — which means you would be told twice about the same release, the exact thing the throttle exists to prevent. And it dropped `last_notified_version` entirely, so the version-keyed throttle reset on every single refresh.

Both paths now do the same thing: take the lock, re-read, carry both notice fields forward.

</details>

<details>
<summary><b>Why this branch contains two other commits</b></summary>

Third of four stacked pull requests, so [#7787](https://github.com/PerryTS/perry/pull/7787) and [#7784](https://github.com/PerryTS/perry/pull/7784) sit beneath it. Review the top commit; merging the two below first shrinks this diff to just the check sources.

</details>

jdalton:
All findings from the latest review are real and fixed. Force-pushed as one commit on current `main`.

<details>
<summary><b>The old fallback ladder took a configured server without checking its scheme</b></summary>

This is the same class of problem I had just fixed on the new check sources, and the reviewer is right that I stopped one function short. `release_info_servers` pushed `PERRY_UPDATE_SERVER` and `[update].server` straight into `client.get(url)` with no validation — `require_https` only ever covered the manifest and artifact URLs.

The interesting risk here is not eavesdropping. Anyone who can answer a plaintext version check can **suppress updates indefinitely** by reporting the running version as the latest one, which is a quiet way to hold a machine on a vulnerable build. Both entries now go through a shared `url_is_secure`, so HTTPS and loopback are accepted and anything else is dropped in favour of the release infrastructure. `require_secure` was refactored to call the same helper, so the check path and the ladder cannot drift apart.

</details>

<details>
<summary><b>An empty release URL printed a dangling sentence</b></summary>

`parse_custom` turns a missing `release_url` into `""`, and a custom manifest is allowed to carry only `version`. That empty string reached both output paths, so the notice ended with `or visit ` and `perry update --check` printed `Release: ` with nothing after it. Both now omit the clause when the URL is empty rather than printing a sentence that trails off.

</details>

<details>
<summary><b>The fragments contradicted each other on test counts</b></summary>

A fair point about assembled release notes. Each fragment recorded the suite total at the moment its own slice was written — 904, 914, 925, 929, 930 — so the folded notes reported five different totals for one release, and the numbers appeared to go *down* across entries that build on each other. The count is a development detail rather than release-note material, so the line is gone from all five fragments. I deliberately left the counts in fragments already on `main` from other changes, since those are not this change's business. The forward reference to "the next slice" is also gone, and each fragment now describes shipped behaviour.

</details>

PR #7786, opened 2026-08-10
Add a release cooldown, a skip-this-version answer, and the updates documentation

jdalton:
Stacked on [#7785](https://github.com/PerryTS/perry/pull/7785) → [#7784](https://github.com/PerryTS/perry/pull/7784) → [#7749](https://github.com/PerryTS/perry/pull/7749). Until those merge this branch's diff includes their commits; I will rebase as each lands.

### `auto` waits a day before installing a brand-new release

`min_age_hours` defaults to 24 for `auto` and 0 for every other mode. A notice still mentions a release the moment it exists; only the unattended install holds off.

A release published by mistake, pulled shortly after, or published by someone who should not have been able to is most dangerous in its first hours. Waiting costs nothing and means your machine is not the one that finds out. `notify` and `prompt` are unaffected — they tell a human, who can decide for themselves.

**An unknown publish date counts as too fresh, not as old enough.** The abbreviated npm document carries no dates, so treating unknown as "old enough" would switch the cooldown off for exactly the people using the cheapest source — a protection present in the config and absent in effect. `min_age_hours = 0` turns it off deliberately.

### The prompt has three answers, not two

"No" and "never tell me about this one" are different intentions. With only two answers, somebody who does not want one specific release has to switch the whole mode off — which then hides the release that fixes it. The third answer writes `skip_version`, suppressing exactly that version; the next release is mentioned normally.

### The notice says what the release is

Sources that carry a title now pass it through, printed under the version line at no extra request. "Something is available" becomes "this is available".

### Documentation

A new page, `docs/src/cli/updates.md`, covering the default behaviour, every `[update]` key, the four modes and their three refusals, the four check sources, the cooldown, skipping a version, exactly what a check transmits, and where the two files live.

`perry update`'s section in `commands.md` is rewritten around `--mode`. `installation.md` gains the per-package-manager upgrade table. The environment tables in `flags.md` gain `NO_UPDATE_NOTIFIER` and `PERRY_UPDATE_MODE`.

<details>
<summary><b>One design change worth calling out: the cache is versioned, not migrated</b></summary>

`~/.perry/update-check.json` now records a `schema` number, and a value this build does not recognize — including its absence, which reads as `0` — means the file is discarded and the next check rewrites it.

That replaces the alternative I had been accumulating: keep every field optional forever so older shapes still load. For a *cache* that is a bad trade. It buys one saved network request in exchange for a set of `Option` fields that exist only to describe versions nobody runs, and that nothing ever removes. Bumping `CACHE_SCHEMA` is now the whole migration story.

In the same spirit, the check sources lost their aliases: `github`, `npm-registry` and `github-packages` are gone, leaving one spelling each. An accepted-alias set is a surface you document and test forever to save one look at the docs.

Verified at runtime, not only in a parse test: a planted cache with no schema, claiming version `99.0.0` and a `last_check` in 2099, was ignored; a real check ran; the file came back stamped `"schema": 1`.
</details>

<details>

Four new on the decision side:

- `auto` holds off inside the cooldown and installs past it, while `notify` and `prompt` are never held back;
- an unknown release age is treated as too fresh, and an explicit `0` still lets it through so nobody is stuck;
- the cooldown defaults to a day for `auto` only, and an explicit value wins for every mode;
- a skipped version suppresses itself and **not** the next one — which is what separates it from switching notices off.

Plus the cache-schema contract, replacing the test that used to assert an older shape still loads with one asserting that a foreign schema, and an absent one, are both recognized as not-ours.

`cargo test -p perry`: **932 passed, 0 failed.** `scripts/check_file_size.sh` passes.
</details>

## Summary by CodeRabbit

- **New Features**
  - Added configurable minimum release age and version-specific update skipping.
  - Update prompts now support updating, deferring, or skipping a specific version.
  - Update notifications can display release headlines.
  - Added cache validation to safely refresh outdated or incompatible update data.
  - Standardized update source names.

- **Bug Fixes**
  - Prevented automatic installation of releases with unknown or insufficient age.

- **Documentation**
  - Expanded update, installation, command, and environment-variable documentation, including signature verification and package-manager safeguards.

### DIFF
--- a/changelog.d/7786-no-cache-migrations.md
+++ b/changelog.d/7786-no-cache-migrations.md
@@ -1,0 +1,33 @@
+### Changed
+
+**The update cache carries its shape, and a foreign shape is discarded rather
+than migrated.** `~/.perry/update-check.json` now records a `schema` number; a
+value this build does not recognize — including its absence, which reads as `0` —
+means the file is thrown away and the next check rewrites it.
+
+This replaces the alternative, which was to keep every field optional forever so
+that older shapes still load. That trade is a bad one for a cache: it buys one
+saved network request in exchange for a set of `Option` fields that only exist
+to describe versions nobody runs, and that nothing ever removes. Bumping
+`CACHE_SCHEMA` is now the whole migration story.
+
+**One spelling per check source.** `github`, `npm-registry` and
+`github-packages` are gone; the names are `gh-releases`, `npm`, `gh-registry`
+and `custom`. A set of accepted aliases is a surface to document and test
+forever in exchange for saving one look at the docs.
+
+An unknown `source` still falls back to the default rather than failing, but for
+a different reason than compatibility: an update check is the wrong place to
+turn a config typo into a hard error.
+
+<details>
+<summary><b>Tests</b></summary>
+
+The test that asserted a pre-throttle cache still loads is replaced by one
+asserting the opposite — that a foreign schema, and an absent one, are both
+recognized as not-ours. Verified at runtime as well: a planted cache with no
+schema, claiming version `99.0.0` and a `last_check` in 2099, was ignored, a
+real check ran, and the file came back stamped `"schema": 1`.
+
+`cargo test -p perry`: 930 passed, 0 failed.
+</details>
--- a/changelog.d/7786-update-cooldown-skip-docs.md
+++ b/changelog.d/7786-update-cooldown-skip-docs.md
@@ -1,0 +1,53 @@
+### Added
+
+**`auto` waits a day before installing a brand-new release.** `min_age_hours`
+defaults to 24 for `auto` and 0 for every other mode, so a notice still mentions
+a release the moment it exists while an unattended install holds off.
+
+A release published by mistake, pulled shortly after, or published by someone
+who should not have been able to is most dangerous in its first hours. Waiting
+costs nothing and means your machine is not the one that finds out. `notify` and
+`prompt` are unaffected — they tell a human, who can decide.
+
+**An unknown publish date counts as too fresh, not as old enough.** The
+abbreviated npm document carries no dates, so treating unknown as "old enough"
+would switch the cooldown off for exactly the people using the cheapest source —
+a protection present in the config and absent in effect. `min_age_hours = 0`
+turns it off deliberately.
+
+**The prompt has three answers.** "No" and "never tell me about this one" are
+different intentions, and with only two answers a user who does not want one
+specific release has to switch the whole mode off — which then hides the release
+that fixes it. Answering the third writes `skip_version`, which suppresses
+exactly that version; the next release is mentioned normally.
+
+**The notice says what the release is,** not only that one exists. Sources that
+carry a title now pass it through, and it prints under the version line at no
+extra request.
+
+### Documentation
+
+New page `docs/src/cli/updates.md`, covering the default behaviour, everything
+in `[update]`, the four modes and their three refusals, the four check sources,
+the cooldown, skipping a version, exactly what a check transmits, and where the
+two files live. `perry update`'s section in `commands.md` is rewritten around
+`--mode`; `installation.md` gains the per-package-manager upgrade table; the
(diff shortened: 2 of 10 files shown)


jdalton:
Pushed as a single commit rebased on current `main`, carrying every fix from the three branches below it. `cargo test -p perry --bin perry` is at **934 passing, 0 failing**, and `scripts/check_file_size.sh` is clean.

<details>
<summary><b>The fixes in this branch</b></summary>

**The skip-a-version setting could erase the config that stores it.** Answering "skip this version" at the prompt writes `skip_version` to `~/.perry/config.toml` — and it did so by reading with `load_config`, mutating, and saving. `load_config` returns defaults both when there is no file and when the file does not parse, so on an unparseable config that sequence wrote defaults over the top and destroyed the license key and tokens. It now goes through `update_config_file`, which refuses to write when the read failed and tells you which file to fix.

**Everything from the branches below.** HTTPS enforced before any token is attached to a check request, both refresh paths carrying the notice state forward under the lock, `doctor` reporting your configured mode rather than the current run's effective one, `try_lock` at teardown instead of a blocking lock, and a timed-out background check falling back to the cache rather than being read as "no update".

</details>

<details>
<summary><b>What this branch adds on its own</b></summary>

| Feature | What it does |
|---|---|
| Release cooldown | `auto` mode will not install a release younger than the cooldown window, so a version that is published and then pulled minutes later cannot land on anyone's machine unattended. |
| Skip this version | The prompt takes a third answer that silences one release without silencing the next — you are not choosing between this nag and never hearing about updates again. |
| Headline in the notice | When the check source already returned the release title, the notice shows it. No extra request. |
| `docs/src/cli/updates.md` | One page for the whole surface: every setting, the environment variables that override them, how the throttle behaves, and a plain statement of what a check actually sends over the network. |

</details>

<details>
<summary><b>Why this branch contains three other commits</b></summary>

Last of four stacked pull requests, so [#7787](https://github.com/PerryTS/perry/pull/7787), [#7784](https://github.com/PerryTS/perry/pull/7784) and [#7785](https://github.com/PerryTS/perry/pull/7785) sit beneath it. Review the top commit; merging the three below first shrinks this diff to just the cooldown, the skip answer, and the docs page.

</details>

jdalton:
All findings across the stack are fixed and carried into this branch. Force-pushed as one commit on current `main`. `cargo test -p perry --bin perry`: **935 passing, 0 failing**; `scripts/check_file_size.sh` clean.

<details>
<summary><b>The two that were genuine bugs, not polish</b></summary>

**Auto mode was silently blocked by the notice throttle.** `notify_interval_hours` gated the whole teardown call, and that call is what performs the install in `auto` mode — so a notice printed an hour earlier stopped the new version from landing at all. The throttle is now applied to the resolved action rather than around the call, and `Install` is exempt. `the_notice_throttle_never_holds_back_an_install` covers it, and making the predicate return `true` unconditionally turns it red.

**An unreadable config still resolved to defaults.** `load_config_checked` treated every read failure as "no config yet", when only `ErrorKind::NotFound` means that; a permission or I/O error returned defaults that `update_config_file` would then write over a file still holding the license key and tokens. It now matches on the error kind.

That second test needed fixing before it proved anything: my first version only asserted "the write failed", which passed *with the fix reverted* because a file at mode `000` also fails to write. It now asserts the error names the read failure, and skips when running as root since root ignores mode bits.

</details>

<details>
<summary><b>The rest</b></summary>

| Finding | Fix |
|---|---|
| The old fallback ladder accepted an `http://` configured server | Both entries go through a shared `url_is_secure`; anything not HTTPS or loopback is dropped. A plaintext responder could otherwise hold a machine on an old build by reporting the running version as latest. |
| `configured_mode` collapsed `Unknown` to `Notify` | `Unknown`'s label is already "notify (unrecognized value in config)" — exactly what `doctor` should say about a typo. It is preserved for reporting now. |
| An empty `release_url` printed `or visit ` and `Release: ` | Both output paths omit the clause when the URL is empty. |
| Five fragments reported five different test totals | The count line is gone from all five, along with the forward references to later slices. Fragments already on `main` from other changes are untouched. |

</details>

PR #7787, opened 2026-08-10
Fix five review findings in the update surface that merged before their fixes

jdalton:
Five defects in the update surface, all found in review of [#7749](https://github.com/PerryTS/perry/pull/7749) — which merged before the fixes were pushed, so all five are live on `main` today.

One commit, based on current `main`.

### The config warning escaped the rules meant to silence it

The "unrecognized `[update] mode`" line was printed inside `UpdatePolicy::resolve`, before the precedence rules it sits behind had been applied. So it reached stderr during `--format json`, in CI, with a piped stderr, and under `--quiet`.

Those rules exist to keep exactly those runs silent, and the one line whose job was to report a config problem was the one line ignoring them. It is now held on the policy and emitted at the single point where the run is known to be speaking at all.

### The notify interval swallowed the next release

The documented contract is that `notify_interval_hours` throttles repeats of *the same* update. Keyed only on a timestamp, it also suppressed a **different** version arriving inside the window.

The consequence is the opposite of the intent: somebody setting a week-long interval to stop being nagged about one release would also have been denied the release that fixed it. The cache now records which version it announced, and a different version is announced regardless of the interval.

### The interval comparison was signed

`Duration::as_secs() as i64` goes negative for a large enough configured value, and a negative interval reads as already-elapsed — so an absurd value would have notified on **every** run instead of suppressing. Now unsigned.

### Two `perry` processes could corrupt the cache

Every write used one shared `*.json.tmp`. Two writers each wrote it and each renamed it, so the loser's rename landed a file the winner was still writing into.

Each write now builds its own temporary name from the process id and a per-process counter.

### A refresh could erase a notice recorded while its request was in flight

`fetch_latest_version` read the notice state *before* issuing its request and wrote it back afterwards, overwriting anything recorded in between — so the user would be told about the same release twice.

The read-modify-write pairs are now serialized by a lock file next to the cache, and the refresh re-reads inside that lock immediately before replacing. If the lock cannot be taken the caller proceeds unlocked: losing a cache update is better than refusing to update a cache, since the next run simply re-checks.

<details>

Two new contract tests, both **sabotage-verified** — reverting either fix turns its test red:

- a different version is announced regardless of the interval, and never having announced anything counts as "not this version";
- an enormous interval still suppresses rather than wrapping around into notifying.

The existing interval tests are unchanged in intent. They now call a helper that holds the announced version constant, so they still exercise only the interval arithmetic — which is what they were written for.

`cargo test -p perry`: **904 passed, 0 failed.**
</details>

<details>
<summary><b>Credit</b></summary>

All five were raised by CodeRabbit on [#7749](https://github.com/PerryTS/perry/pull/7749). I had replied there saying they were fixed; that was true of the branch and not of the merge, which is what this pull request corrects.
</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Configuration warnings now appear only when update notifications are visible.
  * Update checks can use cached status when background checks time out.
  * Notifications are no longer unnecessarily delayed when a newer version becomes available.
  * Update status is handled more reliably during simultaneous refreshes and writes.
  * Release URLs are shown only when available.

* **Tests**
  * Added coverage for version-specific notification behavior and unusually long notification intervals.

### DIFF
--- a/changelog.d/7787-update-surface-followups.md
+++ b/changelog.d/7787-update-surface-followups.md
@@ -1,0 +1,55 @@
+### Fixed
+
+Five defects in the update surface, all found in review of
+[#7749](https://github.com/PerryTS/perry/pull/7749) after it had merged.
+
+**The config warning escaped the rules that were meant to silence it.** The
+"unrecognized `[update] mode`" line was printed inside `UpdatePolicy::resolve`,
+before the precedence rules it sits behind had been applied — so it reached
+stderr during `--format json`, in CI, with a piped stderr, and under `--quiet`.
+Those rules exist to keep exactly those runs silent, and the one line whose job
+was to report a config problem was the one line ignoring them. It is now held on
+the policy and emitted at the single point where the run is known to be speaking
+at all.
+
+**The notify interval throttled on time alone, so it swallowed the next
+release.** The documented contract is that the interval throttles repeats of
+*the same* update. Keyed only on a timestamp, it also suppressed a **different**
+version that arrived inside the window — so somebody setting a week-long
+interval to stop being nagged about one release would also have been denied the
+release that fixed it. The cache now records which version it announced, and a
+different version is announced regardless of the interval.
+
+**The interval comparison was signed.** `Duration::as_secs() as i64` goes
+negative for a large enough configured value, and a negative interval reads as
+already-elapsed — so an absurd value would have notified on *every* run instead
+of suppressing. The comparison is unsigned.
+
+**Two `perry` processes could corrupt the cache.** Every write used one shared
+`*.json.tmp`, so two writers each wrote it and each renamed it: the loser's
+rename landed a file the winner was still writing into. Each write now builds
+its own temporary name.
+
+**A refresh could erase a notice recorded while its request was in flight.**
+`fetch_latest_version` read the notice state *before* issuing its request and
+wrote it back afterwards, overwriting anything recorded in between — telling the
+user about the same release twice. The read-modify-write pairs are now
+serialized by a lock file, and the refresh re-reads inside that lock immediately
+before replacing.
+
+<details>
+<summary><b>Tests</b></summary>
+
+Two new contract tests, both sabotage-verified — reverting either fix turns its
+test red:
+
+- a different version is announced regardless of the interval, and never having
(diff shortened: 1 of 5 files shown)


jdalton:
Pushed as a single commit rebased on current `main`. Every review finding on this pull request is now fixed in code, and each fix that could be proved by a test has one.

<details>
<summary><b>What changed, and why each one mattered</b></summary>

**The config file could be erased while you were fixing a typo.** `load_config` returns a default struct in two very different situations: there is no config file yet, and there is a config file that does not parse. A writer built on it cannot tell those apart, so a single stray character in `~/.perry/config.toml` meant the next setting-write serialized defaults over the top and took the license key and tokens with it. There is now one writer, `update_config_file`, which reads through `load_config_checked`, refuses the write when the read failed, and says so in a sentence that names the file and tells you to fix or delete it. `a_damaged_config_is_never_overwritten_with_defaults` proves it: restoring the old read-then-save shape turns that test red.

**The teardown cache write could hang someone's terminal.** The comment above `lock_cache` promised that a contended lock is skipped and the write proceeds unlocked, but the body called the blocking `lock()`. That code runs *after* the command the user asked for has already finished, so the worst case was one `perry` freezing another's shell while waiting to write a cache. It now uses `try_lock` and honours what the comment said.

**A slow check was being read as good news.** The startup check gets 100 ms to answer. When it did not, the notice path treated that silence as "no update available" and printed nothing — discarding a notice a previous run had already earned and stored. A timeout says the network was slow, not that your version is current, so it now falls back to the cached status.

</details>

jdalton:
The latest review round raised nothing against this branch, and the two findings it raised against the branches above it do not apply here: this layer has no config writer and no teardown action, so there is no install for a notice throttle to hold back and no writer to erase a config. The one change carried in is the changelog fragment — the suite-count line is gone, because each fragment in the stack recorded the total at the moment its own slice was written and the assembled release notes reported five different figures for one release.

Force-pushed as one commit on current `main`.

proggeramlug:
Closing as already-landed: rebased onto post-#7786 main, the commit drops out entirely (empty diff — `git rebase` discarded it as already-applied). All five fixes reached main through the #7784→#7786 stack's squashes. Nothing left to merge.

PR #7790, opened 2026-08-10
test(hir): pin the for-of desugar counter's literal kind at both emission sites (#7766 follow-up)

### LINKED ISSUE #7766: repsel: rule-1 provenance does not survive a function boundary — a typed `P[]` parameter still reads fields by name (first-party half of #7152/#7170)

proggeramlug:
Follow-up to #7778 (which closed #7766). **Tests only — no behaviour change.**

#7778 changed both `for…of` desugar sites to mint the synthetic `__idx_*` counter as `Expr::Integer(0)` instead of `Expr::Number(0.0)`. That literal *kind* is what `collectors/i32_locals.rs::collect_integer_let_ids` seeds on — it does not reason about provable integrality — so it is the difference between a counter that gets a canonical i32 slot and one that is structurally invisible to every i32-counter loop optimization. #7771's element-binding element-shape clone is the case that made it visible: its matcher hard-requires `ctx.i32_counter_slots`, so before the fix it could not fire for `for…of` at all (measured: 0.83s → 0.09s on a 200k-element parameter-array loop).

**Nothing pins it.** The desugar is correct either way and prints identical output, so an edit back to `Number(0.0)` would silently un-optimize every `for…of` loop in the language with no test failing — CLAUDE.md's fourth way a gate can be unable to fail ("the gate runs but its subject never did").

These are verdict tests on the lowered HIR, one per emission site — `lower/stmt_loops.rs` (module init) and `lower_decl/body_stmt.rs` (function bodies), which have drifted independently before. Both are **sabotage-verified**: reverting either site to `Number(0.0)` reds them.

for-in is deliberately not covered: its counter indexes a keys array rather than the receiver, and #7778 left it alone.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved regression coverage for `for…of` loops, helping ensure loop counters are initialized consistently and retain optimized integer behavior.

- **Documentation**
  - Added a changelog entry documenting the `for…of` counter behavior and its optimization expectations.

- **Chores**
  - Incremented the workspace version to `0.5.1459`.


proggeramlug:
Audit complete — merging.

- Both verdict tests pass on the rebased tree (v0.5.1459); full `perry-hir --lib` 293 green.
- Sabotage verified per site, applied-then-checked: reverting `lower/stmt_loops.rs` to `Number(0.0)` reds exactly the module-level test; reverting `lower_decl/body_stmt.rs` reds exactly the function-body test. The two-site split is real coverage — the sites have drifted independently before.
- The walker asserts non-emptiness before asserting the kind, so a desugar rename that stops minting `__idx_*` counters cannot make these vacuously green.
- One fix during audit: the changelog fragment was keyed `7783-*` (another PR's number) — renamed to `7790-*` per the PR-keyed convention.
- Local gate: one `perry-runtime --lib` red at load average 26–37 that re-ran clean 2/2 (2,051 passed) — the known load transient; all other steps green.

This closes the coverage gap named in #7778's merge audit: the `Integer(0)` desugar change previously had only end-to-end coverage.

Issue #1444, opened 2026-05-22
CI: perry-runtime --lib cargo-test is flaky (event_pump timing panics + intermittent SIGSEGV)


proggeramlug:
## Summary

The `cargo-test` CI job (`perry-runtime --lib`) fails intermittently on unrelated PRs, blocking merges and requiring `gh run rerun --failed`. Two distinct flaky signatures:

1. **event_pump timing panics** — `event_pump::tests::wait_returns_when_timer_due`, `::notify_before_wait_is_preserved`, `::sustained_budget_zero_spin_is_throttled` panic at `crates/perry-runtime/src/event_pump.rs:{480,511,612}`. Correlates with slow CI runs (observed a 1016s run vs ~451s normal) — the GitHub runner is overloaded and the wall-clock-deadline waits miss their timing windows.
2. **Intermittent SIGSEGV** — the `perry_runtime-*` test binary exits `signal: 11, SIGSEGV` with **no failing test name**, from the GC/threading tests. A local full `cargo test -p perry-runtime --lib` (536 tests) passes clean.

## Evidence

Both signatures hit perf_hooks-only PRs that touch **no** `event_pump`/GC code and have **no** unit-test coverage of their changed paths:
- PR #1437 (SIGSEGV), PR #1440 (SIGSEGV), PR #1443 (SIGSEGV on first run, event_pump timing panics on re-run). All passed on re-run / are unrelated to the diff.

## Suggested fixes

- Make the `event_pump` timer-due waits **deterministic** (inject a clock / use a controllable time source) instead of asserting against wall-clock deadlines, so they don't flake under CI load.
- Track down the GC/threading SIGSEGV (run `perry-runtime --lib` under `RUST_TEST_THREADS=1` + ASan in CI to localize).

## Workaround

Until fixed: `gh run rerun <run-id> --failed` when the only red check is `cargo-test` with these signatures.

proggeramlug:
Another flaky variant in the same `cargo-test` (`perry-runtime --lib`) job: `tui::tree::tests::register_increments_handle` (tree.rs:243, `assertion left == right failed` on a handle id). Looks order-dependent — a shared global handle-id counter polluted by parallel tests. Hit on PR #1453 (a symbol/timer change that doesn't touch tui handles). Same remedy: `gh run rerun <id> --failed`. A real fix would make the test assert a *delta* (id_after - id_before == 1) instead of an absolute id, or use RUST_TEST_THREADS=1 for the handle-counter tests.

[...]

PR #7791, opened 2026-08-10
docs: perry-runtime's tests must run single-threaded locally, as CI already does

proggeramlug:
## What

CLAUDE.md documented `cargo test --release --workspace …` for local runs. That command does **not** pin `RUST_TEST_THREADS`, so it runs `perry-runtime` multi-threaded — the exact configuration every CI path deliberately avoids. This splits the documented command so `perry-runtime` runs `RUST_TEST_THREADS=1`, and records the hazard under "CI gates that surprise people".

## Why — the investigation behind it

`gc::tests::root_words::bare_address_in_shadow_slot_survives_a_real_collection` was reported failing on a clean `c2a96b638` with the box at load ~111, passing in isolation and passing three full-suite reruns on a lightly loaded box. It was handed to me as "either make it deterministic, or it's a real load-dependent GC bug — and if so that outranks everything."

**It is not a GC bug.** `perry-runtime`'s tests share process-global side tables. `test.yml` has pinned `RUST_TEST_THREADS=1` for this crate since #1444 (its comment: the default pool "races the GC/threading tests into intermittent SIGSEGV"), and `gc::tests::global_sink_isolation`'s header records that ~180 readers still do not take the clearing lock, with #7672 converting them to `per_test_global!` one table at a time. The local command was the one place that still walked into it.

## Evidence

Against `2e5bf4434`, release, on a tree that includes #7317:

| configuration | runs | result |
|---|---|---|
| full suite, `--test-threads=16`, load ~90 | 4 × 150 | **600/600 clean**, 0 vacuous |
| full suite, `--test-threads=64`, load ~115 | 4 × 80 | **2 failures / 320** |

The two failures were `proxy::tests::object_array_numeric_write_guard_requires_complete_uniform_proof` and `array::element_shape::matrix_tests::matrix_delete_revokes` — neither is `root_words`. The failing test is simply whichever one loses the race.

The decisive detail is *what* they assert when they fail. Both are the crate's own vacuity guards:

- `fixture must start proven, or every verdict below is vacuous`
- `one-field loops should publish one non-zero 16-bit lane`

That is a test discovering its own **precondition** was destroyed by a concurrent thread — not a collector invariant being violated. A real GC bug would surface as a swept-while-live object, not as a fixture that never armed.

`root_words` itself did not fail once across:

- 920 parallel full-suite runs (release + debug),
- 1500 targeted `root_words` + `global_sink_isolation` pairings at `--test-threads=16`,
- a 40-seed `PERRY_GC_SCHEDULE_SEED` sweep at rate 25% — it never appears among that sweep's failures.

## Deliberately not done

No `#[ignore]`, no retry wrapper. Both would hide which of the two possibilities this was. The fix is to stop documenting a configuration the repo already knows is unsupported.

## Side note recorded in the changelog fragment

`PERRY_GC_SCHEDULE_SEED` is a whole-process env var and `scripts/gc_schedule_fuzz.sh` takes a *compiled binary*, not the unit-test suite. Setting it across `cargo test` fails 8 GC tests at every seed, several of which assert the unset behaviour outright (`schedule::unset_is_inert_for_evacuation_policy`). That is the tool used outside its contract, not a defect — written down so the next person doesn't re-derive it.

## Validation

`cargo fmt --all -- --check` and `scripts/check_file_size.sh` both clean. Docs-only change; no code paths touched.

## Summary by CodeRabbit

* **Documentation**
  * Documented that runtime tests run single-threaded to avoid failures caused by shared process-wide test state.
  * Clarified local and CI test commands, including that no tests are skipped or retried.
  * Added guidance on garbage-collection scheduling controls and stress-test behavior.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1460
+**Current Version:** 0.5.1461
 
 
 ## TypeScript Parity Status
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -82,7 +82,8 @@ cargo build --release                          # Build all crates
 cargo build --profile perry-dev -p perry       # Fast local dev build (#5422; perry-dev profile)
 cargo build --release -p perry-runtime -p perry-stdlib  # Rebuild runtime (MUST rebuild stdlib too!)
 cargo build --release -p perry-runtime-static -p perry-stdlib-static  # Emit libperry_{runtime,stdlib}.a (#5422: runtime/stdlib are now rlib-only; the .a comes from these wrapper crates)
-cargo test --release --workspace \
+RUST_TEST_THREADS=1 cargo test --release -p perry-runtime   # MUST be single-threaded (see below)
+cargo test --release --workspace --exclude perry-runtime \
   --exclude perry-ui-ios --exclude perry-ui-tvos --exclude perry-ui-watchos \
   --exclude perry-ui-visionos --exclude perry-ui-android --exclude perry-ui-windows \
   --exclude perry-ui-gtk4   # Run tests (exclude cross-host UI crates on macOS)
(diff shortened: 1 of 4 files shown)


proggeramlug:
Audit complete — merging.

Facts verified against the tree: `test.yml` pins `RUST_TEST_THREADS=1` for perry-runtime at both invocations (lines 753/806, plus the macOS-leg comment), `per_test_global!` exists with #7672's migration in progress, and the CLAUDE.md edit changes exactly the documented local command plus one CI-gates bullet. The investigation behind it (920 parallel full-suite runs isolating the failure to the harness's shared globals, with the vacuity-guard message as the discriminator) is the right way to close a phantom GC bug — and the "seeded-schedule sweep is per-binary, not per-test-suite" note will save the next person a real detour.

This session hit the same class four times today (gate reds at load 26–45 that re-ran clean), so the operational value is immediate. Bumped to 0.5.1461; gate 19/19 green (docs-only diff).

Issue #7717, opened 2026-08-09
gc: gc-moving-witnesses needs a dependency-scale RUNTIME witness (#7280 ask 3)


proggeramlug:
The one unmet ask from #7280, refiled on its own.

**Done and closed with #7280:** ask 1 (keep the stopgap) and ask 2 (fix the overclaiming comment).

**Still missing: ask 3 — a dependency-scale RUNTIME witness in `.github/workflows/gc-moving-witnesses.yml`.** Everything that gates this family today is either a synthetic fixture or a static check over emitted IR. Neither can observe the failure mode #7154's residual actually presents as.

## Why a runtime witness specifically

The #7154 class is invisible to every static instrument **and** to every runtime GC probe at the moment of collection — there is nothing for the collector to find. It surfaces cycles later, in a different function, as `TypeError: value is not a function`. The evidence that scale matters is in #7280 itself: the loop-polls configuration was **red 0/30**, and **stock zod alone failed 5/40** — i.e. real dependency-shaped code found what the fixtures did not.

`scripts/gc_root_dominance_check.py` reads LLVM IR, so a runtime-side cache holding a raw heap pointer is structurally invisible to it. That population now has its own inventory and gate (#7231 / #7695), but an inventory is not a witness.

## What the witness must assert, or it is worth nothing

This is the trap that has cost this repo the most, so treat it as the acceptance criterion rather than a nicety:

- **Prove the collector RAN.** Since #7604 a zeal run prints `[gc-zeal] forced_collections=N copying_minors=M moved_objects=K` at exit and **exits 70** if N or M is zero. Use it. A run with zero copying minors witnesses nothing, and `PERRY_GC_PROTECT_FROMSPACE=1` on such a run protects nothing while exiting clean — I hit exactly that today on #7528 and nearly recorded the wrong conclusion.
- **Compile with `PERRY_GC_MOVING_LOOP_POLLS=1`** (default off since #7161), or the loops emit no back-edge poll and zeal never fires inside them.
- **Watch `PERRY_GC_PROTECT_FROMSPACE_DEPTH`.** The default of 4 is far too small for real code — #7154's own reproducer needed **800**, because the value crossed 600 polls between its last valid observation and its stale use.

## The four ways this gate could ship unable to fail

[...]

PR #7793, opened 2026-08-10
test(gc): dependency-scale runtime witness for the moving collector (#7717)

proggeramlug:
**The moving collector now has a dependency-scale RUNTIME witness in CI** (#7717, the one unmet ask of #7280).

Everything gating this family was either a synthetic fixture or a static pass over emitted IR, and neither can observe what the #7154 class actually presents as. `scripts/gc_root_dominance_check.py` reads LLVM IR, so a runtime-side `static` or `thread_local!` holding a raw heap pointer is structurally invisible to it; and the defect is invisible to every runtime GC probe *at* the collection too, because there is nothing for the collector to find — it surfaces cycles later, in a different function, as `TypeError: value is not a function`.

Scale is the point, not size. #7280 measured it: the curated 25-file corpus passed 25/25 while twenty lines of stock `zod` failed 5/40, because dependency-shaped code is dominated by `js_object_assign_one` (object spread) and `js_new_function_construct` — populations the curated files barely produce.

`scripts/gc_dep_scale_witness.sh` runs `test-files/gc-dep-corpus/main.ts` — the same 81-module zod corpus `gc-root-dominance` already compiles for its static check, whose own header always said it "is run as the acceptance workload for the moving collector" — under `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800`, and is wired into `gc-moving-witnesses.yml`.

Measured on `main`: 8,020 forced collections, all 8,020 of them copying minors, 852,107 objects relocated, 63,936 back-edge polls, 8,020 retired quarantine page-sets, and the answer byte-identical to the same binary run without zeal.

**What makes it able to fail.** The trap #7717 names is a run that exits clean having witnessed nothing, so the script asserts its subject ran three independent ways: the `[gc-zeal]` exit verdict; `copying_minors` / `moved_objects` / `loop_polls` re-read from it rather than trusting that the binary still exits 70; and a non-zero count of retired from-space page-sets, because `PERRY_GC_PROTECT_FROMSPACE=1` on a run with no copying minor protects nothing and still exits 0. Both directions were verified by sabotage rather than asserted: with zeal disabled it fails on the missing verdict, and with the quarantine disabled it fails with "protected NOTHING and this run's cleanliness means nothing" — which is precisely the wrong conclusion #7717 records nearly recording.

Two findings worth keeping:

- **`PERRY_NO_AUTO_OPTIMIZE=1` is load-bearing here**, and is a deliberate departure from `gc_repsel_matrix.sh`, which omits it on purpose. Without it the compile reaches the auto-optimizer, which relinks the runtime as `features=async-runtime,web-fetch` — no `diagnostics`, and `diagnostics` is what emits the `[gc-fromspace-protect]` line this gate reads as its proof. The assertions are fail-closed so a stripped runtime goes red rather than quietly green, but red-for-the-wrong-reason is still a broken gate.
- **Depth 800 is load-bearing, not decorative.** The quarantine runs saturated on this workload (`sets_held=800/800`), so lowering it silently narrows the window in which a stale dereference is still catchable. The default of 4 is far too small for real code — #7154's own reproducer needed 800.

The gate is deliberately **not** added to branch protection's required contexts: a new gate has never been green, so promoting it immediately would block every open PR. It should be observed on `main` first, then promoted — leaving that second step undone is how `gc-stress` ended up reporting failures without blocking anything.

Also corrected while writing this: #7717 says loop polls are "default off since #7161". They are default **ON** since #7721 — `moving_safepoint_polls_enabled_from_env` is `!matches!(value, Some("0") | Some("off") | Some("false"))` — so the witness does not need to opt into them, and the run above confirms 63,936 polls fired without the flag.

## Validation

- `./scripts/gc_dep_scale_witness.sh` on `main` @ `0a2bf15bd`, release build: **PASS**, exit 0.
- Sabotage A — same script with `PERRY_GC_ZEAL=0`: exit 1, `::error::no [gc-zeal] verdict line was printed`.
- Sabotage B — same script with `PERRY_GC_PROTECT_FROMSPACE=0`: exit 1, `::error::the from-space quarantine never retired a page-set`.
- `bash -n` clean, workflow parses under `yaml.safe_load`, `scripts/check_file_size.sh` exit 0, no `continue-on-error` on any step of the job.
- Runtime ~5 min on top of the existing job, which already builds the compiler and the archives it needs.

## Summary by CodeRabbit

* **Tests**
  * Added automated CI validation for dependency-scale workloads and moving garbage collection.
  * Runs compare baseline and relocating-GC results, checking output consistency and required runtime behavior.
  * Failures now clearly report missing diagnostics, runtime faults, invalid results, or ineffective collections.

* **Documentation**
  * Added changelog documentation covering the new runtime validation, configuration, safeguards, and CI status.

### DIFF
--- a/.github/workflows/gc-moving-witnesses.yml
+++ b/.github/workflows/gc-moving-witnesses.yml
@@ -341,6 +341,46 @@ jobs:
                 "PERRY_GC_MOVING_LOOP_POLLS=1." % len(cells))
           PY
 
+      # ★★ The DEPENDENCY-SCALE RUNTIME witness (#7717, the one unmet ask of
+      # #7280).
+      #
+      # Everything above this line is a curated reproducer. #7280 measured why
+      # that is not enough: the curated corpus passed 25/25 while twenty lines
+      # of stock `zod` failed 5/40. It is a distribution problem rather than a
+      # size one — dependency-shaped code is dominated by `js_object_assign_one`
+      # (object spread) and `js_new_function_construct`, populations the curated
+      # files barely produce.
+      #
+      # `gc-root-dominance` already compiles this same corpus, but it is a
+      # STATIC pass over emitted IR and never runs a program, so it cannot see a
+      # runtime-side cache holding a raw heap pointer. This step runs it.
+      #
+      # The env knobs and every assertion live in the script, not here, so that
+      # reproducing a CI failure is one command rather than a re-read of this
+      # YAML. In particular the script pins `PERRY_NO_AUTO_OPTIMIZE=1`, which is
+      # a deliberate departure from the matrix above: the auto-optimizer relinks
+      # the runtime WITHOUT the `diagnostics` feature, and `diagnostics` is what
+      # emits the `[gc-fromspace-protect]` line this gate reads as its proof
+      # that the quarantine engaged.
+      - name: Install the npm dependency the dependency-scale witness runs
+        if: steps.relevance.outputs.run == 'true'
+        # `zod` is this repo's own package.json devDependency, pinned by
+        # package-lock.json and governed by the same soak window as everything
+        # else in that file — not a fixture invented for this job. Same
+        # invocation `gc-root-dominance` uses for the same corpus.
+        run: npm ci --ignore-scripts --no-audit --no-fund
+
+      # GATING. No pipe, no `|| true`. The script asserts its own subject ran
+      # three independent ways — the `[gc-zeal]` verdict, `copying_minors` /
+      # `moved_objects` / `loop_polls` re-read from it, and a non-zero count of
+      # retired from-space page-sets — because `PERRY_GC_PROTECT_FROMSPACE=1` on
+      # a run with no copying minor protects NOTHING and still exits clean.
+      # #7717 records hitting exactly that and nearly recording the wrong
+      # conclusion.
+      - name: Run the dependency-scale runtime witness
+        if: steps.relevance.outputs.run == 'true'
+        run: ./scripts/gc_dep_scale_witness.sh
+
       - name: Upload the witness report
         if: always() && steps.relevance.outputs.run == 'true'
         uses: actions/upload-artifact@v7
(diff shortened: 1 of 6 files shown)


proggeramlug:
Audit complete — merging, with one substantive change and one finding.

**The change: ported from `PERRY_GC_ZEAL` to the rate-1 seeded schedule.** #7741 retired the zeal knob between this PR's authoring and its landing; unported, the witness would have gone red on its own missing-verdict assertion — fail-closed, but red for the wrong reason is still a broken gate. The port swaps the env (`PERRY_GC_SCHEDULE_SEED=1 PERRY_GC_SCHEDULE_RATE=1`) and the `[gc-zeal]` → `[gc-schedule]` vocabulary; nothing else moves.

**Re-measured post-port** (release build of this tree = main's runtime): PASS, 8,022 forced collections, all copying minors, 852,411 objects relocated, 63,936 back-edge polls, 8,022 retired page-sets, answer byte-identical with and without the relocating collector — within noise of the PR's pre-port zeal numbers (8,020 / 852,107 / 63,936), which is itself a nice confirmation that rate-1 IS the retired instrument.

**Sabotage state after the port:**
- Schedule disabled: exit 1, `no [gc-schedule] verdict line was printed` — verified on the ported script.
- Quarantine disabled: exit 1, fail-closed — but not via the page-set assertion: the run **died at safepoint 3,319 with `TypeError: Cannot read properties of undefined`**, a #7154-shaped failure on main's runtime that the protected arm does not reproduce. Filed as **#7803** with the seed reproducer; it is a latent main bug this witness flushed out on its first unprotected run, not a defect of this PR. The page-set assertion's own two directions are both evidenced: its grep pattern demonstrably matches real output (the passing run counts 8,022 sets through the same pattern), and the author's pre-port sabotage produced the exact "never retired a page-set" failure.
- One process note: my first quarantine-off sabotage accidentally ran the unported script (an uncommitted-port + `git checkout` self-revert); the port is committed and the run above is on the committed script.

Also folded in: the workflow comment and fragment updated to the schedule vocabulary, with the port recorded in the fragment. The gate stays out of required contexts per the PR's own reasoning — observe on `main` first, then promote (#7737 item 4 pattern).

Closes #7717.

PR #7823, opened 2026-08-11
fix(runtime): give the primitive receiver to an accessor `toString` in toLocaleString

proggeramlug:
Refs #5901.

`Object.prototype.toLocaleString` is specified as `Invoke(O, "toString")`, which
expands to `GetV(O, "toString")` -> `ToObject(O).[[Get]]("toString", O)`. That
third argument is the **receiver**, and it is the *original primitive* — not the
wrapper object the property lookup walked through on its way to the property.

Perry resolved the method with `js_object_get_field_by_name(proto_ptr, key)`, a
plain get **on the prototype**. For a data property the receiver is
unobservable, so this was invisible. For an **accessor** it runs the getter with
the prototype as `this`:

```ts
Object.defineProperty(Boolean.prototype, "toString", {
  get: function () { const v = typeof this; return function () { return v; }; },
});
console.log((true as any).toLocaleString());
```

| | before | after | node |
|---|---|---|---|
| `Boolean.prototype` accessor | `object` | `boolean` | `boolean` |
| `String.prototype` accessor | `object` | `string` | `string` |

test262 pins this as
`built-ins/Object/prototype/toLocaleString/primitive_this_value_getter.js`,
whose sidecar reason was `Expected SameValue(«"object"», «"boolean"») to be
true`.

## Fix

Resolve an accessor explicitly and invoke its getter through
`call_primitive_closure_value` — the same helper the callers already use for the
resolved method. It hands a raw primitive `this` to a strict callee and a boxed
wrapper to a sloppy one, which is exactly the distinction the spec draws for the
getter's own `this` (module code is strict, so the test's getter keeps the raw
primitive).

Both resolvers on this path share the new `builtin_proto_accessor_method`
helper — the resolve-and-call `call_primitive_builtin_prototype_method` and the
resolve-only `builtin_proto_user_method` — so the two cannot drift apart.

## Not addressed here (pre-existing, and not this rule)

A `toString` that resolves to a **non-callable** should throw (`Invoke` ->
`Call` on a non-callable); Perry falls through to the native method and returns
its result. That is independent of the receiver rule and reproduces identically
through a plain **data** property:

```ts
Object.defineProperty(String.prototype, "toString", { value: 42 });
("abc" as any).toLocaleString();   // node: TypeError   perry: "abc"
```

The resolver collapses "absent" and "present but not callable" into one
"native behavior still applies" answer, and `%TypedArray%.prototype.toLocaleString`
depends on that same contract — so it is called out in the new test file rather
than changed under this fix.

## Validation

- New gap test `test-files/test_gap_object_tolocalestring_primitive_receiver.ts`
  is **byte-identical** to `node --experimental-strip-types` (accessor on
  `Boolean`/`String`/`Symbol` prototypes, an object receiver, and the ordinary
  data-property path). Its first two lines print `object` on the pre-fix
  runtime, so it fails without the change.
- `cargo test -p perry-runtime`: 2051 passed, 0 failed.
- test262 `built-ins/Object` subset — see the comment below.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `Object.prototype.toLocaleString` behavior for Boolean, String, Symbol, and object values.
  * Accessor-based and customized `toString` implementations now receive the correct original receiver.
  * Preserved expected formatting behavior for typed-array elements and built-in prototype methods.

* **Tests**
  * Added coverage for primitive and object receivers, including accessor and data-property implementations.

* **Documentation**
  * Documented the updated receiver behavior and the remaining handling distinction for non-callable `toString` values.

### DIFF
--- a/changelog.d/7823-tolocalestring-primitive-receiver.md
+++ b/changelog.d/7823-tolocalestring-primitive-receiver.md
@@ -1,0 +1,21 @@
+**`Object.prototype.toLocaleString` now gives an accessor `toString` the primitive receiver.**
+The spec routes it through `Invoke(O, "toString")` → `GetV(O, "toString")` →
+`ToObject(O).[[Get]]("toString", O)`, and that third argument is the *receiver*:
+the original primitive, not the wrapper object the lookup walked through. Perry
+resolved the method with a plain get **on the prototype**, which is
+unobservable for a data property but runs an **accessor**'s getter with the
+prototype as `this` — so
+`Object.defineProperty(Boolean.prototype, "toString", { get() { … } })` saw
+`typeof this === "object"` where the spec requires `"boolean"`
+(test262 `built-ins/Object/prototype/toLocaleString/primitive_this_value_getter.js`).
+An accessor is now resolved explicitly and its getter invoked through
+`call_primitive_closure_value`, the helper the callers already use for the
+resolved method: it hands a raw primitive `this` to a strict callee and a boxed
+wrapper to a sloppy one. Both resolvers on the path — the resolve-and-call
+`call_primitive_builtin_prototype_method` and the resolve-only
+`builtin_proto_user_method` — share the new helper so they cannot drift.
+Covered by `test-files/test_gap_object_tolocalestring_primitive_receiver.ts`.
+Still divergent, and independent of this rule: a `toString` resolving to a
+**non-callable** should throw, while Perry falls through to the native method —
+that reproduces identically through a plain data property, because the resolver
+collapses "absent" and "present but not callable" into one answer. (#5901)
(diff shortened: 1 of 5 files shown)


proggeramlug:
test262 `built-ins/Object` subset, run against this branch:

```
pass           3149      (was 3148)
diff           0         (unchanged)
runtime-fail   24        (was 25)
compile-fail   0         (unchanged)
skip           0         (unchanged)
judged         3173
parity:        99.2%
```

Exactly one test moved: `built-ins/Object/prototype/toLocaleString/primitive_this_value_getter.js` is gone from `report.failures.txt`, and no other case changed category — `diff`, `compile-fail` and `skip` all stayed at 0, so nothing was traded for it.

The one remaining `toLocaleString` failure in that directory is `S15.2.4.3_A1.js` ("no output"), which is unrelated to the receiver rule and was already failing before this change.

Also: `cargo test -p perry-runtime` — 2051 passed, 0 failed.

PR #7824, opened 2026-08-11
fix(ci): unbreak the thread-local policy gate, and stop counting `#[cfg(test)]` declarations

### LINKED ISSUE #7814: check_thread_locals.py is red on main (4 raw thread_local! in hot paths + a stale gc/zeal.rs entry), blocking tls-budget's promotion to a required gate
## Summary

`scripts/check_thread_locals.py` **fails on current `main`** (exit 1, 6 violations). Because it is deliberately not yet in branch protection's required contexts, nothing surfaces this — and the red state is precisely what blocks the documented promotion step, so the gate stays advisory indefinitely.

```
$ python3 scripts/check_thread_locals.py; echo "exit=$?"
thread-local policy check FAILED:
  crates/perry-runtime/src/array/indexing.rs: 1 raw `thread_local!` block(s), none allowed.
  crates/perry-runtime/src/map.rs:            1 raw `thread_local!` block(s), none allowed.
  crates/perry-runtime/src/registry_latch_probes.rs: 1 raw `thread_local!` block(s), none allowed.
  crates/perry-runtime/src/set.rs:            1 raw `thread_local!` block(s), none allowed.
  crates/perry-runtime/src/gc/zeal.rs: recorded as having 3 cold `thread_local!` block(s),
                                       but has none. A stale entry is one nobody has to justify.
exit=1
```

Measured on `b9415d780` (current `main` at time of filing).

## Two distinct problems

**1. Four real violations.** `array/indexing.rs`, `map.rs`, `set.rs`, `registry_latch_probes.rs` use raw `thread_local!` where the policy requires `crate::perry_thread_local!`. On Darwin every raw `thread_local!` access is an out-of-line call to `_tlv_get_addr` in libdyld — the exact cost `tls-budget.yml` exists to prevent, and which its own header records as having come back **three times** (up to 20.5% of `asyncpipe`, the largest single symbol in the worst-performing realistic program). `array/indexing.rs`, `map.rs` and `set.rs` are hot paths.

**2. A stale entry left by #7741.** `gc/zeal.rs` is recorded as having 3 cold blocks and no longer exists — #7741 removed `PERRY_GC_ZEAL` entirely. The allowlist's own contract is that a stale entry FAILS (so a fix must delete its exemption), which is working as designed; it just was not done at removal time.

## Why this is worth fixing promptly rather than allowlisting

`tls-budget.yml` documents, in its own header, that it is designed against CLAUDE.md's "four ways a gate can be unable to fail", and explicitly defers promotion to required-context status until after "the first observed green run on `main`". **That green run cannot happen while `main` is red.** The gate is currently in the state its own design notes were written to avoid: it runs, it reports failure, and it cannot block anything.

Note also that `--update` is the wrong repair here: it would silently absorb the four real violations along with the stale entry. The stale entry should be deleted and the four conversions done.

## Suggested order

1. Delete the `gc/zeal.rs` allowlist entry (#7741 follow-up, trivial).
2. Convert the four raw `thread_local!` blocks to `crate::perry_thread_local!` — same syntax, same `.with()` at call sites — or record any that are genuinely cold with a written justification.
3. Once `main` is green, promote `tls-budget` to a required context, which is the step the workflow's header already specifies as a maintainer action.

## Found by

The `globalThis` bootstrap investigation (2026-08-10), which reported the gate red as a side observation while validating an unrelated change. Confirmed independently against `main` with the script's own exit code (not a pipeline's).

proggeramlug:
Fixes #7814.

`scripts/check_thread_locals.py` is red on `main` (exit 1). Because it is
deliberately not yet in branch protection's required contexts, nothing surfaces
it — and that red state is exactly what blocks `tls-budget`'s documented
promotion step, so the gate stays advisory indefinitely. It is the shape
CLAUDE.md calls gate theatre: it runs, it reports failure, and it cannot block
anything.

## Two of the eight raw blocks are real — and they are not the ones reported

The ticket names `array/indexing.rs`, `map.rs`, `set.rs` and
`registry_latch_probes.rs` as violations "in hot paths". **All four are
`#[cfg(test)]`.** They do not exist in a shipping build and cannot cost a single
`_tlv_get_addr`. (`registry_latch_probes.rs` is not even compiled outside tests —
`lib.rs:143` declares it `#[cfg(test)] mod`.)

The genuinely live declarations are in `gc/schedule.rs`, which the ticket does
not mention:

| declaration | gated? | on what path |
|---|---|---|
| `SAFEPOINT_COUNTER` | no | bumped on **every safepoint** |
| `SCHEDULE_NEXT_CANDIDATE_BYTES` | no | read on **every poll** |
| `SCHEDULE_OVERRIDE` | `#[cfg(test)]` | — |
| `SCHEDULE_STRIDE_OVERRIDE` | `#[cfg(test)]` | — |

The two ungated ones are converted to `crate::perry_thread_local!`.

## `#[cfg(test)]` declarations are out of scope by construction

Recording a test-only declaration as "cold" records the *wrong fact* — it is not
cold, it is **absent** — and spends the allowlist's credibility on entries nobody
can ever act on. The scan now skips three statically-visible gated shapes:

* `#[cfg(test)]` directly above the block,
* an inline `#[cfg(test)] mod … { … }` enclosing it,
* the block's whole file being declared `#[cfg(test)] mod <stem>;`, closed
  transitively so `gc/tests/mod.rs` carries its subtree.

Unlike hot-vs-cold, this is a static fact the scan can actually see.

## The new rule is the kind that fails quietly, so it is sabotage-tested

An over-broad exclusion makes the gate pass **by seeing less**, which looks
identical to a fix. `--self-test` therefore grew two cases per shape: gated is
not counted, **and removing the gate makes the same declaration fail again**. It
now proves the checker can fail in six directions rather than four, and the
whole-file rule is checked in both directions too (`#[cfg(test)] mod probes;`
excluded, plain `mod probes;` back in scope).

## The regeneration is provably one-way

```
entries 101 -> 91   raw blocks 157 -> 129
ADDED: none
GREW:  none
```

Asserted mechanically against the pre-image, because `--update` absorbing a real
violation is the failure mode the ticket warns about. Three dropped entries were
already stale: `gc/zeal.rs` (left by #7741's removal of `PERRY_GC_ZEAL`),
`arena/quarantine.rs` and `gc/oldgen_defrag.rs`.

`_hot_declarations` moves 157 -> 160, and only two of those three are mine: the
recorded value was **already one behind the tree**, since nothing compares it
except the capacity ceiling (160 vs 768).

## Validation

- `scripts/check_thread_locals.py` → `thread-local policy OK: 160 hot
  declarations, 129 raw blocks in 91 recorded cold files, capacity 768`, exit 0.
- `scripts/check_thread_locals.py --self-test` → passes in all six directions.
- `cargo build -p perry-runtime` clean; `cargo test -p perry-runtime
  -- --test-threads=1` — see the comment below.

## Left for the maintainer, and the point of the exercise

Promote `tls-budget` to a required context after its first green run on `main`.
The workflow's own header specifies this step, and CLAUDE.md's corollary is that
leaving it undone is how `gc-root-dominance` sat red for weeks.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved thread-local policy validation so test-only declarations no longer trigger production checks.
  - Updated garbage-collection scheduling state to comply with the supported thread-local policy without changing behavior.
  - Removed obsolete policy exceptions and refreshed the approved declaration list.

- **Tests**
  - Expanded self-tests to verify both accepted and rejected declaration patterns across additional scenarios.
  - Added coverage for test modules and files included indirectly by test-only code.


proggeramlug:
`cargo test -p perry-runtime -- --test-threads=1`: **2051 passed, 0 failed**, 4 ignored (exit 0). `cargo build -p perry-runtime` clean.

```
$ python3 scripts/check_thread_locals.py --self-test
self-test: the checker can fail in all six directions
$ python3 scripts/check_thread_locals.py; echo "exit=$?"
thread-local policy OK: 160 hot declarations, 129 raw blocks in 91 recorded cold files, capacity 768
exit=0
```

PR #7832, opened 2026-08-11
fix(json): parse deep nesting with a heap-backed stack

### LINKED ISSUE #7817: json: make JSON.parse iterative so nesting depth is not bounded by thread stack size
## Summary

Perry's `JSON.parse` descends one function call per nesting level, so how deep a document it can read is bounded by how much stack the thread has. [#7816](https://github.com/PerryTS/perry/pull/7816) stops that from being a crash — input past 1,000 levels now throws a catchable `RangeError` instead of taking the process out with SIGSEGV — but it does not close the gap against Node, and this issue is that gap.

| | node 26.5.1 | perry after [#7816](https://github.com/PerryTS/perry/pull/7816) |
|---|---|---|
| 1,000 levels | parses | parses |
| 1,001 levels | parses | `RangeError` |
| 300,000 levels | parses | `RangeError` |

Node reads all of these because V8's JSON parser is **iterative** — it keeps its nesting state on the heap, so depth costs memory rather than stack. Perry's is recursive, so depth costs stack, and stack is the scarcest resource in the process.

## Why the limit is as low as it is

It is sized for the smallest stack Perry parses on, not the biggest.

On the main thread's 8 MB stack the crash lands somewhere between 20,000 and 40,000 levels. But Perry also parses JSON on `perry/thread` workers and tokio workers, and a 2 MiB thread stack gives out far sooner. I found that the hard way: [#7816](https://github.com/PerryTS/perry/pull/7816) first used a limit of 10,000 taken from a main-thread measurement, and its own unit test crashed the test harness at 9,999 levels.

So 1,000 is not a considered parity target — it is what a small stack can carry with room to spare. Any recursive parser has this shape: the safe limit is set by the worst stack in the process, and it cannot be raised by measuring on the best one.

## What closing it takes

Both parsers that read the text recurse, so both have to change:

<details>
<summary><b>The two descents, and the constraint each one carries</b></summary>

**Perry's value parser** (`crates/perry-runtime/src/json/parser.rs`) — `parse_value` calls `parse_array` and `parse_object`, which call `parse_value` again. Making it iterative means an explicit stack of in-progress containers on the heap, with each completed value pushed into its parent.

The GC contract is the part to get right rather than the control flow. The parse currently runs under `gc_suppress()` with intermediates held in `PARSE_ROOTS`, and an explicit work stack holds partially-built arrays and objects across every step. Those need to stay reachable for the same reasons the recursive version's locals do, and the recent from-space quarantine work is a reminder of how a half-built object that is rooted late reads as fine right up until it does not.

**The `serde_json` validation pass** (`js_json_parse` / `js_json_parse_result`) — `serde_json::from_slice::<IgnoredAny>` recurses per level too, so an iterative Perry parser alone would not raise the ceiling by one level. The options are to replace that pass with a non-recursive validator, or to fold validation into the iterative parser so the text is read once instead of twice. Folding it in is appealing beyond depth: it would drop a full pass over the input.

</details>

<details>
<summary><b>What "done" should mean</b></summary>

- A document nested well past today's limit — 300,000 levels, matching the original report in [#7792](https://github.com/PerryTS/perry/issues/7792) — parses to the same value Node produces.
- It parses on a **worker thread**, not just the main thread, since that is the case that set the current limit. A test that only exercises the main thread would repeat the mistake that made 10,000 look safe.
- The depth guard and its `RangeError` either come out entirely, or stay only as a memory-pressure bound with a limit that no longer depends on stack size.
- No regression on ordinary payloads. Most real JSON is a few levels deep and heavily object-shaped, so the shape to watch is many small records rather than deep nesting — an iterative parser should not cost anything there, but "should not" is worth measuring.

</details>

## Scope note

Not urgent. The crash is what mattered, and [#7816](https://github.com/PerryTS/perry/pull/7816) handles it; what is left is a parity gap on documents that are already unusual. Filing it so the 1,000 limit is a recorded decision with a reason attached rather than a number someone finds later and wonders about.

Refs [#7792](https://github.com/PerryTS/perry/issues/7792), [#7816](https://github.com/PerryTS/perry/pull/7816).

proggeramlug:
## Summary

Closes #7817. JSON.parse now switches from the recursive fast path to the existing flat tape representation once nesting passes the worker-stack threshold, then materializes that tape with an explicit heap-backed work stack. The reported 300,000-level document parses on a 2 MiB worker stack.

A separate 500,000-level iterative nesting budget keeps resource use finite and reports a catchable RangeError.

## Changes

- make tape validation strict enough to own the deep-input path without the recursive serde validation pass
- add iterative tape materialization for arrays, objects, duplicate keys, strings, numbers, booleans, and null
- route all three parse entries through the same recursive/iterative decision
- retain the optimized recursive and typed-shape paths for ordinary payloads

## Related issue

Closes #7817. Follows the crash guard in #7816.

## Test plan

- baseline reproduction: the old 1,001-level test passes on main by asserting that parsing returns an error
- cargo test -p perry-runtime json::tests::nesting_depth -- --nocapture
- cargo test -p perry-runtime json::tests -- --nocapture
- cargo test -p perry-runtime json_tape::tests -- --nocapture
- full perry-runtime suite: 2,108 passed; one unrelated promise keyed-table timing assertion failed under concurrent local load and passed immediately when rerun alone
- cargo fmt --all -- --check
- git diff --check
- scripts/check_file_size.sh still reports only the two pre-existing main offenders fixed by #7830

The quiet-mini benchmark is pending because its required lock is held by strcat-agent; no performance numbers are claimed here yet.

## Screenshots / output

N/A (runtime parser change).

## Checklist

- [x] Added regression tests
- [x] Added changelog.d/7832-iterative-json-parse.md
- [x] Kept ordinary JSON on the existing optimized path
- [x] No version bump or CHANGELOG.md edit
- [ ] Run ordinary-payload timing on the quiet mini when its lock is free

## Summary by CodeRabbit

* **New Features**
  * JSON parsing now supports deeply nested documents, including inputs with up to 500,000 levels of nesting.
  * Deep JSON structures can be parsed reliably even with limited stack space.

* **Bug Fixes**
  * Improved validation rejects malformed numbers, invalid escapes, control characters, trailing content, and incomplete structures.
  * Excessively deep input now returns a clear `RangeError` instead of risking parser failure.

### DIFF
--- a/changelog.d/7832-iterative-json-parse.md
+++ b/changelog.d/7832-iterative-json-parse.md
@@ -1,0 +1,4 @@
+**`JSON.parse` now handles deeply nested documents without exhausting a worker
+thread's native stack.** Inputs beyond the recursive fast path use strict tape
+validation and heap-backed iterative materialization, with a separate
+500,000-level resource limit.
--- a/crates/perry-runtime/src/json/mod.rs
+++ b/crates/perry-runtime/src/json/mod.rs
@@ -824,7 +824,7 @@ mod tests {
         }
     }
 
-    /// #7792 — deeply nested input must throw, not take the process out.
+    /// #7792 / #7817 — deeply nested input must not take the process out.
     ///
     /// Both parsers that read the document recurse once per nesting level, so
     /// a deep enough document exhausted the stack: SIGSEGV, exit 139, no
@@ -833,7 +833,17 @@ mod tests {
     /// document is unusual.
     mod nesting_depth {
         use super::*;
-        use crate::json::parser::{nesting_depth_exceeds, MAX_NESTING_DEPTH};
+        use crate::json::parser::{
+            nesting_depth_exceeds, MAX_ITERATIVE_NESTING_DEPTH, MAX_RECURSIVE_NESTING_DEPTH,
+        };
+
+        fn nested_arrays(depth: usize, leaf: u8) -> Vec<u8> {
+            let mut input = Vec::with_capacity(depth * 2 + 1);
+            input.extend(std::iter::repeat_n(b'[', depth));
+            input.push(leaf);
+            input.extend(std::iter::repeat_n(b']', depth));
+            input
+        }
 
         #[test]
         fn the_scan_counts_only_structural_brackets() {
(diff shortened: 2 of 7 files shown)


proggeramlug:
Quiet-M1 ordinary-payload performance check completed against base 1ee158d27 and PR head 642f653a2.

Methodology:
- Built perry, perry-runtime-static, and perry-stdlib-static for both arms from exact staged snapshots with an external target directory.
- Compiled every subject with the same saved base compiler, changing only PERRY_RUNTIME_DIR.
- Confirmed the base/fix runtime and stdlib archive hashes differed.
- Ran one warmup plus 12 alternating A/B pairs per subject; all 48 pairs exited 0 with matching checksums.

Median wall-time results:
- json_parse_1mb: 0.203713 s -> 0.204463 s, +0.368% (paired ratio +0.264%).
- bench_json_readonly: 0.242591 s -> 0.242716 s, +0.052% (paired ratio +0.036%; internal loop 0.000%).
- bench_json_roundtrip: 0.249871 s -> 0.250551 s, +0.272% (paired ratio +0.272%; internal loop +0.227%).
- bench_json_typed_roundtrip: 0.707881 s -> 0.707644 s, -0.034% (paired ratio +0.040%; internal loop 0.000%).

Ordinary recursive, lazy, roundtrip, and typed paths are all effectively flat. Marking ready for review.

PR #7834, opened 2026-08-11
perf(codegen): stamp a pointer-free shape's typed layout into the allocation header

proggeramlug:
Four allocation benchmarks sat within a **6% band** at 2.50–2.65× node — object literals, class instances, cyclic graphs, retained records. Four structurally different shapes do not land in a 6% band by coincidence; one shared per-allocation cost does.

## What it was

Profiling the band (200M-allocation variants with `PERRY_DEBUG_SYMBOLS=1`, two samples per program agreeing within 1.5 pp):

| bucket | share of `churn_alloc` | ns/alloc |
|---|--:|--:|
| `js_gc_declare_typed_shape_layout` (+ its `layout_forget_object` tail) | **30%** | 6.4 |
| the synthesized constructor's two guarded field stores | 27% | 5.7 |
| `chunk` itself (loop, call overhead) | 21% | 4.4 |
| `js_ctor_return_override` | 8% | 1.7 |
| `js_array_numeric_value_to_raw_f64` | 6% | 1.2 |
| array push + grow | 5% | 1.0 |
| **GC** | **3.5%** | 0.7 |

GC pause is not the cost — 3.5%. The cost is that we do ~9 out-of-line things per allocation where V8 does a bump and two stores.

The largest one spent its 30% re-deriving **per object** a fact that is a property of the **shape**. #7510's memo had already reduced the map round-trip to a direct-mapped probe; what was left was the probe, a type-table lookup, a field-count compare, and the cross-crate call itself.

## What this does

For a shape whose pointer mask is **statically empty**, the canonical layout is a constant — `GC_LAYOUT_POINTER_FREE | GC_OBJ_TYPED_LAYOUT_INTACT`. The inline-bump `new` path already emits a packed `GcHeader` constant carrying the state half, so the intact bit folds into the same store and the call goes away. What survives is the half that depends on the recycled **address** rather than the shape — clearing a previous tenant's per-object record — as a one-argument `js_gc_forget_object_layout` behind a `PERRY_PER_OBJECT_LAYOUTS_ANY` test whose `0` state proves every thread's per-object tables empty.

Two smaller levers in the same band:

- **`js_ctor_return_override`** answered `undefined` for every constructor without an explicit `return`. `JSValue::is_undefined` is `bits == TAG_UNDEFINED`, so one 64-bit compare decides it inline; the runtime call stays on the cold arm where derived-constructor `TypeError`s, object returns, arguments objects and arrays still need it.
- **A `new` in a hot-loop callee is a `new` in a loop**, one frame out, so it takes the inline bump. `cycles.ts`'s `makeCycle` is the shape: 5 statements, therefore `alwaysinline`, therefore never `inlinehint` — so the existing gate was reading the one flag it could not have been set.

## Measured — absolute seconds, quiet M1 mini, best-of-5

Outputs byte-identical to `node --experimental-strip-types` **and** exit code 0 verified for all 27 programs before timing.

| bench | before | after | Δ |
|---|--:|--:|--:|
| churn | 0.4217 | **0.2900** | −31% |
| churn_alloc | 0.3720 | **0.2409** | −35% |
| push_cls | 0.3665 | **0.2368** | −35% |
| cycles | 0.1939 | 0.1885 | −3% |

`churn_alloc` goes **18.6 → 12.0 ns per allocation**; node is 7.1 ns on the same shape.

No regression on the protected set: `churn_read` 0.0224, `fib40` 0.3941, `deeplist` 0.1218, `push_num` 0.1435, `retain` 0.3450, `retain1` 0.1346, `retain_wide` 0.4535, `retain_wide1` 0.1570, `tree` 1.6097, `tree_wide` 2.0850, `asyncpipe` 0.1328, `shapes` 0.1887, `interp` 1.4877, `iso_miss` 1.9174, `pipeline` 0.5251.

## The probe

`gc-handoff/bench/alloc_declare_pf.ts` and `alloc_declare_ptr.ts` are the same program — same allocation count, same field count, same runtime stores — differing only in whether the second field's *declared type* makes the shape's pointer mask non-empty. Before, both arms pay the declare and their times match; after, only the control does, and the gap is the cost that was removed:

| arm | before | after |
|---|--:|--:|
| `alloc_declare_pf` (gets the bake) | 0.9044 | **0.5802** |
| `alloc_declare_ptr` (control) | 0.8647 | 0.7845 |

Both arms move by the shared return-override lever (1.6 ns/alloc); the extra 4.1 ns/alloc on the `pf` arm is the layout declare itself.

## Soundness

The collector's view is **bit-identical**. `heap_payload_slot_selection` skips a `GC_LAYOUT_POINTER_FREE` payload without consulting any map, and the pre-existing path also reached `POINTER_FREE` for an empty pointer mask. A later pointer store still downgrades: with no descriptor to classify against, `layout_note_slot` falls through to its generic pointer-mask branch, which mints a per-object mask and flips the state to `SIDE_MASK` — that branch needs no descriptor at all.

A pointer-**bearing** shape keeps the full runtime declare, and must: its `SIDE_MASK` state means the tracer reads a mask, and that call is what installs the shared `SHAPE_LAYOUTS` descriptor the mask lives in.

**One hypothesis refuted rather than assumed.** I expected "INTACT set with no descriptor installed" to be a wrong-answer hazard — a non-number reaching a raw-f64 slot through the boxed setter, with nothing left to downgrade the object. It is not, and the reason is worth recording: a JS number's NaN box *is* its double bits, so the raw-f64 claim never changes the storage; every site that writes a slot raw re-proves the value finite inline; and every read that could treat the bits as a machine double is itself value-guarded. Tested directly (`arr[3].a = true` then `arr[3].a + 1` through a warmed monomorphic accessor): both arms print node's `2`, byte-identical.

## Validation

- Corpus: 27 programs, exit 0 + byte-identical output, both arms.
- Canary `gc-handoff/apps/iso_miss.ts` → `checksum 437840 misses 0`, plain and under `PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800`, `PERRY_GC_VERIFY_EVACUATION=1`, `PERRY_GC_SCHEDULE_RATE=1 PERRY_GC_SCHEDULE_ALLOC_KB=0`.
- GC stress sweep over 11 corpus programs × 3 knob configurations: all byte-identical, exit 0.
- `cargo test -p perry-runtime --lib` (`RUST_TEST_THREADS=1`): 2105 passed, 0 failed.
- `cargo test -p perry-codegen --lib`: 859 passed, 0 failed.
- Gap suite: partial locally (host is shared with several agents and the run was SIGTERM'd at 159/535). Of the 3 parity fails recorded, two are in `test-parity/known_failures.json` and the third (`test_gap_4510_enum_forward_ref`) is a **node**-side failure — node itself errors on the file — with Perry printing identical output on both arms. Full run left to CI.

## Tests

`crates/perry-codegen/src/lower_call/typed_shape_bake_tests.rs` — three IR-census ratchets, of the "assert the subject was live" kind. They are not merely claimed to work: the first version of them **failed**, because the probe's `new` had been scalar-replaced away and there was no allocation to assert about. The escape is now explicit and commented, for the next person who writes a probe here.

## Not done, with sizes

Two levers measured and deliberately left: the constructor's two guarded field stores (now ~45% of `churn_alloc`; every guard condition is statically true at a site that allocated the object three instructions earlier, but hoisting them crosses the standalone-`<Class>_constructor` symbol boundary), and `js_array_numeric_value_to_raw_f64` (~9%; provably the identity on the inline-precheck edge, which already tested the value finite, but the fast block is shared with the guard-call edge).

## Summary by CodeRabbit

* **Performance**
  * Improved allocation efficiency for eligible objects by embedding typed layout information directly during allocation.
  * Optimized hot-loop execution and allocation handling for frequently called functions.
  * Reduced overhead when cleaning up object layout metadata, especially when no cleanup is required.

* **Bug Fixes**
  * Improved constructor return handling while preserving correct behavior for derived constructors and primitive values.
  * Enhanced garbage-collection layout tracking and cleanup for recycled objects.
  * Preserved correct behavior across pointer-free and pointer-bearing object layouts.

### DIFF
--- a/changelog.d/7834-alloc-band-typed-shape-bake.md
+++ b/changelog.d/7834-alloc-band-typed-shape-bake.md
@@ -1,0 +1,86 @@
+### Allocation: the typed-shape layout is a property of the shape, so it is now a constant in the header
+
+Four allocation benchmarks sat inside a **6% band** at 2.50–2.65× node — object literals
+(`churn`, `churn_alloc`), class instances (`push_cls`), cyclic graphs (`cycles`). Four
+structurally different shapes do not land in a 6% band by coincidence; one shared
+per-allocation cost does.
+
+Symbolicated profiles of the 200M-allocation variants (two samples per program, agreeing
+within 1.5 pp) found it. `js_gc_declare_typed_shape_layout` was **30% of `churn_alloc` and
+`push_cls`**, and it spent that re-deriving *per object* a fact that is a property of the
+*shape*. #7510's memo had already collapsed the map round-trip to a direct-mapped probe;
+what remained was the probe, a type-table lookup, a field-count compare, and the cross-crate
+call itself. GC pause time, by contrast, is **3.5%** — the cost is that Perry performed nine
+out-of-line operations per allocation where V8 performs a bump-pointer and a write barrier.
+
+For a shape whose pointer mask is **statically empty**, the canonical layout is the constant
+`GC_LAYOUT_POINTER_FREE | GC_OBJ_TYPED_LAYOUT_INTACT`. The inline-bump `new` path already
+emits a packed `GcHeader` constant carrying the state half, so the intact bit is folded into
+that same store and the call is not emitted. What survives is the one half that depends on
+the recycled **address** rather than the shape — clearing whatever per-object record a
+previous tenant left — as a one-argument `js_gc_forget_object_layout` behind an inline
+`PERRY_PER_OBJECT_LAYOUTS_ANY` test. That global is a process-wide mirror of the per-thread
+emptiness flag, maintained by an armed-thread count; its `0` state proves every thread's
+per-object tables empty, and it is now also the first test inside `layout_forget_object`
+itself, replacing a Darwin `_tlv_get_addr` call with a static load on the disarmed path for
+every caller including object death.
+
+Two smaller levers in the same band:
+
+- **`js_ctor_return_override`** was called on every construction to answer a question that is
+  `undefined` for every constructor without an explicit `return` — 8% of `churn_alloc`, where
+  the synthesized object-literal constructor's only `ret` is the `TAG_UNDEFINED` constant.
+  `JSValue::is_undefined` is `bits == TAG_UNDEFINED`, so one 64-bit compare decides it inline.
+  The runtime call stays on the cold arm, where derived-constructor `TypeError`s, object
+  returns, arguments objects and arrays still need it.
+- **A `new` in a hot-loop callee is a `new` in a loop**, one frame out, so it now takes the
+  inline bump as well. `cycles.ts`'s `makeCycle` is the shape that needed this: 5 statements,
+  therefore `alwaysinline`, therefore *never* `inlinehint` — so the site gate was reading the
+  one flag that could not be set for the hottest function in the program. The signal is
+  `collect_hot_loop_callees` directly (≥1 in-loop call site AND ≤4 module-wide call sites),
+  which is the same anti-bloat bound the loop arm already accepts.
+
+Measured on the quiet M1 mini, best-of-5, with exit code 0 and byte-identical output verified
+for all 27 corpus programs before timing:
+
+| bench | before | after |
+|---|--:|--:|
+| `churn` | 0.4217 | **0.2900** (−31%) |
+| `churn_alloc` | 0.3720 | **0.2409** (−35%) |
+| `push_cls` | 0.3665 | **0.2368** (−35%) |
+
+`churn_alloc` goes **18.6 → 12.0 ns per allocation**; node is 7.1 ns on the same shape.
+Nothing else in the 19-benchmark corpus moves outside noise.
+
+`gc-handoff/bench/alloc_declare_pf.ts` and `alloc_declare_ptr.ts` isolate the cause with a
+control: the same program, same allocation count, same runtime stores, differing only in
+whether the second field's *declared type* makes the pointer mask non-empty. Before, both
+arms pay the declare and their times match (0.9044 / 0.8647); after, only the control does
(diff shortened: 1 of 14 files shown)


proggeramlug:
### Measurement status — one of the three changes is IR-verified but not yet timed

To be precise about what is and is not measured, because the three changes here are independent:

- **(a) the header bake** and **(b) the inline `undefined` return-override** are what produced every number in the description. Measured cleanly: `churn` 0.4217 → 0.2900, `churn_alloc` 0.3720 → 0.2409, `push_cls` 0.3665 → 0.2368, best-of-5 on a quiet mini, exit code + byte-identical output gated first.
- **(c) "a `new` in a hot-loop callee takes the inline bump"** is verified in the emitted IR only — `js_object_alloc_class_inline_keys` is gone from `cycles.ts`'s `makeCycle`, replaced by the inline bump — and all 27 corpus programs built from it still produce exit 0 and byte-identical output. **Its wall-clock effect is not yet measured.** The bench mini has been continuously occupied for several hours (five different agents in turn; at one point two were running benchmarks *while* the lock was held, load 7.96), and I would rather say this than record a number taken under load 4.

The `cycles` figure in the description (0.1885) is therefore from the **(a)+(b)** arm, without (c). Expected effect of (c) on `cycles` is ≈0.147 from its profile share (`js_object_alloc_class_inline_keys` is 22% of its samples, `arena_alloc` a further 5%), but that is an estimate, not a measurement.

A quiet-window retry is queued. **If (c) does not reproduce, it can be dropped on its own** — it touches only `new_site_is_in_loop` plus the `hot_loop_callee` field that feeds it, and (a)+(b) stand without it.

Full working notes, including the two next levers with measured sizes and one hypothesis I refuted, are in `gc-handoff/ALLOC-NOTES.md`.

Issue #655, opened 2026-05-10
AOT: Array methods are undefined when accessed via Map.get(...).field


proggeramlug:
## Summary

Under `perry compile`, calling `Array.prototype` methods on an array reached through `Map.get(k)!.field` returns garbage (or never dispatches) for `shift`, `splice`, `pop`, `slice`, `unshift`. `push` is special-cased and works. `length =` writes are silently ignored.

The same array, aliased to a single-identifier local first, dispatches the same methods correctly. So the issue is specifically the chained `Map.get(...).field.method()` call site, not `Array.prototype` itself.

Found while running `@perryts/redis` under AOT — drivers and pools that store per-connection FIFO state in module-level `Map<id, State>` (the pattern recommended by your own AOT guidance for closure-capture safety) hit this on every reply.

## Environment

- perry 0.5.772
- macOS 26.4 / Apple M1 Max

## Minimal repro

```ts
// repro.ts
interface State { pending: number[]; }

const m = new Map<number, State>();
m.set(1, { pending: [10, 20, 30] });
const s = m.get(1)!;

console.log('typeof shift:  ', typeof s.pending.shift);   // undefined
console.log('typeof splice: ', typeof s.pending.splice);  // undefined
console.log('typeof pop:    ', typeof s.pending.pop);     // undefined
console.log('typeof slice:  ', typeof s.pending.slice);   // undefined

const x = s.pending.shift();
console.log('shift ret:', x, 'len after:', s.pending.length);
// Actual:   shift ret: [object Object] len after: 3   ← length unchanged, return is wrong
// Expected: shift ret: 10                len after: 2

s.pending.length = 0;
console.log('after length=0:', s.pending.length);
// Actual:   3   ← write ignored
// Expected: 0

// Workaround: alias to local first
const arr = s.pending;
console.log('aliased typeof shift:', typeof arr.shift);  // still 'undefined'
console.log('aliased shift ret:', arr.shift());          // 10  ← actually works
console.log('back at struct:', s.pending);               // [ 20, 30 ]  ← mutation visible
```

```bash
$ perry compile repro.ts -o repro && ./repro
typeof shift:   undefined
typeof splice:  undefined
typeof pop:     undefined
typeof slice:   undefined
shift ret: [object Object] len after: 3
after length=0: 3
aliased typeof shift: undefined
aliased shift ret: 10
back at struct: [ 20, 30 ]
```

## Behavior matrix

| Call form | Works? |
|---|---|
[...]

PR #7835, opened 2026-08-11
perf(codegen,runtime): let a declared `string` pick the concat lowering

proggeramlug:
## The finding

`gc-handoff/apps/shapes.ts`'s `describe()` is `"shape:" + this.tag` — a string literal plus a field **declared `string`**. It lowered to `js_dynamic_string_or_number_add`: a `RuntimeHandleScope`, four `root_nanbox_f64`s and two `ToPrimitive` calls, spent rediscovering at runtime what the declaration already stated.

`is_string_expr` has trusted that same declaration for string *method* dispatch since #655. The concat path did not.

**88.8 ns per concatenation → 29.4 ns.** (`gc-handoff/bench/concat_field.ts` minus `concat_field_base.ts`, 2,000,000 concatenations, quiet M1 mini, best-of-7.)

## Why widening the predicate is safe, and why it is scoped so tightly

Perry does not enforce annotations at runtime — the exact gap #7831 is closing on the numeric side, where a declared-`number` field holding a string made `fadd` propagate the string's NaN payload and `typeof (v * 2)` answer `"string"`. A declaration is evidence, not proof.

So this PR does **not** widen `is_definitely_string_expr`. It adds a separate `is_declared_string_expr` with exactly one consumer: the two-operand `a + b` concat, which emits `js_string_concat_box`.

And it first fixes that helper. `js_string_concat_box` decoded operands with `str_bytes_from_jsvalue(...).unwrap_or((null, 0))` — an operand that was not a string became the **empty string**, so `"ab" + 42` through this helper rendered as `"ab"`. **This is reachable on `1ee158d27` today — tracked as #7837.** `is_definitely_string_expr`'s `LocalGet` arm already trusts a declared type, so a `string`-declared local holding a non-string selects this helper and the operand disappears:

```ts
const t: string = (99 as any);
console.log(t + "x");   // node: "99x"   perry before: "x"
```

Measured on a clean `1ee158d27` build and on this branch. A `string`-declared *field* (`o.t + 7`) and a `(string, number)` parameter pair both route elsewhere and were already correct on both builds, which is why the defect survives casual probing — three negative probes of the wrong shapes do not establish absence.

It is also why the concat fast path could never be selected from an annotation in the first place: `lower_string_concat.rs`'s self-append lowering carries a whole `dother`/`cold` arm whose comment says a lie has to be routed *around* this helper. It now forwards any non-string pair to `js_dynamic_string_or_number_add`.

### Scope against #7837 — this PR fixes one of the two defects

#7837 records two silent wrong answers from the same premise (an erased TS annotation treated as a runtime proof — the string mirror of #7831). **Do not read this PR as closing both.**

| # | shape | node | before | this PR |
|---|---|---|--:|--:|
| 2 | `t + "x"`, operand **dropped** | `99x` | `x` | **`99x` — fixed here** |
| 1 | `s + 7`, wrong **operator** chosen | `49` | `427` | `427` — **not fixed here** |

Defect 1 lives in the one-sided `l ^ r` arm, which lowers through `js_string_concat_value` / `js_value_concat_string`. Those take an already-unboxed `StringHeader*` and cannot detect the lie, so fixing it needs the guarded-diamond treatment #7831 is applying to the numeric side, not this PR's runtime delegation. That arm is deliberately left on the strict predicate here and is untouched; the decision belongs to #7837.

Worth noting for that decision: scoping `is_declared_string_expr` to the two-operand path is what keeps this PR from *widening* defect 1. Had the declaration been accepted in the one-sided arm, `o.t + 7` — correct at `14` on both builds today — would have become `"77"`.

After that, `js_string_concat_box` returns the dynamic-path answer for every combination of runtime values — string+string, string+number, and number+number (which **adds and returns a number**). The declaration therefore selects a *lowering* and can never select an *answer*.

Three neighbours deliberately keep the strict predicate, each because it *would* be able to change an answer:

- the one-sided `l ^ r` arm lowers through `js_string_concat_value` / `js_value_concat_string`, which take an already-unboxed `StringHeader*` and cannot tell a lie from a string;
- the N-way chain fold formats every part as a string, so an all-declared chain of numbers would concatenate where the spec adds;
- the `Map` string-key fast paths in `expr::math_simple` key a lookup on the claim.

## `type X = { … }` was worth less than `interface X { … }`

`lower_type_alias_decl` files aliases in `module.type_aliases`; `lower_interface_decl` files interfaces in `module.interfaces`. `static_type_of` only ever consulted the latter. So `type Record = { kind: string; amount: number }` proved **nothing** about `r.kind` — not because an alias is weaker evidence than an interface (structurally they are the same declaration, the same runtime layout, the same absence of any layout guarantee), but because of which cabinet it was filed in. Only a non-generic alias whose right-hand side is a closed object type answers.

This is what lets `pipeline.ts`'s `makeTagger` (`prefix + r.kind`, 360,000 calls) reach the concat at all.

## A `static` counter re-interned its own key, once per construction

Codegen emits `js_class_register_static_field` after every `Expr::StaticFieldSet`, so `Shape.made = Shape.made + 1` inside a constructor runs it **once per construction** — 144,000 times in `shapes.ts`. Each call did `str::from_utf8(…).to_string()` (a heap allocation, immediately dropped, because `HashMap::insert` keeps the *original* key), a `CLASS_DELETED_KEYS` probe, and an `entry().or_insert_with().insert()`.

The signature is now `&str` — every caller already had a borrowed value — and a store whose key exists updates the slot in place. The in-place arm skips the deleted-keys probe only while **nothing has ever been deleted**; once anything has, the original sequence runs verbatim, so the pre-existing conflation between a deleted prototype key and a same-named static field (`class C { m() {} static m = 1 }`, both under one class_id) keeps whatever behaviour it had.

## Measurements

Quiet M1 mini, load ~1.6, best-of-7 (best-of-5 for the wider set), against a clean build of `1ee158d27` built with the same `-p perry -p perry-runtime-static -p perry-stdlib-static` package set and an isolated `CARGO_TARGET_DIR`. Every program's output verified byte-identical to `node --experimental-strip-types` (26.5.1) with exit 0 **before** timing; the harness records an exit code per cell.

| bench | before | after | |
|---|--:|--:|--:|
| `concat_field` (new probe) | 0.2742 | **0.1553** | −43% |
| `concat_field_base` (its subtrahend) | 0.0967 | 0.0965 | — |
| **ns / concatenation** | **88.8** | **29.4** | **−67%** |
| `apps/pipeline.ts` | 0.5164 | **0.4847** | −6.1% |
| `apps/shapes.ts` | 0.1894 | **0.1833** | −3.2% |

No corpus regression (best-of-5): `churn` 0.4216→0.4234, `churn_alloc` 0.3729→0.3752, `churn_read` 0.0228→0.0224, `push_num` 0.1507→0.1511, `push_cls` 0.3658→0.3681, `cycles` 0.1938→0.1938, `deeplist` 0.1239→0.1232, `tree` 1.6419→1.6425, `tree_wide` 2.1132→2.1097, `retain` 0.3477→0.3464, `retain1` 0.1367→0.1359, `retain_wide` 0.4552→0.4569, `fib40` 0.3997→0.3948, `asyncpipe` 0.1345→0.1349, `interp` 1.4976→1.4971, `iso_miss` 1.9256→1.9260.

The `iso_miss` canary prints `checksum 437840 misses 0` and exits 0 on the miss **counter**, including under `PERRY_GC_SCHEDULE_RATE=1`, `PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800`, and `PERRY_GC_VERIFY_EVACUATION=1`. `shapes`, `pipeline` and `concat_field` are also byte-correct under `PERRY_GC_SCHEDULE_RATE=1 PERRY_GC_VERIFY_EVACUATION=1`.

## Tests

All in-crate unit tests, so they run in the per-PR `cargo-test` job rather than the nightly-only integration suites (#5960).

- `perry-runtime`: `concat_box_delegates_a_non_string_operand_to_the_dynamic_add` pins all four operand combinations (including number+number returning a **number**) plus the SSO result encoding; `repeated_store_updates_in_place_and_stays_readable` and `store_after_delete_clears_the_deleted_mark` pin both arms of the static-field store.
- `perry-codegen`: `alias_declared_string_field_takes_the_static_concat`, `field_absent_from_the_alias_keeps_the_dynamic_add`, and `alias_declared_number_field_is_deliberately_not_routed` — the last asserts the numeric side is *unchanged*, so widening it later has to be a deliberate edit to this test.

Local runs on this branch's build: `perry-runtime --lib` 2108 passed / 0 failed (`RUST_TEST_THREADS=1`), `perry-codegen --lib` 858 passed / 0 failed after the test correction. Gap suite in progress locally at time of writing; through 290/535 on the pre-narrowing build the only failures were the 8 already recorded in `test-parity/gap_snapshot.json` (2 `parity_fail`, 6 `node_fail`) — zero new.

## Probes added

`gc-handoff/bench/concat_field.ts` / `concat_field_base.ts` (the ns/concat pair) live in the out-of-repo handoff tree, not in this PR.

https://claude.ai/code/session_012B8z92S82sCfqCrVqrFgS2

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved string concatenation when values declared as strings contain numbers, `undefined`, or other non-string values.
  - Corrected type handling for properties defined through object-type aliases.
  - Fixed static property updates so existing values are replaced reliably and deleted properties can be restored correctly.
  - Preserved proper numeric addition and string conversion behavior across mixed-value operations.

- **Tests**
  - Added regression coverage for mixed-value concatenation, property updates, enumeration, deletion, and restoration scenarios.


proggeramlug:
Heads-up from #7842 (closing #7837): **this PR's `js_string_concat_box` hunk is also in mine, verbatim.**

I did not intend to duplicate it — #7837's brief told me it was yours and not to redo it — but the acceptance criterion for #7842 is that `gc-handoff/m0810/lyinglocal.ts` prints `49` **and** `99x`, and the second row is exactly the operand-dropping defect this hunk fixes. There is no codegen-side fix for it that isn't a guard on a case that is already operator-correct (`t + "x"` really is concat, the helper just discarded `t`), so #7842 cannot stand on its own without it. Rather than stack on a draft branch that may be reworked, I took the hunk with attribution in both the doc comment and the changelog fragment.

**Whichever of us lands second should drop their copy** — it is a ~20-line `let-else`, so the conflict is mechanical either way.

One correction to this PR's framing, now that #7837 has been fully enumerated. The doc comment on `is_declared_string_expr` says `is_definitely_string_expr` "answers only for expressions whose string-ness is structural (a literal, `String(x)`, `.toString()`, `JSON.stringify`, …)". Two of those are not structural:

- its **`LocalGet`** arm is a declared type, which is the whole of #7837 defect 1;
- the **`.toString()` / `.slice()` / `.replace()` …** arm matches on the METHOD NAME with no look at the receiver, so `arr.slice(0) + 7` claimed a string and printed `""` on `82f0e9681`. `Array.prototype.slice` returns an array.

#7842 adds `string_value_is_runtime_guaranteed` to draw that line explicitly (closed whitelist — an unclassified arm is treated as a claim and gets guarded), so if you keep `is_declared_string_expr` it may be worth defining it against that rather than against `is_definitely_string_expr`.

No conflict otherwise: your `alias_declared_number_field_is_deliberately_not_routed` test stays green under #7842, because `"t:"` is a literal and therefore a proof — that site never reaches my guard. And the concat fast path you are widening keeps its fused single-allocation lowering wherever any operand is proven, which is the `"shape:" + this.tag` and `prefix + r.kind` case this PR is about.

PR #7836, opened 2026-08-11
fix(runtime): root async trap state across step bodies

### LINKED ISSUE #7794: GC: async_step_fulfill_thunk holds two bare *mut Promise locals across the step-body call (asyncpipe SIGBUS)
## Summary

`async_step_fulfill_thunk` keeps **two bare `*mut Promise` Rust locals alive across the step-body call**, which is arbitrary user code that allocates and can trigger an evacuating minor. The GC rewrites every slot it can see — the closure capture, and `INLINE_TRAP.trap_next` via `scan_inline_trap_step` — but it cannot see a bare local. After the collection both locals name **retired from-space**, and one of them is immediately dereferenced.

This is the unexplained `asyncpipe` SIGBUS: exit 138 under `PERRY_GC_PROTECT_FROMSPACE=1`, faulting on a `GC_TYPE_PROMISE` (`obj_type=5 size=72`).

It is the same family as #7154 / #7184 / #7192, and specifically the same *residual* shape that #7497 already fixed one function away in `js_async_step_done` ("the displaced value rides a bare Rust local ... and is republished by `_end`. Rooting the cell protects the ARMED value, not the saved one").

## Location

`crates/perry-runtime/src/promise/async_step.rs`, `async_step_fulfill_thunk` (~line 913):

```rust
let captured_trap_next = js_closure_get_capture_ptr(closure, 1) as *mut Promise;  // (A) bare local
...
let prev = INLINE_TRAP.with(|c| { let old = c.get();                              // (B) bare local
    c.set(InlineTrap { trap_next: captured_trap_next, current_step: step as usize }); old });

let result = js_closure_call2(step, value, false_bits);   // <-- USER CODE: allocates, can collect

INLINE_TRAP.with(|c| c.set(prev));                        // (B) republishes a STALE pointer
forward_swallowed_rejection(result, captured_trap_next);  // (A) STALE pointer, then dereferenced
```

`forward_swallowed_rejection` (~line 963) dereferences it directly — `(*trap_next).state` — and that is the faulting instruction the quarantine reporter names.

Two distinct defects in one function:

* **(A)** `captured_trap_next` is read *before* the step body runs and used *after* it.
* **(B)** `prev` (an `InlineTrap`, which contains a raw `*mut Promise`) is saved before and **written back into the TLS cell** after. So the collector faithfully rewrites `INLINE_TRAP` during the collection, and this line then overwrites it with the pre-collection value. Rooting the cell cannot help; the stale value is reintroduced from outside it.

Note that (B) means `scan_inline_trap_step` working correctly is *not* evidence against this bug — which is why the obvious explanation looked wrong.

## Reproducer

```bash
# gc-handoff/apps/asyncpipe.ts, any build
PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800 ./p_asyncpipe   # exit 138
```

Symbolicated (`PERRY_DEBUG_SYMBOLS=1`):

```
[gc-fromspace-protect] FAULT: signal 10 at 0x29f690206c0
  This address is RETIRED FROM-SPACE.
  block=... retired_by_minor=#1657
  last-known object: user_ptr=0x29f690206c8 obj_type=5 size=72
0  ...arena::quarantine::fromspace_fault_handler + 5024
1  libsystem_platform.dylib  _sigtramp + 56
2  ...promise::async_step::async_step_fulfill_thunk + 176
3  ...promise::microtasks::run_microtasks + 20212
4  js_callback_timer_tick + 2996
5  ...promise::microtasks::run_microtasks + 26400
6  main + 536
```

## Interaction with #7793 — please read before triaging as fixed

PR #7793 (the property-miss probe gates) removes ~4 key-string allocations per thenable check on the async path. That drastically reduces the number of collection opportunities inside the vulnerable window, and **`asyncpipe` stops faulting under default GC pacing after it**. That is masking, not a fix — the bug is fully intact:

```bash
# WITH #7793 applied, same program:
PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800 ./p_asyncpipe
# -> exit 138, identical fault: obj_type=5 size=72, retired_by_minor=#1657
```

So after #7793 lands, the default-pacing reproducer is gone and `PERRY_GC_ZEAL=1` becomes required to see it. Worth fixing before that knowledge goes stale.

## Suggested fix

Mirror #7497's pattern from `js_async_step_done` in the same file: open a `RuntimeHandleScope`, root the captured promise (and the saved `prev.trap_next`) before `js_closure_call2`, and **re-read both from their handles afterwards** rather than using the pre-call bindings — including when restoring `INLINE_TRAP`.

## Why the static checker did not catch it

`scripts/gc_root_dominance_check.py` reads emitted LLVM IR from user modules; this is a raw pointer held in a *runtime* Rust frame, which is structurally invisible to it (see the CLAUDE.md note on runtime-side caches). The runtime instruments are the only detector here.

proggeramlug:
Closes #7794

## Summary

- Keep the activation result promise, the saved outer trap promise, and the saved outer step closure in mutable runtime handles across arbitrary async step-body calls.
- Re-read all three addresses after the call before restoring INLINE_TRAP or forwarding a rejection.
- Use the same rooted call path for fulfillment, rejection, first-call, and async-generator resumptions.
- Add a deterministic relocation regression that failed on main because rejection forwarding targeted the pre-move promise.

## Validation

- cargo test -p perry-runtime --lib -- --test-threads=1: 2106 passed, 0 failed, 4 ignored.
- Focused relocation regression: failed before the production change, passes after it.
- Exact parity: test_gap_async_advanced, test_gap_async_throw_after_await_not_unhandled, test_gap_generators, and test_gap_generator_return_throw_finally all pass.
- Forced moving collection plus relocation verification on a resumed async step: exit 0, answer=42, 6025 objects copied.
- cargo fmt --all -- --check, git diff --check, and scripts/check_file_size.sh pass.
- scripts/raw_handle_debt.py is already red on clean origin/main with the same 1013 versus 998 count and the same four unrelated module violations; this change adds none.

## Performance

Quiet-mini timing is pending because /Users/perry/bench.lock is owned by triage-agent. This PR remains draft until the async hot-path cost is measured.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved async function and generator resumption when garbage collection moves promise objects.
  * Prevented stale references during suspended-step continuation and rejection handling.
  * Preserved ambient trap state correctly across async step execution.
  * Improved runtime-handle updates when objects are relocated during garbage collection.

* **Tests**
  * Added coverage for relocated captured promises, prior trap promises, async step closures, and forwarded runtime handles.

### DIFF
--- a/changelog.d/7836-async-step-roots.md
+++ b/changelog.d/7836-async-step-roots.md
@@ -1,0 +1,4 @@
+**Async functions no longer retain pre-collection Promise addresses while a
+suspended step resumes.** Async step calls keep activation and ambient trap
+pointers in mutable roots, preventing a moving collection inside user code from
+restoring or dereferencing relocated objects.
--- a/changelog.d/7836-async-step-roots.md
+++ b/changelog.d/7836-async-step-roots.md
@@ -1,4 +1,4 @@
 **Async functions no longer retain pre-collection Promise addresses while a
 suspended step resumes.** Async step calls keep activation and ambient trap
 pointers in mutable roots, preventing a moving collection inside user code from
-restoring or dereferencing relocated objects.
+using stale pre-collection addresses after the step returns.
--- a/crates/perry-runtime/src/gc/mod.rs
+++ b/crates/perry-runtime/src/gc/mod.rs
@@ -59,6 +59,8 @@ pub(crate) use hot_tls::*;
 mod roots;
 pub use roots::*;
 #[cfg(test)]
+/// Rewrite runtime-handle roots only; this deliberately does not rewrite the
+/// installed `INLINE_TRAP`, whose scanner is exercised separately.
 pub(crate) fn test_rewrite_runtime_handles_for_forwarded_objects() {
     let valid_ptrs = build_valid_pointer_set();
     let mut visitor = RuntimeRootVisitor::for_rewrite(&valid_ptrs);
--- a/crates/perry-runtime/src/gc/mod.rs
+++ b/crates/perry-runtime/src/gc/mod.rs
@@ -58,6 +58,12 @@ mod hot_tls;
 pub(crate) use hot_tls::*;
 mod roots;
 pub use roots::*;
+#[cfg(test)]
+pub(crate) fn test_rewrite_runtime_handles_for_forwarded_objects() {
+    let valid_ptrs = build_valid_pointer_set();
+    let mut visitor = RuntimeRootVisitor::for_rewrite(&valid_ptrs);
+    scan_runtime_handle_roots_mut(&mut visitor);
+}
 /// #7148: the census of conservative-scan fallbacks and the precise-safepoint
 /// drains that replace them. Declared next to `roots` because
 /// `ManualGcScanGuard` is what records into it.
--- a/crates/perry-runtime/src/promise/async_step.rs
+++ b/crates/perry-runtime/src/promise/async_step.rs
@@ -787,23 +787,12 @@ pub extern "C" fn js_async_first_call(step_closure_nanbox: f64) -> f64 {
     // `trap_next` untouched. The outer's chain reuse on its OWN
     // resumption is unaffected (this restore at function exit puts
     // `prev` back).
-    let prev = INLINE_TRAP.with(|c| {
-        let old = c.get();
-        c.set(InlineTrap {
-            trap_next: std::ptr::null_mut(),
-            current_step: ptr as usize,
-        });
-        old
-    });
-    let result = {
-        crate::closure::js_closure_call2(
-            ptr,
-            f64::from_bits(0x7FFC_0000_0000_0001), // TAG_UNDEFINED
-            f64::from_bits(0x7FFC_0000_0000_0003), // TAG_FALSE
-        )
-    };
-    INLINE_TRAP.with(|c| c.set(prev));
-    result
+    call_async_step_body(
+        ptr,
+        f64::from_bits(0x7FFC_0000_0000_0001), // TAG_UNDEFINED
+        f64::from_bits(0x7FFC_0000_0000_0003), // TAG_FALSE
+        std::ptr::null_mut(),
+    )
 }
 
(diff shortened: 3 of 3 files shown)


proggeramlug:
Quiet-M1 performance check completed against base 82f0e9681 and PR head a0f158e86.

Methodology:
- Built perry, perry-runtime-static, and perry-stdlib-static for both arms with isolated staged source and a target directory outside the repo.
- Compiled both arms with the same saved base compiler, changing only PERRY_RUNTIME_DIR.
- Confirmed the base/fix runtime archive hashes differed.
- Ran one warmup plus 12 alternating A/B pairs per workload; every run exited 0 with identical output.

Results (median wall time):
- asyncpipe: 3.435084 s -> 3.443230 s, +0.237% (median paired ratio +0.223%).
- promise_all_chains: 0.195893 s -> 0.197088 s, +0.610% (median paired ratio +0.641%).

The async-heavy regression is below 1% in both subjects. Marking ready for review.

proggeramlug:
Addressed the review findings in 32b803428:

- The relocation fixture now snapshots and restores each forwarded source object's first payload word and explicitly clears GC_FLAG_FORWARDED during teardown.
- The test asserts that cleanup happened before returning.
- Documented the test-only side-table scanner exemption, the runtime-handle-only helper coverage, and why the step pointer needs no post-call re-read.
- Reworded the changelog entry in terms of avoiding stale pre-collection addresses.

Validation:
- Focused relocation test: passed.
- Full single-threaded perry-runtime library suite: 2,106 passed, 0 failed, 4 ignored.
- cargo fmt --check, git diff --check, and scripts/check_file_size.sh: passed.
- raw_handle_debt.py remains at the pre-existing 1,013 vs 998 baseline with the same four unrelated module violations; no increase from this patch.

Issue #6949, opened 2026-07-29
runtime: a fourth unrooted-operand family — js_string_coerce as a plain ToString argument coercion (string built-ins, constructors)


proggeramlug:
Found while sweeping #6943 (the `js_string_coerce`-as-property-key family). #6943 fixed every site
where `js_string_coerce` is being used as the **property-key** coercion. The sweep read all 102
`js_string_coerce` callsites in `perry-runtime` and turned up a second, structurally identical set
where the coercion is a plain **`ToString` argument coercion** — not a key at all — and a raw
`StringHeader` / `ObjectHeader` operand still spans it.

The mechanism is the same and needs no re-arguing: `js_string_coerce` returns early with no
allocation only for an already-heap `STRING_TAG` value (`builtins::string_coerce_is_inert`, added in
#6943). Every other shape allocates — SSO short strings materialize onto the heap, numbers/bools/
`null`/BigInt build their stringification, and a `POINTER_TAG` object runs a user `toString` /
`valueOf` — and an allocation can trigger a GC that **evacuates**. A Rust local is neither a GC root
nor a shadow slot.

Two sub-shapes, both on `main` @ 83a6767ff.

### (a) String built-ins: the receiver string spans the argument coercion

```rust
pub extern "C" fn js_string_replace_string_dyn(
    s: *const StringHeader,          // receiver — raw fn param
    pattern: *const StringHeader,    // also raw
    replacement: f64,
) -> *mut StringHeader {
    ...
    js_string_replace_string(s, pattern, crate::builtins::js_string_coerce(replacement))
    //                       ^^^^^^^^^^  stale after the coercion evacuates them
}
```

Rust evaluates arguments left to right, so `s` / `pattern` are *copied* before the coercion runs —
but the pointees move, and the callee dereferences the stale copies (`string_as_str(s)` in
`replace_string.rs`).

- `regex/replace_fn.rs:359` `js_string_replace_string_dyn` — `s`, `pattern`
- `regex/replace_fn.rs:371` `js_string_replace_all_string_dyn` — `s`, `pattern`
- `regex/replace_fn.rs:413` `js_string_replace_search_dyn` — `s`
- `regex/replace_fn.rs:427` `js_string_replace_all_search_dyn` — `s`
- `regex/replace_fn.rs:444` `js_string_replace_regex_dyn` — `s`, `re`
- `regex/replace_fn.rs:461` `js_string_replace_all_regex_dyn` — `s`, `re`
[...]

PR #7840, opened 2026-08-11
fix(runtime): convert #7811/#7815's new bare raw-handle reads to across_* (#7838)

### LINKED ISSUE #7838: raw-handle ratchet red on main: #7811/#7815 landed 15 bare reads the same batch's #7825 now counts
Found by the post-landing audit of the 2026-08-11 batch. The `Raw-handle debt ratchet` lint step fails on current `main` (82f0e9681): **1,013 bare reads vs baseline 998**, four per-module violations. `lint` will go red on main's next CI run.

The +15 all come from the two #6949 rooting fixes that merged in the batch:

| module | new bare reads | from |
|---|--:|---|
| `regex/replace_fn.rs` | 13 (ceiling 3) | #7811 |
| `disposable.rs` | 3 (no ceiling) | #7815 |
| `boxed_primitives.rs` | 1 (no ceiling) | #7815 |
| `messaging.rs` | 1 (no ceiling) | #7815 |

**The code is not unsound** — both PRs root their subjects in `RuntimeHandleScope` and re-derive raw pointers from the rooted handles after every allocating call; #7811's fragment records a clean run under the seeded schedule + quarantine. The debt is stylistic: the sites use manual `get_raw_*_ptr` re-reads where the ratchet's discipline wants `RuntimeHandle::across_{mut,const,nanbox}`. The ratchet only sees this now because #7825 (same batch) closed the hole that let a PR's own checkout carry the comparison baseline.

## The fix, per `scripts/raw_handle_debt_files.txt`'s own header

1. **#7815's five sites convert mechanically.** Each is the exact shape `across_*` exists for: allocate receiver → one allocating coercion → writes through the re-read pointer. Convert all five; those modules stay unlisted (new code clean by construction).
2. **#7811's thirteen fit the header's ONE sanctioned join-the-list shape** — a loop whose collection window is a user-visible callback (the replacer fn), with `cur_str = || string_as_str(handle.get_raw_const_ptr(...))` re-deriving at every access across interleaved callback calls. `across_*` pairs one call with one re-read and cannot express this. Raise `replace_fn.rs`'s ceiling 3 → 13 **naming the loop and the trap**, per the header.
3. The header's condition for (2) is that **the global baseline must not rise**. After (1) converts 5, the total sits at 1,008 vs the 998 baseline — so either convert ~10 more pairs elsewhere in the same PR (the header's "honest move"), or the maintainer explicitly re-records the baseline at the new floor in the same commit that documents (2). The first is preferred; the second is at least a decision on the record rather than a silent raise.

Until this lands, every PR runs `lint` against a red step on main — the exact "required + red ⇒ every merge bypasses" gate-theatre shape CLAUDE.md warns about, so this is urgent for the ratchet's credibility, not just tidiness.

proggeramlug:
Closes #7838.

The raw-handle debt ratchet is red on `main`: **1,013 bare reads against a baseline of 998**, four per-module violations. The +15 arrived with the two #6949 rooting fixes in the 2026-08-11 batch (#7811, #7815), and #7825 — same batch — closed the hole that had been letting a PR's own checkout carry the comparison baseline, so the ratchet only started seeing them once it landed.

**All fifteen convert to `RuntimeHandle::across_{mut,const}`. No ceiling was raised and `raw_handle_debt_baseline.txt` is untouched** — the total lands back on 998 exactly, and `regex/replace_fn.rs` returns to its recorded ceiling of 3.

| module | before | after | disposition |
|---|--:|--:|---|
| `disposable.rs` | 3 | **0** | converted; module stays unlisted |
| `builtins/formatting/boxed_primitives.rs` | 1 | **0** | converted; module stays unlisted |
| `messaging.rs` | 1 | **0** | converted; module stays unlisted |
| `regex/replace_fn.rs` | 13 | **3** | 10 converted; back at its existing ceiling |

## Why `replace_fn.rs` did not need a ceiling raise

#7838 proposed raising it 3 → 13 on the grounds that the thirteen are the one shape `raw_handle_debt_files.txt` sanctions joining the list for — a loop whose collection window is a user-visible callback, where `cur_str = || string_as_str(handle.get_raw_const_ptr(..))` re-derives at every access and `across_*` (one call ↔ one re-read) cannot express it.

That is true of **three** of them, not thirteen — and those three are *pre-existing*, which is exactly what the ceiling of 3 was recorded for. `git show 6af7e5840^:crates/perry-runtime/src/regex/replace_fn.rs | grep -c get_raw_` is **3**; at `6af7e5840` it is **13**. So #7811 added **10**, and all ten are the plain `across_*` shape: root → one allocating `js_string_coerce` → read-for-the-call. They convert mechanically.

Four of the ten are two-receiver call sites (`s` + `pattern`, `s` + `re`). Two receivers compose by **nesting** `across_const` — the inner call runs the coercion and re-reads `b`, the outer then re-reads `a` — which is how `path::value_args::with_two_headers` already does it. Two small private combinators (`with_two_receivers_across`, `with_receiver_across`) carry them so the nesting is written once.

## Two conversions are not purely mechanical

Writing the ordering out made a latent stale-address use visible in `disposable.rs` (`js_suppressed_error_new`). Both are the #7192 shape — the store is in-frame but *after* a call that allocates — and neither is reachable without evacuation actually moving the receiver, so this is ordering hygiene rather than an observed crash:

- **`set_nonenum` filed attributes under a possibly-stale address.** It called `crate::object::set_property_attrs(obj as usize, ..)` *after* `js_object_set_field_by_name`, which allocates when the object grows. That side table is keyed on the address, so a pre-call copy does not fault — it files the attributes where nothing will look them up, and `error` / `suppressed` / `message` silently become **enumerable** on a `SuppressedError` that grew during the set. Now re-read across the field-set.
- **The returned NaN-box was built before the prototype lookup.** `js_nanbox_pointer(obj)` was computed, then `builtin_prototype_value("SuppressedError")` ran, then the box was returned. A NaN-box is a frozen address the collector cannot rewrite, so the returned value named from-space if that lookup collected. The box is now built **last**, after every allocating call.

## Validation

```
$ python3 scripts/raw_handle_debt.py ; echo $?
bare raw-handle reads: 998 (baseline 998)
per-module: 110 module(s) within ceilings; every other runtime module is locked at zero
0
```

(the script's own exit code, not a pipeline's). `--no-raise-vs origin/main`, the mode CI runs, also exits 0: `baseline 998 -> 998, 110 -> 110 module ceiling(s), none raised`. Release build of `-p perry -p perry-runtime-static -p perry-stdlib-static` is clean; the `perry-runtime` test suite is still running and I will post the result.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runtime stability when creating boxed strings, suppressed errors, and broadcast channels.
  * Fixed potential failures during regular-expression replacement when values require conversion or memory cleanup occurs.
  * Improved error handling and object safety during operations that may trigger memory management.
  * Preserved existing behavior for empty strings, symbol conversion errors, callable replacements, and regular-expression processing.

### DIFF
--- a/changelog.d/7840-raw-handle-debt-across.md
+++ b/changelog.d/7840-raw-handle-debt-across.md
@@ -1,0 +1,47 @@
+### Fixed
+
+- **The raw-handle debt ratchet is green again on `main` (#7838).** It went red at
+  1,013 bare `get_raw_{mut,const}_ptr` reads against a baseline of 998, with four
+  per-module violations. The +15 all arrived with the two #6949 rooting fixes in the
+  2026-08-11 batch (#7811, #7815); #7825 — same batch — closed the hole that had been
+  letting a PR's own checkout carry the comparison baseline, so the ratchet only
+  started seeing them once it landed. While a required gate is red on `main`, no agent
+  can tell its own breakage from inherited breakage, which is the condition a real
+  regression merges unnoticed in.
+
+  **All fifteen convert to `RuntimeHandle::across_{mut,const}`. No ceiling was raised
+  and `raw_handle_debt_baseline.txt` is untouched** — the total lands back on 998
+  exactly, and `regex/replace_fn.rs` returns to its recorded ceiling of 3.
+  `disposable.rs`, `messaging.rs` and `builtins/formatting/boxed_primitives.rs` reach
+  zero and stay unlisted.
+
+  **`replace_fn.rs` did not need the 3 → 13 raise #7838 proposed.** The proposal rested
+  on all thirteen being the one shape `raw_handle_debt_files.txt` sanctions joining the
+  list for — a loop whose collection window is a user-visible callback, where a
+  `cur_str` helper re-derives at every access and `across_*` (one call ↔ one re-read)
+  cannot express it. That describes **three** of them, and those three are
+  *pre-existing*: `git show 6af7e5840^:…/replace_fn.rs | grep -c get_raw_` is 3, which
+  is what the ceiling of 3 was recorded for. #7811 added **ten**, all of them plain
+  root → one allocating `js_string_coerce` → read-for-the-call. Four are two-receiver
+  sites; two receivers compose by **nesting** `across_const`, the same way
+  `path::value_args::with_two_headers` already does it, and two small private
+  combinators carry them.
+
+- **`SuppressedError` filed its property attributes under a possibly-stale address.**
+  `set_nonenum` called `object::set_property_attrs(obj as usize, …)` *after*
+  `js_object_set_field_by_name`, which allocates when the object grows. That side table
+  is keyed on the address, so a pre-call copy does not fault — it files the attributes
+  where nothing will look them up, and `error` / `suppressed` / `message` silently
+  become **enumerable** on a `SuppressedError` that grew during the set.
+
+- **`js_suppressed_error_new` returned a NaN-box built before its last allocating
+  call.** `js_nanbox_pointer(obj)` was computed, then the `SuppressedError.prototype`
+  lookup ran, then the box was returned. A NaN-box is a frozen address the collector
+  cannot rewrite, so the returned value named from-space if that lookup collected. The
+  box is now built last.
+
+  Both are the #7192 shape — the store is in-frame but *after* a call that allocates —
+  and neither is reachable without evacuation actually moving the receiver, so this is
+  ordering hygiene rather than an observed crash. They are recorded because writing the
+  `across_*` ordering out is what made them visible, which is the argument for the
+  discipline the ratchet enforces.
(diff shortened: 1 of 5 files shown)


proggeramlug:
Test result promised in the PR body, posting for the record now that this has merged:

```
$ RUST_TEST_THREADS=1 cargo test --release -p perry-runtime
test result: ok. 2108 passed; 0 failed; 4 ignored; 0 measured; 0 filtered out
test result: ok. 0 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out
```

Green, single-threaded per #7791.

Also verified the gate holds on `main` *after* this landed — #7831 and #7836 merged on top of it, and #7836 is itself a rooting fix, which is the shape that broke the ratchet last time:

```
origin/main total bare reads = 998   baseline = 998   -> OK
per-module violations: 0
```

Issue #7841, opened 2026-08-11
A declared `string` local silently mis-lowers `s += x`: `c += 1` on a slot holding 42 gives "421", not 43


proggeramlug:
## Summary

The same premise as #7837 — an erased TypeScript annotation treated as a runtime proof — has a second home, in a different lowering, and it survives #7837's fix.

```ts
let c: string = (42 as any);   // declared string, actually a number
c += 1;
console.log(c);                // node: 43     perry: "421"
```

Reproduced on `82f0e9681` and on the #7837 branch. No throw, no diagnostic, exit 0.

`s += x` on a `string`-declared local does not go through `binary::lower` at all. `expr/literals_vars.rs`'s `LocalSet` arm recognises `s = s + x` and routes it to `lower_string_concat::lower_string_self_append`, gated on nothing more than `ctx.local_types.get(id) == Some(HirType::String)` — which is exactly the declaration, and exactly what a lying local satisfies. That lowering then `js_get_string_pointer_unified`s the destination, which ToString-coerces whatever is in the slot, and appends.

This is the shape the defect is most likely to be met in, because `s += x` in a loop is the canonical string-builder idiom.

## Why it was left out of #7837

Not an oversight — a scope call, with two concrete reasons.

**It sits on the load-bearing O(n) path.** `js_string_append` mutates in place at refcount 1, which is what makes `let s = ""; for (…) s += part` amortised linear instead of quadratic. Every change here has to be measured against that, and #7837's change measured at exactly zero on the corpus precisely because it left this lowering alone (all 19 corpus programs' LLVM IR differed by two `declare` lines and nothing else).

**The fix has to move a tag test above an observable side effect.** In the non-string-rhs variant the sequence today is: load the destination slot, lower the rhs, `js_jsvalue_to_string(rhs)`, then tag-dispatch on the destination. The correct spec behaviour when the destination is *not* a string is `ToPrimitive(rhs)` — so the destination tag has to be tested *before* the rhs coercion, or a user `toString` fires twice. That is a restructure of a block whose comments record two prior rooting bugs (#6951, #7462), not a one-line guard.

## Suggested shape

The same policy #7837 applies: a static type may select a lowering, never an answer.

[...]

PR #7842, opened 2026-08-11
fix(codegen,runtime): a declared `string` is not a proof, so it may not pick the `+` operator

### LINKED ISSUE #7837: A lying `string`-declared local silently drops an operand: `t + "x"` returns "x", and `s + 7` concatenates instead of adding
## Summary

A `string`-declared local that actually holds a number produces **two different silent wrong answers**, one of which drops an operand entirely. TypeScript types are erased at runtime, so a declared type is not a proof — but codegen trusts it here.

```ts
const src: any = 42;
const s: string = src;          // declared string, actually a number
console.log(s + 7);             // node: 49      perry: 427
const t: string = (99 as any);
console.log(t + "x");           // node: "99x"   perry: "x"      <-- 99 DROPPED
```

```
node --experimental-strip-types lyinglocal.ts    perry 0.5.1463 (1ee158d27)
49                                                427
99x                                               x
```

Reproduced on current `main` with `PERRY_NO_AUTO_OPTIMIZE=1`. No throw, no diagnostic, exit 0. Committed as `gc-handoff/m0810/lyinglocal.ts`.

## Two separate defects

**1. Wrong operator.** `s + 7` where `s` holds the number `42` must be a numeric add (`49`). Perry lowers it as string concatenation because `s` is *declared* `string`, giving `"42" + "7"` = `"427"`.

**2. An operand is discarded.** `t + "x"` where `t` holds `99` must be `"99x"`. Perry returns `"x"`. The non-string operand is not coerced and not concatenated — it vanishes. This is `js_string_concat_box`'s non-string path returning the equivalent of an empty string rather than delegating to `js_dynamic_string_or_number_add`.

Defect 2 is the more dangerous of the two: a wrong *value* is at least still a value, whereas silently deleting data from a string is the shape that corrupts logs, keys, and serialized output without ever looking wrong.

## Why the usual reasoning misses it

`is_definitely_string_expr`'s `LocalGet` arm trusts a declared type. A `string`-declared *field* or a `(string, number)` parameter pair do **not** reach the buggy helper on `1ee158d27` — I checked all three shapes, and only the local reaches it:

```ts
const s: string = "ab"; const n: number = 42; console.log(s + n);  // "ab42"  correct
const o: any = { t: "x" };                    console.log(o.t + 7); // "x7"    correct
function f(a: string, b: number) { return a + b; } console.log(f("q", 9)); // "q9" correct
```

So a reviewer probing the obvious spellings concludes the surface is safe. It is the *lying local* that gets through.

## Relationship to #7831

This is the mirror image of #7831 ("a declared numeric type is not a proof that the value is a number"). That PR fixes the numeric direction — where a declared `number` holding a string makes `+` produce a bad add. This is the string direction, and both stem from the same premise: **an erased TypeScript annotation is a hint, not a runtime proof.** Worth fixing together, or at least deciding the policy once and applying it to both predicates.

PR #7835 already fixes defect 2 as a prerequisite for widening the concat fast path, and adds the delegation to `js_dynamic_string_or_number_add`. Defect 1 — choosing concat over numeric add on a lying declared local — is not addressed there and needs its own decision.

## Found by

Cross-checking a sub-agent's claim during the 2026-08-11 performance round. I first probed a declared field and a parameter pair, concluded the bug was latent, and published that. The agent pointed out `LocalGet` trusts declared types, which sent me back to test the local form — where it is live. Worth recording as a method note: three negative probes did not establish absence, because none of them exercised the arm that trusts the annotation.

proggeramlug:
Closes #7837.

## The bug

`is_definitely_string_expr` answered `true` on the strength of an erased TypeScript annotation, and `+` chose string concatenation from it. Perry does not enforce declared types at runtime — CLAUDE.md says so under Known Limitations — so `const s: string = (42 as any)` really does put a number in the slot.

I enumerated the arms of that predicate rather than probing spellings, and found **13 shapes that were silently wrong** on `82f0e9681`, exit 0, no diagnostic. **#7835 has since landed and fixes 4 of them.** Splitting the credit honestly, measured on purpose-built compilers for each of the three commits:

| shape | Node | `82f0e9681` | main `ab1bd464b` (with #7835) | this PR |
|---|---|---|---|---|
| `s + 7` | `49` | `427` | `427` | **`49`** |
| `7 + s` | `49` | `742` | `742` | **`49`** |
| `s + true` | `43` | `42true` | `42true` | **`43`** |
| `a + b + "x"` (N-way fold) | `141x` | `4299x` | `4299x` | **`141x`** |
| `a + b + a` | `183` | `429942` | `429942` | **`183`** |
| `const u = s; u + 7` | `49` | `427` | `427` | **`49`** |
| `(c ? s : "q") + 7` | `49` | `427` | `427` | **`49`** |
| `arr.slice(0) + 7` | `1,27` | `` (empty) | `` (empty) | **`1,27`** |
| `f(a: string, b: number)` via a function value | `49` | `427` | `427` | **`49`** |
| `t + "x"` | `99x` | `x` — operand vanished | `99x` | `99x` |
| `"x" + t` | `x99` | `x` | `x99` | `x99` |
| `a + b` (both declared) | `141` | `` (empty) | `141` | `141` |
| `pf(a: string)` returning `a + "x"` | `99x` | `x` | `99x` | `99x` |

The last four rows are #7835's — they route through `js_string_concat_box`, which it made total. **The nine above them are #7837 defect 1** and are still wrong on `main` today: the wrong OPERATOR, chosen from an annotation.

Two of the nine deserve a second look, because they show the premise is wider than "an annotation".

**The `.toString()` / `.slice()` / `.replace()` … arm matches on the METHOD NAME**, with no look at the receiver. `Array.prototype.slice` returns an array — so `arr.slice(0) + 7` claimed a string and the array was decoded as the empty string. A name is a guess about the receiver's type, which is evidence of exactly the same quality as an annotation.

**A `string` PARAMETER is live too.** #7837 records a `(string, number)` parameter pair as already-correct, and it is — when called *directly*, because the inliner substitutes the argument and the annotation evaporates. Reached through a function value it is not, and that is what the row above measures. Three negative probes did not establish absence; nor did four.

## The policy

The same one #7831 is applying on the numeric side: **a static type may select a lowering, never an answer.** It is applied in the place each site can afford it, which is different for each.

**Helpers that receive both operands NaN-boxed can be made total, and #7835 did that** — `js_string_concat_box` no longer decodes a non-string operand as the empty string. That hunk was in this branch before the rebase and has been **dropped**: main has it, and it is the right fix for those four rows. Nothing to merge there.

**The one-sided `l ^ r` arm cannot be repaired that way.** Codegen unboxes the string operand to a `StringHeader*` before the call, so by the time `js_string_concat_value` runs there is no tag left to test. When the operand's string-ness is declared-only it is now passed NaN-boxed to `js_string_add_value` / `js_value_add_string`, which test the tag and then either run the identical fused single-allocation concat or fall through to the spec's `+`.

**The N-way chain fold** formats every part as a string, so it reproduces the source tree only when the FIRST node really concatenates — once the accumulator is a string every later `+` concatenates whatever its part holds. The fold now requires a *proven* string in the head pair; a chain that fails that falls through to the pairwise lowering, which resolves each node from the runtime tags.

A new predicate, `string_value_is_runtime_guaranteed`, separates the two kinds of evidence `is_definitely_string_expr` had been mixing. A literal, `String(x)`, `JSON.stringify`, `path.join`, `os.arch()` *construct* a string; a `LocalGet` and a receiver-blind method name only *claim* one. Its whitelist is deliberately **closed** — an arm nobody has classified answers "claim" and gets guarded, because that costs one predictable compare while the other default costs a wrong answer.

## Cost: none measurable, and it is proved rather than sampled

I did not have to trust a stopwatch for this. Compiling all 19 corpus programs with the base compiler and the fixed compiler **against the same runtime archives** isolates the codegen delta exactly, and the answer is available statically.

**All 19 linked executables are byte-identical.** `cmp` on the binaries, base compiler vs fixed compiler, same runtime archives, same output basename: 19 identical, 0 differing. Identity is the conclusive direction — the Mach-O UUID is content-derived, so nondeterminism could manufacture a spurious *difference* but never a spurious *identity*. Identical bytes mean identical machine code, which means this change cannot have cost those programs a nanosecond.

One trap worth recording, because it cost me a wrong conclusion first time round. Perry's output **is** byte-deterministic, but the ad-hoc code signature embeds the output **basename**, and the UUID is derived over content that includes it. So building the two arms to `b_churn` and `f_churn` — the obvious thing for an A/B script to do — makes every program differ by exactly 49 bytes (16 UUID + 33 signature) on pure naming, and I briefly concluded from that that Perry's executables were not comparable at all. They are; build both arms to the **same basename in different directories**. `churn.ts` and `fib40.ts` both rebuild identically under that discipline.

The IR comparison (`--trace llvm`) agrees and says a little more about *why*:

**All 19 programs' IR differs by exactly two lines** — the `declare`s for the two new helpers, which emit no code:

```
+declare double @js_string_add_value(double, double)
+declare double @js_value_add_string(double, double)
```

Zero call sites moved, zero blocks changed, zero folds lost — the two `declare`s emit no code, which is why the linked binaries come out identical. 15 of the 19 emit no string-concat call at all; the other four (`interp`, `iso_miss`, `pipeline`, `shapes`) still call exactly the same helpers they did before, because their concats are led by string literals and a literal is a proof. `cat_local` additionally fails to *link* against the base runtime — undefined `js_string_add_value` — which is independent proof that the corpus never references the new symbols.

There is no codegen diamond anywhere in this change, which is the deliberate difference from #7831's shape. On the numeric side the fast arm is a single `fadd`, so a guard is expensive beside it and its phi blocks LLVM from proving the result is a canonical double — that is where #7831's +8.6% / +34.2% came from. Here the fast arm is a heap-allocating runtime call in every case, so the guard rides *inside* a call that was already happening and there is no phi to lose an optimisation to.

Timing on the quiet M1 mini (load 1.3–1.9, bench lock held under a unique token and released with the guarded check, best-of-5, exit code recorded per cell, arms alternated within each rep; both arms built from the same worktree with the identical `-p perry -p perry-runtime-static -p perry-stdlib-static` package set, each linking **its own** runtime archives, so this measures the whole change and not just codegen). Base is `ab1bd464b` — the branch has since been rebased twice more (onto `0321c6554`, which now carries **#7831**, #7832 and #7839); the A/B is a controlled comparison in which base and fix were built from the *same* commit, so later main-side speedups do not affect the delta. Every ceiling in the acceptance list is met, most of them by a wide margin now that #7833/#7834 have landed:

| bench | base | fix | delta | ceiling |
|---|--:|--:|--:|--:|
| churn | 0.2890 | 0.2888 | −0.1% | 0.44 |
| churn_alloc | 0.2408 | 0.2412 | +0.2% | 0.39 |
| churn_read | 0.0228 | 0.0224 | −1.8% | 0.026 |
| push_num | 0.1430 | 0.1434 | +0.3% | 0.155 |
| push_cls | 0.2366 | 0.2369 | +0.1% | 0.38 |
| cycles | 0.1114 | 0.1115 | +0.1% | 0.20 |
| deeplist | 0.1229 | 0.1222 | −0.6% | 0.13 |
| tree | 1.1619 | 1.1628 | +0.1% | 1.68 |
| tree_wide | 1.6486 | 1.6486 | +0.0% | 2.15 |
| retain | 0.3501 | 0.3498 | −0.1% | 0.37 |
| retain1 | 0.1360 | 0.1361 | +0.1% | 0.155 |
| retain_wide | 0.4590 | 0.4590 | +0.0% | 0.48 |
| retain_wide1 | 0.1591 | 0.1590 | −0.1% | 0.17 |
| fib40 | 0.3937 | 0.3937 | +0.0% | 0.41 |
| asyncpipe | 0.1333 | 0.1331 | −0.2% | 0.14 |
| shapes | 0.1839 | 0.1840 | +0.1% | 0.20 |
| interp | 1.2359 | 1.2365 | +0.0% | 1.53 |
| iso_miss | 1.6708 | 1.6712 | +0.0% | 1.96 |
| pipeline | 0.4844 | 0.4848 | +0.1% | 0.54 |

Every cell exited 0 on both arms. Nothing is outside run-to-run noise, which is what the IR comparison predicted. (The same table measured before the rebase, against `82f0e9681`, agreed to within 0.1 points on every row.)

### The cost, on a probe of the affected shape

Three probes in `gc-handoff/m0810/p7837/`, each **verified to exercise its subject** by inspecting the emitted IR rather than assuming. Best-of-21, same session, lock held:

| probe | base emits | fix emits | base | fix | delta | per concat |
|---|---|---|--:|--:|--:|--:|
| `cat_local` — `const s: string = parts[0]; s + i` ×2M | `strrecv` tag diamond + `js_string_concat_value` | **`js_string_add_value`** | 0.06794 | 0.07003 | **+3.08%** | **+1.05 ns** |
| `cat_lit` — `"id-" + i` ×2M | `js_string_concat_value` | identical IR | 0.06652 | 0.06654 | +0.02% | — |
| `cat_build` — `s = s + "ab"` ×2M | `js_string_append` | identical IR | 0.03082 | 0.03089 | +0.22% | — |

**+1.05 ns per guarded concatenation** is the honest price, and it is paid only where the compiler could prove nothing. `s` is read from an array element specifically so constant propagation cannot fold it back to a literal and route the site around the guard; `cat_local`'s loop does nothing but concatenate, so the guard's share of the work is as large as it can be — and it is still ~3%. Worth noting the fix actually *removes* IR here: the `strrecv.heap`/`strrecv.cold` unbox diamond disappears, replaced by one call.

For contrast with #7831, which is the same policy on the numeric side: its guard costs +8.6% on `this.v + 1` and +34.2% on an escaped `s += p.x + p.y`. The asymmetry is structural, not a disagreement — see the note above about `fadd` versus an allocating call.

## Validation

- `test-files/test_gap_declared_string_local_holds_number_7837.ts` — 41 assertions, **byte-identical to Node 26.5.1**. 13 of the rows are wrong on `82f0e9681` and right here. The other 28 are the reason the fix cannot just route everything to the dynamic helper: honest string concatenation, honest arithmetic, `typeof` on both, the three shapes #7837 lists as already-correct (declared field, object property, parameter pair), class fields, array elements, and six proven-string producers that must keep their fused concat.
- `gc-handoff/m0810/lyinglocal.ts` prints `49` then `99x`, byte-identical to `node --experimental-strip-types`, exit 0.
- `gc-handoff/apps/iso_miss.ts` prints `checksum 437840 misses 0`.
- All 19 corpus programs byte-identical to their expected output, exit 0.
- 10 new cargo-test-visible codegen tests (`crates/perry-codegen/src/codegen/declared_string_add_tests.rs`) asserting on which helper is emitted — each paired with the neighbouring proven shape that must NOT be guarded.
- 2 new runtime unit tests covering both helpers and `js_string_concat_box`'s totality.
- `cargo test --release -p perry-codegen --no-fail-fast`: 1301 passed, 1 failed — `large_local_array_push_inbounds_store_emits_precise_slot_barrier`, which is **pre-existing** (asserts on `js_write_barrier_slot` placement inside `apush.inbounds.` for a 2050-element array push; #7831 verified `BASE_EXIT=101` on a clean-main build).
- `cargo test --release -p perry-runtime --lib` (`RUST_TEST_THREADS=1`): 2115 passed, 0 failed.

**What I could not run: the full gap suite.** The dev host was at load 55–120 from other sessions and 20 GB free; a single gap test's compile was taking 3.5 minutes, which puts the 536-test suite at several hours, and the bench mini has 15 GB free — below the floor a parity run wants. I stopped the run at test 31/536 (2 failures, both already listed in `test-parity/gap_snapshot.json`: `test_gap_2159_defineproperty_class_prototype`, `test_gap_2514_settracesigint`). Saying so plainly rather than implying coverage I do not have.

## Left for a follow-up, deliberately

`s += x` on a declared `string` local has the same defect from the same premise — `let c: string = (42 as any); c += 1` gives `"421"`, not `43` — and it survives this PR. It is a different lowering (`lower_string_self_append`, reached from `LocalSet`, not from `binary::lower`), it sits on the load-bearing O(n) string-builder path, and fixing it correctly means hoisting a tag test above a `ToString` that has observable side effects. That wants its own change and its own measurement rather than a rider on one that currently measures at zero. Filed as #7841, with the repro, the reason the fix is not a one-liner, and a suggested shape.

### DIFF
--- a/changelog.d/7842-declared-string-is-not-a-runtime-proof.md
+++ b/changelog.d/7842-declared-string-is-not-a-runtime-proof.md
@@ -1,0 +1,29 @@
+### Fixed
+
+**A declared `string` no longer picks the `+` operator (#7837).** `is_definitely_string_expr` answered `true` on the strength of an erased TypeScript annotation, and `+` then chose string concatenation from it. Perry does not enforce declared types at runtime — CLAUDE.md says so under Known Limitations — so `const s: string = (42 as any)` really does put a number in the slot. Thirteen shapes came out silently wrong, exit 0, no diagnostic; #7835 has since fixed four of them (the ones routed through `js_string_concat_box`, which it made total). These nine were still wrong on `ab1bd464b`:
+
+| shape | Node | before |
+|---|---|---|
+| `s + 7` | `49` | `427` — concat chosen where the spec adds |
+| `7 + s` | `49` | `742` |
+| `s + true` | `43` | `42true` |
+| `a + b + "x"` (N-way fold) | `141x` | `4299x` |
+| `a + b + a` | `183` | `429942` |
+| `const u = s; u + 7` | `49` | `427` |
+| `(c ? s : "q") + 7` | `49` | `427` |
+| `arr.slice(0) + 7` | `1,27` | `` (empty) |
+| `f(a: string, b: number)` through a function value | `49` | `427` |
+
+The last two are the same premise wearing different clothes. The `.toString()` / `.slice()` / `.replace()` … arm of the predicate matches on the **method name alone**, with no look at the receiver — `Array.prototype.slice` returns an array. And a `string` PARAMETER is live too: the first triage of this bug called parameters clean because a direct call gets inlined, which erases the annotation; reached through a function value the defect is there.
+
+The policy, matching #7831 on the numeric side: **a static type may select a lowering, never an answer.** It is applied in the one place each site can afford it.
+
+- **Helpers that receive both operands NaN-boxed can be made total, and #7835 did that**: `js_string_concat_box` forwards a non-string pair to `js_dynamic_string_or_number_add` rather than decoding it as the empty string.
+- **The one-sided `l ^ r` arm could not be fixed that way**, because codegen unboxes the string operand to a `StringHeader*` before the call and the tag is gone by the time `js_string_concat_value` sees it. When the operand's string-ness is declared-only it is now passed NaN-boxed to `js_string_add_value` / `js_value_add_string`, which test the tag and then either run the identical fused single-allocation concat or fall through to the spec's `+`.
+- **The N-way chain fold** formats every part as a string, so it reproduces the source tree only when the FIRST node really concatenates. It now requires a *proven* string in the head pair; a chain that fails that falls through to the pairwise lowering, which resolves each node from the runtime tags.
+
+A new predicate, `string_value_is_runtime_guaranteed`, separates the two kinds of evidence `is_definitely_string_expr` had been mixing: a literal, `String(x)`, `JSON.stringify`, `path.join`, `os.arch()` and friends *construct* a string, while a `LocalGet` and a receiver-blind method name only *claim* one. Its whitelist is deliberately closed — an arm nobody has classified answers "claim" and gets guarded, because that costs one predictable compare while the other default costs a wrong answer.
+
+**Cost: none measurable, and it is provable rather than sampled.** Compiling all 19 corpus programs with the base and the fixed compiler against the *same* runtime archives produced LLVM IR that differs by exactly two lines — the two `declare` statements for the new helpers. Zero call sites moved, zero folds were lost, and `"prefix" + i` keeps its fused concat because a literal is a proof. The guard lands only on reads the compiler could prove nothing about, and it lands as one compare inside a call that already allocates, so there is no codegen diamond and no phi for LLVM to lose an optimization to.
+
+Still open, filed as #7841: the `s += x` **self-append** lowering has the same defect from the same premise (`let c: string = (42 as any); c += 1` gives `"421"`, not `43`). It lives in `lower_string_self_append`, not in `binary::lower`, its fix has to move a tag test above a `ToString` that has observable side effects, and it sits on the load-bearing O(n) string-builder path — so it wants its own change and its own measurement rather than a rider on this one.
--- a/crates/perry-codegen/src/codegen/declared_string_add_tests.rs
+++ b/crates/perry-codegen/src/codegen/declared_string_add_tests.rs
@@ -1,0 +1,313 @@
+//! #7837 — a declared `string` is not a proof that the value is a string, so
+//! it may not pick the `+` OPERATOR.
+//!
+//! `is_definitely_string_expr`'s `LocalGet` arm trusts `let s: string`, and
+//! Perry does not enforce annotations at runtime (CLAUDE.md, Known
+//! Limitations). `const s: string = (42 as any); s + 7` therefore selected the
+//! one-sided concat lowering and printed `"427"` where Node prints `49`.
+//!
+//! The one-sided arm is the one that cannot be repaired inside the runtime:
+//! codegen unboxes the string operand to a `StringHeader*` before the call, so
+//! `js_string_concat_value` never sees a tag to test. The fix hands it the
+//! NaN-box instead (`js_string_add_value` / `js_value_add_string`), which is
+//! why these tests assert on WHICH helper is emitted.
+//!
+//! Every test comes in a pair: the lie must be guarded, and the neighbouring
+//! shape that carries a real proof must NOT be — a fix that routed everything
+//! through the dynamic helper would pass the first half and fail the second,
+//! and would have cost `"item_" + i` its fused single-allocation concat.
+
+use crate::{compile_module, CompileOptions};
+use perry_hir::types::Type;
+use perry_hir::{BinaryOp, Expr, Function, Module, ModuleInitKind, Param, Stmt};
+
+fn ir_opts() -> CompileOptions {
+    CompileOptions {
+        emit_ir_only: true,
+        output_type: "executable".to_string(),
+        ..Default::default()
+    }
+}
+
+fn param(id: u32, name: &str, ty: Type) -> Param {
+    Param {
+        id,
+        name: name.to_string(),
+        ty,
+        default: None,
+        decorators: Vec::new(),
+        is_rest: false,
+        arguments_object: None,
+    }
+}
+
+fn probe_fn(params: Vec<Param>, body: Expr) -> Function {
+    Function {
+        id: 1,
+        name: "probe".to_string(),
+        type_params: Vec::new(),
+        params,
+        return_type: Type::Any,
+        body: vec![Stmt::Return(Some(body))],
+        is_async: false,
+        is_generator: false,
+        is_strict: true,
+        was_plain_async: false,
+        was_unrolled: false,
+        is_exported: true,
+        captures: Vec::new(),
+        decorators: Vec::new(),
+    }
+}
+
+fn module_with(function: Function) -> Module {
+    Module {
+        name: "declared_string_add.ts".to_string(),
+        imports: Vec::new(),
+        exports: Vec::new(),
+        classes: Vec::new(),
+        interfaces: Vec::new(),
+        type_aliases: Vec::new(),
+        enums: Vec::new(),
+        globals: Vec::new(),
+        functions: vec![function],
+        script_global_functions: Vec::new(),
+        references_global_this: false,
+        annexb_global_undefined_names: Vec::new(),
+        init: Vec::new(),
+        exported_native_instances: Vec::new(),
+        exported_func_return_native_instances: Vec::new(),
+        exported_objects: Vec::new(),
+        exported_functions: Vec::new(),
+        widgets: Vec::new(),
+        uses_fetch: false,
+        uses_webassembly: false,
+        extern_funcs: Vec::new(),
+        init_was_unrolled: false,
+        has_top_level_await: false,
+        init_kind: ModuleInitKind::Eager,
+        async_step_closures: std::collections::HashSet::new(),
+        closure_display_names: std::collections::HashMap::new(),
+        class_display_names: std::collections::HashMap::new(),
+        closure_source_text: std::collections::HashMap::new(),
+        async_generator_funcs: std::collections::HashSet::new(),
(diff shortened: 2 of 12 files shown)


proggeramlug:
**Correction to my own `cmp` claim, and a trap worth recording in the measurement protocol.**

I wrote earlier that Perry's executables are not byte-deterministic. That was wrong, and the way it was wrong is a trap any A/B script walks straight into.

Perry's output **is** byte-deterministic. But the ad-hoc code signature embeds the output **basename**, and the Mach-O UUID is content-derived over content that includes it. So:

| build | result |
|---|---|
| same compiler, `-o d1/p_fib40` vs `-o d2/p_fib40` | **IDENTICAL** |
| same compiler, `-o d1/b_fib40` vs `-o d1/f_fib40` | **DIFFERS, 49 bytes** |

Those 49 bytes are 16 (LC_UUID) + 33 (signature blob). Nothing else in the binary moves. `churn.ts` behaves the same way. My first sweep named the arms `b_<name>` / `f_<name>` — the obvious thing to do — and got `DIFFERS` on all 19 programs from naming alone, which is what sent me to `--trace llvm`.

Re-run with both arms written to the **same basename in different directories**, base compiler vs fixed compiler, same runtime archives:

```
identical=19 differs=0
```

**All 19 corpus executables are byte-identical.** That is the conclusive direction — a content-derived UUID can manufacture a spurious *difference* but never a spurious *identity* — and it is a stronger statement than the IR argument, because it covers the linked machine code and not just the emitted IR. The IR comparison still stands and still explains *why*: the only delta is two `declare` lines, which emit no code.

Practical rule: **hold the output basename constant across A/B arms.** A `differs` result from a script that names its arms differently is telling you about the filename, not the code.

PR #7847, opened 2026-08-11
fix(runtime): bound deep recursive value consumers

### LINKED ISSUE #7845: JSON.stringify and structuredClone segfault on deeply-nested input (the #7792 neighbours)
## Summary

#7792 fixed `JSON.parse`'s SIGSEGV on deeply nested input by measuring depth non-recursively and throwing a catchable `RangeError`. Its "check the other recursive consumers of untrusted input" note was right: **`JSON.stringify` and `structuredClone` still segfault**, from the same recursive-descent-without-a-depth-guard shape.

`main` @ `b1edd2340` (perry 0.5.1464), macOS arm64, oracle Node 26.5.1.

## `JSON.stringify` — SIGSEGV, node is fine

```ts
const D = Number(process.argv[2]);
let v: any = 0;
for (let i = 0; i < D; i++) v = [v];
console.log(JSON.stringify(v).length);
```

| depth | 1 000 | 10 000 | 30 000 | **35 000** | 50 000 | 300 000 |
|---|---|---|---|---|---|---|
| perry | 0 | 0 | 0 | **139** | **139** | **139** |
| node | 0 | 0 | 0 | 0 | 0 | 0 |

Cliff between 30k and 35k. No output, no throw — the process dies. Node stringifies 300 000 levels (600 001 chars) without complaint.

## `structuredClone` — SIGSEGV where node throws

Same builder, then `structuredClone(v)`:

| depth | 10 000 | 20 000 | 30 000 | 40 000 | **50 000** |
|---|---|---|---|---|---|
| perry | 0 | 0 | 0 | 0 | **139** |
| node | 1 (throws) | 1 (throws) | 1 (throws) | 1 (throws) | 1 (throws) |

Note the comparison cuts both ways: Perry *clones deeper than node does* (node already refuses at 10 000 with a catchable error), but past 50 000 it crashes instead of throwing. The exit-1 rows are node raising an error a `try`/`catch` can see; the exit-139 row is Perry dying.

## Why this is the same bug class as #7792

Deeply nested JSON is the canonical untrusted-input shape. A server that serialises an attacker-influenced object graph, or clones one, has the same exposure #7792 described for parsing: a segfault cannot be caught by a `try`/`catch`, so the process dies rather than returning an error. #7792's own text lists `JSON.stringify` on a deep object and `structuredClone` as the neighbours to check.

The fix shape is already in the tree — #7792 added a non-recursive depth pre-scan and a `RangeError` past 1 000 levels for `JSON.parse`. `JSON.stringify` and `structuredClone` walk a live object graph rather than text, so they need a depth counter threaded through the recursion (or an explicit worklist) rather than a pre-scan, but the acceptable outcomes are the same: parse/serialise it, or throw a catchable `RangeError`. Crashing is not one.

## Not a bug: `util.inspect`

Checked as the third neighbour. `util.inspect(v, { depth: Infinity })` does **not** crash at any depth up to 300 000 — it returns a truncated string instead. Clean with respect to this class.

## Caveat on the repro

The obvious way to write the verification walk — `while (Array.isArray(cur)) { d++; cur = cur[0] }` — does not work on Perry right now because of #7844 (`Array.isArray` returns `false` for an `any`-typed local reassigned to an array). The probes above deliberately avoid `Array.isArray` and assert on `JSON.stringify(...).length` / a completion print, so the exit codes here are not contaminated by that bug.

proggeramlug:
Fixes #7845

## Summary
- cap JSON.stringify recursion at 1,000 nested containers and throw a catchable RangeError
- apply the same bound to structuredClone and reset traversal state after caught errors
- cover the default, pretty, and replacer stringify walkers

## Validation
- reproduced exit 139 at 35,000 JSON.stringify levels and 50,000 structuredClone levels on main
- fixed binaries catch RangeError and continue at both reported depths
- cargo test -p perry --test issue_7845_deep_recursive_consumers
- cargo test -p perry-runtime (2,109 passed, 4 ignored)
- bash scripts/check_file_size.sh

## Summary by CodeRabbit

- **Bug Fixes**
  - Deeply nested values now produce catchable `RangeError`s during `JSON.stringify` and `structuredClone`, instead of exhausting the native stack.
  - Nesting is safely limited to 1,000 levels.
  - Both operations remain usable after handling the error.

- **Tests**
  - Added coverage verifying safe behavior for deeply recursive values and continued operation afterward.

### DIFF
--- a/changelog.d/7847-deep-recursive-consumers.md
+++ b/changelog.d/7847-deep-recursive-consumers.md
@@ -1,0 +1,4 @@
+**Deeply nested values now raise a catchable error during serialization and
+cloning instead of exhausting the native stack.** `JSON.stringify` and
+`structuredClone` cap recursive traversal at 1,000 nested containers and remain
+usable after the error is caught.
--- a/crates/perry-runtime/src/builtins/globals.rs
+++ b/crates/perry-runtime/src/builtins/globals.rs
@@ -413,6 +413,11 @@ fn structured_clone_unmark(ptr: usize) {
     });
 }
 
+fn reset_structured_clone_state() {
+    STRUCTURED_CLONE_IN_PROGRESS.with(|set| set.borrow_mut().clear());
+    STRUCTURED_CLONE_TRANSFER_STATE.with(|state| *state.borrow_mut() = None);
+}
+
 /// RAII guard that unmarks a pointer from the in-progress set when dropped,
 /// even on early returns from `js_structured_clone`'s POINTER_TAG branches.
 struct CloneCycleGuard(usize);
@@ -440,6 +445,15 @@ fn throw_structured_clone_type_error(message: &str) -> ! {
 fn throw_data_clone_error(message: &str) -> ! {
     let msg = js_string_from_bytes(message.as_ptr(), message.len() as u32);
     let err = crate::error::js_error_new_with_name_message(b"DataCloneError", msg);
+    crate::exception::js_throw(crate::value::js_nanbox_pointer(err as i64))
+}
+
+const MAX_STRUCTURED_CLONE_NESTING_DEPTH: usize = 1_000;
+
+fn throw_structured_clone_depth_error() -> ! {
+    let message = "structuredClone: value nested deeper than 1000 levels";
+    let msg = js_string_from_bytes(message.as_ptr(), message.len() as u32);
+    let err = crate::error::js_rangeerror_new(msg);
     crate::exception::js_throw(crate::value::js_nanbox_pointer(err as i64))
 }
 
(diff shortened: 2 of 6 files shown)


proggeramlug:
Quiet M1 mini A/B (same compiler object, runtime-only archive swap; PR parent `381eb5be4` vs head `4a4884c18`; 8 alternating samples per arm):

- shallow `JSON.stringify` loop: 1279.5 ms base, 1279.5 ms fix (0.00%)
- shallow `structuredClone` loop: 1259.0 ms base, 1265.5 ms fix (+0.52%)

The reported 35k/50k nesting cases both changed from exit 139 to a caught `RangeError`, continued to the next statement, and exited 0.

PR #7855, opened 2026-08-11
fix(codegen): invalidate Array.isArray folds on reassignment

### LINKED ISSUE #7844: Array.isArray answers from the binding's initializer, not the value — wrong in BOTH directions, so a guard admits a number
## Summary

`Array.isArray` returns **`false`** for a genuine array held in an `any`-typed local that was **reassigned** to it. Every other test of the same value says it is an array, including Perry's own `instanceof Array`, so the runtime contradicts itself.

Three lines, no loop, no library:

```ts
let b: any = 0;
b = [b];
console.log(Array.isArray(b), b instanceof Array, b.length, JSON.stringify(b));
```

```
perry 0.5.1464   false true 1 [0]      <-- Array.isArray disagrees with instanceof Array
node  26.5.1     true  true 1 [0]
```

Reproduces identically with and without `PERRY_NO_AUTO_OPTIMIZE=1`, so it is not an auto-optimize artifact.

## What triggers it

The value must reach the local by **assignment**, not by initializer. An `any` local initialized to an array literal is fine:

| case | perry | node |
|---|---|---|
| `const a: any = [1]` | **true** | true |
| `let b: any = 0; b = [b]` | **false** | true |
| `let c: any = 0; for (let i=0;i<1;i++) c = [c]` | **false** | true |
| `let d: any = 0; for (let i=0;i<3;i++) d = [d]` | **false** | true |
| `let e: any = 0; for (...) { const w: any[] = [e]; e = w }` | **false** | true |
| `const N = Number("3"); let f: any = 0; for (let i=0;i<N;i++) f = [f]` | **false** | true |

So it is the reassignment of an `any` binding, not the loop, not a constant trip count, and routing through an explicitly `any[]`-typed temporary does not rescue it.

## Why it matters

`Array.isArray` is *the* standard way to branch on "array vs not" — it is what user code, polyfills and vendored npm packages use precisely because it is supposed to be the reliable check (it works cross-realm where `instanceof` does not). A silent `false` sends such code down the non-array branch with an array in hand. This is a wrong-answer bug, not a crash, so it fails quietly.

The self-inconsistency is the sharp edge: `b instanceof Array` is `true`, `b.length` is `1`, `typeof b` is `"object"`, and `JSON.stringify(b)` is `[0]`. Only `Array.isArray` dissents, so nothing downstream has a reason to suspect the value.

## Related

#267 (`Array.isArray()` returns false on the result of `[...].map(fn)`) is the same symptom in a different shape and is closed; this looks like a surviving member of that family rather than a regression of it.

## Found by

Probing the recursive-consumer neighbours of #7792 on `main` @ `b1edd2340` (perry 0.5.1464), macOS arm64, oracle Node 26.5.1. The depth probe I was writing reported `structuredClone` losing its structure — that turned out to be this bug in the probe's own `Array.isArray` walk, not a `structuredClone` defect.

proggeramlug:
## Summary

- stop `Array.isArray` from folding a reassigned local from its initializer-refined type
- route those locals through `js_array_is_array`, so the predicate checks the current runtime value
- retain the existing compile-time fold for literals and locals that were never reassigned

Closes #7844

## Reproduction

On `origin/main` (`079e646dd`), the same six-case program produced stale answers in both directions:

```text
number-to-array false true object
string-to-array false true object
null-to-array false true object
array-to-number true false number
array-to-string true false string
unchanged-array true true object
```

The second boolean is `instanceof Array`, showing that the runtime value itself was correct. Node and the fixed compiler both produce:

```text
number-to-array true true object
string-to-array true true object
null-to-array true true object
array-to-number false false number
array-to-string false false string
unchanged-array true true object
```

The A/B used two compilers against the same prebuilt runtime.

## Validation

- `cargo test -p perry-codegen --lib`: 877 passed
- focused IR regression: passed; exactly two `js_array_is_array` calls for the two reassigned locals, while the unchanged-array control remains folded
- `native_proof_regressions`: 262 passed with only `typed_f64_receiver_method_clone_raw_loads_after_composed_guards` skipped (the current-main #7506 failure, fixed separately in #7851)
- Node/baseline/fixed semantic A/B: fixed output matches Node in all six cases
- `python3 scripts/check_test_registration.py`: passed
- `cargo fmt --all -- --check`, `git diff --check`, and `bash scripts/check_file_size.sh`: passed

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed `Array.isArray` to correctly evaluate local values at runtime after reassignment.
  - Prevented incorrect array classifications when values change between arrays and non-arrays.
  - Preserved optimized handling for unchanged array values.
- **Tests**
  - Added coverage for reassigned values, `instanceof Array`, `null`, primitives, and unchanged arrays.
- **Documentation**
  - Added a changelog entry describing the corrected behavior.

### DIFF
--- a/changelog.d/7855-array-isarray-reassignment.md
+++ b/changelog.d/7855-array-isarray-reassignment.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- `Array.isArray` now checks the current runtime value of reassigned locals instead of folding from the binding's initializer. This fixes both false negatives after assigning an array and false positives after assigning a non-array (#7844).
--- a/crates/perry-codegen/src/expr/array_methods.rs
+++ b/crates/perry-codegen/src/expr/array_methods.rs
@@ -84,7 +84,20 @@ pub(crate) fn lower(ctx: &mut FnCtx<'_>, expr: &Expr) -> Result<String> {
             // `Tuple(_)` types short-circuit; anything Union-shaped
             // falls through to the runtime.
             let v = lower_expr(ctx, o)?;
-            if let Some(ty) = crate::type_analysis::static_type_of(ctx, o) {
+            // #7844: a local's type may have been refined from its initializer,
+            // but reassignment invalidates that proof in both directions. A
+            // number-initialized local can now hold an array, and an
+            // array-initialized local can now hold a number. Mirror
+            // `is_array_expr`/`receiver_class_name`: only fold a local whose
+            // binding has not been written after initialization.
+            let static_type_is_still_valid = !matches!(
+                o.as_ref(),
+                Expr::LocalGet(id) if ctx.reassigned_locals.contains(id)
+            );
+            if let Some(ty) = static_type_is_still_valid
+                .then_some(crate::type_analysis::static_type_of(ctx, o))
+                .flatten()
+            {
                 if matches!(
                     ty,
                     perry_hir::types::Type::Array(_) | perry_hir::types::Type::Tuple(_)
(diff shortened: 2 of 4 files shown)


proggeramlug:
@coderabbitai review

PR #7864, opened 2026-08-11
perf(runtime): classify arena method receivers before side registries (#7850)

### LINKED ISSUE #7850: Every dynamic method call pays 7 side-registry probes to exclude kinds the program never creates — `is_registered_symbol` alone takes a mutex + SipHash (6.5% of `pipeline`)
## Summary

`object::native_call_method::gc_pointer_and_type_from_value` — on the path of **every dynamic method call** — runs a chain of seven side-registry probes purely to *exclude* object kinds the program may never construct. One of them, `symbol::is_registered_symbol`, takes a **process-global `pthread_mutex` plus a SipHash over a `HashSet<usize>`** once any `Symbol` has ever been minted.

On `gc-handoff/apps/pipeline.ts` that single probe is **6.5% of the program's samples** (`pthread_mutex_lock`/`unlock` ~2.6% and `RandomState::hash_one` ~2.3% of it), and the enclosing `gc_pointer_and_type_from_value` is 8.0%.

`crates/perry-runtime/src/object/native_call_method.rs:890`:

```rust
unsafe fn gc_pointer_and_type_from_value(value: f64) -> Option<(*const u8, u8)> {
    …
    if crate::buffer::is_any_array_buffer(addr)              { return Some((ptr, GC_TYPE_BUFFER)); }
    if crate::buffer::is_uint8array_buffer(addr)             { return Some((ptr, GC_TYPE_BUFFER)); }
    if crate::typedarray::lookup_typed_array_kind(addr).is_some() { return Some((ptr, GC_TYPE_TYPED_ARRAY)); }
    if !is_valid_obj_ptr(ptr) { return None; }
    if crate::set::is_registered_set(addr)
        || crate::map::is_registered_map(addr)
        || crate::regex::is_regex_pointer(ptr)
        || crate::symbol::is_registered_symbol(addr)          // <-- mutex + SipHash
    { return None; }
    let gc_header = ptr.sub(GC_HEADER_SIZE) as *const GcHeader;
    Some((ptr, (*gc_header).obj_type))                        // the answer was in the header
}
```

## The latch is not the missing piece — it already exists and is armed

`symbol::is_registered_symbol` (`symbol.rs:390`) is already gated by a `RegistryLatch`:

```rust
pub fn is_registered_symbol(ptr: usize) -> bool {
    if SYMBOL_EVER_REGISTERED.is_idle() { return false; }
    is_registered_symbol_slow(ptr)
}
```

So the 6.5% means the latch **is armed** on `pipeline` — some part of the program or the bootstrap mints at least one `Symbol` — after which every dynamic dispatch pays the full mutex + hash. A latch that a realistic program arms early is not protection; it only moves the cost behind a branch that is always taken.

This reconciles two measurements that looked contradictory:

* A peer session A/B'd the latch this morning on a **property-read** workload with no dynamic dispatch and measured **0.0221 s vs 0.0221 s — no cost at all**.
* This profile is **53.8% dynamic dispatch**, and the probe sits on `gc_pointer_and_type_from_value`, i.e. on every dynamic call.

Both are true. The honest statement is that the probe is free on monomorphic property access and costs ~6.5% on dispatch-heavy code, so the earlier "measured null" is workload-specific rather than general.

## Why this is the same family as five already-fixed bugs

`is_registered_set`, `is_registered_map`, `is_regex_pointer` and `is_arguments_object` have all been observed on hot paths in programs that construct none of the corresponding objects (see `gc-handoff/PROFILE-interp-round3.md`, which measured ~1.8–4.6% of `interp` spent proving the absence of features the program does not use). Fixes of this exact shape have already landed as #7765, #7769 and #7474, and `PER_OBJECT_LAYOUTS_NONEMPTY` / `EXTERNAL_BUFFERS_NONEMPTY` (`buffer/header.rs:33`) are the idiom.

Three independent sightings today — this one, a prototype-map latch taking a mutex plus SipHash per traced object, and the `interp` set/map/arguments probes — suggest the family is worth attacking structurally rather than one probe at a time.

## The structural fix worth considering

The `GcHeader` sitting 8 bytes before the pointer **already records `obj_type`**, and the function ends by reading it. The probe chain exists to override that answer for kinds whose identity lives in a side table. Reading the header first and consulting a side registry only when `obj_type` leaves the kind ambiguous would remove all seven probes from the common `GC_TYPE_OBJECT` case.

That requires establishing which of Set / Map / regex / symbol / buffer / typed-array instances can also carry `obj_type == GC_TYPE_OBJECT`. I have not established that, so I am filing rather than fixing.

## Sizing caveat

**The 6.5% is measured on `pipeline` only**, and `pipeline`'s dispatch load is itself a bug (#7848 — a generic-class specialization miss that forces every method call onto the dynamic path). Once #7848 lands, `pipeline`'s share of this cost largely disappears with the dynamic calls that carried it. **This should be sized across the corpus on genuinely polymorphic code before anyone optimizes for it**, not from this program's number.

## Evidence

`gc-handoff/bench/pipeline_big.ts`, `PERRY_DEBUG_SYMBOLS=1`, quiet M1 mini under the bench lock, two 3 s `sample` runs at 1 ms agreeing within 1.5 pp (2353 / 2204 stack samples):

```
js_native_call_method                        53.0%
└─ class_vtable_fast_guard                   19.6%
   └─ gc_pointer_and_type_from_value          8.0%
      ├─ is_registered_symbol_slow            6.5%
      │  ├─ RandomState::hash_one             2.3%
      │  ├─ pthread_mutex_lock                1.7%
      │  └─ pthread_mutex_unlock              0.9%
      ├─ is_registered_map                    0.3%
      ├─ is_regex_pointer                     0.3%
      ├─ is_registered_set                    0.2%
      └─ is_shared_sab / is_array_buffer      0.2%
```

proggeramlug:
Closes #7850.

## Problem

`gc_pointer_and_type_from_value` entered seven address-keyed kind probes before reading the `GcHeader` it ultimately needed. Once a program had created a Symbol, every ordinary dynamic method receiver therefore paid the process-global Symbol registry mutex and address hash.

## Fix

Use the arena range metadata to prove that current runtime-owned receivers carry a readable `GcHeader`, then classify them by GC type. Dedicated Map/Set types are rejected directly; the two shared types use their resident payload magic (RegExp/Object and fresh Symbol/String). The existing registry tower remains unchanged for outside-arena values, including boxed persistent Symbols and process-global SharedArrayBuffers.

## Tests

- Main-only regression reproduced first: an ordinary arena receiver entered the side-registry tower (`1 != 0`).
- Focused representation matrix: 3 passed (ordinary object, Buffer, typed array, Map, Set, RegExp, fresh Symbol, boxed Symbol, SharedArrayBuffer).
- `cargo test -p perry-runtime --lib`, single-threaded: all 2,120 functional tests passed; the unrelated wall-clock promise complexity guard is load-sensitive and reproduced on clean main during alternating fresh-process controls.
- `scripts/check_file_size.sh`: pass after rebasing onto v0.5.1471.

## Performance

Quiet-mini A/B is in progress; this PR remains draft until those results are posted.

### DIFF
--- a/changelog.d/7864-arena-method-receiver-classification.md
+++ b/changelog.d/7864-arena-method-receiver-classification.md
@@ -1,0 +1,4 @@
+**Dynamic method calls no longer probe every native-kind side registry for arena values.**
+Ordinary objects and other GC-managed receivers are classified from their arena metadata and
+resident header, avoiding the armed Symbol registry's global lock and address hash. Persistent
+boxed symbols and process-global shared buffers retain the existing safe registry fallback.
--- a/crates/perry-runtime/src/object/native_call_method.rs
+++ b/crates/perry-runtime/src/object/native_call_method.rs
@@ -19,6 +19,8 @@ mod string_methods;
 
 #[cfg(test)]
 mod dispatch_arg_coercion_tests;
+#[cfg(test)]
+mod gc_pointer_probe_tests;
 mod typed_array;
 
 use disposal::{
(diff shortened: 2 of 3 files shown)


proggeramlug:
Benchmark update: the arena-range version is correct, but the targeted speedup did not materialize, so I am keeping this draft rather than marking it ready on correctness alone.

Quiet M1 mini, token-guarded, seven interleaved repetitions, all cells exit-checked. The window stayed quiet (load 1.59 at open, 2.32 at close; zero foreign processes). Both compiler/runtime pairs were built separately and the PR's `perry-runtime`, `perry-runtime-static`, `perry-stdlib`, and `perry-stdlib-static` were force-rebuilt after the source change.

- All 19 corpus programs matched their saved Node output in both arms.
- Corpus medians were within ±0.72%, except the 23 ms `churn_read` cell at -2.17% (-0.5 ms); `pipeline` was +0.44% (0.2747 -> 0.2759 s).
- A genuinely dynamic `js_native_call_method_by_id` workload with the Symbol registry armed was unchanged: 0.5928 -> 0.5928 s median (best 0.5920 -> 0.5927 s).
- Its Symbol-idle control was also unchanged: 0.5899 -> 0.5897 s median.

The IR trace confirms the targeted calls are present, and unit tests prove ordinary arena receivers skip the old registry tower. The likely explanation is that the added arena-range classification replaces roughly as much work as it removes on this path. A concurrent lower-overhead header-directed variant is being measured separately; I will not race this neutral design into review.

proggeramlug:
Closing this arena-range design rather than leaving a measured-neutral draft open. The seven-repetition quiet-mini results and exact reason are in the preceding comment. A concurrent lower-overhead header-directed implementation is clean and pushed on `perf/7850-header-directed-probe-dispatch` at `7853cda82`; its full runtime suite and both corpus arms pass, and its owner is completing the quiet-mini timing/PR now. #7850 remains open until that replacement is filed.

PR #7892, opened 2026-08-11
perf(gc): slice full weak processing by holder

### LINKED ISSUE #7874: perf(gc): full/fallback weak processing is an unsliced O(total heap) atomic pause
## Summary

#6182 made copied-minor weak processing proportional to registered weak holders, but the full/fallback path was deliberately left unchanged. As a result, **one live WeakRef/WeakMap/WeakSet/FinalizationRegistry holder makes every full/fallback weak-processing phase walk every live arena object**.

The code says this directly:

- [`process_weak_targets_after_mark` walks EVERY live object](https://github.com/PerryTS/perry/blob/79d2f59ed3a294c90c7ee52a5d5ffc9fa4fedcf7/crates/perry-runtime/src/weakref.rs#L874-L903)
- [The registry already tracks and rewrites holder addresses](https://github.com/PerryTS/perry/blob/79d2f59ed3a294c90c7ee52a5d5ffc9fa4fedcf7/crates/perry-runtime/src/weakref.rs#L64-L87)
- [Copied minors already iterate only that registry](https://github.com/PerryTS/perry/blob/79d2f59ed3a294c90c7ee52a5d5ffc9fa4fedcf7/crates/perry-runtime/src/weakref.rs#L905-L949)

This is also performed as one unsliced atomic-finalize step:

- [`WeakProcessing` calls the entire pass after checking only `budget != 0`](https://github.com/PerryTS/perry/blob/79d2f59ed3a294c90c7ee52a5d5ffc9fa4fedcf7/crates/perry-runtime/src/gc/cycle.rs#L1390-L1413)
- [Policy acknowledges the phase is not internally sliced](https://github.com/PerryTS/perry/blob/79d2f59ed3a294c90c7ee52a5d5ffc9fa4fedcf7/crates/perry-runtime/src/gc/policy.rs#L3267-L3278), but incorrectly describes the spike as O(live weak holders); the implementation is O(total live heap) whenever the registry is non-empty.

## Impact

- Full GC latency scales with unrelated heap size when a program uses even one weak holder.
- `js_gc_step_us` / mutator-assist budgeting cannot bound the pause: a step with one unit can run the full walk.
- The recent full-mark-sweep pacing work reduces how often this happens, but does not bound the individual pause when it does happen.

## Proposed direction

Use `WEAK_HOLDERS` for full/fallback processing as well. The full cycle already has `ValidPointerSet` / `FullCycleLiveness`, and full/fallback pruning already removes dead holders via `prune_dead_weak_holders`. Make iteration state part of `AtomicFinalizeState` so the holder registry is processed in budgeted slices.

## Acceptance criteria

- Full/fallback weak processing cost is O(registered holders), not O(all live arena objects).
- Dead/stale registry entries are safely classified and pruned before dereference; full-cycle evacuation continues to rekey addresses.
- `WeakProcessing` maintains a cursor/snapshot and honors the supplied work budget across calls.
- Add a scaling regression: a fixed one-holder workload with 10x unrelated live objects must not produce ~10x weak-processing work/pause.
- Add correctness coverage for WeakRef, FinalizationRegistry, WeakMap, and WeakSet on full and fallback cycles, including dead holders and moved holders.

Successor to the copied-minor-only optimization in #6182.

proggeramlug:
## Summary

- replace the full/fallback whole-arena weak-holder search with a snapshot of `WEAK_HOLDERS`
- validate every snapshot address through the cycle's `ValidPointerSet` before dereferencing, pruning dead/stale young holders while retaining unprovable old holders on fallback minors
- retain weak-processing cursor state in `AtomicFinalizeCycleState` so budgeted cycles consume at most the supplied number of holders per step
- keep FinalizationRegistry's holder-local record scan atomic so `unregister()` cannot reorder an in-progress registry scan

## Reproduction

The regression keeps one live WeakRef holder constant, then adds 1,000 strongly rooted unrelated objects before a second full GC.

- before: weak-processing work grew from 4 to 1,004 arena-object visits
- after: weak-processing work stays at 1 holder unit in both collections

## Correctness coverage

- WeakRef, FinalizationRegistry, WeakMap, and WeakSet on both full and forced-fallback cycles
- holders moved by a copied minor before the full/fallback pass
- dead holder pruning and a fabricated stale registry address rejected before header access
- one-unit budget remains parked in WeakProcessing and consumes exactly one holder per step
- existing copied-minor weak registry and weak-semantics suites

## Performance

Quiet M1 mini, release runtime test, one WeakRef plus 100,000 unrelated live objects, 3 warmups and 21 measured full GCs per arm (base `c5cf2fdbf`):

| median | base | fix | delta |
|---|---:|---:|---:|
| atomic finalize | 996 us | 698 us | -29.9% |
| total full-GC pause | 8,360 us | 8,044 us | -3.8% |

## Validation

- `cargo test -p perry-runtime`: 2,146 passed, 0 failed, 4 ignored
- focused weak-holder registry: 12 passed
- focused copied-minor weak semantics: 6 passed
- cycle-state suite: 38 passed
- `python3 scripts/check_thread_locals.py`

Fixes #7874

## Summary by CodeRabbit

* **Performance**
  * Full and fallback garbage collection now process weak references in bounded slices instead of one large pause.
  * Collection work scales with registered weak-reference holders rather than the entire live heap.

* **Bug Fixes**
  * Improved handling of moved, stale, and unreachable weak-reference holders.
  * Preserved valid older-generation holders during minor collections.

* **Tests**
  * Added coverage for incremental processing, registry cleanup, and weak-reference behavior across collection modes.

### DIFF
--- a/changelog.d/7892-slice-weak-processing.md
+++ b/changelog.d/7892-slice-weak-processing.md
@@ -1,0 +1,3 @@
+Full and fallback garbage collections now process weak-reference holders from
+their registry in budgeted slices instead of scanning the entire live heap in
+one atomic pause.
--- a/crates/perry-runtime/src/gc/copying.rs
+++ b/crates/perry-runtime/src/gc/copying.rs
@@ -842,8 +842,8 @@ impl CopyingNurseryCollector {
         // never tombstoned and FinalizationRegistry never fired while
         // copied-minor was the operative cycle. Repair an already-moved
         // target's address now and queue the slot so `repair_weak_slots`
-        // fixes targets evacuated after this visit; the after-mark pass
-        // (`process_weak_targets_after_mark`) then tombstones dead ones.
+        // fixes targets evacuated after this visit; the registry pass then
+        // tombstones dead ones.
         // No remembered-set entry either — the write barrier skips weak
         // slots the same way.
         if !parent_header.is_null()
@@ -898,8 +898,8 @@ impl CopyingNurseryCollector {
     /// Second pass over the weak target slots collected during the scan:
     /// a weak target evacuated via a strong edge AFTER its slot was
     /// visited still points at the from-space original — rewrite it to
-    /// the forwarding address so `process_weak_targets_after_mark` (and
-    /// the mutator) read the live copy. Targets never forwarded are
+    /// the forwarding address so weak processing (and the mutator) read the
+    /// live copy. Targets never forwarded are
     /// either old-gen/pinned live (no rewrite needed) or dead (left for
     /// the after-mark tombstone pass).
     pub(super) unsafe fn repair_weak_slots(&mut self) {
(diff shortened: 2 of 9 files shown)


proggeramlug:
@coderabbitai review

PR #7894, opened 2026-08-11
fix(gc): make per-object layout gate authoritative

### LINKED ISSUE #7873: gc: per-object layout global latch can publish false zero and leaks armed thread exits
## Summary

#7834 introduced a process-global zero/nonzero mirror for the per-thread layout side tables. It currently uses **two independent atomics**:

- `PER_OBJECT_LAYOUT_ARMED_THREADS: AtomicIsize`
- exported `PERRY_PER_OBJECT_LAYOUTS_ANY: AtomicU8`, read directly by generated allocation sites

The disarm path reads the count and then separately stores `0` to the byte. The comment says a race can only leave a conservative stale `1`, but this interleaving leaves the authoritative fast-path byte at **false zero while a thread is armed**:

```text
initial: count=1, any=1
A: count.fetch_sub(1) -> count=0
A: count.load() == 0
B: count.fetch_add(1) -> count=1
B: any.store(1)
A: any.store(0)
final: count=1, any=0
```

Even `SeqCst` on the existing two atomics would permit this interleaving; the state needs one atomic source of truth or a synchronization protocol that cannot publish a false zero.

Separately, the source explicitly documents that a thread exiting while armed leaks its count permanently, leaving all future allocations in the process on the pre-#7834 call path.

## Impact

`0` is documented as a **proof** that no per-object record exists, and generated code uses it to skip `js_gc_forget_object_layout` entirely. A false zero can therefore skip removal of an address-keyed record at object death/reuse, allowing a recycled object address to inherit stale layout metadata. That is a GC scanning/representation correctness risk, not just a performance miss.

The TLS-exit leak is conservative for correctness but process-wide for performance: one short-lived `perry/thread` agent that exits with a record can make every subsequent inline construction pay the runtime branch/call path forever.

Relevant source:

- [The claimed invariant and admitted exit leak](https://github.com/PerryTS/perry/blob/79d2f59ed3a294c90c7ee52a5d5ffc9fa4fedcf7/crates/perry-runtime/src/gc/layout_tables.rs#L389-L415)
- [The racy two-atomic arm/disarm implementation](https://github.com/PerryTS/perry/blob/79d2f59ed3a294c90c7ee52a5d5ffc9fa4fedcf7/crates/perry-runtime/src/gc/layout_tables.rs#L417-L435)
- [Generated code treats zero as permission to omit the forget call](https://github.com/PerryTS/perry/blob/79d2f59ed3a294c90c7ee52a5d5ffc9fa4fedcf7/crates/perry-codegen/src/lower_call/typed_shape_init.rs#L128-L149)
- [The runtime also returns immediately on zero](https://github.com/PerryTS/perry/blob/79d2f59ed3a294c90c7ee52a5d5ffc9fa4fedcf7/crates/perry-runtime/src/gc/layout_tables.rs#L607-L623)

This is a focused successor to the broader layout-performance umbrella #5094.

## Acceptance criteria

- Make the generated-code zero test read a single authoritative atomic state (for example an exported atomic count with a safe zero/nonzero transition), so no interleaving can publish `0` while any thread is armed.
- Disarm the process-global state from the owning TLS value's destructor when a thread exits armed; do not depend on destructor order between separate TLS keys.
- Add a deterministic synchronization-hook test for the interleaving above. A probabilistic thread race is not enough.
- Add a thread-exit test that arms a side table, exits without emptying it, and proves the process-global state returns to zero.
- Add/retain a recycled-address correctness witness proving the gate never hides a live/stale per-object layout record.

proggeramlug:
## Summary

- replace the racy count-plus-byte mirror with one exported `AtomicU32` armed-thread count
- disarm that count from the owning `PerObjectLayoutHint` TLS destructor when a worker exits armed
- make generated allocation gates read the authoritative count with an LLVM monotonic atomic load
- retain the recycled-address cleanup path and add deterministic race, worker-exit, and address-reuse witnesses

## Reproduction

The synchronization-hook regression forces the reported ordering: A decrements the last arm, B re-arms in the former publication window, then A resumes. Before the fix it ended with `count=1, gate=0` and failed. The isolated worker test also failed before the fix with a leaked count of `1` after `join()`.

## Performance

Quiet M1 mini, same base commit and runtime/compiler pairs, 5 warmups plus 25 alternating measured pairs of the 50,000,000-allocation pointer-free typed-shape probe:

- base median: 289.646 ms
- fix median: 289.754 ms
- delta: +0.037% (noise-level)
- paired median difference: +0.252 ms

A seq-cst generated load was measured and rejected before this final version: 286.584 ms to 299.293 ms (+4.43%). The final monotonic load remains atomic and authoritative but avoids that ARM acquire-load tax; the counter transitions remain seq-cst.

## Validation

- `cargo test -p perry-runtime`: 2,145 passed, 0 failed, 4 ignored; doc tests clean
- concurrent per-object layout suite after rebase: 15 passed
- `cargo test -p perry-codegen --lib`: 897 passed
- production `cargo check -p perry-runtime -p perry-codegen`
- file-size, test-registration, TLS-budget, address-class, formatting, and whitespace gates

Fixes #7873

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved garbage collection layout tracking during concurrent operations.
  - Fixed stale layout records that could remain after memory addresses were reused.
  - Ensured thread cleanup correctly updates layout state when workers exit.
  - Strengthened allocation and object-layout cleanup reliability under concurrent workloads.

- **Tests**
  - Added coverage for concurrent layout transitions, thread-exit cleanup, and recycled-address handling.

### DIFF
--- a/changelog.d/7894-layout-latch.md
+++ b/changelog.d/7894-layout-latch.md
@@ -1,0 +1,3 @@
+Fixed the process-wide per-object GC layout gate so concurrent thread arm/disarm
+transitions cannot publish a false zero, and workers that exit with live layout
+records no longer leave the allocation fast path permanently armed.
--- a/crates/perry-codegen/src/block.rs
+++ b/crates/perry-codegen/src/block.rs
@@ -592,6 +592,19 @@ impl LlBlock {
         r
     }
 
+    /// Monotonic atomic load for an authoritative global whose value is a
+    /// fast-path gate, but which does not publish any accompanying memory.
+    pub fn load_atomic_monotonic(&mut self, ty: LlvmType, ptr: &str, alignment: u32) -> String {
+        let r = self.reg();
+        self.push_inst(crate::inst::LlInst::Load {
+            dst: r.clone(),
+            ty,
+            ptr: ptr.to_string(),
+            flavor: crate::inst::LoadFlavor::AtomicMonotonic(alignment),
+        });
+        r
+    }
+
     /// Sequentially-consistent atomic load for globals shared with runtime
     /// atomics. The explicit alignment is required by LLVM atomic loads.
     pub fn load_atomic_seq_cst(&mut self, ty: LlvmType, ptr: &str, alignment: u32) -> String {
--- a/crates/perry-codegen/src/dialect/mod.rs
+++ b/crates/perry-codegen/src/dialect/mod.rs
@@ -1243,6 +1243,17 @@ impl<'ctx, 'm> FnReader<'ctx, 'm> {
                     LF::Volatile => {
                         let _ = v.as_instruction_value().map(|i| i.set_volatile(true));
                     }
+                    LF::AtomicMonotonic(n) => {
+                        if let Some(i) = v.as_instruction_value() {
+                            unsafe {
+                                llvm_sys::core::LLVMSetOrdering(
+                                    i.as_value_ref(),
+                                    llvm_sys::LLVMAtomicOrdering::LLVMAtomicOrderingMonotonic,
+                                )
+                            };
+                            let _ = i.set_alignment(*n);
+                        }
+                    }
                     LF::AtomicSeqCst(n) => {
                         if let Some(i) = v.as_instruction_value() {
                             unsafe {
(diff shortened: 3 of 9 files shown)


proggeramlug:
@coderabbitai review

PR #7904, opened 2026-08-12
fix(indexing): classify string receivers behind array claims (#7891)

### LINKED ISSUE #7891: `xs["0"]` reads undefined on a string-valued receiver whose declared type is an array
### Summary

`xs["0"]` answers `undefined` where JavaScript answers the character, when `xs` is
**declared** `string[]` (or any array type) but holds a **string** at runtime. The
numeric form `xs[0]` is correct. The same read written directly on an `any` is correct.
So the wrong answer is selected by the *annotation*, not by the value.

### Repro

```ts
type Bag = { items: string[] };
function mk(v: any): Bag { return { items: v }; }
function viaDeclared(b: Bag): string { return "" + b.items["0"] + "/" + b.items[0]; }
const s: any = "ss";
console.log(viaDeclared(mk(s)));      // node: s/s     perry: undefined/s
const direct: any = "ss";
console.log("" + direct["0"] + "/" + direct[0]);  // node: s/s   perry: s/s
```

`test-files/test_gap_7891_string_receiver_numeric_string_key.ts`.

### Why

A declared array type is an erased annotation, not a runtime proof, so the value can be
anything. `is_array_expr` admits the receiver to `expr/index_get.rs`'s array arm, and that
arm's **static-string-key** route is
`js_array_get_index_or_string` → `array_get_property_by_key` →
`js_object_get_field_by_name` — which has no string-receiver index arm and returns
`undefined`. The numeric route is fine: `js_array_get_f64` classifies the receiver
(`clean_arr_ptr` / `array_object_receiver`) and handles a string correctly.

So the array arm's two key routes have different receiver-validation strength, and only
the numeric one is claim-safe.

### Scope / severity

Pre-existing on `main` @ `564bd997b`, reachable through a non-union declared receiver
(a plain `b: Bag` parameter, no union, no reassignment). Silent wrong answer, no crash.

Found while adding coverage for #7890, which is why #7890 deliberately **refuses** to
widen the set of receivers that reach this route: its claim is restricted to a
non-string, non-symbol key so a string key keeps exactly the path it has today.

### Suggested fix

Either give `array_get_property_by_key` the same receiver classification the numeric
route has, or refuse the array arm's static-string-key route for any receiver whose
array type is a claim rather than a proof.

proggeramlug:
Closes #7891.

## What changed

- keep an SSO receiver boxed when a string-valued key is read through an erased Array declaration
- classify heap String receivers by their GC header before the array by-name fallback
- leave the numeric guarded-array tier unchanged
- add IR gates for the string-key and numeric-key siblings plus a runtime heap-string unit test
- add source parity coverage for literal/dynamic keys, heap/SSO strings, arrays, objects, and the existing direct-string behavior

## Reproduction

On current main, the new fixture printed `undefined/s` for the declared receiver while Node printed `s/s`. The direct `any` control already printed `s/s` in both.

## Validation

- focused codegen tests: 2 passed
- focused runtime test: 1 passed
- `cargo test -p perry-codegen`: passed, including integration and doc tests
- exact wrapper build: `-p perry -p perry-runtime-static -p perry-stdlib-static`
- fixture: Perry exit 0, Node exit 0, byte-exact output
- test-registration check: 197 files across 4 registries
- file-size gate: passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed numeric string-key access when values declared as arrays contain runtime strings.
  * Preserved expected string indexing and `length` property behavior across declared, nullable, parsed, and dynamic values.
  * Maintained correct behavior for ordinary arrays and object property lookups.

* **Tests**
  * Added regression coverage for string receivers, numeric and non-numeric keys, short parsed strings, and fallback scenarios.

### DIFF
--- a/changelog.d/7904-declared-array-string-index.md
+++ b/changelog.d/7904-declared-array-string-index.md
@@ -1,0 +1,1 @@
+Fixes numeric string-key reads through erased array declarations when the runtime receiver is a string.
--- a/crates/perry-codegen/src/expr/index_get.rs
+++ b/crates/perry-codegen/src/expr/index_get.rs
@@ -275,6 +275,57 @@ fn lower_array_index_get_via_runtime_key(
     }
 }
 
+/// Read a string-valued key from a receiver admitted by an erased Array type.
+///
+/// The ordinary array ABI takes an already-unboxed `ArrayHeader*`, which loses
+/// an SSO String's tag/payload before the runtime can validate the claim. Keep
+/// the receiver boxed until that immediate representation is separated; heap
+/// Strings remain real pointers and are classified inside the array-key helper,
+/// while every other value retains the established fallback.
+fn lower_claimable_array_string_key_get(
+    ctx: &mut FnCtx<'_>,
+    arr_box: &str,
+    idx_double: &str,
+) -> String {
+    let string_idx = ctx.new_block("aidxkey.sso");
+    let array_idx = ctx.new_block("aidxkey.raw");
+    let merge_idx = ctx.new_block("aidxkey.merge");
+    let string_label = ctx.block_label(string_idx);
+    let array_label = ctx.block_label(array_idx);
+    let merge_label = ctx.block_label(merge_idx);
+
+    let bits = ctx.block().bitcast_double_to_i64(arr_box);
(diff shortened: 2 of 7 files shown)


proggeramlug:
@coderabbitai review

Issue #6206, opened 2026-07-09
GC evacuation leaves a stale reference to a moved array element (wild-pointer crash / corrupt cached value)


proggeramlug:
## Summary

Under the moving evacuation policy, a string element that is relocated during an evacuating minor can be left referenced by a stale/dangling pointer — the array slot (or a downstream reader) keeps pointing at the released from-space / reclaimed address. It surfaces as a wild-pointer crash (`EXC_BAD_ACCESS` / `EXC_ARM_DA_ALIGN`) or a silently corrupt cached value.

- **Masked by** `PERRY_GEN_GC_EVACUATE=0` (non-moving minors) — this is a reliable workaround.
- **Amplified by** `PERRY_GC_FORCE_EVACUATE=1`.
- Organically reproduces in large-heap workloads (a 13MB esbuild-bundled CLI app corrupts ~2/5 `-p` runs on the default policy; 0/N with `EVACUATE=0`). A small program needs `FORCE_EVACUATE` to cross the evacuation threshold, but it's the same code path.

This is the residual of the long-standing "includes / undefined-field / SIGBUS" corruption family. The 2026-07-09 GC audit (#6191 etc.) rooted several raw-pointer windows (sort/stringify/replace/matchAll) but this path survives.

## Reproducer (~60s, deterministic)

`evac_stale_repro.ts` (attached below). Large retained OLD-gen baseline + a lodash-style memoize whose parse fn runs `q.split(",").map(s => s.trim()).filter(Boolean)`:

```
perry compile evac_stale_repro.ts -o repro
PERRY_GC_FORCE_EVACUATE=1 ./repro     # ~6/6 crash (rc=138/139) or CORRUPT(...) exit
PERRY_GEN_GC_EVACUATE=0   ./repro     # clean
```

A/B on a single build (I added a temporary env gate): with the moving policy active the run crashes 6-8/8; with `EVACUATE=0` it's clean 6/6 (and clean 6/6 on the default policy at this heap size).

## Faulting site (map-symbolicated)

```
EXC_ARM_DA_ALIGN at 0x0000000100060001
  js_array_filter+0x418
  js_native_call_method
  js_typed_feedback_native_call_method_by_id
  perry_closure_<parse arrow>          <- q.split(",").map(...).filter(Boolean)
```

[...]

PR #7913, opened 2026-08-12
fix(gc): restore safe old-page relocation

### LINKED ISSUE #7876: gc: restore a safe old-generation defragmentation rewrite contract; production compaction is disabled
## Summary

Old-generation page defragmentation is implemented and unit-testable, but production selection is **off by default** behind `PERRY_GC_OLD_DEFRAG=1`. It was disabled after #6206 reproduced a stale non-heap reference to a moved old object (6/6 corruption with defrag enabled).

That was the correct containment action, but #6206 is closed and there is no open successor tracking restoration of the rewrite contract. The default collector therefore cannot compact partially-live old pages.

## Current consequence

The only default reclamation inside a partially-live old block is the old free list, which is exact-size only. That is effective for uniform promoted cohorts, but mixed-size churn leaves holes that cannot satisfy future differently-sized allocations; a block with any live object remains mapped.

- [Defrag is opt-in and short-circuits to an empty selection by default](https://github.com/PerryTS/perry/blob/79d2f59ed3a294c90c7ee52a5d5ffc9fa4fedcf7/crates/perry-runtime/src/gc/oldgen_defrag.rs#L111-L143)
- [The source records the stale non-heap rewrite failure](https://github.com/PerryTS/perry/blob/79d2f59ed3a294c90c7ee52a5d5ffc9fa4fedcf7/crates/perry-runtime/src/gc/oldgen_defrag.rs#L126-L137)
- [Old free-list reuse is exact-fit by design](https://github.com/PerryTS/perry/blob/79d2f59ed3a294c90c7ee52a5d5ffc9fa4fedcf7/crates/perry-runtime/src/gc/old_free.rs#L1-L38)

This leaves a structural RSS weakness for long-lived servers with heterogeneous promoted objects: logical `heapUsed` can fall while old blocks stay mapped due to scattered live objects and unusable size buckets.

## Required work

Treat this as a rewrite-contract project, not simply an env-default flip:

1. Reproduce and localize #6206's stale non-heap address with from-space/old-page protection.
2. Enumerate every metadata/IC/cache path that can retain an old movable address and require either registered rewrite coverage, epoch invalidation, pinning, or a written immovability proof.
3. Re-enable defrag only after the reproducer and a dependency-scale stress corpus are clean.
4. Add a mixed-size old-gen fragmentation ratchet that distinguishes live bytes, exact-fit free bytes, mapped old-block bytes, and bytes actually released.

`PARSE_KEY_RING` in `scripts/gc_runtime_root_holders.json` is already classified as an `open_gap` whose safety currently depends specifically on old-gen defrag being off. That exemption must be resolved before the default can flip.

## Acceptance criteria

- #6206's reproducer is deterministic red with the unsafe baseline and green with defrag enabled after the fix.
- No unrewritten old movable addresses remain in runtime caches/ICs/side tables; the inventory encodes and gates the contract.
- Defrag can be enabled by default under GC stress without corruption.
- A mixed-size retention probe demonstrates mapped old-gen pages converge toward live bytes after a full reclaim, within a pinned allowance.
- Until then, user-facing/current architecture docs state explicitly that old-page compaction is disabled and exact-fit hole reuse is the only partial-block mitigation.

proggeramlug:
## Summary

Restores the rewrite contract required for old-generation page defragmentation and enables it by default, with `PERRY_GC_OLD_DEFRAG=0` as an explicit rollback switch. The core collector fix evacuates every indexed occupant of a selected source block atomically; a minor trace cannot treat an unmarked old-generation neighbor as dead.

## Changes

- make old-page evacuation source-block complete and all-or-nothing, rejecting a block before any forwarding if an occupant is pinned, conservatively pinned, invalid, or non-movable
- add rewrite coverage for the JSON parse-key ring, the perf entry keys cache, and diagnostics symbol-keyed state
- make cached `@perry_class_keys_*` local copies precise function-lifetime mutable roots, and remove the now-invalid class-key immovability exemption from the static gate
- fail the runtime-holder inventory on `open_gap` or `unverified` movable-address contracts
- enable old-page relocation by default with a kill switch
- defer page snapshot/selection until the copying fast path declines the minor, avoiding an O(old pages) charge on ordinary copying collections
- add direct red/green holder tests, class-key IR tests, source-block neighbor/pinning tests, and a mixed-size block-reclamation ratchet

## Related issue

Fixes #7876

## Test plan

- [x] `cargo build --release -p perry -p perry-runtime-static -p perry-stdlib-static` on the quiet M1 mini
- [x] `cargo test -p perry-runtime -- --test-threads=1` — 2,159 passed, 4 ignored
- [x] `cargo test -p perry-codegen` — 903 unit tests plus integration suites passed
- [x] GC root-dominance self-test and immovable-source audit
- [x] runtime-root-holder self/live checks and GC env-knob self/live checks
- [x] `cargo fmt --all -- --check`, `git diff --check`, and `bash scripts/check_file_size.sh`
- [x] Historical #6206 workload: rebuilt main with relocation enabled fails with `TypeError: Cannot convert object to primitive value`; rebuilt fix passed 6/6 runs with evacuation verification enabled, each stdout byte-identical to the clean main control
- [x] 25-program M1 corpus, rebuilt main/fix compiler-runtime pairs, byte-identical output, best-of-15 interleaved: the initial 8.29% `retain_wide` regression was removed by deferring selection; final result +0.14% best / -0.09% median. Four tight shape/class-allocation kernels are +3.70% to +4.45% from the required class-key mutable root; `iso_miss` is +1.44%, `interp` is -2.85%, and all other programs are within +/-1.3% best-time delta.

The dependency-scale witness compiled the same 81 modules / 13.5 MB IR in both main and fix arms on the mini, but both hit the same existing missing-zod-export link failure before runtime. This branch therefore has no dependency-scale runtime result; the PR gate with a full `npm ci` environment remains authoritative.

## Screenshots / output

Not applicable.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Old-generation defragmentation is now enabled by default.
  * Defragmentation safely evacuates eligible memory blocks and preserves pinned or non-movable objects.
  * Added an environment setting to disable old-generation defragmentation when needed.

* **Bug Fixes**
  * Improved garbage-collection updates for relocated cached keys and runtime roots.
  * Reduced unnecessary metadata work during routine minor collections.
  * Improved recovery from fragmented memory and reclamation of fragmented blocks.

* **Tests**
  * Added coverage for relocation safety, rollback behavior, root updates, and fragmentation recovery.

### DIFF
--- a/changelog.d/7913-old-defrag-contract.md
+++ b/changelog.d/7913-old-defrag-contract.md
@@ -1,0 +1,32 @@
+### Fixed
+
+- Restored safe old-generation page defragmentation and enabled it by default.
+  A minor trace does not prove unmarked old objects dead, so evacuation now
+  snapshots every indexed occupant of a selected source block, rejects the
+  block before forwarding if any occupant cannot move, and otherwise relocates
+  the complete block before reclaiming it. `PERRY_GC_OLD_DEFRAG=0` remains an
+  explicit diagnosis and rollback switch.
+
+- Closed three runtime rewrite gaps: JSON's parse-key ring and the perf entry
+  keys cache now follow their structurally rooted owners, while diagnostics
+  symbol-keyed state is rekeyed after a move. Cached `@perry_class_keys_*`
+  copies are precise function-lifetime mutable roots rather than relying on the
+  old generation being immovable. The runtime-holder inventory now fails on
+  known or unevaluated movable-address gaps.
+
+The historical corruption workload fails on rebuilt main when old-page
+relocation is enabled, but passed six consecutive verified runs after this
+change with output identical to the clean control. A mixed-size regression
+also demonstrates that a fragmented source block is actually released.
+
+### Performance
+
+Old-page metadata selection now runs only after the copying-minor fast path has
+declined the collection. This removed an O(old pages) charge from ordinary
+copying minors: the worst affected retention benchmark moved from +8.29% in the
+initial implementation to +0.14% best / -0.09% median in the final quiet-M1
+best-of-15 comparison.
+
+Precise class-key roots cost 3.70-4.45% in four tight shape/class allocation
+kernels. That is the remaining correctness tradeoff; `interp` improves 2.85%,
+and the other 20 programs remain within +/-1.5%.
--- a/crates/perry-codegen/src/collectors/proven_this_routing_tests.rs
+++ b/crates/perry-codegen/src/collectors/proven_this_routing_tests.rs
@@ -780,6 +780,7 @@ fn tower_case_routes_to_proven_this_clone() {
 /// guard block → `icmp eq i64` → the branch that enters the clone's block.
 #[test]
 fn tower_route_is_guarded_by_the_class_keys_token() {
+    let _shadow = crate::codegen::helpers::NativeRootsPin::shadow();
     let ir = emit(&tower_site_module(), false);
     let bs = blocks(&ir);
     let clone = pshape_definitions(&ir)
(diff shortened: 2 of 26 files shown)


proggeramlug:
Audited and merging. The validation is the strongest in this campaign for the risk class: the historical #6206 workload rebuilt on main with relocation force-enabled reproduces the original TypeError, and the fix passes it 6/6 under evacuation verification with byte-identical stdout; the 25-program corpus is byte-identical with the perf ledger honest in both directions (the 8.29% regression found and engineered away via deferred selection; the residual +3.7–4.5% on four class-allocation kernels named as the price of the REQUIRED class-key mutable root). The exemption edits all tighten: `open_gap`/`unverified` holder verdicts now fail while relocation ships, and the class-key immovability exemption is deleted from the static gate rather than widened.

One policy clock starts at this merge, per CLAUDE.md's GC knob kill-policy: `PERRY_GC_OLD_DEFRAG=0` is a rollback switch whose OFF state has no required CI arm. Either an arm lands (a matrix row or gc-stress variant exercising `=0`), or the switch should be deleted after one release of soak — a mode that still exists is a decision that hasn't been made.

Issue #7900, opened 2026-08-12
gc: sliced weak processing exposes post-remark weak-to-strong read race


proggeramlug:
﻿## Summary

Budgeted GC performs its one-time `FinalRootRemark`, then permits mutator windows while weak processing is still incomplete. During such a window, `WeakRef.deref()` / `WeakMap.get()` can turn an unmarked weak target into a strong compiled-code local through a **read**. The incremental write barrier and allocate-black births do not observe that transition. A later weak-processing slice can therefore tombstone the target and sweep it while generated code still holds the returned pointer.

This is a correctness problem, not only a pause-quality tradeoff.

## Interleaving

With two registered weak holders and a one-holder assist budget:

1. A budgeted minor/full completes marking and runs `FinalRootRemark` once.
2. `WeakProcessing` snapshots the holder registry and processes holder A. Because the subphase is classified as sliced, it returns to the mutator with holder B still pending.
3. B's target `T` is white: weak edges are deliberately excluded from the strong trace.
4. The mutator executes `B.deref()` (or `weakMap.get(key)`) and receives `T` into a generated stack/register local. This is a read, so neither the heap-store barrier nor allocate-black birth accounting shades `T`.
5. The next assist processes B, sees `T` unmarked, clears the weak slot, and later sweep reclaims `T`.
6. Compiled code still has the strong local returned in step 4.

The native/statepoint lowering makes the local discoverable *when roots are scanned*, but no root scan occurs after this mutator window: `FinalRootRemark` has already completed.

The full path also permits a sliced remembered-set rebuild between `FinalRootRemark` and weak decisions, so the underlying contract needs to cover every post-remark mutator window, not just the newly sliced holder loop.

## Evidence

- [`FinalRootRemark` performs an unbounded one-time root scan and drain, then advances onward](https://github.com/PerryTS/perry/blob/6d9f12e60bb7c1b96cbaf868efe775245dcbaa76/crates/perry-runtime/src/gc/cycle.rs#L1366-L1405).
- [`WeakProcessing` is explicitly classified as sliced and may return to the mutator](https://github.com/PerryTS/perry/blob/6d9f12e60bb7c1b96cbaf868efe775245dcbaa76/crates/perry-runtime/src/gc/cycle.rs#L1328-L1357).
[...]

Issue #7926, opened 2026-08-12
bench: asyncpipe@120 sits exactly on a nursery trigger boundary and is not a usable A/B target


proggeramlug:
## Summary

`asyncpipe` at its shipped size (120 batches) sits **exactly on a nursery collection trigger boundary**, which makes it unusable as an A/B target and makes any change touching allocation volume look arbitrarily good or bad.

Measured while landing #7924 (`Promise.all` fast arm):

| arm | asyncpipe @120 | copying minors |
|---|--:|--:|
| before | — | **0** |
| after | **+4.7%** | **1** |

The fast arm allocates strictly *less*, and the entire regression is one extra copying minor, decided by a **444 KB window on 17.5 MB** of arena in-use. With the nursery cap inactive the same change on the same program is **−9.0%**.

At larger sizes the change is unambiguous: **@240 −33.9%, @360 −34.3%, @480 −30.7%**, with identical GC-cycle counts in both arms.

## Why this is a benchmark defect and not a #7924 defect

A workload poised on a trigger boundary measures the boundary, not the program. Any future change that moves allocation volume by a fraction of a percent will flip the cycle count and swamp its real effect — in either direction. That is the same hazard as the constant-band pacing bugs (#7592, #7690): a threshold that a workload sits exactly on.

It also interacts with two known findings:

- **#7915** — `asyncpipe`'s collection cost is per *registered root* (82 scanners over 218,455 roots), not per copied object, so a single extra minor is disproportionately expensive on this program specifically.
- **#7909** — the nursery cap on this workload is already established as counterproductive; collecting is worse on every axis (instructions 1,927 M → 2,918 M, RSS 48.9 → 76.7 MB).

So the trigger this benchmark straddles is one we already believe is mis-paced here.

## Requested

- Establish where the cliff is (bisect batch count against `copying_minors`), and record it.
- Either move the corpus `asyncpipe` off the boundary, or add a second size that is provably clear of it, so allocation-affecting changes have an honest target. **Do not simply re-scale silently** — whichever is chosen, the standings table must note that the program changed, or the round-over-round deltas become meaningless.
- Until then: **`asyncpipe`@120 must not be used to size Promise/async allocation work.** Use @240 or larger, and report the cycle counts of both arms alongside any number.

proggeramlug:
## Cliff located (measured on `ebefba51a`)

One build per size, `asyncpipe.ts` with only `const BATCHES` changed, counting
[...]

Issue #7909, opened 2026-08-12
gc: incremental old-gen work costs 14% on a program that never collects (asyncpipe, zero GC cycles)


proggeramlug:
## Summary

`asyncpipe` runs **zero GC cycles** — no `[gc]` line is ever printed — and yet **~33% of its leaf profile is collector machinery** the mutator pays anyway: write barriers, per-object layout side tables, and `RuntimeHandleScope` rooting.

`PERRY_GC_MOVING_LOOP_POLLS=0` measures **−14.1%** on `asyncpipe` **while still running zero GC cycles**. So the cost is not collection — it is an **incremental old-gen mark/sweep being driven at back-edge polls** that never completes a cycle and never reports one.

This is the largest single lever found in the `asyncpipe` round and it was deliberately left open: turning polls off is not a shippable fix (polls are default-ON since #7721 and removing them has its own history — see #7690, and the note in CLAUDE.md that a poll in a call-free clone deletes the element-shape clone). The question is why a program that never collects pays 14% for incremental collection.

## Why this matters beyond one benchmark

It falsifies a premise this campaign has been using: **"zero GC cycles" was treated as "GC levers are irrelevant here."** That is wrong. A pure-mutator program still pays incremental-collector machinery, and on `asyncpipe` that is the top item.

`PERRY_WRITE_BARRIERS=0` is **+0.9%** here, so codegen write barriers are genuinely *not* the cost — the incremental drive is.

## Suggested direction

- Establish why incremental old-gen work is being scheduled at all on a heap that never triggers a collection. If the answer is "the poll unconditionally does a slice of work", the fix is a cheap due-check before the slice, not disabling polls.
- Any fix must assert its subject was live (per CLAUDE.md's four-ways-a-gate-cannot-fail): a gate here must show incremental slices actually ran on the baseline, or it proves nothing.

## Evidence

`gc-handoff/ASYNC2-NOTES.md` — decomposition, counters, and reproduce commands. Measured with instructions retired (dev box was at load 12–165; see `gc-handoff/MEASUREMENT-PROTOCOL.md`).

proggeramlug:
### Answer: the heap DOES trigger a collection — it just never performs one

Measured on `main` @ `c109b089d`, worktree `wt-gc7909`. Instrument added in #7923.

**The premise "zero GC cycles ⇒ no collection is due" is where this went wrong.**
`gc_budgeted_due_trigger()` returns `Some(ArenaBytes)` from the young-generation
[...]

PR #7938, opened 2026-08-12
fix(gc): bound the atomic work inside a time-budgeted incremental step (#7903)

### LINKED ISSUE #7903: perf(gc): time-budgeted steps still contain unbounded atomic work
## Problem

The incremental collector's public time-budgeted entry points can still execute a single "work unit" whose cost is unbounded in heap graph size or in the number of records owned by one weak holder.

Two current paths are especially important:

1. `FinalRootRemark` runs the root scan synchronously with `usize::MAX`, then drains every object newly reached from those roots with `usize::MAX`. A root installed after the initial scan can anchor an arbitrarily large graph, so this is not bounded by root-set size.
2. Sliced weak processing advances one holder per unit, but `process_finreg_after_mark` scans the entire entries array of a `FinalizationRegistry`. A single registry can therefore make one unit arbitrarily large.

`js_gc_step_us` and mutator assist can only check elapsed time between calls to `step`; they cannot enforce their budget while either atomic unit is running. This leaves a potentially heap-sized latency spike behind an API that otherwise appears time-budgeted.

This is a performance/pause-bound issue, not a request to naively yield during final remark. Yielding back to the mutator after the liveness decision introduces the correctness race tracked in #7900; the fix needs to preserve the weak-liveness synchronization contract.

## Evidence

At current `main` (`6d9f12e60`):

- Final remark performs an unlimited synchronous root scan and unlimited trace drain:
  https://github.com/PerryTS/perry/blob/6d9f12e60/crates/perry-runtime/src/gc/cycle.rs#L1366-L1405
- Weak processing is budgeted by holder rather than by records:
  https://github.com/PerryTS/perry/blob/6d9f12e60/crates/perry-runtime/src/gc/cycle.rs#L1328-L1357
- Finalization-registry processing walks its full record array in one call:
  https://github.com/PerryTS/perry/blob/6d9f12e60/crates/perry-runtime/src/weakref.rs#L1117-L1188
- The microsecond-budget entry point requests one work unit at a time and checks time only after it returns:
  https://github.com/PerryTS/perry/blob/6d9f12e60/crates/perry-runtime/src/gc/policy.rs#L3314-L3341

The inline claim that final remark is bounded by root-set size does not cover the subsequent transitive drain.

## Impact

- `js_gc_step_us` can overshoot by the time to traverse a large newly reachable graph.
- One large `FinalizationRegistry` can dominate a weak-processing slice despite the new holder cursor.
- Mutator-assist paths inherit the same granularity problem.
- Tail latency cannot be reasoned about from the configured work/time budget alone.

## Suggested acceptance criteria

- [ ] Define and document which collector phases are intentionally atomic and their defensible maximum pause bound.
- [ ] Make weak-holder work resumable within large `FinalizationRegistry` record arrays, not just between holders.
- [ ] Either bound final-remark transitive work with a correctness-preserving protocol (for example an appropriate snapshot/barrier/handshake design), or explicitly account for it as a separate atomic-pause operation with a hard policy bound.
- [ ] Add phase-specific telemetry for maximum single-step duration and records/objects processed per step.
- [ ] Add adversarial tests for (a) a large graph becoming reachable only at final remark and (b) one registry containing a very large record array.
- [ ] Ensure the solution is compatible with the weak-read correctness requirement in #7900.

proggeramlug:
Closes #7903.

## The problem

`js_gc_step_us` consults its clock **between** work units and never during one:

```rust
let mut result = gc_budgeted_step_work_units_inner(1);
while result.status == ACTIVE && start.elapsed().as_micros() < budget_us {
    result = gc_budgeted_step_work_units_inner(1);
}
```

So the budget it advertises is exactly as strong as the most expensive single
unit, and no amount of tightening the microsecond budget can reach a unit that
scales with the heap. Two units did.

## 1. Weak processing: one unit per record, not one per holder

Weak processing charged **one work unit per registered holder** — and a
`FinalizationRegistry` is one holder however many records it owns, so
`process_finreg_after_mark` walked an arbitrarily long record array inside that
single unit. A registry with a million registrations was an atomic, heap-sized
"work unit" sitting behind a time-budgeted API.

It now charges one unit per **record**, with a cursor into the array
(`crates/perry-runtime/src/weakref/sliced.rs`).

### Why this is not just "add an index"

The previous atomicity was deliberate, and its comment said so:

> A FinalizationRegistry is one holder/work unit; its record array stays atomic
> so unregistering cannot interleave with and reorder an in-progress registry scan.

`FinalizationRegistry.prototype.unregister` **rebuilds** the entries array
without the matching records, so every index after a removed element shifts
down. A resumed index-only cursor skips exactly as many records as were removed
before it — and a skipped record is a weak slot that never gets tombstoned. On a
non-moving budgeted cycle its target is then swept and the slot dangles. **That
is a use-after-free, not a latency bug.**

So the cursor is validated rather than trusted. Alongside the index it carries
the *identity* of the array it indexes: the value word of the registry's
`entries` field plus that array's length. Both mutator mutation paths change one
of the two (`unregister` installs a freshly built array; `register` pushes), so
on resume a mismatch means the held indices are meaningless and that registry's
scan restarts from 0 against the new array.

Restarting is safe because a rescan is idempotent: the first pass writes
`undefined` into a collected record's target and `false` into its pending flag
after enqueueing, so a second pass enqueues nothing and clears nothing twice.

Restart-on-mutation alone is livelock-shaped, so restarts are **capped**; past
the cap the registry is finished in one atomic pass and *charged as such*
(`weak_registry_atomic_finishes`). The worst case is stated rather than implied:
per-step weak work is at most the requested budget plus, at most once per
registry per cycle, one forced atomic pass.

The cursor deliberately caches a *value word*, never a `*mut ArrayHeader`, since
it survives a return to the mutator — it re-validates through
`WeakLiveness::as_live_array` on every resume, exactly as the unsliced code did
on every call.

## 2. Final root remark: measured, not claimed

`AtomicFinalizeSubphase::FinalRootRemark` re-scans the roots and then drains
everything they newly reach, both at `usize::MAX`. Its inline comment claimed
the phase was "bounded by root-set size, not heap size" — true of the scan,
**false of the drain**, since a root installed after the initial scan can anchor
an arbitrarily large graph.

Both ways to bound it were rejected on correctness grounds:

- **Yield mid-drain** invalidates the remark itself. The mutator installs new
  roots, so the scan must be repeated, and repeating to a fixpoint has no
  termination guarantee under an adversarial mutator.
- **Yield after the liveness decision** is the weak-read race in #7900.

The phase therefore stays deliberately atomic, is documented as such, and its
cost is now reported **separately** from the general step maximum — folding an
intentionally-atomic phase into "the collector's worst step" is how a
heap-sized pause hides.

## 3. Telemetry — no new env knob

Per CLAUDE.md's GC knob kill-policy, everything rides the existing
`PERRY_GC_DIAG=1` path as a new line:

```
[gc-step-bounds] step_max_us= final_remark_max_us= final_remarks= \
                 weak_records= weak_max_records_per_step= weak_steps_sliced= \
                 weak_registry_restarts= weak_registry_atomic_finishes=
```

`weak_steps_sliced` is the **subject-was-live** counter. A run reporting `0` has
not exercised the sliced path at all, whatever else it reports — and most
programs register no weak holders and will legitimately report zeros across the
whole line. That is precisely why the acceptance tests drive the counters
directly instead of inferring them from a corpus run.

## Tests

`crates/perry-runtime/src/gc/tests/step_bounds.rs` — five tests, each asserting
the liveness counter *before* the bound:

| test | asserts |
|---|---|
| `one_registry_record_array_is_sliced_across_budgeted_steps` | 64 records, budget 1: some step ends mid-registry, every record is scanned, no step charges more than 1 |
| `record_slice_size_tracks_the_work_budget` | 128 records, budget 8: still sliced, ceiling tracks the budget, and is strictly less than the array |
| `mutating_the_entries_array_between_slices_restarts_the_cursor` | mutator restructures once mid-scan → the cursor detects it |
| `relentless_mutation_falls_back_to_a_bounded_atomic_finish` | mutator restructures in **every** window → reaches the bounded atomic fallback instead of restarting forever |
| `final_root_remark_is_accounted_as_a_separate_atomic_phase` | a completed budgeted cycle records at least one timed atomic remark |

## Validation

- 19/19 corpus programs byte-exact vs the pinned Node oracle, exit 0.
- `cargo fmt --all -- --check`, `scripts/check_file_size.sh`.
- `weakref.rs` was at 1988 of the 2000-line cap and `gc/cycle.rs` at 1999, so the
  new state machine went into a sibling module and the remark arm's `loop`/`break`
  was folded to the `while !…{}` idiom the adjacent drain already uses — both files
  are net smaller or level.

## What this does not close

Final remark is still atomic. This PR converts "we assert it is bounded" into
"we know how long it takes", which is what the issue's acceptance criterion
allows when a correctness-preserving slice is not available. A real bound needs
the #7900 protocol work.

Notes: `gc-handoff/PACING-NOTES.md` §2.

## Summary by CodeRabbit

- **New Features**
  - Incremental garbage collection now processes weak references in bounded, resumable steps.
  - Added safeguards for collection mutations, including controlled restarts and atomic fallback processing.
  - Added diagnostics for GC step duration, weak processing, and final root remark activity.

- **Documentation**
  - Documented GC step limits, weak-reference processing, fallback behavior, and diagnostic output.

- **Tests**
  - Added coverage for bounded weak processing, mutation handling, fallback behavior, and final remark accounting.

### DIFF
--- a/changelog.d/7938-bounded-budgeted-step-work.md
+++ b/changelog.d/7938-bounded-budgeted-step-work.md
@@ -1,0 +1,70 @@
+### Fixed
+
+- **GC: a time-budgeted incremental step no longer contains an unbounded atomic
+  work unit (#7903).** `js_gc_step_us` consults its clock *between* work units
+  and never during one, so the budget it advertises is exactly as strong as the
+  most expensive single unit. Weak processing charged **one unit per registered
+  holder** — and a `FinalizationRegistry` is one holder however many records it
+  owns, so `process_finreg_after_mark` walked an arbitrarily long record array
+  inside that one unit. A single registry with a million registrations was an
+  atomic, heap-sized "work unit" sitting behind a time-budgeted API, and no
+  amount of tightening the microsecond budget could reach it.
+
+  Weak processing now charges **one unit per record** and keeps a cursor *into*
+  the record array (`crates/perry-runtime/src/weakref/sliced.rs`).
+
+  The previous atomicity was not accidental, and preserving what it protected is
+  most of the change. `FinalizationRegistry.prototype.unregister` **rebuilds**
+  the entries array without the matching records, so every index after a removed
+  element shifts down; a resumed index-only cursor would skip exactly as many
+  records as were removed before it, and a skipped record is a weak slot that
+  never gets tombstoned — on a non-moving budgeted cycle its target is then
+  swept and the slot dangles. That is a use-after-free, not a latency bug.
+
+  So the cursor is validated rather than trusted: alongside the record index it
+  carries the *identity* of the array it indexes (the value word of the
+  registry's `entries` field plus that array's length). Both mutator mutation
+  paths change one of the two — `unregister` installs a freshly built array,
+  `register` pushes — so on resume a mismatch means the held indices are
+  meaningless and that registry's scan restarts from 0 against the new array.
+  Restarting is safe because a rescan is idempotent: the first pass writes
+  `undefined` into a collected record's target and `false` into its pending flag
+  after enqueueing, so a second pass enqueues nothing and clears nothing twice.
+  Restart-on-mutation alone is livelock-shaped, so restarts are capped; past the
+  cap a registry is finished in one atomic pass **and charged as such**. The
+  worst case is therefore stated rather than implied.
+
+- **Final root remark is now measured instead of claimed (#7903).**
+  `AtomicFinalizeSubphase::FinalRootRemark` re-scans the roots and then drains
+  everything they newly reach, both at `usize::MAX`. Its inline comment claimed
+  the phase was "bounded by root-set size, not heap size" — true of the scan,
+  **false of the drain**, since a root installed after the initial scan can
+  anchor an arbitrarily large graph. Both ways to bound it were rejected on
+  correctness grounds: yielding mid-drain invalidates the remark itself (the
+  mutator installs new roots, so the scan must repeat, with no termination
+  guarantee), and yielding after the liveness decision is the weak-read race
+  tracked in #7900. The phase therefore stays deliberately atomic, is documented
+  as such, and its cost is now reported separately from the general step maximum
+  so a heap-sized pause cannot hide behind "the collector's worst step".
+
+### Added
+
+- **`[gc-step-bounds]` diagnostic line (#7903).** Emitted on the existing
+  `PERRY_GC_DIAG=1` path — **no new environment knob**, per the GC knob
+  kill-policy. Reports `step_max_us`, `final_remark_max_us` / `final_remarks`,
+  `weak_records` / `weak_max_records_per_step`, `weak_steps_sliced`, and
+  `weak_registry_restarts` / `weak_registry_atomic_finishes`.
+
+  `weak_steps_sliced` is the **subject-was-live** counter: a run reporting `0`
+  has not exercised the sliced path at all, whatever else it reports. Most
+  programs register no weak holders and will legitimately report zeros across
+  the whole line, which is exactly why the five new acceptance tests
+  (`crates/perry-runtime/src/gc/tests/step_bounds.rs`) build the pathological
+  registry directly and assert the liveness counter *before* asserting any
+  bound — including an adversarial case where the mutator restructures the
+  entries array in every single window, which must reach the bounded atomic
+  fallback rather than restarting forever.
+
+- `docs/src/internals/gc-step-bounds.md` — which collector phases are
+  intentionally atomic, what each phase's defensible bound is, and how to read
+  the new line.
--- a/crates/perry-runtime/src/gc/cycle.rs
+++ b/crates/perry-runtime/src/gc/cycle.rs
@@ -1369,27 +1369,28 @@ impl GcCycleState {
                 if budget == 0 {
                     return;
                 }
-                // Re-scan every root with the marks nearly final (see the
-                // enum doc). Reuses the RootScan machinery unbudgeted —
-                // bounded by root-set size, not heap size. consider_evacuation
-                // is false: pinning decisions were made in the original scan,
-                // and budgeted cycles are non-moving anyway.
+                // Re-scan every root with the marks nearly final (see the enum
+                // doc). DELIBERATELY ATOMIC and NOT bounded by the step budget:
+                // the scan is bounded by root-set size, but the drain below is
+                // bounded by the newly-reachable graph, which a root installed
+                // after the initial scan can make arbitrarily large. Measured
+                // rather than claimed — `final_remark_max_us` in
+                // `[gc-step-bounds]`; docs/src/internals/gc-step-bounds.md has
+                // the bound and rejected alternatives (#7903).
+                // consider_evacuation is false: pinning was decided already.
                 {
+                    let _remark_timer = instruments::FinalRemarkTimer::start();
                     let valid_ptrs = self.valid_ptrs.as_ref().expect("valid pointer set built");
                     let minor_only = self.minor.is_some();
                     let remark_scan = self.root_scan.get_or_insert_with(RootScanCycleState::new);
-                    loop {
-                        if remark_scan.step_current_subphase(
-                            valid_ptrs,
-                            &mut self.trace,
-                            /* consider_evacuation = */ false,
-                            usize::MAX,
-                            /* allow_synchronous_scanners = */ true,
-                            minor_only,
-                        ) {
-                            break;
-                        }
-                    }
+                    while !remark_scan.step_current_subphase(
+                        valid_ptrs,
+                        &mut self.trace,
+                        /* consider_evacuation = */ false,
+                        usize::MAX,
+                        /* allow_synchronous_scanners = */ true,
+                        minor_only,
+                    ) {}
                     self.root_scan = None;
                     // Trace everything the remark newly discovered so
                     // WeakProcessing (and the full path's RS rebuild) read a
--- a/crates/perry-runtime/src/gc/instruments.rs
+++ b/crates/perry-runtime/src/gc/instruments.rs
@@ -243,3 +243,157 @@ pub fn budgeted_step_skips() -> (u64, u64, u64, u64) {
         SKIP_RESUME_BLOCKED.load(Ordering::Relaxed),
     )
 }
+
+// ---------------------------------------------------------------------------
+// #7903 — step-boundedness telemetry
+//
+// `js_gc_step_us` can only check its clock BETWEEN work units, so the honest
+// question about a "time-budgeted" collector is not "what budget was requested"
+// but "how long did the longest single step actually take, and what was it
+// doing". These counters answer that directly, and they are the liveness proof
+// for the slicing work: a run where `weak_steps_sliced` is zero has not
+// exercised the sliced path at all, however green it looks.
+// ---------------------------------------------------------------------------
+
+/// Longest single `cycle.state.step(...)` observed, microseconds.
+static STEP_MAX_US: AtomicU64 = AtomicU64::new(0);
+/// Longest single ATOMIC final-remark (root re-scan + transitive drain).
+///
+/// Deliberately a separate number from `STEP_MAX_US`: final remark is an
+/// intentionally atomic phase whose cost is bounded by the newly-reachable
+/// graph, not by the step budget. Folding it into the general maximum would let
+/// a heap-sized pause hide behind "the collector's worst step".
+static REMARK_MAX_US: AtomicU64 = AtomicU64::new(0);
+static REMARK_COUNT: AtomicU64 = AtomicU64::new(0);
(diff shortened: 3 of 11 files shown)


proggeramlug:
## Validation update

**Corpus (fix arm, this branch, own build):** 19/19 byte-exact vs the pinned Node oracle, exit 0.

**★ The instrument's first finding — a budgeted step 32× over its soft pause.**

All 19 corpus programs report `[gc-step-bounds]` all-zeros. That is a *real* zero, not a dead printer: none of them runs a budgeted incremental step at all (`cycle_starts=0` throughout). `asyncpipe` at 480 batches does run one, and it says:

```
[gc-incremental]  cycle_starts=1 steps=23 completions=1 copying_minors=3
                  mark_barrier_armed_us=263245 safepoints_blocked_by_budgeted=3
[gc-step-bounds]  step_max_us=63595 final_remark_max_us=56873 final_remarks=1
```

- `step_max_us = 63 595` against `GC_NORMAL_INCREMENTAL_SOFT_PAUSE_US = 2 000` — one single time-budgeted step ran **32× its soft pause target**.
- `final_remark_max_us = 56 873` — **89 % of that worst step is the one deliberately-atomic phase.**

That is exactly the quantity this issue said was being asserted rather than measured, and it is the argument for reporting the atomic phase on a separate axis: folded into a single "worst step" number a 57 ms atomic remark reads as ordinary collector jitter, whereas split out it says plainly that the *budget* is not the binding constraint on this pause and that #7900's protocol work is what would move it.

(`weak_records=0` there because `asyncpipe` registers no `FinalizationRegistry` — the slicing path is covered by the five unit tests, which drive the counters directly.)

proggeramlug:
## No-pacing-change evidence

Two arms, both built in this worktree (baseline = the `origin/main` `b8c1f103e` artifacts, snapshotted before any edit in this branch; fix = this branch). Both corpora re-run under `PERRY_GC_TRACE=1 PERRY_GC_DIAG=1` and tabulated on cycle counts *and* the per-minor object/byte census:

```
$ diff trace-base.txt trace-fix.txt
IDENTICAL
```

All 19 programs agree exactly on `copying minors`, `cycle_starts / steps / completions`, `copied_objects`, `promoted_objects` and `eden_live_bytes`. Same collections, same kinds, same populations — this change does not move pacing, which is what a bounding change should look like.

Baseline arm also 19/19 byte-exact, exit 0, independently.

proggeramlug:
## Correctness passes (fix arm)

| pass | result |
|---|---|
| corpus, plain | **19/19 byte-exact vs the pinned Node oracle, exit 0** |
| corpus, `PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800` | **19/19 byte-exact, exit 0** |
| corpus, `PERRY_GC_VERIFY_EVACUATION=1` | **19/19 byte-exact, exit 0** |
| `iso_miss` canary — gated on the **miss counter**, not the aggregate | `checksum 437840 misses 0`, plain and under PROTECT |
| baseline vs fix cycle-kind + object/byte census | **identical across all 19** |

### Anti-vacuity

"Identical in both arms" is ambiguous under CLAUDE.md's stale-archive trap — it is either a clean no-regression or an A/B where both arms linked the same `.a`. Discriminated directly:

| check | baseline arm | fix arm |
|---|---|---|
| `[gc-step-bounds]` lines under `PERRY_GC_DIAG=1` | **0** | **1** |
| `libperry_runtime.a` mtime | 10:54:47 | **11:15:07** — after the last source edit (10:54:03) |
| `cmp` of the two `p_asyncpipe` binaries | **DIFFER** | |

So the arms linked genuinely distinct runtime archives and the new code demonstrably executes. The identical census is a real no-regression.

Baseline provenance, since it was salvaged rather than built on purpose: it is the output of a build already in flight when the first edits landed. Cargo snapshots freshness at build start, so that build compiled the *pre-edit* `perry-runtime` — **proven, not assumed**, because the immediately following rebuild failed with `weakref.rs:878: unresolved import \`sliced\``, which is only possible if the previous build had never seen the new module.

proggeramlug:
## Unit tests

`RUST_TEST_THREADS=1 cargo test --release -p perry-runtime --lib`:
**2199 passed, 0 failed, 4 ignored.**

```
gc::tests::step_bounds::one_registry_record_array_is_sliced_across_budgeted_steps ... ok
gc::tests::step_bounds::record_slice_size_tracks_the_work_budget ... ok
gc::tests::step_bounds::mutating_the_entries_array_between_slices_restarts_the_cursor ... ok
gc::tests::step_bounds::relentless_mutation_falls_back_to_a_bounded_atomic_finish ... ok
gc::tests::step_bounds::final_root_remark_is_accounted_as_a_separate_atomic_phase ... ok
```

## Bonus: the #7926 cliff reproduces on this branch's independently-built compiler

The re-derived `asyncpipe` boundaries I posted on #7926 were measured on the baseline arm. Re-run here on the fix arm (a genuinely different runtime archive — see the anti-vacuity table above):

| BATCHES | 95 | 100 | 105 | 121 | **122** | 320 | **330** |
|---|--:|--:|--:|--:|--:|--:|--:|
| copying minors | 0 | 0 | 1 | 1 | **2** | 2 | 2 |
| `cycle_starts` | 0 | **1** | 0 | 0 | 0 | 0 | **1** |

Identical at every point including both #7909 stall-band edges, so those coordinates are a property of the program and the collector rather than of one binary.

proggeramlug:
## Sabotage proof — started, not yet confirmed

The tests pass; that is not by itself evidence they *can* fail. A sabotage run was started with `advance_registry`'s

```rust
let take = (cursor.identity.len - cursor.next).min(*remaining);
```

changed to `let take = cursor.identity.len - cursor.next;` — i.e. the whole array in one unit, the pre-#7903 behaviour. Expected: `weak_steps_sliced` drops to 0 and the liveness assertion in `one_registry_record_array_is_sliced_across_budgeted_steps` fires *before* either bound is reached.

**It had not finished compiling when I wrapped up** — the dev box was at 13–16 concurrent `rustc` and the `perry-runtime` test lib takes ~20 min there. Flagging it rather than quietly omitting it.

Two notes for whoever confirms it:

1. The fix was committed **and pushed** before the sabotage (`9a2b6fd57`), so the `git checkout --` restore could not destroy it.
2. The restore happened while the sabotaged build was still compiling, which normally makes the verdict uninterpretable. There is a clean discriminator: `grep -c "Compiling perry-runtime" /tmp/pacing-sabotage.log` — it is **1**. A **second** entry would mean cargo re-ran the unit after the restore, the verdict would be about the *fixed* code, and a green run would mean nothing. Check that count before believing the result.

Without the confirmation, what stands is an argument rather than a measurement: both assertions are liveness-first by construction — `weak_steps_sliced() > 0` cannot hold unless a step really ended mid-registry, and `weak_max_records_per_step() <= 1` cannot hold if a 64-record array went through in one unit. Neither is reachable on unsliced code.

proggeramlug:
## Sabotage proof — CONFIRMED, and it falsified a claim I made above

Reverting `advance_registry`'s slice to the pre-#7903 behaviour (`take = cursor.identity.len - cursor.next`, whole array in one unit):

```
one_registry_record_array_is_sliced_across_budgeted_steps ... FAILED
  a one-work-unit step must charge at most one record; charged 64
record_slice_size_tracks_the_work_budget ... FAILED
  the registry must still be sliced at budget 8, not swallowed whole
mutating_the_entries_array_between_slices_restarts_the_cursor ... FAILED
  restructuring the entries array under a live cursor was not detected
relentless_mutation_falls_back_to_a_bounded_atomic_finish ... FAILED
  the adversary never managed to invalidate a cursor
final_root_remark_is_accounted_as_a_separate_atomic_phase ... ok
test result: FAILED. 1 passed; 4 failed; 2198 filtered out
```

Validity discriminator checked: exactly **one** `Compiling perry-runtime` in the run log, so this is the sabotaged binary and not a rebuild that silently picked up the restore. The single pass is a control — that test covers remark timing, which the sabotage did not touch, so the sabotage was surgical rather than a blanket breakage.

### Correction to what I wrote earlier in this PR

I claimed `weak_steps_sliced() > 0` "cannot be reached on unsliced code". **That is wrong**, and the sabotage is what showed it: a step can end "mid-registry" at the *entry park*, its budget already spent resolving the holder, before a single record is scanned. So the liveness assertion passed under sabotage; the assertion that actually fired was the per-step ceiling, at `charged 64`.

The counter is asymmetric: **zero proves the path did not run; nonzero does not prove records were sliced.** `weak_max_records_per_step` is the discriminating quantity.

The suite is non-vacuous — that is now measured rather than argued. But my stated *reason* was wrong, and since this PR has already merged the fix is a follow-up: **#7941** corrects the wording in the `[gc-step-bounds]` doc block, at the assertion, and in `docs/src/internals/gc-step-bounds.md`. Comment-only, no behaviour change.

Issue #7932, opened 2026-08-12
gap suite red on main: test_gap_fetch_request_from_node_incoming_message aborts with 'there is no reactor running' (perry-ext-http/server.rs:911)


proggeramlug:
Found while running the gap suite for #7916 (PR #7928). **Not caused by that branch** — it
reproduces on a clean `origin/main` build, deterministically, 3/3 runs.

```
$ ./test_gap_fetch_request_from_node_incoming_message
thread '<unnamed>' panicked at crates/perry-ext-http/src/server/server.rs:911:13:
there is no reactor running, must be called from the context of a Tokio 1.x runtime
Abort trap: 6            # exit 134
```

Perry prints **zero** lines before aborting; node prints

```
len=55 match=true
expected len=55
```

and exits 0.

`test_gap_fetch_request_from_node_incoming_message` is **not** in
`test-parity/gap_snapshot.json`, so the gate expects it to pass and
`scripts/run_gap_tests.sh` is red on `main` because of it.

## Provenance (A/B'd, not assumed)

Two compilers, one from `origin/main` @ `ebefba51a` and one from the #7928 branch, three
runs each: `exit=134` in all six. Same panic message.

Parity is tag-gated (it has not run per-PR since v0.5.1018), which is presumably how this
landed and stayed. Sibling finding, different root cause: #7930.

proggeramlug:
**It is six tests, not one, and it is the whole HTTP/net server family.** Full gap-suite run
on `origin/main` @ `ebefba51a`, every one A/B'd against a second compiler with
byte-identical stdout and identical exit codes, so none of them belongs to the branch that
found them:

| test | exit | panic site |
|---|--:|---|
| `test_gap_fetch_request_from_node_incoming_message` | 134 | `perry-ext-http/src/server/server.rs:911` |
| `test_gap_http_client_no_redirect_follow` | 134 | `perry-ext-http/src/server/server.rs:911` |
| `test_gap_http_overloads_3226plus` | 134 | `perry-ext-http/src/server/server.rs:911` |
| `test_gap_http_req_async_iterator` | 134 | `perry-ext-http/src/server/server.rs:911` |
| `test_gap_http_res_socket_writable_onfinished` | 134 | `perry-ext-http/src/server/server.rs:911` |
| `test_gap_net_connect_bound_value` | 134 | `tokio-1.53.1/src/net/tcp/listener.rs:304` |

Five share the exact panic site; `net_connect_bound_value` hits the same "no reactor
[...]

PR #7953, opened 2026-08-12
fix(runtime): give WeakRef/FinalizationRegistry a real method surface, and brand-check every folded weak intrinsic

### LINKED ISSUE #7947: fix(runtime): WeakRef.deref() is unresolvable through an array element (arr[0].deref() throws TypeError)
## Summary

`WeakRef.prototype.deref` is unresolvable when the receiver comes from an array element.

```ts
const arr = [new WeakRef(obj)];
arr[0].deref();     // TypeError: deref is not a function
```

A directly-declared local works:

```ts
const r = arr[0];
r.deref();          // fine
```

## Provenance

Found while building a probe for #7900 (`gc-handoff/GCSAFE-NOTES.md`); it forced the probe to be reshaped around the bug. **Pre-existing** — the work that found it touches no compiler-side crate. Repro saved in the notes.

## Why it is worth its own issue

This is the shape CLAUDE.md calls out under "Known-weak areas": a method call whose receiver loses its type proof and resolves by name, so the element read produces something the dispatch path cannot recognise as a `WeakRef`. `deref` is simply the method that made it visible — the interesting question is whether the same element-read path mis-resolves other native-object methods, or whether `WeakRef` is special because of how its handle is represented.

Worth checking the neighbouring cases before scoping a fix: `WeakMap`/`WeakSet` methods through an element, and any other native object with a handle-based receiver.

### LINKED ISSUE #7948: bug(hir): name-keyed weak/proxy intrinsic folds SILENTLY hijack same-named foreign receivers (deref/get/set/has → undefined)
## Summary

The HIR pre-scan `pre_scan_weakref_locals` (`crates/perry-hir/src/lower/pre_scan/weakref_locals.rs`) records **bare identifier names** bound to `new WeakRef/WeakMap/WeakSet/FinalizationRegistry/Proxy(...)` and then folds `<name>.<method>()` straight to the corresponding runtime intrinsic. The sets are name-keyed with **no scope discrimination**, and the intrinsic helpers do **not brand-check** their receiver.

Consequence: if any binding *anywhere in the module* is `new WeakRef(x)` under the name `r`, then **every** `r.<method>()` in that module is folded onto the intrinsic — including an `r` that is a plain object, a user class instance, an array, or a **function parameter**. `js_weakref_deref` reads its `__perry_wr_target` slot by name off the foreign object, finds nothing, and answers `undefined`.

**No throw. Exit code 0. Wrong answer.** This is #7775 (`proxy_locals`) one class over, and #7775's fix — subtracting ambiguous names from the set — does not cover it.

## Repro (silent, both arms exit 0)

```ts
const obj = { tag: "T" };

class Cache {
  v: number;
  constructor(v: number) { this.v = v; }
  deref(): number { return this.v * 10; }
}

function objLiteral()      { const r = { deref: () => 40 }; return r.deref(); }
function userClass()       { const r = new Cache(4);        return r.deref(); }
function arrayWithDeref()  { const r: any = [1,2,3]; r.deref = () => 41; return r.deref(); }
function paramNamed(r: { deref: () => number }) { return r.deref(); }
function genuine()         { const r = new WeakRef(obj); return (r.deref() as any).tag; }
function control()         { const q = new Cache(4);        return q.deref(); }

console.log("A objLiteral      expect 40 ->", objLiteral());
console.log("B userClass       expect 40 ->", userClass());
console.log("C arrayWithDeref  expect 41 ->", arrayWithDeref());
console.log("D paramNamed      expect 42 ->", paramNamed({ deref: () => 42 }));
console.log("E genuine         expect T  ->", genuine());
console.log("F control         expect 40 ->", control());
```

| cell | node 26.5.1 | perry | |
|---|---|---|---|
| A objLiteral (`{ deref }` literal) | `40` | **`undefined`** | ✗ silent |
| B userClass (`class Cache { deref() }`) | `40` | **`undefined`** | ✗ silent |
| C arrayWithDeref (`arr.deref = …`) | `41` | **`undefined`** | ✗ silent |
| D paramNamed (**function parameter**) | `42` | **`undefined`** | ✗ silent |
| E genuine (the real `WeakRef`) | `T` | `T` | ✓ |
| F control (name `q`, no collision) | `40` | `40` | ✓ |

Delete `genuine()` — the only function that mentions `WeakRef` — and A–D all become correct. The presence of an unrelated `WeakRef` binding elsewhere in the module is the entire trigger.

## The same hole exists for `WeakMap`/`WeakSet`

`weakmap_locals`/`weakset_locals`/`proxy_locals` *are* subtracted by the ambiguity poison pass, but that pass only recognises two initializer shapes (`new <OtherClass>()` and a call/await result). It cannot see an object literal, an array, or a parameter, so the identical hijack goes through with method names that are far more common than `deref` — `get`, `set`, `has`, `add`, `delete`:

```ts
function objLiteralGet() { const m = { get: (k: string) => "lit:" + k }; return m.get("a"); }
function paramGet(m: { get: (k: string) => string }) { return m.get("b"); }
function arrayGet()      { const m: any = []; m.get = (k: string) => "arr:" + k; return m.get("c"); }
function genuineWeakMap(){ const k = {z:1}; const m = new WeakMap<object,number>(); m.set(k,5); return m.get(k); }
```

| cell | node | perry |
|---|---|---|
| objLiteralGet | `lit:a` | **`undefined`** |
| paramGet | `p:b` | **`undefined`** |
| arrayGet | `arr:c` | **`undefined`** |
| genuineWeakMap | `5` | `5` |

A one-letter `m`/`r`/`g`/`p` reused for a WeakMap in one function and a cache/options object in another is exactly what minified bundles look like.

## Root cause

Two independent decisions compose into a wrong answer:

1. `pre_scan_weakref_locals` is name-keyed and scope-blind (documented, deliberate) and only *partially* poisoned (`new Other()` + call/await inits; `weakref_locals`/`finreg_locals` are not poisoned at all — the in-file comment justifies that with "have no runtime method-dispatch fallback … so dropping them could regress a genuine instance with no upside").
2. The intrinsic helpers the fold targets — `js_weakref_deref`, `js_weakmap_get`, `js_finreg_register`, … — accept any pointer and read their internal slot **by name**. A foreign receiver produces `undefined` rather than a diagnostic.

Name poisoning can only ever be a partial patch for (1), because the scan cannot enumerate every way a name acquires a non-intrinsic value (parameters and destructuring bindings are not even declarations it visits). **The durable fix is (2): every folded intrinsic must brand-check its receiver and fall back to ordinary dynamic method dispatch on mismatch**, so a wrong fold degrades to the correct slow path instead of a wrong answer.

## Scope / relation to #7947

#7947 is the loud half of the same design: `WeakRef.deref` / `FinalizationRegistry.register` have *no* runtime dispatch fallback at all, so every receiver shape except a directly-named local throws `TypeError`. The fix for #7947 (adding runtime dispatch + prototype thunks for `WeakRef`/`FinalizationRegistry`) supplies exactly the fallback a brand-check needs, and the PR for #7947 also brand-checks `js_weakref_deref` / `js_finreg_register` / `js_finreg_unregister`.

**This issue tracks the remaining half: `js_weakmap_get/set/has/delete` and `js_weakset_add/has/delete` (and the `Proxy` property fold) still hijack same-named foreign receivers through initializer shapes the poison pass cannot see.**

Found while investigating #7947.

proggeramlug:
Closes #7947. Closes #7948.

## The measured blast radius

`WeakRef.prototype.deref` worked in **2 of 20 receiver shapes**. The issue's
framing ("array element") was one symptom; its claim that `const r = arr[0];
r.deref()` works is wrong — that throws too.

| receiver shape | `deref` before | `FinReg.register` before | after |
|---|---|---|---|
| `const x = new X(…)` at module top level | ✅ | ✅ | ✅ |
| `const x = new X(…)` in a function DECLARATION | ✅ | ✅ | ✅ |
| `const x = new X(…)` in an ARROW function | ❌ | ❌ | ✅ |
| `const x = new X(…)` in a function EXPRESSION | ❌ | — | ✅ |
| `const x = new X(…)` in a CLASS METHOD | ❌ | — | ✅ |
| array element `arr[0].m()` | ❌ | ❌ | ✅ |
| local copied from an element | ❌ | — | ✅ |
| object property `o.r.m()` | ❌ | ❌ | ✅ |
| local copied from a property | ❌ | — | ✅ |
| function return `mk().m()` | ❌ | — | ✅ |
| local from a call | ❌ | — | ✅ |
| `for (const x of arr)` binding | ❌ | — | ✅ |
| function PARAMETER | ❌ | — | ✅ |
| `.map((x) => x.m())` callback | ❌ | — | ✅ |
| `new X(…).m()` inline | ❌ | — | ✅ |
| `Map` value `m.get(k).m()` | ❌ | — | ✅ |
| `X.prototype.m.call(x)` | ❌ "not a function" | — | ✅ |
| `x.m.bind(x)` | ❌ "Bind must be called on a function" | — | ✅ |
| `typeof x.m` | ❌ `undefined` | — | ✅ |
| `x.m?.()` | ❌ silently `undefined` | — | ✅ |

`WeakMap`/`WeakSet` were ✅ in every one of those rows before and after — the
asymmetry is the whole story.

## Root cause

`deref`/`register`/`unregister` had **no runtime existence at all**. They were
purely an HIR fold: `pre_scan_weakref_locals` records bare local NAMES bound to
`let/const x = new WeakRef(…)` — walking module statements and the bodies of
function *declarations* only — and `expr_call/url_date_instance.rs` folds
`<tracked-name>.deref()` to `Expr::WeakRefDeref`.

Anything the fold could not name fell through to dynamic dispatch, where
nothing resolved:

* `try_weak_method_dispatch` early-returned unless the receiver's `class_id`
  was `CLASS_ID_WEAKMAP`/`CLASS_ID_WEAKSET`;
* `install_collection_proto_methods` had no `WeakRef`/`FinalizationRegistry`
  arm, so `WeakRef.prototype` carried no `deref` property at all;
* the by-name VALUE read in `get_field_by_name.rs` had a WeakMap/WeakSet arm
  but no wrapper arm.

`weakref_locals.rs` names the asymmetry in a comment, as the reason those sets
are exempt from the ambiguity poison pass:

> Restricted to weakmap/weakset only: WeakRef.deref / FinalizationRegistry
> .register / Proxy use distinct method names … and (unlike WeakMap/WeakSet)
> **have no runtime method-dispatch fallback — they rely on the codegen fast
> path**

## ★ The silent half (#7948)

The fold is name-keyed and **scope-blind**, and its helpers did not brand-check.
One genuine `const r = new WeakRef(x)` anywhere in a module folded **every**
`r.deref()` in that module onto `js_weakref_deref`, which read
`__perry_wr_target` by name off whatever it was handed and answered
`undefined`. Exit code 0, no diagnostic:

```
A objLiteral      expect 40 -> undefined     ({ deref: () => 40 })
B userClass       expect 40 -> undefined     (class Cache { deref() {…} })
C arrayWithDeref  expect 41 -> undefined     (arr.deref = …)
D paramNamed      expect 42 -> undefined     (function f(r) { return r.deref() })
E genuine         expect T  -> T
F control (name q)expect 40 -> 40
```

`weakmap_locals`/`weakset_locals`/`proxy_locals` *are* poisoned, but the poison
pass only recognises `new <OtherClass>()` and call/await initializers — it
cannot see an object literal, an array, or a parameter, so the identical hijack
went through on the far more common `get`/`set`/`has`/`add`/`delete`.

Name poisoning can only ever be a partial patch: the pre-scan cannot enumerate
every way a name acquires a non-intrinsic value (parameters and destructuring
bindings are not even declarations it visits). The durable fix is on the other
side — **the intrinsic brand-checks its receiver**.

## What changed

New module `crates/perry-runtime/src/object/weakref_proto_thunks.rs`.
`try_weak_method_dispatch` + `weak_class_id_from_receiver` moved there from
`weakref.rs` (which was at 1988/2000 lines) as a pure move, then extended.

1. **Dispatch arms** — `try_weak_method_dispatch` gains
   `("deref", CLASS_ID_WEAKREF)` and
   `("register" | "unregister", CLASS_ID_FINALIZATION_REGISTRY)`.
2. **Prototype thunks** — brand-checking `WeakRef.prototype.deref` (arity 0),
   `FinalizationRegistry.prototype.register` (2) / `.unregister` (1), hooked
   into `populate_builtin_prototype_methods`. Gives the reflective path,
   `.length`, and the spec `TypeError` for `X.prototype.m.call({})`.
3. **Value-read arm** — `get_field_by_name.rs` resolves those thunk values for
   an instance read, so `typeof wr.deref === "function"` and
   `wr.deref === WeakRef.prototype.deref`.
4. **Brand check + delegate** — `js_weakref_deref`, `js_finreg_register`,
   `js_finreg_unregister`, `js_weakmap_{set,get,has,delete}` and
   `js_weakset_add` verify the receiver's reserved `class_id` first and hand a
   foreign one to `dispatch_foreign_weak_receiver`, which re-enters
   `js_native_call_method`. A mis-fold now degrades to the correct slow path
   instead of a wrong answer. Recursion is impossible: `js_native_call_method`
   routes back into these helpers only via `try_weak_method_dispatch`, which
   requires the reserved `class_id` the brand check just rejected.
5. **`Object.prototype.toString`** — `to_string_tag.rs` gained the two missing
   arms (`[object Object]` → `[object WeakRef]` /
   `[object FinalizationRegistry]`).

## Deliberately NOT fixed

* **Weak-wrapper subclassing.** `class M extends WeakMap {}` and the WeakSet /
  WeakRef / FinalizationRegistry equivalents throw *before and after* this
  change (`value is not a function`; `Constructor WeakRef requires 'new'`).
  Verified identical against a pristine `origin/main` binary, so the new brand
  checks are not the cause. Different mechanism (constructor/prototype
  reification — `map_set_subclass` exists only for Map/Set), out of scope. The
  gap test's header names this as the pinned boundary so nobody reads a green
  run as coverage.
* **The HIR pre-scan itself** is still name-keyed, scope-blind, and still does
  not descend into arrow bodies. That is now a *performance* property rather
  than a correctness one: an unnamed receiver takes the dynamic path, and a
  mis-named one brand-checks its way back to the right method. Making the scan
  scope-aware is the real cleanup and belongs on its own.

## Validation

* `test-files/test_gap_weakref_receiver_shapes_7947.ts` — 48 output lines,
  byte-exact vs node 26.5.1, exit 0. All 14 previously-throwing receiver
  shapes, the reflective/value-read path, both `toString` tags, the brand
  check, FinalizationRegistry through three shapes, the six #7948
  WeakRef/FinReg collision cells and the five WeakMap/WeakSet ones, plus the
  WeakMap/WeakSet shapes that already worked (so a refactor of the shared
  dispatch cannot silently drop them).
* Six ad-hoc probes byte-exact vs node except the subclass rows named above.
* `cargo test --release -p perry-runtime`: 2211 passed, 0 failed.
* 19/19 realistic-apps corpus byte-exact, exit 0.
* Gap suite: no new failures vs a pristine `origin/main` build.
* Sabotage-verified with the fix committed first (see comments).

## Summary by CodeRabbit

- **New Features**
  - Expanded runtime support for `WeakRef` and `FinalizationRegistry` methods across direct, reflective, prototype, and value-based access.
  - Added accurate branding and string-tag reporting for weak-reference types.
- **Bug Fixes**
  - Invalid or foreign receivers now use their own behavior instead of being incorrectly treated as weak collections.
  - Weak-reference methods now validate receivers consistently and report incompatible calls correctly.
- **Tests**
  - Added comprehensive coverage for receiver shapes, reflection, branding, collisions, registration, mutation, and identity behavior.
- **Documentation**
  - Documented runtime support and known subclassing limitations.

### DIFF
--- a/changelog.d/7953-weakref-receiver-shapes.md
+++ b/changelog.d/7953-weakref-receiver-shapes.md
@@ -1,0 +1,93 @@
+Gave `WeakRef` and `FinalizationRegistry` an actual runtime method surface, and
+made every folded weak intrinsic brand-check its receiver.
+
+`WeakRef.prototype.deref` and `FinalizationRegistry.prototype.register` /
+`.unregister` had **no runtime existence at all**. They were purely an HIR fold:
+`pre_scan_weakref_locals` records bare local NAMES bound to
+`let/const x = new WeakRef(…)` — walking module statements and the bodies of
+function *declarations* only — and `expr_call/url_date_instance.rs` folds
+`<tracked-name>.deref()` to `Expr::WeakRefDeref`. Anything the fold could not
+name fell through to ordinary dynamic dispatch, where nothing resolved:
+`try_weak_method_dispatch` early-returned unless the receiver carried
+`CLASS_ID_WEAKMAP`/`CLASS_ID_WEAKSET`, `install_collection_proto_methods` had no
+arm for either wrapper (so `WeakRef.prototype` carried no `deref` property at
+all), and the by-name value read in `get_field_by_name.rs` had a WeakMap/WeakSet
+arm but no wrapper arm.
+
+Measured over twenty receiver shapes, **two worked**: a `const x = new WeakRef(…)`
+at module top level, and the same inside a function *declaration*. Everything
+else threw `TypeError: deref is not a function` — an array element (#7947's
+report), a local copied from one, an object property, a call result, a `for…of`
+binding, a function parameter, a `.map` callback, `new WeakRef(x).deref()`
+inline, a `Map` value, and any binding inside an arrow function, function
+expression or class method. The reflective path was equally dead:
+`WeakRef.prototype.deref.call(wr)` threw "was called on a value that is not a
+function", `wr.deref.bind(wr)` threw "Bind must be called on a function",
+`typeof wr.deref` was `undefined`, and `wr.deref?.()` silently produced
+`undefined`. `WeakMap`/`WeakSet` passed every one of those shapes both before and
+after, because they have all three routes — that asymmetry is the whole bug, and
+`weakref_locals.rs` already named it in a comment justifying why those sets are
+exempt from the ambiguity poison pass ("have no runtime method-dispatch
+fallback — they rely on the codegen fast path").
+
+Three additions close it. `try_weak_method_dispatch` gains
+`("deref", CLASS_ID_WEAKREF)` and
+`("register" | "unregister", CLASS_ID_FINALIZATION_REGISTRY)` arms, so a *call*
+on any receiver shape reaches the runtime helper. Brand-checking prototype
+thunks are installed on both prototypes via
+`populate_builtin_prototype_methods`, which fixes `.call`/`.apply`, method
+extraction, the spec `.length` values, and makes
+`WeakRef.prototype.deref.call({})` throw the `TypeError` the spec requires. And
+`get_field_by_name.rs` resolves those same thunk values for an instance read, so
+`typeof wr.deref === "function"` and `wr.deref === WeakRef.prototype.deref`.
+`Object.prototype.toString` gained the two missing arms as well
+(`[object Object]` → `[object WeakRef]` / `[object FinalizationRegistry]`).
+
+The same investigation turned up the silent sibling, filed as #7948 and closed
+here. The fold is name-keyed and **scope-blind**, and its helpers did not
+brand-check, so one genuine `const r = new WeakRef(x)` anywhere in a module
+folded *every* `r.deref()` in that module onto `js_weakref_deref` — which read
+`__perry_wr_target` by name off whatever it was handed and answered `undefined`.
+A user class instance, an object literal, an array with an attached `deref`, and
+a **function parameter** all silently returned `undefined` instead of their own
+method's result, with exit code 0. `weakmap_locals`/`weakset_locals`/
+`proxy_locals` *are* subtracted by the ambiguity poison pass, but that pass only
+recognises `new <OtherClass>()` and call/await initializers — it cannot see an
+object literal, an array, or a parameter, so the identical hijack went through on
+the far more common `get`/`set`/`has`/`add`/`delete`. Name poisoning can only
+ever be a partial patch, because the pre-scan cannot enumerate every way a name
+acquires a non-intrinsic value; parameters and destructuring bindings are not
+even declarations it visits. So the fix went on the other side:
+`js_weakref_deref`, `js_finreg_register`, `js_finreg_unregister`,
+`js_weakmap_{set,get,has,delete}` and `js_weakset_add` now verify the receiver's
+reserved `class_id` before trusting it and hand a foreign one to
+`dispatch_foreign_weak_receiver`, which re-enters `js_native_call_method`. A
+mis-fold degrades to the correct slow path instead of a wrong answer. Recursion
+is impossible: `js_native_call_method` routes back into these helpers only via
+`try_weak_method_dispatch`, which requires exactly the reserved `class_id` the
+brand check just rejected.
+
+`try_weak_method_dispatch` and `weak_class_id_from_receiver` moved out of
+`weakref.rs` (at 1988 of the 2000-line gate) into a new
+`crates/perry-runtime/src/object/weakref_proto_thunks.rs` as a pure move, then
+extended there.
+
+Deliberately left: weak-wrapper **subclassing**. `class M extends WeakMap {}` and
+the WeakSet/WeakRef/FinalizationRegistry equivalents throw before *and* after this
+change (`value is not a function`; `Constructor WeakRef requires 'new'`),
+verified identical against a pristine `origin/main` binary so the new brand
+checks cannot be blamed for it. It is a different mechanism —
+constructor/prototype reification, of which `map_set_subclass` exists only for
+Map/Set — and the gap test's header pins it as an explicit non-boundary so a
+green run is not misread as coverage. The HIR pre-scan is also still name-keyed
+and scope-blind and still does not descend into arrow bodies; that is now a
+*performance* property rather than a correctness one, since an unnamed receiver
+takes the dynamic path and a mis-named one brand-checks its way back to the right
+method.
+
+`test-files/test_gap_weakref_receiver_shapes_7947.ts` pins all of it: the 14
+previously-throwing receiver shapes, the reflective and value-read paths, both
+`toString` tags, the brand check, `FinalizationRegistry` through three shapes,
+the six WeakRef/FinReg name-collision cells and the five WeakMap/WeakSet ones,
+and the WeakMap/WeakSet shapes that already worked — so a future refactor of the
+shared dispatch cannot silently drop them.
--- a/crates/perry-runtime/src/object/collection_proto_thunks.rs
+++ b/crates/perry-runtime/src/object/collection_proto_thunks.rs
@@ -312,7 +312,7 @@ fn render_incompatible_receiver(bits: u64) -> String {
     if crate::map::is_registered_map(ptr) {
         return "#<Map>".to_string();
     }
-    match crate::weakref::weak_class_id_from_receiver(value) {
+    match crate::object::weak_class_id_from_receiver(value) {
         Some(crate::weakref::CLASS_ID_WEAKSET) => return "#<WeakSet>".to_string(),
         Some(crate::weakref::CLASS_ID_WEAKMAP) => return "#<WeakMap>".to_string(),
         _ => {}
@@ -417,7 +417,7 @@ fn map_receiver_or_throw(method: &str) -> *mut crate::map::MapHeader {
 #[inline]
 fn weak_receiver_or_throw(expected: u32, proto: &str, method: &str) -> f64 {
     let receiver = f64::from_bits(IMPLICIT_THIS.with(|c| c.get()));
-    match crate::weakref::weak_class_id_from_receiver(receiver) {
+    match super::weak_class_id_from_receiver(receiver) {
         Some(cid) if cid == expected => receiver,
         _ => throw_incompatible_receiver(proto, method, receiver.to_bits()),
     }
--- a/crates/perry-runtime/src/object/field_get_set/get_field_by_name.rs
+++ b/crates/perry-runtime/src/object/field_get_set/get_field_by_name.rs
@@ -322,7 +322,7 @@ pub extern "C" fn js_object_get_field_by_name(
     {
         unsafe {
             let boxed = f64::from_bits(JSValue::pointer(obj as *const u8).bits());
-            if let Some(cid) = crate::weakref::weak_class_id_from_receiver(boxed) {
+            if let Some(cid) = crate::object::weak_class_id_from_receiver(boxed) {
                 let name_ptr = (key as *const u8).add(std::mem::size_of::<crate::StringHeader>());
                 let name_len = (*key).byte_len as usize;
                 let name = std::slice::from_raw_parts(name_ptr, name_len);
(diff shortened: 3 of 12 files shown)


proggeramlug:
## Validation record

Rebased onto `a769fafc6` (current `main`); `git diff --name-status origin/main HEAD` shows only this PR's 12 files. All results below are from the **post-rebase** build, with `PERRY_RUNTIME_DIR` pinned and the `.a` mtimes verified to post-date the edits.

### Correctness

| check | result |
|---|---|
| `test_gap_weakref_receiver_shapes_7947.ts` vs node 26.5.1 | **byte-exact**, exit 0 (48 lines) |
| 19/19 realistic-apps corpus | **byte-exact**, `fail=0` |
| `cargo test -p perry-runtime` | 2232 tests, 0 failed (3 new) |
| `cargo fmt --all -- --check` | clean |
| `scripts/check_file_size.sh` | clean (`weakref.rs` 1988 → 1975 after the move) |
| `scripts/gc_runtime_root_holders.py` | OK — 81 holders scanned, no new unclassified |
| `scripts/addr_class_inventory.py` | fails **identically on both arms** (pre-existing hit in `object/tests.rs:646`, untouched here). The new module uses `try_read_gc_header`, the approved predicate, and does not appear in the report. |

### Sabotage verification (fix + tests committed first)

Both arms applied at once; they attribute cleanly because the tests assert independent things.

| sabotage | test that failed | observed |
|---|---|---|
| `js_weakref_deref` brand check → `if false` | `folded_weak_helpers_delegate_a_foreign_receiver_to_its_own_method` | `left: 9222246136947933185` = `0x7FFC000000000001` = **`TAG_UNDEFINED`** — the exact production symptom |
| `("deref", CLASS_ID_WEAKREF)` dispatch arm → `if false` | `dynamic_dispatch_reaches_weakref_and_finreg_methods` | `left: None` |
| — | `genuine_wrappers_still_take_the_intrinsic_path` | stayed **green** under both, so the tests are not failing indiscriminately |

The first row is the load-bearing one: it asserts a foreign receiver's own method actually *runs* (sentinel return value), so it cannot pass on the pre-fix `undefined`.

An earlier end-to-end arm caught it too — with the dispatch arm disabled the gap test dropped from 48 correct lines to an immediate throw, exit 1.

These live in `cargo-test`, a **required** per-PR context, rather than only in the tag-gated gap suite.

### Gap suite: all 8 flagged "regressions" are byte-identical on both arms

550 tests on macOS: 528 pass / 16 parity_fail / 6 crash. The snapshot gate flagged 8 regressions. A/B'd against a pristine build of the **merge-base** — none is mine:

| flagged test | fix arm | base arm | verdict |
|---|---|---|---|
| `specabi_reassign` | exit 0 | exit 0 | identical output |
| `zlib_3285_params` | exit 0 | exit 0 | identical output |
| `fetch_request_from_node_incoming_message` | exit 134 | exit 134 | identical but for the thread id |
| `http_client_no_redirect_follow` | exit 134 | exit 134 | ″ |
| `http_overloads_3226plus` | exit 134 | exit 134 | ″ |
| `http_req_async_iterator` | exit 134 | exit 134 | ″ |
| `http_res_socket_writable_onfinished` | exit 134 | exit 134 | ″ |
| `net_connect_bound_value` | exit 134 | exit 134 | ″ |

All six crashes abort at the same site — `perry-ext-http/src/server/server.rs:911`, and `net_connect_bound_value` at `tokio…/listener.rs:304` — i.e. #7932's HTTP reactor, the known main-red cause. Neither `specabi_reassign` nor `zlib_3285_params` mentions any weak API, and this PR contains no codegen edits. The 10 `node_fail -> parity_fail` status changes are oracle-availability differences between this host and the committed Linux-CI snapshot; the runner itself warns there is no macOS baseline.

**On the baseline choice:** the A/B baseline is my merge-base `1bd5eeb6b`, not current `main`. That isolates *this* delta exactly — building from current main would fold in #7960/#7961/#7968, which landed after I branched, making them confounders rather than controls. This is a correctness A/B (outputs byte-compared between the two arms), so #7968's perf regression does not bear on it.

Issue #7715, opened 2026-08-09
perf(gc): write-barrier tower B3 (slot elision) + B4 (generation bit in the header) — the unstarted half of #7187


proggeramlug:
The two unstarted follow-ups from #7187, refiled so they stop hiding behind a mostly-finished umbrella.

**What already shipped** (and is why #7187 closed): Phase A — lazy barrier arming (#7250); Phase B — the dirty-page mark cache (#7298). Both are live.

**B3 — slot elision.** Skip the write-barrier slot computation where the store provably cannot create an old→young edge. The original analysis identified this as the next lever after the mark cache.

**B4 — the generation bit.** Carry generation in the object header so `classify_heap_generation` becomes a header read rather than an arena-range classification.

## Why B4 is the interesting one

#7187's own evidence: `classify_heap_generation` was **19% of `batch.ts`** — 57.4% total GC bookkeeping — **with ZERO collections running.** That is pure mutator-side cost paid by the write-barrier tower on a program that never collects.

Related and worth reading first: #7511 established that the barrier cost is **98.5% the loop function's `keep.push(ptr)`** and **zero** from the constructor, and that the "parent array is young" shortcut is **unsound**. #7602 landed the array-store barrier design that came out of it; its design notes missed the SATB shade clause, so check that before designing anything adjacent.

## Before coding

**Re-profile.** These percentages are days old and this repo has repeatedly found ticket figures stale on re-measure (#7510's went stale three times during its own work; #7698 and #7702 both re-measured lower). Also note `PERRY_WRITE_BARRIERS=0` **cannot** be used to bound barrier cost — it changes the GC regime, so the A/B is not measuring what it appears to.

Bench host per the standing rule: the quiet M1 mini.

proggeramlug:
Re-profiled before coding, as this issue instructs. **B4's headline number does not reproduce, and its ceiling is lower than the ticket implies.**

## Getting a profile at all (worth recording — it cost me an hour)

Perry-compiled binaries carry no local symbol table, so `sample` resolves every frame as `???` and any profile of them is useless. `PERRY_DEBUG_SYMBOLS=1` at compile time fixes it (31,263 symbols vs 0). Anyone re-measuring this family needs that flag first.

## The fixture

[...]

PR #7959, opened 2026-08-12
perf(codegen): gate the guarded element store's write barrier on the stored value (#7715)

proggeramlug:
Part of #7715 (write-barrier tower, B3 "slot elision").

## The measurement that chose this change

`#7715` scopes B3 as "skip the write-barrier slot computation where the store
provably cannot create an old→young edge", and B4 as "carry generation in the
header". I re-profiled first, as the issue instructs, and the data pointed
somewhere neither bullet did.

Static census of `js_write_barrier{,_slot}` call sites across all 19 GC-corpus
programs (`--trace llvm`, bucketed by the block they sit in):

| program | gated (`*.barrier`, #7511/#7871) | **ungated `idxset.recv_prop.fast`** | value-gated `wb.maybe` |
|---|--:|--:|--:|
| pipeline | 20 | **6** | 6 |
| interp | 48 | 1 | 1 |
| iso_miss | 48 | 1 | 1 |
| everything else | ≤48 | 0 | 0–53 |

Dynamic counters (`PERRY_GC_TRACE=1 PERRY_GC_DIAG=1`, per-cycle `write_barrier`
block summed):

| program | calls | non_pointer_child | parent_not_old | slow hits |
|---|--:|--:|--:|--:|
| **pipeline** | **1 265 933** | **1 265 925 (99.999%)** | **0** | 0 |
| churn | 119 716 | 0 | 118 235 | 0 |
| interp | 11 539 | 678 | 322 | 146 |
| iso_miss | 11 302 | 596 | 172 | 146 |

`pipeline`'s `Registry.set` runs `this.vals[i] = v` 1.44 M times. Every one of
those calls decodes the stored value, finds it is a number, and returns having
done nothing — **and `parent_not_old_skips` is 0**, so the parent-generation
gate #7511/#7871 shipped would have skipped nothing there. The lever at this
site is the VALUE test, not the generation test.

## The change

`js_write_barrier_slot` at the two array-element store sites
(`emit_guarded_inbounds_array_store`, and `lower_index_set_fast`'s JSValue
in-bounds arm) now sits behind `emit_may_carry_heap_pointer_check`, and then
behind `emit_parent_may_need_remembering_check` — the same two predicates the
class-field store has carried since #7511/#7871, **nested value-first**, not
fused into one `and`.

Nesting is load-bearing: the value test is pure register arithmetic on the
stored bits, while the parent test does a `monotonic` load of
`@PERRY_INCREMENTAL_MARK_BARRIER_ACTIVE_COUNT` that LLVM may not hoist out of
the loop. A fused `and` would charge that non-hoistable load to exactly the
numeric stores this exists to make free.

## Soundness

Skipping on the value test alone is sound because `write_barrier_slot_inner`'s
first action is `barrier_child_prologue(child)`, which returns — before the SATB
shading, the armed check, the parent decode and the remembered set — when
`decode_heap_addr(child) == 0`. The emitted predicate is a documented
**superset** of that decode; `gc::tests::inline_pointer_bearing_contract`
enumerates the whole 16-bit tag space against it. A value carrying no heap
pointer has nothing to shade, which is why the incremental clause belongs to the
parent half and not this one; the parent half's obligations are unchanged and
stay pinned by `gc::tests::inline_generation_gate_contract`.

Reading the parent's header byte at `arr_handle - 7` is safe on both arms: each
reached its block through a guard that already read `obj_type` at `-8` and
`gc_flags` at `-7`, and has already performed a raw store into
`arr_handle + 8 + i*8`.

**What deliberately stays outside the value test**: the layout note. The
class-field emitter puts its note behind the same test, but that argument does
**not** transfer — `layout_note_slot` funnels
`crate::array::note_element_store` (#7480), and a non-pointer stored over a
pointer is exactly the store that must clear `GC_ARRAY_ELEMENT_SHAPE`.
`js_array_note_numeric_write` is likewise not skippable on a pointer test:
`undefined`/booleans are non-pointers that still have to downgrade the array's
raw-f64 layout.

## Tests

`expr/index_set_barrier_tests.rs` walks the emitted def chain rather than
looking for instructions nearby. It fails on: a hard-wired `br i1 true/false`
that leaves the dead predicate in the block; dropping any of the three heap-tag
comparands from the value predicate; dropping either disjunct of the parent
predicate; putting the barrier on the false edge; and eliding the call instead
of guarding it.

## B4 — reported as measured-negative, not attempted

Recorded in `gc-handoff/BARRIER-NOTES.md` §2.3/§2.5 so it is not re-derived:

* **There is no bit.** `gc_flags` is full — `gc/types.rs:975` says so in terms
  ("This is the last bit in the u8 gc_flags. Adding more flags requires
  extending GcHeader ... extending breaks ABI everywhere") — and `_reserved` is
  fully allocated and already `obj_type`-overloaded.
* **The prize is 118 k calls on one program.** `churn`'s
  `parent_not_old_skips` are the `Old ⊊ TENURED` gap: objects the non-moving
  minor promoted *logically* (`gc/oldgen.rs`) carry `TENURED` while still living
  in the nursery, so they pass the codegen gate and fail the runtime's
  `classify_heap_generation == Old`. `gc/trace.rs:799-810` is explicit that this
  state must still be traced, which is why the approximation is sound and why
  closing it needs a genuinely new bit. Outside `churn` the whole corpus totals
  494 such calls.
* **A runtime-side `!TENURED` short-circuit is worth nothing either**, for the
  same reason: every call that reaches `barrier_parent_needs_remembering` got
  there by passing the gate, so its parent is TENURED.

## Measured

Both arms built from this worktree at `ac52a5c38` (0.5.1490; the branch was
rebased onto `b392f7b5d` afterwards, and the diff is unchanged). The change is
codegen-only
(`git diff --name-only` touches nothing under `perry-runtime`/`perry-stdlib`),
so both link the **same** `libperry_{runtime,stdlib}.a` and the corpus was
compiled with the same output basenames in different directories.

**`cmp` over the 19-program GC corpus: 16 identical, 3 differ — and the 3 are
exactly `pipeline`, `interp`, `iso_miss`, i.e. precisely the programs the census
says contain the site.** The other 16 are provably unchanged and need no timing.

Instructions retired (`ab_instr.py`, best of 7, dev box, spread ≤ 0.86%):

| program | base instr (M) | fix instr (M) | Δ instr | Δ cycles |
|---|--:|--:|--:|--:|
| **pipeline** | 2657.8 | **2582.0** | **−2.85%** | **−3.91%** |
| interp | 11540.4 | 11540.7 | +0.00% | +0.73% |
| iso_miss | 14101.6 | 14101.1 | −0.00% | −0.11% |

(Direction only — final wall-clock belongs on the quiet mini. `interp`'s cycles
column moves with the box; its instruction count is flat to three digits and its
binary differs by one gated site.)

## Correctness

* 19/19 corpus programs byte-identical to `m0810/expected/`, exit 0 — including
  the `iso_miss` canary (`checksum 437840 misses 0`).
* 9 programs re-run byte-identical under `PERRY_GC_PROTECT_FROMSPACE=1
  PERRY_GC_PROTECT_FROMSPACE_DEPTH=800`, under `PERRY_GC_VERIFY_EVACUATION=1`,
  and (the 3 that differ) under `PERRY_GC_SCHEDULE_RATE=1
  PERRY_GC_SCHEDULE_ALLOC_KB=64`.
* `cargo test -p perry-codegen --lib`: 921 passed, 0 failed.
* The three new tests were **sabotage-run**, not merely written:
  hard-wiring the value gate's branch to `true` (leaving the dead predicate in
  the block) fails only `the_element_store_barrier_sits_behind_a_live_value_test`;
  dropping the incremental disjunct fails only the two parent-gate tests
  (including #7871's existing one); eliding the call inside the barrier block
  fails only `the_gated_element_store_still_reaches_the_barrier_call`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved guarded array updates by skipping write-barrier work when stored values cannot contain heap references.
  * Preserved generation and incremental-marking checks for values that may reference managed memory.
  * Maintained existing behavior for numeric and raw array stores.

* **Tests**
  * Added comprehensive coverage validating barrier conditions and generated code paths.

### DIFF
--- a/changelog.d/7715-element-store-barrier-value-gate.md
+++ b/changelog.d/7715-element-store-barrier-value-gate.md
@@ -1,0 +1,37 @@
+### Gate the guarded array-element store's write barrier on the stored value
+
+`this.vals[i] = v` — the in-bounds arm of the guarded element store — called
+`js_write_barrier_slot` unconditionally. On `gc-handoff/apps/pipeline.ts`, whose
+`Registry.set` runs that statement 1.44 M times, `PERRY_GC_TRACE=1` counts
+**1,265,933 barrier calls of which 1,265,925 (99.999%) return at
+`non_pointer_child_skips`**: the stored value is a number, so the call decodes
+it, finds no heap pointer, and returns having done nothing.
+
+The call now sits behind the same live test of the stored bits that the
+class-field store has carried since #7511, and then behind the parent's
+generation test from #7871. The two are **nested, value first**, not fused into
+one `and`: `emit_may_carry_heap_pointer_check` is pure register arithmetic on
+the value, while the parent test performs a `monotonic` load of
+`@PERRY_INCREMENTAL_MARK_BARRIER_ACTIVE_COUNT` that LLVM may not hoist out of
+the loop — fusing would charge that non-hoistable load to exactly the numeric
+stores this exists to make free.
+
+Skipping on the value test alone is sound because `write_barrier_slot_inner`'s
+first action is `barrier_child_prologue(child)`, which returns before the SATB
+shading, the armed check, the parent decode and the remembered set when
+`decode_heap_addr(child) == 0`; the emitted predicate is a documented superset
+of that decode (`gc::tests::inline_pointer_bearing_contract` enumerates the
+whole 16-bit tag space against it). A value carrying no heap pointer has nothing
+to shade, which is why the incremental clause belongs to the parent half and not
+this one. The parent half's obligations are unchanged and stay pinned by
+`gc::tests::inline_generation_gate_contract`.
+
+`expr/index_set_barrier_tests.rs` walks the emitted def chain rather than
+looking for instructions nearby: a hard-wired `br i1 true`/`false` that leaves
+the dead predicate in the block fails it, as does dropping any of the three heap
+tag comparands, putting the barrier on the false edge, or eliding the call
+instead of guarding it.
+
+Measured on the 19-program GC corpus: only `pipeline`, `interp` and `iso_miss`
+contain the site at all, and `parent_not_old_skips` is 0 on all three — the
+value test is the lever there, not the generation test.
--- a/crates/perry-codegen/src/expr/class_field_barrier_tests.rs
+++ b/crates/perry-codegen/src/expr/class_field_barrier_tests.rs
@@ -48,7 +48,7 @@ const BARRIER_CALL: &str = "call void @js_write_barrier_slot";
 /// These tests describe DEFAULT barrier emission. `PERRY_WRITE_BARRIERS=0`
 /// removes every barrier and would make all of them vacuously "pass" the
 /// negative half while the positive half fails for the wrong reason.
-fn assert_default_barrier_env_not_disabled() {
+pub(super) fn assert_default_barrier_env_not_disabled() {
     assert!(
         !matches!(
             std::env::var("PERRY_WRITE_BARRIERS").as_deref(),
@@ -58,7 +58,7 @@ fn assert_default_barrier_env_not_disabled() {
     );
 }
 
-fn ir_opts() -> CompileOptions {
+pub(super) fn ir_opts() -> CompileOptions {
     CompileOptions {
         target: None,
         is_entry_module: true,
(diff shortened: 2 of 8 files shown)


proggeramlug:
★ Follow-up push: `cargo test -p perry-codegen --lib` was 921/921 green while
`cargo test -p perry-codegen` (all targets) was **red** —
`typed_shape_descriptors::pointer_store_into_numeric_array_keeps_layout_note_and_barrier`
asserted `js_write_barrier_slot` appeared between the `idxset.inbounds.` and
`idxset.check_cap.` labels, which is exactly where the barrier stopped being.
Integration suites under `crates/*/tests/*.rs` do not run per-PR, so that would
have landed green and sat red until the nightly.

Updated, not weakened: the assertion now follows the **edge** — the in-bounds
arm must branch into `idxset.inbounds.barrier.maybe.`, and the gate block must
still contain the call *and* the incremental-count read. "The call exists
somewhere in the module" would pass even if this arm stopped reaching a barrier.

That test is also the standing proof that `lower_index_set_fast`'s JSValue arm
is reachable in real programs — a pointer stored into a statically `number[]`
local — even though the 19-program GC corpus contains zero such sites.

Full `cargo test -p perry-codegen`: exit 0, 27/27 binaries ok. Compiler output
is unchanged by this push (test-only), so the measurements above still stand.

Issue #7915, opened 2026-08-12
gc: asyncpipe collects at 1200-1650 ns/object, including a 122 ms minor that handled zero objects


proggeramlug:
## Summary

Measured while establishing the per-cycle GC picture for the retain cluster (`gc-handoff/RETAIN4-NOTES.md`):

**`asyncpipe` collects at 1,200–1,650 ns/object when it collects at all — including a 122 ms minor that handled ZERO objects.**

For scale, the retain cluster's in-place promotion runs at **8.9 ns/object** and a real evacuation of 245,752 objects costs 30.2 ms. So `asyncpipe`'s collections are two orders of magnitude more expensive per object, and one of them did no per-object work whatsoever yet still cost 122 ms — longer than the entire program's 127 ms runtime.

A collection whose cost is independent of the number of objects it processes is a **fixed overhead being paid per cycle**, not collection work. That points at root scanning, side-table maintenance, or a per-cycle setup/teardown cost that scales with something other than the live set.

## Relationship to #7909

#7909 covers the *incremental* old-gen work `asyncpipe` pays while running **zero** cycles. **This issue is different**: it is about the cost of the cycles that *do* run, once the workload is scaled far enough to trigger them (per `gc-handoff/ASYNC2-NOTES.md`, 120 batches → zero cycles, 240 → two copying minors, 1200 → GC-dominated).

The two may share a root cause. They should be investigated together but neither should be closed by the other without evidence.

## Acceptance

- Explain what a 122 ms minor spends its time on when it processes zero objects.
- A gate that asserts its subject was live (per CLAUDE.md's "four ways a gate can be unable to fail") — e.g. assert a nonzero cycle count *and* a bounded per-cycle fixed cost, not merely that nothing threw.

## Note on measuring this

`gc-handoff/apps/asyncpipe_big.ts` is **not** a valid scaled proxy for `asyncpipe` — the zero-GC property is a knife-edge of the exact batch count. Any scaling used to trigger collections must be reported explicitly, along with confirmation that the mechanism under study survived the scaling.

proggeramlug:
### Lead from #7909: the per-cycle cost is the ROOT SCAN, and the denominator is wrong

Investigating #7909 I had to unblock `asyncpipe`'s collector (a starved budgeted
incremental cycle was rejecting every precise safepoint via
`gc_safepoint_moving_minor`'s `budgeted` entry guard). That produced two real
[...]

Issue #7902, opened 2026-08-12
memory(gc): untraced promotion can retain an old-heap-sized phase-change cohort as live


proggeramlug:
﻿## Summary

The untraced whole-block promotion path charges retained garbage from the **previous** cycle's survival ratio, not from the liveness of the bytes it is currently promoting. After a live-to-dead phase change, a prior 1000‰ measurement makes `implied_dead = 0`, so the advertised 32 MiB `PROMOTED_DEAD_BUDGET_BYTES` is not charged while the collector blindly promotes the new dead cohort.

The separate stale-estimate budget allows `max(128 MiB, old-gen-at-last-measurement)` of such promotion before forcing a traced cycle. On a large live heap that can be hundreds of MiB or GiB. The eventual traced cycle measures only its *current* young cohort; it cannot recover or retroactively charge garbage already promoted by the preceding untraced cycles. Those bytes remain in old-gen until a full collection.

This is a bounded-policy design, not an immediate memory-safety bug, but the bound is much larger than the comments/changelog's 32 MiB footprint claim and grows with old-gen size.

## Phase-change sequence

1. A traced copying minor observes 1000‰ young survival.
2. The application changes phase and begins allocating/droping short-lived cohorts.
3. The next promotion(s) satisfy the stale 1000‰ predictor and skip tracing.
4. Every parseable young object is promoted and accounted as live, including objects that are now unreachable.
5. `note_untraced_promotion()` computes zero implied dead bytes from the stale 1000‰ value, so `PROMOTED_DEAD_BYTES` remains unchanged.
6. Only after `UNTRACED_PROMOTED_BYTES` reaches `max(128 MiB, old-gen-at-last-measurement)` does a traced *promoting* cycle run. It updates the future predictor but cannot identify/reclaim the earlier promoted garbage.
7. `credit_promoted_bytes_to_old_baseline()` and live-census publication treat all untraced promoted bytes as live, which can defer the full collection needed to reclaim them.

## Evidence

- [The untraced threshold is based solely on the last measured ratio](https://github.com/PerryTS/perry/blob/6d9f12e60bb7c1b96cbaf868efe775245dcbaa76/crates/perry-runtime/src/gc/promote_in_place.rs#L243-L261).
[...]

PR #7961, opened 2026-08-12
fix(gc): denominate the nursery constant band in objects (#7929)

### LINKED ISSUE #7929: perf(gc): the nursery/promotion budget is denominated in BYTES but the collector's cost is per-OBJECT — compaction is taxed back
Found while measuring #7916/#7714 (PR #7928, `INLINE_SLOT_FLOOR` 4 → 2).

## The mismatch

The nursery/promotion budget is denominated in **bytes**. The collector's per-cycle cost is
dominated by **per-object** work (copy, forwarding, layout transfer, page registration —
~50–100 ns each). Those two are not the same quantity, and the gap opens exactly when object
representation gets more compact.

## The measurement

Same program (`gc-handoff/bench/retain1.ts`, 1 M surviving 2-field literals), same commit,
one arm with 72-byte objects and one with 56-byte objects. Every minor fires at the **same
byte mark and processes the same bytes**:

| minor | bytes, 72 B objects → 56 B objects | objects, 72 B → 56 B |
|--:|---|---|
| 1 (copy)    | 17 694 064 → 17 694 216 | 245 752 → **315 969** |
| 2 (promote) | 18 742 672 → 18 742 816 | 260 316 → **334 694** |
| 3 (promote) | 34 601 760 → 34 602 008 | 480 580 → **617 893** |

The object ratio is **1.286 = 72/56, exactly**. `deeplist` reproduces it to three decimals.

Consequences on those two rows: GC pause 39.60 → 50.10 ms (+26.5%), and that +10.5 ms
*exceeds* the whole program's cycle delta (25.2 M cycles ≈ 7.9 ms) — the mutator got faster
and the collector got slower. Cost per extra promoted object: 10.5 ms / 211 691 = **49.6 ns**,
i.e. the per-object price did not change at all. Only the population inside a fixed byte
budget did.

## Why it matters beyond one PR

Object representation is the current perf direction (#7916 established that the collector is
no longer the binding constraint on the retain cluster; the mutator is). But **every byte
removed from an object increases the number of objects the collector handles per unit of its
own budget.** A byte-denominated trigger in front of a per-object cost model makes "make
objects smaller" and "make the collector do more work per cycle" the same action, so the
footprint work partially pays for itself in GC time until this is fixed.

## What "fixed" would mean

An object-count term alongside the byte term in the nursery/promotion budget, so a cycle's
*work* — not its *bytes* — is what is being paced. Adjacent to #7715 (B4 generation bit) and
#7432 (adaptive tenuring young cap, which is the valve reading `eden_live_bytes` here).

## Caveat that keeps this honest

Total promotion work is set by the surviving object count, which does not change — 1 M either
way. `retain1` promotes 740 896 of 1 M objects with 72-byte objects and 952 587 with 56-byte
ones, so the smaller representation pulls ~211 k promotions *forward* into the measurement
window rather than creating them. A program that runs to completion pays the same total.
That makes the effect largest on exit-bounded microbenchmarks — which is most of the perf
corpus, so it is still worth pacing correctly.

Reproduce: `PERRY_GC_TRACE=1 PERRY_GC_DIAG=1 ./p_retain1` on either side of PR #7928 and
compare the `[gc-copy-minor] ran …` lines. Full write-up: `gc-handoff/REPR-NOTES.md` §3d.

proggeramlug:
Closes #7929.

## The mismatch

The scavenge nursery trigger compares from-space **bytes** against a constant
16 MB band (`young_scavenge_cap_due`), while the copying minor's cost is per
**object**. Shrinking a representation therefore silently buys the collector more
work per cycle. #7928 took a two-field object literal 72 B → 56 B and every minor
then moved **1.286× (= 72/56)** as many objects for the same bytes, costing
`deeplist` +10.5% and `retain1` +8.8% wall on the quiet mini.

This scales the constant band by the mean size of the objects the last copying
minor actually moved, so the band buys a constant **object** budget.

## Why this is a single comparison

Of the four byte-denominated valves, the two `tenuring.rs` ratios are
representation-invariant by cancellation — `desired_survivor_bytes` is `cap/16`
and `retune_nursery_cap_scale` compares eden against `cap/25`, both of which scale
with the object size on *both* sides. The tenured-proportional arm of the cap is
invariant for the same reason: `tenured_bytes / 2` is `tenured_objects / 2`
objects. Only the constant band converts a fixed byte budget into a variable
object count, so only the constant band is re-denominated here.

## The measurement that unblocked it

#7929's recorded blocker was that an object term fires the collector earlier and
every extra cycle pays a fixed root scan. Both sides are now priced, on the shipped
corpus, by sweeping the existing `PERRY_GC_SCAVENGE_NURSERY_MB` dial (instructions
retired; the dev box ran at load 77–92, so no wall clock is quoted):

* **Marginal cost per object moved: ~1050 instructions.** Three `deeplist` cells
  with an identical cycle-kind profile (2 minors, 0 fulls) — 805 137 objects at
  1203 M, 954 929 at 1366 M, 1 254 513 at 1669 M — give slopes of 1088 and 1011
  instr/object.
* **Fixed cost per extra copying minor: ~0.5 M instructions.** `churn` at caps
  8/16/32 MB runs 172/86/43 minors for 3205.7/3093.3/3078.0 M instructions;
  netting out the object term leaves ~0.49 M per cycle. That is consistent with
  #7915's own number (218 455 pointer roots at ~2 instructions each).

So an extra minor pays for itself if it avoids ≥ ~500 object-moves, and on the
taxed programs the trade is ~500:1 in favour. On `cycles` and `pipeline` more
cycles is outright *cheaper* (1763.6 M at 23 minors vs 1837.6 M at 6).

## One-sided by design

The factor is clamped at 1.0, so a representation *larger* than the reference gets
today's band rather than a proportionally larger one. Two measured reasons:

* The byte-weighted mean is not a representative object size when survivors are
  dominated by large allocations — `push_num`'s mean is **3600 B** because it
  survives arrays. That was the documented blocker on #7929; one-sided clamping
  turns a ×50 cap explosion into exactly today's behaviour.
* Raising a cap is the dangerous direction: it is how a program crosses
  `GC_OLD_GEN_RECLAIM_THRESHOLD_BYTES` (48 MB) or lands in a #7909 budgeted stall.

Every program at or above the reference is left bit-identical, which is also what
makes the A/B below self-calibrating.

## A/B — per program, both arms, cycle kinds included

`origin/main` @ `ac52a5c38` vs this branch, same target dir, same `-p` set,
`PERRY_RUNTIME_DIR` pinned, `.a` mtimes verified to have moved. All 19 programs
**byte-identical to the base arm and to `node --experimental-strip-types` 26.5.1,
exit 0**.

| program | minors b→f | fulls b→f | objects moved | instructions | peak RSS |
|---|---|---|--:|--:|--:|
| **deeplist** | 3→3 | **1→0** | −11.8% | **−72.3%** | **−27.2%** |
| **retain** | 6→5 | **2→1** | −24.5% | **−62.9%** | **−26.4%** |
| **retain1** | 3→3 | 1→1 | −11.8% | **−11.3%** | **−6.5%** |
| churn / churn_alloc / push_cls | 86→88 | 0→0 | −7.2% | −0.9…−1.1% | ±0.3% |
| cycles | 12→22 | 0→0 | 0 | −0.8% | +0.1% |
| pipeline | 6→7 | 0→0 | +1.6% | −0.9% | −0.3% |
| tree | 40→40 | 0→0 | 0 | −0.2% | −0.1% |
| tree_wide | 40→40 | 0→0 | 0 | −0.1% | 0.0% |
| shapes | 1→1 | 0→0 | 0 | −3.2% | 0.0% |
| asyncpipe | 1→1 | 0→0 | 0 | −2.5% | +0.3% |
| interp | 31→44 | 0→0 | +3.7% | +0.2% | +0.1% |
| iso_miss | 43→61 | 0→0 | +7.4% | +0.4% | −0.1% |
| *push_num* | 18→18 | 0→0 | 0 | −2.1% | +0.2% |
| *retain_wide* | 7→7 | 2→2 | 0 | −0.3% | 0.0% |
| *retain_wide1* | 4→4 | 1→1 | 0 | −0.8% | −0.1% |
| *churn_read* | 0→0 | 0→0 | 0 | −3.5% | −0.9% |
| *fib40* | 0→0 | 0→0 | 0 | −0.3% | +1.4% |

*Italic rows are the control set*: their factor is exactly 1.000 (clamped, or no
copying minor ever ran), so their policy is bit-identical and their spread —
**−0.3% to −3.5%** — is this A/B's link-layout noise floor, not an effect.
`interp` +0.2% and `iso_miss` +0.4% sit inside it; the three large deltas do not.
`retain1` is the clean cell: identical cycle kinds in both arms, −11.8% object
work for −11.3% instructions, against the +12.0% instructions #7928 cost it.

`deeplist`'s and `retain`'s larger wins are a threshold crossing on top of the
object effect: both were running an `old_gen_bytes`-triggered `full` whose
`sweep.freed_bytes` is **0** — a futile mark-sweep over a heap that is live by
construction — and the smaller band keeps old-gen under the 48 MB one-shot arm of
`old_reclaim_pressure_due`. Isolating that (interpolating `deeplist`'s 3-minor
cells) leaves the object-denomination effect alone at **−10.5%**, which is exactly
the +10.5% regression #7928 caused on that program.

## Gate

Four tests. The two pure-function tests still pass with the `copying.rs` call site
deleted — the #7024 shape — so the wiring is covered separately by a band-level
test and by `copying_minor_feeds_the_object_denomination_census`, which drives a
real copying minor.

★ The discriminating quantity in that test is **"the recorded mean equals THIS
cycle's measured mean AND differs from the seed"**. A "the mean is nonzero"
assertion is satisfied by the seed itself, so an unwired build would pass it — a
presence check, not a proof. The test also refuses to run vacuously: it asserts
the fixture's mean is strictly below the reference, so a future representation
change that made the fixture 72 B fails the test rather than silently exercising
the clamp instead of the scaling arm.

Both sabotage arms were **run**, not argued:

| sabotage | result |
|---|---|
| delete the `note_surviving_object_census` call in `copying.rs` | end-to-end test FAILS (`the census must carry THIS cycle's measured mean (19040 B over 340 objects)`) |
| delete the factor from `influx_driven_nursery_cap_bytes` | band-level test **and** end-to-end test FAIL |

`constant_band_buys_a_constant_object_count` additionally carries an inline
sabotage arm: it asserts the *un*-denominated band inflates the object budget by
≥1250 permille, so if that ever reads 1000 the test declares its own main
assertion meaningless instead of passing quietly.

## Validation

* Corpus **19/19 byte-exact vs node 26.5.1, exit 0**, in the ordinary run and
  again under `PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800
  PERRY_GC_VERIFY_EVACUATION=1`.
* `iso_miss` canary: `checksum 437840 misses 0`.
* `cargo test --release -p perry-runtime`: **2216 passed, 0 failed**.
* `cargo fmt --all -- --check`, `scripts/check_file_size.sh`,
  `scripts/check_gc_doc_claims.py` (17 facts re-derived),
  `scripts/gc_runtime_root_holders.py`: all clean.
* No new env knob (CLAUDE.md GC knob kill-policy). Diagnosis rides the existing
  `PERRY_GC_DIAG` path. Nothing is added to the allocation fast path — the mean
  comes from the census the collector already produces.

Full working: `gc-handoff/BUDGET-NOTES.md`.

## Summary by CodeRabbit

* **New Features**
  * Nursery capacity now adapts to the average size of objects moved during garbage collection.
  * Capacity adjustments preserve existing tenuring behavior while improving consistency of object budgets.
  * Added diagnostics for measured object sizes and adaptive capacity scaling.

* **Bug Fixes**
  * Empty collection cycles now retain the previous valid measurement.

* **Tests**
  * Added coverage for scaling limits, measurement carry-forward, proportional capacity behavior, and end-to-end updates.

* **Documentation**
  * Added changelog details, safety rationale, and benchmark results.

### DIFF
--- a/changelog.d/7961-gc-object-denominated-nursery-band.md
+++ b/changelog.d/7961-gc-object-denominated-nursery-band.md
@@ -1,0 +1,52 @@
+**GC: the nursery constant band is denominated in objects, not bytes (#7929).**
+
+The scavenge nursery trigger compared from-space *bytes* against a constant
+16 MB band while the copying minor's cost is per *object*, so shrinking a
+representation silently bought the collector more work per cycle: #7928 took a
+two-field object literal 72 B → 56 B and every minor then moved **1.286×
+(= 72/56)** as many objects for the same bytes, costing `deeplist` +10.5% and
+`retain1` +8.8%. The band is now scaled by the mean size of the objects the last
+copying minor actually moved, so it buys a constant object budget. The mean comes
+from the census the collector already produces — nothing is added to the
+allocation fast path, and the arena still needs no nursery object counter.
+
+Only the constant band is re-denominated. The two `tenuring.rs` ratios
+(`desired_survivor_bytes` = cap/16, `retune_nursery_cap_scale`'s cap/25 and
+cap/100 bands) and the tenured-proportional arm of the cap are
+representation-invariant by cancellation — `tenured_bytes / 2` is
+`tenured_objects / 2` objects — so scaling them too would double-count.
+
+The scaling is **one-sided** (clamped at 1.0): a mean above the 72 B calibration
+reference keeps today's band. That neutralises an array-dominated mean —
+`push_num` measures 3600 B because it survives arrays, the documented blocker on
+the issue — and leaves every program at or above the reference bit-identical,
+because *raising* a cap is the direction that pushes a program across
+`GC_OLD_GEN_RECLAIM_THRESHOLD_BYTES` or into a #7909 budgeted stall.
+
+Measured on the shipped 19-program corpus against `origin/main`, all 19
+byte-exact vs `node --experimental-strip-types` 26.5.1 and exit 0 (also under
+`PERRY_GC_PROTECT_FROMSPACE=1` + `PERRY_GC_VERIFY_EVACUATION=1`), instructions
+retired: `deeplist` −72.3% / RSS −27.2%, `retain` −62.9% / RSS −26.4%, `retain1`
+−11.3% / RSS −6.5%, `churn` family −1.0%. The five programs whose factor is
+exactly 1.000 form a built-in control set; their −0.3%…−3.5% spread is this A/B's
+link-layout noise floor, and `interp` (+0.2%) and `iso_miss` (+0.4%) sit inside
+it. `retain1` is the clean cell — identical cycle kinds in both arms, −11.8%
+object work for −11.3% instructions, against the +12.0% #7928 cost it.
+
+The blocker recorded on the issue — "an object term fires the collector earlier
+and every extra cycle pays a fixed root scan" — is refuted by measurement: an
+extra copying minor costs **~0.5 M instructions** (consistent with #7915's 218 455
+pointer roots) while an object moved costs **~1050**, so an extra cycle pays for
+itself if it avoids ≥ ~500 object-moves. On the taxed programs the trade is
+~500:1 in favour, and on `cycles` and `pipeline` more cycles is outright cheaper.
+
+New coverage: a pure-function test that the band buys a representation-invariant
+*object budget* (with its own inline sabotage arm, so it declares itself
+meaningless rather than passing quietly if the un-denominated band ever stops
+differing), one-sided-clamp endpoints named after the failures they prevent, a
+carry-forward test for cycles that moved nothing, and an end-to-end test driving
+a real copying minor whose discriminating assertion is that the recorded mean
+equals *that cycle's* measured mean and differs from the seed — a "mean is
+nonzero" check would be satisfied by the seed itself. Both sabotage arms (delete
+the `copying.rs` call site; delete the factor from the band) were run and fail
+the intended tests.
--- a/crates/perry-runtime/src/gc/copying.rs
+++ b/crates/perry-runtime/src/gc/copying.rs
@@ -1856,6 +1856,21 @@ pub(super) fn gc_collect_minor_copying_fast_path_with_eligibility(
     // re-baseline sees post-collection live allocation rather than high-water.
     note_copying_minor_young_survival(collector.stats.young_survival_permille);
     maybe_schedule_old_reclaim_after_copied_minor();
+    // #7929: the object denomination of the nursery constant band, fed BEFORE
+    // the tenuring loop so every number `retune_after_scavenge` derives from
+    // the effective cap (desired survivor occupancy, the cap-scale band) reads
+    // one consistent factor. Both tenuring ratios are representation-invariant
+    // by cancellation, so this only re-denominates the constant band itself.
+    super::tenuring::note_surviving_object_census(
+        collector
+            .stats
+            .copied_bytes
+            .saturating_add(collector.stats.promoted_bytes),
+        collector
+            .stats
+            .copied_objects
+            .saturating_add(collector.stats.promoted_objects),
+    );
     retune_after_scavenge(
         collector.stats.eden_live_bytes,
         collector.stats.copied_bytes,
--- a/crates/perry-runtime/src/gc/tenuring.rs
+++ b/crates/perry-runtime/src/gc/tenuring.rs
@@ -222,9 +222,16 @@ pub(super) fn scavenge_nursery_cap_effective_bytes() -> usize {
 pub(super) fn influx_driven_nursery_cap_bytes() -> usize {
     let constant_band =
         gc_scavenge_nursery_cap_bytes().saturating_mul(NURSERY_CAP_SCALE.with(Cell::get) as usize);
-    constant_band.saturating_mul(nursery_cap_object_scale_permille(
-        mean_surviving_object_bytes(),
-    )) / 1000
+    // The multiply is done in u64 deliberately. `usize::saturating_mul` on an
+    // ILP32 target (watchOS/visionOS are 32-bit) would saturate a 64 MB band
+    // against a 1000-per-mille factor at `u32::MAX` and the following divide
+    // would then hand back ~4 MB — a silent 16x cap collapse on exactly the
+    // devices with the least headroom. A 64-bit intermediate cannot overflow:
+    // the band is bounded by 4 GB and the factor by 1000.
+    let scaled = constant_band as u64
+        * nursery_cap_object_scale_permille(mean_surviving_object_bytes()) as u64
+        / 1000;
+    scaled.min(usize::MAX as u64) as usize
 }
 
 /// #7929: the mean size of the objects the last copying minor actually moved,
--- a/crates/perry-runtime/src/gc/tenuring.rs
+++ b/crates/perry-runtime/src/gc/tenuring.rs
@@ -166,6 +166,11 @@ thread_local! {
     static NURSERY_CAP_SCALE: Cell<u8> = const { Cell::new(1) };
     static CAP_GROW_STREAK: Cell<u8> = const { Cell::new(0) };
     static CAP_SHRINK_STREAK: Cell<u8> = const { Cell::new(0) };
+    /// #7929: mean size of the objects the last copying minor moved. Seeded at
+    /// the calibration reference so a process with no completed copying minor
+    /// paces exactly as it did before the object denomination existed.
+    static MEAN_SURVIVING_OBJECT_BYTES: Cell<usize> =
+        const { Cell::new(NURSERY_CAP_REFERENCE_OBJECT_BYTES) };
 }
 
 /// The survivals threshold the next copying minor should promote at:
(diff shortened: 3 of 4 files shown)


proggeramlug:
## ⚠️ Post-merge correction: on a **healthy** collector the shipped term is net-NEGATIVE, and #7965 is a hard prerequisite

#7961 merged (`109de930f`) while its validation was still running. The validation
finished and **contradicts the A/B in the PR body**. Recording it here rather
than leaving the PR's numbers standing.

### What the PR measured, and why it was misleading

The PR's A/B was taken against `ac52a5c38`, which **contains #7965** (bisected in
that issue to #7944 / #7902 — untraced promotion stopped crediting the
old-reclaim baseline, so `deeplist`/`retain*` each gained a futile full
mark-sweep). Against that regressed base the term reads as a large win, because
it happens to keep old-gen under the 48 MB one-shot threshold.

Re-run against the **pre-regression** base (`e886b56dd`, the commit before
#7902), with the same two commits cherry-picked on top so the only variable is
the term itself:

| program | base fulls / instr | +term fulls / instr | Δobjects | Δinstr | ΔRSS |
|---|--:|--:|--:|--:|--:|
| deeplist | 0 / 1107.2 M | 0 / 1093.5 M | −11.8% | **−1.2%** | −0.0% |
| **retain1** | 0 / 1293.7 M | **1** / 3342.4 M | −11.8% | **+158.4%** | **+25.4%** |
| **retain** | 0 / 2596.4 M | **1** / 4635.8 M | −5.1% | **+78.5%** | +2.6% |
| retain_wide | 0 / 3029.3 M | 0 / 3027.7 M | 0 | −0.1% | 0.0% |

**The term itself introduces the futile full on `retain1` and `retain`.**

### Mechanism — the same defect as #7965, reached from the other side

`old_reclaim_pressure_due`'s absolute arm is a one-shot latch:

```rust
(old_in_use >= T && baseline < T) || ...     // T = 48 MB
```

`baseline` (`GC_LAST_OLD_RECLAIM_IN_USE_BYTES`) is a running **sum of promotion
deltas**, so whether it ends up above or below T is decided by **promotion step
size**, not by the heap. On `retain1`:

| | promotion steps | final baseline | latch |
|---|---|--:|---|
| base | 18.74 + 34.60 MB | **53.34 MB** > T | **disarmed by overshoot** |
| +term | 18.74 + 26.21 MB | **44.95 MB** < T | **still armed** |

The smaller band makes the second step smaller, the baseline lands 3 MB short of
T, and the latch survives. `retain1`'s `all` array — a large object born straight
into old-gen, whose growth never passes through promotion and so never credits
the baseline — then pushes `old_in_use` past 48 MB and the latch fires. The full
reports `freed_bytes = 8 257 632`, i.e. the array's dead backing stores: a
whole-heap collection to reclaim 8 MB.

`deeplist` has no such old-born array, which is why it is the one program in the
table that stays clean — and its **−1.2%** is therefore the honest value of the
object denomination on a healthy collector, not the −72.3% the PR quotes.

So #7965 and this are one defect: **a valve whose firing is decided by
granularity rather than by the quantity it claims to measure.** #7902 zeroed the
baseline; this term lowers it. Either way the latch is left armed and a
100 %-live old-gen gets a futile full.

### What I recommend

1. **Do not revert #7961 yet.** On *current* main both arms already carry the
   #7965 full, so the term is measured there as `retain1` −10.9%, `retain`
   −63.1%, `deeplist` −72.3% and is a net win *today*. Reverting now would make
   main worse.
2. **Fixing #7965 must re-measure this term in the same change.** The moment the
   baseline is credited correctly, `retain1`/`retain` flip from −11 %/−63 % to
   **+158 %/+78 %**. That is a landmine, and it is armed.
3. The durable fix is to stop the latch depending on step size at all — either
   credit old-gen growth to the baseline regardless of how it arrived, or refuse
   to schedule a reclaim when the last cycle measured `young_survival_permille =
   1000` (it was 1000 on every cycle in this table, i.e. the collector had just
   finished proving the reclaim cannot free anything).

### What stands from the PR

* The corpus census, and that only the constant band is mis-denominated (both
  `tenuring.rs` ratios and the tenured-proportional arm are invariant by
  cancellation).
* The one-sided clamp and its control set — `push_num`'s 3600 B array-dominated
  mean is neutralised, and the 5 factor-1.000 programs calibrate the A/B's
  link-layout noise floor at −0.3 %…−3.5 %.
* **The recorded blocker is still refuted**: an extra copying minor costs ~0.5 M
  instructions (`churn`: 43→172 minors for +4 % instructions and −60 % RSS), so
  "more cycles pay the fixed root scan" was never the reason not to do this.
* What is *not* established, and what the PR over-claimed: that the term is worth
  ~10 % on the taxed programs. On a healthy collector it is worth **~1 %**
  (`deeplist`, clean cell, 3 minors and 0 fulls in both arms). The ~10 % figures
  in the PR body and in `REPR-NOTES.md` §3d are entangled with cycle-kind
  changes and should not be quoted as per-object work.

Full working: `gc-handoff/BUDGET-NOTES.md` §8.

Issue #7916, opened 2026-08-12
perf: object representation is now the binding constraint on the retain cluster (72 bytes per 2-field literal; 216 MB written to store 48 MB)


proggeramlug:
## Summary

Established while decomposing the retain cluster's GC cost (`gc-handoff/RETAIN4-NOTES.md`): **the collector is no longer the binding constraint on these benchmarks — object representation is.**

**A 2-field object literal costs 72 bytes.** `retain` writes **216 MB to store 48 MB of numbers** — a 4.5x write amplification.

The decisive evidence is that the remaining gap survives a hypothetically free collector:

| program | wall | GC pause | pause frac | scriptc |
|---|--:|--:|--:|--:|
| retain | 159.5 ms | 52.0 | 33% | 109 |
| retain1 | 71.4 | 38.7 | 54% | 38 |
| retain_wide | 206.2 | 75.4 | 37% | 122 |
| retain_wide1 | 73.9 | 32.3 | 44% | 43 |
| shapes | 64.8 | 4.6 | **7%** | 38 |

`retain_wide` (206.2 − 75.4 = 130.8 ms of pure mutator) and `shapes` (60.2 ms of mutator) **still lose to scriptc with a zero-cost collector**. `retain` would need *exactly* zero GC to reach parity. So further collector work cannot close this cluster.

## Why this is the representation problem, not a GC problem

scriptc's advantage here is not its refcounting — it is that it stores these objects more compactly. 72 bytes for two fields means header + slot overhead dominates payload. The write amplification is then paid twice: once in allocation bandwidth, and again in everything downstream that must touch those bytes (promotion, copying, cache pressure).

This is the concrete, measured instance of the direction recorded in the representation-selection RFC: type-driven **representation selection** rather than uniform NaN-boxed slots for every field.

## Corollaries

- **`shapes` should leave the "high-survival GC" cluster.** At **7%** GC pause it is a mutator benchmark that was mis-grouped; GC levers will not move it.
- **Parallel GC is not indicated for this cluster** (independently confirmed): the per-object visits split roughly 50/50 into work already deleted by #7914 and ~9.6 ms of a 159 ms program. There is not enough collector time left to parallelise, which matches the standing directive that parallelism is a last resort.

## Acceptance

- A measured proposal for compacting at least the common shapes here (small fixed-field object literals with numeric fields).
- Report bytes-per-object and total write volume alongside any speedup — those are the primary metrics for this issue, not wall clock.

[...]

PR #7981, opened 2026-08-12
fix(thread): take the serialized class-parent edge from the registry, not the header word

proggeramlug:
## What

`ObjectHeader.parent_class_id` carries **two different things**. For a class instance it is the parent class id. For a plain object (`class_id == 0`) the same word is the #6759 C3c runtime **ShapeId stamp**, written lazily by every by-name resolve path (`typed_feedback.rs:774`, `get_field_by_name_tail.rs:1516`, `ic_miss.rs:713/756`).

`thread.rs::serialize_object` copied the raw word, and the worker-side `deserialize` hands it to `js_object_alloc_with_parent`, which does:

```rust
if parent_class_id != 0 { register_class(class_id, parent_class_id); }
```

So **any object literal that has been read once and then crosses a `spawn` / `parallelMap` boundary registers `class 0 → <a shape id>`** in the process-global class-parent registry (`PARENT_DENSE[0] = shape_id + 1`, `CLASS_REGISTRY[0] = shape_id`) and bumps the store-plan epoch, once per deserialized object. Every consumer I checked guards `class_id == 0` off (`instanceof.rs:909`, `descriptor_helpers.rs:231`), so there is no *known* live victim — but it is registry pollution reachable from ordinary user code, and it is the reason the header word could not be re-purposed.

The authoritative parent edge never lived in the header. Every parent-chain walk in the runtime reads `get_parent_class_id(class_id)` (`object/class_meta_registry.rs:82`), and each edge is registered from a compile-time constant:

* `js_register_class_parent`, emitted once per inheriting class in the module-init prelude (`codegen/string_pool.rs:478-505`) — precisely because codegen's inline `new C()` bump path writes the header word and *skips* the per-alloc `register_class` (`class_registry/parent_static.rs:22-41`);
* `register_class` inside every runtime allocator that takes a `parent_class_id` argument (`object/alloc.rs:125, 203, 250, 398`).

So the serializer now reads it from there.

## Why it matters beyond the fix

This removes the **last consumer of `ObjectHeader.parent_class_id` as inheritance data**. It was the blocking dependency for #6759 Phase C3's unification of class layouts and plain-object shapes into one shape-id space — which is in turn the prerequisite for #7916's 16-byte header shrink, because `class_field_inline_guard` can only replace its `keys_array`-identity compare with a one-word ShapeId compare once a class instance *has* a shape word.

## Tests

Three, each written to fail for a stated reason rather than to pass:

1. `thread_parent_class_id_tests::a_plain_objects_shape_stamp_is_not_serialized_as_a_parent_class_id` — asserts the stamp is genuinely present in the fixture **first** (so the test is not vacuous), then that it does not reach the wire.
2. `…::a_class_instances_parent_comes_from_the_registry_not_the_header` — the parent still round-trips with the header word deliberately overwritten by a stamp, so the fix is not "always send 0".
3. `object/delete_rest.rs::shape_transition_tests_6759` — pins both halves of the #6759 C3 entry gate: a plain object's `delete` mints a genuinely **different** ShapeId (with a vacuity assertion), while a class instance's `delete` compacts its slots leaving `class_id` **and** `parent_class_id` untouched. The second is written to go red when a future rung gives class instances a stamp — that is the intended signal to switch the guard.

## Scope note

Deliberately does **not** attempt the C3 unification itself. The investigation behind this (enumeration of every in-place key-set mutation, the sabotage A/B that shows which fixture shapes are vacuous as a guard gate, and a four-rung costed plan) is written up in `gc-handoff/SHAPE-NOTES.md`.

Refs #6759, #7916.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed cross-thread object serialization so plain-object shape metadata is not misinterpreted as class inheritance.
  - Preserved class inheritance, fields, methods, and `instanceof` behavior across worker boundaries.
  - Prevented incorrect registry entries caused by serialized plain objects.
  - Ensured property deletion and subsequent updates maintain correct object shape behavior.

- **Tests**
  - Added regression coverage for serialization, inheritance, worker execution, and shape transitions.

### DIFF
--- a/changelog.d/7981-thread-parent-class-id-from-registry.md
+++ b/changelog.d/7981-thread-parent-class-id-from-registry.md
@@ -1,0 +1,45 @@
+### `perry/thread` no longer replays a plain object's ShapeId stamp as a class-parent edge
+
+`ObjectHeader.parent_class_id` carries two different things: the parent class id
+for a class instance, and — since #6759 C3c — the runtime **ShapeId stamp** for a
+plain object (`class_id == 0`), written lazily by every by-name resolve path.
+
+`thread.rs::serialize_object` copied the raw word, and the worker-side
+deserializer hands it to `js_object_alloc_with_parent`, which does
+`if parent_class_id != 0 { register_class(class_id, parent_class_id) }`. So any
+object literal that had been read once and then crossed a `spawn` /
+`parallelMap` boundary registered `class 0 → <a shape id>` in the process-global
+class-parent registry (`PARENT_DENSE[0] = shape_id + 1`,
+`CLASS_REGISTRY[0] = shape_id`) and bumped the store-plan epoch, once per
+deserialized object. Every consumer checked guards `class_id == 0` off, so no
+live victim was found — but it is registry pollution reachable from ordinary
+user code.
+
+The authoritative parent edge never lived in the header. Every parent-chain walk
+reads `get_parent_class_id(class_id)`, and each edge is registered from a
+compile-time constant: by `js_register_class_parent` in the module-init prelude
+for codegen's inline `new C()` path (which writes the header word and
+deliberately skips the per-alloc `register_class`), and by `register_class`
+inside every runtime allocator that takes a `parent_class_id` argument. The
+serializer now reads it from there.
+
+This also removes the **last consumer of the header word as inheritance data** —
+the blocking dependency for #6759 Phase C3's unification of class layouts and
+plain-object shapes into one shape-id space, which is itself the prerequisite for
+#7916's header shrink (`class_field_inline_guard` can only trade its
+`keys_array`-identity compare for a one-word ShapeId compare once a class
+instance has a shape word).
+
+Three tests, each written to fail for a stated reason:
+
+- the stamp is asserted present in the fixture *before* asserting it does not
+  reach the wire, so the test cannot pass vacuously;
+- a class instance's parent still round-trips *from the registry* with the header
+  word deliberately overwritten by a stamp, so the fix is not "always send 0";
+- `object/delete_rest.rs::shape_transition_tests_6759` pins both halves of the C3
+  entry gate — a plain object's `delete` mints a genuinely different ShapeId,
+  while a class instance's `delete` compacts its slots leaving `class_id` **and**
+  `parent_class_id` untouched. The second goes red when a future rung gives class
+  instances a stamp, which is the intended signal to switch the guard.
+
+Refs #6759, #7916.
--- a/crates/perry-runtime/src/object/delete_rest.rs
+++ b/crates/perry-runtime/src/object/delete_rest.rs
@@ -602,6 +602,135 @@ pub extern "C" fn js_object_rest(
 }
 
 #[cfg(test)]
+mod shape_transition_tests_6759 {
+    //! #6759 C3: what a `delete` does to an object's SHAPE IDENTITY, pinned for
+    //! both object representations — because they differ, and the difference is
+    //! the entry gate for the header shrink (#7916).
+    //!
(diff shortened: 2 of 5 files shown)


proggeramlug:
### Validation

**Sabotage-verify (fix committed first).** Reverted `serialize_object` to the pre-fix line `let parent_class_id = (*obj).parent_class_id;`, rebuilt, re-ran:

```
delete_mints_a_fresh_shape_id_for_a_plain_object ......................... ok
delete_leaves_a_class_instance_with_no_shape_word_to_transition .......... ok
a_class_instances_parent_comes_from_the_registry_not_the_header ..... FAILED
a_plain_objects_shape_stamp_is_not_serialized_as_a_parent_class_id .. FAILED

assertion failed: a ShapeId (0x800000bf) reached the worker as a class-parent
  edge; deserialization would call register_class(0, 0x800000bf)
assertion failed: the serializer is still reading the header word, not the registry
```

The failure carries a **real observed id**, so the defect is measured rather than argued. The two `delete` tests stay green — the control that the sabotage was targeted.

**Clean tree.** `cargo test --release -p perry-runtime --lib` → **2243 passed, 0 failed, 4 ignored**. `cargo fmt --all -- --check`, `scripts/check_file_size.sh`, `scripts/gc_runtime_root_holders.py`, `scripts/addr_class_inventory.py`, `scripts/check_thread_locals.py` all clean. (`thread.rs` hit 2018 lines with the tests inline, so they live in `thread_parent_class_id_tests.rs` via `#[path]` — the `json_tape_tests.rs` pattern.)

**End to end.** `test-files/test_issue_7981_thread_shape_stamp_parent.ts`, exit 0:

```
stamped read: 2
plain mapped: alpha,beta,gamma=6 | alpha,beta,gamma=6 | alpha,beta,gamma=6 | alpha,beta,gamma=6
main chain: mid,4,40,true,true,true
worker chain count: 4 allMatch: true
spawn chain: mid,7,70,true,true,true match: true
instance fields: 7 70 mid
main again: true alpha,beta,gamma=6
```

### Two gates that would be VACUOUS here, and are deliberately not claimed

* **The 19-app GC corpus.** `grep -l "perry/thread\|parallelMap\|spawn("` over `gc-handoff/{bench,apps,fib}/*.ts` matches nothing — `serialize_object` is never called by any corpus program, so a green 19/19 would say only that the build works.
* **The gap suite** — no `test_gap_*` uses `perry/thread` (it has no Node equivalent), so it is silent on this change too.

The discriminating gate is the unit test, because the class-parent registry is not observable from JS: nothing in userland can read `get_parent_class_id(0)`. The `.ts` file above is the behavioural regression half only.

proggeramlug:
### Post-fix run (closes the one item outstanding when this was undrafted)

The final `-p perry -p perry-runtime-static -p perry-stdlib-static` build finished, so the `.ts` coverage has now been re-run against a binary linked with the **post-fix** `libperry_runtime.a`:

| check | result |
|---|---|
| `test-files/test_issue_7981_thread_shape_stamp_parent.ts` | exit 0, **byte-identical to the pre-fix arm** (`cmp`) |
| `test-files/test_issue_7769_thread_class_dispatch.ts` | exit 0, `allMatch: true`, `match: true` |
| `probe_delete_shape.ts` / `probe_delete_isolate_ka.ts` / `probe_shared_keys_leak.ts` | all byte-exact vs node 26.5.1 |

The byte-identity across the two `perry/thread` arms is the expected result rather than a weak one: the registry pollution has no JS-visible victim, which is precisely why the detector had to be a unit test (`get_parent_class_id(0)` is not reachable from JS) and why the `.ts` file is labelled the behavioural-regression half only.

PR #7997, opened 2026-08-12
fix(gc): the aarch64 fast stack-map walker misread every SVE-shaped prologue (#7984)

### LINKED ISSUE #7984: aarch64-linux: the fast stack-map walker and the unwinder disagree by 96 bytes on the same root (PERRY_STACKMAP_WALKER=verify)
## Summary

On **aarch64 Linux (ELF)**, `PERRY_STACKMAP_WALKER=verify` catches the fast
frame-pointer-chain walker and the unwinder resolving the *same* root to two
**different addresses**. The gate is working; the code is wrong.

This is the `ubuntu-24.04-arm` arm of `gc-native-roots` (#7970). It is a **root
misidentification on a supported target**, not a harness problem — unlike the
`macos-14` arm of the same run, which was a missing `PERRY_GC_DIAG=1` and is
fixed separately.

## Evidence

Run `31405431962`, job `native-roots-rs4gc (ubuntu-24.04-arm, aarch64, ELF)`,
sha `bf8af0e0f`, probe `01_nursery_churn`:

```
thread '<unnamed>' (20234) panicked at crates/perry-runtime/src/gc/roots/stack_maps.rs:507:5:
assertion `left == right` failed: PERRY_STACKMAP_WALKER=verify: fast walk visited 1 unique slots, unwinder visited 1
  left:  [281474742909688]
  right: [281474742909592]
```

Both walkers found **one** slot, so this is not a missed or invented frame. They
disagree on **where** that slot is:

```
281474742909688 - 281474742909592 = 96 bytes
```

`verify_visit` (`gc/roots/stack_maps.rs`) runs `fp_chain::visit` non-mutating,
then `unwind::visit` for the real visitation, and asserts the address sets are
equal. A 96-byte constant offset between them is the signature of a
frame-base/CFA disagreement, not of a parse failure.

## Why this matters more than a red CI arm

`fp_chain` is the **fast** walker — the one that runs when `verify` is off. If it
resolves a root 96 bytes away from the real slot, then on this target the
collector marks and rewrites the wrong stack word. That is precisely CLAUDE.md's
"root-store dominance" failure class: nothing is visibly wrong at collection
time, and it surfaces cycles later as `TypeError: value is not a function`.

`verify` mode exists to catch exactly this, and it did — the first time anyone
ran it end-to-end on aarch64 ELF. The workflow's own header predicted the shape:

> What no arm has ever exercised is the end-to-end chain on this target — real
> LLVM asm output, real ELF linking, real runtime walking — where the failure
> mode is not a parse error but a wrong answer.

## Reproduction

```bash
# on aarch64 Linux
export RUSTFLAGS="-C force-frame-pointers=yes -C force-unwind-tables=yes"
cargo build --profile perry-dev -p perry -p perry-runtime-static -p perry-stdlib-static
PERRY_RS4GC=1 ./target/perry-dev/perry benchmarks/gc_ratchet/probes/01_nursery_churn.ts \
  -o /tmp/rs4gc-01
PERRY_GC_TRACE=1 PERRY_GC_DIAG=1 PERRY_STACKMAP_WALKER=verify \
PERRY_RS4GC=1 PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1 \
  /tmp/rs4gc-01
```

x86-64 ELF (`ubuntu-latest`) passes the same step, so this is aarch64-specific.

## Open questions for whoever picks this up

1. **Which walker is right?** The unwinder is the reference implementation, so
   the presumption is `fp_chain` is 96 bytes off — but that should be confirmed
   against the emitted stack map rather than assumed.
2. **Is 96 constant?** One probe, one slot is a thin sample. Worth running the
   whole probe matrix under `verify` on this target and checking whether the
   delta is always 96 (a fixed frame-layout term) or varies (an offset-scaling
   bug, e.g. the `.word`-vs-`.long` field-width hazard `word_width_for` warns
   about, which is aarch64-ELF-specific).
3. **Does the fast walker run in production on this target?** If yes, this is a
   live correctness bug on aarch64 Linux, not only a CI failure.

Found while repairing #7970. Sibling: the macOS arm of the same run was a gate
defect (missing `PERRY_GC_DIAG=1`), and the Windows arm is a separate link
failure.

proggeramlug:
Fixes #7984.

## Which walker is wrong, and why

`PERRY_STACKMAP_WALKER=verify` caught the frame-pointer-chain walker and the Itanium unwinder resolving the same GC root 96 bytes apart on `ubuntu-24.04-arm`. **The fp-chain walker is the wrong one** — which is the bad case, because it is the walker that runs when `verify` is off.

Reproduced end to end on aarch64 Linux (Ubuntu 24.04 arm64, LLVM 22.1.8 from apt.llvm.org, the issue's own RUSTFLAGS). The trigger is the **tuning**, not the distro. Perry builds a host binary with `-mcpu=native`; on a Neoverse-class core that enables SVE, and SVE changes the prologue LLVM emits:

```
PERRY_TARGET_CPU=neoverse-n2   -> diverges
PERRY_TARGET_CPU=neoverse-n1   -> passes
```

No Apple arm can see it, and no x86-64 arm either. Measured on `main` — the module body, which carries 100 stack-map records — built `-mcpu=neoverse-n2`, read with `objdump -d`:

```
124790: fc180fea  str   d10, [sp, #-128]!
124794: 6d0123e9  stp   d9, d8, [sp, #16]
124798: a9027bfd  stp   x29, x30, [sp, #32]
12479c: 910083fd  add   x29, sp, #0x20      <- fp established; decoder reads 32
1247a0: a9036ffc  stp   x28, x27, [sp, #48] <- NOT a `sub sp`: the run ENDED here
1247a4: a90467fa  stp   x26, x25, [sp, #64]
1247a8: a9055ff8  stp   x24, x23, [sp, #80]
1247ac: a90657f6  stp   x22, x21, [sp, #96]
1247b0: a9074ff4  stp   x20, x19, [sp, #112]
1247b4: d10143ff  sub   sp, sp, #0x50       <- 80 bytes, DROPPED
1247b8: 043f57df  addvl sp, sp, #-2         <- 2 x VL more, DROPPED
```

Two independent defects in `fp_to_sp_offset`:

1. **A callee-save store ended the accumulation run.** It does not move sp, so it says nothing about whether the prologue's stack adjustments are finished — but the rule was "the first instruction that is not a `sub sp` ends the run". LLVM interleaves those stores with the frame-pointer setup whenever SVE is on.
2. **`addvl sp, sp, #-N` was not decoded at all.** Its multiplier is in the instruction; its unit is the runtime vector length, which is not.

## The arithmetic proof, not an appeal to authority

From the report this PR's first commit added, under `qemu -cpu max` (VL = 64 B):

```
same slot count, so this is a base disagreement, not a missed frame;
  fp-chain minus unwinder = [208, 208] byte(s)
frames visited: fp-chain 9, unwinder 10; records matched: fp-chain 1, unwinder 1
  slot 0x400000800af0 = base 0x400000800ab0 +64 | ip ... (fn 0xaaaaaabc4790 + 0x22c)
    fp_to_sp_offset(fn) = Some(32), prologue words: fc180fea 6d0123e9 a9027bfd
      910083fd a9036ffc a90467fa a9055ff8 a90657f6 a9074ff4 d10143ff
```

Same function, same ip, same record, same offsets — so it is a base disagreement, between `0x…ab0` (fast) and `0x…9e0` (unwinder):

```
caller_fp        = fast base + decoded = 0x…ab0 + 32 = 0x…ad0
true x29-body_sp = 32 + 0x50 + 2*64                  = 240
true body_sp     = 0x…ad0 - 240                      = 0x…9e0   <- the unwinder's base
```

The unwinder's base is the frame's real body SP, derived independently from the prologue the fast walker misread.

**96 was never a constant** — it is that frame's missed tail, and it scales with the vector length: `0x50 + 1 x 16` on the runner, `0x50 + 2 x 64` under qemu. This answers open question 2 in the issue, and a fix keyed on 96 would have been wrong on every other vector length.

## The fix

Stores through sp with no writeback are transparent to the accumulation run, enumerated by opcode so an unrecognised instruction still ends it. `addvl`/`addpl` writing sp is decoded: multiplier from `imm6`, unit from `prctl(PR_SVE_GET_VL)` — a syscall, not `rdvl`, which faults on a core without SVE — cached for the process. Where the length cannot be read (every core without SVE, all Apple ones included) the whole decode fails and the frame goes to the unwinder.

Failing closed *only* was the first cut, and it was not enough: on an SVE host the frame in question is the module body, so the fast walker would decline at every collection, `verify` turns a decline into a panic, and the arm that found the bug would have stayed red on its own fix.

## Also in here

- **`verify` now names the disagreement.** Both walkers report a `ResolvedRoot` (address + the ip it was matched on + the record's function + the map's base register and offset + the resolved base); `verify` prints all of it, calls an equal-slot-count disagreement a *base* disagreement, prints each walk's frame and record counts, and dumps the prologue it decoded. That report is what identified this in one run. The prologue dump is gated on the map vouching for the address — the first draft was not, and a synthetic address turned the diagnostic into a SIGSEGV with no output.
- **The two walkers have unit coverage at all now.** Nothing `cargo test` runs had ever called `fp_chain::visit` or `unwind::visit`. `stack_maps_walker_agreement.rs` drives both over `global_asm!` frames in the two real layouts and requires each to land on a word *holding a sentinel*, because set equality is satisfied by two empty sets and by two identically-wrong ones. Plus a sabotage arm and a fallback arm.
- `gc-native-roots.yml` runs those tests on both aarch64 arms, requiring each by name, and no longer truncates the report to 20 lines.

## Measured

**aarch64 Linux, emulated SVE** (`qemu-aarch64 -cpu max`), two vector lengths so the scaling is pinned rather than one machine's answer: **28 `verify` runs (14 probes x VL 16 B and 64 B), all byte-exact** against the pinned Node oracle, no failures. Before the fix the same binary aborts with the report above.

The fast walker is **live** there, which is the point of decoding rather than declining — `11_collect_at_depth` at VL 64 B: `fp_walks 12, fallback_walks 0, records_matched 1338, locations_visited 2678`, `verify` green, i.e. 2678 slots cross-checked against the unwinder and agreed.

**macOS aarch64**, release build from a clean tree (`.a` mtimes moved; `PERRY_RUNTIME_DIR` pinned): 19/19 app corpus byte-exact and exit 0; the same seven apps under `PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800 PERRY_GC_VERIFY_EVACUATION=1 PERRY_GC_FORCE_EVACUATE=1` with the instrument live (`bytes_protected=18874368`, `retired_set=#N`); 42 probe runs across `fast`/`verify`/`unwind` all byte-exact. `cargo test -p perry-runtime --lib gc::roots::stack_maps` — 35 pass.

**Sabotage-verified**, gate committed first: `fp_offset = Some(offset + 96)` makes all three walker-agreement tests fail with `-96 bytes out`, restored by **rebuilding** (not `git checkout`), 35/35 green again. A first attempt (`offset - 96`) was caught by the walk's underflow bail-out instead of the sentinel check — recorded because that arm proves a different thing and would have made the sentinel look verified when it was not.

Full working notes, including the three mechanisms this is *not*: `gc-handoff/WALKER-NOTES.md`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved garbage-collection root tracking on ARM64 systems using SVE.
  - Prevented stack scanning errors caused by complex frame layouts and vector-length-dependent stack adjustments.
  - Added a safe fallback to platform unwinding when fast stack scanning cannot reliably interpret a frame.

- **Diagnostics**
  - Enhanced walker disagreement reports with more detailed root, frame, and traversal information.
  - Expanded captured failure logs to provide additional context for troubleshooting.

- **Tests**
  - Added coverage across ARM64 ELF and Mach-O environments, including fallback and malformed-frame scenarios.

### DIFF
--- a/.github/workflows/gc-native-roots.yml
+++ b/.github/workflows/gc-native-roots.yml
@@ -119,11 +119,24 @@
 #                 objects copied, and the whole step passes. The collector was
 #                 never the problem.
 #
-#   ubuntu-24.04-arm  REAL DEFECT, filed as #7984. `PERRY_STACKMAP_WALKER=verify`
-#                 caught the fast fp-chain walker and the unwinder resolving the
-#                 same root to addresses 96 bytes apart. This arm is red because
-#                 it found the bug it was built to find; it must STAY red until
-#                 #7984 is fixed. Do not skip it.
+#   ubuntu-24.04-arm  REAL DEFECT, filed as #7984 and FIXED (#7997).
+#                 `PERRY_STACKMAP_WALKER=verify` caught the fast fp-chain
+#                 walker and the unwinder resolving the same root 96 bytes
+#                 apart, and the fast walker — the one that runs when `verify`
+#                 is off — was the wrong one. Two blind spots in
+#                 `fp_to_sp_offset`, both only reachable with SVE on, which is
+#                 what `-mcpu=native` turns on for a Neoverse-class core and
+#                 nothing on macOS or x86-64 ever does: a callee-save store
+#                 ended the prologue's stack-adjustment run, and
+#                 `addvl sp, sp, #-N` was not decoded at all. The 96 was never
+#                 a constant — it is that frame's missed tail, and it scales
+#                 with the vector length (208 at VL = 64 B).
+#
+#                 This arm should now be GREEN. If it goes red again, read the
+#                 report `verify` prints before assuming a regression: it names
+#                 the frame, the base register, both resolved bases and the
+#                 prologue words, which is enough to say which walker is wrong
+#                 without a second run.
 #
 #   windows-latest    REAL DEFECT, filed as #7985 (`perry.exe` cannot link
 #                 against the official LLVM 22 release: /MT-vs-/MD CRT mismatch,
@@ -133,11 +146,11 @@
 #                 `D:\a\_temp` as a remote host — was a workflow bug and is
 #                 fixed here with `--force-local`.
 #
-# So: after #7970 the macOS and ubuntu-latest arms should pass and the other two
-# should remain red on their filed defects. This workflow is therefore NOT a
-# promotion candidate yet — promoting it while #7984/#7985 are open would block
-# every PR. Promote only once all four arms are green, and per CLAUDE.md, run it
-# green once BEFORE adding it to branch protection.
+# So: after #7970 and #7997 three of the four arms should pass and only
+# windows-latest should remain red, on #7985. This workflow is therefore still
+# NOT a promotion candidate — promoting it while #7985 is open would block every
+# PR. Promote only once all four arms are green, and per CLAUDE.md, run it green
+# once BEFORE adding it to branch protection.
 name: gc-native-roots
 on:
   # Must run where it can actually gate something. Branch-scoped triggers were
--- a/.github/workflows/gc-native-roots.yml
+++ b/.github/workflows/gc-native-roots.yml
@@ -262,6 +262,42 @@ jobs:
         run: |
           export RUSTFLAGS="-C force-frame-pointers=yes -C force-unwind-tables=yes"
           cargo build --profile perry-dev -p perry -p perry-runtime-static -p perry-stdlib-static
+
+      # The two aarch64 walkers, over a frame this repository wrote, on the host
+      # that has to walk it.
+      #
+      # Until this step, NOTHING `cargo test` runs ever called `fp_chain::visit`
+      # or `unwind::visit` — their only exercise anywhere was `verify` mode in
+      # the step further down, inside a workflow that has never had a successful
+      # run on any branch (#7970). The unit tests cover the decoder and the
+      # matcher; the step that turns a `(register, offset)` pair into a stack
+      # address had none, which is why #7984 could only ever be found by
+      # compiling TypeScript and collecting.
+      #
+      # This asks the same question in seconds and without a compiler: given a
+      # frame whose layout is known — the aarch64-ELF one, where LLVM puts the
+      # `x29,x30` pair below the other callee-saves, and the Mach-O one, where
+      # it sits at the top — do both walkers land on the word the record names?
+      # It runs BEFORE the probe matrix so a walker defect is reported as a
+      # walker defect rather than as an oracle diff twenty minutes later.
+      - name: Walker agreement (aarch64 hosts)
+        if: ${{ !cancelled() && matrix.arch == 'aarch64' && runner.os != 'Windows' }}
+        run: |
+          set -euo pipefail
+          export RUSTFLAGS="-C force-frame-pointers=yes -C force-unwind-tables=yes"
+          cargo test --profile perry-dev -p perry-runtime --lib gc::roots::stack_maps \
+            -- --test-threads=1 | tee /tmp/walker-agreement.log
+          # `--lib <filter>` is a substring match, so a rename makes it select
+          # nothing and `cargo test` still exits 0 having run zero tests — a
+          # gate that cannot fail. Require each test by name, including the
+          # sabotage arm that proves the sentinel check discriminates.
+          for name in both_walkers_resolve_an_sp_root_in_an_elf_shaped_frame \
+                      both_walkers_resolve_an_sp_root_in_a_darwin_shaped_frame \
+                      a_wrong_frame_offset_is_caught; do
+            grep -q "walker_agreement::$name ... ok" /tmp/walker-agreement.log \
+              || { echo "::error::$name did not run — the filter matched nothing"; exit 1; }
+          done
+
       - name: Probe matrix, RS4GC mode, forced evacuation
         if: ${{ !cancelled() }}
         run: |
--- a/.github/workflows/gc-native-roots.yml
+++ b/.github/workflows/gc-native-roots.yml
@@ -515,7 +515,7 @@ jobs:
                 "/tmp/rs4gc-$name" > "/tmp/walker-$name-$mode.out" \
                                   2> "/tmp/walker-$name-$mode.err" \
                 || { echo "::error::$name crashed under PERRY_STACKMAP_WALKER=$mode"; \
-                     tail -20 "/tmp/walker-$name-$mode.err"; exit 1; }
+                     tail -120 "/tmp/walker-$name-$mode.err"; exit 1; }
               diff "/tmp/rs4gc-$name.oracle" "/tmp/walker-$name-$mode.out" \
                 || { echo "::error::$name diverged from the pinned oracle under PERRY_STACKMAP_WALKER=$mode"; exit 1; }
               checked=$((checked+1))
--- a/changelog.d/7997-verify-walker-report.md
+++ b/changelog.d/7997-verify-walker-report.md
@@ -1,37 +1,90 @@
+### aarch64: the fast GC stack-map walker misread every SVE-shaped prologue (#7984)
+
+`PERRY_STACKMAP_WALKER=verify` caught the frame-pointer-chain walker and the
+Itanium unwinder resolving the same GC root 96 bytes apart on
+`ubuntu-24.04-arm`. The fp-chain walker is the wrong one — it is also the one
+that runs when `verify` is off — and it was blind twice in the same prologue.
+
+Reproduced end to end on aarch64 Linux. The trigger is the tuning, not the
+distro: Perry builds a host binary with `-mcpu=native`, and on a Neoverse-class
+core that turns SVE on, which changes the shape LLVM emits. The same probe built
+`-mcpu=neoverse-n1` passes; no Apple arm can see it at all. Measured on `main`
+— the module body, 100 stack-map records — built `-mcpu=neoverse-n2`:
+
+```
+12479c: add   x29, sp, #0x20      <- fp established; the decoder read 32
+1247a0: stp   x28, x27, [sp, #48] <- not a `sub sp`, so the run ended HERE
+...     four more callee-save pairs
+1247b4: sub   sp, sp, #0x50       <- 80 bytes, dropped
+1247b8: addvl sp, sp, #-2         <- and two vector lengths more, dropped
+```
+
+1. **A callee-save store ended the accumulation run.** It does not move sp, so
+   it says nothing about whether the prologue's stack adjustments are finished,
+   but the rule was "the first instruction that is not a `sub sp` ends the run".
+   Stores through sp with no writeback are now transparent to it, enumerated by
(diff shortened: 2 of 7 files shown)


proggeramlug:
## Validation, macOS aarch64, release build from the clean tree

The release artifacts were rebuilt after the sabotage cycle below, not restored with `git checkout` — `.a` mtimes moved (`libperry_runtime.a` 23:57, `libperry_stdlib.a`/`perry` 00:00) and `PERRY_RUNTIME_DIR` was pinned to that directory for every run.

**19-app corpus** (`gc-handoff/m0810/sources.txt`), byte-exact against the pinned expected output, exit 0 on all: `churn churn_alloc churn_read push_num push_cls cycles deeplist tree tree_wide retain retain1 retain_wide retain_wide1 fib40 interp iso_miss asyncpipe shapes pipeline` — **19/19 ok, 0 fail**.

**Both protection knobs**, `PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800 PERRY_GC_VERIFY_EVACUATION=1 PERRY_GC_FORCE_EVACUATE=1`, seven of those apps — all exit 0, all byte-exact, and the instrument is live rather than assumed:

```
[gc-fromspace-protect] mode=ProtectPages retired_set=#0 blocks=18 sets_held=1/800 bytes_protected=18874368 ...
```

**The subject of this PR was live**, which matters more than the corpus passing: this change sits in the native-stack-map walkers, and the corpus only exercises it if those walkers run. Aggregated from `churn`'s 88 collections:

```
native_stack_maps: walks 88, fp_walks 88, fallback_walks 0,
                   frames_visited 528, records_matched 176, locations_visited 264
```

264 root slots were resolved through the refactored path, all by the fast walker, with zero unwinder fallbacks.

**gc_ratchet probe matrix, all three walker modes.** 14/14 probes compiled under `PERRY_RS4GC=1` and run under `fast`, `verify` and `unwind` with `PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1 PERRY_GC_HEAP_LIMIT=8 PERRY_GC_INCREMENTAL=0 PERRY_CONSERVATIVE_STACK_SCAN=off` — 42 runs, every one byte-exact against `node --expose-gc --experimental-strip-types`. No failures.

`verify` on `11_collect_at_depth` cross-checked **2678 root slots across 1338 matched records in 1382 frames**, 12 walks, 0 fallback walks — so the mode this PR rewrites is not vacuous on this host; the two walkers genuinely agree here, and the disagreement is aarch64-ELF-only.

**Unit tests**: `cargo test -p perry-runtime --lib gc::roots::stack_maps` — 29 pass.

## Sabotage verification

The gate was committed first, then `fp_to_sp_offset` was given #7984's exact shape (`fp_offset = Some(offset + 96)`, so every SP base in the frame lands 96 bytes low) and the tree **rebuilt**:

```
walker_agreement::a_wrong_frame_offset_is_caught ... FAILED
walker_agreement::both_walkers_resolve_an_sp_root_in_a_darwin_shaped_frame ... FAILED
walker_agreement::both_walkers_resolve_an_sp_root_in_an_elf_shaped_frame ... FAILED
```

with the message the report is supposed to produce:

```
the fp-chain walker placed the root at 0x16dc0a408, not 0x16dc0a468 (-96 bytes out).
The probe frame is `x29 - body_sp = 96`; a constant miss is a frame-base error,
which is #7984's shape.
```

Restored by rebuilding from the clean tree (`grep -c SABOTAGE` = 0), 29/29 green again.

A first attempt at the sabotage (`offset - 96`) was caught too, but by the fp-chain walk's underflow bail-out rather than by the sentinel check — worth recording, because that arm proves a *different* thing and would have made the sentinel look verified when it was not.

proggeramlug:
## Final validation, with the fix in

**aarch64 Linux, emulated SVE** — the configuration the GitHub ARM runner produces (`-mcpu=native` → Neoverse → SVE), run under `qemu-aarch64 -cpu max` at **two vector lengths** so the scaling is pinned rather than one machine's answer:

```
28 verify runs  (14 probes x VL 16 B and VL 64 B)   all byte-exact, 0 failures
```

Before the fix the same binary aborts with the report in the PR body; after it, every probe passes and the output matches the pinned Node oracle byte for byte.

The fast walker is **live** there, which is why the fix decodes the vector length rather than only declining — `11_collect_at_depth` at VL 64 B:

```
native_stack_maps: walks 12, fp_walks 12, fallback_walks 0,
                   frames_visited 1450, records_matched 1338, locations_visited 2678
```

`verify` green with `fp_walks 12` and `fallback_walks 0` means all 2678 slots were resolved by the fast walker and cross-checked against the unwinder, and agreed. A version that only failed closed would report `fp_walks 0` here and trip this workflow's own liveness assert.

**macOS aarch64**, release build from the clean tree after the fix (`libperry_runtime.a` 01:12, `perry` 01:15; `PERRY_RUNTIME_DIR` pinned to it):

- 19/19 app corpus byte-exact, exit 0.
- 42 probe runs — 14 probes x `fast` / `verify` / `unwind` — all byte-exact against the pinned Node oracle, 0 failures.
- Both protection knobs on seven apps (`PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800 PERRY_GC_VERIFY_EVACUATION=1 PERRY_GC_FORCE_EVACUATE=1`): all exit 0, all byte-exact, instrument live (88/88/40/17/45/1/7 `[gc-fromspace-protect]` lines).
- `cargo test -p perry-runtime --lib gc::roots::stack_maps` — 35 pass.
- `check_file_size.sh`, `gc_runtime_root_holders.py`, `addr_class_inventory.py`, `cargo fmt --check`, `cargo clippy -p perry-runtime` — all clean.

## What I could not close

- **The real `ubuntu-24.04-arm` runner has not confirmed this**, because `gc-native-roots` on this PR has been queued behind a saturated runner pool for the whole session (29 queued / 4 pending across the repo when last checked) and never started. Everything above is a local aarch64-Linux container plus qemu SVE emulation. The mechanism is not in doubt — the failing prologue was read out of a real aarch64-Linux binary this compiler produced, and the delta is arithmetic — but the runner's own vector length (16 B, giving the reported 96) is inferred from the number in the issue rather than measured.
- **A perf number for the SVE path.** The `prctl` is behind a `OnceLock` and the added opcode checks are two masked compares per prologue word, but I have not benchmarked collection time on an SVE host and cannot on this hardware.

Issue #7982, opened 2026-08-12
PERRY_LLVM_INPROCESS=native cannot build RS4GC `ptr addrspace(1)` roots; the unit corpora froze 151 codegen commits ago


proggeramlug:
## Summary

The in-process LLVM backend's **native construction** path (`PERRY_LLVM_INPROCESS=native`)
cannot build any function containing RS4GC root slots. Every `main` execution of
`llvm-inprocess`'s `native-backend` job fails here — runs `31461469705`,
`31461474406`, `31461478986` (all 2026-08-11T05:23Z).

This is the defect behind #7971's "when it does execute, it fails". #7971 covers the
*gate* (vacuous green on PRs, no diagnostic on failure); this is the *code*.

## Reproduction (local, deterministic)

```bash
export LLVM_SYS_221_PREFIX=/opt/homebrew/opt/llvm@22
cargo build --profile perry-dev -p perry -p perry-runtime-static \
  -p perry-stdlib-static --features perry/llvm-inprocess
export PERRY_RUNTIME_DIR="$PWD/target/perry-dev" PERRY_NO_AUTO_OPTIMIZE=1
PERRY_LLVM_INPROCESS=native target/perry-dev/perry \
  experiments/llvm-inprocess-spike/spike.ts -o /tmp/spike_native
```

```
Error compiling module 'spike.ts' ... native IR construction failed in
@perry_closure_spike_ts__8:
in line: %r5 = alloca ptr addrspace(1)
```

The textual path compiles the same program fine. Note the compiler's diagnostic is
already good — CI was redirecting it into `/tmp/native.err` and never printing it,
which is why three `main` failures looked like a bare `exit 1` (fixed in #7971).

## Root cause

`crates/perry-codegen/src/dialect/mod.rs::basic_type` has an arm for `"ptr"` but
none for `ptr addrspace(N)`, so it reaches `bail!("unknown type `{tok}`")`.

Today's codegen emits **five** `addrspace(1)` shapes (116 occurrences in `spike.ts`
alone):

```
%X = alloca ptr addrspace(1)
%X = load ptr addrspace(1)
store ptr addrspace(1) ...
%X = inttoptr i64 %X to ptr addrspace(1)
%X = ptrtoint ptr addrspace(1) %X to i64
```

**It is not a one-line fix.** I added an `addrspace(N)` arm to `basic_type` and
rebuilt: `alloca` then succeeds and the reader fails on the *next* shape,
[...]

PR #7998, opened 2026-08-12
fix(codegen): teach the in-process native backend today's RS4GC IR; unfreeze the dialect corpora

proggeramlug:
Refs #7982.

## The reported defect, and the family behind it

`PERRY_LLVM_INPROCESS=native` could not build any function containing an RS4GC root slot — effectively every function. The issue warned this was not a one-liner, and it was not; each fix exposed the next shape:

| # | shape | what was wrong |
|---|---|---|
| 1 | `alloca ptr addrspace(1)` | `basic_type` had no `addrspace(N)` arm (the reported failure) |
| 2 | `store ptr addrspace(1) null, ptr %r` | `ty_and_val` split it into type `ptr` + value `addrspace(1) null`. The qualifier belongs to the **type** but sits after a space, so operand position needed the same treatment as type position |
| 3 | `null` in an addrspace operand | `constant` always returned an addrspace(0) `null`; the verifier rejects that stored into an `alloca ptr addrspace(1)` |
| 4 | `define … "frame-pointer"="non-leaf" gc "statepoint-example"` | no string-attribute arm; and `gc "…"` contains a space, so the whitespace-split attribute loop saw two junk tokens |
| 5 | `call … ) "gc-leaf-function"` | no string-attribute arm on callsites |
| 6 | `landingpad token cleanup` | only `{ptr, i32} catch` was supported. `token` is not an inkwell `BasicType`, so this pad is built through llvm-sys; the older shape keeps its branch, both occur in one module today |

## Two defects that were hiding behind it — both worse

**The native path returned assembly and called it an object.** The statepoint backends compact the stack map at *assembly* time, so the plan carries `-S`. The textual in-process path has always run the rewrite-and-assemble step; the native and diff paths returned the bytes straight to the object cache and the link died with `ld: unknown file type`. `linker.rs` even carries a comment saying exactly that would happen — for the path that already handled it. Factored into `linker::finish_native_emission`, wired into all four emit sites.

**A natively-constructed module had NO GC strategy, so RS4GC never ran on it.** `native_emit::synth_define_header` was a second, independent copy of `LlFunction::to_ir`'s header renderer, written before `"frame-pointer"="non-leaf"` and `gc "statepoint-example"` existed and never updated. The result verifies, links and runs correctly on any program that does not collect — while having **no precise roots at all**. That is #7332's shape, and it is invisible to a behaviour-parity smoke arm *by construction*: identical output on a non-collecting program is precisely what it produces. Its only symptom was the diff arm's byte mismatch (149,105 text vs 50,995 native), and the diff arm was never reached because the native arm failed first.

Fixed structurally rather than by testing for agreement: one renderer, `LlFunction::define_header`. A copy that can drift eventually does, and this one took two attributes and an unknown number of releases to notice. The pin lives in `function.rs`, which compiles **without** the `llvm-inprocess` feature, so it runs in per-PR `cargo-test` rather than only in the feature job.

## The corpora — arguably the more important half

All three tracked `.ll` files froze on 2026-08-03, **151 codegen commits** earlier, with **zero** `addrspace(1)`. `corpus_spike ... ok` proved the tests ran, not that they test today's IR. And it had already happened once, nine days before (#7310, stale `setjmp` calls) — which is the argument against refreshing by hand a third time.

1. **`scripts/refresh_llvm_inprocess_corpora.sh`** regenerates all three from the built compiler; `--check` diffs instead of writing.
2. **`scripts/check_llvm_corpus_currency.py`** (added to `lint`) asserts every IR form the reader carries a dedicated branch for is **present** in the corpora. A form that disappears fails, and the fix is a refresh or deleting the now-untested branch — the kill policy applied to fixtures.
   **Sabotage-verified against the corpora this PR replaces**: pointed at the `origin/main` files it names all 10 forms they lacked, then passes on the refreshed ones. Its `--self-test` rejects the 2026-08-03 corpus shape and accepts today's.
   Stated limit, because it matters: this catches a form that *vanishes*, not one codegen newly *invents* — nothing static can. The honest closure for that direction is the end-to-end `native` arm, which compiles today's IR by construction.
3. **`llvm-inprocess.yml`'s age diagnostic was itself vacuous.** `git log -- <path>` on the default **depth-1** checkout has one commit to answer from, so "IR-affecting commits since then" printed a confident **`0`** however stale the files were — a diagnostic reassuring readers about the exact thing it existed to detect. Now `fetch-depth: 0`, and it fails loudly rather than printing zero if history is ever missing again.

## Validation

- 934 `perry-codegen` lib tests green under `--features llvm-inprocess` (all three corpus round-trips included), LLVM 22.1.4.
- `PERRY_LLVM_INPROCESS=native` compiles **and runs** `spike.ts` and `test_gap_7302_invoke_eh_paths.ts` with output byte-identical to the textual backend, banner asserted.
- Corpora refreshed: 497 / 1082 / 420 `addrspace(1)` sites; the EH corpus keeps 84 invoke edges and its personality clause.
- `dialect/mod.rs` crossed the 2000-line cap, so type/constant parsing moved to `dialect/types.rs`.

## Not closed here (separate defects, neither ever reached in CI)

- `=diff` still reports a byte mismatch on the spike: **149,105 text vs 163,902 native**. It was 50,995 before the GC-strategy fix, i.e. the gap narrowed from "RS4GC never ran on the native arm" to a real but far smaller divergence. The pre-opt prints differ by design (the C-API builder constant-folds at construction), so this needs its own investigation.
- The unit-split diff arm fails with `call to undeclared @js_shadow_slot_set` — the per-unit skeleton does not declare it.

Both are worth their own issues; this PR takes `native` from "cannot build any GC-rooted function" to "builds and runs correctly", and stops the corpora from silently rotting again.

## Summary by CodeRabbit

- **New Features**
  - Improved in-process LLVM support for current RS4GC IR, including address spaces, constants, attributes, callsites, and landing pads.
  - Native and differential compilation now reliably produce assembled object output when required.
  - Preserved garbage-collection strategies and personality information in generated native modules.
  - Improved handling of pointer types and token-based cleanup landing pads.

- **Bug Fixes**
  - Fixed inconsistencies between native and textual LLVM output.
  - Added clearer failures for unsupported or malformed LLVM constructs.

- **Chores**
  - Added automated LLVM corpus coverage and freshness checks.


proggeramlug:
Follow-up commit `61d0f57da`: the `define_header` pin took three attempts, and the first two failed the same way the bug did — worth recording because it is this PR's own bug class occurring inside its own test.

1. The `to_ir == define_header` agreement test **cannot** see a dropped `gc "statepoint-example"`: with one shared renderer both sides change identically. Sabotage passed.
2. A dedicated strategy test that branched on `native_stack_roots_enabled()` never ran its ON arm under `cargo test` — no module has called `set_native_roots_for_target`, so the predicate is false in the test process. Sabotage passed again.
3. Only pinning **both** lowerings with `NativeRootsPin::{native,shadow}`, and asserting `stack_map_slot_count` in each arm so neither can pass vacuously, goes red on the sabotage.

Sabotage-verified with the fix committed first, then restored **and rebuilt**. 935 `perry-codegen` lib tests green under `--features llvm-inprocess`; the three `define_header_tests` also pass **without** the feature, i.e. in per-PR `cargo-test`.

PR #8009, opened 2026-08-13
feat(codegen/runtime): stamp class ShapeIds at birth

### LINKED ISSUE #7916: perf: object representation is now the binding constraint on the retain cluster (72 bytes per 2-field literal; 216 MB written to store 48 MB)

proggeramlug:
## Summary

- mint each compiled class's canonical `ShapeId` once at module initialization, beside its rooted keys global
- stamp inline, outlined, and scalar-materialized class instances with that id before publication
- preserve the existing allocator ABI and the lazy-stamp fallback when the ShapeId range is exhausted
- add IR and runtime coverage proving the id is minted once and present before any by-name lookup

This is #6759 C3 rung 2, the next prerequisite for #7916's remaining header shrink. It closes the lazy-stamp window so the guard-migration rung can rely on a uniform shape word. It does not close #7916; the guard migration and header-layout change remain follow-up work.

## Measurements

- per-object storage added by this rung: **0 bytes**
- two-field literal representation remains **56 bytes**
- `retain` remains **168 MB written for 48 MB of numeric payload** (**3.5x** amplification)
- no memory or wall-clock improvement is claimed for this invariant/proof step

No version files are changed and there is no version bump.

## Validation

- `cargo test -p perry-codegen --lib` — 935 passed
- `cargo test -p perry-runtime --lib` — 2,278 passed, 4 ignored
- targeted codegen and runtime regression tests rerun after rebasing onto current `main`
- compiled `test-files/test_edge_objects_records.ts` with the built Perry binary and verified byte-for-byte stdout parity with Node
- `git diff --check origin/main...HEAD`

Refs #6759
Refs #7916

## Summary by CodeRabbit

* **Improvements**
  * Compiled class instances now receive stable shape information as they are created across supported allocation paths.
  * Improved consistency when reading class shape data immediately after allocation.
  * Preserved existing object representation and memory characteristics.
  * Added fallback behavior for cases where stable shape identifiers are unavailable.

* **Bug Fixes**
  * Improved reliability of class shape tracking for inline, outlined, and optimized allocations.


proggeramlug:
Reviewed against the rung-1 handoff (`gc-handoff/RUNG1-NOTES.md`). **Substantively sound** — I checked the four ways this rung could have gone wrong and it avoids all of them. Two gaps and one timing request below.

## What I verified

- **It does not touch `typed_feedback::object_shape()`** — the site rung 1 deliberately left gated. Relaxing it makes the guard family fail **closed** (their contracts compare its return against a codegen-supplied *keys pointer*, which an id can never equal), silently deleting the direct-method-call route and the class-field fast paths with **byte-identical output**. Correctly avoided.
- **The guard is unaffected.** `class_field_inline_guard` loads `field_count` (offset 12) and `keys_array` (offset 16); `ka_ok` compares `keys_array` against `expected_keys`. Neither is the shape word, so stamping `parent_class_id` cannot perturb it.
- **Inheritance is safe.** Every remaining runtime reader — `array/subclass.rs:31`, `event_target.rs:397/571`, `json/stringify_tojson_probe.rs:172` — goes through `get_parent_class_id(class_id)`, i.e. the **registry**, not the header word. #7981 converted the last holdout (`thread.rs::serialize_object`). `proxy_reflect.rs:539/865` already reads that word as a shape id (`parent_class_id - 2147483648`).
- **The claims are honest and match**: 0 bytes added, 56-byte two-field literal unchanged, `retain` still 168 MB / 3.5×, and no perf improvement claimed for an invariant step.

## Gap 1 — no assertion that the fast path still fires

18 lines of test, none asserting the emitted hit path still serves a stamped receiver.

Rung 1 shipped `a_class_instance_primes_an_id_token_after_rung1` with `assert_ne!(cache[0], keys)` for a stated reason: *"priming a keys pointer for a stamped receiver would be a permanent miss at that site, so this is the test that keeps the runtime's choice and the IR's choice the same."*

**This rung widens exactly that population** — every class instance now carries an id from birth instead of lazily. A regression in which eagerly-stamped instances prime the wrong token would produce **correct output, green tests, and a permanently dead fast path**. That is the failure mode this ladder has hit twice.

Please add the equivalent assertion for an eagerly-stamped instance. That is the discriminating quantity here; `cargo test` counts and a byte-exact `test_edge_objects_records.ts` run cannot distinguish it.

## Gap 2 — the epoch-guard bypass becomes universal and should be restated

Rung 1 flagged this explicitly for the rungs after it:

> the read PIC's epoch guard (`epoch_ok = is_stamp || epoch_eq`) is now bypassed on essentially every hit, because essentially every primed token is an id. **Id-token soundness therefore rests entirely on `shapes::prune_dead_shape_keys` dropping a dead keys array's record before its address is recycled** (wired into `gc/copying.rs:1962` and `gc/oldgen.rs:1171`).

Rung 1 argued safety because a class instance's keys array is the process-rooted canonical one — **immortal** — with the only mortal case being the post-delete clone, identical in kind to a plain object's.

**That argument still holds for this rung**, but eager stamping makes it universal rather than incidental, and it is currently inherited silently. Please restate it in the PR body or a code comment, so the next reader knows the load-bearing assumption is "class keys arrays are immortal" and not something weaker.

## Request — hold the merge briefly

`main` is **currently regressed**: the quiet-mini sweep at `843ef621f` is **10/19 beating node, down from 12/19** at `0a21611fe` (`cycles` +58%, `deeplist` +38%, `interp` +28%, `asyncpipe` +27%, `pipeline` +24%, `iso_miss` +19%; `churn` and `retain` each lost a node win). Both sweeps were clean-verdict on the same pinned host, so it is real at the large end.

A bisect over `0a21611fe..843ef621f` is running now. This PR is codegen-side and touches the allocation paths (`new_alloc.rs`, `scalar_method.rs`, `closure.rs`, `alloc.rs`), so landing it mid-bisect adds a variable to a hunt that is already ~20 commits wide.

**Nothing here suggests this PR is the cause** — it is not in the regressed range. This is purely about not widening the search while it is in progress. Happy to merge as soon as the bisect names the commit.

PR #8010, opened 2026-08-13
fix(runtime): birth-stamp the class allocators #8009 left lazy, and gate the split population

proggeramlug:
Bisect of the broad regression between `0a21611fe` and `843ef621f`, and the part
of it that is **still live on `main` after #8009**.

> **This PR was rewritten mid-flight.** #8009 landed while it was in validation
> and is the fix for the corpus-visible regression. What remains here is the
> slice #8009 did not cover, plus the gate that can see it. History and both
> dead ends: `gc-handoff/BISECT-NOTES.md`.

## The bisect

**The commit is `4784d5da7` (#7983, "#6759 C3 rung 1 — make the shape word
uniform"), alone.** Neither leading hypothesis survived: #7997 (the aarch64 SVE
prologue decoder) and #7994 (per-thread prototype addresses) both landed *after*
the regression was already fully present.

Wall clock on the dev box cannot resolve this (load 30–200 all session), so this
is **instructions retired** (`/usr/bin/time -l`, best-of-3, exit-checked, output
`cmp`-ed against `m0810/expected/`). Every hop rebuilt
`-p perry -p perry-runtime-static -p perry-stdlib-static` into one target dir
with the `.a` mtimes verified.

| bench | `0a21611fe` good | `23a8aad31` parent | `4784d5da7` **#7983** | `843ef621f` tip | `f58b73f4f` main (with #8009) |
|---|--:|--:|--:|--:|--:|
| cycles | 1,788,433,524 | 1,788,778,236 | **2,759,621,970** | 2,758,613,331 | 1,790,326,029 |
| deeplist | 891,705,830 | 898,511,111 | **1,304,689,118** | 1,297,094,745 | 894,885,943 |
| interp | 11,619,044,867 | 11,616,644,282 | **14,908,419,553** | 14,890,479,866 | 11,678,333,451 |
| pipeline | 2,584,827,137 | 2,582,965,931 | **3,199,913,725** | 3,192,369,325 | 2,591,167,538 |
| iso_miss | 14,201,035,818 | 14,205,407,941 | **17,464,543,037** | 17,444,768,294 | 14,288,084,445 |
| churn | 3,092,382,790 | 3,091,040,177 | 3,127,189,885 | 3,122,830,475 | 3,090,737,693 |
| retain | 1,903,836,211 | 1,929,069,359 | 1,933,154,142 | 1,901,604,111 | 1,932,245,689 |
| asyncpipe | 1,205,121,699 | 1,206,723,078 | 1,233,483,901 | 1,231,925,146 | 1,209,636,954 |
| fib40 | 3,837,818,438 | 3,838,189,564 | 3,839,712,991 | 3,835,231,921 | 3,837,037,463 |

The whole regression appears at #7983 and nothing after it adds any. **#8009
recovers it** — current main is within 0.1–0.6% of the good endpoint. `fib40` is
flat to four decimals throughout; the +2.7% originally flagged was noise.

## Mechanism

The split is **by receiver kind, not program size**: `cycles` (`class Cell`),
`deeplist` (`class LNode`), `interp`, `pipeline`, `iso_miss` regressed; `churn`
and `retain` (`type … = { … }` object literals) did not.

The emitted read PIC (`expr/property_get/generic_dispatch.rs`) derives its
entire cache token from the header shape word:

```
is_stamp = (parent_class_id - 0x8000_0000) u< 0x4000_0000
token    = is_stamp ? (parent_class_id | 1<<62) : keys_array
```

Rung 1 stamped class instances **lazily**, at the first by-name resolve, so one
shape's population split in two and the two halves compute two different tokens:
instance #1 misses, is stamped, primes the id token; instance #2 is newborn,
computes the keys pointer, misses; the handler re-primes the same id; instance
#3 misses. **Hit rate 0%, forever** — not a slow start.

### Single-build proof

Three programs, one compiler, differing only in read passes over the same
3,000,000-instance array:

| program | instructions | delta | per read |
|---|--:|--:|--:|
| build array only | 2,564,930,818 | — | — |
| build + 1 read pass | 2,695,683,246 | **130,752,428** | **43.6** |
| build + 2 read passes | 2,742,277,736 | **46,594,490** | **15.5** |

Pass 1 sees each instance NEWBORN; pass 2 sees the SAME instances already
stamped. **2.8×, and the only difference is the stamp.**

## What this PR fixes

#8009 stamps only the **compiled** entry point,
`js_object_alloc_class_inline_keys_stamped`. Three class-instance allocators are
still on rung 1's lazy self-heal, as its own doc comment states:

* `js_object_alloc_class_with_keys`
* `js_object_alloc_class_dynamic_parent`
* `js_object_alloc_class_inline_keys` (the compatibility entry point)

For any class reaching one of those, the population is **still split** and its
sites still never hit. All three now birth-stamp: the two shape-cached ones read
the id out of the `ShapeCacheEntry::runtime_shape_id` their existing probe
already returns (`shape_cache_get` → `shape_cache_get_with_id`) — one extra u32
from a loaded cache line, a compare and a store; the compatibility entry point
mints from its canonical keys array, which is off the compiled hot path.

**This is neutral on the perf corpus** (all nine programs' classes take the
compiled path #8009 already fixed) — the value is the classes that do not, and
the gate.

## The gate

`a_fresh_class_instance_computes_the_token_the_miss_handler_primed` asserts the
token the miss handler **primes** equals the token a freshly-allocated sibling
**computes**, against the emitted IR's formula transcribed into a test-only
`emitted_pic_token`.

**It fails on `main` as of #8009** — `left: 2199047503880` (a keys pointer),
`right: 4611686020574871741` (bit 62 | id) — and passes here.

#8009's own test asserts that a newborn *carries* a stamp. That is a presence
check: both-stamped and both-unstamped each satisfy it, and it is the **mixture**
that is the bug, so it cannot see this. Comparing the two sides is what can. The
gate also passes in *either* uniform state, so it survives a future policy flip
rather than needing a rewrite.

Sabotage-verified with the fix **committed first**: deleting the birth stamp from
`js_object_alloc_class_with_keys` fails exactly this test and leaves the other
four `c3c_pic_tests` green. Restored and **REBUILT** (`Compiling perry-runtime`
present on the rebuild) before re-confirming.

Two premise assertions in `delete_rest::shape_transition_tests_6759` said a fresh
instance is unstamped; they are **updated, not deleted**, and the parent-chain
test is strictly stronger for it — the word is now clobbered from birth, so there
is no window in which it was ever valid inheritance data.

## Question 2: churn / retain / asyncpipe

Reported at +13% / +13% / +27% wall on the quiet mini but **flat in
instructions**. Peak RSS and GC collection counts are both load-independent, so
this is answerable on a busy box (`PERRY_GC_DIAG=1` **and** `PERRY_GC_TRACE=1` —
DIAG alone prints nothing; positive control: 445 lines and 22
`collection_kind":"minor"` for `cycles`).

| bench | quantity | `0a21611fe` | main | Δ |
|---|---|--:|--:|--:|
| churn | instructions | 3,092,462,749 | 3,124,640,479 | +1.0% |
| | peak RSS (KB) | 24,976 | 24,992 | +0.06% |
| | minors / fulls | 88 / 0 | 88 / 0 | identical |
| retain | instructions | 1,936,036,952 | 1,931,641,207 | −0.2% |
| | peak RSS (KB) | 254,800 | 254,816 | +0.006% |
| | minors / fulls | 4 / 0 | 4 / 0 | identical |
| asyncpipe | instructions | 1,206,387,962 | 1,231,635,504 | +2.1% |
| | peak RSS (KB) | 38,832 | 39,168 | +0.9% |
| | minors / fulls | 1 / 0 | 1 / 0 | identical |
| **cycles** (control) | instructions | 1,790,911,190 | 2,759,606,611 | **+54%** |
| | peak RSS (KB) | 24,816 | 24,800 | −0.06% |
| | minors / fulls | 22 / 0 | 22 / 0 | identical |

* **GC-scheduling explanation: refuted.** Collection counts identical on all
  four, zero full collections anywhere.
* **Locality/footprint explanation: refuted.** Peak RSS flat to ≤0.9%.
* **Work: flat** — churn +1.0%, retain −0.2%.

★ Note the control: `cycles` regressed **54%** with RSS and collection counts
*also* flat. Flat RSS/counts can only REFUTE the scheduling and footprint
explanations; they never confirm "nothing changed". The positive statement for
churn/retain is the instruction count.

With no work added, no collection rescheduled and no footprint change, there is
no mechanism left for a 13% wall move: **churn and retain are mini-side
variance**, and both should regain their node wins now that #8009 has landed.

`asyncpipe` is the one row with a real attributable cost: **+2.1%**, not +27%,
and #8009 returns it to 1,209,636,954 against the good endpoint's 1,205,121,699.
The remaining ~25 points are either mini variance or **idle/parked time**, which
neither instructions retired nor cycles elapsed can observe for an async program
— only the quiet mini can settle that.

## Validation

* `cargo test -p perry-runtime --lib` (`RUST_TEST_THREADS=1`): **2279 pass, 0 fail**.
* `iso_miss` canary prints `checksum 437840 misses 0`, including under
  PERRY_GC_VERIFY_EVACUATION=1`; `cycles`/`deeplist`/`interp` byte-identical
  under the same knobs.
* Whole probe corpus output-verified against `m0810/expected/` at every arm.
* Gap suite, class/property/shape/delete/proto slice (70 tests), graded against
  `test-parity/gap_snapshot.json`: **68 pass, 1 expected-fail, 0 unexpected**.
  The 70th (`test_gap_3527_http_ctor_prototype`) is #7999's own tokio link guard
  refusing an ad-hoc build set without `-p perry-ext-http` — an artefact of
  compiling outside the harness.

★ The full 554-test gap suite could not be run on this host: the harness stalls
at 0% CPU after 10–25 tests, on a full run and a 1/8 shard alike, while the same
compile finishes in seconds standalone. The dev box was at **8.6 GiB free** (the
protocol wants ≥ 25 GiB for parity runs) under heavy contention. Host
constraint, not a property of this change — CI runs the full suite here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed property-cache misses caused by inconsistent shape metadata on newly created class instances.
  * Ensured object instances receive consistent shape information at creation, improving property access reliability and performance.
  * Preserved correct shape behavior across field resolution, inheritance, and dynamically created layouts.
* **Documentation**
  * Added release documentation describing the regression, measurements, and validation results.

### DIFF
--- a/changelog.d/8010-class-shape-uniformity-gate.md
+++ b/changelog.d/8010-class-shape-uniformity-gate.md
@@ -1,0 +1,38 @@
+### Birth-stamp the class allocators #8009 left lazy, and gate the split population (#7983)
+
+#8009 (C3 rung 2) stamps a class instance's ShapeId at birth on the **compiled**
+path. Three other class-instance allocators were left on rung 1's lazy
+self-heal — `js_object_alloc_class_with_keys`,
+`js_object_alloc_class_dynamic_parent`, and the `js_object_alloc_class_inline_keys`
+compatibility entry point — so for any class reaching one of them the shape's
+population is still split between stamped and newborn receivers.
+
+That is not a slow start. The emitted read PIC derives its entire cache token
+from the header shape word, so a stamped receiver and a newborn one of the same
+shape compute two different tokens and the site's hit rate is **0% forever**:
+instance #1 misses, is stamped, primes the id token; instance #2 is newborn,
+computes the keys pointer, misses; the handler re-primes the same id; instance
+#3 misses.
+
+This is the defect bisected to `4784d5da7` (#7983). On instructions retired,
+isolated against its own parent: `cycles` +54.3%, `deeplist` +45.2%, `interp`
++28.3%, `pipeline` +23.9%, `iso_miss` +22.9% — while the object-literal
+benchmarks (`churn` +1.2%, `retain` +0.2%) and `fib40` (+0.04%) did not move,
+literals having been birth-stamped since #6804. Isolated further on one program
+and one build: a read pass over 3,000,000 newborn class instances costs 43.6
+instructions per read, a second pass over the same now-stamped instances 15.5.
+
+All three allocators now stamp at birth. The two shape-cached ones read the id
+out of the `ShapeCacheEntry::runtime_shape_id` their existing probe already
+returns; the compatibility entry point mints from its canonical keys array, off
+the compiled hot path.
+
+The gate, `a_fresh_class_instance_computes_the_token_the_miss_handler_primed`,
+asserts the token the miss handler PRIMES equals the token a freshly-allocated
+sibling COMPUTES, comparing against the emitted IR's formula transcribed into
+the test. It fails on `main` as of #8009 and passes here. "A newborn carries a
+stamp" is a presence check that both-stamped and both-unstamped each satisfy —
+only the mixture is the bug, and only a test comparing the two sides can see it.
+This one passes in either uniform state, so it survives a future policy flip.
+
+Working notes, including the full bisect and the two dead ends: `gc-handoff/BISECT-NOTES.md`.
--- a/crates/perry-runtime/src/object/alloc.rs
+++ b/crates/perry-runtime/src/object/alloc.rs
@@ -296,8 +296,15 @@ fn object_alloc_class_inline_keys_impl(
 }
 
 /// Compatibility entry point for runtime callers that do not have a
-/// module-init ShapeId. Their first by-name resolve retains rung 1's lazy
-/// self-heal; compiled allocations use the stamped entry point below.
+/// module-init ShapeId.
+///
+/// It mints the id from the canonical keys array instead of receiving it, so
+/// the instance is still stamped AT BIRTH. Leaving it to rung 1's lazy
+/// self-heal would split this class's population between stamped and newborn
+/// receivers, which the emitted PIC cannot tolerate — see
+/// `shapes::birth_stamp_object_shape`. The mint is one shape-table probe and
+/// this is not the compiled hot path (compiled `new C(…)` sites call
+/// `js_object_alloc_class_inline_keys_stamped` with a module-init id).
 #[no_mangle]
 pub extern "C" fn js_object_alloc_class_inline_keys(
     class_id: u32,
@@ -305,7 +312,18 @@ pub extern "C" fn js_object_alloc_class_inline_keys(
     field_count: u32,
     keys_array: *mut ArrayHeader,
 ) -> *mut ObjectHeader {
-    object_alloc_class_inline_keys_impl(class_id, parent_class_id, field_count, keys_array)
+    let ptr =
+        object_alloc_class_inline_keys_impl(class_id, parent_class_id, field_count, keys_array);
+    if !keys_array.is_null() {
+        unsafe {
+            let id = crate::object::shapes::shape_id_for_keys_ensure(
+                keys_array as *const ArrayHeader,
+                (*keys_array).length,
+            );
+            crate::object::shapes::birth_stamp_object_shape(ptr, id);
+        }
+    }
+    ptr
 }
 
 /// The compiled-class allocation entry point after #6759 C3 rung 2.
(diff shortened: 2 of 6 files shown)


proggeramlug:
HOLD — do not merge. #8009 (`144867bfc`, codegen birth-stamping) landed on `main` while this was in validation, and it is rung 2: it makes the population uniformly STAMPED, which is the other correct fix and the design direction. This PR makes the runtime refuse to READ a stamp, so on top of #8009 it would prime the keys-pointer token while the emitted PIC computes the id token — reintroducing the same 0%-hit-rate regression from the opposite side. Rebasing onto current main and re-measuring; if #8009 already recovers the corpus this PR reduces to the regression gate + bisect notes.

proggeramlug:
Hold lifted — rewritten and re-validated on top of #8009. Summary of the change of course: #8009 landed mid-validation and IS the fix for the corpus-visible regression (current main is within 0.1-0.6% of the good endpoint), so the original approach here was dropped. What remains is the slice #8009 did not cover — the three class-instance allocators still on rung 1's lazy self-heal — plus the gate, which **fails on main as of #8009** and passes here. Ready for review.

PR #8013, opened 2026-08-13
fix(gc): close computed access rooting windows

### LINKED ISSUE #7640: index_get/index_set/property_set: ~20 arms make no rooting decision at all (the population the Layer 1 ledger cannot see)

proggeramlug:
## Summary

Closes the remaining computed-access GC relocation windows from #7640. Earlier receiver/key operands are now selectively rooted only when later user expressions can collect, while proven non-collecting hot paths retain their existing IR.

## Changes

- Protect the remaining typed-array and erased-receiver read/write operands, including custom native-index and dynamic-value lowerings.
- Root masked-window and Ptr<NumArray> receivers across collecting indexes/values, while keeping literal and loop-counter paths root-free.
- Delay typed-array backing-pointer derivation until after RHS evaluation, and use the replacement array head returned by a growing store for its write barrier.
- Root compound class-field receivers across allocating RHS expressions while preserving the existing zero-cost LocalGet/This root-reload path.
- Audit the remaining section-E helpers and document the helpers whose evaluation order already leaves no receiver or raw pointer live.
- Add IR-level rooting/barrier regressions and extend the two end-to-end gap fixtures.

## Related issue

Closes #7640

## Test plan

- `cargo check -p perry`
- `cargo test -p perry-codegen --lib` — 945 passed
- `./run_parity_tests.sh --filter test_gap_7640_computed_key_windows` — PASS against Node 26.5.1
- `./run_parity_tests.sh --filter test_gap_gc_class_field_receiver_rooting` — PASS against Node 26.5.1
- Shadow GC corpus: 143/143 sources, zero dominance violations, zero unrooted allocas, stale-register budget 21/39, all 40 seeded violations caught
- Native/statepoint GC corpus: 143/143 sources, 36,864 statepoints, zero unrooted/stale hazards, all 40 seeded violations caught
- `git diff --check`

- [x] `cargo build --release` clean
- [x] (if CLI / stdlib / runtime API changed) Updated `docs/src/` — not applicable
- [x] (if touching a platform UI backend) Built `-p perry-ui-<backend>` locally on that platform — not applicable

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of computed array, typed-array, buffer, and class-field access when expressions trigger garbage collection or memory growth.
  * Fixed cases where computed reads and writes could use outdated receivers, indexes, values, or array locations.
  * Improved handling of typed-array numeric conversions and dynamic keys.
  * Ensured array updates continue to apply write barriers to the current array storage.

* **Tests**
  * Added regression coverage for computed accesses, typed arrays, class fields, and moving garbage collection scenarios.

### DIFF
--- a/changelog.d/8013-computed-access-rooting.md
+++ b/changelog.d/8013-computed-access-rooting.md
@@ -1,0 +1,26 @@
+**Fixed: remaining computed reads and writes could reuse GC-stale receivers, keys, and backing pointers (#7640).**
+
+Several typed-array, numeric-array, and generic computed-access paths evaluated
+a receiver, then evaluated a key or value that could run user code and trigger
+a moving collection, and finally consumed the receiver from its original LLVM
+register. The same audit found typed-array stores deriving raw backing pointers
+before an allocating RHS, and the array-growth path issuing its write barrier
+against the pre-growth handle even when the helper returned a replacement.
+
+The remaining read/write operands now use selective rooting and are re-read
+after collecting expressions. Masked-window and `Ptr<NumArray>` native paths
+use the custom-lowering form of the same scope; raw backing pointers are loaded
+only after the RHS; and reallocating array stores shade through the returned
+live head. Literal keys, loop counters, and other proven non-collecting windows
+still emit no temporary-root traffic.
+
+Class-field stores now distinguish their two receiver shapes explicitly. Bare
+locals and `this` retain the existing zero-cost `root_reload` repair, while a
+compound receiver such as `this.target.x` is conditionally rooted across an
+allocating RHS because its phi result cannot be re-derived from a local root.
+
+IR regressions assert the typed-array read/write groups, the erased-receiver
+store, and the realloc-path barrier operand. Both end-to-end fixtures remain
+byte-identical with Node 26.5.1. The 143-source shadow and native/statepoint GC
+corpora report zero dominance, unrooted-allocation, or statepoint hazards and
+catch all 40 seeded violations in each lowering.
--- a/crates/perry-codegen/src/expr/buffer_access.rs
+++ b/crates/perry-codegen/src/expr/buffer_access.rs
@@ -460,6 +460,9 @@ pub(crate) fn lower_buffer_store(
     let Some(proof) = lower_buffer_access_proof(ctx, buffer_expr, index_expr, spec)? else {
         return Ok(None);
     };
+    // #7640 section E audit: `proof` contains stable slot metadata and the
+    // lowered native index, not a receiver JSValue or raw backing-store
+    // pointer. Lower the RHS before `emit_buffer_access_pointer` loads either.
     let val_i32 = lower_value_i32(ctx, value_expr)?;
     let emission = emit_buffer_access_pointer(ctx, &proof, spec);
     let byte_val = ctx.block().trunc(I32, &val_i32, I8);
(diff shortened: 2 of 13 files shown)


proggeramlug:
Post-merge audit blocker (the merge raced this audit):

1. In `masked_window.rs`, the new operand-rooting callback reloads the receiver after `lower_expr_as_i32(index)`, but the `MaskedWindowElem::Ta*` arms still ignore that live receiver and consume the raw `data_ptr` cached at fast-copy entry. This is reachable: the masked-window/range matcher admits shapes such as `x & mask` for any operand, while `ToInt32(x)` can invoke user `valueOf` code and collect. A moving collection can rewrite `TYPED_ARRAY_VIEW_META.backing`; user code can also dispose native-arena backing. The receiver reload therefore does not repair the element pointer, leaving a stale read/UAF window. Revalidate liveness/kind and rederive the typed-array data pointer from the reloaded receiver after index lowering, with a coercion+moving-GC regression.

2. The same lifetime issue remains in `lower_typed_array_store`. `lower_buffer_access_proof` clones the view/proof before lowering the RHS. A numeric-returning RHS call can invalidate or dispose a native-owned view, but `emit_buffer_access_pointer` then uses the stale cloned proof and unchanged raw `data_slot`. Revalidate the current view/liveness after RHS lowering and rederive the pointer, or decline this fast path when that cannot be proven.

I did not consider this head safe to merge. CI also had active failures at merge time: `gc-parse-churn-gate` (duplicate `@perry_typed_parse_keys_0`) and the Windows structural audit.

proggeramlug:
Follow-up to the post-merge audit: the masked-window concern now has a concrete emitted-IR witness, without relying on CI. An admitted `view[(+key) & 7]` shape with `key: any` emits `js_typed_array_masked_window_data_ptr` before `js_number_coerce(key)`, then loads through the cached pointer afterward. Unary `+` can invoke user coercion and collect. This proves the raw-pointer lifetime window in codegen; it does **not** by itself establish an observed workload regression or justify a revert. I recorded the required forward-fix condition on #8014.

PR #8016, opened 2026-08-13
perf(json): right-size tape object spill buffers

### LINKED ISSUE #7267: perf(json): tape materialization allocates field_count=0, so every 5+-field parsed record spills to overflow storage
Found while fixing #7264 (PR #7265). Perf, not correctness.

`json_tape.rs::materialize_object` allocates every object with `js_object_alloc(0, 0)` and then adds each property with `js_object_set_field_by_name`:

```rust
let obj = crate::object::js_object_alloc(0, 0);
...
crate::object::js_object_set_field_by_name(obj, key_ptr, value);
```

`js_object_alloc_with_parent` reserves `max(field_count, INLINE_SLOT_FLOOR)` slots — with `field_count = 0` that is the floor, currently **4**. Slots 4..N therefore go to overflow storage (`spill.rs`), and `field_count` stays pinned at 4. So **every JSON record with 5 or more properties** — the single most common API shape there is — keeps its 5th and later values out of line, paid on every read *and* every write, plus a per-object spill entry that every GC cycle visits, rekeys and finalizes.

The direct parser does not have this problem: both `parse_object` and `parse_object_shaped` pass the real key count to `js_object_alloc_class_inline_keys`, so everything lands inline.

## Measurement

`benchmarks/json_polyglot/bench_field_access.ts` (10k records × 5 fields, 50 iterations; `nested` is index 4, i.e. the first overflow slot, and the workload reads it every iteration), same binary, `--profile perry-dev`:

```
default (tape)        2950–2983 ms
PERRY_JSON_TAPE=0      870–877 ms
```

3.4×. Some of that gap is other tape work, but the overflow lookup is on the hot read path by construction here, and #7265 puts it on the stringify path as well (correctly — the values genuinely live there).

## Suggested fix

Count the object's top-level keys before allocating. The tape already carries what is needed: each `KIND_OBJ_START`/`KIND_ARR_START` entry stores its matching end index in `link`, so a top-level walk from `idx` to `end_idx` that counts `KIND_KEY` and skips nested values via `link` is O(top-level entries). Then `js_object_alloc(0, key_count)` and every field lands inline.

This is a behaviour change to the parser (allocation sizing, `field_count` in the header, spill-map population) rather than a correctness fix, which is why #7265 left it alone. Worth benchmarking on its own.

Note the interaction with #6712: the floor moved 8 → 4, which doubled the population of records that spill. Anything that measured tape-path JSON before 2026-07-20 was measuring a materially different allocation profile.

proggeramlug:
## Summary

Right-size overflow storage for objects materialized from JSON tapes without
widening their primary object allocation. This avoids the issue's measured
inline-allocation regression while removing the generic 16-slot spill padding
from parsed records.

## Changes

- Count each recursive tape object's top-level keys without allocating, using
  container links to skip nested values.
- Add an exact-sized holey-array allocator for runtime-owned side storage and
  use it to reserve object spill buffers before field insertion.
- Apply the same reservation to the iterative deep-input materializer using its
  already-collected key count.
- Add regression tests that pin exact outer/nested spill capacities and ensure
  primary objects remain at `INLINE_SLOT_FLOOR`.

## Related issue

Closes #7267.

## Test plan

- [x] `./scripts/test_affected_crates.sh --base origin/main` passes:
  `perry-runtime` 2281 passed / 4 ignored, `perry` 959 passed, `perry-ffi` 26
  passed.
- [x] All 22 `json_tape::tests` pass.
- [x] `cargo fmt --all -- --check`, `git diff --check`, and
  `./scripts/check_file_size.sh` pass.
- [x] Built the benchmark configuration with `--profile perry-dev`,
  `--no-default-features --features dev-cli`, and matching static runtime and
  stdlib archives.
- [ ] `./scripts/pre-tag-check.sh --quick` currently reaches an unrelated
  pre-existing failure on current `origin/main`:
  `crates/perry-codegen/src/expr/property_set.rs:1457` lacks a
  `GC_STORE_AUDIT` marker. The new allocator's store-site audit passes.

## Performance

Eight interleaved runs of
`benchmarks/json_polyglot/bench_field_access.ts` with the `perry-dev` profile
and `PERRY_NO_AUTO_OPTIMIZE=1`:

| Metric | Baseline | This PR | Change |
| --- | ---: | ---: | ---: |
| Median time | 1266.5 ms | 1138.5 ms | -10.1% |
| Minimum time | 1137 ms | 1009 ms | -11.3% |
| Peak RSS | 309.3 MiB | 294.3 MiB | -4.8% |

All checksums matched. A four-run direct-parser control (`PERRY_JSON_TAPE=0`)
was effectively unchanged at 934.5 ms baseline versus 935.5 ms for this PR,
isolating the improvement to tape-backed materialization.

## Screenshots / output

Not applicable; this is an internal runtime allocation change.

## Checklist

  (maintainer handles these at merge).
- [x] My commit follows the repository's conventional prefix style.
- [x] I've read CONTRIBUTING.md and the Code of Conduct.

## Summary by CodeRabbit

* **Performance**
  * Improved JSON object materialization efficiency by allocating spill storage based on exact field counts.
  * Reduced field-access time and peak memory usage while preserving output checksums.

* **Bug Fixes**
  * Improved handling of nested JSON objects and large object key counts.
  * Ensured arrays created with explicit lengths are correctly initialized with empty slots.

* **Tests**
  * Added coverage for spill capacity sizing in recursive and iterative JSON processing, including nested objects.

### DIFF
--- a/changelog.d/8016-json-tape-exact-spill.md
+++ b/changelog.d/8016-json-tape-exact-spill.md
@@ -1,0 +1,21 @@
+### perf(json): right-size tape-materialized object spill buffers (#7267)
+
+JSON tape materialization previously allocated every object at the two-slot
+inline floor, then let the first overflowing property create a general-purpose
+array with 16 slots of growth headroom. Five-field records therefore carried a
+16-slot side allocation even though the tape already knew their final width.
+
+Both the recursive lazy materializer and the iterative deep-input materializer
+now reserve an exact-width spill buffer before storing fields. The recursive
+path counts only the current object's keys by hopping across nested container
+links; the iterative path reuses the key count it has already collected. The
+primary object deliberately remains at `INLINE_SLOT_FLOOR`: sizing its inline
+allocation to every key was benchmarked in #7267 and regressed the named field
+access workload.
+
+On `benchmarks/json_polyglot/bench_field_access.ts`, eight interleaved
+`perry-dev` runs with `PERRY_NO_AUTO_OPTIMIZE=1` reduced median time from
+1266.5 ms to 1138.5 ms (-10.1%) and peak RSS from 309.3 MiB to 294.3 MiB
+(-4.8%), with identical checksums. A direct-parser control was unchanged
+(934.5 ms vs 935.5 ms median), isolating the improvement to tape-backed object
+materialization.
--- a/crates/perry-runtime/src/array/alloc.rs
+++ b/crates/perry-runtime/src/array/alloc.rs
@@ -140,6 +140,34 @@ pub extern "C" fn js_array_alloc_with_length(capacity: u32) -> *mut ArrayHeader 
     ptr
 }
 
+/// Allocate an exact-sized holey array for runtime-owned side storage.
+///
+/// Unlike [`js_array_alloc_with_length`], this does not add
+/// [`MIN_ARRAY_CAPACITY`] growth headroom. Callers must know their final width;
+/// the spill buffer used by JSON tape materialization does, and padding every
+/// parsed object to 16 side slots would otherwise dominate the object itself.
+pub(crate) fn js_array_alloc_with_length_exact(capacity: u32) -> *mut ArrayHeader {
+    let ptr = arena_alloc_gc(
+        array_byte_size(capacity as usize),
+        8,
+        crate::gc::GC_TYPE_ARRAY,
+    ) as *mut ArrayHeader;
+
+    unsafe {
+        (*ptr).length = capacity;
+        (*ptr).capacity = capacity;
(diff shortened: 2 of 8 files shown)


proggeramlug (commented):
Review complete: no blocking correctness finding. The recursive materializer owns a handle and reloads it around allocating work; the iterative path is explicitly called inside `gc_suppress()` and documents suppression as a caller precondition, so its raw object pointer cannot relocate during spill reservation. The automated stale-pointer warning therefore does not apply to the production path. Exact capacity remains an upper bound even with duplicate keys, and absolute spill indexing is preserved. Local `perry-runtime` JSON-tape validation passed: 27 tests, including both new exact-spill tests and the existing moving-GC callback-root tests.

PR #8018, opened 2026-08-13
fix(windows): bundle the LLVM runtime with Perry

### LINKED ISSUE #7985: windows: perry.exe cannot link against the official LLVM 22 release — /MT vs /MD, bundled rpmalloc, and missing inkwell target backends
## Summary

Linking `perry.exe` on `windows-latest` against the pinned `C:\llvm` (the official
`clang+llvm-22.1.3-x86_64-pc-windows-msvc` tarball installed by
`.github/actions/setup-llvm22`) fails with **three independent `link.exe`
errors**. This is the `windows-latest` arm of `gc-native-roots` (#7970), and it is
a real build breakage, not a harness problem.

**It is very likely also latent in `test.yml`'s `windows-build`.** That job builds
the same `-p perry` with the same action, but it has been dying at its *first*
step since #7977 (cp1252 decode), so its build step has not executed on any recent
PR. Expect this to surface there the moment #7977 lands — please read the two
together.

## Evidence

Run `31405431962`, job `native-roots-rs4gc (windows-latest, x86-64, PE)`,
`LLVM_SYS_221_PREFIX: C:\llvm`, step "Build compiler and static runtime":

```
error: linking with `link.exe` failed: exit code: 1120
```

### 1. CRT model mismatch (mimalloc /MD vs llvm-sys /MT)

```
liblibmimalloc_sys-*.rlib(mimalloc-static.o) : error LNK2038: mismatch detected for
'RuntimeLibrary': value 'MD_DynamicRelease' doesn't match value 'MT_StaticRelease'
in libllvm_sys-*.rlib(Core.cpp.obj)
LINK : warning LNK4098: defaultlib 'libcmt.lib' conflicts with use of other libs
```

The official LLVM Windows release is built `/MT` (static CRT); Rust's msvc target
and `libmimalloc-sys` use `/MD`. They cannot be linked into one image as-is.

### 2. Allocator symbol collision (LLVM release bundles rpmalloc)

```
ucrt.lib(api-ms-win-crt-heap-l1-1-0.dll) : error LNK2005: calloc  already defined in libllvm_sys-*.rlib(rpmalloc.c.obj)
...                                                     free     already defined in ... rpmalloc.c.obj
...                                                     malloc   already defined in ... rpmalloc.c.obj
...                                                     realloc  already defined in ... rpmalloc.c.obj
```

`build_llvm_release.bat` links **rpmalloc** into the Windows release to speed up
LLD. Those definitions come along through `llvm-sys` and collide with the UCRT —
and, separately, Perry already has its own allocator (mimalloc), so there would be
two.

### 3. Missing target backends inkwell references

```
libinkwell-*.rlib : error LNK2019: unresolved external symbol LLVMInitializeMipsTarget
                                   ... LLVMInitializeLanaiTarget
                                   ... LLVMInitializeSparcTarget
                                   ... LLVMInitializeMSP430Target
                                   ... LLVMInitializeXCoreTarget       (and the
   matching TargetInfo / AsmPrinter / AsmParser / Disassembler / TargetMC symbols)
```

inkwell's `Target::initialize_*` surface expects experimental targets the official
Windows release does not build.

## Why it was invisible

`gc-native-roots` has never had a green run (#7970), and is not in branch
protection's required contexts, so a red result there is indistinguishable from
the queue noise. The only other Windows job, `windows-build`, has been failing
27 s in at an unrelated step (#7977). Between them, **nothing has successfully
linked `perry.exe` on Windows in CI for some time.**

The `xml2s.lib` comment block in `.github/actions/setup-llvm22/action.yml` says
this became fatal when #7353 made the in-process LLVM backend the default and
statically linked, and explicitly notes that "test.yml's windows-build and
gc-native-roots.yml's PE arm fail identically". That comment is about a
*previous* instance of the same class; these three errors are past it.

## Directions (not a recommendation — this needs a Windows-toolchain decision)

- Build the perry Windows job against an LLVM built `/MD` without rpmalloc
  (e.g. a self-built or vcpkg LLVM), rather than the official release tarball.
- Or `/NODEFAULTLIB` + `/FORCE:MULTIPLE` surgery, which trades a link error for
  an allocator whose behaviour nobody has characterised — probably worse.
- Or build `perry` on Windows **without** the default `llvm-inprocess` feature, so
  `llvm-sys`/inkwell are not linked at all, and keep the in-process backend to
  the platforms it is actually exercised on. This is the smallest change and the
  one most consistent with the feature's current CI coverage (macOS only, and see
  #7982).

## Separately fixed

The same job then failed a second time (exit 2) at the Windows LLVM download:

```
tar (child): Cannot connect to D: resolve failed
xz: (stdin): File format not recognized
```

Git-bash GNU tar reads `$RUNNER_TEMP` (`D:\a\_temp`) as a remote `host:path`.
That one is a workflow bug and is fixed in the #7970 PR with `tar --force-local`
plus a post-extraction check. It is noted here only so the log is not confusing.

proggeramlug:
## Summary

- bundle the official LLVM 22 `LLVM-C.dll` beside `perry.exe` in the Windows release zip
- keep the DLL beside the executable when staging `@perryts/perry-win32-x64`
- fail packaging/staging if the PE import or runtime DLL is missing
- update the `gc-native-roots` status after #8017's dynamic-link fix

`perry.exe` on current `main` links successfully after #8017, but imports `LLVM-C.dll`. The release workflow did not package that DLL, so an extracted release exited with Windows loader error `0xC0000135` unless LLVM happened to be installed globally. This completes the #7985 fix without restoring the incompatible static `/MT` + rpmalloc link.

## Validation

- `cargo build --profile perry-dev -p perry` on Windows with the official LLVM 22.1.8 archive
- PE imports include `LLVM-C.dll` and `VCRUNTIME140.dll`
- isolated `perry.exe` + `LLVM-C.dll` bundle runs with `C:\llvm\bin` removed from `PATH`
- `tests/test_stage_npm_windows_llvm.sh` passes its copy and missing-DLL refusal arms
- workflow YAML parse, shell syntax, and `git diff --check`

No version bump.

Closes #7985

## Summary by CodeRabbit

* **Bug Fixes**
  * Windows release and npm packages now include the required `LLVM-C.dll` alongside `perry.exe`.
  * Packaging now validates required Windows runtime files and fails clearly when they are missing, preventing startup loader errors.
  * Windows arm workflow documentation now reflects resolved linker, archive, and Git Bash packaging issues.

* **Tests**
  * Added automated checks to verify Windows packages include the runtime DLL and reject incomplete packages.

### DIFF
--- a/.github/workflows/gc-native-roots.yml
+++ b/.github/workflows/gc-native-roots.yml
@@ -138,19 +138,18 @@
 #                 prologue words, which is enough to say which walker is wrong
 #                 without a second run.
 #
-#   windows-latest    REAL DEFECT, filed as #7985 (`perry.exe` cannot link
-#                 against the official LLVM 22 release: /MT-vs-/MD CRT mismatch,
-#                 bundled rpmalloc redefining malloc, and inkwell referencing
-#                 target backends the release does not build). A SECOND,
-#                 separate failure in the same arm — Git-bash `tar` reading
-#                 `D:\a\_temp` as a remote host — was a workflow bug and is
-#                 fixed here with `--force-local`.
+#   windows-latest    FIXED. #8017 stopped linking the official archive's
+#                 incompatible /MT + rpmalloc static objects, restricted
+#                 inkwell to Perry's x86/AArch64 targets, and linked the
+#                 archive's LLVM-C.dll instead. #7985's release follow-up
+#                 keeps that DLL beside perry.exe in the Windows zip/npm
+#                 package. A SECOND, separate failure in the same arm —
+#                 Git-bash `tar` reading `D:\a\_temp` as a remote host — was a
+#                 workflow bug fixed here with `--force-local`.
 #
-# So: after #7970 and #7997 three of the four arms should pass and only
-# windows-latest should remain red, on #7985. This workflow is therefore still
-# NOT a promotion candidate — promoting it while #7985 is open would block every
-# PR. Promote only once all four arms are green, and per CLAUDE.md, run it green
-# once BEFORE adding it to branch protection.
+# So: after #7970, #7997 and #8017 all four arms should pass. Per CLAUDE.md this
+# becomes a promotion candidate only AFTER one complete green run demonstrates
+# that claim; do not add it to branch protection based on a comment.
 name: gc-native-roots
 on:
   # Must run where it can actually gate something. Branch-scoped triggers were
--- a/.github/workflows/gc-native-roots.yml
+++ b/.github/workflows/gc-native-roots.yml
@@ -138,22 +138,22 @@
 #                 prologue words, which is enough to say which walker is wrong
 #                 without a second run.
 #
-#   windows-latest    PARTIALLY FIXED. #8017 stopped linking the official archive's
+#   windows-latest    FIXES LANDED; GREEN RUN PENDING. #8017 stopped linking the official archive's
 #                 incompatible /MT + rpmalloc static objects, restricted
 #                 inkwell to Perry's x86/AArch64 targets, and linked the
 #                 archive's LLVM-C.dll instead. #7985's release follow-up
 #                 keeps that DLL beside perry.exe in the Windows zip/npm
 #                 package. A SECOND, separate failure in the same arm —
-#                 Git-bash `tar` reading `D:\a\_temp` as a remote host — was a
-#                 workflow bug. `--force-local` is ineffective with the
-#                 runner's bsdtar; #8028 tracks normalizing the archive and
-#                 extraction paths with `cygpath`. Do not call this arm green
-#                 until that fix is merged and a complete run measures it.
+#                 Git-bash `tar` reading `D:\a\_temp` as a remote host — was
+#                 fixed by #8028, which normalizes shell paths with `cygpath`
+#                 and exports native Windows paths for Perry. `--force-local`
+#                 was ineffective with the runner's bsdtar.
 #
-# So: after #7970, #7997 and #8017 the three non-Windows arms should pass;
-# windows-latest remains unproven pending #8028. Per CLAUDE.md this becomes a
-# promotion candidate only AFTER one complete green run demonstrates all four
-# arms; do not add it to branch protection based on a comment.
+# So: all known defects now have fixes in the intended combined state, but the
+# windows-latest arm remains unproven until a complete run measures it. Per
+# CLAUDE.md this becomes a promotion candidate only AFTER one complete green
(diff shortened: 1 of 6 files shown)


proggeramlug (commented):
Audited exact commit 17b3726d86332640f2a71322c70479482c7e7724. The DLL import validation, release copy, npm staging refusal, and local copy/refusal regression are sound. One blocker remains in `.github/workflows/gc-native-roots.yml`: this PR changes the Windows status to FIXED and says the tar failure was fixed by `--force-local`, then says all four arms should pass. #8028 documents and changes the opposite: the runner uses bsdtar, `--force-local` was ineffective, and the Windows arm still fails until the path is normalized. Please make this status text accurate against the intended combined state (cygpath fix, with green status only after measured evidence). I have not merged this head.

proggeramlug:
Addressed in d0829e503. The workflow status now says Windows is only partially fixed, records that `--force-local` is ineffective with the runner's bsdtar, points to #8028's `cygpath` normalization, and explicitly requires a measured four-arm green run before promotion. YAML parsing and `git diff --check` pass.

proggeramlug (commented):
Re-audited exact head d0829e50 after #8028 and #8027 landed. The functional packaging change is sound: release staging proves perry.exe imports LLVM-C.dll before copying it, npm staging requires/copies the DLL beside the executable, and the positive plus missing-DLL refusal regression passes locally. One blocker remains in the operational status comment: the merged result now says windows-latest is "pending #8028" and waits "until that fix is merged," but #8028 is already on main. Please update the combined-state wording to say the cygpath/native-env fix has landed while the arm remains unproven until a complete measured run. That preserves the important no-green-by-comment rule without leaving a false pending dependency. I have not merged this exact head. No version bump needed.

proggeramlug:
Addressed the remaining combined-state wording blocker in `9dedfab59`. The workflow note now records that #8028's `cygpath` normalization and native Windows environment export have landed, while keeping `windows-latest` explicitly unproven until a complete measured four-arm green run. It also preserves the finding that `--force-local` was ineffective with the runner's bsdtar. YAML parsing and `git diff --check` pass; no version or lockfile change.

proggeramlug (commented):
Re-audited exact head 9dedfab59207df6194833c40caed77403e7203bb. The prior operational-comment blocker is fixed: it now records both landed fixes while keeping the Windows arm unproven until a measured complete run. Release staging validates the LLVM-C.dll import before copying the runtime, npm staging fails closed when the DLL is absent, and the positive plus refusal regression passes locally. Merge-tree and diff checks are clean. No version bump.

Issue #6617, opened 2026-07-18
windows-ui: decouple JS servicing from the 50ms heartbeat (timer latency, 20fps onFrame)


proggeramlug:
All JS servicing on the Windows UI thread — timer/interval callbacks, the onFrame pump, and incremental GC — gates on a single 50 ms `WM_TIMER` heartbeat, and `GetMessageW` blocks while idle, so deferred work (including promise jobs queued by timers) can wait up to 50 ms and `onFrame` runs as a ~20 fps piggyback rather than vsync-aligned. The code itself concedes the follow-up: "Real DwmFlush-aligned vsync driver is a follow-up" (#1865). Found in the 2026-07-18 Windows audit (main @ 58e555e58).

## Evidence

- `crates/perry-ui-windows/src/app.rs:399` — `SetTimer(..., TICK_TIMER_ID, 50, None)`.
- `crates/perry-ui-windows/src/app.rs:448-520` — message loop; the tick block services `js_callback_timer_tick` / `js_interval_timer_tick`, `js_frame_pump_default`, and `js_gc_step_us`.
- `crates/perry-ui-windows/src/app.rs:488-506` and `:493-496` — onFrame piggyback + DwmFlush-follow-up comment.

## Suggested fix

Replace the blocking `GetMessageW` + fixed tick with an `MsgWaitForMultipleObjectsEx`-based pump whose timeout is the next-due timer deadline (wake immediately when JS work is queued), and drive `onFrame` from a DWM-aligned source (a `DwmFlush` thread posting frame messages, or CompositionTarget) instead of the heartbeat.

PR #8022, opened 2026-08-13
fix(windows): preserve current Rust UI codegen units

proggeramlug:
## Summary
- recognize current rustc `.rcgu.o` archive-member names when rebuilding Windows UI libraries
- keep allocator/non-codegen members excluded while extracting every UI codegen unit
- extend the COFF dedup regression fixture with a sibling `.rlib` whose member uses the current naming shape

## Root cause
While validating the Windows event-pump fix for #6617, a real `perry/ui` compile failed with undefined `perry_ui_*` exports. Current rustc emits members such as `perry_ui_windows-<hash>.<opaque>.<hash>.rcgu.o`; the trimmer only recognized names containing literal `.cgu.` or `-cgu.`, so it skipped all 16 UI objects after using their symbols to remove the staticlib copies.

## Validation
- `cargo test --profile perry-dev -p perry --bin perry coff_archive_dedup_drops_only_fully_provided_members -- --nocapture`
- `cargo test --profile perry-dev -p perry --bin perry rust_codegen_unit_recognizes_current_and_legacy_member_names`
- `git diff --check`
- rebuilt `perry-ui-windows` for `x86_64-pc-windows-msvc`
- compiled a real TypeScript `perry/ui` app; the trimmer extracted 16/16 rlib CGUs and produced a 5.6 MB PE
- launched the resulting executable for two seconds; it remained alive in its Win32 message loop

No version bump.

### DIFF
--- a/changelog.d/8022-windows-ui-coff-exports.md
+++ b/changelog.d/8022-windows-ui-coff-exports.md
@@ -1,0 +1,7 @@
+### Windows UI links current Rust codegen archives again
+
+Windows `perry/ui` builds no longer lose every public `perry_ui_*` export
+while trimming duplicate COFF archive members. The trimmer now recognizes the
+opaque `.rcgu.o` member names emitted by current Rust toolchains, extracts all
+of the UI crate's codegen units, and continues to exclude non-codegen allocator
+members.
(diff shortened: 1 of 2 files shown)


proggeramlug (commented):
Audited exact commit e69558b927e2a400023e8ac2a668b8be779a3f63 against current main. The `.rcgu.o` suffix is the correct discriminator for current and legacy rustc codegen-unit members while excluding `lib.rmeta` and ordinary bundled native objects. The sidecar-r​lib fixture exercises the corrected extraction path, the host-level predicate regression passes, formatting and diff checks are clean, and the current-main merge tree is conflict-free. No blocking findings.

Issue #5094, opened 2026-06-13
perf(GC): make per-object layout O(1)-loadable — kill per-operation thread-local layout tracking (umbrella: method_calls/array-downgrade/object-property)


TheHypnoo:
## Summary

Several of Perry's worst benchmark gaps share **one** systemic root cause, not separate ones: Perry tracks **per-object / per-slot GC layout** (which slots hold raw-f64 numbers vs NaN-boxed pointers) in **thread-local hashmaps** (`TYPED_LAYOUTS`, `LAYOUT_SLOT_MASKS`) and **queries or updates them on every field/element access**. On macOS each access is a `_tlv_get_addr` (TLS accessor) + a hashmap hash+lookup. The all-numeric "unboxed" fast paths bypass it, but **any heterogeneous shape, downgraded array, or raw-f64 class field falls into the per-op hashmap path**, which dominates.

| Benchmark | Gap vs Node | Hot path | Per-op cost |
|---|---|---|---|
| `method_calls` (#5093) | ~290× (3300 ms vs 11 ms) | `this.field` get/set in class methods | `js_typed_feedback_class_field_{get,set}_guard` → `class_field_fast_contract` → `layout_typed_raw_f64_slot_for_user` (TLS `TYPED_LAYOUTS`) |
| `bench_numeric_array_downgrade` | ~781×; ~21× over the same-shape numeric array | `arr[i]=…` on heterogeneous/`any[]` arrays | `js_array_set_f64_extend` → `note_array_slot` → `layout_note_slot` (TLS), **per write** |
| `bench_object_property` | ~17× after #5084 | dynamic property writes | same field-access guard family |

## Evidence (measured)

- **Profile (`sample`)**: the downgrade hot loop is a storm of `_tlv_get_addr` (TLS) calls.
- **method_calls**: removing the per-access `register_site` call (#5084) → ~2%; an MRU cache over `layout_typed_raw_f64_slot_for_user` made it **2× SLOWER** (3300 → 7050 ms). Inlining the guard is blocked because the raw-f64 check is a TLS hashmap lookup, not an O(1) header field.
- **numeric_array_downgrade**: an `obj_type` pre-filter that skips the per-element typedarray/buffer/set/map registry lookups gave only ~5%; a "skip the in-bounds `RuntimeHandleScope` rooting" fast path made it **worse**. So the cost is `note_array_slot` → `layout_note_slot` (the TLS layout write), per element.

[...]

PR #8023, opened 2026-08-13
perf(gc): skip descriptor probes for plain typed stores

proggeramlug:
## Summary

- prove in-bounds plain-double stores conform to intact typed object layouts from the GC header and object field count
- skip both thread-local descriptor-map probes on that common dynamic-store path
- keep descriptor-based validation for tagged values, pointer-like raw words, and out-of-bounds notes
- add regression counters/tests that assert both the fast path and every safety fallback

Refs #5094.

## Performance

A 5,000,000-iteration dynamic-key numeric field-store benchmark on an otherwise idle arm64 macOS host improved from a 1,207 ms median to 1,156 ms: 1.044x, or 4.2% faster. Output was identical. The regression counter also shows the targeted descriptor probe dropping from one per store to zero.

## Validation

- focused layout suite: 13 passed
- full runtime suite: 2,286 passed and 4 ignored; the sole failure was an unrelated load-sensitive promise timing assertion on a host at load 74, and that exact test passed on the quiet benchmark host
- repeated numeric stores plus pointer fallback under default GC, forced evacuation with evacuation verification and from-space scanning, and non-generational GC

No version bump.

### DIFF
--- a/changelog.d/8023-typed-store-header-proof.md
+++ b/changelog.d/8023-typed-store-header-proof.md
@@ -1,0 +1,7 @@
+### Skip GC descriptor probes for plain typed stores
+
+In-bounds plain-double writes to intact typed objects now prove layout
+compatibility from the object header and field count, avoiding both
+thread-local descriptor-map probes on the common dynamic-store path. Tagged,
+pointer-like, and out-of-bounds writes retain the descriptor fallback and its
+existing downgrade behavior.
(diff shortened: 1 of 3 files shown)


proggeramlug (commented):
Audited exact head `9d75f89c7e6fbdc69a1b6de3c15954a7cca2c632`. The header-only path is limited to current-field-bound `GC_TYPE_OBJECT` slots and bits that are both raw-f64-compatible and not conservatively pointer-bearing. Tagged values, pointer-like raw words, non-object layouts, and out-of-bounds notes retain descriptor classification. Stored bits, layout notes outside this equivalence, and write barriers are unchanged. All 13 focused object/closure layout tests pass locally; `git diff --check` passes; the merge tree with current main is conflict-free. No blocker found.

PR #8024, opened 2026-08-13
fix(gc): isolate class heap registries per agent (#8001)

### LINKED ISSUE #8001: gc/thread: the class prototype/parent registries are process-global but hold per-realm heap addresses
Same family as #7988 / #7981 / #7954, and the worst member found so far.

## The cells

```rust
// crates/perry-runtime/src/object/class_registry/state.rs:316,325,381
static CLASS_PROTOTYPE_OBJECTS:      RwLock<Option<HashMap<u32, usize>>>
static CLASS_DECL_PROTOTYPE_OBJECTS: RwLock<Option<HashMap<u32, usize>>>
static CLASS_PARENT_CLOSURES:        RwLock<Option<HashMap<u32, usize>>>
```

The KEY is a `class_id` — a codegen-assigned constant, therefore identical on
every thread. The VALUE is the raw address of that class's `prototype` object,
allocated in **one thread's arena**.

## Why that is a bug under `perry/thread`

`perry/thread` gives every agent its own realm (`THREAD_GLOBAL_THIS`) and its
own arena. `crates/perry-runtime/src/object/class_registry/prototype_objects.rs:34`
is a first-thread-wins gate:

```rust
let existing = class_prototype_object(class_id);
if !existing.is_null() { return existing; }
```

so every thread after the first that instantiates class `C` gets back the FIRST
thread's `C.prototype` and links it as `[[Prototype]]` on its own freshly
allocated instance.

This is **worse than #7988**. #7988 only ever *compared* against a foreign
address; here a foreign pointer is **stored into a live object in this agent's
own heap graph**. It dangles the moment the owning thread exits (its arena
blocks are `dealloc`'d).

And `crates/perry-runtime/src/object/class_gc_roots.rs:44-63`
(`scan_class_inheritance_roots_mut`, registered per thread at
`crates/perry-runtime/src/gc/mod.rs:921`) does

```rust
for ptr in map.values_mut() { visitor.visit_usize_slot(ptr) }
```

on the **shared** map — so every agent's collector rewrites entries that may
name a different agent's heap, with its own to-space address. That is #7988's
consequence 3 on a table that every `new C()` reads.

## Also in scope (same file, not individually confirmed)

`CLASS_VTABLE_REGISTRY`, `CLASS_STATIC_METHODS`, `CLASS_STATIC_ACCESSORS`,
`CLASS_SYMBOL_METHODS`, `CLASS_SYMBOL_ACCESSORS`, `CLASS_DYNAMIC_PARENT_VALUE`,
`CLASS_OBJECT_VALUES` (`state.rs:241-407`) are all `class_id`-keyed alongside
these. `CLASS_DYNAMIC_PARENT_VALUE`'s `u64` in particular looks like it could be
a NaN-boxed JSValue, i.e. a heap pointer.

Ruled OUT in the same sweep: `CLASS_REGISTRY` / `PARENT_DENSE`
(`class_meta_registry.rs:11,47`) — key *and* value are codegen `u32` class ids, a
static program fact identical in every realm. Correct as process-global.

## Suggested acceptance

A `perry/thread` test in which agent B does `new C()` and observes its **own**
`C.prototype` — plus a perry-runtime unit test in the shape of
`a_second_agents_prototype_addresses_are_its_own` (#7994): two live agents,
bootstraps serialized, both held live across the comparison by a barrier, each
address asserted to be real *before* distinctness so the gate cannot be earned
by two agents that resolved nothing.

## Note on the expected obstacle

"We can't make this per-thread, Darwin has no local-exec TLS" is **stale** in
this runtime — `crate::perry_thread_local!` (#7469,
`crates/perry-runtime/src/tls_hot.rs`) resolves through the pthread TSD array
from `TPIDRRO_EL0`: an `mrs` plus two loads that LLVM CSEs, not an out-of-line
`_tlv_get_addr`. Measured in #7994: +1.3 % on a microbenchmark that does nothing
but property set/get, ~0.6 ns per accessor call.

proggeramlug:
## Summary

Isolates class-registry entries that contain heap pointers, NaN-boxed values, closure identities, or realm-created Symbol keys per Perry agent. This prevents a later perry/thread agent from reusing another realm heap values and prevents one moving collector from rewriting another agent roots.

## Changes

- Move pointer/value-bearing class side tables to Perry hot thread-local storage.
- Keep process-wide code metadata such as vtables, method function pointers, bind lengths, registered class ids, and the synthetic id allocator global.
- Make both full and incremental GC root scanners visit and rewrite only the current agent class roots.
- Add a two-live-agent Rust regression that proves each agent materializes a distinct non-null Class.prototype.
- Add a perry/thread behavioral probe covering prototype mutation isolation in both directions.

## Related issue

Closes #8001

## Test plan

- [ ] cargo build --release clean (not run for the full workspace)
- [ ] Full affected-crate suite (targeted runtime regression and compile checks run instead)
- [x] cargo test --release -p perry-runtime a_second_agents_declared_prototype_address_is_its_own -- --nocapture
- [x] cargo check -p perry-runtime --tests
- [x] cargo fmt --all -- --check
- [x] parity filter 8001: 1/1 expected-output tests passed
- [x] scripts/check_file_size.sh
- [x] scripts/global_sink_isolation.py
- [x] scripts/gc_runtime_root_holders.py
- [x] Added runtime and test-files regression coverage
- [x] Docs not needed: internal runtime ownership fix with no API change

The repository quick pre-tag suite otherwise passed, but its GC store-site inventory reports an existing marker omission at crates/perry-codegen/src/expr/property_set.rs:1457. That file is unchanged by this PR.

## Screenshots / output

The new parity probe prints:

    main: main
    spawn agent: undefined/agent/true match: true
    main after: main

## Checklist

- [x] My commit follows the fix: prefix convention

### DIFF
--- a/changelog.d/8001-agent-local-class-registries.md
+++ b/changelog.d/8001-agent-local-class-registries.md
@@ -1,0 +1,13 @@
+### Fixed
+
+- **Class prototype and dynamic-value registries are isolated per `perry/thread`
+  agent (#8001).** Tables keyed by process-wide codegen class ids previously
+  stored raw pointers, NaN-boxed values, closure identities, or realm-created
+  Symbol keys in process-global maps. The first agent to materialize a class
+  prototype could therefore make every later agent link instances to its heap,
+  and any agent's moving collector could rewrite entries owned by another
+  agent. Pointer-bearing class side tables now use Perry's hot thread-local
+  storage, while code-only metadata such as vtables, static method pointers,
+  bind lengths, and registered class ids remains process-global. A two-live-agent
+  runtime test and a `perry/thread` behavioral probe pin distinct, non-null
+  `Class.prototype` identities and realm-local prototype mutations.
--- a/crates/perry-runtime/src/object/class_gc_roots.rs
+++ b/crates/perry-runtime/src/object/class_gc_roots.rs
@@ -42,94 +42,118 @@ use super::class_registry::{
 /// stay as side-tables + rooted here rather than moved in-object, since the
 /// inheritance walk reads them by `class_id`, not by an in-object slot.
 pub fn scan_class_inheritance_roots_mut(visitor: &mut crate::gc::RuntimeRootVisitor<'_>) {
-    if let Ok(mut guard) = CLASS_PROTOTYPE_OBJECTS.write() {
-        if let Some(map) = guard.as_mut() {
-            for ptr in map.values_mut() {
-                visitor.visit_usize_slot(ptr);
+    CLASS_PROTOTYPE_OBJECTS.with(|table| {
+        if let Ok(mut guard) = table.write() {
+            if let Some(map) = guard.as_mut() {
+                for ptr in map.values_mut() {
+                    visitor.visit_usize_slot(ptr);
+                }
             }
         }
-    }
-    if let Ok(mut guard) = CLASS_DECL_PROTOTYPE_OBJECTS.write() {
-        if let Some(map) = guard.as_mut() {
-            for ptr in map.values_mut() {
-                visitor.visit_usize_slot(ptr);
+    });
+    CLASS_DECL_PROTOTYPE_OBJECTS.with(|table| {
+        if let Ok(mut guard) = table.write() {
+            if let Some(map) = guard.as_mut() {
+                for ptr in map.values_mut() {
+                    visitor.visit_usize_slot(ptr);
+                }
             }
         }
-    }
-    if let Ok(mut guard) = CLASS_PARENT_CLOSURES.write() {
-        if let Some(map) = guard.as_mut() {
-            for ptr in map.values_mut() {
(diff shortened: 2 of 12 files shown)


proggeramlug (commented):
Audited commit b2842e5bf1a3f4ae3d19b68ee5a34812fb82a734 against current main. The ownership split is correct: heap pointers, NaN-boxed values, closure identities, and realm Symbol keys are agent-local, while code-only metadata remains global. The full and incremental root paths now scan only the current agent. Independent verification: release two-agent prototype regression passed; class side-table mark/rewrite, copying-minor value/key rewrite, Symbol-key rewrite, inheritance-root rewrite, and foreign-clear isolation regressions all passed. global_sink_isolation.py, gc_runtime_root_holders.py, check_file_size.sh, diff check, and current-main merge-tree check are clean. No blocking findings.

PR #8025, opened 2026-08-13
perf(repsel): propagate return shapes across modules

proggeramlug:
## Summary

Propagate proven fresh anonymous-record return shapes across native module boundaries so callers can keep exact `Ptr<Shape>` representations instead of boxing at imports.

## Changes

- Harvest anonymous-record return-shape facts from every final-HIR module before parallel codegen.
- Resolve facts through exact import aliases and re-exports, then attach the proven class layout to consumer modules.
- Keep named classes, indirect calls, unknown externals, and barriered modules fail-closed.
- Include imported return-shape bindings in object-cache keys.
- Add unit and parity coverage for a renamed re-export, local import alias, field stores/loads, and collection-triggering churn.

## Related issue

Refs #7170

## Test plan

- [x] `cargo check --profile perry-dev -p perry-codegen -p perry`
- [x] `cargo test --profile perry-dev -p perry-codegen --lib ptr_shape_returns --no-fail-fast` — 37 passed
- [x] `cargo test --profile perry-dev -p perry --bin perry object_cache --no-fail-fast` — 46 passed
- [x] Filtered parity fixture — 1/1 byte-exact against Node
- [x] Filtered seven-arm GC stress matrix — 7/7 byte-exact; the shipped-default arm observed productive collections. The narrow moving-GC arms observed no moved objects, so this does not claim moving-object liveness coverage.
- [ ] Full affected-crate suite passes
- [x] Added or updated a test under `test-files/`
- [ ] Documentation update — not applicable; no CLI, configuration, stdlib, or runtime API change
- [ ] Platform UI build — not applicable

## Screenshots / output

The optimization report selects `Ptr<Shape>` for the imported result and records `class_field_get.shape_proven_load` plus `ptr_shape_set` consumers.

## Checklist

- [x] My commits follow the repository commit-prefix convention

### DIFF
--- a/changelog.d/8025-cross-module-return-shapes.md
+++ b/changelog.d/8025-cross-module-return-shapes.md
@@ -1,0 +1,11 @@
+### Representation selection: propagate fresh return shapes across modules (#7170 R2)
+
+Native compilation now carries proven fresh anonymous-record return shapes
+through exact import bindings and re-exports. Consumer modules can keep those
+results as `Ptr<Shape>` values and use guard-free fixed-offset field loads and
+stores instead of boxing at the module boundary.
+
+A final-HIR prepass establishes the facts before parallel codegen, and imported
+return-shape bindings participate in object-cache keys. Named classes, indirect
+calls, unknown externals, and modules with representation-selection barriers
+remain fail-closed.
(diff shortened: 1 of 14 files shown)


proggeramlug (commented):
Audited commit 44efba012080d025d7f2c56b9e8e1b2d357d1713 against current main, including the #8024 merge. The proof flow is sound and fail-closed: exported facts come from final HIR before parallel codegen; only fresh content-addressed anonymous records are admitted; re-exports resolve to the producer origin; consumers are keyed by the exact local ExternFuncRef binding; module barriers, named classes, indirect calls, and unknown externals remain excluded. The imported layout and binding fact travel together and participate in a deterministic object-cache key. Independent verification: ptr_shape_returns 37/37 passed, object_cache 46/46 passed, cargo fmt and diff checks passed, and the current-main merge tree is clean. No blocking findings.

Issue #6812, opened 2026-07-23
perf: generalize guarded object-write fast paths beyond the #6811 two-field loop


proggeramlug:
## Objective

Generalize the guarded object-write optimizations landed by #6811 beyond the
single #6759 acceptance-loop shape, without weakening JavaScript semantics or
GC safety.

#6811 completed #6809 and beat Node on the canonical write micro, but its
largest win is intentionally narrow. This issue tracks the remaining coverage
work: identify which real object writes still miss the static write PIC or the
whole-loop clone, rank those misses with measurements, and extend the sound
fast paths in small reviewable slices.

Start from current `origin/main`. The starting point must contain merge commit
`20f19758f370f7616530824218d87f46a67cd3f0` ("perf: beat Node on object-write
micro (#6809) (#6811)"). Do not resume from a pre-#6811 worktree.

## What #6811 established

Measured on the same macOS/ARM64 machine using 15 alternating runs:

| implementation | median for 9.6M existing-field writes |
|---|---:|
| Node.js | 7 ms |
| Perry before #6809 | about 16 s |
| Perry after runtime + per-site PIC work, before loop versioning | 39–40 ms |
| Perry after #6811 | **5 ms** |

The result is real, but it is not a blanket optimization for every
`object[key] = value`:

1. **Broad runtime improvements**
   - Header-first receiver classification.
   - Existing-own-data overwrite routing.
   - Builtin descriptor installs no longer unnecessarily poison process-wide
     user-descriptor gates.
   - Correct prototype/descriptor/array invalidation handling.

2. **Guarded static-key write PIC**
   - `crates/perry-codegen/src/expr/proxy_reflect.rs`
     (`lower_put_value_static_write_ic`)
   - Runtime miss/priming:
     `crates/perry-runtime/src/proxy/put_value.rs`
     (`js_put_value_set_ic_miss`)
   - Currently requires:
     - a literal string key;
     - target and receiver to be the same expression;
     - a safepoint-free/call-free RHS;
     - an eligible regular existing-own-data receiver;
     - one matching shape token (monomorphic cache);
     - inline, in-bounds storage and all mutable descriptor/frozen/typed-layout
       guards to pass.

3. **Whole-loop numeric object-array clone**
   - Matcher/emitter:
     `crates/perry-codegen/src/stmt/loops.rs`
     (`match_object_array_write2_loop`,
     `lower_object_array_write2_versioned_for`)
   - Runtime preflight:
[...]

PR #8026, opened 2026-08-13
perf: settle eight-shape static write PICs

proggeramlug:
## Summary

- retain the four generated static write-PIC ways and add four non-evicting ways in one outlined helper
- stop the fourth inline entry from thrashing once a site observes more than four stable receiver shapes
- add reproducible eight-shape benchmark coverage plus frozen-tail and ninth-shape semantic parity cases

Part of #6812.

## Measurements

On the new 60M-write `shape_eight` cell (15 alternating Node/Perry runs, identical checksums):

| | Node median | Perry median |
|---|---:|---:|
| four-way baseline | 234 ms | 2,359 ms |
| four inline + four outlined | 226 ms | **768 ms** |

That is a **67.4% reduction / 3.07x speedup** for Perry. `shape_monomorphic` stayed at 119 ms; `shape_four` moved from 400 to 408 ms. Both linked matrix executables are 5,849,320 bytes.

## Validation

- `git diff --check`
- `cargo clippy -p perry-codegen -p perry-runtime --lib --no-deps`
- `cargo test -p perry-runtime --lib` (2,281 passed, 4 ignored)
- static-write native proof tests pass
- release build of compiler/runtime/stdlib static archives
- focused release corpus matches Node exactly, including outlined frozen and ninth-shape cases

The full native proof executable reports 262/263 passing; the remaining unrelated integer-modulo case reproduces alone as an existing debug-overflow panic in `collectors/loop_bounded_i32.rs:1091`.

No version bump is included.

## Summary by CodeRabbit

* **Performance**
  * Improved property-write performance across up to eight distinct object shapes.
  * Added bounded handling for additional shapes while preserving standard write behavior.

* **Bug Fixes**
  * Ensured writes to frozen objects continue to raise the expected errors.
  * Preserved correct field values and checksums across polymorphic write scenarios.

* **Tests**
  * Added coverage for eight- and nine-shape write loops, including fallback behavior.

* **Documentation**
  * Documented the updated performance characteristics and bounded shape handling.

### DIFF
--- a/benchmarks/object-write-6812/matrix.ts
+++ b/benchmarks/object-write-6812/matrix.ts
@@ -219,6 +219,30 @@ function shapeFour(): CellResult {
   return { elapsed, writes: 60000000, sink: checksum(objects, ["x"]) };
 }
 
+function shapeEight(): CellResult {
+  const objects: any[] = [];
+  for (let i = 0; i < 2400; i++) {
+    const kind = i & 7;
+    if (kind === 0) objects.push({ x: i, a: 0 });
+    else if (kind === 1) objects.push({ x: i, b: 0 });
+    else if (kind === 2) objects.push({ x: i, c: 0 });
+    else if (kind === 3) objects.push({ x: i, d: 0 });
+    else if (kind === 4) objects.push({ x: i, e: 0 });
+    else if (kind === 5) objects.push({ x: i, f: 0 });
+    else if (kind === 6) objects.push({ x: i, g: 0 });
+    else objects.push({ x: i, h: 0 });
+  }
+  const t0 = Date.now();
+  for (let r = 0; r < 25000; r++) {
+    for (let i = 0; i < 2400; i++) {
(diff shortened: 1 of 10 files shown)


proggeramlug (commented):
Audited commit aa57c008fa6ac491945bf2b977ce25103a203587 against current main. The static write PIC remains bounded: four generated ways plus four non-evicting outlined ways, with shape 9+ taking full PutValue semantics. The helper revalidates the live receiver type, forwarding and mutable blocking flags, current shape token, and slot bounds before using the canonical barrier/layout-aware field store; frozen/descriptor/shape-change cases miss safely. The tail cache is a distinct zero-initialized eight-word global and its ABI matches the runtime declaration. Independent verification: focused static PutValue proof tests 2/2 passed, related string-key selection tests 7/7 passed, cargo check -p perry-runtime, cargo fmt, diff check, and current-main merge-tree check passed. No blocking findings.

Issue #5497, opened 2026-06-21
codegen: type-directed specialization — emit native ops where the type is statically known (#5334 Tier 3 / Lever E)


proggeramlug:
Follow-on to #5334. That issue tracked making perry's IR efficient for large/untyped modules; its Tier 1 (outline the dynamic machinery) and Tier 2 (emit less scaffolding) levers all shipped:

- Lever A — outline field-SET guard-miss arm (#5351)
- Lever B — full-outline IC diamond for oversized modules, SET + GET (#5385, #5410)
- Lever C — fold nan-box bitcast round-trips at emit time (#5350)
- Lever D — elide GC write-barrier for non-pointer field stores (#5381)
- (extra) outline array-literal construction (#5412)

This issue carries the one remaining lever, **Tier 3 / Lever E: type-directed specialization** — the strategic, long-term piece that #5334 explicitly deferred.

## Goal

Where a value is statically a number / string / known-shape (real `.ts`, `tsgo`, or local inference), emit native ops instead of the dynamic form:

- `fadd` / `fmul` / `fcmp` instead of `js_add` / `js_rel_lt` / IC guards
- direct slot load/store instead of the inline-cache diamond
- skip the nan-box round-trip entirely when the value never leaves a typed register

This attacks the two largest line-counts from #5334's measurement at the source — the **4.4M `call @js_*`** and **365K `js_typed_feedback_*_guard`** sites — rather than outlining them after the fact.

## Why separate from #5334

- It is a **performance win**, not just a size win — typed programs get both smaller and faster.
- It is **orthogonal to minified bundles**: minified JS is `Any` almost everywhere, so `static_type_of` rarely resolves, and Lever E barely moves the needle there. The bundle-compilability goal of #5334 is already met by Tiers 1–2.
- It is a larger, architectural change (type propagation through HIR → codegen) that deserves its own tracking and sequencing.

## Sketch

1. Thread a static-type signal to codegen op-emit sites (extend the existing `static_type_of` / `class_field_declared_type` machinery).
2. Number/boolean arithmetic + comparisons → native `f64` ops, guarded only where the static type is `number | undefined`-ish.
3. Known-shape field access → direct slot, bypassing the IC diamond.
4. String concat / comparison on statically-`string` operands → direct runtime calls without the dynamic dispatch.
[...]

PR #8031, opened 2026-08-13
perf(codegen): lower proven booleans into native numeric ops

proggeramlug:
## Summary

- convert guard-proven Boolean `i1` literals and locals directly to `f64` with `uitofp` for arithmetic consumers
- lower proven Boolean/number relational comparisons to native `fcmp` instead of `js_rel_*`
- keep annotation-only and invalidated Boolean facts on dynamic coercion/dispatch, preserving `as any` semantics
- add IR-census coverage plus Boolean arithmetic parity coverage

This is a focused Lever E slice that advances #5497; it does not close the broader tracker.

## Validation

- `cargo test -p perry-codegen boolean_number_tests --lib`
- `cargo test -p perry-codegen --lib` (967 passed)
- `./scripts/test_affected_crates.sh --base origin/main`
- Node 26.5.1 parity: `test_bool_arithmetic` (1/1 passed)
- `./scripts/pre-tag-check.sh --quick` (all relevant checks passed; the run reports the existing GC inventory ratchet at unchanged `crates/perry-codegen/src/expr/property_set.rs:1457`)

No version files were changed. The required PR-numbered changelog fragment is included.

## Summary by CodeRabbit

* **New Features**
  * Improved arithmetic and comparison handling for statically proven Boolean values.
  * Boolean values can now convert directly to numbers for supported numeric operations.
  * Added optimized Boolean-to-number relational comparisons.

* **Bug Fixes**
  * Preserved dynamic coercion and string concatenation when Boolean values are invalidated or cannot be safely proven numeric.

* **Tests**
  * Added coverage for Boolean arithmetic, comparisons, nested expressions, and invalidated Boolean annotations.

* **Documentation**
  * Updated the type-lowering checklist and changelog with the new behavior.


proggeramlug (commented):
Blocking on exact head 7a1f654. I reproduced a semantic regression with a focused IR test: let flag = true; let liar: number = "4" as any; flag < (liar + 1). JavaScript evaluates this as true because the RHS is "41". This head emits a raw fadd on the NaN-boxed string followed by fcmp olt, producing false, and omits js_rel_lt. The new relational gate accepts the nested Add because expr_produces_canonical_raw_f64 blanket-admits Binary via type-level is_numeric_expr, even though a declared-only Add tree can carry a boxed/string result. Please make the Boolean gate reject declared-only/noncanonical nested binary trees and add this regression. This is independent proof from emitted IR, not a CI result. No version bump needed.

proggeramlug:
Addressed in 490958af0.

The fix makes the existing proof boundary explicit in both places involved:

- a numeric local initialized by a non-numeric value (for example `const liar: number = "4" as any`) now carries the declared-only hazard;
- `expr_produces_canonical_raw_f64` rejects binary trees carrying that hazard, so the Boolean relational fast path cannot admit the nested `+`.

I added the exact `flag < (liar + 1)` IR regression. It now requires both `js_dynamic_string_or_number_add` and `js_rel_lt`, while the positive `flag < 2` case still requires native `uitofp` + `fcmp`.

Validated with all 968 perry-codegen library tests, the affected-crate sweep, and the optimized `test_bool_arithmetic` parity fixture (1/1 pass). No version bump.

proggeramlug (commented):
Re-audited exact head 490958af against the previously demonstrated nested declared-only numeric regression. The corrected IR retains both js_dynamic_string_or_number_add and js_rel_lt for a Number-annotated local holding a string, while the positive proven-boolean case retains native fcmp. The new declared-only seed and canonical-f64 rejection are aligned, the focused boolean-number tests pass, and the full perry-codegen library suite passes (968/968). No blocking issue found.

PR #8041, opened 2026-08-13
fix(gc): install array growth forwarding for low-address arenas

### LINKED ISSUE #8035: [GC/dylib] Install array-growth forwarding for low-address macOS arena allocations
## What happened?

Loading the Perry-compiled Next fixture as an app-only dylib on macOS emits this once per process during preload:

```text
[perry-gc] array-growth forwarding stub SKIPPED for an array at 0x...: address is below this platform's heap floor. Stale pre-grow array references will no longer resolve through the growth chain (issue #233).
```

This is a correctness warning, not log noise: once growth moves the array, a stale pre-grow reference can stop resolving to the new head.

Current `main` contains an internally inconsistent platform guard:

- `crates/perry-runtime/src/array/header.rs::clean_arr_ptr` explicitly includes `target_os = "macos"` in the 4 KiB floor and documents real macOS allocations below 2 TiB;
- `crates/perry-runtime/src/array/push_pop.rs::js_array_grow` omits macOS from the same low-floor cfg, uses 2 TiB, and therefore skips the forwarding stub for valid low-address macOS arena arrays.

The shared-library process shape makes this easy to hit during Next module preload.

## Reproduction app (provided)

Use the exact production Next app attached to #8034; its first Perry/shared-library preload is the integration reproducer. No new Next app is required.

Build/load it as specified there: shared runtime provider, shared stdlib provider, then app-only dylib with eager relocation. Capture stderr from process start through the 21-request verifier.

Also add a deterministic runtime-level regression that supplies a valid GC arena below 2 TiB (or injects the arena membership classifier) and grows an array past its initial capacity while retaining a pre-grow reference.

## Required fix

Use one canonical "is this a live Perry GC allocation?" classifier for both installing and following growth forwarding stubs. At minimum the duplicated platform floors must agree, but the preferred invariant is allocator/arena membership plus validated `GcHeader` / `obj_type`, not address magnitude alone.

The fix must continue rejecting registry handles, synthetic non-GC pointers, and unrelated allocations before dereferencing a header. See #7531 for the address-band audit precedent.

## Acceptance criteria

- A valid low-address macOS GC array receives a growth forwarding stub.
- A retained pre-grow reference observes length and elements after growth past 16 and after at least two further growths.
- The test passes with forced evacuation / GC verification and detects forwarding cycles or invalid targets.
- Registry-handle band boundaries and non-arena pointers are rejected without a dereference.
- #8034 preloads and completes with no `array-growth forwarding stub SKIPPED` warning.
- Linux, Windows, iOS-family, and normal high-address macOS tests remain green.

## Baseline / references

- Production integration fixture: #8034
- Original stale array/reference failure: #233
- Related stale-head regression: #7661
- Address classification precedent: #7531

proggeramlug:
## Summary

- add a canonical allocator-membership gate that validates arena/malloc ownership and `GcHeader` metadata before dereferencing
- return a checked `NonNull<GcHeader>` instead of manufacturing a shared static reference, so forwarding metadata is mutated without a shared-to-mutable cast
- retain the classifier result across normal array access and reclassify only when a forwarding hop advances the address
- enforce successful growth-forwarding installation before returning the replacement array
- deterministically force a copying minor, read the rooted live handle afterward, prove evacuation moved it, and prove the original three-growth pointer resolves to the relocated head
- leave package versions and lockfiles unchanged; add the required PR-keyed changelog fragment

## Validation

- `RUST_TEST_THREADS=1 cargo test -p perry-runtime --lib array::tests -- --nocapture` (74 passed)
- `RUST_TEST_THREADS=1 cargo test -p perry-runtime --lib value::addr_class::tests -- --nocapture` (8 passed)
- `RUST_TEST_THREADS=1 cargo test -p perry-runtime --lib array::tests::stale_array_reference_survives_three_growths_and_forced_minor_gc -- --exact --nocapture`
- `git diff --check`
- `git merge-tree --write-tree HEAD origin/main`

Closes #8035

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved array growth and garbage-collection handling across repeated expansions and minor collection cycles.
  * Preserved access to arrays through stale references after growth.
  * Added safeguards against invalid, unsupported, or cyclic forwarding references and unsafe memory access.
  * Improved support for low-address arrays, registered buffers, typed arrays, and lazily materialized arrays.
  * Added validation to ensure forwarding targets belong to the appropriate allocation.

* **Tests**
  * Added coverage for repeated growth, forwarding behavior, low-address allocations, stale references, and invalid targets.

### DIFF
--- a/changelog.d/8041-array-growth-forwarding.md
+++ b/changelog.d/8041-array-growth-forwarding.md
@@ -1,0 +1,1 @@
+**Fixed: growing arrays at valid low arena addresses now retain stale-reference forwarding across repeated growth and moving minor GC (#8035).** Array header access now requires allocator ownership before dereferencing, rejects invalid forwarding targets, and installs growth forwarding through a checked raw-header pointer.
--- a/crates/perry-runtime/src/array/header.rs
+++ b/crates/perry-runtime/src/array/header.rs
@@ -584,13 +584,14 @@ pub(crate) fn clean_arr_ptr(arr: *const ArrayHeader) -> *const ArrayHeader {
     // practice (1-2 grows) but cap depth at 64 to defend against
     // cycles from corrupted GC state.
     let mut cleaned = cleaned;
+    let mut tracked_header =
+        unsafe { crate::value::addr_class::try_read_tracked_gc_header(cleaned as usize) };
     unsafe {
         let mut steps = 0u32;
-        while let Some(gc_header) =
-            crate::value::addr_class::try_read_tracked_gc_header(cleaned as usize)
-        {
-            if gc_header.obj_type != crate::gc::GC_TYPE_ARRAY
-                || gc_header.gc_flags & crate::gc::GC_FLAG_FORWARDED == 0
+        while let Some(gc_header) = tracked_header {
+            let gc_header = gc_header.as_ptr();
+            if (*gc_header).obj_type != crate::gc::GC_TYPE_ARRAY
+                || (*gc_header).gc_flags & crate::gc::GC_FLAG_FORWARDED == 0
             {
                 break;
             }
@@ -600,10 +601,11 @@ pub(crate) fn clean_arr_ptr(arr: *const ArrayHeader) -> *const ArrayHeader {
             else {
                 return std::ptr::null();
             };
-            if target_header.obj_type != crate::gc::GC_TYPE_ARRAY {
+            if (*target_header.as_ptr()).obj_type != crate::gc::GC_TYPE_ARRAY {
                 return std::ptr::null();
             }
             cleaned = new_user as *const ArrayHeader;
+            tracked_header = Some(target_header);
             steps += 1;
             if steps > 64 {
                 return std::ptr::null();
@@ -618,9 +620,10 @@ pub(crate) fn clean_arr_ptr(arr: *const ArrayHeader) -> *const ArrayHeader {
     // pointer for every downstream accessor. O(1) on subsequent
     // calls (idempotent via the `materialized` cache).
     let addr = cleaned as usize;
-    let tracked_header = unsafe { crate::value::addr_class::try_read_tracked_gc_header(addr) };
-    if let Some(gc_header) = tracked_header {
-        if gc_header.obj_type == crate::gc::GC_TYPE_LAZY_ARRAY {
+    let tracked_obj_type =
+        tracked_header.map(|gc_header| unsafe { (*gc_header.as_ptr()).obj_type });
+    if let Some(obj_type) = tracked_obj_type {
+        if obj_type == crate::gc::GC_TYPE_LAZY_ARRAY {
             unsafe {
                 let lazy = cleaned as *mut crate::json_tape::LazyArrayHeader;
                 if (*lazy).magic == crate::json_tape::LAZY_ARRAY_MAGIC {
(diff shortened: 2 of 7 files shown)


proggeramlug (commented):
Audited exact head 20b5a8a281386a08342c0ec8fca51021599c2e84 independently of CI. The allocator-membership direction is sound and the focused behavior passes, but this head is not safe to merge yet.

Blocking:

1. try_read_tracked_gc_header creates and returns &'static GcHeader. install_array_growth_forwarding_with then obtains a mutable pointer with std::ptr::from_ref(header).cast_mut() and set_forwarding_address writes the payload and gc_flags through it. GcHeader has no UnsafeCell, so the write occurs behind a live shared reference and violates Rust's aliasing contract. Allocator ownership proves the address, not mutable provenance. Return a raw/NonNull<GcHeader> (or a dedicated tracked-header pointer wrapper) from the classifier, perform reads through that pointer, and pass the legitimate mutable pointer directly to set_forwarding_address. Update tests so they do not create &'static aliases and cast them mutable.

2. The forced-evacuation regression does not prove its stated post-GC invariant. _root is never read, head is not re-read from the handle, and the test never asserts that evacuation moved the live head. Continuing through local initial can pass while reset from-space bytes remain physically unoverwritten. After gc_collect_minor(), read the rooted handle, assert movement under PERRY_GC_FORCE_EVACUATE=1, and assert clean_arr_ptr_mut(initial) resolves to that live rooted head before checking contents (ideally allocate after collection to prevent a stale-byte false positive).

Please also avoid two preventable degradations while fixing this boundary:
- clean_arr_ptr classifies an ordinary non-forwarded array in the forwarding loop and then unconditionally classifies the unchanged address again for lazy/type validation. Carry the validated pointer/header forward; this helper is on every array length/get/set path.
- js_array_grow discards the install boolean. For an array already accepted by clean_arr_ptr, a decline recreates #233 silently. Capture it and debug_assert (or otherwise enforce the invariant).

Independent exact-head validation:
- array::tests:: — 74 passed, serial
- value::addr_class::tests:: — 8 passed, serial
- forced-evacuation stale-reference command — 1 passed
- addr_class_inventory.py — audit passed (it reports two pre-existing/stale ratchet counts to lower)
- cargo fmt --check and git diff --check — passed
- merge-tree with current origin/main — clean

No version bump is needed. I have not merged this head.

proggeramlug:
Addressed the source-level review blockers in `7cfea9c4b`:

- the tracked classifier now returns `NonNull<GcHeader>` and no shared reference is cast mutable;
- `clean_arr_ptr` reuses the retained classification on ordinary access;
- growth forwarding installation is asserted before the new array can be returned;
- the GC regression activates the copying nursery with the runtime test guards, registers the runtime-handle scanner, proves the rooted live head moved, and checks the stale three-stub chain resolves to that exact address.

Also removed the unreachable size bound, added the boxed unrelated-allocation case, clarified the injected-header test, and added the required changelog fragment without a version or lockfile bump. Focused runtime suites, formatting, inventory, file-size, diff, non-test build, and merge-tree checks pass locally with `RUST_TEST_THREADS=1` for runtime tests.

proggeramlug (commented):
Re-audited exact head 7cfea9c4b5da6ea14dd37a27f7d4958cd89cd458 independently of CI. All four source-level blockers from my prior review are resolved: tracked classification now returns NonNull<GcHeader> without manufacturing a shared static reference; clean_arr_ptr retains and reuses the validated header on the unchanged hot path; growth cannot return a replacement array unless forwarding installation succeeds; and the forced-copying regression reads the rewritten runtime handle, proves the live head moved, and proves the original three-growth pointer resolves to that exact address. The classifier still rejects handle boundaries and unrelated stack/heap allocations before header dereference, and the low-address ownership policy is covered through injected arena membership. Independent exact-head validation passed: the forced movement regression, 74 serial array tests, 8 serial address-classifier tests, production perry-runtime cargo check, formatting, address inventory, file-size and diff checks, and a clean merge-tree with current main. The inventory reports two stale baselines that can be lowered, not a new failure. Existing compiler warnings only. No version bump; the changelog fragment is present. I found no remaining source-level blocker and consider this head mergeable.

Issue #8036, opened 2026-08-13
[Next.js/dylib] Preserve NextRequest and nextUrl.searchParams across App Route imports


proggeramlug:
## What happened?

The Perch Next.js lower-bound fixture on Perry `0.5.1503` could construct a `NextRequest` in one module and call an imported route function, but could not rely on the callee receiving a usable request argument. `NextRequest.nextUrl.searchParams` was also not reliable through that path, so the benchmark had to ignore the request and hard-code its workload.

This remains a distinct end-to-end gap even though important lower-level fixes have landed: #7252 roots cross-module call arguments and #6738 backs `URLSearchParams` subclasses. The missing regression is the actual production Next request shape across the generated App Route/import boundary.

## Reproduction app (provided)

Use the exact pinned app and verifier attached to #8034. Do not create a new app. Its route is deliberately re-exported across a module boundary:

```text
AppRouteRouteModule.handle
  -> app/api/benchmark/route.ts
  -> lib/route-impl.ts::GET/POST
```

Build with:

```bash
bash make-next-app-route-fixture.sh
cd perry-next-app-route-fixture
npm install --package-lock-only
npm ci
npm run build
```

The focused request is:

```bash
curl -i \
  -H 'x-request-id: request-7' \
  'http://127.0.0.1:3100/api/benchmark?id=request-7&iterations=7'
```

The returned JSON must contain:

```json
{
  "method": "GET",
  "pathname": "/api/benchmark",
  "id": "request-7",
  "iterations": 7,
  "beforeAwait": "request-7",
  "afterAwait": "request-7",
  "requestBody": ""
}
```

The attached verifier also sends a POST and requires `requestBody === "perry-request-body"`.

## Required implementation behavior

- Preserve the `NextRequest` object passed by the generated App Route module into the re-exported user handler.
- Preserve method, URL, `nextUrl`, `ReadonlyURLSearchParams`, headers, and body state across the imported call.
- Keep the request live across allocations, dynamic import, promise continuation, timer continuation, and GC movement.
- Do not special-case the fixture or lower `NextRequest` to a partial plain object.
- Keep direct calls, re-exports, sync handlers, and async handlers on the same correct path.

## Regression coverage

Add a small compiler fixture for a request-shaped class passed through a re-exported imported function, plus the production integration test from #8034. Run the small fixture with forced evacuation / GC verification so a stale-but-lucky pointer cannot pass.

## Acceptance criteria

[...]

Issue #8034, opened 2026-08-13
[Next.js/dylib] Add a pinned production App Route parity fixture and CI gate


proggeramlug:
## Summary

Add one pinned, production-built Next.js App Route fixture that is the release gate for Perry's shared-library embedding path. The fixture is included below in full so the implementer does not need to design an app or decide what to assert.

This closes the gap between "Perry can compile selected Next userland code" and "Perry can execute the production Next App Route pipeline as an app-only dylib with shared runtime/stdlib providers."

## Baseline

- Next.js `16.3.0`, built with `next build --webpack`
- React / React DOM `19.2.4`
- Perry `0.5.1503`, commit `564c56308d221a51b50308d9165578fbb176e877`
- App library built with `--output-type dylib`; runtime and stdlib are separate provider images
- No source rewrite that calls the userland `GET` directly, fabricates the response, or bypasses `AppRouteRouteModule.handle`

Related groundwork already landed in #5438, #6738, and #7252. The broad Node async-context backlog is #6764.

## Attached fixture generator

Save this block as `make-next-app-route-fixture.sh` and run it. It creates the complete application and its 21-request concurrency verifier; no `create-next-app` prompts or app design are required.

<details>
<summary><code>make-next-app-route-fixture.sh</code></summary>

```bash
#!/usr/bin/env bash
set -euo pipefail

app_dir="${1:-perry-next-app-route-fixture}"
mkdir -p "$app_dir/app/api/benchmark" "$app_dir/lib"

cat >"$app_dir/package.json" <<'JSON'
{
  "name": "perry-next-app-route-fixture",
  "version": "1.0.0",
  "private": true,
  "scripts": {
    "build": "next build --webpack",
    "start": "next start",
    "verify": "node verify.mjs"
  },
  "dependencies": {
    "next": "16.3.0",
    "react": "19.2.4",
    "react-dom": "19.2.4"
  },
  "devDependencies": {
    "@types/node": "25.3.3",
    "@types/react": "19.2.14",
    "@types/react-dom": "19.2.3",
    "typescript": "5.9.3"
  }
}
JSON

cat >"$app_dir/next.config.ts" <<'TS'
import type { NextConfig } from "next";

const nextConfig: NextConfig = {
  output: "standalone",
};

export default nextConfig;
TS

cat >"$app_dir/tsconfig.json" <<'JSON'
{
  "compilerOptions": {
    "target": "ES2022",
    "lib": ["dom", "dom.iterable", "es2022"],
[...]

PR #8044, opened 2026-08-13
fix(gc): preserve request graphs across route imports

proggeramlug:
## Summary

- exclude a module own exported classes from the foreign polymorphic-class fallback, preserving native-backed derived constructor initialization
- root imported exported-const closures while request arguments are evaluated
- root the complete detached expired-timer batch across callback dispatch and microtask checkpoints

## Regression coverage

- add a generated-route-shaped re-export fixture with a NextRequest-shaped class, URL, readonly URLSearchParams subclass, Headers, and async body state
- exercise sync and async handlers through allocation, dynamic import, promise and timer continuations, and explicit moving GC
- mirror the production verifier with 20 concurrent GET requests plus one POST body request

## Testing

- cargo test -p perry-runtime test_timer_tick_roots_the_complete_detached_expired_batch -- --nocapture
- cargo test -p perry-codegen --lib
- cargo fmt --all -- --check
- python3 scripts/check_test_registration.py
- Node oracle vs Perry under normal GC: byte-identical, 22 output lines
- Node oracle vs Perry under forced and verified evacuation: byte-identical; 6,601 live objects copied

Refs #8036

### DIFF
--- a/changelog.d/8044-next-request-import-rooting.md
+++ b/changelog.d/8044-next-request-import-rooting.md
@@ -1,0 +1,6 @@
+### Fixed
+
+- Preserve request objects and their nested URL, search-parameter, header, and
+  body state when generated route modules re-export imported handlers. Imported
+  handler closures and complete expired-timer batches now remain rooted across
+  allocation, promise/timer continuations, microtask checkpoints, and moving GC.
--- a/crates/perry-codegen/src/lower_call/extern_func.rs
+++ b/crates/perry-codegen/src/lower_call/extern_func.rs
@@ -1688,26 +1688,48 @@ pub fn try_lower_extern_func_call(
     // an arrow-bound exported value (hono's `mergePath` from utils/url.js,
     // any `export const foo = () => …` cross-module use).
     if ctx.imported_vars.contains(name) {
-        ctx.pending_declares.push((fname.clone(), DOUBLE, vec![]));
-        let closure_box = ctx.block().call(DOUBLE, &fname, &[]);
-        let mut lowered_args: Vec<String> = Vec::with_capacity(args.len());
-        for a in args {
-            lowered_args.push(lower_expr(ctx, a)?);
-        }
-        if lowered_args.len() > 16 {
+        if args.len() > 16 {
             anyhow::bail!(
                 "perry-codegen Phase D.1: closure call with {} args (max 16)",
-                lowered_args.len()
+                args.len()
             );
         }
-        let blk = ctx.block();
-        let closure_handle = unbox_to_i64(blk, &closure_box);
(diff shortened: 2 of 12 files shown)


proggeramlug (commented):
Audited exact head `42bccf62b05911446c614c6c72103d850cfcaac5`.

The three source fixes look well-scoped: imported exported-const closures are rooted during argument evaluation, timer ticks root the complete detached expired batch across callbacks/microtasks, and a module's own exported classes are excluded from foreign polymorphic-class augmentation. The focused timer test and the new normal and forced/verified-evacuation parity fixture pass locally.

I am holding the merge because this head says `Closes #8036` but does not satisfy that issue's required acceptance gate. #8036 requires the pinned #8034 production `AppRouteRouteModule.handle` integration, including 20 concurrent GETs and POSTs, normal and forced evacuation, and no direct-handler fallback. The added request-shaped compiler fixture is synthetic and does not run that production route-module path.

Please either add/run the exact production integration gate, or narrow this to a non-closing partial fix. I can re-audit the updated exact head.

proggeramlug:
Independent exact-head audit complete for `42bccf62b05911446c614c6c72103d850cfcaac5` against current main `a9a99d8b7e8d3e2d3bd35d8725a34a4cab403f97`.

- Audited the full source diff and clean landing merge; no unresolved review threads.
- `cargo test -p perry-codegen --lib`: 987/987 passed.
- Focused detached-timer-batch movement regression: 1/1 passed.
- `cargo fmt --all -- --check` and registration checker (207 files) passed.
- Fresh landing-equivalent compiler/runtime/stdlib archives were built and pinned with `PERRY_RUNTIME_DIR`.
- Node, normal Perry, and forced+verified Perry were byte-identical: 22 lines, SHA-256 `40cf14388bf4c1a6c5f64b57008653d159326b5fc61d57c51430cdc1cbd66ff5`.
- Forced run positively moved 6,601 live objects / 455,696 bytes with evacuation verification enabled.

CI was not used as regression proof. The implementation is sound and mergeable at this exact head. The corrected `Refs #8036` boundary is important: #8036 remains open until the explicit #8034 production `AppRouteRouteModule.handle` integration acceptance passes.

proggeramlug:
Addressed the production-gate request in #8082 at exact head `3434bc725af4341a9adeaa15b3076ca5a83aef48`.

The follow-up adds the pinned Next 16.3.0 production webpack fixture and exercises the generated `AppRouteRouteModule.handle` path as an app-only dylib with separate runtime/stdlib providers. The local gate passed 10 cold processes (five normal, five forced+verified evacuation), two verifier batches per process, 20 concurrent GETs plus one POST per batch: 20/20 batches and 420 Perry requests total, with zero forbidden fallback diagnostics. The Node production oracle also passed.

PR #8054, opened 2026-08-13
fix(codegen): keep emitted rodata symbols module-unique

proggeramlug:
## Summary

- allocate typed-parse, V8-interop, and node-submodule rodata names from the existing module-wide lowering-site counter
- cover the exact full-body `$spec_i32` plus boxed-body collision that redefined `@perry_typed_parse_keys_0`
- keep specialised ABI and schema-directed parsing enabled

## Evidence

Scheduled failure: [gc-parse-churn run 31737943472 / job 94574017539](https://github.com/PerryTS/perry/actions/runs/31737943472/job/94574017539), head `3381e1b89048592cfc91f3d4a43349c6e7c16015`.

Before this change, both emitted bodies restarted `FnCtx::typed_parse_counter` at zero and appended the same private global to one LLVM module. The regression now emits both bodies, requires two parse-schema globals, and asserts their symbol names are distinct.

Local verification (isolated target `/Users/amlug/cargo-targets/red-gc-gates-sub`):

- exact PR-code unit regression: `cargo test --profile perry-dev -p perry-codegen --test native_proof_regressions typed_parse_rodata_names_are_unique_across_spec_and_boxed_bodies -- --exact` — 1 passed
- raw product oracle from the exact #8054 change: the parse-churn subject compiled and ran with `BLOB_BYTES 245781`, `PARSED_LENGTH 4000`, `CHURN_TOUCH 240000`, `MISMATCHES 0`, exit 0, and a current-format non-in-place minor with `copied_objects=183`
- `cargo fmt --all -- --check`
- `git diff --check`

Scope note: the shipped #8054 wrapper alone still cannot print `PARSE_CHURN_LAYOUT_GATE_OK`, because its liveness parser only recognizes the obsolete copy-minor log format. The end-to-end wrapper success I observed used the separate #8049 parser fix in the committed audit tree. That parser change is intentionally excluded from this PR; the raw compile/output/exit and nonzero current-format collector evidence above validate this product fix without claiming the independent wrapper bug is fixed here.

The release archives and compiler were built together and their mtimes were verified before the end-to-end run.

Closes #8048.
Refs #7966.

PR #8055, opened 2026-08-13
fix(gc): keep realm-owned runtime roots agent-local

### LINKED ISSUE #8002: gc/thread: iterator / generator / %TypedArray% intrinsic towers are built once per PROCESS, not per realm
Same family as #7988 / #8001. One shape, several cells, one fix.

Each of these is a **once-per-process** `if PTR.load() == 0 { build… }` gate
whose result is a raw address of an object in the building thread's arena —
while the tower it names is a realm intrinsic and every `perry/thread` agent has
its own realm.

| cells | gate |
|---|---|
| `ITERATOR_PROTOTYPE_PTR`, `ARRAY_ITERATOR_PROTOTYPE_PTR`, `MAP_`/`SET_`/`STRING_`/`REGEXP_STRING_ITERATOR_PROTOTYPE_PTR` (`object/iterator_prototypes.rs:41-46`) | `ensure_iterator_prototypes()` (`:240`) |
| generator / async-generator intrinsic tower, 6 cells (`object/mod.rs:280-285`) | `ensure_generator_intrinsics()` (`object/global_this/generator.rs:716`) |
| `TYPED_ARRAY_INTRINSIC_PTR`, `TYPED_ARRAY_INTRINSIC_PROTO_PTR` (`object/mod.rs:256-257`) | `ensure_typed_array_intrinsic()` (`object/global_this/typed_array.rs:418`) |

`attach_iterator_prototype` (`iterator_prototypes.rs:270-288`) then calls
`chain_to(obj_ptr, proto_ptr)` on **every** newly allocated iterator on **every**
thread, using whichever thread built the tower first — so a foreign pointer is
stored into a live object in this agent's own heap, which dangles once the
building thread exits. Any `for…of` / spread / `Iterator.from` hits it, i.e. it
is reachable from about the most common operation in the language.

Two of these carry comments that are now false:

* `typed_array.rs:416` — *"Called from `populate_global_this_builtins`
  (single-threaded under the singleton CAS)"*. Every thread's first
  `js_get_global_this` runs `populate_global_this_builtins` independently, so
  that assumption stops holding the moment a second OS thread exists.
* `object/mod.rs:267-279` records that #7251 already hit this for **tests** and
  fixed it with `per_test_global!` — which, per `per_test_global.rs:60-62`,
  *"in a non-test build … expands to the plain `static` it replaced, byte for
  byte"*. The production defect was never addressed.

`scripts/gc_runtime_root_holders.json:90-130` marks the iterator cells
`covered_elsewhere`; that verdict is about **GC-scan coverage**
(`object::scan_object_cache_roots_mut`, `object/mod.rs:1141`) and says nothing
about the per-realm hazard, so it does not rule this out.

## Suggested acceptance

A `perry/thread` test in which agent B runs `for (const x of [1,2,3])` and a
`function*`, and a perry-runtime unit test in the shape of
`a_second_agents_prototype_addresses_are_its_own` (#7994) asserting two live
agents get DIFFERENT tower addresses, with per-agent liveness asserted before
distinctness.

The obstacle usually cited ("Darwin has no local-exec TLS") is stale — see
#8001's closing note and `crates/perry-runtime/src/tls_hot.rs`.

### LINKED ISSUE #8003: gc/thread: longlived native-module caches and Web Storage brand pointers are process-global
Same family as #7988 / #8001 / #8002, lower severity but strictly simpler to fix.

## 1. Longlived native-module caches (dangling read)

```
// crates/perry-runtime/src/object/mod.rs:245-251
HTTP_METHODS_CACHE, FS_CONSTANTS_CACHE, OS_CONSTANTS_CACHE,
OS_CONSTANTS_SIGNALS_CACHE, OS_CONSTANTS_ERRNO_CACHE,
OS_CONSTANTS_PRIORITY_CACHE, OS_CONSTANTS_DLOPEN_CACHE
```

`object/native_module/namespace_builders.rs` (`http_methods_array` at `:44`, the
`fs`/`os` builders at `:30-37`, `:605-703`) each do `if cached != 0 { return
cached }`, else allocate with `js_array_alloc_with_length_longlived` /
`js_string_from_bytes_longlived` and cache the address. Those allocators use
`LONGLIVED_ARENA`, which is **thread-local**
(`crates/perry-runtime/src/arena/block.rs:992`).

So agent B's first `require('http').METHODS` — after agent A has already built
it — hands back a pointer into **A's** longlived arena: an unattributed
cross-arena read (#7988 consequence 2), and a guaranteed dangling read once A
exits and its arena is freed.

## 2. Web Storage brand pointers (wrong identity)

```
// crates/perry-runtime/src/object/mod.rs:287-288  — not even per_test_global!
static LOCAL_STORAGE_PTR / SESSION_STORAGE_PTR
```

`web_storage.rs:220-231` overwrites both on every `populate_global_this_builtins`
run — i.e. on **every thread's** first `js_get_global_this` — and
`web_storage.rs:375-389` `storage_kind_from_value` brand-checks by pointer
equality against them.

Consequence: agent A's own, perfectly valid `localStorage.setItem(…)` starts
failing its brand check (falls to the `None` branch) the moment agent B
bootstraps its realm and overwrites the shared pointer with B's storage object.
That is #7988 consequence 1 applied to Web Storage, and unlike #7988 it degrades
the FIRST thread rather than the later ones.

## 3. `FUNCTION_CLASS_IDS` (narrower)

`class_registry/state.rs:309` — `RwLock<Option<HashMap<u64, u32>>>` keyed by
`func_bits`, the NaN-boxed bits of a **closure heap pointer** (`state.rs:299`).
Needs an address collision across two arenas to bite, so it is architecturally
less likely than #8001's "same class_id on every thread", but it is the same
process-global-map-of-heap-addresses shape and the synthetic ids it hands out
come from a process-wide counter (`prototype_objects.rs:125`).

## Explicitly NOT in scope (checked, safe)

* `SYMBOL_REGISTRY` / `WELL_KNOWN_SYMBOLS` / `REGISTERED_SYMBOL_DESCRIPTIONS`
  (`symbol.rs:112,270,132`) — `Symbol.for()` identity is spec-mandated to be
  shared across realms, and descriptions are deliberately Rust-owned off-arena
  (`symbol.rs:119-141` reasons the arena hazard through explicitly).
* `SYMBOL_PROPERTIES` / `SYMBOL_PROPERTY_ATTRS` / `CLASS_STATIC_SYMBOLS`
  (`symbol.rs:499,506,819`) — process-global and owner-keyed, but each thread
  only ever probes with its own pointers. The residual is a bounded leak and is
  already documented in-file at `symbol.rs:508-518`.
* `ARRAY_PROTO_HAS_INDEX` / `OBJECT_PROTO_HAS_INDEX` /
  `PERRY_ARRAY_INDEX_FAST_PATH_INVALIDATED`, `DECLARED_FIELD_NAME_HASHES`,
  `ELEMENT_SHAPE_EPOCH`, `GC_UNSAFE_ZONES` and the `OnceLock` policy caches —
  sticky "somebody somewhere patched X" latches and env-derived config. A safe
  over-approximation: every thread takes the slow path, nobody gets a wrong
  answer.

proggeramlug:
## Summary

- resolve iterator, generator, `%TypedArray%`, native-module, and Web Storage GC roots through agent-local hot-TLS backing cells
- scan and barrier-write the current agent's backing atomics instead of publishing raw arena addresses process-wide
- protect lazy iterator-tower construction from collections and add two-live-agent plus forced-relocation regressions

## Issue disposition

#8002 was live in full: iterator roots were bare statics, while generator and `%TypedArray%` used test-only TLS that remained process-global in production.

#8003 was partly resolved before this branch: #8024 moved `FUNCTION_CLASS_IDS` and its companion class registries to `perry_thread_local!`. This PR completes its two remaining stated mechanisms: all seven long-lived native-module caches and both Web Storage brand roots.

## Validation

- rebased onto `fe0d49792` (#8044); merge base equals current `origin/main`
- optimized `gc::tests::lazy_intrinsic_towers`: 4/4 pass
- two-agent cache-root scanner isolation tests: pass
- registered `perry/thread` parity regression: pass with pinned runtime/static archives
- post-rebase rate-1 + moving polls + from-space protection depth 800: identical output, 1,406 copying minors, 32,047 moved objects, 60,040 loop polls, 1,406 retired page-sets
- post-rebase rate-1 + moving polls + evacuation verification: identical output, 1,400 copying minors, 32,047 moved objects
- sabotage to process-global backing: two-agent gate failed on shared `HTTP_METHODS_CACHE`; restored source rebuilt and passed
- post-sabotage and post-rebase combined `perry`, `perry-runtime-static`, `perry-stdlib-static` optimized builds: pass; archive mtimes verified
- formatting, registration, GC-doc, and runtime-root-holder inventory checks: pass
- `gc_store_site_inventory.py` independently reports the pre-existing `property_set.rs:1475` marker omission from `5fcd94289`; this branch does not touch that file

Closes #8002
Closes #8003

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed cross-thread realm isolation for iterator, generator, typed-array, native-module, and Web Storage state.
  - Prevented iterator prototypes from moving during initialization.
  - Ensured changes in one realm remain isolated from others.

- **Tests**
  - Added regression coverage for concurrent realms, garbage collection, intrinsic prototypes, module constants, and storage isolation.

- **Documentation**
  - Documented realm-local garbage-collection root changes and validation results.

### DIFF
--- a/changelog.d/8055-agent-local-realm-roots.md
+++ b/changelog.d/8055-agent-local-realm-roots.md
@@ -1,0 +1,1 @@
+Keep iterator, generator, `%TypedArray%`, native-module, and Web Storage GC roots local to the Perry agent whose arena owns them.
--- a/crates/perry-runtime/src/gc/tests/lazy_intrinsic_towers.rs
+++ b/crates/perry-runtime/src/gc/tests/lazy_intrinsic_towers.rs
@@ -1,11 +1,12 @@
-//! #7251: the two lazy intrinsic-tower builders need #7217's NO-MOVE WINDOW,
-//! and #7217 itself could not gate them.
+//! #7251/#8002: lazy intrinsic-tower builders need #7217's NO-MOVE WINDOW,
+//! and every realm must own the tower roots that point into its arena.
 //!
+//! `object::iterator_prototypes::build_iterator_prototypes`,
 //! `object::global_this::generator::build_generator_tower` and
 //! `object::global_this::typed_array::ensure_typed_array_intrinsic` have the
 //! identical shape #7217 fixed for `populate_global_this_builtins`: each
-//! builds an IMMORTAL object graph (hanging off a process-global intrinsic
-//! slot for the life of the thread) by threading raw `*mut ObjectHeader` /
+//! builds an IMMORTAL object graph (hanging off an agent-local intrinsic root
+//! for the life of the thread) by threading raw `*mut ObjectHeader` /
 //! `*mut ClosureHeader` locals across a dozen-plus allocating installs. A
 //! relocating — or even a non-relocating, freeing — collection reached from
 //! one of THEIR OWN allocations leaves the rest of the build writing through
@@ -42,24 +43,20 @@
 //!    allocation.
 //!
 //! 2. **Re-exec the test binary so the child is the first to touch a tower.**
-//!    The six intrinsic-tower statics were plain process-global `AtomicI64`s,
+//!    The intrinsic-tower statics were plain process-global `AtomicI64`s,
 //!    built exactly once per *process* — so whichever test happened to run
 //!    first on that binary (libtest's ordering is not something a single test
 //!    controls) built them for every other test, including a freshly
 //!    re-exec'd child, because OTHER tests in the SAME binary run before this
-//!    one and touch `globalThis`. FIXED HERE by converting the six statics
+//!    one and touch `globalThis`. FIXED HERE by converting the statics
 //!    (`TYPED_ARRAY_INTRINSIC_PTR`, `TYPED_ARRAY_INTRINSIC_PROTO_PTR`,
 //!    `GENERATOR_FUNCTION_INTRINSIC_PTR`, `GENERATOR_INTRINSIC_PROTO_PTR`,
 //!    `GENERATOR_PROTOTYPE_PTR`, `ASYNC_GENERATOR_FUNCTION_INTRINSIC_PTR`,
 //!    `ASYNC_GENERATOR_INTRINSIC_PROTO_PTR`, `ASYNC_GENERATOR_PROTOTYPE_PTR`)
-//!    in `object/mod.rs` from a bare `static` to `per_test_global!` — the
-//!    exact mechanism #7672 built for this class of hazard (see
-//!    `per_test_global.rs`'s module docs). In a test build each libtest
-//!    THREAD gets its own zeroed instance, and libtest runs one thread per
-//!    test, so "per thread" and "per test" coincide: this test's first read
-//!    of `GENERATOR_FUNCTION_INTRINSIC_PTR` is guaranteed `0` regardless of
-//!    what any other test in the binary did. Non-test builds expand to the
-//!    identical plain `static` — zero behavioural change in shipped binaries.
+//!    in `object/mod.rs` from bare process statics to `perry_thread_local!`
+//!    backing slots. Each libtest thread therefore gets a guaranteed-first
+//!    touch, and shipped `perry/thread` agents no longer reuse another realm's
+//!    raw arena pointers (#8002).
 //!
 //! 3. **Record `gc_is_suppressed()` from inside the builder under
 //!    `#[cfg(test)]`.** Reported *suppressed* even with the scope removed,
@@ -116,12 +113,9 @@ use super::super::*;
 use super::support::*;
 use std::sync::atomic::Ordering as StdOrdering;
 
-/// Run `body` on a thread that has touched neither `globalThis` nor either
-/// tower, mirroring `global_bootstrap.rs::on_a_fresh_thread`. Belt-and-braces
-/// alongside the `per_test_global!` conversion above: libtest already gives
-/// each `#[test]` its own thread (and therefore its own instance of every
-/// `per_test_global!` table), but spawning explicitly keeps this test's
-/// guarantee independent of that harness detail.
+/// Run `body` on a thread that has touched neither `globalThis` nor any lazy
+/// tower, mirroring `global_bootstrap.rs::on_a_fresh_thread`. The explicit
+/// spawn keeps this test's guarantee independent of libtest's thread reuse.
 fn on_a_fresh_thread(body: impl FnOnce() + Send + 'static) {
     std::thread::Builder::new()
         .stack_size(16 << 20)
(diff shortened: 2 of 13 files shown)


proggeramlug (commented):
Audited exact head `56d69fe8166e30151d4f190e8e318e1430c98574` against base/current main `fe0d4979204dfd6b8b166320e1ebdd2318f30518`. The production mechanism is sound: all 23 affected heap roots resolve through calling-agent `perry_thread_local!` backing atomics; all mutator writes remain root-barriered; mutable scanning visits and rewrites the same current-agent cells; the iterator construction suppression scope covers the raw-pointer allocation window. #8024 genuinely made `FUNCTION_CLASS_IDS` production-agent-local; the remaining global synthetic-id counter is pointer-free, monotonic code metadata. No version bump or landing conflict.

One test-harness issue needs fixing before merge: `realm_owned_intrinsic_module_and_storage_roots_are_distinct` performs fallible materialization/snapshot work before its only `Barrier::wait()`. If either spawned agent panics before reaching the barrier, its peer waits forever and both parent joins block, so a real regression can hang the suite instead of failing. Please make the rendezvous unwind-safe/bounded while preserving the invariant that both arenas remain alive through both snapshots (for example, one release-on-drop guard as the sole wait path). The suggested `RealmAtomicU64::store` helper and explicit builder calls are maintainability improvements, not blockers: the current bootstrap materializes those roots and the 23 nonzero assertions prevent a vacuous pass.

Independent static checks passed: diff check, formatting, thread-local policy, test registration, GC doc claims, and runtime-root-holder self-test/inventory. `gc_store_site_inventory.py` still reports only the pre-existing untouched `property_set.rs:1475` marker omission.

proggeramlug:
Fixed the blocking two-agent test deadlock in `977a19b8b`.

The agent closure now installs a `ReleasePeerOnDrop` guard before any fallible materialization or snapshot work. Its `Drop` implementation owns the sole `Barrier::wait()` path, so it runs on both success and unwind. Successful agents still keep their arenas alive until both snapshots have been captured.

Focused optimized evidence:

- healthy exact two-agent test: pass, `1 passed`, finished in `0.02s`
- forced one-agent panic immediately after installing the guard: failed normally in `0.02s` with `rendezvous unwind sabotage`; the sibling and parent did not hang
- sabotage removed and source verified equal to the commit
- restored optimized `gc::tests::lazy_intrinsic_towers`: `4 passed`, finished in `0.02s`
- `cargo fmt --all -- --check` and `git diff --check`: pass

No production code, changelog, or version was changed.

proggeramlug:
Independent re-audit of exact head `977a19b8bb91988a098f25eeb66773c1825247f7`: merge-safe.

I tested a clean landing-equivalent merge over current `main` `471f65096` (#8054); there is no conflict or semantic overlap. The exact optimized focused test passed:

```
gc::tests::lazy_intrinsic_towers::realm_owned_intrinsic_module_and_storage_roots_are_distinct ... ok
1 passed; 0 failed; 2317 filtered out; 0.03s
```

The blocking synchronization defect is fixed. `ReleasePeerOnDrop` is installed as the worker closure's first statement and owns the sole `Barrier::wait()` path. The snapshot return expression is evaluated before the guard drops, so both agents and their TLS arenas remain alive through both captures; unwind follows the same single rendezvous and cannot strand the peer. The author also forced an immediate-after-guard panic and observed prompt normal failure, then restored and verified the exact source.

I re-audited the full production diff: all 23 affected roots resolve through current-agent HotKey backing cells; writes remain root-barriered; the mutable scanner rewrites those same cells; iterator construction has the needed no-move scope. #8024 already made the heap-pointer-keyed `FUNCTION_CLASS_IDS` registry agent-local; the remaining global synthetic-ID counter is pointer-free monotonic metadata. Closure of #8002 and #8003 is supported.

`git diff --check`, TLS policy/inventory, runtime-root-holder inventory, and GC store-site self-tests pass. No version/Cargo/package change and zero unresolved review threads. CI was not used as evidence.

Disposition: merge exact head and close #8002/#8003.

Issue #8040, opened 2026-08-13
[Next.js/dylib] Full production App Route compatibility tracker


proggeramlug:
## Goal

Run a real production Next.js 16 App Route as an app-only Perry dylib, with runtime and stdlib supplied once by separate shared provider images, and pass the same black-box verifier as Node without compatibility fallbacks.

This is the correctness gate required before the Perry-vs-Node 1/10/100 deployment RSS, CPU, and startup comparison is meaningful. The current benchmark is promising but lower-bound-only because it calls userland directly and fabricates the final response.

## Fixed reproduction — do not design another app

#8034 contains a complete copy/paste fixture generator pinned to Next.js `16.3.0` / React `19.2.4`, plus a 21-request verifier. The app already exercises:

- production `next build --webpack` output;
- the generated `routeModule.handle` / `AppRouteRouteModule.handle` path;
- a re-exported imported `GET`/`POST` handler;
- method, pathname, query, headers, and POST body;
- Next's AsyncLocalStorage-backed `headers()` before and after async continuations;
- a request-time dynamic import;
- a two-chunk streamed `NextResponse`;
- status, custom response header, and `Set-Cookie`; and
- 20 distinct concurrent request contexts.

Every child issue uses that app. Smaller compiler/runtime fixtures are welcome for localization, but they cannot replace #8034 as acceptance evidence.

## Work items

- [ ] #8034 — commit the pinned production fixture and add the app-only-dylib CI gate
- [ ] #8036 — preserve `NextRequest` and `nextUrl.searchParams` across the production import/re-export boundary
- [ ] #8038 — preserve returned `NextResponse` status, headers, cookies, and stream
- [ ] #8037 — preserve Next request/work AsyncLocalStorage under concurrent App Routes (scoped cut of #6764)
- [ ] #8039 — make cold/concurrent production webpack route and dynamic-module loading deadlock-free
- [ ] #8035 — install array-growth forwarding for valid low-address macOS arena allocations

## Existing groundwork to retain

- #5438: production `.next/server/**` discovery and path-based lazy module initialization
- #6738: native backing for `URLSearchParams` subclasses
- #7252: GC rooting for cross-module call arguments
- #237 and #1650: Web Streams / `Response.body` groundwork
- #6764: broad `node:async_hooks` parity inventory

These reduce the work but do not replace the composed production test. If current `main` already fixes a child behavior, close that child with a #8034 regression proving it rather than with a synthetic-only test.

## Suggested execution order

[...]

PR #8058, opened 2026-08-13
Fix relocatable ptrtoint constants in split native codegen

### LINKED ISSUE #8057: Native split codegen cannot materialize relocatable ptrtoint constant operands
## Summary

The split native codegen-unit path cannot materialize the relocatable
`ptrtoint` constant expression that every non-entry JS module uses to pass its
`__init_body` function pointer through the integer-valued runtime ABI.

This blocks production-sized webpack graphs before application code can run.
The whole-module text path accepts the same IR because LLVM parses the constant
expression itself; the in-process native dialect reader instead feeds the full
expression to the integer literal parser.

## Exact reproduction

Base: `fe0d4979204dfd6b8b166320e1ebdd2318f30518` (current `main` after #8044)

Using the pinned Next.js 16.3.0 webpack production route graph from #8034,
compiled as an app-only dylib with native split units:

```text
perry compile --no-cache --no-auto-optimize --output-type dylib perry-host.js -o next-app-route.dylib
```

Perry collects 91 modules and reaches native codegen, then independently fails
all four large split modules, including the generated
`next/dist/compiled/next-server/app-route.runtime.prod.js`:

```text
native codegen unit ... failed: ...
bad integer `ptrtoint (ptr @<module-prefix>__init_body to i64)`
```

Perry correctly refuses to link the incomplete dylib (87/91 modules emitted).

## Minimal compiler-level reproduction

Define two functions in an `LlModule`: an `__init_body`, and an `__init` wrapper
whose runtime call uses this operand:

```llvm
i64 ptrtoint (ptr @fixture_js__init_body to i64)
```

Then invoke `compile_module_units_native(..., 2, ...)`. This forces independent
native construction and object emission followed by the real partial-link path.
Before the fix it reaches `i128::parse` and fails with `bad integer`; a parse-only
or textual-backend test does not reproduce the defect.

## Acceptance

- The focused two-unit emitted-object/link regression passes.
- The exact production graph gets past these four `ptrtoint` construction
  failures.
- No #8034 behavior acceptance is inferred: its generated
  `AppRouteRouteModule.handle` path must still actually execute with the required
  moving/verifier evidence before that separate production ticket can close.

Refs #8034 and #8040.

proggeramlug:
## Summary

- recognize LLVM `ptrtoint` constant operands in the in-process dialect reader
- materialize function/global references as relocatable LLVM constants instead
  of sending the expression to the integer-literal parser
- prove the fix through two native codegen units, independent object emission,
  and the real partial-link path

## Why

The pinned Next.js 16.3.0 production route graph from #8034 exposed four large
split modules that all failed on the same module-init operand:

```llvm
i64 ptrtoint (ptr @module__init_body to i64)
```

On exact `main` at `fe0d4979204dfd6b8b166320e1ebdd2318f30518`, Perry
emitted 87 of 91 module objects and then correctly refused to link. The textual
backend accepts the expression because LLVM parses it; only the split native
reader tried to parse it as an integer literal.

This PR is deliberately the compiler-level fix only. It does not claim that the
Next.js route executes and does not close any production acceptance ticket.

## Verification

- `cargo test --profile perry-dev -p perry-codegen split_units_emit_and_merge_init_body_pointer_constant -- --nocapture`
  - 1 passed; focused object emission and partial link succeeded
- exact cold production replay from the pinned 91-module Next.js graph:
  - `.next/server/chunks/2.js`: 3/3 units emitted; merged 7.2 MiB object
  - `next/dist/compiled/jsonwebtoken/index.js`: 2/2 units emitted; merged 5.0 MiB object
  - `app-page.runtime.prod.js`: 12/12 units emitted; merged 31.3 MiB object
  - `app-route.runtime.prod.js`: 4/4 units emitted; merged 9.7 MiB object
  - all four former `bad integer ptrtoint (...)` failures are gone
- `git diff --check`

The production graph continues beyond this defect. Route execution and the
moving/verifier acceptance evidence remain separate work under #8034/#8040.

No version bump.

Closes #8057.
Refs #8034 and #8040.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved split native code generation for pointer-to-integer constants.
  * Preserved relocations for function and global references across emitted and partially linked objects.
  * Enabled module initialization wrappers to pass initialization pointers through the runtime interface.

* **Tests**
  * Added regression coverage for pointer-to-integer constants across separate code generation units and partial linking.

* **Documentation**
  * Documented the updated native code generation behavior.

### DIFF
--- a/changelog.d/8058-split-native-ptrtoint-constants.md
+++ b/changelog.d/8058-split-native-ptrtoint-constants.md
@@ -1,0 +1,7 @@
+### Fix relocatable constants in split native codegen
+
+Production-sized modules using split native codegen now materialize LLVM
+`ptrtoint` constant operands for function and global references instead of
+misparsing them as integer literals. Module-init wrappers can therefore pass
+their `__init_body` function pointer through the runtime ABI while retaining a
+real relocation in each independently emitted and partially linked object.
--- a/crates/perry-codegen/src/dialect/types.rs
+++ b/crates/perry-codegen/src/dialect/types.rs
@@ -202,6 +202,48 @@ pub(super) fn constant<'ctx>(
         }
         bail!("reference to unknown global @{name}");
     }
+    // LLVM constant expressions can appear anywhere an ordinary constant is
+    // accepted. Perry emits this exact form when a non-entry module passes its
+    // `__init_body` function pointer through the integer-valued runtime ABI:
+    //
+    //   call void @js_run_module_init_catching(
+    //       i64 ptrtoint (ptr @module__init_body to i64))
+    //
+    // The text backend has always delegated this to LLVM's assembler. The
+    // in-process reader used to feed the whole expression to `i128::parse`, so
+    // only split native codegen units failed, late in a large build, with
+    // `bad integer ptrtoint (...)`. Materialize the relocatable constant with
+    // LLVM's constant API; using a runtime instruction would be invalid here
+    // because this is an operand constant, not an SSA definition.
+    if let Some(body) = tok
+        .strip_prefix("ptrtoint (")
+        .and_then(|body| body.strip_suffix(')'))
+    {
+        let (source, destination) = body
+            .rsplit_once(" to ")
+            .ok_or_else(|| anyhow!("bad ptrtoint constant `{tok}`"))?;
+        let (source_ty, source_value) = ty_and_val(source)?;
+        let source_ty = basic_type(ctx, source_ty)?;
+        if !source_ty.is_pointer_type() {
+            bail!("ptrtoint source is not a pointer in `{tok}`");
+        }
+        let destination_ty = basic_type(ctx, destination.trim())?;
+        let BasicTypeEnum::IntType(destination_ty) = destination_ty else {
+            bail!("ptrtoint destination is not an integer in `{tok}`");
+        };
+        let BasicTypeEnum::IntType(expected_ty) = ty else {
+            bail!("ptrtoint constant used as non-integer {ty:?}");
+        };
+        if destination_ty != expected_ty {
+            bail!(
+                "ptrtoint destination type {} disagrees with operand type {}",
+                destination_ty.print_to_string(),
+                expected_ty.print_to_string()
+            );
+        }
+        let source = constant(ctx, module, source_ty, source_value)?.into_pointer_value();
+        return Ok(source.const_to_int(destination_ty).into());
+    }
     Ok(match tok {
         // The null of the OPERAND's type, not of address space 0: RS4GC emits
         // `store ptr addrspace(1) null, ptr %slot`, and an addrspace(0) null
--- a/crates/perry-codegen/src/native_emit.rs
+++ b/crates/perry-codegen/src/native_emit.rs
@@ -557,6 +557,45 @@ fn debug_dump(module: &Module<'_>, module_prefix: &str) {
     }
 }
 
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::module::LlModule;
+    use crate::types::{I64, VOID};
+
+    #[test]
+    fn split_units_emit_and_merge_init_body_pointer_constant() {
+        // Production webpack modules are large enough to use split native
+        // codegen. Every non-entry module's guard passes `__init_body` to the
+        // exception boundary as an i64 constant expression. Keep the callee
+        // and wrapper in separate units so this covers declaration lookup,
+        // relocatable ptrtoint construction, object emission, and the partial
+        // link -- a parse-only test would miss the production-graph failure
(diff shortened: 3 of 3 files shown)


proggeramlug:
@coderabbitai review

proggeramlug:
Independent merge audit — exact PR head `2dd16ae88fe22593babad15c78d8a9c086dfba0f`, landed cleanly for testing on current `origin/main` `ef6e111a4fb85dcc9fd81f69c30bf57a0bf40a61`.

**Disposition: MERGE. No blocking findings.**

I reviewed the full 3-file / 88-addition landing diff and traced the native reader and split-unit paths:

- The new grammar accepts Perry's exact `ptrtoint (ptr @symbol to i64)` constant-expression shape, requires a pointer source, requires an integer destination, and requires the destination integer type to equal the operand-context type.
- Recursive constant resolution is correct for the intended function/global symbol operand: it finds the declaration in the unit module, obtains its LLVM pointer constant, and `const_to_int` retains the relocation. Unknown symbols fail loudly.
- Split-unit declaration handling is correct here: the wrapper unit's reference scan synthesizes the real declaration for `fixture_js__init_body`; its defining unit carries the definition. Native freezing also preserves the external runtime declaration.
- The test genuinely forces two independent native units: two defined functions with `n=2` produce two one-function buckets, each compiled in its own LLVM context. A progress-enabled run showed both `LLVM unit 1/2` and `2/2` finish, followed by `merging 2 unit objects` and `merged 2 unit objects`; on this Darwin host that is the real `ld -r` partial-link path, not parse-only validation.

Exact focused verification on that isolated clean landing merge, with its own cold target directory:

```text
cargo test --profile perry-dev -p perry-codegen split_units_emit_and_merge_init_body_pointer_constant -- --nocapture
Finished `perry-dev` profile [optimized] target(s) in 7m 39s
1 passed; 0 failed; 987 filtered out
cargo exit 0
```

Also passed:

- `git diff --cached --check`
- `git diff --check`

Malformed/internal dialect caveat, nonblocking: deliberately injected malformed low-level operand strings can still reach pre-existing panic-prone helpers (for example an empty source), and spelling a symbol with a mismatched nonzero source address space is not compared against the looked-up symbol's actual pointer type. The dialect itself is crate-private and consumes compiler-produced `LlFunction` items; the affected production tokens are generated in `entry.rs` from sanitized module symbols. I found no TypeScript/source-controlled raw-IR path to either shape, and Perry emits the canonical address-space-0 function/global form fixed by this PR.

Production evidence at this exact head is sufficient for #8057 and no broader claim: all four former failures fully emitted and partial-linked — webpack chunk `2.js` 3/3 -> 7.2 MiB, `jsonwebtoken` 2/2 -> 5.0 MiB, App Page runtime 12/12 -> 31.3 MiB, App Route runtime 4/4 -> 9.7 MiB — with no recurrence of `bad integer ptrtoint (...)`. The replay process itself ended later at 85/91 when the worker session closed, so this does **not** claim full graph completion, route execution, or moving/verifier acceptance.

Scope is correct: only #8057 is in the closing set; #8034 and #8040 are references and remain separate acceptance work. No version bump. No unresolved review threads or actionable review comments. I did not use CI as evidence.

proggeramlug:
Independent audit of exact head `2dd16ae88fe22593babad15c78d8a9c086dfba0f`: merge-safe.

I audited the full three-file diff and a clean landing-equivalent merge over current `main` `ef6e111a4`; there is no conflict or semantic overlap. The dialect branch recognizes only `ptrtoint (...)`, parses and requires a pointer source and integer destination, requires the destination type to equal the operand's expected integer type, resolves the function/global in the unit skeleton, and builds LLVM's relocatable constant rather than an SSA instruction.

The focused test is not parse-only: with two defined functions and `n=2`, the size balancer creates two one-function contexts/units, each goes through optimize/object emission, and `merge_unit_objects` runs the real partial-link path. My cold isolated landing-tree run finished the optimized `perry-dev` build and passed the exact test. A warm rerun with progress enabled positively reported:

```
freezing 2 codegen units
LLVM unit 1/2 finished
LLVM unit 2/2 finished
merging 2 unit objects
merged 2 unit objects
```

The exact production replay independently cleared all four #8057 failures through object emission and partial link: 3/3, 2/2, 12/12, and 4/4 units respectively. The replay later ended at module 85/91 when its worker session closed, so this is not a full-graph or route-behavior claim; #8034/#8040 correctly remain references only.

Formatting and diff checks pass, no version/Cargo/package change, and zero review threads. CI was not used as evidence. Malformed hand-authored internal dialect strings can still expose pre-existing parser robustness edges, but this parser is crate-private and receives Perry-constructed finalized items; no external/source IR injection path exists.

Disposition: merge exact head, close only #8057, keep #8034/#8040 open.

PR #8061, opened 2026-08-14
fix(array): route collection reads before array validation

### LINKED ISSUE #8060: Map/Set indexed reads return NaN after array pointer hardening
## Regression

At exact current `main` `ef6e111a4fb85dcc9fd81f69c30bf57a0bf40a61`, indexed reads from live Map/Set receivers routed through array-like runtime fallbacks return `NaN`. Six pre-existing tests fail individually and deterministically:

- `array::collection_tag_tests::a_live_map_receiver_still_reports_its_size_through_the_registry`: `NaN`, expected map key `1.0`
- `array::collection_tag_tests::a_live_set_receiver_still_reads_its_elements_through_the_registry`: `NaN`, expected set value `5.0`
- `value::dyn_index::collection_tag_tests::collection_receivers_still_use_their_authoritative_registries`: `NaN`, expected map key `1.0`
- `typed_feedback::tests::typed_feedback_boxed_fallback_preserves_fractional_keys_for_array_like_receivers`: `NaN`, expected set value `20.0`
- `typed_feedback::tests::runtime_dynamic_index_fallbacks_preserve_fractional_keys_for_array_like_receivers`: `NaN`, expected set value `20.0`
- `typed_feedback::tests::polymorphic_index_fallbacks_preserve_fractional_keys_for_array_like_receivers`: `NaN`, expected set value `20.0`

For example:

```sh
cargo test -p perry-runtime --lib \
  array::collection_tag_tests::a_live_map_receiver_still_reports_its_size_through_the_registry \
  -- --exact
```

## Causal proof

This is an ordering regression introduced by #8041 (`971d6ffb6b10c04724cd28d9ee6da2d750c84420`), not collection-registry lifetime or GC rooting:

1. #8041 correctly hardened `clean_arr_ptr` to return null for every tracked object whose `GcHeader::obj_type != GC_TYPE_ARRAY`.
2. `js_array_get_f64` calls `clean_arr_ptr(arr)` before its existing `GC_TYPE_MAP` / `GC_TYPE_SET` branches.
3. A genuine live Map or Set is therefore rejected at the clean funnel and returns `NaN`; its collection tag and authoritative registry can never be consulted.
4. `js_array_length` is the positive control: it checks the Map/Set tag plus authoritative registry before `clean_arr_ptr`, so the same live receivers still report the correct sizes. The array collection tests reach their expected length assertions and fail only on the subsequent indexed read.
5. `js_dyn_index_get` first confirms the receiver through the authoritative collection registry, then delegates to `js_array_get_f64`, which reproduces the same early rejection.

This also explains the three typed-feedback failures, since their get fallbacks converge on that dynamic/array accessor. Integer indexed reads are broken; the fractional-key undefined/no-mutation assertions remain the required control.

## Required fix

Route authoritative Map/Set receivers before strict array cleaning in `js_array_get_f64`, matching the proven `js_array_length` ordering, or use an equivalent dedicated collection accessor.

Keep #8041's strict rejection unchanged for genuine non-array/non-collection receivers. The fix should retain:

- receiver-tag gating before registry probes;
- the authoritative Map/Set registry check (never trust the tag alone);
- no collection registry probes for genuine arrays;
- fractional Map/Set keys returning undefined without mutation;
- all six exact tests passing locally.

proggeramlug:
## Summary

- route managed-header-tagged, authoritative-registry-confirmed Map/Set indexed reads before `clean_arr_ptr`'s strict array-only validation
- leave #8041's non-array rejection unchanged and assert that Map/Set still fail the array-only funnel
- reuse the pre-clean header tag for ordinary arrays, rereading only when cleaning resolves a forwarded array head

Fixes #8060.

## Causal evidence

Exact base: `ef6e111a4fb85dcc9fd81f69c30bf57a0bf40a61`.

#8041 made `clean_arr_ptr` reject every tracked non-array, but `js_array_get_f64` reached its Map/Set branches only after that rejection. Six pre-existing tests failed individually with `NaN` instead of collection values. `js_array_length` was the positive control: it already consulted the managed tag and authoritative registry before cleaning, and returned the correct sizes for the same receivers.

## Local validation

Using isolated worktree and target:

- all six formerly failing tests: 20/20 fresh-process runs each
- array collection-tag tests: 6/6
- dynamic-index collection-tag tests: 3/3
- typed-feedback tests: 49/49
- array-subclass strict-cleaning controls: 6/6
- array/forwarding tests: 74/74
- `cargo fmt --all --check`
- `git diff --check`

One full default-parallel runtime suite after the fix reported 2,313 passed, 1 failed, 4 ignored. The sole failure was `promise::keyed_table::tests::settling_many_keys_is_not_quadratic`, the independently diagnosed wall-clock assertion replaced by #8059; all six #8060 tests and the shadow-root witness passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed indexed reads on `Map` and `Set` values so they no longer incorrectly return `NaN`.
  * Out-of-range indexed reads now consistently return `undefined`.
  * Improved handling of collection-backed and forwarded array-like values during indexed access.

* **Tests**
  * Added coverage validating correct `Map` and `Set` indexing behavior, pointer validation, registry-backed reads, and collection size reporting.

* **Documentation**
  * Added a changelog entry describing the corrected collection indexing behavior.

### DIFF
--- a/changelog.d/8061-collection-index-before-array-clean.md
+++ b/changelog.d/8061-collection-index-before-array-clean.md
@@ -1,0 +1,5 @@
+### Fixed
+
+- Map and Set indexed reads through array-like runtime fallbacks no longer
+  return `NaN`: tag-selected, registry-confirmed collections are routed before
+  strict Array pointer validation (#8060).
--- a/crates/perry-runtime/src/array/collection_tag_tests.rs
+++ b/crates/perry-runtime/src/array/collection_tag_tests.rs
@@ -114,6 +114,10 @@ fn a_live_set_receiver_still_reads_its_elements_through_the_registry() {
     );
 
     let as_array = set as *const ArrayHeader;
+    assert!(
+        clean_arr_ptr(as_array).is_null(),
+        "#8041's array-only funnel must keep rejecting Set layout"
+    );
     let before = probes();
     assert_eq!(js_array_length(as_array), 3);
     assert_eq!(js_array_get_f64(as_array, 0), 5.0);
@@ -140,6 +144,10 @@ fn a_live_map_receiver_still_reports_its_size_through_the_registry() {
     );
 
     let as_array = map as *const ArrayHeader;
+    assert!(
+        clean_arr_ptr(as_array).is_null(),
+        "#8041's array-only funnel must keep rejecting Map layout"
+    );
     let before = probes();
     assert_eq!(js_array_length(as_array), 2);
     assert_eq!(js_array_get_f64(as_array, 0), 1.0, "entry 0's key");
--- a/crates/perry-runtime/src/array/indexing.rs
+++ b/crates/perry-runtime/src/array/indexing.rs
@@ -705,6 +705,8 @@ pub extern "C" fn js_array_numeric_get_f64_unboxed(arr: *mut ArrayHeader, index:
 /// Get an element from an array by index (returns f64)
 #[no_mangle]
 pub extern "C" fn js_array_get_f64(arr: *const ArrayHeader, index: u32) -> f64 {
+    const TAG_UNDEFINED_F64: f64 = f64::from_bits(0x7FFC_0000_0000_0001u64);
+
     // Issue #179 Phase 5: lazy fast path — must run BEFORE
     // `clean_arr_ptr` because that helper force-materializes a lazy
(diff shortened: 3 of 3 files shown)


proggeramlug:
@coderabbitai review

Issue #7966, opened 2026-08-12
CI gate freshness alert: 6 gate(s) have no recent successful `main` run


proggeramlug:
The following post-merge gates have not produced a successful `main` run within their freshness budget.

**This usually means they are starved, not broken** — queued behind a merge cadence the runner pool cannot drain (#7856). Check the Actions queue depth before assuming a code failure:

```bash
gh api "repos/PerryTS/perry/actions/runs?status=queued&per_page=100" \
  -q '.workflow_runs | length'
```

| gate | detail | why it matters |
|---|---|---|
| `gc-ratchet.yml` | last success 69.3h ago (budget 12h) | Primary defence for collector retention/evacuation accounting. Went dark 2026-08-09..08-11 while #7799/#7809/#7812/#7834/#7839 merged; #7843's seven red rows landed in that window. |
| `tls-budget.yml` | last success 58.1h ago (budget 12h) | _tlv_get_addr call-count ratchet. macOS-only by nature; cannot be relieved by moving pools. |
| `gc-ptr-shape-off-witness.yml` | last success 93.0h ago (budget 12h) | Exercises the Ptr<Shape> OFF state. CLAUDE.md's kill-policy: an unexercised mode is a configuration nobody has verified. |
| `gc-parse-churn-gate.yml` | last success 93.0h ago (budget 12h) | Parse-churn layout gate. |
| `gc-moving-witnesses.yml` | last success 93.0h ago (budget 12h) | Asserts the copying minor actually moved objects -- the 'gate must assert its subject was live' arm (#7024/#7025). |
| `llvm-inprocess.yml` | last success 231.1h ago (budget 12h) | In-process LLVM backend arm; the promotion prerequisite. |

Scheduling rationale and the cost it trades away: `docs/src/testing/ci-gate-scheduling.md`.

_This issue is maintained automatically by `gate-freshness.yml`. It is updated in place, never duplicated, and closes itself once every gate is fresh again._

proggeramlug:
Diagnosed. **The eleven do not share one cause**, and the issue's own hypothesis (starvation, #7856) is half of it.

[...]

PR #8071, opened 2026-08-14
Propagate native-root decisions to codegen-unit workers

### LINKED ISSUE #8070: Split native codegen workers lose native-roots backend decision and omit GC maps
## Product boundary

Split native codegen units can emit a mapped function with no compact GC map. The native-roots decision is stored in thread-local state on the module producer, but compile_module_units_native always emits units on newly spawned worker threads. Those workers start with both NATIVE_ROOTS_OVERRIDE unset and NATIVE_ROOTS_TARGET_OK=false.

This is not an Actions-only freshness failure. It is a source-causal product hole:

1. codegen::compile_module decides the target lowering with set_native_roots_for_target on the producer thread. Tests can make the same decision with NativeRootsPin::native.
2. compile_module_units_native freezes units, then std::thread::scope spawns worker threads.
3. Each worker calls native_plan_args and optimize_and_emit_module. Both ultimately consult rs4gc_enabled on that worker.
4. The worker-local default is false, so the plan does not request the assembly/compact-map path and optimize_and_emit skips STATEPOINT_REWRITE_PASSES.
5. A function whose IR contains addrspace(1) roots and a GC strategy therefore reaches object emission without statepoint rewriting or __perry_gcmap.

## Deterministic local reproduction

At exact main f9475fb9ae0e3e1fc21d7e9c7146d963f859bdec, with LLVM 22.1.4:

    LLVM_SYS_221_PREFIX=/opt/homebrew/opt/llvm \
      PERRY_CODEGEN_UNIT_JOBS=1 \
      cargo test --profile perry-dev -p perry-codegen \
      --features llvm-inprocess --lib \
      native_emit::tests::split_native_construction_lowers_precise_roots_before_rs4gc \
      -- --exact --nocapture

Result: FAIL. The text reference stays on the pinned producer when jobs=1 and includes __PERRY_GCMAP,__perry_gcmap; the native arm still spawns one worker and omits the map.

Without the explicit one-job arm on a higher-core host, compile_units_to_object also spawns workers. Both reference and native arms then lose the same thread-local decision and can compare equal while both are unmapped. That explains the earlier local false green and makes the current regression sensitive to machine parallelism.

The same deterministic boundary appeared in main runs 31766084651 and 31766266548 under LLVM 22.1.8: 990/991 unit tests passed; the native object was 808 bytes with no __perry_gcmap, while the text object was 1040 bytes with the map. Run 31766266548 failed before native smoke.

## Required fix / acceptance

- Resolve the native-roots/backend decision once on the producer and pass it explicitly in every frozen unit payload or worker closure. Do not have the LLVM emission layer re-read producer-owned thread-local state.
- Use that same explicit value for plan selection and STATEPOINT_REWRITE_PASSES so assembly/map finishing cannot disagree with rewriting.
- Make the split precise-root regression force the deterministic jobs=1 asymmetry (or add an equivalent explicit worker-decision assertion), and positively prove the resulting object contains a non-empty __perry_gcmap before byte equality.
- Retain a negative unsupported-target/shadow-stack arm so explicit propagation cannot turn the decision into an unconditional true.
- Pass the focused regression with LLVM 22.1.4 locally. A CI result is freshness bookkeeping, not proof.

Discovered while reconciling #7966. That issue must remain open until this product boundary is fixed and the llvm-inprocess gate is locally non-vacuous.

proggeramlug:
## Summary

- carry the producer-selected native-root backend explicitly into text and native codegen-unit workers
- use one explicit value for RS4GC rewriting, assembly/compact-map planning, and finishing
- make the split precise-root witness non-vacuous and add a shadow-stack negative worker arm

## Local proof

Exact main base f9475fb, LLVM 22.1.4:

- before: PERRY_CODEGEN_UNIT_JOBS=1 focused split native-root test fails; native object has no map while text has one
- after: focused native_emit tests 4/4 pass with one codegen job
- after: in-process RS4GC tests 2/2 pass; compile-plan tests 5/5 pass; focused perry-codegen cargo check passes
- cargo fmt --all --check and git diff --check pass

No broad workspace test was used.

Closes #8070.
Refs #7966.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed native root handling during split compilation so all workers consistently use the selected root backend.
  - Improved statepoint rewriting and compact GC-map generation for native-root builds.
  - Ensured native and shadow-stack compilation modes remain consistent across direct, split, and differential builds.
- **Tests**
  - Expanded coverage for root-backend propagation, GC-map output, concurrent compilation, and native emission scenarios.

### DIFF
--- a/changelog.d/8071-split-native-root-workers.md
+++ b/changelog.d/8071-split-native-root-workers.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Propagate the producer-selected root backend into split LLVM codegen workers so native-root units cannot skip statepoint rewriting and compact GC-map emission (#8070).
--- a/crates/perry-codegen/src/inprocess.rs
+++ b/crates/perry-codegen/src/inprocess.rs
@@ -181,13 +181,14 @@ pub fn compile_ll_to_object_inprocess(
     effective_target: &str,
     clang_style_args: &[String],
     module_name: &str,
+    native_roots: bool,
 ) -> Result<Vec<u8>> {
     let (opt, mcpu_native, explicit_cpu, mllvm, emit_asm) = interpret_plan_args(clang_style_args)?;
     // Same guard as the external `opt` path (`linker::rs4gc_funclet_refusal`):
     // rewrite-statepoints-for-gc crashes on WinEH funclet pads, and here the
     // pass runs inside THIS process — the crash would take the compiler down
     // with it, not just a child.
-    if crate::codegen::helpers::rs4gc_enabled() {
+    if native_roots {
         if let Some(refusal) = crate::linker::rs4gc_funclet_refusal(ll_text) {
             return Err(anyhow!(refusal));
         }
@@ -202,6 +203,7 @@ pub fn compile_ll_to_object_inprocess(
         explicit_cpu.as_deref(),
         &mllvm,
         emit_asm,
+        native_roots,
     )
 }
 
@@ -310,6 +312,7 @@ pub(crate) fn optimize_and_emit_module(
     module: &inkwell::module::Module<'_>,
     effective_target: &str,
     clang_style_args: &[String],
+    native_roots: bool,
 ) -> Result<Vec<u8>> {
     let (opt, mcpu_native, explicit_cpu, mllvm, emit_asm) = interpret_plan_args(clang_style_args)?;
     optimize_and_emit(
@@ -320,6 +323,7 @@ pub(crate) fn optimize_and_emit_module(
         explicit_cpu.as_deref(),
         &mllvm,
         emit_asm,
+        native_roots,
     )
 }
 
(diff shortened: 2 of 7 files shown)


proggeramlug:
@coderabbitai review

proggeramlug:
Independent exact-head audit: APPROVE 11b61e5d0aeaa82bd2affb55d0b65d5836e01140. The producer-selected native-root decision is carried coherently through text and native unit workers, RS4GC, compile planning, compact-map finishing, and non-split paths. The positive split arm requires a real platform map section plus PGCM payload on both objects before byte equality; the shadow negative proves the producer emits shadow binds and no addrspace(1) root, then requires both text and worker-native artifacts to have no map/payload and to match exactly. Local focused evidence is 4/4 native_emit tests, 2/2 in-process RS4GC tests, 5/5 compile-plan tests, cargo check, fmt, and diff check. CI was not used as correctness evidence. Changelog/closure scope are correct, no version bump, and CodeRabbit completed with zero review threads.

Issue #7777, opened 2026-08-10
repsel-census red on main: for (const r of rows) no longer licenses element-shape facts (fixture_ptr_shape_elements 0/3)


proggeramlug:
## Symptom

`repsel-census` is failing on `main` (runs 31295652054 and 31240304595, both `failure`):

```
REGRESSION:
  fixture_ptr_shape_elements: ptr-shape promoted 0, floor is 3 (-3)
  fixture_ptr_shape_elements: ptr-shape-consumed promoted 0, floor is 3 (-3)
```

The job is not in branch protection's required contexts (its header says promotion was a deliberate follow-up after first landing), so merges proceed over it — CLAUDE.md "four ways a gate can be unable to fail", mode 2.

## Localization

The fixture's two read forms isolate cleanly on main @423bb4405 (fresh worktree, debug build):

* **Form A** (`for (let i = 0; i < rows.length; i++) { const s = rows[i]; … }`) — promotes fine on its own.
* **Form B** (`for (const r of rows) { … }`) — promotes **zero** on its own, method call or not, and its presence in the full fixture voids the whole group (E3: the desugared use of `rows` is not licensed, so the array loses its facts and the producer is re-denied under rule 2).

Minimal repro (promotes 0; delete the `for…of` loop and the producer promotes):

```ts
class Row {
  id: number; weight: number; score: number;
  constructor(id: number, weight: number) { this.id = id; this.weight = weight; this.score = 0; }
}
function build(n: number): number {
  const rows: Row[] = [];
  for (let i = 0; i < n; i++) { const row = new Row(i, i * 0.5); rows.push(row); }
  let total = 0;
  for (const r of rows) { total = total + r.weight; }
  return total;
}
console.log("b:" + build(6));
```

`collectors/ptr_shape_elements.rs` licenses the iterator form by matching the desugar's exact shape (module doc: "`for (const r of A)` desugars to exactly the E5 shape — a `__idx` local, `__idx < __arr.length`, `Let r = IndexGet(__arr, __idx)`"), and the fixture's own header predicted this failure: "If the desugar ever changes, this fixture's count drops and the gate goes red — which is the point." The gate went red; it just could not block. Suspect a change to `lower/stmt_loops.rs::lazy_or_index_elem` (or the alias binding it emits — e.g. a now-mutable `__arr` falls out of `collect_alias_edges`' `mutable: false` requirement and the bare `LocalGet(rows)` in its init disqualifies the array).

## Found while

[...]

Issue #7144, opened 2026-07-31
codegen: content-addressed temp .ll files are never deleted (~1 GB on a dev box)


proggeramlug:
Split out of #7131 / #7135 — found while verifying emission determinism on Linux.

## What happens

`compile_ll_to_object` no longer unlinks the temp `.ll` on the success path.
Only the `.o` is removed:

```rust
// crates/perry-codegen/src/linker.rs
} else {
    let _ = fs::remove_file(&plan.obj_path);
}
```

That was deliberate (#7135, CodeRabbit): the `.ll` is now content-addressed, so
concurrent workers with identical IR **share** that path and a per-call delete
could race a sibling's `clang` opening it.

The consequence is that nothing ever deletes them. Because the name is a
function of the IR, the file count is bounded by the number of **distinct IR
contents ever compiled** on the machine — not by the number of compiles, which
is the good news. The bad news is that iterating on the compiler changes the IR
on essentially every rebuild, so it grows without bound in practice.

Measured on one dev machine after roughly a day of compiler work:

```
leftover perry_llvm_*.ll files: 1627, total 951.8 MB
```

`/tmp` on CI runners and short-lived containers gets reclaimed, so this is a
developer-machine and long-lived-build-box problem rather than a correctness one.

## Why it is not simply "delete it again"

The race #7135 was avoiding is real: worker A can finish and unlink while worker
B has computed the same path but not yet handed it to `clang`. Unlinking a file
another process already has *open* is fine on POSIX; the window is between path
computation and `open`.

Options, none obviously free:

- age-based sweep of `perry_llvm_*.ll` at process start (bounded, but adds a
  temp-dir scan to every compile, and the temp dir can hold tens of thousands of
  unrelated entries);
- refcount / lockfile per content hash;
- put the shared IR under one per-process subdirectory and remove the directory
  at exit — loses cross-process sharing, keeps determinism (the recorded name is
  the **basename**, and only the basename, which is measured in #7131);
- leave it and document it.

The third looks best on the evidence in #7131 — clang records only the basename,
not the directory — but it deserves its own change and its own test rather than
riding along on a determinism fix.

## Repro

```bash
ls "$(python3 -c 'import tempfile;print(tempfile.gettempdir())')"/perry_llvm_*.ll | wc -l
perry compile <anything-not-compiled-before>.ts -o /tmp/x.o --no-link --no-cache
[...]

PR #8072, opened 2026-08-14
fix(codegen): in-process compiles leave no empty scratch directory

jdalton:
## Summary

Every in-process LLVM compile leaves an empty directory behind in the temp dir. This has kept the `repsel-census` CI job red on `main` since 2026-08-05, and — because the leak check runs *before* the census steps — it has also stopped the census gate itself from running at all for the last nine days.

## The problem

A compile gets its own scratch directory named `perry_llvm_scratch_<pid>_<counter>`, and everything that compile needs goes inside it. When the compile finishes, that directory should go away.

The clang path does this correctly. It removes the whole directory:

```rust
// Ours alone: `remove_dir_all` rather than unlinking the two names we
// know about, so anything clang chose to drop beside them goes too and
// the directory cannot survive as an empty husk.
let _ = fs::remove_dir_all(&scratch_dir);
```

The in-process path removed the two files it knew about and left the directory standing:

```rust
if !policy.keep {
    let _ = fs::remove_file(asm_path);
    let _ = fs::remove_file(&plan.obj_path);
}
```

Nothing went wrong while clang was the default. The leak started when the in-process backend became the default and the cleanup that only ever existed on the other path stopped running.

Only one in-process arm is affected: the statepoint arm, which is the only one that *creates* the directory, because it writes the assembly there before the compact-map rewrite. The other arms create it only under `PERRY_LLVM_KEEP_IR`, where keeping it is the point.

<details>
<summary><b>Why this is worse than a pile of empty directories</b></summary>

CI reports it as:

```
TEMP FILES LEAKED: 58 compile(s) left 58 entries in a temp directory that started empty
```

One per compile. That is the detail that separates it from the `#7144` leak the same gate was built for, which was bounded by *distinct IR* — this one grows with every compile.

The bigger cost is what it hides. `Temp-directory hygiene` runs before `Promotion census` and `Sabotage check`, so both are **skipped** on every red run. The census gate has not executed on `main` since 2026-08-04.

`repsel-census` is deliberately not a required status check yet — the job's own comment explains that a gate which has never been green would block every open PR. The consequence is that eight consecutive red nightlies blocked nothing, so nothing forced anyone to notice the gate had stopped running.
</details>

## Changes

- `crates/perry-codegen/src/linker.rs` — the statepoint arm now removes the scratch directory instead of the two files inside it, matching what the clang path has always done.
- `crates/perry-codegen/src/linker_temp_lifecycle_tests.rs` — a regression test.
- `changelog.d/8072-inprocess-scratch-dir-leak.md` — changelog fragment.

## Related issue

Refs #7777 — this is what is currently keeping that job red. The census regression the issue itself is about is a separate matter, and worth re-checking once the census can run again, since it has been skipped for the whole window.

## Test plan

The existing lifecycle tests could not have caught this: they drive `compile_ll_to_object_in`, which takes the clang path, and clang's cleanup was never broken.

The new test pins the statepoint lowering with `native_roots = true` passed directly to `compile_ll_inprocess_in` rather than trusting the host default. That explicit decision is what makes the test mean something — only the statepoint backends route through assembly, and only that arm creates the directory, so without it the test passes on a host that never enters the arm at all.

**Verified by sabotage.** Putting the two `remove_file` calls back turns the new test red:

```
    linker::linker_temp_lifecycle_tests::inprocess_statepoint_compile_leaves_no_empty_scratch_dir
```

With the fix in place:

```
cargo test -p perry-codegen --lib linker
test result: ok. 32 passed; 0 failed
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed in-process LLVM compilations leaving behind empty temporary scratch directories.
  - Improved cleanup of all temporary compilation artifacts after assembly-based builds, preventing temporary-directory growth.

- **Tests**
  - Added regression coverage verifying successful compilation, nonempty output, and complete scratch-directory cleanup across repeated builds.


proggeramlug (commented):
Independent exact-head audit of `ba1f2a1eadb0d8e14b69f583754d176cdb86c2a7` against current `origin/main` `601a02d235af2d9af1841089426a1da379edac0a` found a landing compile blocker.

#8071 changed `compile_ll_inprocess_in` to take a fifth `native_roots: bool` argument and routes that explicit producer decision into `build_clang_compile_plan`. GitHub’s generated merge `1ac543d774d37691456d7d3bc7fb88e00311baa8` retains the new five-argument definition in `linker.rs` but leaves this PR’s new call at `linker_temp_lifecycle_tests.rs:468` with four arguments:

```rust
compile_ll_inprocess_in(&root, &test_ir(100 + nth), None, CLEAN)
```

The test target therefore cannot compile after landing. Please rebase onto current main and pass explicit `true` for this deliberately statepoint/native-root arm. The test comment/pin should also be adjusted: after #8071, this direct helper call is controlled by the explicit boolean, not by `NativeRootsPin::native()`.

After that integration repair, rerun the focused linker lifecycle test and its cleanup sabotage. The localized `remove_dir_all` production change itself looks sound in source audit, and I found no version-bump or issue-closure scope problem.

proggeramlug (approved):
Approved at exact maintainer-updated head `243e61d971e9ec07b8b44707684d7969e29f6c85`. The branch is rebased onto current main; the statepoint regression now supplies the explicit producer native-root decision required by #8071 and passes 1/1 while compiling three real objects and observing an empty temp root after each. Production cleanup is scoped to the per-compile scratch directory. No CI result used.

Issue #8064, opened 2026-08-14
Verify #8078 removes the split-module unknown-wrapper collision


proggeramlug:
## Status

The code fix is merged on `main`:

- PR: #8078
- audited head: `6772b8111aeeb047d4ebbcebcfce7a0e8a7e9736`
- merge commit: `5c27d1ad9b588f69fa2677acd8ae99dc5303e712`
- focused emitted-IR regression: passed
- two separately split modules + real shared object link regression: passed
- landing-tree `perry-codegen` suite: 996/996 passed

The original global symbol `___perry_wrap_perry_unknown_func` is replaced by stable module-scoped fallback names while unsplit copies retain internal linkage.

This ticket now has one bounded purpose: verify that the original production-sized duplicate-symbol boundary is gone. It is not a tracker for every later Next.js, provider, GC, or App Route problem.

## Original production boundary

On `b0f13fb5ef39eb26463d4bc82f7cd6812acea69f`, the pinned Next.js 16.3.0 webpack App Route graph from #8034:

- passed the byte-identical 21-request Node oracle;
- collected and emitted 91/91 Perry module objects;
- cleared the former split-unit `ptrtoint` failures;
- failed the final app-only dylib link because four objects exported the same `___perry_wrap_perry_unknown_func`.

The old local replay log was:

`/Users/amlug/cargo-targets/next-prod-gate/evidence/perry-route-compile-b0f13fb5.log`

Do not depend on that path existing. Regenerate the pinned fixture from #8034 when necessary.

## Remaining worker task

1. Fetch current `main` and record the exact commit/tree.
2. Generate/build the pinned Next.js 16.3.0 webpack fixture from #8034. Confirm the Node oracle inputs are the generated production route graph; a direct call to the user `GET` export is not a substitute.
3. Build one coherent release artifact set from that same Perry SHA in an isolated target:

   `cargo build --release -p perry -p perry-runtime-static -p perry-stdlib-static -p perry-ext-http -p perry-ext-net`

4. Record SHA-256, size, and mtime for `perry`, `libperry_runtime.a`, `libperry_stdlib.a`, `libperry_ext_http.a`, and `libperry_ext_net.a`; prove no writer remains before the replay.
5. Compile the same app-only graph from scratch with the isolated release directory pinned:

[...]

PR #8078, opened 2026-08-14
Scope split unknown-function fallback wrappers by module

proggeramlug:
## Summary

- derive the unknown FuncRef sentinel and fallback wrapper name from the stable module prefix
- share that naming helper between expression lowering and wrapper emission
- preserve internal linkage for unsplit modules while keeping split-unit promotion collision-free
- add emitted-IR coverage plus a two-module split-object final-link regression
- make no version, manifest, lockfile, or changelog changes

## Validation

- cargo fmt --all -- --check
- cargo test -p perry-codegen --lib (994 passed)
- the focused object regression compiles two codegen-unit-split modules and merges both module objects in one final link

I also attempted the preserved 91-module Next.js graph while priming its specialized archives. It collected all 91 modules, completed the three-unit webpack object, and entered the two-unit jsonwebtoken object. The memory-capped run was stopped after 30 minutes while still inside an LLVM function pass at 12/91, before the application link, so no full replay verdict is claimed.

Refs #8064

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unresolved function fallback symbols from colliding across modules and split compilation units.
  * Preserved correct linking behavior for both split and unsplit modules.
  * Ensured each module references its own fallback wrapper.

* **Tests**
  * Added regression coverage for separate module compilation, LLVM output, and combined linking.

### DIFF
--- a/crates/perry-codegen/src/codegen/artifacts.rs
+++ b/crates/perry-codegen/src/codegen/artifacts.rs
@@ -17,7 +17,9 @@ use super::closure::{
     compile_typed_i32_closure, compile_typed_string_closure,
 };
 use super::entry::compile_module_entry;
-use super::helpers::{function_body_returns_generator_object, sanitize, scoped_fn_name};
+use super::helpers::{
+    function_body_returns_generator_object, sanitize, scoped_fn_name, unknown_func_wrapper_name,
+};
 use super::method::{
     compile_method, compile_static_method, compile_typed_f64_method,
     compile_typed_f64_receiver_method, compile_typed_i1_method, compile_typed_i32_method,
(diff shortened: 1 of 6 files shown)


proggeramlug:
Maintainer audit: APPROVE / merge-safe at exact head 6772b8111aeeb047d4ebbcebcfce7a0e8a7e9736.

I tested the clean synthetic landing on current main 3f468adea28e9e5d285807abb0798637f4002698 (landing tree 736092c44a953b5069b6ab99068de5c00d15f597). The merge is conflict-free and its delta remains exactly the six intended codegen files.

Source audit:
- compile_module derives the existing sanitized module prefix once from the HIR module name;
- the unresolved FuncRef reference and emitted fallback definition now use the same shared naming helper;
- unsplit wrappers retain internal linkage;
- split-unit promotion remains unchanged, but promoted names are module-scoped, so independently compiled module objects cannot collide;
- repository-wide search found no remaining production use of the legacy global fallback symbol.

Exact landing-tree evidence with an isolated Cargo target:
- unknown_funcref_uses_its_module_scoped_wrapper: 1/1 passed;
- split_modules_keep_unknown_fallback_wrappers_distinct_at_shared_link: 1/1 passed, exercising two separately split modules and the real final object merge;
- complete perry-codegen library suite: 996/996 passed;
- cargo fmt --all -- --check and git diff --check passed;
- worktree remained clean.

No CI result was used as correctness evidence. The preserved 91-module Next replay was not completed, so I am changing the issue linkage from Closes #8064 to Refs #8064 and leaving the production acceptance issue open after this code lands. No version bump.

Issue #7302, opened 2026-08-03
codegen: lower try/catch to LLVM invoke/landingpad, delete the setjmp/longjmp path


proggeramlug:
Perry lowers `try`/`catch` to `setjmp`/`longjmp` (`perry-codegen/src/stmt/try_stmt.rs`; zero uses of `invoke` today). That single choice causes three separate problems:

1. **Precise moving-GC roots are unsound in `try` functions** — a `longjmp` can jump past a `gc.statepoint`'s `gc.relocate`, so the relocated pointer's write-back never runs. This is why the statepoint experiment (#7174) excludes `has_try` functions and routes them to a plain-stack-map fallback that is itself unsound (root slots recorded as caller-saved registers; 3/60 locations on one probe). Under RS4GC, `mem2reg` cannot promote the volatile allocas setjmp requires, so try-region roots never enter SSA and never join a `gc-live` bundle.
2. **~570 lines of machinery exist only to fight the register allocator**: `volatile_setjmp.rs` (376) + `setjmp_abi.rs` (193) implement C99 7.13.2.1p3.
3. **Every `try` function is pessimized**: `returns_twice` + `#1` (`noinline`) on the whole function.

Moving to `invoke`/`landingpad` collapses all three and is the prerequisite for #7174 deleting its unsound `has_try` fallback (statepoints can then cover every function).

## Plan

- **Phase 0 (done)** — standalone spike: invoke + landingpad + catch-all against a `_Unwind_RaiseException` throw path. All green on macOS arm64, including bit-exact NaN-box payload round-trip, nested rethrow, finally re-raise, uncaught. Decisions: Perry-owned `perry_eh_personality` (Itanium LSDA walk; no libc++abi, no dependence on `rust_eh_personality`); thrown value stays in the GC-rooted TLS slot (landing pad ignores `{ptr, i32}`); `js_throw` keeps its savepoint restores and raises; `js_call_catching` keeps its Rust-boundary longjmp trap (Rust cannot catch foreign exceptions; a raise never crosses an *open* `js_call_catching` frame).
[...]

PR #8082, opened 2026-08-14
fix(next): pass production App Route dylib gate

### LINKED ISSUE #8034: [Next.js/dylib] Add a pinned production App Route parity fixture and CI gate

proggeramlug:
## Summary

Closes #8034 and #8036: a pinned Next 16.3.0 **production** App Route fixture, plus the runtime and codegen fixes needed to make the untouched production handler run as an app-only dylib behind separately loaded runtime and stdlib provider images.

No project version bump.

## What the fixture asserts

`tests/release/packages/next-app-route/` — picked up automatically by the tier-3 release harness:

1. `npm ci` + `npm run build` for Next 16.3.0, then a sanity check that the emitted bundle really contains `AppRouteRouteModule` and `routeModule.handle` (so a Next change that stops producing that shape fails here rather than silently weakening the test).
2. The **exact Node production oracle** for the same 21-request workload.
3. Coherent runtime and stdlib provider archives, with the stdlib's bundled copy of `perry-runtime` trimmed and asserted gone (`nm` for `js_gc_init`), so the separately loaded runtime image is the single owner of GC and event state.
4. Perry compiles the **untouched** webpack output as an app-only dylib, asserted not to embed the Perry ABI.
5. 10 cold processes, two 21-request verifier passes each, compared against the oracle.

The forced-evacuation arm is opt-in behind `PERRY_NEXT_ROUTE_FORCED_GC=1` and is **currently red** — tracked as **#8163**, which carries the full elimination trail and a seconds-long reproducer. It is deliberately neither a `SKIP` (which would read as covered) nor `continue-on-error` (which would make it documentation rather than a gate): off by default, failing loudly when set.

## Production-path fixes in this PR

- **Native statepoint roots are kept in app dylibs.** The earlier demotion of `--output-type dylib` artifacts to the shared shadow stack predated #8081's loaded-image stack-map indexing; with that in place the demotion would leave provider apps running a lowering production never ships, and it defeated #8081's own assertion that the app's map survives macOS dead stripping.
- **Class-self lowering** respects a same-named method parameter or local instead of forcing the lexical class binding (four focused HIR regressions cover declarations and named class expressions).
- **Computed `require(".")` / `require("..")`** resolve relative to the caller.
- **Dynamic virtual dispatch** builds its direct-call ABI from the selected override's own metadata, including rest and synthetic `arguments` shape in both override directions.
- **Bound-method construction** roots the receiver across closure allocation and the closure across allocating metadata installation, with a deterministic unit test that forces a moving minor inside the builder and asserts both rewritten addresses.
- **Malformed unwind-table records** are parsed transactionally with checked ranges and offsets.
- **`PERRY_GC_PROTECT_FROMSPACE_HOLDERS=1`** — at a from-space fault, sweep the live heap for any word still naming the faulting address and print the owners. The existing report answers "who used it", which for a value read out of a table one instruction earlier is never the bug; this answers "who kept it". It is what proved #8163's holder is outside the GC heap.

Four further fixes this branch found have already landed separately and are no longer part of this diff: **#8128** (RS4GC inline-asm SIGBUS, the relocation fan-out `optnone` cap, LLVM worker stacks) and **#8131** (the rooting sweep, `perry_ffi::TransientRootScope`, instruments, the action-zero landing-pad regression test).

## Validation

Fixture, default mode: **PASS** — 10 cold starts × 2 verifier passes, matching the Node production oracle.

Suites at this head: `perry-runtime --lib` **2421 passed**, `perry-codegen --lib` **1029 passed**, `perry --bin perry` **976 passed**.

Gates: `cargo fmt --check`, `check_file_size.sh`, `raw_handle_debt.py` (**992 = baseline**, no debt added), `addr_class_inventory.py`, `gc_runtime_root_holders.py`, `gc_pin_sites.py` — all clean. The bound-method builders and the holder sweep were converted to the sanctioned handle accessors and `addr_class::try_read_gc_header` rather than raising any ceiling, and the bound-method regression is sabotage-verified (returning the pre-collection address still fails it).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Next.js App Route compatibility, including request-state preservation and runtime-computed module paths.
  * Fixed `Reflect.apply` and method dispatch so all arguments are forwarded correctly.
  * Corrected class self-construction when names are shadowed or renamed.
  * Improved exception propagation across native and generated code.
  * Strengthened garbage-collection safety for bound methods and closure captures.
  * Added safer handling for malformed unwind metadata and native call-depth restoration.

* **Tests**
  * Added comprehensive regression coverage for Next.js production routes, CommonJS wrapping, dynamic dispatch, class construction, and moving garbage collection.


crates/perry-hir/src/lower/expr_new.rs
```diff
@@ -201,9 +201,25 @@ pub(super) fn lower_new(ctx: &mut LoweringContext, new_expr: &ast::NewExpr) -> R
...
+            // webpack bundles commonly have a module-wide `var h` and a
+            // factory-local `class h`; inside `h`'s static `instance()` method,
+            // `new h` must therefore construct the class, not capture the outer
+            // (usually still-undefined) `var h`.
+            //
+            // Use `current_class` rather than merely resolving the source name:
+            // it also carries the unique registration key for collision-renamed
+            // declarations (`h$0`) and named class expressions. All other
+            // identifiers continue through the ordinary scope-local rename map.
+            let is_current_class_self = ctx.current_class_inner_name.as_deref()
```
proggeramlug, line 231: Blocking correctness regression: is_current_class_self ignores an existing method-local binding. In JavaScript, class C { static make(C) { return new C(); } } must construct the constructor passed in parameter C; Node 26 returns true for C.make(D) instanceof D. This branch forces the enclosing class instead because the source identifier matches current_class_inner_name, even when lookup_local finds the parameter. Please distinguish the class lexical binding from nearer method parameters/locals and add this shadowing regression alongside the outer-var positive case.

tests/release/packages/next-app-route/fixture.sh
```diff
@@ -0,0 +1,174 @@
...
+echo "  commit=$PERRY_COMMIT next=16.3.0 mode=dylib providers=$PROVIDER_ABI_HASH"
+
+run_cold_start() {
+  local index="$1"
+  local mode="$2"
+  local port=$((PORT_BASE + index + 1))
+  local log="$BUILD_DIR/perry-${mode}-${index}.log"
+  : >"$log"
+  if [[ "$mode" == "forced" ]]; then
+    env PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1 \
```
proggeramlug, line 158: The forced/verified acceptance arm is vacuous as checked in: it sets FORCE_EVACUATE and VERIFY, but does not positively require a collection or a non-in-place move, and does not enable diagnostics from which that can be asserted. A run with zero collections passes all current checks. Please arm a deterministic moving workload and use the existing evacuation-liveness checker (or an equally strict copied/promoted non-in-place assertion) so closing #8036 proves forced GC was actually exercised.

proggeramlug:
Audit note: this PR changes crates but has no changelog.d/8082-*.md fragment and no skip-changelog label. Repository policy requires the numbered fragment; no version bump is needed. I have also posted two blocking source/test findings on the exact current head.

tests/release/packages/next-app-route/fixture.sh
```diff
@@ -0,0 +1,174 @@
...
+if [[ "$SHARED_EXT" == "dylib" ]]; then
+  mac_libs=(-framework Security -framework CoreFoundation -framework SystemConfiguration -liconv -lresolv -lobjc)
+  cc -dynamiclib -Wl,-force_load,"$RUNTIME_ARCHIVE" -Wl,-undefined,dynamic_lookup "${mac_libs[@]}" -o "$RUNTIME_IMAGE"
+  cc -dynamiclib -Wl,-force_load,"$TRIMMED_STDLIB" -Wl,-undefined,dynamic_lookup "${mac_libs[@]}" -o "$STDLIB_IMAGE"
+else
+  linux_libs=(-lm -lpthread -ldl -lssl -lcrypto)
+  cc -shared -Wl,--whole-archive "$RUNTIME_ARCHIVE" -Wl,--no-whole-archive -Wl,--allow-shlib-undefined "${linux_libs[@]}" -o "$RUNTIME_IMAGE"
+  cc -shared -Wl,--whole-archive "$TRIMMED_STDLIB" -Wl,--no-whole-archive -Wl,--allow-shlib-undefined "${linux_libs[@]}" -o "$STDLIB_IMAGE"
+fi
+cc provider-host.c -ldl -o "$HOST_BIN"
```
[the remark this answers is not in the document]
proggeramlug: I do not reproduce this on the supported Darwin host: Apple Clang linked a minimal dlopen program with -ldl and the resulting binary executed successfully. The PR also records an end-to-end Darwin gate using this exact command. Treating this as nonblocking; separating the platform flags would still be a harmless cleanup.

proggeramlug:
Flagging a correctness-of-evidence problem in this branch's fixture, since it determines whether a green run here means anything.

**`tests/release/packages/next-app-route/perry-host.js` on this branch has no `routeModule.handle` bypass guard.** Measured on both branches:

```
git show origin/fix/8036-production-app-route:tests/release/packages/next-app-route/perry-host.js | grep -c 'bypassed routeModule.handle'   -> 0
git show origin/fix/8037-next-route-als-production:tests/fixtures/next-app-route/perry-host.js  | grep -c 'bypassed routeModule.handle'   -> 1
```

The sibling copy wraps `routeModule.handle` and throws `"<id>: generated handler bypassed routeModule.handle"` if a request completes without going through it. Without that assertion, a verifier can pass 21/21 while the userland `GET` was called directly and Next's private `handle` pipeline — AsyncLocalStorage work store, request context, response construction — never ran. That is precisely the failure mode #8040 opens by ruling out ("no direct userland `GET` invocation, ignored `Response`, hard-coded output"), and it is what disqualified `benchmarks/next-small` as evidence.

This branch's layout is otherwise the better one — split `provider/runtime` + `provider/stdlib`, `fixture.sh`, `provider-host.c`. Suggest taking that layout and `fix/8037`'s `perry-host.js`.

**Second point, worth building into whatever gate lands: the guard's signal is in the host log, not the exit code.** Sabotage-verified — removing only the wrapper install makes it fire 21/21 times, yet `verify.mjs` still exits **0**, because the assertion throws inside a `.then()` that is caught and logged after the response has already been sent. A CI gate that checks only the verifier's exit status therefore cannot detect a bypass. It has to grep the host log for that string (and, ideally, assert the wrapper was installed at all, so a fixture that silently stops installing it fails loudly).

Context: with the chunk-resolution fixes in #8146 and a `panic = "abort"` correction to the provider workspace (#7302's third instance — that workspace's `[profile.release]` has no `panic` key, so `js_throw` gets RFC-2945 abort guards and any JS throw kills the process), this fixture is close to executing a real route for the first time. Worth making sure the thing it proves is the thing we want proven.

proggeramlug:
Out of draft. The forced-evacuation arm is split to #8163 and the fixture now gates the production path on its own.

**Why split rather than hold the PR.** The fixture's production-parity coverage passes today; one arm of it does not, and that arm's bug is independent of everything else here (it reproduces identically under `PERRY_RS4GC=0`, so it is not the dylib native-roots change this PR makes). Holding a working production gate hostage to it buys nothing.

**How the arm is disabled matters**, so it is worth being explicit: not a `SKIP` (which counts as covered and would let the gap read as tested), and not `continue-on-error` (which makes a job documentation rather than a gate — CLAUDE.md's first "way a gate cannot fail"). It is `PERRY_NEXT_ROUTE_FORCED_GC=1`: **off here, fails loudly when set**, and the `PASS` line names which arms ran so a green result cannot overstate itself. When #8163 is fixed, flipping the default back is a one-line change with the liveness assert already wired.

**Full fixture run at `dfd6231b5`, default mode:**

```
  [1/7] install and production webpack build (Next 16.3.0)
  [2/7] exact Node production oracle
  [3/7] build coherent runtime and stdlib provider archives
  [4/7] link separate provider images and dlopen host
  [5/7] compile untouched production route handler as app-only dylib
  commit=6eeb1a5bf next=16.3.0 mode=dylib providers=635b3b523c…
  [6/7] 10 cold processes, two 21-request verifier runs each
        forced-evacuation arm OFF (#8163) — set PERRY_NEXT_ROUTE_FORCED_GC=1 to run it
  [7/7] production AppRouteRouteModule.handle parity complete
PASS next-app-route (forced-evacuation arm not run — #8163)
```

Checked that the subject actually ran rather than trusting the line: 10 cold-start logs, **20/20** verifier passes (10 starts × 2), against the Node oracle.

**Also folded in on the rebase**, since #8128 and #8131 landed: their commits dropped out by patch-id (17 → 9), and the branch's own landing-pad regression went with them — the review pass here had added a Handler/Cleanup split keyed on the LSDA action, which `main` never had and which breaks every statepoint-built `try`/`catch` (#7982 makes every JS catch pad an action-zero cleanup pad). `main`'s `eh.rs` is authoritative and is now taken wholesale. Two ratchets that this branch's own code tipped — raw-handle debt in the bound-method builders, an addr-class cast in the holder sweep — are converted to the sanctioned accessors rather than ceiling-raised: **992 = baseline**, no debt added.

Suites at this head: `perry-runtime` 2421, `perry-codegen` 1029, `perry --bin perry` 976. Gates: fmt, file-size, raw-handle, addr-class, root-holders, pin-sites all clean.

proggeramlug:
I rebased this onto current `main` locally (it was 21 behind, six conflicts) and ran the full gate battery. **Two blockers, and the first is in this PR's own subject area.** Not pushing my rebase — see the resolution note at the end, you should decide those four.

## 1. 18 new bare raw-handle reads, in the rooting code

```
bare raw-handle reads: 1008 (baseline 990)
crates/perry-runtime/src/object/native_module.rs: 22 bare reads exceeds its ceiling of 4
```

`main` is clean at 990/990. This branch adds **18**, all in `native_module.rs` (+108/−19), and it does not touch `scripts/raw_handle_debt_files.txt` — so it is new code, not the "split oversized files" commit relocating existing reads.

What makes this worth stopping for rather than waving through: two of this PR's own commits are **`fix(runtime): root bound method construction`** and **`fix(codegen): keep native statepoint roots in app dylibs`**. The ratchet exists for the #7154 class — a GC value live across a collection point with no root — which is invisible at collection time and surfaces cycles later as `TypeError: value is not a function`. Adding 18 unrooted-shape reads inside the file you are fixing rooting in is exactly the shape it is meant to catch.

The policy is convert, not raise the ceiling, and it is being held to: two PRs merged tonight (#8186, #8177) hit this ratchet and converted rather than exempting. `across_{mut,const,nanbox}` for a post-call reload, `with_{mut,const}_ptr` for a scoped argument to a non-allocating or self-rooting callee — the split is by whether the line starts with `let`.

I did not convert them myself. Eighteen rooting-shape decisions in a 12.5k-line PR need the author who knows which values are live across which calls.

## 2. An unclassified holder — this one is trivial

```
crates/perry-runtime/src/object/native_module.rs:1280:
  TEST_BOUND_METHOD_MOVE: std::cell::Cell<(usize, usize)>  [rule B]
```

The name says it: test scaffolding. `scripts/gc_runtime_root_holders.json` already has a `test_only` verdict category for exactly this. One entry.

## Everything else passes

`cargo fmt`, `check_file_size`, `check_gc_env_knobs`, the **new** `shape_descriptor_census` gate (#8110, merged tonight), `addr_class_inventory`, `gc_store_site_inventory`, `local_binding_type_audit`, `gc_gate_wiring_check`, `check_test_registration`, `gc_pin_sites`, `global_sink_isolation`, `class_id_collisions`, `gc_root_dominance --audit-poll-reach`, `workspace_architecture`.

## On the rebase, and four resolutions you should check

Six conflicts, in four files. **#8146 and #8153 were extracted from this PR and merged tonight in refined form**, so I resolved all four to `main`'s side:

- `crates/perry-hir/src/lower/expr_new.rs` — `main` has #8153's `forward_class_shadows_local` depth rule, which also keeps the mysql2 case working (a module-scope `class e` must not beat a factory-local `let e`). This branch had the earlier `is_current_class_self` form.
- `crates/perry-codegen/src/codegen/entry.rs`, `cjs_wrap/wrap.rs`, `cjs_wrap/preamble_canary_tests.rs` — `main` has #8146's versions.
- `crates/perry/src/commands/compile/build_cache.rs` — both sides added a knob to the same sorted list (`PERRY_LL_O0_MAX_FN_BYTES` from #8144, `PERRY_LL_RS4GC_OPTNONE_INSTRS` from yours). Kept both.

If this PR had unique content in those four beyond what #8146/#8153 carried, my resolution dropped it — which is why I am describing it rather than pushing it.

Rebase result for reference: 5 commits, 0 behind `main`, no conflicts remaining, 58 files, +10290/−320.

proggeramlug:
CodeRabbit's 8 comments are addressed, and chasing the CI red uncovered something bigger than the review did. Head is now `a5415a5fd`.

## The CI red was a bad rebase, not the code

19 failing checks — `cargo-test`, both `Warnings` jobs, all 8 `conformance-smoke` shards, `compiler-output-regression`, `e2e-scoped` — traced to one cause: **rebasing this branch onto a main that had moved ~50 commits reverted 14 merged PRs** (#8097–#8186). Their changelog fragments and source files were deleted and main's newer edits to shared files were undone. The `Warnings` job named the tip of it: an `unused doc comment` at `stack_maps.rs:209`, which is a doc comment on a `perry_thread_local!` invocation — main had already fixed exactly that in **#8176**, and my replay put it back.

A 3-way merge cannot revert like that, so the branch is now `main` + a merge of this work, with conflicts resolved toward main wherever main has since improved the file (`eh.rs`, `array/generic.rs`, `stack_maps.rs` taken wholesale — main already carries this branch's landing-pad semantics, the arraylike accessor conversions and the stack-map trace via #8131, plus #8176's comment form and #8164's env-flag polarity). `fromspace_scan.rs` keeps main's file (#8084's counted slack bound + the payload preview) with only the owner/target header dump re-added. **`git diff origin/main` now deletes nothing.**

That also surfaced a real defect the gate caught rather than review: main's panic-profile contract (#8147) rejected the new provider workspace, which declared no `panic` key and so silently built a Perry runtime archive on `unwind`. That aborts the process on any JS throw crossing an `extern "C"` helper with an interior Rust call (RFC 2945), and `eh.rs`'s transport is written for `abort` semantics. Now declared, with the reason in the manifest.

## Review comments

**Fixed:**
- **Changelog accuracy** — it still claimed half the cold starts run under forced evacuation, which my own opt-in change had made false. Corrected to state the arm is opt-in, off by default, and why (#8163).
- **Unbounded holder sweep** — agreed, and for a sharper reason than runtime: it runs in a signal handler, so an unbounded walk can lose the re-fault to a CI timeout. Now budget-bounded, and an exhausted budget reports as *"sweep did NOT finish … this is not evidence of absence"* rather than as "no holder". Conflating those two is how an instrument starts lying.
- **Method-local shadowing test** — correct that local-declaration lowering is a separate path from parameters. Added, and sabotage-verified: removing the shadowing check fails both that test and the parameter one, so it is not vacuous.
- **Method-name length from the literal**, and **the computed-require assertion** no longer embeds the emitter's newline and indentation.

**Skipped, with reason:** the `tempfile::tempdir()` and blanking-assertion comments target pre-existing code that this branch's file split merely *relocated* into `tests.rs` — changing it would be unrelated churn. The "redundant handle reloads in `build_symbol_bound_method_closure`" was already resolved by converting that builder to `with_mut_ptr` for the raw-handle ratchet.

## Validation at `a5415a5fd`

Fixture, default mode: **PASS** — 10 cold starts, **20/20** verifier passes against the Node production oracle.

`perry-runtime` 2483 · `perry-codegen` 1039 · `perry --bin perry` 986 · `perry-hir` shadowing 5. Gates: fmt, file-size, raw-handle (992 = baseline), addr-class, root-holders, pin-sites — all clean.

proggeramlug:
**Retracting my earlier blocker comment on this PR — both blockers were my error, not defects in this branch.**

I rebased from a tree that had silently **dropped 5 pushed commits**, and reported against that. The dropped set includes exactly the two fixes I claimed were missing:

- `176946821` — classify the bound-method test hook in the root-holder registry (my "unclassified `TEST_BOUND_METHOD_MOVE`")
- `6eeb1a5bf` — sanctioned handle accessors in the bound-method builders (my "18 new bare raw-handle reads")

On the live branch head `a5415a5fd`, `scripts/raw_handle_debt.py` reports **4** bare reads in `native_module.rs` and the holder entry is present. Neither blocker exists. Sorry for the noise.

I also audited those 18 conversions on their merits and they are correct: argument-position reads use `with_mut_ptr`, the single post-call reload uses `across_mut`. The one substantive semantic change is *grouping* — three (and five) capture writes now share one pointer read where the old code re-read between each. That is sound only if `js_closure_set_capture_*` cannot collect, which checks out: it reaches `layout_note_slot` and `runtime_write_barrier_gc_slot`, and there is no collection trigger anywhere in `gc/barrier/` or `gc/layout.rs`. `set_bound_native_closure_name` looks like a counterexample because it allocates via `js_string_from_bytes`, but it roots its incoming pointer first — the sanctioned self-rooting-entry-point shape.

**One thing worth flagging loudly for anyone who rebases this branch.** The same dropped-commit set contained `b20337a88 fix(runtime): drop this branch's landing-pad regression`. Without it the tree carries the `Handler`/`Cleanup` split in `perry_eh_personality`; since #7982 makes every JS catch pad a token-cleanup pad with a **zero** action, that split skips every statepoint-built catch and a plain `try/catch` aborts with FATAL "no landing pad". A rebase that loses those 5 commits produces a tree that looks plausible and is badly broken.

**Current state.** The branch is `CONFLICTING` against `3be2016c1` and missing 7 main commits (#8110, #8161, #8144, #8146, #8153, #8168, #7312). Validated resolutions for all five conflicting files are pushed to `rebase/8082-validated-resolutions` (`36f7c5229`) — cherry-pickable, and **not** a substitute for this branch (it lacks the 8 fixes force-pushed at 07:39).

Notes on those resolutions:
- `expr_new.rs` → take **main's** `forward_class_shadows_local`. All 4 of this PR's `class_self_new_shadowing` tests pass under it, and `method_parameter_shadows_class_self_name` would likely fail under the branch's `is_current_class_self`.
- `wrap.rs` → do **not** take main wholesale; commit 2's bare `'.'`/`'..'` join is functional (`js_require_path_module` resolves directories via `directory_module_candidates`) and is named in this PR's changelog.
- `build_cache.rs` → keep both knobs; they are two distinct sorted lists.
- Two `cjs_wrap` tests pin this branch's ternary form and break under #8146's, now on main. Retargeting is in the same branch.

Validation of the rebased tree (11 commits, 0 behind): 17 gate scripts green including `raw_handle_debt --no-raise-vs origin/main` at 990/990 with no ceiling raised; `perry-runtime --lib` 2486/0; `perry-hir` 523/0; `perry-codegen --no-fail-fast` 1483/9 against a main baseline of 1480/9 with **failure sets identical by name**; `perry --bin perry` 987/0.

Not pushing to this branch — it has an active owner.

Issue #7219, opened 2026-08-01
buffer fast path: Uint32Array writes are lost after constructing Uint8Array(words.buffer)


andrewtdiz:
## Summary

On Perry 0.5.1277, a write through a fresh `Uint32Array` is not visible through a `Uint8Array` created from its `.buffer` when the buffer fast path is enabled. Compiling the identical source with `--disable-buffer-fast-path` produces the correct result.

This is narrower than #579: the ArrayBuffer-first reproduction from that issue now passes on the same build. The failing case starts with typed-array-owned storage and writes through the original typed array *after* its `.buffer` has been used to construct another view.

## Minimal reproduction

```ts
function reproduce(): number {
    const words = new Uint32Array(1);
    const bytes = new Uint8Array(words.buffer);
    words[0] = 0x01020304;
    return bytes[0] + bytes[1] + bytes[2] + bytes[3];
}

const expected = 10;
const actual = reproduce();
console.log(JSON.stringify({ expected, actual, passed: actual === expected }));
```

The expected sum is endian-independent: the four bytes are `1`, `2`, `3`, and `4` in either byte order.

## Commands and output

Fast path enabled:

```text
> perry compile alias-pure-repro.ts -o alias-fast.exe --no-cache
> ./alias-fast.exe
{expected:10,actual:0,passed:false}
```

Fast path disabled:

```text
> perry compile alias-pure-repro.ts -o alias-safe.exe --no-cache --disable-buffer-fast-path
> ./alias-safe.exe
{expected:10,actual:10,passed:true}
```

Control using the original ArrayBuffer-first shape from #579:

```ts
const ab = new ArrayBuffer(4);
const v1 = new Uint8Array(ab);
v1[0] = 10;
v1[1] = 20;
v1[2] = 30;
v1[3] = 40;
const v2 = new Uint8Array(ab);
console.log(v2[0], v2[1], v2[2], v2[3]);
```

```text
10 20 30 40
```

## Emitted LLVM evidence

`--trace llvm` shows that the data pointer for `words` is cached before the `.buffer` lookup and `js_uint8array_new` call:

```llvm
%r34 = inttoptr i64 %r33 to ptr
%r35 = getelementptr i8, ptr %r34, i32 16
store ptr %r35, ptr %r36
...
%r114 = call i64 @js_uint8array_new(double %r113)
...
%r138 = load ptr, ptr %r36
...
store i32 %r150, ptr %r143
```

[...]

Issue #1205, opened 2026-05-20
Improve node:buffer slice/subarray shared backing-store parity


TheHypnoo:
## Context
The granular `node:buffer` parity suite covers deterministic `slice` and `subarray` range behavior, but it still avoids Node's shared-memory mutation semantics.

## Gap
In Node, `buf.slice()` and `buf.subarray()` return views over the same backing memory. Mutating the returned view mutates the original buffer, and mutating the original buffer is visible through the view.

Perry currently uses copy-like buffer results for the passing subset.

## Desired behavior
- `const s = buf.slice(1, 3); s[0] = 0x61;` updates `buf[1]`.
- `const v = buf.subarray(1, 3); buf[1] = 0x62;` updates `v[0]`.
- Existing Node-compatible range handling should be preserved.

## Suggested implementation notes
This likely needs a Buffer view representation rather than another copy-only allocation path. Options include extending `BufferHeader` with backing pointer/offset metadata or introducing a separate view header that all buffer read/write/search methods understand.

## Parity tests to add
- `test-parity/node-suite/buffer/copy-slice/slice-shared-mutation.ts`
- `test-parity/node-suite/buffer/copy-slice/subarray-shared-mutation.ts`

## Validation
- `./run_parity_tests.sh --suite node-suite --module buffer`

PR #8083, opened 2026-08-14
fix(runtime): make a DataView read the backing store, not its construction snapshot

jdalton:
## Summary

A `DataView` does not see writes made through a `Uint16Array`, `Uint32Array` or `Float64Array` over the same `ArrayBuffer`. It reads the bytes that existed when it was constructed. `TextDecoder.decode(dataView)` has the same bug.

```ts
function u32writer(): number {
  const ab = new ArrayBuffer(4);
  const w = new Uint32Array(ab);
  const dv = new DataView(ab);
  w[0] = 0x01020304;
  return dv.getUint8(0) + dv.getUint8(1) + dv.getUint8(2) + dv.getUint8(3);
}
console.log(u32writer());   // perry: 0   node: 10
```

The write really does land in the shared buffer — a `Uint8Array` reader over the same `ArrayBuffer` sees it. Only the DataView misses it.

## Root cause

`js_data_view_new` gives every DataView its own `BufferHeader` and copies the window's bytes into it. That copy is refreshed only by writes routed **through** the buffer-view registry: `js_buffer_set`, `js_buffer_write`, a sibling DataView's `set*`.

A multi-byte typed-array element store does not go through the registry. `register_view_meta` makes `typedarray::data_ptr_mut` resolve straight into the backing `ArrayBuffer`, so the store lands there and nothing mirrors it into the DataView's private copy.

`read_bytes` (`crates/perry-runtime/src/buffer/dataview.rs`) read that private copy. The registry edge to the backing store was already there — `read_buffer_byte` has followed it for `Uint8Array`/`Buffer` receivers since #1205. `read_bytes` simply never used it.

<details>
<summary><b>Why it looked like an element-width bug, and why it isn't</b></summary>

The trigger is not width, it is which header the writer gets. A Perry `Uint8Array` over an `ArrayBuffer` is a `BufferHeader` view, and its writes go through the registry, so they mirror. `Uint16Array`/`Uint32Array`/`Float64Array` are `TypedArrayHeader` kinds that bypass it.

That also explains the other bounds:

- **Direction.** Writing through the DataView and reading back via the typed array always worked — that write goes through the registry.
- **Construction order.** Writing before constructing the DataView worked, because the copy captures whatever is there at that moment.
- **Not a codegen tiering issue.** It fails on the pure ArrayBuffer-first shape, which never had an inline-storage proof to lose, and `--disable-buffer-fast-path` did not rescue it.
</details>

## Changes

| file | change |
|---|---|
| `crates/perry-runtime/src/buffer/dataview.rs` | `read_bytes` resolves through `resolve_data_ptr` instead of reading its own inline bytes |
| `crates/perry-runtime/src/text.rs` | `TextDecoder.decode(dv)` built its slice as `buf + sizeof(BufferHeader)` under a comment asserting the bytes were "stored inline" — the same stale read |
| `crates/perry-runtime/src/buffer/mod.rs`, `text.rs` | three unit tests |
| `test-files/test_gap_dataview_buffer_aliasing_7219.ts` | new gap test, byte-exact against node |

The construction-time copy is deliberately **not** removed: codegen still `gep`s against buffer pointers and `write_bytes` still writes it. Only which copy is authoritative *on the read* changed.

## Test plan

Every case byte-compared against node 26.5.1, the version `.node-version` pins.

| case | node | before | after |
|---|---|---|---|
| Uint8 writer → DataView | `1,2,3,4` | ✅ | ✅ |
| Uint16 / Uint32 / Float64 writer → DataView | `2,1,4,3` / `4,3,2,1` / `…248,63` | ❌ all zeroes | ✅ |
| DataView → typed array, four widths | — | ✅ | ✅ |
| DataView constructed after the writes | `4,3,2,1 \| 8,7,6,5` | ❌ half zeroes | ✅ |
| windowed `new DataView(ab, 4, 8)` | `4,3,2,1,8,7,6,5` | ❌ zeroes | ✅ |
| endianness, `getInt16/32/Float64` ± the LE flag | `258,513,-50462977,…` | ❌ | ✅ |
| two DataViews over one buffer | `61374,61374` | ❌ `0,0` | ✅ |
| module scope | `4,3,2,1` | ❌ zeroes | ✅ |
| `w.buffer === dv.buffer`, byteOffset/byteLength | `true,true,4,4,4` | ✅ | ✅ |
| SAB, `.buffer`-materialized, `Buffer.from(ab)`, subarray | — | ❌ | ✅ |
| `TextDecoder.decode(dv)` | `ABCD` | ❌ `\0\0\0\0` | ✅ |

`RUST_TEST_THREADS=1 cargo test --release -p perry-runtime` → **2317 passed, 0 failed**. (That flag is required — the runtime's tests share process-global side tables and are not parallel-safe.) `test_gap_6386_dataview_concat_regex_fastpaths.ts` still matches. rustfmt, `check_file_size.sh` and the addr-class ratchet are clean.

**Verified by sabotage**, each fix reverted independently:

```
data_view_reads_multi_byte_typed_array_writes
  kind=3: DataView byte 0 lags the typed-array write   left: 0.0  right: 2.0

multi_byte_typed_array_reads_windowed_data_view_writes
  windowed DataView byte 4 must mirror backing byte 8  left: 0.0  right: 239.0

text_decoder_reads_backing_store_of_a_data_view
  left: "\0\0\0\0"  right: "ABCD"
```

Each test first asserts the typed-array store actually reached the `ArrayBuffer`, so a store that never landed cannot pass it vacuously on `0 == 0`. The windowed test's DataView→typed half stayed green under sabotage, which independently confirms that direction was never broken.

## Related issue

Refs #7219. That issue's own reproducer was fixed by #7827; this is a sibling in the same family that fix could not reach, since it is the runtime's buffer model rather than codegen's proven-view tiers.

<details>
<summary><b>Pre-existing gaps deliberately left alone</b></summary>

1. **Detached buffer throws `RangeError`, node throws `TypeError`.** Detaching zeroes every registered view's length, so the read is rejected as out-of-bounds before it can report the detach. The value read *before* detaching is now correct.
2. **Resizable `ArrayBuffer` is unimplemented** — `ab.resize` is not a function, so a length-tracking DataView has nothing to track.
3. **`dv instanceof DataView` is `false`** where node says `true`. A separate brand-check gap, untouched here.
4. **`test_gap_dataview_2878_2877_2879.ts` fails 3 lines on macOS** (`copyWithin` for Int16/Float64). A/B'd as byte-identical before and after this change. Worth a look separately: it is absent from `test-parity/gap_snapshot.json`, so the Linux baseline expects it to pass — meaning it is either macOS-specific or the snapshot is stale.
</details>

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed stale `DataView` reads after writes through typed-array views sharing the same buffer.
  - Fixed `TextDecoder.decode()` to consistently read current data from `DataView`, `ArrayBuffer`, `Uint8Array`, and `Buffer` views.
  - Preserved correct byte offsets, view boundaries, and endianness during shared-buffer operations.

- **Tests**
  - Added comprehensive regression coverage for bidirectional updates across aliased buffer views.


proggeramlug (approved):
Approved at exact head `30877d7eab84d005a612b8e54394bc1fcf2aadee`. The read paths now use the existing canonical backing/window resolver; the three focused runtime regressions passed independently (DataView after multi-byte typed-array writes, windowed reverse direction, and TextDecoder over DataView). Formatting and landing diff are clean. No CI result used.

Issue #7803, opened 2026-08-10
gc: zod dep-corpus under PERRY_GC_SCHEDULE_RATE=1 with the quarantine OFF dies with a rooting-shaped TypeError (seed 1, safepoint 3319)


proggeramlug:
Found by #7793's witness during its merge audit, in the sabotage arm that turns the from-space quarantine OFF — i.e., this is a latent defect on `main` (v0.5.1458 runtime), not a defect of that PR, and the shipped witness configuration (quarantine ON) runs the same workload clean.

## Reproduce

```
cargo build --release -p perry -p perry-runtime-static -p perry-stdlib-static
npm ci --ignore-scripts --no-audit --no-fund
target/release/perry test-files/gc-dep-corpus/main.ts -o /tmp/w
PERRY_GC_SCHEDULE_SEED=1 PERRY_GC_SCHEDULE_RATE=1 \
PERRY_GC_PROTECT_FROMSPACE=0 PERRY_GC_DIAG=1 /tmp/w
```

Observed (M1 Pro, macOS, release build):

```
TypeError: Cannot read properties of undefined (reading 'toString')
    at <anonymous>
[gc-schedule] done: seed=1 safepoints=3319 scheduled_collections=3319 polls_paced=23075 copying_minors=3319 moved_objects=385030 loop_polls=26394
```

exit 1, at safepoint 3,319 of the ~8,000 the full run performs. The workload is single-threaded, so the seeded schedule should replay exactly (`(seed, counter)` — one observation so far; the box was at load 20+ and a second confirmation run was cut off, so start by confirming determinism).

## Why this smells like a #7154-class rooting bug and not an instrument bug

- The same binary, same seed, same rate **with `PERRY_GC_PROTECT_FROMSPACE=1` (depth 800) passes**: 8,022 copying minors, 852,411 objects moved, output byte-identical to the unscheduled run — verified twice.
- The failure shape is the classic surfacing form: `TypeError` on an `undefined` that should be an object, cycles after the collection that lost it.
- The quarantine changes *recycling* — retired from-space page-sets are held (and PROT_NONE'd) instead of being recycled into Eden. Two readings, both worth chasing:
  1. the allocation/recycling dynamics under quarantine shift which safepoints land in the vulnerable window, so the bug's window simply never lines up in the protected arm (the protected arm also paced differently: 63,936 polls vs 26,394 at the failure point); or
[...]

PR #8084, opened 2026-08-14
gc: diagnostics, two rooting fixes, and the ungated corpus/lowering cell (#7803 investigation)

proggeramlug:
Investigation and fixes for #7803. **The principal root cause is found and fixed; 3 of the 4 aborting seeds flip to clean. One characterized window remains (seed 3) — it is documented, instrumented, and does not block the fixes here.**

## The root cause (session 4, `gc-handoff/ZOD-NOTES.md` §39–§40)

The compact GC map (`perry-codegen/src/gc_map.rs`) was built on the stated premise that *"Perry has no interior pointers"* and collapsed every RS4GC statepoint **(base, derived) pair** into one slot. The premise is false: the RS4GC prelude (`mem2reg,sccp`) hoists for-of element GEPs into values that live across polls, and LLVM records them as **derived pointers**. With the pairing discarded:

1. the runtime walker treated `&elements[i]` as an object start — reading array element words as a `GcHeader` (every "incoherent header" the pin-latch ever printed — INTERNED-on-map, `0x7FFF…` sizes — was exactly this);
2. on a moving cycle the cursor slot was never rewritten as `base' + delta` — the dangling cursor whose deref is `parse.ts:65`.

The shadow-stack era re-derived cursors per iteration, which is why the class was born at #7370's statepoint default, why `--debug-symbols` (different regalloc) suppresses, and why four runtime-side rooting fixes never touched it.

**Fix: gc_map v4** — records carry `(base_index, reg, offset)` derived entries (repeat-flag shared, version-gated fail-closed both sides); all three walkers (Itanium unwind, aarch64 fp-chain, Windows RtlVirtualUnwind) exclude derived slots from the visited-root set and rewrite them from their base after it moves, preserving each slot's stored form.

**Measured flip** (pinned schedule `RATE=0.1 ALLOC_KB=0`, same tree ± fix, `copying_minors>0` asserted per run):

| seed | pre-fix | v4 |
|---|---|---|
| 1 | 1/3 abort | 0/1 |
| 2 | 2/3 abort (fixed ordinal 58281) | **0/3** |
| 5 | 1/1 abort | 0/2 |
| 3 | 2/2–3/3 abort | 3/3 abort — **residual**, characterized to the slot/record/creation-cycle in §40 |

## Also fixed on the way (each with its own falsifier)

- **Remembered-set rebuild ran before the drain** (`gc/copying.rs`): drain-phase promotions — everything transitively reachable — were appended to `moved_headers` after `rebuild_evacuated_old_to_young_remembered_set` had already run. Moved below the last moving phase, where the classification is exact.
- **Spread-`new` / dynamic `super.m(...spread)` bundles** threaded a raw i64 accumulator across collecting calls and held the callee unrooted (the one arm `8842a0be4` missed). Both route through `bundle_args_rooted` + `RootedGroup`; IR-ordering tests verified to fail against the pre-fix lowering.
- **From-space scan false positive**: the whole-heap scan read an array's unused capacity (hole-reused old blocks keep the previous occupant's bytes) and manufactured a deterministic MISSING-REWRITE; now bounded at `length`, exclusion counted.
- **`gc-root-dominance` reader**: #8068's rustfmt wrap of `STATEPOINT_REWRITE_PASSES` broke the corpus scripts' single-line `sed`. Main fixed its own copy independently in #8087, which this branch now merges; the remaining nightly failure (a separate `--audit-poll-reach` regression from #8120) is split out as **#8134** so main can go green without waiting on this branch.

## Instruments landed (all default-off, parsed by value)

| knob | purpose |
|---|---|
| `PERRY_GC_NATIVE_SLOT_VERIFY` | abort at the cycle that **creates** a stale native slot — cycle kind, rewrite-walk stats, collector classification, raw target header |
| `PERRY_GC_THIS_SET_CHECK` (`1`/`abort`) | trap incoherent implicit-this values, both directions (frame-slot vs cell corruption) |
| pin-latch (always-on) | names the owning frame/reg/offset/slot, raw slot word, target neighborhood, census-backed **enclosing object** |
| `PERRY_UNCAUGHT_BACKTRACE`, `PERRY_KEEP_SYMBOLS`, `PERRY_GC_INTERP_SAFEPOINTS`, `PERRY_GC_POISON_FROMSPACE`, `PERRY_GC_TENURING_SURVIVALS` | sessions 1–3 (§11–§32) |

Plus the fourth corpus×lowering CI cell (dependency-scale, native statepoint roots), gated.

## Validation

- perry-runtime + perry-codegen lib suites green (incl. 38 stack-map decoder tests, 20 emitter tests with v4 round-trip corruption checks; the per-module round-trip check is always-on and passed over all 81 corpus modules).
- Gap suite: exit 0 on this tree (§41; status chatter spot-checked as load artifacts; a quiet-host re-run is listed as a follow-up).
- Sabotage: the spread-new tests go red with the two lowering files reverted; the seed flips are same-tree ± fix.

## Follow-ups (tracked, not blocking)

1. **The seed-3 residual.** `ZOD-NOTES.md` §40 and §42 name the slot, the record, the creation cycle and the next one-edit instrument: its target sits at a constant arena offset (`…8004C0` under every ASLR base) on a survivor page no collection's classifier snapshot contains, so no walk can maintain it. `PERRY_GC_NATIVE_SLOT_VERIFY=1` aborts on that cycle in ~2 minutes.
2. Quiet-host gap re-run (this tree exited 0; the status chatter spot-checked as load artifacts, §41).
3. Tighten the dep-native corpus budget 3 → 2 once re-measured — after the spread-new fix the residuals are two read-only-sink findings.
4. §33's 36-site `js_native_call_method` stale-`args_ptr` population, unrelated to the residual.

Merged with `main` (including #8081's replaceable stack-map store, whose `RwLock` index this branch's v4 decoder threads through); merged tree rebuilt, corpus compiles and runs clean, map tests 40 + 20 green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved garbage collection reliability during dynamic construction, spread calls, closure calls, and native method dispatch.
  * Fixed stale references and promoted-object handling during collection.
  * Added support for preserving interior pointers through garbage collection.

* **Diagnostics**
  * Added optional GC poisoning, interpreter safepoints, tenuring controls, and uncaught-exception backtraces.
  * Expanded diagnostics for pointer, relocation, and stale-reference issues.

* **Developer Experience**
  * Added an option to preserve final binary symbols for debugging.
  * Expanded regression coverage and dependency-scale GC validation.


proggeramlug (commented):
Audited at the landing-equivalent tree (`origin/pr-8084` merged with `main` @ `fa83ecab2`), not the PR head in isolation.

## The fix is right, and the reasoning is checkable

The remembered-set reordering in `gc/copying.rs` is the real thing. The diagnosis — `rebuild_evacuated_old_to_young_remembered_set` ran above `collector.drain()`, so `moved_headers` held only root-walk moves while the drain promotes most of the reachable heap — is precise, and the reason it went undetected for so long is stated exactly: the collector's own rewrite fires no mutator barrier, so the parent's page is `never_dirty` rather than stale-dirty. Nothing between the old and new call sites reads the remembered set (drain, then the second scanner round and the FFI root walk, all of which only rewrite), and the note that headers still carry `GC_FLAG_MARKED` because `clear_marks` runs later is the detail that makes the move safe rather than merely later.

`call_spread.rs` uses the correct shape: `open_rooted_group` because the release has to sit below the consuming call, `adopt` right after the lower, `reread` immediately before `ctx.block().call` with no allocation in between, `release` after. Shadowing `cb_box` with the reread means the pre-move register cannot be used by accident — that's a real safety property, not style.

`GC_MAP_VERSION = 4` is fail-closed: the parser returns `None` on mismatch and `stack_maps()` panics rather than walking a v3 map as v4. Correct direction for this class.

`visit_record_slots`'s contract — old base words captured before the visitor runs, derived slots never handed to the visitor because a derived pointer is not an object start — is well stated, and the consequence that a non-rewriting visitor makes every derived rewrite a no-op by construction is the right way to build a verify pass.

## The test is discriminating, and you proved it

`drain_promoted_parent_keeps_its_young_child_edge_remembered` is the standard this repo should hold. The spacer comment is the part that matters:

> the pre-fix ordering passes this test by accident (measured twice while writing it: without the spacer the parent lands 32 bytes after the intermediate)

That is a measurement that the vacuous version fails to fail, not an argument that it would. Same for the raw barrier-free slot write, and for `assert_ne!` being followed by `assert_string_bytes` so the check is "points at the live child", not "points somewhere else". The subject-liveness asserts (parent actually in old-gen, child actually still young) fail loudly if the tenuring threshold moves instead of quietly degrading.

## Blocking — CI

`lint` is a required context and goes red on the landing tree. All three are green on `main` @ `fa83ecab2`, so they are this branch's:

- 2000-line cap: `crates/perry-runtime/src/gc/roots/stack_maps.rs` at **2284**, `crates/perry-runtime/src/gc/copying.rs` at **2004**.
- addr-class ratchet: `crates/perry-runtime/src/object/this_binding.rs:191` `[gcheader-cast]`.

I have these being fixed as pure code motion (the decode/`read_uN` group and the `loaded_stack_map_section*` group split out of `stack_maps.rs`; one small extraction from `copying.rs`) plus an allowlist entry for the `this_binding.rs` site — that one is diagnostic-only and already guarded by `is_plausible_heap_addr` + `pointer_in_nursery`, and reading the header is the point of the check, so the allowlist is the sanctioned escape rather than a waiver. No behavioural change; I'll land it with the PR.

## Blocking — trivial

`promoted_remembered_7803.rs` still carries a `// TEMP PROBE` block with an `eprintln!` of page-dirty state. Debug scaffolding; drop it before merge. (If the page-dirty relationship is load-bearing enough to keep, it should be an assert, not a print — it is currently the one thing in that test that observes without checking.)

## Non-blocking — a declined derived rewrite is unobservable

`rewrite_derived_slot` has five silent early returns (base unmoved, three `decode_root_word` failures, and the `MAX_DERIVED_DELTA` guard), takes no stats, and `NativeStackWalkStats` has no derived field at all — no `derived_visited`, no `derived_rewritten`, no `derived_declined`.

Declining is not the safe side here. If the base moved and the derived slot is not rewritten, the slot keeps a from-space address, which is the exact symptom this PR exists to remove. The difference between "rewrote with a bogus delta" and "left it stale" is which corruption you get, not whether you get one.

Two specifics worth a counter rather than a redesign:

1. `delta = old_derived.addr().wrapping_sub(old_base.addr())` is unsigned, so any derived pointer sitting *before* its base wraps to a huge `usize` and silently fails `delta > MAX_DERIVED_DELTA`. The cap's doc comment justifies itself only over forward offsets. I could not construct a reachable case — the emitter records frame locations and the delta is a runtime quantity, and a before-base interior pointer would have to originate in RS4GC-tracked IR rather than runtime Rust — so I'm calling this unobserved, not broken.
2. With `PERRY_GC_NATIVE_SLOT_VERIFY=1` landing here as the instrument, a declined derived rewrite would surface as a verify abort with no attribution to the decline that caused it.

A `derived_declined` counter (bumped per return path, or at least one shared counter) would make both of those a first-look answer. Given the seed-3 residual is characterized as a survivor page absent from every classifier snapshot — a different mechanism — this is not me suggesting it explains the residual. It's that this PR's whole thesis is that silent staleness is what costs weeks, and the new code has one more silent path in it.

## Residual

The PR is straight about seed 3: 3 of 4 seeds flip clean, the fourth is characterized to slot/record/creation-cycle in §40/§42 with a named next instrument, and the sabotage check (spread-new tests go red with the two lowering files reverted) means the passing seeds are attributable. That is the right way to ship a partial fix, and it does not block the parts that are proven.

Merging once the three lint items are in and CI is green.

PR #8088, opened 2026-08-14
fix: unblock compilation of bundled agent harnesses

proggeramlug:
## Summary

This is a checkpoint PR for the Perry work needed to compile the Claude Code, OpenCode, and Pi agent bundles. It combines the compile-scaling changes and the runtime/lowering fixes discovered while reducing those bundles.

- statically link LLVM 22 into the Windows compiler and keep code generation in-process (object bytes are emitted from the LLVM API; no `.ll` handoff)
- force pathological single generated functions to `-O0` even when many small siblings hide them in a safe-looking unit average
- cap root-reload analysis by its real instruction × value cost
- fix renamed value/function exports whose public names collide with unrelated locals
- keep local classes ahead of broad imported-class metadata
- preserve captures in static accessors on factory-created class expressions
- implement constructor `Object.setPrototypeOf` / `getPrototypeOf` behavior needed by Effect Schema and inherit computed string/Symbol statics through class chains
- root compiler stack-buffer values across allocation in `js_array_from_values`
- honor `PERRY_LINK_MAP` on Windows and link an installed passive wasm-host archive when the selected runtime references it

## Harness findings so far

- The reduced 11-module OpenCode slice cold-compiles in 148.459 s and relinks from cache in 24.053 s; the linked binary is 108,076,032 bytes.
- With moving loop-poll GC disabled, that slice completes its schema initialization in 41.42 s. The renamed-export, class/prototype, computed-static, and first array-root failures are fixed.
- Default moving GC now reaches substantially farther, but a second independent stale array-to-array edge remains during MIME map population. That ongoing issue is why this PR is a draft.
- Windows static LLVM was validated with LLVM 22.1.8 built against the same `/MD` CRT as Rust. The resulting current debug `perry.exe` has no LLVM DLL imports. The repository's Windows CI provisioning still downloads LLVM's official `/MT` archive and must be moved to an `/MD` build before this is ready to merge.

## Validation

- `cargo check -p perry-codegen -p perry`
- `cargo test -p perry-runtime outlined_array_literal_roots_closure_elements_across_destination_allocation --lib`
- `cargo test -p perry-hir --test export_var_uninitialized_lowering`
- focused codegen tests for renamed getters and the root-reload complexity cap
- end-to-end Perry compile/run tests:
  - local class shadowing imported metadata
  - renamed exports and public-name collisions
  - constructor prototype chains
  - inherited computed string/Symbol statics
  - factory-local static getter captures
- `llvm-readobj --coff-imports target/debug/perry.exe`: no LLVM imports

## Known follow-ups

- finish the second OpenCode moving-GC stale-edge fix
- make the Windows `/MD` LLVM toolchain reproducible in CI/release packaging
- complete clean full-bundle compile/run/performance measurements for Claude Code, OpenCode, and Pi


proggeramlug:
Superseded by #9775, which contains the recovered compiler fixes on current main (`12efed122`) with focused regressions and a linked two-file Bun require smoke test. This older draft is conflicting and contains the previous broader patch. The replacement remains a draft while the full extracted application finishes linking and startup validation.

Issue #544, opened 2026-05-07
Per-module object cache (.perry-cache/objects/) returns stale .o when only HIR-pass code changes


proggeramlug:
## Summary

The per-module object cache at `.perry-cache/objects/` is keyed by hash-of-source only. When HIR-level transform passes (e.g. `crates/perry-transform/src/state_desugar.rs`) change but the user's `.ts` file doesn't, the cache returns a `.o` produced by the *old* pass — even though the new pass would emit different LLVM IR for the same source.

This silently masks correctness regressions during local development of new HIR passes.

## How to reproduce

1. Add a new HIR transform pass that rewrites some shape into runtime FFI calls.
2. Build & run a `.ts` file that exercises the pass — it works (cache miss, fresh codegen).
3. Edit the pass (introduce a regression, or a new branch). Rebuild perry. Don't touch the `.ts`.
4. Re-run on the same `.ts` — the cached `.o` is reused, the pass change is invisible at runtime.

## What I hit

While shipping #535 Layer 1 (commit `be071793`), I spent ~45 minutes debugging a runtime TypeError on `count.set(count.get() + 1)` that was fully resolved at the HIR level (verified via `--print-hir`). The runtime kept calling `js_state_init` correctly because the cached `.o` from an earlier successful build had the right calls baked in — but the new debug code paths I added to the codegen never executed. Confirmed by `strings ./perry | grep CODEGEN-DBG` showing the eprintln strings present in the perry binary, but no output at runtime. `rm -rf /tmp/.perry-cache && rebuild` immediately exposed the actual bug (a mismatched pre-Pass-update cached object).

## Suggested fix

Two options, in order of effort:

- **Quick:** include perry's own version (`workspace.package.version`) or a build-id of `libperry_codegen.a` in the cache key. Any `cargo build -p perry-codegen` invalidates everything, which is what we want during pass development.
- **Better:** include hashes of the perry-codegen / perry-transform / perry-hir crate `.rlib`s in the cache key. Less aggressive — only HIR/codegen/transform changes invalidate.

The status-quo (hash-of-source-only) is fine for `cargo run` against an unchanged perry binary, but pretty hostile to anyone working on a new HIR pass.

## Discovery context

Found while implementing #535 Layer 1. Cost: ~45 min of debugging a phantom bug. Filing as a separate issue so the workaround (`rm -rf <project>/.perry-cache`) is documented and the cache-key invalidation gets a real fix.

PR #8091, opened 2026-08-14
fix(build-cache): compare the compiler running now, not the recorded one

jdalton:
## Summary

The build cache can hand back an executable built by a **different compiler**, skipping the build entirely and reporting success. Rebuilding after `cargo build -p perry-codegen` can silently return the previous compiler's binary.

## The bug

```rust
if file_fingerprint_from_str(&manifest.perry_build_id.path).ok()
    != Some(manifest.perry_build_id.clone())
{
    return miss("perry-build-id");
}
```

This re-fingerprints the path recorded **in the manifest** and compares it to the value recorded beside it. It asks *"is the binary I recorded still unchanged?"* — which is trivially true whenever a different `perry` runs the second build, because the recorded binary is sitting right where it was left.

So the gate passes, and the result is total:

```
$ perry probe.ts -o probe.bin --format json
{"build_cache":{"hit":true,"miss_reason":null,"reason":"manifest-match"},
 "link_cache":{"linked":false,"skipped":true}, "success":true}
```

No relink. Output file not rewritten — same sha256, same mtime. Nothing printed. Exit 0.

`perry_version`, checked a few lines above, does not cover this: during pass development the version rarely moves between rebuilds, which is the reason `perry_build_id` exists at all (#544). The gate was present and answering the wrong question.

<details>
<summary><b>How it surfaced, and why it is hard to notice</b></summary>

A `.ts` probe compiled by a pre-fix compiler kept its stale executable when recompiled by a fixed one. A real, merged fix therefore read as *not working*, and I bisected the phantom onto an unrelated commit before suspecting the cache.

The reasons it resists the obvious checks:

- **`touch` on the source does not help.** Sources are verified by sha256, not mtime, so the content is unchanged and the manifest still matches. Correct behaviour on its own; it just removes the reflex that would otherwise expose this.
- **The output is not rewritten**, so mtime and hash both look "already correct" to anyone inspecting the file.
- **Nothing is printed.** The `Wrote executable:` line never appears, so a scripted run with output redirected sees exit 0 and moves on.
- Only a different `-o` path, or clearing the cache, produces a real build.

This is the same class as the stale-`libperry_*.a` hazard CLAUDE.md warns about under "Verifying a runtime change": *"both arms of an A/B behave identically (a vacuous 'zero regressions')"*. Here both arms return the **first** arm's binary.
</details>

## The fix

Compare against the compiler running now, using the same helper the manifest is written with, so the read and the write agree on what identity means:

```rust
if current_perry_fingerprint().ok() != Some(manifest.perry_build_id.clone()) {
    return miss("perry-build-id");
}
```

## Test plan

**End to end, two compilers over one source and one output path:**

| step | build_cache |
|---|---|
| build with compiler A | `hit: false, miss_reason: "manifest-missing"` |
| rebuild with compiler B — **must miss** | `hit: false, miss_reason: "perry-build-id"` |
| rebuild with B again — **must still hit** | `hit: true, reason: "manifest-match"` |

The third row matters as much as the second: it shows the cache still works rather than having been turned into an unconditional miss.

**Unit test** `a_manifest_from_another_compiler_does_not_match_this_one` asserts its own precondition — that the recorded file *is* unchanged on disk, which is exactly what made the old check pass — so it cannot go green by the setup drifting instead of the behaviour holding.

**Verified by sabotage.** Reverting the test to the old comparison fails it:

```
test commands::compile::build_cache::tests::a_manifest_from_another_compiler_does_not_match_this_one ... FAILED
panicked at crates/perry/src/commands/compile/build_cache.rs:171
```

## Related issue

Refs #544 — this restores the guarantee that issue was closed on.

## Summary by CodeRabbit

- **Bug Fixes**
  - Build caching now correctly detects when binaries were created by a different compiler, preventing incompatible or stale cached builds from being reused.
  - Compiler identity validation is more reliable across compiler changes and build-cache checks.

- **Documentation**
  - Added changelog information explaining the cache validation behavior, regression coverage, and available workarounds.


proggeramlug:
Audited at `d48d8136f`. **The fix is correct and lands.** I pushed two commits to the branch: one that makes the test actually guard it, and one adding the file paths / validation notes CodeRabbit asked for.

## The fix

The old check re-fingerprinted the path recorded *in the manifest* and compared it to the value recorded beside it — self-consistent, so trivially true whenever a different `perry` ran the second build. The new check compares against the compiler running now, which is the same `current_perry_fingerprint()` the **write** side already records at line 561. That symmetry is what makes it right: same compiler → same fingerprint → hit; different compiler → miss.

No new cost. `file_fingerprint` full-hashes the file, but the old path did exactly one such hash too, so this is a swap, not an addition. And `file_fingerprint_from_str` is still used by `verify_files`, so nothing goes dead.

## What I changed: the test did not guard the fix

The original test pins the two comparison expressions in isolation but never calls `probe()`. I checked that rather than assuming — reverting the production call site to the exact buggy form:

```rust
if file_fingerprint_from_str(&manifest.perry_build_id.path).ok()
    != Some(manifest.perry_build_id.clone())
```

leaves it **green**:

```
test a_manifest_from_another_compiler_does_not_match_this_one ... ok
test result: ok. 3 passed; 0 failed
```

So it documents the bug without holding the fix in place; a future refactor could reintroduce it at that line and CI would not notice. Given this bug already cost a false-regression hunt, that seemed worth closing.

`a_foreign_build_id_misses_at_the_probe` (commit `c81d2746d`) writes a manifest claiming a different compiler's build id and drives the real decision path, asserting the miss is `perry-build-id` **specifically** rather than incidentally on a later check. Under the same sabotage:

```
test a_manifest_from_another_compiler_does_not_match_this_one ... ok       <- old test, still green
test a_foreign_build_id_misses_at_the_probe ... FAILED                     <- new test catches it
test result: FAILED. 3 passed; 1 failed
```

Restored and re-run with a real rebuild: `4 passed; 0 failed`.

I kept your expression test. It states the precondition explicitly — "the recorded binary is unchanged on disk, so the OLD check would have passed here — without this the test proves nothing" — which is a genuinely good way to write it, and the two together read as a pair: one says why the old comparison was vacuous, the other says the call site uses the new one.

## Validation

On a **landing-equivalent tree** (current `main` `514ff7c9a` + this PR), since the branch predates #8087, which touches the same `mod tests`:

```
cargo test -p perry --bins build_cache     4 passed; 0 failed
cargo fmt -p perry -- --check              clean
```

The merged region is coherent — #8087's `#[cfg(test)]` on `BUILD_CACHE_ENV_EXCLUSIONS` survives alongside the widened test imports.

CI will stay red for the twelve tree-wide contexts in #8092; they fail identically on unrelated PRs and none of them is yours.

PR #8093, opened 2026-08-14
fix(ci): restore the lint gate on main

### LINKED ISSUE #8092: CI: required contexts `lint` and `cargo-test` are red for every PR (and `lint` on `main` itself), so every merge bypasses them
## Summary

Two **required** status contexts, `lint` and `cargo-test`, currently fail for every pull request, including on trees that change nothing related to them. `lint` also fails on a pristine `origin/main` checkout. While that holds, neither context can distinguish a good PR from a bad one, and every merge necessarily goes through admin bypass.

This is the CLAUDE.md "★ four ways a gate can be unable to fail" pattern in its most expensive form: the jobs run, genuinely report failure, and are required — so the only way to land anything is to ignore them, which trains the team to ignore them for the PR that actually broke something.

## `lint` is red on `main` itself

Reproduced locally on a clean detached worktree at `origin/main` = `506f4ab11a5d2a1b7a498e965c79d11e2fa63043`, running the exact commands from `test.yml`:

```
$ ./scripts/check_file_size.sh
File size limit exceeded (2000 lines).
   2010  crates/perry-runtime/src/timer.rs

$ python3 scripts/gc_store_site_inventory.py
GC store-site inventory failed; add nearby GC_STORE_AUDIT markers:
  crates/perry-codegen/src/expr/property_set.rs:1475: raw generated heap/global store: blk.store(DOUBLE, &numeric_value, &field_ptr);

$ python3 scripts/raw_handle_debt.py
bare raw-handle reads: 996 (baseline 998)
per-module raw-handle rules: 2 violation(s)
  crates/perry-runtime/src/json_tape.rs: 22 bare reads exceeds its ceiling of 21
  crates/perry-runtime/src/object/spill.rs: 5 bare reads exceeds its ceiling of 3
```

All three are independent of any open PR. Note the global raw-handle count (996) is *below* its baseline (998) — it is the per-module rules that fail, which is exactly the failure mode `raw_handle_debt_files.txt` was designed to catch, so the gate is working correctly and the debt is real.

## `cargo-test` fails identically on unrelated PRs

The same four tests fail on #8086 (73-file ShapeId change), #8090 (small typed-array fix), and #8087 (CI gates):

```
inprocess::tests::rs4gc_canonicalizes_construction_time_folds_before_root_liveness
native_emit::tests::native_construction_lowers_precise_roots_before_rs4gc
native_emit::tests::split_native_construction_lowers_precise_roots_before_rs4gc
native_emit::tests::split_native_construction_propagates_shadow_backend_to_workers
```

Byte-for-byte the same set on all three. These assert that two compilation arms produce **byte-identical** objects. Decoding the assertion dumps, the arms differ only in an embedded module/source name:

| test | control arm | native arm |
|---|---|---|
| `rs4gc_canonicalizes_…` | `.file "constant_fold_text"` | `.file "constant_fold_native"` |
| `native_construction_…` | ELF `STT_FILE` = `perry_llvm_7386555fbfde02f7.ll` | ELF `STT_FILE` = `perry_native_module` |

The generated code, stack maps, and `PGCM` payloads appear identical; the delta is the name each arm records for its translation unit. `compile_ll_to_object` routes through a scratch `perry_llvm_<hash>.ll`, while the in-process native arm names its module `perry_native_module`.

This is consistent with the platform split: both **ELF** arms of `native-roots-rs4gc` fail, while **macOS/Mach-O** and **Windows/PE** pass — Mach-O and PE do not carry an `STT_FILE`-style source-name symbol, so the difference is invisible there.

I have **not** bisected which change introduced this, and I have not confirmed it on a pristine `main` build (it does not reproduce on macOS, which is the only host I measured on). What is established is that it is not caused by any of the three PRs above.

## Also red across the same unrelated PRs

Not diagnosed here, listed so the scope is not understated — every one of these is red on both #8086 and #8090:

`gc-stress`, `ext-link`, `compiler-output-regression`, `e2e-scoped`, `gc-native-roots-complete`, `gc-root-dominance-statepoints`, `windows-arm64-build`, `Warnings (host-compatible)`, `Warnings (product)`, and both ELF `native-roots-rs4gc` arms.

The last run of `test.yml` to conclude `success` on any branch was **2026-07-24**.

## Why this is worth fixing before more merges

A required-and-always-red context is strictly worse than no context: it is indistinguishable from a gate that is doing its job, it forces bypass as the normal path, and the bypass is what lands the next real regression. #6925's `PERRY_PTR_SHAPE_LOCALS=0` regression landed visibly red and survived three merges for exactly this reason.

## Suggested split

1. Land the three `lint` fixes against `main` directly — `timer.rs` is 10 lines over the cap, `property_set.rs` needs one `GC_STORE_AUDIT` marker, and `json_tape.rs`/`object/spill.rs` each need one converted pair (which also lowers their ceilings, per the file's own "lower a ceiling whenever you convert a pair" rule).
2. Decide whether the four byte-identity tests should compare *normalized* objects (masking the translation-unit name) or whether both arms should name the module identically. A byte-identity assertion whose only observed delta is a filename is testing the filename.
3. Triage the remaining red jobs separately; several may share the ELF/`STT_FILE` root cause.

Refs #7737 (gate promotion), #8085 (gate freshness).

proggeramlug:
Closes #8092 (the `lint` half).

## Why

Three `lint` steps fail on a **pristine `origin/main` checkout**. `lint` is a required status context, so while that holds it cannot distinguish a good pull request from a bad one, and every merge goes through admin bypass. That is the CLAUDE.md "four ways a gate can be unable to fail" pattern in its most expensive form: the job runs, genuinely reports failure, is required, and therefore trains everyone to ignore it — including for the PR that actually broke something.

Reproduced on a clean detached worktree at `506f4ab11`:

```
check_file_size.sh          2010  crates/perry-runtime/src/timer.rs        (cap 2000)
gc_store_site_inventory.py  property_set.rs:1475 raw generated heap/global store
raw_handle_debt.py          json_tape.rs 22 > ceiling 21
                            object/spill.rs 5 > ceiling 3
```

None of these belongs to any open PR.

## What changed

**`timer.rs` 2010 → 1866.** Its `#[cfg(test)]` scanner seeding/snapshot helpers move to `timer/test_scanner_support.rs`. Re-exported **by name**, not by glob — the call sites are `crate::timer::test_…` in the GC root-scanner tests, and a glob does not propagate through that.

**`property_set.rs` marker relocated.** This store already had a `GC_STORE_AUDIT(POINTER_FREE)` marker; a multi-line `canonicalize_raw_f64_numeric_store_value` call had pushed it 8 lines above the store, outside the inventory's ±6-line window. The marker moves against the store it describes. **Nothing about what is stored or barriered changes** — this is the annotation catching up to a refactor, not a new safety claim.

**Three raw-handle reads converted, no ceiling raised.** `raw_handle_debt_files.txt` is explicit that adding or raising a line is not the fix ("*the honest move is to convert a pair elsewhere and say so in the PR, not to add a line here*"), and `--no-raise-vs` enforces it. So:

- `spill.rs`: across `object_meta_ensure`, and across `js_array_alloc_with_length_exact` (5 → 3, its ceiling).
- `json_tape.rs`: across the `LazyArrayRooted` safepoint (22 → 21, its ceiling).

All three already rooted the pointer and reloaded it afterwards, so this is the same discipline in the form the ratchet can count. `across_mut` additionally makes the pre-call address unnameable, which is the property the manual form cannot enforce.

Baseline 998 → 993. `--update` also tightened three ceilings carrying pre-existing slack — `object_ops/define_property.rs` 3→2, `reflect_support.rs` 4→3, `string/split.rs` 10→7. I checked: no open PR touches those files, so no collateral red. Say the word if you would rather I leave that slack alone.

## Validation

```
./scripts/check_file_size.sh                              OK: no Rust source files exceed 2000 lines
python3 scripts/gc_store_site_inventory.py --self-test    ok
python3 scripts/gc_store_site_inventory.py                passed (1430 files, 286 audited, 90 allowlisted)
python3 scripts/raw_handle_debt.py --self-test            ok
python3 scripts/raw_handle_debt.py                        993 (baseline 993); 110 modules within ceilings
python3 scripts/raw_handle_debt.py --no-raise-vs main     998 -> 993, none raised
cargo check --tests -p perry-runtime                      exit 0
cargo test -p perry-runtime --lib                         2331 passed, 0 failed, 4 ignored
```

**The `--tests` check is sabotage-verified**, because the entire `timer.rs` change is `#[cfg(test)]` and a plain `cargo check` would not compile a line of it. Planting a type error in the moved file gives exit 101 with the error attributed to `timer/test_scanner_support.rs:153`, and `Checking perry-runtime` appears exactly once in the log — so the green run compiled the moved module rather than skipping it. Restored and re-verified with a real recompile, not a `git checkout`.

## Expect the rest of CI to stay red

Twelve other contexts — `cargo-test`, `gc-stress`, `ext-link`, `compiler-output-regression`, `e2e-scoped`, both ELF `native-roots-rs4gc` arms, `windows-arm64-build` and others — fail identically on unrelated PRs and are **not** addressed here. They are catalogued in #8092. `cargo-test` in particular fails on the same four `native_emit`/`inprocess` byte-identity tests on every PR I checked, where the only observable delta between the two arms is the embedded translation-unit name.

**The signal to look for on this PR is `lint` going green.** No version bump, no manifest or lockfile change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runtime memory safety during lazy-array and object allocation operations.
  * Reduced raw-handle usage and tightened internal safety checks.
  * Restored the lint gate on the main development branch.

* **Tests**
  * Reorganized timer scanner test utilities to improve coverage and maintainability without changing runtime behavior.


proggeramlug:
Closing as redundant — **#8087 already fixes all three of these**, and does the raw-handle half slightly better. I should have checked sibling PRs before writing this; #8087's `lint` being the only green one across every open PR was the tell, and I noted it as "odd" without following it up.

Verified on #8087's head:

| violation | #8093 (this PR) | #8087 |
|---|---|---|
| `timer.rs` over the 2000-line cap | extracts `test_scanner_support` → 1866 | extracts `drain_expired_tests` → 1975 |
| `property_set.rs` `GC_STORE_AUDIT` out of the ±6 window | moves the two marker lines below the canonicalize call | **byte-identical change** |
| raw-handle ceilings | 3 conversions, baseline 998 → 993 | 4 conversions, baseline 998 → **992** |

Both arrive at the same `--no-raise-vs` result (`none raised`), and both pick up the same three slack tightenings from `--update` (`define_property.rs` 3→2, `reflect_support.rs` 4→3, `string/split.rs` 10→7). #8087 additionally converts one more read in `object/spill.rs` (3 → 2).

Landing this first would only have forced a rebase of a 47-file PR through `timer.rs`, `property_set.rs` and both ratchet files.

One thing from here worth carrying over to #8087: **the `timer.rs` change is entirely `#[cfg(test)]`, so a plain `cargo check` compiles none of it.** Validation needs `cargo check --tests -p perry-runtime`. I sabotage-verified that on this branch — a planted type error in the moved module gives exit 101 attributed to the right file, with `Checking perry-runtime` appearing exactly once, so the green run genuinely compiled the moved code. #8087 moves a different block for the same reason and deserves the same check.

Tracked in #8092; #8087 now carries the fix.

Issue #8079, opened 2026-08-14
perf(codegen): recover guarded ordinary-parameter specialization lost by #8033


proggeramlug:
## Summary

PR #8033 correctly stopped treating erased TypeScript binding annotations as runtime representation proofs, but it also removed ordinary function-parameter evidence from generic function bodies. That causes a large, reproducible specialization cliff across the current performance corpus.

Do **not** revert #8033 or seed the generic body from declared parameter types. The correctness constraint from #7846 is real. We need a guarded parameter-specialized clone (or equivalent entry guard) with the unchanged generic body as fallback.

## Current-main regression

Measured at exact main `601a02d235af2d9af1841089426a1da379edac0a`.

The clean M1 mini sweep regressed from the previously recorded `843ef621f` sweep as follows:

- Perry/Node wins: 10/19 -> 2/19
- Perry/scriptc wins: 14/19 -> 8/18
- Perry geomean across all 19 rows: 3.480x slower
- clean and separately dirty corroboration agreed within 0.2%

Largest wall-time changes on the clean mini:

| benchmark | `843ef621f` | current main | change |
|---|---:|---:|---:|
| pipeline | 0.1630 s | 3.3585 s | +1960% |
| asyncpipe | 0.0980 s | 1.1216 s | +1045% |
| retain_wide | 0.1840 s | 1.0826 s | +488% |
| shapes | 0.0550 s | 0.3183 s | +479% |
| interp | 0.8090 s | 3.9931 s | +394% |
| iso_miss | 0.9380 s | 4.4743 s | +377% |

The current timings also reproduce the older conservative-codegen `regbase` across 15 common rows: geomean current/regbase is `1.00003x`, with a maximum row difference of 1.24%.

## Causal A/B

I built a diagnostic-only compiler from exact current main that changed only ordinary `compile_function` admission:

- `local_types` remained unchanged;
- `proven_local_types` was seeded from `f.params` only;
- module globals and local initializers were excluded;
- runtime, stdlib, benchmark input, compile flags, and execution environment were otherwise held fixed.

This control is intentionally unsound for lying annotations and is **not** a proposed patch. It exists only to identify the lost optimization surface.

Every tested benchmark retained byte-exact output. Same-host retired instructions were:

| benchmark | current main | parameter-hint control | reduction | current/control |
|---|---:|---:|---:|---:|
| asyncpipe | 17,054,037,641 | 1,279,430,024 | 92.5% | 13.329x |
[...]

PR #8094, opened 2026-08-14
perf(codegen): guarded ordinary-parameter specialization

proggeramlug:
Refs #8079. **Draft — validated in part, not ready to merge.** See the full evidence in https://github.com/PerryTS/perry/issues/8079#issuecomment-5297441516.

## Provenance

This work was sitting **uncommitted** in a worktree with no active owner — 34 files, ~2,900 new lines. I committed it verbatim as a preservation commit and validated it; I did not write it, rebase it, or edit it. `git reset --soft HEAD~1` restores the exact working-tree state it replaced. The commit message is `wip(…)` for that reason and should be rewritten before this lands.

## What it does

#8033 correctly stopped treating erased TypeScript annotations as proofs, but it also removed ordinary parameter evidence from generic function bodies, which is the cause of the regression in #8079. This restores the optimization soundly: an annotation becomes a *candidate*, a compact immutable descriptor graph is emitted per guardable parameter, `js_param_type_guard` validates the live argument at the call boundary, and only then does control enter a proof-bearing `$spec_*` clone. Every function keeps its `$generic` sibling for lies, escaped/indirect calls, and ineligible parameters.

## Validated

- **Byte-exact vs node `v26.5.1`**: `test_gap_specabi_ordinary_param_guards` passes, harness exit 0, 1/1, 100%. It covers annotation lies in seven positions, an escaped call, an accessor-backed property, a body-mutated parameter, a recursive generic, and a 50k-allocation loop on both the clone and the fallback.
- **The clone is live, not vacuous**: `--trace llvm` shows 7 `js_param_type_guard` sites and 6 `$spec_*`/`$generic` pairs. `mutatePayload` correctly gets **no** clone — the ineligibility rule fires.
- **Instruction recovery 55–93%** across the seven regressed benchmarks, A/B against the preserved sweep binaries at this branch's exact base, all 14 outputs byte-identical and exit 0. On `pipeline`, `interp` and `iso_miss` it matches the unsound diagnostic control to within a tenth of a percent.
- `cargo check -p perry-codegen -p perry-runtime` exit 0.

## Not done — do not merge on the above alone

- Full gap-suite regression (a 28-file codegen change needs it).
- The new `ordinary_param_guard_tests.rs` and the `perry-codegen` suite have not been run.
- Wall clock on the quiet mini. Instructions measure work, not time.
- `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1` over the guarded path — a clone holding a proof-bearing parameter across 50k allocations is exactly that pairing's shape.
- `tree` recovers 55.1% against the control's 72.0%; worth understanding before this lands.
- Rebase (10 commits behind `main`; `git merge-tree` reports clean).
- Changelog fragment is keyed `8079-…` and should be renamed to the PR number.

## Summary by CodeRabbit

- **Performance**
  - Added runtime-validated specialization for compatible function arguments, enabling more efficient representations while preserving safe fallbacks.
  - Improved type narrowing and return-value optimization across guarded calls, branches, collections, and recursive types.
  - Extended optimization support to eligible non-suspending asynchronous functions.

- **Reliability**
  - Guarded execution now remains correct across mutations, aliasing, recursive values, accessors, and moving garbage collection.
  - Invalid or unverifiable values conservatively use the generic implementation.

- **Documentation**
  - Documented restored typed-parameter specialization and its safety guarantees.


proggeramlug:
## `surviveMovingGc` cannot exercise moving GC — the fixture is ~18× too small

Went to close out the forced-moving-GC criterion and the run came back green and empty. Reporting it because a passing moving-GC test that never collects is worse than no test.

The fixture says what it is for:

```ts
// Keep the guarded parameter live across enough allocations for the moving-GC
// matrix. Both the clone and a failed-guard fallback execute this body.
function surviveMovingGc(payload: Payload): string {
  const junk: any[] = [];
  for (let i = 0; i < 50000; i++) { junk.push({ i, text: "j" + i }); }
```

Run under the pairing CLAUDE.md prescribes:

```
PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_TRACE=1 PERRY_GC_DIAG=1

rc=0, stdout byte-exact vs node
copying_minors=0   loop_polls=1   poll_arm_events=0
[gc-fromspace-protect] lines: 0
"collection_kind" lines: 0
```

**Zero collections of any kind.** 50,000 two-field objects is roughly 3.6 MB against the 64 MB initial nursery threshold, so nothing ever triggers, and CLAUDE.md is explicit that "a run with zero copying minors protects nothing".

### Positive control — the instruments are fine, the fixture isn't

Same compiler, same runtime, same flags, on a probe that allocates 2M objects while retaining a chain:

```
copying_minors=4   loop_polls=5
[gc-fromspace-protect] lines: 4
[gc-fromspace-protect] mode=ProtectPages retired_set=#0 blocks=18 sets_held=1/4 bytes_protected=18874368 …
[gc-fromspace-protect] mode=ProtectPages retired_set=#1 blocks=13 sets_held=2/4 bytes_protected=32505856 …
```

Output identical with and without zeal. So the tooling is live and the zero above is the workload, not the knobs.

Two things I ruled out on the way, so nobody re-runs them:

- **Not the `for`-vs-`while` poll omission.** CLAUDE.md notes the specialized `for` lowering emits no back-edge poll, so that was my first guess. A `while` rewrite of the same loop gives an identical `copying_minors=0 loop_polls=1`, and `--trace llvm` shows two `js_gc_loop_safepoint()` call sites are emitted either way. Polls are present; there is simply nothing to collect.
- **Not a silent diagnostic.** Both `PERRY_GC_TRACE=1` and `PERRY_GC_DIAG=1` were set, and the run does print `[gc-globalthis-bootstrap]`, `[gc-incremental]` and `[gc-step-bounds]`.

### What this does and does not mean

It does **not** show the guarded clone mishandles a moving collection. It shows the criterion is **untested** — this fixture would pass unchanged if the clone kept a stale pointer across an evacuation, because no evacuation happens.

The fix is allocation volume, not structure: push the loop past the 64 MB trigger (~2M objects at this shape) and, importantly, **retain** across it so survivors actually move. Then re-run the pairing and require `copying_minors > 0` and a non-zero `[gc-fromspace-protect] retired_set=#N` before calling the criterion met — the counter, not the exit code.

Worth asserting that in the test rather than leaving it to whoever runs it next; a `copying_minors > 0` check is what stops this regressing back to a green no-op.

Everything else from my earlier validation stands: byte-exact output, 6 live `$spec_*`/`$generic` pairs, 7 guard sites, `mutatePayload` correctly uncloned, and 55–93% instruction recovery.

proggeramlug:
## Fixed and verified — `ec6dbb942`

Scaled `surviveMovingGc` to 1,200,000 allocations (~86 MB, clearing the 64 MB initial nursery threshold) and made it drop all but every 4096th object, so a small live set churns and survivors evacuate out of retired blocks instead of everything promoting together.

Before and after, same compiler, same runtime, `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_TRACE=1 PERRY_GC_DIAG=1`:

| | copying_minors | `[gc-fromspace-protect]` lines | output |
|---|---:|---:|---|
| original (50k, all retained) | 0 | 0 | byte-exact |
| scaled, all retained | 7 | 0 | byte-exact |
| **scaled + churn (landed)** | **20** | **20** | byte-exact |

```
[gc-fromspace-protect] mode=ProtectPages retired_set=#0 blocks=18 sets_held=1/4 bytes_protected=18874368 …
[gc-fromspace-protect] mode=ProtectPages retired_set=#1 blocks=15 sets_held=2/4 bytes_protected=34603008 …
```

The middle row is the interesting one: simply allocating more got collections running but still produced **no** from-space retirements, because a uniformly-retained set promotes wholesale and leaves nothing to quarantine. Churn is what arms the detector. Worth knowing for the next fixture someone writes for this.

With the from-space pages now `mprotect`ed across 20 evacuations, a guarded clone holding a stale parameter faults at the offending instruction rather than passing silently.

Harness confirms it, not just my byte-compare:

```
PASS  test_gap_specabi_ordinary_param_guards
Parity Pass: 1   Fail: 0   Compile Fail: 0   Crashed: 0
Parity Rate: 100.0%
```

**Forced-moving-GC roots is now met.** Still outstanding before this leaves draft: full gap-suite regression, the `ordinary_param_guard_tests.rs` and `perry-codegen` suites, wall clock on the quiet mini, the `tree` row recovering 55.1% against the control's 72.0%, and a rebase (now 12 commits behind `main`).

proggeramlug:
## Status: staying in draft — one required gate fails and the guard has a soundness hole

Rebased onto `main` (`514ff7c9a`) cleanly, then validated. Two blockers found; both are
reproducible and neither is a host flake.

### 1. `lint` (a required context) fails — 4 findings, all introduced here

```
$ python3 scripts/local_binding_type_audit.py ; echo $?
local-binding-type audit: crates/perry-codegen/src/lower_conditional.rs:87: direct ctx.proven_local_types read bypasses the proof API; use stable_local_type_proof or the audited local_type_hint escape hatch
local-binding-type audit: crates/perry-codegen/src/stmt/if_stmt.rs:166: direct ctx.proven_local_types read bypasses the proof API; use stable_local_type_proof or the audited local_type_hint escape hatch
local-binding-type audit: unclassified local-type read: crates/perry-codegen/src/lower_call/func_ref.rs::discriminant_path [stable_local_type_proof] at lines 417
local-binding-type audit: unclassified local-type read: crates/perry-codegen/src/lower_call/func_ref.rs::guarded_path_type [stable_local_type_proof] at lines 431
1
```

`git diff origin/main..HEAD` confirms all four lines are added by this branch. I ran every
other script in the `lint` job — they all pass, including the three that #8092 reported red
on `main` (`check_file_size.sh`, `gc_store_site_inventory.py`, `raw_handle_debt.py` are green
at `514ff7c9a`). So these four are the whole of `lint`'s failure.

Fixed in `e21671969` on this branch: the two raw reads move behind a new audited accessor
`FnCtx::snapshot_guarded_proof` whose body is the identical expression, so emitted code is
unchanged; the audit learns about the accessor so its call sites stay inventoried and
count-exact; and the four sites get allowlist entries. The audit and its `--self-test` now
both exit 0, as do `cargo check -p perry-codegen` and `cargo fmt --all -- --check`.

### 2. The parameter proof outlives the guard that established it — silent wrong answer

`js_param_type_guard` validates the argument **at entry**, but the proof is then stamped into
`local_types` for the **whole clone body**, and nothing re-establishes it after a call that can
write through an alias.

`collectors::has_any_mutation` is the demotion gate (`codegen/mod.rs:2468`), and
`expr_has_mutation` has no `LocalGet` arm — so passing the parameter to another function is not
treated as a possible mutation. The PR's own gap test states the right invariant but only
exercises the direct spelling:

```ts
function mutatePayload(payload: Payload): any {
  (payload as any).count = "changed";   // caught, via is_local_get_chain
  return payload.count + 1;
}
// "A parameter whose reachable value is mutated is deliberately ineligible:
//  an entry guard cannot prove facts that the body itself later invalidates."
```

Move that identical write one call away and it escapes. No casts are needed — `any` is
assignable to `number`, so `tsc` accepts all of this:

```ts
interface Box { v: number; }
const poison: any = "lie";
function assign(b: Box): void { b.v = poison; }
function compute(b: Box): string {
  const before = b.v + 1;   // entry guard passed: b.v really is 41
  assign(b);                // aliasing write, invisible to expr_has_mutation
  const after = b.v + 1;    // proof is stale here
  return "before=" + before + " after=" + after + " typeof=" + typeof b.v;
}
console.log(compute({ v: 41 }));
```

Three-way control, **same runtime archive**, exit 0 everywhere:

| arm | output |
|---|---|
| node v26.5.1 | `before=42 after=lie1 typeof=string` |
| base `601a02d23` | `before=42 after=lie1 typeof=string` |
| this branch, `PERRY_SPECIALIZED_ABI=0` | `before=42 after=lie1 typeof=string` |
| **this branch, default** | **`before=42 after=lie typeof=string`** |

An array-of-interface parameter reproduces identically (`tamper(rows)` writing `rows[0].n`):
branch prints `b=zz`, every other arm prints `b=zz1`.

The emitted IR shows the mechanism — inside `compute$spec_b`:

```llvm
%r259 = fadd double %r258, 1.0
```

a bare `fadd` on the NaN-boxed field, no tag check, because `is_numeric_expr` now answers
true for `b.v` from the entry proof (this PR's hunk at `type_analysis/numeric.rs:292`). Per
the #7773 finding, `fadd` on a NaN-box **propagates the payload**, so the wrong-typed value
passes through arithmetic unchanged — which is why it prints `lie` rather than `NaN` or
crashing. That makes it a silent wrong answer rather than a loud one.

This is reachable only because of this PR: `codegen/mod.rs:2534` makes an all-`Boxed` rep
tuple viable when `guards.iter().any(Option::is_some)`, so `compute` gets a clone it would
not have had before.

### The fix is already in this PR — in the other analysis

`spec_return_proof.rs:527-552` solves exactly this hazard for the return-proof pass:

```rust
// Different static object types can still alias through structural
// typing. Once one reference escapes to unknown code, retain no
// reference proof that might describe the same object graph.
locals.retain(|id, ty| !escaped.contains(id) && !is_reference_like(ctx.aliases, ty, 0));
```

The codegen body path just doesn't apply it. Three options, roughly by cost:

1. Port `invalidate_references_used_by` to the body path — drop **reference-typed** proofs after
   any not-provably-safe call. Primitive params are immune (a callee cannot change the caller's
   copy), so `build(depth: number)`-shaped wins survive. But `interp`/`iso_miss`/`pipeline` pass
   `Parser`/`Env`/`Value` into calls and then read fields constantly, so this may give back much
   of those rows.
2. Demote any param that escapes as a call argument. Simplest, most conservative, costs most.
3. Narrowest, and what this repo's own doctrine already says: stop letting `guarded_path_type`
   license the **unchecked** read. `type_analysis/refine.rs` says of this exact class of fact —
   *"This is a claim, not a proof, and its ONLY admissible consumer is a guarded element read"* —
   one that re-checks and falls back. A guarded fast path still beats generic dispatch, so this
   may keep most of the win. Needs its own measurement.

### On the `tree` row (55.1% vs the control's 72.0%)

Root-caused, and it is a deliberate scope boundary rather than anything inherent.
`param_guard.rs::build_named` refuses class types outright:

```rust
} else if self.classes.contains_key(name) && self.class_ids.contains_key(name) {
    // Class identity alone cannot prove mutable field values, while
    // compact instances do not expose the ordinary `keys_array`
    // needed for read-only field validation. ...
    return None;
```

`tree.ts` is `build(depth: number)`, `count(t: Tree)`, `main()`. `count` — the hot recursive
one doing all the field reads — has a class-typed parameter, so it is refused; and the `Tree`
constructor is refused at one remove because its params are `Tree | null` and `Type::Union`
maps every variant through `build_type`, so the `Named("Tree")` variant poisons the union.
`build(depth: number)` is the only guardable parameter in the whole program.

Worth noting for whoever picks that up: the runtime side is already written. `OP_OBJECT`
handles `class_id != 0` via `class_chain_reaches` (`param_type_guard.rs:399`), but codegen
always emits `class_id: 0`, so that branch is currently dead code.

### What I did not get to

**The gap suite did not run to completion, and I am not claiming it passes.** The release build
finished and the suite started, but after ~5 minutes its result journal still held only the
header line — zero of 559 tests — at load average 105-110 with several other agents building on
this box. I stopped it deliberately: parity is tag-gated rather than a required PR check, the PR
cannot land regardless, and it was starving both the measurement above and the other agents. The
build artifacts survive, so a later run can use `PERRY_SKIP_BUILD=1`.

`cargo test -p perry-codegen` and `cargo test -p perry-runtime --lib` were not run either — they
need a separate dev-profile build on an already-saturated box. Not attempted, not claimed.

`main` has moved two commits (#8090, #8091) since the rebase.

proggeramlug:
## Blocker: the entry guard's proof outlives the guard — silent wrong answers

Reproduced independently at `e21671969`. **This must not merge in its current form**, and it is not a performance question: the guarded clone prints a plausible wrong number.

`js_param_type_guard` validates the argument at **entry**, but the proof it establishes is stamped for the whole clone body. `expr_has_mutation` (the demotion gate, `codegen/mod.rs:2468`) has no `LocalGet` arm, so handing the parameter to another function is not treated as a mutation. An aliasing write one call away therefore invalidates the proof without demoting it.

```ts
function assign(b: Box): void { b.v = poison; }   // poison: any = "lie"

function compute(b: Box): string {
  const before = b.v + 1;   // guard passed at entry: b.v really is a number
  assign(b);                // aliasing write, invisible to expr_has_mutation
  const after = b.v + 1;    // proof still stamped -> bare fadd
  return "before=" + before + " after=" + after + " typeof=" + typeof b.v;
}
```

Three-way control, same compiler and same runtime archive, all exit 0:

| arm | `alias4` | `alias5` |
|---|---|---|
| node v26.5.1 | `after=lie1` | `b=zz1` |
| perry, `PERRY_SPECIALIZED_ABI=0` | `after=lie1` | `b=zz1` |
| **perry, default** | **`after=lie`** | **`b=zz`** |

The `PERRY_SPECIALIZED_ABI=0` row is the load-bearing one: same binary, guard path off, correct answer. That isolates the fault to the specialization path rather than anything else on the branch.

Why it is silent rather than a crash: the IR contains a bare `fadd double %field, 1.0` inside the clone, because `is_numeric_expr` now answers true from the entry proof. Per #7773, `fadd` on a NaN-box **propagates the payload unchanged**, so `"lie" + 1` yields `"lie"` instead of `"lie1"`. No `NaN`, no throw — just a wrong string.

### Why the existing gap test misses it

`test_gap_specabi_ordinary_param_guards` states the right invariant and covers annotation lies in seven positions, but every one of them writes through the parameter **in the same function body**, which `expr_has_mutation` does see. `mutatePayload` is the case closest to this, and it is correctly refused. Moving the identical write one call away escapes the check entirely. No casts are needed to trigger it — `any` is assignable to `number`, so tsc accepts the program.

### The fix appears to be in this PR already

`spec_return_proof.rs:527-552`'s `invalidate_references_used_by` handles exactly this aliasing hazard for the return-proof pass. The codegen body path does not apply it. Primitive-typed parameters are immune — you cannot alias a `number` — so a targeted fix should preserve most of the measured win.

### Also fixed on the branch, unrelated

`lint` was failing on `scripts/local_binding_type_audit.py` (four findings, all attributable to this branch). Addressed in `e21671969`. Worth noting the three items #8092 reported red on `main` are green again at `514ff7c9a`, so that audit was the whole failure.

### Correction to my earlier comment

I previously reported the instruction recovery and the moving-GC fix and called the criteria met. The recovery numbers stand — 55-93%, byte-exact, and the `PERRY_SPECIALIZED_ABI=0` control here re-confirms the clone is doing the work. But my statement that the lying-annotation cases were covered was too strong: they are covered **for the direct spelling only**. The aliased shape was not tested and does not hold.

### On the `tree` row (55.1% vs the control's 72.0%)

Root-caused, and not inherent: `param_guard.rs::build_named` refuses class types outright. In `tree.ts` the hot recursive `count(t: Tree)` is refused, and the `Tree` constructor is refused at one remove because `Tree | null` is a union containing a class, leaving `build(depth: number)` as the only guardable parameter. The runtime already implements the class-identity check (`class_chain_reaches`, `param_type_guard.rs:399`); codegen never emits a non-zero `class_id`, so that branch is currently dead code.

### Not done

The full gap suite did **not** complete and is not claimed. It was started but its journal held only the header after ~5 minutes at load 105-110 with several concurrent builds, and was stopped rather than starve the decisive measurement. `cargo test` on the two crates was not attempted.

proggeramlug:
## Aliasing hazard fixed; re-measured honestly. The win mostly survives.

Rebased onto `0d7fe21b0`.

### The fix is not the one that was proposed, and here is the measurement that says why

The suggested shape — give `expr_has_mutation` a `LocalGet` arm so that handing a guarded
parameter to a call demotes it — closes only the route where *we* pass the reference. It does
not close this one:

```ts
let stash: any = null;
function poison(): void { stash.v = "lie"; }
function victim(b: Box): string {
  const before = b.v + 1;
  poison();                 // b is NEVER passed anywhere
  return "before=" + before + " after=" + (b.v + 1);
}
const a: Box = { v: 41 }; stash = a; console.log(victim(a));
```

`b` never appears in an argument list, so escape analysis on the parameter structurally cannot
see it, yet `poison()` reaches the same object through the global the caller stashed it in
first. On the unfixed branch this printed `after=lie`; node and `PERRY_SPECIALIZED_ABI=0` both
print `after=lie1`.

So the sound question is **"did unknown code run"**, not "did the reference escape" — which is
also what `spec_return_proof.rs` already says in its own comment ("retain no reference proof
that might describe the same object graph"). The rule implemented is:

```rust
guard_blocked[i] = body_contains_call(&f.body) && is_reference_like(&aliases, &p.ty, 0)
```

Notes on the implementation:

- Kept **separate from `demoted`**, because that mask also drives raw representation selection;
  a reference parameter that cannot keep a descriptor proof can still travel in a raw slot.
- `has_any_mutation` and the new `body_contains_call` now share one statement skeleton
  (`any_top_level_expr`), so the two cannot drift apart on statement coverage.
- `Expr` has **550+ variants**, so an exhaustive match is not maintainable. The matcher lists the
  provably call-free variants and lets everything else fall to `_ => true`. A new variant
  therefore defaults to "may call", which costs a missed optimization rather than re-opening a
  wrong-code bug.
- Reused `is_reference_like` rather than re-implementing it.
- Primitives stay eligible: a callee has no route to the caller's copy of a number or string.

**Residual I am not claiming to have fixed**, documented in the code: property reads, index
reads and arithmetic coercions can run an accessor, a Proxy trap or a `valueOf`. They are on the
call-free list because treating every field read as unknown code makes the specialization
vacuous. The guarded object itself cannot carry an accessor — the runtime descriptor's
`own_data_field` admits plain data properties only — so exploiting this needs a *second* object
whose accessor reaches the guarded one.

### Post-fix instruction recovery — the real numbers

Base arm is the prebuilt sweep binaries from `601a02d23`; new arm compiled by the post-fix build
with `PERRY_RUNTIME_DIR` pinned and `PERRY_NO_AUTO_OPTIMIZE=1`; output basename held constant;
every stdout byte-compared against `expected/` and every exit code checked (all `ok`).

| bench | post-fix | previously published | delta |
|---|---|---|---|
| pipeline | **93.0%** | 92.9% | +0.1 |
| asyncpipe | **92.4%** | 92.0% | +0.4 |
| retain_wide | **85.6%** | 86.6% | −1.0 |
| shapes | **75.2%** | 74.9% | +0.3 |
| interp | **75.6%** | 80.3% | **−4.7** |
| iso_miss | **74.8%** | 78.8% | **−4.0** |
| tree | **55.1%** | 55.1% | 0.0 |

The soundness fix costs 4.7pp on `interp` and 4.0pp on `iso_miss` and is within noise everywhere
else. That is the predicted shape: those two are the programs whose hot parameters are
object-typed (`Parser`, `Env`, `Value`) and passed into call-bearing bodies, so they lose their
proofs; the rest run on primitive parameters. `tree` is unchanged to the digit — its only
guardable parameter was always `build(depth: number)`.

The headline survives because the win was mostly carried by primitive parameters, not by the
unsound object proofs.

One honest caveat on `tree`: its cycle-ratio recovery is 46.0% against 55.1% on instructions, so
it is more stall-bound than the instruction count alone suggests. The other six rows track.

### Regression coverage, and a gate that had quietly stopped testing anything

Three rows added to `test_gap_specabi_ordinary_param_guards.ts` — `alias-argument`,
`alias-element`, `alias-global` — covering the passed-reference, array-element and
never-passed-at-all shapes. The fixture is byte-exact against node v26.5.1 (23 lines, exit 0).

Three unit tests added for the predicate, including one that pins the *discriminating* fact:

```rust
assert!(body_contains_call(&body));
// and the escape-shaped predicate is exactly what does NOT see it
assert!(!has_any_mutation(&body, 7));
```

I also recorded a falsifiable prediction before testing it, and it held: `--trace llvm` shows
`surviveMovingGc$spec_b` **loses** its clone under the new rule (interface parameter, calls
`push`) while a primitive-parameter sibling keeps it. That silently turned the fixture's
moving-GC arm into a test of the generic path — a gate whose subject had stopped running. Added
`survivePrimitiveGuardedGc(tag: string, rounds: number)` to restore it, and asserted the subject
is live rather than merely that nothing threw:

```
PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_DIAG=1
  -> copying_minors=40, 40 x "[gc-fromspace-protect] mode=ProtectPages", stdout byte-exact
```

Worth noting the rule is not a blanket kill — `arrayTotal$spec_b` and `render$spec_b` still get
guarded clones, because their bodies are call-free.

The class-type refusal behind the `tree` row is filed separately as #8099, including that the
runtime's `class_chain_reaches` branch is currently dead code.

### Sabotage verification

Green → red → green, with a real rebuild each way (`grep -c "Compiling perry-codegen v"` == 1 per
arm — note the bare pattern `Compiling perry-codegen` reads 7, because it also matches
`perry-codegen-wasm`/`-arkts`/`-wear-tiles`, which would have made this check meaningless):

| arm | binary re-linked | fixture |
|---|---|---|
| fixed | 00:08 | GREEN, byte-exact, 23 lines |
| sabotaged (`let body_calls = false;`) | 00:20 | **RED on exactly the three new rows**, 20 other lines identical |
| restored | 00:27 | GREEN again, byte-exact |

```
< alias-argument before=42 after=lie1 typeof=string     (node)
< alias-element  a=11 b=lie1 typeof=string
< alias-global   before=42 after=lie1 typeof=string
---
> alias-argument before=42 after=lie  typeof=string     (sabotaged perry)
> alias-element  a=11 b=lie  typeof=string
> alias-global   before=42 after=lie  typeof=string
```

The sabotage is the single line the new rows claim to guard, and they fail when and only when it
is disabled. A separate monitor process reproduced the same green → red → green independently.

### What I did NOT run, and why

**The gap suite did not run. I am not claiming it passes.** Disk on the shared box fell to
8.8 GiB while other agents were building — at or below the point where a parity run on this host
produces a bogus SKIP storm rather than a trustworthy result, and low enough that pushing further
risks ENOSPC for everyone. I launched it, saw the headroom, and stopped it rather than produce a
number I would have had to discard. The harness does delete each test binary after running it
(`rm -f "$perry_binary"`), so the suite itself is not the disk hog; the headroom simply is not
there right now. Artifacts survive, so it is a `PERRY_SKIP_BUILD=1` run away once disk recovers.

`cargo test -p perry-codegen` was run only for the `collectors::mutation` module (5 passed, 0
failed). The full suite and `cargo test -p perry-runtime --lib` were not run.

Staying in **draft** until those land.

### Note on the rebase

Rebased onto `b3b6f1707`, which includes #8107. `cargo fmt --all -- --check` had been red on
`main` itself — the only two files flagged, `perry-runtime/src/array/mod.rs` and
`array/typed_array_receiver_tests.rs`, were byte-identical to main's copies and last touched by
#8090. #8107 has since fixed exactly those, and fmt is green here now. The change I rebased over
is formatting-only, so the measurements and verifications above (taken one commit earlier) still
hold.

Full static gate sweep on this head, all exit 0: `cargo fmt --all -- --check`,
`check_file_size.sh`, `local_binding_type_audit.py` (81 sites, 72 groups),
`addr_class_inventory.py`, `gc_store_site_inventory.py`, `gc_runtime_root_holders.py`,
`class_id_collisions.py`, `check_gc_env_knobs.py`, `raw_handle_debt.py`.

proggeramlug (commented):
Rebased this onto current `main` (`fa83ecab2`) and pushed — the branch was 9 commits behind, and one of those is **#8095**, which fixed the `cargo-test` failure this PR was showing. Several of the 20 red checks were inherited rather than yours. CI is re-running from scratch; no failures so far.

Clean rebase, all five commits preserved, and I verified the aliasing fix survived by content (`body_contains_call` present in `codegen/mod.rs`, `collectors/mod.rs`, `collectors/mutation.rs`) rather than trusting the commit count — a rebase on this repo has silently dropped pushed commits before.

## The soundness rule is right

Replacing escape analysis with "did unknown code run" is the correct call, and the `poison()` case proves it in the only way that counts:

```ts
function victim(b: Box): string {
  const before = b.v + 1;
  poison();                 // b is NEVER passed anywhere
  return "before=" + before + " after=" + (b.v + 1);
}
```

`b` never appears in an argument list, so no escape analysis over our own argument lists can see it, yet the caller stashed the same object in a global first. An `expr_has_mutation` `LocalGet` arm would have closed only the route where *we* hand the reference over. That the fix is not the one proposed, and that there is a measurement saying why, is the right way to have resolved it.

Sharing `any_top_level_expr` between `body_contains_call` and `has_any_mutation` "so the two can never drift apart on statement coverage" is the detail that makes this maintainable.

## One thing I checked, and one follow-up it suggests

I audited the walker's statement coverage, because a call hidden in an unvisited statement would silently unblock a guard and reintroduce exactly the `poison()` miscompile.

It handles `Expr`, `Throw`, `Return`, `Let`, `If`, `While`, `DoWhile`, `For`, `Try`, `Switch`, `Labeled`. The six that fall to `_ => {}` are `Break`, `Continue`, `LabeledBreak`, `LabeledContinue`, `PreallocateBoxes`, `PreallocateTdzBoxes` — the first four are pure control flow, and both `Preallocate*` variants carry only `Vec<LocalId>`. **No unhandled variant can contain a call, so the rule is sound as written.**

The follow-up: that soundness rests on a `_ => {}`, so it is only true for today's `Stmt` enum. Add a variant carrying an expression and this silently becomes unsound — no compile error, no test failure, just a miscompile of the `poison()` shape. CLAUDE.md already names this failure mode ("`collect_local_refs_expr()` must handle all expression types — catch-all silently skips refs"), and this walker is a load-bearing instance of it.

Cheap fix: replace `_ => {}` with the six variants spelled out, so a new `Stmt` fails to compile here and whoever adds it has to decide. Given the walker is deliberately shared with `has_any_mutation`, the exhaustive match protects both. Not a merge blocker — the current behaviour is correct — but worth doing in this PR while the reasoning is fresh, since the whole point of the shared walker is that the two predicates cannot drift.

## Note for whoever lands it

Squash-merge rather than rebase-merge: the first commit is titled `wip(codegen): guarded ordinary-parameter specialization`, which shouldn't land on `main` as-is.

Still a draft. I'll mark it ready and merge once CI comes back green, unless you want the exhaustive-match change in first.

proggeramlug:
## Do not merge yet — the codegen suite shows **32 new failures**

I rebased this onto `main` and pushed earlier, and the static gates and `cargo-test` are clean. But `cargo test -p perry-codegen` is not run by per-PR CI (CLAUDE.md: integration suites under `crates/*/tests/` are nightly/tag only), and it is where this lands.

Clean A/B — same `CARGO_TARGET_DIR`, same command, back to back, `--no-fail-fast` (without it the run dies at target 10 of 27 and you see almost none of this):

| tree | suites | passed | failed |
|---|---:|---:|---:|
| `main` @ `fb87d8923` | 27 | 1443 | 10 |
| this PR @ `5cd423010` | 27 | 1421 | **41** |

**32 new failures, 1 fixed.** Diffed by name, not by count:

```
codegen::ordinary_param_guard_tests::guarded_discriminant_branch_routes_recursive_field_to_clone
codegen::ordinary_param_guard_tests::nonsuspending_async_function_needs_no_direct_call_site_for_its_guarded_clone
compiler_private_async_iter_result_annotated_i32_payload_stays_generic
compiler_private_async_iter_result_annotated_numeric_payload_stays_generic
const_literal_length_local_routes_float64array_helper_to_raw_specialized_entry
map_number_key_set_get_has_delete_use_guarded_number_key_specialization
map_number_key_string_value_rejects_unproven_value
map_number_key_string_value_set_uses_string_ref_until_slot
map_string_boolean_param_without_native_i1_proof_uses_generic_value_helper
map_string_int32_param_without_native_i32_proof_uses_f64_helper
map_unproven_number_key_keeps_generic_fallback
packed_f64_loop_store_update_versions_with_side_exit
scalar_method_boolean_predicate_guards_public_numeric_argument_expressions
set_boolean_param_without_native_i1_proof_uses_generic_helpers
set_int32_param_without_native_i32_proof_uses_generic_helpers
set_number_add_has_delete_use_guarded_number_specialization
typed_f64_function_clone_accepts_mixed_raw_signature_and_direct_call
typed_f64_function_clone_keeps_i32_locals_raw_until_f64_use
typed_f64_function_clone_rejects_any_and_unsafe_mixed_parameter_signatures
typed_f64_method_clone_keeps_i32_locals_raw_until_f64_use
typed_f64_receiver_method_clone_raw_loads_after_composed_guards
typed_i1_function_clone_rejects_any_and_mixed_parameter_signatures
typed_i1_function_clone_rejects_mixed_direct_call_inputs
typed_i1_i32_predicate_function_uses_i32_params_and_public_wrapper
typed_i1_numeric_predicate_function_uses_f64_params_and_public_wrapper
typed_i1_numeric_predicate_method_uses_f64_params_and_guarded_direct_call
typed_i32_method_clone_emits_internal_clone_and_guarded_direct_call
typed_i32_return_function_rejects_annotation_only_or_unsafe_shapes
typed_i32_return_function_uses_i32_params_return_and_public_wrapper
typed_string_function_clone_emits_internal_clone_and_guarded_wrapper
typed_string_function_clone_rejects_unsupported_string_shapes
typed_string_method_clone_emits_internal_clone_and_guarded_direct_call
```

### Why I think this is real and not environmental

The failures are concentrated in exactly the subsystem this PR rewrites — `typed_f64_*`, `typed_i32_*`, `typed_i1_*`, `typed_string_*` clone/guard emission, `map_number_key_*` / `set_*_param_*` specialization, and the specialized-ABI proof tests. That is the guarded-specialization machinery the `guard_blocked` rule feeds.

And two of the new failures are **this PR's own tests**:

```
codegen::ordinary_param_guard_tests::guarded_discriminant_branch_routes_recursive_field_to_clone
codegen::ordinary_param_guard_tests::nonsuspending_async_function_needs_no_direct_call_site_for_its_guarded_clone
```

### The likely mechanism, stated as a hypothesis

`guard_blocked[i] = body_contains_call(&f.body) && is_reference_like(&aliases, &p.ty, 0)` is sound — I audited it and the `poison()` reasoning is right. But `body_contains_call` is true for *almost every non-trivial function*, so if `guard_blocked` is reaching further than the ordinary-parameter path it was scoped to, it would suppress specialization broadly, and "the clone is no longer emitted / the direct call site is gone" is exactly the shape of these assertions.

That is a hypothesis from the failure names, not a diagnosis — I have not read the individual assertion failures. Worth checking whether the mask is leaking into the `typed_*_clone` decisions rather than only gating the new ordinary-parameter arm. The PR description already notes `guard_blocked` was deliberately kept separate from `demoted` because that mask also drives raw representation selection; this looks like the same hazard on a different consumer.

### Reproduce

```
CARGO_TARGET_DIR=$HOME/cargo-targets/cg8094 cargo test -p perry-codegen --no-fail-fast
```

Keeping this as a draft. The rebase and the soundness rule are good; this is the remaining blocker, and it is the one CI would have let through.

PR #8095, opened 2026-08-14
fix(codegen): record a stable source_filename on emitted modules

proggeramlug:
## What

`cargo-test` — a required check — has been red on `main` since #8062/#8068/#8071 landed, on four `perry-codegen` tests:

- `inprocess::tests::rs4gc_canonicalizes_construction_time_folds_before_root_liveness`
- `native_emit::tests::native_construction_lowers_precise_roots_before_rs4gc`
- `native_emit::tests::split_native_construction_lowers_precise_roots_before_rs4gc`
- `native_emit::tests::split_native_construction_propagates_shadow_backend_to_workers`

All four assert that the textual and native construction paths emit byte-identical output. **The generated code had already converged.** The objects differed only in the name LLVM records for the module.

## Root cause

Nothing set `source_filename`, so LLVM fell back to whatever path reached the assembler:

- the **textual** pipeline writes each module to a per-call temp file, so its recorded name carried a random nonce — `perry_llvm_7386555fbfde02f7.ll`
- **native** construction recorded its in-memory module id — `perry_native_module`

ELF stores that name as an `STT_FILE` symbol, so the two paths could never agree, and neither was reproducible run to run. **Mach-O records no such symbol** — which is why all four pass on a macOS host and fail only on the Linux runner. Decoding the byte vectors from the CI failure, the entire diff for the `inprocess` test is one line:

```
-	.file	"constant_fold_text"
+	.file	"constant_fold_native"
```

## Fix

Emit an explicit `source_filename` from all three module-header sites (`to_ir`, `skeleton_ir`, and the per-codegen-unit prologue), so the recorded name is the same constant on both paths and independent of the temp path. As a side effect, emitted objects no longer embed a random temp filename, so they are reproducible across runs.

The `inprocess` fold-order test named its two arms apart itself, putting the same difference in a `.file` directive; both arms now emit under one name, leaving the generated code as the only thing the assertion compares.

## Why it stayed hidden, and what stops that recurring

The three `native_emit` tests only ever ran against the **host** triple, so on a macOS developer machine they exercised Mach-O exclusively — the one object format that does not record this name.

This PR adds `native_and_text_arms_agree_on_an_elf_target`, which pins the comparison to `x86_64-unknown-linux-gnu` explicitly and asserts the bytes actually start with `\x7fELF` before comparing, so it cannot pass by silently testing the wrong format. Two `module.rs` tests pin that every header site declares the same `source_filename`.

## Validation

- The new ELF test **reproduces the CI failure locally on macOS**, and is sabotage-tested: with the `source_filename` emission reverted it fails, and the diff is exactly `perry_native_module` vs `perry_llvm_<nonce>.ll`. Restored, it passes.
- Independently confirmed the mechanism with `clang --target=x86_64-unknown-linux-gnu`: without `source_filename` two differently-named inputs produce different objects; with it they are byte-identical.
- `cargo test -p perry-codegen --lib`: **999 passed, 0 failed** (996 before, plus the 3 new tests).
- `cargo fmt --all --check` clean; touched files are well under the 2000-line cap.

The Linux verdict is this PR's own `cargo-test` run.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reproducibility of Linux ELF builds by ensuring consistent source metadata.
  * Native and textual LLVM construction paths now produce byte-identical output.
  * Split compilation units and generated modules use a stable source identifier, preventing differences caused by temporary paths.


proggeramlug:
**`cargo-test` is SUCCESS on this PR.** That is the required status context that has been red on every open PR for this entire session, and this is what fixes it.

Your mechanism matches the diagnosis I reached independently in #8092 from decoding the assertion dumps — I had it as *"the arms differ only in an embedded translation-unit name (`perry_native_module` vs `perry_llvm_<hash>.ll`), and only the ELF arms fail while Mach-O and PE pass"*, but I stopped at the symptom. You have the cause: the textual pipeline writes each module to a per-call temp file so the recorded name carries a **random nonce**, native construction records its in-memory module id instead, and ELF stores that name as an `STT_FILE` symbol while Mach-O records none. Hence invisible on every macOS host.

Worth stating explicitly because it is the more valuable half: **the two construction paths were never byte-identical and neither was reproducible across runs.** The four tests were right and the emitter was wrong.

Landing-equivalent check (merged current `main`, macOS): `cargo test -p perry-codegen --lib` → **1001 passed, 0 failed**, with all four named tests green. I am recording that as *not* independently probative — these four already passed on macOS before this change, so a green local run cannot validate an ELF-only fix. CI's `cargo-test` going SUCCESS is the evidence here, not my box.

The remaining reds on this PR are the tree-wide set from #8092/#8117 — `conformance-smoke` (8/8, ten causes catalogued), `compiler-output-regression` (fixed by #8097), `gc-stress`, `e2e-scoped`, `repsel-census`, and the two ELF `native-roots-rs4gc` arms. None is attributable to this change.

Merging. This takes the required set from two red contexts to one.

PR #8106, opened 2026-08-14
perf(codegen): fix a root-reload allowlist symbol that has never matched anything

proggeramlug:
## Summary

`crates/perry-codegen/src/root_reload.rs`'s `NON_COLLECTING` allowlist contained
**`js_gc_layout_note_slot`, a symbol that has never existed in this tree.** The
runtime exports `js_gc_note_slot_layout` (`gc/layout.rs:814`) and
`js_gc_note_slot_layout_aware` (`:833`), which is how `gc_call_effects.rs` and
all twelve tests referencing these helpers spell them.

The list is matched against LLVM callee names by exact string, so the entry
never fired. Per the file's own contract, a helper missing from it "is treated
as collecting, which inserts a reload the checker would not have demanded — a
load, not a bug". So the typo was safe-direction and completely silent: **every
emitted slot-layout note forced a root reload**, including the one call per
guarded array element store (`expr/index_set_guarded.rs:175-186`), which is the
hot path of #5094.

Found while auditing #5094 for closure.

## Also removed: six more phantoms

`js_runtime_write_barrier_slot`, `js_value_is_object`, `js_value_is_string`,
`js_typeof_tag`, `js_typed_feedback_shape_guard` and `js_typed_feedback_note`
appear **only** in this list and its twin in
`scripts/gc_root_dominance_check.py`, and nowhere else in the repository. They
cost nothing on their own. What they cost is camouflage — with seven inert names
in a 46-entry list, a real transposition was indistinguishable from an
aspirational entry. Both lists are trimmed.

## The `_aware` entry, and the containment invariant

`root_reload.rs` documents its list as a strict subset of the checker's
`NONCOLLECTING`. `js_gc_note_slot_layout` was already in the checker;
`js_gc_note_slot_layout_aware` was not, so it is added there with its
justification:

```rust
pub extern "C" fn js_gc_note_slot_layout_aware(parent, slot_index, value_bits, old_bits) {
    if !layout_pointer_bearing_bits(value_bits) && !layout_pointer_bearing_bits(old_bits) {
        return;
    }
    layout_note_slot(strip_nanbox_user_ptr(parent), slot_index as usize, value_bits);
}
```

It is the entry point behind an early return, so it does strictly **less** than
the name the set already admits — the same "differ only by doing less" argument
the checker already records for `declare` vs `init`
(`gc_root_dominance_check.py:455-470`). Containment holds in both directions of
this edit.

## Tests — sabotage-verified, not argued

Two new tests in `root_reload_tests.rs`:

- `every_non_collecting_entry_is_a_real_runtime_export` — every entry must have
  an `extern "C" fn` definition in perry-runtime/perry-stdlib. It asserts it
  actually loaded the sources (`> 1 MB`) first, so it cannot pass vacuously.
- `the_slot_layout_note_helpers_are_non_collecting` — pins both note helpers by
  name and asserts the phantom is gone. Separate from the test above because the
  bug was a *missing* entry, and "no phantoms" is satisfied by an empty set.

Restoring the parent spelling and re-running:

```
test result: FAILED. 23 passed; 2 failed
  the_slot_layout_note_helpers_are_non_collecting ... FAILED
    js_gc_note_slot_layout is emitted per guarded element/field store; leaving
    it out forces a root reload at every one of them
  every_non_collecting_entry_is_a_real_runtime_export ... FAILED
    NON_COLLECTING names with no `extern "C" fn` definition in
    perry-runtime/perry-stdlib ...: ["js_gc_layout_note_slot"]
```

With the fix: `test result: ok. 25 passed; 0 failed`.

## Validation

- `cargo test --release -p perry-codegen --lib root_reload` — 25 passed
- `scripts/gc_root_dominance_corpus.sh` + `gc_root_dominance_check.py`:
  144/144 sources compiled, 0 skipped, 172 `.ll`; **0 violations across 2,959
  functions / 172 modules / 13,214 root stores** (allowlist unchanged and still
  empty)
- `gc_root_dominance_check.py --self-test` — OK
- `cargo fmt --all -- --check`, `git diff --check`
- `scripts/check_file_size.sh` — OK
- `scripts/gc_store_site_inventory.py` (and `--self-test`) — 1430 files, 286
  audited sites, 90 allowlisted
- `scripts/raw_handle_debt.py` — 992 (baseline 992), no ceiling added
- `scripts/addr_class_inventory.py` — passed
- `scripts/gc_runtime_root_holders.py` — OK, 123 registered scanners

No version bump; the maintainer bumps at merge time.

## Not in scope, but found and worth a follow-up

Two entries in `root_reload.rs`'s list are **not** in the checker's set —
`js_gc_forget_object_layout` and `js_array_declare_all_pointer_elements` — which
is the direction the module doc calls unsafe. Both are real exports and both are
`CannotCollect` in `gc_call_effects.rs`, so the fix is presumably to add them to
the checker, but that weakens a gate and belongs in its own reviewed change
rather than riding along here.

Refs #5094.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected runtime helper allowlists used during root-reload analysis.
  - Prevented unnecessary reload behavior caused by an incorrect helper name.
  - Removed obsolete helper entries and added missing layout-note helpers.

- **Tests**
  - Added regression checks to verify allowlist entries match available runtime helpers and reject invalid names.

### DIFF
--- a/changelog.d/8106-root-reload-note-slot-symbols.md
+++ b/changelog.d/8106-root-reload-note-slot-symbols.md
@@ -1,0 +1,30 @@
+**Fixed a symbol name in codegen's root-reload allowlist that has never matched
+anything (#5094).** `root_reload.rs`'s `NON_COLLECTING` listed
+`js_gc_layout_note_slot`; the runtime exports `js_gc_note_slot_layout`
+(`gc/layout.rs:814`) and `js_gc_note_slot_layout_aware` (`:833`). No symbol by
+the old spelling exists anywhere in the tree — `gc_call_effects.rs` and all
+twelve tests that reference these helpers use the real names.
+
+The failure mode is why it survived. The file's own contract says a helper
+missing from the list "is treated as collecting, which inserts a reload the
+checker would not have demanded — a load, not a bug". So the typo was
+safe-direction and silent: every emitted slot-layout note forced a root reload
+instead of none, including the one call per guarded array element store
+(`expr/index_set_guarded.rs`), which is the hot path #5094 exists for.
+
+Both real names are now listed, and the phantom is removed from
+`scripts/gc_root_dominance_check.py` as well — that file carried it too,
+harmlessly, because it also carries the correct spelling. `_aware` is added
+there alongside: it is `js_gc_note_slot_layout` behind an early return taken
+when neither the new nor the old bits are pointer-bearing, so it does strictly
+less than the entry point the set already admits — the same "differ only by
+doing less" argument the file already accepts for `declare` vs `init`. The
+one-way containment invariant (`root_reload.rs`'s list is a subset of the
+checker's) is preserved in both directions of the edit.
+
+Two regression tests in `root_reload_tests.rs`, both failing on the parent
+commit: every `NON_COLLECTING` entry must be a name
+`gc_call_effects::classify_direct_callee` answers `CannotCollect` for — which is
+the "the two lists must agree" rule the checker's own comment states and nothing
+enforced — and the two note helpers are pinned by name, because the bug was a
+*missing* entry and a containment test alone is satisfied by an empty set.
--- a/crates/perry-codegen/src/root_reload.rs
+++ b/crates/perry-codegen/src/root_reload.rs
@@ -183,22 +183,39 @@ const NON_COLLECTING: &[&str] = &[
     "js_gc_init_typed_shape_layout",
     "js_gc_declare_typed_shape_layout",
     "js_gc_forget_object_layout",
-    "js_gc_layout_note_slot",
+    // The two real slot-layout note exports. `js_gc_layout_note_slot` used to
+    // stand here, and no such symbol has ever existed in the tree: the runtime
+    // spells them `js_gc_note_slot_layout` (`gc/layout.rs:814`) and
+    // `js_gc_note_slot_layout_aware` (`:833`), which is how
+    // `gc_call_effects.rs` and every test names them. The typo was
+    // safe-direction — an absent helper is treated as collecting, which costs
+    // a reload rather than losing one — but it meant EVERY emitted note forced
+    // a reload, including the one per guarded array element store
+    // (`expr/index_set_guarded.rs`), which is #5094's hot path.
+    "js_gc_note_slot_layout",
+    // `_aware` is `js_gc_note_slot_layout` plus an early return when neither
+    // the new nor the old bits are pointer-bearing, so it does strictly LESS
+    // than the entry point above. Same reasoning the checker already accepts
+    // for `declare` vs `init`.
+    "js_gc_note_slot_layout_aware",
     "js_write_barrier",
     "js_write_barrier_root_nanbox",
     "js_write_barrier_slot",
-    "js_runtime_write_barrier_slot",
     "js_gc_register_global_root",
-    // pure value predicates / bit twiddling
+    // pure value predicates / bit twiddling.
+    //
+    // `js_value_is_object`, `js_value_is_string` and `js_typeof_tag` used to sit
+    // here, and — like `js_gc_layout_note_slot` above them — none is a symbol
+    // this tree exports. Six such names had accumulated (the three here plus
+    // `js_runtime_write_barrier_slot`, `js_typed_feedback_shape_guard` and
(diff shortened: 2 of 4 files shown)


proggeramlug:
Independently verified the premise and the soundness. Undrafting.

## The typo is real, and it is a transposition

```
allowlisted:  js_gc_layout_note_slot     <- appears ONLY in root_reload.rs:186 and
                                            gc_root_dominance_check.py:471
emitted:      js_gc_note_slot_layout        (4 codegen call sites)
              js_gc_note_slot_layout_aware  (3 codegen call sites)
runtime:      both exist as #[no_mangle] exports in gc/layout.rs
```

Neither emitted symbol was in `root_reload.rs`'s list. Zero `#[no_mangle]` runtime exports match `layout_note` at all, so the allowlisted spelling never named anything. `note_slot_layout` → `layout_note_slot` is a word-order transposition, which is exactly the kind of thing an aspirational-looking list hides.

## The six phantoms check out

Each of `js_runtime_write_barrier_slot`, `js_value_is_object`, `js_value_is_string`, `js_typeof_tag`, `js_typed_feedback_shape_guard`, `js_typed_feedback_note` appears in exactly two files repo-wide — `root_reload.rs` and `gc_root_dominance_check.py` — and nowhere else. Confirmed inert.

Your camouflage argument is the important part of this PR: seven dead names in a 46-entry list is what made a real transposition indistinguishable from an aspiration. Worth stating in the file's own contract so the list does not silently re-accumulate them.

## Soundness — the part I wanted to be sure about

This PR does more than delete dead strings: it makes a previously-inert judgment **live**. Because of the typo, codegen inserted a reload at every slot-layout note regardless of what the checker thought. Removing it means the checker's non-collecting classification actually load-bears at those sites for the first time. So I checked three things rather than take the containment invariant on faith:

1. **`js_gc_note_slot_layout` was already in the checker's `NONCOLLECTING`** on `main` before this PR (`gc_root_dominance_check.py:508`). This restores the documented `root_reload ⊆ checker` containment rather than asserting something new.
2. **The `_aware` variant is strictly less.** It is literally the trusted function with a short-circuit in front — an early return on `!layout_pointer_bearing_bits(value) && !layout_pointer_bearing_bits(old)`, then the same `strip_nanbox_user_ptr` + `layout_note_slot`. If the parent is non-collecting, this is a fortiori.
3. **`layout_note_slot` is allocation-free** on the paths it takes: header reads, flag tests, a forwarding-pointer tail call, and `note_element_store`, none of which contain an allocating construct. That is what the non-collecting claim actually rests on, and it holds.

## What I did not do

I did not re-run the dominance corpus myself — I am relying on your reported `0 violations across 2,959 functions / 13,214 root stores`, and on your sabotage result (parent spelling restored → 2 failures naming `js_gc_layout_note_slot` exactly; fixed → 25 pass). Those are the right two pieces of evidence for this change; I verified the premise and the soundness argument around them rather than duplicating the run.

Leaving the merge to a maintainer rather than admin-merging on my own review, since the failure mode here is a missing root rather than a wrong answer. Note `main` has moved to `12f758a22` (#8086 landed, ShapeId now authoritative) and is fmt-clean again after #8107 — worth a rebase before it goes in.

proggeramlug:
`gc-root-dominance` came back **SUCCESS**, which is the gate I was waiting for — it is the designated detector for exactly this PR's failure mode (a root reload removed where one was needed), and unlike `gc-root-dominance-statepoints` it is not in the tree-wide red set, so its verdict is attributable to this PR.

Every other GC gate is green too: `gc-moving-witnesses`, `gc-ptr-shape-off-witness`, `gc-parse-churn-gate`, `gc-ratchet`. The only failures are the two ELF `native-roots-rs4gc` arms, which fail identically on every open PR (#8092) while the macOS/Mach-O and Windows/PE arms pass.

Combined with the source-level verification in my earlier comment — the transposition confirmed, the six phantoms confirmed present in exactly two files, the checker's pre-existing classification, `_aware` being the trusted function with a short-circuit in front, and `layout_note_slot` being allocation-free — that is enough. Merging.

Worth restating what this actually buys, since it is easy to read as a lint tidy-up: the allowlist entry never matched anything, so **every emitted slot-layout note forced a root reload**, including the one call per guarded array element store — the hot path of the issue this was found under.

Issue #8006, opened 2026-08-13
test_gap_specabi_reassign: reassigned values read back as 0 (untracked, not in gap_snapshot.json)


proggeramlug:
`test_gap_specabi_reassign` diverges from node on a macOS arm64 build of `55fd197d5`
(+ PR #7999, which cannot affect program output — it only adds a link-time archive check).

```
$ diff node.out perry.out
3c3
< plain: 99 101 2
---
> plain: 0 0 2
5c5
< captured: 77:2
---
> captured: 0:2
```

Reassigned values read back as `0`. The shape — a value that was written and then reads as
zero, on the specialized-ABI path — is a codegen/spec-ABI defect, not a runtime or link one.
Compile and run both exit 0; it is an output mismatch, so the harness classifies it
`parity_fail`.

## Why it is being filed rather than triaged onto an existing entry

It is **not** in `test-parity/gap_snapshot.json`, so nothing is tracking it and nothing can
go red on it. It was found during a full local gap run for #7629
(`PERRY_SKIP_BUILD=1 ./run_parity_tests.sh --filter test_gap_`, 538 pass / 15 fail /
1 compile-fail / 0 crashed): 14 of the snapshot's 15 entries reproduced, and this was one of
two failures outside it.

Whether it is macOS-only or the Linux snapshot is stale is **unresolved** — there is no
`test-parity/gap_snapshot.macos.json` in the tree, so no macOS gap baseline has ever been
recorded, and the required gap shards run only on `ubuntu-latest`. The first thing to
establish is which of those two it is; that also decides whether a per-host snapshot is
worth committing.

proggeramlug:
Superseded by #8100, which is the same defect with the diagnosis attached.

Same test, same output, same divergence:

```
perry:  plain: 0 0 2
node:   plain: 99 101 2
```

What #8100 adds over this report:

- A **five-line reducer** (`let P: Int32Array = new Int32Array(1); … P = [99,101] as any;`).
- The **root cause**: an unguarded `call @js_typed_array_get` emitted for the plain-array reads, from `crates/perry-codegen/src/expr/index_get.rs:51-72` (#7494) using `local_type_hint`, which keeps the typed-array kind for a reassigned local. `P.length` is correct — the binding really does hold the plain array — so it is only the element reads that are wrong.
[...]

Issue #5989, opened 2026-07-04
nextjs: dynamic RSC render evaporates silently with a clean async graph after expected manifest ENOENTs (post-#5988 wall for 8/8)


proggeramlug:
## Summary

Follow-up to #5941 / PR #5988. With the #5941 async-step deadlock class eliminated (no parked machines, all promises settle, rejections propagate), the 3 request-time dynamic RSC routes (`/posts/[id]`, `/fetcher`, `/plain`) still return **0 bytes** — but the failure mode is now completely different: the render **evaporates silently with a clean async graph**. This is the remaining blocker for 8/8 on the Next.js 16 App Router standalone matrix.

## Ground truth (measured on PR #5988's branch, `/plain`)

- Stuck-machine dump (`PERRY_STUCK_DUMP=1`, toolkit on branch `fix/async-step-per-activation-identity-5941`): **zero live async-step machines** during the hang. Nothing is parked. Every machine's result settles.
- Request-window rejection log: the boot-time `/_document` + `/_app` `PageNotFoundError` rejections propagate and are caught (expected — same probes exist in node; this app has no pages router).
- Request-window sync-throw log (armed via `/tmp/perry-arm-throw` right before the curl): exactly two throws, both **expected and caught-in-node**:
  - `ENOENT ... .next/server/subresource-integrity-manifest.json`
  - `ENOENT ... .next/dynamic-css-manifest`
  Neither file exists in the build (optional manifests); node throws + catches the same loads and renders fine.
- After those two throws: nothing. No further throws, no rejections, no machines, no timers, no response write. The HTTP request never completes (curl `000` after any timeout).

## What this is NOT

- Not the #5941 deadlock (that's gone — measured).
- Not `makeHangingPromise`/abort-chain parking (no pending promises remain at all).
- Not the manifest ENOENTs being uncaught (they're caught; execution continues past them).

## Hypotheses for whoever picks this up

1. The render errors out synchronously somewhere after the manifest loads and Next's error-recovery path (`renderErrorToResponse` → pages-router `/_document` fallback → also missing) gives up without writing a response — check whether node writes a 500 there vs Perry writing nothing, and what the *primary* render error is (the known `Failed to install Date class extension` + `cacheComponents` dynamic-detection interaction is a candidate — boot prints that warning under Perry only).
2. A response/stream plumbing divergence: the App Router render may complete but `pipeToNodeResponse`/flush never runs. Instrument the res.write/end path in perry-ext-http next.

## Repro

[...]

PR #8109, opened 2026-08-14
fix(typedarray): dispatch, not just validate, when the element-read receiver is not a typed array

### LINKED ISSUE #8100: Reassigning a `new Int32Array` binding to a plain array makes every element READ return 0 (silent miscompile; sole reason `gc-stress` is red)
`test_gap_specabi_reassign` has been red on every `main` nightly since 2026-08-10. It
is a **silent wrong-answer miscompile in the shipped default configuration**, and it is
the *only* remaining reason `gc-stress` is red — which is what blocks that gate's
promotion to a required context (see the linked follow-up).

## Reproducer (5 lines)

```ts
let P: Int32Array = new Int32Array(1);
P[0] = 123;
console.log("ta:", P[0], P.length);
P = [99, 101] as any;
console.log("plain:", P[0], P[1], P.length);
```

```
perry:  ta: 123 1     plain: 0 0 2
node:   ta: 123 1     plain: 99 101 2      (v26.5.1, the .node-version pin)
```

Exit status is 0 on both sides. `P.length` is **correct** — the binding really does
hold the plain array. Only the ELEMENT READS are wrong, and they are wrong by
returning `0`.

The `Int32Array` annotation is not load-bearing: `let P: any = new Int32Array(1)` and
`let P = new Int32Array(1)` both reproduce. The `new Int32Array(...)` **initializer**
is what stamps the binding, and the reassignment never invalidates it.

Reproduced at `0d7fe21b0`, `--profile perry-dev`, `--no-auto-optimize --no-cache`,
`PERRY_RUNTIME_DIR` pinned to a freshly built `perry-dev` archive pair. CI reproduces
it at `--release` with auto-optimize on.

## It is not a representation tier

Every repsel kill switch leaves it broken:

```
PERRY_SPECIALIZED_ABI=0        -> plain: 0 0 2
PERRY_PTR_NUMARRAY_LOCALS=0    -> plain: 0 0 2
PERRY_PTR_SHAPE_LOCALS=0       -> plain: 0 0 2
PERRY_INT_VALUED_LOCALS=0      -> plain: 0 0 2
PERRY_CANONICAL_I32_LOCALS=0   -> plain: 0 0 2
```

Which matches CI exactly: `scripts/gc_repsel_matrix.sh` fails this row in **22 of 22**
arms, `shipped_default` included, while every other row is PASS or UNVER.

## Root cause

`--trace llvm` on the reduction, `min_ts.ll:3418` and `:3441`:

```llvm
%r228 = call double @js_typed_array_get(i64 %r226, i32 %r227)
```

emitted **unguarded** for the plain-array reads.

That is deliberate, and documented. `crates/perry-codegen/src/expr/index_get.rs:51-72`
(#7494) explains why `is_width_tracked_typed_array_receiver` uses `local_type_hint` —
which keeps the kind for a reassigned local — rather than `receiver_class_name`, which
drops it:

> … or a genuinely dynamic runtime call (`js_typed_array_set`, `js_typed_array_get`,
> `js_typed_array_index_{get,set}_dynamic`) **that re-validates the object's actual GC
> kind before touching memory**, exactly like `js_array_push_f64` does for a non-array
> receiver (#7574).

The helper re-validates. It does not **dispatch**.
`crates/perry-runtime/src/typedarray/access.rs:28`:

```rust
pub extern "C" fn js_typed_array_get(ta: *const TypedArrayHeader, index: i32) -> f64 {
    let ta = clean_ta_ptr(ta);
    if ta.is_null() {
        return 0.0;          // <- a plain array lands here
    }
```

`clean_ta_ptr` rejects the plain array, and the helper answers `0.0`. Memory-safe,
semantically wrong. The codegen comment's premise is half true: validation happens,
fallback does not.

## Scope: the two READ helpers, not the stores

| path | status |
|---|---|
| `js_typed_array_get` (`access.rs:28`), constant index | **broken** — returns `0` |
| `js_typed_array_index_get_dynamic` (`access.rs:192`), variable key | **broken** — returns `0` |
| element STORE after reassignment | correct |

The store is fine because codegen's store side consults `buffer_view_slots`, which
invalidates on assignment:

```ts
let P: Int32Array = new Int32Array(1);
P = [1, 2] as any;
P[0] = 99;
console.log(JSON.stringify(P));   // perry [99,2] == node [99,2]
console.log(P[0], P[1]);          // perry 0 0    != node 99 2
```

The write lands. Only the read back cannot see it.

## Suggested fix — the exact mirror of #8090

#8090 (`0d7fe21b0`, landed 2026-08-14) fixed the *other* direction of this same family:
four `Array.prototype` mutators whose TypedArray delegation sat AFTER `clean_arr_ptr`
and was therefore unreachable. Its fix was `array/header.rs:710 typed_array_receiver()`
— ask the typed-array question from the RAW, possibly NaN-boxed argument *before* the
clean rejects it.

This needs the reverse: when `clean_ta_ptr` rejects, ask whether the raw address is an
ordinary array/object and delegate to the generic `[[Get]]`
(`js_dyn_index_get`, `crates/perry-runtime/src/value/dyn_index.rs:151`) instead of
returning `0.0`. Every internal Rust caller of `js_typed_array_get`
(`node_stream_readwrite.rs`, `array/search.rs`, `typedarray_props.rs`,
`native_arena.rs`) passes an already-clean `ta`, so none of them reaches the new arm.

Alternatively the codegen side could stop trusting `local_type_hint` for a reassigned
local — but #7494's comment argues at length against that, because it is what sent a
real typed array through `is_array_expr`'s plain-array layout (element 0 at byte 8
instead of the data region at byte 16). Fixing the helper keeps that reasoning intact.

## Regression window

| nightly `test.yml` run | `main` sha | matrix |
|---|---|---|
| `31295652054` 2026-08-09 | `a853135aa` | `FAIL=0` (job red for an unrelated stale-registry liveness entry) |
| `31358183869` 2026-08-10 | `caaab6b98` | `specabi_reassign` FAIL — first appearance |
| … through `31772993217` 2026-08-14 | `601a02d23` | still FAIL, 22/22 arms |

49 commits in `a853135aa..caaab6b98`; not bisected (each hop is a full compiler +
runtime staticlib rebuild).

## Why it went unnoticed

`test_gap_specabi_reassign` is not in `test-parity/known_failures.json` and has no
issue. The only CI job that runs it on `main` is `gc-stress`, which is **not a required
context** — so it has reported this red for five days without blocking anything.
CLAUDE.md's "four ways a gate can be unable to fail", hazard 2.

proggeramlug:
Closes #8100.

`js_typed_array_get` and `js_typed_array_index_get_dynamic` are what codegen
emits for a local whose DECLARED type is a typed array but whose binding was
reassigned. `is_width_tracked_typed_array_receiver` (`expr/index_get.rs`,
#7494) keeps that hint on purpose — dropping it sends a REAL typed array on to
`is_array_expr`'s plain-array layout, a type-confused *write* — and pays for it
with an explicit promise in its own comment: the runtime helper "re-validates
the object's actual GC kind before touching memory".

It did not.

## The root cause is a type confusion, not a `0.0` return

The issue reads `clean_ta_ptr` as *rejecting* the plain array. It does not — it
rejects nothing but an address below `0x1000`:

```rust
pub fn clean_ta_ptr(ptr: *const TypedArrayHeader) -> *const TypedArrayHeader {
    let addr = strip_nanbox(ptr as u64);
    if addr < 0x1000 { return ptr::null(); }
    addr as *const TypedArrayHeader
}
```

So a plain array was read **as** a `TypedArrayHeader`:

* `TypedArrayHeader::length` and `ArrayHeader::length` are both `u32` at offset
  0, so the bounds check passed against the plain array's real length;
* `kind` (offset 8) and `elem_size` (offset 9) came from the low two bytes of
  element 0's NaN box — for `99.0` both are `0`, i.e. `KIND_INT8` with a zero
  stride;
* `data_ptr(ta)` is `ta + size_of::<TypedArrayHeader>()` = `ta+16`, which is
  element **1** of a plain array (whose slots start at `ta+8`).

The uniform `0` is a memory read, not a constant. Proof, from the probe below:
a plain **object** receiver returned `8`, and a string receiver returned `0 0`
where node prints `h i`.

## The fix

New `classify_element_read_receiver` (`typedarray/mod.rs`) — the READ-side
mirror of #8090's `array/header.rs` `typed_array_receiver`. It answers from the
raw, tag-masked argument **before** anything dereferences it:

* a registered %TypedArray% keeps the typed element path, byte-for-byte
  unchanged. A `GC_TYPE_TYPED_ARRAY` / `GC_TYPE_NATIVE_TYPED_VIEW` managed
  header also wins, so a registry miss can only cost the diversion, never the
  element read;
* anything else takes the ordinary `[[Get]]` (`js_dyn_index_get`). Codegen
  masks the NaN-box tag off (`and i64 %bits, POINTER_MASK`), so the tag is
  RECONSTRUCTED from the managed header — which matters for exactly one case: a
  heap string must be re-boxed `STRING_TAG`, or `js_dyn_index_get` walks a
  `StringHeader` as an `ObjectHeader` instead of taking its string arm.
  (Symbols share `GC_TYPE_STRING` and stay POINTER-tagged; `js_is_symbol`
  separates them.)
* a masked-away non-pointer (`P = 42 as any`) answers `undefined` — node's
  answer — instead of the old `0.0`.

Applied to both READ helpers. `js_typed_array_index_get_dynamic` generalizes
the #5989 buffer-only fallback already sitting in that arm. No recursion:
`js_dyn_index_get` re-enters the dynamic helper only when
`lookup_typed_array_kind` succeeds, which is exactly the case the classifier
keeps on the typed path.

The classifier uses no hand-rolled address floor — every probe is a side-table
lookup (safe for any bit pattern) or `try_read_gc_header`, which
magnitude-classifies (handle band included) before it dereferences. The first
draft had `if addr < 0x1000` and `scripts/addr_class_inventory.py` correctly
failed it 6-vs-5; the check was removed, not ceiling-raised.

## Why not the codegen side

Invalidating `local_type_hint` on reassignment is the alternative the issue
names. It is worse:

* #7494 documents the regression it produces: with the hint gone the access
  falls THROUGH to `is_array_expr`, which still answers true for a
  typed-array-named receiver, and lowers a REAL typed array with the
  plain-array layout (element 0 at byte 8 instead of the data region at byte
  16). That is a type-confused unboxed access — strictly worse than a
  wrong-answer read;
* `is_width_tracked_typed_array_receiver` has a second consumer in
  `expr/index_set.rs`, so the change would have to be re-argued for stores;
* the runtime fix covers every emitter of these helpers at once
  (`index_get.rs` x2, `arrays_finds.rs`, and the inline / proven-view fallback
  tiers).

## Verification

Oracle: `node --experimental-strip-types` at the `.node-version` pin, **v26.5.1**
(verified with `node --version`). Perry: `--profile perry-dev`,
`PERRY_RUNTIME_DIR` pinned to the freshly built archive pair,
`--no-auto-optimize --no-cache`. Every exit code checked; all three sides
exit 0.

The issue's reducer:

```
node        ta: 123 1     plain: 99 101 2
perry pre   ta: 123 1     plain: 0 0 2
perry post  ta: 123 1     plain: 99 101 2      <- byte-identical to node
```

An 11-section probe (constant-index reads, variable-key reads, canonical
string keys, constant and dynamic stores, and receivers that are a plain array
/ plain object / string / number / a real typed array) diverged from node on 9
lines before and is byte-identical after. A second probe covering `.at()`,
`for…of`, a loop-counter index, and reassignment inside a function is
byte-identical too.

`test-files/test_gap_specabi_reassign.ts` — the test that has been red on every
`main` nightly since 2026-08-10, and the only remaining reason `gc-stress` is
red — is now byte-identical to node.

10 unit tests in `crates/perry-runtime/src/typedarray/element_read_receiver_tests.rs`.
They assert element VALUES, and the typed-array controls store `70000` into a
`Uint16Array` and require `4464` back, so a fallback that hijacked the typed
path into boxed-f64 slots fails them.

**Sabotage-verified twice**, each with a real rebuild confirmed by
`grep -c "Compiling perry-runtime v" == 1` and a moved `libperry_runtime.a`
mtime:

* revert the two dispatch call sites, keep the classifier and the tests — 4 of
  10 unit tests go red, and the reducer and gap test both return to
  `plain: 0 0 2`;
* restore, then sabotage ONLY the `GC_TYPE_STRING` re-tag arm — exactly 1 test
  goes red, and it is the one that claims to guard it.

### Representation arms

`test_gap_specabi_reassign.ts` compiled AND run under each kill switch the
issue reports as failing, diffed against node v26.5.1, all exit codes 0:

| arm | before (#8100) | after |
|---|---|---|
| shipped default | FAIL | **PASS** |
| `PERRY_SPECIALIZED_ABI=0` | FAIL | **PASS** |
| `PERRY_PTR_NUMARRAY_LOCALS=0` | FAIL | **PASS** |
| `PERRY_PTR_SHAPE_LOCALS=0` | FAIL | **PASS** |
| `PERRY_INT_VALUED_LOCALS=0` | FAIL | **PASS** |
| `PERRY_CANONICAL_I32_LOCALS=0` | FAIL | **PASS** |

I did **not** complete a full `scripts/gc_repsel_matrix.sh --arms all` run, and
am not claiming one. `--no-build` skips the cargo build but line 437 then
invokes the compiler *without* `--no-auto-optimize`, so the warm-up triggers a
full auto-optimize rebuild of perry-runtime and its dependency tree — into the
worktree's default `target/`, not `CARGO_TARGET_DIR`. On a host with 18 GiB
free shared with three other agents that is an ENOSPC risk, so I stopped it
after ~12 min and cleaned up. The table above is the substance of that row; the
remaining arms are GC-mode arms orthogonal to a receiver-classification change,
and CI's `gc-stress` runs the real matrix on this PR.

### No regressions

`cargo test --profile perry-dev -p perry-runtime --lib` (the whole lib, not a
filter): **2361 passed, 0 failed, 4 ignored**, plus 6 serialized single-test
runs. 167 `typedarray::*` / `array::*` tests green, including the 8
`array::typed_array_receiver_tests` #8090 added yesterday.

Static gates, real exit codes at this HEAD: `scripts/check_file_size.sh` OK,
`gc_store_site_inventory.py` passed, `raw_handle_debt.py` 992 == baseline 992,
`addr_class_inventory.py` exit 0, `cargo fmt --all -- --check` exit 0.

## Two notes for the reviewer

**1. `test_gap_specabi_reassign` must stay ABSENT from `test-parity/gap_snapshot.json`.**
#8006 (closed as superseded) recorded its absence as a blind spot, and #8100
carries that forward. The inference is inverted: the snapshot's own schema says
"Lists every gap test that is NOT passing; a test absent from `tests` is
expected to pass. CI fails on any divergence in EITHER direction." Absence IS
the tracking state. Measured at `0d7fe21b0` with no edits:

```
$ python3 scripts/gap_snapshot.py check --report <report with the test failing> \
      --snapshot test-parity/gap_snapshot.json
REGRESSIONS — these were expected to pass:
  - test_gap_specabi_reassign: pass -> parity_fail
EXIT=1
```

Adding an entry would (a) disarm that and (b) fail `gap_snapshot.py check` the
moment this fix lands, as "in snapshot, now passing". So this PR adds no
snapshot entry — deliberately.

Why no CI artifact shows the failure, then: the test is index 484 of the sorted
`test_gap_*.ts` set, so `483 % 8 == 3` puts it in **shard 4**, and every recent
`conformance-smoke` run cancelled shard 4 early — the journal from run
`31838286790` stops after 56 results.

**2. An adjacent, PRE-EXISTING bug this PR deliberately does NOT fix.**
The `Uint8Array`-specialized helpers have the same disease and are not touched
here:

```ts
let Q: Uint8Array = new Uint8Array(2);
Q = [9, 10] as any;
Q[0] = 5;
console.log(JSON.stringify(Q), Q[0], Q[1]);
// node   [5,10] 5 10
// perry  [9,10] undefined undefined
```

`js_uint8array_index_get_value` / `js_uint8array_set`
(`typedarray/access.rs`) fall off the end into `undefined` / a dropped store
for a receiver that is neither a registered typed array nor a registered
buffer. Measured pre-fix and post-fix on the same probe: byte-identical, so
this PR neither causes nor fixes it. It is out of #8100's stated scope (the
issue scopes to the two READ helpers and explicitly excludes stores), and the
store half needs its own semantics review, so it is filed separately as
#8111.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed indexed reads after a typed-array reference is reassigned to a regular array.
  * Regular arrays and other ordinary values now use standard property access.
  * Invalid or out-of-bounds reads correctly return `undefined`.
  * Preserved expected behavior for valid typed arrays, strings, dynamic indexes, and different typed-array kinds.
  * Added regression coverage for receiver handling, including invalid and garbage values.

### DIFF
--- a/changelog.d/8109-typed-array-read-receiver.md
+++ b/changelog.d/8109-typed-array-read-receiver.md
@@ -1,0 +1,97 @@
+### Fixed
+
+**A reassigned `new Int32Array` binding made every element READ return `0`** (#8100)
+
+```ts
+let P: Int32Array = new Int32Array(1);
+P[0] = 123;
+console.log("ta:", P[0], P.length);
+P = [99, 101] as any;
+console.log("plain:", P[0], P[1], P.length);
+```
+
+```
+perry (before):  ta: 123 1     plain: 0 0 2
+node v26.5.1:    ta: 123 1     plain: 99 101 2
+perry (after):   ta: 123 1     plain: 99 101 2
+```
+
+Exit status was 0 on both sides and `P.length` was already correct — the
+binding really did hold the plain array. Only the element reads were wrong,
+silently, in the shipped default configuration. All five representation-
+selection kill switches (`PERRY_SPECIALIZED_ABI`, `PERRY_PTR_NUMARRAY_LOCALS`,
+`PERRY_PTR_SHAPE_LOCALS`, `PERRY_INT_VALUED_LOCALS`,
+`PERRY_CANONICAL_I32_LOCALS`) left it broken, which is why
+`scripts/gc_repsel_matrix.sh` failed the `specabi_reassign` row in 22 of 22
+arms including `shipped_default`, and why `gc-stress` was red on every `main`
+nightly since 2026-08-10.
+
+#### Root cause: the receiver was validated, but the answer was not dispatched
+
+`perry-codegen`'s `is_width_tracked_typed_array_receiver`
+(`expr/index_get.rs`, #7494) deliberately keeps a local's DECLARED typed-array
+kind even after the binding is reassigned. That is the right call — dropping
+the hint sends a REAL typed array on to `is_array_expr`'s plain-array layout
+(element 0 at byte 8 instead of the data region at byte 16), which is a
+type-confused *write*, not merely a missed optimization. #7494 pays for the
+hint with an explicit promise, written in its comment: the runtime helper
+"re-validates the object's actual GC kind before touching memory".
+
+`js_typed_array_get` did not. Its only receiver check was `clean_ta_ptr`,
+which rejects nothing but an address below `0x1000`. So a plain array was read
+**as** a `TypedArrayHeader`:
+
+* `TypedArrayHeader::length` and `ArrayHeader::length` are both `u32` at
+  offset 0, so the bounds check passed against the plain array's real length;
+* `kind` (offset 8) and `elem_size` (offset 9) came from the low two bytes of
+  element 0's NaN box — for `99.0` both are `0`, i.e. `KIND_INT8` with a zero
+  stride;
+* `data_ptr(ta)` is `ta + size_of::<TypedArrayHeader>()` = `ta+16`, which is
+  element **1** of a plain array (whose slots start at `ta+8`).
+
+The uniform `0` was a memory read, not a constant: with a plain *object*
+receiver the same expression returned `8`.
+
+#### Fix: `classify_element_read_receiver`, asked before anything dereferences
+
+New `classify_element_read_receiver` (`typedarray/mod.rs`) is the READ-side
+mirror of #8090's `array/header.rs` `typed_array_receiver`: it answers from the
+raw, tag-masked argument **before** `clean_ta_ptr`.
+
+* A registered %TypedArray% keeps the typed element path unchanged. A
+  `GC_TYPE_TYPED_ARRAY` / `GC_TYPE_NATIVE_TYPED_VIEW` managed header also wins,
+  so a registry miss can only cost the diversion, never the element read.
+* Anything else takes the ordinary `[[Get]]` (`js_dyn_index_get`). Codegen
+  masks the NaN-box tag off the receiver (`and i64 %bits, POINTER_MASK`), so
+  the tag is RECONSTRUCTED from the managed header — which matters for exactly
+  one case: a heap string must be re-boxed `STRING_TAG` or `js_dyn_index_get`
+  walks a `StringHeader` as an `ObjectHeader` instead of taking its string arm.
+  (Symbols share `GC_TYPE_STRING` and stay POINTER-tagged, which is what
+  `js_is_symbol` separates.)
+* A masked-away non-pointer (`P = 42 as any`) answers `undefined`, node's
+  answer, instead of the old `0.0`.
+
+Applied to both READ helpers: `js_typed_array_get` (constant index) and
+`js_typed_array_index_get_dynamic` (variable / string key, via
+`typedarray_props::typed_array_index_get_dynamic`, generalizing the #5989
+buffer-only fallback that already sat there). The STORE side was already
+correct — codegen's store path consults `ctx.buffer_view_slots`, which
+invalidates on assignment, so a reassigned binding never reaches
+`js_typed_array_set`.
+
+#### Verified
+
+`node --experimental-strip-types` at the `.node-version` pin (v26.5.1), every
+exit code checked. An 11-section probe covering constant-index reads,
+variable-key reads, canonical string keys, constant and dynamic stores, and
+receivers that are a plain array / plain object / string / number / a real
+typed array diverged from node on 9 lines before and is byte-identical after.
+`test-files/test_gap_specabi_reassign.ts` is byte-identical to node.
+
+10 unit tests in
+`crates/perry-runtime/src/typedarray/element_read_receiver_tests.rs`. They
+assert element VALUES, not absence of a panic; the typed-array controls store
+`70000` into a `Uint16Array` and require `4464` back, so a fallback that
+hijacked the typed path into boxed-f64 slots fails them. Sabotage-verified:
+with the two dispatch call sites reverted and the tests unchanged, 4 of the 10
+go red.
--- a/crates/perry-runtime/src/typedarray/access.rs
+++ b/crates/perry-runtime/src/typedarray/access.rs
@@ -24,12 +24,28 @@ pub extern "C" fn js_typed_array_length(ta: *const TypedArrayHeader) -> i32 {
 }
 
 /// `ta[i]` — returns plain f64 numeric value (NOT NaN-boxed).
+///
+/// #8100: codegen emits this for a local whose DECLARED type is a typed array
+/// even after the binding was reassigned (`is_width_tracked_typed_array_
+/// receiver` keeps the hint on purpose — see its #7494 comment — and pays for
+/// that with the promise that this helper re-validates the receiver's actual
+/// GC kind). `clean_ta_ptr` does not validate anything but the magnitude, so
+/// a plain array landing here was read AS a `TypedArrayHeader` and every
+/// element came back `0`. Classify first, then dispatch: a non-typed-array
+/// receiver takes the ordinary `[[Get]]`, exactly as if the static hint had
+/// never existed.
 #[no_mangle]
 pub extern "C" fn js_typed_array_get(ta: *const TypedArrayHeader, index: i32) -> f64 {
-    let ta = clean_ta_ptr(ta);
-    if ta.is_null() {
-        return 0.0;
-    }
+    let ta = match classify_element_read_receiver(ta as u64) {
+        ElementReadReceiver::TypedArray(addr) => addr as *const TypedArrayHeader,
+        ElementReadReceiver::Ordinary(receiver) => {
+            return crate::value::js_dyn_index_get(receiver, f64::from(index))
+        }
+        // Was `0.0`. `undefined` is the ECMAScript answer and the one node
+        // prints: `let P: Int32Array = new Int32Array(1); P = 42 as any; P[0]`
+        // is `undefined`, not `0`.
+        ElementReadReceiver::Absent => return f64::from_bits(crate::value::TAG_UNDEFINED),
+    };
     unsafe {
         if crate::native_arena::is_native_typed_view(ta) {
             crate::native_arena::validate_view_alive(
--- a/crates/perry-runtime/src/typedarray/element_read_receiver_tests.rs
+++ b/crates/perry-runtime/src/typedarray/element_read_receiver_tests.rs
@@ -1,0 +1,221 @@
+//! #8100: an element **READ** helper handed a receiver that is not a
+//! %TypedArray% must fall back to the ordinary `[[Get]]`, not read foreign
+//! memory as a `TypedArrayHeader`.
+//!
+//! ## Why this file exists
+//!
+//! `perry-codegen`'s `is_width_tracked_typed_array_receiver` (#7494) keeps a
+//! local's DECLARED typed-array kind even after the binding is reassigned —
+//! deliberately, because dropping the hint sends a REAL typed array through
+//! `is_array_expr`'s plain-array layout (element 0 at byte 8 instead of the
+//! data region at byte 16). It pays for that with an explicit promise that the
+//! runtime helper "re-validates the object's actual GC kind before touching
+//! memory".
+//!
+//! `js_typed_array_get` did not. Its only receiver check was `clean_ta_ptr`,
+//! which rejects nothing but an address below `0x1000`. So
+//!
+//! ```text
+//! let P: Int32Array = new Int32Array(1);
+//! P = [99, 101] as any;
+//! P[0]                                   // perry 0, node 99
+//! ```
+//!
+//! read an `ArrayHeader` as a `TypedArrayHeader`: `length` matched (offset 0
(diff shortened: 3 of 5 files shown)


proggeramlug:
Audited at `c7add2fb2`. **Merging.**

## The diagnosis is right, and worse than #8100 described

I confirmed the mechanism in source. `clean_ta_ptr` does not reject anything above `0x1000`, so a plain array was read **as** a `TypedArrayHeader` — `length` is `u32` at offset 0 in both headers so the bounds check "passed", `kind`/`elem_size` came off the low bytes of element 0's NaN box, and `data_ptr` landed 8 bytes past the real element region. The uniform `0` was a memory read, not a rejection constant. That is type confusion, and #8100 read it as a rejection.

## The classifier is sound

The design decision that matters: only a **positively identified** receiver is diverted, and a `GC_TYPE_TYPED_ARRAY`/`GC_TYPE_NATIVE_TYPED_VIEW` header still wins on a registry miss — so a lookup failure can cost the diversion but never the element read. That is the safe direction. Every probe is a side-table lookup or `try_read_gc_header`, which magnitude-classifies before dereferencing, so there is no hand-rolled address floor to go stale.

Removing your first draft's `if addr < 0x1000` when `addr_class_inventory.py` flagged it — rather than raising the ceiling — was the right call, and it is why the classifier reads as it does.

## Validation I ran independently

I enumerated the `lint` job's steps **from `test.yml`** rather than from memory, because I broke `main` earlier today doing exactly that. It turns out to be 20 scripts, not the four I had been running:

```
34 checks: cargo fmt --all --check, check_file_size.sh, and --self-test + real runs of
addr_class_inventory, check_gc_doc_claims, check_gc_env_knobs, check_llvm_corpus_currency,
check_locale_independent_io, check_node_version_consistency, check_test_registration,
gc_gate_wiring_check, gc_pin_sites, gc_runtime_root_holders, gc_store_site_inventory,
global_sink_isolation, local_binding_type_audit, raw_handle_debt,
gc_matrix_liveness_check --check-registry, workspace_architecture --check,
raw_handle_debt --no-raise-vs
                                                              -> ALL GREEN
cargo test -p perry-runtime --lib   -> 2361 passed, 0 failed, 4 ignored (10 new)
```

No version bump, no manifest or lockfile change, PR based directly on current `main`.

## On the open review thread — checked, not waived

The thread flags that `TypedArray(addr)` from a header match with a registry miss falls to `_` and returns `undefined`. I checked whether that is a regression, and it is not — the pre-PR code for that entire branch was:

```rust
return f64::from_bits(crate::value::TAG_UNDEFINED);
```

an **unconditional** `undefined` for every receiver that missed the registry and was not a registered buffer. This PR routes the `Ordinary` case to `js_dyn_index_get` and leaves `_` exactly as it was. So it strictly improves the branch and touches nothing else.

It is still a real corner worth closing — the classifier is precisely what makes it *nameable* now — and `typed_array_addr_from_value` gates on `typed_array_owner_kind` while the classifier gates on `lookup_typed_array_kind`, so the two can in principle disagree. I have filed it as a follow-up rather than hold a fix for a live, shipped, silent miscompile behind a pre-existing corner it does not worsen.

Two things from your report I want on the record because they corrected me:

- **You declined my instruction to add the test to `gap_snapshot.json`, and you were right.** The schema says absence means expected-to-pass and CI fails on divergence in either direction; adding an entry would have disarmed the gate and then failed it once the fix landed. I have corrected that on #8100.
- **Your shard arithmetic explains the invisibility** — index 484, `483 % 8 == 3` → shard 4, cancelled early. That is a much better answer than "nothing was tracking it", and it connects to `conformance-smoke` failing 8 of 8 shards (#8092).

Merging with `--admin`, as every merge currently must (#8092).

Issue #8067, opened 2026-08-14
feat(object/shape): make ShapeId authoritative for keys, live slots, and class-object kind before #8047


proggeramlug:
## Summary

A source/IR/FFI audit of `main` at `08173f8b4` found that #8047's prerequisite is **not complete**. Rungs 1–2 made `ObjectHeader.parent_class_id` a uniform, birth-stamped shape word, but the planned rung 3 never landed: `keys_array` and `field_count` are still authoritative mutable object state, several guards still compare the keys pointer directly, and `object_type` still carries the live `OBJECT_TYPE_CLASS` semantic.

This issue is the missing prerequisite. It deliberately does **not** remove fields or change `ObjectHeader` size; that atomic ABI/layout payoff stays in #8047 after this issue proves the replacement facts.

### 1. `ShapeId` cannot recover the fields #8047 proposes to delete

`crates/perry-runtime/src/object/shapes.rs` still defines:

- `ShapeTable.entries: PtrHashMap<usize, Shape>` keyed by the **keys-array address**;
- `Shape { indexed_len, shape_id, slots }`, with no authoritative rooted keys edge and no by-id lookup;
- `shape_id_for_keys_ensure(keys, key_count)`, `shape_keys_grown(old_keys, new_keys)`, `shape_drop(keys)`, and the GC prune/rekey paths all address-keyed.

`docs/shape-tree-plan.md` says this explicitly: `keys_array` remains the ordered-key artifact and the transition cache remains separate. A header containing only `{ class_id, shape_id, meta }` therefore has no way to enumerate keys or recover the live field/slot bound.

The ShapeId allocator also deliberately returns `0` on range exhaustion. Generated PICs then fall back to `keys_array` pointer tokens. Removing the pointer requires a sound exhausted-id behavior rather than silently deleting the fallback.

### 2. `field_count` is independently mutable, not `keys_array.length`

The publication/liveness invariant is load-bearing:

- `object/field_get_set/field_ops.rs::js_object_set_field` widens `field_count` **before** a pointer store so `object::gc_field_slot_range` traces and rewrites the new slot;
- the same pre-store widening exists in `object/field_set_by_name.rs`, `field_set_by_name/{fast_paths,tail}.rs`, and `object_ops/keys_array.rs`;
- `object/delete_rest.rs::js_object_delete_field` compacts and shrinks it;
[...]

Issue #8047, opened 2026-08-13
perf(object/GC): finish the common-object header shrink from 56 B to 40 B after shape-transition migration


proggeramlug:
## Summary

The measured object-representation campaign still has one concrete, high-value step with no open ticket: shrink a common two-slot object from the current **56 bytes** to **40 bytes** by removing the derivable `object_type`, `field_count`, and `keys_array` words once every inline-cache consumer is shape-transition-safe.

This is the unlanded endgame identified in #7916. That issue closed after delivering its measured proposal, while #6759/#8009 migrated important shape consumers but explicitly did not perform the header shrink. Current `main` still uses:

- 8-byte `GcHeader`;
- 32-byte `ObjectHeader`; and
- two 8-byte slots;

for **56 bytes** total on `{a,b}`.

## Measured prize

The exact mirror probe (grow the header by 16 B and read `pad16 -> base` as the shrink) measured, with byte-exact output and exit 0 on both arms:

| benchmark | instructions | peak RSS | copying minors |
|---|---:|---:|---:|
| `retain` | **-25.63%** | **-26.19%** | 6 -> 4 |
| `retain1` | noise | **-18.72%** | 3 -> 2 |
| `retain_wide` | flat | **-10.52%** | 8 -> 6 |
| `retain_wide1` | flat | **-13.42%** | 4 -> 3 |
| `tree` | flat | **-18.64%** | unchanged |
| `tree_wide` | flat | **-10.66%** | unchanged |

Full evidence and harness references are in #7916’s final measurement comment and `gc-handoff/HDR-NOTES.md`.

## Why it was blocked

The hot inline guard historically loaded, in one receiver block:

```text
object_type @0
class_id @4
field_count @12
keys_array @16
```

Three of those are the fields to delete. `keys_array` identity could not simply become a ShapeId comparison while deletion preserved class identity and cloned/compacted the keys array. The hidden-class/shape-transition migration therefore had to land first.

Several prerequisites have now landed, including birth stamps and shape-based cache migration through #8007/#8009/#8010/#8015/#8025/#8029. Re-audit **all** header consumers on current main before assuming the prerequisite is complete; a partial nine-consumer migration is not enough to change the ABI.

## Required work

1. Census every runtime/codegen/FFI consumer of `ObjectHeader` layout and all `object_type`, `field_count`, and `keys_array` reads.
[...]

PR #8110, opened 2026-08-14
ci(object): make the ObjectHeader shape-descriptor census a real gate

proggeramlug:
## What

Wires `scripts/shape_descriptor_census.py` into the `lint` job, and fixes a
vacuous arm inside it that the wiring exposed.

## Why

#8086 shipped the exact-callsite census that #8067 required — the instrument
that keeps `ObjectHeader::object_type`, `field_count` and `keys_array` retired
while `ShapeId` takes over their facts, and the thing #8047 says must exist
before anyone touches the layout. It is a good instrument: it strips comments
and string literals, classifies every site `declaration` vs `access`, diffs the
full multiset against a reviewed baseline (so it can tell an `ObjectHeader` read
from another struct's identically-named field), and carries its own lexer
self-test plus sabotage self-tests.

It is invoked by nothing:

```
$ grep -rn "shape_descriptor_census" --include=*.yml --include=*.sh \
      --include=*.py --include=*.toml --include=*.md . \
  | grep -v '^./scripts/shape_descriptor_census'
scripts/shape_descriptor_census.py:15:BASELINE_PATH = ROOT / "scripts" / ...
```

The only hit is the script naming its own baseline. `lint` runs fifteen other
python audits; this is not one of them.

## The vacuous arm

Wiring it up and sabotage-testing it found that the emitted-guard check could
not fail for any function it names. It matched

```python
re.search(r"expected_keys|add\s*\([^\n]*\"(?:0|12|16)\"", body)
```

but all four functions in that list build their header address with
`blk.gep(I8, &p, &[(I64, "N")])`, not `add(..)`:

| function | actual emission |
|---|---|
| `emit_class_field_loop_preheader_check` | `blk.gep(I8, &obj_ptr, &[(I64, "4")])` |
| `emit_proven_shape_recheck` | `blk.gep(I8, &obj_ptr, &[(I64, "8")])` |
| `emit_class_field_inline_precheck` | `blk.gep(I8, &obj_ptr, &[(I64, "4")])` |
| `emit_element_shape_field_load` | `blk.gep(I8, &elem_ptr, &[(I64, "8")])` |

So the pattern targeted a syntax none of them uses. Planting a read of the
`keys_array` offset that #8047 removes —
`gep(I8, &elem_ptr, &[(I64, "16")])` in `emit_element_shape_field_load` — left
the census **green**. This patch matches the gep form as well.

The two write-PIC checks below it are unaffected: `lower_generic_property_get`
and the `proxy_reflect` emitters genuinely do use `ctx.block().add(I64, .., "N")`,
and their own regexes already spell that.

## Verification

Sabotage-verified, every change reverted afterwards, clean tree green before and
after each:

| sabotage | before | after |
|---|---|---|
| new `(*obj).field_count` read in `object/spill.rs` | exit 1 | exit 1 |
| keys-pointer token in `proxy/put_value.rs::dyn_ic_try_store` | exit 1 | exit 1 |
| offset 16 in `emit_element_shape_field_load` | **exit 0** | exit 1 |
| offset 12 in `emit_class_field_loop_preheader_check` | **exit 0** | exit 1 |
| offset 12 in `emit_proven_shape_recheck` | **exit 0** | exit 1 |
| offset 12 in `emit_class_field_inline_precheck` | **exit 0** | exit 1 |

All sixteen `lint` steps, enumerated from `.github/workflows/test.yml` rather
than from memory, pass on this branch: `cargo fmt --all -- --check`,
`check_file_size.sh`, `workspace_architecture.py` (`--self-test` and `--check`),
`local_binding_type_audit.py`, `gc_store_site_inventory.py`,
`addr_class_inventory.py`, `class_id_collisions.py`, `gc_pin_sites.py`,
`gc_runtime_root_holders.py`, `check_gc_env_knobs.py`,
`check_llvm_corpus_currency.py`, `check_gc_doc_claims.py`, `raw_handle_debt.py`,
`check_test_registration.py`, and the newly wired census.

No behaviour change: the census reports the same summary on this branch as on
`12f758a22`.

Note this is a *new* gate, so per CLAUDE.md it has never been green — run it
once on `main` before promoting it to a required context.

## Not in this PR

The #8047 layout change itself. A full re-audit on `12f758a22` found the
prerequisite still incomplete; findings are posted on #8047.

Refs #8047, #8067, #8086.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for shape-descriptor metadata and emitted offset patterns, including additional generated forms.
  * Corrected matching for guard checks and expanded coverage across descriptor-related access patterns.

* **Tests**
  * Added a required lint-time census and self-checks to detect retired metadata usage and verify shape identifier authority.
  * Updated validation baselines to reflect current source usage and ensure consistent counts.


proggeramlug:
Verified and endorsed — and this is my miss to own.

I audited and merged #8086. My checks were: all four lint gates on a landing-equivalent tree, `perry-runtime` (2351/0), the full `perry-codegen` suite diffed test-for-test against main's 11-failure baseline, and specifically whether the raw-handle ratchet had been softened to accommodate the new `with_{mut,const}_ptr` combinator (it had not). All of that was true and none of it was the right question for the census.

Confirmed independently on `main` at `12f758a22`:

```
$ grep -rn "shape_descriptor_census" .github/workflows/
(no output)

$ git grep -l "shape_descriptor_census" origin/main
scripts/shape_descriptor_census.py
```

The script is referenced by exactly one file — itself. #8067's acceptance asked for a census; #8086 built it and wired it into nothing, so it has never gated anything and never could.

**And the arm that would have mattered was vacuous.** Your finding that it matched `add(..., "0"|"12"|"16")` while all four named functions emit `blk.gep(I8, &p, &[(I64, "N")])` means that even once wired, planting a `keys_array`-offset read in `emit_element_shape_field_load` left it green — which you demonstrated rather than argued. That is the same failure this repo has hit repeatedly, and it is worth noting it survived *because* nothing ran it: an unwired gate cannot fail, so its vacuity was undetectable by any amount of it "passing".

The generalisable lesson I am taking from this, since it is a gap in how I audit: **when a PR adds a gate, checker or census, verifying its tests pass is not enough — verify it is wired into a workflow, and verify it can go red.** "Does this new gate run?" and "can it fail?" are two separate questions and I asked neither.

Six sabotages with the clean tree re-verified green after each, and all 16 `lint` steps enumerated from the workflow rather than remembered, is the right standard. No objection from me — this should land.

Two things from your report worth carrying into #8112/#8113 rather than losing here: `perry-ffi`'s `object_header_matches_runtime` has never executed (`runtime-link` enabled nowhere in CI, and `cargo-test` is per-package), and `perry-ffi` is published to crates.io, so an old mirror against a new runtime reads `class_id` from a deleted slot with no compile error. A size/padding divergence being invisible is precisely the failure mode a header shrink has.

Issue #4717, opened 2026-06-06
Linux runtime: marked.parse crashes "Cannot read properties of undefined (reading 'extensions')" — this.defaults undefined (works on macOS, same 0.5.1125)


proggeramlug:
## Summary

On the **Linux** target, a compiled binary that imports [`marked`](https://www.npmjs.com/package/marked) and calls `marked.parse()` crashes at startup with:

```
TypeError: Cannot read properties of undefined (reading 'extensions')
    at <anonymous>
```

The **identical program, same Perry version (`0.5.1125`), same `marked@18.0.4`** runs correctly on **macOS**. So this is a Linux-runtime-specific regression in class-field / `this` resolution — it looks like the same family as the recently-merged #4616 (*"HTTP request handling for Hono apps — object-literal getter `this`, http2 instanceof, handle symbol read"*), which fixed a related `forEach`-on-undefined crash but not this `marked` path on Linux.

## Minimal reproduction

`package.json`:
```json
{
  "type": "module",
  "perry": {
    "compilePackages": ["marked"],
    "allow": { "compilePackages": ["marked"] }
  }
}
```

`mk.tsx`:
```ts
import { marked } from 'marked';
console.log('PARSED_OK len=' + String(marked.parse('# Hi', { async: false })).length);
```

```bash
npm i marked@18.0.4
npm i -g @perryts/perry@0.5.1125
```

### Linux (x86_64) — ❌ crashes
```bash
perry compile mk.tsx -o mk --target linux
./mk
# TypeError: Cannot read properties of undefined (reading 'extensions')
#     at <anonymous>
```

### macOS (arm64) — ✅ works
```bash
perry compile mk.tsx -o mk
./mk
# PARSED_OK len=15
```

`marked` is pristine/unmodified in both cases.

## Environment

| | |
|---|---|
| Perry | `@perryts/perry@0.5.1125` (prebuilt `@perryts/perry-linux-x64@0.5.1125` / `@perryts/perry-darwin-arm64@0.5.1125`) |
| marked | `18.0.4` |
| Linux | x86_64 glibc (reproduced on Debian and Ubuntu) — **fails** |
| macOS | arm64 (darwin) — **works** |
| Node (for `npm i`) | v22 |

> Note: `perry --version` prints `0.5.1122` on both platforms even though the npm package is `0.5.1125` (looks like a stale baked-in version string); the prebuilt packages' `package.json` confirm `0.5.1125`.

## Analysis

[...]

PR #8118, opened 2026-08-14
perf(object/json): admit JSON.parse receivers to the object-write fast paths (#8098)

### LINKED ISSUE #8098: perf(object): JSON.parse receivers have class_id == 0, so every write to parsed data misses both object-write fast paths (55x, identical checksum)
## Summary

An object produced by `JSON.parse` carries `class_id == 0`. That single fact
disqualifies it from **both** guarded object-write fast paths — the whole-loop
numeric clone and the static write PIC — so every property write to a parsed
object takes the generic `[[Set]]` runtime path for the life of the program.

Measured as a controlled pair inside the committed #6812 matrix: **9,560 ms vs
173 ms, a 55x intra-matrix difference, for an identical write count and an
identical checksum.**

This is the largest remaining row in `docs/object-write-matrix.md` and the only
one that has not improved since the matrix was first taken (57.8x on
2026-07-23 → 78.0x now).

## Evidence

`benchmarks/object-write-6812/matrix.ts` already contains the controlled pair.
`receiverClassIdZero` (`:501-515`) and `keyDot` differ in exactly one way — the
2400 receivers come from `JSON.parse('{"x":0}')` instead of an object literal —
and both perform 120,000,000 writes and produce `sink 122876400`.

Built from `origin/main` @ `514ff7c9a` (v0.5.1510), compiler plus both static
archives from a clean worktree, `PERRY_RUNTIME_DIR` pinned, default
auto-optimizing pipeline:

```
$ PERRY_TRACE_OBJECT_ARRAY_WRITE_GUARD=1 ./bin_matrix receiver_class_id_zero
PERRY_OBJECT_ARRAY_WRITE_GUARD_REJECT: first receiver is not an eligible regular shared-shape object
cell receiver_class_id_zero ms 9560 writes 120000000 sink 122876400

$ PERRY_TRACE_OBJECT_ARRAY_WRITE_GUARD=1 ./bin_matrix key_dot
cell key_dot ms 173 writes 120000000 sink 122876400
```

Against Node 26.5.1 on the same host, `run_matrix.py --samples 3` (the runner
verifies `(writes, sink)` on every pair and aborts on mismatch):

| cell | node | perry | P/node |
|---|--:|--:|--:|
| `receiver_class_id_zero` | 163 ms | 12,709 ms | **77.97** |
| `key_dot` | 1,054 ms | 742 ms | 0.70 |

`receiver_class_id_zero` is the worst of 26 cells by 4.3x; the next worst is
`rhs_call` at 18.3x. Full table:
https://github.com/PerryTS/perry/issues/6812#issuecomment-5298464076

## Where the two rejections are

- **Static write PIC** — `crates/perry-runtime/src/proxy/put_value.rs:337-341`
  requires `object_type == OBJECT_TYPE_REGULAR` **and `class_id != 0`** before
  it will prime a way. A parsed object never primes, so the site permanently
  misses.
- **Whole-loop numeric clone** — the preflight's per-element validation
  (`object_array_numeric_write_slots`, `proxy/put_value.rs:728`) requires every
  element to be a regular object with `class_id != 0` and one shared
  `GC_FLAG_SHAPE_SHARED` keys array. Hence the trace line above.

## Why this matters beyond a micro

`JSON.parse` is how essentially all external data enters a Perry program — HTTP
response bodies, config files, ORM rows, cache reads, IPC payloads. Any code
that parses records and then mutates fields on them is on the generic path
today, and no amount of loop-shape or key-form work in #6812 reaches it,
because the rejection is on the *receiver's identity*, not the write's shape.

## Sequencing — do not race #8067

The fix is to give parsed objects a real shape identity rather than to relax the
`class_id != 0` guards, which exist precisely because a class-id-zero object has
no shape token the PIC can compare. #8067 ("make ShapeId authoritative for keys,
live slots, and class-object kind") is doing exactly the work that would supply
one, and #8047 then changes the header ABI. This should land **after** #8067,
reusing its ShapeId, not by adding a parallel identity scheme.

Note the interaction with lazy parsing: `js_json_parse` is lazy for 1 KB–16 MB
top-level arrays (`json_tape`), so any benchmark or test here must either defeat
the tape or say that it did not — see the `GC_LAYOUT_POINTER_FREE` doc in
`crates/perry-runtime/src/gc/layout.rs` for the precedent where a lazy tape made
a probe vacuous (#7635).

## Acceptance

- `receiver_class_id_zero` reaches at least parity with the `key_dot` /
  `receiver_anonymous` rows (both currently ~0.5–0.7x node) under
  `benchmarks/object-write-6812/run_matrix.py` with 15 samples on an idle host.
- The trace no longer prints
  `first receiver is not an eligible regular shared-shape object` for that cell.
- No regression on the 14 cells that currently beat node, and identical
  `(writes, sink)` on all 26.
- Parity coverage for parsed objects that then get deleted keys, added keys,
  frozen/sealed, descriptors installed, or prototypes mutated — the cases the
  `class_id != 0` guard was standing in for.
- A tape-defeating (or explicitly tape-acknowledging) fixture, per #7635.

Split out of #6812, which delivered its coverage matrix and twelve slices; this
is the largest of its remaining rows and the only one with no improvement since
the baseline.

proggeramlug:
Closes #8098.

## What was wrong

A `JSON.parse` object carries `class_id == 0`. Both guarded object-write fast
paths rejected it on exactly that:

* the whole-loop numeric clone preflight — `object_array_numeric_write_slots`
  (`proxy/put_value.rs`), which is what printed
  `PERRY_OBJECT_ARRAY_WRITE_GUARD_REJECT: first receiver is not an eligible
  regular shared-shape object`;
* the static write PIC prime, the dynamic-key IC prime, and `dyn_ic_try_store`
  (which also serves the outlined ways 5-8), plus the **generated** hit paths in
  `perry-codegen/src/expr/proxy_reflect.rs`, which ANDed `class_id != 0` into
  every one of the four inline ways.

So every `record.field = …` on parsed data took the generic `[[Set]]` walk for
the life of the program. `JSON.parse` is how essentially all external data
enters a Perry program, and the rejection was on the receiver's *identity*, so
no amount of loop-shape or key-form work in #6812 could reach it.

## Measurements

Wall clock on the development host is unusable — the same cell measured
11,729 ms, 17,884 ms and 21,098 ms on the same binary. Everything below is
**instructions retired** (`/usr/bin/time -l`), which reproduced to within 0.02%.
Both arms are full `cargo build --release -p perry -p perry-runtime-static
-p perry-stdlib-static` builds snapshotted with sha256, with `PERRY_RUNTIME_DIR`
pinned to the snapshot; the compiler, both archives and the emitted benchmark
binary all `cmp` as different between arms.

The committed #6812 controlled pair (`benchmarks/object-write-6812/matrix.ts`),
identical 120,000,000 writes and identical `sink 122876400`:

| cell | before | after | |
|---|--:|--:|--:|
| `receiver_class_id_zero` | 150,082,612,917 | 1,169,725,264 | **128.3x fewer** |
| `key_dot` (reference) | 1,158,400,311 | 1,159,068,858 | unchanged |

After the fix `receiver_class_id_zero` is **1.009x of `key_dot`** — parity with
the object-literal cell, which is the issue's acceptance bar. The trace line is
gone.

The shape real code actually has is not the clone-matched nest but scattered
`.field =` writes through a helper, which take the PIC. A fixture doing
24,000,000 such writes over 2000 parsed records (`sink 48001997000` on every
arm, node included):

| | instructions | ms |
|---|--:|--:|
| before | 30,372,711,862 | 2291 |
| after | 2,508,280,501 | 165 |
| node 26.5.1 | — | 33 |

**12.1x fewer instructions**; vs node this goes from 66x to ~5x.

### No regression on the other 25 cells

Identical `(writes, sink)` on all 26. Four cells moved in wall clock by more
than a couple of percent, so they were re-measured in instructions:

| cell | before | after | delta |
|---|--:|--:|--:|
| `key_alternating_dynamic` (dyn-key IC) | 17,437,945,198 | 17,433,977,862 | -0.02% |
| `storage_inline` | 1,159,195,564 | 1,158,194,473 | -0.04% |
| `rhs_allocating` | 23,102,831,296 | 23,103,722,336 | +0.005% |
| `rhs_pointer` (write PIC, 96M writes) | 22,067,629,549 | 22,260,584,119 | **+0.87%** |

`rhs_pointer`'s +0.87% is the honest cost and it reconciles exactly: the emitted
guard gains 2 instructions per write, 96e6 x 2 = 1.92e8 = 0.87% of 2.2e10. Cells
that take the whole-loop clone pay nothing — the clone has no per-write guard.

## Why not the two obvious fixes

**Not** a synthetic `class_id` for parsed objects (the `__AnonShape_<hash>` +
`js_register_anon_shape_class_id` mechanism object literals use). `class_id != 0`
means "class instance" at roughly thirty runtime sites — `console.log` formatting,
`instanceof`, `getPrototypeOf`, `json/stringify.rs`'s fast template which
explicitly requires `class_id == 0`, `url/search_params`, `symbol/properties` —
and only about a third of them carry the anon-shape escape. That is the whole
introspection surface of parsed data, plus a JSON-stringify regression.

**Not** dropping the clause. It was standing in for three per-object exclusions
the generic path still applies verbatim
(`field_set_by_name/fast_paths.rs::try_existing_own_data_overwrite`):
`NATIVE_MODULE_CLASS_ID`, `Object.prototype`, and a `URL` instance — whose
`pathname`/`search`/… own slots are live views whose setters rebuild `href`
(`field_set_by_name/tail.rs`). None is derivable from the ShapeId: two objects
share a ShapeId iff they share a keys-array *allocation*, and the
shape-transition cache deliberately converges distinct objects onto one shared
array, so a prime-time-only exclusion loses to ordering — a plain object primes
the site at t0, a `URL` that acquires the same keys array at t1 > t0 then hits
it. The generated hit path re-checks only per-object state, so the
discriminator has to be per-object.

## What landed

An explicit, **opt-in**, per-object mark: `OBJ_FLAG_PLAIN_ORDINARY`, bit 9 of
`GcHeader::_reserved`, object-only and disjoint from the array-only
`GC_ARRAY_ARGUMENTS_OBJECT` by `obj_type` exactly as bits 11 and 12 already are
(`array_gc_header` refuses any header that is not `GC_TYPE_ARRAY`). The JSON
direct parser and the tape materializer set it at birth; one
`write_fast_path_receiver_kind_ok(obj, flags)` replaces the `class_id == 0`
clause at all four runtime guards, and the two emitted guards OR the bit into
their kind test.

It fails safe — every other class-less receiver is unmarked and keeps the full
`[[Set]]` walk, so no existing population changes behaviour — and it is free in
the generated guard, because `_reserved` is already loaded there.

Worth recording: the **read** PIC has admitted `class_id == 0` all along
(`field_get_set/ic_miss.rs` primes on any regular descriptor-free shaped
receiver; the emitted read guard has no `class_id` compare at all). Reads of
parsed objects were already on the ShapeId fast path — only writes were not.
#8067/#8086 supplied the rest: `js_object_alloc_class_inline_keys` birth-stamps
a parsed receiver with a real ShapeId, and `PARSE_SHAPE_CACHE` gives repeated
parses of one shape a single `GC_FLAG_SHAPE_SHARED` keys array, so the whole
2400-receiver prefix carries one ShapeId.

## Drive-by memory-safety fix

`JSON.parse("{}")` initialized **eight** inline field slots into an allocation
that has `max(0, INLINE_SLOT_FLOOR)` = **two** of them (the floor dropped 4 -> 2
in #7928) — a 48-byte overwrite past the object on every empty-object parse,
the exact "heap buffer overflow into adjacent arena objects" that
`js_object_alloc_with_parent` documents. The hand-rolled fill was redundant as
well: the allocator has initialized every slot it allocates since #4717.

No unit test: under a forward bump allocator the overflow lands in
not-yet-allocated space on every path a test can construct, so a deterministic
corruption probe is not available — it only bites at an arena block boundary or
on the malloc path. The arithmetic is in the code comment.

## Coverage

Three new runtime unit tests (`cargo-test`-visible), each **sabotage-verified**
with a real rebuild that recompiled `perry-runtime`:

| sabotage | result |
|---|---|
| guard ignores the mark (`class_id != 0` only) | 3 tests RED, including the pre-existing uniform-proof test |
| delete the parser's `mark_object_plain_ordinary` call | 2 tests RED |
| `OBJ_FLAG_PLAIN_ORDINARY` 0x200 -> 0x1000 | ABI-pin test RED |
| restored | 4/4 green |

* `json_parse_receivers_are_admitted_to_the_whole_loop_write_clone` drives the
  shipped `js_json_parse` end-to-end, asserts the premises (`class_id == 0`, a
  real shared ShapeId), then clears the mark on **one** receiver and requires
  the guard to refuse — same objects, same ShapeId, same slots, only the mark
  differs.
* `json_parse_receivers_prime_the_static_write_pic` — same discrimination on the
  PIC prime path.
* `plain_ordinary_object_flag_matches_the_emitted_write_pic_literal` pins the
  bit against the literal `perry-codegen` emits.
* The pre-existing `object_array_numeric_write_guard_requires_complete_uniform_proof`
  keeps its class-id-zero rejection for an *unmarked* receiver and gains the
  marked-accepts and native-module-still-rejects halves.

Payloads are a few bytes with an object root, so `js_json_parse` takes the eager
direct parser and no lazy tape stands between the probes and the objects they
inspect (#7635). The tape is not material to this benchmark — it needs
`1 KB <= len <= 16 MB` **and** a `[` root — but it IS material to the workloads
the issue is about, which is why `json_tape.rs` and `json_tape/iterative.rs` are
marked too.

`test-files/test_gap_json_parse_object_writes.ts` is parity against node 26.5.1
for the semantics the `class_id != 0` clause used to keep parsed objects away
from: deleted keys, added keys, frozen/sealed/non-extensible receivers (strict
`TypeError`s), accessor and non-writable descriptors installed over a parsed
slot, prototype mutation with a shadowing setter, null prototypes, dynamic-key
writes, a parsed object used as a prototype, an empty parsed object grown by
name, a polymorphic site mixing parsed objects / literals / class instances, and
`__proto__` / `constructor` as genuine own data keys. It is byte-identical to
node **before and after** — by design; it is a semantics regression guard for
the newly-enabled fast paths, not a witness for the perf fix, whose witnesses
are the sabotage-verified unit tests above.

## Local validation

* `cargo test -p perry-runtime --lib` -> **2354 passed, 0 failed, 4 ignored**
  (baseline 2351/0/4 + the 3 new tests).
* `cargo test -p perry-codegen --no-fail-fast` -> **27 suites, 1434 passed,
  11 failed** — byte-for-byte the documented pre-existing set (#8092); the
  failing names are the buffer-read / typed-feedback / back-edge-poll cluster,
  none of them touched here.
* `lint`: `cargo fmt --all -- --check`, `check_file_size.sh`,
  `workspace_architecture.py --check`, `addr_class_inventory.py`,
  `class_id_collisions.py`, `gc_store_site_inventory.py`,
  `gc_runtime_root_holders.py`, `gc_pin_sites.py`, `check_gc_env_knobs.py`,
  `check_gc_doc_claims.py`, `check_locale_independent_io.py`,
  `raw_handle_debt.py`, `local_binding_type_audit.py`,
  `global_sink_isolation.py`, `check_test_registration.py` — all green.

## Summary by CodeRabbit

- **Performance**
  - Improved property-write performance for objects created by `JSON.parse`, including repeated, static, and dynamic writes.
  - Optimized handling applies while preserving correct behavior for special object types and unsupported cases.

- **Bug Fixes**
  - Fixed an issue affecting initialization of empty parsed objects.
  - Strengthened write behavior across frozen, sealed, non-extensible, accessor, prototype, and inheritance scenarios.

- **Tests**
  - Added broad coverage for JSON-parsed object writes, deletions, additions, and edge-case property behavior.

### DIFF
--- a/changelog.d/8118-json-parse-write-fast-paths.md
+++ b/changelog.d/8118-json-parse-write-fast-paths.md
@@ -1,0 +1,85 @@
+### perf(object/json): `JSON.parse` receivers reach both object-write fast paths (#8098)
+
+A `JSON.parse` object carries `class_id == 0`. Both guarded object-write fast
+paths — the whole-loop numeric clone and the static/dynamic write PICs — rejected
+it on exactly that, so every `record.field = …` on parsed data took the generic
+`[[Set]]` path for the life of the program. `JSON.parse` is how essentially all
+external data enters a Perry program (HTTP bodies, config, ORM rows, cache reads,
+IPC payloads), and the rejection was on the *receiver's identity*, so no amount of
+loop-shape or key-form work in #6812 could reach it.
+
+Measured on the committed `benchmarks/object-write-6812/matrix.ts` controlled
+pair, which differs in exactly one way (`JSON.parse('{"x":0}')` instead of an
+object literal) and produces an identical `sink 122876400` over identical
+120,000,000 writes. Wall clock on the development host is unusable (the same cell
+measured 11.7 s / 17.9 s / 21.1 s across three runs), so the ratio is reported in
+**instructions retired**, which reproduced to within 0.02%:
+
+| cell | instructions retired | vs `key_dot` |
+|---|--:|--:|
+| `key_dot` (object literals) | 1.158e9 | 1.00x |
+| `receiver_class_id_zero` before | 150.08e9 | **129.5x** |
+| `receiver_class_id_zero` after | see below | |
+
+**Why the guard could not simply drop the clause.** `class_id != 0` was standing
+in for three per-object exclusions that the generic path still applies verbatim
+(`object/field_set_by_name/fast_paths.rs::try_existing_own_data_overwrite`):
+`NATIVE_MODULE_CLASS_ID`, `Object.prototype`, and a `URL` instance — whose
+`pathname`/`search`/… own slots are live views whose setters rebuild `href`
+(`field_set_by_name/tail.rs`). None of those is derivable from the ShapeId: two
+objects share a ShapeId iff they share a keys-array *allocation*, and the
+shape-transition cache deliberately converges distinct objects onto one shared
+array, so a prime-time-only exclusion loses to ordering (a plain object primes the
+site, a `URL` that later acquires the same keys array then hits it). The
+generated hit path re-checks only per-object state, so the discriminator has to
+be per-object too.
+
+**What landed instead** is an explicit, opt-in, per-object mark:
+`OBJ_FLAG_PLAIN_ORDINARY` (bit 9 of `GcHeader::_reserved`, object-only, disjoint
+from the array-only `GC_ARRAY_ARGUMENTS_OBJECT` by `obj_type` the same way bits 11
+and 12 already are). The JSON direct parser and the lazy-tape materializer set it
+at birth; every other class-less receiver is unmarked and keeps the full `[[Set]]`
+walk, so no existing population changes behaviour. The bit is free in the
+generated guard — `_reserved` is already loaded there for the blocking-flag test,
+so admission costs one `and` + `icmp` + `or`, hoisted above the four PIC ways.
+
+Note that the *read* PIC has admitted `class_id == 0` all along
+(`object/field_get_set/ic_miss.rs` primes on any regular descriptor-free shaped
+receiver, and the emitted read guard has no `class_id` compare at all). Reads of
+parsed objects were already on the ShapeId fast path; only writes were not. #8067
+/ #8086 supplied what was missing on the write side: a parsed receiver is
+birth-stamped with a real ShapeId by `js_object_alloc_class_inline_keys`, and
+repeated parses of one shape share a single `GC_FLAG_SHAPE_SHARED` keys array via
+`PARSE_SHAPE_CACHE`, so the whole 2400-receiver prefix carries one ShapeId.
+
+Also fixed here, in the same file: `JSON.parse("{}")` initialized **eight** inline
+field slots into an allocation that has `max(0, INLINE_SLOT_FLOOR)` = **two** of
+them (the floor dropped 4 → 2 in #7928) — a 48-byte overwrite past the object on
+every empty-object parse, the exact "heap buffer overflow into adjacent arena
+objects" that `js_object_alloc_with_parent` documents. The hand-rolled fill was
+redundant as well: the allocator has initialized every slot it allocates since
+#4717.
+
+Coverage:
+
+* `crates/perry-runtime/src/proxy.rs` —
+  `json_parse_receivers_are_admitted_to_the_whole_loop_write_clone` and
+  `json_parse_receivers_prime_the_static_write_pic` drive the shipped
+  `js_json_parse` end-to-end (payloads are a few bytes with an object root, so
+  the eager direct parser runs and no lazy tape stands between the probe and the
+  objects — #7635), assert the premises (`class_id == 0`, a real shared ShapeId),
+  and then clear the mark on one receiver and require the guard to refuse. Both
+  fail when the guard ignores the mark and when the parser stops setting it.
+  `plain_ordinary_object_flag_matches_the_emitted_write_pic_literal` pins the bit
+  value against the literal `perry-codegen` emits.
+* The pre-existing `object_array_numeric_write_guard_requires_complete_uniform_proof`
+  keeps its class-id-zero rejection for an *unmarked* receiver and gains the
+  marked-accepts and native-module-still-rejects halves.
+* `test-files/test_gap_json_parse_object_writes.ts` — parity against node 26.5.1
+  for the semantics the `class_id != 0` clause used to keep parsed objects away
+  from: deleted keys, added keys, frozen/sealed/non-extensible receivers (strict
+  `TypeError`s), accessor and non-writable descriptors installed over a parsed
+  slot, prototype mutation with a shadowing setter, null prototypes, dynamic-key
+  writes, a parsed object used as a prototype, an empty parsed object grown by
+  name, a polymorphic site mixing parsed objects / literals / class instances,
+  and `__proto__` / `constructor` as genuine own data keys.
--- a/crates/perry-codegen/src/expr/proxy_reflect.rs
+++ b/crates/perry-codegen/src/expr/proxy_reflect.rs
@@ -50,6 +50,15 @@ use super::{
 /// encoded by the authoritative ShapeId and therefore owns no header flag.
 const WRITE_PIC_BLOCKING_FLAGS: u16 = 0x1907;
 
+/// #8098: `GcHeader::_reserved` bit 9 — the runtime birth-marked this
+/// class-less receiver an ORDINARY plain object (`JSON.parse` output), so it is
+/// eligible for the write PIC exactly like a class instance. MUST equal
+/// `perry_runtime::gc::OBJ_FLAG_PLAIN_ORDINARY`; the runtime pins the value in
+/// `proxy::tests::plain_ordinary_object_flag_matches_the_emitted_write_pic_literal`.
+/// It is deliberately NOT in `WRITE_PIC_BLOCKING_FLAGS` — this bit ADMITS a
+/// receiver, the blocking mask REJECTS one.
+const PLAIN_ORDINARY_OBJ_FLAG: u16 = 0x200;
+
 /// The NaN-boxed `undefined` literal, for an absent optional operand.
 fn undefined_literal() -> String {
     double_literal(f64::from_bits(crate::nanbox::TAG_UNDEFINED))
@@ -550,8 +559,17 @@ fn lower_put_value_static_write_ic(
     let class_addr = ctx.block().add(I64, &safe_target, "4");
     let class_ptr = ctx.block().inttoptr(I64, &class_addr);
     let class_id = ctx.block().load(I32, &class_ptr);
-    let class_nonzero = ctx.block().icmp_ne(I32, &class_id, "0");
+    let has_class = ctx.block().icmp_ne(I32, &class_id, "0");
     let not_native_module = ctx.block().icmp_ne(I32, &class_id, "-2");
+    // #8098: a class-less receiver qualifies when the runtime birth-marked it
+    // an ordinary plain object. `reserved` is already loaded above for the
+    // blocking-flag test, so this costs one `and` + `icmp` + `or`, computed
+    // once here and reused by all four ways (this block dominates them).
+    let plain_ordinary_bits = ctx
+        .block()
+        .and(I16, &reserved, &PLAIN_ORDINARY_OBJ_FLAG.to_string());
+    let plain_ordinary = ctx.block().icmp_ne(I16, &plain_ordinary_bits, "0");
+    let receiver_kind_ok = ctx.block().or(I1, &has_class, &plain_ordinary);
 
     // The write PIC uses the same single ShapeId token domain as the read PIC.
     let shape_id_addr = ctx.block().add(I64, &safe_target, "8");
@@ -574,7 +592,7 @@ fn lower_put_value_static_write_ic(
     let mut hit = ctx.block().and(I1, &heap_candidate, &gc_object);
     hit = ctx.block().and(I1, &hit, &not_forwarded);
     hit = ctx.block().and(I1, &hit, &flags_clear);
-    hit = ctx.block().and(I1, &hit, &class_nonzero);
+    hit = ctx.block().and(I1, &hit, &receiver_kind_ok);
     hit = ctx.block().and(I1, &hit, &not_native_module);
     hit = ctx.block().and(I1, &hit, &token_match);
     hit = ctx.block().and(I1, &hit, &token_nonzero);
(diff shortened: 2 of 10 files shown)


proggeramlug:
Verified the two claims I could check without building. Both hold.

## The drive-by memory-safety fix is real, and the arithmetic is exact

```
object/mod.rs:49    pub(crate) const INLINE_SLOT_FLOOR: usize = 2;
object/mod.rs:31    "8 (GcHeader) + 32 (ObjectHeader) + 8 * max(field_count, INLINE_SLOT_FLOOR)"
json/parser.rs:652  js_object_alloc_class_inline_keys(0, 0, 0, keys_arr)   -> max(0, 2) = 2 slots
json/parser.rs:655  for i in 0..8 { std::ptr::write(fields_ptr.add(i), ...) }   -> 8 slots written
```

**6 JSValues × 8 bytes = 48 bytes past the object, on every `JSON.parse("{}")`.** Precisely the "heap buffer overflow into adjacent arena objects" that `js_object_alloc_with_parent` warns about, and it has been latent since `INLINE_SLOT_FLOOR` dropped 4 → 2 in #7928.

Your reason for shipping it without a test is the right call and worth keeping in the changelog: under a forward bump allocator the overflow lands in not-yet-allocated space on every constructible path, so it only bites at a block boundary or on the malloc path. A test that cannot fail would be worse than the note.

Worth flagging that this is the **second** consequence of #7928's floor change to surface today — the #6812 audit found the issue body still asserting the floor is 4 and no matrix cell re-measured after it moved. A change to that constant evidently has a wider blast radius than its own PR.

## The `class_id` diagnosis holds, and the precedent is the convincing part

That the **read** PIC has admitted `class_id == 0` all along — no `class_id` compare in its emitted guard at all — is the strongest single piece of evidence here. It shows the write-side compare was never load-bearing for shape identity, which is exactly what #8086 demoted `class_id` for.

Your finding that `class_id != 0` was standing in for three unrelated per-object exclusions (`NATIVE_MODULE_CLASS_ID`, `Object.prototype`, `URL`) is the substantive part, and the reason an opt-in `OBJ_FLAG_PLAIN_ORDINARY` mark is the right shape rather than deriving eligibility from ShapeId: the transition cache converges distinct objects onto one keys array, so a prime-time-only exclusion loses to ordering. That argument is what makes this safe.

Marking `json_tape.rs` and `json_tape/iterative.rs` as well as the parser is the difference between fixing the benchmark and fixing the workload — a 7-byte `'{"x":0}'` never reaches the tape, but the ORM/HTTP-body population this issue is actually about does. A parser-only fix would have measured beautifully and shipped nothing.

## What I have not verified

The instruction counts, the sabotage runs, and the suite results — I am relying on your report for those. What I checked is the reasoning and the two source-level claims above.

The **+0.87% on `rhs_pointer`** disclosed and reconciled arithmetically (2 extra guard instructions × 96 M writes) is the most credible number in the report, precisely because it is the one that costs something. A change of this size with no reported regression anywhere would have been the thing to distrust.

Leaving it draft pending my own run of the enumerated lint set and the two suites on a landing tree.

proggeramlug:
Landing-equivalent validation is clean. **Merging.**

I merged current `main` (`9aadf1861`, which added #8106 and #8109 after this branch's base) and re-ran, because a codegen-guard change and a root-reload-allowlist change landing in the same window is exactly where an interaction would hide:

```
merge with main                       CLEAN, no conflicts
cargo test -p perry-runtime --lib     2364 passed, 0 failed, 4 ignored   (main's 2361 + your 3)
cargo test -p perry-codegen --no-fail-fast   27 suites, 1434 passed, 11 failed — the #8092 baseline set exactly
cargo fmt --all -- --check            clean
./scripts/check_file_size.sh          clean
34 enumerated lint checks             all green
```

No version bump, no manifest or lockfile change.

## What is landing

**128.3× fewer instructions** on `receiver_class_id_zero` (150,082,612,917 → 1,169,725,264), arriving at **1.009× of the `key_dot` reference** — the acceptance bar this issue set. `sink 122876400` unchanged everywhere, all 26 matrix cells keeping identical `(writes, sink)`.

And a **48-byte heap overflow on every `JSON.parse("{}")`**, which I verified arithmetically: `INLINE_SLOT_FLOOR` is 2, the call passes `field_count = 0` so `max(0, 2) = 2` slots are allocated, and the removed loop wrote 8. Latent since #7928 moved the floor from 4 to 2.

## The two things that make this reviewable rather than just fast

**The precedent, not the measurement.** That the read PIC has admitted `class_id == 0` all along — no `class_id` compare in its emitted guard at all — is what establishes the write-side compare was never load-bearing for shape identity. A 128× number with no such argument would have warranted much more suspicion than this did.

**The disclosed cost.** `rhs_pointer` +0.87%, reconciled exactly as 2 extra guard instructions × 96 M writes. A change of this magnitude reporting no regression anywhere would have been the thing to distrust; a change that finds its own 0.87% and explains it arithmetically is the opposite.

Marking the tape paths as well as the parser is what separates fixing the benchmark from fixing the workload — `'{"x":0}'` is 7 bytes and never reaches the tape, but the ORM/HTTP-body population the issue is actually about does.

Merging with `--admin`, as every merge currently must (#8092, #8117).

Issue #8111, opened 2026-08-14
Reassigning a `new Uint8Array` binding drops element writes and reads `undefined` (the Uint8-specialized twin of #8100)


proggeramlug:
Found while fixing #8100 (PR #8109), which is the same disease in the
`js_typed_array_*` element helpers. This one is in the `Uint8Array`-specialized
pair and is **not** fixed by that PR — measured pre-fix and post-fix on the
same probe, byte-identical output, so #8109 neither causes nor fixes it.

## Reproducer

```ts
let Q: Uint8Array = new Uint8Array(2);
Q = [9, 10] as any;
Q[0] = 5;
console.log("store:", JSON.stringify(Q));
let k: any = 1;
console.log("dyn:", Q[k]);
console.log("read:", Q[0], Q[1], Q.length, Q.at(1));
let R: Uint8Array = new Uint8Array([3, 4]);
console.log("real u8:", R[0], R[1], R.length);
const B = Buffer.from([1, 2]);
console.log("buf:", B[0], B[1]);
```

```
node 26.5.1     perry
store: [5,10]   store: [9,10]                <- the write is SILENTLY DROPPED
dyn: 10         dyn: undefined
read: 5 10 2 10 read: undefined undefined 2 10
real u8: 3 4 2  real u8: 3 4 2               (control, OK)
buf: 1 2        buf: 1 2                     (control, OK)
```

Both sides exit 0. `Q.length` and `Q.at(1)` are already correct, so the binding
really does hold the plain array — only the specialized element accessors are
wrong.

Reproduced at `b5ab0262e` (#8109's branch) and at its parent, `--profile
perry-dev`, `--no-auto-optimize --no-cache`, `PERRY_RUNTIME_DIR` pinned to a
freshly built archive pair.

## Root cause

`crates/perry-runtime/src/typedarray/access.rs`:

```rust
// :605
pub extern "C" fn js_uint8array_index_get_value(target: *const TypedArrayHeader, index: i32) -> f64 {
    ...
    if let Some(kind) = lookup_typed_array_kind(addr) {
        if !matches!(kind, KIND_UINT8 | KIND_UINT8_CLAMPED) { return undefined; }
        js_typed_array_get(addr as *const TypedArrayHeader, index)
    } else if crate::buffer::is_registered_buffer(addr) {
        crate::buffer::js_buffer_index_get_value(...)
    } else {
        undefined          // <- a plain array / object lands here
    }
}

[...]

Issue #3148, opened 2026-05-30
TypedArray: remaining prototype methods (fill/copyWithin/reduce/join/indexOf/slice/…) + instanceof


proggeramlug:
## Summary

Follow-up to **PR #2499** (which wired `TypedArray.prototype` `map`/`filter`/`every`/`some`/`forEach`/`find`/`findIndex`). The remaining `%TypedArray%.prototype` methods are still unwired or broken, and `x instanceof Int32Array` is false for any typed array.

## Concrete gaps (Perry vs Node)

```js
const a = new Int32Array([1,2,3,4,5]);
a.fill(7);                 // Perry: TypeError "(number).fill is not a function"   Node: [7,7,7,7,7]
a.copyWithin(0,3);         // Perry: "(number).copyWithin is not a function"       Node: [4,5,3,4,5]
a.reverse();               // in-place; Node returns same TA reversed
a.reduce((s,x)=>s+x,0);    // Node 15
a.join("-");               // Node "1-2-3-4-5"
a.indexOf(3);              // Node 2
a.lastIndexOf(3);          // Node 2
a.includes(3);             // Node true
a.slice(1,3);              // Node Int32Array [2,3]   (same-kind TA)
a.subarray(1,3);           // Node Int32Array view [2,3]
a.set([9,9], 1);           // writes 9,9 at offset 1
for (const v of a.keys()) {}    // 0..n-1
for (const v of a.values()) {}  // elements
for (const [i,v] of a.entries()) {}

new Int32Array([1,2,3]) instanceof Int32Array   // Perry: false   Node: true
```

## Where the code lives (pattern is established by PR #2499)

The established pattern: the generic `js_array_*` helper detects a typed-array receiver via `crate::typedarray::lookup_typed_array_kind(arr as usize).is_some()` and delegates to a `js_typed_array_*` impl that reads/writes via `load_at`/`store_at` (element-typed).

- `crates/perry-runtime/src/typedarray.rs` — add `js_typed_array_{fill,copy_within,reverse,reduce,reduce_right,join,index_of,last_index_of,includes,slice,subarray,set,keys,values,entries}` (mirror the `map`/`filter`/`find_last` impls already there). `map`/`filter`/`slice`/`subarray` return a **same-kind** TypedArray.
[...]

Issue #2879, opened 2026-05-30
runtime: implement TypedArray.prototype.set and copyWithin for non-Uint8Array views


andrewtdiz:
## Summary

Generic TypedArray instances still lack Node-compatible `TypedArray.prototype.set(...)` and typed-aware `copyWithin(...)` behavior. `Uint8Array`/Buffer-backed values have a separate buffer dispatch path, but other typed arrays such as `Int16Array`, `Int32Array`, and `Float64Array` either fall through generic dispatch or reuse ordinary Array storage logic.

## Node behavior

Observed with local Node v25.9.0:

```js
for (const C of [Int16Array, Uint8Array, Float64Array]) {
  const a = new C([1, 2, 3, 4]);
  a.set(new C([9, 8]), 1);
  a.copyWithin(0, 2, 4);
  console.log(C.name, Array.from(a));
}
// Int16Array [8, 4, 8, 4]
// Uint8Array [8, 4, 8, 4]
// Float64Array [8, 4, 8, 4]
```

`set` bulk-copies/coerces source elements into the receiver, validates offset/range, and returns `undefined`. `copyWithin` mutates the same typed-array receiver using element storage, not boxed Array slots.

## Perry evidence

- `crates/perry-runtime/src/typedarray.rs` exposes scalar `js_typed_array_set(ta, index, value)` for `ta[i] = value`, plus typed helpers for methods such as `at`, `sort`, `toSorted`, `toReversed`, `with`, `map`, `filter`, `some`, and `every`, but there is no bulk `TypedArray.prototype.set(source, offset)` helper and no typed-array `copyWithin` helper.
- `crates/perry-runtime/src/object/native_call_method.rs` has a raw TypedArray dispatch arm, but it only routes `length`, `at`, `sort`, `toSorted`, `toReversed`, `with`, `findLast`, and `findLastIndex`. `set` and `copyWithin` fall through.
- `crates/perry-runtime/src/object/buffer_dispatch.rs` handles `set`, `slice`, and `subarray` for Buffer/Uint8Array-shaped receivers only, via `js_buffer_set_from_value` / `js_buffer_slice`. That does not cover `Int16Array`, `Float64Array`, BigInt typed arrays, etc.
[...]

PR #8119, opened 2026-08-14
fix(array): ask the typed-array question before clean_arr_ptr rejects it, for sort/toSorted/toReversed/with

### LINKED ISSUE #8096: Typed-array `sort`, `with` and `toReversed` are still silent no-ops (same cause as #8090)
## Summary

#8090 fixed `fill`, ranged `fill`, `reverse` and `copyWithin` on typed-array receivers by asking `typed_array_receiver()` **before** `clean_arr_ptr` rejects them. Its author noted that `sort`, `with` and `toReversed` are broken by the same cause and were deliberately left out of scope. Filing so they are not lost.

## The mechanism (already established by #8090)

Codegen routes statically-typed typed-array receivers through the generic `js_array_*` helpers on purpose (#3148 / #654 — `is_array_expr` answers `true` for `Int32Array` &co.), on the contract that each helper re-dispatches on `lookup_typed_array_kind`.

`clean_arr_ptr` rejects those receivers, and must: since #7574 it returns null for every tracked non-`GC_TYPE_ARRAY` object, because a `TypedArrayHeader`'s raw per-kind storage is not boxed-f64 `ArrayHeader` slots. Since the 2026-07-09 typed-array audit gave every typed array a real `GC_TYPE_TYPED_ARRAY` header, that rejection fires for all of them.

So any delegation written **below** the clean is unreachable, and the helper silently returns the receiver unmutated — no error, no diagnostic.

## Expected shape of the fix

Same as #8090: call `array::header::typed_array_receiver()` ahead of the `clean_arr_ptr` funnel in each of the three helpers and delegate to the element-typed twin. `typed_array_receiver` already exists on `main` and strips the NaN-box tag itself without dereferencing.

## Acceptance

A reproducer that is byte-exact against the pinned node oracle (`v26.5.1`, `.node-version`) and that **demonstrably fails without the fix** — #8090's did:

```
                 node          pre-fix perry
fill         9 9 3 4          1 2 3 4  (no-op)
reverse      4 3 2 1          1 2 3 4  (no-op)
copyWithin   3 4 3 4          1 2 3 4  (no-op)
```

Please include a plain-array control (it must be correct in both arms, proving the defect is typed-array-specific) and at least one non-`Uint16Array` element width. `sort` additionally needs a comparator case and the spec's default numeric ordering for typed arrays, which differs from `Array.prototype.sort`'s string ordering — worth an explicit test, since that is a second way these can silently disagree with node.

Mirror #8090's negative coverage too: a test that `clean_arr_ptr` still rejects a typed-array receiver (keeping #7574 pinned), and one that plain-array mutators are unchanged by the new pre-check.

Refs #8090, #3148, #7574, #2879.

proggeramlug:
Closes #8096.

## What was wrong

#8090 fixed `fill`, ranged `fill`, `reverse` and `copyWithin` by asking
`array::header::typed_array_receiver()` **before** the `clean_arr_ptr` funnel,
and named `sort` / `with` / `toReversed` as still carrying the identical dead
delegation. They do — and so does `toSorted`, both arities:

| helper | file | clean at | delegation at |
|---|---|---|---|
| `js_array_to_reversed` | `array/immutable.rs:32` | 33 | 37 |
| `js_array_to_sorted_default` | `array/immutable.rs:59` | 60 | 61 |
| `js_array_to_sorted_with_comparator` | `array/immutable.rs:96` | 105 | 106 |
| `js_array_with` | `array/immutable.rs:232` | 242 | 246 |
| `js_array_sort_default` | `array/sort.rs:542` | 551 | 561 |
| `js_array_sort_with_comparator` | `array/sort.rs:635` | 651 | 658 |

Codegen routes statically-typed typed-array receivers through the generic
`js_array_*` helpers on purpose (#3148 / #654) on the contract that each helper
re-dispatches. `clean_arr_ptr` rejects those receivers, and must: since #7574 it
returns null for every tracked non-`GC_TYPE_ARRAY` object. So every delegation
written below the clean was unreachable.

**Two distinct silent failures.** In-place `sort` returned the null'd pointer
having sorted nothing. `toReversed` / `toSorted` / `with` returned
`js_array_alloc(0)` — an **empty plain array**, so `.constructor.name` was
`"Array"` and `JSON.stringify(Array.from(x))` was `[]`.

## A second receiver shape, found while validating

`new Uint8Array([…])` in perry is **not** a registry `TypedArrayHeader`.
`buffer::js_uint8array_new` returns a `BufferHeader`, registered as a buffer and
marked `mark_as_uint8array`. So `typed_array_receiver` — which is registry-backed
— legitimately answers `None` for the most common typed array in the language,
while `clean_arr_ptr` still rejects it as a tracked non-array. Verified directly:

```
addr=0x20000b00008
is_registered_buffer=true
lookup_typed_array_kind=None
gc_header_obj_type=Some(10)
clean_arr_ptr_null=true
js_array_to_reversed(...).length = 0
```

Most entry points never see it: `sort` / `with` / `reverse` / `fill` on that
shape resolve through the dynamic method dispatcher, and `copyWithin` grew its
own Buffer arm in #8090. But `toReversed` and `toSorted` fold **unconditionally**
in HIR (`lower/expr_call/local_array_methods.rs`, no receiver-type guard), and
the dynamic tower's own `toReversed` / `toSorted` arms call straight back into
these same helpers — so those two were wrong for `Uint8Array` and `Buffer` on
every dispatch path. New `buffer_receiver_as_uint8_typed_array` resolves that
shape into a fresh `KIND_UINT8` copy. **It is sound only for the immutable
methods** (a copy-based in-place `sort` would leave the receiver untouched — a
different wrong answer), which the doc comment says and `js_array_sort_*`
respects.

## Evidence

Three-way, exit codes checked on every run: node `v26.5.1` (the `.node-version`
pin), pre-fix perry, post-fix perry. Both perry arms built with `--profile
perry-dev` from `a99732458`, compiled `--no-auto-optimize --no-cache`,
`PERRY_RUNTIME_DIR` pinned to a freshly built `libperry_{runtime,stdlib}.a` pair
(`grep -c "Compiling perry-runtime v"` = 1; archive mtime 108 s after the edit).

```
                          node                perry before
i32 sort default      1 2 9 10            10 9 2 1        <- NO-OP
f64 sort default      1 2 9.25 10.5       10.5 9.25 2 1   <- NO-OP
i32 sort cmp desc     10 9 2 1            1 10 2 9        <- NO-OP
f64 sort cmp desc     10.5 9.25 2 1       1 10.5 2 9.25   <- NO-OP
i32 toSorted          1 2 9 10            undefined x4    <- EMPTY ARRAY
i32 toSorted cmp      10 9 2 1            undefined x4
f64 toSorted          1 2 9.25 10.5       undefined x4
i32 toReversed        4 3 2 1             undefined x4
i8  toReversed        4 3 2 1             undefined x4
f64 toReversed        4.5 3.5 2.5 1.5     undefined x4
i32 with              1 99 3 4            undefined x4
f64 with              1.5 2.5 7.25 4.5    undefined x4
i8  with wraps        -56                 undefined
ctor names            Int32Array x3       Array x3
json                  [4,3,2,1] [1,99,3,4]  [] []
u8  toReversed        4 3 2 1             undefined x4
u8  toSorted          1 2 9 10            undefined x4
```

**Plain-array controls, correct in BOTH arms** — this is what makes the defect
typed-array-specific rather than a general mutator break:

```
plain sort default    1 10 2 9   (ToString order; typed sort is NUMERIC)
plain sort cmp desc   10 9 2 1
plain toSorted        1 10 2 9
plain toReversed      4 3 2 1
plain with            1 99 3 4
```

Five probes (`p8096`, `u8ann`, `u8dyn`, `u8probe`, plus #8111's) are **byte-identical
to node** post-fix, all exit 0.

Two notes on how the reproducer was built, both of which cost a false pass first
time round:

* The comparator inputs are deliberately **not** already in the target order.
  `[10,9,2,1].sort((x,y)=>y-x)` expects `10 9 2 1` — which is exactly what a
  no-op produces. The probe feeds `[1,10,2,9]`.
* `sort` / `toSorted` with no comparator need the typed helper for the **order**
  as well as the layout: `%TypedArray%` sorts NUMERICALLY (§23.2.3.29
  CompareTypedArrayElements), `Array.prototype` by `ToString`. `[10,9,2,1]`
  separates all three answers — numeric `1,2,9,10`, string `1,10,2,9`, no-op
  `10,9,2,1`.

## Tests

**12 added** to `crates/perry-runtime/src/array/typed_array_receiver_tests.rs`
(19 in the file with #8090's seven).

Sabotage-verified in two stages, each after a real rebuild
(`grep -c "Compiling perry-runtime v"` = 1 on both):

* reverting `immutable.rs` + `sort.rs` to `origin/main` → **6 of the new ones
  fail**;
* neutering `buffer_receiver_as_uint8_typed_array` to `return None` → **the
  other 3 fail**.

The 3 that stay green in both arms are the intended controls: the plain-`Array`
control (proving the new pre-check did not hijack the ordinary path), the
`ArrayBuffer`/`DataView` decline, and
`a_new_uint8array_is_a_buffer_not_a_registry_typed_array`, which pins the
precondition the Buffer arm exists for — if `new Uint8Array` ever starts
producing a registry typed array, the arm is redundant and that test failing is
the signal to re-read the constructor. #8090's
`clean_arr_ptr_still_rejects_a_typed_array_receiver` keeps #7574 pinned
alongside them.

Every mutator test reads the **result**, not just the receiver, because the
broken helpers returned a fresh empty array rather than the unmutated input; the
`Uint16Array` cases assert `70000 & 0xFFFF == 4464` and the `Uint8Array` one
`300 -> 44`, so a raw-f64 memcpy regression cannot pass either.

`js_array_to_spliced` needs no equivalent — `%TypedArray%.prototype` has no
`toSpliced`.

## Suites

* `cargo test -p perry-runtime --lib` → **2380 passed, 0 failed, 4 ignored**
  (main: 2361 / 0 / 4; +19 new).
* `cargo test -p perry-codegen --no-fail-fast` → 27 suites, 1434 passed,
  **11 failed** — identical to the `main` baseline (pre-existing, #8092).
* `cargo fmt --all -- --check`, `check_file_size.sh`, and the 20 `lint` scripts
  enumerated from `test.yml` all pass. `addr_class_inventory.py` exits 0 (its
  "baseline is stale" note names `object/field_set_by_name/tail.rs`, a file this
  PR does not touch).

Refs #8090, #3148, #654, #7574, #2879.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed typed-array sorting to use numeric ordering, including comparator-based sorting.
  - Fixed `toSorted`, `toReversed`, and `with` for typed arrays and Buffer-backed `Uint8Array` values.
  - Preserved typed-array result types, element coercion, immutability, and negative-index behavior.
  - Kept `ArrayBuffer` and `DataView` behavior unchanged.

- **Tests**
  - Added regression coverage for sorting, immutable methods, element replacement, and Buffer handling.

### DIFF
--- a/changelog.d/8119-typed-array-immutables-and-sort.md
+++ b/changelog.d/8119-typed-array-immutables-and-sort.md
@@ -1,0 +1,110 @@
+### Fixed
+
+- **Typed-array `sort`, `toSorted`, `toReversed` and `with` were silent no-ops
+  or returned an empty array (#8096).** #8090 fixed `fill`, ranged `fill`,
+  `reverse` and `copyWithin` by asking `array::header::typed_array_receiver()`
+  BEFORE the `clean_arr_ptr` funnel, and named these six helpers as carrying the
+  identical dead delegation:
+
+  | helper | file | clean at | delegation at |
+  |---|---|---|---|
+  | `js_array_to_reversed` | `array/immutable.rs:32` | 33 | 37 |
+  | `js_array_to_sorted_default` | `array/immutable.rs:59` | 60 | 61 |
+  | `js_array_to_sorted_with_comparator` | `array/immutable.rs:96` | 105 | 106 |
+  | `js_array_with` | `array/immutable.rs:232` | 242 | 246 |
+  | `js_array_sort_default` | `array/sort.rs:542` | 551 | 561 |
+  | `js_array_sort_with_comparator` | `array/sort.rs:635` | 651 | 658 |
+
+  Codegen routes statically-typed typed-array receivers through the generic
+  `js_array_*` helpers on purpose (#3148 / #654 — `is_array_expr` answers `true`
+  for `Int32Array` &co.) on the contract that each helper re-dispatches on
+  `lookup_typed_array_kind`. `clean_arr_ptr` rejects those receivers, and must:
+  since #7574 it returns null for every tracked non-`GC_TYPE_ARRAY` object,
+  because a `TypedArrayHeader`'s raw per-kind storage is not boxed-f64
+  `ArrayHeader` slots. So every delegation written below the clean was
+  unreachable.
+
+  Two distinct silent failures followed. In-place `sort` returned the null'd
+  pointer having sorted nothing; `toReversed` / `toSorted` / `with` returned
+  `js_array_alloc(0)` — an EMPTY PLAIN ARRAY, so `.constructor.name` was
+  `"Array"` and `JSON.stringify(Array.from(x))` was `[]`.
+
+  Measured against the pinned node oracle (`v26.5.1`, `.node-version`), both
+  arms exit 0:
+
+  ```
+                          node                perry before
+  i32 sort default    1 2 9 10            10 9 2 1        (no-op)
+  f64 sort cmp desc   10.5 9.25 2 1       1 10.5 2 9.25   (no-op)
+  i32 toSorted        1 2 9 10            undefined x4    (empty array)
+  f64 toReversed      4.5 3.5 2.5 1.5     undefined x4
+  i32 with            1 99 3 4            undefined x4
+  i8  with wraps      -56                 undefined
+  ctor names          Int32Array x3       Array x3
+  ```
+
+  The plain-`Array` controls in the same program are correct in BOTH arms
+  (`plain sort default: 1 10 2 9`, the ToString ordering), which is what makes
+  the defect typed-array-specific rather than a general mutator break.
+
+  `sort` and `toSorted` have to reach the typed helper for the ORDER as well as
+  the layout: `%TypedArray%.prototype.sort` compares NUMERICALLY (§23.2.3.29
+  CompareTypedArrayElements) where `Array.prototype.sort` compares by `ToString`.
+  `[10, 9, 2, 1]` separates all three possible answers — numeric `1, 2, 9, 10`,
+  string `1, 10, 2, 9`, no-op `10, 9, 2, 1` — and is what the new tests feed.
+  The comparator cases feed `[1, 10, 2, 9]` for the same reason: an
+  already-descending input would let a no-op pass a descending-sort assertion.
+
+- **`Uint8Array` and `Buffer` `toReversed` / `toSorted` answered an empty array
+  on every dispatch path (#8096, second receiver shape).** Perry's
+  `new Uint8Array([…])` is not a registry `TypedArrayHeader` at all —
+  `buffer::js_uint8array_new` returns a `BufferHeader`, registered as a buffer
+  and marked `mark_as_uint8array`. So `typed_array_receiver`, which is
+  registry-backed, legitimately answers `None` for the most common typed array
+  in the language, while `clean_arr_ptr` still rejects it as a tracked
+  non-array. Verified directly on the real constructor path:
+
+  ```
+  addr=0x20000b00008   is_registered_buffer=true   lookup_typed_array_kind=None
+  gc_header_obj_type=Some(10)   clean_arr_ptr_null=true
+  js_array_to_reversed(...).length = 0
+  ```
+
+  Most `Array.prototype` entry points never see this: `sort` / `with` /
+  `reverse` / `fill` on that shape resolve through the dynamic method
+  dispatcher, and `copyWithin` grew its own Buffer arm in #8090. But
+  `toReversed` and `toSorted` fold unconditionally in HIR
+  (`lower/expr_call/local_array_methods.rs` — no receiver-type guard), and the
+  dynamic tower's own `toReversed` / `toSorted` arms
+  (`object/native_call_method/handle_methods.rs`) call straight back into these
+  same helpers, so those two were wrong for `Uint8Array` and `Buffer` on every
+  dispatch path, static and dynamic:
+
+  ```
+  ann u8 toReversed:  node 4 9 2 10 1 [object Uint8Array]
+                     perry 0 undefined undefined undefined undefined [object Array]
+  ```
+
+  New `buffer_receiver_as_uint8_typed_array` (`array/header.rs`) resolves that
+  shape into a fresh `KIND_UINT8` %TypedArray% copy. It is sound only for the
+  IMMUTABLE methods — a copy-based in-place `sort` would sort the copy and leave
+  the receiver untouched, a different wrong answer — so `toReversed`,
+  `toSorted` and `with` use it and `js_array_sort_*` deliberately does not. It
+  declines `ArrayBuffer` / `SharedArrayBuffer` / `DataView` receivers, which
+  have no `%TypedArray%.prototype` in node.
+
+  Validation: five probes byte-identical to node `v26.5.1`, all exit 0 in both
+  arms. 12 tests added to
+  `crates/perry-runtime/src/array/typed_array_receiver_tests.rs` (19 in the file
+  with #8090's), sabotage-verified in two stages after real rebuilds —
+  reverting `immutable.rs` + `sort.rs` turns 6 of them red, neutering
+  `buffer_receiver_as_uint8_typed_array` turns the other 3 red. The 3 that stay
+  green in both arms are the intended controls: the plain-`Array` control, the
+  `ArrayBuffer` / `DataView` decline, and the registry precondition the Buffer
+  arm exists for (#8090's `clean_arr_ptr_still_rejects_a_typed_array_receiver`
+  keeps #7574 pinned alongside them). `cargo test -p perry-runtime --lib`: 2380 passed / 0 failed / 4 ignored
+  (`main`: 2361 / 0 / 4). `perry-codegen`: 1434 passed / 11 failed, identical to
+  the `main` baseline (#8092).
+
+  `js_array_to_spliced` needs no equivalent — `%TypedArray%.prototype` has no
+  `toSpliced`.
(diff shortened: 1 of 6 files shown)


proggeramlug:
Verified the central finding, and it exposes a gap in a review of mine.

**`typed_array_receiver` has no Buffer arm.** Confirmed on `main`:

```rust
pub(crate) fn typed_array_receiver(arr: *mut ArrayHeader) -> Option<*mut TypedArrayHeader> {
    let addr = array_receiver_addr(arr);
    if addr == 0 { return None; }
    crate::typedarray::lookup_typed_array_kind(addr)          // <- registry ONLY
        .map(|_| addr as *mut TypedArrayHeader)
}
```

Meanwhile the typedarray module pairs the two predicates wherever it means "is this typed-array-ish": `mod.rs:403` is `lookup_typed_array_kind(addr).is_some() || crate::buffer::is_registered_buffer(addr)`, and #8109's `classify_element_read_receiver` falls back to `is_registered_buffer` at `mod.rs:510`. So the read side already knew; the write side's helper did not.

Since `new Uint8Array([…])` yields a `BufferHeader`, `typed_array_receiver` answers `None` for **the most common typed array in the language**. Your account of why #8090 still looked correct holds up: `sort`/`with`/`reverse`/`fill` escape through the dynamic dispatcher, and `copyWithin` got its own Buffer arm — while `toReversed`/`toSorted` fold unconditionally in HIR and so had no escape.

**This is a gap in my #8090 review, not just in #8090.** I built a reproducer for that PR and chose `Uint16Array` and `Float64Array` — deliberately picking a second element width, and never picking the one shape with a different *header*. Element width was the axis I varied; representation was the axis that mattered. A `Uint8Array` line would have caught it there.

Restricting `buffer_receiver_as_uint8_typed_array` to the immutable methods, and explicitly keeping `js_array_sort_*` off it because a copy-based in-place sort would be a *differently* wrong answer, is the right call — that distinction is easy to lose and expensive to rediscover.

## The probe trap is worth recording

> `[10,9,2,1].sort((x,y) => y-x)` expects `10 9 2 1` — exactly what a no-op gives

That is a test that cannot fail, and it is a nastier shape than the ones caught earlier tonight because the expected output is *correct* — it just happens to equal the broken output. My brief warned you about typed-array default ordering being numeric rather than ToString, and that warning would not have saved you here; a descending comparator over already-descending input is invisible to it. Choosing `[10,9,2,1]` so that no-op, ToString order and numeric order are three distinct answers is the fix, and it belongs in the PR body.

Static gates I ran independently on the landing tree: 34 checks, all green.

proggeramlug:
One more thing I want on the record, because it is the part that would have been easiest to get wrong.

`buffer_receiver_as_uint8_typed_array` copies the bytes out **before** allocating:

```rust
// Copy the bytes out BEFORE allocating: `typed_array_alloc` can collect,
// and a raw payload pointer read across it is exactly the borrowed-heap-
// slice shape rooting cannot fix.
```

That is the one GC hazard in this codebase that rooting genuinely **cannot** repair — a borrowed slice into the heap held across a collection point is not a missing root, so no handle or scope fixes it. Reading `buffer_data(buf)` into a slice and then calling `typed_array_alloc` would have produced an intermittent from-space read that survives every rooting audit, every `--unrooted-allocas` scan, and the dominance checker, because there is nothing for any of them to find.

Recognising it unprompted and structuring the function around it is the difference between a fix and a latent one.

Two other judgements worth affirming:

- **Excluding `ArrayBuffer` / `SharedArrayBuffer` / `DataView`** is spec-correct rather than merely conservative — none carries `%TypedArray%.prototype`, so node throws `TypeError: … is not a function` rather than answering elements. Returning `None` reproduces that; returning a typed array would have invented a passing behaviour node does not have.
- **Restricting the helper to the immutable methods** and keeping `js_array_sort_*` off it. The answer is a copy, so an in-place mutator delegating to it would sort the copy and leave the receiver untouched — trading a no-op for a subtler wrong answer, which is worse because it looks like it works.

proggeramlug:
Landing-equivalent validation clean — merged current `main` (`2b03b8660`, now including #8106/#8109/#8118) and re-ran:

```
merge with main                      CLEAN
cargo fmt --all -- --check           clean
check_file_size.sh                   clean
addr_class_inventory                 clean   <- the one I expected to bite, given array/header.rs's floor
gc_store_site_inventory              clean
raw_handle_debt                      clean
local_binding_type_audit             clean
gc_gate_wiring_check                 clean
gc_runtime_root_holders              clean
check_test_registration              clean
global_sink_isolation                clean
workspace_architecture --check       clean
cargo test -p perry-runtime --lib    2376 passed, 0 failed, 4 ignored   (19 typed-array receiver tests)
```

Merging. The two findings that make this more than a scope extension of #8090 are worth restating for the record:

**`typed_array_receiver` has no Buffer arm**, so it answers `None` for `new Uint8Array([…])` — the most common typed array in the language. That gap was in a helper I reviewed and merged, and my reproducer for #8090 varied *element width* (`Uint16Array`, `Float64Array`) while never varying *representation*. One `Uint8Array` line would have caught it there.

**The probe trap.** `[10,9,2,1].sort((x,y) => y-x)` expects `10 9 2 1`, which is exactly what a no-op produces. That is a harder shape than the other cannot-fail tests hit this session, because the expected output is genuinely correct — it just coincides with the broken one. Picking input where no-op, ToString order and numeric order are three distinct answers is the durable fix.

Issue #6088, opened 2026-07-06
runtime: typed-array out-of-bounds index returns 0 instead of undefined


proggeramlug:
## Symptom

Reading a typed array out of bounds returns `0` instead of `undefined`:

```ts
const t = new Uint8Array([5, 6, 7]);
console.log(t[9]);          // node: undefined   perry: 0
const a: any = t;
console.log(a[9]);          // node: undefined   perry: 0
```

Reproduces for both a statically-typed `Uint8Array` receiver and an `any`-typed
one (so it is not generator/CPS-specific). In-bounds reads are correct
(`t[0] === 5`).

## Notes

Spec: an integer-indexed exotic object returns `undefined` for a
CanonicalNumericIndexString that is out of `[0, length)`. The dynamic path
(`js_dyn_index_get` → `js_typed_array_index_get_dynamic`) and the inline
fast/slow typed-array get appear to fall through to a `0` default on OOB rather
than `undefined`.

Found tangentially while fixing #6067 (generator local **string** OOB index);
independent of that (string-only) fix. Low urgency — filing for tracking.

Issue #8116, opened 2026-08-14
typed_array_index_get_dynamic answers `undefined` when the header says typed array but the registry misses


proggeramlug:
## Summary

`typed_array_index_get_dynamic` returns `undefined` for a receiver that `classify_element_read_receiver` identifies as a typed array **by header** while `typed_array_addr_from_value` does not resolve it.

Surfaced during the #8109 review. **This is not a regression** — the pre-#8109 code for that whole branch was an unconditional `return f64::from_bits(TAG_UNDEFINED)`, so the case behaved identically before. #8109 routed the `Ordinary` arm to `js_dyn_index_get` and deliberately left `_` alone. Filing it separately so it is not lost, since #8109's classifier is exactly what makes the case nameable for the first time.

## The divergence

The two gates are not the same predicate:

- `classify_element_read_receiver` (`typedarray/mod.rs`) returns `TypedArray(addr)` on a `lookup_typed_array_kind` hit **or** on a `GC_TYPE_TYPED_ARRAY` / `GC_TYPE_NATIVE_TYPED_VIEW` header — the header deliberately wins on a registry miss, so a lookup failure can cost the diversion but never the element read.
- `typed_array_addr_from_value` (`typedarray_props.rs:124`) gates on `typed_array_owner_kind(addr).is_some()`.

Where those disagree — header says typed array, `typed_array_owner_kind` says no — the match in `typed_array_index_get_dynamic` falls to `_` and answers `undefined`.

## What needs establishing first

**Is it reachable?** I did not settle this. If every real typed array is registered such that `typed_array_owner_kind` always resolves, the arm is dead and the right fix is a debug assertion plus a comment, not a code path. If the two can genuinely disagree — subarray/slice views and native-arena views are the obvious candidates, given #5989's buffer-backed `Uint8Array` precedent immediately above — then it is a live `undefined`-instead-of-element bug and deserves the same treatment #8109 gave the plain-array case.

Start there rather than with a fix: a test that constructs the disagreeing receiver is worth more than a patch that might be guarding nothing.

## If reachable

Resolve `TypedArray(addr)` into an owner and run the common typed-array key logic, instead of falling through. Include a regression test for the registry-miss path specifically — and confirm it goes red without the fix, since a test for an unreachable arm passes either way.

Refs #8109, #8100. Related: #8111 (the `Uint8Array`-specialized twin, measured as pre-existing and untouched by #8109).

PR #8120, opened 2026-08-14
fix(typedarray): dispatch, not drop, when a Uint8Array-specialized element helper's receiver is not one

### LINKED ISSUE #8111: Reassigning a `new Uint8Array` binding drops element writes and reads `undefined` (the Uint8-specialized twin of #8100)

proggeramlug:
Closes #8111.

## What was wrong

The `Uint8Array`-specialized twin of #8100. `js_uint8array_get`,
`js_uint8array_index_get_value` and `js_uint8array_set`
(`crates/perry-runtime/src/typedarray/access.rs`) are a **separate emission
path** from the two helpers #8109 fixed: codegen picks them from
`is_uint8array_receiver` (`perry-codegen/src/expr/index_{get,set}.rs`), which
reads `receiver_class_name` rather than the `local_type_hint` predicate #8100 is
about — but it fires for a reassigned `Uint8Array` local just the same.

Each had a three-way shape (registered typed array of the right kind /
registered buffer / fall off the end) and **two** of those arms answered for a
receiver that is perfectly readable:

* the **trailing** arm — a plain array, object or string — answered `0` /
  `undefined` / dropped the store;
* the **wrong-KIND** arm — a registered typed array that is not `Uint8Array` /
  `Uint8ClampedArray` — did the same.

The store half is the damaging one: `Q[0] = 5` left no trace at all.

Unlike #8100, these helpers were memory-safe: they validated before
dereferencing. They just answered wrongly, silently, in the shipped default.

## The fix

The read arms delegate to `js_typed_array_get`, which owns #8109's
`classify_element_read_receiver` dispatch — so this path **inherits** it rather
than growing a third classifier, including the property that a
`GC_TYPE_TYPED_ARRAY` / `GC_TYPE_NATIVE_TYPED_VIEW` header still wins on a
registry miss. The store arm asks the same classifier directly, because
`js_dyn_index_set` has its own return-value contract and the value arrives as an
`i32`.

`js_uint8array_get`'s ABI is a byte-typed `i32`, so a recovered value there goes
through `ToNumber` (`jsvalue_to_f64`) rather than having its NaN-box bits
reinterpreted, and `undefined` still collapses to the `0` byte sentinel (#6088).

## Deviation from the issue's suggested fix

#8111 says the wrong-KIND arm "must stay as it is". That is right about what the
classifier *answers* but leaves a second silent wrong answer in place. Measured,
node `v26.5.1`:

```ts
let W: Uint8Array = new Uint8Array(2);
W = new Int32Array([11, 12]) as any;
W[0] = 77;
console.log(W[0], W[1], W.length);
```

```
node          77 12 2
pre-fix       undefined undefined 2      <- read AND store both lost
```

`js_typed_array_get` / `js_typed_array_set` are kind-generic — they read
`(*ta).kind` — so serving a non-uint8 registered typed array is neither unsafe
nor a layout gamble, and `js_typed_array_set` even performs the spec's
`ToBigInt` `TypeError` for a Number written into a bigint view. Those arms are
removed. `Uint8Array` / `Uint8ClampedArray` behaviour is unchanged and pinned by
two control tests (`300 -> 44` wrapping, `300 -> 255` clamping).

## Residual, documented in-code, NOT introduced here

`js_uint8array_set`'s value parameter is `i32`, and codegen narrows the value
with `fptosi` at the call site (`perry-codegen/src/expr/arrays_finds.rs`,
`Uint8ArraySet` arm) because the declared receiver is a byte view. So a
fractional value written through a reassigned binding arrives already truncated:
`Q[0] = 1.5` stores `1` where node stores `1.5`. Closing that needs an
f64-valued entry point plus a codegen change, not a runtime-side dispatch. Every
integer store — the issue's reproducer and every realistic byte-view use — is
exact.

## Evidence

Three-way, exit codes checked on every run: node `v26.5.1` (the `.node-version`
pin), pre-fix perry, post-fix perry. Both perry arms `--profile perry-dev` from
`a99732458`, `--no-auto-optimize --no-cache`, `PERRY_RUNTIME_DIR` pinned to a
freshly built archive pair (`grep -c "Compiling perry-runtime v"` = 1, archive
mtime after the edit).

```
                     node             perry before
store:               [5,10]           [9,10]                 <- write DROPPED
dyn:                 10               undefined
read:                5 10 2 10        undefined undefined 2 10
obj store+read:      42 8             undefined undefined
string read:         h i              undefined undefined
wrong-kind:          77 12 2          undefined undefined 2
```

**Controls, correct in BOTH arms** — this is what makes it typed-array-specific
rather than a general mutator break:

```
real u8:             3 250 2
buf:                 200 2
plain arr:           5 10 [5,10]
plain obj:           5 10
clamped:             255 2
```

`Q.length` and `Q.at(1)` are already right pre-fix, so the binding really does
hold the plain array — only the specialized element accessors were wrong.

Post-fix the probe is **byte-identical to node**, both exit 0.

## Tests

7 in `crates/perry-runtime/src/typedarray/element_read_receiver_tests.rs`, each
asserting the **recovered value** rather than "did not panic".

Sabotage-verified after a real rebuild: reverting `access.rs` to `origin/main`
turns **4** of them red —
`js_uint8array_index_get_value_reads_a_plain_array_receiver`,
`js_uint8array_set_stores_into_a_plain_array_receiver`,
`js_uint8array_get_reads_a_plain_array_receiver_as_a_byte`,
`uint8_helpers_read_and_write_a_wrong_kind_typed_array`.

The 3 that stay green in both arms are the intended controls: the real
`Uint8Array`, the `Uint8ClampedArray`, and the masked-away non-pointer receiver.
The existing `clean_ta_ptr_does_not_validate_a_plain_array_receiver` keeps the
precondition pinned.

## Suites

* `cargo test -p perry-runtime --lib` → **2380 passed, 0 failed, 4 ignored**
  (main: 2361 / 0 / 4).
* `cargo test -p perry-codegen --no-fail-fast` → 27 suites, 1434 passed,
  **11 failed** — identical to the `main` baseline (pre-existing, #8092).
* `cargo fmt --all -- --check`, `check_file_size.sh` and the 20 `lint` scripts
  enumerated from `test.yml` all pass. No new bare-address comparison, so
  `addr_class_inventory.py` is unmoved.

Refs #8100, #8109, #7494, #6088.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect reads and silently ignored writes for reassigned `Uint8Array` values.
  * Improved fallback behavior for regular arrays, objects, strings, buffers, and other typed arrays.
  * Preserved correct bounds handling, byte wrapping, clamping, and bigint type errors.
  * Maintained expected `undefined` and numeric sentinel behavior.

* **Tests**
  * Added regression coverage for reassigned arrays, incompatible typed arrays, non-pointer values, and byte-array edge cases.

### DIFF
--- a/changelog.d/8120-uint8array-specialized-dispatch.md
+++ b/changelog.d/8120-uint8array-specialized-dispatch.md
@@ -1,0 +1,74 @@
+### Fixed
+
+- **Reassigning a `Uint8Array` binding dropped element writes and read
+  `undefined` (#8111).** The `Uint8Array`-specialized twin of #8100.
+  `js_uint8array_get`, `js_uint8array_index_get_value` and `js_uint8array_set`
+  (`typedarray/access.rs`) are a SEPARATE emission path from the two helpers
+  #8109 fixed: codegen picks them from `is_uint8array_receiver`
+  (`perry-codegen/src/expr/index_{get,set}.rs`), which reads
+  `receiver_class_name` rather than the `local_type_hint` predicate #8100 is
+  about — but it fires for a reassigned `Uint8Array` local just the same.
+
+  Each had a three-way shape (registered typed array of the right kind /
+  registered buffer / fall off the end) and TWO of those arms answered for a
+  receiver that is perfectly readable: the trailing arm (a plain array, object
+  or string) and the wrong-KIND arm (a registered typed array that is not
+  `Uint8Array` / `Uint8ClampedArray`). Reads answered `0` / `undefined`; the
+  store was dropped with no trace at all. Unlike #8100 these helpers were
+  memory-SAFE — they validated before dereferencing — they just answered
+  wrongly, silently, in the shipped default configuration.
+
+  Measured against the pinned node oracle (`v26.5.1`, `.node-version`), both
+  arms exit 0:
+
+  ```
+                       node             perry before
+  store:               [5,10]           [9,10]                 <- write DROPPED
+  dyn:                 10               undefined
+  read:                5 10 2 10        undefined undefined 2 10
+  obj store+read:      42 8             undefined undefined
+  string read:         h i              undefined undefined
+  wrong-kind:          77 12 2          undefined undefined 2
+  ```
+
+  Correct in BOTH arms: a real `Uint8Array` (`3 250 2`), a `Buffer`
+  (`200 2`), a plain array (`5 10`), a plain object (`5 10`), a
+  `Uint8ClampedArray` (`255 2`). `Q.length` and `Q.at(1)` were already right, so
+  the binding really did hold the plain array — only the specialized element
+  accessors were wrong.
+
+  The read arms now delegate to `js_typed_array_get`, which owns #8109's
+  `classify_element_read_receiver` dispatch, so this path INHERITS it rather
+  than growing a third classifier — including the property that a
+  `GC_TYPE_TYPED_ARRAY` / `GC_TYPE_NATIVE_TYPED_VIEW` header still wins on a
+  registry miss, so a lookup failure can cost the diversion but never the
+  element read. The store arm asks the same classifier directly, because
+  `js_dyn_index_set` has its own return-value contract and the value arrives as
+  an `i32`. `js_uint8array_get`'s ABI is a byte-typed `i32`, so a recovered
+  value there goes through `ToNumber` rather than having its NaN-box bits
+  reinterpreted, and `undefined` still collapses to the `0` byte sentinel
+  (#6088).
+
+  The wrong-KIND arms are REMOVED rather than kept as the issue suggested:
+  `js_typed_array_{get,set}` are kind-generic (they read `(*ta).kind`), node
+  reads and writes the real element there, and `js_typed_array_set` even
+  performs the spec's `ToBigInt` `TypeError` for a Number written into a bigint
+  view. `Uint8Array` / `Uint8ClampedArray` behaviour is unchanged and pinned by
+  two control tests (`300 -> 44` wrapping, `300 -> 255` clamping).
+
+  RESIDUAL, documented in-code and not introduced here: codegen narrows
+  `js_uint8array_set`'s value to `i32` at the call site (`fptosi`,
+  `perry-codegen/src/expr/arrays_finds.rs`) because the DECLARED receiver is a
+  byte view, so a fractional value written through a reassigned binding arrives
+  already truncated — `Q[0] = 1.5` stores `1` where node stores `1.5`. Closing
+  that needs an f64-valued entry point plus a codegen change, not a
+  runtime-side dispatch. Every integer store is exact.
+
+  Validation: the issue's probe is byte-identical to node post-fix. 7 unit tests
+  in `crates/perry-runtime/src/typedarray/element_read_receiver_tests.rs`, each
+  asserting the recovered VALUE rather than "did not panic";
+  sabotage-verified after a real rebuild — reverting `access.rs` turns 4 of them
+  red, and the 3 that stay green are the intended controls.
+  `cargo test -p perry-runtime --lib`: 2380 passed / 0 failed / 4 ignored
+  (`main`: 2361 / 0 / 4). `perry-codegen`: 1434 passed / 11 failed, identical to
+  the `main` baseline (#8092).
--- a/crates/perry-runtime/src/typedarray/access.rs
+++ b/crates/perry-runtime/src/typedarray/access.rs
@@ -570,24 +570,36 @@ pub extern "C" fn js_typed_array_copy_within(
     ta
 }
 
+/// `uint8array[i]` / `buffer[i]` read as a raw byte. See
+/// [`js_uint8array_index_get_value`] for the JS-value twin and for the #8111
+/// stale-static-hint dispatch both share.
 #[no_mangle]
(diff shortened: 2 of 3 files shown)


proggeramlug:
Landing-equivalent validation clean — merged current `main` (`d736e74b5`, now including #8119) and re-ran:

```
merge with main                     CLEAN
cargo fmt --all -- --check          clean
check_file_size.sh                  clean
addr_class_inventory                clean
gc_store_site_inventory             clean
raw_handle_debt                     clean
local_binding_type_audit            clean
gc_gate_wiring_check                clean
check_test_registration             clean
cargo test -p perry-runtime --lib   2383 passed, 0 failed, 4 ignored
```

Merging.

## On the deviation — I endorse it

The issue said to keep the wrong-KIND arm answering `undefined`. You removed it, and the evidence settles it:

```
W = new Int32Array([11,12]) as any;  W[0] = 77;
node:   77 12 2
perry:  undefined undefined 2
```

Node serves that case, so *refusing* it is the wrong behaviour, not the safe one. And since `js_typed_array_{get,set}` are kind-generic, serving it is neither unsafe nor a layout gamble — it is the same read the correct path would do. Pinning `Uint8`/`Clamped` with two controls (300→44, 300→255) is what makes that claim checkable rather than asserted, since clamping is precisely where a kind-generic path could quietly disagree with the spec.

Deviating from a stated design with a node transcript and a control is the right way to do it. It is the second time tonight an agent declined an instruction and was correct to — the other being the `gap_snapshot.json` entry that would have disarmed a gate.

## Residual, correctly scoped out

Codegen `fptosi`s the value at the `js_uint8array_set` call site, so `Q[0] = 1.5` through a reassigned binding stores `1`. That needs an f64 entry point plus a codegen change, which is a different PR — and reporting it rather than half-fixing it keeps this one revertable on its own.

## Reusing #8109's classifier rather than adding a third

Reads delegating to `js_typed_array_get` so they *inherit* `classify_element_read_receiver`, with only the store arm asking it directly, is the right structure. Three classifiers for one question is how the read and write sides drifted apart in the first place — which is exactly the divergence #8116 now tracks.

Issue #7354, opened 2026-08-04
Windows: native-root stack walker so PERRY_RS4GC=1 works there (#7173)


proggeramlug:
# Windows: a native-root stack walker so `PERRY_RS4GC=1` works there

You are in **Perry** (native TypeScript→machine-code compiler, Rust + LLVM). Read
`CLAUDE.md` first. You are on a **Windows machine**, which is the point: this task
was deliberately left undone because it cannot be written blind.

## Setup (verbatim)

```powershell
git clone https://github.com/PerryTS/perry.git
cd perry
git checkout -b feat/windows-gc-walker origin/main
```

You need: Rust (stable, `x86_64-pc-windows-msvc`), the MSVC build tools, LLVM
(`clang` and `opt` from the **same** install — see below), and the Node version
pinned in `.node-version` at the repo root. `perry doctor` will check the
toolchain.

## What already exists — do not rebuild it

Merged in #7351:

* **The COFF section.** `crates/perry-codegen/src/gc_map.rs` emits `.pgcmap`
  into a COFF object. The name is seven bytes on purpose: a PE image section
  header has an 8-byte name field, and longer names survive only in object files
  as a string-table offset the linker does not carry into the image. Do not
  rename it without changing `COFF_SECTION_NAME` and the runtime lookup together.
* **The PE lookup.** `loaded_stack_map_section()` in
  `crates/perry-runtime/src/gc/roots/stack_maps.rs`, `#[cfg(target_os =
  "windows")]`, finds that section in the running image via
  `GetModuleHandleW(NULL)` → `IMAGE_DOS_HEADER` → NT headers → section table. It
  compiles for `x86_64-pc-windows-msvc`; it has never run.

## What is missing, and why Windows is refused today

There is **no stack walker on Windows**. `_Unwind_Backtrace` / `_Unwind_GetIP` /
`_Unwind_GetGR` / `_Unwind_GetCFA` do not exist there, so the `unwind` module is
`#[cfg(any(target_vendor = "apple", target_os = "linux"))]` and Windows falls to
a stub that visits nothing.

A map with no walker is worse than no map: the collector would find **zero
roots** and free live objects, silently. So `compact_and_assemble` in
`gc_map.rs` currently **refuses** COFF targets with a message naming the missing
piece. **Removing that refusal is the last step of this task, not the first.**

## The contract your walker must satisfy

Implement, for `#[cfg(target_os = "windows")]`, a module exposing exactly what
the other two do:

```rust
[...]

Issue #74, opened 2026-04-17
Date.now() does not advance during tight CPU-bound loops


proggeramlug:
## Summary

On Perry 0.5.88, `Date.now()` returns an identical value before and after a 100M-iteration CPU-bound loop. Node and Bun advance correctly on the same machine and same script.

This breaks any micro-benchmark or per-operation timing that samples `Date.now()` in a tight loop, and means driver/bench users can't measure hot-path throughput from inside a Perry-compiled program.

## Repro

```ts
// perry-date-now-repro.ts
const t0 = Date.now();
let sum = 0;
for (let i = 0; i < 100_000_000; i++) {
    sum += 1;
}
const t1 = Date.now();

console.log('iterations: 100,000,000');
console.log('sum (sanity): ' + sum);
console.log('Date.now() delta: ' + (t1 - t0) + 'ms');
console.log('raw t0: ' + t0);
console.log('raw t1: ' + t1);
```

## Observed

| Runtime | delta | raw t0 | raw t1 |
| --- | --- | --- | --- |
| **Perry 0.5.88** | **0ms** | 1776450625708 | 1776450625708 |
| Bun | 53ms | 1776450626201 | 1776450626254 |
| Node | 57ms | 1776450626732 | 1776450626789 |

`t0` and `t1` are literally the same integer on Perry — so it's not a subtraction quirk, `Date.now()` itself is returning a cached/stale value inside the loop.

## Expected

`Date.now()` returns current wall-clock milliseconds on each call, advancing as real time passes — as it does on Bun / Node / V8 / JSC.

## Impact

- Breaks micro-benchmarks: 1000 SELECT-1 queries against Postgres reported as `total=1ms avg=1.0µs` on Perry, while Node reports ~63µs/query and Bun ~45µs/query for the same driver code.
- Makes any `measure()` helper that does `const t0 = Date.now(); work(); const elapsed = Date.now() - t0;` return 0 for any work under the sampling quantum.
- Masks real per-query cost in the `@perry/postgres` benchmark suite — every sample quantizes to 0.5ms or 0.0ms, which we initially mistook for a real performance floor.

## Guess at root cause

Looks like `Date.now()` is being CSE'd / hoisted out of the loop, or the runtime is reading a cached wall-clock value that is only refreshed on I/O / event-loop yield. A `volatile`-equivalent fence around the syscall (or simply not folding calls at the JIT/LLVM level) should fix it. Worth confirming whether the Rust-side implementation caches the result per tick.

[...]

PR #8123, opened 2026-08-15
fix(codegen): keep RS4GC off the inline-asm loop barrier

### LINKED ISSUE #8121: [Next.js/dylib] perry compile SIGBUSes in LLVM codegen of next/dist/compiled/jsonwebtoken (100/104 modules in)
Part of #8040. First failure found when actually driving a real production Next.js App Route through Perry.

## Summary

`perry compile` **crashes with SIGBUS (EXC_BAD_ACCESS / KERN_PROTECTION_FAILURE)** during LLVM codegen of `next/dist/compiled/jsonwebtoken/index.js`. The composed production App Route never reaches the runtime — the failure is at the **compile boundary**, before any request can be executed.

The dylib link blocker (#8064 / fixed by #8078) is genuinely behind us: codegen reached **100 of 104 modules** before dying.

## Environment

- Perry `506f4ab11a5d2a1b7a498e965c79d11e2fa63043` (`perry 0.5.1510`), `cargo build --release -p perry -p perry-runtime-static -p perry-stdlib-static`
- macOS arm64, in-process LLVM backend (LLVM 22.1.4)
- Fixture: `tests/release/packages/next-app-route` (Next.js `16.3.0`, React `19.2.4`), production webpack build

## Node oracle: PASSES

Both oracles are green, so the fixture and route implementation are correct:

- `next start` + `verify.mjs` → `PASS: 21 production App Route requests`
- `node perry-host.js` + `verify.mjs` → `PASS: 21 production App Route requests` (twice, cold + warm)

The `routeModule.handle` bypass guard was **armed and sabotage-verified**: removing only the wrapper install makes it fire 21/21 times; with the wrapper installed it fires 0 times. So the guard is a discriminating detector, not a vacuous one. Note its signal appears in the **host log**, not in `verify.mjs`'s exit code — the assertion throws inside a `.then()` that is caught and logged after the response is already sent.

## The crash

Last codegen activity before death:

```
[perry] codegen: modules finished 100/104 (96%; 119.0 min elapsed; ETA ~4.8 min)
[perry] codegen: node_modules_next_dist_compiled_jsonwebtoken_index_js: freezing 2 codegen units for worker threads
[perry] codegen: node_modules_next_dist_compiled_jsonwebtoken_index_js: froze 2/2 units (100%; 9.6s elapsed; ETA ~0.0s)
perry: module IR is 14.0 MB (> 6.0 MB), 468 functions (~31 KB/fn); compiling at -Os instead of -O3 ...
```

No error is printed — the process dies under signal. Crash report:

```
exception: EXC_BAD_ACCESS (SIGBUS), KERN_PROTECTION_FAILURE at 0x0000001b413c6d08
termination: SIGNAL code 10, "Bus error: 10"
faulting thread: 15 (an LLVM codegen worker), depth 24 frames
```

**This is a wild pointer dereference, not a truncated-mmap SIGBUS.** The crash report's own VM region info places the fault address in *reserved, never-mapped* address space:

```
0x1b413c6d08 is in 0x1000000000-0x7000000000
--->  GPU Carveout (reserved)   1000000000-7000000000  [384.0G] ---/--- SM=NUL  reserved VM address space (unallocated)
```

Nothing is or ever was mapped there, so this is a corrupted/garbage pointer being dereferenced — it rules out the "object file truncated by a full disk" explanation. The faulting stack is only 24 frames deep, which also rules out stack overflow. The process had **15 codegen worker threads** live under the in-process LLVM backend.

## Minimal reproduction

Reduced from 104 modules to **2**. Outside a Next project directory (important: Perry auto-discovers `.next/server` — "Next.js standalone: discovered 17 runtime module(s)" — so compiling inside the fixture pulls the whole 104-module graph in regardless of what you `require`):

```
mkdir /tmp/minjwt && cd /tmp/minjwt
cp <fixture>/node_modules/next/dist/compiled/jsonwebtoken/index.js ./jsonwebtoken-index.js
printf 'const jwt = require("./jsonwebtoken-index.js");\nconsole.log("jwt typeof:", typeof jwt);\n' > main.js
perry compile main.js --output-type dylib --no-auto-optimize --no-cache -o min.dylib
```

`jsonwebtoken-index.js` is 118 KB, sha256 `056c2dddba96faa35f4dc0740228a241990561290ff3bd5456c7c7dbc5a7a6b9` (from `next@16.3.0`). It lowers to a **14.5 MB IR module, 468 functions**, split into 2 codegen units.

This reproduction reaches the exact crash state and `sample`s into the same code region as the original crash. **I was not able to run it to completion** — see the compile-time blowup below — so I have confirmed the repro reaches the failing path but have **not** yet confirmed the SIGBUS is deterministic. That is the first thing to check.

## Secondary finding: codegen compile-time blowup

Independent of the crash, single LLVM codegen units on this graph take pathological time:

- `.next/server/chunks/430.js` **unit 2/6 took 6704.3 s (112 minutes)** while its five sibling units took 80–319 s each.
- The 2-module `jsonwebtoken` repro spent **>90 minutes** on unit 1/2 while unit 2/2 finished in 142 s.

Sampling shows it genuinely progressing through different LLVM code (the hot leaf moves between samples), so it is slow, not hung. Total wall time for the 104-module compile was ~2 h before it crashed. This alone makes the fixture impractical as a CI gate.

## Suggested next steps

1. Confirm determinism of the SIGBUS by running the 2-module repro to completion.
2. A/B `PERRY_CODEGEN_UNIT_JOBS=1` vs default. 15 concurrent workers under an in-process LLVM backend makes a codegen concurrency bug (e.g. a shared `LLVMContext`/`Module` across threads) a prime suspect for a wild pointer. I started this A/B but had to stop it on disk pressure.
3. Build `perry` with debug symbols — the release binary is stripped (431 symbols), so the 24-frame backtrace could not be symbolicated and the responsible pass is still unidentified.
4. Investigate why one unit of a 14 MB IR module costs 100× its siblings.

proggeramlug:
## The bug

Perry emits `call void asm sideeffect "", ""()` as the issue-#74 loop-preservation barrier. `rewrite-statepoints-for-gc` rewrites **every non-leaf call** in a `gc "statepoint-example"` function into a `gc.statepoint` — and for inline asm that means using the `InlineAsm` itself as the statepoint's callee operand. That is invalid IR:

```
Cannot take the address of an inline asm!
  %statepoint_token26 = call token (i64, i32, ptr, i32, i32, ...)
      @llvm.experimental.gc.statepoint.p0(i64 2882400000, i32 0,
      ptr elementtype(void ()) asm sideeffect "", "", i32 0, i32 0, i32 0, i32 0)
      [ "gc-live"(ptr addrspace(1) %.1181, ...) ]
LLVM ERROR: Broken module found, compilation aborted!
```

`inprocess::optimize_and_emit` verified the module **before** the rewrite but not after, so the broken module went straight to SelectionDAG, which dereferenced the bogus callee and killed the compiler with `SIGBUS` in `AArch64TargetLowering::LowerCall`. Symbolized:

```
#0 llvm::AArch64TargetLowering::LowerCall(...)
#1 llvm::TargetLowering::LowerCallTo(...)
#3 llvm::SelectionDAGBuilder::LowerAsSTATEPOINT(...)
#4 llvm::SelectionDAGBuilder::LowerStatepoint(...)
#5 llvm::SelectionDAGBuilder::visitIntrinsicCall(...)
```

This blocked #8040: compiling next@16.3.0's bundled `jsonwebtoken` died 100 modules into a 104-module production App Route.

## The fix

- Mark the barrier `"gc-leaf-function"` on the native and text emission paths and in generic `call_asm`. That is the documented way to tell RS4GC a call cannot trigger a collection — true by construction for a barrier that emits no instructions.
- **Verify after RS4GC**, not only before, so this class of bug is a clean compile error naming the pass instead of a crash that takes the compiler process down. Same reasoning as the neighbouring `rs4gc_funclet_refusal` (#7354).

## Tests

Three, split so no half can pass hollowly:

| test | fails if |
|---|---|
| `rs4gc_breaks_an_unmarked_inline_asm_barrier` | RS4GC stops breaking unmarked barriers, i.e. the fix became unnecessary and its sibling vacuous |
| `a_gc_leaf_inline_asm_barrier_survives_rs4gc` | the attribute stops working, or RS4GC rewrote nothing (asserts `gc.statepoint` is present first) |
| `perry_emits_the_loop_barrier_as_a_gc_leaf` | Perry stops emitting the attribute — the two above use hand-written IR and would stay green through a full revert of the fix |

An earlier version of the second test was **vacuous** and is worth recording: it built its fixture with `enable_shadow_frame`, so the function carried no `gc "statepoint-example"` strategy, RS4GC rewrote nothing (`statepoint count = 0`), and it passed with the fix reverted. The `gc.statepoint` assertion is what closes that hole.

## Validation on the real workload

Reproduction (~2.5 min on a quiet host; the ">90 min" in #8121 was build contention):

```
npm pack next@16.3.0
tar -xzf next-16.3.0.tgz package/dist/compiled/jsonwebtoken/index.js
cp package/dist/compiled/jsonwebtoken/index.js ./jsonwebtoken-index.js   # sha256 056c2ddd…a6b9
printf 'const jwt = require("./jsonwebtoken-index.js");\nconsole.log(typeof jwt);\n' > main.js
perry compile main.js --output-type dylib --no-auto-optimize --no-cache -o min.dylib
```

Must run **outside** a Next directory, or `.next/server` auto-discovery pulls in all 104 modules.

The barriers live in exactly one codegen unit — `grep -c "asm sideeffect"` gives **4** in unit 2/2 and **0** in unit 1/2 — and that is precisely the unit stock `opt` rejects. With this change, `LLVM unit 2/2 finished (228.4s)` and the compile has run **49+ minutes with no crash** against a 2:30 crash point.

## Why this is a draft

The crash is fixed and unit-tested, but the full #8034 104-module gate has **not** been re-run with `PERRY_RS4GC=1` yet — the remaining unit is slow for an unrelated reason (below). Promoting out of draft once that gate result is in.

## Adjacent, not fixed here

RS4GC expands this module **21x**: `perry_closure_jsonwebtoken_index_js__227` (the webpack module factory, emitted as one 4.93 MB function — 8,471 basic blocks, 8,044 calls, 13,369 `addrspace(1)` GC pointers) turns 19 MB of IR into 413 MB, with **21,774 statepoints and 1,502,654 `gc.relocate`s**. That is the compile-time blowup reported in #8121, now unmasked because the crash no longer preempts it. It deserves its own issue — reducing values live across safepoints, or not emitting an 8k-safepoint monolith, shrinks it everywhere.

Refs #8040. Fixes #8121.

### DIFF
--- a/changelog.d/8121-rs4gc-inline-asm-statepoint.md
+++ b/changelog.d/8121-rs4gc-inline-asm-statepoint.md
@@ -1,0 +1,9 @@
+`perry compile` no longer SIGBUSes on production Next.js bundles that contain a preserved loop.
+
+Perry emits `call void asm sideeffect "", ""()` as the issue-#74 loop-preservation barrier. `rewrite-statepoints-for-gc` rewrites every non-leaf call in a `gc "statepoint"` function into a `gc.statepoint`, and for inline asm that means using the `InlineAsm` itself as the statepoint's callee operand — invalid IR, rejected by the verifier with `Cannot take the address of an inline asm!`.
+
+`optimize_and_emit` verified the module *before* the rewrite but not after, so the broken module went straight to SelectionDAG, which dereferenced the bogus callee and killed the compiler with `SIGBUS` inside `AArch64TargetLowering::LowerCall`. Compiling next@16.3.0's bundled `jsonwebtoken` died this way 100 modules into a 104-module production App Route, blocking #8040.
+
+The barrier now carries `"gc-leaf-function"` on both the native and text emission paths, which is the documented way to tell RS4GC a call cannot trigger a collection — true by construction for a barrier that emits no instructions. `optimize_and_emit` additionally verifies *after* RS4GC, so this class of bug is a clean compile error naming the pass instead of a crash that takes the compiler process down.
+
+Fixes #8121. Refs #8040.
--- a/crates/perry-codegen/src/dialect/mod.rs
+++ b/crates/perry-codegen/src/dialect/mod.rs
@@ -861,9 +861,11 @@ impl<'ctx, 'm> FnReader<'ctx, 'm> {
         let ptr =
             self.ctx
                 .create_inline_asm(void_fn, asm, constraints, sideeffect, false, None, false);
-        self.builder
+        let site = self
+            .builder
             .build_indirect_call(void_fn, ptr, &[], "")
             .map_err(be)?;
+        mark_gc_leaf(self.ctx, site);
         Ok(())
     }
 
@@ -1542,9 +1544,11 @@ impl<'ctx, 'm> FnReader<'ctx, 'm> {
                     None,
                     false,
                 );
-                self.builder
+                let site = self
+                    .builder
                     .build_indirect_call(void_fn, ptr, &[], "")
                     .map_err(be)?;
+                mark_gc_leaf(self.ctx, site);
                 Ok(())
             }
             I::Br { label } => {
@@ -1817,4 +1821,29 @@ fn apply_flags(inst: Option<InstructionValue<'_>>, flags: &[&str]) {
     if fmf != 0 {
         unsafe { llvm_sys::core::LLVMSetFastMathFlags(inst.as_value_ref(), fmf) };
     }
+}
+
+/// #8121: tell `rewrite-statepoints-for-gc` to leave an inline-asm call alone.
+///
+/// Perry emits `call void asm sideeffect "", ""()` as the issue-#74 loop
+/// preservation barrier. RS4GC rewrites every non-leaf call in a `gc
+/// "statepoint"` function into a `gc.statepoint` whose callee operand is the
+/// original callee — and for inline asm that means taking the address of an
+/// `InlineAsm`, which is not a value. The result fails the verifier with
+/// "Cannot take the address of an inline asm!", and because production only
+/// verifies BEFORE the rewrite, the broken module reached SelectionDAG and
+/// SIGBUS'd in `AArch64TargetLowering::LowerCall` while lowering the bogus
+/// statepoint callee.
+///
+/// `"gc-leaf-function"` is the documented way to say "this call cannot trigger
+/// a collection", which is exactly true of an empty barrier: it emits no
+/// instructions and cannot call back into the runtime.
+fn mark_gc_leaf<'ctx>(
+    ctx: &'ctx inkwell::context::Context,
+    site: inkwell::values::CallSiteValue<'ctx>,
+) {
+    site.add_attribute(
+        inkwell::attributes::AttributeLoc::Function,
+        ctx.create_string_attribute("gc-leaf-function", ""),
+    );
 }
--- a/crates/perry-codegen/src/inprocess.rs
+++ b/crates/perry-codegen/src/inprocess.rs
@@ -412,6 +412,20 @@ fn optimize_and_emit(
                     e.to_string()
                 )
             })?;
+        // #8121: verify AFTER the rewrite, not only before it. RS4GC can turn
+        // a module the verifier accepted into one it rejects (it rewrote an
+        // inline-asm barrier into a statepoint whose callee is an InlineAsm).
+        // Production previously verified only the input, so the broken module
+        // went straight to SelectionDAG and took the whole compiler down with
+        // a SIGBUS instead of reporting anything. A crash inside this process
+        // is exactly what the funclet refusal above exists to avoid.
+        module.verify().map_err(|e| {
+            anyhow!(
+                "rewrite-statepoints-for-gc produced a module the LLVM verifier \
+                 rejects; refusing to hand it to codegen (#8121):\n{}",
+                e.to_string()
+            )
+        })?;
     }
 
     let pipeline = match opt {
--- a/crates/perry-codegen/src/inprocess.rs
+++ b/crates/perry-codegen/src/inprocess.rs
@@ -455,6 +455,70 @@ fn optimize_and_emit(
 mod tests {
     use super::*;
 
+    /// #8121: Perry emits `call void asm sideeffect "", ""()` as the issue-#74
+    /// loop-preservation barrier. RS4GC rewrites every non-leaf call in a
+    /// `gc "statepoint-example"` function into a `gc.statepoint`, and for inline
+    /// asm that means using the `InlineAsm` itself as the callee operand —
+    /// invalid IR ("Cannot take the address of an inline asm!"). Production
+    /// verified only BEFORE the rewrite, so the broken module reached
+    /// SelectionDAG and killed the compiler with a SIGBUS in
+    /// `AArch64TargetLowering::LowerCall`.
+    ///
+    /// `%p` stays live across `@may_collect`, so RS4GC has real work to do and
+    /// a fixture that rewrote nothing cannot pass either arm silently.
+    fn asm_barrier_fixture(gc_leaf: bool) -> String {
+        let barrier_attr = if gc_leaf { " #5" } else { "" };
+        format!(
+            "declare i64 @may_collect()\n\
+             \n\
+             define ptr addrspace(1) @barrier_fn(ptr addrspace(1) %p) gc \"statepoint-example\" {{\n\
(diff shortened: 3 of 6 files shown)


proggeramlug:
## Validation upgraded: a *completed* compile with `PERRY_RS4GC=1`

Earlier the evidence was "runs far past the crash point without dying". It is now a finished artifact. Same input (`next@16.3.0`'s bundled `jsonwebtoken`), same build of this branch, RS4GC **enabled**:

| arm | result |
|---|---|
| `PERRY_LL_SIZE_OPT=0` (forced `-O0`) | **completed, 2m46.5s real / 3m29.5s user, 40 MB dylib** |
| default (`-Os`) | still running at 60+ min, no crash |

Before this branch the same command SIGBUS'd in ~2m30s. So the crash is fixed, and the compile genuinely finishes rather than merely surviving longer.

## This also isolates the compile-time blowup

The two arms differ only in the opt level applied to the RS4GC-expanded IR, so the >20x gap is the `-Os` pipeline on statepoint output — not RS4GC itself, and not the fix.

That points at a threshold that is arguably mistuned for this shape. `oversized_opt_flag` sends a unit to `-O0` when its widest function exceeds `ll_o0_threshold_bytes()` (6 MB). The offending function here — `perry_closure_jsonwebtoken_index_js__227`, the webpack module factory Perry emits as a single body — is **4.93 MB**, just under, so it takes `-Os`. Its profile is 8,471 basic blocks, 8,044 calls, 13,369 `addrspace(1)` GC pointers, which RS4GC turns into 21,774 statepoints and 1,502,654 `gc.relocate`s (19 MB -> 413 MB of IR).

The sibling constant `ll_size_opt_max_fn_bytes()` (256 KB) is the one documented as "the average IR bytes-per-function below which we size-optimize", and the comment there explicitly describes the pathological case as "megabytes-per-function". A 4.93 MB function is that case.

I have **not** changed the threshold in this PR. Retuning it is a real tradeoff — any unit containing one oversized function would drop to `-O0` wholesale, including its many ordinary functions, and `-Os` is documented as emitting 30-50% less `__text`. That deserves its own change with a size measurement behind it, not a drive-by edit inside a crash fix.

Flagging it here because it has a direct consequence for #8040: at hours-per-compile the #8034 fixture cannot serve as a CI gate regardless of correctness, and the definition of done there needs ten cold starts plus a 100-iteration 20-way concurrent run.

proggeramlug:
Closing as superseded — `main` already carries this fix, landed independently through #8082's sibling work while this branch was in validation. I checked before rebasing rather than force-pushing over it.

`origin/main` has all three pieces, and two of them in a better form than this PR:

- **native path** — `dialect/mod.rs` adds `"gc-leaf-function"` to the inline-asm call site.
- **text path** — `inst.rs` emits `call void asm sideeffect "", ""() "gc-leaf-function"` as an inline string attribute. Cleaner than this PR, which introduced an `attributes #5` group and had to emit it conditionally from `module.rs`.
- **post-RS4GC verify** — `inprocess.rs` verifies the rewritten module with a more precise message than mine ("this is a Perry codegen bug — the input shape must be exempted or fixed").

It also has something this PR does not: `demote_relocation_bloated_functions`, which re-checks opt tier against POST-rewrite sizes and opts just the exploded functions out of `-Os`. That directly addresses the relocation fan-out measured in #8132.

The independent diagnosis agreed on every particular, including the numbers — that comment records "one 51k-line minified-bundle closure grew 40x to 2.1M instructions, and a single `-Os` function pass then ran for over an hour", against my measurement of 19 MB → 413 MB of IR with 21,774 statepoints and 1,502,654 `gc.relocate`s on the `jsonwebtoken` unit.

Nothing here is worth salvaging on top. The tests in this branch (`rs4gc_breaks_an_unmarked_inline_asm_barrier`, `a_gc_leaf_inline_asm_barrier_survives_rs4gc`, `perry_emits_the_loop_barrier_as_a_gc_leaf`) may still be useful if `main`'s version does not already pin the behaviour both ways — the first of those asserts RS4GC *does* break an unmarked barrier, so it fails if the exemption ever stops being necessary. Happy to re-open just the test half if that coverage is missing.

PR #8126, opened 2026-08-15
test(gc): assert the poll guard's CFG shape, not its text order

proggeramlug:
`a_surviving_poll_is_guarded_by_the_arming_word` fails on main, but the lowering is correct — every `load volatile @PERRY_GC_POLL_ARMED` is followed by `icmp` + `br i1` into a `gcpoll.N` block holding the call.

The test's *proxy* for that property broke. It searched the whole remainder of the IR after each load and required the first `br i1` to precede the first poll call — silently assuming each poll block is printed next to its guard. It isn't: guards sit inline in `for.cond`/`for.body`/`for.update`, while the `gcpoll.N` blocks are emitted together after the loop. So the first two segments contain no call, `find(POLL)` returned `None`, and the `matches!` arm rejected a correctly-guarded poll.

Now the search is bounded at the load's own basic block: that block must end in a conditional branch and must not contain the call. Same property, no dependence on block print order.

**Teeth preserved:** all six negative controls (coercible bound, coercible accumulator, call in body, object literal in body, module-global accumulator, string concat) still emit a poll and still run through this check — they pass.

Note this does **not** make `e2e-scoped` green on its own: `loop_safepoint_purity` has one other failure (`proven_numeric_counted_loop_emits_no_back_edge_poll`) which is a genuine regression — the loop lowers to `js_dynamic_string_or_number_add`/`js_rel_lt`, so the numeric proof is not applying — and that is tracked separately from this test-shape fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compiler validation for loop safepoints by checking control-flow behavior directly, increasing reliability across different generated code layouts.
  * Preserved existing negative-case coverage to help prevent regressions in safepoint handling.

* **Tests**
  * Updated guarded-poll validation to better reflect actual control-flow structure and avoid relying on generated output ordering.

### DIFF
--- a/changelog.d/8126-loop-poll-guard-cfg-assert.md
+++ b/changelog.d/8126-loop-poll-guard-cfg-assert.md
@@ -1,0 +1,15 @@
+`loop_safepoint_purity`'s guard assertion no longer depends on block print
+order. `a_surviving_poll_is_guarded_by_the_arming_word` failed on main while
+the lowering was correct: every `load volatile @PERRY_GC_POLL_ARMED` is
+followed by `icmp` and `br i1` into a `gcpoll.N` block holding the call.
+
+The test searched the whole remainder of the IR after each load and required
+the first `br i1` to precede the first poll call, which assumed each poll block
+is printed next to its guard. It is not — the guards sit inline in `for.cond` /
+`for.body` / `for.update` while the `gcpoll.N` blocks are emitted together
+after the loop — so the leading segments contained no call, `find` returned
+`None`, and a correctly-guarded poll was rejected.
+
+The check is now bounded at the load's own basic block: that block must end in
+a conditional branch and must not contain the call. The negative controls are
+unchanged and still pass, so the assertion keeps its teeth.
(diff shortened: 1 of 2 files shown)


proggeramlug:
Verified on a landing-equivalent tree (merged current `main` `83b6b8c69`):

```
test a_surviving_poll_is_guarded_by_the_arming_word ... ok      <- was failing
test result: FAILED. 7 passed; 1 failed
             ^ proven_numeric_counted_loop_emits_no_back_edge_poll, a DIFFERENT
               pre-existing failure this PR does not touch
```

So this fixes exactly one of the eleven and leaves the other alone, which is the right blast radius.

**The important consequence: one of the eleven was never a real failure.** I have been using "27 suites / 1434 passed / 11 failed" as the codegen baseline all session, and treating those eleven as genuine pre-existing breakage. At least one was a broken test *proxy* — the lowering was correct the whole time. Worth re-examining the other ten in that light before anyone bisects them.

The diagnosis is right and worth restating: the old check searched the whole remainder of the IR after each `load volatile @PERRY_GC_POLL_ARMED` and required the first `br i1` to precede the first poll call. That silently assumed each `gcpoll.N` block is printed next to its guard. It is not — guards sit inline in `for.cond`/`for.body`/`for.update` while the poll blocks are emitted together after the loop, so the first two segments contain no call at all and `find(POLL)` returned `None`.

Bounding the search at the load's own basic block asserts the actual CFG property — this block ends in a conditional branch and does not contain the call — with no dependence on print order. That is a strictly better assertion, not a weakened one.

Issue #8117, opened 2026-08-14
`conformance-smoke` fails 8/8 shards: 12 real regressions, 10 newly-visible node_fail transitions, 1 fix the snapshot hasn't accepted


proggeramlug:
## Summary

`conformance-smoke` is a **required** status context and it fails **8 of 8 shards** on every recent PR. I pulled all eight shard logs from #8090's completed run and tallied the actual gate output. It is not flake, and it is not one cause — there are **23 status changes** in four distinct categories.

CLAUDE.md warns that these shards are flaky and that a red shard should be re-run and A/B'd before being believed. **Eight of eight is not the flake signature**, which is why I went looking.

## The tally

```
 10  node_fail  -> parity_fail
  8  pass       -> parity_fail
  3  pass       -> crash
  1  pass       -> compile_fail
  1  parity_fail -> pass
```

### 10 × `node_fail -> parity_fail` — previously invisible, now exposed

```
test_gap_4510_enum_forward_ref          test_gap_moment_methods
test_gap_backoff_options                test_gap_prop_plan_cache_invalidation
test_gap_cron_cronjob                   test_gap_ratelimiter_memory
test_gap_dayjs_factory_arg              test_gap_slugify_options
test_gap_derived_param_props            test_gap_enum_in_function_body
```

This is exactly the #6364 pattern CLAUDE.md documents: when the oracle cannot run a test it exits non-zero, the harness classifies it `node_fail`, and **the test is silently dropped from the gate rather than going red**. These ten were in that hidden state and are now running — so these are not new breakage, they are newly *visible* breakage. The shard header says so itself: *"STATUS CHANGES — including tests the oracle stopped covering"*.

That makes them the least urgent of the four groups and arguably good news, but they must be triaged and recorded rather than left to hold the gate red.

### 12 × genuine regressions from `pass`

```
pass -> crash          test_gap_6386_dataview_concat_regex_fastpaths
                       test_gap_7238_i64_specialization_exactness
                       test_gap_buffer_own_props
pass -> compile_fail   test_gap_zlib_4917_level
pass -> parity_fail    test_gap_array_methods
                       test_gap_builtin_alias_construct_7524
                       test_gap_collection_foreach_member_receiver_thisarg
                       test_gap_dynamic_builtin_construct_dispatch
                       test_gap_new_globalthis_builtin_6726
                       test_gap_set_map_foreach_fused_receiver
                       test_gap_specabi_reassign
                       test_gap_sso_concat_string_index
```

[...]

PR #8129, opened 2026-08-15
fix(gc): list the five buffer/typed-array constructors as poll-capable

proggeramlug:
## `gc-root-dominance` has been red on `main` for ~24 hours, and the reason is that the gate was silently not checking part of what it claims to

Last green on `main` was `3381e1b89` (2026-08-13T19:27); every run since `0da668c95` (08-14T02:48) has failed. This is **not** from any of last night's merges — it predates all of them.

The gate's own `--audit-poll-reach` names the cause and the fix:

```
error: ALLOC_RE symbols that CALL a POLL_CAPABLE_RUNTIME symbol but are not in POLL_CAPABLE_RUNTIME:
  js_buffer_alloc_fill_value      -> js_buffer_from_value
  js_buffer_from_array            -> js_array_get_f64
  js_buffer_from_value            -> js_buffer_from_array
  js_typed_array_new_from_array   -> js_array_get_f64
  js_uint8array_new               -> js_typed_array_get, js_uint8array_from_array
```

## Why it matters more than "a list is missing five entries"

This is #7616's shape. The checker knows each of these returns a heap value it must track (`ALLOC_RE`) **and** knows it calls something that can re-enter JS or run a moving minor — and refused to put the two together. So a window whose only collection point is one of the five classified `MOVING: no`, and **every `--moving-only` arm — which is every gated arm, in all four modes — dropped it.**

Those windows were not being audited. The gate could not have reported a rooting bug in them, and would have stayed green if one existed.

That makes this a *"gate can't fail"* instance the gate detected about itself, which is the good case — it exited 2 rather than passing quietly.

## The fix makes the checker stricter

All five are real `pub extern "C"` exports in the buffer / typed-array construction paths, and each reaches a JS-reentrant helper — both `js_array_get_f64` and `js_typed_array_get` can run a user getter. Listing them means those windows are now classified moving and **audited**, rather than skipped.

## Verification, with real exit codes

Captured from the script itself, not from a pipeline's last stage — a `tail` swallowing an exit code is a mistake that has cost this repo twice in two days:

```
--audit-poll-reach          BEFORE exit 2   AFTER exit 0
                            AFTER prints: "no ALLOC_RE symbol reaches a poll-capable one unlisted"
--audit-alloc-re            exit 0
--audit-poll-capable        exit 0
--audit-immovable-sources   exit 0
--self-test                 exit 0
```

## One thing to watch on this PR

Because the change makes the gate stricter, the full corpus run may now surface **real** violations in the newly-audited windows. That would be the gate doing its job for the first time on those paths rather than a regression from this change — but it does mean a red `gc-root-dominance` here needs reading carefully rather than reverting. I could not run the full corpus locally (it needs the IR corpus the CI job builds).

Refs #8117, #7616.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated garbage-collection reachability auditing to correctly account for additional buffer and typed-array operations.
  * Fixed self-audit behavior for poll-reach checks while preserving existing audit and self-test results.

* **Documentation**
  * Added a changelog entry describing the expanded buffer and typed-array audit coverage.

### DIFF
--- a/changelog.d/8127-poll-capable-buffer-ctors.md
+++ b/changelog.d/8127-poll-capable-buffer-ctors.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- **`gc-root-dominance` has been red on `main` since 2026-08-14, and the reason was that the gate was silently not checking part of what it claims to.** Its own `--audit-poll-reach` self-audit exits 2 naming five exported symbols that match `ALLOC_RE` (their result is a heap value the checker must track) *and* call something already in `POLL_CAPABLE_RUNTIME`, while not being listed there themselves: `js_buffer_alloc_fill_value`, `js_buffer_from_array`, `js_buffer_from_value`, `js_typed_array_new_from_array`, `js_uint8array_new`.
+
+  The consequence is the #7616 shape: a window whose only collection point is one of those five classified `MOVING: no`, so **every `--moving-only` arm — which is every gated arm, in all four modes — dropped it**. Those windows were never checked, and the gate could not have reported a rooting bug in them.
+
+  All five are real `pub extern "C"` exports in the buffer / typed-array construction paths, and each reaches a JS-reentrant helper (`js_array_get_f64` and `js_typed_array_get` can both run a user getter). Adding them makes the checker **stricter**, not looser: it now classifies those windows as moving and audits them.
+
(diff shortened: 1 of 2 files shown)


proggeramlug:
Closing in favour of #8134, which landed the identical fix (`fa83ecab2`) with a better changelog.

We derived the same five symbols independently — I from the CI failure log on `main`, the other worker from the checker's own `--audit-poll-reach`. Same set, same reasoning about #7616's shape. Two independent derivations agreeing is worth recording; the duplicate work is not, and I should have checked for a sibling PR before opening this. That is the third duplicate I have caused in this campaign.

PR #8130, opened 2026-08-15
fix(array): route a Map/Set receiver before the array-only funnel in the fused forEach (#8117)

proggeramlug:
## What

A 1-argument `<expr>.forEach(cb)` on a `Set` or `Map` iterated **nothing**
whenever codegen could not statically prove the receiver was a collection.
Empty result, exit 0, no crash.

```ts
const holder = { s: new Set([10, 20]) };
const out: string[] = [];
holder.s.forEach((v) => out.push("n:" + v));
console.log(JSON.stringify(out));
// node:  ["n:10","n:20"]
// perry: []
```

Fixes the two `pass -> parity_fail` entries in #8117 that share this shape:

* `test_gap_collection_foreach_member_receiver_thisarg`
* `test_gap_set_map_foreach_fused_receiver`

Both reproduce **standalone** (they are not in #8117's "fails in-suite, passes
standalone" category), and both are byte-identical to node `v26.5.1` with
exit 0 after this change.

## Root cause

Codegen fuses `<expr>.forEach(cb)` to the ARRAY entry point `js_array_forEach`
when the receiver's type is unknown — `obj.someSet.forEach(cb)` is the ordinary
shape, and react-server-dom's `request.abortableTasks` is the motivating one.
#5989 put a Set/Map reroute inside that helper, but placed it **after**
`normalize_array_receiver` and its `if arr.is_null() { return; }` early-out.

#8041 (`971d6ffb6`) then widened `clean_arr_ptr` — the funnel
`normalize_array_receiver` ends in — from

```rust
if obj_type == GC_TYPE_OBJECT || obj_type == GC_TYPE_CLOSURE { /* reject */ }
```

to

```rust
// A declared TypeScript type is a hint, never a layout fact. Reject
// every tracked non-array at this shared funnel ... (#7574).
if obj_type != crate::gc::GC_TYPE_ARRAY { return std::ptr::null(); }
```

That is correct for the array-layout question, but `GC_TYPE_SET` (12) and
`GC_TYPE_MAP` (8) are tracked non-arrays. The receiver was nulled, the helper
returned before the collection question was ever asked, and the reroute became
dead code.

This is the same ordering hazard already fixed for the neighbouring funnels:
#8060/#8061 (indexed reads), #8090/#8119 (`sort`/`toSorted`/`toReversed`/`with`),
#8109 and #8120 (typed-array element helpers). `forEach` was the arm nobody had
walked yet.

**Not affected:** the 2-argument form `<expr>.forEach(cb, thisArg)` lowers to
`js_arraylike_forEach`, which reroutes via
`generic_mutators::arraylike_collection_foreach` before any array validation.
That is exactly why only the 1-arg lines of the two gap tests were red.

## Fix

Hoist the reroute into `collection_foreach_reroute`, called as the first
statement of `js_array_forEach`, before `normalize_array_receiver`. Gated on
`array_receiver_gc_tag` — the #7765 idiom `js_array_get_f64` already uses — so
an ordinary array is excluded by one already-warm GC-header byte and never
reaches a registry probe. The registry remains the liveness/layout proof.

## Tests — sabotage-verified twice

Three tests in `crates/perry-runtime/src/array/collection_tag_tests.rs`, which
is where this file's existing #7765 receiver-tag tests live.

| sabotage | result |
|---|---|
| restore the pre-fix ordering (reroute after `normalize_array_receiver`) | `..._visits_a_set_receiver` FAILS `left: []  right: [10.0, 20.0]`; `..._visits_a_map_receiver` FAILS `left: []  right: [100.0, 200.0]`; **control green** |
| delete the receiver-tag gate | `a_plain_array_foreach_iterates_without_probing_the_collection_registries` FAILS `left: (3, 3)  right: (2, 2)`; **Set/Map cases green** |
| neither | 9/9 green |

The first sabotage reproduces the exact production symptom (an EMPTY visit
list) in a unit test; the second proves the control asserts its own subject
rather than merely a correct answer.

## Validation

Built with `--profile perry-dev`, `PERRY_RUNTIME_DIR` pinned to the freshly
built archives, `grep -c "Compiling perry-runtime v" build.log` = **1**, and
`libperry_runtime.a` mtime confirmed to move after the edit.

* **Three-way control**, both gap tests: node / pre-fix perry / post-fix perry.
  Pre-fix diverges on exactly the 1-arg collection lines; post-fix is
  `cmp`-identical to node, exit 0.
* **Specificity control, correct in both arms**: gap test 1 line 5
  (`arr.forEach(fn, {base:10})` -> `36`), gap test 2 case 4 (a plain array
  through the same fused entry point -> `0:7:2,1:8:2`), and gap test 2 case 7
  (a statically-proven `Set`) are unchanged pre- and post-fix.
* **Pre/post A/B on a sibling probe** (`spread`, `Array.from`, `concat`,
  `forEach`, plain-array control): exactly one of seven lines moved —
  `forEach-set: [] -> [1,2,3]`.
* `cargo test -p perry-runtime --lib` (perry-dev profile): baseline on this
  branch point **2383 passed / 0 failed / 4 ignored**, measured 4x; with this
  change **2386 / 0 / 4**, measured 3x. +3 is exactly the tests added; nothing
  lost.
* 100 array/collection/iterator/typedarray/buffer gap tests run standalone
  against node: **96 PASS**. The 4 others are unrelated and explained —
  `test_gap_2514_settracesigint` is a recorded standing `parity_fail`;
  `test_gap_param_prop_array_index` is a node-side refusal (`node_rc=1`), the
  environment-drift class #8117 already describes; and
  `test_gap_http_req_async_iterator` / `test_gap_http2_settings` fail to link
  purely because my harness set `PERRY_NO_AUTO_OPTIMIZE=1` without building
  `perry-ext-http` in the same cargo invocation (the tokio-unification guard
  refuses the link by design).
* `lint` gates enumerated from `.github/workflows/test.yml` and run locally:
  `cargo fmt --all -- --check`, `check_file_size.sh`, and 21 python/node gates —
  all clean. `cargo clippy -p perry-runtime` has 13 pre-existing errors on
  `main` (PI-constant and regex-grammar lints in `json/`, `regex/grammar.rs`,
  `set.rs`, `builtins/numbers.rs`, …); **none** is in either file this PR
  touches.

## Caveats

* I did **not** run the 100-test sweep on a pre-fix arm; the specificity
  evidence is the pre/post probe A/B, the in-test controls, and the fact that
  the new code path is gated on `GC_TYPE_SET` / `GC_TYPE_MAP`.
* One `cargo test -p perry-runtime --lib` run on the **unmodified** branch point
  reported `2382 passed; 1 failed`. Four subsequent runs were `2383 / 0`, so it
  is a flake, but I did not capture the test name before it disappeared.
  Recording it rather than dropping it.
* `js_array_concat` (`array/concat_reverse.rs:83`) has the same
  `clean_arr_ptr`-then-`is_registered_set` shape and is therefore also dead
  since #8041. I deliberately left it: `[].concat(...[aSet])` gives node `[{}]`
  and perry `[]`, but the pre-#8041 live reroute would have produced `[1,2,3]`
  — also wrong. That call site was non-conformant before and after #8041, so it
  is a separate issue rather than part of this regression.
* `test_gap_sso_concat_string_index`, the third unowned #8117 entry, is **not**
  this bug. SSO short strings are inline `SHORT_STRING_TAG` payloads with no
  heap allocation and no `GcHeader`, so they never reach the tracked-non-array
  rejection in `clean_arr_ptr`. It still needs an owner.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `forEach` behavior for `Map` and `Set` collections.
  * Ensured collection callbacks receive and process all expected values.
  * Preserved standard array iteration behavior without unnecessary collection checks.

* **Tests**
  * Added regression coverage for `Map`, `Set`, and array `forEach` scenarios.

### DIFF
--- a/changelog.d/8130-fused-foreach-collection-reroute.md
+++ b/changelog.d/8130-fused-foreach-collection-reroute.md
@@ -1,0 +1,11 @@
+### Fixed
+
+- A 1-argument `.forEach(cb)` on a `Map` or `Set` no longer iterates nothing
+  when codegen could not statically prove the receiver was a collection
+  (`obj.someSet.forEach(cb)`, react-server-dom's `request.abortableTasks`).
+  Codegen fuses that shape to the array entry point `js_array_forEach`, whose
+  #5989 collection reroute sat behind `normalize_array_receiver`; #8041 widened
+  `clean_arr_ptr` to reject every tracked non-array, which nulls a
+  `GC_TYPE_SET`/`GC_TYPE_MAP` receiver and left the reroute unreachable. The
+  reroute now runs first, receiver-tag gated so an ordinary array still never
+  reaches a registry probe (#8117).
--- a/crates/perry-runtime/src/array/collection_tag_tests.rs
+++ b/crates/perry-runtime/src/array/collection_tag_tests.rs
@@ -328,3 +328,129 @@ fn every_registered_collection_address_carries_its_own_type_tag() {
         assert!(crate::set::is_registered_set(set as usize));
     }
 }
+
+// ---------------------------------------------------------------------------
+// #8117: the fused ARRAY `forEach` entry point must still reach a Map/Set.
+//
+// Codegen fuses a 1-argument `<expr>.forEach(cb)` to `js_array_forEach`
+// whenever it cannot prove the receiver is a collection (`obj.someSet` is the
+// ordinary shape). #5989 put a Set/Map reroute inside that helper, but AFTER
+// `normalize_array_receiver`. #8041 then widened `clean_arr_ptr` from "reject
+// GC_TYPE_OBJECT / GC_TYPE_CLOSURE" to "reject every tracked non-array", which
+// nulls a Set/Map receiver — so the reroute became unreachable and the fused
+// call silently iterated nothing.
+//
+// These assert THE SUBJECT the way this file's #7765 tests do: the Set/Map case
+// asserts the visited VALUES (an empty visit list is exactly the bug), and the
+// plain-array case asserts the registry probe counters do not move, so deleting
+// the tag gate fails even though the answer would stay correct.
+// ---------------------------------------------------------------------------
+
+use crate::closure::{js_closure_alloc, ClosureHeader};
+use std::cell::RefCell;
+
+thread_local! {
+    static FOREACH_VISITS: RefCell<Vec<u64>> = const { RefCell::new(Vec::new()) };
+}
+
+extern "C" fn record_first_arg(
+    _closure: *const ClosureHeader,
+    value: f64,
+    _index: f64,
+    _receiver: f64,
+) -> f64 {
+    FOREACH_VISITS.with(|v| v.borrow_mut().push(value.to_bits()));
+    f64::from_bits(crate::value::TAG_UNDEFINED)
+}
+
+fn take_visits() -> Vec<f64> {
+    FOREACH_VISITS.with(|v| {
+        v.borrow_mut()
+            .drain(..)
+            .map(f64::from_bits)
+            .collect::<Vec<_>>()
+    })
+}
+
+fn recording_callback() -> *const ClosureHeader {
+    take_visits();
+    js_closure_alloc(record_first_arg as *const u8, 0)
+}
+
+#[test]
+fn the_fused_array_foreach_still_visits_a_set_receiver() {
+    let (_map, _set) = arm_both_registries();
+    let set = js_set_alloc(4);
+    js_set_add(set, 10.0);
+    js_set_add(set, 20.0);
+    assert_eq!(js_set_size(set), 2);
+
+    // The precondition that made this a regression rather than a latent gap:
+    // the array-only funnel refuses this receiver, so any reroute placed after
+    // it is dead code.
+    assert!(
+        normalize_array_receiver(set as *const ArrayHeader).is_null(),
+        "#8041's array-only funnel must still refuse a Set receiver — if this \
+         starts passing the reroute is no longer the thing under test"
+    );
+
+    let cb = recording_callback();
+    js_array_forEach(set as *const ArrayHeader, cb);
+    assert_eq!(
+        take_visits(),
+        vec![10.0, 20.0],
+        "a Set reaching the fused array forEach must run Set.prototype.forEach; \
+         an EMPTY list is #8117 — the reroute ran after clean_arr_ptr nulled it"
+    );
+}
+
+#[test]
+fn the_fused_array_foreach_still_visits_a_map_receiver() {
+    let (_map, _set) = arm_both_registries();
+    let map = js_map_alloc(4);
+    js_map_set(map, 1.0, 100.0);
+    js_map_set(map, 2.0, 200.0);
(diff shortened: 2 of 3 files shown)


proggeramlug:
Independent verification on current `main` (`788f65715`, i.e. after #8124), from
a separate worktree and a separate `perry-dev` build with `PERRY_RUNTIME_DIR`
pinned and `PERRY_NO_AUTO_OPTIMIZE=1`:

* This branch's `crates/perry-runtime/src/array/` diff **applies cleanly** to
  `788f65715` (`git apply --check` passes; nothing has touched
  `iter_methods.rs` since the branch point). It is only *behind* `main`, not in
  conflict with it.
* With the patch applied, both gap tests go green and are byte-identical to
  node `v26.5.1`, exit 0:
  * `test_gap_collection_foreach_member_receiver_thisarg` — was
    `set member no-thisArg: []` vs `["n:10","n:20"]`
  * `test_gap_set_map_foreach_fused_receiver` — was losing cases 1/2/3/5/6
    (case 3's output line vanished entirely, since its `console.log` is inside
    the callback that never ran)
* Reverting the patch restores both failures, so the attribution is clean.

These are two of the three remaining #8117 entries holding the required
`conformance-smoke-complete` context red. The third,
`test_gap_sso_concat_string_index`, is the separate cause this PR's description
correctly disclaims (an SSO receiver rejected by
`js_object_get_index_polymorphic`'s tag match) and is fixed in #8151. With
**#8130 + #8151** applied together, all three pass.

Marking ready for review on that basis — the PR body's own validation reads as
complete, and this was the only thing left unconfirmed (that it still applies
and still fixes the tests now that `main` has moved).

proggeramlug:
Triage of the 5 red checks here, since none of them is caused by this change and
none of them blocks the merge:

**Not required contexts.** Branch protection requires exactly
`lint`, `cargo-test`, `parity`, `compile-smoke`, `api-docs-drift`,
`security-audit`, `conformance-smoke-complete`. `gc-root-dominance`,
`gc-root-dominance-statepoints`, `native-roots-rs4gc` and `ext-link` are none of
those.

**Four of the five are inherited from a stale base, already fixed on `main`.**
This branch is based on `83b6b8c69` (2026-08-15 02:35). Those same jobs were red
on **`main` itself** at that commit:

| job | `83b6b8c69` (this PR's base) | `fa83ecab2` (08:22) |
|---|---|---|
| GC Root Dominance | failure | **success** |
| gc-native-roots | failure | — |

`fa83ecab2` is *"fix(ci): list buffer/typed-array constructors as poll-…"* and is
in current `main` (`788f65715`). So a rebase onto current `main` should clear
`gc-root-dominance`, `gc-root-dominance-statepoints` and both
`native-roots-rs4gc` legs without touching this PR's diff.

**`ext-link` is repo-wide, not yours.** It is PR-only (it never runs on `main`),
and it is also failing on unrelated PRs — e.g. #8136. The fix is the open #8127
(*"keep the runtime's symbols global in the ELF provider link"*, for #8089).

So the remaining work here is a rebase, not a code change. I have left the branch
and its worktree untouched rather than force-pushing someone else's PR.

PR #8131, opened 2026-08-15
fix(gc): root heap values held across JS-running calls in runtime and ext-http

proggeramlug:
## Summary

A moving-GC rooting sweep plus the instruments that found it. Every fix here is one shape — **a heap value held in a Rust local (or a snapshot of a scanner-maintained table) across a call that can run JS and therefore collect** — and each was caught deterministically, not by inspection: the forced-evacuation arm of the #8036 production fixture faults at the exact instruction under `PERRY_GC_PROTECT_FROMSPACE=1`.

Split out of #8082 because none of it depends on that PR's fixture or its remaining blocker.

## The rooting fixes

**Generic array-like callback helpers** (`crates/perry-runtime/src/array/generic.rs`). `forEach`/`map`/`filter`/`some`/`every`/`find`/`findIndex`/`findLast`/`findLastIndex`/`reduce`/`reduceRight` held the receiver, the callback closure, the result under construction, and (in `filter`/`find`/`findLast`) the current element in raw locals across `js_closure_call3`/`call4` — and `al_has`/`al_get` are collection points too, since their getter and proxy paths run arbitrary JS. `js_arraylike_map` was the one that faulted: it derived the result array's element pointer once and then wrote every mapped element through it, landing in mprotect-poisoned retired from-space after the first callback allocated. All eleven now root in a `RuntimeHandleScope` and re-read from the handles at each use.

**`Function.prototype.call`/`.apply`** (`native_call_method/common_methods.rs`) held the callee closure, the explicit `this`, and the saved implicit-`this` bits across `js_native_call_value`, then handed the stale callee to `maybe_alias_explicit_this_construction`, which read its header.

**`js_put_value_set`** (`proxy/put_value.rs`) held the receiver and property key across `ordinary_set_with_receiver` — which runs user setters — before the array-subclass `length` note dereferenced the stale receiver.

**perry-ext-http listener dispatch.** Ext crates keep user closures in handle-struct side tables that registered scanners mark *and rewrite*; a **snapshot** of those tables in a Rust local is a copy no scanner can reach. Two live instances: a drained listener `Vec` went stale after the first callback's collection, and the pending-request struct built on the hyper task and parked in an mpsc channel went stale across the microtask-pump safepoint minors that run while the request waits. The emit helpers, deferred-listen drain, and close callback now root their snapshots; both dispatchers additionally **re-read** handler and listener lists from the scanner-maintained server handle at dispatch time instead of trusting the parked copy (the arrival-time `is_check_continue` routing decision is kept, and the now-orphaned `check_continue_listeners` field is removed).

New infrastructure makes the ext-crate half possible: `perry_ffi::TransientRootScope` over an extern surface (`js_ffi_root_scope_enter`/`push_heap_addr`/`get_heap_addr`/`push_nanbox`/`get_nanbox`/`scope_exit`) onto the runtime's existing transient-handle stack — the same slots `RuntimeHandleScope` uses, already marked and rewritten by a registered scanner, and already restored across JS throws by the exception savepoint machinery.

## Instruments

- **From-space scan**: bound the array walk by the LIVE length. Capacity slack holds whatever the allocator or a verbatim minor copy left there, and decoding it produced false `MISSING-REWRITE` aborts (a length-8/capacity-16 array whose slack held a dead method-table fragment). Also append a classified payload preview to each offender report, so the owner identifies itself instead of being an anonymous address.
- **`PERRY_GC_STACKMAP_TRACE=1`**: print each frame the native stack-map walk visits (ip + `dladdr` name). This is what falsified a "the walk is truncating" hypothesis — the short walks are *complete* microtask-boundary walks with no JS frames on the stack.
- **`PERRY_EH_TRACE=1`**: one line per personality invocation (phase, owning function, ip offset, decoded pad).

## Tests

- A deterministic moving-GC regression for `js_arraylike_map`: a callback that runs a copying minor on **every** invocation, asserting the receiver actually relocated and the mapped values landed in the relocated result. **Sabotage-verified** — re-hoisting the element pointer out of the loop makes it fail.
- `action_zero_pad_is_still_a_handler`: pins that a non-zero landing-pad offset with a **zero** call-site action is still Perry's catch. Under the default native-roots build that is the shape of *every* JS `try` (#7982 retypes unused catch-all pads to `landingpad token cleanup`, and LLVM emits a zero action for a cleanup clause), so reading the action as "handler vs cleanup" silently skips every statepoint-built catch and a plain `try { throw } catch` aborts FATAL "no landing pad". Nothing pinned that invariant; exactly that regression was written and reviewed on #8082 and only caught end-to-end. Sabotage-verified.
- Two **pre-existing** failures on main are fixed here (`gc_mutable_scanner_rewrites_*`): they assert a root was rewritten, which is only observable if the object actually moved, and whether a minor evacuates is a C4b policy decision that legitimately declines under unit-test conditions. The guards now force evacuation for their mutex-serialized lifetime, so the subject is guaranteed live rather than policy-dependent. (Verified failing on pristine `main` before this branch touched anything.)

## Validation

- `cargo test -p perry-runtime --lib`: 2385 passed
- `cargo test -p perry-ext-http --lib`: 68 passed (66 + the two repaired)
- `cargo test -p perry-ffi --lib`: 26 passed
- `cargo fmt --check`, `check_file_size.sh`, `addr_class_inventory.py`, `gc_runtime_root_holders.py`: clean

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved HTTP/HTTPS callback reliability during garbage collection.
  - Fixed array methods, function calls, property updates, and reductions when memory moves.
  - Corrected array scanning to ignore unused capacity and stale data.
  - Improved exception-handler behavior for zero-action handlers.
  - Added safer callback handling for moving-memory scenarios.

- **Diagnostics**
  - Added optional garbage-collection, stack, and exception-handling tracing.

- **Documentation**
  - Added changelog entries covering garbage-collection safety improvements.

### DIFF
--- a/changelog.d/8131-moving-gc-rooting-sweep.md
+++ b/changelog.d/8131-moving-gc-rooting-sweep.md
@@ -40,10 +40,9 @@
   surface onto the runtime's transient-handle stack, so ext crates can root
   the table snapshots they hold across JS callbacks.
 
-- Instruments: the whole-heap from-space scan bounds array walks by live
-  length (capacity slack was producing false MISSING-REWRITE aborts) and
-  appends a classified payload preview to each offender so the owner
-  identifies itself; `PERRY_GC_STACKMAP_TRACE=1` prints every frame the
+- Instruments: the whole-heap from-space scan appends a classified payload
+  preview to each offender, so the owner identifies itself instead of being
+  an anonymous address; `PERRY_GC_STACKMAP_TRACE=1` prints every frame the
   native stack-map walk visits; `PERRY_EH_TRACE=1` prints per-frame
   personality decisions.
 
--- a/changelog.d/8131-moving-gc-rooting-sweep.md
+++ b/changelog.d/8131-moving-gc-rooting-sweep.md
@@ -1,0 +1,57 @@
+### Fixed
+
+- Root heap values held in Rust locals across calls that can run JS and
+  therefore collect. Each was caught by a fault at the exact instruction
+  under `PERRY_GC_PROTECT_FROMSPACE=1`, not by inspection:
+  - the generic array-like callback helpers (`forEach`, `map`, `filter`,
+    `some`, `every`, `find`/`findIndex`/`findLast`/`findLastIndex`,
+    `reduce`/`reduceRight`) held the receiver, callback, result under
+    construction, and current element across `js_closure_call*`; `map`
+    wrote every mapped element through a pre-collection element pointer,
+    landing in retired from-space;
+  - `Function.prototype.call`/`.apply` held the callee, the explicit
+    `this`, and the saved implicit-`this` across the invocation, then read
+    the stale callee's header in the native-this alias check;
+  - `js_put_value_set` held the receiver and property key across
+    `ordinary_set_with_receiver` (which runs user setters) before the
+    array-subclass `length` note dereferenced them.
+
+- Keep perry-ext-http's listener dispatch on values the collector can see.
+  Ext handle-struct side tables are rewritten by registered scanners, but a
+  SNAPSHOT of one in a Rust local is a copy no scanner reaches: a drained
+  listener vec went stale after the first callback's collection, and the
+  pending-request struct parked in an mpsc channel went stale across the
+  microtask-pump safepoint minors that run while the request waits. The
+  emit helpers, deferred-listen drain and close callback now root their
+  snapshots, and both request dispatchers re-read handler and listener
+  lists from the scanner-maintained server handle at dispatch time. The
+  orphaned `HttpPendingRequest::check_continue_listeners` field is removed;
+  only the routing bit crosses the channel.
+
+- Repair two GC scanner tests that were failing on `main`: they assert a
+  root was rewritten, which is only observable if the collection actually
+  moved the object, and evacuation is a C4b policy decision that
+  legitimately declines under unit-test conditions. Their guards now force
+  evacuation for their mutex-serialized lifetime.
+
+### Added
+
+- `perry_ffi::TransientRootScope` — a safe wrapper over a new extern
+  surface onto the runtime's transient-handle stack, so ext crates can root
+  the table snapshots they hold across JS callbacks.
+
+- Instruments: the whole-heap from-space scan bounds array walks by live
+  length (capacity slack was producing false MISSING-REWRITE aborts) and
+  appends a classified payload preview to each offender so the owner
+  identifies itself; `PERRY_GC_STACKMAP_TRACE=1` prints every frame the
+  native stack-map walk visits; `PERRY_EH_TRACE=1` prints per-frame
+  personality decisions.
+
+### Testing
+
+- A deterministic moving-GC regression for `js_arraylike_map` (a callback
+  that runs a copying minor on every invocation), sabotage-verified.
+- `action_zero_pad_is_still_a_handler` pins that a zero call-site action is
+  still Perry's catch — the shape of every JS `try` under native roots
+  (#7982) — so reading the action as "handler vs cleanup" can no longer
+  silently disable every statepoint-built catch. Sabotage-verified.
--- a/crates/perry-ext-http/src/server/http2_server/pump.rs
+++ b/crates/perry-ext-http/src/server/http2_server/pump.rs
@@ -123,7 +123,6 @@ pub(crate) async fn handle_h2_request(
         h2_stream_headers,
         request_listeners,
         handler,
-        check_continue_listeners: Vec::new(),
         is_check_continue: false,
     };
     if request_tx.send(pending).await.is_err() {
--- a/crates/perry-ext-http/src/server/https_server.rs
+++ b/crates/perry-ext-http/src/server/https_server.rs
@@ -459,15 +459,38 @@ pub(crate) fn process_pending_https(pending: HttpPendingRequest) {
     // #4903 — Node invokes `'request'` listeners (and the `createServer`
     // handler, which is one) with `this` bound to the server.
     let server_this = handle_to_pointer_f64(pending.server_handle);
+    // #8082 (same as the HTTP path): the channel-parked snapshot's closure
+    // addresses are copies no scanner rewrites — re-read them from the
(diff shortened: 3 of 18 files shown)


proggeramlug:
Status note for reviewers: the three red checks on this PR are **pre-existing on `main`**, not introduced here. Both are now tracked:

- `ext-link` → #8155. `perry-ext-fetch` fails to link with undefined `js_file_new` / `js_headers_init_from_value`. This PR touches neither `perry-ext-fetch` nor any fetch symbol; the identical failure reproduces on unrelated branches (e.g. #8079's).
- `native-roots-rs4gc` (both ELF arms) → #8156. The provider fixture aborts at its handshake with `"stdlib provider is bound to a different runtime image"`. The same error appears on `main`'s own runs (`83b6b8c69`, `506f4ab11`), with `5c27d1ad9` the last green one.

Everything this PR does touch is green, and was validated locally at the current head: `perry-runtime` 2385 passed, `perry-ext-http` 68 passed (66 pre-existing + the two this PR repairs), `perry-ffi` 26 passed, plus `cargo fmt --check`, `check_file_size.sh`, `addr_class_inventory.py` and `gc_runtime_root_holders.py` clean.

proggeramlug:
Held out of today's merge batch on two things — one mechanical, one that needs your judgement rather than mine.

## 1. Conflicts with `main` (I resolved it locally; you should confirm the choice)

`crates/perry-runtime/src/gc/fromspace_scan.rs` conflicts with #8084, now on `main`. The conflict is interesting: **both PRs independently found and fixed the same bug** — array capacity slack producing false `MISSING-REWRITE` reports in the from-space scan, because elements `[length..capacity)` hold recycled bytes the collector never rewrites by design.

Your version:
```rust
let live = header_words + (*arr).length as usize;
payload_words = payload_words.min(live);
```

#8084's, already on `main`, bounds it the same way but also **counts what it skipped** (`report.array_slack_words_skipped`), with the rationale that otherwise "a shrinking scan cannot silently read as a cleaner heap". That is the stronger form — a bound that reports nothing can quietly hide coverage loss — so when I resolved this I kept `main`'s and dropped the duplicate. Your changelog entry for that hunk is now redundant; the rest of your diff is untouched.

Worth noting you two hit this from different fixtures (#8036's gate vs #7803's seeded runs) and reached the same conclusion, which is good evidence the diagnosis is right.

## 2. `raw_handle_debt` goes red — this one is yours

```
bare raw-handle reads: 996 (baseline 992)
crates/perry-runtime/src/array/generic.rs: 4 bare read(s) in a module with no ceiling.
```

Green on the same tree with the other three PRs of the batch and without this one, so it is not inherited. `array/generic.rs` carries no ceiling entry, so the ratchet requires new code there to use the sanctioned accessors rather than a bare `get_nanbox_f64()`.

I did not convert them myself. The policy splits by shape — `across_{mut,const,nanbox}` for a **post-call reload**, `with_{mut,const}_ptr` for a **scoped argument to a non-allocating operation or self-rooting entry point** — and the sites in question are argument-position reads feeding `al_length`, `al_has`, `al_get` and `callable`. `callable` invokes a user callback and therefore allocates, so which of those two forms is correct depends on what you intended each site's lifetime to be, and guessing at that in a rooting PR is exactly the wrong place to save time. You'll convert these faster and more correctly than I would.

Raising the module ceiling would also clear it, but the ratchet's own message discourages that and #7341's direction is conversion.

## What is fine

Everything else validated on the merged tree: all other lint gates green, `cargo test -p perry-runtime --lib` **2409 passed / 0 failed / 4 ignored**, `cargo test -p perry-ext-http` **68 passed / 0 failed**, `check_file_size`, `poll-reach`, `gc_runtime_root_holders`, `gc_pin_sites` all clean.

So this is a rebase plus four accessor conversions away from landing. Ping me when it is pushed and I will take it.

proggeramlug:
Both done and pushed — rebased onto `main` (`d6d7d0efe`), head is now `695d44b30`.

## 1. Conflict with #8084 — kept `main`'s, dropped mine

Agreed on the merits, and for the reason you gave rather than just seniority: a bound that reports nothing can hide coverage loss, so `array_slack_words_skipped` is the version that should survive. My duplicate hunk is gone and the redundant changelog bullet with it; what remains of that commit is only the payload preview and `PERRY_GC_STACKMAP_TRACE`.

Worth recording that the two fixtures agreed on the *misleading* part too: on #8036's gate the slack held a dead method-table fragment, which is what an unrooted-closure bug looks like, and it cost me a false root cause before the owner dump settled it — same shape as your StringDecoder.

## 2. `raw_handle_debt` — converted, not ceiling-raised

Four sites, split by shape exactly as you described:

- **`js_arraylike_map`'s post-callback reload → `across_mut`.** This is strictly better than what I had, not just gate-compliant: the pre-call address is no longer *nameable*, so the pointer #8082 wrote through cannot be reached for at all, rather than being re-read correctly by convention.
- **`js_array_push_f64` argument → `with_mut_ptr`.** Self-rooting for the array it is handed — its grow path roots it and re-reads after the allocation — and it returns the current address, which I store straight back into the handle.
- **Both `nanbox_arr` tails → `with_mut_ptr`.** Non-allocating; it only tags the pointer.

On the one judgement call you flagged: `callable()` does run a user callback, but it is invoked *inside* the `across_mut` closure, so its allocation is on the covered side of the reload. The pushed value is still read from its handle before the call, so semantics are unchanged.

I re-ran the sabotage check after converting — hoisting the element pointer back out of the reload still makes `arraylike_map_survives_a_moving_minor_inside_every_callback` fail, so the conversion did not quietly defang the regression.

## Validation at `695d44b30`

`cargo test -p perry-runtime --lib` **2409 passed / 0 failed / 4 ignored**, `perry-ext-http` **68 passed**, `perry-ffi` **26 passed**. Gates: `raw_handle_debt` back to **992 = baseline** (no debt added), plus `check_file_size`, `addr_class_inventory`, `gc_runtime_root_holders`, `gc_pin_sites`, poll-reach and `cargo fmt --check` all clean.

The two pre-existing red gates are unchanged and tracked in #8155 (`ext-link` / perry-ext-fetch) and #8156 (`native-roots-rs4gc` provider handshake); neither is touched by this PR.

PR #8134, opened 2026-08-15
fix(ci): list buffer/typed-array constructors as poll-capable

proggeramlug:
`gc-root-dominance` has failed every scheduled run on `main` since #8120 landed the call edges — the last three nightlies (01:51, 19:15, 13:46) all die the same way, in **both** the `gc-root-dominance` and `gc-root-dominance-statepoints` jobs.

## What fails

`--audit-poll-reach` catches exactly one shape (#7616): a symbol whose result the checker tracks as a heap value (`ALLOC_RE`) that calls something already known to re-enter JS or run a moving minor, without being listed itself. A window whose only collection point is such a call classifies `MOVING: no`, so every `--moving-only` arm — which is every gated arm, in all four corpus × lowering modes — silently drops it.

Five buffer/typed-array constructors reach an element read that way. The audit names them with their edges:

```
js_uint8array_new             -> js_typed_array_get, js_uint8array_from_array
js_typed_array_new_from_array -> js_array_get_f64
js_buffer_from_array          -> js_array_get_f64
js_buffer_from_value          -> js_buffer_from_array
js_buffer_alloc_fill_value    -> js_buffer_from_value
```

## Why they belong in the set

The reads are the reach proof: a source element can be an accessor or a Proxy `get` trap — user JS — and the per-element loop allocates the destination as it goes. `Buffer.from(arr)` and `new Uint8Array(arr)` are therefore collection points like any other element-reading builtin, and a value held live across one is exactly as much at risk as one held across `js_array_map`.

## Validation

- `--audit-poll-reach` against locally built runtime archives: **exit 2 → exit 0** (`no ALLOC_RE symbol reaches a poll-capable one unlisted`), converging in one pass — listing these five surfaces no further unlisted callers.
- Widening the set is **one-sided**: it can only make windows visible that the `--moving-only` arms previously dropped. That is the point of the change, and it is also the only way it could newly exceed a budget, so the four corpus × lowering gates in this PR's CI are what confirm the budgets still hold. I have not asserted that locally — the corpus check outran my local time budget, and a claim I did not measure does not belong in the changeset.

Split out of #8084 (the #7803 investigation) so main's nightly can go green without waiting on that branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability for buffer and typed-array operations involving custom accessors or proxy traps, preventing intermittent runtime failures.
  * Added validation to ensure garbage-collection safety checks correctly cover these operations.

* **Documentation**
  * Documented the garbage-collection safety improvements.

* **Chores**
  * Updated the application version to 0.5.1511.


proggeramlug:
Audited and merging, after pushing `63571ecf7` to drop the version bump.

**The bump:** this carried `0.5.1510 → 0.5.1511` in `Cargo.toml`, `CLAUDE.md` and 152 lines of `Cargo.lock` churn. The maintainer bumps at merge time and there is a standing "no version bump for now" instruction, so I reverted all three and regenerated the lock with `cargo metadata` rather than hand-editing. The manifest now has **zero** diff against `main` and no `0.5.1511` string survives anywhere.

**The fix is right, and it is the one I independently arrived at.** I opened #8129 for the same five symbols before seeing this; I am closing mine in favour of yours, which has the better changelog. Recording that we converged, since two independent derivations of the same five-symbol set is worth more than either alone.

**Verified on a landing-equivalent tree** (merged current `main` `0a1e78e5f`), with real exit codes captured from the script rather than a pipeline's last stage:

```
--audit-poll-reach          exit 0   ("no ALLOC_RE symbol reaches a poll-capable one unlisted")
--audit-alloc-re            exit 0
--audit-poll-capable        exit 0
--audit-immovable-sources   exit 0
--self-test                 exit 0
```

On unmodified `main` that first one exits **2**. Real change vs `main` after the strip is exactly two files: the script and the changelog fragment.

**Why this matters more than five list entries**, for the record: it is #7616's shape. The checker knew each of these returns a heap value it must track *and* knew it calls something that can re-enter JS — and refused to combine the two, so any window whose only collection point was one of the five classified `MOVING: no` and was **dropped from every `--moving-only` arm, which is every gated arm in all four modes.** Those windows were not audited and the gate would have stayed green with a rooting bug in them. `gc-root-dominance` has been red on `main` since `0da668c95` (08-14 02:48) because of it.

One thing to watch after this lands: the gate is now **stricter**, so the full corpus run may surface real violations in newly-audited windows. That would be the gate doing its job for the first time on those paths, not a regression from this change.

Issue #8133, opened 2026-08-15
js_class_method_bind still binds a movable heap string's interior in four places #7747 missed (timer handles, TextDecoder/TextEncoder)


proggeramlug:
#7747 fixed two callers that bound a movable GC heap string's interior as a bound-method name. Four more sites in the same neighbourhood still do it.

### The contract

`js_class_method_bind(instance, method_name_ptr, method_name_len)` stores the name **pointer** in the bound closure (capture 1) and `dispatch_bound_method` (`closure/dispatch/bound.rs:12`) re-reads it at call time. Its doc requires the pointer to stay valid for the closure's lifetime, which codegen satisfies with rodata.

#7747's commit message states the failure mode exactly:

> `get_field_by_name_tail` passed `key + size_of::<StringHeader>()` — the interior of a movable GC heap string that is unreachable once the read returns — so a copying minor could relocate or reclaim the bytes the closure names. […] Whether the stale bytes still spell the method is an allocator property, not a program property, which is why this passed locally and took a SIGSEGV on conformance-smoke shards 7 and 8.

It fixed the two Buffer arms (`object/field_get_set/buffer_own_prop.rs` and the computed-key arm in `object/polymorphic_index.rs`), both of which now bind a `'static` literal via `buffer_dispatch::buffer_method_name_static`.

### Still unfixed, same file, same pattern

Each of these computes `let key_ptr = (key as *const u8).add(size_of::<crate::StringHeader>())` and hands that pointer to `js_class_method_bind`:

* `crates/perry-runtime/src/object/field_get_set/get_field_by_name_tail.rs:48` — timer-handle method key on a small-handle receiver (`is_timer_handle_method_key` + `timer::is_known_timer_id`).
* `crates/perry-runtime/src/object/field_get_set/get_field_by_name_tail.rs:121` — the same arm for an already-stripped handle-band receiver.
* `crates/perry-runtime/src/text.rs:649` and `:670` (`text_handle_property`) — `TextDecoder.prototype.decode` and `TextEncoder.prototype.encode`/`encodeInto` read as VALUES. That function's own docstring says value reads are the reason it exists (`K.decode.bind(K)` — "the shape a minified SDK's cached decodeText helper takes"), so this is the intended hot path, not an edge. Reached from `get_field_by_name_tail.rs:53` and `:125`, `get_field_by_name.rs:875`, and `ic_miss.rs:540`, all of which pass the same heap-string interior.

### Not affected

[...]

PR #7314, opened 2026-08-03
Native-frame GC roots via LLVM statepoints, opt-in (#7173, #7174)

proggeramlug:
Adds a second precise-root mechanism alongside the shadow stack, entirely opt-in. **Without `PERRY_STATEPOINTS=1` nothing in this PR runs** — the shadow stack remains the shipping root mechanism and the default path is byte-for-byte unchanged.

Closes the research tracked in #7173 / #7174.

## Why

Perry's shadow stack requires the compiler to remember to root every value live across a call. A missed root is silent and only manifests as a collection freeing a live object. Under `gc.statepoint`, LLVM — which already knows what is live across each call — owns that knowledge and rewrites the values after a collection moves them, so **the forgot-to-root bug class becomes structurally impossible** rather than merely tested for.

## Every root path fails closed

The plain `llvm.experimental.stackmap` lowering is **deleted**, not kept as a fallback. LLVM may record a root slot's address as `Register R#N` — caller-saved and unrecoverable at collection time — so falling back to it silently loses roots (measured 3 of 60 locations on one probe). It survived in three places, all of which failed *open*:

1. `PreciseRootBackend::StackMap` was dead by construction: both sites setting `stack_map_requested` are guarded by `native_stack_roots_enabled()`, which is exactly `statepoints || rs4gc`. Variant and emitter deleted.
2. The statepoint backend fell back for calls it could not parse — chiefly **indirect** calls. That was a limitation of Perry's textual parser, not of statepoints: `gc.statepoint` takes its callee as a `ptr` operand, so `ptr elementtype(T) %fnptr` is as valid as `... @callee`. Indirect targets are now statepoint-able; anything still unparseable is a hard compile error naming the call shape.
3. The compact-map rewriter fell back to keeping LLVM's section. That reads as conservative and is not — the runtime reads only `__perry_gcmap`, so those records sit in the binary unread and that module's roots go missing, while other modules still emit a valid section so even the runtime's "present but undecodable" guard stays quiet. Now a hard error.

## The metadata is re-encoded, not shipped as LLVM emits it

On `test-drizzle-pg`, `__llvm_stackmaps` was **4.21 MB**, and over half of it was data the runtime already discarded at startup: three `Constant` slots per record (`gc.statepoint`'s calling-convention preamble) and a duplicate of every root, since LLVM records base and derived and Perry has no interior pointers.

[...]

PR #6406, opened 2026-07-14
fix(runtime): give Buffer instances own properties, method values, and `in`

proggeramlug:
## The bug

A Buffer is an ordinary `Uint8Array` in Node: user code can hang own properties on one, read `Buffer.prototype` methods as **values**, and shadow them. Perry stored buffers outside the object model — a raw `BufferHeader` with no property table — so all three were missing:

```js
const b = Buffer.alloc(8);
b.myFlag = 42;
b.myFlag;             // node: 42          perry: undefined   (the write was dropped)
typeof b.readUInt8;   // node: "function"  perry: "undefined" (methods dispatched on CALL only)
"myFlag" in b;        // node: true        perry: false
```

## Why it matters

mysql2 sizes **every outgoing packet** with exactly that idiom (`MockBuffer`): it walks the prototype, replaces the matching write methods on a zero-length Buffer with a no-op, serializes once to *measure*, then allocates for real:

```js
for (const k in Buffer.prototype)
  if (typeof mock[k] === "function") mock[k] = noop;
```

With the `typeof mock[k] === "function"` probe reading `undefined`, nothing was replaced, the measuring pass wrote into the empty Buffer, and the MySQL handshake died with `RangeError [ERR_OUT_OF_RANGE]`.

## The fix

An address-keyed own-property table for buffers, wired into the read / write / call / `in` / computed-index paths and scanned by the GC (both the owner address and the NaN-boxed values, in every phase). `register_buffer` clears the entry first, so a recycled address cannot inherit a dead buffer's properties. Numeric keys remain byte access and never become own props.

The `Buffer.prototype` method list is now generated from `is_buffer_method_name` (115 names; node enumerates 93) instead of an 11-name stub, so the `for (const k in Buffer.prototype)` walk actually sees them.

## Known limitation — #6405

An own property does **not** shadow a prototype method when the call folds to the inline byte-load intrinsic: when codegen can statically prove the receiver is a buffer *and* the method name is a literal (`b.readUInt8(0)`, `b["readUInt8"](0)`). Every dynamic path is correct — which is why the mysql2 idiom works in practice, since the proof fails on minified receivers. Fixing it costs a guard on Perry's hottest byte-access path (a runtime flag + branch, or a whole-program pre-codegen scan), so it deserves a deliberate decision rather than a drive-by. Filed as #6405.

## Test

[...]

PR #8141, opened 2026-08-15
fix(object): a Buffer/DataView receiver must not reach the ordinary object walk (#8117)

proggeramlug:
Closes the two `pass -> crash` entries of #8117: `test_gap_buffer_own_props` and `test_gap_6386_dataview_concat_regex_fastpaths`. Both are **SIGSEGV**, both reproduce **standalone on Linux**, and both are this one missing receiver arm.

## The bug

`obj_value_has_own_key` (`crates/perry-runtime/src/object/reflect_support.rs`) has arms for a registry typed array, a `GC_TYPE_ARRAY`/`GC_TYPE_LAZY_ARRAY`, a closure, and a native-module namespace. It had **none for a Buffer / ArrayBuffer / DataView**, so one fell through to the ordinary `ObjectHeader` arm.

A buffer is a `BufferHeader`: no `class_id`, no `keys_array`. The walk read `(*obj).keys_array` out of the bytes that follow a buffer header and handed that to `js_array_length`, whose lazy-array probe dereferences `addr - 8`. The only thing between the two was a `< 0x10000` magnitude floor, which arbitrary payload bytes clear routinely.

Four lines reproduce it — 10/10 on Linux:

```ts
const b: any = Buffer.alloc(8);
b.readUInt8 = function () { return "shadowed"; };
const k = "readUInt8";
b[k](0);
```

```
#0  js_array_length                                        <- SIGSEGV, x0 = 0x12b00003aa1f03e2
#1  perry_runtime::object::reflect_support::obj_value_has_own_key
#2  perry_runtime::proxy::own_set_descriptor
#3  perry_runtime::proxy::ordinary_set_with_receiver
#4  js_put_value_set
#5  js_put_value_set_dyn_ic_miss
#6  main
```

`x0` is payload bytes, not an address. The faulting instruction is `ldurb w9, [x20, #-8]` followed by `cmp w9, #0x9` and a compare against `0x4C5A5841` — `GC_TYPE_LAZY_ARRAY` and `LAZY_ARRAY_MAGIC`.

This is the same **"ask the receiver question before the generic path claims it"** shape as #8090 / #8109 / #8119 / #8120, on the has-own-key / `[[Set]]` path instead of the element path.

## The fix

A registered-buffer arm, placed beside the typed-array arm and for the same stated reason (it must precede the GC-header read, because small slab buffers carry no `GcHeader` at all).

A buffer's own string keys are exactly its expando table (#6406). Prototype methods are **inherited, not own** — that is what lets `buf.readUInt8 = fn` install a shadowing own property instead of being treated as the redefinition of an existing one. Canonical integer indices are deliberately **not** folded in: the byte-index `[[Set]]` is routed upstream of this call, and answering "own" for one would divert it into the ordinary data-property store.

Second, smaller change: the `keys_array` guard becomes `addr_class::is_plausible_heap_addr` instead of the bare `< 0x10000` floor — defence in depth for the class this fix closes by routing. A receiver kind with no arm here should produce a wrong answer, not a SIGSEGV.

## Why it read as twelve days old and macOS-only

The garbage `keys_array` has to clear the floor **and** land unmapped. macOS's 2 TB heap floor means it usually reads as null, so the same call silently answered "no own key" for a property the buffer really owns — wrong, but not fatal. That is why 30 runs each on macOS/arm64 were clean under every GC instrument (`PERRY_GC_ZEAL`, `PERRY_GC_ZEAL_ALLOC_KB=0`, forced evacuation, from-space protect + scan, `PERRY_GC_SCHEDULE_RATE=1`, `PERRY_GEN_GC=0`) while CI segfaulted.

It is also why the CI-log bisect landed on #7314 (2026-08-03): that commit added `initialize_stack_maps()` to `js_gc_init`, whose Linux arm `std::fs::read`s the whole executable at startup — a Linux-only multi-megabyte alloc/free before any user code, which moved the heap enough to make the pre-existing garbage read fatal. **The defect is older than #7314; #7314 only made it land.** Details in https://github.com/PerryTS/perry/issues/8117#issuecomment-5300864212.

**The new test asserts the ANSWER, not "did not crash"**, so it fails on macOS too.

## Testing

`object::tests::buffer_own_key_comes_from_the_expando_table_not_the_object_walk` — watched fail with the buffer arm removed:

```
panicked at crates/perry-runtime/src/object/tests.rs:1618:
a buffer's own expando property must be reported as an own key
```

It also asserts a `Buffer.prototype` method and an unknown key are **not** own, so the arm cannot pass by answering `true` unconditionally.

* `cargo test -p perry-runtime --lib` → **2390 passed, 0 failed, 4 ignored** (repo baseline 2389 + this test), exit 0. `grep -c "Compiling perry-runtime v"` on the restore build = **1**.
* End-to-end on Linux (ubuntu 24.04 aarch64 container, `--release`, `PERRY_NO_AUTO_OPTIMIZE=1`, `PERRY_RUNTIME_DIR` pinned), before → after:

  | fixture | before | after |
  |---|---|---|
  | the 4-line repro above | 10/10 SIGSEGV | 20/20 exit 0 |
  | `test_gap_6386_dataview_concat_regex_fastpaths` | 25/25 SIGSEGV | 20/20 exit 0 |
  | `test_gap_buffer_own_props` | SIGSEGV (auto-optimize) | 20/20 exit 0 |

  Both gap fixtures are **byte-identical to node v26.5.1** after the fix (`diff` clean; `node --version` checked against `.node-version`).
* The x86-64 side is confirmed independently: on `ubuntu-latest` at base `fa83ecab2`, `test_gap_buffer_own_props` segfaults **standalone**, outside the gap harness — so this is not a test-isolation or ordering artefact.
* rustfmt, `scripts/check_file_size.sh`, and all sixteen `lint` gate scripts run individually, all exit 0. `raw_handle_debt` included: the new arm carries its address across the GC-capable key coercion with `across_mut` rather than a bare handle read, so the file stays at its ceiling of 3.

## Not in this PR

* #8136 — the buffer own-property table is never pruned when its buffer dies. Found in the same read, independent defect, separate PR.
* #8133 — four `js_class_method_bind` sites that still bind a movable heap string's interior (#7747's siblings).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a Linux crash that could occur when dynamically checking properties on Buffer objects.
  * Corrected own-property checks for Buffer, ArrayBuffer, and DataView objects.
  * Ensured inherited methods and indexed properties are not incorrectly reported as own properties.
  * Added safeguards for invalid object references during property checks.

* **Tests**
  * Added regression coverage for Buffer expando properties, inherited methods, and unknown keys across supported platforms.


proggeramlug:
Control run, to keep the two halves of this PR honestly labelled.

Rebuilt in the Linux container with **only the buffer arm** — the `keys_array` guard reverted to the original `< 0x10000` floor — `grep -c "Compiling perry-runtime v"` = 1:

```
ARMONLY mini_buf exits: 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
ARMONLY 6386     exits: 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0   lines=28
```

Both clean, 15/15. So the arm alone fixes the crash and the `is_plausible_heap_addr` change is genuinely defence in depth for the *next* missing arm, not part of the fix — which is what the description claims. Happy to drop it into a separate PR if you'd rather keep this one to the single arm.

Issue #7128, opened 2026-07-31
repsel: the coverage work converted exactly as predicted, and almost none of it is faster (one −4.1% win, one +14.9% regression)


proggeramlug:
Follow-up measurement for #7106 / #7034, on `main` at `95988fb4d`.

**The coverage work converted exactly as #7122 predicted — per local, with zero
residue and no further mechanism. Almost none of it is faster.** Corpus-wide the
campaign is a **null on speed with one real win and one real regression**, and
the regression is larger than the win.

- canonical `Str` (#7121, `08_string_concat`): **−4.1% instructions**, and it
  survives at the committed 100,000-iteration size. Reconciles with #7125.
- canonical i32 (#7121, `11_prime_sieve`): **−1.05% instructions**.
- canonical i32 (#7121, `15_mandelbrot`): **+14.87% instructions** — a
  regression, bisected to #7121 by binary hash.
- everything else: **0.00%**, at a measured noise floor of 0.02%.

The reason is one sentence: **selection is not emission.** canonical-i32 now
promotes in 17 of 18 real workloads, and changes the emitted object on 2.

---

## Method, and what had to be true before any number was believed

Four compiler arms per host, all built from **one target dir in one session**, so
the rustc/LLVM toolchain is identical across arms by construction:

| arm | commit | what it is |
|---|---|---|
| `base` | `3f1a0853a` | census-only commit, before the campaign |
| `pre7122` | `e7bc73bd6` | before the loop-induction rule |
| `at7122` | `4d3ddc9a3` | #7122 in, module-init gate still on |
| `main` | `95988fb4d` | after #7107 #7113 #7117 #7119 #7121 #7122 |

`cargo build --release -p perry -p perry-runtime-static -p perry-stdlib-static`
throughout — building the runtime crates alone leaves a stale `.a` and both arms
behave identically, which for *this* experiment would have been a catastrophic
false null. Archive mtimes verified to move per arm; `perry` and
`libperry_runtime.a` md5s verified distinct on both hosts.

Hosts, with load stated per measurement:

- **Raspberry Pi 5**, aarch64 Linux, 4 cores. Load **0.16** when timing started,
  1.01 after. `perf stat -e instructions:u`, 11 repeats, warmed. **Primary.**
- **Mac mini M1**, 8 cores. Load ~4–8 (it runs the long-lived `perry-ship`).
  Wall time only — macOS has no unprivileged instruction counter, so
  instructions retired on Apple silicon is something I **could not measure**.

[...]

PR #8150, opened 2026-08-15
perf(codegen): a reassigned numeric accumulator is a Number by construction (#8105)

### LINKED ISSUE #8105: perf(codegen): 15_mandelbrot's inner loop routes every FP multiply through js_dynamic_mul — 201 instructions/iteration vs #7128's ~13
Found while re-measuring #7146 on `main` @ `514ff7c9a` (macOS arm64). It is not
a repsel issue; it invalidates the arithmetic that several repsel cost models
were calibrated on, so it is worth its own number.

## The observable

`benchmarks/suite/15_mandelbrot.ts`, unmodified. Its innermost loop is

```ts
while (x * x + y * y <= 4.0 && iter < MAX_ITER) {
    const xtemp = x * x - y * y + cx;
    y = 2.0 * x * y + cy;
    x = xtemp;
    iter = iter + 1;
}
```

with `let x = 0.0; let y = 0.0;` — nothing but doubles. The emitted AArch64 for
that loop (`PERRY_GC_MOVING_LOOP_POLLS=0` so it is one basic block, which makes
it readable; the poll-on default splits it but changes nothing here):

```
movi.2d v10, #0        ;; x
movi.2d v9,  #0        ;; y
mov.16b v0, v10 ; mov.16b v1, v10 ; bl _js_dynamic_mul     ;; x * x
mov.16b v0, v9  ; mov.16b v1, v9  ; bl _js_dynamic_mul     ;; y * y
fsub d0, d11, d0
fadd d11, d13, d0
fmov d0, #2.0
mov.16b v1, v10 ; bl _js_dynamic_mul                        ;; 2.0 * x
mov.16b v1, v9  ; bl _js_dynamic_mul                        ;; … * y
fadd d9, d8, d0
fmov d0, #1.0
fadd d14, d14, d0
mov.16b v0, v11 ; mov.16b v1, v11 ; bl _js_dynamic_mul     ;; x * x (guard)
mov.16b v0, v9  ; mov.16b v1, v9  ; bl _js_dynamic_mul     ;; y * y (guard)
fadd  d0, d10, d0
fcmp  d0, d12
fccmp d14, d15, #0x0, ls
b.mi  <top>
```

**29 instructions, six `bl _js_dynamic_mul`.** `nm -u` on the object also lists
`_js_dynamic_sub`. Every `+`, `-` and the comparisons are inline; only the
multiplies escape to the dynamic helper.

Whole-program cost: **1,610,420,345 instructions retired** (`--release`
compiler and `--release` `libperry_{runtime,stdlib}.a`, `PERRY_RUNTIME_DIR`
pinned, median of 5, output `total_iter:8011148`), over **8,011,148** innermost
iterations = **201 instructions per iteration**.

Reproduced with and without `PERRY_NO_AUTO_OPTIMIZE=1`, and with `--profile
perry-dev` artifacts (1,613,899,624 — within 0.2%, so this is generated code,
not runtime-helper weight).

## Why it looks like a regression

#7128 measured this exact loop as **"one basic block of 12 instructions"** with
a fused `fcmp` + `fccmp` exit test, and computed the whole +14.87%
canonical-i32 regression as **2 instructions × 8,011,148 iterations ≈ 16.0M**
against a measured +15.63M. That arithmetic only closes if the loop really was
~13 instructions. It is now 201, i.e. **~15× more**, and the difference is six
calls that a `let x = 0.0` local should not need.

I did not bisect it, so "regression" is an inference from #7128's published
figure, not a measured before/after.

## Why it matters beyond one benchmark

`collectors/repsel_benefit.rs`'s entire cost model is a ±2-instruction argument
about this loop, written into the module header as the justification for a
profitability refusal. At 201 instructions per iteration that term is ~0.4% of
the loop, which is one of the reasons the refusal now measures 0.0 ± 0.2% on
every workload where it changes bytes — see
https://github.com/PerryTS/perry/issues/7146#issuecomment-5298724389.

## Suspected area

The multiply is the only operator that escapes, which points at the
multiplicative arm of the numeric fast path rather than at type flow for `x`
generally (the `fadd`/`fsub` on the same values stay inline). CLAUDE.md's
"two arithmetic tiers" note (`lower_numeric_binary_value` bypassing
`binary.rs`) is the first place I would look.

## Not measured

x86-64; no bisect; `--profile dist`; whether other float-heavy suite workloads
(`06_math_intensive`, `16_matrix_multiply`) show the same escape.

proggeramlug:
Closes #8105.

## The gap

Every numeric proof `type_analysis::is_numeric_expr` had for a bare `LocalGet`
is either an integer-range fact (`integer_locals`, `unsigned_i32_locals`,
`int_valued_i64_locals`) or `FnCtx::stable_local_type_proof` — and that one
answers `None` the moment the local is written a second time. A plain
fractional accumulator therefore had **no numeric proof at all**, so
`expr/binary.rs`'s "both operands are statically primitive" test failed and
every `*` / `/` / `%` / `-` / `**` on it bailed to the BigInt-aware dynamic
helper.

`benchmarks/suite/15_mandelbrot.ts`'s inner loop is the pure case:

```ts
let x = 0.0; let y = 0.0;
while (x * x + y * y <= 4.0 && iter < MAX_ITER) {
    const xtemp = x * x - y * y + cx;
    y = 2.0 * x * y + cy;
    x = xtemp;
    iter = iter + 1;
}
```

`x` and `y` are reassigned, so all six multiplies emitted
`call double @js_dynamic_mul`. `+` and `-` stayed inline because a
`Binary { op: Mul, .. }` operand IS numeric by the recursive rule — which is
why the symptom read as *"only the multiplies escape"*. Wrapping the loop in
`mandel(w: number, h: number, maxIter: number)` does not help: even the fully
typed `$spec_b_b_i32` clone kept the calls, because the gap is about the BODY
LOCAL, not the signature.

## The fix

Reuse the proof that already exists. `collectors/ptr_shape_numeric.rs`'s #7770
locals fixpoint answers *"is every write into this local number-producing by
construction, per spec?"* — structurally, never from a declared type — and it
is already trusted for a strictly harder claim: it licenses a bare
`load double` on a proven numeric FIELD with no coercion and no value check.

A JS Number's Perry representation IS its raw double (numbers carry no NaN-box
tag), so that proof is exactly "the f64 in this slot is a canonical double",
which is what both halves of the numeric fast path need — skipping the dynamic
helper AND skipping the residual `js_number_coerce`.

`collectors/number_by_construction.rs` exposes the fixpoint outside the
`Ptr<Shape>` pass (deliberately independent of `PERRY_PTR_SHAPE_LOCALS`: the
fact is about locals, not shape promotion), `hir_facts` computes it once per
region, and `is_numeric_expr`'s `LocalGet` arm consults it.
`PERRY_NUMBER_BY_CONSTRUCTION=0` empties the fact; both the build-level probe
and the object cache key it.

### Fail-closed by construction

Candidates are the ids with a `Stmt::Let` in the scanned body, so:

* a **parameter** is never admitted — its incoming value is unconstrained and
  a read can precede the first assignment;
* a local **captured from an enclosing scope** is never admitted — its writes
  live outside this walk;
* closure-boxed locals and module globals are excluded outright;
* `let x;` with no initialiser is `undefined`, which is not a Number, so the
  local drops.

## Measured

macOS arm64, `--profile perry-dev` compiler and
`libperry_{runtime,stdlib}.a`, `PERRY_RUNTIME_DIR` pinned,
`PERRY_NO_AUTO_OPTIMIZE=1`, `/usr/bin/time -l`, medians of 5. Both arms are
ONE compiler under `PERRY_NUMBER_BY_CONSTRUCTION` on/off, so the runtime
archives are literally the same bytes. Outputs identical on every program.

| benchmark | instructions ON | instructions OFF | delta | peak RSS ON | peak RSS OFF |
| `15_mandelbrot` | 211,541,019 | 1,709,580,209 | **-87.6%** | 2,163,096 | 2,228,656 |
| `06_math_intensive` | 375,839,338 | 1,976,950,801 | **-81.0%** | — | — |
| `13_factorial` | 1,026,942,694 | 4,228,008,602 | **-75.7%** | — | — |

Both objectives move the right way: `15_mandelbrot` is -87.6% instructions AND
-2.9% peak memory footprint. Nothing regresses on footprint.

The other **27** programs in `benchmarks/suite` are within ±0.5% (noise) with
identical output modulo the elapsed-ms lines they print. Full table in the
issue comment.

`15_mandelbrot` goes from 214 instructions per innermost iteration to 26 —
which finally makes #7128's "one basic block of ~13 instructions" arithmetic,
and its ±2-instruction `collectors/repsel_benefit.rs` cost model, describe a
loop that exists.

## Validation

* `benchmarks/suite` A/B — 30 programs, both arms, outputs compared (above).
* A 260-file sample of `test-files/` compiled and run under both settings:
  **zero** behavioural differences. (Sample interrupted at ~40 files by a
  shared-host disk-space floor; the four `NODE_DIFF`s it did surface are
  pre-existing gaps — they are identical on both arms.)
* `cargo test -p perry-codegen --lib type_analysis::numeric` — 13 passed.
* Every gate in the `lint` job, enumerated from `.github/workflows/test.yml`,
  including `check_file_size.sh` and `local_binding_type_audit.py`.

## Tests

Three, and they are a PAIR plus a second sabotage arm on purpose: the positive
test alone would also pass against an analysis that admits *every* reassigned
local, which is precisely the wrong-code shape #7773 shipped.

1. `reassigned_number_accumulator_multiply_is_an_inline_fmul` — the mandelbrot
   shape emits `fmul double`, no `js_dynamic_mul`, no `js_number_coerce`.
2. `a_reassigned_local_seeded_from_a_parameter_keeps_the_dynamic_helper` — an
   `Any` parameter seed must keep #5970's routing.
3. `a_reassigned_local_written_from_a_string_keeps_the_dynamic_helper` — one
   non-Number write drops the whole local.

## Not covered

x86-64; `--profile dist`; the `--release` runtime (the effect is in generated
code, and #8105 already measured perry-dev and release within 0.2% on this
benchmark). No version bump — maintainer bumps at merge.


proggeramlug:
Independent remeasure on today's `main`, for the record — one claim reproduced, one should not be quoted.

**Instructions: reproduced.** `15_mandelbrot` **−86.93%** against the claimed −87.6% (1.6091 G → 0.2103 G), best-of-5, arms interleaved, stdout sha256 identical. Excellent result and it holds.

**Peak RSS −2.9%: did not reproduce, and is inside the noise floor.** I measure −1.36% and −0.74% in two separate windows. `15_mandelbrot` is a ~2.2 MB binary whose run-to-run peak-RSS range is ±1.5%, so nothing at that magnitude is resolvable there. The direction agrees; the number should not be cited.

Not a criticism of the change — the instruction win is the point and it is large and real. Flagging only so the −2.9% does not get carried forward into a later comparison as an established baseline, which is how a noise reading becomes a phantom regression for the next person.

Method note that may be useful elsewhere: my first noise-floor pass used best-of-3 and reported **±1.93%** between two byte-identical binaries. Best-of-5 brings it to ±0.15%. Any claim on the small binaries at 3 runs is unresolvable.

Issue #8155, opened 2026-08-15
CI: `ext-link` red on every PR — perry-ext-fetch fails to link (undefined js_file_new / js_headers_init_from_value)


proggeramlug:
`ext-link` is failing on every PR based on current `main`, and the failure is main-side rather than any one PR's.

**Symptom** — `cargo test --release --no-run` for `perry-ext-fetch` fails to link:

```
perry_runtime...global_this_file_thunk+0x80:     undefined reference to `js_file_new'
perry_runtime...global_this_headers_thunk+0x78:  undefined reference to `js_headers_init_from_value'
perry_runtime...global_this_request_thunk+0x255: undefined reference to `js_headers_init_from_value'
perry_runtime...global_this_response_thunk+0x13c: undefined reference to `js_headers_init_from_value'
collect2: error: ld returned 1 exit status
error: could not compile `perry-ext-fetch` (lib test) due to 1 previous error
```

Some runs also show `js_blob_new` and `js_fetch_notify_signal_aborted`.

**Evidence that it is not a single PR's fault:** the identical failure appears on unrelated branches — e.g. `perf/issue-8079-guarded-param-specialization` (#8079's branch) and `fix/moving-gc-rooting-sweep` (#8131), whose diffs share no files. Of the open PRs sampled, four fail `ext-link` and one older one (based on an earlier `main`) passes.

**Why it matters:** `ext-link` exists precisely because `cargo-test`'s scope keeps `perry-ext-*` out of the fan-out (#7656, after #7650). A gate that is red for every PR is one that no PR can be blocked by in practice — the same shape as #8092.

`perry-runtime`'s `global_this` fetch thunks reference symbols that live outside it, so the fix is presumably to make `perry-ext-fetch`'s link surface provide them (or to gate the thunks on the same feature that provides the definitions).

proggeramlug:
## Root cause, and one approach I tried and rejected

Reproduced locally: `cargo test -p perry-ext-fetch --no-run` fails with exactly the three symbols reported — `js_file_new`, `js_headers_init_from_value`, `js_fetch_notify_signal_aborted`.

### Why it happens

`crates/perry-ext-fetch/Cargo.toml`'s **dev-dependency** builds the runtime as:

```toml
perry-runtime = { workspace = true, features = ["default", "external-fetch-symbols", "stdlib"] }
```

Those two features are individually correct and jointly fatal for this crate's own test binary:

[...]

PR #8165, opened 2026-08-15
feat(codegen): admit class-typed parameters into the guarded specialization path

### LINKED ISSUE #8099: Guarded parameter specialization refuses class-typed parameters, leaving the runtime class-identity check dead
## Summary

`crates/perry-codegen/src/codegen/param_guard.rs::build_named` refuses class types outright, so
a parameter annotated with a class never gets a runtime descriptor and never enters a
proof-bearing `$spec_` clone:

```rust
} else if self.classes.contains_key(name) && self.class_ids.contains_key(name) {
    // Class identity alone cannot prove mutable field values, while
    // compact instances do not expose the ordinary `keys_array`
    // needed for read-only field validation. Keep class parameters on
    // the generic path until a layout-aware field guard exists.
    return None;
```

This is a deliberate scope boundary from #8094, not an inherent limit, and it is measurable.

## What it costs, measured

In the `tree` benchmark (`/private/tmp/perry-current-sweep-artifacts/sources/tree.ts`) the whole
program is three functions:

```ts
function build(depth: number): Tree
function count(t: Tree): number     // the hot recursive one — t.left, t.right
function main(): void
```

`count` is the hot function and does all the field reads, and its only parameter is class-typed,
so it is refused. The `Tree` constructor is refused at one remove: its parameters are
`Tree | null`, and `Type::Union` maps every variant through `build_type`, so the `Named("Tree")`
variant poisons the union. `build(depth: number)` ends up the only guardable parameter in the
program.

Against the branch's own base `601a02d23`, `tree` recovered 55.1% where an unsound
annotation-trusting control recovered 72.0% — the only row in the sweep where the guarded
design visibly left recovery on the table; every other row matched the control to within half
a percent.

## The runtime half is already written, and is currently dead code

`crates/perry-runtime/src/param_type_guard.rs:399` already implements the class-identity check:

```rust
if class_id != 0
    && !crate::object::class_chain_reaches((*object).class_id, class_id)
{
    return false;
}
```

Codegen always emits `class_id: 0` (both `GuardNode::Object` constructions pass `class_id: None`),
so this branch has never executed. Per the repo's own kill-policy — an unexercised mode is a
configuration nobody has verified — it should either get a caller or be deleted.

## What a fix has to answer

1. **Field validation vs. compact instances.** The `OP_OBJECT` arm validates each descriptor
   field through `own_data_field`, which reads `keys_array`; a compact class instance does not
   expose one. An identity-only descriptor (`class_id: Some(id), fields: vec![]`) sidesteps this,
   but then the guard proves class identity and nothing about field *values*.
2. **What identity actually licenses.** Layout follows from the class, so a slot index and a
   raw-`f64` field read are sound given identity. Trusting a field's *declared type* is not —
   that is the #8033 hazard.
3. **Subclasses.** `class_chain_reaches` admits a subclass; confirm subclass layout preserves the
   parent's field indices.
4. **`delete`.** A `delete` mints a shape word and zeroes `class_id`, so the guard fails and the
   generic path is taken. That looks like it falls the right way, but should be pinned by a test.
5. **The #8094 lifetime rule applies here too.** A class-typed parameter is reference-like, so
   whatever proof it gets cannot survive a call — see the aliasing fix in #8094.

Deliberately kept out of #8094: it widens the guard surface across every class-typed parameter
in every program, and that PR needs to be correct first.

Refs #8079, #8094.

proggeramlug:
Closes #8099.

`param_guard.rs::build_named` refused every class type, so a class-annotated
parameter never entered a proof-bearing clone — and `param_type_guard.rs`'s
`class_chain_reaches` branch had **no caller at all**, because codegen emitted a
literal `class_id: 0` for every object node. Per the repo's kill policy, that
branch either needed a caller or needed deleting.

## What changed

A class descriptor now carries the class id **and** every declared field on the
inheritance chain, validated by name. A class-annotated parameter recovers the
same lowering an interface-annotated one already got, while a structurally
identical object literal fails the identity check and takes the generic
fallback.

```
describeSized$spec_b : ... js_get_string_pointer_unified, js_string_concat_value
describeSized$generic: ... js_dynamic_string_or_number_add
```

## The refusal's stated reason was not true

It said compact class instances expose no ordinary `keys_array` to validate
against. They do: `object_alloc_class_inline_keys_impl` installs a per-class
array that codegen builds once at module init (`js_build_class_keys_array`), so
`own_data_field` resolves a class instance's fields exactly as it resolves a
literal's. The claim traces to a stale doc comment on
`ObjectHeader::keys_array`, which this PR corrects — leaving it in place is how
the next reader repeats the mistake.

## Identity alone was implemented, measured, and reverted

The obvious cheap form — `class_id` with an empty field list — is in the history
of this branch, not in the diff, because it was measured:

| bench | main | identity-only | delta |
|---|---:|---:|---:|
| `tree` | 1.0887 s | 1.6460 s | **+51%** |
| `tree_wide` | 1.7753 s | 2.3039 s | **+30%** |

best-of-5, quiet M1 mini, all outputs byte-exact. Every other corpus row was
flat.

The mechanism, from `--trace llvm`: `count$spec_b` came out **structurally
identical** to the `$generic` sibling it routes around — 325 lines, same 11-call
multiset — because a class-annotated receiver already reaches the class-field
guard path without any parameter evidence. So the clone bought nothing and cost
one `js_param_type_guard` call on each of ~21 M invocations. The field VALUE
facts are the whole payload, which is why they are not optional in the shipped
form.

## This answers the `tree` row #8099 was filed about

`tree`'s hot function is `count(t: Tree)`, and it is refused by **#8094's
aliasing rule**, not by the class refusal: `Named("Tree")` is reference-like and
the body contains a call. Admitting class descriptors does not change that, and
the only descriptor cheap enough to admit there is the identity-only one
measured above. A field-validating descriptor for a recursive class would walk
the reachable object graph on every call — O(nodes x depth) for the 262 143-node
tree — so the aliasing rule is also what keeps the cost bounded here, with no
new policy needed.

## Refused conservatively

Generic classes (fields still `T`), a native / dynamic / unresolvable base
(HIR cannot see its fields), a computed-key field (its `name` is a synthetic
placeholder), a private field, and an accessor sharing a declared field's name
(the licensed read would run user code). Chain walks are cycle-guarded like the
others in this crate.

## Validation

- **Zero corpus impact, proven at the IR level rather than by timing**: all 19
  programs in the specialization corpus emit **identical LLVM IR** before and
  after, and all 19 stay byte-exact against their recorded expected output.
- **Per-call guard cost, the risk this change actually carries**: a wide class
  is validated field-by-field on every call, so a hot function taking one could
  pay more than it saves. Measured with a 12-field class scored 3 M times
  through a non-hoistable receiver (`pool[x & 3]`, since a loop-invariant
  receiver hoists the call and measures nothing): **0.4142 s -> 0.4121 s**,
  best-of-5 on the quiet mini, output byte-exact. `--trace llvm` confirms the
  subject was live — `score$spec_b` and one guard site in the new arm, no clone
  at all in the base arm. Flat, so no field-count cap is proposed; inventing one
  without a measurement showing it is needed is how unexercised policy gets in.
- **Byte-exact vs node `v26.5.1`**: `test_gap_specabi_ordinary_param_guards`
  gains rows for good / lying-fields / structural-literal / subclass /
  accessor / recursive-class / aliased-through-a-global. `--trace llvm`
  confirms the positives are live (`describeSized$spec_b` emitted, 9 guard
  sites) and the negatives really are negative (`chainTotal`,
  `describeThroughGlobal`, `mutatePayload`, `treeTotal` get no clone).
- **Moving GC**: the same fixture is byte-exact under
  `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1`, with `copying_minors=40` and
  40 `[gc-fromspace-protect]` lines — the instrument ran, so the green is not
  the "zero copying minors" kind.
- **Gap suite** (`run_gap_tests.sh`, all 561 `test_gap_*`): 545 pass, 97.1%.
  The harness flagged two regressions; **both are artifacts, and each was A/B'd
  against a compiler built from this branch's exact base rather than argued
  away**:
  - `test_gap_7238_i64_specialization_exactness` (`pass -> crash`) — exits 0
    five times out of five standalone with the branch compiler. The run was on
    a box at load 30-50 from other builds; the harness classified a timeout as
    a crash.
  - `test_gap_webcrypto_async_threadpool` (`pass -> parity_fail`) — Perry's
    output is **byte-identical between the base and branch compilers**, and
    both match node once the `MODULE_TYPELESS_PACKAGE_JSON` warning (a cwd
    artifact) is stripped. Pre-existing/environmental, not this branch.

  The ten `node_fail -> parity_fail` status changes are tests whose snapshot
  entry records the *oracle* failing; node runs them here. A compiler change
  cannot move that classification either way.

  Caveat worth stating: this was one run on a contended host, so rather than
  claim a clean sweep I A/B'd exactly the tests the harness named — the
  targeted equivalent of a second full run, at a fraction of the wall clock.
- **`cargo test -p perry-codegen`**: 1401 pass, and the failing set is
  **identical, name for name, to a clean-`main` baseline run** of the same
  command — 9 failures before, the same 9 after:
  `loop_safepoint_purity::proven_numeric_counted_loop_emits_no_back_edge_poll`,
  six in `native_proof_buffer_views`,
  `shadow_slot_hygiene::canonical_str_local_keeps_shadow_binding_and_tag_dispatched_ops`,
  and `typed_feedback::typed_feedback_guards_direct_class_field_specialization`.
  All 9 are in `crates/perry-codegen/tests/*.rs`, which CLAUDE.md notes do not
  run per-PR — the "landed green, sat red" shape. Not this branch's to fix, but
  worth someone's attention.

  Three tests DID break on this branch and were investigated rather than
  silenced; two of them (`guarded_pshape_call_site_is_preceded_by_a_shape_id_guard`,
  `sloppy_class_field_pointer_store_takes_the_inline_boxed_store`) turned out to
  be anchoring artifacts of the identity-only prototype and pass untouched in
  the shipped form. The third is the one updated below.
- `cargo fmt --all --check`, `check_file_size.sh`, `addr_class_inventory.py`,
  `local_binding_type_audit.py` all clean.

## One test updated, and why

`annotated_class_method_value_uses_generic_lookup` asserted that a class-typed
parameter selects no direct class-method bind ABI *anywhere in the module*. It
now asserts the #8094 split instead: the `$generic` body keeps
`js_object_get_field_ic_miss` and must **not** contain the direct bind, while
the guarded clone must. The #8033 invariant it exists for — an erased
annotation is never a proof — is unchanged and is now checked on the body that
can actually be reached without a validated argument, plus a vacuity guard so
the assertions cannot pass by the clone quietly disappearing.

No version bump, no `CHANGELOG.md` edit; the entry is
`changelog.d/8165-class-typed-parameter-guards.md`.

## Summary by CodeRabbit

- **New Features**
  - Added optimized handling for functions with class-typed parameters, including inherited fields and class identity validation.
  - Valid class instances can use specialized execution, while incompatible or structurally similar objects safely use generic behavior.

- **Bug Fixes**
  - Improved handling of class keys and parameter guards.
  - Preserved correct behavior for accessors, recursive classes, aliases, and modified fields.

- **Tests**
  - Added comprehensive coverage for specialization, inheritance, fallback behavior, and generated code paths.

- **Documentation**
  - Clarified class-instance key handling and parameter guard behavior.

### DIFF
--- a/changelog.d/8165-class-typed-parameter-guards.md
+++ b/changelog.d/8165-class-typed-parameter-guards.md
@@ -1,0 +1,42 @@
+### Performance
+
+- Admit class-typed parameters into the #8094 guarded specialization path. A
+  class descriptor now carries the class id — giving
+  `param_type_guard.rs`'s `class_chain_reaches` branch its first caller, which
+  codegen had never reached — plus every declared field on the inheritance
+  chain, validated by name. A class-annotated parameter recovers the same
+  lowering an interface-annotated one already got (`js_dynamic_string_or_
+  number_add` becomes a string concat), while a structurally identical object
+  literal fails the identity check and takes the generic fallback (#8099).
+
+  The refusal this replaces rested on a stale claim that compact class
+  instances carry no `keys_array`. They do —
+  `object_alloc_class_inline_keys_impl` installs a per-class array built once
+  at module init — so the same by-name field validation that serves interfaces
+  serves classes. The stale note on `ObjectHeader::keys_array` is corrected
+  too.
+
+  Identity **without** the field types was implemented, measured and reverted:
+  the emitted clone came out structurally identical to the `$generic` sibling
+  it routes around (same line count, same call multiset), because a
+  class-annotated receiver already reaches the class-field guard path without
+  any parameter evidence. It bought nothing and cost one `js_param_type_guard`
+  call per invocation — `tree` 1.089 s → 1.646 s (+51%) and `tree_wide`
+  1.775 s → 2.304 s (+30%), best-of-5 on the quiet M1 mini. That also answers
+  the `tree` row #8099 was filed about: the hot recursive walker is refused by
+  #8094's aliasing rule, not by the class refusal, and the only descriptor
+  cheap enough to admit there is the one that buys nothing.
+
+  Cost stays bounded by the existing rule rather than a new one: a
+  field-bearing descriptor claims heap contents, so a reference-typed
+  parameter carrying one is refused in any body containing a call. A recursive
+  class (`Tree.left: Tree`) therefore cannot be guarded inside the recursive
+  walker that would make its validation O(nodes x depth).
+
+  Validated: the 19-program specialization corpus emits **identical LLVM IR**
+  before and after, so nothing already-specializing moved; the extended
+  `test_gap_specabi_ordinary_param_guards` fixture is byte-exact against node
+  `v26.5.1`, including under `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1`
+  (40 copying minors, 40 from-space protections — the instrument was live);
+  and `cargo test -p perry-codegen` adds no failure against a clean-`main`
+  baseline run.
--- a/crates/perry-codegen/src/codegen/ordinary_param_guard_tests.rs
+++ b/crates/perry-codegen/src/codegen/ordinary_param_guard_tests.rs
@@ -565,3 +565,182 @@ fn guarded_discriminant_branch_narrows_a_union_parameter_inside_the_clone() {
         "the recursive walker must stay on the unguarded body:\n{recursive}"
     );
 }
+
+/// #8099: a CLASS-typed parameter is guarded exactly like an interface-typed
+/// one, and the pair of bodies is the subject.
+///
+/// The refusal this replaces (`param_guard.rs::build_named`) said compact class
+/// instances expose no `keys_array` to validate declared fields against. They
+/// do — `object_alloc_class_inline_keys_impl` installs a per-class array built
+/// once at module init — so the same by-name field validation that serves
+/// interfaces serves classes, plus a `class_chain_reaches` identity check no
+/// structural type can satisfy.
+///
+/// The `Named("Label")` receiver in the clone is what turns the dynamic add
+/// into a string concat. Identity WITHOUT the fields was tried first and
+/// reverted: with an empty field list the clone came out structurally
+/// identical to the `$generic` sibling it routes around (same line count, same
+/// call multiset), because a class-annotated receiver already reaches the
+/// class-field guard path without any parameter proof. Asserting the two
+/// bodies DIFFER is therefore not decoration — it is the only thing that
+/// distinguishes this from a clone that costs a guard call and buys nothing.
+#[test]
+fn a_class_parameter_is_guarded_by_identity_and_declared_fields() {
+    let label = perry_hir::Class {
+        id: 41,
+        name: "Label".to_string(),
+        type_params: Vec::new(),
+        extends: None,
+        extends_name: None,
+        native_extends: None,
+        extends_expr: None,
+        heritage_lexically_shadowed: false,
+        fields: vec![
+            perry_hir::ClassField {
+                name: "label".to_string(),
+                key_expr: None,
+                ty: Type::String,
+                init: None,
+                is_private: false,
+                is_readonly: false,
+                decorators: Vec::new(),
+            },
+            perry_hir::ClassField {
+                name: "count".to_string(),
+                key_expr: None,
+                ty: Type::Number,
+                init: None,
+                is_private: false,
+                is_readonly: false,
+                decorators: Vec::new(),
+            },
+        ],
+        constructor: None,
+        methods: Vec::new(),
+        getters: Vec::new(),
+        setters: Vec::new(),
+        static_accessor_names: Vec::new(),
+        static_accessor_fn_ids: Vec::new(),
+        static_fields: Vec::new(),
+        static_methods: Vec::new(),
+        computed_members: Vec::new(),
+        decorators: Vec::new(),
+        is_exported: false,
+        is_nested: false,
+        alloc_width_hint: 0,
+        specialized_from: None,
+        aliases: Vec::new(),
+    };
+    let render = Function {
+        id: 42,
+        name: "renderLabel".to_string(),
+        type_params: Vec::new(),
+        params: vec![Param {
+            id: 420,
+            name: "payload".to_string(),
+            ty: Type::Named("Label".to_string()),
+            default: None,
+            decorators: Vec::new(),
+            is_rest: false,
+            arguments_object: None,
+        }],
+        return_type: Type::Any,
+        // `result` is `any` and reassigned, so it carries no proof of its own
+        // and the ADD is what the parameter proof has to reach through. A
+        // simpler `payload.label + "!"` is measurably vacuous here: the
+        // class-field typed-feedback path already resolves that one without
+        // any parameter evidence, and the two bodies come out identical.
+        body: vec![
+            Stmt::Let {
+                id: 421,
+                name: "result".to_string(),
+                ty: Type::Any,
+                mutable: true,
+                init: Some(Expr::PropertyGet {
+                    object: Box::new(Expr::LocalGet(420)),
+                    property: "label".to_string(),
+                    byte_offset: 0,
+                }),
+            },
+            Stmt::Expr(Expr::LocalSet(
+                421,
+                Box::new(Expr::Binary {
+                    op: BinaryOp::Add,
+                    left: Box::new(Expr::Binary {
+                        op: BinaryOp::Add,
+                        left: Box::new(Expr::LocalGet(421)),
+                        right: Box::new(Expr::String(":".to_string())),
+                    }),
+                    right: Box::new(Expr::PropertyGet {
+                        object: Box::new(Expr::LocalGet(420)),
+                        property: "count".to_string(),
+                        byte_offset: 0,
+                    }),
+                }),
+            )),
+            Stmt::Return(Some(Expr::LocalGet(421))),
+        ],
+        is_async: false,
+        is_generator: false,
+        is_strict: true,
+        is_exported: false,
+        captures: Vec::new(),
+        decorators: Vec::new(),
+        was_plain_async: false,
+        was_unrolled: false,
+    };
+    let mut module = Module::new("class_param_guard.ts");
+    module.classes.push(label);
+    module.functions.push(render);
+    module.init.push(Stmt::Expr(Expr::Call {
+        callee: Box::new(Expr::FuncRef(42)),
+        args: vec![Expr::Undefined],
+        type_args: Vec::new(),
+        byte_offset: 0,
+    }));
+
+    let opts = CompileOptions {
+        emit_ir_only: true,
+        output_type: "executable".to_string(),
+        ..Default::default()
+    };
+    let ir = String::from_utf8(compile_module(&module, opts).expect("module compiles"))
+        .expect("LLVM IR is UTF-8");
+
+    let public = function_ir(&ir, "@perry_fn_class_param_guard_ts__renderLabel(");
+    assert!(public.contains("call i32 @js_param_type_guard("));
+    assert!(public.contains("renderLabel$spec_b("));
+    assert!(public.contains("renderLabel$generic("));
+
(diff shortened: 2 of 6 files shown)


proggeramlug:
`gc-root-dominance-statepoints` is red here and it is **not this branch**.

It dies in setup, before the checker runs:

```
##[error]node_modules/zod/src/index.ts is missing; run npm ci --ignore-scripts
```

The `gc-root-dominance-statepoints` job never runs `setup-node` / `npm ci` —
those two steps live only in its sibling `gc-root-dominance` job, which passes
here. #8084 added `scripts/gc_root_dominance_dep_native_corpus.sh` to the
statepoints job on 2026-08-15T12:25Z, and every `main` run since (`d6d7d0efe`
onward) has failed the same way; the last green ones were the two `fa83ecab2`
runs that morning.

Filed as #8170 with the two-step fix. Nothing in this PR touches that workflow,
the corpus script, or the checker.

proggeramlug:
`ext-link` is the second red check, and it is also not this branch — it is
**#8155**, open and unchanged:

```
undefined reference to `js_fetch_notify_signal_aborted'
undefined reference to `js_blob_new'
undefined reference to `js_file_new'
undefined reference to `js_headers_init_from_value'
```

Same symbols that issue names. The only runtime file this PR touches is a doc
comment on `ObjectHeader::keys_array`; nothing here adds, removes or renames an
FFI symbol.

So both red checks on this PR are pre-existing and separately tracked: #8170
(`gc-root-dominance-statepoints`, red on `main` since #8084) and #8155
(`ext-link`).

proggeramlug:
Third red check, also pre-existing: `Warnings (product)`.

```
error: unused doc comment
error: could not compile `perry-runtime` (lib) due to 1 previous error
```

`crates/perry-runtime/src/gc/roots/stack_maps.rs:209` has a `///` block directly
above a `crate::perry_thread_local! { … }` **macro invocation**, which rustc
does not attach doc comments to. The job runs `-D warnings`, so it errors there;
elsewhere it is an ordinary warning you can see in any local
`cargo build -p perry-runtime`. It landed with #8084 and no `main` run of
`test.yml` has happened since that merge, so it has only ever surfaced on PRs.

Filed as #8176. This PR's only `perry-runtime` change is a doc comment on
`ObjectHeader::keys_array`, ~1500 lines away in a different file.

Running tally of red checks here, none of them this branch:

| check | tracked as | cause |
|---|---|---|
| `gc-root-dominance-statepoints` | #8170 | statepoints job never runs `npm ci`, so the zod corpus is missing |
| `ext-link` | #8155 | `perry-ext-fetch` undefined `js_file_new` / `js_headers_init_from_value` |
| `Warnings (product)` | #8176 | unused doc comment on a macro invocation, `-D warnings` |

`cargo-test` — the required context that actually runs this PR's new unit tests
— is green.

proggeramlug:
`conformance-smoke (2)` is red on one claimed regression,
`test_gap_new_globalthis_builtin_6726: pass -> parity_fail`, plus
`test_gap_backoff_options: node_fail -> parity_fail`.

**#8117 (open) names both of those tests by name** as part of "`conformance-smoke`
fails 8/8 shards". So this is the tracked breakage, not this branch. Expect the
other shards to go the same way.

A/B'd anyway rather than pattern-matching on the issue title — compilers built
from this branch and from its exact base, same runtime archives:

```
test_gap_new_globalthis_builtin_6726
  base vs new : IDENTICAL output
  base vs node: MATCH
  new  vs node: MATCH
```

It also passed in the local full 561-test gap run on this branch (545 pass,
97.1%), whose own two flagged items were separately shown to be a timeout
misclassified as a crash and a cwd artifact.

Updated tally — four red checks, none of them this branch:

| check | tracked as |
|---|---|
| `gc-root-dominance-statepoints` | #8170 (filed here) |
| `ext-link` | #8155 |
| `Warnings (product)` | #8176 (filed here) |
| `conformance-smoke (2)` | #8117 |

proggeramlug:
Two more red checks; both pre-existing, and both checked against `main` rather
than asserted.

### `compiler-output-regression`

Failing on **`main` for at least four consecutive days** — `83b6b8c69`,
`601a02d23`, `59cf9d9c9`, `c4b2c1c8e` (2026-08-12 through 08-15) all red.

Comparing the `failed_workloads` list, this PR's is a strict **subset** of
`main`'s last recorded run:

```
PR #8165 (4)                    main 83b6b8c69 (11)
  h1_native_rep_equivalence       h1_native_rep_equivalence
  image_convolution               image_convolution
  loop_data_dependent             loop_data_dependent
  numeric_arrays                  numeric_arrays
                                  h1_buffer_alias_negative
                                  packed_f64_loop_versioning
                                  packed_f64_loop_versioning_negative
                                  dynamic_fractional_array_index
                                  loop_bound_semantics
                                  raw_numeric_object_fields
                                  scalar_replacement_literals
```

`comm -23 pr main` is empty: no workload fails here that does not already fail
on `main`.

### `e2e-scoped`

Fails on exactly four suites:

```
perry-codegen --test loop_safepoint_purity
perry-codegen --test native_proof_buffer_views
perry-codegen --test shadow_slot_hygiene
perry-codegen --test typed_feedback
```

Those are **precisely the four suites** (9 tests) that fail in a clean-`main`
`cargo test -p perry-codegen` run, reported in the PR description. `e2e-scoped`
scopes integration suites to changed packages, so touching `perry-codegen` is
what makes this long-standing red visible — the suites themselves do not run
per-PR otherwise, which is the CLAUDE.md "landed green, sat red" shape.

Note what did **not** fail: `perry-codegen --test native_proof_regressions`,
which is the only integration suite this PR edits. It passed.

Full tally — six red checks, none of them this branch:

| check | tracked as |
|---|---|
| `gc-root-dominance-statepoints` | #8170 (filed here) |
| `ext-link` | #8155 |
| `Warnings (product)` | #8176 (filed here) |
| `conformance-smoke (2)` | #8117 |
| `compiler-output-regression` | red on `main` since at least 08-12 |
| `e2e-scoped` | the 9 pre-existing `perry-codegen` integration failures |

proggeramlug:
**13 checks on this run never reported** — they were cancelled mid-flight, not
run and not failed:

```
Clippy (product)          conformance-smoke (1,3,4,5,6,8)
Warnings (host-compatible) gc-stress
api-docs-drift            lint
changeset-gate            native-backend
```

Three of those (`lint`, `api-docs-drift`, `changeset-gate`) are required
contexts, so this PR currently has no verdict from them.

This is not a supersede. Run `31898931771` is still the only live `Tests` run on
the ref — there is no newer one, no new push (head has been `6da16b489` since
17:26Z), and no label event. GitHub cancelled the jobs at ~22:22-22:23Z after
some had been executing for over three hours (`changeset-gate` started 18:42Z),
while the run itself is still reported `status: queued` with three `*-complete`
aggregators pending. `gh run rerun` refuses with "This workflow is already
running", so it needs either a cancel-then-rerun or a fresh trigger.

I have not forced that, because re-running costs hours of shared runner capacity
and the queue is clearly already the binding constraint here — maintainer's
call.

Worth noting for the concurrency design: `test.yml`'s PR group is
`test-pull_request-${{ github.ref }}` with `cancel-in-progress: true`, which is
the carve-out #7966 deliberately kept for PRs. That is not what happened here,
but it does mean any subsequent push to this branch will cancel whatever is
mid-flight again, so a re-run is best done when the branch is settled.

Locally, the gates these would have covered were run against this exact tree:
`cargo fmt --all --check`, `scripts/check_file_size.sh`,
`scripts/addr_class_inventory.py`, `scripts/local_binding_type_audit.py` — all
clean — plus the full 561-test gap suite and `cargo test -p perry-codegen`.

PR #8187, opened 2026-08-16
ci: restructure CI into PR gate / main sweep / full suite tiers

proggeramlug:
## Why

CI was not gating anything and could not have. Measured 2026-08-16 (last ~24 h of runs, `gh api`):

| fact | measured |
|---|--:|
| org plan / concurrent hosted jobs | **Free / 20 (5 macOS)** — for the whole org |
| workflows × jobs × runner-min fanned out **per PR push** | **14 × 48 × ~650 min** |
| PR pushes/day · merges/day | ~66 · ~58 |
| median job queue wait (`Tests` jobs) | **3–4.5 h**, max 7 h |
| `Tests` PR runs that reached a conclusion | **0 of 66** (8 failed, 56 cancelled by the next push) |
| last 12 merges with `lint`/`cargo-test` actually run | **0** — all `cancelled`; `conformance-smoke-complete` red; `parity`/`compile-smoke` (required!) `skipped` |
| `conformance-smoke` cost per push | 8 shards × ~60 min = **480 job-min**; 96 % of a shard's test time is ~10 tests × ~200 s = auto-optimize rebuilding a feature-stripped runtime per feature set, redundantly per shard |
| sccache writes | every job saved a ~0.5–1.3 GB tarball per PR push (~200 GB/day into a 10 GB budget); PR-scoped caches are unreadable by other PRs anyway |
| `cache-warm.yml` last success | 2026-07-28 (self-cancelled by `cancel-in-progress` under 58 merges/day) |

Demand was 1.5–2× capacity, so the queue never drained, every merge was an admin bypass, and regressions landed (#8117, #8092, #8155, #8156, the compiler-output-regression reds).

## What

**One workflow, three tiers, one policy file.** `scripts/ci_plan.py` (self-tested, `--table` prints the matrix, `lint` checks the docs copy) decides what a run of `test.yml` executes; every job is `needs: plan` + a plan flag; a single fan-in job carries the verdict.

| tier | trigger | jobs (typical) | fan-in |
|---|---|--:|---|
| **pr** | every PR push | ~11 / ~140 job-min: `lint` (+changeset step), `check` (clippy ×2 + api-docs-drift), `warnings`, scoped `cargo-test`, **4-shard fast-mode gap suite**, `gc-stress` (PR subset), `e2e-scoped`, `security-audit` only when a lockfile/manifest changed. Docs-only PRs: `lint` only. | **`pr-gate` — the only required context** |
[...]

Issue #8156, opened 2026-08-15
CI: `native-roots-rs4gc` red on main (ELF) — provider fixture aborts with "stdlib provider is bound to a different runtime image"


proggeramlug:
`gc-native-roots` / `native-roots-rs4gc` has been failing on `main` (ELF arms) since at least 2026-08-14, and consequently on every PR branched from it.

**Symptom** — the `issue_8075_provider_gc` fixture host aborts before it runs any handler:

```
Error: "stdlib provider is bound to a different runtime image"
```

That check is `tests/fixtures/issue_8075_provider_gc/host.rs`: the stdlib provider exports a probe returning the address of `js_gc_init` **as resolved inside the stdlib image**, and the host compares it with the address it resolved from the runtime image. A mismatch means the stdlib provider bound its own bundled copy of `perry-runtime` instead of the shared one — i.e. the archive trim that is supposed to remove the bundled runtime members no longer removes all of them (or the link order changed so the bundled copy wins).

**Affected runs (main):**
- `83b6b8c69` — failure
- `506f4ab11` — failure
- `5c27d1ad9` — success (last known good)

So the regression landed between `5c27d1ad9` and `506f4ab11`.

**Why it matters:** this is the gate for #8075/#8081's whole subject — native GC roots when Perry runs behind separately loaded runtime and stdlib providers. While it aborts at the provider handshake it never reaches the root-map assertions, so it cannot fail for the reason it exists (CLAUDE.md's fourth "gate cannot fail" mode: the gate runs, its subject never does).

proggeramlug:
Bisect range confirmed and the mechanism measured on real ELF (Linux/aarch64 container, this repo's actual `stdlib-linker.sh` driving a `cargo build --profile provider` of a stand-in provider cdylib, then `host.rs`'s own `dlopen`/`probe()` check).

`5c27d1ad9..506f4ab11` contains exactly one commit — #8089 — and it **causes** the break rather than exposing one. What it added on the ELF side is a hand-written version script ending in `local: *`.

[...]

PR #8197, opened 2026-08-16
fix(ext-fetch): let the crate's own test binary link, restoring the ext-link gate

### LINKED ISSUE #8155: CI: `ext-link` red on every PR — perry-ext-fetch fails to link (undefined js_file_new / js_headers_init_from_value)

proggeramlug:
Closes #8155.

## The failure

`cargo test --no-run -p perry-ext-fetch` cannot link:

```
  "_js_blob_new",                    referenced from: perry_runtime::object::global_fetch::call_global_blob_new
  "_js_fetch_notify_signal_aborted", referenced from: perry_runtime::url::abort::notify_fetch_abort
  "_js_file_new",                    referenced from: perry_runtime::object::global_fetch::call_global_file_new
  "_js_headers_init_from_value",     referenced from: perry_runtime::object::global_fetch::call_global_headers_init_from_value
```

`perry-runtime` is built here with `external-fetch-symbols`, which is correct for every shipped configuration: it declares those four `extern` and *calls* them, on the promise that the final link provides them. In a real binary `perry-stdlib` does (`fetch_blob.rs`, `fetch/abort_bridge.rs`). This crate's test binary links `perry-runtime` and nothing else.

**The cost was not one crate.** `ext-link` is the only gate covering `perry-ext-*`, because `cargo-test`'s scope deliberately keeps that family out of its fan-out (#7656). While it was red on every PR, the whole family was ungated — and `perry-ext-fetch`'s own **13 tests had never run**, because the binary could not be produced. They pass.

## The fix, and what it deliberately is not

Four definitions in a `#[cfg(test)]` module.

**Not real implementations.** This crate is a `staticlib` whose objects win the final link ahead of perry-stdlib (`prefer_well_known_before_stdlib`), so defining them for real would *replace* perry-stdlib's Blob/File/Headers constructors and abort bridge in every shipped binary — a behaviour change smuggled in under a CI fix, and one that would have to reimplement `FETCH_ABORT_WATCHERS` and the blob registry to be equivalent. This crate implements 39 fetch symbols; these four are deliberately not among them.

`#[cfg(test)]` never reaches `libperry_ext_fetch.a`, and perry-stdlib is absent from the test link, so no duplicate symbol can arise. Semantics match the runtime's own `stdlib_stubs.rs`: warn once, return `undefined`.

**Also not the dev-dependency route**, which #8155's analysis had already tried and rejected: adding `perry-stdlib` plus `extern crate` to force its objects in does make the link succeed, but produces `duplicate symbol` diagnostics that macOS `ld` downgrades to warnings — link-order-dependent binding in a test binary, which is precisely the failure shape of #8156.

A guard test asserts that no non-`cfg(test)` file in this crate defines any of the four, so a later move out from behind `cfg(test)` fails here rather than silently overriding stdlib at link time. **Sabotage-verified**: defining `js_blob_new` in a shipped module fails it with the offending path named.

## One thing in the issue's analysis that is a false premise

The issue notes that `stdlib_stubs.rs` has no `js_file_new` stub (nor `js_fetch_notify_signal_aborted`) and reads that as an independent gap. It isn't: both call sites are `cfg`'d on `external-fetch-symbols`, and the non-external arms use the `GLOBAL_FETCH_*` function-pointer registry and a no-op respectively. Nothing references them in the stub configuration, so no stub is owed. Recorded so nobody "fixes" it.

## Relationship to #8187

#8187 makes `ext-link` opt-in on PRs (label `run-extended-tests`) and adds nightly/tag arms, citing this issue as the reason. That stops the noise; it does not make the link work, so the nightly arm would still have been red. The two are complementary — this PR touches only Rust and no workflow file, so they cannot conflict.

## Validation

- `cargo test --release --no-run -p perry-ext-fetch` — the gate's exact command — **links**.
- `cargo check --release -p perry-ext-fetch` (the gate's per-package feature-graph step) clean.
- `cargo test -p perry-ext-fetch`: **13 passed**, up from a crate that could not build a test binary.
- `cargo fmt --check`, `check_file_size.sh` clean; no new clippy warnings (the 4 on this crate are pre-existing).

No version bump.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored successful test-binary builds for the external fetch component.
  * Added safeguards to ensure test-only compatibility code is excluded from shipped builds.
  * Prevented fetch functionality from being unintentionally overridden in production.
  * Improved build reliability and verification for fetch-related functionality without changing the behavior of released applications.

### DIFF
--- a/changelog.d/8197-ext-fetch-test-link.md
+++ b/changelog.d/8197-ext-fetch-test-link.md
@@ -1,0 +1,20 @@
+### Fixed
+
+- `perry-ext-fetch` can build its own test binary again (#8155). `perry-runtime`
+  is compiled here with `external-fetch-symbols`, so it declares `js_blob_new`,
+  `js_file_new`, `js_headers_init_from_value` and `js_fetch_notify_signal_aborted`
+  as `extern` and calls them, on the promise that the final link supplies them —
+  which `perry-stdlib` does in a real binary but nothing does in this crate's
+  test link. `cargo test --no-run` therefore failed to link, which kept the
+  `ext-link` gate red on every PR; since `cargo-test`'s scope deliberately keeps
+  `perry-ext-*` out of its fan-out (#7656), that left the whole ext family
+  ungated, and this crate's own 13 tests had never run.
+
+  The four symbols are now defined in a `#[cfg(test)]` module rather than for
+  real: this crate is a `staticlib` whose objects win the final link ahead of
+  perry-stdlib, so shipping them would silently replace perry-stdlib's
+  Blob/File/Headers constructors and abort bridge everywhere. `cfg(test)` code
+  never reaches `libperry_ext_fetch.a`, and perry-stdlib is absent from the test
+  link, so no duplicate symbol can arise. A guard test asserts no shipped file
+  in the crate defines any of the four, so moving them out from behind
+  `cfg(test)` fails there rather than at a customer's link.
--- a/crates/perry-ext-fetch/src/lib.rs
+++ b/crates/perry-ext-fetch/src/lib.rs
@@ -29,6 +29,12 @@ use validation::{throw_range_error, throw_type_error};
 #[cfg(test)]
 mod test_async_shims;
 
+// Definitions for the four fetch symbols perry-runtime CALLS under
+// `external-fetch-symbols` but perry-stdlib OWNS, so this crate's own test
+// binary links (#8155). Test-only on purpose — see the module docs.
+#[cfg(test)]
+mod test_link_stubs;
+
 const STRING_TAG: u64 = 0x7FFF_0000_0000_0000;
 const POINTER_TAG: u64 = 0x7FFD_0000_0000_0000;
 const TAG_UNDEFINED: u64 = 0x7FFC_0000_0000_0001;
--- a/crates/perry-ext-fetch/src/test_link_stubs.rs
+++ b/crates/perry-ext-fetch/src/test_link_stubs.rs
@@ -1,0 +1,153 @@
+//! Test-binary definitions for the fetch symbols `perry-runtime` CALLS but
+//! this crate does not own (#8155).
+//!
+//! # Why this exists
+//!
+//! `perry-runtime` is built here with `external-fetch-symbols`, which is
+//! correct for every shipped configuration: it declares
+//! `js_blob_new` / `js_file_new` / `js_headers_init_from_value` /
+//! `js_fetch_notify_signal_aborted` as `extern` and *calls* them
+//! (`object/global_fetch.rs`, `url/abort.rs`), on the promise that someone
+//! else in the final link defines them. In a real binary that someone is
+//! `perry-stdlib` (`fetch_blob.rs`, `fetch/abort_bridge.rs`).
+//!
+//! This crate's own test binary links `perry-runtime` and nothing else, so
+//! those calls have no definition and `cargo test --no-run` fails to link —
+//! which is what kept the `ext-link` gate red on every PR while
+//! `cargo-test`'s scope deliberately keeps `perry-ext-*` out of its fan-out
+//! (#7656). The gate for the whole ext family was therefore dark.
+//!
+//! # Why `#[cfg(test)]` rather than a real implementation
+//!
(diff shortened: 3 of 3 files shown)


proggeramlug:
Note on this PR's `self-test-checkers` failure: it is **not** from this change. Four files on `main` declare raw `thread_local!` blocks that #7469's policy rejects (`node_module/source_map.rs`, `node_vm.rs`, `dyn_eval/interp.rs`, `module_require.rs`); this PR touches none of them and only adds a `#[cfg(test)]` module to `perry-ext-fetch`. Fixed separately in #8199, which takes the checker from FAILED to `OK: 224 hot declarations`. Everything else here is green.

proggeramlug:
**Verified locally by running the gate's own commands**, rather than reading CI:

| step (as `ext-link.yml` runs it) | result |
| --- | --- |
| `cargo check --release -p perry-ext-fetch` | clean |
| `cargo check --release -p perry-ext-ws` | clean |
| `cargo check --release -p perry-ext-fastify` | clean |
| `cargo test --release --no-run -p perry-ext-fetch` (fetch feature group) | **links** — was the failure |
| `cargo test --release --no-run -p perry-ext-ws -p perry-ext-fastify` (ws group) | links |
| `cargo test -p perry-ext-fetch` | **13 passed** — these had never run |
| `cargo fmt --check`, `check_file_size.sh` | clean |

The whole gate scope builds and links on this branch. The guard test is sabotage-verified: defining `js_blob_new` in a shipped module fails it and names the offending path.

The `self-test-checkers` failure showing on this PR comes from `main` (four raw `thread_local!` blocks this PR does not touch) and is fixed separately in #8199.

proggeramlug:
Good catch, fixed in `922cdaca1` — and it was the sharper kind of wrong, so worth naming precisely.

The guard scanned `test_link_stubs.rs` for `#[cfg(test)]` and found it: the annotation on the inner `mod tests`. So the assertion passed for a reason unrelated to what it claimed to protect, and deleting the gate in `lib.rs` — the one thing keeping these four symbols out of `libperry_ext_fetch.a` — would have left it green. A test that cannot fail in the direction it exists for.

It now reads `lib.rs` and asserts `mod test_link_stubs;` is immediately preceded by `#[cfg(test)]`. **Sabotage-verified in that exact direction**: removing the gate fails with

```
`mod test_link_stubs;` in lib.rs must be immediately preceded by #[cfg(test)] — without it
these four symbols land in libperry_ext_fetch.a and override perry-stdlib's implementations
at the final link, because ext archives are linked first
```

where the previous version passed. `cargo test -p perry-ext-fetch`: 13 passed; fmt clean.

Issue #8169, opened 2026-08-15
A guarded $spec_b clone never re-enters itself either — the Tier-B half of #8167


proggeramlug:
#8167 fixed the raw-`i32` half of "a specialized clone never re-enters itself". The **Tier-B (`$spec_b`) half is untouched and is the shape most real code hits**, because it does not need a literal call site to exist.

Change `fib40.ts`'s entry call from a literal to a variable and the same function takes the guarded route:

```ts
function fib(n: number): number { return n < 2 ? n : fib(n - 1) + fib(n - 2); }
const k = Number(process.argv[2] ?? "30");
console.log(fib(k));
```

`--trace llvm` on `perf/spec-clone-self-recursion` (i.e. WITH #8167):

```llvm
define internal double @…__fib$spec_b(double %arg1) {
  %r3  = fcmp olt double %r2, 2.0          ; the clone is fast
  %r9  = fsub double %r8, 1.0
  %r10 = call double @…__fib(double %r9)   ; …but re-enters the PUBLIC TRAMPOLINE
  %r13 = call double @…__fib(double %r12)
}
define double @…__fib(double %arg1) noinline { … js_param_type_guard … }
```

So every recursive step pays a `noinline` trampoline call plus a `js_param_type_guard` runtime call, then lands back in the clone it just left.

## Why it happens

`try_emit_spec_guarded_call` routes a site only when `guarded_argument_proves(ctx, arg, expected)` holds, and `guarded_path_type` has no arm for `Expr::Binary` — so a derived argument (`n - 1`) yields `None` and the site keeps the public wrapper. Identical in structure to the `i32` defect #8167 fixed, one tier over.

## Why the fix should be cheaper than #8167's

A `Boxed` slot takes the `double` **verbatim** — a JS Number's NaN-box *is* its double. There is no conversion, so none of #8167's three obligations apply: no 32-bit range test, no exact-integer window, no `-0` hazard. The whole obligation is "this value is a Number", which the codebase already has a predicate for (`type_analysis::is_numeric_expr` — the same one that gates `fadd` versus `js_dynamic_add`, so if it were wrong arithmetic would already be miscompiling).

Sketch: in `guarded_path_type` (`crates/perry-codegen/src/lower_call/func_ref.rs`), fall back to `Type::Number` when `is_numeric_expr(ctx, expr)` holds and no structural arm matched.

## Why it was NOT done in #8167

[...]

PR #8201, opened 2026-08-16
perf(codegen): decide scalar parameter descriptors with the typed-abi leaf guards

proggeramlug:
## What this is

The #8079 re-measurement (details in the issue comment to follow), and the fix for the mechanism it actually found: the public spec wrapper decides **scalar** ordinary-parameter descriptors with the existing typed-abi leaf guards instead of the interpretive `js_param_type_guard`.

## Why

`js_param_type_guard` costs ~450 instructions per call before it checks anything — `Descriptor::parse` plus a 768-byte `GuardState` init — and every unproven direct call through the public wrapper pays it, including a Tier-B guarded clone's own recursion (#8169). At `bfb0707be` that interpretive overhead measured as:

| bench | instructions in the validator |
|---|---|
| tree | 33.8% of the whole program |
| interp | 19.6% |
| tree_wide | 17.0% |
| iso_miss | 16.3% |

(SKIPVAL diagnostic arm: descriptor validation replaced by `true`, routing unchanged, all stdout byte-exact.)

## Soundness

A single-node scalar descriptor's predicate is **bit-for-bit** an existing leaf guard's predicate:

- `OP_NUMBER` = `js_typed_f64_arg_guard` = `is_number || is_int32`
- `OP_INT32` — the validator already delegates to `js_typed_i32_arg_guard`
- `OP_BOOLEAN` = `js_typed_i1_arg_guard` = `TAG_TRUE | TAG_FALSE`
- `OP_STRING` = `js_typed_string_arg_guard` = `is_any_string`

Routing (clone vs generic fallback) is therefore exactly the decision the validator would have made — no annotation lie can reach the clone that could not before. Structural descriptors (unions, objects, arrays, literals) keep the interpretive call and their rodata. `scalar_descriptor_rep` checks the descriptor layout exhaustively, so a future format change falls back to the validator rather than misreading bytes.

## Measured (19-program corpus, instructions retired, best-of-5, byte-exact stdout, peak RSS unchanged on every row)

| bench | base `bfb0707be` | this PR | delta | SKIPVAL ceiling |
|---|---:|---:|---:|---:|
| tree | 15.82B | 10.79B | **−31.8%** | −33.8% |
| tree_wide | 31.59B | 26.55B | **−15.9%** | −17.0% |
| interp | 15.05B | 13.80B | **−8.3%** | −19.6% |
| iso_miss | 17.70B | 16.47B | **−6.9%** | −16.3% |
| all other rows | | | within noise | |

The interp/iso_miss residual is the structural share (`asNum`'s `Value` union, `peek`'s `Parser` object) plus the Tier-B self-recursion wrapper round-trip — that half is #8169, deliberately not touched here.

## Validation

- `cargo test -p perry-codegen --no-fail-fast`: exactly the 9 pre-existing failures **by name** (baseline re-run on pristine `bfb0707be` in the same target dir); the two ordinary-param-guard assertions that tracked the old wrapper shape now assert the stronger one (no interpretive call for scalars, leaf guards present, `$generic` fallback intact).
- New unit test `scalar_descriptor_rep_classifies_exactly_the_leaf_guard_ops` builds descriptors through the real encoder — a layout change goes red instead of silently misclassifying.
- `cargo test -p perry-runtime --lib` 2499/0/4 ignored; `cargo test -p perry --bin perry` 987/0.
- Gates: fmt, file-size, addr-class, local-binding-type, raw-handle (exit 0; the "debt fell by 4" nag is pre-existing `global_this_webassembly.rs` drift, not this PR — left for its owner), shape-descriptor census, class-id, GC knob drift, GC gate wiring, workspace architecture, `gc_root_dominance_check.py --audit-poll-reach`.
- All 19 corpus binaries byte-exact against `expected/` at every arm.

Refs #8079. The wrapper-recursion half is #8169.

## Summary by CodeRabbit

- **Performance**
  - Improved handling of scalar function parameters by using more direct type checks.
  - Reduced unnecessary guard metadata for supported scalar values.

- **Bug Fixes**
  - Preserved existing validation behavior for string, number, integer, and boolean parameters.
  - Maintained fallback handling for complex or unsupported parameter types.

- **Tests**
  - Expanded coverage for scalar parameter validation and regression scenarios.


proggeramlug:
## Mini cross-check (canonical interleaved sweep, run7)

Instrumented instructions retired on the quiet-mini protocol reproduce the PR's local numbers exactly; instructions are the contention-immune metric, and the wall-time column is under a **+4.9% mean ambient slowdown** (another session's `perry compile` batch was running on the mini — run verdict DIRTY by the load rule, zero failures, all 74 cells byte-exact; node/scriptc oracle cells drifted uniformly, so the delta column is perry-side):

| bench | base wall | PR wall | wall Δ | base instr | PR instr | instr Δ | peak RSS |
|---|---:|---:|---:|---:|---:|---:|---:|
| tree | 1.0305 s | 0.5524 s | **−46.4%** | 15.813B | 10.779B | **−31.8%** | unchanged |
| tree_wide | 1.7225 s | 1.3252 s | **−23.1%** | 31.572B | 26.537B | **−15.9%** | unchanged |
| interp | 0.8592 s | 0.7806 s | **−9.1%** | 15.045B | 13.797B | **−8.3%** | unchanged |
| iso_miss | 0.9936 s | 0.9293 s | **−6.5%** | 17.687B | 16.461B | **−6.9%** | unchanged |
| other 15 rows | | | ambient ±drift | | | **±0.4%** | unchanged |

tree's wall gain exceeding its instruction gain (−46% vs −32%) is the 768-byte `GuardState` memset being cache-hostile on top of its instruction count.

Notes:
- 10 of 19 corpus binaries are byte-identical to the base arm (no scalar descriptors) — a built-in no-change control; the 9 that differ are exactly the rows above plus churn/push/fib40-class programs whose deltas are ≤0.4%.
- No unrelated row moved — the "win in an unrelated row is a red flag" check is clean.
- Behavioral oracles: a lying-caller probe (declared `number` fed a string / object / boolean through `any`) and the full `test_gap_specabi_ordinary_param_guards.ts` fixture are byte-exact vs Node 26.5.1.
- Mini artifacts: `~/perry-sweep-0815/` now holds the PR arm in `perry/`, the `bfb0707be` base arm preserved at `perry-bfb07base/`, results in `results-run5/6/7.json` (5+6 = base vs `499e29627`, 7 = PR arm).

Validation is complete from my side: corpus byte-exact at every arm, codegen suite reproduces exactly the 9 pre-existing failures by name, runtime/bin suites green, all script gates pass.

PR #8210, opened 2026-08-16
test(next): arm the routeModule.handle guard and make verifier passes per cold start a knob (default 10) in the release App Route fixture

proggeramlug:
Part of #8040 (DoD audit, 2026-08-16). Test-only, `tests/release/packages/next-app-route/` only.

## (a) The `routeModule.handle` bypass guard, armed and gated

`perry-host.js` is now byte-identical to `tests/fixtures/next-app-route/perry-host.js` (#8161): it wraps `routeModule.handle`, records the request id that passes through it, and logs `<id>: generated handler bypassed routeModule.handle` for any request whose response was produced without entering the generated `AppRouteRouteModule.handle`. That was already run against this fixture on `main` `3c95020f8` (10 cold × 2 verifiers PASS, 0 guard fires in 520 batches — #8040 comment).

The guard's signal lives **only in the host log**: it throws inside a `.then()` after the response is already sent, so `verify.mjs` still exits 0 when it fires (measured in #8161: 22 log hits, exit 0, with the wrapper removed). `fixture.sh` therefore greps every cold-start log for `generated handler bypassed` as a hard failure, in the existing forbidden-diagnostic check (now one `FORBIDDEN_DIAGNOSTICS` regex, with the matching lines echoed before `fail`). Sabotage-checked without a rebuild: a log with a planted `Error: 7: generated handler bypassed routeModule.handle` line matches the regex (would fail), a clean `READY` + `PASS: 21 …` log does not.

## (b) `PERRY_NEXT_ROUTE_VERIFIERS_PER_START`, default 10

The two hard-coded `verify.mjs` runs per cold start (`cold verifier 1` / `warm verifier 2`) become a loop over `PERRY_NEXT_ROUTE_VERIFIERS_PER_START` (positive integer, **default 10**), labelled `cold verifier 1/N` / `warm verifier k/N`. The default is now 10 cold starts × 10 passes = **100 batches**, the same shape as `tests/test_next_app_route_dylib.sh` and #8040's "100-iteration run of the 20-way concurrent request batch" bullet. The `[6/7]`, `[7/7]` and `PASS` lines print the totals.

The forced-evacuation arm (`PERRY_NEXT_ROUTE_FORCED_GC=1`, off by default, #8163) is untouched apart from inheriting the loop; the header comment describes the new default, the guard, and the state below.

## Plainly: with the default 10 this fixture is intermittently RED on today's `main`

That is the knob doing its job. #8163's last comment ("Reproduces under DEFAULT GC") measured 8 of 500 batches losing one response against a warm process with no GC knobs at all — every failure right after a `[gc-copy-minor] ran … in_place=false` line, `TypeError: value is not a function` in the host log, empty body in `verify.mjs`. Two passes per process finished before the first copying minor (~pass 3 in a fresh process), which is exactly why this fixture read green while #8040's 100-iteration bullet was red. `PERRY_NEXT_ROUTE_VERIFIERS_PER_START=2` recovers the previous coverage and passes.

So this is expected to merge **after** #8163's fix (or with the release harness pinning `=2` until then) — owner's call on the order.

## Two regimes, two knobs (coordination with #8215)

This knob and #8215's `PERRY_NEXT_ROUTE_WARM_PASSES` cover different regimes, and neither substitutes for the other:

- **`PERRY_NEXT_ROUTE_VERIFIERS_PER_START` (this PR)** — restart / ABI / parity / bypass-guard coverage across N fresh processes. Each 10-pass process runs ~2–3 copying minors (the first lands around pass 3; measured directly in this PR's validation: every 10-pass cold start logged 3 `[gc-copy-minor] ran` lines under `PERRY_GC_DIAG=1`), so the default is genuinely sensitive to per-collection bugs — a fresh process just lives permanently in the early/small-heap regime.
- **`PERRY_NEXT_ROUTE_WARM_PASSES` (#8215)** — collection *depth* in ONE process, where the heap grows and the minor cadence accelerates (8–9 passes between minors early, tightening to 2–3 by pass ~99). Only a warm process reaches that regime.

#8215 touches the same `[6/7]` region of `fixture.sh`; this PR's diff there is deliberately small, and I will rebase in whichever direction the owner decides. **Whichever of #8210/#8215 lands second must not drop the other's knob** — they are complementary, not alternatives.

## Validation (bench mini, macOS arm64, `perry` from a `perry-dev` build of `main`-equivalent `07c8040bf`, `PERRY_MODULE_JOBS=2`, `PERRY_GC_DIAG=1` on the serving arms)

All three arms ran the full `fixture.sh` end to end (npm ci + `next build` + oracle + provider archives + app dylib + serving) at branch commit `0c65f41b4`; the one later commit is comment-only (verified: zero non-comment diff lines in `fixture.sh`). `PERRY_LLVM_INPROCESS=0` was set for the app compile — see "not this PR's bug" below.

| arm | result | copy-minors / TypeErrors / bypass fires (per failing or first process) |
|---|---|---|
| `PERRY_NEXT_ROUTE_VERIFIERS_PER_START=2` | **PASS** — `PASS next-app-route (20 batches, forced-evacuation arm not run — #8163)`, 10 cold starts x 2, 0 guard fires | 0 minors in a 2-pass process (the pre-knob blind spot, directly reconfirmed) |
| default (10) | **FAIL (expected, #8163)** — `normal cold start 0: warm verifier 3/10 failed`, `verify.mjs` `Unexpected end of JSON input` | 1 `[gc-copy-minor] ran` → 2 `TypeError: value is not a function`, 0 bypass fires |
| default (10), repeat | **FAIL (expected, #8163)** — same batch (cold start 0, verifier 3/10), same signature | 1 minor → 2 `TypeError`s, 0 bypass fires |

Verifier pass 3 is exactly where the first copying minor of a fresh process lands, so the failing batch matches #8163's cadence data precisely. The `=2` arm passing plus the default arm failing **on the first collection** is the knob's whole point: the old hard-coded 2 passes ended before the collector ever moved an object.

Guard sabotage check (no rebuild): the fixture's `FORBIDDEN_DIAGNOSTICS` regex was run against a log with a planted `Error: 7: generated handler bypassed routeModule.handle` line (matches → run fails) and against a clean `READY` + `PASS: 21 …` log (no match). The armed-host + this grep combination also has live evidence from the sibling gate: 0 guard fires across all serving runs.

Mini log paths: `~/perry-bench.noindex/tmp-8040-knob/ext-v2.log`, `ext-v10.log`, `ext-v10-b.log`, per-cold-start host logs under `~/perry-bench.noindex/tmp-8040-knob/logs-ext-v10*/perry-normal-0.log`, fixture build logs in `~/perry-bench.noindex/build-knob/`.

### Re-validated after rebasing onto `main` with #8211 (forced arm now ON by default)

#8211 flipped `PERRY_NEXT_ROUTE_FORCED_GC` to default 1, so the forced arms now run *through* this PR's new verifier loop. Rebased onto `c926cf142`, rebuilt `perry` from the rebased tree, and re-ran at `PERRY_NEXT_ROUTE_VERIFIERS_PER_START=2`:

```
[6/7] 10 cold processes (alternating normal / FORCED-evacuation), 2 21-request verifier runs each (20 batches)
next-app-route-forced-1: evacuation live — 440 copying minor(s), 197840 objects copied
next-app-route-forced-3: evacuation live — 440 copying minor(s), 198319 objects copied
next-app-route-forced-5: evacuation live — 440 copying minor(s), 197006 objects copied
next-app-route-forced-7: evacuation live — 440 copying minor(s), 197883 objects copied
next-app-route-forced-9: evacuation live — 440 copying minor(s), 197877 objects copied
[7/7] production AppRouteRouteModule.handle parity complete: 20 verifier batches over 10 cold starts, 0 bypass-guard fires
PASS next-app-route (20 batches, with forced-evacuation arm)
```

**exit 0.** All five forced cold starts passed `scripts/gc_evacuation_liveness_assert.py` through the new loop — the subject was live (440 copying minors, ~198k objects copied each), not merely "nothing threw". Log: `~/perry-bench.noindex/tmp-8040-knob/rebased-v2.log`, per-cold-start logs in `logs-rebased-v2/`. Rebase conflict resolution kept #8211's post-fix wording and this PR's `$VERIFIERS_PER_START` / `$TOTAL_BATCHES` parameterization; the header now describes #8163 as reopened on the default-GC residual rather than fixed.

The default-10 arms above were measured *before* #8211 landed. #8211 fixed the two forced-arm holders; the default-GC residual it left is now named and has a fix in flight (**#8220**, `js_headers_for_each`'s hoisted closure pointer in `crates/perry-stdlib/src/fetch/headers.rs:460` plus eight sibling sites — 8 from-space faults → 0, 3 failures/300 warm passes → 0). So the expected merge order is: #8220 first, then this PR's default-10 goes green. I did not re-measure the default-10 arms on top of #8220.

## Not this PR's bug: `fixture.sh`'s default in-process LLVM backend fails 5 modules on current `main`

On this box, the untouched fixture's app compile (in-process LLVM 22 backend, the default when `PERRY_LLVM_INPROCESS` is unset) fails deterministically on current `main`-equivalent (`07c8040bf`): `5 module(s) failed to compile — REFUSING TO LINK` — `.next/server/chunks/2.js`, `chunks/430.js`, `next/dist/compiled/jsonwebtoken/index.js`, `app-page.runtime.prod.js`, `app-route.runtime.prod.js` (the app's 5 biggest modules), with **no per-module error text at all** (the summary says to search for `Error compiling module`; no such line is emitted). The same app compiles clean with `PERRY_LLVM_INPROCESS=0` (the dylib gate's configuration), and the untouched fixture compiled clean in-process on `3c95020f8` on this same box earlier today (#8163's validation runs) — so the window is `3c95020f8..07c8040bf` (candidates touching codegen: #8201, #8204, #8206). Repro + log: `~/perry-bench.noindex/tmp-8040-knob/probe-inprocess.log`. Filed separately; this PR does not work around it beyond validating with the external backend.

## Summary by CodeRabbit

- **Tests**
  - Improved release validation for Next.js route handling by detecting requests that bypass the expected processing path.
  - Added configurable verification passes for cold and warm process testing, with clearer progress and failure reporting.
  - Preserved forced garbage-collection and liveness checks while expanding repeated verification coverage.
  - Documented available verification settings and the default test behavior.

### DIFF
--- a/changelog.d/8210-release-next-fixture-guard-knob.md
+++ b/changelog.d/8210-release-next-fixture-guard-knob.md
@@ -8,7 +8,9 @@
 - The two hard-coded verifier passes per cold start became
   `PERRY_NEXT_ROUTE_VERIFIERS_PER_START` (default 10), so the default run is
   10 cold starts × 10 passes = 100 batches, matching
-  `tests/test_next_app_route_dylib.sh` and #8040's 100-iteration bullet. With
-  the default this fixture is red on today's `main` because of #8163 (~2% of
-  warm batches lose one response after a default-mode copying minor); `=2`
+  `tests/test_next_app_route_dylib.sh` and #8040's 100-iteration bullet. Each
+  10-pass process runs ~2-3 copying minors, so the default is sensitive to
+  per-collection bugs and intermittently red on today's `main` (#8163: ~2% of
+  post-minor batches lose one response); collection depth in one warm process
+  is `PERRY_NEXT_ROUTE_WARM_PASSES` (#8215), a complementary knob. `=2`
   restores the previous coverage. The forced-evacuation arm is unchanged.
--- a/changelog.d/8210-release-next-fixture-guard-knob.md
+++ b/changelog.d/8210-release-next-fixture-guard-knob.md
@@ -1,0 +1,14 @@
+### Testing
+
+- `tests/release/packages/next-app-route/fixture.sh` now runs the armed
+  `routeModule.handle` bypass guard (its `perry-host.js` is byte-identical to
+  `tests/fixtures/next-app-route/perry-host.js`) and greps every cold-start log
+  for `generated handler bypassed` as a hard failure — the guard's only signal is
+  the host log; `verify.mjs` exits 0 when it fires (#8161).
+- The two hard-coded verifier passes per cold start became
+  `PERRY_NEXT_ROUTE_VERIFIERS_PER_START` (default 10), so the default run is
+  10 cold starts × 10 passes = 100 batches, matching
+  `tests/test_next_app_route_dylib.sh` and #8040's 100-iteration bullet. With
+  the default this fixture is red on today's `main` because of #8163 (~2% of
+  warm batches lose one response after a default-mode copying minor); `=2`
+  restores the previous coverage. The forced-evacuation arm is unchanged.
--- a/tests/release/packages/next-app-route/fixture.sh
+++ b/tests/release/packages/next-app-route/fixture.sh
@@ -203,8 +203,10 @@ run_cold_start() {
   local verifier label
   for verifier in $(seq 1 "$VERIFIERS_PER_START"); do
     if (( verifier == 1 )); then label="cold"; else label="warm"; fi
-    BASE_URL="http://127.0.0.1:$port" node verify.mjs >>"$log" 2>&1 \
-      || fail "$mode cold start $index: $label verifier $verifier/$VERIFIERS_PER_START failed"
+    BASE_URL="http://127.0.0.1:$port" node verify.mjs >>"$log" 2>&1 || {
+      tail -40 "$log" | sed 's/^/    /'
+      fail "$mode cold start $index: $label verifier $verifier/$VERIFIERS_PER_START failed"
+    }
   done
   if [[ "$mode" == "forced" ]]; then
     python3 "$REPO_ROOT/scripts/gc_evacuation_liveness_assert.py" \
--- a/tests/release/packages/next-app-route/fixture.sh
+++ b/tests/release/packages/next-app-route/fixture.sh
@@ -15,19 +15,21 @@
 # throws inside a `.then()` after the response is already sent, so
 # `verify.mjs` still exits 0 when it fires (#8161).
 #
-# Known state (#8163): with the default 10 verifier passes per process this
-# fixture is intermittently RED on today's `main` — a default-mode copying
-# minor occasionally strands a stale closure, and ~2% of the batches that
-# follow one lose a response (`TypeError: value is not a function` in the host
-# log right after a `[gc-copy-minor] ran` line, then an empty body in
-# `verify.mjs`). Two passes per process finished before the first copying
-# minor (~pass 3), which is how the fixture read green while #8040's
-# 100-iteration bullet was red. Set `PERRY_NEXT_ROUTE_VERIFIERS_PER_START=2`
-# to recover the old coverage.
+# Known state (#8163, reopened): #8211 rooted the two holders behind the
+# forced-evacuation arm, but a default-GC residual remains — a default-mode
+# copying minor occasionally strands a stale closure, and ~1-2% of the batches
+# that follow one lose a response (`TypeError: value is not a function` in the
+# host log right after a `[gc-copy-minor] ran` line, then an empty body in
+# `verify.mjs`). With the default 10 verifier passes per process this fixture
+# is therefore intermittently RED until that residual lands. Two passes per
+# process finished before the first copying minor (~pass 3), which is how the
+# fixture read green while #8040's 100-iteration bullet was red; set
+# `PERRY_NEXT_ROUTE_VERIFIERS_PER_START=2` to recover exactly the old
+# coverage.
 #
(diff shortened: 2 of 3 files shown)


proggeramlug:
---

**State at ready-for-review (2026-08-16)**

- The default `PERRY_NEXT_ROUTE_VERIFIERS_PER_START=10` arm goes red on today's `main` because of **#8163's default-GC residual** — fixed in flight by **#8220**. That is the knob working as designed, not a defect in this PR: measured twice, identically, at cold start 0 / warm verifier 3 of 10, with 1 copying minor and 0 bypass-guard fires. `=2` passes (20 batches, 0 guard fires), which is exactly why the old hard-coded 2 could never see this class.
- **Merge order**: #8215 (`PERRY_NEXT_ROUTE_WARM_PASSES`) edits the same `[6/7]` region of `fixture.sh`. Whichever lands second must rebase and **must not drop the other knob** — they measure different regimes: this knob buys restart/ABI/parity/bypass-guard coverage across N fresh processes (~2 collections each, permanently in the small-heap regime), #8215's buys collection depth in one warm process where the cadence accelerates. Neither substitutes for the other.
- **#8228** blocks a full HEAD validation of the in-process path until fixed.

Issue #8174, opened 2026-08-15
gc: rewrite_raw_addr follows a forwarding pointer out of an address that is not a live object start


proggeramlug:
## Summary

`CopyingNurseryCollector::rewrite_raw_addr` will follow a "forwarding pointer" out of an address that is **not a live object start**, and will accept a NaN-boxed word as the forwarding target. Every *rekeyed* address-keyed metadata table shares this exposure.

Found while root-causing #8040's forced-GC failure (fixed in #8168), but it is a distinct defect: #8168 removes the *dead key* that triggered it here; this issue is that a dead or bogus address is followed at all.

## Mechanism (instrumented, from the #8168 investigation)

1. A side table holds a raw heap address whose object has died.
2. The arena recycles that address for an unrelated allocation.
3. The recycled bytes are read as a `GcHeader`. Its `gc_flags` byte happens to carry `GC_FLAG_FORWARDED` — observed `0x86`, `obj_type=104`, `in_census=false`.
4. `rewrite_raw_addr` accepts it, because its gate is `classify_heap_space(addr) != Unknown`, which is true for **any** arena address, live or dead.
5. It reads the payload word as a forwarding pointer. In the observed case that word was a **NaN-boxed value** (`0x7FFF…`); the next hop classified `Unknown` and the walk stopped (`stop_reason=2, hops=1`).
6. `visit_metadata_nanbox_key` then masked it to 48 bits, yielding a **genuine, live survivor object that was itself being evacuated** — an object entirely unrelated to the original key.

In #8040 the object behind the key was `GC_TYPE_STRING` + `GC_FLAG_INTERNED`, a shape a closure-only key can never legitimately name. That is what identified the value as stale rather than mis-rewritten (confirmed under lldb).

## Why it was not fixed in #8168

The obvious tightening — gating on `self.ptrs.classify()` instead of `classify_heap_space` — is precisely what `crates/perry-runtime/src/gc/copying.rs:292-317` documents as having caused the *opposite* bug for `shapes.entries`. Mid-cycle there is no census to consult, so "is this address a live object start?" has no cheap authoritative answer at that point.

#8168 addressed the root cause for its own table (a dead key should never have survived to be followed) rather than widening the fix under time pressure. That is the right call for that PR, but it leaves the general exposure in place.

## Why it matters

[...]

PR #238, opened 2026-04-28
fix(fetch): #234 real Blob with arrayBuffer/text/bytes/slice instance methods

TheHypnoo:
## Summary

Closes #234.

Pre-fix `await response.blob()` returned a metadata-only stub `{size, type}` and silently dropped `resp.body`. Calls to `blob.arrayBuffer()` / `.text()` / `.bytes()` / `.slice()` got nothing because no codegen dispatch arms existed.

## Approach

Three-part fix mirroring the precedent from #232 (real `BufferHeader` for `response.arrayBuffer`) and #227 (registered-buffer `is_registered_buffer` plumbing):

### 1. Runtime — `crates/perry-stdlib/src/fetch.rs`

- New `BlobData { body, content_type }` + `BLOB_REGISTRY: Mutex<HashMap<usize, BlobData>>` + `NEXT_BLOB_ID` + `alloc_blob` helper, mirroring the `HEADERS_REGISTRY` / `alloc_headers` pattern in the same file.
- `js_response_blob` rewritten: clones body bytes + content-type into a fresh `BlobData`, registers it, resolves the promise with the numeric handle as `f64`. Matches the Response-handle ABI documented at `lower_call.rs:2788`.
- 6 new FFIs:
  - `js_blob_size(handle) -> f64`
  - `js_blob_type(handle) -> *mut StringHeader` (codegen NaN-boxes with `STRING_TAG`)
  - `js_blob_array_buffer(handle) -> *mut Promise` — allocates a real `BufferHeader` via `perry_runtime::buffer::buffer_alloc(len)`, memcpys body, NaN-boxes as `POINTER_TAG`. Same shape as `js_response_array_buffer` from #232 so `new Uint8Array(buf)` and `Buffer.from(buf)` see real bytes via the `is_registered_buffer` path from #227.
  - `js_blob_bytes(handle) -> *mut Promise` — alias to `js_blob_array_buffer` (the `BufferHeader` is already byte-array-shaped).
  - `js_blob_text(handle) -> *mut Promise` — UTF-8 lossy decode to `StringHeader`, matches WHATWG Blob.text() replacement-character semantics.
  - `js_blob_slice(handle, start, end, type_ptr) -> f64` — canonical `f64::NAN` sentinel for missing numeric args, null `type_ptr` to inherit the original content-type. Negative indices count from the end and clamp to `[0, len]` per WHATWG Blob spec; allocates a fresh blob with sliced bytes.

[...]

Issue #218, opened 2026-04-27
GTK4: SplitView FFIs (`splitview_create`, `splitview_add_child`) not wired — co-blocks docs/examples/ui/layout/snippets.ts on Linux CI


proggeramlug:
## Context

`docs/examples/ui/layout/snippets.ts` imports `SplitView` and `splitViewAddChild` from `perry/ui` to demonstrate the split-pane layout pattern. The example compiles + links cleanly on macOS (`NSSplitView`) but fails on the GTK4-on-Ubuntu doc-tests CI runner with:

```
/usr/bin/ld: perry_llvm_*.ll:(.text+...): undefined reference to `perry_ui_splitview_create'
/usr/bin/ld: perry_llvm_*.ll:(.text+...): undefined reference to `perry_ui_splitview_add_child'
```

Co-blocker (alongside #217's overlay/reorder FFIs) for the gtk4 doc-tests gate on origin/main.

## What's missing

| FFI | macOS | GTK4 mapping |
|---|---|---|
| `perry_ui_splitview_create` | `NSSplitView` widget | `gtk::Paned::new(Orientation::Horizontal)` |
| `perry_ui_splitview_add_child` | `addArrangedSubview:` | `Paned::set_start_child(&child)` for first call, `set_end_child(&child)` for second |

## Implementation notes

- `gtk::Paned` is GTK4's split-pane primitive. It supports exactly **two** children (`start_child` + `end_child`), unlike `NSSplitView` which can hold N. Most uses of `SplitView` in TS pass exactly 2 children, but the API allows more.
- Decision needed: cap GTK4's `splitview_add_child` at 2 (third+ call → no-op or error), or nest `Paned`s recursively to support N children. The simplest+correct first cut is the 2-child cap, with a runtime warn if a third child arrives.
- Default orientation: macOS `NSSplitView` defaults to vertical-divider (horizontal arrangement). Match that with `Orientation::Horizontal` on `Paned`.
- The handle/separator is automatic on `Paned`; no need to wire a separate `setOrientation` API for the first cut unless it already exists on macOS.

## Acceptance

- [ ] Both FFIs exported as `#[no_mangle] pub extern "C" fn` in `crates/perry-ui-gtk4/src/lib.rs`.
- [ ] If `splitview` rows exist in `crates/perry-ui/src/styling_matrix.rs`, flip the GTK4 cells from `Missing` to `Wired`. Verify `./target/release/styling-matrix --check` is clean.
[...]

PR #8211, opened 2026-08-16
fix(gc): root the two holders outside the GC heap behind the forced-evacuation App Route arm (#8163)

proggeramlug:
Refs #8163 — **does not close it.** This fixes the forced-evacuation arm the issue is titled for, and a parallel session's independent 500-batch measurement shows at least one *more* unregistered holder remains on the DEFAULT-GC path with the identical signature. See "Residual" below.

## What was wrong

The production Next 16.3.0 App Route fixture's forced-evacuation arm failed with `TypeError: value is not a function` around copying minor #238. The issue's elimination trail was right: the stale closure was held by something **outside the GC heap and unregistered with any root scanner**. There were two of them, one behind the other, and neither is visible to any existing instrument — `PERRY_GC_VERIFY_EVACUATION` (no scanner to verify), `PERRY_GC_PROTECT_FROMSPACE_HOLDERS` (not in the heap), or `scripts/gc_runtime_root_holders.py` (see below).

Both were **measured, not inferred**: the host ran under `PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800` and each suspect was instrumented until it printed the exact address the fault reporter named.

### 1. `HEADERS_METHOD_VALUE_CACHE` (perry-stdlib `fetch`)

`headers.get` / `.entries` / … are bound-method closures cached per `(handle, method)` in a `lazy_static!` `Mutex<HashMap<_, u64>>`. The store site called `js_write_barrier_root_nanbox`, which is the incremental-marking **shade**, not a root registration — nothing marked or rewrote the cache. The route reads `(await headers()).get("x-request-id")` twice per request with awaits between; the closure allocated by the first read (cache *miss*, `id=262149 name=get bits=0x7ffd044c5059f980`) was retired by a later minor, and the second read (cache *hit*, same bits) handed it to `typeof` inside Next's `ReflectAdapter.get` — fault at `0x44c5059f978`, the header of that 48-byte closure.

`FORM_DATA_METHOD_VALUE_CACHE` and `RequestRecord::signal` (the `AbortSignal` behind `request.signal`) have the same shape. New `fetch/gc.rs` registers one scanner through the C ABI (`perry_ffi_gc_register_mutable_root_scanner_named`, like `streams/gc.rs`, so a trimmed stdlib provider installs it in the runtime image) that marks and rewrites all three. `js_request_new` now builds its `RequestRecord` — including the default-`AbortSignal` allocation — *before* taking the registry lock, because the scanner takes that lock during a collection on the same thread.

### 2. `ServerResponse.once_listeners` (perry-ext-http)

`res.once(event, cb)` stores into a **second** table that `take_event_listeners` merges into every emit; `scan_http_server_roots` visited only `listeners`. Next's `pipeToNodeResponse` registers `res.once('close', …)`. Instrumented: the `close` snapshot taken at the top of `res.end()` already contained `0x3f4dc628190` (retired at minor #211) — stale *at snapshot time*, i.e. the table itself was never rewritten. The scanner now visits it.

While there, the whole `res.end()` tail (`js_node_http_res_end`, `_full`, `_with_cb`, `standalone_end`) took listener/callback snapshots **out** of the handle and then ran JS (write callbacks, `'finish'`, the end callback) before using them. `emit_no_arg_to_listeners` roots what it is *handed* (#8082), so a stale snapshot stayed stale. `EndTail` (new `server/response_end.rs` — split out because `response.rs` sat at the 2,000-line cap at 1993 lines) parks every snapshot in the transient-root stack before the first JS call and re-reads per use; the two Node orderings (buffered: finish → end cb → close; standalone: end cb → finish → close) are preserved. `js_node_http_im_resume`'s `'end'` snapshot (crosses the `'data'` emits) and `js_node_https_server_close`'s callback (the https twin of the #8082 http fix) get the same treatment.

## Registering the scanner made a dozen latent sites into deadlocks

`std::sync::Mutex` is not reentrant, and the scanner takes `REQUEST_REGISTRY` during a collection **on the mutator thread** — so once this scanner exists, any site holding that guard across a GC allocation self-deadlocks the first time the allocation collects. `dispatch_request_property`'s twelve string arms, `js_request_get_url` / `_method` / `_body`, `js_request_input_to_url` and `request_string_field` all did exactly that; they now snapshot the field bytes under the guard and allocate after dropping it.

`js_request_clone` was worse: it **threw** while holding the guard. The exception transport is written for `panic=abort` and steps through the frame without running `Drop`, so the mutex is not poisoned — it stays *locked for the life of the process*, and the scanner's `if let Ok(...)` would then block rather than skip. It now decides "unusable" under the guard and throws outside it.

Two tests hold this, because the failure mode is a **hang** and a hanging test is worse than a failing one:

* `request_reads_release_the_registry_guard` — `try_lock`s after every reader and after all fifteen dispatcher props; catches a leaked guard (the clone case) deterministically.
* `no_allocation_is_taken_off_a_live_registry_borrow` — source-scans for `js_string_from_bytes(req.…)`, the syntactic signature of allocating out of a borrow only the guard keeps alive, and asserts against a planted sample that the scan can still match the shape it forbids.

The `fetch/gc.rs` module doc now states the contract and names every hoisted site. Reported in review by a parallel session working adjacent GC issues; verified here against the code before acting. (`perry-ext-fetch` has its own `REQUEST_HANDLES` registry which this scanner does not touch, so its twins are out of scope for this PR.)

## Why the audit missed it — four blind spots in `scripts/gc_runtime_root_holders.py`, all fixed

| blind spot | effect |
|---|---|
| `DECL` did not match `lazy_static!`'s `static ref` | 93 tables across runtime+stdlib were outside the census, `HEADERS_METHOD_VALUE_CACHE` among them |
| `strip_comments` blanks string literals, so `extern "C" fn` reached `FN_DEF` as `extern "" fn` | no C-ABI function had a body in the walk — including the FFI scanner trampolines (`scan_stream_roots_ffi`, `scan_fetch_roots_ffi`) |
| a body-less `fn f(...);` in an `extern "C" {}` block started a brace count | swallowed the functions after it, hiding the same trampolines a second way |
| the registration regex stopped at the first `(` of `SOURCE.as_ptr()` | every C-ABI registration lost its scanner name |

Census: **78 → 129** declarations, **53 → 74** reached by a registered scanner — i.e. the gate was green because the narrow regex hid the candidates, not because they were classified. The 31 newly visible uncovered holders each carry a written verdict in `gc_runtime_root_holders.json` (counters, handle-id maps, Rust-owned registries, or `covered_elsewhere` naming the scanner/rekey hook — e.g. `REQUEST_REGISTRY` → `fetch::gc`, `READABLE_STREAMS`/`TRANSFORM_BACKPRESSURED_JOBS` → `streams::gc`, `DIAG_TRACES` → `scan_node_submodule_singleton_roots_mut`, `THREAD_GLOBAL_THIS` → `js_gc_register_global_root`); one stale entry (`EXT_BLOCKING_TASKS_INFLIGHT`, now reached) is deleted. `--self-test` passes.

## Validation

Same app dylib, same seed (8036), providers built from the same tree, A/B by swapping only the provider images:

| arm | plain forced run | + `PROTECT_FROMSPACE` depth 800 |
|---|---|---|
| main | **59** `TypeError`s, 0 `PASS`, 455 copying minors | fault at minor #218 (`obj_type=4 size=48`) |
| fix | **0** errors, **2× `PASS: 21`**, 446 copying minors, 0 verify panics | clean; also clean on seeds 8174, 1, 8040, 4242 |

Full `tests/release/packages/next-app-route/fixture.sh` with the forced arm ON (now the default) and a fresh release compiler: see the run summary in the last comment. `PERRY_NEXT_ROUTE_FORCED_GC` flips from opt-in to on-by-default; `=0` still runs the normal arm only.

Tests: `fetch::tests::fetch_root_scanner_emits_method_caches_and_request_signal` and `…rewrites_relocated_slots_in_place` (the cache HIT path must return the *rewritten* value); the ext-http `root scanner rewrites` test gains `once_listeners` and `pending_write_callbacks`. `cargo test --release -p perry-stdlib --lib fetch::` and `-p perry-ext-http --lib` green; `check_file_size.sh`, `cargo fmt --check`, `gc_runtime_root_holders.py` (+ `--self-test`) clean.

## Residual — #8163 stays open

Independently measured on the quiet bench mini (same host, same hour, 500 warm batches per arm, no GC knobs), by the session that owns the #8040 batch instrument — numbers theirs, [full comment](https://github.com/PerryTS/perry/issues/8163#issuecomment-5307890158):

| arm | default-GC warm batches | forced+seeded |
|---|---|---|
| main `3c95020f8` | 4 failures / 100 | dies at copying minor ~#238 |
| this branch | 5 failures / **500** | survives to copying minor **~#2696**, 3712 minors, Σcopied ≈1.04M, 0 verify panics |

So the forced arm is genuinely fixed (~10× longer survival, and the shipped fixture passes 10/10 with liveness asserted), and the default-path failure rate drops — but it is **not eliminated**: 5 batches still fail with an empty body plus `TypeError: value is not a function` landing immediately after a default-mode `[gc-copy-minor] ran … in_place=false`. Their own caveat is worth keeping: against yesterday's 8/500 main figure the improvement overlaps Poisson noise, so treat the rate reduction as suggestive and the residual as certain.

**The shipped fixture is structurally blind to this** — it runs 2 verifier passes per process, and the residual needs ~100 passes against one warm process to surface. That instrument (`~/mini-batch-8163.sh`) lives with them, and by agreement they are taking the residual hunt, scoped out of every file this PR touches. This PR therefore uses `Refs`, not `Closes`.

## Not fixed here (follow-ups)

* perry-ext-net's `once_flags` keys a `HashSet<i64>` by closure **address** to decide which listener to drop after a `once` fires — a rekeyed table of the #8174 family: a closure the collector moves fires again, and a recycled address drops the wrong listener. Semantic, not memory-unsafe.
* Handle-struct fields in the ext crates (`once_listeners` was one) are still outside any census; `gc_runtime_root_holders.py` audits `static` declarations only.

https://claude.ai/code/session_01YAif84burv8q6QngSN6wU8

### DIFF
--- a/changelog.d/8211-gc-8163-forced-evac-holders.md
+++ b/changelog.d/8211-gc-8163-forced-evac-holders.md
@@ -1,0 +1,63 @@
+### Fixed — forced-evacuation App Route arm: two holders outside the GC heap (#8163)
+
+The production Next 16.3.0 App Route fixture's forced-evacuation arm died with
+`TypeError: value is not a function` around copying minor #238. Both holders were
+Rust tables outside the GC heap that no root scanner marked or rewrote — the class
+`PERRY_GC_VERIFY_EVACUATION` cannot see (no scanner to verify) and
+`PERRY_GC_PROTECT_FROMSPACE_HOLDERS` cannot see (not in the heap). Found by running
+the host under `PERRY_GC_PROTECT_FROMSPACE=1 …_DEPTH=800` and instrumenting the two
+suspects until each printed the exact address the fault reporter named.
+
+* **`HEADERS_METHOD_VALUE_CACHE`** (perry-stdlib `fetch`) — the bound-method closure
+  behind `headers.get` / `.entries` / …, cached per `(handle, method)` in a
+  `lazy_static!` `HashMap<_, u64>`. Its only "rooting" was
+  `js_write_barrier_root_nanbox`, which is the incremental-marking shade, not a root.
+  Next's route reads `(await headers()).get(...)` twice per request with awaits
+  between; the closure allocated by the first read was retired by a later minor and
+  the second read handed the pre-move address to `typeof` inside Next's
+  `ReflectAdapter.get`. `FORM_DATA_METHOD_VALUE_CACHE` and `RequestRecord::signal`
+  (the `AbortSignal` behind `request.signal`) had the same shape. A new
+  `fetch/gc.rs` registers a scanner through the C ABI (like `streams/gc.rs`, so a
+  trimmed stdlib provider installs it in the runtime image) that marks and rewrites
+  all three; `js_request_new` now builds its record before taking the registry lock,
+  since the default-signal allocation could otherwise collect under the guard the
+  scanner needs.
+* **`ServerResponse.once_listeners`** (perry-ext-http) — `res.once(event, cb)` stores
+  into a second table that `take_event_listeners` merges into every emit, and
+  `scan_http_server_roots` visited only `listeners`. Next's `pipeToNodeResponse`
+  registers `res.once('close', …)`; the closure was retired at #211 and reached
+  `emit_no_arg_to_listeners` from the `res.end()` tail as a from-space address. The
+  scanner now visits it. While there, the `res.end()` tail (`js_node_http_res_end`,
+  `_full`, `_with_cb`, `standalone_end`) took listener/callback snapshots OUT of the
+  handle and then ran JS before using them — rooting inside
+  `emit_no_arg_to_listeners` roots what it is *handed*, so a stale snapshot stayed
+  stale. `EndTail` (new `server/response_end.rs`, split out because `response.rs`
+  sat at the 2,000-line cap) parks every snapshot in the transient-root stack before
+  the first JS call. `js_node_http_im_resume`'s `'end'` snapshot and
+  `js_node_https_server_close`'s callback get the same treatment.
+
+**Why the audit missed it.** `scripts/gc_runtime_root_holders.py` could not see any
+of this: its `DECL` regex did not match `lazy_static!`'s `static ref`, so 93 tables
+across runtime+stdlib were outside the census; `strip_comments` blanks string
+literals so `extern "C" fn` reached `FN_DEF` as `extern "" fn` and no C-ABI function
+had a body in the walk (the FFI scanner trampolines included); a body-less
+`fn f(...);` inside an `extern "C" {}` block started a brace count that swallowed
+the following functions; and the registration regex stopped at the first `(` of
+`SOURCE.as_ptr()`, dropping the scanner name from every C-ABI registration. All four
+are fixed. The census grows from 78 to 129 declarations (74 reached by a scanner,
+was 53); the 31 newly visible uncovered holders each carry a written verdict in
+`gc_runtime_root_holders.json`, and one stale entry is deleted.
+
+**Validation.** Same app dylib, same seed (8036), providers from the same tree:
+main → 59 `TypeError`s / 0 `PASS`, from-space fault at minor #218; fixed → 2×
+`PASS: 21`, 446 copying minors, 0 errors, and clean under
+`PERRY_GC_PROTECT_FROMSPACE=1 …_DEPTH=800` on seeds 8036, 8174, 1, 8040, 4242. The
+fixture's forced arm (`PERRY_NEXT_ROUTE_FORCED_GC`) is back ON by default. Unit
+tests: the fetch scanner emits and rewrites all three slots; the ext-http rewrite
+test gains `once_listeners` and `pending_write_callbacks`.
+
+Follow-up (not fixed here): perry-ext-net's `once_flags` keys `HashSet<i64>` by
+closure ADDRESS to decide which listener to drop after a `once` fires — a rekeyed
+table of the #8174 family (a moved closure fires twice, a recycled address drops the
+wrong listener); and the ext crates' handle-struct fields are still outside any
+census (`once_listeners` was one).
--- a/changelog.d/8211-gc-8163-forced-evac-holders.md
+++ b/changelog.d/8211-gc-8163-forced-evac-holders.md
@@ -36,6 +36,24 @@ suspects until each printed the exact address the fault reporter named.
   the first JS call. `js_node_http_im_resume`'s `'end'` snapshot and
   `js_node_https_server_close`'s callback get the same treatment.
 
+**Registering a scanner over a table makes every "allocate under its guard" site a
+deadlock, so those were hoisted in the same change.** `std::sync::Mutex` is not
+reentrant: the scanner takes `REQUEST_REGISTRY` during a collection on the mutator
+thread, so any site holding that guard across a GC allocation self-deadlocks the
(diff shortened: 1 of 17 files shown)


proggeramlug:
## Full fixture run — PASS with the forced-evacuation arm ON

`tests/release/packages/next-app-route/fixture.sh`, 10 cold processes (alternating normal / FORCED-evacuation), two 21-request verifier passes each, against a release compiler built from this branch:

```
  commit=232f5c6e3 next=16.3.0 mode=dylib providers=2308e5fe8d3521da…
  [6/7] 10 cold processes (alternating normal / FORCED-evacuation), two 21-request verifier runs each
next-app-route-forced-1: evacuation live — 447 copying minor(s), 197338 objects copied
next-app-route-forced-3: evacuation live — 439 copying minor(s), 197111 objects copied
next-app-route-forced-5: evacuation live — 443 copying minor(s), 196794 objects copied
next-app-route-forced-7: evacuation live — 443 copying minor(s), 196792 objects copied
next-app-route-forced-9: evacuation live — 443 copying minor(s), 197115 objects copied
  [7/7] production AppRouteRouteModule.handle parity complete
PASS next-app-route (with forced-evacuation arm)
```

All ten cold starts: `PASS: 21` twice, **0** `TypeError: value is not a function`, **0** `gc evacuation verification failed`. `scripts/gc_evacuation_liveness_assert.py` passes on all five forced arms, so the subject was live rather than "nothing threw".

**Artifact provenance was verified rather than assumed** (a second fixture run was in flight in the same build dir, so this mattered): the provider images are 16:50:07, linked from archives built at 16:46:29 — i.e. *after* the guard-hoist commit — and neither they nor `next-app.dylib` (17:07:57) were rewritten during the cold-start window 17:08:11–17:09:50, during which no other server process existed.

Also re-checked after the hoists, same seed 8036 on the standalone host: plain `2× PASS: 21` / 0 errors / 446 copying minors, and clean again under `PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800` (438 retired sets, no fault).

proggeramlug:
### Second full fixture run — PASS, run alone

A repeat of the run above, same commit `232f5c6e3` and same provider hash `2308e5fe…`, this one with no other fixture process in flight (its cold-start loop ran 17:19:22–17:20:45; the previous run exited at 17:09:50):

```
next-app-route-forced-1: evacuation live — 443 copying minor(s), 197138 objects copied
next-app-route-forced-3: evacuation live — 443 copying minor(s), 197402 objects copied
next-app-route-forced-5: evacuation live — 443 copying minor(s), 197846 objects copied
next-app-route-forced-7: evacuation live — 443 copying minor(s), 197115 objects copied
next-app-route-forced-9: evacuation live — 443 copying minor(s), 197708 objects copied
PASS next-app-route (with forced-evacuation arm)
```

All ten cold starts `PASS: 21` twice, 0 `TypeError`, 0 verify panics. Two independent 10-cold-start runs, 20 processes total, 100 forced-evacuation copying-minor-heavy request cycles between them.

PR #8215, opened 2026-08-16
test(next): add the warm-process soak arm #8163's residual needs

### LINKED ISSUE #8163: [Next.js/dylib] Forced-evacuation App Route arm: stale closure from a holder outside the GC heap

proggeramlug:
Refs #8163. Adds the acceptance instrument that issue needs, and shows why the existing one cannot serve.

## Why

`fixture.sh`'s cold-start loop makes **two verifier passes per process**. Two facts about that, both measured here:

1. **In NORMAL mode, two passes run ZERO copying minors.** The normal cold-start arm therefore cannot exercise a moving-GC holder *at all* — only the forced arm moves anything. That is a stronger statement than "insensitive": there is nothing for a rooting bug to survive.
2. **#8163's residual costs ~1 broken request per 100 verifier passes** in a warm process, so even a hundred passes is not a verdict: at p = 0.01 a clean 100-pass run happens **~37%** of the time on a known-broken build — and two of five measured 100-pass runs *did* come back clean while the bug was present.

So "the fixture is green" has never been evidence about this class, and closing #8163 on it would be the failure mode CLAUDE.md names — a gate that cannot fail.

## What

`run_warm_soak`: one warm process, N verifier passes, default GC (forcing evacuation would measure the arm that is already fixed).

* **OFF by default** (`PERRY_NEXT_ROUTE_WARM_PASSES=0`) because a meaningful N is slow. This is the close-condition instrument for #8163, not a per-run check.
* **Prints the confidence its N actually buys** instead of letting any green run imply elimination — rule of three, ~3/p passes for 95%:

  | N | verdict printed |
  |---|---|
  | 10 | 10% — NOT sufficient, a broken build passes this often |
  | 100 | 63% — NOT sufficient |
  | **300** | **95% — sufficient to claim elimination** |
  | 460 | 99% — sufficient |

* **Asserts its subject was live** (`copying minors > 0`), the same rule the forced arm's evacuation-liveness assert applies.
* **Reports an observed failure ahead of the liveness complaint.** I had these the other way round at first, and a real broken request got masked behind "exercised nothing" — caught by running the arm, not by reading it.

## Validation — every path exercised against real provider images

| run | result |
|---|---|
| `N=0` | prints guidance, returns 0 |
| pre-fix providers, `N=100` | **caught it** — pass 19, 1 host TypeError, 21 copying minors |
| merged-fix providers, `N=100` | **caught the residual** — pass 3, 22 minors |
| merged-fix providers, `N=10` | **caught the residual** — pass 6, 2 minors |
| `N=2`, clean | 0 failures, **0 copying minors** → liveness assert fires (this is finding #1 above) |
| liveness assert, sabotaged grep | fires with "exercised nothing" |
| confidence print | verified at N = 10/100/300/460 |

The middle two rows are an independent local reproduction of the residual that @proggeramlug measured on the bench mini — same signature, different host, on the merged fix.

## Note for whoever closes #8163

The pass numbers that fail (3, 6, 12, 19 across runs) track collection boundaries, not request counts: `N=10` produced 2 copying minors and one failure. The per-*collection* failure rate is much higher than the per-pass rate, so if the hunt wants a faster reproducer, drive collections rather than requests.

## Summary by CodeRabbit

- **Tests**
  - Added an optional warm-process soak test for Next.js route releases.
  - Repeated verification passes detect intermittent failures, collection issues, and copying-minor activity.
  - Reports observed failures and statistical confidence against residual failure rates.
  - Remains disabled by default and supports configurable pass counts.

- **Documentation**
  - Added release documentation covering configuration, validation scenarios, and expected behavior.


proggeramlug:
### Validation status vs current `main`, stated explicitly

The arm in this PR was validated against real provider images built from `b8d32ab19`, plus the pinned app dylib — every path exercised, both asserts sabotage-tested (see the PR body).

It **cannot currently be re-validated at HEAD**, and not for any reason belonging to this change: #8228 means the in-process LLVM backend fails 5 of the fixture's 104 modules on current `main` (`insertelement <2 x i64>` from #8204's module-init header compose has no case in `perry-codegen/src/dialect/mod.rs`), and `app-route.runtime.prod.js` is among them — so `fixture.sh` step 5 cannot produce an app dylib on `main` at all right now. That gates every arm of this fixture equally, not just the new one.

This PR is orthogonal to that: it adds a shell function and touches no compiled code, so it neither causes nor worsens #8228, and landing it does not make the fixture any less runnable than it already is. Flagging it so a future reader does not mistake "validated" for "validated against HEAD" — the artifacts it was proven on predate #8204.

PR #8220, opened 2026-08-16
fix(gc): root the Headers/FormData iteration frame slots (#8163, #8217)

### LINKED ISSUE #8217: gc: js_headers_for_each/keys/values/entries hold heap pointers across allocations and a JS callback
Found by code-reading `crates/perry-stdlib/src/fetch/headers.rs` while answering an ownership question during the #8163 residual hunt. **Reachability on the Next App Route path is NOT proven** — filing because the shape is unambiguous and it is the #7154 family, not because it is known to fire. Four sites, all in that one file:

## 1. `js_headers_for_each` — a raw closure pointer live across user JS

```rust
let cb_ptr = (cb_bits & 0x0000_FFFF_FFFF_FFFF) as i64;
let closure = cb_ptr as *const perry_runtime::ClosureHeader;
for (k, v) in entries {
    let v_ptr = js_string_from_bytes(...);   // allocates
    let k_ptr = js_string_from_bytes(...);   // allocates
    let v_nan = JSValue::string_ptr(v_ptr).bits();   // <-- v_ptr may have MOVED
    ...
    perry_runtime::js_closure_call2(closure, ...);   // <-- runs user JS; `closure` may have MOVED
}
```

Two defects in one loop:

* `closure` is a bare `*const ClosureHeader` in a Rust local, reused on every iteration, and `js_closure_call2` runs user JS — so a collection during iteration *n* leaves every later iteration calling a pre-move address. A `headers.forEach` whose callback allocates is the trigger.
* `v_ptr` is allocated, then `k_ptr` is allocated (a second collection point), and only *afterwards* is `v_ptr` read into `v_nan`. A collection inside the `k_ptr` allocation moves the value string, and the callback receives a dangling one.

## 2–4. `js_headers_keys` / `js_headers_values` / `js_headers_entries` — the array is live across allocations

```rust
let mut arr = perry_runtime::js_array_alloc(entries.len() as u32);
for (k, _) in entries {
    let k_ptr = js_string_from_bytes(...);        // allocates -> can move `arr`
    arr = perry_runtime::js_array_push_f64(arr, ...);  // stale `arr`
}
```

`arr` is held across `js_string_from_bytes`. The `arr = push(arr, …)` idiom handles the array *growing*, not the collector *moving* it. `js_headers_entries` is the worst of the three: `k_ptr` is live across the `v_ptr` allocation, `pair` is live across two `js_array_push_f64` calls (which allocate), and `arr` is live across the whole `pair` construction.

## Why the existing instruments miss it

* `scripts/gc_root_dominance_check.py` reads emitted LLVM IR and is structurally blind — these are Rust runtime locals.
* `scripts/gc_runtime_root_holders.py` enumerates `static` declarations; these are stack locals, not declarations.
* The Next App Route fixture's cold-start loop runs **zero copying minors in normal mode** (measured, #8215), so it could not surface this even if the path were hot.

## Fix shape

The same one #8131/#8211 applied elsewhere: `RuntimeHandleScope` + `across_*` around each allocation and each `js_closure_call*`, re-reading the closure pointer and the string/array pointers after every call that can collect. Note `raw_handle_debt.py` distinguishes reload shape (`across_*`) from argument position (`with_*_ptr`).

## Coordination

`crates/perry-stdlib/src/fetch/**` was excluded from the #8163 residual hunt at my request; I have released it. If the hunt's own `headers.rs` change turns out to be a candidate fix that covers some of these, it should absorb them; if it is a throwaway instrument, these four still want fixing and I will take them.

proggeramlug:
Closes #8217. Refs #8163, #8040.

## The holder

`js_headers_for_each` (`crates/perry-stdlib/src/fetch/headers.rs:452`) stripped the
NaN-box off its `callback` argument into a bare `*const ClosureHeader` **before** the
loop and never re-read it:

```rust
let closure = cb_ptr as *const perry_runtime::ClosureHeader;   // line 460
for (k, v) in entries {
    let v_ptr = js_string_from_bytes(v.as_ptr(), v.len() as u32);   // allocates
    let k_ptr = js_string_from_bytes(k.as_ptr(), k.len() as u32);   // allocates
    …
    perry_runtime::js_closure_call2(closure, …);                    // runs user JS
}
```

That copy lives only in a native Rust frame — callee-saved `x20` in the emitted code.
The precise root map does not cover Rust frames, and production resolves the
conservative stack scan to `SkipDisabled` (the copying minors in the reproducer all
report `eligible=true fallback=none`), so the collector never rewrites it. After the
first copying minor inside the loop, every remaining iteration calls a pre-move address.

## Why it produces an empty body and never a wrong value

The consumer is Next's `pipeToNodeResponse` header copy, present verbatim in the
fixture's compiled route chunk:

```js
c.headers.forEach((a, c) => { … b.appendHeader(c, d) … })
```

`js_closure_call2` → `get_valid_func_ptr` reads `CLOSURE_MAGIC` at
`closure + CLOSURE_TYPE_TAG_OFFSET` (12). Once the retired block has been recycled into
Eden and partly overwritten, the magic no longer matches, `get_valid_func_ptr` returns
null, and `dispatch_proxy_callee_or_throw` falls through to `throw_not_callable()` —
**`TypeError: value is not a function`**, thrown mid-header-copy. Next never writes the
body (`E180 failed to pipe response`) and the client sees an **empty body**, never a
wrong id/header/cookie: the response content was already correct, only the write was
abandoned. That is every discriminator #8163 recorded.

It also explains the ~10× gap between stale derefs and visible failures: a from-space
block is recycled into Eden and bump-allocated over gradually, so most stale derefs
still read an intact `CLOSURE_MAGIC` and call the right function.

## Evidence chain

| step | instrument | result |
|---|---|---|
| 1 | `PERRY_GC_PROTECT_FROMSPACE=1`, **default GC, no other knobs** | SIGBUS at `js_headers_for_each + 200` |
| 2 | fault report | `obj_type=4` (GC_TYPE_CLOSURE) `size=48`, fault at `user_ptr + 12` = `CLOSURE_TYPE_TAG_OFFSET` |
| 3 | `PERRY_GC_PROTECT_FROMSPACE_HOLDERS=1` | `(none in 90262 objects — the holder is outside the arena: a runtime side table, an FFI structure, or a register/stack slot)` |
| 4 | source | the `let closure = …` hoist above |

**Fault census, 12 protect runs on the unfixed binary: 8 faults, 51 copying minors, and
every single fault at `js_headers_for_each + 200` with `obj_type=4 size=48`.**
`retired_by_minor` spans #0/#1/#1/#1/#2/#3/#3, so they are distinct collections — one
holder, caught eight times.

## Measurements

Matched A/B on the mini (quiet M1, 8 GB): identical app dylib and provider-host, the two
arms differing **only** by this patch. Verified the right artifact was swapped — the
fixed `js_headers_for_each` is 4× larger and now performs a `TPIDRRO_EL0` read (the
`RUNTIME_HANDLE_STACK` TLS), which is *undefined* in the stdlib image and *defined* in
the runtime image, so the handles land on the same stack the runtime's scanner walks.

| arm | config | passes | failed batches | host `TypeError` | `[gc-copy-minor] ran` |
|---|---|---|---|---|---|
| before | protect, default GC | 12 runs | — | **8 faults** | 51 |
| after | protect, default GC | 6 runs | — | **0 faults** | 60 |
| before | default GC, no knobs, interleaved | 300 | **3** | **3** | — |
| after | default GC, no knobs, interleaved | 300 | **0** | **0** | — |
| before | default GC, `HEAP_LIMIT=8` | 75 | **1** | **1** | 6334 |
| after | default GC, `HEAP_LIMIT=8` | 75 | **0** | **0** | 6339 |
| before | forced+seeded (`SEED=8036`) | 30 | 0 | 0 | 3723 (1.035 M copied) |
| after | forced+seeded (`SEED=8036`) | 30 | 0 | 0 | 3719 (1.038 M copied) |

The protect arm is the discriminating one: it faults on **every** stale deref rather
than only the ones whose bytes were already clobbered. Under the before-arm rate
(8 faults / 51 minors) the probability of seeing 0 faults in the after arm's 60 minors
is ≈ 8 × 10⁻⁵. No second holder emerged in 360 protected passes.

Every one of the 375 after-arm passes is 21 requests checked byte-for-byte against the
Node production oracle — status, `x-perry-repro` header, `set-cookie`, and the full JSON
body — so ~7,900 request-level validations also cover the `keys`/`values`/`entries`
refactor. `bypassed routeModule.handle` = 0 throughout.

## Why the A/B base is `3b01efba4` and not merged main

`b8d32ab19` **cannot compile the fixture's app dylib** — five modules fail codegen on
`insertelement <2 x i64>`, the module-init header-image compose added by `bf8fd868e`
(#8204), which `crates/perry-codegen/src/dialect/mod.rs:1014` has no case for. Filed as
attribution on #8228. An app dylib compiled *before* that commit also does not run
against providers built *after* it (a clean-main control with no patch fails identically),
so the two cannot be mixed either.

The A/B therefore runs on `3b01efba4`, where `crates/perry-stdlib/src/fetch/headers.rs`
is **byte-identical** to `b8d32ab19` (`git diff 3b01efba4 b8d32ab19 -- …/headers.rs` is
empty) and `gc/roots/runtime_handles.rs` differs only by a doc comment — so the code
under test and the rooting machinery it uses are the same on both. The residual has also
been reproduced independently on merged-main providers on a second host, so it is not a
property of this base.

## Scope

Nine functions, all the same defect:

- `headers.rs` — `js_headers_for_each` (the one that fires on the Next path),
  `js_headers_keys`, `js_headers_values`, `js_headers_entries`,
  `js_headers_get_set_cookie`
- `body_metadata.rs` — `js_form_data_for_each` (a verbatim copy of the `for_each`
  defect), and three more the new test found on its first run that #8217 does not name:
  `form_data_string_array` (backing `js_form_data_keys`/`_values`) and
  `js_form_data_entries`

`arr = push(arr, …)` handled the array *growing*; it never handled the collector
*moving* it. `js_headers_entries` was worst: `k_ptr` live across the `v_ptr` allocation,
`pair` live across two allocating pushes, `arr` across all of it.

All six now root through `RuntimeHandleScope` — the idiom this file already uses on the
`Headers` constructor path (`read_headers_record_entries`,
`materialize_headers_init_iterable`, …) — and re-read after every call that can collect.

## Test

A source-shape scan, because the failure mode is a stale pointer rather than a crash:
reproducing it in a unit test means landing a collection inside one specific loop
iteration. This mirrors `no_allocation_is_taken_off_a_live_registry_borrow`, shipped in
#8211 for the same reason. It has two halves — a forbidden-binding list and a positive
"this entry point must open a scope" requirement, because the negative list alone is
satisfiable by renaming — and **both halves are sabotage-checked**: a planted pre-fix
binding must still match the forbidden list, and a planted unrooted body must still be
isolated by the body extractor.

## Why no existing instrument caught this

The holder is a **native Rust frame slot**, and all three static instruments are blind
to that shape by construction:

- `scripts/gc_runtime_root_holders.py` audits `static` / `thread_local!` declarations.
- `scripts/gc_root_dominance_check.py` reads emitted LLVM IR from compiled JS.
- `scripts/raw_handle_debt.py` counts bare reads out of handles that *already exist*, and
  only within `crates/perry-runtime/src` — a site that never rooted anything is invisible,
  and `perry-stdlib` is out of scope entirely.

A scan for the general shape ("a raw heap pointer bound, then used after another
allocating call") over `perry-stdlib` + the `perry-ext-*` crates reports ~150 candidate
sites. That is a ratchet-sized follow-up, not this PR.

## Note for reviewers: a pre-existing flaky test in this file

`fetch::tests::request_reads_release_the_registry_guard` (added in #8211) fails roughly
**1 run in 10** under the default parallel test harness, always with
`dispatch_request_property("cache") left the registry guard held`. It `try_lock`s
`REQUEST_REGISTRY` while its sibling `fetch_root_scanner_*` tests hold the same mutex on
another thread. Measured on this branch's test binary: 22/25 pass with all ten tests,
and **23/25 with my test excluded from the pool** — same rate — so it is pre-existing and
not caused by this change. It passes 100% under `--test-threads=1`. Worth its own fix
(serialise on a test-local mutex, or `try_lock` in a retry loop); `cargo-test` is a
required context, so it will intermittently red unrelated PRs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed failures when iterating over headers and form data during memory allocations or JavaScript callbacks.
  * Improved reliability of header keys, values, entries, cookies, and form-data iteration without changing public behavior or ordering.

* **Tests**
  * Added regression coverage to prevent similar runtime memory-handling issues.

### DIFF
--- a/changelog.d/8220-gc-headers-frame-roots.md
+++ b/changelog.d/8220-gc-headers-frame-roots.md
@@ -1,0 +1,56 @@
+### Fixed
+
+- **`headers.forEach` handed the collector's from-space to `js_closure_call2` — the
+  residual #8163 holder, and the one that is production-visible under the DEFAULT GC
+  (#8217).**
+
+  `js_headers_for_each` stripped the NaN-box off its `callback` argument into a bare
+  `*const ClosureHeader` **before** the loop and never re-read it. That copy lives only
+  in a native Rust frame — the precise root map does not cover Rust frames, and
+  production resolves the conservative stack scan to `SkipDisabled` — so the first
+  copying minor inside the loop left it naming retired from-space. Every iteration
+  allocates twice (`js_string_from_bytes` for the value and the key) and then runs
+  arbitrary user JS, so the window is wide open.
+
+  The consumer is Next's `pipeToNodeResponse` header copy, which the production App
+  Route fixture runs once per response: `c.headers.forEach((a, c) => … b.appendHeader(c,
+  d) …)`. `js_closure_call2` → `get_valid_func_ptr` reads `CLOSURE_MAGIC` at
+  `closure + CLOSURE_TYPE_TAG_OFFSET` (12). Once the retired block has been recycled into
+  Eden and partly overwritten the magic no longer matches, `get_valid_func_ptr` returns
+  null, and `dispatch_proxy_callee_or_throw` falls through to `throw_not_callable()` —
+  `TypeError: value is not a function`, thrown mid-header-copy, so Next never writes the
+  body (`E180 failed to pipe response`) and the client sees an **empty body**. That is
+  every discriminator #8163 recorded, including why the failure is never a wrong
+  id/header/cookie: the response content was already correct, only the write was
+  abandoned. It also explains the ~10× gap between stale derefs and visible failures — a
+  from-space block is recycled into Eden and bump-allocated over gradually, so most stale
+  derefs still read an intact magic and call the right function.
+
+  `js_headers_keys` / `_values` / `_entries` / `_get_set_cookie` hold `arr` (and
+  `entries` also `k_ptr` and `pair`) across the same allocations: the
+  `arr = push(arr, …)` idiom handles the array *growing*, not the collector *moving* it.
+  `js_form_data_for_each` is a verbatim copy of the `for_each` defect. All six now root
+  through `RuntimeHandleScope`, the idiom this file already uses on the `Headers`
+  constructor path, and re-read after every call that can collect.
+
+  **Attribution.** `PERRY_GC_PROTECT_FROMSPACE=1` under the *default* GC — no forced
+  evacuation, no seeded schedule — faults at `js_headers_for_each + 200` on a retired
+  `obj_type=4` (GC_TYPE_CLOSURE), `size=48`, at `user_ptr + 12`. Twelve protect runs on
+  the unfixed binary produced 8 faults over 51 copying minors, **every one at that single
+  site**; `PERRY_GC_PROTECT_FROMSPACE_HOLDERS=1` reported `(none in 90262 objects — the
+  holder is outside the arena: a runtime side table, an FFI structure, or a register/stack
+  slot)`. It is the stack slot.
+
+  **Measured**, matched A/B differing only by this patch, identical app dylib: protect
+  arm 8 faults / 51 minors → **0 faults / 60 minors**; default GC with no knobs 3 failed
+  batches / 300 passes → **0 / 300**; default GC under `PERRY_GC_HEAP_LIMIT=8` 1 failure
+  over 6,334 copying minors → **0 over 6,339**; forced+seeded (`SEED=8036`) unchanged at
+  30/30 with ~1.04 M objects copied per arm and zero evacuation-verify panics. No second
+  holder emerged in 360 protected passes.
+
+  **Why nothing caught it.** The holder is a native Rust frame slot, invisible to all
+  three static instruments by construction: `scripts/gc_runtime_root_holders.py` audits
+  `static`/`thread_local!` declarations, `scripts/gc_root_dominance_check.py` reads
+  emitted LLVM IR from compiled JS, and `scripts/raw_handle_debt.py` counts bare reads out
+  of handles that *already exist* and only inside `crates/perry-runtime/src`. A site that
+  never rooted anything, in `perry-stdlib`, is outside all of them.
--- a/crates/perry-stdlib/src/fetch/body_metadata.rs
+++ b/crates/perry-stdlib/src/fetch/body_metadata.rs
@@ -547,13 +547,19 @@ pub extern "C" fn js_form_data_for_each(handle: f64, callback: f64) -> f64 {
     if cb_ptr == 0 {
         return f64::from_bits(TAG_UNDEFINED);
     }
-    let closure = cb_ptr as *const perry_runtime::ClosureHeader;
+    // #8163: the raw `ClosureHeader*` and the first of the two strings are held
+    // across an allocation and across user JS. See `js_headers_for_each`.
+    let scope = perry_runtime::gc::RuntimeHandleScope::new();
+    let cb_handle = scope.root_nanbox_f64(perry_runtime::value::js_nanbox_pointer(cb_ptr));
     for (name, value) in entries {
+        let inner = perry_runtime::gc::RuntimeHandleScope::new();
         let name_ptr = js_string_from_bytes(name.as_ptr(), name.len() as u32);
+        let name_handle = inner.root_nanbox_u64(JSValue::string_ptr(name_ptr).bits());
         let value_ptr = js_string_from_bytes(value.as_ptr(), value.len() as u32);
-        let name_value = f64::from_bits(JSValue::string_ptr(name_ptr).bits());
         let value_value = f64::from_bits(JSValue::string_ptr(value_ptr).bits());
-        perry_runtime::js_closure_call3(closure, value_value, name_value, handle);
+        let closure = perry_runtime::js_nanbox_get_pointer(cb_handle.get_nanbox_f64())
+            as *const perry_runtime::ClosureHeader;
+        perry_runtime::js_closure_call3(closure, value_value, name_handle.get_nanbox_f64(), handle);
     }
     f64::from_bits(TAG_UNDEFINED)
 }
--- a/crates/perry-stdlib/src/fetch/body_metadata.rs
+++ b/crates/perry-stdlib/src/fetch/body_metadata.rs
@@ -145,15 +145,18 @@ unsafe fn form_data_value_string(value: f64) -> String {
 }
 
 fn form_data_string_array(values: Vec<String>) -> f64 {
-    let mut arr = perry_runtime::js_array_alloc(values.len() as u32);
+    // #8163: `arr` must survive the per-value string allocations below.
+    let scope = perry_runtime::gc::RuntimeHandleScope::new();
+    let arr_handle = scope.root_raw_mut_ptr(perry_runtime::js_array_alloc(values.len() as u32));
     for value in values {
         let value_ptr = js_string_from_bytes(value.as_ptr(), value.len() as u32);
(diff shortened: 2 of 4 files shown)


proggeramlug:
## Qualification on this PR's end-to-end numbers: they were measured on a pre-#8204 object layout

Raised by the #8163 author while trying to reproduce these numbers independently, and **verified here rather than taken on faith**:

- The A/B in this PR was measured on base `3b01efba4`. `git merge-base --is-ancestor bf8fd868e 3b01efba4` → **false**: that base is **pre-#8204**.
- This branch's base is **post-#8204** (`bf8fd868e`, "recover the instruction cost of the 56 B → 48 B header shrink"), which removed the derivable `object_type` and `field_count` header words.

I previously argued the cross-base transfer was sound because `crates/perry-stdlib/src/fetch/headers.rs` is blob-identical (`a3120f1600…`) at `3b01efba4`, `b8d32ab19` and `2e7a12818`, and `gc/roots/runtime_handles.rs` differs only in doc comments. **That argument is correct but insufficient.** It establishes that the *code under test* is identical; it does not establish that the *object layout the collector operates on* is. For a GC rooting/evacuation defect, a 56 B → 48 B header change is not a neutral variable — it moves every field offset the copying minor and the from-space fault reporter reason about.

**What still stands, unaffected:**
- The **attribution** — the fault at `js_headers_for_each + 200`, `obj_type=4 size=48`, `user_ptr + 12` = `CLOSURE_MAGIC`, `HOLDERS=1` reporting none-in-heap, and the hoisted `let closure` visible in the source and in `x20` — is a static property of the code, not of the header layout.
- The **defect is real on current main**: the hoisted raw pointer is held across two allocating calls and a callback that runs user JS, on `main` today, regardless of layout generation.
- The unit tests and `cargo fmt`/file-size gates ran on this branch's own (post-#8204) base.

**What is not yet true:** that the before/after was measured on the layout this code will actually run under. **#8228 is exactly what prevents making it true** — the in-process backend cannot compile 5 of the fixture's 104 modules on `main`, so the fixture cannot be built at HEAD.

**Interim workaround under test** (credit to the #8163 author): `PERRY_LLVM_INPROCESS=off` selects the mature text transport instead of native in-process construction (`native_units_mode()`, `crates/perry-codegen/src/native_emit.rs:61`, returns `NativeMode::Off` for `0|off|false`), and #8228 is a missing `insertelement` case in the **dialect reader**, which is only on the native path. If that holds, the matched post-#8204 re-run becomes a re-run rather than a blocker. They are running the 300-pass close condition on a matched pair on an independent host and will post the numbers to #8163.

**Reviewers:** please treat the fix and its attribution as reviewable now, and the end-to-end A/B as pending re-measurement on a matched layout. I would rather flag this myself than have it found in review.

One methodological note worth recording, since it cost the #8163 author a confusing failure: the provider-swap trick (pin the app dylib, rebuild only the providers) **only works within a header-layout generation**. Swapping post-#8204 providers under a pre-#8204 app dylib dies at startup with `TypeError: Cannot set properties of null or undefined (setting 'id')` — ABI skew, not a GC bug.

PR #8238, opened 2026-08-16
perf(codegen, runtime): stop recording parameter-guard visits that can never be consulted

### LINKED ISSUE #8202: perf(runtime): js_param_type_guard's interpretive walk dominates the remaining ~11pp (fixed per-call overhead now largely removed)
## Where this number comes from

Measured during the #8079 re-sweep at `bfb0707be` (quiet-host instructions retired, best-of-5, stdout byte-exact). A diagnostic arm that replaces every `js_param_type_guard` call with `true` (routing otherwise unchanged) prices the interpretive validator's total cost; #8201 recovers the scalar share by predicate-equivalent leaf guards. What remains is the **structural** share:

| bench | validator total | recovered by #8201 (scalar) | remaining (structural) |
|---|---:|---:|---:|
| interp | −19.6% | −8.3% | **−11.3pp (~1.70B instr)** |
| iso_miss | −16.3% | −7.0% | **−9.4pp (~1.66B instr)** |
| tree | −33.8% | −31.8% | −2.0pp (leaf-call + wrapper hop on 10.5M Tier-B self-recursion round-trips — that part is #8169) |
| tree_wide | −17.0% | −15.9% | −1.1pp (same, #8169) |

The interp/iso_miss residual is concentrated in exactly two wrappers: `asNum(v: Value)` (a 3-arm union of small object literals, called per arithmetic op in `evalNode`) and `peek(p: Parser)` (an object whose `toks: Token[]` walk validates every element per call).

## Why the per-call cost is mostly fixed overhead

`js_param_type_guard` (crates/perry-runtime/src/param_type_guard.rs) pays, on every invocation:

1. `Descriptor::parse` — re-validates magic/root/offset-table of an **immutable, compiler-emitted rodata blob** that cannot change between calls;
2. `GuardState` construction — a 768-byte `inline_visited: [(usize, u32); 64]` zero-init plus spill bookkeeping, even when the value is a 2-field `{kind, num}` object whose walk touches 3 nodes;
3. `seen_or_insert` — a linear scan of up to 64 entries per node visited.

## Candidate shapes (in the order I would try)

1. **Lazy `GuardState`**: construct the visited set only when the first container op actually recurses; a flat object/small union check needs none of it.
2. **Parse-once descriptor cache**: the descriptor pointer is a rodata address — key a one-time-validated `Descriptor` by pointer (append-only table; the blob is immutable), skipping step 1 per call.
3. **Caller-side static proofs** for the `evalNode -> asNum(l)` shape, so proven sites bypass the wrapper entirely (`spec_return_proofs` already exists for return values; `asNum`'s argument is `evalNode`'s return).

Deliberately separate: #8169 (Tier-B guarded clones re-entering through their own public wrapper) owns the tree/tree_wide residual.

Refs #8079, #8201, #8169.

proggeramlug:
Refs #8202.

`js_param_type_guard` runs on every unproven call into a guarded ordinary-parameter clone. #8201 moved the scalar descriptors to the typed-abi leaf guards; what stayed on the interpretive validator is the structural half, and #8202 priced its **fixed** per-call component: descriptor re-parse, a ~1 KB `GuardState` zero-init, and a linear `seen_or_insert` scan per node visited.

This removes two of those three.

## The visited set was unconditional

Every container the walk touched went into the set — **per array element**. Validating `p: { toks: Token[], pos: number }` on every `peek(p)` therefore recorded one entry per token, and not one of them could ever be consulted: `Token` lies on no descriptor cycle and is reachable by exactly one path, so a second arrival at the same `(address, node)` pair is impossible.

The set is load-bearing for exactly two facts, and both are properties of the immutable compiler-emitted graph rather than of the value:

* **termination** — a value cycle (`env.parent === env`) can only walk forever through a node that reaches itself;
* **no re-walk blowup** — a node the traversal can enter twice with the same address must memoize, or a shared graph re-walks exponentially.

So the compiler decides it. `visit_tracking_bits` runs Tarjan over the graph it just built and propagates a saturating "ways in" count from the root, then sets the high bit of the op byte on exactly the container nodes that need recording. `js_param_type_guard` masks the op byte and reads the bit.

On `interp.ts` that is: `peek(p: Parser)` — 6 nodes, **0 tracked**, was one record per token; `asNum(v: Value)` — 123 nodes, **15 tracked**, exactly the recursive `Node`/`Env` cluster, 108 nodes stop recording. Descriptor length is unchanged (the bit rides in a byte that only ever held ops 0–16).

The magic goes `PGT1` → `PGT2`, so a mismatched compiler/runtime pair fails closed on the magic — guard returns 0, caller takes the generic function — instead of reading a v1 blob as one that opts out of tracking everywhere.

## `GuardState` zeroed 1 KB of stack per call

`inline_visited` is now `MaybeUninit`. Only `[..inline_visited_len]` is ever read, and after the change above most guarded calls never write a slot at all.

## Measured

Instructions retired (`/usr/bin/time -l`), best-of-3, both arms built from their own tree with the same `-p perry -p perry-runtime-static -p perry-stdlib-static`, `PERRY_RUNTIME_DIR` pinned per arm, all 19 corpus stdouts byte-compared and exit-checked. `iso_miss` keeps `misses 0`.

**Exactly 2 of the 19 rows emit a `js_param_type_guard` call site** — `interp` and `iso_miss`, two sites each (`asNum`, `peek`). Both improve:

| row | instructions | peak RSS |
|---|---:|---:|
| `interp` | 13.814 B → 13.589 B (**−1.63%**) | −0.06% |
| `iso_miss` | 16.489 B → 16.264 B (**−1.36%**) | +0.13% |

Isolating the validator's *own* cost, by differencing against the same runtime archive (`PGT1` blob + `PGT2` runtime = every guard fails, so binary-layout effects cancel):

| row | validator cost, `main` | with this change | delta | share of program |
|---|---:|---:|---:|---:|
| `interp` | 1.671 B | 1.422 B | **−14.9%** | 12.05% → 10.45% |
| `iso_miss` | 1.611 B | 1.400 B | **−13.1%** | 9.76% → 8.60% |

### The other 17 rows are not attributable, in either direction

The two arms' `libperry_runtime.a` differ in **exactly two functions out of 11,185** in the crate's codegen unit — `js_param_type_guard` (808 → 316 bytes) and `GuardState::matches` (+28) — every other function byte-identical. The 17 rows that emit no guard call site execute neither, so their movement (`pipeline` −3.9%, `retain_wide1` +0.6%, `retain1` +0.3%, `deeplist` −0.5%, `retain` −0.3%, the remaining 12 within ±0.1%) is address-layout noise, not effect. Two `main` builds from identical source came out **byte-identical** (archive and `perry` binary alike), and repeat runs of one binary spread ~0.1%, so the build itself is deterministic and `pipeline`'s ±4% is what an address-hash-sensitive program does when the heap moves. I am claiming none of it.

## Tests

`a_tree_shaped_descriptor_records_no_visits`, `a_container_on_a_cycle_records_its_visits`, `a_shared_container_records_its_visits` (codegen, through the real encoder); `a_tracked_node_terminates_on_a_cyclic_value` — which also asserts the *untracked* form of the same descriptor conservatively returns 0 rather than hanging — `an_untracked_shared_node_decides_the_same_way`, and `the_previous_descriptor_format_is_refused` (runtime).

`cargo test --release -p perry-runtime --lib -p perry-codegen` green.

## What this does NOT fix

#8202's premise was that the fixed per-call overhead dominates. It does not: it is ~15% of the validator's cost, and the validator is ~12% of `interp`. The structural walk is the other ~10.5pp, and the measurement in the issue thread shows the specialization it gates is worth ~0.2%. That is a policy question for #8094/#8079, not a per-call-overhead one — see the issue comment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parameter validation for cyclic and shared data structures.
  * Added safeguards to limit repeated traversal and prevent excessive validation work.
  * Invalid or outdated guard descriptors now fail closed instead of being accepted.

* **Performance**
  * Reduced runtime tracking overhead by enabling visit tracking only where necessary.
  * Added bounded handling for deeply repeated or duplicated values.


proggeramlug:
## Validation

- `cargo test --release -p perry-runtime --lib -p perry-codegen --lib` — green. Every pre-existing guard test still passes alongside the new ones: `scalar_descriptor_rep_classifies_exactly_the_leaf_guard_ops`, `recursive_alias_serializes_as_a_finite_graph`, `a_class_descriptor_carries_its_class_id_and_its_declared_fields`, and the four `ordinary_param_guard_tests`.
- **Gap corpus A/B, Perry vs Perry.** All 37 `test_gap_*.ts` that declare a structural parameter type — the only shapes that can reach the validator — compiled under both arms and byte-compared on stdout, stderr **and** exit code: **37 identical, 0 differing, 0 compile failures**. Seven of them genuinely exercise the subject rather than merely not crashing, 30 guard call sites in total: `test_gap_specabi_ordinary_param_guards` (7), `test_gap_7890_declared_array_receiver_element_read` (7), `test_gap_declared_field_type_refine_guarded` (5), `test_gap_repsel_element_shape_loop_clone` (4), `test_gap_7891_string_receiver_numeric_string_key` (3), `test_gap_repsel_element_shape_param_binding` (2), `test_gap_gc_alloc_point_no_move` (2).
- 19-program perf corpus: all stdouts byte-compared, `iso_miss` keeps `misses 0`.
- `cargo fmt --all -- --check`, `scripts/check_file_size.sh`, `addr_class_inventory.py`, `gc_runtime_root_holders.py`, `check_thread_locals.py` — all clean.

Not run here: the full 569-test gap suite against node, and `crates/perry-codegen/tests/native_proof_regressions.rs`. The dev host is saturated (load average 110+ from unrelated jobs) and the gap runner was getting ~0.5 s of CPU per 10 minutes, so I stopped it rather than add to the contention; the 37-test structural subset above is the part of it that can reach this code.

### Why the behavioural surface is narrow

The bit only controls memoization, so it cannot make the guard accept something `main` rejects. Over-tracking reproduces today's behaviour exactly. Under-tracking can only cost a memo hit — the walk then depth-caps and returns 0, the caller takes the generic function, and the observable result is unchanged. The magic bump is symmetric, so a mismatched compiler/runtime pair also just routes everything generic; that is the configuration the `interp`/`iso_miss` decomposition above measures, and it prints the same checksums.

PR #8250, opened 2026-08-16
fix(runtime): preserve structured clone graph identity

### LINKED ISSUE #8232: [parity] structuredClone: cycles lose reference identity, and non-cloneable values throw TypeError instead of DataCloneError
Split out of #6766 after PR #8229. Not a `perf_hooks` bug — `perf_hooks` is just where the node-suite happens to pin it, because mark/measure `detail` is passed through `structuredClone`.

Two gaps in Perry's structured clone:

**1. Cycles are not preserved.** The clone succeeds (no throw, which is already correct — a cycle is cloneable) but the reference identity inside the copy is lost.

`test-parity/node-suite/perf_hooks/mark/detail-circular`:
```js
const o: any = {}; o.self = o;
const mark = performance.mark("c", { detail: o });
mark.detail !== o          // node true,  perry true   ✅
mark.detail.self === mark.detail   // node true,  perry false  ❌
```
`test-parity/node-suite/perf_hooks/measure/detail-clone-failures` hits the same thing via `performance.measure("cycle", { start: 0, duration: 1, detail: cycle })`.

The fix is the standard one: a memo map from source object to its clone, consulted before cloning each reference, so a cycle resolves to the already-created copy.

**2. A non-cloneable value throws the wrong error type.** Node raises a `DataCloneError` `DOMException`; Perry raises a plain `TypeError`.

`measure/detail-clone-failures`:
```
node:   function: DataCloneError
perry:  function: TypeError
```
`crates/perry-runtime/src/event_target.rs` already has `js_dom_exception_new(message, name)` and maps `"DataCloneError"` to legacy code 25, so this is a matter of routing the existing throw through it.

Both cases are exercised by `./run_parity_tests.sh --suite node-suite --module perf_hooks`, but the underlying behaviour is `structuredClone`'s and affects every caller.

proggeramlug:
## Summary

- replace the recursion-only cycle guard with a GC-visible source-to-clone memo
- preserve cycles and repeated reference identity across objects, arrays, maps, sets, regexps, and buffers
- throw DataCloneError as a DOMException and remove the perf_hooks TypeError workaround
- clear traversal roots after success and failure to avoid retaining cloned graphs

## Tests

- cargo fmt --all -- --check
- cargo test -p perry-runtime structured_clone_preserves_cycle_identity --lib
- ./run_parity_tests.sh --suite node-suite --module perf_hooks --filter detail-circular
- ./run_parity_tests.sh --suite node-suite --module perf_hooks --filter detail-clone-failures

Closes #8232

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `structuredClone` handling for cyclic references, preserving object identity across nested arrays, objects, sets, maps, buffers, and regular expressions.
  * Improved garbage-collection safety during cloning.
  * Clone failures now report the correct `DataCloneError`.
  * Transferable values are detached only after cloning completes successfully.
  * Structured cloning no longer incorrectly rejects closures with a `TypeError`.
* **Documentation**
  * Added documentation covering graph identity preservation and cloning error behavior.

### DIFF
--- a/changelog.d/8250-structured-clone-graph-identity.md
+++ b/changelog.d/8250-structured-clone-graph-identity.md
@@ -1,0 +1,18 @@
+**`structuredClone` preserves graph identity.**
+
+The recursion-only cycle guard is replaced by a GC-visible source-to-clone memo, so
+cycles and repeated references keep their identity across objects, arrays, maps,
+sets, regexps and buffers — a value reached twice now clones once and both edges
+point at the same clone, as the structured-clone algorithm requires. Clone failures
+raise `DataCloneError` as a `DOMException`, which let the `perf_hooks` `TypeError`
+workaround go, and traversal roots are cleared on both the success and failure paths
+so a completed clone does not retain the graph it walked.
+
+The memo's entries are visited as mutable roots (`scan_structured_clone_memo_roots_mut`)
+and its address-keyed lookup is rebuilt after visitation, because a copying collection
+relocates either side. Registering that scanner is what the map clone test now does
+explicitly: the GC test harness clears `MUTABLE_ROOT_SCANNERS` so a collection sees
+exactly the roots a test installs, so without it the memo is decorative under test and
+a clone's address is never rewritten across a forced copying minor.
+
+Closes #8232.
--- a/crates/perry-runtime/src/builtins/globals.rs
+++ b/crates/perry-runtime/src/builtins/globals.rs
@@ -376,19 +376,22 @@ fn hex4_chars(cs: &[char]) -> Option<u16> {
 // structuredClone
 // ============================================================
 
-// Cycle-detection state for `js_structured_clone` (#1512). Tracks the source
-// pointers currently mid-clone on this thread. On re-entry for a pointer
-// already in the set, we return the original value rather than recursing
-// — that breaks the spec's "preserve reference identity" guarantee but
-// keeps cycles from infinite-recursing into a stack overflow, which is
-// what previously caused `performance.mark("n", { detail: o.self = o })`
-// to crash. Full reference-identity preservation would need a src→dst
-// map; deferred until a real user-facing need surfaces.
 thread_local! {
-    static STRUCTURED_CLONE_IN_PROGRESS: std::cell::RefCell<std::collections::HashSet<usize>>
-        = std::cell::RefCell::new(std::collections::HashSet::new());
+    static STRUCTURED_CLONE_MEMO: std::cell::RefCell<StructuredCloneMemo>
+        = std::cell::RefCell::new(StructuredCloneMemo::default());
     static STRUCTURED_CLONE_TRANSFER_STATE: std::cell::RefCell<Option<StructuredCloneTransferState>>
         = const { std::cell::RefCell::new(None) };
+}
+
+#[derive(Default)]
+struct StructuredCloneMemo {
+    entries: Vec<StructuredCloneMemoEntry>,
+    by_source: std::collections::HashMap<usize, usize>,
+}
+
+struct StructuredCloneMemoEntry {
+    source: u64,
+    clone: u64,
 }
 
 #[derive(Default)]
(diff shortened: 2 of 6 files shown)


proggeramlug:
Holding this one — it regresses a GC rooting test that main passes.

## The failure

```
gc::tests::runtime_roots::hook_dispatch_handles::test_structured_clone_map_runtime_handles_survive_nested_copied_minor_gc
crates/perry-runtime/src/gc/tests/runtime_roots/hook_dispatch_handles.rs:374
assertion failed: crate::map::is_registered_map(cloned_map as usize)
```

`cargo test -p perry-runtime --lib` on this branch merged with current `main`: **2546 passed, 1 failed.**

## It is this branch, not a pre-existing failure

The same test passed on two other `main`-based trees I ran today, within the hour:

- the #8238 tree — `... survive_nested_copied_minor_gc ... ok`
- the #8245 tree — `... survive_nested_copied_minor_gc ... ok`

Both were `main` + one unrelated PR, so `main` passes it and this branch does not.

## What the test pins

It clones an outer `Map` holding an inner `Map`, with `test_force_next_map_helper_gc()` armed twice so a **copying minor runs inside the nested clone**, then asserts the returned pointer is still a *registered* map. So the assertion fails either because the returned pointer is stale (the clone moved and the value handed back predates the move) or because the clone never enters `MAP_REGISTRY` on the new path. `MAP_REGISTRY` is address-keyed, which makes it the same hazard class you already handled correctly for `by_source` — the memo scanner rewrites `entry.source`/`entry.clone` and then rebuilds the address-keyed index. Worth checking whether the map registry needs the equivalent treatment across the copying minor this test forces.

I traced the obvious candidate and it is **not** the top-level return: the Map branch does re-read `structured_clone_memo_value(memo_index)` before returning, and the buffer/regexp branches that return a local `cloned` are leaf clones with no collection point between `record` and `return`. So I have not localized it, and I would rather hand you the reproducer than guess further.

## Everything else is clean

- all four script gates pass: `gc_runtime_root_holders`, `shape_descriptor_census`, `check_thread_locals`, `addr_class_inventory`
- `check_file_size` passes
- merges into current `main` with zero conflicts

The design reads well otherwise — registering `scan_structured_clone_memo_roots_mut` in the same commit is right, visiting both sides as mutable roots is right, and rebuilding `by_source` after visitation is the subtle part that is easy to miss. Every memo borrow is confined to an allocation-free helper, so there is no double-borrow-during-GC hazard. Happy to re-run the suite as soon as the map case is addressed.

proggeramlug:
Re-checked against a newer `main` so the blocker isn't stale.

`main` has since taken **#8251** (`fix(gc): reject fabricated Map/Set headers in plausible_gc_header`), which tightens Map/Set classification to require `size == 24`. That is close enough to this failure's subject that it was worth ruling in or out. Merged this branch onto that `main` (conflict-free) and re-ran:

```
test_structured_clone_map_runtime_handles_survive_nested_copied_minor_gc ... FAILED
structured_clone_preserves_cycle_identity ... ok
structured_clone_keeps_overflow_properties ... ok
test result: FAILED. 2 passed; 1 failed
```

So it is **not** an interaction with #8251 — the regression is this branch's, and it survives the newer classification path. Worth noting the other direction too: your own two new tests pass, so the PR does what it says on cycles and overflow properties; it is specifically the "cloned Map is still a registered map after a copying minor runs *inside* the nested clone" case that breaks.

No other change needed from my side — flagging so you don't spend time on the #8251 hypothesis. Everything else I reported still holds: all script gates green, `check_file_size` green, merges clean.

proggeramlug:
Merging — and the blocker turned out **not** to be a rooting bug in this PR. I owe you the correction, since I flagged it twice.

## What was actually wrong

`test_structured_clone_map_runtime_handles_survive_nested_copied_minor_gc` failed because the **GC test harness clears `MUTABLE_ROOT_SCANNERS`** so a collection sees exactly the roots a test installs — that is documented on `register_runtime_handle_root_scanner_for_tests`. The test installed the runtime-handle scanner and nothing else, so `scan_structured_clone_memo_roots_mut` never ran and the memo was decorative *under test*.

I measured it rather than reasoned about it. Instrumenting the Map return:

```
MEMODIAG handle=0x200015201a0 memo=0x200008800c8 equal=false
```

The map moved; the handle tracked it, the memo did not. Then a probe in the scanner produced no output at all, and a probe at the `reg_scanner!` site produced none either — `gc_init`'s registration block simply does not run in a unit test. So the memo could not have been rewritten there by construction.

Registering the memo scanner in the test makes it pass, with no production change:

```
cargo test -p perry-runtime --lib   →  2557 passed; 0 failed; 4 ignored
```

**So the memo design is sound where `gc_init` installs the scanner.** My earlier framing — "the regression is this branch's" — was right that the branch changed behaviour (main returned a handle-backed re-read, this returns a memo-backed one) but wrong to imply the memo was unrooted in production. It isn't.

## What I pushed

1. `test(gc)` — installs the memo scanner in that test, with a comment explaining why the harness requires it.
2. `docs(changelog)` — the PR had no `changelog.d/` fragment, and fragments are what become the release notes.

## One thing genuinely worth following up

Because the harness installs only declared scanners, **the memo's rooting had no test coverage at all** before this. That test is now the only thing exercising it, and it covers the Map arm only. Seven other sites use `structured_clone_memo_value` as a live-pointer source — the Set return, both Array mid-loop reads and its return, and both Object mid-loop reads and its return. Each is sound on the same assumption and none is covered. A follow-up that registers the scanner in a shared helper and adds a cycle-through-a-moved-object case for the Array and Object arms would close that.

Verified on top of current `main`, conflict-free: runtime suite **2557/0/4**, `cargo fmt`, `check_file_size`, and the five script gates including the new `unrooted_local_shape` ratchet.

Issue #8233, opened 2026-08-16
gc: the raw-handle ratchet cannot see unrooted locals in perry-stdlib / perry-ext-* — 90% of that surface never adopted the rooting API


proggeramlug:
Found while filing #8217. That issue is four instances; this is the reason nothing was going to catch them, or the next four.

## The gap

`scripts/raw_handle_debt.py` is the repo's instrument for "a raw heap pointer copy held across a collection point". It has two properties that compound:

```python
SRC = ROOT / "crates" / "perry-runtime" / "src"
PAT = re.compile(r"\.get_raw_(?:mut|const)_ptr\b")
```

1. **Scope is `perry-runtime` only.** `perry-stdlib` and every `perry-ext-*` crate are outside it.
2. **It counts bare reads *out of a `RuntimeHandle`*** — i.e. it measures debt in code that has *already adopted* the rooting API and then degraded a use. Code that never roots at all has no `get_raw_*_ptr` to count and scores **zero**, the ratchet's best possible result.

So a file holding four unrooted heap pointers with no `RuntimeHandleScope` anywhere is indistinguishable, to this instrument, from a perfectly rooted one. Its own docstring is careful to call it "a DEBT COUNTER, not a soundness proof" — this issue is not that it overclaims, it is that its *denominator* excludes the surface where the bugs are being found.

## Scale

Files under `crates/perry-stdlib/src` and `crates/perry-ext-*/src` that call an allocator (`js_string_from_bytes`, `js_array_alloc`, `js_object_alloc`, `js_closure_alloc`, `alloc_string`, `alloc_buffer`, `js_array_push*`) or invoke JS (`js_closure_call*`):

```
files that allocate or call JS:                    177
  of those, with NO rooting-API usage at all:      161   (90%)
```

Top density, zero rooting: `streams.rs` (72), `crypto/sign.rs` (66), `lodash.rs` (57), `mysql2/result.rs` (51), `ioredis.rs` (50), `perry-ext-fetch/lib.rs` (50), `perry-ext-events/lib.rs` (40).

**This is an exposure surface, not a bug count** — most of those sites allocate and return without holding anything across a second collection point, which is fine. The number says how much of the surface no instrument is watching, not how much is broken.

## Two independent confirmations that the shape is really there

Sampled two files, both hit:

* **`fetch/headers.rs`** (#8217) — four sites, including a bare `*const ClosureHeader` reused across `js_closure_call2` in a loop.
[...]

PR #8253, opened 2026-08-16
fix(gc): instrument unrooted locals in perry-stdlib and perry-ext-*

proggeramlug:
Refs #8233. Builds the instrument that issue asks for, as a report-then-ratchet. Does **not** fix the sites it finds — that is the campaign #8233 says wants an owner's priority call.

## The gap

`scripts/raw_handle_debt.py` is the repo's instrument for "a raw heap pointer copy held across a collection point", and two properties compound:

- **Scope is `perry-runtime` only.** `perry-stdlib` and every `perry-ext-*` crate are outside it.
- **It counts bare reads *out of* a `RuntimeHandle`** — debt in code that has *already adopted* the rooting API and then degraded a use. Code that never roots at all has no `get_raw_*_ptr` to count and scores **zero**, the ratchet's best possible result.

So a file holding four unrooted heap pointers with no `RuntimeHandleScope` anywhere is indistinguishable, to that instrument, from a perfectly rooted one. Its docstring is careful to call itself a debt counter rather than a soundness proof; the problem is its *denominator*, not its claim.

Nothing else covers it either. `gc_root_dominance_check.py` reads emitted LLVM IR and is structurally blind to Rust runtime locals. `gc_runtime_root_holders.py` enumerates `static`/`thread_local!` declarations — these are stack slots — and names the ext crates in its own "cannot see" list.

## What this adds

`scripts/unrooted_local_shape.py` detects the *shape*: a local bound from an allocator return, used again after an intervening call that can allocate or run JS.

```
unrooted-local shape: 218 finding(s) across 51 file(s)
    27  crates/perry-stdlib/src/mysql2/result.rs
    16  crates/perry-stdlib/src/crypto/sign.rs
    12  crates/perry-stdlib/src/sqlite/better.rs
    11  crates/perry-stdlib/src/events/events_on.rs
    10  crates/perry-stdlib/src/streams.rs
```

**This is an exposure surface, not a bug count.** Most of those sites allocate and return without holding anything across a second collection point. The number says how much of the surface no instrument was watching.

## Validated against independent ground truth, not just its own fixtures

It reproduces both sites #8233 names by hand. The `events_on.rs` one exactly:

```
events_on.rs:40: `state`  bound at :37, may have moved at :38
events_on.rs:40: `buffer` bound at :38, may have moved at :39
```

That is the case the issue calls out as worse than a transient misread — the push writes a **pre-move address into the heap**, so the damage outlives the frame. (Reading it for this PR turned up a second latent problem at the same site: `js_array_push_f64` returns a possibly-reallocated pointer and the code discards it with `let _ =`. A fix needs to thread that return, not only root the locals.)

## It can go red, and I checked rather than assumed

Per CLAUDE.md's four ways a gate can be unable to fail:

- **Sabotage test.** Planting one stale-local site: `--check` exits **1** with `REGRESSION: 219 findings exceeds baseline 218`. Clean tree exits **0**.
- **`--self-test`** plants a stale-local site and asserts it is flagged, *and* asserts an allocation-free function is not — so a detector that silently stops matching fails loudly instead of reporting a comfortable zero.
- **No `continue-on-error`**, and it is wired into `lint`, which is a required context.
- **Ratchet, not a zero target.** It fails only on an *increase* over `scripts/unrooted_local_shape_baseline.json`, so it is green on merge — a new gate has never been green, and promoting one that isn't blocks every open PR.

## Honest limits

Rust has no effect system marking "this call may allocate", so the collection-point list is a curated denylist and binding detection is line-order over source text. Expect false positives where an allocation provably cannot collect, and false negatives wherever a pointer flows through a shape this does not spell. Functions that open a `RuntimeHandleScope` are skipped — those belong to `raw_handle_debt.py`'s denominator, not this one's. The value is that the number can only be paid down, and that the two crate families are now *in* a denominator at all.

## Summary by CodeRabbit

* **New Features**
  * Added automated detection for potentially unsafe unrooted local allocation patterns.
  * Added baseline tracking, regression checks, verbose reporting, and self-testing capabilities.

* **Documentation**
  * Documented detection coverage, baseline behavior, known limitations, and reported findings.

* **Chores**
  * Integrated the check into the continuous integration lint workflow across supported library and extension components.


proggeramlug:
Correcting one claim in the description before this goes in, since it is a PR about gates that cannot fail and I should hold it to its own standard.

I wrote *"wired into `lint`, which is a required context."* That is imprecise. `lint` is **not** a required context — `pr-gate` is the only one. The accurate chain is:

- the `gate` job lists `lint` among its `needs` and renames itself `pr-gate` on `pull_request`;
- it exits 1 on any need whose result is `failure` or `cancelled`;
- it deliberately tolerates `skipped`, because the plan decides what runs — which is exactly the shape that has burned this repo before, so it matters whether `lint` can be skipped;
- `scripts/ci_plan.py:75` has `"lint": ("pr", "sweep", "full")`, it is special-cased on at line 303 for the `pr` tier, and the self-test asserts `docs-only PR: lint on`.

So `lint` runs in every tier and the path from a red step here to a blocked merge is unbroken — but via `pr-gate`'s fan-in, not because `lint` is required itself. The changelog fragment now says this.

Everything else in the description stands as written and was checked rather than assumed: sabotage gives `REGRESSION: 219 findings exceeds baseline 218` and exit 1, the clean tree exits 0, `--self-test` asserts both directions, and the detector reproduces both ground-truth sites named in the issue.

Issue #8234, opened 2026-08-16
[parity] node:perf_hooks timerify — async settlement, construct, entry arguments, and name/length descriptors


proggeramlug:
Split out of #6766 after PR #8229, which made `timerify`'s histogram option real (`timerify(fn, { histogram })` now validates the handle and records durations) but left five behaviours unimplemented. All five live in `perf_timerify_wrapper` / `js_perf_timerify` in `crates/perry-runtime/src/perf_hooks.rs`.

Reproduce with `./run_parity_tests.sh --suite node-suite --module perf_hooks`.

## 1. Async settlement — the wrapper records at return, not at settle (2 cases)

Node defers the measurement until the returned promise settles. Perry records synchronously, so an async function's entry/sample lands one turn early.

`timerify/async-fulfill`:
```
node:   pending records: 0 / result: 5 / settled records: 1
perry:  pending records: 1 / result: 5 / settled records: 0
```
`timerify/histogram-async`:
```
node:   pending: 0 / result: 6 / settled: 1 / positive: true
perry:  pending: 1
```
Needs the wrapper to detect a thenable result and attach a native continuation that emits the entry (and records into the optional histogram) on settlement.

## 2. `new` through the wrapper (`timerify/constructor`)

```js
class Value { marker = 1 }
const Wrapped = timerify(Value);
new Wrapped()          // node: a Value instance, marker 1;  perry: instance false, marker undefined
Wrapped()              // node: TypeError (class called without new);  perry: "call ok"
```

## 3. Entry arguments (`timerify/entry-arguments`)

Node's function entry carries the call arguments both as `entry.detail` (an array) and as indexed properties on the entry itself:
```js
`${entry.detail.length}:${entry.detail[0]}:${entry.detail[1] === argument}:${entry[0]}:${entry[1] === argument}`
node:   arguments: 2:1:true:1:true
perry:  TypeError: Cannot read properties of null (reading 'length')   (detail is null)
```

## 4. `name`/`length` descriptor flags (`timerify/name-length-descriptors`)

The values are already right; the attributes are not. Node defines both as enumerable and non-configurable on the wrapper:
```
node:   name timerified sample false true false     (value, writable, enumerable, configurable)
        length 2 false true false
perry:  name timerified sample false false true
        length 2 false false true
```
This one needs per-property descriptor attributes on a closure, so it may be blocked on the same machinery as #8231.

PR #8254, opened 2026-08-16
fix(perf_hooks): implement class prototype machinery

### LINKED ISSUE #8231: [parity] node:perf_hooks — the perf classes have no real prototype: no illegal-constructor guard, Symbol.toStringTag, or descriptors
Split out of #6766 after PR #8229 took `node:perf_hooks` from 86/148 to 132/148. These six cases all need the same missing machinery — real prototype objects for the native perf classes, carrying brand checks and per-property descriptors — so they are one piece of work rather than six.

Perry models the perf classes as bound-method closures and their instances as plain shaped objects. That gets `instanceof` right (via the class-id chain) but leaves no prototype object to hang `Symbol.toStringTag`, descriptors, or a construct trap on.

## Failing cases

`test-parity/node-suite/perf_hooks/`, reproduce with `./run_parity_tests.sh --suite node-suite --module perf_hooks`.

**`global/illegal-constructors`** — the classes are not constructible in Node:
```
node:   Performance: TypeError ERR_ILLEGAL_CONSTRUCTOR   (same for PerformanceEntry,
        PerformanceMeasure, PerformanceObserverEntryList, PerformanceResourceTiming)
perry:  Performance: no throw
```

**`shapes/prototype-tags`** — `Symbol.toStringTag` on each prototype:
```
node:   mark [object PerformanceMark] / measure [object PerformanceMeasure] / observer [object PerformanceObserver]
        PerformanceMark PerformanceMark   (and the same for Measure/Observer/ObserverEntryList/ResourceTiming)
perry:  mark [object Object] / measure [object Object] / observer [object Object]
        (the Ctor.prototype descriptor reads produce nothing)
```

**`shapes/public-descriptors`** — `Object.getOwnPropertyDescriptor(Object.getPrototypeOf(performance), "now")` etc:
```
node:   now function true true true          (value/writable/enumerable/configurable)
        timeOrigin function undefined true true          (an accessor)
        supportedEntryTypes function undefined false true
perry:  TypeError: Cannot read properties of undefined (reading 'value')
```

**`resource-timing/illegal-constructor-tag`**:
```
node:   tag: PerformanceResourceTiming / TypeError ERR_ILLEGAL_CONSTRUCTOR
perry:  tag: undefined / no throw
```

**`mark/direct-constructor`** — `new PerformanceMark("direct", { startTime: 7, detail })` is the one perf class Node *does* let you construct:
```
node:   shape: direct mark 7 0 / instances: true true / detail cloned: {"value":1} true
perry:  shape: undefined undefined undefined undefined / instances: false false
```

**`observer/entry-list-validation`** — the wrong-receiver arm, which needs a real `PerformanceObserverEntryList.prototype.getEntries` with a brand check:
```
node:   wrong receiver TypeError ERR_INVALID_THIS
perry:  wrong receiver TypeError undefined
```
(the other three arms in that file pass)

## Where

`crates/perry-runtime/src/perf_hooks.rs` (entry shapes, `is_perf_entry_object_instance_of`), `crates/perry-runtime/src/object/native_module/callable_exports.rs` (`nm_attach_perf_hooks`, which is where `PerformanceObserver.supportedEntryTypes` is already hung off the class closure as a dynamic prop — the same hook could carry a prototype).

proggeramlug:
Closes #8231.

## Summary

- install real `perf_hooks` constructor prototypes, inheritance, descriptors, getters, methods, and `Symbol.toStringTag` metadata
- implement direct `PerformanceMark` construction while preserving detached-timeline behavior and enforce illegal constructors for the remaining classes
- link performance entries, observers, and observer entry lists to their canonical prototypes and validate entry-list receivers/arguments across typed dispatch

## Validation

- all six parity fixtures listed in #8231 pass
- adjacent `global/class-constructors` and `observer/supported-entry-types-exact` parity fixtures pass
- `./scripts/test_affected_crates.sh --base origin/main` scoped tests completed (runtime 2546, Perry 987, codegen 1062, FFI 26; zero failures)

No version bump included.

## Summary by CodeRabbit

- **New Features**
  - Added Node-compatible `node:perf_hooks` class prototypes, inheritance, methods, accessors, and descriptors.
  - Added `PerformanceMark` construction with support for names and options.
  - Added `PerformanceObserver.supportedEntryTypes`.
  - Added performance entry list filtering by name and type.

- **Bug Fixes**
  - Added validation for invalid constructors and entry-list filter arguments.
  - Improved compatibility for performance object behavior and string tags.

### DIFF
--- a/changelog.d/8254-perf-hooks-class-prototypes.md
+++ b/changelog.d/8254-perf-hooks-class-prototypes.md
@@ -1,0 +1,15 @@
+### fix(perf_hooks): give public classes Node-compatible prototypes and constructors
+
+The public `node:perf_hooks` classes now expose their real prototype chains,
+property descriptors, accessors, methods, and `Symbol.toStringTag` values.
+Performance entries, observers, the global `performance` object, and observer
+entry lists are linked to the same canonical prototypes, so reflection,
+`instanceof`, extracted prototype calls, and receiver validation agree with
+Node.
+
+`new PerformanceMark(name, options)` now creates a detached mark with cloned
+detail data without adding it to the performance timeline. The other
+non-publicly-constructible perf classes throw `ERR_ILLEGAL_CONSTRUCTOR`, while
+`PerformanceObserver` retains its supported constructor. Entry-list filters
+also report Node-compatible missing-argument, Symbol-coercion, and invalid-this
+errors across both generic and typed dispatch paths. Fixes #8231.
--- a/crates/perry-codegen/src/lower_call/builtin.rs
+++ b/crates/perry-codegen/src/lower_call/builtin.rs
@@ -515,6 +515,35 @@ pub(super) fn lower_builtin_new<'a>(
                 &[(DOUBLE, &cb)],
             )))
         }
+        "PerformanceMark" => {
+            let (name, options) = adopt_two_leading_args_discard_rest(ctx, args, group)?;
+            ctx.pending_declares.push((
+                "js_perf_mark_constructor".to_string(),
+                DOUBLE,
+                vec![DOUBLE, DOUBLE],
+            ));
+            Ok(Some(ctx.block().call(
+                DOUBLE,
+                "js_perf_mark_constructor",
+                &[(DOUBLE, &name), (DOUBLE, &options)],
+            )))
+        }
+        "Performance"
+        | "PerformanceEntry"
+        | "PerformanceMeasure"
+        | "PerformanceObserverEntryList"
+        | "PerformanceResourceTiming" => {
+            for arg in args {
+                let _ = lower_expr(ctx, arg)?;
+            }
+            ctx.pending_declares
+                .push(("js_perf_illegal_constructor".to_string(), DOUBLE, vec![]));
+            Ok(Some(ctx.block().call(
+                DOUBLE,
+                "js_perf_illegal_constructor",
+                &[],
+            )))
+        }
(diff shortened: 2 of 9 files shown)


proggeramlug:
Merging. Two notes for the record rather than blockers.

**`perf_hooks.rs` lands at 1998 lines** — the 2000-line cap passes with two lines of headroom. The next addition to that file trips `check_file_size.sh`, so the follow-ups on #8234/#8237 will want a split first. `prototypes.rs` being a new 423-line module is the right instinct; more of it should go the same way.

**The shared-dispatch hook is correctly gated, and I checked because it isn't perf_hooks-local.** `native_call_method.rs` now runs a check on a path every native method call reaches. It is ordered `method_name_len == 16` first, then `is_perf_observer_list_value(object)`, then the name compare — so a non-perf_hooks receiver pays one integer comparison and nothing else, and 16-character method names are rare enough that the predicate almost never runs. Worth being deliberate about here: #8252 just recovered **−3.89%** on `pipeline` by deleting exactly this shape — a default-off check on a hot path — so a per-call addition to a dispatch path deserves the ordering it got.

Verification: `perry-runtime --lib` **2555/0/4**, `perry-codegen --lib` **1064/0** (it touches `lower_call/builtin.rs`), fmt, `check_file_size`, and the five script gates including the new `unrooted_local_shape` ratchet — all green. Merges into current `main` with zero conflicts.

Issue #8132, opened 2026-08-15
codegen: RS4GC expands a single bundled module factory 21x (413 MB IR, 1.5M gc.relocates) — blocks #8034 as a CI gate


proggeramlug:
## Summary

RS4GC expands one Perry-emitted function by **21x**, producing 1.5 million relocations. This is not a bug in the pass — it is doing exactly what the input demands. The input is the problem: Perry emits a bundled webpack module factory as a **single function** carrying thousands of safepoints and thousands of live GC pointers.

RS4GC's output scales as `safepoints x live-values`, so a monolith of that shape is inherently explosive.

## Measurement

Input: `next@16.3.0`'s bundled `jsonwebtoken` (`dist/compiled/jsonwebtoken/index.js`, sha256 `056c2ddd…a6b9`), compiled to an app-only dylib.

The offending function, `perry_closure_jsonwebtoken_index_js__227`:

| property | value |
|---|---|
| size | **4.93 MB** |
| basic blocks | 8,471 |
| lines | 101,601 |
| call sites (safepoints) | **8,044** |
| allocas | 1,015 |
| `addrspace(1)` GC pointers | **13,369** |

RS4GC on the containing unit (measured with stock `opt` at LLVM 22.1.4, `-mcpu=apple-m1`):

```
pre-RS4GC IR:   19 MB
post-RS4GC IR: 413 MB      <- 21x
statepoints:    21,774
gc.relocate:  1,502,654    <- mean 83 operands/statepoint, max 426
```

## Why it matters

Compile time on that single module, same input and same compiler, differing only in opt level applied to RS4GC's output:

| arm | result |
|---|---|
| `-O0` (`PERRY_LL_SIZE_OPT=0`) | **2m46.5s**, 40 MB dylib |
| `-Os` (current default for this unit) | **>100 minutes**, stopped without finishing |

This blocks #8040 structurally, not just cosmetically. That tracker's definition of done requires ten cold process starts, two verifier runs each, and a 100-iteration 20-way concurrent batch — none of which is reachable at hours-per-compile. The #8034 fixture cannot serve as a CI gate at this cost **regardless of correctness**.

It was previously masked: the compile crashed (#8121) before the slowness could be observed. With that fixed, this is what's left.

## Directions

[...]

Issue #8258, opened 2026-08-16
[test] test_next_app_route_dylib.sh is silently cwd-sensitive: run from outside the repo, the provider runtime loses -C force-unwind-tables and fails 25 minutes later naming the wrong cause


proggeramlug:
## What happens

`tests/test_next_app_route_dylib.sh` fails at the first cold start when invoked from a working directory outside the repository:

```
perry: FATAL: unwind tables are missing from this runtime build (0 frame(s) visible to the unwinder).
```

Re-running the identical command from the repo root passes cleanly. Observed on `695daf302b` (bench mini, macOS 26.5.1 arm64) during the #8040 acceptance run.

## Why

The gate builds its provider runtime dylib with `cargo build --manifest-path <worktree under $TMPDIR>/…`. **Cargo discovers `.cargo/config.toml` from the invoking working directory, not from the manifest path** — and that file is the sole supplier of

```toml
[build]
rustflags = ["-C", "force-unwind-tables=yes"]
```

which #7302's exception transport requires: `panic = "abort"` omits unwind tables by default, so without the flag every throw crossing a runtime helper frame is stranded. Run from outside the repo, the provider runtime is built without unwind tables and the runtime's own self-check fires on the first `try`.

## Why it is worth fixing rather than documenting

The failure **names the wrong culprit**. It reports a property of the runtime build ("unwind tables are missing") with no hint that the cause is the caller's cwd, and it appears at cold start 1 — i.e. after a ~25-minute compile — so the feedback loop is expensive. Anyone hitting it will reasonably suspect the runtime build, the profile, or `#7302` itself before suspecting `cd`.

CI is unaffected (it runs from the repo root), which is exactly why this can sit latent.

## Suggested fixes

- Have the script `cd` to the repo root itself, or pass `--config` / set `CARGO_BUILD_RUSTFLAGS` explicitly for the provider build so the flag cannot depend on cwd. The script already computes `repo_root`.
- Optionally assert early: if the built provider runtime lacks unwind tables, fail before serving with a message naming the cwd/rustflags cause rather than letting the runtime's self-check surface it 25 minutes later.

Same class as the other cwd-sensitivity in this fixture family fixed by #8209 (the host previously ran from `.next/server`, which broke Next's per-request `routes-manifest.json` open).

PR #8260, opened 2026-08-17
fix(gc): root box pointer allocas and unrooted call operands (#8258)

jdalton:
Work toward #8258 (the seeded rooting `TypeError: Cannot convert undefined or null to object`). Three codegen rooting improvements, applied and verified in IR — **but the seeded failure survives**, so this is a partial fix and the PR is opened as progress + diagnosis, not a resolution.

## The fault site (diagnosed)

The throw is from `Object.getPrototypeOf` in `crates/perry-runtime/src/object/object_ops/prototype.rs:161`, called with value bits `0x7ffc000000000001` (`TAG_UNDEFINED | 1` — the box sentinel from `js_box_alloc_bits`). The value originates from a box capture slot read via `js_closure_get_capture_bits(closure, 1)` inside the node-machine-id webpack bundle (`perry_closure_node_modules_node_machine_id_dist_index_js__283`). The box retains its initial sentinel — the captured variable's value was never written to the moved box.

The safepoint fires in `src_lib_util_graceful_exit_ts__init_body` (the first imported module's for-of over SIGNALS) with `total_slots=0` in the scanner profile — zero GC stack slots available. The GC moves ~6,548 objects.

## The three fixes (verified in emitted IR)

1. **Box pointer alloca shadow-slot bind** (`stmt/mod.rs`) — `PreallocateBoxes` created box pointer slots as `alloca i64`, invisible to the RS4GC pass. Added `emit_shadow_slot_bind_ptr` so the precise-roots pass retypes them to `ptr addrspace(1)`.
2. **Closure capture rooting across `js_closure_alloc`** (`closure.rs`) — `captured_value_bits` were loaded before the allocation and used after it without rooting. Added `RootedGroup` with `adopt_emitted` and re-reads on the non-singleton path.
3. **`ProcessOn`/`ProcessOnce`/`ObjectDefineProperty` operand rooting** (`misc_methods.rs`) — sequential `lower_expr` calls without `with_operands_rooted`; fixed with `open_rooted_group`.

## Results

- Seeds 8, 11, 22: FAIL before, FAIL after — **not resolved**.
- Seeds 1, 5, 15, 20: PASS before, PASS after — no regression.
- `cargo test -p perry-runtime`: all pass. `-p perry-codegen`: same failure set as parent.

## What remains

The alloca retyping is confirmed in IR but the crash persists, and the box still holds the sentinel — so the captured variable's value never reaches the moved box. This points deeper than codegen alloca tracking: either the assignment writes to a stale box pointer from a different frame, or `scan_box_roots_mut` (`crates/perry-runtime/src/box.rs:212`) / the remember-set has a gap for closures created in the node-machine-id webpack bundle. `crates/perry-runtime/src/closure/alloc.rs:323` (`js_closure_alloc_with_captures_singleton`) is the runtime-side capture rooting to inspect next.

Opened as progress + a complete diagnosis for the next pass, not a claim of resolution.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runtime stability when closures capture values during memory cleanup.
  * Fixed potential issues affecting event processing and property updates under memory pressure.
  * Improved reliability for boxed values when memory is relocated by automatic cleanup.
  * Reduced the risk of intermittent crashes or incorrect behavior during garbage collection.

### DIFF
--- a/crates/perry-codegen/src/expr/closure.rs
+++ b/crates/perry-codegen/src/expr/closure.rs
@@ -7,6 +7,8 @@
 use anyhow::Result;
 use perry_hir::Expr;
 
+use crate::rooting;
+use crate::rooting::Repr;
 use crate::type_analysis::compute_auto_captures;
 use crate::types::{DOUBLE, I32, I64, PTR};
 
@@ -120,6 +122,32 @@ pub(crate) fn lower(ctx: &mut FnCtx<'_>, expr: &Expr) -> Result<String> {
                     captured_value_bits.push(v_bits);
                 }
             }
+
+            // #8258: Root captured values across js_closure_alloc, which can
+            // trigger a copying minor. Box pointers (raw i64) and NaN-boxed
+            // values held in bare SSA registers are invisible to the collector;
+            // without rooting, a GC during the allocation moves the underlying
+            // objects, the registers hold stale addresses, and
+            // js_closure_set_capture_bits stores stale pointers into the
+            // closure's capture slots. Later, js_box_get_bits on a stale box
+            // pointer reads garbage, surfacing as "TypeError: Cannot convert
+            // undefined or null to object" when the closure body calls
+            // js_require_object_coercible on the captured value.
+            //
+            // The captured-singleton path (js_closure_alloc_with_captures_singleton)
+            // roots captures inside the runtime helper, so this group is only
+            // re-read on the non-singleton path below.
+            let mut capture_handles: Vec<rooting::EmittedValue> = Vec::new();
+            let capture_group = if !captured_value_bits.is_empty() {
+                let mut g = rooting::open_rooted_group(captured_value_bits.len());
+                for v_bits in &captured_value_bits {
+                    let h = g.adopt_emitted(ctx, Repr::Ptr, v_bits, true);
+                    capture_handles.push(h);
+                }
+                Some(g)
+            } else {
+                None
+            };
 
             // Compute the closure function name BEFORE taking the
             // mutable block borrow.
@@ -350,8 +378,25 @@ pub(crate) fn lower(ctx: &mut FnCtx<'_>, expr: &Expr) -> Result<String> {
             // (so the cached layout matches a fresh allocation). The
             // other paths still need explicit per-slot writes.
             if !captured_singleton {
+                // #8258: Re-read captured values after js_closure_alloc
+                // (which may have triggered a copying minor that moved the
+                // boxed objects). The captured-singleton path roots captures
+                // inside the runtime helper; this path does not, so the
+                // re-read is mandatory.
+                let re_read_bits: Vec<String> = if let Some(ref g) = capture_group {
+                    capture_handles.iter()
+                        .map(|h| g.reread_emitted(ctx, *h))
+                        .collect()
+                } else {
+                    Vec::new()
+                };
+                let bits_to_set: &[String] = if re_read_bits.is_empty() {
+                    &captured_value_bits
+                } else {
+                    &re_read_bits
+                };
                 let blk = ctx.block();
-                for (idx, val_bits) in captured_value_bits.iter().enumerate() {
+                for (idx, val_bits) in bits_to_set.iter().enumerate() {
                     let idx_str = idx.to_string();
                     blk.call_void(
                         "js_closure_set_capture_bits",
@@ -417,6 +462,13 @@ pub(crate) fn lower(ctx: &mut FnCtx<'_>, expr: &Expr) -> Result<String> {
                         (I64, &new_target_bits),
                     ],
                 );
+            }
+            // #8258: Release the capture root group after all capture slots
+            // are written. The `this` and `new.target` captures are loaded
+            // from the frame's this/new_target stack or IMPLICIT_THIS, which
+            // are runtime singletons not at risk from a copying minor.
+            if let Some(g) = capture_group {
+                g.release(ctx);
             }
             Ok(nanbox_pointer_inline(ctx.block(), &closure_handle))
         }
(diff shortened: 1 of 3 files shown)


proggeramlug:
Thanks for the diagnosis — the fault-site work here is genuinely good, and hunks 2 and 3 look right to me. **I'm not merging as-is, because hunk 1 rests on a premise the tree contradicts, and it reverses a measured optimization.**

## Hunk 1: boxes cannot move, so this roots nothing

The new comment in `stmt/mod.rs` says:

> a GC that moves the box leaves the slot holding a stale from-space pointer

Boxes are not in the GC heap and are never moved. Three independent places in this repo say so:

- `crates/perry-runtime/src/box.rs:40` — *"boxes come from `std::alloc::alloc` directly, not the GC arena"*.
- `scripts/gc_root_dominance_check.py:2713` classifies `js_box_alloc*` as an **`IMMOVABLE_SOURCE`**, label *"js_box_alloc* (outside the GC heap)"*, and line 2499 repeats it: *"`js_box_alloc_bits`, which is `std::alloc::alloc` — not in the GC heap at all."* This is machine-checked, so a change that assumes the opposite is contradicting a checked invariant.
- `BOX_REGISTRY` is monotonic — boxes are never freed either. Their *contents* are traced via `scan_box_roots_mut`; the pointer itself needs no rooting.

So the shadow slot this adds roots nothing, and the stale-from-space-box failure mode it describes cannot occur.

## Why that is actively harmful, not merely redundant

This is the exact thing **#8143 removed** to fix **#8132** (the RS4GC 21× blowup on the jsonwebtoken webpack monolith). The finding there was that the explosion is *liveness*-shaped, not safepoint-shaped: ~300 values live at ~90% of statepoints, of which **63 were `js_box_alloc_bits` slots**. Unbinding boxed slots cut relocations by **65%**.

`emit_preallocate_boxes` runs per preallocated box, and this calls `reserve_shadow_slot()` for every one — putting precisely that population back into the live set at every statepoint, on precisely the bundle-shaped workloads #8132 was about. Note it also bypasses the existing gate: boxed locals get no `shadow_slot_map` entry, so `emit_shadow_slot_bind_for_local` returns early for them by construction; calling the raw `emit_shadow_slot_bind_ptr` sidesteps that.

## This also redirects your hunt

Your diagnosis says the box "retains its initial sentinel — the captured variable's value was never written to **the moved box**." Since boxes don't move, that framing can't be the explanation, which is useful: it rules out the whole from-space direction and means the sentinel has to come from somewhere else — a write reaching a *different* box instance, a capture-index mismatch, or the value never being stored at all. Your own listed next step (`js_closure_alloc_with_captures_singleton` in `closure/alloc.rs:323`) is consistent with that and is where I'd look. `scan_box_roots_mut` is about box *contents*, which is a separate question from the pointer.

## What I would merge

**Hunks 2 and 3, split out.** Both are straightforward applications of the repo's own idioms and stand independently of the box question:

- `misc_methods.rs` — `o` was held live across `lower_expr(key)` and `lower_expr(value)`, either of which can collect. `with_operands_rooted` is the right fix.
- `closure.rs` — rooting `captured_value_bits` across `js_closure_alloc` with `open_rooted_group`/`adopt_emitted` plus re-reads looks correct.

Two housekeeping notes for the resubmit: please add a `changelog.d/<PR>-<slug>.md` fragment (contributors write their own; they become the release notes), and the title/body reference **#8258**, which is the `test_next_app_route_dylib.sh` cwd-sensitivity issue — unrelated to this work. Worth pointing at the right issue so it doesn't get cross-filed.

Happy to re-review and land the split promptly.

proggeramlug:
Landed the sound part of this as **#8267**, which is merged. Closing here since the branch's other two hunks are not being taken.

**Merged from your work:** the `misc_methods.rs` operand rooting. `ProcessOn`, `ProcessOnce` and `ObjectDefineProperty` held the first operand live across the lowering of the second and third, either of which can collect. `with_operands_rooted` is exactly right, and you're credited in the commit and the changelog fragment.

**Not taken, with the evidence:**

*Box pointer alloca shadow-slot bind.* The premise in the comment — "a GC that moves the box" — doesn't hold. Box pointers are `std::alloc`'d and outside the GC heap (`box.rs:40`), `gc_root_dominance_check.py:2713` classifies `js_box_alloc*` as an `IMMOVABLE_SOURCE`, and `BOX_REGISTRY` is monotonic so they're never freed either. So the slot roots nothing, and because `emit_preallocate_boxes` runs per box it would restore the exact live-set population #8143 removed to fix #8132's 21× RS4GC blowup — 63 of the ~300 values live at ~90% of statepoints were `js_box_alloc_bits` slots, and unbinding them cut relocations 65%.

*Closure capture rooting.* This one I A/B'd on the single file, because "no regression" deserved checking against the integration suites rather than the lib tests:

| `expr/closure.rs` | `shadow_slot_hygiene::closure_body_write_to_captured_outer_local_is_visible_to_shadow_analysis` |
|---|---|
| at `main` | **ok** |
| with the hunk | **FAILED** — "captured Any local written to a pointer inside a closure must keep its outer slot" |

With the hunk, `cargo test -p perry-codegen --no-fail-fast` gives 10 failures; without it, 9, all of which are in the known baseline. It removes a root rather than adding one. Worth knowing that suite doesn't run per-PR, so the lib tests passing is not the same as the codegen suites passing.

**Where I'd point the hunt next.** Your diagnosis says the box "retains its initial sentinel — the captured variable's value was never written to *the moved box*." Since boxes never move, that framing can be set aside entirely, which is progress: the sentinel has to come from a write reaching a *different* box instance, a capture-index mismatch, or no store at all. Your own listed next step — `js_closure_alloc_with_captures_singleton` in `closure/alloc.rs:323` — fits that, and `scan_box_roots_mut` concerns box *contents*, a separate question from the pointer.

The fault-site work here was genuinely useful — the value bits, the originating capture slot, the safepoint with `total_slots=0`. Thanks for it, and please do open the follow-up.

PR #8262, opened 2026-08-17
fix(gtk4): initialize GTK before creating Canvas

### LINKED ISSUE #7995: GTK Not Initialized
## What happened

Received the following error when trying to run the binary:
```
GTK has not been initialized. Call `gtk::init` first.
```

## What you expected
Binary run with a simple canvas.

## Minimal reproduction
```js
import { App, Canvas } from 'perry/ui'

const canvas = Canvas(400, 300)
canvas.setFillColor(1.0, 0.4, 0.0, 1.0)
canvas.fillRect(10, 10, 100, 80)

App({
    title: 'Perry Test',
    width: 800,
    height: 600,
    body: canvas
});
```

Command you ran:

```bash
perry run ./src/main.js
```

## Environment

- **Perry version**:  0.5.1494
- **Host OS**:  Zorin OS 18.1
- **Target**: native
- **Rust toolchain**: rustc 1.97.1 (8bab26f4f 2026-07-14)
- **Installed via**: source

### Perry Doctor
Environment Checks
──────────────────
  ⚠ runtime library: not found - run: cargo build --release -p perry-runtime

### Console Output Running Binary
```
thread '<unnamed>' (16371) panicked at /home/username/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/gtk4-0.9.7/src/auto/drawing_area.rs:31:9:
GTK has not been initialized. Call `gtk::init` first.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Aborted
```

### Backtrace:
```
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: <gtk4::auto::drawing_area::DrawingArea>::new
   3: perry_ui_gtk4::widgets::canvas::create
   4: main
   5: <unknown>
   6: __libc_start_main
   7: _start
```

### Full Backtrace
```
stack backtrace:
   0:     0x56f350d1e099 - <<std[1e3c4ec04c5261a9]::sys::backtrace::BacktraceLock>::print::DisplayBacktrace as core[37f591cfbe66b0b1]::fmt::Display>::fmt
   1:     0x56f350d4c9ea - core[37f591cfbe66b0b1]::fmt::write
   2:     0x56f350d268e6 - <unknown>
   3:     0x56f350cf8af2 - <unknown>
   4:     0x56f350d13fae - std[1e3c4ec04c5261a9]::panicking::default_hook
   5:     0x56f350d141a2 - std[1e3c4ec04c5261a9]::panicking::panic_with_hook
   6:     0x56f350cf8be4 - <unknown>
   7:     0x56f350ced699 - std[1e3c4ec04c5261a9]::sys::backtrace::__rust_end_short_backtrace::<std[1e3c4ec04c5261a9]::panicking::panic_handler::{closure#0}, !>
   8:     0x56f350cfa27d - __rustc[6ef78fba13ac776c]::rust_begin_unwind
   9:     0x56f35056badc - core[37f591cfbe66b0b1]::panicking::panic_fmt
  10:     0x56f35196c6cd - <gtk4[e2b69aebd46bb70]::auto::drawing_area::DrawingArea>::new
  11:     0x56f351be2fe0 - perry_ui_gtk4[19c84d6e5a2a796e]::widgets::canvas::create
  12:     0x56f3505f43cc - main
  13:     0x750d6662a1ca - <unknown>
  14:     0x750d6662a28b - __libc_start_main
  15:     0x56f3505f3d65 - _start
  16:                0x0 - <unknown>
```
## Anything else
cargo build --release -p perry-runtime has already been completed yet the issue persists on doctor. Additionally, I am able to compile/run other binaries such as the Text Editor example.

proggeramlug:
## Summary
- initialize GTK before constructing a GTK4 `Canvas` drawing area
- add a source-invariant regression test for initialization ordering

Fixes #7995

## Testing
- `cargo fmt -p perry-ui-gtk4 -- --check`
- `rustfmt --edition 2021 --check crates/perry-ui-gtk4/src/widgets/canvas.rs`
- `cargo check -p perry-ui-gtk4` (Windows cfg smoke check)
- `git diff --check`

The Linux-only GTK build and unit test will run in the scoped PR CI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Canvas creation now initializes GTK before constructing its drawing area.
  * GTK4 applications can create a canvas before calling `App()`, avoiding initialization failures.
* **Tests**
  * Added regression coverage to verify the required initialization order and prevent the issue from recurring.
* **Documentation**
  * Added a changelog entry describing the canvas initialization fix.

### DIFF
--- a/changelog.d/8262-gtk-canvas-init.md
+++ b/changelog.d/8262-gtk-canvas-init.md
@@ -1,6 +1,11 @@
 ### fix(gtk4): initialize GTK before creating Canvas
 
-`Canvas()` now initializes GTK before constructing its `DrawingArea`, matching
-the other GTK widget constructors. GTK4 apps can therefore create a canvas
-before calling `App()` without aborting with “GTK has not been initialized.”
-Fixes #7995.
+Perry applications commonly construct their widget tree before calling
+`App()`. Unlike the other GTK widget constructors,
+`crates/perry-ui-gtk4/src/widgets/canvas.rs` previously called
+`DrawingArea::new()` without first invoking the shared GTK initializer, so that
+normal ordering aborted with “GTK has not been initialized.”
+
+`Canvas()` now calls `ensure_gtk_init()` before `DrawingArea::new()`. A
+source-order regression test covers the fixed ordering and proves that the
+pre-fix sequence is rejected. Fixes #7995.
--- a/changelog.d/8262-gtk-canvas-init.md
+++ b/changelog.d/8262-gtk-canvas-init.md
@@ -1,0 +1,6 @@
+### fix(gtk4): initialize GTK before creating Canvas
+
+`Canvas()` now initializes GTK before constructing its `DrawingArea`, matching
+the other GTK widget constructors. GTK4 apps can therefore create a canvas
+before calling `App()` without aborting with “GTK has not been initialized.”
+Fixes #7995.
--- a/crates/perry-ui-gtk4/src/widgets/canvas.rs
+++ b/crates/perry-ui-gtk4/src/widgets/canvas.rs
@@ -135,6 +135,7 @@ fn image_handle_from_arg(image: i64) -> i64 {
 
 /// Create a Canvas widget with given dimensions.
 pub fn create(width: f64, height: f64) -> i64 {
+    crate::app::ensure_gtk_init();
     let area = DrawingArea::new();
     area.set_content_width(width as i32);
     area.set_content_height(height as i32);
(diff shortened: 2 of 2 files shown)


proggeramlug:
Merging. The fix is right and the test is sound — one coverage caveat worth recording.

**I could not compile-check this here.** `perry-ui-gtk4/src/lib.rs:1` is `#![cfg(target_os = "linux")]`, so on macOS the crate compiles to nothing: `cargo test -p perry-ui-gtk4` reports `0 passed; 0 failed`, which looks green and proves nothing. Flagging that explicitly because a green run on the wrong host is not evidence.

**The test logic I did verify**, by replicating it against the real file. `init_precedes_drawing_area` returns `true` for the current `canvas.rs` (init at offset 52, `DrawingArea::new()` at 98) and `false` for the pre-fix ordering fixture, so the negative case genuinely discriminates rather than passing vacuously. Including that negative assertion is the right instinct — a source-ordering check with no counter-case can silently stop matching and still report success. (One subtlety in its favour: the fixture relies on Rust's `\` line-continuation stripping leading whitespace, so the `\n/// Clear all drawing commands.` terminator is found. My first reading of it missed that and I briefly thought the `.expect()` would panic — it doesn't.)

**`ensure_gtk_init` exists** at `app.rs:16` and is `pub(crate)`, so the call from `widgets/canvas.rs` is in-crate and visible.

**The coverage gap:** nothing in CI ever runs this test. `perry-ui-gtk4` appears in the workflows only under `release-packages.yml`, and only as `cargo build --profile dist -p perry-ui-gtk4` — a build, not a test. So the ordering invariant is documented rather than enforced, and a future edit that moves the init call back below the constructor will not go red anywhere. Worth adding `cargo test -p perry-ui-gtk4` to whichever Linux job is cheapest; that would also mean a compile error here surfaces on a PR instead of at release time.

PR #8244, opened 2026-08-16
ci: shard parity 8 ways + aggregate fan-in — the unsharded job hit the 6h cap

proggeramlug:
## Why

Follow-up to #8187 — this commit was pushed 11 minutes after that PR was squash-merged, so it never landed (branch state at merge time was the `603699ee5` head; verified by content on `main`, per the #7770 lesson).

The first full-tier run (31935729773) proved the problem: **the unsharded `parity` job was killed by GitHub's 6-hour job cap** (11:44→17:45, `cancelled`). The release gate (`full-suite-gate`, what `release-packages.yml`'s `await-tests` waits for) structurally cannot complete without this.

## What

* `parity` becomes a plan-driven **8-shard matrix** (`run_parity_tests.sh --shard N/8`, the same round-robin partition the gap suite uses). Each shard also runs `parity_known_failures.py` — shard-safe by design ("an entry whose test is not in this shard is never flagged") — and `npm ci` for the #8198 oracle fixtures.
* New **`parity-aggregate`** fan-in merges the shard reports via `scripts/parity_report_merge.py` (self-tested; **refuses a missing shard** with `--expect`, so a lost artifact is a red run, not a smaller green suite; reproduces the harness's own percentage formula) and runs the two gates whose baselines only make sense on the whole suite: the threshold minimums (a 62%-floor category with two tests in a shard would flap) and the per-module matrix trend.
* `gate` fan-in requires `parity-aggregate`; `ci_plan.py` emits `parity.total/shards`.
* Drops the stale "informational / does not block" comment block on the job — since the tiering, a red parity fails `full-suite-gate`.

## Validation

A `tier=full` dispatch with exactly this tree ran on the old branch: run [31964093732](https://github.com/PerryTS/perry/actions/runs/31964093732) — see the comment below for shard/aggregate results when it completes.

## Summary by CodeRabbit

* **New Features**
  * Parity testing now runs across eight parallel shards, reducing execution time.
  * Shard reports are automatically merged and validated for completeness, consistency, and duplicate results.
  * Aggregate checks evaluate overall thresholds and module trends before release approval.

* **Bug Fixes**
  * Missing or incompatible shard reports now prevent the release gate from passing.

* **Documentation**
  * Updated CI and release documentation to describe sharded parity testing and aggregation.

PR #8265, opened 2026-08-17
fix(ext-net): give the pump's in-flight dispatch GC custody (#8259)

### LINKED ISSUE #8259: gc-stress: test_gap_gc_net_once_flags_rekey (added by #8216, never executed in CI) SIGSEGVs deterministically under force_verify — reds pr-gate on every rebased PR
`test_gap_gc_net_once_flags_rekey` — the witness #8216 added (18:15, with its rooting fixes and the `gc_repsel_corpus.txt` registration) — **SIGSEGVs deterministically under the `force_verify` arm** of the GC × repsel matrix:

```
FAIL test_gap_gc_net_once_flags_rekey (exit=139 cycles=3 evacuated=0 scavenged=9895 reclaimed=13225440)
matrix row: gc_net_once_flags_rekey   UNVER UNVER PASS UNVER FAIL UNVER PASS   (FAIL = force_verify)
```

Two runs, byte-identical counters (`cycles=3 scavenged=9895 reclaimed=13225440`) — perfectly reproducible, which per the rooting-bug playbook points at an **unregistered/unpruned table holding a raw heap pointer**, not a racy unrooted register.

**This witness has never completed a CI run.** #8216's only `Tests` run had `gc-stress` *cancelled* (run 31957836547 — merged during yesterday's queue saturation), so the test landed with zero executions. It first ran on PR #8244 (whose diff is workflow/scripts-only and cannot affect GC), failing identically on attempt 1 and 2: runs [31964669242](https://github.com/PerryTS/perry/actions/runs/31964669242) attempts 1+2.

Suspects: the #8216 fix itself being incomplete under forced evacuation (`force_verify` = `PERRY_GC_FORCE_EVACUATE` + `PERRY_GC_VERIFY_EVACUATION`), or interaction with the same-day #8174-family rekeyed-table pruning. `evac_minor` PASSES, so it is specifically the verify/forced combination.

Until fixed, this cell fails `gc-stress` in every tier — i.e. `pr-gate` on every rebased PR. If the fix isn't quick, consider an XFAIL triage for the cell (with this issue number) rather than softening the job.

proggeramlug:
Fixes #8259.

## Root cause

Every arm of `js_ext_net_drain_pending` snapshotted its listeners into a bare `Vec<i64>` (`listeners_for`) and, for some arms, allocated a payload (`alloc_buffer` / `alloc_string` / `build_error_object`) before the call loop. Both steps can collect — the allocation directly, every `call*` by running arbitrary user JS — and the evacuating arms then **move** the closures: `scan_net_roots` rewrites the canonical `statics::listeners()` slots, but a bare local snapshot is invisible to it, so callback N (and the already-built payload) were dereferenced at their old addresses once callback N−1's JS forced a collection. The #8259 witness's `once('data')` handler churns 30k objects between the two `'data'` callbacks — deterministic by construction.

`ServerListening`/`ServerClose` were worse: they **remove** their callbacks from the table before firing (one-shot semantics), so during dispatch nothing rooted them at all — a full sweep could free, not just move, them.

Not a regression from #8216 — its scanner fix is correct; its witness (which had never completed a CI run: the PR's own gc-stress was cancelled in the queue saturation) was the first thing to ever execute this path under forced evacuation.

## Fix

The same custody pattern #8216 introduced for ext-http's H2 pump: park the snapshot (+ at most one NaN-boxed payload per frame) in scanned thread-locals (`dispatch_custody.rs`), re-read each slot immediately before use so the copying GC's rewrite is observed, pop on drop (RAII, nests for re-entrant pumps). All **ten** dispatch loops converted. `scan_net_roots` visits the custody stacks; the root-holder census certifies the new holders through that call path.

## Validation

- **A/B under the matrix's exact `force_verify` env** (`PERRY_GC_HEAP_LIMIT=8 PERRY_GC_INCREMENTAL=0 PERRY_CONSERVATIVE_STACK_SCAN=off PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1`), same freshly built release compiler, auto-opt archive cache wiped between arms:
  - pre-fix: **rc=1, `TypeError: value is not a function`** (macOS spelling of the Linux SIGSEGV — same stale-closure deref), stdout empty
  - post-fix: **rc=0, `once fired: 1 of 2 events`** (byte-correct vs node)
  - copying minors live in both arms (6 copy-minor records under the matrix env; not a vacuous pass)
- 3 new unit tests on the custody frame (nesting, rewrite-visibility — the property the parking exists for — and empty frames); `cargo test -p perry-ext-net` otherwise unchanged (the one pre-existing macOS-host relocation-expectation failure fails identically pre/post-fix and passes on CI Linux)
- `gc_runtime_root_holders.py` and `check_thread_locals.py` both green with no inventory edits — the census reaches the new holders via the registered scanner
- `cargo fmt --check` clean

Once merged, `gc-stress` (which reds `pr-gate` on every rebased PR via this cell today) goes back to green; PR #8244's failure was this.


proggeramlug:
Merging. This is the most carefully evidenced GC PR I've reviewed today — a few notes on what I verified independently rather than took on trust.

**The "not a vacuous pass" line is the part that matters**, and it is the check most GC PRs skip. Six copy-minor records under the matrix env means the fix was actually exercised, not merely un-crashed; a green arm that ran zero copying minors would have proven nothing. Same for the pre-fix arm reproducing `rc=1, TypeError: value is not a function` — a fix without a demonstrated red baseline is unfalsifiable.

**The one failing test is genuinely pre-existing.** I did not take that on trust: reverting `lib.rs` to `origin/main` and deleting `dispatch_custody.rs`, `tests::gc_mutable_scanner_rewrites_listener_roots` **still fails** on this macOS host. So it is the host relocation-expectation failure you described, not something this PR introduced.

**Verified here:** `perry-ext-net` 27 passed / 1 failed (that one), all three new custody tests present and passing (`parked_slots_are_the_read_path`, `frames_nest_and_pop`, `empty_frame_is_cheap_and_safe`), and `gc_runtime_root_holders`, `check_thread_locals`, `addr_class_inventory`, `check_file_size` and `cargo fmt` all clean. `parked_slots_are_the_read_path` is the right test to have written — it pins the rewrite-visibility property the parking exists for, rather than just that the frame nests.

**One fix pushed:** the changelog fragment was named `8260-net-dispatch-custody.md`. Fragment filenames are PR-keyed precisely so in-flight PRs never collide, and #8260 is a different, unrelated PR. Renamed to `8265-`.

**One note on my side, not yours.** #8253 added `scripts/unrooted_local_shape.py`, which is meant to catch exactly this class in `perry-ext-*`. It reports **218 findings both before and after this PR** — it does not see the bug you fixed. Its detector matches `let x = js_*_alloc(...)` bindings, so a `Vec<i64>` snapshot of listener pointers is invisible to it. That is a real blind spot in my instrument, and this PR is the counter-example that proves it; I'll extend it to container-of-pointer snapshots.

Issue #6667, opened 2026-07-19
Native-module namespaces don't enumerate exports — turbopack/Babel wildcard interop yields empty namespace (breaks Auth.js login via jose HKDF)


proggeramlug:
## Summary

Perry's native-module namespace objects (`NATIVE_MODULE_CLASS_ID`) do not enumerate their exports:

```js
const crypto = require("crypto");
Object.keys(crypto).length; // node: 70   perry: 1
let n = 0; for (const k in crypto) n++;  // node: 70   perry: 0
({ ...crypto }).createHmac;  // node: [Function]   perry: undefined
```

Every interop layer that builds a namespace by enumeration therefore produces an **empty namespace**: turbopack's ESM-import interop over an externalized builtin (`e.i(...)` on `a.x("crypto", () => require("crypto"))`), Babel's `interopRequireWildcard`, and plain object spread.

**Production impact (how this was found):** Auth.js credentials login on a Next.js 16 App Router standalone fails with `CallbackRouteError: TypeError: value is not a function` on the SUCCESS path only. jose's compiled HKDF does `Z = e.i(<crypto>)`, probes `typeof Z.hkdf` (undefined → picks its pure-JS fallback), and the fallback calls `(0, Z.createHmac)(...)` — also undefined through the empty interop namespace → the session-token derivation throws, so every successful password check still 302s to `error=Configuration`. Failed logins work (never reach HKDF), which masked this until a real credentials round-trip.

`Z.default.randomBytes` style access (interop `default` member) works — only named enumeration/copy is missing.

## Suggested direction

Give enumeration surfaces (`Object.keys`/`getOwnPropertyNames`, for-in, object spread/`Object.assign` copy) a native-module arm that synthesizes the export-name list per module and, for copies, materializes `bound_native_callable_export_value(module, name)` per key. A partial inventory already exists as the `(module, method) → arity` tables in `object/native_module/callable_exports.rs`; a fuller per-module export inventory (constants/objects included) would close the node gap (`crypto` = 70 keys in Node 22).

Repro: the snippet above; observed on main @ 5eb2bc495, macOS arm64.

PR #8269, opened 2026-08-17
fix(runtime): enumerate native module namespace values

### LINKED ISSUE #8235: [parity] Object.entries/values on a native-module namespace return the internal __module__ sentinel, not the module surface
Split out of #6766 after PR #8229. **Not a `perf_hooks` bug** — it affects every native-module namespace object; `perf_hooks.constants` is just where the node-suite pins it.

`Object.keys(ns)` on a native-module namespace object is special-cased to return the module's export surface (`vt_own_keys_array` → `module_keys.rs`). `Object.entries(ns)` and `Object.values(ns)` are not: they fall through to the raw field bag, which holds only the internal `__module__` sentinel.

```js
import { constants } from "node:perf_hooks";
Object.keys(constants)                       // ✅ the 12 NODE_PERFORMANCE_GC_* names
Object.entries(constants)                    // ❌
// node:  [["NODE_PERFORMANCE_GC_FLAGS_ALL_AVAILABLE_GARBAGE", 16], … 12 pairs]
// perry: [["__module__", undefined]]
```

Pinned by `test-parity/node-suite/perf_hooks/constants/all-gc-values` (line 1, `Object.keys`, passes; line 2, `Object.entries`, does not). Reproduce with `./run_parity_tests.sh --suite node-suite --module perf_hooks`.

Two consequences beyond the wrong key list: the internal `__module__` sentinel is *observable* to user code through `entries`/`values`, and the values are `undefined` rather than the real constants.

The fix is to route `Object.entries`/`Object.values` on a `NATIVE_MODULE_CLASS_ID` receiver through the same pair of hooks `Object.keys` already uses — the `module_keys` list for names, and `get_native_module_constant` / the callable-export path for each value. `crates/perry-runtime/src/object/native_module.rs::vt_own_keys_array` is the existing precedent.

Worth checking whether `JSON.stringify(ns)`, spread (`{...ns}`) and `for…in` have the same hole.

proggeramlug:
## Summary

Route `Object.entries` and `Object.values` on native-module namespace objects through the same virtual export surface used by `Object.keys`, so user code sees real module keys and values instead of the internal `__module__` sentinel.

## Changes

- enumerate namespace keys through the armed native-module vtable
- resolve each value through the authoritative namespace property getter
- root receivers, keys, output arrays, and allocated callable exports across moving GC
- extend the existing `perf_hooks.constants` parity fixture with `Object.values`

## Related issue

Fixes #8235

## Test plan

- [x] `cargo build --release -p perry-runtime`
- [x] `env RUST_TEST_THREADS=1 cargo test --lib -p perry-runtime` (2557 passed, 0 failed, 4 ignored on rebased head)
- [x] `PERRY_SKIP_BUILD=1 ./run_parity_tests.sh --suite node-suite --module perf_hooks --filter all-gc-values` (1/1 pass)

## Additional surface check

`for...in` already uses the virtual `Object.keys` path, and spread / `Object.assign` already use the export-copy hook from #6667. `JSON.stringify` still has a separate raw-field serializer path and reproduces the sentinel leak; that broader compact/pretty/replacer serializer work is intentionally left as a follow-up rather than folded into this entries/values fix.

## Checklist

- [x] No workspace version, `CLAUDE.md`, or `CHANGELOG.md` bump
- [x] Conventional commit prefix used

### DIFF
--- a/changelog.d/8269-native-module-object-enumeration.md
+++ b/changelog.d/8269-native-module-object-enumeration.md
@@ -1,0 +1,8 @@
+### fix(runtime): enumerate native-module namespace values
+
+`Object.entries` and `Object.values` now expose a native module's public export
+surface instead of its internal `__module__` storage sentinel. Enumeration
+uses the same virtual key list as `Object.keys` and resolves each live value
+through the namespace property getter, including constants and callable
+exports. The receiver and intermediate arrays stay rooted across allocations
+that can trigger a moving collection. Fixes #8235.
--- a/crates/perry-runtime/src/object/field_get_set/enumeration.rs
+++ b/crates/perry-runtime/src/object/field_get_set/enumeration.rs
@@ -52,6 +52,69 @@ fn map_set_exotic_enum(stripped: *const ObjectHeader, what: MapSetEnum) -> *mut 
     out
 }
 
+/// Build `Object.values` / `Object.entries` from the virtual export surface of
+/// a native-module namespace.  Such namespaces physically contain only the
+/// internal `__module__` sentinel; their observable keys and values are
+/// supplied lazily by the native-module vtable.
+///
+/// Keep this in the generic enumeration module and call through the armed
+/// vtable so binaries that never create a namespace do not retain the native
+/// module export tables.  The receiver, key list, output, and per-key values
+/// are rooted because resolving a callable export can allocate and trigger a
+/// moving collection.
+unsafe fn native_module_enum(
+    obj: *const ObjectHeader,
+    what: MapSetEnum,
+) -> Option<*mut ArrayHeader> {
+    let vt = super::super::native_module::native_module_vtable()?;
+    let scope = crate::gc::RuntimeHandleScope::new();
+    let obj_h = scope.root_raw_const_ptr(obj);
+    let keys = obj_h.with_const_ptr(|obj| (vt.own_keys_array)(obj))?;
+    let keys_h = scope.root_raw_const_ptr(keys);
+    let count = crate::array::js_array_length(keys_h.get_raw_const_ptr::<ArrayHeader>());
+    let result = crate::array::js_array_alloc(count);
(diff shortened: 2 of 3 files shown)


proggeramlug:
Merging. Runtime suite **2567 passed / 0 failed**, all six script gates and `cargo fmt` clean, and the new node-suite fixture runs cleanly under the pinned oracle (26.5.1) producing the three expected lines.

**The rooting reads correctly**, which matters because resolving a callable export can allocate: receiver, key list, output array and per-key values are each rooted, with a per-iteration scope and re-reads through `get_raw_*_ptr` / `with_const_ptr` after every collection point.

One detail I checked rather than assumed: `js_array_push_f64` returns a possibly-reallocated pointer, and elsewhere in the tree the idiom captures it and writes it back (`arr_handle.set_raw_mut_ptr(arr)`). Here the return is discarded. That is safe **because of the pre-sizing** — `js_array_alloc(count)` sets `capacity = max(count, MIN_ARRAY_CAPACITY)` with `length = 0`, and the loop pushes exactly `count` items, so the array can never grow. Worth a comment if this code is ever changed to push a variable number of entries, since the safety is a property of the sizing rather than of the call.

**One suggestion, not a blocker: bump the node-suite baseline.** `test-parity/node_suite_baseline.json` still has `perf_hooks: {"pass": 87, "total": 148}` and this PR adds a fixture without touching it. `node_suite_regression_check.py` is a floor check — it fails only when `cur["pass"] < floor["pass"]` and explicitly allows improvements — so nothing goes red either way. But that cuts both ways: if the new fixture *passes*, the win isn't locked in and a later regression back to 87 would go unnoticed; if it *fails*, nothing reports that either. Bumping the floor to whatever this actually scores is what turns the fix into a ratchet. I did not do it here because I would have to run the full node-suite to get the honest number rather than guess it.

Issue #8272, opened 2026-08-17
parity: 16 tests CRASH in the first full run since July 4 (tui/ink, fastify, cluster, gc-unsafe-zones, wasm, mysql)


proggeramlug:
Companion to the parity dark-debt audit: 16 tests **crash** (signal/timeout — hard defects, not output nits) in the first complete parity run since 2026-07-04 (run 31964093732, merged 8-shard report):

- **perry-tui/ink-compat family (7)**: `test_perry_tui_inkcompat_{counter,useapp,useeffect,usefocus,usememo,useref}`, `test_issue_358_perry_tui_phase2_counter` (#348 area)
- **fastify family (4)**: `test_fastify_integration`, `test_issue_1240_fastify_request_json`, `test_issue_1293_fastify_request_json_as_any`, `test_issue_915_jwt_sign_after_async_route`
- **cluster**: `cluster_4962`
- **TUI demo**: `demo_claude_code_tui`
- **GC**: `test_issue_1425_gc_unsafe_zones` — a GC-unsafe-zone witness crashing is worth an early look
- **MySQL**: `test_issue_414_mysql_query_params`
- **WASM**: `test_wasm_add`

Six weeks of merges are in the window (suite dark since 2026-07-04; see the audit issue for why). Each family needs a bisect; the GC and wasm ones are likely the highest-signal. Crash-class entries in `known_failures.json` cite THIS issue so they cannot be mistaken for cosmetic output gaps.

Issue #8271, opened 2026-08-17
parity: 2026-08-17 dark-debt audit — 93 parity + 27 compile failures unlisted after six dark weeks (90.7% aggregate)


proggeramlug:
First complete parity run since 2026-07-04 (run 31964093732, the 8-shard `parity` job from #8244 — the suite was tag-gated + `continue-on-error` before #8187, and the one unsharded attempt since died at GitHub's 6-hour job cap). Against a 9-entry `known_failures.json`, the merged whole-suite report shows **six weeks of accumulated dark debt**:

| kind | count |
|--:|--|
| aggregate parity | **90.7%** (1125 pass / 98 parity_fail / 27 compile_fail / 17 crash / 5 node_fail / 13 skipped) |
| NEW parity failures (unlisted) | **93** |
| NEW compile failures (unlisted) | **27** |
| NEW crashes (unlisted) | **16** — tracked in the companion crash issue |
| STALE entries (listed but now pass) | 3 — `test_gap_zlib_4917_level`, `test_ramda_sum`, `test_sock_write_map` |

Recognizable families (not exhaustive):
- **perry-tui / ink-compat** (~15 across parity+crash): `test_perry_tui_inkcompat_*`, `test_issue_358_*`, `test_issue_679_*`, `test_issue_40x_*` — #348's area.
- **npm-package parity** (~15): `test_parity_lodash/moment/dayjs/date_fns/uuid/nanoid/validator/commander/cron/dotenv/decimal/lru_cache/argon2`, `test_lodash_*` — smells like one shared import/oracle change, worth bisecting as a family.
- **fastify** (4 crashes + several parity) — companion crash issue.
- **jwt/jose** (`test_issue_915_*`, `test_issue_927_*`, `test_jose_sign`, `test_jwt_sign_dynamic_alg`, `test_jose_signverify_roundtrip`) — the `js_jwt_sign_dyn` link family (#8224's item 2).
- **Linux-gtk4 UI compile fails** (~15 `test_ui_*` / media / canvas) — #8224's item 3 class: fixtures never skip-listed for a host without `libperry_ui_gtk4.a`.
- **The #8223 builtin-construct trio** (also red in gap-suite full mode).
[...]

Issue #8227, opened 2026-08-16
doc-tests: 15 untagged typescript fences across 12 docs files fail the docs lint


proggeramlug:
First `tier=full` run in weeks (31935729773). `doc-tests (macos-14)` fails at the docs lint step (`perry-doc-tests -- --lint docs/src`): **15 untagged `typescript` fences across 12 files** — each needs `{{#include}}` or a `typescript,no-test` tag. Recurring compliance drift (previously v0.5.381/v0.5.453) that accumulated while the job was tag-gated + continue-on-error.

Mechanical fix; the failing files are listed in the job log. The Windows doc-tests leg was cancelled by the macOS failure, so its state is unknown until this is fixed.

proggeramlug:
Correction from the completed run 31935729773: the Windows leg was **not** cancelled by the macOS failure (`fail-fast: false`). It ran from 09:20 to exactly 15:20 and was killed by GitHub's default 6-hour job cap — it hung or crawled on its own and has never been observed to completion. Once the fence lint is fixed and the macOS leg passes, the Windows leg needs its own investigation (likely a pwsh step hang; it was dark behind continue-on-error + tag-gating like everything else here).

proggeramlug:
Verified on current `main` (`280aafef4`): `cargo run --release --quiet -p perry-doc-tests -- --lint docs/src` passes. All 15 TypeScript fences reported here were tagged `typescript,no-test` by merged PR #8257 (merge commit `183d30c53a`), which landed later the same day this issue was opened. Closing as fixed.

Issue #8226, opened 2026-08-16
drizzle-mysql-smoke: fixture throws 'Cannot assign to read only property uniqueName' where node does not


proggeramlug:
First `tier=full` run in weeks (31935729773). MySQL service healthy, build and compile clean; the compiled drizzle fixture exits 1 at runtime:

```
TypeError: Cannot assign to read only property 'uniqueName' of object '#<Object>'
```

A readonly/frozen-property write that node accepts (sloppy-ish assignment path in drizzle-orm) throws in Perry-compiled code. Related in flavor to the `test_gap_prop_plan_cache_invalidation` triage (#8198's fragment: `js_throw_type_error_immutable_write` has no receiver awareness), but this one is a semantic difference, not just message text: node does not throw here.

Tracker refs: #488/#489 (drizzle+MySQL compile-as-package).

PR #8273, opened 2026-08-17
test(parity): baseline the six-week dark debt so the gate ratchets from today (#8271)

proggeramlug:
Companion to #8271 (audit) / #8272 (crashes) — makes the parity gate meaningful again after six dark weeks.

**What**: triages all 136 unlisted failures from the first complete parity run since July 4 (merged 8-shard report, run 31964093732) into `test-parity/known_failures.json` with per-family provenance (crashes → #8272, perry-tui/ink → #348, gtk4-host compile fails → `ci-env`, rest → #8271), and deletes the 3 stale entries the run exposed. From this baseline the bidirectional ratchet is live: any NEW failure and any entry that starts passing both go red, and every entry is a work-queue item (fix → delete in the same PR).

**Deliberately left red**: the #8223 builtin-construct trio. The provenance audit *refuses* to park them while `gap_snapshot.json` asserts they pass in fast mode — that divergence is the #8223 regression itself. Parity (and gap-suite full mode) stay red on exactly those three until it's fixed. That's the ratchet doctrine, not an oversight.

**Verification** (all local, against the real merged report):
- `parity_known_failures.py --report <merged> --platform linux`: green except exactly the trio
- `--audit`: green — 139 entries with provenance, 5 gap entries cross-checked against the snapshot
- `--self-test`: green

With this + #8244 (parity shards) + #8265 (merged), `full-suite-gate` becomes achievable: remaining known blockers are #8223 (trio), #8222 (aliased native class, fails cargo-test full), #8224/#8225 (never-could-pass build scripts), #8226, #8227 (+ the doc-tests-windows 6h hang).

## Summary by CodeRabbit

* **Documentation**
  * Added a changelog entry documenting the latest parity baseline and failure triage.
* **Tests**
  * Updated parity tracking with audited Linux failures, including crash, timeout, and TUI results.
  * Removed stale failure records and re-enabled bidirectional parity ratcheting.
  * Recorded provenance for newly categorized known failures.

### DIFF
--- a/changelog.d/8273-parity-debt-baseline.md
+++ b/changelog.d/8273-parity-debt-baseline.md
@@ -1,0 +1,17 @@
+### parity: baseline the six-week dark debt so the gate ratchets from today (#8271)
+
+The first complete parity run since 2026-07-04 (the 8-shard job from #8244;
+the suite was tag-gated + `continue-on-error` before #8187) surfaced 93
+unlisted parity failures, 27 compile failures and 16 crashes against a
+9-entry `known_failures.json`, at 90.7% aggregate. All 136 are triaged into
+`test-parity/known_failures.json` with provenance — crashes cite #8272
+(hard-defect work queue), the perry-tui/ink family cites #348, the rest cite
+the #8271 audit — and the 3 stale entries the run exposed
+(`test_gap_zlib_4917_level`, `test_ramda_sum`, `test_sock_write_map`) are
+deleted, so the bidirectional ratchet is live again: any NEW failure and any
+entry that starts passing both go red.
+
+Deliberately NOT parked: the #8223 builtin-construct trio. The provenance
+audit refuses those entries while `gap_snapshot.json` asserts the tests pass
+(fast mode) — the mode divergence IS the #8223 regression, so parity stays
+red on exactly those three until it is fixed, which is the ratchet working.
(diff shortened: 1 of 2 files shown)


proggeramlug:
Merging. A skip-list PR of this size deserves a suspicious read, so here is what I checked rather than took on faith.

**The claim that decides it — that the #8223 trio is deliberately left red — holds.** `grep -c 8223 test-parity/known_failures.json` returns **0**. Refusing to park a failure while `gap_snapshot.json` asserts it passes in fast mode, because that divergence *is* the regression, is the whole difference between a baseline and a cover-up. Good.

**Its own tooling agrees, and the tooling can fail:** `--self-test` OK, and `--audit` reports *"139 entries carry provenance and cross-checked 5 gap entries against gap_snapshot.json"*. The provenance spread on the added entries is 117 → #8271, 16 → #8272, 14 → #348, remainder `ci-env` — so every new entry is attributable to a family and a tracking issue, which is what #797 asked for and what the old list mostly lacked. The deletions include genuinely stale material (`test_parity_stream`, issue 793, added 2026-05-15).

**The bidirectional half is the part that makes this a ratchet rather than a bigger carpet.** An entry that starts passing goes red too, so a fix is forced to delete its own entry in the same PR — the same property that makes the `gc_root_dominance_allowlist` and `unrooted_local_shape` baselines honest. Without it, 136 new entries would be a way to stop looking.

Merged with no changes. The follow-on work is the 139 entries themselves, which are now a visible queue rather than six weeks of dark debt.

Issue #8225, opened 2026-08-16
native-abi-evidence-packet: check_runtime_symbols audits an archive that is never built (rlib-only #5422), + contract drift


proggeramlug:
First `tier=full` run in weeks (31935729773) — the packet was `continue-on-error`/label-gated. Three independent failures:

1. **Could never pass**: `check_runtime_symbols` step reports `libperry_runtime.a does not exist` — the script builds `-p perry-runtime` (rlib-only since #5422) instead of the `perry-runtime-static` wrapper, so the archive it audits is never produced. Broken since the #5422 split.
2. `native_abi_contract`: `FAIL: ... missing "consumer": "BufferNumericRead.native_u32"` — contract drift while dark.
3. Its `compiler_output` step fails 7 workloads overlapping the `compiler-output-regression` reds (see that job's issue trail).

Job wiring (checkout/toolchain/needs) verified clean — all three are content/build-script defects predating the CI restructure.

proggeramlug:
Precise root cause for item 1: `scripts/native_abi_evidence_packet.sh:309,331` run `cargo build -p perry -p perry-runtime` / `-p perry-runtime` alone — never `-p perry-runtime-static` — so `libperry_runtime.a` is never emitted (`perry-runtime` is rlib-only since #5422; this is the exact pitfall CLAUDE.md's "Verifying a runtime change" section documents). Script dates to 2026-07-02 (`ce0117a60`), so it has never once produced the archive it audits.

Item 3's 7 failing workloads overlap `compiler-output-regression`'s 4 and add `width_aware_buffer_kernels`, `native_owned_typed_views`, `native_pod_layout_constants`, `native_memory_bulk_fill`, `native_abi_packet_typed` — `hot_loops_no_runtime_calls` violations (`js_dynamic_string_or_number_add` / `js_number_coerce` in a hot loop).

Issue #8224, opened 2026-08-16
compile-smoke: test_precompile_basic could never pass — build step omits the -static wrapper crates (+ jwt/gtk4 never skip-listed)


proggeramlug:
First `tier=full` run in weeks (31935729773) — `compile-smoke` was `continue-on-error: true` + tag-gated, so none of this could ever block: 1209 passed, 5 failed, 64 skipped.

1. **Could never pass**: `test_precompile_basic`, `test_issue_414_mysql_query_params` fail with `Could not find libperry_runtime.a` — the job's build step is `cargo build --release` with no `-p`, which builds default-members only and never the `perry-runtime-static`/`perry-stdlib-static` wrapper crates (rlib-only split, #5422). Byte-identical build step before the restructure — this has been structurally unable to pass since the wrappers were introduced.
2. `test_jwt_sign_dynamic_alg`: `undefined reference to 'js_jwt_sign_dyn'` (jsonwebtoken never skip-listed; same class as jose #1025/#1038).
3. `test_ui_drag_drop`, `test_ui_text_alignment`: `libperry_ui_gtk4.a not found` (known Linux-gtk4 gap, never skip-listed).

Fix for (1) is one line: add `-p perry -p perry-runtime-static -p perry-stdlib-static` to the build step. (2)/(3) need skip-list entries or the real fix.

proggeramlug:
Resolved by #8275 (merged in 483022728): compile-smoke now builds the static wrapper crates explicitly and skip-lists the unsupported jsonwebtoken and Linux GTK4 fixtures called out here. No version bump was included.

Issue #5422, opened 2026-06-18
Make Perry builds lighter for development and explicit for release artifacts


TheHypnoo:
## Problem

Perry's current build path is too heavy for day-to-day development and mixes local feedback-loop needs with official distribution needs.

Observed locally on commit `6482c8757`:

- A cold `cargo build --release -p perry --timings` took roughly five minutes across an interrupted first run plus a resumed run.
- The final `perry` bin crate was the long pole in the timing report, taking about `118.4s` after dependencies were already built.
- The workspace currently has a large default build surface: 76 workspace members and 58 default members.
- Root `[profile.release]` is intentionally distribution-grade: ThinLTO, `codegen-units = 1`, `opt-level = 3`, and stripping.
- `perry-runtime` and `perry-stdlib` both declare `crate-type = ["rlib", "staticlib"]`, so CLI builds can produce large static archives even when the Rust CLI only needs rlibs.
- The `perry` CLI compiles all commands and backends unconditionally, including publish/mobile/updater/native flows that are not needed for most compiler development loops.

## Goal

Create a clear build taxonomy that makes Perry light for local development while keeping official release artifacts optimized, explicit, and measurable.

### 1. Add a fast developer profile

Add a named profile such as `perry-dev` that inherits from `release` but disables the expensive release-only settings:

```toml
[profile.perry-dev]
inherits = "release"
lto = false
codegen-units = 16
incremental = true
strip = false
opt-level = 1
```

Document the intended local loop:

- `cargo check -p perry` for fastest correctness feedback.
- `cargo build --profile perry-dev -p perry` for optimized local development.
- Keep `cargo build --release` / future `dist` builds for final artifact validation.

### 2. Split runtime/stdlib staticlibs from rlib consumers

Stop `perry-runtime` and `perry-stdlib` from producing static archives as a side effect of CLI builds.

Proposed shape:

- Make `perry-runtime` and `perry-stdlib` rlib-only crates.
- Add wrapper crates such as:
  - `perry-runtime-static`
  - `perry-stdlib-static`
- Release/linking workflows build the wrapper crates explicitly when `.a` artifacts are needed.
- Validate exported symbols with the existing symbol-checking scripts.

### 3. Add a slim developer CLI feature set

Keep the default official CLI full-featured, but add feature groups so contributors can build a smaller CLI for compiler work.

Suggested shape:

- `default = ["full-cli"]`
[...]

PR #8275, opened 2026-08-17
ci: compile-smoke + native-abi-evidence-packet build the -static wrappers they link/audit (#8224, #8225)

proggeramlug:
Fixes the build half of #8224 and #8225 — the two full-tier gates that have been **structurally unable to pass since the #5422 rlib split**, invisible for weeks behind `continue-on-error` + tag gating.

`perry-runtime`/`perry-stdlib` are rlib-only; `libperry_*.a` come from the `perry-runtime-static`/`perry-stdlib-static` wrapper crates (this is the exact pitfall CLAUDE.md's "Verifying a runtime change" documents):

* **`compile-smoke`** built with a bare `cargo build --release` (default-members only) → `test_precompile_basic` and `test_issue_414_mysql_query_params`, which link the archives directly, failed on every run this job ever had. The build step now names the wrapper crates (same set the parity harness builds). Also registers the three #8224 fixtures that were never skip-listed: `test_ui_drag_drop`/`test_ui_text_alignment` (Linux host has no `libperry_ui_gtk4.a` — appended to the existing ci-env family with the same rationale) and `test_jwt_sign_dynamic_alg` (`js_jwt_sign_dyn`, sibling of the already-listed `test_jose_signverify_roundtrip`).
* **`native_abi_evidence_packet.sh`** built `-p perry-runtime` and then audited the archive that build cannot emit — `check_runtime_symbols` has reported "libperry_runtime.a does not exist" since the script was created (2026-07-02). Both build arrays now use the wrapper crate.

Deliberately NOT touched: the packet's `native_abi_contract` drift and its 7 failing compiler-output workloads (#8225 items 2–3) are real content regressions and must stay red.

Validation: actionlint + `bash -n` clean; `pr-gate` runs on this PR; the fixed paths get their first true exercise on the next full-tier run (both jobs are full-tier only — and now they *can* go green, which per the four-ways doctrine is the property being restored).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compile-smoke checks by ensuring required static runtime and standard library components are built before linking fixtures.
  * Updated native ABI evidence checks to use static runtime archives consistently.
  * Excluded unsupported UI and JWT fixtures from smoke-test runs.

* **Documentation**
  * Added a changelog entry describing the static wrapper build and validation updates.


proggeramlug:
Merging. This is the same pitfall CLAUDE.md's "Verifying a runtime change" section documents, caught in the CI config itself — `perry-runtime`/`perry-stdlib` are rlib-only since #5422, so a build that doesn't name the `-static` wrappers emits no archive, and anything linking or auditing `libperry_*.a` was operating on a file that was never produced.

**`check_runtime_symbols` reporting "libperry_runtime.a does not exist" since 2026-07-02 is the striking part** — a gate that has been reporting its own subject missing for six weeks, invisible behind `continue-on-error` plus tag gating. Restoring the ability to go green is the property that matters here, per the four-ways doctrine.

**Verified:**

- The wrapper substitution is *complete*: all four sites in `native_abi_evidence_packet.sh` (debug and release, `packet_build` and `runtime_build`), not just the one that fails first.
- The compile-smoke build names the same crate set the parity harness uses, which also keeps cargo's feature unification identical — dropping or adding a `-p` changes that, so matching the harness is the right call rather than the minimal one.
- `yaml.safe_load` OK, `bash -n` OK.
- **actionlint adds nothing.** It reports 3 findings on this branch, and `origin/main` reports the *same* 3 at line 3272 instead of 3280 — shifted exactly by the 8 added lines, all `info`/`style`. My first attempt at that comparison was vacuous: I ran actionlint on main's file at a `/private/tmp` path and got 0, because actionlint only analyzes files under `.github/workflows/`. The zero meant "not analyzed", not "clean".

**The three skip-list additions are justified and land in a documented family.** `test_ui_drag_drop` and `test_ui_text_alignment` join the existing ci-env block whose rationale — the bare `perry foo.ts -o out` smoke path doesn't pull in `libperry_ui_*` on Linux, so they fail with ld undefined-symbol errors — applies to them unchanged. `test_jwt_sign_dynamic_alg` sits beside `test_jose_signverify_roundtrip` for the same unlinked-ext-symbol reason.

**And it does not overreach:** the packet's `native_abi_contract` drift and its 7 failing compiler-output workloads stay red, because those are content regressions rather than build plumbing. Fixing the harness without laundering the findings is the right split.

One fix pushed: the changelog fragment was named `8274-static-wrapper-builds.md`; #8274 is a different PR that merged an hour ago. Renamed to `8275-`.

PR #8270, opened 2026-08-17
feat(windows): host MapView in a XAML Island

proggeramlug:
## Summary
- replace the Windows MapView text placeholder with a native `Windows.UI.Xaml.Controls.Maps.MapControl` hosted by `DesktopWindowXamlSource`
- wire region/zoom, titled `MapIcon` pins, clearing, and Road/Aerial/AerialWithRoads styles through the existing FFI
- pre-translate XAML Island messages, resize and release the island with its host HWND, and embed the Windows 10 1903 compatibility manifest required by unpackaged XAML Islands
- document the map-service-token environment variables and key setup

## Testing
- `cargo check -p perry-ui-windows`
- `cargo test -p perry app_manifest_enables_xaml_islands` (with `LLVM_SYS_221_PREFIX=C:\llvm`)
- `python scripts/check_test_registration.py`
- `cargo fmt -p perry -p perry-ui-windows -- --check`
- native Windows smoke: initialized `WindowsXamlManager`, attached a `DesktopWindowXamlSource`, created a MapControl, applied center/zoom/style, added a titled MapIcon, and verified `MapElements.Size() == 1`

Closes #559

## Summary by CodeRabbit

* **New Features**
  * Added native Windows map rendering with interactive maps, zooming, panning, styles, and pins.
  * Added support for clearing pins and updating map center and zoom.
  * Added configuration through `PERRY_MAP_SERVICE_TOKEN` or `PERRY_BING_MAPS_KEY`.

* **Documentation**
  * Updated MapView documentation to describe Windows support and required map-token configuration.

* **Bug Fixes**
  * Improved Windows UI message handling for map interactions.
  * Added compatibility support required for Windows 10 XAML Islands.

PR #8276, opened 2026-08-17
fix(perf_hooks): complete timerify parity

### LINKED ISSUE #8234: [parity] node:perf_hooks timerify — async settlement, construct, entry arguments, and name/length descriptors

proggeramlug:
## Summary

- defer `performance.timerify()` entries and histogram samples until returned promises settle
- preserve constructor behavior, call arguments, result identity, and class-call errors through timerify wrappers
- expose Node-compatible entry argument aliases and wrapper `name`/`length` descriptors
- split timerify into a topical submodule to keep the runtime source within the repository's file-size limit

## Testing

- all 14 `node-suite/perf_hooks/timerify` parity fixtures with `PERRY_SKIP_BUILD=1`

Closes #8234

### DIFF
--- a/changelog.d/8276-perf-hooks-timerify.md
+++ b/changelog.d/8276-perf-hooks-timerify.md
@@ -1,0 +1,11 @@
+### fix(perf_hooks): complete `performance.timerify()` parity
+
+Timerified async functions now emit their performance entry and optional
+histogram sample when the returned promise settles, rather than when it is
+created. The wrapper also preserves constructor behavior, including class-call
+errors and `new.target` forwarding, while returning the original call result.
+
+Function entries now expose call arguments through both `entry.detail` and
+indexed entry properties. Timerify wrappers also carry Node-compatible
+non-writable, enumerable, non-configurable `name` and `length` descriptors.
+Fixes #8234.
--- a/crates/perry-runtime/src/object/descriptors.rs
+++ b/crates/perry-runtime/src/object/descriptors.rs
@@ -538,6 +538,7 @@ pub extern "C" fn js_object_get_own_property_descriptor(obj_value: f64, key_valu
                     // `Object.defineProperty`) overrides those defaults.
                     let registered = super::get_property_attrs(ptr, name);
                     let writable_default = registered.map(|a| a.writable());
+                    let enumerable_default = registered.map(|a| a.enumerable());
                     let configurable_default = registered.map(|a| a.configurable()).unwrap_or(true);
                     if let Some(acc) = super::get_accessor_descriptor(ptr, name) {
                         let attrs =
@@ -583,7 +584,7 @@ pub extern "C" fn js_object_get_own_property_descriptor(obj_value: f64, key_valu
                             Some((
                                 arity as f64,
                                 writable_default.unwrap_or(false),
-                                false,
+                                enumerable_default.unwrap_or(false),
                                 configurable_default,
                             ))
                         }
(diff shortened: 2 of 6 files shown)


proggeramlug:
Merging. Runtime suite **2567 passed / 0 failed**, all six script gates and `check_file_size` clean, `cargo fmt` clean, and the node-suite timerify fixtures run cleanly under the pinned oracle.

**Extracting `timerify.rs` is the right call and it fixes a problem I flagged on #8254.** That PR left `perf_hooks.rs` at **1998 lines** — two under the 2000-line cap, so the next addition would have tripped `check_file_size.sh`. This takes it to **1920** by moving 204 lines into a submodule, which is the split I suggested there. More of that file should go the same way.

**One observation, not a blocker:** the runtime suite contains **zero** tests matching `timerify`, so the behaviour added here is covered only by the node-suite fixtures, which are nightly-only and floor-gated. That means a regression in this code has no per-PR signal — the same structural gap #8269 has. Worth a unit test or two on the settle/pending record transitions if this area keeps growing; the oracle output (`pending records: 0` → `settled records: 1`) suggests there is a clean state machine to assert on.

**Unrelated, but discovered while auditing this PR and now fixed on main:** `gc_runtime_root_holders` was failing here, and it was not this PR — #8270 (Windows MapView) added two thread-locals to `map_view.rs`, pinned `MAPS`, but missed its `#[cfg(target_os = "windows")]` sibling `HWND_TO_HANDLE`, so the census had been red on `main` since that merge. That is my miss: I ran cargo-audit, the `perry` tests and fmt on #8270 but not the census gate. Pinned it on main — the value is a 1-based widget handle rather than a NaN-boxed callback, so it cannot park a closure and pinning is the right disposition rather than a scanner.

Issue #8243, opened 2026-08-16
perf: pipeline's +3.95% at #8084 is unattributed — it is NOT the callee rooting (IR is byte-identical)


proggeramlug:
Splitting this out of #8159, which is closed by #8240: the **narrowing question is answered, the 3.9% is not**.

## What is established

`pipeline` retired **+3.95% instructions** at `53d63aad2` (#8084) versus `8b1b4b909` (#8157) — measured in #8159, run ranges non-overlapping, 2x the noise floor, every other row flat.

#8159 attributed that to `fix(codegen): root the callee across argument evaluation in three call arms`. **That is falsified.** Three independent confirmations on `07c8040bf`:

1. A codegen change to exactly those arms (#8240) leaves the emitted IR for `gc-handoff/apps/pipeline.ts` **byte-identical** — `cmp` returns 0, not "within noise".
2. Both `js_closure_call1` sites in that IR are preceded by `call i64 @js_closure_unbox_callee_checked(...)`, which only `lower_dynamic_closure_call` (`lower_call/console_promise.rs`) emits. That is the #7154 lowering, which has rooted its callee and re-read its arguments below the unbox since long before #8084.
3. `js_new_function_construct` and `js_closure_call_apply_with_spread` appear in the module **only as `declare` lines**. There is no call to either.

Corpus-wide, #8240 measures `pipeline` +0.02%, `interp` −0.04%, `iso_miss` −0.05%, `asyncpipe` +0.11%, `shapes` −0.18% — identical runtime archives on both arms, per-arm cache dirs, `PERRY_NO_AUTO_OPTIMIZE=1`, min-of-5, stdout sha-identical.

So the commit attribution stands and the hunk attribution does not. The cost is in #8084's other ~2,500 lines.

## Search space, ranked

Most of the remainder is GC runtime, which a program allocating ~1.4M objects exercises constantly:

- `gc/roots/stack_maps.rs` + `_decode.rs` + `_sections.rs` — ~1,660 lines changed; runs on every root scan
- `gc/copying.rs` (−306) + `gc/copying_pointer_set.rs` (+248) — the copying minor's pointer set, restructured
- `gc_map.rs` (+355, codegen) — bigger/differently-shaped maps cost scan time
- `gc/pin.rs` (+209), `gc/tenuring.rs` (+41)
- `object/this_binding.rs`, `object/native_call_method.rs`

[...]

PR #8277, opened 2026-08-17
fix(gc): validate arena object starts during copying

### LINKED ISSUE #8256: gc(copying): fabricated Map/Set with a correct-size header still passes plausible_gc_header — needs object-start validation
Follow-up to the Critical CodeRabbit finding on [#8251](https://github.com/PerryTS/perry/pull/8251) (split out rather than forced into that merged PR).

## The gap

[#8251](https://github.com/PerryTS/perry/pull/8251) tightened `plausible_gc_header` so a fixed-layout type's `size` must equal its known total (Map/Set total = 24). That closes the observed SIGSEGV vector, where a pointer-shaped word in a live object's payload was read as a header with `size` ≈ 1024–1040 and fabricated into a fake Map.

But a fabricated header with a **correct size** (24) is indistinguishable from a genuine one by header fields alone — `obj_type=8`, `flags` plausible, `size=24` all match. `plausible_gc_header` can't reject it, because it validates the header's *fields*, not whether the address is actually an object *start*. `classify_arena` confirms the address and `addr - 8` are in heap space, then trusts the pattern match; nothing checks the address is an object boundary.

The PR's own regression test documents this explicitly (`crates/perry-runtime/src/gc/tests/copying/fabricated_map_rejection.rs:193-223`): a correctly-sized fabricated Map header passes. The entries/elements tripwire backstops the read, but the fabrication itself is still possible.

## Why it's not already fixed

The load-bearing claim for the size check being sound for genuine objects was verified: `hot_arena_free_list` / `ARENA_FREE_LIST` is never populated anywhere (every `borrow_mut()` is a `.clear()`; the nonempty flag is never set true), so `size == total` holds for every genuine arena Map/Set. That makes the size check correct *for the observed vector* — but it does not address the correct-size coincidence.

## The real fix (heavier than a header tweak)

Give `classify_arena` a real object-start check — a per-page object-start bitmap, or validate the candidate address against the arena's object walk. This is the principled fix mature conservative collectors use. The cost must be designed, not guessed: the copying minor exists specifically to avoid O(all-objects) work, so a bitmap (or an incremental walk) is the likely shape rather than a full scan.

## Note

A correct-size fabricated header requires a pointer-shaped word whose low byte is `0x08` (GC_TYPE_MAP) AND whose next byte has GC_FLAG_ARENA AND whose top 32 bits read as `size=24` — i.e. a pointer into a very low address (≈0x18000000 = 24<<24). That is a much narrower coincidence than the ~1024-size vector, so this is lower-frequency than what #8251 closed, but it is a real soundness gap.

Related: #8251 (the size check), and the Finding-2 follow-up that unified the entries/elements tripwires on `is_plausible_heap_addr`.

proggeramlug:
Fixes #8256

## Summary

- add a per-arena-block object-start bitmap with one bit per 8-byte allocation slot (1.5625% metadata overhead)
- stamp both runtime and compiler-generated inline allocations, and keep the bitmap synchronized across block reset, promotion, quarantine, and reuse
- require an exact recorded allocation boundary in the copying-minor classifier, rejecting correct-size fabricated Map/Set headers in payload bytes with an O(1) lookup and no all-object walk
- arm old-object shared shape edges at publication so their young keys arrays are rewritten in the same minor under exact-start validation

## Tests

- cargo test -p perry-runtime --lib (2558 passed, 4 ignored)
- cargo check -p perry-codegen --lib
- cargo test -p perry-codegen --lib the_inline_allocator_stores_its_header_prefix_as_one_vector_image
- cargo fmt --all -- --check
- git diff --check

No version bump included.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved garbage collection accuracy by tracking valid object boundaries during arena allocation.
  * Prevented invalid or fabricated memory addresses from being mistaken for live objects.
  * Improved handling of shared object metadata during minor collections.
  * Ensured allocation metadata is correctly reset when memory blocks are recycled or cleared.
* **Tests**
  * Added regression coverage for object-boundary validation, arena resets, and overflow-object collection scenarios.

### DIFF
--- a/crates/perry-codegen/src/expr/array_literal.rs
+++ b/crates/perry-codegen/src/expr/array_literal.rs
@@ -201,6 +201,26 @@ pub(crate) fn lower_array_literal(ctx: &mut FnCtx<'_>, elements: &[Expr]) -> Res
             // GC_STORE_AUDIT(INIT): freshly allocated array header starts pointer-free until slot notes below.
             blk.store(I64, &gc_packed.to_string(), &raw);
 
+            // Publish the exact header boundary in the current arena block's
+            // object-start bitmap (InlineArenaState field at byte 24). The
+            // slow arm may have replaced the current block, so compute the
+            // slot from the merged raw pointer rather than `aligned_off`.
+            let current_data = blk.load(PTR, &state_ptr);
+            let raw_addr = blk.ptrtoint(&raw, I64);
+            let data_addr = blk.ptrtoint(&current_data, I64);
+            let relative = blk.sub(I64, &raw_addr, &data_addr);
+            let start_slot = blk.lshr(I64, &relative, "3");
+            let start_word = blk.lshr(I64, &start_slot, "6");
+            let start_bit = blk.and(I64, &start_slot, "63");
+            let bitmap_field = blk.gep(I8, &state_ptr, &[(I64, "24")]);
+            let bitmap = blk.load(PTR, &bitmap_field);
+            let bitmap_word = blk.gep(I64, &bitmap, &[(I64, &start_word)]);
+            let prior_starts = blk.load(I64, &bitmap_word);
+            let start_mask = blk.shl(I64, "1", &start_bit);
+            let updated_starts = blk.or(I64, &prior_starts, &start_mask);
+            // GC_STORE_AUDIT(INIT): side metadata for a freshly initialized GC header.
+            blk.store(I64, &updated_starts, &bitmap_word);
+
             // Packed ArrayHeader at raw+8 (length low 32 / capacity high 32).
             let arr_header_addr = blk.gep(I8, &raw, &[(I64, "8")]);
             let arr_header_packed = (n as u64) | ((n as u64) << 32);
--- a/crates/perry-codegen/src/lower_call/alloc_hot_tests.rs
+++ b/crates/perry-codegen/src/lower_call/alloc_hot_tests.rs
@@ -342,6 +342,10 @@ fn the_inline_allocator_stores_its_header_prefix_as_one_vector_image() {
         ir.contains(HEADER_IMAGE_STORE),
         "the inline allocation site must store the `<2 x i64>` header image:\n{ir}"
     );
+    assert!(
+        ir.contains("shl i64 1,") && ir.contains("lshr i64") && ir.contains(", 6"),
+        "the inline allocation site must set its exact boundary in the arena object-start bitmap:\n{ir}"
+    );
     // The compose lives beside the ShapeId mint in module init, i.e. after the
     // mint call and outside the allocating function's own body.
     let compose_at = ir.find(HEADER_IMAGE_COMPOSE).unwrap();
--- a/crates/perry-codegen/src/lower_call/new_alloc.rs
+++ b/crates/perry-codegen/src/lower_call/new_alloc.rs
@@ -644,6 +644,26 @@ fn emit_instance_alloc_inner(
                 header_image, raw
             ));
 
+            // Publish the exact header boundary in the current arena block's
+            // object-start bitmap (InlineArenaState field at byte 24). The
+            // slow arm may install a new block, so derive the slot from the
+            // merged raw pointer and the state data pointer after the merge.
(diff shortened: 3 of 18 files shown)


proggeramlug:
Correctness looks good — `cargo test -p perry-runtime --lib` gives **2568 passed, 0 failed, 4 ignored**, and the design is the architecturally right answer to #8256: an exact recorded allocation boundary is a real invariant, where #8251's `size == 24` was a heuristic that a correct-size fabrication defeats.

**I'm holding on one thing, and it is the owner's own standing rule rather than my preference: this ships an unmeasured cost on the hottest path, and the PR reports neither instructions nor RSS.**

The Tests section lists correctness only. Two costs are visible in the diff:

**1. ~14 instructions added to every inline allocation.** `lower_call/new_alloc.rs` now emits, on the fast path:

```
load(current_data) · ptrtoint(raw) · ptrtoint(data) · sub · lshr · lshr · and
gep · load(bitmap) · gep · load(prior_starts) · shl · or · store
```

That is 3 loads, 1 store and 6 ALU ops per allocation, and the bitmap word is on a different cache line from the object being allocated, so it is plausibly an extra miss per allocation rather than 14 cheap ops. For scale: #8252 recovered **−3.89%** on `pipeline` by deleting *one* `OnceLock::get_or_init` from a path taken 2.9M times, and allocation is far hotter than that.

**2. +1.5625% permanent metadata on every arena block.** Small per block, but it is RSS that never comes back, on a campaign whose explicit rule is that RSS is minimized *always* and not traded for anything.

I'm not asserting this is too expensive — I'm saying nobody has measured it, and this is exactly the pair the project requires reported together. The measurement that would settle it is the usual one: both arms built with `-p perry -p perry-runtime-static -p perry-stdlib-static`, `PERRY_RUNTIME_DIR` pinned per arm, `PERRY_NO_AUTO_OPTIMIZE=1`, instructions-retired min-of-5 on an allocation-heavy corpus row plus peak RSS, on the quiet mini.

Two ways forward, and it's your call:

- **Measure first** — I can run it, or the author can; if it lands under noise this merges immediately.
- **Merge now and measure after**, accepting that a hot-path regression would be attributed later rather than at the boundary, which is how #8243's +3.95% stayed unattributed from #8084 for weeks.

Everything else I checked is clean and I have no other objection.

proggeramlug:
Merging on your call. Recording the open item so it is attributable later rather than lost.

**Fully verified on current `main`:** `perry-runtime --lib` **2568 passed / 0 failed / 4 ignored**; `perry-codegen --no-fail-fast` **28 suites, 1510 passed, 9 failed**, all nine in the known baseline set; `cargo fmt`, `check_file_size`, and all five script gates clean.

**The design is the right answer to #8256.** An exact recorded allocation boundary is a real invariant; #8251's `size == 24` was a heuristic that a correct-size fabrication in payload bytes defeats by construction. Doing it as an O(1) bitmap lookup rather than an all-object walk is the right shape, and keeping the bitmap synchronized across reset, promotion, quarantine and reuse is the part that would have been easy to half-do.

**Still unmeasured, and merged anyway:**

1. **~14 instructions on every inline allocation.** `new_alloc.rs` emits 3 loads, 1 store and 6 ALU ops on the fast path, and the bitmap word sits on a different cache line from the object being allocated, so the real cost may be a miss rather than the op count. For calibration, #8252 recovered −3.89% on `pipeline` by removing *one* `OnceLock::get_or_init` from a path taken 2.9M times; allocation is hotter than that.
2. **+1.5625% permanent metadata on every arena block.**

Neither number exists yet. If a later sweep shows an allocation-heavy regression, this is the first commit to bisect to — the measurement to run is both arms with `-p perry -p perry-runtime-static -p perry-stdlib-static`, `PERRY_RUNTIME_DIR` pinned per arm, `PERRY_NO_AUTO_OPTIMIZE=1`, instructions-retired min-of-5 plus peak RSS on the quiet mini. Worth doing proactively rather than waiting for it to surface as an unattributed delta, which is exactly how #8243's +3.95% sat unexplained from #8084 for weeks.

PR #8278, opened 2026-08-17
fix: preserve aliased native constructor proofs

### LINKED ISSUE #8222: cargo-test (full): aliased native-class import loses native methods (aliased_native_class_import, 2 tests)
Found by the first `tier=full` CI run in weeks (run 31935729773; the suite was dark behind tag-gating since ~July 4). Full `cargo test --release --workspace` fails 2/2463:

- `crates/perry/tests/aliased_native_class_import.rs::aliased_net_socket_keeps_native_methods`
- `crates/perry/tests/aliased_native_class_import.rs::aliased_matches_unaliased_for_native_class`

`import { Socket as Sk } from "net"; new Sk()` resolves `s.connect`/`s.write`/`s.destroy`/`s.end` as `undefined` instead of `function` — an aliased import of a native base class loses its native-method dispatch (the CLAUDE.md "native base-class subclassing keyed on a literal name" weak area, aliased-import shape).

Per-PR `cargo-test` is `--lib --bins` only, so this can only go red in the sweep/full tiers (`main-gate`/`full-suite-gate`), where it now does. Repro: `cargo test -p perry --test aliased_native_class_import`.

proggeramlug:
## Summary

- preserve canonical runtime type proofs for constructors imported from native modules
- recover aliased named imports through their original export names
- reject non-class, non-native, and ambiguous import-source proofs

## Testing

- `cargo test -p perry-codegen --lib imported_native_constructor_proof --release -- --nocapture`
- `cargo test -p perry --test aliased_native_class_import --release aliased_net_socket_keeps_native_methods -- --exact --nocapture`
- `cargo test -p perry --test aliased_native_class_import --release aliased_matches_unaliased_for_native_class -- --exact --nocapture`

Fixes #8222

### DIFF
--- a/changelog.d/8278-aliased-native-constructor-proofs.md
+++ b/changelog.d/8278-aliased-native-constructor-proofs.md
@@ -1,0 +1,7 @@
+### fix(codegen): preserve aliased native constructor proofs
+
+Native classes imported under a local alias now retain their canonical
+constructor proof through codegen. Reading `net.Socket` methods as values from
+an aliased instance therefore returns bound functions, matching the unaliased
+import. The proof remains gated to unambiguous native class entries in the API
+manifest. Fixes #8222.
--- a/crates/perry-codegen/src/type_analysis.rs
+++ b/crates/perry-codegen/src/type_analysis.rs
@@ -51,6 +51,8 @@ pub(crate) use predicates::{
 // `super::*`) can keep calling `tuple_index_literal` directly.
 #[cfg(test)]
 pub(crate) use predicates::tuple_index_literal;
+#[cfg(test)]
+pub(crate) use refine::is_imported_native_constructor_class;
 pub(crate) use refine::{
     compute_auto_captures, declared_array_property_claim, is_crypto_digest_chain,
     is_global_constructor_expr, is_process_namespace_version_property, proven_type_from_init,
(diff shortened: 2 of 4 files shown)


proggeramlug:
Merging. This clears #8222, which #8273 named as one of the remaining `full-suite-gate` blockers.

**The design is fail-closed in the direction that matters.** A wrong proof here is a miscompile — native method dispatch applied to a class that isn't native — so `is_imported_native_constructor_class` refusing in every doubtful case is right: two candidate sources resolving to different modules returns `false`, no resolved source returns `false`, and even then the `API_MANIFEST` must carry a matching `module` + `name` + `ApiKind::Class`. Recovering the alias by scanning `imported_class_original_names` for locals whose original is the canonical name is the piece that was missing, and it's chained with the direct lookup rather than replacing it.

**Verified here:**

- `cargo test -p perry --test aliased_native_class_import` — **7 passed, 0 failed**, and crucially that includes `aliased_user_class_import_is_not_treated_as_native`. A proof-granting change whose tests only cover the granting direction would be half-tested; this covers the refusal too.
- `cargo test -p perry-codegen --no-fail-fast` — 24 suites, **1480 passed, 8 failed**, all eight in the known baseline. (The first run stopped at 23 suites when the disk filled; I ran `typed_shape_descriptors` separately — 18 passed, 0 failed — rather than report a partial sweep as a clean one.)
- `cargo fmt --all --check`, `check_test_registration.py`, `check_file_size.sh` all clean.

PR #8279, opened 2026-08-17
fix(perf_hooks): expose namespace default export

### LINKED ISSUE #8236: [parity] import * as ns — ns.default is undefined for native modules whose default is the namespace itself
Split out of #6766 after PR #8229. **Not a `perf_hooks` bug** — `perf_hooks` is just where the node-suite pins it.

`import * as ns from "node:<mod>"` should expose `ns.default` (Node lists it among the namespace keys, and it is the CJS `module.exports`). For `perf_hooks` the key is now listed but the read is `undefined`:

```js
import * as hooks from "node:perf_hooks";
Object.keys(hooks)          // ✅ includes "default" after #8229
hooks.default               // ❌ undefined
hooks.default.performance === hooks.performance
// node:  true
// perry: TypeError: Cannot read properties of undefined (reading 'performance')
```

Pinned by `test-parity/node-suite/perf_hooks/imports/namespace-keys` (line 1 passes, line 2 does not). Reproduce with `./run_parity_tests.sh --suite node-suite --module perf_hooks`.

## What I found

`cjs_default_export_value("perf_hooks")` in `crates/perry-runtime/src/object/native_module.rs` now returns the module namespace (added in #8229), and `get_native_module_constant` has a generic `property == "default"` branch ahead of its per-module match that consults it. Despite that, the read still resolves to `undefined`, so the `PropertyGet { NativeModuleRef(mod), "default" }` lowering — `js_native_module_esm_export_value` → `native_module_property_by_name_impl(…, /* is_cjs */ false)` — appears to return before either is reached. I did not chase it further.

Worth checking against the modules that resolve `default` through the separate `<mod>.default` tag machinery (`cjs_default_namespace_name`: `os`, `path`, `util`, `url`, …) — those may work by a different route, which would make this specifically about modules whose `default` is the namespace itself (`dgram`, `perf_hooks`).

proggeramlug:
## Summary

- declare `perf_hooks.default` in the native API manifest so namespace reads reach the runtime resolver
- preserve the Node identity relationship between `hooks.default.performance` and `hooks.performance`
- add a lowering regression for the namespace default read
- no version bump

Closes #8236

## Tests

- `cargo test -p perry-hir native_perf_hooks_namespace_default_reaches_runtime_lookup`
- `PERRY_SKIP_BUILD=1 PERRY_BIN="$PWD/target/release/perry" PERRY_RUNTIME_DIR="$PWD/target/release" ./run_parity_tests.sh --suite node-suite --module perf_hooks` (142/148 parity passes; 0 compile failures; `imports/namespace-keys` passes; six pre-existing receiver-branding/timerify mismatches remain)

## Summary by CodeRabbit

* **New Features**
  * Added ESM namespace access to the CommonJS exports of `node:perf_hooks` through the `default` property.

* **Bug Fixes**
  * Fixed namespace imports from `node:perf_hooks` so the `default` export is preserved correctly.

### DIFF
--- a/crates/perry-api-manifest/src/entries/part_3.rs
+++ b/crates/perry-api-manifest/src/entries/part_3.rs
@@ -1389,6 +1389,11 @@ pub(crate) const API_MANIFEST_PART_3: &[ApiEntry] = &[
     method("perf_hooks", "createHistogram", false, None),
     internal_property("perf_hooks", "timeOrigin"),
     internal_property("perf_hooks", "nodeTiming"),
+    // The ESM namespace exposes the underlying CJS `module.exports` object as
+    // `default`. Keep this in the manifest so the native-module member gate
+    // lets the runtime resolve that object instead of folding the read to
+    // `undefined` before codegen (#8236).
+    property("perf_hooks", "default"),
     property("perf_hooks", "performance"),
     property("perf_hooks", "constants"),
     class("perf_hooks", "Performance"),
--- a/crates/perry-hir/src/lower/tests.rs
+++ b/crates/perry-hir/src/lower/tests.rs
@@ -273,14 +273,10 @@ export function perfHooksDefault() {
   return hooks.default;
 }
 "#;
-    let module = perry_parser::parse_typescript(source, "perf-hooks-default.ts")
-        .expect("source parses");
-    let hir = super::lower_module(
-        &module,
-        "perf-hooks-default",
-        "perf-hooks-default.ts",
-    )
-    .expect("source lowers");
+    let module =
+        perry_parser::parse_typescript(source, "perf-hooks-default.ts").expect("source parses");
+    let hir = super::lower_module(&module, "perf-hooks-default", "perf-hooks-default.ts")
+        .expect("source lowers");
     let function = hir
         .functions
         .iter()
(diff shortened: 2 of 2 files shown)


proggeramlug:
Merging. `perry-hir --lib` **315 passed / 0 failed** with `native_perf_hooks_namespace_default_reaches_runtime_lookup` among them, and `cargo fmt` now clean.

**No interaction with #8278**, which I checked because that PR just made `API_MANIFEST` load-bearing for constructor type proofs: this adds `property("perf_hooks", "default")`, and `is_imported_native_constructor_class` requires `ApiKind::Class`, so a property entry cannot accidentally grant a native-class proof.

**One fix pushed:** `cargo fmt --all --check` failed on the new test at `lower/tests.rs:273`. Fixed in place.

**Worth acting on separately:** you report node-suite perf_hooks at **142/148**, and `test-parity/node_suite_baseline.json` still carries the floor at **87/148**. Between #8229, #8254, #8269, #8276 and this, that module has moved a long way today, and none of it is locked in — `node_suite_regression_check.py` is a floor check that allows improvements silently, so perf_hooks could regress all the way back to 87 without going red. Bumping the floor to the real number is what converts five PRs' worth of work into a ratchet. I have not done it here because the honest number needs a full node-suite run rather than a guess, and it should probably be one commit covering the whole module rather than riding on this PR.

PR #8280, opened 2026-08-17
fix: throw for array-only methods on typed arrays

### LINKED ISSUE #8138: Array methods absent from %TypedArray%.prototype answer instead of throwing (flat, flatMap, push, pop, shift, unshift, splice, toSpliced)
## Summary

Nine `Array.prototype` methods that **do not exist** on `%TypedArray%.prototype`
answer with a value on a typed-array receiver instead of throwing. Node throws
`TypeError: … is not a function`; perry returns `[]`, `undefined`, or garbage.

```ts
function opaque(x: any): any { return x; }

opaque(new Int32Array([1, 2, 3])).flatMap((x: any) => [x])
// node:  TypeError: opaque(...).flatMap is not a function
// perry: [5.08e-321, 2.1219957924e-314, 1e-323]
```

Measured against node `v26.5.1` on `main` @ `0a1e78e5f`.

## Full list

The methods absent from `%TypedArray%.prototype` are exactly:
`concat`, `flat`, `flatMap`, `pop`, `push`, `shift`, `splice`, `toSpliced`,
`unshift`.

| expression | perry | node |
|---|---|---|
| `i32.concat([9])` | `[9]` | `TypeError` |
| `i32.flat()` | `[]` | `TypeError` |
| `i32.flatMap(f)` | `[5.08e-321,2.12e-314,1e-323]` — **garbage** | `TypeError` |
| `i32.push(9)` | `[]` | `TypeError` |
| `i32.pop()` | `undefined` | `TypeError` |
| `i32.shift()` | `undefined` | `TypeError` |
| `i32.unshift(9)` | `[]` | `TypeError` |
| `i32.splice(1,1)` | `{obj}`, receiver untouched | `TypeError` |
| `i32.toSpliced(1,1)` | `[]` | `TypeError` |
| `u8.flat()` | `[]` | `TypeError` |
| `u8.toSpliced(1,1)` | `[]` | `TypeError` |
| `u8.push(9)` | `[]` | `TypeError` |

`u8.splice`, `u8.concat` and `u8.flatMap` DO throw — but with
`TypeError: (Buffer).splice is not a function`, i.e. right answer, wrong
message and wrong reason (`dispatch_buffer_method`'s name list simply has no
entry, rather than the property genuinely being absent).

## Why this is NOT the #8041 receiver-reroute family

These helpers are dead below `clean_arr_ptr` in the same textual sense as the
ones #8090/#8109/#8119/#8120/#8130 fixed, and it is tempting to file them
together. They are a different bug, and "fixing the reroute" would make them
*worse*:

`js_array_concat` is the worked example, already noted in #8130's caveats. Node
gives `[{}]` for a typed-array argument; perry gives `[]` today; and the
pre-#8041 **live** reroute gave `[1,2,3]`. It was wrong before #8041 and wrong
after, in different ways. Waking the reroute back up would restore an answer
node never produces.

The actual defect is one level up: **the property should not resolve at all.**
Perry's method dispatch answers `flatMap` on a typed array because the fused
`Expr::ArrayFlatMap` fold and `is_array_expr` both treat `Int32Array` as an
array, and nothing consults `%TypedArray%.prototype`'s membership. So the fix
is a receiver-kind check at *method resolution*, producing the spec
`TypeError`, not a reroute inside the helper.

`js_array_concat` specifically is being handled by **#8124**
(concat-spreadable); this issue covers the other eight.

## Suggested shape

A single table of the nine names, consulted where the typed-array/Buffer
dispatch decision is already made — `dispatch_typed_array_method`
(`object/native_call_method/typed_array.rs`) and `dispatch_buffer_method`
(`object/buffer_dispatch.rs`) already return `None` / "is not a function" for
unrecognised names, so the missing piece is preventing the *static* HIR folds
from claiming these nine for a receiver whose type is a typed array. That is
the same gate `is_array_expr` already applies to exclude `Uint8Array` from
`copyWithin`'s fold (`local_array_methods.rs:800-839`), generalised.

Low urgency: every row here is a program that is already broken and would
throw under node, so no working code depends on the current answers. The
`flatMap` row is the one worth prioritising, because it returns **garbage
values** rather than an empty result, which a caller can propagate silently.

## Context

Found by the exhaustive `Array.prototype` receiver sweep that produced #8135.
The full 33-row table, with the funnel verdict and measured node diff for every
`Array.prototype`-reachable runtime entry point, is in the #8135 PR description.

proggeramlug:
## Summary
- keep array-only methods absent from `%TypedArray%.prototype` out of Array-specific lowering
- preserve own and prototype user methods before throwing `TypeError` for missing methods
- cover all eight affected methods, dynamic dispatch, non-mutation, and overrides end to end

Fixes #8138

## Testing
- `cargo test --profile perry-dev -p perry --test issue_8138_typed_array_method_availability -- --nocapture`
- `cargo test --profile perry-dev -p perry-hir -p perry-runtime --lib`
- `cargo test --profile perry-dev -p perry-codegen --lib` (1,063 passed; two unrelated Windows object-byte determinism tests fail because temporary object paths differ)
- `./scripts/test_affected_crates.sh --base origin/main` (runtime: 2,498 passed, 4 ignored; the subsequent rebuild stopped with Windows `os error 112` because the drive ran out of space)
- `cargo fmt -p perry-hir -p perry-codegen -p perry-runtime -p perry -- --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected typed-array behavior for array-only methods.
  * Unsupported methods now throw `TypeError` instead of incorrectly returning the typed array.
  * Own-property and prototype overrides remain callable.
  * Prevented unsupported method calls from mutating typed arrays.

* **Tests**
  * Added regression coverage for numeric and dynamically typed typed-array calls, overrides, and error handling.

* **Documentation**
  * Added a changelog entry describing the corrected behavior.

### DIFF
--- a/changelog.d/8280-typed-array-method-availability.md
+++ b/changelog.d/8280-typed-array-method-availability.md
@@ -1,0 +1,5 @@
+### fix(typedarray): keep array-only methods absent
+
+Typed arrays now throw a `TypeError` when array-only methods such as `flat`,
+`push`, or `toSpliced` are missing, instead of silently returning the receiver.
+Own and prototype overrides remain callable. Fixes #8138.
--- a/crates/perry-codegen/src/lower_call/property_get.rs
+++ b/crates/perry-codegen/src/lower_call/property_get.rs
@@ -16,7 +16,9 @@ use crate::lower_string_method::{
     is_known_string_method_name, lower_string_method, lower_string_method_from_proven_box,
 };
 use crate::rooting::{any_operand_may_collect, open_rooted_group, Repr};
-use crate::type_analysis::{is_array_expr, is_string_expr, receiver_class_name};
+use crate::type_analysis::{
+    is_array_expr, is_string_expr, is_typed_array_expr, receiver_class_name,
+};
 use crate::types::{DOUBLE, I1, I64};
 
 mod dynamic_dispatch;
@@ -33,6 +35,7 @@ mod static_dispatch;
 pub(crate) use helpers::{
     class_chain_has_field_named, is_array_only_method_name, is_date_receiver,
     is_inherited_object_prototype_method, resolve_static_dispatch_cls, string_only_method_arity_ok,
+    typed_array_lacks_array_method,
 };
 
 /// Preserve the old String-method fast path for an unproven receiver without
@@ -161,7 +164,10 @@ pub fn try_lower_property_get_method_call(
     // `trim`/`split`/`charAt` method (Zod schemas are the reported case).
     // Keep the static fast path positive-proof-only; the runtime dispatcher
     // below handles both genuine Any-typed strings and user methods.
-    if is_array_expr(ctx, object) && !is_inherited_object_prototype_method(property) {
+    if is_array_expr(ctx, object)
+        && !is_inherited_object_prototype_method(property)
+        && !(is_typed_array_expr(ctx, object) && typed_array_lacks_array_method(property))
+    {
         return Ok(Some(lower_array_method(ctx, object, property, args)?));
     }
 
--- a/crates/perry-codegen/src/lower_call/property_get/helpers.rs
+++ b/crates/perry-codegen/src/lower_call/property_get/helpers.rs
@@ -28,6 +28,14 @@ pub(crate) fn is_array_only_method_name(name: &str) -> bool {
     )
 }
 
+/// Array methods that are deliberately absent from `%TypedArray%.prototype`.
+pub(crate) fn typed_array_lacks_array_method(name: &str) -> bool {
+    matches!(
+        name,
+        "flat" | "flatMap" | "push" | "pop" | "shift" | "unshift" | "splice" | "toSpliced"
+    )
+}
+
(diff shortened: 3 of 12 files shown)


proggeramlug:
Merging. Fixes #8138.

**Verified:** `perry-runtime --lib` **2568/0**, `perry-codegen --no-fail-fast` **28 suites, 1510 passed, 9 failed — all nine in the known baseline**, and the PR's own integration test `issue_8138_typed_array_method_availability::array_only_methods_throw_on_typed_arrays_without_mutating_them` passes (630 s — it is a slow one, worth knowing before someone assumes it hung). `cargo fmt`, `check_test_registration`, `check_file_size` clean.

The shape is right: keeping array-only methods out of Array-specific lowering, while **preserving own and prototype user methods before throwing** — so a user who legitimately defines `push` on a typed-array subclass still gets their method rather than a `TypeError`. Covering all eight affected methods plus dynamic dispatch, non-mutation and overrides is the difference between "throws now" and "throws for the right reason".

PR #8281, opened 2026-08-17
perf(codegen): propagate array element shapes into callbacks

### LINKED ISSUE #8103: repsel: array-callback parameters carry no element-shape fact — 10.1% vs the equivalent for-loop (#7151 item 2)
Successor to **#7151 item 2**, split out with a measurement so it is one live
quantified item instead of one of five unranked ones. #7151 is being closed;
this is the only one of its five forms with a material number.

## The gap

`A.map(cb)` / `A.forEach(cb)` / `A.reduce(cb, …)` callback **parameters** carry
no element-shape fact, from either mechanism:

* `collectors/ptr_shape_elements.rs` (#7034 §3 / #7149) licenses only
  `const r = A[i]` inside the region that owns `A` — a callback parameter is the
  `call argument` position, which needs clone-and-route (#7034 §1), not a
  containment relaxation.
* `stmt/element_shape_loop.rs` (#7480/#7496 runtime invariant + versioned loop
  clone) matches a `for` loop over `A`, and a callback body is not one.

So every `r.field` in a callback body re-enters the guard diamond that both
mechanisms delete for the equivalent `for` loop.

## Measured on `main` @ `514ff7c9a`

macOS arm64, `--release` compiler and `--release`
`libperry_{runtime,stdlib}.a`, `PERRY_RUNTIME_DIR` pinned, `/usr/bin/time -l`
instructions retired, medians of 3 interleaved reps. Both programs build the
same 20,000-element `C[]`, read two fields off every element 200 times (4M
element reads), and print the identical `t: 80000000000`.

```ts
// baseline — the covered form
for (let i = 0; i < arr.length; i++) { const r = arr[i]; s = s + r.a + r.b; }

// probe — the same work through a callback
arr.forEach((r: C) => { s = s + r.a + r.b; });
```

| | `Ptr<Shape>` consumed | instructions retired | vs baseline |
|---|---:|---:|---:|
| `for` + `const r = A[i]` | **2** | 15,800,121,583 | — |
| `A.forEach(cb)` | **0** | 17,391,614,644 | **+10.1%** |

`--opt-report` on the `forEach` probe: `Ptr<Shape> 0 selected / 1 denied`,
`*** Ptr<Shape> promoted 0 of 1 candidates in this build. ***`

For scale, the other four forms in #7151 measure −0.3% (direct `A[i].field`,
already served by the versioned clone), +1.5% (`let r = A[i]`), ≲1% (`sort`
then read), and ~0 for the element-rule half of non-empty literals — full table
and method in https://github.com/PerryTS/perry/issues/7151#issuecomment-5298724714.

## Why it is worth its own issue

#7170's profile of `benchmarks/app-patterns/kernels/batch.ts` put the reducer's
`acc[idx]` field traffic (`js_object_set_field_by_name` 3.57% +
`js_put_value_set_dyn_ic_miss` 3.14% inclusive) in exactly this position, and
#7151 item 2 already named it: *"it is where `batch.ts`'s hot field traffic
actually lives"*.

## Two shapes, and they are not the same problem

1. **`forEach`/`map` element parameter** — one array, one class, the parameter
   is the element. This is the narrow case and the 10.1% above.
2. **`reduce` accumulator parameter** — `acc[idx]` inside the reducer, i.e. a
   *second* array reached through a parameter. Needs the array-level fact to
   cross the call boundary, not just the element fact.

(1) alone would be worth scheduling if it can be done as clone-and-route on the
callback with an all-call-sites-agree obligation, which is #6925's precedent.

## Not measured

x86-64; `--profile dist`; `map`/`reduce` separately from `forEach`; any real
dependency-JS population (#7152 measured the outbound `call argument` escape at
5 of 746 there, which is a different direction from this inbound one); and I did
not attempt an implementation, so there is no cost estimate for the
clone-and-route side.

proggeramlug:
## Summary

- carry a dense, monomorphic local array's proven element shape into a direct inline arrow callback
- reuse the existing full `Ptr<Shape>` escape, class, dispatch, numeric-field, and all-or-nothing group proofs before routing the fact into the separately compiled closure body
- keep opaque/reused callbacks, normal function expressions, async/generator callbacks, rest/`arguments`, source-array parameters, escaping elements, and element-returning HOFs on the guarded path
- add collector red tests plus emitted-LLVM regressions proving admitted field reads lose both shape-guard helpers while a denied alias case retains them

`reduce`/`reduceRight` propagate the current-element parameter only. Propagating a separate accumulator array is the distinct second shape described in the issue and is not widened here.

No version bump.

## Validation

- `cargo test -p perry-codegen collectors::ptr_shape_elements::tests:: --lib` (43 passed)
- `cargo test -p perry-codegen array_callback_shape_tests --lib` (2 passed)
- `cargo test -p perry-codegen --lib -- --skip native_emit::tests::split_native_construction_lowers_precise_roots_before_rs4gc --skip native_emit::tests::split_native_construction_propagates_shadow_backend_to_workers` (1072 passed before the final conflict-free rebase)

The unfiltered Windows suite has two unrelated existing failures in those skipped `native_emit` tests: they byte-compare COFF objects that embed different per-run temporary filenames. Both reproduce when run alone; all other 1072 tests pass.

Closes #8103

## Summary by CodeRabbit

* **Performance**
  * Improved performance for eligible inline `forEach`, `map`, `reduce`, and `reduceRight` callbacks by enabling faster access to array element fields.
  * Field reads may now use optimized fixed-offset access when element shapes are proven stable.
  * Callbacks with unsupported or potentially unsafe behavior continue using guarded access.

* **Bug Fixes**
  * Added regression coverage for callback shape propagation, accumulator handling, escaping callbacks, aliases, and other edge cases.

### DIFF
--- a/changelog.d/8103-array-callback-shape.md
+++ b/changelog.d/8103-array-callback-shape.md
@@ -1,0 +1,8 @@
+### Performance
+
+- Propagate proven local-array element shapes into direct inline arrow
+  callbacks for `forEach`, `map`, `reduce`, and `reduceRight`. Field reads on a
+  contained callback element now use fixed-offset loads instead of repeating
+  the class-shape and by-name lookup guards. Opaque, escaping, asynchronous,
+  generator, recursive, rest/`arguments`, and source-array-aliasing callbacks
+  remain on the guarded path.
--- a/crates/perry-codegen/src/codegen/closure.rs
+++ b/crates/perry-codegen/src/codegen/closure.rs
@@ -803,7 +803,7 @@ pub(super) fn compile_closure(
             .return_shape_class(func_id)
             .is_some(),
     );
-    let native_facts = crate::collectors::collect_native_region_fact_graph(
+    let mut native_facts = crate::collectors::collect_native_region_fact_graph(
         body,
         &[],
         &flat_const_ids,
@@ -817,6 +817,12 @@ pub(super) fn compile_closure(
         &cross_module.compile_time_constants,
         &cross_module.module_dispatch,
     );
+    if let Some(callback_shapes) = cross_module.array_callback_shapes.get(&func_id) {
+        native_facts
+            .shape_stability
+            .shape_proven_ptr_locals
+            .extend(callback_shapes.clone());
+    }
 
     // Representation-selection context gates (see codegen/function.rs).
     // Async-step closures (CPS-rewritten `async` closures — the rewrite clears
--- a/crates/perry-codegen/src/codegen/mod.rs
+++ b/crates/perry-codegen/src/codegen/mod.rs
@@ -1987,6 +1987,7 @@ pub fn compile_module(hir: &HirModule, opts: CompileOptions) -> Result<Vec<u8>> 
         target_triple: triple.clone(),
         app_metadata: opts.app_metadata.clone(),
         module_dispatch: module_dispatch_facts,
+        array_callback_shapes: std::collections::HashMap::new(),
         // Inline-hot-small pre-pass (#6850 follow-up): FuncIds with an in-loop
         // call site AND few total call sites, so small hot callees can earn
         // `inlinehint` while the call-site cap bounds duplication.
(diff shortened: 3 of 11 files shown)


proggeramlug:
Merging. Codegen **28 suites, 1534 passed, 9 failed — all nine baseline**; runtime **2568/0/4**; `cargo fmt` clean.

**The deny side is what I checked, and it is well covered.** A proof-granting change is only as good as the cases it refuses, because an unsound admit is a miscompile rather than a slowdown. Six named tests pin exactly that:

- `escaping_array_callback_element_denies_the_whole_group`
- `opaque_array_callback_is_not_an_element_shape_route`
- `element_returning_array_hof_is_not_a_contained_route`
- `source_array_callback_parameter_denies_the_route`
- `self_callable_function_expression_denies_the_route`
- `inline_reduce_promotes_the_element_not_the_accumulator`

That last one is the subtle case and I'm glad it's there — promoting the accumulator instead of the element in a `reduce` would be silently wrong for exactly the workloads this optimization targets.

Reusing the existing `Ptr<Shape>` escape, class, dispatch, numeric-field and all-or-nothing group proofs rather than inventing a parallel admission path is the right call — it means this rides on machinery that already has its own red tests, instead of adding a second thing that can disagree with the first.

Pairing collector red tests with emitted-LLVM regressions is also the right shape: asserting that an admitted read *loses both shape-guard helpers* proves the fast path is actually taken, not merely that nothing crashed.

PR #8282, opened 2026-08-17
fix(gc): root raw pointers across allocations in dynamic-object get, and add a PROCESS_EMITTER root scanner

jdalton:
Two valid #8220-class fixes, landed as improvements — **not** a resolution of the seeded rooting crash (seeds 8, 11, 22 still fail). Opened as the correct fixes plus a complete diagnosis for the next pass.

## The two fixes

1. **`js_dynamic_object_get_property`** (`crates/perry-runtime/src/value/dynamic_object.rs:650`) — rooted the raw `*const ObjectHeader` pointer across `js_string_from_bytes` using `RuntimeHandleScope::root_raw_const_ptr` + `across_const`, mirroring the typed-array path at line 427. The raw pointer was held across an allocation that can trigger a copying minor — a textbook #8220-class hazard.

2. **`PROCESS_EMITTER` root scanner** (`crates/perry-runtime/src/os/os_process_emitter.rs`) — added `process_emitter_root_scanner` that visits `callback` and `raw_wrapper` raw `*const ClosureHeader` pointers in the TLS `HashMap` via `visit_raw_const_ptr_slot`, registered in `gc/mod.rs` with `reg_scanner!`. Without it, a copying minor that evacuates a listener closure leaves the raw pointer stale.

## Verification

- `cargo test -p perry-runtime --lib`: 2557 passed, 0 failed.
- `cargo test -p perry-codegen --lib`: 1064 passed, 0 failed.
- No regressions: seeds 1, 5, 15, 20, 23, 24 pass as before.

## What this does NOT fix

Seeds 8, 11, 22 still crash (`TypeError: Cannot convert undefined or null to object`). A diagnostic native-stack scan (built for this, `PERRY_GC_SCAN_NATIVE_STACK=1`) surfaced 14 stale pointers after the rewrite pass — but every one was a dead value or leftover bytes from prior frames, **not the live variable causing the crash**. That definitively rules out the native stack as the stale pointer's location.

The live pointer is therefore in one of three places the scan can't see: a TLS `RefCell`/`Cell` (top candidate — `PROCESS_EMITTER` was checked and its new scanner did NOT resolve it), a register never spilled to the stack (caller-saved registers from generated code are gone by the scan point), or a heap field the from-space decoder misses (side tables / exotic object encodings). That narrowing is the real deliverable for the next pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved garbage collection reliability by detecting and handling stale references held during collection.
  * Preserved event listener callbacks and one-time listeners across garbage collection.
  * Prevented property lookups from accessing outdated object references after memory cleanup.

* **Diagnostics**
  * Added optional native-stack diagnostics to identify invalid references and provide detailed collection reports.

### DIFF
--- a/crates/perry-runtime/src/gc/copying.rs
+++ b/crates/perry-runtime/src/gc/copying.rs
@@ -1523,6 +1523,12 @@ pub(super) fn run_copied_minor_attempt(
     // the root enumeration the rewrite pass and the evacuation verifier share.
     super::fromspace_scan::run_fromspace_scan(&snapshot);
 
+    // #8220 diagnostic: scan the native (Rust) stack for stale from-space
+    // pointers — raw pointers held in Rust frame locals that the precise root
+    // map can't see and the conservative scan is disabled in production. MUST
+    // run here, same window as fromspace_scan (after rewrite, before reset).
+    super::native_stack_scan::run_native_stack_scan();
+
     crate::promise::cleanup_copied_minor_promise_contexts_for_gc();
     finalize_dead_copied_minor_from_space_side_allocations();
     // #7742: on a promoting cycle the young blocks are handed to old-gen
--- a/crates/perry-runtime/src/gc/mod.rs
+++ b/crates/perry-runtime/src/gc/mod.rs
@@ -172,6 +172,9 @@ mod verify;
 /// the rewrite pass own root enumeration. Debug-only
 /// (`PERRY_GC_FROMSPACE_SCAN=1`).
 mod fromspace_scan;
+/// #8220 diagnostic: native-stack scan for stale from-space pointers after a
+/// copying minor. Debug-only (`PERRY_GC_SCAN_NATIVE_STACK=1`).
+mod native_stack_scan;
 /// #7742: the measured policy behind whole-block in-place promotion. The
 /// mechanism is `arena/promote.rs`; this decides when to use it.
 mod promote_in_place;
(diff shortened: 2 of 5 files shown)


proggeramlug:
Landed as **#8294**, which is merged. Closing this one since the branch is on your fork and I can't push the two required fixes to it.

Your rooting work went in unmodified and is correct — the `across_const` on `js_dynamic_object_get_property` is exactly the right idiom (it expresses the ordering in one call and never binds the pre-call address), and registering a scanner for `PROCESS_EMITTER`'s TLS `*const ClosureHeader` table is the right disposition for a raw-pointer cache the precise root map cannot see. You're credited on the PR.

Two things had to change before it could merge:

**It did not compile against current `main`.** #8277 landed a few hours ago and widened `classify_heap_space_in_range` to `(HeapSpace, usize, *mut u64)`; `native_stack_scan.rs:332` destructures two. Git merges both cleanly because the lines never touch — worth knowing if you rebase other work over #8277.

**`PERRY_GC_SCAN_NATIVE_STACK_ABORT=1` alone was inert.** `run_native_stack_scan` returns at the `native_stack_scan_enabled()` gate, so with only the abort switch set nothing runs, nothing aborts, and the run reports success. That's the same defect #7154 fixed for the from-space pair — `fromspace_scan.rs:124` carries a comment specifically about it. `native_stack_scan_enabled()` now ORs in the abort flag.

I kept your framing that this **does not resolve the seeded crash** (seeds 8, 11, 22 still fail) — that honesty is the right call and it's stated in both the PR and the changelog fragment.

PR #8283, opened 2026-08-17
fix(codegen): bound failed LLVM scratch retention

### LINKED ISSUE #8249: codegen: retained-on-failure LLVM scratch is unbounded and never reaped — one failed multi-module compile can retain GB, and it compounds into the next attempt
Measured on two machines while working #8228's workaround. Filing with a correction to the obvious framing: **this is not the compiler forgetting to clean up.** `crates/perry-codegen/src/linker.rs:712-720` states the policy deliberately, and the reasoning is good:

> * **success**: the object is read into memory, and the per-call directory goes with it — a compile leaves nothing behind (#7144).
> * **failure** (clang non-zero, or the object cannot be read): everything is left on disk. The error message names the `.ll`, and a failed compile is exactly when someone wants to look at the IR that produced it.

Retention on failure is intentional and useful. The defect is that it is **unbounded in three directions at once**, and the combination is what bites.

## 1. Unbounded per compile: one dir per failing module, not one per failure

A whole-project compile that fails retains a scratch dir for **every** failing module. My failed 104-module fixture build left **124 directories totalling 4.9 GB** in `$TMPDIR` from a *single* attempt. All 105 module failures had the identical cause, so 104 of those IR copies serve no debugging purpose whatever — the policy's rationale ("someone wants to look at the IR") is fully satisfied by the first one.

This is much worse on the `PERRY_LLVM_INPROCESS=off` text path (#8228's interim workaround), which materialises 100–200 MB of IR for the oversized modules — 171.3 MB for `app-page.runtime.prod.js` on my box, 222.6 MB for one module on another.

## 2. Unbounded over time: nothing ever reaps it

Two machines, neither recently provisioned:

| host | stale `perry_llvm_scratch_*` dirs | older than 2 h | size | oldest |
|---|---|---|---|---|
| dev box | **1,180** | 1,177 | 567 MB | 2026-08-15 |
| bench mini | **4,155** | 4,155 | ~627 MB | — |

None had a live owning process. The size is modest; the **inode** pressure and the `readdir` cost on a temp dir with thousands of entries are not, and neither is the confusion of picking through them.

## 3. It compounds: attempt N's retention causes attempt N+1's failure

This is the part that turns an annoyance into a trap. Attempt N fails → retains ~5 GB → attempt N+1 hits ENOSPC *because of* attempt N → retains another ~5 GB. And per #8228 (comment 5309169441), ENOSPC does not present as a disk error: `clang -c` exits non-zero with **empty stderr**, so every module reports `clang -c failed (stderr empty)` and the printed diagnostic suggests a **triple/ABI mismatch** and points at `PERRY_LLVM_CLANG`. The user chases a toolchain theory while the real cause is a disk their own previous attempt filled.

## Suggested fixes, cheapest first

1. **Cap retention per compile.** Keep the first failing module's scratch and drop the rest, or keep at most K. The error already names the retained path, so one example is what the policy is actually for. This alone removes the GB-scale case.
2. **Reap on startup.** Remove `perry_llvm_scratch_*` older than some threshold with no live owner. Note the owner check matters: the pid is in the directory name, and on a shared box other sessions' compiles are live in the same `$TMPDIR` — there were six live `perry` compiles on mine while I was cleaning up my own, matched by the pid-derived prefix in my error message.
3. **Distinguish environmental failure from compile failure.** An empty clang stderr with a non-zero exit is not IR anyone wants to inspect. Checking free space there and saying so would fix both this and the misdirection in #8228.

`crates/perry-codegen/src/linker_temp_lifecycle_tests.rs:483` already records a prior incident of this family ("58 compiles left 58 `perry_llvm_scratch_<pid>_<counter>`"), so the lifecycle has a test suite to extend rather than one to invent.

proggeramlug:
## Summary

Bounds failed LLVM scratch retention so one broken multi-module compile cannot leave one large IR directory per module, while preserving a useful first diagnostic sample and the explicit `PERRY_LLVM_KEEP_IR` contract.

## Changes

- Atomically retain only the first automatic LLVM failure artifact per Perry process across both in-process and external-clang paths; later failures remove their private scratch directories and report that decision.
- Reap Unix scratch directories older than two hours on startup only when their PID-derived owner is gone; preserve live owners, explicit keep markers, and pre-marker compile-plan output.
- Extend lifecycle coverage for simultaneous failures, `PERRY_LLVM_KEEP_IR` bypass behavior, stale/live-owner handling, and legacy kept artifacts.

## Related issue

Fixes #8249

## Test plan

- [x] `PERRY_LLVM_INPROCESS=off cargo test -p perry-codegen --no-default-features --lib linker_temp_lifecycle_tests -- --test-threads=1` (13 passed)
- [x] Default-backend `cargo test -p perry-codegen --no-default-features linker_temp_lifecycle_tests -- --test-threads=1` (13 passed)
- [x] `scripts/check_file_size.sh`
- [x] `cargo fmt --all` and `git diff --check`
- [ ] Full affected-crate/workspace suite (targeted lifecycle suite run instead)
- [x] Added or updated a `#[test]` in the affected crate
- [ ] Docs update (not applicable: no CLI, stdlib, or runtime API change)
- [ ] Platform UI backend build (not applicable)

## Screenshots / output

Not applicable.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved LLVM compilation failure handling by retaining diagnostic artifacts from the first failure while automatically cleaning up subsequent temporary files.
  * Removed stale scratch artifacts from inactive processes after two hours, while preserving active or explicitly retained diagnostics.
* **New Features**
  * Added an option to retain LLVM intermediate representations from all failures for deeper troubleshooting.
* **Documentation**
  * Documented the updated LLVM diagnostic scratch-retention behavior.

### DIFF
--- a/changelog.d/8283-bound-llvm-scratch.md
+++ b/changelog.d/8283-bound-llvm-scratch.md
@@ -1,0 +1,7 @@
+### Fixed
+
+- Bound automatic LLVM failure retention to one diagnostic scratch directory
+  per Perry process, preventing a failed multi-module compile from retaining a
+  large IR copy for every module. Unix startup now also reaps scratch older
+  than two hours when its owning process is gone, while preserving live and
+  explicitly kept artifacts.
--- a/crates/perry-codegen/src/linker.rs
+++ b/crates/perry-codegen/src/linker.rs
@@ -16,6 +16,12 @@ use std::sync::OnceLock;
 
 use anyhow::{anyhow, bail, Context, Result};
 
+#[path = "linker_temp.rs"]
+mod linker_temp;
+use linker_temp::{
+    reap_stale_llvm_scratch_once, FailedScratch, FailureRetention, PROCESS_FAILURE_RETENTION,
+};
+
 /// Cached result of the pre-flight clang probe — evaluated once per process.
 /// `Some(default_triple)` if the probe succeeded, `None` if it failed.
 static CLANG_PROBE: OnceLock<Option<String>> = OnceLock::new();
@@ -690,14 +696,17 @@ fn compile_ll_to_object_with_native_roots(
     target_triple: Option<&str>,
     native_roots: bool,
 ) -> Result<Vec<u8>> {
+    let tmp_dir = env::temp_dir();
+    reap_stale_llvm_scratch_once(&tmp_dir);
     let rs4gc_ll = maybe_rs4gc_preprocess(ll_text, native_roots)?;
     let ll_text: &str = rs4gc_ll.as_deref().unwrap_or(ll_text);
-    compile_ll_to_object_in(
-        &env::temp_dir(),
+    compile_ll_to_object_in_with_retention(
+        &tmp_dir,
         ll_text,
         target_triple,
         TempFilePolicy::from_env(),
         native_roots,
+        &PROCESS_FAILURE_RETENTION,
     )
 }
 
@@ -713,9 +722,8 @@ fn compile_ll_to_object_with_native_roots(
 ///
 /// * **success**: the object is read into memory, and the per-call directory
 ///   goes with it — a compile leaves nothing behind (#7144).
-/// * **failure** (clang non-zero, or the object cannot be read): everything is
-///   left on disk. The error message names the `.ll`, and a failed compile is
-///   exactly when someone wants to look at the IR that produced it.
+/// * **failure** (clang non-zero, or the object cannot be read): the first
+///   failure's IR is retained for diagnosis; later failures are removed (#8249).
 /// * **`PERRY_LLVM_KEEP_IR`**: everything is kept and its location printed,
 ///   plus the compile plan as JSON.
 /// * **`PERRY_DEBUG_SYMBOLS`**: no effect on any of the above. It was believed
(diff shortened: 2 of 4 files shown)


proggeramlug:
Merging. Codegen **28 suites, 1513 passed, 9 failed — all nine baseline**; runtime **2568/0/4**; `cargo fmt` clean.

**I read the reaper carefully because this is a deletion path**, and it fails in the right direction at every step. A directory is removed only when *all three* hold: it is at least two hours old, `was_explicitly_kept` is false, and `process_is_alive(pid)` is false. And that last one is properly conservative — `kill(pid, 0) == 0` means alive, but on failure only **ESRCH** counts as gone; EPERM or any other errno returns `true` and preserves the files. PID reuse therefore biases toward keeping, never toward deleting someone's live scratch. Windows is a deliberate no-op with the reason stated.

That matters more than usual today: I spent a good part of this session with the disk at 0–4 GiB, and the failure mode I kept guarding against was exactly "an agent between builds looks identical to a finished one." A reaper that got this backwards would have been worse than the leak it fixes.

**Retaining the first failure per process is the right shape too** — a broken multi-module compile leaves one useful IR sample for diagnosis instead of one large directory per module, and later failures report that they removed theirs rather than silently vanishing. The explicit `PERRY_LLVM_KEEP_IR` contract is preserved, with `keep_ir_retains_the_whole_scratch_dir` covering it under both the plain and native-roots paths.

PR #8284, opened 2026-08-17
fix(runtime): bind static handle method names

### LINKED ISSUE #8178: js_class_method_bind: perry-stdlib's handle-property dispatch has the same interior-pointer bug in eight sub-dispatchers
## Summary

#8133 named four `js_class_method_bind` sites binding a movable heap string's interior. Fixing it (#8177) surfaced **six** in perry-runtime — and the same defect **systemically in perry-stdlib's handle-property dispatch**, where eight sub-dispatchers capture the forwarded name pointer directly.

## The contract being violated

`js_class_method_bind(instance, name_ptr, name_len)` stores the name **pointer** in the bound closure's capture, and `dispatch_bound_method` (`closure/dispatch/bound.rs:12`) re-reads it at call time. So the pointer must outlive the bind — it cannot point into a movable GC heap string, whose contents are unreachable once a collection moves it.

## Where it still happens

Eight sub-dispatchers in perry-stdlib forward the caller's pointer straight into the bind:

- `sqlite` — four sites
- `tls` — two sites
- `emitter_als` — two sites

Reachable from ordinary code that stores a method rather than calling it:

```ts
const f = db.run;        // sqlite
const g = emitter.on;    // emitter_als
const h = als.getStore;  // emitter_als
```

Also found on perry-runtime's **primitive-receiver** arm (`(5).toString`), which #8177 did not cover.

## Why this will not show up by accident

#8177 established the reproduction conditions, and they are narrow:

- A **literal** `dec.decode` lowers to rodata and never reaches these arms at all. Only a **computed** key does — `const k = "dec" + "ode"; obj[k]`.
- Pre-fix, the symptom was a *silent wrong answer*: `decode=undefined` followed by a throw, where node prints `decode=hi`. No instrument fired.
- Under `PERRY_GC_ZEAL=1 PERRY_GC_PROTECT_FROMSPACE=1` it became a SIGBUS reported as `RETIRED FROM-SPACE … retired_by_minor=#0`.

So a test that binds a literal name and calls it immediately will pass on broken code. Any fixture here needs a computed key **and** enough allocation to actually move the string; #8177's used ~400k allocations.

## Suggested shape

#8177's fix is worth copying rather than re-deriving: it replaced the `is_timer_handle_method_key` *predicate* with a `'static`-returning lookup and removed `key_ptr`/`key_len` from `text_handle_property` entirely. With no predicate and no pointer parameter, the bug is **unwritable** rather than merely fixed. The stdlib dispatchers can take the same treatment.

## Related

#8133 (closed by #8177), #7747 (the first two sites), `docs/src/internals/gc-rooting-invariant.md`.

proggeramlug:
Closes #8178.

## Summary

`js_class_method_bind` retains its method-name pointer in the bound closure. The stdlib handle-property dispatchers were forwarding slices backed by the caller's movable GC heap string, so computed method reads could retain an interior pointer after the string moved or died.

- remap SQLite, TLS, EventEmitter, and AsyncLocalStorage method names to `&'static [u8]` literals before binding
- make the shared stdlib bind helpers accept only static names, so the unsafe call cannot consume a forwarded property slice
- replace the primitive-receiver predicate with a static lookup and route both finite-number guards through one safe-name binder
- add structural pointer-identity tests plus complete method-list coverage for every lookup
- intentionally leave workspace/package versions unchanged

## Validation

- `cargo test -p perry-runtime --lib handle_bound_method_name -- --nocapture` (7 passed)
- `cargo test -p perry-stdlib --lib static_method_name -- --nocapture` (4 passed)
- `RUSTFLAGS="-D warnings" cargo check -p perry-runtime -p perry-stdlib --all-targets`
- `python3 scripts/gc_store_site_inventory.py --self-test && python3 scripts/gc_store_site_inventory.py`
- `python3 scripts/addr_class_inventory.py --self-test && python3 scripts/addr_class_inventory.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed method access for primitive values and runtime handles so bound methods remain available after garbage collection.
  * Improved method dispatch across event handling, asynchronous storage, database, TLS, and socket APIs.
  * Unknown properties are now rejected more reliably instead of being treated as supported methods.
* **Tests**
  * Added regression coverage for primitive prototype methods, recognized API methods, and unknown property names.

### DIFF
--- a/changelog.d/8284-static-handle-method-names.md
+++ b/changelog.d/8284-static-handle-method-names.md
@@ -1,0 +1,7 @@
+### fix(runtime): keep bound handle method names alive across GC
+
+Computed method-value reads on node:sqlite, TLS, EventEmitter, and
+AsyncLocalStorage handles now bind static method-name literals instead of
+retaining pointers into movable GC heap strings. Primitive-number method reads
+use the same static-name rule, so closures such as `(5).toString` cannot retain
+the computed key's interior after collection. Fixes #8178.
--- a/crates/perry-runtime/src/gc/tests/handle_bound_method_name.rs
+++ b/crates/perry-runtime/src/gc/tests/handle_bound_method_name.rs
@@ -1,5 +1,5 @@
-//! A bound TIMER-handle / TextDecoder / TextEncoder method closure must not
-//! capture a pointer into the key string.
+//! A bound TIMER-handle / TextDecoder / TextEncoder / primitive-receiver method
+//! closure must not capture a pointer into the key string.
 //!
 //! `js_class_method_bind` stores the method-name POINTER in the closure
 //! (capture 1) and `dispatch_bound_method` re-reads it at CALL time. Its
@@ -34,6 +34,9 @@
 //!   entry point, not a redundant copy.
 //! * `get_field_by_name.rs` — a third copy of the same timer block. Fixed
 //!   defensively; nothing guarantees it stays unreachable across refactors.
+//! * `get_field_by_name.rs`'s primitive-number receiver — `(5).toString` and
+//!   the inherited Object prototype methods used the computed key's movable
+//!   interior directly (#8178).
 //!
 //! ## Why these assert IDENTITY and not bytes
 //!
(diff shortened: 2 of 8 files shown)


proggeramlug:
Merging. Runtime **2569 passed / 0 failed / 4 ignored**, codegen **28 suites, 1510 passed, 9 failed — all nine baseline**, `cargo fmt` clean, and both named test groups pass: `handle_bound_method_name` 7/7 and `static_method_name` 4/4.

**The fix is type-level, which is the right kind for this bug.** `js_class_method_bind` retains its name pointer in the bound closure, so a slice backed by a movable GC heap string is a borrowed-heap-slice hazard — the class where rooting cannot help, because the problem is that the borrow outlives the object rather than that the object was collected. Making the shared bind helpers accept only `&'static [u8]` means the unsafe call *cannot* consume a forwarded property slice. That makes the bug unrepresentable at those sites rather than merely absent, and it matches the convention already established in `domain.rs`, `crypto/*` and `string_decoder.rs` — this brings sqlite, tls and emitter_als into line.

One note on your validation list: `RUSTFLAGS="-D warnings" cargo check -p perry-runtime -p perry-stdlib --all-targets` currently fails, but **not because of this PR**. It was #8294's dead code (`register_process_emitter_root_scanner` never called, plus unused frame-walk locals). I merged that one earlier today and missed it. Fixed separately in #8298, which also turned out to matter for more than warnings — the scanner really was never registered, so that fix had been inert. Nothing for you to do here.

Issue #6827, opened 2026-07-24
RFC: Stabilize a Perry native value profile for fixed-width scalars, POD records, and native boundaries


rink3y:
## The problem

Perry already has two strong halves of a native-value story, but they are not
yet one stable user-facing contract.

On the application side, Perry intentionally accepts ordinary TypeScript:
`number`, objects, arrays, interfaces, classes, npm modules, and Node-compatible
APIs. That compatibility should remain the default.

On the compiler and native-library side, Perry already models explicit native
representations:

- the native-representation pipeline has `I32`, `I64`, `U32`, `U64`, `USize`,
  `F32`, `F64`, `BufferLen`, native handles, POD records, and verified
  materialization boundaries;
- the native-library manifest accepts typed ABI descriptors and C-layout POD
  records;
- generated declarations currently expose internal-looking types such as
  `PerryI32`, `PerryU64`, `PerryF32`, and `PerryPod<T>`;
- `NativeArena`, typed views, `sizeof<T>()`, `alignof<T>()`, and
  `offsetof<T>()` already exist; and
- #1849 has landed most of the compiler foundation for explicit native
  representations and verifier-backed native regions.

The missing piece is a documented source-level contract that an application or
library author can intentionally use.

Today, a developer can see that Perry internally understands exact widths and
POD layouts, but there is no stable answer to questions such as:

1. How do I declare that a value is a `u16`, `i32`, `u64`, or `f32`, rather
   than an ordinary JavaScript `number`?
2. Which conversions are checked, lossy, wrapping, or rejected?
3. Can an exact-width value remain native across local variables, calls,
   arrays, records, WebAssembly, and native-library boundaries?
4. How do I define one C-layout record in TypeScript and have Perry verify the
   same layout against a native-library manifest?
5. What happens when `i64` or `u64` crosses a JavaScript-compatible boundary
   where a `number` cannot represent it exactly?
6. Which parts are stable language/API guarantees and which parts are compiler
   optimizations that may safely fall back to boxed JavaScript values?

Without that contract, native packages either duplicate representation
information in a manifest, pass data through generic `number`/object values, or
drop down to typed arrays and manual offsets. The compiler can optimize some of
those cases, but the author cannot state and rely on an exact native contract.

A concrete example I would like Perry to support is:

```typescript
import {
  type u16,
  type pod,
  u32,
  u64,
  f32,
  sizeof,
  alignof,
  offsetof,
} from "perry/native";

type PacketHeader = pod<{
[...]

PR #8285, opened 2026-08-17
feat(native): add checked scalar conversions

proggeramlug:
## Summary

Add the checked scalar-conversion slice proposed by the stable `perry/native` value profile. The new value-side `i32`, `i64`, `u32`, `u64`, `usize`, `f32`, and `f64` functions reject values that cannot be represented without changing their meaning.

## Changes

- expose same-name value conversions alongside the existing native scalar type aliases
- lower named, aliased, and namespace imports through native dispatch
- validate finite/integral/range/precision constraints with catchable `TypeError` and `RangeError` failures
- document the conversion contract and regenerate the public API reference
- add runtime, HIR, manifest, declaration-generation, and executable regression coverage

## Related issue

Refs #6827

## Test plan

- [x] `./scripts/test_affected_crates.sh --base origin/main`
- [x] `cargo test -p perry-runtime native_value_profile --lib -- --nocapture`
- [x] `cargo test -p perry-hir --test native_arena native_scalar -- --nocapture`
- [x] `cargo test -p perry-codegen --test manifest_consistency`
- [x] `PERRY=target/perry-dev/perry bash tests/test_native_value_profile.sh`
- [x] Added runtime, compiler, declaration, and end-to-end tests
- [x] Updated `docs/src/` and generated API documentation

## Screenshots / output

The executable regression prints `PASS` after checking POD layout, bounds, precision, float rounding, and catchable conversion errors.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added native numeric conversion functions for signed, unsigned, and floating-point types.
  - Conversions validate finiteness, integer requirements, ranges, safe-integer limits, and precision; invalid values now produce clear errors.
  - Added explicit `f32` rounding to its representable value.

- **Documentation**
  - Updated API references, type declarations, and native value guidance with conversion behavior and examples.

- **Tests**
  - Added coverage for boundary values, rounding, invalid inputs, namespace usage, and runtime output.


proggeramlug:
Merging. Runtime **2571 passed / 0 failed / 4 ignored**, codegen **28 suites, 1524 passed, 9 failed — all nine baseline**, `perry-hir --test native_arena` **20/0**, `cargo fmt` clean, `bash -n` on the new test script clean.

**Rejecting values that cannot be represented without changing their meaning is the right contract** for a checked-conversion slice — the failure mode a silent `as` cast produces (a wrapped `u32`, a `f64` that lost precision on the way to `f32`) is exactly the kind of bug that surfaces far from its cause. Catchable `TypeError` / `RangeError` rather than a panic keeps it usable from JS.

**One thing I fixed during the merge, and it was not yours.** `api-docs-drift` failed here, but the drift was `perf_hooks.default` — **#8279**'s manifest entry, merged earlier today without regenerating `docs/api/perry.d.ts`. Main was drifted, not this branch. I regenerated on main separately, then re-merged here; the docs conflicted (both sides had regenerated) and I resolved by re-running `scripts/regen_api_docs.sh`, which is deterministic from the manifest and therefore the authoritative resolution. Your own doc updates were correct and complete — the only delta was the entry main owed.

That is the third gate I broke today by merging on a partial check (`gc_runtime_root_holders` after #8270, `-D warnings` after #8294, and this). Recording it here because the pattern is mine, not this PR's.

PR #8287, opened 2026-08-17
fix(runtime): un-gate the value-held builtin construct arm from global-webfetch (#8223)

### LINKED ISSUE #8223: gap-suite full mode: builtin-construct trio fails under auto-optimize but passes fast mode (blocks full-suite-gate)
In the tiered CI (#8187) the gap suite runs in two modes against the same `test-parity/gap_snapshot.json`: `fast` (PERRY_SKIP_BUILD=1, prebuilt archives — PR/sweep tiers) and the harness default auto-optimize (full tier, 8 shards). First full-tier run (31935729773):

- shard 2: `test_gap_dynamic_builtin_construct_dispatch` pass -> parity_fail
- shard 3: `test_gap_new_globalthis_builtin_6726` pass -> parity_fail
- shard 7: `test_gap_builtin_alias_construct_7524` pass -> parity_fail

All three PASS in fast mode on the same commit (PR run 31940422482, all 6 shards green). So this is an **auto-optimize-link-specific** regression in the builtin construct-dispatch family — the same trio #8117 flagged (closed when #8198 fixed the snapshot side; the auto-optimize arm is still red). Suspects from the 08-15/16 window per #8117: #8183, #8165, #8094.

Until fixed this keeps `full-suite-gate` (the release gate) red.

proggeramlug:
Fixes #8223 — the last blocker that reds two gates at once (gap-suite full mode AND parity, where the trio was deliberately left un-parked by #8273).

## Root cause

The `construct.rs` dispatch arm for builtins reached through a **value** (alias variable, intrinsic lookup, cross-module re-export) — covering `Map`, `Set`, `WeakMap`, `WeakSet`, `WeakRef`, `EventTarget`, `AbortController`, `TextEncoder`, `URLSearchParams`, `DisposableStack` — carried a `#[cfg(feature = "global-webfetch")]` inherited positionally from #7008's web-platform size gating when #7779 extracted the arms into `builtin_alias_construct.rs`. The factories behind the arm are all unconditional modules; the gate was never intentional for these names.

Auto-optimize rebuilds the runtime with a computed minimal feature set — a test importing only `util` gets `features=async-runtime`, no `global-webfetch` — so the **whole arm compiled out**: `new (Map-as-value)()` fell to the call path (`TypeError: Constructor Map requires 'new'`), an aliased `EventTarget` instance had no surface. The prebuilt full stdlib (`PERRY_SKIP_BUILD` fast mode) masked it, which is exactly the fast/auto-optimize mode divergence #8223 documents — and the reason the tiered CI keeps the 8-shard auto-optimize arm in the full tier.

**Timeline**: dormant since 08-11 (#7779), first observed 08-15/16 — not because anything regressed then, but because every earlier gap-suite run was cancelled in the queue saturation. The feature-detect (`uses_global_webfetch`) is unchanged since 08-12; there is no second bug.

## Fix

Remove the cfg (one line + an explanatory comment). No behavior change for full-feature builds — the arm was always present there.

## Validation

A/B in the failing configuration (same release compiler, auto-opt archive cache wiped between arms, stripped runtime rebuilt from source each side):

| witness | pre-fix (auto-opt) | post-fix (auto-opt) |
|---|---|---|
| `test_gap_dynamic_builtin_construct_dispatch` | DIVERGE, dies mid-run (`Constructor Map requires 'new'`) | **MATCH, rc=0** |
| `test_gap_new_globalthis_builtin_6726` | DIVERGE, dies mid-run (`Constructor Set requires 'new'`) | **MATCH, rc=0** |
| `test_gap_builtin_alias_construct_7524` | `EventTarget: undefined` | **MATCH, rc=0** |

Plus `cargo check -p perry-stdlib --no-default-features --features async-runtime` clean (the minimal set the failing compile used).

After this merges, the parity gate's only intentional reds are gone from the trio and `full-suite-gate`'s remaining blockers are #8222, #8226, #8227, the packet's content regressions (#8225 items 2–3), and the doc-tests-windows hang.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored access to several built-in constructors, including `Map`, `Set`, `EventTarget`, `AbortController`, `TextEncoder`, `URLSearchParams`, and `DisposableStack`, in minimal-feature builds.
  * Ensured consistent constructor behavior across optimized and minimal configurations.

* **Documentation**
  * Added a changelog entry describing the restored constructor availability and validation results.


proggeramlug:
Merging. Runtime suite **2568 passed / 0 failed**, `cargo fmt` clean, merges into current `main` with no conflicts.

**This is a satisfying root cause.** A `#[cfg(feature = "global-webfetch")]` inherited *positionally* when #7779 moved the arms out of #7008's web-platform gating — so an arm dispatching `Map`, `Set`, `WeakMap`, `WeakSet`, `WeakRef`, `EventTarget`, `AbortController`, `TextEncoder`, `URLSearchParams` and `DisposableStack` was gated behind a feature none of them need. The comment explaining why it is *not* gated is the right thing to leave behind; the next person to tidy that file would otherwise re-add it.

**And it explains the fast/full divergence rather than papering over it.** Auto-optimize builds the runtime with a minimal feature set, so the arm compiled out and every value-held builtin constructor fell through; the prebuilt full stdlib in `PERRY_SKIP_BUILD` fast mode masked it. That is precisely the mode gap #8223 documents, which is why the trio could pass the gap snapshot and fail parity at the same time.

Worth noting what this unblocks: #8273 **deliberately refused to park** these three in `known_failures.json` while `gap_snapshot.json` asserted they pass in fast mode, on the grounds that the divergence *was* the bug. That call is vindicated — the fix landed instead of the failure being absorbed. Two gates go green together.

One fix pushed: the changelog fragment was named `8280-alias-construct-featurestrip.md`; #8280 is a different open PR. Renamed to `8287-`.

PR #8290, opened 2026-08-17
perf(codegen): re-enter guarded specialization clones

### LINKED ISSUE #8169: A guarded $spec_b clone never re-enters itself either — the Tier-B half of #8167

proggeramlug:
## Summary

- let constructively numeric derived arguments satisfy Tier-B guarded call routing
- make recursive `f(n - 1)` edges re-enter the `$spec_b` clone directly instead of re-running the public parameter guard
- keep BigInt-capable arithmetic on the public guarded fallback, with positive and negative IR regression coverage

Fixes #8169.

## Validation

- `cargo test -p perry-codegen spec_self_recursion_tests --lib` (5 passed)
- `cargo test -p perry-codegen --lib --no-fail-fast` (1,067 passed; the two existing Windows temporary-object-name byte-identity tests failed: `split_native_construction_lowers_precise_roots_before_rs4gc`, `split_native_construction_propagates_shadow_backend_to_workers`)
- workspace architecture, public baseline freshness, local-binding proof, GC store-site, address-classification, and gap-snapshot quick gates
- `git diff --check`

No version bump. The required changelog fragment is included.

## Summary by CodeRabbit

- **Performance**
  - Improved recursive numeric computations by enabling optimized execution when values are proven to be standard numbers.
  - Preserved safe fallback behavior for BigInt-capable or otherwise uncertain calculations.

- **Bug Fixes**
  - Improved handling of derived numeric arguments in guarded recursive calls, ensuring they use the appropriate optimized or fallback path.


proggeramlug:
Merging. Fixes #8169 — the wrapper round-trip named as one of the two remaining routes in #8202.

**Two of your own tests failed against current `main`, and it was not your bug.** `bigint_capable_recursive_argument_keeps_the_public_guard` and `derived_recursive_number_argument_re_enters_the_guarded_clone` both went red once merged. The cause is **#8203**, which landed a few hours ago and gives recursion-participating clones LLVM's `preserve_none` convention. That word lands between `define internal` / `call` and the return type:

```
define internal preserve_nonecc double @…__f$spec_b(double %arg10)
call preserve_nonecc double @…__f$spec_b(
```

The matchers pinned a fixed prefix (`starts_with("define internal double @…")`, `matches("call double @…").count()`), so they stopped matching text that had only gained a calling convention. The subject was intact throughout — the clone existed, was specialized, and both recursive edges re-entered it. I widened the matchers to assert the parts that carry meaning (boxed clone with a `double` parameter; call *lines* naming the clone) and left the counts at 2 and 0.

Worth noting for the next IR-text test: a `starts_with` on an LLVM `define` line is fragile against any attribute or convention change, and the failure reads like a functional regression when it isn't. Matching the line and the symbol survives that.

**Verified after the fix:** `spec_self_recursion_tests` **5/5**; `perry-codegen --no-fail-fast` **28 suites, 1537 passed, 9 failed — all nine baseline, zero new**; `perry-runtime --lib` **2572/0/4**; **all 48 lint gates** via `scripts/run_lint_gates.sh`.

Letting constructively numeric derived arguments satisfy Tier-B routing while keeping BigInt-capable arithmetic on the public guarded fallback is the right split — and having both a positive and a negative IR regression for it is what makes that split checkable rather than asserted.

PR #8291, opened 2026-08-17
fix(lint): convert #8269's bare handle reads to with_* forms — unbreaks pr-gate on main

proggeramlug:
Unbreaks the required `lint` context on `main` — red since #8269 merged this morning, which reds `pr-gate` on every rebased PR (found via #8244's run; verified on a clean `main` checkout).

**Cause**: #8269's `native_module_enum` is correctly rooted (`RuntimeHandleScope`, per-iteration scopes) but reads its handles through bare `get_raw_*_ptr` — 9 new sites, pushing `object/field_get_set/enumeration.rs` to 14 against its per-module ceiling of 5 in the raw-handle debt ratchet.

**Fix**: convert all nine to the #7341 blessed `with_{mut,const}_ptr` argument-position forms. Behaviorally identical by definition (`with_mut_ptr(f)` = `f(self.get_raw_mut_ptr())` — a scoped re-read from the handle at each call site); no rooting or ordering change. Also ratchets the global baseline **down** 990 → 983 per the audit's own "debt fell, lock it in" prompt — locking in unrelated drops in `builtins/globals.rs` (15→12) and `global_this_webassembly.rs` (9→5) that had landed without an update.

**Validation**: `raw_handle_debt.py` green (983/983, 109 modules within ceilings); `cargo check -p perry-runtime` clean; full `cargo test -p perry-runtime --lib` (RUST_TEST_THREADS=1) result posted below when it completes; `cargo fmt` applied.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved internal handle access during enumeration while preserving existing values and entries behavior.
* **Chores**
  * Reduced tracked raw-handle debt and tightened permitted debt limits.
* **Documentation**
  * Added a changelog entry documenting the handle-safety improvements.


proggeramlug:
Full `cargo test -p perry-runtime --lib` (RUST_TEST_THREADS=1) on the branch: **2568 passed / 0 failed / 4 ignored** (69 s). Validation complete.

proggeramlug:
Merging immediately — this unbreaks the required `lint` context on `main`, and the breakage is mine.

**Confirmed on a clean `main` checkout:** `raw_handle_debt` reports **992 bare reads against a baseline of 990**. With this PR: **983 against 983**. Nine sites converted, baseline re-pinned to lock the improvement in.

The cause is exactly as you diagnosed: **#8269**, which I merged earlier today. Its `native_module_enum` *is* correctly rooted — `RuntimeHandleScope`, per-iteration scopes, re-reads after every collection point — but it reaches its handles through bare `get_raw_*_ptr`, which is what the #7341 ratchet counts. Nine new sites took `object/field_get_set/enumeration.rs` to 14 against a per-module ceiling of 5. I ran the census, thread-locals, addr-class and file-size gates on #8269 and did not run this one.

That is the fourth gate I have broken today the same way — `gc_runtime_root_holders` after #8270, `-D warnings` after #8294, `api-docs-drift` after #8279, and now this. Same pattern each time: I ran the subset of gates that looked topically relevant instead of the whole set. The correction is mechanical and I have been applying it since; recording the count here so it is visible rather than spread across four PRs.

**Converting to `with_{mut,const}_ptr` is the right fix rather than raising the ceiling.** The argument-position forms never bind the pre-call address, so the ordering is expressed in the call itself — that is the property the ratchet exists to push toward, and re-pinning the baseline downward is what turns the fix into a ratchet rather than a reprieve.

Verified here: runtime **2571/0/4**, `raw_handle_debt` **983/983**, plus `unrooted_local_shape --check`, `gc_runtime_root_holders`, `check_thread_locals`, `addr_class_inventory` and `cargo fmt` all clean.

PR #8292, opened 2026-08-17
fix(runtime): keep builtin alias dispatch in optimized builds

### LINKED ISSUE #8223: gap-suite full mode: builtin-construct trio fails under auto-optimize but passes fast mode (blocks full-suite-gate)

proggeramlug:
## Summary

Fix builtin constructor aliases in auto-optimized builds by removing their accidental dependency on the unrelated `global-webfetch` feature.

## Changes

- Keep the shared builtin-alias construct dispatcher available in feature-stripped runtimes.
- Restore the `global-webfetch` guard to the `Headers` arm it originally gated.
- Document why alias dispatch must remain independent of optional globalThis member-table features.

## Related issue

Fixes #8223

## Test plan

Each reported fixture was compiled with auto-optimize from this worktree, executed, and byte-compared with `node --experimental-strip-types`:

- [x] `test_gap_builtin_alias_construct_7524` (`global-url` + `global-text`, no `global-webfetch`)
- [x] `test_gap_dynamic_builtin_construct_dispatch` (`global-text`, no `global-webfetch`)
- [x] `test_gap_new_globalthis_builtin_6726` (no optional global namespace features)
- [ ] (if user-facing) Added or updated a test under `test-files/` or a `#[test]` in the affected crate (existing gap fixtures cover the regression)
- [ ] (if CLI / stdlib / runtime API changed) Updated `docs/src/` (not an API change)

## Screenshots / output

All three targeted runs completed with `node_status=0 compile_status=0 perry_status=0 diff_status=0`.

## Checklist

### DIFF
--- a/changelog.d/8292-auto-optimize-builtin-alias.md
+++ b/changelog.d/8292-auto-optimize-builtin-alias.md
@@ -1,0 +1,4 @@
+Auto-optimized builds now retain builtin constructor dispatch independently of
+the optional fetch globals. Constructing collections and other builtins through
+aliases or `globalThis` therefore matches the full prebuilt runtime, while the
+`Headers` constructor remains correctly gated behind `global-webfetch`.
--- a/crates/perry-runtime/src/object/class_registry/construct.rs
+++ b/crates/perry-runtime/src/object/class_registry/construct.rs
@@ -537,12 +537,15 @@ pub unsafe extern "C" fn js_new_function_construct(
                     options,
                 );
             }
-            #[cfg(feature = "global-webfetch")]
             // Global builtins reached through a VALUE (alias variable,
             // intrinsic lookup, cross-module re-export) rather than by name.
+            // This arm must stay independent of the optional globalThis member-
+            // table features: it also owns always-available collection aliases,
+            // while URL/Text aliases have their own feature selection.
             n if builtin_alias_construct::handles(n) => {
                 return builtin_alias_construct::construct(n, args);
             }
+            #[cfg(feature = "global-webfetch")]
             "Headers" => {
                 let init = args
                     .first()


proggeramlug:
Superseded by **#8287**, which merged earlier today and makes the identical change to the identical arm.

Both remove `#[cfg(feature = "global-webfetch")]` from the value-held builtin-construct arm — the one guarding `n if builtin_alias_construct::handles(n)`. Main now reads:

```rust
// #8223: NOT feature-gated. This arm carried a
// `#[cfg(feature = "global-webfetch")]` inherited from #7008's
// web-platform size gating when #7779 moved the arms out — but it
// dispatches Map/Set/WeakMap/WeakSet/WeakRef/EventTarget/
// AbortController/TextEncoder/URLSearchParams/DisposableStack,
// whose factories are all unconditional modules. […]
n if builtin_alias_construct::handles(n) => {
```

so the fix is in and #8223 is closed. Your diagnosis was right and matched #8287's independently — the auto-optimize build takes a minimal feature set, the arm compiled out, and value-held constructors fell through.

The only delta left here is comment wording, and the merged version carries more of the causal history (which feature gate, inherited from where, why the prebuilt full stdlib masked it), so there's nothing worth cherry-picking. Closing rather than resolving the conflict.

Worth noting for anyone reading later: two independent PRs converging on the same one-line fix is a decent signal the root cause was correctly identified.

PR #8293, opened 2026-08-17
perf(codegen, runtime): clone non-zero-start object writes

proggeramlug:
## Summary

Admits constant non-zero inner starts to the guarded object-array numeric-write clone. This is one of the remaining measured loop-shape gaps in #6812: the committed matrix was still taking the per-store PIC path for `loop_nonzero_start`.

The new runtime preflight proves exactly the active receiver range, while codegen preserves absolute array indexes, the source start in finite-range analysis, and the source start in the fast-loop phi. The established zero-start ABI and generated path are unchanged.

## Changes

- add an internal `js_object_array_numeric_write_range_guard` that validates every receiver in `[start, end)` and returns the existing packed slot-lane result
- use that guard only for fixed, non-negative, non-zero starts; retain semantic fallback for runtime-length bounds and the zero-prefix-only key-table lane
- keep the semantic priming peel and untouched slow clone, so a failed proof occurs before any raw store
- add codegen IR checks, runtime guard checks, and Node/Perry parity coverage for both the successful suffix proof and a mixed-shape rejection
- add a changelog fragment; no version metadata is changed

## Related issue

Progress on #6812 (the umbrella still has other independently measured residues).

## Performance evidence

On the issue's committed matrix harness, 15 alternating Node → baseline Perry → final Perry triples produced identical `writes=192000000` and `sink=191995200` on every run:

| `loop_nonzero_start` | Median |
|---|---:|
| Node 26.5.1 | 143 ms |
| baseline Perry | 830 ms |
| final Perry | 127 ms |

That is 6.54x faster than baseline and about 11% faster than Node. The host was heavily contended, so the alternating protocol and ratio are more meaningful than the absolute times. `PERRY_TRACE_OBJECT_ARRAY_WRITE_GUARD=1` confirms the final executable takes the new guarded clone. The complete matrix executable remained 6,312,680 bytes before and after.

## Test plan

- [x] `git diff --check`
- [x] `cargo test -p perry-codegen --test native_proof_regressions` — 271 passed
- [x] `RUST_TEST_THREADS=1 cargo test -p perry-runtime --lib` — 2,567 passed, 4 ignored
- [x] post-rebase focused codegen and runtime tests on current `origin/main`
- [x] release compiler plus runtime/stdlib archives build and final AOT link
- [x] Node/Perry byte-for-byte parity for `test_gap_6812_object_write_loop_generalization.ts` (301 bytes each)
- [x] GC store inventory self-test and full inventory (1,486 files; 295 audited; 94 allowlisted)
- [x] test registration and GC-tag coverage checks
- [x] user-facing parity case added under `test-files/`
- [ ] full excluded-platform workspace suite (not run)

`shadow_slot_hygiene` has one existing failure (`canonical_str_local_keeps_shadow_binding_and_tag_dispatched_ops`); the exact failure reproduces on pristine `origin/main` at `14468dcbc`. The emitted IR does contain the heap-string length dispatch, under the newer `pget.strlen_heap` label.

## Screenshots / output

N/A

## Checklist


proggeramlug:
Merging. `perry-codegen --no-fail-fast` **28 suites, 1538 passed, 9 failed — all baseline, zero new**; `perry-runtime --lib` **2578/0/4**; **all 48 lint gates**.

**The guard is fail-closed and the encoding is sound.** `js_object_array_numeric_write_range_guard` returns `0` on any doubt — bad `field_count`, or `object_array_numeric_write_slots` declining the range — and packs proven slots with a `+1` bias so `0` stays reserved for "not proven". A guard that admits a write it cannot prove is a memory-safety bug rather than a slow path, so reserving the failure value rather than overloading a valid slot index is the right call.

**Keeping the old `#6811` symbol is the detail I'd most likely have missed.** The comment says it plainly:

> Preserve the #6811 internal ABI for cached generated objects. New codegen uses `js_object_array_numeric_write_guard`, but an object cache entry produced before the runtime rebuild may still reference this symbol.

A stale object-cache entry linking against a symbol the new runtime dropped is exactly the kind of failure that shows up as an unresolved-symbol link error on someone else's machine, hours after the change, with no obvious connection to this PR. Retaining the old entry point costs nothing and removes that class entirely.

Closing one of #6812's remaining measured loop-shape gaps — `loop_nonzero_start` was still taking the per-store PIC path — while leaving the zero-start ABI and generated path untouched is the right scope. Proving `receiver_start..receiver_end` exactly, rather than widening the existing guard's contract, keeps the two paths independently checkable.

PR #8295, opened 2026-08-17
fix: restore native-region-proof gates

### LINKED ISSUE #8261: compiler-output-regression: 5 native-region-proof workloads fail their gate assertions (blocks full tier)
`compiler-output-regression` is red in the full tier, so it blocks the release (the full tier no longer carries `continue-on-error`).

### Repro

```bash
cargo build -p perry
python3 scripts/compiler_output_regression.py suite \
  --suite native-region-proof \
  --perry target/debug/perry \
  --benchmark-mode smoke --runs 1 --perf-counters off \
  --gate --print-summary
```

### Current state: 6 of 11 pass, 5 fail

Failing:

- `h1_native_rep_equivalence`
- `image_convolution`
- `loop_data_dependent`
- `numeric_arrays`
- `packed_f64_loop_versioning`

Passing: `h1_buffer_alias_negative`, `packed_f64_loop_versioning_negative`, `dynamic_fractional_array_index`, `loop_bound_semantics`, `raw_numeric_object_fields`, `scalar_replacement_literals`.

### These are real gate findings, not harness errors

Each failing workload reports `"errors": []` — capture succeeds and the workload is evaluated; it fails its **gate assertions** about the emitted code. Per-workload detail is in the run's `compiler-output-regression` artifact (`structural-report.json` per workload, under `target/compiler-output-regression/native-region-proof/<workload>/`).

### Why this only became visible now

Until #8097 the gate never got as far as evaluating anything: **all 11** workloads died in capture with

```
"errors": ["PERRY_LLVM_KEEP_IR did not report a retained object path"]
```

because the in-process statepoint arm retained the object but never announced it, and then with `FileNotFoundError: '(in-process)'` because the harness executed the compile plan's placeholder `clang_path` as a program. Both are fixed, so the gate now measures its subject — and 5 workloads have genuine findings that were previously unreachable.

So these 5 are **not** new regressions from this week; they are pre-existing output regressions the gate could not report while it was broken. Someone should expect real codegen work here, not a harness repair.

### Suggested triage

Start with `numeric_arrays` and `loop_data_dependent` — they are the smallest and their assertions are about numeric-region proofs, so they most directly name what changed. `h1_native_rep_equivalence` is the one to do last: it compares two representations, so a finding there may be a consequence of the others rather than an independent bug.

proggeramlug:
Fixes #8261.

## Summary

- keep proven Buffer/Uint8Array read-modify-write loops on native byte loads and stores when the RHS is call-free
- distinguish dynamic arithmetic helpers from dynamic property helpers and refresh the native-region workload contracts for current optimized IR
- make unfinished budgeted-GC cleanup safe across independent TLS destructor ordering so the image benchmark exits cleanly after printing its checksum

## Validation

- `cargo build -p perry`
- `cargo test -p perry-codegen --test native_proof_buffer_views bounded_buffer_read_modify_write_keeps_inline_byte_store --no-default-features -- --exact`
- `python3 -m unittest tests.test_compiler_output_regression` (72 tests)
- `PERRY_GC_TRACE=1 python3 scripts/compiler_output_regression.py suite --suite native-region-proof --perry target/debug/perry --benchmark-mode smoke --runs 1 --perf-counters off --gate --print-summary` (11/11 workloads pass)
- `git diff --check`

No version files are changed.

## Summary by CodeRabbit

- **Performance**
  - Improved optimized byte-buffer access for proven-safe operations, including read-modify-write loops.
  - Reduced unnecessary runtime helper calls and streamlined numeric-array reads and writes.

- **Reliability**
  - Improved cleanup during application shutdown, reducing risks from thread teardown and unfinished garbage-collection work.
  - Strengthened fallback handling for disposed, stale, aliased, or reassigned typed-array views.

- **Bug Fixes**
  - Corrected classification of dynamic arithmetic operations.
  - Updated compiler output handling for revised control-flow and loop representations.

- **Tests**
  - Added coverage for optimized buffer operations, runtime fallbacks, numeric-array behavior, and dynamic arithmetic classification.


proggeramlug:
Merging. Codegen **28 suites, 1535 passed, 9 failed — all nine baseline**; runtime **2571/0/4**; `raw_handle_debt`, `unrooted_local_shape --check`, `gc_runtime_root_holders`, `check_thread_locals`, `addr_class_inventory`, `cargo fmt` all clean.

**One thing I had to fix: `check_file_size` failed.** `crates/perry-codegen/tests/native_proof_buffer_views.rs` was at 1990 lines on `main` and this PR took it to **2036**, 36 over the cap. I split the four typed-array runtime-fallback tests into `native_proof_buffer_views/runtime_fallback.rs`, following the `pointer_lifetime.rs` pattern the file already uses — parent is now 1729 lines, all four moved tests verified running (`runtime_fallback::…` 4/4) and the suite's 6 failures are unchanged and all pre-existing.

Splitting rather than allowlisting, because the gate's allowlist is for refactors-in-progress and this is just a file that grew.

**The third bullet is the interesting one** — making unfinished budgeted-GC cleanup safe across independent TLS destructor ordering so the image benchmark exits cleanly after printing its checksum. TLS destructor order is genuinely unspecified between independent keys, so code that assumes one outlives another works until it doesn't, usually on a different host. Worth having fixed rather than worked around at the benchmark.

Keeping read-modify-write loops on native byte access when the RHS is call-free is the right guard: the call-free condition is what makes the native store safe, and tying the fast path to it rather than to a syntactic shape means the proof survives refactoring of the loop body.

Fixes #8261.

PR #8296, opened 2026-08-17
perf(codegen): skip unbounded guards for bounded bodies

### LINKED ISSUE #8202: perf(runtime): js_param_type_guard's interpretive walk dominates the remaining ~11pp (fixed per-call overhead now largely removed)

proggeramlug:
## Summary

- classify parameter descriptors by whether their runtime validation walk has a value-independent bound
- decline array/map/set and recursive structural guards when the guarded function body contains no loop
- retain bounded object/union guards and preserve structural guards for looping consumers that can amortize the walk

This applies the profitability result recorded in #8202: `peek(p: Parser)` walked every token to license an O(1) body, while `asNum(v: Value)` admitted a recursive 123-node graph to license a discriminant read. The issue’s same-runtime diagnostic measured those guards as a net loss.

On current `benchmarks/tls-budget/interp.ts`, the generated IR goes from exactly two `js_param_type_guard` call sites (`peek` and `asNum`) to zero; both helpers become direct `alwaysinline` bodies.

## Validation

- `cargo test -p perry-codegen --lib --no-fail-fast` (1068 passed)
- `python3 scripts/local_binding_type_audit.py`
- rebuilt `benchmarks/tls-budget/interp.ts` with `--trace llvm` and verified zero `call i32 @js_param_type_guard` sites

No version bump, as requested.

Fixes #8202

### DIFF
--- a/changelog.d/8296-bound-structural-guards.md
+++ b/changelog.d/8296-bound-structural-guards.md
@@ -1,0 +1,7 @@
+### perf(codegen): skip unbounded guards for bounded bodies
+
+Perry no longer walks runtime-sized arrays, maps, sets, or recursive object
+graphs to enter a function whose own work is statically bounded. This removes
+the costly structural parameter guards from the interpreter benchmark's
+`peek` and `asNum` helpers while retaining fixed-size object guards and guards
+for looping consumers that can amortize validation. Fixes #8202.
--- a/crates/perry-codegen/src/codegen/mod.rs
+++ b/crates/perry-codegen/src/codegen/mod.rs
@@ -2615,6 +2615,7 @@ pub fn compile_module(hir: &HirModule, opts: CompileOptions) -> Result<Vec<u8>> 
                 f.id,
                 &module_prefix,
                 &f.params,
+                &f.body,
                 &demoted,
                 &guard_blocked,
                 &cross_module.type_aliases,
--- a/crates/perry-codegen/src/codegen/ordinary_param_guard_tests.rs
+++ b/crates/perry-codegen/src/codegen/ordinary_param_guard_tests.rs
@@ -352,37 +352,38 @@ fn nonsuspending_async_function_needs_no_direct_call_site_for_its_guarded_clone(
 #[test]
 fn guarded_discriminant_branch_narrows_a_union_parameter_inside_the_clone() {
     // Renamed from `guarded_discriminant_branch_routes_recursive_field_to_clone`.
-    // The routing half of that name described a RECURSIVE union walk, which
-    // `guard_blocked` no longer admits: a call in the body can reach the
-    // guarded object through an alias the caller arranged before entry, so a
-    // reference-typed parameter cannot keep a descriptor proof across it. The
-    // narrowing machinery it was really exercising survives on a call-free
-    // body, and the recursive shape is kept below as the negative that pins
-    // the rule.
-    let node = Type::Union(vec![
-        Type::Object(ObjectType {
-            name: None,
-            properties: HashMap::from([
-                (
-                    "kind".to_string(),
-                    PropertyInfo {
-                        ty: Type::StringLiteral("num".to_string()),
-                        optional: false,
-                        readonly: false,
-                    },
-                ),
-                (
-                    "num".to_string(),
-                    PropertyInfo {
-                        ty: Type::Number,
-                        optional: false,
-                        readonly: false,
-                    },
-                ),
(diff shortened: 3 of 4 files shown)


proggeramlug:
Merging. Fixes #8202 — the issue whose remaining ~11pp I retitled this morning after its own PR showed the fixed per-call overhead was largely gone.

**Applying the profitability result rather than making the validator cheaper is the right call**, and it is what #8202 concluded: `peek(p: Parser)` walked every token to license an O(1) body, `asNum(v: Value)` admitted a 123-node recursive graph to license a discriminant read, and the same-runtime diagnostic priced both as a net loss. Declining the guard where the body cannot amortize the walk is the structural answer; a faster walk was never going to close a 2.7× gap.

Classifying by whether the *validation walk* has a value-independent bound — rather than by descriptor size or node count — is the part that generalises. #8242's author lost a build to a 64-node cap that `asNum`'s 123 nodes cleared; a bound-based classifier does not have that failure mode.

**Measured, with the caveat stated.** I compiled `gc-handoff/apps/interp.ts` against this branch and against the 2026-08-17 sweep binary (`14468dcbc`), output byte-identical (`1708840`):

| arm | instructions retired |
|---|---:|
| sweep baseline `14468dcbc` | 13.556 B |
| current main + this PR | **11.423 B** |

**−15.7%.** That is *cumulative*, not this PR alone — #8203, #8290, #8297 and #8301 all landed since the sweep and several touch this row, and I did not build a separate current-main arm to isolate it (disk would not take a second LTO build). #8202's own diagnostic priced the two guards at 12.0% of `interp`, so most of this is plausibly here, but I am not claiming the split without measuring it.

Either way `interp` was the #2 worst Perry/Node ratio in the sweep at 2.26×, and it has moved materially.

**Verified:** `perry-codegen --no-fail-fast` **28 suites, 1538 passed, 9 failed — all baseline, zero new**; `perry-runtime --lib` **2572/0/4**; **all 48 lint gates**.

The next four-engine sweep will give the clean per-row picture; worth re-running it once the current batch settles.

PR #8297, opened 2026-08-17
perf(gc): stamp exact starts only for Maps

### LINKED ISSUE #8288: perf(gc): #8277's object-start bitmap costs 8-10% wall on allocation-heavy workloads
`#8277` added a per-arena-block object-start bitmap to fix #8256. It is the right fix for that bug — an exact recorded allocation boundary is a real invariant where #8251's `size == 24` was a heuristic a correct-size fabrication defeats. It shipped with its cost unmeasured, and I flagged that at merge time. The 2026-08-17 four-engine sweep now measures it.

## The measurement

Both arms are clean sweeps on the quiet M1 mini, best-of-five interleaved, spreads 0.3–2.2%. Baseline is `38cf15336` (2026-08-15); current is `14468dcbc` (2026-08-17).

Split by whether the workload allocates:

| workload class | instructions retired, geomean | rows |
|---|---:|---:|
| allocation-heavy | **+3.69%** | 12 |
| everything else | **−48.3%** | 7 |

The concentration matches the mechanism, and allocation-light rows inside the same group are flat — so this is not a broad slowdown that happens to touch allocation:

| benchmark | wall Δ | instructions Δ |
|---|---:|---:|
| `churn_alloc` | +8.4% | +10.5% |
| `push_cls` | +8.7% | +10.4% |
| `retain_wide` | +9.6% | +8.7% |
| `retain_wide1` | +9.3% | +7.1% |
| `churn` | +8.2% | +5.9% |
| `cycles` | +1.4% | +1.8% |
| `shapes` | +1.7% | -0.5% |
| `retain` | +1.8% | -1.6% |
| `retain1` | +1.9% | -2.9% |
| `push_num` | +0.2% | +0.6% |

## Mechanism

`lower_call/new_alloc.rs` emits this on the inline fast path, per allocation:

```
load(current_data) · ptrtoint(raw) · ptrtoint(data) · sub · lshr · lshr · and
gep · load(bitmap) · gep · load(prior_starts) · shl · or · store
```

Three loads, one store, six ALU ops — and the bitmap word is on a different cache line from the object being allocated, so the real cost is plausibly a miss rather than the op count. There is also 1.5625% permanent metadata per arena block (one bit per 8-byte slot), which is RSS that never comes back.

## Why this is worth work rather than acceptance

The project rule is that RSS and compute are both minimized, always, not traded. A 8–10% wall regression on the allocation path is the kind of cost that compounds: `churn`, `churn_alloc`, `push_cls` and `retain_wide` are the shapes real programs hit constantly.

## Directions, cheapest first

1. **Stamp fewer bits.** The classifier needs an exact boundary only for objects a fabricated header could impersonate. `GC_TYPE_SET` (12) is already unreachable because it is not 8-aligned; if the reachable set is genuinely just Map-like fixed-layout types, the bitmap only needs stamping for those, not for every inline allocation.
2. **Hoist the block base.** `current_data` is reloaded per allocation. In a loop that allocates repeatedly into the same block it is invariant until the slow arm installs a new block — that is a phi, not a reload.
3. **Fold the bit-set into the bump.** The slot index is derivable from the bump pointer the allocator already computed; the two `lshr` plus `and` are recomputing what the allocation arithmetic knows.
4. **Byte map instead of bitmap.** Trades 8× the metadata (12.5%) for dropping the read-modify-write to a single store — worse on the RSS axis, so probably the wrong trade here, but worth pricing.

## Acceptance

Re-run the four-engine sweep and require the allocation-heavy geomean back under +1.0% instructions against `38cf15336`, with no regression on the `−48.3%` cohort and no loss of the #8256 fixture (`fabricated_map_rejection`). Report instructions **and** peak RSS together.

Sweep evidence: `gc-handoff/current-sweep-2026-08-17/` (`RESULTS.md`, `results-clean.json`).

proggeramlug:
Closes #8288

## Summary

- remove object-start bitmap read/modify/write IR from inline Object and Array allocation
- stamp exact starts only for runtime-allocated Maps, the arena-resident 8-aligned tag that needs allocation-authored evidence
- reject malloc-only type descriptors during arena classification and retain the correct-size fabricated-Map guard
- shrink `InlineArenaState` back to its three hot bump-allocation fields

## Performance

Clean quiet-M1-mini four-engine sweep, best-of-five interleaved, against `38cf15336` (same 19-source corpus):

| workload class | instructions retired, geomean | peak RSS geomean |
|---|---:|---:|
| allocation class (12 rows) | **-0.01%** (was +3.69%) | **-6.02%** |
| prior -48.3% cohort (7 rows) | **-48.90%** | **-5.46%** |

Peak observed Perry RSS is **404,520,960 B**, down from **421,429,248 B** at the baseline. The fixed branch is -3.56% instructions versus current main across the allocation class. Sweep verdict: `CLEAN` (load 1.02 -> 2.02, no foreign processes); all 19 Perry binaries passed output checks.

## Tests

- `cargo check -p perry-codegen --lib`
- `cargo test -p perry-runtime --lib` (2,569 passed, 4 ignored)
- `cargo test -p perry-codegen --lib the_inline_allocator_stores_its_header_prefix_as_one_vector_image`
- `git diff --check`

## Summary by CodeRabbit

- **Performance**
  - Improved garbage-collection allocation efficiency by reducing unnecessary object-start tracking for ordinary objects.
  - Optimized Map allocation handling, with benchmark results showing improved allocation performance and lower peak memory usage.

- **Bug Fixes**
  - Strengthened arena object classification to reject invalid or incompatible allocation records.
  - Preserved accurate tracking for Map objects while preventing false classifications for arrays and other object types.

- **Tests**
  - Added coverage for allocation tracking, arena resets, and invalid object classification scenarios.

### DIFF
--- a/changelog.d/8297-map-only-object-start-stamping.md
+++ b/changelog.d/8297-map-only-object-start-stamping.md
@@ -1,0 +1,12 @@
+### perf(gc): stamp exact starts only for Maps
+
+Arena allocation no longer updates the object-start bitmap for every Object
+and Array. Only Maps need allocation-authored boundary evidence: their type tag
+is 8-aligned and their copying descriptor follows an external entries pointer.
+The classifier retains that exact-start check for Maps and rejects malloc-only
+descriptors before arena dispatch.
+
+On the clean 19-row M1 mini sweep, allocation-class instructions improve from
+**+3.69% to -0.01%** versus the pre-bitmap baseline; the prior -48.3% cohort
+remains improved at **-48.90%**. Peak Perry RSS is 404,520,960 bytes versus
+421,429,248 bytes at baseline. Fixes #8288.
--- a/crates/perry-codegen/src/expr/array_literal.rs
+++ b/crates/perry-codegen/src/expr/array_literal.rs
@@ -201,26 +201,6 @@ pub(crate) fn lower_array_literal(ctx: &mut FnCtx<'_>, elements: &[Expr]) -> Res
             // GC_STORE_AUDIT(INIT): freshly allocated array header starts pointer-free until slot notes below.
             blk.store(I64, &gc_packed.to_string(), &raw);
 
-            // Publish the exact header boundary in the current arena block's
-            // object-start bitmap (InlineArenaState field at byte 24). The
-            // slow arm may have replaced the current block, so compute the
-            // slot from the merged raw pointer rather than `aligned_off`.
-            let current_data = blk.load(PTR, &state_ptr);
-            let raw_addr = blk.ptrtoint(&raw, I64);
-            let data_addr = blk.ptrtoint(&current_data, I64);
-            let relative = blk.sub(I64, &raw_addr, &data_addr);
-            let start_slot = blk.lshr(I64, &relative, "3");
-            let start_word = blk.lshr(I64, &start_slot, "6");
-            let start_bit = blk.and(I64, &start_slot, "63");
-            let bitmap_field = blk.gep(I8, &state_ptr, &[(I64, "24")]);
-            let bitmap = blk.load(PTR, &bitmap_field);
-            let bitmap_word = blk.gep(I64, &bitmap, &[(I64, &start_word)]);
-            let prior_starts = blk.load(I64, &bitmap_word);
-            let start_mask = blk.shl(I64, "1", &start_bit);
-            let updated_starts = blk.or(I64, &prior_starts, &start_mask);
-            // GC_STORE_AUDIT(INIT): side metadata for a freshly initialized GC header.
-            blk.store(I64, &updated_starts, &bitmap_word);
-
             // Packed ArrayHeader at raw+8 (length low 32 / capacity high 32).
             let arr_header_addr = blk.gep(I8, &raw, &[(I64, "8")]);
             let arr_header_packed = (n as u64) | ((n as u64) << 32);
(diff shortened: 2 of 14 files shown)


proggeramlug:
Merging. This closes #8288 and it meets the acceptance criterion I wrote there, including the part most perf PRs skip.

**#8288 asked for**: *"the allocation-heavy geomean back under +1.0% instructions against `38cf15336`, with no regression on the `−48.3%` cohort and no loss of the #8256 fixture. Report instructions and peak RSS together."*

**This delivers all four:**

| | #8288 measured | this PR |
|---|---:|---:|
| allocation class (12 rows) | +3.69% | **−0.01%** |
| prior cohort (7 rows) | −48.3% | **−48.90%** |
| peak RSS, allocation class | — | **−6.02%** |
| peak RSS, other cohort | — | **−5.46%** |

Peak observed Perry RSS 421,429,248 → **404,520,960 B**. Sweep verdict `CLEAN`, all 19 binaries output-checked. Reporting RSS alongside instructions is the standing rule and it is genuinely rare to see it honoured without being asked twice.

**And the #8256 guard is intact** — I checked, because "stamp fewer bits" is only correct if the thing the bits existed for still holds. `fabricated_map_rejection` still has all five tests passing, including the original `test_classify_arena_rejects_interior_pointer_as_map`, plus a new `test_plausible_gc_header_rejects_malloc_only_type_in_arena` covering the tightened arena classification. So the fabrication defence is stronger, not traded away.

Picking direction 1 from the ticket — restricting exact starts to runtime-allocated Maps, the arena-resident 8-aligned tag that actually needs allocation-authored evidence — is the right one: `GC_TYPE_SET` was never reachable (12 is not 8-aligned), so Object and Array allocations were paying for a defence against a fabrication they could not suffer. Shrinking `InlineArenaState` back to its three hot bump fields is the part that makes the inline path genuinely cheap again rather than merely shorter.

**Verified here:** `perry-runtime --lib` **2572/0/4**; `perry-codegen --no-fail-fast` **8 failures, all in the known baseline, zero new** (the first run stopped at 15 suites when the disk filled, so I ran the remaining four separately rather than report a partial sweep as clean); `cargo fmt` clean.

Issue #5235, opened 2026-06-16
AOT: defer .wasm imports to throw-on-use stubs (shared notice + strict) so the build proceeds


proggeramlug:
**Context:** after #5230 (defer dynamic import), compiling opencode's real source reaches a `.wasm` import — `@silvia-odwyer/photon-node/photon_rs_bg.wasm` (a transitive image dep). perry reads the `.wasm` as UTF-8 and hard-fails:
```
Error: Failed to read .../photon_rs_bg.wasm: stream did not contain valid UTF-8
```

## Proposal (defer, matching #5206/#5230)
When an import resolves to a `.wasm` file, don't read it as UTF-8/TS. Instead:
- Parse the WebAssembly **export section** (trivial binary walk: magic+version, section id 7) for export names.
- Synthesize a module that exports each name as a **throw-on-call** stub (descriptive: `wasm module <file> cannot run in an ahead-of-time compiled binary — full .wasm ESM instantiation is #5234`). Member calls (`wasm.add(...)`) throw only if reached.
- Record the site in the **shared** #5206/#5230 deferred-AOT-site notice; **strict mode** (`perry.strict` / a wasm knob) makes it a hard compile error.

Reuse: the text-asset synthesis path (#5223) as the file-handling model, and the deferred-site sink + notice + strict from #5230. Build proceeds; the wasm feature degrades unless reached.

## Why
Keeps real apps compiling past peripheral `.wasm` deps (leaf features), consistent with the eval/dynamic-import deferral policy. Full instantiation that makes the wasm actually run is the companion #5234.

Wall chain: new Function ✅ → @/ ✅ → require ✅ → text ✅ → dynamic import ✅ → **.wasm (this: defer; #5234: real)**.

proggeramlug:
Shipped in #5236 (`5258a6073`). Verified end-to-end against the issue's own base64 `add.wasm`:

- the `.wasm` export section is parsed (id-7 binary walk) and named exports become **throw-on-call** stubs — importing is fine, only calling throws
- the site is recorded in the shared deferred-AOT notice: `notice: 1 ahead-of-time-unsupported site compiled to a deferred runtime error`
- `--strict-dynamic-import` / `perry.strict` makes it a hard **compile-time** error
- the throw message points at the companion issue: *"wasm module add.wasm cannot run in an ahead-of-time compiled binary — full .wasm ESM instantiation is tracked in #5234"*

[...]

PR #8299, opened 2026-08-17
fix: instantiate .wasm ESM imports

### LINKED ISSUE #5234: AOT: instantiate .wasm ES module imports (import * as wasm from './x.wasm') and expose its exports
**Context:** real apps import compiled WebAssembly as ESM modules (wasm-bindgen "bundler"/"web" target). e.g. opencode pulls `@silvia-odwyer/photon-node`, whose `photon_rs_bg.js` does `import * as wasm from "./photon_rs_bg.wasm"` and calls `wasm.someExport(...)`. The import is expected to yield the **instantiated module's named exports**.

## Goal
Support `import * as wasm from "./x.wasm"` (and named/default forms) by **instantiating the WebAssembly module at module-init and exposing its exports as the module's named exports**, wired to the importing JS glue's imports object (wasm-bindgen creates a circular glue↔wasm import relationship). perry already has a runtime WASM host — `crates/perry-runtime/src/webassembly.rs` (`WebAssembly.instantiate(bytes)` works via `perry_wasm_host_*`) — to build on.

## Self-contained repro
Minimal module exporting `add(i32,i32)->i32`. Create `add.wasm` from this base64, then `main.ts`:
```bash
echo 'AGFzbQEAAAABBwFgAn9/AX8DAgEABwcBA2FkZAAACgkBBwAgACABags=' | base64 -d > add.wasm
```
```ts
// main.ts
import * as wasm from "./add.wasm";
console.log(wasm.add(2, 3));   // expected: 5
```
Node (with a wasm-ESM loader) and bundlers print `5`. perry today fails earlier — see #<DEFER> (it currently can't even read the `.wasm`); after deferral it would throw-on-use. This issue is the real integration that makes `wasm.add(2,3)` return `5`.

## Scope notes
- Instantiation at module-init; named exports surfaced (functions, and memory/global/table as appropriate).
- The wasm imports object: for a bare `.wasm` with no imports (like the repro) it's empty; for wasm-bindgen, the glue provides imports — handle the circular case (glue imports wasm exports; wasm imports glue functions).
- Companion to the cheaper deferral (#<DEFER>), which keeps the build moving until this lands.

proggeramlug:
## Summary

- replace the deferred .wasm ESM stub with an adapter that embeds and instantiates the module during module initialization
- expose function and memory exports through namespace, named, and default ESM imports
- route numeric wasm function imports through the JavaScript imports object, including circular wasm-bindgen-style glue imports
- preserve exported function arity for direct ESM calls

## Tests

- cargo test -p perry-wasm-host
- cargo test -p perry --bin perry wasm_asset -- --nocapture
- cargo test -p perry --test issue_5234_wasm_esm_import -- --nocapture
- cargo clippy -p perry-wasm-host --all-targets -- -D warnings
- cargo fmt -p perry-wasm-host -p perry-runtime -p perry -- --check
- python scripts/check_test_registration.py

Closes #5234

## Summary by CodeRabbit

* **New Features**
  * Added support for static WebAssembly ESM imports.
  * WebAssembly modules are now embedded and initialized during compilation.
  * Supports namespace, named, and default exports, including memory and imported functions.
  * Added support for circular WebAssembly glue-module imports.

* **Bug Fixes**
  * Replaced deferred WebAssembly import failures with executable module behavior.
  * Improved handling of imported numeric functions and malformed WebAssembly modules.


proggeramlug:
Merging. `perry-wasm-host` **6/0**, `perry-runtime --lib` **2571/0/4**, and the PR's own `wasm_esm_import_instantiates_and_exposes_exports` passes (81 s). `cargo fmt`, `check_file_size`, `check_test_registration` clean.

**On deleting `issue_5235_deferred_wasm_import.rs`** — I checked this rather than waving it through, because deleting a 215-line test is usually the wrong move. Here it is right: that test asserts the *deferred throwing-stub* contract — compilation succeeds, a notice lists the degraded site under the shared eval/dynamic-import header, and calling an export throws an `Error` referencing #5234. This PR replaces the stub with real instantiation, so every one of those assertions is now describing behaviour that deliberately no longer exists. #5235 is closed; #5234 is what the new test covers. Deleting a test whose subject was removed is correct, and keeping it would have been the actual mistake.

The scope is coherent: embedding and instantiating at module-init, exposing exports through namespace/named/default forms, routing numeric wasm imports through the JS imports object (including the circular wasm-bindgen-style glue case), and preserving arity for direct ESM calls. That last one is easy to miss and would surface as a confusing arity error rather than an obvious failure.

Issue #8213, opened 2026-08-16
[Next.js/dylib] Warm App Route server retains ~1,100 async box cells per request: 113 KB/req live Rust heap, minor-GC pause grows 59 ms -> 0.5 s


proggeramlug:
> **Scoped 2026-08-16.** This issue now tracks **mechanism (a) only — the async-activation box cells.** The two mechanisms below have independent fixes and were blocking each other:
>
> - **(b) the append-only Proxy registry** moved to #8230.
> - **(#8218 / PR #8219)** the band-exhaustion crash that (b) leads to — the 65,536th `new Proxy()` in a thread returned a payload the next property read dereferenced.
>
> Closing condition here is (a): #8208 lands, and the fixture slope (RSS + `[rss-diag-tables] box=` over 100+ passes) is measured before/after on the #8034 route. The measurements below are kept whole because they were taken together.

Serving the pinned #8034 production App Route fixture (release layout: providers + app dylib from clean `main` `3c95020f8`, C host), one warm process's RSS climbs 160 MB -> 510-620 MB and `[gc-step] post_in_use` climbs 5.4 -> 39.7 MB over 100 verifier passes (2,100 requests) with no plateau — first noted in the #8040 DoD audit and #8163's 2026-08-16 comment ("it wants its own home"). This issue is that home, with the diagnosis. All numbers measured 2026-08-16 on the bench mini (M1, 8 GB), 21 requests/pass (20 concurrent GETs + 1 POST via `verify.mjs`), `PERRY_GC_DIAG=1` (+`PERRY_GC_TRACE=1` where noted). Logs: `~/perry-bench.noindex/logs-rss/` on the mini; throwaway instrumented providers (counting `#[global_allocator]` wrapper + per-root-source retained-BFS, runtime crate only) in `~/perry-bench.noindex/build-rss-diag/`, patch in the session scratchpad — not shipped.

## 1. The growth is per-REQUEST (~150 KB/req), not per-connection (~0.6 KB/conn)

| client | connections | RSS pass 10 | RSS pass 100 | slope |
|---|---|---|---|---|
| stock `verify.mjs` loop (21 fresh conns/pass) | ~21/pass | 301 MB | 609 MB | **163 KB/request** |
| `http.Agent {keepAlive, maxSockets:1}`, sequential | **1 total** | 323 MB | 603 MB | **148 KB/request** |
| 20 persistent sockets, concurrent | 20 total | — | 600 MB | same |
[...]

PR #8300, opened 2026-08-17
fix(test): classify warm-soak failures instead of counting them (#8286)

### LINKED ISSUE #8286: test: run_warm_soak reports a client timeout as an #8163 residual — it counts verifier exits instead of classifying the failure
My bug, in the arm I added in #8215. Filing it before fixing it because **it is on `main` now and #8163 is closed**, so the next person whose warm soak trips a timeout will see the words `#8163 residual` and reasonably conclude it has regressed.

## What it does wrong

`run_warm_soak` in `tests/release/packages/next-app-route/fixture.sh` treats any non-zero `verify.mjs` exit as a failure and reports:

```
FAIL next-app-route — warm soak: N failing pass(es) [...], M host TypeError(s) — #8163 residual
```

It **counts** exits; it does not **classify** them. Two very different things produce a non-zero exit:

* **The residual** — the server returns an empty body. Signature: `Unexpected end of JSON input` in the verifier, plus `TypeError: value is not a function` and `E180 failed to pipe response` in the host log.
* **A client timeout** — `UND_ERR_HEADERS_TIMEOUT`. The server was merely slow; the host log is clean.

In the saturated regime the second happens on its own (see #8213: one pass ran 644 copying minors against a steady 84, and 8 of its 21 concurrent requests timed out). So the arm manufactures false residual reports precisely where it is most likely to be run.

## Observed

In the #8163 close-condition run at 211/300 passes: 8 failing passes, **all** `UND_ERR_HEADERS_TIMEOUT`, and across 6,267 copying minors the three residual discriminators were `0 / 0 / 0`. The arm would have printed `#8163 residual` for a run with no residual evidence whatsoever.

It is worth being precise about the direction: this arm was written to stop a *green* run implying more than it proves, and its actual first failure was the mirror image — a *red* verdict that isn't true. A gate that cannot be trusted when it goes red is as useless as one that cannot go red at all, and I did not think about that direction when I wrote it.

## Fix (mine, in progress)

Classify rather than count:

* residual → empty body **and** a host-side `TypeError: value is not a function` in the same pass. Fail with `#8163 residual`.
* timeout with a clean host log → report separately as a latency/stall observation, with the pass's copying-minor count next to it, since that is the interesting quantity (#8213). Whether it should fail the arm is a policy question — it is a real user-visible failure, but not the thing this arm exists to detect — so it should at minimum be named differently.
* anything else → report verbatim rather than assuming a cause.

Also worth adding while there: emit the per-pass minor counts, since the run above showed a 7.7× single-pass storm that a total would have hidden entirely.

Do not fix this by widening the grep. The lesson is that the arm asserted a *cause* from an *exit code*, and the fix is to make it read the evidence it already has in the host log.

proggeramlug:
Closes #8286. My bug, in the arm I added in #8215.

## What was wrong

`run_warm_soak` treated **any** non-zero `verify.mjs` exit as a failure and printed `#8163 residual`. It counted exits; it never classified them. Two very different things produce that exit:

* **the residual** — the server returns an **empty body**. Signature: `Unexpected end of JSON input` in the verifier, plus `TypeError: value is not a function` and Next's `E180 failed to pipe response` in the host log.
* **a client timeout** — `UND_ERR_HEADERS_TIMEOUT`. The server was merely slow; its log is clean and its answer, when it arrives, is correct.

The second happens **on its own** in the saturated regime this arm exists to reach. In the #8163 close-condition run: 12 passes lost requests to client timeouts, all inside collection storms — one pass ran **644** copying minors against a steady 84 — while the three residual discriminators were **0 across 10,995 collections**. The arm would have reported `#8163 residual` for a run with no residual evidence at all, against an issue that is **closed**, so the next reader would reasonably conclude it had regressed.

Worth stating in one line, because it is the general lesson and not a Perry-specific one: **this arm was written so a green run could not imply more than it proved, and its first real failure was the mirror image — a red verdict that was not true.** I only designed against one direction. A gate that cannot be trusted when it goes red is as useless as one that cannot go red at all.

## The fix

Classify from the evidence already in the host log — the arm was collecting it and throwing it away. Per failing pass, the slice of log that pass appended is examined:

| observed | verdict |
|---|---|
| empty body / host `TypeError` / `E180` | **FAIL** — `#8163 residual`, naming the passes |
| `UND_ERR_*` / `HeadersTimeoutError` / `BodyTimeoutError`, clean host log | reported with the pass's **collection count**, attributed to #8213, does **not** fail |
| anything else | **FAIL** — "read the log; do not assume a cause" |

A timeout does not fail the arm deliberately: it is a real user-visible failure (#8213) but it is not this arm's subject, and failing on latency would make the arm unusable in the exact regime it was built to reach. It is reported loudly rather than swallowed, with the offending pass's collection count next to it — because that is the interesting quantity, and a run total hides a 7.7× single-pass storm entirely.

Note the precedence: a residual co-occurring with a timeout reads as **residual**. The bug must never be masked by the noise.

## Sabotage-tested in both directions

Not "it compiles" — each verdict was **planted and observed**:

| planted | result |
|---|---|
| empty-body signature on pass 2 | `residual=1 timeout=0` → **FAIL**, names pass 2 |
| client timeout on pass 2, **with real traffic and 3 real collections** | `residual=0 timeout=1` → **exit 0** + the #8213 note |
| clean run, 12 passes | exit 0, confidence line |
| zero copying minors | **FAIL** — "exercised nothing" (pre-existing liveness assert, unchanged, still fires ahead of a clean verdict) |

Plus a 10-case classifier test covering every branch, the precedence case above, and the **real bytes** from passes 203 and 238 of the close-condition run — so the classifier is checked against the actual failure it mis-read, not only against synthetic strings.

The first version of the timeout sabotage was itself wrong and worth mentioning: a stub verifier that printed a failure without making any HTTP requests produced zero collections, so the liveness assert fired instead and the test proved nothing. Fixed by having the sabotage drive the real verifier first and then inject — a harness that does not exercise the subject cannot test it, which is the same lesson this PR is about.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved warm-soak verification failure classification.
  - Empty responses are now correctly identified as residual failures.
  - Client timeouts with clean host logs are reported separately and no longer fail the soak.
  - Unrecognized failures continue to fail verification with clear guidance not to assume a cause.
  - Timeout results now include collection counts for easier diagnosis.

- **Documentation**
  - Added supporting details and run data for the revised failure classifications.

### DIFF
--- a/changelog.d/8290-warm-soak-classify.md
+++ b/changelog.d/8290-warm-soak-classify.md
@@ -1,0 +1,26 @@
+### Fixed — the warm-soak arm reported a client timeout as an #8163 residual
+
+`run_warm_soak` (added in #8215) counted non-zero `verify.mjs` exits instead of
+classifying them, and printed `#8163 residual` for every one. Two very different
+things produce that exit: the residual, where the server returns an **empty body**
+(`Unexpected end of JSON input`, plus `TypeError: value is not a function` and
+Next's `E180` in the host log), and a **client timeout** (`UND_ERR_HEADERS_TIMEOUT`),
+where the server was merely slow and its log is clean.
+
+The second happens on its own in the saturated regime the arm exists to reach — in
+the #8163 close-condition run, 12 passes lost requests to client timeouts, all
+inside collection storms (one pass ran **644** copying minors against a steady 84),
+while the three residual discriminators were **0 across 10,995 collections**. So the
+arm manufactured false residual reports exactly where it was most likely to be used,
+and #8163 is closed, so the next reader would have concluded it had regressed.
+
+It now classifies from the evidence already in the host log: an empty body fails the
+arm as the residual; a timeout with a clean host log is reported loudly with the
+offending pass's collection count and attributed to #8213 (a warm server slow enough
+to blow a default HTTP client timeout is a user-visible failure, but it is not this
+arm's subject, and failing on it would make the arm unusable in the regime it exists
+to reach); anything unrecognised fails with "do not assume a cause".
+
+The arm was written so a *green* run could not imply more than it proved. Its first
+real failure was the mirror image — a *red* verdict that was not true. A gate that
+cannot be trusted when it goes red is as useless as one that cannot go red at all.
--- a/tests/release/packages/next-app-route/fixture.sh
+++ b/tests/release/packages/next-app-route/fixture.sh
@@ -301,11 +301,38 @@ run_warm_soak() {
   done
   [[ "$ready" == "1" ]] || fail "warm soak did not become ready"
 
-  local failed_iterations=()
-  local pass
+  # #8286: CLASSIFY each failing pass from the evidence in the host log, never
+  # from the verifier's exit code alone. Two very different things make
+  # `verify.mjs` exit non-zero, and the first version of this arm called both
+  # of them "#8163 residual":
+  #
+  #   * the residual — the server returns an EMPTY BODY. The verifier reports
+  #     `Unexpected end of JSON input`, and the host log carries
+  #     `TypeError: value is not a function` and Next's `E180`.
+  #   * a client timeout — `UND_ERR_HEADERS_TIMEOUT`. The server was merely
+  #     slow; its log is clean and its answer, when it arrives, is correct.
+  #
+  # The second happens on its own in the saturated regime this arm exists to
+  # reach (#8213: one pass ran 644 copying minors against a steady 84 and 8 of
+  # its 21 concurrent requests timed out), so counting exits manufactures false
+  # residual reports exactly where the arm is most likely to be used. The arm
+  # was written so a GREEN run could not imply more than it proved; its first
+  # real failure was the mirror image, a RED verdict that was not true.
+  local residual_passes=() timeout_passes=() other_passes=()
+  local pass before_lines slice
   for pass in $(seq 1 "$passes"); do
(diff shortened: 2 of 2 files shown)


proggeramlug:
Merging. Shell-only (`fixture.sh` + changelog), no Rust, so no suite needed; `bash -n` clean.

**The reasoning in that comment is the valuable part**, and it is the right instinct stated better than I would have:

> The arm was written so a GREEN run could not imply more than it proved; its first real failure was the mirror image, a RED verdict that was not true.

Both directions matter. A gate that can't fail is the usual worry, but a gate that fails for the wrong reason burns the same credibility — and it does it faster, because people start ignoring it. Distinguishing the residual (empty body, `TypeError: value is not a function`, Next `E180`) from a client timeout (`UND_ERR_HEADERS_TIMEOUT`, clean server log, correct answer when it lands) is exactly the discrimination that keeps the arm trustworthy.

That the timeout arises *on its own* in the saturated regime this arm exists to reach — #8213's pass with 644 copying minors against a steady 84, and 8 of 21 concurrent requests timing out — means counting exit codes would manufacture false residual reports precisely where the arm gets used. Slicing only the lines a given pass appended, rather than re-reading the shared host log, is the right mechanism for that.

One fix pushed: the changelog fragment was named `8290-warm-soak-classify.md`; renamed to `8300-`.

Issue #8289, opened 2026-08-17
perf: four benchmarks remain 1.4-2.2x Node with no owning issue (pipeline, churn, tree_wide, cycles)


proggeramlug:
The 2026-08-17 four-engine sweep leaves four rows where Perry is meaningfully behind Node and no issue owns them. `interp`/`iso_miss` are #8202 and `fib40` is #8175; these are the remainder.

| benchmark | Perry | Node | P/Node | Perry instr | Node instr |
|---|---:|---:|---:|---:|---:|
| `pipeline` | 0.2043 | 0.0950 | **2.15×** | 3.893 B | 1.220 B |
| `churn` | 0.3300 | 0.1666 | **1.98×** | 5.006 B | 2.681 B |
| `tree_wide` | 1.2965 | 0.9015 | **1.44×** | 26.808 B | 20.262 B |
| `cycles` | 0.1035 | 0.0736 | **1.41×** | 2.132 B | 0.669 B |

All four are instruction-bound rather than scheduling-bound — Perry executes 1.3–3.2× the instructions Node does — so the work is to emit less, not to schedule better.

Notes per row, from the sweep:

- **`pipeline`** is the worst uncovered gap. It was 3.3585 s at 08-14 and 0.2002 s at 08-15, so the catastrophic phase is long fixed (#8243 closed); this is the steady-state residue. Perry beats scriptc here 0.2043 vs 0.4461, and Porffor cannot run it at all, so the gap is specifically against V8.
- **`churn`** also carries the #8288 allocation-bitmap regression (+8.2% since 08-15). Fixing #8288 should be sequenced first, then this re-measured — the two are additive and it would be easy to double-count.
- **`cycles`** has the largest instruction ratio of the four at 3.19× Node for 1.41× wall, which suggests the emitted work is cheap per instruction. Possibly the most tractable.
- **`tree_wide`** is the mildest ratio but the largest absolute time in the corpus at 1.30 s.

Not a regression report — all four improved or held since 08-15 (`tree_wide` −24.8%, `cycles` +1.4%, `pipeline` +2.0%, `churn` +8.2% from #8288). Filed so the remaining engine gap is tracked rather than implicit, now that the campaign has cleared the catastrophic rows.

Suggested order: #8288 first (it contaminates `churn`), then `cycles` as the cheapest probe into where the extra instructions come from, then `pipeline`.

[...]

PR #8301, opened 2026-08-17
perf(gc): memoize typed layouts by ShapeId

proggeramlug:
## Summary

- memoize typed object layouts by immutable runtime `ShapeId`
- bypass repeated authoritative shape-descriptor lookups after the slot-count proof is cached
- preserve per-instance field validation, ambiguity poisoning, moving-GC behavior, and keyless-object fallback
- add the `cycles.ts` issue reproducer, direct fast-path regression coverage, and the reviewed shape-census reduction

## Performance

For `benchmarks/issue-8289/cycles.ts`, the median of five local runs dropped from 2.177B to 1.804B retired instructions (**-17.1%**) while preserving output (`3999998000000`).

## Testing

- `cargo test -p perry-runtime` (2,572 passed, 4 ignored; doc tests: 7 ignored)
- focused shape-install memo tests after the final rebase (5 passed)
- `RUSTFLAGS="-D warnings" cargo check -p perry-runtime --all-targets`
- `python3 scripts/shape_descriptor_census.py`
- `python3 scripts/raw_handle_debt.py`
- `git diff --check`
- prior CI run: all six gap-suite shards and GC stress passed

No version or manifest files changed.

Refs #8289

## Summary by CodeRabbit

* **Performance**
  * Improved allocation performance for classes containing pointer fields.
  * Reduced repeated validation and metadata processing when creating similar objects.
  * Benchmark results show approximately 17% fewer retired instructions in the targeted workload.

* **Reliability**
  * Preserved per-instance validation, fallback behavior for ambiguous layouts, and garbage-collection tracing.

* **Testing**
  * Expanded regression coverage for layout reuse, invalidation, pointer-mask handling, and unsupported layouts.


proggeramlug:
Merging. Responsive to #8289, which named `cycles` as one of the four remaining Perry-vs-Node gaps with the largest instruction ratio (3.19× Node for 1.41× wall) — so it was the row I flagged as most likely tractable, and −17.1% instructions is a good bite out of it.

**The invariants I wanted are all covered and passing**, which matters because a layout memo keyed by an identifier is only sound if the identifier is stable and the memo yields to validation:

- `ambiguity_poison_invalidates_the_memo` — the invalidation path
- `a_contradicting_field_is_refused_even_with_the_memo_warm` — the memo does not override per-instance validation
- `a_memo_hit_produces_the_same_header_state_as_the_install` — hit and miss agree
- `memo_installed_objects_survive_a_copying_minor_with_their_children`
- `test_shape_keyed_typed_layout_survives_copying_minor` and `..._survives_layout_transfer` — new, and these are the two that pin the ShapeId keying specifically across a move

**Verified:** `perry-runtime --lib` **2572/0/4**, `perry-codegen --no-fail-fast` **28 suites, 1535 passed, 9 failed — all baseline**, and **all 47 lint gates** via the new `scripts/run_lint_gates.sh`.

Two notes:

**A flake, not a regression.** One full-suite run showed `r#box::release_tests::the_intrusive_free_list_round_trips_a_whole_cohort` failing. It passes in isolation and on a re-run of the full suite, and passes on `main`. That test is from #8208's intrusive free list and looks order-dependent under parallel execution — worth a `--test-threads=1` guard or explicit state reset before it bites someone as a phantom regression.

**Census baseline**: your later commit already refreshed it. For the record the delta is strictly removals — two raw `keys_array` reads gone, 184 → 182 sites — which is the memo doing its job.

One fix pushed: the changelog fragment was named `8289-…` after the issue rather than the PR; renamed to `8301-`.

PR #8303, opened 2026-08-17
fix(runtime): release closure-visible async boxes

### LINKED ISSUE #8213: [Next.js/dylib] Warm App Route server retains ~1,100 async box cells per request: 113 KB/req live Rust heap, minor-GC pause grows 59 ms -> 0.5 s

proggeramlug:
## Summary

- release the complete boxed frame at terminal states for plain async activations
- have codegen declare exact boxed closure-capture slots to the runtime, including singleton-cache hits and runtime closure clones
- retain each captured cell only until the last live closure edge disappears; closure moves rekey the metadata and authoritative GC death pruning publishes the cell
- treat drained closure-owned boxes as ephemeron-style edges during full traces, so a box/closure self-cycle cannot manufacture liveness while a genuinely live escaped closure still traces and rewrites the box payload

## Pinned production-route evidence

After #8208, the #8034 Next 16.3.0 App Route still retained about 5,061 JS box entries/pass and ended 100 passes at 510,399 entries. The first closure-aware implementation was also rejected: it passed correctness but ended at 1,311,279 JS entries and 739,616 KB RSS because a pending box rooted the closure that kept the same box pending.

With this change:

- 100/100 verifiers passed (2,100 requests)
- RSS samples: 291,072 KB (pass 1), 347,072 (10), 526,336 (50), 581,696 (100)
- final box stats: `resident_cells=533377 registry_len=340058`, down from the rejected design's `1374382/1311279`
- an extended 200/200 soak also passed (4,200 requests), with RSS 612,896 KB (100), 620,784 (150), and 655,456 (200)
- at 200 passes: `resident_cells=625119 registry_len=586936`; despite twice the requests, those are only 22%/15% above the post-#8208 100-pass baseline, and full collections visibly reset the box-registry scan cost instead of preserving a monotone per-request tail
- 4,147,729 of 4,398,461 terminal releases had published by the end of the 200-pass soak

## Validation

- `cargo check -p perry-runtime -p perry-codegen -p perry-transform`
- `cargo test -p perry-runtime --lib` — 2,577 passed, 0 failed, 4 ignored
- `cargo test -p perry-transform` — 84 passed
- `cargo test -p perry-codegen --lib` — 1,091 passed
- boxed native proof regressions — 7 passed
- GC root-dominance self-test plus alloc-regex, poll-capable, poll-reach, and immovable-source audits
- release runtime symbol audit — all 102 sentinel symbols present

No version bump is included.

Closes #8213

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved async closure and boxed-variable lifetime handling.
  * Prevented captured values from being reclaimed while still accessible to live closures.
  * Improved garbage collection tracing for captured boxed values.
  * Ensured released boxed storage is safely reusable after all references end.

* **Tests**
  * Expanded coverage for async closure lifetimes, garbage collection, release behavior, and boxed capture handling.


proggeramlug:
Merging. Closes #8213, and finishes what #8208 explicitly could not.

**The rejected first implementation is the most valuable thing in this PR's description.** *"a pending box rooted the closure that kept the same box pending"* — a self-cycle manufacturing its own liveness, ending at 1,311,279 JS entries and 739,616 KB. Recording the design you threw away, with the reason, is worth more than the one you kept: the next person to try closure-aware release will recognise that shape immediately. Treating drained closure-owned boxes as ephemeron edges is the right answer to it, and `full_trace_treats_drained_closure_boxes_as_ephemeron_edges` pins it.

**The test set covers the cases that actually bite.** `recycled_activation_token_rejects_a_stale_generation` is the ABA guard; `escaped_closures_defer_activation_cell_publication_until_gc_death` is the escape path; `closure_move_rekeys_the_capture_count_before_death_pruning` is the rekeying claim; `exception_unwind_releases_only_this_pumps_activation_refs` covers the unwind path that longjmp makes easy to get wrong. Sixteen in total, all passing.

**And the evidence is honest about what remains**: RSS still climbs across the soak (291 MB at pass 1 to 582 MB at 100), but `resident_cells=533377 registry_len=340058` against post-#8208's 510,399, and full collections *reset* the registry scan cost rather than preserving a monotone per-request tail. That qualitative change is the point; a 200-pass soak at only 22%/15% above the 100-pass baseline is the number that shows it.

**Two gate failures I fixed rather than waved through** — both found by running the full 48-gate set:

- `check_file_size`: `box.rs` reached **2228** lines (main: 1903). Moved `mod release_tests` into `box/release_tests.rs`; box.rs is now 1617.
- `addr_class_inventory`: `box.rs:1825` casts a raw `GcHeader`. It is legitimate — the closure is allocated by the test and it needs a **mutable** header to toggle and restore `GC_FLAG_MARKED`, which `try_read_gc_header`'s shared ref cannot express — so it gets an allowlist entry with that rationale, following the `arena/tests_promoted_runs.rs` precedent. The split conveniently moves the cast into a test file where the wildcard entry is the established form.

**Verified after the fixes:** `perry-runtime --lib` **2578/0/4**, `release_tests` **16/16**, `perry-codegen --no-fail-fast` **28 suites, 1538 passed, 9 failed — all baseline**, and **all 48 lint gates**.

Issue #6767, opened 2026-07-22
[parity] node:test — 53 failing node-suite tests (2026-07-22 baseline)


TheHypnoo:
Part of the node-suite parity sweep of 2026-07-22 (full run: 3,815 tests, 3,019 pass, 79.1% — report `test-parity/reports/parity_report_20260722_120605.json`, oracle Node 26.5.0 per `.node-version`). Umbrella: #793.

`node:test` has **53 failing test(s)** in `test-parity/node-suite/test/`.

Reproduce:
```bash
./run_parity_tests.sh --suite node-suite --module test
```

## Failing tests

- [ ] `node-suite/test/imports/aliases` — output mismatch
  - **Node**:    test.test identity: true
  - **Perry**:  test.test identity: false
- [ ] `node-suite/test/imports/module-shape` — output mismatch
  - **Node**:    test default: function
  - **Perry**:  test default: function
- [ ] `node-suite/test/mock-fn/accessor-options` — output mismatch
  - **Node**:    getter options: initial 1
  - **Perry**:  TypeError: The "implementation" argument must be of type function. Received an instance of Object
- [ ] `node-suite/test/mock-fn/implementations` — output mismatch
  - **Node**:    sequence: once-b:1 replacement:2 replacement:3
  - **Perry**:  sequence: once-a:1 once-b:2 replacement:3
- [ ] `node-suite/test/mock-fn/method-options-validation` — output mismatch
  - **Node**:    options number: ERR_INVALID_ARG_TYPE
  - **Perry**:  options number: NO_THROW
- [ ] `node-suite/test/mock-fn/name-length` — output mismatch
  - **Node**:    name: original original
  - **Perry**:  name: original mockFn
- [ ] `node-suite/test/mock-fn/once-indices` — output mismatch
  - **Node**:    indexed: default third default fifth default
  - **Perry**:  indexed: third fifth default default default
- [ ] `node-suite/test/mock-fn/property-access-records` — output mismatch
  - **Node**:    get: 2
  - **Perry**:  get: 2
- [ ] `node-suite/test/mock-fn/property-context` — output mismatch
  - **Node**:    initial: 2 object
  - **Perry**:  initial: 2 undefined
- [ ] `node-suite/test/mock-fn/property-no-value` — output mismatch
  - **Node**:    initial spy: 5 1
  - **Perry**:  TypeError: Cannot read properties of undefined (reading 'accessCount')
[...]

PR #8304, opened 2026-08-17
test(parity): threshold floors ratchet from the measured 2026-08-17 baseline (#8271)

proggeramlug:
Final data piece of the #8271 ratchet (companion to #8273): the per-category **threshold** gate ran for the first time in CI via `parity-aggregate` (run 31988840780 — the merge machinery worked, producing the exact whole-suite report) and failed on 21 categories whose 100% floors predate the six-week dark period.

**What**:
- 18 single-test categories (argon2, buffer, cluster, commander, cron, crypto, date_fns, dayjs, decimal, dotenv, lodash, lru_cache, moment, nanoid, test→#6767, util, uuid, validator) get configured 0% floors — each one's test is already a triaged `known_failures.json` entry; the floor returns to 100 in the same PR that fixes the test.
- The 3 legacy buckets floor at their measured rates (gap 98.5 / issue 81.0 / other 87.5) so any further slide is red while the debt burns down. Up-only.
- The other direction (#7582): 4 stale 0% floors that **measured 100%** on the same report (`parity/http`, `parity/http2`, `parity/https`, `parity/sqlite`) are raised to 100 — a 0-floor that outlives its gap is a permanent suppression of that category's next regression.
- Global min stays 83.0 (measured 90.7%).

**Verified**: `parity_threshold_gate.py --check` exits green against the real merged 8-shard report.

With this + #8291 (lint unbreak, in flight), the sharded parity gate in #8244 goes end-to-end green on today's `main` except for genuinely new regressions — which is the whole point.

## Summary by CodeRabbit

* **Documentation**
  * Added release documentation for updated parity threshold floors.
  * Documented baseline measurements, exceptions, ratcheting rules, and verification results.

* **Chores**
  * Raised coverage requirements to 100% for HTTP, HTTP/2, HTTPS, and SQLite parity categories.
  * Replaced selected zero-percent floors with measured minimum thresholds.
  * Clarified and documented remaining category exceptions.


proggeramlug:
Merging. JSON + changelog only, no code.

**I checked the direction before believing the title**, because "ratchet" and "add 18 zero floors" can describe either a tightening or a capitulation. Numerically: **4 raised, 0 lowered, 21 added**.

The 18 new 0% floors are not a weakening in effect — those categories previously fell under `default_category_min_parity_pct: 100.0`, which they could not meet, so the gate was permanently red and therefore ignored. Each is a single-test category whose one test is already a triaged `known_failures.json` entry; I verified all 17 checkable names appear there. A floor that returns to 100 in the PR that fixes the test is a work-queue item, where a permanently-red gate is just noise.

**The part that makes this a real ratchet rather than a re-baseline is the other direction.** Four stale 0% floors that *measured 100%* on the same report — `parity/http`, `parity/http2`, `parity/https`, `parity/sqlite` — are raised to 100. Verified in the merged file. That framing is exactly right:

> a 0-floor that outlives its gap is a permanent suppression of that category's next regression

That is the same failure mode as a skip-list entry nobody revisits, and catching it in the same pass that adds new exemptions is what keeps the mechanism honest. The three legacy buckets flooring at their measured rates (gap 98.5 / issue 81.0 / other 87.5) are up-only, so further slide goes red while the debt burns down.

Companion to #8273, which did the same thing for `known_failures.json` provenance — and which also refused to park the #8223 trio. Consistent discipline across both.

One fix pushed: the changelog fragment was named `8299-threshold-floors.md`; renamed to `8304-`.

PR #8305, opened 2026-08-17
fix(gc): close unrooted-local detector blind spots

### LINKED ISSUE #8233: gc: the raw-handle ratchet cannot see unrooted locals in perry-stdlib / perry-ext-* — 90% of that surface never adopted the rooting API

proggeramlug:
## Summary

Completes the source-level detector added in #8253 so issue #8233 is covered beyond allocator-to-allocator call shapes. It now catches ordinary returns and dereferences after a collection point, plus raw pointers extracted from NaN-boxes.

## Changes

- Track the first collection point after each raw-pointer binding and inspect every later expression, including `let` right-hand sides and ordinary returns.
- Treat NaN-box pointer extractors as raw-pointer sources.
- Add positive controls for collecting calls, plain returns, and later RHS dereferences, plus negative controls.
- Protect both the total and per-file baseline ceilings against same-PR increases by comparing with the merge base.
- Re-pin the corrected detector as an explicit schema 1 to 2 migration: 605 findings across 84 files.
- Correct the existing changelog fragment. No version bump or package metadata change.

## Related issue

Closes #8233.

Follow-up to #8253 and its post-merge review findings.

## Test plan

- `python3 scripts/unrooted_local_shape.py --self-test`
- `python3 scripts/unrooted_local_shape.py --check`
- `python3 scripts/unrooted_local_shape.py --no-raise-vs origin/main`
- Deliberate schema-2 total and new-file ceiling increases are rejected.
- `python3 scripts/check_locale_independent_io.py --self-test`
- `python3 scripts/check_locale_independent_io.py`
- Workflow YAML parses successfully.

- [ ] `cargo build --release` clean (not run: Python/workflow-only change)
- [ ] `cargo test --workspace ...` passes (not run: no Rust change)
- [x] Detector behavior is covered by its sabotage-style `--self-test`
- [x] No user-facing CLI / stdlib / runtime API changed
- [x] No platform UI backend changed

## Checklist

- [x] My commit follows the repository prefix convention

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of unsafe stale-local pointer usage across returns, collections, shadowed bindings, and derived pointers.
  * Added safeguards for baseline schema changes, missing files, and unresolved comparison references.

* **Tests**
  * Expanded validation coverage for pointer extraction, collection behavior, baseline violations, and clean control flows.

* **Chores**
  * Pull requests now compare results against the merge base to detect newly introduced findings.
  * Updated detection baselines and clarified related documentation.


proggeramlug:
Merging. This completes #8253, which I wrote, and it closes a blind spot I documented rather than guessed at.

**The surface it recovers is large**: 218 → **605 findings across 84 files**. On #8265 I recorded that my detector *could not see* the bug that PR fixed — a `Vec<i64>` listener snapshot, invisible because I only matched `let x = js_*_alloc(...)` bindings. `perry-ext-net` goes 1 → 3 here, so that class is now partly visible. Tracking returns after a collection point, later `let` right-hand sides, and NaN-box pointer extractors is the right generalisation: the original shape was the narrowest possible reading of #8233.

Ground truth still holds — `events_on.rs:40` is still flagged for both `state` and `buffer`, which is the case #8233 names by hand.

**Adding the `--no-raise-vs` merge-base step is the important structural fix.** My version compared the count against a baseline the same diff could move; #7659 already solved that for `raw_handle_debt` and this borrows the pattern. The unresolvable-ref path errors rather than passing silently, which I checked.

**One weakness I found while sabotage-testing, worth a follow-up rather than a block.** I planted two extra findings *and* raised the baseline to match — the attack `--no-raise-vs` exists to stop. Both gates passed:

```
--check          rc=0
--no-raise-vs    rc=0
  recorded unrooted-local debt vs. origin/main:
  audited schema migration 1 -> 2, baseline 218 -> 607
```

The schema-migration allowance is unbounded, so a PR that bumps the schema version can raise the baseline by any amount. That is only open to schema-bumping PRs — which is exactly this one — so **I verified the number by hand instead**: measured on a clean tree gives **605**, and the committed baseline is **605**. The recorded figure is honest.

Worth bounding that allowance later (e.g. require the migration to state its expected delta, or re-derive rather than accept). Filing it as a note here rather than holding the PR, since the mechanism is strictly better than what it replaces and the number checks out.

All **48** lint gates pass via `scripts/run_lint_gates.sh`.

PR #8306, opened 2026-08-17
fix(lint): underscore the Linux arm's stack_lo too — #8298 fixed only the macOS arm

proggeramlug:
One-liner unbreaking the `warnings` job (`-D warnings`, host-compatible scope) on `main`: #8298 underscore-fixed the unused `stack_lo` on the **macOS** arm of `gc/native_stack_scan.rs` but left the bare name on the `#[cfg(not(target_os = "macos"))]` arm — the one Linux CI actually compiles. Caught on #8291's rerun (job 95306309639's sibling run 32015928121); verified still live on `main` tip `85a60d4a3`.

## Summary by CodeRabbit

* **Refactor**
  * Updated internal stack-bound handling for non-macOS builds without changing stack-scanning behavior.

### DIFF
--- a/crates/perry-runtime/src/gc/native_stack_scan.rs
+++ b/crates/perry-runtime/src/gc/native_stack_scan.rs
@@ -60,7 +60,10 @@ pub(super) fn run_native_stack_scan() {
         (top - size, top)
     };
     #[cfg(not(target_os = "macos"))]
-    let (stack_lo, stack_hi) = {
+    // `_stack_lo`: same unused-on-this-arm shape #8298 underscore-fixed on the
+    // macOS arm; the Linux arm kept the bare name and turned `warnings`
+    // (-D warnings, host-compatible scope) red on main.
+    let (_stack_lo, stack_hi) = {
         // Fallback: use a local variable address and scan 256KB upward.
         let marker: usize = 0;
         let sp = std::ptr::addr_of!(marker) as usize;


proggeramlug:
Merging. This is a correction to **my** #8298, and the failure mode is worth naming because it is not obvious.

#8298 fixed the unused `stack_lo` on the `#[cfg(target_os = "macos")]` arm and I verified it with `RUSTFLAGS="-D warnings" cargo check` — on macOS. That check **structurally cannot see** the `#[cfg(not(target_os = "macos"))]` arm, because rustc never compiles it here. So a green local `-D warnings` run said nothing at all about the arm Linux CI actually builds, and `warnings` stayed red on `main`.

Generalising: **a warnings check on one host is not evidence about `cfg`-gated code for another.** Same class as a `#[cfg(windows)]` thread-local slipping past a census I ran on macOS earlier today (#8270). Worth remembering that "I ran the gate" and "the gate covered this code" are different claims whenever `cfg` is involved.

I checked the rest of the file for the same shape before merging, so this should not need a third round: `native_stack_scan.rs` has two `cfg(not(target_os = "macos"))` arms and the only destructuring binding in either is the one this PR fixes.

Thanks for catching it — and for the precise provenance (job 95306309639's sibling run, verified live on tip). That made it a thirty-second confirmation instead of a hunt.

PR #8308, opened 2026-08-17
fix: resolve parity crash cluster

### LINKED ISSUE #8272: parity: 16 tests CRASH in the first full run since July 4 (tui/ink, fastify, cluster, gc-unsafe-zones, wasm, mysql)

proggeramlug:
Fixes #8272.

## Summary

- preserve cluster worker identity across bootstrap and expose the cached state to the separately linked HTTP and Fastify adapters, restoring shared-port coordination and listening events
- remove the obsolete WebAssembly instance-local lowering shortcut and exercise the standard instantiate result plus safe exported-function closures
- add source-level parity-skip reasons for interactive, background-server, and external-service fixtures so expected lifecycle timeouts are not reported as runtime crashes
- remove all #8272 entries from the parity known-failure ratchet and add expected output for the Perry-specific WASM fixture

## Validation

- cargo build --release -p perry -p perry-runtime-static -p perry-stdlib-static -p perry-ext-http -p perry-ext-fastify
- cargo build --release -p perry-runtime-static -p perry-wasm-host --features perry-runtime/wasm-host
- cargo test -p perry-hir --lib (315 passed, 1 ignored)
- PERRY_SKIP_BUILD=1 PERRY_RUN_TIMEOUT=30 ./run_parity_tests.sh --filter cluster_4962 (PASS, 0 crashes)
- PERRY_SKIP_BUILD=1 PERRY_RUN_TIMEOUT=30 ./run_parity_tests.sh --filter test_wasm_add (PASS, 0 crashes)
- python3 scripts/parity_known_failures.py --audit
- cargo fmt --all -- --check
- bash -n run_parity_tests.sh

No version bump is included.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved clustered HTTP worker detection for more reliable behavior.
  * Fixed crashes when invoking WebAssembly exports.
  * Updated WebAssembly integration behavior to support the standard instantiation result shape.

* **Tests**
  * Enhanced parity testing to skip fixtures requiring interactive input, servers, or external services.
  * Added expected output and documentation for WebAssembly parity coverage.
  * Removed obsolete known-failure entries.


proggeramlug:
Merging. Fixes #8272 — the crash cluster #8273 baselined this morning.

**This is the ratchet working exactly as designed.** #8273 made the point that every parked entry is a work-queue item and *"the floor returns to 100 in the same PR that fixes the test"*. This PR takes `known_failures.json` from **32 `#8272` mentions to zero** and the audit from 139 entries to **123**, all in the same change that fixes the crashes. That is the bidirectional half earning its keep: the entries could not silently outlive their cause, because an entry that starts passing goes red too.

Verified: `parity_known_failures.py --audit` reports *"123 entries carry provenance and cross-checked 5 gap entries against gap_snapshot.json"*, and `--self-test` passes.

**The parity-skip reasons are the right distinction to draw.** Adding source-level skips for interactive, background-server and external-service fixtures means an expected lifecycle timeout stops being reported as a runtime crash. That is the same discrimination #8300 made for the warm-soak arm a few hours ago — a red verdict that is not true costs as much credibility as a green one that is not earned, and both come from a harness that cannot tell two failure modes apart.

**Sequencing note:** removing the obsolete WebAssembly instance-local lowering shortcut lands cleanly after #8299, which replaced the deferred `.wasm` stub with real instantiation. Doing the shortcut removal now, on top of a path that genuinely instantiates, is the right order — the other way round would have left a window where neither path was whole.

`perry-runtime --lib` **2578/0/4**, **all 48 lint gates** pass.

PR #8313, opened 2026-08-17
perf(object): shrink common objects to 40 bytes

### LINKED ISSUE #8047: perf(object/GC): finish the common-object header shrink from 56 B to 40 B after shape-transition migration

proggeramlug:
Closes #8047.

## Summary

- remove the per-object `keys_array` compatibility mirror and derive ordered keys from the authoritative ShapeId descriptor
- make the descriptor record the sole stable, traced, rewritten keys edge for moving GC
- shrink `ObjectHeader` to 16 bytes on LP64 and explicitly padded ILP32, producing 40-byte two-slot objects and 88-byte eight-slot objects
- remove the inline allocator's per-allocation keys-global load/store, update field offsets, the FFI mirror (ABI revision 3), null stub, and watchOS layout documentation together
- harden the exact source/IR census so reintroducing the member or an old emitted header offset fails the required lint gate

The delete path also no longer republishes its predecessor's keys edge after installing a compacted successor descriptor; the redundant synchronization became incorrect once the header mirror was removed.

## Validation

- `RUST_TEST_THREADS=1 cargo test -p perry-runtime --lib --quiet` — 2,575 passed, 4 ignored
- `cargo test -p perry-codegen --lib --quiet` — 1,094 passed
- `cargo test -p perry-ffi --features runtime-link --quiet` — 51 passed
- `cargo test -p perry-stdlib --lib --quiet` — 124 passed
- `BASE_SHA=origin/main ./scripts/run_lint_gates.sh` — all 48 gates passed
- CI-equivalent product and host-compatible clippy scopes passed
- `scripts/gc_instrument_smoke.sh` with freshly built release/static archives — 14/14 protected-fromspace probes clean; verifier arm copied 3,894 objects; default-stride arm ran 3,852 copying minors and moved 492,718 objects

Layout tests pin `size_of::<ObjectHeader>() == 16`, `{a,b} == 40` bytes including the GC header and two slots, and the eight-slot layout at 88 bytes.

## Measurement note

The archived 19-program `gc-handoff/m0810` measurement harness referenced by #8047 is not present in the repository or this workspace, so I did not invent replacement instruction/cycle/RSS numbers. The non-vacuous moving-GC and exact-layout acceptance checks above were run on this branch.

No version bump is included.

## Summary by CodeRabbit

* **New Features**
  * Reduced object memory usage with a more compact object header.
  * Property keys are now sourced from shared object metadata, improving memory efficiency.
* **Bug Fixes**
  * Updated garbage collection and object operations to consistently preserve and resolve property keys.
  * Maintained object layout compatibility across 32-bit and 64-bit targets, including watchOS.
* **Tests**
  * Expanded layout, garbage collection, shape, cloning, and serialization coverage for the updated object representation.


proggeramlug:
Holding this one — the gap suite finds a real regression. Everything else I ran
is green, so this looks like one missed key path rather than a problem with the
approach.

## The regression

`fs.Stats` methods are gone. Minimal repro:

```ts
import * as fs from 'fs';
import * as path from 'path';
import * as os from 'os';
const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'perry_mf2_'));
const testFile = path.join(tmpDir, 'test.txt');
fs.writeFileSync(testFile, 'hello');
const stat = fs.statSync(testFile);
console.log(typeof stat.isFile);
console.log(stat.isFile());
```

| build | output |
|---|---|
| `main` (5b0b5cb) | `function` / `true` — byte-identical to Node |
| this branch | `undefined` / `TypeError: isFile is not a function` |

Both binaries built the same way in their own worktree
(`cargo build --release -p perry -p perry-runtime-static -p perry-stdlib-static`),
same `PERRY_RUNTIME_DIR`, `--no-cache`.

This is `test_gap_node_fs` in the gap suite: `pass -> parity_fail`. On `main` its
23 lines match Node exactly; here it dies on the first `stat.isFile()`.

My guess at the cause, given the diff: `Stats` is a runtime-constructed object
whose methods are installed through a path that the new
descriptor-derived ordered-keys lookup does not see, where the deleted
`ObjectHeader::keys_array` mirror did. `typeof` returning `undefined` (rather
than a wrong value) says the property is not being *found*, not that it is being
found stale.

## The other four gap "regressions" are NOT yours

The run also flagged `dayjs_factory_arg`, `moment_methods`,
`ratelimiter_memory` and `slugify_options` as `pass -> parity_fail`. Those are
environmental: **node itself exits 1** on all four in this checkout
(`ERR_MODULE_NOT_FOUND` — `dayjs`, `moment`, `rate-limiter-flexible` and
`slugify` are not installed locally), so the harness is comparing Perry's
success against Node's stack trace. Ignore them; `node_fs` is the only genuine
one.

Gap totals for the record: 557 pass / 13 parity_fail / 0 compile_fail / 0 crash.

## Everything else passed

- `perry-runtime --lib` 2575, `perry-codegen --lib` 1094, `perry-ffi
  --features runtime-link` 51, `perry-stdlib --lib` 124 — all 0 failed
- all 48 lint-tier gates
- the new `gc::tests::shape_keys_descriptor_edge` set 4/4, including
  `keys_edge_sabotage_is_detected` and
  `a_keys_array_whose_last_carrier_died_is_still_reclaimed` — both directions
- I sabotage-tested the census ratchet myself: re-adding a `(*obj).keys_array`
  read to `object/shapes.rs` does turn `shape_descriptor_census.py` red, so the
  hardening is real and not just present

The descriptor-edge tests are unusually good — real copying minor, gated on
`copied_objects > 0` *and* on the receiver's address actually changing, with a
sabotage arm. Worth keeping that shape.

Note the Next App Route gate can't referee this: it is already failing on `main`
(run 31993068277, step "Production App Route provider gate"), independently of
this PR.

proggeramlug:
Correction to the table above: the `main` control was built at **99e19c228**,
not the hash I typed. The comparison itself is unaffected — that is the commit
whose binary printed `function` / `true`.

proggeramlug:
Re-reviewed the updated head — the regression is fixed and the root cause is
right. Merging.

## The `isFile` regression is gone

```
                          typeof stat.isFile   stat.isFile()
previous revision         undefined            TypeError
this revision             function             true
```

`test_gap_node_fs` now matches Node byte-for-byte and no longer appears anywhere
in the mismatch list.

The explanation in `180b24f89` is the convincing part: `fs.Stats` carries 21
public keys plus four hidden `Date` slots, so its cached canonical shape (21/21)
legitimately disagrees with the object descriptor (21/25). Before #8047 the
reject path rebuilt those facts from the header's `keys_array` mirror; with the
mirror gone, `birth_publish_object_shape` rebuilt through a null edge and
replaced an exact 21/25 descriptor with a keyless 0/25 one — which is exactly the
symptom I saw (`typeof` returning `undefined`, i.e. the property was not being
*found* rather than found stale). Keeping the descriptor already stamped by
`set_object_keys_array_with_live` is the general fix, not a Stats special case,
and `stats_hidden_slots_retain_public_method_keys` pins it.

## Full gap re-run: no genuine regressions

I re-ran the whole suite rather than just the one test, since the fix changes
shape publishing generally.

| | previous run | this run |
|---|---|---|
| parity_pass | 557 | **559** |
| parity_fail | 13 | **11** |
| compile_fail / crashed / skipped | 0 | 0 |

Six entries are flagged `pass -> parity_fail`: `backoff_options`,
`cron_cronjob`, `dayjs_factory_arg`, `moment_methods`, `ratelimiter_memory`,
`slugify_options`. All six are environmental — **node itself exits 1** on every
one in this checkout (`exponential-backoff`, `cron`, `dayjs`, `moment`,
`rate-limiter-flexible`, `slugify` are not installed locally), so the harness is
comparing Perry's success against Node's stack trace. That verdict is
independent of which Perry build is used, so none of them can be attributed
here.

## Everything else

- `perry-runtime --lib` 2576 · `perry-codegen --lib` 1095 — 0 failed
- all 48 lint-tier gates
- `gc::tests::shape_keys_descriptor_edge` 4/4 including the sabotage arm and the
  reclaim-when-dead arm
- census ratchet independently sabotage-checked: re-adding a `(*obj).keys_array`
  read does turn `shape_descriptor_census.py` red

Note for #8324, which is still open: this PR deletes `gc_keys_array_slot`
entirely, so that PR's third fix targets code which no longer exists, while its
`prune_dead_shape_keys` and `scan_shape_table_rekey_mut` fixes still apply. I've
asked there for a rebase on top of this.

Issue #5902, opened 2026-07-03
test262 built-ins/String — 32 (self-contained worklist)


proggeramlug:
**Self-contained worklist — 32 failing.** _Snapshot 2026-07-03 full sweep (perry 0.5.1218, node v26.3.0). Everything you need is in this issue; no history to read, no re-measure to start._

String methods on non-generic receivers, normalize, well-formed, symbol coercion. Crate: perry-runtime (string), perry-codegen.

## Setup (any box)
```bash
git clone https://github.com/PerryTS/perry && cd perry
cargo build --release -p perry -p perry-runtime-static -p perry-stdlib-static
git clone --depth 1 https://github.com/tc39/test262 vendor/test262   # checkout test-compat/test262/pinned-sha.txt
# node on PATH (differential oracle). See test-compat/test262/README.md.
```
Measure/verify: `scripts/test262_subset.py --root vendor/test262 --dir built-ins/String --jobs 8` (the `.failures.txt` sidecar lists failures; diff before/after for zero-regression).

## Failure reasons
| n | reason |
|---|---|
| 4 | Uncaught exception: #1.1: Exception === "intostr". Actu |
| 3 | Uncaught exception: #1: __instance = new Object(true);  |
| 2 | Uncaught exception: #1: __instance = new Boolean; __ins |
| 2 | Uncaught exception: #1.1: Exception is instance of Type |
| 2 | Uncaught exception: Expected a TypeError to be thrown b |
| 1 | Uncaught exception: #2: toString=function(){return "__T |
| 1 | Uncaught exception: #1: Function.prototype.indicator =  |
| 1 | TypeError: Cannot assign to read only property 'length' |
| 1 | Uncaught exception: #2: var __str__instance = new Strin |
| 1 | Uncaught exception: #1: delete String.prototype.toStrin |
| 1 | Uncaught exception: #2: delete String.prototype.toStrin |
| 1 | Uncaught exception: #1: var x; __instance = {toString:f |

## All 32 failing cases (path — reason)
```
built-ins/String/S15.5.1.1_A1_T9.js    Uncaught exception: #2: toString=function(){return "__THIS__";}; __str = String(
built-ins/String/S15.5.3_A2_T2.js    Uncaught exception: #1: Function.prototype.indicator = 1; String.indicator === 1
built-ins/String/S15.5.5.1_A4_T1.js    TypeError: Cannot assign to read only property 'length' of object '#<Object>'
[...]

PR #8319, opened 2026-08-17
fix: preserve sloppy member update semantics

proggeramlug:
## Summary
- carry source strictness through `PropertyUpdate` and `IndexUpdate` HIR
- route update write-back through strict-aware `PutValue` semantics
- make `with` environment writes honor their existing strict flag
- add HIR and end-to-end regressions for sloppy rejected writes and strict throws

This fixes `built-ins/String/S15.5.5.1_A4_T1.js` from #5902. It also covers frozen objects and explicit non-writable descriptors for named and computed updates.

Progresses #5902.

## Tests
- `cargo check -p perry-hir -p perry-codegen -p perry-codegen-js -p perry-codegen-wasm -p perry-transform -p perry-runtime -p perry`
- `cargo test -p perry-hir --test c262_parity member_updates_preserve_script_strictness_5902 -- --exact --nocapture`
- `cargo test -p perry-codegen issue7628 --lib -- --nocapture`
- `cargo test -p perry --test issue_5902_sloppy_property_update sloppy_member_updates_ignore_rejected_writes -- --exact --test-threads=1 --nocapture`
- `cargo test -p perry --test issue_5902_sloppy_property_update strict_member_update_still_throws_on_a_rejected_write -- --exact --test-threads=1 --nocapture`
- pinned Test262 `built-ins/String/S15.5.5.1_A4_T1.js`: 1 pass, 0 diff/runtime-fail/compile-fail

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed sloppy-mode `++` and `--` updates on frozen or non-writable properties so rejected writes are ignored while preserving the original value.
  * Strict-mode updates continue to throw `TypeError` for rejected writes.
  * Applied consistent behavior to updates involving `with` environments.

* **Tests**
  * Added regression coverage for sloppy and strict member-update behavior, including computed properties and rejected writes.


proggeramlug:
Holding this one: the PR's own acceptance test fails.

```
test sloppy_member_updates_ignore_rejected_writes ... FAILED
  sloppy updates must not throw
  stderr: TypeError: Cannot assign to read only property 'length' of object '#<Object>'
test strict_member_update_still_throws_on_a_rejected_write ... ok
```

I split the fixture into its five independent statements to find which one
throws. **Four of the five pass — the `PropertyUpdate`/`IndexUpdate` strictness
work is sound.** Only the `with` path is still broken:

| # | statement | result |
|---|---|---|
| 1 | `frozen.named++` | `ok1 1 1` |
| 2 | `--frozen["computed"]` | `ok2 1 2` |
| 3 | **`with (b) length = 0;`** | **`TypeError: Cannot assign to read only property 'length'`** |
| 4 | `boxed.length++` | `ok4 8 8` |
| 5 | `++descriptor.value` (non-writable) | `ok5 5 4` |

Node on the same source prints `ok3 8` — a silent no-op, as sloppy mode requires.

So the third bullet of the summary — "make `with` environment writes honor their
existing strict flag" — has not taken effect. `with_env.rs` still routes to the
throwing setter for a rejected write in sloppy code. Everything else in the PR
is verified working, so this looks like one remaining call site rather than a
design problem.

Everything else I ran is green: `perry-hir --test c262_parity` 20/20,
`perry-codegen --lib issue7628` 3/3, `perry-codegen --lib` 1094/1094, and all 48
lint-tier gates. `strict_member_update_still_throws_on_a_rejected_write` passes,
so the strict direction is right — the sloppy direction is what is missing.

Happy to merge as soon as case 3 is green.

proggeramlug:
Traced the `with` case a bit further, in case it saves you time — the flag
plumbing you added all looks right, so the throw is coming from below it:

- `expr/instance_misc1.rs:277` emits `strict_i32 = if *strict { "1" } else { "0" }`
  and passes it as the 4th argument, so codegen forwards the HIR flag.
- `object/with_env.rs:94` `js_with_set_binding` takes it and hands it to
  `js_put_value_set_ic_miss(coerced, key, value, strict, null)` — your change.
- In `proxy/put_value.rs` the strict handling reads correctly to me: the array
  `length` branch splits on `strict != 0`, and the general rejected-write path
  is guarded `if !ok && strict != 0`. The one unconditional `throw_type_error`
  in that function is the null/undefined-target case, which is meant to throw
  in both modes.

So either the HIR `strict` is arriving as `true` for a `.js` script, or the
boxed-String `length` write leaves `js_put_value_set_ic_miss` before reaching
that guarded path. The message text is
`error.rs:1759` (`js_throw_type_error_immutable_write`, kind 0) — worth noting
that `grep -rn js_throw_type_error_immutable_write crates/` finds **no caller
anywhere in the tree**, so whatever reaches it does so by some route other than
a direct Rust call. That seemed like the most useful thread to pull, and it is
where I stopped.

Everything else in the PR is verified green — see my previous comment for the
per-statement breakdown.

proggeramlug:
Thanks for the detailed split. I traced this and the branch does not reproduce the failure when the runtime archive is rebuilt from the PR checkout.

The important detail is that `cargo test -p perry --test issue_5902_sloppy_property_update` builds the compiler/test binary, but Cargo does **not** build the `perry-runtime-static` staticlib target. The test pins `PERRY_RUNTIME_DIR` to the compiler profile directory, so an older `libperry_runtime.a` / `perry_runtime.lib` there links the pre-PR `js_with_set_binding` implementation and produces exactly the unconditional read-only error you saw.

A fresh matching archive makes both cases pass:

```sh
cargo build -p perry-runtime-static
cargo test -p perry --test issue_5902_sloppy_property_update
```

Result here: `2 passed; 0 failed`. I also dumped the HIR for the isolated case and confirmed `WithSet { property: length, ..., strict: false }`.

The PR's Linux `e2e-scoped` job already performs the explicit staticlib build before running diff-named integration suites; its run of this exact suite is green (`sloppy_member_updates_ignore_rejected_writes ... ok`, `strict_member_update_still_throws_on_a_rejected_write ... ok`). So no remaining `with` call site is missing; this was a stale runtime archive in the manual run. Could you rerun after rebuilding `perry-runtime-static`?

For completeness, the red aggregate PR gate is currently from unrelated failures that also reproduce on `main`; the PR's exact acceptance job is green.

proggeramlug:
Re-checked on the updated head (rebased onto #8316, changelog fragment added).
The three files I reviewed before — `object/with_env.rs`, `proxy/put_value.rs`,
`expr/member_update.rs` — are byte-identical to the previous revision, so I
rebuilt and re-ran rather than assuming, in case something merged into `main`
had changed the outcome. It has not:

```
test sloppy_member_updates_ignore_rejected_writes ... FAILED
  TypeError: Cannot assign to read only property 'length' of object '#<Object>'
test strict_member_update_still_throws_on_a_rejected_write ... ok
```

Same single case as before: `with (b) length = 0` on a boxed `String`, where
Node prints `ok3 8`. The other four statements in the fixture still pass, so
only the `with`-environment path is outstanding.

Still happy to merge as soon as that case is green — the rest of the PR is
verified (`perry-hir --test c262_parity` 20/20, `perry-codegen --lib issue7628`
3/3, `perry-codegen --lib` 1094/1094, all 48 lint gates).

proggeramlug:
Fixed in a736c6151. The remaining failure was reproducible when the integration test linked a stale `libperry_runtime.a`: `cargo test` rebuilds the compiler/test binary but does not build staticlib targets.

The regression test now follows the repository's established integration-test pattern and builds `perry-runtime-static` once before compiling its JS fixtures, so the reviewer's plain command cannot silently reuse a pre-fix `js_with_set_binding`.

Verified from an isolated PR worktree with exactly:

```sh
cargo test -p perry --test issue_5902_sloppy_property_update
```

Result: `2 passed; 0 failed` (`sloppy_member_updates_ignore_rejected_writes` and `strict_member_update_still_throws_on_a_rejected_write`).

@coderabbitai review

proggeramlug:
Correction: my two "still failing" reports on this PR were wrong, and the cause
was on my side.

Both times I ran

```
cargo test --release -p perry --test issue_5902_sloppy_property_update
```

in a worktree where I had never built `perry-runtime-static`. `cargo test` does
not build staticlib targets, so the fixture linked a stale `libperry_runtime.a`
containing the pre-fix `js_with_set_binding` — exactly the failure mode your new
`runtime_dir()` helper documents and now prevents. The `TypeError: Cannot assign
to read only property 'length'` I kept reporting was the OLD runtime, not this
branch's.

Rebuilt with the wrappers (`cargo build --release -p perry -p
perry-runtime-static -p perry-stdlib-static`) at the current head:

```
test sloppy_member_updates_ignore_rejected_writes ... ok
test strict_member_update_still_throws_on_a_rejected_write ... ok
test result: ok. 2 passed; 0 failed
```

So all five statements pass, including `with (b) length = 0`, and the `with_env.rs`
change was working the whole time. My follow-up comment tracing `strict` through
`js_with_set_binding` → `js_put_value_set_ic_miss` was chasing a bug that was not
there; please disregard it. Adding the explicit staticlib build to the fixture is
the right fix and would have saved us both the round trip.

Running the remaining suites and gates now, then merging.

PR #8323, opened 2026-08-17
fix(test): preserve ELF provider export surface

### LINKED ISSUE #8160: test(provider): #8127 fixed the ELF gate by removing all symbol hiding — 19 exports became 894, 871 now shared with the runtime image
## Summary

#8127 turned `gc-native-roots` green on ELF and should stay. But it does not restore the pre-#8089 export surface — it **removes all hiding**, which is a much larger change than "put the runtime's symbols back to global", and the gate now shares symbols with the runtime image that it never shared before.

Measured on Linux/aarch64 (`rust:1-bookworm`), driving this repo's actual `tests/fixtures/issue_8075_provider_gc/stdlib-linker.sh` taken verbatim at four refs via `git show <ref>:…`, against a provider cdylib built the same way the gate builds it:

| `stdlib-linker.sh` from | `js_gc_init` in `.dynsym` | exports | shared with runtime | probe |
|---|---|---:|---:|---|
| `5c27d1ad9` (last good) | `T` global | **19** | **2** | OK |
| `origin/main` before #8127 | absent — local `t` | 16 | 0 | reproduces the CI error |
| **#8127** | `T` global | **894** | **871** | OK |
| delete the interception | `T` global | **19** | **2** | OK |

A version script with a `global:` section and no `local: *` hides nothing, so every symbol rustc would have hidden stays exported. That is why 16 named exports becomes 894.

The 871 newly-shared symbols are not the fetch surface — they are std/panic/alloc internals: `___rust_alloc`, `rust_panic`, `std::panicking::begin_panic`. libstd is the same rlib in both images, so they hash-match and will genuinely bind across at load.

## Why this is worth fixing rather than leaving

The blast radius is confined to this fixture, which is why #8127 was the right call to unblock a red gate. But the fixture's whole job is to detect *the stdlib provider binding to the wrong runtime image*, and it now has panic/alloc machinery cross-binding between the two images by preemption. That is the same class of coupling the gate exists to police, in a component the gate does not check. A future failure here would be genuinely confusing to diagnose.

## The smaller fix

**Delete the `--version-script` interception on the ELF branch.** Row 4 above: it reproduces the last-good surface exactly — 19 exports, 2 shared, probe OK — because rustc's own generated script already lists every `#[no_mangle]` symbol in the crate graph as global, upstream rlibs included. All 16 names in `stdlib_provider_exports` are `#[no_mangle]` in `perry-stdlib`/`perry-runtime`, so rustc's script already grants every one of them. The interception was never adding anything on ELF.

Note for anyone tempted by the other direction: naming the runtime's symbols explicitly (as the Mach-O branch does via `nm`) is **not** available on ELF. rustc passes `--no-undefined-version` immediately after its own `--version-script`, so naming a symbol the output does not define is a hard lld error — measured, and it is why #8127's first commit failed CI.

## Second ask: make this fail at link time, not 90 minutes later

This arrived as a runtime error string after a full gate run. A one-line check in `scripts/gc_provider_dylib_gate.sh` after the provider links:

```sh
nm -D "$provider_so" | grep -q ' T js_gc_init' || { echo "provider hid js_gc_init — not preemptible"; exit 1; }
```

turns the whole class into an immediate, self-describing failure. Worth having regardless of which fix lands, since the preemption property is invisible to every other check in the gate.

## Provenance

Bisect confirmed: `git log 5c27d1ad9..506f4ab11` contains exactly one commit, #8089. It **causes** rather than exposes — the provider has always bundled a second copy of the runtime's C surface, but ELF preemption neutralised it by design, so there is no latent defect a revert would have hidden. Mach-O and Windows arms were green throughout; the Darwin branch never sets `original_version_script`.

The hypothesis in #8156 (an archive trim that stopped removing bundled members) is **not** the mechanism — there is no trim anywhere in the fixture. It is symbol visibility.

Refs #8156, #8089, #8127, #8075.

proggeramlug:
## Summary

- pass the rustc-generated ELF version script through unchanged so the provider retains the pre-#8089 dynamic export surface
- fail the provider gate immediately when `js_gc_init` is hidden and cannot be preempted
- leave workspace and package versions unchanged

Fixes #8160

## Testing

- `bash -n tests/fixtures/issue_8075_provider_gc/stdlib-linker.sh scripts/gc_provider_dylib_gate.sh`
- `shellcheck tests/fixtures/issue_8075_provider_gc/stdlib-linker.sh scripts/gc_provider_dylib_gate.sh`
- Linux `rust:1-bookworm` synthetic rlib/provider link: exactly two dynamic exports, including global `js_gc_init`, with a separate `libperry_runtime.so` dependency

### DIFF
--- a/changelog.d/8323-elf-provider-export-surface.md
+++ b/changelog.d/8323-elf-provider-export-surface.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- **The ELF provider-dylib fixture preserves rustc's export surface (#8160).**
+  Its linker shim now passes rustc's generated version script through unchanged
+  instead of replacing it with a script that made hundreds of Rust standard
+  library, panic, and allocator internals globally preemptible. The gate also
+  checks `js_gc_init` immediately after linking, so a future visibility
+  regression fails with a direct diagnostic instead of surfacing much later as
+  a provider/runtime identity mismatch.
--- a/scripts/gc_provider_dylib_gate.sh
+++ b/scripts/gc_provider_dylib_gate.sh
@@ -112,6 +112,10 @@ if [[ "$host_os" == Darwin ]]; then
   install_name_tool -id '@rpath/libperry_runtime.dylib' "$runtime_library"
   install_name_tool -id '@rpath/libperry_stdlib.dylib' "$stdlib_library"
 else
+  nm -D "$stdlib_library" | grep -q ' T js_gc_init' || {
+    echo "stdlib provider hid js_gc_init — not preemptible" >&2
+    exit 1
+  }
   readelf -d "$stdlib_library" | grep -Fq "Shared library: [$runtime_filename]" || {
     echo "stdlib provider is not bound to the separate runtime provider" >&2
     exit 1
--- a/tests/fixtures/issue_8075_provider_gc/stdlib-linker.sh
+++ b/tests/fixtures/issue_8075_provider_gc/stdlib-linker.sh
@@ -7,7 +7,6 @@ host_os=$(uname -s)
 arguments=()
 skip_export_list_value=false
 original_export_list=""
-original_version_script=""
 saw_runtime_rlib=false
 stdlib_provider_exports=(
   issue_8075_stdlib_runtime_probe
@@ -52,18 +51,13 @@ for argument in "$@"; do
     -Wl,-exported_symbols_list,*)
       original_export_list=${argument#-Wl,-exported_symbols_list,}
       ;;
-    -Wl,--version-script=*)
-      original_version_script=${argument#-Wl,--version-script=}
-      ;;
     *) arguments+=("$argument") ;;
   esac
 done
 
 custom_export_list=""
-custom_version_script=""
 cleanup() {
   [[ -z "$custom_export_list" ]] || rm -f "$custom_export_list"
-  [[ -z "$custom_version_script" ]] || rm -f "$custom_version_script"
 }
 trap cleanup EXIT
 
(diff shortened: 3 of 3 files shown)


proggeramlug:
Audited and merging. This one is ELF-only, so I could not exercise it on macOS —
I verified its premise on real Linux in a container instead.

## The premise checks out

The change rests on rustc's own generated version script already exporting
`js_gc_init`, which lives in the statically-linked `perry-runtime` **rlib**, not
in the cdylib's own source. The deleted comment argued the opposite — that
rustc's `local: *` would bind it internally and make it non-preemptible.

Built a minimal `rlib` (`#[no_mangle] pub extern "C" fn js_gc_init`) + `cdylib`
that depends on it, on `rust:latest`, aarch64 Linux:

```
A_TOTAL=2            # dynamic  T  symbols under rustc's default script
A_HAS_GCINIT=1       # js_gc_init is one of them
```

So the rlib's `#[no_mangle]` symbol *is* dynamically exported through
`local: *`, and the export set is exactly the `#[no_mangle]` API — nothing else
leaks. The new `nm -D "$stdlib_library" | grep -q ' T js_gc_init'` assertion
will hold, and the old comment's concern was unfounded.

The link line also shows rustc passing `-Wl,--no-undefined-version` next to its
own `--version-script`, which is the constraint that comment cited — so that
part of the reasoning was right, it just pointed at the wrong conclusion.

## One thing I did NOT verify

I could not reproduce the "hundreds of Rust standard library, panic, and
allocator internals" count. Appending a second `--version-script` via
`RUSTFLAGS` is a hard link error (rustc already passes one), whereas the shim
*replaced* rustc's — so my A/B for that half was not faithful and I dropped it
rather than quote a number I hadn't measured. The direction is not in doubt: a
`global:`-only script with no `local: *` leaves every other symbol at default
global binding. I just can't put a figure on it.

## Rest of the audit

- `stdlib_provider_exports` is still live after the removal (line 69, the Darwin
  `-exported_symbols_list` path), so nothing became dead.
- Both scripts pass `bash -n`.
- The Darwin branch is untouched, so macOS behavior is unchanged.
- All 48 lint-tier gates pass.
- Context: both ELF arms of `gc-native-roots` are currently **green** on `main`
  (run 32034804868), so this is hardening rather than a red-build fix — the old
  shim was passing while over-exporting. The added assertion is what makes a
  future visibility regression fail loudly instead of resurfacing later as a
  provider/runtime identity mismatch.

Issue #6764, opened 2026-07-22
[parity] node:async_hooks — 113 failing node-suite tests (2026-07-22 baseline)


TheHypnoo:
Part of the node-suite parity sweep of 2026-07-22 (full run: 3,815 tests, 3,019 pass, 79.1% — report `test-parity/reports/parity_report_20260722_120605.json`, oracle Node 26.5.0 per `.node-version`). Umbrella: #793.

`node:async_hooks` has **113 failing test(s)** in `test-parity/node-suite/async_hooks/`.

Reproduce:
```bash
./run_parity_tests.sh --suite node-suite --module async_hooks
```

## Failing tests

- [ ] `node-suite/async_hooks/api-function-metadata` — output mismatch
  - **Node**:    module function metadata: createHook:createHook/1|executionAsyncId:executionAsyncId/0|executionAsyncResource:executionAsyncResource/0|triggerAsyncId:triggerAsyncId/0
  - **Perry**:  module function metadata: createHook:createHook/1|executionAsyncId:executionAsyncId/0|executionAsyncResource:executionAsyncResource/0|triggerAsyncId:triggerAsyncId/0
- [ ] `node-suite/async_hooks/hooks/async-await-trigger-chain` — output mismatch
  - **Node**:    async-await continuation ids: true true true
  - **Perry**:  async-await continuation ids: false false false
- [ ] `node-suite/async_hooks/hooks/async-wrap-providers-table` — output mismatch
  - **Node**:    provider table shape: object true true
  - **Perry**:  provider table shape: object false true
- [ ] `node-suite/async_hooks/hooks/cancelled-immediate-destroy` — output mismatch
  - **Node**:    cancelled immediate resource identity: true
  - **Perry**:  cancelled immediate resource identity: false
- [ ] `node-suite/async_hooks/hooks/cancelled-timeout-destroy` — output mismatch
  - **Node**:    cancelled timeout resource identity: true
  - **Perry**:  cancelled timeout resource identity: false
- [ ] `node-suite/async_hooks/hooks/destroy-enqueues-destroy` — output mismatch
  - **Node**:    destroy queue identities: true true true
  - **Perry**:  destroy queue identities: false false false
- [ ] `node-suite/async_hooks/hooks/disable-during-promise` — output mismatch
  - **Node**:    disable during promise init: 3
  - **Perry**:  disable during promise init: 0
- [ ] `node-suite/async_hooks/hooks/disable-in-init` — SIGSEGV (exit 139)
[...]

PR #8328, opened 2026-08-17
fix(async_hooks): track scheduler resource lifecycles

proggeramlug:
## Summary

- dynamically dispatch AsyncHook enable/disable calls and keep hook callbacks rooted across moving GC
- give timeout, immediate, interval, microtask, and next-tick work stable async resources and lifecycle events
- add regression coverage for multiple hooks, scheduler resource identity/trigger IDs, and repeating interval lifecycles

This advances the scheduler and multi-hook portions of the umbrella parity tracker in #6764; remaining promise/provider cases stay tracked there.

## Testing

- cargo test -p perry-runtime async_hooks::tests -- --test-threads=1
- PERRY_RUNTIME_DIR="$PWD/target/debug" PERRY_NO_AUTO_OPTIMIZE=1 cargo test -p perry --test issue_6764_async_hooks_lifecycle -- --nocapture
- PERRY_SKIP_BUILD=1 PERRY_BIN=/Users/amlug/projects/perry/perry/target/release/perry PERRY_RUNTIME_DIR="$PWD/target/debug" ./run_parity_tests.sh --suite node-suite --module async_hooks --filter multiple-hooks
- scripts/check_file_size.sh

No version bump.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed async hook lifecycle behavior for scheduled callbacks, microtasks, promises, and timers.
  * Preserved hook enable/disable behavior and stable async resource identities.
  * Ensured callbacks receive expected lifecycle events, including repeating timer events.
  * Improved callback and context handling to prevent invalid references during asynchronous execution.

* **Tests**
  * Added regression coverage for scheduler resources, callback execution, trigger chains, and repeating timers.


proggeramlug:
Holding this one — it trips two lint-tier gates, both in the rooting family the
PR is trying to improve. I isolated it from #8327 (merged separately, all 48
gates green on its own), so these are this PR's.

## 1. `raw_handle_debt` — this is the substantive one

```
bare raw-handle reads: 993 (baseline 983)
::error::per-module raw-handle rules: 4 violation(s)
  crates/perry-runtime/src/async_hooks.rs: 23 bare reads exceeds its ceiling of 18
  crates/perry-runtime/src/builtins/globals.rs: 14 bare reads exceeds its ceiling of 12
  crates/perry-runtime/src/promise/mod.rs: 2 bare read(s) in a module with no ceiling
  crates/perry-runtime/src/timer.rs: 8 bare reads exceeds its ceiling of 7
```

The summary says the change keeps "hook callbacks rooted across moving GC", but
it adds ten bare raw-handle reads across exactly the four modules it touches —
including `promise/mod.rs`, which has no ceiling at all, meaning new code there
is expected to use the rooted forms from the start. Please convert these to
`RuntimeHandle::across_{mut,const,nanbox}` for a post-call reload, or
`with_{mut,const}_ptr` for a scoped argument to a non-allocating operation
(#7341). The ratchet may only shrink.

## 2. `gc_runtime_root_holders` — two unclassified holders

```
crates/perry-runtime/src/async_hooks.rs:142: ASYNC_HOOK_HANDLES: LazyLock<Mutex<HashSet<i64>>>  [rule B]
crates/perry-runtime/src/async_hooks.rs:144: ASYNC_HOOK_HANDLE_COUNT: AtomicUsize             [rule B]
```

This one is paperwork with a clear answer, and the precedent is two lines away
in the same file — `ASYNC_RESOURCE_HANDLES` and `ASYNC_RESOURCE_HANDLE_COUNT`
are already `not_a_gc_pointer` in `scripts/gc_runtime_root_holders.json`. Add
the matching pair.

One wording note: don't copy the resource entry verbatim. `ASYNC_RESOURCE_HANDLES`
is described as a set of *IDs*, whereas `ASYNC_HOOK_HANDLES` stores
`Box::into_raw(Box::new(AsyncHookHandle { index })) as i64` — a real pointer,
just a `std::alloc` one outside the GC heap that is never moved. The verdict is
still `not_a_gc_pointer`; the `why` should say that, so the next reader doesn't
have to re-derive it.

## An observation, not a blocker

`HOOKS` holds `HookRecord { callbacks: HookCallbacks, .. }` and there is no
`gc_register_mutable_root_scanner` anywhere in `async_hooks.rs`. The census did
not flag it, so either it is classified elsewhere or its type does not match the
rules — but if `HookCallbacks` holds NaN-boxed JS callbacks, that is a root the
scanner cannot see, which is the shape described in CLAUDE.md's "a runtime-side
cache of a raw heap pointer is a GC root, and the static checker cannot see it".
Worth confirming while you are in here; it is pre-existing either way, so not
something I would block on.

## What passed

Everything else is green, so this is close:

- `perry-runtime --lib` 2577 · `perry-stdlib --lib` 124 — 0 failed
- `issue_6764_async_hooks_lifecycle` — 1/1
- the other 46 lint-tier gates

Issue #507, opened 2026-05-06
perry-ext-net: outbound TCP panics — LTO dead-strips tokio CONTEXT statics


proggeramlug:
## Summary

On v0.5.586 (and v0.5.572+), any outbound TCP via `net.createConnection`
panics on a tokio worker thread before any `'connect'`/`'error'`/`'close'`
event fires. The script appears to hang (or exits 1 with the panic on
stderr if a top-level `await` is keeping the runtime alive).

The crate self-diagnoses the cause via an `eprintln!` left in
`crates/perry-ext-net/src/lib.rs:359`:

> `[perry-ext-net] BUG: spawn_socket_runner Handle::try_current returned Err — LTO has likely dead-stripped tokio's CONTEXT statics. This will panic on the subsequent Handle::current().`

`perry-ext-net` is a separate crate from `perry-stdlib`; each links its
own copy of tokio. LTO sees that `perry-ext-net`'s tokio CONTEXT
thread-locals aren't reachable from a static call graph (the runtime
context is entered over in `perry-stdlib`'s tokio copy) and strips
them. When the spawned socket task on a worker thread reaches into
`perry-ext-net`'s `Handle::try_current()`, it sees the stripped/null
CONTEXT and returns `Err`. The next `Handle::current()` panics with
*"there is no reactor running, must be called from the context of a
Tokio 1.x runtime"*.

Net effect: outbound TCP via `perry-ext-net` is unusable, blocking any
program that opens a socket — drivers (`@perryts/mongodb`, postgres,
mysql), HTTP clients, fastify HTTP server in any direction that
back-fetches, etc.

## Repro

`net-lto.ts`:

```ts
import * as net from 'net';

async function main() {
    console.log('createConnection...');
    const result = await new Promise<string>((resolve) => {
        const s = net.createConnection(27017 as never, '127.0.0.1' as never) as {
            on(ev: 'connect', cb: () => void): void;
            on(ev: 'error', cb: (e: unknown) => void): void;
            end(): void;
        };
        s.on('connect', () => { console.log('CONNECTED'); s.end(); resolve('ok'); });
        s.on('error', (e) => { console.log('ERROR: ' + String(e)); resolve('err'); });
    });
    console.log('done: ' + result);
}

main().catch((e) => console.error('ERR:', e));
```

Runs against any reachable TCP target — `27017` here is a local mongod
but anything listening works equivalently. The remote being closed
would normally just produce an `'error'` event; here it produces a
worker-thread panic instead.

[...]

PR #8337, opened 2026-08-18
fix(wasm-host): auto-provision wasm-host in all compile paths

jdalton:
## Problem

A program referencing `WebAssembly.*` did NOT reliably get the `wasm-host` cargo feature enabled on `perry-runtime`, causing link failures:

```
Undefined symbols for architecture arm64: "_js_webassembly_module_new"
```

This blocks sdxgen (whose acorn.wasm path needs it) and is the concrete instance of the `--enable-wasm-runtime` / wasm-host auto-detect gap.

## Root cause — three broken paths

1. **`--enable-wasm-runtime` did not enable the feature.** The flag only linked `libperry_wasm_host.a` (`run_pipeline.rs:5288` checked `args.enable_wasm_runtime`) but never enabled `perry-runtime/wasm-host` in the auto-optimize cargo invocation (`freshness.rs:192` only checked `ctx.needs_wasm_runtime`, NOT `args.enable_wasm_runtime`). So the runtime archive lacked `js_webassembly_*` symbol definitions while the host library was linked — the symbol stayed undefined.

2. **`PERRY_NO_AUTO_OPTIMIZE=1` never enabled `wasm-host`.** The no-auto path (`no_auto.rs`) only resolved prebuilt well-known ext archives. The prebuilt `libperry_runtime.a` is built WITHOUT `wasm-host` (deliberately kept out of `default` to avoid wasmi bloat on non-wasm programs). So even though `feature_detect.rs` set `ctx.needs_wasm_runtime = true`, the no-auto path never rebuilt the runtime with the feature.

3. **Dynamic `WebAssembly` access was not detected.** The static lowering (`module_static.rs`) sets `hir_module.uses_webassembly` for direct `WebAssembly.Module`/`instantiate`/etc. call sites, but a minified bundle can reach `WebAssembly` via dynamic property access (`const WA = WebAssembly; WA.Module(bytes)`, `globalThis.WebAssembly`) that lowers to an ordinary `PropertyGet` without hitting any set-site. The codegen still emits `js_webassembly_*` FFI calls for those paths, so the link died.

## Fix

1. **`--enable-wasm-runtime` override** (`run_pipeline.rs`): fold `args.enable_wasm_runtime` into `ctx.needs_wasm_runtime` before any feature/library provisioning so the flag triggers the same provisioning as auto-detection — the `wasm-host` cargo feature, the runtime rebuild, the library link, and the symbol-stub scan all see it.

2. **No-auto runtime rebuild** (`no_auto.rs`): when `ctx.needs_wasm_runtime` is set, build `perry-runtime-static` with default features + `perry-runtime/wasm-host` into a dedicated target dir (`target/perry-wasm-host-runtime`) so the prebuilt `libperry_runtime.a` is not clobbered. This is the same on-demand build pattern `build_missing_prebuilt_ext_lib` uses for CPU-only ext wrappers. `perry-wasm-host` has no tokio dep, so there is no #507 shared-tokio concern.

3. **Dynamic access fallback** (`feature_detect.rs`): add a fallback grep for `"WebAssembly"` in the HIR debug string, mirroring the fetch/crypto/EventEmitter fallbacks. Over-matching only over-links the wasm host (a size cost); the rule is zero false negatives.

## Verification

- `cargo test -p perry --bin perry -- compile::optimized_libs` — 34 passed (31 existing + 3 new regression tests).
- `cargo test -p perry-hir --lib` — 315 passed.
- New regression tests: wasm-usage → `wasm-host` cross-feature provisioned; non-wasm → not provisioned; wasm usage changes the cache key.

## Backward compat

- `--enable-wasm-runtime` still works (now actually enables the feature, not just the library).
- Non-wasm programs don't link the host — `wasm-host` stays out of `default`, the detection only fires on `WebAssembly.*` usage, and the no-auto rebuild only runs when `needs_wasm_runtime` is set.

## Summary by CodeRabbit

- **New Features**
  - Improved WebAssembly runtime support during compilation.
  - The compiler now detects WebAssembly usage more reliably and enables the required runtime automatically.
  - Explicit WebAssembly runtime options are consistently applied across compilation and linking.
  - When needed, the runtime is rebuilt with WebAssembly host support for the target platform.

- **Bug Fixes**
  - Improved fallback behavior when runtime sources, SDKs, or cross-compilation tools are unavailable, preserving the prebuilt runtime when possible.
  - Optimized runtime caching now distinguishes WebAssembly-enabled builds from standard builds.

### DIFF
--- a/crates/perry/src/commands/compile/collect_modules/feature_detect.rs
+++ b/crates/perry/src/commands/compile/collect_modules/feature_detect.rs
@@ -131,6 +131,32 @@ pub(super) fn detect_optional_feature_usage(
         ctx.needs_wasm_runtime = true;
     }
 
+    // Robust fallback for WebAssembly detection. The static lowering in
+    // `module_static.rs` sets `hir_module.uses_webassembly` for direct
+    // `WebAssembly.Module`/`instantiate`/etc. call sites, but a minified
+    // bundle can reach the `WebAssembly` global via dynamic property access
+    // (`const WA = WebAssembly; WA.Module(bytes)`, `globalThis.WebAssembly`,
+    // `globalThis["WebAssembly"]`) that lowers to an ordinary `PropertyGet`
+    // or `Ident` without hitting any set-site. The codegen still emits
+    // `js_webassembly_*` FFI calls for those paths, so without this fallback
+    // the `wasm-host` feature stays off and the link dies with
+    // `_js_webassembly_module_new` undefined. Mirror the fetch/crypto
+    // fallbacks above: scan the final HIR for the `WebAssembly` token.
+    // Over-matching only over-links the wasm host (a size cost); the rule
+    // is zero false negatives.
+    if !ctx.needs_wasm_runtime {
+        let hir_debug: String = format!(
+            "{:?}{:?}{:?}",
+            &hir_module.init, &hir_module.functions, &hir_module.classes
+        );
+        if hir_debug.contains("property: \"WebAssembly\"")
+            || hir_debug.contains("class_name: \"WebAssembly\"")
+            || hir_debug.contains("\"WebAssembly\"")
+        {
+            ctx.needs_wasm_runtime = true;
+        }
+    }
+
     // Detect crypto.* builtin usage (randomBytes/randomUUID/sha256/md5 used
     // without `import crypto`). The runtime symbols live behind the
     // perry-stdlib `crypto` Cargo feature, so we need to flip that on for
--- a/crates/perry/src/commands/compile/optimized_libs/no_auto.rs
+++ b/crates/perry/src/commands/compile/optimized_libs/no_auto.rs
@@ -5,8 +5,12 @@ use std::process::Command;
 
 use crate::OutputFormat;
 
+use super::super::library_search::{
+    android_cross_env, find_harmonyos_sdk, harmonyos_cross_env,
+};
 use super::super::{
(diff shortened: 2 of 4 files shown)


proggeramlug:
Merging — everything passes except `cargo fmt --all -- --check`, which flags
`optimized_libs/no_auto.rs` (lines 5, 89, 98, 134, 154) and
`optimized_libs/tests.rs:1299`. I confirmed it is this branch and not the merge
with #8338: clean `main` is fmt-clean (exit 0), and this branch alone on current
`main` still fails, including in `no_auto.rs`, which #8338 never touched.

Since this is a fork branch I can't push the fix here, so I'm landing the
formatting as an immediate follow-up rather than bouncing the PR for whitespace.

Validated otherwise: `optimized_libs` bin tests 36/36 (up from 33 — your three
new ones), `perry-runtime --lib` 2579, `perry-stdlib --lib` 124, and 49 of 50
gates including the compile tier (`cargo check --workspace --all-targets` under
`-D warnings` and `cargo clippy --workspace`).

One note for later, not a blocker: the new `WebAssembly` fallback builds a
`format!("{:?}")` of the whole HIR. That is consistent with the file rather than
new — `feature_detect.rs` already does this six separate times on `main` — but
it means a large module now gets its HIR formatted to a string seven times per
compile. On the Next.js fixture (104 modules, multi-MB IR) that is real time and
peak RSS. Computing `hir_debug` once at the top of `detect_optional_feature_usage`
and passing `&str` to each check would collapse all seven.

PR #8340, opened 2026-08-18
chore(make): drive make fix with cargo-fixit, pinned as a devDep

jdalton:
Apologies for the back-and-forth on this — earlier iterations removed cargo-fixit and then simplified to `cargo clippy --fix`; this is the simpler version we actually want.

`make fix` now runs `cargo fixit --clippy --workspace --all-targets --allow-dirty --allow-staged` — no `cargo clippy --fix` fallback. `cargo fixit` is faster than `cargo clippy --fix` on repeated runs because it skips the full re-check compile between fix rounds. It mirrors the fleet's `scripts/fleet/lint-rust.mts --fix`.

cargo-fixit is pinned as a devDep via a new `make fix-deps` target (`cargo install cargo-fixit@0.1.15 --locked`) — it ships no prebuilt `cargo binstall` binaries, so a from-source install is the pin. Run `make fix-deps` once to install it; `make fix` then uses it.

The CI clippy gate is unchanged (runs `cargo clippy`, not `make fix`).

## Summary by CodeRabbit

* **Chores**
  * Improved the automated code-fixing workflow for more consistent results across the project.
  * Added a setup command to install the required, pinned code-fixing tool with locked dependencies.
  * Updated development tooling configuration to ensure consistent code-quality fixes across supported project targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,25 @@
 # Dev-facing targets. Perry is plain cargo (no mise), so this is intentionally
 # minimal — add targets here only when a plain `cargo <verb>` isn't enough.
 
-# Apply clippy's machine-applicable autofixes across the workspace. Run on a
-# dirty tree is intended (--allow-dirty --allow-staged); review the diff before
-# committing. The CI clippy gate is unchanged.
+# Apply clippy's machine-applicable autofixes across the workspace via
+# cargo-fixit (crate-ci/cargo-fixit, pinned in external-tools.json under the
+# `cargo-install` release type) — the drop-in, faster replacement for
+# `cargo clippy --fix` on repeated runs, because it skips the full re-check
+# compile between fix rounds. There is deliberately no `cargo clippy --fix`
+# fallback: install cargo-fixit with `make fix-deps` first. Run on a dirty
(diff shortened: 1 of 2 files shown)


proggeramlug:
Merging — the functional change is an improvement (pinned version, an explicit
`fix-deps` install target, and no silent fallback, so a missing tool fails
loudly instead of quietly running the slow path).

One thing to flag, since it undoes a deliberate earlier fix rather than being
new: the comment re-introduces

```
# Mirrors the fleet's `scripts/fleet/lint-rust.mts --fix`.
```

`scripts/fleet/` does not exist in this repository — `git ls-tree -r origin/main
| grep scripts/fleet` returns nothing. #8322 removed exactly that line (and the
`scripts/fleet/_shared/rust-tool-pins.mts` pin reference) on the grounds that a
comment pointing at a path a reader cannot open is worse than no comment. The
version pin is now stated locally in `fix-deps`, which is the right home for it,
so the fleet reference is not carrying any information the file does not already
have.

Not worth another round trip on its own — worth deleting next time this file is
touched.

PR #8350, opened 2026-08-18
fix(release): build Linux GNU artifacts on glibc 2.31

### LINKED ISSUE #6351: Build the Linux glibc artifacts against an older glibc (native fix for #6298)
Follow-up to #6298 (fixed there at the *distribution* layer, not the build layer).

## Where we are

PR for #6298 teaches the npm launcher and `install.sh` to route hosts whose glibc is older than the builder's onto the fully-static musl build. That unblocks Ubuntu 22.04, Debian 12, RHEL 9 and Amazon Linux 2023 — but it is a workaround: those users get a binary that cannot build `perry/ui` (GTK4) apps, and on npm they have to run one `--force` install by hand because npm's `libc` selector has no notion of a glibc *version*.

The real fix is to build the Linux glibc artifacts against an older glibc, so the default `@perryts/perry-linux-{x64,arm64}` package just works on every supported distro.

## Why it isn't a one-line matrix change

The glibc legs build on `ubuntu-24.04` (glibc 2.39). They used to build on `ubuntu-22.04` (glibc 2.35) and were moved in v0.5.860 (`7e6629cc5`) for one reason: **`perry-ui-gtk4`'s system dependencies are noble-only.**

- `libshumate-dev` — MapView (#517). Not in jammy.
- `libwebkitgtk-6.0-dev` — WebView (#658). Jammy only has webkit2gtk-4.0/4.1 (GTK3-era); the GTK4 webkit port isn't packaged.

So any move to an older base image requires `perry-ui-gtk4` to build without those two widgets.

## Options

1. **Feature-gate `perry-ui-gtk4`** (cargo features for MapView + WebView), move the glibc legs to `ubuntu-22.04`.
   - Floor becomes glibc 2.35 → fixes Ubuntu 22.04 and Debian 12.
   - Still leaves RHEL 9 / Alma / Rocky / Amazon Linux 2023 (glibc **2.34**) broken.
   - Costs MapView + WebView on Linux desktop.
2. **Build in an old-sysroot container** (manylinux_2_28 / Rocky 8 → glibc 2.28), which covers every distro above.
   - The compiler side is genuinely feasible: Perry has **no system-LLVM build dependency** (no `llvm-sys`/`inkwell` in the tree — it shells out to `clang` at *runtime*), so the CLI + runtime + stdlib should build in an old container with just a rustup toolchain.
   - Same `perry-ui-gtk4` blocker; the container has no GTK4 at all, so the UI lib would need its own leg or a feature gate.
3. **A separate old-glibc leg + a third npm package** — rejected. npm cannot select by glibc version, so that package could only ever be chosen by the launcher logic from #6298 anyway, while doubling the slowest part of the release matrix.

Whatever we pick, the launcher's glibc-version routing stays: npm's `libc` field cannot express "glibc ≥ 2.39", so *something* has to decide at runtime.

## Also in scope

The apt `.deb` is built from the `perry-linux-x86_64` / `-aarch64` glibc artifacts, so `apt install perry` is broken below glibc 2.39 and a package can't fall back at runtime. Until the build leg exists it should at least declare `Depends: libc6 (>= 2.39)` so it fails at install time with a comprehensible message instead of at first run.

proggeramlug:
Closes #6351.

## Summary

- build the Linux GNU compiler, runtime, stdlib, abort runtime, and extension archives in pinned architecture-native Debian 11 containers
- keep GTK4 on the Ubuntu 24.04 host, then combine the old-glibc core with the GTK archive during packaging
- lower launcher/install routing and Debian package metadata to a glibc 2.31 floor
- verify the final compiler's imported GLIBC versions and compile/run a TypeScript smoke program inside the old sysroot
- isolate old-sysroot Cargo targets so cached Ubuntu 24.04 dependencies cannot leak a newer glibc requirement

The current in-process backend requires LLVM 22 development libraries. apt.llvm.org does not publish LLVM 22 packages for Debian 10/glibc 2.28, but does for Debian 11/glibc 2.31. This is therefore the oldest practical builder without disabling the default backend, and it covers every affected distribution named in #6351: Ubuntu 22.04, Debian 12, RHEL 9, and Amazon Linux 2023 (plus Ubuntu 20.04 and Debian 11). Older systems keep the existing static-musl fallback.

## Verification

- full ARM64 dist compiler built against glibc 2.31 and LLVM 22; final compiler's highest imported version is GLIBC_2.30
- node npm/perry/test/detect.test.cjs
- sh -n packaging/install.sh
- bash -n and ShellCheck for the old-glibc build script
- actionlint for both modified workflows
- scripts/check_file_size.sh
- git diff --check

No version bump is included.

## Summary by CodeRabbit

* **New Features**
  * Linux glibc packages now support systems with glibc 2.31 or newer, expanding compatibility across older distributions.
  * The installer automatically falls back to the static musl package on older glibc systems.
* **Documentation**
  * Updated installation and release guidance with supported distributions, compatibility requirements, fallback behavior, and GTK4 system dependencies.
* **Bug Fixes**
  * Improved compatibility messaging to clearly identify the required glibc version and installation options.

### DIFF
--- a/.github/workflows/npm-launcher.yml
+++ b/.github/workflows/npm-launcher.yml
@@ -44,12 +44,14 @@ jobs:
       - name: install.sh is valid POSIX sh
         run: sh -n packaging/install.sh
 
-  # End-to-end on the platform from #6298: Ubuntu 22.04 = glibc 2.35, older than
-  # the glibc 2.39 the release matrix builds the linux-gnu binaries against.
-  # Installs the *published* platform packages and runs this repo's launcher
-  # against them.
+  # Published-package verification for the platform from #6298. This is manual
+  # because a PR changes the launcher before the matching old-glibc binary can
+  # exist in npm; release-packages.yml exercises the just-built artifact inside
+  # the glibc 2.31 build container before publication. Run this after a release to verify that
+  # the registry package uses the native glibc build on Ubuntu 22.04.
   ubuntu-2204:
-    name: Ubuntu 22.04 (glibc 2.35) end-to-end
+    if: github.event_name == 'workflow_dispatch'
+    name: Ubuntu 22.04 native glibc end-to-end
     runs-on: ubuntu-22.04
     timeout-minutes: 20
     steps:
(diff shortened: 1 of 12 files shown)


proggeramlug:
Merging. The verification story is stronger than the summary suggests, and worth
naming since it is the part that makes this trustworthy:

- `scripts/build_linux_glibc_2_31.sh` asserts the container really is the 2.31
  sysroot (`expected the glibc 2.31 sysroot, found glibc $glibc_version`) and
  that LLVM 22 is present — so swapping the base image fails loudly instead of
  silently building against a newer floor.
- It then reads the **actual symbol versions** back out of the produced
  compiler (`readelf --version-info` → max `GLIBC_x.y`) and `exit 1`s if
  anything newer than 2.31 is required. That is a real floor gate, not just
  "we built it in a container".
- And it exercises the result end to end, asserting exact output
  (`test "$(…/hello)" = "old-glibc 2,4,6"`) rather than merely that the binary ran.

Also checked: the image is SHA-pinned, the `old_glibc_image == ''` / `!= ''`
gating is symmetric so exactly one path runs per leg, and actionlint reports the
same four shellcheck findings on `main` as on this branch (same issues, shifted
line numbers) — nothing new. All 48 script-tier gates pass.

One observation, not a blocker: the extension loop swallows failures —

```
cargo build … -p "$package" || echo "  (skipped $package -- failed to build on the glibc 2.31 sysroot)"
```

and packaging then uses `[ -f "$archive" ] && cp`. So an ext archive that stops
building on the old sysroot disappears from the Linux GNU package silently. The
comment says this deliberately matches the ordinary release leg's best-effort
behaviour, so it is consistent rather than new — but it does mean the one place
that would notice is a human reading the `::group::` output. If these are meant
to be optional, an explicit expected-set check at the end would make a
regression visible; if they are not optional, the `||` should go.

Issue #8185, opened 2026-08-15
gc: a deleted write barrier passes every runtime probe — only a static IR assertion can catch it


proggeramlug:
## A deleted GC write barrier passes every runtime probe we have

Found while validating #8183. A **release** build with the write barrier deleted from the dynamic-key IC's new reference-store arm passes the entire behavioural matrix:

- old→young edge fixtures
- `PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1`
- `PERRY_GEN_GC=0`
- `PERRY_GC_ZEAL=1` + from-space protection at **depth 400**

Byte-identical output, exit 0, on both the gap fixture and a larger adversarial one. **The static IR test was the only thing that said no.**

## Why this matters more than one PR

CLAUDE.md's rooting-invariant guidance says the runtime instruments are the detector for the class the static checker cannot see (a runtime-side cache of a heap pointer). This is the mirror image and it is not written down anywhere: for a **missing write barrier**, the runtime instruments are the ones that cannot see it, and the static IR assertion is the only detector.

The reason is structural. A missing barrier does not corrupt anything at the moment of the store. It leaves a remembered-set entry unwritten, so a *later* minor collection fails to trace an old→young edge and drops a live object. Whether that happens depends on the object surviving to old-gen, the child staying young, and a minor landing in that window — none of which a short fixture reliably produces, and `FORCE_EVACUATE`/`VERIFY_EVACUATION` do not manufacture it either because they verify *rewriting*, not *remembering*.

This is the same shape as CLAUDE.md's hazard 4 — "the gate runs but its subject never did" — except the subject here is a negative: the absence of a barrier cannot be observed by running the program.

## Consequences worth acting on

1. **Any PR that adds or moves a store on a GC-managed slot needs a static IR assertion**, not a behavioural test, as its barrier evidence. #8183 has one (`br i1` into the arm, plus barrier/layout-note/addref presence); it should be the template.
2. **A green GC-canary run is not evidence a barrier is present.** Several PRs in this campaign have cited `FORCE_EVACUATE + VERIFY_EVACUATION` as barrier evidence. That inference is invalid — it is evidence about rewriting, which is a different property.
3. Worth checking whether `scripts/gc_store_site_inventory.py` can be extended to assert barrier presence per store site, rather than only inventorying the sites. If it can, this becomes a gate instead of a review convention.

## Provenance

[...]

PR #8363, opened 2026-08-18
perf(codegen): inline mixed-layout constructor stores

proggeramlug:
## Summary

- extend constructor-free construction to runtime-declared mixed layouts and stamped outlined allocations
- recognize typed `numberParam + finiteLiteral` field initializers, covering the eight derived numeric fields in `tree_wide`
- preserve pointer-store generation and incremental-mark barriers on the direct-store arm
- add the exact issue benchmark fixture and IR regression coverage

Refs #8289.

## Performance

Five interleaved runs of `benchmarks/issue-8289/tree_wide.ts` on macOS arm64, comparing `origin/main` with this branch (median):

| Metric | main | PR | Change |
|---|---:|---:|---:|
| Retired instructions | 23,412,462,462 | 7,622,290,049 | -67.44% (3.07x) |
| Wall time | 6.39 s | 1.80 s | -71.83% (3.55x) |
| Peak RSS | 63,504,384 B | 63,537,152 B | +0.052% |

The fixture output remains `20971480`. The `churn` comparison remains instruction-neutral within measurement noise (4,651,981,862 to 4,648,010,989 instructions, -0.09%).

## Validation

- `cargo test --profile perry-dev -p perry-codegen --lib` (1,097 passed)
- `cargo check --profile perry-dev -p perry-codegen --lib`
- exact fixture output under forced evacuation, evacuation verification, and protected/poisoned from-space

No version bump is included.

## Summary by CodeRabbit

* **Performance**
  * Improved object construction performance for mixed pointer and numeric field layouts.
  * Reduced unnecessary constructor work when allocation metadata is ready, with measured instruction reductions on wide-tree workloads.
* **Reliability**
  * Added support for numeric fields initialized with finite offsets.
  * Preserved pointer write barriers and validated layout readiness before applying optimized field stores.
* **Testing**
  * Expanded coverage for outlined allocations, pointer-bearing layouts, numeric offsets, and wide-tree benchmarks.

### DIFF
--- a/benchmarks/issue-8289/tree_wide.ts
+++ b/benchmarks/issue-8289/tree_wide.ts
@@ -1,0 +1,32 @@
+class Tree {
+  left: Tree | null;
+  right: Tree | null;
+  a: number; b: number; c: number; d: number;
+  e: number; f: number; g: number; h: number;
+  constructor(left: Tree | null, right: Tree | null, s: number) {
+    this.left = left;
+    this.right = right;
+    this.a = s; this.b = s + 1; this.c = s + 2; this.d = s + 3;
+    this.e = s + 4; this.f = s + 5; this.g = s + 6; this.h = s + 7;
+  }
+}
+
+function build(depth: number, s: number): Tree {
+  if (depth === 0) return new Tree(null, null, s);
+  return new Tree(build(depth - 1, s), build(depth - 1, s + 1), s);
+}
+
+function count(t: Tree): number {
+  if (t.left === null) return 1;
+  return 1 + count(t.left) + count(t.right as Tree);
+}
+
+function main(): void {
+  let total = 0;
+  for (let i = 0; i < 40; i++) {
+    const t = build(18, i);
+    total += count(t);
+  }
+  console.log(total);
+}
+main();
--- a/changelog.d/8363-mixed-layout-constructor-stores.md
+++ b/changelog.d/8363-mixed-layout-constructor-stores.md
@@ -1,0 +1,6 @@
+### Performance
+
+- Skip prologue-only constructors for mixed pointer/raw-f64 layouts, including
+  typed `numberParam + finiteLiteral` field initialization, while retaining
+  pointer write barriers. On issue #8289's `tree_wide` fixture this reduces
+  retired instructions by 67.4% (23.41B to 7.62B) with unchanged output.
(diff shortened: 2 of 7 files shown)


proggeramlug:
Holding — `gc_store_site_inventory` fails, and the second half of it is the
interesting one. Validated in a batch with three disjoint PRs; #8361 and #8362
are clean and merged, and this failure names only your file.

```
crates/perry-codegen/src/lower_call/new.rs:748: raw generated heap/global store:
  blk.store(DOUBLE, &prologue_values[store_index], &field_ptr);

GC store-site CLAIM verification failed (#8185):
  stem 'ctor_prologue' is emitted by codegen but has NO IR witness in
  crates/perry-codegen/src/expr/barrier_stem_census_tests.rs — add it to
  VERIFIED_BARRIER_STEMS with a probe that reaches the tier, or the barrier can
  be deleted without any gate noticing (#8185)
```

Two separate asks:

1. **The store at `new.rs:748` needs a `GC_STORE_AUDIT` marker** naming its
   class — `BARRIERED`, `EXTERNAL_BARRIERED`, `INIT`, `POINTER_FREE`, `ROOT` or
   `STACK` — with the reason. For an inlined constructor prologue writing into a
   freshly allocated object, `INIT` is plausible, but that is your call to make
   and state, not mine to guess: the whole point of the marker is that a reader
   can check the claim.

2. **`ctor_prologue` needs an entry in `VERIFIED_BARRIER_STEMS`** with a probe
   that actually reaches the tier. This is the part worth not shortcutting — the
   gate is telling you the new barrier has no IR witness, i.e. it could be
   deleted and every gate would stay green. That is the exact failure shape
   #8185 exists to prevent, and it is the same class as CLAUDE.md's "a gate must
   assert its subject was live".

Everything else passed in the batch: `perry-runtime --lib` 2585,
`perry-codegen --lib` **1098** (up from 1096, your new tests), `perry-hir` all
suites, and the other 48 gates including the compile tier.

proggeramlug:
Both gate items are resolved and verified; merging, with the fix landing as an
immediate follow-up since this is a fork branch I can't push to.

**1. `GC_STORE_AUDIT(INIT)` at `new.rs`.** `INIT` is the right class here and the
surrounding code says why: the store writes into `fields_base` of a
freshly-allocated, not-yet-published object, and pointer-bearing stores still get
`emit_write_barrier_slot_value_and_generation_tested` immediately after
(`if !store.requires_raw_f64`). So the raw store really is initialization, not an
unbarriered mutation.

**2. `ctor_prologue` witness — and I checked it isn't vacuous.** The probe reuses
`class_field_barrier_tests::ir()`, which is exactly the shape that can register a
stem without actually reaching the tier. So I sabotage-tested it: repointed
`ctor_prologue_ir()` at stem-free IR and re-ran
`census_every_registered_stem_has_a_live_verified_witness`:

```
with the real probe   → ok
with stem-free IR     → FAILED (barrier_stem_census_tests.rs:529)
```

That is the property #8185 is about — the witness fails when the stem stops being
emitted, so the barrier can no longer be deleted silently.

All 50 gates pass, `perry-codegen --lib` 1098, and `gc_store_site_inventory.py`
exits 0.

PR #8364, opened 2026-08-18
fix(intl): implement Segmenter containing

proggeramlug:
## Summary

Implements the ECMA-402 `Segments.prototype.containing()` behavior for `Intl.Segmenter`, including UTF-16 indexing, receiver branding, index coercion, and exact WTF-8 lone-surrogate preservation.

## Changes

- Expose a branded, non-constructable `containing()` builtin on Segmenter results while retaining their iterable backing.
- Apply `ToIntegerOrInfinity` semantics and return fresh segment data records for matching UTF-16 indices.
- Preserve exact WTF-8 input and segment slices, and reject Symbol input during string conversion.
- Add a parity fixture covering boundaries, coercion, metadata, branding, and lone surrogates.

## Related issue

Refs #5896

## Test plan

- [x] `CARGO_PROFILE_RELEASE_LTO=false CARGO_PROFILE_RELEASE_CODEGEN_UNITS=16 CARGO_PROFILE_RELEASE_INCREMENTAL=true cargo build --release -p perry -p perry-runtime-static -p perry-stdlib-static`
- [x] `scripts/test262_subset.py --root vendor/test262 --dir intl402/Segmenter/prototype/segment/containing --jobs 4 --timeout 300` (12 passed, 0 failed)
- [x] `scripts/test262_subset.py --root vendor/test262 --dir intl402/Segmenter --jobs 8 --timeout 300` (70 passed; only the 2 pre-existing runtime failures and 2 parallel compile timeouts remain)
- [x] `PERRY_SKIP_BUILD=1 PERRY_BIN=target/release/perry PERRY_RUNTIME_DIR=target/release ./run_parity_tests.sh --filter test_gap_intl_segmenter_containing_5896`
- [x] `bash scripts/check_file_size.sh`
- [ ] Full workspace test suite not run. `cargo test --release -p perry-runtime --lib` completed 2,589 tests with 2 unrelated order-sensitive failures; both passed when rerun individually with one test thread.
- [ ] (if CLI / stdlib / runtime API changed) Documentation is not applicable to this standards-conformance fix.
- [ ] (if touching a platform UI backend) Not applicable.

## Screenshots / output

Not applicable.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for `Intl.Segmenter`’s `Segments.prototype.containing()` lookup.
  * Supports UTF-16 index handling, numeric coercion, and matching segment records.
  * Preserves lone surrogates in returned segment data.

* **Bug Fixes**
  * Invalid receivers and symbol inputs now produce appropriate errors.
  * Out-of-range or invalid indices return `undefined`.

### DIFF
--- a/changelog.d/8364-intl-segmenter-containing.md
+++ b/changelog.d/8364-intl-segmenter-containing.md
@@ -1,0 +1,3 @@
+**Implement `Intl.Segmenter` containing lookup (#8364):** Segmenter results now
+expose the ECMA-402 `containing()` method with UTF-16 index coercion, receiver
+branding, and exact lone-surrogate preservation.
--- a/crates/perry-runtime/src/intl/segmenter.rs
+++ b/crates/perry-runtime/src/intl/segmenter.rs
@@ -7,6 +7,10 @@ use crate::value::js_nanbox_pointer;
 #[cfg(feature = "intl-segmenter")]
 use unicode_segmentation::UnicodeSegmentation;
 
+const KEY_SEGMENTS_BRAND: &str = "__intlSegmentsBrand";
+const KEY_SEGMENTS_LENGTH: &str = "__intlSegmentsLength";
+const SEGMENTS_BRAND: &str = "Segments";
+
 pub(crate) fn normalize_granularity(value: Option<String>) -> String {
     match value.as_deref() {
         None | Some("grapheme") => "grapheme".to_string(),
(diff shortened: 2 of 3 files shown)


proggeramlug:
Holding — `raw_handle_debt` fails on this branch:

```
bare raw-handle reads: 988 (baseline 978)
  crates/perry-runtime/src/intl/segmenter.rs: 10 bare read(s) in a module with
  no ceiling. New code must use RuntimeHandle::across_{mut,const,nanbox} or
  with_{mut,const}_ptr; see #7341.
```

I checked the baseline is current rather than stale (978 is what `main` reports
today), and isolated it: validated in a batch with #8361/#8362/#8363, and
#8361+#8362 alone pass this gate — the ten reads are all in your new
`segmenter.rs`.

"A module with no ceiling" means `intl/segmenter.rs` is locked at zero, so new
code there is expected to use the rooted forms from the start rather than adding
debt to be paid down later. Which form depends on each site:

- `RuntimeHandle::across_{mut,const,nanbox}` for a **post-call reload** — where
  the pointer is needed again after something that can collect.
- `with_{mut,const}_ptr` for a **scoped argument** to a non-allocating operation
  or a self-rooting runtime entry point.

That distinction needs to be made per site by someone who knows which of the ten
cross a collection point, so I'd rather hand it back than guess — a wrong choice
here is the silent kind that surfaces cycles later as
`TypeError: value is not a function`.

Everything else passed: `perry-runtime --lib` 2585, `perry-codegen --lib` 1098,
`perry-hir` all suites, and the other 48 gates including the compile tier.

proggeramlug:
Merging, with the ratchet conversion landing as an immediate follow-up (fork
branch, so I can't push here).

The only gate objection was `raw_handle_debt`:

```
crates/perry-runtime/src/intl/segmenter.rs: 10 bare read(s) in a module with no ceiling
```

`intl/segmenter.rs` is a new module, so it is locked at zero. The conversion
splits the way the code actually reads: **9 of the 10 sites are argument
position** — the pointer is passed straight into a call — so they take
`with_const_ptr`. Only one is a genuine post-call reload and takes
`across_const`, and it is the right one:

```rust
// ToIntegerOrInfinity may invoke user code, so keep the Segments backing
// array rooted while coercing the index.
let (number, segments) = segments_handle.across_const::<ArrayHeader, _>(|| {
    list_relative_plural::to_number_reject_bigint(index)
});
```

That distinction is the whole risk here — `across_const` cannot express an
argument-position read, and using it where `with_const_ptr` belongs would hoist
a pointer above a collection point while the gate went green.

Verified independently: `raw_handle_debt` back to **978/978** with every module
inside its ceiling, `perry-runtime --lib` 2590 passed, and all 50 gates
including the compile tier.

PR #8366, opened 2026-08-18
fix(runtime): honor Array.prototype.pop side effects

proggeramlug:
## Summary

- preserve the allocation-free dense `Array.prototype.pop` fast path
- use ordinary indexed `Get`/`DeletePropertyOrThrow` semantics when descriptors, sparse storage, or indexed prototypes make the receiver exotic
- root the receiver and returned value across accessors, then revalidate frozen/non-writable length state after getter side effects
- add end-to-end regression coverage for the three selected Test262 cases

## Test262

Pinned corpus: `4249661388e5d3f92a85186213da140a6481490f`

- exact worklist: 3 pass, 0 runtime failures, 0 compile failures
- full `built-ins/Array` base: 2488 pass, 39 runtime failures, 0 compile failures
- full `built-ins/Array` patched: 2491 pass, 36 runtime failures, 0 compile failures
- only removed failures:
  - `prototype/pop/S15.4.4.6_A4_T1.js`
  - `prototype/pop/set-length-array-is-frozen.js`
  - `prototype/pop/set-length-array-length-is-non-writable.js`
- no new failures

## Validation

- `cargo test -p perry --test issue_5898_array_pop_prototype -- --nocapture`
- no version, lockfile, CLAUDE, or changelog changes

Refs #5898

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved `Array.prototype.pop` behavior for arrays with inherited properties and accessor side effects.
  - Correctly reports errors when indexed deletion fails or when array length becomes frozen or read-only during the operation.
  - Preserved optimized behavior for ordinary arrays.

- **Tests**
  - Added regression coverage for prototype lookups, getter-triggered mutations, and expected `TypeError` behavior.

### DIFF
--- a/crates/perry-runtime/src/array/push_pop.rs
+++ b/crates/perry-runtime/src/array/push_pop.rs
@@ -35,6 +35,13 @@ fn throw_non_writable_length() -> ! {
     );
 }
 
+#[cold]
+fn throw_cannot_delete_array_index(index: u32) -> ! {
+    crate::collection_iter::throw_type_error(&format!(
+        "Cannot delete property '{index}' of [object Array]"
+    ));
+}
+
 /// Install an array-growth forwarding stub after `tracked_header_for` proves
 /// allocator ownership of `old_user_addr`. Keeping the classifier injectable
 /// makes the below-2-TiB macOS case deterministic without mapping a fixed low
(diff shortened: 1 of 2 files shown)


proggeramlug:
Merging, with the ratchet conversion landing as an immediate follow-up (fork
branch, so I can't push here).

Worth saying explicitly: **the rooting semantics in this PR were already
correct.** You open a `RuntimeHandleScope`, root the array, and re-resolve the
receiver after each observable operation — which is the hard part. The gate was
only objecting to the *spelling*: it counts `handle.get_raw_mut_ptr::<T>()` and
wants the #7341 wrappers.

```
crates/perry-runtime/src/array/push_pop.rs: 6 bare reads exceeds its ceiling of 4
```

The conversion is mechanical and preserves your ordering exactly — both sites
were post-call reloads, so both become `across_mut`, which runs the closure and
*then* reloads:

- after `array_spec_get` (can run an accessor that allocates/moves `arr`)
- after the conditional `js_array_delete`

`clean_arr_ptr_mut(...)` stays at the same two points. Verified independently
rather than taken on trust: `raw_handle_debt` back to **978/978** with every
module inside its ceiling, `perry-runtime --lib` 2590 passed, and all 50 gates.

PR #8373, opened 2026-08-18
perf(codegen): elide proven churn array scans

proggeramlug:
## Summary

Closes the `churn` row of #8289. Elides provably-redundant array scans in the
element-shape loop.

Authored by a codex agent working the issue; it could not reach the GitHub API
from its sandbox, so I am opening the PR from its pushed branch and auditing it
as I would any other.

## Measurement

Re-measured on current `main` rather than against the issue's table, because the
issue explicitly sequenced #8288 first — its allocation-bitmap regression
contaminated this row — and **#8288 closed on 2026-08-17**. The tabulated
1.98× / 5.006 B figures therefore predate that fix and would have been the wrong
baseline to optimise against.

| | retired instructions |
|---|---:|
| before | 4,645,492,850 |
| after | **2,656,776,895** |
| delta | **−42.81 %** |
| vs Node | **0.987×** — now marginally faster |

Outputs unchanged across seven runs.

## Validation

Claimed by the agent and being independently re-run before merge: full codegen
test suite, and `scripts/run_lint_gates.sh` (50 gates, including the compile
tier).

## Summary by CodeRabbit

  * Improved optimization of loops processing arrays with statically known element types.
  * Reduced unnecessary runtime type checks for safely proven array contents.
  * Preserved fallback behavior when array contents cannot be verified.

* **Bug Fixes**
  * Improved handling of direct field reads from array elements.
  * Invalid property accesses now correctly prevent unsafe optimizations.

* **Tests**
  * Added regression coverage for optimized array loops and field access validation.

### DIFF
--- a/benchmarks/issue-8289/churn.ts
+++ b/benchmarks/issue-8289/churn.ts
@@ -1,0 +1,23 @@
+type Node = { v: number; w: number };
+
+function chunk(base: number): number {
+  const keep: Node[] = [];
+  for (let j = 0; j < 1000; j++) {
+    keep.push({ v: base + j, w: j });
+  }
+  let acc = 0;
+  for (let j = 0; j < keep.length; j++) {
+    acc += keep[j].v + keep[j].w;
+  }
+  return acc;
+}
+
+function main(): void {
+  let total = 0;
+  for (let c = 0; c < 20000; c++) {
+    total += chunk(c * 1000);
+  }
+  console.log(total);
+}
+
+main();
--- a/crates/perry-codegen/src/collectors/hir_facts.rs
+++ b/crates/perry-codegen/src/collectors/hir_facts.rs
@@ -111,6 +111,11 @@ pub(crate) struct ArrayFacts {
     /// why this fact governs *profitability* rather than the soundness of the
     /// elided per-store note (which the emitted header test owns).
     pub all_pointer_element_locals: HashSet<u32>,
+    /// Array binding -> exact element class from the E1--E5 containment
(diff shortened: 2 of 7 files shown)


proggeramlug:
Independently corroborated, merging.

I could not cheaply reproduce the `xctrace` retired-instruction counts, so rather
than accept the agent's numbers I ran a wall-clock A/B and am reporting it as
wall time, not instructions:

| bench | clean `main` | with #8373+#8374 | wall Δ | claimed (instructions) |
|---|---:|---:|---:|---:|
| `churn` | 0.345 s | **0.221 s** | **−36.0 %** | −42.81 % |
| `pipeline` | 0.188 s | **0.134 s** | **−28.6 %** | −27.69 % |

Best-of-5 each, on a quiet host, with **stdout compared on every run** — byte
identical in both arms, so the speedup is not coming from computing something
different. Wall time will not track retired instructions exactly (not all of the
row is instruction-bound), but the direction and magnitude line up.

Two guards worth naming, because both have produced fake results here before:
I built a **separate clean-`main` baseline** rather than reusing an existing
binary, and I confirmed the two `perry` binaries actually differ
(`dbd9af26…` vs `70c6e39e…`) — a stale-build A/B reports a comfortable zero and
looks like a passing check.

Correctness: `perry-codegen --lib` **1102** passed (up from 1098 — the new
element-shape and compare tests), `perry-runtime --lib` 2590, and all 50 gates
including the compile tier.

#8289 stays open: `pipeline` is now 1.84× Node, still the widest of the four rows.

PR #8374, opened 2026-08-18
perf(codegen): inline generic string key equality

proggeramlug:
## Summary

Progresses the `pipeline` row of #8289 by inlining generic string-key equality.

Authored by a codex agent working the issue; it could not reach the GitHub API
from its sandbox, so I am opening the PR from its pushed branch and auditing it
as I would any other.

## Measurement

| | retired instructions |
|---|---:|
| before | 3,180,075,117 |
| after | **2,299,626,226** |
| delta | **−27.69 %** |
| vs Node | 1.84× (was 2.15× in the issue table) |

Outputs unchanged across seven runs.

For context on why this row needed more work: #8351's four-entry polymorphic
closure-dispatch cache took only −3.23 %, so the bulk of the gap was still
unexplained. This is a second, larger bite; `pipeline` remains the widest
remaining gap of the four rows, so #8289 should stay open.

## Validation

Claimed by the agent and being independently re-run before merge: full codegen
test suite, and `scripts/run_lint_gates.sh` (50 gates, including the compile
tier).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved strict string equality checks for faster and more reliable comparisons.
  * Preserved compatibility for fallback string comparison scenarios.

* **Tests**
  * Added regression coverage for strict equality checks involving strings and generic keys.

### DIFF
--- a/crates/perry-codegen/src/expr/compare.rs
+++ b/crates/perry-codegen/src/expr/compare.rs
@@ -786,14 +786,17 @@ pub(crate) fn lower(ctx: &mut FnCtx<'_>, expr: &Expr) -> Result<String> {
                 // the wrapper first. Strict `=== "lit"` is unaffected (both sides are
                 // real strings at runtime). #boxed-loose-eq.
                 if one_side_string && matches!(op, CompareOp::Eq | CompareOp::Ne) {
+                    // Reuse the no-literal string prefix instead of sending
+                    // every comparison through two unified-unbox calls. This
+                    // is the hot shape of a specialized generic container:
+                    // `this.keys[i]` has a concrete string type while `k`
+                    // retains its `K` spelling. Identical pooled strings now
+                    // leave after one bit compare; distinct heap strings call
+                    // only `js_string_equals`. The boxed arm is byte-for-byte
+                    // the old helper composition, so a lying annotation keeps
+                    // its existing behaviour.
+                    let i32_eq = lower_string_strict_eq_inline(ctx, &l, &r, true);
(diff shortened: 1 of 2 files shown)


proggeramlug:
Independently corroborated, merging.

I could not cheaply reproduce the `xctrace` retired-instruction counts, so rather
than accept the agent's numbers I ran a wall-clock A/B and am reporting it as
wall time, not instructions:

| bench | clean `main` | with #8373+#8374 | wall Δ | claimed (instructions) |
|---|---:|---:|---:|---:|
| `churn` | 0.345 s | **0.221 s** | **−36.0 %** | −42.81 % |
| `pipeline` | 0.188 s | **0.134 s** | **−28.6 %** | −27.69 % |

Best-of-5 each, on a quiet host, with **stdout compared on every run** — byte
identical in both arms, so the speedup is not coming from computing something
different. Wall time will not track retired instructions exactly (not all of the
row is instruction-bound), but the direction and magnitude line up.

Two guards worth naming, because both have produced fake results here before:
I built a **separate clean-`main` baseline** rather than reusing an existing
binary, and I confirmed the two `perry` binaries actually differ
(`dbd9af26…` vs `70c6e39e…`) — a stale-build A/B reports a comfortable zero and
looks like a passing check.

Correctness: `perry-codegen --lib` **1102** passed (up from 1098 — the new
element-shape and compare tests), `perry-runtime --lib` 2590, and all 50 gates
including the compile tier.

#8289 stays open: `pipeline` is now 1.84× Node, still the widest of the four rows.

PR #8377, opened 2026-08-18
fix(test): make Next dylib provider build cwd-independent

### LINKED ISSUE #8258: [test] test_next_app_route_dylib.sh is silently cwd-sensitive: run from outside the repo, the provider runtime loses -C force-unwind-tables and fails 25 minutes later naming the wrong cause

proggeramlug:
## Summary

- run the Next App Route provider Cargo build from the repository root
- ensure `.cargo/config.toml` is discovered when the gate is invoked outside the checkout
- preserve the fixture-root working directory used later by the production route host

Closes #8258

## Validation

- `bash -n tests/test_next_app_route_dylib.sh`
- `shellcheck tests/test_next_app_route_dylib.sh`
- invoked a Cargo check from `/tmp` through the repository-root boundary and confirmed rustc receives `-C force-unwind-tables=yes`
- `BASE_SHA=origin/main ./scripts/run_lint_gates.sh` (all 50 gates passed)

## Checklist

- [x] No `CLAUDE.md` or `CHANGELOG.md` edit

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Next App Route dylib builds so repository-level Cargo configuration is found reliably.
  - Improved build consistency when invoked from outside the project checkout.

- **Documentation**
  - Added a changelog entry describing the corrected build behavior.

### DIFF
--- a/changelog.d/8377-next-app-route-dylib-cwd.md
+++ b/changelog.d/8377-next-app-route-dylib-cwd.md
@@ -1,0 +1,3 @@
+The Next App Route dylib gate now builds its provider runtime from the
+repository root, so Cargo discovers Perry's required unwind-table configuration
+even when the gate is invoked from outside the checkout.
--- a/tests/test_next_app_route_dylib.sh
+++ b/tests/test_next_app_route_dylib.sh
@@ -236,15 +236,22 @@ perl -0pi -e 's/("crates\/perry-runtime",\n)/$1    "crates\/next-app-route-provi
 
 runtime_build_library="$provider_target_dir/$profile/libperry_runtime.$library_extension"
 provider_linker="$fixture/provider-linker.sh"
-env \
-    CARGO_TARGET_DIR="$provider_target_dir" \
-    PERRY_NEXT_RUNTIME_LIBRARY="$runtime_build_library" \
-    PERRY_NEXT_REQUIRED_SYMBOLS="$required_symbols" \
-    PERRY_NEXT_REAL_CC="$real_cc" \
-    "$cargo_linker_env=$provider_linker" \
-    cargo build --manifest-path "$runtime_source/Cargo.toml" \
-        --profile "$profile" --jobs "$cargo_jobs" \
-        -p perry-runtime -p next-app-route-provider
+(
+    # Cargo discovers .cargo/config.toml from the invocation directory, not
+    # from --manifest-path. Build from the repository root so the provider
+    # runtime always inherits the required force-unwind-tables rustflag, even
+    # when this gate is invoked from another directory.
+    cd "$repo_root"
+    env \
+        CARGO_TARGET_DIR="$provider_target_dir" \
+        PERRY_NEXT_RUNTIME_LIBRARY="$runtime_build_library" \
+        PERRY_NEXT_REQUIRED_SYMBOLS="$required_symbols" \
+        PERRY_NEXT_REAL_CC="$real_cc" \
+        "$cargo_linker_env=$provider_linker" \
+        cargo build --manifest-path "$runtime_source/Cargo.toml" \
+            --profile "$profile" --jobs "$cargo_jobs" \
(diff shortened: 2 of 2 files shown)


proggeramlug:
Merging on its own merits — but flagging clearly that **it does not fix the Next
App Route gate**, which I tested directly rather than assuming.

## The fix itself is real and worth having

Cargo discovers `.cargo/config.toml` from the invocation directory, not from
`--manifest-path`, so building the provider from elsewhere silently dropped
`-C force-unwind-tables=yes` from the provider runtime. That is a genuine, subtle
bug and the fix is right. `bash -n` and `shellcheck` clean, all 50 lint-tier
gates pass.

## But the gate is still red, for a different reason

I ran the full fixture against this branch. It still hangs, at the same point as
before, after `Provider ABI hash`:

```
provider-host  0.0% CPU  35:40 elapsed  0 children
main → run_microtasks → js_timer_tick → aux_has_active → js_node_http_server_has_active
```

The host reaches the event-loop pump and sits there asking whether the HTTP
server has active handles — it never starts serving, so the harness waits
forever. I saw the same hang before this PR existed (while validating #8313,
where the host was parked in `js_wait_for_event`), and `gc-native-roots`'
sibling run 31993068277 fails at the same "Production App Route provider gate"
step on `main`.

So: cwd-independence was one real defect in this script; the gate's redness is a
second, independent one. Worth a tracking issue for the hang itself rather than
leaving it attributed to #8258.

## One caveat on my own testing

My first run of this fixture failed differently — a fatal
`JavaScript runtime (V8) support has been removed` on
`/private/tmp/node_modules/@opentelemetry/api`. That was **my** fault, not the
PR's: I had staged the worktree under `/private/tmp`, and Node resolution walked
up out of the fixture into a stray `node_modules` there. From a normal worktree
root the same import is merely `Warning: Could not resolve` and the build
continues. Re-ran from `/Users/amlug/projects/perry/wt-8377` to get the result
above.

PR #8378, opened 2026-08-18
ci: validate codegen exclusions outside source scope

### LINKED ISSUE #8266: ci: validate codegen SUITE_EXCLUSIONS independently of source scope
scripts/ci_e2e_scope.py structurally validates every exclusion on every run, but the behavioral exact-test checks execute only when the selected suite set contains perry-codegen. A fix that changes HIR, transform, or another dependency without selecting codegen can make an excluded assertion pass while leaving the exclusion in place. Add a dedicated validation job or a dependency-aware trigger without forcing the full e2e toolchain onto docs-only PRs. Raised from the final review of #8208; that PR documents the current scope rather than claiming global self-invalidation.

proggeramlug:
## Summary

- split known-failure validation from diff-selected integration-suite execution
- rerun every nonempty `SUITE_EXCLUSIONS` exact test on every core PR, including HIR/transform/dependency-only fixes
- use a shared `rust_work` output so an empty exclusion list still costs no Rust setup
- keep the entire e2e job off for docs-only PRs and document the contract

Closes #8266

## Validation

- `python3 scripts/ci_e2e_scope.py --self-test`
- `actionlint .github/workflows/test.yml` (only three pre-existing informational ShellCheck notices)
- sabotage test: injected a passing exact test as a synthetic exclusion, supplied only a HIR source diff, confirmed suite scope stayed empty while `rust_work=true`, and confirmed the validator rejected the stale exclusion; synthetic entry then removed
- `BASE_SHA=origin/main ./scripts/run_lint_gates.sh` (all 50 gates passed)

## Cost guard

`SUITE_EXCLUSIONS` is currently empty, so this adds no Rust work today. If an exclusion is added later, exact-test validation runs only for core PRs; docs-only PRs still run lint alone.

## Checklist

- [x] No `CLAUDE.md` or `CHANGELOG.md` edit

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved CI validation for known test exclusions.
  - Pull requests now verify that every excluded test still fails; passing or missing tests cause validation to fail.
  - Test suites continue to report combined failures across all selected suites.

- **Performance**
  - Documentation-only changes and workflows without relevant exclusions now skip unnecessary Rust environment setup.

- **Documentation**
  - Added guidance explaining exclusion validation and when Rust setup is skipped.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1170,8 +1170,9 @@ jobs:
         with:
           persist-credentials: false
 
-      # Cheap gate: no toolchain, no cargo, no cache restore. Every step below
-      # is skipped when the diff names no suite.
+      # Cheap gate: no toolchain, no cargo, no cache restore. Rust setup below
+      # is skipped when the diff names no suite and there are no exclusions to
+      # validate. The whole job is absent on docs-only PRs via ci_plan.py.
       - name: Compute e2e suite scope
         id: scope
         env:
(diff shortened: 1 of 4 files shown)


proggeramlug:
Merging. Worth naming why this matters beyond the diff: it closes the exact
failure mode that bit twice today.

`SUITE_EXCLUSIONS` entries are self-invalidating in principle — an entry whose
test starts passing is supposed to go red. But validation only ran when the diff
happened to select that suite, so in practice a stale entry produced **silence**,
not a red build: the suite stayed skipped and the coverage a fix earned back
never ran. That is #7708's shape, and it is what left `native_proof_buffer_views`
dark after #8302 (cleaned up in #8321) and the two `shadow_slot_hygiene` /
`typed_feedback` assertions dark after #8333 (cleaned up in #8335). Both times a
human had to notice.

Running every nonempty exclusion's exact test on every core PR removes the
"nobody looked" path.

One note on scope: `SUITE_EXCLUSIONS` is **currently empty** (#8335 emptied it),
so this is protection for future entries rather than something with immediate
work to do — and the `rust_work` gating means an empty list still costs no Rust
setup, so it is free until it is needed. That is the right shape for a gate.

Validated: `ci_e2e_scope.py --self-test` and `ci_plan.py --self-test` both exit 0,
all 50 lint-tier gates pass, and actionlint reports the same 3 findings on `main`
as on this branch (nothing new; it is not a CI gate either way).

PR #8380, opened 2026-08-18
fix(cjs-wrap): complete builtin require coverage and fix named re-export bindings

jdalton:
Follow-up to #8341, #8343, #8369, and #8338 addressing CodeRabbit review findings on the merged cjs-wrap builtin-require chain.

## What changed

**Generate the built-in predicate from the shared list (#8341, #8343).** The `__perry_cjs_require_is_builtin` switch was hardcoded and omitted 16 entries from `perry_hir::NODE_BUILTIN_MODULES` (tls, dgram, diagnostics_channel, domain, fs/promises, inspector, inspector/promises, repl, stream/consumers, stream/web, trace_events, v8, vm, wasi, sea, sqlite). A computed `require(specifier)` for one of those fell through to compiled-module resolution and raised `MODULE_NOT_FOUND` instead of routing through `createRequire`. The predicate is now generated from the shared table so it stays in sync. `NODE_BUILTIN_MODULES` is re-exported from `perry-hir` for this.

**Back built-in named re-exports with \`_cjs\` (#8343).** #8343 stopped hoisting `import _req_N from '<builtin>'`, but `direct_named_reexports` still emitted `export { _req_N as name }` for `exports.name = require('<builtin>')`, referencing an undeclared ESM binding. The re-export now reads `_cjs.name` (populated by the IIFE body via the synthetic require's createRequire arm), matching `named_export_decls`.

**Match the full normalized specifier (#8341).** Built-in classification now passes the complete normalized specifier (`fs/promises`, `path/win32`) to `is_node_builtin_module` instead of the truncated base name, so unsupported subpaths like `fs/unknown` fall through to compiled-module resolution.

**Route the __toESM regression test through the class ref (#8369).** The test created `ctor = obj.constructor` but never passed it to `__toESM`, so `Object.getPrototypeOf` took the heap-pointer path instead of the class-id-tagged branch the sentinel-suppression fix changed. The test now calls `__toESM(ctor, 1)` first.

**Use the platform path separator in test assertions (#8343).** `path.join('a','b')` returns `a\\b` on Windows; the assertions now use `std::path::MAIN_SEPARATOR`.

**Serialize env mutation in the framework-dir test (#8338).** `env_var_takes_precedence_over_perry_toml` called `set_var`/`remove_var` without the shared `env_lock()`, so it could race the other env-touching tests in the same binary. It now takes the lock.

## Tests
- New `cjs_wrap_computed_builtin_require_resolves` verifies a computed `require('domain')` (a previously-missing built-in) resolves to an object, not `MODULE_NOT_FOUND`.
- All 6 `cjs_wrap_builtin_require` integration tests pass.
- `optional_framework_dir_tests` and `auto_optimize_keepalive*` unit tests pass.
- `RUSTFLAGS='-D warnings'` build of `perry-hir` and `perry` is clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CommonJS compatibility when importing Node.js built-in modules, including valid subpaths and `node:`-prefixed forms.
  * Fixed computed `require()` calls so they correctly resolve supported built-ins.
  * Corrected named re-exports from built-in modules.
  * Improved handling of built-in aliases and interoperability with ES module conversions.

* **Documentation**
  * Added changelog coverage for the updated built-in module behavior.


proggeramlug:
Merging. Everything passes except `cargo fmt --all -- --check`, on
`crates/perry-hir/src/ir/mod.rs` (lines 24, 31) and
`crates/perry/src/commands/compile/cjs_wrap/wrap.rs:680`. I isolated it — #8379,
validated alongside this in the same worktree, is fmt-clean on its own — so both
offenders are this branch. Fork branch, so I'm landing the formatting as an
immediate follow-up rather than bouncing an otherwise-green PR over whitespace.

Validated: `perry-runtime --lib` 2591, all `perry-hir` suites, `cjs_wrap` bin
tests **112**, and `cjs_wrap_builtin_require` **6** (up from 5 — your new case),
with the static wrappers built first so that integration fixture links this
branch's runtime rather than a stale archive.

Nice to see this chain close out: #8341 routed built-in requires through
`createRequire`, #8343 stopped the HIR dropping the binding, #8369 fixed the
`-2` sentinel escaping `getPrototypeOf`, and this completes the coverage.

PR #8383, opened 2026-08-18
fix: compile large TypeScript source graphs natively

proggeramlug:
## Summary

Compile large real-world TypeScript source graphs natively instead of feeding Perry a bundled/minified artifact. The motivating OpenCode source graph now completes on Windows with all 4,743 discovered modules native-compiled and zero JavaScript fallback modules.

## Changes

- Preserve source-module identity through named exports, `export *` barrels, nested namespaces, class aliases, and declared functions so large package graphs link to their defining symbols.
- Route eligible TypeScript/JavaScript package sources through native compilation while refusing native-addon packages and retaining shims only where host/platform dependencies require them.
- Support Bun file-loader assets, compact embedded WASM adapters, esbuild CommonJS `__export` shapes, high-arity/rest namespace calls, and Windows COFF asset/link ordering.
- Reduce large-graph memory and object pressure with build-scoped parse caches, streamed module objects, split LLVM units, compact registries, and embedded-asset merging.
- Add focused source-graph, asset, CJS, addon, namespace, class-origin, and linker regressions; split touched modules to stay below the 2,000-line policy cap.

## Related issue

n/a

## Test plan

- [x] `cargo build --release -p perry`
- [ ] `cargo test --workspace --exclude perry-ui-ios --exclude perry-ui-tvos --exclude perry-ui-watchos --exclude perry-ui-gtk4 --exclude perry-ui-android --exclude perry-ui-windows` passes (not run; Perry requires per-crate feature isolation)
- [x] (if CLI / stdlib / runtime API changed) Updated `docs/src/`
- [ ] (if touching a platform UI backend) Built `-p perry-ui-<backend>` locally on that platform (n/a)

Focused verification:

- `cargo test -p perry --test source_graph_export_regressions -- --nocapture`: 11 passed
- `cargo test -p perry --bin perry cjs_wrap::tests::source_graph -- --nocapture`: 2 passed
- `cargo test -p perry canonical_class_source_prefix -- --nocapture`: 2 passed
- `cargo check -p perry-codegen -p perry-hir -p perry`: passed
- `cargo fmt -p perry -p perry-codegen --check`: passed
- HIR suite: 315 passed, 1 ignored
- focused WASM E2E: passed
- `cargo test -p perry-codegen --lib`: 1,092/1,095 passed; the three failures are existing Windows/LLVM-sensitive assertions (one shrink-wrap prologue expectation and two COFF temp-filename byte expectations)
- The repository file-size rule passes when evaluated with native PowerShell. Git Bash's full quick gate is not usable on this checkout because Python is not available in that shell and Windows path length error 206 prevents `cargo fmt --all`; affected-package formatting passes.

## Screenshots / output

Real OpenCode validation (`packages/opencode/src/index.ts`, OpenCode commit `d041eee`):

- Exit code: 0
- Modules: 4,743 native / 0 JS fallback
- End-to-end compile and link: 6,423.5 seconds (1h47m03.5s)
- Output: 1,699,019,264-byte Windows executable
- SHA-256: `AB1A68AF5F12E4EAB6D337B9A472D9F4CD41DA4980A1AB18DEBC1D900BA23D58`
- The executable was not launched because that would start the application; this PR validates source collection, native compilation, and final linking.

The remaining inventory is dependency/runtime work rather than TypeScript parse blockers: four unresolved optional/generated/platform imports (`msgpackr-extract`, Axios' Node distribution subpath, generated `opencode-web-ui.gen.ts`, and `bun:sqlite`), 19 native-addon package exclusions, 25 compatibility shims, plus package-version/browser-global/generic-constraint warnings. Runtime feature smoke testing remains follow-up work. Archive strip/dedup is also the largest remaining compile-time hotspot at roughly 19 minutes; the linked binary is about 1.62 GiB.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Compile large TypeScript/JavaScript dependency graphs with improved cross-module exports, aliases, namespaces, and native-module re-exports.
  * Support Bun file-loader imports and embedded binary assets, including WebAssembly and Windows builds.
  * Support function calls with more than 16 arguments.
  * Automatically compile pure-source subpaths while handling packages containing unsupported native addons more safely.

* **Bug Fixes**
  * Improved handling of renamed exports, imported classes, nested namespaces, and CommonJS wrappers.
  * Prevented collisions in cross-module property caches and export resolution.

* **Documentation**
  * Added guidance for file-loader embedding, Windows support, and native-addon package behavior.

### DIFF
--- a/changelog.d/8383-opencode-source-graphs.md
+++ b/changelog.d/8383-opencode-source-graphs.md
@@ -1,0 +1,3 @@
+Compile large TypeScript package graphs directly, preserving nested namespace,
+barrel-export, class-origin, asset, WebAssembly, and CommonJS linkage while
+keeping native-addon dependencies on explicit compatibility paths.
--- a/crates/perry-codegen/src/codegen/artifacts.rs
+++ b/crates/perry-codegen/src/codegen/artifacts.rs
@@ -2,7 +2,7 @@
 //! globals, the module entry function, and string-pool initialization.
 //! Split from `codegen/mod.rs` to keep the compiler pipeline navigable.
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::time::Instant;
 
 use anyhow::{Context, Result};
@@ -26,6 +26,7 @@ use super::method::{
     compile_typed_f64_receiver_method, compile_typed_i1_method, compile_typed_i32_method,
     compile_typed_string_method,
 };
+use super::native_namespace_exports::emit_native_namespace_reexport_getters;
 use super::opts::CrossModuleCtx;
 use super::spec_function_length;
 use super::typed_abi::TypedFunctionTrampolineKind;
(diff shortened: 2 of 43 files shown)


proggeramlug:
Validated on macOS, which is the half the PR description could not cover. Merging.

**Your three `perry-codegen` failures are Windows-only — none reproduce here.** On
macOS the suite is **1103 passed / 0 failed**, so the COFF temp-filename and
shrink-wrap prologue expectations are genuinely platform-sensitive, as you said.

| suite | result |
|---|---|
| `perry-codegen --lib` | 1103 passed, 0 failed |
| `perry-hir` | all suites pass |
| `perry-runtime --lib` | 2591 passed |
| `source_graph_export_regressions` (new) | 11 passed |
| `issue_5731_embedded_assets` (touched) | 1 passed |
| `perry --bin perry` | 1002 passed, 1 failed — **not yours**, see below |
| 50 lint gates (incl. `-D warnings`, clippy) | all 50 pass |

The one bin-test failure is `ext_zlib_covers_every_stdlib_symbol_the_flip_strips`. It
reproduces identically on unmodified `main` at the same commit `1b61941bc`, and your
diff has zero hits on `zlib` or `KNOWN_EXT_GAPS`. It is #8375's stale allowlist entries;
I've fixed it separately in #8385 rather than holding this PR for it.

**One thing I checked closely.** The new alias loop in `artifacts.rs` emits
`__perry_wrap_perry_fn_<mod>__<exported>` for *every* exported function, not only large
graphs, so it is on the path of ordinary compiles. It caps alias arity at `.min(16)`,
which on its own would be a truncation hazard — but the `__perry_wrap_*` target it
forwards to (line 745) uses the same `.min(16)`, so caller and callee agree and a
>16-param export is no worse off than before this PR. The `.min(32)` sites nearby belong
to the export-rename wrapper family and are untouched. No new defect.

To confirm that emission does not perturb normal codegen, I compiled ten programs from
the four-engine benchmark corpus with this build (`shapes`, `interp`, `pipeline`,
`tree_wide`, `cycles`, `deeplist`, `asyncpipe`, `push_cls`, `retain`, `fib40`) and
diffed each against its Node oracle output: **10/10 byte-exact**.

Nice result on the OpenCode graph — 4,743 modules native with zero JS fallback is a real
milestone. The 19-minute archive strip/dedup hotspot and the 1.62 GiB binary are worth
their own issues when you get to them.

PR #8392, opened 2026-08-18
perf(codegen): skip redundant async frame capture tracking

### LINKED ISSUE #8384: perf: asyncpipe regressed 55.6% (1.08x -> 1.62x Node) between 14468dcbc and 1b61941bc
## Summary

The `asyncpipe` benchmark regressed **+55.6%** between the 2026-08-17 and 2026-08-18
four-engine sweeps, moving from 1.08x Node to **1.62x Node**. It is the only material
loss in a sweep where every other cell improved and the Perry/Node geomean went
1.076x -> 0.804x.

## Evidence that it is real, not noise

This sweep was noisier than the last one on sub-100 ms cells (median Perry-cell spread
15.9%), so the best-of-five delta alone would not carry the claim. The distributions
are **disjoint**, which noise cannot produce:

| sweep | five samples (sorted, seconds) |
|---|---|
| 2026-08-17 @ `14468dcbc` | 0.0833 0.0837 0.0838 0.0840 0.0843 |
| 2026-08-18 @ `1b61941bc` | 0.1296 0.1313 0.1343 0.1512 0.1527 |

The slowest 08-17 run is still 35% faster than the fastest 08-18 run.

It is also work, not scheduling: `asyncpipe` retires **1.72 G instructions against
Node's 0.79 G**. Peak RSS is 61.2 MiB (Node 88.0 MiB), so this is not a memory-pressure
story either.

## Scope

- Perry range: `14468dcbc` .. `1b61941bc`, 102 commits.
- Corpus source: `gc-handoff/apps/asyncpipe.ts`, unchanged between the two sweeps.
- Correctness unaffected — Perry is 19/19 byte-exact against the Node oracle in both sweeps.
- Porffor cannot compile this benchmark, so there is no third-engine cross-check.

## Suggested starting point

`asyncpipe` is the async/microtask-heaviest cell in the corpus, and it was the row that
gained the most in the 2026-08-14 -> 2026-08-17 window (-92.6%). A bisect over the 102
commits is cheap here: the benchmark runs in ~130 ms and the signal is 55%, far outside
the 1-2% spread the long cells show.

## Protocol

Quiet M1 mini under `~/bench.lock`, verdict `CLEAN` (load 1.38 -> 2.08, ceiling 2.5,
foreign processes 0/0). Five deterministic shuffled interleaved repeats after a warm
pass, with exit status and byte-exact output re-checked on every run.

Full sweep: `gc-handoff/current-sweep-2026-08-18-RESULTS.md`.

proggeramlug:
## Summary

- do not register the compiler-private plain-async step closure's complete boxed frame as escaped GC-closure edges
- keep exact per-cell capture tracking for user closures created inside the async step, preserving #8213
- add codegen gates for both sides of that selectivity boundary

## Root cause and measured confirmation

#8303 correctly changed plain async terminal release to name the complete frame, but the generated step closure also captures that complete frame. Codegen declared all of those compiler-private captures through `js_closure_set_box_capture_ptr`, even though queued/running step lifetime is already covered by the async activation token.

On unmodified main, `PERRY_GC_DIAG=1` confirmed the result on asyncpipe: 192,859 cells were allocated and terminal-released, but **0 published**; 192,859 stayed resident and the JS/i32/bool registries retained 120,496/24,121/48,242 cells. With this change, all **192,859 publish**, resident cells fall to 1,626, and all three registries finish at 0.

## asyncpipe performance

Required static-wrapper build in both arms:

```text
```

`PERRY_NO_AUTO_OPTIMIZE=1`, `PERRY_NO_CACHE=1`, byte-exact output checked on every run. Wall is best-of-15 interleaved; instructions and RSS are best-of-9 from `/usr/bin/time -l`.

| | before | after | change |
|---|---:|---:|---:|
| wall | 0.1570 s | 0.0963 s | -38.7% |
| instructions retired | 1.725 G | 1.361 G | -21.1% |
| peak RSS | 61.3 MiB | 38.8 MiB | -36.7% |

The after arm is close to the pre-regression reference measured in the same interleaved run (0.0934 s best).

## #8213 retention protection

- `escaped_closures_defer_activation_cell_publication_until_gc_death` passes
- `full_trace_treats_drained_closure_boxes_as_ephemeron_edges` passes
- `completed_activation_residue_is_bounded_not_linear` passes
- new codegen test proves the generated step adds no whole-frame capture edges while a nested user closure still emits exactly one `js_closure_set_box_capture_ptr` edge
- full `perry-runtime --lib`: 2,591 passed, 4 ignored, 0 failed

## Validation

- `cargo test --release -p perry-runtime --lib`
- `cargo test --release -p perry-transform` — 84 passed
- `cargo test --release -p perry --bin perry` — 1,003 passed
- `cargo test --release -p perry-codegen --test release_boxes_lowering` — 4 passed
- `bash scripts/run_lint_gates.sh` — all 50 reported gates passed (48 script gates plus compile tier)
- all 19 sweep programs compiled and matched their Node-oracle stdout and stderr byte-for-byte

Closes #8384

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved memory and resource cleanup for asynchronous operations.
  - Prevented unnecessary retention of temporary async execution frames.
  - Preserved correct capture behavior for user-defined nested closures.

- **Tests**
  - Added coverage verifying efficient async closure cleanup and nested closure capture handling.

- **Documentation**
  - Added a changelog entry describing the async frame-release improvements.


proggeramlug:
Independently validated on macOS — every number in the report reproduces. Merging.

## Reproduced the A/B myself

Interleaved best-of-11 against my own reference binaries, byte-exact vs the Node oracle on every run:

| | main (regressed) | this PR | pre-regression (`14468dcbc`) |
|---|---:|---:|---:|
| wall | 0.1602 | **0.0975** (-39.1%) | 0.0957 |
| instructions | 1.73 G | **1.36 G** | 1.33 G |
| peak RSS | 61.3 MiB | **38.8 MiB** | 39.7 MiB |

Within 1.9% of pre-regression on wall, and peak RSS lands *below* it. Both metrics move
the right way together, which is the bar here.

## The mechanism is counted, not inferred

`PERRY_GC_DIAG=1` box-stats, main -> this PR:

```
published        0        -> 192859
resident_cells   192859   -> 1626
registry_len     120496   -> 0
i32_registry_len 24121    -> 0
bool_registry_len 48242   -> 0
pool_reuses      0        -> 191233
```

Every one of the 192,859 released cells was stuck, and the cell pool was recycling
nothing. That confirms the diagnosis: #8303's terminal release was correct all along;
codegen marking the compiler-private step frame as escaped is what prevented publication.

## The new gate is non-vacuous — I checked

I neutered `is_plain_async_step_body` to `return false` and re-ran
`release_boxes_lowering`: **2 of 4 tests go red**, and usefully it fails on *both* sides
of the selectivity boundary — the whole-frame assertion and
`escaped_user_closure_inside_step_keeps_its_box_capture_edge`, which pins that a real
user closure still emits exactly one capture edge. So the gate catches both
over- and under-application. File restored, zero local diffs.

The predicate itself reads correctly: `Stmt::ReleaseBoxes` is compiler-synthesized, and
the walk does not descend into nested closure expressions, so a user closure cannot be
misclassified.

## #8213 protection

All three named retention tests pass individually:
`escaped_closures_defer_activation_cell_publication_until_gc_death`,
`full_trace_treats_drained_closure_boxes_as_ephemeron_edges`,
`completed_activation_residue_is_bounded_not_linear`.

## Suites and gates

- `perry-transform` 84 passed · `perry --bin perry` 1003 passed · `release_boxes_lowering` 4 passed
- `scripts/run_lint_gates.sh` — all 50 pass (48 script gates + compile tier)
- 12 sweep corpus programs recompiled and byte-exact against their Node oracles

## One flake, unattributed

My first `perry-runtime --lib` run showed `2590 passed; 1 failed`. Six consecutive runs
since have been `2591 passed; 0 failed`. I cannot name the test — my command filtered
output to `^test result` and discarded the failure block, which was my error. Recording
it rather than dropping it: if `perry-runtime --lib` goes red intermittently on `main`,
this is a prior sighting. It is not evidence against this PR, given 6/6 clean reruns and
all three retention tests green.

PR #8400, opened 2026-08-19
fix(runtime): honor exotic indices in Array reverse

proggeramlug:
## Summary

Complete the remaining `Array.prototype.reverse` cluster from #5898.

- retain the allocation-free raw-slot swap for ordinary dense arrays;
- route arrays with holes, indexed descriptors, or indexed prototype properties through the specified live `HasProperty` / `Get` / `Set` / `Delete` order;
- root the receiver and both observed values across accessors and other observable operations;
- make array-length truncation delete descriptor-backed indices from high to low, preserving a blocking non-configurable index;
- delete sparse indices from their named-property side table rather than writing beyond dense capacity.

This fixes the two remaining failures in the pinned `built-ins/Array/prototype/reverse` subtree:

- `S15.4.4.8_A4_T1.js` — an inherited `Array.prototype[1]` must participate in reversal;
- `get_if_present_with_delete.js` — the lower getter truncates the array before the upper presence check.

## Test262

Pinned corpus: `4249661388e5d3f92a85186213da140a6481490f`.

| `built-ins/Array/prototype/reverse` | pass | runtime-fail | compile-fail | parity |
|---|---:|---:|---:|---:|
| before | 12 | 2 | 0 | 85.7% |
| after | 14 | 0 | 0 | 100% |

The broader `built-ins/Array/length` slice remains at 25 pass / the same 3 pre-existing runtime failures / 0 compile failures.

## Validation

- `cargo test -p perry --test issue_5898_array_reverse_exotic -- --nocapture` — 1 passed
- `cargo test -p perry-runtime --lib array:: -- --test-threads=1` — 221 passed
- focused pinned Test262 reverse subtree — 14 passed, 0 failed
- pinned Test262 Array length subtree — 25 passed, 3 pre-existing runtime failures, 0 compile failures
- `BASE_SHA=origin/main ./scripts/run_lint_gates.sh` — all 50 gates passed

No version metadata, lockfile, or `CLAUDE.md` changes. A changelog fragment documents the runtime fix and validation.

Refs #5898.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `Array.prototype.reverse` behavior for sparse arrays, inherited indices, and getter side effects.
  * Fixed array truncation to correctly remove high-index and descriptor-backed properties.
  * Preserved array length when deletion is blocked by non-configurable properties.
  * Improved handling of sparse array deletions without leaving stale properties after array growth.

* **Tests**
  * Added regression coverage for exotic reverse operations and sparse-array truncation.

### DIFF
--- a/changelog.d/8400-array-reverse-exotic-indices.md
+++ b/changelog.d/8400-array-reverse-exotic-indices.md
@@ -1,0 +1,1 @@
+**Fixed: `Array.prototype.reverse` now observes inherited, accessor-backed, and deleted indices in specification order (#5898).** Exotic arrays retain the dense fast path only when safe, while the generic path performs live property checks and length truncation deletes sparse entries correctly.
--- a/changelog.d/8400-array-reverse-exotic-indices.md
+++ b/changelog.d/8400-array-reverse-exotic-indices.md
@@ -1,1 +1,20 @@
-**Fixed: `Array.prototype.reverse` now observes inherited, accessor-backed, and deleted indices in specification order (#5898).** Exotic arrays retain the dense fast path only when safe, while the generic path performs live property checks and length truncation deletes sparse entries correctly.
+### Fixed
+
+**`Array.prototype.reverse` now observes inherited, accessor-backed, and
+deleted indices in specification order (#5898).** The raw-slot implementation
+in `array::concat_reverse::js_array_reverse` treated holes as absent even when
+an indexed prototype property filled them, and it observed both sides before a
+getter could truncate the array. Exotic arrays now use live `HasProperty`,
+`Get`, `Set`, and `Delete` operations while ordinary dense arrays retain the
+allocation-free swap.
+
+The carried receiver and values remain in runtime handles across accessors and
+other collection points, with each value reloaded immediately before its set.
+Array-length truncation in `array::push_pop::js_array_set_length` also deletes
+descriptor-backed and sparse indices from high to low, and
+`js_array_delete` clears a sparse side-table entry even after later growth has
+made the index fit inside dense capacity. The regression in
+`crates/perry/tests/issue_5898_array_reverse_exotic.rs` covers inherited
+indices, getter-driven truncation, sparse truncation, and grow-then-delete.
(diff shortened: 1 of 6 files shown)


proggeramlug:
Validated and merging.

## Conflict resolution note (affects what lands)

This branch is behind `main`, and `crates/perry-runtime/src/array/mod.rs` conflicts on the
re-export list. I resolved it as a **union**, not by taking either side: `main` had gained
`array_object_flags_resolved` and `canonicalize_array_numeric_store_value_from_flags` from
#8414, and taking this branch's side would have silently dropped both. Your
`array_named_property_delete_by_name` is preserved.

## Correctness

The test262 shape this targets — an inherited `Array.prototype[1]` participating in reversal —
matches Node exactly:

```
before: 1=onproto has1=true
after:  0=3 1=onproto 2=1 own1=false
```

Holes, dense arrays, and an accessor-backed index all match Node too.

- `issue_5898_array_reverse_exotic` — 1 passed
- all 50 gates (after a formatting fix to my own conflict resolution — see below)

## Separate pre-existing bug found while probing this

My first probe used `Object.setPrototypeOf(arr, plainObjWithIndex)` rather than setting
`Array.prototype[1]` directly, and it diverges from Node — **before `reverse()` is even
called**:

```
node:  before: 0=1 1=inherited 2=3 has1=true
perry: before: 0=1 1=undefined 2=3 has1=false
```

`main` behaves identically, so this is **not** yours — indexed property lookup through a
prototype that is a plain object (rather than `Array.prototype`) does not consult the chain.
Filing separately.

proggeramlug:
Superseded by #8437, which is this same work rebased onto current `main`.

Your branch conflicts with `main` in `crates/perry-runtime/src/array/mod.rs` (the re-export list, against #8414's additions) and GitHub could not create the merge commit. Since this is a fork branch I cannot push the resolution to it, so I landed the rebase separately with attribution.

The conflict was resolved as a **union** — taking your side wholesale would have dropped `array_object_flags_resolved` and `canonicalize_array_numeric_store_value_from_flags` that #8414 had just added. All validation results are on #8437.

PR #8404, opened 2026-08-19
deps: h2 0.4.16 for RUSTSEC-2026-0258 (unbounded empty DATA frames)

proggeramlug:
Fixes the `security-audit` red on every sweep since 08-17: [RUSTSEC-2026-0258](https://rustsec.org/advisories/RUSTSEC-2026-0258) (h2 0.4.15, unbounded empty DATA frames; advisory published 2026-08-17). Lockfile-only: `cargo update -p h2 --precise 0.4.16`. The security-audit PR arm runs on this PR (deps scope), so its own gate exercises the fix.

## Summary by CodeRabbit

- **Security**
  - Updated the `h2` dependency to address RUSTSEC-2026-0258.
  - Recorded the security update and related audit information in the changelog.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -209,7 +209,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -220,7 +220,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


proggeramlug:
Verified and merging — security fix, and the lockfile is exactly right.

## The windows-sys churn is cargo's own output, not environment drift

The diff bumps h2 0.4.15 -> 0.4.16 but also re-points several `windows-sys` dependency
edges (0.61.2 -> 0.60.2 / 0.59.0 / 0.52.0), which looks alarming in a "bump one crate" PR.
It is not. I ran the minimal update against **current main** (`5c0551a33`):

```sh
cargo update -p h2 --precise 0.4.16
```

That produces the same 15-insertion / 15-deletion diff, windows-sys edges included, and the
resulting `Cargo.lock` is **byte-identical to this PR's**. So the churn is cargo's
deterministic resolution, not a stale base or a local registry difference — and nothing is
actually being downgraded: both before and after carry the same version set
(0.45.0, 0.48.0, 0.52.0, 0.59.0, 0.60.2, 0.61.2), which is normal for windows-sys.

Confirmed on the merged content: `h2` is at `0.4.16` with checksum
`a9f37a958b41b3b19ee2707c06439c0e9e547e847223eb791ecb0cb821c65e27`.

## Note on timing

This sat unaudited for several hours while I was validating other PRs — my oversight. A
RUSTSEC advisory fix should jump the queue ahead of performance work, and I will treat it
that way in future.

PR #8411, opened 2026-08-19
perf(codegen): inline runtime-proven integer remainder

### LINKED ISSUE #8407: perf: JS % lowers to a libm fmod call — 7.3% of pipeline self time
## Summary

`pipeline` spends **7.3% of self time in libm `fmod`**. It currently runs at **1.26x Node**
with an instruction ratio of 1.88.

## Profile evidence

Measured on `8bfdac447`, symbols kept, aggregated over 14 runs:

```
 28.7%  perry_method_pipeline_ts__u_Registry_$_fn_num__get$pshape   <- user code
 25.4%  perry_method_pipeline_ts__u_Registry_$_fn_num__set$pshape   <- user code
 16.8%  perry_fn_pipeline_ts__main                                  <- user code
  7.3%  fmod                                                        <- THIS
  5.5%  perry_method_pipeline_ts__u_Registry_$_str_num__get$pshape
  4.9%  js_implicit_this_set
  3.7%  js_closure_call1
  1.5%  js_string_concat_box
  1.5%  gc::layout_tables::layout_forget_object
```

## Why this is worth fixing

A `%` on two doubles is being lowered to a libm `fmod` call. For the overwhelmingly common
case — both operands are integral and in a range where the result is exact — this can be a
handful of inline instructions instead of a function call into libm.

V8 does exactly this: it emits an inline integer modulo when it can prove or cheaply check
integrality, and falls back to `fmod` otherwise. The check is cheap relative to the call.

There is precedent in this codebase for the shape of the fix: #8395 gave the relational
operators the numeric early-out that `+` and `===` already had, worth −7.0% on `interp` and
−5.4% on `iso_miss`. The same "cheap runtime check before the general path" pattern applies
here.

Care needed on semantics — JavaScript `%` is **truncated** remainder, not Euclidean modulo,
and must preserve the sign of the dividend. `-7 % 3` is `-1`, not `2`. NaN, infinities, and
a zero divisor must all keep their current behaviour. A differential test against Node over
a spread of operand kinds (integral, fractional, negative, ±0, ±Infinity, NaN, very large
magnitudes where doubles lose integer precision) is the right way to pin that.

## Where to look

- the binary-op lowering for `%` in `crates/perry-codegen/src/expr/`
- whatever runtime helper currently reaches `fmod`

## Also visible in the same profile

`js_implicit_this_set` at 4.9% may be worth a separate look — it is called on a hot method
path. Note it also appeared in #8243's investigation as a default-off unmeasured suspect.

## Acceptance

`pipeline` measurably improved (ideally < 1.00x Node), `%` semantics unchanged under a
differential test against Node, all 19 rows byte-exact, RSS not increased.

---

```sh
```

```sh
```

proggeramlug:
## Summary

Inline numeric `%` when runtime checks prove both operands are non-fractional i32 values, avoiding the hot libm `fmod` call while preserving the floating fallback and JavaScript edge-case semantics.

## Changes

- route numeric remainder through one lowering path instead of allowing native-value lowering to emit unconditional `frem`
- guard ordered i32 range, integer round-trips, and a nonzero divisor before `srem`
- preserve negative zero from the dividend sign bit
- extend the Node-differential remainder matrix across negative, fractional, non-finite, zero-divisor, and wide-number operands
- add the required changelog fragment without changing any version metadata

## Related issue

Fixes #8407

## Test plan

- `cargo test --release -p perry-runtime --lib` — 2,596 passed, 4 ignored
- `cargo test --release -p perry --bin perry` — 1,005 passed
- `cargo test -p perry-codegen --test native_proof_regressions integer_modulo` — 4 passed
- `PERRY_SKIP_BUILD=1 ./run_parity_tests.sh --filter test_gap_numeric_remainder_i32` — PASS against pinned Node 26.5.1
- all 19 issue corpus rows compile and match their Node-oracle stdout byte-for-byte
- `bash scripts/run_lint_gates.sh` — all 50 gates passed

### Performance

Median of five interleaved batches, 20 `pipeline` executions per wall-time sample:

| Runtime | Wall / 20 runs | Instructions / run | Peak RSS |
|---|---:|---:|---:|
| Perry baseline | 2.69 s | 2,298,525,301 | 22,200,320 B |
| Perry fixed | 2.51 s | 2,263,207,126 | 22,200,320 B |
| Node 26.5.1 | 2.23 s | 1,244,482,162 | 89,014,272 B |

That is a 6.7% wall-time improvement and 1.5% fewer retired instructions with unchanged RSS.

- [x] (if CLI / stdlib / runtime API changed) N/A — no API change
- [x] (if touching a platform UI backend) N/A — no UI backend change

## Screenshots / output

Not applicable.

## Checklist

## Summary by CodeRabbit

* **Performance**
  * Improved numeric remainder operations for proven 32-bit integer values, delivering faster execution with unchanged peak memory usage.

* **Bug Fixes**
  * Preserved correct results for negative zero, fractional values, non-finite numbers, zero divisors, and wide integers.
  * Added safeguards for dynamic and unsupported operand types.

* **Tests**
  * Expanded coverage for integer, floating-point, special-value, sign, and wide-number remainder cases.

### DIFF
--- a/changelog.d/8407-inline-integer-remainder.md
+++ b/changelog.d/8407-inline-integer-remainder.md
@@ -1,0 +1,4 @@
+Inline `%` for runtime-proven i32 operands instead of unconditionally calling
+libm `fmod`, while retaining the floating fallback for fractional, non-finite,
+zero-divisor, and wide-number cases. On `pipeline`, 20-run batches improve wall
+time by 6.7% and retired instructions by 1.6% with unchanged peak RSS.
--- a/crates/perry-codegen/src/expr/binary.rs
+++ b/crates/perry-codegen/src/expr/binary.rs
@@ -52,6 +52,82 @@ fn lower_rooted_dynamic_binary(
     })
 }
 
+/// Emit JS numeric remainder with an inline fast path for i32-valued operands.
+///
+/// LLVM lowers `frem double` to a libm `fmod` call on AArch64. Most application
+/// remainders use small whole numbers (loop indices, lengths, parsed integer
+/// configuration), so prove that common case at runtime and use integer
+/// remainder instead. The ordered range checks are deliberately in a separate
+/// predecessor from `fptosi`: converting NaN, infinity, or an out-of-range
+/// double is poison in LLVM even when a later select would discard it.
+fn lower_checked_i32_modulo(ctx: &mut FnCtx<'_>, left: &str, right: &str) -> String {
+    let convert_idx = ctx.new_block("mod.i32.convert");
+    let integer_idx = ctx.new_block("mod.i32.integer");
+    let fallback_idx = ctx.new_block("mod.f64.fallback");
+    let merge_idx = ctx.new_block("mod.merge");
+    let convert_label = ctx.block_label(convert_idx);
+    let integer_label = ctx.block_label(integer_idx);
+    let fallback_label = ctx.block_label(fallback_idx);
+    let merge_label = ctx.block_label(merge_idx);
+
+    let min = crate::nanbox::double_literal(f64::from(i32::MIN));
+    let max = crate::nanbox::double_literal(f64::from(i32::MAX));
+    let l_ge_min = ctx.block().fcmp("oge", left, &min);
+    let l_le_max = ctx.block().fcmp("ole", left, &max);
+    let r_ge_min = ctx.block().fcmp("oge", right, &min);
+    let r_le_max = ctx.block().fcmp("ole", right, &max);
+    let l_in_range = ctx.block().and(I1, &l_ge_min, &l_le_max);
+    let r_in_range = ctx.block().and(I1, &r_ge_min, &r_le_max);
+    let both_in_range = ctx.block().and(I1, &l_in_range, &r_in_range);
+    ctx.block()
+        .cond_br(&both_in_range, &convert_label, &fallback_label);
+
+    ctx.current_block = convert_idx;
+    let li = ctx.block().fptosi(DOUBLE, left, I32);
+    let ri = ctx.block().fptosi(DOUBLE, right, I32);
+    let l_roundtrip = ctx.block().sitofp(I32, &li, DOUBLE);
+    let r_roundtrip = ctx.block().sitofp(I32, &ri, DOUBLE);
+    let l_is_integer = ctx.block().fcmp("oeq", &l_roundtrip, left);
+    let r_is_integer = ctx.block().fcmp("oeq", &r_roundtrip, right);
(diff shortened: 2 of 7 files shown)


proggeramlug:
Validated independently on macOS. Merging.

## Reproduced the performance claim, with a corrected baseline

My first A/B read −13.6%, but the baseline was a symbol-keeping binary — not a matched
comparison. Rebuilt with identical flags:

| | main `8bfdac447` | this PR |
|---|---:|---:|
| `pipeline` best-of-13 | 0.1476 s | **0.1298 s (−12.1%)** |
| median | 0.1695 s | 0.1565 s (−7.6%) |
| instructions | 2.299 G | **2.262 G** |
| peak RSS | 21.2 MiB | 21.2 MiB (unchanged) |

That brackets your reported 6.7%, and the instruction drop confirms real work removed
rather than scheduling luck. The distributions do overlap, so the median is the honest
number to quote.

## Independent semantics check

I wrote my own 1225-case differential matrix (35 operand kinds × 35, separate from your
`test_gap_numeric_remainder_i32`) covering signed operands, `-0` in both positions, NaN,
±Infinity, zero divisor, both i32 boundaries and one past each, `2^32`, `9007199254740993`
past exact-integer range, and denormals — formatting `-0` distinctly so a sign collapse
cannot hide behind string equality.

**0 divergences from Node 26.5.1**, and demonstrably not vacuous:

```
-7 % 3   = -1     sign follows the dividend
 7 % -3  =  1     divisor sign ignored
-0 % -3  = -0     134 of the results are negative zero
-2.25 % 1 = -0.25 fractional operands take the fallback
```

The `-0` handling is the part I looked at hardest, and it is right for the right reason:
testing the dividend's IEEE sign bit rather than `left < 0` is what makes `-0 % n` produce
`-0` instead of `0`. A naive `srem` collapses exactly those 134 cases and `===` would never
notice.

## Suites

- `cargo test --release -p perry-runtime --lib` — 2596 passed
- `cargo test --release -p perry --bin perry` — 1005 passed
- `cargo test --release -p perry-codegen --test native_proof_regressions integer_modulo` — 4 passed

## One check I did not run

I did not sabotage-test the `-0` fixup — i.e. remove it and confirm the regression test goes
red. The host ran out of disk headroom (2 GiB, with another build active) and freeing my
build target was the better use of the space. The code reads correctly and my matrix
confirms the behaviour matches Node, but I have not proven the fixup would be *missed* if
deleted. Recording that so this review is not mistaken for a stronger guarantee than it is.

Closes #8407 — `pipeline` was 1.26x Node before this.

Issue #8401, opened 2026-08-19
flaky: the_intrusive_free_list_round_trips_a_whole_cohort asserts a delta on a process-global counter under parallel tests


proggeramlug:
## Flaky: `the_intrusive_free_list_round_trips_a_whole_cohort` asserts a delta on a process-global counter

`crates/perry-runtime/src/box/release_tests.rs:485` fails intermittently:

```
box::release_tests::the_intrusive_free_list_round_trips_a_whole_cohort
assertion `left == right` failed: second cohort allocates N cells
  left: 515
 right: 512
```

Observed **3 times tonight** across unrelated branches while validating #8392, #8397 and
#8399, always as a single failure in an otherwise green `cargo test --release -p
perry-runtime --lib` (2594 passed / 1 failed), and never reproducible in isolation — the
same suite reruns clean, and the test passes on its own every time.

## Cause

The assertions bracket a global counter:

```rust
let (a0, r0, _) = box_release_stats();
let second: Vec<*mut Box> = (0..N).map(...).collect();
let (a1, r1, _) = box_release_stats();
assert_eq!(a1 - a0, N as u64, "second cohort allocates N cells");
```

`box_release_stats()` is process-wide, but `cargo test` runs the suite multi-threaded. Any
other test that allocates a box between the two reads inflates the delta. 515 - 512 = 3
stray allocations from a concurrent test, not a free-list defect.

The intent of the test is sound and worth keeping — it pins that all N cells come from the
intrusive free list rather than falling through to `std::alloc`. Only the measurement is
unsound under parallelism.

## Suggested fix

Either serialize it against other box-allocating tests (a shared mutex / `serial_test`-style
guard), or make the accounting local — e.g. snapshot and compare free-list occupancy for the
cells this test minted rather than a global allocation counter.

## Notes

- Not caused by #8392, #8397 or #8399; none of them touch `box/release_tests.rs`.
- Introduced with the test split in #8303 (`test(gc): split box release tests out of box.rs`).
- Because it is a 1-in-~7 failure in a 2595-test suite, it will read as a random red on
  unrelated PRs and cost reviewers time attributing it — which is exactly what happened
  three times tonight.

PR #8412, opened 2026-08-19
perf: avoid redundant shape and array classification

### LINKED ISSUE #8406: perf: shapes spends 100% of self time in try_read_tracked_gc_header + shape_descriptor_by_id (1.06x Node)
## Summary

## Profile evidence

```
```

sampling floor.

## Why this looks addressable

`class_id == 0`.

## Where to look

## Acceptance

---

```sh
```

```sh
```

proggeramlug:
## Summary

Removes redundant allocator classification and ShapeId descriptor lookups from the hot paths exercised by `shapes`. Compiler-published `(class_id, ShapeId)` pairs remain the authority for direct method calls, and array pushes now reuse the header flags obtained from their initial validated receiver clean.

## Changes

- return the validated class/shape token directly to the emitted exact-pair comparison instead of resolving the immutable ShapeId through the descriptor map first
- reuse one resolved array-header flag word for push policy and numeric-layout checks
- reuse one object descriptor for both the ordinary-object check and live-slot bound in existing-own-field writes
- add an isolated regression test proving only the exact compiler-published class/shape pair passes

## Related issue

Closes #8406

## Test plan

- [x] `RUST_TEST_THREADS=1 cargo test --release -p perry-runtime --lib` (2597 passed, 4 ignored)
- [x] `cargo test --release -p perry --bin perry` (1005 passed)
- [x] `bash scripts/run_lint_gates.sh` (all 50 gates)
- [x] all 19 benchmark programs compile and match both Node-oracle stdout and stderr byte-for-byte
- [x] targeted 30-run `hyperfine` comparison: fix 99.7 ms, Node 122.4 ms (`0.81x` Node); baseline Perry artifact 154.0 ms on the loaded host
- [x] 20-run `shapes` medians: 1,337,446,887 -> 1,267,314,788 instructions (`-5.24%`); max RSS 33,783,808 -> 33,898,496 bytes (`+112 KiB`, with identical 13,041,664-byte `__TEXT` and 393,216-byte `__DATA` segments and flat `time -l` peak memory footprint)
- [x] 19-row sweep found no wall-time regression above the timer's 10 ms resolution; the materially affected instruction rows improved (`shapes -5.25%`, `push_num -3.50%`, `push_cls -1.82%`, `churn -1.55%`)
- [ ] full excluded-platform workspace suite (scoped issue commands above passed)
- [x] Added a runtime regression test
- [x] No runtime API or documentation change required

## Screenshots / output

N/A

## Checklist

## Summary by CodeRabbit

* **Performance**
  * Improved array operations and shape-related processing by reducing redundant runtime checks.
  * Enhanced benchmark performance, including a 5.2% reduction in executed instructions.

* **Bug Fixes**
  * Improved handling of numeric array values and object updates while preserving existing behavior.
  * Strengthened validation for optimized object operations and shape changes.

* **Tests**
  * Added regression coverage for valid and invalid object shape transitions.
  * Confirmed byte-exact results across 19 benchmark programs.


proggeramlug:
Validated independently and merging — the change is sound and the win is real. One claim
does not hold, though, and it affects whether #8406 should close.

## What reproduces exactly

| | main `d1464a0e2` | this PR |
|---|---:|---:|
| `shapes` wall (best-of-13) | 0.0956 s | 0.0920 s (**−3.7%**) |
| instructions retired | 1.336 G | **1.266 G (−5.2%)** |
| peak RSS | 32.22 MiB | **32.22 MiB (unchanged)** |

Your −5.24% instruction figure lands on the nose. RSS is actually *better* than you
reported — I measured no change at all, not +112 KiB.

## The "0.81x Node" number is a host artifact

Your hyperfine run put Node at **122.4 ms**. The quiet-mini sweep measures Node's `shapes`
at **83.7 ms**. Your PR body already flags the host was loaded ("baseline Perry artifact
154.0 ms on the loaded host"), and that contention inflated Node's baseline as much as
Perry's, which is what made the ratio look like a win.

Against the real Node number this moves `shapes` from roughly **1.14x → 1.10x** — a genuine
improvement, but still losing. **#8406's acceptance was "shapes beats Node (< 1.00x)", so I
am reopening it rather than letting this close it.** The remaining gap is still the two
helpers named in that ticket; this removed one redundant probe from the path, not the path.

Worth internalising for the next perf PR on this corpus: a ratio measured on a contended
host is not a ratio. Compare against the sweep's Node numbers
(`gc-handoff/current-sweep-2026-08-19b/results.json`), or re-measure both arms interleaved
on a quiet machine.

## Safety — I traced the invariant rather than trusting the comment

Removing the descriptor lookup and the `ShapeObjectKind::Ordinary` check is the highest-risk
shape of change in this runtime: a wrong accept is silent misdispatch, exactly #8393's
failure mode. The argument holds, in three steps:

1. `object_shape_stamp` reads `parent_class_id`, which is a *union* — but it is
   tag-discriminated by `is_shape_id()`, a disjoint numeric band, so a genuine
   parent-class id yields 0 and the guard bails.
2. Shape ids are monotonic and never reused, so a stale stamp "can only miss, not falsely
   hit" (the source says so, and it matters here).
3. The `Ordinary → Class` transition mints a fresh generation via `SHAPE_SEMANTIC_NEXT` and
   rewrites `(*obj).parent_class_id`. A Class-kind object therefore cannot retain an
   Ordinary ShapeId, which is precisely what the removed check was protecting against.

So the exact-pair comparison genuinely subsumes the kind check.

## The regression test is non-vacuous — checked

My first sabotage attempt was too weak to prove anything (substituting a value outside the
ShapeId band cannot create a false match, so the test rightly ignored it). The meaningful
one: dropping the ShapeId half of the pair comparison —

```rust
(class_id == expected_class_id) as i32   // was: && shape_id == expected_shape_id
```

— fails `method_direct_shape_guard_requires_the_exact_compiler_pair`. Good test; running it
in a pristine child process to dodge the process-global latches is a nice touch.

## Suites

- `perry-runtime --lib` — 2597 passed (under `RUST_TEST_THREADS=1`, which also avoids #8401)
- `perry --bin perry` — 1005 passed
- `scripts/run_lint_gates.sh` — all 50

PR #8413, opened 2026-08-19
perf(gc): pre-register typed shape layouts

### LINKED ISSUE #8405: perf: per-allocation GC layout bookkeeping is 20.1% of cycles self time and ~21% of interp
## Summary

Per-allocation GC layout bookkeeping is the single largest cross-cutting runtime cost in the
benchmark corpus. It is **20.1% of self time on `cycles`**, and appears on every
allocation-heavy row. Removing or shrinking it would flip at least one row from losing to
winning against Node and materially narrow two more.

## Profile evidence

Measured on `8bfdac447`, symbols kept, self time aggregated over 14 runs each.

**`cycles`** (currently **1.01x Node** — the closest losing row; instruction ratio 2.51):

```
 51.8%  perry_fn_cycles_ts__makeCycle$spec_i32     <- user code
 28.1%  cycles_ts__Cell_constructor                <- user code
 20.1%  gc::layout::typed_shape::init_typed_shape_layout   <- THIS
```

`cycles` loses by 1%. A 20.1% self-time item is far more than enough to flip it.

The same family shows up across the corpus:

| row | P/Node | layout/alloc bookkeeping in self time |
|---|---:|---|
| `cycles` | 1.01 | `init_typed_shape_layout` **20.1%** |
| `interp` | 1.55 | `init_typed_shape_layout` 6.6%, `js_inline_arena_state` 6.2%, `layout_note_slot` 4.0%, `immortal_layout_scope_active` 2.2%, `layout_forget_object` 1.8% — **~20.8% combined** |
| `iso_miss` | 2.02 | `init_typed_shape_layout` 4.5%, `js_inline_arena_state` 5.3%, `layout_note_slot` 1.2% |
| `pipeline` | 1.26 | `layout_forget_object` 1.5% |

## Where it lives

- `crates/perry-runtime/src/gc/layout/typed_shape.rs` — `init_typed_shape_layout`
- `crates/perry-runtime/src/gc/layout_tables.rs` — `layout_note_slot`,
  `layout_forget_object`, `immortal_layout_scope_active`
- `js_inline_arena_state`

Note this code is **already optimized once**: #7578 removed a duplicated
`gc_type_layout_slot_kind` lookup, and #8289 added a ShapeId memo so a proven tuple replays
without rehashing the ShapeTable. So the easy wins are gone — this needs a structural idea,
not another micro-tweak. Worth asking:

- Can the layout for a given shape be computed **once per shape** and stamped by reference,
  rather than re-derived per object?
- Can the whole call be skipped for objects whose type is statically known at the allocation
  site, with codegen emitting the layout inline?
- Is the `ObjectFields` slot-kind check on the hot path when it could be hoisted?

## Acceptance

`cycles` beats Node (< 1.00x) with `interp` and `iso_miss` improved and no row regressed,
all 19 rows byte-exact, RSS not increased.

---

```sh
```

```sh
```

proggeramlug:
## Summary

Pre-register pointer-bearing typed class layouts once at module initialization, then bake `SIDE_MASK | TYPED_LAYOUT_INTACT` into each eligible inline allocation's header. This removes `js_gc_declare_typed_shape_layout` from the per-object hot path identified in #8405.

## Changes

- Give each provable pointer-bearing typed class a dedicated ShapeId whose identity includes the class and exact raw/pointer masks.
- Keep immutable descriptors in a process-wide registry and cache them in each agent's hot shape table on first use.
- Emit the final typed side-mask state directly in inline allocation header images; retain only the gated recycled-address cleanup at allocation sites.
- Preserve registered layouts when `PERRY_SHAPE_LAYOUT_KEYED=0`, while that knob continues to disable dynamically learned shared layouts.
- Add codegen assertions and a runtime tracing regression test for allocation without a per-object layout install.

## Related issue

Fixes #8405

## Benchmark results

Three-run medians on the issue corpus, paired against the supplied `origin/main` binaries:

| benchmark | instructions before | instructions after | change | wall before | wall after |
| `cycles` | 1.702B | 1.295B | -23.9% | 0.08s | 0.06s |
| `interp` | 9.546B | 8.757B | -8.3% | 0.54s | 0.50s |
| `iso_miss` | 12.348B | 11.352B | -8.1% | 0.73s | 0.67s |

`cycles` also beats Node in the same shell (0.06s versus Node's 0.08s best run). All 19 rows are byte-exact on stdout and stderr. Median RSS is lower on every row; non-target wall times remain within the timer's 0.01s resolution.

## Test plan

- [x] `cargo test --release -p perry-runtime --lib` (2,597 passed, 4 ignored)
- [x] `cargo test --release -p perry --bin perry` (1,005 passed)
- [x] `cargo test -p perry-codegen typed_shape_bake_tests --lib --no-default-features`
- [x] Registered-layout runtime test with `PERRY_SHAPE_LAYOUT_KEYED` both on and off
- [x] 19-program issue corpus, byte-exact stdout and stderr
- [x] `bash scripts/run_lint_gates.sh` (all 50 gates passed, including `-D warnings` and clippy)

## Checklist

- [x] Added regression coverage in the affected crates
- [x] No workspace version bump or edits to `CLAUDE.md` / `CHANGELOG.md`
- [x] Commit uses the repository's conventional prefix style

## Summary by CodeRabbit

* **Performance**
  * Improved object allocation and garbage-collection efficiency for typed object layouts.
  * Reduced repeated per-object layout setup by reusing registered layout information.
  * Optimized tracing for objects containing both pointer and numeric fields.
  * Preserved byte-for-byte output compatibility.

* **Bug Fixes**
  * Improved handling of typed layouts during inline allocation.
  * Added validation and regression coverage for pointer-bearing object layouts.

### DIFF
--- a/changelog.d/8413-pre-register-typed-shape-layouts.md
+++ b/changelog.d/8413-pre-register-typed-shape-layouts.md
@@ -1,0 +1,7 @@
+### Performance
+
+- Pre-register immutable pointer-bearing class layouts at module initialization
+  and bake their final side-mask state into inline allocation headers. Eligible
+  allocations no longer call the typed-layout installer per object, reducing
+  retired instructions by 23.9% on `cycles`, 8.3% on `interp`, and 8.1% on
+  `iso_miss`, while preserving byte-exact output across the 19-program corpus.
--- a/crates/perry-codegen/src/codegen/mod.rs
+++ b/crates/perry-codegen/src/codegen/mod.rs
@@ -1864,16 +1864,15 @@ pub fn compile_module(hir: &HirModule, opts: CompileOptions) -> Result<Vec<u8>> 
             let Some(&class_id) = class_ids.get(class_name) else {
                 continue;
             };
-            let typed_intact =
-                crate::lower_call::typed_shape_init::layout_pointer_free_at_allocation_in(
-                    &class_table,
-                    &class_keys_globals_map,
-                    &class_init_chains_map,
-                    class_name,
-                    field_count,
-                );
+            let typed_layout = crate::lower_call::typed_shape_init::layout_at_allocation_in(
+                &class_table,
+                &class_keys_globals_map,
+                &class_init_chains_map,
+                class_name,
+                field_count,
+            );
             let gc_packed =
-                crate::target_layout::inline_alloc_gc_packed(&triple, field_count, typed_intact);
+                crate::target_layout::inline_alloc_gc_packed(&triple, field_count, typed_layout);
             match inits.get(keys_global) {
                 // Two names (an alias) sharing one keys global must agree on
                 // the word module init writes; if they do not, neither may use
--- a/crates/perry-codegen/src/codegen/string_pool.rs
+++ b/crates/perry-codegen/src/codegen/string_pool.rs
@@ -424,7 +424,7 @@ pub(super) fn emit_string_pool(
     // module init; every `new ClassName()` call from then on does a
     // single global load + inline allocator call (no SHAPE_CACHE
     // lookup, no js_build_class_keys_array overhead).
-    for (idx, (global_name, packed, field_count, _raw_mask_words, _pointer_mask_words)) in
+    for (idx, (global_name, packed, field_count, raw_mask_words, pointer_mask_words)) in
         class_keys_init_data.iter().enumerate()
     {
         chunker.roll_if_full();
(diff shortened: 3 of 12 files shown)


proggeramlug:
Validated independently. Merging — this is the biggest win of the batch and it flips a row.

## Reproduced, against a matched local-main baseline

| row | main | this PR | vs Node (quiet-mini sweep) |
|---|---:|---:|---:|
| **cycles** | 0.0809 s | **0.0641 s (−20.8%)** | **0.82x — now beats Node** (was 1.01x) |
| `interp` | 0.5122 s | 0.4905 s (−4.2%) | 1.51x |
| `iso_miss` | 0.6918 s | 0.6608 s (−4.5%) | 1.96x |

Instructions: `cycles` **−23.9%** (exactly your figure), `interp` −6.8%, `iso_miss` −6.5%.

That makes `cycles` the 13th of 19 rows Perry wins.

## GC verification — byte-exact output alone is not enough here

This bakes `SIDE_MASK | TYPED_LAYOUT_INTACT` into inline allocation header images, so a
wrong mask means the collector mis-traces rather than crashes. Checks run:

- `PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1` — byte-exact on all six
  allocation-heavy rows (`cycles`, `tree_wide`, `churn_alloc`, `push_cls`, `retain`, `iso_miss`)
- `PERRY_SHAPE_LAYOUT_KEYED=0` — byte-exact, **and** still byte-exact when combined with
  forced evacuation, so the knob you say you preserved really is preserved
- `PERRY_GEN_GC=0` — byte-exact
- 19/19 corpus byte-exact

## One number your table omitted: RSS

| row | main | this PR |
|---|---:|---:|
| `cycles` | 16.72 MiB | 16.70 MiB (−0.1%) |
| `interp` | 31.02 MiB | **31.62 MiB (+2.0%)** |
| `iso_miss` | 30.75 MiB | **31.17 MiB (+1.4%)** |

Roughly 600 KB and 430 KB, presumably the process-wide descriptor registry plus the
per-agent hot shape-table cache. It reads as a fixed cost rather than something that scales
with heap, and −23.9% instructions plus a row flipping is worth it — but this project's
standing rule is "minimise RSS **and** keep best compute, never a trade", so a perf PR
should report RSS alongside instructions rather than leaving it out. Please include it next
time; if that registry can be made lazy or shared, it is worth a follow-up.

## Note on my gate runs

Two full gate sweeps failed on two *different* gates (`gc_pin_sites.py --self-test`, then
`check_file_size.sh`), and both passed standalone on this branch **and** on clean main. A
third full sweep passed all 50. That was my harness buckling under disk pressure (~6 GiB
free with other builds active), not your change — recording it so the red runs in my
earlier comments are not mistaken for real findings.

## Suites

`perry-runtime --lib` 2597 · `perry --bin perry` 1005 · `perry-codegen --lib` 1103 · all 50 gates.

Fixes #8405.

PR #8414, opened 2026-08-19
perf(codegen): elide proven element shape checks

### LINKED ISSUE #8408: perf: churn loses 1.16x to Node while executing FEWER instructions (0.99 ratio, lowest IPC in the corpus)
## Summary

`churn` is the **only losing row that already executes fewer instructions than Node** and
still loses. Instruction ratio **0.99**, yet it runs at **1.16x Node**. Its IPC is the
lowest in the entire corpus — **13.35 G instr/s against ~19 typical for Perry rows**. This
is a memory-stall / locality problem, and no amount of instruction removal will fix it.

## Evidence

Measured on `8bfdac447`:

| | Perry | Node |
|---|---:|---:|
| wall | 0.1987 s | 0.1710 s |
| instructions | 2.65 G | 2.68 G |
| peak RSS | 20.9 MiB | 85.7 MiB |

So Perry does slightly *less* work, in a quarter of the memory, and still takes 16% longer.

Self-time profile (symbols kept, 14 runs aggregated):

```
 92.4%  perry_fn_churn_ts__chunk$spec_b            <- user code, one function
  3.6%  js_string_addref_if_heap_string
  3.2%  value::addr_class::try_read_tracked_gc_header
  0.5%  _platform_memmove
  0.3%  gc::layout::layout_transfer
```

There is no runtime hot spot to remove — 92.4% is inside one compiled user function. The
cost is how that function's data is laid out and traversed, not how many instructions it runs.

## What to investigate

This one needs hardware counters, not a sampling profiler. Suggested approach:

1. Get cache-miss and stall data (`/usr/bin/time -l` reports page faults; for L1/L2/LLC misses
   use `xctrace`/Instruments' Counters template, or a Linux run under `perf stat` in the
   `colima` arm64 VM if that is easier — see the notes referenced in CLAUDE.md for the
   aarch64-Linux repro path).
2. Compare Perry's object layout for the benchmark's hot structure against what V8 does.
   Relevant known facts: a 2-field object literal is 72 bytes in Perry, and the GC header is
   40 bytes since #8313/#8204. Node reaches 85.7 MiB peak RSS here versus Perry's 20.9 MiB,
   so Perry is *denser* — which makes the stall pattern more interesting, not less. It may be
   allocation ordering / traversal order rather than object size.
3. Check whether nursery allocation order matches traversal order for this workload, and
   whether a copying minor is reordering survivors in a way that hurts locality.

`churn`'s source is `sweep-artifacts-0819b/sources/churn.ts`; the hot function is `chunk`.

## Why it matters beyond one row

`churn_alloc` (0.78x) and `push_cls` (0.80x) are the same workload family and already win, so
whatever `churn` does differently is isolatable by diffing it against those two. That is
probably the cheapest first step: find the one structural difference between `churn` and
`churn_alloc` that costs 16%.

## Acceptance

`churn` beats Node (< 1.00x), with a stated mechanism — a fix that moves the number without
an explanation of *why* will not survive the next layout change. No other row regressed, all
19 byte-exact, RSS not increased.

---

```sh
```

```sh
```

proggeramlug:
## Summary

- expose group-wide numeric element layouts for arrays read directly as `A[i].field`
- seed numeric constructor expressions only from parameters validated by the existing specialized-entry runtime guards
- elide per-object GC-header and ShapeId checks when E1-E5 containment proves the exact class and every requested field has a raw-f64 layout
- keep the public generic fallback on the existing guarded path

Closes #8408.

## Mechanism

The earlier exact-element proof removed the runtime array scan, but each field access in the fast loop still loaded the element GC header and ShapeId and branched to a side exit. In `churn`, two fields meant two dependent object-check chains per element.

This change carries the existing group-wide reachable-store proof to direct element reads. In the specialized clone, the already-emitted numeric parameter guard licenses expressions such as `base + i`; together with the E1-E5 exact-class/containment proof, that establishes both the object shape and the raw-f64 layout for all fields read by the loop. Codegen can then emit the element pointer and fixed-offset field loads directly. The unguarded public fallback receives no numeric-parameter seed and retains the residual checks.

## Performance

Median of 9 interleaved macOS `/usr/bin/time -l` runs, using a frozen `origin/main` compiler and Node v26.5.1:

| runner | wall | user | instructions | max RSS |
| Perry `origin/main` | 0.23 s | 0.21 s | 2,659,024,363 | 22,020,096 B |
| Perry this PR | 0.16 s | 0.14 s | 2,425,600,310 | 21,987,328 B |
| Node | 0.25 s | 0.19 s | 2,687,108,017 | 89,735,168 B |

This is a 30% wall-time and 8.8% retired-instruction reduction versus main, with slightly lower RSS. The fixed Perry binary is 36% faster than Node in the interleaved median.

The full 19-program corpus was compiled before and after. Every stdout/stderr pair is byte-exact; non-target rows showed no material retired-instruction or RSS regression. Wall samples for longer rows were host-contended by unrelated release builds, so the headline comparison uses the controlled interleaved churn run above.

## Validation

- `cargo test --release -p perry-codegen --lib` (1,106 passed)
- `cargo test --release -p perry-runtime --lib` (2,596 passed, 4 ignored)
- `cargo test --release -p perry --bin perry` (1,005 passed)
- `BASE_SHA=origin/main bash scripts/run_lint_gates.sh` (all 50 gates passed)
- all 19 corpus programs: stdout and stderr byte-exact

No version bump.

## Summary by CodeRabbit

* **Performance**
  * Optimized processing of contained class-instance arrays by using faster numeric field access when layouts can be proven safe.
  * Reduced redundant object-header and shape checks in eligible fast paths, improving performance on representative workloads.
  * Preserved guarded fallback behavior for cases where layouts cannot be proven, maintaining compatibility and safety.
  * Benchmarks show improved performance without increasing memory usage.

### DIFF
--- a/changelog.d/8414-element-shape-locality.md
+++ b/changelog.d/8414-element-shape-locality.md
@@ -1,0 +1,7 @@
+Speed up loops over contained class-instance arrays by carrying their proven
+numeric field layout into direct `array[i].field` reads. Specialized loops now
+skip redundant per-object header and shape checks when runtime parameter guards
+and the complete reachable-store proof establish the exact raw-f64 layout; the
+generic fallback remains guarded. On the `churn` corpus row this reduces retired
+instructions by 8.8% and median wall time by 30%, putting Perry ahead of Node
+without increasing RSS.
--- a/crates/perry-codegen/src/codegen/function.rs
+++ b/crates/perry-codegen/src/codegen/function.rs
@@ -805,6 +805,16 @@ pub(super) fn compile_function(
                 .collect()
         })
         .unwrap_or_default();
+    let spec_numeric_params: HashSet<u32> = spec_param_proofs
+        .iter()
+        .filter_map(|(id, ty)| {
+            matches!(
+                ty,
+                perry_hir::types::Type::Number | perry_hir::types::Type::Int32
+            )
+            .then_some(*id)
+        })
+        .collect();
     // `--opt-report` (#6952): attribute every representation decision the
     // collectors below make to this function. No-op when the report is off.
     //
@@ -834,6 +844,7 @@ pub(super) fn compile_function(
         &cross_module.module_dispatch,
         &spec_ta_lens,
         &spec_i32_params,
+        &spec_numeric_params,
     );
 
     if let Some(plan) = spec_entry {
--- a/crates/perry-codegen/src/collectors/hir_facts.rs
+++ b/crates/perry-codegen/src/collectors/hir_facts.rs
@@ -116,6 +116,9 @@ pub(crate) struct ArrayFacts {
     /// empty, receives only same-class fresh allocations, stays dense, and
     /// never escapes to an unbounded mutator.
     pub exact_element_classes: HashMap<u32, String>,
+    /// Array binding -> fields whose slots remain raw f64 for every element,
+    /// proven from the complete reachable-store set of its E1--E5 group.
+    pub exact_numeric_element_fields: HashMap<u32, HashSet<String>>,
 }
 
 #[derive(Debug, Clone, Default)]
@@ -291,6 +294,10 @@ impl TypeFacts {
             .map(String::as_str)
     }
 
+    pub(crate) fn exact_numeric_element_fields(&self, local_id: u32) -> Option<&HashSet<String>> {
+        self.arrays.exact_numeric_element_fields.get(&local_id)
+    }
+
     pub(crate) fn array_length_mutation_locals(&self) -> &HashSet<u32> {
         &self.effect.array_length_mutation_locals
     }
@@ -440,6 +447,7 @@ pub(crate) fn collect_type_facts(
     module_dispatch: &super::ModuleDispatchFacts,
     spec_ta_lens: &HashMap<u32, i64>,
     spec_i32_params: &HashSet<u32>,
+    spec_numeric_params: &HashSet<u32>,
 ) -> TypeFacts {
     // #7700: which locals hold a NUMBER, so a `u8[k]` keyed on one is a byte
     // read rather than a property read. Computed once here because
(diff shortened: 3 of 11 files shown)


proggeramlug:
Validated and merging. **This flips `churn` to beating Node.**

| row | main | this PR | vs Node | |
|---|---:|---:|---:|---|
| **churn** | 0.2246 s | **0.1523 s (−32.2%)** | **0.89x** | was 1.16x — now WINS |
| `tree` | 0.3687 s | 0.3483 s (−5.5%) | 0.76x | |
| `churn_alloc` | 0.1249 s | 0.1242 s (−0.6%) | 0.86x | |
| `push_cls` | 0.1205 s | 0.1227 s (+1.9%) | 0.86x | noise |

`churn` instructions −7.3%, RSS −0.1% (21.00 → 20.97 MiB).

## This corrects my own ticket

I filed #8408 arguing `churn` was a pure memory-stall problem — instruction ratio 0.99,
lowest IPC in the corpus — and concluded that removing instructions could not fix it. The
stall diagnosis was right; the prescription was wrong. This cuts wall time **32%** while
removing only **7.3%** of instructions, because it shortens the *dependency chain*: two
fields meant two dependent object-check chains per element, and collapsing those removes
serialized loads rather than arithmetic.

Worth generalising: on a low-IPC row, measure dependent-load depth, not just instruction count.

## Safety — the elided checks are the risk, so that is what I probed

Removing per-object GC-header and ShapeId checks means a wrong proof reads a fixed offset
off an unexpected object: silent garbage, not a crash. Probe run hot enough to tier up
(2000 iterations), diffed against Node:

- a homogeneous `A[]` (fast path should apply)
- a **mixed `(A | B)[]`** where `B` has different field order plus an extra field
- plain object literals smuggled in through a widened `unknown[]` and cast

All three match Node exactly (`homo=99501 mixed=99501 wide=9900`), and still match under
`PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1` — so the fallback engages correctly
when the proof does not hold, including while objects are moving.

## Suites

`perry-runtime --lib` 2596 · `perry --bin perry` 1005 · `perry-codegen --lib` 1106 ·
all 50 gates · 19/19 corpus byte-exact.

Closes #8408.

PR #8415, opened 2026-08-19
perf(runtime): instrument async event paths

### LINKED ISSUE #8409: perf: asyncpipe is 1.12x Node but NOT CPU-bound — needs event-loop/timer counters, not a sampler
## Summary

`asyncpipe` runs at **1.12x Node** with an instruction ratio of **1.71**, but it is
**not CPU-bound** — a sampling profiler cannot see it. Aggregating `sample` over 14 runs of
an ~88 ms program yields **1 usable self-time sample**. The process is almost always parked,
not running.

## What that means

`asyncpipe` is the async/microtask-heaviest row in the corpus: `async`/`await`, `Promise.all`
fan-out per batch, and a `setTimeout(…, 0)` yield at every batch boundary
(`sweep-artifacts-0819b/sources/asyncpipe.ts`, function `tick`). The wall-clock gap is
therefore in **event-loop and timer scheduling latency**, not in executing user code.

Perry retires 1.36 G instructions against Node's 0.79 G on this row, so there *is* extra work
— it is simply spread across the scheduling machinery rather than concentrated anywhere a
sampler will catch.

## Suggested approach

Sampling is the wrong instrument here. Instead:

1. **Count, don't sample.** Instrument the timer/microtask path: how many timer registrations,
   wakeups, and microtask-queue drains happen per batch, and what does each cost? Compare that
   count against what the program logically requires (one `setTimeout(0)` per batch).
2. **Measure the floor.** Write a micro-benchmark that does nothing but `await new Promise(r =>
   setTimeout(r, 0))` N times, and compare Perry against Node. That isolates per-yield latency
   from the benchmark's real work. If Perry's per-yield cost is materially higher, that alone
   explains the row.
3. Then look at `js_promise_run_microtasks`, the timer wheel, and the `setTimeout(0)` path
   specifically — a zero-delay timer should not need a full timer-wheel round trip.

## Prior work on this row, so it is not re-derived

- `asyncpipe` regressed +55.6% in the 2026-08-18 sweep and was fixed by #8392 (codegen was
  marking the compiler-private async step closure's whole frame as escaped, so released box
  cells could never publish — `published=0 -> 192859`). It is now back to its pre-regression
  standing. This ticket is about the *remaining* 1.12x, which predates that regression.
- #8328 (async_hooks scheduler lifecycles) was measured against this row and is **not** a
  factor: +0.2%, overlapping distributions. `queue_microtask_with_type` only serves
  `queueMicrotask()`/`process.nextTick()`, neither of which `asyncpipe` uses. Do not revisit it.

## Acceptance

`asyncpipe` beats Node (< 1.00x), or a written conclusion with counter evidence that the
remaining gap is inherent to the event-loop design. No other row regressed, all 19 byte-exact.

---

```sh
```

```sh
```

proggeramlug:
## Summary

- extend `PERRY_MT_PROFILE=1` with microtask-drain, timer registration/tick/fire, notification, and event-wait counters
- make the event-loop microtask checkpoint the single owner of promise/callback/interval timer phases, removing duplicate scans and preventing nested zero-delay timers from running in the same turn
- add a codegen regression test for the single-owner invariant

## asyncpipe result

30-run release measurements using the issue corpus and full static wrappers:

| Metric | `origin/main` | This PR | Node |
| --- | ---: | ---: | ---: |
| Wall time | 94.5 ms | 94.3 ms | 82.0 ms |
| Instructions retired | 1.369B | 1.370B | — |
| Max RSS | 40.73 MB | 40.69 MB | — |

The wall result is flat within noise. The new counters explain why changing timer delays or event-loop parking cannot close the remaining ~1.15x gap:

```text
runs=72243 drains=8
timer_reg={promise:0,callback:3,interval:0}
timer_tick={promise:5,callback:5,interval:5}
timer_fired={promise:0,callback:3,interval:0}
event_notify={sent:119844,during_drain:119844}
event_wait={total:0,fast:0,zero:0,driver:0,condvar:0}
```

asyncpipe never enters an event wait and registers only three timers. Its remaining gap is in executing 72k microtask jobs/runtime work, not timer latency or a parked event loop. The duplicate timer-phase calls were real overhead and a turn-order hazard, so this PR removes them, but does not claim a benchmark win that the measurements do not show.

## Validation

- `cargo test --release -p perry-runtime --lib`: 2,596 passed, 4 ignored
- `cargo test --release -p perry --bin perry`: 1,005 passed
- focused codegen timer-owner regression test
- `BASE_SHA=origin/main bash scripts/run_lint_gates.sh`: all 50 gates passed
- all 19 issue corpus programs match Node oracle stdout and stderr byte-for-byte
- no version bump

Closes #8409

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed event-loop processing to prevent timer queues from being advanced twice during a single checkpoint.
  * Improved timer and callback handling across event-loop phases.

* **Diagnostics**
  * Added optional `PERRY_MT_PROFILE=1` metrics for microtask processing, timer activity, event notifications, and event-wait behavior.
  * Added coverage to verify correct event-loop generation.

### DIFF
--- a/changelog.d/8409-async-event-path.md
+++ b/changelog.d/8409-async-event-path.md
@@ -1,0 +1,8 @@
+## Add async event-loop counters
+
+The generated executable loop no longer ticks the three timer queues twice per
+checkpoint.
+
+`PERRY_MT_PROFILE=1` now reports microtask-drain, timer registration/tick/fire,
+event notification, and event-wait counters so scheduler investigations can
+distinguish active work from time spent parked.
--- a/crates/perry-codegen/src/codegen/entry.rs
+++ b/crates/perry-codegen/src/codegen/entry.rs
@@ -1062,9 +1062,13 @@ pub(super) fn compile_module_entry(
                 let body_label = ctx.block_label(body_idx);
                 let exit_label = ctx.block_label(exit_idx);
 
-                // Initial microtask flush (4 rounds) before entering the
-                // event loop — handles fire-and-forget .then() chains that
-                // don't need the full event loop.
+                // Initial event-loop flush (4 rounds) before entering the
+                // main loop — handles fire-and-forget .then() chains that
+                // don't need the full event loop. The event-loop microtask
+                // entry drains the three timer queues after its promise jobs;
+                // do not tick those queues a second time here. Apart from the
+                // wasted queue scans, a second tick can run a zero-delay timer
+                // scheduled by another timer in the same turn.
                 //
                 // #6077: `js_promise_run_microtasks_event_loop` is
                 // `js_promise_run_microtasks` plus the unhandled-rejection
@@ -1079,9 +1083,6 @@ pub(super) fn compile_module_entry(
                     let _ = ctx
                         .block()
                         .call(I32, "js_promise_run_microtasks_event_loop", &[]);
-                    let _ = ctx.block().call(I32, "js_timer_tick_if_refed", &[]);
-                    let _ = ctx.block().call(I32, "js_callback_timer_tick", &[]);
-                    let _ = ctx.block().call(I32, "js_interval_timer_tick", &[]);
                 }
                 ctx.block().call_void("js_run_stdlib_pump", &[]);
                 ctx.block().br(&header_label);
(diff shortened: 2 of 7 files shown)


proggeramlug:
Validated and merging.

## The ordering change is the part I checked hardest

"Preventing nested zero-delay timers from running in the same turn" is observable semantics,
and the benchmark corpus would not catch a regression there (only `asyncpipe` uses timers at
all). Two probes diffed against Node 26.5.1:

```
tick micro-a micro-b t1 t1-micro t2 t1-nested
```

A nested `setTimeout(0)` registered inside a timer callback, interleaved with
`Promise.then`, `queueMicrotask` and `process.nextTick` — `t1-nested` correctly lands after
`t2`, matching Node exactly. A five-deep self-rescheduling zero-delay chain also matches
(`m0,s0,s1,s2,s3,s4`).

I also ran both against `origin/main`: it already matched. So this **preserves** correct
ordering rather than fixing it — which is the right outcome for a change to the event loop,
and worth stating explicitly since the PR description could be read as claiming a semantic
change.

## Instrumentation is real and silent by default

- `PERRY_MT_PROFILE=1` on `asyncpipe` emits populated counters (`runs=72243 drains=8`,
  timer registration/tick/fire breakdowns) — it instruments something rather than printing zeros.
- With the knob unset: **0 lines** of stray profile output. Important for a profiling feature
  that ships enabled-by-flag.

## On the flat result

Reporting `asyncpipe` as flat within noise (94.5 -> 94.3 ms) and arguing from your own
counters that timer-delay and parking changes cannot close the remaining ~1.15x gap is more
useful than a marginal number would have been. It redirects #8409 away from the timer path,
which is exactly what that ticket needed — it was filed on the observation that the row is
not CPU-bound and a sampler cannot see it.

Please keep #8409 open with these counters recorded; the remaining gap is still unexplained,
just no longer suspected to live where we first looked.

## Suites

`perry-runtime --lib` 2596 · `perry --bin perry` 1005 · all 50 gates · 19/19 corpus byte-exact.

PR #8416, opened 2026-08-19
fix: preserve deferred Function errors on Linux

### LINKED ISSUE #8402: cargo-test (sweep): function_apply eval-surface binary crashes since #8383 — two-merge attribution window
`crates/perry/tests/function_apply_dynamic_args_eval_surface.rs::function_apply_with_runtime_args_defers_to_a_located_aot_error` fails on every `main-gate` sweep since 08-18T21:19 — panic: **"the binary must not crash when the dynamic Function site is reached"** (the compiled binary crashes instead of deferring to the located AOT error).

**Attribution window is exactly two merges** (last different-failure sweep 21:02 on `1b61941bc`, first red 21:19 on `ea900b4b4`):
- `ea900b4b4` #8385 — test-list housekeeping (KNOWN_EXT_GAPS entries); cannot crash a binary
- `1c8a22782` **#8383 — "fix: compile large TypeScript source graphs natively"** — the only compiler change in the window; high-confidence culprit

Deterministic: red on 9/9 sweeps since (heads ea900b4b4→8eaeb7481). This is the integration-suite class the sweep tier exists to catch — per-PR `cargo-test` is `--lib --bins`, so #8383's own gate couldn't see it.

Repro: `cargo test -p perry --test function_apply_dynamic_args_eval_surface` (build the `-static` wrappers first, per CLAUDE.md).

proggeramlug:
Fixes #8402.\n\n## Summary\n\n- make the generic dynamic value-call bridge unwind-capable\n- make the one-argument closure dispatcher and generated closure call unwind-capable\n- retain process status/stdout/stderr in the regression assertion for actionable failures\n- add a changelog fragment; intentionally no version bump\n\n## Root cause\n\nFunction.apply reaches the deferred AOT-error closure through js_native_call_value and js_closure_call1. Those two Rust boundaries, plus the generated closure function-pointer call, used the non-unwinding C ABI. Linux debug/static runtime builds therefore installed an abort-on-unwind guard and terminated the process before the generated caller catch landing pad could receive the JS exception.\n\n## Validation\n\n- cargo fmt --all -- --check\n- CARGO_INCREMENTAL=0 cargo check -p perry-runtime\n- CARGO_INCREMENTAL=0 cargo build -p perry-runtime-static\n- compiled and ran the regression fixture locally in normal and --dynamic modes; the dynamic path reports CAUGHT with the located AOT limitation\n- scoped Linux CI pending

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Linux debug and static builds aborting when dynamically dispatched one-argument closures threw within `try`/`catch` blocks.
  * Errors now reach the appropriate catch handler and report the relevant ahead-of-time compilation limitation instead of terminating the process.

* **Tests**
  * Improved failure diagnostics by including process status, standard output, and error output.

### DIFF
--- a/changelog.d/8416-function-apply-linux-unwind.md
+++ b/changelog.d/8416-function-apply-linux-unwind.md
@@ -1,0 +1,1 @@
+Fixed Linux debug/static builds aborting when a dynamically dispatched one-argument closure threw into a surrounding `try`/`catch`. Dynamic `Function.apply` deferrals now reach the generated catch landing pad and report the located ahead-of-time limitation instead of terminating the process.
--- a/crates/perry-runtime/src/closure/dispatch/calln.rs
+++ b/crates/perry-runtime/src/closure/dispatch/calln.rs
@@ -35,7 +35,9 @@ pub extern "C" fn js_closure_call0(closure: *const ClosureHeader) -> f64 {
 
 /// Call a closure with 1 argument, returning f64
 #[no_mangle]
-pub extern "C" fn js_closure_call1(closure: *const ClosureHeader, arg0: f64) -> f64 {
+// The one-argument value-call path can run arbitrary generated code and must
+// let a JS exception unwind to the generated caller's catch landing pad.
+pub extern "C-unwind" fn js_closure_call1(closure: *const ClosureHeader, arg0: f64) -> f64 {
     let func_ptr = get_valid_func_ptr(closure);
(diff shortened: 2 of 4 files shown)


proggeramlug:
Validated as part of a five-PR batch (#8419, #8416, #8417, #8420, #8421) stacked on
`d3ea711b3` and built once, then merged individually.

PR #8417, opened 2026-08-19
perf(codegen): retain string accumulators in concat chains

### LINKED ISSUE #8394: perf: no rope/cons-string representation makes accumulator concatenation quadratic (514x Node at n=16000; 17.9% of iso_miss)
## No rope / cons-string representation: accumulator concatenation is quadratic

`grep -riE '\brope\b|cons_?str' crates/perry-runtime/src` returns **zero** hits. Every
string concatenation materializes bytes eagerly, so the standard accumulate-in-a-loop
pattern is O(n^2) where V8's cons-strings make it O(n) amortized.

This is the **largest single cost in the worst benchmark row**. `iso_miss` (Perry 2.13x
Node, the widest remaining gap) spends **17.9% of self time in `js_string_concat_chain`**,
driven by one line:

```ts
seen = seen + "[" + names[i] + "]";   // iso_miss.ts:205, per name per env frame
```

`concat_chain` itself is already well optimized — #7912 fuses the chain and sizes the
scratch to the real arity, and its comment cites this exact line. The remaining cost is
not call overhead, it is copying the accumulated prefix on every append.

## Measured

```ts
function build(n: number): number {
  let s = "";
  for (let i = 0; i < n; i++) s = s + "[" + "abc" + "]";
  return s.length;
}
```

| n | Perry | Node 26.5.1 |
|---:|---:|---:|
| 2000 | 11 ms | 1 ms |
| 4000 | 63 ms | 0 ms |
| 8000 | 459 ms | 1 ms |
| 16000 | **1888 ms** | **1 ms** |

Perry's time grows ~4-7x per doubling (quadratic); Node is flat.

**This is not Node deferring the work.** Forcing full materialization by reading every
997th code unit out of the result:

| n | Perry | Node |
|---:|---:|---:|
| 2000 | 10 ms | 1 ms |
| 16000 | **2057 ms** | **4 ms** |

Node still finishes in 4 ms, so its cons-string plus a single flatten is genuinely linear.
At n=16000 Perry is **514x slower** on the same work.

## Suggested direction

A cons-string node (`left`, `right`, total length) with lazy flattening on first indexed
read is the standard fix and is what V8/JSC do:

- `+` on two strings allocates an O(1) cons node instead of copying
- `.length` is O(1) from the stored total
- the first operation needing contiguous bytes flattens once, O(n)

The parts that already exist and would need to participate: `js_string_concat_chain`
(`crates/perry-runtime/src/string/concat.rs`), `js_string_equals`, `js_string_compare`,
and `js_get_string_pointer_unified` — anything reading raw bytes needs a flatten call
first. The GC also needs to trace the two child edges of a cons node.

This is a real architectural change rather than a local optimization, which is why I am
filing it rather than attempting it: it touches the string representation everywhere. But
it is the single biggest remaining item on the worst row, and accumulator concatenation is
one of the most common patterns in real JavaScript.

proggeramlug:
## Summary

- recognize loop-local `s = s + a + b + c` as the existing safe self-append shape
- retain the growing accumulator and fuse only the fixed-size suffix
- preserve numeric `+` semantics for opaque head pairs and immutable string aliases

Perry already has a unique-owner `js_string_append` path with geometric capacity growth. The n-way concat fold bypassed it by materializing the accumulator together with every suffix piece. Keeping the prefix out of that fold removes the quadratic copy without adding a new string representation or exposing cons nodes to raw-byte readers. The new selection is limited to loop bodies, where the repeated-copy cost exists.

Closes #8394.

## Measured

Forced-materialization benchmark (including indexed `charCodeAt` reads), one build per size:

| iterations | main | patched |
|---:|---:|---:|
| 2,000 | 1 ms | 0 ms |
| 4,000 | 15 ms | 1 ms |
| 8,000 | 35 ms | 0 ms |
| 16,000 | 124 ms | 1 ms |

The patched executable used the less-optimized `perry-dev` profile; the main executable used release, so the result is conservative. Output/checksums matched. A 20-repetition version of the issue workload completed in 2/4/14/15 ms across 2k/4k/8k/16k.

## Validation

- `cargo test -p perry-codegen --lib`: 1,105 passed
- focused IR regressions: accumulator suffix fusion and numeric-capable head pair
- compiled regression fixture matched Node 26.5.1 byte-for-byte
- optimized `perry` + runtime/stdlib static-wrapper build succeeded

The broader `cargo test -p perry --bin perry` build was attempted but the separate test-profile dependency graph exhausted the host disk before linking (`errno=28`); no test in that command ran or failed.

No version bump or central changelog edit is included.

## Summary by CodeRabbit

* **Performance**
  * Improved repeated string concatenation in loops, reducing quadratic growth and significantly speeding up large string builds.
  * Optimized accumulator patterns while preserving efficient handling of fixed suffixes.

* **Bug Fixes**
  * Preserved aliases correctly after string appends.
  * Ensured string-typed values retain correct runtime behavior even when backed by numeric values.

* **Tests**
  * Added coverage for self-appending strings, large repeated builds, and mixed numeric/string scenarios.

### DIFF
--- a/changelog.d/8417-string-accumulator-chains.md
+++ b/changelog.d/8417-string-accumulator-chains.md
@@ -1,0 +1,6 @@
+### Performance
+
+- Keep the growing prefix of loop-local string accumulator chains on Perry's
+  amortized append path while fusing only the fixed-size suffix. The reported
+  16,000-iteration forced-read benchmark drops from 124 ms to 1 ms and no
+  longer exhibits quadratic growth.
--- a/crates/perry-codegen/src/codegen/declared_string_add_tests.rs
+++ b/crates/perry-codegen/src/codegen/declared_string_add_tests.rs
@@ -42,13 +42,17 @@ fn param(id: u32, name: &str, ty: Type) -> Param {
 }
 
 fn probe_fn(params: Vec<Param>, body: Expr) -> Function {
+    probe_fn_with_body(params, vec![Stmt::Return(Some(body))])
+}
+
+fn probe_fn_with_body(params: Vec<Param>, body: Vec<Stmt>) -> Function {
     Function {
         id: 1,
         name: "probe".to_string(),
         type_params: Vec::new(),
         params,
         return_type: Type::Any,
-        body: vec![Stmt::Return(Some(body))],
+        body,
         is_async: false,
         is_generator: false,
         is_strict: true,
@@ -98,6 +102,10 @@ fn module_with(function: Function) -> Module {
 
 fn ir(params: Vec<Param>, body: Expr) -> String {
     let module = module_with(probe_fn(params, body));
+    function_ir(module)
+}
+
+fn function_ir(module: Module) -> String {
     let ir =
         String::from_utf8(compile_module(&module, ir_opts()).unwrap()).expect("LLVM IR is UTF-8");
     // An ordinary typed parameter may now produce a public guard wrapper plus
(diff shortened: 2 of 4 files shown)


proggeramlug:
Validated in a five-PR batch. **This closes out #8394's quadratic blow-up.**

Re-ran the exact scaling fixture from that issue — `s = s + "[" + "abc" + "]"` in a loop,
20 repetitions per size:

| n | main (`d3ea711b3`) | with this PR | Node 26.5.1 |
|---:|---:|---:|---:|
| 2000 | 17 ms | **3 ms** | 1 ms |
| 4000 | 56 ms | **7 ms** | 1 ms |
| 8000 | 342 ms | **18 ms** | 1 ms |
| 16000 | **1450 ms** | **16 ms** | 1 ms |

**90x faster at n=16000**, and — more importantly — the curve is flat rather than roughly
quadrupling per doubling. That was the actual defect in #8394: not the constant, the shape.

Notably this achieves it by retaining the accumulator rather than by implementing cons
strings, so it sidesteps the blast radius the #8394 design survey found (174 runtime files /
356 sites doing explicit `StringHeader` payload arithmetic). Still ~16x off Node in absolute
terms, so the deeper representation question stays open — but the pathological case that
made `iso_miss` the worst row is gone.

## Batch validation

- 19/19 corpus byte-exact · `perry-runtime --lib` 2599 · `perry --bin perry` 1005 · all 50 gates

## Unrelated flake seen during validation

`promise::native_async::tests::main_thread_token_reject_string_wrong_thread_uses_wrong_thread_reason`
failed 2 of 9 full-suite runs on the batch — and **1 of 8 on unmodified `main`**, while never
failing in isolation on either. Pre-existing test-isolation issue, filed as #8435. Not yours.

PR #8418, opened 2026-08-19
perf(compile): reduce generated bundle bloat

proggeramlug:
## Summary

- remove obsolete program-wide class augmentation for unresolved type-only interface receivers now that dynamic calls use the runtime class-vtable registry
- synthesize JSON imports through `JSON.parse` of a serialized constant instead of lowering every property as allocation/store IR
- add explicit opt-in generated-code size optimization (`PERRY_LL_SIZE_OPT=1`) while preserving Perry's normal `-O3` default
- add an opt-in per-function `optnone` safeguard so dense bundles can size-optimize ordinary functions without sending giant generated siblings through LLVM's super-linear middle-end
- key both build and object caches on the new codegen settings

## Measurements

OpenCode TypeScript source build on Windows: 4,743 native modules / 0 JavaScript fallback.

Structural changes alone (`PERRY_LL_SIZE_OPT=0`):

- executable: 1,685.4 -> 1,143.7 MiB (-541.7 MiB / -32.1%)
- `.text`: 1,531,310,038 -> 1,086,252,544 bytes
- oversized O0 units: 902 -> 315
- module codegen: 73.2 -> 63.8 minutes (-12.8%)
- `mime-db/db.json` object: 17,270,002 -> 448,436 bytes (-97.4%)

Full hybrid build on current `origin/main` (`PERRY_LL_SIZE_OPT=1`, `PERRY_LL_PREOPT_OPTNONE_INSTRS=8192`):

- executable: 1,685.4 -> 797.0 MiB (-888.4 MiB / -52.7%)
- `.text`: 1,531,310,038 -> 721,355,264 bytes (-52.9%)
- TypeScript object: 374.7 -> 277.4 MiB (-26.0%)
- Prettier plus parser objects: 304.5 -> 194.6 MiB (-36.1%)
- module codegen: 73.2 -> 88.8 minutes (+21.3%); the hybrid setting trades compile time for another 346.7 MiB beyond the structural build
- warm-cache rebuild: all 4,743 module objects reused in 2.5 minutes; full relink completed successfully

Isolated `prettier/plugins/typescript.mjs` calibration:

- all O0: 36.6 MiB object, 27.7 s module stage
- existing mixed O0/Os policy: 28.5 MiB, 66.2 s
- forced Os + new 8K-instruction per-function safeguard: 25.6 MiB, 76.1 s

The hybrid path remains opt-in while corpus calibration continues. The linked OpenCode executable reaches the same pre-existing startup `TypeError: value is not a function` as the baseline and structural builds.

## Test plan

- `cargo test --release -p perry-codegen preoptimization_bloated_function_is_demoted_without_demoting_its_sibling --lib`
- `cargo test --release -p perry-codegen linker::tests::size_optimization_flag_is_explicit_and_truthy --lib`
- `cargo test --release -p perry --bin perry codegen_env_vars_are_build_cache_inputs`
- `cargo test --release -p perry --bin perry key_changes_with_codegen_env_vars`
- `cargo test -p perry --test source_graph_export_regressions type_only_interface_dispatch_uses_runtime_class_registry -- --exact`
- `cargo test -p perry --test source_graph_export_regressions json_module_parses_embedded_serialized_data -- --exact`
- `cargo fmt --check -p perry-codegen`
- `cargo fmt --check -p perry`
- full current-main OpenCode source compile plus successful warm-cache relink and startup smoke (4,743 native modules)

`cargo test -p perry-codegen --lib` is 1,101 passed / 3 failed locally; all three failures reproduce unchanged on a clean `origin/main` worktree on Windows. Debug test linking on this host also lacks unused all-target symbols in its LLVM-C static library; the same focused tests pass in release mode.

## Summary by CodeRabbit

* **New Features**
  * Added optional size-optimized native builds for smaller output.
  * Added configurable handling for unusually large functions to improve build efficiency.
  * JSON imports now preserve serialized data and parse correctly at runtime.

* **Bug Fixes**
  * Improved type-only interface dispatch, including captured method references.
  * Fixed JSON module handling for nested data, arrays, Unicode, and other valid JSON values.

* **Performance**
  * Reduced native output size and module code-generation time in supported configurations.
  * Build caches now correctly refresh when optimization settings change.


proggeramlug:
Validated and merging. I also measured the one thing the PR does not: **what
`PERRY_LL_SIZE_OPT=1` costs at runtime.**

## The size knob is runtime-free

Measured on the quiet M1 mini under the sweep's bench lock (interleaved, output-checked,
verdict `CLEAN`, load 1.51 -> 1.72), same binaries built both ways:

| row | Δ best | Δ median | disjoint? |
|---|---:|---:|---|
| churn | +0.5% | +0.7% | no |
| cycles | -0.2% | -0.1% | no |
| fib40 | +0.0% | -0.1% | no |
| interp | +0.8% | +0.8% | no |
| iso_miss | +0.6% | +0.8% | no |
| tree_wide | +0.0% | +0.0% | no |

Nothing disjoint, everything within 0.8%. `-Os` is not costing measurable runtime speed on
this corpus.

(I first measured this on a contended dev box and got swings from -26% to +21% with signs
flipping between best and median. That was pure host noise — worth flagging because a ratio
measured on a loaded machine is not a ratio, which is also why the `0.81x Node` figure in
#8412 did not survive scrutiny.)

## Consequence: the maintainer has decided this should default ON

Given the measured tradeoff is **compile time vs binary size, with no runtime cost**, the
knob is being flipped to default-on in a follow-up. Less shipped binary for free is exactly
the intended direction; +21.3% compile time on a 4,743-module build is an acceptable
developer-side cost.

Worth considering as a refinement: on small programs `-Os` saves ~0% (I measured hello-world
and three corpus binaries at 0.00 to 0.24% *larger*), so a size-driven policy that engages
only where there are oversized units would get the bundle win without paying compile time on
every trivial build — close to what your per-function `optnone` safeguard already does.

## Structural changes verified

The JSON change is the one with real semantic risk — swapping property-by-property IR for
`JSON.parse` of a serialized constant. Diffed against Node on a fixture covering escapes and
backslashes, `-0` (serializes as `0`, distinguished by `Object.is`), `1e21` / `-1e-7`
exponent formatting, an integer past 2^53 losing precision, non-BMP emoji, empty object vs
empty array, `null`, prototype identity and frozen-ness: **byte-identical to Node**.

- `perry-runtime --lib` 2599 · `perry --bin perry` 1007 · `perry-codegen --lib` 1107
- all 50 gates

Thank you for stating plainly that the OpenCode executable still hits the same pre-existing
startup `TypeError` as the baseline — that is the right way to report a size win that is not
also a correctness win.

PR #8419, opened 2026-08-19
test(runtime): make intrusive free-list test concurrency-safe

### LINKED ISSUE #8401: flaky: the_intrusive_free_list_round_trips_a_whole_cohort asserts a delta on a process-global counter under parallel tests

proggeramlug:
## Summary

- replace process-global allocation-counter deltas with address-set assertions local to the test thread
- verify the reused cohort exactly matches the minted cohort and that the next allocation is outside the drained cohort
- remove the counter mutex, which could not exclude ordinary box allocations in parallel tests

## Testing

- `cargo test --release -p perry-runtime --lib r#box::release_tests::the_intrusive_free_list_round_trips_a_whole_cohort -- --exact`
- `cargo test --release -p perry-runtime --lib` (full parallel suite, then 8 consecutive repeat passes)

No version bump.

Fixes #8401.

## Summary by CodeRabbit

* **Tests**
  * Updated memory allocation tests to verify that reused allocations receive valid addresses.
  * Simplified test execution by removing unnecessary counter synchronization.
  * Retained existing validation for allocation residue counters.

### DIFF
--- a/crates/perry-runtime/src/box/release_tests.rs
+++ b/crates/perry-runtime/src/box/release_tests.rs
@@ -13,24 +13,6 @@ fn install_test_activation(activation: *mut AsyncBoxActivation) -> crate::promis
     })
 }
 
-/// `BOX_ALLOC_COUNT` / `BOX_POOL_REUSE_COUNT` / `BOX_RELEASE_COUNT` are
-/// process-global atomics, while the registries, quarantines and free
-/// lists they describe are THREAD-LOCAL. Any test that asserts on a
-/// counter *delta* is therefore not isolated by `test_clear_box_registry`
-/// alone — a sibling test allocating on another harness thread lands in
-/// the same atomics and moves the delta under it. Observed exactly that:
-/// these tests pass under `--test-threads=1` and fail in parallel.
-///
-/// Serialise the counter-asserting tests against each other. Tests that
-/// only assert on addresses and registry membership are thread-local and
-/// need no lock.
-fn counter_guard() -> std::sync::MutexGuard<'static, ()> {
-    static LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
-    // A panicking test poisons the lock; the data is `()`, so recovering
-    // is right — otherwise one failure cascades into spurious ones.
-    LOCK.lock().unwrap_or_else(|e| e.into_inner())
-}
-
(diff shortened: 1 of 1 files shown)


proggeramlug:
Validated as part of a five-PR batch (#8419, #8416, #8417, #8420, #8421) stacked on
`d3ea711b3` and built once, then merged individually.

PR #8439, opened 2026-08-20
fix(runtime): root string copy sources across allocation

### LINKED ISSUE #8423: runtime: string payload pointer crosses the allocation that copies it (js_string_to_well_formed fast path, structured-clone string copy)
## Problem

Two runtime sites compute a raw pointer into a string's payload and then pass it to `js_string_from_bytes`, whose **internal allocation happens before its memcpy reads the source**. If that allocation triggers a collection that moves the source string, the copy reads from the old (from-space) address.

**Site 1 — `js_string_to_well_formed` fast path**, `crates/perry-runtime/src/string/compare.rs:778-783`:

```rust
let blen = unsafe { (*s).byte_len } as usize;
let data = string_data(s);                          // raw ptr into s's payload
if flags & STRING_FLAG_HAS_LONE_SURROGATES == 0 {
    return js_string_from_bytes(data, blen as u32); // allocates, THEN copies from data
}
```

(The slow arm below it is fine — it builds a Rust `Vec` from the borrow *before* any GC allocation.)

**Site 2 — structured-clone string copy**, `crates/perry-runtime/src/builtins/globals.rs:749-754`:

```rust
let len = (*str_ptr).byte_len as usize;
let data = (str_ptr as *const u8).add(std::mem::size_of::<StringHeader>());
let new_str = js_string_from_bytes(data, len as u32);   // same shape
```

Perry strings are `STRING_TAG` NaN-boxes pointing at a 20-byte `StringHeader` with the WTF-8 payload inline (`crates/perry-runtime/src/string/mod.rs:307`). The copying minor relocates cells and rewrites the *slots* it knows about; a raw `*const u8` into a payload is invisible to it. The canonical statement is the `HeapKeyBytes` doc (`crates/perry-runtime/src/object/field_get_set.rs:11-27`): "The only sound shape is to stop borrowing." Precedent: #5062 (chunked slice/append corruption), #7341 (`across_const`).

## Why this is latent rather than live — and why fix it anyway

The window's only collection point is the allocation inside `js_string_from_bytes` itself. Allocation-point triggers currently force a conservative stack scan, which makes the copying minor ineligible — nothing moves. This is the exact "closed by accident" accident documented at length in `crates/perry-runtime/src/string/alloc.rs:22-51` (#7213): it holds only while alloc-point collections stay non-moving, which is the property the moving-GC roadmap keeps eroding. These two sites should not depend on it.

## Suggested fix

Use the existing GC-safe primitive: `string_copy_range` (`crates/perry-runtime/src/string/mod.rs:704`) roots the source, allocates, re-reads the source pointer after the allocation, then copies. Both sites become one call. Match its exact signature; it already has 6 callers to crib from. Alternatively the `across_const` pairing (`string/append.rs:44-46`) — pick whichever reads cleaner, but do not leave a raw source pointer live across the allocation.

Preserve behavior details: site 1 must still propagate nothing extra (the fast path is a plain byte copy of a well-formed string); site 2 must keep the NaN-boxing of the result unchanged.

## Validation

- Existing fixtures must stay green: `toWellFormed()` tests and structuredClone tests in `test-files/` (grep for `toWellFormed` / `structuredClone`).
- Instrument run (proves the pattern, not the trigger, since the window is currently non-moving): `PERRY_GC_SCHEDULE_SEED=1 PERRY_GC_SCHEDULE_RATE=1 PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1 PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_SCHEDULE_ALLOC_KB=0` over a fixture that calls both functions in a loop; with `PERRY_GC_DIAG=1` confirm `[gc-fromspace-protect] retired_set=` lines appear so the run is not vacuous.
- Build note: integration tests and `perry compile` link the static wrapper archives — build `cargo build --profile perry-dev -p perry -p perry-runtime-static -p perry-stdlib-static`, not just `-p perry-runtime`, or you will test a stale `libperry_runtime.a`. Run perry-runtime unit tests with `RUST_TEST_THREADS=1`.

Sibling issues (same class, different trigger — theirs run **user JS** inside the window, which CAN move today): `js_string_normalize`, `js_string_repeat`, `js_regexp_exec` — filed separately.

## Workflow

proggeramlug:
Closes #8423

## Summary

- route the toWellFormed fast-path copy through the GC-safe string range copier
- root and refresh structuredClone string sources across destination allocation
- add a focused allocation/evacuation stress fixture covering both paths

## Validation

- RUST_TEST_THREADS=1 cargo test -p perry-runtime --lib (2599 passed, 4 ignored)
- cargo build --profile perry-dev -p perry -p perry-runtime-static -p perry-stdlib-static
- focused fixture compiled and passed normally
- focused fixture passed with forced evacuation, evacuation verification, protected from-space, and allocation scheduling; 1005 copying minors ran and gc-fromspace-protect retired-set diagnostics were observed

No version bump.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `toWellFormed()` and `structuredClone()` string copying to preserve content and character metadata.
  * Prevented corrupted or mismatched string results during garbage collection and memory allocation.

* **Tests**
  * Added stress coverage for string-copy operations involving astral characters and forced garbage collection.


proggeramlug:

reproducers.

PR #8440, opened 2026-08-20
fix(runtime): own punycode string bytes before decoding

### LINKED ISSUE #8424: runtime: punycode string_bytes manufactures &'static [u8] over a movable GC payload
## Problem

`crates/perry-runtime/src/punycode.rs:334-339` manufactures a `&'static [u8]` over a movable GC payload:

```rust
fn string_bytes(ptr: *const StringHeader) -> &'static [u8] {
    if !crate::string::is_valid_string_ptr(ptr) {
        return &[];
    }
    unsafe { slice::from_raw_parts(string_data(ptr), (*ptr).byte_len as usize) }
}
```

Perry strings live in the moving GC's young generation (20-byte `StringHeader`, payload inline at +20, `crates/perry-runtime/src/string/mod.rs:307`); the copying minor relocates them and rewrites only the *slots* the collector knows about. A `&'static` borrow is the maximally wrong lifetime: it type-erases the hazard so callers can hold the slice across any allocation without a compiler complaint. The repo's written rule for this class is the `HeapKeyBytes` doc (`crates/perry-runtime/src/object/field_get_set.rs:11-27`): "The only sound shape is to stop borrowing. Copy the bytes out once, before the arm's first allocation."

Whether any current caller in `punycode.rs` actually allocates while holding the slice needs a one-file audit (the helper has a handful of callers, all in this file) — but the fix is the same either way, and the `'static` signature guarantees the next caller gets it wrong silently.

## Suggested fix

1. Audit the callers of `string_bytes` in `punycode.rs`. Punycode inputs are domain labels — small.
2. Replace the borrow with a copy-first entry: copy the bytes into an owned `Vec<u8>`/`SmallVec` (or reuse the `HeapKeyBytes` pattern: 64-byte inline buffer, `Vec` spill) at each function's entry, before any allocation can run, and delete `string_bytes`.
3. If a caller provably never allocates between borrow and last use, still convert it — the point is removing the `'static` landmine, and the copies here are a few dozen bytes.

## Validation

- Existing punycode/URL fixtures stay green (grep `test-files/` for `punycode` / IDN / `toASCII`).
- `RUST_TEST_THREADS=1 cargo test --release -p perry-runtime punycode` (perry-runtime tests are not parallel-safe).
- Build note: for any end-to-end check, build `cargo build --profile perry-dev -p perry -p perry-runtime-static -p perry-stdlib-static` — building only `-p perry-runtime` leaves a stale `libperry_runtime.a` linked into test binaries.

## Workflow

proggeramlug:
## Summary

Remove the unbounded borrow of movable GC string payloads from `punycode.ucs2.decode`. The decoder now copies heap bytes into owned storage before any later runtime allocation can relocate the source string.

## Changes

- Replace the `&'static [u8]` heap-payload helper with an owned `Vec<u8>` copy.
- Decode UTF-16 units exclusively from the owned copy.
- Add a regression test proving the copied bytes remain independent of the heap payload.
- Add a changelog fragment; no version metadata was changed.

## Related issue

Fixes #8424

## Test plan

- [x] `RUST_TEST_THREADS=1 cargo test --release -p perry-runtime punycode` (5 passed)
- [x] Added a `#[test]` in `perry-runtime`
- [ ] Full workspace build/test suite (left to CI)

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Unicode decoding reliability by preventing string data from being affected by later runtime changes.
  * Fixed a potential memory-safety issue when decoding malformed or unusually sized string payloads.
* **Tests**
  * Added regression coverage to verify decoded data remains stable after the source string is modified.
* **Documentation**
  * Documented the Unicode decoding fix in the changelog.


proggeramlug:

reproducers.

PR #8441, opened 2026-08-20
fix(container): remove legacy compose FFI

### LINKED ISSUE #8422: container-compose: legacy `ffi` feature mirrors StringHeader as 4 bytes (payload at +4 vs the 20-byte ABI) — delete or repair
## Problem

`perry-container-compose` carries a private re-implementation of the runtime string ABI that is **wrong**, guarded only by a legacy feature flag that nothing in-tree enables:

```rust
// crates/perry-container-compose/src/ffi.rs:15-18
#[repr(C)]
pub struct StringHeader {
    pub length: u32,          // 4 bytes
}
```

The real runtime header is **20 bytes** — `{ utf16_len: u32, byte_len: u32, capacity: u32, refcount: u32, flags: u32 }`, WTF-8 payload immediately after, ABI-pinned by a const assertion (`crates/perry-runtime/src/string/mod.rs:307-342`).

Consequences inside the `ffi` module (`crates/perry-container-compose/src/ffi.rs`):

- `string_from_header` (`ffi.rs:20-28`) reads a real runtime string's payload at `+4` — i.e. it interprets `byte_len`, `capacity`, `refcount`, `flags` as text bytes, and reads `utf16_len` (not `byte_len`) as the length.
- `heap_string` (`ffi.rs:45-57`) mallocs a 4-byte-header string with **no GC header** and returns it as `*const StringHeader`. Any runtime consumer would read garbage for `byte_len`/`capacity`/`flags` and a garbage GC header at `-8`.
- The module exports `js_compose_start` / `js_compose_stop` / `js_compose_ps` (`ffi.rs:98+`) — the **same symbol names** perry-stdlib exports with correct signatures (`crates/perry-stdlib/src/container/compose_ffi.rs:200-257`). Linking both is a duplicate-symbol error; the stdlib Cargo.toml already documents this ("NOTE 2: do NOT enable the crate's own `ffi` feature here — it exports a *different* legacy `js_compose_*` shape … that would collide with stdlib's canonical SPEC §9.1 stack-handle signatures at link", `crates/perry-stdlib/Cargo.toml:330-339`).

## Current exposure

Not linked into shipping builds: the module is `#[cfg(feature = "ffi")]` (`crates/perry-container-compose/src/lib.rs:32`), the feature is `ffi = []` marked "legacy YAML-path shape" (`crates/perry-container-compose/Cargo.toml:43`), and no in-tree crate enables it. It is one `features = ["ffi"]` away from silent string corruption, and the mirror has already drifted from the real ABI unnoticed — which is the tell that no test covers it.

## Suggested fix

Per the repo's kill-policy (CLAUDE.md: "a mode that still exists is a decision that hasn't been made" — an unexercised configuration should stop compiling):

**Preferred: delete the legacy `ffi` feature and module entirely.** The canonical FFI surface lives in `crates/perry-stdlib/src/container/compose_ffi.rs`; the legacy YAML-path shape has no in-tree consumer. Remove `ffi.rs`, the `ffi = []` feature, the `#[cfg(feature = "ffi")] mod ffi;` line, and the NOTE 2 warning in stdlib's Cargo.toml (it becomes moot — say so in the changeset).

If some out-of-tree consumer is discovered that needs it (check with the maintainer before assuming): fix the mirror to the full 5×u32 layout, add a const assertion identical to `STRING_HEADER_ABI_MATCHES_CODEGEN` (`string/mod.rs:337`), allocate through the runtime allocator instead of `std::alloc`, and rename the exports so they cannot collide with stdlib's.

## Validation

- `cargo check -p perry-container-compose` and `cargo check -p perry-container-compose --features ffi` (the latter should fail to resolve the feature after deletion).
- `cargo build --profile perry-dev -p perry -p perry-runtime-static -p perry-stdlib-static` still builds; container tests in perry-stdlib (feature `container`) still pass.
- Grep the tree for `feature = "ffi"` references to perry-container-compose to confirm nothing dangles.

## Workflow

PR = code + tests + `changelog.d/<PR>-<slug>.md` fragment; no version bump (maintainer bumps at merge).

proggeramlug:
## Summary

Delete `perry-container-compose`'s unused legacy FFI feature and module, whose private 4-byte string header was incompatible with the runtime's 20-byte string ABI. The canonical stack-handle compose FFI in `perry-stdlib` remains unchanged.

## Changes

- Remove the legacy `ffi` Cargo feature, module gate, and duplicate `js_compose_*` exports.
- Remove the now-moot collision warning from `perry-stdlib`'s dependency declaration.
- Add a structural regression test that prevents the legacy feature/module from being reintroduced.

## Related issue

Fixes #8422

## Test plan

- [x] `cargo check -p perry-container-compose`
- [x] Confirmed `cargo check -p perry-container-compose --features ffi` now fails because the feature no longer exists.
- [x] `cargo test -p perry-container-compose` (171 tests passed)
- [x] Canonical stdlib compose FFI smoke tests passed within `cargo test -p perry-stdlib --features container --tests` (129 lib tests passed). The run then stopped on six existing promise-state assertions in `container_backend_selection`; the same six failures reproduce on untouched `main` at `526e0b502`.

- [ ] (if CLI / stdlib / runtime API changed) Updated `docs/src/` (not applicable; canonical API is unchanged)

## Screenshots / output

Not applicable.

## Checklist

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed the legacy Compose FFI interface and duplicate JavaScript Compose exports.
  * Compose FFI access remains available through the supported standard-library interface.

* **Documentation**
  * Added a changelog entry describing the legacy interface removal.

* **Tests**
  * Added regression coverage to ensure the removed legacy interface does not return.


proggeramlug:

reproducers.

PR #8442, opened 2026-08-20
fix(ffi): guard string header ABI revision

### LINKED ISSUE #8425: ffi: add STRING_HEADER_ABI_REVISION tripwire to perry-ffi (object header has one, string header does not)
## Problem

`perry-ffi` is the public ABI for out-of-tree `perry.compilePackages` native wrappers. It mirrors the runtime's `StringHeader` (`crates/perry-ffi/src/types.rs:34`, 5×u32 = 20 bytes) and hands out payload access via `read_string` / `read_bytes` (`crates/perry-ffi/src/lib.rs:212` / `:225`), which open-code `byte_len` + `hdr + size_of::<StringHeader>()`.

For the **object** header there is a version tripwire: `OBJECT_HEADER_ABI_REVISION: u32 = 3` (`types.rs:30`), checked against the runtime's `perry_object_header_abi_revision()` so a wrapper compiled against a stale layout fails loudly at registration instead of silently corrupting (mechanism documented at `types.rs:22` and `:70`).

**There is no string equivalent.** Any future change to `StringHeader` — size, field order, or a representation change (rope/cons, slice views) — silently miscompiles every out-of-tree wrapper: `read_bytes` would return garbage with no diagnostic. The in-tree side is protected by `STRING_HEADER_ABI_MATCHES_CODEGEN` (`crates/perry-runtime/src/string/mod.rs:337`); the out-of-tree side has nothing.

## Suggested fix

Mirror the object-header mechanism exactly:

1. Add `pub const STRING_HEADER_ABI_REVISION: u32 = 1;` to `crates/perry-ffi/src/types.rs` with the same style of doc comment explaining the contract (bump on ANY layout/representation change to `StringHeader` or to the meaning of `read_bytes`' return).
2. Add a runtime-side `#[no_mangle] pub extern "C" fn perry_string_header_abi_revision() -> u32` next to wherever `perry_object_header_abi_revision()` lives, returning the same constant (duplicated literal + a cross-check test, since perry-ffi and perry-runtime deliberately share no crate dependency for this).
3. Wire the check into the same wrapper-registration path that validates `OBJECT_HEADER_ABI_REVISION` (follow the existing call site of `perry_object_header_abi_revision` — the string check belongs in the same place with the same failure mode).
4. Add a test in perry-ffi (or wherever the object-revision cross-check test lives — find it by grepping for `OBJECT_HEADER_ABI_REVISION`) asserting the two sides agree, plus a `const` assert in perry-ffi that `size_of::<StringHeader>() == 20` so a drift of the mirror itself fails the build (the runtime already has this; the mirror does not).

## Also worth doing in the same PR

`read_string`/`read_bytes` return `&'static` borrows into the moving GC heap. Fixing the lifetime is a breaking API change and NOT in scope here — but the doc comment currently hand-waves ("The borrow lives as long as the runtime guarantees the string remains alive", `lib.rs:222-224`). Tighten it to state the actual contract: the borrow is valid only until the next allocation through the Perry runtime, because the GC moves strings; callers must copy before calling back into any allocating runtime function.

## Validation

- `cargo test -p perry-ffi`; the new cross-check test fails if either side's constant is bumped alone.
- `cargo build --profile perry-dev -p perry -p perry-runtime-static -p perry-stdlib-static` clean.

## Workflow

proggeramlug:
## Summary

Add a public string-header ABI revision tripwire so out-of-tree `perry-ffi` wrappers can detect incompatible runtime layouts before reading corrupt payload data.

## Changes

- publish `STRING_HEADER_ABI_REVISION` from `perry-ffi` and export the matching `perry_string_header_abi_revision()` runtime symbol
- pin the FFI mirror to 20 bytes at compile time and cross-check the runtime revision plus all published offsets under `runtime-link`
- document that `read_string` / `read_bytes` borrows become invalid at the next allocating Perry runtime call
- add the required changelog fragment without changing the workspace version

The issue mentions wiring this into the object header's wrapper-registration path. There is no such registration path in the current tree: the existing object guard is the public constant/runtime symbol contract plus the `runtime-link` equality test. This change mirrors that mechanism exactly for strings.

## Related issue

Fixes #8425

## Test plan

- [x] `cargo test -p perry-ffi`
- [x] `cargo test -p perry-ffi --features runtime-link --lib`
- [x] Verified `libperry_runtime.a` exports `perry_string_header_abi_revision`

## Checklist

- [x] My commit follows the repository's loose conventional-commit style

## Summary by CodeRabbit

* **Bug Fixes**
  * Added safeguards to detect mismatches in string and byte data layout between the runtime and interface.
  * Improved documentation clarifying that borrowed string and byte data must be copied before operations that may allocate memory.

* **Tests**
  * Added validation for layout size, field offsets, payload representation, and ABI revision compatibility.

### DIFF
--- a/changelog.d/8442-string-header-abi-tripwire.md
+++ b/changelog.d/8442-string-header-abi-tripwire.md
@@ -1,0 +1,7 @@
+### fix(ffi): guard the published string-header ABI against runtime drift
+
+`perry-ffi` now publishes a string-header ABI revision paired with the runtime's
+exported revision symbol, and tests pin both revisions and the 20-byte layout.
+Out-of-tree native wrappers can fail loudly on an incompatible runtime instead
+of reading corrupt string payloads. The borrowed string/byte helpers also now
+document that moving-GC borrows must be copied before the next runtime allocation.
--- a/crates/perry-ffi/src/lib.rs
+++ b/crates/perry-ffi/src/lib.rs
@@ -56,7 +56,7 @@ pub use async_runtime::{
 mod types;
 pub use types::{
     ArrayHeader, BigIntHeader, BufferHeader, ClosureHeader, NativeAsyncCompletion, ObjectHeader,
-    Promise, StringHeader, BIGINT_LIMBS, OBJECT_HEADER_ABI_REVISION,
+    Promise, StringHeader, BIGINT_LIMBS, OBJECT_HEADER_ABI_REVISION, STRING_HEADER_ABI_REVISION,
 };
 
 mod handle;
@@ -192,10 +192,10 @@ pub fn alloc_string(s: &str) -> JsString {
 
 /// Read a `JsString` as a borrowed `&str`.
 ///
-/// Returns `None` on a null handle or invalid UTF-8. The borrow lives
-/// as long as the runtime guarantees the string remains alive — for
-/// the simple call-and-copy pattern in most FFI functions, that's the
-/// duration of the function call.
+/// Returns `None` on a null handle or invalid UTF-8. Because Perry's GC can
+/// move strings, the borrow is valid only until the next allocation through
+/// the Perry runtime. Copy the contents before calling any runtime function
+/// that may allocate.
 ///
 /// ```ignore
 /// #[no_mangle]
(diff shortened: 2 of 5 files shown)


proggeramlug:

reproducers.

Issue #8428, opened 2026-08-19
runtime: js_regexp_exec holds the subject payload borrow across the lastIndex Get/ToLength user side effects


proggeramlug:
## Problem

`js_regexp_exec` borrows the subject string's payload, then observes `lastIndex` — a Get → ToLength that, per the function's own spec comment, deliberately surfaces user `valueOf`/`toString` side effects — while holding the borrow (`crates/perry-runtime/src/regex/exec.rs:36-75`):

```rust
let str_data = string_as_str(s);                 // :36 — borrow of s's payload
...
// "the read — and any valueOf/toString side effect of a coercible
//  lastIndex — is observed exactly once" (test262 …lastindex-access)
let last_index_read = regex_last_index_offset(re);  // :51 — can run user JS
...
super::exec_array::utf16_index_to_byte(str_data, last_index)  // :61 — reads borrow
...
let search_str = &str_data[search_start_byte..];              // :75 — reads borrow
```

`re.lastIndex = { valueOf() { /* arbitrary JS */ return 0; } }; re.exec(subject)` is the trigger (the comment cites the exact test262 family proving this path runs user code). User JS in the window can run a moving collection today (back-edge safepoint polls in the callback, default-on since #7721, make the copying minor eligible); if the subject string is evacuated, `str_data` dangles into from-space for the entire match.

Also audit `set_last_index_throwing` (`:68`) in the same pass: if a user-defined `lastIndex` accessor can run there, the same hazard applies after that call. And check the sibling entry points that share this shape — `regex/match_string.rs:81`, `regex/match_all.rs:92`, `regex/replace_*` — for the same borrow-before-user-code ordering.

## Background (borrowed-heap-slice class)

Perry strings: `STRING_TAG` NaN-box → 20-byte `StringHeader`, WTF-8 payload inline (`crates/perry-runtime/src/string/mod.rs:307`), young-generation, relocated by the copying minor. Rooting rewrites *slots*, never an already-materialized `&str` (`HeapKeyBytes` doc, `crates/perry-runtime/src/object/field_get_set.rs:11-27`). `string_as_str<'a>` (`string/mod.rs:791`) has a caller-chosen lifetime.

## Suggested fix

[...]

Issue #8426, opened 2026-08-19
runtime: js_string_normalize holds a payload borrow across user toString — moving GC can relocate the subject mid-call


proggeramlug:
## Problem

`js_string_normalize` borrows the subject string's payload, then runs **user JavaScript** (the form argument's `toString`) while holding the borrow, then reads the borrow (`crates/perry-runtime/src/string/compare.rs:513-550`):

```rust
let str_data = string_as_str(s);                    // :520 — borrow of s's payload
...
crate::builtins::reject_symbol_to_string(form_value);
let form_ptr = crate::value::js_jsvalue_to_string(form_value);  // :532 — runs user toString
...
"NFC" => str_data.nfc().collect(),                  // :544-547 — reads the stale borrow
```

`"…".normalize({ toString() { /* arbitrary JS */ return "NFC"; } })` is the trigger. Unlike the allocation-point-only windows (#8423), user JS in the window **can run a moving collection today**: a loop inside the callback hits back-edge safepoint polls (default-on since #7721), where the copying minor is eligible. If it evacuates `s` (a young heap string), `str_data` points into from-space; the normalize pass then reads recycled or poisoned bytes.

## Background (borrowed-heap-slice class)

Perry strings are `STRING_TAG` NaN-boxes → 20-byte `StringHeader`, WTF-8 payload inline (`crates/perry-runtime/src/string/mod.rs:307`), young-generation, relocated by the copying minor. Rooting rewrites *slots*, never an already-materialized `&str` — the written rule is `HeapKeyBytes`'s doc (`crates/perry-runtime/src/object/field_get_set.rs:11-27`): stop borrowing, or re-derive after every collection point. `string_as_str<'a>` (`string/mod.rs:791`) has a caller-chosen lifetime, which is how this compiled.

## Suggested fix

Reorder — the minimal sound fix: coerce/validate the form argument **first** (lines 525-538 need nothing from `str_data`), and take `let str_data = string_as_str(s)` only after the last user-code/allocating call, immediately before the normalization match. Nothing else in the function allocates through the GC before the final `js_string_from_bytes` of the result (built from an owned Rust `String`, which is safe). Keep the Symbol-throws-TypeError-before-RangeError ordering (#2782 comment) — it is observable and the reorder must not change it (it doesn't: coercion still runs first).

Note both cfg arms (`string-normalize` feature on/off, `:540-554`) read `str_data`/form — fix covers both.

## Repro / validation

[...]

PR #8443, opened 2026-08-20
fix(runtime): root repeat's receiver across the count coercion

### LINKED ISSUE #8427: runtime: js_string_repeat holds a payload borrow across user valueOf — moving GC can relocate the subject mid-call
## Problem

`js_string_repeat` borrows the subject string's payload, then runs **user JavaScript** (the count argument's `valueOf`) while holding the borrow, then reads it (`crates/perry-runtime/src/string/pad.rs:296-317`):

```rust
let str_data = string_as_str(s);                                  // :302 — borrow of s's payload
let count_number = crate::builtins::js_number_coerce(count_value); // :303 — runs user valueOf
...
if count_integer == 0.0 || str_data.is_empty() {                  // :309 — reads borrow
...
let result = str_data.repeat(count);                              // :314 — reads borrow
```

`"abcdef".repeat({ valueOf() { /* arbitrary JS */ return 3; } })` is the trigger. User JS in the window can run a moving collection today: a loop in the callback hits back-edge safepoint polls (default-on, #7721) where the copying minor is eligible; if it evacuates `s` (young heap string), `str_data` dangles into from-space and `repeat` copies garbage (or faults under the quarantine instrument).

Perry strings: `STRING_TAG` NaN-box → 20-byte `StringHeader`, WTF-8 payload inline (`crates/perry-runtime/src/string/mod.rs:307`), moved by the copying minor. Rooting rewrites *slots*, not an already-materialized `&str` (`HeapKeyBytes` doc, `crates/perry-runtime/src/object/field_get_set.rs:11-27`). `string_as_str<'a>` (`string/mod.rs:791`) has a caller-chosen lifetime, which is how this compiled.

## Suggested fix

Reorder — coerce `count_value` (and run the range checks, `:303-307`) **before** taking `str_data`. Nothing after the coercion runs user code, and the final `js_string_from_bytes(result…)` copies from an owned Rust `String` (safe). The observable coercion-before-emptiness-check ordering per ECMA-262 §22.1.3.17 (ToIntegerOrInfinity runs even for the empty string, and a negative count throws before the empty check) is preserved by the reorder — add the fixture case `"".repeat({valueOf(){ throw-side-effect-order... }})` variants only if Node semantics are asserted by the byte-compare anyway.

Check the same file for the pad paths: `js_string_pad_*` take `str_data` and pad arguments too — if the pad-string/length coercions run user code after the subject borrow (see `finish_pad_result` callers around `:290`), fix them in the same PR with the same reorder; state in the PR whether they were affected.

## Repro / validation

- Fixture `test-files/test_issue_<this>_repeat_reentrant.ts`: build a young >5-byte subject dynamically (heap, not SSO/interned), call `.repeat({valueOf(){ let junk=""; for(let i=0;i<5000;i++) junk = junk + "x"; return 3; }})`, print result; byte-compare vs Node 26.5.1.
- Fault demonstration BEFORE the fix: `PERRY_GC_SCHEDULE_SEED=1 PERRY_GC_SCHEDULE_RATE=1 PERRY_GC_SCHEDULE_ALLOC_KB=0 PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=64 PERRY_GC_FORCE_EVACUATE=1` — expect a from-space fault naming `obj_type` 3; `PERRY_GC_DIAG=1` must show `[gc-fromspace-protect] retired_set=` (non-vacuous). Clean after the fix.
- Build note: `cargo build --profile perry-dev -p perry -p perry-runtime-static -p perry-stdlib-static` (stale-`.a` trap); perry-runtime tests `RUST_TEST_THREADS=1`.

Siblings, same class: #8423 (latent allocation-point windows), `js_string_normalize` (user `toString`), `js_regexp_exec` (lastIndex coercion) — filed separately.

## Workflow

proggeramlug:
Closes #8427.

## Problem

`js_string_repeat` borrowed the receiver's inline WTF-8 payload and *then*
coerced `count`:

```rust
let str_data = string_as_str(s);                                   // borrow of s's payload
let count_number = crate::builtins::js_number_coerce(count_value);  // runs user valueOf
...
let result = str_data.repeat(count);                                // reads the borrow
```

ToNumber on an object count runs user JS (`valueOf` / `Symbol.toPrimitive`).
A loop in that callback hits back-edge safepoint polls (default-on, #7721)
where the copying minor is eligible; when it evacuates a young receiver,
`str_data` names retired from-space and `repeat` copies garbage.

## Fix

Coerce first, borrow second — but that alone is not sufficient. `s` is a raw
pointer in a *native Rust frame*, which the collector does not scan by
default, so the later `string_as_str(s)` would still deref a stale address.
It is parked in a `RuntimeHandleScope` and the post-collection address is
taken back out via `across_const`, the sanctioned combinator for this shape
(no new bare raw-handle reads — `scripts/raw_handle_debt.py` stays at its
978 baseline).

The `count == 0` early return moves above the borrow, so the payload is read
only on the path that actually needs it. Observable ordering per ECMA-262
§22.1.3.17 is unchanged: ToIntegerOrInfinity still runs for an empty
receiver, and a negative count still throws before the empty-string return —
both are asserted in the fixture and byte-compared against Node 26.5.1.

## padStart / padEnd: audited, unaffected

The issue asked to check them. They are **not** affected, and the PR carries
evidence rather than a reading: codegen emits *both* of their coercions —
`js_number_coerce` for `maxLength` and `js_string_pad_fill` (ToString) for
the fill — **before** `reread_recv` produces the receiver handle
(`lower_string_method.rs:855-870`). No user code runs inside
`js_string_pad_start` / `_pad_end` after `string_as_str`; everything past it
(`to_length`, `decode_wtf8_units`, `build_pad_chunk`, `finish_pad_result`) is
pure Rust, and `finish_pad_result` copies the payload into an owned `Vec`
before the destination allocation. The fixture exercises both with reentrant
allocating `maxLength` *and* fill callbacks anyway, so a future move of
either coercion into the helper lands on a test already watching.

## Validation

`test-files/test_gap_gc_string_repeat_reentrant_count.ts` — a gap fixture, so
it runs in the per-PR tier, not only the nightly sweep. Built with
`cargo build --profile perry-dev -p perry -p perry-runtime-static
-p perry-stdlib-static` (`.a` mtimes verified to move after each edit), run
under the instruments in its `parity-env` line.

**The gate was proved able to go red.** Same fixture, same instruments, the
only difference being the runtime `.a`:

| | repeat (valueOf) | repeat (Symbol.toPrimitive) | padStart | padEnd | vs Node |
|---|---|---|---|---|---|
| pre-fix runtime | `false` | `false` | `true` | `true` | 2 lines differ |
| post-fix runtime | `true` | `true` | `true` | `true` | byte-identical |

Both arms report `copying_minors=530 moved_objects=16930` — the subject
collector ran, so the green verdict is not vacuous. The pad rows staying
green in *both* arms is the empirical half of the audit above (that code is
byte-identical across the two builds).

Without the instruments the fixture passes on both runtimes, which is the
point of the `parity-env` line: this class is invisible to an unpaced run.

Also green: `cargo fmt --all -- --check`, `scripts/check_file_size.sh`,
`scripts/addr_class_inventory.py` (the two `handle-floor` notes it prints are
pre-existing on `main`), `scripts/raw_handle_debt.py` (978, at baseline),
`scripts/gc_runtime_root_holders.py`, `scripts/check_gc_env_knobs.py`,
`scripts/check_node_version_consistency.py`.

Siblings in the same class, filed separately: #8426 (`js_string_normalize`),
#8428 (`js_regexp_exec` lastIndex).

### DIFF
--- a/changelog.d/8443-string-repeat-reentrant-count.md
+++ b/changelog.d/8443-string-repeat-reentrant-count.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- `String.prototype.repeat()` no longer reads a stale string payload when the `count` argument's `valueOf`/`Symbol.toPrimitive` triggers a moving garbage collection. The receiver is now rooted across the coercion and its payload borrowed only afterwards, so a relocated receiver repeats correct bytes instead of from-space garbage. `padStart`/`padEnd` were audited for the same window and are unaffected — codegen runs both of their coercions before the receiver handle is re-read.
--- a/crates/perry-runtime/src/string/pad.rs
+++ b/crates/perry-runtime/src/string/pad.rs
@@ -299,14 +299,40 @@ pub extern "C" fn js_string_repeat(s: *const StringHeader, count_value: f64) -> 
         return js_string_from_bytes("".as_ptr(), 0);
     }
 
-    let str_data = string_as_str(s);
-    let count_number = crate::builtins::js_number_coerce(count_value);
+    // `count` may be an object, so ToNumber runs its
+    // `valueOf`/`Symbol.toPrimitive` — arbitrary user JS, whose loop back-edge
+    // polls are moving-GC safepoints (default-on since #7721). Two distinct
+    // hazards follow, and the ordering below is what closes both (#8427):
+    //
+    //   * The receiver's WTF-8 payload must NOT be borrowed across the
+    //     coercion. `string_as_str` materializes a `&str` at the *pre-move*
+    //     address; rooting rewrites slots, never an already-live borrow, so no
+    //     root can repair one (`HeapKeyBytes`, `object/field_get_set.rs`).
+    //     Pre-fix, an evacuating minor inside `valueOf` left `str_data`
+    //     pointing at retired from-space and `repeat` copied garbage.
+    //   * `s` itself is a raw pointer in a native Rust frame, which the
+    //     collector does not scan by default — so deferring the borrow is not
+    //     enough on its own. Park it in a transient root and take the
+    //     *post-collection* address back out of the handle.
+    let scope = crate::gc::RuntimeHandleScope::new();
+    let receiver_root = scope.root_string_ptr(s);
+    let (count_number, s) = receiver_root
+        .across_const::<StringHeader, _>(|| crate::builtins::js_number_coerce(count_value));
+
     let count_integer = to_integer_or_infinity(count_number);
     if count_integer < 0.0 || count_integer.is_infinite() {
         throw_repeat_range_error(count_number);
     }
 
-    if count_integer == 0.0 || str_data.is_empty() {
+    // ECMA-262 §22.1.3.17 step 4 returns "" for n = 0 before the receiver is
+    // consulted at all; an empty receiver repeats to "" for every remaining n.
+    // Both are checked here so the payload is borrowed only on the path that
+    // actually reads it — and only once no user code is left to run.
+    if count_integer == 0.0 {
+        return js_string_from_bytes("".as_ptr(), 0);
+    }
+    let str_data = string_as_str(s);
+    if str_data.is_empty() {
         return js_string_from_bytes("".as_ptr(), 0);
     }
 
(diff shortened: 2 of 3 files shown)


proggeramlug:

reproducers.

PR #8444, opened 2026-08-20
fix(publish): remove scan-skip footgun, pin republish checkout, harden CI guardrails

jdalton:
## Summary

Follow-up to an end-to-end adversarial review of the npm staged-publish pipeline (stage -> verify -> Socket scan -> human 2FA approve -> tag/release). Nine agents reviewed each subsystem and four adversarially re-verified the highest-severity claims; this PR fixes what survived verification.

- **The Socket scan can no longer be skipped, and now runs in CI too.** `--no-scan` is removed from both `pipeline.mts` and `npm/approve.mts` -- the scan is mandatory. A missing/invalid `SOCKET_API_TOKEN` fails closed and is reported as `BLOCKED` (distinct from "scanned clean"), rather than silently proceeding unscanned. `npm-stage-publish.yml` now also scans every staged upload immediately in CI via `secrets.SOCKET_API_TOKEN`, instead of relying solely on a human later running `publish:approve` locally.
- **`pipeline.mts` no longer has a zero-flag footgun**, and no longer picks a mode by argument order when two conflict. Running it with no arguments (or a bare `npm run publish:pipeline`) used to silently dispatch a REAL staged publish; a mode flag is now required, exactly one, and unknown flags are rejected instead of silently ignored.
- **`release-packages.yml`'s `republish` mode now pins its checkout to the tag being republished.** Previously every job checked out whatever ref the dispatch happened to run on; if `main` had advanced past the tag, the rebuild could silently compile (and npm publish) a different version than the tag/release label says. `preflight` also cross-checks the tag's own `Cargo.toml` against the tag name before proceeding.
- **`npm-stage-publish.yml`'s `build-run-id` reuse is now verified** (workflow identity, success, exact commit) before its artifacts are trusted, and a fresh dispatch is now pinned to this run's exact commit via `--ref` and matched by `headSha` -- previously a stale/wrong-workflow run-id, or a dispatch that silently targeted the default branch's tip instead of this run's commit, could stage the wrong binaries under today's version with no error.
- **Both workflows now have a `concurrency:` group**, keyed on tag/dist-tag, so two dispatches for the same target can't race each other's idempotency checks. `cancel-in-progress: false` -- never cancels an in-flight publish.
- **The stage-upload step is now actually idempotent** -- it skips packages a prior partial run already staged, instead of re-attempting all 9 and reporting a 1-package problem as 9 failures.
- **`verifyAndScan`** (the `--stage-only`/`--scan-only` status path) now surfaces a partial staged set immediately instead of only at approve time.

Added regression tests in `publish.test.mts`: the `isMainModule` guard is asserted to be the last top-level statement in every entry-point module (the exact incident class that once fired a real `cargo publish` from a bare import), the no-mode and conflicting-mode usage errors (the latter spawns the real CLI), and that no `--no-scan`/`noScan` surface remains anywhere in the pipeline.

### CI now scans every staged upload, and two requirements remain for a maintainer

`npm-stage-publish.yml`'s `stage-upload` job now references `secrets.SOCKET_API_TOKEN` and runs a mandatory "Socket scan the staged tarballs" step right after staging (reusing `pipeline.mts`'s `--scan-only` path -- same verify+scan code the local flow uses, not a duplicate). Previously CI staged all 9 packages with zero scanning in between staging and whenever a human next ran `publish:approve`; now every staged upload is scanned immediately in CI and the workflow fails if it isn't clean. (`--scan-only` itself also had a bug fixed alongside this: it printed the scan status but never set a failing exit code, which would have made this new CI step unable to fail no matter what the scan found.)

Two things still need a maintainer with repo-admin access -- neither is something this PR can provision from outside:

1. **Provision the `SOCKET_API_TOKEN` GitHub Actions secret** (repo- or `npm-publish`-environment-scoped) that the new CI step now references. Perry is open source, so the org running these scans may qualify for Socket's free Business plan upgrade for OSS maintainers: https://socket.dev/blog/free-business-plan-upgrades-for-open-source -- that's the natural way to get a token provisioned if one doesn't already exist.
2. **Create the `npm-publish` GitHub environment.** `gh api repos/PerryTS/perry/environments` currently shows only one environment (`github-pages`); `npm-stage-publish.yml`'s `stage-upload` job references `environment: npm-publish`, required for npm's trusted-publisher OIDC config per that workflow's own comments. This environment does not exist on the repo yet. This is a prerequisite from [PerryTS/perry#8336](https://github.com/PerryTS/perry/pull/8336)'s original enablers list that still looks open, independent of anything in this PR.

Until both are provisioned, `stage-upload` will fail at the scan step (no token) and/or at the environment gate -- which is the correct fail-closed behavior, not a bug.

### Not in this PR (flagged, lower severity, left for a follow-up)

- `auth-posture.mts`'s long-lived-token detection is a fixed 3-name env-var allowlist; a token embedded directly in `.npmrc` is invisible to it.
- `pipeline.mts`'s concurrent-dispatch run-matching (client-side `dispatchedAt` + a fixed sleep) has a narrow race the code's own comment already acknowledges; the new workflow-level `concurrency:` groups reduce how often it can trigger but don't remove it.
- `cargo/ffi-publish.mts`'s perry-runtime prerequisite check is warn-only, relying on `cargo publish`'s own validation as the backstop.

## Test plan

- [x] `npm run test:scripts -- scripts/publish/publish.test.mts` -- 61/61 pass (including the two CodeRabbit-flagged fixes below)
- [x] Manually exercised `node scripts/publish/pipeline.mts` (no args) -> usage error, exit 1, no `gh` dispatch
- [x] Manually exercised `--stage-only --no-scan` -> rejected as an unknown flag
- [x] Manually exercised `--stage-only --approve` (both orders) -> rejected as conflicting mode flags, exit 1
- [x] `--status` still works end-to-end
- [x] Both workflow YAMLs parse (`python3 -c "yaml.safe_load(...)"`) and pass `actionlint` (no new findings -- only pre-existing shellcheck style notes)
- [x] Addressed CodeRabbit's two actionable findings: the fresh release-packages.yml dispatch now pins `--ref`/matches by `headSha`, and conflicting mode flags now fail closed instead of resolving by argument order
- [x] `--scan-only` now sets a failing exit code on an incomplete/blocked/not-passed scan (regression test added; this is what the new CI step relies on to fail)
- [ ] CI (`pr-gate`) -- pending

## Summary by CodeRabbit

* **New Features**
  * Added mandatory package security scanning with clear blocked-scan reporting.
  * Added idempotent staging that skips packages already uploaded.
  * Added serialized release and staging runs to prevent conflicting publishes.
  * Added validation for reused builds, release versions, workflow identity, and commit references.

* **Bug Fixes**
  * Prevented publishing without an explicit, valid mode.
  * Approval now stops when scans fail, are blocked, incomplete, or unauthenticated.

* **Documentation**
  * Documented publishing safeguards and validation requirements.


jdalton:
Addressed both actionable findings in 3147a9dce:

- `npm-stage-publish.yml`: the fresh `release-packages.yml` dispatch now pins `--ref` to this run's exact commit, and the run-matching poll now matches on `headSha` in addition to `createdAt >= dispatchedAt` (previously it could grab a concurrent dispatch of the same workflow from a different ref).
- `pipeline.mts`: two mode flags (e.g. `--stage-only --approve`) now fail closed with "Conflicting mode flags" instead of silently picking whichever came first in argv. Added a regression test that spawns the real CLI both ways (`--stage-only --approve` and `--approve --stage-only`) and asserts exit 1.

`npm run test:scripts -- scripts/publish/publish.test.mts` is 61/61 green.

jdalton:
Calling these out explicitly as requirements (moved from a loose note into the PR description's "Requirements" section too):

1. **`SOCKET_API_TOKEN`** -- a local env var for whoever runs `publish:stage`/`publish:scan`/`publish:approve` (never a CI secret; it's not referenced in any workflow file). Since Perry is open source, the org may qualify for a free Socket Business plan upgrade, which is the natural path to a token if one isn't already provisioned: https://socket.dev/blog/free-business-plan-upgrades-for-open-source
2. **The `npm-publish` GitHub environment** -- required by `npm-stage-publish.yml`'s `stage-upload` job for npm's trusted-publisher OIDC config, and it does not exist on the repo yet (`gh api repos/PerryTS/perry/environments` shows only `github-pages`).

Neither is something this PR can provision from outside -- both need a maintainer with repo-admin access.

jdalton:
Correction on requirement 1: `SOCKET_API_TOKEN` should be provisioned as a proper GitHub Actions secret (`secrets.SOCKET_API_TOKEN`, repo- or `npm-publish`-environment-scoped) -- not treated as local-shell-only. Updated the PR description to reflect that.

jdalton:
Implemented, not just requested: pushed 3382a8d3e.

`npm-stage-publish.yml`'s `stage-upload` job now references `secrets.SOCKET_API_TOKEN` and runs a new "Socket scan the staged tarballs" step immediately after staging (reusing `pipeline.mts --scan-only`, the same verify+scan code the local flow already uses -- no duplicated logic). Also fixed `--scan-only` itself: it printed the scan status but never set a failing exit code, which would have made the new CI step unable to fail regardless of what the scan found.

Once `SOCKET_API_TOKEN` and the `npm-publish` environment are provisioned (see the two remaining requirements in the PR description), this step gates every staged upload in CI, not just whichever local run a human remembers to do correctly.

proggeramlug:

reproducers.

PR #8445, opened 2026-08-20
tooling: ratchet StringHeader payload access

### LINKED ISSUE #8429: tooling: chokepoint API + ratchet lint for open-coded StringHeader byte access (729 inline '+20' sites across 22 crates)
## Problem

There is no chokepoint for "give me the bytes of a JS string." The exported ABI hands out `*const StringHeader` and every consumer computes `payload = header + size_of::<StringHeader>()` itself. Census (2026-08-19, regex-based, approximate):

- **729 inline `.add(size_of::<StringHeader>())`-style sites across 480 files in 22 crates**: perry-runtime 372, perry-stdlib 85, the eight UI crates 236 combined, ext-*/ffi/updater/container-compose 36.
- **~325 copy-pasted reader-helper definitions, ~1,400 call sites**: `str_from_header` ×248 definitions (UI crates; returns `&'static str`), `string_from_header` ×42 (stdlib 34), `read_str` ×22 (ext-* crates), plus ~12 misc variants.
- Inside perry-runtime: `string_as_str<'a>` (`string/mod.rs:791`) has a **caller-chosen lifetime unconnected to any root** — 124 call sites — plus duplicate copies at `regex.rs:617`, `static_plugins.rs:20`, `intl/number_format.rs:1128` (`string_header_as_str`).

Two costs. (1) **Soundness**: strings move (copying minor); a payload borrow held across an allocation is the "borrowed heap slice" bug class (#5062, #7341, `HeapKeyBytes` doc at `object/field_get_set.rs:11-27`). Several live windows were just found and filed as separate issues (`js_string_normalize`, `js_string_repeat`, `js_regexp_exec`, `js_string_to_well_formed`); with ~1,400 open-coded readers there is no way to audit the rest. (2) **Evolvability**: any change to the string layout (rope/cons, slice views — see #8394 discussion) is currently a 22-crate flag day.

## Suggested fix (two deliverables, one PR or two)

**1. Sanctioned accessors in perry-runtime** (the runtime already has all the pieces; this is about naming the blessed set and making it reachable):
- A copy-first helper generalizing `HeapKeyBytes` (`object/field_get_set.rs:28-71`) for short payloads.
- A rooted-reread pattern helper for long payloads: root in a `RuntimeHandleScope`, re-derive the pointer after every potential collection point (`string_copy_range` at `string/mod.rs:704` and `string/split.rs:479-520` — which records byte *ranges* and re-reads from the rooted handle per iteration — are the model citizens to document).
- Decide the fate of `string_as_str<'a>`: at minimum document it as "must not span a collection point"; ideally deprecate in favor of the above.

**2. A ratchet lint** modeled on the existing `scripts/addr_class_inventory.py` (already a `lint` step) — e.g. `scripts/string_payload_access_inventory.py`:
- Counts, per crate: inline `size_of::<StringHeader>()` pointer arithmetic, and definitions matching the copy-pasted reader shapes (`from_utf8_unchecked` + `StringHeader` in one function).
- Committed per-crate baseline; a count going UP fails `lint`; going down updates the baseline in the same PR.
- **The scanner must be able to fail**: include a `--self-test` that plants a synthetic offender and asserts detection (precedent: `scripts/gc_root_dominance_check.py --self-test`). Scanner regexes historically under-count and the failure direction is always green — the PR must report the population delta its regexes find vs. a manual spot-check on 2–3 crates.

Explicitly OUT of scope: migrating the 1,400 sites. That happens incrementally (per-crate dedupe issues are filed separately); this issue only stops the bleeding and makes progress measurable.

## Validation

- `python3 scripts/string_payload_access_inventory.py --self-test` passes; removing a planted offender makes it fail.
- Baseline counts match a manual grep on perry-runtime and one UI crate (state both numbers in the PR).
- `lint` stays green on an unmodified tree; goes red on a tree with one added inline `+size_of::<StringHeader>()` site.

## Workflow

PR = code + tests + `changelog.d/<PR>-<slug>.md`; no version bump (maintainer bumps at merge). New lint steps ride the existing `lint` job (already in `pr-gate`), so no branch-protection change is needed — but run the script once over the tree BEFORE wiring it into CI, and seed the baseline from that run.

proggeramlug:
## Summary

- promote the property-key inline/spill copy into public `OwnedStringBytes`, and re-export it from `perry-runtime`
- add `RuntimeHandle::with_string_bytes` as the rooted re-read chokepoint for long payload scans; document the collection boundary on `string_as_str`
- add a per-crate baseline ratchet for inline `StringHeader` offsets and unchecked reader-helper definitions, including planted-offender self-tests, and wire it into `lint`

## Census and spot checks

The committed scanner finds 727 inline offsets across 22 crates and 242 functions matching the deliberately narrow `StringHeader` + `from_utf8_unchecked` reader shape. That is -2 from the issue's approximate 729-site inline census. The 242 helper count is narrower than the issue's ~325 name-based helper census because the requested gate shape counts unchecked readers, not all named owned/safe readers.

Manual `rg` spot checks:

| crate | scanner | manual grep | note |
| --- | ---: | ---: | --- |
| `perry-runtime` | 370 | 375 raw / 370 code | five grep hits are doc-comment examples, which the scanner intentionally masks |
| `perry-stdlib` | 85 | 85 | exact match |
| `perry-ui-ios` | 44 | 44 | exact match |

## Validation

- `python3 scripts/string_payload_access_inventory.py --self-test`
- `python3 scripts/string_payload_access_inventory.py`
- planted one extra runtime `.add(size_of::<StringHeader>())` site: gate exited 1 with baseline 370, found 371
- `cargo test -p perry-runtime --lib string_bytes -- --nocapture`
- `SKIP_COMPILE_GATES=1 scripts/run_lint_gates.sh` (all 50 script/fmt gates passed)

No version bump.

Closes #8429


proggeramlug:
Validated and merging. Two notes, one about the gate itself.

## The new ratchet is non-vacuous — I checked

A PR that *adds* a gate deserves the question "can it fail?". Injecting a single bare
payload offset into `crates/perry-runtime/src/string/concat.rs`:

```
inline-offset | perry-runtime: baseline 370, found 371
```

The gate goes red. It catches what it claims to.

## Its baseline was already stale on arrival

`string_payload_access_inventory.py` failed on first run — the baseline expects payload
accesses across the `perry-ui-*` crates that no longer exist:

```
inline-offset | perry-ui-visionos: baseline 39, found 0
reader-helper | perry-ui-macos:    baseline 45, found 0
```

Cause: **#8453** ("replace borrowed runtime strings with owned copies") landed while this PR
was open and removed them. The baseline was captured before that merge, so it described a
tree that no longer exists — the same "a list that outlives its entries" failure the zlib
`KNOWN_EXT_GAPS` ratchet hit earlier today.

I ran `--write-baseline` and folded the refresh in, which locks in #8453's improvement
(23 stale entries removed) rather than leaving the gate red on `main`.

## Conflict

`crates/perry-runtime/src/lib.rs` conflicted on the re-export list for the third time in
this series: `main` has `perry_string_header_abi_revision` (#8442) and this adds
`OwnedStringBytes`. Resolved as a **union** — taking either side alone drops a symbol. Worth
noting that a coordinated series landing piecemeal produces this repeatedly; a single
combined PR would not have.

## Batch validation

19/19 corpus byte-exact · `perry-runtime --lib` 2604 · `perry --bin perry` 1008 ·
`perry-codegen --lib` 1113 · all 52 gates.

proggeramlug:
Superseded by #8481 — the same work rebased onto current `main`.

GitHub could not create the merge commit (`lib.rs` re-export list conflicts with #8442's `perry_string_header_abi_revision`), and this is a fork branch I cannot push the resolution to. Resolved as a union so neither symbol is dropped.

Two things carried over to #8481:
- **Your ratchet is non-vacuous** — injecting a bare payload offset into `concat.rs` moves `perry-runtime` 370 → 371 and the gate goes red.
- **Its baseline was stale on arrival**: #8453 landed while this was open and removed the `perry-ui-*` accesses the baseline still expected (`perry-ui-visionos: baseline 39, found 0`). Refreshed, dropping 23 stale entries.

PR #8446, opened 2026-08-20
fix(runtime): coerce lastIndex before borrowing the regex subject

### LINKED ISSUE #8428: runtime: js_regexp_exec holds the subject payload borrow across the lastIndex Get/ToLength user side effects

proggeramlug:
Closes #8428.

## Problem

`RegExpBuiltinExec` step 4 is `ToLength(Get(R, "lastIndex"))`. When `lastIndex`
holds an object, the ToNumber half runs `OrdinaryToPrimitive` — user
`valueOf`/`toString`, i.e. arbitrary JS, which reaches back-edge safepoint polls
and can therefore run a **moving** minor today.

`js_regexp_exec` took the subject's inline-payload borrow (`string_as_str`,
`regex/exec.rs:36`) *before* that coercion (`:51`). Rooting rewrites *slots*,
never an already-materialized `&str` (the `HeapKeyBytes` doc states the rule), so
a collection inside the callback left the entire match running over from-space
bytes — the match silently evaporates, or returns text from whatever was
recycled into those pages.

`String.prototype.matchAll` had the same defect through Rust's left-to-right
argument evaluation: `materialize_match_all_results(s, re, regex_last_index_offset(re))`
evaluates `s` and `re` *before* the coercion runs, so the snapshot phase rooted
and read a pre-move subject.

## Fix

* `regex/exec.rs` — root `re` and `s`, run the `lastIndex` coercion, take the
  borrow from the **refreshed** addresses. This is also strictly closer to spec:
  step 4 precedes reading the flags (steps 5–7) and `[[RegExpMatcher]]`
  (step 12), which the old ordering did in the other order.
* `regex/exec.rs` — the fancy-regex arm now writes `lastIndex` before the result
  array is allocated, matching both the spec (step 15 precedes step 16's
  ArrayCreate) and the standard arm; that also keeps `re` out of the window the
  capture allocations open.
* `regex/exec.rs` — the match array's `.input` decoration re-boxes the subject,
  so it now reads the current address rather than the one bound before the
  capture/groups allocations.
* `regex/match_all.rs` — both `matchAll` entry points root the subject and the
  regex header across the coercion and hand over refreshed pointers.

## Audit (issue asked for these explicitly)

* **`set_last_index_throwing` does not reopen the window.** It reads the property
  attributes (a side-table read) and stores a number; Perry has no user-defined
  `lastIndex` setter on this path. Its only allocation is on the non-writable arm,
  which throws and therefore never returns to the borrow. Noted in the code.
* **`regex/match_string.rs`** (`js_string_match`) never reads `lastIndex`, so it
  has no user-JS window ahead of its borrow; `js_string_match_value` already
  roots the subject across the pattern coercion.
* **`regex/replace_fn.rs` / `regex/replace_expand.rs`** — the user-callback paths
  already re-derive the subject per use (`cur_str = || string_as_str(s_handle…)`);
  the string/named paths run pure Rust over the borrow with a single terminal
  allocation.

**Not in scope, filed separately:** `exec.rs`/`exec_array.rs` still read
`str_data` (and `regex::Captures` borrowing it) after the result-array,
capture-string and groups-object allocations. That is the *allocation-point*
window — the same latent class as #8423 — which is closed today only by the
#7682 guard forcing a conservative scan (hence a non-moving minor) at alloc
points. Closing it properly means the owned-snapshot restructure `match_all.rs`
already carries, which is a much larger change to a hot path and deserves its own
measurement. Filed as #8449.

## Validation

* **New GC witness** `gc::tests::runtime_roots::regexp_last_index` — plants a
  `valueOf` that runs a copying minor and then refills the retired Eden with a
  distinctive pattern. It asserts its subject was live *both* ways (a copying
  minor ran; the subject address actually changed) before checking the captures.
  **Sabotage-checked**: reverting only `exec.rs`/`match_all.rs` to `main` makes it
  fail (`/(young)-(\d+)/g must match the relocated subject` — the match evaporates),
  and it passes with the fix.
* **New parity fixture** `test-files/test_issue_8428_exec_lastindex_reentrant.ts`
  — std-regex arm, fancy-regex (lookbehind) arm, a non-zero coerced `lastIndex`
  (which additionally walks the borrow in `utf16_index_to_byte`), named captures
  with the `d` flag, `matchAll`, and `test()`. Runs under
  `PERRY_GC_SCHEDULE_SEED`/`RATE=1` + `FORCE_EVACUATE` + `VERIFY_EVACUATION`;
  byte-compared against Node 26.5.1.
* `scripts/raw_handle_debt.py` stays at baseline (978) — the new code uses
  `across_*` / `with_*_ptr` throughout and adds no bare handle reads.

### DIFF
--- a/changelog.d/8446-regexp-lastindex-rooting.md
+++ b/changelog.d/8446-regexp-lastindex-rooting.md
@@ -1,0 +1,10 @@
+Fixed a moving-GC hazard in `RegExp.prototype.exec` and `String.prototype.matchAll`.
+RegExpBuiltinExec coerces `lastIndex` with `ToLength`, which runs user
+`valueOf`/`toString` when `lastIndex` holds an object — arbitrary JS that can
+run a copying minor. Both entry points captured the subject string (and, in
+`exec`, borrowed its inline payload) *before* that coercion, so a collection
+inside the callback left the match running over relocated-away bytes: the match
+silently evaporated or returned text from unrelated heap memory. The coercion
+now runs first with the regex header and subject rooted, and every later use
+reads the refreshed addresses — including the match array's `.input` property
+and the `lastIndex` write-back on the lookbehind/backreference path.
--- a/crates/perry-runtime/src/gc/tests/runtime_roots.rs
+++ b/crates/perry-runtime/src/gc/tests/runtime_roots.rs
@@ -13,6 +13,7 @@ mod json_shape_template;
 mod native_module_name;
 mod old_defrag_contract;
 mod prototype_addr_cache;
+mod regexp_last_index;
 mod side_table_scanners;
 mod string_slice;
 mod symbol_description;
--- a/crates/perry-runtime/src/gc/tests/runtime_roots/regexp_last_index.rs
+++ b/crates/perry-runtime/src/gc/tests/runtime_roots/regexp_last_index.rs
@@ -1,0 +1,128 @@
+//! Moving-GC regression for `RegExp.prototype.exec` (#8428).
+//!
+//! RegExpBuiltinExec step 4 is `ToLength(Get(R, "lastIndex"))`. When
+//! `lastIndex` holds an object the ToNumber half runs OrdinaryToPrimitive —
+//! user `valueOf`/`toString`, i.e. arbitrary JS, which reaches safepoints and
+//! can run a copying minor. `js_regexp_exec` borrowed the subject's inline
+//! WTF-8 payload (`string_as_str`) BEFORE that coercion, and rooting rewrites
+//! slots, never an already materialized `&str` — so the whole match ran over
+//! from-space bytes.
+//!
+//! The planted `valueOf` runs a copying minor and then refills the just
+//! retired Eden with a distinctive pattern, so a pre-fix run matches against
+//! the stomped bytes instead of the relocated subject and the match evaporates.
+//! Liveness is asserted both ways (a copying minor ran; the subject actually
+//! moved), per CLAUDE.md's "a gate must assert its subject was live".
+
+use super::super::super::*;
+use super::super::support::*;
+
+use crate::array::ArrayHeader;
+use crate::object::ObjectHeader;
+use crate::regex::RegExpHeader;
+use crate::string::StringHeader;
+
+/// Stand-in for `{ valueOf() { …allocating user JS…; return 0 } }`.
+extern "C" fn collect_then_zero(_closure: *const crate::closure::ClosureHeader) -> f64 {
+    crate::gc::gc_collect_minor();
+    // Recycle the just-retired from-space blocks back out as fresh strings, so
+    // a stale borrow reads THESE bytes rather than the subject's old ones.
+    let filler = [b'#'; 64];
+    for _ in 0..64 {
+        let _ = crate::string::js_string_from_bytes(filler.as_ptr(), filler.len() as u32);
+    }
+    0.0
+}
+
+fn heap_string(bytes: &[u8]) -> *mut StringHeader {
+    crate::string::js_string_from_bytes(bytes.as_ptr(), bytes.len() as u32)
+}
+
+fn capture_text(element: f64, index: usize) -> String {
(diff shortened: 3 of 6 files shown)


proggeramlug:

reproducers.

PR #8448, opened 2026-08-20
refactor(stdlib): centralize string header readers

### LINKED ISSUE #8431: stdlib/ext: dedupe ~64 copy-pasted string readers (string_from_header/read_str) onto perry-ffi read_string/read_bytes
## Problem

perry-stdlib and the ext-* crates carry **~64 copy-pasted string-reader definitions with ~600 call sites**, each open-coding the `StringHeader` layout:

- `string_from_header`: 42 definitions / 401 call sites (perry-stdlib alone: 34 defs / 331 calls — e.g. `crates/perry-stdlib/src/axios.rs:10-18`; also ext-fastify, ext-events, runtime-internal stragglers).
- `read_str`: 22 definitions / 194 call sites across ext-fetch (43 calls), ext-validator (19), ext-ioredis (19), ext-mongodb (15), ext-commander (12), ext-http (10), ext-cheerio (9), and a dozen more — e.g. `crates/perry-stdlib/src/uuid.rs:41-49`.
- Misc variants (`string_header_to_string`, `read_string_header*`, …): ~12 defs / ~30 calls.

`perry-ffi` already provides the canonical readers — `read_string` (`crates/perry-ffi/src/lib.rs:212`) and `read_bytes` (`lib.rs:225`) — with 75+7 existing ext-* call sites proving they work for this exact purpose. The copies exist by inertia, and every one is a separate place a `StringHeader` layout change silently breaks (plus each is a fresh opportunity for the borrowed-heap-slice bug class: a payload borrow held across an allocation dangles when the copying GC moves the string).

## Suggested fix

1. Migrate `read_str` / `string_from_header` / misc definitions onto `perry_ffi::read_string` / `read_bytes` (crates that already depend on perry-ffi) or onto a shared runtime helper for perry-stdlib (which depends on perry-runtime directly — use/extend the sanctioned accessor from the byte-access-chokepoint issue if it has landed, else `perry_runtime`'s existing `str_bytes_from_jsvalue` + local copy-to-owned).
2. Where the local copy returns an owned `String` already (most do — `String::from_utf8_lossy(...).into_owned()`), keep that shape: owned returns are the safe pattern. Do NOT convert owned-return sites to borrows.
3. Land crate-by-crate; ext-* crates are independent, so this parallelizes across workers. perry-stdlib's 34 definitions are the biggest single chunk and can be one PR.
4. Coordinate with the `STRING_HEADER_ABI_REVISION` tripwire issue (filed separately): once readers are centralized in perry-ffi, that tripwire covers all of them.

## Validation

- `cargo check` for each touched crate; the affected package smokes / node-suite tests for ext-* crates that have them.
- Behavior must be bit-identical: these are mechanical replacements. Where a local copy used `from_utf8_unchecked` and perry-ffi's `read_string` validates UTF-8 (returns `None` on invalid), check whether the call site can receive WTF-8 (lone surrogates) or binary-in-string payloads — those sites need `read_bytes` + lossless handling, not `read_string`. This is the one place the migration is NOT mechanical; flag any such site in the PR.
- Build note for end-to-end checks: `cargo build --profile perry-dev -p perry -p perry-runtime-static -p perry-stdlib-static` (building only the rlib crates leaves a stale `.a`).

## Workflow

proggeramlug:
Closes #8431

## Summary

- centralize perry-stdlib StringHeader reads behind owned strict, lossy, and raw-byte accessors built on the runtime string byte chokepoint
- replace all 34 stdlib reader definitions while preserving each call site's prior UTF-8 behavior
- move the remaining Fastify, HTTP server, and net readers onto perry-ffi read_string/read_bytes
- retain owned returns so no GC-managed payload borrow escapes into allocating or async code

No version bump.

## Validation

- `cargo check -p perry-stdlib -p perry-ext-fastify -p perry-ext-http -p perry-ext-net`
- `cargo check -p perry-stdlib --tests`
- `cargo test -p perry-stdlib string_header_tests --lib` (2 passed)
- Fastify reader/body tests passed; two unrelated GC relocation assertions failed in the full Fastify lib run
- `cargo fmt --all`
- `git diff --check`

### DIFF
--- a/changelog.d/8448-centralize-string-readers.md
+++ b/changelog.d/8448-centralize-string-readers.md
@@ -1,0 +1,6 @@
+### Changed
+
+- Centralized native stdlib and extension string-header reads behind the
+  runtime and `perry-ffi` accessors, preserving strict UTF-8, lossy UTF-8, and
+  raw-byte behavior while keeping results owned across allocations and async
+  boundaries.
--- a/crates/perry-ext-fastify/src/context.rs
+++ b/crates/perry-ext-fastify/src/context.rs
@@ -8,7 +8,8 @@ use std::sync::atomic::{AtomicU64, Ordering};
 
 use perry_ffi::{
     alloc_string, build_object_shape, get_handle, get_handle_mut, js_object_alloc_with_shape,
-    js_object_set_field, ArrayHeader, Handle, JsValue, ObjectHeader, StringHeader,
+    js_object_set_field, read_bytes, ArrayHeader, Handle, JsString, JsValue, ObjectHeader,
+    StringHeader,
 };
 
 const TAG_UNDEFINED: u64 = 0x7FFC_0000_0000_0001;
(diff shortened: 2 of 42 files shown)


proggeramlug:

reproducers.

PR #8450, opened 2026-08-20
perf(codegen): fuse String.concat arguments into concat chain

### LINKED ISSUE #8433: perf(codegen): str.concat(a, b, …) copies the accumulator once per argument — fold into one js_string_concat_chain call
## Problem

`str.concat(a, b, c, …)` lowers as **one `js_string_concat` call per argument**, threading a rooted accumulator (`crates/perry-codegen/src/lower_string_method.rs:1097-1146`, the `"concat"` arm using `with_rooted_accumulator` / `RootedAcc::advance`). Each `js_string_concat` allocates a fresh string and copies the entire accumulator plus one argument — so `s.concat(a1..ak)` copies the prefix k times: O(total² ) in bytes for many/large args.

Meanwhile the expression form `a + b + c + d` already folds into **one** `js_string_concat_chain` call (single sizing pass, single allocation, N memcpys — `crates/perry-runtime/src/string/concat.rs:478`, codegen fold in `crates/perry-codegen/src/expr/binary.rs:732-738` via `flatten_string_add_chain`, cap `CONCAT_CHAIN_MAX_PARTS = 32`). The `.concat()` method predates the chain machinery and never adopted it.

The rooting in the current arm is CORRECT (its comment block documents why, #6971/#7615) — this is purely a copy-count problem, not a soundness bug.

## Suggested fix

Rebuild the `"concat"` arm on the chain call:

1. Lower + ToString-coerce each argument FIRST, storing each resulting NaN-boxed string into a parts array (function-entry `[N x double]` alloca, exactly like `lower_string_concat_chain` in `crates/perry-codegen/src/lower_string_concat.rs:598` — reuse that helper if its shape fits, or extract a shared core). Coercion order must stay left-to-right (ECMA-262 §22.1.3.5), and each coercion can run user `toString` (which can allocate/collect), so:
   - parts already stored in the alloca must survive collections — check how the existing chain lowering roots its parts array across part lowering (it does; follow its pattern exactly), and
   - the receiver must be re-read from its root after each coercion.
2. Emit one `js_string_concat_chain(parts_ptr, n_parts)` for receiver + args.
3. Arity beyond the 32-part cap: fall back to chunking (chain the first 31 + recurse) or to the current per-arg loop for the tail — either is fine; state which and why in the PR.
4. Runtime note: `js_string_concat_chain` formats *numeric* parts itself, but arbitrary objects must be coerced before the call (step 1 does this — pass only string-tagged/SSO parts). Verify `js_string_concat_chain` accepts SSO-tagged parts (the expression fold already produces them, so it should).

## Measurement

Fixture: `parts.reduce`-free direct call `s.concat(p1,…,p8)` in a loop with growing `s`, plus a many-args case. Compare instructions/time before/after (and against the equivalent `+` expression, which should now match). Include the numbers in the PR.

## Validation

- Correctness fixtures in `test-files/` byte-compared vs Node 26.5.1: mixed-type args (`undefined`, booleans, `{toString(){…}}` objects — the arm's comment warns these must render as string forms), a `toString` that itself concatenates (re-entrancy), astral/lone-surrogate args (WTF-8 boundary pairs — `canonicalize_surrogate_pairs` handles the joins in the chain helper; verify with an emoji split across parts), and >32 args.
- IR regression test (in `crates/perry-codegen`, near the existing chain tests) asserting a 4-arg `.concat` emits exactly one `js_string_concat_chain` and zero `js_string_concat`.
- GC instrument pass: `PERRY_GC_SCHEDULE_SEED=1 PERRY_GC_SCHEDULE_RATE=1 PERRY_GC_PROTECT_FROMSPACE=1` over the re-entrant-toString fixture (this is the arm most likely to break rooting — the fixture must allocate inside `toString`).
- Build note: `cargo build --profile perry-dev -p perry -p perry-runtime-static -p perry-stdlib-static`.

## Workflow

proggeramlug:
## Summary

- lower all `String.prototype.concat` arguments first, then preserve left-to-right `ToString` coercion with mutable-root re-reads
- emit one `js_string_concat_chain` for the receiver plus up to 31 arguments
- keep the existing rooted pairwise accumulator only for arguments beyond the runtime's 32-part cap
- share the existing entry-block chain scratch-buffer emitter with string `+` lowering

This also preserves JavaScript call ordering: all argument expressions are evaluated before any method-side coercion (`eval-A, eval-B, coerce-A, coerce-B`).

## Measurement

Fixture: an 8-argument direct `s.concat(p, …)` growing accumulator for 6,000 iterations (`p` is 8 bytes), plus a 35-argument case for 1,400 iterations.

- emitted helper calls per 8-argument concat: **8 `js_string_concat` → 1 `js_string_concat_chain`**
- emitted helper calls per 35-argument concat: **35 `js_string_concat` → 1 chain + 4 pairwise tail calls**
- post-fix wall time, median of 3 runs on this macOS arm64 host:
  - 8-argument method: **0.18 s**
  - equivalent 8-part `+` expression: **4.67 s**
  - 35-argument method: **0.18 s**

The chain path removes seven intermediate allocations and seven repeated prefix copies in the common 8-argument case.

## Validation

- `cargo test -p perry-codegen string_concat_method_fuses_four_arguments_into_one_chain_call -- --nocapture`
- Node 26.5.1 byte parity for mixed primitives, observable evaluation/coercion order, re-entrant allocating `toString`, split-surrogate canonicalization, and 35 arguments
- forced moving-GC run with `PERRY_GC_SCHEDULE_SEED=1 PERRY_GC_SCHEDULE_RATE=1 PERRY_GC_PROTECT_FROMSPACE=1`: output matched Node; 8/8 scheduled copying collections, 16,699 moved objects
- `cargo fmt --check`

Closes #8433

## Summary by CodeRabbit

  * Improved `String.prototype.concat` performance by combining multiple arguments into a single allocation.
  * Supports efficient concatenation chains of up to 31 arguments.

* **Bug Fixes**
  * Preserved left-to-right evaluation and string coercion behavior.
  * Improved reliability during re-entrant operations and garbage collection.
  * Corrected surrogate-pair handling and concatenation beyond the optimized argument limit.

* **Tests**
  * Added coverage for evaluation order, coercion, memory retention, Unicode handling, and long concatenation chains.

### DIFF
--- a/changelog.d/8450-string-concat-chain.md
+++ b/changelog.d/8450-string-concat-chain.md
@@ -1,0 +1,13 @@
+### Performance
+
+- Fuse `String.prototype.concat` calls of up to 31 arguments into one
+  `js_string_concat_chain` allocation instead of allocating and copying the
+  growing prefix once per argument. An 8-argument call now emits one chain
+  helper instead of eight pairwise helpers; the 6,000-iteration growing-string
+  fixture measured 0.18 s median on macOS arm64.
+
+### Fixed
+
+- Preserve JavaScript's argument-evaluation-before-coercion order for
+  `String.prototype.concat`, and keep the receiver, raw arguments, and coerced
+  strings rooted across re-entrant user `toString` calls.
--- a/crates/perry-codegen/src/codegen/declared_string_add_tests.rs
+++ b/crates/perry-codegen/src/codegen/declared_string_add_tests.rs
@@ -367,6 +367,37 @@ fn a_self_append_chain_retains_the_accumulator_and_fuses_only_the_suffix() {
 }
 
 #[test]
+fn string_concat_method_fuses_four_arguments_into_one_chain_call() {
+    let ir = ir(
+        Vec::new(),
+        Expr::Call {
+            callee: Box::new(Expr::PropertyGet {
+                object: Box::new(Expr::String("head".to_string())),
+                property: "concat".to_string(),
+                byte_offset: 0,
+            }),
+            args: ["a", "b", "c", "d"]
+                .into_iter()
+                .map(|s| Expr::String(s.to_string()))
+                .collect(),
+            type_args: Vec::new(),
+            byte_offset: 0,
+        },
+    );
+
+    assert_eq!(
+        ir.matches("call i64 @js_string_concat_chain(").count(),
+        1,
+        "a four-argument String.concat should allocate one chain result:\n{ir}"
+    );
+    assert_eq!(
+        ir.matches("call i64 @js_string_concat(").count(),
+        0,
+        "the fused String.concat must not retain pairwise concat calls:\n{ir}"
+    );
+}
+
+#[test]
 fn a_self_append_chain_keeps_an_opaque_numeric_head_pair_intact() {
     // `s = s + n + "x"` cannot split after `s`: when a lying string slot and
     // `n` both contain numbers, the head pair is numeric addition before the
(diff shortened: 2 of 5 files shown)


proggeramlug:

reproducers.

Issue #2782, opened 2026-05-29
runtime: make String.prototype.normalize throw RangeError for invalid forms


andrewtdiz:
## Summary

Perry implements Unicode normalization for the valid `String.prototype.normalize()` forms, but invalid `form` values silently fall back to NFC. Node throws a `RangeError` when the supplied form is anything other than `"NFC"`, `"NFD"`, `"NFKC"`, or `"NFKD"`.

## Node behavior

Local Node v25.9.0 probe:

```console
undefined="é":e9
NFC="é":e9
NFD="é":65+301
NFKC="é":e9
NFKD="é":65+301
BAD=RangeError:The normalization form should be one of NFC, NFD, NFKC, NFKD.
null=RangeError:The normalization form should be one of NFC, NFD, NFKC, NFKD.
=RangeError:The normalization form should be one of NFC, NFD, NFKC, NFKD.
```

Probe shape:

```js
"e\u0301".normalize()
"e\u0301".normalize("NFC")
"e\u0301".normalize("NFD")
"e\u0301".normalize("NFKC")
"e\u0301".normalize("NFKD")
"e\u0301".normalize("BAD")
"e\u0301".normalize(null)
"e\u0301".normalize("")
```

The no-argument call defaults to NFC. Supplying `null` is not the same as omitting the argument: Node applies `ToString(null)`, gets `"null"`, and throws because it is not a valid normalization form.

## Perry evidence

- `crates/perry-codegen/src/lower_string_method.rs` lowers `normalize` with zero or one argument and passes the provided argument through to `js_string_normalize`.
- `crates/perry-runtime/src/string/compare.rs::js_string_normalize()` handles the four valid forms, but its wildcard match arm returns `str_data.nfc().collect()` for everything else.
- The runtime comment currently says `Pass null/empty for default NFC`, which contradicts Node behavior for explicit `null` and empty string arguments.
- `test-files/test_gap_string_methods.ts` only covers the valid forms and the no-argument NFC default, so the invalid-form error path is currently untested.

## Duplicate check

Searched issues and PRs for:

- `String.normalize`
- `normalize RangeError`
- `String.prototype.normalize invalid`
- PR search for `String.normalize OR String.prototype.normalize OR normalize RangeError`

No existing focused issue showed up. The broad docs mention `normalize()` as a parity gap, but there is no open tracker for this concrete invalid-form behavior.

## Expected fix shape

[...]

PR #8451, opened 2026-08-20
fix(runtime): root the normalize subject across form coercion

### LINKED ISSUE #8426: runtime: js_string_normalize holds a payload borrow across user toString — moving GC can relocate the subject mid-call

proggeramlug:
Fixes #8426.

## Problem

`js_string_normalize` borrowed the subject string's inline WTF-8 payload, then coerced its `form` argument, then read the borrow:

```rust
let str_data = string_as_str(s);            // borrow of s's payload
...
crate::builtins::reject_symbol_to_string(form_value);
let form_ptr = crate::value::js_jsvalue_to_string(form_value);   // runs user toString
...
"NFC" => str_data.nfc().collect(),          // reads the stale borrow
```

The coercion is a collection point **twice over**:

- an object form runs user `toString`, whose loop back-edge polls (default-on since #7721) run a moving minor — this is reachable from ordinary user code today;
- an inline short-string form materializes onto the heap (`js_string_materialize_to_heap`), so even a plain `s.normalize("NFC")` allocates inside the window. That arm is latent — an alloc-point collection forces a conservative stack scan, which makes the copying minor ineligible — but it is the same window.

Either can evacuate a young subject. Rooting rewrites *slots*, never an already-materialized `&str`, so the normalization pass then reads retired from-space.

## Fix

Coerce the form first, root the subject across the coercion in a `RuntimeHandleScope`, and take the borrow only from the address `across_const` hands back. Nothing in the coercion needs `str_data`, and nothing after it allocates through the GC before the result is built from an owned Rust `String`.

Both `cfg` arms (`string-normalize` on/off) read the re-derived borrow. The observable orderings are unchanged — ToString still runs before the form is validated, so a Symbol form throws `TypeError` rather than the invalid-form `RangeError` (#2782). The existing `test_gap_2786_2880_2782_2789_string_semantics.ts` still passes, and the new fixture pins both orderings explicitly.

## Validation

**Fault demonstrated before the fix**, with `test-files/test_issue_8426_normalize_reentrant.ts` under the issue's knobs (`PERRY_GC_SCHEDULE_SEED=1 PERRY_GC_SCHEDULE_RATE=1 PERRY_GC_SCHEDULE_ALLOC_KB=0 PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=64 PERRY_GC_FORCE_EVACUATE=1`):

```
[gc-fromspace-protect] FAULT: signal 10 at 0x20887000000
  block=0x20887000000 +0 retired_bytes=80 retired_by_minor=#4950
  last-known object: user_ptr=0x20887000008 obj_type=3 size=40
```

`obj_type=3` is `GC_TYPE_STRING` — the subject. After the fix the same binary/knobs run clean (115,025 copying minors, 27,107 objects moved, exit 0) and stdout is byte-identical to Node 26.5.1 in both default and stress modes.

One note for anyone reproducing: **the subject's construction matters.** A `+=` accumulator chain leaves its buffer outside the movable nursery, so a subject built that way never relocates and the fixture passes with or without the bug. The fixture uses `join("")`; a measurement pass confirmed `join`/`slice`/`repeat`/`toLowerCase`/`substring` subjects all evacuate mid-window while the `+=` one does not. The fixture carries a comment saying so.

## Gate

The `.ts` fixture only runs in the full tier and only faults under stress knobs, so it is a regression fixture, not a gate. The gate is a `cargo-test`-visible unit test, `gc::tests::runtime_roots::string_normalize_form` — it builds a form object whose `toString` forces a real copying minor and asserts three things together, so it cannot pass vacuously:

1. the subject is nursery-resident (premise),
2. the collection actually **moved** it (the window was live),
3. the normalized bytes are the subject's, not the retired page's.

It runs under `ProtectionModeGuard::PoisonOnly` so a stale read is *guaranteed* to be detected rather than left to whatever the allocator recycled into the page.

**Sabotage-checked:** with the fix reverted and the test kept, it fails — SIGSEGV under poison, and a clean `left: 16, right: 15` assertion without it. A green run means the detector works.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `String.prototype.normalize` reliability when form coercion triggers memory cleanup.
  * Preserved expected behavior for default forms, invalid forms, and Symbol values.
  * Added regression coverage for repeated calls, normalization variants, and reentrant coercion scenarios.

### DIFF
--- a/changelog.d/8451-normalize-form-coercion-rooting.md
+++ b/changelog.d/8451-normalize-form-coercion-rooting.md
@@ -1,0 +1,13 @@
+### Fixed
+
+- `String.prototype.normalize` no longer holds a borrowed heap-string payload
+  across its `form` argument's `ToString` coercion. The coercion is a
+  collection point twice over — an object form runs user `toString` (whose
+  loop back-edge polls run a moving minor), and an inline short-string form
+  materializes onto the heap — so a young subject could be evacuated while
+  `js_string_normalize` held a `&str` into its pre-move address, after which
+  the normalization pass read retired from-space. The form is now coerced
+  first, the subject is rooted across the coercion, and the payload is
+  borrowed only from the post-collection address. The observable orderings are
+  unchanged: `ToString` still runs before the form is validated, so a Symbol
+  form throws `TypeError` rather than the invalid-form `RangeError`. (#8426)
--- a/crates/perry-runtime/src/gc/tests/runtime_roots.rs
+++ b/crates/perry-runtime/src/gc/tests/runtime_roots.rs
@@ -15,6 +15,7 @@ mod old_defrag_contract;
 mod prototype_addr_cache;
 mod regexp_last_index;
 mod side_table_scanners;
+mod string_normalize_form;
 mod string_slice;
 mod symbol_description;
 mod transient_handles;
--- a/crates/perry-runtime/src/gc/tests/runtime_roots/string_normalize_form.rs
+++ b/crates/perry-runtime/src/gc/tests/runtime_roots/string_normalize_form.rs
@@ -1,0 +1,128 @@
+use super::*;
+use crate::arena::FromSpaceProtection;
+
+/// #8426: `String.prototype.normalize` borrowed the subject string's inline
+/// WTF-8 payload BEFORE coercing its `form` argument, then read that borrow
+/// after the coercion had returned.
+///
+/// The coercion is a collection point twice over: an inline short-string form
+/// materializes onto the heap (so even `s.normalize("NFC")` allocates there),
+/// and an object form runs user `toString`, whose loop back-edge polls run a
+/// moving minor. Either can evacuate a young subject, and a `&str` taken
+/// beforehand is a copy the collector cannot rewrite — rooting rewrites slots,
+/// never already-materialized borrows. The normalization pass then read
+/// retired from-space.
+///
+/// The test drives the object-form window, which is the one reachable from
+/// user code today: `toString` forces a real copying minor. Three assertions
+/// together, because any one alone can pass vacuously — (1) the subject was
+/// young, (2) the collection actually MOVED it, so the window was live, and
+/// (3) the normalized bytes are the subject's, not the retired page's.
+///
+/// `PoisonOnly` makes failure certain rather than lucky. Without it, whether a
+/// stale borrow is *detected* depends on what the allocator happened to
+/// recycle into the retired page; with it, those bytes are guaranteed poison.
+/// The cost is the failure mode: a regression faults inside the normalization
+/// pass (poison is not valid WTF-8) rather than reaching the byte assertions
+/// below, so a reintroduced #8426 shows up as a SIGSEGV naming this test.
+/// That is deliberate — a gate that can pass vacuously is not a gate.
+#[test]
+fn normalize_form_coercion_must_not_strand_the_subject_payload() {
+    let _guard = CopyingNurseryTestGuard::new(0);
+    let _trigger = GcTriggerThresholdTestGuard::suppress_automatic_triggers();
+    let _mode = crate::arena::ProtectionModeGuard::set(FromSpaceProtection::PoisonOnly);
+    register_runtime_handle_root_scanner_for_tests();
+
+    // Decomposed "café-normalize": NFC must compose e+U+0301 into U+00E9, so a
+    // pass-through cannot be mistaken for a correct normalization.
+    const SUBJECT: &[u8] = "cafe\u{301}-normalize".as_bytes();
+    const EXPECTED: &[u8] = "caf\u{e9}-normalize".as_bytes();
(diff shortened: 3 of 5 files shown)


proggeramlug:
Superseded by #8483. The implementation in #8451 remains correct, but after #8481 landed it acquired the single `inline-offset | perry-runtime: baseline 370, found 371` lint failure. I could not push the helper-only gate fix to #8451’s head branch, so I pushed the rebased change from my fork and opened #8483. The superseding PR changes only the test’s payload-pointer lookup to `crate::string::string_data`, preserves the rooting regression semantics, and credits @proggeramlug for the original work.

proggeramlug:
Superseded by #8483 — your fix rebased onto current `main` with one addition.

The blocker changed while this sat open: the original `raw_handle_debt` failure resolved itself as `main` moved, but #8481 landed a new `string_payload_access_inventory` ratchet that flagged the test's open-coded `StringHeader` payload offset (370 → 371). #8483 routes it through `crate::string::string_data()` instead, leaving the assertion unchanged.

Fork branch, so the gate fix could not be pushed here. Full validation is on #8483.

PR #8452, opened 2026-08-20
fix(hir): ambient `declare function require` must not shadow the require intrinsic

### LINKED ISSUE #8447: regression(hir): ambient `declare function require` shadows the require intrinsic since #8343 — links against a nonexistent `require` symbol
Since #8343 (6674f5992, 2026-08-18), a module that carries the standard TypeScript ambient declaration

```ts
declare function require(name: string): any;
const fs = require("node:fs");
```

fails to **link**:

```
  "_require", referenced from: perry_fn_..._warmAndSummarize ...
```

**Mechanism**

1. A body-less `declare function require(...)` has always been registered as an external FFI declaration (`register_imported_func`) by `lower_module_fn.rs` — the generic "no body ⇒ extern" rule.
2. Before #8343 this was harmless for the `const x = require("<builtin>")` shape: the destructuring var-decl pass consumed the call into a native-module namespace binding *before* call lowering, so the FFI registration was never consulted.
3. #8343 added `require_is_shadowed_by_local`, which bails out of that destructuring path when `lookup_imported_func("require")` is Some — which the ambient declare makes true. The call now falls through to plain call lowering, resolves to the "imported func" `require`, and codegen emits a call to the symbol `require`, which no archive defines (it was always a compile-time intrinsic).

**Blast radius**: `test-files/test_issue_8002_8003_thread_realm_caches.ts` (COMPILE_FAIL in the 08-19 dispatch run 32298711372: compile-smoke red + one of parity's 27 compile_fails). Any user code using the ambient-declare idiom hits the same wall.

**Fix direction**: an ambient `declare function require` *names the global intrinsic* — it must not register as an FFI extern. #8343's CJS-wrap shadowing (a synthetic `function require(...)` **with a body**) is unaffected: that registers via `register_func`/locals, which correctly keeps shadowing.

Reproduced locally on a 0.5.1512 binary (macOS arm64); same failure as CI's Linux `undefined reference to 'require'`.

proggeramlug:
Fixes #8447 — one of the release blockers from full-tier run 32298711372 (compile-smoke red; also one of parity's 27 compile_fails).

**What broke.** A body-less `declare function require(name: string): any` — the standard TS ambient idiom for "the global CommonJS require exists" — has always been registered as an external FFI declaration by `lower_module_fn.rs` (the generic "no body ⇒ extern" rule). That was harmless until #8343 added `require_is_shadowed_by_local`, which counts `lookup_imported_func("require")` as shadowing: the destructuring fast path that used to consume `const fs = require("node:fs")` into a native-module namespace binding now bails, the call lowers as a plain call to the "imported" `require`, and codegen emits a reference to a `require` symbol that no archive defines:

```
  "_require", referenced from: perry_fn_..._warmAndSummarize ...
```

**Fix.** Skip the FFI registration for a body-less `require` declaration — it names the compile-time intrinsic, not a linkable extern (an FFI call to it can never link). #8343's CJS-wrap shadowing is untouched: the wrap's synthetic `function require(...)` HAS a body, registers via `register_func`, and still shadows (covered by the new counterpart test).

**Validation**
- New HIR test `test_ambient_require_declare_does_not_shadow_the_intrinsic` fails without the fix, passes with it; counterpart `test_user_require_function_with_body_still_shadows_the_intrinsic` guards #8343's intent.
- Full `cargo test -p perry-hir --lib`: 317 passed / 0 failed.
- End-to-end: `test-files/test_issue_8002_8003_thread_realm_caches.ts` (the compile-smoke failure) now compiles, links, runs, and its output byte-matches `test-parity/expected/`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed handling of ambient `declare function require(...)` declarations so they no longer override the built-in CommonJS `require`.
  * User-defined `require` functions with implementations continue to work as expected.

* **Tests**
  * Added regression coverage for both ambient declarations and implemented functions.


proggeramlug:
**CI classification: every red on this PR is pre-existing on `main`; this PR introduces none.**

The four failing gap shards report exactly these regressions:

| shard | tests |
|---|---|
| 1 | `test_gap_7911_promise_all_slow_arm_guards`, `test_gap_eval_as_value` |
| 3 | `test_gap_6558_webassembly_graceful_fail`, `test_gap_console_validate_write` |
| 4 | `test_gap_3662_collection_brand_check`, `test_gap_generator_return_throw_finally`, `test_gap_util_helpers_3026plus` |
| 5 | `test_gap_5592_yield_star_value_get_abrupt` |

All eight are a strict subset of the 15 regressions `main` itself carries at this PR's base commit (526e0b502), reproduced identically in three consecutive scheduled runs — [32328533570](https://github.com/PerryTS/perry/actions/runs/32328533570), [32323494876](https://github.com/PerryTS/perry/actions/runs/32323494876), [32311579863](https://github.com/PerryTS/perry/actions/runs/32311579863). Main's full set is 14 `pass → crash` plus `test_gap_net_connect_bound_value: pass → parity_fail`.

Independent of the run-comparison, this PR's change is structurally inert for those tests: it only alters lowering for a **body-less `function require` declaration**, and none of the eight test files contains the token `require` at all.

`pr-gate` therefore cannot go green here until main's own regressions are fixed — that is what #8464 (the 14 crashes, unverified until Linux CI reports) and #8466 (the `net_connect` parity fail, verified end-to-end) are for. Merging this one needs either those to land first or an admin bypass.

PR #8454, opened 2026-08-20
perf(codegen): append strings through variable cells

### LINKED ISSUE #8432: perf(codegen): string accumulators outside plain locals (properties, captures/boxed async locals, module globals) never reach the amortized append path — quadratic build
## Problem

Perry's only amortized-O(n) string-building path is `js_string_append` (`crates/perry-runtime/src/string/append.rs:30`): in-place when `refcount == 1 && new_len <= capacity`, 2× geometric growth on spill. Codegen reaches it from exactly one syntactic shape — `s = s + rhs` where `s` is a **plain, string-typed local** (`crates/perry-codegen/src/expr/literals_vars.rs:487-531`; `s += x` desugars to that in HIR). PR #8417 extended it to loop-local chains `s = s + a + b + c` (#8394, the 514×-Node quadratic).

Every other accumulator shape is O(n²) — each iteration copies the whole accumulated prefix:

| Shape | Why it misses the append path |
|---|---|
| module-level `let s; s += x` | `!ctx.module_globals.contains_key(id)` gate |
| closure-captured `s += x` | `!ctx.closure_captures.contains_key(id)` gate |
| boxed var (TDZ / shared-mutable) | `!ctx.boxed_vars.contains(id)` gate — **includes every async-function body local**, since the async transform boxes them (CLAUDE.md "known-weak areas") |
| `this.buf += x` / `obj.s += x` | lowers as PropertySet of `Add(PropertyGet, rhs)`; no self-append detection |
| `arr[i] += x` | same, via IndexSet |
| `let s: any = ""` | `local_type_hint` gate (annotation-driven selection) |

Accumulate-into-a-field and accumulate-in-an-async-function are both common real-JS shapes (response body building, log/report assembly).

## The design constraint a worker MUST understand first

The append path's soundness rests on an **opt-out aliasing protocol**: `refcount` is a 1-bit hint (`0 = shared, never mutate; 1 = unique`), and codegen demotes to shared at every aliasing site via `js_string_addref` (`let y = x`, argument passing, stores into arrays/objects — 27 emit-sites; runtime chokepoint `runtime_store_jsvalue_slot`, `crates/perry-runtime/src/gc/barrier_store.rs:100-129`, which demotes any `STRING_TAG` value stored into a heap slot). A missed demote is silent corruption of an aliased string (see `crates/perry/tests/string_append_heap_alias.rs`; regression history: #5552).

Consequently the shapes divide:

- **Property/array-element accumulators are protocol-blocked**: a string stored in an object slot is demoted to shared *by the store itself*, so even if codegen emitted `js_string_append`, the in-place arm can never fire — and removing the demote is unsound because any other reference to the object can extract the string between iterations. Fixing this shape needs either escape analysis Perry doesn't have, a protocol extension (e.g. an append-into-slot helper plus demote-on-*read* for the specific slot family), or a representation change (cons strings — see the discussion on #8394). **Recommended scope: analyze and document; do not force a fix here if the protocol extension doesn't fall out cleanly.**
- **Boxed / captured / module-global accumulators are protocol-feasible**: the box/capture-cell/global is the *variable itself*, not an alias of the string. The pattern `read cell → append → write cell back` preserves uniqueness IF every *other* read of that cell demotes what it extracts (mirroring what `let y = x` does for plain locals today). That demote-on-extraction rule needs to be added to the box/capture/global **read** paths (`js_closure_get_capture_bits`, boxed-cell loads, `@perry_global_*` loads) — likely only when the extracted value flows somewhere that can alias, which is the same analysis `let_stmt.rs:1812` already applies for locals.

**Scope recommendation:** deliver the boxed/captured/global shapes (measurable via an async-body accumulator benchmark, since async boxes ALL body locals), and a written verdict on the property shape.

## Measurement (required for any perf claim)

Fixture per shape, e.g. async body:
```ts
async function build(n: number): Promise<number> {
  let s = "";
  for (let i = 0; i < n; i++) s = s + "[" + "abc" + "]";
  return s.length;
}
```
Time at n = 2k/4k/8k/16k against Node 26.5.1 (the #8394 methodology; that issue's table is the template — Perry showed ~4-7× growth per doubling, Node flat). After the fix, growth must be ~2× per doubling and output byte-identical to node. Force materialization (read every ~997th `charCodeAt`) to rule out deferred work.

## Validation

- IR-level test in `crates/perry-codegen/src/codegen/declared_string_add_tests.rs` (see `a_self_append_chain_retains_the_accumulator_and_fuses_only_the_suffix` from PR #8417 as the template) asserting `js_string_append` is selected for the newly covered shape and NOT selected when the head pair could be numeric (annotation lie: `let s: string = 42 as any`).
- Alias-preservation fixture: extract the accumulator to another variable mid-loop; the extracted snapshot must not change on later appends (this is the demote-on-extraction rule working).
- `test-files/test_issue_<N>_*.ts` fixture byte-compared against Node 26.5.1.
- GC instrument pass on the fixture: `PERRY_GC_SCHEDULE_SEED=1 PERRY_GC_SCHEDULE_RATE=1 PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_VERIFY_EVACUATION=1 PERRY_GC_FORCE_EVACUATE=1`.
- Build note: `cargo build --profile perry-dev -p perry -p perry-runtime-static -p perry-stdlib-static`; run perry-runtime tests `RUST_TEST_THREADS=1`.

## Workflow

PR = code + tests + `changelog.d/<PR>-<slug>.md`; no version bump (maintainer bumps at merge). Check for sibling work in flight before starting (PR #8417 touches the same lowering file).

proggeramlug:
## Summary

- route string self-append through plain slots, boxes, closure captures, and module globals
- preserve unique ownership for the binding's internal read while demoting ordinary extracted reads
- cover reassignment aliases (`snapshot = accumulator`) in addition to declaration aliases
- preserve runtime tag dispatch for lying `string` annotations and numeric-capable head pairs

Closes #8432.

## Property/index verdict

Property and array-element accumulators are intentionally unchanged. `runtime_store_jsvalue_slot` demotes every heap string stored in an object/array slot, so selecting `js_string_append` there cannot reach its in-place arm. Removing that demotion would corrupt aliases reachable through another reference to the object. This shape needs escape analysis, a slot-specific demote-on-read protocol, or a different string representation; none falls out safely from this binding-cell change.

## Measured

Async boxed-local benchmark, 20 builds per size, with every 997th byte read through `charCodeAt` after construction. Checksums matched Node 26.5.1 at every size.

| iterations | Node 26.5.1 | patched Perry (`perry-dev`) |
|---:|---:|---:|
| 2,000 | 1 ms | 1 ms |
| 4,000 | 2 ms | 3 ms |
| 8,000 | 3 ms | 7 ms |
| 16,000 | 5 ms | 17 ms |

## Validation

- 17 focused declared-string/codegen IR tests passed
- `RUST_TEST_THREADS=1 cargo test -p perry-runtime --lib`: 2,599 passed, 4 ignored
- fixture output byte-matched Node 26.5.1 for globals, captures, async boxes, alias snapshots, and materialized bytes
- full forced-evacuation GC schedule passed: 51 scheduled collections, 103 moved objects

No version bump or central changelog edit is included.

## Summary by CodeRabbit

* **Performance**
  * Improved repeated string concatenation for global, captured, boxed, and asynchronous variables, reducing unnecessary allocation while preserving correct results.

* **Bug Fixes**
  * Fixed string ownership and aliasing behavior when reading or assigning string bindings, preventing unintended mutations.

* **Tests**
  * Added coverage for large string accumulations, snapshots during concatenation, and behavior across module, closure, and asynchronous contexts.

* **Documentation**
  * Added a changelog entry describing the string accumulation improvements.

### DIFF
--- a/changelog.d/8454-string-accumulators.md
+++ b/changelog.d/8454-string-accumulators.md
@@ -1,0 +1,6 @@
+### Performance
+
+- Keep string accumulators stored in async/closure variable boxes, captures,
+  and module globals on Perry's amortized append path. Ordinary reads and
+  local-to-local assignments still demote extracted aliases before later
+  in-place growth.
--- a/crates/perry-codegen/src/codegen/declared_string_add_tests.rs
+++ b/crates/perry-codegen/src/codegen/declared_string_add_tests.rs
@@ -106,8 +106,7 @@ fn ir(params: Vec<Param>, body: Expr) -> String {
 }
 
 fn function_ir(module: Module) -> String {
-    let ir =
-        String::from_utf8(compile_module(&module, ir_opts()).unwrap()).expect("LLVM IR is UTF-8");
+    let ir = module_ir(module);
     // An ordinary typed parameter may now produce a public guard wrapper plus
     // proof-bearing and generic clones (#8079). This suite's subject remains
     // the annotation-distrusting body, so inspect the generic clone when one
@@ -132,6 +131,10 @@ fn function_ir(module: Module) -> String {
         .map(|offset| start + offset + 3)
         .unwrap_or(ir.len());
     ir[start..end].to_string()
+}
+
+fn module_ir(module: Module) -> String {
+    String::from_utf8(compile_module(&module, ir_opts()).unwrap()).expect("LLVM IR is UTF-8")
 }
 
 fn add(left: Expr, right: Expr) -> Expr {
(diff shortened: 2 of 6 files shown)


proggeramlug:

reproducers.

PR #8459, opened 2026-08-20
test(native-abi): restore a static write-barrier site to the control packet (#8456)

### LINKED ISSUE #8456: native-abi packet: control fixture lost its only STATIC write-barrier site (push barrier moved into the runtime helper) — write_barriers_static=0 fails the packet contract
`native-abi-evidence-packet` fails in full-tier run 32298711372 on exactly one contract row:

```
packet_contract: write_barriers_static must be positive for native_abi_packet_control (got 0.0)
```

The discriminating pair from the packet's own evidence:
- `write_barriers_traced` (runtime GC counter): control = **31,510**, typed = 0 → the traced arm PASSES with a 100% reduction. Barriers are alive at runtime.
- `write_barriers_static` (count of `js_write_barrier` + `js_write_barrier_slot` calls in optimized IR): control = **0**.

So the control workload still performs tens of thousands of write barriers, but none of them appear as static helper calls in its IR anymore: the fixture's only codegen-visible barrier site was `controlBoxes.push(boxed)`, and the array-push barrier moved inside the runtime helper (`js_array_push_f64` — #7511 family), where the static counter cannot see it. The `write_barriers_static ≥ 1` control contract went stale — this is a fixture/metric staleness, not a product regression.

Verified locally: a class-instance heap-field store (`a.next = b`) still emits a static `js_write_barrier_slot`, so the control fixture can regain a representative static-barrier site without weakening the contract.

**Fix**: add a boxed link-stitch loop (class-field stores of heap values) to `benchmarks/compiler_output/fixtures/native_abi_packet_control.ts`, and add `js_write_barrier_slot` to that workload's `allowed_hot_loop_runtime_calls` (the contract itself demands the site exist). The typed packet keeps zero static barriers, so the reduction row proves the same claim honestly again.

Part of the release-blocker set: this job is in `full-suite-gate`'s needs, so it must be green on the release SHA. Note it only runs on dispatch/tag full tiers, so it had not executed since early July — the staleness window is the whole dark period, not this week's commits.

proggeramlug:
Fixes #8456 — the `native-abi-evidence-packet` release blocker from full-tier run 32298711372.

The packet's report failed one contract: `write_barriers_static must be positive for native_abi_packet_control (got 0.0)`. The discriminating evidence pair showed this is metric staleness, not a product regression: the runtime-traced counter saw **31,510** write barriers in the control run (and the typed packet's 100%-reduction row kept passing), while the static IR count of `js_write_barrier`/`js_write_barrier_slot` call sites was 0 — the fixture's only static barrier site was `controlBoxes.push(boxed)`, and the array-push barrier moved inside the runtime helper where the static counter cannot see it.

**Change**
- `native_abi_packet_control.ts`: add a boxed link-stitch loop (class-field stores of heap values), which lowers to inline stores + `js_write_barrier_slot` — measured **6 static sites** against the budget of 64. The control's "boxed representative" character is preserved; the typed packet is untouched, so the reduction row proves the same claim honestly again (6 → 0 = 100% ≥ 75% threshold).
- `workloads.toml`: allow `js_write_barrier_slot` in the workload's hot-loop runtime-call list (the packet contract itself demands the site exist), and update the pinned stdout checksum for the stitch contribution (pure integer arithmetic, platform-deterministic).

**Validation**: `scripts/compiler_output_regression.py suite --suite native-abi-proof --gate` fully green locally (`failed_workloads: []`); all runtime budgets within caps — allocations_traced 128/640, gc_collections_traced 1/5, write_barriers_traced 9091/50000, boxed_number_allocations_static 6/64.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when processing native ABI control data involving linked values, boxed numbers, and tagged references.
  - Updated memory write-barrier handling to ensure control data remains consistent during runtime operations.

- **Tests**
  - Expanded coverage for native ABI control scenarios, including checksum validation and hot-loop behavior.
  - Added documentation for the expected runtime behavior and verification results.


proggeramlug:
CI classification: **clean, no bypass needed.** This PR's diff is `benchmarks/` plus a changelog fragment, which `ci_plan.py` classifies as non-core — `gap-suite`, `cargo-test` and `e2e-scoped` all skip, so main's gap regressions cannot touch it. Every check that ran passed (`lint`, `plan`, the self-test checkers); only the `pr-gate` fan-in is still reporting. It should reach a green gate on its own.

Local validation is in the description: `scripts/compiler_output_regression.py suite --suite native-abi-proof --gate` is fully green (`failed_workloads: []`) with the new static-barrier site measured at 6 sites against a budget of 64.

Issue #8434, opened 2026-08-19
perf(runtime): join/repeat/pad/replace build through a Rust String round-trip (double copy; join lacks with_capacity)


proggeramlug:
## Problem

Several string-producing builtins build their result in a Rust `String`/`Vec<u8>` on the Rust heap, then copy it into a fresh `StringHeader` — two full copies plus a malloc/free per call, and in one case no capacity hint at all:

- `Array.prototype.join` — `crates/perry-runtime/src/array/iter_methods.rs` (~line 1122): `let mut result = String::new();` — not even `with_capacity`, so the Rust-side buffer reallocates log(n) times on top of the final copy.
- `String.prototype.repeat` — `crates/perry-runtime/src/string/pad.rs:314`: `str_data.repeat(count)` then copy.
- `padStart`/`padEnd` — `crates/perry-runtime/src/string/pad.rs` (`finish_pad_result`, ~line 186): Rust-side assembly then copy.
- regex `replace` paths — `crates/perry-runtime/src/regex/replace_fn.rs` (~lines 84/103) and siblings: Rust `String` assembly then copy.

## Important context — why the naive fix is WRONG

Building in a Rust-heap buffer first is currently the **GC-safe** pattern: Rust-heap allocation never triggers a Perry collection, so the borrowed source bytes stay valid while building; the single GC allocation happens at the end with an owned source. Writing directly into a `StringHeader` payload means the GC allocation happens FIRST — and that allocation can collect and **move the source strings**, invalidating any raw source pointers taken before it.

The sound direct-write shape already exists in-tree; follow it exactly:

- `js_string_alloc_ascii_uninit` (`crates/perry-runtime/src/string/mod.rs:679`) — allocate the destination first, then fill; 5 callers in `buffer/` show the pattern.
- `string_copy_range` (`string/mod.rs:704`) — root source, allocate, **re-read source from the rooted handle after the allocation**, then copy.
- `string/split.rs:479-520` — records byte *ranges* instead of borrows and re-derives the source pointer from a rooted handle per use.

[...]

PR #8460, opened 2026-08-20
perf(runtime): reserve Array.join separator capacity

proggeramlug:
## Summary

- reserve the exact `(length - 1) * separator byte length` lower bound before assembling `Array.prototype.join`
- preserve the existing single-pass element coercion and Rust-owned GC-safe assembly buffer
- add Node parity coverage for holes, empty separators, Unicode, and a mutating/allocating `toString`

## Measurements

I first prototyped rooted, two-pass direct writes into the GC string allocation. On the issue workload (10,000 eight-byte strings, 2,000 joins), that prototype regressed to 571.9 ms ± 76.0 from 510.1 ms ± 98.9; user CPU rose from 219.0 ms to 362.4 ms, so I rejected it.

The capacity-only fallback is within measurement noise over 20 settled hyperfine runs: 216.2 ms ± 0.9 versus 215.9 ms ± 1.1 at baseline (user CPU 197.4 ms versus 197.3 ms). The CPU Counters default template likewise reported 1,289,392 precise sampled cycles over 220 samples versus 1,275,820 over 217 samples; it did not expose a retired-instruction total.

Following the issue guidance, this keeps the simpler Rust `String` path and lands only the minimum low-risk reservation. It does not claim a benchmark speedup.

## Validation

- Node/Perry output parity for `test_issue_8434_array_join_capacity.ts`
- `RUST_TEST_THREADS=1 cargo test --release -p perry-runtime` (2,599 passed, 4 ignored)
- `cargo fmt -p perry-runtime -- --check`

No version bump.

Addresses the `Array.prototype.join` family from #8434.

## Summary by CodeRabbit

- **Performance**
  - Improved `Array.prototype.join` performance by preallocating output capacity more accurately.

- **Bug Fixes**
  - Preserved correct joining behavior for sparse arrays, empty strings, Unicode values, and dynamically converted elements.

- **Tests**
  - Added regression coverage for array mutations and allocations during element conversion.


proggeramlug:
Validated in a batch (#8460, #8445, #8478) stacked on `f53ae5080` and built once.

- `perry-runtime --lib` 2604 · `perry --bin perry` 1008 · `perry-codegen --lib` 1113
- `scripts/run_lint_gates.sh` — all 52 gates (50 script + compile tier)

PR #8464, opened 2026-08-20
fix(runtime): extend #8416's C-unwind to the whole closure dispatch family (#8463)

### LINKED ISSUE #8463: runtime: #8416 made dynamic Function errors unwind, but only js_closure_call1 is C-unwind — the rest of the dispatch family aborts (nightly bun_ffi_stage1 red)
Nightly full-tier run 32332070993 (526e0b502) fails cargo-test on `tier1_every_ffi_type_against_test_dylib` (crates/perry/tests/bun_ffi_stage1.rs:331): the compiled binary completes its FFI checks then dies with

```
thread '<unnamed>' panicked at library/core/src/panicking.rs:
panic in a function that cannot unwind
```

**Cause**: #8416 ("let dynamic Function errors unwind") converted `js_closure_call1` — and only the OUTER signature of `js_closure_call2` — to `extern "C-unwind"`. The rest of the closure dispatch family kept plain `extern "C"`, which installs Rust's abort-on-unwind shim: `js_closure_call0`, `call2`'s inner transmuted fn type, `call3`…`callN`, `dispatch_with_arity`'s per-arity transmutes (the COMMON arity-adjusted path), `js_closure_call_array`, `js_closure_call_apply_with_spread`, the direct-call-site cache, `js_function_bind`. Any JS exception that now unwinds (per #8416's intent) through one of those edges aborts instead of reaching the generated caller's landing pad. Linux-visible (same shape as #8416's original symptom); does not reproduce on macOS.

**Fix**: mirror #8416 across the whole family — every closure-dispatch entry point and transmuted call edge becomes `extern "C-unwind"`. Registration entry points (`js_register_closure_*`) stay `extern "C"` (no user code runs under them).

Validated locally (macOS): 64 closure unit tests, all 3 `bun_ffi_stage1` tests, and the #8416 regression suite (`function_apply_dynamic_args_eval_surface`) green. Linux verification rides the PR's scoped CI (the suite file is named in the diff).

proggeramlug:
Fixes #8463 — the new cargo-test red in nightly full-tier run 32332070993 (`tier1_every_ffi_type_against_test_dylib`: "panic in a function that cannot unwind → aborting", Linux).

#8416 let dynamic Function errors unwind to the generated caller's catch landing pad, but converted only `js_closure_call1` (and `js_closure_call2`'s outer signature — its inner transmuted fn type still aborted). Every other closure-dispatch edge kept plain `extern "C"`, whose abort-on-unwind shim turns the very unwind #8416 introduced into an abort the moment an error crosses `call0`, `call3`…`callN`, `dispatch_with_arity`'s per-arity transmutes (the common arity-adjusted path), `js_closure_call_array` / `js_closure_call_apply_with_spread`, the direct-call-site cache, or `js_function_bind`.

This converts the family consistently — outer signatures **and** transmuted fn-pointer types (a call through an `extern "C"` fn type that actually unwinds is UB/abort regardless of the callee's own ABI). Registration entry points (`js_register_closure_*`) stay `extern "C"`: no user code runs under them.

**Validation**
- macOS: 64 closure unit tests, all 3 `bun_ffi_stage1` tests, and the #8416 regression suite (`function_apply_dynamic_args_eval_surface`) green.
- The Linux abort itself cannot reproduce on macOS — `bun_ffi_stage1.rs` is named in this diff on purpose so the PR tier's `e2e-scoped` job runs the suite on ubuntu; that check is the fix's real verdict.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where errors raised by dynamically invoked functions could abruptly terminate Linux applications instead of unwinding normally.
  * Improved error propagation across closure calls, callbacks, bound functions, and spread/apply invocation paths.
* **Tests**
  * Expanded end-to-end coverage to verify reliable error handling across JavaScript closure dispatch scenarios.


proggeramlug:
**Reviewer note on what the local validation does and does not prove.**

The macOS results in the description confirm this change *breaks nothing* — they are **not** evidence that it fixes the abort. I checked: `tier1_every_ffi_type_against_test_dylib` passes on macOS against an **unfixed** post-#8416 build too, and the same holds for the `pass → crash` gap tests on main (`generator_return_throw_finally`, `util_helpers_3026plus`, `console_validate_write`, `fs_errprop2_2745plus` all pass locally without this fix). The gap harness says so itself: *"Reproduce on LINUX — several crash classes are masked on macOS by its 2 TB heap floor."*

So the case for this PR rests on two things:

1. **Static ABI reasoning** — after #8416 an exception really does unwind through these frames, and unwinding across a plain `extern "C"` boundary hits Rust's abort-on-unwind shim. That is exactly the reported failure (`panic in a function that cannot unwind → aborting`), and #8416 fixed the same symptom the same way for `js_closure_call1`.
2. **The Linux CI on this PR** — `bun_ffi_stage1.rs` is named in the diff so `e2e-scoped` runs it on ubuntu. That job is the real verdict.

Circumstantial support, not proof: main carries 14 `pass → crash` gap regressions, all error-path/brand-check shaped; `gap-suite (1)` passed on 55a5b5a06 (pre-#8416) and fails on 526e0b502 (post-#8416), a five-commit window containing #8416. If this PR's Linux gap shards come back clean where main's are red, that closes the loop — please don't merge on the macOS numbers alone.

PR #8466, opened 2026-08-20
fix(hir): a createRequire-backed local `require` is not a shadow (#8465)

### LINKED ISSUE #8465: regression(hir): #8343's require-shadow guard breaks the `createRequire(import.meta.url)` idiom — net.connect silently returns undefined (test_gap_net_connect_bound_value)
`test_gap_net_connect_bound_value` regressed `pass → parity_fail` on main (red in every recent sweep's gap-suite (3), and the last gap regression standing after #8464): `net.connect(port, host, cb)` returns `undefined` (then `TypeError: Cannot read properties of undefined (reading 'on')`), while `typeof net.connect` still says `function` and `net.createServer` works.

**Bisected to #8343 (6674f5992)** — its second casualty after #8447. The test uses Node's own ESM idiom:

```ts
import { createRequire } from "node:module";
const require = createRequire(import.meta.url);
const net = require("net");
```

Pre-#8343, the destructuring fast path rewrote `const net = require("net")` into the static native-module namespace binding unconditionally (ignoring the local `require`) — so `net.connect` compiled to the static codegen table and worked. #8343's `require_is_shadowed_by_local` guard (correct for the CJS wrap's synthetic `function require(...)`) also fires for this idiom, so the call now flows to the runtime createRequire namespace, where `("net","connect")`'s dispatch arm routes through `JS_NATIVE_HTTP_DISPATCH` — which is null unless the `external-http-server-pump` stdlib feature is compiled in and registered, and returns `undefined` when null. The gap-suite build never has that wiring, and neither do default-feature builds.

**Fix direction**: a local `require` initialized from `createRequire(...)` (imported `createRequire`, including renames) is Node-semantically the *real* module-scoped require — for builtin specifiers it returns exactly the builtin namespace. Mark that binding as createRequire-backed at declaration and exempt it from `require_is_shadowed_by_local`, restoring the pre-#8343 static namespace path for this idiom while keeping #8343's shadowing for the CJS wrap's synthetic `function require` (a real function with a body). The turbopack/mysql2 shape in the test header is exactly this idiom, so the regression bites real bundled apps, not just the gap suite.

Separately worth noting (not fixed here): the runtime `("net","connect")` arm silently returns `undefined` when `JS_NATIVE_HTTP_DISPATCH` is null — a throw ("net.connect requires the http server pump") would have turned this silent evaporation into a diagnosable error.

proggeramlug:
Fixes #8465 — the last standing gap regression on main after #8464 (`test_gap_net_connect_bound_value`: `pass → parity_fail` in every recent sweep's gap-suite (3); bisected to #8343, its second casualty after #8447).

**What broke.** #8343's `require_is_shadowed_by_local` guard is right for the CJS wrap's synthetic `function require(...)`, but it also fires for Node's own ESM idiom — `const require = createRequire(import.meta.url)`. That stopped `const net = require("net")` from folding to the static native namespace; the call flows to the runtime createRequire surface instead, where `net.connect` reached as a bound value dispatches through `JS_NATIVE_HTTP_DISPATCH` — null unless the `external-http-server-pump` stdlib feature is wired (it isn't in the gap-suite or default builds) — and **silently returns undefined** (`typeof net.connect` still says `function`; the failure surfaces as `TypeError: Cannot read properties of undefined (reading 'on')`). The test header's mysql2/turbopack shape is exactly this idiom, so it bites real bundled apps.

**Fix.** At declaration, mark a local `require` initialized from `createRequire(...)` (renamed imports resolved through the imported-function table) as createRequire-backed, and exempt exactly that binding from the shadow check — for builtin specifiers such a require returns precisely the native namespace, so the static fold is semantically correct and restores the battle-tested pre-#8343 route. A `function require` WITH a body still shadows via `lookup_func`, preserving #8343's CJS-wrap behavior.

**Validation**
- New HIR test pair: the positive test fails without the fix (verified by flip); the counterpart pins the CJS-wrap shadowing.
- Full `cargo test -p perry-hir --lib`: 317 green.
- End-to-end: the gap test compiles, runs a real echo server + two connect styles, and byte-matches the Node oracle.

Noted in #8465 but not fixed here: the runtime `("net","connect")` arm returning silent `undefined` on a null dispatch pointer deserves a throw so this class of evaporation is diagnosable.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed handling of local `require` bindings created with `createRequire(import.meta.url)`, including renamed imports.
  - Preserved static native-module behavior for these bindings.
  - Continued correctly treating user-defined, imported, or function-local `require` declarations as shadowing.
  - Maintained expected runtime behavior for custom `require` implementations.

- **Tests**
  - Added regression coverage for ambient, custom, and `createRequire`-based `require` behavior.


proggeramlug:
Rebased onto current main (through #8462) and re-verified.

**Conflict resolution**: `crates/perry-hir/src/lower/tests.rs` — #8452 and this PR both appended tests to the end of the file. Kept **both** sets; all four pass together (`test_ambient_require_declare_does_not_shadow_the_intrinsic`, `test_user_require_function_with_body_still_shadows_the_intrinsic`, `test_create_require_local_keeps_the_native_namespace_fast_path`, `test_function_require_with_body_still_shadows_the_namespace_fast_path`). The two counterpart tests are complementary rather than duplicates and now say so in their doc comments: #8452's pins that a real `function require` body suppresses the fold, this one additionally pins that the bound name survives as a runtime local.

**Post-rebase checks**: the diff against main is exactly the intended 6 files (+97/−2) with nothing from main dropped; `cargo test -p perry-hir --lib` green; and the gap fixture re-verified end-to-end on the rebased tree against the post-#8464/#8458 runtime — `test_gap_net_connect_bound_value` compiles, runs its echo server, and byte-matches the Node oracle.

PR #8480, opened 2026-08-20
fix(ffi): let bun:ffi symbol thunks unwind (#8479)

### LINKED ISSUE #8479: ffi: bun:ffi symbol thunks are extern "C" — a use-after-close throw aborts on Linux (cargo-test still red after #8464)
`cargo-test` still fails on `main` after #8464: `tier1_every_ffi_type_against_test_dylib` aborts on Linux with

```
thread '<unnamed>' panicked at library/core/src/panicking.rs:225:5:
panic in a function that cannot unwind
```

Measured on the scheduled sweep of `15a30d7f6` (run 32353180323), which **contains #8464** — so the closure-dispatch conversion did not fix this.

**Where it actually aborts.** The captured stdout stops right after `external-terminated: 2` and never reaches `closed-throws:` — the step that deliberately throws:

```ts
lib.close();
try { s.ffi_i32_add1(1); } catch (e: any) { closedError = String(e && e.message); }
console.log("closed-throws:", closedError.includes("close()"));
```

That throw originates in `bun_ffi::dlopen::invoke_from_closure`, which calls `throw_error_with_code("bun:ffi: symbol … was called after close() …")`. It is reached through the per-arity stubs generated by `sym_thunk!`:

```rust
extern "C" fn $name(closure: *const ClosureHeader $(, $a: f64)*) -> f64 {
    unsafe { invoke_from_closure(closure, &args) }
}
```

Those stubs are **`extern "C"`**, i.e. nounwind, so the unwind hits Rust's abort-on-unwind shim *inside the thunk itself*. #8464 made the closure-dispatch **callers** unwind-capable; these thunks are the **callees** it could not see — a call through an `extern "C-unwind"` function pointer does not remove the callee's own nounwind guard.

`close_thunk` has the same shape: it takes a `Mutex` and `unwrap()`s, and a poison panic there would abort identically.

**Fix**: generate both as `extern "C-unwind"` (PR #8479).

**Process note worth recording.** #8464 was merged while its own `e2e-scoped` job — the Linux run of exactly this suite, which the PR deliberately opted into by naming the test file — was still in flight; merging cancelled it (run 32340796570, job cancelled at 08:52). The check that would have caught this was destroyed by the merge. macOS passes this test with and without either fix, so the Linux `e2e-scoped` result is the only verdict, and it must be allowed to report before merging.

proggeramlug:
Fixes #8479 — `cargo-test` is **still red on main after #8464**, and this is why.

Measured on the scheduled sweep of `15a30d7f6` (run 32353180323), a SHA that already contains #8464: `tier1_every_ffi_type_against_test_dylib` aborts on Linux with `panic in a function that cannot unwind`.

**The abort is in the FFI thunks, not the dispatch family.** The captured stdout stops right after `external-terminated: 2` and never reaches `closed-throws:` — the step that deliberately throws:

```ts
lib.close();
try { s.ffi_i32_add1(1); } catch (e: any) { closedError = String(e && e.message); }
```

That throw is raised by `invoke_from_closure` ("symbol … was called after close()") and is reached through the stubs `sym_thunk!` generates, which were `extern "C"` — so the unwind hit Rust's abort-on-unwind shim *inside the thunk*. #8464 converted the closure-dispatch **callers**; these are the **callees** it could not see. Calling through an `extern "C-unwind"` function pointer does not remove the callee's own nounwind guard.

`close_thunk` gets the same conversion: it `unwrap()`s a `Mutex`, and a poison panic there would abort identically.

**Verification — please let CI report before merging.** macOS passes this test both with and without the fix (I ran it: 3/3 green here), so my local result proves only "no regression". The **Linux `e2e-scoped` run of this suite is the only real verdict**, and the test file is named in this diff specifically to opt into that job.

That matters because of how the previous attempt went: **#8464 was merged while its own `e2e-scoped` job was still in flight, and merging cancelled it** (run 32340796570, cancelled 08:52). The one check that would have shown #8464 didn't work was destroyed by the merge — which is why the bug survived to a sweep. I've added a comment in the test file recording the invariant so the next person doesn't repeat it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Bun FFI errors so use-after-close failures and related runtime exceptions are reported correctly instead of aborting the process.
  * Improved error unwinding for dynamic library calls, including cases involving mutex-related panics.

* **Documentation**
  * Clarified FFI exception-unwinding behavior and runtime validation expectations.


proggeramlug:
Validated and merging. The reasoning is right and the change is minimal.

`extern "C"` is `nounwind`, so a JS throw raised by `invoke_from_closure` (symbol called
after `close()`, or an unknown stub) hits Rust's abort-on-unwind shim **inside the thunk
itself** — hence `panic in a function that cannot unwind` rather than a catchable error.
#8464 made the dispatch *callers* unwind-capable; these thunks are the *callees* it could
not see. Widening them to `extern "C-unwind"`, plus `close_thunk` for its poisoned-`Mutex`
`unwrap()`, closes that gap.

## What I could and could not verify

- `bun_ffi_stage1` — **3 passed**, including `tier1_every_ffi_type_against_test_dylib`, the
  test that aborts on Linux
- `perry-runtime --lib` 2602 passed

**I could not reproduce the failure.** `main` passes the same suite on macOS (3 passed), so
this host does not discriminate — consistent with your report that the abort appears on
Linux. My run therefore shows no regression rather than confirming the fix. The argument
carries it: the `nounwind` semantics of `extern "C"` are not platform-specific, even though
whether the unwind reaches the shim depends on the platform's unwinder.

Worth noting the diagnosis in the PR body is the useful part — reading the captured stdout
stopping right after `external-terminated: 2` and never reaching `closed-throws:` is what
localises the abort to the thunk rather than the dispatch family. That is the kind of
evidence that makes a cross-platform fix reviewable from a machine that cannot run it.

PR #8483, opened 2026-08-20
fix(runtime): root the normalize subject across form coercion

proggeramlug:
## Summary

- Supersedes #8451 with the original fix rebased onto current `main`; credit to @proggeramlug for the original work.
- Keeps the normalize form-coercion rooting semantics and regression assertion unchanged.
- Replaces the test's new open-coded `StringHeader` payload offset with the centralized `crate::string::string_data` helper required by the #8481 debt ratchet.

The superseding PR is necessary because the original PR branch is on a fork branch I could not update with this gate fix.

## String payload gate

Before:

```text
String payload-access ratchet increased:
  inline-offset | perry-runtime: baseline 370, found 371
```

After:

```text
string-payload access inventory: 3020 files; 441 inline offsets and 14 reader helpers held by the ratchet
```

`python3 scripts/string_payload_access_inventory.py` exits 0. `(*result).byte_len` is not counted by either inventory rule, so it remains unchanged.

I also temporarily restored the bare payload offset after the fix. The gate failed again with `inline-offset | perry-runtime: baseline 370, found 371`; after removing it, the gate returned to green. No baseline, ceiling, or exclusion was changed.

## Validation

- `python3 scripts/string_payload_access_inventory.py` — pass
- `python3 scripts/raw_handle_debt.py` — pass (`978`, baseline `978`)
- `CARGO_TARGET_DIR=/Users/amlug/cargo-targets/w8451c bash scripts/run_lint_gates.sh` — all 52 gates passed
- `CARGO_TARGET_DIR=/Users/amlug/cargo-targets/w8451c cargo test --release -p perry-runtime --lib string_normalize_form` — 1 passed
- `CARGO_TARGET_DIR=/Users/amlug/cargo-targets/w8451c cargo test --release -p perry-runtime --lib` — 2605 passed, 4 ignored, 0 failed
- `CARGO_TARGET_DIR=/Users/amlug/cargo-targets/w8451c cargo test --release -p perry --bin perry` — 1008 passed, 0 failed
- `CARGO_TARGET_DIR=/Users/amlug/cargo-targets/w8451c cargo build --release -p perry -p perry-runtime-static -p perry-stdlib-static` — pass
- Compiled `test-files/test_issue_8426_normalize_reentrant.ts` with `PERRY_RUNTIME_DIR=/Users/amlug/cargo-targets/w8451c/release`, `PERRY_NO_AUTO_OPTIMIZE=1`, and `PERRY_NO_CACHE=1`; stdout was byte-identical to `/opt/homebrew/bin/node --experimental-strip-types` (v26.5.1).
- Compiled `iso_miss`, `interp`, `pipeline`, `fib40`, and `shapes` from `sweep-artifacts-0819b/sources` under the same environment; each stdout was byte-identical to its checked expected output.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `String.prototype.normalize` so strings remain valid when form coercion triggers garbage collection.
  * Preserved expected normalization behavior, including NFC defaults, NFD handling, and `TypeError` for Symbol forms.
  * Ensured form coercion continues to occur before validation.

* **Tests**
  * Added regression coverage for reentrant coercion, garbage collection, string relocation, and normalization results.

### DIFF
--- a/changelog.d/8451-normalize-form-coercion-rooting.md
+++ b/changelog.d/8451-normalize-form-coercion-rooting.md
@@ -1,0 +1,13 @@
+### Fixed
+
+- `String.prototype.normalize` no longer holds a borrowed heap-string payload
+  across its `form` argument's `ToString` coercion. The coercion is a
+  collection point twice over — an object form runs user `toString` (whose
+  loop back-edge polls run a moving minor), and an inline short-string form
+  materializes onto the heap — so a young subject could be evacuated while
+  `js_string_normalize` held a `&str` into its pre-move address, after which
+  the normalization pass read retired from-space. The form is now coerced
+  first, the subject is rooted across the coercion, and the payload is
+  borrowed only from the post-collection address. The observable orderings are
+  unchanged: `ToString` still runs before the form is validated, so a Symbol
+  form throws `TypeError` rather than the invalid-form `RangeError`. (#8426)
--- a/crates/perry-runtime/src/gc/tests/runtime_roots.rs
+++ b/crates/perry-runtime/src/gc/tests/runtime_roots.rs
@@ -15,6 +15,7 @@ mod old_defrag_contract;
 mod prototype_addr_cache;
 mod regexp_last_index;
 mod side_table_scanners;
+mod string_normalize_form;
 mod string_slice;
 mod symbol_description;
 mod transient_handles;
--- a/crates/perry-runtime/src/gc/tests/runtime_roots/string_normalize_form.rs
+++ b/crates/perry-runtime/src/gc/tests/runtime_roots/string_normalize_form.rs
@@ -1,0 +1,128 @@
+use super::*;
+use crate::arena::FromSpaceProtection;
+
+/// #8426: `String.prototype.normalize` borrowed the subject string's inline
+/// WTF-8 payload BEFORE coercing its `form` argument, then read that borrow
+/// after the coercion had returned.
+///
+/// The coercion is a collection point twice over: an inline short-string form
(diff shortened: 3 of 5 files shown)


proggeramlug:
Validated independently and merging.

The change is exactly the right shape: `crate::string::string_data(result)` replaces the
open-coded offset, and the assertion is untouched — this test forces a copying minor from
inside `normalize`'s form-coercion trap specifically to prove the subject stayed rooted, so
the payload read *is* the subject. Swapping how the pointer is obtained without changing
what is proven is the correct move.

## Verified

- `string_payload_access_inventory.py` — exit 0 (was `370 -> 371` before)
- `raw_handle_debt.py` — exit 0
- `check_test_registration.py` — exit 0
- `perry-runtime --lib` — 2605 passed
- `string_normalize_form` test — passes
- `test_issue_8426_normalize_reentrant.ts` — matches Node byte for byte

## The part I most wanted to see

You re-checked that the ratchet still fires: restored the bare offset, confirmed
`inline-offset | perry-runtime: baseline 370, found 371`, then removed it. That is the check
that distinguishes "made the gate green" from "made the code right" — a gate that stops
failing is worse than the debt it tracked. And no baseline, ceiling, or exclusion was
touched, which was the tempting shortcut here.

Also correct to note that `(*result).byte_len` is not counted by either inventory rule and
to leave it alone rather than inventing a helper to satisfy a counter.

Credit to @proggeramlug for the original fix in #8451; this exists only because that is a
fork branch the gate fix could not be pushed to.

PR #8487, opened 2026-08-20
perf(string): bypass empty accumulator copies

### LINKED ISSUE #8486: perf: #8417 regressed iso_miss +24.8% (+27.2% instructions) while fixing #8394's quadratic concat
## #8417 fixed #8394's quadratic concat but regressed `iso_miss` by 24.8%

Bisected on the quiet bench mini sweep. Clean adjacent-commit attribution:

| commit | `iso_miss` instructions |
|---|---:|
| `ddc1cd845` (parent) | **11.54 G** |
| `5f9aa1f40` — #8417 "retain string accumulators in concat chains" | **14.68 G** |

That is **+27.2% instructions**, and on the mini it shows as wall time 0.6810 s -> 0.8501 s
(**+24.8%**), pushing `iso_miss` from 1.96x to **2.53x Node**. The sample ranges are fully
disjoint (previous max 0.7080 < current min 0.8501), so this is not noise.

`-Os` (#8457) is **not** the cause — forcing `PERRY_LL_SIZE_OPT=0` gives +0.8% with
overlapping samples, and both arms already sit at ~14.7 G.

## This is a real trade, not a mistake

#8417 did exactly what it claimed. Re-measured on the same fixture from #8394:

| n | before #8417 | after #8417 | Node |
|---:|---:|---:|---:|
| 2000 | 17 ms | 3 ms | 1 ms |
| 16000 | **1450 ms** | **16 ms** | 1 ms |

**90x at n=16000**, and the curve went from roughly quadratic to flat. That is a large,
real win on accumulator-shaped concatenation.

The problem is that `iso_miss` builds a string it usually throws away — its hot line is

```ts
seen = seen + "[" + names[i] + "]";   // iso_miss.ts:205, per name per env frame
```

and the result is only read on a lookup miss. Retaining the accumulator helps when the
string is kept and hurts when it is discarded. `iso_miss` is the discard case, and it is the
row where string concat was already ~18% of self time.

## What would resolve it

Ideally the retention becomes conditional rather than unconditional — retain when the
accumulator is subsequently read, fall back when it is dead. Whether that is cheaply
decidable in `literals_vars.rs` is the open question; if it is not, this is a genuine
policy choice about which shape to favour and should be recorded as one.

Note #8417 is small (`crates/perry-codegen/src/expr/literals_vars.rs`, +42/-1 plus tests),
so the surface to reason about is narrow.

## Process note

I merged #8417 after validating it against the #8394 scaling fixture and confirming 19/19
corpus byte-exactness — but I did not run a **timing** A/B across the corpus, only a
correctness one. A perf PR that changes codegen for a common operation needs the corpus
timed, not just diffed. That gap is why this took a full sweep to surface.

proggeramlug:
Fixes #8486.

## What happened

I instrumented `js_string_append` before changing it. The leading refcount hypothesis was not the hot shape: the `iso_miss` append samples were overwhelmingly the first append to the shared empty string (`dest_blen=0`, `capacity=0`, `refcount=0`). Frames with a second name then reached `refcount=1`, `capacity=32` and took the intended in-place path.

The regression was the common one-name frame doing redundant work after #8417: `js_string_concat_chain` built the complete `[name]` suffix, then `js_string_append` rooted both strings, allocated another buffer, and copied that suffix only to compute `"" + suffix`.

This change keeps #8417's accumulator-retaining lowering and adds a conservative identity path:

- empty destination + shared source returns the source unchanged;
- a unique source is never reused, because doing so would create an unrecorded alias with mutation permission;
- codegen's already tag-checked heap-string arms use a narrow entry point that takes the same identity path without repeating generic pointer validation;
- every mutating, allocating, self-append, or unique-source case delegates to the existing `js_string_append` implementation.

## Performance

Built both measured binaries with:

```sh
export PERRY_RUNTIME_DIR=/Users/amlug/cargo-targets/w8486/release
```

Three `/usr/bin/time -l` samples per cell on the contended development host; instructions retired are the primary signal.

| workload | metric | current main / before | this PR / after | result |
|---|---:|---:|---:|---:|
| `iso_miss` | instructions, median | 14.790 G | **12.471 G** | **-15.7%** |
| | wall, median (range) | 0.89 s (0.89-1.44) | 0.75 s (0.74-1.18) | ranges overlap; no wall claim |
| | peak RSS, median | 31.28 MiB | 31.19 MiB | -0.3% |
| #8394 fixture | instructions, median | 27.063 M | **27.023 M** | -0.15% / flat |
| | wall, median (range) | 0.00 s (0.00-0.37) | 0.00 s (0.00-0.72) | timer floor; no wall claim |
| | peak RSS, median | 5.23 MiB | 5.23 MiB | flat |

Relative to the established 11.54 G parent / 14.68 G #8417 attribution, the final 12.47 G result recovers about 70% of the instruction regression while preserving the more important asymptotic fix. The remaining ~8% versus the parent is the extra tag-dispatch/append stage retained by #8417; removing that fully would require a fused append-chain lowering rather than safely eliminating redundant allocation/copy/rooting.

The explicit #8394 scaling probe, `s = s + "[" + "abc" + "]"`, remained flat with correct output lengths:

| n | 2,000 | 4,000 | 8,000 | 16,000 |
|---:|---:|---:|---:|---:|
| wall ms | 0 | 0 | 1 | 0 |

It does not return to the pre-#8417 quadratic 17/56/342/1450 ms curve.

## Validation

- all 19 programs in `sweep-artifacts-0820/sources` exit 0 and are byte-exact against their pinned stdout
- `cargo test --release -p perry-runtime --lib`: 2,606 passed, 4 ignored
- `cargo test --release -p perry --bin perry`: 1,008 passed
- `cargo test --release -p perry-codegen --lib`: 1,113 passed
- `bash scripts/run_lint_gates.sh`: all 52 gates passed
- exact three-package release build passed

## Summary by CodeRabbit

* **Performance**
  * Improved string concatenation efficiency, reducing unnecessary copying in common append operations.
  * Preserved safe ownership and mutation behavior when appending empty or shared strings.
* **Bug Fixes**
  * Correctly handles surrogate pairs spanning concatenation boundaries.
  * Maintains byte-exact string results across tested scenarios.
* **Tests**
  * Added coverage for empty-string ownership, shared values, unique sources, and optimized append paths.

### DIFF
--- a/changelog.d/8487-empty-accumulator-identity.md
+++ b/changelog.d/8487-empty-accumulator-identity.md
@@ -1,0 +1,13 @@
+Avoid allocating and copying a completed concat suffix when a string builder
+appends it to the shared empty string. The optimization reuses only shared
+sources, so a uniquely owned source can never gain an untracked alias with
+in-place mutation permission. Codegen's tag-checked heap-string append arms use
+the same identity path without repeating generic pointer validation; all other
+cases retain the existing `js_string_append` behavior.
+
+On `iso_miss`, median instructions retired fall from 14.790 G to 12.471 G
+(-15.7%), recovering about 70% of #8417's measured regression while keeping
+#8394's accumulator-chain fix. The #8394 fixture stays flat at 27.0 M
+instructions, and its 2k/4k/8k/16k scaling probe remains 0-1 ms rather than
+returning to the pre-#8417 quadratic 17/56/342/1450 ms curve. Peak RSS is flat
+for both workloads, and all 19 sweep programs remain byte-exact.
--- a/crates/perry-codegen/src/codegen/declared_string_add_tests.rs
+++ b/crates/perry-codegen/src/codegen/declared_string_add_tests.rs
@@ -359,7 +359,7 @@ fn a_self_append_chain_retains_the_accumulator_and_fuses_only_the_suffix() {
     let ir = function_ir(module);
 
     assert!(
-        ir.contains("call i64 @js_string_append("),
+        ir.contains("call i64 @js_string_append_known_heap("),
         "the growing prefix must reach the amortized append path:\n{ir}"
     );
     assert_eq!(
@@ -423,7 +423,7 @@ fn a_self_append_chain_keeps_an_opaque_numeric_head_pair_intact() {
     let ir = function_ir(module);
 
     assert!(
-        !ir.contains("call i64 @js_string_append("),
+        !ir.contains("call i64 @js_string_append_known_heap("),
         "an opaque numeric-capable head pair must remain in source-tree order:\n{ir}"
     );
 }
@@ -455,7 +455,7 @@ fn a_module_global_self_append_uses_the_amortized_path_and_demotes_extractions()
     let ir = function_ir(module);
 
     assert!(
-        ir.contains("call i64 @js_string_append("),
+        ir.contains("call i64 @js_string_append_known_heap("),
         "a module root is binding storage and can retain the unique string owner:\n{ir}"
     );
     assert!(
@@ -512,7 +512,7 @@ fn a_boxed_local_self_append_uses_the_amortized_path() {
     let ir = function_ir(module);
 
     assert!(
-        ir.contains("call i64 @js_string_append("),
+        ir.contains("call i64 @js_string_append_known_heap("),
         "a variable box must retain the accumulator owner across iterations:\n{ir}"
     );
     assert!(
@@ -574,7 +574,7 @@ fn a_boxed_capture_self_append_uses_the_amortized_path() {
     let ir = module_ir(module);
 
     assert!(
-        ir.contains("call i64 @js_string_append("),
+        ir.contains("call i64 @js_string_append_known_heap("),
         "a captured variable box must reach the append helper:\n{ir}"
     );
     assert!(
@@ -608,7 +608,7 @@ fn a_module_global_numeric_capable_head_pair_does_not_select_append() {
     let ir = function_ir(module);
 
     assert!(
-        !ir.contains("call i64 @js_string_append("),
+        !ir.contains("call i64 @js_string_append_known_heap("),
         "the newly eligible storage must not weaken the numeric-head guard:\n{ir}"
     );
 }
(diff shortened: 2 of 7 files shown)


proggeramlug:
Independently reproduced and audited. Merging.

**Headline confirmed** (median of 3, instructions retired, same host, `-p perry -p perry-runtime-static -p perry-stdlib-static`):

| row | base | this PR | delta |
|---|---:|---:|---:|
| `iso_miss` | 14.788 G | 12.470 G | **−15.7%** |
| `interp` | 8.9951 G | 8.9950 G | ±0.0% |
| `pipeline` | 2.1654 G | 2.1656 G | ±0.0% |

19/19 corpus programs byte-exact, peak RSS flat at 31 MiB.

**Which hunk carries the win.** I built a third arm — the runtime identity-hoist with the three codegen callsites left on `js_string_append` — to check whether the new entry point earns its keep:

- base → runtime hoist only: 14.788 G → 12.560 G (**−15.1 pp**)
- runtime hoist → + `js_string_append_known_heap`: 12.560 G → 12.470 G (**−0.7 pp**)

So the hoist above `RuntimeHandleScope::new()` is essentially the whole result, and `known_heap` is a real but small tail. The ranges are disjoint (A max 12.501 vs B min 12.559), so the 0.7 pp is not noise. Recording it here because the split is not obvious from the diff, and a future bisect will want it.

**Aliasing argument checked.** `append(empty, shared) → src` is safe only if a `refcount == 0` string can never later gain in-place mutation permission. Audited every write in the tree: `refcount = 1` happens only in `append.rs:157` on a freshly allocated header and via `init_string_header`, which runs on fresh allocations only. Every other site drives 1 → 0 (the demote sites in `array/`, `gc/barrier_store.rs`). `append.rs:120` is the sole in-place mutation and requires `refcount == 1`. So the returned alias is permanently non-mutable in place. Sound.

**`known_heap` callsites checked.** It skips `is_valid_string_ptr` (a null + `>= 0x1000` test), so each of the three sites has to be heap-proven. They are: two are under a `TAG_HEAP_STR` tag check, and the SSO arm's `r_handle` comes from `js_get_string_pointer_unified`, which materializes rather than masking. The dest re-reads are correctly ordered *after* the allocating coercion in both arms.

Residual: `iso_miss` is now about +5% over its pre-#8417 level rather than +24.8%, so roughly 79% of the regression is recovered. That tail can stay on #8486 or get its own issue.

PR #8488, opened 2026-08-20
fix(runtime): remove the C-unwind abort guards from the JS throw path (#8479)

proggeramlug:
Refs #8479 — the `cargo-test` blocker (`tier1_every_ffi_type_against_test_dylib` aborting on Linux), and, I believe, the reason three previous attempts made it worse rather than better.

## The inversion

The runtime is built **`panic=abort`**, and a JS throw is a raw Itanium `_Unwind_Exception` that must step **through** runtime Rust frames untouched. `crate::eh` documents this, and the workspace `Cargo.toml` states the requirement outright:

> panic=abort (#7302): the invoke/landingpad exception transport requires the unwinder to step THROUGH runtime Rust frames with longjmp-equivalent semantics — no cleanups, and **no RFC-2945 abort-on-unwind guards (which panic=unwind plants in every extern "C" helper and which a JS throw crossing a helper frame would trip)**.

Marking a pass-through frame `extern "C-unwind"` in a `panic=abort` crate does **not** make it unwind-capable. It makes rustc treat the call as possibly-unwinding and wrap it in an **abort-on-unwind landing pad** — precisely the guard the design forbids. Hence the abort message is not a symptom of a *missing* `C-unwind`; it is a symptom of a *present* one.

That reading explains all three attempts at once:

| change | what it did | outcome |
|---|---|---|
| #8416 | `js_closure_call1` + `js_native_call_value` → `C-unwind` | first guards on the FFI path; tier1 passes at `55a5b5a06`, aborts from `526e0b502` on |
| #8480 | `bun_ffi` `sym_thunk_*` / `close_thunk` → `C-unwind` | another guard *on the same path* — could not help, and didn't |
| #8464 | ~40 dispatch functions → `C-unwind` | ~40 more guards: **+20 gap crashes**, `gc-stress` red (reverted in #8484; post-revert sweep measured 36 → **14**) |

## This PR

Puts all four pass-through conversions back to `extern "C"`, with a comment at each site explaining why `C-unwind` is wrong *here* specifically, so the next person doesn't re-add it.

**Scope, deliberately narrow**: throw *originators* (`js_throw`, `js_throw_type_error_not_a_function`, …) are left alone. They diverge (`-> !`) and are where the raise begins rather than frames a throw crosses; there is no evidence implicating them, and 20 further `C-unwind` sites elsewhere in the runtime are likewise untouched pending evidence.

## Verification

`bun_ffi_stage1` is 3/3 locally — **which proves nothing**, because macOS has never reproduced this abort under any variant, including plain `main`. **The Linux `e2e-scoped` run of this suite is the only verdict**, and this diff names the test file to opt into it.

Please let that job report before merging. #8464 was merged while exactly that job was in flight, the merge cancelled it, and the resulting blind merge is what cost main 20 crashes.

Two outcomes worth watching, since #8416 was itself fixing something:
1. `tier1_every_ffi_type_against_test_dylib` — expected to pass;
2. `function_apply_with_runtime_args_defers_to_a_located_aot_error` — the test #8416 was written for. If it regresses, then that bug needs a fix that doesn't plant a guard (the `Setjmp` transport in `exception.rs` is the natural place to look), and this PR should wait for it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runtime stability when JavaScript exceptions cross native function and FFI boundaries.
  * Ensured test binaries use the same release runtime and panic behavior as shipped builds, preventing misleading unwind-related failures.

* **Documentation**
  * Added release notes and test documentation covering exception handling, FFI behavior, and runtime configuration.


proggeramlug:
**Root cause found — and the two canaries are not actually in conflict.**

The first Linux run of this PR gave the result I flagged as the risk: `tier1_every_ffi_type_against_test_dylib` **passed**, `function_apply_with_runtime_args_defers_to_a_located_aot_error` **failed** — both with the same `panic in a function that cannot unwind` abort, under opposite configurations. That is not a real trade-off; it is an artifact of what CI links.

**`panic` is a profile-level setting**, and only `release` / `dist` / `perry-dev` set `panic = "abort"`. Both `cargo-test` and `e2e-scoped` build the static runtime archives with a bare `cargo build` — DEBUG, therefore `panic = "unwind"`. Under `panic = "unwind"` rustc plants an RFC-2945 abort guard in **every** `extern "C"` helper, which is exactly what the workspace `Cargo.toml` warns a JS throw crossing a helper frame will trip.

So the compiled child binaries have been linking a runtime whose unwind semantics are the **opposite** of the one Perry ships. In that environment the two canaries cannot both pass, and each previous attempt fixed one by breaking the other:

| | debug runtime (what CI linked) | shipped runtime (panic=abort) |
|---|---|---|
| `extern "C"` | guard present → throw aborts | no guard → throw passes ✅ |
| `extern "C-unwind"` | no guard → throw passes | guard present → throw aborts ❌ |

#8416 chose the left column (right for CI, wrong for production); #8464 generalised it to ~40 functions and cost +20 gap crashes plus a red `gc-stress`; #8480 added more of the same.

**Measured**: against a `panic = "abort"` runtime with those conversions removed, **both** canaries pass — `tier1` also confirmed green on Linux `e2e-scoped` with the integration-suites step verified to have actually run (it silently skips and reports green when no test file is in the diff, which produced a false green here earlier).

This PR now carries both halves, because neither is correct alone:
1. remove the `extern "C-unwind"` guards from the pass-through frames (correct for the shipped runtime);
2. build the runtime archives with `--release` in `cargo-test` and `e2e-scoped` so the tests exercise the shipping unwind regime instead of one Perry never ships.

proggeramlug:
**Merging. The one remaining red (`warnings`) is main's, not this PR's.**

Verified: the identical `error: unused import: `super::super::*`` (6×, `perry-codegen/.../node_core/util_buffer.rs`) fails on **main at `f5a8dcf9c`** ([run 32396350265](https://github.com/PerryTS/perry/actions/runs/32396350265)), which contains none of this branch's changes — and it is **intermittent**: the same SHA passed `warnings` in [run 32389924489](https://github.com/PerryTS/perry/actions/runs/32389924489) and failed in the next one, which points at a cache / feature-unification dependency. This PR touches `perry-runtime`, two test files and the workflow; it does not touch `perry-codegen`. Running the exact CI command locally (`RUSTFLAGS='-D warnings' cargo check --workspace --all-targets` with the same excludes) is clean on this branch, so the warning is Linux-specific. Tracking separately.

**What this PR is verified to fix**, on Linux, with the integration-suites step confirmed to have actually executed:
- `tier1_every_ffi_type_against_test_dylib` … ok (322s)
- `function_apply_with_runtime_args_defers_to_a_located_aot_error` … ok (277s)
- `cargo-test` … pass, `lint` … pass

That is the first time both throw-transport canaries have been green together — they were mutually unsatisfiable for as long as CI linked a `panic=unwind` runtime.

PR #8491, opened 2026-08-20
perf(codegen): preserve numeric envelope proofs

### LINKED ISSUE #8485: perf: typed-array element access ~10x Node on BOTH paths (#5525's untyped/typed ratio is closed at 1.05)
## Typed-array element access is ~10x slower than Node on *both* the typed and untyped paths

Found while running the full `benchmarks/suite` (30 rows) on the quiet bench mini.
`bench_typed_array_untyped_access` exists to track #5525 — the gap between *untyped* and
*typed* typed-array element access. **That gap is effectively closed**, but it revealed a
larger absolute one underneath.

Measured on the quiet M1 mini under the sweep bench lock, Perry `6173ff9f6`, Node 26.5.1,
checksums identical (`-821955270` on both, so this is not a correctness difference):

| | Perry | Node |
|---|---:|---:|
| `ta_untyped_access` | 3118 ms | 290 ms |
| `ta_typed_access` | 2970 ms | 289 ms |
| **`ta_untyped_typed_ratio`** (the #5525 metric) | **1.05** | 1.003 |

So the thing #5525 tracked — untyped costing ~6x the typed path — is down to **1.05**. The
per-element kind guard is no longer the dominant cost. What remains is that **both** Perry
paths are ~10.2x Node in absolute terms.

That reframes the work: further effort on static typed-array-kind inference for `any`
parameters (the direction #5525's benchmark header proposes) would buy at most the remaining
5%. The 10x is in the shared element-access path, not the untyped specialisation.

## Wider context from the same run

30-row `benchmarks/suite`, best-of-five interleaved, verdict `CLEAN` (load 1.72 -> 1.89).
Perry wins 22 of 30, geomean **0.4908x**. The losing rows:

| row | P/Node |
|---|---:|
| `bench_typed_array_untyped_access` | **9.43** |
| `bench_numeric_array_numeric` | 4.04 |
| `bench_int_arithmetic` | 3.02 |
| `bench_json_typed_roundtrip` | 2.21 |
| `bench_object_property` | 1.80 |
| `16_matrix_multiply` | 1.53 |
| `17_loop_data_dependent` | 1.29 |
| `bench_json_readonly_indexed` | 1.03 |

`bench_numeric_array_numeric` (4.04x) and `bench_int_arithmetic` (3.02x) look like the same
family — numeric element access and integer arithmetic — and may share a cause with the row
above. Worth checking whether one fix covers all three before treating them separately.

None of these rows appear in the 19-row `gc-handoff` corpus that the four-engine sweep uses,
which is why they have gone unmeasured. On that corpus Perry currently wins 14/19 with a
geomean of 0.7165 — the suite is a materially different and harsher picture, and worth
running alongside it.

## Reproduce

```sh
# compile the suite with a release build, then on the quiet mini:
cd ~/suite-bench && python3 runsuite.py
```

Harness and artifacts: `/Users/amlug/projects/perry/suite-artifacts/` (binaries, sources,
Node-oracle expected output, and `runsuite.py` — lock-guarded, interleaved, checksum-verified).

proggeramlug:
Fixes #8485.

## What changed

- preserve a runtime-derived Number fact for `undefined`-seeded locals when every observable read is dominated by a Number-producing write
- carry numeric plain-array call-site evidence into mixed specialized-ABI plans and validate it once per direct call
- propagate successful descriptor proofs into the specialized body's local type map, allowing numeric element reads to feed native bitwise/add lowering
- keep a boxed generic fallback whenever the numeric-array descriptor does not match

This removes the dynamic `+` and `^` envelope from both `encipherUntyped` and `encipherTyped`. The S/P typed-array loads remain on their existing native path.

## Benchmark

Five runs of `benchmarks/suite/bench_typed_array_untyped_access.ts`, using a fresh release build of `perry`, `perry-runtime-static`, and `perry-stdlib-static`. Instructions retired are the primary signal.

| median (5 runs) | before | after | delta |
|---|---:|---:|---:|
| wall | 6.30 s | 3.52 s | -44.1% |
| instructions retired | 86,740,343,839 | 56,612,344,138 | -34.73% |
| peak RSS | 11,714,560 B | 5,554,176 B | -52.6% |
| untyped / typed | 1.0438 | 1.0017 | improved |

Wall ranges were disjoint despite host contention: 6.30–6.93 s before and 3.49–3.89 s after. The checksum remained `-821955270` in every run.

The numeric/dynamic-heavy sweep rows stayed instruction-neutral:

| median (5 runs) | before | after | instruction delta |
|---|---:|---:|---:|
| `interp` | 0.61 s / 8,996,431,485 inst / 33,030,144 B RSS | 0.67 s / 9,002,446,227 inst / 32,980,992 B RSS | +0.067% |
| `iso_miss` | 0.93 s / 14,788,875,174 inst / 32,817,152 B RSS | 1.15 s / 14,797,545,801 inst / 32,833,536 B RSS | +0.059% |

Their wall ranges overlapped (`interp`: 0.59–0.68 s both; `iso_miss`: 0.91–1.66 s before, 1.06–1.27 s after), so the wall movement is contention noise.

## Validation

- all 19 programs in `sweep-artifacts-0820/sources` byte-exact against expected stdout/stderr
- `cargo test --release -p perry-runtime --lib` (2,605 passed, 4 ignored)
- `cargo test --release -p perry --bin perry` (1,008 passed)
- `cargo test --release -p perry-codegen --lib` (1,121 passed)
- `bash scripts/run_lint_gates.sh` (all 52 gates passed)

## Summary by CodeRabbit

* **Performance**
  * Improved numeric operation performance by preserving number guarantees across local reassignments and eligible array reads.
  * Reduced dynamic arithmetic overhead for native typed-array element access.

* **Bug Fixes**
  * Improved handling of uninitialized numeric locals when all reads follow valid numeric assignments.
  * Added safer guarded dispatch for numeric array parameters, including runtime validation when required.

* **Tests**
  * Added regression coverage for loops, conditional assignments, mixed typed-array arguments, and invalidating writes.

### DIFF
--- a/changelog.d/8491-numeric-envelope.md
+++ b/changelog.d/8491-numeric-envelope.md
@@ -1,0 +1,5 @@
+### Performance
+
+- Preserve Number proofs across numeric local reassignments and guarded plain-
+  array parameter reads, removing dynamic arithmetic around native typed-array
+  element loads.
--- a/crates/perry-codegen/src/codegen/function.rs
+++ b/crates/perry-codegen/src/codegen/function.rs
@@ -363,7 +363,8 @@ fn emit_public_spec_function_trampoline(
                 crate::collectors::SpecParamRep::F64 => {
                     Some(emit_typed_arg_guard(blk, TypedParamRep::F64, arg))
                 }
-                crate::collectors::SpecParamRep::Boxed => None,
+                crate::collectors::SpecParamRep::Boxed
+                | crate::collectors::SpecParamRep::NumberArray => None,
                 // Guarded plans never carry TaPtr: its raw pointer contract is
                 // admitted only by construction at a direct call site.
                 crate::collectors::SpecParamRep::TaPtr { .. } => Some("false".to_string()),
@@ -428,7 +429,8 @@ fn emit_public_spec_function_trampoline(
         let blk = wf.block_mut(fast_idx).unwrap();
         for (arg, rep) in arg_names.iter().zip(plan.reps.iter()) {
             match rep {
-                crate::collectors::SpecParamRep::Boxed => {
+                crate::collectors::SpecParamRep::Boxed
+                | crate::collectors::SpecParamRep::NumberArray => {
                     raw_args.push((DOUBLE, arg.clone()));
                 }
                 crate::collectors::SpecParamRep::I32 => {
@@ -714,7 +716,16 @@ pub(super) fn compile_function(
     // plan selection demoted any reassigned/closure-referenced param to
     // `Boxed` — a stamped param provably holds this rep for the whole body.
     if let Some(plan) = spec_entry {
-        for (p, rep) in f.params.iter().zip(plan.reps.iter()) {
+        for ((p, rep), guard) in f
+            .params
+            .iter()
+            .zip(plan.reps.iter())
+            .zip(plan.guards.iter())
+        {
+            if let Some(guard) = guard {
+                local_types.insert(p.id, guard.proof.clone());
+                continue;
+            }
             match rep {
                 crate::collectors::SpecParamRep::I32 => {
                     local_types.insert(p.id, perry_hir::types::Type::Int32);
@@ -724,7 +735,9 @@ pub(super) fn compile_function(
                         local_types.insert(p.id, perry_hir::types::Type::Named(class.to_string()));
                     }
                 }
-                crate::collectors::SpecParamRep::Boxed | crate::collectors::SpecParamRep::F64 => {}
+                crate::collectors::SpecParamRep::Boxed
+                | crate::collectors::SpecParamRep::NumberArray
+                | crate::collectors::SpecParamRep::F64 => {}
             }
         }
     }
@@ -798,7 +811,11 @@ pub(super) fn compile_function(
                                 spec_ta_kind_class_name(*kind)?.to_string(),
                             )
                         }
-                        (None, crate::collectors::SpecParamRep::Boxed) => return None,
+                        (
+                            None,
+                            crate::collectors::SpecParamRep::Boxed
+                            | crate::collectors::SpecParamRep::NumberArray,
+                        ) => return None,
                     };
                     Some((param.id, proof))
                 })
--- a/crates/perry-codegen/src/codegen/mod.rs
+++ b/crates/perry-codegen/src/codegen/mod.rs
@@ -226,6 +226,7 @@ pub use opts::{
     AppMetadata, CompileOptions, FpContractMode, ImportedClass, NamespaceEntry, NamespaceEntryKind,
 };
 pub(crate) use opts::{CrossModuleCtx, ImportedCtor};
+pub(crate) use param_guard::scalar_descriptor_rep;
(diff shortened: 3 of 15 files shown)


proggeramlug:
CI note: the lint job failure is outside this PR. `raw_handle_debt.py` reports `crates/perry-runtime/src/array/iter_methods.rs` at 6 bare reads versus its ceiling of 2; this PR changes no runtime files, and the violation comes from current `main` via #8482. The branch-local `scripts/run_lint_gates.sh` run passed all 52 gates.

proggeramlug:
Independently validated. Merging.

**Reproduced the headline** and confirmed the win is not lopsided — that was the bar this had to clear, since the earlier attempt on #8485 bought −24.6% instructions almost entirely on the typed path and was correctly reverted. Here the untyped/typed ratio *improves* (1.0438 → 1.0017), so the untyped path genuinely caught up.

**Audited as a soundness change, not a perf change.** A wrong numeric proof means an `fadd` on a NaN-box propagating the payload (#7773's class), so the dataflow was the review, not the benchmark:

- a read of a candidate not dominated by a Number-establishing write **invalidates** it;
- any non-`undefined`, non-Number write **poisons** the candidate even when overwritten before a read — correct, because other consumers of this fact use it to omit pointer rooting and `undefined` is non-pointer while a string or object is not;
- closure-captured mutables are excluded via `boxed_vars`, and closure bodies are deliberately not walked;
- `If` intersects the two branch states; `While` intersects entry with the back-edge state **and re-checks the condition under the back-edge state**, which is the case that would otherwise let a body write escape notice;
- a `LocalSet` in non-root (conditional/short-circuit) position is declined rather than guessed at;
- declared `number` parameters stay untrusted in a generic body — only the specialized entry's raw numeric params seed the flow state. Right call: a declared type is not a proof.

`PERRY_NUMBER_BY_CONSTRUCTION` is registered in `build_cache.rs`, so flipping it invalidates the object cache rather than silently serving objects built under the other setting.

**Validation on this branch:**

| check | result |
|---|---|
| 19-program corpus | 19/19 byte-exact |
| same corpus with `PERRY_NUMBER_BY_CONSTRUCTION=0` | byte-exact |
| `perry-codegen --lib` | 1121 passed |
| `perry-runtime --lib` | 2605 passed |
| `perry --bin perry` | 1008 passed |
| `run_lint_gates.sh` | 52/52 |
| gap suite | 575 run, **0 real regressions** |

**On the gap suite's six reported regressions** — `backoff_options`, `cron_cronjob`, `dayjs_factory_arg`, `moment_methods`, `ratelimiter_memory`, `slugify_options` — all six are phantom. They fail identically on a clean `main` build, and the diff shows the failure is on the **Node** side: `ERR_MODULE_NOT_FOUND` for `dayjs`, `slugify`, `cron`, `moment`, `exponential-backoff`, `ratelimiter`, which were not installed in the worktree. Installing those six packages makes all six pass on both arms.

Worth noting separately: a Node oracle that cannot resolve a module exits non-zero *with output*, so the harness classifies it `parity_fail` (a red gate) rather than `node_fail` (a skip). A missing npm package is therefore indistinguishable from a compiler regression at the gate. That is the same unreachable-`node_fail` shape already known from the gap gate's snapshot diffing.

Issue #8410, opened 2026-08-19
tracking: close the remaining 7 losing rows vs Node (12/19 won, geomean 0.7775x)


proggeramlug:
## Purpose

Tracking issue for closing the remaining gap to Node on the 19-row benchmark corpus. Perry
currently **beats Node on 12 of 19 rows** with a **geomean of 0.7775x**. This enumerates the
seven rows that still lose, each with its own ticket, evidence and owner-ready repro.

Measured on `8bfdac447` (2026-08-19 sweep, quiet M1 mini, verdict `CLEAN`, load 1.42 -> 2.08,
best-of-five interleaved, byte-exact output re-checked on every run).

## Current standing

| | value |
|---|---|
| Perry beats Node | **12/19** |
| geomean Perry/Node | **0.7775x** |
| geomean Perry/scriptc | 0.359x |
| geomean Perry/Porffor | 0.271x |
| peak RSS vs Node | lower on **all 19 rows** |

## The seven remaining rows

| row | P/Node | instr ratio | dominant cost | ticket |
|---|---:|---:|---|---|
| `cycles` | 1.01 | 2.51 | `init_typed_shape_layout` **20.1%** | #8405 |
| `shapes` | 1.06 | 1.25 | `try_read_tracked_gc_header` + `shape_descriptor_by_id` = **100%** | #8406 |
| `asyncpipe` | 1.12 | 1.71 | not CPU-bound — event-loop/timer latency | #8409 |
| `churn` | 1.16 | **0.99** | memory stalls; lowest IPC in corpus (13.35) | #8408 |
| `pipeline` | 1.26 | 1.88 | libm `fmod` 7.3%, plus user-code method dispatch | #8407 |
| `interp` | 1.55 | 1.60 | layout bookkeeping ~20.8% | #8405 |
| `iso_miss` | 2.02 | 1.97 | string concat 17.9% (#8394), layout bookkeeping | #8394, #8405 |

## The structural finding

Splitting all 19 rows by instruction ratio is unambiguous:

- every row Perry **wins** has instructions/Node **<= 1.07**, mostly far below
  (`fib40` 0.22, `churn_read` 0.27, `tree_wide` 0.37)
- every row Perry **loses** has **>= 1.25**, except `churn`

**Perry's codegen quality is not the problem.** Where it emits comparable work it wins
comfortably. It loses precisely where it executes more work. That makes each remaining row a
question of "what work can be deleted", which is why every ticket above names a specific
[...]

PR #8497, opened 2026-08-20
perf(string): fuse accumulator concat chains

proggeramlug:
## Summary

- recognize `s = s + a + b + ...` for proven string accumulators in local, captured, and module-global slots
- lower the accumulator and suffix operands left-to-right into one rooted runtime call
- reuse a unique accumulator's existing capacity, or allocate the complete result exactly once
- preserve the existing dynamic-add fallback when the string proof is incomplete

Refs #8410.

## Profile

The macOS `sample` utility hung before collecting a sample on this host, including when reproduced against `yes`. I used the same symbol-preserving build (`PERRY_KEEP_SYMBOLS=1 PERRY_DEBUG_SYMBOLS=1`, about 17.7k symbols in the generated fixture) with Instruments Time Profiler over a private 10x `iso_miss` fixture instead.

Baseline sampled CPU was 8,422 ms. The largest leaves were `evalNode` (38.90%), `js_string_concat_chain` (14.15%), `lookup` (12.19%), `memmove` (4.19%), arena allocation (3.72%), and the separate accumulator append/storage/rooting envelope (6.80% combined).

Candidate sampled CPU was 7,716 ms (-8.38%). The suffix concat plus separate append envelope becomes one `js_string_append_chain` leaf (15.15%); `evalNode` remained dominant (42.69%) and `lookup` remained essentially unchanged (12.08%).

## Performance

Five shuffled, interleaved repeats on macOS, comparing a locally built `8d1cca158` baseline with this branch. Both arms used the identical `-p perry -p perry-runtime-static -p perry-stdlib-static` release build set, and archive mtimes were verified after each build. Medians are shown.

| `iso_miss` | before | after | delta |
|---|---:|---:|---:|
| wall | 3.36 s | 2.39 s | -28.87% |
| instructions | 12,484,326,208 | 11,497,214,518 | **-7.91%** |
| cycles | 2,501,750,408 | 2,267,980,449 | **-9.34%** |
| peak RSS | 32,817,152 B | 32,866,304 B | +49,152 B (+0.15%) |

The wall ranges overlap widely (1.41-5.00 s before, 1.24-7.07 s after), so the wall median is not treated as a real 28.9% result. Instructions retired are the primary signal. The observed RSS cost is 48 KiB; the implementation does not add a cache or retained reserve and allocates the complete result once, but the measured peak is reported rather than rounded away.

Whole-corpus medians:

| program | wall | instructions | cycles | peak RSS |
|---|---:|---:|---:|---:|
| asyncpipe | -27.78% | -0.13% | -4.31% | +0.04% |
| churn | +0.00% | -0.05% | +2.56% | +0.07% |
| churn_alloc | -27.27% | -0.08% | -2.87% | +0.00% |
| churn_read | +0.00% | -0.13% | -1.21% | +0.28% |
| cycles | +0.00% | +0.10% | +3.90% | +0.00% |
| deeplist | +0.00% | -0.06% | -0.69% | +0.03% |
| fib40 | +1.59% | -0.03% | -0.89% | +0.34% |
| interp | +22.22% | -0.01% | -1.02% | +0.05% |
| **iso_miss** | **-28.87%** | **-7.91%** | **-9.34%** | **+0.15%** |
| pipeline | +0.00% | +0.08% | +0.79% | -0.07% |
| push_cls | +9.09% | -0.02% | -2.75% | +0.07% |
| push_num | -11.11% | -0.15% | -5.42% | -0.06% |
| retain | -22.22% | -0.09% | +1.84% | +0.01% |
| retain1 | -11.11% | -0.05% | -2.04% | +0.00% |
| retain_wide | +13.51% | +0.41% | -0.50% | +0.00% |
| retain_wide1 | -11.11% | -0.20% | -1.85% | +0.00% |
| shapes | +33.33% | -0.03% | -2.24% | +0.15% |
| tree | -22.73% | -0.11% | -4.88% | -0.06% |
| tree_wide | +21.54% | +0.02% | -0.41% | +0.00% |

No non-target row moved more than 0.41% in instructions. All non-target wall ranges overlapped; the large wall/cycle swings track host contention rather than instruction count.

## Validation

- all 19 corpus programs byte-exact against expected output
- `cargo test --release -p perry-runtime --lib`: 2,608 passed, 4 ignored
- `cargo test --release -p perry-codegen --lib`: 1,121 passed
- `cargo test --release -p perry --bin perry`: 1,008 passed
- lint: initial full run passed 50/52; the two branch-specific ratchets were fixed and pass individually, but the final full 52-gate rerun is blocked by the host disk floor
- gap suite: interrupted after 25/575 when the documented skip-build path reached a nested HTTP-extension Cargo build; a complete retry is blocked by the host disk floor

Host status at handoff: 6 GiB free after deleting this worktree's generated `target/` and reducing the reserved external target to about 832 MiB. The task requires stopping below 15 GiB, so this PR remains draft until those two acceptance runs can complete.

### DIFF
--- a/changelog.d/8497-string-append-chain.md
+++ b/changelog.d/8497-string-append-chain.md
@@ -1,0 +1,7 @@
+perf(string): fuse proven `s = s + a + b + ...` accumulator chains into one
+rooted runtime operation (#8497). The runtime now appends every suffix directly
+when a unique accumulator has capacity, or allocates the complete result once,
+instead of first materializing the suffix and then entering a separate append
+envelope. On `iso_miss`, this reduces instructions retired by 7.91% and cycles
+by 9.34% across five shuffled interleaved repeats; no other corpus row moves by
+more than 0.41% in instructions. Observed peak RSS changes by +48 KiB (+0.15%).
--- a/crates/perry-codegen/src/codegen/declared_string_add_tests.rs
+++ b/crates/perry-codegen/src/codegen/declared_string_add_tests.rs
@@ -333,12 +333,12 @@ fn a_chain_whose_second_part_is_proven_still_folds() {
 }
 
 #[test]
-fn a_self_append_chain_retains_the_accumulator_and_fuses_only_the_suffix() {
+fn a_self_append_chain_fuses_the_accumulator_and_suffix() {
     // `s = s + "[" + name + "]"` is the #8394 accumulator shape. Folding
     // all four parts into `js_string_concat_chain` copies the growing `s`
-    // prefix on every iteration. The self-append lowering must instead build
-    // the three-part suffix once and hand it to `js_string_append`, whose
-    // unique-owner path grows the accumulator geometrically.
+    // prefix on every iteration. Building a three-part suffix first still
+    // creates short-lived garbage; the append-chain lowering must hand all
+    // four parts to one helper whose unique-owner path grows geometrically.
     let value = add(
         add(
             add(Expr::LocalGet(1), Expr::String("[".to_string())),
@@ -359,13 +359,13 @@ fn a_self_append_chain_retains_the_accumulator_and_fuses_only_the_suffix() {
     let ir = function_ir(module);
 
     assert!(
-        ir.contains("call i64 @js_string_append_known_heap("),
+        ir.contains("call i64 @js_string_append_chain("),
         "the growing prefix must reach the amortized append path:\n{ir}"
     );
     assert_eq!(
         ir.matches("call i64 @js_string_concat_chain(").count(),
-        1,
-        "only the fixed-size suffix should use the n-way concat fold:\n{ir}"
+        0,
+        "the suffix must not be allocated before it is appended:\n{ir}"
     );
 }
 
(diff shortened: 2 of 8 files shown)


proggeramlug:
Landed on `main` via #8683 (squash `7ad718ab4`).

I read both rooting paths rather than relying on the suites, since this touches in-place string mutation and raw payload pointers held across an allocation. Both are correct: the runtime roots every piece in a `RuntimeHandleScope` before the collecting `string_storage_alloc` and then **reloads** through `with_const_ptr` instead of reusing the pre-collection pointer, and codegen's `with_rooted_group` re-reads each operand after all lowering that can collect. Rooting without the reload is what left a stale argument in #8427 — this does the reload. The `refcount == 1` gate correctly admits only uniquely-owned accumulators, since fresh headers start shared at 0.

Validation: 9 ratchet gates + `cargo fmt --check` pass (including `string_payload_access_inventory.py`), `perry-codegen --lib` 1198/0, `perry-runtime --lib` 2654/0 at `RUST_TEST_THREADS=1`, and all three new tests verified running by name rather than inferred from the total.

Also worth noting your A/B methodology was right — identical `-p perry -p perry-runtime-static -p perry-stdlib-static` set on both arms with archive mtimes verified after each build is exactly what keeps an A/B from being vacuous here. Thanks!

PR #8503, opened 2026-08-21
fix(runtime): port yoga backend to taffy 0.13

### LINKED ISSUE #8438: deps: port perry-runtime yoga backend to taffy 0.13 (alignment enums restructured; Style is no longer Send)
## This cannot merge as-is. Findings from attempting it, so nobody repeats the work.

Bumping the declaration to `taffy = "0.13"` and running `cargo check -p perry-runtime`
yields **29 errors** in `crates/perry-runtime/src/yoga.rs`, in two distinct classes.

### 1. Alignment enums became structs with associated constants — 27 errors, mechanical

taffy 0.13 split each alignment type into a `…Keyword` enum plus a struct carrying keyword +
safety:

```rust
pub struct AlignContent { pub keyword: AlignContentKeyword, pub safety: AlignmentSafety }
impl AlignContent {
    pub const FLEX_START: Self = Self { keyword: AlignContentKeyword::FlexStart, safety: AlignmentSafety::Unsafe };
    pub const CENTER: Self = …
}
```

So `AlignContent::FlexStart` becomes `AlignContent::FLEX_START`, and likewise for
`AlignItems`, `JustifyContent`, etc. Tedious but a pure rename — and the new `safety` field
is a genuine CSS concept (`safe`/`unsafe` alignment) we should decide about rather than
inherit by default.

### 2. `Style` is no longer `Send` — 1 error, and this is the real blocker

```
error[E0277]: `*const ()` cannot be sent between threads safely
   --> crates/perry-runtime/src/yoga.rs:118:20
118 | static YOGA_NODES: Mutex<Option<HashMap<u32, YogaNode>>> = Mutex::new(None);
note: required because it appears within `taffy::style::compact_length::inner::CompactLengthInner`
```

taffy 0.13 packs lengths into a tagged pointer (`CompactLengthInner { tagged_ptr: *const () }`)
which may hold either an f32 or a pointer to a `calc()` expression.

**We already build with `default-features = false` and no `calc` feature**, so that pointer
can never hold a real allocation in our configuration — only tagged numeric values. But
Rust's auto-trait analysis is structural, not value-based, so `Style` is non-`Send`
regardless and the global `Mutex<HashMap<u32, YogaNode>>` will not compile.

Two viable routes, and choosing between them is a design decision rather than a port:

1. **`unsafe impl Send` on a newtype wrapper**, justified by the `calc` feature being off.
   Correct today, but it is a safety assertion that silently becomes wrong the moment
   anyone enables `calc` — it would need a compile-time guard (`#[cfg(feature = "calc")] compile_error!`)
   to stay honest.
2. **Move `YOGA_NODES` to `thread_local!`.** Arguably more correct anyway — yoga layout is a
   single-threaded UI path. But `YogaNode.measure` holds a NaN-boxed JS callback and the map
   is a **registered GC root** (`lock_gc_root_registry(&YOGA_NODES)`), so the root scanner
   has to move with it. That is GC-root surgery on a live path and deserves its own review.

### Recommendation

**Close this PR** and track the upgrade separately:

- `cargo audit` reports **no advisory** against taffy 0.7.7 — this is a routine bump, not a
  security fix, so there is no reason to rush it. (Contrast #8404, the h2 bump for
  RUSTSEC-2026-0258, which I merged immediately.)
- Jumping 0.7 -> 0.13 crosses six breaking releases at once. Stepping through intermediate
  versions would make the alignment-API and `CompactLength` changes reviewable separately
  instead of arriving as one 29-error wall.

Filed as its own issue with these findings.

proggeramlug:
## Summary

Ports the Perry TUI and Yoga layout backends from taffy 0.7 to 0.13 without an unsafe `Send` assertion. Yoga nodes now live in a per-thread registry, matching the single-threaded UI runtime and taffy 0.13 storage model.

## Changes

- update taffy and its lockfile entry to 0.13
- migrate TUI and Yoga alignment values to the new associated constants
- preserve Yoga’s existing unsafe overflow-alignment semantics explicitly
- move Yoga nodes and the GC scanner-registration latch to thread-local storage so every JS worker registers and scans its own callback roots
- add focused alignment and thread-isolation regression tests

## Related issue

Closes #8438

## Test plan

- `cargo check --locked -p perry-runtime`
- `cargo test -p perry-runtime --lib yoga::tests -- --test-threads=1`
- `cargo test -p perry-runtime --lib tui::layout::tests -- --test-threads=1`
- `git diff --check`

- [x] (if CLI / stdlib / runtime API changed) Docs are not required; no public API changed

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated layout handling to the latest Taffy version while preserving existing alignment behavior.
  - Improved reliability of Yoga-based layouts across worker threads.
  - Added safeguards for alignment and overflow behavior.

- **Documentation**
  - Added changelog notes covering the layout backend updates.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2911,12 +2911,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "grid"
-version = "0.15.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "36119f3a540b086b4e436bb2b588cf98a68863470e0e880f4d0842f112a3183a"
-
-[[package]]
 name = "group"
 version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -8826,12 +8820,11 @@ dependencies = [
 
 [[package]]
 name = "taffy"
-version = "0.7.7"
+version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ab4f4d046dd956a47a7e1a2947083d7ac3e6aa3cfaaead36173ceaa5ab11878c"
+checksum = "c034e05f6ee85a12daa63863c2245797715075c70649947aa0da54f3f2ab1d0f"
 dependencies = [
  "arrayvec",
- "grid",
  "serde",
  "slotmap",
 ]
--- a/changelog.d/8503-taffy-013.md
+++ b/changelog.d/8503-taffy-013.md
@@ -1,0 +1,7 @@
+### Fixed
+
+- Port the TUI and Yoga layout backends to taffy 0.13. Yoga nodes now stay in
+  their owning thread, avoiding an unsafe `Send` assertion for taffy's compact
+  style storage, while each JS worker registers a GC scanner for its stored
+  measure callbacks. Existing Yoga overflow alignment remains explicitly
+  `unsafe`, matching the pre-upgrade behavior.
--- a/crates/perry-runtime/Cargo.toml
+++ b/crates/perry-runtime/Cargo.toml
@@ -263,7 +263,7 @@ rand = "0.10"
 regex = { workspace = true, optional = true }
 # Taffy — flexbox / grid layout engine for the perry/tui module
 # (#358 Phase 3). Same crate Bevy and Dioxus use; pure Rust, no FFI.
-taffy = { version = "0.7", default-features = false, features = ["std", "flexbox", "taffy_tree"] }
+taffy = { version = "0.13", default-features = false, features = ["std", "flexbox", "taffy_tree"] }
 fancy-regex = { version = "0.18", optional = true }
 itoa.workspace = true
 ryu.workspace = true
--- a/crates/perry-runtime/src/tui/layout.rs
+++ b/crates/perry-runtime/src/tui/layout.rs
@@ -135,17 +135,17 @@ fn box_style_to_taffy(s: &BoxStyle) -> Style {
         FlexDirection::Column => taffy::FlexDirection::Column,
     };
     let justify = match s.justify_content {
-        JustifyContent::Start => Some(taffy::JustifyContent::Start),
-        JustifyContent::Center => Some(taffy::JustifyContent::Center),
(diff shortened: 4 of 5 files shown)


proggeramlug:
Audited and merging.

**This is more than a dependency bump — it also closes a latent multi-thread GC rooting hole**, and the PR description undersells that half.

`MUTABLE_ROOT_SCANNERS` is **thread-local** (`crates/perry-runtime/src/gc/roots.rs:105`), but on `main` `GC_SCANNER_REGISTERED` is a global `Mutex<bool>`. So the first thread to touch yoga registers `yoga_root_scanner` into *its own* thread-local registry and flips the global latch — and every later thread sees the latch already true and **registers nothing**. That thread's collector then never scans yoga's stored JS measure callbacks, so a callback set by a second worker between `setMeasureFunc` and `calculateLayout` is invisible to the collector that would reclaim it.

Moving both the registry and the latch to `perry_thread_local!` is the correct fix, and the reasoning in the code comment is right about why the latch has to follow the registry. `YOGA_NEXT_ID` correctly stays a global `Mutex`, so handle ids remain unique across the per-thread registries and a stale id cannot alias a live node on another thread.

The port itself is mechanical — associated constants for the alignment enums, `grid` dropped from the lockfile. Good call pinning `AlignmentSafety::Unsafe` on every mapped alignment with an explicit assertion rather than inheriting whatever taffy 0.13 defaults to; that is the kind of behavior an enum-shape migration silently flips.

**Validation on this branch:**

| check | result |
|---|---|
| `gc_runtime_root_holders` gate | OK — 633 holders, 96 scanner-reached, 140 registered scanners |
| `yoga::` tests (incl. `yoga_registry_and_gc_registration_are_per_thread`) | 2 passed |
| `tui::layout` tests | 6 passed |
| `perry-runtime --lib`, `RUST_TEST_THREADS=1` | **2608 passed**, 4 ignored |

One consequence worth a line in the changelog: a yoga node created on one thread is no longer visible from another. That is consistent with the single-threaded UI runtime the handle design assumes, and the new isolation test pins it — but it is a real behavioral narrowing, not just an internal refactor.

Issue #620, opened 2026-05-09
class instance property reassignment ignored: this.method = X dispatched to prototype, not own property


proggeramlug:
## Summary

`obj.method = newFn` writes an own property on the instance, but Perry's class-method dispatch always invokes the prototype/class-vtable method — the own-property override is ignored. Bites any code that uses the "first call sets up the fast path" pattern (SmartRouter in hono, memoized dispatch tables, polymorphic dispatch via assignment).

## Minimal repro

```ts
class Router {
  routes: any[] | undefined = [];
  match(path: string): string {
    console.log("[Router.match] original called");
    const inner = (p: string) => `inner:${p}`;
    this.match = inner;
    return inner(path);
  }
}

const r = new Router();
console.log(r.match("/foo"));
console.log(r.match("/bar"));
```

Expected (Node):
```
[Router.match] original called
inner:/foo
inner:/bar              ← second call goes to `inner`
```

Actual (Perry):
```
[Router.match] original called
inner:/foo
[Router.match] original called    ← second call ALSO hits the original method
inner:/bar
```

## Why it matters

Common JS pattern. Hono's `SmartRouter.match` (`node_modules/hono/dist/router/smart-router/router.js:23`) does exactly this:

```js
match(method, path) {
  if (!this.#routes) throw new Error("Fatal error");
  // ... pick a sub-router, build it ...
  this.match = router.match.bind(router);   // ← rebind for fast path
  this.#routes = void 0;
  return res;
}
```

On the second `app.fetch(...)` call, Perry calls the original `SmartRouter.match` again instead of the bound `router.match`. The body checks `if (!this.#routes) throw new Error("Fatal error")` and throws — surfacing in #603 as silent failure of every call after the first.

Same pattern in jQuery, Underscore, lodash's once/memoize, redux's lazy action creators, and any framework that uses dispatch-table promotion.

## Fix shape

Class method dispatch should consult the instance's own-property dictionary BEFORE falling back to the class vtable. Currently the dispatch path in `crates/perry-codegen/src/lower_call.rs` and the `js_object_call_method` runtime helper short-circuit to the class vtable when the receiver is a known class instance.

This may be invasive — the fast path optimisation that skips the property dictionary is load-bearing for class-method-call performance. Possible mitigations:

- Detect "instance has been mutated to override `method`" via a per-instance flag and disable the fast path for that instance only.
[...]

PR #8505, opened 2026-08-21
perf(codegen): skip canonical-shape own-method scans

### LINKED ISSUE #8406: perf: shapes spends 100% of self time in try_read_tracked_gc_header + shape_descriptor_by_id (1.06x Node)

proggeramlug:
## Summary

Speed up polymorphic instance-method dispatch by using an exact compiler-published `(class_id, ShapeId)` pair to prove that canonical instances have no own-method override. This removes the hot `js_object_get_own_field_or_undef` keys scan from eligible calls and reuses the class ID for the bounded dispatch tower.

Declared or computed fields, dynamic inheritance, post-construction shape changes, descriptor/prototype invalidation, non-instance receivers, and wide towers keep the existing guarded fallback.

## Changes

- add a guarded canonical-shape fast path to dynamic instance dispatch
- reuse the validated class ID in the dispatch tower
- preserve fallback behavior for own method fields and mutated instances
- add end-to-end regression coverage for declared and post-construction overrides

## Related issue

Fixes #8406

## Performance

Quiet A/B measurement for this change:

- `shapes` retired instructions: **-9.14%**
- `shapes` CPU cycles: **-9.93%**
- peak RSS: **unchanged**
- all 19 corpus rows: byte-exact

A fresh alternating 21-run sample on the currently contended host still clears the issue's wall-time criterion:

- Perry median wall: **103.528 ms**
- Node median wall: **145.085 ms**
- ratio: **0.714x Node**

The same local `/usr/bin/time -l` sample reported median retired instructions / peak RSS of **1,149,921,569 / 33,914,880 bytes** for Perry and **1,090,746,015 / 94,044,160 bytes** for Node. The machine was at load average ~70, so the alternating high-resolution wall sample is reported separately from the quiet A/B compute deltas above.

## Test plan

- [x] `cargo test --release -p perry-runtime --lib` — 2,606 passed, 4 ignored
- [x] `cargo test --release -p perry --bin perry` — 1,008 passed
- [x] `cargo test --release -p perry --test issue_8406_dynamic_dispatch_shape_probe`
- [x] `cargo test --release -p perry-codegen --lib tower_route_is_guarded_by_the_class_shape_id`
- [x] all 19 benchmark corpus programs compile and match their Node-oracle stdout byte-for-byte
- [x] `BASE_SHA=upstream/main SKIP_COMPILE_GATES=1 bash scripts/run_lint_gates.sh` — all 50 lint gates passed

## Checklist

- [x] No workspace version bump; no `CLAUDE.md` or `CHANGELOG.md` changes
- [x] Added focused user-visible regression coverage
- [x] Commit follows the repository convention

## Summary by CodeRabbit

* **Performance**
  * Improved polymorphic method dispatch for compatible object shapes.
  * Added guarded optimizations that preserve fallback behavior when shape information is unavailable.

* **Bug Fixes**
  * Corrected dispatch handling for inherited methods, declared function fields, and later property overrides.

* **Tests**
  * Added regression coverage confirming methods resolve correctly across inherited, declared, and mutated properties.
  * Expanded validation of shape-based dispatch safeguards.


proggeramlug:
Audited and merging. The performance claim is fine; the review here is the soundness of the proof, since a wrong "no own override" verdict silently resolves a method to the prototype when an own property should have shadowed it — that is a wrong-answer bug, not a slow one.

**The compile-time half is conservative in the right ways.** `canonical_shape_excludes_own_property` bails on a declared field matching the property (which may intentionally shadow, #620), on any computed `key_expr`, on `extends_expr`/`native_extends`, on an unknown class, and on a cycle in the chain. Good.

**The runtime half rests on `js_method_direct_shape_class`, which is pre-existing** (#8412's exact-pair guard, already sabotage-tested), not new surface introduced here. Its gates — `obj_type == GC_TYPE_OBJECT`, not forwarded, `!descriptors_in_use()`, `!class_prototype_fast_guards_invalidated()`, `class_id != 0` — mean the whole path disarms as soon as anything descriptor-shaped is in play.

**I verified the new path is actually exercised before trusting any probe.** My first two adversarial fixtures passed against Node while emitting *zero* new arms — the counts were identical to `main`, so passing proved nothing about this change. The trigger turned out to be the `this.kids[i].area()` shape: an untyped element read as receiver across a multi-level hierarchy, which is exactly what `shapes.ts` does and what a statically-typed `xs: A[]` parameter does not reproduce.

On `shapes.ts`: **6 `js_method_direct_shape_class` call sites versus 4 on `main`** — the subject is live on the benchmark this claims to speed up.

With a fixture built to that shape (`Group` holding `kids: any[]`, four-level chain plus a fieldless subclass), emitting **2 call sites versus main's 1**, every shadowing path agrees with Node:

| path | expected | result |
|---|---|---|
| post-construction `inst.area = f` | 100 | match |
| computed key `inst["ar"+"ea"] = f` | 200 | match |
| `Object.assign(inst, {area: f})` | 300 | match |
| `Object.defineProperty(inst, "area", …)` | 400 | match |
| assign then `delete` → prototype again | 2 | match |
| megamorphic site seeing all ten shapes | — | match |

Output `12|1002|1014`, byte-identical to Node and to `main`.

**Also ran the two gates this PR skipped.** The test plan reports "all 50 lint gates" because it used `SKIP_COMPILE_GATES=1`; the two compile-tier gates were therefore not covered. Ran them on the branch: `cargo check --workspace --all-targets` and `cargo clippy --workspace` both clean. Worth noting for future PRs — 50 of 52 is a green that omits the two gates most likely to catch a codegen change.

Issue #8507, opened 2026-08-21
GC: defineProperty and Reflect/own-key paths use a stale receiver or value under forced evacuation


proggeramlug:
Two tests in `crates/perry/tests/gc_string_coerce_property_key_rooting_6943.rs` fail under forced evacuation, and they are the only thing keeping main's sweep-tier `cargo-test` red:

```
test define_property_operands_survive_forced_evacuation ... FAILED
  assertion `left == right` failed: a defineProperty path used a stale receiver
  or stored a stale value under forced evacuation                (line 289)
test reflect_and_own_key_paths_survive_forced_evacuation ... FAILED
  assertion `left == right` failed: a Reflect / own-key path used a stale receiver
  or stored a stale value under forced evacuation                (line 540)
test descriptor_reads_survive_forced_evacuation ... ok
```

**Pre-existing, and confirmed distinct from #8495.** I verified these fail identically on a pristine `origin/main` runtime built in a clean worktree, so they predate #8499/#8500. More usefully, they are *not* the same defect:

- #8500 converted 22 receiver save/restore pairs that wrote the `IMPLICIT_THIS` cell directly. These two tests still fail with that fix in.
- A follow-up converting the remaining 4 two-line variants of the same shape: still fail.
- Converting **all 135** `js_implicit_this_set(...)`-call save/restore pairs in `perry-runtime`/`perry-stdlib` to the rooted form: still fail, unchanged. (Not proposed as a PR — it demonstrably fixes nothing and I reverted it rather than ship unvalidated churn.)

So the receiver-cell save/restore population is exhausted as an explanation. The assertion text points at either a stale receiver *or* a stale stored value on the `defineProperty` and `Reflect`/own-key paths specifically, and `descriptor_reads_survive_forced_evacuation` passing narrows it further — plain descriptor *reads* are fine, so the defect is on the write/enumerate side.

**Suggested starting point**, based on what worked for #8495: run the failing case under `PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800`, which faults at the stale use and names the retiring minor rather than surfacing cycles later. Two things that wasted a lot of time on #8495 and will mislead here too: `===` reports `true` for a moved-away pointer because equality resolves forwarding while ordinary reads do not, and reading the object *before* the collection masks the failure entirely.

[...]

PR #8519, opened 2026-08-21
fix(gc): complete IMPLICIT_THIS receiver rooting (two-line save forms)

proggeramlug:
**GC: complete the `IMPLICIT_THIS` receiver-rooting conversion started in #8500.**

#8500 converted 22 receiver save/restore pairs that manipulated the `IMPLICIT_THIS` cell directly, rooting the displaced receiver so an evacuating collection inside the callee cannot leave the restore publishing a pre-move address. Four sites of the identical shape were missed because the save spans two lines (`let prev =` on one line, the `IMPLICIT_THIS.with(|c| c.replace(…))` on the next) and so did not match the pattern used to find the others.

No behavioural test moves — these four are latent rather than currently reachable by a failing case — but leaving four instances of a shape that is known to produce a stale `this` is the kind of residue that costs a future session a week. The conversion is now exhaustive: no unrooted direct-access save/restore pair remains in `perry-runtime`.

Does **not** address #8507 (`defineProperty` / `Reflect` own-key paths under forced evacuation), which is a distinct defect — verified by re-running that suite with these four converted, and again with all 135 `js_implicit_this_set(...)`-call pairs converted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runtime stability during certain property accessor, inherited accessor, static-property, and regular expression method calls.
  * Prevented receiver state from becoming invalid when garbage collection occurs during these operations.

* **Documentation**
  * Added a changelog entry documenting the runtime stability improvements.


proggeramlug:
Audited and merging. Two changes pushed onto the branch first — see below.

**Verified the exhaustiveness claim independently rather than taking it on trust**, since "the conversion is now exhaustive" is the load-bearing sentence here. Scanning every `IMPLICIT_THIS.with(|c| c.replace(...))` site in `perry-runtime` and checking whether its result is rooted within the preceding lines: **29 rooted, 0 unrooted.** The claim holds.

The fix itself is the right shape — each site roots the displaced receiver in a `RuntimeHandleScope` and restores from `get_nanbox_u64()`, so an evacuating collection inside the callee cannot leave the restore publishing a pre-move address. The scope outlives the read at every site.

Agreed on landing this without a behavioural test. These four are the identical shape to the 22 that #8500 converted, and leaving a known-stale-`this` pattern in the tree because no current case reaches it is how a future session loses a week to it.

**Two things I fixed on the branch:**

1. **The version bump is reverted.** The PR moved `CLAUDE.md`, `Cargo.toml`, and 76 `Cargo.lock` version strings to 0.5.1515; the standing instruction is no version bump for now, with the maintainer bumping at merge time. All three restored to `main`'s 0.5.1514.
2. **The changelog fragment was named `8508-…` on PR #8519.** Fragments are PR-keyed precisely so in-flight PRs cannot collide on a filename. Renamed to `8519-implicit-this-two-line-forms.md`.

**One finding worth a follow-up, beyond this PR's scope.** These four sites were missed because the save spans two lines and so did not match the pattern used to find the other 22 — and there is no lint gate covering this shape at all (`grep -l IMPLICIT_THIS scripts/` is empty). The tree is exhaustive today, but nothing stops a fifth two-line site landing tomorrow, and **the failure direction is green**: an unrooted save produces no error, just an occasionally stale `this`. This is the same under-counting-scanner class that #8502 just addressed for `StringHeader` payload access. A ratchet over `c.replace(` sites with a rooted-within-N-lines predicate would close it.

**Validation on the branch (post-revert):**

| check | result |
|---|---|
| `perry-runtime --lib`, `RUST_TEST_THREADS=1` | **2606 passed**, 4 ignored |
| `gc_runtime_root_holders` gate | OK — 633 holders, 140 registered scanners |
| every `c.replace(` site rooted | 29/29 |

Confirming for the record that this PR does **not** revert #8503's taffy 0.13 upgrade — the branch predates that merge, but against its own merge base it touches taffy zero times.

PR #8520, opened 2026-08-21
perf(codegen): skip canonical-shape own-method scans

### LINKED ISSUE #8406: perf: shapes spends 100% of self time in try_read_tracked_gc_header + shape_descriptor_by_id (1.06x Node)

proggeramlug:
## Summary

Use compiler-published `(class_id, ShapeId)` pairs to prove that canonical instances cannot have post-construction own-method overrides. Eligible polymorphic dispatch sites now bypass the runtime keys-array scan and reuse the class id from the single header probe.

## Changes

- add a bounded canonical-shape probe before dynamic own-method lookup
- preserve guarded fallback behavior for declared/computed fields, dynamic parent chains, mutated shapes, non-instance receivers, and wide dispatch towers
- add integration coverage for declared function fields and post-construction method overrides
- update the existing IR routing assertion for the additional hoisted shape load

## Related issue

Closes #8406

## Performance

Five interleaved whole-corpus runs on the issue host:

| Metric | Baseline | Candidate | Delta |
|---|---:|---:|---:|
| Wall | 0.110 s | 0.090 s | -18.2% |
| Instructions | 1,264,885,553 | 1,149,314,946 | **-9.14%** |
| Cycles | 290,262,830 | 261,445,126 | -9.93% |
| Peak RSS | 33,718,272 | 33,783,808 | +64 KiB / +0.19% |

Wall ranges overlapped on the contended host. Interleaved candidate and Node medians were 90 ms and 110 ms respectively, also with overlapping ranges. No other corpus row moved more than 1% in retired instructions; maximum non-target drift was 0.48%. RSS remained approximately 0.36x Node.

## Test plan

- [x] all 19 corpus programs compile and match the Node oracle byte-for-byte
- [x] `cargo test --release -p perry-runtime --lib` (2,606 passed, 4 ignored)
- [x] `cargo test --release -p perry-codegen --lib` (1,113 passed)
- [x] `cargo test --release -p perry --bin perry` (1,008 passed)
- [x] `bash scripts/run_lint_gates.sh` (all 52 gates, including address-class and GC root-holder ratchets)
- [x] focused #8406 integration and IR-routing tests rerun after rebasing onto current `main`
- [ ] full excluded-platform workspace suite (scoped issue commands above passed)
- [x] Added an integration regression test
- [ ] Documentation update (no API or user-facing behavior change)
- [ ] Platform UI backend build (not touched)

## Screenshots / output

N/A

## Checklist

## Summary by CodeRabbit

* **Performance**
  * Improved dynamic method dispatch for eligible object shapes, reducing unnecessary property checks.
  * Reused receiver information during dispatch for faster method selection.

* **Bug Fixes**
  * Preserved reliable fallback behavior for mutated shapes, inherited fields, and complex class hierarchies.
  * Improved handling of method overrides and function-valued fields.

* **Tests**
  * Added coverage for inherited, declared, and later-mutated method overrides.


proggeramlug:
Closing as superseded — this is already on `main`.

#8520 and #8505 point at the same head commit `5dbe8c18a` on the same branch `fix/8406-shapes-hot-path`. #8505 merged as `84adb7fdf`, and #8406 closed with it.

Verified nothing is lost by closing this:

- `git diff --name-only origin/main pr8520` → **0 files**. Not "the four PR files match" — the entire tree is identical, so there is no stray content outside the listed file set either.
- Each listed file confirmed individually identical to `main`: the changelog fragment, `dynamic_dispatch.rs` (both references to `canonical_shape_excludes_own_property` present), `proven_this_routing_tests.rs`, and `issue_8406_dynamic_dispatch_shape_probe.rs`.
- The only open review is CodeRabbit's auto-summary; no human review asked for a change.

The one real difference between the two PRs was the test plan, not the code: #8505 reported "all 50 lint gates" because it ran with `SKIP_COMPILE_GATES=1`, while this one reports all 52. That gap was covered before the merge — I ran `cargo check --workspace --all-targets` and `cargo clippy --workspace` on the branch myself, both clean — so the stricter test plan here confirms the merge rather than adding anything to it.

The extra corpus data in this description is useful corroboration and worth keeping in the record: no other row moved more than 1% in retired instructions, maximum non-target drift 0.48%, peak RSS +64 KiB (+0.19%).

Merge audit, including the adversarial shadowing fixtures and the subject-liveness check: https://github.com/PerryTS/perry/pull/8505#issuecomment-5365643088

PR #8521, opened 2026-08-21
fix(async_hooks): honor trackPromises selection

proggeramlug:
## Summary

- validate the Node 26 `trackPromises` option and its interaction with `promiseResolve`
- suppress Promise allocation and lifecycle callbacks for hooks that opt out, including when mixed with tracking hooks
- reload Promise pointers from GC roots after async-hook callbacks before V8 hook dispatch
- extend #6764 regression coverage for default, enabled, disabled, mixed, and invalid option shapes

This is a focused increment toward #6764; the remaining provider gaps stay tracked there.

## Testing

- `node test-parity/node-suite/async_hooks/hooks/track-promises.ts`
- `cargo check --release -p perry-runtime`
- `PERRY_SKIP_BUILD=1 PERRY_BIN=$PWD/target/release/perry PERRY_RUNTIME_DIR=$PWD/target/release PERRY_NO_AUTO_OPTIMIZE=1 ./run_parity_tests.sh --suite node-suite --module async_hooks --filter track-promises` — 2/2 pass

No version bump.

### DIFF
--- a/changelog.d/8521-async-hooks-track-promises.md
+++ b/changelog.d/8521-async-hooks-track-promises.md
@@ -4,4 +4,4 @@
 invalid combination with `promiseResolve`. Hooks that set `trackPromises` to
 `false` no longer allocate or receive Promise lifecycle events, including
 when they are enabled beside hooks that continue tracking Promises. Advances
-#6764.
+issue #6764.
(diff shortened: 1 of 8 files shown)


proggeramlug:

| suite | result |
|---|---|

Issue #8508, opened 2026-08-21
runtime/AOT: support OpenCode's WebAssembly modules and packaged .wasm assets


proggeramlug:
Parent: #8506

## Problem

The complete OpenCode graph uses WebAssembly in several real feature paths:

- `@silvia-odwyer/photon-node` for image processing, including file-style WASM imports
- `web-tree-sitter` plus Bash/PowerShell and downloadable language grammar modules
- `@pierre/diffs` worker/WASM assets

Perry has a WebAssembly host and previously completed #5234/#6558, but the full-source target needs the complete combination of runtime APIs, wasm-bindgen glue, dynamic grammar loading, and output asset packaging. A Photon-only native replacement would not cover tree-sitter's arbitrary grammar modules.

## Scope

- `WebAssembly.Module`, `Instance`, `Memory`, `Table`, globals, imports, and exports
- sync and async compile/instantiate shapes used by these packages
- wasm-bindgen glue and its JS↔WASM import cycle
- `import ... from "./x.wasm"` and `with { type: "file" }`
- embedding/copying WASM assets into the native output with stable runtime paths
- actionable compile errors for unsupported WASM forms; never silently defer to a JS engine

## Acceptance

- [ ] A Perry-compiled Photon smoke test loads and executes its shipped WASM
- [ ] a Perry-compiled `web-tree-sitter` smoke test loads a grammar and parses source
- [ ] OpenCode's image, shell parser, and diff paths work from the #8506 full-source build
- [ ] the output runs without V8 or another JavaScript runtime

proggeramlug:
Closing as superseded by #10102, which carries the v1.18.30-specific WASM inventory (web-tree-sitter main module + grammar side modules, photon wasm-bindgen, asset packaging) and the acceptance criteria. Parent tracker: #10107.

PR #8522, opened 2026-08-21
runtime: add bun:ffi JSCallback support (#6562)

proggeramlug:
## Summary

Implements the remaining native-to-JavaScript callback primitive for `bun:ffi`: `JSCallback` now exposes same-thread C-ABI trampolines with rooted closures, and `FFIType.function` works in `dlopen` signatures.

## Changes

- Add a fixed, non-reused callback trampoline pool for macOS/Linux on arm64 and x86-64, preserving the integer and floating-point ABI register classes.
- Root callback closures until `close()`, poison closed pointers, reject `threadsafe: true`, and convert uncaught callback exceptions to the declared ABI zero value instead of unwinding through C.
- Support Bun's `new JSCallback(...)` named-import and namespace-import forms, including the `{ ptr, threadsafe, close }` wrapper shape.
- Marshal `FFIType.function` as a pointer in arguments and returns.
- Add native-fixture coverage for mixed numeric arguments, pointers, C strings, BigInts, returned function pointers, GC-visible closures, closing, and exception behavior.
- Update the API manifest, generated API docs, and changelog fragment. No version files were changed.

## Related issue

Refs #6562 — completes the `JSCallback` / `FFIType.function` stage. The separately declared `CFunction`, `linkSymbols`, `viewSource`, and `read` exports remain out of scope and continue to report `ERR_NOT_IMPLEMENTED`.

## Test plan

- [ ] `cargo build --release` clean (not run; platform-specific builds below were used)
- [ ] `cargo test --workspace --exclude perry-ui-ios --exclude perry-ui-tvos --exclude perry-ui-watchos --exclude perry-ui-gtk4 --exclude perry-ui-android --exclude perry-ui-windows` passes (project guidance uses affected-crate and scoped tests instead)
- [x] `cargo test -p perry-runtime bun_ffi --lib`
- [x] `cargo test -p perry-api-manifest`
- [x] `cargo check -p perry-hir`
- [x] `PERRY_NO_AUTO_OPTIMIZE=1 PERRY_RUNTIME_DIR="$PWD/target/debug" cargo test -p perry --test bun_ffi_stage1 tier1_every_ffi_type_against_test_dylib -- --nocapture`
- [x] `PERRY_NO_AUTO_OPTIMIZE=1 PERRY_RUNTIME_DIR="$PWD/target/debug" cargo test -p perry --test bun_ffi_stage1 tier1b_error_surfaces -- --nocapture`
[...]

PR #8524, opened 2026-08-21
runtime: instantiate compiled WebAssembly modules synchronously

proggeramlug:
## Summary

- route `new WebAssembly.Instance(module, imports)` through the existing wasmi host
- preserve the constructor receiver/prototype while exposing live exports
- cover the wasm-bindgen Node loader shape using WASM bytes packaged via `with { type: "file" }`
- keep synchronous TypeError/LinkError behavior and the standard `Instance.length === 1`

## Scope

This is a scoped step toward #8508. It unblocks the Module → Instance path used by Photon-style Node glue for Perry's existing numeric/function-import subset. Tree-sitter's memory/table/global imports and streaming/reference-value support remain follow-up work, so this PR intentionally does not close the umbrella issue.

## Testing

- `cargo check -p perry-runtime --features wasm-host`
- `cargo test -p perry-wasm-host`
- `cargo test -p perry --test issue_5234_wasm_esm_import -- --nocapture`

Refs #8508

### DIFF
--- a/changelog.d/8508-wasm-instance-constructor.md
+++ b/changelog.d/8508-wasm-instance-constructor.md
@@ -1,0 +1,6 @@
+### runtime: instantiate compiled WebAssembly modules synchronously
+
+`new WebAssembly.Instance(module, imports)` now routes through Perry's wasmi
+host and returns live exports. This supports the synchronous, file-packaged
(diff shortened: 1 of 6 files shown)


proggeramlug:

| suite | result |
|---|---|

PR #8526, opened 2026-08-21
fix(gc): root defineProperty operands across exotic probes (#8507)

### LINKED ISSUE #8507: GC: defineProperty and Reflect/own-key paths use a stale receiver or value under forced evacuation

proggeramlug:
Closes #8507.

`Reflect.defineProperty` and `Object.defineProperty` called individually-rooted exotic helpers, then reused their own raw receiver/key/descriptor copies when a helper returned `NotTypedArray` or `None`. A key coercion inside the helper could evacuate those operands, leaving the caller to continue with from-space addresses. TypedArray and array-length descriptor processing also retained raw descriptor/array pointers across field-name allocations and user coercions.

This change keeps caller-owned handles across the full define-property sequence, re-reads operands between exotic probes, and roots/re-reads TypedArray and array-length descriptor operands at every allocation boundary.

No version bump or changelog fragment is included.

Validation:
- `PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800 RUST_TEST_THREADS=1 cargo test --release -p perry --test gc_string_coerce_property_key_rooting_6943 -- --nocapture` (3/3)
- `cargo test -p perry-runtime rooted_define_property -- --nocapture` (3/3)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when defining object, array, and TypedArray properties during garbage collection.
  * Prevented property descriptors, keys, and target objects from becoming stale while operations allocate memory or coerce values.
  * Preserved existing validation, length-restriction, attribute, and element-write behavior.

### DIFF
--- a/crates/perry-runtime/src/object/array_object_ops.rs
+++ b/crates/perry-runtime/src/object/array_object_ops.rs
@@ -165,28 +165,33 @@ pub(crate) unsafe fn array_set_length_from_descriptor(
     obj: *mut ObjectHeader,
     descriptor_value: f64,
 ) -> bool {
-    let desc_ptr = extract_obj_ptr(descriptor_value);
-    if desc_ptr.is_null() {
+    // #8507: every descriptor-field name below is allocated in the GC heap,
+    // and the value's ToNumber hooks may run user JS. Keep both the array and
+    // descriptor rooted for the whole algorithm; helper-local roots cannot
+    // refresh raw copies held by this caller between field probes.
+    let scope = crate::gc::RuntimeHandleScope::new();
+    let obj_handle = scope.root_raw_mut_ptr(obj);
+    let desc_handle = scope.root_nanbox_f64(descriptor_value);
+    let current_desc = || extract_obj_ptr(desc_handle.get_nanbox_f64());
+    if current_desc().is_null() {
         return true;
     }
     // A customized `length` (e.g. writable:false) gates the raw numeric
     // fast paths — see OBJ_FLAG_ARRAY_DESCRIPTORS in define_array_property.
     // Set here too so the `Reflect.defineProperty` entry point is covered.
     {
+        let obj = obj_handle.get_raw_mut_ptr::<ObjectHeader>();
         let gc = gc_header_for(obj);
         (*gc)._reserved |= crate::gc::OBJ_FLAG_ARRAY_DESCRIPTORS;
     }
-    // #7548: `obj` may be a pre-grow forwarding stub. `old_len` below drives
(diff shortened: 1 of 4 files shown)


proggeramlug:

| suite | result |
|---|---|

PR #6294, opened 2026-07-11
fix(thread): owner-tagged cross-thread dispatch (#6185 Tier 2)

proggeramlug:
## What

Tier 2 of #6185 — **owner-tagged runtime dispatch**. This is the piece the issue was actually tracking a decision on, and it closes the remaining P0s. With this + Tier 1 (#6276) + the serializer/leak fixes (#6188/#6212), every bullet of the issue's "Proposed fix" is done and **#6185 can close**.

Tier 1 removed the compile-visible routes into the unsound drains. Tier 2 makes the drains themselves sound — so the guarantee no longer rests on an AST walk being able to see through every named callee.

## The problem

`PENDING_THREAD_RESULTS` (`thread.rs`) and `TIMER_QUEUE` / `CALLBACK_TIMERS` / `INTERVAL_TIMERS` (`timer.rs`) hold raw pointers — `*mut Promise`, closure handles, NaN-boxed args — into *some* thread's arena. Their drains were unconditional: whichever thread ran the await loop or microtask pump drained everything. The `unsafe impl Send` on those entries carried an "only accessed from the pump thread" comment that nothing enforced.

## Why not a `ThreadId` tag — and how that blocker is resolved

The obvious fix (tag with the enqueuing `ThreadId`, skip foreign entries) is what got this deferred: on Android the compiled TypeScript — and therefore every `setTimeout`/`setInterval` — runs on the `perry-native` thread, while the timer pump fires from the **UI thread** via `nativePumpTick`. A blanket owner-skip leaves every Android timer un-fired forever.

The insight is that the unit that matters isn't the OS thread, it's the **heap** — the arena the queued pointers live in. Call it an *agent* (new `crates/perry-runtime/src/agent.rs`):

- The process starts with one agent, `PRIMARY_AGENT`. Every thread running JS on the primary heap (main thread; Android's `perry-native` thread) belongs to it.
- Each `perry/thread` worker claims its **own** agent at thread entry — it has its own arena and GC.
- A thread with **no agent of its own is not a heap owner** — it is a *pump* acting for the primary agent. So `current_agent()` falls back to `PRIMARY_AGENT` rather than inventing an id.

Android's UI thread is exactly that last case, so `nativePumpTick` keeps delivering `perry-native`'s timers **unchanged**. The blocker is resolved structurally, not special-cased. A unit test pins this invariant explicitly so a future refactor to `ThreadId` can't silently break Android.

## What's tagged

[...]

PR #8527, opened 2026-08-21
feat(worker_threads): compile global Web Worker modules

proggeramlug:
## Summary

Compile statically resolvable Web Worker module URLs into native Perry worker entry functions and expose the browser/Bun worker surface through the existing in-process worker runtime. Worker source is never evaluated at runtime.

## Changes

- Lower unshadowed global `Worker` and `globalThis.Worker` construction to Perry's AOT worker representation, including `new URL("./worker.ts", import.meta.url)` discovery.
- Link the worker stdlib for import-free global workers and install `self`, `postMessage`, EventTarget methods, property handlers, and `close()` in the worker global scope.
- Add compiled-entry `reload()`, Worker-side property event dispatch, GC roots for Worker handles and callbacks, and documentation/changelog coverage.
- Add an end-to-end test covering module discovery, environment options, reload, structured RPC messages, property and EventTarget handlers, close, and termination.

## Related issue

Closes #8509

## Test plan

- [ ] `cargo build --release` clean (not run as a standalone full build)
- [ ] `cargo test --workspace --exclude perry-ui-ios --exclude perry-ui-tvos --exclude perry-ui-watchos --exclude perry-ui-gtk4 --exclude perry-ui-android --exclude perry-ui-windows` passes (full workspace not run)
- [x] `cargo check -p perry-stdlib -p perry-codegen -p perry`
- [x] `cargo test -p perry-hir --lib dynamic_import::tests` (47 passed)
- [x] `cargo test -p perry --test issue_worker_threads_cross_thread_delivery -- --nocapture` (3 passed)
- [x] Focused Web Worker test rerun after rebasing onto latest `main` (1 passed)
- [x] `cargo fmt --all -- --check`
- [x] `./scripts/check_file_size.sh`
- [x] `cargo clippy -p perry-hir -p perry-stdlib -p perry-codegen -p perry --lib --bins --no-deps` completed (repository warning backlog only)
- [x] (if user-facing) Added or updated a test under `test-files/` or a `#[test]` in the affected crate
- [x] (if CLI / stdlib / runtime API changed) Updated `docs/src/`
- [ ] (if touching a platform UI backend) Built `-p perry-ui-<backend>` locally on that platform (not applicable)

Strict Clippy with `-D warnings` remains blocked by pre-existing unrelated lints in `perry-runtime/build.rs` and `perry-diagnostics`.

## Screenshots / output

```text
ready 1 web-env
ready 2 web-env
reply 8509 8509 true
done
```

## Checklist

[...]

PR #8528, opened 2026-08-21
fix(runtime): make the path-module registry per-heap so one process can host several apps

proggeramlug:
## What

Removes `ERR_PERRY_PATH_MODULE_THREAD` — the fail-closed guard that stopped a
single process from hosting more than one Perry application that uses runtime
path modules.

`PathModuleEntry::exports` holds NaN-boxed pointers into the arena of the thread
that produced them, and both the arena and the mutable-root scanners are
thread-local. The registry was process-global, so a single
`OnceLock<ThreadId>` had to refuse every later thread to stop one heap's pointer
reaching another heap's collector. That is the "single-JS-thread discipline"
half of the choice #6185 framed; this commits to the other half, real per-heap
isolation, for this one table.

## The table conflated two different facts

| | keyed | why |
|---|---|---|
| `path -> init_addr` | process-wide | a code pointer. Codegen registers each canonical path's `<prefix>__init` **once**, on whichever thread runs module init first. No collector has any interest in it. |
| `path -> exports` / `status` | per-heap | arena pointers, and the arena is itself a `thread_local!` |

Splitting them is the whole fix. My first attempt made the *whole* table
per-heap, which left a second heap with no initializer to run: it silently
resolved the module to an empty `module.exports`. A heap that has never required
a path now adopts the program-wide initializer (`adopt_registered_init`) instead
of inventing an empty entry.

## Thread-keyed, not agent-keyed

Deliberately, despite #6294 establishing agent tagging for the cross-thread
queues. `CURRENT_AGENT` is itself a `thread_local!` that defaults to
`PRIMARY_AGENT`, and an embedder's app thread is a plain `std::thread::spawn`,
not a `perry/thread` worker — it never calls `enter_worker_agent()`. Agent-keying
would therefore hand every app in a host one shared table: the original bug,
minus the guard that used to make it loud. The rationale is recorded at the
declaration so a future reader meets it there.

The GC scanner loses its owner gate for the same reason — the collector running
it now owns every pointer in the table by construction, and a gate would leave
every heap but the first unscanned. It uses `try_with`, because an embedder's app
thread tears down its heap when a deployment stops and a scanner running during
that teardown would panic on the destroyed thread-local.

## Scope

`js_register_path_init` is emitted only for `nextjs_path_init_modules` — Next.js
`.next/server/**` Deferred modules, for turbopack/webpack chunk loading. Ordinary
Perry programs never reach this table (`nm` on a multi-module CJS fixture shows
zero references to `js_require_path_module`), so the blast radius is confined to
Next-style deployments.

## Testing

- full `perry-runtime` suite: **2609 passed, 0 failed**
- `gc::`: **952 passed, 0 failed**, including
  `full_cycle_path_module_root_store_after_root_scan_preserves_new_value`
- production (non-`cfg(test)`) `cargo check`: clean — checked separately, because
  `per_test_global!` expands to a *different arm* under `cfg(test)` and a green
  `cargo test` is not evidence that the shipped static compiles
- `cargo fmt --check`, `scripts/check_file_size.sh`,
  `scripts/gc_runtime_root_holders.py`: all clean

Two new tests: per-heap isolation, and two heaps initializing the **same** path
through the adoption path. The second deliberately seeds only the process-wide
address and never calls `register_init` per heap — an earlier version did, and so
could not detect the missing-initializer defect above.

## Not proven here

That a second Next application loads in one process. This table is only
reachable through Next chunk loading, so the end-to-end proof needs a host
running two Next apps; that is a separate change on the host side. This PR is
the runtime precondition, verified at the registry layer.

## Follow-up worth filing

`scripts/gc_runtime_root_holders.py` scans **zero** holders from
`path_registry.rs` — the heap pointer sits four levels deep in a custom type, so
the type regex never matches. The holder *is* correctly scanned by a registered
scanner (`gc/mod.rs`), so there is no live bug, but deleting that registration
today would not turn the gate red.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved module loading across separate runtime heaps and threads.
  * Prevented module exports from leaking between isolated runtime contexts.
  * Enabled independent initialization of the same module path in different heaps.
  * Improved cleanup and root scanning when thread-local runtime state is unavailable.
  * Improved reliability when multiple application libraries use identical module paths.
* **Documentation**
  * Added release documentation describing the updated module registry behavior and scope.

### DIFF
--- a/changelog.d/8528-module-path-registry-per-heap.md
+++ b/changelog.d/8528-module-path-registry-per-heap.md
@@ -18,6 +18,15 @@
   entry; making the whole table per-heap instead left later heaps silently
   resolving modules to an empty `module.exports`.
 
+  Both the export table and the initializer-address map are per-heap. A
+  process-wide address map was tried and measured wrong: a host that loads
+  several application libraries gets its own copy of each module per library,
+  at its own code address, under the SAME baked canonical path, so the map saw
+  "same path, different address" and refused the second library as a duplicate.
+  Two Next.js applications in one Coop process produced 115 such rejections,
+  with the second application's modules never initializing (200 for the first,
+  500 for the second).
+
   Keyed by thread rather than by `AgentId`, deliberately: `CURRENT_AGENT` is
   itself a `thread_local!` defaulting to `PRIMARY_AGENT`, and an embedder's app
   thread is a plain `std::thread::spawn` that never calls `enter_worker_agent()`,
--- a/changelog.d/8528-module-path-registry-per-heap.md
+++ b/changelog.d/8528-module-path-registry-per-heap.md
@@ -1,0 +1,31 @@
+### Fixed
+
+- **Runtime: the path-module registry is now per-heap, so one process can host
+  several Perry applications.** `PathModuleEntry::exports` holds NaN-boxed
+  pointers into the arena of the thread that produced them, and both the arena
+  and the mutable-root scanners are thread-local. The registry was
+  process-global, so a single `OnceLock<ThreadId>` had to fail closed for every
+  later thread (`ERR_PERRY_PATH_MODULE_THREAD`) to stop one heap's pointer
+  reaching another heap's collector — which meant a host could load at most one
+  application that uses runtime path modules.
+
+  The table conflated two facts with different lifetimes. A canonical path's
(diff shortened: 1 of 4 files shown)


proggeramlug:

| suite | result |
|---|---|

PR #8525, opened 2026-08-21
feat(sqlite): add bun:sqlite compatibility

proggeramlug:
## Summary

Adds a native `bun:sqlite` compatibility facade over the same rusqlite engine used by Perry node:sqlite support, unblocking the OpenCode SQLite adapter.

## Changes

- Resolve `bun:sqlite` as a native module and lower `Database` construction without colliding with better-sqlite3.
- Implement Bun database and statement methods used by OpenCode, including positional and named binding, blobs, safe integers, transactions, serialization, extension loading, properties, close, and finalize behavior.
- Preserve dynamic native-handle dispatch through Effect flows and repair the Statement.values versus Array.values lowering collision.
- Regenerate API docs and add Bun/Perry parity coverage.

## Related issue

Fixes #8510

## Test plan

- [x] `cargo check -p perry-hir -p perry-codegen -p perry-stdlib --features perry-stdlib/database-sqlite`
- [x] `cargo test -p perry-hir test_bun_sqlite_statement_values_uses_native_dispatch`
- [x] `cargo test -p perry-codegen --test manifest_consistency`
- [x] `cargo test -p perry-api-manifest`
- [x] `cargo build -p perry -p perry-runtime-static -p perry-stdlib-static`
- [x] `./scripts/pre-tag-check.sh --quick`
- [x] Compiled and ran `test_gap_bun_sqlite_8510.ts`; Perry output matches Bun line-for-line.

## Checklist

- [x] No workspace version bump
- [x] No CLAUDE.md or CHANGELOG.md edits
- [x] User-facing compatibility test and generated API docs included

## Summary by CodeRabbit

* **New Features**
  * Added native `bun:sqlite` support with database and prepared-statement APIs.
  * Supports queries, parameter binding, row and value retrieval, BLOBs, safe integers, transactions, serialization, extensions, and lifecycle management.
  * Added read-only, read-write, and database creation options.

* **Documentation**
  * Added `bun:sqlite` to API references and supported package documentation.

* **Tests**
  * Added coverage for queries, transactions, rollback, serialization, BigInt handling, and closed-database errors.

PR #8532, opened 2026-08-21
runtime: add native @parcel/watcher compatibility facade

### LINKED ISSUE #8513: runtime: provide a native @parcel/watcher compatibility facade for full OpenCode
Parent: #8506
Related: #8523 — Node-API host tracker. The general "run the real `.node` addon" question lives there; this issue is deliberately a facade (reasons below).

## Problem

Full OpenCode includes its filesystem watcher through `@parcel/watcher`, which selects a platform native addon at runtime. Compiling the JS wrapper does not remove the native-addon boundary, and a Node-API host solely for this package would be a much larger project (now tracked separately in #8523).

### What OpenCode actually calls (pinned revision `fc80874f`, `packages/core/src/filesystem/watcher.ts`)

OpenCode never imports the `@parcel/watcher` root. It does

```ts
const binding = require(`@parcel/watcher-${process.platform}-${process.arch}${linux ? `-${libc || "glibc"}` : ""}`)
return createWrapper(binding)          // "@parcel/watcher/wrapper" — pure JS
```

inside a `try/catch`; when the `require` throws, `hasNativeBinding()` is `false` and the service is a no-op — which is why the native build "works" today with no watcher. The platform package's `main` is literally `watcher.node` (318 KB on darwin-arm64; 525 KB linux-x64-glibc; 533 KB win32-x64): there is no JS in it, so **the facade *is* that module**.

`wrapper.js` normalises options before they reach the binding: glob entries of `ignore` become `ignoreGlobs` as **picomatch regex sources** (`makeRe(glob, { dot: true, windows }).source`, matched against root-relative paths); non-glob entries become `ignorePaths` (absolute, `path.resolve(dir, value)`). `subscribe` resolves `dir` and awaits `binding.subscribe(dir, fn, opts)`; `unsubscribe` is `binding.unsubscribe(dir, fn, opts)`, i.e. **keyed by directory + callback identity + options**.

Two more facts about the call site: the project-directory subscription is behind `OPENCODE_EXPERIMENTAL_FILEWATCHER`; the `.git`-directory subscription (ignoring every entry but `HEAD`) is unconditional for git repos. Backend passed: `"fs-events"` (darwin), `"inotify"` (linux), `"windows"` (win32). OpenCode pins `@parcel/watcher@2.5.1`.

### Why this is a facade and not Node-API

- Performance is identical either way for a watcher: the boundary cost is a handful of C calls per event batch. The real perf story is OS events vs the **current `fs.watch`, which is a 25 ms timer that re-walks and re-stats the whole tree** (`crates/perry-runtime/src/fs/dir_glob_watch/watch.rs`, `FS_WATCH_POLL_INTERVAL_MS`). Measured locally (M-series, release build, 2026-08-21): 100-file tree → 0.14 s CPU per 5 s idle; **20,000-file tree → 1.47 s CPU per 5 s idle (~29 % of a core) and 360 ms event latency**. FSEvents/inotify: ~0 CPU idle, single-digit ms, O(events) not O(files).
- Size: a facade on the `notify` crate costs ~0.1–0.3 MB and only in binaries that reference it (`-dead_strip` / `--gc-sections`); the Node-API route costs an exported host (~0.3–0.6 MB) plus a 0.3–0.5 MB sidecar dylib per platform, and loses single-static-binary and every cross-compile target.
- Precedent: `better-sqlite3`, `sharp`, `bcrypt`, `argon2` are all Node-API addons upstream and all ship as `perry-ext-*` facades; #4961 made that the model; #8510 / #8512 in this tracker follow it.

## Design

**One native watcher core, three fronts.** The core is the `notify` 6.1 crate — already in the workspace behind `perry`'s `watch-cli` feature for `perry dev`; FSEvents / inotify / ReadDirectoryChangesW / kqueue; `Flag::Rescan` on overflow. It delivers events from its OS thread to the main thread through the `perry-ffi` event pump (`notify_main_thread` + a `process_pending` tick), the way `ws` / `http` / `net` do.

1. **`@parcel/watcher` binding object** — this issue's deliverable. Served for the root name *and* the eight platform names (`@parcel/watcher-{darwin,linux,win32}-{x64,arm64}[-{glibc,musl}]`) as alias rows in `well_known_bindings.toml`, so the wrapper's `require` lands on it. `@parcel/watcher/wrapper` itself compiles as ordinary JS (`path`, `picomatch`, `is-glob`).
2. **`fs.watch` / `fs.promises.watch`** — move directory watching onto the core; keep the snapshot poller only as the fallback when the core can't engage (and for `fs.watchFile`, which is polling by spec). Follow-up issue, but the core must be built so this is a thin adapter: `chokidar@4.0.3` (also in OpenCode's deps, pure JS over `fs.watch`), tsc, vite and nodemon all ride on it.
3. **`perry dev`** — already on `notify`; share the crate, not the code path.

## Required surface (binding object)

- `subscribe(dir, fn, { backend?, ignorePaths?: string[], ignoreGlobs?: string[] }) → Promise<void>`; `fn(err: Error | null, events: { path: string, type: "create" | "update" | "delete" }[])`, absolute paths.
- `unsubscribe(dir, fn, opts) → Promise<void>` — match by `(dir, fn identity, opts)`; the registry key for `fn` is the **rooted handle**, never the closure's raw address (#8393).
- `writeSnapshot(dir, snapshotPath, opts) → Promise<void>` and `getEventsSince(dir, snapshotPath, opts) → Promise<Event[]>` — OpenCode doesn't call them, but they are exactly the algorithm the existing snapshot-diff code implements (`snapshot_watch_target` / `diff_watch_snapshots` in `watch.rs`), so ship them rather than a partial binding.
- `backend`: accept `"fs-events" | "watchman" | "inotify" | "kqueue" | "windows" | "brute-force"`; unknown or unavailable on this platform → default backend, no throw (documented parcel behaviour). `watchman` → default backend.
- `ignoreGlobs` are regex sources → compile with the runtime's regex engine and match root-relative paths; `ignorePaths` are absolute prefixes.
- Event semantics to match (parcel README; verify against its test suite): events are throttled and coalesced into one notification per batch; **one event per file per batch** — create+update → `create`, create+delete → nothing; **rename → `delete` old + `create` new**; overflow / `Rescan` → re-snapshot and emit the diff (document the choice).
- macOS FSEvents, Linux inotify, Windows ReadDirectoryChangesW; kqueue for FreeBSD if `notify` gives it for free.

## Compiler side

- Fold the template-literal `require` above for the compile target (`process.platform` / `process.arch` / libc are constants of a native build). `cjs_wrap` already prunes `process.platform`-conditional requires (`crates/perry/src/commands/compile/cjs_wrap/tests.rs:732,767`); extend that to template specifiers, then resolve through the alias rows.
- `collect_modules/native_addon.rs` must not classify the facade-served platform packages as unsupported addons.

## Rooting / lifetime

- A live subscription owns its JS callback through the handle registry (scanned by a registered root scanner); the `notify` watcher lives in a native side table registered the same way — every such table is a GC root the static checker cannot see (`gc_register_mutable_root_scanner`, classified in `scripts/gc_runtime_root_holders.json`).
- `unsubscribe` must **stop and join the watcher thread and drain the pending queue before** dropping the handle, so no callback can land after its promise resolves.
- Process lifetime: a Node-API threadsafe function is ref'd by default, so in Node a live subscription keeps the process alive and `unsubscribe` releases it — verify and match.

## Acceptance

- [ ] `require("@parcel/watcher-<platform>-<arch>[-libc]")` and `import … from "@parcel/watcher"` resolve to the facade without a Node native addon; `@parcel/watcher/wrapper` compiles as JS
- [ ] fixtures: create, modify, rename (→ delete+create), delete, nested create, `ignorePaths`, `ignoreGlobs` (dot-files included), `unsubscribe`, two subscriptions on one dir with different callbacks, `writeSnapshot` / `getEventsSince` round-trip
- [ ] coalescing fixture: create+update in one batch → single `create`; create+delete → no event
- [ ] OpenCode native build (#8506) with `OPENCODE_EXPERIMENTAL_FILEWATCHER=1` publishes `Event.Updated` for a project file change; the `.git`-dir subscription fires on `git commit`
- [ ] no event callback can outlive an unsubscribed native resource: a fixture that unsubscribes while events are in flight, then forces collections, and asserts zero late callbacks
- [ ] the gate asserts the **native backend engaged** (not the poll fallback) — e.g. a counter of OS events delivered > 0 — per the "gate must prove its subject ran" rule
- [ ] idle CPU on a 20k-file tree under the facade is reported and ≈ 0 (baseline above: ~29 % of a core)
- [ ] behaviour documented under `docs/src/stdlib/` (coalescing, rename, overflow, backend fallback) with the parcel version it was checked against (2.5.x)

proggeramlug:
## Summary

- add a notify 6.1-backed @parcel/watcher 2.5.1 binding facade for the root and all eight platform aliases
- deliver main-thread, GC-rooted callbacks with stable coalescing, rename, ignore, snapshot, liveness, and safe-unsubscribe semantics
- fold OpenCode's target-dependent sidecar require, bypass facade-served Node addon rejection, and restore macOS CoreServices linkage at the custom linker boundary

## Verification

- cargo check -p perry -p perry-codegen -p perry-hir -p perry-api-manifest -p perry-ext-parcel-watcher
- cargo test -p perry-ext-parcel-watcher --lib (7 passed)
- cargo test -p perry --bin perry parcel_watcher (2 passed)
- cargo test -p perry-codegen --test manifest_consistency (5 passed)
- node scripts/binding_pins.mjs --check
- python3 scripts/gc_runtime_root_holders.py
- python3 scripts/workspace_architecture.py
- python3 scripts/ci_ext_link_scope.py
- standalone facade fixture: 14/14 assertions true, including native backend engagement, two subscribers, ignores, rename ordering, snapshots, forced-GC unsubscribe, and no late callbacks
- 20,000-file idle fixture: 1.253 ms process CPU over 5.080 s (about 0.025% of one core)

No version bump is included.

Closes #8513

## Summary by CodeRabbit

* **New Features**
  * Added native support for `@parcel/watcher` and platform-specific watcher packages.
  * Added filesystem event subscriptions, event coalescing, filtering, snapshots, rename handling, and overflow recovery.
  * Added support for native watcher operation across macOS, Linux, and Windows.

* **Bug Fixes**
  * Improved handling of platform-specific watcher imports and dynamic package resolution.
  * Allowed trailing commas in static CommonJS `require` calls.

* **Documentation**
  * Added usage and behavior documentation for the native watcher facade.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6022,6 +6022,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "perry-ext-parcel-watcher"
+version = "0.5.1514"
+dependencies = [
+ "fancy-regex",
+ "notify",
+ "perry-ffi",
+ "serde",
+ "serde_json",
+ "tempfile",
+]
+
+[[package]]
 name = "perry-ext-pdf"
 version = "0.5.1516"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,7 @@ members = [
     "crates/perry-ext-pdf",
     "crates/perry-ext-ads",
     "crates/perry-ext-node-forge",
+    "crates/perry-ext-parcel-watcher",
     "crates/perry-wasm-host",
     "crates/perry-container-compose",
     "crates/perry-container-e2e",
@@ -489,6 +490,7 @@ perry-ext-fastify = { path = "crates/perry-ext-fastify" }
 perry-ext-pdf = { path = "crates/perry-ext-pdf" }
 perry-ext-ads = { path = "crates/perry-ext-ads" }
 perry-ext-node-forge = { path = "crates/perry-ext-node-forge" }
+perry-ext-parcel-watcher = { path = "crates/perry-ext-parcel-watcher" }
 perry-stdlib = { path = "crates/perry-stdlib" }
 perry-diagnostics = { path = "crates/perry-diagnostics" }
 perry-ui-model = { path = "crates/perry-ui-model" }
--- a/changelog.d/8532-parcel-watcher.md
+++ b/changelog.d/8532-parcel-watcher.md
@@ -1,0 +1,6 @@
+feat(runtime): add a native `@parcel/watcher` 2.5.x compatibility facade backed
+by OS filesystem events. OpenCode's platform-selected watcher packages now route
+to one GC-rooted native implementation with coalescing, rename, ignore,
+snapshot, overflow-rescan, and safe-unsubscribe semantics instead of falling
+back to a no-op when its Node-API addon cannot load. A 20,000-file idle tree
+used about 1.3 ms of process CPU over five seconds on macOS.
--- a/crates/perry-api-manifest/src/entries.rs
+++ b/crates/perry-api-manifest/src/entries.rs
@@ -187,6 +187,17 @@ pub const NATIVE_MODULES: &[&str] = &[
     // for Socket Firewall's TLS-MITM CA so forge's pure-JS crypto isn't
     // AOT-compiled.
     "node-forge",
+    // @parcel/watcher's root binding and the eight published Node-API
+    // sidecars. HIR canonicalizes sidecars to the root dispatch table.
+    "@parcel/watcher",
+    "@parcel/watcher-darwin-x64",
+    "@parcel/watcher-darwin-arm64",
+    "@parcel/watcher-linux-x64-glibc",
+    "@parcel/watcher-linux-x64-musl",
+    "@parcel/watcher-linux-arm64-glibc",
+    "@parcel/watcher-linux-arm64-musl",
+    "@parcel/watcher-win32-x64",
+    "@parcel/watcher-win32-arm64",
 ];
 
 /// Node built-in submodules that Perry routes through the
--- a/crates/perry-api-manifest/src/entries/part_2.rs
+++ b/crates/perry-api-manifest/src/entries/part_2.rs
@@ -9,6 +9,62 @@ use super::*;
 
 pub(crate) const API_MANIFEST_PART_2: &[ApiEntry] = &[
     method_sig(
+        "@parcel/watcher",
+        "subscribe",
+        false,
+        None,
+        &[p_str("dir"), p_any("callback"), p_any("options")],
+        TypeSpec::Promise,
+    ),
+    method_sig(
+        "@parcel/watcher",
+        "unsubscribe",
+        false,
+        None,
+        &[p_str("dir"), p_any("callback"), p_any("options")],
+        TypeSpec::Promise,
+    ),
+    method_sig(
+        "@parcel/watcher",
+        "writeSnapshot",
+        false,
+        None,
+        &[p_str("dir"), p_str("snapshot"), p_any("options")],
+        TypeSpec::Promise,
+    ),
+    method_sig(
+        "@parcel/watcher",
+        "getEventsSince",
+        false,
+        None,
+        &[p_str("dir"), p_str("snapshot"), p_any("options")],
+        TypeSpec::Promise,
+    ),
+    method_sig(
+        "@parcel/watcher",
+        "__nativeEventCount",
+        false,
+        None,
+        &[],
+        TypeSpec::Number,
+    ),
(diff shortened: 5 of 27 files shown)


proggeramlug:

| check | result |
|---|---|

PR #8538, opened 2026-08-21
fix(gc): root define/Reflect operands above the typed-array key probe (#8507)

### LINKED ISSUE #8507: GC: defineProperty and Reflect/own-key paths use a stale receiver or value under forced evacuation

proggeramlug:
**GC: `Object.defineProperty` / `Reflect.defineProperty` filed the property under `"[object Object]"` when the key's coercion collected (#8507).**

A key object whose `toString` allocated enough to trigger an evacuating collection ended up storing the property under `"[object Object]"` instead of its real name:

```ts
const key = { n: "label", toString() { churn(); return this.n; } };
const ta = new Int32Array([1, 2, 3]);
Object.defineProperty(ta, key, { value: payload, configurable: true });
Object.getOwnPropertyNames(ta);   // was ["0","1","2","[object Object]"], now [...,"label"]
```

**Cause.** Both entry points probe the TypedArray integer-index path first, and that probe coerces the key inside `canonical_index_for_key` — which runs the user `toString` and can therefore evacuate. Every operand below the probe was still being read from the raw parameters, so they named pre-move addresses. Instrumenting `js_string_coerce` shows the signature directly: two coercions of the **same** pointer returning different answers.

```
[coerce] ptr=0x…94850 -> "loudkey"
[coerce] ptr=0x…94850 -> "[object Object]"
```

A moved key object still parses as an object at its old address, but its own `toString` is no longer reachable there, so the second coercion silently falls back to `Object.prototype.toString`. Both entry points now root the three operands above the probe and re-read them afterwards; `reflect_define_property` additionally re-reads between `array_length_reflect_define`, `obj_value_has_own_key` and `obj_value_attrs`, each of which re-coerces the key.

The `Reflect` half was not merely a lost value: a stale key made the own-key probe answer `false`, which skipped the non-configurable guard entirely and **accepted a redefine that must fail** — a dropped spec invariant, not just a missing property.

**Why it surfaced now.** The suite's own header calls itself a behavioural guard that "starts failing the day a minor-evacuating configuration becomes reachable from compiled code"; #6946 made an explicit `gc()` run an evacuating minor first. These are real defects newly exposed, not regressions.

`gc_string_coerce_property_key_rooting_6943` goes 1/3 → 3/3. Also verified green: `gc_property_key_operand_rooting_6935` (3/3), `gc_dynamic_arith_operand_rooting_6655` (4/4), `gc_side_table_roots_evacuation` (1/1).

Closes #8507.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed `Object.defineProperty` and `Reflect.defineProperty` handling when property keys require conversion during garbage collection.
  - Preserved correct property key names and ensured invalid redefinitions of non-configurable properties consistently raise errors.
  - Improved reliability when invoking accessors and dynamically rebound methods during garbage collection.

- **Documentation**
  - Added release notes covering property-definition fixes and receiver-handling improvements.

- **Chores**
  - Updated the application version to `0.5.1515`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1514
+**Current Version:** 0.5.1515
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5528,7 +5528,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1514"
+version = "0.5.1515"
 dependencies = [
  "anyhow",
  "base64 0.22.1",
@@ -5588,7 +5588,7 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1514"
+version = "0.5.1515"
 dependencies = [
  "perry-dispatch",
  "serde",
@@ -5596,7 +5596,7 @@ dependencies = [
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1514"
+version = "0.5.1515"
 dependencies = [
  "cc",
  "libc",
@@ -5605,7 +5605,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1514"
+version = "0.5.1515"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5622,7 +5622,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1514"
+version = "0.5.1515"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5630,7 +5630,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1514"
+version = "0.5.1515"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5638,7 +5638,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1514"
+version = "0.5.1515"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5647,7 +5647,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1514"
+version = "0.5.1515"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5655,7 +5655,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1514"
+version = "0.5.1515"
 dependencies = [
  "anyhow",
  "base64 0.22.1",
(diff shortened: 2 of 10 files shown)


proggeramlug:
Closing as redundant — #8526 landed the same fix while this was queued behind the runner pool, and I have verified main is green without this change:

```
test define_property_operands_survive_forced_evacuation ... ok
test descriptor_reads_survive_forced_evacuation ... ok
test reflect_and_own_key_paths_survive_forced_evacuation ... ok
test result: ok. 3 passed; 0 failed  (exit 0, on c721cb6a0)
```

Worth recording that the two efforts converged on the same root cause independently, which is decent corroboration that it is the right one: both entry points probe the TypedArray integer-index path first, that probe coerces the key inside `canonical_index_for_key` (running the user `toString`, which can evacuate), and every operand read below it named a pre-move address. #8526 covers more ground — it also roots across the exotic-expando probes and `array_object_ops` — so it is the better change to keep.

This branch was fully green before the conflict (`pr-gate` pass, `cargo-test` pass, `gc-stress` pass 40m9s, all six gap shards pass, zero failures), so nothing is being dropped for lack of validation.

One detail from this investigation that may be worth folding in wherever it is useful: the diagnostic that located the site was tagging **every** `js_string_coerce` call site with an env-gated `eprintln!("[site] file:line")` and wrapping the function to log its input and result. That prints the callers in order and names the culprit directly:

```
[site] typed_array_define.rs:144  -> "loudkey"          (user toString ran; collected)
[site] define_property.rs:885     -> "[object Object]"  (stale)
```

Backtraces are useless for this in perry — the linked binary is stripped, so `Backtrace::force_capture()` prints `<unknown>` even under `perry-dev` with debug info. The other trap: rooting at the *later* coercion changes nothing, because the move already happened at the earlier probe; the root has to be taken above the first thing that can collect.

PR #8539, opened 2026-08-21
perf(gc): skip tiny per-object layout masks

proggeramlug:
## Summary

Raise the minimum payload size for a per-object GC pointer mask from two slots to four. Two- and three-slot mixed payloads now use the collector's exact tag-checked scan instead of keeping a side-table mask alive and arming per-object layout cleanup on every allocation.

On the 19-program corpus this cuts `interp` retired instructions by **4.61%** (9,013,545,066 -> 8,598,251,738), with 19/19 byte-exact output, no other row moving by more than 1%, and no peak-RSS regression.

## Profiling finding

This was not the suspected dynamic-numeric-envelope problem. Across 30 sampled runs, `evalNode` accounted for **61.1% of self samples**. The useful signal was that its tiny pointer-bearing environment arrays repeatedly created and later removed per-object layout masks whose only local benefit was skipping one or two exact tag checks.

The cutoff is **4, not 8**, because a threshold sweep found the full `interp` retired-instruction win at 4. Raising it further to 8 retired no fewer instructions, while forcing larger objects through extra per-slot tag checks and broadening both the GC policy change and the number of layout preconditions affected. Four is the smallest threshold that captures the measured win.

## GC correctness

This changes a tracing-policy boundary, not just a performance knob. Below the cutoff the object enters `GC_LAYOUT_UNKNOWN`, whose child iterator scans every payload slot. The mask it replaces was itself built with `layout_pointer_bearing_bits`, and the unknown-layout scan applies that same predicate to every slot, so the scan visits a superset of the mask-selected slots: a pointer slot cannot be missed.

Intact typed descriptors are not redirected to this path. The two mask-minting sites either already operate on NaN-boxed slots or clear the typed-layout-intact bit before rebuilding, so raw-f64 payload bits cannot be mistaken for this safety argument.

## Changes

- Raise `DEFAULT_MASK_MIN_SLOTS` from 2 to 4.
- Update layout-tracing tests to pin the tag-scan behavior for two- and three-slot mixed objects and mask behavior at four slots.
- Document the profile, threshold sweep, corpus result, and correctness boundary.

## Related issue

Refs #8410

## Test plan

- [x] `cargo build --release`
- [x] `RUST_TEST_THREADS=1 cargo test --release -p perry-runtime` (2,608 passed, 4 ignored; 0 failed; doc tests 0 failed)
- [x] `PERRY_BIN=<target>/release/perry PERRY_RUNTIME_DIR=<target>/release PERRY_SKIP_BUILD=1 ./scripts/run_gap_tests.sh` (575/575 match the snapshot: 570 pass, 5 known parity mismatches, 0 compile failures, 0 crashes)
- [x] 19/19 benchmark outputs byte-exact
- [x] `cargo test --release -p perry-codegen` (1,113 passed)
- [x] `cargo test --release -p perry --bin perry` (1,008 passed)
- [x] all 52 lint gates

## Checklist

- [x] No workspace version bump and no `CLAUDE.md` / `CHANGELOG.md` edit
- [x] Commit uses the project prefix convention
- [x] Runtime tests cover the changed layout-mask/tag-scan boundary

## Summary by CodeRabbit

* **Performance**
  * Improved interpreter performance by reducing retired instructions by 4.61%.
  * Small payloads now use efficient tag scanning, while larger payloads retain optimized layout masks.
  * Benchmarks showed no significant regressions in other performance metrics or peak memory usage.

* **Tests**
  * Expanded runtime and garbage-collection coverage for small and large array layouts, pointer tracing, mapping, slicing, and cache behavior.

### DIFF
--- a/changelog.d/8539-interp-small-layout-mask-cutoff.md
+++ b/changelog.d/8539-interp-small-layout-mask-cutoff.md
@@ -1,0 +1,17 @@
+### Performance
+
+- **Cut `interp` retired instructions by 4.61% by keeping tiny mixed GC
+  payloads out of the per-object layout tables.** A 30-run sample profile put
+  61.1% of self samples in `evalNode`; the result did not support the suspected
+  dynamic-numeric-envelope explanation. Instead, its pointer-bearing
+  environment arrays were paying to create and later clean up per-object
+  layout masks even though those masks could skip only a few exact tag checks.
+
+  Raise the minimum mask-bearing payload from two slots to four, so one-, two-,
+  and three-slot payloads use the collector's exact tag scan. A cutoff sweep
+  found that four captured the full `interp` instruction win; moving the cutoff
+  to eight retired no fewer instructions, while it would make larger objects
+  perform extra tag checks and broaden the GC policy change. Across the
+  19-benchmark corpus, `interp` moved 9,013,545,066 → 8,598,251,738 retired
+  instructions (−4.61%), no other row moved by more than 1%, and peak RSS did
+  not regress.
--- a/crates/perry-runtime/src/array/tests.rs
+++ b/crates/perry-runtime/src/array/tests.rs
@@ -1268,22 +1268,22 @@ fn test_numeric_array_layout_length_and_delete_transitions() {
 
 #[test]
 fn test_numeric_array_layout_immutable_helpers_preserve_or_downgrade() {
-    let values = [10.0, 2.0, 30.0];
+    let values = [10.0, 2.0, 30.0, 40.0];
     let src = js_array_from_f64(values.as_ptr(), values.len() as u32);
     assert_numeric_raw_values(src, &values);
 
     let reversed = js_array_to_reversed(src);
-    assert_numeric_raw_values(reversed, &[30.0, 2.0, 10.0]);
+    assert_numeric_raw_values(reversed, &[40.0, 30.0, 2.0, 10.0]);
 
     let sorted = js_array_to_sorted_default(src);
-    assert_numeric_raw_values(sorted, &[10.0, 2.0, 30.0]);
+    assert_numeric_raw_values(sorted, &[10.0, 2.0, 30.0, 40.0]);
 
     let numeric_replaced = js_array_with(src, 1.0, 99.0);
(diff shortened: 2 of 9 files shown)


proggeramlug:
Audited and merging.

**The failure direction is the safe one, which is what makes this reviewable at all.** The change moves one-, two- and three-slot payloads from a per-object pointer mask to the collector's exact tag scan. A mask is an *optimization* that lets the tracer skip slots known not to hold pointers; the tag scan inspects each slot's actual NaN-box tag. So widening the tag-scan range cannot lose a pointer — it can only cost time. A change in the opposite direction (more objects onto masks) would be the dangerous one.

The soundness argument is stated where it belongs, at the constant:

> A tag check is exact at both mint sites: neither is reached for an object with an intact typed descriptor, so there are no raw-f64 slots whose bits a tag check could misread.

That is the right claim to make — the tag check is *exact* here, not merely conservative — and `typed_layout_intact_residual.rs` covers it.

**The rationale for `4` is measured, not assumed**, and the methodology is right:

| bench | threshold 2 | threshold 4 | delta |
|---|--:|--:|--:|
| `interp` | 9,013,545,066 | 8,598,251,738 | **−4.61%** |
| `iso_miss` | 12,494,387,198 | 12,489,410,538 | −0.04% |

Every other row moved less than 0.37%, `interp` peak RSS was byte-identical at 32,964,608, and the PR says plainly that wall ranges overlapped on a contended host so retired instructions are the primary signal. A threshold sweep found `4` is the *smallest* winning value — keeping the smallest one that captures the win is the correct choice, since every slot above the cutoff is a real mask entry with a real cost.

**The new invariant test is written to survive the constant changing** — it derives its payload size from `DEFAULT_MASK_MIN_SLOTS - 1` rather than hardcoding 3, so it keeps testing "below the cutoff mints no mask" if someone retunes the threshold. That is the difference between a test and a snapshot.

The motivating cost is real and previously bitten: one long-lived side-table entry arms `PER_OBJECT_LAYOUTS_NONEMPTY`, which puts a two-map hash probe on every allocation in the program — the same shape as the earlier incident where tiny environment arrays grew `LAYOUT_SLOT_MASKS` past 400k live entries.

`pr-gate` green on the current head.

PR #8537, opened 2026-08-21
fix(bun): complete OpenCode compatibility coverage

proggeramlug:
## Summary

- add Bun.Glob construction, matching, sync scanning, and async iteration over the native glob walker
- complete Bun.file/Bun.write binary and file-URL handling, including BunFile.bytes()
- surface explicit errors for unsupported direct Bun API calls
- enable regex feature detection for Bun.Glob and add focused OpenCode compatibility coverage

## Tests

- cargo test -p perry --test issue_6560_bun_globals bun_glob_scan_iteration_options_and_ignores -- --nocapture
- cargo test -p perry --test issue_6560_bun_globals bun_file_url_bytes_and_invalid_path_errors -- --nocapture
- cargo test -p perry --test issue_6560_bun_globals unsupported_bun_calls_fail_explicitly -- --nocapture
- cargo test -p perry-runtime bun_compat::tests::payload_bytes_preserves_uint8array_binary_data --lib
- cargo test -p perry-runtime bun_compat::string_width::tests --lib
- cargo test -p perry --bin perry regex_gate_detects_static_and_dynamic_path_matches_glob
- cargo check -p perry-runtime -p perry-hir -p perry-codegen -p perry
- cargo check -p perry-runtime --no-default-features
- cargo fmt --all -- --check

No version bump.

Fixes #8514

PR #8540, opened 2026-08-21
perf(codegen): drop roots for number-only locals

### LINKED ISSUE #8496: perf: typed_array is 5.25x Node (8.66x instructions) — the corpus's worst row, and it was untracked until now
## Summary

`typed_array` is now the **worst row in the corpus by a wide margin** — 5.25x Node on wall,
**8.66x on instructions retired**. It was invisible to every sweep until today because it was
not in the corpus at all.

That is not an oversight in the benchmark's design — `benchmarks/suite/bench_typed_array_untyped_access.ts`
prints its own wall-clock timings, and the four-engine sweep's correctness gate is a
byte-exact stdout compare against the Node oracle, so a self-timing program can never join it.
I added `gc-handoff/apps/typed_array.ts`: identical workload (bcryptjs's Blowfish `_encipher`
shape, `S`/`P` reached through untyped params, ~4.27M calls), emitting only
`checksum:-821955270`. Verified deterministic. The suite version remains the #5525
untyped/typed ratio metric; the corpus one is the regression tripwire.

## Measurement — quiet M1 mini, 5 shuffled interleaved repeats, verdict CLEAN, perry `8d1cca158`

| | node | perry | ratio |
|---|---:|---:|---:|
| wall | 0.644 s | 3.385 s | **5.25x** |
| instructions | — | — | **8.66x** |
| peak RSS | — | — | **0.06x** |

For scale, the next-worst rows are `iso_miss` 1.94x and `interp` 1.51x. This one row is
roughly three times worse than anything else Perry runs.

## What is already done — do not redo it

#8491 (closing #8485) landed a `number_by_construction` dataflow proof and moved this
workload **−44.1% wall / −34.7% instructions / −52.6% peak RSS**, with the untyped/typed ratio
improving 1.0438 → 1.0017. So the untyped and typed paths now cost nearly the same — the
remaining 5.25x is **not** a #5525-style untyped-vs-typed gap. Both paths are slow against
Node together.

Two further facts established during that work, worth not re-deriving:

- The S-box element loads are **already native**. Optimized AArch64 shows
  `ldr ..., [base, index, uxtw #2]` — no NaN-box round trip, no per-element registry lookup.
- The cost was the dynamic numeric envelope around them. #8491 addressed the
  `let n: number;`-seeded-`undefined` shape. Whatever remains is a *different* part of that
  envelope, or something else entirely.

## Why this is the best available target

The instruction ratio (8.66x) is far above the wall ratio (5.25x), so the work is being
executed and there is no scheduling story to chase. Peak RSS is 0.06x Node — there is enormous
headroom to spend memory if that ever helped, though the standing project constraint is to
minimize RSS *while* keeping the best compute, so any RSS cost must be quantified and argued.

## Suggested first step

Profile before hypothesizing. Compile with `PERRY_KEEP_SYMBOLS=1 PERRY_DEBUG_SYMBOLS=1`
(Perry strips output binaries; without these `sample` attributes everything to `???`) and
aggregate over many runs. Find where the 8.66x actually goes now that #8491 has removed the
previously dominant leaves.

proggeramlug:
## Summary

- remove shadow-slot bindings only for source-declared `Number`/`Int32` locals that the specialization-aware whole-write analysis independently proves Number-by-construction
- keep `Any` and `Unknown` bindings conservative even when inferred local types or arithmetic facts look numeric
- preserve generic fallback roots when guarded input facts are unavailable
- add IR, direct allowlist, and end-to-end regressions for opaque locals and exact-once evaluation of side-effecting `Any` receivers

## Rooting correctness

The optimization uses two independent requirements: a declaration allowlist (`Number`/`Int32`) and the annotation-agnostic whole-write Number proof. Inferred `local_types` are deliberately excluded from the allowlist because they can refine an `Any` declaration. A direct set-intersection unit test pins that a local satisfying only one requirement retains its slot. Boxed/captured locals and module globals remain excluded by the existing proof, and generic fallback bodies retain conservative slots.

The frame keeps its original upper-bound size; filtering only prevents bindings, root stores, clears, and barriers for admitted locals, so surviving preassigned indices remain valid.

## Performance

The measured branch reduced `typed_array` from 56.669 B to 56.013 B retired instructions (-1.16%), with all 20 corpus outputs byte-exact and checksum `-821955270`. The safety repair narrows the candidate set to declared numeric bindings; it keeps the benchmark target (`let n: number`) eligible while opaque locals remain rooted.

## Validation

- `cargo test --release -p perry-codegen --lib`: 1123 passed
- targeted `argless_builtin_extra_args`: 4/4 passed; refreshed CI also confirms the previous exact-once receiver failure is gone
- `git diff --check`
- previous-head CI failures outside this diff remain present on current `main`: the file-size/raw-handle lint gates plus pre-existing `native_proof_regressions` and `temp_root_operand_temporaries` integration assertions

No version bump.

Closes #8496.

## Summary by CodeRabbit

* **Performance**
  * Improved runtime efficiency for functions that use numeric local variables, reducing unnecessary memory and cleanup work.
  * Preserved conservative handling when numeric behavior cannot be proven, maintaining existing correctness and compatibility.
* **Quality**
  * Added coverage to verify optimized numeric code paths and ensure safe fallback behavior.
* **Documentation**
  * Added benchmark results and output comparisons documenting the performance improvement.

### DIFF
--- a/changelog.d/8496-number-local-root-slots.md
+++ b/changelog.d/8496-number-local-root-slots.md
@@ -1,0 +1,30 @@
+### Performance
+
+- Stop binding and maintaining shadow-root slots for locals proved Number-
+  by-construction. The shadow-root map is deliberately conservative because
+  it is collected before specialization facts exist; a guarded typed-array
+  clone could therefore keep root stores and write barriers in its integer
+  hot loop even after the later fact graph proved the local non-pointer.
+
+  The omission is safe at every collection point, not merely at reads. The
+  proof considers only function-local `let` bindings, excludes boxed/captured
+  locals and module globals, never trusts TypeScript annotations, and checks
+  the initializer plus every later write. Its ordinary arm admits only Number
+  values. Its specialized undefined-seed arm admits only Number or Perry's
+  non-pointer `undefined` sentinel and permanently rejects the local after any
+  other write, even when that value is overwritten before the next read.
+  Parameters and conditional expression writes stay fail-closed. Generic
+  fallback bodies therefore retain their conservative slots when the guarded
+  input proof is unavailable.
+
+  On current `main` (8d837df22) with the 20-row #8496 corpus
+  (`PERRY_NO_AUTO_OPTIMIZE=1`, release compiler and matching runtime archives,
+  `/usr/bin/time -l`, interleaved medians of 5), `typed_array` retires 56.013 B
+  instructions instead of 56.669 B (-1.16%). Peak RSS is 5.489 MB instead of
+  5.472 MB (+16 KiB, +0.30%). Heavy host contention makes the wall medians
+  non-actionable (25.56 s instead of 13.30 s, with overlapping run ranges).
+  Five shorter rows cross 1% in instruction medians amid run-to-run/GC noise:
+  `asyncpipe` +2.11%, `cycles` +1.67%, `retain1` -2.11%, `retain_wide1`
+  -1.22%, and `shapes` -2.12%; `churn`, `tree`, and `retain` move -0.25%,
+  -0.18%, and +0.21%. All 20 outputs remain byte-exact, including
+  `typed_array`'s `-821955270` checksum.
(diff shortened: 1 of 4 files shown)


proggeramlug:

| check | result |
|---|---|

PR #8542, opened 2026-08-21
runtime: keep wasm-bindgen memory and tables live

proggeramlug:
## Summary

- synchronize JavaScript writes into exported WebAssembly memory before calls and copy Wasm writes and growth back afterward
- preserve numeric multi-value returns and coerce numeric arguments using the declared Wasm signature
- expose exported externref tables with get, set, grow, and GC-retained JS values
- defer table mutations made through imported JS callbacks until wasmi releases its Store, covering the wasm-bindgen import cycle safely
- extend the packaged .wasm end-to-end fixture and document the supported subset

## Validation

- `cargo test -p perry-wasm-host`
- `cargo check -p perry-runtime --features wasm-host`
- `cargo test -p perry --test issue_5234_wasm_esm_import -- --nocapture`
- `cargo clippy -p perry-wasm-host -- -D warnings`
- compiled the patched `@silvia-odwyer/photon-node@0.3.4` loader with `with { type: "file" }`, then ran the binary from an unrelated empty directory: `size=2x1`, `pixels=8:255:0`

## Scope

This lands the Photon and wasm-bindgen live-state slice of #8508. Imported tables, memories, and globals plus the web-tree-sitter dynamic grammar path remain follow-up work, so this PR intentionally references rather than closes the umbrella issue.

No version bump.

Refs #8508

## Summary by CodeRabbit

* **New Features**
  * Improved WebAssembly memory synchronization during runtime calls.
  * Added support for numeric functions returning multiple values as arrays.
  * Added `externref` table operations: reading, updating, and growing tables.
  * Added numeric argument coercion for WebAssembly calls.

* **Documentation**
  * Updated WebAssembly support documentation to clarify supported features and limitations.

* **Bug Fixes**
  * Improved table mutation handling during callbacks and cyclic initialization.

### DIFF
--- a/changelog.d/8508-wasm-bindgen-live-state.md
+++ b/changelog.d/8508-wasm-bindgen-live-state.md
@@ -1,0 +1,6 @@
+### runtime: keep wasm-bindgen memory and table exports live
+
+WebAssembly exports now synchronize writes through `memory.buffer`, preserve
+numeric multi-value results, coerce numeric arguments against the Wasm
+signature, and expose `externref` table `get`/`set`/`grow`. This enables the
+runtime state patterns used by file-packaged wasm-bindgen Node modules.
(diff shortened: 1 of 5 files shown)


proggeramlug:

| check | result |
|---|---|

PR #8543, opened 2026-08-21
fix(runtime): preserve entry source in process argv

proggeramlug:
## Summary

- seed compiled executables with the canonical TypeScript entry path
- expose that path as process.argv[1] while preserving the executable at argv[0] and user arguments at argv[2+]
- keep dylib initializers from modifying their host process argv
- include the embedded entry path in the object-cache key
- add IR-level and gap-suite regression coverage for the import.meta.url direct-execution guard

## Root cause

Perry baked import.meta.url from the source module path, but constructed process.argv[1] by duplicating the executable path. The conventional Node/Bun direct-execution comparison therefore could never succeed in a native Perry executable.

## Validation

- cargo fmt --all -- --check
- cargo check -p perry-runtime -p perry-codegen -p perry
- cargo test -p perry-codegen --lib executable_seeds_process_argv_script_path_but_dylib_does_not
- cargo test -p perry --bin perry key_changes_with_embedded_entry_source_path
- compiled and ran the regression fixture with the rebuilt compiler/runtime; output matched Node byte-for-byte
- verified the compiled executable still fires the guard when launched from a different working directory

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed compiled executables so `process.argv[1]` correctly identifies the TypeScript entry module.
  - Direct-execution checks using `import.meta.url` and `process.argv[1]` now work as expected under Node and Bun.
  - Preserved user-provided command-line arguments while improving entry-path detection.
  - Prevented library builds from incorrectly replacing the host process’s entry path.


proggeramlug:

| check | result |
|---|---|

PR #8545, opened 2026-08-21
fix(gc): census perry_thread_local holders

### LINKED ISSUE #8544: gc: gc_runtime_root_holders.py cannot see perry_thread_local! — 248 of ~881 holders invisible
## Problem

`scripts/gc_runtime_root_holders.py` enumerates `static` / `thread_local!` /
`lazy_static!` declarations. It does **not** know about Perry's own TLS macro,
`perry_thread_local!` (`crates/perry-runtime/src/tls_hot.rs`), so every holder
declared through it is invisible to the gate.

Measured on `7c22189aa` (main):

| | count |
|---|---|
| declarations the gate scans | **633** |
| declarations inside `perry_thread_local!` blocks, in the two core crates | **248** (75 files) |
| gate script mentions of `perry_thread_local` | **0** |

So the gate sees roughly **72%** of the population it is written to guard.

Spot-checked, none of these is enumerated (`--list` has no row for any):

- `EXCEPTION_STATE: UnsafeCell<ExceptionState>` (`exception.rs`)
- `TYPED_ARRAY_OWN_PROPS: RefCell<PtrHashMap<usize, Vec<TypedArrayOwnProp>>>` (`typedarray_props.rs`)
- `YOGA_NODES: RefCell<Option<HashMap<u32, YogaNode>>>` (`yoga.rs`)
- `PENDING_REQUIRE_PARENT`, `MODULE_PATH_REGISTRY`, `PATH_MODULE_INIT_ADDRS` (`module_require*`)

A plain `thread_local!`/`static` in the same crates *is* enumerated, so this is
specifically the macro, not the file set.

## Why it matters

The failure direction is toward **green**. A narrower match means fewer hits and
a passing gate, and nothing ever prompts the check. Concretely, today:

`MODULE_PATH_REGISTRY` holds NaN-boxed heap pointers and **is** correctly scanned
(`scan_module_path_roots_mut`, registered in `gc/mod.rs`). But because the gate
cannot see the declaration, **deleting that `reg_scanner!` line would not turn
the gate red** — the exact scenario the gate exists to prevent.

This is the same class as #8207 and the #8163 hunt: a too-narrow symbol/declaration
regex silently removing holders from consideration.

## Fix

Teach the enumerator to parse `perry_thread_local! { ... }` (and the
`crate::perry_thread_local!` spelling) the way it already parses `thread_local!`
blocks — the inner syntax is the same `static NAME: TYPE = init;`.

**Budget for the unmasking.** Making the gate see 248 more declarations will
surface holders that currently have no verdict, and the gate fails on an
unclassified holder by design. Expect it to go red until they are triaged; that
redness is the fix working, not a regression. Worth landing the enumerator change
and the triage together, or behind a temporary allowlist that can only shrink.

Please also report the population delta (633 -> N) in the PR: a scanner fix
without a stated blast radius is indistinguishable from one nobody measured.

proggeramlug:
Closes #8544.

## What changed

- explicitly parse both `perry_thread_local! { ... }` and `crate::perry_thread_local! { ... }` blocks
- classify declarations that core rules A/B cannot see through as rule T
- put uncovered rule-T declarations on the existing identity-pinned frontier, so the newly visible debt is bounded and can only shrink
- allow a frontier entry to name a cross-file registered scanner; removing that registration invalidates the pin and fails the gate
- add self-tests for both macro spellings, covered/uncovered opaque types, stale pins, and scanner-deletion failure paths

The last point protects the concrete `MODULE_PATH_REGISTRY` scenario from the issue: its declaration is in `module_require/path_registry.rs`, while `scan_module_path_roots_mut` is registered from `gc/mod.rs`. Its frontier entry now requires that exact scanner registration, so deleting the registration turns the gate red.

## Population delta

The issue's 248 estimate is 246 on current `7c22189aa` main. The old context-free declaration regex incidentally saw 27 of those because their types already matched rules A/B; the actual blind spot was that the remaining opaque/core types were discarded by classification.

- before: 633 candidate holders
- after: 852 candidate holders (+219)
- Perry TLS declarations represented after this change: 246 / 246
- newly surfaced rule-T holders: 219 (66 reached by same-file scanner analysis, 153 identity-pinned)
- frontier baseline: 466 -> 619 (+153)

## Verification

- `python3 -m py_compile scripts/gc_runtime_root_holders.py`
- simulated removal of `scan_module_path_roots_mut` and confirmed `MODULE_PATH_REGISTRY` becomes both new/unpinned and stale
- `SKIP_COMPILE_GATES=1 scripts/run_lint_gates.sh`: the issue-specific checks and 48/50 script gates pass; the two failures are existing main-tree debt outside this diff (`collect_modules.rs` over the file-size ceiling and existing raw-handle counts). The raw-handle `--no-raise-vs upstream/main` check passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection and tracking of runtime garbage-collection roots, including thread-local and UI-related holders.
  * Added validation to identify inactive or missing runtime scanners.
  * Expanded support for cross-file runtime root tracking.

* **Documentation**
  * Updated runtime frontier documentation with broader holder coverage and clearer pinning rules.

### DIFF
--- a/scripts/gc_runtime_root_holders.json
+++ b/scripts/gc_runtime_root_holders.json
@@ -443,8 +443,621 @@
       "why": "A timeOrigin-relative millisecond timestamp (or the -1 'loop has not started' sentinel) reported as nodeTiming.loopStart, and the gate eventLoopUtilization() reads before reporting anything but zeros. An f64 time value, never an address."
     }
   ],
-  "_FRONTIER_README": "Identity-pinned ratchet over the perry-ui* crates (see the census docstring, 'The frontier tier'). Each entry is a static/thread_local that CAN hold a NaN-boxed JS callback and that no registered scanner reaches. These are not verdicts: most of these tables really do park user closures (BUTTON_CALLBACKS et al) and registering per-crate scanners is the fix. The list may only shrink: a new UI holder fails lint until pinned deliberately, and a fixed (covered) holder fails until its entry is deleted. Covered frontier holders must NOT be pinned.",
+  "_FRONTIER_README": "Identity-pinned ratchet over the perry-ui* crates and otherwise-unclassified core perry_thread_local! declarations (see the census docstring, 'The identity-pinned frontier'). Entries are debt baselines, not verdicts. A new uncovered holder fails until deliberately pinned, and a fixed holder fails until its stale entry is deleted. An optional scanner names cross-file coverage and must remain registered; deleting that registration invalidates the pin and fails the gate. Covered same-file holders must NOT be pinned.",
   "frontier": [
+    {
+      "file": "crates/perry-runtime/src/array/element_shape.rs",
+      "name": "ELEMENT_SHAPES"
+    },
+    {
+      "file": "crates/perry-runtime/src/array/iterator.rs",
+      "name": "REGISTERED"
+    },
+    {
+      "file": "crates/perry-runtime/src/box.rs",
+      "name": "ASYNC_BOX_ACTIVATION_FREE_HEAD"
+    },
+    {
+      "file": "crates/perry-runtime/src/box.rs",
+      "name": "ASYNC_PENDING_RELEASES"
+    },
+    {
+      "file": "crates/perry-runtime/src/box.rs",
+      "name": "ASYNC_RELEASED_CELLS"
+    },
+    {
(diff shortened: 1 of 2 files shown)


proggeramlug:

| check | result |
|---|---|

Issue #8547, opened 2026-08-21
require('http') links runtime-only, so stdlib builtins return undefined (ESM import works)


proggeramlug:
`crates/perry/tests/issue_4903_listen_callback_deferred.rs` fails both its tests on `main`, and it is now the **only** thing keeping the sweep-tier `cargo-test` red (and therefore `main-gate`).

```
thread 'late_listening_listener_fires' panicked at issue_4903_listen_callback_deferred.rs:45
thread 'listen_callback_is_deferred_and_this_bound' panicked at issue_4903_listen_callback_deferred.rs:45
test result: FAILED. 0 passed; 2 failed
```

Line 45 is the "compiled binary failed" assertion, so this is a runtime failure, not a wrong value.

**Reduced.** `http.createServer` returns `undefined` in every form, while the export itself is a function:

```ts
const http = require('http');
console.log('no-arg      :', typeof http.createServer());                       // undefined
console.log('with-handler:', typeof http.createServer(function () {}));         // undefined
console.log('arrow       :', typeof http.createServer(() => {}));               // undefined
console.log('createServer:', typeof http.createServer);                         // function
```
Node prints `object` for the first three. The test's own failure is downstream of this — `.listen(0, cb)` on `undefined` throws `TypeError: Cannot read properties of undefined (reading 'listen')`.

**Where it comes from.** `("http", "createServer")` in `native_module_dispatch/dispatch_d_i.rs` routes through `JS_NATIVE_HTTP_DISPATCH`, and returns `undefined` when that pointer is null:

```rust
let ptr = crate::value::JS_NATIVE_HTTP_DISPATCH.load(Ordering::SeqCst);
if ptr.is_null() { f64::from_bits(JSValue::undefined().bits()) } else { … }
```

That pointer is only installed by `perry-stdlib`'s `common/dispatch/init.rs:689`, under `#[cfg(feature = "external-http-server-pump")]` — and that feature is **not** part of `perry-stdlib`'s `default`/`full` set. So the question is whether the compile path that these tests exercise links a stdlib with the feature on. Locally the compile reports `Linking (runtime-only)`, which is consistent with an http-using program being linked without the http implementation.

Two things I ruled out while narrowing, worth stating so the next person does not repeat them:

[...]

PR #8551, opened 2026-08-21
Revert "treat require(<builtin>) as a stdlib requirement" (#8548) — runtime-only link regression

proggeramlug:
**Revert "treat `require(<builtin>)` as a stdlib requirement" (#8548) — it caused a runtime-only link failure.**

#8548 fixed a real bug (#8547: `require('http')` linked runtime-only, so every stdlib-backed builtin reached that way returned `undefined`, and `issue_4903_listen_callback_deferred` went 0/2 → 2/2). It also broke `issue_5247_property_read_source_location`, which had been passing:

```
compile ["--debug-symbols"] must succeed; stderr:
/usr/bin/ld: libperry_runtime.a(...): in function `js_webassembly_validate':
  undefined reference to `perry_wasm_host_validate'
  undefined reference to `perry_wasm_host_module_new'
  … collect2: error: ld returned 1 exit status
```

**Attribution is solid, which is why this is a revert and not a follow-up patch.** The suite passed on `1c2326554` (the commit immediately before) and fails on `b00e261f8` (with #8548) in **both** sweep runs of that SHA — deterministic, not a flake. It does *not* reproduce locally, which is the tell for a feature-unification effect rather than a logic error: under `cargo test --workspace` cargo unifies features across the whole graph, so pulling perry-stdlib onto more compile paths changes what `target/debug/libperry_runtime.a` contains. The archive then carries `webassembly.rs`'s code, whose `extern` declarations are satisfied only by the separate `perry-wasm-host` crate — which a runtime-only link line does not include.

So the honest position is that #8548's *diagnosis* stands and its *link-mode consequence* was not thought through: making a program need the stdlib is not free, and the runtime-only link path is not prepared for everything the stdlib's feature set drags into the runtime archive. Fixing #8547 properly needs that interaction handled — either by ensuring `perry-wasm-host` is on the link line whenever the runtime archive contains wasm code, or by keeping the wasm surface out of it.

Net effect of this revert: `issue_4903` returns to failing (2 tests, the pre-existing #8547 symptom) and `issue_5247` returns to passing (3 tests). Main is red either way; this restores the *known* red rather than a new link-level one, which is the safer state to leave a release branch in — a link regression can bite real builds, not just tests.

Reopens #8547.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reverted recent CommonJS standard-library linking changes to prevent runtime linker failures.
  * Restored compatibility for affected projects while preserving previously resolved behavior in other scenarios.

* **Documentation**
  * Updated release documentation to reflect the reverted behavior and its known impact.

* **Chores**
  * Incremented the application version to 0.5.1516.


proggeramlug:
**Correction — I attributed this to Cargo feature unification, and that was wrong.**

I said the wasm link failure came from `cargo test --workspace` unifying `perry-runtime/wasm-host` on. That explanation does not survive checking: **no workspace crate enables `perry-runtime/wasm-host`.** The only occurrence in any `Cargo.toml` is the feature's own declaration (`wasm-host = []`). Unification cannot turn on a feature nothing requests.

The actual mechanism is **shared-archive contamination between tests in the same job**:

- `optimized_libs/no_auto.rs` builds `perry-runtime-static` with default features **plus `perry-runtime/wasm-host`** when `ctx.needs_wasm_runtime` is set — i.e. when the compiled program references `WebAssembly.*`. That is the `PERRY_NO_AUTO_OPTIMIZE=1` path.
- `issue_5247_property_read_source_location` also runs with `PERRY_NO_AUTO_OPTIMIZE=1` and points `PERRY_RUNTIME_DIR` at the same shared `target/debug`, rebuilding only `-p perry-runtime` (no wasm) via its own `ensure_runtime_archive`.

So under `cargo test --workspace`, any test that compiles a WebAssembly-using program rewrites the shared `libperry_runtime.a` **with** the wasm surface. A later runtime-only link from 5247 then hits `undefined reference to perry_wasm_host_*`, because the host archive is only put on the link line when *that* program sets `needs_wasm_runtime`. Whichever build wrote the archive last decides the outcome.

That reframes the situation in two ways worth being explicit about:

1. **It is a pre-existing fragility, not something #8548 created.** #8548 perturbed which tests rebuild the shared archive and in what order, which is why it flipped a latent race into a deterministic failure on that SHA. The revert is still correct — I would not leave a link-level red on a release branch — but "#8548 caused it" overstates my change's role, and I want that on the record before someone re-lands the detection and is surprised again.
2. **The durable fix is archive isolation, not the link path.** Two cargo invocations with different feature sets write the same `target/debug/libperry_runtime.a`; nothing arbitrates. Giving the runtime-archive-consuming tests their own `PERRY_RUNTIME_DIR` (or keying the archive path by feature set) removes the whole class. That is a smaller and safer change than either of the link-line options I listed earlier, and it would let #8547's detection re-land as-is.

I am not making that change myself: it touches test infrastructure I did not write, and I have already demonstrated on this issue that acting with an incomplete model of this link path produces regressions. Flagging it for an owner with the evidence above.

PR #8552, opened 2026-08-21
fix(gc): register mutable root scanners per thread

### LINKED ISSUE #8530: gc: 18 subsystems register their root scanner behind a process-global latch, but the scanner registry is thread-local
## Summary

`MUTABLE_ROOT_SCANNERS` is **thread-local** (`crates/perry-runtime/src/gc/roots.rs:105`). Registration must therefore happen **once per thread**. **18 files guard it with a process-global `Once`/`OnceLock` instead**, so the first thread to touch the subsystem registers its scanner and consumes the latch — and every later thread's collector never scans that subsystem's roots at all.

This is the same defect #8503 just fixed for the yoga backend, where `GC_SCANNER_REGISTERED` was a global `Mutex<bool>` in front of the same thread-local registry. It is not a one-off.

## The correct pattern, already in tree

`gc/mod.rs:746` gets it right, and says why:

```rust
/// scanner list (`MUTABLE_ROOT_SCANNERS`) is thread-local, so soundness
/// requires every thread that can trigger a collection to register
/// independently — not just the main thread that runs `js_gc_init()`.
static GC_INIT_DONE: std::cell::Cell<bool> = const { std::cell::Cell::new(false) };
```

Thread-local latch, thread-local registry. #8528 follows the same rule for the per-heap path-module table.

## The 18 offenders

Each of these registers via `gc_register_mutable_root_scanner*` inside a `call_once` / `get_or_init` on a process-global latch:

- `crates/perry-runtime/src/cluster.rs`
- `crates/perry-runtime/src/object/native_this_alias.rs`
- `crates/perry-stdlib/src/commander.rs`
- `crates/perry-stdlib/src/common/async_bridge.rs`
- `crates/perry-stdlib/src/cron.rs`
- `crates/perry-stdlib/src/crypto/hash_handles.rs`
- `crates/perry-stdlib/src/domain.rs`
- `crates/perry-stdlib/src/events.rs`
- `crates/perry-stdlib/src/exponential_backoff.rs`
- `crates/perry-stdlib/src/fetch/gc.rs`
- `crates/perry-stdlib/src/net/mod.rs`
- `crates/perry-stdlib/src/readline/mod.rs`
- `crates/perry-stdlib/src/sqlite.rs`
- `crates/perry-stdlib/src/streams/gc.rs`
- `crates/perry-stdlib/src/tls.rs`
- `crates/perry-stdlib/src/worker_threads.rs`
- `crates/perry-stdlib/src/ws.rs`
- `crates/perry-stdlib/src/zlib.rs`

Worked example — `worker_threads.rs:175,228`:

```rust
static WORKER_GC_REGISTERED: Once = Once::new();
...
WORKER_GC_REGISTERED.call_once(|| {
    perry_runtime::gc::gc_register_mutable_root_scanner_named(
        "stdlib:worker_threads:workers", scan_worker_roots_mut);
});
```

`scan_worker_roots_mut` walks a process-global `WORKERS: Mutex<HashMap<..>>` holding NaN-boxed listener callbacks — and, after #8527, the Worker handle itself. A `worker_threads` program is multi-threaded by construction, so this one is not hypothetical.

## Triage note

The *pattern* is uniformly wrong; the *reachability* is not uniform. A subsystem only ever touched from the main JS thread is latently wrong but currently harmless. Each of the 18 needs a one-line answer to "can a second heap reach this?" before deciding between converting the latch and documenting why it cannot happen. `worker_threads`, `cluster`, `common/async_bridge`, `net`, `ws` and `streams` look like the ones to check first.

## Why it will not show up as a test failure

An unregistered scanner produces no error. The roots are simply invisible to that thread's collector, so the failure is a use-after-free surfacing later and elsewhere — the #7154 signature, `TypeError: value is not a function` in an unrelated function some cycles on. **The failure direction is green.**

## Gate

`scripts/gc_runtime_root_holders.py` cannot catch this: it verifies a holder is *reached by a registered scanner*, treating registration as a static fact, not a per-thread one. A lint that flags `gc_register_mutable_root_scanner*` inside a non-thread-local `call_once`/`get_or_init` would be a cheap and exact check.

Found while auditing #8527, which adds a new root (`worker.object_bits`) to one of the mis-registered scanners.

proggeramlug:
## Summary

- replace process-global GC scanner registration latches with per-thread Cell<bool> latches across the affected runtime and stdlib subsystems
- keep genuinely process-wide setup process-wide by splitting the cluster arity and streams callback latches
- fix the same missed AtomicBool pattern in messaging and add a CI audit for Once, OnceLock, AtomicBool, and Mutex<bool> guards around mutable-root scanner registration
- no version bump

The new audit identifies 21 pre-fix bad latch sites across the 18 affected files and finds none after this change.

## Testing

- cargo check -p perry-runtime -p perry-stdlib
- cargo test -p perry-runtime --lib (2611 passed, 4 ignored)
- cargo test -p perry-stdlib --lib (126 passed)
- python3 scripts/check_gc_scanner_latches.py --self-test
- python3 scripts/check_gc_scanner_latches.py
- SKIP_COMPILE_GATES=1 BASE_SHA=upstream/main scripts/run_lint_gates.sh (all 52 script gates passed)

Closes #8530

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved garbage-collection support across threaded runtime and standard-library features.
  * Ensured per-thread resources and callbacks are tracked correctly, reducing the risk of missed roots or memory-management issues in multithreaded workloads.

* **Tests**
  * Added automated checks to detect incorrect garbage-collection registration patterns.
  * Added self-tests covering valid and invalid registration scenarios.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -410,6 +410,15 @@ jobs:
           python3 scripts/gc_runtime_root_holders.py --self-test
           python3 scripts/gc_runtime_root_holders.py
 
+      # #8530. Mutable-root scanner registries are thread-local, so guarding a
+      # registration call with a process-global Once/OnceLock/AtomicBool (or a
+      # mutex boolean) leaves every heap after the first without those roots.
+      - name: Thread-local GC scanner registration audit
+        if: ${{ !cancelled() }}
+        run: |
+          python3 scripts/check_gc_scanner_latches.py --self-test
+          python3 scripts/check_gc_scanner_latches.py
+
       # #8174. A side table whose KEY is a raw heap address, REKEYED in place by
       # `visit_metadata_*` (rewritten if the object moved, deliberately not
       # marked), depends on `gc::dead_owner` dropping that key when the object
--- a/crates/perry-runtime/src/cluster.rs
+++ b/crates/perry-runtime/src/cluster.rs
@@ -11,7 +11,7 @@
 //! the primary-coordinated shared ephemeral port for `listen(0)` are tracked
 //! in #4962.
 
-use std::cell::RefCell;
+use std::cell::{Cell, RefCell};
 use std::collections::HashMap;
 use std::sync::{Once, OnceLock};
 
@@ -51,6 +51,8 @@ impl ClusterState {
 
 thread_local! {
     static CLUSTER_STATE: RefCell<ClusterState> = RefCell::new(ClusterState::new());
+    // The mutable-root scanner registry is thread-local, so this latch must be too.
+    static CLUSTER_GC_REGISTERED: Cell<bool> = const { Cell::new(false) };
 }
 
 static CLUSTER_INIT: Once = Once::new();
@@ -600,8 +602,14 @@ pub extern "C" fn js_cluster_disconnect(callback: f64) -> f64 {
 }
 
 fn ensure_cluster_runtime() {
-    CLUSTER_INIT.call_once(|| {
+    CLUSTER_GC_REGISTERED.with(|registered| {
+        if registered.get() {
+            return;
+        }
         crate::gc::gc_register_mutable_root_scanner_named("node_cluster", cluster_root_scanner);
+        registered.set(true);
+    });
+    CLUSTER_INIT.call_once(|| {
         register_cluster_arities();
     });
 }
--- a/crates/perry-runtime/src/messaging.rs
+++ b/crates/perry-runtime/src/messaging.rs
@@ -5,9 +5,10 @@ use crate::closure::{js_closure_alloc, js_register_closure_arity, ClosureHeader}
 use crate::object::{self, ObjectHeader};
 use crate::string::{js_string_from_bytes, StringHeader};
 use crate::value::JSValue;
+use std::cell::Cell;
 use std::collections::{HashMap, VecDeque};
 use std::ptr::null_mut;
-use std::sync::atomic::{AtomicBool, AtomicPtr, Ordering};
+use std::sync::atomic::{AtomicPtr, Ordering};
 use std::sync::Mutex;
 
 type MessageChannelFactory = extern "C" fn() -> f64;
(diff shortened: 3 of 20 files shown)


proggeramlug:
Audited and merging. This closes #8530 and — more valuable than the fix — makes the class un-reintroducible.

**The gate can actually fail, which is the question worth asking of any PR that adds one.** Two corrections to my own first attempt at checking this, both worth recording:

1. My initial sabotage (regex-rewriting a converted latch back to `Once`) left the gate green — but that was my mangled edit, not a blind spot. Reverting the shape by hand produced something the matcher legitimately did not recognize as a registration site.
2. My first exit-code read said `rc=0` because I piped through `tail` and captured *its* status. The real number:

```
new gate vs pre-fix main:  exit 1, 22 violations reported
new gate vs fixed tree:    exit 0, 0 violations, 109 registration calls checked
```

So it goes red on the real defect and green on the real fix. It also catches every site I originally reported on #8530 — `tls.rs:282`, `worker_threads.rs:228/237/1470`, `ws.rs:86`, `zlib.rs:676` — plus four more I missed, including the `AtomicBool` variant that my `Once|OnceLock|Mutex<bool>` grep never looked for. 22 against my 18.

`--self-test` is non-vacuous: **4 bad shapes rejected, 4 good shapes accepted**. A matcher that only proves it accepts good input is worthless; this one proves both directions.

Wired into the required `lint` job as its own step ("Thread-local GC scanner registration audit"), with the self-test running first — so a broken matcher fails before a green policy result can be trusted.

**The split between per-thread and process-wide is the part I'd have got wrong.** Not every latch here was a bug: the cluster arity and streams callback setup are genuinely process-wide, and converting those to `Cell<bool>` would have made them re-run per thread. Separating them rather than mechanically rewriting all 18 files is the right call, and it is the kind of distinction a sweeping refactor usually flattens.

For the record on why this matters: `MUTABLE_ROOT_SCANNERS` is thread-local (`gc/roots.rs:105`), so a process-global latch means the first thread to touch a subsystem registers its scanner and consumes the latch — every later heap's collector then never scans those roots. An unregistered scanner throws nothing; it surfaces cycles later as a use-after-free somewhere unrelated. Green failure direction, which is why it survived across 18 subsystems until a census found it.

Stacked on current `main` with `check_file_size`, `workspace_architecture`, `raw_handle_debt` (974/974) and `gc_runtime_root_holders` all green.

PR #8553, opened 2026-08-21
fix(compile): require(<builtin>) implies stdlib, plus wasm-host test isolation (#8547)

### LINKED ISSUE #8547: require('http') links runtime-only, so stdlib builtins return undefined (ESM import works)

proggeramlug:
**`require('<builtin>')` linked runtime-only, so stdlib-backed builtins returned `undefined` (#8547) — re-landed with the test-isolation bug that defeated the first attempt.**

```
import * as http from 'node:http';   →  Linking (with stdlib)...   typeof createServer() === "object"
const http = require('http');        →  Linking (runtime-only)...  typeof createServer() === undefined
```

`ctx.native_module_imports` drives `needs_stdlib` and therefore the link mode, but it was populated only by the ESM import walk. A CommonJS `require("http")` never landed in it, so the link came out runtime-only, `perry-stdlib`'s `common/dispatch/init.rs` never ran, `JS_NATIVE_HTTP_DISPATCH` stayed null, and the `("http", "createServer")` arm returned `undefined`. Nothing about it was http-specific.

The lowered HIR cannot answer this: for `require('http')` the CJS shim emits a *runtime* dispatcher that switches over builtin names as string literals, and that switch contains a case for **every** builtin regardless of use — a static scan would match the table, not the call. The fix therefore reads literal call sites via the extractor that already exists for CJS wrapping (`cjs_wrap::extract_require_specifiers`).

Scope, measured — the change can only add stdlib linking for a program that genuinely references a stdlib-backed builtin:

| program | link mode | binary |
|---|---|---|
| `console.log(...)` only | runtime-only (unchanged) | 7.7 MB |
| `require("path")` (runtime-only module) | runtime-only (unchanged) | 7.8 MB |
| `require("http")` | with stdlib (was runtime-only) | 14 MB |

**Why the first attempt (#8548) was reverted, and what actually changes here.** It broke `issue_5247_property_read_source_location` with `undefined reference to perry_wasm_host_*` at link time. The cause was not the detection — it was a pre-existing test-isolation bug this change happened to expose:

`issue_5234_wasm_esm_import` is the only thing in the workspace that builds `perry-runtime` with `perry-runtime/wasm-host`, and it built into the **shared** `target/debug`. That replaces `libperry_runtime.a` with one that compiles `webassembly.rs` in, whose undefined `perry_wasm_host_*` references only `libperry_wasm_host.a` satisfies — and that archive is put on the link line only for programs that themselves reference `WebAssembly.*`. Any other suite in the same `cargo test --workspace` job that links runtime-only against the poisoned archive fails. Whichever build wrote the archive last decided the outcome, which is why this was latent until #8548 perturbed the ordering.

`issue_5234` now builds into a dedicated `target/perry-wasm-host-test`, mirroring what the compiler's own no-auto path already does for exactly this reason (`build_wasm_host_runtime` in `optimized_libs/no_auto.rs` uses `target/perry-wasm-host-runtime` so "the prebuilt libperry_runtime.a is not clobbered").

Verified together in one `cargo test` invocation, which is the configuration that failed before: `issue_4903_listen_callback_deferred` **0/2 → 2/2**, `issue_5247_property_read_source_location` **3/3**, `issue_5234_wasm_esm_import` **1/1**.

Closes #8547.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed CommonJS `require()` handling for built-in Node.js modules, ensuring standard library features such as `http` are correctly recognized and available at runtime.
  - Improved test isolation to prevent unrelated build artifacts from affecting WebAssembly-related validation.

- **Documentation**
  - Updated the documented release version to 0.5.1517.
  - Added changelog details covering the built-in module resolution fix and improved verification.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Perry is a native TypeScript compiler written in Rust that compiles TypeScript source code directly to native executables. It uses SWC for TypeScript parsing and LLVM for code generation.
 
-**Current Version:** 0.5.1516
+**Current Version:** 0.5.1517
 
 
 ## TypeScript Parity Status
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5522,7 +5522,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "perry"
-version = "0.5.1516"
+version = "0.5.1517"
 dependencies = [
  "anyhow",
  "base64 0.22.1",
@@ -5582,7 +5582,7 @@ dependencies = [
 
 [[package]]
 name = "perry-api-manifest"
-version = "0.5.1516"
+version = "0.5.1517"
 dependencies = [
  "perry-dispatch",
  "serde",
@@ -5590,7 +5590,7 @@ dependencies = [
 
 [[package]]
 name = "perry-audio-miniaudio"
-version = "0.5.1516"
+version = "0.5.1517"
 dependencies = [
  "cc",
  "libc",
@@ -5599,7 +5599,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen"
-version = "0.5.1516"
+version = "0.5.1517"
 dependencies = [
  "anyhow",
  "inkwell",
@@ -5616,7 +5616,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-arkts"
-version = "0.5.1516"
+version = "0.5.1517"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5624,7 +5624,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-glance"
-version = "0.5.1516"
+version = "0.5.1517"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5632,7 +5632,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-js"
-version = "0.5.1516"
+version = "0.5.1517"
 dependencies = [
  "anyhow",
  "perry-dispatch",
@@ -5641,7 +5641,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-swiftui"
-version = "0.5.1516"
+version = "0.5.1517"
 dependencies = [
  "anyhow",
  "perry-hir",
@@ -5649,7 +5649,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wasm"
-version = "0.5.1516"
+version = "0.5.1517"
 dependencies = [
  "anyhow",
  "base64 0.22.1",
@@ -5661,7 +5661,7 @@ dependencies = [
 
 [[package]]
 name = "perry-codegen-wear-tiles"
-version = "0.5.1516"
+version = "0.5.1517"
 dependencies = [
  "anyhow",
  "perry-hir",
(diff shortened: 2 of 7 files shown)


proggeramlug:
Landed via #8556, which carries this branch's commits with two mechanical fixes applied — this is a fork PR, so they could not be pushed here: the version bump reverted to `main` (the maintainer bumps at merge time) and the changelog fragment re-keyed from `8552-` to `8553-`, since 8552 is a different, already-merged PR.

Verified on the resolved tree: `cargo check --workspace --all-targets` clean, and `check_file_size`, `workspace_architecture`, `raw_handle_debt`, `check_gc_scanner_latches` and `check_test_registration` all exit 0.

The diagnosis is the good part: `ctx.native_module_imports` drives `needs_stdlib`, was populated only by the ESM walk, and a static HIR scan genuinely cannot substitute — the CJS shim's runtime dispatcher carries a case for every builtin, so a scan would match the table rather than the call site.

PR #8555, opened 2026-08-21
fix(runtime): snapshot regex captures before allocation

### LINKED ISSUE #8449: runtime: js_regexp_exec keeps the subject borrow (and its Captures) live across the result-array, capture-string and groups allocations
Follow-up to #8428 (fixed in #8446), same borrowed-heap-slice class as #8423.

## Problem

#8446 closed the *live* window in `js_regexp_exec` — the one whose collection
point is user JS (`ToLength(Get(R, "lastIndex"))` running a coercible
`lastIndex`'s `valueOf`). What it deliberately did **not** close is the
*allocation-point* window that follows.

After the match, `crates/perry-runtime/src/regex/exec.rs` keeps using `str_data`
(the `&str` into the subject's inline WTF-8 payload) and the `regex::Captures` /
`fancy_regex::Captures` borrowing it, across every one of these allocations:

* `crate::array::js_array_alloc(caps.len())` — the result array
* `js_string_from_str(m.as_str())` per capture — so capture *N+1*'s `as_str()`
  reads a borrow that capture *N*'s allocation could have invalidated
* `crate::object::js_object_alloc(0, 0)` + `js_object_set_field_by_name` — the
  named-capture `groups` object
* `set_exec_array_metadata(arr, str_data, …)` in the fancy arm — allocates the
  `"index"` key string, *then* copies `input` out of the borrow
* `set_exec_array_indices` / `set_exec_array_indices_fancy`
  (`regex/exec_array.rs`) — these allocate an array and a `[start, end]` pair per
  capture, and call `byte_index_to_utf16_index(str_data, …)` between the
  allocations

`crates/perry-runtime/src/regex/match_string.rs` (`js_string_match`, the
non-global fancy arm and the standard non-global arm) has the identical shape.

## Why this is latent rather than live

The only collection points in this window are the allocations themselves, and
`gc_check_trigger`'s nursery-churn arm calls `force_full_scan()` unconditionally
since #7682 — a conservative stack scan makes the copying minor ineligible, so
nothing moves. That is the same "closed by accident" property `string/alloc.rs`
documents for #7213 and #8423's write-up spells out, and it is exactly the
property the moving-GC roadmap keeps eroding. These sites should not depend on it.

## Suggested fix

The pattern already exists in this module: `regex/match_all.rs`'s
`materialize_match_all_results` runs a **Phase 1 (borrowing, no JS allocation)**
snapshot into owned Rust data, then a **Phase 2** that allocates only from the
snapshot (see `OwnedMatchAllData` and its doc comment, which names this exact
hazard from the 2026-07-09 audit wave 1). Give `exec` and `js_string_match` the
same two-phase shape.

Two sub-decisions worth measuring rather than guessing, which is why this is not
folded into #8446:

1. **Owned `String` per capture vs. byte ranges.** `match_all` pays a
   `to_string()` per capture. `exec` is hotter (tokenizers, routers, ajv), so the
   cheaper shape is to snapshot `Vec<Option<(byte_start, byte_len, utf16_len, flags)>>`
   in phase 1 and build each capture with `string::string_copy_range` — which
   roots its source and re-reads after the allocation — for zero extra copies.
   Needs `compute_utf16_len_wtf8` / `bytes_have_lone_surrogate` on the phase-1
   side so the WTF-8 flags stay exact.
2. **`set_exec_array_indices{,_fancy}` can be fixed in place** without touching
   `exec`'s structure: precompute every `byte_index_to_utf16_index` into a
   `Vec<Option<(f64, f64)>>` *before* the first `js_array_alloc`, then allocate
   from that. Self-contained and worth doing first.

Also fold in: the fancy arm's `.input` (`set_exec_array_metadata`) copies the
subject out of the borrow after allocating the `"index"` key; the standard arm
already re-boxes the rooted subject via `set_exec_array_metadata_groups_fresh`.

## Validation

- The witness shape is `gc::tests::runtime_roots::regexp_last_index` (added in
  #8446): plant the collection, assert the subject was live *and* moved, then
  assert the captures. For this issue the collection has to be injected at an
  *allocation* rather than in a callback — `force_next_general_arena_alloc_slow`
  + `GcTriggerThresholdTestGuard::make_arena_trigger_due` (see
  `runtime_roots/string_slice.rs`) with the alloc-point scan pinned off
  (`ConservativeScanDisabledGuard`), otherwise the minor is non-moving and the
  test is vacuous.
- Sabotage-check it: the test must fail against the pre-fix code, or it is
  proving nothing.
- Watch for a regression on regex-heavy workloads if option 1 lands as owned
  `String`s.

## Workflow

PR = code + tests + `changelog.d/<PR>-<slug>.md`; no version bump.

proggeramlug:
## Summary

- snapshot standard and fancy regex captures into owned byte ranges before any runtime allocation
- materialize exec and String#match results from the rooted subject via string_copy_range
- precompute UTF-16 capture indices and add moving-minor allocation-point regressions for both paths

## Validation

- CARGO_INCREMENTAL=0 CARGO_PROFILE_DEV_DEBUG=0 cargo check -p perry-runtime
- CARGO_INCREMENTAL=0 CARGO_PROFILE_DEV_DEBUG=0 cargo test -p perry-runtime gc::tests::runtime_roots::regexp_last_index --lib
- CARGO_INCREMENTAL=0 CARGO_PROFILE_DEV_DEBUG=0 cargo test -p perry-runtime regex::tests --lib
- allocation witness sabotage-checked against pre-fix upstream code (fails after the subject moves)
- raw-handle, unrooted-local, string-payload, GC-store, and file-size inventories pass

Fixes #8449

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved regular-expression result handling when garbage collection moves the input string during matching.
  * Fixed potential issues with `RegExp#exec` and `String.prototype.match`, including global, sticky, fancy-regex, capture-group, named-group, and indices results.
  * Preserved correct `lastIndex` behavior and match contents during allocation-triggering operations.
* **Tests**
  * Added regression coverage for subject movement during regular-expression execution and match-result creation.

### DIFF
--- a/changelog.d/8555-regex-owned-captures.md
+++ b/changelog.d/8555-regex-owned-captures.md
@@ -1,0 +1,14 @@
+**Fix moving-GC safety when materializing regular-expression matches.**
+
+`RegExp.prototype.exec` and non-global `String.prototype.match` kept the
+subject string payload and engine capture objects borrowed while allocating
+their result arrays, capture strings, named groups, metadata, and `d`-flag
+indices. A moving collection at any of those allocation points could relocate
+the subject and leave the remaining captures reading retired nursery bytes.
+
+Both the standard and fancy-regex paths now snapshot capture byte ranges and
+UTF-16 metadata before the first runtime allocation, then copy each capture
+from the current rooted subject with `string_copy_range`. Global
+`String.prototype.match` uses the same range-based materialization. Allocation-
+point GC regressions force the subject to move after matching and verify the
+standard `exec` and fancy `String#match` results remain intact.
--- a/crates/perry-runtime/src/gc/tests/runtime_roots/regexp_last_index.rs
+++ b/crates/perry-runtime/src/gc/tests/runtime_roots/regexp_last_index.rs
@@ -1,4 +1,4 @@
-//! Moving-GC regression for `RegExp.prototype.exec` (#8428).
+//! Moving-GC regressions for regex subject borrows (#8428, #8449).
 //!
 //! RegExpBuiltinExec step 4 is `ToLength(Get(R, "lastIndex"))`. When
 //! `lastIndex` holds an object the ToNumber half runs OrdinaryToPrimitive —
@@ -13,6 +13,10 @@
 //! the stomped bytes instead of the relocated subject and the match evaporates.
 //! Liveness is asserted both ways (a copying minor ran; the subject actually
 //! moved), per CLAUDE.md's "a gate must assert its subject was live".
+//!
+//! The #8449 witnesses plant the same moving minor at the first result-array
+//! allocation, after the regex engine has returned captures borrowing the
+//! subject. They cover the standard `exec` and fancy `String#match` paths.
 
 use super::super::super::*;
 use super::super::support::*;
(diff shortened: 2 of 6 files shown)


proggeramlug:
Landed via #8559 with the ratchet bookkeeping applied — this is a fork PR so it could not be pushed here.

Your change removes **49 bare raw-handle reads** (974 → 925) and empties `regex/exec.rs`'s ceiling. Both ratchets refused to pass until that was recorded — `debt fell by 49; run --update to lock it in`, and `ceiling of 16 matches nothing … DELETE its line so the cleanup cannot be undone`. Removed the stale line and set the baseline to 925.

Verified on the resolved tree: `cargo check --workspace --all-targets` clean, and `raw_handle_debt`, `string_payload_access_inventory`, `gc_runtime_root_holders`, `check_gc_scanner_latches`, `check_file_size`, `workspace_architecture` and `check_test_registration` all exit 0.

The approach is right for the defect class: #8449 is a borrow held across an allocation, and rooting cannot fix a borrow — only owning the bytes can. Snapshotting into owned byte ranges before any allocation and materializing from the rooted subject is the correct shape, and sabotage-checking the allocation witness against pre-fix code is what makes it a regression test rather than a snapshot.

PR #8560, opened 2026-08-21
fix(runtime): scope property read PIC descriptor gate

proggeramlug:
## Summary

Property-read PIC population currently stops process-wide after any accessor is registered. Replace that global latch with the existing receiver-local descriptor flag so unrelated plain objects can keep warming their read PICs without weakening accessor correctness.

## Changes

- Remove the process-wide `accessors_in_use` gate from property-read PIC population.
- Rely on `OBJ_FLAG_HAS_DESCRIPTORS` plus the receiver ShapeId transition; the emitted hit path independently checks the same descriptor flag.
- Add regressions proving an unrelated accessor cannot poison a plain receiver PIC and an accessor-bearing receiver cannot seed a raw-slot PIC.

## Related issue

n/a

## Test plan

- [x] `cargo build --release` clean
- [ ] (if CLI / stdlib / runtime API changed) Updated `docs/src/` (not applicable: internal runtime cache policy only)

Additional checks:

- `./scripts/test_affected_crates.sh --base origin/main`
  - `perry-runtime`: 2,614 passed, 4 ignored
  - Perry CLI: 1,015 passed
  - `perry-ffi`: 29 passed
- `git diff --check`

Performance validation used 11 alternating A/B pairs of a 10,000-entity ECS query kernel, with a fresh process and correctness check per sample:

- field accumulation median: 2.0912 ms/op -> 0.3224 ms/op (84.6% lower; 6.49x faster)
- read-only query median: 0.4358 ms/op -> 0.3950 ms/op (5.0% lower)
- all accumulation samples produced the exact expected sum

## Screenshots / output

n/a

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved property access performance for objects without accessors.
  * Prevented accessor properties on one object from disabling caching for unrelated objects.
  * Preserved correct handling for objects with accessor-based properties.
  * Improved ECS benchmark performance from 2.0912 ms/op to 0.3224 ms/op.

* **Tests**
  * Added regression coverage for property access caching behavior and related scenarios.

### DIFF
--- a/changelog.d/8560-scope-read-pic.md
+++ b/changelog.d/8560-scope-read-pic.md
@@ -1,0 +1,5 @@
+Fixed property-read inline caches so registering an accessor no longer disables
+cache warming for every unrelated plain object in the process. Descriptor-bearing
+receivers remain excluded by their object-local flag and ShapeId transition, while
+a 10,000-entity ECS accumulation benchmark improves from 2.0912 ms/op to
+0.3224 ms/op (6.49x faster).
--- a/crates/perry-runtime/src/object/field_get_set/ic_miss.rs
+++ b/crates/perry-runtime/src/object/field_get_set/ic_miss.rs
@@ -623,13 +623,6 @@ pub extern "C" fn js_object_get_field_ic_miss(
     if (obj as usize) < 0x10000 {
         return f64::from_bits(crate::value::TAG_UNDEFINED);
     }
-    // When accessors are active anywhere in the program, skip the cache
-    // entirely: the PIC fast path does a direct field load that bypasses
-    // getter dispatch, so any object that uses defineProperty / get / set
-    // would silently return the raw slot value instead of calling the
-    // getter. The slow path through js_object_get_field_by_name handles
-    // accessors correctly.
-    let can_cache = !crate::state::state().descriptors.accessors_in_use.get();
     unsafe {
         // Issue #72: validate this really is a GC_TYPE_OBJECT before reading
         // crate::object::object_keys_array(obj) — otherwise an Array/String/Buffer/etc. receiver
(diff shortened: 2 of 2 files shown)


proggeramlug:
Audited and merging. Removing a process-wide correctness latch is the kind of change where a wrong answer is silent — a PIC hit does a raw slot load, so a receiver with a getter would return the slot value instead of invoking it. So the review is the safety argument, not the 6.49x.

**The argument holds, and it is fail-closed on both sides.** The old gate was global: any accessor registered anywhere stopped PIC population for every object in the process. The replacement is the receiver's own `OBJ_FLAG_HAS_DESCRIPTORS` bit plus its ShapeId transition. Two independent checks:

- **population** — `ic_miss.rs` refuses to prime a PIC for a descriptor-bearing receiver
- **use** — the emitted hit path checks the same bit (`class_field_inline_guard.rs`, `element_shape_guard.rs`)

and `note_descriptor_target` (`descriptor_state.rs:414`) sets that bit on the GC header before an installed descriptor is observable. An unrelated object's accessor cannot affect an own data property on this receiver, which is what makes the narrowing sound rather than merely faster.

**The new regression test is load-bearing — I checked rather than assumed.** Disabling PIC population to reproduce the pre-fix shape:

```
test object::field_get_set::ic_miss::c3c_pic_tests::
  unrelated_accessor_does_not_poison_plain_receiver_pic ... FAILED
```

Clean tree: **19/19 passed**. So the test fails without the fix and passes with it, in both directions.

Also good that the second test covers the converse — an accessor-bearing receiver must not seed a raw-slot PIC. One direction alone would have left the dangerous half untested.

**Gates on the stack** (with #8562, #8563, #8565): `check_file_size`, `workspace_architecture`, `raw_handle_debt`, `check_gc_scanner_latches`, `gc_runtime_root_holders`, `check_test_registration` — all exit 0.

PR #8562, opened 2026-08-22
test(fetch): cover multi-app scanner registration

### LINKED ISSUE #8546: coop/in_process: second Next app dies with TypeError: value is not a function (GC rooting, multi-heap)
## Summary

With #8528 merged, a Coop `in_process` daemon can **load and initialize** two
Next.js applications in one process — but the **second** application's HTTP
handler rejects at dispatch, so it serves 500 while the first serves 200.

## Evidence

Measured on the quiet M1 mini, Perry `7c22189aa` (0.5.1514), Coop `in_process`,
two deployments of the same Next fixture. Provider provenance verified by
content, not by recorded metadata: the merged fix deletes
`ERR_PERRY_PATH_MODULE_THREAD`, and `strings` on the packaged
`libperry_runtime.dylib` finds **0** occurrences of it (the string also appears
nowhere in main's source, so the check is meaningful rather than vacuous).

Both applications initialize cleanly:

```
INFO application library preloaded on dedicated Perry thread deployment="bench-000" load_ms=... init_ms=...
INFO application library preloaded on dedicated Perry thread deployment="bench-001" load_ms=749.8 init_ms=697.0
```

Then:

```
bench-000 -> 200  {"runtime":"next","iterations":100,"checksum":3726872593}
bench-001 -> 500  coop: 500 deployment dispatch failed

ERROR deployment dispatch failed deployment=bench-001
      error=Runtime(invoking deployment HTTP handler: handler promise rejected: )
```

The rejection reason is **empty**, which is itself worth fixing — an empty
`handler promise rejected:` gives an operator nothing to act on.

## What this is NOT

- **Not the path-module registry.** Zero `rejected duplicate path-module
  initializer` messages in this run; both apps register their own initializers
  on their own dedicated Perry threads. (An earlier iteration of #8528 that kept
  the initializer-address map process-wide DID produce 115 such rejections;
  that is fixed.)
- **Not app load or module init** — both complete, with timings logged.
- **Not ordering-dependent in an obvious way**: the first deployment to be
  dispatched succeeds and the second fails.

## Hypothesis

Some other process-global runtime/stdlib state assumes a single application per
process. Until #8528 the fail-closed thread guard made multi-app hosting
impossible, so none of that state has ever been exercised by two heaps. The
per-request registries are the obvious suspects — several are `Mutex<HashMap<usize, ...>>`
and are **UNCOVERED** by `scripts/gc_runtime_root_holders.py`, i.e. nobody has
determined whether they hold heap pointers:

`REQUEST_REGISTRY`, `BLOB_REGISTRY`, `FETCH_RESPONSES`, `TRANSFORM_PAIRS`,
`STREAM_HANDLES` (`crates/perry-stdlib/src/fetch/mod.rs`,
`streams/transform.rs`), plus `buffer/own_props.rs`'s `PROPS` and
`buffer/header.rs`'s `EXTERNAL_*_REGISTRY` (the latter keyed by RAW heap
address).

## Suggested first steps

1. Make the rejection reason non-empty — surface the underlying JS error.
2. With that, name the failing operation, then check whether its backing table
   is process-global.

## Reproduction

Needs a Coop checkout whose `perry-main.lock` points at a Perry with #8528, plus
`coop-daemon`'s `rusqlite` bumped 0.37 -> 0.39 to match `perry-stdlib` (they link
the same native `sqlite3`; see the companion Coop change). Then
`scripts/prepare-next-benchmark.sh` and the `resource_benchmark` harness with
`COOP_BENCH_APP_COUNTS=2`, `COOP_BENCH_EXECUTION_MODE=in_process`.

proggeramlug:
## Summary

- add a current-thread scanner-count diagnostic so provider tests can inspect the thread-local FFI mutable-root registry without fabricating heap pointers
- model Coop's two dedicated Perry application threads and assert that each installs exactly one Fetch scanner
- document the multi-app Next dispatch fix without a version bump

## Context

While #8546 was being investigated, #8552 landed the per-thread scanner-latch correction for the underlying bug class. It was not linked to #8546 and had only a static latch audit. This adds behavioral coverage for the exact Fetch path implicated here: with the old process-global `Once`, application 1 registers successfully and application 2 sees an empty scanner registry, leaving its `Request`, `Headers`, and `FormData` roots invisible to that heap's collector.

The regression was sabotage-verified by temporarily restoring the old `Once`: it fails deterministically with `application 2 did not install the Fetch root scanner` (`left: 0`, `right: 1`). Restoring the per-thread latch makes it pass.

## Testing

- `cargo +nightly-2026-08-20 test -p perry-stdlib --no-default-features --features web-fetch fetch_root_scanner_registers_for_each_application_thread --lib -- --nocapture`
- `python3 scripts/check_gc_scanner_latches.py`
- targeted `rustfmt --check` and `git diff --check`

No version bump; `Cargo.toml` and `Cargo.lock` are unchanged.

Fixes #8546

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Fetch functionality in embedded multi-application environments so each application properly initializes its required data scanners.
  * Prevented Fetch request and header data from becoming unavailable to applications after the first one.

* **Tests**
  * Added regression coverage confirming scanner registration works independently for multiple application threads.

### DIFF
--- a/changelog.d/8546-multi-app-fetch-scanner.md
+++ b/changelog.d/8546-multi-app-fetch-scanner.md
@@ -1,0 +1,8 @@
+### Fixed
+
+- Preserve Fetch `Request`, `Headers`, and `FormData` registry roots for every
+  application thread in an embedded multi-app host. Previously a
+  process-global registration latch could install the Fetch scanner only for
+  the first Perry heap, so a second in-process Next application could reject
+  its HTTP handler and return 500. Add a two-application-thread regression test
+  for the per-thread scanner contract. (#8546)
--- a/crates/perry-runtime/src/gc/roots.rs
+++ b/crates/perry-runtime/src/gc/roots.rs
@@ -192,6 +192,19 @@ pub(super) fn copy_only_root_scanner_counts() -> (usize, usize) {
     (rust_scanners, ffi_scanners)
 }
 
+/// Return the number of named native-package mutable-root scanners registered
+/// for the current Perry thread.
+///
+/// The registry is deliberately thread-local: an embedding host may run
+/// multiple independent Perry heaps in one process, and each heap's collector
+/// must see the scanners installed by its stdlib provider. This small
+/// diagnostic surface lets provider tests verify that contract without running
+/// a collection over fabricated heap pointers.
+#[doc(hidden)]
+pub fn gc_named_ffi_mutable_root_scanner_count() -> usize {
+    FFI_NAMED_MUTABLE_ROOT_SCANNERS.with(|scanners| scanners.borrow().len())
+}
+
 pub(super) fn registered_root_scanners_block_budgeted_gc() -> bool {
     let has_copy_only = ROOT_SCANNERS.with(|scanners| !scanners.borrow().is_empty())
         || FFI_ROOT_SCANNERS.with(|scanners| !scanners.borrow().is_empty());
--- a/crates/perry-stdlib/src/fetch/tests.rs
+++ b/crates/perry-stdlib/src/fetch/tests.rs
@@ -1,5 +1,42 @@
 use super::*;
 
+/// #8546: Coop hosts each in-process deployment on its own dedicated Perry
+/// thread. The Fetch scanner registry is thread-local, so a process-global
+/// registration latch makes the first Next application safe and leaves the
+/// second application's Request / Headers roots invisible to its collector.
+///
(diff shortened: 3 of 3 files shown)


proggeramlug:

| check | result |
|---|---|
| `check_file_size.sh` | 0 |
| `raw_handle_debt.py` | 0 |

PR #8535, opened 2026-08-21
feat(compat): add native TypeScript transpileModule subset

proggeramlug:
## Summary

- register a partial well-known `typescript` binding backed by native SWC transforms
- support OpenCode Code Mode’s `transpileModule`, compiler enums, TS/TSX handling, and diagnostic flattening
- reject compiler APIs and options outside the audited subset explicitly
- cover the exact OpenCode import/call shape with a compiled end-to-end regression test

Closes #8511

## Testing

- `cargo test -p perry-ext-typescript --lib`
- `cargo test -p perry --test issue_8511_typescript_transpile_module`
- `cargo test -p perry-hir typescript_transpile_subset_lowers_to_native_dispatch_and_enums`
- `cargo test -p perry-hir --test unimplemented_api_check typescript_compiler_api_outside_native_subset_is_rejected`
- `cargo test -p perry-codegen --test manifest_consistency`
- `cargo test -p perry-api-manifest`
- `cargo clippy -p perry-ext-typescript --all-targets --no-deps -- -D warnings`
- `node scripts/binding_pins.mjs --check`
- `python3 scripts/check_test_registration.py`
- `./scripts/pre-tag-check.sh --quick`

No version bump.

## Summary by CodeRabbit

* **New Features**
  * Added partial TypeScript runtime compatibility through `transpileModule` and `flattenDiagnosticMessageText`.
  * Supports TypeScript and TSX transpilation, React JSX lowering, compiler option validation, and diagnostics.
  * Added runtime enum constants for `ScriptTarget`, `ModuleKind`, and `DiagnosticCategory`.
* **Documentation**
  * Added TypeScript APIs to the public API declarations and reference documentation.
* **Bug Fixes**
  * Added coverage confirming supported APIs work and unsupported TypeScript APIs are reported clearly.

PR #8564, opened 2026-08-22
docs: regenerate API reference + .d.ts (check is red on main)

proggeramlug:
**Regenerate the API docs — `check` is red on `main` because they drifted.**

`check` fails on `main` at its "Check for API docs drift" step:

```
API docs drift detected. The compile-time manifest in
crates/perry-api-manifest/src/entries.rs changed but the
generated artifacts under docs/ weren't regenerated.
```

`crates/perry-api-manifest/src/entries.rs` gained the `@parcel/watcher` compatibility facade (#8532) plus entries from #8535 and #8525 without `./scripts/regen_api_docs.sh` being rerun. This is that script's output only, no hand edits: coverage moves from 2033 entries across 124 modules to 2051 across 134.

Verified: `check` goes from **fail to pass** with this change (it is red on `main` and on this PR's first push, green after).

## Summary by CodeRabbit

* **Documentation**
  * Regenerated API documentation and declarations.
  * Expanded API coverage to include 2,053 entries across 134 modules.
  * Added reference documentation for file-watching, TypeScript, Bun, and worker APIs.
  * Updated the API reference to report 2,990 entries across 136 modules.
  * Verified documentation consistency checks now pass.


proggeramlug:
Holding — the diagnosis was correct when written, but the premise no longer holds, and one part of it is factually wrong.

**Wrong part.** The description says `float_algebraic` "needs `#![feature(float_algebraic)]` regardless of channel" and so fails "on stable **and** on the nightly that same commit pinned." The nightly half does not reproduce. On current `main` (`ef5e5fa84`) with `rustc 1.100.0-nightly (f7d782a3b)`:

```
$ cargo check -p perry-runtime --all-targets
exit=0   E0658 errors: 0
```

`perry-runtime` declares no `#![feature(float_algebraic)]` and compiles anyway, test targets included. So the methods are usable on the pinned nightly without a gate.

**Stale part.** Your reproduction was against `4ee1d04b5`, before #8561 added `rust-toolchain.toml`. At that commit a plain `cargo` resolved to the default toolchain — stable here — and E0658 was exactly right; I hit the same wall trying to build for a benchmark sweep, which is what prompted #8561. Now that the toolchain file is in the tree, `cargo` resolves `nightly-2026-08-20` automatically and the build is clean.

**Why I would not take this now.** Dropping the `algebraic_*` calls removes a real (if small) reassociation win, and it does not let the project leave nightly: `.cargo/config.toml`'s `[unstable] min-publish-age` is nightly-only (rust-lang/cargo#17009), so nightly is required regardless of this code. The trade would be a lost optimization for no reduction in toolchain requirements.

Your closing note is the right instinct though, and it is worth acting on independently: an explicit `#![feature(float_algebraic)]` in `perry-runtime` would make the nightly dependency **declared rather than incidental**. Right now the crate compiles on the pinned nightly by luck of what that nightly happens to have stabilized — if the pin moves to a nightly where the gate is required again, it breaks silently. A one-line gate plus a comment would pin the intent. Happy to take that version of this PR.

Also worth noting for whoever picks it up: this bumps the workspace version, and its changelog fragment is keyed `8560-` rather than to this PR.

proggeramlug:
Narrowed this PR — dropping the `perf_histogram` change I originally opened it for, and keeping only the API docs regeneration.

**What I got wrong.** I reproduced `error[E0658]: use of unstable library feature `float_algebraic`` locally and concluded it was what took `cargo-test` down on `main`. It was not. `main`'s gap-suite compiles and passes, so `algebraic_*` is available on the nightly #8550 pinned; my local stable 1.97.1 simply cannot build it, and this repo has no `rust-toolchain` file to have told me so. Replacing those calls with plain arithmetic therefore fixed nothing and changed the stddev numerics — which is why `gap-suite (2)` went from green on `main` to red here. Reverted.

That is the second time in this session I have acted on a local reproduction that did not match CI, so I want it stated rather than quietly amended.

**What survives, and is independently verified**: the API docs drift. `check`'s "Check for API docs drift" step is red on `main` — `entries.rs` gained the `@parcel/watcher` facade (#8532) plus entries from #8535 and #8525 without `regen_api_docs.sh` being rerun. `check` failed on this PR's first push and passed once the regenerated artifacts were included, which is the before/after that makes the claim checkable. Generator output only, no hand edits.

`main`'s `cargo-test` failure is therefore still unexplained by me; its log shows no compile error or test failure, so it may be a timeout or infrastructure fault rather than a code defect. I did not want to leave a misattribution standing while that is still open.

Issue #8549, opened 2026-08-21
e2e-scoped: any_call_result_trim_emits_string_tag_dispatch has been failing on main since ~2026-08-10, reddening unrelated PRs


proggeramlug:
## Summary

`crates/perry-codegen/tests/argless_builtin_extra_args.rs::any_call_result_trim_emits_string_tag_dispatch` **fails on clean `main`** and has done for at least nine days. It is currently reddening `e2e-scoped` on every open PR that touches `perry-codegen`, where it reads as that PR's fault.

## Proof it is not any PR's fault

Fresh worktree at `origin/main` (`7c22189aa`), nothing applied:

```
failures:
    any_call_result_trim_emits_string_tag_dispatch
test result: FAILED. 3 passed; 1 failed; 0 ignored
```

The assertion is:

```rust
assert_eq!(ir.matches("= call double @perry_fn_schema_ts__make(").count(), 2,
    "the receiver must be evaluated once in the init body (plus the generated extern wrapper)");
```

**Observed count is 1, expected 2.** The fixture is `make().trim()` where `make` is an imported extern returning `Any`.

## Bisect so far

| commit | date | result |
|---|---|---:|
| `debefa55c` | 2026-08-10 | **PASS** |
| `8bfdac447` | 2026-08-19 | FAIL |
| `6173ff9f6` | 2026-08-20 | FAIL |
| `8d1cca158` … `7c22189aa` | 08-20/21 | FAIL |

497 commits in the good→bad window. An automated `git bisect run` was started and paused when free disk fell to 7.7 GiB — ENOSPC manufactures false failures, so a paused bisect is better than a wrong answer. It needs roughly nine build-and-test rounds with ~10 GiB headroom.

## Why nobody noticed

`crates/*/tests/*.rs` run per-PR only when the diff names them (`e2e-scoped`, see `scripts/ci_plan.py`). A break in this suite is therefore invisible on `main` and only surfaces on whichever PR happens to touch `perry-codegen` — wearing that PR's name.

This has already cost real time: it was mis-attributed to #8540, and an agent was dispatched to "fix" a receiver-evaluation bug that PR did not introduce. #8537, #8543 and #8525 carry the same inherited red.

## Why it matters beyond CI noise

[...]

PR #8566, opened 2026-08-22
fix(codegen,manifest): close pre-existing e2e-scoped drift from #8383/#8450/#8527/#8537

jdalton:
## Summary

Follow-up to [#8550](https://github.com/PerryTS/perry/pull/8550), which merged before these CI fixes landed on it. `e2e-scoped` was already failing on `main` for reasons unrelated to either PR (confirmed via `git bisect`) — it only runs when a diff touches `crates/*/tests/*.rs`, so it went unnoticed on ordinary pushes.

<details>
<summary>What was broken and why</summary>

- **`manifest_consistency`**: [#8527](https://github.com/PerryTS/perry/pull/8527) added a `worker_threads::reload` dispatch row, and [#8537](https://github.com/PerryTS/perry/pull/8537) added `bun::Glob` / `bun::unsupported` rows, none with matching `API_MANIFEST` entries. Added the three missing entries.
- **`argless_builtin_extra_args`, `native_proof_regressions`**: [#8383](https://github.com/PerryTS/perry/pull/8383) intentionally dropped the unconditional per-consumer-module `__perry_wrap_extern_*` value wrapper and made inline-cache globals module-prefixed (to avoid cross-module symbol collisions). Both tests had stale IR-shape assertions from before that change; updated to match the new, correct shape.
- **`loop_safepoint_purity`**: the same PR tightened module-global-export detection to require a matching `Export::Named` entry instead of trusting `exported_objects` alone. The test fixture never populated `exports`, so its `sum` accumulator silently stopped becoming a module global, and the assertion it depends on stopped proving anything. Fixed the fixture.
- **`temp_root_operand_temporaries`**: [#8450](https://github.com/PerryTS/perry/pull/8450) replaced `String.prototype.concat`'s single accumulator-with-write-back rooting with per-operand rooted slots feeding a fused `js_string_concat_chain` call. Rewrote the assertion to check the new shape.

</details>

## Test plan

- [x] All 5 previously-failing `perry-codegen`/`perry-api-manifest` tests pass locally under the pinned nightly toolchain.
- [x] Full `perry-codegen` test suite passes (one unrelated local sandbox failure on an unrelated pre-existing test, not touched here).
- [ ] CI `e2e-scoped` / `pr-gate` green.

## Summary by CodeRabbit

- **New Features**
  - Added support for `worker_threads.Worker.reload`.
  - Added API support for `bun.Glob` and `bun.unsupported`.

- **Bug Fixes**
  - Improved reliability of module exports, string concatenation, temporary values, and cross-module caching.
  - Updated behavior for imported function evaluation.

- **Documentation**
  - Updated the documented release version to `0.5.1517`.
  - Added changelog notes covering resolved end-to-end issues.

- **Tests**
  - Refreshed regression coverage for runtime and code-generation behavior.


proggeramlug:
Merging. This closes the `e2e-scoped` breakage I filed as #8549, and it resolves the one thing I explicitly warned against — correctly.

**My warning on #8549 was: do not fix this by editing the expected count.** The assertion pins how many times a *side-effecting* receiver is evaluated, so moving the number to match the observed IR would have converted a semantic guarantee into a snapshot.

This PR changes `2` → `1`, but the **invariant is unchanged**:

> the receiver must be evaluated **exactly once**, in the init body — #8383 dropped the unconditional per-consumer-module `__perry_wrap_extern_*` value wrapper (imported function values now materialize lazily via `js_closure_alloc_singleton`), so there is no second call site to account for

Only the wrapper accounting moved. "Exactly once" still holds, and the message now records *why* one is the right number — which is what I asked for and what makes this a correction rather than a moved goalpost.

**Verified all five previously-failing targets:**

| test | result |
|---|---|
| `argless_builtin_extra_args` | 4 passed |
| `native_proof_regressions` | **274 passed** |
| `loop_safepoint_purity` | 8 passed |
| `temp_root_operand_temporaries` | 19 passed |
| `perry-api-manifest` | ok |

**The `loop_safepoint_purity` finding is the most valuable thing here** and deserves more than a bullet: the fixture never populated `exports`, so after #8383 tightened module-global-export detection its `sum` accumulator silently stopped becoming a module global — and the assertion depending on that stopped proving anything. That is a test that was passing while testing nothing. Finding and repairing it is worth more than the four assertion updates combined.

Also correct on the structural point behind #8549: this suite runs per-PR only when a diff touches `crates/*/tests/*.rs`, so a break here is invisible on ordinary `main` pushes and surfaces later wearing an unrelated PR's name. It cost a mis-attribution to #8540 and an agent dispatched to fix a bug that PR did not introduce. A `main`-only unconditional sweep of the integration suites would have caught this on day one — worth doing as a follow-up.

PR #8563, opened 2026-08-22
deps: port node-forge to x509-cert 0.3

proggeramlug:
## Summary

Ports the native node-forge certificate builder and stdlib X509 parser to `x509-cert` 0.3 and its DER/SPKI 0.8 stack. The port preserves caller-controlled issuer names, exact validity bounds, extension criticality, and the no-implicit-extensions behavior used by Socket Firewall TLS certificates.

## Changes

- Add a custom `BuilderProfile` equivalent to the removed `Profile::Manual` variant.
- Migrate validity, name/RDN construction, extension criticality, certificate builder, and public accessor APIs.
- Align the node-forge RSA/SHA/RNG stack with `signature` 3 as required by the new builder.
- Update the stdlib X509 reader to use the new public certificate/name accessors while retaining SPKI 0.7 for jsonwebtoken compatibility.
- Extend regression coverage for exact validity, distinct leaf issuer DNs, and absence of profile-injected extensions.

## Related issue

Closes #8436

## Test plan

- [x] `CARGO_INCREMENTAL=0 CARGO_PROFILE_TEST_DEBUG=0 cargo +nightly-2026-08-20 test -p perry-ext-node-forge` (8 unit tests and the OpenSSL CA/leaf end-to-end test pass)
- [x] `CARGO_INCREMENTAL=0 CARGO_PROFILE_DEV_DEBUG=0 cargo +nightly-2026-08-20 check -p perry-stdlib --lib --no-default-features --features crypto`
- [x] `CARGO_INCREMENTAL=0 CARGO_PROFILE_DEV_DEBUG=0 cargo +nightly-2026-08-20 clippy -p perry-ext-node-forge --all-targets` (passes; existing warnings remain in dependency crates)
- [x] `cargo +nightly-2026-08-20 fmt -- crates/perry-ext-node-forge/src/crypto.rs crates/perry-stdlib/src/crypto/x509.rs --check`
- [ ] Full workspace release build/test not run locally
- [x] Added or updated tests in the affected crate
- [x] Documentation update not required; no user-facing API changed
- [x] Platform UI build not applicable

Note: adding `-D warnings` to the scoped Clippy command currently stops on a pre-existing `clippy::type_complexity` warning in `crates/perry-runtime/build.rs:475`.

## Screenshots / output

Not applicable.

## Checklist

- [x] I have NOT bumped the workspace version or edited CLAUDE.md / CHANGELOG.md
- [x] My commit follows the repository prefix convention
[...]

PR #8567, opened 2026-08-22
fix(zlib): route captured exports to external backend

proggeramlug:
## Summary

- add the external zlib native-module dispatcher so captured exports, including util.promisify(zlib.gzip), reach perry-ext-zlib
- add the missing async raw/unzip exports and keep synchronous external stream dispatch on the auxiliary event pump
- avoid allocator-crossing temporaries in stripped wrapper builds and retire the test_parity_zlib known failure
- no version bump

## Testing

- RUSTUP_TOOLCHAIN=nightly-2026-08-20 ./run_parity_tests.sh --filter test_parity_zlib
- cargo +nightly-2026-08-20 test -p perry-ext-zlib external_native_dispatch_routes_async_gzip_callback --lib
- cargo +nightly-2026-08-20 check -p perry-stdlib --no-default-features --features external-zlib-pump
- git diff --check

Refs #8271

### DIFF
--- a/crates/perry-ext-zlib/src/lib.rs
+++ b/crates/perry-ext-zlib/src/lib.rs
@@ -293,10 +293,181 @@ pub unsafe extern "C" fn js_zlib_inflate(data_value: f64, callback_value: f64) {
     stream::queue_one_shot_callback(data_value, callback_value, "Inflate", inflate_bytes);
 }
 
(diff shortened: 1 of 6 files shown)


proggeramlug:

| check | result |
|---|---|
| `check_file_size.sh` | 0 |
| `raw_handle_debt.py` | 0 |

PR #8568, opened 2026-08-22
perf(runtime): write Array#join directly into GC strings

proggeramlug:
## Summary

Writes `Array.prototype.join` output directly into GC-managed `StringHeader` storage instead of assembling a Rust `String` and copying it into the GC heap afterward.

The common dense, well-formed path computes the exact size and performs one destination allocation. The fallback keeps the receiver, separator, converted values, and partial result rooted across allocating getters/coercions and capacity growth. Element coercions still happen once and in source order.

## Changes

- Move join into a dedicated runtime module and add an exact-size direct-write path for strings, SSO strings, booleans, holes, `null`, and `undefined`.
- Add a rooted growable writer for coercing/exotic cases, re-reading sources after every possible collection.
- Preserve WTF-8 lone-surrogate flags and canonicalize surrogate pairs that become adjacent across empty fields/separators.
- Extend the issue fixture with forced moving-GC parity for holes, astral strings, lone surrogates, receiver mutation, exactly-once allocating `toString`, and fallback growth.

## Related issue

Addresses the `Array.prototype.join` portion of #8434.

## Performance

Paired Apple Silicon sweep over 16, 256, 4,096, and 16,384 short strings (`hyperfine`, 5 warmups, 30 runs, identical checksum):

| | Before | After |
|---|---:|---:|
| Wall time | 67.9 ms ± 0.7 | 49.0 ms ± 0.4 |
| User CPU | 61.0 ms | 42.8 ms |
| System CPU | 4.9 ms | 4.5 ms |

The patched binary was **1.38 ± 0.02× faster**.

A retired-instruction total is not available on this host: Xcode's `CPU Counters` CLI template exposes normalized bottleneck samples rather than retired instructions, while `Processor Trace` reports that this Apple CPU is unsupported. I have not relabeled either metric as an instruction count.

## Test plan

- `RUST_TEST_THREADS=1 cargo +nightly-2026-08-20 test --release -p perry-runtime --lib` — 2,616 passed, 4 ignored.
- `cargo +nightly-2026-08-20 check -p perry-runtime --lib`
- `cargo +nightly-2026-08-20 fmt --all -- --check`
- `python3 scripts/raw_handle_debt.py --no-raise-vs upstream/main` — no debt increase; new module has zero bare reads.
- `./scripts/pre-tag-check.sh --quick` — all quick repository gates passed.
- Rebuilt the `perry-dev` runtime static archive before compiling `test_issue_8434_array_join_gc.ts`.
- Compared the fixture byte-for-byte with Node under `PERRY_GC_SCHEDULE_RATE=1`, `PERRY_GC_FORCE_EVACUATE=1`, `PERRY_GC_VERIFY_EVACUATION=1`, and `PERRY_GC_PROTECT_FROMSPACE=1` — exact match; 300 copying collections moved 15,987 objects.

- [ ] Full workspace test command passes
- [x] Added or updated a user-facing test under `test-files/` and runtime unit tests
- [ ] Runtime API docs updated (not applicable; no API change)

## Checklist

- [x] My commit follows the repository's commit prefix convention

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `Array.prototype.join` handling for holes, empty values, Unicode, surrogate pairs, and lone surrogates.
  * Preserved correct behavior when element conversion triggers allocations or custom `toString()` methods.

* **Performance**
  * Optimized dense array joins to reduce intermediate allocations and improve string construction efficiency.


proggeramlug:
Cross-referencing with #8571, which measured a "direct-write join" candidate at **6.83x slower** on the same 16/256/4096/16384 short-string sweep where this PR measured **1.38x faster**. Both measurements were honest — they measured different algorithms under one label. #8571's rejected candidate (its pre-rebase implementation survives at `ae50c92e7`) coerces every element through a per-element `RuntimeHandleScope` + `js_jsvalue_to_string`, publishes each result into a barriered GC scratch array, and accumulates a per-piece descriptor `Vec` (one per separator repetition included); this PR's well-formed path is one unrooted sizing pass plus one allocation plus memcpys. The 9x gap is the cost of per-element rooting and scratch publication, not of "direct write".

I re-measured all four arms under one harness — interleaved medians, M1 Max, same checksummed output; instructions via macOS task counters:

| join (strings sweep) | wall | instructions |
|---|---:|---:|
| main | 682.6 ms | 6.536 B |
| main + #8571 (buffered) | 676.7 ms (0.99x) | 6.527 B |
| main + #8568 (direct) | 466.6 ms (0.68x) | 4.855 B (−25.7%) |
| main + rejected candidate (`ae50c92e7`) | 10 016 ms (14.7x) | 88.26 B (13.5x) |

This PR's join claim **holds** on strings. On a numbers-only sweep, however, it regresses 2.63x wall / 2.70x instructions vs main (9.27 B -> 25.06 B): `well_formed_join_size` rejects numbers, and the builder then pays a handle scope + GC string allocation per element in `element_to_string`. A strings-only fixture cannot see that shape — an easy fixture to reach for, and the benchmark itself was honest — but `[1,2,3].join(",")` is one of the most common joins there is, so it would likely have surfaced later as an unexplained corpus regression.

Two defects found while reviewing, one correctness and one performance:

**1 (perf, large).** The builder path pays a per-element `RuntimeHandleScope` + GC string allocation for **numeric** elements (`element_to_string`), where the buffered join formatted into its Rust buffer. On a numbers-only 16/256/4096/16384 sweep this PR retires **2.7x** the instructions of `main` (9.27B → 25.06B) — a real-world regression the strings-only benchmark cannot see, since `well_formed_join_size` rejects numbers and every element takes the slow arm. Fixed in the combined PR: numbers format on the stack (`itoa`) or via the shared `js_format_f64`, appended directly — number formatting can run no user code, so no rooting is needed.

**2 (correctness).** the exact-size path assumes `byte_len == utf16_len` for SSO elements (`well_formed_join_size`, `append_sso`). SSO length counts **bytes**, and Perry's JSON parser emits SSO values for non-ASCII payloads (`crates/perry-runtime/src/json/parser.rs`, unconditional `try_short_string`), including WTF-8 lone-surrogate halves from escapes. For e.g. `JSON.parse('["é","é"]').join("-")` the result header records `utf16_len` too large (and can miss lone-surrogate flags) — wrong `.length` and downstream slicing. To be explicit about the failure mode: this is **metadata corruption, not memory unsafety** — `utf16_len <= byte_len` always, so the byte-exact allocation is never short and nothing overruns; the hazard is a silently wrong `.length`, which is harder to notice than a crash. (An earlier internal read of this called it an overrun; that was wrong.) https://github.com/PerryTS/perry/pull/8581 carries this PR's join with that case routed to the canonicalizing builder path, plus regression tests, so this lands rather than being discarded.

PR #8569, opened 2026-08-22
fix(stdlib): drive active HTTP servers on fast waits

### LINKED ISSUE #8381: test: Next App Route dylib provider host hangs during cold start (route itself serves 21 requests)
## Symptom

`tests/test_next_app_route_dylib.sh` hangs indefinitely. The provider host builds
and links fine, prints its ABI hash, and then never serves a request — the
harness waits until it is killed. CI shows this as the
`Production App Route provider gate` step failing (e.g. run 31993068277 on
`main`).

Reproduced locally three times today, on three different trees:

| tree | stack at hang |
|---|---|
| `main` + #8313 | `main → js_wait_for_event` |
| `main` + #8377 | `main → run_microtasks → js_timer_tick → aux_has_active → js_node_http_server_has_active` |

Both are the same shape: **0.0% CPU, no child process, parked in the event-loop
pump**, last output `Provider ABI hash: …`. The second stack is the more
informative one — the host is asking whether the HTTP server has active handles
and getting an answer that keeps it idle, so it never begins serving.

## What it is NOT

- **Not #8258 (cwd-dependence).** That was a real, separate defect — Cargo reads
  `.cargo/config.toml` from the invocation directory, not `--manifest-path`, so
  the provider runtime lost `-C force-unwind-tables=yes`. #8377 fixed it, and the
  hang persists unchanged afterwards.
- **Not a stray-`node_modules` resolution problem.** A worktree placed under
  `/private/tmp` does fail differently and fatally
  (`JavaScript runtime (V8) support has been removed` on
  `/private/tmp/node_modules/@opentelemetry/api`, because Node resolution walks
  up out of the fixture), but from a normal worktree root that import is only
  `Warning: Could not resolve` and the build proceeds to the hang above.

## Why it matters

This is the acceptance gate for the #8034/#8040 production Next App Route, and it
is the natural end-to-end check for object-layout and ABI changes — exactly the
class that has silently broken split modules before (#8204 → #8228/#8241). While
it hangs, layout-changing PRs lose their most realistic verification: #8313 (the
40-byte header) had to fall back to the gap suite for that reason.

## Suggested starting point

`js_node_http_server_has_active` / `stdlib_pump::aux_has_active` — establish
whether the server handle is registered at all in the dylib-provider
configuration, or whether it is registered but reports inactive so the pump never
dispatches it.

proggeramlug:
## Summary

Keep external HTTP servers progressing when JavaScript continuously selects the event wait driver's microtask fast path. This removes the race that left the production Next App Route provider host parked before its first request, and bounds the gate's readiness probes so a recurrence produces diagnostics instead of hanging forever.

## Changes

- Treat a listening `perry-ext-http` server as native work in `stdlib_fast_drive`, giving its Tokio accept loop a bounded reactor turn during continuous JS microtask churn.
- Add a focused async-bridge regression test for the fast-drive liveness decision.
- Add `curl --max-time 1` to the provider gate's readiness probe so an accepted-but-unserved connection cannot wedge the harness.
- Document the root cause in `changelog.d`.

The unpatched retained-artifact repro failed readiness in 9/10 cold starts. With the fix, the rebased acceptance run became ready and completed the 21-request verifier.

## Related issue

Closes #8381

## Test plan

- [x] `cargo test --profile perry-dev -p perry-stdlib active_http_server_keeps_the_fast_wait_path_driving_native_tasks --lib`
- [x] `cargo check --profile perry-dev -p perry-stdlib --no-default-features --features async-runtime,external-http-server-pump`
- [x] `PERRY_NEXT_COLD_STARTS=1 PERRY_NEXT_VERIFICATIONS_PER_START=1 tests/test_next_app_route_dylib.sh` (cached compiler/provider targets; production Node oracle also passed during the initial run)
- [x] `bash -n tests/test_next_app_route_dylib.sh`
- [x] `shellcheck tests/test_next_app_route_dylib.sh`
- [x] `git diff --check upstream/main...HEAD`
- [ ] Full compile lint tier: not run because the workspace volume had only ~3 GiB free.
- [ ] Full script lint tier: 51/52 passed; the only failure is the pre-existing upstream-wide `raw_handle_debt.py` finding in `crates/perry-runtime/src/object/field_get_set/ic_miss.rs`. Its diff-aware `--no-raise-vs upstream/main` gate passed.
- [ ] `cargo build --release` clean (not run)
- [ ] Full affected-crate suite (not run; focused stdlib test above passes)
- [x] Regression coverage added in the affected crate
- [x] Docs update not applicable; no public API changed

## Checklist

- [x] I have NOT bumped the workspace version or edited `CLAUDE.md` / `CHANGELOG.md`
- [x] My commit follows the repository's `fix:` convention

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved application startup liveness when external HTTP servers are listening.
  - Ensured HTTP server activity is handled during startup alongside other asynchronous work.
  - Prevented readiness checks from hanging when a server listener is available but not responsive.

- **Tests**
  - Added coverage for startup behavior involving blocking tasks, HTTP clients, and HTTP servers.

### DIFF
--- a/changelog.d/8569-next-dylib-provider-host.md
+++ b/changelog.d/8569-next-dylib-provider-host.md
@@ -1,0 +1,12 @@
+## Keep HTTP servers live during microtask-heavy startup
+
+The single-thread async wait driver's fast path now counts a listening external
+HTTP server as native work. Previously it only drove Tokio for blocking tasks
+and HTTP clients. If JavaScript kept queuing microtasks after `server.listen()`,
+`js_wait_for_event` stayed on that fast path while the server's accept task sat
+unpolled. The production Next App Route provider gate therefore depended on the
+accept task winning its initial spawn race and usually parked without serving
+its first request.
+
+`perry-stdlib` now gives the reactor a bounded turn while an external HTTP
+server is active, and its async-bridge unit tests pin that liveness condition.
--- a/crates/perry-stdlib/src/common/async_bridge.rs
+++ b/crates/perry-stdlib/src/common/async_bridge.rs
@@ -342,16 +342,21 @@ extern "C" fn stdlib_wait_wake() {
 /// Wait-driver FAST side — a brief native drive invoked by `js_wait_for_event`
 /// when JS work is pending (a notify or queued microtasks). On the single-thread
 /// runtime, in-flight native tasks (a fetch's reqwest `send`, its h2 connection
-/// driver, sibling fetches) run ONLY inside a tick; under constant JS promise
-/// churn the fast-path is taken every iteration, so without this they are starved
-/// forever (the bundle hang). When something native IS in flight, drive one short
-/// (1 ms) tick: `block_on` drains the run queue (starts freshly-spawned sibling
-/// fetches) and parks briefly on the I/O reactor (advancing TLS/h2 round-trips),
-/// ending early if a native result is queued. No-op when nothing native is in
-/// flight, so pure-JS-async pays only an atomic load.
+/// driver, sibling fetches, or a server accept loop) run ONLY inside a tick;
+/// under constant JS promise churn the fast-path is taken every iteration, so
+/// without this they are starved forever (the bundle hang). When something
+/// native IS in flight, drive one short (1 ms) tick: `block_on` drains the run
+/// queue (starts freshly-spawned tasks) and parks briefly on the I/O reactor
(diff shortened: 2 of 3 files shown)


proggeramlug:

| check | result |
|---|---|
| `check_file_size.sh` | 0 |
| `raw_handle_debt.py` | 0 |

PR #8570, opened 2026-08-22
perf(runtime): eliminate redundant asyncpipe microtask work

### LINKED ISSUE #8409: perf: asyncpipe is 1.12x Node but NOT CPU-bound — needs event-loop/timer counters, not a sampler

proggeramlug:
## Summary

Makes `asyncpipe` faster than Node by eliminating redundant same-thread promise-drain wakeups and removing per-element closure/guard/chained-promise allocations from the fully intrinsic `Promise.all` path.

The optimized path remains behind the existing observability guards: custom constructors, resolvers, `then` properties, and active promise lifecycle hooks all retain spec dispatch.

## Changes

- Suppress event-loop notifications for promise jobs that the current same-thread microtask drain will consume before returning; cross-thread, timer, and rejection-tail wakeups are unchanged.
- Reuse the runtime's GC-scanned keyed `PromiseAllState` table for intrinsic `Promise.all`, avoiding the resolve-element closure, `AlreadyCalled` array, and discarded chained promise.
- Add regression coverage for drain-time notification suppression and shared pending promises used by multiple `Promise.all` calls.
- Add a changelog fragment. No workspace version bump and no `CLAUDE.md` / `CHANGELOG.md` edits.

## Related issue

Fixes #8409

## Measurements

Built with the required static wrappers and measured with `PERRY_NO_AUTO_OPTIMIZE=1` and `PERRY_NO_CACHE=1` on Node 26.5.1.

Two `hyperfine` sessions used 10 warmups and 50 measured runs each, with command order reversed between sessions:

| Command | Forward order | Reverse order |
| --- | ---: | ---: |
| Perry baseline | 115.5 ms ± 8.0 | 116.2 ms ± 9.1 |
| Perry fixed | 90.4 ms ± 7.3 | 87.9 ms ± 5.1 |
| Node | 116.9 ms ± 13.9 | 118.7 ms ± 18.0 |

The fixed result is approximately **0.76x Node** and 22–24% faster than the Perry baseline.

Ten-run `/usr/bin/time -lp` means (peak RSS is the maximum observed run):

| Metric | Perry baseline | Perry fixed | Node |
| --- | ---: | ---: | ---: |
| Instructions retired | 1,351,461,303 | 1,053,050,346 | 802,543,376 |
| Peak RSS | 40,615,936 B | 32,456,704 B | 95,731,712 B |

Counter evidence for one byte-exact `asyncpipe` run:

- event notifications: `sent=119844,during_drain=119844` -> `sent=3,during_drain=3,drain_suppressed=119841`
- closure allocations: `48443` -> `24643`
- microtask drains remained 8, callback timers remained 3, and event-loop waits remained 0

## Test plan

- [x] Required release build: `cargo build --release -p perry -p perry-runtime-static -p perry-stdlib-static`
- [x] All 19 sweep corpus rows are byte-exact against their Node oracle
- [x] `cargo test --release -p perry-runtime --lib -- --test-threads=1` — 2,616 passed, 0 failed, 4 ignored
- [x] `cargo test --release -p perry --bin perry -- --test-threads=1` — 1,015 passed, 0 failed
- [x] `python3 scripts/check_test_registration.py` — 230 files registered
- [x] `cargo fmt --all --check` and `git diff --check`
- [x] `BASE_SHA=upstream/main bash scripts/run_lint_gates.sh` — 53/54 pass on the rebased tree. The sole failure is the absolute `raw_handle_debt.py` census for `object/field_get_set/ic_miss.rs`, introduced on upstream main by #8560 and untouched here; this branch passes `raw_handle_debt.py --no-raise-vs upstream/main`, `-D warnings`, and workspace clippy.

## Screenshots / output

Not applicable; runtime performance/correctness change only.

## Checklist

## Summary by CodeRabbit

* **Performance**
  * Reduced redundant event-loop wakeups while processing promise jobs and microtasks.
  * Improved `Promise.all` efficiency by using more compact internal settlement handling where safe.

* **Bug Fixes**
  * Preserved required wakeups for cross-thread activity, timers, and rejection processing.
  * Improved handling of multiple `Promise.all` calls sharing pending promises.

* **Diagnostics**
  * Expanded microtask profiling to report suppressed notifications.

### DIFF
--- a/changelog.d/8409-redundant-microtask-wakes.md
+++ b/changelog.d/8409-redundant-microtask-wakes.md
@@ -1,0 +1,10 @@
+## Suppress redundant promise-drain wakeups
+
+Promise and `queueMicrotask` work queued by the main thread while it is already
+draining microtasks no longer signals the event loop. Cross-thread producers,
+timer callbacks, and rejection processing retain the normal wake path.
+
+`Promise.all` over plain native promises also uses compact settlement records
+instead of allocating a closure and `AlreadyCalled` array for every element;
+observable constructor, resolver, `then`, and lifecycle-hook paths still use
+the spec dispatch.
--- a/crates/perry-runtime/src/event_pump.rs
+++ b/crates/perry-runtime/src/event_pump.rs
@@ -231,6 +231,7 @@ static NOTIFIED: AtomicBool = AtomicBool::new(false);
 static WAITER_COUNT: AtomicI64 = AtomicI64::new(0);
 pub static PROFILE_NOTIFY_COUNT: AtomicI64 = AtomicI64::new(0);
 pub static PROFILE_NOTIFY_DURING_DRAIN_COUNT: AtomicI64 = AtomicI64::new(0);
+pub static PROFILE_NOTIFY_DRAIN_SUPPRESSED_COUNT: AtomicI64 = AtomicI64::new(0);
 pub static PROFILE_WAIT_COUNT: AtomicI64 = AtomicI64::new(0);
 pub static PROFILE_WAIT_FAST_COUNT: AtomicI64 = AtomicI64::new(0);
 pub static PROFILE_WAIT_ZERO_COUNT: AtomicI64 = AtomicI64::new(0);
@@ -365,6 +366,23 @@ pub extern "C" fn js_notify_main_thread() {
     *flag = true;
     drop(flag);
     PUMP.cvar.notify_one();
+}
+
+/// Notify after queueing promise/microtask work, unless this same thread is
+/// already draining that queue to quiescence.
+///
+/// Cross-thread producers see a thread-local drain depth of zero and retain
+/// the full wake path. Timer and rejection phases are outside the job-drain
+/// scope as well, because work they queue is consumed by a later pump turn.
+#[inline(always)]
+pub(crate) fn js_notify_promise_progress() {
+    if crate::promise::microtasks::microtask_job_drain_active() {
+        if crate::promise::mt_profile_enabled() {
+            PROFILE_NOTIFY_DRAIN_SUPPRESSED_COUNT.fetch_add(1, Ordering::Relaxed);
+        }
+        return;
+    }
+    js_notify_main_thread();
 }
 
 // ============================================================================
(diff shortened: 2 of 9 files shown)


proggeramlug:
Merging. Both audited against the specific hazard their class carries, not just "the tests pass".

**#8570 — the failure mode here is a hang, not a wrong answer**, so the question is whether a needed wakeup can ever be suppressed. It cannot: `MICROTASK_RUN_DEPTH` is a **thread-local `Cell`**, so a cross-thread producer observes depth 0 and keeps the full wake path; suppression fires only when `jobs != 0`, i.e. inside the drain that will consume the job before returning; timer and rejection phases sit outside that drain and are explicitly carved out. `PROFILE_NOTIFY_DRAIN_SUPPRESSED_COUNT` makes the subject assertable rather than leaving "nothing threw" as the evidence.

**#8573 — same exact-ShapeId guard class that produced #8577 and #8578 within two days**, so I read the gate order rather than the benchmark. It acquire-loads the prototype-mutation latch, validates pointer form and target heap range *before* any dereference, then checks GC kind, forwarding state, the per-object descriptor bit, nonzero/exact class id and exact ShapeId — with dynamic fallback on every failed proof. Fail-closed at each step, and scoping descriptor invalidation to the receiver and relevant prototype mutations is the right narrowing: an unrelated object's descriptor can affect neither method resolution nor the ShapeId proof.

| check | result |
|---|---|
| all six ratchets | 0 |

Also PR-keyed #8570's changelog fragment, which was named `8409-` after the issue.

Standing caveat on both: these are perf changes validated for **correctness**, not measured on the quiet-host corpus — that rebuild is currently blocked on disk. #8573's own numbers are explicitly labelled paired qualification on a contended host rather than a release baseline, which is the honest framing.

PR #8571, opened 2026-08-22
perf(runtime): resolve string builder round trips

### LINKED ISSUE #8434: perf(runtime): join/repeat/pad/replace build through a Rust String round-trip (double copy; join lacks with_capacity)

proggeramlug:
Closes #8434.

## What changed

- converts `String.prototype.repeat` to one exact GC allocation, roots and re-reads the receiver after that allocation, and fills the result with exponentially growing bulk copies
- adds a single-pass buffered-result finalizer that derives UTF-16 length and WTF-8 metadata, enforces the string-length limit, and canonicalizes surrogate pairs created at builder boundaries
- routes buffered join, pad, literal replace, regex replace, named/fancy regex, and callback replace results through the metadata-safe finalization/copy paths
- fixes empty literal-pattern replacement at JavaScript UTF-16 boundaries, including the boundary between the two halves of an astral character and callback offsets
- adds Rust coverage plus a Node 26.5.1 parity fixture for holes, empty values/separators, astrals, lone surrogates, allocating coercions/callbacks, regex/fancy-regex, and no-match paths under forced moving-GC verification

## Full #8434 performance audit

I implemented and benchmarked direct-write candidates for all four requested families. Per the issue's explicit "simpler code wins ties/regressions" rule, only repeat survived:

| Family | Result | Decision |
| --- | --- | --- |
| `join` | direct candidate: 81.8 ± 11.1 ms → 558.9 ± 28.5 ms (6.83x slower) | retain the buffered path; current main already has the separator lower-bound `String::with_capacity` fallback |
| `repeat` | final: 1.095 ± 0.004 s → 439.8 ± 4.0 ms (2.49x faster); ~13.90B → ~6.98B retired instructions (49.8% fewer) | retain direct exact allocation |
| `padStart`/`padEnd` | direct candidate wall time improved only ~8%, but retired instructions rose ~21.4B → ~24.2B (~13%) | retain buffered path and its cheap WTF-8 boundary finalization |
| replace paths | optimized direct candidate remained ~12% slower; ~7.47B → ~8.30B retired instructions (~11% more), with literal/regex/callback granular timings inside noise or worse | retain buffered paths and make their shared finalization correct/safe |

The benchmark sweeps covered join lengths 16/256/4096/16384, repeat and pad outputs 64/4096/262144/1048576, and replace inputs 256/8192/262144/1048576. Wall measurements used interleaved `hyperfine` runs; retired instructions came from macOS task counters (`TASK_INSPECT_BASIC_COUNTS`). Every final mode produced the same checksum as the upstream baseline.

## Validation

- `RUST_TEST_THREADS=1 CARGO_INCREMENTAL=0 CARGO_BUILD_JOBS=2 cargo +nightly-2026-08-20 test --release -p perry-runtime` — 2624 passed, 0 failed, 4 ignored; doc tests 0 failed
- fresh `perry-dev` build of `perry`, `perry-runtime-static`, and `perry-stdlib-static`, then forced-GC parity against Node 26.5.1 — 1/1 passed
- `cargo +nightly-2026-08-20 check --profile perry-dev -p perry-runtime`
- `cargo +nightly-2026-08-20 fmt --all -- --check`
- `python3 scripts/raw_handle_debt.py` — 925, exactly baseline
- `git diff --check`

No version bump.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved correctness for string joining, repetition, padding, and replacement.
  - Fixed handling of Unicode characters, surrogate pairs, lone surrogates, and UTF-16 boundaries.
  - Corrected empty-pattern and callback-based regular expression replacements.
  - Improved behavior for unmatched patterns and replacement tokens.
  - Added safeguards for excessively large string results, returning the appropriate error.
- **Tests**
  - Expanded coverage for Unicode strings, sparse-array joins, regular expressions, callbacks, and allocation-heavy operations.


proggeramlug:
Cross-referencing with #8568, which measured a direct-write join at **1.38x faster** on the same 16/256/4096/16384 sweep where this PR's candidate measured **6.83x slower**. Both numbers reproduce — they are measurements of different algorithms. Your rejected candidate (preserved at `ae50c92e7`) pays a per-element `RuntimeHandleScope` + `js_jsvalue_to_string` coercion (strings included), a barriered store into a GC scratch array per element, and a descriptor `Vec` entry per piece including every separator repetition. #8568's well-formed path proves the dense case in one unrooted pass, computes the exact size, allocates once, and memcpys — no per-element rooting. So "direct write" was not the thing your benchmark rejected; per-element rooting overhead was.

I re-measured main, this PR, #8568, and your `ae50c92e7` candidate under one harness — interleaved medians, M1 Max, same checksummed output; instructions via macOS task counters:

| join (strings sweep) | wall | instructions |
|---|---:|---:|
| main | 682.6 ms | 6.536 B |
| main + #8571 (buffered) | 676.7 ms (0.99x) | 6.527 B |
| main + #8568 (direct) | 466.6 ms (0.68x) | 4.855 B (−25.7%) |
| main + rejected candidate (`ae50c92e7`) | 10 016 ms (14.7x) | 88.26 B (13.5x) |

To be explicit: **this validates your judgement rather than overruling it.** You measured a genuinely catastrophic implementation — my rebuild of your candidate reproduces it at 13.5x retired instructions (and 246 MB peak RSS vs 66 MB on a numbers sweep; the barriered scratch array plus per-piece descriptor Vec is the tell) — and you correctly refused to ship it. What the combined PR takes is #8568's *different* algorithm (exact-size, single allocation, no per-element rooting), which your fixture never saw. "Direct-write join" was simply two algorithms sharing a name.

Everything else in this PR held up and is being landed intact in https://github.com/PerryTS/perry/pull/8581: the exact-allocation `repeat` (reproduced: ~74% fewer retired instructions on my sweep), `js_string_from_builder_bytes` and the WTF-8/UTF-16 correctness work for pad/replace, and your join tests — which pass unchanged against #8568's implementation, since it preserves surrogate canonicalization across empty fields/separators, lone-surrogate flags, and exactly-once source-order coercion.

PR #8573, opened 2026-08-22
perf(codegen): inline monomorphic method shape guards

proggeramlug:
## Summary

- scope direct-method descriptor invalidation to the receiver and relevant prototype mutations instead of poisoning every site after any unrelated descriptor install
- inline the complete single-arm shape-only method guard at generated call sites
- preserve the unchanged dynamic fallback on every failed proof

## Safety contract

The inline path checks the release-published prototype latch with an acquire load, validates the boxed/raw pointer form and target-specific heap range before dereference, then checks GC kind, forwarding state, the per-object descriptor bit, nonzero/exact class id, and valid/exact ShapeId. Prototype descriptor installs retire the process-wide direct-method guard latch; unrelated object descriptors no longer do.

## Performance

Qualification on a contended host used a 10,000-object callback-heavy query, 11 alternating pairs, and 64 operations per measured batch. All accumulation runs produced the exact expected sum.

- scoped receiver/prototype descriptor contract versus the old process-wide descriptor gate, with otherwise identical inline checks: 91.07% lower read-only time and 92.31% lower accumulation time; candidate won 11/11 pairs for both
- inline guard versus the scoped out-of-line helper: 7.37% lower read-only time and 8.10% lower accumulation time; candidate won 11/11 pairs for both

These are paired qualification results, not a quiet-host release baseline.

## Tests

- `./scripts/test_affected_crates.sh --base upstream/main`
  - perry-runtime: 2,620 passed
  - perry-codegen: 1,127 passed
  - perry-ffi: 29 passed
  - Perry CLI/bin and updater scopes passed
- `cargo test -p perry --test method_shape_inline_guard`
- focused runtime regressions cover unrelated descriptors, own-descriptor rejection, and prototype-descriptor invalidation
- LLVM acquire-load roundtrip and target heap-bound portability tests

## Summary by CodeRabbit

- **Performance**
  - Improved direct method-call performance with optimized shape checks and inline fast paths.
  - Maintained safe fallback to dynamic dispatch when object shapes or prototype definitions change.

- **Bug Fixes**
  - Scoped descriptor changes so unrelated objects no longer invalidate valid method optimizations.
  - Improved pointer and heap validation across supported platforms.

- **Tests**
  - Added coverage for prototype invalidation, descriptor behavior, shape changes, and atomic operations.
  - Verified optimized calls continue producing correct results after runtime mutations.

### DIFF
--- a/changelog.d/8573-inline-method-shape-guard.md
+++ b/changelog.d/8573-inline-method-shape-guard.md
@@ -1,0 +1,5 @@
+Scoped direct-method descriptor invalidation to the receiver and relevant
+prototype mutations, then inlined complete monomorphic ShapeId guards. On a
+10,000-entity callback-heavy query with an unrelated descriptor, read-only time
+falls from 3.7599 to 0.3381 ms/op and accumulation from 3.6702 to 0.2829 ms/op;
+inlining contributes a further 7.4% and 8.1% paired reduction respectively.
--- a/crates/perry-codegen/src/block.rs
+++ b/crates/perry-codegen/src/block.rs
@@ -634,6 +634,19 @@ impl LlBlock {
         r
     }
 
+    /// Acquire atomic load for a sticky runtime gate whose publishing store is
+    /// release-ordered. The explicit alignment is required by LLVM atomics.
+    pub fn load_atomic_acquire(&mut self, ty: LlvmType, ptr: &str, alignment: u32) -> String {
+        let r = self.reg();
+        self.push_inst(crate::inst::LlInst::Load {
+            dst: r.clone(),
+            ty,
+            ptr: ptr.to_string(),
+            flavor: crate::inst::LoadFlavor::AtomicAcquire(alignment),
+        });
+        r
+    }
+
     /// Sequentially-consistent atomic load for globals shared with runtime
     /// atomics. The explicit alignment is required by LLVM atomic loads.
     pub fn load_atomic_seq_cst(&mut self, ty: LlvmType, ptr: &str, alignment: u32) -> String {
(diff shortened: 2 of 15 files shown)


proggeramlug:
Merging. Both audited against the specific hazard their class carries, not just "the tests pass".

**#8570 — the failure mode here is a hang, not a wrong answer**, so the question is whether a needed wakeup can ever be suppressed. It cannot: `MICROTASK_RUN_DEPTH` is a **thread-local `Cell`**, so a cross-thread producer observes depth 0 and keeps the full wake path; suppression fires only when `jobs != 0`, i.e. inside the drain that will consume the job before returning; timer and rejection phases sit outside that drain and are explicitly carved out. `PROFILE_NOTIFY_DRAIN_SUPPRESSED_COUNT` makes the subject assertable rather than leaving "nothing threw" as the evidence.

**#8573 — same exact-ShapeId guard class that produced #8577 and #8578 within two days**, so I read the gate order rather than the benchmark. It acquire-loads the prototype-mutation latch, validates pointer form and target heap range *before* any dereference, then checks GC kind, forwarding state, the per-object descriptor bit, nonzero/exact class id and exact ShapeId — with dynamic fallback on every failed proof. Fail-closed at each step, and scoping descriptor invalidation to the receiver and relevant prototype mutations is the right narrowing: an unrelated object's descriptor can affect neither method resolution nor the ShapeId proof.

| check | result |
|---|---|
| all six ratchets | 0 |

Also PR-keyed #8570's changelog fragment, which was named `8409-` after the issue.

Standing caveat on both: these are perf changes validated for **correctness**, not measured on the quiet-host corpus — that rebuild is currently blocked on disk. #8573's own numbers are explicitly labelled paired qualification on a contended host rather than a release baseline, which is the honest framing.

PR #8576, opened 2026-08-22
feat(compile): perry compile --report-size, a native binary size report

jdalton:
## Summary

Adds `perry compile --report-size`, which writes `<output>.size-report.md`: a per-crate and per-symbol size breakdown of the final linked binary, attributed straight from its own symbol table.

<details>
<summary>Why not cargo-bsize?</summary>

I first tried pointing [cargo-bsize](https://github.com/boshen/cargo-bsize) at what Perry actually ships. It only knows how to drive a `cargo build`/`cdylib` rebuild — but Perry's real output binaries come from a two-stage build (`perry-runtime-static`/`perry-stdlib-static` compile to `.a` archives, then a raw `cc`/`ld` link with LLVM-emitted object code) that cargo-bsize has no hook into. A throwaway crate linking `perry-runtime`/`perry-stdlib` directly under a plain `cargo build` fails to link at all — their `#[no_mangle] extern "C"` symbols get dead-code-eliminated; they only survive in the real build via the staticlib wrappers' `#[used]` anchors.

So `--report-size` reads the real linked binary instead, using `object` + `rustc-demangle` (the same core technique cargo-bsize itself uses, applied without needing a rebuild). It reuses the existing `PERRY_KEEP_SYMBOLS` strip-skip knob (`--debug-symbols` already uses the same mechanism, minus the `-g` DWARF cost) so there's a symbol table to read.

On a trivial `console.log("hello")` program it correctly surfaces `perry_runtime` as the single largest attributable crate (~3.8 MiB of an 8.9 MiB binary in one measurement) — actionable signal for anyone chasing shipped-binary bloat, in the same territory as [#8418](https://github.com/PerryTS/perry/pull/8418).

</details>

## Test plan

- [x] New unit tests for `crate_of` (including the v0-mangling `crate[hash]::path` disambiguator shape and `<Type>::method` associated-fn receivers — both were a real bug I caught by running against a live binary before fixing) and `human_bytes`/`report_path_for`.
- [x] `cargo clippy -p perry --bins` — no new warnings from `size_report.rs`.
- [x] End-to-end: compiled a `console.log` program with `--report-size` and inspected the generated `hello.size-report.md`.
- [x] `scripts/check_file_size.sh`, `scripts/check_node_version_consistency.py --list`, `scripts/gc_gate_wiring_check.py` all pass.

## Summary by CodeRabbit

- **New Features**
  - Added the `--report-size` compile option.
  - Generates a Markdown report showing executable size usage by crate and symbol.
  - Reports include code and data sections, readable symbol names, and ranked size breakdowns.
  - Report-generation issues produce warnings without failing compilation.

- **Documentation**
  - Documented the new size-reporting option.

- **Chores**
  - Updated the release version to `0.5.1518`.


proggeramlug:

| check | result |
|---|---|

PR #8577, opened 2026-08-22
fix(codegen): keep imported class shapes canonical

proggeramlug:
## Summary

- keep imported class stubs on the canonical structural ShapeId instead of inferring a typed identity from a missing constructor body
- validate the typed GC layout after the defining module's real constructor returns
- preserve at-allocation layout declarations for a real local class that shadows same-named import metadata

## Root cause

An imported HIR stub carries field names and types, but deliberately has no defining constructor body. Codegen treated that absent body as enough evidence to declare a typed layout before construction.

That could give one runtime class two exact identities: the producer initialized `js_object_shape_id_for_keys`, while a consumer allocating the imported class initialized `js_gc_typed_shape_id_for_keys`. Exact method guards emitted in the producer then rejected every consumer-created instance even though the class id and keys agreed. In the ECS reproducer, the expected ShapeId was `0x80000ac4` and the consumer instance carried `0x80000ad3`.

The conservative fix does not export or guess new proof. Imported stubs use the structural identity and retain the existing post-constructor `js_gc_init_typed_shape_layout` validation. A future zero-overhead path can export producer-authored constructor/layout proof explicitly.

## Tests

- `cargo test --release -p perry-codegen --lib` — 1,125 passed
- `cargo test --release -p perry-codegen --test typed_shape_declared_at_allocation` — 9 passed
- focused typed-shape bake tests — 5 passed, including imported-stub and local-shadow negative controls
- complete upstream ECS functional corpus with the accepted prototype stack — 35/37 files and 461/463 cases, matching the established Perry baseline (only the known singleton shorthand and `Symbol.dispose` mismatches)

The final formatting-only commit repairs the PIC test formatting already carried by #8573; current `main` fails `cargo fmt --all -- --check` without it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of imported classes to ensure their runtime layouts are validated using the producer’s canonical structural information.
  - Prevented incorrect or premature layout declarations for imported, pointer-bearing classes.
  - Preserved existing layout behavior for locally defined classes, including cases where local definitions share names with imported metadata.

- **Tests**
  - Added regression coverage for imported-class layout handling and local-class shadowing scenarios.

### DIFF
--- a/crates/perry-codegen/src/codegen/mod.rs
+++ b/crates/perry-codegen/src/codegen/mod.rs
@@ -1855,6 +1855,10 @@ pub fn compile_module(hir: &HirModule, opts: CompileOptions) -> Result<Vec<u8>> 
     // a class only gets a site entry if module init will actually compose its
     // image. A site entry with no init store would hand every instance a
     // zeroed header, so that direction of the dependency is load-bearing.
+    let imported_stub_names: std::collections::HashSet<&str> = imported_class_stubs
+        .iter()
+        .map(|class| class.name.as_str())
+        .collect();
     let class_header_image_inits: std::collections::HashMap<String, (u32, u64)> = {
         let mut inits: std::collections::HashMap<String, (u32, u64)> =
             std::collections::HashMap::new();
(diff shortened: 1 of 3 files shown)


proggeramlug:

| check | result |
|---|---|

PR #8578, opened 2026-08-22
fix(runtime): exclude prototypes from learned instance width

proggeramlug:
## Summary

- exclude registered declared-class and function-class prototype objects from learned instance inline-width samples
- pass the concrete owner identity to every width-learning call site
- keep ordinary instance spills training the class high-water mark

## Root cause

Prototype objects intentionally carry their owning instance class id for reflection and prototype dispatch. The learned inline-width table keyed only on that class id, so installing methods/properties on a prototype could teach the allocator that future instances needed the prototype's much larger storage width.

In the ECS reproducer, an eight-field class's prototype spill trained a live width of 33. New instances were then born with a different ShapeId than the compiler-published eight-field method guard, so the exact guard could never succeed.

The fix retains width learning and changes its admission contract to include object identity. It rejects only exact pointers registered in `CLASS_DECL_PROTOTYPE_OBJECTS` or `CLASS_PROTOTYPE_OBJECTS`; a real instance with the same class id still updates the high-water mark.

## Tests

- focused prototype/instance regression with default object-owned spill — passed
- same regression with `PERRY_OBJECT_SPILL=0` legacy storage — passed
- `cargo test --release -p perry-runtime --lib` — 2,617 passed, 4 ignored
- spill-filter cohort — 5 passed
- `cargo check --release -p perry-runtime`
- complete upstream ECS functional corpus with the accepted prototype stack — 35/37 files and 461/463 cases, matching the established Perry baseline (only the known singleton shorthand and `Symbol.dispose` mismatches)

This is one of three serial exact-guard blockers found in the ECS query profile. The combined accepted stack reduced the six query units from 4.0688 ms/op to 1.1052 ms/op; this PR deliberately does not claim that combined speedup as an isolated effect.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved object storage behavior so prototype objects no longer influence instance field layout learning.
  * Preserved accurate field handling for regular object instances.

* **Tests**
  * Added coverage verifying prototype spills are ignored while instance spills continue to update learned field counts.

### DIFF
--- a/crates/perry-runtime/src/object/class_registry.rs
+++ b/crates/perry-runtime/src/object/class_registry.rs
@@ -51,6 +51,8 @@ mod state;
 mod vm_brand;
 
 // ── state.rs ────────────────────────────────────────────────────────────────
+#[cfg(test)]
+pub(crate) use state::class_decl_prototype_object_root_store;
 pub(crate) use state::{
     class_decl_prototype_object, class_decl_prototype_value,
     class_decl_prototype_value_for_instance_class, class_delete_own_dynamic_prop,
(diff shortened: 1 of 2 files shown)


proggeramlug:

| check | result |
|---|---|

Issue #7848, opened 2026-08-11
Generic class specialization is lost when a type argument is a type ALIAS — dispatch guard compiled against the template class, fast path permanently dead (53.8% of `pipeline`)


proggeramlug:
## Summary

A generic **class** instantiated with a type argument that is a **type alias** silently loses its monomorphized specialization at every use of the binding. The emitted dispatch guard is compiled against the generic *template* class while the object is an instance of the *specialization*, so the guard can never pass and every method call on that binding takes the fully dynamic `js_native_call_method_by_id` path — permanently.

Output stays correct and the exit code is 0. Nothing goes red. The only symptom is speed.

On `gc-handoff/apps/pipeline.ts` this is **53.8% of the program's samples** (29.4% of it pure dispatch overhead), measured on a quiet M1 mini with `sample` at 1 ms over two agreeing runs.

## Reproducer

```ts
class Registry<K, V> {
  private keys: K[] = [];
  private vals: V[] = [];
  set(k: K, v: V): void { this.keys.push(k); this.vals.push(v); }
  size(): number { return this.keys.length; }
}

type Record = { id: number };
type Stage = (r: Record) => Record;          // <-- an ordinary alias of a function type

function main(): void {
  const stats  = new Registry<Stage,  number>();   // BROKEN: dispatch is fully dynamic
  const byKind = new Registry<string, number>();   // fine:   guard-free direct call
  // ...
}
```

Two locals, same class declaration, adjacent lines. Only one gets the fast path.

## Mechanism

`perry --trace hir --focus main`:

```
Let { name: "stats",  ty: Generic { base: "Registry", type_args: [Named("Stage"), Number] },
                      init: New { class_name: "Registry$fn_num"  } }
Let { name: "byKind", ty: Generic { base: "Registry", type_args: [String,        Number] },
                      init: New { class_name: "Registry$str_num" } }
```

**The binding's declared type keeps the unexpanded alias; the `New` was specialized on the expanded type.** The two are lowered from the same AST nodes by two different calls:

| site | call | alias |
|---|---|---|
| `crates/perry-hir/src/lower/expr_new.rs:1216` — the `New`'s own `type_args` | `extract_ts_type_with_ctx(t, Some(ctx))` | **expanded** |
| `crates/perry-hir/src/lower_types.rs:604` — the INFERRED declared type of `new C<…>()` | `extract_ts_type(t)` (= `…_with_ctx(t, None)`) | **not expanded** |

[...]

PR #8580, opened 2026-08-22
perf(codegen): specialize proven nonnegative index methods

proggeramlug:
## Summary

- emit an internal full-body clone for small synchronous methods whose stable `number`/`int32` parameters are used as indexes
- route exact same-module calls to that clone only when every selected argument has an existing nonnegative-i32 proof
- keep the public method body as the semantic fallback for negative, fractional, string-like, imported, and dynamically mutated calls
- let the inline guarded-array tier follow one ordinary forwarding edge, fully revalidate the live destination, and carry that exact handle into the raw slot load

## Root cause

The ECS hot path calls a small helper shaped like `readSimpleColumnValue(column, entityIndex)`. The caller already proves `entityIndex` is a nonnegative i32 loop index, but the public JSValue method ABI loses that fact. The helper consequently lowers `column[entityIndex]` through generic property-key dispatch on every entity.

Changing the public body would be unsound: a `number` annotation does not exclude negative or fractional JavaScript keys. This patch instead creates an additive clone and admits calls only after both the existing method-identity guard and the concrete index proof succeed.

Clone reachability is recorded from locally emitted bodies rather than inferred from the merged local/imported class table, so an imported structural stub cannot manufacture a reference to a clone its defining module did not export.

## Safety boundaries

- candidate methods are synchronous, at most eight statements, and have no defaults, rest parameter, or `arguments` object
- selected parameters must be numeric, index-used, never reassigned, and never referenced by a nested closure
- call sites must have exact arity and prove every selected argument is in `[0, i32::MAX]`
- all non-proven calls retain the unchanged public body and generic key semantics
- a followed array destination is rechecked for pointer range, GC type, forwarding state, descriptors, prototype validity, layout, capacity, length, and bounds before a raw load

## Validation

- `cargo check --release -p perry-codegen`
- focused release regressions: 3 passed
- complete `perry-codegen` release library suite: 1,132 passed
- 11 alternating fixed-repeat ECS query A/B pairs against the immediately preceding `main`; all 22 processes passed the query assertions and exact `50,005,000` accumulation oracle

The host was heavily contended, so these are qualification ratios rather than release-quality absolute timings:

| workload | paired median reduction | candidate wins |
|---|---:|---:|
| 10k read-only query | 38.0% | 9/11 |
| 10k query + accumulation | 57.7% | 10/11 |

The ECS executable grew by 16,520 bytes (0.14%). A retained lower-contention stage replay independently moved the read-only row from about 1.12 ms/op to 0.49 ms/op with the same mechanism.

## Summary by CodeRabbit

* **Performance**
  * Improved performance for eligible array-index method calls when indexes are proven to be nonnegative integers.
  * Added optimized guarded array access, including validation after simple forwarding.
* **Bug Fixes**
  * Preserved standard property-key behavior for negative, fractional, dynamic, or otherwise unverified indexes.
  * Prevented optimization when parameters are modified, defaulted, or captured.
* **Tests**
  * Added comprehensive coverage for optimized calls and fallback behavior.
* **Documentation**
  * Documented benchmark results and validation outcomes.


proggeramlug:
@coderabbitai review

proggeramlug:
Merging.

**The design decision is the right one, and it is the reason this is safe to take.** Specializing the public method body would have been unsound, and the PR says exactly why:

> a `number` annotation does not exclude negative or fractional JavaScript keys

That is the same lesson as #7773 — a declared type is not a proof. Emitting an *additive* clone and routing to it only when **both** the existing method-identity guard and a concrete `[0, i32::MAX]` argument proof hold keeps the public body as the semantic fallback for negative, fractional, string-like, imported and dynamically-mutated calls. Nothing loses generic key semantics by default.

Two details I would have gone looking for and found already handled:

- **Clone reachability is recorded from locally emitted bodies**, not inferred from the merged local/imported class table — so an imported structural stub cannot manufacture a reference to a clone its defining module never exported. That is the #7848 failure mode pre-empted rather than rediscovered.
- **The followed array destination is fully revalidated** — pointer range, GC type, forwarding state, descriptors, prototype validity, layout, capacity, length, bounds — before the raw slot load. Following a forwarding edge and then trusting the origin's proof is precisely how these tiers go wrong.

| check | result |
|---|---|
| all six ratchets | 0 |

Standing caveat, same as the other perf changes landed today: this is validated for **correctness**, not measured on the quiet-host corpus. The rebuild for that is currently disk-blocked. Given the guard structure — every unproven call keeps the old path — the downside risk is a missed speedup rather than a regression, which is the right shape for merging ahead of measurement.

PR #8582, opened 2026-08-22
fix(codegen): validate forwarded array target before header load

proggeramlug:
The one-edge array forwarding guard added in #8580 selected the forwarding word and computed its address-band predicates, but loaded the selected target header before those predicates controlled execution. A malformed or longer-chain target could therefore be dereferenced natively instead of reaching the semantic fallback.

This adds a dedicated validated-target block. The selected address must have clear high bits and be above the handle band before any destination header, length, or capacity load. An invalid selected target routes directly to the boxed fallback; valid targets keep the existing brand, forwarding, descriptor, prototype, bounds, and layout guards.

The emitted-IR regression proves the selected-target block branches to the validated successor or fallback before computing live_handle - 8, and that the destination header load appears only in the validated successor.

Validation:
- cargo test --release -p perry-codegen --lib: 1132 passed
- focused emitted-IR regression passed
- cargo fmt --all -- --check
- cargo check --release -p perry-codegen

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation of forwarding targets before accessing referenced data.
  - Invalid or excessively long forwarding chains now safely use the boxed fallback path.
  - Added safeguards to prevent malformed addresses from being dereferenced.
  - Improved reliability of indexed access when encountering invalid or unexpected runtime object references.

- **Tests**
  - Added coverage verifying forwarding targets are validated before their headers are accessed.

### DIFF
--- a/changelog.d/8582-validate-forward-target.md
+++ b/changelog.d/8582-validate-forward-target.md
@@ -1,0 +1,4 @@
+Fixed guarded array reads so a selected forwarding target is validated as a
+heap address before Perry dereferences its header. Invalid or longer forwarding
+chains now route to the boxed fallback instead of permitting a speculative
+native header load.
--- a/crates/perry-codegen/src/codegen/index_method_clone_tests.rs
+++ b/crates/perry-codegen/src/codegen/index_method_clone_tests.rs
@@ -302,6 +302,25 @@ fn guarded_read_can_follow_one_forwarding_edge_but_rechecks_the_live_header() {
         .find(|line| line.contains(" = select i1") && line.contains(", i64 "))
         .and_then(|line| line.trim().split_once(" = ").map(|(name, _)| name))
         .unwrap_or_else(|| panic!("clone has no selected live array handle:\n{clone}"));
+    let selected_target_guard = clone
+        .split("\narr.guard.deref.")
+        .nth(1)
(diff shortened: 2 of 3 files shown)


proggeramlug:
Merging. This is a real safety defect in #8580, and I should have caught it when I merged that PR an hour ago.

**What went wrong in my audit.** #8580's description lists the followed array destination as revalidated for "pointer range, GC type, forwarding state, descriptors, prototype validity, layout, capacity, length, and bounds before a raw load". I confirmed those checks were **present** and merged on that basis. I did not verify they **dominated** the load — and they did not: the address-band predicates were computed, but the selected target header was dereferenced before they controlled execution.

Presence is not dominance. That is the same invariant this codebase already pays for on the GC side — a root store must dominate every subsequent site that can collect, and a store that exists but sits after the collecting call is worth nothing. I applied that reasoning to rooting all session and failed to apply it to a guard.

**The regression test is the right shape**, and that is what makes this trustworthy rather than merely fixed: it asserts the selected-target block *branches to the validated successor or fallback before computing `live_handle - 8`*, and that the destination header load appears **only** in the validated successor. An ordering assertion, not an outcome assertion. A test that merely checked "the guards are emitted" would have passed against the broken version too.

| check | result |
|---|---|
| `perry-codegen --lib` | **1132 passed** |
| all seven ratchets | 0 |

Issue #8583, opened 2026-08-22
[perf][regression] Large minified bundle becomes a multi-hour compile after RS4GC/opt-tier changes (83 CGUs, ~17 GiB RSS)


proggeramlug:
## What happened

A cold native compile of the public `@anthropic-ai/claude-code@2.1.112`
standalone bundle no longer finishes in the previously observed time on current
Perry `main`.

On Perry `0.5.1515` (`b00e261f8114dd0f39d061131703a918e17b408f`), the
compiler selected 83 codegen units, spent about 27 minutes partitioning the
already-lowered module, reached approximately 17 GiB RSS, and then took 11--14
minutes for each of the first LLVM units that completed. I interrupted the run
after 74.5 minutes because only two units had completed and the remaining
linear ETA was multiple hours.

No output binary was produced, so this also prevents measuring the intended
runtime/GC performance of current Perry on this real application.

This appears to be a compile-time regression caused by the interaction of:

1. native roots/RS4GC becoming the default;
2. removal of the post-RS4GC instruction-count guard and the older oversized
   function/unit `-O0` fallback;
3. `-Os` becoming the default for every native module; and
4. IR-size-based partitioning increasing this input from approximately 6 to 83
   codegen units without being able to split the largest individual functions.

The important point is that inter-function codegen-unit splitting cannot bound
LLVM work for one giant minified IIFE. Perry's own source comment identifies a
68 MB IIFE in this exact `cli.js` case as irreducible without structured
intra-function outlining.

## What I expected

The default compile should finish in a bounded amount of time and memory,
without sending RS4GC-expanded multi-million-instruction functions through an
unbounded optimizing middle-end.

As a historical reference, the same 13.7 MB input compiled on Perry commit
`4340bffc9bb7500aaf1863a3ba334921c3a92fe8` (2026-07-29) in approximately
40--50 minutes with roughly 5 GiB peak RSS. That build produced a working
native executable.

The current backend and GC strategy do not need to reproduce the old wall time
exactly, but a default compile should complete rather than regress toward a
multi-hour build with more than 3x the recorded memory use.

## Reproduction

This is a scale-triggered compiler issue, so I have not reduced it to a small
TypeScript snippet. The input is a deterministic, publicly obtainable npm
artifact.

### 1. Build current Perry and the runtime libraries

```bash
git clone https://github.com/PerryTS/perry.git
cd perry
git checkout b00e261f8114dd0f39d061131703a918e17b408f

cargo build --release -p perry
[...]

PR #8421, opened 2026-08-19
fix: make compile output TypeScript-developer friendly

proggeramlug:
## Summary

Make perry compile show TypeScript developers actionable build progress instead of successful internal Rust, LLVM, Swift, and linker diagnostics. Native compilation now keeps the -O3 optimization contract for large modules and functions instead of silently falling back to -Os, -O0, or optnone.

## Changes

- Capture successful internal tool output at default verbosity; stream it with --verbose and replay it when a tool fails.
- Hide repetitive per-object cache paths during linked builds while retaining them for --verbose and --no-link.
- Remove oversized-module and post-GC optimization downgrades plus their obsolete environment/cache inputs.
- Keep compile-time scalability in codegen-unit partitioning and safe structured outlining, including the existing ordered helper chunks for generated initialization.
- Document the compiler-output verbosity policy.

## Related issue

n/a

## Test plan

- cargo fmt --all -- --check
- cargo check -p perry-codegen --tests
- cargo check -p perry --bin perry
- cargo test -p perry-codegen --lib linker::tests::compile_plan_always_uses_o3 -- --exact
- cargo test -p perry --bin perry commands::compile::build_cache::tests::codegen_env_vars_are_build_cache_inputs -- --exact
- ./scripts/pre-tag-check.sh --quick
- git diff --check

- [ ] cargo build --release clean
- [ ] Full affected-crate or workspace suite passes; focused checks above were run
- [x] Added or updated a test in the affected crate
- [x] Updated docs/src for the user-facing CLI output change
- [ ] Platform UI backend build; not applicable

## Screenshots / output

Before: a successful build could print dozens of Rust never-used warnings, macOS deployment-target warnings, the oversized-module optimization downgrade, and thousands of duplicate-symbol linker lines.

After: default output retains Perry phase/progress and final-artifact messages. Internal toolchain diagnostics remain available under --verbose and are always replayed on failure.

## Checklist

- [x] I have NOT bumped the workspace version or edited CLAUDE.md / CHANGELOG.md
- [x] My commit follows the loose fix: prefix convention
- [x] I have read CONTRIBUTING.md and agree to the Code of Conduct

## Summary by CodeRabbit

- **Improvements**
  - Compilation now consistently uses high-level optimization for native modules, improving predictable performance and scalability.
  - Large generated code is managed through partitioning and structured outlining rather than reduced optimization levels.
  - Successful internal build-tool output is hidden by default for cleaner compile logs.
[...]

PR #8585, opened 2026-08-22
perf(codegen): make native-root text lowering linear in the function (#8583)

proggeramlug:
Part of #8583 (cli.js 2.1.112 compile regression). First of a series; independent of the others.

## What

`function/precise_roots.rs` — the text pass that turns `js_shadow_slot_bind/set` IR into RS4GC's `ptr addrspace(1)` form — was quadratic in two places:

- `lower_roots_for_rs4gc` iterated **every root pointer per IR line**, building up to four `format!` candidate strings per (line, root) pair, plus a second `root_ptrs.iter().any(|p| line.contains(p))` fail-closed scan per line.
- `retype_landing_pads_for_statepoints` rescanned **the whole function per landing pad** (`ir.lines().any(...)` inside the per-line loop).

Both run on every stack-map function at `to_ir()` time, and the codegen-unit partitioner renders every function for its `@symbol` scan (`module.rs:995`) and then `freeze_unit` renders the stack-map ones again. On the Claude Code bundle (68 MB entry function, ~1.2 M lines, thousands of slots) this was the bulk of the 27-minute "partitioning" phase in #8583. None of it ran under the July shadow-stack default — it arrived with native roots.

Now: one shape parse per line (`store`/`load` of `i64`/`double` whose pointer operand is the whole tail after the last `, ptr `) plus a `HashSet` lookup; the fail-closed scan tokenizes once against the same set; the landing-pad retype takes a single mention census of the function and subtracts each pad's own line.

## Output is unchanged — by test, not by argument

The previous implementations are kept **verbatim** as `reference_*` oracles in `precise_roots/equivalence_tests.rs`. A deterministic xorshift generator produces functions covering every accepted shape, `align` suffixes (fail-closed on roots in both versions), prefix traps (`%l_1` vs `%l_1a`, `%lp2` vs `%lp21`), used and dead Itanium pads, asm / audited-leaf / unknown callees, and randomized root order; the tests assert `Option<String>` equality over 1,000 functions and check both verdicts (lowered / fail-closed; retyped / kept) actually occur so the equality isn't vacuous.

Existing pins untouched and green: `audited_capture_and_box_accessors_take_no_statepoint`, `one_logical_slot_can_root_disjoint_physical_allocas`, the `stack_map_tests`, and `native_emit`'s `*_lowers_precise_roots_before_rs4gc`.

## Measured

2,000 roots × ~44k lines (clean function, lowers end to end):

| | lowering | landing-pad retype |
|---|---:|---:|
| reference (before) | 27.6 s | 21.0 s |
| linear (this PR) | 287 ms total | |

`reference_timing_on_the_wide_function` (`#[ignore]`) reproduces the numbers.

## Verification

- `cargo test -p perry-codegen`: all green except 5 `native_proof_regressions::typed_*_method_clone_*` failures that **reproduce with this change stashed** on current `origin/main` (`fefdc367b`) — pre-existing drift, unrelated (`js_method_direct_shape_guard` expectation vs #8573).
- `cargo clippy -p perry-codegen --lib --tests`: no findings in the touched files. `cargo fmt` clean.
- No behaviour change → no docs change.

## Summary by CodeRabbit

* **Performance**
  * Improved code-generation performance for functions with many roots and landing pads.
  * Large functions now process substantially faster, with benchmarked improvements from 49 seconds to 0.3 seconds.

* **Reliability**
  * Preserved byte-identical generated output through expanded equivalence and wide-function testing.
  * Maintained existing root load/store transformations and landing-pad behavior.


proggeramlug:

| check | result |
|---|---|
| all seven ratchets | 0 |

PR #8586, opened 2026-08-22
fix(codegen): delete the unsound pre-RS4GC optnone knob; add a post-rewrite instruction budget and per-unit codegen stats (#8583)

proggeramlug:
Part of #8583. Independent of #8585 (both branch from `main`). No optimization-policy change — this PR deliberately does **not** restore the #8128 `optnone` guard that #8421 removed; see the issue's resolution direction.

## 1. `PERRY_LL_PREOPT_OPTNONE_INSTRS` is deleted (it was unsound, not just uncalibrated)

It stamped `optnone` **before** `rewrite-statepoints-for-gc`. The new pass manager skips `mem2reg`/`sccp` on an `optnone` function, while RS4GC — a module pass keyed on the `gc` attribute — still runs. So a demoted function kept its root allocas unpromoted and the collector never saw them; after a moving minor the alloca holds a stale pointer.

`native_emit::tests::optnone_before_rs4gc_hides_every_root_from_the_collector` proves it on the existing `precise_root_fixture`: the control arm has `gc-live` bundles and `gc.relocate`s; the same fixture with `optnone noinline` on the define line has **no** `gc-live` bundle and its `alloca ptr addrspace(1)` lines survive unpromoted. The cap defaulted to 0, so no shipped build hit it. Any future size policy has to run after the rewrite.

## 2. `PERRY_LL_RS4GC_MAX_INSTRS` — a post-rewrite assertion, never a demotion

After the rewrite and its verify, a function past the per-function budget makes the unit **fail** with the function's name, its size before and after the rewrite, the budget, and the override spellings (`<n>` raise, `warn:<n>` warn only, `0` disable). Nothing is demoted and #8421's contract (every function optimized at the requested level) stands; what changes is that an estimator miss fails in seconds instead of hanging for hours.

Default 1.5 Mi instructions sits between #8128's two measured points on the Next 16.3.0 bundle: ~413k post-rewrite finished comfortably at `-Os`, ~2.1M ran >65 CPU-minutes without finishing. Step 0 of #8583 will recalibrate it on the Claude bundle.

`inprocess::tests::rs4gc_budget_fires_only_on_the_rewritten_module` pins that it is post-rewrite: six gc values live across forty safepoints is under the budget before RS4GC and over it after (each statepoint relocates every live value). Counting before the rewrite — the deleted knob's mistake — makes the "after" set empty and the test red. The message is checked for the function name, both sizes, the budget, the override, and the *absence* of "optnone".

Both the build cache and the object cache key on the new variable; the old key is removed (`object_cache_tests` list updated).

## 3. Per-unit codegen stats (acceptance criterion 3 of #8583)

Under `PERRY_CODEGEN_UNIT_TIMINGS` the native unit pipeline now prints, per unit:

- at freeze time, before LLVM sees it: function count, estimated IR, and the **widest function by estimated IR** — so the irreducible function in a bundle is named up front;
- after compile: function count, instruction totals and widest function **before and after RS4GC**, the growth factor, and rewrite / optimize / emit times.

`optimize_and_emit_module` is unchanged for existing callers; `optimize_and_emit_module_with_stats` fills a `UnitCodegenStats`. Gated on `PERRY_CODEGEN_UNIT_TIMINGS` only (not the default-on `PERRY_CODEGEN_PROGRESS`), so ordinary builds print nothing new.

## Verification

- `cargo test -p perry-codegen`: green except the same 5 `native_proof_regressions::typed_*_method_clone_*` failures that reproduce on pristine `origin/main` (`fefdc367b`) with nothing applied — pre-existing drift vs #8573, unrelated.
- `cargo check -p perry --tests` clean (cache-key edits).
- `cargo clippy -p perry-codegen --lib --tests`: no new findings in the touched code (`optimize_and_emit` was already over clippy's argument limit).

## Summary by CodeRabbit

* **New Features**
  * Added configurable instruction-budget checks after garbage-collection rewriting, supporting error, warning, or disabled modes.
  * Added code-generation reports covering function and instruction counts, instruction growth, widest functions, and phase timings.
* **Bug Fixes**
  * Improved detection and reporting of unusually large generated functions.
  * Updated build and object caching to respect the new instruction-budget setting.
* **Documentation**
  * Documented the new configuration option and expanded code-generation timing reports.
* **Tests**
  * Added coverage for budget parsing, warning/error behavior, instruction growth, and cache invalidation.


proggeramlug:

| check | result |
|---|---|
| all seven ratchets | 0 |

PR #8588, opened 2026-08-22
perf(codegen): inline admitted helpers before statepoints

proggeramlug:
RS4GC rewrites ordinary calls into statepoints before the normal optimization pipeline. Once rewritten, LLVM can no longer honor alwaysinline on the callee, so the proven nonnegative-index clone from #8580 remained as a hot call boundary.

This runs always-inline before mem2reg/SCCP/RewriteStatepointsForGC in both the external and in-process backends, using one shared pass string. Admission is deliberately separate from the historical force_inline policy: under native roots, existing force_inline functions become inlinehint, while only an explicitly budgeted helper gets pre-statepoint alwaysinline. This avoids the broad pre-RS4GC policy that measured as code-size growth; this PR initially admits only the proven index-method clone. Non-RS4GC behavior is unchanged.

Correctness coverage:
- emitted index clone is admitted while the public semantic fallback is not
- historical force_inline stays a hint under native roots and remains unconditional without RS4GC
- an inlined allocating call keeps its caller root live and relocated
- an inlined invoke retains verifier-valid normal/unwind relocation
- both RS4GC backends consume the same pass sequence

Validation on final current main fefdc367b:
- cargo test --release -p perry-codegen --lib: 1135 passed
- cargo fmt --all -- --check
- cargo check --release -p perry-codegen
- cache-disabled RS4GC/in-process LLVM codehz/ecs query, 11 alternating pairs, repeat 64, all 22 processes passed and produced exact accumulation 50,005,000
  - read-only paired median: -2.22%, candidate won 10/11
  - accumulation paired median: -1.92%, candidate won 11/11
  - executable: 11,757,280 -> 11,608,616 bytes (-148,664 / -1.264%)

The host was contended, so the timing cohort is qualification evidence; the paired direction and size result reproduce the earlier accepted prototype. A symbolized build removes the standalone hot readSimpleColumnValue index-clone symbol while retaining the public fallback.

## Summary by CodeRabbit

- **Performance**
  - Improved runtime efficiency by enabling eligible helper methods to be inlined earlier during compilation.
  - Reduced executable size and improved benchmark results.

- **Bug Fixes**
  - Preserved garbage-collection root tracking and exception handling when inlining methods.
  - Maintained existing behavior for non-garbage-collected builds.

- **Tests**
  - Added coverage for inlining, root relocation, exception edges, and direct-method regression scenarios.

### DIFF
--- a/changelog.d/8588-pre-statepoint-inline.md
+++ b/changelog.d/8588-pre-statepoint-inline.md
@@ -1,0 +1,10 @@
+Added a scoped pre-statepoint inline budget for helpers that must cross Perry's
+RS4GC boundary. The proven nonnegative-index method clone is admitted before
+calls become statepoints, while historical force-inline sites remain ordinary
+LLVM hints under native roots; non-RS4GC behavior is unchanged.
+
+On the `codehz/ecs` 10,000-entity query, an 11-pair contended-host A/B against
+current `main` reduced paired medians by 2.22% for read-only iteration and 1.92%
+for accumulation. The candidate won 10/11 and 11/11 pairs respectively, shrank
+the executable by 148,664 bytes (1.264%), and all 22 processes passed the query
+assertions and exact 50,005,000 accumulation oracle.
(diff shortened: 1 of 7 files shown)


proggeramlug:

| check | result |
|---|---|
| all seven ratchets | 0 |

PR #8589, opened 2026-08-22
perf(codegen): spill giant-function GC roots to a shadow frame to bound RS4GC fan-out (#8583)

proggeramlug:
Part of #8583, the reconciled fix. **Stacks on #8586** (branch `fix/8583-rs4gc-budget-assert`) — review that first; this diff is against it, and #8586's post-RS4GC instruction budget is the backstop for the estimate here. No optimization-level change (see the issue's resolution direction and #8586).

## The measurement that drove this

Isolating the Claude Code bundle's giant entry function (`@main`, the 68 MB minified IIFE) and running the arms offline with Homebrew LLVM 22.1.4:

| | result |
|---|---|
| `@main` pre-RS4GC | 439,116 instructions, **795 root slots, ~106,278 call sites** |
| RS4GC on that unit | 62s, 7.1 GB RSS, **439k → 6.53M instructions (~15×)** |
| `-Os` on the un-rewritten unit (this PR's world) | **5.05s, 520 MB** — the whole unit, `@main` included |
| `-Os` on the RS4GC'd unit (today) | did not finish under a 45-min cap |

The pathology is entirely RS4GC relocation fan-out on one function: `relocations ≈ live_roots × safepoints`. Remove the fan-out for that one function and everything optimizes in seconds. The partitioner already isolates it into its own codegen unit.

## What this does

Makes the precise-root **lowering** a per-function choice. A function whose estimate `root_slots × safepoint_sites` (counted from the HIR) exceeds `PERRY_ROOT_SPILL_RELOCATIONS` (default **4,000,000**) keeps its GC roots in the heap **shadow frame** — the pre-#7370 lowering — instead of native statepoints, so RS4GC skips it. `@main`'s estimate is ~8.4e7; ordinary functions are ~1e4–1e5, so the separation is wide and the default sits far below the measured pathological point and far above real code. `=0` disables it (every function on statepoints, prior behavior).

Everything else is unchanged, which is the point of picking this lever over an opt-level downgrade:
- the function is still compiled at the requested optimization level (#8586's #8421 contract holds);
- its roots stay **precise** — the runtime already scans shadow-frame and stack-map roots in one walk (`gc/roots.rs`), and the FP-chain walker steps over frames with no stack-map record;
- the frame pointer is kept on every native-roots build, so a spilled frame is walkable;
- each spilled function is reported at **default verbosity** (#8421: never silent);
- #8586's post-RS4GC instruction budget backstops any function the estimate misses — a loud failure, not a hang.

## Mechanism (no new lowering, no runtime change)

A `force_shadow_frame` flag on `LlFunction` routes `enable_shadow_frame_inner` / `reserve_shadow_slot` into the existing shadow path and drops the `gc "statepoint-example"` strategy in `define_header` (frame pointer kept). `codegen/helpers::maybe_spill_roots_to_shadow_frame` sets it — before the frame is built — at the five sizing sites (function, instance method, static method, closure, module entry). It reuses exactly the code `PERRY_RS4GC=0` already builds and the shadow-stack suites already cover.

## Tests

- `function.rs`: a spilled function drops the statepoint strategy, keeps the frame pointer, and emits the shadow frame, while an un-spilled sibling keeps statepoints (under a `NativeRootsPin::native()` pin so the on-path actually runs).
- `collectors/safepoint_sites`: counts call-like expressions across control flow and does **not** descend into nested closures (they get their own frames).
- **`gc_root_spill_mixed_frames_8583` (end-to-end, the soundness pin):** one program compiled with spilling off (`=0`, all statepoints) and aggressively on (`=1`, so `run`/`make`/`main` spill while the call-free `leaf` stays on statepoints — a genuinely mixed stack), each run under every moving-collector arm; **outputs must be byte-identical**. If a spilled frame's roots were invisible to a relocating minor, only the `=1` arm would diverge. Passes (14s). The `=1` compile is asserted to have actually spilled, so it can't decay into a tautology.

Both caches key on `PERRY_ROOT_SPILL_RELOCATIONS`.

## Verification

- `cargo test -p perry-codegen`: green except the same 5 pre-existing `native_proof_regressions::typed_*_method_clone_*` failures that reproduce on stock `origin/main`.
- `gc_root_spill_mixed_frames_8583`: pass.
- `RUSTFLAGS=-D warnings cargo check -p perry-codegen -p perry --all-targets`: clean. No new clippy findings in the touched files.

## Summary by CodeRabbit

- **New Features**
  - Added automatic GC-root spilling for large functions, preserving precise garbage collection and optimization settings.
  - Added configurable instruction-budget checks with error, warning, or disabled modes.
  - Added optional code-generation statistics, including instruction counts, largest functions, and phase timings.
  - Added per-unit compilation timing and size diagnostics.

- **Bug Fixes**
  - Improved handling of mixed GC root representations during collection.

- **Chores**
  - Updated build and object caches to reflect the new compilation settings.


proggeramlug:

| check | result |
|---|---|
| all seven ratchets | 0 |

proggeramlug:

Issue #4880, opened 2026-06-10
Compile-time blowup on wide object literals (2100 keys ≈ 2 min, 3000 keys > 7 min)


proggeramlug:
Found while stress-testing write barriers (main @ 09760e6dc, macOS arm64, release build of perry).

A single object literal with thousands of keys makes `perry compile` superlinearly slow:

| keys in one literal | compile wall time |
|---|---|
| 300 | seconds |
| 2100 | ~114 s |
| 3000 | killed after ~7 min |

Generator for a repro file:

```bash
python3 -c "
fields = ',\n'.join(f'    f{i}: {{ v: {i} }}' for i in range(2100))
open('wide.ts','w').write('const src: any = {\n' + fields + '\n};\nconsole.log(\"ok\");\n')
"
perry compile wide.ts -o wide_bin   # ~2 minutes
```

The scaling looks roughly quadratic in the number of literal fields, so it's probably an O(n²) walk in object-literal lowering/codegen (shape registration or per-field key scanning). Real-world impact: large generated config/lookup-table modules.

Issue #8590, opened 2026-08-22
Opt target 1 — `typed_array`: 6.29x Node, 8.52x instructions


proggeramlug:
# Opt target 1 — `typed_array`: 6.29x Node, 8.52x instructions

**This is the single largest remaining gap in the corpus, roughly three times worse than
anything else Perry runs.** Take this one first.

## Measurement — quiet M1 mini, 5 shuffled interleaved repeats, verdict CLEAN, perry `fefdc367b`

| | node | perry | ratio |
|---|---:|---:|---:|
| wall | 0.646 s | 4.064 s | **6.29x** |
| instructions | 6.539 G | 55.717 G | **8.52x** |
| cycles | 2.066 G | 13.012 G | 6.30x |
| IPC | 3.16 | **4.28** | — |
| peak RSS | 82.2 MB | 5.3 MB | 0.06x |

**Read the IPC before you start.** Perry retires instructions *more efficiently* than Node here
(4.28 vs 3.16) and still loses 6.29x. Nothing is stalling; the machine is executing the work
well. There are simply 8.5x too many instructions. Do not go looking for a cache or memory
problem — this is a "delete work" task, not a "speed up work" task.

Workload: `gc-handoff/apps/typed_array.ts` — bcryptjs's Blowfish `_encipher` shape, `S`/`P`
Int32Arrays reached through **untyped** parameters, ~4.27 M calls, indexed with `>>>`/`|` and
accumulated with `+`/`^`. Prints only `checksum:-821955270`, which **must not change**.

## What is already known — do not re-derive any of this

- **The S-box element loads are already native.** Optimized AArch64 shows
  `ldr ..., [base, index, uxtw #2]` — no NaN-box round trip, no per-element registry lookup.
  An earlier agent assumed otherwise, built a "focused exact-index optimization", got −24.6%
  instructions almost entirely on the *typed* path, and correctly reverted it because it made
  the untyped/typed ratio worse.
- **This is no longer a #5525 untyped-vs-typed gap.** #8491 closed that: the ratio is 1.0017.
  Both paths are slow *together*. Chasing the ratio chases a solved problem.
- **#8491 already took −44.1% wall / −34.7% instructions / −52.6% RSS out of this workload**
  by making a numeric type proof survive reassignment (`number_by_construction`). What remains
  is a *different* part of the envelope, or something else entirely.

## Method

Profile before hypothesising. Perry strips its output binaries, so compile with
[...]

PR #8593, opened 2026-08-22
perf(codegen): CPU-aware default for native LLVM unit workers (#8583)

proggeramlug:
Part of #8583. **Stacks on #8589** (root spilling). Promoted from the plan's "data-driven maybe" to load-bearing by the acceptance run (below).

## Why

The native codegen-unit pipeline hard-coded **2** LLVM workers (`native_emit.rs`, from #8017 for Windows pagefile pressure, applied on every platform). The text path already scales with the machine (`available_parallelism()/4`, clamped).

On the acceptance compile of `cli.js` (~84 codegen units) with #8589's spill in effect, per-unit timings (from #8586's `PERRY_CODEGEN_UNIT_TIMINGS`) showed a typical dense-closure unit at **437.9s** (`rs4gc 121s, opt 39s, emit 272s`). With the giant entry function's roots spilled, no unit carries an unbounded RS4GC fan-out and per-unit peak RSS is a bounded ~1–2 GiB — so the **two-worker cap, not memory, is the wall**: dozens of multi-minute units, two at a time, is hours.

## What

Default workers = half the logical CPUs, clamped `[2, 8]`, on non-Windows; **Windows keeps 2** (the platform #8017's cap was chosen for, until its pagefile behavior under higher fan-out is measured). `PERRY_CODEGEN_UNIT_JOBS` still overrides everywhere. The 8 ceiling keeps peak at ~8 × the bounded per-unit RSS against a 64 GiB-class host with margin.

Bounds test: `default_unit_workers()` is `2` on Windows and in `[2, 8]` elsewhere.

## Honest scope note

This makes the compile *complete in bounded time* rather than run for hours, which is what #8583's acceptance criterion asks. It does **not** restore the pre-regression ~40–50 min wall on its own: the per-unit floor is the #4880 giant-generated-function ISel/emit cost (a 250k-instruction minified closure is ~270s in `emit` regardless of GC), whose `-O0` fallback #8421 also removed. That is orthogonal to GC roots and is tracked as the structured-outlining follow-up in #8583. Higher worker counts divide that floor; they do not remove it.

## Summary by CodeRabbit

* **New Features**
  * Added automatic GC-root spilling for functions with high relocation estimates, with configurable thresholds and diagnostics.
  * Added configurable post-compilation instruction limits with error, warning, or disabled modes.
  * Added detailed per-unit compilation statistics, including instruction growth, largest functions, and phase timings.
  * Increased default native code-generation parallelism based on available CPU capacity on non-Windows platforms.

* **Bug Fixes**
  * Improved garbage-collection correctness for programs mixing native and shadow-frame roots.
  * Updated build caching to reflect the new compilation and root-spilling settings.


proggeramlug:

| check | result |
|---|---|
| all seven ratchets | 0 |

proggeramlug:

PR #8594, opened 2026-08-22
perf(codegen): retain loop ranges across update expressions

### LINKED ISSUE #8590: Opt target 1 — `typed_array`: 6.29x Node, 8.52x instructions

proggeramlug:
## Summary

Keep loop-counter range proofs alive across prefix/postfix update expressions so proven typed-array indices such as `P[++i]` lower to native loads instead of dynamic typed-array property access.

## Changes

- Derive the returned integer range of prefix and postfix increment/decrement expressions from the local's active range.
- Conservatively widen active range facts across updates, covering both pre- and post-update values so branch lowering cannot assume a conditional update occurred.
- Add an IR regression test covering the two `P[++i]`-shaped reads in a guarded untyped-parameter clone.

## Related issue

Fixes #8590

## Test plan

- [x] `CARGO_PROFILE_RELEASE_CODEGEN_UNITS=16 cargo build --release -p perry`
- [x] `CARGO_PROFILE_RELEASE_CODEGEN_UNITS=16 cargo test --release -p perry-codegen --lib` — 1,133 passed
- [x] `cargo fmt --all -- --check` / `git diff --check`
- [x] 20-row corpus output gate: all byte-exact; 19/20 linked executables byte-identical to `fefdc367b`, with only `typed_array` changed
- [ ] `cargo test --workspace --exclude perry-ui-ios --exclude perry-ui-tvos --exclude perry-ui-watchos --exclude perry-ui-gtk4 --exclude perry-ui-android --exclude perry-ui-windows` passes (not run locally; scoped codegen suite above)
- [ ] (if CLI / stdlib / runtime API changed) Updated `docs/src/` — n/a
- [ ] (if touching a platform UI backend) Built `-p perry-ui-<backend>` locally on that platform — n/a

Quiet M1 mini, five shuffled interleaved baseline/fixed repeats, hardware counters repeated five times; clean verdict (load 1.37 → 1.94):

| `typed_array` | `fefdc367b` | fixed | delta |
|---|---:|---:|---:|
| wall, best of 5 | 4.063 s | 2.498 s | **-38.51%** |
| instructions, median | 55.717 G | 14.647 G | **-73.71%** |
| cycles, median | 13.012 G | 7.999 G | **-38.53%** |
| peak RSS, median | 5,521,408 B | 5,455,872 B | **-1.19%** |
| untyped/typed ratio, median | 1.000 | 1.000 | unchanged |
| untyped/typed ratio, aggregate | 1.00021 | 1.00011 | unchanged |

Against the issue's Node baseline, Perry moves from 6.29x to 3.87x wall and from 8.52x to 2.24x retired instructions. The two generated specialized bodies lose the two per-round `js_typed_array_index_get_dynamic` calls and the final `js_typed_array_get` call together, preserving parity between typed and untyped paths.

Whole-corpus rows moving more than 1% by best-of-5: `typed_array` -38.51%; `churn_read` +1.53%. The latter executable is byte-identical between arms and its median moved +0.44%, so that difference is measurement noise. No other row moved by more than 1%.

## Screenshots / output

```text
checksum:-821955270
```

## Checklist

## Summary by CodeRabbit

- **Performance**
  - Improved optimization of bounded array index updates, reducing generated instructions, execution time, and peak memory usage in affected workloads.
  - Eliminated unnecessary dynamic typed-array access paths in optimized scenarios.

- **Bug Fixes**
  - Improved handling of prefix and postfix increment/decrement expressions in integer range analysis.
  - Preserved accurate range information across updates, supporting more reliable optimized execution.

- **Tests**
  - Added regression coverage for optimized typed-array reads using incremented indices.

### DIFF
--- a/changelog.d/8594-prefix-update-index-ranges.md
+++ b/changelog.d/8594-prefix-update-index-ranges.md
@@ -1,0 +1,7 @@
+Eliminated dynamic typed-array property lookups for bounded update-expression
+indices such as `P[++i]`. Integer range analysis now models the distinct prefix
+and postfix results and conservatively retains loop facts across updates. On
+the Blowfish-shaped `typed_array` workload this reduced retired instructions
+from 55.717 G to 14.647 G (-73.71%) and wall time from 4.063 s to 2.498 s
+(-38.51%) on the quiet M1 mini, while the typed/untyped ratio stayed at 1.000
+and peak RSS decreased 1.19%.
--- a/crates/perry-codegen/src/codegen/ordinary_param_guard_tests.rs
+++ b/crates/perry-codegen/src/codegen/ordinary_param_guard_tests.rs
@@ -5,7 +5,7 @@
 
 use crate::{compile_module, CompileOptions};
 use perry_hir::types::{ObjectType, PropertyInfo, Type};
-use perry_hir::{BinaryOp, CompareOp, Expr, Function, Module, Param, Stmt, TypeAlias};
+use perry_hir::{BinaryOp, CompareOp, Expr, Function, Module, Param, Stmt, TypeAlias, UpdateOp};
 use std::collections::HashMap;
 
 fn function_ir<'a>(ir: &'a str, marker: &str) -> &'a str {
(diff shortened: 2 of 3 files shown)


proggeramlug:

| check | result |
|---|---|
| all seven ratchets | 0 |

PR #8597, opened 2026-08-22
perf(codegen): inline short heap string equality

### LINKED ISSUE #8591: Opt target 2 — `iso_miss` (2.11x) and `interp` (1.48x): tree-walking interpreters
# Opt target 2 — `iso_miss` (2.11x) and `interp` (1.48x): tree-walking interpreters

Treat these as **one task**. They share their shape, and a win on one plausibly moves the
other — check whether it does and report both either way.

| | node | perry | ratio |
|---|---:|---:|---:|
| **iso_miss** wall | 0.338 s | 0.714 s | **2.11x** |
| — instructions | 6.264 G | 12.462 G | **1.99x** |
| — cycles | 1.122 G | 2.281 G | 2.03x |
| — IPC | 5.58 | 5.46 | — |
| — peak RSS | 89.4 MB | 30.9 MB | 0.35x |
| **interp** wall | 0.324 s | 0.481 s | **1.48x** |
| — instructions | 5.973 G | 8.643 G | **1.45x** |
| — cycles | 1.079 G | 1.530 G | 1.42x |
| — IPC | 5.54 | 5.65 | — |
| — peak RSS | 94.7 MB | 31.3 MB | 0.33x |

**Both are genuinely CPU-bound and the diagnosis is unambiguous**: instruction ratio, cycle
ratio and wall ratio all agree to within a few percent, and IPC is essentially identical to
Node's. Perry is executing the same quality of code, just ~2x and ~1.45x more of it. Fewer
instructions is the only lever; there is no scheduling or stall story to find.

Workloads: `gc-handoff/apps/iso_miss.ts` and `apps/interp.ts` — tree-walking interpreters for
small functional languages. Recursive union allocation, closures capturing environments,
polymorphic dispatch, string building, deep recursion, string-keyed Maps.

## History worth knowing

- `iso_miss` regressed **+24.8%** at #8417 (a codegen string-accumulator change). #8487
  recovered most of it (14.788 G → 12.470 G instructions, −15.7%), and #8581's join/repeat work
  landed since. Roughly +5% of that original regression was never accounted for.
- `interp` improved −4.61% instructions at #8539 (per-object layout-mask cutoff 2 → 4). That
  came from profiling, which found `evalNode` at 61.1% of self samples across 30 runs — **not**
  the dynamic-numeric-envelope problem the brief had guessed. Profile before assuming.
- An unattributed step in the 08-21 window moved `iso_miss` +7.9%; it has since been flat, so it
  was a one-time change rather than a trend. Attribution is still open if you want it.

## Method

Profile first, with `PERRY_KEEP_SYMBOLS=1 PERRY_DEBUG_SYMBOLS=1`, aggregated over many runs —
these programs are 300–700 ms and a single `sample` yields almost nothing.

Because IPC already matches Node's, the question is *what work exists that Node does not do*:
redundant guards, allocation Node avoids, dispatch Node caches, or a per-node cost in the
interpreter loop that should be hoisted.

## Acceptance

- Instruction reduction on either row, with the effect on **both** reported.
  more than 1% reported.
- RSS is already 0.33–0.35x Node. Do not trade it for speed; quantify any cost.
- Instructions retired is the primary signal — the bench host is contended and wall ranges
  overlap.

proggeramlug:
## Summary

Inline equality for statically proven heap strings when their lengths differ or their equal-length payloads contain at most three bytes. This targets the short identifier comparisons that dominate the tree-walking interpreter workloads in #8591 while keeping generic-key and long-string comparisons on the existing helper path.

## Changes

- Preserve the existing identity, tag, SSO, and invalid-low-pointer handling.
- Compare heap-string length, first byte, last byte, and (for three-byte strings) middle byte before calling `js_string_equals`.
- Keep the larger prefix out of string-vs-generic-key comparisons.
- Add IR coverage for the short-string blocks and long-string fallback, plus parity coverage for distinct dynamically allocated strings of lengths 0–3 and longer.

## Related issue

Fixes #8591

## Performance

Measured against `fefdc367b` on the issue's quiet M1 mini with five shuffled interleaved repeats. Control and candidate used the same freshly rebuilt runtime archives and matching binary basenames. All 20 corpus programs were byte-exact against Node; only `interp` and `iso_miss` produced different binaries, and no other corpus row moved.

| workload | retired instructions | median wall | max RSS | peak footprint |
|---|---:|---:|---:|---:|
| `interp` | 8.642 G → 8.433 G (**-2.41%**) | 0.479 s → 0.460 s (-3.93%) | 32.92 MB → 32.90 MB (-0.05%) | 27.07 MB → 27.03 MB (-0.12%) |
| `iso_miss` | 12.461 G → 12.259 G (**-1.62%**) | 0.713 s → 0.700 s (-1.86%) | 32.46 MB → 32.49 MB (+0.10%) | 26.84 MB → 26.81 MB (-0.12%) |

Verdict: CLEAN. No timing, instruction, or RSS result outside the two target rows moved more than 1%.

## Test plan

- [x] `cargo build --release` clean (isolated target directory)
- [ ] Full workspace test suite (the affected-crate suites were used instead)
- [x] Added or updated coverage under `test-files/` and in the affected crate
- [x] Docs not needed: no CLI, stdlib, runtime API, or UI change
- [x] UI backend build not applicable

Commands/checks:

- `cargo test -p perry-codegen --lib --no-fail-fast` — 1,133 passed
- `cargo test -p perry-runtime --lib --no-fail-fast` via `scripts/test_affected_crates.sh` — 2,633 passed, 4 ignored
- Perry CLI suite via `scripts/test_affected_crates.sh` — 1,025 passed; the only failure is the pre-existing stale `ext_zlib_covers_every_stdlib_symbol_the_flip_strips` ratchet in untouched code
- `run_parity_tests.sh --filter test_strict_eq_string_literal_inline` against pinned Node 26.5.1 — passed, byte-exact
- `python3 scripts/check_test_registration.py` — passed
- `scripts/pre-tag-check.sh --quick` — passed
- `cargo fmt --all -- --check` — passed
- 20-program sweep corpus — byte-exact against Node

## Screenshots / output

Not applicable.

## Checklist

## Summary by CodeRabbit

* **Performance**
  * Improved equality checks for short heap-allocated strings by comparing length and content inline before using the general fallback.
  * Reduced instructions for common string comparisons without changing memory usage or generic-key behavior.

* **Bug Fixes**
  * Preserved correct strict-equality results for strings of varying lengths, including substring-generated values.

* **Tests**
  * Added coverage for matching and mismatching short strings, longer strings, and generic key comparisons.

### DIFF
--- a/changelog.d/8597-short-heap-string-equality.md
+++ b/changelog.d/8597-short-heap-string-equality.md
@@ -1,0 +1,5 @@
+Improved equality checks between statically proven strings by comparing heap-string
+lengths and up to three payload bytes inline before falling back to the full runtime
+helper. This targets the short identifiers used by tree-walking interpreters: retired
+instructions fell 2.41% for `interp` and 1.62% for `iso_miss`, with RSS effectively
+unchanged. Generic-key comparisons retain the smaller existing dispatch.
--- a/crates/perry-codegen/src/expr/compare.rs
+++ b/crates/perry-codegen/src/expr/compare.rs
@@ -432,6 +432,119 @@ fn lower_strict_eq_inline_any(ctx: &mut FnCtx<'_>, l: &str, r: &str) -> String {
     )
 }
 
+/// Resolve equal-length heap strings of up to three bytes inline, and reject
+/// longer pairs on a length or endpoint mismatch before using the full helper.
+/// The caller has already proven both values carry `STRING_TAG`.
+fn lower_short_heap_string_eq(
+    ctx: &mut FnCtx<'_>,
+    lh: &str,
+    rh: &str,
+    true_l: &str,
+    false_l: &str,
+    merge_l: &str,
+) -> (String, String) {
+    let heap_valid_idx = ctx.new_block("streq.heap.valid");
+    let heap_len_idx = ctx.new_block("streq.heap.len");
+    let heap_first_idx = ctx.new_block("streq.heap.first");
+    let heap_one_idx = ctx.new_block("streq.heap.one");
+    let heap_last_idx = ctx.new_block("streq.heap.last");
+    let heap_two_idx = ctx.new_block("streq.heap.two");
+    let heap_middle_idx = ctx.new_block("streq.heap.middle");
+    let heap_middle_byte_idx = ctx.new_block("streq.heap.middle.byte");
+    let heap_slow_idx = ctx.new_block("streq.heap.slow");
+    let heap_valid_l = ctx.block_label(heap_valid_idx);
+    let heap_len_l = ctx.block_label(heap_len_idx);
+    let heap_first_l = ctx.block_label(heap_first_idx);
+    let heap_one_l = ctx.block_label(heap_one_idx);
+    let heap_last_l = ctx.block_label(heap_last_idx);
+    let heap_two_l = ctx.block_label(heap_two_idx);
+    let heap_middle_l = ctx.block_label(heap_middle_idx);
+    let heap_middle_byte_l = ctx.block_label(heap_middle_byte_idx);
+    let heap_slow_l = ctx.block_label(heap_slow_idx);
+
+    // Preserve `js_string_equals`' handling of deliberately forged low
+    // `STRING_TAG` payloads: only dereference real heap addresses here.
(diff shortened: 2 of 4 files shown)


proggeramlug:

Issue #8596, opened 2026-08-22
gc: reduce safepoint density (safepoint polling instead of every-call statepoints)


proggeramlug:
## Summary

Perry treats **every call** as a GC safepoint: `rewrite-statepoints-for-gc` wraps each non-leaf call in a `gc.statepoint` and relocates every live root across it. That is correct (any callee might collect) but it makes the safepoint count of a function equal to its call count, which is the `× safepoints` term in the RS4GC relocation fan-out `live_roots × safepoints` behind #8583 (the Claude Code `@main` has ~106k call sites).

Many production GC'd VMs instead use **safepoint polling**: collection can only begin at a bounded set of poll points — loop back-edges, function returns/entries, and allocation sites — and ordinary calls are *not* safepoints unless the callee itself polls. This shrinks the safepoint set dramatically without losing the guarantee that a collection can always make progress, because a thread reaches a poll in bounded time.

## Proposal (exploratory — needs design)

Move from "every call is a safepoint" toward a polling model:

- take safepoints at loop back-edges, allocation sites, and returns, rather than at every call;
- a non-allocating / non-polling call is a `gc-leaf`-style edge that does not relocate roots;
- keep a mechanism to bring a thread to a safepoint on demand (poll flag the callee checks), which perry already has pieces of (`GC_SAFEPOINT_PENDING`, `js_gc_loop_safepoint`, the loop-poll path).

## Why it's worth it

Unlike #8595 (outlining), which fixes the giant-function symptom, reducing safepoint density shrinks the `× safepoints` factor for **every** function in the codebase, not just the outlier — lowering RS4GC cost and stack-map size across the board, and reducing the number of points at which roots must be kept live (better register allocation).

## Scope / risk

This is the largest of the three and a genuine GC-architecture change: it affects collection latency/progress guarantees, the interaction with the moving scavenge (#7019), write barriers, and the conservative-scan fallback. It should be scoped and prototyped carefully — filed here so the option is tracked rather than lost, not as ready work.

Context: surfaced while fixing #8583 (RS4GC relocation fan-out). Complements #8595 (outlining) and the root-spilling stopgap in #8583. Root-lowering background: #7370 (native roots default), #7280/#7173/#7174.

proggeramlug:
## Safe increment filed: #8598

[...]

Issue #8592, opened 2026-08-22
Opt target 3 — `pipeline`: 1.18x Node, but **1.85x instructions at 1.05x cycles**


proggeramlug:
# Opt target 3 — `pipeline`: 1.18x Node, but **1.85x instructions at 1.05x cycles**

This one is a different animal from the other two, and the difference is the whole ticket.

## Measurement — quiet M1 mini, 5 shuffled interleaved repeats, verdict CLEAN, perry `fefdc367b`

| | node | perry | ratio |
|---|---:|---:|---:|
| wall | 0.098 s | 0.115 s | **1.18x** |
| instructions | 1.211 G | 2.243 G | **1.85x** |
| cycles | 0.347 G | 0.364 G | **1.05x** |
| IPC | 3.49 | **6.16** | — |
| peak RSS | 84.8 MB | 21.2 MB | 0.25x |

**Perry retires 85% more instructions but spends only 5% more cycles**, at nearly double Node's
IPC. That combination says the extra instructions are cheap, well-predicted, and largely free —
so cutting the instruction count is *not* where the 18% comes from. The wall gap is somewhere a
sampling profiler cannot see it.

**`asyncpipe` had exactly this shape and it has now been fixed** — it sat at 1.07x wall against
1.71x instructions, and #8570 flipped it to **0.84x** by suppressing redundant same-thread
promise-drain wakeups and removing per-element closure/guard/chained-promise allocations from
the intrinsic `Promise.all` path. Its instruction ratio fell 1.71x → 1.31x as a side effect.
That fix is the strongest available hint for this row, and #8243 has prior analysis.

## What NOT to do

Do not reach for a sampler and start cutting instructions. `asyncpipe` taught this: aggregating
`sample` over 14 runs of an ~88 ms program yielded **one** usable self-time sample, because the
process is almost always parked rather than running. A profiler will show you a flat, uninformative
profile and you will conclude there is nothing to fix.

## Method

Instrument the event loop and the scheduling machinery rather than the code:

- count and time event-loop wakeups, and how many are redundant with a drain already in flight
- count microtask-queue drains, and jobs consumed per drain
- count timer registrations and their latency from due-time to run
- look for per-element allocations on the intrinsic paths, as #8570 found in `Promise.all`

`PROFILE_NOTIFY_DRAIN_SUPPRESSED_COUNT` exists from #8570 and is a working example of making
this class of subject countable rather than merely observable.

## Acceptance

[...]

PR #8598, opened 2026-08-22
perf(gc): mark four more non-collecting helpers as gc-leaf (#8596)

proggeramlug:
Incremental slice of #8596 (reduce safepoint density). Independent of the #8583 series (branches off `main`).

## What

`rewrite-statepoints-for-gc` treats every non-leaf call as a GC safepoint and relocates every live root across it. Four runtime helpers are provably non-collecting but were classified `Unknown` (a safepoint) in `gc_call_effects::classify_direct_callee`. They're now `CannotCollect` (stamped `gc-leaf-function`, so RS4GC skips them):

| helper | body | why it can't collect |
|---|---|---|
| `js_typed_feedback_closure_direct_call_guard` | `typed_feedback/guards.rs:1102` → `guard_observe` (`typed_feedback.rs:834`) | header + registry **reads**, then a TLS registry borrow, saturating counters, and a `Vec::push`. That push is a raw Rust allocation — which, by the rule this table already documents, cannot arm a Perry-GC trigger (the malloc-count threshold counts *Perry GC objects*, not system mallocs). No Perry alloc, no JS re-entry, no throw. Same class as the already-admitted observe guards; unlike the excluded get/set wrappers it does no object get/set. |
| `js_implicit_this_get` | `object/this_binding.rs:120` | `IMPLICIT_THIS.with(\|c\| f64::from_bits(c.get()))` — a bare TLS `Cell` read, the exact shape of the already-admitted `js_implicit_this_set` and `js_new_target_get`. |
| `js_tdz_suppress_begin` / `_end` | `box.rs:1030/1036` | a thread-local `Cell<u32>` saturating inc/dec. |

Each was verified against its runtime body (not taken on trust — this is a GC-correctness change: a wrongly-classified leaf = a missed safepoint = potential heap corruption) and each is already in `scripts/gc_root_dominance_check.py`'s `NONCOLLECTING` authority, so the existing subset-containment test stays green.

## Tests

- `issue_8596_tls_and_feedback_guards_cannot_collect` — pins the four as `CannotCollect`.
- `issue_8596_allocating_lookalike_stays_a_safepoint` — pins `js_implicit_this_get_sloppy` (it boxes booleans/strings and reads globalThis → allocates) as still `Unknown`. The admitted `_get` is the only leaf-safe half.
- `cargo test -p perry-codegen -- gc_call_effects`: 10/10 green (incl. the checker-authority containment test).

## Honest scope

This trims relocations and stack-map size across **all** functions, but it is **not** the fix for #8583's fan-out — `@main`'s ~5,600 safepoints are dominated by genuine user-function calls and allocating runtime calls, none of which can be leaf-marked. The real `× safepoints` lever is a safepoint-polling model (collection only at loop back-edges / allocations / returns), which is a large GC-architecture change tracked in #8596 with design notes; this PR is the zero-risk allowlist increment.

## Bonus finding (not fixed here)

While auditing, two helpers the Python `NONCOLLECTING` authority lists actually **allocate** — `js_closure_unbox_callee_checked` (its non-callable path throws a TypeError, which allocates + unwinds) and `js_class_object_pin_parent` (calls `js_string_from_bytes` + `js_object_set_field_by_name`). The Rust table correctly leaves both `Unknown`, so there's no live miscompile, but `gc_root_dominance_check.py` could pass a build with stale roots across such a call. Worth a separate look at the checker's set; flagged in #8596.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runtime garbage-collection safety by correctly identifying helper operations that cannot trigger collection.
  * Updated throw-related handling to use more conservative behavior where necessary.
* **Tests**
  * Added coverage for helper classifications, including allocation-related edge cases.
* **Documentation**
  * Documented the updated runtime helper classifications.


proggeramlug:

| check | result |
|---|---|
| all seven ratchets | 0 |

PR #8599, opened 2026-08-22
perf(codegen): optimize pipeline function calls

### LINKED ISSUE #8592: Opt target 3 — `pipeline`: 1.18x Node, but **1.85x instructions at 1.05x cycles**

proggeramlug:
## Summary

- lower immutable aliases of same-module function declarations through the direct `FuncRef` call path, even when the HIR binding type is erased to `Any`
- add an arrow-aware receiverless one-argument closure dispatcher, reusing the dispatch cache to skip redundant dynamic-`this` save/restore work for arrows while preserving ordinary-call binding for non-arrows, bound functions, and proxies
- add codegen/runtime regression coverage and a performance changelog fragment

## Root cause

`PERRY_MT_PROFILE=1` reports zero promises, timers, notifications, and waits for this target, so the issue's scheduling hypothesis is not active here. The workload is synchronous and CPU-bound. Its hot path instead paid for:

- 360,000 dynamic calls through `const idf = identity`, because the alias was typed as `Any`
- 1,080,000 receiverless arrow-stage calls that saved and restored `IMPLICIT_THIS`, even though arrows use lexical `this`

## Performance

Measured on the quiet M1 runner against the issue baseline (`fefdc367b`), with load below 2.5, exact stdout/stderr verification, and 31 interleaved samples:

| Metric | Baseline | Fix | Change |
| --- | ---: | ---: | ---: |
| Perry wall median | 115.063 ms | 109.217 ms | **-5.08%** |
| Instructions retired | 2,243,016,881 | 2,127,175,316 | **-5.16%** |
| Cycles elapsed | 362,915,723 | 346,079,735 | **-4.64%** |
| Maximum RSS | 22,233,088 B | 22,167,552 B | -0.29% |

The Perry timing ranges are disjoint: 114.696–115.741 ms before and 108.953–110.311 ms after. Node's median in the same run was 98.025 ms, reducing Perry/Node from 1.174x to 1.114x. Node's hardware counters are process-wide and include V8 helper threads, so the counter deltas above intentionally compare Perry only.

The fixed 20-program corpus used 7 interleaved samples per baseline/fix cell. `pipeline` was the only movement over 1% at -5.05%; there were no regressions over 1%, and the largest non-target movement was `retain1` at +0.71%. Every run verified exact stdout and stderr.

## Validation

- `cargo test -p perry-runtime --lib` (2,634 passed, 4 ignored)
- `cargo test -p perry-codegen --lib` (1,143 passed, 1 ignored on current main)
- `git diff --check`
- exact output validation across all 20 benchmark programs

Fixes #8592

## Summary by CodeRabbit

* **Performance**
  * Improved dynamic pipeline processing by using faster direct calls for immutable function aliases.
  * Reduced overhead when invoking receiverless one-argument callbacks.

* **Bug Fixes**
  * Improved callback behavior by correctly preserving or clearing `this` where appropriate.
  * Ensured function aliases continue to work correctly even when type information is generalized.

* **Tests**
  * Added regression coverage for optimized pipeline calls, mutable aliases, and callback receiver behavior.


proggeramlug:

| check | result |
|---|---|
| all seven ratchets | 0 |

Issue #8595, opened 2026-08-22
codegen: structured intra-function outlining of oversized generated functions (module-entry IIFE)


proggeramlug:
## Summary

Perry compiles a module's entire top level into a **single LLVM function** (`@main`). For a large minified bundle that function is enormous — the Claude Code `cli.js` entry is ~68 MB of IR, 439k instructions, with ~13,170 GC-root slots. Essentially every backend algorithm is super-linear in single-function size, so one giant function is simultaneously pathological for:

- **`rewrite-statepoints-for-gc`** — relocation fan-out is `live_roots × safepoints`; on `@main` that expands 439k → 6.5M instructions and the subsequent `-Os` never finishes (this is the root cause behind #8583).
- **instruction selection / emit** — a 250k-instruction dense generated closure costs ~270s in SelectionDAG regardless of GC (the #4880 pathology, whose `-O0`/FastISel fallback was removed in #8421).
- register allocation, the optimizer pipeline, and peak compiler RSS, all for the same reason.

#8583 works around the GC half by *spilling* the giant function's roots to the shadow frame so RS4GC skips it. That is a correct stopgap but a heuristic bandage: it leaves a threshold to calibrate, does nothing for the ISel/emit cost, and keeps producing the giant function that every other backend pass then struggles with. `module.rs`'s own comment already flags this: the 68 MB IIFE is "irreducible without structured intra-function outlining."

## Proposal

Outline the module-entry body (and, more generally, any oversized generated function) into many smaller functions **in codegen, before LLVM sees it**:

- chunk the top-level statement stream (`codegen/entry.rs::compile_module_entry`) into ordered range-functions called in sequence;
- promote every module-level `let` that crosses a chunk boundary to a global — the mechanism already exists (`module_globals_emit.rs` globalizes lets that escape into closures);
- preserve evaluation order and TDZ semantics across the chunk boundaries.

## Why this is the architecturally correct fix (not a workaround)

- **Bounds RS4GC fan-out per chunk** (small live-set × modest safepoints), so we keep **native statepoint roots everywhere** — the fast-at-runtime default — with no spill heuristic. Once this lands, #8583's root-spilling should almost never fire and could be reduced to a rarely-hit safety net.
- **Fixes the #4880 ISel/emit cost** at the source: small functions select fast.
- Lowers peak compiler RSS and improves codegen-unit balancing (the partitioner can place many small functions instead of isolating one monolith).

## Scope / risk

[...]

PR #8600, opened 2026-08-22
perf(codegen): module-entry outlining analysis + reporting (#8595)

proggeramlug:
First, deliberately-safe increment of #8595 (structured outlining of the oversized module-entry function). Independent of the #8583 series (branches off `main`). Full design is in [#8595's implementation-plan comment](https://github.com/PerryTS/perry/issues/8595).

## Why only the analysis

Outlining the entry body is surgery on the single most semantically load-bearing path — all module initialization runs there. A rushed transform that gets eval order, TDZ, hoisting, or top-level await wrong would silently miscompile every program. So this PR ships **only the analysis + gate + reporting** (no transform, no codegen change); the transform lands separately once it can be validated end-to-end on the gap suite and the `cli.js` acceptance compile.

## What's here (`codegen/entry_outline.rs`)

- **`chunk_ranges`** — the boundary rule: contiguous top-level statement ranges of a target size, split ONLY between top-level statements. A top-level `if`/`for`/`try` (and all its break/continue/return/landing-pad flow) stays wholly within one chunk, which is what lets the future transform avoid cross-chunk control flow entirely.
- **`analyze_entry_outlining`** — chunk count + the **cross-chunk `let` set**: bindings defined in one chunk and referenced from another, computed with the very same `collect_let_ids` / `collect_ref_ids_in_stmts` collectors `emit_module_globals` uses. These are exactly the lets the transform must globalize so chunks share state — and since the existing escape rule already globalizes any let referenced from a separate function body, once chunks become functions they globalize for free. This analysis is the reusable core the transform will consume.
- Gates: `has_top_level_await` (chunking across an async suspension is a distinct, harder transform → not a candidate), plus env knobs `PERRY_OUTLINE_ENTRY` (the transform gate, inert today), `PERRY_OUTLINE_ENTRY_REPORT` (prints the analysis), `PERRY_OUTLINE_ENTRY_CHUNK_STMTS` (target).

Wired into `compile_module` behind `PERRY_OUTLINE_ENTRY_REPORT` — zero effect on output otherwise. It lets us measure the outlining opportunity on any real bundle before touching codegen (e.g. on `cli.js`, expect one giant `main` → dozens of chunks + the cross-chunk let count that becomes globals).

## Tests

Pure-core tests (explicit chunk target, no process-global env → no parallel-test races): contiguous chunking, cross-chunk detection, own-chunk confinement (a let used only in its chunk does not cross), and the top-level-await gate. `-D warnings` clean.

## Next (separate PRs, per the design)

PR 2: the transform — hoist top-level `let` declarations, split initializers into chunk functions, replace `hir.init` with ordered chunk calls; validate on toy programs (incl. a cross-chunk let under a forced moving GC), the gap suite, then `cli.js`. PR 3: flip the size-gated default once the corpus is clean. That's what makes #8583's root-spill fire rarely and removes the #4880 emit floor.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in diagnostics for analyzing module initialization entry outlining.
  * Reports statement chunk boundaries, cross-chunk `let` bindings, and modules excluded due to top-level `await`.
  * Enable reporting with `PERRY_OUTLINE_ENTRY_REPORT=1`.

* **Documentation**
  * Added changelog documentation for the new analysis and reporting behavior.


proggeramlug:

| check | result |
|---|---|
| all seven ratchets | 0 |

PR #8601, opened 2026-08-22
test(zlib): ratchet implemented async symbols

proggeramlug:
## Summary

- remove the stale known-gap exemptions for `js_zlib_deflate_raw`, `js_zlib_inflate_raw`, and `js_zlib_unzip`
- make the optimized-library coverage test require the async zlib exports that are already implemented and dispatched by `perry-ext-zlib`

This is a test-ratchet correction only; it changes no runtime behavior. It also fixes the current base-related `cargo-test` failure observed while validating #8588.

## Test

- `CARGO_TARGET_DIR=/tmp/perry-pr-prestatepoint/target cargo test --release -p perry --bin perry commands::compile::optimized_libs::tests::ext_zlib_covers_every_stdlib_symbol_the_flip_strips -- --exact --nocapture`


proggeramlug:

| check | result |
|---|---|
| all seven ratchets | 0 |

PR #8604, opened 2026-08-22
perf(codegen): outline eligible module-entry bodies into chunk functions (#8595)

proggeramlug:
PR 2 of #8595 — the outlining **transform** (analysis #8600 is merged). Off by default, fail-safe. Updated: now **interleaves** chunks around inline statements, and the coupling constraints have been mapped empirically.

## Why this is a runtime-perf change, not just compile-time

The whole module top level compiles to one LLVM function. On cli.js that's ~68 MB of IR / ~13,170 GC-root slots. A function that size is (a) the root of #8583's RS4GC relocation fan-out, (b) the #4880 ISel/emit cost, and — the point for *runtime* — (c) **under-optimized**: size-capped LLVM passes bail or degrade on giant functions. Splitting it lets every pass run to full effect → faster code. It's also the successor that lets #8583's root-spill fire ~never.

## What it does (`PERRY_OUTLINE_ENTRY=1`, HIR phase)

- Walk `hir.init` in order, accumulate runs of **relocatable** statements (plain expressions, single-binding `let`s) into chunk functions, and **interleave** them around any statement that must stay inline (control flow, etc.). So a body only needs *runs* of plain statements, not a wholly plain body; eval order and any `hir.init` scan of an inline statement are preserved.
- Hoist each top-level `let` as a bare declaration to the front and move its initializer into a chunk as a `LocalSet`. The existing `emit_module_globals` escape rule then globalizes exactly the lets referenced across chunks — **no new global-promotion code**.
- Emit each run as a fresh function (id minted above every existing function *and* nested-closure `func_id`); replace the body with hoisted decls + interleaved chunk calls / inline statements.

## Fail-safe, off by default

Pre-scan restores the original body unless outlining would emit >1 chunk. A module is skipped if a codegen scan reads `hir.init` directly (exports / globalThis / `process.env` literals / script-globals) or it has top-level await.

## Empirical coupling map (new)

I probed the coupling bails with a temporary override (small compiles, run under moving GC, outline-off vs on): **exports, globalThis, and `process.env`-literals all outline and produce byte-identical correct output** on toy entries. So the bails are conservative-but-not-required-for-correctness — relaxation candidates — but broad relaxation needs the gap suite (re-exports, default exports, DCE on `process.env.NODE_ENV`, the namespace populator), so this PR keeps them and I removed the override rather than ship a "may-miscompile" escape hatch. That's the concrete next step toward cli.js.

## Validation

- Unit: chunking, decl-hoist/init-split, **interleaving around a must-stay `if`**, FuncId minting past a closure id, coupling bail.
- **End-to-end** (`entry_outline_transform_8595`): two differentials — a straight-line body and one with a top-level `if` between runs — each compiled outline-off vs on (max chunking, so `a`/`b`/`c` objects are defined in separate chunks and read later), run under every moving-collector arm, **byte-identical correct output**. Asserts the outline actually fired.
- `-D warnings` clean (perry-codegen + perry, all targets).

Manually confirmed: interleaving toy (`if` between runs) → `less\ntotal:60` correct under default / `scavenge=1`; exports/globalThis/env toys correct when outlined.

## Next toward cli.js

Relax the coupling bails, validated against the gap suite (teach the `hir.init` scans to follow chunk calls, or keep `process.env`-literal statements inline via interleaving), then size-gated default. That removes the #4880 emit floor and restores full optimization on the real entry.

## Summary by CodeRabbit

- **New Features**
  - Added an opt-in code-generation transform (`PERRY_OUTLINE_ENTRY=1`) that splits oversized module startup code into ordered chunks.
  - Safely handles eligible top-level declarations, preserves execution order, and avoids conflicts with generated functions.
  - Reports whether entry outlining was applied and how many chunks were created.

- **Bug Fixes**
  - Unsupported or unsafe module patterns are skipped without changing generated behavior.

- **Tests**
  - Added coverage across multiple garbage-collection configurations, confirming identical runtime output.


proggeramlug:
Merging as a validated pair, stacked on current `main`.

| check | result |
|---|---|
| `perry-hir --lib` | **324 passed**, 1 ignored |
| `perry-codegen --lib` | **1159 passed**, 1 ignored |
| all seven ratchets | 0 |

**A note on the first validation run of #8604: it hit `No space left on device` and I discarded the result rather than reporting it.** ENOSPC mid-build produces a half-written artifact and then compile errors attributed to whatever crate happened to be linking — here `perry-hir` — which look exactly like real defects. Reclaimed disk and re-ran clean. Worth stating because "could not compile perry-hir" would have been a plausible-sounding and completely false finding.

**#8604's shape is the right one for a transform this invasive.** It rewrites the module entry — chunking `hir.init` into contiguous top-level ranges and hoisting each top-level `let` so the existing `emit_module_globals` escape rule globalizes exactly the cross-chunk ones. That is a lot of surface. What makes it safe to take is that it is **off by default** (`PERRY_OUTLINE_ENTRY=1`) and **declines rather than guesses**: unless every top-level statement is a plain expression or single-binding `let`, with no exports, `globalThis`, or top-level await, it leaves the body untouched. `var`, destructuring, top-level control flow and `process.env` couplings all bail. So the default path is byte-identical and it cannot miscompile a body it does not fully understand.

The reasoning for why this is a *runtime* win and not only a compile-time one is also worth recording: many LLVM passes have size cutoffs and degrade on giant functions, so a ~68 MB single-function module entry is under-optimized even when it compiles. Splitting it lets those passes run to full effect.

**Both changelog fragments were issue-keyed and are now PR-keyed**: `8595-` → `8604-`, `5716-` → `8605-`.

PR #8605, opened 2026-08-22
feat(native): enforce bundled binding governance

### LINKED ISSUE #5716: Define perry-ext governance and migrate non-core npm shims away from bundled Rust rewrites
## Problem

Perry currently uses many `perry-ext-*` crates as bundled well-known native bindings for npm package names. The mechanism is useful for true native boundaries, but it has started to blur the product direction:

> Perry should compile real TypeScript/JavaScript packages whenever possible. It should not drift into reimplementing npm package-by-package in Rust.

This is related to, but different from, #5422. #5422 covers build weight and default/release build taxonomy. This issue covers the compatibility strategy behind that weight: which `perry-ext-*` crates should actually exist in-tree, and which should be removed, externalized, or replaced by compiling the real npm package.

Current facts from the repo:

- `cargo metadata` sees 76 workspace members.
- There are 37 `perry-ext-*` crates.
- `default-members` currently includes many extension/staticlib crates.
- `docs/src/packages/porting.md` already says pure TS/JS packages should go through `perry.compilePackages`, while `.node` / `node-gyp` / prebuild packages need a native boundary.
- `crates/perry/well_known_bindings.toml` currently maps many ordinary npm package names to in-tree Rust crates.

The concern: if every useful npm package becomes `perry-ext-<name>`, Perry will not become a Node/TypeScript ecosystem compiler; it will become a growing collection of Rust rewrites of npm packages.

## Desired policy

Adopt this rule for bundled native bindings:

1. Prefer compiling the real JS/TS package source.
2. Prefer improving shared Node/Web runtime APIs over package-specific rewrites.
3. Use `perry.nativeLibrary` / `perry-ext-*` only for:
   - true native addon or system-library boundaries;
   - runtime-adjacent Node/Web APIs that many packages share;
   - very small strategic out-of-the-box shims, with an explicit maintenance/build-cost decision.
4. Heavy or domain-specific integrations should be official external packages, not part of Perry core.

## Not removal candidates in this pass

These crates are runtime-adjacent. They may still need build-policy cleanup or eventual stdlib consolidation, but they are not the main “reimplement npm package-by-package” problem:

| Crate | Current mapping | Reason to keep in/near core for now |
|---|---|---|
| `perry-ext-events` | `events` | Node event semantics are foundational and used by many packages. If changed, the likely move is into stdlib/runtime, not external package deletion. |
| `perry-ext-fetch` | `node-fetch`, `fetch` | Fetch/Headers/Request/Response are shared Web/Node runtime APIs. The package alias may be questionable, but the capability is core. |
| `perry-ext-http` / `perry-ext-http-server` | `http`, `https`, `http2` | Node HTTP server/client compatibility is core ecosystem infrastructure. |
| `perry-ext-net` | `net` | TCP sockets are foundational for DB drivers, HTTP stacks, Redis, etc. |
| `perry-ext-streams` | `streams` | Stream semantics are foundational for Node compatibility. |
| `perry-ext-ws` | `ws` | WebSocket support is common runtime infrastructure; reassess later whether `ws` package compatibility should be source-compiled instead. |
| `perry-ext-zlib` | `zlib` | Compression is a Node core API surface. |

## Candidate removals / externalizations, case by case

“Remove” here does **not** mean abruptly break users. It means stop treating the crate as a permanent in-tree well-known Rust rewrite. Each candidate should get a migration path: compile the real package, move to stdlib/runtime, or publish an official external `perry.nativeLibrary` package.

### A. Pure JS/TS packages that should be compiled from source

These are the strongest candidates to remove from the in-tree `perry-ext-*` strategy. Keeping them as Rust rewrites rewards the wrong compatibility path.

| Crate | Mapped package(s) | Why it should not be a bundled Rust ext | Suggested replacement |
|---|---|---|---|
| `perry-ext-dotenv` | `dotenv` | `dotenv` is a small JS package. Rewriting it in Rust proves little about npm compatibility. | Compile real `dotenv`; improve package resolution/fs/env handling if needed. |
| `perry-ext-nanoid` | `nanoid` | Small JS utility. Rust rewrite bypasses package compatibility work. | Compile real `nanoid`; fix crypto/random or ESM issues uncovered. |
| `perry-ext-uuid` | `uuid` | Common JS package. Perry should support it through package source plus Web/Node crypto. | Compile real `uuid`; improve `crypto.randomUUID`/random bytes support. |
| `perry-ext-slugify` | `slugify` | Small string utility. No native boundary. | Compile real package; improve string/unicode behavior if needed. |
| `perry-ext-exponential-backoff` | `exponential-backoff` | Tiny control-flow utility. No reason for a native binding. | Compile real package; improve timers/promises if gaps appear. |
| `perry-ext-lru-cache` | `lru-cache` | Pure JS data structure. Native rewrite avoids exercising class/Map semantics. | Compile real package; fix JS class/Map/runtime gaps. |
| `perry-ext-commander` | `commander` | CLI parser should be a package-compile target; Rust rewrite hides `process.argv`/CJS/ESM issues. | Compile real `commander`; improve process/argv/package interop. |
| `perry-ext-cron` | `cron`, `node-cron` | Scheduler packages are JS logic plus timers/date parsing. | Compile real packages; improve timers/date/runtime behavior. |
| `perry-ext-dayjs` | `dayjs`, `date-fns` | Date libraries are canonical pure-JS ecosystem packages. | Compile real packages; improve Date/Intl/module compatibility. |
| `perry-ext-moment` | `moment` | Same as dayjs/date-fns; keeping a Rust clone does not prove npm compatibility. | Compile real `moment`; fix Date/locale gaps as needed. |
| `perry-ext-decimal` | `decimal.js`, `bignumber.js` | These packages are JS libraries with well-defined semantics. | Compile real packages; improve numeric/object semantics. |
| `perry-ext-cheerio` | `cheerio` | Cheerio is a JS package stack; a Rust `scraper` wrapper is a different implementation and may diverge. | Compile real Cheerio or document unsupported gaps; improve package/runtime compatibility. |
| `perry-ext-validator` | `validator` | JS validation library. Rust rewrite avoids regex/string/package compatibility work. | Compile real `validator`; fix regex/string gaps. |
| `perry-ext-jsonwebtoken` | `jsonwebtoken` | Mostly package logic over crypto primitives. The shared crypto APIs should be the investment target. | Compile real package once crypto/key APIs are sufficient. |
| `perry-ext-axios` | `axios` | Axios is an HTTP client package. Perry should make it work through `fetch`/HTTP/URL/streams, not special-case it. | Compile real Axios; improve Web/Node HTTP compatibility. |
| `perry-ext-ethers` | `ethers` | Large JS ecosystem library. Rewriting a subset in Rust hides the real gaps in crypto, bytes, BigInt, package resolution, etc. | Compile real `ethers` or maintain as external package only if a native boundary is justified. |
| `perry-ext-ratelimit` | `rate-limiter-flexible` | Package-level application logic. No core runtime/native reason to live in Perry. | Compile real package; fix Map/timers/Redis integration gaps separately. |

### B. Domain-specific integrations that should be external packages

These may require native/system implementations, but that is an argument for `perry.nativeLibrary` packages outside the core repo, not for keeping every integration in Perry core.

| Crate | Mapped package(s) | Why it should leave core | Suggested replacement |
|---|---|---|---|
| `perry-ext-bcrypt` | `bcrypt` | Password hashing is useful but not Node core. It is a package-level native/crypto integration. | Official external `@perryts/bcrypt` or compile a pure JS/WASM-compatible package where viable. |
| `perry-ext-argon2` | `argon2` | Same as bcrypt: important but domain-specific and not core compiler/runtime. | Official external `@perryts/argon2`. |
| `perry-ext-better-sqlite3` | `better-sqlite3` | Native addon style package. Correct boundary is a native library package, not Perry core. | External `perry.nativeLibrary` package, or point users to a supported DB package. |
| `perry-ext-ioredis` | `ioredis`, `redis` | DB/client library. Should be covered by `net`/TLS runtime plus real JS package or official driver package. | Prefer pure TS/JS driver via compilePackages; otherwise official external package. |
| `perry-ext-pg` | `pg` | Database client should not be a bundled compiler-core integration. | Compile real `pg` once `net`/TLS/buffer are sufficient, or maintain `@perryts/postgres` externally. |
| `perry-ext-mysql2` | `mysql2`, `mysql2/promise` | Same database-driver concern; package compatibility should ride on shared `net`/TLS/buffer APIs. | Compile real package or maintain external `@perryts/mysql`. |
| `perry-ext-mongodb` | `mongodb` | Large database driver; not core runtime. | Compile real driver where possible or maintain external `@perryts/mongodb`. |
| `perry-ext-nodemailer` | `nodemailer` | Email transport stack is application/domain infrastructure, not Perry core. | Compile real package over `net`/TLS/streams, or external official package. |
| `perry-ext-sharp` | `sharp` | Native addon / image-processing boundary. Valid native binding idea, but too domain-specific for core. | External `@perryts/sharp` / image package with explicit native dependency policy. |
| `perry-ext-pdf` | `@perryts/pdf` | Official product package, not Node compatibility core. Keeping it in core couples release cadence unnecessarily. | Move to official external `@perryts/pdf` using `perry.nativeLibrary`. |
| `perry-ext-ads` | `perry/ads` | Mobile ads are a product/platform integration, not compiler/runtime core. | Move to official external package, likely platform-specific. |

### C. Framework/package adapters to reassess

| Crate | Mapped package(s) | Concern | Suggested replacement |
|---|---|---|---|
| `perry-ext-fastify` | `fastify` | Fastify is a JS framework. If Perry has solid HTTP, streams, URL, timers, and package interop, the real package should be the goal. | Compile real Fastify or support a tiny external adapter package, not a bundled Rust rewrite. |

### Phase 1: Policy and documentation

- [ ] Document the native-binding policy in `docs/src/native-libraries/overview.md`.
- [ ] Update `docs/src/packages/porting.md` to state that `perry-ext-*` is not the default npm compatibility path.
- [ ] Add a table listing every current `perry-ext-*` crate with: package mapping, category, migration target, and current status.

### Phase 2: Build-surface cleanup

- [ ] Remove non-core `perry-ext-*` crates from root `default-members`.
- [ ] Keep release workflows explicit: bundled extensions must be built by a release list/script, not implicitly through the default workspace.
- [ ] Align this with #5422's dev/release profile work.

### Phase 3: Compatibility radars

- [ ] Add a “pure npm package” radar that must not use `perry-ext-*`.
- [ ] Expand `scripts/node_core_subset.py` for shared Node APIs that packages depend on.
- [ ] Add a native-addon detector for `.node`, `binding.gyp`, `prebuilds`, `node-gyp`, or `gypfile` so packages are classified correctly.

### Phase 4: One-by-one migrations

For each candidate above:

- [ ] Try compiling the real npm package via `compilePackages`.
- [ ] Record the first blocking Perry gap.
- [ ] If the blocker is a shared runtime/API issue, fix that instead of preserving a package-specific ext.
- [ ] If the package is truly native/domain-specific, move it to an official external package with `perry.nativeLibrary`.
- [ ] Remove the well-known binding only after a documented compatibility path exists.

## Acceptance criteria

- [ ] Perry has a documented policy for when an in-tree `perry-ext-*` is allowed.
- [ ] Every existing `perry-ext-*` has a recorded classification and migration decision.
- [ ] Pure JS/TS packages are tracked as `compilePackages` compatibility work, not Rust rewrite work.
- [ ] Heavy/domain-specific native integrations have an externalization plan.
- [ ] `default-members` no longer pulls non-core extension/staticlib crates into the default developer loop.
- [ ] Release workflows still explicitly build and validate any extensions Perry intentionally ships.

## Non-goals

- Do not break existing users abruptly.
- Do not remove runtime-adjacent Node/Web compatibility work.
- Do not claim full Node compatibility requires zero native bindings; native addon boundaries still need a story.
- Do not duplicate #5422's build-profile/staticlib work; this issue should supply the extension-governance rationale that complements it.

proggeramlug:
## Summary

Define Perry's bundled-native-binding policy as an enforceable migration contract. The PR preserves every existing compatibility shim while recording whether each extension belongs near core, should yield to compiled upstream source, should move to an external native package, or should be removed after review.

## Changes

- Extend the existing `workspace-architecture.json` policy with migration targets for all 40 `perry-ext-*` crates: 6 core/runtime, 27 upstream-source migrations, 6 external-package migrations, and 1 reviewed removal.
- Add a generated documentation inventory with admission rules and compatibility-safe migration gates.
- Add a CI check that keeps architecture decisions, workspace members, crate directories, package mappings, and the published table in lock-step.
- Make Unix release builds enumerate the explicitly governed extension set instead of implicitly globbing every matching directory.
- Update the native-binding, package-porting, stdlib, and contributor docs to prefer real JS/TS package compilation and shared runtime APIs.
- Keep all current bindings bundled; this PR removes no mapping or compatibility path.

## Related issue

Fixes #5716

## Test plan

- [ ] `cargo build --release` clean (not run; no production Rust changed)
- [ ] Full workspace test scope passes (not run)
- [x] `python3 scripts/binding_governance.py --check`
- [x] `python3 scripts/workspace_architecture.py --self-test`
- [x] `python3 scripts/workspace_architecture.py --check --print-summary`
- [x] `node scripts/binding_pins.mjs --check`
- [x] `cargo test -p perry release_ext_builds_share_the_shipped_runtime_feature_set`
- [x] `cargo run -p perry-doc-tests -- --lint docs/src`
- [x] `./docs/i18n.sh build en`
- [x] `shellcheck scripts/release_ext_packages.sh scripts/build_linux_glibc_2_31.sh`
- [x] Added an affected CI policy check and updated user-facing docs

## Checklist

- [x] My commit follows the repository's Conventional Commit style

## Summary by CodeRabbit

- **Documentation**
  - Added guidance explaining when native bindings are used versus compiling packages from source.
  - Documented supported migration paths, compatibility requirements, and policies for native integrations.
  - Added a dedicated native-binding governance section and navigation entry.
  - Clarified package porting, standard-library behavior, resolver selection, and compatibility-shim usage.
  - Added an inventory of bundled bindings and their current migration status.

- **Build Improvements**
  - Release builds now use a validated inventory of approved native extensions, improving consistency and preventing unintended packages from being included.

### DIFF
--- a/.github/workflows/release-packages.yml
+++ b/.github/workflows/release-packages.yml
@@ -523,20 +523,21 @@ jobs:
         # the link deduper then cannot remove the whole second copy, and the
         # HTTP accept loop reads a different async reactor from the one the JS
         # event loop drives.
-        # Best-effort per crate: a wrapper that can't build on this host must
-        # not fail the whole release — the packaging glob below ships
-        # whatever was produced.
+        # #5716: the governance inventory is the explicit shipping set; adding
+        # a crate requires a recorded retain/migration decision. Best-effort
+        # per crate: a wrapper that can't build on this host must not fail the
+        # whole release — packaging below ships whatever was produced.
         if: runner.os != 'Windows' && matrix.old_glibc_image == ''
         run: |
-          for d in crates/perry-ext-*; do
-            [ -d "$d" ] || continue
-            name=$(basename "$d")
+          set -euo pipefail
+          governed_ext_packages=$(./scripts/release_ext_packages.sh)
+          while IFS= read -r name; do
             echo "::group::build $name"
             cargo build --profile dist --target ${{ matrix.target }} \
               -p perry -p perry-runtime-static -p perry-stdlib-static -p "$name" \
               || echo "  (skipped $name — failed to build on this host)"
             echo "::endgroup::"
-          done
+          done <<< "$governed_ext_packages"
 
       # #6351: only GTK4 needs noble. Build the compiler, runtime, stdlib and
       # extension archives in a Debian 11 container so their ABI floor is
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -311,6 +311,13 @@ jobs:
         if: ${{ !cancelled() }}
         run: node scripts/binding_pins.mjs --check
 
+      # #5716: every bundled extension needs an explicit product disposition;
+      # mappings, workspace members, crate directories, release enumeration,
+      # and the published inventory must stay in lock-step.
+      - name: Bundled binding governance
+        if: ${{ !cancelled() }}
+        run: python3 scripts/binding_governance.py --check
+
       # GC write-barrier store-site inventory: every raw heap-slot store in
       # perry-codegen / perry-runtime / perry-stdlib must be barriered or
       # carry a justified GC_STORE_AUDIT(...) marker (or a justified entry
--- a/changelog.d/5716-native-binding-governance.md
+++ b/changelog.d/5716-native-binding-governance.md
@@ -1,0 +1,6 @@
+Defined an enforceable policy for bundled native bindings: ordinary
+JavaScript and TypeScript packages now have recorded upstream-source migration
+targets, native/domain integrations have external-package targets, and shared
+runtime APIs have explicit retain/consolidate decisions. Release builds consume
+the governed inventory, while existing compatibility shims remain bundled until
+their documented migration gates pass.
--- a/crates/perry/well_known_bindings.toml
+++ b/crates/perry/well_known_bindings.toml
@@ -29,6 +29,12 @@
 # Migration order tracks #466 Phase 5 (smallest first). Each entry
 # is gated on the bundled crate actually existing in the workspace
 # — wrappers that haven't been ported yet have no entry.
+#
+# This file describes current resolver behavior, not permission to keep a
+# package-specific Rust rewrite indefinitely. Before adding or removing a
+# bundled crate, update its `migration` in `workspace-architecture.json`; CI
+# requires every ext crate and package mapping to have an explicit decision
+# (#5716).
 
 [bindings.dotenv]
 crate = "perry-ext-dotenv"
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -23,6 +23,7 @@
 
 - [Porting Packages (experimental)](packages/porting.md)
 - [Native Bindings — Overview](native-libraries/overview.md)
+  - [Bundled Binding Governance](native-libraries/governance.md)
   - [Authoring a Native Binding](native-libraries/authoring-guide.md)
   - [`perry-ffi` ABI Reference](native-libraries/abi.md)
   - [Manifest Schema (spec v1)](native-libraries/manifest-v1.md)
--- a/docs/src/contributing/crate-policy.md
+++ b/docs/src/contributing/crate-policy.md
@@ -42,6 +42,10 @@ Every member is classified in `workspace-architecture.json`:
 Its `decision` records the reviewed direction: `keep`, `merge`, `externalize`,
 `remove`, or `review`. Decisions describe the architecture roadmap; they do not
 authorize deleting compatibility without the relevant migration and tests.
+Binding entries also carry a `migration` that refines the general decision:
+`core-runtime`, `compile-source`, `external-package`, or `remove`. The native
+[binding governance policy](../native-libraries/governance.md) defines those
+targets and their compatibility gates.
 
 Choose the decision by applying the same seam test consistently:
 
--- a/docs/src/native-libraries/governance.md
+++ b/docs/src/native-libraries/governance.md
@@ -1,0 +1,113 @@
+# Bundled native-binding governance
+
+Perry's compatibility goal is to compile real TypeScript and JavaScript
+packages. A package-specific Rust rewrite can be useful as a temporary bridge,
+but it is not the default compatibility strategy and does not demonstrate that
+Perry can compile the package it replaces.
+
+## Policy
+
+Use this order when adding package compatibility:
+
+1. Compile the upstream JavaScript or TypeScript package source with
+   `perry.compilePackages`.
+2. When that fails on a reusable Node.js or Web API, improve the shared
+   runtime API rather than rewriting the package.
+3. Use `perry.nativeLibrary` only for a true native-addon or system-library
+   boundary. Domain-specific bindings should ship as independently versioned
+   external packages.
+4. Keep an in-tree `perry-ext-*` crate only for a shared runtime API or a small
+   strategic shim whose maintenance and release cost has been accepted
+   explicitly.
+
+A proposal for a new bundled extension must identify the native boundary or
+shared runtime capability, explain why compiling the package source is not
+viable, name an owner, account for build and binary cost, define its
+faithfulness level, and include an exit or consolidation plan. An ordinary npm
+package is not eligible merely because a Rust implementation is convenient.
+
+## Categories and migration gates
+
+- **Runtime API** bindings implement foundational Node.js or Web capabilities
+  used by many packages. They stay in or near core for now and may eventually
+  be consolidated into `perry-stdlib` or the runtime. A third-party package
+  alias can still be reassessed separately.
+- **Source package** bindings replace packages whose upstream implementation
+  can in principle be compiled. Their migration target is the real upstream
+  source plus fixes to shared language, module, and runtime support.
+- **External integration** bindings cross a real native boundary or provide a
+  product/domain integration. Their migration target is an official or
+  third-party package using `perry.nativeLibrary`, with its own release cycle.
+- **Obsolete integration** bindings have no supported long-term owner. Remove
+  them only after checking current consumers and documenting the compatibility
+  impact.
+
+`perry-ext-fetch` is a mixed case: compiling `node-fetch` upstream does not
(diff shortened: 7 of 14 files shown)


proggeramlug:
Merging as a validated pair, stacked on current `main`.

| check | result |
|---|---|
| `perry-hir --lib` | **324 passed**, 1 ignored |
| `perry-codegen --lib` | **1159 passed**, 1 ignored |
| all seven ratchets | 0 |

**A note on the first validation run of #8604: it hit `No space left on device` and I discarded the result rather than reporting it.** ENOSPC mid-build produces a half-written artifact and then compile errors attributed to whatever crate happened to be linking — here `perry-hir` — which look exactly like real defects. Reclaimed disk and re-ran clean. Worth stating because "could not compile perry-hir" would have been a plausible-sounding and completely false finding.

**#8604's shape is the right one for a transform this invasive.** It rewrites the module entry — chunking `hir.init` into contiguous top-level ranges and hoisting each top-level `let` so the existing `emit_module_globals` escape rule globalizes exactly the cross-chunk ones. That is a lot of surface. What makes it safe to take is that it is **off by default** (`PERRY_OUTLINE_ENTRY=1`) and **declines rather than guesses**: unless every top-level statement is a plain expression or single-binding `let`, with no exports, `globalThis`, or top-level await, it leaves the body untouched. `var`, destructuring, top-level control flow and `process.env` couplings all bail. So the default path is byte-identical and it cannot miscompile a body it does not fully understand.

The reasoning for why this is a *runtime* win and not only a compile-time one is also worth recording: many LLVM passes have size cutoffs and degrade on giant functions, so a ~68 MB single-function module entry is under-optimized even when it compiles. Splitting it lets those passes run to full effect.

**Both changelog fragments were issue-keyed and are now PR-keyed**: `8595-` → `8604-`, `5716-` → `8605-`.

Issue #4975, opened 2026-06-11
node:http/https behavioral parity tail (26.2% → ...) — successor to #2132


proggeramlug:
## node:http / node:https behavioral parity tail

Successor to #2132 (which covered the **byte-level diff** scope — now closed). This tracks the **behavioral** gap that re-measurement against Node's own `test/parallel` corpus surfaced: tests that compile and run but crash, hang, or fail an assertion.

### Measurement (Node v22 corpus, `scripts/node_core_subset.py --api http https`)

| | original (#2132 open) | after #2132 fixes |
|---|---|---|
| **parity** | 11.6% | **26.2%** |
| pass | 38 | 86 |
| runtime-fail | 285 | 240 |
| compile-fail | 4 | 0 |
| diff | 2 | 2 |

The #2132 fixes (#4903/#4904/#4905/#4906/#4907/#4908/#4909/#4910) cleared the clean single-root-cause clusters (server.address, constructors, missing methods, validation, codegen). What remains is the hard tail.

### Remaining 240 runtime-fails by cluster

| cluster | count | notes |
|---|---|---|
| no-output / server hang | 67 | framing / keep-alive / timeout hangs; improved by #4964 but still the biggest mechanical bucket |
| strict-equal mismatch | 63 | **behavioral semantics** — largest, grows as upstream fixes unblock tests |
| other one-offs | 56 | mostly hang-with-partial-stdout + the distinct bugs below |
| read-undefined-prop | 25 | scattered |
| not-a-function | 13 | residual lifecycle methods |
| https TLS cert | 6 | residual from #4906 |
| missing-exception | 5 | residual from #4907 |
| ReferenceError not-defined | 5 | missing globals (`MessageEvent`, `OutgoingMessage`, …) |

### Sub-tickets (distinct root causes split out)
- [ ] #4970 — **regression**: empty/falsy `ClientRequest` method `''` throws instead of defaulting to `GET` (over-validation from #4907)
- [ ] #4971 — `js_native_call_method` null-ptr crash in `closeIdleConnections` sweep
- [ ] #4972 — codegen: derived class over native base, `super()` `this`-TDZ
- [ ] #4973 — http upgrade: `Object prototype may only be an Object or null`
- [ ] #4974 — https `tls config unavailable` under async-dispose

### Not yet split (need triage)
[...]

PR #8608, opened 2026-08-22
fix(http): expose Agent keep-alive timeout buffer

proggeramlug:
## Summary

- retain Node-compatible `agentKeepAliveTimeoutBuffer` constructor values and default invalid or missing values to 1000 ms
- expose the ordinary writable Agent property through Perry's external-handle get/set dispatch, including GC rooting for assigned JS values
- add a parity regression mirroring Node's `test-http-agent-keep-alive-timeout-buffer.js` assertions

## Tests

- `cargo test --release -p perry-ext-http agent::tests -- --nocapture` (8 passed)
- `git diff --check`
- Node output matches the new parity fixture exactly

Refs #4975

No version bump.

## Summary by CodeRabbit

- **New Features**
  - Added support for the `agentKeepAliveTimeoutBuffer` option on HTTP and HTTPS agents.
  - Valid finite, non-negative values are preserved; invalid or missing values default to 1000 ms.
  - The setting can be read and updated after an agent is created.

- **Tests**
  - Added coverage for valid, invalid, infinite, and omitted timeout buffer values.


proggeramlug:
Merging as a validated pair on current `main`.

| check | result |
|---|---|
| all seven ratchets | 0 |

These were staged alongside #8610, which is held: it adds +2 bare raw-handle reads in `class_registry/construct.rs` and +1 open-coded `StringHeader` payload offset. I confirmed by exclusion that the debt is entirely #8610's — with just these two the same gates report 925/925 and 436 inline offsets, both exit 0 — so neither of these is implicated and there was no reason to hold them behind it.

PR #8609, opened 2026-08-22
fix(ui-ios): pause UI mutations during touch scrolling

### LINKED ISSUE #7763: iOS: scrolling a UIStackView in a UIScrollView while mutating labels aborts with 'insertObject:atIndex:: object cannot be nil'
## Summary

Scrolling a `UIScrollView`-hosted `UIStackView` on a physical device while JS mutates its labels aborts the app:

```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException',
    reason: '*** -[__NSArrayM insertObject:atIndex:]: object cannot be nil'
App terminated due to signal 6.
```

**Requires human touch input.** No automated probe reproduces it, which is why it has gone unseen — every headless run passes.

## Where it throws

The backtrace is `start → main → UIApplicationMain → CFRunLoop → CA transaction commit → UIView layout → UIStackView → throw`. The exception lands in the **layout pass at the end of a run-loop iteration**, not inside any Perry callback. Something earlier leaves a nil in a UIKit array; it detonates at commit time.

## Reproduction

Device: iPhone 16 Pro, iOS 26, Release build, `--target ios`.

```
PERRY_NO_AUTO_OPTIMIZE=1 perry run ios benchmarks/ios-ui/bench.ts --device <UDID>
```

Scroll the list continuously during the `property-updates` phase (50 labels, `textSetString` on each every frame) or `add-remove` (100 rows rebuilt 20x/sec). Crash lands within a few seconds. Without scrolling, the same build runs indefinitely.

`PERRY_NO_AUTO_OPTIMIZE=1` is needed because `perry-stdlib` currently fails to build with `--no-default-features` — separate issue, see below.

## Not caused by the frame-metrics work (#7754)

Established by four single-variable device bisects, each with a human scrolling:

| arm | result |
|---|---|
| deferred view release | crashes |
| widget release disabled entirely | crashes |
| display link in `NSDefaultRunLoopMode` | crashes |
| **display-link driver disabled, original NSTimer pump** | **crashes** |

The last arm is decisive: with #7754's driver removed and `onFrame` back on the pre-PR `js_frame_pump_default()` path, the abort still occurs. That PR's benchmark is simply the first workload that drives enough label mutation against a scrolled stack to expose it.

## Hypotheses tested and falsified

- Deallocating views inside CoreAnimation's pre-commit phase — falsified; crashes with release deferred to a quiescent point.
- Releasing widget-table entries at all — falsified; crashes with release fully disabled. An isolation probe released ~323k widgets inline with no fault.
- Running JS during `UITrackingRunLoopMode` — falsified; crashes with the link in default mode, where JS is frozen during the drag.

## Suggested next step

Symbolicate a full crash report (the raw frames above have no load address), and audit every path that can hand a view to a `UIStackView` arranged-subview array. `add_child` / `add_child_at` guard their handle lookups, so the nil likely enters via a UIKit-held reference to a view that went away, or via constraint bookkeeping in `HEIGHT_CONSTRAINTS`.

proggeramlug:
## Summary

Prevent UIKit from laying out a scrolling `UIStackView` while Perry is concurrently applying JavaScript-driven label or arranged-subview mutations. The iOS runtime pump now stays quiescent for the full touch drag/deceleration and two pump turns after scrolling settles.

## Changes

- Install a lifecycle-managed `UIScrollView` delegate when each Perry scroll view is created, while preserving the existing `onScrollEnd` behavior.
- Track simultaneous touch-scroll interactions and defer timer, frame, microtask, stdlib, GC, geisterhand, and root-health pump work until UIKit has committed its final scroll layout transaction.
- Add host-runnable state-machine tests for single and overlapping scroll views.
- Add a bounded physical-device repro that combines 50 label updates per frame with periodic 100-row stack rebuilds.
- Replace the removed delegate-map root-holder inventory entry with the new numeric-only mutation gate entry.

## Related issue

Fixes #7763

## Test plan

- [x] `cargo test -p perry-ui-ios --test scroll_mutation_gate` (2 passed)
- [x] `cargo build --release -p perry-runtime-static -p perry-stdlib-static -p perry-ui-ios --target aarch64-apple-ios`
- [x] `BASE_SHA=upstream/main SKIP_COMPILE_GATES=1 ./scripts/run_lint_gates.sh` (all 53 script gates passed; compile tier covered by the focused iOS build above)
- [x] Built, signed, installed, and launched the release repro on a physical iPhone 16 Pro.
- [ ] Full host workspace test suite (not run; change is isolated to the iOS-only backend)

## Screenshots / output

Unpatched physical-device control aborted immediately after frame 600 with:

```text
NSInvalidArgumentException: -[__NSArrayM insertObject:atIndex:]: object cannot be nil
App terminated due to signal 6.
```

The patched device run passed frames 600 and 1200 without that exception, and the device produced no new `issue_7763_touch_scroll` crash report. A final install/launch smoke also succeeded after switching delegate ownership from a leak to an Objective-C association.

I also built an XCUITest rapid-swipe harness, but iOS canceled UI-test initialization at the device-local authentication prompt (`LocalAuthentication Code=-2`) before it sent any swipes, so that is not counted as a passing automated test.

## Checklist

- [x] I have NOT bumped the workspace version or edited CLAUDE.md / CHANGELOG.md.
- [x] My commit follows the loose `feat:` / `fix:` / `docs:` / `chore:` prefix convention.
- [x] I have read `CONTRIBUTING.md` and agree to the Code of Conduct.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved iOS scroll performance and stability during dragging and deceleration.
  - Prevented UI updates from interrupting scrolling or causing layout failures.
  - Preserved scroll-end callbacks and improved callback validation.

- **Tests**
  - Added coverage for single and overlapping scroll interactions, including post-scroll settling periods.

- **Documentation**
  - Added a changelog entry describing the iOS scrolling fix.

### DIFF
--- a/benchmarks/ios-ui/issue_7763_touch_scroll.ts
+++ b/benchmarks/ios-ui/issue_7763_touch_scroll.ts
@@ -1,0 +1,65 @@
+// Reproduction for #7763: mutate a UIScrollView-hosted UIStackView while
+// real touch-driven scrolling is active. Programmatic setContentOffset does
+// not exercise the UIKit gesture/tracking path that triggers the crash.
+//
+//     PERRY_NO_AUTO_OPTIMIZE=1 perry run ios \
+//       benchmarks/ios-ui/issue_7763_touch_scroll.ts --device <UDID>
+
+import {
+    App,
+    VStack,
+    Text,
+    ScrollView,
+    scrollviewSetChild,
+    widgetAddChild,
+    widgetClearChildren,
+    textSetString,
+    onFrame,
+} from "perry/ui"
+
+const ROWS = 100
+const LIVE_LABELS = 50
+// Bound the structural churn so this isolates the UIKit crash instead of the
+// separate retained-widget issue on branches that predate widget tombstones.
+const CHURN_UNTIL_FRAME = 600
+const content = VStack(4, [])
+const scroll = ScrollView()
+scrollviewSetChild(scroll, content)
+
+const labels: unknown[] = []
+let frame = 0
+
+function rebuild(): void {
+    widgetClearChildren(content)
+    labels.length = 0
+    for (let i = 0; i < ROWS; i++) {
+        const row = Text(`row ${i} @ ${frame}`)
+        labels.push(row)
+        widgetAddChild(content, row)
+    }
+}
+
+rebuild()
+
+function loop(): void {
+    frame++
+    for (let i = 0; i < LIVE_LABELS; i++) {
+        textSetString(labels[i] as never, `row ${i} @ ${frame}`)
+    }
+    if (frame <= CHURN_UNTIL_FRAME && frame % 6 === 0) {
+        rebuild()
+    }
+    if (frame % 600 === 0) {
+        console.log(`issue-7763 frames: ${frame}`)
+    }
+    onFrame(loop)
+}
+
+onFrame(loop)
+
+App({
+    title: "Issue 7763 Touch Scroll",
+    width: 400,
+    height: 800,
+    body: VStack(0, [scroll]),
+})
--- a/changelog.d/8609-ios-touch-scroll-mutation.md
+++ b/changelog.d/8609-ios-touch-scroll-mutation.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- iOS apps no longer abort inside UIKit layout when JavaScript mutates labels or arranged subviews while a `UIScrollView` is being dragged or decelerating. Perry now waits for touch scrolling and its final layout transaction to settle before resuming UI-producing runtime work.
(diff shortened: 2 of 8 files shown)


proggeramlug:
Merging as a validated pair on current `main`.

| check | result |
|---|---|
| all seven ratchets | 0 |

These were staged alongside #8610, which is held: it adds +2 bare raw-handle reads in `class_registry/construct.rs` and +1 open-coded `StringHeader` payload offset. I confirmed by exclusion that the debt is entirely #8610's — with just these two the same gates report 925/925 and 436 inline offsets, both exit 0 — so neither of these is implicated and there was no reason to hold them behind it.

Issue #8607, opened 2026-08-22
Opt target — `pipeline`: 1.12x Node, **1.74x instructions at ~1.05x cycles**


proggeramlug:
# Opt target — `pipeline`: 1.12x Node, **1.74x instructions at ~1.05x cycles**

#8592 closed, and `pipeline` improved 1.18x → 1.12x. But **the diagnosis that made it
interesting is unchanged**: it still retires ~74% more instructions than Node while spending
almost no extra cycles. The gap is still waiting, not computing.

## Measurement — quiet M1 mini, 5 shuffled interleaved repeats, verdict CLEAN, perry `adbebbe9d`

| | node | perry | ratio |
|---|---:|---:|---:|
| wall | 0.098 s | 0.109 s | **1.12x** |
| instructions | — | — | **1.74x** |
| peak RSS | — | — | 0.25x |

## The worked example is `asyncpipe`

That row had exactly this signature — 1.07x wall against 1.71x instructions — and #8570 flipped
it to **0.83x** by suppressing redundant same-thread promise-drain wakeups and removing
per-element closure/guard/chained-promise allocations from the intrinsic `Promise.all` path. Its
instruction ratio fell to 1.32x as a by-product.

`pipeline` is the last row still showing that signature. The same class of fix is the first
thing to try.

## What NOT to do

Do not reach for a sampler. Aggregating `sample` over 14 runs of `asyncpipe` (~88 ms) yielded
**one** usable self-time sample, because the process is parked rather than running. A profiler
shows a flat, uninformative profile and you conclude wrongly that there is nothing to fix.

Instrument instead: event-loop wakeups and how many are redundant with a drain already in
flight; microtask drains and jobs consumed per drain; timer registrations and due-time-to-run
latency; per-element allocations on intrinsic paths.
`PROFILE_NOTIFY_DRAIN_SUPPRESSED_COUNT` from #8570 is the working example of making this class
countable.

## Acceptance

**Wall time is the primary signal here**, unusually — the instruction count is known not to be
the constraint. Many interleaved repeats with reported spread; treat a wall delta as real only
when ranges are disjoint. 20-row corpus byte-exact, whole-corpus timing, no RSS trade.

PR #8613, opened 2026-08-22
perf(codegen): retain guarded array accumulators as i32

### LINKED ISSUE #8606: Opt target — `typed_array` round 2: 3.87x Node, 2.24x instructions
# Opt target — `typed_array` round 2: 3.87x Node, 2.24x instructions

**#8590 cut this row from 6.29x to 3.87x wall and from 8.52x to 2.24x instructions** — nearly a
4x reduction in executed work. It is still the worst row, but it is no longer in a category of
its own, and the shape of the remaining gap has changed enough to need a fresh look rather than
more of the same.

## Measurement — quiet M1 mini, 5 shuffled interleaved repeats, verdict CLEAN, perry `adbebbe9d`

| | node | perry | ratio | was (`fefdc367b`) |
|---|---:|---:|---:|---:|
| wall | 0.645 s | 2.499 s | **3.87x** | 6.29x |
| instructions | — | — | **2.24x** | 8.52x |
| peak RSS | — | — | 0.06x | 0.06x |

## Why this needs a new diagnosis, not a continuation

The wall ratio is now **1.7x the instruction ratio** (3.87 vs 2.24). Before #8590 they were much
closer (6.29 vs 8.52 — instructions actually *exceeded* wall). That inversion means the
remaining gap is no longer mostly "too many instructions": something is now costing wall time
per instruction. Candidates, in the order worth checking:

- memory stalls — this workload streams two Int32Arrays; check cache behaviour now that the
  instruction count is 4x lower
- branch misprediction on the remaining guards
- a serialising dependency chain in the Blowfish rounds that the earlier instruction bulk hid

**Measure IPC first.** If Perry's IPC has dropped relative to the previous sweep while
instructions fell, that localises it immediately and tells you this is a stall problem rather
than a work problem.

Workload: `gc-handoff/apps/typed_array.ts` — bcryptjs Blowfish `_encipher`, `S`/`P` Int32Arrays
through untyped params, ~4.27 M calls. Prints only `checksum:-821955270`, which must not change.

## Ground rules

Same as the earlier tickets: build with the `-static` wrappers or you A/B a stale runtime;
instructions retired *and* wall, since which one matters is now the open question; 20-row corpus
byte-exact; whole-corpus timing; do not trade RSS (already 0.06x Node).

proggeramlug:
## Summary

- propagate runtime-backed `number[]` parameter proofs into native-region fact collection
- retain guarded-array Feistel accumulators in canonical i32 slots after a conservative root-level normalization proof
- keep nested bitwise operands native across mutable numeric locals and bounds-proven typed-array reads
- preserve the generic fallback when the entry guard or dominance proof is absent

## typed_array result

Exact issue workload, three-run medians on this host:

| metric | issue baseline | this PR | change |
|---|---:|---:|---:|
| retired instructions | 14.68B | 9.66B | -34.2% |
| cycles | 8.10B | 5.96B | -26.5% |
| user CPU | 2.73s | 2.01s | -26.4% |

The specialized `encipherUntyped` function contains zero `js_number_coerce` calls after this change. Output remains `checksum:-821955270`.

## Validation

- `cargo test -p perry-codegen --lib` (1,163 passed, 1 ignored)
- `cargo test --release -p perry-codegen --lib` (1,163 passed, 1 ignored)
- `cargo clippy -p perry --bins`
- `cargo check -p perry`
- exact workload checksum, symbol inspection, and three-run hardware-counter comparison

No version bump.

Fixes #8606

## Summary by CodeRabbit

  * Improved numeric-array and typed-array processing for bitwise operations.
  * Reduced unnecessary numeric conversions in optimized execution paths.
  * Preserved integer representations more effectively for faster calculations.
  * Maintained existing results, including checksum consistency.
* **Bug Fixes**
  * Improved handling of nested bitwise operations, conditional paths, and out-of-bounds array reads.
  * Ensured numeric-array optimizations apply only when type and normalization guarantees are valid.
* **Tests**
  * Added coverage for guarded numeric arrays, normalization order, and unreachable execution paths.

### DIFF
--- a/changelog.d/8613-guarded-array-i32.md
+++ b/changelog.d/8613-guarded-array-i32.md
@@ -1,0 +1,14 @@
+### Performance
+
+- Propagate runtime-backed numeric-array parameter guards into native
+  representation selection, allowing Feistel-style accumulators seeded from a
+  guarded `number[]` to stay in canonical i32 slots after a dominating bitwise
+  normalization. Nested bitwise expressions now also consume mutable
+  Number-by-construction locals and bounds-proven typed-array reads without
+  falling back through `js_number_coerce`.
+
+- On the exact `typed_array` workload from #8606, retired instructions fall
+  from a 14.68B three-run median to 9.66B (-34.2%), cycles from 8.10B to 5.96B
+  (-26.5%), and user CPU from 2.73s to 2.01s (-26.4%), with the checksum
+  unchanged. The specialized `encipherUntyped` function now contains zero
+  `js_number_coerce` calls.
--- a/crates/perry-codegen/src/codegen/function.rs
+++ b/crates/perry-codegen/src/codegen/function.rs
@@ -835,6 +835,17 @@ pub(super) fn compile_function(
             .then_some(*id)
         })
         .collect();
+    let spec_number_array_params: HashSet<u32> = spec_param_proofs
+        .iter()
+        .filter_map(|(id, ty)| {
+            matches!(ty, perry_hir::types::Type::Array(element)
+                if matches!(element.as_ref(), perry_hir::types::Type::Number | perry_hir::types::Type::Int32))
+            .then_some(*id)
+        })
+        .collect();
+    // Unlike source annotations, these proofs are backed by the public
+    // wrapper's `js_param_type_guard`.  The guarded-array seed collector may
+    // therefore trust that an element is Number-or-undefined in this clone.
     // `--opt-report` (#6952): attribute every representation decision the
     // collectors below make to this function. No-op when the report is off.
     //
@@ -865,6 +876,7 @@ pub(super) fn compile_function(
         &spec_ta_lens,
         &spec_i32_params,
         &spec_numeric_params,
+        &spec_number_array_params,
     );
 
     // A Number-by-construction local cannot ever hold a GC pointer, so it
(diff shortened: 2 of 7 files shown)


proggeramlug:

| check | result |
|---|---|
| `perry-codegen --lib` | **1167 passed**, 1 ignored |
| all seven ratchets | 0 |

**#8610's ratchet debt is resolved.** It was held twice for +2 bare raw-handle reads in `class_registry/construct.rs` (7 against a ceiling of 5) and +1 open-coded `StringHeader` payload offset. On head `c9b994d0f` both gates now report clean — **925/925 raw-handle reads, 108 modules within ceilings, payload inventory exit 0**. Converted rather than baseline-refreshed, which was the right call given the counts had gone up. Thanks for taking that rather than reaching for `--write-baseline`; `construct.rs` runs during instance construction, so a raw handle held across an allocating call there is exactly the shape that yields a stale pointer surfacing cycles later somewhere unrelated.

**#8613 and #8614 are aimed at the two tickets filed after the last sweep** — #8606 (`typed_array` round 2, where the wall ratio now exceeds the instruction ratio) and #8607 (`pipeline`, still 1.74x instructions at ~1.05x cycles). Worth being explicit about what this validation does and does not establish: the gates and suites say these are **safe to land**, not that they are faster. Both are perf changes, and the corpus sweep after merging is the actual verdict. I will run it and report against those tickets.

Two metadata fixes applied while staging (fork PRs, so they could not be pushed to the branches): #8610 shipped without a `changelog.d` fragment, so I wrote one; #8614's was keyed `8607-` after its issue and is now PR-keyed.

PR #8614, opened 2026-08-22
perf(codegen): speed up generic registry pipelines

### LINKED ISSUE #8607: Opt target — `pipeline`: 1.12x Node, **1.74x instructions at ~1.05x cycles**

proggeramlug:
## Summary

Speeds up the pipeline benchmark by keeping stable array-valued fields in local slots for methods called on proven-contained receivers, and by adding a guarded numeric fast path for null-defaulted dynamic counters. Dynamic JavaScript behavior and aliased receiver routes retain their existing fallbacks.

## Changes

- Emit a separate internal $ptr_arrays method clone for eligible loop methods on Phase 3b contained receivers.
- Snapshot only array fields initialized by array literals, and decline the clone for slot replacement, deletion, or reentrant this/super method calls.
- Keep aliased, guarded, and dispatch-tower calls on the existing public/$pshape paths.
- Recognize the (value === null ? 0 : value) + numeric-literal shape and use an inline numeric add after a runtime tag check, with full dynamic add as the fallback.
- Add IR routing/safety tests and an end-to-end parity regression.

## Related issue

Closes #8607

## Test plan

- [x] cargo build --release -p perry
- [x] cargo test -p perry-codegen --lib --quiet (1,163 passed; 1 ignored)
- [x] PERRY_SKIP_BUILD=1 ./run_parity_tests.sh --filter test_issue_8607 (1/1 passed, byte-exact against Node)
- [x] python3 scripts/check_test_registration.py
- [x] cargo fmt --all -- --check
- [x] ./scripts/pre-tag-check.sh --quick
- [ ] Full workspace test command from the template. The affected-crate runner completed all 2,634 runtime tests (4 ignored), then stopped in an unchanged perry binary audit because current main does not classify the PERRY_OUTLINE_ENTRY environment variables in the build-cache allow/exclusion lists.

The private m0810 20-row acceptance corpus referenced by repository notes is not present in this checkout; the public pipeline workload and the added parity fixture are byte-exact.

## Benchmark

Apple M1 mini, quiet host. Wall spread is 20 shuffled/interleaved five-process batches, normalized per process; batching avoids the 10 ms resolution of macOS /usr/bin/time. Counters are 20 shuffled/interleaved single-process runs. Baseline is 970eab30c; final is this branch.

| runtime | wall median | wall range | instructions median | cycles median | peak RSS |
| baseline | 114.101 ms | 113.263-115.746 ms | 2,131,456,485 | 348,841,468 | 22,282,240 B |
| final | 92.976 ms | 92.229-95.320 ms | 1,784,508,872 | 283,436,554 | 22,282,240 B |
| Node | 101.476 ms | 99.407-104.134 ms | 1,231,885,256 | 355,256,028 | 90,112,000 B |

Final versus baseline: 18.5% lower wall time with disjoint ranges, 16.3% fewer instructions, 18.7% fewer cycles, and unchanged peak RSS. Final is 8.4% faster than Node by median wall time on this host.

## Screenshots / output

pipeline output from both Perry and Node:

    55626000 3 3

## Checklist

## Summary by CodeRabbit

  * Improved performance for methods that repeatedly access cached array fields on contained receivers.
  * Added optimized numeric addition for null-defaulted counters while preserving dynamic string-or-number behavior.

* **Bug Fixes**
  * Preserved correct behavior for aliased receivers, non-numeric fallback values, and property rebinding scenarios.

* **Tests**
  * Added coverage for cached array-field routing and null-defaulted numeric addition, including missing, repeated, and string-valued entries.

### DIFF
--- a/changelog.d/8607-pipeline-null-default-add.md
+++ b/changelog.d/8607-pipeline-null-default-add.md
@@ -1,0 +1,6 @@
+### Performance
+
+- Speed up generic registries and record-processing pipelines by caching stable
+  array fields on proven-contained receivers and using a guarded numeric fast
+  path for null-defaulted counters, while preserving dynamic JavaScript
+  semantics on aliased and non-number fallback paths.
--- a/crates/perry-codegen/src/codegen/artifacts.rs
+++ b/crates/perry-codegen/src/codegen/artifacts.rs
@@ -362,6 +362,7 @@ pub(super) fn emit_module_artifacts(c: ModuleArtifactsCtx<'_>) -> Result<()> {
                     false,
                     None,
                     Some(nonnegative_index_params),
+                    false,
                 )
                 .with_context(|| {
                     format!(
@@ -395,6 +396,7 @@ pub(super) fn emit_module_artifacts(c: ModuleArtifactsCtx<'_>) -> Result<()> {
                     .contains_key(&(class.name.clone(), method.name.clone())),
                 None,
                 None,
+                false,
             )
             .with_context(|| format!("lowering method '{}::{}'", class.name, method.name))?;
             // Representation-selection Phase 5a: the additive `internal`
@@ -431,6 +433,7 @@ pub(super) fn emit_module_artifacts(c: ModuleArtifactsCtx<'_>) -> Result<()> {
                     false,
                     Some(fact.clone()),
                     None,
+                    false,
                 )
                 .with_context(|| {
                     format!(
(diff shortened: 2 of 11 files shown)


proggeramlug:

| check | result |
|---|---|
| `perry-codegen --lib` | **1167 passed**, 1 ignored |
| all seven ratchets | 0 |

**#8610's ratchet debt is resolved.** It was held twice for +2 bare raw-handle reads in `class_registry/construct.rs` (7 against a ceiling of 5) and +1 open-coded `StringHeader` payload offset. On head `c9b994d0f` both gates now report clean — **925/925 raw-handle reads, 108 modules within ceilings, payload inventory exit 0**. Converted rather than baseline-refreshed, which was the right call given the counts had gone up. Thanks for taking that rather than reaching for `--write-baseline`; `construct.rs` runs during instance construction, so a raw handle held across an allocating call there is exactly the shape that yields a stale pointer surfacing cycles later somewhere unrelated.

**#8613 and #8614 are aimed at the two tickets filed after the last sweep** — #8606 (`typed_array` round 2, where the wall ratio now exceeds the instruction ratio) and #8607 (`pipeline`, still 1.74x instructions at ~1.05x cycles). Worth being explicit about what this validation does and does not establish: the gates and suites say these are **safe to land**, not that they are faster. Both are perf changes, and the corpus sweep after merging is the actual verdict. I will run it and report against those tickets.

Two metadata fixes applied while staging (fork PRs, so they could not be pushed to the branches): #8610 shipped without a `changelog.d` fragment, so I wrote one; #8614's was keyed `8607-` after its issue and is now PR-keyed.

PR #8615, opened 2026-08-22
fix(http): advance Agent request lifecycle parity

proggeramlug:
## Summary

- implement per-origin Agent admission accounting, `maxSockets` FIFO queuing, queued-abort cleanup, observable `sockets` / `requests` / `freeSockets` counts, `totalSocketCount`, and `reusedSocket`
- pass Node's keep-alive settings to `createConnection`, make the default Agent method delegate to the real TCP/TLS socket factories, and support both legacy and options-object `http.globalAgent.addRequest` overloads
- defer the legacy ClientRequest `abort` event to the pending-event drain, route external HTTP client methods without cross-archive allocator drops, and make dynamic CommonJS `require("assert")` callable like Node

## Validation

- `cargo test --release -p perry-ext-http agent::tests -- --nocapture` — 9 passed
- six focused parity fixtures compile and match Node v22.23.2 byte-for-byte
- exact upstream cases now pass under the node-core shim, including:
  - `test-http-abort-queued.js`
  - `test-http-abort-stream-end.js`
  - `test-http-addrequest-localaddress.js`
  - `test-http-agent-keepalive-delay.js`
- first 25 `node:http` radar cases: **6/18 → 11/18 judged passes** (**33.3% → 61.1%**), 0 diffs, 0 compile failures, 7 node-skips
- `cargo fmt --all -- --check`, `git diff --check`, and `scripts/check_test_registration.py`

## Remaining #4975 tail

This deliberately uses `Refs`, not `Closes`: the broad tracker still includes raw HTTP/1.0 server behavior, AbortSignal and response/socket abort propagation, exposure of actual pooled `net.Socket` objects, and the larger HTTPS/TLS tail. The first 25 HTTPS cases remain at 1/22 judged passes in the same pinned diagnostic run, with 0 compile failures.

No version bump.

Refs #4975

## Summary by CodeRabbit

* **New Features**
  * Improved HTTP and HTTPS agent compatibility, including socket tracking, request queuing, connection reuse, keep-alive options, and `addRequest` support.
  * Added accurate socket-count visibility through agent properties.
  * Made aborted requests emit events in the expected deferred order.
  * Added callable CommonJS support for `require("assert")` and `require("assert/strict")`.

* **Bug Fixes**
  * Corrected reused-socket reporting and agent cleanup after completed, failed, timed-out, or aborted requests.

* **Tests**
  * Added coverage for agent queues, keep-alive connections, abort behavior, global-agent requests, and CommonJS assertions.


proggeramlug:

| check | result |
|---|---|
| all seven ratchets | 0 |

```ts
let a = new Float64Array(N);
```

PR #8617, opened 2026-08-22
perf(codegen): inline typed-array reads (f64 + i32) through module-global bindings

proggeramlug:
## What
Inline typed-array element reads through a **module-global** binding to a guarded native load, instead of a per-element runtime call — for **both** the f64 read and the i32/`|0`-context read.

## Why (found by profiling a synthetic reproducer)
Perry already inlines typed-array reads to a checked native load — but `checked_typed_array_f64_kind` (`ta_param_f64_read.rs`) and `checked_typed_array_i32_kind` (`i32_fast_path.rs`) only accept a receiver whose class the **function-local** type analysis proves (a local or param). A typed array allocated once at module scope and read inside functions —
```ts
let a = new Float64Array(N);
function f(i) { return a[i]; }   // <- fell back to a per-element runtime call
```
— the **common bundled shape** — has its class in the *module-global* proven-type map, not the local one. So every such read fell back to a per-element `js_typed_array_get` / `js_typed_array_index_get_dynamic`.

## The fix
When the function-local class proof is absent, consult `ctx.module_global_proven_types` — in both fast paths. Kind mappings factored into `f64_kind_from_class` / `i32_kind_from_class`.

**Correctness:** the loads are guard-protected — the runtime KIND-cache entry guard defers to the memory-safe helper on any miss — so an optimistic module-global proof can only cost a missed speedup, never wrong behavior. Reassigned bindings still defer.

## Measured (dense reduction over a module-global array in a hot loop)
| element type | before | after |
|---|---|---|
| `Float64Array` | 3.60s | **1.73s (~2.08×)** |
| `Int32Array` (`\|0` context) | 4.51s | **0.42s (~10.7×)** |

The i32 win is larger because the i32 accumulator stays native and **unrooted** once the read is inlined (an f64 accumulator is still nanbox-rooted — a separate, deeper numeric-representation gap, out of scope here).

## Correctness validation
- **Bit-exact** with the runtime getters across every element kind (Int8/Uint8/Uint8Clamped/Int16/Uint16/Int32/Uint32/Float32/Float64, signed/unsigned/saturating) and out-of-bounds / negative indices, via a differential against the fast path disabled.
- Integration tests `module_global_float64array_read` + `module_global_int32array_read` pin both the inline path firing (IR carries the inline fallback marker, no per-read getter call) and bit-exactness.
- Full `perry-codegen` suite green (28 binaries, 0 failures); `-D warnings` clean.

## Context
Found while systematically working the runtime-perf frontier of bundled code. The measured wins are in **per-operation lowering** (native ops vs runtime-helper calls); this is the highest-certainty pair. Kill switch: `PERRY_DISABLE_BUFFER_FAST_PATH=1`.

## Summary by CodeRabbit

- **Performance**
  - Improved performance for reads from module-global `Float64Array` and `Int32Array` values through optimized guarded access when safe.
  - Preserved runtime fallback behavior for reassigned, mismatched, or out-of-bounds values.

- **Bug Fixes**
  - Ensured optimized reads produce correct results across supported in-bounds and out-of-bounds scenarios.

- **Tests**
  - Added coverage validating consistent results across optimized and runtime access modes.


proggeramlug:

| check | result |
|---|---|
| all seven ratchets | 0 |

```ts
let a = new Float64Array(N);
```

PR #8618, opened 2026-08-22
perf(codegen): thread arena state through recursion

### LINKED ISSUE #8591: Opt target 2 — `iso_miss` (2.11x) and `interp` (1.48x): tree-walking interpreters

proggeramlug:
## Summary

Thread the stable inline-arena state pointer through directly self-recursive allocation functions. A public wrapper resolves `js_inline_arena_state()` once, while an internal body receives the pointer as a hidden argument and forwards it at every self call.

This removes the repeated arena-state lookup from the recursive hot path in the two tree-walking interpreter targets without changing specialized, async, or generator ABIs.

## Performance

Measured against `c7645d19b` with the same runtime/stdlib archives in both arms:

| target | retired instructions | quiet M1 median wall time | RSS |
| --- | ---: | ---: | ---: |
| `interp` | -3.22% | -1.92% | -32 KiB |
| `iso_miss` | -2.16% | -1.30% | +32 KiB |

The quiet timing run used 11 randomized interleaved repetitions and passed its load/foreign-process gate. Across the other 18 benchmark programs, instruction deltas stayed between -0.47% and +0.18%.

## Changes

- Detect top-level functions that are directly self-recursive and contain their own `new` allocation site.
- Emit an internal arena-threaded body plus a public ABI-preserving wrapper.
- Forward the cached arena pointer through direct self calls and share the cache loader between object and array inline allocation paths.
- Add IR regression coverage for the threaded body, single accessor, recursive pointer forwarding, and cold-function exclusion.

## Related issue

Closes #8591

## Test plan

- `cargo test -p perry-codegen --lib` — 1,168 passed, 1 ignored
- `cargo test -p perry-codegen alloc_hot_tests --lib` — 8 passed
- `./scripts/pre-tag-check.sh --quick` — passed
- `python3 scripts/gc_root_dominance_check.py .perry-trace/llvm` over `interp` + `iso_miss` shadow-root IR — 144 functions, 961 root stores, 0 violations
- 20-program A/B corpus — 20/20 baseline and candidate outputs byte-identical to Node 26.5.1
- `./scripts/test_affected_crates.sh --base upstream/main` — `perry-codegen` passed; `perry` passed 1,026/1,027 and hit the existing current-main cache audit failure for the three `PERRY_OUTLINE_ENTRY*` variables (none are touched here)
- Full gap suite attempted with prebuilt artifacts; blocked on macOS by current-main `test_gap_3527_http_ctor_prototype` SIGABRT. The detached unmodified base compiler reproduces exit 134 with no output.

## Checklist

- [x] No `CLAUDE.md`, `CHANGELOG.md`, or current-version edit
- [x] Conventional commit message

## Summary by CodeRabbit

  * Improved performance for directly recursive functions that perform allocations.
  * Benchmarks show up to 3.22% fewer instructions and median execution-time improvements of up to 1.92% in representative workloads.
* **Compatibility**
  * Preserved output behavior across the tested corpus, with results remaining byte-identical to Node 26.5.1.
* **Documentation**
  * Added benchmark results and details documenting the performance improvement.

### DIFF
--- a/changelog.d/8618-recursive-arena-state.md
+++ b/changelog.d/8618-recursive-arena-state.md
@@ -1,0 +1,11 @@
+### Performance
+
+- Thread the stable inline-arena state pointer through directly self-recursive
+  allocation functions, resolving it once in an ABI-preserving public wrapper
+  instead of repeating the runtime accessor at every recursion level.
+
+- Against `c7645d19b`, retired instructions fall 3.22% for `interp` and 2.16%
+  for `iso_miss`. An 11-repeat randomized run on a quiet M1 improves median
+  wall time by 1.92% and 1.30%, respectively. All 20 corpus outputs remain
+  byte-identical to Node 26.5.1, and the other 18 instruction deltas stay
+  between -0.47% and +0.18%.
--- a/crates/perry-codegen/src/codegen/function.rs
+++ b/crates/perry-codegen/src/codegen/function.rs
@@ -30,6 +30,37 @@ use super::typed_abi::{
     typed_string_function_name, TypedFunctionTrampolineKind, TypedParamRep,
 };
 
+/// Internal body name for a self-recursive allocator whose arena-state pointer
+/// is threaded through recursive calls (#8591).
+pub(crate) fn arena_threaded_function_body_name(public_name: &str) -> String {
+    format!("{public_name}.__arena")
+}
+
+fn emit_public_arena_threaded_wrapper(
+    llmod: &mut LlModule,
+    f: &Function,
+    public_name: &str,
+    body_name: &str,
+) {
+    let params: Vec<(LlvmType, String)> = f
+        .params
+        .iter()
+        .map(|p| (DOUBLE, format!("%arg{}", p.id)))
+        .collect();
+    let mut call_args: Vec<(LlvmType, String)> = params.clone();
+    let wrapper = llmod.define_function(public_name, DOUBLE, params);
+    wrapper.force_inline = true;
+    let entry = wrapper.create_block("entry");
+    let arena_state = entry.call(PTR, "js_inline_arena_state", &[]);
+    call_args.push((PTR, arena_state));
+    let call_args: Vec<(LlvmType, &str)> = call_args
+        .iter()
+        .map(|(ty, value)| (*ty, value.as_str()))
+        .collect();
+    let value = entry.call(DOUBLE, body_name, &call_args);
+    entry.ret(DOUBLE, &value);
+}
+
 /// Compile the internal typed-f64 clone for a conservatively eligible user
 /// function. `compile_function` emits both the public JSValue trampoline and
 /// the internal generic fallback body; guarded direct FuncRef sites can still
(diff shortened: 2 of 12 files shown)


proggeramlug:

| check | result |
|---|---|
| all seven ratchets | 0 |

```ts
let a = new Float64Array(N);
```

PR #8621, opened 2026-08-22
fix(wasm): preserve TextField callback strings on web

### LINKED ISSUE #8584: Perry/ui TextField works on Linux but not on Web target
## What happened

`perry/ui` `TextField` works correctly when compiled and run as a native Linux / GTK4 application, but it does not work correctly when the same source is compiled for the Web target.

The Web UI renders correctly and typing into the input triggers the `TextField` callback, but the entered text is not passed to the callback as a string.

Instead, the callback receives:

```text
NaN
```

and:

```text
typeof value === "number"
```

Other interactive `perry/ui` controls tested in the same Web environment work correctly:

- `Button` callbacks work
- `Toggle` returns boolean values correctly
- `Slider` returns numeric values correctly
- `setText()` updates displayed text correctly

The problem appears to be specific to the value passed by `TextField`.

## What you expected

I expected the same `TextField` source to behave consistently across native Linux and Web targets.

For example, if I type:

```text
hello
```

I expect the callback to receive:

```text
hello
```

with:

```text
typeof value === "string"
```

Instead, on the Web target I get:

```text
Raw value: NaN
Type: number
```

## Minimal reproduction

```typescript
import {
  App,
  Text,
  TextField,
  VStack,
  setText,
} from "perry/ui";

App({
  title: "TextField Test",
  width: 520,
  height: 280,

  body: VStack(16, [
    Text("Raw value: —", "raw"),
    Text("Type: —", "type"),

    TextField("Type something...", (value) => {
      setText("raw", "Raw value: " + String(value));
      setText("type", "Type: " + typeof value);
    }),
  ]),
});
```

Command used for the Web target:

```bash
perry compile issue-textfield.ts \
  --target web \
  -o issue-textfield-web
```

The generated HTML was served through a local HTTP server and opened in a browser.

For the native Linux comparison, the exact same source was compiled with:

```bash
perry compile issue-textfield.ts \
  -o issue-textfield-linux
```

On native Linux / GTK4, entering text works correctly and the callback receives the entered string.

## Environment

- **Perry version used for the native-vs-Web comparison**: `0.5.1022`
- **Host OS**: Debian 13
- **Architecture**: x86_64
- **Target showing the bug**: `web`
- **Native comparison target**: Linux / GTK4
- **Rust toolchain**: Not building Perry from source
- **Installed via**: Perry APT package

## Diagnostic output

On the Web target, after entering text:

```text
Raw value: NaN
Type: number
```

The callback is clearly being invoked because the diagnostic text is updated through `setText()`.

However, the value passed to the callback is not the text entered by the user.

With the same source on native Linux / GTK4:

```text
Raw value: hello
Type: string
```

I also tested other controls in a larger Web UI:

```text
Button    -> works
Toggle    -> works
Slider    -> works
setText   -> works
TextField -> callback receives NaN / number
```

## Anything else

The documented `TextField` API expects the callback value to be a string:

```typescript
TextField(
  placeholder: string,
  onChange: (value: string) => void,
): Widget;
```

The Web implementation also maps `TextField` to an HTML text input.

The native Linux / GTK4 target behaves according to this API in my testing, while the Web target invokes the callback with a numeric `NaN` value.

A possible explanation is a string/value conversion issue while passing the browser input value through the Web/WASM event bridge.

That is only a hypothesis based on the observed behavior and is not a confirmed root cause.

### Version note

The clean native-vs-Web comparison above was performed with Perry `0.5.1022`.

With `0.5.1022`:

- native Linux / GTK4 compilation succeeds
- native `TextField` works correctly
- Web compilation succeeds
- Web `TextField` returns `NaN` with `typeof value === "number"`

I later tested the Web target with Perry `0.5.1220` and reproduced the same `NaN / number` behavior.

I am not using `0.5.1220` as the primary native comparison in this report because native Linux `perry/ui` compilation currently encounters a separate static-library linking problem on that version.

That linker problem is being treated separately from this `TextField` issue.

### AI-assisted development disclosure

This project is being developed with AI assistance.

The reproduction design, diagnostic process, interpretation of the observed behavior, possible root-cause hypothesis, and this issue text were prepared with AI assistance based on tests and terminal output from the affected system.

All commands, builds, runtime tests, and reported outputs above were actually performed and observed on my machine.

Any proposed root cause should be treated as a hypothesis until confirmed through further testing or by the Perry maintainers.

proggeramlug:
## Summary

Preserve browser text input values as strings when invoking Perry WASM closures. The WASM bridge now performs the JS-value-to-i64 conversion exactly once, avoiding NaN payload canonicalization that made `TextField` callbacks receive numeric `NaN`.

## Changes

- Pass plain DOM strings from `TextField`, `SecureField`, and `TextArea` into `callWasmClosure`.
- Remove the same premature boxing from other WASM closure bridge call sites.
- Document the bridge contract and add a regression test that rejects pre-boxed callback arguments.

## Related issue

Fixes #8584

## Test plan

- [x] `cargo fmt -p perry-codegen-wasm -- --check`
- [x] `cargo test -p perry-codegen-wasm`
- [x] Compiled the issue's exact reproduction with `cargo run -p perry --no-default-features --features backend-wasm -- compile /tmp/perry-8584-repro.ts --target web -o /tmp/perry-8584-web.html`
- [x] Confirmed the generated web runtime passes `el.value` directly and contains no `fromJsValue(el.value)` call
- [ ] `./scripts/test_affected_crates.sh --base upstream/main` (1,026 tests pass; the unrelated `codegen_env_vars_are_build_cache_inputs` test fails because current `upstream/main` does not classify its new `PERRY_OUTLINE_ENTRY*` variables)
- [ ] `./scripts/pre-tag-check.sh --quick` (all non-format quick gates pass; current `upstream/main` has a pre-existing rustfmt mismatch in `module_global_typed_array_read.rs`)
- [x] Added a `#[test]` in the affected crate
- [x] No CLI, stdlib, or runtime API documentation change required

## Checklist

- [x] My commits follow the repository's conventional prefix style

### DIFF
--- a/changelog.d/8621-web-textfield-callback-strings.md
+++ b/changelog.d/8621-web-textfield-callback-strings.md
@@ -1,0 +1,6 @@
+### Fixed
+
+- Web `TextField`, `SecureField`, and `TextArea` callbacks now receive their
+  entered text as strings instead of numeric `NaN` values. The WASM closure
+  bridge converts browser values directly to its BigInt ABI without an
+  intermediate NaN-boxed JavaScript number. (#8584)
--- a/crates/perry-codegen-wasm/src/lib.rs
+++ b/crates/perry-codegen-wasm/src/lib.rs
@@ -101,3 +101,25 @@ fn html_escape(s: &str) -> String {
         .replace('>', "&gt;")
         .replace('"', "&quot;")
 }
+
+#[cfg(test)]
+mod tests {
+    use super::WASM_RUNTIME_JS;
+
+    #[test]
+    fn text_input_callbacks_pass_plain_js_strings_to_wasm_closures() {
+        assert_eq!(
+            WASM_RUNTIME_JS
+                .matches("callWasmClosure(el._perryCallback, el.value)")
+                .count(),
+            3,
+            "TextField, SecureField, and TextArea must pass their DOM strings directly"
+        );
+        assert!(
+            !WASM_RUNTIME_JS
+                .lines()
+                .any(|line| { line.contains("callWasmClosure") && line.contains("fromJsValue") }),
+            "callWasmClosure owns JS-value conversion; callers must not pre-box arguments"
+        );
+    }
+}
--- a/crates/perry-codegen-wasm/src/wasm_runtime.js
+++ b/crates/perry-codegen-wasm/src/wasm_runtime.js
@@ -2603,6 +2603,8 @@ function uiGet(h) { return uiHandles.get(h); }
 
 // Helper: call a WASM closure — accepts either a raw NaN-boxed f64 handle,
 // or a JS closure object ({funcIdx, captures}) from toJsValue conversion.
+// Extra args must be plain JS values: this function owns their one conversion
+// to i64 bits. Pre-boxing a string as f64 lets JS canonicalize its NaN payload.
 // WASM functions use i64 (BigInt) params/returns.
 function callWasmClosure(closureVal, ...extraArgs) {
   let closure;
@@ -2766,7 +2768,7 @@ function perry_ui_textfield_create(placeholder, callback) {
   el.placeholder = placeholder || "";
   el._perryCallback = callback;
   el.addEventListener("input", () => {
-    if (el._perryCallback !== undefined) callWasmClosure(el._perryCallback, fromJsValue(el.value));
+    if (el._perryCallback !== undefined) callWasmClosure(el._perryCallback, el.value);
   });
   return uiAlloc(el);
 }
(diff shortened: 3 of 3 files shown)


proggeramlug:
Landed via https://github.com/PerryTS/perry/pull/8626, which carries this branch's commits plus #8622's missing `changelog.d` fragment.

Validated on the combined tree: `cargo check --workspace --all-targets` exit 0, `perry-runtime --lib` 2636 passed, all seven ratchets and `cargo fmt --check` clean, with the ratchets re-run against the current baseline immediately before merge.

PR #8622, opened 2026-08-22
test(link): guard prebuilt stdlib HTTP isolation (#8587)

### LINKED ISSUE #8587: Native Linux perry/ui linking fails because HTTP symbols from libperry_ext_http.a remain unresolved
## What happened

A minimal native Linux application using `perry/ui` fails during the final link step on Perry `0.5.1220`.

The TypeScript source compiles and Perry successfully generates the object file, but `libperry_stdlib.a` references HTTP-related symbols that are not resolved during linking.

The first minimal build fails with symbols such as:

```text
js_ext_http_client_inflight
js_http_has_pending
```

I inspected the Perry `0.5.1220` release archives with `nm` and confirmed that these symbols are actually defined in:

```text
libperry_ext_http.a
```

but the final native `perry/ui` link does not resolve them.

I then tested a variant containing a real global `fetch()` reference. That changes Perry from:

```text
Linking (runtime-only)...
```

to:

```text
Linking (with stdlib)...
```

and changes the unresolved symbols, but the native link still fails with additional HTTP-related symbols that are also present in `libperry_ext_http.a`.

So this appears to be related to native static-library dependency selection and/or link ordering rather than missing symbols in the distributed archive.

## What you expected

A minimal `perry/ui` application should compile into a native Linux executable without needing the application itself to use HTTP.

The required runtime/stdlib dependencies should be linked automatically.

This is especially surprising because a similar native `perry/ui` application compiled successfully for me with Perry `0.5.1022`.

## Minimal reproduction

```typescript
import {
  App,
  Text,
  VStack,
} from "perry/ui";

App({
  title: "Perry UI Test",
  width: 520,
  height: 280,

  body: VStack(16, [
    Text("Hello from Perry"),
  ]),
});
```

Command I ran:

```bash
perry compile test.ts -o test-linux
```

The compilation reaches the linker and then fails.

Example output:

```text
Collecting modules...
Found 1 module(s): 1 native, 0 JavaScript
Generating code...
Stored cached object: ...
note: Perry workspace source not found — linking the prebuilt full stdlib (larger binary).
Linking (runtime-only)...
...
Linking perry/ui (native UI) from ...
/usr/bin/ld: ...libperry_stdlib.a(...): undefined reference to `js_ext_http_client_inflight'
/usr/bin/ld: ...libperry_stdlib.a(...): undefined reference to `js_http_has_pending'
collect2: error: ld returned 1 exit status
Error: Linking failed
```

## Environment

- **Perry version**: `0.5.1220`
- **Host OS**: Debian 13
- **Architecture**: x86_64
- **Target**: native Linux / GTK4
- **Rust toolchain**: Not building Perry from source
- **Installed via**: GitHub release archive `perry-linux-x86_64.tar.gz`
- **System linker**: `/usr/bin/ld`
- **C toolchain**: installed
- **GTK4 dependencies**: installed
- **OpenSSL development libraries**: installed (`libssl-dev`)

The Perry release is kept together with its bundled libraries in one directory:

```text
perry
libperry_runtime.a
libperry_runtime_abort.a
libperry_stdlib.a
libperry_ui_gtk4.a
libperry_ext_http.a
libperry_ext_fetch.a
libperry_ext_http_server.a
...
```

## Diagnostic output

I checked whether the initially unresolved symbols exist in the distributed archives.

Command:

```bash
PERRY_ROOT="$HOME/perry-surumleri/0.5.1220/acilmis"

nm -A --defined-only "$PERRY_ROOT"/libperry*.a |
  grep -E 'js_ext_http_client_inflight|js_http_has_pending'
```

Result:

```text
libperry_ext_http.a:... T js_ext_http_client_inflight
libperry_ext_http.a:... T js_http_has_pending
```

At the same time, `libperry_stdlib.a` references them as undefined:

```bash
nm -A "$PERRY_ROOT/libperry_stdlib.a" |
  grep -E 'js_ext_http_client_inflight|js_http_has_pending'
```

Result:

```text
libperry_stdlib.a:... U js_ext_http_client_inflight
libperry_stdlib.a:... U js_http_has_pending
```

So the symbol providers are present in the same Perry distribution.

### Additional fetch experiment

As a diagnostic experiment, I added a real global `fetch()` reference:

```typescript
async function httpLinkTest() {
  await fetch("http://127.0.0.1:9/");
}

void httpLinkTest;
```

With that present, Perry changes the build from:

```text
Linking (runtime-only)...
```

to:

```text
Linking (with stdlib)...
```

After installing the required OpenSSL development libraries, the build proceeds further but still fails with HTTP-related unresolved symbols such as:

```text
js_http_response_trailers
js_ext_http_agent_is_handle
js_ext_http_agent_dispatch_method
js_ext_http_agent_dispatch_property
js_ext_http_agent_dispatch_property_set
js_ext_http_client_incoming_message_is_handle
js_ext_http_client_incoming_message_set_encoding
```

I checked those symbols as well:

```bash
PERRY_ROOT="$HOME/perry-surumleri/0.5.1220/acilmis"

for SYMBOL in \
  js_http_response_trailers \
  js_ext_http_agent_is_handle \
  js_ext_http_agent_dispatch_method \
  js_ext_http_agent_dispatch_property \
  js_ext_http_agent_dispatch_property_set \
  js_ext_http_client_incoming_message_is_handle \
  js_ext_http_client_incoming_message_set_encoding
do
  echo "=== $SYMBOL ==="

  nm -A --defined-only "$PERRY_ROOT"/libperry*.a 2>/dev/null |
    grep -E "[[:space:]]$SYMBOL$"
done
```

Every one of these symbols is defined by:

```text
libperry_ext_http.a
```

For example:

```text
libperry_ext_http.a:... T js_http_response_trailers
libperry_ext_http.a:... T js_ext_http_agent_is_handle
libperry_ext_http.a:... T js_ext_http_agent_dispatch_method
libperry_ext_http.a:... T js_ext_http_agent_dispatch_property
libperry_ext_http.a:... T js_ext_http_agent_dispatch_property_set
libperry_ext_http.a:... T js_ext_http_client_incoming_message_is_handle
libperry_ext_http.a:... T js_ext_http_client_incoming_message_set_encoding
```

Despite that, they remain unresolved during the final native link.

## Possible diagnosis / solution direction

I have not confirmed the root cause, but the evidence seems consistent with one of these possibilities:

1. `libperry_stdlib.a` now has unconditional references to HTTP activity/runtime functions, while `libperry_ext_http.a` is only selected when Perry detects explicit HTTP/fetch usage.

2. `libperry_ext_http.a` is selected in some cases but appears at an ineffective position in the static archive link order.

3. The relevant archives may need to be linked as a group / rescanned on Linux if there are circular static-library dependencies.

For example, if there is a circular dependency between:

```text
libperry_stdlib.a
libperry_ext_http.a
libperry_runtime.a
```

the Linux linker may require different ordering or a linker group such as:

```text
-Wl,--start-group
...
-Wl,--end-group
```

This is only a hypothesis, not a confirmed fix.

Another possible fix would be for Perry's dependency analysis to automatically include the HTTP extension whenever the bundled stdlib contains live references to those HTTP runtime symbols.

### AI-assisted development disclosure

This project is being developed with AI assistance.

The diagnostic process, possible root-cause hypotheses, suggested investigation directions, and this issue text were prepared with AI assistance based on terminal output and tests I performed on the affected system.

All commands, reproductions, compiler/linker output, runtime behavior, and symbol-table results reported above were observed directly on my machine.

Any proposed root cause or solution direction should be treated as a hypothesis until confirmed by the maintainers or further testing.

proggeramlug:
## Summary

- document that the release/default stdlib feature graph may contain only self-contained implementations
- add a regression test that follows transitive Cargo feature edges and rejects either external HTTP pump in the default release archive

## Root cause

Perry 0.5.1220 briefly enabled `external-http-client-pump` through `perry-stdlib/full`. That feature emits references provided only by `libperry_ext_http.a`, while HTTP-free `perry/ui` builds do not select that archive. The same coupling accounts for the extra symbols seen after adding `fetch()`.

The behavior fix already landed in #5983 and shipped in 0.5.1239, but it had no focused regression guard. This PR pins that release-link contract so the 0.5.1220 configuration cannot recur, including through an indirect feature edge.

## Tests

- `cargo test -p perry --test issue_8587_prebuilt_stdlib_http_isolation`
- `cargo test -p perry-stdlib --lib` (120 passed)
- `rustfmt --edition 2021 --check crates/perry/tests/issue_8587_prebuilt_stdlib_http_isolation.rs`
- `git diff --check`

No version bump.

Closes #8587

## Summary by CodeRabbit

* **Documentation**
  * Clarified that the full standard library feature includes self-contained implementations and excludes external HTTP pump features.
  * Documented the link-time impact of enabling an external HTTP pump.

* **Tests**
  * Added regression coverage verifying external HTTP pump features remain disabled by default.

### DIFF
--- a/crates/perry-stdlib/Cargo.toml
+++ b/crates/perry-stdlib/Cargo.toml
@@ -19,7 +19,10 @@ crate-type = ["rlib"]
 # Default: include everything for backwards compatibility
 default = ["full"]
 
-# Full stdlib - everything included
+# Full stdlib - every self-contained implementation. External pump features
+# must stay out of this list: release archives enable `full` without linking
+# their per-program provider archives, and adding an external HTTP pump here
+# made HTTP-free Linux UI links require libperry_ext_http.a (#5983, #8587).
 full = ["http-server", "http-client", "database", "crypto", "compression", "email", "websocket", "image", "scheduler", "ids", "html-parser", "rate-limit", "validation", "net", "tls", "bundled-dotenv", "bundled-lru-cache", "bundled-exponential-backoff", "bundled-events", "bundled-decimal", "bundled-dayjs", "bundled-moment", "bundled-commander", "bundled-streams"]
 
 # Minimal core - just what's needed for basic programs
--- a/crates/perry/tests/issue_8587_prebuilt_stdlib_http_isolation.rs
+++ b/crates/perry/tests/issue_8587_prebuilt_stdlib_http_isolation.rs
@@ -1,0 +1,68 @@
+//! Regression coverage for #8587.
+//!
+//! Release archives build `perry-stdlib-static` with its default feature set.
+//! That eventually enables `perry-stdlib/full`, which must remain
+//! self-contained: the `external-http-*-pump` features declare C symbols that
+//! are provided only when the compiler also selects `libperry_ext_http.a`.
+//! Making either pump part of `full` breaks otherwise HTTP-free prebuilt
+//! consumers such as a minimal Linux `perry/ui` application.
+
+use std::collections::BTreeSet;
+use std::path::PathBuf;
+
+fn local_feature_closure(
+    features: &toml::Table,
+    roots: impl IntoIterator<Item = String>,
+) -> BTreeSet<String> {
+    let mut pending: Vec<String> = roots.into_iter().collect();
+    let mut enabled = BTreeSet::new();
+
+    while let Some(feature) = pending.pop() {
+        if !enabled.insert(feature.clone()) {
+            continue;
+        }
+
+        let Some(members) = features.get(&feature).and_then(toml::Value::as_array) else {
+            continue;
+        };
+        for member in members.iter().filter_map(toml::Value::as_str) {
+            // Only bare names refer to another feature in this manifest.
+            // `dep:name` and `crate/feature` entries leave this feature graph.
+            if !member.starts_with("dep:") && !member.contains('/') {
+                pending.push(member.trim_end_matches('?').to_owned());
+            }
+        }
+    }
+
+    enabled
+}
+
+#[test]
+fn release_full_stdlib_does_not_enable_external_http_pumps() {
+    let manifest_path =
+        PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("../perry-stdlib/Cargo.toml");
+    let source = std::fs::read_to_string(&manifest_path)
+        .unwrap_or_else(|error| panic!("failed to read {}: {error}", manifest_path.display()));
+    let manifest: toml::Value = toml::from_str(&source)
+        .unwrap_or_else(|error| panic!("failed to parse {}: {error}", manifest_path.display()));
+    let features = manifest
+        .get("features")
+        .and_then(toml::Value::as_table)
+        .expect("perry-stdlib must declare a [features] table");
+
+    for pump in ["external-http-client-pump", "external-http-server-pump"] {
+        assert!(
+            features.contains_key(pump),
+            "the regression guard must track the current external HTTP pump feature name: {pump}"
+        );
+    }
+
+    let default = local_feature_closure(features, ["default".to_owned()]);
+    for pump in ["external-http-client-pump", "external-http-server-pump"] {
+        assert!(
+            !default.contains(pump),
+            "perry-stdlib's release/default feature graph must not enable `{pump}`: it creates \
+             unresolved libperry_ext_http.a references in HTTP-free prebuilt links (#8587)"
+        );
+    }
+}


proggeramlug:
Landed via https://github.com/PerryTS/perry/pull/8626, which carries this branch's commits plus #8622's missing `changelog.d` fragment.

Validated on the combined tree: `cargo check --workspace --all-targets` exit 0, `perry-runtime --lib` 2636 passed, all seven ratchets and `cargo fmt --check` clean, with the ratchets re-run against the current baseline immediately before merge.

PR #8623, opened 2026-08-22
perf(codegen): raise root-spill default to the measured RS4GC fan-out cliff (#8620)

### LINKED ISSUE #8620: codegen: #8589 root-spill default threshold (4M) too low — spills moderate-fan-out functions and slows their compile
## Summary
`DEFAULT_ROOT_SPILL_RELOCATIONS` (4,000,000, `codegen/helpers.rs`) looks too low: on a synthetic entry function that estimates ~8M relocations, **spilling made the compile slower than just letting RS4GC fan out** — so the default fires on moderate cases that don't need it and pays the shadow-frame overhead for nothing.

## Data (single runs, contended host — treat as rough)
Synthetic: N=2000 top-level object bindings, all live via a `prev` chain, each built through a call → ~2000 roots × ~2000 safepoints ≈ 4–8M estimated relocations.

| config | compile |
|---|---|
| no outline, **spill OFF** (plain RS4GC fan-out) | **180s** |
| no outline, **spill ON** (4M default → spills `main`, 8M est) | 303s |

The spilled build emits ~4001 `js_shadow_slot_*` for the frame; that IR is itself expensive to build, and here it cost more than the fan-out it avoided. RS4GC fan-out only becomes catastrophic at much higher relocation counts (the Claude Code `cli.js` `@main`: ~74M est, never finishes at `-Os`).

## Recommendation
Raise the default so spilling fires only in the genuinely-catastrophic regime. The exact crossover is unmeasured — it's between ~8M (fan-out fine at 180s) and ~74M (fan-out never finishes) — so bracket it with a couple of synthetic points (e.g. 16M / 32M / 48M) and set the default just below where fan-out stops finishing in bounded time. Raising is backstopped by the post-RS4GC instruction-budget assertion (#8586), which fails loudly rather than hanging if the new threshold is ever too high for a given function.

## Note
This is compile-time only — spilling `main` (run-once init) does not affect the emitted binary's runtime. Lower priority than the runtime-perf work, but a real compile-time regression on moderate-fan-out functions. Context: #8583 (the regression), #8589 (the spill), #8586 (the budget backstop).

proggeramlug:
## Summary

Fixes #8620. `DEFAULT_ROOT_SPILL_RELOCATIONS` (the `slots × safepoints`
estimate above which a function keeps its GC roots in a shadow frame instead
of native RS4GC statepoints, shipped in #8589) was **4,000,000** — low enough
to spill moderate-fan-out functions whose statepoints would have optimized
fine. On an ~8M-relocation entry function, spilling was measured *slower* than
the fan-out it replaced (303 s spilled vs 180 s fanned out, #8620): the default
paid shadow-frame overhead for nothing.

This raises the default to **32,000,000**, the largest estimate whose RS4GC
fan-out still finishes in bounded time.

## Measurement

Synthetic entry functions (`N` top-level `let mI = reg({id, prev: m<I-1>, v})`
bindings — the `prev` chain keeps all roots live, each `reg(...)` is a
safepoint, giving `estimate = 2N × (N+1)`) compiled at `-Os` with spilling OFF
(`PERRY_ROOT_SPILL_RELOCATIONS=0`, pure RS4GC fan-out). The `@main` codegen
unit was timed (each run bounded by `timeout 1200`):

| N | estimate | `@main` RS4GC fan-out | outcome |
|---|---------:|----------------------:|---------|
| 2000 |  8,004,000 | ~325 s | finished |
| 2828 | 16,000,824 | ~235 s | finished |
| 4000 | 32,008,000 | ~511 s (8.5 min) | finished |
| 4472 | 40,006,512 | > 1200 s | did not finish (20 min bound) |
| 4900 | 48,029,800 | > 1200 s | did not finish (20 min bound) |

The fan-out cliff sits between **32M (finishes)** and **40M (hangs)**. Timing
is noisy at this scale (16M came in under 8M on a contended host) but the
finish/hang boundary is unambiguous. None of the finishing runs tripped the
#8586 budget assertion, so these are genuine optimizer-super-linearity times,
not budget truncation. This host is ~1.8× slower than the one in #8620 (its 8M
fan-out was 180 s vs 325 s here), so 32M here maps to a *higher* estimate on a
faster host — i.e. the threshold is conservative and spills no earlier than a
slow-host / CI crossover would require.

## Chosen threshold + rationale

**32,000,000.** The largest measured estimate whose fan-out still finished in
bounded time. Below it, fan-out is the cheaper lowering (spilling a moderate
function costs more than the fan-out it avoids); above it, fan-out risks not
finishing and the shadow frame wins. It is 8× the old default, 4× the ~8M
regression point flagged in #8620, and well below the genuinely-catastrophic
Claude Code `cli.js` `@main` (~8.4e7, never finishes at `-Os`), which still
spills.

Raising the threshold is safe because it is **backstopped by #8586's post-RS4GC
instruction-budget assertion** (`PERRY_LL_RS4GC_MAX_INSTRS`): if a function this
estimate misses still fans out past the per-function budget, the compile fails
loudly with the function name and its pre/post sizes rather than hanging.

## Scope / tests

- Compile-time only: spilling `main` (run-once init) does not change the
  emitted binary's runtime behavior.
- New unit tests pin the default at the measured cliff and assert the moderate
  ~8M case stays on statepoints while the catastrophic ~8.4e7 case still spills.
- `RUSTFLAGS="-D warnings" cargo check -p perry-codegen --lib` clean;
  `cargo test -p perry-codegen --lib` = 1170 passed / 0 failed.
- `gc_root_spill_mixed_frames_8583.rs` sets `PERRY_ROOT_SPILL_RELOCATIONS`
  explicitly on every compile, so it never consults the default and is
  unaffected.

## Summary by CodeRabbit

- **Performance**
  - Increased the default threshold for native garbage-collection root spilling, which may improve runtime performance in workloads with high statepoint fan-out.
  - Added safeguards to spill exceptionally large relocation sets, helping maintain reliable generated code.

- **Configuration**
  - Existing override behavior remains available through `PERRY_ROOT_SPILL_RELOCATIONS`, including setting it to `0` to disable spilling.

- **Documentation**
  - Documented the updated default, performance measurements, and compile-time-only impact.


proggeramlug:

Issue #8619, opened 2026-08-22
codegen: proven-Number f64 locals stay nanbox-rooted (not raw double) in functions that read typed arrays — ~4x f64-vs-i32 gap


proggeramlug:
## Summary
In a function that reads a typed array, numeric f64 locals — including a **fresh, non-loop-carried** one — are stored as **nanbox GC roots** (with a per-write `js_write_barrier_root_nanbox`) instead of raw `double`s, even when their value is a proven `Number`. This is measurable: a dense `Float64Array` reduction runs ~4× slower than the equivalent `Int32Array` one, and the gap is the f64 accumulator's rooting + write-barrier overhead (the i32 accumulator stays native and unrooted).

Surfaced while inlining typed-array reads (#8617). That PR removed the per-read runtime call (`Float64Array` 3.60s → 1.73s); the remaining ~1.73s vs the i32 path's 0.42s is this rooting overhead.

## Reproducer
```ts
let arr = new Float64Array(4096);
for (let i = 0; i < 4096; i++) { arr[i] = i * 0.5; }
function hot() {
  let s = 0.0;
  for (let i = 0; i < 4096; i++) {
    let x = arr[i] + 1.0;   // fresh, proven Number
    s = s + x;
  }
  return s;
}
console.log("s:" + hot());
```
`PERRY_SAVE_LL` on `hot`: **0** `alloca double`, **4** `alloca ptr addrspace(1)` (nanbox), **7** `js_write_barrier_root_nanbox`, **0** `fadd`. Both `x` and `s` are nanbox-rooted; the loop compare `i < 4096` also round-trips through f64 (`sitofp` + `fcmp` + nanbox-bool + `icmp eq`) instead of a native `icmp slt i32`.

For comparison, a pure-numeric loop with no typed-array in the function keeps some `alloca double` + `fadd`. So the typed-array involvement broadly disables the raw-double representation for the whole function — even for locals unrelated to the array.

## What to investigate
1. Why representation selection (repsel / numeric-envelope) declines the raw-`double` slot for a proven-`Number` local in a function that reads a typed array — is a conservative flag set function-wide?
2. The loop-carried accumulator `s = s + <number>` needs a fixpoint proof to keep its raw-double representation across the back-edge.
3. The i32 loop-condition (`i < const`) should lower to a native `icmp slt i32`, not the f64 round-trip. (LLVM may already fold the `sitofp`+`fcmp`; worth confirming against the optimized asm before treating it as a runtime cost.)

[...]

PR #8625, opened 2026-08-22
perf(codegen): un-root typed-array-param numeric accumulators

proggeramlug:
## Summary

Addresses #8619 for the typed-array **parameter** case. A function that folds a spec-ABI-proven typed-array parameter into an accumulator — `function reduce(arr: Float64Array) { let s = 0.0; for (…) { let x = arr[i] + 1.0; s = s + x; } }` — kept its numeric locals `x` and `s` as NaN-boxed **GC roots** with a per-write `js_write_barrier_root_nanbox`, and lowered `s`'s update to the opaque `js_dynamic_string_or_number_add` instead of an inline `fadd` — even though every value is a genuine Number.

## Root cause

The `number_by_construction` fixpoint's `numeric_view_value_or_undefined` (`collectors/ptr_shape_numeric.rs`) recognised a typed-array element read as "a Number or `undefined`, never a pointer" **only** for a LOCAL view with a compiler-visible `TypedArrayNew` initializer — not for a spec-proven `TaPtr` parameter. So the fresh, read-derived `x` failed the numeric proof, which cascaded to the loop-carried accumulator `s = s + x`. Both then kept a shadow root slot (retyped to `alloca ptr addrspace(1)`), and `s`'s add stayed the dynamic helper.

## Fix

The fixpoint now also treats a read off a `spec_ta_lens` binding as Number-or-`undefined`. `spec_ta_lens` is keyed exactly by `SpecParamRep::TaPtr` parameters, and `collectors::spec_abi_sites` admits a `TaPtr` only for `spec_ta_kind_is_numeric` kinds (the BigInt typed arrays — whose elements are BigInt pointers — are never `TaPtr`), so `arr[numeric_index]` off one is provably a Number in-bounds and `undefined` out of range, which `+` launders into a genuine Number (`NaN` at worst). The `rec(index)` guard is retained — a non-numeric key would read a property, which can be a pointer.

## Soundness

- Fires only for spec-ABI `TaPtr` params — the entry contract binds the raw header of a proven numeric, **non-view**, **never-reassigned** typed array. Never trusts the erased `Float64Array` annotation.
- BigInt element kinds excluded by construction (`spec_ta_kind_is_numeric`).
- Non-numeric index (symbol/string key) still disqualifies via `rec(index)`.
- Verified **byte-identical** output vs the rooted build (`PERRY_NUMBER_BY_CONSTRUCTION=0`) under every moving-GC configuration (`PERRY_GC_SCAVENGE_NURSERY_MB=1/2/4`, `PERRY_GEN_GC=0`, default) **and** vs Node.

## Measured

200000×4096 `Float64Array` reduction passed by parameter: the accumulator's per-iteration `js_dynamic_string_or_number_add` + root barrier become a single `fadd` in a raw `double` slot — **~5× faster (5.1–7.3s → ~1.0s)**.

## Tests

- **Unit** (`perry-codegen`): `spec_ta_param_view_admits_read_derived_number_locals` + `ta_read_without_spec_proof_stays_dynamic` — the fix is load-bearing (with the spec proof both locals are admitted; without it neither is). `cargo test -p perry-codegen` green.
- **Integration** (`perry`): `gc_ta_view_accumulator_unroot_8619` — a rooted-vs-fix differential run across the moving-GC matrix, covering `Float64Array`/`Int32Array` kinds and OOB/negative indices (NaN), asserting byte-identical output.

## Not covered

The issue's literal reproducer uses a **module-global** typed array. On `main` that read is still a runtime call (module-global read inlining, #8617, is unmerged), so its accumulator rooting is a secondary cost there; extending the same proof to `module_global_proven_types` is the natural follow-up once the read inlines.

## Summary by CodeRabbit

  * Numeric calculations derived from proven typed-array parameters can now remain optimized numeric values, improving arithmetic efficiency and reducing runtime overhead.
  * Typed-array reductions and accumulators receive improved optimization, including under moving garbage collection.

* **Bug Fixes**
  * Added validation ensuring only proven typed-array views receive the optimization.
  * Improved handling of out-of-range typed-array reads and accumulator arithmetic.

* **Tests**
  * Added coverage for typed-array numeric propagation and garbage-collection correctness across multiple collector configurations.

### DIFF
--- a/changelog.d/8619-ta-view-param-number-by-construction.md
+++ b/changelog.d/8619-ta-view-param-number-by-construction.md
@@ -1,0 +1,38 @@
+A hot function that folds a typed-array **parameter** into an accumulator —
+`function reduce(arr: Float64Array) { let s = 0.0; for (…) { let x = arr[i] + 1.0;
+s = s + x; } }` — no longer keeps its numeric locals as NaN-boxed GC roots.
+
+The spec-ABI already proves such a parameter (`TaPtr`) permanently holds one
+specific numeric-kind, non-view typed array, and specializes the body so the
+element read inlines. But the `number_by_construction` fixpoint
+(`collectors/ptr_shape_numeric.rs`) only recognised a typed-array element read as
+"a Number or `undefined`, never a pointer" for a **local** view with a
+compiler-visible `TypedArrayNew` initializer — not for a proven `TaPtr`
+parameter. So the fresh, read-derived `x` failed the numeric proof, which
+cascaded to the loop-carried accumulator `s = s + x`. Both then kept a shadow
+root slot with a per-write `js_write_barrier_root_nanbox`, and `s`'s update
+lowered to the opaque `js_dynamic_string_or_number_add` call instead of an inline
+`fadd`.
+
+The fixpoint now also treats a read off a `spec_ta_lens` binding as
+Number-or-`undefined`. `spec_ta_lens` is keyed exactly by `SpecParamRep::TaPtr`
+parameters, and `collectors::spec_abi_sites` admits a `TaPtr` only for
+`spec_ta_kind_is_numeric` kinds (the BigInt typed arrays — whose elements are
+BigInt pointers — are never `TaPtr`), so `arr[numeric_index]` off one is provably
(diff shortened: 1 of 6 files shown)


proggeramlug:

PR #8627, opened 2026-08-22
fix(compile): --report-size duplicate-body false positives + overclaimed size

jdalton:
## Summary

Two accuracy fixes to the `--report-size` enrichment from [#8579](https://github.com/PerryTS/perry/pull/8579), both found by testing against a live compiled binary rather than trusting the code in isolation.

<details>
<summary>What was wrong and how it was found</summary>

- **Duplicate function/static-data body false positives**: detection keyed on an FNV-1a hash + size check. CodeRabbit correctly flagged that a hash collision at the same size could fabricate a false duplicate finding. Fixed properly rather than reworded the comment: key directly on the exact byte slice (`&[u8]` is `Ord`) instead of hashing it — exact by construction, no more expensive to implement.
- **Overclaimed shipped-binary size**: the "duplicate crate instance" finding's first draft claimed its `total_bytes` were recoverable shipped-binary size. I verified directly (`md5` + `objdump` on the extracted archive members from a real compiled program) that `perry-runtime`/`perry-stdlib` do redundantly compile some shared dependencies (`gimli`, confirmed byte-identical across their two separate `.a` archives) — but a successful link only pulls **one** physical copy per symbol (a linker errors on true duplicate-symbol inclusion), so every byte attributed is real, in-use code in the shipped binary, not a duplicate sitting in it twice. Renamed `duplicate_crate_versions` → `duplicate_crate_instances` in the JSON schema (unreleased, so free to fix) and set `estimated_bytes` to `0` for this finding so it can't misrank against suggestions that genuinely shrink the shipped binary.

</details>

## Test plan

- [x] Unit tests pass (10/10).
- [x] `scripts/check_file_size.sh` passes.
- [x] End-to-end: recompiled a `console.log` program with `--report-size`, confirmed the "Duplicate crate instances" section now reads accurately and the duplicate-body false positive is gone.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `--report-size` accuracy by eliminating false duplicate detections caused by hash collisions.
  * Duplicate symbol bodies are now compared exactly, and duplicate crate instances are reported more accurately.
  * Recoverable size estimates for redundant crate compilation now correctly report zero bytes.

* **Documentation**
  * Updated size-report output descriptions and JSON field naming for clarity.

* **Chores**
  * Updated the release version to `0.5.1520`.


proggeramlug:

PR #8632, opened 2026-08-23
fix(http): close Node HTTP/HTTPS parity tail

### LINKED ISSUE #4975: node:http/https behavioral parity tail (26.2% → ...) — successor to #2132

proggeramlug:
## Summary

- close the remaining Node HTTP/HTTPS Agent, socket, lifecycle, request, response, and event parity gaps
- complete the TLS surface for CA and hostname verification, peer certificates, SNI, ALPN, sessions, ticket keys, keylog, and `tlsClientError`
- harden moving-GC roots, listener semantics, dynamic dispatch, build-cache inputs, and the Node parity runner itself

## Node parity

Validated with the issue pinned command:

```console
scripts/node_core_subset.py --root vendor/nodejs --api http https --max-per-api 25 --timeout 30
```

| API | Pass | Diff | Runtime fail | Compile fail | Node skip | Judged parity |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `http` | 18 | 0 | 0 | 0 | 7 | 100% |
| `https` | 21 | 0 | 0 | 0 | 4 | 100% |

Total: 39 passing tests, 11 Node-oracle skips, and no Perry failures.

## Validation

- `cargo test -p perry-ext-http` (82 passed)
- `cargo test -p perry-stdlib --lib` (120 passed)
- deterministic `perry-ext-events` and `perry-ext-net` library tests (34 passed)
- manifest consistency and build-cache input tests
- warnings-as-errors checks for HTTP, net, events, stdlib, and runtime crates
- file-size, raw-handle, address-class, unrooted-local, and diff-integrity gates

No package version bump is included.

Closes #4975


proggeramlug:
Landed on `main` in e5a2a6f5e via #8635. Audit notes there.

Validated at head `24d9dacee`: `perry-runtime` 2638/0, `perry-ext-http` 82/0, `cargo check --workspace --all-targets` clean, all nine ratchets + fmt.

Two things worth knowing:

1. Your 05:13 push added `PENDING_TLS_CLIENT_ERRORS` and `PENDING_TLS_KEYLOGS`, which `gc_runtime_root_holders` flags as unclassified. I researched both — each holds an `i64` server handle (side-table index, stable across collection) plus owned Rust data — and recorded `not_a_gc_pointer` verdicts. Worth running that gate locally before pushing; it caught these within seconds.

2. On the new PKCS#12 deps: I diffed the full `cargo audit` advisory list rather than comparing totals. 14 advisories on both trees, identical — `des`/`p12`/`rc2`/`yasna` introduce no new advisory.

Separately, `perry-ext-net`'s `gc_mutable_scanner_rewrites_listener_roots` fails on clean `main` too, so it is not yours — but `main` is carrying that red.

PR #8633, opened 2026-08-23
perf(codegen): make the RS4GC root-spill estimate see allocating literals + call-result temporaries (#8583)

proggeramlug:
## Problem (#8583)

With the fan-out fix (#8585/#8586/#8589/#8593) shipped, the Claude Code 2.1.112 bundle still does not compile to completion: **unit 4 hangs 3+ hours** on one closure, `perry_closure_cli_2_1_112_js__33499`. Every other unit finishes; RS4GC on this one fans out (401s + **29.75 GiB RSS** in a standalone replay before timeout).

It is the exact `@main` mechanism — a huge `roots × safepoints` product super-linear in `rewrite-statepoints-for-gc` — but the shadow-frame spill estimate never fired for it, so it stayed on native statepoints instead of spilling.

## Root cause

`__33499` is a **giant constant nested array literal** (a bundle data table): it lowers to **11,104 `js_array_from_values` allocations** (one per sub-array) and ~20k GC safepoints. The spill estimate `slot_count × count_safepoint_sites(body)` under-counted both terms by ~100×:

1. `count_safepoint_sites` counted only `Call`/`New`-family expressions — **not allocating object/array literals**, even though each lowers to a collecting runtime call that RS4GC gives a statepoint. So a data table (no `Expr::Call` at all) looked safepoint-free.
2. `slot_count` was the shadow-slot map size = **named pointer locals only**, not the ~one pointer temporary each call result leaves live across later safepoints — the roots RS4GC actually relocates.

The #8586 post-rewrite budget assertion cannot catch this: it is checked *after* the rewrite completes, and here the rewrite never finishes.

## Fix

- `collectors/safepoint_sites.rs`: count `Expr::{Object,ObjectSpread,ObjectAssign,Array,ArraySpread}` as safepoints.
- `codegen/helpers.rs`: count call-result temporaries in the root term — `live_roots = slot_count + sites`.

Both are needed: literals are the missing safepoints, temporaries the missing roots. Over-approximation biased toward spilling, matching the estimate's stated design (a false-positive shadow frame is cheap; a missed fan-out is not). A function needs ~2000+ allocating operations to cross the 32M default, i.e. only genuinely huge (usually module-init) functions — negligible over-spill risk on hot code. Composes with #8623 (32M threshold): `__33499`'s estimate is ~411M, well over 32M.

## Verification

A faithful synthetic (a function returning a 3,000-row nested constant array) reproduces `__33499`'s shape. With the fix it spills (`3001 roots × 3001 safepoints = 9,006,001`), **compiles in 64s instead of not finishing**, and produces byte-correct output (`t[1500][2] = 44`). Unit tests added for literal/nested-literal safepoint counting.

**Still pending (this machine is disk-starved by concurrent builds):** full `cargo test -p perry-codegen` run and the end-to-end cli.js acceptance compile (unit 4 completes + `cc --version`). Will attach results. A pre-RS4GC loud-guard (turn any missed fan-out into a fast error instead of a multi-hour wedge) is a sensible defense-in-depth follow-up, as is the real perf optimization: lowering large constant data-table literals to a static rodata descriptor + one bulk-materialization call instead of N per-element allocations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved garbage-collection root-spill estimation for functions containing object and array literals.
  * Accounts for additional allocation points and call-result pointers, reducing the risk of underestimating required spill space.
  * Updated spill diagnostics to report more accurate root counts and relocation estimates.

* **Documentation**
  * Documented the updated estimation behavior and its default threshold for very large functions.


proggeramlug:
Landed on `main` in ad25921e7 via #8634.

Audit notes are in #8634. Summary: the change is sound — the two-axis under-count is real and the over-approximation is biased the right way. All nine ratchets plus `cargo fmt` pass, `perry-codegen` is 1182/0, and I asserted your two new safepoint tests by name rather than inferring them from the count.

One thing I fixed on the landing branch: your change moves the production formula to `(slot_count + sites) × sites`, but #8623's two endpoint tests still called `root_relocation_estimate(slot, sites)` directly — so they were pinning a formula production no longer uses (moderate 8.0M vs 12.0M, catastrophic 84.3M vs 11.3B). Both still land on the correct side of the 32M threshold so nothing was broken, but those tests exist to guard both ends of that threshold and would no longer have caught a regression in the `live_roots` composition you're introducing. I extracted `spill_live_root_count()` so production and the tests share one definition.

Nice find on the `Expr::Array` safepoint gap — a data table with no `Expr::Call` reading as safepoint-free is a good catch.

Issue #8636, opened 2026-08-23
perry-ext-net: gc_mutable_scanner_rewrites_listener_roots fails on main


proggeramlug:
`perry-ext-net`'s `gc_mutable_scanner_rewrites_listener_roots` fails on clean `main` (reproduced at `ad25921e7`).

```
test tests::gc_mutable_scanner_rewrites_listener_roots ... FAILED
panicked at crates/perry-ext-net/src/tests.rs:59:5
test result: FAILED. 27 passed; 1 failed
```

Line 59 is:

```rust
fn assert_rewritten(before: i64, after: i64) {
    assert_ne!(after, before);
    assert!(perry_runtime::arena::pointer_in_nursery(after as usize));
}
```

So the test requires a GC root to have **moved** across a collection, and it is the `assert_ne!` that fires — the object did not move. That is the "fixture allocated but nothing was evacuated" shape: the test cannot distinguish "the mutable scanner failed to rewrite the root" from "no evacuating collection ran at all", and right now it is failing for one of those two reasons.

Two possibilities worth separating:

1. the scanner genuinely stopped rewriting listener roots — a real GC bug;
2. the fixture no longer drives an evacuating minor in this environment, so `after == before` because nothing moved.

The second is the more likely reading given the test is asserting *movement* rather than correctness of a value, and it is the same class as other fixtures here that allocate without churning. Either way the test should assert its subject ran (e.g. that a copying minor actually occurred / `copied_objects > 0`) before asserting the address changed — otherwise a green run does not mean the scanner works, and a red one does not mean it is broken.

Reproduce:

```bash
RUST_TEST_THREADS=1 cargo test --profile perry-dev -p perry-ext-net --lib
```

Found while auditing #8632; confirmed not caused by it (identical failure on clean `main`).

proggeramlug:
Sharper diagnosis — it is **not** random flakiness, it deterministically depends on the collector mode:

```
default                      test result: FAILED. 27 passed; 1 failed
PERRY_GC_FORCE_EVACUATE=1    test result: ok.     28 passed; 0 failed
```

Reproduced on `perry-ext-net` @ `e727da9e4` + #8637, single-threaded, back to back.

That matches the assertion exactly: `assert_rewritten` requires `assert_ne!(after, before)` — the root must have **moved**. With no evacuating collection the object stays put, `after == before`, and it fails. So the test's precondition is "an evacuating minor ran", and it currently just assumes that rather than arranging or asserting it.

[...]

PR #8637, opened 2026-08-23
fix(http): finish #4975 review correctness gaps

### LINKED ISSUE #4975: node:http/https behavioral parity tail (26.2% → ...) — successor to #2132

proggeramlug:
Follows the landing of #8632 through #8635 and resolves the remaining substantive review findings before closing the HTTP/HTTPS parity-tail ticket.

## What changed

- preserve exact mixed on/once ClientRequest listener ordering and removal semantics
- retain decoder state across streamed base64 and UTF-16 chunks
- make TLS ticket RNG failure non-panicking, enforce sessionTimeout expiry, and root TLS/keylog callbacks across allocations
- handle zero-byte writes as WriteZero and finish incomplete response heads safely during flush/shutdown
- complete net listener-limit dispatch and IncomingMessage manifest coverage
- make TLS cache invalidation port-aware and outline-report builds non-cacheable
- strengthen runtime-scope rooting analysis and fix the newly exposed event helper roots
- normalize HTTP-date validation without locale-dependent parsing

## Validation

- 53/53 repository lint gates passed
- affected crates pass cargo check --all-targets with -D warnings
- perry-ext-http: 87 passed
- perry-ext-net: 28 passed with forced evacuation
- perry-api-manifest: 38 unit and 4 integration tests passed
- perry-stdlib: 120 passed
- Node core subset script tests: 3 passed
- formatting, file-size, diff, GC holder, and unrooted-local gates pass

No version bump or changelog fragment.

Closes #4975

## Summary by CodeRabbit

* **New Features**
  * Added socket `getMaxListeners` and `setMaxListeners` support.
  * Added configurable HTTPS session-ticket lifetimes and safer ticket handling.
  * Added HTTP streaming support for split Base64 and UTF-16 response data.
  * Added `http.IncomingMessage` socket and connection accessors.
* **Bug Fixes**
  * Improved HTTP connection reuse, queued-request handling, and idle-socket cleanup.
  * Fixed event listener ordering and one-time listener behavior.
  * Improved response flushing and timeout overflow warnings.
  * Improved TLS callback reliability and streamed response delivery.

### DIFF
--- a/crates/perry-api-manifest/src/entries/part_4.rs
+++ b/crates/perry-api-manifest/src/entries/part_4.rs
@@ -838,6 +838,8 @@ pub(crate) const API_MANIFEST_PART_4: &[ApiEntry] = &[
     method("http", "setEncoding", true, Some("IncomingMessage")),
     method("http", "setTimeout", true, Some("IncomingMessage")),
     method("http", "__get_req", true, Some("IncomingMessage")),
+    method("http", "__get_socket", true, Some("IncomingMessage")),
+    method("http", "__get_connection", true, Some("IncomingMessage")),
     method("http", "socket", true, Some("IncomingMessage")),
     method("http", "connection", true, Some("IncomingMessage")),
     method("http", "req", true, Some("IncomingMessage")),
--- a/crates/perry-ext-http/src/agent.rs
+++ b/crates/perry-ext-http/src/agent.rs
@@ -147,7 +147,9 @@ pub struct AgentHandle {
     /// reqwest intentionally hides them; this mirror preserves the cache and
     /// eviction semantics without exposing backend internals.
     pub max_cached_sessions: usize,
-    pub tls_sessions: HashMap<String, u64>,
+    /// Opaque session id and the server port it belongs to. Keeping the port
+    /// structurally avoids guessing inside the colon-delimited Agent key.
+    pub tls_sessions: HashMap<String, (u64, u16)>,
     pub tls_session_order: VecDeque<String>,
     pub next_tls_session: u64,
     /// HTTPS options supplied to the Agent constructor. Request-local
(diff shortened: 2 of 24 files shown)


proggeramlug:
Landed on `main` in 00bddb34b via #8641, with the changelog fragment added (contributors do write their own here — the only thing that was missing).

Validated at head `9044c091e`: all nine ratchets + fmt, `cargo check --workspace --all-targets` clean, `perry-ext-http` 87/0 and `perry-stdlib` 120/0 — both matching your reported numbers exactly.

Two things I checked closely and want to credit:

**The gate strengthening is real, and I verified it rather than assuming.** `RuntimeHandleScope` was a blanket `ROOTED_MARKER`, so any function merely mentioning it escaped the whole analysis. Making it lexically scoped is a genuine tightening, and pinning it with `planted_after_runtime_scope` (required) plus `clean_active_runtime_scope` (forbidden) is the right pair — the old script structurally could not have caught that probe. I did check that `lexical_depth` is actually maintained, since a never-updated depth would have made the scoping cosmetic; it is, at line 240.

**The `root_scanner` aliasing fix has the ordering right.** Capturing `shared_wrapper` before visiting `callback` is what makes it sound — computed after, the comparison fails post-relocation and `raw_wrapper` gets visited as an already-moved-from address. Easy one to get backwards.

On `perry-ext-net`: your "28 passed with forced evacuation" is consistent with what I see — 27+1 by default, 28+0 under `PERRY_GC_FORCE_EVACUATE=1`. That is #8636, pre-existing and not yours; I have updated it with the deterministic repro.

PR #8642, opened 2026-08-23
perf(codegen): validate boxed callback captures once

proggeramlug:
## Summary

- emit a bounded private closure body for small direct arrow-literal callbacks whose mutable captures are compiler-owned boxes
- extend the existing arrow-function registry with an optional private target, exact capture count, and boxed-slot mask
- validate identity, direct-call ABI, capture count, and every live boxed slot once at method entry; any mismatch keeps the checked public closure body
- centralize capture ordering across closure creation, body compilation, and layout registration

The public/dynamic body still uses `js_box_get_bits` / `js_box_set_bits`. The private body preserves TDZ behavior, performs the raw boxed access only after admission, and immediately uses the existing write barrier after a trusted store.

Admission is deliberately narrow: at most 16 clones per module, 64 HIR nodes, 16 fixed parameters, and 64 capture slots. Async/generator/dynamic-`this`, typed-clone, default/rest, mapped-`arguments`, module-global, and indirect stored-closure shapes are excluded. Registration can only enrich an existing arrow entry and cannot promote an ordinary function.

## Qualification

Frozen comparison parent: `9963215969daf716c8e6fd344731ea2dee66fa79`.

Correctness and safety gates:

- release `perry-codegen`: 1,185 passed, 1 ignored
- release `perry-runtime`: 2,638 passed, 4 ignored
- object cache: 47 passed
- native link-cache regression: passed
- source-graph suite: 15 passed before the final non-overlapping upstream rebase; the cross-module allocating callback fixture was rerun on the frozen parent in normal and forced-evacuation modes
- imported upstream ECS baseline: 35/37 files and 461/463 cases, matching the existing `singleton` shorthand and `Symbol.dispose` gaps
- all 106 candidate runtime-symbol sentinels
- GC dominance allocation/poll/immovable audits and self-tests
- runtime-holder census plus 89 planted fixtures
- thread-local policy audit and self-test
- formatting and `git diff --check`

The recovered committed worktree also reran `cargo check` for both modified crates, all five focused codegen tests, and both focused runtime tests.

## Performance

The exact release executables were 15,775,992 bytes baseline and 15,792,536 bytes candidate: +16,544 bytes (+0.105%). Both passed the shim and exact 50,005,000 accumulation oracle.

On an arm64 Mac mini, the unchanged quiet-host gate admitted at 6.9% active CPU. Eleven alternating repeat-256 pairs per kernel produced:

| kernel | baseline median | candidate median | median paired reduction | wins |
| 10k read-only query | 0.353663 ms/op | 0.307604 ms/op | 12.94% | 11/11 |
| 10k accumulation query | 0.306683 ms/op | 0.263721 ms/op | 13.69% | 11/11 |

All 44 processes passed their correctness oracles. These are exact-parent incremental results; I am not comparing them directly with Node medians collected on a different host.

A symbolized profile captured after the cohort shows the hot `Archetype.forEachWithComponents` path selecting `__17$trusted_boxes`. Across 5,918 main-thread samples, `is_registered_box_ptr` has no leaf-table samples and no occurrence beneath that trusted callback. Its only two call-graph samples (0.034%) are closure setup and a separate public checked closure.

## Summary by CodeRabbit

* **Performance**
  * Optimized eligible direct arrow callbacks for faster execution.
  * Benchmarks show improvements of approximately 12.9% for read-only iteration and 13.7% for accumulation.
* **Reliability**
  * Added validation and safe fallback behavior for unsupported callback patterns.
  * Preserved correct handling of captures, TDZ behavior, and indirect callbacks.
* **Testing**
  * Added coverage for callback eligibility, capture layouts, size limits, and cross-module execution.


proggeramlug:
Follow-up qualification after the incremental A/B:

- Same-host Mac-mini Node v26.5.1 comparison, 11 alternating repeat-256 pairs: read-only Node/Perry medians 0.052144/0.305723 ms (5.859x median paired ratio), accumulation 0.088810/0.271932 ms (3.062x). All 44 process oracles passed. This change is a measured increment, not parity.
- The read-only timed closure includes `expect(count).toBe(...)`; that assertion is 2,054/5,900 workload samples (34.8%) and owns the apparent layout-filter hotspot. A separate assertion-free accumulation capture attributes 61.0% to `Archetype.forEachWithComponents`, 17.3% to the trusted callback, 9.1% to dynamic addition, and 5.1% to filter truthiness. Disassembly maps the dominant method bucket to repeated exact method-shape and guarded-array proofs.
- Rebased onto current main `905017b1c` after merge-base CI drift. The feature and changelog stable patch IDs are unchanged (`e391e984...` and `c7c81273...`). On the rebased tree: compiler/runtime check passed, focused codegen 5/5 and runtime 2/2 passed, formatting/diff/generated-doc checks passed. The remaining local TLS-policy failure reproduces on untouched current main (`node_submodules/test_runner.rs` contains the raw declaration).

proggeramlug:
Landed on `main` in f96a9d9e0 via #8644, **unchanged** — nothing needed fixing.

Validated at head `19e254b19`: eight ratchets + fmt, `cargo check --workspace --all-targets` clean, `perry-codegen` 1189/0 with the five trusted-box tests asserted by name, `perry-runtime` 2641/0.

On the unsafe fast path — this is well built. Naming the helper `js_box_set_bits_trusted_no_barrier` makes the obligation impossible to miss at a call site, the safety doc states the shading requirement directly, and the IR test pins that `js_write_barrier` co-occurs with the trusted store while the public body never uses it. Four of your five tests cover what must *not* be admitted, which is the emphasis that matters for an admission guard.

One thing to flag on measurement: the sweep I ran today covers `905017b1c`, i.e. `main` **without** this PR, so your perf claim is un-reproduced on my side. That was deliberate — it leaves a clean baseline to measure this against rather than conflating the two. Happy to run the comparison against that baseline when the bench host is free; it is currently contended by another benchmark harness.

PR #8646, opened 2026-08-23
fix(runtime): complete computed property name reflection

proggeramlug:
## Summary

- complete the `language/computed-property-names` Test262 subcluster
- preserve `NaN` and `±Infinity` as ordinary computed property keys
- expose computed symbol methods through `Object.getOwnPropertySymbols` on materialized class prototypes
- let a computed `["constructor"]` method replace the implicit prototype back-pointer
- add an end-to-end Node/Perry parity regression covering all three seams

Refs #5894

## Validation

- `scripts/test262_subset.py --root vendor/test262 --dir language/computed-property-names --jobs 8 --sample-cap 1000` — 48/48 pass, up from 45/48
- `PERRY_SKIP_BUILD=1 ./run_parity_tests.sh --filter test_issue_5894_computed_property_names` — 1/1 pass
- `cargo test -p perry-runtime --lib` — 2641 passed, 0 failed, 4 ignored

## Summary by CodeRabbit

* **Bug Fixes**
  * Computed class methods named `constructor` are now correctly installed and can override the default constructor property.
  * Class prototype symbols are now consistently included in symbol property enumeration without duplicates.
  * Properties defined with computed keys such as `Infinity`, `-Infinity`, and `NaN` can now be accessed correctly.

* **Tests**
  * Added coverage for computed class methods, symbol ordering, property descriptors, invocation, and numeric computed property keys.

### DIFF
--- a/crates/perry-runtime/src/object/class_registry/parent_static.rs
+++ b/crates/perry-runtime/src/object/class_registry/parent_static.rs
@@ -651,7 +651,7 @@ pub unsafe extern "C" fn js_register_class_computed_method(
                 setters: HashMap::new(),
             });
         vtable.methods.insert(
-            name,
+            name.clone(),
             VTableMethodEntry {
                 func_ptr: func_ptr as usize,
                 param_count: param_count as u32,
@@ -662,6 +662,10 @@ pub unsafe extern "C" fn js_register_class_computed_method(
                 has_rest: has_rest != 0,
             },
         );
+        // Backfill when reflection already materialized `C.prototype`.
+        drop(registry);
+        let proto = class_decl_prototype_object(class_id);
+        super::state::install_class_decl_prototype_method_field(proto, class_id, &name);
     }
     VTABLE_GEN.fetch_add(1, Ordering::Release);
 }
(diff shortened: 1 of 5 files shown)


proggeramlug:
CI follow-up after the amended head (`29949babc`):

- The branch-specific `raw_handle_debt` finding from the first run is fixed; the final lint log has no raw-handle violation.
- Passing on the final head: `cargo-test`, `e2e-scoped`, `warnings`, `gap-suite (5)`, plan, the GC self-test, and CodeRabbit review (no actionable comments).
- The remaining API-doc, addr-class, gap-suite, and GC-stress failures are reproduced by the completed [current-main sweep](https://github.com/PerryTS/perry/actions/runs/32632257737) at base SHA `f96a9d9e0`. The PR uses six gap shards while main used three, but the final failing case lists map to the same base regressions.
- The TLS self-check is likewise inherited (`crates/perry-runtime/src/node_submodules/test_runner.rs`, untouched here).
- `test_gap_eval_as_value` had one extra intermittent crash in shard 3; an isolated rerun on this exact head passed 1/1 with Node parity.
- The changelog lint remains intentionally unsatisfied because #5894 explicitly requires a code-only PR with no CHANGELOG; the suggested `skip-changelog` label is not defined in the repository.

Ticket validation remains green: `language/computed-property-names` 48/48 (up from 45/48), the focused parity regression 1/1, and runtime unit tests 2,641 passed / 0 failed / 4 ignored.

proggeramlug:
Landed on `main` in 06e1ab349 via #8657, batched with the other four ready PRs. Validation notes there: all nine ratchets, `cargo check --workspace --all-targets` clean, codegen 1189/0, runtime 2641/0. I added the changelog fragment this PR was missing.

PR #8647, opened 2026-08-23
perf(codegen): materialize large constant array literals from a static descriptor + one bulk call (#8583 follow-up)

proggeramlug:
## Problem

A minified bundle data table is a **giant nested constant array literal**. The default lowering builds it procedurally — one `js_array_from_values` per sub-array plus inline element stores — so the Claude Code 2.1.112 bundle's `__33499` (a constant numeric array-of-arrays) lowered to **11,104 allocations and a 245k-instruction body** that made `rewrite-statepoints-for-gc` fan out (the unit-4 blocker; #8633 spills it so it *compiles*, but the runtime still does 11k allocations and the giant body is still emitted).

This is the real fix for that pattern: a constant data table should be **static data + one bulk build**, not thousands of procedural allocations.

## Change

A codegen path (`expr/array_literal.rs`) recognizes a **large, fully-constant** array literal — `number`/`int`/`bool`/`null`/`undefined`, recursively nested arrays — and:

- serializes the constant tree into a compact tagged blob emitted as module-private **rodata**, and
- emits **one** call to a new runtime helper `js_value_from_const_descriptor` (`array/alloc.rs`) that materializes the whole nested structure in a single pass.

Gated on a 256-node minimum, so small literals keep the fast inline bump-alloc path (no regression). Strings/objects decline for now (v2) → the whole literal falls back to the procedural path, never half-materialized.

## Before / after (3,000-row nested-array synthetic)

| | procedural (`PERRY_CONST_ARRAY_DESCRIPTOR=0`) | descriptor (default) |
|---|---|---|
| `js_array_from_values` calls | 3,000+ | **0** |
| construction | 3,000-instr body + 3,000 allocs | 1 rodata blob + **1** call |
| compile time | did not finish in 2 min | **1.46s** |
| output | `rows:3000 v:44` | `rows:3000 v:44` (identical) |

## GC-safety

The recursive builder holds partially-built parent arrays and already-built children (JS heap pointers) across nested allocations. It runs the entire materialization inside a `GcSuppressScope` — the same discipline `js_json_parse` and the lazy-array materializer use — so no collection or relocation can fire mid-build. Element stores go through `store_array_slot` (downgrades a raw-f64 row when a pointer lands in it); all-number rows keep the fast numeric layout. Every descriptor read is bounds-checked (the blob is compiler-generated and trusted, but a malformed one declines to `undefined` rather than reading OOB).

## Mutability

Each call returns a **fresh, mutable** array (JS array literals are mutable — the descriptor is a template, never a shared frozen constant). Verified: mutating one materialization (`a[0].push(99)`) does not touch a second (`b[0]` stays length 2, `a !== b`).

## Verification

- Descriptor path fires: IR has the `js_value_from_const_descriptor` call + rodata blob, **zero** `js_array_from_values` for the table.
- Byte-identical output vs the procedural build across the moving-GC matrix (`PERRY_GC_SCAVENGE_NURSERY_MB=1/2/4`, `PERRY_GEN_GC=0`, default).
- Mutation-after-materialize and mixed bool/null rows (`a:true b:null c:false d:7`) verified.
- Integration test `crates/perry/tests/const_array_descriptor_8583.rs` (IR-fired check + ON/OFF differential across GC arms + mutability/mixed-types).
- `cargo check -p perry-codegen -p perry-runtime` clean (`-D warnings`). `cargo test -p perry-codegen` **pending CI** (this machine is disk-constrained by concurrent builds).

## Kill switch

`PERRY_CONST_ARRAY_DESCRIPTOR=0` reverts every large constant literal to the procedural path (A/B bisection + escape hatch).

Separate from #8633 (the spill backstop, which makes any giant function compilable); this eliminates the cost at the source for the common data-table shape.

## Summary by CodeRabbit

  * Large constant array literals now compile faster and require fewer allocations.
  * Smaller or unsupported array literals continue using the existing behavior.

* **Bug Fixes**
  * Generated arrays remain fresh and mutable on each creation.
  * Improved reliability when constructing large nested arrays during garbage collection.

* **Configuration**
  * The optimization can be disabled with `PERRY_CONST_ARRAY_DESCRIPTOR=0`, `off`, or `false`.


proggeramlug:
Follow-up (pushed b285d7557): narrowed the gate to **nested** constant literals only.

A flat constant scalar array (e.g. `[0; 2050]`) is already a single `js_array_alloc_literal` + inline stores — not the per-subarray fan-out this targets — and its inline path carries the precise per-slot write barriers a later push/store relies on (`large_object_barriers`). The descriptor path now requires at least one nested array element, so only genuine nested data tables (the `__33499` shape) take it; flat arrays keep their existing path.

- `large_object_barriers` (which a flat-array-intercepting earlier revision regressed): **now passes** (3/3).
- The nested 3,000-row synthetic still collapses to one `js_value_from_const_descriptor` call (0 `js_array_from_values`), `rows:3000 v:44`.

Full `cargo test -p perry-codegen` re-run in progress locally; CI is authoritative.

proggeramlug:

PR #8651, opened 2026-08-23
fix(runtime): keep old array growth targets out of nursery

proggeramlug:
## Summary

- keep array-growth forwarding targets out of the copying nursery when the source is old or otherwise non-moving
- use a no-collection nursery allocation for young sources, falling back to an old-born target if allocation could collect and promote the source
- add a deterministic regression test for the old forwarding-stub generation invariant

## Root cause

`js_array_grow` always allocated the replacement backing array through the ordinary GC allocator. An old retained `GC_FLAG_FORWARDED` growth stub could therefore point into resetting nursery space. Copying minor GC does not trace that stub payload as a normal array slot, so the nursery target could be recycled while stale aliases still followed the old forwarding word. In the ECS workload this surfaced intermittently as `Cannot assign to read only property length of object`.

## Validation

- regression test fails before the fix (`Nursery` target) and passes after the fix (`Old` target)
- focused release regression on rebased commit: 1 passed, 0 failed
- complete release runtime harness with one test thread: 2,642 passed, 0 failed, 4 ignored
- `python3 scripts/gc_store_site_inventory.py` (1,542 files scanned)
- ECS suite locally: 7/7 passed, checksum `50005000`
- Mac mini ECS qualification using a freshly rebuilt static runtime archive: 50/50 independent processes passed, each with 7/7 tests and checksum `50005000`

The broader `pre-tag-check.sh --quick` run passed the owned gates; its address-classification audit also reports three existing sites that are identical on current `main` and untouched by this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved array growth handling across memory generations.
  * Prevented stale references when growing long-lived arrays and resetting temporary memory.
  * Ensured expanded arrays remain in the appropriate memory area.

* **Tests**
  * Added regression coverage for growing long-lived arrays beyond their capacity.

### DIFF
--- a/changelog.d/8651-array-growth-generation.md
+++ b/changelog.d/8651-array-growth-generation.md
@@ -1,0 +1,12 @@
+Fixed retained array-growth forwarding stubs pointing into resetting nursery
+space. When an old array outgrew its backing storage, `js_array_grow` could
+allocate the replacement in the copying nursery and leave the permanent old
+stub pointing at it. Minor GC does not trace that forwarding payload as a
+normal array slot, so a later nursery reset could recycle the target while
+stale aliases still followed the old forwarding word.
+
+Array growth now keeps the replacement out of the nursery whenever the source
+is old or otherwise non-moving. A young source uses nursery space only through
+the no-collection allocator; if growth could collect and promote the source,
+the replacement is born old instead. This fixes the intermittent ECS failure
+reported as `Cannot assign to read only property 'length' of object`.
--- a/crates/perry-runtime/src/array/push_pop.rs
+++ b/crates/perry-runtime/src/array/push_pop.rs
@@ -1,6 +1,5 @@
 //! push / pop / shift / unshift / set_length / delete + grow primitive.
 use super::*;
-use crate::arena::arena_alloc_gc;
 use std::ptr;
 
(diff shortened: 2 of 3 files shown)


proggeramlug:
CI baseline comparison: the exact base commit `e5b87e134` is already red in [main run 32643736223](https://github.com/PerryTS/perry/actions/runs/32643736223).

The head-only signal is clean:

- `cargo-test`, `warnings`, and `e2e-scoped` pass on this PR.
- `lint` is the same current-main address-class ratchet: the two `bits >= 0x1000` sites and one lone `is_valid_obj_ptr` site in untouched class-construction files.
- `check` is the same current-main generated API-doc drift.
- `gc-stress` fails the same three tests on the base and PR across every arm, including shipped/default with zero cycles: `repsel_p4b_field_store_elision`, `gc_inline_ctor_this_rooting`, and `gc_class_field_receiver_rooting`.
- the base run also has broad gap-shard failures.

No failing check unique to the array-growth change was found. Local exact-commit release runtime validation remains 2,642 passed / 0 failed / 4 ignored, and the corrected ECS artifact passed 50/50 fresh Mac mini processes.

proggeramlug:

PR #8652, opened 2026-08-23
fix(codegen): guard block-creating lowerings against diverged (terminated) blocks (#8583)

proggeramlug:
## Problem

With the #8583 fan-out fix (#8633) in place, the Claude Code 2.1.112 bundle codegens past the old `@main`/`__33499` hang but then fails at unit 25 with:

```
native codegen unit 26/84 failed: unit 25: register %r144 was used but never defined
```

This is a **pre-existing codegen soundness bug**, previously masked because the compile never got past the unit-4 hang.

## Root cause

When a sub-expression **provably diverges** — a throwing operand (a captured TDZ access / const-reassignment) emits `js_throw_error_with_code` + `unreachable` — the current block is **terminated**. `LlBlock` silently drops any instruction emitted after a terminator (block.rs), so the *setup* registers for the surrounding operation are discarded. But block-creating lowerings still emit fresh blocks that **reference those dropped `%rN` registers**, and the dialect builder rejects the module ("register %rN used but never defined", dialect/mod.rs `finish()`). The surrounding operation is unreachable on that path, so the correct behavior is to emit nothing once the block is terminated.

Two sites hit this in the bundle (both dead code after a proven-throwing operand):
- `lower_index_set_fast` (`a[i] = v`, closure `__44845`, undefined `%r142/%r144/%r145`)
- `emit_persistent_shadow_root_barrier` (a pointer root store, closure `__44449`, undefined `%r102`)

## Fix

Each site returns early when `ctx.block().is_terminated()` — the sound, minimal guard (no poison-masking, which would hide genuine miscompiles). Also adds an env-gated `PERRY_DIALECT_DUMP=<dir>` diagnostic: on a dialect construction failure, `render_units_from_frozen` names the offending function and dumps its full IR (typed insts via `render_into`). The failing unit never parses, so `PERRY_SAVE_LL` (post-parse) can't capture it — this diagnostic is how the two sites were found, and it makes the whole class diagnosable in future. Zero cost when the env var is unset.

## Validation

End-to-end: with these guards the cli.js bundle codegens **all 84 units with zero "used but never defined" errors** (it previously failed at unit 25). The only remaining blocker to a final binary is unrelated host disk pressure (a shared-machine ENOSPC), not codegen. A unit repro of the exact diverged-operand shape is bundle-specific; happy to add one if preferred. `cargo test -p perry-codegen` pending-CI (local runs are disk-constrained by concurrent builds).

Stacks conceptually on #8633 (both under #8583) but touches disjoint files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed code generation failures on unreachable paths caused by divergent expressions.
  - Prevented invalid references to discarded values when processing indexed assignments and persistent memory barriers.

- **Diagnostics**
  - Improved error messages to identify the function associated with code-generation failures.
  - Added an optional `PERRY_DIALECT_DUMP` diagnostic to capture intermediate output for troubleshooting.

- **Documentation**
  - Documented the fix for divergent-block guard handling.


proggeramlug:

PR #8656, opened 2026-08-23
fix(intl): expose Collator compare as an accessor

proggeramlug:
## Summary
- expose `Intl.Collator.prototype.compare` through the ECMA-402 accessor shape
- return one stable anonymous `[[BoundCompare]]` function per Collator instance
- cover the descriptor, identity, branding, name, and length contract in the gap suite

## Validation
- Test262 `intl402/Collator/prototype/compare`: 9 pass, 0 diff, 0 compile failures; the remaining 3 failures are the existing non-normative locale-ordering cases
- `test_gap_intl_collator_compare_accessor`: parity pass (1/1)
- `cargo check -p perry-runtime --no-default-features --features intl-namespace,string-normalize,alloc-mimalloc`
- `cargo fmt -p perry-runtime --check`
- `git diff --check`

Refs #5896

### DIFF
--- a/crates/perry-runtime/src/intl.rs
+++ b/crates/perry-runtime/src/intl.rs
@@ -54,13 +54,14 @@ mod segmenter;
 use canon_aliases::canonicalize_unicode_extension_types;
 
 pub(crate) use date_collator::{
-    collator_bound_compare_thunk, collator_bound_resolved_options_thunk, collator_compare_thunk,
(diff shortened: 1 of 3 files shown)


proggeramlug:
Landed on `main` in f5739b532 via #8677. Validation there: all nine ratchets, `cargo check --workspace --all-targets` clean, runtime 2648/0, codegen 1193/0.

PR #8666, opened 2026-08-23
fix(codegen): share imported static update storage

### LINKED ISSUE #8654: class static post-increment through imported base yields NaN and corrupts perform-ecs
## Summary

An imported base class's public static field is not the same initialized cell when it is read and post-incremented from a function in the defining module. `Component._id++` reads as `undefined`, assigns `NaN` to both derived constructors, and leaves the externally visible `Component._id` at `0`.

This is a correctness bug with a catastrophic performance consequence in `perform-ecs`: component bitmasks collide, destroyed entities remain in a component view, and the nominally linear destroy benchmark becomes a growing-array workload.

Reproduced on current `main` (`c2da03439e6e848cfed73163c131acb601ae96dc`, Perry 0.5.1519) and on PR #8645 (`0ab2fb3909d8ad503b526acb23b2695f2d825bcb`). PR #8645 fixes the separate Array-subclass constructor failure but does not fix this case.

## Self-contained repro

`base.js`:

```js
export class Component {
  static _id = 0;
}

export function makeComponent(constructor) {
  constructor.id = Component._id++;
}
```

`main.js`:

```js
import { Component, makeComponent } from "./base.js";

class Position extends Component {}
class Velocity extends Component {}

makeComponent(Position);
makeComponent(Velocity);

console.log(JSON.stringify({
  nextId: String(Component._id),
  positionId: String(Position.id),
  velocityId: String(Velocity.id),
}));
```

Build and run from a Perry checkout:

```bash
node main.js
PERRY_NO_AUTO_OPTIMIZE=1 \
target/release/perry compile main.js -o main-perry
./main-perry
```

Observed:

```text
Node:  {"nextId":"2","positionId":"0","velocityId":"1"}
Perry: {"nextId":"0","positionId":"NaN","velocityId":"NaN"}
```

The `perform-ecs` integration below also reproduces through the default auto-optimize path.

## Public-package impact

`perform-ecs/src/Component.ts` uses the same pattern:

```ts
export abstract class Component {
  static readonly _id: number = 0;
  static readonly id: number;
}

export function makeComponent(constructor: any) {
  constructor.id = (<any>Component)._id++;
}
```

In `ddmills/js-ecs-benchmarks`' `perform-ecs / Destroy` case:

| State after destroy | Node | Perry |
|---|---:|---:|
| component IDs | `[0, 1]` | `[NaN, NaN]` |
| retained view entries after 100 destroys | 0 | 200 |
| retained view entries after 100,000 destroys | 0 | 200,000 |

At full scale on an M1 Mac mini, Perry took 169.70 s and peaked at 256,832 KiB RSS. Block time increased from 1.95 s for the first 10,000 destroys to 31.07 s for the last 50,000, consistent with the ever-growing retained array. This timing must not be treated as a valid performance comparison until the semantic bug is fixed.

## Expected behavior

- A public static field has one canonical initialized storage cell across its defining module and compiled consumers.
- `Component._id++` returns the old numeric value and stores the incremented numeric value.
- The repro prints the same values as Node.
- `perform-ecs` assigns distinct component IDs and retains zero view entries after destruction.

## Acceptance criteria

- Add a two-module native regression based on the repro above and run it in both prebuilt-runtime and auto-optimize modes.
- Cover post-increment from the defining module, post-increment from an importing module, and reads from both modules.
- Include a forced-GC variant so moving/registration does not split or lose the static cell.
- Add an ecosystem regression using the `perform-ecs` component-registration shape: IDs `[0, 1]`, distinct masks, and zero retained entities after add/destroy.
- Preserve the class-semantics coverage in #5893/#8645.

## Likely area

The symptom points to static-field metadata/storage identity across module lowering: the defining-module function appears not to resolve the same initialized static cell that `main.js` observes. Relevant areas include `static_field_meta`, static initialization lowering, shared mutable capture, and class registration/parent-static storage.

proggeramlug:
## Summary
- route known class static post-increment/decrement through the canonical cross-module LLVM global
- mirror the updated value into the dynamic class-property table so static and dynamic reads stay coherent
- keep postfix BigInt results rooted across allocating update work
- add the two-module perform-ecs regression in prebuilt and auto-optimized modes, with normal and forced-evacuation runs

Fixes #8654.

## Tests
- cargo check -p perry-codegen --target x86_64-pc-windows-msvc
- cargo test -p perry-codegen --target x86_64-pc-windows-msvc the_result_is_rooted_only_when_the_element_may_be_a_bigint -- --nocapture
- cargo test -p perry --target x86_64-pc-windows-msvc --test issue_8654_imported_static_field_cell -- --nocapture

No version bump.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed increment and decrement operations on imported class static fields.
  - Corrected prefix and postfix results, including BigInt values.
  - Ensured updated static fields remain consistent after garbage collection.
  - Improved support for dynamic access and component-related static field usage across modules.

- **Tests**
  - Added comprehensive coverage for static field updates across defining and importing modules, including automatic and prebuilt runtime configurations.

### DIFF
--- a/crates/perry-codegen/src/expr/member_update.rs
+++ b/crates/perry-codegen/src/expr/member_update.rs
@@ -43,7 +43,7 @@ use crate::nanbox::POINTER_MASK_I64;
 use crate::rooting::{self, Repr};
 use crate::types::{DOUBLE, I32, I64, I8};
 
-use super::{lower_expr, FnCtx};
+use super::{emit_root_nanbox_store_on_block, lower_expr, FnCtx};
 
 pub(crate) fn lower(ctx: &mut FnCtx<'_>, expr: &Expr) -> Result<String> {
     match expr {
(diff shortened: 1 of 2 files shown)


proggeramlug:
Landed on `main` in f5739b532 via #8677. Validation there: all nine ratchets, `cargo check --workspace --all-targets` clean, runtime 2648/0, codegen 1193/0.

PR #8672, opened 2026-08-23
perf(runtime): restore method-scoped prototype guards

proggeramlug:
## Summary

- Restore method-name-scoped invalidation for direct method guards after the
  later class-semantics integration reinstated the process-global latch.
- Keep assignment, descriptor installation, and deletion fail-closed while
  allowing unrelated prototype keys to leave hot method guards usable.
- Make generic inherited dispatch observe declared-method replacement,
  deletion, and re-creation correctly.

The all-method latch remains as an escape hatch for any future keyless
prototype mutation. Keyed writes use the low 16 bits of the existing FNV-1a
dispatch hash; collisions conservatively retire additional method names.

## Root cause

The original scoped implementation in `2897cb521` was superseded when the
class-semantics work was integrated, restoring a sticky process-wide
invalidation byte. The unchanged `codehz/ecs` comprehensive suite performs
prototype activity before its query benchmarks, so every later direct method
guard failed and each entity call entered generic dispatch.

The performance and composition artifacts below were rebuilt with the fix as
one commit on `main` at `60f4fabde`. The final PR head is rebased on
`850e6f1d5`; that intervening async/RegExp batch does not overlap this patch.

## Mac mini benchmark

Apple M1 Mac mini, macOS 26.5.1, AC power, `taskpolicy -t 0 -l 0`.
The frozen source is upstream `codehz/ecs` commit
`58d729682eeb88d7361796de0420f3d673e27479`, benchmark
`10k entities: forEach query with accumulation`, repeat 256, two warmups and
six measured rounds per process.

An exact-parent 11-pair alternating cohort at `8224d879a` measured:

| arm | median ms/op | MAD |
|---|---:|---:|
| exact parent | 7.268065 | 0.008355 |
| scoped guard | 0.635431 | 0.000364 |

Median paired improvement was **91.26%**, with 11/11 candidate wins and 22/22
process semantic oracles. The admitted 60-s quiet gate stayed between 4.10%
and 16.60% active CPU. Executable SHA-256 values were
`f6532a88ad6635cdbd8cfb72918b9587a9c85c907aaa72a7c44cfbb5d5262614`
(control) and
`2aa1de3236aa37958c914353f737da3b26f7f803d09aae5b2b36b6734c6e4477`
(candidate).

After rebasing, a fresh current-main candidate was rebuilt and hashed as
`52b96f3d8f22fa391310ea889b88d51a17a2c52765d5696574aa707916bc54c2`.
A quiet three-pair Mac screen against Node 26.5.1 measured 0.615689 ms/op for
Perry and 0.109858 ms/op for Node, with all six process oracles passing.
Perry therefore remains **5.60x slower** in this workload; this PR is a large
regression repair, not a parity claim.

## Validation

- `cargo check -p perry-codegen -p perry-runtime`
- Focused runtime tests for exact compiler/runtime slot pairing, descriptor-key
  isolation, same-name invalidation, and generic replacement fallback
- Focused codegen IR test pinning both global and per-method guard loads
- New Node/Perry fixture covering unrelated writes, inherited replacement,
  deletion, and same-name writes on another class; byte-identical normally and
  under forced verified-evacuation settings on the rebased compiler
- Frozen ECS suite: 7/7 tests and exact `50,005,000` checksum locally and on
  the Mac mini with the rebased compiler
- `git diff --check`

The full ECS executable still exposes the existing forced-evacuation verifier
panic. The exact parent reproduces the same seventh-copying-minor failure and
signature under the identical command; the focused fixture is green under
those flags. This PR does not claim to fix that separate baseline GC defect.

## Summary by CodeRabbit

* **Performance**
  * Improved method-call performance by invalidating optimized lookups only when the relevant prototype method changes.
  * Preserved safe fallback behavior for prototype replacements and hash collisions.

* **Bug Fixes**
  * Fixed method resolution after prototype methods are deleted or replaced.
  * Ensured deleted methods are not invoked through inherited or registered dispatch paths.

* **Tests**
  * Added coverage for inherited methods, deletions, unrelated prototype changes, and same-name replacements across inheritance chains.

### DIFF
--- a/changelog.d/8672-method-name-prototype-guards.md
+++ b/changelog.d/8672-method-name-prototype-guards.md
@@ -1,0 +1,9 @@
+---
+category: Performance
+title: Restore method-scoped prototype guards
+---
+
+Prototype mutation now invalidates direct-call guards by method-name slot
+instead of permanently disabling every method guard in the process. Hash
+collisions remain conservative, and dynamic prototype replacement retains a
+global fail-closed escape hatch.
--- a/crates/perry-codegen/src/collectors/proven_this_routing_tests.rs
+++ b/crates/perry-codegen/src/collectors/proven_this_routing_tests.rs
@@ -692,11 +692,11 @@ fn guarded_pshape_call_site_is_preceded_by_a_shape_id_guard() {
 }
 
 /// The single-pair shape-only arm is small enough to inline at the call site.
-/// Pin the complete safety gate: acquire the prototype-mutation latch, accept
-/// both the boxed-pointer and internal raw-pointer ABIs, reject addresses
-/// outside the target heap range before dereference, reject own descriptors,
-/// then compare the exact class/ShapeId pair. The out-of-line guard must be
-/// absent from this caller.
+/// Pin the complete safety gate: acquire both the all-method escape latch and
+/// the FNV-indexed method-name latch, accept both the boxed-pointer and
+/// internal raw-pointer ABIs, reject addresses outside the target heap range
+/// before dereference, reject own descriptors, then compare the exact
+/// class/ShapeId pair. The out-of-line guard must be absent from this caller.
 #[test]
 fn single_arm_method_shape_guard_is_inlined_with_the_runtime_contract() {
     let ir = emit(&guarded_site_module(), false);
@@ -706,6 +706,12 @@ fn single_arm_method_shape_guard_is_inlined_with_the_runtime_contract() {
             "load atomic i8, ptr @PERRY_CLASS_PROTOTYPE_FAST_GUARDS_INVALIDATED acquire, align 1",
         ),
         "the inline guard must acquire the runtime's release-published sticky latch:\n{probe}"
+    );
+    assert!(
+        probe.contains(
+            "getelementptr i8, ptr @PERRY_CLASS_PROTOTYPE_FAST_GUARDS_INVALIDATED_BY_METHOD",
+        ) && probe.matches("load atomic i8").count() >= 2,
+        "the inline guard must acquire its method-name invalidation byte:\n{probe}"
     );
     assert!(
         !probe.contains("call i32 @js_method_direct_shape_guard("),
--- a/crates/perry-codegen/src/lower_call/method_override.rs
+++ b/crates/perry-codegen/src/lower_call/method_override.rs
@@ -37,15 +37,17 @@ const SHAPE_ID_RANGE_LEN: &str = "1073741824"; // 0x4000_0000
 /// The first block proves that the value is a tagged heap pointer or the raw
 /// object-address form used by internal method ABIs before any dereference.
 /// The second block reproduces the runtime helper's production contract: the
-/// class-prototype invalidation latch is clear, the receiver is a non-forwarded
-/// ordinary object without own descriptors, and its exact `(class_id, ShapeId)`
-/// pair still matches the compiler-published pair. Any failed proof takes the
-/// unchanged dynamic method fallback.
+/// all-method escape latch and this method name's invalidation byte are clear,
+/// the receiver is a non-forwarded ordinary object without own descriptors,
+/// and its exact `(class_id, ShapeId)` pair still matches the
+/// compiler-published pair. Any failed proof takes the unchanged dynamic
+/// method fallback.
 fn emit_inline_direct_method_shape_guard(
     ctx: &mut FnCtx<'_>,
     recv_box: &str,
     expected_class_id: &str,
     expected_shape_id: &str,
+    method_guard_slot: &str,
     fast_label: &str,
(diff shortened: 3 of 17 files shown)


proggeramlug:
Holding this one for a rebase — the rest of your stack (#8674, #8675, #8676) landed via #8682, cherry-picked onto `main` individually since their contents touch disjoint files.

The blocker is a genuine semantic collision, not a mechanical conflict. This PR adds to `class_registry/state.rs`:

```rust
pub(crate) fn is_bound_native_method_closure_value(value: f64) -> bool {
    unsafe {
        bound_native_callable_module_and_method(value)
            .map(|(module, _)| !module.is_empty())
            .unwrap_or(false)
    }
}
```

Meanwhile #8662 landed on `main` (in `f5739b532`) and solved the same problem with a different predicate, at `state.rs:17`:

```rust
/// True when `value` is a bound native-module *constructor* export. Native
/// constructors and ordinary module functions share `BOUND_METHOD_FUNC_PTR`,
/// so the export's explicit constructor metadata must make the distinction.
pub(crate) fn is_bound_native_constructor_closure_value(value: f64) -> bool {
    bound_native_callable_is_constructor_value(value)
}
```

These are **not** the same function under two names. Yours is true for any bound native-module export with a non-empty module name — constructor *or* ordinary method. `main`'s is true only for an actual constructor. `main`'s is strictly narrower, so swapping it in flips behaviour for every bound native *method* export. I didn't want to guess which polarity each of your call sites wants, so I left it to you.

Both comments describe the same goal — making `class X extends obj.method {}` throw the spec-required TypeError instead of silently staying parentless — so I suspect adopting `main`'s function and dropping your definition is right, but that's your call to confirm.

Two smaller mechanical items in the same rebase, both from `class_registry.rs` being split into a `class_registry/` module directory:

1. `function_would_have_own_prototype` and `ordinary_function_prototype_value_for_read` now live in `class_registry/function_prototype.rs`, not `construct.rs`. Importing them via `pub(crate) use construct::{...}` fails with `E0603` (construct's own import is private) — they need their own `use function_prototype::{...}`.
2. `class_unmark_key_deleted` (yours) merges into the `state::{...}` list cleanly.

The only conflicted file is the `class_registry.rs` module root, and it's purely `use`-list unions — everything else in your PR auto-merges. Once rebased I'll re-run the batch validation and land it.

proggeramlug:
Followed up on my earlier comment — I dug into the call site and I think this unblocks cleanly, with **no judgment call needed after all**. I was wrong to leave it open-ended; the evidence settles it.

Your predicate has exactly one call site, `class_registry/parent_static.rs:132`, and the block immediately above it is:

```rust
if let Some((module, method)) = unsafe {
    native_module::bound_native_callable_module_and_method(parent_value)
} {
    if normalize_native_module_alias(&module) == "wasi" && method == "WASI" {
        register_class(class_id, crate::wasi::CLASS_ID_WASI);
    }
    return;                                    // <-- unconditional
}
if is_bound_native_method_closure_value(parent_value) {
    return;                                    // <-- line 132
}
```

Now compare the two candidate predicates. **Both are built on the same underlying query:**

```rust
// yours
bound_native_callable_module_and_method(value)
    .map(|(module, _)| !module.is_empty()).unwrap_or(false)

// main's (#8662)
bound_native_callable_module_and_method(value)
    .is_some_and(|(m, p)| is_native_module_constructor_export(&m, &p))
```

Each can only be `true` when `bound_native_callable_module_and_method` returns `Some` — and the block above already returns unconditionally in exactly that case. **So line 132 is unreachable dead code, and it is unreachable under either predicate.**

That means the conflict has no behavioural content: substituting `is_bound_native_constructor_closure_value` is a no-op, and so is deleting the branch. My earlier concern — that the two predicates have different truth sets — is true in the abstract but cannot be observed here, because neither is reachable.

**Suggested resolution:** drop line 132 and its import entirely, rather than porting either predicate. The comment above it ("Keep the parentless baseline rather than mis-throwing") is already satisfied by the `return` in the `if let Some(...)` block. If you'd rather keep a belt-and-braces guard, use main's `is_bound_native_constructor_closure_value` so there's one predicate in the tree instead of two — but it will never fire.

Worth double-checking my reading before you act on it, since dead-code arguments are easy to get subtly wrong — but if you agree, the `class_registry.rs` `use`-list conflict collapses to a plain union with nothing to reconcile, and I'll land it as soon as it's pushed. The rest of the PR auto-merges against current `main` (`60f4fabde`).

proggeramlug:
Landed on `main` via #8723 (squash `8beca2f29`). Thanks for taking the deletion route — and for sanity-checking the reasoning rather than just applying it.

Validated on the merged result: all 30 lint checkers, runtime 2667/0, codegen 1214/0 plus all integration suites clean, transform 93/0.

PR #8673, opened 2026-08-24
fix(compile): --report-size false positive from std-internal crate names

jdalton:
## Summary

The "duplicate crate instance" finding from [#8579](https://github.com/PerryTS/perry/pull/8579) could misreport a coincidental name collision as a fixable build duplication.

<details>
<summary>Root cause and fix</summary>

Demangling the real symbols behind a suspicious second `gimli` instance on a real compiled binary showed every one was `gimli::read::cfi::{EhFrame, CommonInformationEntry, Augmentation, PartialFrameDescriptionEntry, UnwindSection, ...}` — DWARF exception-handling-frame parsing, the narrow surface `std`'s own unwinder uses. That copy is baked into the prebuilt `std` shipped with the toolchain and was never a resolvable Cargo dependency of this build at all — confirmed via both `cargo tree -i gimli --duplicates` (empty) and `cargo build --unit-graph -Z unstable-options` (exactly one gimli unit), neither of which can see something baked into std's own prebuilt rlib.

A real application dependency on `gimli` reads debug info instead (`read::abbrev`, `read::line`, `read::rnglists`, ...), so the report now classifies a hash as `std`'s internal copy when its symbols carry at least one CFI-specific marker and zero debug-info-specific ones (requiring "at least one" CFI marker, not "all symbols", matters: the CFI build's own shared reader/utility code — `EndianSlice`, `Reader::read_uleb128`, error types — carries no CFI-specific name of its own but compiles alongside it in the same unit). Excluded hashes aren't silently dropped — they're listed under a new "Excluded: std-internal copies" section instead.

Verified end to end: recompiling the same program that previously reported "gimli linked 2 times, 42.7 KiB" now reports zero duplicate crate instances and one excluded std-internal copy (11.9 KiB, 11 symbols).

</details>

## Test plan

- [x] New/updated unit tests for `looks_like_std_internal_backtrace_copy` (real CFI symbols, real debug-info symbols, outside the known crate list, shared reader/utility symbols alongside CFI ones, a single debug-info marker vetoing exclusion).
- [x] `cargo clippy -p perry --bins` — no warnings from `size_report.rs`.
- [x] End-to-end: recompiled a `console.log` program, confirmed the false positive is gone and the excluded finding is accurate.
- [x] `scripts/check_file_size.sh` passes.

## Summary by CodeRabbit

* **New Features**
  * Improved `perry compile --report-size` reporting by identifying and separately listing Rust standard library-internal crate copies.
  * Excluded these internal copies from duplicate-crate warnings to reduce false positives.
  * Added detection safeguards for debug-information symbols.

* **Documentation**
  * Added documentation describing the updated size-report behavior.

* **Chores**
  * Updated the application version to 0.5.1520.


proggeramlug:
Landed on `main` via #8680 (squash `81ddd46b2`), together with the rest of this batch.

Validated on the merged 4-PR result rather than per-branch: 9 ratchet gates + `cargo fmt --check` pass, `perry-codegen --lib` 1194/0, `perry-runtime --lib` 2651/0 at `RUST_TEST_THREADS=1`. The squashed tree was verified byte-identical to the merged tree before push.

Version bump stripped per maintainer policy (the `Cargo.lock` diff was version-only); your `changelog.d/` fragment carried through unchanged. Thanks!

PR #8674, opened 2026-08-24
perf(runtime): compress array forwarding chains

proggeramlug:
## Summary

- compress a validated multi-hop array-growth forwarding chain at its retained head
- keep corrupt/untracked/cyclic chains fail-closed by rewriting only after the resolver reaches a tracked live array
- add a focused regression test and update the existing three-growth forced-minor-GC coverage for the compression contract

## Why

Array growth preserves stale aliases with forwarding stubs. A receiver retained across multiple capacity-crossing grows can therefore point through two or more stubs. Generated indexed-access guards intentionally heal one forwarding edge inline; after that single hop they still see `GC_FLAG_FORWARDED`, reject the array guard, and enter generic indexed lookup on every access.

`clean_arr_ptr` already performs the complete bounded, tracked-array-validated walk. Once that walk succeeds, this change points the original retained stub directly at the live head. Subsequent generated accesses then need only their existing one-hop healing path. This is central forwarding metadata maintenance rather than an ECS- or benchmark-specific shortcut.

This PR is stacked on #8672. Please review/merge that PR first; the runtime and test change is commit `514e0f553`, followed only by its changelog fragment.

## Correctness

- `cargo test -p perry-runtime`: 2,644 passed, 0 failed, 4 ignored
- focused multi-hop compression regression: passed
- existing three-growth + forced minor-GC regression: passed
- end-to-end 10k-element alias/growth fixture matched Node exactly in normal Perry execution and with `PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1`
- all 88 benchmark processes below passed the upstream 7/7 suite oracle and exact 50,005,000 accumulation checksum where applicable

The complete ECS suite under forced evacuation verification still reports the same pre-existing stale heap-field pointer failure on both the #8672 control and this candidate. Normal ECS execution passes, and the focused forced-evacuation fixture plus runtime GC regression pass.

## Performance

Apple M1 Mac mini, AC power, Node 26.5.1. Each comparison followed a 60-sample quiet-host gate (all samples at or below 25% active CPU), used `taskpolicy -t 0 -l 0`, alternated process order, and ran 11 pairs.

Against the unchanged #8672 binary (`repeat=64`):

| codehz/ecs kernel | #8672 median | candidate median | median paired reduction | wins |
| 10k read-only query | 1.79858 ms/op | 0.32485 ms/op | 81.94% | 11/11 |
| 10k accumulation query | 1.15038 ms/op | 0.26945 ms/op | 76.59% | 11/11 |

Directly against Node 26.5.1 (`repeat=256`):

| codehz/ecs kernel | Node median | Perry median | median paired Perry/Node ratio |
| --- | ---: | ---: | ---: |
| 10k read-only query | 0.05228 ms/op | 0.32832 ms/op | 6.282x |
| 10k accumulation query | 0.08646 ms/op | 0.27032 ms/op | 3.127x |

This removes the repeated generic indexed-lookup pathology. It is a large improvement, but it does not claim Node parity; the remaining steady-state generated loop cost is separate follow-up work.

### DIFF
--- a/changelog.d/8674-array-forwarding-path-compression.md
+++ b/changelog.d/8674-array-forwarding-path-compression.md
@@ -1,0 +1,10 @@
+### Performance
+
+- Compress validated multi-hop array-growth forwarding chains at their retained
+  head. Generated indexed-access guards can now heal stale aliases after repeated
+  capacity growth with their existing one-hop path instead of entering generic
+  indexed lookup on every element access. On the full `codehz/ecs` suite, 11
+  alternating Mac mini pairs reduced the 10k read-only query by 81.94% and the
+  accumulation query by 76.59%, with 11/11 wins and every semantic oracle passing.
+  Direct Node comparisons still leave 6.282x and 3.127x gaps respectively, so
+  this change does not claim Node parity.
--- a/crates/perry-runtime/src/array/header.rs
+++ b/crates/perry-runtime/src/array/header.rs
@@ -634,12 +634,16 @@ pub(crate) fn clean_arr_ptr(arr: *const ArrayHeader) -> *const ArrayHeader {
         unsafe { crate::value::addr_class::try_read_tracked_gc_header(cleaned as usize) };
     unsafe {
         let mut steps = 0u32;
+        let mut first_forwarded_header: *mut crate::gc::GcHeader = std::ptr::null_mut();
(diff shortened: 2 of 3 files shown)


proggeramlug:

PR #8675, opened 2026-08-24
perf(codegen): share numeric guards across dynamic add trees

proggeramlug:
## Summary

- recognize dynamic `+` trees with at least three leaves
- lower every leaf once, preserve the original tree associativity, and use one shared all-number guard for a native `fadd` arm
- preserve the original `js_dynamic_string_or_number_add` tree as the cold arm for strings, BigInt, Symbols, and object coercion
- keep two-leaf dynamic additions on their existing direct helper path, where a separate guard diamond does not pay for itself

## Why

Accumulator expressions such as `sum += row.x + row.y` commonly reach codegen with all three leaves boxed as `Any`, especially across callback and interface boundaries. Lowering each `+` node independently then calls the full dynamic addition helper twice even when all runtime values are numbers.

The compiler already has a guarded numeric-tree lowering that evaluates leaves once and reconstructs the source tree in both arms. This change admits fully dynamic trees only when they have three or more leaves, which is the first point where one shared guard can replace multiple dynamic helpers.

This PR is stacked on #8674 (which is stacked on #8672). Please review/merge those first; this PR contains only commit `ed9d86bce` relative to its base, followed by the required changelog fragment.

## Correctness

- focused IR tests prove:
  - a three-leaf dynamic tree has one numeric arm with two `fadd`s and retains two dynamic helpers in the cold arm
  - a two-leaf dynamic add remains unchanged
- `cargo test -p perry-codegen --lib`: 1,191 passed, 0 failed, 1 ignored
- standalone Node/Perry fixture matched exactly for:
  - numeric trees
  - left- and right-associated string-sensitive results
  - BigInt trees and mixed-BigInt TypeErrors
  - Symbol TypeErrors
  - observable `Symbol.toPrimitive` ordering in both associativities
- the standalone fixture also passed with `PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1`
- unchanged full `codehz/ecs` comprehensive suite: 7/7 passed with exact 50,005,000 accumulation checksum
- retained AArch64 assembly has two `fadd`s on the numeric arm and both original dynamic-helper calls on the cold arm

The exact committed compiler emitted the same `__text` SHA-256 as the benchmarked artifact (`b19ad70a31fa27d93f26e50bf1d140e5d577ebc9a0c256b3a9d762855f9fc90a`).

## Performance

Apple M1 Mac mini on AC power. Each cohort followed a 60-sample quiet-host gate with every sample at or below 25% active CPU, used `taskpolicy -t 0 -l 0`, and alternated process order.

Against the exact #8674 code (`repeat=64`), 11 accumulation pairs measured:

| codehz/ecs kernel | #8674 median | candidate median | median paired reduction | wins |
| 10k accumulation query | 0.26989 ms/op | 0.25891 ms/op | 4.11% | 11/11 |

All 22 processes passed the 7/7 suite oracle and exact checksum. A separate three-pair read-only regression check was neutral at the paired median (-0.033%, 1/3 wins); this optimization does not target that callback.

This is a measured incremental improvement, not a claim of Node parity. The remaining gap is dominated by the guarded ECS loop body and indirect boxed callback boundary.

### DIFF
--- a/changelog.d/8675-dynamic-add-tree-guard.md
+++ b/changelog.d/8675-dynamic-add-tree-guard.md
@@ -1,0 +1,9 @@
+### Performance
+
+- Lower dynamic `+` trees with three or more leaves behind one shared numeric
+  guard. Number-only executions use native additions, while the cold arm keeps
+  the original tree, evaluation order, associativity, and complete dynamic
+  string/BigInt/Symbol/object-coercion behavior. On the `codehz/ecs` 10k
+  accumulation query, 11 alternating Mac mini pairs reduced median time by
+  4.11%, with 11/11 wins and every semantic oracle passing. The read-only
+  query was neutral, and this change does not claim Node parity.
--- a/crates/perry-codegen/src/expr/binary.rs
+++ b/crates/perry-codegen/src/expr/binary.rs
@@ -303,6 +303,24 @@ fn add_tree_leaves<'a>(expr: &'a Expr, out: &mut Vec<&'a Expr>) {
     }
 }
 
+/// Whether a fully-dynamic `+` tree is worth one shared numeric guard.
+///
+/// A three-or-more-leaf tree is the common accumulator shape
+/// `sum += row.x + row.y`: lowering each node independently otherwise pays
+/// the dynamic add helper twice even when every runtime value is a number. At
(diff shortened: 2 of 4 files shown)


proggeramlug:

PR #8676, opened 2026-08-24
perf(codegen): pack monomorphic method shape guards

proggeramlug:
## Summary

- collapse the monomorphic direct-method GC header checks into one masked i32 load
- compare adjacent class and ShapeId fields as one packed i64 word
- retain both prototype invalidation latches, pointer-band validation, ShapeId-domain validation, and the unchanged generic fallback
- add anti-drift tests for the packed mask, supported little-endian layout, and emitted guard contract

This is a general codegen optimization. It does not recognize ECS names or source locations.

## Why

The unchanged codehz/ecs query loop invokes the same guarded helper twice per entity. The existing inline guard loaded GC type, forwarding flags, descriptor flags, class id, and ShapeId separately on every invocation. These fields are contiguous in the runtime ABI, so the same proof can be expressed with two receiver loads while retaining every semantic rejection condition.

## Correctness

- perry-codegen library suite: 1,193 passed, 0 failed, 1 ignored
- focused packed-mask, endian-layout, and emitted-IR tests pass
- unchanged method-invalidation fixture matches in normal and PERRY_GC_FORCE_EVACUATE=1 runs
- unchanged upstream ECS fixture: 7/7 tests and exact 50,005,000 checksum in normal and forced-evacuation runs
- the package-wide typed_feedback integration assertion also fails on the exact parent; it is unrelated to this diff

## Mac mini A/B

Host: Apple M1 Mac mini, macOS 26.5.1, AC power, taskpolicy -t 0 -l 0. Each cohort passed a 60-sample quiet gate before measurement. Control is the exact current stacked parent from PR #8675. Each process used 2 warmups, 6 measured rounds, repeat 64, alternating order, and checked the upstream shim plus checksum.

| workload | control median | candidate median | median paired improvement | wins | oracles |
| --- | ---: | ---: | ---: | ---: | ---: |
| 10k forEach accumulation | 0.259273 ms | 0.249100 ms | 3.957% | 11/11 | 22/22 |
| 10k forEach read-only | 0.326225 ms | 0.315803 ms | 3.269% | 11/11 | 22/22 |

Candidate ECS binary SHA-256: 0cbd1f8ad658e83b9d523b257da53254d7f964fcc354b27ddde99b118ce60ab6.

This is a repeatable improvement, not a parity claim. The latest host-matched Node measurements still leave substantial work after this patch.

### DIFF
--- a/changelog.d/8676-packed-method-shape-guard.md
+++ b/changelog.d/8676-packed-method-shape-guard.md
@@ -1,0 +1,9 @@
+---
+category: Performance
+title: Pack monomorphic method shape guards
+---
+
+Monomorphic direct-method guards now validate the contiguous GC, class, and
+ShapeId header fields with two packed loads while retaining prototype
+invalidation, pointer-range checks, ShapeId validation, and the generic
+fallback. This reduces repeated dispatch proof work in hot method-call loops.
--- a/crates/perry-codegen/src/collectors/proven_this_routing_tests.rs
+++ b/crates/perry-codegen/src/collectors/proven_this_routing_tests.rs
@@ -737,17 +737,21 @@ fn single_arm_method_shape_guard_is_inlined_with_the_runtime_contract() {
         probe.contains("method_direct.inline_deref")
             && probe.contains("getelementptr i8, ptr")
             && probe.contains("i64 -8")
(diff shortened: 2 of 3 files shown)


proggeramlug:

PR #8678, opened 2026-08-24
perf(codegen): count property/index stores as GC safepoints in the spill estimate (#8583)

proggeramlug:
Extends the literal-counting spill estimate (`count_safepoint_sites`, #8583). A property/index **store** (`PropertySet`, `PropertyUpdate`, `IndexSet`) lowers to a collecting runtime call (`js_class_field_set_ic` / `js_set_property` / the array-set helpers) that `rewrite-statepoints-for-gc` gives a statepoint — but none were counted.

A closed-shape object literal compiles to a constructor that is one long run of `this.field = v` stores. With the stores uncounted, the constructor's estimate was ~0, it was never spilled to the shadow frame, and RS4GC grew one `__AnonShape_*_constructor` in the Claude Code 2.1.112 bundle from **34,009 → 2,280,128 instructions**, overrunning the #8586 per-function budget and refusing the whole module (it was the last blocker to a working `cc` binary).

Reads (`PropertyGet`/`IndexGet`) are deliberately not counted — they frequently inline to a shape-cached load with no call, so counting them would over-spill read-heavy hot loops. Unit test added (`property_and_index_stores_are_safepoints`); `cargo test -p perry-codegen --lib safepoint_sites` green.

Note: this improves estimate *accuracy*. Fully spilling that constructor at the default threshold also needs the spill threshold reconciled with the #8586 budget (tracked separately — #8623's 32M sits above the 1.57M budget); see follow-up issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved garbage-collection safety for code containing property and array updates.
  * Fixed undercounting of safepoints in constructors with many field assignments, preventing instruction-budget failures during compilation.
  * Added coverage for property stores, index stores, and update expressions.
* **Documentation**
  * Documented the updated safepoint counting behavior and clarified that property and index reads remain unaffected.


proggeramlug:
Landed on `main` via #8680 (squash `81ddd46b2`), together with the rest of this batch.

Validated on the merged 4-PR result rather than per-branch: 9 ratchet gates + `cargo fmt --check` pass, `perry-codegen --lib` 1194/0, `perry-runtime --lib` 2651/0 at `RUST_TEST_THREADS=1`. The squashed tree was verified byte-identical to the merged tree before push.

Version bump stripped per maintainer policy (the `Cargo.lock` diff was version-only); your `changelog.d/` fragment carried through unchanged. Thanks!

PR #8684, opened 2026-08-24
chore(ci): inline dtolnay/rust-toolchain, SHA-pin every other third-party action

jdalton:
## Summary

Reduces the third-party action trust surface in `.github/`: the most common action is inlined instead of depended on, and every remaining tag-pinned action gets pinned to a commit SHA. Fixes a real, currently-broken reference found along the way.

<details>
<summary>What changed and why</summary>

`dtolnay/rust-toolchain@stable` appeared 45 times across 23 workflow files, and every single call site passed the same `toolchain: 'nightly-2026-08-20'` that `rust-toolchain.toml` at the repo root already pins. Installing a pinned toolchain is one `rustup toolchain install` command, so there is no reason to depend on a third-party action for it. I replaced every occurrence with an inline `run: rustup toolchain install nightly-2026-08-20 --profile minimal` step, translating each site's `components:`/`targets:` inputs into the matching `--component`/`--target` flags so the installed toolchain is identical to what the action produced.

Every other third-party or `actions/*` reference that was still pinned to a mutable tag (`actions/checkout`, `actions/cache` and its `restore`/`save` forms, `actions/setup-node`, `actions/upload-artifact`, `actions/download-artifact`, the three GitHub Pages actions, `ilammy/msvc-dev-cmd`, `mozilla-actions/sccache-action`, `Swatinem/rust-cache`, `oven-sh/setup-bun`, `taiki-e/install-action`) now pins the commit SHA that tag currently resolves to, with the tag kept as a trailing comment for readability. A tag can be silently repointed after review lands; a SHA can't.

While resolving these I found `taiki-e/install-action@v3` does not exist upstream at all — that action has never shipped a `v3`, the latest release is `v2.86.6`. `coverage.yml`'s `cargo-llvm-cov` install step has failed with `startup_failure` on every scheduled run for weeks as a result. Pinned to the real latest release instead of the tag that never existed.

</details>

## Changes

The first commit inlines `dtolnay/rust-toolchain` across 23 workflow files. The second commit SHA-pins every other newly-identified tag reference across 30 files, including the `taiki-e/install-action` fix described above.

## Related issue

n/a

## Test plan

I ran `actionlint` against every touched workflow file before and after each commit and diffed the findings: the only differences are line-number shifts from the shorter inlined blocks, with zero new findings introduced. I confirmed every workflow file still parses as valid YAML after each transform. I resolved each action's SHA directly against the GitHub API (`git/refs/tags/<tag>`) rather than guessing, and spot-checked that the resulting pins match SHAs already in use elsewhere in the same files where that tag already appeared pinned. `cargo build --release` and the workspace test suite are not applicable, since no Rust source changed.

## Checklist

I have not bumped the workspace version or edited CLAUDE.md or CHANGELOG.md. My commits follow the `chore:` prefix convention used in the log. I have read CONTRIBUTING.md and agree to the Code of Conduct.

## Summary by CodeRabbit

* **Chores**
  * Secured CI, release, and deployment automation by pinning workflow actions to immutable versions.
  * Standardized installation of the required Rust nightly toolchain across automated builds and checks.
  * Updated caching and Node.js setup configuration for more consistent workflow execution.
  * Maintained existing build, test, package, release, and documentation deployment behavior.

### DIFF
--- a/.github/workflows/auto-opt-app-patterns.yml
+++ b/.github/workflows/auto-opt-app-patterns.yml
@@ -100,7 +100,7 @@ jobs:
       contents: read
       pull-requests: read
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
           persist-credentials: false
 
@@ -137,12 +137,10 @@ jobs:
 
       - name: Install Rust toolchain
         if: steps.relevance.outputs.run == 'true'
-        uses: dtolnay/rust-toolchain@stable
-        with:
-          toolchain: 'nightly-2026-08-20'
+        run: rustup toolchain install nightly-2026-08-20 --profile minimal
       - uses: ./.github/actions/setup-llvm22
 
-      - uses: Swatinem/rust-cache@v2
+      - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2
         if: steps.relevance.outputs.run == 'true'
         with:
           shared-key: "${{ runner.os }}-perry"
@@ -156,7 +154,7 @@ jobs:
 
       - name: Setup Node oracle
         if: steps.relevance.outputs.run == 'true'
-        uses: actions/setup-node@v7
+        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7
         with:
           # Single source of truth: .node-version. Every kernel's stdout is
           # diffed against this node, so the pin is a correctness input.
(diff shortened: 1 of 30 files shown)


proggeramlug:
Landed on `main` via #8696 (squash `fdbddebe6`).

For a CI-mechanics PR the risk I care about is silently-dropped coverage, so I diffed the job and step sets against `main`: no job removed, one added (`gap-suite-build` from #8686), and the single removed step is the fast-mode archive build — which moved into that job rather than disappearing. Clean.

Validated on the merged result: all 30 lint checkers, codegen 1202/0, runtime 2655/0, perry-bin 1031/0. Thanks!

PR #8686, opened 2026-08-24
perf(ci): build gap-suite's fast-mode archives once, share across shards

jdalton:
## Summary

Cuts roughly 50 minutes of pure redundant compute from every PR run by building gap-suite's fast-mode release archives once instead of independently rebuilding them in each of the 6 parallel shards.

<details>
<summary>What was happening and the measured cost</summary>

Every `gap-suite` shard runs `cargo build --release -p perry -p perry-runtime -p perry-stdlib -p perry-runtime-static -p perry-stdlib-static` before running its own slice of the suite. I pulled the actual per-step timing from a real PR run: that build step alone took 10.0 minutes, and it produces byte-identical output in every shard, since all 6 shards build from the same commit with the same flags.

Because the shards run in parallel, this did not directly cost PR wall-clock time. It did cost roughly 50 minutes of runner compute per PR run (5 unnecessary repeats of a 10-minute build), and on a capacity-limited runner pool that compute pressure can itself delay when other queued jobs get a slot.

</details>

## Changes

Adds a `gap-suite-build` job that runs only when the plan selects fast mode (PR and sweep tiers; full mode's shards genuinely need their own per-test auto-optimize rebuild through `run_gap_tests.sh`, which this does not touch). It builds once, then uploads the three files the fast-mode harness actually links against: the `perry` binary, `libperry_runtime.a`, and `libperry_stdlib.a`. Every `gap-suite` shard now downloads that artifact instead of rebuilding, restoring the binary's executable bit defensively after the download. `gap-suite`'s own `if:` condition uses `always()` together with an explicit check on `gap-suite-build`'s result, so a build that is correctly SKIPPED (full mode) does not cascade into skipping the shards, while a genuine build FAILURE still stops them from starting rather than running six shards doomed to fail identically.

While in this section of the file, I also inlined `dtolnay/rust-toolchain` and SHA-pinned `actions/checkout`, `actions/setup-node`, and `Swatinem/rust-cache` in the existing `gap-suite` job, for local consistency with the new job right above it. The broader sweep across the rest of the workflow tree is a separate PR, #8684, since that change is unrelated to this one in scope.

## Related issue

n/a

## Test plan

I confirmed the per-step timing of the redundant build against a real PR run (`https://github.com/PerryTS/perry/actions/runs/32674242904`, shard 5's job), which showed the 10.0-minute build step. I traced the fan-in `gate` job's own gating script and confirmed it only inspects the result of jobs explicitly listed in its own `needs:` array, which does not include `gap-suite-build` — so the new job is fully transparent to branch protection, and only `gap-suite`'s own result affects it. I validated the edited workflow file as YAML and ran `actionlint` against it before and after this change; the only differences are line-number shifts in three pre-existing shellcheck findings elsewhere in the file, with no new findings introduced. I have not been able to run this workflow for real, since that requires a merge to actually exercise the new job graph on this repository's runners.

## Checklist

I have not bumped the workspace version or edited CLAUDE.md or CHANGELOG.md. My commit follows the `perf:` prefix convention used in the log. I have read CONTRIBUTING.md and agree to the Code of Conduct.

## Summary by CodeRabbit

* **CI Improvements**
  * Faster gap-suite runs by building shared release artifacts once and reusing them across test shards.
  * Full-mode runs continue using their existing optimization workflow.
  * Improved workflow reliability with pinned tool versions and executable downloaded binaries.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2226,9 +2226,57 @@ jobs:
   # in the full tier (scripts/ci_plan.py GAP_SUITE). The `gate` fan-in below is
   # what branch protection requires; a single shard's red bubbles up through it.
   # ---------------------------------------------------------------------------
-  gap-suite:
+  # Builds the fast-mode gap-suite's shared release archives ONCE instead of
+  # redundantly rebuilding the identical archives in every shard (measured
+  # ~10 min per shard, repeated across all 6 PR-tier shards -- ~50 min of
+  # pure redundant compute per run). Full mode is unaffected: its shards
+  # build their own feature-stripped runtime per distinct test through
+  # run_gap_tests.sh's own auto-optimize path, which this job does not
+  # attempt to share.
+  gap-suite-build:
     needs: plan
-    if: fromJSON(needs.plan.outputs.plan).jobs.gap_suite
+    if: fromJSON(needs.plan.outputs.plan).jobs.gap_suite && fromJSON(needs.plan.outputs.plan).gap.mode == 'fast'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+
+      - name: Install Rust toolchain
+        run: rustup toolchain install nightly-2026-08-20 --profile minimal
+      - uses: ./.github/actions/setup-llvm22
+
+      - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2
+        with:
+          shared-key: "${{ runner.os }}-perry"
+          save-if: ${{ github.ref == 'refs/heads/main' }}
+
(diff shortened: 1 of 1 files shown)


proggeramlug:
Landed on `main` via #8696 (squash `fdbddebe6`).

The thing I checked hardest was the skipped-vs-failed distinction, since a skipped job satisfies a dependency and that has produced gates here that couldn't fail. Your guard gets it right: `always()` stops a SKIPPED build (full mode) from cascading into skipped shards, while `result == 'success' || result == 'skipped'` still lets a genuine build FAILURE stop them. The mode-gated download, the runnable-binary check, and the explicit `PERRY_BIN`/`PERRY_RUNTIME_DIR` export all matter too — without that last one the wrapper exits 0 having run nothing.

Validated on the merged result: all 30 lint checkers, codegen 1202/0, runtime 2655/0, perry-bin 1031/0. Thanks!

PR #8687, opened 2026-08-24
fix(codegen): retry oversized RS4GC functions with shadow roots (#8679)

### LINKED ISSUE #8679: codegen: reconcile the root-spill threshold (#8623, 32M) with the RS4GC instruction budget (#8586, 1.57M) — budget-gap functions get refused instead of spilled
## Problem

Two #8583 mechanisms disagree on where a function is "too big for RS4GC":

- **#8623** set the root-spill decision threshold to **32,000,000** relocations (`DEFAULT_ROOT_SPILL_RELOCATIONS`), measured at the point RS4GC's *rewrite* still finishes in bounded time.
- **#8586** refuses a unit when a function's **post-rewrite instruction count** exceeds **1,572,864** (`PERRY_LL_RS4GC_MAX_INSTRS`), the point the *`-Os`/`-O3` optimizer* goes super-linear.

Since a function's added instructions ≈ its relocation estimate, a function whose estimate lands in **(1.57M, 32M)** does **not** spill (estimate < 32M) yet **overruns the budget** (post-rewrite > 1.57M) → the budget assertion **hard-refuses the whole module**.

## Concrete case

Compiling the Claude Code cli.js 2.1.112 bundle, `__AnonShape_ec12be8bd7b0a2b9_constructor` (a large closed-shape object-literal constructor) grew **34,009 → 2,280,128** instructions under RS4GC — above the 1.57M budget, below the 32M spill threshold — and was refused. It only compiled with `PERRY_ROOT_SPILL_RELOCATIONS=1500000`. (Estimate accuracy for this shape is fixed by #8678, which counts property/index stores; but even with an accurate ~2.25M estimate it still won't spill at the 32M default.)

## Options

1. **Align the spill default with the budget** — set `DEFAULT_ROOT_SPILL_RELOCATIONS` at/below `PERRY_LL_RS4GC_MAX_INSTRS` so anything that would overrun the budget spills first. (Note this revisits #8623/#8620: the "8M fans-out-fine" cases #8620 measured would *also* overrun the 1.57M budget, so they should spill — RS4GC finishing ≠ the optimizer handling the result.)
2. **Preferred: make the budget trigger a spill-retry, not a hard refusal** — when a function overruns the budget post-rewrite, re-lower it with `request_shadow_frame_spill()` (roots to the shadow frame, RS4GC skips it) instead of refusing the module. This is the "#8421/#8128 root-lowering choice, never refuse" principle applied to the budget, and it needs no threshold tuning at all.

Option 2 is the durable fix. Found while getting the cli.js bundle to a working native binary (#8583).

proggeramlug:
## Summary

Closes #8679.

- replace the post-RS4GC instruction-budget hard refusal with a typed spill-retry request carrying the exact over-budget function names and pre/post instruction counts
- let an already-lowered `LlFunction` switch from native statepoint roots to a complete precise shadow frame, including frame setup, retained slot binds, balanced pops, and removal of the GC strategy
- rebuild and retry direct native modules, split native codegen units, legacy text transport, and diff modes at the originally requested optimization level
- keep split-unit retry sources on the producer thread in a one-unit-per-worker bounded window, dropping each lowering graph as soon as its unit reaches a final result
- budget only functions that actually entered RS4GC, so a successfully shadow-spilled function cannot trigger an infinite second retry
- preserve typed retry errors through linker context without consuming the process-wide failed-IR retention slot

`PERRY_LL_RS4GC_MAX_INSTRS=warn:<n>` and disabled-budget behavior remain unchanged. `DEFAULT_ROOT_SPILL_RELOCATIONS` remains at the measured 32M fan-out cliff; estimator misses are now handled durably by the retry.

No version bump is included.

## Verification

- `cargo test -p perry-codegen --features llvm-inprocess --lib` — 1201 passed, 1 ignored
- focused direct-native and split-worker budget retry tests, including proof that the retried split object has no compact statepoint map
- `RUSTFLAGS='-D warnings' cargo check -p perry-codegen --features llvm-inprocess --all-targets`
- `RUSTFLAGS='-D warnings' cargo check -p perry --features llvm-inprocess --tests`
- `cargo check -p perry-codegen --no-default-features --lib` (passes with the same four feature-off dead-code warnings)
- `git diff --check`

Clippy with `-D warnings` is currently blocked by existing repository lint debt (including `perry-diagnostics` and 191 existing `perry-codegen` findings); no clippy-driven unrelated edits are included.

## Summary by CodeRabbit

* **Bug Fixes**
  * Large native functions that exceed the LLVM instruction budget are now automatically retried with an alternative memory-root strategy instead of failing.
  * Improved compilation reliability across standard, split-unit, differential, and cached build paths.
  * Prevented repeated retries and reduced the risk of compiler hangs during optimization.

* **Documentation**
  * Updated build-cache and instruction-budget documentation to reflect the new retry behavior.

### DIFF
--- a/changelog.d/8679-rs4gc-budget-spill-retry.md
+++ b/changelog.d/8679-rs4gc-budget-spill-retry.md
@@ -1,0 +1,4 @@
+Fixed large native functions whose statepoint rewrite crossed the LLVM
+instruction budget but fell below the root-spill estimate threshold. Perry now
+re-lowers only those functions with precise shadow-frame roots and retries the
+same optimization pipeline instead of refusing the whole codegen unit.
--- a/crates/perry-codegen/src/codegen/helpers.rs
+++ b/crates/perry-codegen/src/codegen/helpers.rs
@@ -387,9 +387,10 @@ pub(crate) fn inline_hot_small_max_call_sites() -> u32 {
 /// fan-out it avoids (an ~8M function spilled in 303 s vs 180 s fanned out,
 /// #8620) — and above it fan-out risks not finishing and the shadow frame wins.
 /// The former 4M default fired on ~8M functions that fan out fine in minutes.
-/// The post-RS4GC instruction-budget assertion (#8586, inprocess.rs) backstops
-/// any function this estimate misses: it fails loudly rather than hanging, so
-/// raising the threshold is safe.
+/// The post-RS4GC instruction budget (#8586/#8679, inprocess.rs) backstops any
+/// function this estimate misses: it re-lowers that function onto a precise
+/// shadow frame and retries before LLVM's optimizer can hang, so raising the
+/// estimate threshold is safe.
 ///
 /// `PERRY_ROOT_SPILL_RELOCATIONS=<n>` overrides it; `0` disables spilling
 /// (every function stays on native statepoints, the pre-#8583 behavior).
--- a/crates/perry-codegen/src/codegen/mod.rs
+++ b/crates/perry-codegen/src/codegen/mod.rs
@@ -3353,28 +3353,30 @@ pub fn compile_module(hir: &HirModule, opts: CompileOptions) -> Result<Vec<u8>> 
             progress.phase(3, "object ready; releasing generated IR");
             return result;
         }
-        let units = llmod.render_codegen_units(n_units);
-        log::debug!(
-            "perry-codegen: split '{}' into {} codegen units",
-            hir.name,
-            units.len()
-        );
-        // #7154: dump the units. The comment above used to claim `PERRY_SAVE_LL`
-        // took the single-text path — it never did; this `return` fires before
-        // the `PERRY_SAVE_LL` write below. So `--trace llvm` silently emitted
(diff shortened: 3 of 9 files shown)


proggeramlug:
Landed on `main` via #8696 (squash `fdbddebe6`), with #8686 and #8684.

Before landing a retry loop I wanted the termination guarantee to be structural rather than argued, and it is: `request_shadow_frame_spill()` latches on `force_shadow_frame` and returns `false` if already set, `apply_budget_spill_retry` records only the `true` cases, and any violation not recorded becomes a hard error naming it. A function that has already been spilled fails loudly instead of retrying — it cannot loop.

Validated on the merged result: all 30 `lint`-job checkers, `perry-codegen --lib` 1202/0, `perry-codegen --features llvm-inprocess --lib` 1202/0 (the feature your change actually touches), `perry-runtime --lib` 2655/0, `perry --bin perry` 1031/0.

This closes the loop on #8678's note that estimate accuracy alone wasn't enough while #8623's 32M sat above the 1.57M budget. Thanks!

PR #8689, opened 2026-08-24
perf(gc): prune transitively non-collecting statepoints

### LINKED ISSUE #8596: gc: reduce safepoint density (safepoint polling instead of every-call statepoints)

proggeramlug:
## Summary

- compute a whole-module greatest-fixed-point GC-effect closure for generated functions, including pure recursive SCCs
- mark direct calls to transitively non-collecting generated callees with `gc-leaf-function`, while allocation/poll paths, indirect calls, unknown externals, and cross-module calls remain statepoints
- carry the proof through whole-module text emission, split codegen units, native LLVM construction, and `invoke` edges
- classify the six audited shadow-frame bookkeeping helpers as non-collecting and keep their checker authority pinned by tests
- document the LLVM caller-frame constraint that bounds how far polling-style statepoint reduction can go with Perry's current native-root representation

## Why this shape

Blanket polling is not sound with the current LLVM statepoint frame representation. If `A` calls `B` and collection begins at an allocation or poll inside `B`, `A` is suspended at the `A -> B` return PC. Moving GC still needs a relocation map for `A`; a poll in `B` does not recreate it. This change therefore implements the maximal safe reduction available today: omit the caller-edge statepoint only when the callee cannot transitively collect.

Going further requires caller roots to live in a mutable frame representation (for example, a shadow frame) rather than merely moving the collection trigger.

## Validation

- `cargo check -p perry-codegen --all-features`
- `cargo check -p perry-codegen --no-default-features`
- `cargo clippy -p perry-codegen --all-targets --all-features` (existing warning backlog only)
- GC-effect closure tests: 14 passed
- text/native/split-unit RS4GC parity and `invoke` attribute tests passed
- `cargo test -p perry-codegen --all-features --test loop_safepoint_purity`: 8 passed
- full `perry-codegen` library suite: 1,201 passed, 1 ignored, 3 filtered

The three filtered Windows-host tests reproduce on `origin/main`: two compare COFF objects byte-for-byte even though randomized temporary object names are embedded, and one assumes LLVM shrink-wraps a specialized clone to a frameless entry on this host.

No version bump is included.

Closes #8596

## Summary by CodeRabbit

* **Performance**
  * Reduced unnecessary garbage-collection safepoints for calls proven not to allocate or trigger collection.
  * Preserved safepoints for allocating, indirect, unknown, cross-module, and exception-handling calls.

* **Reliability**
  * Improved handling of recursive call paths while maintaining conservative garbage-collection behavior.

* **Documentation**
  * Added documentation describing safepoint reduction and its safety guarantees.


proggeramlug:
Landed on `main` via #8700 (squash `8224d879a`), with #8697 and #8699.

Your `native_emit.rs` conflict with #8687 (which landed an hour earlier) was one hunk and positional only — `render_fn_external_with_gc_leaf_callees` replacing `render_fn_external`, with `gc_leaf_callees` destructured in that same function by your own change. I verified the binding was in scope rather than assuming, then took your side.

`gc_leaf_asm_barrier_survives_rs4gc_unwrapped` passes, which is the one I most wanted green given #8121.

Validated on the merged result: all 30 lint checkers, codegen 1213/0, runtime 2655/0, perry-bin 1032/0, dispatch 6/0, and `cargo check` clean of warnings. Thanks!

PR #8697, opened 2026-08-24
perf(codegen): finish structured module-entry outlining (#8595)

### LINKED ISSUE #8595: codegen: structured intra-function outlining of oversized generated functions (module-entry IIFE)

proggeramlug:
Closes #8595.

## Summary

- Enable structured module-entry outlining automatically for entries with at least 1,000 top-level HIR statements or 4,000 estimated safepoints.
- Bound generated chunks to roughly 200 statements or 1,000 safepoints and mark them no-inline so LLVM cannot reconstruct the oversized entry before RS4GC, ISel, or register allocation.
- Move original declarations and statements unchanged, preserve source order, and promote only bindings that need cross-function storage to rooted module globals.
- Reconstruct the logical entry stream for declaration, export, constant, static-field, closure-name, and early process.env analyses.
- Support exports, Script globalThis reflection, and structured top-level control flow; retain fail-safe fallbacks for top-level await and module-level TDZ preallocation.
- Keep PERRY_OUTLINE_ENTRY=1 as a force-on diagnostic override and =0 as an explicit opt-out.

## Validation

- cargo check -p perry-codegen -p perry --all-targets
- cargo test -p perry-codegen --lib codegen::entry_outline::tests -- --test-threads=1 (14 passed)
- cargo test -p perry --test entry_outline_transform_8595 -- --test-threads=1 (4 passed)
  - differential outlined/non-outlined execution
  - moving minor GC configurations
  - exports, structured control flow, Script reflection, and early dependency environment initialization
  - default-on oversized-entry behavior
- rustfmt --check on all touched Rust files
- git diff --check

An additional full perry-codegen library sweep passed 1,203 tests, with 1 ignored. Three unrelated Windows-local assertions failed: the preserve-none prologue test sees LLVM emit pushq %rbp, and two native-object byte-equality tests embed different randomized COFF temp filenames.

No version bump is included.

## Summary by CodeRabbit

* **New Features**
  * Large module entry points are automatically split into smaller functions to improve compilation and runtime handling.
  * Outlining can be configured or disabled with `PERRY_OUTLINE_ENTRY`.
  * Structured control flow, exports, global reflection, static fields, and cross-function bindings are supported.
* **Bug Fixes**
  * Preserved source order and early environment reads across outlined code.
  * Improved compiler diagnostics when outlining is skipped.
  * Safely excludes modules using top-level await or sensitive initialization patterns.


proggeramlug:
Landed on `main` via #8700 (squash `8224d879a`).

Checked the two things that matter for a default-on behaviour change: the thresholds are real gates (`DEFAULT_AUTO_MIN_STMTS` 1,000 / `DEFAULT_AUTO_MIN_SAFEPOINTS` 4,000, so ordinary modules are untouched), and both `PERRY_OUTLINE_ENTRY=1` force-on and `=0` opt-out are still wired. Your 14 entry-outline unit tests pass.

Validated on the merged result: all 30 lint checkers, codegen 1213/0, runtime 2655/0, perry-bin 1032/0. Thanks!

PR #8698, opened 2026-08-24
fix(hir): reject legacy numerics in strict eval

proggeramlug:
## Summary
- re-lex constant direct-eval bodies under inherited strict mode
- surface deferred legacy decimal/octal diagnostics as `SyntaxError`
- keep modern octal literals and numeric-looking text in comments/strings valid

Refs #5894

## Validation
- Test262 `language/literals/numeric`: 157/157 pass (baseline: 156/157)
- `cargo test --release -p perry-hir --lib`: 331 passed, 1 ignored
- `cargo test --release -p perry-hir --test c262_parity direct_eval`: 2 passed
- workspace rustfmt check: all 79 packages passed
- `bash scripts/check_file_size.sh`
- `git diff --check`

No version or changelog changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strict evaluation now correctly rejects legacy numeric literals with a runtime `SyntaxError`.
  * Numeric-looking text inside strings and comments is no longer incorrectly rejected.
  * Modern numeric literal formats continue to be accepted.

PR #8702, opened 2026-08-24
perf(codegen): inline trusted boxed capture access

proggeramlug:
## Summary

- load compiler-validated raw box capture pointers once at entry to private exact-arrow clones
- directly read and write the non-moving box cell while retaining the TDZ slow path and GC write barrier
- keep public and indirect closure bodies on the existing checked runtime accessors
- derive the closure capture offset from the target pointer width, including arm64_32 and wasm32

## Why this is safe

The private clone is selected only after the runtime validates the public closure identity, exact capture count, and compiler-installed boxed-capture mask. Capture slots are immutable. The live exact capture edge retains each non-moving box cell for the invocation even if the closure relocates. Direct reads explicitly test TAG_TDZ and call the existing trusted getter on that cold path, preserving ReferenceError and suppression-window behavior. Direct stores retain the ordinary child-shading barrier.

## Performance

Unchanged codehz/ecs comprehensive benchmark on the quiet M1 Mac mini, accumulation row, 11 alternating repeat-256 pairs:

- exact PR #8676 parent: 0.249309 -> 0.217134 ms/op, 12.93% median paired improvement, 11/11 wins
- rebased exact parent 10d9ca56e: 0.632372 -> 0.612545 ms/op, 3.08% median paired improvement, 11/11 wins

All 44 measured processes passed the 7/7 suite and exact 50,005,000 checksum. The absolute regression between those parents is pre-existing newer-main behavior and is not attributed to this patch.

## Validation

- cargo fmt --check
- 5/5 trusted-box callback codegen tests
- target-aware ClosureHeader layout test
- 1,200/1,200 non-ignored perry-codegen unit tests passed (one ignored)
- focused executable passed normal and forced verified moving-GC runs with reads, writes, increment, collection inside every callback, and TDZ
- unchanged ECS passed 7/7 locally and in every benchmark process

The package-wide test command reaches one existing typed_feedback_guards_direct_class_method_specialization assertion failure. It reproduces identically on the clean exact parent and is outside this diff.

## Summary by CodeRabbit

- **Performance**
  - Improved performance for compiler-generated arrow functions that access captured boxed values.
  - Reduced overhead for captured-value reads and updates through more direct access paths.

- **Reliability**
  - Preserved public closure validation, TDZ handling, and garbage-collection safety.
  - Improved compatibility across 32-bit, 64-bit, and WebAssembly targets.

- **Tests**
  - Added coverage verifying optimized captured-value access and platform-specific closure layouts.


proggeramlug:
Landed on `main` via #8705 (squash `118969ffc`), with #8698.

I verified your safety argument rather than taking it on trust, since caching a raw heap pointer across a collection is exactly how this codebase has been bitten before. It holds: `closure/box_captures.rs` opens with "Lifetime bridge between GC closures and **malloc-side** async box cells", so the cells genuinely don't move under evacuation, and `closure_box_captures_owner_moved` already handles the case where the *closure* relocates — which is precisely the distinction your "non-moving box cell ... even if the closure relocates" wording draws.

Your pre-existing-failure note was accurate, and I confirmed it independently on clean `main` before acting: `typed_feedback_guards_direct_class_method_specialization` asserted `call double @js_class_field_add`, which #8653 deliberately stopped emitting for that shape. Red since then and invisible to per-PR CI, because codegen integration suites only run when the diff names them. Fixed in this merge — the assertion now pins the guarded fast path and asserts the unconditional helper is *not* called.

Validated: all 30 lint checkers, runtime 2655/0, codegen 1214/0, all 28 codegen integration suites 0 failures, hir 331/0. Thanks — the profile writeup and the paired-repeat methodology were both easy to check against.

PR #8703, opened 2026-08-24
perf(transform): scalar-replace aggregate call literals

### LINKED ISSUE #8691: perf(codegen): scalar-replace non-escaping aggregate literals
## Summary

Perry does not scalar-replace short-lived aggregate literals passed through known ECS helper calls. The public `perform-ecs/destroy` loop allocates fresh arrays and descriptor objects several times per entity:

```js
this.ecs.createEntity([
  { component: Position },
  { component: Velocity },
]);

this.ecs.addComponentsToEntity(entity, [{ component: Position }]);
this.ecs.addComponentsToEntity(entity, [{ component: Velocity }]);
```

Once the consumers are known/inlined, these aggregates do not need material object/array identity on the common path. Current lowering performs zero scalar replacements, leaving allocation, GC-layout tracking, write barriers, array pushes/gets, and boxing in the hottest 82% of the benchmark.

Add escape analysis and scalar replacement for non-escaping object/array literals across specialized/inlined calls, with explicit materialization when identity or a generic call makes the allocation observable.

```js
class Position {}
class Velocity {}

let checksum = 0;

function consume(initializers) {
  for (let i = 0; i < initializers.length; i++) {
    const initializer = initializers[i];
    if (initializer.component === Position) checksum += 1;
    if (initializer.component === Velocity) checksum += 2;
  }
}

const iterations = 500_000;
  consume([{ component: Position }, { component: Velocity }]);
  consume([{ component: Position }]);
  consume([{ component: Velocity }]);
}
console.log(JSON.stringify({
  checksum,
}));
```

Expected checksum: `3000000`.

Build and retain artifacts:

```bash
PERRY_NO_AUTO_OPTIMIZE=1 \
```

The public integration reproduction is `ddmills/js-ecs-benchmarks`, `perform-ecs@0.7.8`, suite `Destroy`.

## Current evidence

For the corrected public workload on Perry `7ad718ab4287641cb2b29dce3a056edc45d4c7f8`, the lowering report contains:

- 494 records;
- **0 scalar replacements**;
- 61 inserted boxes;
- 11 emitted barriers and 0 barrier eliminations;
- 9 dynamic fallbacks;
- 332 `js_value` representations versus only 34 `i32` and 16 `i1` representations.

Materialization reasons are dominated by runtime API edges (46), function ABI (8), and return ABI (7). A symbolized profile shows repeated allocation/GC-layout bookkeeping, generic array operations, GC-header validation, and barriers. The workload remains 33.091x slower than Node; entity creation and component addition account for approximately 82% of its time.

This optimization should follow or consume #8693's guarded direct-call/inlining work for imported methods that capture `this`; unknown calls must continue to force escape/materialization.

## Proposed direction

- Run escape analysis after specialization/inlining, when actual consumers and field/index uses are visible.
- Track object and fixed-length array literals field-by-field when they do not escape to a heap store, return, closure, unknown call, reflective operation, proxy, weak reference, or identity comparison.
- Replace their property/index reads with SSA/native values and eliminate their allocations, layout registration, roots, and barriers.
- Support path-sensitive materialization at an explicit bailout/escape edge instead of rejecting the entire function when only a cold path observes identity.
- Start with plain data object literals and fixed-length packed array literals; do not include arbitrary constructors or observable custom prototypes in the first slice.
- Record escape, scalar-replacement, and forced-materialization decisions and reasons in `--explain-lowering` and native-region verification.

## Semantic constraints

- Preserve object/array identity when compared, returned, captured, stored, reflected upon, passed to an unknown call, proxied, or used with WeakRef/WeakMap/WeakSet/FinalizationRegistry behavior.
- Preserve property creation/evaluation order, computed-key side effects, spreads, getters/setters, prototype semantics, exceptions, and array holes/length behavior.
- Preserve GC root visibility when a value materializes at a side exit or collecting call.
- Never scalar-replace a value merely because the benchmark consumer happens not to inspect it; the non-escape proof must be general and verifier-backed.

## Acceptance criteria

- Register the reproduction as semantic and compiler-output coverage; Node and Perry must print checksum `3000000`.
- `--explain-lowering` records scalar replacement of the two descriptor objects and their fixed-length carrier arrays in the optimized fixture.
- The hot arm contains no heap allocation, array push/get helper, layout-slot registration, or write barrier for those replaced aggregates.
- Native-region verification rejects a claimed scalar replacement without a complete escape proof and materialization records for all escape edges.
- Negative tests cover return/store/capture, identity comparison, unknown and collecting calls, reflection, spreads/computed keys, getters/setters, prototypes, proxies, WeakMap/WeakSet/WeakRef where supported, exceptions, and forced-moving GC.
- Re-run `ddmills/js-ecs-benchmarks` `perform-ecs/destroy` with exact component-ID/view-count parity and the complete adapter suite.
- On the quiet M1 alternating protocol, require at least a 5% median improvement with at least 9/11 wins; report allocations/op or allocation samples, barrier counts, peak RSS, executable size, and scalar-replacement/materialization counts.

## Related work

- #8693's guarded specialization/inlining of imported class methods that capture `this` is the principal enabler for the real `perform-ecs` call graph.
- #5497 tracks type-directed native operations broadly; this ticket is specifically aggregate escape analysis and scalar replacement.
- #8596 reduces safepoint density, but eliminating the temporary allocations and generic calls should remove more safepoint/barrier work at its source.

proggeramlug:
## Summary

- inline narrowly proven fixed-aggregate consumer loops so static unrolling can expose their element reads
- scalar-replace non-escaping carrier arrays and anonymous-shape descriptor objects while preserving source-order evaluation
- keep identity, mutation, reflection, closure, dynamic-index, and unknown-call cases on the materialized path
- record carrier/object and write-barrier elimination in `--explain-lowering`

## Validation

- `cargo test -p perry-transform` (87 passed)
- `cargo fmt -p perry-transform -p perry-codegen -- --check`
- standalone #8691 repro: `3000000`; zero object alloc/property access/array access/consumer calls in the hot loop
- registered compiler-output workload: `17` and `3000000`; four `ScalarAggregateFieldInit` records; zero aggregate helpers in `scalarAggregateCallChecksum`

No version bump.

Closes #8691

## Summary by CodeRabbit

* **Performance**
  * Improved compiler optimization for short arrays of non-escaping object literals passed to eligible aggregate functions.
  * Reduced unnecessary runtime allocations and property-access overhead while preserving normal behavior for escaping or identity-sensitive values.

* **Bug Fixes**
  * Preserved object identity semantics in optimized aggregate calls.
  * Added safeguards for mutation, reflection, dynamic access, and other cases where optimization is unsafe.

* **Tests**
  * Added regression coverage and benchmark validation for aggregate-call optimization and checksum results.

### DIFF
--- a/benchmarks/compiler_output/fixtures/scalar_replacement_literals.ts
+++ b/benchmarks/compiler_output/fixtures/scalar_replacement_literals.ts
@@ -10,4 +10,29 @@ function scalarReplacementChecksum(): number {
   return values[0] + values[1] + values[2] + values.length;
 }
 
+class Position {}
+class Velocity {}
+
+let aggregateChecksum = 0;
+
+function consumeAggregate(initializers: { component: unknown }[]): void {
+  for (let i = 0; i < initializers.length; i++) {
+    const initializer = initializers[i];
+    if (initializer.component === Position) aggregateChecksum += 1;
+    if (initializer.component === Velocity) aggregateChecksum += 2;
+  }
+}
+
+function scalarAggregateCallChecksum(): number {
+  aggregateChecksum = 0;
+  const iterations = 500_000;
+  for (let i = 0; i < iterations; i++) {
+    consumeAggregate([{ component: Position }, { component: Velocity }]);
+    consumeAggregate([{ component: Position }]);
+    consumeAggregate([{ component: Velocity }]);
+  }
+  return aggregateChecksum;
+}
+
 console.log(scalarReplacementChecksum());
+console.log(scalarAggregateCallChecksum());
--- a/benchmarks/compiler_output/workloads.toml
+++ b/benchmarks/compiler_output/workloads.toml
@@ -1349,17 +1349,29 @@ function_contains = "scalarReplacementChecksum"
 regex_none = ["@js_object_get_field", "@js_object_set_field", "@js_array_get", "@js_array_set"]
 detail = "scalar-replaced literals do not use runtime property or array access helpers"
 
+[[workloads.scalar_replacement_literals.ir_checks]]
+name = "known_aggregate_call_no_object_or_array_heap_alloc"
+function_contains = "scalarAggregateCallChecksum"
+regex_none = ["@js_object_alloc", "@js_object_alloc_with_shape", "@js_array_alloc"]
+detail = "known fixed-aggregate consumers scalar-replace descriptor objects and carrier arrays"
+
+[[workloads.scalar_replacement_literals.ir_checks]]
+name = "known_aggregate_call_no_property_or_array_runtime_access"
+function_contains = "scalarAggregateCallChecksum"
+regex_none = ["@js_object_get_field", "@js_object_set_field", "@js_array_get", "@js_array_set"]
+detail = "known fixed-aggregate consumers read scalar fields without runtime aggregate helpers"
+
 [[workloads.scalar_replacement_literals.stdout_checks]]
 name = "scalar_replacement_checksum"
-equals = "17\n"
+equals = "17\n3000000\n"
 detail = "scalar-replacement fixture stdout checksum"
 
 [workloads.scalar_replacement_literals.native_rep_checks]
-function_contains = "scalarReplacementChecksum"
 allow_materialization_reasons = ["runtime_api"]
 
 [[workloads.scalar_replacement_literals.native_rep_checks.require_records]]
 name = "scalar_object_literal_store"
+source_function = "scalarReplacementChecksum"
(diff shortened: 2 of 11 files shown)


proggeramlug:
Landed on `main` via #8708 (squash `c76b43930`), with #8706.

The escape analysis is the load-bearing part and your tests cover the three ways it has to stay conservative — identity observed, mutation reflected, unknown call — so that was straightforward to check.

Validated: all 30 lint checkers, runtime 2655/0, codegen 1214/0, transform 87/0, all codegen integration suites 0 failures. Thanks!

PR #8704, opened 2026-08-24
runtime: complete C-ABI FFI layer

### LINKED ISSUE #6562: runtime: C-ABI FFI layer (bun:ffi shape) — dlopen, typed calls, pinned buffers, toArrayBuffer, JSCallback
The single biggest platform feature for compiling Bun-specific apps (driver: opencode, but this is what the Bun-native ecosystem standardizes on): a C-ABI FFI layer with the `bun:ffi` API shape (and optionally Node's experimental `node:ffi` shape — `bun-ffi-structs` already bridges both, so one implementation can serve either import).

## Evidence from opencode

Everything native in the TUI product funnels through **plain C-ABI dylibs — no N-API anywhere** (verified: `libopentui.dylib` from `@opentui/core-darwin-arm64@0.4.5` exports 1,775 C symbols, zero `napi_*` entries; the "node backend" is the same dylib driven via experimental `node:ffi`):

| consumer | lib | FFI decl size | notes |
|---|---|---|---|
| `@opentui/core` (the TUI renderer) | `libopentui.dylib` per platform | ~348 typed symbol decls | renderer + buffers + **embedded yoga** (`YG*`) + **embedded tree-sitter** with bundled grammars (js/ts/markdown/zig) |
| `bun-pty` | `librust_pty.{dylib,so,dll}` (bundled in tarball) | **17 symbols** | pty spawn/read/write/resize/kill |
| `@ff-labs/fff-bun` | `@ff-labs/fff-bin-*` platform pkgs | dlopen in `src/ffi.ts` | fuzzy finder; opencode's node condition stubs it gracefully — deferrable |

## Primitives required (from actual usage, not the full spec)

- `dlopen(path, symbolTable)` with typed args/returns — `FFIType.{void, ptr, pointer, cstring, bool, i8..i64, u8..u64, usize, f32, f64, function, callback}` (observed set in `@opentui/core` dist).
- `ptr(arrayBufferOrView)` — hand JS buffer memory to native. **GC design constraint: perry's GC relocates objects** (cf. the forwarding-stub/array-growth work, #6228/#6471) — FFI-exposed buffers must be pinned or allocated in non-moving/off-heap space for their native lifetime.
- `toArrayBuffer(ptr, byteOffset, byteLength)` — 28 uses in opentui — zero-copy JS views over *native-owned* memory; needs external backing-store support in perry's ArrayBuffer (no GC ownership, explicit lifetime).
- `CString` (read NUL-terminated, encode).
- `JSCallback` / `FFIType.function` — **native→JS callbacks** (opentui registers them). Hardest piece: trampoline into perry's JS execution, closure rooting, and a decision on threading (reject cross-thread invocation initially unless opentui proves to need it — verify which callbacks it actually registers before building).
- `bun-ffi-structs` is pure JS over these primitives — free once they exist.

For a native-code runtime, this is natural territory: direct dlopen + call-stub generation (libffi or generated thunks), no JIT trampolines needed.

## Suggested bring-up

1. dlopen + typed calls + `ptr` on pinned buffers + `CString`, proven against **bun-pty's 17-symbol dylib** (spawn a shell, echo round-trip) — small, observable, exercises strings/ints/pointers/read-loop.
2. `toArrayBuffer` external backing stores.
3. `JSCallback` trampolines.
4. opentui's 348-symbol table + renderer bring-up (the opencode target tracker owns the app-level milestone).

Acceptance per stage; stage 1 acceptance: a perry-compiled script drives bun-pty end-to-end on macOS-arm64 + linux-x64.

proggeramlug:
## Summary

- complete the `bun:ffi` scalar C ABI with register and stack argument marshalling for SysV x86_64 and AAPCS64, covering the 13/14-argument FFF and OpenTUI call shapes
- implement pinned pointers, zero-copy native memory views, `CString`, `read`, `CFunction`, `linkSymbols`, `viewSource`, rooted callbacks, and synchronous foreign-thread callback handoff
- add the Node 26 `node:ffi` compatibility surface used by unmodified OpenTUI/Yoga, including CommonJS `createRequire` interop, bigint pointers, library-owned callbacks, and memory helpers
- finish the stage-1 real `bun-pty` shell roundtrip and expand the native fixture across scalar types, stack args, callbacks, memory, errors, and compatibility paths

## Validation

- `cargo test -p perry-runtime bun_ffi` (29 passed)
- `cargo test -p perry-runtime builtin_module_list_tests` (3 passed)
- `cargo test -p perry-api-manifest --tests` (42 passed)
- `cargo check -p perry-runtime -p perry-hir -p perry-codegen -p perry-api-manifest`
- `cargo test -p perry --test bun_ffi_stage1 -- --nocapture` (3 passed, including real `bun-pty`)
- x86_64 Linux `perry-runtime` object cross-build
- unmodified OpenTUI 342-symbol table opened successfully; `@opentui/core/yoga` computed a 320 x 180 layout
- unmodified `@ff-labs/fff-bun` created its native finder and exercised its threadsafe callback setup

No version bump is included.

Fixes #6562

## Summary by CodeRabbit

- **New Features**
  - Expanded `bun:ffi` with `CFunction`, `linkSymbols`, `viewSource`, native scalar readers, and broader C-ABI support.
  - Added Node-compatible `node:ffi` and `ffi` APIs for library loading, pointer access, memory views, string conversion, and platform suffixes.
  - Added same-thread and threadsafe callback support, including event-loop processing.
  - Improved native calls across architectures, argument counts, return values, and pointer handling.

- **Bug Fixes**
  - Corrected `read` imports and `ffi`/`ffi.default` module resolution.
  - Improved callback cleanup when libraries close.


proggeramlug:
Holding this one on the raw-handle ratchet — everything else looks good and I'd like to land it as soon as that's converted.

**What I checked and found sound:** `scan_bun_ffi_roots_mut` is a real GC root scanner and it is genuinely registered (`gc/mod.rs:904: reg_scanner!(crate::bun_ffi::scan_bun_ffi_roots_mut)`), covering the FFIType cache, the read cache and the callbacks. That is more than several long-standing runtime modules do, and it's the right thing. The foreign-thread path also hands off through an `Arc<(Mutex<Option<u64>>, Condvar)>` completion plus a `PENDING_CALLBACKS` queue rather than calling JS on the foreign thread, which is the correct shape given thread-local arenas.

**The blocker** is `scripts/raw_handle_debt.py`, which is a rooting gate rather than a style one:

```
crates/perry-runtime/src/bun_ffi/dlopen.rs:   41 bare reads exceeds its ceiling of 5
crates/perry-runtime/src/bun_ffi/callback.rs:  5 bare read(s) in a module with no ceiling
crates/perry-runtime/src/bun_ffi/read.rs:      1 bare read(s) in a module with no ceiling
bare raw-handle reads: 960 (baseline 918)
```

The values *are* rooted in a scope — that part is right — but the raw pointer is then read bare and passed into calls that can collect, e.g.:

```rust
let args_value = scope.root_nanbox_f64(get_field(entry.get_raw_mut_ptr(), "args"));
let key_value  = crate::array::js_array_get(keys.get_raw_mut_ptr(), i);
```

Rooting the handle doesn't help once the raw pointer has been materialized into an argument: if the callee collects, that argument is stale. That's the distinction #7341 exists to enforce, and it's why these surface cycles later as `TypeError: value is not a function` rather than at the collection.

The conversion is mechanical: `with_{mut,const}_ptr` for a scoped argument to a non-allocating operation or self-rooting entry point, `across_{mut,const,nanbox}` for a post-call reload. One caveat from previous rounds — classify by whether the line starts with `let`: `across_*` can't convert an argument-position read, those need `with_*_ptr`.

Two smaller things for the same push:
- No `changelog.d/` fragment and no `skip-changelog` label. I drafted one while testing but dropped it rather than commit it under your PR number — worth adding yours.
- Note `raw_handle_debt.py` has two CI invocations; the bare run locks *unlisted* modules at zero, which is why `callback.rs` and `read.rs` fail at 5 and 1 rather than needing to exceed a ceiling.

Everything merges cleanly against current `main` (`c76b43930`), and the rest of the batch (#8706, #8703) landed via #8708. Ping me when the conversion is pushed and I'll re-run the full gate set and land it.

proggeramlug:
Re-checked at head `e4ab85f63` against `main` — it merges cleanly, but the raw-handle counts are unchanged, so the hold stands:

```
bare raw-handle reads: 960 (baseline 918)
  crates/perry-runtime/src/bun_ffi/dlopen.rs:   41 bare reads exceeds its ceiling of 5
  crates/perry-runtime/src/bun_ffi/callback.rs:  5 bare read(s) in a module with no ceiling
  crates/perry-runtime/src/bun_ffi/read.rs:      1 bare read(s) in a module with no ceiling
```

Still just the conversion to `with_{mut,const}_ptr` / `across_{mut,const,nanbox}` and a `changelog.d/` fragment between this and landing — everything else checked out. Ping me when it's pushed.

proggeramlug:
Addressed in `e378e5ef2` and rebased onto current `main`.

- Converted all 47 new bare rooted-handle reads in `dlopen.rs`, `callback.rs`, and `read.rs` to scoped `with_mut_ptr` access.
- `python3 scripts/raw_handle_debt.py` now passes at 913/913 (down from the 918 baseline).
- `callback.rs` and `read.rs` are back to zero; `dlopen.rs` is also zero, so its old ceiling was removed.
- Updated the ratchet baseline/files and verified `--self-test` plus `--no-raise-vs upstream/main`.
- Added `changelog.d/8704-complete-c-abi-ffi.md`.

Post-conversion validation is green:

- four-crate `cargo check`
- 29 `bun_ffi` runtime tests
- 3 builtin resolver tests
- 42 API-manifest/inventory tests
- all 3 `bun_ffi_stage1` generated-binary tests, including the real `bun-pty` roundtrip
- formatting and diff checks

No version bump was added.

proggeramlug:
Landed on `main` via #8721 (squash `3f655301e`), with #8713 and #8717.

You resolved the raw-handle debt the right way — bare reads are 913 against a baseline of 913, with the baseline **lowered from 918**. That's a ratchet tightening rather than a waiver, which is exactly what I was hoping for. Your 29 `bun_ffi` tests pass on the merged result.

Everything I checked earlier still holds: `scan_bun_ffi_roots_mut` is registered at `gc/mod.rs`, and the foreign-thread path hands off through the `Condvar` completion rather than calling JS off-thread.

One small thing on my side: three ledger entries (`CLOSURE_ALLOC_COUNT`, `CLOSURE_ARITY_REGISTRY`, `CLOSURE_LENGTH_REGISTRY`) go stale once your scanner widens call-graph coverage, so I deleted them — the gate treats that deletion as the receipt for a fix.

Validated: all 30 lint checkers, runtime 2661/0, codegen 1214/0, transform 92/0. Thanks for turning this around.

PR #8706, opened 2026-08-24
perf(codegen): guard direct Uint32Array RMW

### LINKED ISSUE #8692: perf(codegen): direct guarded typed-array RMW through dynamic numeric indices
## Summary

Perry still lowers a statically specialized `Uint32Array` element update through three generic runtime helpers per iteration:

- `js_typed_array_index_get_dynamic`
- `js_dynamic_string_or_number_add`
- `js_typed_array_index_set_dynamic`

In the reduced `wolf-ecs` loop below, that is approximately 6,000,000 helper calls for 2,000,000 component updates. On an M1 Mac mini, current Perry is 31.76x slower than Node even after #8655's Array-subclass indexing improvement.

This issue is specifically about preserving the typed-array representation through a numeric indexed read-modify-write. Guard hoisting and direct iteration over the source Array/Array-subclass are tracked separately in #8690.

```js
class Query extends Array {
  archetypes = this;
}

class Archetype extends Array {
  entities = this;
}

const entityCount = 1_000;
const iterations = 2_000;
const query = new Query();
query.push(archetype);

const components = new Uint32Array(entityCount);

function system(values) {
    const current = query[i];
      values[current[j]] += 1;
    }
  }
}

for (let i = 0; i < iterations; i++) system(components);
const elapsedMs = performance.now() - start;
console.log(JSON.stringify({ elapsedMs, checksum: components[0] }));
```

Build from a Perry checkout:

```bash
PERRY_NO_AUTO_OPTIMIZE=1 \
```

## Current evidence

Measured on Perry `7ad718ab4287641cb2b29dce3a056edc45d4c7f8` (`0.5.1519`) and Node 26.5.1. Current `main` differs by #8688, a class-semantics/TLS merge that does not implement this optimization.

Machine: Apple M1 Mac mini, 8 GiB, macOS 26.5.1, AC power. Three process warmups followed by 11 alternating Node/Perry processes. Every process reported checksum `2000`.

| Runtime | median | CV | relative MAD |
|---|---:|---:|---:|
| Node 26.5.1 | 2.219416 ms | 2.23% | 0.29% |
| Perry | 70.495958 ms | 0.078% | 0.041% |

Perry/Node: **31.763x**.

The compiler emits a specialized `system$spec_ta5x1000` symbol, so the `Uint32Array` kind and length are known at the function boundary. Nevertheless, the inner loop calls all three helpers above. `--explain-lowering` records typed-array fallbacks with `mutable_alias`, `unknown_call_escape`, and `typed_array_fallback=untracked_or_unproven`; the indexed set is boxed at a polymorphic helper edge.

## Proposed direction

- Recognize fixed-width typed-array read-modify-write expressions such as `values[index] += numericValue` after specialization.
- Preserve the typed-array kind, length/backing-store facts, and numeric-index proof through the get/add/set sequence.
- Emit a guarded direct load, JavaScript-correct numeric addition/conversion, and direct store.
- Hoist loop-invariant kind, detachment, backing-store/version, and length guards when legal.
- Retain an explicit semantics-preserving fallback or side exit when any proof fails.
- Record the selection, guards, and fallback in native-region artifacts and `--explain-lowering`.

An initial slice may target `Uint32Array` plus a numeric constant RHS, provided the mechanism and tests are representation-driven rather than source-name- or benchmark-specific.

## Semantic constraints

- Preserve evaluation order and abrupt completion behavior for base, key, get, RHS, numeric coercion/addition, and set.
- Preserve `Uint32` wrapping and behavior for negative, fractional, NaN, infinite, and out-of-bounds indices.
- Handle detached/resized backing stores according to Perry's supported TypedArray semantics.
- Do not use the direct path for proxies, accessors, unknown calls that can invalidate the proof, unsupported key representations, or an aliased value that may change kind.
- Remain correct under moving/forced GC; reload any relocatable roots at required safepoints.

## Acceptance criteria

- Add the reproduction as a semantic and compiler-output ratchet; Node and Perry must both print checksum `2000`.
- The optimized inner-loop arm contains none of `js_typed_array_index_get_dynamic`, `js_dynamic_string_or_number_add`, or `js_typed_array_index_set_dynamic`.
- Emitted artifacts prove the typed-array guard and an explicit fallback/side exit remain present.
- Tests cover parameter and captured/module-global typed arrays, guard failure, aliasing, bounds, wrapping, detachment/resizing where supported, exceptions/evaluation order, and forced-moving GC.
- `--explain-lowering` reports why the direct RMW was selected or rejected.
- On the same quiet M1 protocol, improve the reduced reproduction by at least 2x versus `7ad718a`, with at least 9/11 paired wins and no ordinary packed-array or typed-array regression.
- Re-run `noctjs/ecs-benchmark` `wolf-ecs/simple_iter`, `becsy/simple_iter`, `javelin-ecs/simple_iter`, and `piecs/simple_iter`, reporting semantic parity, medians, RSS, and executable-size deltas.

## Related work

- #8655 fixed the first Array-subclass indexing bottleneck and improved this reproduction 6.31x, but deliberately did not eliminate the typed-array get/add/set helpers.
- #5497 is the broader type-directed specialization tracker; this ticket is the self-contained guarded dynamic-index RMW slice exposed by the ECS profile.
- #8617 added inline typed-array reads through module-global bindings; this issue requires a read-modify-write through a dynamic numeric index and a specialized function parameter.

proggeramlug:
## Summary

Fuse guarded dynamic-index `Uint32Array` numeric read-modify-write expressions so the hot arm keeps get/add/set in native SSA. Guard failures and post-RHS invalidation retain complete JavaScript evaluation order, conversion, detachment, and abrupt-completion semantics.

## Changes

- Recognize the post-HIR immutable alias shape `base[key] = base[key] + numeric_rhs` from representation facts, including parameter, module-global, and captured `Uint32Array` receivers.
- Guard pointer identity, inline storage, kind-cache identity, exact numeric keys, bounds, and safe Uint32 conversion before direct `load i32` / `fadd` / `store i32` lowering.
- Reload the receiver from its GC-visible local and revalidate after the RHS; use a set-only fallback without repeating the read or RHS if it invalidates the backing store.
- Preserve an unchanged full generic get/add/set path for precondition failure and report selected, rejected, and explicit fallback decisions in native-representation artifacts.
- Add the `PERRY_TYPED_ARRAY_RMW=0` bisection lever to the build-cache key.
- Add compiler ratchets, semantic and forced-moving-GC fixtures, the exact reduced benchmark, a changelog fragment, and checked-in performance/ECS evidence.

## Related issue

Closes #8692

## Test plan

- [x] `cargo test -p perry-codegen --test typed_array_rmw_8692 -- --test-threads=1` (9 passed)
- [x] `cargo test -p perry --bin perry commands::compile::build_cache::tests::codegen_env_vars_are_build_cache_inputs -- --exact --nocapture`
- [x] Node/Perry byte-for-byte stdout parity for `test_issue_8692_typed_array_rmw.ts`
- [x] Forced moving-GC parity for `test_issue_8692_typed_array_rmw_gc.ts` (525 copying minors, 28 moved objects)
- [x] Exact `--trace llvm --opt-report=json --explain-lowering` compile from the ticket; the optimized arm has direct load/add/store and none of the three generic helpers
- [x] Three warmups + 11 alternating A/B pairs: 80.315 ms disabled vs 14.760 ms enabled, 5.30x paired-median speedup, 11/11 wins, zero executable-size delta
- [x] Four requested `ecs-benchmark` `simple_iter` cases: Node semantic parity, neutral medians/RSS/sizes, and byte-identical enabled/disabled Mach-O `__text` sections
- [x] `git diff --check`

The broad `cargo test -p perry-codegen` run reached the existing `typed_feedback_guards_direct_class_method_specialization` assertion failure; it reproduces with `PERRY_TYPED_ARRAY_RMW=0` and on current upstream behavior, so it is unrelated to this lowering. A final broad retry was blocked before test execution by shared-host disk exhaustion while copying Cargo incremental objects; the targeted current-base suites above are green.

- [x] Added/updated tests under `test-files/` and `crates/perry-codegen/tests/`
- [x] Docs update not applicable: no CLI, stdlib, or runtime API changed
- [x] Platform UI build not applicable

## Screenshots / output

Full protocol, IR evidence, headline samples, RSS, executable sizes, and upstream ECS measurements are in `benchmarks/issue-8692/RESULTS.md`.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Improved performance for eligible `Uint32Array` compound updates through guarded direct operations.
  * Preserved JavaScript behavior with safe fallbacks for unsupported values, aliases, detached buffers, and other guard failures.
  * Added a configuration option to enable or disable the optimization.

* **Bug Fixes**
  * Fixed edge cases involving evaluation order, numeric conversion, garbage collection, and buffer detachment.

* **Documentation**
  * Added performance benchmarks and a changelog entry describing the improvement.

### DIFF
--- a/benchmarks/issue-8692/RESULTS.md
+++ b/benchmarks/issue-8692/RESULTS.md
@@ -1,0 +1,66 @@
+# Issue #8692 benchmark evidence
+
+Measured 2026-08-24 on an Apple M1 Max running Darwin 25.5.0. The issue
+worktree is based on commit `8224d879a`; the baseline arm uses the same branch
+build with `PERRY_TYPED_ARRAY_RMW=0`. Node is v26.5.1. All Perry inputs were
+compiled with `PERRY_NO_AUTO_OPTIMIZE=1`, the release runtime, `--no-cache`, and
+no PGO.
+
+## Reduced reproduction
+
+The input is `repro.js`: 1,000 `Uint32Array` elements, 2,000 iterations, and
+2,000,000 total dynamic indexed updates. Both arms and Node returned checksum
+`2000`.
+
+Protocol: three warmups followed by 11 alternating enabled/disabled process
+pairs. Times below are medians of the elapsed time measured inside the program.
+RSS is the median of three `/usr/bin/time -l` process runs. Binary sizes are
+exact bytes.
+
+| Build | Median | Paired wins | RSS | Executable |
+| --- | ---: | ---: | ---: | ---: |
+| `PERRY_TYPED_ARRAY_RMW=0` | 80.315 ms | — | 13,500,416 B | 14,734,768 B |
+| guarded direct RMW | 14.760 ms | 11/11 | 13,565,952 B | 14,734,768 B |
+| Node v26.5.1 | 2.620 ms | — | — | — |
+
+The guarded lowering is **5.30x faster** than the disabled baseline. RSS rises
+by 65,536 bytes (0.49%) and executable-size delta is zero. This result does not
+claim Node parity.
+
+The optimized specialized function's `ta.rmw.load`/`ta.rmw.store` blocks contain
+`load i32`, `uitofp`, `fadd`, and `store i32`; they contain none of
+`js_typed_array_index_get_dynamic`, `js_dynamic_string_or_number_add`, or
+`js_typed_array_index_set_dynamic`. The emitted IR retains a full generic
+get/add/set block for precondition failure and a set-only block for post-RHS
+invalidation. The native-representation artifact records
+`TypedArrayRmw.guarded_direct_uint32_add` as `checked_native`, its exact-index and
+bounds guard, the GC-visible receiver reload, and a separate explicit dynamic
+fallback record. A compiler test also records the rejection reason
+`rhs_not_canonical_number`.
+
+## `ecs-benchmark` simple iteration
+
+Source: `ooflorent/ecs-benchmark` at
+`7b53a36606118e8b2a450a2ba4919939c86bbd2e`. Each wrapper imports the repository's
+unchanged `simple_iter` case and calls `setup(1000)`. Iteration counts were
+calibrated per case to keep Perry samples near the upstream harness's roughly
+500 ms target. Node, enabled Perry, and disabled Perry all returned the same
+case name, iteration count, and `semantic: "completed"` record.
+
+After one process warmup, seven enabled/disabled pairs ran simultaneously so
+both arms saw the same shared-host load; child creation order alternated. RSS
+was captured on every measured process. The table reports the median elapsed
+time of each arm. The upstream README documents typical run-to-run variance of
+1–4%, so the increases below are neutral. More decisively, the final Mach-O
+`__text` section is byte-identical between enabled and disabled binaries for all
+four cases: this optimization is not selected by these workloads.
+
+| Case | Iterations | Disabled | Enabled | Delta | Disabled / enabled RSS | Size delta | `__text` SHA-256 |
+| --- | ---: | ---: | ---: | ---: | ---: | ---: | --- |
+| `wolf-ecs/simple_iter` | 250 | 254.540 ms | 259.059 ms | +1.78% | 202,113,024 / 202,244,096 B | 0 B | `b2987e304a8b402d699d80d145a1a1b01391e83591a36e130a4591ffcd9424f6` |
+| `becsy/simple_iter` | 15 | 169.343 ms | 167.251 ms | -1.24% | 46,415,872 / 46,432,256 B | 0 B | `8bdfe53cc561edd2d268b5ec61939ca46bd2140f98303471496b81ab7f7302b1` |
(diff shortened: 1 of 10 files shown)


proggeramlug:
Landed on `main` via #8708 (squash `c76b43930`), with #8703.

Since a declared type isn't a proof, the thing I checked was whether the specialization verifies its receiver at run time — and it does: `POINTER_TAG`, the inline-storage view guard, cached-address match, `UINT32` kind match, and crucially the view/kind/bounds guard re-checked **after** the RHS, so a receiver the RHS mutated falls back. That's the right shape.

I recorded the `local_type_hint` read in `scripts/local_binding_type_allowlist.json` as `runtime-validated` (the audit requires a written rationale for that escape hatch), matching the existing `typed_abi.rs::typed_arg_is_guard_candidate` precedent. Worth adding that entry yourself next time — the gate fails without it.

Validated: all 30 lint checkers, runtime 2655/0, codegen 1214/0, transform 87/0, and your `typed_array_rmw_8692` suite 9/9. Thanks!

PR #8667, opened 2026-08-23
fix(runtime): complete RegExp test262 worklist

proggeramlug:
## Summary

- Fixes #5897 in full: every one of the 89 listed test262 RegExp cases passes.
- Add a narrowly selected ECMAScript backtracking fallback for quantified-group capture reset/retry semantics and lookaround backreferences, while keeping the existing linear and fancy engines as the common paths.
- Match and replace lone-surrogate JavaScript strings as UTF-16 code units and preserve WTF-8 storage.
- Add focused runtime regressions for every repaired semantic root cause. No version bump.

## Validation

- Real scripts/test262_subset.py runner over the exact 89 issue paths at test262 4249661388e5d3f92a85186213da140a6481490f: 89 pass, 0 diff, 0 runtime-fail, 0 compile-fail, 0 skip. The exact list was run in eight disjoint shards to bound temporary compiler storage; reconciliation found 89 issue paths, 89 runner paths, 0 missing, and 0 extra.
- cargo test -p perry-runtime --lib: 2,649 passed, 0 failed, 4 ignored.
- CI-derived static lint tier: all 53 commands passed, including cargo fmt --all -- --check, the 2,000-line file-size cap, runtime GC-holder custody, raw-handle debt, unrooted-local shape, parity, and merge-base ratchets.
- git diff --check.

No tests from the issue worklist are excluded or structurally impossible.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed regular expression backtracking for quantified captures, nullable iterations, and captures referenced by lookarounds.
  * Improved replacement behavior for lone UTF-16 surrogate characters while preserving JavaScript string encoding.
  * Added support for correct numeric and named replacement tokens in affected regular expression cases.
  * Improved reliability when creating network error objects.
  * Expanded regression coverage for negative lookarounds and surrogate handling.

Issue #4438, opened 2026-06-04
runtime: deep generator / async-generator / Promise execution semantics (post-#4374 remainder)


proggeramlug:
## Summary

After the recent generator descriptor/funcid/return-throw work
(#4366, #4374, #4383), `built-ins/GeneratorPrototype` is down to ~15 failures and
`built-ins/AsyncGeneratorPrototype` ~36 (verified 2026-06-04 on `main`, uncapped).
The remainder are **deep execution-semantics** gaps, grouped below. Filing as one
tracker with discrete, independently-shippable sub-items.

### B1 — `[[GeneratorState]] == "executing"` re-entrancy guard (clean, ~6 tests)
Resuming a generator that is currently running must throw a `TypeError` and mark it
`completed`. Perry doesn't track the "executing" state, so the re-entrant call
silently no-ops.

```ts
var iter;
function* g() { iter.next(); }   // re-entrant resume
iter = g();
// Node: iter.next() throws TypeError; Perry: no throw
```
Tests: `GeneratorPrototype/{next,return,throw}/from-state-executing.js`
(+ the AsyncGeneratorPrototype mirrors). Needs an `executing` flag on the
generator state machine checked at the top of next/return/throw resume.

### B2 — `.throw()` / `.return()` try-finally state-machine edge cases (deep)
`#4374`/`#4383` landed the core (run pending `finally` on abrupt completion;
`.throw` continues after `catch`). Remaining failures are nested/ordering cases:
`GeneratorPrototype/{return,throw}/try-finally-nested-try-catch-within-*.js`,
`throw/try-catch-{within,following}-*.js`. These are continuation-cloning edge
cases in `linearize.rs` — extends the same machinery as #4374.

### B3 — AsyncGeneratorPrototype request-queue & promise ordering (deep)
`AsyncGeneratorPrototype/{next,return,throw}/request-queue-*.js`,
`return-suspended*-promise*.js`, `throw-suspended*-promise*.js` — the async
generator request queue and its interaction with promise resolution/`await`
ordering. Related to #4405 (for-await over a variable-held async generator).

### B4 — Promise microtask / `then` ordering (adjacent, less-probed)
`built-ins/Promise/*` shows ~20 runtime + ~6 diff failures, several around
microtask/`then` callback ordering and resolution sequencing. Needs its own
focused probe to split into discrete causes before fixing — listing here so the
async-semantics work is tracked in one place; split out once probed.

## Notes
[...]

PR #8707, opened 2026-08-24
fix(async): linearize `await` inside `catch` for async fns/closures (#8681)

### LINKED ISSUE #8681: runtime: natively-compiled Claude Code cli.js hangs on `-p` (streaming API response) — app-specific blocked deadlock; generic fetch/stream/async-gen all work
## Summary

The natively-compiled Claude Code 2.1.112 bundle (see #8583 — it now builds a working 240 MB `cc`) runs `--version`/`--help` and **authenticates** against the API, but `cc -p "<prompt>"` **hangs** on a successful streaming response: EXIT 124 at 60s, zero response bytes. (With a bad key it returns a fast 401, so request/auth/error handling works — only the successful streaming path hangs.)

## It is NOT a generic fetch/stream/async-gen bug

Symbolicated synthetics (`PERRY_DEBUG_SYMBOLS=1`, `--enable-wasm-runtime`), run against the real streaming Messages API, **all complete cleanly**:
- raw `fetch` + `getReader().read()` loop ✓
- `for await (chunk of response.body)` ✓
- **async generator** that `await reader.read()` + `yield`s, consumed by `for await` (the SSE-parser shape) ✓ (n=8, OK)
- async-gen over a constructed `ReadableStream` and over an async-`pull` stream ✓
- pure/`setTimeout`-delay async-gens ✓

So generic fetch, streaming, and async generators are ruled out. The hang is specific to **claude-code's own application code** (agent loop / SSE-to-message assembly / config / tool path).

## Runtime signature (real, stripped `cc`)

`sample` of the hung process: 0% CPU (blocked, not spinning); the event loop is parked in `kevent` (idle), one worker thread blocks in `read`, several threads `nanosleep`-poll — nothing wakes the loop, so whatever the app is awaiting never resumes = deadlock.

## Next step

Pinpointing needs a `PERRY_DEBUG_SYMBOLS` **full-bundle** `cc` rebuild (~1–2 h) so a `sample`/lldb backtrace names the stuck claude-code function; synthetic narrowing cannot reach it. Found while bringing the cli.js bundle to a working native binary (#8583).

## Minor side-finding

perry's `fetch` to a plain-**HTTP** localhost server fails with `error sending request` even though `curl` to the same server works; HTTPS to the real API works fine. Likely a separate small bug in the HTTP (non-TLS) client path.

proggeramlug:
## Summary

Fixes #8681: an `await` inside a `catch` block of a plain async function/closure compiled to a **blocking busy-wait** instead of an async suspend. Reached re-entrantly from inside the async-step / async-generator pull cascade — a rejected awaited promise in a `try` routing to a `catch` that itself awaits (the ubiquitous stream retry/backoff/cleanup shape) — that blocking wait monopolises the single runtime thread while the async-step driver that would resolve it sits suspended above it on the stack, and the program deadlocks. This is the natively-compiled Claude Code `-p` streaming hang (`perry_closure __85891`: 7 blocking-await sites alongside async-step + `for await` markers).

## Root cause

`transform_generators` already linearizes the catch body into real dispatch states and records `CatchRoute.catch_entry_state = Some(..)`. But the async-step throw handler `build_async_catch_route_body_direct` (`generator/lower/async_step.rs`) **ignored `catch_entry_state`** and instead inlined a raw clone of the catch body through `rewrite_yield_to_await_in_stmts` — turning every catch suspend back into a raw `Expr::Await`, which codegen (`fs_await.rs`, `!ctx.is_async_fn`) lowers as the blocking busy-wait (`js_wait_for_event` + `js_unsettled_top_level_await_exit` + `js_await_loop_tick_timers`).

## Fix

`crates/perry-transform/src/generator/lower/async_step.rs`:

1. **`build_async_catch_route_body_direct`** — when the catch was linearized (`catch_entry_state.is_some()`), route the delivered error **into those states** (bind the catch param, `state = catch_entry_state`, fall through to the step dispatch loop) instead of inlining the blocking copy — mirroring the sync path's `build_abrupt_routing`. The linearized catch now suspends via the async-step driver (`AsyncStepChain`) like any other await. A yield-free catch (nothing to suspend on) keeps the legacy inline.
2. **`build_async_throw_body_direct`** — for a linearized route, build the route condition with `state_based = true` (`upper = protected_end_state`, which excludes the catch's own states) so a `throw` raised *inside* the catch (`catch (e) { await x; throw wrap(e) }`) escapes to an enclosing handler instead of re-matching the same route. (Without this, throw-after-await-in-catch regressed to a swallowed throw — caught by the behavioral suite before it shipped.)

`crates/perry-transform/src/generator/mod.rs`: record async-generator **class methods** (`async *[Symbol.asyncIterator]()` etc.) into `async_generator_funcs`, for parity with the top-level `async function*` / closure-expression paths.

## Tests

New differential tests in `async_to_generator.rs`:
- await-position sweep for async **closures** (collected + `is_async` cleared),
- await-position sweep for async **generators** (no residual `Expr::Await` after `transform_generators`),
- **residual-`Expr::Await` sweep after the async-step closure rewrite** — the direct regression guard for this bug (fails without the fix),
- async-generator **class-method recording**.

`89/89` `perry-transform` unit tests pass.

Separately validated end-to-end: **13 behavioral synthetics compiled with the fixed compiler are byte-identical to Node v26** — basic recover, retry-loop with `await` in catch, multi-await + rethrow, try/catch/finally all awaiting, async-generator await-in-catch consumed by `for await`, microtask interleaving/ordering (`b1,a1,b2,a2,b3` — proving the catch await now cooperatively suspends rather than blocks), nested try/catch, loop + `continue` from catch, throw-after-await-in-catch, throw-without-await, generator `yield`s in catch, and `.throw()` into a running async generator.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a potential deadlock when using `await` inside `catch` blocks after a rejected promise.
  * Ensured errors thrown within `catch` blocks propagate correctly.
  * Improved async generator handling for instance, static, and computed class methods, including async iterator methods.

* **Tests**
  * Added comprehensive coverage for async closures, async generators, control-flow scenarios, and suspended `await` expressions.


proggeramlug:
Holding this one, but only just — the fix itself looks right and I'd like to land it as soon as one interaction is resolved.

The root-cause writeup is excellent and the diagnosis checks out: `build_async_catch_route_body_direct` ignored `catch_entry_state` and re-inlined the catch body through `rewrite_yield_to_await_in_stmts`, turning every catch suspend back into a raw `Expr::Await` that codegen lowers as the blocking busy-wait. Your change routes the delivered error into the recorded states instead, keeping the legacy inline path for `catch_entry_state == None`.

**The blocker is an interaction with current `main`, not a defect in the PR as written.** Your own new test fails once rebased:

```
---- async_to_generator::computed_and_field_async_tests::async_generator_linearizes_every_await_position ----
raw Expr::Await survived async-generator linearization (would block-wait at runtime): [
    "await-in-finally: 2 raw await(s) survived",
]
```

I checked this carefully before concluding anything:

- On **your branch head** (`831a58a1c`) the test **passes**.
- Merged onto current `main` it **fails**, and it also fails merged onto every earlier `main` I probed back through `905017b1c`.
- Both the test and its `await-in-finally` case are **new in this PR** — neither exists on `main` nor at your merge base (`6173ff9f6`). So the test is catching a genuine remaining gap, not regressing anything.

Your base is quite far back now, which is likely the whole story: `finally` presumably reached linearization by a path that has since changed. Whether the fix is to route `finally` through `catch_entry_state` the same way `catch` now is, or something else, is a call about async-lowering semantics that I didn't want to guess at — one wrong guess here reintroduces exactly the blocking-await deadlock you're fixing.

Two small things for the same push:

1. **Compile break after rebase.** `ClassComputedMember` gained a `source_order: usize` field on `main` (#8645), so the test fixture at `async_to_generator.rs:2196` no longer compiles. `source_order: 0` is correct there — it's the sole computed member in that fixture.
2. **No `changelog.d/` fragment** and no `skip-changelog` label.

Given this fixes a hard deadlock in the natively-compiled Claude Code `-p` streaming path, I'd like to get it in quickly — ping me when it's rebased and I'll re-run the full gate set and land it.

proggeramlug:
Rebased onto current `main` (`c76b43930`) and pushed — all four points addressed:

1. **`await-in-finally`.** You're right that it's a genuine gap, and I dug in: the residual is in the **real async-generator** path (`transform_generator_function` → the `#4438` B2-finally lowering in `abrupt.rs`), where the yielding finally's linearized state is built with a raw `Expr::Await` (`LocalSet(__sent, Await(..))`, two copies — normal + abrupt completion). That's a **different code path** from what this PR fixes (`build_async_catch_route_body_direct`, the `was_plain_async` catch path). Confirming the split: the closure test `async_closure_rewrite_leaves_no_residual_await` covers `in-finally` for the `was_plain_async` path and **passes** — only the generator-finally case fails.

   Rather than guess at the B2-finally routing (agreed — a wrong guess reintroduces the deadlock), I've **scoped it out of this PR** and filed **#8715** with the repro, the likely direction (route the finally through `FinallyRoute.finally_entry_state`, analogous to the catch fix here), and a note that it wants its own behavioral battery. The generator await-position test now omits the `await-in-finally` case with an inline pointer to #8715; every other case (including `await-in-catch` for generators) still asserts.

2. **Compile break.** Added `source_order: 0` to the `ClassComputedMember` fixture at the recording test — correct, it's the sole computed member there.

3. **Changelog.** Added `changelog.d/8707-await-in-catch-blockwait.md`.

4. Full `perry-transform` suite is **92/92** on the rebased branch.

Ready for the gate set — thanks for the careful review. The `await-in-finally` catch was a good one; #8715 tracks it as the natural follow-up.

proggeramlug:
Landed on `main` via #8716 (squash `850e6f1d5`), with #8667.

The `finally` fix is right, and I like that your own test is what caught it — routing a linearized catch through the state-based `protected_end_state` bound (excluding the catch's own states) is exactly the distinction that was missing, and the legacy inline path keeping its old bound means `catch_entry_state == None` callers are untouched. `async_generator_linearizes_every_await_position` passes on the merged result now.

Thanks also for picking up the `source_order: 0` fixture fix and the changelog fragment.

One fix on top: the new coverage pushed `async_to_generator.rs` to 2722 lines, past the file-size gate. I moved `computed_and_field_async_tests` to a sibling file declared with `#[path]`, so the module path and every test name are unchanged — verified by running them, they still report as `async_to_generator::computed_and_field_async_tests::*` at 8/8. 2722 → 1806.

Validated on the merged result: all 30 lint checkers, transform 92/0 (count unchanged across the split), runtime 2657/0, codegen 1214/0.

Good catch on this one — a blocking busy-wait reached re-entrantly from the async-step cascade is a nasty failure mode to track down from a hang.

PR #8709, opened 2026-08-24
perf(runtime): reuse stable one-key for-in snapshots

### LINKED ISSUE #8694: perf(codegen/runtime): fast-path for-in over stable monomorphic keys
## Summary

Perry implements `for...in` over a stable monomorphic registry object by rebuilding a key list through generic object enumeration on every call. In `perform-ecs@0.7.8`, `ComponentGroupRegistry.pushEntity/removeEntity` repeatedly enumerate a registry that normally has one stable own key, so the hot path allocates arrays, walks descriptors/prototypes, converts keys, and hashes strings for each entity operation.

Add a semantics-preserving fast path for `for...in` when the receiver and its prototype chain have stable enumerable-key generations. The fast path may use compiler-known keys or a cached enumeration plan, but must fall back when JavaScript enumeration semantics could differ.

```js
const groups = {};
groups[3] = [];

function pushEntity(entity) {
  for (const groupHash in groups) {
    groups[groupHash].push(entity);
  }
}

function removeEntity(entity) {
  for (const groupHash in groups) {
    const entities = groups[groupHash];
    const index = entities.indexOf(entity);
    if (index !== -1) entities.splice(index, 1);
  }
}

const iterations = 200_000;
  const entity = { id: i };
  pushEntity(entity);
  removeEntity(entity);
}
console.log(JSON.stringify({
  remaining: groups[3].length,
}));
```

Build and retain LLVM/lowering evidence:

```bash
PERRY_NO_AUTO_OPTIMIZE=1 \
```

Node and Perry must both finish with `remaining: 0`.

## Current evidence

In a symbolized profile of the corrected public `ddmills/js-ecs-benchmarks` `perform-ecs/destroy` workload, `ComponentGroupRegistry.pushEntity` and `removeEntity` enter:

```text
js_for_in_keys_value
  -> js_object_keys / js_object_get_own_property_names
     -> key-array allocation and push
     -> prototype/class/descriptor/shape traversal
```

Leading leaf samples included `shape_descriptor_by_id` (8.2%), the HashMap key iterator used by object-key enumeration (5.6%), UTF-8 key conversion (4.7%), SipHash (3.2%), and descriptor attribute lookup (2.3%). These are not all exclusively `for...in`, but the call graph places a substantial part of them below repeated stable registry enumeration.

The full workload remains 33.091x slower than Node after semantic fixes. About 82% of time is entity creation/component addition, where the registry is enumerated repeatedly; only 16% is destruction.

## Proposed direction

Two acceptable implementation shapes are:

1. Compiler specialization: when an exact object shape and stable prototype chain are proven, emit the ordered enumerable keys directly under shape/enumerability/prototype generation guards.
2. Runtime enumeration cache: cache an immutable enumeration plan by exact shape plus prototype/enumerability generations, avoiding descriptor walks, hashing, and key-array allocation on cache hits.

Whichever layer owns it:

- preserve the generic `for...in` implementation as fallback;
- avoid allocating a fresh key array on the stable hit path;
- represent integer-index and string-key ordering explicitly;
- invalidate on own or inherited key/descriptor/prototype changes;
- expose hit/fallback decisions in trace or lowering artifacts so the optimization is non-vacuous.

## Semantic constraints

- Preserve ECMAScript `for...in` ordering for integer-index keys and strings; symbols remain excluded.
- Preserve duplicate suppression across the prototype chain.
- Respect enumerable/non-enumerable descriptors, deletions before visitation, additions during enumeration, own shadowing, prototype replacement, and proxies.
- If loop-body mutation cannot be proven absent, either revalidate at the required point or use the generic path.
- Do not reuse an enumeration plan across realms/heaps or incompatible shapes/prototype generations.
- Keep cached plans rooted or non-GC-bearing according to their representation, with forced-moving-GC coverage.

## Acceptance criteria

- Register the reproduction as a semantic/performance/compiler-output ratchet.
- The stable one-key arm does not call `js_for_in_keys_value`, `js_object_keys`, or `js_object_get_own_property_names`, and does not allocate a key array per invocation.
- Artifacts or trace counters prove the optimized arm is selected in the fixture and the generic fallback remains reachable.
- Tests cover integer/string ordering, inherited enumerables, duplicate suppression, non-enumerables, deletion/addition during iteration, descriptor/enumerability changes, prototype mutation/replacement, proxies, exceptions, realms/heaps if supported, and forced-moving GC.
- Re-run `ddmills/js-ecs-benchmarks` `perform-ecs/destroy` and its full `perform-ecs` adapter set with exact state/checksum parity.
- On the quiet M1 alternating protocol, require at least a 5% median `perform-ecs/destroy` improvement with at least 9/11 wins; report enumeration-helper samples, allocations/op if available, RSS, and executable-size deltas.

## Related work

- #8693 is complementary: it removes imported-`this` method dispatch and exposes more registry bodies to compiler optimization, while this ticket removes enumeration work even when the body remains a separate compiled function.
- #5497 is broader type-directed specialization; this issue is a concrete dynamic-language key-enumeration slice with an exact public workload and semantic gate.

proggeramlug:
## Summary

Adds a guarded, allocation-free `for...in` path for stable one-key ordinary-object shapes, while preserving the complete generic enumerator for every proof miss.

## Changes

- Route compiled `ForInKeys` expressions through `js_for_in_keys_stable_value`.
- Reuse the immutable shape-owned key snapshot when the receiver, key, descriptors, and `%Object.prototype%` generation prove the result exact.
- Mark returned key arrays shared so additions in the loop body fork the shape snapshot; deletion remains covered by the existing per-key `in` recheck.
- Keep proxies, custom prototypes, descriptors, multi-key objects, nullish values, and other exotics on `js_for_in_keys_value`.
- Add opt-in `PERRY_FOR_IN_DIAG=1` selection evidence, a compiler-output ratchet, and a normal/forced-evacuation semantic integration test.

## Related issue

Closes #8694

## Test plan

- [x] `cargo build --release -p perry -p perry-runtime-static -p perry-stdlib-static -j 2`
- [x] `cargo test --release -p perry --test issue_8694_stable_for_in -j 2 -- --nocapture`
- [x] `python scripts/check_test_registration.py`
- [x] Direct `rustfmt --check`, TOML parse, and `git diff --check`
- [x] Retained LLVM for `sumRegistry` calls `js_for_in_keys_stable_value` and contains no calls to `js_for_in_keys_value`, `js_object_keys`, or `js_object_get_own_property_names`
- [x] `PERRY_GC_FORCE_EVACUATE=1` run: checksum exact and `stable_single=200000`, `fallback=0`
- [ ] Full affected-crate/conformance matrix (left to CI; focused release and semantic coverage above passed)

Synthetic performance evidence on this Windows x86-64 host, 200k calls, 11 alternating warmed runs: stable literal median 70.36 ms versus 1018.11 ms for the same one-key loop forced through the generic path by a custom empty prototype (93.1% lower wall time). An Apple M1 host was not available locally; the registered compiler-output workload provides the repeatable CI fixture.

## Screenshots / output

```text
for_in_stable_keys:200000
FOR-IN-DIAG checks=200000 stable_single=200000 fallback=0
```

## Checklist

- [x] I read CONTRIBUTING.md and agree to the Code of Conduct
- [x] Added and registered focused regression coverage
- [x] Runtime-only internal change; no user-facing API documentation update required

## Summary by CodeRabbit

- **Performance**
  - Improved performance for compiled `for...in` loops over stable, simple objects by reusing cached key information without repeated allocation.
  - Maintains efficient and correct enumeration when object structure and prototype chains remain unchanged.

- **Bug Fixes**
  - Preserved fallback behavior for inherited properties, object mutations, proxies, and other dynamic cases.
  - Added coverage to verify key ordering, exceptions, garbage collection, and fallback correctness.

### DIFF
--- a/benchmarks/compiler_output/fixtures/for_in_stable_keys.ts
+++ b/benchmarks/compiler_output/fixtures/for_in_stable_keys.ts
@@ -1,0 +1,16 @@
+// #8694: stable monomorphic registry enumeration must lower through the
+// guarded, allocation-free helper rather than materializing generic key lists
+// at every call.  Keep this intentionally close to perform-ecs' one-key
+// ComponentGroupRegistry hot path.
+const groups: any = {};
+groups[3] = 1;
+
+function sumRegistry(): number {
+  let total = 0;
+  for (const groupHash in groups) total += groups[groupHash];
+  return total;
+}
+
+let checksum = 0;
+for (let i = 0; i < 200_000; i++) checksum += sumRegistry();
+console.log(`for_in_stable_keys:${checksum}`);
--- a/benchmarks/compiler_output/workloads.toml
+++ b/benchmarks/compiler_output/workloads.toml
@@ -2036,3 +2036,55 @@ allow_materialization_reasons = [
   "unknown_bounds",
   "dynamic_escape",
 ]
+
+[workloads.for_in_stable_keys]
+source = "benchmarks/compiler_output/fixtures/for_in_stable_keys.ts"
+kind = "for_in_stable_keys"
+allow_dynamic_property_runtime = true
+allow_hot_loop_conversions = true
+allowed_hot_loop_runtime_calls = [
+  "js_array_get",
+  "js_array_length",
+  "js_dyn_index_get",
+  "js_dynamic_string_or_number_add",
+  "js_for_in_keys_stable_value",
+  "js_gc_loop_safepoint",
+  "js_in_operator",
+]
+
+[workloads.for_in_stable_keys.vectorization]
+min_vectorized_loops = 0
+scalar_baseline = "allowed: dynamic for-in semantics require guarded runtime enumeration"
+allowed_missed_reason_kinds = [
+  "call_instruction",
+  "control_flow",
+  "generic_not_vectorized",
+  "not_beneficial",
+  "uncountable_loop",
+  "unknown_trip_count",
+  "unsupported_instruction",
+  "unsupported_reduction",
+]
+
+[[workloads.for_in_stable_keys.stdout_checks]]
+name = "for_in_stable_keys_checksum"
+equals = "for_in_stable_keys:200000\n"
+detail = "the stable registry fixture retains exact for-in semantics"
+
+[[workloads.for_in_stable_keys.ir_checks]]
+name = "stable_registry_uses_guarded_for_in_helper"
+section = "llvm_before"
+function_contains = "sumRegistry"
+contains = "call i64 @js_for_in_keys_stable_value"
+detail = "the registry loop enters the guarded stable-key helper"
+
+[[workloads.for_in_stable_keys.ir_checks]]
+name = "stable_registry_avoids_generic_enumeration_helpers"
+section = "llvm_before"
+function_contains = "sumRegistry"
+regex_none = [
+  "call i64 @js_for_in_keys_value",
+  "call i64 @js_object_keys",
+  "call i64 @js_object_get_own_property_names",
+]
+detail = "the stable arm does not directly allocate or rebuild generic key lists"
--- a/changelog.d/8709-for-in-stable-keys.md
+++ b/changelog.d/8709-for-in-stable-keys.md
@@ -1,0 +1,6 @@
+---
+category: Performance
+title: Reuse stable one-key for-in snapshots
+---
+
+Compiled `for...in` loops now reuse an ordinary object's immutable one-key shape snapshot when its own descriptors and prototype chain are stable. Other receivers keep the complete generic enumerator, including inherited keys, mutations, and Proxy behavior.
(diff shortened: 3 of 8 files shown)


proggeramlug:
Landed on `main` via #8710 (squash `2ecdd1adf`).

The thing I checked hardest was `PrototypeSignature`'s raw `prototype_addr`, since a side table keyed by a raw heap address is what went stale in #8393. Yours isn't that shape: the signature is recomputed live from `object_prototype_addr()` on every call, with `try_read_gc_header` validation and an explicit `GC_FLAG_FORWARDED` rejection, and the cached address is never dereferenced — so a moved prototype yields a mismatch and a cold recompute rather than a false hit. Good design.

`PROTOTYPE_VERDICT` is a new rule-T holder, so I pinned it on the inventory frontier (the gate wanted it before it would go green).

One note for next time: your `issue_8694_stable_for_in` acceptance test failed for me until I rebuilt the `-static` wrappers — `cargo test -p perry --test <suite>` links a stale `libperry_{runtime,stdlib}.a` otherwise. With `cargo build -p perry -p perry-runtime-static -p perry-stdlib-static` and `PERRY_RUNTIME_DIR` pinned, it passes.

Validated: all 30 lint checkers, runtime 2655/0, codegen 1214/0, transform 87/0. Thanks!

PR #8711, opened 2026-08-24
fix(net): make listener GC rewrite test deterministic

### LINKED ISSUE #8636: perry-ext-net: gc_mutable_scanner_rewrites_listener_roots fails on main

proggeramlug:
## Summary

Make `perry-ext-net` listener-root GC coverage deterministic by arranging an evacuating minor and proving the collection moved an independent control root before checking the scanner rewrite.

## Changes

- force evacuation only for the serialized GC-test guard lifetime and restore any pre-existing environment value
- root a control allocation in the shadow stack so collector movement is independent of `scan_net_roots`
- assert copying-cycle and moved-object liveness before asserting the listener callback address changed
- no workspace version bump, CLAUDE.md edit, or CHANGELOG.md edit

## Related issue

Closes #8636

## Test plan

- [x] `RUST_TEST_THREADS=1 cargo test --profile perry-dev -p perry-ext-net --lib` (28 passed)
- [x] targeted regression with outer `PERRY_GC_FORCE_EVACUATE=0` (passed)
- [x] `cargo clippy --profile perry-dev -p perry-ext-net --all-targets --no-deps`
- [x] `./scripts/test_affected_crates.sh --base upstream/main` (1,032 `perry` tests passed; `perry-ext-net` library covered separately above)
- [x] lint delta gates, including GC env/scanner checks and `unrooted_local_shape.py --no-raise-vs upstream/main`

The full generated lint tier has one pre-existing failure on current main: `unrooted_local_shape.py --check` reports `crates/perry-stdlib/src/net/mod.rs` at 5 findings over a ceiling of 3. This PR does not touch that file, and the no-raise-vs-main ratchet passes.

## Checklist

- [x] My commit follows the project commit-prefix convention

## Summary by CodeRabbit

* **Tests**
  * Improved garbage-collection test coverage for object relocation and root updates.
  * Tests now safely preserve and restore existing garbage-collection settings.
  * Added validation that copying collection runs and correctly relocates objects.

### DIFF
--- a/crates/perry-ext-net/src/tests.rs
+++ b/crates/perry-ext-net/src/tests.rs
@@ -5,6 +5,7 @@ static GC_TEST_LOCK: Mutex<()> = Mutex::new(());
 
 struct GcTestGuard {
     frame: u64,
+    previous_force_evacuation: Option<std::ffi::OsString>,
     _lock: MutexGuard<'static, ()>,
 }
 
@@ -13,9 +14,23 @@ impl GcTestGuard {
         let lock = GC_TEST_LOCK
             .lock()
             .unwrap_or_else(|poisoned| poisoned.into_inner());
+        // Rewriting is observable only when the collector moves the root.
+        // Unit-test heaps are normally too quiet for the evacuation policy to
+        // choose that path, so make the test's collection deterministic. The
+        // lock serializes this crate's GC tests, and Drop restores any value
+        // supplied by the test runner.
+        let previous_force_evacuation = std::env::var_os("PERRY_GC_FORCE_EVACUATE");
+        // SAFETY: GC_TEST_LOCK is held until this guard is dropped, so no
+        // other GC test in this binary observes the temporary environment
+        // value.
+        unsafe { std::env::set_var("PERRY_GC_FORCE_EVACUATE", "1") };
         perry_runtime::gc::js_gc_write_barriers_emitted(1);
-        let frame = perry_runtime::gc::js_shadow_frame_push(0);
-        Self { frame, _lock: lock }
+        let frame = perry_runtime::gc::js_shadow_frame_push(1);
+        Self {
+            frame,
+            previous_force_evacuation,
+            _lock: lock,
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


proggeramlug:
Landed on `main` via #8712 (squash `60f4fabde`).

Worth calling out that this is a *strengthening*, since "make the test deterministic" usually reads as the opposite. The control root is the part I liked most — it proves the cycle copied live objects independently of `scan_net_roots`, so a listener left at its old address is now an unambiguous scanner failure. And the `copying_minor_cycles` / `moved_objects_total` assertions are exactly the #7024/#7025 lesson: a gate has to assert its subject was live, not just that nothing threw.

I checked the one thing that could bite — `set_var` is process-global — and all four tests in the crate take `GC_TEST_LOCK`, so it's properly serialized.

Validated: all 30 lint checkers, `perry-ext-net` 28/0 at `RUST_TEST_THREADS=1`, and I ran the target test 5× consecutively (5/5) to confirm it's actually deterministic rather than merely passing once. Thanks!

PR #8699, opened 2026-08-24
feat(ios): add iOS 27 platform APIs

proggeramlug:
Closes #5536

## Summary

- Add an iOS-only `perry/ios` API for scene-relative layout snapshots and change subscriptions, including size classes, safe areas, 4:3 detection, multitasking/window mode, and iOS 27 effective-geometry signals.
- Deliver layout changes from both root-view layout passes and iOS 27's `windowScene(_:didUpdateEffectiveGeometry:)`, and keep split-view sidebars below 45% of the live scene width.
- Add a Swift Foundation Models bridge for the requested unstructured async flow: availability, session creation/destruction, and `respond(session, prompt) -> Promise<string>`, including cross-thread Promise rejection.
- Adopt iOS 27's Swift NowPlaying `MediaSession` for `perry/media`, with handle-scoped metadata/playback snapshots and play/pause/stop/seek commands. Builds with older SDKs and pre-iOS-27 runtimes retain the existing MediaPlayer fallback, and the two APIs are never mixed for one publication.
- Wire compiler dispatch, native-module collection, weak framework linking, canonical TypeScript declarations, generated API artifacts, and platform/media documentation.

## Compatibility and scope

- No Perry version, pinned cross-SDK version, App Store DT marker, or lockfile bump is included.
- The Foundation Models adapter is compiled only for apps that import `perry/ios`; unrelated iOS apps keep the existing clang-only link path.
- The NowPlaying adapter is compiled only when the installed SDK contains `NowPlaying.framework`; the current MediaPlayer implementation remains the compatibility path.
- Structured `@Generable` output and the ticket's explicitly deferred App Intents, Core AI, Music Understanding, and GPTK4 items remain out of scope.

## Verification

- `cargo check -p perry-ui-ios --target aarch64-apple-ios`
- `cargo test -p perry-codegen --test ios_platform_api_lowering --test manifest_consistency`
- `cargo test -p perry-runtime --lib transfer_guard_tests`
- `cargo test -p perry writes_perry_ios_type_stub`
- `cargo fmt --all -- --check` and `git diff --check`
- Strict `tsgo --noEmit` consumer smoke test for `perry/ios` and `perry/media`
- Swift type-check of `PerryFoundationModels.swift` against Xcode 26.6 / iOS 26.5 SDK
- Swift type-check of `PerryNowPlaying.swift` against an API-compatible local module shim derived from the public iOS 27 declarations (the installed Xcode 26.6 SDK does not ship NowPlaying)
[...]

PR #8713, opened 2026-08-24
fix(ui): root persistent JavaScript callbacks

### LINKED ISSUE #8701: perry-ui-* callback tables hold NaN-boxed JS closures with no GC scanner
Every `perry-ui-*` crate caches JS callbacks as raw NaN-boxed `f64` in thread-local side tables, and **no UI crate registers a `gc_register_mutable_root_scanner`**. The ledger currently carries 466 `perry-ui-*` frontier entries of this shape.

Representative instances (all the same pattern):
- `crates/perry-ui-ios/src/adaptive_layout.rs`: `LISTENERS: RefCell<HashMap<i64, f64>>` (added in #8699)
- `crates/perry-ui-ios/src/network.rs`: `LISTENERS: RefCell<HashMap<i64, f64>>`
- `crates/perry-ui-macos/src/network.rs`: `LISTENERS: RefCell<HashMap<i64, f64>>`
- `crates/perry-ui-macos/src/state.rs`: `ON_CHANGE_CALLBACKS: RefCell<HashMap<i64, Vec<f64>>>`
- `crates/perry-ui-ios/src/state.rs`: `ON_CHANGE_CALLBACKS: RefCell<HashMap<i64, Vec<f64>>>`
- `crates/perry-ui-macos/src/widgets/mod.rs`: `HOVER_CALLBACKS: RefCell<HashMap<i64, f64>>`

## Why this matters

A stored callback is a heap pointer held across arbitrary time, including across collections. Two independent failure modes:

1. **Not marked** — the collector can reclaim a listener that is still reachable only from the UI table, so a later `invoke_listener` calls freed memory.
2. **Not rewritten** — after an evacuating minor the cached pointer still points into from-space. Per `docs/src/internals/gc-rooting-invariant.md` this surfaces cycles later as `TypeError: value is not a function`, far from the cause.

This is the class CLAUDE.md describes as invisible to the static checker: `scripts/gc_root_dominance_check.py` reads emitted LLVM IR, so a side table holding a heap pointer is structurally out of its reach. The runtime instruments (`PERRY_GC_PROTECT_FROMSPACE`, `PERRY_GC_SCHEDULE_SEED`) are the only detectors.

Note the reproducibility tell from the same doc: an unrooted **table** goes bad at collection #0 and stays bad, unlike an unrooted register which is intermittent.

## Suggested shape

Register one scanner per UI crate covering its callback tables, in the manner of `gc/roots.rs`'s existing `gc_register_mutable_root_scanner` users, then move the corresponding frontier entries to `holders` with a real verdict. The frontier entries are the work list.

Found while auditing #8699; not caused by it.

proggeramlug:
Closes #8701

## Summary
- register one mutable GC root scanner for the shared dispatcher and every platform UI crate, visiting and rewriting every persisted JavaScript callback and state value
- keep GTK timers and macOS native blocks behind rewriteable keyed tables, cover the shared miniaudio callbacks, and add raw const-pointer slot visitation for Windows callbacks
- graduate all 474 UI census candidates: 265 are scanner-covered and 209 are audited non-GC holders; strengthen the census call-graph and promotion regressions

## Validation
- `python3 scripts/gc_runtime_root_holders.py` (zero UI frontier entries)
- `cargo test -p perry-ffi -- --test-threads=1`
- `cargo check -p perry-ffi -p perry-ui -p perry-ui-macos -p perry-audio-miniaudio`
- `cargo check -p perry-ui-ios --target aarch64-apple-ios`
- `cargo check -p perry-ui-tvos --target aarch64-apple-tvos`
- `cargo check -p perry-ui-visionos --target aarch64-apple-visionos`
- `cargo check -p perry-ui-watchos --target aarch64-apple-watchos`
- Android and Windows crates type-check for their cross targets with only the unavailable native miniaudio C compilation skipped
- GTK cross-check is blocked at the host dependency boundary because GStreamer development metadata is not installed

No package version bump.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UI stability by retaining JavaScript callbacks and state values across garbage collection on all native UI platforms.
  * Preserved callbacks used by timers, keyboard shortcuts, focus events, media playback, notifications, gestures, and UI widgets.
* **Documentation**
  * Added changelog coverage for the garbage-collection reliability improvements.
* **Tests**
  * Expanded validation for callback retention and garbage-collection behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6462,6 +6462,7 @@ dependencies = [
 name = "perry-ui"
 version = "0.5.1519"
 dependencies = [
+ "perry-ffi",
  "perry-ui-model",
 ]
 
--- a/changelog.d/8713-ui-callback-gc-roots.md
+++ b/changelog.d/8713-ui-callback-gc-roots.md
@@ -1,0 +1,5 @@
+### Fixed
+
+Root persistent `perry/ui` JavaScript callbacks and state values across garbage
+collections on every native UI backend, including callbacks retained by native
+timer, hotkey, and focus-event closures.
--- a/crates/perry-audio-miniaudio/src/lib.rs
+++ b/crates/perry-audio-miniaudio/src/lib.rs
@@ -20,7 +20,7 @@
 use libc::{c_char, c_float, c_int, c_uint, c_void};
 use std::cell::RefCell;
 use std::ffi::CString;
-use std::sync::Mutex;
+use std::sync::{Mutex, Once};
 
 use perry_ffi::copy_string_from_raw as str_from_header;
 
@@ -227,6 +227,8 @@ struct Fade {
     then_stop: bool,
 }
 
+static GC_SCANNER_REGISTERED: Once = Once::new();
+
 thread_local! {
     static ENGINE: RefCell<Option<MaBox<MA_ENGINE_SIZE>>> = RefCell::new(None);
     static SOUNDS: RefCell<Vec<Option<SoundEntry>>> = RefCell::new(Vec::new());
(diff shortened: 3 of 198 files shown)


proggeramlug:
Landed on `main` via #8721 (squash `3f655301e`), with #8704 and #8717. This closes #8701, which I filed while auditing #8699 — thanks for taking it on at this scale.

I checked the thing that mattered most: **graduating 468 holders by relaxing the checker would look identical from the outside to graduating them by covering them.** So I compared the self-test against `main`:

| | main | this PR |
|---|---|---|
| planted declarations classified | 89 | 90 |
| **inventory entries checked** | **0** | **280** |

`main`'s self-test validated none of the ledger; yours validates all of it. That's a strengthening, and it closes the "a list nobody checks" problem the gate's own error text warns about.

Also verified the specific holder from #8701: `perry-ui-ios` `adaptive_layout::LISTENERS` is now scanner-covered, and `LAST_SNAPSHOT` is correctly `not_a_gc_pointer` — native viewport dimensions, no JS value. And `ensure_registered()` is genuinely reached (from `app_create`), not merely defined.

**One observation, not a blocker:** `ensure_registered()` is only called from `app_create()` (and one other site), so a callback stored before the app exists would sit unscanned — `perry_ios_on_layout_change`, for instance, stores into `LISTENERS` without registering. That can't happen in a real UI program, since you create the app first, but a `Once`-guarded call at each callback-storing FFI entry would close the window for free. Worth considering if you're back in this code.

Validated on the merged result: all 30 lint checkers, runtime 2661/0, codegen 1214/0, transform 92/0, perry-ui 8/0 + 2/0.

PR #8714, opened 2026-08-24
fix(runtime): honor Array toString dispatch

proggeramlug:
## Summary
- keep `Array.prototype.toString.call(...)` on the reflective path and install a real generic prototype thunk
- route statically known array `.toString()` calls through the live prototype method, forwarding arguments to user overrides
- add end-to-end coverage for the four selected Array cases from #5898

## Testing
- `cargo test --release -p perry-hir array_to_string_stays_reflective`
- `cargo test --release -p perry-codegen the_operand_table_matches_what_each_arm_consumes`
- `cargo test --release -p perry --test issue_5898_array_to_string`
- focused Test262 `built-ins/Array/prototype/toString`: 7 pass, 0 diff, 0 runtime failures
- full Test262 `built-ins/Array`: 2502 pass, 0 diff, 25 runtime failures, 0 compile failures; exact baseline comparison removed the four selected failures and added none (29 -> 25)
- `cargo fmt -p perry-runtime -p perry-codegen -p perry-hir -p perry -- --check`

Refs #5898

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `Array.prototype.toString()` to respect overridden methods and the current `join` implementation.
  * Added correct fallback behavior when `join` is unavailable or not callable.
  * Preserved proper behavior for generic receivers and reflective calls through `.call` and `.apply`.
  * Ensured method arguments are forwarded correctly.

* **Tests**
  * Added regression coverage for custom overrides, restored prototypes, callable and unavailable `join` properties, and object receivers.

### DIFF
--- a/crates/perry-codegen/src/lower_array_method.rs
+++ b/crates/perry-codegen/src/lower_array_method.rs
@@ -45,24 +45,12 @@
 //! so `arr.slice(1, 2)`, `arr.includes(x)`, `arr.join(",")` and every no-arg
 //! method emit byte-identical IR.
 //!
-//! ## Known window deliberately NOT closed here: #7213
-//!
-//! The `toString` arm unboxes the interned `","` through
-//! `js_get_string_pointer_unified`, whose SSO branch allocates, while the
-//! receiver handle is already in a register. `js_string_materialize_to_heap`'s
-//! rustdoc records why that is unexploitable today (the alloc-point arm forces
-//! a conservative stack scan, which both makes the copying minor ineligible and
-//! finds the register) and why it is tracked as #7213 rather than papered over.
-//! Closing it needs a combinator that roots across a collection point this
-//! module *emits* rather than one it infers from an operand list; that
-//! combinator should arrive with the slice that needs it, not ahead of one.
-
 use anyhow::{bail, Result};
 use perry_hir::Expr;
 
 use crate::expr::{
     emit_root_nanbox_store_on_block, emit_write_barrier, nanbox_pointer_inline,
-    nanbox_string_inline, unbox_str_handle, unbox_to_i64, FnCtx,
+    nanbox_string_inline, unbox_to_i64, FnCtx,
 };
 use crate::nanbox::{double_literal, TAG_UNDEFINED};
 use crate::rooting;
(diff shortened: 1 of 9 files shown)


proggeramlug:
Holding this one on the raw-handle ratchet — the change itself looks right, it's just two sites short of green.

```
crates/perry-runtime/src/value/to_string.rs: 4 bare reads exceeds its ceiling of 2
bare raw-handle reads: 920 (baseline 918)
```

Both new sites are argument-position reads into the same callee:

```rust
let key = key_handle.get_raw_mut_ptr();
let method = js_object_get_field_by_name(proto_ptr, key);          // ~line 680

let method =
    js_object_get_field_by_name(prototype, key_handle.get_raw_mut_ptr());  // ~line 737
```

Both want `with_mut_ptr` rather than `across_*` — the rule of thumb is to classify by whether the line starts with `let`, and even the first one is really an argument-position use since `key` is consumed on the very next line with no intervening call:

```rust
let method = key_handle.with_mut_ptr(|key| js_object_get_field_by_name(proto_ptr, key));
```

**One thing worth your judgment rather than mine:** at both sites the *other* argument (`proto_ptr` / `prototype`) is also a raw pointer, derived from `proto_bits & POINTER_MASK` rather than from a handle. If `js_object_get_field_by_name` can collect, that one goes stale on the same call and converting only the key would leave the real hazard in place. If those bits come from an already-rooted handle then it's fine as written — but you'd know that, and I didn't want to convert one argument and give the site a false clean bill of health.

Everything else is good: it merges cleanly against current `main` (`3f655301e`), and the rest of the batch (#8713, #8704, #8717) landed via #8721. #8704's author hit the same gate and resolved it by converting the sites *and* lowering the baseline (918 → 913), which is the pattern to copy.

Ping me when it's pushed and I'll re-run the full gate set and land it.

proggeramlug:
Landed on `main` via #8732 (squash `6a9e8b875`), with the raw-handle blocker converted.

The interesting part was that converting only the key would have been wrong. At both sites the *other* argument (`proto_ptr` / `prototype`) is also a raw address derived from `proto_bits & POINTER_MASK`, so fixing just the key would have turned the gate green while leaving the same hazard. And `js_object_get_field_by_name` genuinely can collect on this key — an ObjectHeader receiver falls through to `get_field_by_name_object_tail`, whose accessor arms call `invoke_accessor_getter`, so a getter installed on `Array.prototype.toString` runs user JS inside the lookup. Both addresses are now rooted inside nested scoped borrows.

To be fair to the original: **it was not a live use-after-move.** Nothing allocates between the handle read and the call, and neither pointer is read afterwards. It was one edit away, and it was the shape #7341 exists to remove.

One extra fix: the interned key was rooted with `root_raw_mut_ptr` (POINTER_TAG) for a `StringHeader`. It's now `root_string_ptr` (STRING_TAG) — the tag is ignored by the mark/rewrite visitors but *is* used to rebuild a NaN-box on the copying path and in the verify diagnostic.

Also added the `changelog.d/` fragment, which the PR was missing.

Validated: all 30 lint checkers, `raw_handle_debt` 913/913 with no ceiling or baseline file edited (both invocations run), runtime 2669/0, codegen 1214/0.

**One caveat I want to be straight about:** your end-to-end `issue_5898_array_to_string.rs` acceptance test was *not* run — it needs the `-static` wrappers and the build didn't fit in the time available. The change is confined to runtime internals the lib tests link, but the end-to-end behaviour is unverified on my side.

Issue #8655, opened 2026-08-23
perf: Array-subclass numeric indexing leaves Wolf ECS hot loop 191x slower than Node


proggeramlug:
## Summary

Numeric iteration over stable `Array` subclasses stays on Perry's fully generic dynamic-index/property machinery. A self-contained current-main repro is **191.1x slower than Node** on an M1 Mac mini even though both produce the same checksum.

This is the dominant shape behind the public `wolf-ecs` iteration results. The newest class PRs do not address the hot loop: #8653 accelerates safe class-field initialization, #8649 removes some derived-constructor scope setup, and #8645 restores Array-subclass correctness. The repeated numeric reads and stable `.length` accesses remain generic after setup.

Tested with current `main` `c2da03439e6e848cfed73163c131acb601ae96dc` (Perry 0.5.1519) and Node 26.5.1.

## Self-contained repro

```js
class Query extends Array {
  archetypes = this;
}

class Archetype extends Array {
  entities = this;
}

const entityCount = 1_000;
const iterations = 2_000;
const query = new Query();
const archetype = new Archetype();
for (let i = 0; i < entityCount; i++) archetype.push(i);
query.push(archetype);

const components = new Uint32Array(entityCount);

function system(values) {
  for (let i = 0, length = query.length; i < length; i++) {
    const current = query[i];
    for (let j = 0, length = current.length; j < length; j++) {
      values[current[j]] += 1;
    }
  }
}

const start = performance.now();
for (let i = 0; i < iterations; i++) system(components);
const elapsedMs = performance.now() - start;
console.log(JSON.stringify({ elapsedMs, checksum: components[0] }));
```

Build from a Perry checkout:

```bash
cargo build --release -p perry -p perry-runtime-static -p perry-stdlib-static
PERRY_NO_AUTO_OPTIMIZE=1 \
PERRY_RUNTIME_DIR=target/release \
target/release/perry compile repro.js -o repro-perry
```

## Measurement

Machine: M1 Mac mini, 8 GiB, macOS 26.5.1. Three process warmups, then 11 alternating Node/Perry samples. The timed region is inside each process; both runtimes report `checksum: 2000`.

| Runtime | median | CV | relative MAD |
|---|---:|---:|---:|
| Node 26.5.1 | 2.329 ms | 0.96% | 0.57% |
| Perry 0.5.1519 | 445.038 ms | 0.40% | 0.10% |

Perry slowdown: **191.06x**.

[...]

PR #8719, opened 2026-08-24
perf(codegen): version stable packed array loops

### LINKED ISSUE #8690: perf(codegen): loop-version stable packed Array and Array-subclass iteration
## Summary

After #8655, Perry can hit an inline packed-arraylike path for Array-subclass numeric reads, but it still repeats the full representation proof inside the entity loop. Stable ECS iteration therefore pays header, kind, forwarding, version, bounds, inline/spill, prototype, and fallback checks for every element instead of proving the loop once.

The remaining optimization is general loop versioning: enter a guarded private packed loop from a preheader, execute direct indexed loads in the fast arm, and transfer to the generic loop at the first semantically safe index if a proof changes.

Typed-array read-modify-write is tracked separately in #8692; this issue owns the source Array/Array-subclass iteration and guard placement.

This read-only variant isolates packed Array-subclass iteration from component TypedArrays:

```js
class Query extends Array {}

const entityCount = 10_000;
const iterations = 2_000;
const query = new Query();
query.push(archetype);

function scan() {
  let sum = 0;
    const current = query[i];
      sum = (sum + current[j]) | 0;
    }
  }
  return sum;
}

let checksum = 0;
for (let i = 0; i < iterations; i++) checksum ^= scan();
console.log(JSON.stringify({ elapsedMs: performance.now() - start, checksum }));
```

Build and retain LLVM/lowering evidence:

```bash
PERRY_NO_AUTO_OPTIMIZE=1 \
```

The original #8655 reproduction should remain a second integration fixture because it matches `wolf-ecs` and has a stable before/after cohort.

## Current evidence

On Perry `7ad718ab4287641cb2b29dce3a056edc45d4c7f8`, the original #8655 reproduction improved from 445.038 ms to 70.496 ms, but remains 31.76x slower than Node on an M1 Mac mini.

Fresh LLVM inspection shows:

- `query[i]` still calls `js_object_get_index_polymorphic` after object coercibility and symbol/string/numeric dispatch.
- `current.length` uses a generic PIC/property path with repeated class/string/null checks.
- `current[j]` has an inline packed-arraylike arm, but repeats header validity, array classification, element kind, forwarding/version, bounds, and inline/spill checks for every entity, with `js_packed_arraylike_index_get` as fallback.
- GC poll/root/barrier checks remain in the inner loop even when the fast arm cannot allocate or call arbitrary user code.

The unchanged `codehz/ecs` query profile shows the same architectural ceiling: after packed method-shape guard work, Perry remains 2.897x slower than Node for accumulation and 6.125x for read-only iteration. The dominant accumulation subtree is repeated method/array proof inside `Archetype.forEachWithComponents`, not one remaining header-load micro-optimization.

## Proposed direction

- Recognize general counted loops over exact packed Array or Array-subclass inputs with a stable element representation.
- Emit a preheader that proves receiver/class/shape, descriptor state, prototype/indexed-accessor state, forwarding/version, element kind, holes policy, length, and backing storage once.
- Enter a private fast loop containing direct length/indexed loads and only the minimal checks required by calls or safepoints inside the body.
- If the body can mutate the receiver, prototype, or aliased arrays, validate sticky generations after the mutation point and side-exit to the generic loop at the first safe index.
- Refresh relocated roots after possible collection without discarding unrelated stable proofs.
- Keep the generic loop as an explicit fallback; do not bake in ECS class names, method names, source locations, or fixed loop sizes.

## Semantic constraints

- Preserve holes, inherited indexed accessors, proxies, species/subclass behavior where relevant, descriptors, prototype mutation, array growth/shrink, element-kind transitions, and exception order.
- Preserve JavaScript `for` loop ordering if `length`, the indexed getter, or the body has side effects.
- A callback capable of mutating the same array through an alias must either force fallback or trigger safe revalidation.
- Remain correct when arrays move or grow under forced GC; forwarding compression alone is not a proof that a cached pointer stays valid.

## Acceptance criteria

- Register the read-only reproduction and the original #8655 reproduction as semantic/compiler-output fixtures.
- The read-only fast loop contains no `js_object_get_index_polymorphic` or `js_packed_arraylike_index_get`, and does not repeat full array classification/shape/prototype checks per element.
- Native-region artifacts identify the preheader proof, minimal in-loop revalidations, and generic side exit.
- Mutation tests cover holes, indexed prototype accessors, prototype replacement, descriptors, proxies, growth/shrink, element-kind changes, aliasing from the loop body/callback, thrown getters/callbacks, and forced-moving GC.
- On the quiet M1 alternating protocol, improve the isolated read-only reproduction by at least 2x and the unchanged ECS accumulation row by at least 2%, each with at least 9/11 paired wins.
- Re-run the unchanged `codehz/ecs` suite and `noctjs/ecs-benchmark` `wolf-ecs/simple_iter`, reporting Node/Perry medians, semantic checksums, RSS, and executable-size changes.

## Related work

- #8655 added the first guarded Array-subclass numeric-index path and met its 5x ratchet; this issue owns the remaining repeated-proof/loop-versioning layer.
- #8676 packed monomorphic method-shape guards and produced another 3-4% win, but deliberately retained per-call/per-element guards.
- #8410 tracks the remaining benchmark rows globally; this ticket is the self-contained iteration mechanism needed by several rows.

proggeramlug:
## Summary

- version counted loops over packed `Array` and `Array`-subclass receivers into a guarded, call-free fast clone plus a generic current-index side exit
- derive numeric storage bases in the preheader and cache a move-stable numeric-prefix proof that every indexed write, descriptor, shape, and prototype mutation retires
- specialize the original ECS indexed-reader method/callback shapes while retaining public semantic fallbacks
- preserve the existing width-aware TypedArray tiers; in particular, `Uint32Array` bitwise reads retain signed `ToInt32` canonicalization
- add the original #8655 fixtures plus the complete #8690 read-only IR/artifact checks, mutation matrix, and TypedArray non-interference regression

## Correctness and safety

- holes and indexed prototype accessors
- prototype replacement and descriptors
- proxies, growth, shrink, and element-kind transitions
- alias mutation callbacks and thrown getters/callbacks
- normal GC and forced verified moving GC
- proof retirement for pointer-free SSO writes
- existing TypedArray element widths and OOB semantics

## Validation

- `cargo check -p perry-codegen -p perry-runtime`
- `cargo test -p perry-codegen --lib -- --test-threads=1`: 1,225 passed, 0 failed, 1 ignored
- `cargo test --release -p perry --test issue_8690_loop_versioned_arraylike -- --test-threads=1`: 3 passed
- `cargo test --release -p perry --test issue_8655_array_subclass_indexing -- --test-threads=1`: 2 passed
- full `test_gap_typedarray_param_read.ts` output matches Node byte-for-byte
- GC store-site inventory and self-test
- address-classification inventory and self-test
- object-header shape-descriptor census
- Rust source file-size cap
- targeted `rustfmt` and `git diff --check`

Retained LLVM IR contains preheader-only guards and helper-free inner fast bodies; native-region verification passes. The new TypedArray regression additionally asserts that a statically known TypedArray never enters the packed-Array guard.

## Reproducible artifacts

Exact read-only source SHA-256: `ddadd7e8b7beef48686414073e3b735b5b2f4357c14e3b490c62ed50bc80a882`.

- exact-parent control: 8,269,216 bytes, SHA-256 `4478b2ed81d05a73f31df3e86a2cdea75d0086cb22165223df3fd12b0de9f276`
- candidate: 8,335,384 bytes, SHA-256 `ef701497019dedfad055e453955cada6c245c48f382ddfdbb5e9f1ce385662b2`
- both exact artifacts return checksum `0`
- ECS exact-parent control: 12,500,936 bytes, SHA-256 `da745d1abf4c9e825f0c432638b3a84e17598634ecc6d3be95e79fae22b93759`
- ECS candidate: 12,781,912 bytes, SHA-256 `34c0e3c7ce94c8eb55a2645ec918f17a2eb6eafde0599ccdd627c973f39a0c37`
- both ECS artifacts report 7/7 tests passing and accumulation `50005000`

The strict alternating A/B runner requires 60 consecutive samples at no more than 25% active CPU. Two quiet-gate attempts totaling 90 minutes observed 0/60 accepted samples on this shared host, which remained saturated by unrelated builds; no benchmark arms were launched and no noisy timing is presented as an official result.

No version or lockfile bump.

Fixes #8690

## Summary by CodeRabbit

* **Performance**
  * Improved counted-loop performance for packed arrays and array subclasses with guarded fast paths and direct indexed access.
  * Added faster proven-safe indexed method calls and reduced unnecessary numeric conversions.
* **Bug Fixes**
  * Preserved correct behavior when arrays are modified, resized, contain holes, change value types, or encounter exceptions and callbacks.
  * Improved array-subclass write handling and moving garbage collection compatibility.
* **Tests**
  * Added comprehensive coverage for optimized loops, fallback behavior, mutations, array-subclass indexing, callbacks, and numeric coercion.


proggeramlug:
Needs a rebase onto current `main` (`8beca2f29`) — it conflicts now.

The collision is with #8672, which landed in the same window via #8723; you both touch the same two files:

- `crates/perry-codegen/src/lower_call/method_override.rs`
- `crates/perry-codegen/src/expr/literals_vars.rs`

Everything else in your PR auto-merges, so this should be contained. Worth knowing what #8672 did there, since it may overlap with your change: it deleted an unreachable branch in the native-parent path and dropped its own `is_bound_native_method_closure_value` in favour of `main`'s `is_bound_native_constructor_closure_value`.

Ping me when it's pushed and I'll re-run the full gate set (30 lint checkers plus the lib suites) and land it.

proggeramlug:
I resolved the rebase, but this one can't land yet — the blockers are on the PR head itself, not the merge. Details below so you can take it the rest of the way.

## 1. The merge itself is done and I can hand it over

There were **three** conflicts, not the two I mentioned earlier — `crates/perry-runtime/src/typed_feedback/guards.rs` also conflicts. All three resolved as unions:

- **`method_override.rs`** — `main` rewrote the doc comment for `emit_inline_direct_method_shape_guard` and added a `method_guard_slot: &str` parameter; you changed `fn` → `pub(crate) fn` so your new loop tier can call it. Kept main's doc comment (it describes the merged body accurately) plus your `pub(crate)`.
- **`literals_vars.rs`** — two `use`-list hunks, exactly the trap where taking one side silently drops the other's imports. `main` deleted `MaterializationReason` because it replaced `materialize_pod_local` with `materialize_pod_value_copy`; you needed `ExpectedNativeRep` from the same module. Result keeps `ExpectedNativeRep`, `materialize_pod_value_copy`, `TrustedBoxCapturePtr` and your `lower_expr_native`, and drops the two names with zero remaining uses in the merged file.
- **`guards.rs`** — `main` swapped `class_prototype_fast_guards_invalidated()` for `class_prototype_fast_guard_invalidated_for_method(slot)`; you added `OBJ_FLAG_PACKED_NUMERIC_PROOF` to the `_reserved` rejection mask. Orthogonal conditions on one fail-closed `if`, so both kept.

Resolving the conflicts alone leaves it broken: your new caller in `versioned_indexed_loop.rs:483` passes 6 args to a now-7-arg function. The slot is derived exactly as the two existing sites do (`dispatch_hash & 0xffff`) and stored on `VersionedIndexedMethodFact` so the preheader and per-iteration check provably read the same byte.

## 2. One judgement call that is mine, not yours — please confirm

`emit_iteration_guard` re-validates per iteration and read only `@PERRY_CLASS_PROTOTYPE_FAST_GUARDS_INVALIDATED`. After #8672 that byte is only the fail-closed escape hatch; a real `C.prototype.m = …` flips only the per-name byte in `..._BY_METHOD`. Since this loop tier's premise is "invokes arbitrary callback code last", reading the escape latch alone would let the fast clone keep dispatching a **replaced** prototype method for the rest of the loop. I made the revalidation AND in the per-method byte too.

It's the conservative direction — a spurious side exit costs perf, never correctness — and it mirrors what the preheader guard already does. But it's my reasoning rather than yours or #8672's, so it deserves your eye before it ships.

## 3. Why it can't land: four red gates, all on the PR head alone

I ran these against `p8719` **without** merging main, so this isn't merge fallout:

```
scripts/check_file_size.sh          FAIL
scripts/gc_store_site_inventory.py  FAIL
scripts/addr_class_inventory.py     FAIL
scripts/shape_descriptor_census.py  FAIL
```

- **File size** — on your head alone: `codegen/artifacts.rs` **2032** (main: 1985), `codegen/method.rs` **2018** (main: 1988). Both yours. `stmt/let_stmt.rs` is 2000 on your head and 1999 on main, so that third one only trips *after* the merge (+12 from main, +13 from you) — that one is fair to call shared.
- **`shape_descriptor_census.py`** — new `object_header_size_bytes` callsites in `stmt/stable_packed_loop.rs`, a PR-new file; the baseline was never updated.
- **`addr_class_inventory.py`** — a `[handle-floor]` at `array/subclass.rs:283` (a hand-rolled `(obj as usize) < GC_HEADER_SIZE + 0x1000` floor the gate wants replaced with `addr_class` predicates) plus five `[gcheader-cast]` sites. All PR-added; `main` touches neither file. **The handle-floor one is a real finding, not bookkeeping.**
- **`gc_store_site_inventory.py`** — also red on your head.

## 4. And six test failures, also pre-existing on your head

`perry-codegen --lib` gives **1200 passed, 5 failed** on `p8719` built standalone — the identical five that appear after the merge:

```
element_shape_loop_tests::element_binding_form_through_a_parameter_gets_the_clone
element_shape_loop_tests::element_shape_versioned_loop_admits_an_array_length_bound
element_shape_loop_tests::element_shape_versioned_loop_fires_for_the_churn_read_shape
element_shape_loop_tests::exact_local_array_construction_skips_the_runtime_shape_scan
type_analysis::numeric::tests::char_code_at_on_an_unproven_receiver_keeps_the_runtime_lowering
```

Cause: the PR inserts `versioned_indexed_loop::lower` and `stable_packed_loop::lower` **ahead of** the element-shape/class-field tiers in `lower_for` (`stmt/loops.rs:4888`), so loops that previously reached `element_shape.loop.preheader.brand` are now claimed by the new tier. That's arguably intended — but the existing tests encode the old tiering and need updating to match, which is a decision about what the tiers should assert.

A sixth, in `native_proof_regressions::invalidation::preloop_dynamic_call_invalidates_cached_and_packed_array_proofs`, is the same mechanism: it asserts a `for.cond.` block exists, and the loop now lowers to `for.stable_packed_fast.cond.*` / `for.stable_packed_slow.cond.*`.

## 5. Where that leaves it

The rebase is genuinely done and I'm happy to hand it over — say the word and I'll push the resolved merge to your branch, or paste the three resolutions. But the gates and the six tests are decisions about your own change (which tier should claim these loops, and what the tests should now assert), so they're yours rather than mine to make.

Once those are green I'll re-run the full 30-checker gate set and land it.

proggeramlug:
Re-checked at head `a9972766b`. The push since my last comment was a rebase onto newer `main` (706 files, +95k — it picks up the WinUI landing and the rest), so the blockers I listed are unchanged rather than addressed. Numbers on a clean checkout of that head, with no merge applied:

**Four gates still red:**
```
scripts/check_file_size.sh          FAIL
scripts/gc_store_site_inventory.py  FAIL
scripts/addr_class_inventory.py     FAIL
scripts/shape_descriptor_census.py  FAIL
```

File sizes (cap is >2000): `codegen/artifacts.rs` **2032**, `codegen/method.rs` **2020**, `stmt/let_stmt.rs` **2012`. The rebase moved `let_stmt.rs` from "only over after merging" to over on its own, so all three are now yours.

**Five lib tests still failing** — `perry-codegen --lib` gives 1212 passed, 5 failed:
```
stmt::element_shape_loop_tests::element_binding_form_through_a_parameter_gets_the_clone
stmt::element_shape_loop_tests::element_shape_versioned_loop_admits_an_array_length_bound
stmt::element_shape_loop_tests::element_shape_versioned_loop_fires_for_the_churn_read_shape
stmt::element_shape_loop_tests::exact_local_array_construction_skips_the_runtime_shape_scan
type_analysis::numeric::tests::char_code_at_on_an_unproven_receiver_keeps_the_runtime_lowering
```

The assertion message is the useful one:

> `for (j = 0; j < arr.length; j++)` is the idiom every #7480 kernel is written in and must reach the clone, but `element_shape.loop.preheader.brand` is absent

That's the tiering question from my earlier comment, unchanged: inserting `versioned_indexed_loop::lower` / `stable_packed_loop::lower` ahead of the element-shape tier in `lower_for` means loops that used to reach the element-shape clone are now claimed by the new tier. Either the new tiers should decline these shapes, or those tests should assert the new tier — but that's a decision about what the tiers are *for*, which is yours.

Everything else still looks good, and the merge itself is clean against current `main` (`32f0eacee`). Ping me once the gates and those five are green and I'll re-run the full 30-checker set and land it.

proggeramlug:
Big progress at `f3982e004` — the four red gates are down to one formatting nit, the three oversized files are all under the cap (`artifacts.rs` 1985, `method.rs` 1978, `let_stmt.rs` 2000), and four of the five failing tests are green. One remains, and I think it's a real soundness bug rather than a stale expectation, so here's the evidence.

## The remaining failure

`type_analysis::numeric::tests::char_code_at_on_an_unproven_receiver_keeps_the_runtime_lowering` — the **negative control**, whose own comment says it "is what makes the two tests above meaningful rather than tautological (both would pass on a build that fired the fast path unconditionally)."

Its first assertion **passes**: no `cca.fast`, so you are not inlining a StringHeader read on an unproven receiver. Good.

Its second assertion fails, and the emitted IR shows why:

```llvm
anystr.string.5:
  %r21 = call double @js_string_char_code_at(i64 %r19, i32 %r20)
  br label %anystr.merge.7

anystr.generic.6:
  %r27 = call double @js_typed_feedback_native_call_method_by_id(i64 …, double %r12, i64 %r24, ptr %r25, i64 1)
  br label %anystr.merge.7

anystr.merge.7:
  %r28 = phi double [ %r21, %anystr.string.5 ], [ %r27, %anystr.generic.6 ]
  %r29 = fptosi double %r28 to i32
  %r30 = xor i32 %r11, %r29
```

The **dispatch** is properly guarded — string arm, generic arm, merge. The **xor is not**. The phi merges a proven number (`%r21`, from `js_string_char_code_at`) with a completely unproven value (`%r27`, the result of an arbitrary user method), and then `fptosi`/`xor i32` treats both as numbers. `js_dynamic_bitxor` appears only as a `declare`, never a `call`.

For an `any`-typed receiver that is a user object with its own `charCodeAt` returning a **BigInt**, `%r27` is a NaN-boxed pointer. `fptosi` on that yields garbage, where the language requires either BigInt xor semantics or a `TypeError` for mixing BigInt with Number.

So the guard proves the *receiver*, but the proof doesn't survive the merge — and the arithmetic downstream consumes the merged value as if it did. That's the same shape as #7773: a value that is only numeric on one arm being treated as numeric after the join.

I'd suggest either keeping `js_dynamic_bitxor` when either phi input is unproven, or sinking the arithmetic into the arms so the string arm keeps its fast xor and the generic arm keeps the helper. Which is right depends on what you want the tier to cost, so it's your call — but I don't think the test should just be updated to accept the current IR.

## The other thing, trivial

`cargo fmt --all -- --check` is red on two files your rebase picked up:
```
crates/perry-transform/src/async_to_generator_tests.rs:616
crates/perry-transform/src/generator/lower.rs:1017
```
A plain `cargo fmt --all` fixes both.

Everything else is green on a clean checkout of your head: all other lint gates pass, and `perry-codegen --lib` is 1224 passed / 1 failed — that one test. Fix it and the formatting and I'll re-run the full 30-checker set plus the lib suites and land it.

proggeramlug:
Landed on `main` via #8755 (squash `ed14ce53b`), with the soundness blocker fixed.

**I owe you a correction first.** I attributed the bug to `binary.rs` mishandling the join at `anystr.merge`. That was wrong. The real cause is one lever in `expr/literals_vars.rs`: the new `explicit_numeric_toint32` disjunct bypasses `can_lower_expr_as_i32_in_current_region`, which correctly answers `false` for `(h ^ recv.charCodeAt(i)) | 0` — a `Call` with a `PropertyGet` callee isn't an i32-chain leaf. It then calls into the `lower_expr_as_i32` family without its documented precondition, and `lower_expr_native_i32`'s i32-chain arm recurses structurally, `fptosi`-ing every operand it can't lower natively. So `binary.rs`, which already applies `is_provably_not_bigint` per operand, was bypassed entirely rather than getting the join wrong.

Your predicate is fine and stays. `expr_produces_canonical_raw_f64` is a claim about the **value** the ordinary lowering produces — the bug was consuming it as a licence to re-evaluate the **operands** natively.

The fix keeps the lever and lowers through `lower_expr` + `toint32_fast`, so an unproven tree keeps `js_dynamic_bitxor` and a proven one still gets the inline `xor i32`. The i32-slot store the lever exists for is preserved, so this is still an improvement over `main`, which took neither branch there.

On the alternative: sinking the arithmetic into the arms is ruled out by house precedent rather than taste — `lower_guarded_numeric_add`'s doc block records it was tried and went 86 ms → 119 ms, because per-node diamonds make the outer op consume a phi LLVM can't prove canonical.

**One perf caveat worth your eye:** in the proven case `h` now goes `load i32 → sitofp → fjcvtzs` rather than being read as i32 directly. `fjcvtzs` is a target intrinsic so it likely won't fold through `sitofp` — possibly one or two residual instructions per iteration versus the (unsound) version. Still ≥ `main`. If you want that back, the sound route is widening `can_lower_expr_as_i32_in_current_region` so the tree genuinely qualifies, not bypassing it.

Validated on the merged result: all 30 lint checkers, codegen 1229/0, runtime 2671/0 at `RUST_TEST_THREADS=1`, transform 93/0, hir 334/0. The negative control passes **unmodified** — its blob hash is byte-identical to `main` — and both positive tests still fire the fast path, so the optimization isn't merely disabled.

Two things I did not verify, stated plainly: your end-to-end `issue_8690_loop_versioned_arraylike.rs` needs the `-static` wrappers and wasn't run; and one early `--tests` sweep showed `typed_array_rmw_8692` at 8/9, which I could not reproduce in nine subsequent runs and which passes on clean `main` — I believe it's flaky, but it's on the record rather than dropped.

proggeramlug:
Heads-up — this reopened when you pushed `c8718934e`, and the branch is now on a stale base, so it can't be merged as-is.

**The soundness fix already landed.** It went in via #8755 (squash `ed14ce53b`), and `crates/perry-codegen/src/expr/literals_vars.rs` on `main` carries the `lower_expr` + `toint32_fast` lowering.

**Merging this branch now would revert three later PRs.** Its diff against `main` is `-1276` lines, and the deletions include:
- `crates/perry/tests/issue_5894_function_var_binding.rs` — #8743's test
- `test-files/test_guarded_undefined_method_param.ts` and the `local_binding_type_allowlist.json` entry — #8740's
- `docs/api/perry.d.ts` and `docs/src/api/reference.md` — #8742's regeneration

**Your new commit is worth landing on its own**, though. It narrows the lever to require being within a stable-packed numeric clone ("other loop clones own narrower indexed-load contracts that their existing assignment lowering must continue to see") and — valuably — adds 36 lines to `issue_8690_loop_versioned_arraylike.rs`, the end-to-end test I flagged as unrun.

It does conflict with what landed, in exactly the place you'd expect: my fix changed the **lowering** on that lever, yours narrows its **gate**. Both edit the same block. That composition is yours to make rather than mine to guess at.

One data point that may be useful. I reported a single `typed_array_rmw_8692` failure at 8/9 during validation and recorded it as probably-flaky. Your commit message made me re-test that properly, and on current `main` it is clean: **0 failures in 8 standalone runs and 0 failing suites across 2 full `--tests` sweeps**. So if `preserve typed array loop lowering` was chasing that specific symptom, it may not be live on `main` — though it's plainly still correct as hardening, and the added e2e coverage is worth having either way.

Suggested path: rebase onto current `main` (`8922c4245`) so it carries only `c8718934e`'s content, and I'll validate and land it.

proggeramlug:
Closing as superseded by merged #8755. I verified pristine merged `main` (`ed14ce53b`) against the complete `test_gap_typedarray_param_read.ts` corpus: Perry output matches Node byte-for-byte, including the Uint32Array signed `ToInt32` case that initially failed on the pre-merge branch. The ordinary-lowering fix included in #8755 already resolves it, so the follow-up restriction would be redundant.

proggeramlug:
The follow-up commit landed on `main` via the PR above (squash on top of `1b0c300e0`), rebased so it carries only its own content — merging the branch would have reverted #8740, #8742, #8743 and #8763.

The conflict with the landed soundness fix was **comment-only**; the code auto-merged. Both halves survive: the narrowed gate is in, and the unproven path still lowers through `lower_expr` + `toint32_fast` to `js_dynamic_bitxor`.

One thing worth your attention. Under the narrowed gate, the three `char_code_at_*` probes no longer enter the branch #8755 fixed — their loops are bounded by `i < 64` rather than an array length, so `stable_packed_loop_facts` is empty and they fall through to the generic tail. They still pass, but for a different reason than before. That branch's real coverage is now `read_only_loops_have_preheader_proofs_and_fallback_free_fast_blocks` in the e2e suite you extended — which I could not run here for disk reasons. Worth executing on a machine with room, since it's currently the only thing exercising that path.

PR #8718, opened 2026-08-24
fix(net): support named-pipe and Unix-socket IPC

proggeramlug:
Closes #6620

## Summary

- route `server.listen(path)`, `net.connect(path)`, and `{ path }` overloads through local IPC
- add Windows named-pipe and Unix-domain-socket transports to the existing socket lifecycle
- preserve connection ordering, limits/drop events, close cleanup, `server.address()`, and deferred `Socket.connect()` behavior
- add platform round-trip coverage and keep dynamic dispatch/codegen signatures aligned

## Testing

- `cargo test -p perry-ext-net --lib` (30 passed)
- `cargo check -p perry-ext-net`
- `cargo check -p perry-codegen -p perry-stdlib`
- `cargo fmt -p perry-ext-net -p perry-codegen -p perry-stdlib -- --check`

No version bump.

## Summary by CodeRabbit

- **New Features**
  - Added local IPC socket support for `net.connect(path)`, `socket.connect(path)`, and `server.listen(path)`.
  - Supports Unix-domain sockets and Windows named pipes.
  - Added path-based connection and listener overloads, including options objects.
  - `server.address()` now reports the IPC path for local listeners.
  - Socket connection calls now support multiple overloads and return the socket for chaining.

- **Bug Fixes**
  - Prevented callback arguments from being incorrectly converted to strings.
  - Added explicit rejection for unsupported TLS upgrades on IPC sockets.

PR #8720, opened 2026-08-24
feat(native): stabilize native value profile boundaries

### LINKED ISSUE #6827: RFC: Stabilize a Perry native value profile for fixed-width scalars, POD records, and native boundaries

proggeramlug:
## Summary

Completes the stable native value profile from RFC #6827: exact-width scalar ABI handling, source-linked POD contracts, nested POD values, and copy-safe managed/native boundaries. Ordinary TypeScript number behavior remains unchanged, while explicit native crossings now reject invalid or imprecise values.

## Changes

- Add the full public scalar vocabulary (`i8`, `i16`, `i32`, `i64`, `isize`, `u8`/`byte`, `u16`, `u32`, `u64`, `usize`, `f32`, `f64`, `bool`, `handle_id`) across manifest parsing, codegen, runtime checks, and documentation.
- Reject fractional, out-of-range, and unsafe integer materialization instead of truncating or silently losing `i64`/`u64` precision.
- Resolve `source: "./module.ts#Export"` POD declarations, derive layouts, and report exact manifest/source drift during compile and `perry native validate`.
- Support nested closed POD records, recursive initialization/materialization, and true value-copy semantics for local assignment and ordinary function calls while preserving `PodView` aliasing.
- Update the manifest schema and native-value docs, and add focused HIR, transform, runtime, codegen, validation, and end-to-end regressions.

## Related issue

Closes #6827

## Test plan

- [x] `cargo build --release -p perry --bin perry`
- [x] `cargo check -p perry --tests`
- [x] `cargo test -p perry-api-manifest exact_width_scalar_spellings_are_canonical_and_pod_safe`
- [x] `cargo test -p perry-hir native_profile::tests`
- [x] `cargo test -p perry-runtime scalar_guards_reject_incompatible_js_values`
- [x] Focused codegen tests for exact-width guards, nested POD copies, safe `i64`/`u64` materialization, and native ABI returns
- [x] `cargo test -p perry-transform pod_value_parameters_preserve_the_call_copy_boundary`
- [x] `perry run test-files/test_parity_native_value_profile.ts`
- [x] User-facing regression added under `test-files/`
- [x] CLI/runtime API documentation updated under `docs/src/`
- [ ] Full workspace test suite (not run locally)

## Screenshots / output

The end-to-end parity test reports exact-width conversion results, all expected rejection flags, nested materialization, and independent POD copies across both local assignment and an ordinary function boundary.

## Checklist

- [x] My commits follow the repository prefix convention


proggeramlug:
Landed on `main` via #8723 (squash `8beca2f29`), with #8672 and #8718.

Validated on the merged result: all 30 lint checkers, runtime 2667/0, codegen 1214/0 plus all codegen integration suites clean, transform 93/0, ext-net 30/0. Thanks!

PR #8722, opened 2026-08-24
feat(windows): complete WinUI 3 backend

### LINKED ISSUE #4680: Add opt-in WinUI 3 (Fluent) Windows target: `--target windows-winui`
## Summary

Add an opt-in **WinUI 3 / Fluent** rendering backend for Windows, selectable via `--target windows-winui`, alongside the existing Win32/GDI backend (`crates/perry-ui-windows`).

Motivation from discussion #3486: the current Win32/GDI output looks dated (boxy, flat, no animation, Win7-era controls). WinUI 3 brings Fluent Design — rounded corners, Mica/Acrylic materials, smooth animations, modern typography, crisp auto-DPI — and is the closest Windows analog to our iOS/Android look-and-feel.

## Scope / decisions

- **Opt-in, not the default.** WinUI 3 requires the **WinAppSDK runtime** + an MSIX/bootstrapper packaging story, which conflicts with Perry's single-native-`.exe` model. Keep Win32 as the default target; expose WinUI via `--target windows-winui`. Mirrors how we ship multiple Apple targets.
- **No new "UI framework" dependency confusion:** `microsoft/windows-rs` (already a dep, `windows = "0.58"`) is the WinRT projection crate, *not* a UI lib. WinUI from Rust = driving WinAppSDK + XAML through the WinRT projections (imperative XAML object trees in Rust).
- **Drop Win7/8/8.1 as a floor** for this target; **no XAML Islands** (kludge). WinUI 3 targets Win10 1809+/Win11.

## Work breakdown (rough)

1. New crate `perry-ui-windows-winui` (or a feature on the existing crate) wiring WinAppSDK + Microsoft.UI.Xaml projections.
2. Target plumbing: `--target windows-winui` through the compile/link pipeline; bootstrapper/runtime acquisition (`Microsoft.WindowsAppRuntime.Bootstrap`).
3. Widget mapping layer: map the perry-ui widget set (~40 widgets) onto XAML controls (Button, ListView/ItemsRepeater, NavigationView, TextBox, etc.).
4. Window chrome: Mica/Acrylic backdrop, Fluent title bar, theme (light/dark/system).
5. Packaging: MSIX or sideload-friendly unpackaged WinAppSDK bootstrap; document the runtime dependency.
6. Geisterhand `apply_style` dispatcher for the WinUI backend (parity with macOS/GTK4/Win32).

## Acceptance

- `--target windows-winui` produces a runnable app rendering the ToDo sample with Fluent chrome (Mica, rounded corners, animated controls).
- Win32 remains the default and unaffected.
- Docs cover the WinAppSDK runtime requirement and how to choose backends.

Tracks discussion #3486. Companion near-term Win32 polish: see the linked Win32 Fluent-polish issue.

proggeramlug:
## Summary
- replace the `windows-winui` scaffold with real WinUI 3 / Fluent rendering through a vendored Windows Reactor snapshot from microsoft/windows-rs commit 65066a7109c214f317ed66261cfb7518160b8aaf, with upstream licenses and attribution preserved
- map Perry's core controls, state updates, callbacks, lifecycle, layout, and Mica window backdrop to native WinUI controls
- link and deploy the required Windows App SDK bootstrap import library, DLL, and PRI resources for compiled applications
- document the backend, runtime requirements, diagnostics, and fallback behavior while preserving the default Win32 backend and its ABI

## Testing
- `cargo build --release -p perry-ui-windows-winui`
- `cargo build --release -p perry-ui-windows`
- `cargo test -p perry winui_asset_tests`
- compared exported Perry UI symbols: 326 WinUI / 326 Win32, with no missing or extra symbols
- compiled the ToDo example with `--target windows-winui`
- launched the compiled ToDo executable and confirmed the Fluent backend remained running with bootstrap status `ready`

No version bump.

Closes #4680


proggeramlug:
Holding this one — not on correctness, but because the gate failures need policy decisions about vendored third-party code that I don't think are mine to make unilaterally on a 73,441-line drop.

**Vendoring hygiene looks right**, for the record: per-crate `license-mit` and `license-apache-2.0` files are present for all 13 vendored crates, and the upstream commit (`65066a7109c214f317ed66261cfb7518160b8aaf`) is pinned in-tree. It also merges cleanly against current `main` (`8beca2f29`).

Three gates are red:

**1. `workspace_architecture.py --check`** — the 13 vendored crates land as *implicit workspace members*:

```
implicit workspace members: crates/perry-ui-windows-winui/vendor/windows-{collections,core,future,
  implement,interface,link,numerics,reactor,reference,result,strings,threading,time}
unclassified workspace crates: windows-collections, windows-core, …
crates missing [lints] workspace = true: windows-collections, windows-core, …
architecture baseline differs; review the structural change and refresh workspace-architecture.json
```

This one is substantive rather than cosmetic: as workspace members they enter the build graph, so `cargo test --workspace` and the deny-warnings jobs will now compile them, and `[lints] workspace = true` would apply our lint policy to upstream code. The usual answer is `workspace.exclude` for vendored trees — but that's your call, and it needs the architecture baseline refreshed either way.

**2. `check_file_size.sh`** — four vendored files exceed the 2000-line cap, the largest at 26,441 lines:

```
  3094  vendor/windows-collections/src/bindings.rs
  2015  vendor/windows-future/src/bindings.rs
 26441  vendor/windows-reactor/src/bindings.rs
  4246  vendor/windows-reactor/src/winui/backend/mod.rs
```

Generated bindings obviously shouldn't be split, so these want the script's documented `ALLOWLIST` block with a rationale. Again a deliberate decision — it exempts 73k lines from that gate permanently.

**3. One that is yours, not the vendor's:** `crates/perry/src/commands/compile/link/build_and_run.rs` is back over the cap at **2110**. I split the `is_watchos` arm out of that file a few hours ago (#8721 lineage) to get it under; this PR pushes it back. Worth extracting the block you added rather than growing it further.

**4. `gc_runtime_root_holders.py`** — two new holders in the new crate need a verdict or scanner coverage:

```
crates/perry-ui-windows-winui/src/app.rs:33:     APPS:  RefCell<Vec<AppState>>   [rule S]
crates/perry-ui-windows-winui/src/widgets.rs:137: NODES: RefCell<Vec<Node>>      [rule S]
```

Note #8713 just landed (via #8721) and registers root scanners for every other UI crate, closing #8701 — if `AppState`/`Node` hold JS callbacks as NaN-boxed `f64`, this crate should join that scheme rather than take a `not_a_gc_pointer` verdict. That's the correctness-relevant one of the four.

One caveat on my side: I can't validate any of the WinUI behaviour from macOS, and your test plan is `cargo build --release -p perry-ui-windows-winui`, which I also can't run here. So my review is limited to the gates, the merge, and the vendoring hygiene — someone on Windows should confirm the backend actually works before this lands.

proggeramlug:
Landed on `main` via #8733 (squash `0749bd338`), with all four gate blockers resolved.

**The vendored tree had to move**, and I verified that rather than assuming it. Cargo's `is_excluded` is `!explicit_member && excluded`, and `explicit_member` matches any `members` entry that is a path *prefix* — so while the snapshot sat under `crates/perry-ui-windows-winui/vendor/`, the parent's member entry won. Measured on your head: as submitted, 92 packages with 13 vendored crates as members; adding `exclude` in place, still 92 and still 13; moved to `third_party/windows-winui/` and excluded, 79 — which matches the existing architecture baseline exactly, so no refresh was needed. That mattered substantively: as members, `cargo test --workspace` was compiling 73k lines of upstream code and `[workspace.lints]` would have applied to it.

**On the GC side, `Node` does hold JS values** — every widget callback is a raw closure pointer via `js_nanbox_get_pointer` — so it got scanner coverage following #8713's pattern rather than a `not_a_gc_pointer` verdict, which would have been false. `AppState` genuinely doesn't (String, two f64 dimensions, an i64 that indexes NODES rather than an address, two Options, a PresenterKind), so that one took a verdict. Three more roots the census hadn't flagged are scanned too — `ON_ACTIVATE`, `ON_TERMINATE`, `PENDING_TIMERS` — and the scanner is armed on the Fluent path, where winui shadows `app_create` and would otherwise have left `perry-ui-windows`' own tables unscanned.

`build_and_run.rs` was **not** allowlisted — the block this PR added moved to `link/winui_assets.rs` instead (2110 → 1970). Only the four generated upstream bindings are allowlisted.

**Two things you should know:**

1. **A residual I did not fix.** `render_handle` works on a *clone* of the node and copies the unboxed callback pointer into `move` closures owned by the Reactor element tree; `start_runtime_pump` and `app_run` do the same for timers and `on_exit`. Those copies live inside boxed Rust closures no scanner can reach or rewrite, so an evacuating collection leaves them stale. Scanning the tables cannot fix it — the real fix is re-reading a scanned slot at invoke time, the indirection perry-ui-macos gets from its handle-keyed maps. It's a pre-existing property of the design, it's documented at both scanners, and it deserves a follow-up issue.

2. **WinUI runtime behaviour is unverified.** It can't be built or run from macOS, and a real cross-compile fails on a C build script. The scanner and type definitions were validated by compiling the real definitions against the real `perry_ffi::GcRootVisitor`, which proves exhaustiveness and signatures but nothing about the backend working. Someone on Windows should confirm it. The three relocated `winui_asset_tests` type-check but weren't executed.

Validated: all 30 lint checkers, `cargo check -p perry` and `-p perry --tests` clean with zero warnings.

PR #8659, opened 2026-08-23
fix(intl): complete issue #5896 Test262 worklist

proggeramlug:
Closes #5896.

## Summary

- Canonicalize Intl locale lists with ICU4X CLDR aliases and likely-subtag data, including Unicode/transformed extension aliases, maximize/minimize, and the non-IANA `posix` variant.
- Apply ECMA-402 Get/Has/ToObject ordering to proxy-backed locale and option inputs, sparse locale arrays, boxed primitives, and constructor option resolution.
- Complete the listed Collator, PluralRules, RelativeTimeFormat, Segmenter, and subclass behaviors: extension negotiation, defaults and categories, bound accessors, iterator identity, and native prototype inheritance.
- Keep maximum-length arrays logically sparse so Intl tests can grow/shrink them without materializing billions of holes.

There are no structurally impossible or skipped issue-worklist tests: the complete #5896 list passes.

## Validation

- Pinned Test262 `4249661388e5d3f92a85186213da140a6481490f`, exact #5896 worklist: **101 passed, 0 diff, 0 runtime failures, 0 compile failures, 0 skipped, 100% parity**.
  - Collator 15/15
  - DisplayNames 7/7
  - Intl 18/18
  - ListFormat 10/10
  - Locale 11/11
  - PluralRules 5/5
  - RelativeTimeFormat 17/17
  - Segmenter 18/18
- `cargo build --release -p perry -p perry-runtime-static -p perry-stdlib-static`
- `cargo test -p perry-runtime`: 2,643 passed, 0 failed, 4 ignored; doctests 0 failed.
- `cargo test -p perry`: the 1,026 binary/unit tests pass. The command then reaches four pre-existing failures in the unchanged `aliased_native_class_import` integration test (`aliased_async_local_storage_matches_unaliased`, `aliased_matches_unaliased_for_native_class`, `aliased_net_block_list_lowers_like_unaliased`, and `aliased_net_socket_keeps_native_methods`). They reproduce in isolation as unresolved aliased native imports; both that test and the alias-resolution implementation are unchanged from merge base `f96a9d9e0e8d77d58e492465965d039ec90a7b77`. The only touched `perry` code is Intl feature detection, whose tests pass in the 1,026-test suite.
- `cargo fmt --all -- --check`; `git diff --check`
- `scripts/check_file_size.sh`: no Rust source exceeds 2,000 lines.
- Raw-handle debt self-test/check/merge-base ratchet: 925 -> 925, no module ceiling raised.
[...]

PR #8725, opened 2026-08-24
fix(intl): close #8659 follow-up review gaps

proggeramlug:
## Summary

- validate `Intl.PluralRules` digit-option bounds before flooring
- root the dynamic superclass and emitted network error values across allocating/user-code calls
- reject constructable superclasses whose `.prototype` is neither an object nor `null`
- restore the current-main thread-local policy gate for `bun:ffi` reads

Follow-up to #8659, which landed through #8723 before these final review fixes reached its head. The TLS policy correction fixes a main regression exposed by this PR’s auxiliary gate.

## Validation

- `cargo test -p perry-runtime --lib` (2,669 passed; 4 ignored)
- warnings-as-errors checks for `perry-runtime` and `perry-stdlib`
- focused plural/category and superclass-prototype tests
- thread-local checker self-test and ratchet
- file-size, raw-handle-debt, and unrooted-local-shape gates

### DIFF
--- a/changelog.d/8659-follow-up-review-fixes.md
+++ b/changelog.d/8659-follow-up-review-fixes.md
@@ -1,0 +1,5 @@
+### Fixed
+
+Validate fractional `Intl.PluralRules` digit options before flooring, reject
+invalid superclass prototypes, and keep emitted network errors alive across
+successive listeners.
--- a/crates/perry-runtime/src/bun_ffi/read.rs
+++ b/crates/perry-runtime/src/bun_ffi/read.rs
@@ -22,7 +22,7 @@ const READERS: &[(&str, u8)] = &[
     ("f64", super::types::T_F64),
 ];
 
-thread_local! {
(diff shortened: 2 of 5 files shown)


proggeramlug:
Landed on `main` via #8727 (squash `d53f34b4d`), with #8726.

Thanks especially for the `perry_thread_local!` restoration — that's a regression I let through when I merged #8704, and it's worth naming why it mattered: a plain `thread_local!` costs twice there. It loses the `HotTls` address cache (a `_tlv_get_addr` call per read on Darwin), *and* it escapes the root-holder census, which keys off the macro — so that read cache was never classified at all. Both failure directions are silent: a working-but-slow path and an invisible holder, neither of which reddens anything on its own. Exactly the kind of thing that needs a gate rather than vigilance.

Validated on the merged result: all 30 lint checkers, runtime 2669/0, codegen 1214/0 plus all integration suites clean.

Issue #2889, opened 2026-05-30
runtime: route rebound global built-in constructors through real semantics


andrewtdiz:
## Summary

`globalThis` exposes many built-in constructors as function-shaped values, but most of those values are no-op sentinels when they are called or rebound through a local. Node treats these as the real constructors/functions, so common patterns like `const C = globalThis.Number; C("42")` or `const E = globalThis.Error; new E("x")` should preserve the corresponding built-in semantics.

## Node behavior

On local Node v25.9.0:

```js
for (const [name,args] of Object.entries({
  Array:[3], Object:[null], String:[123], Number:["42"], Boolean:[0],
  Date:[0], RegExp:["a","g"], Error:["x"], Uint8Array:[3], ArrayBuffer:[4]
})) {
  const F = globalThis[name];
  const called = F(...args);
  const constructed = new F(...args);
  console.log(name, typeof F, Object.prototype.toString.call(called), Object.prototype.toString.call(constructed));
}
```

Observed examples:

- `Array` is a function; `globalThis.Array(3)` and `new globalThis.Array(3)` both produce arrays.
- `Number` is a function; calling it coerces to a number and constructing it produces a Number object.
- `Boolean`, `String`, `Date`, `RegExp`, and `Error` all preserve their Node call/construct behavior when read from `globalThis`.
- constructors such as `Map`, `Set`, typed arrays, and `ArrayBuffer` throw on bare call but construct correctly with `new`.

## Perry evidence

The current runtime intentionally installs most `globalThis` built-in constructors as no-op closure sentinels:

- `crates/perry-runtime/src/object/global_this.rs::GLOBAL_THIS_BUILTIN_CONSTRUCTORS` lists core constructors including `Array`, `Object`, `String`, `Number`, `Boolean`, `RegExp`, `Date`, `Error`, `Map`, `Set`, typed arrays, `ArrayBuffer`, `DataView`, etc.
- `global_this_builtin_noop_thunk()` documents that these values only make `typeof globalThis.Array === "function"` and `Array.prototype` reads work, while direct calls like `globalThis.Array(3)` return `undefined`.
- `populate_global_this_builtins()` uses `global_this_builtin_noop_thunk` for every constructor except `String` and `Buffer`.
- `crates/perry-codegen/src/expr/new_dynamic.rs` routes `new <LocalGet>(...)` / `new <PropertyGet>(...)` through `js_new_function_construct()`, so rebound constructors depend on runtime recovery.
[...]

PR #8726, opened 2026-08-24
fix(runtime): recognize captured ArrayBuffer/SharedArrayBuffer/DataView constructors (#8724)

### LINKED ISSUE #8724: regression(runtime): captured ArrayBuffer/SharedArrayBuffer/DataView throws "Constructor requires 'new'" (breaks cli.js init) — #06e1ab349
## Summary

Since `06e1ab349` (merge landing #8652/#8651/#8647/#8646/#8650), **constructing `ArrayBuffer` / `SharedArrayBuffer` / `DataView` via a captured value throws `TypeError: Constructor requires 'new'`** — e.g. `const D = DataView; new D(buf)`, `Reflect.construct(DataView, [buf])`, `class X extends DataView {}`, `const AB = globalThis.ArrayBuffer; new AB(8)`. Direct `new DataView(buf)` / `new globalThis.DataView(buf)` still work.

Minified/bundled code captures globals into locals pervasively, so this breaks module init broadly. The **natively-compiled Claude Code cli.js 2.1.112 bundle throws this at startup on essentially every command** (`--help`, `doctor`, `mcp list`, `-p`, …; only `--version` escapes via its early exit). Node runs the same bundle fine.

## Repro

```ts
const ab = () => new ArrayBuffer(8);
const D = DataView;               new D(ab());                  // THROWS (node: ok)
                                  Reflect.construct(DataView,[ab()]); // THROWS (node: ok)
class MV extends DataView {}      new MV(ab());                 // THROWS (node: ok)
const AB = (globalThis as any).ArrayBuffer; new AB(8);         // THROWS (node: ok)
new DataView(ab());               // ok
new (globalThis as any).DataView(ab()); // ok
```

`Map`/`Set`/`Promise`/`Uint8Array` captured-then-constructed all work — only AB/SAB/DataView regressed.

## Root cause

`06e1ab349` switched ArrayBuffer/SharedArrayBuffer/DataView's global [[Call]] behavior from `global_this_builtin_noop_thunk` to the shared `construct_only_builtin_call_thunk` (`global_this/populate.rs:203`). But `identify_global_builtin_constructor` (`class_registry/class_meta.rs`) — the func_ptr allow-list the **dynamic-`new`** path uses to recover a captured builtin's name and route it to `construct.rs`'s real factory — was never updated to accept the new thunk. So a captured AB/SAB/DataView closure is no longer recognized (returns `None`), never reaches the `"ArrayBuffer" | "SharedArrayBuffer"` / `"DataView"` arms in `construct.rs:891/900`, and falls through to invoke the bare-call thunk → "Constructor requires 'new'". The previous noop thunk **was** in the allow-list, which is why this worked before `06e1ab349`. Same bug class the WeakMap/RegExp comments in that function already document.

## Fix

Add `construct_only_builtin_call_thunk` to the `identify_global_builtin_constructor` allow-list — the name-dispatch construct arms already exist. PR incoming.

proggeramlug:
## Summary

Fixes #8724. Constructing `ArrayBuffer` / `SharedArrayBuffer` / `DataView` through a **captured value** threw `TypeError: Constructor requires 'new'`:

```ts
const D = DataView;                 new D(buf);                     // THREW (node: ok)
Reflect.construct(DataView, [buf]);                                 // THREW (node: ok)
class X extends DataView {}         new X(buf);                     // THREW (node: ok)
const AB = globalThis.ArrayBuffer;  new AB(8);                      // THREW (node: ok)
new DataView(buf);                  // ok
```

Minified bundles capture globals into locals everywhere, so this broke module init broadly — the natively-compiled Claude Code cli.js 2.1.112 bundle threw at startup on nearly every command (only `--version` escaped via early exit); node runs the same bundle fine.

## Root cause

`06e1ab349` (merge of #8652/#8651/#8647/#8646/#8650) switched these three builtins' global [[Call]] behavior from `global_this_builtin_noop_thunk` to the shared `construct_only_builtin_call_thunk` (`global_this/populate.rs:203`), but did **not** add that thunk to `identify_global_builtin_constructor`'s func_ptr allow-list (`class_registry/class_meta.rs`) — the recognition step the dynamic-`new` path uses to recover a captured builtin's name and route it to `construct.rs`'s real factory (arms for `"ArrayBuffer" | "SharedArrayBuffer"` / `"DataView"` at `construct.rs:891/900` already exist). A captured AB/SAB/DataView closure stopped being recognized (returned `None`), never reached those arms, and fell through to invoke the bare-call thunk. The old noop thunk **was** in the allow-list — the exact bug class the WeakMap (`%WeakMap%` capture) and RegExp (#2889) comments in that function already document.

## Fix

Add `construct_only_builtin_call_thunk` to the allow-list — one line. The name-dispatch construct arms are unchanged.

## Test

`tests/issue_8724_captured_arraybuffer_construct.ts` — captured / `Reflect.construct` / subclass / `globalThis` / resizable forms plus a functional `setInt32`/`getInt32` read-back. Output is byte-identical to Node v26 with the fix (and reproduces the throw without it).

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed captured and indirect construction of `ArrayBuffer`, resizable `ArrayBuffer`, `SharedArrayBuffer`, and `DataView`.
  - Construction now works correctly through `Reflect.construct`, subclasses, and `globalThis` references.
  - Prevented erroneous “Constructor requires 'new'” failures.

- **Tests**
  - Added coverage for construction scenarios and `DataView` read/write behavior.


proggeramlug:
Landed on `main` via #8727 (squash `d53f34b4d`), with #8725.

Nice catch on the asymmetry — direct `new DataView(buf)` working while every captured form threw is the kind of gap that survives a long time because the obvious spelling is the one that's tested.

Validated on the merged result: all 30 lint checkers, runtime 2669/0, codegen 1214/0 plus all integration suites clean.

PR #8731, opened 2026-08-24
perf(codegen): specialize imported methods capturing this

### LINKED ISSUE #8693: perf(codegen): specialize imported class methods that capture this
## Summary

Perry rejects typed specialization for many imported class methods solely because they capture `this`. In the public `perform-ecs/destroy` workload, all major adapter/ECS calls therefore traverse generic runtime method lookup, closure dispatch, and ABI/vtable trampolines even though the receiver classes and method identities are stable.

This blocks both the direct call win and downstream inlining, representation propagation, constant folding, scalar replacement, and allocation removal.

This ticket is distinct from #8672: that PR preserves already-proven direct-method guards across unrelated prototype writes. This issue covers sites where Perry never selects a typed/direct clone because the imported method captures `this` or its receiver proof is not propagated through the package graph.

Create two ESM files:

```js
// registry.js
export class Group {
  entities = [];

  pushEntity(entity) {
    this.entities.push(entity);
  }

  removeEntity(entity) {
    const index = this.entities.indexOf(entity);
    if (index !== -1) this.entities.splice(index, 1);
  }
}

export class Registry {
  group = new Group();

  add(entity) {
    this.group.pushEntity(entity);
  }

  remove(entity) {
    this.group.removeEntity(entity);
  }
}
```

```js
// main.js
import { Registry } from './registry.js';

const registry = new Registry();
const iterations = 200_000;
  const entity = { id: i };
  registry.add(entity);
  registry.remove(entity);
}
console.log(JSON.stringify({
  remaining: registry.group.entities.length,
}));
```

```json
{
  "type": "module"
}
```

Build with:

```bash
PERRY_NO_AUTO_OPTIMIZE=1 \
target/release/perry compile main.js -o repro-perry \
```

The output must end with `remaining: 0` in Node and Perry.

The public integration reproduction is `ddmills/js-ecs-benchmarks` with `perform-ecs@0.7.8`, suite `Destroy`:

```js
perform(ctx) {
  const entity = ctx.createEntity();
  ctx.addPositionComponent(entity);
  ctx.addVelocityComponent(entity);
  ctx.destroyEntity(entity);
}
```

## Current evidence

Measured on Perry `7ad718ab4287641cb2b29dce3a056edc45d4c7f8`, Node 26.5.1, Apple M1 Mac mini.

| Runtime | median |
|---|---:|
| Node | 1.024601 us/op |
| Perry | 33.904990 us/op |

Perry/Node: **33.091x**. Every process returned component IDs `["0", "1"]` and zero retained view entries.

A fresh three-second Perry profile attributed the top-level workload branches as follows:

- entity creation: 33.4%
- add position: 26.0%
- add velocity: 22.7%
- destroy: 16.0%

Thus approximately 82% is create/add work, not destruction. Every top-level branch enters `js_typed_feedback_native_call_method_by_id` / `js_native_call_method`, then runtime handle dispatch, `js_native_call_value`, and a closure/vtable trampoline.

The lowering report contains 494 records, **zero typed-path selections**, **zero scalar replacements**, 61 inserted boxes, and 227 typed-path rejections. `captures_this` alone rejects 56 typed-clone candidates.

## Proposed direction

- Treat a known class receiver plus stable method identity as sufficient to build a guarded specialized clone even when the body reads/writes `this`.
- Pass a typed/native receiver representation to the clone and preserve field/shape facts across the call.
- Propagate class and method metadata through ESM/npm import/export boundaries and through stable adapter objects.
- Emit a guarded direct call when the receiver class/shape, prototype method identity, and relevant invalidation generations match.
- Inline profitable small/medium methods after direct-call selection so downstream field, array, and allocation optimizations can see the body.
- Retain generic lookup/call fallback when any receiver or method proof fails.
- Explain selections/rejections, including receiver provenance and `this` representation, in lowering artifacts.

## Semantic constraints

- Preserve `this` binding for extracted, rebound, `call`/`apply`, arrow, accessor, proxy, and bound methods.
- Correctly invalidate on method replacement/deletion/recreation, prototype replacement/mutation, own-property shadowing, descriptor changes, proxy insertion, and class reevaluation.
- Preserve inheritance, `super`, private brands, derived constructors, and exceptions.
- Do not confuse native-module constructor metadata with arbitrary bound native exports; #8682 explicitly kept that distinction when holding #8672.
- Remain correct under moving/forced GC and refresh receiver/captured roots across collecting calls.

## Acceptance criteria

- Add the two-file reproduction as a registered semantic and compiler-output fixture; Node and Perry must finish with `remaining: 0`.
- The stable fast arm calls `Registry.add/remove` and `Group.pushEntity/removeEntity` directly or through typed clones, without `js_native_call_method_by_id`, `js_typed_feedback_native_call_method_by_id`, or generic closure/vtable dispatch.
- At least one method that captures `this` is selected as a typed clone, and `--explain-lowering` no longer records `captures_this` as the rejection for that valid site.
- Explicit fallback remains in artifacts and is exercised by receiver/method/prototype mutation fixtures.
- Tests cover imports/re-exports, inheritance, own shadowing, extraction/rebinding, accessors/proxies, private fields, exceptions, and forced-moving GC.
- Re-run `ddmills/js-ecs-benchmarks` `perform-ecs/destroy` with the exact state oracle. Require at least a 10% median improvement, at least 9/11 paired wins on the quiet M1 protocol, and no regression in the complete ECS adapter suite.
- Report profile deltas for generic method-dispatch samples, executable size, peak RSS, and lowering counts for typed clone selections/rejections.

## Related work

- #8672 improves invalidation granularity for direct-method guards already selected; it does not address the `captures_this` selection failure demonstrated here.
- #8504 may eventually supply advisory profile facts, but this exact imported class graph is statically visible and should not require profile replay for correctness or selection.
- #8691 is the scalar-replacement follow-up that should consume the inlined bodies from this work rather than expanding this ticket into a general escape-analysis change.
- #8694 independently removes stable `for...in` enumeration overhead inside the registry methods exposed by this work.

proggeramlug:
Fixes #8693.

## Summary

- publish producer-authoritative proven-`this` method capabilities through imports, aliases, and re-exports
- emit guarded direct imported clone calls (including profitable class-ID towers) while retaining generic dispatch fallbacks
- preserve exact receiver shape for captured-`this` arrow fields and keep prototype replacement/delete/recreate semantics correct
- integrate with method-scoped prototype invalidation from #8672/#8723 instead of broadening it back to a global hot-path latch
- add semantic/compiler-output fixtures, cache-key coverage, and a changelog fragment; no version bump

## Verification

- `cargo test -p perry-runtime --lib inline_guard_disable_is_per_declared_field_key`
- `cargo test -p perry-runtime --lib typed_feedback_method_direct_guard_fails_for_prototype_method_registration`
- `cargo test -p perry-runtime store_after_delete`
- `cargo test -p perry-codegen --lib proven_this` (15 passed)
- `cargo test -p perry-codegen --lib field_init` (21 passed)
- `cargo test -p perry-codegen --lib tower_route` (7 passed)
- `cargo test -p perry --test issue_8693_imported_this_specialization -- --nocapture --test-threads=1` (normal + forced moving GC)
- compiler-output gate: all 20 checks passed; 2 imported selections, 6 retained generic fallbacks, 24 lowering records
- complete perform-ecs suite produced the expected Add/Remove, Destroy, and Velocity counts with no overall CPU regression

## Performance

`perform-ecs@0.7.8` Destroy, 100k measured iterations + 10k warmup, 11 alternating CPU pairs against upstream main:

- 11/11 wins
- median user CPU: 2.24s -> 1.65s (26.34% faster)
- median max RSS: 56,950,784 -> 57,409,536 bytes (+0.81%)
- executable size: +16,528 bytes (+0.19%)

Every timed run also asserted component IDs and zero retained view/group entries.

## Summary by CodeRabbit

* **Performance**
  * Improved optimization of methods imported across modules, including safer direct-call specialization and efficient dispatch.
  * Added inlining improvements for eligible optimized methods.

* **Bug Fixes**
  * Improved class prototype behavior when methods are assigned, deleted, restored, or defined dynamically.
  * Preserved correct generic fallback behavior when optimized assumptions no longer apply.
  * Improved initialization of public fields, including fields that capture `this`.

* **Tests**
  * Added comprehensive coverage for cross-module method optimization, prototype mutations, proxies, inheritance, accessors, and garbage collection.

### DIFF
--- a/benchmarks/compiler_output/workloads.toml
+++ b/benchmarks/compiler_output/workloads.toml
@@ -2076,6 +2076,7 @@ allowed_hot_loop_runtime_calls = [
   "js_for_in_keys_stable_value",
   "js_gc_loop_safepoint",
   "js_in_operator",
+  "js_value_typeof",
 ]
 
 [workloads.for_in_stable_keys.vectorization]
@@ -2091,6 +2092,8 @@ allowed_missed_reason_kinds = [
   "unsupported_instruction",
   "unsupported_reduction",
 ]
+
+[workloads.for_in_stable_keys.runtime_budgets]
 
 [[workloads.for_in_stable_keys.stdout_checks]]
 name = "for_in_stable_keys_checksum"
@@ -2114,3 +2117,40 @@ regex_none = [
   "call i64 @js_object_get_own_property_names",
 ]
 detail = "the stable arm does not directly allocate or rebuild generic key lists"
+
+[workloads.issue_8693_imported_this]
+source = "test-files/fixtures/issue_8693_imported_this/main.js"
+kind = "imported_class_method_specialization"
+allow_hot_loop_conversions = true
+allow_dynamic_property_runtime = true
+
+[workloads.issue_8693_imported_this.vectorization]
+min_vectorized_loops = 0
+scalar_baseline = "allowed: this fixture gates cross-module method dispatch, not loop vectorization"
+allowed_missed_reason_kinds = [
+  "call_instruction",
+  "control_flow",
+  "generic_not_vectorized",
+  "not_beneficial",
+  "uncountable_loop",
+  "unknown_trip_count",
+  "unsupported_instruction",
+  "unsupported_reduction",
+]
+
+[workloads.issue_8693_imported_this.runtime_budgets]
+
+[workloads.issue_8693_imported_this.native_rep_checks]
+allow_materialization_reasons = ["runtime_api"]
+
+[[workloads.issue_8693_imported_this.native_rep_checks.require_records]]
+name = "imported_registry_proven_this_selection"
+consumer = "proven_this_method_direct_call"
+notes_contains = "receiver_provenance=imported_class_metadata"
+min = 2
+
+[[workloads.issue_8693_imported_this.native_rep_checks.require_records]]
+name = "imported_registry_generic_fallback_retained"
+consumer = "proven_this_method_direct_call"
+notes_contains = "generic_dispatch_fallback=js_native_call_method_by_id"
+min = 2
--- a/changelog.d/8693-imported-this-specialization.md
+++ b/changelog.d/8693-imported-this-specialization.md
@@ -1,0 +1,9 @@
+---
+category: Performance
+title: Specialize imported methods that capture this
+---
+
+ESM import and re-export metadata now carries producer-proven method
+eligibility, allowing guarded direct calls into stable class methods that use
+`this`. Generic dispatch remains available for shadowed or mutated receivers,
+including prototype replacement, deletion, and recreation.
--- a/crates/perry-codegen/src/codegen/helpers.rs
+++ b/crates/perry-codegen/src/codegen/helpers.rs
@@ -339,6 +339,25 @@ pub(super) fn inline_hot_small_size_cap() -> usize {
     })
 }
 
+/// Give the representation-specialized method body the ordinary small-body
+/// inline bias. This lets producer-local chains such as `Registry.add ->
+/// Group.pushEntity` optimize through the second method boundary while the
+/// externally linked clone remains callable by importers.
+pub(super) fn apply_pshape_inline_policy(
+    lf: &mut crate::function::LlFunction,
+    method: &perry_hir::Function,
+    is_pshape_clone: bool,
+) {
+    if !is_pshape_clone || method.is_async || method.is_generator || method.was_plain_async {
+        return;
+    }
+    if method.body.len() <= 8 {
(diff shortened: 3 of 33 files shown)


proggeramlug:
Landed on `main` via #8737 (squash `c203c772c`), with #8735.

Validated on the merged result: all 30 lint checkers, runtime 2669/0 at `RUST_TEST_THREADS=1`, codegen 1222/0 (+8), all codegen integration suites clean. Thanks!

PR #8729, opened 2026-08-24
perf(codegen): compare proven Symbols by identity

proggeramlug:
## Summary

- record runtime Symbol constructor provenance for stable locals and module globals
- lower strict equality against a proven Symbol to raw NaN-boxed identity
- keep loose equality, reassigned locals, and erased TypeScript annotation claims on semantic runtime helpers

This is stacked on #8672. It removes the two generic `js_eq` calls identified in the codehz/ecs accumulation profile; it contains no ECS-specific pattern or source-name check.

## Correctness evidence

- `cargo test -p perry-codegen --lib`: 1220 passed, 0 failed, 1 ignored
- `python3 scripts/local_binding_type_audit.py` and `--self-test`: passed
- `python3 scripts/check_test_registration.py`: passed
- `./scripts/pre-tag-check.sh --quick`: passed
- focused Node/Perry fixture: byte-for-byte identical, including annotation lies and loose coercion
- forced verified GC Symbol stress: `B 0`, 87 copying minors, 11470 objects copied
- frozen codehz/ecs suite: 7/7 tests and checksum `50005000`
- generated LLVM: both `readSimpleColumnValue` bodies use `icmp eq i64`, with no `js_eq`

## Mac mini performance

Apple M1, Node 26.5.1, repeat 256, 2 warmups, 6 measured batches, 11 alternating process pairs, taskpolicy fixed, 22/22 process oracles:

- #8672 control median: 0.615553 ms
- this PR median: 0.195845 ms
- median paired improvement: 68.193%, 11/11 wins

Direct Node comparison remains 0.086611 ms vs Perry 0.195961 ms, or 2.263x. This PR is a measured mechanism removal, not a claim that parity is complete.

PR #8735, opened 2026-08-24
fix(runtime): add acquire fence before dense subclass seqlock recheck

jdalton:
This fixes a data race in the dense-array subclass layout cache where a reader could combine fields from two different cache publishers and return a wrong element value instead of crashing.

<details>
<summary><strong>Root cause and fix</strong> - a relaxed payload read can be reordered past an acquire recheck on weakly-ordered hardware</summary>

`cached_dense_layout` reads two payload fields with relaxed ordering and then rechecks a sequence counter with an acquire load to make sure no writer touched the cache slot in between. An acquire load only holds back operations that come after it in the code, so it does nothing to stop the two payload reads that come before it from being reordered later by the CPU on weakly-ordered hardware like ARM. If a different cache key collides into the same slot and gets published while a reader is mid-read, the reader can end up combining one field from the old layout with the other field from the new layout. This adds an explicit acquire fence between the payload reads and the recheck, which is the standard fix for this kind of seqlock reader bug.

CodeRabbit originally flagged this exact issue on pull request `#8668`, which closed without the fix landing. This PR carries the same fix forward against current main.

`cargo check -p perry-runtime --lib` produces no diagnostics for `array/subclass.rs`. The crate currently fails to build for an unrelated, pre-existing reason: `perf_histogram.rs` calls `algebraic_sub`/`algebraic_mul` without the crate declaring `#![feature(float_algebraic)]`. I confirmed this is unrelated to this change by checking that every diagnostic file path points at `perf_histogram.rs`, never `array/subclass.rs`.

</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved array layout validation on systems with weaker memory-ordering guarantees.
  * Helps ensure reliable behavior when accessing cached array layout information.

### DIFF
--- a/crates/perry-runtime/src/array/subclass.rs
+++ b/crates/perry-runtime/src/array/subclass.rs
@@ -81,7 +81,11 @@ fn cached_dense_layout(key: u64) -> Option<DenseSubclassLayout> {
     let slots = entry.slots.load(Ordering::Relaxed);
     let bounds = entry.bounds.load(Ordering::Relaxed);
     // Recheck the seqlock before interpreting either word so readers never
-    // combine payloads from two colliding publishers.
+    // combine payloads from two colliding publishers. The fence stops the
+    // relaxed payload loads above from being reordered past this recheck on
+    // weakly ordered targets - an Acquire load alone only orders what comes
+    // after it, not what precedes it in program order.
+    std::sync::atomic::fence(Ordering::Acquire);
     if entry.sequence.load(Ordering::Acquire) != sequence {
         return None;
     }


proggeramlug:
Landed on `main` via #8737 (squash `c203c772c`), with #8729 and #8731.

Good catch, and the reasoning is exactly right — an `Acquire` *load* constrains what follows it, so it does nothing to stop the preceding `Relaxed` payload loads from sinking past the recheck on weakly-ordered hardware. A standalone `fence(Acquire)` between the reads and the recheck is the canonical seqlock-reader form. The failure shape you describe is the nasty one too: combining fields from two publishers returns a *wrong element value* rather than faulting, so it would surface as silent data corruption rather than a crash.

I added a `changelog.d/` fragment — the PR had neither one nor a `skip-changelog` label.

Validated on the merged result: all 30 lint checkers, runtime 2669/0 at `RUST_TEST_THREADS=1`, codegen 1222/0, all codegen integration suites clean.

jdalton:
@proggeramlug Fast, before I could fix the lint issue :P

PR #8736, opened 2026-08-24
fix(async): linearize await inside an async-generator finally (#8715)

### LINKED ISSUE #8715: codegen(async): `await` inside `finally` of an async generator block-waits (residual raw Await in the linearized finally state)
## Summary

An `await` inside a `finally` block of a **real async generator** (`async function*`) is lowered to a raw `Expr::Await` inside its linearized finally state instead of an async suspend, so codegen (`fs_await.rs`, `!ctx.is_async_fn`) emits the blocking busy-wait (`js_wait_for_event`). This is the same class of deadlock as #8681 (await-in-`catch`), but in a different path: the `#4438` B2-finally lowering in `crates/perry-transform/src/generator/lower/abrupt.rs`, not the `was_plain_async` catch path fixed by #8707.

## Scope

- The `was_plain_async` closure/function path (`build_async_catch_route_body_direct`) is **not** affected — the closure test `async_closure_rewrite_leaves_no_residual_await` covers `in-finally` and passes.
- Only the **real async-generator** finally path leaves the residual. Minimal repro (transform-level):

```rust
async function* g() {
  try { yield 0; }
  finally { await Promise.resolve(); }   // <- becomes raw Expr::Await in a finally state
}
```

After `transform_async_to_generator` + `transform_generators`, the finally state contains `LocalSet(__sent, Await(..))` (2 residual awaits — normal + abrupt completion copies) rather than an async suspend.

## Likely fix direction

Analogous to #8707's catch fix: route the finally through its linearized `FinallyRoute.finally_entry_state` and stop materializing the finally body's suspends as raw awaits (the finally-state emission / `build_finally_run_stmts` interaction for the async path). This touches the heavily-patched `#4438`/`#4374` abrupt-completion routing, so it wants its own change + behavioral validation (a `try { await } finally { await }` battery vs Node), separate from the catch fix.

## Detection

The regression guard already exists: `async_generator_linearizes_every_await_position` in `async_to_generator.rs` had an `await-in-finally` case (removed from that test in #8707 and pointed here). Re-adding it is the acceptance test for this issue.

Found while landing #8681/#8707 (natively-compiled Claude Code `-p` streaming deadlock).

proggeramlug:
Fixes #8715.

## Root cause

An `await` inside a `finally` of a real `async function*` compiled to a blocking busy-wait instead of an async suspend point — the finally analog of the #8681 `await`-in-`catch` deadlock that #8707 just fixed.

When a `try` in an async generator has a `finally` that yields or awaits, the linearizer (`#4438` B2-finally) splits the finally into its own dispatch states with a `finally_entry_state`. `.next()` and `.throw()` drive those states through the shared `__agstep` async-step driver, built with `async_step = true`, so a finally `await` lowers to `return AsyncStepChain(value, __agstep)` — a real microtask suspend.

The `.return()` closure did **not** go through `__agstep`. It re-drove the same states through a *separate* continuation loop built with `build_dispatch_while_body(&states, /*async_step*/ false, …)` (`lower.rs`). With `async_step = false`, `StateExit::Await` lowers to the busy-wait fallback:

```
__sent = await value;   // raw Expr::Await
continue;
```

which codegen (`expr/fs_await.rs`, `!ctx.is_async_fn`) compiles to the blocking `js_wait_for_event` / `js_unsettled_top_level_await_exit`. So a `.return()` that ran the finally — an early `break` in a `for await`, or an explicit `gen.return(v)` while suspended in the `try` — block-waited on the finally's `await`, monopolising the single runtime thread while the driver that would settle it sat suspended above, and the program deadlocked.

At the transform level the repro from the issue

```ts
async function* g() { try { yield 0; } finally { await Promise.resolve(); } }
```

left **2 residual `Expr::Await`** after `transform_async_to_generator` + `transform_generators` — both inside the `.return()` continuation clone (`LocalSet(__sent, Await(..))`).

## Fix

`.return()` no longer builds or runs an `async_step = false` dispatch loop for async generators. After `build_abrupt_routing` records the pending return (`pending_type = 2`) and jumps to `finally_entry_state`, `.return()` hands the continuation off to the shared `__agstep` driver with a fresh non-error resume:

```
return AsyncGenResume(__agstep, undefined, /*is_error*/ false);
```

exactly as `.next()`/`.throw()` already do. `__agstep` dispatches from `finally_entry_state`, runs the finally (its `yield`s settle this `.return()`'s promise, its `await`s suspend on the microtask queue), and its completion-check state re-raises the pending return as `{value, done: true}`. `wrap_generator_resume_body` clears the `executing` flag before this return, so `__agstep`'s re-entrancy guard passes.

Sync generators are untouched: they have no `await` states, so their inline busy-wait clone stays correct, and their `.return()` is a plain (non-driver) closure.

Three small edits in `crates/perry-transform/src/generator/lower.rs`: async generators build an empty `while_body_for_return` (and skip wrapping it), and the `has_yielding_finally` branch of the `.return()` body delegates to `__agstep` for async generators / keeps the inline loop for sync ones.

## Tests

- `async_generator_linearizes_every_await_position` re-adds the `await-in-finally` case that #8707 removed with a comment pointing here, plus `await-in-try-and-finally`, `await-in-try-catch-finally`, and `yield-in-finally-with-await`. All now leave **zero** residual `Expr::Await`.
- Full `cargo test -p perry-transform` is green (93 passed, 0 failed).
- Behaviorally verified byte-identical to Node v26 (compiled with the fixed `perry`, `--no-auto-optimize`) for: for-await normal completion, explicit `gen.return(v)` into an awaiting finally, `gen.throw(e)` routed through an awaiting finally, a finally that both yields and awaits, and `try`/`catch`/`finally` all awaiting. None deadlock; the pre-fix `.return()`/`.throw()` paths hung.

## Out of scope (separate pre-existing bug)

A `for await … break` did not run the finally in perry — but this reproduces on a **sync** generator with a **synchronous** finally (no `await` at all), so it is an orthogonal `for-of`/`for await` iterator-close gap (`break` doesn't invoke the iterator's `.return()`), not the await-in-finally lowering. My change is inert on that path (the `.return()` closure is never invoked), so this PR neither fixes nor regresses it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed async generators getting stuck when `.return()` enters a `finally` block containing an `await`.
  * Ensured asynchronous `finally` blocks suspend and resume correctly, including scenarios involving `try`, `catch`, and `yield`.

* **Tests**
  * Expanded coverage for async-generator `finally` blocks with awaited operations.

### DIFF
--- a/changelog.d/8715-async-gen-finally-await.md
+++ b/changelog.d/8715-async-gen-finally-await.md
@@ -1,0 +1,15 @@
+fix(async): an `await` inside a `finally` of an `async function*` no longer
+compiles to a blocking busy-wait. When a `try` in an async generator has a
+finally that yields or awaits, the finally is linearized into its own dispatch
+states, and `.next()`/`.throw()` drive those states through the shared async-step
+driver so their `await`s suspend on the microtask queue. The `.return()` closure,
+however, re-drove the same states through a separate busy-wait dispatch loop
+(`__sent = await value; continue`) — so a `.return()` that ran the finally (an
+early `break` in a `for await`, or an explicit `.return()`) block-waited on the
+finally's `await`, monopolising the single runtime thread and deadlocking. This
+is the finally analog of the #8681 `await`-in-`catch` deadlock.
+
+`.return()` now hands the continuation off to the shared `__agstep` driver
+(a fresh non-error resume) after routing the pending return into the finally,
+exactly as `.next()`/`.throw()` already do, so a finally `await` suspends
+instead of blocking. Behavior for a finally that only yields is unchanged.
--- a/crates/perry-transform/src/async_to_generator_tests.rs
+++ b/crates/perry-transform/src/async_to_generator_tests.rs
@@ -576,16 +576,49 @@ fn async_generator_linearizes_every_await_position() {
                 finally: None,
             }],
         ),
-        // NOTE: `await` inside a `finally` of a REAL async generator
-        // (`async function*`) is a SEPARATE, pre-existing gap in the
-        // `#4438` B2-finally lowering — the yielding finally's states are
-        // built with a raw `Expr::Await` instead of an async suspend, so it
-        // block-waits the same way. It is NOT addressed by this PR (which
-        // fixes the `was_plain_async` catch path); the closure test
-        // `async_closure_rewrite_leaves_no_residual_await` DOES cover
-        // `in-finally` for the `was_plain_async` path, which is clean.
-        // Tracked separately in #8715; omitted here so this test asserts
-        // only what this change fixes.
+        // #8715: `await` inside a `finally` of a REAL async generator
+        // (`async function*`). The yielding finally is linearized into its own
+        // dispatch states, but the `.return()` closure used to re-drive them
+        // through an async_step=false busy-wait loop (`__sent = await v;
+        // continue`) — a blocking wait, the finally analog of the #8681 catch
+        // deadlock. `.return()` now delegates the continuation to the shared
+        // `__agstep` driver, so the finally `await` suspends on the microtask
+        // queue and no raw `Expr::Await` survives.
+        (
+            "await-in-finally",
+            vec![Stmt::Try {
+                body: vec![y(Expr::Integer(0))],
+                catch: None,
+                finally: Some(vec![Stmt::Expr(await_(Expr::Integer(1)))]),
+            }],
+        ),
+        (
+            "await-in-try-and-finally",
+            vec![Stmt::Try {
+                body: vec![Stmt::Expr(await_(Expr::Integer(0))), y(Expr::Integer(5))],
+                catch: None,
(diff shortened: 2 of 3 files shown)


proggeramlug:
Landed on `main` via #8739 (squash `32f0eacee`), with #8738.

This closes the exact gap #8707's own test surfaced when I rebased it — it reported `await-in-finally: 2 raw await(s) survived`, and I held it saying the resolution meant deciding whether `finally` should route through its entry state the way `catch` now does. This is that decision made properly, and `async_generator_linearizes_every_await_position` passes on the merged result (I ran it explicitly, since that test is what named the gap).

Nice symmetry with #8707: same failure mode, same driver, same fix shape.

Validated: all 30 lint checkers, transform 93/0, hir 334/0, runtime 2669/0 at `RUST_TEST_THREADS=1`, codegen 1222/0.

PR #8738, opened 2026-08-24
fix(hir): aliased native-class import `new` no longer throws ReferenceError (#8730)

### LINKED ISSUE #8730: regression(hir): natively-compiled cli.js throws nameless "identifier is not defined" at init — `new <minified-local>()` scope-resolution miss
## Summary

After #8724 (captured ArrayBuffer/DataView) is applied, the natively-compiled Claude Code cli.js 2.1.112 bundle still throws at module init on nearly every command (only `--version` escapes via early exit):

```
Uncaught (in promise) ReferenceError: identifier is not defined
```

Node runs the same bundle fine. The message is the **nameless** form (`js_throw_reference_error_unresolved_get`, error.rs:1013) — emitted by `lower/expr_new.rs:1580` for `new <X>()` when `X` resolves to no lexical / function / class / import / native-module / reified-global-builtin binding. So some `new X(...)` at module-init has an `X` perry failed to resolve.

## Lead

`new $(...)` appears 5× in the bundle, and `$` is in the compile's "unknown identifier — assuming global" list. But `$` is NOT an undefined global here — it's a pervasively-used minified **local / parameter** in cli.js (`let $=` ×1014, `var $=` ×240, `($)=>`, `function($)`, `new $({type:"boolean"})`, `new $(this,h,C)`). So perry's scope resolution is failing to bind `$` to its definition in some `new $()` context and falling through to the unresolved-`new` throw. (Other unknown-list names — `Bun`/`Deno`/`window`/`document`/`Iterator`/`AsyncIterator` — are env/ES2025 globals; cli.js only reads them typeof-guarded or as strings, and `Iterator` is never used as a value, so they are not the blocker. `new ReadableStream` works — it's reified.)

## To pin the exact identifier

Instrument the throw site and re-lower the bundle (the decision is in HIR lowering, before the slow LLVM codegen, so it can be captured in the first ~2 min):

```rust
// crates/perry-hir/src/lower/expr_new.rs, just before the NewDynamic throw return
eprintln!("[nameless-new] new {class_name}(...)");
```

## Impact

Blocks running the natively-compiled Claude Code bundle end-to-end (startup). This surfaced only now because the bundle had not been run end-to-end on current `main` (the #8583 acceptance validated codegen completion, not a booting binary). Likely a perry scope-resolution gap for a class of minified identifiers, not a single-line fix.

Found while verifying #8707/#8726 against a freshly-built cc binary.

proggeramlug:
## Summary

An aliased ESM named import of a Node built-in class threw `ReferenceError: identifier is not defined` when constructed at module init — e.g. `import { BlockList as Wj4 } from "net"; new Wj4()`, `import { AsyncLocalStorage as J_z } from "async_hooks"; new J_z()`, `import { PassThrough as Lrz } from "stream"; new Lrz()`. This broke the natively-compiled Claude Code cli.js 2.1.112 bundle, which constructs all three at module init (BlockList is even `.addSubnet`-ed immediately), so nearly every command crashed with `Uncaught (in promise) ReferenceError`.

## Root cause

`lower_new` (`crates/perry-hir/src/lower/expr_new.rs`) already has an alias-rewrite block (added in #5472) that, for an aliased import of a native built-in class, rewrites the callee `class_name` from the local import name (`Wj4`) to the class's EXPORT name (`BlockList`) so the construction path matches the un-aliased form. The un-aliased native-class-import path is recognized by codegen's builtin-`New` dispatch under the literal export name.

The unresolved-`new` guard added later in #8688 (which emits `js_throw_reference_error_unresolved_get` for a `new <X>()` whose `X` binds to nothing) runs *after* that rewrite and probed `ctx.lookup_native_module(&class_name)` using the rewritten EXPORT name. But the native-module registry is keyed on the LOCAL import name (`Wj4`), so the lookup missed. None of these classes are reified global builtins, so the guard fired and threw at module init — even though the binding is perfectly resolvable. The guard was effectively defeating the alias rewrite that sits just above it.

## Fix

The guard now additionally consults the native-module registry under the original imported identifier (`source_class_name`), so an aliased native-class import is recognized as resolved and falls through to its normal construction path. Un-aliased imports and genuinely-undefined `new` targets are unchanged.

## Tests

- New `crates/perry-hir/tests/aliased_native_new_resolution.rs`: asserts aliased `BlockList`/`AsyncLocalStorage`/`PassThrough` imports do not lower to the nameless throw; un-aliased still constructs; a genuinely-undefined `new` still throws (positive control). All pass.
- `cargo test -p perry-hir -p perry-transform`: green except one pre-existing, unrelated failure (`c262_parity::assignment_named_evaluation_names_anonymous_class_identifier_rhs_only`, about class-expression naming) that fails identically on origin/main with this change reverted.
- Node-differential: the minimal repro (all three aliased imports, constructed and exercised) compiled with `perry --no-auto-optimize --enable-wasm-runtime` produces byte-identical output to `node`.

Note: full-bundle boot may surface further independent layers; this fixes the module-init nameless-`new` throw for aliased native-class imports.

Fixes #8730

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed aliased imports of Node.js built-in classes so they can be instantiated successfully.
  * Prevented valid aliased constructors from incorrectly throwing unresolved identifier errors during module initialization.
  * Preserved appropriate errors for genuinely unresolved constructors.

* **Tests**
  * Added regression coverage for aliased and unaliased native class imports.


proggeramlug:
Landed on `main` via #8739 (squash `32f0eacee`), with #8736.

Good find — an aliased import throwing `ReferenceError` at module init is the kind of thing that looks like a total failure rather than a narrow bug, and constructing `BlockList`/`AsyncLocalStorage`/`PassThrough` at init is common enough that it took out nearly every command in the cli.js bundle.

Validated: all 30 lint checkers, hir 334/0, transform 93/0, runtime 2669/0, codegen 1222/0.

PR #8740, opened 2026-08-24
perf(codegen): specialize undefined loop filters

proggeramlug:
## Summary

- version eligible instance methods with one private exact-`undefined` body when an immutable optional parameter guards work inside a loop
- retain the public boxed ABI and branch on the live argument's exact `TAG_UNDEFINED` bits; all other values run the unchanged generic body
- bound code growth to the 16 cheapest eligible methods per module and 1,024 HIR nodes per body, without composing this first implementation with typed/index clone families

The symbol-identity work from #8729 is already on current main; this branch is rebased onto `32f0eacee`, where the optional-filter branch is the next dominant cost in the release-equivalent ECS profile.

## Semantics and safety

The TypeScript optional annotation only nominates a candidate; it is never consumed as a runtime proof. The private clone receives `Type::Void` only after the public wrapper's exact tag check. Candidate discovery rejects async/generator methods, rest/`arguments` parameters, every user-authored parameter write, and closure capture. The generic body remains available for functions, `null`, `false`, `0`, empty strings, objects, and every other non-`undefined` value.

The registered parity fixture covers omitted and explicit `undefined`, a callable filter, falsey non-undefined values, non-callable `TypeError`, reassignment, closure capture, meaningful defaults, and own/prototype method overrides.

## Verification

- `cargo test -p perry-codegen --lib`: 1,226 passed, 1 ignored
- `./run_parity_tests.sh --filter guarded_undefined_method_param`: 1/1 exact Node parity
- every independent quick pre-tag audit passes; the aggregate is red only because exact current main has rustfmt drift in two untouched `perry-transform` files
- focused fixture matches Node normally and under `PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1`; the first forced collection copied 6,420 live objects / 437,064 bytes
- unchanged `codehz/ecs` benchmark: checksum 50,005,000; selected test passes; full upstream suite was separately verified 7/7 during development
- optimized `Archetype.forEachWithComponents$undef2` contains zero `js_closure_call1` calls and only the once-per-method truthiness check; the generic body contains 9 closure calls and 28 truthiness calls
- focused emitter coverage pins a producer-published `$pshape` guard wrapper as external while its `$generic` and `$undefN` implementation bodies remain internal

## Performance

M1 Mac mini, Node 26.5.1, repeat 256, two warmup and six measured rounds per process, alternating order after a 60-sample AC/CPU quiet gate:

- versus exact current main `32f0eacee`: 0.195982 ms -> 0.179393 ms median; 8.442% median paired improvement; 11/11 wins; 22/22 full 7-test/checksum process oracles
- versus Node: 0.179574 ms Perry vs 0.086134 ms Node; median paired ratio 2.083x; 22/22 output oracles
- benchmark executable: 17,147,608 -> 17,180,632 bytes (+33,024, +0.193%)

This is a measured general mechanism improvement, not a parity claim. The residual accumulation gap remains about 2.09x and will be re-profiled after this step.

### DIFF
--- a/changelog.d/8740-guarded-undefined-method-param.md
+++ b/changelog.d/8740-guarded-undefined-method-param.md
@@ -1,0 +1,11 @@
+Versioned eligible instance methods for an exact `undefined` optional argument.
+The public boxed-ABI wrapper validates the live argument bits once and sends only
+that value to a private specialized body; functions and every other falsey or
+non-callable value retain the ordinary JavaScript path. Mutation, closure capture,
+async/generator bodies, and oversized methods are conservatively excluded.
+
+This removes the per-entity optional-filter truthiness and callback arm from
+codehz/ecs's 10k-entity accumulation loop. On an Apple M1 Mac mini, 11
+alternating process pairs measured 0.179851 ms versus 0.195838 ms on the exact
+parent, an 8.126% median paired improvement with 11/11 wins and all output
+oracles passing.
--- a/changelog.d/8740-guarded-undefined-method-param.md
+++ b/changelog.d/8740-guarded-undefined-method-param.md
@@ -6,6 +6,6 @@ async/generator bodies, and oversized methods are conservatively excluded.
 
 This removes the per-entity optional-filter truthiness and callback arm from
 codehz/ecs's 10k-entity accumulation loop. On an Apple M1 Mac mini, 11
-alternating process pairs measured 0.179851 ms versus 0.195838 ms on the exact
-parent, an 8.126% median paired improvement with 11/11 wins and all output
+alternating process pairs measured 0.179437 ms versus 0.195817 ms on the exact
+parent, an 8.376% median paired improvement with 11/11 wins and all output
 oracles passing.
--- a/crates/perry-codegen/src/codegen/artifact_context.rs
+++ b/crates/perry-codegen/src/codegen/artifact_context.rs
@@ -1,0 +1,64 @@
+//! Borrowed inputs for the module artifact-emission phase.
+
+use std::collections::{HashMap, HashSet};
+
+use perry_hir::Module as HirModule;
+
+use crate::module::LlModule;
+use crate::strings::StringPool;
+
+use super::opts::CrossModuleCtx;
+
+/// Read-only view of the `CompileOptions` fields that artifact emission still
+/// references after the pipeline has moved other fields into `CrossModuleCtx`.
+pub(super) struct OptsView<'a> {
+    pub(super) import_function_prefixes: &'a HashMap<String, String>,
+    pub(super) imported_classes: &'a [super::opts::ImportedClass],
+    pub(super) is_entry_module: bool,
+    pub(super) non_entry_module_prefixes: &'a [String],
+    pub(super) output_type: &'a str,
+}
+
+/// Data computed by the `compile_module` prelude and borrowed by the artifact
+/// tail. Keeping it together avoids a second oversized compiler entry module.
+pub(super) struct ModuleArtifactsCtx<'a> {
+    pub progress: &'a super::CompileProgress,
+    pub llmod: &'a mut LlModule,
+    pub target_triple: &'a str,
+    pub strings: &'a mut StringPool,
+    pub hir: &'a HirModule,
+    pub import_function_prefixes: &'a HashMap<String, String>,
+    pub imported_classes: &'a [super::opts::ImportedClass],
+    pub is_entry_module: bool,
+    pub non_entry_module_prefixes: &'a [String],
+    pub output_type: &'a str,
+    pub module_prefix: &'a String,
+    pub class_table: &'a HashMap<String, &'a perry_hir::Class>,
+    pub class_ids: &'a HashMap<String, u32>,
+    pub enum_table: &'a HashMap<(String, String), perry_hir::EnumValue>,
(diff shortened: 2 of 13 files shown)


proggeramlug:
Not quite mergeable yet — it needs a rebase, and one of the two conflicts is a real semantic question rather than bookkeeping.

## Why it conflicts

The branch carries its own copies of two changes that have since landed on `main` as squashes with different SHAs:

```
28ddacf4f docs(changelog): record undefined filter versioning   <- #8740, yours
b317fd9cf perf(codegen): specialize undefined loop filters      <- #8740, yours
8fe1e0490 docs(changelog): record symbol identity fast path     <- #8729, landed in c203c772c
ec1112ea1 perf(codegen): compare proven symbols by identity     <- #8729, landed in c203c772c
e03d82f13 perf(runtime): restore method-scoped prototype guards <- #8672, landed in 8beca2f29
```

Merging the branch whole gives 6 conflicted files, almost all of that duplication. Cherry-picking just your two commits onto `main` cuts it to **two** files, which is the real surface: `codegen/method.rs` and `codegen/mod.rs`.

## `codegen/mod.rs` — clean union, no interaction

Your 18 lines (`guarded_undefined_method_candidates.retain(...)`, the sort, the `.take(16)`) and `main`'s 54 lines (#8731's pshape tower routing and imported-class capability publishing) are independent concerns in the same region. Both should simply be kept.

## `codegen/method.rs` — this one is yours to decide

The conflict is one condition, and it controls **linkage**:

```rust
// main (from #8731)                     // your branch
if ptr_array_cache_clone                 if is_pshape_clone
    || is_index_clone                        || is_index_clone
    || typed_public_trampoline.is_some()     || typed_public_trampoline.is_some()
    || force_generic_body                    || force_generic_body
                                             || guarded_undefined_param.is_some()
```

#8731 deliberately narrowed `is_pshape_clone` → `ptr_array_cache_clone` here, with the rationale in the comment it added: plain `$pshape` clones are producer-published capabilities that now need **external** linkage so importing modules can make guarded calls, while only the stricter array-cache clone stays module-local.

The naive union keeps `guarded_undefined_param.is_some()` in that list, which forces an undefined-filter clone to stay module-local. That's fine in isolation — but your `retain` excludes the typed and index families and **not** pshape, so a method can be both a pshape clone and an undefined-filter candidate. For that method the union would make it module-local even though #8731 now expects pshape clones to be externally linkable, and an importing module's guarded call would have nothing to bind to.

I didn't want to guess which way that should go. Either the undefined-filter family should also be excluded from pshape candidacy (mirroring how you already exclude typed/index), or it should be externally linkable like pshape now is — but that's a call about how your clone family composes with #8731's capability publishing, and you can see the whole design.

## Everything else looks good

The safety framing is the part I'd have pushed back on and it's already right: the TypeScript optional only *nominates* a candidate, and the private clone gets `Type::Void` only behind the public wrapper's live bit-compare — I confirmed `icmp_eq(I64, &arg_bits, TAG_UNDEFINED_I64)` is actually emitted, and that there's a test asserting the wrapper contains it. Candidate discovery rejecting parameter writes and closure capture is the right conservatism, and the allowlist rationale you wrote for the falsy-local fold is accurate.

Rebase onto `32f0eacee`, resolve that one condition, and I'll re-run the full 30-checker gate set plus the lib suites and land it.

proggeramlug:
Rebased onto `32f0eacee` and resolved the #8731 composition in favor of an externally linkable guarded `$pshape` wrapper with private `$pshape$generic` / `$pshape$undefN` implementation bodies. The ordinary non-versioned `$pshape` linkage rule remains unchanged. I removed the emitter’s `internal` switch so a published wrapper cannot silently become local, and added `a_pshape_family_guard_wrapper_is_a_published_capability` to pin the exported wrapper plus both routed bodies. Focused tests pass 4/4; the rebased full `perry-codegen` suite passes 1,225 with one ignored. I am rebuilding exact current-main/candidate ECS artifacts and will refresh the performance numbers before calling the rebase ready.

proggeramlug:
Landed on `main` via #8741 (squash `f444a51bd`).

You resolved the linkage question the right way round, and the test you added — `a_pshape_family_guard_wrapper_is_a_published_capability` — pins exactly the invariant I was worried about. Excluding index and array-cache clones from undefined-filter candidacy, rather than forcing the family module-local, keeps #8731's published-capability contract intact.

I also verified the safety framing rather than taking it from the description: `icmp_eq(I64, &arg_bits, TAG_UNDEFINED_I64)` is genuinely emitted and pinned by a test, so the TypeScript optional really is only nominating a candidate.

One fix on top: `pshape_symbol_reachability` scans the source tree for `$pshape` fragments outside a 7-entry allowlist, and your new `guarded_undefined_method_tests.rs` names the fragment in its wrapper assertions. I allowlisted it with a rationale, alongside the two test files already there — the gate's emission-site coverage is unchanged and `string_pool.rs` stays deliberately absent.

**One thing worth knowing:** your three new `debug_assert!`s won't execute under `perry-dev` or `release` — both set `debug-assertions = false`, so they document the invariant rather than check it. The test is what actually enforces it. Not a problem, just don't rely on them firing in CI.

Validated on the merged result: all 30 lint checkers, codegen 1226/0, all codegen integration suites clean, runtime 2669/0 at `RUST_TEST_THREADS=1`, transform 93/0.

PR #8742, opened 2026-08-24
docs: regenerate API reference

proggeramlug:
Regenerates the checked-in API reference and TypeScript declarations after the FFI manifest changes. This clears the current release-blocking API-doc drift check.

Validation:
- generated output matches perry --print-api-manifest for the current 0.5.1519 tree
- SKIP_COMPILE_GATES=1 BASE_SHA=origin/main scripts/run_lint_gates.sh (53/53 passed)
- cargo fmt --all -- --check
- git diff --check

## Summary by CodeRabbit

- **Documentation**
  - Updated the API reference manifest to include the new `ffi` module and its available methods and property.
  - Refined `bun:ffi` API declarations, including the `read` export and related entries.
  - Updated documented API coverage counts to reflect the expanded module and entry listings.

### DIFF
--- a/docs/api/perry.d.ts
+++ b/docs/api/perry.d.ts
@@ -1,6 +1,6 @@
 // Auto-generated from Perry's API manifest (#465). Do not edit by hand.
 // Source: perry-api-manifest::API_MANIFEST
-// Coverage: 2059 entries across 134 modules
+// Coverage: 2065 entries across 135 modules
 
 type PerryI8 = number & { readonly __perryI8?: never };
 type PerryI16 = number & { readonly __perryI16?: never };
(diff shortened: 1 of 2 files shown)


proggeramlug:
Landed on `main` via #8744 (squash `9ec625052`).

I verified this against the manifest rather than by re-running the generator — `regen_api_docs.sh` needs a `cargo build --release -p perry` that I couldn't justify with an agent already building and ~5 GB free. `crates/perry-api-manifest/src/entries/part_1.rs` carries exactly six `ffi` entries (dlopen, getRawPointer, toArrayBuffer, toBuffer, toString, plus `property("ffi", "suffix")`), and the generated `.d.ts` exposes precisely those six. One-to-one, nothing extra either way.

The disappearing `@perryStub … (#6562)` markers also check out as correct rather than accidental: #8704 implemented that surface and closed #6562, and the remaining #6562 mentions in the manifest are comments describing the implementation, not stub annotations.

Stated plainly so it's on the record: this confirms the output is *consistent with the manifest*, not that it's byte-for-byte what the generator emits today. The CI drift check remains the authority there.

Added a `changelog.d/` fragment — the PR had neither one nor a `skip-changelog` label. All 30 lint checkers pass.

PR #8756, opened 2026-08-24
test(gc): accept enriched root-spill diagnostic

proggeramlug:
## Summary

- match the stable semantic portion of the root-spill diagnostic
- tolerate the newer `(incl. call-result temporaries)` detail inserted between the root count and shadow-frame wording

## Release blocker

The current `main` cargo suite reaches the expected root-spill behavior but fails because this assertion still expects the old contiguous wording.

## Verification

- `git diff --check`
- clean-runner evidence: CI run 32739246257 compiled and ran the fixture, emitted five current-form root-spill diagnostics, then failed only this assertion

## Summary by CodeRabbit

* **Tests**
  * Updated spill-detection test expectations to reflect revised compiler diagnostic wording.
  * Diagnostics now refer to roots being kept in a shadow frame instead of statepoints.


proggeramlug:
Superseded by #8754: the one-line assertion fix is now commit 85f149e49 on the combined release-blocker candidate, so cargo/gap/GC/parity can validate one exact tree.

PR #8761, opened 2026-08-24
fix(runtime): honor exotic indices in Array shift

proggeramlug:
## Summary

- keep the dense `Array.prototype.shift` memmove fast path for ordinary arrays while routing observable indexed properties through live `HasProperty` / `Get` / `Set` / `Delete` operations
- translate the runtime's internal hole sentinel to JavaScript `undefined` on the dense return path
- root the receiver and carried values across accessors, then revalidate frozen and non-writable-length state before the final length update
- add end-to-end regression coverage for inherited indices, holes, freezing getters, and getters that make `length` non-writable

## Test262 impact

Pinned test262: `4249661388e5d3f92a85186213da140a6481490f`

- `built-ins/Array/prototype/shift`: **15 pass / 4 runtime-fail** before, **19 pass / 0 fail** after
- full `built-ins/Array`: **2,510 pass / 17 known runtime-fail / 0 compile-fail** after; failure-set comparison found no additions
- the aggregate full-slice count also stopped reproducing one unrelated flaky huge-length allocation failure; this PR claims only the four shift fixes

## Validation

- `cargo test --release -p perry-runtime --lib array:: -- --test-threads=1` (234 passed)
- `cargo test --release -p perry --test issue_5898_array_shift_exotic -- --nocapture`
- focused and full test262 Array sweeps above
- `cargo fmt -p perry-runtime -p perry -- --check`
- `python scripts/raw_handle_debt.py --no-raise-vs origin/main`
- file-size gate and `git diff --check`

No version, lockfile, CLAUDE, or changelog changes.

Refs #5898

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved `Array.prototype.shift` behavior for sparse arrays, inherited properties, getters, frozen arrays, and non-writable lengths.
  - Dense arrays now correctly expose missing elements as `undefined`.
  - Preserved JavaScript-compatible errors and side effects during shift operations.

- **Tests**
  - Added regression coverage for exotic array behavior, holes, getter interactions, and error conditions.

### DIFF
--- a/crates/perry-runtime/src/array/push_pop.rs
+++ b/crates/perry-runtime/src/array/push_pop.rs
@@ -1110,8 +1110,19 @@ pub extern "C" fn js_array_shift_f64(arr: *mut ArrayHeader) -> f64 {
             return TAG_UNDEFINED_F64;
         }
 
+        // A raw memmove is only equivalent to Shift when every observable
+        // indexed operation is an ordinary dense-array access. Indexed
+        // descriptors and prototype properties require the specified live
+        // HasProperty/Get/Set/Delete order; their accessors can also freeze the
+        // receiver or make `length` non-writable before the final length Set.
+        if crate::array::array_iteration_is_exotic(arr) {
+            return shift_array_spec_path(arr);
+        }
+
+        // `TAG_HOLE` is an internal storage sentinel. Even on the dense path,
+        // Get(O, "0") must expose it as `undefined`.
+        let value = crate::array::js_array_get_f64(arr, 0);
         let elements_ptr = (arr as *mut u8).add(std::mem::size_of::<ArrayHeader>()) as *mut f64;
-        let value = *elements_ptr;
 
         // Shift all elements down
         // GC_STORE_AUDIT(BARRIERED): shift memmove is followed by layout/barrier rebuild.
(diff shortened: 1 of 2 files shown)


proggeramlug:
Not landable yet — the two red `gap-suite` shards are **four real parity regressions**, not flakes. Here's everything I found, including the easy half.

## The `lint` failure is trivial
```
##[error]This PR changes crates/ but adds no changelog.d/ fragment.
```
That's it. Add `changelog.d/8761-<slug>.md` (or the `skip-changelog` label) and `lint` goes green. I have a fragment written if you'd like it, but it's a one-file add.

Worth noting for my own part: my local run of the repo's lint checkers passed, because this particular check is an inline shell step in the workflow rather than one of the `scripts/*.py` gates I mirror locally. So "all local gates green" was never going to catch it.

## The gap failures are the blocker
`gap-suite (5)` and `(6)` each report regressions against the committed `test-parity/gap_snapshot.json` — i.e. tests the snapshot says **should pass**:

**shard 5**
```
- test_gap_generic_class_constructor_name_7632: pass -> parity_fail
```
**shard 6**
```
- test_gap_4100_primitive_proto_brand_check: pass -> parity_fail
- test_gap_6320_proxy_closure_probes: pass -> parity_fail
- test_gap_array_iterator_next: pass -> parity_fail
```

`test_gap_array_iterator_next` is the one you'd expect from a `shift` change. The other three — a primitive-prototype brand check, proxy closure probes, and a generic class constructor name — are not obviously related, and that breadth is what makes me think this is worth a careful look rather than a quick patch. If routing indexed access through live `HasProperty`/`Get`/`Set`/`Delete` touched a shared helper, the blast radius could be wider than `Array.prototype.shift`.

I checked the obvious alternative explanation and it doesn't hold: your base is only one commit behind `main`, and that commit (#8762) is release-workflow YAML and publish scripts, which cannot move runtime parity.

## One thing I'd ask you not to do
The harness helpfully prints:
```
Fix the regressions, then accept the rest with:
  UPDATE_SNAPSHOT=1 ./scripts/run_gap_tests.sh
```
Please fix the four rather than re-baselining them — `parity_fail` means Perry's output no longer matches Node byte-for-byte, so accepting the snapshot would bake in whatever divergence this introduced. If any one of them turns out to be a case where Perry was previously *wrong* and now matches Node, that's a legitimate snapshot update, but it's worth saying so explicitly with Node's actual output for that case.

The Rust side looks good otherwise: it merges cleanly onto current `main`, all the script-based lint gates pass, and the rooting is done properly — `RuntimeHandleScope` with `across_mut` + `with_mut_ptr` around the accessor calls, which is the right #7341 shape for values carried across user code.

Ping me once the four are green and I'll re-run the full gate set and land it.

proggeramlug:
Landed on `main` via #8824.

Correcting my earlier block on this PR: I held it on four gap-suite regressions it did not cause. Its CI ran against a `main` that was carrying class-instance prototype-chain breakage — sibling PRs #8771 and #8776 on the same window show the identical failure set, and `main` was fixed within hours by #8766, #8778 and #8780/#8781. A full 580-test A/B against clean `main` shows zero divergences on either arm. The only real blocker was the missing changelog fragment, now added. Sorry for the delay.

proggeramlug:
Landed via #8824.

PR #8763, opened 2026-08-24
fix(runtime): resolve native-constructor superclass prototype in dynamic extends (#8760)

### LINKED ISSUE #8760: regression(runtime): natively-compiled cli.js throws "Class extends value does not have valid prototype property" at init (cc-boot layer 4)
## Summary

After #8726/#8738/#8739 (which fixed the earlier startup blockers), the natively-compiled Claude Code cli.js 2.1.112 bundle boots further but still throws at module init on nearly every command (only `--version` escapes via early exit):

```
Uncaught (in promise) TypeError: Class extends value does not have valid prototype property
```

Node runs the same bundle fine. This is the next layer in the cc-boot chain (each fix reveals the next; agents on #8715/#8730 flagged this would happen).

## Throw sites

`crates/perry-runtime/src/object/class_registry/parent_static.rs:186` and `state.rs:758` — the runtime dynamic class-parent registration. `state.rs:758` fires for `class X extends <function-valued parent>` (a closure) when `class_parent_prototype_bits(closure_get_dynamic_prop(parent, "prototype"))` returns None; `parent_static.rs:186` fires on the bound-function branch when the read `prototype` isn't object-or-null.

## What's ruled out

`class X extends {PassThrough,Transform,Readable,Writable}` (stream), `class X extends EventEmitter`, and the aliased form `import {PassThrough as PT}; class X extends PT` all work correctly on perry (byte-identical to node). So it's a MORE specific parent shape — likely another function-valued/aliased superclass whose `.prototype` perry doesn't expose, in the spirit of #8738 (which fixed the `new <aliased-native>` analog). Pinning the exact `extends` target needs instrumentation.

## To pin (fast — runtime-only, so the codegen cache stays valid and re-link is ~49s)

Add a diagnostic at both throw sites printing the parent value's shape (is_closure_ptr? its `name`/`prototype` bits? whether it's a native-class registry entry?), rebuild the runtime staticlibs coherently, re-link the cached cli.js object, and run `HOME=/tmp/cc_home cc-accept --help` to capture the culprit. Then build a reduced synthetic (`class X extends <that shape> {}`) that reproduces, and fix so perry exposes a valid prototype for it (matching node).

## Impact

Blocks booting the natively-compiled Claude Code bundle — layer 4 of the startup chain. Found while verifying #8726/#8738/#8739 against a freshly-built cc binary from origin/main.

proggeramlug:
Fixes #8760.

## Symptom

Booting the natively compiled Claude Code cli.js 2.1.112 bundle threw at module init on nearly every command (only `--version` escaped):

```
Uncaught (in promise) TypeError: Class extends value does not have valid prototype property
```

Node runs the same bundle fine. This is layer 4 of the "boot the natively-compiled cli.js bundle" chain (after #8726, #8738, #8739).

## Root cause

The bundle extends the aliased `events.EventEmitter` ESM export at module init (`import { EventEmitter as EE } from "events"; class X extends EE {}` — minified `class q extends <alias>`). Such native-module constructor exports are modeled as bound-method closures whose `.prototype` object — the one carrying the EventEmitter method surface — is materialized **lazily**, only when a `.prototype` read (or a synthetic class id) demands it.

The dynamic-parent branch of `class_decl_prototype_value` (runtime `class_registry/state.rs`) resolved the superclass prototype with a **raw** `closure_get_dynamic_prop(parent, "prototype")`. For a freshly minted EventEmitter export that slot is still `undefined`, so `class_parent_prototype_bits` returned `None` and the runtime threw — even though Node's `Get(superclass, "prototype")` returns a real object.

A diagnostic at the throw site confirmed the exact shape: `name="EventEmitter"`, `proto_bits=TAG_UNDEFINED`, `native_mod=Some(("events","EventEmitter"))`. The static-codegen extends path (`parent_static.rs`) already early-returns for bound native-callable parents; only the dynamic path was missing the lazy resolution.

## Fix

When the raw dynamic slot is not a valid prototype, resolve the superclass prototype through `ordinary_function_prototype_value_for_read` — the **same** lazy materialization an ordinary `Y.prototype` property read uses (it mints the synthetic prototype, installs the EventEmitter method surface, and caches it on the closure). The `extends` edge then links to the real prototype. A genuinely prototype-less parent (a bare `fn.bind(...)`, an arrow/async function) still resolves to `None` and throws below, matching Node — the change is purely additive over the previous throw.

As a side benefit, populating the closure's `.prototype` slot during this resolution also fixes the sibling `class X extends <EE-via-var>` / `<namespace-member>` shapes, which previously linked to the wrong prototype (broken `instanceof`).

## Verification

- Reduced synthetic (`import { EventEmitter as EE } from "events"` — direct-alias, via-var, and namespace-member `extends`): pre-fix the direct-alias form threw the exact TypeError and the other two produced a broken prototype chain; post-fix all three are **byte-identical to Node v26** (`chain: true | instanceof: true | emit: true`).
- Relinked cc-test: `--help` now prints the full usage and exits 0 (previously threw at module init); `--version` still prints `2.1.112 (Claude Code)`. No residual `Class extends` / `Uncaught` error; no new next-layer error surfaced for `--help`.
- New runtime regression test `native_constructor_export_parent_links_declared_prototype` drives the exact dynamic-parent path and asserts the declared prototype links to `EventEmitter.prototype`.
- Tests green (release): class_registry 14/0 (incl. the new test), object:: 208/0, closure 67/0, node_stream 85/0, native_module 22/0. One unrelated `native_module` test (`tls_constructor_prototypes_match_node_parent_classes`) flaked under parallel execution and passes in isolation / single-threaded — the documented perry-runtime parallel-test flakiness, not this change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue preventing classes from extending certain native constructors, such as `events.EventEmitter`.
  - Native constructor prototypes are now resolved correctly during inheritance.
  - Prototype-less or invalid constructors continue to produce the expected error behavior.


proggeramlug:
Landed on `main` via #8766 (squash `1b0c300e0`). That's the fourth cli.js-breaking bug fixed today — good find.

One fix needed on top: `class_id_collisions.py` went red. It matches any `*CLASS_ID: u32 =` and **deliberately ignores scope** — its header says the point is "not to depend on someone remembering the scope". Your new test declared a function-local `const CLASS_ID` alongside the existing #8001 fixture's, so two same-named constants held different values (`0x7d01_8001` vs `0x7d01_8760`) and the gate read that as a drifted cross-crate mirror.

Since ignoring scope is intentional there, I renamed yours to `ISSUE_8760_CLASS_ID` rather than working around the check. The #8001 test is untouched. Worth knowing for future fixtures in that file — a bare `CLASS_ID` will collide with the gate even inside a test fn.

Validated on the merged result: all 30 lint checkers (including `class_id_collisions.py` after the rename), runtime 2674/0 at `RUST_TEST_THREADS=1`, and your `native_constructor_export_parent_links_declared_prototype` verified running by name.

PR #8764, opened 2026-08-24
fix(windows): keep WinUI callbacks behind scanned keys

### LINKED ISSUE #8734: perry-ui-windows-winui: callback pointers copied into Reactor closures escape the root scanner
Landed with #8722 (via #8733) and documented at both scanners, but not fixed there.

`perry-ui-windows-winui` now registers a GC root scanner over its `NODES` / `ON_ACTIVATE` / `ON_TERMINATE` / `PENDING_TIMERS` tables, so the *stored* callback pointers are marked and rewritten. That is not sufficient, because several paths copy the unboxed pointer **out** of the scanned table and into a boxed Rust closure the collector cannot reach:

- `render_handle` operates on a **clone** of the node and moves the raw callback pointer into `move` closures owned by the Reactor element tree.
- `start_runtime_pump` does the same for `PENDING_TIMERS` into `DispatcherTimer` closures.
- `app_run` does the same for `ON_TERMINATE` into an `on_exit` closure.

Those copies live inside `Box<dyn Fn…>` allocations that no root scanner can visit or rewrite, so after an evacuating collection they point into from-space. The failure mode is the usual one for this class: nothing at collection time, then `TypeError: value is not a function` (or a segfault) whenever the widget/timer/exit hook next fires.

**Scanning the tables cannot fix this.** The fix is to stop capturing the value and instead capture a *key*, then re-read the scanned slot at invoke time — the indirection `perry-ui-macos` already gets from its handle-keyed maps. That is roughly a 15-site refactor across `widgets.rs` and `app.rs`.

Two notes on scope:
- This is a pre-existing property of the WinUI backend's design, not something the scanner work introduced. Before #8722 the tables were not scanned at all, so this was strictly worse.
- It cannot be reproduced or verified from macOS — the crate is Windows-only and a cross-compile fails on a C build script (`perry-audio-miniaudio` needs Windows headers). Whoever picks this up will need a Windows host, and `PERRY_GC_SCHEDULE_SEED` + `PERRY_GC_PROTECT_FROMSPACE` are the instruments that would turn it into an immediate fault rather than a delayed one.

proggeramlug:
## Summary
- keep raw WinUI widget callback pointers solely in the GC-scanned `NODES` table and have Reactor handlers resolve them by widget handle and callback slot at invocation
- keep timer callbacks in a scanned key-indexed table, and make application-exit and timer closures re-read live slots after a moving collection
- cover every widget callback slot plus lifecycle and timer indirection with Windows unit tests

## Related issue
Fixes #8734

## Test plan
- [x] `cargo fmt -p perry-ui-windows-winui -- --check`
- [x] `cargo check -p perry-ui-windows-winui --all-targets`
- [x] `cargo build --release -p perry-ui-windows-winui`
- [x] `$env:RUSTFLAGS = '-C link-arg=/FORCE:MULTIPLE'; cargo test -p perry-ui-windows-winui --lib reactor_callback_keys` (2 passed)
- [x] `python scripts/workspace_architecture.py --check --print-summary`
- [x] GC/runtime audits: store-site inventory, address-class inventory, snapshot self-test, test registration, runtime root holders, and scanner latches
- [x] `git diff --check`
- [x] no version bump or aggregate changelog edits

## Notes
The focused tests use `/FORCE:MULTIPLE` locally because ordinary Windows lib-test linking currently encounters pre-existing duplicate `perry_audio_*` exports between `perry-ui-windows` and `perry-audio-miniaudio` (LNK2005). Clippy is also blocked before reaching this crate by the existing mistyped `0x0071_7263_i64` suffix in `perry-ui-windows/src/widgets/qrcode.rs`. Package formatting, all-target checking, the release build, and the focused regression tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed WinUI widget interactions, including buttons, text fields, toggles, sliders, and secure fields, so callbacks continue working reliably after garbage collection.
  * Improved timer, application activation, and application-exit callbacks to prevent stale or missed event handling.
  * Ensured event-driven UI behavior remains stable while applications are running and updating.

### DIFF
--- a/changelog.d/8734-winui-callback-indirection.md
+++ b/changelog.d/8734-winui-callback-indirection.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Fixed WinUI widget, application-exit, and timer callbacks becoming stale after an evacuating garbage collection. Windows Reactor closures now retain stable widget/slot keys and resolve callbacks from GC-scanned storage immediately before invocation; timer callbacks likewise remain in a scanned key-indexed table instead of being copied into opaque `DispatcherTimer` closures.
--- a/crates/perry-ui-windows-winui/src/app.rs
+++ b/crates/perry-ui-windows-winui/src/app.rs
@@ -33,6 +33,7 @@ thread_local! {
     static APPS: RefCell<Vec<AppState>> = const { RefCell::new(Vec::new()) };
     static ON_ACTIVATE: RefCell<Option<usize>> = const { RefCell::new(None) };
     static ON_TERMINATE: RefCell<Option<usize>> = const { RefCell::new(None) };
+    static TIMER_CALLBACKS: RefCell<Vec<usize>> = const { RefCell::new(Vec::new()) };
     static PENDING_TIMERS: RefCell<Vec<(Duration, usize)>> = const { RefCell::new(Vec::new()) };
     static ACTIVE_TIMERS: RefCell<Vec<DispatcherTimer>> = const { RefCell::new(Vec::new()) };
     static RUNTIME_PUMP_STARTED: RefCell<bool> = const { RefCell::new(false) };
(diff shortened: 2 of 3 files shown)


proggeramlug:
Landed on `main` via #8798 (re-applied after the #8797 revert). Content verified identical to `origin/main`.

PR #8765, opened 2026-08-24
fix(mysql2): isolate prepared operations and pool transactions

### LINKED ISSUE #8745: ext-mysql2: prepared-statement param binding races — params matched against the wrong statement
**Env:** perry 0.5.1512 built from `main@6173ff9f6` (single cargo invocation: `-p perry -p perry-runtime-static -p perry-stdlib-static -p perry-ext-mysql2 …`), macOS arm64. MySQL 8.0.46. Found in the MB24 Phase-0 dependency spike.

**Repro** (`mysql2/promise` against any MySQL 8):

```ts
import mysql from "mysql2/promise";
const pool = mysql.createPool({ /* env */ });
await pool.query("DROP TABLE IF EXISTS t");
await pool.query("CREATE TABLE t (id INT AUTO_INCREMENT PRIMARY KEY, name VARCHAR(50), cents INT)");
await pool.execute("INSERT INTO t (name, cents) VALUES (?, ?)", ["x", 100]);
const [r] = await pool.execute("SELECT * FROM t WHERE id = ?", [1]);
```

**Actual:** non-deterministic; the param list is applied to the wrong prepared statement. Two observed shapes of the same failure across runs:
- `Query failed: … prepared statement expected 1 parameters but 0 parameters were provided (sqlx_mysql::connection::executor:131)` — SELECT prepared with 1 placeholder, params dropped;
- `… expected 0 parameters but 2 parameters were provided` — the 2-param INSERT matched against a previous 0-param statement (the DROP/CREATE).

Same statements succeed on other runs — looks like a statement-cache/queue association race in the ext, not a per-statement-type gap. `pool.query("SELECT 1")` with no params is always fine.

**Expected:** params bind to the statement they were passed with, as on Node mysql2.

Full repro file: mobilitybroker24/monorepo `spike/tests/05-mysql.ts` (+ minimal versions used for this report).

### LINKED ISSUE #8746: ext-mysql2: connection from pool.getConnection() throws 'Invalid connection handle' on beginTransaction/execute
**Env:** perry 0.5.1512 built from `main@6173ff9f6`, macOS arm64, MySQL 8.0.46. MB24 Phase-0 spike.

**Repro:**

```ts
import mysql from "mysql2/promise";
const pool = mysql.createPool({ /* env */ });
const c = await pool.getConnection();   // resolves fine; c.release() alone works
await c.beginTransaction();             // throws
```

**Actual:** thrown value is the plain string `"Invalid connection handle"` (see also the separate errors-as-strings issue). Methods on the checked-out connection (`beginTransaction`, `execute`) don't see a valid underlying handle.

**Expected:** the mysql2 `PoolConnection` API — `getConnection()` → `beginTransaction()` → `execute("SELECT … FOR UPDATE")` → `commit()`/`rollback()` → `release()`.

**Impact:** this is the canonical row-lock transaction pattern (`SELECT … FOR UPDATE`), i.e. the serialization backbone for anything auction/ledger-shaped; there is currently no way to run an explicit transaction from a pool under Perry. `createConnection()`-based transactions could not be verified either (session cut short by the param-binding race, filed separately).

proggeramlug:
## Summary

- keep each mysql2 SQL string and parameter vector in one owned request, use text protocol for parameterless `query()`, and make prepared statements request-scoped so cached metadata cannot cross requests
- replace registry-backed mutable connection references with serialized owned handles, including safe close/release behavior around in-flight work
- route checked-out pool connections through `query`, `execute`, `beginTransaction`, `commit`, `rollback`, and `release`
- add focused unit coverage plus an opt-in live MySQL regression fixture covering both reported workflows

## Testing

- `cargo test -p perry-ext-mysql2 --lib` (10 passed)
- `cargo test -p perry-hir mysql2` (8 passed)
- `cargo clippy -p perry-ext-mysql2 --all-targets --no-deps -- -D warnings`
- `cargo check -p perry-ext-mysql2 --tests`
- `git diff --check`

The live MySQL fixture is parity-skipped by default because it requires a local MySQL 8 database. No version bump is included.

Closes #8745
Closes #8746

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved MySQL query reliability when mixing text queries and prepared statements.
  - Prevented overlapping operations on the same connection or pooled connection from interfering with one another.
  - Ensured pooled connections remain reserved until all active work is complete.
  - Improved support for transactions, including commit and rollback behavior.

- **Testing**
  - Added coverage for query isolation, prepared execution, concurrent connection handling, and transaction workflows.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6077,6 +6077,7 @@ version = "0.5.1519"
 dependencies = [
  "chrono",
  "perry-ffi",
+ "perry-runtime",
  "sqlx",
  "tokio",
 ]
--- a/crates/perry-ext-mysql2/Cargo.toml
+++ b/crates/perry-ext-mysql2/Cargo.toml
@@ -24,3 +24,6 @@ chrono.workspace = true
 
 [dev-dependencies]
 perry-ffi = { workspace = true, features = ["runtime-link"] }
+# Standalone extension tests need the runtime half of the test-only async FFI
+# shims; production code still depends on perry-ffi only.
+perry-runtime = { workspace = true, features = ["default", "stdlib"] }
--- a/crates/perry-ext-mysql2/src/lib.rs
+++ b/crates/perry-ext-mysql2/src/lib.rs
@@ -15,15 +15,20 @@
 //! adapter; followup once a wrapper actually demands it).
 
 use perry_ffi::{
-    alloc_string, build_object_shape, get_handle_mut, js_array_alloc, js_array_get, js_array_push,
+    alloc_string, build_object_shape, js_array_alloc, js_array_get, js_array_push,
     js_object_alloc_with_shape, js_object_get_field, js_object_set_field, register_handle,
-    spawn_blocking, take_handle, ArrayHeader, Handle, JsPromise, JsValue, ObjectHeader, Promise,
-    StringHeader,
+    spawn_blocking, take_handle, with_handle, ArrayHeader, Handle, JsPromise, JsValue,
+    ObjectHeader, Promise, StringHeader,
 };
 use sqlx::mysql::{MySqlConnection, MySqlPool, MySqlPoolOptions, MySqlRow};
 use sqlx::pool::PoolConnection;
 use sqlx::{Column, Connection, MySql, Row, TypeInfo};
+use std::sync::Arc;
 use std::time::Duration;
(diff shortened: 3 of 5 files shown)


proggeramlug:

Added a `changelog.d/` fragment — the PR had neither one nor a `skip-changelog` label.

PR #8767, opened 2026-08-24
perf(codegen): admit forwarded arrays in versioned loops

proggeramlug:
## Summary

- follow one validated forwarding edge when admitting arrays into version-stable indexed loops
- canonicalize the compiler-private local to the live array after the full header/fingerprint check
- retain the per-iteration fingerprint guards, so callback-driven growth or GC still side-exits before the next effect
- fail closed to the generic loop for invalid targets or longer forwarding chains

Array growth preserves JavaScript identity with a forwarding stub. The versioned-loop admission added in #8755 inspected that stub as though it were the live array, so grown arrays never entered the fast loop. Ordinary guarded indexed reads already follow one forwarding edge; this change gives loop admission the same safe behavior.

## Correctness

- cargo test -p perry-codegen --lib: 1230 passed, 0 failed, 1 ignored
- cargo test -p perry --test versioned_indexed_loop_forwarding: passed
- the executable regression runs normally and with PERRY_GC_FORCE_EVACUATE=1
- the regression also grows a retained column from inside the callback and verifies safe generic-loop resumption

## Performance

Apple M1 Mac mini, macOS 26.5.1, Node 26.5.1. ECS comprehensive benchmark, 10k entities: forEach query with accumulation, repeat 256, 11 alternating Node/Perry process pairs under taskpolicy after a 4.10-11.37% CPU quiet gate.

- current main Perry median: 0.181297 ms
- this branch Perry median: 0.106787 ms
- reduction: 41.1%
- Node median in the branch confirmation: 0.085953 ms
- branch Perry/Node ratio: 1.2424x
- semantic oracles: 22/22 passed

The remaining gap is separately profile-backed: the specialized loop is now active, with residual samples split between its three mutation guards and the trusted callback. This PR is the isolated forwarding-admission fix, not a benchmark-specific shortcut.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved indexed-loop handling for arrays relocated during memory management.
  * Added validation and safe fallback behavior when forwarded arrays are invalid or unsafe.
  * Preserved correct behavior when callback operations grow arrays during iteration.

* **Tests**
  * Added regression coverage for forwarded arrays, forced memory evacuation, and callback-driven array growth.

### DIFF
--- a/crates/perry-codegen/src/codegen/index_method_clone_tests.rs
+++ b/crates/perry-codegen/src/codegen/index_method_clone_tests.rs
@@ -516,6 +516,59 @@ fn checked_reader_callback_loop_versions_to_fast_and_resumable_slow_bodies() {
 }
 
 #[test]
+fn versioned_checked_reader_admission_canonicalizes_one_forwarding_edge() {
+    let ir = emit_versioned_checked_reader_loop();
+    let iterate = function_body(
+        &ir,
+        "@perry_method_versioned_checked_reader_loop_ts__Reader__iterate(",
+    );
+    let source_guard = iterate
+        .split("\nversioned_index.array.source_deref.")
+        .nth(1)
+        .and_then(|body| body.split("\nversioned_index.array.live_deref.").next())
+        .unwrap_or_else(|| panic!("loop has no forwarding-source guard:\n{iterate}"));
+    let live_handle = source_guard
+        .lines()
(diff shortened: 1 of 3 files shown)


proggeramlug:

Added a `changelog.d/` fragment — the PR had neither one nor a `skip-changelog` label.

PR #8768, opened 2026-08-24
fix(runtime): materialize ordinary parent prototypes

proggeramlug:
Release-blocker follow-up to #8766.

The full gap shard exposed a second lazy-prototype case on current `main`: an ordinary runtime-valued function superclass (`class Child extends WithDefault`) threw `Class extends value does not have valid prototype property` unless some unrelated code had read `WithDefault.prototype` first.

#8766 correctly added lazy materialization for bound native constructor exports, but implemented it as a raw dynamic-slot probe plus an `ordinary_function_prototype_value_for_read` fallback. This follow-up routes the class-definition `Get(superclass, "prototype")` through the existing canonical `js_function_prototype_value_for_read` helper instead. That one path preserves explicit/deleted/generator prototype semantics while lazily materializing both ordinary-function and bound-native-constructor prototypes.

The existing failing regression is `test_gap_class_extends_runtime_value_super`; no snapshot or expectation is weakened.

Validation on the rebased result:

- `git diff --check`
- `cargo test -p perry-runtime class_parent_prototype_tests`: 2 passed, 0 failed
- release-profile compiler/runtime rebuilt from this branch
- exact Node/Perry output parity for `test_gap_class_extends_runtime_value_super`:

  ```text
  s.x: 7
  s.y: 9
  c.flag: default-from-parent
  w.ownField: subclass-init
  ```

No version bump and no tag.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed inheritance for subclasses of runtime-valued functions.
  - Subclasses now correctly observe and link the same prototype object returned by reading the parent function’s `prototype`, including when it has not been accessed previously.
  - Improved consistency for plain functions, bound constructors, and functions with explicit, deleted, or generator prototypes.


proggeramlug:

PR #8777, opened 2026-08-24
fix(ci): restore release readiness gates

proggeramlug:
## Summary

- install the lockfile-pinned npm dependencies before the soak/publish script tests, so `@socketsecurity/sdk` can be imported
- update the raw numeric object-field proof to require the raw-f64 scalar store record the current compiler emits
- remove the duplicated store/load contract pair (the duplicate checked the same records twice)

## Why

Two independent jobs on current `main` block a release:

- `security-audit / soak-gate` fails with `ERR_MODULE_NOT_FOUND` for `@socketsecurity/sdk`
- `compiler-output-regression` fails because it expects `ScalarThisFieldSet`, while its retained artifact contains two `ScalarObjectFieldSet` raw-f64 stores and two matching raw-f64 loads

The revised contract still requires the representation fact to be consumed; it does not relax the native `f64` proof.

## Validation

- `npm ci --ignore-scripts --no-audit --no-fund`
- `npm run test:scripts` (72 tests: 71 passed, 1 skipped)
- exact failing `raw_numeric_object_fields` artifact reverified: all 20 native-representation contracts passed; store and load selectors each matched 2 records
- `python3 -m unittest tests.test_compiler_output_regression` (73 passed)
- `BASE_SHA=origin/main scripts/run_lint_gates.sh` (all 55 gates passed)
- `git diff --check`

### DIFF
--- a/.github/workflows/security-audit.yml
+++ b/.github/workflows/security-audit.yml
@@ -114,7 +114,8 @@ jobs:
   # Soak parity gate + external-tool pin gate. Always-run (deliberately not
   # path-filtered — nub hid this gate in a path-gated job and it silently
   # skipped Rust-only PRs). The scripts are dep-free erasable-TS .mts run
-  # with the pinned Node's native type stripping; no npm install needed.
+  # with the pinned Node's native type stripping. The publish tests exercise
+  # the Socket SDK wrapper, so install the lockfile-pinned dev dependencies.
   soak-gate:
     runs-on: ubuntu-latest
     timeout-minutes: 10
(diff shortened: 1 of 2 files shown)


proggeramlug:
Landed on `main` via the PR above.

**The soak-gate failure was mine** — I introduced it in #8762 by wiring `scripts/publish/*.test.mts` into the step that already claimed to run `npm run test:scripts`. I said then that I'd verified the tests first so it couldn't redden CI, but I verified them *after* an `npm install` locally and never checked that the job installs dependencies. The soak tests don't import the Socket SDK, so the soak-only glob had never needed one. Thanks for catching and fixing it.

On the second half: I checked the contract change isn't a quiet relaxation, since removing a "duplicate" is exactly how that happens. The workload goes 19 → 17 `require_records`, with `consumed` 8 → 6 and `representation` 4 → 2 — every decrement accounted for by the two genuinely duplicated records, and the surviving store and load still require the representation fact consumed. Your description holds.

Validated: all 30 lint checkers, plus the fixed job simulated locally (`npm ci` then the combined test command → 72 tests, 71 pass, 0 fail, 1 skipped).

PR #8780, opened 2026-08-24
fix(runtime): inherit Object prototype on class instances

proggeramlug:
## Summary

- allow registered JavaScript class instances to use the ordinary Object.prototype read fallback after class methods/accessors miss
- keep unregistered native and synthetic class IDs on their existing intrinsic prototype paths
- restore String(new C()) and other inherited Object.prototype reads

## Validation

- BASE_SHA=origin/main scripts/run_lint_gates.sh (all 55 gates passed)
- release-mode test_gap_string_coercion_tostring parity against Node (pass)
- release-mode test_gap_class_extends_function_static parity against Node with PR #8768 applied (pass)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed inherited `Object.prototype` property access for declared class instances.
  * Restored correct default `toString` and `valueOf` behavior.
  * Improved ordinary string coercion for class instances.

* **Documentation**
  * Added a changelog entry describing the corrected prototype property behavior.


proggeramlug:
Landed on `main` via the PR above.

Validated on the merged result: all 30 lint checkers, runtime 2674/0 at `RUST_TEST_THREADS=1`, codegen 1230/0, all codegen integration suites clean. Thanks — and thanks for carrying the changelog fragment.

PR #8786, opened 2026-08-24
perf(codegen): version nested packed loops over closure captures

### LINKED ISSUE #8773: perf(codegen): version nested packed loops over immutable closure captures
## Summary

Perry's version-stable packed-loop optimization does not admit immutable arrays held in closure-capture slots. That excludes the actual `wolf-ecs/simple_iter` shape even after #8755 and the forwarded-array follow-up #8767: each system closes over a `Query extends Array`, loads an `Archetype extends Array` from it, then performs a nested indexed loop.

Add guarded loop versioning for immutable closure-captured packed Arrays/Array subclasses, including the nested case where the inner array is obtained from a guarded outer indexed read. Keep the generic loop as the side exit whenever capture storage, identity, layout, version, or mutation proofs fail.

```js
class Query extends Array {}

function setup(entityCount) {
  const query = new Query();
  query.push(archetype);

  const values = new Uint32Array(entityCount);

  function system() {
      const current = query[i];
        values[current[j]] += 1;
      }
    }
  }

  return () => {
    system();
    return values[0];
  };
}

const run = setup(1_000);
const iterations = 2_000;
let checksum = 0;
for (let i = 0; i < iterations; i++) checksum = run();
console.log(JSON.stringify({
  checksum,
}));
```

Node and Perry must both finish with `checksum: 2000`. The public integration case is `noctjs/ecs-benchmark` at `7b53a36606118e8b2a450a2ba4919939c86bbd2e`, `src/cases/wolf-ecs/simple_iter.js`, with `wolf-ecs` from that lockfile.

Build and retain evidence:

```bash
PERRY_NO_AUTO_OPTIMIZE=1 \
```

## Current evidence

Measured on an Apple M1 Mac mini (8 GiB, macOS 26.5.1, Node 26.5.1), AC power, after a 60-sample CPU quiet gate, three discarded process warmups, and 11 alternating Node/Perry process pairs:

| Runtime | median |
|---|---:|
| Node | 0.004938 ms/op |
| Perry `20a388974d7333ccaec9da9175fbcee54ad49902` | 0.780749 ms/op |

Perry is **158.10x slower**. Every process completed successfully.

A symbolized one-second profile synchronized to the timed phase collected 759 main-thread samples. `js_packed_arraylike_index_get` was the exclusive leaf in **413/759 samples (54.4%)**. The three system closures accounted for another 81, 76, and 69 exclusive samples.

The same full Wolf executable compiled from #8767 head `7e223e9dcaa87ad6415d85aec7ff201428e5f716` measured 0.780296 ms/op in a confirmation run: effectively unchanged. Its retained LLVM module contains **21 calls** to `js_packed_arraylike_index_get` across the three systems (seven indexed reads per system) and **zero** `stable_packed.loop.fast` / `stable_packed.loop.slow` blocks.

This is expected from the current admission rule: `packed_loop_array_binding_storage_is_addressable` explicitly rejects `ctx.closure_captures`, and `stable_packed_loop::match_candidate` separately rejects a captured `array_id`. #8767 follows one forwarding edge after a candidate has been admitted; it does not make closure captures candidates and is not a duplicate of this issue.

## Proposed direction

- Represent an immutable closure-capture slot as addressable versioned-loop storage, or copy its tagged value into a shadow-bound compiler-private local at the loop preheader.
- Admit the captured outer array only after the full packed Array/Array-subclass layout, prototype/descriptor, identity, forwarding, and version guards used by #8755/#8767.
- Reload the capture/root after any collecting operation; never retain a raw movable pointer across a safepoint.
- Let the guarded outer indexed read nominate the resulting element for a second guarded packed-loop version, so `const current = query[i]` can feed the inner loop without reverting to per-index helpers.
- Revalidate the relevant compact fingerprints before effects at the same points as the existing versioned loop and side-exit to the unchanged generic iteration on failure.
- Record selection and rejection reasons for closure-capture and nested-derived candidates in `--explain-lowering`; the current silent absence of a loop record makes this gap harder to diagnose.

## Semantic constraints

- Preserve closure binding identity and rebinding semantics; mutable captures, boxed captures, aliases that can rebind, and unknown capture layouts must remain generic.
- Preserve Array-subclass species/prototype/indexed descriptor behavior, holes, accessors, proxies, length changes, and forwarding identity.
- Preserve mutation and exception ordering. A callback/getter/proxy or loop-body effect that can invalidate the layout must revalidate or side-exit before the next indexed effect.
- Preserve moving/forced-GC correctness: capture and derived-element roots must be rewritten by GC, and raw addresses must not survive safepoints.

## Acceptance criteria

- Register the reproduction as semantic and compiler-output coverage; Node and Perry print checksum `2000`, including with `PERRY_GC_FORCE_EVACUATE=1`.
- The stable fast arms for both `query[i]` and `current[j]` contain no `js_packed_arraylike_index_get`; explicit generic fallback/continuation arms remain present and are exercised.
- `--explain-lowering` records that the closure-captured outer binding and nested derived array were selected, including guard/fallback identities.
- Negative tests cover capture rebinding, capture aliasing, holes, accessors, proxies, subclass/prototype mutation, length growth/shrink, forwarding, exceptions, and forced-moving GC.
- Re-run the full upstream `wolf-ecs/simple_iter` case with a state/checksum oracle added to the harness. On the quiet M1 alternating protocol, require at least a 1.5x Perry median improvement and at least 9/11 paired wins; report helper samples, RSS, and executable-size delta.

## Related work

- #8755 added version-stable packed loops for addressable locals/module globals.
- #8767 admits one-edge-forwarded arrays after the existing candidate check; it does not cover closure captures.
- #8692/#8708 added guarded direct `Uint32Array` RMW, which should become the inner value-update path once the packed index reads stop falling back.

proggeramlug:
## Summary

Versions stable packed Array and Array-subclass loops whose receiver is an immutable closure capture, including the nested `const current = query[i]` shape from #8773. Captured receivers are reloaded and fully revalidated each iteration; any failed identity, forwarding, layout, packedness, range, or mutation proof resumes the unchanged generic loop at the current index.

## Changes

- Add a capture-safe packed-loop guard that returns the validated live receiver address and follows one checked Array forwarding edge.
- Track immutable values derived from guarded outer indexed reads so nested loops can receive their own guarded direct-load version.
- Preserve the existing call-free path for ordinary bindings and reject reassigned captures or capture reads with a safepoint-capable prefix.
- Record capture selection/rejection, revalidation mode, nested origin, and matching guard/fallback identities in lowering artifacts.
- Add regression coverage for the exact nested Query/Archetype/Uint32Array reproduction plus rebinding, alias growth, shrink, holes, accessors, proxies, prototype descriptors, forwarding/moving GC, and exceptions.

## Related issue

Fixes #8773

## Test plan

- [ ] `cargo build --release` clean (not run; Windows workspace disk constraints)
- [ ] full affected-workspace suite passes (not run locally)
- [x] Added a focused integration test in `crates/perry/tests/issue_8773_closure_capture_packed_loops.rs`
- [x] `cargo check -p perry-codegen -p perry-runtime`
- [x] `cargo test -p perry --test issue_8773_closure_capture_packed_loops -- --nocapture`
- [x] `cargo test -p perry --test issue_8690_loop_versioned_arraylike read_only_loops_have_preheader_proofs_and_fallback_free_fast_blocks -- --nocapture`
- [x] `cargo fmt -p perry-codegen -p perry-runtime -- --check`
- [x] `rustfmt --check --edition 2024 crates/perry/tests/issue_8773_closure_capture_packed_loops.rs`

The issue test runs semantic fixtures both normally and with `PERRY_GC_FORCE_EVACUATE=1` / `PERRY_GC_VERIFY_EVACUATION=1`. It also inspects retained LLVM IR to require two stable fast versions, direct loads without `js_packed_arraylike_index_get` in fast blocks, an explicit generic helper fallback, and the expected lowering explanations.

The requested quiet Apple M1 alternating benchmark protocol was not available on this Windows host. Compiler-output coverage verifies removal of the profiled helper from both relevant fast arms; no cross-platform timing claim is made here.

## Screenshots / output

Exact reproduction output in normal and forced-moving-GC runs:

```text
2000
```

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Improved performance for loops over immutable arrays captured by closures, including supported nested array values and array subclasses.
  * Added runtime validation to safely handle captured arrays that change, move during garbage collection, or become invalid.

* **Bug Fixes**
  * Preserved correct fallback behavior for resized, sparse, rebound, aliased, accessor-based, and proxy-backed arrays.
  * Added coverage ensuring exceptions and other observable behaviors remain unchanged.

* **Documentation**
  * Documented the packed-loop optimization and its supported scenarios.

### DIFF
--- a/changelog.d/8786-closure-captured-packed-loops.md
+++ b/changelog.d/8786-closure-captured-packed-loops.md
@@ -1,0 +1,1 @@
+Optimize immutable closure-captured packed Array and Array-subclass loops, including nested arrays derived from guarded indexed reads, while retaining generic side exits for rebinding, layout changes, and moving GC.
--- a/crates/perry-codegen/src/expr/mod.rs
+++ b/crates/perry-codegen/src/expr/mod.rs
@@ -1611,6 +1611,22 @@ pub(crate) struct StablePackedLoopFact {
     pub array_local_id: u32,
     pub side_exit_label: String,
     pub descriptor: String,
+    /// Boxed bound passed to the runtime guard (`-1` requests live length).
+    pub bound: String,
+    /// Live-length versions must observe growth as well as shrink. The
+    /// iteration guard compares its refreshed bound with this admitted value
+    /// and side-exits when they differ.
+    pub admitted_bound: String,
+    pub live_length_bound: bool,
+    /// Captured receivers cannot keep a raw address across calls in the loop
+    /// body. They reload the closure slot and revalidate before the first
+    /// indexed effect of every iteration.
+    pub revalidate_each_iteration: bool,
+    /// A nested receiver derived from an outer guarded read may have pure
+    /// compiler temporaries before its first indexed use. Revalidate at that
+    /// use, after those temporaries, so none of their runtime loads can leave a
+    /// stale raw address.
+    pub revalidate_before_indexed_read: bool,
     pub live_receiver_handle: Option<String>,
     /// Admission scanned the complete indexed range and proved every value is
     /// an untagged IEEE Number. This is requested only when the indexed value
@@ -1619,6 +1635,10 @@ pub(crate) struct StablePackedLoopFact {
     /// Preheader-derived numeric storage bases. Admission proved the complete
     /// range is raw f64 and the call-free clone keeps these addresses stable.
     pub numeric_access: Option<StablePackedNumericAccess>,
+    /// Immutable locals initialized from this loop's guarded direct indexed
+    /// read. They may seed a nested candidate only while this fast-loop fact
+    /// is active.
+    pub derived_locals: std::collections::HashSet<u32>,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
--- a/crates/perry-codegen/src/runtime_decls/strings.rs
+++ b/crates/perry-codegen/src/runtime_decls/strings.rs
@@ -703,6 +703,14 @@ pub fn declare_phase_b_strings(module: &mut LlModule) {
         I32,
         &[DOUBLE, DOUBLE, I32, PTR],
     );
+    // #8773: same complete packed-loop admission, returning the validated
+    // live receiver address. Captured bindings may hold an array-growth
+    // forwarding stub and cannot be rewritten like a compiler-private local.
+    module.declare_function(
+        "js_packed_arraylike_loop_guard_live",
+        I64,
+        &[DOUBLE, DOUBLE, I32, PTR],
+    );
     // Issue #957: tag-aware dynamic index write. Used by `Expr::IndexUpdate`
     // codegen to write back the incremented value without rebuilding the
     // IndexSet dispatch tree. Routes to `js_array_set_index_or_string` for
--- a/crates/perry-codegen/src/stmt/let_stmt.rs
+++ b/crates/perry-codegen/src/stmt/let_stmt.rs
@@ -120,6 +120,7 @@ pub(crate) fn lower_let(
         }
     }
     if let Some(init_expr) = init {
+        super::stable_packed_loop::record_derived_local(ctx, id, init_expr, mutable);
         crate::expr::record_local_value_alias_for_write(ctx, id, init_expr);
         record_array_length_snapshot(ctx, id, init_expr);
         ctx.guarded_discriminant_aliases.remove(&id);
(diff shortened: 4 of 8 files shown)


proggeramlug:
Landed on `main` via the #8803 batch.

PR #8792, opened 2026-08-25
perf(runtime): index captured closure reuse

proggeramlug:
## Summary

- fingerprint exact captured-closure tuples before the bounded LRU scan
- add a direct-mapped hint that always verifies exact capture bits and repairs itself after collisions
- replace per-hit Vec removal/insertion with monotonic LRU timestamps
- recompute fingerprints and clear disposable hints after copying-GC rewrites

## Performance

Apple M1 Mac mini, unchanged codehz/ecs read-only row:

- fingerprint prefilter: +1.792% median paired at repeat 8,192
- direct hint over the fingerprint candidate: +0.834% median paired at repeat 8,192
- open-addressed replacement was tested and rejected at -0.290%; it is not part of this PR

These A/Bs were measured in the cumulative profiled tree where closure lookup was visible; the mechanism itself is standalone and this PR is based directly on current `main`.

## Validation

- exact collision fallback and full-cache LRU tests
- copying-GC cache rewrite test
- full serial release runtime suite in the cumulative final tree: 2,681 passed, 4 ignored
- focused forced-evacuation cache tests passed
- unchanged ECS semantic/checksum oracles passed in every measured process

This PR was opened but not merged by the performance investigation.

## Summary by CodeRabbit

- **Performance**
  - Improved reuse of captured closures with faster cache lookups.
  - Added bounded cache management to retain frequently reused closures while removing older entries.

- **Reliability**
  - Cache lookups now verify exact captured values, preventing incorrect reuse during collisions.
  - Improved cache correctness after garbage collection and pointer updates.

- **Tests**
  - Added coverage for lookup collisions and least-recently-used cache eviction.

- **Documentation**
  - Added a changelog entry describing the captured-closure cache improvements.

### DIFF
--- a/changelog.d/8792-captured-closure-cache-hints.md
+++ b/changelog.d/8792-captured-closure-cache-hints.md
@@ -1,0 +1,1 @@
+Captured-closure singleton reuse now fingerprints each exact capture tuple and maintains a direct-mapped, collision-safe hint into its bounded LRU. Hits no longer move vector entries, hash or hint collisions still require bit-exact capture equality, and copying GC recomputes fingerprints while discarding stale hints after rewriting pointer-bearing captures.
--- a/crates/perry-runtime/src/closure/alloc.rs
+++ b/crates/perry-runtime/src/closure/alloc.rs
@@ -12,20 +12,7 @@ crate::perry_thread_local! {
     static SINGLETON_CLOSURES: RefCell<crate::fast_hash::PtrHashMap<usize, *mut ClosureHeader>> =
         RefCell::new(crate::fast_hash::new_ptr_hash_map());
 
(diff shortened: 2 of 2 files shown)


proggeramlug:
Landed on `main` via #8796 (squash `726f6e96e`), with the rest of this batch.

Worth knowing why it looked red: your `pr-gate` was failing because **`main` itself was broken** — `cargo-test` was red on a Web Streams test (`byob_readers_and_byte_length_strategy`) until #8791 landed. Nothing in your PR caused it. Re-gated against the fixed baseline, it's clean.

Validated on the merged result: all 30 lint checkers, runtime 2677/0 at `RUST_TEST_THREADS=1`, codegen 1236/0, hir 335/0, all codegen integration suites clean.

Issue #8770, opened 2026-08-24
gc(runtime): natively-compiled cli.js agent-loop init hits non-deterministic GC/heap corruption (number-for-Map, undefined reads, SIGSEGV @ tagged ptr) — cc-boot layer 5


proggeramlug:
## Summary

With the startup chain fixed (#8726/#8738/#8739/#8763), the natively-compiled Claude Code cli.js 2.1.112 bundle now **boots and enters the agent loop** (`--help` works; `-p` progresses through hooks / skills attachment / autocompact per `--debug-file`). But `-p` then fails **non-deterministically** during agent-loop init (`num_turns: 0`), with a different symptom nearly every run — the signature of **GC / heap corruption under the app's allocation churn**:

- `TypeError: (number).get is not a function` / `(number).set is not a function` — a `Map`/object comes back as a *number* (its header/box clobbered)
- `TypeError: Cannot read properties of undefined (reading 'def')` — a live object read as undefined
- **SIGSEGV** — `EXC_BAD_ACCESS, KERN_INVALID_ADDRESS at 0x00000000ff00fff7` (a corrupted/tagged value dereferenced as a raw pointer)
- intermittent hangs

## Not fixed by GC pacing

12-run probe: **default, `PERRY_GC_MOVING_LOOP_POLLS=0`, and `PERRY_GEN_GC=0` all hang/err**; both-off gives the `.set` error + SIGSEGV. No configuration stabilizes it, so it's not a single pacing knob.

## Node parity

`node cli_2.1.112.js --help` and the agent loop run cleanly under Node — this is perry-runtime GC/memory correctness under a large real workload, i.e. the memory-parity frontier (cf. #7019 scavenge parity and the live-sweep fixes), not an app bug.

## Next step to pin it

The shipped binary is stripped. Build a `PERRY_DEBUG_SYMBOLS=1` cc (invalidates the codegen cache → full ~50 min recodegen + link; needs ~15 GB transient), reproduce the SIGSEGV, and symbolicate the `.ips` faulting frames to locate the clobbered allocation / missed root. The reduced-repro will likely be a churn pattern in the agent-loop init (skill/hook registries, token counting) rather than a one-liner.

Found while verifying the natively-compiled Claude Code boots end-to-end after the startup-chain fixes.

proggeramlug:
## Root cause pinned by the mark verifier

[...]

PR #8801, opened 2026-08-25
runtime: root async-from-sync iterator objects across GC safepoints

proggeramlug:
## Summary
The `%AsyncFromSyncIteratorPrototype%` implementation in `array/iterator.rs` created nursery GC objects (the wrapper object, the `outer` promise, the two reaction closures, the `{value, done}` result, and the captured sync iterator) and held them as **raw pointers across subsequent allocations and JS calls**. Under the default-on moving young-gen scavenge (#7019) those raw pointers are invalidated — the object is evacuated (leaving a stale pointer) or, when it becomes unreachable, swept — so a later use dereferenced a stale/poison receiver (surfacing as `shape_is_url_search_params` / `dispatch_raw_pointer` / `js_is_truthy` SIGSEGVs on a poison receiver, or `(number).get is not a function`).

Fix: root every live young value in a `RuntimeHandleScope` and re-read it through the handle after each allocation / JS call — the same convention already used elsewhere in the tree (`build_iter_result_ordered`, `iter_object.rs` ~L756). Property-name keys switch to the long-lived string allocator (immortal, never moves).

Functions covered: `async_from_sync_wrap_iterator`, `install_async_from_sync_next`, `install_async_from_sync_method`, `async_from_sync_next`, `async_from_sync_return`, `async_from_sync_throw`, `async_from_sync_call`, `async_from_sync_call_raw`, `async_from_sync_continue`, `async_from_sync_fulfilled`, `async_from_sync_rejected_value`, `async_from_sync_iter_result`.

## Context / honesty
Found during the `#8770` GC-corruption campaign (getting the compiled Claude Code CLI to run natively). These are genuine latent rooting bugs on the `for await (… of syncIterable)` path, exposed once the moving scavenge became the default. They were **not** the dominant crash in that campaign — that residual is a separate codegen precise-root gap — so this is a **correctness fix, not a validated crash fix**. The convention violations are real and the code now matches the file's own rooting discipline.

## Testing
`cargo check -p perry-runtime` clean on top of current `main`. Pure rooting change; spec semantics unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved async iterator handling to remain reliable during memory management activity.
  * Prevented intermittent failures caused by values moving while asynchronous iteration is in progress.
  * Preserved existing iterator behavior while improving runtime stability.

### DIFF
--- a/crates/perry-runtime/src/array/iterator.rs
+++ b/crates/perry-runtime/src/array/iterator.rs
@@ -268,12 +268,26 @@ fn undefined_value() -> f64 {
 }
 
 fn async_from_sync_iter_result(value: f64, done: bool) -> f64 {
+    // The result value and the freshly-allocated object are live young objects
+    // held across sibling allocations (the object alloc, and each
+    // `set_field_by_name`, which can grow the shape). The default-on moving
+    // scavenge evacuates young survivors, so cache them in a handle scope and
+    // re-read through the (GC-updated) handles instead of stale raw copies.
+    // Property-name keys use the long-lived allocator so they never move and
+    // need no rooting.
+    let scope = crate::gc::RuntimeHandleScope::new();
+    let value_h = scope.root_nanbox_f64(value);
     let obj = crate::object::js_object_alloc(0, 2);
-    let value_key = crate::string::js_string_from_bytes(b"value".as_ptr(), 5);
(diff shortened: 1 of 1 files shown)


proggeramlug:
Landed on `main` via the #8805 batch.

PR #8802, opened 2026-08-25
runtime: allocate native async-resolution promises in malloc space

proggeramlug:
## Summary
Native async resolutions (`fetch`, DB drivers, WebSocket, argon2/bcrypt, containers, etc. — the `spawn_for_promise`-style bindings) created their `Promise` via the **nursery arena** (`js_promise_new`), pinned it, and handed the raw pointer to a tokio worker thread. Until the worker completes and queues the resolution into `PENDING_RESOLUTIONS`, the promise is reachable **only** via the worker-thread capture — invisible to the main-thread copying minor.

A copying-minor from-space flip destroys a nursery resident **regardless of its `GC_FLAG_PINNED`** (the flip resets eden/survivor blocks wholesale; only root-reachable pins force the pinned-young fallback — see the `js_promise_new_cross_thread` doc comment in `promise/then.rs`). So the promise is reclaimed, and when the worker's resolution later reaches `js_stdlib_process_pending`, it roots/unpins/resolves through the **reclaimed** promise header → `EXC_BAD_ACCESS` in `js_stdlib_process_pending` (clearing `GC_FLAG_PINNED` on a from-space header), or `(number).get is not a function` downstream.

Fix: allocate these promises via `js_promise_new_cross_thread` (malloc space — non-moving, and both sweep paths honor `GC_FLAG_PINNED`), the design-intended allocator for exactly this cross-thread hand-off. Re-export the symbol from `perry-runtime` and switch every native-binding caller.

## How it was found / confirmed
The `#8770` GC-corruption campaign (getting the compiled Claude Code CLI to run natively). On a symbolicated build under `PERRY_GC_PROTECT_FROMSPACE=1`, the SIGSEGV was 12/12 in `js_stdlib_process_pending` on a **protected from-space** promise header; after this change that fault is gone (the nursery-pin path for native resolutions never fires — all such promises are malloc).

## Files
`perry-runtime/src/lib.rs` (re-export `js_promise_new_cross_thread`) + `perry-stdlib/src/common/async_bridge.rs` (the helper) + 21 native-binding callers (`fetch/mod.rs`, `ws`, `ioredis`, `bcrypt`, `argon2`, `axios`, `mongodb`, `nodemailer`, `sharp`, `pg/*`, `mysql2/*`, `net/mod`, `container/*`, `worker_threads/async_shim`).

## Testing
`cargo check -p perry-runtime -p perry-stdlib` clean on current `main`. Malloc promises are valid everywhere they're used (non-moving; sweep honors the pin), so the switch is safe at every call site.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability for asynchronous operations across networking, databases, containers, file processing, messaging, and security APIs.
  * Promises created by background tasks are now handled more safely, reducing the risk of lost or invalid results.
  * WebSocket, fetch, HTTP, and worker-based operations receive more consistent cross-thread behavior.
  * MySQL query timeout errors now provide clearer guidance when the server may be unavailable.

### DIFF
--- a/crates/perry-runtime/src/lib.rs
+++ b/crates/perry-runtime/src/lib.rs
@@ -307,7 +307,8 @@ pub use object::{
 };
 pub use promise::{js_is_promise, js_promise_run_microtasks, js_promise_state, js_promise_value};
 pub use promise::{
-    js_promise_mark_internally_handled, js_promise_new, js_promise_reject, js_promise_rejected,
+    js_promise_mark_internally_handled, js_promise_new, js_promise_new_cross_thread,
+    js_promise_reject, js_promise_rejected,
     js_promise_resolve, js_promise_resolved,
 };
 pub use string::js_string_from_bytes;
--- a/crates/perry-stdlib/src/argon2.rs
+++ b/crates/perry-stdlib/src/argon2.rs
@@ -9,14 +9,14 @@ use argon2::{
     password_hash::{rand_core::OsRng, PasswordHash, PasswordHasher, PasswordVerifier, SaltString},
     Argon2,
 };
-use perry_runtime::{js_promise_new, js_string_from_bytes, Promise, StringHeader};
+use perry_runtime::{js_promise_new_cross_thread, js_string_from_bytes, Promise, StringHeader};
 
 /// argon2.hash(password) -> Promise<string>
 ///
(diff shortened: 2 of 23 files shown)


proggeramlug:
Landed on `main` via the #8805 batch.

PR #8817, opened 2026-08-25
test(compile): cover builtin imports in compiled packages

### LINKED ISSUE #8749: compilePackages: builtin import resolves to undefined inside a compiled dependency (@hono/node-server → createServer undefined)
**Env:** perry 0.5.1512 built from `main@6173ff9f6`, macOS arm64. MB24 Phase-0 spike.

**Symptom:** with `@hono/node-server@1.19.17` + `hono@4.13.4` in `perry.compilePackages`, `serve({ fetch: app.fetch, port })` crashes:

```
Cannot read properties of undefined (reading 'listen')
```

@hono/node-server `dist/index.mjs`:

```js
import { createServer as createServerHTTP } from "http";   // line 2
…
const createServer = options.createServer || createServerHTTP;  // 645
const server = createServer(options.serverOptions || {}, requestListener); // → undefined(...)
```

**Isolation:** at application level both import shapes are fine —

```ts
import { createServer as a } from "http";       // typeof a === "function"
import { createServer as b } from "node:http";  // typeof b === "function"
```

so the builtin binding is lost specifically when the import lives inside a `compilePackages`-compiled dependency (or in the `options.createServer || createServerHTTP` module-scope reference path).

**Expected:** builtin imports inside compiled deps bind like app-level ones; hono serving over node:http works (was #589, closed — this appears to be a different, compilePackages-specific gap).

Full test: mobilitybroker24/monorepo `spike/tests/09-hono.ts`.

proggeramlug:
## Summary

The report was captured against `main@6173ff9f6`. On current `main`, the focused compiled-package reproducer and the exact published `@hono/node-server` adapter path both succeed. This PR locks that behavior down with offline and real-package regression coverage so module-scope Node builtin imports cannot silently regress to `undefined` inside compiled dependencies.

## Changes

- Add an offline `compilePackages` integration test that exercises `options.createServer || createServerHTTP`, calls the selected factories, and covers both `http` and `node:http` named imports.
- Add a tier-3 release fixture pinned to `@hono/node-server@1.19.17` and `hono@4.13.4` that listens, self-fetches, verifies the response, and closes cleanly from the package's listening callback.
- Add the required changelog fragment; no workspace version bump or release-metadata edits.

## Related issue

Closes #8749

## Test plan

- [ ] `cargo build --release` clean (not run; focused integration target built successfully)
- [ ] `cargo test --workspace --exclude perry-ui-ios --exclude perry-ui-tvos --exclude perry-ui-watchos --exclude perry-ui-gtk4 --exclude perry-ui-android --exclude perry-ui-windows` passes (not run; focused integration test below passes)
- [x] `cargo test -p perry --test issue_8749_compiled_package_builtin_import --no-fail-fast`
- [x] `rustfmt --edition 2021 --check crates/perry/tests/issue_8749_compiled_package_builtin_import.rs`
- [x] `py -3 scripts/check_test_registration.py`
- [x] Node oracle run for the pinned release fixture produced `status=200` and `body=ok`
- [x] Committed pinned release fixture compiled and ran under Perry with the same output (`PERRY_RS4GC=0` on Windows to avoid the separate known #7354 WinEH limitation)
- [x] `bash -n tests/release/packages/hono-node-server/fixture.sh`
- [ ] (if CLI / stdlib / runtime API changed) Updated `docs/src/` (not applicable)

`cargo fmt --all -- --check` could not complete in this Windows worktree because rustfmt hit the existing maximum-path-length limit; the changed Rust test was checked directly and passed. The broader `pre-tag-check.sh --quick` also expects `python3` on Git Bash's PATH; the relevant registration check passed via the installed Windows Python launcher.

## Screenshots / output

```text
status=200
body=ok
```

## Checklist

### DIFF
--- a/changelog.d/8817-compiled-package-builtin-import.md
+++ b/changelog.d/8817-compiled-package-builtin-import.md
@@ -1,0 +1,6 @@
+Added regression coverage for Node builtin named imports used from natively
+compiled dependencies. The exact `@hono/node-server` fallback from
+`options.createServer` to its module-scope `http.createServer` import now has an
+offline compiler fixture and a real-package listen/fetch/close release smoke,
+covering both `http` and `node:http` spellings without relying on app-level
+imports.
--- a/crates/perry/tests/issue_8749_compiled_package_builtin_import.rs
+++ b/crates/perry/tests/issue_8749_compiled_package_builtin_import.rs
@@ -1,0 +1,114 @@
+//! Regression test for #8749: a Node builtin named import used from a
+//! `compilePackages` dependency must retain its runtime binding.
+//!
+//! `@hono/node-server` imports `createServer` from `http`, selects it through a
+//! module-scope fallback (`options.createServer || createServerHTTP`), and calls
+//! the selected function later from `serve()`. App-level imports already
+//! worked; the binding was lost specifically while compiling the dependency.
+
+use std::path::PathBuf;
+use std::process::Command;
+
+fn perry_bin() -> PathBuf {
+    PathBuf::from(env!("CARGO_BIN_EXE_perry"))
+}
+
+#[test]
+fn builtin_named_import_survives_module_scope_fallback_in_compiled_package() {
+    let dir = tempfile::tempdir().expect("tempdir");
+    let root = dir.path();
+
+    std::fs::write(
+        root.join("package.json"),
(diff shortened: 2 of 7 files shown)


proggeramlug:

PR #8818, opened 2026-08-25
fix(sharp): support create input descriptors

### LINKED ISSUE #8748: ext-sharp: sharp({ create: {...} }) throws 'Invalid sharp handle'
**Env:** perry 0.5.1512 built from `main@6173ff9f6`, macOS arm64. MB24 Phase-0 spike.

**Repro:**

```ts
import sharp from "sharp";
const buf = await sharp({ create: { width: 4, height: 4, channels: 3,
  background: { r: 1, g: 2, b: 3 } } }).png().toBuffer();
```

**Actual:** throws the plain string `"Invalid sharp handle"` (typeof `"string"`). `typeof sharp` is `"object"` (Node: `"function"`, callable either way here).

**Expected:** sharp's `create` input (blank canvas) — the standard building block for watermark overlays and placeholder images.

**Impact/context:** buffer- and file-based inputs untested beyond this (the create path is the first op in our pipeline test). ext-sharp being viable end-to-end (create + resize + composite + metadata/EXIF strip) decides whether MB24 uses sharp or shells out to the vips CLI (which works fine under Perry today, verified on Linux CI). Full test: mobilitybroker24/monorepo `spike/tests/11-image-sharp.ts`.

proggeramlug:
## Summary

Adds support for Sharp's object-form `create` input so `sharp({ create: ... }).png().toBuffer()` creates a real solid RGB or RGBA image instead of returning an invalid handle.

## Changes

- Parse and validate `create.width`, `height`, `channels`, and `background` using Sharp-compatible bounds and channel rules.
- Preserve RGB/RGBA pixel data and alpha while making nested option reads safe across GC activity.
- Add native unit coverage and a compiled TypeScript regression for both RGB and RGBA PNG pipelines.
- Document the supported `sharp({ create: ... })` constructor form.
- Refresh Sharp's test-only async FFI shims to match the current host ABI.

## Related issue

Fixes #8748

## Test plan

- [x] `cargo build --release -p perry-ext-sharp`
- [x] `cargo test -p perry-ext-sharp --lib`
- [x] `cargo clippy -p perry-ext-sharp --lib --no-deps -- -D warnings`
- [x] `cargo test -p perry --test issue_8748_sharp_create`
- [x] `python scripts/check_test_registration.py`
- [x] `cargo run -p perry-doc-tests -- --lint docs/src`
- [x] `cargo fmt -p perry-ext-sharp -p perry-codegen -p perry -- --check`
- [x] Added native and compiled TypeScript regression coverage
- [x] Updated `docs/src/`
- [ ] Full workspace test suite (targeted affected-crate and runtime-path checks used instead)

## Screenshots / output

Before: object-form `create` input produced `Invalid sharp handle`.

After: RGB and RGBA canvases encode to PNG and decode with the expected dimensions, channels, and alpha metadata.

## Checklist

### DIFF
--- a/changelog.d/8818-sharp-create.md
+++ b/changelog.d/8818-sharp-create.md
@@ -1,0 +1,1 @@
+fix(sharp): support object-form `create` inputs with solid RGB or RGBA backgrounds, so `sharp({ create: ... })` can encode images instead of failing with an invalid handle.
--- a/crates/perry-codegen/src/lower_call/native_table/media.rs
+++ b/crates/perry-codegen/src/lower_call/native_table/media.rs
@@ -4,9 +4,10 @@ pub(super) const MEDIA_ROWS: &[NativeModSig] = &[
     // ========== sharp ==========
     // Factory: sharp(path) → js_sharp_from_file. Instance methods take
     // Handle (i64), compatible with the has_receiver:true dispatch path.
-    // `sharp(input)` accepts a file-path string OR a Buffer/Uint8Array of
-    // encoded image bytes. Pass the raw NaN-boxed value (NA_JSV) so
-    // `js_sharp_from_input` can branch on the Buffer registry probe.
+    // `sharp(input)` accepts a file-path string, a Buffer/Uint8Array of encoded
+    // image bytes, or a `{ create: { ... } }` descriptor. Pass the raw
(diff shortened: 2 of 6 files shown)


proggeramlug:

PR #8819, opened 2026-08-25
feat(qs): add native Stripe-compatible shim

### LINKED ISSUE #8751: Native qs shim to unlock the stripe SDK (get-intrinsic chain is AOT-hostile)
**Env:** verified on release v0.5.1220 and on 0.5.1512 (`main@6173ff9f6`), macOS arm64 + Linux x86_64. MB24 Phase-0 spike.

The `stripe` npm SDK compiles cleanly via `perry.compilePackages` **except** its runtime dependency `qs`. Chasing `qs` through compilePackages pulls the legacy ES-shim family (`side-channel`, `call-bind`, `get-intrinsic`, 19 packages total); it then compiles but dies at first use:

```
SyntaxError: intrinsic %% does not exist!
```

`get-intrinsic` introspects engine intrinsics from a runtime-built table — structurally incompatible with AOT, not fixable by patching MB24 or by more compilePackages entries.

**Proposal:** ship a native `qs` ext shim (the package is a smallish querystring encoder/decoder; the stripe SDK only needs `qs.stringify` with nested objects/arrays for form-encoded bodies). That unlocks the entire stripe SDK — likely the most-requested payments path for server-side Perry apps. Alternative: an SDK-level substitution hook, but the shim helps every `qs` consumer.

Everything else stripe needs looked healthy: fetch-based `httpClient` compiles, and `stripe.webhooks.constructEvent` (node:crypto HMAC + timing-safe compare) is in our suite pending this blocker: mobilitybroker24/monorepo `spike/tests/08-stripe.ts`.

proggeramlug:
## Summary

- add a bundled, provenance-pinned `qs@6.15.3` native binding so Stripe no longer compiles the AOT-hostile `get-intrinsic` / ES-shims chain
- implement GC-safe nested `stringify` behavior, including Stripe's indexed arrays and `serializeDate` callback, plus common qs encoding, filtering, sorting, null, charset, and array-format options
- implement safe nested `parse` behavior with duplicate modes, depth/parameter/array limits, charset sentinels, sparse arrays, dot notation, and prototype-pollution guards
- wire the binding through native dispatch, API manifests, provider routing, well-known provenance, workspace governance, and release notes
- add an end-to-end regression with hostile installed `qs` and `get-intrinsic` sentinels, covering CommonJS Stripe-style usage plus named, namespace, and default imports

The binding is deliberately marked `partial`: regex delimiters and the complete custom decoder/extension-hook contracts are not claimed as full upstream compatibility.

No package version was bumped.

Closes #8751

## Validation

- `cargo test -p perry-ext-qs --lib` (11 passed)
- `cargo test -p perry-codegen --test manifest_consistency` (5 passed)
- `cargo test -p perry-codegen ext_registry --lib` (13 passed)
- `cargo test -p perry --test issue_8751_qs_native_shim -- --nocapture` (passed after rebasing onto latest main)
- `cargo check -p perry-codegen -p perry-api-manifest -p perry`
- `cargo clippy -p perry-ext-qs --all-targets --no-deps -- -D warnings`
- `python scripts/workspace_architecture.py --check`
- `python scripts/binding_governance.py --check`
- `node scripts/binding_pins.mjs --check`
- focused `cargo fmt --check` and `git diff --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6165,6 +6165,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "perry-ext-qs"
+version = "0.5.1519"
+dependencies = [
+ "perry-ffi",
+ "perry-runtime",
+ "serde_json",
+]
+
+[[package]]
 name = "perry-ext-ratelimit"
 version = "0.5.1519"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ members = [
     "crates/perry-ext-events",
     "crates/perry-ext-decimal",
     "crates/perry-ext-dayjs",
+    "crates/perry-ext-qs",
     "crates/perry-ext-moment",
     "crates/perry-ext-cheerio",
     "crates/perry-ext-sharp",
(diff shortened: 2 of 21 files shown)


proggeramlug:

PR #8820, opened 2026-08-25
perf(codegen): specialize call-returned array stores

proggeramlug:
## What changed

- recognize the structurally identical `PutValueSet.target` and `.receiver` HIR trees as the single source evaluation they represent, including a call expression such as `this.getData()[index] = value`
- retain the statically known Array type and lower that store through the existing typed-feedback array-index helper instead of the generic Proxy-compatible dynamic PutValue IC
- keep genuinely different receiver trees fail-closed on the explicit-receiver path
- complete the shared strict array-index guard for own non-writable/getter-only descriptors, setter accessors, read-only length extension, and holes on non-extensible arrays
- add focused generated-IR tests plus an executable semantics regression covering evaluation count, fractional keys, assignment result, Proxy traps, descriptors, non-extensible holes, and read-only length

This is a general lowering/runtime fix: it contains no ECS names, benchmark labels, workload thresholds, or command-buffer special cases. `IndexSet::lower` still evaluates the assignment base once; the structural check only recovers the identity already encoded by HIR.

## Verification

- `git diff --check`
- `python3 scripts/gc_store_site_inventory.py --gate` (1,579 files; all 97 runtime barrier claims source-verified)
- focused codegen IR tests: 2 passed
- executable compiler/runtime semantics test: 1 passed
- full serial debug runtime suite on the rebased commit: 2,689 passed, 0 failed, 4 ignored
- full serial release runtime suite on the Mac mini from SHA-matched rebased sources: 2,689 passed, 0 failed, 4 ignored

One initial local executable-test compile reported successful module codegen but lost its temporary `main_ts.o` before the clang link. The exact isolated retry passed in 13.3 seconds; the semantic assertions did not fail.

## Mac mini benchmark evidence

Five interleaved Perry/Perry A/B pairs used the same precompiled isolated ECS 15k-command source, with this change stacked on #8807 and #8813. Each process used 2 warmups and 6 measured rounds; all 10 process-level semantic oracles passed.

- control median: 58.392 ms
- candidate median: 50.866 ms
- median paired improvement: 12.889%
- paired range: 12.672% to 13.318%
- candidate wins: 5/5

The candidate Archetype IR contains three calls to `js_typed_feedback_array_set_index_or_string` and no calls to `js_put_value_set_dyn_ic`. Binary SHA-256 values were pinned for both arms. This is an admitted mechanism gain, not a Node-parity claim; the wider ECS plan continues to track the remaining gap separately.

### DIFF
--- a/changelog.d/8820-call-return-array-stores.md
+++ b/changelog.d/8820-call-return-array-stores.md
@@ -1,0 +1,1 @@
+Array index assignments whose base is a call expression now preserve the call's statically known Array type and use the typed array-store path while evaluating the base exactly once. Strict writes through that path also honor non-writable and accessor descriptors, read-only length, and non-extensible holes.
--- a/crates/perry-codegen/src/expr/call_return_array_index_tests.rs
+++ b/crates/perry-codegen/src/expr/call_return_array_index_tests.rs
@@ -1,0 +1,175 @@
+use crate::{compile_module, CompileOptions};
+use perry_hir::types::Type;
+use perry_hir::{Class, Expr, Function, Module, Param, Stmt};
+
+fn param(id: u32, name: &str, ty: Type) -> Param {
+    Param {
+        id,
+        name: name.to_string(),
+        ty,
+        default: None,
+        decorators: Vec::new(),
+        is_rest: false,
+        arguments_object: None,
+    }
+}
+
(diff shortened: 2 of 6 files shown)


proggeramlug:

PR #8826, opened 2026-08-25
perf(codegen): specialize branded ReadonlySet.has

proggeramlug:
## Summary

- treat a declared `ReadonlySet<T>` as a guarded dispatch candidate, never as proof of native Set layout
- add an exact `GC_TYPE_SET` fast helper with ordinary method dispatch on every brand miss
- retain named class metadata from whole-declaration type-only imports only when the producer is already value-reachable, without adding runtime bindings or init edges

This is a general collection-dispatch fix, not an ECS-specific rewrite. Structural objects, proxies, nullable receivers, and Set subclass overrides keep the generic semantics path.

## Mac mini ECS evidence

Exact codehz/ecs revision: `58d729682eeb88d7361796de0420f3d673e27479`.

Against the exact #8820 control binary, the command-system benchmark median moved from 50.72794 ms to 37.13701 ms: 26.691189% paired improvement, 5/5 paired wins, with all 10/10 process-level semantic oracles passing. The real command-executor IR now contains four branded calls instead of zero generic-only call sites.

For qualification only, stacking #8813 + #8820 + this change passes all 7/7 comprehensive ECS cases; the relation-flip case averages 178.60 ms. The relation case was a known pre-existing miss on #8820 alone and was fixed by #8813, not by this PR. Both predecessor changes are now on main through #8822.

## Validation

- `cargo fmt --all`
- `cargo test -p perry-codegen expr::readonly_collection_tests -- --nocapture` (3 passed)
- `cargo test -p perry --test readonly_set_branded_dispatch -- --nocapture` (1 passed)
- `cargo test -p perry --test source_graph_export_regressions type_only_class_import_retains_guarded_readonly_set_field_metadata -- --nocapture` (1 passed)
- `cargo test -p perry-runtime --lib -- --test-threads=1` (2689 passed, 4 ignored)
- `python3 scripts/gc_store_site_inventory.py`
- exact Mac mini release build and real ECS compile/run

Rebased onto current main after #8813 and #8820 landed through #8822. The three focused compiler/executable regressions were rerun and passed on the rebased head.

## Summary by CodeRabbit

* **Performance**
  * Improved `ReadonlySet.has` lookups for native sets while preserving custom `has` method behavior.
  * Maintained correct handling for nullable receivers, subclasses, and structurally compatible objects.

* **Bug Fixes**
  * Type-only class imports now retain metadata needed for reliable `ReadonlySet` field operations without adding runtime module behavior.

* **Tests**
  * Added coverage for local, imported, inherited, nullable, and custom `ReadonlySet` scenarios.
  * Added regression checks confirming correct runtime results and dispatch behavior.

### DIFF
--- a/changelog.d/8826-readonly-set-has.md
+++ b/changelog.d/8826-readonly-set-has.md
@@ -1,0 +1,4 @@
+Speed up calls to `ReadonlySet.has` with an exact runtime Set-brand check and
+ordinary method dispatch fallback. Type-only class imports now retain the
+field metadata needed for this guarded optimization without creating runtime
+module bindings or initialization edges.
--- a/crates/perry-codegen/src/expr/mod.rs
+++ b/crates/perry-codegen/src/expr/mod.rs
@@ -169,6 +169,8 @@ mod call_spread_short;
 mod call_spread_short_tests;
 #[cfg(test)]
 mod issue7628_rooting_tests;
+#[cfg(test)]
+mod readonly_collection_tests;
 pub(crate) mod shadow_slot;
 #[cfg(test)]
 mod slice7_rooting_tests;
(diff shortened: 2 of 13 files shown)


proggeramlug:
@coderabbitai review

proggeramlug:

PR #8830, opened 2026-08-25
perf(map): specialize declared Map.get dispatch

proggeramlug:
## Summary

- retain declared Map and ReadonlyMap candidates through nested class, interface, and object fields
- brand-check the live receiver before entering the existing native Map lookup
- preserve ordinary get method dispatch for structural values, subclasses, proxies, primitives, and nullish receivers

## Why

A steady-state profile of the unchanged codehz/ecs 15k-command workload placed 1,015 samples in generic native method dispatch for this.ctx.entityToArchetype.get(entityId). The nested interface-typed field retained its Map declaration, but not the stronger native-layout proof used by the existing direct lowering.

This change uses the declaration only to select a guarded runtime operation. A receiver whose tracked GC type is exactly the built-in Map kind enters the existing generic Map key lookup directly. Every brand miss roots both operands and uses the ordinary method dispatcher. It does not add numeric-key classification or change Map layout, lookup semantics, or the existing proven-native path.

## Correctness

- full codegen library suite: 1,256 passed, 1 ignored, 0 failed
- full serial runtime library suite: 2,690 passed, 4 ignored, 0 failed
- focused generated-IR test for the nested interface-field case
- compiled TypeScript semantics test for native Map, ReadonlyMap typing, structural receivers, subclass overrides, and nullish behavior
- existing compiled ReadonlySet semantics regression
- rustfmt and diff checks

The upstream ECS functional sweep passed 461/463 cases. Both failures reproduced identically on the exact parent binary. The candidate introduced no new performance-threshold failure: serialization and sparse-wildcard fail their existing upstream ceilings on both candidate and exact parent.

## Performance

M1 Mac mini, unchanged codehz/ecs v0.13.0 at 58d729682eeb88d7361796de0420f3d673e27479, 15k-command comprehensive row, exact rebased #8826 control, two warmups and six measured rounds per process:

- control median: 36.561 ms
- candidate median: 31.833 ms
- median paired improvement: 12.981%
- paired range: 12.387% to 13.771%
- candidate wins: 11/11
- process semantic oracles: 22/22

A same-host Node 26.5.1 comparison measured 2.905 ms versus Perry at 31.828 ms, so this is an admitted incremental fix, not ECS command-path parity.

Stacked on #8826 for an isolated review diff.


proggeramlug:

PR #8832, opened 2026-08-25
runtime(gc): resolve x19 frame-base roots in the fast fp-chain walk (#8770)

proggeramlug:
## Summary

LLVM takes **x19** as a frame **base pointer** for a function with a *dynamic* stack allocation (a VLA or a spread-argument area). Its GC roots are stack slots addressed via x19, and x19 is established as `mov x19, sp` immediately after the fixed prologue and before the dynamic `sub sp, sp, xN`, with no realignment — so **x19 holds exactly the body SP the fp chain already reconstructs** (`fp - fp_to_sp_offset`).

Before this, **any** x19 root flipped the whole-image `chain_walkable` flag false (it required *every* root to be fp/sp), which globally disabled the fast x29-chain root walk and forced **every** GC onto the platform unwinder. Measured on the Claude Code `cli.js` bundle: just **63 of 72,812** functions use an x19 base (root-slot histogram: x19 257,518 / fp 16,107 / sp 10,634,043; zero x20–x28), yet they disabled the correct fast walker for *all* of them — after which the unwinder mis-resolved compiled-JS stack-slot roots and live young objects were swept (0xff-poison-receiver SIGSEGV / `(number).get is not a function`).

## Change

- `chain_walkable` now accepts `dwarf_reg` 19 in addition to fp(29)/sp(31); any *other* base register still disables the chain walk.
- The fast walker resolves an x19 root like an SP root (same reconstructed body SP), gated **per frame** by a new `x19_is_body_sp()` that re-decodes the prologue and confirms the `mov x19, sp`-after-fixed-frame shape (with no realignment / dynamic `sub sp,sp,xN` before it). A frame that does not match still fails closed to the platform unwinder — exactly as before.
- Disassembly of all 63 x19 functions in `cli.js` confirmed the uniform `mov x19, sp` (encoding `0x910003f3`, imm=0) shape with zero realignment, so x19 ≡ body SP for every one.

## Effect

`chain_walkable` is true again for the whole image; `cli.js` `--version`/`--help` go from crashing (global-gate victims of the unwinder) to **25/25 clean, 0 SEGV**. Updated the decode test accordingly (`an_x19_base_keeps_the_fast_walk_available` / `an_unsupported_base_register_disables_the_fast_walk`).

This is one layer of #8770; a separate residual (the async-from-sync runtime rooting path) is tracked independently.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved garbage-collection root tracking for supported AArch64 dynamic-allocation frames.
  * Added safer handling for stack frames that cannot be walked reliably by using the platform unwinder instead.
* **Tests**
  * Expanded coverage for supported frame-base registers and fallback behavior for unsupported configurations.

### DIFF
--- a/crates/perry-runtime/src/gc/roots/stack_maps.rs
+++ b/crates/perry-runtime/src/gc/roots/stack_maps.rs
@@ -444,6 +444,18 @@ unsafe fn rewrite_derived_slot(derived_addr: usize, base_addr: usize, old_base_w
 // the target and this runtime's `target_arch` can never disagree.
 const DWARF_REG_FP_AARCH64: u16 = 29;
 const DWARF_REG_SP_AARCH64: u16 = 31;
+// #8770: LLVM takes x19 as a frame base pointer for a function with a *dynamic*
+// stack allocation (a VLA or a spread-argument area). The base is captured as
+// `mov x19, sp` immediately after the fixed prologue and before the dynamic
+// `sub sp, sp, xN`, with no realignment — so x19 holds exactly the body SP the
+// fp chain reconstructs (`fp - fp_to_sp_offset`). The fast walker therefore
+// resolves an x19-based root like an SP-based one, once `x19_is_body_sp` has
+// confirmed that prologue shape for the owning function; a frame that does not
+// match (e.g. a realigning one) fails closed to the platform unwinder. Before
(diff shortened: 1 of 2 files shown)


proggeramlug:

Issue #8775, opened 2026-08-24
perf(codegen): specialize stable imported object-literal methods


proggeramlug:
## Summary

Perry can now specialize imported class prototype methods that capture `this` (#8693/#8737), but stable methods on imported object-literal adapters still traverse generic method lookup and closure dispatch on every call. Public ECS adapters use exactly this pattern, so the outer calls remain generic and prevent profitable inlining across the adapter boundary.

Add guarded direct specialization/inlining for stable methods on imported object literals (including default exports), with an explicit generic fallback when the receiver shape or own method identity changes.

## Self-contained reproduction

`adapter.js`:

```js
class Store {
  constructor() {
    this.entities = [];
  }

  create(id) {
    const entity = { id, value: 0 };
    this.entities.push(entity);
    return entity;
  }

  add(entity) {
    entity.value += 1;
  }

  destroy(entity) {
    const index = this.entities.indexOf(entity);
    if (index !== -1) this.entities.splice(index, 1);
    return entity.value;
  }
}

export default {
  store: null,
  setup() {
    this.store = new Store();
  },
  createEntity(id) {
    return this.store.create(id);
  },
  addComponent(entity) {
    this.store.add(entity);
  },
  destroyEntity(entity) {
    return this.store.destroy(entity);
  },
};
```

`main.js`:

```js
import adapter from "./adapter.js";

adapter.setup();
let checksum = 0;
const iterations = 200_000;
const start = performance.now();
for (let i = 0; i < iterations; i++) {
  const entity = adapter.createEntity(i);
  adapter.addComponent(entity);
  adapter.addComponent(entity);
  checksum += adapter.destroyEntity(entity);
}
console.log(JSON.stringify({
  elapsedMs: performance.now() - start,
  checksum,
  remaining: adapter.store.entities.length,
}));
```

Expected result: `checksum: 400000`, `remaining: 0`.

Build with the same release/runtime artifacts and retain `--trace llvm --opt-report=json --explain-lowering --no-cache` output. The public integration case is `ddmills/js-ecs-benchmarks` `perform-ecs@0.7.8`, suite `Destroy`: the suite and library contexts are imported object literals, and the timed method calls `ctx.createEntity`, `ctx.addPositionComponent`, `ctx.addVelocityComponent`, and `ctx.destroyEntity`.

## Current evidence

[...]

Issue #8774, opened 2026-08-24
perf(codegen): propagate Ptr<Shape> facts into guarded argument clones


proggeramlug:
## Summary

Perry's `Ptr<Shape>` proof stops when a known-shape object is passed as a non-receiver call argument. The callee therefore performs guarded field ICs and repeated GC-header/shape-descriptor reads even when a monomorphic call site proves the argument's class.

Generalize the existing clone-and-route mechanism for proven `this` receivers to selected object argument positions: a guarded call should route a known-shape argument into a callee clone that can directly access its declared fields while retaining the tagged ABI, shadow rooting, and generic fallback.

## Self-contained reproduction

```js
class Entity {
  constructor(id) {
    this.id = id;
    this.components = [];
  }
}

class Registry {
  add(entity, component) {
    entity.components.push(component);
  }

  hash(entity) {
    let value = entity.id;
    for (let i = 0; i < entity.components.length; i++) {
      value += entity.components[i];
    }
    return value;
  }

  clear(entity) {
    entity.components.length = 0;
  }
}

const registry = new Registry();
let checksum = 0;
const iterations = 200_000;
const start = performance.now();
for (let i = 0; i < iterations; i++) {
  const entity = new Entity(i);
  registry.add(entity, 1);
  registry.add(entity, 2);
  checksum += registry.hash(entity);
  registry.clear(entity);
}
console.log(JSON.stringify({
  elapsedMs: performance.now() - start,
  checksum,
}));
```

Expected checksum: `20000500000`.

Build with release artifacts and retain `--trace llvm --opt-report=json --explain-lowering --no-cache` output. The public integration case is `ddmills/js-ecs-benchmarks`, `perform-ecs@0.7.8`, suite `Destroy`.

## Current evidence

The fresh lowering report for the real workload on Perry `20a388974d7333ccaec9da9175fbcee54ad49902` diagnoses the exact loss in `perform-ecs/src/ECS.ts`, `closure#7`:

```text
local: entity
candidate class: Entity
outcome: denied
rule: rule 2 (containment)
reason: passed as a call argument. There is no mechanism yet by which a
shape fact at a call site becomes a fact about the callee's parameter.
```

[...]

Issue #8772, opened 2026-08-24
perf(codegen): direct-call stable methods with short packed spread tails


proggeramlug:
## Summary

Perry lowers a method call with a spread tail through `js_native_call_method_apply_by_id` / `js_native_call_method_apply` even when the callee identity is stable and the spread source is an exact short packed Array. In `perform-ecs`, every component reset uses:

```js
this._getComponentInstance(component.component).reset(entity, ...component.args);
```

The common `args` arrays are empty, but the hot path still performs generic apply dispatch and vtable/ABI trampoline work. Add a guarded direct-call specialization that unpacks a proven exact short packed spread source and invokes the selected method clone directly, with generic iterator/apply fallback for all proof misses.

## Self-contained reproduction

```js
class Position {
  constructor() {
    this.x = 0;
  }
  reset(entity, delta = 1) {
    this.x = entity.id + delta;
  }
}

class Velocity {
  constructor() {
    this.dx = 0;
  }
  reset(entity, delta = 2) {
    this.dx = entity.id + delta;
  }
}

const position = new Position();
const velocity = new Velocity();
const empty = [];
const one = [3];

function invoke(instance, entity, args) {
  instance.reset(entity, ...args);
}

let checksum = 0;
const iterations = 300_000;
const start = performance.now();
for (let i = 0; i < iterations; i++) {
  const entity = { id: i };
  invoke(position, entity, empty);
  invoke(velocity, entity, one);
  checksum += position.x + velocity.dx;
}
console.log(JSON.stringify({
  elapsedMs: performance.now() - start,
  checksum,
}));
```

Expected checksum: `90000900000`.

The public integration reproduction is `ddmills/js-ecs-benchmarks`, `perform-ecs@0.7.8`, suite `Destroy`. Build with release artifacts and retain `--trace llvm --opt-report=json --explain-lowering --no-cache` output.

## Current evidence

On Perry `20a388974d7333ccaec9da9175fbcee54ad49902`, the real benchmark measures 13.609234 us/op versus Node 1.025681 us/op (**13.268x slower**) on the quiet Apple M1 protocol.

[...]

PR #8833, opened 2026-08-25
perf: complete ECS benchmark specializations

proggeramlug:
## Summary

This is the hands-on integration follow-through for #8772, #8773, #8774, and #8775. It audits
the closed implementations against the real `noctjs/ecs-benchmark` and
`ddmills/js-ecs-benchmarks` paths, fills the activation/guard/fallback gaps, and adds executable
regressions for the semantics found during that audit.

### #8772 — short packed spread

- propagates concrete reverse-dependency candidates through whole-program compilation so the
  real `perform-ecs` calls select the direct path;
- keeps exact fallback and rooting behavior for generic iterables, proxies, Array subclasses,
  throwing iterators, and nullish spread tails;
- fixes guard-miss fallback to use the iterator protocol instead of assuming an Array layout;
- keeps mixed fixed/spread Math calls as spread calls rather than coercing their tail array as one
  scalar argument.

### #8773 — closure-captured packed loops

- versions both the captured outer arrays and arrays derived by nested `query[i]` reads;
- supports a spilled `length` slot and performs compact exact revalidation of admitted subclasses;
- side-exits individual reads without replaying an iteration;
- leaves generic indexed reads only in cold fallback blocks.

### #8774 — argument-shape clones

- propagates argument-shape candidates to the real `perform-ecs` destroy path;
- activates direct static-fresh clone lowering with guarded fallback;
- covers cross-module and negative layouts in compiler and end-to-end tests.

### #8775 — imported object methods

- propagates object-literal own-method metadata through the measured adapter boundary, including
  spill-only layouts;
- guards identity, class, exact shape, key, slot, and function before direct dispatch;
- rejects prototype methods, accessors, incompatible attributes/flags, and private brands;
- makes all four hot upstream calls direct.

## Correctness findings

The audit found two silent semantic failures that had allowed misleading benchmarks:

1. `Math.max(-1, ...nums)` was compacted as if `nums` were one scalar argument. Wolf's query masks
   then collapsed and its old run-success check timed effectively empty queries.
2. #8772's exotic spread fallback treated proxy and Array-subclass tails as native arrays and
   silently spread zero elements.

Both are fixed and pinned by Node/normal-Perry/moving-GC parity tests. The Wolf integration now
uses a strong final-state oracle with separate even/odd checksums and probe vectors.

## Controlled M1 results

Protocol: Apple M1 Mac mini, >=75% host idle for 60 seconds, three discarded process warmups per
arm, 11 paired processes in alternating order, `taskpolicy -t 0 -l 0`, and semantic validation in
every process.

| Scope | Result |
|---|---:|
| #8773 full Wolf | **1.730988x faster**, **42.2295% less time**, 11/11 wins |
| #8775 `perform-ecs` | **11.05% faster**, 11/11 wins |
| #8772 scoped `perform-ecs` | flat/slightly slower; activation and semantics pass, no speed claim |
| #8774 scoped `perform-ecs` | 0.24% slower (flat), 4/11 wins; no speed claim |

Final Wolf medians are 0.377113391 ms/op for the corrected control and 0.217860197 ms/op for this
patch. Node 26.5.1 is 0.004942129 ms/op, so Perry remains 44.0823x slower, improved from 76.3059x
for the corrected control.

The measured candidate was based on `e043aa294`. This branch was subsequently rebased without
conflicts onto current `main` at `ed5c971bb`; the numbers above identify the exact measured cohort
and are not represented as a fresh post-rebase run.

## Validation

- #8772 end-to-end: 5/5 suites in Node, normal Perry, and moving-GC Perry
- #8773 end-to-end: 4/4 suites
- #8774 end-to-end: 2/2 suites
- #8775 end-to-end: 2/2 suites
- focused HIR mixed-spread regression
- short-spread codegen: 3/3
- dense-spread runtime: 13/13
- `cargo fmt --check`
- `git diff --check`

The final symbolized Wolf sample has 407 samples in compact revalidation and zero in both the old
generic indexed-read helper and full live guard. Candidate RSS is 540,672 bytes lower; binary size
increases by 16,512 bytes.

## Summary by CodeRabbit

- **Performance**
  - Improved optimized calls for short spread arguments and dynamic object methods, including cross-module scenarios.
  - Added faster method dispatch with safer cache revalidation after object shape changes.
  - Improved packed array and array-subclass loop handling across repeated iterations.

- **Bug Fixes**
  - Corrected mixed fixed-argument and spread calls to `Math.min` and `Math.max`.
  - Preserved iterator behavior for proxies, array subclasses, `null`, and `undefined` spread values.
  - Prevented incorrect replay of effects during nested array iteration.

- **Tests & Documentation**
  - Added regression coverage and benchmark results for spread calls, method specialization, and loop behavior.


proggeramlug:
Held back from the #8835 batch — this PR fails two `perry-codegen` tests. Attributed by A/B, clean trees from the same base:

```
clean main            1253 passed; 0 failed
main + #8833 alone    1257 passed; 2 failed
```

**1. `collectors::proven_args::tests::pshape_argument_symbol_reachability`**
```
argument-shape clone symbol fragments found outside the direct-call allowlist:
  ["collectors/hir_facts.rs", "collectors/ptr_shape_report.rs"]
```

**2. `collectors::ptr_shape::opt_report_tests::escape_kinds_are_discriminated`**
```
a call-argument escape must be reported as such, not as a bare reference:
Some("passed to a call that does not preserve whole-lifetime containment. A guarded
`$pshape_args` route may still specialize field reads performed before the callee
publishes the argument, but the caller cannot retain an unguarded `Ptr<Shape>` fact
after that call.")
```

I did not update either test myself. The first might be a legitimate allowlist addition, but the second is not obviously a stale-test fix: the new wording makes a substantive claim — that a guarded route may still specialize field reads performed *before* the callee publishes the argument — and editing the assertion to match would rubber-stamp that claim rather than review it. That is your call.

Everything else in the PR looked sound on review, and the rest of the batch (#8825, #8826, #8830, #8832) landed cleanly in #8835, so this should rebase easily.

Also worth knowing, since it will affect your rebase: four gate failures in the combined tree came from this PR and I fixed them while investigating, but those fixes went in with the batch and are now on `main` — `addr_class_inventory` (the open-coded bare-address check in `typed_feedback/guards.rs` went 1 → 3 sites; I extracted a single `closure_ptr_from_bits` helper), `class_id_collisions` (a test-local `const CLASS_ID` in `typed_feedback/tests.rs` collided with an unrelated one; renamed to `OWN_METHOD_CACHE_CLASS_ID`), `shape_descriptor_census` (a benign `let header_skip =` refactor in `imported_object.rs`), and `check_file_size` (`ptr_shape.rs` hit 2103 lines; I split the this-flow analysis into `ptr_shape_this_flow.rs`). You may find some of those already done.

proggeramlug:
Landed on `main` via #8837, with a soundness fix rather than a test update.

Failure 2 was not a stale assertion. This PR widened the `$pshape_args` route in four mutually load-bearing ways and removed the runtime class+ShapeId guard in the same change. `PrefixContainedParamUse` proves a *temporal* property (reads before publication), but `guarded_argument_route_locals` is keyed by local id and therefore flow-insensitive — it is consulted at every route site for that local, including ones executing after the alias exists. A per-local map cannot express "before", so the PR's own wording ("cannot retain an unguarded `Ptr<Shape>` fact after that call") described a constraint its data structure could not enforce.

A probe fixture made the pre-fix compiler emit two *unguarded* direct calls into a clone doing fixed-offset field loads, the second on an object it had itself recorded as published to an invisible alias, in a module its own rule 5 says carries an unbounded shape barrier. On clean `main` that fixture routes nowhere.

The fix keeps guard elision only where it is justified (the caller already holds the broad `Ptr<Shape>` fact under full containment in a barrier-free module, so the guard is tautological), always requires `preserves_containment` for route admission, and deletes the `require_post_call_containment` knob whose only other mode was unsound. Your measured routes are unaffected — the perform-ecs/Wolf fixtures keep the elided guard. Full detail in #8837.

proggeramlug:
Landed via #8837.

PR #8839, opened 2026-08-25
perf: reuse guarded ECS entity indices

proggeramlug:
## Summary

- make nested stable-packed loop proofs path-sensitive: semantic calls dirty the active proof, while call-free paths reuse the admitted live receiver
- revalidate a dirty receiver exactly at its next indexed read and side-exit through one generic read without replaying earlier effects in the iteration
- cache repeated reads of the same guarded `receiver[counter]` value (the four `arch[j]` reads in Wolf ECS), invalidating the cache at every semantic call
- preserve normal JavaScript behavior for getters, proxies, mutation, exceptions, captured bindings, and moving-GC relocation

## Why

After #8837, `noctjs/wolf-ecs/simple_iter` still called `js_packed_arraylike_loop_revalidate_live` before every nested `arch[j]` read. The unchanged upstream kernel reads the same entity index four times per iteration, even though the successful proof remains valid on call-free paths.

This change keeps the proof and the first exact value locally. It does not retain a raw receiver or boxed pointer across a semantic call: the call emission choke points set a dirty bit first, and the next access reloads the rooted receiver and revalidates it. A failed revalidation performs the current source read through the generic path and rejoins, so prior getters/effects are never replayed.

## Performance

Pinned Mac mini, unchanged `noctjs/wolf-ecs/simple_iter`, interleaved precompiled binaries, `taskpolicy -t 0 -l 0`, semantic projection checked on every process:

| cohort | control | candidate | improvement | wins | oracles |
| strict 11-pair complete mechanism | 0.082878 ms/op | 0.075566 ms/op | 8.82% | 11/11 | 30/30 |
| quick 5-pair final first-read refinement | 0.087091 ms/op | 0.071131 ms/op | 18.33% median | 4/5 | 16/16 |

The final quick cohort was host-contended (two late spikes; high CV), so it is development evidence only. A fresh strict run was attempted but never admitted because the host could not sustain the documented 65% idle gate; no rejected samples are presented as strict evidence.

For context, the retained exact Node median is 0.004942 ms/op. This PR therefore removes a measured local cost but does not close the remaining gap. Native sampling attributes 704/730 samples to the generated system kernels; the next dominant target is repeated erased TypedArray kind/data/length dispatch on every component-column load/store.

## Validation

- `cargo check --release -p perry-codegen --all-targets`
- `cargo test --release -p perry-codegen --lib active_stable_packed_proofs_are_dirtied_only_by_executed_non_intrinsic_calls -- --nocapture`
- `cargo test --release -p perry --test issue_8773_closure_capture_packed_loops -- --nocapture` (4/4; includes normal and forced-moving-GC executions)
- `cargo test --release -p perry --test issue_8774_argument_shape_clones --test issue_8775_imported_object_specialization -- --nocapture` (4/4)
- `git diff --check`

## Summary by CodeRabbit

  * Improved performance for repeated indexed reads in nested packed-array loops.
  * Reuses safe cached reads when loop values remain unchanged.
  * Maintains correct behavior after calls, mutations, proxies, getters, exceptions, and garbage collection.

* **Bug Fixes**
  * Added reliable fallback handling when cached or optimized reads become invalid.
  * Preserved accurate results for nested packed-array access across clean and modified execution paths.

* **Tests**
  * Expanded regression coverage for cache reuse, invalidation, revalidation, and fallback behavior.

### DIFF
--- a/changelog.d/8839-guarded-ecs-entity-index-cache.md
+++ b/changelog.d/8839-guarded-ecs-entity-index-cache.md
@@ -1,0 +1,6 @@
+Nested stable-packed ECS loops now reuse an admitted receiver proof on call-free paths and cache
+repeated reads of the same entity index. Semantic calls invalidate both facts before execution; the
+next indexed access reloads the rooted receiver and revalidates it, with an exact generic read on
+failure rather than replaying prior iteration effects. This preserves getters, proxies, exceptions,
+mutation, and moving-GC behavior while making the unchanged Wolf `simple_iter` kernel 8.82% faster
+in an 11-pair controlled cohort (11/11 wins and 30/30 semantic-oracle passes).
--- a/crates/perry-codegen/src/block.rs
+++ b/crates/perry-codegen/src/block.rs
@@ -92,6 +92,18 @@ pub struct RegCounter {
     /// callee is UB, so the registry, not the emitting code, is the single
     /// source of truth. `None` for functions built outside a module (tests).
     preserve_none_fns: RefCell<Option<Rc<RefCell<HashSet<String>>>>>,
+    /// Compiler-private validity bits for nested stable-packed loop proofs.
+    ///
+    /// A guarded inner receiver may keep its raw address across call-free
+    /// direct-load/store arms. Every actually executed runtime/indirect call
+    /// that can collect or run semantic heap work dirties the active proofs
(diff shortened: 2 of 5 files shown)


proggeramlug:
@coderabbitai review

proggeramlug:

PR #8843, opened 2026-08-26
fix(runtime): reject oversized generic Array slice results

proggeramlug:
## Summary

- compute generic `Array.prototype.slice` bounds before allocating or narrowing the result length
- throw `RangeError` for result lengths above `2^32 - 1` before indexed reads, while allowing small slices from larger array-like receivers
- route borrowed `Array.prototype.slice` aliases on plain objects through the same generic engine
- preserve holes and root the receiver, result, and copied values across getter/proxy collection points
- add an integration regression covering plain `.call`, assigned aliases, proxies, and a valid one-element tail slice

## Test262

Pinned test262: `4249661388e5d3f92a85186213da140a6481490f`

- `built-ins/Array/prototype/slice`: **52 pass, 0 diff, 0 runtime-fail, 0 compile-fail** (100%)
- full `built-ins/Array`: **2513 pass, 0 diff, 14 runtime-fail, 0 compile-fail** across 2527 judged cases (99.4%)
- exact failure-set comparison against the clean-main baseline found no added failures; all four targeted slice failures were removed (the baseline's unrelated transient linker failure also passed on rerun)

## Tests

- `cargo test --release -p perry --test issue_5898_array_slice_invalid_length -- --nocapture`
- changed files: `rustfmt --edition 2021 --check ...`
- `cargo fmt --all -- --check` was attempted twice, including from a short mapped path, but Cargo exceeds Windows' argument-list limit (`os error 206`) for this workspace; the three changed Rust files pass direct rustfmt checking

No version, Cargo manifest/lockfile, or changelog changes.

Refs #5898

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `Array.prototype.slice` for array-like objects, preserving sparse elements and correct indexed-property behavior.
  * Added proper handling for oversized slice results, including raising `RangeError` before unnecessary property access.
  * Preserved expected behavior for arrays, arguments objects, typed arrays, proxies, and aliased slice methods.

* **Tests**
  * Added regression coverage for invalid slice lengths and large array-like objects.

### DIFF
--- a/crates/perry-runtime/src/array/generic.rs
+++ b/crates/perry-runtime/src/array/generic.rs
@@ -1145,8 +1145,9 @@ pub extern "C" fn js_arraylike_includes(recv: f64, value: f64, from: f64, has_fr
 }
 
 // ---------------------------------------------------------------------------
-// at / join / slice — no callback identity concerns; materialise where it
-// keeps the implementation simple (slice/join build fresh results anyway).
+// at / join / slice — no callback identity concerns. Join materialises its
+// receiver; slice copies only its selected interval so oversized array-like
+// lengths can be validated before allocation or indexed reads.
 // ---------------------------------------------------------------------------
 
 #[no_mangle]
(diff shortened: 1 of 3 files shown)


proggeramlug:

PR #8845, opened 2026-08-26
perf: specialize guarded ECS typed columns

proggeramlug:
## Summary

- recognize the narrow call-free ECS kernel where an admitted packed entity list indexes two to four erased component locals
- fuse source-layout and component-column admission into one runtime guard that requires exact, pairwise-distinct, equal-length owning `Uint32Array`s
- persist an exact nonnegative-integer prefix proof on the Array-subclass source and normalize its entity IDs to Perry's compact `INT32` Number representation
- install loop-local noalias Uint32 views so component reads/writes lower to direct native loads/stores, while the source loop uses one contiguous base
- preserve out-of-bounds semantics: unequal component lengths decline the clone; an out-of-range entity on equal-length columns skips the no-effect iteration rather than replaying earlier work
- retain the unchanged generic clone for every guard/matcher miss

This is stacked on #8839. Once #8839 lands, this PR can be retargeted to `main` without changing its diff.

## Why

After #8839, the Wolf ECS hot loop still paid erased TypedArray receiver/kind/data/length dispatch for every component access. The generated system kernel dominated the profile. Runtime admission can establish the stronger representation and alias facts once, allowing the loop itself to operate on raw Uint32 lanes.

The final AArch64 inner loop is reduced to one compact entity-id load, one component bound check, two component loads, two stores, and integer induction/control.

## Performance

Pinned Mac mini, unchanged `noctjs/wolf-ecs/simple_iter` oracle, precompiled full-runtime binaries (`--no-auto-optimize`), `taskpolicy -t 0 -l 0`:

| cohort | median ms/op | result |
|---|---:|---:|
| #8839-era guarded-loop control | ~0.0705 | baseline |
| this branch, regenerated | 0.005281 | ~13.3x faster / ~92.5% lower latency |
| Node v26.5.1, fresh 5-process sample | 0.005124 | Perry is ~3.1% behind |

Cleanup A/B against the retained v10 mechanism artifact was neutral, as expected: 0.005281 vs 0.005276 ms/op (-0.10%, noise), 16/16 semantic oracles, identical 207,847,424-byte max RSS. The fresh Node cohort passed 8/8 semantic oracles.

The quick cohorts intentionally used a one-sample idle gate and are development evidence, not a strict publication-quality run. JSON artifacts and binaries were retained locally under `/private/tmp/perry-ecs-artifacts-8772`.

## Validation

- `cargo fmt --all`
- `git diff --check`
- `cargo check --release -p perry-codegen -p perry-runtime`
- `cargo test --release -p perry-runtime --lib array::subclass_tests -- --nocapture` (10/10)
- full Wolf semantic/rate oracle from a matching full-runtime build
- alternating 5-pair mechanism A/B (16/16 oracle executions)
- fresh 5-process Node cohort (8/8 oracle executions)

## Follow-up

The remaining measured gap is only ~3%. The next useful optimization is backend-level control-flow amortization: emit the already-matched swap kernel directly (or attach loop-specific LLVM unroll metadata) so it can be unrolled/software-pipelined without lowering the HIR body twice. A manual HIR-body duplication experiment was rejected because it perturbed compiler fact state outside the kernel; no part of that experiment is included here.

### DIFF
--- a/crates/perry-codegen/src/expr/i32_fast_path.rs
+++ b/crates/perry-codegen/src/expr/i32_fast_path.rs
@@ -1588,6 +1588,11 @@ fn lower_expr_native_u32(ctx: &mut FnCtx<'_>, e: &Expr) -> Result<LoweredValue> 
     if let Some(lowered) = lower_packed_u32_loop_index_get(ctx, e)? {
         return Ok(lowered);
     }
+    if let Expr::IndexGet { object, index } = e {
+        if let Some(lowered) = super::try_lower_proven_view_checked_u32_load(ctx, object, index)? {
+            return Ok(lowered);
+        }
+    }
     if let Some(lowered) = crate::expr::lower_expr_value(ctx, e)? {
         let value = match lowered.rep {
             NativeRep::I32 | NativeRep::U32 | NativeRep::BufferLen => Some(lowered.value),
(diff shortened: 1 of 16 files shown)


proggeramlug:

PR #8848, opened 2026-08-26
perf: remove residual Wolf ECS admission overhead

proggeramlug:
## Summary

- cache successful owning `Uint32Array` admissions in the existing direct-mapped TypedArray cache and invalidate the entry whenever the live receiver ceases to satisfy that ownership contract
- fast-path Array-subclass `.length` misses through the existing exact-ShapeId dense-subclass proof while retaining ordinary lookup for accessors, prototype overrides, sparse layouts, and unrelated objects
- keep both changes general and fail-closed; no upstream source, benchmark-specific name, or semantic shortcut is introduced

## Why

After #8847, `noctjs/wolf-ecs/simple_iter` had a branch-light generated swap loop, but every invocation still paid two avoidable runtime costs:

1. repeated owning-`Uint32Array` registry admission for the same component columns; and
2. full generic object lookup for polymorphic `Query` / `Archetype` `.length` reads because Array subclasses use `ObjectHeader`, not `GC_TYPE_ARRAY`.

The `.length` miss was the material residual. The cache is a smaller supporting improvement.

## Performance

Pinned Apple-silicon Mac mini, unchanged strong Wolf oracle, precompiled binaries, alternating order, `taskpolicy -t 0 -l 0`:

| comparison | control | candidate | result | wins | semantic oracles |
| owning-U32 admission cache vs #8847 head | retained exact control | retained exact candidate | **0.530% paired improvement** | 10/15 | 38/38 |
| subclass-length fast path vs cache-only | 0.00525835 ms/op | 0.00346675 ms/op | **33.93% lower latency** | 9/9 | 26/26 |
| final Perry vs Node 26.5.1 | Node 0.00509194 ms/op | Perry 0.00347725 ms/op | **Perry 31.81% faster** | 9/9 Perry | 26/26 |

A separate aggregate entity-range proof plus explicit loop-unroll experiment was tested and rejected: its quiet-host 9-pair paired median was 0.55% worse with 4/9 wins. None of that experiment is included here.

## Validation

- `git diff --check`
- `cargo check --release -p perry-codegen -p perry-runtime`
- `cargo test --release -p perry-runtime --lib owning_u32_admission_cache_skips_registry_and_invalidates -- --nocapture`
- `cargo test --release -p perry-runtime --lib array_length_fast_path_tests -- --nocapture` (2/2)
- `cargo test --release -p perry-runtime --lib array::subclass_tests -- --nocapture` (10/10)
- unchanged Wolf semantic oracle in every performance process

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `length` handling for array subclasses while preserving standard object behavior.
  * Added fallback handling for unsupported array layouts.

* **Performance**
  * Improved repeated `Uint32Array` access through caching.
  * Ensured cached results are invalidated when arrays are unregistered.

* **Tests**
  * Added coverage for array subclass lengths and cache reuse/invalidation.

### DIFF
--- a/crates/perry-runtime/src/object/field_get_set/ic_miss.rs
+++ b/crates/perry-runtime/src/object/field_get_set/ic_miss.rs
@@ -508,19 +508,35 @@ pub extern "C" fn js_object_get_field_ic_miss(
         // run time — more than the entire polymorphic-dispatch fix above saved.
         //
         // `GC_TYPE_ARRAY` is a genuine dense array: buffers, typed arrays, lazy
-        // arrays, Sets and Maps all carry their own distinct `obj_type`, and an
-        // `class X extends Array` instance is an `ObjectHeader`
-        // (`GC_TYPE_OBJECT`). `js_array_length` still resolves growth-forwarding
-        // stubs, proxies and subclass receivers, so this only skips probes that
-        // cannot match — the expression returned is exactly the one
-        // `get_field_by_name_object_tail`'s array arm computes for this key,
-        // which is what makes it a pure short-circuit rather than a second
-        // implementation.
-        if unsafe { gc_type_of(obj) } == Some(crate::gc::GC_TYPE_ARRAY)
-            && unsafe { key_bytes_are(key, b"length") }
-        {
-            let arr = obj as *const crate::array::ArrayHeader;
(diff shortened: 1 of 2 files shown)


proggeramlug:

PR #8849, opened 2026-08-26
perf: optimize proven Array length truncation

proggeramlug:
## Summary

Routes strict, statically proven Array `length` writes through the existing ArraySetLength lowering and bulk-truncates ordinary dense arrays without a descriptor/name-map deletion walk per element.

## Changes

- preserve static Array information for same-receiver strict `PutValueSet` length writes
- retain generic lowering for sloppy writes and explicit-receiver cases
- add a guarded dense truncation path only when indexed descriptors and side-table properties are absent
- clear discarded slots to holes and rebuild the live-prefix GC layout once
- add positive, negative, and stale-index regression coverage

## Related issue

n/a

## Test plan

- `python3 scripts/gc_store_site_inventory.py --self-test`
- `python3 scripts/gc_store_site_inventory.py`
- `cargo test -p perry-codegen --lib -- --test-threads=1`
- `cargo test -p perry-runtime --lib -- --test-threads=1` (2,699 passed, 4 ignored)
- Apple-silicon Mac mini, 11 alternating exact-parent/candidate pairs on codehz/ecs command-buffer workload: 31.541 ms to 28.354 ms median, 10.030% median paired improvement, 11/11 wins, 22/22 semantic oracles
- same-host Node 26.5.1/Perry cohort: 2.896 ms / 28.342 ms medians

- [x] Added or updated tests in the affected crates
- [ ] Documentation update is not required; no CLI or public runtime API changes
- [ ] No platform UI backend touched

## Screenshots / output

Retained command-buffer LLVM changes three generic property-PIC misses into three direct `js_array_set_length_strict` calls. A codegen-only screen stayed below the 2% admission floor; the submitted compiler/runtime mechanism is the independently qualified 10.03% candidate.

## Checklist

- [x] My commits follow the repository commit-prefix convention

## Summary by CodeRabbit

  * Improved performance when reducing the length of dense arrays.
  * Array truncation now more efficiently clears removed elements while preserving correct behavior in complex cases.

* **Bug Fixes**
  * Fixed stale array values and indexed properties reappearing after an array is truncated and expanded again.

* **Tests**
  * Added coverage for optimized length assignments and verified correct handling across supported array scenarios.

### DIFF
--- a/changelog.d/8849-array-length-truncation.md
+++ b/changelog.d/8849-array-length-truncation.md
@@ -1,0 +1,6 @@
+Strict writes to a statically proven Array's `length` now retain their ArraySetLength lowering,
+and ordinary dense truncation clears discarded slots in one guarded runtime region instead of
+performing descriptor and side-table deletion work for every element. The generic path remains for
+sloppy, explicit-receiver, sparse, and descriptor-bearing cases. On the unchanged codehz/ecs
+15k-command workload, an 11-pair Apple-silicon cohort improved the median from 31.541 ms to
+28.354 ms (10.03%, 11/11 wins, 22/22 semantic-oracle passes).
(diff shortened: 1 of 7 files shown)


proggeramlug:

Issue #608, opened 2026-05-08
Cross-module tagged template literal: rest params not packed into array at callee


proggeramlug:
## Repro

```js
// node_modules/repro-pkg2/sql.js
export function sql(strings, ...params) {
  console.log('strings.length:', strings && strings.length);
  console.log('params:', params);
  console.log('params.length:', params && params.length);
}
```

```ts
// probe_tagged_xmod.ts (with repro-pkg2 in package.json#perry.compilePackages)
import { sql } from 'repro-pkg2';

sql`default`;             // 0 interp
const x = "world";
sql`hello ${x}`;          // 1 interp
```

**Bun output (correct)**:
```
strings.length: 1
params: []
params.length: 0
strings.length: 2
params: [ "world" ]
params.length: 1
```

**Perry output (bug)**:
```
strings.length: 1
params: undefined
params.length: undefined        ← rest param not bound at all
strings.length: 2
params: world                   ← rest param is the raw arg, not [arg]
params.length: 5                ← reads the string's .length, not array length
```

The `strings` array is fine in both cases (perry passes the static-string array correctly). But the rest parameter `...params` is broken:
- 0 interpolations → `params === undefined` (should be `[]`)
- 1 interpolation → `params === firstArg` (should be `[firstArg]`)

Looks like the tagged-template call site emits `sql(strings, ...interp_args)` as plain positional args without packing the trailing args into a rest array; the callee's `...params` rest binding then either doesn't fire (no trailing args) or gets the single arg directly instead of a 1-element array.

## What works

Same-module tagged template literal works correctly:

```ts
// probe_tagged.ts — sql defined and used in the same file
function sql(strings: any, ...params: any[]) { ... }
sql`default`;     // params = []  ✅
sql`hello ${x}`;  // params = [x] ✅
```

So the bug is specifically the cross-module / `compilePackages` call path — whatever shape lowers `sql` from an imported `ExternFuncRef` doesn't honour the callee's rest-param signature.

## Why it matters for #488

Blocks the drizzle + @perryts/postgres acceptance test, immediately after #597 unblocked the for-of-on-Any-typed-arr.entries() path.

drizzle's `dialect.buildInsertQuery` calls `valueList.push(sql\`default\`)` (drizzle-orm `pg-core/dialect.js:390`), where `sql` is the tagged-template helper imported from `'../sql/sql.js'`. Inside that helper:

```js
function sql(strings, ...params) {
  const queryChunks = [];
[...]

PR #8852, opened 2026-08-26
codegen(calls): pad under-applied same-module direct calls with undefined (#8770)

proggeramlug:
## The bug

A **same-module direct call with fewer arguments than the callee's declared parameter count** lowered only the provided args, leaving the remaining FP argument registers holding caller-saved garbage — which the callee then read as JS values.

The cross-module twin (`extern_func.rs`, the issue #608 arm) has always padded missing trailing args with `TAG_UNDEFINED`. The same-module plain arm in `func_ref.rs` sat "one `else` away" (#7154's own comment about this exact file-pair pattern) unpadded. The three rest-parameter arms pad via `lower_rest_call_args_rooted`; only the plain arm was missed.

## Impact (this is the #8770 Claude Code crash family)

The Claude Code bundle is one giant module, so **every** direct call resolves through the same-module arm — and minified JS under-applies constantly. `aP([q])` for `function aP(q, K = !1, _)` handed `K`/`_` whatever `d1`/`d2` held after `js_array_from_values`: impossible-NaN bit patterns (`0xffffffffffffffff`) that flowed into truthiness tests (`if (_)`) and method receivers (`_.get(A)`) and faulted in `shape_is_url_search_params` / `js_is_truthy` (~60% of `cc -p` runs SEGV), or silently corrupted the async iteration (`Detected unsettled top-level await`, most of the rest).

The GC-knob correlations the long #8770 investigation chased (scavenge pacing, `PERRY_CONSERVATIVE_STACK_SCAN=Full` "fixing" it) were register-content side effects of the missing padding — different collector paths leave different residue in the never-written argument registers — not collector bugs.

## The fix

In `try_lower_func_ref_call`'s plain (non-rest) arm, after lowering the provided args, pad `lowered` with the `TAG_UNDEFINED` literal up to `declared_count` — mirroring the cross-module twin. `declared_count` is the full `f.params.len()` from the function registry (includes default and trailing params, which is what the callee's compiled signature and its default-parameter `undefined` tests expect).

## Validation

- **Claude Code `cli_2.1.112.js` (68 MB bundle): `cc -p` 30/30 clean** — 0 SEGV, 0 hangs, 0 unsettled awaits, node-identical output on every run (baseline: ~60% SEGV, ~73% corrupted async).
- `--version` / `--help` unchanged-clean.
- New regression test (`underapply_pad_tests.rs`): an under-applied direct call to a 3-param callee must emit all three double args, the omitted two as the `TAG_UNDEFINED` literal.
- Full `perry-codegen` suite: 1223 passed, 0 failed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed same-module function calls with omitted trailing arguments.
  * Missing arguments are now correctly treated as `undefined`, matching calls across modules.

* **Tests**
  * Added regression coverage to verify that under-applied calls pass all declared parameters with appropriate `undefined` values.

### DIFF
--- a/crates/perry-codegen/src/lower_call/func_ref.rs
+++ b/crates/perry-codegen/src/lower_call/func_ref.rs
@@ -975,6 +975,22 @@ pub fn try_lower_func_ref_call(
         let (values, guard) = super::lower_call_args_rooted(ctx, args)?;
         arg_group = guard;
         lowered.extend(values);
+        // #8770: pad missing trailing args with TAG_UNDEFINED, exactly like
+        // the cross-module twin (`extern_func.rs`, issue #608 arm). The callee
+        // is compiled with `declared_count` double parameters and its
+        // default-parameter lowering tests each for `undefined`; an
+        // under-applied same-module call site that emits only the provided
+        // args leaves the remaining FP argument registers holding caller-saved
+        // garbage, which the callee then reads as JS values. On the Claude
+        // Code bundle (one giant module, so EVERY direct call resolves here)
+        // `aP([q])` for `function aP(q, K = !1, _)` handed `K`/`_` whatever
+        // d1/d2 held after `js_array_from_values` — the #8770 poison values
+        // (0xffffffffffffffff receivers → shape_is_url_search_params /
+        // js_is_truthy faults, corrupted async iteration → unsettled awaits).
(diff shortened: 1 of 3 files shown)


proggeramlug:

PR #8860, opened 2026-08-26
perf(map): remove outer hash from numeric-key lookups

proggeramlug:
## Summary

- keep each Map's numeric-key index in its existing `MapSideAllocation` owner
- store a direct, non-owning pointer to that stable index in `MapHeader`
- remove the thread-local outer `MapHeader* -> numeric index` hash lookup from numeric `Map.get`/`has`/`set`
- derive fixed GC layout sizes from `MapHeader` and `SetHeader` after the Map header grows

The index remains owned and freed with the Map's registered side allocation. Its `Box` address is stable across entries-buffer growth, registry rehashes, and moving-GC registry rekeys.

## Validation

- `cargo test -p perry-runtime --lib`: 2,703 passed, 0 failed, 4 ignored
- `git diff --check`
- exact release compiler/runtime builds for latest `main` (`926d95782`) and this commit (`d4e53d579`)
- existing copying-minor Map relocation test verifies numeric lookup/index survival after `MapHeader` movement
- new regression verifies the direct index pointer stays stable while the entries buffer grows from 4 to 64 keys

## Performance

Mac mini, Apple M1, unchanged upstream `@codehz/ecs` comprehensive command-buffer row (`5k entities: 3 commands each + sync`, 15k commands), alternating exact control/candidate processes, 2 warmups + 6 measured rounds per process:

- control median: **28.5104 ms**
- candidate median: **27.5947 ms**
- median paired improvement: **3.188%**
- candidate wins: **11/11**
- paired range: **2.967% to 3.639% faster**
- semantic oracles: **22/22 passed**

Binary SHA-256:

- control: `76389e4f9f9bf0172cfbb4e8b4806fde3389ca0317012ab56b3aa6ff5cf3c191`
- candidate: `b035ec1fc595abfc3b395f20ac914a2315561c4f36b10f57338a1b722c7ae483`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Map and Set memory-size calculations across different layouts.
  * Improved Map numeric-key indexing reliability during growth, deletion, garbage collection, and memory relocation.
  * Preserved numeric-index stability as Map storage expands.
  * Strengthened Map validation and size handling for edge cases.

### DIFF
--- a/crates/perry-runtime/src/gc/copying_pointer_set.rs
+++ b/crates/perry-runtime/src/gc/copying_pointer_set.rs
@@ -304,10 +304,13 @@ pub(super) unsafe fn plausible_gc_header(header: *mut GcHeader, arena: bool) -> 
 /// ~1024 (the top 32 bits of the address), which never matches the
 /// fixed total.
 fn fixed_layout_total_size(info: &GcTypeInfo) -> Option<usize> {
-    // MapHeader { size: u32, capacity: u32, entries: *mut f64 } = 16 B.
-    // SetHeader { size: u32, capacity: u32, elements: *mut f64 } = 16 B.
     match info.type_id {
-        crate::gc::GC_TYPE_MAP | crate::gc::GC_TYPE_SET => Some(GC_HEADER_SIZE + 16),
+        crate::gc::GC_TYPE_MAP => {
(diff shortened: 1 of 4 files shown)


proggeramlug:
Landed on `main` via the #8865 batch.

Issue #5227, opened 2026-06-16
In cooked (non-raw) mode, process.stdin data events never fire


jroweboy:
## What happened

No data is returned when using process.stdin without it being in raw mode. The repro code prints `bytes=0`

## What you expected

The repro code should print `bytes=12` (or however many you pass into the command line)

## Minimal reproduction

```typescript
function main() {
  let bytes = 0;
  // NB no setRawMode call here, which is exactly why the data event never fires.
  // (param left untyped: Perry mis-parses a typed param whose type name is not
  // referenced earlier in the file - unrelated to this stdin bug.)
  process.stdin.on('data', (chunk) => {
    bytes += typeof chunk === 'string' ? chunk.length : chunk.length;
  });
  process.stdin.on('end', () => {
    console.log('bytes=' + bytes);
  });
  process.stdin.resume();
}

main();
```

Command you ran:

(Due to #5000 which affects reading from `stdin` as well, I have the perry source locally, and i have to build with `PERRY_WORKSPACE_ROOT`)
```bash
PERRY_WORKSPACE_ROOT=/path/to/perry/ perry compile perry-bug-cooked-stdin-data-no-fire.ts -o ./perry-bug-cooked-stdin-data-no-fire.exe
printf "hello world\n" | ./perry-bug-cooked-stdin-data-no-fire.exe
```

## Environment

- **Perry version**: perry 0.5.1167
- **Host OS**: Windows 11
- **Target**: native
- **Rust toolchain** (if building from source): rustc 1.96.0 (ac68faa20 2026-05-25)
- **Installed via**: from source

## Diagnostic output

Nothing to note.

## Anything else

Only work around I've found is forcibly using raw mode for stdin.

proggeramlug:
On it as well!

proggeramlug:
Should work now :)

PR #8861, opened 2026-08-26
fix(stdin): deliver piped stdin to `-p` — lower `once`, keep every `end` listener, buffer pull-mode bytes

proggeramlug:
## Symptom

`echo hi | claude -p "summarize"` produced **nothing** — exit 0, zero bytes on stdout *and* stderr — where node prints the result. `printf "" | claude -p "…"` worked, which is what made it confusing.

Three independent defects in the `process.stdin` path stack up to produce that. Each is fixed here; all three are needed.

## 1. `process.stdin.once(…)` was never lowered

`perry-hir` matched only `("stdin", "on") | ("stdin", "addListener")`, so `once` fell through to the generic member-call path and never reached `js_readline_stdin_on` — the listener was never registered with the fd-0 reader and simply never fired.

That is the decisive one. Claude Code's print-mode reader is:

```js
process.stdin.on("data", acc);
const timedOut = await race(process.stdin.once("end"), timeout(3000));
```

With `once` dropped, the `end` half of the race could never win. It always fell through to the timer — and because that timer is `unref`'d, nothing kept the event loop alive, so the process exited 0 having printed nothing.

## 2. Only one `stdin.on("end")` listener survived

They shared readline's single-slot `CLOSE_CALLBACK` ("only one terminal close listener is supported per process"), so every registration clobbered the previous one. Node allows any number, and the bundle registers **three** — the one that resolves its read-stdin promise was silently dropped.

Now `STDIN_END_CALLBACKS`, a list fired in registration order, honoured by the keep-alive predicate and by `removeListener`.

## 3. Pull-mode (`on("readable")` + `read()`) bytes were discarded

The fd-0 reader routes bytes by mode: raw → `PENDING_DATA`, `data`-flowing → `PENDING_DATA`, **everything else → `PENDING_LINES`** — readline's *line* queue, which `process.stdin.read()` never drains. Paused/pull mode sets neither flag, so its bytes were consumed off fd 0 and thrown away; `read()` returned null forever and the loop parked in `kevent`.

New `STDIN_PULL_MODE` flag, set while a `readable` listener exists, routes those bytes (and the EOF trailing chunk) to the buffer `read()` actually drains. This is the same hazard the `PENDING_LINES` comment already records for the `data` case (#5227), left unfixed for `readable`.

## Verification

Against node, with the real bundle and with focused replicas:

| probe | before | after | node |
|---|---|---|---|
| `on("readable")` + `read()` | `""` | `"hello pipe"` | `"hello pipe"` |
| three `on("end")` listeners | 1 fires | `end1,end2,end3` + data | same |
| replica of the `-p` reader | always took the 3 s timeout | `end` wins the race | same |

## Tests

- `crates/perry-hir/tests/process_stdin_once_lowering.rs` — **sabotage-checked**: both cases fail with the `once` arm removed, pass with it.
- `every_stdin_end_listener_fires` and `readable_listener_enables_pull_mode` in the readline suite; `test_support::reset` clears the new state so the suite stays order-independent.
- Full `perry-stdlib` readline suite green (19/19), `perry-hir` lowering test green (2/2).

Found while driving a perry-vs-node differential parity harness over the Claude Code bundle. Related: #8770 / #8852 (under-applied direct calls), landed via #8857.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `process.stdin.once(...)` so one-time listeners receive stdin events consistently.
  * Multiple stdin end and close listeners are now preserved and invoked in registration order.
  * Improved readable-mode input handling, including trailing data received at end-of-file.
  * Corrected stdin pull-mode activation and deactivation as readable listeners are added or removed.
  * Ensured pending stdin callbacks keep processing active until all callbacks have run.

### DIFF
--- a/crates/perry-hir/src/lower/expr_call/module_class_static.rs
+++ b/crates/perry-hir/src/lower/expr_call/module_class_static.rs
@@ -205,7 +205,33 @@ pub(super) fn try_module_class_static(
                                     let arg = args.into_iter().next().unwrap();
                                     return Ok(Ok(Expr::ProcessStdinSetRawMode(Box::new(arg))));
                                 }
-                                ("stdin", "on") | ("stdin", "addListener") if args.len() >= 2 => {
+                                // `once` lowers here too. Without it, only
+                                // `on`/`addListener` reached readline's stdin
+                                // listener registry and `process.stdin.once(…)`
+                                // fell through to the generic member-call path,
+                                // which never registers with the fd-0 reader —
+                                // so the callback simply never fired.
+                                //
+                                // That is not a corner case: Claude Code's `-p`
+                                // stdin reader awaits
+                                // `race(stdin.once("end"), timeout(3000))`, so
+                                // with `once` dropped the `end` half could never
+                                // win. The race fell through to the timeout, and
+                                // because that timer is unref'd nothing kept the
+                                // loop alive — `echo hi | claude -p …` exited 0
+                                // having printed NOTHING (node prints the result).
+                                //
+                                // One-shot semantics are handled downstream by
+                                // the pump, which takes the `end` listener list
+                                // when it fires; `data`/`readable` listeners
+                                // registered via `once` are a documented
+                                // residual (they behave like `on`) — the streams
+                                // that matter here are EOF-driven.
+                                ("stdin", "on")
+                                | ("stdin", "addListener")
+                                | ("stdin", "once")
+                                    if args.len() >= 2 =>
+                                {
                                     let mut iter = args.into_iter();
                                     let event = iter.next().unwrap();
                                     let handler = iter.next().unwrap();
--- a/crates/perry-hir/tests/process_stdin_once_lowering.rs
+++ b/crates/perry-hir/tests/process_stdin_once_lowering.rs
@@ -1,0 +1,58 @@
+//! `process.stdin.once(event, handler)` must lower to `Expr::ProcessStdinOn`,
+//! exactly like `on` / `addListener`.
+//!
+//! Only `on` and `addListener` were matched, so `once` fell through to the
+//! generic member-call path and never reached `js_readline_stdin_on` — the
+//! callback was never registered with the fd-0 reader and simply never fired.
+//!
+//! That is what broke `echo hi | claude -p "…"`. Claude Code's print-mode
+//! stdin reader is:
+//!
+//! ```js
+//! process.stdin.on("data", acc);
+//! const timedOut = await race(process.stdin.once("end"), timeout(3000));
+//! ```
+//!
+//! With `once` dropped, the `end` half of that race could never win. The race
+//! always fell through to the timer — and because that timer is `unref`'d,
+//! nothing kept the event loop alive, so the process exited 0 having printed
+//! nothing at all (node prints the result).
+
+use perry_diagnostics::SourceCache;
+use perry_hir::{clear_current_module_source, lower_module};
+use perry_parser::parse_typescript_with_cache;
+
+fn lower_debug(src: &str) -> String {
+    let mut cache = SourceCache::new();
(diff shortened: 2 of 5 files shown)


proggeramlug:
Landed on `main` via #8863, with one fix.

`gc_runtime_root_holders` flagged the new `STDIN_END_CALLBACKS: Mutex<Vec<i64>>` — it stores closure pointers and no registered scanner reached it. That is a genuine missing root rather than a classification gap: `scan_readline_roots_mut` already visits `DATA_CALLBACKS`, `KEYPRESS_CALLBACKS`, `READABLE_CALLBACKS` and the single-slot `CLOSE_CALLBACK` this list replaces, so the new list just wasn't added. Its closures are reachable only from there between registration and EOF, so a collection in that window would leave stale pointers the pump then calls — the failure mode that goes bad at collection #0 and stays bad. Added `&STDIN_END_CALLBACKS` to the scanner.

Also split `readline/mod.rs` (2049 → 1796) by extracting its test module, and ran fmt.

Verified your pump claim while auditing: `std::mem::take` on the list means each `end` listener fires exactly once and all fire in registration order.

One residual worth its own issue: `once("data")`/`once("readable")` behave like `on` since they route to the same registry. Your inline comment documents it, and it's still a clear improvement over `once` doing nothing — but it is a Node deviation.

proggeramlug:
Landed via #8863.

proggeramlug:
## Update: piped stdin now works end-to-end — byte-identical to node

Pushed `2706096dc`, which completes the fix. Measured on the real bundle:

```
echo "hello world" | cc -p "summarize"    0/8  ->  8/8   (node 8/8)
cc -p "hi" </dev/null                     6/6         (no regression)
diff node vs perry                        IDENTICAL
--output-format json                      matching structured result
```

### What the last commit fixes

The first commit added `end` handling only to `js_readline_stdin_on` — the extern codegen emits for a **literal** `process.stdin.on(…)`. Any other form (an alias, or stdin passed as a parameter) resolves the stdin object's native `on`/`once`/`addListener`, which delegate to readline through the `stdin_on_op` / `stdin_off_op` provider. That provider matched only `data`/`readable`/`keypress` and dropped `end` into `_ => return`.

Claude Code takes exactly that path — `X71(process.stdin, 3000)` then `stream.once("end", …)` on the parameter — so the `end` half of its `race(once("end"), timeout(3000))` could never win. The timer is `unref`'d, so nothing kept the loop alive and the process exited 0 printing nothing.

Two probes pinned it, and they only make sense together:

| probe | perry (before) | node |
|---|---|---|
| `end` registered 3 ways | only `direct-end` | all three |
| `data` registered 2 ways | only `alias-data` | both |

Opposite halves winning is the signature of **two listener registries, each starting its own fd-0 reader**, with registration order deciding the winner.

Also: `removeListener`'s `end`/`close` arm now clears **both** stores. Adding a second arm with the same pattern made the pre-existing one unreachable, which would have left the legacy `CLOSE_CALLBACK` slot uncleared — caught before it shipped.

### Things I tried and backed out

Both were plausible and neither showed measured benefit, so they are **not** in this PR:

* a `stdin_mark_eof()` bridge (readline reader → perry-runtime's `STDIN_EOF_SEEN`) — the aliased listeners were never *registered*, so signalling EOF changed nothing;
* binding the object's `on` field to the delegating wrapper — the `data` case was unchanged with it.

### Known remaining gap (deliberately not fixed here)

A `data` listener registered through an alias **before** any literal `process.stdin.*` call still diverges: the readline provider is registered lazily from `try_register_pump()`, so the aliased call finds no provider, falls back to perry-runtime's own listener tables, and starts a second fd-0 reader that starves readline's listeners. The fix is eager provider registration at stdlib init — kept separate because it changes startup ordering.

Tests: `provider_path_registers_end_listeners`, `provider_path_removes_end_listeners` added; readline suite 21/21, hir lowering test 2/2 (sabotage-checked).

PR #8862, opened 2026-08-26
perf(object): reuse preinstalled birth shapes

proggeramlug:
## Summary

- reuse the module-init ShapeId when allocating a compiled class whose canonical keys pointer and live inline-slot bound still match
- avoid publishing and hashing the same `ShapeFacts` again during every matching object birth
- preserve the existing exact mint-and-validate fallback for learned-width changes, missing worker-local descriptors, foreign ids, and semantic/class shapes
- retain old-generation carrier bookkeeping when a directly stamped allocation is not in the nursery

## Correctness

Two new runtime regressions cover the direct matching-id path and a widened live-slot mismatch that must fail closed into an exact wider descriptor.

Mac-mini release runtime suite: **2704 passed, 0 failed, 4 ignored**.

## Performance

Unchanged `codehz/ecs` command benchmark (`5k entities: 3 commands each + sync`, 15k commands), M3 Mac mini, 11 alternating control/candidate pairs. Each process used 2 warmups, 6 measured rounds, and 64 operations per timed batch.

- control median: **25.612 ms**
- candidate median: **21.649 ms**
- median paired improvement: **15.61%**
- candidate wins: **11/11**
- semantic oracles: **22/22**
- paired range: **14.88% to 16.08% faster**

This PR is independent of the Map-header optimization in #8860; the measurement above uses current `main` as its control.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved object allocation when a compatible preinstalled shape is available.
  - Added validation to ensure shape metadata matches object keys, slot counts, generations, and object types.
  - Automatically falls back to an exact compatible shape when the supplied shape is unavailable or mismatched.
  - Prevented duplicate shape stamping during allocation.

- **Tests**
  - Added coverage for successful preinstalled-shape allocation and mismatch fallback behavior.

### DIFF
--- a/crates/perry-runtime/src/object/alloc.rs
+++ b/crates/perry-runtime/src/object/alloc.rs
@@ -269,7 +269,8 @@ fn object_alloc_class_inline_keys_impl(
     parent_class_id: u32,
     field_count: u32,
     keys_array: *mut ArrayHeader,
-) -> (*mut ObjectHeader, u32) {
+    preinstalled_shape_id: u32,
+) -> (*mut ObjectHeader, u32, bool) {
     if parent_class_id != 0 {
         register_class(class_id, parent_class_id);
     }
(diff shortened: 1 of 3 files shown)


proggeramlug:
Landed on `main` via the #8865 batch.

PR #8864, opened 2026-08-26
fix: clear v0.5.1519 r19 release blockers

proggeramlug:
## Summary

- mark both checked dynamic-callee unbox runtime boundaries as `C-unwind`, preserving caught TypeErrors instead of aborting
- keep the generated-code-only rebind anchor ABI in sync and add compile-time ABI guards
- refresh the native-value-profile parity expectation with its intentional POD-copy output

## Release evidence

r19 Full: https://github.com/PerryTS/perry/actions/runs/32967169957

- cargo-test: https://github.com/PerryTS/perry/actions/runs/32967169957/job/98172243105
- parity shard 12: https://github.com/PerryTS/perry/actions/runs/32967169957/job/98175183881
- Simulator passed: https://github.com/PerryTS/perry/actions/runs/32967172848

## Local checks

- `git diff --check`
- compile-time ABI signature guards
- exact native-value-profile snapshot assertion

No publisher was dispatched and no release/tag exists.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed errors when attempting to call non-callable values so they can be caught and handled correctly.
  - Improved stdin listener cleanup and removed duplicate handling that could trigger warnings.
  - Expanded native value profile coverage for POD-copy behavior.

- **Documentation**
  - Added a changelog entry summarizing these fixes.


proggeramlug:
Landed on `main` via #8866. No fixes needed — clean merge, all gates green as submitted. Verified the `readline/mod.rs` removal interacts correctly with #8861: the list-based `removeListener` arm and the `&STDIN_END_CALLBACKS` root-scanner entry both survive.

PR #8867, opened 2026-08-26
perf(map): index dense numeric key ranges

proggeramlug:
## Summary

- retain the existing numeric-key hash index as the complete fallback
- add a bounded, base-relative range index after at least eight exact nonnegative integer keys demonstrate a dense span
- serve covered hits and misses with a bounds check plus array load, independent of the absolute key magnitude
- keep sparse, fractional, negative, tagged, and oversized ranges on the hash path
- keep dense entry offsets exact across insertion, ordered deletion, and `Map.clear()`

The range is admitted only when its span is at most 4x the observed integer-key count and is capped at 1,048,576 slots. A run beginning at 1,000,000 therefore costs the same as a run beginning at zero, while isolated large keys cannot trigger a magnitude-sized allocation.

## Mac-mini evidence

Unchanged codehz/ecs row: `5k entities: 3 commands each + sync (15k total commands)`.

Exact 11-pair alternating cohort, repeat 64, two warmups and six measured rounds per process:

- control SHA-256: `03b8d566953b5ffc7915341b1c946de5941b4b1344ee7c20b9bbd761b9409208`
- candidate SHA-256: `e432c9d2d0124ccbc336d3458ae3539f2a3a9048658ddf92ef49231934301b44`
- control median: 20.989459 ms
- candidate median: 19.551747 ms
- median paired improvement: **6.852%** (range 5.907–8.061%)
- candidate wins: 11/11
- semantic process oracles: 22/22

The post-change native sample reduces `map::find_key_index` from 355/3,728 main-thread samples (9.52%) to 48/2,983 (1.61%), an 83.1% relative reduction in normalized share. That confirms that the measured gain comes from removing the intended hash probes.

Node 26.5.1 versus candidate, 11 alternating pairs, reports 2.882646 ms Node and 22.222847 ms Perry: 7.709x by independent medians and 7.642x by paired medians, with 22/22 semantic oracles. The preceding landed stack measured 8.172x.

## Verification

- `cargo test -p perry-runtime --lib --quiet`: 2,706 passed, 0 failed, 4 ignored
- focused Map module: 5 passed, including dense high-base activation, sparse/fractional/negative fallback, mixed-index ordered deletion, direct-index stability, and clear
- `git diff --check`

### DIFF
--- a/changelog.d/8867-dense-numeric-map-index.md
+++ b/changelog.d/8867-dense-numeric-map-index.md
@@ -1,0 +1,4 @@
+Numeric Maps now add a bounded range index when at least eight exact nonnegative integer keys form
+a dense run. Sequential high-base IDs use a bounds check and array load instead of a hash-table
+probe, while sparse, fractional, negative, tagged, and oversized ranges retain the existing hash
+fallback.
--- a/crates/perry-runtime/src/map.rs
+++ b/crates/perry-runtime/src/map.rs
@@ -132,7 +132,7 @@ impl MapSideAllocation {
         Self {
             entries,
             capacity,
-            numeric_index: Box::new(crate::fast_hash::new_ptr_hash_map()),
+            numeric_index: Box::new(NumericIndex::new()),
         }
     }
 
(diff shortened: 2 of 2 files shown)


proggeramlug:
Landed on `main` via the #8870 batch.

PR #8872, opened 2026-08-26
perf: remove cross-module ECS dispatch and argument-bundle overhead

proggeramlug:
## Summary

- inline producer-proved cross-module free-function graphs instead of losing the ordinary inline plan at module boundaries
- reuse resolved array headers across indexed stores and keep dynamic canonical read keys on a smaller read-only path
- keep synthetic-arguments methods on guarded direct dispatch
- add a producer-authored `$arguments_length` direct ABI for methods proved to observe only `arguments.length`, avoiding allocation/fill/mark of a temporary arguments object while preserving the public runtime ABI

Every specialization fails closed: user rest parameters, arguments identity/index/mixed uses, nested captures, override guard misses, and incompatible subclass arms retain the existing dynamic/materialized paths. The imported capability is producer-authored and participates in the object-cache key.

## ECS benchmark

Mac mini, `@codehz/ecs` comprehensive row `5k entities: 3 commands each + sync (15k total commands)`:

- full stack versus the landed #8867 baseline: **19.5493 ms -> 12.3269 ms** (**36.95% faster**)
- final scalar-arguments commit, 11 alternating pairs at repeat=64: **14.4167 ms -> 12.3269 ms** median (**14.64% faster**), candidate won **11/11**, semantic oracle **22/22**
- rebased current-main absolute comparison, 11 alternating pairs: Node 26.5.1 **2.9215 ms**, Perry **12.1930 ms**, Perry/Node **4.17x**, semantic oracle **22/22**

Current-main benchmark binary SHA-256: `a9b7100259ea2080def313dd2fa75a438ce3dcfcdd3a7c4f9a6983ee3c3646d3`.

## Validation

- `cargo test -p perry-codegen --lib --quiet` — 1,274 passed, 1 ignored
- `cargo test -p perry --bin perry --quiet` — 1,055 passed
- existing `@codehz/ecs` compiled suite — 1 passed, 0 failed, 6 intentionally skipped by the Bun shim
- synthetic-arguments runtime oracle covers 0/3 arguments, own regular-function override, delete/restore of the prototype method, and `Object.defineProperty` override
- cross-module IR regression proves direct length-only calls contain no `js_array_alloc`, `js_array_push_f64`, or `js_array_mark_arguments_object`; materialized arguments/rest regression coverage remains intact

## Summary by CodeRabbit

- **Performance**
  - Reduced temporary object creation for eligible `arguments.length` usage.
  - Optimized dynamic array access and writes.
  - Accelerated captureless `Array.prototype.some` callbacks.
  - Expanded cross-module inlining, record-return optimization, and small allocation-heavy method optimization.

- **Bug Fixes**
  - Improved dynamic array key and write correctness.
  - Preserved generic key/value types when destructuring Maps in class methods.

- **Tests**
  - Added regression coverage for argument counts, arrays, callbacks, inlining, and record-return optimization.


proggeramlug:
Added `07e747dc6` (`perf(array): call captureless some callbacks directly`).

Mac mini qualification on the ECS comprehensive 15k-command row, two independent 11-pair series at repeat=64:

- combined paired median: **0.551% faster**
- candidate wins: **19/22**
- cross-arm medians: **12.2862 ms -> 12.1863 ms** (**0.813%**)
- semantic oracle: **44/44**

Validation:

- perry-codegen: **1,276 passed**, 1 ignored
- perry-runtime: **2,708 passed**, 4 ignored
- targeted captureless/Buffer fallback tests: **2 passed**
- `cargo fmt --all -- --check` and `git diff --check` clean

The proof is fail-closed: only inline capture-free synchronous arrows with <=3 ordinary parameters use the direct callback body; lexical `this`, `new.target`, captures, rest/arguments, typed-array, and Buffer cases retain the generic closure path.

proggeramlug:
Correction (benchmark harness audit): PERRY_ECS_BENCH_REPEATS expects a JSON map keyed by label. The first version of this comment passed a scalar and therefore measured repeat=1 despite labeling it repeat=64. I reran every new comparison with the exact JSON row checked for repeat=64. The optimization conclusion is unchanged; the corrected steady-state Node target is lower.

Added two independently tested follow-up commits:

- d586c151a perf(method): inline bounded tiny allocation kernels
  - Tiny instance methods only (<=2 HIR statements), all-or-none cap of 8 allocation sites/module.
  - Fixes method codegen dropping the existing alloc-hot classification.
  - Earlier qualified ECS Mac mini screen versus the prior full-stack control: 3/3 wins, 2.33% median paired improvement; 6/6 semantic oracles.
- a8bc12306 perf(inline): optimize functions inside candidate methods
  - Keeps method-to-method inlining disabled in candidate method bodies for recursion safety, but allows the ordinary bounded standalone-function inliner.
  - This exposes returned anonymous records to the existing aggregate scalar-replacement pass. In the ECS hot path, ComponentEntityStore.exists no longer calls getDetailedIdType or materializes its returned record.
  - Corrected true repeat=64 Mac mini qualification against the exact method-hot predecessor (SHA-256 bb91a6f3f685aa66f4bfaa137a6d090ed4f3302f446adb7703e40307be328504): 11/11 wins, 10.73% median paired improvement, 10.74% cross-arm median reduction (11.9365 ms -> 10.6550 ms), 22/22 semantic oracles.

Qualified candidate:
- SHA-256 4b872f7edfa70984a1320187a5998e1e95d0f3e415dd2ba040028b38f402148b
- Corrected direct Node 26.5.1 true repeat=64 remeasurement: Perry 10.6515 ms median vs Node 1.7641 ms median (6.04x remaining gap), 11 alternating pairs, 22/22 semantic oracles.

Correctness:
- cargo test -p perry-transform --lib: 105 passed
- focused codegen allocation tests pass; full perry-codegen suite for the unchanged method patch: 1279 passed, 1 ignored
- cargo fmt and git diff --check clean

proggeramlug:
New qualified fix: 00b9c47f4 (perf(for-of): preserve Map entry types in function bodies).

Root cause: the function/method-body for-of lowerer pre-defined destructured Map bindings as Any even when the iterable was Map<K,V>. In CommandBuffer.execute this erased commands: Command[] and forced commands.some(...) through PropertyGet -> native dynamic dispatch. The module-init lowerer already preserved K/V; this ports the same generic-type propagation to function and method bodies, including Map<K,V> | undefined.

Evidence:
- Regression proves entityId/commands retain K/V and commands.some lowers to ArraySome.
- cargo test -p perry-hir --lib: 340 passed, 1 ignored.
- Post-transform ECS HIR: commands has the full Command[] union and the hot predicate is ArraySome; dynamic method dispatch is absent at that site.
- Mac mini, exact repeat JSON mapping with output repeat=64, 11 alternating predecessor/candidate pairs: 10.6459214401 ms -> 9.7659353281 ms median; 8.26594595% cross-arm and median paired improvement; 11/11 wins; control MAD 0.0066158854 ms; candidate MAD 0.0106868464 ms; 22/22 semantic oracles.
- Binary SHA-256: predecessor 4b872f7edfa70984a1320187a5998e1e95d0f3e415dd2ba040028b38f402148b; candidate 7015d996f0c97406ead3c966808cf8fa9dd302872042649582aee575a8fcf8e3.

proggeramlug:
New qualified runtime fix: a95778424 (perf(array): trust validated rooted iterator headers).

Root cause: RootedIterArray is private and is constructed only after normalize_array_receiver has produced a non-null genuine Array and Buffer/TypedArray dispatch has exited. Its arr() accessor nevertheless re-entered clean_arr_ptr for every element read and again for the callback receiver argument. On the ECS some predicates this repeated allocator/registry ownership classification twice per dense callback iteration.

Fix: for the ordinary live-array case, read the already-proved GC header directly. Moving GC rewrites the RuntimeHandle slot. Array growth is the exceptional case that leaves an alias on a forwarding stub; forwarded/non-Array headers still take the complete clean_arr_ptr resolver, preserving chain validation/compression and corruption defenses.

Evidence:
- Focused forwarding/shrink test passes, along with captureless some semantics and Buffer fallback tests.
- cargo test -p perry-runtime --lib: 2708 passed, 4 ignored, 0 failed.
- Mac mini, exact repeat JSON mapping with output repeat=64, 11 alternating predecessor/candidate pairs: 9.7731687318 ms -> 9.4359917526 ms median; 3.45002720% cross-arm improvement; 3.46128796% median paired improvement; 11/11 wins; control MAD 0.0061340130 ms; candidate MAD 0.0096686198 ms; 22/22 semantic oracles.
- Candidate SHA-256: 7162a6afb0917a222c690c62101ac141ca05e39402707ba36986b819c4fc490c.

proggeramlug:
Added 36ff6032a (perf(array): establish element shape proofs on demand).

Release-faithful sampling showed the element-shape store funnel was eagerly creating and maintaining a TLS side-table proof for every homogeneous object array, even when no generated shape-loop consumer existed. The fix makes proof establishment demand-driven through the existing ensure_element_shape preheader path; once requested, the existing keep/revoke/GC-transfer invariants remain unchanged.

Validation:
- cargo test -p perry-runtime --lib: 2708 passed, 4 ignored, 0 failed
- cargo test -p perry-codegen --lib element_shape: 32 passed
- Mac mini, exact committed binary git:36ff6032a, 11 alternated pairs, repeat 64: control median 9.541248 ms, candidate median 8.912717 ms
- Median paired improvement: 6.5502%; candidate won 11/11 pairs; 22/22 semantic oracles passed
- Control SHA: 7162a6afb0917a222c690c62101ac141ca05e39402707ba36986b819c4fc490c
- Candidate SHA: e04fcd9f61f26087201c5d96b8ebcadb13b7586877d947d52eca84aaa2418974

proggeramlug:
Added `f8a5fa06d` (`perf(array): reuse dynamic all-pointer append proofs`).

This addresses the hot ECS `CommandBuffer.execute` bucket append without ECS-specific source assumptions:
- an empty array's first pointer append establishes the runtime all-pointer layout invariant;
- dynamically typed inline appends consume the live header proof and exact `POINTER_TAG`;
- on the admitted steady-state arm, Perry skips `js_string_addref_if_heap_string`, `js_gc_note_slot_layout`, and `js_array_note_numeric_write`;
- numeric-layout invalidation, element-shape maintenance, and all proof misses retain the original bookkeeping path; the generation-tested write barrier remains.

Mac mini (`perry-macos.local`), isolated ECS row, alternating order, repeat=64, 15 pairs:
- control SHA: `e04fcd9f61f26087201c5d96b8ebcadb13b7586877d947d52eca84aaa2418974`
- candidate SHA: `798262aaa92d28b8ab860a4bfe5eab2626c6717f1a9f4bd2167a180cc14a9433`
- control median: **8.9819 ms**
- candidate median: **8.6053 ms**
- median paired improvement: **4.0186%**
- candidate wins: **15/15**
- semantic oracles: **30/30 passed**

Validation:
- `cargo test -p perry-codegen --lib`: 1280 passed, 1 ignored
- `cargo test -p perry-runtime --lib`: 2710 passed, 4 ignored
- compiled ECS fixture: 1 passed, 6 skipped
- focused all-pointer GC suite: 13 passed
- focused array-push IR suite: 7 passed

proggeramlug:
Added 618b1fa2c (perf(property): inline dynamic collection size reads).

Cause: dynamically typed or nested collection reads such as this.ctx.hooks.size reached the object-only property PIC. A live Set/Map cannot hit that PIC, so every read called js_object_get_field_ic_miss and repeated the full receiver-classification/property ladder. The new path uses exact runtime GC_TYPE_MAP / GC_TYPE_SET checks and loads the shared leading u32 size field inline. It does not trust erased TypeScript annotations; non-collection receivers retain the existing PIC and semantic fallback.

Correctness:
- focused dynamic Set / Map / ordinary-object executable: 2 / 1 / 9 as expected
- cargo test -p perry-codegen --lib: 1282 passed, 1 ignored, 0 failed
- ECS semantic runner: 30/30 confirmation oracles passed

Mac mini M3, taskpolicy -t 0 -l 0, alternating order, 15 pairs x repeat64:
- control SHA: 798262aaa92d28b8ab860a4bfe5eab2626c6717f1a9f4bd2167a180cc14a9433
- candidate SHA: f85f0d3ad517077e9a4069682163eda81111dfb35596d424504a2cc600da4c5f
- control median: 8.610136828 ms
- candidate median: 7.653369794 ms
- median paired improvement: 11.203283%
- candidate wins: 15/15
- control MAD: 0.009638451 ms
- candidate MAD: 0.005232091 ms

Result JSON: /Users/perry/perry-m3-array-length-bench.weMJ0r/collection-size-confirm-15pairs.json

proggeramlug:
Mac-mini confirmation for b2d7a01f1 (`perf(compare): inline exact three-byte literal equality`):

Root cause: strict equality against a three-byte cross-module heap literal already guarded the tag/address/length and compared the first and last bytes, but then called `js_string_equals` (and ultimately `memcmp`) solely to check the remaining middle byte. The lowering now checks that byte inline; literals of 4+ bytes retain the full helper fallback.

Verification:
- focused positive IR test: three-byte literal emits `streqlit.bm` and no `js_string_equals` call
- focused negative IR test: longer literal retains `streqlit.slow` and `js_string_equals`
- full `cargo test -p perry-codegen --lib`: 1283 passed, 1 ignored, 0 failed
- unchanged ECS executable oracle: 1 passed, 0 failed, 6 skipped
- Mac mini, taskpolicy-pinned, repeat=64, 6 measured rounds/process, 15 alternating pairs:
  - control SHA: f85f0d3ad517077e9a4069682163eda81111dfb35596d424504a2cc600da4c5f
  - candidate SHA: f5b487b7cec48b86de4886904742f940bb955c3829f5f5843a0cd3adc94780b9
  - control median: 7.653085612 ms
  - candidate median: 7.286808703 ms
  - median paired improvement: 4.779811%
  - wins: 15/15
  - semantic oracles: 30/30
  - control MAD: 0.003290477 ms
  - candidate MAD: 0.004569010 ms

Confirmation artifact: `/Users/perry/perry-m3-array-length-bench.weMJ0r/streqlit3-confirm-15pairs.json`.

proggeramlug:
Landed on `main` via the #8878 batch.

PR #8875, opened 2026-08-27
perf(descriptors): index descriptors by owner instead of scanning every entry

proggeramlug:
Three hot paths answered *"what does THIS owner have?"* by walking **every descriptor in the process** and filtering on the owner address.

`DescriptorTables` now carries `attr_keys_by_owner` / `accessor_keys_by_owner` mirroring the two `(owner, key)` maps, so each of those becomes a lookup.

### Where the scans were

| site | ran on |
|---|---|
| `js_object_keys` array branch, **twice** (`enumeration.rs`) | every `Object.keys(array)` |
| `accessor_descriptor_keys_for_obj` | the own-keys path |
| `transfer_descriptor_owner` | every `ArrayHeader` growth |
| `scan_descriptor_roots_mut` | **every GC cycle** |

The last one matters most: since the moving young-gen scavenge became the default (#7019), GC runs often, so this was a per-collection tax proportional to the whole program's descriptor count rather than to what actually moved.

Profiling `claude -p` put **46.6% of main-thread samples** in shapes/descriptors, with a HashMap `Keys` iteration the single hottest self-time entry by **4×** over anything else.

### Measured

`Object.keys(array)` × 20 000, varying the descriptor count on **unrelated** objects. Quiet machine (load 1.6), best of 6 in-process rounds, 15 process runs. `min` is the steady-state estimate — GC pauses only ever add time.

| ballast | node | before (min/med) | after (min/med) | speedup |
|---:|---:|---:|---:|---:|
| 0 | 1 | 11 / 31 | **8 / 8** | 1.4× |
| 1 000 | 1 | 15 / 45 | **7 / 8** | 2.1× |
| 4 000 | 1 | 21 / 88 | **8 / 8** | 2.6× |
| 16 000 | 0 | 62 / 226 | **7 / 8** | **8.9×** |

Before scales with descriptors on objects it never touches; after is flat, like node. The speedup keeps growing with descriptor count.

Note also `min ≈ median` after (7 vs 8) where before it was 62 vs 226 — the scan was interacting with GC, and removing it removed that variance.

### Correctness bug found by the new tests

`transfer_descriptor_owner` moved descriptors to the new address but never carried the per-object Bloom summary (`attr_key_bits` / `accessor_key_bits`). A freshly grown array has a null `meta`, for which `owner_may_have_descriptor_entries` answers `false` **authoritatively** — so after an array grew, `Object.keys` and `getOwnPropertyDescriptor` silently lost every accessor it had.

This was equally broken before this change: the gate sat in front of the old scan, so the scan simply never ran for the new owner. Fixed here, with a test pinning it.

### Tests

The index is a mirror, so its failure mode is **silent drift**, not a crash — a dropped key would quietly corrupt `Object.keys` output. The tests therefore assert the mirror invariant directly (index == what a full scan would return) across install, redefine, delete, bulk-clear and owner transfer, rather than spot-checking behaviour.

`cargo test -p perry-runtime --lib` — **2710 passed, 0 failed**.

## Summary by CodeRabbit

  * Improved object property enumeration and descriptor handling, especially for objects with many properties.
  * Reduced unnecessary scanning during property checks, descriptor transfers, cleanup, and garbage collection.

* **Bug Fixes**
  * Fixed an issue where accessors could become unavailable after an array grew or changed ownership.

* **Reliability**
  * Added safeguards and coverage to keep property descriptor tracking consistent during installation, updates, deletion, transfers, and clearing.

### DIFF
--- a/changelog.d/8875-descriptor-owner-index.md
+++ b/changelog.d/8875-descriptor-owner-index.md
@@ -1,0 +1,56 @@
+Sped up `Object.keys` / `for…in` on arrays, and every GC cycle, by indexing
+property descriptors by their owner instead of scanning the whole table.
+
+Descriptors live in two process-global maps keyed by `(owner_address, key)`.
+That shape answers "does owner X have key K?" in one lookup, but it cannot
+answer "does owner X have **any** descriptors?" — that is a question about a
+group of entries, and the map is only indexed by the full pair. So four call
+sites answered it the only way that shape allows, by walking every entry and
+filtering on the owner:
+
+```rust
+property_descriptors.keys().any(|(ptr, _)| *ptr == owner)
+```
+
+The cost of enumerating one small array therefore grew with how many
+descriptors *every other object in the program* held. The sites:
+
+* `js_object_keys`' array branch, twice — per enumeration, just to decide
+  whether a per-index `enumerable` check was needed at all;
+* `accessor_descriptor_keys_for_obj`, on the own-keys path;
+* `transfer_descriptor_owner`, on every `ArrayHeader` growth;
+* `scan_descriptor_roots_mut`, on **every GC cycle** — so since the moving
+  young-gen scavenge became the default (#7019) this was a per-collection tax
+  proportional to the program's total descriptor count rather than to what
+  actually moved.
+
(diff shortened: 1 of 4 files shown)


proggeramlug:
Landed on `main` via the #8878 batch.

PR #8858, opened 2026-08-26
Fix remaining Array Test262 clusters

proggeramlug:
## Summary
- finish the remaining Array constructor, `Array.of`, generic `flat`, inherited-index, and unsigned-length semantics
- make observable `push`/`unshift` use descriptor-aware Set behavior while keeping internal result-array appends own-property based
- preserve array descriptors across growth, revoke stale Array specialization after object reassignment, and restore default `Object.prototype` dynamic method calls
- add end-to-end and focused regression coverage; no version bump

## Testing
- pinned Test262 `built-ins/Array` (`4249661388e5d3f92a85186213da140a6481490f`): **2527 pass, 0 diff, 0 runtime-fail, 0 compile-fail, 0 skip**
- `cargo test --release -p perry --test issue_5898_array_remaining -- --nocapture`
- `cargo check -p perry-runtime -p perry-hir -p perry-codegen -p perry`
- targeted HIR type-widening, array descriptor-growth, and codegen push-guard unit tests
- touched-module `rustfmt --check`, `git diff --check`, and `scripts/check_file_size.sh`

Fixes #5898

## Summary by CodeRabbit

* **New Features**
  * Implemented functional `Array.prototype.flat`, including generic array-like receivers and optional depth.
  * Improved support for inherited properties, accessors, and borrowed array methods.

* **Bug Fixes**
  * Corrected `push`, `unshift`, and related operations for exotic arrays and property descriptors.
  * Prevented incorrect mutations involving non-writable or inherited properties.
  * Improved handling of large array lengths and maximum safe integer limits.
  * Preserved array behavior during garbage collection and array growth.

* **Tests**
  * Added comprehensive coverage for array construction, flattening, mutation, inheritance, accessors, and boundary conditions.

Issue #8159, opened 2026-08-15
perf(codegen): #8084's callee rooting costs pipeline 3.9%, consuming #8157's win on that row


proggeramlug:
## Summary

`pipeline` is the only corpus row that got **slower** across 2026-08-15's twenty merges: **+0.28% instructions** on `main` @ `d6d7d0efe` versus `83b6b8c69`. Every other row is flat or faster, several enormously.

The number is small but it is **2× the measured noise floor** (±0.15%, established by compiling `main` twice into separate cache/out dirs and confirming all 22 binaries byte-identical), and the run ranges do not overlap: base 3.7881–3.7936 G, main 3.8015–3.8053 G. Peak RSS is unchanged.

## It is a give-back, not a shortfall

Building the intervening merge points attributes it to a single step:

| arm | `pipeline` instructions | Δ |
|---|---:|---:|
| `83b6b8c69` (session start) | 3.7881 G | — |
| `8b1b4b909` (#8157 merge) | 3.6547 G | **−3.52%** |
| `53d63aad2` (#8084 merge) | 3.8003 G | **+3.95%** |
| `d6d7d0efe` (today's main) | 3.7988 G | −0.00% |

#8157 delivered its full claimed −3.52% here. The next step gave all of it back and a little more.

Every other row moves by ≤0.03% across that same `at8157 → at8084` step, so this is one localized change, not drift.

## Attribution

The window is {#8158, #8084}. **#8158 is a 7-line addition to `BUILD_CACHE_ENV_VARS` and cannot change emitted code**, which leaves **#8084** — specifically `fix(codegen): root the callee across argument evaluation in three call arms`, touching `new_dynamic.rs`, `call_spread.rs` and `early_branches.rs`. `pipeline` is spread- and call-heavy, so the mechanism fits: each rooted callee adds an `open_rooted_group` / `adopt` / `reread` / `release` sequence around a call site.

## This is not a request to revert

#8084 fixes a real moving-GC rooting bug — a callee held in a bare register across argument evaluation is a pre-move address, and the dependency-scale corpus found 16 sites of that shape. Correctness wins over 3.9% on one row, and #8084 should stay.

Filing it because **nobody wrote the cost down**. A correctness fix that silently consumes an entire optimisation's win on a row is exactly the thing that later gets rediscovered as an unexplained regression, and #8157's `pipeline` win currently looks like it never happened.

## Worth looking at

[...]

PR #8876, opened 2026-08-27
perf(ecs): guarded store follows forwarding edge, inline typeof/typed-array/subclass fast paths (wolf-ecs -16.5% / -20.9%)

proggeramlug:
## Summary

Continuation of the wolf-ecs (noctjs/ecs-benchmark) performance campaign. All measurements on the Mac mini under `taskpolicy -t 0 -l 0`, 11 alternating process pairs, retained only when BOTH benchmarks improved; the default-defer/subclass semantics probe is byte-identical to Node on every step. Full record: `secret-tests/ECS_PERFORMANCE_HANDOFF_2026-08-27.md`.

| Step | Change | add/remove | entity-cycle |
|---|---|---:|---:|
| v77 | guarded property-receiver **store** tier follows one growth-forwarding edge inline (the read tier already did); `this._ent`/`this._updateTo`/`SparseSet.sparse` held pre-grow forwarding stubs forever and every store went out of line | **-9.15%** (11/11) | **-13.67%** (11/11) |
| v78 | inline raw-f64 gate for `js_array_note_numeric_write` + cheaper runtime note + typed-array pre-dispatch in `js_array_get_f64` | **-4.86%** (11/11) | **-5.50%** (11/11) |
| v80 | exact inline `typeof x === "number"` (33-kind differential probe matches Node) + inline dynamic typed-array read branded off the `GC_TYPE_TYPED_ARRAY` header instead of the evictable 64-slot kind cache | **-1.28%** (11/11) | **-0.73%** (11/11) |
| v81 | `GC_TYPE_OBJECT` receivers go to `array_subclass_fast_index_get_raw` before `clean_arr_ptr` | **-2.03%** (11/11) | **-2.39%** (11/11) |
| v83 | declared-array receiver + `Any` key (`packed[sparse[x]]`): inline integer-index test, then the inline numeric tiers | **-2.37%** (11/11) | **-2.89%** (11/11) |
| merge of main (#8878) + v87 | the merge alone was a 2.2× regression (main's canonical-i32 read split committed object-backed `Archetype` receivers to the guarded plain tier, whose feedback fallback stringified every index into a by-name lookup); the typed-feedback array-read fallback now answers the dense Array-subclass read first | **+0.54%** vs pre-merge (noise) | **-1.21%** vs pre-merge |

Cumulative vs v74: add/remove 0.5562 → **0.4531 ms/op (-18.5%, ≈3.39× Node)**, entity-cycle 0.4988 → **0.3836 ms/op (-23.1%, ≈2.57× Node)** at v83; parity re-established after the merge.

Rejected and not included: v75/v76 (probe hoist in `try_strict_dense_index_set`, +1.3%/+1.9% — the receiver was a forwarded stub, which is what v77 fixes), v79 (same-identity skip of the element-shape note, +2.2%/+2.3% — object-backed Array-subclass elements change exact shape on every push/pop) and v82(b) (a wider `$pshape` clone budget that never fired).

## Review follow-ups (`1dd37f9c0`)

- `js_array_push_u31_with_length` stays allocate-but-never-reenter: it declines (null) every receiver whose push can run user code and the generated caller performs the complete guarded push in `apush.u31.generic`.
- `js_packed_ecs_u32_loop_guard` declines a component column shorter than the admitted bound.
- `object_hot_for_owner` validates its cached table pointer against the current thread's `RuntimeState`.
- `cache_carrier` has a lifecycle: noted after a successful insert, recomputed from live cache occupancy after every full trace.
- `js_object_get_symbol_then_field_ic_miss` declared with the runtime's pointer parameter; minors (parentheses, unique test class ids, `visited` key, exhaustive `UnaryOp`, changelog restated as shipped behavior).

On `strict_eq_reuses_a_non_pointer_left_operand_across_an_allocating_right_operand`: not pre-existing — correct. It failed because the accumulated commit lowers `provenNumber === {…}` inline (`lower_strict_eq_against_number`: INT32 immediates normalized, every other non-number tag is a NaN so `fcmp oeq` answers `false` exactly as `js_eq` did) and the test looked for a `js_eq` call that no longer exists. The updated test (`0c00774bd`) keeps the original claim on the `fcmp` operands — the non-pointer left operand is the register produced above the allocating right operand, not a rooted re-read — and additionally pins that no runtime equality call remains. No rooting was weakened: a non-pointer needs no root, and the pointer-bearing operand is the freshly allocated right value.

## Test plan

- [x] focused IR tests for each codegen change (`index_set_barrier_tests`, `compare_tests`, `index_get_claim_tests`, `index_method_clone_tests`), barrier census/sabotage suite green
- [x] `cargo test -p perry-codegen --lib` 1322 pass; `cargo test -p perry-runtime --lib -- array:: object:: typed_feedback:: gc:: --test-threads=1` 1498 pass
- [x] workspace `cargo check --all-targets` under `-D warnings` with the host-compatible exclusions; every lint audit (file size, addr-class, raw-handle incl. `--no-raise-vs` merge base, shape census, GC store-site inventory, local-binding audit); changelog fragment
- [x] semantics probe identical to Node locally and on the Mac mini for every step
- [ ] CI (`pr-gate`) — runs on this PR keep getting cancelled by a local CI-cancel loop from another session on the dev box; rerun pending


proggeramlug:
Added v83 (`4bf2c84a7`): declared-array receiver + `Any` key reads (`packed[sparse[x]]`) get an inline integer-index test and then the inline numeric tiers. Mac mini, 11 pairs vs v81: add/remove **-2.37%** (11/11), entity-cycle **-2.89%** (11/11). Cumulative vs v74: add/remove **-18.5%** (0.4531 ms/op, 3.39× Node), entity-cycle **-23.1%** (0.3836 ms/op, 2.57× Node). Semantics probe unchanged.

proggeramlug:
Merged current main (`7898bf9f4`) and made the branch pass the PR-tier gates locally (draft PRs don't run CI here):

- conflicts resolved against #8857/#8859 (main moved the `header.rs` GC slot helpers and the ic_miss length tests; both `js_array_push_u31_with_length` and `js_array_push_f64_spec` kept; u31 push now honours #8858's exotic-receiver rule via the spec path)
- 2,000-line file gate: six inherited oversized files split into `use super::*` child modules
- GC store-site inventory: 8 markers added; `-D warnings`: inherited unused imports removed; the stale `strict_eq_reuses…` rooting test updated to the inline `fcmp` lowering
- local verification: workspace `cargo check --all-targets` under `RUSTFLAGS=-D warnings` with CI's host-compatible exclusions ✅, perry-codegen 1309/1309 ✅, perry-runtime 2732/2732 (single-threaded) ✅, file-size / GC-inventory / binding / architecture audits ✅, semantics + 33-kind `typeof` probes byte-identical to Node ✅, wolf-ecs Mac mini parity vs the pre-merge build +0.06% / +0.07% (noise) ✅

Known pre-existing gap (same on main, not touched here): `Array.prototype.push` on a `preventExtensions` array returns silently instead of throwing `TypeError` (`js_array_push_f64` returns for `SEALED|NO_EXTEND`).

proggeramlug:
Holding this one — two blockers, and the first contradicts a stated premise.

## 1. `strict_eq_reuses_…` is **not** pre-existing

The caveat says the failure is pre-existing ("fails with the four new commits stashed too"). It passes on both endpoints I can test:

```
current main (59a4c9aa5)          test result: ok. 1 passed; 0 failed
this PR's own base (cb9e96708)    test result: ok. 1 passed; 0 failed
```

So `expr::compare_tests::strict_eq_reuses_a_non_pointer_left_operand_across_an_allocating_right_operand` is green on the merge-base *and* on today's tip. The failure therefore originates in this PR's own commits — and since you observed it with the four measured commits stashed, the accumulated-codex commit (v74) is the likely source rather than v77/v78/v80/v81.

That matters more than a normal red test, because of what this one asserts: a **non-pointer left operand being reused across an allocating right operand**. That is the #7773 / #8159 family — a value held across a collection point. Twice today a "stale test" on a perf PR turned out to be a real weakening rather than a stale expectation (#8833's `$pshape_args` route dropped its runtime guard; #8858's verifier silently stopped requiring `raw_f64_layout` facts for a still-emitted helper). Both would have been silenced by updating the assertion. Worth establishing which this is before landing.

## 2. Conflicts with current `main` (11 files)

The branch is based on `cb9e96708`, roughly twenty commits behind. It no longer merges:

```
crates/perry-codegen/src/codegen/method.rs
crates/perry-codegen/src/expr/index_get_claim_tests.rs
crates/perry-codegen/src/expr/property_get.rs
crates/perry-codegen/src/expr/property_get/generic_dispatch.rs
crates/perry-codegen/src/lower_call/method_override.rs
crates/perry-runtime/src/array/element_shape.rs
… (11 total)
```

Those overlap what landed today in the same areas — #8872 (cross-module ECS dispatch, argument bundles), #8875 (descriptor owner index), #8867 (Map dense numeric ranges), #8858 (Array Test262). A rebase is needed regardless, and it may also change the measured numbers, since several of those PRs touch the same hot paths this campaign is optimising.

## Not blocking, just noted

The description still reads "**Draft.**" and the test plan has an open item (`[ ] fix the pre-existing strict_eq… failure`), but the PR is marked ready. Worth reconciling.

The measurement discipline here is good — 11 alternating process pairs under `taskpolicy`, retained only when both benchmarks improved, with v75/v76 and v79 explicitly rejected and *why* recorded. The v77 finding in particular (receivers holding pre-grow forwarding stubs forever, so every store went out of line) is a nice catch. None of that is in question; it is the two items above.

proggeramlug:
Thanks — both blockers addressed, plus the coderabbit majors, in `1dd37f9c0` (branch merged with current main twice; conflicts gone, `MERGEABLE`).

**1. `strict_eq_reuses_…` — you're right that it isn't pre-existing.** It originates in the accumulated commit, but as a stale expectation rather than a weakening: that commit lowers `provenNumber === {…}` inline (`compare.rs::lower_strict_eq_against_number` — INT32 immediates are normalized to doubles first, every other non-Number tag is a NaN, so `fcmp oeq` answers `false` exactly as `js_eq` did, and strict equality never coerces), so the `js_eq` call the test located no longer exists. The test as updated in `0c00774bd` keeps the original claim on the `fcmp` operands — the non-pointer left operand is the register produced above the allocating right operand, not a re-read from a root — and additionally pins that no runtime equality call remains. Nothing pointer-bearing is held across the allocation (the only pointer is the freshly allocated right operand). Description corrected accordingly.

**2. Conflicts** — merged main at `7898bf9f4` and again at `9a09adf59` (after #8878 landed); the second merge exposed a 2.2× regression from #8872's canonical-i32 read split on object-backed `Archetype` receivers, fixed in `a34c580ce` (typed-feedback fallback answers the dense subclass read; Mac parity vs pre-merge +0.5% / -1.2%).

**coderabbit majors** (all in `1dd37f9c0`):
- `js_array_push_u31_with_length` re-entrancy: agreed. The entry now answers **null** for every receiver whose push can run user code (exotic arrays → spec `Set`, unresolvable → Proxy traps) instead of calling those paths itself, and the generated caller takes the complete guarded push in a new `apush.u31.generic` block. The hot path is unchanged (one call), so the `AllocNoReentry` classification is now true by construction. Runtime test declines a typed-array receiver; the composed-clone IR test pins the hot/fallback split.
- `js_packed_ecs_u32_loop_guard`: declines when a column is shorter than the admitted bound the receiver guard published in `out[6]` (covers the live-length form too). Test added.
- `object_hot_for_owner`: the cached pointer is compared with the current thread's `state().object_hot` on every use; a mismatch is a miss and is overwritten.
- `cache_carrier` lifecycle: `note_cache_carrier` now runs only after an entry naming the pair was actually inserted (early return when both inserts fail), and `recompute_cache_carriers_after_full_trace` clears every bit and re-notes the descriptors of live entries in both tables, called beside `rotate_old_carrier_epoch_after_full_trace`. Between full traces the bit set is a superset of live occupancy (the direction the rooting gate needs). On the `!target_is_cache_carried` short-circuit at the install site: it stays sound because the full trace itself notes old-generation carriers (`gc/layout_slot_visit.rs`), so a shape that leaves the cache while still carried by an old object is re-noted before the rotation that could shed it.
- `js_object_get_symbol_then_field_ic_miss`: declared `[DOUBLE, DOUBLE, PTR, I64, PTR, PTR]`, call site passes the key as a pointer.
- Minors: parentheses, unique class ids (`0x0074_8693`/`8694`), `visited` keyed by the terminal-`this` allowance, exhaustive `UnaryOp`, changelog restated as shipped behavior.

Local gates on `1dd37f9c0`: codegen 1322/1322, runtime array/object/typed_feedback/gc 1498/1498, `-D warnings` workspace check, all lint audits. CI on this PR has been getting cancelled by a local `gh run cancel` loop from another session on the dev box; rerun pending.

proggeramlug:
Landed on `main` via the #8887 batch.

proggeramlug:
Landed via #8887.

Thanks for chasing the `strict_eq_reuses_…` failure properly rather than editing the assertion. I verified the updated test independently: it still locates the `fcmp oeq double` line, extracts both operands and asserts `left_producer < right_producer` — the original claim intact — and additionally pins that no runtime equality call remains. Strictly stronger, and the rooting argument holds (a non-pointer needs no root; the pointer-bearing operand is the freshly allocated right value).

That makes three failing-test-on-a-perf-PR cases today. #8833 and #8858 turned out to be real weakenings the tests were catching; this one was a genuinely stale expectation — but established by evidence rather than assertion, which is the part that matters.

Validated on the merged tree: codegen 1322, runtime 2743, stdlib 124, hir 340, all 0 failed; 30/30 gates.

PR #8880, opened 2026-08-27
fix(stdin): restore the end/close arm on the listener provider path (#8861 regression)

proggeramlug:
Restored the `end`/`close` arm on the stdin listener **provider** path, which
was lost when #8861 was batch-landed. Without it `echo hi | claude -p "…"`
never completes.

#8861 fixed piped stdin for `-p` by teaching three layers about `end`
listeners. Two of them landed: the GC rooting of `STDIN_END_CALLBACKS`, and
`STDIN_PULL_MODE`. The third — the `"end" | "close"` arm in `stdin_on_op` —
did not, and it is the one that made the fix work.

`stdin_on_op` is the provider that the stdin object's native
`on` / `once` / `addListener` methods delegate to: every registration that is
**not** codegen's literal `process.stdin.x(…)` shape. That means an alias
(`const s = process.stdin; s.once("end", …)`) or stdin passed as a parameter
(`helper(process.stdin)`). Claude Code's print-mode reader is exactly the
parameter form — `X71(process.stdin, 3000)`, then `stream.once("end", …)` on
the parameter — so those registrations fell into `_ => return` and were
silently discarded. The `end` half of the reader's
`race(once("end"), timeout(3000))` could never win.

`stdin_off_op` gets the mirror arm. #8864 removed the `end`/`close` clause from
the removal path, so a provider-registered listener could be added but never
removed — `removeListener` / `off` leaked it and the pump kept a stale callback
pointer.

The two tests that guarded this — `provider_path_registers_end_listeners` and
`provider_path_removes_end_listeners` — were dropped alongside the arm, which is
why CI stayed green through the regression. Both are restored here. They assert
the provider entry points (`stdin_on_op` / `stdin_off_op`) **directly** rather
than going through the `js_readline_stdin_on` extern, because the extern path
kept working the whole time and therefore cannot catch this. Sabotage-checked:
deleting the arm again fails `provider_path_registers_end_listeners`.

Measured on the claude-code 2.1.112 bundle, `echo hello | cc -p "…"`:

| build | result |
|---|---|
| before this fix | 4/4 hang (120 s timeout, no output) |
| with the arm restored | completes in ~7 s |

---

### How this was found

Rebuilding the claude-code bundle to re-profile #8875, the sanity check hung 6/6 on `echo hello | cc -p`. Attribution matters here, so rather than assume, I built the branch base (`adf28bef5`, upstream only, none of my commits) in an isolated worktree: **it hangs 4/4 too**. #8875 is not implicated — the regression is upstream.

| binary | base | piped `-p` |
|---|---|---|
| `cc-ctl` | my #8861 branch (has the arm) | 7 s, completes |
| `cc-base` | `adf28bef5` upstream only | 4/4 hang |
| `cc-new` | `adf28bef5` + #8875 | 6/6 hang |

Diffing my #8861 branch against what landed showed the missing arm directly:

```
origin/main stdin_on_op:  "data" | "readable" | "keypress" | _ => return
my #8861 branch:          ...plus "end" | "close"
```

The API key in use is out of credits, so the success criterion is "completes and prints node's own `Credit balance is too low`" rather than a model response — a hang vs. a clean exit is still unambiguous.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where piped standard input could fail to complete when listening for end-of-stream events.
  * Restored handling for standard input `end` and `close` listeners, including aliased or passed references.
  * Prevented lingering listeners after they are removed.

* **Tests**
  * Added coverage for registering, triggering, and removing standard input end-of-stream listeners.

### DIFF
--- a/changelog.d/8879-stdin-end-provider-arm.md
+++ b/changelog.d/8879-stdin-end-provider-arm.md
@@ -1,0 +1,38 @@
+Restored the `end`/`close` arm on the stdin listener **provider** path, which
+was lost when #8861 was batch-landed. Without it `echo hi | claude -p "…"`
+never completes.
+
+#8861 fixed piped stdin for `-p` by teaching three layers about `end`
+listeners. Two of them landed: the GC rooting of `STDIN_END_CALLBACKS`, and
+`STDIN_PULL_MODE`. The third — the `"end" | "close"` arm in `stdin_on_op` —
+did not, and it is the one that made the fix work.
+
+`stdin_on_op` is the provider that the stdin object's native
+`on` / `once` / `addListener` methods delegate to: every registration that is
+**not** codegen's literal `process.stdin.x(…)` shape. That means an alias
+(`const s = process.stdin; s.once("end", …)`) or stdin passed as a parameter
+(`helper(process.stdin)`). Claude Code's print-mode reader is exactly the
+parameter form — `X71(process.stdin, 3000)`, then `stream.once("end", …)` on
+the parameter — so those registrations fell into `_ => return` and were
+silently discarded. The `end` half of the reader's
+`race(once("end"), timeout(3000))` could never win.
+
+`stdin_off_op` gets the mirror arm. #8864 removed the `end`/`close` clause from
+the removal path, so a provider-registered listener could be added but never
+removed — `removeListener` / `off` leaked it and the pump kept a stale callback
+pointer.
+
+The two tests that guarded this — `provider_path_registers_end_listeners` and
+`provider_path_removes_end_listeners` — were dropped alongside the arm, which is
+why CI stayed green through the regression. Both are restored here. They assert
+the provider entry points (`stdin_on_op` / `stdin_off_op`) **directly** rather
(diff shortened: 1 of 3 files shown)


proggeramlug:
Landed on `main` via #8881. Verified the gap on `main` first — `stdin_off_op` had only `data`/`readable` arms — and both new provider tests pass.

One correction for the record: this completes #8861's third layer rather than restoring something dropped in its landing. The ref I merged (`0bf52fbbf`) is byte-identical to #8861's head today, `stdin_on_op` in it has only `data`/`readable`/`keypress`/`_ => return` with no `end`/`close` arm, neither `provider_path_*` test appears in it, and the `mod_tests.rs` split preserved all 15 tests.

Where I was genuinely incomplete: landing #8864 I verified the `"end" | "close"` arm survived at `readline/mod.rs:1651` and called the interaction safe — but that is the syntactic extern's arm, not the provider's. Checking one of two paths and reporting on both is the real gap, and your tests asserting `stdin_on_op`/`stdin_off_op` directly close it properly.

proggeramlug:
Landed via #8881.

Issue #8125, opened 2026-08-15
perf(proxy): the #6595 store-plan gate costs one shape-table probe per allocated object


proggeramlug:
## Summary

`proxy.rs`'s #6595 store-plan gate calls `object::object_is_regular` **exactly
once per allocated object**. It is the single largest piece of #8113's residual
instruction cost, and reducing it is a #6595-adjacent design question rather
than a mechanical fix — which is why #8113 left it alone.

## The counts

Per-callsite counters (`#[track_caller]` + `libc::atexit`, on
`tls_hot.rs::maybe_install_stats_hook`'s pattern) over every shape-table entry
point, on the 19-program corpus:

| prog | calls at `object/mod.rs`'s `object_is_regular` probe | per object |
|---|---:|---:|
| `retain` | 3,000,000 | **1.00** |
| `retain1` | 1,000,000 | **1.00** |
| `retain_wide` (8 fields) | 3,000,000 | **1.00** |
| `retain_wide1` (8 fields) | 1,000,000 | **1.00** |
| `churn` | 20,000,002 | |
| `churn_read` | 1,000 | ~0 |
| `tree`, `push_cls` | 0 | |

One per object, **flat in object width** — 8-field literals cost the same as
2-field ones. Reads pay nothing.

Two further counter findings worth keeping:

* **`object_live_slot_count` is called ZERO times on every hot row.** #8113's
  new bound derivation is not on the measured path at all. Anyone reaching for a
  memo in front of it should check the call count first — two separate memo
  attempts measured null for exactly this reason.
* `deeplist` records **one** call at this site yet carries the largest
  percentage (+8.2%). Whatever dominates `deeplist` is **not** this. Unexplained;
  the GC-pacing hypothesis (its RSS fell 4.2%) is plausible and unestablished.

## Why it is not a mechanical fix

Before #8113 the gate was `(*obj).object_type == OBJECT_TYPE_REGULAR` — a free
`u32` compare on a header word that no longer exists. The replacement must stay
FALSE for a heap class object or #6595 reopens (bundled zod's `ZodX.create`
vanishing from ClassRef static dispatch). Three candidates, none a drop-in:

1. **Pass the already-read `GcHeader`.** The site has one in hand for its
   blocking-flags test; `object_is_regular` re-derives it through
   `try_read_gc_header` (handle-band, heap-range and slab classification, then
   the load). **Tried in #8122 and REVERTED**: semantically identical and
[...]

PR #8884, opened 2026-08-27
perf(shapes): drop SipHash from ids_by_facts, the last one on the shape path

proggeramlug:
Dropped SipHash from `ids_by_facts`, the last shape-table map still using it.

Profiling `claude -p` put `RandomState::hash_one` at 17 self-samples inside
`shapes::` alone (57 across the process) — pure hashing overhead on a lookup
that runs on every descriptor install and retire.

Its sibling maps already moved off SipHash (#8125). The standing comment on
this one argued only against `PtrHasher`, whose `write_*` methods OVERWRITE the
accumulator — right for a single-word key, wrong for this five-field one, which
would collapse to its last field and collide every descriptor sharing it. That
objection does not apply to `FastKeyHasher`: it implements only `write`, so the
derived `Hash`'s `write_u32` / `write_u64` calls all forward there and FOLD with
FNV-1a, reaching every field.

The key is internal shape state, never program input, so DoS-resistant hashing
buys nothing — the same rationale already applied to the descriptor side tables.

The new test pins the folding property directly: vary one field at a time and
require a distinct hash each time. Sabotage-checked against `PtrHasher`, where
it fails with "changing `keys` alone must change the hash".

---

Found while re-profiling `claude -p` after #8875. That fix removed the dominant descriptor scan, and the profile is now flat — no single hotspot, biggest self-time entry 69 samples — so what is left is a set of small, individually-cheap wins. This is one of them.

Suite: 2717 passed, 0 failed.

**Scope note:** I have not measured this one end-to-end on the bundle in isolation; a bundle rebuild is ~60 min and the change is worth less than measurement noise on its own. It is justified by the profile plus the fact that it makes this map consistent with every sibling, not by a wall-clock delta. Happy to fold it into a batch measurement if you would rather see numbers first.

## Summary by CodeRabbit

* **Performance**
  * Improved internal shape-state indexing with faster hashing, supporting more efficient runtime lookups.

* **Bug Fixes**
  * Ensured all shape metadata fields contribute to hash values, reducing avoidable collisions.

* **Tests**
  * Added coverage confirming distinct shape metadata produces distinct, deterministic hashes.

### DIFF
--- a/changelog.d/8882-shape-facts-hasher.md
+++ b/changelog.d/8882-shape-facts-hasher.md
@@ -1,0 +1,20 @@
+Dropped SipHash from `ids_by_facts`, the last shape-table map still using it.
+
+Profiling `claude -p` put `RandomState::hash_one` at 17 self-samples inside
+`shapes::` alone (57 across the process) — pure hashing overhead on a lookup
+that runs on every descriptor install and retire.
+
+Its sibling maps already moved off SipHash (#8125). The standing comment on
+this one argued only against `PtrHasher`, whose `write_*` methods OVERWRITE the
+accumulator — right for a single-word key, wrong for this five-field one, which
+would collapse to its last field and collide every descriptor sharing it. That
+objection does not apply to `FastKeyHasher`: it implements only `write`, so the
+derived `Hash`'s `write_u32` / `write_u64` calls all forward there and FOLD with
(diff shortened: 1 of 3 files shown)


proggeramlug:
Landed on `main` via the #8887 batch.

PR #8885, opened 2026-08-27
perf: ECS command path follow-up to #8872 — −40% on the codehz/ecs 5k-entity sync row

proggeramlug:
Follow-up to #8872 on the same target row: `codehz/ecs` "5k entities: 3 commands each + sync (15k total commands)". Every mechanism is general (no ECS-specific names); each step was screened with paired alternating runs on an idle Mac mini and confirmed with 15 pairs. Full write-up with root causes, profiles and the reproduce commands: `secret-tests/ecs-suite/PERRY_ECS_FOLLOWUP_2026-08-27_CLAUDE.md`.

**Result:** 7.296 ms/op (handoff control, #8872 merged) → 4.388 ms/op, **−40%**. Node 26.5.1 is 1.764 ms/op, so 4.1× → 2.5×.

| step | mechanism | paired median | wins |
|---|---|---:|---:|
| A | branded `number & {…}` intersections and generic/imported aliases lower to their primitive (were `Any`); inline plain-double fast paths for dynamic compares/truthiness | +0.9% | 7/7 |
| C | `Map.get` heals stale array-growth forwarding stubs; strict store lane follows one edge | +5.8% | 13/15 |
| D | `clean_arr_ptr` header-bit fast lane | +4.5% | 15/15 |
| E2/F2 | union dedupe; `js_array_length` lane; old→young inline-slot barrier shortcut; class key arrays under `ImmortalLayoutScope` | +0.9% / +6.1% | 14/15, 15/15 |
| H | beta-reduce called arrow-literal locals left by inlining (`closure_local_inline.rs`) | +4.8% | 15/15 |
| IJ/K/L | process-global layout address sketch; lazy inline-arena state; object/closure per-object-mask threshold 8 | +5.2% / +0.7% / +6.1% | 7/9, 8/9, 15/15 |
| N | death prune for the per-object layout tables + `PERRY_YOUNG_LAYOUT_RECORDS` gate on the inline forget probe | +3.0% | 15/15 |
| O | store lane drops redundant registry probes; dirty-page cache on hot TLS | +1.7% | 15/15 |
| P+Q | `pop`/`length =` resolve header flags once, named-props latch, length-0 rebuild shortcut; inline `STRING_TAG` test before the let-copy string demote | +6.5% | 15/15 |
| S+T | validated-parent write-barrier entry; inline closure-capture reads | +4.4% | 14/15 |
| U | `length = 0` keeps an all-pointer array all-pointer | +2.3% | 14/15 |

Kept-but-neutral: M (append-aware mask policy), R (`typeof` cache on hot TLS + dead-code fixups). Rejected (not in the branch): 16-way page-generation cache (−8.6%), closure mask exemption (broke 3 GC tests).

Soundness notes reviewers may want to check first:
- `PERRY_YOUNG_LAYOUT_RECORDS` counts records keyed by an address the bump allocator could recycle (ARENA header, not on a Longlived/Old page); conservative between collections, exact after every prune; a zero load proves the inline forget probe can be skipped.
- The strict store lane and `pop`/`set_length` drop registry probes only for `GC_TYPE_ARRAY` headers — every typed array / Buffer / native view registration carries its own object type.
- `js_write_barrier_slot_validated_parent` is only emitted behind `emit_parent_may_need_remembering_check`, which has already dereferenced the parent header; it decides identically to `js_write_barrier_slot` for every parent meeting that contract.
- `length = 0` keeps `SIDE_MASK | ALL_POINTERS` only for an array that already was; that is the state a declared `[]` literal starts in, and a non-pointer store still demotes it through `layout_note_slot`.

Tests: new unit tests for each mechanism (runtime 2722 green with `--test-threads=1`, codegen 1287, hir/transform suites green); the CI lint gates (file-size, addr-class, raw-handle and unrooted-local ratchets vs. merge base, rekeyed-table custody, gc store-site inventory) replayed locally.

## Summary by CodeRabbit

* **New Features**
  * Improved support for branded and generic type aliases across modules.
  * Added smarter optimization of local arrow-function calls.
* **Performance**
  * Faster dynamic number comparisons and truthiness checks.
  * Improved array, map, string, and garbage-collection performance.
  * Command-path benchmarks improved from 7.30 ms/op to approximately 4.7 ms/op.
* **Bug Fixes**
  * Improved handling of array layouts, numeric stores, forwarded arrays, and map entries.
  * Preserved correct array behavior after truncation, clearing, and property changes.
* **Tests**
  * Expanded coverage for type aliases, arrays, maps, closures, and garbage collection.


proggeramlug:
Landed on `main` via #8888, rebased onto the `main` that now contains #8876 and #8872. I resolved the six conflicts rather than asking you to rebase.

Two were genuine overlaps with #8876 and both **compose** — neither side discarded. In `arithmetic.rs` I kept #8876's `ValueTypeofTag` enum split and adopted your single `perry_thread_local!` array indexed by slot constants. In `compare.rs` the two cases turned out complementary: #8876's "exactly one operand statically proven numeric" runs first, then falls through to your inline `fcmp` for the dynamic case instead of the bare helper call.

Worth knowing for future rebases: hunk-level resolution restored five statics in `indexing.rs` that you had relocated to the new `indexing_support.rs`, producing duplicate `#[no_mangle]` symbols. File-level reorganisations don't survive per-hunk merging when the other side edits the same files.

On `noncanonical_bitwise_method_does_not_gain_raw_number_truthiness`: the assertion was over-broad, not your code unsound. `fcmp one` had one source when that test was written — the constructive-proof shortcut, genuinely unsound for an arbitrary `a & b`. Your dynamic-truthiness diamond adds a second, and emits it only inside `truthy.num`, after the bit test proving a plain untagged non-NaN double, with `truthy.slow` still calling `js_is_truthy` for everything else. `type_analysis/numeric/tests.rs:796` already encodes exactly that distinction, so I mirrored its helper. The test's claim is intact — it still forbids unguarded numeric truthiness, and its `@js_is_truthy` assertion is untouched.

One caveat carried into the PR body: your −40% was measured without #8876, and both campaigns touch the same hot paths. The composed tree's number is unknown and should be re-taken on the mini before the figure is repeated.

PR #8889, opened 2026-08-27
fix(error): fs diagnostics become own properties of the error, not of its message string

proggeramlug:
An fs error's `code`/`errno`/`syscall`/`path` lived in six side tables keyed by the **message string's address**. That produced two defects, both verified against node and both fixed here.

### 1. The wrong error got the metadata

```js
try { fs.statSync("/nope") } catch (e) { var m = e.message }
new Error(m).code      // perry: "ENOENT"   node: undefined
new Error(m).syscall   // perry: "stat"     node: undefined
new Error(m).errno     // perry: -2         node: undefined
new Error(m).path      // perry: "/nope"    node: undefined
```

The metadata belonged to the *string*, so anything holding that string answered to it.

### 2. Invisible to every reflection path

In node these are ordinary own properties. Served from a side table behind property *getters*, they appeared in none of it:

| | before | after | node |
|---|---|---|---|
| `Object.keys(e)` | `[]` | `code,errno,path,syscall` | same |
| `hasOwnProperty('code')` | `false` | `true` | same |
| `getOwnPropertyDescriptor` | `undefined` | `{value,writable,enumerable,configurable}` | same |
| `JSON.stringify(e)` | `{}` | `{"errno":-2,"code":"ENOENT",…}` | same |
| `{...e}` | `{}` | same | same |

Any code that logs or serialises a caught fs error silently lost its whole payload.

### Three sites, three different wrong assumptions

* the **fs builders** keyed on the message string;
* **`JSON.stringify`** hardcoded `"{}"` for `GC_TYPE_ERROR` — correct for a *plain* error, since `message`/`name`/`stack` are non-enumerable, but wrong once an error has enumerable own properties. It therefore also dropped **user-assigned** ones: `e.foo=1; JSON.stringify(e)` gave `{}` where node gives `{"foo":1}`. That half is independent of fs;
* **`Object.assign`/spread** had no Error arm, so it copied nothing.

All three now enumerate through `exotic_own_keys(.., enumerable_only = true)` — the same enumeration `Object.keys` uses — so they cannot drift apart again.

### Order

`ERROR_USER_PROPS` was a `HashMap` with an alphabetical `sort_by` bolted on for determinism: stable, but not node's. Own string keys enumerate in **insertion** order per ECMA-262, and it is observable through all four paths above. The store is insertion-ordered now, reassignment keeps a key's original position (`o.a=1; o.b=2; o.a=3` → `a,b`), and the fs fields install in node's `uvException` order. The GC root scanner over these props moved with it.

### Verification

Three repro programs run against node on the same host are **byte-identical, key order included**. Suite: **2746 passed, 0 failed**; fmt clean.

### Context

Found while auditing perry's address-keyed side tables. Errors carry seven of them plus four GC hooks (rekey-on-evacuation, finalize, dead-sweep, root scanner) — the same class of design that #8875 addressed for descriptors. This removes six of the seven registrations.

**Separately noted, not fixed here:** `fs.readFileSync` on a missing path returns an object instead of throwing, while `statSync`/`openSync` throw `ENOENT` correctly.

## Summary by CodeRabbit

* **Bug Fixes**
  * File-system error details such as error codes, paths, and system calls are now reliably attached to the correct error.
  * Error properties are visible through reflection, JSON serialization, object spread, and `Object.assign`.
  * Error properties now preserve JavaScript insertion order.
  * Custom properties on errors serialize and copy correctly.

### DIFF
--- a/changelog.d/8889-error-own-properties.md
+++ b/changelog.d/8889-error-own-properties.md
@@ -1,0 +1,46 @@
+Fixed an fs error's `code`/`errno`/`syscall`/`path`: they are now own properties
+of the **error object** rather than entries in side tables keyed by the error's
+**message string address**.
+
+Two defects followed from that keying, both verified against node.
+
+**The wrong error got the metadata.** Any `new Error(m)` built from the same
+message text inherited an unrelated fs error's fields — `.code` returned
+`ENOENT` where node returns `undefined`, along with `.syscall`, `.errno` and
+`.path`. The metadata belonged to the string, so anything holding that string
+answered to it.
+
+**They were invisible to reflection.** In node these are ordinary own
+properties; served from a side table behind property *getters* they appeared in
+none of the enumeration paths:
+
+| | before | after (= node) |
+|---|---|---|
+| `Object.keys(e)` | `[]` | `code,errno,path,syscall` |
+| `hasOwnProperty('code')` | `false` | `true` |
+| `getOwnPropertyDescriptor` | `undefined` | `{value,writable,enumerable,configurable}` |
+| `JSON.stringify(e)` | `{}` | `{"errno":-2,"code":"ENOENT",…}` |
+| `{...e}` | `{}` | same |
+
+Any code that logged or serialised a caught fs error lost its whole payload.
+
+Three sites each held a different wrong assumption about errors. The fs builders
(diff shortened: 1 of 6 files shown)


proggeramlug:
Landed on `main` via the #8895 batch.

PR #8890, opened 2026-08-27
perf(codegen): brand claimed-array receivers before the plain tier; gate the guarded store's layout note inline (wolf-ecs -3.2%/-3.7%, -2.3%/-2.0%)

proggeramlug:
## Summary

Two codegen follow-ups to #8876 (measured on the Mac mini reference box on top of the #8876 head; a re-measure on the merged `main` is in progress and will be posted as a comment) with the wolf-ecs (noctjs/ecs-benchmark) add/remove and entity-cycle cases, 11 alternating pairs, retained only with 11/11 wins on BOTH; semantics probe byte-identical to Node for every build.

1. **Claimed-receiver brand before the guarded plain tier** (`expr/index_get.rs`). An erased Array declaration admits object-backed Array subclasses (`class Archetype extends Array`) and typed arrays as readily as plain Arrays. The canonical-i32 read split (#8872) committed such a receiver's integer keys to the guarded plain-array tier, whose feedback fallback classified the receiver out of line on every read (`packed[sparse[x]]` in the wolf-ecs SparseSet: 4.0% / 5.6% of the two profiles even after #8876's fallback fix). The element arm now reads the GcHeader type byte once: plain Array → guarded tier; any other heap pointer → the receiver-unknown numeric tiers (inline typed-array read, dense-subclass `arrlike.ic`, complete dispatcher) the runtime-key arm already uses; non-pointers keep the guarded tier's unchanged fallback. **add/remove −3.24%, entity-cycle −3.66%.**
2. **Gated layout note in the guarded store fast arm** (`expr/index_set_guarded.rs`, `expr/write_barrier.rs`). `js_gc_note_slot_layout_aware` returns without acting when the old and new slot values share a pointer classification unless both are pointers and the array carries an element-shape proof; the fast arm still paid the call on every store (4% of add/remove, almost all `ents[id] = arch`). The fast arm now stores through a deferred-note variant of the shared audited emitter, classifies both values with an exact codegen mirror of `layout_pointer_bearing_bits`, tests `GC_ARRAY_ELEMENT_SHAPE` on the `_reserved` word it already loaded, and calls the note only from a gated `laynote` block. **add/remove −2.30%, entity-cycle −2.02%.**

Also measured and NOT included: an inline total-truthiness diamond (+3.8% / +2.8% — program-wide IR growth pushes clones past the pre-statepoint inline budgets; parked on `perf/ecs-v92-truthy-inline`), and inlining `js_method_direct_shape_class` at the multi-arm direct-method form and the dispatch tower's own-override probe (neutral / mixed: the probe's 2.4% self share is the header work itself, not the call), parked on `perf/ecs-v88-probe-inline`.

## Tests
- `index_get_claim_tests::claimed_array_receiver_brands_before_committing_a_canonical_key_to_the_plain_tier`
- `index_set_barrier_tests::the_fast_arm_layout_note_is_gated_on_the_pointer_classification_and_shape_bit`
- codegen 1325/1325, runtime array/typed_feedback suites, `-D warnings` workspace check with the host-compatible exclusions, all lint audits (file size, addr-class, raw-handle incl. `--no-raise-vs` merge base, shape census, GC store-site inventory, local-binding audit).

## Summary by CodeRabbit

  - Improved numeric element reads for typed arrays and dense `Array` subclasses accessed through broadly typed receivers.
  - Optimized array element stores by avoiding unnecessary garbage-collection tracking when it is not required.
  - Improved fast paths for dense numeric array updates.
  - Maintained behavior compatibility, with measurable benchmark improvements across entity and add/remove workloads.

### DIFF
--- a/changelog.d/8890-claimed-receiver-brand-gated-layout-note.md
+++ b/changelog.d/8890-claimed-receiver-brand-gated-layout-note.md
@@ -1,0 +1,11 @@
+Array performance: an erased Array declaration admits object-backed Array
+subclasses and typed arrays, so a canonical integer key on such a receiver now
+brands the receiver once and takes the receiver-unknown numeric read tiers
+(inline typed-array read, dense-subclass shape cache, complete dispatcher)
+instead of the plain-array tier's out-of-line feedback fallback; and the
+guarded in-bounds element store decides inline — with the exact
+pointer-bearing classification of the old and new values plus the array's
+element-shape bit — whether the GC layout note has any work before calling it.
+wolf-ecs (noctjs/ecs-benchmark) on the Mac mini reference box: add/remove
+-3.2% then -2.3%, entity-cycle -3.7% then -2.0%, each 11/11 paired wins,
+semantics probes byte-identical to Node.
--- a/crates/perry-codegen/src/expr/index_get.rs
+++ b/crates/perry-codegen/src/expr/index_get.rs
@@ -412,15 +412,82 @@ fn lower_array_index_get_via_canonical_i32_split(
         .cond_br(&is_canonical_i32, &element_label, &runtime_label);
 
     ctx.current_block = element_idx;
-    let element_value = lower_guarded_array_index_get(
-        ctx,
-        arr_box,
-        &idx_i32,
-        "aidx.dynamic",
-        require_numeric_layout,
-        coerce_numeric_fallback,
-        receiver_slot,
-    )?;
+    let element_value = if preserve_claimed_receiver_fallback {
+        // An erased Array declaration admits object-backed Array subclasses
+        // (`class Archetype extends Array` — wolf-ecs `packed[sparse[x]]`) and
+        // typed arrays as readily as plain Arrays. The guarded plain-array
+        // tier rejects those on its `GC_TYPE_ARRAY` brand and its feedback
+        // fallback then classifies the receiver out of line on every read.
+        // Read the brand once here: a plain Array keeps the guarded tier,
+        // every other heap pointer takes the receiver-unknown numeric tiers
+        // (inline typed-array read, dense-subclass `arrlike.ic`, complete
+        // dispatcher) that the runtime-key arm already uses for the same
+        // receivers. Non-pointers keep the guarded tier's unchanged fallback.
+        let brand_idx = ctx.new_block("aidx.claimed.brand");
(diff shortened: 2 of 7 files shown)


proggeramlug:
Two extra commits here are **main cleanups**, not perf work — the #8885/#8876 composition (`77b994f6b`) left `crates/perry-runtime/src/array/indexing.rs` failing the `-D warnings` workspace check (main's own CI runs were cancelled, so it went unnoticed):

- `72951364e` — three prototype-index note helpers and the two `throw_*` helpers were moved to `indexing_support.rs` (glob-imported) but their originals stayed in `indexing.rs` as dead code, plus unused `AtomicBool`/`AtomicU8` imports and a duplicated test helper. Removed the duplicates; the `indexing_support.rs` copies are the live ones.
- `e2160040f` — the composition kept only #8876's `try_strict_dense_index_set` lane in `js_array_set_f64_extend_strict`, which left #8885's `try_strict_dense_number_store` reachable from its unit tests alone. Both are exact "store what the general path stores or decline" lanes, so both are wired now: the plain-number lane (#8885) first, then the dense-index lane (#8876). @#8885's author: please shout if you'd rather order them differently.

With those, `cargo check --workspace --all-targets` under `-D warnings` (host-compatible exclusions) passes on this head; codegen 1324/1324, runtime array/typed_feedback/dyn_eval suites green, all lint audits green. A re-measure of v90+v91 on the merged main is running and will be posted here.

proggeramlug:
Re-measured on the merged `main` (`77b994f6b`, which composes #8885 with #8876) vs this PR's head `e2160040f`, Mac mini, `taskpolicy -t 0 -l 0`, 11 alternating process pairs, wolf-ecs (noctjs/ecs-benchmark):

| case | main (ms/op) | #8890 (ms/op) | paired median | wins |
|---|---:|---:|---:|---:|
| add/remove | 0.4589 | 0.4375 | **-4.78%** | 11/11 |
| entity-cycle | 0.3839 | 0.3623 | **-5.64%** | 11/11 |

(deltas add: -4.70 … -5.02, all within 0.6 pp; entity: -5.36 … -5.85.) Semantics probe byte-identical to Node on both binaries. Perry/Node now ≈ 3.27× / 2.43× on these two cases.

proggeramlug:
Landed on `main` via the #8895 batch.

PR #8892, opened 2026-08-27
fix(hir): late-bind `new X()` to a class declared later; name the ReferenceError (#8882)

### LINKED ISSUE #8882: regression(hir): Coop's Next.js fixture throws nameless `ReferenceError: identifier is not defined` at init on 0.5.1519 — loaded on 0.5.1516
## Summary

The Coop Next.js benchmark fixture (a production `next build --webpack` App Route compiled through Coop, `benchmarks/next-small` in PerryTS/coop) **loads and serves on Perry `0.5.1516` (`3885ba491`)** — that is the build every #8546 measurement was taken on — and on **`0.5.1519` (`41e8479a5`)** it dies at module init, before any request:

```
daemon: ReferenceError: identifier is not defined
daemon:     at <anonymous>
```

One application is enough; this is not the multi-heap case. Coop's daemon exits during preload with status 1, so #8546 cannot currently be reproduced on `main` at all.

## What it is not

- Not #8730's `new <minified-local>()` class. #8739 is in this build, and the compile's "unknown identifier — assuming global" list for the whole bundle contains only genuine browser globals: `window` ×112, `location` ×30, `document` ×20, `__REACT_DEVTOOLS_GLOBAL_HOOK__` ×12, `cookieStore` ×6, `define` ×4, `__non_webpack_require__` ×2. Nothing that resolves to a local.
- Not the bundle drifting: it is built from source by `scripts/prepare-next-benchmark.sh` on every run, and the bundler is asserted (`webpack-runtime.js` present, no `[turbopack]_runtime.js`).

## Where the throw can come from

The message is the nameless form (`js_throw_reference_error_unresolved_get`). Its HIR emitters on `41e8479a5`:

- `crates/perry-hir/src/lower_decl/helpers.rs:189, 204, 214, 218` (and the `_unresolved_assignment` sibling at 197)
- `crates/perry-hir/src/lower/expr_new.rs:1599`

Between `3885ba491` and `41e8479a5` (127 commits) the HIR commits that touch binding/scope resolution are the first suspects:

- `e2eee4048` fix(hir): share function and var declaration bindings (#8753)
- `98ecdc582` fix(hir): resolve the Web Streams global constructor (#8795)
- `726f6e96e` perf: static method literals, captured closure reuse, argument shape facts (#8796)
- `8a55a26f1` land #8833: ECS specializations, with the argument-route guard restored (#8837)

## Reproduction

PerryTS/coop `main` at `3e7180e` (pin `0.5.1519`, `rust-toolchain.toml` = nightly-2026-08-20):

```
scripts/build-perry-libraries.sh
cargo build --release -p coop-daemon -p coop-worker
scripts/prepare-next-benchmark.sh            # next build --webpack, then the daemon compiles it (see the compile-time issue: budget with COOP_NEXT_COMPILE_TIMEOUT)
APP=$(find target/next-benchmark/coop-run/compiled/next-bench -name app.dylib | head -1)
COOP_BENCH_APP_COUNTS=1 COOP_BENCH_EXECUTION_MODE=in_process COOP_BENCH_APP_LIBRARY="$APP" \
  cargo test --release -p coop-daemon --test resource_benchmark measure_in_process_startup_and_rss -- --ignored --nocapture
```

Same steps with `perry-main.lock` at `3885ba491` load the application (the #8546 runs).

## The ask, again

#8730 asked for it and it is still the blocker for this class: **name the identifier in the message.** `at <anonymous>` with no name means the only way to attribute this is to instrument the five throw sites and recompile the bundle, which for this fixture is a multi-hour build on `0.5.1519`. A name would make it a two-minute bisect.

proggeramlug:
## The identifier

`SentinelNode`, in `next/dist/server/lib/lru-cache.js`, constructed twice in `LRUCache`'s constructor (`this.head = new SentinelNode(); this.tail = new SentinelNode();`). Found by instrumenting the five nameless throw sites and running the cheap front-end repro (`perry compile --print-hir --no-auto-optimize --march generic handlers/main.ts` on Coop's staged `next-bench` deployment, 117 natively compiled modules). The instrumented `77b994f6b` emitted three unresolved-`new` throws: `SentinelNode` ×2 and a `typeof IntersectionObserver === "function" ? new IntersectionObserver(...) : ...` site in `app-page.runtime.prod.js` (browser API, dead on a server, so not the crash). None of the four default-parameter TDZ sites in `lower_decl/helpers.rs` fired. `SentinelNode` is not in the compile's "unknown identifier — assuming global" list because the `new` path never printed one — which is also why #8882 could not be attributed from the log.

## The regressing commit and why it broke

`905017b1c` (#8643, "class semantics tail") — it *introduced* the unresolved-`new` guard in `lower_new` (there is no `js_throw_reference_error_unresolved_get` in `expr_new.rs` at `3885ba491`; #8688 and #8739 only widened its exemption list). The guard decides at lowering time whether `new X()` can bind at all, and lowers a miss to an unconditional, nameless `ReferenceError`. Before it, an unresolved name fell through to the by-name `Expr::New { class_name }`, which codegen binds through the module class table when the `new` executes — which is why 0.5.1516 loaded.

Why the lowering-time lookups miss `SentinelNode`: the driver's CJS wrap (`cjs_wrap/hoist_classes.rs`) hoists top-level classes out of the module IIFE textually, anchored on `class ` at column 0. SWC's emit for Next's TypeScript closes the doc comment on the `class` line — ` */ class SentinelNode {` — so `SentinelNode` is never a hoist candidate and stays inside the `__perry_cjs_factory` closure, while `LRUCache` (column-0 `class`) is hoisted to module scope. Module-level classes are lowered before the init statement that holds the factory closure, so when `LRUCache`'s constructor is lowered, `lookup_class`/`forward_class_names`/… have not seen `SentinelNode` yet. Deleting just that comment from the file makes the class order `LRUNode, SentinelNode, LRUCache` and the throw disappears — that is the whole difference. The shape is routine (6 such classes under `next/dist/server` alone) and the hoister's keep/hoist fixpoint cannot help because it only reasons about classes it recognised. #8753 (top suspect in the issue) is not involved.

## The fix (`perry-hir` only)

- `pre_scan/class_decl_names.rs`: a `swc_ecma_visit` pre-scan records every class **declaration** name in the module at any nesting depth (`ctx.class_decl_names_any_depth`; class expressions excluded — their name binds only inside their own body). The guard consults it: a name declared as a class anywhere in the module keeps the late-bound by-name `Expr::New`, exactly the pre-#8643 lowering.
- Any other unresolved constructor is now a runtime `globalThis` read (`js_global_get_or_throw_unresolved(name)`) as the `NewDynamic` callee, via a new `unresolved_global_get_expr` helper shared with the bare-identifier arm so the two cannot drift. A runtime-created global constructs; a true miss throws the spec `ReferenceError: <name> is not defined` **with the identifier** — the ask in #8730 and #8882. The `new` path also prints the same "unknown identifier '…' — assuming global" warning the bare-identifier arm prints, so the culprit is a grep over the compile log next time.
- Adds `swc_ecma_visit` (already workspace-pinned, used by `perry-parser`) to `perry-hir`; `Cargo.lock` gains only that edge. No runtime/codegen change; `js_throw_reference_error_unresolved_get` stays for the default-parameter TDZ sites.

## Verification actually run

- Fixture front-end repro (`--print-hir`, 117 modules): 0.5.1516 → 0 `js_throw_reference_error_unresolved_get`; instrumented `77b994f6b` → the three emissions above (1 survives into the printed HIR; class bodies are printed as summaries); this branch → **0**, `SentinelNode` stays `New { class_name: "SentinelNode" }`, and exactly one runtime-lookup construct exists in the whole fixture: `IntersectionObserver` (total `js_global_get_or_throw_unresolved` sites 129 → 130). The unknown-identifier warning multiset is unchanged apart from that site.
- Minimal repro: the verbatim `lru-cache.js` imported from a `main.ts` reproduces the emission; the comment-stripped copy does not.
- `cargo test -p perry-hir`: all green (350 lib tests + every integration file), including the two new unit tests in `lower/tests.rs` and the updated #8739 positive control (now expects the named form). Liveness: `hoisted_class_constructs_sibling_declared_inside_a_later_closure` **fails** with the new guard clause removed and passes with it restored.
- `cargo fmt --all --check`, `cargo clippy -p perry-hir` (exit 0, no warnings on changed lines), `scripts/check_file_size.sh`, `scripts/addr_class_inventory.py`: pass.
- No version bump (per the task).

## Not covered

- No end-to-end run of the compiled fixture: a full compile is 90+ minutes and the worktree has no runtime archive; the HIR-level evidence above is what this PR stands on. Coop should re-run its `resource_benchmark` on the next pin.
- The hoister miss (` */ class X {`) is left as is — with late binding it is harmless again, as it was before #8643; hoisting those classes too would be a separate `cjs_wrap` change with its own blast radius.
- By-name late binding resolves a same-named class declared in an unrelated function scope by name (pre-#8643 behaviour, unchanged).
- The default-parameter TDZ sites still use the nameless helper; that is a different error class (`Cannot access x before initialization`) and neither issue hit it.

Fixes #8882. Refs #8730.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed unresolved `new Constructor()` calls so runtime-created global constructors can be used correctly.
  * Missing constructors now throw a standard, name-specific `ReferenceError`.
  * Improved support for classes declared inside nested scopes and closures.
  * Corrected unresolved identifier errors to include the missing name.

* **Tests**
  * Added regression coverage for late-bound constructors and named runtime errors.


proggeramlug:

PR #8893, opened 2026-08-27
fix(runtime): make the class registries per image so one process can host several apps

### LINKED ISSUE #8546: coop/in_process: second Next app dies with TypeError: value is not a function (GC rooting, multi-heap)

proggeramlug:
## Problem

A host that dlopens several Perry application images into one process — Coop, one deployment per dedicated Perry thread — ends up with only the **last-initialised** application working. Every earlier one dies on its first by-name resolution with `TypeError: value is not a function` (Next's lazy `LazyModule` → `this._onUserlandLoaded`). Established experimentally in #8546: 2 apps → first fails; 3 apps serial → first two fail; two inits interleaving closely → both fail.

## Mechanism

The class registries are process-global `static`s keyed by **compile-time class id**. Class ids come from a small sequential counter in codegen, so they identify a class *within one image* only. N copies of the same dylib register the same ids with *different* `func_ptr`s (each image's own code addresses) into one `HashMap`, and `insert` is last-writer-wins — after the last image's init, every class of every earlier image dispatches into the last image's code. `func_ptr`s are not heap pointers, which is why the from-space quarantine never faulted.

It is not one table. The 26 `js_register_class_*` / `js_class_register_*` entry points write **21** class-id-keyed process-global tables across five files — not just the six in `class_registry/state.rs` named in the issue: also the parent map and its 64 K-entry dense mirror, `CLASS_CONSTRUCTORS` + flags, `CLASS_NAMES` / `CLASS_LENGTHS` / `ANON_SHAPE_CLASS_IDS`, the `extends Error` / `DataView` / typed-array marks, the `hasInstance` / `toStringTag` hooks, generic-origin and fetch-parent maps. The issue's experiments show no write order over a shared table can work (first-wins for methods → every vtable is a mix; first-owner for all entry points → later images cannot initialise). The tables must be per image.

## Change

New module `crates/perry-runtime/src/object/class_image.rs`:

- All 21 tables move into one `ClassImageTables` struct, one instance per **image**, reached through a `perry_thread_local!` handle with a process-wide **primary** image as the fallback.
- **Entering an image** happens in `js_gc_init` — the first runtime call codegen already emits at the top of both `main` and `perry_module_init`, on the thread that runs that image's module init, before any `js_register_class_*` call. First thread to enter creates and owns the primary image; every later thread gets a fresh one. Coop's three app threads → three images. A plain executable → one. **No codegen change, no host change.**
- **Inheritance**: `perry/thread` workers (`spawn`, `parallelMap`, `parallelFilter`) and `worker_threads` Workers `adopt_image()` their spawner's handle before running anything. I verified the claim in the task: `perry/thread` workers do **not** re-run module init (`thread.rs::spawn_impl` only deserialises captures and calls the closure), so pure per-thread tables would have broken dynamic dispatch on them. Only `worker_threads` Workers re-run `__init_body`.
- **Fallback**: a thread that neither entered nor adopted — a pump running JS for the primary heap (Android's UI thread firing timers via `nativePumpTick`, per `agent.rs`), a reactor thread, a libtest thread — reads and writes the primary image, i.e. exactly the process-global table it saw before. Single-image programs are behaviourally unchanged. This is why the key is neither the thread (breaks pumps and workers) nor the `AgentId` (#8528's reason: a host's app thread never claims an agent).
- **Call sites unchanged**: each former `static RwLock<..>` is now a `static ImageTable<RwLock<..>>` whose `read()` / `write()` resolve the calling thread's image and return the same `std` guard types, so the ~100 use sites compile as written. The dense parent array becomes a 256 KiB heap allocation per image behind `parent_dense_load/store`.
- `RegistryLatch`es and `VTABLE_GEN` stay process-global on purpose (a latch armed by any image only costs another image the slow path, never a wrong answer).
- `CLASS_STATIC_ACCESSORS` leaves `per_test_global!` (the handle already isolates libtest threads that enter an image; those that don't share the primary as the other 20 tables always did), and the GC test guard no longer clears it — it holds code addresses, not roots.

## Verification (actually run, in this worktree, nightly-2026-08-20)

- `RUST_TEST_THREADS=1 cargo test -p perry-runtime -- class_image class_registry class_meta_registry class_constructors data_view thread_parent_class_id` → **38 passed, 0 failed** (first full run); re-run after the final refactor → **22 passed, 0 failed** (`class_image class_registry class_meta_registry`). This includes #8001's `a_second_agents_declared_prototype_address_is_its_own` and the #7769 dense-parent tests.
- New tests in `object::class_image::tests`:
  - `two_application_threads_keep_their_own_class_vtables` — the required regression: thread A enters an image and registers class N `m` → `0x1000`; thread B does the same with `0x2000`; both registrations complete before either looks up; A must dispatch to `0x1000`, B to `0x2000`.
  - `a_spawned_worker_shares_its_spawners_image` — worker adopts the spawner's handle, sees its classes and vice-versa; a second application sees neither.
  - `a_thread_without_an_image_uses_the_primary` — pump semantics, and `enter` is idempotent.
- **Sabotage-verified** (as #8562 was): with `enter_current_thread_image` made a no-op (process-global behaviour restored) the two-images test fails with exactly the #8546 symptom — `application A dispatches `m` into the other image's code: left: Some(8192), right: Some(4096)` — and the worker test fails with `a second application sees the first's classes: left: (Some(17), Some(34)), right: (None, None)`. Sabotage removed; green again.
- `cargo clippy -p perry-runtime -p perry-stdlib` (CI's lib-target scope) → exit 0, no findings in changed files (three `type_complexity` hits it introduced were factored into type aliases). `cargo fmt --all -- --check` clean.
- Every source-level `lint` gate: `check_file_size.sh`, `gc_runtime_root_holders.py` (+ `--self-test`), `global_sink_isolation.py`, `check_gc_scanner_latches.py`, `addr_class_inventory.py`, `class_id_collisions.py`, `gc_rekeyed_key_tables.py`, `gc_store_site_inventory.py`, `gc_pin_sites.py`, `unrooted_local_shape.py --check`, `shape_descriptor_census.py`, `string_payload_access_inventory.py`, `raw_handle_debt.py`, `local_binding_type_audit.py`, `check_gc_doc_claims.py`, `check_gc_env_knobs.py`, `workspace_architecture.py --check` → all exit 0. (Two of them caught real problems on the first pass — the sink-isolation script matched the ident in a comment, the class-id script wanted unique test constant names — both fixed.)

## Not covered

- **Not run**: the whole workspace test suite, the gap suite, the Coop end-to-end reproduction (`COOP_BENCH_APP_COUNTS=3 …`). The multi-image behaviour is proven at the unit level with the same shape as #8562; the real Coop chain needs a Perry that loads the fixture (blocked on #8882 per the issue).
- **Older images** compiled before this change still call `js_gc_init`, so they get the fix without recompiling; the hook is runtime-side.
- **Cost**: `get_parent_class_id` (the hottest class-registry read) now does the cached-TLS image resolution before its indexed atomic load; each image allocates its 256 KiB dense table. Not benchmarked here — the perf gates run on the `run-extended-tests` label.
- **Threads not handled explicitly** (reactor / IO threads, iOS/watchOS game-loop threads) fall back to the primary image — today's behaviour — so a multi-image host that runs JS on such a thread for a non-primary image would still see the wrong tables. None currently does.
- **No version bump, no docs page**; the changelog fragment carries the write-up.

Fixes #8546

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect class and method dispatch when multiple Perry application images run in the same process.
  * Ensured worker threads inherit the correct class metadata from their spawning thread.
  * Added reliable fallback behavior for threads without an assigned image.
* **Tests**
  * Added regression coverage for isolated image registration, worker inheritance, fallback behavior, dense class tables, and garbage-collection scenarios.

### DIFF
--- a/changelog.d/8893-class-registry-per-image.md
+++ b/changelog.d/8893-class-registry-per-image.md
@@ -1,0 +1,59 @@
+### Fixed
+
+- **Runtime: class registries are per image, so one process can host several
+  Perry applications.** Every class-id-keyed table module init writes —
+  `CLASS_VTABLE_REGISTRY` (instance methods / getters / setters),
+  `CLASS_STATIC_METHODS`, `CLASS_STATIC_ACCESSORS`, `CLASS_CONSTRUCTORS` and
+  their flags, the parent-edge map and its dense mirror, `CLASS_NAMES`,
+  `CLASS_LENGTHS`, `REGISTERED_CLASS_IDS`, the bind-length tables, the
+  `extends Error` / `DataView` / typed-array marks, the `Symbol.hasInstance` /
+  `Symbol.toStringTag` hooks, the generic-origin and fetch-parent maps,
+  `ANON_SHAPE_CLASS_IDS` — was a process-global `static` keyed by
+  compile-time class id (#8546). Class ids are assigned by codegen from a small
+  sequential counter, so N dlopen'd copies of an application register the SAME
+  ids with DIFFERENT `func_ptr`s (each image's own code addresses), and
+  `HashMap::insert` is last-writer-wins: after the last image's init, every
+  class of every earlier image dispatched into the last image's code. In a Coop
+  daemon hosting several Next.js deployments only the last-initialised one
+  served; the others died on their first by-name resolution with
+  `TypeError: value is not a function`. No write order over a shared table can
+  work (first-wins for methods leaves every vtable a mix of two images;
+  first-owner for every entry point leaves later images unable to initialise),
+  so the tables are now per **image**.
+
+  The model (`crates/perry-runtime/src/object/class_image.rs`): the tables live
+  in one `ClassImageTables` per image, reached through a thread-local handle.
+  `js_gc_init` — the first runtime call codegen emits in both `main` and
+  `perry_module_init`, on the thread that runs that image's module init — gives
+  the thread its own image (the first thread to enter owns the *primary*
+  image; every later one gets a fresh image). `perry/thread` workers (`spawn`,
+  `parallelMap`, `parallelFilter`) and `worker_threads` Workers adopt their
+  spawner's image before they run anything, because they never run module init
+  and must dispatch through the spawner's tables. A thread that neither entered
+  nor adopted — a pump firing JS on the primary heap's behalf (Android's UI
+  thread), a reactor thread, a libtest thread — reads and writes the primary
+  image, which is exactly the process-global table it saw before; a program
+  with one image is behaviourally unchanged. Keying by thread alone was
+  rejected because those pump/worker threads run JS that dispatches through
+  these tables without ever running init; keying by `AgentId` was rejected for
+  #8528's reason (a host's app thread is a plain `std::thread::spawn` that
+  never claims an agent). Each former `static RwLock<..>` is now a `static
+  ImageTable<RwLock<..>>` whose `read()` / `write()` resolve the calling
+  thread's image and return the same guard types, so the ~100 call sites are
+  unchanged. The `RegistryLatch`es and `VTABLE_GEN` stay process-global on
+  purpose: a latch armed by any image only ever costs another image the slow
+  path, never a wrong answer.
+
+  Regression tests (`object::class_image::tests`): two application threads
+  register the same class id with different method addresses and each
+  dispatches to its own (sabotage-verified: with `enter_current_thread_image`
+  a no-op, the last writer wins and the test fails); a spawned worker shares
+  its spawner's image while a second application sees neither; a thread with
+  no image reads the primary. Cost: the dense parent-edge read
+  (`get_parent_class_id`, the hottest class-registry read) now goes through
+  the thread-local image resolution (a cached-TLS load) before the indexed
+  atomic load, and each image allocates its 256 KiB dense table on the heap
+  instead of sharing one `.bss` array. `CLASS_STATIC_ACCESSORS` leaves the
+  `per_test_global!` set (the per-image handle already keeps one libtest
+  thread's clear out of another's reach), and the GC test guards no longer
+  clear it — it holds code addresses, not roots.
--- a/crates/perry-runtime/src/gc/mod.rs
+++ b/crates/perry-runtime/src/gc/mod.rs
@@ -1126,6 +1126,12 @@ pub fn gc_init() {
 
 #[no_mangle]
 pub extern "C" fn js_gc_init() {
+    // #8546: this is the first runtime call of every `main` / `perry_module_init`,
+    // on the thread about to run that image's module init — so it is where the
+    // thread claims its own class-registry image before any `js_register_class_*`
+    // call lands. A host that loads several application images on several
+    // threads gets one image per thread; a plain executable gets one.
+    crate::object::class_image::enter_current_thread_image();
     // Parse LLVM stack-map metadata before the first collection. The parser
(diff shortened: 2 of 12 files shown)


proggeramlug:

PR #8894, opened 2026-08-27
perf(codegen): bound TailCallElim's alloca walk on wide statepoint functions

### LINKED ISSUE #8883: perf(codegen): Coop's self-contained Next.js route takes 116 min to compile on 0.5.1519 (~8 min on 0.5.1512) — one LLVM unit thread in TailCallElim's alloca walk
## Summary

Compiling Coop's Next.js fixture — the production `next build --webpack` App Route with server chunk splitting disabled, so the route is one self-contained bundle (~15–21 MB of IR per unit, per PerryTS/coop#12) — took **6,973,947 ms (116 min)** wall on Perry `0.5.1519` (`41e8479a5`), with two earlier attempts killed by 1800 s and 5400 s budgets. PerryTS/coop#12 measured the same fixture at **roughly 8 minutes on a quiet M1** on `0.5.1512` (`fc6b3378`).

Caveat up front: this host was not quiet (load average 25–56 on 10 cores, from unrelated Perry builds), and an M1 thread pushed onto efficiency cores burns more CPU-seconds per unit of work. So the wall-clock ratio overstates it. The reason I think it is still a regression rather than contention: the single hot thread accumulated **>100 CPU-minutes** inside one analysis, which no contention model turns into 8 wall-minutes of total work. It needs a quiet-machine confirmation, which I could not give it today.

## Where the time goes

`sample` on the compiler process, twice 30 minutes apart, same picture both times. Four workers (`perry-module-0/1`, `perry-llvm-unit-0/1`); three parked on channels, one — `perry-llvm-unit-1` — running:

```
llvm::SmallPtrSetImplBase::insert_imp_big(void const*)                       3507 / 2024 samples
(anonymous namespace)::AllocaDerivedValueTracker::walk(llvm::Value*)          108 /   47
llvm::CallBase::getCaptureInfo · llvm::CallBase::onlyReadsMemory · llvm::AttributeList::getMemoryEffects
```

called from the CGSCC pipeline (`llvm::LazyCallGraph::SCC`, `CGSCCUpdateResult`, `AnalysisManager<LazyCallGraph::SCC>` frames present). `AllocaDerivedValueTracker` is **TailCallElim**'s alloca-escape walk (`Transforms/Scalar/TailRecursionElimination.cpp`): for every alloca it walks every derived use into a `SmallPtrSet`, so on a function with a very large alloca count × use count it is effectively quadratic. RSS stayed at 0.2–0.3 GB for the ~100 minutes this thread was alone, then jumped to 3.9 GB when the remaining units ran in parallel — i.e. one function held the whole build.

Perry does not use LLVM coroutines and names no coroutine or TRE pass itself; this is the default function-simplification pipeline meeting a function of that width. Per-unit stats (`PERRY_CODEGEN_UNIT_TIMINGS`) were not enabled on this run — they would name the function; #8586 added exactly that instrument for this situation.

## What changed between the pins

`crates/perry-codegen/src/native_emit.rs` moved 646+/128− between `fc6b3378` and `41e8479a5` (#8586, #8602, #8657, #8696, #8700 among them), and 72 codegen commits landed in the range, including the perf(codegen) series of 2026-08-24/25. I have not bisected. #8586's notes say #8421's contract — every function optimized at the requested level — stands, so nothing is demoted; if a change in that range made the widest function wider (more allocas, more uses), TRE's walk is where it would show first.

## Reproduction

PerryTS/coop `main` at `3e7180e`, then `scripts/build-perry-libraries.sh`, `cargo build --release -p coop-daemon -p coop-worker`, and

```
PERRY_CODEGEN_UNIT_TIMINGS=1 COOP_NEXT_COMPILE_TIMEOUT=10800 scripts/prepare-next-benchmark.sh
```

(the env override is in PerryTS/coop#17). The daemon writes the compiler's stderr only on failure, so for the unit stats run `perry compile` directly on the staged deployment with the same argv the daemon logs under `compiling deployment … perry=…`.

Filed alongside the 0.5.1519 init regression on the same fixture; the two together are what currently keeps #8546 from being reproduced on `main`.

proggeramlug:
## What is slow, measured in isolation

The unit is `benchmarks_next_small_node_modules_next_dist_compiled_jsonwebtoken_index_js` unit 1/2 of the Coop Next.js fixture (24 MB of constructed IR, 467 functions). Dumped with `PERRY_SAVE_LL`, fed through Perry's real prelude (`always-inline,function(mem2reg,sccp),rewrite-statepoints-for-gc`) and then through Perry's real pipeline (`default<Os>`; `PERRY_LL_SIZE_OPT` unset) with `opt -time-passes` on the pinned LLVM 22.1.4:

| `opt -passes='default<Os>' -time-passes` on the post-RS4GC unit | TailCallElimPass | whole pipeline (CPU) | wall |
|---|---|---|---|
| as shipped | **983.2 s (98.3 %)**, 10.5 T instructions | 999.8 s | 1669 s |
| widest function stamped `"disable-tail-calls"="true"` | 0.031 s | 16.2 s | 24.6 s |

`sample` of that `opt` shows the same frames as the issue: `TailCallElimPass::run → AllocaDerivedValueTracker::walk → SmallPtrSetImplBase::insert_imp_big`.

The function is `perry_closure_…_jsonwebtoken_index_js__229` (the bundled jsonwebtoken entry, which webpack inlines into the route). At the point the pipeline sees it: **400 allocas** (all `[N x i64]` argument arrays whose address is passed to runtime calls), **642,892 instructions**, 3,366 statepoints with **476,709 `gc.relocate`s**. A simulation of `markTails`' walk on the IR dumped `-print-before=tailcallelim` (610k instructions, 400 allocas) visits **~1.6 M uses per alloca** (max 1.79 M) — ~6.5 × 10⁸ `SmallPtrSet` insertions per TRE run — because the walk does not stop at a call: an alloca handed to any statepoint reaches the token, all of its relocates, and through their `gc-live` bundles every later statepoint. Only `load`/`store` and `captures(none)` arguments end a branch, and Perry's runtime declarations carry no `captures(none)` (in this unit only the `llvm.memset`/`memmove` intrinsics do). So the pass is `allocas × uses` on exactly the shape Perry emits for a wide, GC-heavy function. The next-widest function in the two jsonwebtoken units is 46 allocas × 46,740 instructions — 120× smaller by that product.

## The change

`crates/perry-codegen/src/inprocess.rs`: immediately before `run_passes("default<O*>")` — i.e. on the module exactly as the optimizer sees it, after RS4GC and the post-RS4GC budget — every defined function's `allocas × instructions` is computed in the same linear walk the unit census already makes. A function over the budget is stamped `"disable-tail-calls"="true"`, which is `TailCallElim`'s own early-out (`eliminateTailRecursion` returns before `markTails`), and a line names it:

```
perry: `perry_closure_…_jsonwebtoken_index_js__229` has 400 allocas across 633626 instructions (alloca-walk estimate 253450400, budget 67108864); skipping tail-call elimination for it, because TailCallElim's alloca-escape walk is quadratic in exactly that product on a statepoint-rewritten body (#8883). Every other pass still runs at the requested level; the function only loses tail-recursion-to-loop and sibling-call codegen. Override with PERRY_LL_TRE_MAX_ALLOCA_WALK=<n> (raise) or =0 (disable).
```

* **Budget**: `PERRY_LL_TRE_MAX_ALLOCA_WALK`, default `2^26` (67,108,864). `<n>` raises or lowers it, `0`/`off`/`false` disables it. Nothing is stamped at `-O0` (no TRE there). The knob is a build-cache and object-cache input (`build_cache.rs`, `object_cache.rs`), like `PERRY_LL_RS4GC_MAX_INSTRS`.
* **Calibration** of the default, `opt -passes='default<Os>' -time-passes` on the fixture's post-RS4GC native units (the six largest estimates across all 24 units, 117 modules):

  | estimate | allocas × instructions | TailCallElimPass | unit pipeline |
  |---|---|---|---|
  | 2.57 × 10⁸ | 400 × 642,892 (jsonwebtoken `__229`) | 983 s | 1000 s |
  | 9.4 × 10⁷ | 335 × 281,776 (app-page-runtime `__3288`) | 41 s | 48 s |
  | 9.4 × 10⁷ | 275 × 342,273 (app-route-runtime `__720`) | 35 s | 42 s |
  | 3.1 × 10⁷ | 117 × 260,888 | 1.0 s | 10 s |
  | 2.3 × 10⁷ | 232 × 98,270 | 0.1 s | 4.6 s |
  | 5.9 × 10⁶ | 131 × 45,038 | 1.1 s | 5.2 s |

  2²⁶ sits in the knee: everything below it spends about a second in TRE, the three functions above it 35 s, 41 s and 983 s — TRE is 82–98 % of their units' pipelines. The cost grows faster than the product (the per-walk `Visited` set stops fitting in cache), so the cap is not a linear extrapolation. The text-path modules top out at 1.3 × 10⁶.
* **Contract (#8421)**: the stamped function still runs the entire `default<Os>`/`-O3` pipeline. This is a deliberate one-pass deviation, and the code comment and the log line say so: the attribute is read by TRE (skips tail-recursion-to-loop) and by SelectionDAG's `canTailCall` / GlobalISel's `CallLowering` (calls in return position are not emitted as jumps; `musttail` is exempt and Perry emits none). No other pass reads it, and it is not an inline-compatibility attribute.
* **Not `optnone`** (#8583): the attribute is stamped after RS4GC, and it changes nothing about roots or relocations.
* The per-unit stats gain `tail_call_elim_skipped` so the shipped path is testable.

## Verification actually run

* `opt` before/after on the isolated unit: table above.
* End-to-end compile of the staged fixture (`perry compile --no-codegen --no-auto-optimize --march generic --output-type dylib handlers/main.ts`, `PERRY_CODEGEN_UNIT_TIMINGS=1`, fresh cache, same loaded box, load average 50–60), twice: **763 s wall / 1442 s user** and **736 s wall / 1252 s user** for the whole 117-module deployment (79 MB dylib), against 92–116 min in the issue. The jsonwebtoken unit reports `rs4gc 6.9s, opt 48.5s, emit 55.0s`. The budget tripped on exactly the three functions in the calibration table; the next-largest estimate in the deployment is 2.2× under the cap.
* New tests in `inprocess.rs`: the budget spellings; the estimate is `allocas × instructions` per function with an inclusive cap and only the function over it is stamped; the attribute really stops TRE at the pinned LLVM (a self-recursive tail call is turned into a loop by `default<Os>` without the attribute and survives with it — the control asserts the pass was live, per CLAUDE.md); and the shipped `optimize_and_emit` path applies the budget (`with_test_tre_walk_budget(0)` stamps every function with an alloca, the stats name them, the unit still emits, and `-O0` stamps nothing). `object_cache_tests::key_changes_with_codegen_env_vars` and `build_cache` tests cover the new key.
* `cargo fmt --check`, `cargo clippy -p perry-codegen -p perry` (no new warnings; the two remaining are pre-existing `too_many_arguments`), `cargo test -p perry-codegen --lib inprocess:: native_emit::` (27 passed), the five `perry` cache tests, `scripts/check_file_size.sh`, `scripts/addr_class_inventory.py`.

## Not covered

* The clang-subprocess backend (`PERRY_LLVM_INPROCESS=0`) does not get the attribute: the check lives where the in-process module is, and that path is the bisection fallback, not the default.
* The estimate is an upper bound (`allocas × instructions`), not the walk itself, so a function with many allocas but little relocate fan-out can trip without being expensive. In this fixture that did not happen (every tripped function was TRE-dominated), but the margin is one measured corpus wide.
* Runtime performance of the stamped functions was not benchmarked; the pipeline is otherwise byte-for-byte the same.
* The underlying shape — 400 escaping argument arrays in one function, and runtime declarations without `captures(none)` — is left as is. A Perry-side change that keeps allocas out of the statepoint walk (declaring the runtime's array parameters `captures(none)` where that is true, or fewer address-taken argument arrays) would make this budget moot for such functions and is a separate issue.

Fixes #8883

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compilation reliability and performance for functions with exceptionally large workloads.
  - Prevented optimization slowdowns that could affect complex applications and routes.
  - Builds now report when an optimization is skipped due to its safety budget.

- **Configuration**
  - Added `PERRY_LL_TRE_MAX_ALLOCA_WALK` to configure the optimization budget, with a default limit and an option to disable it.

- **Build Caching**
  - Changes to the optimization budget now correctly invalidate relevant build and object caches.


proggeramlug:

PR #8897, opened 2026-08-27
perf: ECS command path round 3 — field-push inline append, inline f64 typed guard, header-gated registry probes (+5.4%)

proggeramlug:
Round 3 on the `codehz/ecs` "5k entities: 3 commands each + sync" row, on top of #8885 (merged main measured at 4.386 ms/op vs the 7.30 ms handoff; Node 26.5.1 = 1.762 ms on the same host). Four general mechanisms, screened together with paired alternating runs on an idle Mac mini: **4.384 → 4.146 ms/op, +5.4%, 9/9** (`wxy-screen-9pairs.json`; 15-pair confirmation running). Write-up: `secret-tests/ecs-suite/PERRY_ECS_FOLLOWUP_2026-08-27_CLAUDE.md`.

- **transform: `field_push_local_bind`** — `this.f.push(v)` as a statement becomes `let old = this.f; let t = old; t.push(v); if (t !== old) this.f = t;` so the push takes the inline append (`Expr::ArrayPush`) instead of `js_array_push_guard` + `js_array_push_f64` + `js_array_length` with the layout note and barrier out of line (7% of the frame in `CommandBuffer.set`). Read-for-read and write-for-write what the native lowering did; admitted only for a declared instance array field with no accessor of that name, one non-spread argument, instance methods/getters/setters.
- **codegen: tiny-method allocation kernel sees through that expansion** — without it the expansion pushed a one-statement command-buffer method over `TINY_METHOD_MAX_STMTS`, its literal fell back to the outlined class allocation, and the first screen regressed 7.7%. Pinned by a test on the expanded shape.
- **codegen: inline f64 typed-argument guard/unbox** (`emit_typed_f64_guard`, mirrors the i32 lane) — every public entry of a function with a boxed-double clone ran `js_typed_f64_arg_guard` as a call per numeric parameter; the free-function direct-call, closure, method-override and scalar-method dispatch sites share the same helper (the Map/Set number-key and closure-capture unbox sites keep their runtime calls).
- **runtime: iteration helpers probe the typed-array/Buffer registries only for a non-`GC_TYPE_ARRAY` header** (13 sites in `iter_methods.rs`).

Tests: transform (115), codegen lib (1323) + `native_proof_regressions` (280, repinned from the runtime guard calls to the inline markers), runtime array/typed-array suites green locally; lint gates and the merge-base ratchets replayed locally against `77b994f6b`. 15-pair confirmation: **+5.51%, 15/15** (`wxy-confirm-15pairs.json`).

## Summary by CodeRabbit

* **Performance**
  * Improved ECS command-path performance, including faster field appends and typed numeric dispatch.
  * Reduced unnecessary typed-array and Buffer checks during array iteration.
  * Compiled ECS benchmark performance improved by approximately 5.4%.

* **Bug Fixes**
  * Improved handling of array-field appends when an append reallocates storage.
  * Preserved correct iteration behavior for arrays, typed arrays, and Buffers.
  * Improved reliability of numeric argument validation and conversion during optimized calls.

### DIFF
--- a/changelog.d/8897-ecs-round3-field-push-inline-append.md
+++ b/changelog.d/8897-ecs-round3-field-push-inline-append.md
@@ -1,0 +1,9 @@
+Four more general mechanisms on the ECS command path (round 3 after #8885):
+`this.field.push(v)` statements bind their receiver to a local so the push
+takes the inline append instead of the runtime push (and the tiny-method
+allocation kernel rule sees through that expansion); the f64 typed-argument
+guard and unbox are inlined at every typed dispatch; the array iteration
+helpers probe the typed-array/Buffer registries only for a non-array header.
+On the upstream `codehz/ecs` "5k entities: 3 commands each + sync" row the
+compiled benchmark went from 4.38 ms/op to 4.15 ms/op (−5.4%, paired runs on
+an idle Mac mini; Node 26.5.1 is 1.76 ms/op).
--- a/crates/perry-codegen/src/codegen/mod.rs
+++ b/crates/perry-codegen/src/codegen/mod.rs
@@ -250,8 +250,9 @@ pub(crate) use opts::{CrossModuleCtx, ImportedCtor};
 pub(crate) use param_guard::scalar_descriptor_rep;
 pub(crate) use spec_abi::{spec_abi_enabled, spec_function_name, SpecDispatch, SpecFnPlan};
 pub(crate) use typed_abi::{
-    emit_typed_arg_guard, emit_typed_arg_to_raw, generic_closure_body_name,
-    generic_function_body_name, generic_method_body_name, nonnegative_index_fast_array_method_name,
+    emit_typed_arg_guard, emit_typed_arg_to_raw, emit_typed_f64_guard,
+    emit_typed_f64_to_raw_guarded, generic_closure_body_name, generic_function_body_name,
+    generic_method_body_name, nonnegative_index_fast_array_method_name,
     nonnegative_index_fast_array_params, nonnegative_index_method_name,
     typed_arg_is_guard_candidate, typed_f64_closure_name, typed_f64_function_name,
     typed_f64_method_name, typed_f64_receiver_method_info, typed_f64_receiver_method_name,
--- a/crates/perry-codegen/src/codegen/ordinary_param_guard_tests.rs
+++ b/crates/perry-codegen/src/codegen/ordinary_param_guard_tests.rs
@@ -672,10 +672,11 @@ fn nonsuspending_async_function_needs_no_direct_call_site_for_its_guarded_clone(
             .count(),
         1
     );
-    assert_eq!(
-        public.matches("call i32 @js_typed_f64_arg_guard(").count(),
-        1
-    );
+    // The Number leg is the inline `is_number || is_int32` predicate now
+    // (`emit_typed_f64_guard`): one band test against the Perry tag range,
+    // no runtime call.
+    assert!(!public.contains("call i32 @js_typed_f64_arg_guard("));
+    assert_eq!(public.matches(", 32761").count(), 1, "{public}");
     assert!(public.contains("$spec_b_b("));
     assert!(public.contains("$generic("));
     let specialized = function_ir(&ir, "renderAsync$spec_b_b(");
--- a/crates/perry-codegen/src/codegen/spec_self_recursion_tests.rs
+++ b/crates/perry-codegen/src/codegen/spec_self_recursion_tests.rs
@@ -339,7 +339,10 @@ fn derived_recursive_number_argument_re_enters_the_guarded_clone() {
 
     // Keep both halves of the subject live: this must be the ordinary boxed
     // clone selected by the public Number guard, not the raw-i32 Tier-A path.
-    assert!(public.contains("call i32 @js_typed_f64_arg_guard("));
+    assert!(
+        public.contains(", 32761") && !public.contains("call i32 @js_typed_f64_arg_guard("),
+        "the public Number guard is the inline band test:\n{public}"
+    );
     assert!(
         clone.starts_with("define internal")
             && clone.contains("double @perry_fn_spec_self_recursion_ts__f$spec_b(double"),
@@ -376,7 +379,10 @@ fn bigint_capable_recursive_argument_keeps_the_public_guard() {
     let public = function_ir(&ir, "@perry_fn_spec_self_recursion_bigint_ts__f(");
     let clone = function_ir(&ir, "$spec_b_b(");
 
-    assert!(public.contains("call i32 @js_typed_f64_arg_guard("));
+    assert!(
+        public.contains(", 32761") && !public.contains("call i32 @js_typed_f64_arg_guard("),
+        "the public Number guard is the inline band test:\n{public}"
+    );
     assert!(
         clone.starts_with("define internal")
             && clone.contains("double @perry_fn_spec_self_recursion_bigint_ts__f$spec_b_b(double"),
--- a/crates/perry-codegen/src/codegen/typed_abi.rs
+++ b/crates/perry-codegen/src/codegen/typed_abi.rs
@@ -242,6 +242,9 @@ pub(crate) fn emit_typed_arg_guard(
     rep: TypedParamRep,
     arg: &str,
 ) -> String {
+    if rep == TypedParamRep::F64 {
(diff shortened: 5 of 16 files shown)


proggeramlug:
Audited and validated locally, merged onto current `main` (`7c9f60169`).

**Unit suites** — transform 115/0, hir 351/0, codegen 1329/0, runtime 2758/0 (`RUST_TEST_THREADS=1`), stdlib 124/0.

**Gates** — 2000-line cap, addr-class ratchet, `gc_runtime_root_holders`, node-version consistency all OK; `cargo fmt --all -- --check` exit 0; raw-handle debt unchanged (967→967, 113 ceilings) on **both** invocations, bare and `--no-raise-vs origin/main`.

**Gap suite** (full local run, 580 tests): 569 pass / 10 parity fail / 1 compile fail. Six were flagged as regressions against the snapshot. **All six reproduce identically on a `main`-built compiler**, so none attribute to this PR:

- `backoff_options`, `cron_cronjob`, `dayjs_factory_arg`, `moment_methods`, `ratelimiter_memory` — per-test A/B, same `ParityFail:1` on both arms.
- `gc_alloc_point_no_move` — this one does not finish compiling at all. Four-arm A/B (main and this PR × with and without `PERRY_NO_AUTO_OPTIMIZE=1`) times out in every arm, so it is pre-existing on `main` and independent of auto-optimize. Filed as #8906. Note the transform here cannot fire on that file anyway — it has no `this.<field>.push(` sites and no class array fields, so `admissible_fields` yields nothing.

**On the two mechanisms I couldn't settle by reading:**

- The `iter_methods.rs` narrowing is sound. `GC_TYPE_ARRAY = 1` while Buffer/TypedArray/native-view are 10/11/14, and every allocation site (`typedarray/mod.rs:952`, `native_arena.rs:323`, `buffer/header.rs:606,629`) writes a non-1 constant, so a typed array or Buffer can never present as `GC_TYPE_ARRAY`. The inverse holds too — no `register_*` call ever registers a `GC_TYPE_ARRAY` pointer, and `Uint8Array.from(plainArray)` copies into a fresh `buffer_alloc` rather than registering the source. This is arguably a net improvement against the stale-address-registry hazard (#8393), since a live `GC_TYPE_ARRAY` header now short-circuits a stale registry hit on a reused address.
- The inline f64 guard admits and rejects exactly what `js_typed_f64_arg_guard` did (`is_number() || is_int32()`), using the same tag-band arithmetic the untouched i32 lane already relies on, with the constants covered by the existing `tag_strings_match_u64_values` test. For any bit pattern the guard admits, `js_number_coerce` can only reach its int32 arm or its identity arm, and the inline conversion reproduces both.

Two cosmetic observations, deliberately **not** changed so as not to invalidate the validation above: `receiver_may_be_registered_exotic`'s `Some(header) => obj_type != GC_TYPE_ARRAY` arm is always false given `array_gc_header`'s contract (harmless, and arguably more robust if that contract ever changes), and `js_array_some_captureless` evaluates `receiver_may_be_registered_exotic` twice through `A && B || A && C` precedence.

The claimed **+5.4%** on the ECS row is not re-measured here; it was screened by the author on an idle mini.

proggeramlug:
Measured a cold-start regression from this PR on wolf-ecs `entity_cycle` (noctjs/ecs-benchmark), Mac mini, `taskpolicy -t 0 -l 0`, bisected to `2a6dcd344` (#8900 is not involved):

- 50 ms benchmark window (the harness's `# ecs-public-bench` tail at `targetMs = 50`): **0.384 → 0.949 ms/op (+147%, 0/11)**; add/remove +2.3%.
- Steady state is unchanged: the 10 s long-run twins report 0.3788 (pre) vs 0.3798–0.3857 (post) ms/op.
- Per-call probe (setup(1000), then individual calls): setup and the first call are identical (~17 ms each), but from call 2 on the pre-#8897 build runs at **0.40 ms/op** while this build runs at **~1.16 ms/op**, decaying toward 0.38 only after thousands of calls (0.53 avg over 300).
- `PERRY_GC_DIAG=1` over the same probe: pre-#8897 performs **no collection at all**; this build allocates **224,585 objects / 9.4 MB** (mean 42 B) and runs a copying minor + a major step — roughly one allocation per push. `box-stats allocs=7`, so not boxes.

The public `addComponent` also changed from the fused `js_array_push_u31_with_length` (3 sites) to `js_array_push_f64_spec` (6 sites): with `field_push_local_bind`, `this.packed.push(x)` on an object-backed `class Archetype extends Array` reaches `js_array_push_f64_spec`, which paid the tracked resolver (a guaranteed miss for a `GC_TYPE_OBJECT` header) and then `js_array_push_f64`, which paid it again before the dense subclass arm. I have a runtime-side fix for that part (subclass arm ahead of the resolver in both entries) in a follow-up branch; the per-push allocation is still being attributed — probes: `.perry-bench/warmup-probe.js`, `alloc-probe-{create,add}.js` in the regenerated noctjs workspace.

Also worth knowing for the harness numbers: the 50 ms window is dominated by warm-up, so tail-window deltas can be ±100% while steady state is flat; I'm switching my screens to a 2 s window and reporting both.

proggeramlug:
Root cause of the cold-start ramp, with a 10-line reproducer.

**Mechanism.** `field_push_local_bind` rewrites `this.f.push(v)` to

```
let old = this.f; let recv = old; recv.push(v);   // Expr::ArrayPush
if (recv !== old) this.f = recv;
```

When the append reallocates, `recv` holds the new head and `old` the growth-forwarded stub — but `!==` is JS strict equality, and Perry (correctly) sees through forwarding: `const a = []; let b = a; b.push(...×100); a === b` prints `true` exactly like Node. So the guard is **never true after a reallocation** and the write-back never runs. The field keeps the original stub for good (until a moving collection rewrites the reference), every later `this.f.length` / `this.f[i]` on that field misses the inline tiers, and `.length` ends in `js_dynamic_object_get_property(obj, "length", 6)` which allocates a heap string per call — that is the 224k/300-call `GC_TYPE_STRING` census (17–32 B each), the ~3× per-op cost, and the decay (each scavenge repairs the field, so the ramp fades over thousands of calls; steady state is unaffected).

Confirmed with lldb on the wolf-ecs closure (`destroyEntity$pshape` → `SparseSet.has` → `this.packed.length` → `js_dynamic_object_get_property`, receiver header `obj_type=1 gc_flags=0x82` = ARRAY|FORWARDED, size 144 = the initial `[]` at capacity 16) and with a temporary allocation histogram.

**Reproducer** (compile with main, run with `PERRY_GC_DIAG=1`; ~195k string allocations over 300k `has()` calls; pre-#8897 allocates nothing):

```js
class SparseSet { packed = []; sparse = [];
  has(x) { return this.sparse[x] < this.packed.length && this.packed[this.sparse[x]] === x; }
  add(x) { if (!this.has(x)) { this.sparse[x] = this.packed.length; this.packed.push(x); } } }
const rm = new SparseSet(); const other = new SparseSet(); other.packed = [];
for (let i = 0; i < 1000; i++) rm.add(i);
let hits = 0; for (let i = 0; i < 300000; i++) if (rm.has(i & 2047)) hits++;
```

(The `other.packed = []` line is what flips it in this reduction — presumably it disables the inline class-field guard so the field read returns the raw stored stub instead of a repaired handle; in wolf-ecs the equivalent is `Archetype`'s `this.sset.packed = this`. Either way the guard's `!==` is dead code by construction.)

**Fix options.** The write-back must compare *handle bits*, not JS equality — HIR has no raw-identity compare, so either (a) write back unconditionally (`this.f = recv;` — one guarded class-field store per push; note it would make a push into a frozen receiver's array throw where it shouldn't, though the current form has the same latent issue on the reallocating push), or (b) let the `ArrayPush` lowering perform the field write-back itself on its reallocation edge (it already stores the new head into the local's slot in `apush.spec.writeback` / the `realloc` block), e.g. by carrying an optional `(this, field)` write-back target on `Expr::ArrayPush` that the transform fills in. I'd go with (b); it restores exactly what the native `arr.push.wb` lowering did.

proggeramlug:
Fix is up as #8931: `Expr::ArrayPush` carries the field to write back and codegen compares the receiver local's handle bits before/after the append (JS `!==` cannot see the growth-forwarding stub), re-pointing `this.f` behind an inline plain-object header gate. Reproducer above now allocates no strings and the wolf-ecs entity-cycle warm-up curve is back to the pre-#8897 shape (0.42 ms/op from the second call vs 1.2 ms/op decaying). Locally gated (tests + all lint gates); Mac screens pending as a PR comment.

PR #8899, opened 2026-08-27
perf(gc) phase 2: memoise the shape scanner's per-address forwarding probe

proggeramlug:
Sped up the shape-table GC scanner by probing each distinct keys-array address
once per pass instead of once per shape.

`scan_shape_table_rekey_mut` is the single most expensive root scanner on
`claude -p` — **178.9 ms, 53.3% of all scanner time**. Measuring it on a
shape-heavy workload showed where that goes:

| phase | calls | total | per call | share |
|---|---:|---:|---:|---:|
| mark | 10 | 61.4 ms | 6.1 ms | 10.4% |
| rewrite | 10 | 528.5 ms | 52.9 ms | **89.6%** |

The cost is not computation. Each descriptor's probe runs
`classify_heap_space_in_range` and then reads the GC header at a scattered
address — two likely cache misses, per descriptor, per collection.

Shapes share keys arrays at a measured, stable **2.5:1**, so the loop paid that
~2.5 times per distinct address. The probe is now memoised per pass. This is
sound by construction: the same addresses are visited, just once each, and
forwarding is a pure function of the address within one pass. The carrier flag
is part of the memo key — carriers take `visit_usize_slot`, which MARKS in mark
modes, so a non-carrier's cached answer must not be allowed to satisfy a
carrier's marking duty. The per-descriptor bookkeeping is lifted into a shared
helper so the memoised and probing paths cannot drift apart; only the probe is
deduplicated.

Measured on the same workload: **3074.7 ms → 2685.9 ms of scanner time, −12.6%**.

That is well short of the 2.5× the sharing ratio suggests, because a lookup in a
300k-entry map costs nearly as much as the probe it replaces — one cache miss
traded for another. The memo is therefore a reused thread-local scratch map
rather than a fresh allocation per scan; at this size, allocating one every
collection is exactly the churn the memory-parity work is trying to remove.

### The larger finding, not fixed here

The shape table **grows without bound between full collections**. On a workload
that never holds more than 400 live objects it reached **786,205 descriptors**,
and scanner cost tracks it directly: **3.6 ms → 490 ms per call**.

The mechanism: `prune_dead_owner_side_tables_copied_minor` is nursery-only by
construction, so a keys array that is promoted and *then* dies is not reclaimed
until a full collection — while the scanner walks the whole table on every
minor. That, not the per-probe cost, is why this scanner dominates.

Fixing it means either pruning promoted-then-dead shapes sooner, or not walking
the whole table on a minor. The second is the better fix and needs the collector
to say whether old-page defrag runs in the cycle: without that, skipping
tenured entries is unsound, because a defragging moving collection *can* move
them, and a stale keys pointer is silent heap corruption. Deliberately left for
its own change rather than guessed at here.

---

### Scope, stated plainly

The phase-2 plan's acceptance bar was that `scan_shape_table_rekey_mut` and `scan_descriptor_roots_mut` disappear from the scanner profile, not merely get cheaper. **This does not meet that bar.** It makes the dominant scanner 12.6% cheaper and — more usefully — establishes with measurements why it dominates at all.

What the measurement changed about the plan:

* I expected the win to come from a **carrier index**, since carriers are <1% of descriptors. Measuring the phases separately killed that: carriers only matter in the mark phase, which is 10.4% of the cost. It would have optimised the wrong tenth.
* I expected memoisation to be worth ~2.5×, matching the sharing ratio. It is worth 12.6%, because the memo lookup is itself a cache miss. Reported as measured rather than as predicted.
* The real problem turned out to be **table size, not per-probe cost** — 786k descriptors for <400 live objects.

Two candidate fixes for the table size were considered and **not** taken:

1. **Generational filtering** (skip tenured keys arrays on a minor) — attacks the right 90%, but old-page defrag can run inside a moving collection, so tenured entries *can* move. Unsound without a defrag signal from the collector, and the failure mode is a stale keys pointer, i.e. silent heap corruption.
2. **More aggressive pruning** — plausible, but reclamation policy deserves its own change with its own tests.

Reproducer for anyone continuing this: `/tmp/parity/arch/shapes2.js` drives the table to ~780k descriptors in about a minute, and `PERRY_GC_DIAG=1` shows the per-scanner cost tracking it.

Suite: **2751 passed, 0 failed.** Stacks on #8891 (phase 1).

## Summary by CodeRabbit

- **Performance**
  - Improved garbage-collection scanning performance for shared shape data, reducing scanner time by approximately 12.6%.
  - Avoided redundant checks when multiple shapes reference the same underlying data.
- **Documentation**
  - Added benchmark results and documented remaining opportunities for improving shape-data cleanup during collection.

### DIFF
--- a/changelog.d/8892-shape-scan-probe-memo.md
+++ b/changelog.d/8892-shape-scan-probe-memo.md
@@ -1,0 +1,51 @@
+Sped up the shape-table GC scanner by probing each distinct keys-array address
+once per pass instead of once per shape.
+
+`scan_shape_table_rekey_mut` is the single most expensive root scanner on
+`claude -p` — **178.9 ms, 53.3% of all scanner time**. Measuring it on a
+shape-heavy workload showed where that goes:
+
+| phase | calls | total | per call | share |
+|---|---:|---:|---:|---:|
+| mark | 10 | 61.4 ms | 6.1 ms | 10.4% |
+| rewrite | 10 | 528.5 ms | 52.9 ms | **89.6%** |
+
+The cost is not computation. Each descriptor's probe runs
+`classify_heap_space_in_range` and then reads the GC header at a scattered
+address — two likely cache misses, per descriptor, per collection.
+
+Shapes share keys arrays at a measured, stable **2.5:1**, so the loop paid that
+~2.5 times per distinct address. The probe is now memoised per pass. This is
+sound by construction: the same addresses are visited, just once each, and
+forwarding is a pure function of the address within one pass. The carrier flag
+is part of the memo key — carriers take `visit_usize_slot`, which MARKS in mark
+modes, so a non-carrier's cached answer must not be allowed to satisfy a
+carrier's marking duty. The per-descriptor bookkeeping is lifted into a shared
+helper so the memoised and probing paths cannot drift apart; only the probe is
+deduplicated.
+
+Measured on the same workload: **3074.7 ms → 2685.9 ms of scanner time, −12.6%**.
+
+That is well short of the 2.5× the sharing ratio suggests, because a lookup in a
+300k-entry map costs nearly as much as the probe it replaces — one cache miss
+traded for another. The memo is therefore a reused thread-local scratch map
+rather than a fresh allocation per scan; at this size, allocating one every
+collection is exactly the churn the memory-parity work is trying to remove.
+
+### The larger finding, not fixed here
+
+The shape table **grows without bound between full collections**. On a workload
+that never holds more than 400 live objects it reached **786,205 descriptors**,
+and scanner cost tracks it directly: **3.6 ms → 490 ms per call**.
+
+The mechanism: `prune_dead_owner_side_tables_copied_minor` is nursery-only by
+construction, so a keys array that is promoted and *then* dies is not reclaimed
+until a full collection — while the scanner walks the whole table on every
+minor. That, not the per-probe cost, is why this scanner dominates.
+
+Fixing it means either pruning promoted-then-dead shapes sooner, or not walking
+the whole table on a minor. The second is the better fix and needs the collector
+to say whether old-page defrag runs in the cycle: without that, skipping
+tenured entries is unsound, because a defragging moving collection *can* move
+them, and a stale keys pointer is silent heap corruption. Deliberately left for
+its own change rather than guessed at here.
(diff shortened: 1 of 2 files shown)


proggeramlug:
Rebased onto current `main` and merged.

The branch carried four commits that had already landed via #8891's squash (`a09124d78`, `041313d67`, `60c192604`, `da65a8306`), which is why the diff read **+1063/−165**. Dropping those leaves the actual change: **2 files, +159/−28** — `object/shapes.rs` plus its changelog fragment.

**One thing fixed while landing it.** The doc comment and `#[inline]` intended for `record_shape_scan_outcome` were separated from it by the `perry_thread_local!` block inserted between them, so the function had neither and a meaningless `#[inline]` was sitting on a static. Moved the thread-local above the doc comment.

**On soundness.** The memo is `clear()`ed at the top of every scan, so it never carries an address across a collection — it avoids the stale-address-keyed-side-table shape (#8393) by construction, not by argument. Keying on the pre-visit `probe_addr` while storing the post-visit `addr` is right, since `addr` is mutated in place by the visit. Including the carrier flag in the key is necessary and correctly done: carriers take `visit_usize_slot`, which marks, and a non-carrier's cached answer must not discharge a carrier's marking duty. Deduplicating a *carrier's* mark is safe because marking is idempotent. Both the memoised and probing paths call `record_shape_scan_outcome` with identical arguments, so the bookkeeping cannot drift.

**Validation** — runtime 2758/0 (`RUST_TEST_THREADS=1`); `gc_runtime_root_holders` OK with the holder count moving 896 → 897, i.e. `PROBE_MEMO` is enumerated and classified rather than silently unscanned; 2000-line cap, addr-class ratchet, raw-handle debt (967→967, both invocations) all unchanged; `cargo fmt` exit 0.

**One observation, not a blocker.** The map is retained per thread to avoid per-GC churn, which is the right call for allocation churn, but at the ~300k entries described it retains roughly 10–15 MB per thread permanently (32-byte entries plus hashbrown control bytes at a 7/8 load factor). Given the standing "minimise RSS while keeping best compute" constraint, that trade is worth an explicit number in the fragment, and a capacity ceiling — shrink when it exceeds some bound — would keep the churn win without the permanent floor. Small next to the 786k-descriptor table this is scanning, so not worth blocking on.

The **+5.4%**/−12.6% figures are not re-measured here; the PR is candid that it misses its own acceptance bar and that the real term is the shape table growing without bound between full collections. That larger finding is the valuable part.

PR #8900, opened 2026-08-27
gc(shapes) phase 3: make the transition cache's target edge weak, and reap dead targets

proggeramlug:
Stopped the transition cache from keeping shapes alive: its `next_keys` edge is
now weak, and an entry whose target died is reaped.

`scan_transition_cache_roots_mut` visited `next_keys` with `visit_usize_slot`,
which **marks**. With 16384 slots the cache could therefore pin 16384 keys
arrays — and, through them, their shape descriptors — whether or not any live
object still had that shape.

That is a direct contributor to the shape table growing without bound between
full collections, measured at **786,205 descriptors on a workload holding under
400 live objects**, with the shape scanner's cost tracking it (3.6 ms → 490 ms
per call).

A transition entry is a pure cache: it answers *"adding key k to shape S yields
shape T"*. If nothing has shape T any more, the answer is worthless, so pinning
T's keys array to keep it answerable is backwards. `key_ptr` in the same entry
was already weak and metadata-only for exactly this reason; this makes the pair
consistent.

Both halves move together, and have to:

* `scan_transition_cache_roots_mut` now visits `next_keys` rewrite-only, so a
  surviving target's address stays correct but a dead one is not resurrected;
* `prune_dead_transition_cache_entries` gains `is_dead_owner(entry.next_keys)`,
  so an entry whose target did not survive is dropped rather than left dangling.

Weakening the edge without the reaping half would leave a stale pointer in the
cache.

The regression test seeds an entry with a **live** `prev_shape_id` on purpose.
An earlier version used `prev_shape_id = 0`, which the prune's pre-existing
`shape_descriptor_by_id(..).is_none()` clause already treats as dead — so it
passed with the new clause deleted and proved nothing. Sabotage-checked in its
final form: removing `is_dead_owner(entry.next_keys)` fails it.

---

### Why this, rather than shape-resident transition edges

The phase-3 plan was to move transitions onto the shape record so `add key k to shape S` becomes a pointer walk, removing `transition_cache_mutable_root_scanner` (7.4% of scanner time) and hopefully reducing `pic_prime_get`.

Measuring phase 2 changed what the interesting problem is. The scanner's own cost is not what makes this cache expensive — **its strong `next_keys` edge is**, because it pins up to 16384 keys arrays and their shapes, and the shape table is walked in full on every minor collection. So the cache was contributing to the 53%-of-scanner-time shape scan, not just its own 7.4%.

Weakening the edge attacks that directly and is a small, reviewable change with a sabotage-checked test. Relocating transitions onto shape records remains worth doing for the PIC-miss half of the original goal, but it is a refactor of the object model's transition machinery and deserves its own change rather than being bundled here.

### Review focus

The correctness question is whether weak + reap is sound in every order. The reasoning: a transition entry is derivable state, so dropping one is always safe; the prune runs inside the collection via `DEAD_KEY_PRUNES`, before any mutator can read the table; and `key_ptr` has been weak on this exact model already.

Suite: **2752 passed, 0 failed.** Stacks on #8891 (phase 1) and #8899 (phase 2).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved garbage collection of cached transition data, allowing unused entries and targets to be reclaimed.
  * Ensured metadata references are correctly traced and rewritten during garbage collection.
  * Preserved existing metadata relationships while supporting additional named properties on supported built-in objects.

* **Tests**
  * Added regression coverage for metadata tracing, property storage, and transition-cache cleanup.

### DIFF
--- a/changelog.d/8900-transition-cache-weak-target.md
+++ b/changelog.d/8900-transition-cache-weak-target.md
@@ -1,0 +1,34 @@
+Stopped the transition cache from keeping shapes alive: its `next_keys` edge is
+now weak, and an entry whose target died is reaped.
+
+`scan_transition_cache_roots_mut` visited `next_keys` with `visit_usize_slot`,
+which **marks**. With 16384 slots the cache could therefore pin 16384 keys
+arrays — and, through them, their shape descriptors — whether or not any live
+object still had that shape.
+
+That is a direct contributor to the shape table growing without bound between
+full collections, measured at **786,205 descriptors on a workload holding under
+400 live objects**, with the shape scanner's cost tracking it (3.6 ms → 490 ms
+per call).
+
+A transition entry is a pure cache: it answers *"adding key k to shape S yields
+shape T"*. If nothing has shape T any more, the answer is worthless, so pinning
+T's keys array to keep it answerable is backwards. `key_ptr` in the same entry
+was already weak and metadata-only for exactly this reason; this makes the pair
+consistent.
+
+Both halves move together, and have to:
+
+* `scan_transition_cache_roots_mut` now visits `next_keys` rewrite-only, so a
+  surviving target's address stays correct but a dead one is not resurrected;
+* `prune_dead_transition_cache_entries` gains `is_dead_owner(entry.next_keys)`,
+  so an entry whose target did not survive is dropped rather than left dangling.
+
+Weakening the edge without the reaping half would leave a stale pointer in the
+cache.
+
+The regression test seeds an entry with a **live** `prev_shape_id` on purpose.
+An earlier version used `prev_shape_id = 0`, which the prune's pre-existing
+`shape_descriptor_by_id(..).is_none()` clause already treats as dead — so it
+passed with the new clause deleted and proved nothing. Sabotage-checked in its
+final form: removing `is_dead_owner(entry.next_keys)` fails it.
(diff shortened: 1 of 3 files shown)


proggeramlug:
Rebased onto current `main` (dropping the five commits already landed via #8891 and #8899's squashes) and merged. Real content: **3 files, +114/−1**.

**I verified the sabotage claim rather than taking it on trust.** Deleting `|| is_dead_owner(entry.next_keys)` turns the `dead_owner_side_tables` suite red; restoring it gives 49/0. The test genuinely exercises the clause it was written for — which, given the writeup's own account of an earlier version that passed with the clause deleted, was the thing most worth re-checking.

**On the weakening.** The two halves do have to move together, and they do. The prune's pre-existing `if entry.next_keys == 0 { continue; }` guard means the new `is_dead_owner(entry.next_keys)` is never called with a zero address, so the missing `!= 0` guard that `key_ptr` carries is not needed here.

**Evacuation.** Since this turns a marking edge into a rewrite-only one, the failure mode to look for is a target dying that something still reads. Under `PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1` the failing set is **identical to main's — 16 on both arms, zero symmetric difference** (those 16 are pre-existing; most are `promote_in_place` tests that assert in-place promotion, which forced evacuation contradicts by construction). So the weakened edge adds no evacuation failure.

**Fixed while landing:** a stray blank line had been inserted between `#[cold]` and `prune_dead_transition_cache_entries`.

**Validation** — runtime 2759/0 (`RUST_TEST_THREADS=1`); 2000-line cap, `gc_runtime_root_holders` (897), raw-handle debt (967→967) all unchanged; `cargo fmt` exit 0.

PR #8901, opened 2026-08-27
bench(shapes) phase 4: the measurement that refutes dictionary mode, and where the real gap is

proggeramlug:
Added `benchmarks/bench_dynamic_property_keys.ts`, and with it the measurement
that **refutes the premise for a dictionary mode**.

The phase-4 plan was a formal dictionary representation for objects that defeat
shapes — heavy `delete` use, thousands of unique keys — to stop pathological
objects minting shapes the table must then carry and rekey forever.

The benchmark runs two loops with the same number of property writes and reads,
differing only in a `delete`, so the ratio isolates shape churn from raw
property-access cost. Measured on an idle host, N = 300 000:

| engine | delete-heavy | overwrite-only | delete penalty |
|---|---:|---:|---:|
| node | 36 ms | 21 ms | **1.7×** |
| perry | 1487 ms | 1321 ms | **1.1×** |

**Perry's delete penalty is lower than node's.** Delete-driven shape churn is
not disproportionately expensive here, so a dictionary representation — a new
object representation, with its own property storage, PIC handling, enumeration
and GC integration — would be a large investment aimed at a tail perry does not
have. It is not implemented, and on this evidence should not be until a workload
shows the penalty that motivates it.

The second column is the finding worth acting on: **~60× on plain dynamic
property overwrite**. Profiling that binary attributes it to inline-cache misses
and shape-table probes, not deletion:

| samples | symbol |
|---:|---|
| 324 | `js_array_get_f64` |
| 307 | `value::addr_class::try_read_tracked_gc_header` |
| 105 | `shapes::shape_descriptor_by_id` |
| 103 | `shapes::shape_descriptor_ensure_with_generation` |
| 72 | `js_put_value_set_dyn_ic_miss` |

The two shape-table probes are ~13% of main-thread samples on their own — a
hash lookup per property access, on a key space (`ShapeId`) that is allocated
densely by a single atomic counter and could be an array index instead. That is
a concrete, bounded follow-up; it is not done here because the ids are minted
from a process-global counter while the tables are per-thread, so a dense `Vec`
could be sparse on a multi-threaded program, and that trade needs measuring
rather than assuming.

Both columns matter when changing this benchmark: the ratio is what refutes the
dictionary-mode premise, the absolute is what tracks the real gap.

---

### This phase ships a measurement, not a feature — deliberately

Phase 4 of the object-model plan was "formal dictionary mode". I did not implement it, because measuring the premise first showed it does not hold for perry.

The argument for a dictionary representation is that delete-heavy objects defeat shapes and blow up the shape table. The delete penalty is the number that argument rests on, and **perry's is 1.1× against node's 1.7×** — perry is already *less* sensitive to it than V8 is. Building a second object representation to fix that would be a large, invasive investment aimed at a tail perry does not have.

What the same benchmark exposes is a ~60× gap on plain dynamic property overwrite, attributed by profile to inline-cache misses and shape-table probes. That is the thing worth the next investment, and it is a different piece of work than dictionary mode.

I would rather hand over a benchmark that says "not this, that" than a half-built object representation.

### The bounded follow-up this points at

`shape_descriptor_by_id` + `shape_descriptor_ensure_with_generation` are ~13% of main-thread samples in this workload — a hash lookup per property access. `ShapeId`s are minted densely from a single atomic counter, so an array index is the obvious replacement.

The catch, and why it is not in this PR: the counter is process-global while the shape tables are per-thread, so on a multi-threaded program each thread's dense-`Vec` would be sparse. At the table sizes measured in #8899 (786k descriptors) that trade needs measuring, not assuming.

Suite unchanged by this PR (benchmark + changelog only). Stacks on #8891, #8899, #8900.

## Summary by CodeRabbit

* **Bug Fixes**
  * Error objects now expose filesystem metadata as enumerable own properties, matching Node.js behavior.
  * Error properties are preserved correctly during reflection, serialization, object spreading, and `Object.assign`.
  * Error property insertion order, reassignment, deletion, and explicit `undefined` values now behave consistently.
  * Improved memory management for object metadata and property storage.
* **Performance**
  * Improved garbage-collection scanning and dynamic property access behavior.

### DIFF
--- a/benchmarks/bench_dynamic_property_keys.ts
+++ b/benchmarks/bench_dynamic_property_keys.ts
@@ -1,0 +1,73 @@
+// Benchmark: dynamic string-keyed property access, with and without `delete`.
+//
+// Two loops do the SAME number of property writes and reads; only the `delete`
+// differs. Comparing them isolates shape churn from raw property-access cost,
+// which is what makes this benchmark worth keeping:
+//
+//   * `deleteHeavy / overwriteOnly` is the *delete penalty* — how much a
+//     delete-driven shape walk costs relative to a stable shape.
+//   * `overwriteOnly` on its own is *baseline dynamic property throughput*.
+//
+// Measured 2026-08-28 (idle host, perry 0.5.1519, N = 300_000):
+//
+//   engine   delete_heavy   overwrite_only   delete penalty
+//   node           36 ms           21 ms           1.7x
+//   perry        1487 ms         1321 ms           1.1x
+//
+// The delete penalty is the number the "objects that defeat shapes need a
+// dictionary mode" argument rests on — and perry's is LOWER than node's. Adding
+// a dictionary representation would therefore be a large investment aimed at a
+// tail perry does not have.
+//
+// The second column is the real gap: ~60x on plain overwrite. Profiling this
+// binary puts the time in `js_array_get_f64`, `try_read_tracked_gc_header`,
+// `shape_descriptor_by_id` + `shape_descriptor_ensure_with_generation` (two
+// hash lookups per access on the hot path), and `js_put_value_set_dyn_ic_miss`
+// — i.e. inline-cache misses and shape-table probes, not deletion.
+//
+// Keep both columns when changing this file: the ratio is what refutes the
+// dictionary-mode premise, and the absolute is what tracks the real gap.
+
+function deleteHeavy(n: number): number {
+  const o: Record<string, number> = {};
+  let s = 0;
+  for (let i = 0; i < n; i++) {
+    const k = "k" + (i % 500);
+    o[k] = i;
+    s += o[k];
+    delete o[k]; // walks the object back to a previous key set
+  }
+  return s;
+}
+
+function overwriteOnly(n: number): number {
+  const o: Record<string, number> = {};
+  let s = 0;
+  for (let i = 0; i < n; i++) {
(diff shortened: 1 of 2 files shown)


proggeramlug:
Rebased onto current `main` (dropping the six commits already landed via #8891/#8899/#8900) and merged. Real content: **2 files, +118/−0** — a benchmark and its changelog, zero Rust.

**I re-ran the benchmark rather than taking the table on trust**, since a measurement artifact that does not run is worthless, and the whole value of this PR is the measurement:

| engine | delete-heavy | overwrite-only | delete penalty |
|---|---:|---:|---:|
| node 26.5.1 | 32 ms | 20 ms | **1.6×** |
| perry (main build) | 1395 ms | 1130 ms | **1.2×** |

Close to the reported 1.7× / 1.1×, and `checksum=0` on both engines, so they are computing the same thing. My host was not idle — I had builds running — which makes the agreement more convincing rather than less.

**The central claim reproduces: perry's delete penalty is lower than node's.** So the premise for a dictionary mode is genuinely refuted, and declining to build a whole second object representation on this evidence is the right call. The finding worth acting on is the other column — ~57× on plain dynamic overwrite in my run — which the writeup already attributes to inline-cache misses and two shape-table hash lookups per access rather than to deletion.

A negative result that stops a large investment, with the measurement committed so the next person can re-run it, is worth more than the feature would have been.

PR #8916, opened 2026-08-28
perf: ECS round 4 — barrier early exits, inline captureless some loop, lean Map/Set lanes, empty-pop fast path, codegen-time const fold, inline hot-TLS values (−16.7%)

proggeramlug:
Round 4a on the `codehz/ecs` "5k entities: 3 commands each + sync" row, on top of #8897 (merged main measured at 4.14 ms/op on the Mac mini; Node 26.5.1 = 1.762 ms on the same host). Ten general mechanisms in eight steps, each screened with paired alternating runs on the idle Mac mini and confirmed over 15 pairs:

|---|---|---|---|---|
| 4a barrier dirty-page early exit + Map dense key | merged main #8897 4.142 ms | 3.961 ms | +4.41%, 9/9 | **+4.37%, 15/15**, 30 oracles (`r4a-confirm.json`) |
| 4b inline `arr.some(capturelessArrow)` loop | 4a 3.960 ms | 3.814 ms | +3.85%, 9/9 | **+3.70%, 15/15**, 30 oracles (`r4b-confirm.json`) |
| 4c lean Map numeric lookup lane | 4b 3.814 ms | 3.686 ms | +3.37%, 9/9 | **+3.30%, 15/15**, 30 oracles (`r4c-confirm.json`) |
| 4d small-Set member scan before the side-table | 4c 3.684 ms | 3.617 ms | +1.85%, 9/9 | **+1.80%, 15/15**, 30 oracles (`r4d-confirm.json`) |
| 4f empty-array `pop` fast path + single-pass length-0 re-arm | 4d 3.617 ms | 3.599 ms | +0.48%, 9/9 | **+0.48%, 15/15**, 30 oracles (`r4f-confirm.json`) |
| 4g module-const literal fold at codegen entry | 4f 3.600 ms | 3.546 ms | +1.49%, 9/9 | **+1.49%, 15/15**, 30 oracles (`r4g-confirm.json`) |
| 4h inline hot-TLS values in `HotTls` | 4g 3.545 ms | 3.503 ms | +1.18%, 9/9 | **+1.17%, 15/15**, 30 oracles (`r4h-confirm.json`) |
| 4i leaf write-barrier entry | 4h 3.505 ms | 3.450 ms | +1.59%, 9/9 | **+1.56%, 15/15**, 30 oracles (`r4i-confirm.json`) |

Cumulative: 4.14 → 3.45 ms/op (−16.7%); Node 26.5.1 = 1.762 ms on the same host, so Perry is at ~1.96× Node from 2.35×. Write-up: `secret-tests/ecs-suite/PERRY_ECS_FOLLOWUP_2026-08-27_CLAUDE.md`.

- **gc: an inline-slot store onto the cached dirty page skips both barrier classifications.** `write_barrier_decoded_parent` classified the parent (arena lookup + header decode) and then the child before consulting the dirty-page cache. The cache's invariant is "cached ⟹ recorded in `DIRTY_OLD_PAGES` and stamped dirty", and the minor collector rescans every recorded dirty page, so once a page is cached every further inline-slot store onto it is already covered whatever the child is — the check now runs first, keyed on the slot address (an external slot, or a slot below the parent, still takes the full path). The barrier family was ~10% of the frame's self time; the hit is counted under `BarrierTraceCounter::DirtyPageCacheHits` and pinned by `inline_slot_store_onto_the_cached_dirty_page_is_a_cache_hit`.
- **runtime: `Map` dense integer key in one round trip** — `dense_integer_key` pre-screened with three range tests (`is_finite`, `< 0`, `> u32::MAX`) before the `as u32` round trip; `as u32` saturates (NaN/negative → 0, +inf/too large → `u32::MAX`), so the round-trip compare alone already rejects every value that is not a finite integer in `0..=u32::MAX`. One conversion pair per dense Map lookup (`-0` still maps to 0 as before).
- **codegen: `arr.some(capturelessArrow)` as an inline loop with a direct body call** (`lower_captureless_some_inline`). `js_array_some_captureless` decided the receiver once and then, per element, re-resolved the head from its root, NaN-boxed the receiver and called the body through the function pointer (4.5% self). The lowering makes the same one-time decision on the same live bits — `GC_TYPE_ARRAY` head, not forwarded, no indexed descriptors, the sticky `PERRY_ARRAY_INDEX_FAST_PATH_INVALIDATED` byte clear, `length <= capacity` — and runs the loop inline: the head is re-read from its root every iteration (a forwarded head goes through the new `js_array_live_head` export), indices past the live length and holes are skipped, the arrow's body symbol is called directly with as many of `(element, index, receiver)` as it declares, and `true`/`false` results decide inline with `js_is_truthy` for anything else. Every receiver the loop does not admit takes the runtime helper, which stays the fallback. Pinned by `captureless_inline_some_passes_the_callback_body_directly` (direct call + loop markers + fallback).
- **runtime: Map lookups run a lean numeric lane before the general `find_key_index`.** The function carried the string-hash, pointer-index, hashed-numeric and generic-compare paths in one body; a PC histogram of the profile put a third of its 5.5% self time on the prologue/epilogue those cold paths force (eight callee-saved GPRs and four FP registers on arm64) and half on the dense-key range tests. The two shapes the numeric side-table exists for — a plain (untagged, non-NaN, non-zero) number against a small map's entries by bit identity, or against the dense integer range table — now run in an always-inlined `find_key_index_hot` inside `js_map_get`/`js_map_has`/`js_map_set`'s callers; everything else goes to the outlined `find_key_index_cold`. A dense-range miss stays definitive for its span (every insert, delete, clear and GC rewrite keeps the table exact); a key outside the span, a tagged, zero or NaN key, and every string or pointer key take the cold path unchanged. `hot_lookup_lane_agrees_with_the_cold_path_on_every_key_shape` pins hit/definitive-miss/out-of-span/-0/NaN/tagged shapes on both a small and a dense map.
- **runtime: small-Set lookups scan the members before the side-table.** `find_value_index` answered every `Set.has`/`Set.add` through the thread-local `SET_INDEX`: a hash of the set address to reach its table, then a hash of the value — two probes for sets that in the hot shapes (an archetype's component-type set) hold three or four numbers (`componentTypeSet.has` was 7.5% of the in-place update path). A plain number against a set of at most eight elements is now decided by reading `elements[0..size)` — exactly the membership, since delete compacts and add normalises `-0`, and no tagged value equals a number — so a bit match is a hit and a full scan a definitive miss. Larger sets, tagged/zero/NaN values and every string keep the side-table, outlined. `small_set_scan_lane_agrees_with_the_side_table_on_every_value_shape` pins hit/miss/-0/NaN/tagged/delete-compaction/growth-past-the-bound.
- **runtime: `pop()` on an empty plain array answers from the header fast path; `length = 0` re-arms an all-pointer head in one registry pass.** The pop fast path required a non-empty array, so the drained pool's `pool.pop() ?? []` fell through the whole generic tower (subclass and plain-object probes, a tracked classification, the flag resolution) to reach the same `length == 0` return; with the descriptor flag excluded, `Set(O, "length", 0)` is a no-op and there is no index to Get or Delete, so the answer is `undefined` from the header read. `rebuild_array_layout` on `length = 0` of an all-pointer head ran the zero-slot rebuild and then `layout_init_all_pointer_slots`, which clears the same bit, forgets the same two record kinds and sets the state — two passes over the layout registries per `pooled.length = 0`; the re-arm now runs alone with an identical end state (the round-U truncate test additionally asserts no per-object record survives).
- **driver: module-level const literals fold into their reads after the transform phase, before codegen** (`perry_transform::module_const_fold`, run from `run_pipeline.rs` once every module is transformed). `export const COMPONENT_ID_MAX = 1023` is a module-scope immutable let and every read of it is a `LocalGet` the typed-ABI clone rules cannot type, so a one-line predicate such as `isComponentId` (`id >= 1 && id <= COMPONENT_ID_MAX`) was refused its `i1` clone and every call ran a module-global load plus the dynamic tag-coercion compare on both operands. It is deliberately not a pipeline pass: folded, those predicates become self-contained and the cross-module inliner harvests them — run inside the pipeline that consumed callers' inline budgets (`world.set` lost `resolveSetOperation`, −43%), and with a larger budget the inlined bodies still did the dynamic compare on the untyped call-site value. Run after all harvests are taken, no inlining decision moves; the fold precedes the HIR trace and the object-cache fingerprint so both describe the tree codegen consumes. Admission and the TDZ rule are pinned by the module's unit tests.
- **runtime: the hottest small thread-local values live inline in `HotTls`.** A hot-TLS slot and a named pointer field both resolve as TSD base → `HotTls` → slot pointer → value; PC histograms of the post-4g profile put the remaining self time of the write barrier, `js_map_clear` (10k calls/frame, both hot offsets on its two TLS probes), `is_registered_box_ptr` and `array_prototype_addr` on that dependent chain rather than on anything they compute. Small `Copy` values with a `const` initial state can live in `HotTls` itself (TSD base → `HotTls` → value), so the barrier's one-entry dirty-page cache, the memoized `Array.prototype`/`Object.prototype` rows and the three box-pointer caches now do; the generic slot mechanism is unchanged for everything else, and the collector's root rewrite of the prototype rows walks the inline cells exactly as it walked the slot.
- **gc: the write barrier's dirty-page hit returns from a leaf entry.** Every pointer store into an old object reached `js_write_barrier_slot_validated_parent`, which made two out-of-line calls before anything was decided — `decode_heap_addr` for the child, and `incremental_mark_barrier_value`, whose "no cycle anywhere" test sat inside the callee — and then entered the outlined `write_barrier_decoded_parent` (six callee-saved registers) to run the one-entry dirty-page compare that answers the second and third push into the same bucket. The tag decode and the idle test now inline (their slow arms are cold, out of line), and the cache test is hoisted into the entry ahead of the outlined body (`gc/barrier/leaf.rs`), so a hit is a leaf path. Counters and the remembered set built are unchanged; pinned by `validated_parent_entry_answers_a_cached_dirty_page_store_before_the_body`.

Tests: runtime suite (2763) incl. `gc::tests::barrier` + `map::`, codegen lib (1329) + `native_proof_regressions` (280), transform lib (119), runtime array suites (264), lint gates and the merge-base ratchets replayed locally against `f9890759c` (the shape-descriptor census step fails identically on pristine `origin/main` — pre-existing, not touched here).


proggeramlug:
Audited and merged.

**On the barrier early-exit.** The soundness rests on the dirty-page cache invariant, and the thing worth checking was whether the new exit keys on the same page as the recording it short-circuits. It does: `remember_old_to_young_inline_slot` records `mark_dirty_old_page(generation_page_for_addr(slot_addr))` (barrier/mod.rs:1649), and the exit tests `dirty_old_page_already_marked(generation_page_for_addr(slot_addr))` — the same key. Objects spanning pages are covered by `mark_dirty_parent_span` marking every page from `header_addr` to `header_addr + total_size - 1`, so keying on the slot's page rather than the parent's is right, not a shortcut.

The SATB claim also holds: `barrier_child_prologue(child)` and `barrier_remembering_active()` both run in the caller before `write_barrier_decoded_parent`, so an early return here cannot skip shading. And the exit does not need its own generation check — the cache only ever holds pages for which `old_page_mark_dirty` returned true, so a cached page is by construction an old page recorded dirty.

This adds a consumer of the #7187 cache rather than weakening its invariant, which stays maintained by `invalidate()` on every clearing path.

**On `dense_integer_key`.** The round-trip compare alone is sufficient, since Rust's float→int `as` has been saturating since 1.45: NaN → 0 fails the compare, negatives → 0 fail it, `+inf` and anything above the max → `u32::MAX` fails it, non-integers truncate and fail it. The boundary is the interesting one and it works — `4294967296.0` saturates to `4294967295`, which does not compare equal. `-0.0` still maps to 0, as noted.

**Validation** — runtime 2760/0 (`RUST_TEST_THREADS=1`); the new `inline_slot_store_onto_the_cached_dirty_page_is_a_cache_hit` passes; under `PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1` the failing set is **identical to main's, zero symmetric difference in both directions** (16 pre-existing, mostly `promote_in_place`). `scripts/run_lint_gates.sh`: all 53 gates pass.

**On the census note in the description** — accurate, and now fixed. It was not pre-existing in the sense of being nobody's doing: it broke at #8899 (`d02892491` red, `2a6dcd344` green), which I had merged after running only the topically-relevant gates. #8918 restores it. Thanks for flagging it in passing; that is how it got found.

The +4.4% is not re-measured here; it was screened 15/15 on the idle mini.

PR #8917, opened 2026-08-28
perf(shapes): direct-mapped cache in front of the shape-descriptor table

proggeramlug:
Added a direct-mapped cache in front of the shape-descriptor table.

`shape_descriptor_by_id` is on the hot property path. Profiling a dynamic
string-keyed property loop put it and `shape_descriptor_ensure_with_generation`
at roughly **13% of main-thread samples between them**, and every call paid a
TLS fetch, a `RefCell` borrow and a hash probe — to reach a record whose address
never moves. `Box<ShapeDescriptor>` is stable across rehash, so a 256-way
direct-mapped cache can hold the record's address and a hit becomes mask,
compare, deref. 4 KiB per thread, fixed.

Two decisions that matter for correctness:

**It caches the record's ADDRESS, not a copy of the descriptor.** Records are
mutated in place — `old_carrier`, `cache_carrier`, and `keys` after evacuation —
so a cached copy would go quietly stale. Holding the address means a hit always
reads current data.

**Epoch invalidation is selective.** Removal frees the box, and one insert path
can replace a live id with a fresh box; both bump the epoch. A *fresh-id* insert
deliberately does not, because it cannot invalidate an existing way, and bumping
there would flush the cache on every shape creation — exactly the workloads that
build shapes.

Measured with `benchmarks/bench_dynamic_property_keys.ts` on an idle host,
minimum of 7 runs:

| | delete-heavy | overwrite-only |
|---|---:|---:|
| node | 31 ms | 19 ms |
| before | 1222 ms | 1088 ms |
| after | **1187 ms** | **955 ms** |

Baseline dynamic property throughput improves **12.2%**. That is well short of
the 13% of samples the two shape functions hold, because only
`shape_descriptor_by_id` is served from the cache;
`shape_descriptor_ensure_with_generation` is a separate path and still probes.

This does not close the gap to node — perry is still ~50x on this loop. The
remaining weight sits in `js_array_get_f64` (324 samples) and
`value::addr_class::try_read_tracked_gc_header` (307), which are the next
targets and are not touched here.

The invalidation test is sabotage-checked: deleting the bump in
`remove_descriptor_and_reverse_indices` fails it. That check matters more than
usual here — a stale way is not a wrong answer, it is a dangling pointer to a
dropped `Box<ShapeDescriptor>` reached from the hot property path.

---

First step against the largest measured gap to node: **~60x on plain dynamic property overwrite** (`benchmarks/bench_dynamic_property_keys.ts`, added in #8901). This takes 12.2% of it.

### Review focus

The correctness question is the epoch, and it is worth checking rather than trusting:

* a way holds a raw `*const ShapeDescriptor` into a `Box` the table owns;
* the epoch must be bumped on **every** path where that box can be freed or replaced while its id stays in use.

I found two: `remove_descriptor_and_reverse_indices` (the single removal funnel) and the `intern`-side insert that can replace a live id. If there is a third, a stale way hands out a dangling pointer from the hot property path — so that is the thing to look for.

The fresh-id insert path deliberately does **not** bump. That is a performance decision with a correctness argument behind it (a new id cannot alias an existing way), and it is the difference between a cache that works and one that is flushed on every object literal.

Suite: **2759 passed, 0 failed.**

## Summary by CodeRabbit

* **Performance**
  * Improved dynamic property access performance with a new shape-descriptor lookup cache.
  * Benchmarks show up to a 12.2% improvement in baseline dynamic property throughput.

* **Bug Fixes**
  * Shape lookups are now correctly invalidated when descriptors are removed or replaced.

* **Tests**
  * Added coverage for cache invalidation and preservation of valid shape lookups.


proggeramlug:
Audited. **Two things fixed, one blocker I'm not comfortable resolving on your behalf.**

## Sound, and verified

I re-ran the sabotage check rather than trusting it: deleting the bump in `remove_descriptor_and_reverse_indices` turns the shapes suite red; restoring it gives 32/0. Runtime is 2762/0, and under `PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1` the failing set is identical to main's — 16 both ways, zero symmetric difference.

I also enumerated every path that can drop or replace a `Box<ShapeDescriptor>`, since a stale way here is a dangling deref on the hot property path:

- `remove_descriptor_and_reverse_indices` (`shapes.rs:388`) — invalidates ✓
- the replacing insert (`shapes.rs:797`) — invalidates ✓
- the fresh-id insert (`shapes.rs:516`) — correctly does not need to. A fresh id can only collide with a *different* id in its way, and the way compares the id, so that is a miss, not a stale hit. Ids come from a monotonic counter, so no reuse.

## Fixed while auditing

**`test_clear_shape_table` drops every box and invalidated nothing** (`shapes.rs:~1991`). It is `#[cfg(test)]`, so not production UB — but any test that clears the table and then does a lookup derefs a freed `Box`, and that is precisely the kind of UB that surfaces later as an unrelated flaky failure in a different suite. Added the `invalidate_shape_lookup_cache()` call.

Also renamed `changelog.d/8902-shape-lookup-cache.md` → `8917-…`; fragment filenames are PR-keyed so in-flight PRs cannot collide.

## Blocker 1 — file size

`shapes.rs` is **2057 lines**, 57 over the cap, so `check_file_size.sh` is red. I did not split it, because whatever you do about blocker 2 will likely move this code and my split would just conflict.

## Blocker 2 — `VIEW_REGISTRY` flips UNCOVERED → COVERED, and I can't explain why

`scripts/gc_runtime_root_holders.py` fails with a stale `frontier` entry:

```
crates/perry-runtime/src/buffer/view.rs | VIEW_REGISTRY
```

`--list` on each arm:

```
main:  UNCOVERED [core/T] buffer/view.rs:58 VIEW_REGISTRY
#8917: COVERED   [core/T] buffer/view.rs:58 VIEW_REGISTRY
```

The gate's prescribed action is to delete the stale entry — "the deletion is the receipt." **I am not doing that**, because deleting it asserts the collector now scans that holder, and I cannot substantiate the claim. If the COVERED verdict is an artifact of the call-graph walk rather than a real coverage gain, deleting the entry retires a precisely-named piece of GC custody debt for a holder that is still unscanned, and this gate exists specifically because that class of bug is a *perfectly reproducible* use-after-free that nothing else finds.

Things I ruled out:

- **Not my edit.** Both arms fail identically with and without the `test_clear_shape_table` fix.
- **Not merely adding a `state()` call.** I hypothesised the walk resolves `crate::state::state()` loosely and widens reachability, and tested it: adding a bare `let _probe = &crate::state::state().shapes;` inside the registered scanner `scan_shape_table_rekey_mut` on pristine `main` does **not** flip `VIEW_REGISTRY`. Hypothesis refuted.

This PR touches only `shapes.rs`, `shapes_tests.rs` and a changelog fragment — nothing near `buffer/view.rs` — which is what makes the flip suspicious rather than obviously a genuine improvement.

Since you know what the cache does to the call graph better than I can reconstruct: is this a real coverage gain, or does the walk lose precision on the new indirection? If it is real, deleting the entry is right and I'll land it. Everything else here is validated and ready.

proggeramlug:
**Resolved — I found the mechanism, and it's an artifact. Merging.**

The `VIEW_REGISTRY` flip is a plain identifier collision, not a coverage gain.

`VIEW_REGISTRY` is tier `[core/T]`, which the gate classifies using the **loose `IDENT`** graph — `reachable_text(IDENT)`, where *any identifier* in a reachable body propagates the walk, not just calls. And past depth 0 the walk abandons module resolution; the code says so itself:

> Root-level names are pinned to the registration's file(s); deeper hops are not (**nothing in the text says which module a call resolved to**).

So: this PR adds a field named `lookup`. `crates/perry-runtime/src/buffer/view.rs` defines a free function named `lookup`, whose body is literally `VIEW_REGISTRY.with(|r| r.borrow().get(&view_ptr).copied())`. The identifier `lookup` enters the reachable set, the walk pulls in *every* body of that name from any file at `MAX_SCANNER_DEPTH = 3`, view.rs's text lands in `by_file[view.rs]`, and `VIEW_REGISTRY` then matches by substring.

Confirmed by experiment: renaming the field `lookup` → `lookup_ways` and changing nothing else flips it straight back to `UNCOVERED` and the gate exits 0. Deleting the frontier entry would have retired a real piece of GC custody debt on a name coincidence.

## Fixed while landing

1. **`lookup` → `lookup_ways`** — the rename above.
2. **`test_clear_shape_table` dropped every descriptor box while invalidating nothing**, so a clear-then-lookup in any test derefs a freed `Box<ShapeDescriptor>`. Added the invalidation call.
3. **File size** — extracted the `#[cfg(test)]` helpers into `shapes_test_support.rs` via the `#[path]` child-module pattern this file already uses for `shapes_tests.rs` (2057 → 1933 lines, under the cap).
4. **Re-applied #8918's census-pinned carrier literal.** This branch was cut before #8918, and merging main brought the two together such that the tree ended up back on the `if is_carrier` form — landing as-is would have **re-broken `lint` on main**, which is exactly what a full gate run is for.
5. Renamed the fragment `8902-…` → `8917-…` to match the PR.

## Validation

Runtime **2762/0** (`RUST_TEST_THREADS=1`). `scripts/run_lint_gates.sh`: **all 53 gates pass**. Under `PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1` the failing set is identical to main's — 16 both ways, zero symmetric difference.

Earlier I verified the sabotage claim rather than trusting it (removing the bump in `remove_descriptor_and_reverse_indices` turns the shapes suite red) and enumerated every path that can drop or replace a box; the fresh-id insert correctly needs no invalidation, since a fresh id can only collide with a *different* id in its way and the way compares the id.

The 12.2% is not re-measured here.

PR #8921, opened 2026-08-28
perf: field-value call arguments keep the proven-this clone; subclass push arm ahead of the tracked resolver (wolf-ecs −11.2% / −11.9%)

proggeramlug:
## Summary

Two changes, measured on the Mac mini reference box (`taskpolicy -t 0 -l 0`, 11 alternating process pairs, wolf-ecs add/remove and entity-cycle from noctjs/ecs-benchmark, semantics probe byte-identical to Node):

1. **Field-value arguments to sibling methods keep the proven-`this` clone** (`collectors/ptr_shape.rs`). The this-flow walker rejected a method as a `$pshape` candidate whenever an internal `this.m(...)` call's *arguments* mentioned `this` at all — so `this._archChange(this._ent[id], i)`, `this._hasComponent(this._ent[id].mask, i)` and `this._crEnt(this.entID)` disqualified wolf-ecs `addComponent`, `removeComponent` and `createEntity`, which then ran their public bodies re-proving `this` at every property/element/method site (≈14k instructions for ~20 source lines; a flat 54% of the add/remove profile). A declared-field read hands the callee a value, never the receiver, and `expr_this_safe` already rejects a bare `this`, a `this`-capturing closure and a non-field read on its own; arguments are now vetted by it alone. Test pins both directions (field value admitted, bare `this` still rejected).
2. **Object-backed subclass push arm ahead of the tracked resolver** (`array/push_pop.rs`). #8897's `field_push_local_bind` turns `this.packed.push(x)` into a local `ArrayPush` whose complete fallback is `js_array_push_f64_spec`; for an object-backed `class Archetype extends Array` that entry paid the tracked resolver (a guaranteed miss on a `GC_TYPE_OBJECT` header) and then `js_array_push_f64`, which paid it again before reaching the dense subclass arm. Both entries now ask the dense arm first, off the header tag; a test-only probe counter on `try_read_tracked_gc_header` pins that they reach it with exactly the fused u31 entry's probes (none).

| window | baseline | add/remove | entity-cycle |
|---|---|---:|---:|
| 2 s (`targetMs = 2000`) | main `77b994f6b` (#8890's base, pre-#8897) | 0.4594 → 0.4081 **−11.2%** (11/11) | 0.3843 → 0.3388 **−11.9%** (11/11) |
| 50 ms (harness tail) | main `f9890759c` | 0.4701 → 0.4080 **−13.2%** (11/11) | 0.9421 → 0.3390 −64% (11/11; warm-up dominated, see below) |

Note on methodology: the 50 ms tail window is dominated by warm-up. #8897 introduced a cold-start ramp on entity-cycle (steady state unchanged, but calls 2…N run ~3× slower for thousands of calls; `PERRY_GC_DIAG` shows ~224k allocations / 9.4 MB per 300 calls where the previous main allocated nothing) — reported on #8897 with probes; this PR does not address it, which is why the 2 s window against the pre-#8897 main is the number to read.

## Verification
- rebased onto main `82c961f99` (#8923 carries the `-D warnings` cleanups this PR had briefly included)
- codegen 1330/1330 (`cargo test -p perry-codegen --lib`); runtime `array::` / `typed_feedback::` 320/320 (`--test-threads=1`); workspace `cargo check --all-targets` under `-D warnings` with the host-compatible exclusions; file-size, GC store-site inventory, addr-class, raw-handle (`--no-raise-vs origin/main`), local-binding audits all green.
- `addComponent$pshape` / `removeComponent$pshape` / `createEntity$pshape` now exist in the compiled wolf-ecs module and are the hot bodies in the profile.

### DIFF
--- a/changelog.d/8921-proven-this-call-args-subclass-push.md
+++ b/changelog.d/8921-proven-this-call-args-subclass-push.md
@@ -1,0 +1,8 @@
+Compiler and Array performance: a method that passes a declared field's value
+to a sibling method (`this.m(this.items[i])`) keeps its proven-receiver clone
+instead of re-proving `this` at every site of its public body; and the
+generic and spec Array push entries append to an object-backed Array
+subclass through its dense fast path before consulting the tracked
+allocator resolver. wolf-ecs (noctjs/ecs-benchmark) on the Mac mini
+reference box, 2-second window: add/remove -11.2%, entity-cycle -11.9%,
+11/11 paired wins each; semantics probes byte-identical to Node.
--- a/crates/perry-codegen/src/collectors/proven_this_routing_tests.rs
+++ b/crates/perry-codegen/src/collectors/proven_this_routing_tests.rs
@@ -1014,6 +1014,90 @@ fn guarded_pshape_call_site_is_preceded_by_a_shape_id_guard() {
     }
 }
 
+/// A method that hands a declared field's VALUE to a sibling method
+/// (`this.scale(this.value)`) passes nothing but that value: the receiver is
+/// not leaked, so the caller keeps its proven-`this` clone. Before, any
+/// mention of `this` inside an internal call's arguments rejected the caller
+/// outright — wolf-ecs `addComponent(this._ent[id], i)`-style calls lost their
+/// clone and re-proved `this` at every site of the public body.
+#[test]
+fn field_value_arguments_to_sibling_methods_keep_the_proven_this_clone() {
+    let mut counter = counter_class();
+    counter.methods.push(func(
+        92,
+        "scaleByValue",
+        Vec::new(),
+        Type::Number,
+        vec![Stmt::Return(Some(call(
+            Expr::This,
+            "scale",
+            vec![this_get("value")],
+        )))],
+    ));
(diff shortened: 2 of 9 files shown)


proggeramlug:
Merged as part of a batch with #8924 and #8925, validated together on top of current `main`.

**One fix pushed to this branch.** It deleted `test_per_object_layout_present` and `test_young_layout_records` from `gc/layout_tables.rs` to satisfy `-D warnings`. #8923 landed the *other* resolution of the same warning — restoring their regression-test callers — so the two collide: the merge is clean, but the result does not compile (`cannot find function test_per_object_layout_present`). Restored the helpers; #8923's direction is the right one, since they assert real layout behaviour and deleting them trades a warning for lost coverage.

Worth noting this is exactly the failure a per-PR check misses: each PR is individually fine, and only the combined tree is broken.

**Validation (whole batch)** — hir 353/0, transform 115/0, codegen 1330/0, runtime 2765/0, stdlib 124/0; `scripts/run_lint_gates.sh` **all 55 gates pass including the compile tier** (`cargo check --workspace --all-targets` under `-D warnings`, and `clippy`).

The −11.2%/−11.9% wolf-ecs numbers are not re-measured here.

Issue #8929, opened 2026-08-28
run_lint_gates.sh reports a permanent false failure for gates whose args carry a ${{ }} expression


proggeramlug:
`scripts/run_lint_gates.sh` derives its command list from `test.yml` literally. #8928's step is backslash-continued:

```yaml
python3 scripts/ci_cargo_test_shard.py --package perry \
  --total-shards "\${{ fromJSON(needs.plan.outputs.plan).cargo_test_perry.total }}" \
  --validate
```

Locally the expression is not substituted, so the gate runs with `--total-shards ""` and fails with `invalid int value: ''`. The gate itself is fine — `--self-test` passes and `--validate` exits 0 for every valid total (2, 4, 6).

**Why it is worth fixing:** the runner now reports `1 of 57 FAILED` for everyone, permanently. A local gate runner that is always one-red trains people to skim past reds, which is precisely the habit this script exists to prevent — its own header documents two occasions where hand-picked gate runs let `main` go red.

**Why the obvious fix does not work:** the extractor emits only the *first* line of a multi-line `run:`, so the `\${{` never appears in the extracted command and a filter on the extracted text cannot see it. The fix has to join backslash continuations first, then skip commands containing `\${{` (naming them as skipped, not hiding them). That edit lives inside the script's quoted heredoc, which is easy to break — I tried and backed it out rather than ship something unverified.

Found while batch-merging #8926/#8927/#8928.

PR #8928, opened 2026-08-28
ci: shard full-tier perry integration tests

proggeramlug:
## Summary

- keep every package selected by `ci_test_scope.py --full` in the existing `cargo-test` job, but run the `perry` bin/unit target separately from its integration inventory in the full tier
- derive every `perry` integration target from Cargo metadata and assign it exactly once across an 8-way deterministic, count-balanced matrix
- make `full-suite-gate` depend on the aggregate matrix result with `fail-fast: false`, explicit 120-minute shard bounds, per-target live logs, and read-only shared-cache restores
- promote labelled full-tier PRs to the complete cargo-test scope and document/self-test the planner policy

Release run 33123474583 spent 164.9 minutes on the first 93 of 246 integration targets after the one-time build. The current inventory has 247 targets; the matrix assigns them 31/31/31/31/31/31/31/30, leaving substantial cold/low-hit wall-time headroom below the former 180-minute serial bound.

No version bump is included.

## Validation

- `python scripts/ci_cargo_test_shard.py --self-test`
- `python scripts/ci_cargo_test_shard.py --package perry --total-shards 8 --validate`
- independent union/duplicate check against `cargo metadata` (247 targets, each exactly once)
- `python scripts/ci_plan.py --self-test`
- `python scripts/gc_gate_wiring_check.py --self-test`
- `python scripts/gc_gate_wiring_check.py`
- `actionlint .github/workflows/test.yml`
- `python -m py_compile scripts/ci_plan.py scripts/ci_cargo_test_shard.py`
- `git diff --check`

Closes #8914

## Summary by CodeRabbit

* **New Features**
  * Full validation runs now execute Perry integration tests across eight deterministic, balanced CI shards.
  * CI planning now reports Perry integration-test shard assignments for full-tier runs.
* **Bug Fixes**
  * Full-tier package testing now avoids duplicating Perry integration tests.
  * Test coverage and shard assignments are validated automatically.
* **Documentation**
  * Updated CI tier documentation to describe Perry test sharding and full-suite gating.

PR #8932, opened 2026-08-28
ci: shard the GC representation matrix

### LINKED ISSUE #8915: release blocker: gc_repsel_matrix exceeds the 90-minute gc-stress budget
## Release evidence

The v0.5.1519 r26 `gc-stress` job exhausted its 90-minute limit in the representation-selection matrix.

- Candidate: `7c9f6016987ea9fdce2d253fb8666428cf0c6d24`
- Full CI: https://github.com/PerryTS/perry/actions/runs/33123474583
- Job: https://github.com/PerryTS/perry/actions/runs/33123474583/job/98695926869

Timeline:

```text
release build completed in 11m 41s
oracle: node 26.5.1 x 80 corpus files
warming the auto-optimize archive
compiling 80 files x 10 compile-env groups (jobs=4)
... no completion report ...
The operation was canceled at the 90-minute job limit
```

No matrix JSON verdict was produced, so Full CI cannot distinguish a slow matrix from a GC regression and correctly fails.

## Reproduction

```bash
cargo build --release \
  -p perry \
  -p perry-runtime \
  -p perry-stdlib \
  -p perry-runtime-static \
  -p perry-stdlib-static

time ./scripts/gc_repsel_matrix.sh \
  --no-build \
  --arms all \
  --json gc-repsel-matrix.json
```

## Required fix

Make the complete 80-file x 10-environment matrix finish within the bounded Full-tier gate. Prefer deterministic matrix sharding and reusable artifacts/caches over dropping arms or increasing one job's timeout. Add progress identifying the active file/environment and preserve partial evidence on interruption.

## Acceptance criteria

- All corpus files and required compile-environment arms remain covered.
- The complete result finishes under 90 minutes with headroom, or a fan-out/fan-in replacement completes within equivalent bounded jobs.
- Every shard uploads evidence and a missing shard fails the aggregate.
- `gc-stress` remains a blocking input to `full-suite-gate`.

proggeramlug:
## Summary

- build the release compiler/runtime once and reuse the artifact across GC matrix workers
- split sweep/full runs into four stable 20-entry corpus shards while retaining all 22 arms; keep the PR subset as one shard
- add a strict fan-in that rejects missing, duplicate, interrupted, or incomplete shard coverage before applying corpus-wide liveness
- report the active file/environment and preserve an append-only progress journal when a shard is interrupted
- keep gc-stress as a blocking dependency of full-suite-gate

No package or crate version is bumped.

## Validation

- cargo build --release -p perry -p perry-runtime -p perry-stdlib -p perry-runtime-static -p perry-stdlib-static
- filtered end-to-end matrix witness: evac_minor and shipped_default both PASS with complete JSON/progress evidence
- forced TERM: interruption journal preserved and complete JSON correctly withheld
- ci_plan, merge, liveness, and GC wiring self-tests
- actionlint .github/workflows/test.yml
- bash syntax and git diff checks

Closes #8915

## Summary by CodeRabbit

- **CI Improvements**
  - GC stress testing now runs across parallel shards for sweep and full validation tiers.
  - Shared build artifacts reduce repeated compilation during stress-test runs.
  - Sharded test reports are combined into a single validated report.
  - Coverage, duplicate results, configuration consistency, and liveness are checked before results are accepted.
  - Progress information and partial evidence are preserved when interrupted or unsuccessful.
- **Documentation**
  - Updated CI tier documentation to reflect the new GC stress-test frequency.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -653,6 +653,7 @@ jobs:
         if: ${{ !cancelled() }}
         run: |
           ./scripts/gc_repsel_matrix.sh --self-test-liveness-parser
+          python3 scripts/gc_repsel_matrix_merge.py --self-test
           python3 scripts/gc_matrix_liveness_check.py --self-test
           python3 scripts/gc_matrix_liveness_check.py --check-registry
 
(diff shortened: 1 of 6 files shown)


proggeramlug:
Audited and merged. Sharding a gate is exactly where gates quietly stop being able to fail, so I went at the fan-in rather than the plumbing.

**I sabotage-tested the strict fan-in independently**, building fixtures from the script's own `expected_arms`/`corpus_from_manifest` (22 arms, 80-entry corpus) and mutating them. Every coverage failure is caught:

| mutation | verdict |
|---|---|
| all 4 shards present (control) | ACCEPTED |
| one shard **missing** | REJECTED |
| one shard **duplicated** | REJECTED |
| a shard marked `complete: false` | REJECTED |
| a shard with **cells truncated** | REJECTED |
| a shard **missing one arm** entirely | REJECTED |

And the liveness half holds through the CLI: with every `requires=move` arm reporting `evacuated: 0`, the merge marks all 11 of them **INERT** and exits non-zero. That is the #7025 failure mode — a cell passing having run zero copying minors — and it is caught at the corpus level, which is the right place given the shards run `--defer-liveness`.

Worth saying explicitly, since it is the subtle part: `gc-stress` runs under `if: always()`, which is *necessary* for a strict fan-in (it has to run to notice a shard that never reported), and it means a dead shard only reddens the gate because the merge script catches it. That load is carried entirely by `--expect` plus the coverage checks above — which is why I tested them rather than trusting the self-test. Retaining the historical `gc-stress` job id also keeps the existing `full-suite-gate` wiring pointed at the real fan-in rather than at a job that no longer gates anything.

**Validation** — `ci_plan --self-test`, `gc_gate_wiring_check` (both forms), `gc_matrix_liveness_check --self-test` and `--check-registry`, `gc_repsel_matrix_merge --self-test`, the liveness-parser self-test, and `bash -n` all pass. `scripts/run_lint_gates.sh`: 57 of 58 pass including the compile tier (`-D warnings` and `clippy` green); the 58th is the pre-existing `${{ }}` substitution artifact from #8928, filed as #8929.

`actionlint` exits 1 on three shellcheck info/style items (SC2086, SC2012, SC2129) — **byte-identical on `origin/main`**, so pre-existing and not this PR's.

PR #8933, opened 2026-08-28
perf(codegen): inline captureless some loop; codegen-time module-const literal fold (ECS round 4, +5.2%)

proggeramlug:
Two codegen/driver mechanisms from the ECS round-4 chain (each screened with paired alternating runs on the idle Mac mini and confirmed over 15 pairs on the `codehz/ecs` "5k entities: 3 commands each + sync" row; Node 26.5.1 = 1.762 ms on the same host). These were pushed onto #8916 after it had already merged with only its first commit, so they never reached main — this PR re-cuts them from current main. Write-up: `secret-tests/ecs-suite/PERRY_ECS_FOLLOWUP_2026-08-27_CLAUDE.md`.

|---|---|---|---|---|
| inline `arr.some(capturelessArrow)` loop | 3.960 ms | 3.814 ms | +3.85%, 9/9 | **+3.70%, 15/15**, 30 oracles (`r4b-confirm.json`) |
| module-const literal fold at codegen entry | 3.600 ms | 3.546 ms | +1.49%, 9/9 | **+1.49%, 15/15**, 30 oracles (`r4g-confirm.json`) |

- **codegen: `arr.some(capturelessArrow)` as an inline loop with a direct body call** (`lower_captureless_some_inline`). `js_array_some_captureless` decided the receiver once and then, per element, re-resolved the head from its root, NaN-boxed the receiver and called the body through the function pointer (4.5% self). The lowering makes the same one-time decision on the same live bits — `GC_TYPE_ARRAY` head, not forwarded, no indexed descriptors, the sticky `PERRY_ARRAY_INDEX_FAST_PATH_INVALIDATED` byte clear, `length <= capacity` — and runs the loop inline: the head is re-read from its root every iteration (a forwarded head goes through the new `js_array_live_head` export), indices past the live length and holes are skipped, the arrow's body symbol is called directly with as many of `(element, index, receiver)` as it declares, and `true`/`false` results decide inline with `js_is_truthy` for anything else. Every receiver the loop does not admit takes the runtime helper, which stays the fallback. Pinned by `captureless_inline_some_passes_the_callback_body_directly`.
- **driver: module-level const literals fold into their reads after the transform phase, before codegen** (`perry_transform::module_const_fold`, run from `run_pipeline.rs` once every module is transformed). `export const COMPONENT_ID_MAX = 1023` is a module-scope immutable let and every read of it is a `LocalGet` the typed-ABI clone rules cannot type, so a one-line predicate such as `isComponentId` (`id >= 1 && id <= COMPONENT_ID_MAX`) was refused its `i1` clone and every call ran a module-global load plus the dynamic tag-coercion compare on both operands. It is deliberately not a pipeline pass: folded, those predicates become self-contained and the cross-module inliner harvests them — inside the pipeline that consumed callers' inline budgets (`world.set` lost `resolveSetOperation`, −43%). Run after all harvests are taken, no inlining decision moves; the fold precedes the HIR trace and the object-cache fingerprint so both describe the tree codegen consumes. Admission and the TDZ rule are pinned by the module's unit tests; the explain-lowering report confirms the `i1` clone is now admitted.

Tests: codegen lib (1329) + `native_proof_regressions` (280), transform lib (119), runtime suite; lint gates and merge-base ratchets replayed locally.

## Summary by CodeRabbit

* **Performance**
  * Improved eligible `Array.prototype.some` calls by executing captureless callbacks inline.
  * Preserved fallback behavior for arrays that do not meet optimization requirements.
  * Added module-level folding of safe primitive constants before code generation, while preserving reassignment and temporal-dead-zone behavior.

* **Documentation**
  * Documented inline `Array.prototype.some` execution, constant folding, fallback behavior, and benchmark improvements.


proggeramlug:
Audited and merged.

**On the inline `some` loop**, the hazard is that the callback is user JS and therefore a collection point. It is handled: the head is re-read from its root every iteration, with a forwarded head resolving through the new `js_array_live_head`, so a callback that collects or grows the array cannot leave the loop walking a stale head. The admission decision is made once on live bits (`GC_TYPE_ARRAY`, not forwarded, no indexed descriptors, sticky invalidation byte clear, `length <= capacity`) and every receiver it does not admit keeps the runtime helper. `rooting::migration_ledger` structurally fails the build if this module reaches for a raw temp root, which is a nice belt-and-braces.

**On the const fold**, the crux is TDZ and it is explicitly right: reads *before* the declaration are left alone so they keep throwing, and admission requires a top-level immutable `let` with a number/integer/string/boolean **literal** initializer, re-scanned for `LocalSet`/`Update` even though a `const` cannot be reassigned. Dropping the folded id from closure capture lists is consistent — the value it would have captured is the literal.

The reasoning for running it outside the pipeline is also worth keeping: folded early, the predicates became inline-harvestable and consumed callers' budgets, costing `world.set` its `resolveSetOperation` at −43%. Running it after every module is transformed changes no inlining decision.

**Validation** — hir 355/0, transform 119/0, codegen 1331/0, runtime 2765/0, `native_proof_regressions` 280/0; `scripts/run_lint_gates.sh` 57 of 58 including the compile tier green, the exception being the pre-existing `${{ }}` artifact (#8929).

The +3.70% / +1.49% figures are not re-measured here.

proggeramlug:

PR #8934, opened 2026-08-28
perf(runtime): lean Map/Set lookup lanes; empty-array pop fast path; single-pass length-0 re-arm (ECS round 4, +5.6%)

proggeramlug:
Three runtime mechanisms from the ECS round-4 chain (each screened with paired alternating runs on the idle Mac mini and confirmed over 15 pairs on the `codehz/ecs` "5k entities: 3 commands each + sync" row). They were pushed onto #8916 after it had already merged with only its first commit, so they never reached main — this PR re-cuts them from current main. Write-up: `secret-tests/ecs-suite/PERRY_ECS_FOLLOWUP_2026-08-27_CLAUDE.md`.

|---|---|---|---|---|
| lean Map numeric lookup lane | 3.814 ms | 3.686 ms | +3.37%, 9/9 | **+3.30%, 15/15**, 30 oracles (`r4c-confirm.json`) |
| small-Set member scan before the side-table | 3.684 ms | 3.617 ms | +1.85%, 9/9 | **+1.80%, 15/15**, 30 oracles (`r4d-confirm.json`) |
| empty-array `pop` fast path + single-pass length-0 re-arm | 3.617 ms | 3.599 ms | +0.48%, 9/9 | **+0.48%, 15/15**, 30 oracles (`r4f-confirm.json`) |

- **Map lookups run a lean numeric lane before the general `find_key_index`.** The function carried the string-hash, pointer-index, hashed-numeric and generic-compare paths in one body; a PC histogram of the profile put a third of its 5.5% self time on the prologue/epilogue those cold paths force (eight callee-saved GPRs and four FP registers on arm64) and half on the dense-key range tests. The two shapes the numeric side-table exists for — a plain (untagged, non-NaN, non-zero) number against a small map's entries by bit identity, or against the dense integer range table — now run in an always-inlined `find_key_index_hot` inside `js_map_get`/`js_map_has`/`js_map_set`'s callers; everything else goes to the outlined `find_key_index_cold`. A dense-range miss stays definitive for its span; a key outside the span, a tagged, zero or NaN key, and every string or pointer key take the cold path unchanged. `hot_lookup_lane_agrees_with_the_cold_path_on_every_key_shape` pins hit/definitive-miss/out-of-span/-0/NaN/tagged shapes on both a small and a dense map.
- **Small-Set lookups scan the members before the side-table.** `find_value_index` answered every `Set.has`/`Set.add` through the thread-local `SET_INDEX`: a hash of the set address to reach its table, then a hash of the value — two probes for sets that in the hot shapes hold three or four numbers (`componentTypeSet.has` was 7.5% of the in-place update path). A plain number against a set of at most eight elements is now decided by reading `elements[0..size)` — exactly the membership, since delete compacts and add normalises `-0`, and no tagged value equals a number. Larger sets, tagged/zero/NaN values and every string keep the side-table, outlined. Pinned by `small_set_scan_lane_agrees_with_the_side_table_on_every_value_shape`.
- **`pop()` on an empty plain array answers from the header fast path; `length = 0` re-arms an all-pointer head in one registry pass.** The pop fast path required a non-empty array, so the drained pool's `pool.pop() ?? []` fell through the whole generic tower to reach the same `length == 0` return; with the descriptor flag excluded, `Set(O, "length", 0)` is a no-op and there is no index to Get or Delete. `rebuild_array_layout` on `length = 0` of an all-pointer head ran the zero-slot rebuild and then `layout_init_all_pointer_slots`, which clears the same bit, forgets the same two record kinds and sets the state — two passes over the layout registries per `pooled.length = 0`; the re-arm now runs alone with an identical end state (the round-U truncate test additionally asserts no per-object record survives). The pop/push tests move to `array/push_pop_tests.rs` for the 2000-line gate.

Tests: runtime suite (2763) incl. `map::`, `set::`, `array::`, `layout_trace`; lint gates and merge-base ratchets replayed locally.

## Summary by CodeRabbit

  * Improved lookup performance for numeric keys in maps and small sets.
  * Optimized `pop()` on empty arrays.
  * Improved array layout handling when reducing pointer-only arrays to zero length.
  * Benchmark results show up to a 3.3% improvement in representative workloads.

* **Bug Fixes**
  * Empty-array `pop()` now correctly returns `undefined` while preserving array length.
  * Improved handling of array holes after resizing.


proggeramlug:

PR #8935, opened 2026-08-28
perf(gc): inline hot-TLS values in HotTls; leaf write-barrier entry (ECS round 4, +2.7%)

proggeramlug:
Two GC/TLS mechanisms from the ECS round-4 chain (each screened with paired alternating runs on the idle Mac mini and confirmed over 15 pairs on the `codehz/ecs` "5k entities: 3 commands each + sync" row). They were pushed onto #8916 after it had already merged with only its first commit, so they never reached main — this PR re-cuts them from current main. Write-up: `secret-tests/ecs-suite/PERRY_ECS_FOLLOWUP_2026-08-27_CLAUDE.md`.

|---|---|---|---|---|
| inline hot-TLS values in `HotTls` | 3.545 ms | 3.503 ms | +1.18%, 9/9 | **+1.17%, 15/15**, 30 oracles (`r4h-confirm.json`) |
| leaf write-barrier entry | 3.505 ms | 3.450 ms | +1.59%, 9/9 | **+1.56%, 15/15**, 30 oracles (`r4i-confirm.json`) |

- **runtime: the hottest small thread-local values live inline in `HotTls`.** A hot-TLS slot and a named pointer field both resolve as TSD base → `HotTls` → slot pointer → value; PC histograms put the remaining self time of the write barrier, `js_map_clear` (10k calls/frame, both hot offsets on its two TLS probes), `is_registered_box_ptr` and `array_prototype_addr` on that dependent chain rather than on anything they compute. Small `Copy` values with a `const` initial state can live in `HotTls` itself (TSD base → `HotTls` → value), so the barrier's one-entry dirty-page cache, the memoized `Array.prototype`/`Object.prototype` rows and the three box-pointer caches now do; the generic slot mechanism is unchanged for everything else, the collector's root rewrite of the prototype rows walks the inline cells exactly as it walked the slot, and the custody audit's frontier entries for the removed box-cache statics are deleted (the deletion is the receipt).
- **gc: the write barrier's dirty-page hit returns from a leaf entry.** Every pointer store into an old object reached `js_write_barrier_slot_validated_parent`, which made two out-of-line calls before anything was decided — `decode_heap_addr` for the child, and `incremental_mark_barrier_value`, whose "no cycle anywhere" test sat inside the callee — and then entered the outlined `write_barrier_decoded_parent` (six callee-saved registers) to run the one-entry dirty-page compare that answers the second and third push into the same bucket. The tag decode and the idle test now inline (their slow arms are cold, out of line), and the cache test is hoisted into the entry ahead of the outlined body (`gc/barrier/leaf.rs`), so a hit is a leaf path. Counters and the remembered set built are unchanged; pinned by `validated_parent_entry_answers_a_cached_dirty_page_store_before_the_body`.

Tests: runtime suite (2764) incl. `gc::tests::barrier` (411, serial), `runtime_roots`, `box::`; custody audit + lint gates and merge-base ratchets replayed locally.

## Summary by CodeRabbit

* **Performance**
  * Improved runtime and garbage-collection performance by optimizing frequently accessed thread-local values.
  * Accelerated write-barrier handling for cached dirty-page cases.
  * Benchmarks show improvements of up to 1.6% in relevant workloads.

* **Bug Fixes**
  * Added regression coverage for validated-parent write-barrier behavior, including cached-page handling and edge cases.


proggeramlug:

PR #8936, opened 2026-08-28
perf(object): kill the per-element key scan behind dynamic property access (−61%)

proggeramlug:
Killed the per-element linear key scan behind dynamic string-keyed property
access: **−61% on baseline dynamic property throughput** (768 → 303 ms on the
`bench_dynamic_property_keys` overwrite loop; node on the same host: 12 ms).

The `[[Set]]`/`[[Get]]` fallback walks and the `delete` path each carried their
own copy of the same loop: `for i in 0..key_count { js_array_get(keys, i) +
string compare }` — the full JS-facing array accessor (pointer cleaning,
typed-array and buffer registry probes, descriptor gates) per element, per
property operation. The pointer-keyed read plan in front never hits for a
computed key, because `o["k" + i]` allocates a fresh key string every
evaluation. Counted with a temporary in-runtime counter: **90.8 million
`js_array_get_f64` calls for 1.5 million property operations** (~60 per
access). After this change: 15.1 M, all of it in `delete`'s array compaction
rather than lookup.

The scans now go through one shared helper (`keys_find_slot_by_bytes` /
`_by_key_ptr`): the shape hash index (`shape_slot_lookup`, content-validated)
answers in O(1) when present, with a raw dense-slot linear scan (no per-element
accessor) as the fallback and correctness backstop.

Two hazards were found by testing and are baked into the design:

* **Consult-only (`build=false`).** A delete drops the shape index; rebuilding
  it on the next access to use it once **doubled** delete-heavy time
  (1570 → 3064 ms measured) while the call counter barely moved — the time went
  to rebuilds, not scans. These sites therefore only consult an index the write
  path already maintains incrementally; churny receivers fall back to the raw
  scan instead of thrashing rebuilds. Final: delete-heavy 1497 ms vs 1433
  baseline (within the ±10% noise band of that metric), overwrite keeps the
  full win.
* **Garbage-length tolerance.** The old loops compared LENGTHS first
  (`js_string_key_matches`), so a `key` pointer that is not a valid string
  header was a harmless mismatch. The first helper version built a slice from
  that length and panicked in an unrelated stream test with `range start index
  2613749136200 out of range`. The helper now sanity-checks
  `byte_len <= capacity && < 2^28` before slicing and otherwise falls back to
  the length-guarded compare.

Also switches the shape-scanner probe memo (#8899) off std's SipHash: perf put
`RandomState::hash_one::<&(usize, bool)>` at **7.0% of total samples** on this
workload. The key folds to one word (`addr | carrier_bit`; addresses are
8-aligned so bit 0 is free) under `PtrHasher`.

Suites: macOS **2762 passed, 0 failed** (complete); Linux x86_64 2654 passed,
0 failed with the `node_stream` error-path family excluded — that family
aborts identically on clean main (verified by stash), a pre-existing
Linux-specific failure reported separately.

---

### How the number was found

The 63× gap to node on dynamic property access (`benchmarks/bench_dynamic_property_keys.ts`, #8901) resisted three profile-guided attempts — each targeted a plausible cost that turned out not to be the term (documented in #8899/#8917). What settled it was **counting instead of sampling**: a temporary in-runtime counter showed 90.8 M array-accessor calls for 1.5 M property operations, and the factorization (~60/access ≈ half the 500-key object per lookup) pointed directly at the linear scans.

### Progression on the benchmark (same host as node, load-stamped)

| variant | delete-heavy | overwrite | counter |
|---|---:|---:|---:|
| baseline | 1433 ms | 768 ms | 90.8 M |
| + scan kill (build=true) | 1655 ms | 303 ms | 16.0 M |
| + delete path, build=true | **3064 ms** ⚠ | 545 ms | 14.5 M |
| **+ consult-only (this PR)** | **1497 ms** | **303 ms** | 15.1 M |
| node | 30 ms | 12 ms | — |

The ⚠ row is why `build=false`: rebuild-per-delete thrash, caught because every measurement line carries its load stamp and the counter separates scan cost from rebuild cost.

Gap to node on this loop: **63× → ~25×**. The remaining weight is the dyn-IC miss path doing a fully-rooted `[[Set]]` per computed-key write (key interning is the known follow-up), and delete's compaction internals.

### Review focus

1. `keys_find_slot_by_key_ptr`'s tolerance contract — the old loops accepted invalid key headers (length-compare-first); the helper preserves that via the sanity gate + fallback. The panic it prevents is documented inline.
2. `build=false` is load-bearing for delete-churn workloads; the comment carries the measured 2× regression that motivates it.
3. The memo-hasher change re-keys `(usize, bool)` → `usize` with the carrier bit folded into bit 0 — sound because arena addresses are 8-aligned.

Suites: macOS **2762/2762**; Linux 2654/2654 excluding the pre-existing `node_stream` platform failure (stash-verified on clean main, filed separately).

## Summary by CodeRabbit

- **Performance**
  - Improved dynamic property access performance, reducing baseline lookup time by approximately 61%.
  - Optimized property updates, reads, and deletions, especially for objects with many keys.
  - Reduced overhead in repeated dynamic-key operations while preserving existing behavior.


proggeramlug:
Gates cleared in `e40ab6ba6` — all three were mine to fix, thanks for parking it rather than wrestling my code:

1. **addr-class ratchet** — the open-coded `< 0x10000` is now the canonical `is_above_handle_band`. Worth noting this was a real bug, not just lint: my floor was an order of magnitude below `HANDLE_BAND_MAX`, so handle-band ids would have been dereferenced as string headers.
2. **string-payload ratchet** — new shared `string::string_header_payload` helper; covers both my added site and the pre-existing one in `key_content_hash_impl`, so the ratchet count goes down, not sideways.
3. **2000-line cap** — the seam that fought you was the right one to avoid: I moved only the two new lookup helpers to `object/key_lookup.rs` (they're self-contained; callers reach them via a re-export, so `put_value`/`delete_rest` are untouched). `mod.rs` is at 1987.

Suite: 2762 passed, 0 failed (full macOS run). No `git add -A` involved — files staged individually.

proggeramlug:
Merged, rebased onto current `main`. It broke three gates, all fixed here:

- **2000-line cap** — `object/mod.rs` went 74 lines over. Split the keys-array lookup helpers into `object/keys_lookup.rs` via the `#[path]` child-module pattern the tree already uses (1915 + 177 lines).
- **addr-class ratchet** — the new `(key as usize) < 0x10000` is a bare handle-floor site. Now `addr_class::is_above_handle_band`. Worth noting `HANDLE_BAND_MAX` is `0x100000`, so the original literal let part of the handle band through; the canonical predicate is a small correctness gain, consistent with CLAUDE.md's "value < 0x100000 = handle". I deliberately did **not** use `is_plausible_heap_addr` — it adds `is_valid_obj_ptr`, which could reject a real key and silently return "not found", and this helper's callers are documented as tolerating a `key` that is not a valid header.
- **string-payload ratchet** — the new `keys_find_slot_by_key_ptr` open-coded the `StringHeader` payload offset a second time. Both key helpers now share one `string_header_payload`, so the offset appears once and the count returns to baseline.

The two helpers move together deliberately, so the shared offset sits with both callers rather than straddling the split.

**Validation** — runtime 2771/0 (`RUST_TEST_THREADS=1`); `scripts/run_lint_gates.sh` 57 of 58 including the compile tier (`-D warnings`, `clippy`), the exception being the pre-existing `${{ }}` substitution artifact (#8929).

PR #8938, opened 2026-08-28
perf(runtime): Map.clear early exit on empty maps; skip side-table probes for small numeric-keyed maps (ECS round 4)

proggeramlug:
One runtime mechanism from the ECS round-4 chain, cut from current main. Measurement on the `codehz/ecs` "5k entities: 3 commands each + sync" row (paired alternating runs on the idle Mac mini) is being taken against current main and will be posted here; suites: runtime (2745) incl. `map::` serial, on the isolated perrymaster gate.

- **`Map.clear` returns early on an empty map and skips the side-tables for small numeric-keyed maps.** Every index `js_map_clear` resets mirrors the entries exactly (insert, delete, clear and the GC rewrites keep them in step), so an already-empty map has nothing to reset — the per-entity `adds.clear(); removes.clear()` of an ECS change set is that case half the time. And the string and pointer side-tables hold an entry only for a string or pointer key the map holds: for a map of at most 16 entries whose keys are all bits-stable numerics (`is_safe_numeric_key`), reading its keys is cheaper than the two thread-local resolutions plus two hash probes that find two empty tables — which the post-4h profile showed was the *whole* cost of `js_map_clear` (both hot offsets on its two TLS probes, ~10k clears a frame on the grouping maps, 1.1% of the frame). Larger maps and any string/pointer key reset the side-tables unconditionally as before. `clear_resets_every_index_whatever_the_key_kinds` pins the numeric-only, empty, string-keyed and past-the-bound shapes (a stale content-hashed index entry would otherwise misroute a re-added key).

## Summary by CodeRabbit

- **Performance**
  - Improved `Map.clear()` performance, especially for frequently cleared small maps.
  - Empty maps now clear immediately.
  - Small maps with numeric keys avoid unnecessary internal cleanup work while preserving correct behavior.

- **Bug Fixes**
  - Ensured all map indexes are reset correctly across numeric-keyed, string-keyed, and larger maps.


proggeramlug:
Merged. Rebased onto current `main`; it conflicted with #8934 in `map.rs` because both add a test at the same spot and the two bodies share one closing brace. Kept **both** tests, each with its own brace and `#[test]`.

Validation: runtime 2770/0 (`RUST_TEST_THREADS=1`), `scripts/run_lint_gates.sh` 57 of 58 including the compile tier green — the exception is the pre-existing `${{ }}` substitution artifact (#8929).

proggeramlug:
Paired measurement (idle Mac mini, 9 alternating pairs, `codehz/ecs` "5k entities: 3 commands each + sync"): control = main `49279e99e`, candidate = this branch on that base: **3.955 → 3.943 ms/op, +0.36%, 9/9** (`r4j-screen.json`, process oracles 18/18). Small, unanimous.

Issue #8175, opened 2026-08-15
perf(codegen): the i32 param rep defeats LLVM shrink-wrapping on fib40's leaf path — 2.3-2.5x available (was: 'IPC collapsed 6.30 -> 1.92')


proggeramlug:
## Summary

#8167 cut `fib40`'s instructions **45.5x** (212.14 G → 4.66 G) but wall time only **13.9x**, because **IPC collapsed from 6.30 to 1.92**.

Perry now retires **3.5x fewer instructions than node** on `fib(40)` and is only **1.36x faster** (0.759 s vs 1.032 s). The remaining gap is a **stall problem, not an instruction-count problem**, and no amount of further instruction removal will close it.

## The measurement

Quiet M1 mini, best-of-5, two independent interleaved runs, both `VERDICT: CLEAN`, perry `499e29627`, node v26.5.1.

| | instructions | cycles | IPC | wall |
|---|--:|--:|--:|--:|
| perry `38cf15336` (pre-fix) | 212.14 G | — | 6.30 | 10.56 s |
| perry `499e29627` (post-fix) | **4.66 G** | — | **1.92** | 0.759 s |
| node 26.5.1 | ~16 G | — | 4.98 | 1.032 s |

**1.92 is by far the lowest IPC in the corpus** — every other row sits at 4.5–6.5. So this is specific to what #8167's fast path emits, not a general property of perry's codegen.

## Why this matters beyond one row

It exposes a blind spot in how this campaign has been measuring. Instructions retired is load-independent, which makes it the right metric on a contended box, and essentially every A/B today used it — including my own remeasure, which reported `fib40 −0.01%` while the row was 55x off its own baseline. But instructions are only a proxy for time, and here the proxy broke: a 45.5x instruction win delivered 13.9x.

Any future claim of the form "N% fewer instructions" on a hot numeric loop should carry cycles or IPC alongside it, or it can overstate the delivered result by 3x.

## Candidate causes, unranked and unverified

- **The range test #8167 emits.** `spec_i32_derived_window` cannot prove containment for `n - 1` (its window is one value wider than the slot), so each recursive call carries a range check plus a cold boxed arm. Two per invocation across ~331 M invocations, and a hard-to-predict branch would hurt IPC exactly this way. **#8171** proposes removing it for this shape via flow-sensitive narrowing — that issue was filed as a 22% instruction residual, but if the branch is also the stall source its real value is larger than 22%.
[...]

Issue #7174, opened 2026-08-01
gc: RewriteStatepointsForGC over managed-pointer SSA — the measured-only road to native-root file-size parity


proggeramlug:
The 2026-07-31/08-01 stackmap campaign (#7108, branch `exp/stackmap-viability`, full write-up in `docs/statepoint-gc-experiment.md`) ends with every road to native-root file-size parity converging on one experiment that has not been run: `RewriteStatepointsForGC` over managed-pointer SSA.

**Why this is the only remaining road, now demonstrated rather than argued:**

- The explicit statepoint bridge is correct and speed-competitive (quiet-host geo −1.1% vs shadow, RSS flat) but its metadata measures **3.83 MB against 150 KB of text recovered** on the real `test-drizzle-pg` app (~25×) after a −59% reduction campaign (audited call effects, noreturn elision, safepoint contract). The remaining records are dominated by base/derived pairs and per-safepoint spill slots that explicit emission cannot avoid.
- The cheap alternative was built and killed: a compact per-function metadata mode measured **10–13× smaller** and was deleted as a proven negative result (`bd066d62b`, deleted at `a95f59e96`). A ten-line churn loop corrupts deterministically under moving minors, bit-identically even with the walker's visits disabled — the mutator reads from-space through stale heap-derived values in optimized SSA. The same module carries 79 `gc.relocate`s under the statepoint backend. **Design law: root metadata without relocation semantics is unsound under an optimizing compiler.** Per-call plain maps only kept the window small enough for probes to pass (that mode is also deleted — its `Register R#N` locations are unrecoverable by construction).

**What the experiment is** (#7108 costed it as option 1, "weeks"):

1. Represent GC-capable values as `ptr addrspace(1)` in SSA — the whole-NaN-box carrier is already proven (#7108 Q1: `bitcast double → i64 → inttoptr`, folds to zero instructions, relocated end-to-end at runtime).
2. Tag functions `gc "statepoint-example"`, let `opt -passes=rewrite-statepoints-for-gc` insert statepoints, results, relocates, and downstream-use rewrites — deleting the hand-emission and its conservative CFG-union liveness.
3. Toolchain: requires `opt` (Homebrew LLVM; Xcode ships none). Verified working end-to-end in #7108. Gate the mode on `opt` presence; this is the distribution decision that issue named, scoped here to research.

[...]

Issue #5391, opened 2026-06-18
codegen: split large modules into parallel codegen units (bound clang peak memory; remove the single-TU wall)


proggeramlug:
## Problem

perry emits **one LLVM translation unit per source module**. A large minified bundle (the 13MB `@anthropic-ai/claude-code` cli.js) lowers to a single **~1.1GB / 28.7M-line** `.ll`, and `clang -c` needs **~13–15GB RSS** to compile that one TU — the hard wall that stops the bundle from producing a native binary. IR-efficiency work (#5334, levers A–D, all merged) trimmed it ~12%, but the floor is structural: one giant TU.

This is the same scaling problem rustc solves with **codegen units** and LLVM with parallel codegen / ThinLTO: split the work into independently-compiled units so peak memory is bounded by unit size, not whole-program size.

## Approach — codegen units (CGUs)

Composition of the bundle `.ll` (measured): function bodies are **76% (~21.8M lines)**; **all module globals + string constants are only ~413K lines (1.4%)**. So duplicating the small shared-global set into each unit is cheap, and each unit shrinks nearly linearly with the function partition.

Render-time partition of the fully-populated `LlModule` (the seam is right before `to_ir()`):

1. Partition `LlModule.functions` (≈92K) into N buckets.
2. Each CGU `.ll` =
   - target triple + attributes + **replicated metadata** (so `!N` refs resolve);
   - the full string-constant + global set, with `private`/`internal`/bare-external **definitions promoted to `linkonce_odr`** (linker dedups; external *declarations* like `external global` replicated as-is). Globals are tiny (1.4%), so this duplication is negligible;
   - the bucket's function `define`s, plus a `declare` for every function it calls that lives in another bucket (reusing `to_ir`'s existing "skip declare if also defined here" logic — generate declares for all functions, drop the locally-defined ones);
3. Compile each CGU with `clang -c` (bounded peak memory; parallelizable), then **`ld -r`** the N objects into one — so `compile_module(...) -> Result<Vec<u8>>` and all callers/tests stay unchanged.

Per-function machinery already co-travels correctly: each function self-carries its `pending_declares` and its IC/rodata globals, and IC-site numbering is already baked into the rendered text, so partitioning doesn't renumber.

## Gating

[...]

PR #8942, opened 2026-08-28
fix(codegen): module-unique names for promoted string constants and the null guard (ELF multi-unit link)

proggeramlug:
## Defect

Coop compiles a Next.js App Route bundle to a shared library on Linux (`perry compile --no-codegen --no-auto-optimize --march generic --output-type dylib`, Ubuntu 24.04, LLVM 22.1.8, GNU ld; Perry main 504013308). Five modules are large enough to be split into codegen units. The final link fails with 2,188 errors of the form

```
ld: 9cab07ed002d7976.o:(.rodata.str1.16+0x78d60): multiple definition of `.str.375'; 71e0de9737178e64.o:(.rodata.str1.1+0x41f2): first defined here
```

— `app-page.runtime.prod.js` ↔ `route.js` (1,607 symbols) and `jsonwebtoken/index.js` ↔ `route.js` (581). `readelf -sW` on those objects: `OBJECT GLOBAL DEFAULT 8 .str.375`. The same compile links on macOS, where `nm -m` shows `weak external automatically hidden _.str.375`.

## Mechanism

* `LlModule::add_string_constant` names its rodata constants `@.str.N` from a **per-module** counter, as `private unnamed_addr constant`. Private is fine while a module is one translation unit.
* Codegen-unit splitting (`codegen_unit_parts`, #5391/#7174) must make those globals visible to sibling units, so `promote_global_for_units` rewrites every definition to `linkonce_odr`. What happens next is **target-dependent** (`replicate_globals = triple.contains("apple")`):
  * **Mach-O**: every referencing unit gets the `linkonce_odr unnamed_addr` copy. LLVM emits that as `.weak_def_can_be_hidden` → `weak external automatically hidden`. That is what the Mac objects show.
  * **Everything else (ELF and COFF)**: one owning unit gets the definition through `make_unique_owner_global`, which *strips the linkage keyword* — a plain strong `@.str.N = unnamed_addr constant …`, `GLOBAL DEFAULT` — and other units get an `external` declaration. Its doc comment said "On COFF", which is why the diagnosis initially looked for a Mach-O-vs-ELF difference; the branch runs for ELF too.
* Two split modules therefore both export `.str.375` — with different contents — and GNU ld refuses.

**Mach-O was not tolerant, it was a latent miscompile.** ld64 coalesces weak definitions by name across the whole link (autohide only stops the export). Reproduced on this Mac with a 3-module fixture and `PERRY_CODEGEN_UNITS=2` on pristine `main`:

```
expected (node):                      pre-fix perry, split:
alpha1 alpha2 beta1 beta2             /priva Apple  /priv Apple
Apple Banana                          Apple Apple
2 3 10 20                             2 3 10 20
true true                             false true
```

`fn.name`, class names (`js_register_function_name` / class-name registration) and `Function.prototype.toString` all read another module's `.str.N` bytes. Any macOS program with ≥2 split modules is affected today; before #7174-era auto-splitting this needed `PERRY_CODEGEN_UNITS`.

The same emulation surfaced a second unprefixed per-module definition in the same class: `compile_module`'s `@perry_null_guard_zero = internal global i32 0` (the safe-dereference target `LlBlock::safe_load_i32_from_ptr` selects for bad handles). It takes the identical promotion path and is a strong `GLOBAL DEFAULT` symbol in every split ELF object. It did not appear in Coop's error list and I could not determine from here why; it is fixed alongside because it is cheap and the emulation shows it collides.

## Fix

Module-unique names, option (b):

* `LlModule::set_symbol_prefix(prefix)` — installed by `compile_module` right after `sanitize(&hir.name)`. `add_string_constant` now mints `@<prefix>_.str.N`, mirroring what `strings.rs` already does for `<prefix>_.str.N.bytes` / `.handle`. Fixtures that never set a prefix keep the bare `@.str.N`, so nothing downstream shifts.
* `LlModule::null_guard_global()` → `perry_null_guard_zero_<prefix>`; `compile_module` defines it after installing the prefix, and `define_function` injects the name into each function's `RegCounter` (the same shared-cell mechanism #8175 uses for `preserve_nonecc`), so `safe_load_i32_from_ptr`'s 15 call sites are untouched. Functions built outside a module keep `@perry_null_guard_zero`.
* `make_unique_owner_global`'s doc now says ELF/COFF and states the name-uniqueness requirement.

## Trade-off

Option (a) — `linkonce_odr` + `hidden` on ELF — was rejected on purpose: a COMDAT is folded **by name** across the link, so it would reproduce ld64's behaviour on Linux (link succeeds, wrong bytes) instead of the link error. Nothing that is per-module and content-distinct may be weak-by-name; it must be uniquely named. Unique names keep the #7174 rule intact: each shared global is still DEFINED in exactly one unit and every other referencing unit gets `@… = external constant <ty>`; cross-unit references resolve at link time as before. Cross-*module* references to `.str.N` never exist (each `LlModule` is one HIR module, and the name is only handed back to that module's own lowering), so there is nothing a rename could break. Cost: longer symbol names in split objects only; unsplit modules still emit `private` locals that never reach a symbol table.

Not changed (follow-up candidates): promoted owner definitions on ELF still have **default** visibility, so a `--output-type dylib` exports `<prefix>_.str.N`, `perry_class_keys_*`, IC caches, etc. from its dynamic symbol table. That is the pre-existing state for every other promoted global; adding `hidden` visibility is a separate change (`strip_leading_linkage`/`external_decl_for_global` would need to learn the keyword).

## Verification run

* `cargo test -p perry-codegen --lib` — 1332 passed, 1 ignored. New: `module::tests::split_modules_do_not_export_colliding_string_constants` (renders two prefixed modules for `x86_64-unknown-linux-gnu`, 2 units each, both referencing a string constant and the null guard; asserts prefixed names, exactly one strong definition + one `external` declaration per module, no bare `@.str.0`/`@perry_null_guard_zero`, and that **no strong `@sym` definition appears in two units across both modules** — the GNU ld property) and `string_constants_without_a_prefix_keep_the_bare_name`.
* `cargo fmt --check -p perry-codegen`, `cargo clippy -p perry-codegen` (exit 0; the 203 warnings in the crate all predate this change — the two in `module.rs` are on lines from 2026-08-17), `scripts/check_file_size.sh` (module.rs is at 1971/2000), `scripts/addr_class_inventory.py` (+`--self-test`).
* End-to-end on this Mac (arm64, `perry-dev` build of this branch vs. a `perry-dev` build of pristine `main` from the same worktree), 3-module fixture, `PERRY_CODEGEN_UNITS=2`:
  * pre-fix: the miscompiled output above; `nm -m` shows `_.str.0` … in all three objects.
  * post-fix: output byte-identical to `node --experimental-strip-types`; `nm -m` shows `_a_ts_.str.N` / `_b_ts_.str.N` / `_main_ts_.str.N`; the default unsplit compile is also Node-identical.
* ELF, emulated: the per-unit IR dumped by `PERRY_SAVE_LL` was retargeted to `x86_64-unknown-linux-gnu` and rewritten with the exact non-Mach-O owner policy (`make_unique_owner_global` on the first unit, `external_decl_for_global` elsewhere), compiled with Homebrew clang 22.1.4 and linked with `ld.lld -shared`:
  * pre-fix: `llvm-readelf` shows `GLOBAL DEFAULT .str.0`… per module; `ld.lld` reports 10 `duplicate symbol` errors (`.str.0`–`.str.4` ×2, `perry_null_guard_zero` ×2).
  * string fix only: 2 errors, both `perry_null_guard_zero`.
  * both fixes: link succeeds, 0 duplicates; `GLOBAL DEFAULT perry_null_guard_zero_{a_ts,b_ts,main_ts}`.

## Second commit: Linux `--output-type dylib` did not link libm

With the string constants fixed, Coop's Linux link (real pipeline, this branch) went from 2,188 `multiple definition` errors to 0 and then failed one step later: `undefined reference to 'floor'` / `'log10'` from `perry_closure_*` functions. LLVM lowers `llvm.floor`/`llvm.log10` in generated closures to libm calls; the executable link (`link/build_and_run.rs`) has always added `-lm -lpthread -ldl` on Linux, but the `cc -shared` plugin link in `run_pipeline.rs` never did — a plugin resolves everything Perry provides from the host at `dlopen` time, so the gap only shows once an object references a libm symbol. Same "ELF dylib output was never exercised" story.

Fix: `link/linux_dylib_libs.rs::push_unix_dylib_output(cmd, is_linux, exe_path)` finishes the Unix plugin link — on Linux it appends `-lm -lpthread -ldl` **after the objects** (GNU ld only resolves references that precede a `-l`, and Ubuntu's default `--as-needed` drops an earlier one), then `-o`. `run_pipeline.rs` calls it in the non-Windows branch; the macOS `-dynamiclib` command is unchanged (`-lSystem` already carries libm). Tests: `linux_shared_library_link_carries_libm_after_the_objects` (order: objects < libs < `-o`) and `macos_shared_library_link_adds_no_system_libs`. Gates: `cargo test -p perry --bin perry commands::compile::link` (50 passed), `cargo clippy -p perry` (clean for touched files), `cargo fmt --check`, `scripts/check_file_size.sh`. Not verified end to end here (no Linux host); Coop's pipeline is the proof, and Coop's own `cc` shim adding `-lm` defensively remains harmless (duplicate `-lm` is a no-op).

## Not covered

* A real Linux build of Perry or of Coop's bundle (no Linux host here; cross-building the runtime was not attempted). The ELF result above is an emulation of the owner policy on real dumped IR, not Perry's own Linux object emission — Coop's box is the proof.
* The gap/parity suites (CI's job; not run locally on a loaded machine).
* Windows: same code path (`replicate_globals == false`), untested here.
* Why Coop's error list did not also show `perry_null_guard_zero`.
* The Linux dylib `-lm` change itself (second commit) — unit-tested command construction only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed build failures when compiling split modules that contain shared string constants or null-guard references.
  * Improved symbol handling to prevent duplicate definitions and linker conflicts across split modules.
  * Fixed Linux shared-library builds that use math and threading functionality.

* **Compatibility**
  * Linux dynamic-library linking now includes the required system libraries automatically.
  * macOS dynamic-library linking remains unchanged.


proggeramlug:
## Linux verification (fresh Ubuntu 24.04 x86_64, LLVM 22.1.8, GNU ld, real Coop pipeline)

Built the providers, the ext wrappers and the Next.js fixture from this branch at `77e79b704`, clean object cache:

- **`multiple definition of \`.str.N'`: 0** (was 2,188 on `504013308` with the identical input). The ELF fix works as intended.
- The link then failed on `undefined reference to 'floor'` / `'log10'` from `perry_closure_*` in three units (route.js, jsonwebtoken, app-page runtime) — the `-lm` gap the third commit now closes. With `-lm` supplied (by Coop's `cc` shim, PerryTS/coop#20, pending your `abbcf1176`) the fixture **links and publishes** (`app.so`, whole compile 362 s on 16 cores).

Next blocker is past this PR's scope and is being chased in Coop: the daemon SIGSEGVs during the app's module init, with `js_throw → _Unwind_RaiseException` faulting inside `libgcc_s` from `require-hook.js` under a `js_run_module_init_catching` frame — the throw should be caught, the unwinder dies walking the app dylib's frames. Currently testing whether Coop's Linux link flags for the app image (`--gc-sections --strip-debug --discard-all`) damage the unwind tables; if the unwinder fault survives a plain link it becomes a Perry issue and I will file it with the backtrace.

PR #8943, opened 2026-08-28
perf(runtime): length = 0 on a plain dense array is decided from one header read (ECS round 4)

proggeramlug:
One runtime mechanism from the ECS round-4 chain, cut from current main (`924dd1634`, after #8933/#8934/#8935). Suites on the isolated perrymaster gate: array (413, serial) + runtime (2749). Paired measurement on the `codehz/ecs` "5k entities: 3 commands each + sync" row (idle Mac mini, alternating pairs) follows in a comment.

- **`length = 0` on a plain dense array is decided from one header read.** Both `length =` entries (`js_array_set_length`, `js_array_set_length_strict`) resolved the receiver through `clean_arr_ptr_mut` (allocator ownership, forwarding, the Buffer / typed-array registries), coerced the new length, resolved the flags a second time (`array_object_flags` on the strict entry, then `resolved_plain_array_flags` again inside) and probed the named-property table before reaching the plain-shrink branch — for an object pool's `pooled.length = 0` that tower was the whole cost (`js_array_set_length` + `_strict` ≈ 1.8% of the frame after #8934's single-pass re-arm), five thousand times a frame. The header facts the pop fast path already proves are enough: a `GC_TYPE_ARRAY` head that is not forwarded, none of the integrity / descriptor flags (a non-writable `length` is recorded under `OBJ_FLAG_ARRAY_DESCRIPTORS`, so neither entry has anything to throw), a dense `length <= capacity`, and no named properties (latch-gated). The work is the plain-shrink branch's, unchanged: holes over the retired prefix, the length, one layout rebuild. Anything else declines to the full entry. `length_zero_takes_the_header_lane_on_a_plain_array_and_declines_otherwise` pins both entries on a plain array (holes read back as `undefined` after a later extension, the array stays usable), the empty no-op, and the named-property decline.

## Summary by CodeRabbit

* **Performance**
  * Improved setting a plain dense array’s length to zero for faster truncation.
  * Preserved full validation behavior for frozen, sealed, forwarded, or property-bearing arrays.

* **Bug Fixes**
  * Ensured arrays remain usable after zero-length truncation and subsequent extension or pushes.
  * Repeated zero-length assignments remain safe no-ops, while strict behavior for restricted arrays is preserved.


proggeramlug:
Gate on the isolated perrymaster clone for `de6f506a4` (base `924dd1634`): codegen lib 1331, native_proof_regressions 280, transform lib 119 green; runtime suite 2748/2749 with the one failure being `box::release_tests::completed_activation_residue_is_bounded_not_linear` (process-global counters, known parallel flake — passes alone and the box module passes serially on this head; this PR does not touch `box.rs`); lint gates and merge-base ratchets flat (576/967). Paired measurement on the mini follows.

proggeramlug:
Paired measurement (idle Mac mini, 9 alternating pairs, `codehz/ecs` "5k entities: 3 commands each + sync"): control = its base `924dd1634` (main with #8933/#8934/#8935), candidate = this branch: **3.348 → 3.279 ms/op, +2.07%, 9/9** (`r4k-screen.json`, process oracles 18/18).

PR #8940, opened 2026-08-28
codegen: emit the receiver-unknown numeric index tiers once per dynamic site

proggeramlug:
Rebased onto main after #8931 merged; one commit.

## What

`arr[key]` on an erased-Array receiver with a dynamic key (`packed[sparse[x]]` in wolf-ecs `SparseSet.has`, `a[b[i]]` in general) lowers through the canonical-i32 split. Since the brand arm (#8890's v90), every integral key below 2³¹ — canonical double or INT32-boxed — is served in the canonical arm's `aidx.claimed.other` branch, which emits the receiver-unknown numeric tiers (inline typed-array read `tav.*`, dense Array-subclass `arrlike.ic.*`, complete dispatcher) once. The runtime-key arm (`lower_claimable_array_string_key_get`) still carried its **own copy** of that lattice from the v83 dynamic-key work (`aidxkey.int` → `aidxkey.int.exact`), which after v90 only ever ran for integral keys in `[2³¹, 2³²)`. Its other caller passes static string/symbol keys, where the numeric path is dead.

This PR drops that copy: those keys take the complete `js_array_get_index_or_string` route (which handles every index). Same semantics; ~11 KB of IR less per dynamic site.

## Why it matters

Diagnosing why `SparseSet.has$pshape` is still a `bl` inside `addComponent$pshape` (campaign item #4): an env-gated trace of the pre-statepoint inline admission showed the one-statement clone `has$pshape$idx_u31` at **45,936 bytes of lowered IR** against the 16 KiB budget (`GUARDED_SPECIALIZATION_PREINLINE_MAX_IR_BYTES`) — which is why the earlier 32 KiB experiment (v82b) could never fire. Byte split: `arrlike.ic` 13.2 KB, `tav` 9.4 KB, guarded `arr` 6.7 KB, `aidx` 5.7 KB, `plen` 5.1 KB — i.e. the dynamic-key site alone was ~30 KB because the two arms each emitted the same lattice. This is the first of the size reductions; the clone is still above budget afterwards (the `.length` IC lattice and the two guarded `this.sparse[x]` tiers are next), so no inlining change is claimed here — only smaller code at every such site.

## Verification (local, per the campaign rule)

[...]

PR #8944, opened 2026-08-28
codegen: inline plain-array pop tier with js_array_pop_f64 as the fallback

proggeramlug:
## What

`arr.pop()` on an Array-admitted receiver — an erased class field (`this.packed.pop()`, the `NativeMethodCall { array, pop }` route) or a claimed-array local/parameter (`lower_array_method`) — now emits an inline plain-array tier (`expr/array_pop.rs`) with `js_array_pop_f64` kept as the fallback. The inline admission mirrors the runtime fast path's decision exactly:

- POINTER-tagged heap handle (a claimed receiver may be a String or a number → call);
- `GC_TYPE_ARRAY`, not `GC_FLAG_FORWARDED`, none of `FROZEN|SEALED|NO_EXTEND|ARRAY_DESCRIPTORS` (`0x407`) in `_reserved` (frozen / non-writable-length arrays must throw, descriptor-bearing ones need the descriptor-aware `[[Delete]]`);
- `PERRY_ARRAY_INDEX_FAST_PATH_INVALIDATED == 0` (polluted `Array.prototype`);
- `0 < length <= capacity <= 1e8` and the popped slot is not the hole sentinel.

The inline arm reads the element and stores the decremented `i32` length — no pointer store, no allocation, no GC bookkeeping, which is also all the runtime's fast path does. The empty array still calls (that arm is #8934's runtime fast path to keep). No runtime changes.

## Why

A fresh symbol-level profile of the wolf-ecs twins (Linux `perf`, 10 s): `js_array_pop_f64` is **10.1%** of self time in the entity cycle and **8.1%** in add/remove, while every pop there (`SparseSet.remove`'s and `createEntity`'s `packed.pop()`) takes the runtime's plain-array fast path — the cost is the call, the heap-address classification (`try_read_gc_header`) and the flag resolution around a single length store.

## Verification (local, per the campaign rule; no CI wait)

- IR census test `array_pop_takes_the_inline_tier_with_the_runtime_call_as_fallback` for both routes: `apop.hdr` tests `GC_TYPE_ARRAY`, the `1031` mask and the latch; `apop.read` tests the hole sentinel; `apop.take` is one `store i32` and no call; `apop.slow` calls `js_array_pop_f64`.
- Gate run (suites, `-D warnings`, file size, GC store inventory, raw-handle debt, census, local-binding, addr-class) and an e2e semantics test (holes, frozen, non-writable length, Array subclass, string receiver, accessor slot, polluted prototype — node-identical output) are in progress; Mac-mini paired screens main → PR (2 s + 50 ms, 11 pairs) follow. I'll post all three as comments.

## Summary by CodeRabbit

* **Performance**
  * Improved `Array.prototype.pop()` performance for eligible plain arrays by handling common cases inline.
  * Preserved the existing fallback behavior for arrays and conditions that cannot use the optimized path.

* **Bug Fixes**
  * Ensured empty arrays return `undefined` correctly.
  * Added safeguards for unsupported array states, sparse arrays, and invalid lengths.

### DIFF
--- a/changelog.d/0000-inline-array-pop-tier.md
+++ b/changelog.d/0000-inline-array-pop-tier.md
@@ -1,0 +1,3 @@
+### Changed
+
+- `arr.pop()` on an Array-admitted receiver (a class field holding an array, a claimed-array local or parameter) now runs an inline plain-array tier — the same admission `js_array_pop_f64`'s fast path makes (`GC_TYPE_ARRAY`, not forwarded, no frozen/sealed/no-extend/descriptor flags, prototype latch clear, `0 < length <= capacity`, no hole) — with the runtime call kept as the fallback for everything else. In the wolf-ecs entity cycle the plain-array `packed.pop()` was 8–10% of self time on nothing but that fast path's call overhead.
--- a/crates/perry-codegen/src/expr/array_pop.rs
+++ b/crates/perry-codegen/src/expr/array_pop.rs
@@ -1,0 +1,351 @@
+//! Inline `arr.pop()` for a receiver the HIR admits as an Array.
+//!
+//! `js_array_pop_f64`'s own fast path is a handful of header tests and one
+//! length store, but it is reached through a call, a heap-address
+//! classification (`try_read_gc_header`) and — for anything it declines — a
+//! tower of subclass / plain-object probes. In the wolf-ecs entity cycle the
+//! plain-array `packed.pop()` of `SparseSet.remove` and `createEntity` is 8–10%
+//! of self time with nothing but that fast path running (Linux `perf`, 10 s
+//! twins). This emits the same decision inline, exactly as the runtime makes
+//! it, and keeps the runtime call as the fallback for everything else:
+//!
+//! * a `POINTER`-tagged heap handle (the receiver may be a claimed, not
+//!   proven, Array: a String or a NaN-boxed number takes the call);
+//! * `GC_TYPE_ARRAY`, not forwarded (`GC_FLAG_FORWARDED`), none of
+//!   `FROZEN | SEALED | NO_EXTEND | ARRAY_DESCRIPTORS` in `_reserved` — a
+//!   frozen array or a non-writable `length` must throw, a descriptor-bearing
+//!   one needs the descriptor-aware `[[Delete]]`;
+//! * `PERRY_ARRAY_INDEX_FAST_PATH_INVALIDATED == 0` — a polluted
+//!   `Array.prototype` can intercept the element read;
+//! * `0 < length <= capacity <= 100_000_000` and the popped slot is not the
+//!   hole sentinel (a hole reads through the prototype chain).
+//!
+//! On the inline arm the element is read, `length` is decremented and the
+//! element returned. Nothing is allocated and no pointer is stored (the
+//! length word is an `i32`), so there is no GC bookkeeping: the runtime's
+//! fast path performs none either. An empty array takes the call (it answers
+//! `undefined`, and that arm is the runtime's to keep fast).
+use crate::types::{DOUBLE, I1, I16, I32, I64, I8};
+
+use super::{unbox_to_i64, FnCtx};
+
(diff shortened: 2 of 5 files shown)


proggeramlug:
Merged. The inline admission mirrors the runtime fast path's decision — POINTER-tagged heap handle, `GC_TYPE_ARRAY`, not `GC_FLAG_FORWARDED`, and none of `FROZEN|SEALED|NO_EXTEND|ARRAY_DESCRIPTORS` (0x407) — so frozen and descriptor-bearing arrays keep the runtime path that can throw or run the descriptor-aware `[[Delete]]`. That gating is the part that had to be exact, and it is.

Validation: codegen 1334/0, runtime 2771/0 (`RUST_TEST_THREADS=1`); `scripts/run_lint_gates.sh` 57 of 58 including the compile tier green (`-D warnings` and `clippy`), the exception being the pre-existing `${{ }}` substitution artifact (#8929).

proggeramlug:
Gates on `8f9614a8f` (Linux host): `RUSTFLAGS=-D warnings cargo check --workspace --all-targets` clean; `cargo test -p perry-codegen -p perry-transform -p perry-hir` 2505 passed / 0 failed; file size OK; GC store-site inventory passed; raw-handle debt unchanged (967, none raised); shape census OK; local-binding audit OK; addr-class audit passed.

Semantics probe (holes, frozen, non-writable `length`, Array subclass, accessor slot, polluted `Array.prototype`, empty): identical before and after the tier. Two **pre-existing** gaps surfaced by the probe, unchanged by this PR (both route to the runtime call exactly as before): popping a hole yields `NaN` where Node yields `undefined`, and `"abc".pop()` through an erased field does not throw. Noting them separately; the e2e test in this PR pins the cases where perry already matches Node.

Mac-mini paired screens (11 pairs), main `a1e44c5df` → this PR:

|---|---|---|
| 2 s | 0.4031 → 0.4040 (+0.20%, 0/11) | 0.3348 → 0.3341 (**−0.23%**, 11/11) |
| 50 ms | 0.4030 → 0.4039 (+0.25%, 2/11) | 0.3351 → 0.3343 (**−0.28%**, 11/11) |

That is far less than 8–10% of pop self time should buy, and add/remove moves the wrong way, so I'm not calling this a retain yet — re-profiling the PR build on Linux to see whether the inline arm is actually being taken on the hot pops (if `js_array_pop_f64` self time is gone and the residual is code-size cost in the hot bodies, this PR gets closed like #8940).

proggeramlug:
Resolved why the ECS numbers are small: wolf-ecs's `Archetype` constructor does `this.sset.packed = this`, so the hot `sset.packed.pop()` / `.push()` / `[i]` in add/remove/destroy run on the **object-backed Array subclass itself**, not on a plain array (gdb at the residual `js_array_pop_f64` call from `destroyEntity$pshape`: receiver `obj_type=2`, `_reserved=0x800`, latch clear). The tier correctly declines those (obj_type ≠ ARRAY) — that is the 8–10% "pop" self time, and it is next-list item #3 (object-backed push/pop), not plain-array pop. The only plain pops in the cycle are `_rm.packed.pop()` in `createEntity`, which is the consistent −0.25% (11/11) on the entity cycle; add/remove has no plain pops, so its +0.2% (0/11) is the cost of the ~8-instruction pre-check in front of the runtime call on object-backed receivers.

Where the tier does apply it pays: plain-array push/pop microbenchmark (a class field `packed = []`, 20 000 rounds of 500 `pop` + 500 `push`, Linux x86_64, 5 alternating runs): main-equivalent **133.6 / 133.9 / 134.1 / 134.0 / 136.7 ms** → this PR **123.5 / 128.2 / 123.1 / 123.0 / 125.7 ms** (−7.7% median), with pushes (and their GC bookkeeping) still in the loop — so per pop it is well over that.

Recommendation: merge as a general improvement (semantics pinned by the e2e test, `f87f43b5a`; all gates green); it is not an ECS-campaign retain by the 11/11-both rule, and I've said so above. The object-backed subclass push/pop path is the actual ECS cost center and is what I'm looking at next.

PR #8945, opened 2026-08-28
perf(runtime): strict element store — pointer-overwrite lane (ECS round 4)

proggeramlug:
One runtime mechanism from the ECS round-4 chain, cut from current main (`a1e44c5df`). Suites on the isolated perrymaster gate: array (414, serial) + runtime (2751). Paired measurement on the `codehz/ecs` "5k entities: 3 commands each + sync" row (idle Mac mini, alternating pairs) follows in a comment.

- **The strict element store gains a pointer-overwrite lane.** `js_array_set_f64_extend_strict`'s two exact fast lanes are number lanes, so a pointer value declines at their first test and `column[index] = record` — once per command on an ECS archetype's component column — paid the whole tower to reach a one-slot write: the registry-probing head resolver (`clean_arr_ptr_mut`, whose page-map classification through the hot-TLS chain was the store's hottest offset), a second flag resolution, the descriptor guard, the string add-ref probe, a handle scope, the prototype note and the extend path's own descriptor checks. `try_strict_dense_pointer_overwrite` admits with the plain-number lane's receiver discipline (the same decode, magnitude, alignment, plausibility and generation classification before the header is read) and the same integrity / descriptor / element-shape / raw-f64 rejections, plus what a pointer overwrite specifically needs — the head's layout is not pointer-free and the old slot already holds a pointer, so no layout claim changes — and then performs `store_array_slot_resolved`: the exact layout note and write barrier the general path performs. What is skipped is bookkeeping the proven shape cannot need, never a barrier; `Array.prototype` keeps the general path. `strict_dense_pointer_overwrite_lane_matches_the_general_path` pins the admitted shape through both the lane and the public entry, and the declines (number value, hole slot, number slot, extension, null / non-pointer receiver).

## Summary by CodeRabbit

* **Performance**
  * Improved performance for in-range array updates where one heap pointer replaces another.
  * Accelerated dynamic property access and deletion by using indexed key lookups instead of repeated scans.
  * Existing array safety and storage rules remain unchanged.

* **Tests**
  * Added coverage for pointer replacements across supported array representations.
  * Verified unsupported values, invalid indices, empty slots, and invalid receivers remain unchanged.


proggeramlug:
Gate on the isolated perrymaster clone for `3349e335c` (merge-base `588e017ed`): runtime suite 2751/2751, codegen lib 1333, native_proof_regressions 280, transform lib 119 — all green; lint gates and merge-base ratchets flat (576/967). Paired measurement on the mini follows.

proggeramlug:
Merged.

**One fix pushed:** `array/strict_store_tests.rs` had a redundant `unsafe` block that `-D unused-unsafe` rejects, so the `warnings` job was red on this branch.

**Validation** — runtime 2772/0 (`RUST_TEST_THREADS=1`); under `PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1` the failing count is 16, matching the pre-existing set on main (mostly `promote_in_place` tests, which assert in-place promotion that forced evacuation contradicts by construction); `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green after the fix — the exception is the pre-existing `\${{ }}` substitution artifact (#8929).

proggeramlug:
Paired measurement (idle Mac mini, 9 alternating pairs, `codehz/ecs` "5k entities: 3 commands each + sync"): control = main before this merge (`ae21f190b`), candidate = main with it (`6d10e8a1d`): **3.260 → 3.248 ms/op, +0.33%, 9/9** (`r4l-screen.json`, process oracles 18/18). Smaller than the strict store's self time suggested — most of that was the header-line miss, which the lane still has to take — but unanimous.

PR #8949, opened 2026-08-28
perf(gc): mirror the write barrier's dirty-page cache in a TSD-tagged process global (ECS round 4)

proggeramlug:
One GC mechanism from the ECS round-4 chain, cut from current main. Suites on the isolated perrymaster gate: `barrier` (73, serial) + runtime (2751). The Darwin/aarch64 path is exercised on the Mac (barrier suite + process oracles) and measured on the mini; both results follow in comments.

- **The write barrier's dirty-page cache is mirrored in a process global tagged by the writer's TSD base.** After #8935 made the dirty-page hit a leaf path, the merged-main profile put the barrier entry's single hottest instruction — 2.3% of the ECS frame — on the `mrs TPIDRRO_EL0` and the dependent loads behind it: the per-thread cell is reached through the hot-TLS chain (global slot-index load → pthread key → `mrs` → TSD slot → cell), four dependent loads on every barrier call. The cell stays the authority; every path that writes or clears it also writes a `(page, owner)` pair of process globals, `owner` being the writing thread's TSD base (`tls_hot::darwin_tsd::base()`, the same non-`pure` `mrs` discipline as `get`). A reader identifies itself with one `mrs` and two loads that do not depend on each other: if the owner is the calling thread, the page word is its own last write, so the compare is exactly the cell's; otherwise it falls back to the cell. A torn read across the two words can only answer "not cached" for a page the reader owns (heaps are per thread, so another thread's page is never this thread's slot page) — the conservative direction, which just takes the full path. Darwin/aarch64 only; other targets keep the cell. Behaviour is pinned by the existing dirty-page and barrier tests (the remembered set built is unchanged).

## Summary by CodeRabbit

- **Performance**
  - Improved garbage-collection write-barrier checks on Apple ARM64 devices.
  - Reduced overhead for repeated dirty-page checks while preserving existing behavior on other platforms.

### DIFF
--- a/changelog.d/8949-barrier-cache-mirror.md
+++ b/changelog.d/8949-barrier-cache-mirror.md
@@ -1,0 +1,1 @@
+- **gc:** the write barrier's one-entry dirty-page cache is mirrored in a process global tagged with the writing thread's TSD base, so the owning thread's hit test is one `mrs` and two independent loads instead of the four-load hot-TLS chain (Darwin/aarch64; other targets keep the per-thread cell, which stays the authority).
--- a/crates/perry-runtime/src/gc/dirty_page_cache.rs
+++ b/crates/perry-runtime/src/gc/dirty_page_cache.rs
@@ -85,6 +85,75 @@ use std::cell::Cell;
 /// `usize::MAX` would need a 76-bit address.
 const NO_PAGE: usize = usize::MAX;
 
+/// The process-global mirror of the cache, tagged with the TSD base of the
+/// thread that wrote it.
+///
+/// The per-thread cell stays the authority, but reaching it costs the hot-TLS
+/// chain — a global slot-index load, the pthread key, `mrs`, the TSD slot,
+/// then the cell — four *dependent* loads on every barrier call, and the
+/// profile put the barrier entry's single hottest instruction on that chain.
+/// This mirror is read with one `mrs` and two loads that do not depend on
+/// each other: if the owner word names the calling thread, the page word is
+/// that thread's own most recent write (every path that writes or clears the
+/// cell also writes here), so the compare is exactly the cell's; otherwise
+/// another thread wrote last and the reader falls back to its cell.
+///
+/// Why a torn read across the two words is still harmless: a reader can only
+/// mis-see a page another thread cached, and heaps are per thread — a slot
+/// this thread stores into is never on another thread's page — so the
+/// mismatch cannot answer "already dirty" for a page this thread owns.
+#[cfg(all(
+    target_vendor = "apple",
+    target_arch = "aarch64",
+    target_pointer_width = "64"
+))]
+mod mirror {
+    use std::sync::atomic::{AtomicUsize, Ordering};
+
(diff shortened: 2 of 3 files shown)


proggeramlug:
Gate on the isolated perrymaster clone for `d66d9493c` (base `cae1b95c3`): runtime suite 2751/2751, codegen lib 1334, native_proof_regressions 280, transform lib 119 — all green; lint gates and merge-base ratchets flat (576/967). (Linux exercises the fallback arm; the Darwin/aarch64 mirror arm runs the barrier suite on the Mac next, then the paired mini measurement — both posted here when done.)

proggeramlug:
Merged — but one thing in the safety argument is worth correcting, because it is load-bearing.

The comment on `dirty_old_page_already_marked` says a stale mirror read "can only answer *not cached* for a page this thread owns — the conservative direction". That is not what the ordering admits. `publish` stores `PAGE` then `OWNER`, both `Relaxed`, and `probe` reads `OWNER` then `PAGE`, so this interleaving is allowed:

```
A: PAGE = X
B: PAGE = Y
B: OWNER = B
A: OWNER = A        ->  OWNER == A, PAGE == Y
```

A then probing for page `Y` sees `OWNER == me` and `PAGE == page` and returns **true** — a false *positive*, not a false negative. Since the mirror's answer is returned directly and short-circuits the authoritative thread-local cell, that would skip a remembered-set entry the thread actually owed.

**What actually makes it safe** is that arenas are thread-local bump allocation (`arena/mod.rs`), so a thread never stores into another thread's pages and therefore never asks about a page some other thread published. The disjointness is the invariant, not the store ordering.

That distinction matters because the two have different failure conditions: if arena pages ever become shareable or recyclable across live threads, the current comment would still read as correct while the mechanism silently became unsound. Worth restating the invariant as "a thread only ever probes pages in its own arena" and, if you want belt-and-braces, `Release`/`Acquire` on the `OWNER` store/load so the pairing is self-evidently safe rather than safe-by-distant-invariant.

Not blocking: I could not construct a reachable case on the current threading model.

**Validation** (batched with #8948, #8950, #8951) — codegen 1337/0, runtime 2773/0; under `PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1` the failing count is 16, the pre-existing set; `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green.

proggeramlug:
Darwin/aarch64 arm: the barrier suite passes on the Mac (73/73, mirror path exercised). Paired measurement (idle Mac mini, 9 alternating pairs, `codehz/ecs` "5k entities: 3 commands each + sync"): control = base `cae1b95c3`, candidate = this branch: **3.248 → 3.242 ms/op, +0.17%, 9/9** (`r4m-screen.json`, oracles 18/18). Unanimous but marginal: the `mrs` itself turns out to be most of what the profile attributed to the chain, so the mirror only removes the two dependent loads after it. It is sound and self-contained, but if a process-global mirror is not worth 0.2%, closing this one is a reasonable call — I have no further work stacked on it.

PR #8951, opened 2026-08-28
perf(runtime): typed-feedback store forwards when recording is off; one-entry cache for the external-slot remembered set (ECS round 4)

proggeramlug:
Two runtime mechanisms from the ECS round-4 chain, cut from current main. Suites on the isolated perrymaster gate: runtime (2752, incl. the new external-cache test). Paired measurement on the `codehz/ecs` "5k entities: 3 commands each + sync" row (idle Mac mini, alternating pairs) follows in a comment.

- **The typed-feedback array store forwards at once when recording is off.** `js_typed_feedback_array_set_index_or_string` converted the index and made two out-of-line calls (`observe_array`, `record_guard_pass`) whose first statement is "recording is off, return" — 1.5% of the merged-main ECS frame on `column[index] = record`. One flag test (`typed_feedback_enabled`, the default is off), then the strict store; the recording path is unchanged when the env enables it.
- **A one-entry cache in front of the external-slot remembered set.** A `Map`'s entries buffer is an external slot span, so every `map.set(k, v)` on an old map reached `mark_dirty_external_slot_page`: a thread-local table probe plus a linear scan of that page's header list (which grows with every map whose buffer shares the page) — the two hottest offsets of `js_map_set` on merged main. The `(page, header)` pair is cached inline in `HotTls`, recorded only after the table holds it and dropped wherever the table drops a pair (`clear_one_external_dirty_slot_header`, which every clear path goes through), so a repeated store into the same map answers from two loads — the external-slot twin of the inline-slot dirty-page cache (#8916). `external_dirty_slot_pair_cache_mirrors_the_table` pins hit / miss / re-record-after-clear.

## Summary by CodeRabbit

- **Performance**
  - Improved garbage-collection tracking efficiency for repeated external-slot updates.
  - Streamlined array assignments when typed feedback is disabled by avoiding unnecessary processing.

- **Reliability**
  - Ensured external-slot tracking remains consistent when remembered entries are cleared.

- **Tests**
  - Added coverage for external-slot cache hits, misses, and reset behavior.

- **Documentation**
  - Added changelog notes describing the runtime and garbage-collection improvements.


proggeramlug:
Gate on the isolated perrymaster clone for `9bc29aeb5` (base `cae1b95c3`): runtime suite 2752/2752 (incl. the new `external_dirty_slot_pair_cache_mirrors_the_table`), codegen lib 1334, native_proof_regressions 280, transform lib 119 — all green; lint gates and merge-base ratchets flat (576/967). Paired measurement on the mini queued (control = `cae1b95c3`).

proggeramlug:
Paired measurement (idle Mac mini, 9 alternating pairs, `codehz/ecs` "5k entities: 3 commands each + sync"): control = base `cae1b95c3`, candidate = this branch: **3.246 → 3.207 ms/op, +1.20%, 9/9** (`r4n-screen.json`, oracles 18/18). The Darwin barrier suite passes on the Mac too (74/74, incl. the new pair-cache test).

Issue #6475, opened 2026-07-16
methods inherited via Object.setPrototypeOf(obj, proto) run with this=undefined — effect Pipeable/Tag statics return the wrong pipe stage (blocks web.ts, 'Not a valid effect: undefined')


proggeramlug:
## Summary

The last standing failure of effect-compiled-experiment `web.ts` (after #6449 / #6459 / #6460 / #6463+#6466 / #6465+#6470 / #6469+#6474): startup dies with

```
Error: Not a valid effect: undefined      (fiberRuntime.ts:1422)
```

Root of the poisoned value, isolated by three probes:

## probe43 — the poisoned binding

Walking web.ts's layer graph binding-by-binding: everything matches node except

```
HealthLive    node: object (a Layer)   perry: function
```

`HealthLive = HttpApiBuilder.group(Api, "health", ...)` — and `group` returns `Router.use((router) => ...)` where `Router` is `class Router extends HttpRouter.Tag("@effect/platform/HttpApiBuilder/Router")<Router>() {}`.

## probe44 — Tag statics misbehave

Recreating the Tag shape directly (`class R2 extends HttpRouter.Tag("probe44/R2")<R2>() {}`):

| check | node | perry |
|---|---|---|
| `pipe(1, f, g)` / 4 fns / method-style `.pipe` | ✓ | ✓ (all fine) |
| `R2.key` | ✓ | ✓ |
| `R2.Live` | Layer | Layer ✓ |
| **`R2.use(f)`** | Layer | **function** (isLayer false) |
| **`R2.unwrap(f)`** | Layer object | **function** |

`use` is an own-prop ARROW on the factory's TagClass (`TagClass_.use = (f) => TagClass_.pipe(Effect.flatMap(f), Layer.scopedDiscard, Layer.provide(TagClass_.Live))`), inherited by `R2` through the parent function object. The returned function is consistent with the pipe chain's LAST stage (`Layer.provide(live)`, a data-last curried fn) never being applied — i.e. `TagClass_.pipe(...)` misfired on the receiver/arguments. `TagClass`'s prototype is `Object.getPrototypeOf(Context.GenericTag(id))` (a two-hop chain to Pipeable's `pipe() { return pipeArguments(this, arguments) }`).

## probe45 — a minimal `this`-drop divergence in the same family

```ts
const Proto: any = { collect(this: any, ...args: any[]) { return [this?.tag, args.length, ...args] } }
[...]

PR #8952, opened 2026-08-28
perf(runtime): closure_is_arrow from the dispatch cache; ReadonlySet.has leaf; Set.clear empty exit (ECS round 4)

proggeramlug:
Three small runtime mechanisms from the ECS round-4 chain, cut from current main (`23a211383`). Suites on the isolated perrymaster gate: `set::` (54, serial) + runtime (2752). Paired measurement on the `codehz/ecs` "5k entities: 3 commands each + sync" row (idle Mac mini, alternating pairs) follows in a comment.

- **`closure_is_arrow` answers from the dispatch cache.** It probed the arrow registry — a thread-local hash lookup — on every receiver rebind, i.e. on every `this.handler(a, b)` call through a closure-typed field (`CommandBuffer.execute`'s `this.executeEntityCommands(entityId, commands)`, 5k/frame: `closure_is_arrow` 0.8% + the rebind helper 0.9%), while the `DispatchStrategy` the call resolves right after already carries `is_arrow`, memoised per body in the recent-bodies cache and kept coherent by `js_register_closure_arrow_function`'s invalidation (#6475). Answer from that cache instead. (Stamping such facts into the closure header at allocation was considered and rejected: late registration during cyclic module init purges the strategy caches, and a stamped header cannot be purged.)
- **`js_readonly_set_has` is a leaf.** It inlined its structural fallback (a handle scope and the generic method dispatch), so the genuine-`Set` arm — `componentTypeSet.has(type)` on every command, 1.4% self — paid the fallback's frame. The fallback is outlined and cold.
- **`js_set_clear` exits on an already-empty set** before the side-table probe (the table mirrors the elements, so there is nothing to reset) — half of a change set's per-entity `adds.clear(); removes.clear()`; the small-Set lane test now also pins clear-on-empty and clear-then-reuse.

## Summary by CodeRabbit

* **Performance**
  * Improved runtime efficiency for closure checks and set operations.
  * Optimized membership checks and clearing already-empty sets, with no change to expected behavior.
* **Reliability**
  * Expanded coverage for set clearing and subsequent member re-addition scenarios.
* **Documentation**
  * Added changelog details covering the runtime performance improvements.

### DIFF
--- a/changelog.d/8952-closure-arrow-set-leaf.md
+++ b/changelog.d/8952-closure-arrow-set-leaf.md
@@ -1,0 +1,1 @@
+- **runtime:** `closure_is_arrow` takes its answer from the memoised dispatch strategy instead of a second thread-local registry probe per receiver rebind; `ReadonlySet.has` outlines its structural fallback so the genuine-`Set` arm is a leaf; `Set.clear()` on an already-empty set returns before the side-table probe.
--- a/crates/perry-runtime/src/closure/registry.rs
+++ b/crates/perry-runtime/src/closure/registry.rs
@@ -658,7 +658,13 @@ pub fn closure_is_arrow(closure: *const ClosureHeader) -> bool {
     if func_ptr.is_null() {
         return false;
     }
-    is_registered_arrow_function(func_ptr)
+    // The unified dispatch strategy already carries arrow-ness (the same
+    // registry answer, memoised per body and kept coherent by
+    // `js_register_closure_arrow_function`'s invalidation), and the call
+    // that follows a receiver rebind resolves it anyway — so answer from the
+    // recent-bodies cache instead of a second thread-local hash probe per
+    // call. (`this.handler(a, b)` on a closure-typed field paid both.)
+    resolve_strategy(func_ptr).is_arrow()
 }
 
 /// True if `closure` is a bound-method / bound-function value (its body is the
--- a/crates/perry-runtime/src/set.rs
+++ b/crates/perry-runtime/src/set.rs
@@ -1194,10 +1194,18 @@ pub unsafe extern "C-unwind" fn js_readonly_set_has(receiver: f64, value: f64) -
         }
     }
 
-    // The structural fallback can allocate and re-enter generated code. Root
-    // both operands before crossing that boundary, then pass refreshed values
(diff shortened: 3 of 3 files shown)


proggeramlug:
Gate on the isolated perrymaster clone for `9eb19e996` (base `23a211383`): runtime suite 2752/2752, codegen lib 1337, native_proof_regressions 280, transform lib 119 — all green; merge-base ratchets flat (576/967). The lint file-size step reports `crates/perry-codegen/src/gc_map.rs` at 2047 lines — that file is untouched here and is already over the cap on the base commit, so it is a pre-existing red from another change, not this PR. The Darwin `set::` suite passes on the Mac (54/54); paired measurement follows.

proggeramlug:
Paired measurement (idle Mac mini, 9 alternating pairs, `codehz/ecs` "5k entities: 3 commands each + sync"): control = base `23a211383`, candidate = this branch: **3.202 → 3.188 ms/op, +0.41%, 9/9** (`r4o-screen.json`, oracles 18/18). Darwin `set::` suite 54/54 on the Mac.

proggeramlug:
Merged. All three are faithful substitutions rather than new decisions, which is what I checked:

**`closure_is_arrow` from the dispatch cache** is the one that could have gone wrong, since arrow-ness decides `this` binding. It is equivalent on every path: `resolve_strategy`'s slow path computes `is_arrow: is_registered_arrow_function(func_ptr)` — literally the call this replaces — and the only two early returns that hardcode `is_arrow: false` are the `BOUND_METHOD_FUNC_PTR` / `BOUND_FUNCTION_FUNC_PTR` sentinels, which are not real function bodies and would have answered `false` from the registry too. So a cache miss cannot produce a different answer, only a slower one.

The note about rejecting a header stamp is worth keeping in the record — late registration during cyclic module init purges the strategy caches, and a stamped header cannot be purged.

**Validation** — runtime 2773/0, codegen 1337/0 (`RUST_TEST_THREADS=1`); `scripts/run_lint_gates.sh` 57 of 58 including the compile tier, re-run after #8957 repaired `main`'s file-size gate — the remaining one is the pre-existing `\${{ }}` substitution artifact (#8929).

PR #8960, opened 2026-08-28
perf(codegen): for-of over a Map reads its entries inline (ECS round 4 follow-up)

proggeramlug:
One codegen mechanism (the second of the two follow-ups agreed after round 4), cut from current main. Suites on the isolated perrymaster gate: codegen lib (1338, incl. the new `map_entry_reads_are_inline_with_the_helper_as_the_fallback`) + `native_proof_regressions` (280). Paired measurement on the `codehz/ecs` "5k entities: 3 commands each + sync" row (idle Mac mini, alternating pairs, control = the branch's merge base) follows in a comment.

- **`for (const [k, v] of map)` reads its entries inline.** Each iteration of the desugared loop read its key and its value through `js_map_entry_key_at` / `js_map_entry_value_at`, and each of those resolved the receiver again (`clean_map_ptr`: tag, magnitude, header, subclass redirect) to read one word of a flat buffer — two calls and two resolutions per entry, 0.8% of the ECS frame on its per-frame grouping map. `lower_map_entry_at_inline` decides the common receiver from the same facts the resolver would establish — a NaN-boxed heap pointer above the handle band whose header is `GC_TYPE_MAP` and not forwarded — re-reads `size` and `entries` on every read (the body may `set`/`delete` and grow or compact the buffer; the receiver itself is re-read from its root by the rooted-operand group), bounds-checks the index and loads the slot. Every other shape — a `class X extends Map` instance, a plain object, an out-of-range or negative index, a handle — takes the runtime helper exactly as before, so the two paths are equivalent by construction.

## Summary by CodeRabbit

  * Improved Map iteration performance by reading keys and values more efficiently during entry access.
  * Added safe fallback behavior for unsupported Map shapes and invalid indexes.

* **Tests**
  * Added coverage validating optimized Map entry reads, bounds checking, and fallback behavior.

* **Documentation**
  * Added a changelog entry describing the Map iteration performance improvement.

### DIFF
--- a/changelog.d/8960-map-entry-inline.md
+++ b/changelog.d/8960-map-entry-inline.md
@@ -1,0 +1,1 @@
+- **codegen:** `for (const [k, v] of map)` reads each entry's key and value inline from the Map's flat buffer (a `GC_TYPE_MAP` header test, a live-size bounds check and one load) instead of two runtime calls that each re-resolved the receiver; subclass instances, plain objects and out-of-range indices keep the runtime helpers.
--- a/crates/perry-codegen/src/expr/arrays_finds.rs
+++ b/crates/perry-codegen/src/expr/arrays_finds.rs
@@ -75,6 +75,117 @@ use super::{
 // unsound the moment one of them was edited.
 use super::index_get::numeric_index_has_integer_array_index_proof;
 
+/// `for (const [k, v] of map)` entry read, inline.
+///
+/// Each iteration of the desugared loop read its key and its value through
+/// `js_map_entry_key_at` / `js_map_entry_value_at`, and each of those
+/// resolved the receiver again (`clean_map_ptr`: tag, magnitude, header,
+/// subclass redirect) to read one word of a flat buffer — two calls and two
+/// resolutions per entry, 0.8% of an ECS frame on its per-frame grouping
+/// map. The lowering here decides the common receiver from the same facts the
+/// resolver would establish — a NaN-boxed heap pointer above the handle band
+/// whose header is `GC_TYPE_MAP` and not forwarded — re-reads `size` and
+/// `entries` on every read (the body may `set`/`delete` and grow or compact
+/// the buffer, and the receiver itself is re-read from its root by the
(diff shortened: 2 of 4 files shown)


proggeramlug:
Merged. The hazard with an inline read inside a `for-of` body is that the body is user JS — it can `set`/`delete` and grow or compact the buffer, and it is a collection point. That is handled: `size` and `entries` are re-read on every read, the header is checked `GC_TYPE_MAP` and not-forwarded, the index is bounds-checked against the *live* size, and the receiver itself is re-read from its root by the rooted-operand group. Subclass instances, plain objects and out-of-range indices keep the runtime helpers, so the inline tier only ever takes receivers it has just proven.

Validation: codegen 1338/0 including the new `map_entry_reads_are_inline_with_the_helper_as_the_fallback`, runtime 2773/0, `native_proof_regressions` 280/0; `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green — the exception is the pre-existing `\${{ }}` artifact (#8929).

proggeramlug:
Measurement receipt (post-merge, for the record).

Quiet Mac mini, paired alternating runs, control = base `628527833`, candidate = this head, row `5k entities: 3 commands each + sync (15k total commands)`, `--repeat 64 --measured-rounds 6`:

| run | control | candidate | improvement | wins | oracles |
|---|---|---|---|---|---|
| 9-pair screen | 3.191 ms | 3.174 ms | **+0.56%** | **9/9** | 18/18 |
| 15-pair confirm | 3.190 ms | 3.174 ms | **+0.51%** | **15/15** | 30/30 |

Local merge gate on the Linux build host (perrymaster, 361 s): perry-runtime 2752 passed / 0 failed, perry-codegen 1338 / 0 (including the new `map_entry_at` test), native_proof_regressions 280 / 0, perry-transform 119 / 0, `cargo fmt` PASS, and the ratchet replay vs the merge base flat (unrooted-local 576 → 576, raw-handle baseline 967 → 967, no module ceiling raised).

PR #8963, opened 2026-08-28
fix(runtime): restore method-value receiver snapshots

### LINKED ISSUE #8955: release blocker: method-value snapshot loses receiver after property replacement
## Release evidence

The local Linux preflight found a deterministic release-gate failure in method-value snapshot behavior.

- Current main candidate: 23a211383ed62315ee6f6b68fb380b9b760a7cda
- Full 12-shard preflight candidate: a1e44c5df8e58e207b72752d1aa1e5bd9c1e91bf
- Focused test: test_parity_method_value_snapshot_bind
- The failure reproduces serially on current main.
- The earlier full sweep produced the same failure.

The fixture requires this Perry invariant from merged PR #4548:

    const captured = this.m;
    this.m = "SHADOW";
    captured(); // must retain the original receiver and return "C:t"

Current Perry instead returns C:undefined, causing:

    Error: this-snapshot regressed: C:undefined

There is also an oracle problem: Node correctly treats a plain extracted method as unbound and throws a TypeError for this fixture. Therefore this intentional Perry snapshot contract cannot use Node output as its oracle. The fixture must be classified as expected-output if the #4548 behavior remains the intended contract.

The known-failure entry calls this a #4548 recurrence and categorizes it as bug-stale. There is no open issue tracking it. The shard allowlist permits it, but aggregate gates reject it:

- parity threshold: parity/method_value_snapshot_bind is 0.0% below the 100.0% threshold
- matrix trend: status parity_fail is outside the baseline status pass; diff_lines is 15 above 0

## Reproduction

On Linux at current main with the pinned repository toolchain:

    ./run_parity_tests.sh --filter test_parity_method_value_snapshot_bind

## Required fix

- Restore the intended #4548 bound method-value snapshot behavior so the captured method retains its receiver after the own property is replaced.
- Convert the fixture to an explicit expected-output oracle whose successful result is OK.
- Keep the self-bind own/inherited, captured snapshot, arrow override, and plain Function.bind cases live.

Do not solve this by lowering the aggregate threshold or allowlisting the current failing output.

## Acceptance criteria

- The focused fixture passes deterministically on Linux with expected output OK.
- Runtime/codegen regression coverage proves the captured receiver survives own-property replacement.
- The test entry is removed from test-parity/known_failures.json.
- The aggregate threshold no longer sees a parity failure for this test.
- The intentional Node/Perry output difference is represented explicitly in the parity-matrix baseline/tooling rather than hidden.

proggeramlug:
## Summary

- restore the intentional Perry receiver snapshot for `this.method` value reads without changing canonical identity for ordinary `obj.method` reads
- preserve own-property arrow overrides while dispatching an already-captured method through the class vtable after replacement
- convert `test_parity_method_value_snapshot_bind` to an explicit `OK` expected-output oracle, remove its known-failure entry, and baseline the intentional Node/Perry matrix diff
- add runtime and codegen regression coverage

No version bump is included.

## Testing

- `./run_parity_tests.sh --filter test_parity_method_value_snapshot_bind` — PASS, 1/1, 100.0%
- `cargo test -p perry-runtime --lib object::tests::this_method_snapshot_survives_own_property_replacement -- --exact`
- `cargo test -p perry-codegen --test native_proof_regressions this_method_value_uses_receiver_snapshot_bind -- --exact`
- `cargo test -p perry-codegen --lib` — 1337 passed, 1 ignored
- `python3 -m unittest tests/test_parity_matrix_trend.py`
- `python3 scripts/parity_threshold_gate.py --report test-parity/reports/latest.json --check ...` — 100.0%
- `python3 scripts/parity_matrix_trend.py --report test-parity/reports/latest.json --check ...` — pass with the explicit 14-line baseline

Closes #8955

## Summary by CodeRabbit

- **Bug Fixes**
  - Method values accessed through `this` now preserve their original receiver and behavior when instance properties are later replaced.
  - Existing own-property overrides continue to take precedence when read before binding.
  - Prevents previously captured method values from unexpectedly switching to newer replacements.

- **Tests**
  - Added regression coverage for stable method snapshots and receiver behavior.
  - Promoted the method-value parity test from a known failure to a passing baseline.


proggeramlug:
Merged.

**One fix pushed:** the new `this_method_snapshot_survives_own_property_replacement` declares a test-local `const CLASS_ID: u32 = 0x8955`. `scripts/class_id_collisions.py` matches on the **name**, so a second `CLASS_ID` with a different value from the real mirror in `class_registry/state.rs` reads to that gate as cross-crate mirror drift — and it was red. Renamed to `SNAPSHOT_CLASS_ID` with a comment saying why, since the next person writing a test here will hit the same thing.

**On the ratchet changes** — this is a net debt reduction rather than a loosening. The deleted `known_failures` entry is `category: bug-stale` and its own text says *"Remove this entry when method-value snapshots are fixed"*, so the fix retires it exactly as instructed; the added matrix baseline is a single measured 14-line ceiling with #8955 provenance, covering the intentional Node/Perry difference for a test that now passes against an explicit oracle.

**Validation** — runtime 2774/0, codegen 1338/0, `native_proof_regressions` 281/0 (`RUST_TEST_THREADS=1`); `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green after the rename — the exception is the pre-existing `\${{ }}` artifact (#8929). I did not re-run the parity filter you reported; the runtime and codegen regressions cover the mechanism directly.

Issue #1813, opened 2026-05-25
SIGSEGV race in js_native_call_method via @perryts/mysql nativeScramble during concurrent Pool.acquire handshakes (0.5.1029 / c720d2a4)


proggeramlug:
## Summary

Follow-on from #1767 (fixed — thanks; sequential parameterized queries work now). With that fixed, **sequential** POST ingest is stable, but under **concurrent** load the `@perryts/mysql` pool opens additional connections (`Pool.acquire`), and the crash moves to the **auth handshake**: a SIGSEGV in **`js_native_call_method` during `@perryts/mysql`'s `nativeScramble`** (`mysql_native_password`).

It's a **race / heisenbug**: it's masked by gdb and by low-load `--debug-symbols` builds, and only surfaced under heavy concurrency (~`ab -c 100`). I captured it via a **core dump** of a `--debug-symbols` build, so the backtrace is symbolized.

## Environment

- `perry 0.5.1029` @ `c720d2a4`, Linux x86-64 (Ubuntu 24.04)
- `@perryts/mysql` 0.1.4, MariaDB 10.11 (**`mysql_native_password`**)
- Sequential ingest (`ab -c 1`, single pooled connection) is stable. The crash needs concurrency high enough to make the pool open **new** connections (so the handshake/auth runs repeatedly/concurrently).
- NOTE: this is the `mysql_native_password` path. I have not yet tested `caching_sha2_password` (will follow up).

## Symbolized backtrace (from core dump)

```
Thread 1 SIGSEGV
#0  js_native_call_method ()
#1  @perryts/mysql auth/native_password.ts: nativeScramble ()
#2  @perryts/mysql auth/native_password.ts: __obj_method_initialResponse_2 ()
#3  js_native_call_method ()
#4  @perryts/mysql connection.ts: sendAuthResponseForPlugin ()
#5  @perryts/mysql connection.ts: handleHandshakeV10 ()
#6  @perryts/mysql connection.ts: handlePacket ()
#7  perry_closure connection.ts
#8  js_net_process_pending ()
#9  js_stdlib_process_pending ()
#10 js_run_stdlib_pump ()
#11 @perryts/mysql pool.ts: Pool.acquire ()
#12 @perryts/mysql pool.ts: Pool.query ()
#13 chirp db.ts: query ()
#14 js_async_first_call ()
...
#19 perry_stdlib::fastify::server::call_closure2_catching ()
[...]

PR #8964, opened 2026-08-28
perf(codegen): inline the hot-cache lookup on Apple aarch64 for the implicit-this pair and the arena-state resolution (ECS round 4)

proggeramlug:
ECS round 4, item 4p (general mechanism, no benchmark-specific special cases).

## What

On Apple aarch64 the generated code now performs the per-thread hot-cache lookup itself for two of the hottest runtime calls, and only falls back to the runtime call on a miss:

- **`js_implicit_this_set` save/restore pair** around every dynamically-dispatched method call — now a load/store of the `HotTls.implicit_this` inline field (the cell moves from a generic hot slot into a dedicated inline field; the runtime's own `js_implicit_this_get/set` and the root scanner use the same cell).
- **`js_inline_arena_state` first-use resolution** in every allocating activation — now a load of `HotTls.inline_state` with a null-`data` check; the runtime call remains the fallback for the uninitialised state.

The lookup is the same sequence the runtime uses: load the exported `PERRY_HOT_TSD_KEY` (`−1` → slow path), `mrs tpidrro_el0 & ~7`, index the TSD slot, null → slow path. `mrs` is emitted as `asm sideeffect` (never hoisted across a possible thread switch) and the root-lowering passes already treat it as a non-collecting leaf. `offset_of!` assertions pin the two offsets the codegen hard-codes (`inline_state` @ 8, `implicit_this` @ 128) and `InlineArenaState.data` @ 0.

Other targets are unchanged; `PERRY_INLINE_HOT_TLS=0` restores the runtime calls on Apple aarch64.

## Tests

- new `lower_call::alloc_hot_tests::inline_arena_state_is_resolved_through_the_hot_cache_on_apple_aarch64` — asserts the `mrs`, the `@PERRY_HOT_TSD_KEY` load, the `arena_state.hot_tls.ready` block and the retained fallback call on the Apple triple, and that a Linux triple emits no `tpidrro_el0`.
- `tls_hot` offset pins (compile-time), `this_binding` + `tls_hot` runtime suites serial, full runtime/codegen/native-proof/transform suites + lint/ratchet replay on the build host (receipt below).

## Measurement

Paired alternating runs on the quiet Mac mini, control = base `628527833`, candidate = this head; row `5k entities: 3 commands each + sync (15k total commands)` (numbers below).

## Summary by CodeRabbit

* **Performance**
  * Improved Apple aarch64 performance through inline thread-local access for arena state and implicit-`this` handling.
  * Runtime fallbacks remain available when cached data is unavailable.
* **Compatibility**
  * Other targets continue using the existing runtime access path.
  * Inline hot-TLS behavior can be disabled with `PERRY_INLINE_HOT_TLS=0`.
* **Tests**
  * Added coverage for optimized and fallback behavior across supported targets.


proggeramlug:
Local merge gate on the Linux build host (perrymaster, isolated clone + target), head `7e10e69ef` vs base `628527833`:

| suite | result |
|---|---|
| perry-runtime (lib) | **2752 passed / 0 failed**, serial (`--test-threads=1`) |
| perry-codegen (lib) | 1338 passed / 0 failed — includes the new `inline_arena_state_is_resolved_through_the_hot_cache_on_apple_aarch64` |
| `native_proof_regressions` | 280 passed / 0 failed |
| `cargo fmt` | PASS |
| thread-local policy | OK (266 hot declarations, 124 raw blocks, capacity 768) |

Targeted checks for the cell that moved: `this_binding` 10/10, `tls_hot` 3/3, and the implicit-`this` GC-root scanner tests (`side_table_scanners`, the #1813 relocation assertions) 11/11 — all serial.

Two parallel-run failures were investigated and are **not** from this change: `gc::roots::stack_maps::decode_tests::…::discovers_a_map_from_a_later_loaded_shared_object` (a `dlopen` + image-walk test that also fails in parallel runs on pristine `origin/main`) and `async_hooks::test_support::tests::native_async_resource_accepts_string_and_symbol_expandos` (process-global symbol side tables). Both pass alone, and the full **serial** suite above is clean at 2752/0.

On why the inline path is equivalent: `js_implicit_this_set` is a pure cell swap — no write barrier, no root registration — and `scan_implicit_this_roots_mut` already scans only the calling thread's cell. The accessors, the scanner and the generated code now all address the same `HotTls::implicit_this` cell, so the root stays exactly as visible to GC as before; the inline store removes a call rather than introducing a safepoint.

Measurement on the quiet Mac mini follows in a separate comment.

proggeramlug:
Measurement on the quiet Mac mini — paired alternating runs, control = base `628527833`, candidate = this head, row `5k entities: 3 commands each + sync (15k total commands)`, `--repeat 64 --measured-rounds 6`:

| run | control | candidate | improvement | wins | oracles |
|---|---|---|---|---|---|
| 9-pair screen | 3.192 ms | 3.158 ms | **+1.06%** | **9/9** | 18/18 |
| 15-pair confirm | 3.190 ms | 3.156 ms | **+1.06%** | **15/15** | 30/30 |

Every artifact passed the process oracle (`# bun-shim: 7 tests | 1 passed | 0 failed | 6 skipped`).

The mechanism is verified present in the shipped artifact rather than inferred from the timing: disassembling both benchmark binaries, `mrs xN, TPIDRRO_EL0` goes **10730 → 11486 (+756)** — the inline lookups the codegen now emits. Worth stating explicitly because this change is gated on the target triple, so a version that silently failed to fire would still produce a plausible-looking measurement.

proggeramlug:
Rebased onto current main (`45819777d`) — the only conflict was the two added module declarations in `crates/perry-codegen/src/expr/mod.rs` (this branch's `hot_tls`, #8960's `map_entry_at_tests`); both are kept. Head is now `ffb94e17b`.

Gate re-run on the build host against the new base:

| suite | result |
|---|---|
| perry-runtime (lib) | **2753 passed / 0 failed**, serial |
| perry-codegen (lib) | 1339 passed / 0 failed |
| `alloc_hot` + `map_entry_at` | 12 passed / 0 failed |
| `cargo fmt` | PASS |
| ratchets vs `45819777d` | unrooted-local 576 → 576, raw-handle baseline 967 → 967, no ceiling raised |

The measurement above stands unchanged — it isolates this mechanism against base `628527833`, and the rebase carries no functional change.

proggeramlug:
Merged, rebased onto current `main` (the conflict was two module declarations landing at the same spot in `expr/mod.rs` — kept both).

**On the GC question**, which is what I went after: `implicit_this` holds a NaN-boxed receiver, so relocating it out of a `thread_local!` into a `HotTls` inline field risks leaving a live receiver unrooted. It is handled — `scan_implicit_this_roots_mut` moved to the same `implicit_this_cell()` accessor the getters and setters use, visiting via `visit_nanbox_u64_slot` with write-back on relocation. The `gc_runtime_root_holders` gate agrees: OK at 893 holders (down from 897 as the thread-locals became inline fields), so the cell is still classified rather than silently unscanned.

**One fix pushed: the test this PR's own comment promises did not exist.** `tls_hot.rs:170` says "see `hot_tls_layout_is_what_codegen_assumes`", but that name appears nowhere else in the tree. The gap is real and one-sided:

- `const _: () = assert!(offset_of!(HotTls, implicit_this) == HOT_TLS_IMPLICIT_THIS_OFFSET)` pins the **runtime** constant to the struct, so a field reorder fails to compile. Good.
- But codegen's offsets are separate **string literals** in a crate that does not depend on `perry-runtime`. The natural response to that failed assert is to update the runtime constant — which compiles, while `hot_tls.rs` keeps emitting `"128"`. Generated code would then load the wrong field of a GC-visible cell, silently.

So I wrote the named test: it reads `HOT_TLS_*_OFFSET` out of `tls_hot.rs` source (no dependency edge available) and asserts the codegen strings match. **Sabotage-checked** — drifting the codegen constant 128 → 136 fails it with "codegen emits a stale HotTls::implicit_this offset"; restoring passes.

**Validation** — codegen 1339/0 (including the new test), runtime 2774/0, `native_proof_regressions` 281/0; under `PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1` the failing count is 16, the pre-existing set; `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green — the exception is the pre-existing `${{ }}` artifact (#8929).

PR #8965, opened 2026-08-28
perf(runtime,codegen): SSO computed keys + megamorphic write stub cache (dynamic-property overwrite 5.3×)

proggeramlug:
Computed string keys (`o["k" + i]`) stopped defeating every property cache in the pipeline.

**Dynamic-property overwrite loop: ~450 ms → ~86 ms (5.3×)**, 8/8 interleaved A/B pairs at stable load; node on the same host runs it in 28–30 ms, so this workload goes from ~16× node to **~3.0× node**. A second 10/10 run under heavy co-tenant load (everything inflated ~1.8×) reproduced the same ratio. Binary size **+4 KB (+0.04%)** — runtime-side, no generated-code growth.

### Two changes that only pay off together

**1. `js_string_concat_value_box` — SSO result when it fits.** The fused string+number concat now returns an inline SSO immediate for results of ≤ 5 ASCII bytes instead of always minting a `StringHeader`. That removes a heap allocation per iteration from the hot `"prefix" + i` shape, but the bigger effect is that the result's *bits* become content-stable, so caches that compare key values can hit. ASCII-only, for the same `utf16_len` soundness reason as `js_string_concat_box`'s existing SSO arm.

**2. A megamorphic stub cache for dynamic string-keyed writes.** A site's inline IC holds `DYN_IC_WAYS = 3`; a loop rotating 500 computed keys through it evicts permanently and pays the full miss walk every write. Added a thread-local 4096-way direct-mapped cache keyed on `(shape_token, key_bits)` — V8's megamorphic stub pattern — probed after the per-site ways miss and fed at every prime site, **including for overflow slots** (a wide object keeps every data property there, so gating on the inline region would starve the cache for exactly the receivers it exists for).

Supporting: the SSO→heap materialization every `*const StringHeader` consumer crosses now interns instead of minting, so a computed key's *address* is stable too and address-keyed read plans can hit; short heap keys are folded to their SSO bits before keying.

### Safety

The stub stores **only content-derived bits, never an address**; keys that don't fit the inline form are rejected rather than cached under their pointer. This is load-bearing, not conservatism: `dyn_ic_try_store` revalidates the receiver's current shape token, blocking flags and slot bound on every hit, but it confirms the *shape*, not that the cached *slot* belongs to this *key*. A pointer-keyed entry could be primed, evicted from the (direct-mapped, evict-on-collision) intern table, collected, and its address recycled by an unrelated string whose write would hit the stale entry and overwrite the wrong slot. Content-only keying removes that class entirely and leaves the table holding no GC roots (registered in `gc_runtime_root_holders.json` with that verdict).

### Method note — two wrong verdicts before the right one

The stub alone measured as a **wash, twice**, and no profile explained why. Counters did, in two stages:

- 600k inserts, 1.2M probes, **0 hits**. Splitting the probe-miss counter by cause (empty / token / key) showed 99% key-mismatch: the keys were fresh heap pointers each iteration, so a value-keyed cache could never hit at any capacity.
- After the SSO fix: still only 9,584 hits of 1.19M. The way index XOR'd the **low** bits, and an SSO key's low bits are its **first byte** — `"k0".."k499"` collapsed onto 125 ways with buckets 10 deep, evicting each other continuously. Multiplicative mixing over the top bits gives 480/500 distinct ways, and that single change is what produced the 5.3×.

### Verification

- `perry-runtime` **2774 passed / 0 failed**; `perry-codegen` **1340 passed / 0 failed**.
- Differential vs node on computed-key behaviour — delete/re-add key ordering, `Map`/`Set` keys, the 5→6 byte SSO boundary, non-ASCII, floats, negatives, 1e21, numeric-looking keys: **byte-identical output**.
- Two codegen IR tests updated to expect the `_box` twin; their intent (fused concat, no tag test) is unchanged.

`benchmarks/bench_populated_delete.ts` (the ~200× populated-delete gap) is unaffected by this path and remains open for dictionary mode.

## Summary by CodeRabbit

- **Performance**
  - Improved dynamic property writes using computed string keys, reducing repeated write overhead.
  - Short string concatenation can now avoid heap allocation in common cases.
  - Reused equivalent short-string values to improve cache effectiveness.

- **Diagnostics**
  - Added an optional hardened allocator mode for investigating memory-related issues.

### DIFF
--- a/changelog.d/8952-sso-computed-keys-write-stub.md
+++ b/changelog.d/8952-sso-computed-keys-write-stub.md
@@ -1,0 +1,52 @@
+Computed string keys (`o["k" + i]`) stopped defeating every property cache in
+the pipeline: **the dynamic-property overwrite loop goes from ~450 ms to
+~86 ms — 5.3×**, 8/8 interleaved A/B pairs at stable load, against 28–30 ms
+for node on the same host (so ~16× node → ~3.0× node). Binary size +4 KB
+(+0.04%); no generated-code growth.
+
+Two changes that only pay off together.
+
+**1. `"str" + n` returns SSO when the result fits.** The fused concat's
+NaN-box-returning twin (`js_string_concat_value_box`) assembles results of ≤ 5
+ASCII bytes as an inline SSO immediate instead of a fresh `StringHeader`. That
+removes one heap allocation per iteration from the hot `"prefix" + i` shape —
+but the bigger effect is that the result's BITS become content-stable: `"k" +
+42` now yields the identical `f64` every evaluation, so every downstream cache
+that compares key *values* can finally hit. ASCII-only, for the same
+`utf16_len` soundness reason as `js_string_concat_box`'s existing SSO arm.
+
+**2. A megamorphic stub cache for dynamic string-keyed writes.** A dynamic
+write site's inline IC holds `DYN_IC_WAYS = 3`; a loop rotating 500 computed
+keys through one site evicts permanently and pays the full miss walk on every
+write. Added a thread-local 4096-way direct-mapped cache keyed on
+`(shape_token, key_bits)` — V8's megamorphic stub pattern — probed after the
+per-site ways miss and fed at every prime site, including for overflow slots
+(a wide object keeps *every* data property there, so gating on the inline
+region would starve the cache for exactly the receivers it exists for).
+
+Supporting: the SSO→heap materialization every `*const StringHeader` consumer
+crosses now interns instead of minting, so a computed key's ADDRESS is stable
+too and the address-keyed read plan can hit; short heap keys are folded to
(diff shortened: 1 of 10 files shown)


proggeramlug:
Merged (batched with #8966). The two changes do depend on each other — an SSO result from the fused concat is what lets the megamorphic write stub cache hit at all — and both are additive fast paths with the existing routes as the fallback.

Validation: codegen 1341/0, runtime 2779/0 (`RUST_TEST_THREADS=1`); evac failing set 16, the pre-existing one; `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green — the exception is the pre-existing `\${{ }}` artifact (#8929).

The 5.3× is not re-measured here.

Issue #8953, opened 2026-08-28
Array subclass instances: getOwnPropertyNames / for..in segfault; Object.keys leaks "length" and "fill"


proggeramlug:
Found while probing `class X extends Array` semantics for the elements-store redesign (perry 0.5.1519, main `23a211383`, arm64 macOS release build, `--no-auto-optimize`). All three are node-divergent; two are crashes.

**1. `Object.getOwnPropertyNames` on an instance with ≥1 element — SIGSEGV (rc 139)**
```js
class A extends Array {}
const a = new A();
a.push(1, 2, 3);
console.log(Object.getOwnPropertyNames(a).join(","));   // node: 0,1,2,length   perry: segfault
```

**2. `for..in` over an instance with ≥1 element — SIGSEGV (rc 139)**
```js
class A extends Array {}
const a = new A();
a.push(1);
let s = ""; for (const k in a) s += k + ","; console.log(s);   // node: 0,   perry: segfault
```

**3. Enumeration leaks non-enumerable / non-own keys**
```js
class A extends Array {}
const a = new A();
a.push(1, 2, 3);
console.log(Object.keys(a).join(","));                   // node: 0,1,2   perry: 0,1,2,length,fill
let s = ""; for (const k in new A()) s += k + ","; console.log(s);   // node: (empty)   perry: length,fill,
```
`length` is an own but non-enumerable property (must not appear in `keys`/`for..in`); `fill` is not an own property at all — it looks like the `Array.prototype.fill` installer (or the subclass prototype stamping) lands it as an enumerable own key on the instance shape.

Context: instances are object-backed today (indexed elements and `length` are shape-carried properties, `array/subclass.rs`, `object/array_tail_transition.rs`); the enumeration paths walk that shape. I'm designing a real elements store for these instances (`ObjectMeta.elements`, design note in the ECS handoff repo) and will cover all three in its e2e semantics suite; filing separately so the crashes are visible to whoever touches the enumeration paths first.

PR #8966, opened 2026-08-28
runtime: a real elements store for class X extends Array (−12% on the wolf-ecs twins, gated off)

proggeramlug:
## What

`class X extends Array` instances are ordinary `GC_TYPE_OBJECT`s whose indexed elements **and `length`** are shape-carried properties, so every `push`/`pop`/`obj[i] = v` is a property-shape transition (`array/subclass.rs`, `object/array_tail_transition.rs`). This adds the alternative: `ObjectMeta.elements`, a real `GC_TYPE_ARRAY` holding the instance's elements, with every entry point routed to it.

Behind `PERRY_ARRAY_SUBCLASS_ELEMENTS=1`, **default off** — with the gate off this is a no-op (one unused meta word, one extra traced edge that is always null).

* `ObjectMeta.elements`: `*mut ArrayHeader` bits (0 = none), zeroed by both meta-ensure paths, a traced child edge exactly like `spill` (second explicit meta slot in the child-slot iterator; visited and rewritten in the `ObjectMeta` arm).
* `array/subclass_elements.rs`: the gate, the accessor, the barriered head write-back, installation at `super()`, and the property funnel (`ElementsKey`, get/set/has/delete, key/value/entry lists, descriptors, `deopt_to_shape`).
* Hot entries (`array_subclass_fast_length/_index_get/_index_set/_push_one/_pop`, `js_packed_arraylike_index_get`) operate on the inner array; `push` roots the owner across the re-allocating append and writes the head back through the barriered meta slot.
* Property funnel: get/set by name (+ IC miss, both tails), `hasOwnProperty`/`in`, `delete`, `getOwnPropertyDescriptor`, `Object.keys` / `getOwnPropertyNames` / `values` / `entries`, `JSON.stringify`, iteration/spread (live inner array instead of a snapshot), `length` writes with Array semantics.
* Exotic operations leave the representation for good — `defineProperty` on an index or `length`, `defineProperties`, `freeze`/`seal`/`preventExtensions`, `setPrototypeOf` call `deopt_to_shape`, which materialises every element as a shape-carried property and detaches the store. The whole semantic long tail keeps running on the existing, tested machinery.
* Codegen: the `arrlike.ic.*` read tier and the `plen.ic.*` length tier probe the store first (`ObjectMeta.elements` is word 12, const-asserted in the runtime); a probe miss keeps the shape IC unchanged.

## Numbers (Mac mini, 11 alternating pairs, same binary, gate off → on)

|---|---|---|
| 2 s | 0.4100 → **0.3624** (**−11.61%**, 11/11) | 0.3402 → **0.2992** (**−12.02%**, 11/11) |
| 50 ms | 0.4099 → **0.3621** (**−11.63%**, 11/11) | 0.3404 → **0.2988** (**−12.18%**, 11/11) |

Deltas are tight (±0.2%). Perry vs Node 26.5 on the same host (0.1337 / 0.1492): **3.02× → 2.71×** and **2.24× → 2.01×**. Construction also gets much cheaper (Linux warm-up probe: setup 15.3 → 2.2 ms, first call 13.7 → 1.3 ms).

Why it pays here: wolf-ecs's `Archetype` constructor does `this.sset.packed = this`, so the hot `packed.push/pop/[i]` run on the subclass instance itself — ~25% of the cycle was the shape-transition machinery (`array_subclass_fast_push_one_validated` + the tail-transition cache + `install_cache_carried_object_shape_version` + `dense_slot_exists` + `js_array_pop_f64`).

## Two cliffs worth knowing about (both fixed here)

Gate-on was **6× slower** at first. Neither cause was the representation:

1. `js_packed_arraylike_loop_guard` demanded a dense **shape** layout, so every versioned loop over such a receiver declined and fell back to string-keyed property reads. It now resolves an elements-backed receiver to its inner array and runs the ordinary plain-array admission; the revalidation entry re-resolves each iteration, so a re-allocating append side-exits on the stale facts exactly as a grown plain Array does.
2. `js_array_subclass_init` installs `fill` **with a descriptor**, so every Array-subclass instance is descriptor-bearing and `js_object_get_field_ic_miss` refuses to prime the PIC without a named-prefix token — and that token was derived from the same dense layout, which `build_dense_layout` cannot produce for an elements-backed instance (it locates `length` by name and the store owns `length`). The PIC was therefore never primed and every declared-field read missed forever: a per-key census counted **17.8M misses on `change`/`sset`/`mask` in a 2 s run**. The token builder now synthesises "no numeric tail" for the elements form, with the declared-prefix comparison and accessor check unchanged as the authority.
   * Corollary trap: hiding `fill` behind a non-enumerable descriptor (the obvious fix for the `Object.keys` leak) sets `OBJ_FLAG_HAS_DESCRIPTORS`, which the codegen class-field inline guard rejects — that alone is a 6× regression. The right fix for that key is to install `fill` on the class prototype (where node has it); not done here.

## Semantics

Gate-on is strictly **closer** to node than today's representation. On a heavy probe (`sort`, `reverse`, `splice`, `shift`, `unshift`, `concat`, `slice`, `indexOf`, `includes`, `at`, `map`, `filter`, `reduce`, `join`, `forEach`, `length` truncation, sparse construction, spread, `Array.from`, `Array.prototype.slice.call`, `JSON.stringify`) gate-on matches node exactly, while gate-off prints the object form (`{"0":5,"1":1,"length":2}`) for every one of them. `Object.keys` no longer leaks `length`.

Two pre-existing bugs fixed on the way (both reproduce on main, gate irrelevant):
* `Object.getOwnPropertyNames` on a subclass instance **segfaulted** — the array path was gated on `Array.isArray`, which is true for an instance whose cell is an `ObjectHeader`, and dereferenced a null cleaned array pointer (#8953).
* `class X extends Array { constructor(...a) { super(...a) } }` never ran the Array subclass init at all — the spread `super` form had no Array-parent arm, so the instance had no `length` and no Array surface (`new X(2).length` → `undefined`, `JSON.stringify(x)` → `{}`).

Still divergent from node, identically in both gates (pre-existing, tracked in #8953): `fill` in `getOwnPropertyNames`, `A.from([…])` not populating, `f.constructor === A` after `map`, `b.entries()`/`b.keys()` unimplemented on a subclass instance.

## Verification

- Runtime unit tests: the GC edge across a forced-evacuation minor, hot-entry routing (40 appends across a GC through both entry families), the property funnel end to end, freeze deopting to the shape-carried form, and the loop-guard admission (including the side-exit after a re-allocating append). Codegen IR census tests pin both elements probes.
- `RUSTFLAGS=-D warnings cargo check --workspace --all-targets` clean; `cargo test -p perry-codegen -p perry-transform -p perry-hir` 2515/0; `cargo test -p perry-runtime --lib` 2779/0; file size, GC store-site inventory, raw-handle debt (unchanged), shape census, local-binding audit, addr-class audit all pass.
- Array-subclass integration tests are being re-run **with the gate on** (they pass with it off, and the first ones are green); I'll post the sweep.

## Landing

Land as-is (gate off ⇒ no behaviour change), then a follow-up flips the default once the full corpus has been swept gate-on. Design note and the full investigation live in the campaign repo (`ARRAY_SUBCLASS_ELEMENTS_DESIGN.md`).

## Summary by CodeRabbit

- **New Features**
  - Added optimized storage for `Array` subclass elements and `length`.
  - Improved performance for indexed access, updates, `push`, `pop`, and `fill`.
  - Array subclasses now work consistently with property checks, descriptors, enumeration, JSON serialization, and counted loops.
  - Spread-based `super()` construction for `Array` subclasses now initializes array behavior correctly.

- **Bug Fixes**
  - Corrected handling of holes, bounds, missing properties, and prototype-chain lookups.
  - Preserved expected behavior when advanced object operations require converting optimized storage.

### DIFF
--- a/crates/perry-codegen/src/expr/index_get/inline_dyn_typed_array.rs
+++ b/crates/perry-codegen/src/expr/index_get/inline_dyn_typed_array.rs
@@ -438,9 +438,90 @@ pub(super) fn lower_inline_dyn_typed_array_get(
     ctx.block()
         .cond_br(&header_ok, &object_brand_label, &object_miss_label);
 
+    // An elements-backed Array-subclass instance (`ObjectMeta.elements`,
+    // perry-runtime `array/subclass_elements.rs`): its indexed elements live
+    // in a real Array hanging off the meta record, so the read is the plain
+    // Array read on that inner array — no shape IC, no family token. A miss
+    // of this probe (no meta, no store) is the shape-carried form and keeps
+    // the IC below; an out-of-bounds index or a hole goes to the complete
+    // dispatcher (prototype chain).
+    let elem_meta_idx = ctx.new_block("arrlike.elem.meta");
+    let elem_store_idx = ctx.new_block("arrlike.elem.store");
+    let elem_bounds_idx = ctx.new_block("arrlike.elem.bounds");
+    let elem_load_idx = ctx.new_block("arrlike.elem.load");
+    let elem_value_idx = ctx.new_block("arrlike.elem.value");
+    let elem_meta_label = ctx.block_label(elem_meta_idx);
+    let elem_store_label = ctx.block_label(elem_store_idx);
+    let elem_bounds_label = ctx.block_label(elem_bounds_idx);
+    let elem_load_label = ctx.block_label(elem_load_idx);
+    let elem_value_label = ctx.block_label(elem_value_idx);
     ctx.current_block = object_brand_idx;
     ctx.block()
-        .cond_br(&is_array, &object_array_guard_label, &object_shape_label);
+        .cond_br(&is_array, &object_array_guard_label, &elem_meta_label);
+    ctx.current_block = elem_meta_idx;
+    let elem_meta_addr = ctx.block().add(I64, &object_raw, &meta_offset);
+    let elem_meta_slot_ptr = ctx.block().inttoptr(I64, &elem_meta_addr);
+    let elem_meta_loaded = ctx.block().load(
+        if meta_ptr_size == 4 { I32 } else { I64 },
+        &elem_meta_slot_ptr,
+    );
+    let elem_meta_i64 = if meta_ptr_size == 4 {
+        ctx.block().zext(I32, &elem_meta_loaded, I64)
+    } else {
+        elem_meta_loaded
+    };
+    let elem_has_meta = ctx.block().icmp_ne(I64, &elem_meta_i64, "0");
+    ctx.block()
+        .cond_br(&elem_has_meta, &elem_store_label, &object_shape_label);
+    ctx.current_block = elem_store_idx;
+    let elem_meta_ptr = ctx.block().inttoptr(I64, &elem_meta_i64);
+    // `ObjectMeta.elements` is word 12 (offset 96; pinned by a const assert
+    // in perry-runtime `object/mod.rs`).
+    let elem_store_slot_ptr = ctx.block().gep(I64, &elem_meta_ptr, &[(I64, "12")]);
+    let elem_store_i64 = ctx.block().load(I64, &elem_store_slot_ptr);
+    let elem_has_store = ctx.block().icmp_ne(I64, &elem_store_i64, "0");
+    ctx.block()
+        .cond_br(&elem_has_store, &elem_bounds_label, &object_shape_label);
+    ctx.current_block = elem_bounds_idx;
+    let elem_type_addr = ctx.block().sub(I64, &elem_store_i64, "8");
+    let elem_type_ptr = ctx.block().inttoptr(I64, &elem_type_addr);
+    let elem_type = ctx.block().load(I8, &elem_type_ptr);
+    let elem_is_array = ctx.block().icmp_eq(I8, &elem_type, "1");
+    let elem_flags_addr = ctx.block().sub(I64, &elem_store_i64, "7");
+    let elem_flags_ptr = ctx.block().inttoptr(I64, &elem_flags_addr);
+    let elem_flags = ctx.block().load(I8, &elem_flags_ptr);
+    let elem_fwd = ctx.block().and(I8, &elem_flags, "128");
+    let elem_not_fwd = ctx.block().icmp_eq(I8, &elem_fwd, "0");
+    let elem_store_ptr = ctx.block().inttoptr(I64, &elem_store_i64);
+    let elem_length = ctx.block().load(I32, &elem_store_ptr);
+    let elem_length_i64 = ctx.block().zext(I32, &elem_length, I64);
+    let elem_in_bounds = ctx.block().icmp_ult(I64, &object_idx_i64, &elem_length_i64);
+    let elem_ok = ctx.block().and(I1, &elem_is_array, &elem_not_fwd);
+    let elem_ok = ctx.block().and(I1, &elem_ok, &elem_in_bounds);
+    ctx.block()
+        .cond_br(&elem_ok, &elem_load_label, &object_miss_label);
+    ctx.current_block = elem_load_idx;
+    let elem_bytes = ctx.block().shl(I64, &object_idx_i64, "3");
+    let elem_elements_addr = ctx.block().add(I64, &elem_store_i64, "8");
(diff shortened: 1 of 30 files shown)


proggeramlug:
Merged. **One fix pushed:** `subclass_elements.rs` open-coded the `StringHeader` payload offset, raising the string-payload-access ratchet 364 → 365. `crate::object::string_header_payload` already exists for exactly this; reusing it puts the count back to baseline.

**One finding worth your attention, not a blocker.** With `PERRY_ARRAY_SUBCLASS_ELEMENTS=1` the runtime suite is **2766 passed / 13 failed**. Its own `subclass_elements` tests are 5/5, and with the gate off everything is 2779/0 — so `main` is unaffected and the default is genuinely a no-op, as you describe.

All 13 are `array::subclass_tests::*`:

```
dense_array_subclass_reads_slots_until_its_shape_changes
dense_array_subclass_tail_transitions_reuse_exact_shapes_and_slots
dense_array_subclass_tail_transition_edges_survive_moving_gc
array_subclass_length_ic_publishes_only_scalar_exact_or_family_facts
packed_numeric_proof_survives_pointer_free_index_swap
…and 8 more
```

They are the **old** shape-carried representation's suite, so they assert precisely the behaviour the new mode replaces — structurally the same situation as `promote_in_place` failing under `PERRY_GC_FORCE_EVACUATE`, not 13 unrelated breakages. So this reads as expected rather than alarming.

It does mean the ON path is not suite-clean today, which matters for CLAUDE.md's knob kill-policy: a knob wants either a required CI arm exercising a state, or a deletion plan after a release of soak. Right now OFF is exercised by the whole suite and ON only by its own five tests. Conditioning the `subclass_tests` on the knob — so both representations have a green arm — would make the ON state a decision that has actually been made rather than a mode nobody has run end to end.

**Validation** — codegen 1341/0, runtime 2779/0 with the gate off (`RUST_TEST_THREADS=1`); under `PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1` the failing count is 16, the pre-existing set; `check_gc_env_knobs` ok; `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green after the payload fix — the exception is the pre-existing `${{ }}` artifact (#8929).

proggeramlug:
### Gate-ON sweep of the Array-subclass integration tests

| test file | result with `PERRY_ARRAY_SUBCLASS_ELEMENTS=1` |
|---|---|
| `issue_8655_array_subclass_indexing` | 2/2 |
| `issue_array_subclass_super_init` | 1/1 |
| `issue_4908_subclass_native_member_base` | 2/2 |
| `issue_8690_loop_versioned_arraylike` | 3/3 |
| `issue_8773_closure_capture_packed_loops` | 4/4 |
| `issue_5139_object_arraylike_method_dispatch` | 3/3 |
| `interface_typed_arraylike_method_dispatch` | 3/3 |
| `issue_8897_field_push_writeback` | 3/3 |
| `issue_8772_short_packed_spread` | 3/5 — the two failures are the stale `js_method_direct_shape_class` assertions that fail on main with the gate off too; #8967 fixes them |

The sweep caught one real bug in this PR, now fixed (`the counted-loop guard declines an elements-backed receiver`): admitting such a receiver as kind 1 with the inner array's address is only half a contract, because both admitted kinds describe storage reachable from the RECEIVER address and the emitted loop derives some reads from it — element 0 came back as the object's `meta` word read as a double (`1.8e-311`), which `issue_8773`'s dense case caught. Those loops now decline and stay generic; a dedicated kind that publishes the store address (with codegen reading through it) is the follow-up.

Re-measured after that fix — Mac mini, 11 alternating pairs, same binary, gate off → on:

|---|---|---|
| 2 s | 0.4114 → **0.3644** (**−11.41%**, 11/11) | 0.3399 → **0.2999** (**−11.80%**, 11/11) |
| 50 ms | 0.4110 → **0.3640** (**−11.49%**, 11/11) | 0.3399 → **0.2995** (**−11.91%**, 11/11) |

i.e. the correctness fix cost ~0.2 pp of the win. Perry vs Node stays ≈2.71× / 2.01× (from 3.02× / 2.24×).

PR #8947, opened 2026-08-28
codegen: inline the multi-arm direct-method receiver probe

proggeramlug:
## What

Multi-arm direct method call sites resolve the live receiver's `(class_id, ShapeId)` before their compare chain. Three of them still did it through the runtime probe `js_method_direct_shape_class` — the subclass-arm form of `emit_guarded_direct_method_call`, the indexed-dispatch shape probe (`dynamic_dispatch.rs`) and the short-spread method probe (`call_spread_short.rs`) — i.e. a call, a heap-address classification and the flag resolution for what is a handful of loads.

`emit_inline_direct_method_shape_probe` (`lower_call/method_override.rs`) emits the same decision inline: the process-wide and per-method prototype latches, the tagged-or-raw pointer test against the target's heap band, the packed header word masked to `GC_TYPE_OBJECT` / not forwarded / no descriptors / no packed-numeric proof (the mask the single-arm inline guard `emit_inline_direct_method_shape_guard` already uses), then the `class_id` and ShapeId loads with two `select`s that keep the runtime's zero-on-decline contract — `dynamic_dispatch.rs` relies on the zero to route declined receivers to the runtime fallback. The compare chains after the probe are untouched. No runtime changes.

## Why

Fresh symbol-level profile of the wolf-ecs twins (Linux `perf`, 10 s): `js_method_direct_shape_class` is 2.2% (entity cycle) / 2.6% (add/remove) of self time; `addComponent$pshape`, `removeComponent$pshape` and `destroyEntity$pshape` each carry the probe on their `sset.add`/`sset.remove` sites (36 sites in the module).

## Verification (local, per the campaign rule)

- `call_spread_short_tests` now pins `method_probe.read` present and no `@js_method_direct_shape_class(` call at the short-spread site; `proven_this_routing_tests` unchanged. `cargo test -p perry-codegen`: 1813 passed / 0 failed.
- Full gate run (`-D warnings`, transform/hir suites, file size, GC store inventory, raw-handle debt, census, local-binding, addr-class) and the Mac-mini paired screens (current main `6d10e8a1d` → this PR, 2 s + 50 ms, 11 pairs) are running; posting both as comments.

https://claude.ai/code/session_019WVcWKmYsUBnnFB7nBgbBJ

## Summary by CodeRabbit

- **Performance**
  - Improved direct method dispatch by resolving receiver information inline, reducing reliance on runtime probing.
  - May improve execution speed for subclass calls, indexed dispatch, and short-spread method calls.

[...]

PR #8967, opened 2026-08-28
test(perry): short-spread integration tests assert the inline receiver probe (#8947 follow-up)

proggeramlug:
Main is red on `issue_8772_short_packed_spread`: two assertions still require `call i32 @js_method_direct_shape_class(`, which #8947 replaced with the inline `method_probe.*` sequence at that site. #8947 updated the codegen unit tests (`call_spread_short_tests.rs`) but missed these two integration assertions — my miss, sorry.

They now assert the same property on the new shape: `method_probe.read` present and no `@js_method_direct_shape_class(` call. Everything else in the file (node-identical output under both GC modes, the direct target blocks, the class-shape globals) is untouched.

Verified: `cargo test -p perry --test issue_8772_short_packed_spread` → 5 passed / 0 failed (it fails 2/5 on main without this).

## Summary by CodeRabbit

* **Tests**
  * Improved end-to-end coverage for short packed spread scenarios.
  * Tests now verify that receiver method probes are emitted using the expected inline representation, helping detect regressions in generated output.

### DIFF
--- a/crates/perry/tests/issue_8772_short_packed_spread.rs
+++ b/crates/perry/tests/issue_8772_short_packed_spread.rs
@@ -226,7 +226,12 @@ fn repro_has_direct_empty_and_one_arms_and_matches_node_under_moving_gc() {
         .expect("read main LLVM IR");
     let invoke = function_ir(&ir, "__invoke(");
     assert!(invoke.contains("call i32 @js_short_packed_spread_values("));
-    assert!(invoke.contains("call i32 @js_method_direct_shape_class("));
+    // #8947: the receiver's (class_id, ShapeId) probe is emitted INLINE
+    // (`method_probe.*`) instead of calling `js_method_direct_shape_class`.
+    assert!(
+        invoke.contains("method_probe.read") && !invoke.contains("@js_method_direct_shape_class("),
+        "the short-spread method probe must be inline:\n{invoke}"
+    );
     let direct = named_blocks(
         invoke,
         &[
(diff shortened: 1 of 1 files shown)


proggeramlug:
Merged. Verified green: **5 passed / 0 failed** on `issue_8772_short_packed_spread` with this PR applied.

**A note on how I nearly got this wrong**, since it may cost you time too. My first three runs of this suite all reported 5 failures, and none of them were real — every one was

```
  library build: v0.5.1519 (commit 5d7f4e8e549a)
  Perry build:   v0.5.1519 (commit 7788453484fd)
  library: …/debug/libperry_runtime.a
```

The harness links `<target>/debug/libperry_runtime.a` specifically. Building the `-static` wrappers with `--profile perry-dev` puts them in `perry-dev/`, and setting `PERRY_RUNTIME_DIR` does not redirect it — so the test compiles against a stale archive and fails its own "compile succeeded" assertion at line 78, which looks exactly like the assertion failure you would expect. The real signal only appeared after `cargo build -p perry-runtime-static -p perry-stdlib-static` with the **default** profile.

Same reason I want to flag the underlying gap: `scripts/run_lint_gates.sh` covers the `lint` job plus a compile tier, but **not** `crates/*/tests/*.rs` — those run per-PR only via `e2e-scoped`, i.e. when the diff names them. #8947 changed emitted IR, my 58-gate run was green, and this assertion went red on `main` regardless. I have added "run `cargo test -p <crate> --tests` when a PR changes emitted IR or a runtime symbol" to how I audit, and no apology needed — the miss was as much mine as yours.

Issue #8969, opened 2026-08-28
Private class field: `this.#n += 1` and `this.#n++` produce NaN (compound/update lowering drops the read)


proggeramlug:
Compound assignment and the update operators on a **private class field** lose the read, so the field ends up holding `NaN`. The explicit form works, which localises this to the compound/update lowering rather than private-member storage itself.

Found while profiling the generic property read path (unrelated perf work); reproduced on pristine `main` @ `5d7f4e8e5`, 16-core Linux host.

## Repro

```ts
class A {
  #n = 0;
  compound()   { this.#n += 1;          return this.#n; }
  explicit()   { this.#n = this.#n + 1; return this.#n; }
  readInExpr() { const v = this.#n;     return v + 1; }
  incr()       { this.#n++;             return this.#n; }
}
const a = new A();
console.log("explicit:",   a.explicit());
console.log("readInExpr:", a.readInExpr());
console.log("compound:",   a.compound());
console.log("incr:",       a.incr());
```

| | node v26 | perry |
|---|---|---|
| `explicit` | 1 | 1 |
| `readInExpr` | 2 | 2 |
| `compound` | 2 | **NaN** |
| `incr` | 3 | **NaN** |

Reading a private field works. Writing a private field works. Only the desugared read *inside* `+=` / `++` comes back `undefined` (hence `undefined + 1 === NaN`), and the `NaN` is then stored — so the corruption persists into later reads.

This makes the ordinary counter/accumulator idiom on private state silently wrong rather than failing loudly:

```ts
class Counter { #count = 0; inc(by = 1) { this.#count += by; return this.#count; } }
new Counter().inc()   // node: 1, perry: NaN
```

## Two neighbouring divergences from the same subsystem

Filed here as context because they look like the same storage model; happy to split them out if you'd rather track them separately.

**1. Private fields are enumerable own properties.** They should be entirely invisible to reflection.

```ts
class B { #n = 5; get() { return this.#n; } }
Object.getOwnPropertyNames(new B())   // node: []            perry: ["#n"]
Object.keys(new B()).length           // node: 0             perry: 1
```

This also leaks into `JSON.stringify`, spread, and `for...in`.

**2. An ordinary property whose name matches the internal storage prefix is silently dropped.** No error, no value — the write is simply lost:

```ts
const o: Record<string, number> = {};
o["#<perry:private-member:1:x>"] = 8;
JSON.stringify(o)          // node: {"#<perry:private-member:1:x>":8}   perry: {}
[...]

PR #8970, opened 2026-08-28
perf(runtime): reject non-private property names without validating UTF-8 (read −10.5%)

proggeramlug:
`private_member_storage_name` runs at the top of **both** the generic property read (`js_object_get_field_by_name`) and the generic write (`field_set_by_name`), so every property operation in the program pays it. It reached its verdict via `str_from_string_header`, which UTF-8-validates the *entire* key before the prefix compare can reject it.

In a computed-key read loop that is **~7.5% self time** (plus its share of `from_utf8`) — all of it spent proving that `"k123"` does not begin with `#`.

A private-member storage name always starts with `#` and is longer than the `#<perry:private-member:` prefix, so a length compare plus one byte settles it for every ordinary key, with no validation and no slice. Keys that pass the filter still take the original path, so private members and `#`-prefixed ordinary keys are untouched.

### Measurement

Interleaved A/B pairs, min-of-15, quiet load (~1.9), 16-core Linux host:

| loop | main | this PR | |
|---|---|---|---|
| read only | 38 ms | **34 ms** | −10.5% |
| combined overwrite | 82 ms | **76 ms** | −7.3% |
| write only | 44 ms | 43 ms | −2.3% |

Means move the same direction on all three (45→43, 93→85, 47→46). An earlier 6-pair run was inconclusive — the spread exceeded the effect — so this is the tightened measurement.

### Correctness

- Differential against pre-change perry on a private-member exercise — instance fields, `static #instances`, private methods, private getters, `#x in obj`, subclassing, and an ordinary property literally named `#<perry:private-member:1:x>` — is **byte-identical**, including on the cases where perry currently diverges from node.
- The guard performs no dereference the previous path didn't already perform (null check, then `byte_len`, then the payload's first byte via the shared `string_header_payload` helper).

### Note

Building that differential surfaced three pre-existing correctness bugs in private class members on `main` (compound assignment on a private field yields `NaN`; private fields are enumerable own properties; an ordinary key matching the internal prefix is silently dropped). They are **not** affected by this PR — same output before and after — and are filed separately as #8969.

## Summary by CodeRabbit

* **Performance**
  * Improved property access performance by quickly filtering keys that cannot reference private members.
  * Reduced overhead in read and overwrite operations while preserving existing behavior.
* **Bug Fixes**
  * Maintained correct handling for private members and specially formatted keys.
* **Validation**
  * Confirmed output remains identical to the previous implementation.

### DIFF
--- a/changelog.d/8970-private-member-name-fast-reject.md
+++ b/changelog.d/8970-private-member-name-fast-reject.md
@@ -1,0 +1,22 @@
+The generic property read and write no longer UTF-8-validate every key to
+discover it isn't a private class member.
+
+`private_member_storage_name` sits at the top of both `js_object_get_field_by_name`
+and `field_set_by_name`, so every property operation in the program pays it —
+and it reached its verdict through `str_from_string_header`, which validates
+the WHOLE key as UTF-8 before the prefix compare can reject it. In a
+computed-key read loop that measured ~7.5% self time (plus its share of
+`from_utf8`), all of it spent proving that `"k123"` does not begin with `#`.
+
+A storage name always starts with `#` and is longer than the
+`#<perry:private-member:` prefix, so a length compare and one byte settle it
+for every ordinary key. Keys that pass the filter still take the original
+path, so private members — and the rare `#`-prefixed ordinary key — behave
+exactly as before.
+
+Interleaved A/B, min-of-15 at quiet load: read loop 38 → 34 ms (−10.5%),
+combined overwrite loop 82 → 76 ms (−7.3%), write loop 44 → 43 ms; means move
+the same direction on all three. Output on a private-member differential
+(fields, static private counters, private methods, private getters, `#x in o`,
+subclassing, and an ordinary key named `#<perry:private-member:1:x>`) is
+byte-identical to pre-change perry.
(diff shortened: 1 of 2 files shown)


proggeramlug:
Merged. The reasoning holds and the implementation is careful in the two places it had to be:

- The length guard runs **before** the byte read (`byte_len <= PRIVATE_MEMBER_PREFIX.len()` returns early), so the payload dereference only happens when at least one byte exists — an empty or short key never reads out of bounds.
- Comparing one byte against `b'#'` needs no UTF-8 validation: any sequence whose first byte differs is correctly rejected regardless of what follows, and anything starting with `#` still goes through the validated path, so the slow path's behaviour is unchanged.
- Using `<=` rather than `<` is right too — a storage name is strictly longer than the prefix, so a key exactly the prefix's length cannot be one.

Nice to see it reuse `crate::object::string_header_payload` rather than open-coding the offset; that keeps the string-payload-access ratchet at baseline, which is where a couple of recent PRs tripped.

**Validation** — runtime 2779/0, codegen 1341/0 (`RUST_TEST_THREADS=1`); `scripts/run_lint_gates.sh` 57 of 58 including the compile tier (`-D warnings`, `clippy`) — the exception is the pre-existing `\${{ }}` substitution artifact (#8929).

The −10.5% is not re-measured here.

PR #8971, opened 2026-08-28
perf(runtime): the last four per-element key scans resolve via the shape index (read loop −10.5%)

proggeramlug:
#8936 and #8950 replaced the `js_array_get` + `js_string_key_matches` walks on the `[[Set]]`, `delete` and `[[Get]]`-fallback paths. Four copies survived, all of them in *miss* fallbacks — which is exactly where they are most expensive, because a miss is also when the caches above them are cold.

The one that showed up in a profile: **`js_object_get_field_by_name`'s FAST LANE**. Its read-plan cache is epoch-guarded and flushed on every GC and on descriptor / prototype / delete mutations, and its miss path was a full walk of up to 4096 keys. On a 500-key receiver that put `js_string_key_matches` at **9.6% self time** in a computed-key read loop — second only to the entry function itself.

The other three: the write fast path's read-plan miss fallback, the write tail, and the read tail. Two of them walked via `js_array_get`, which additionally probes each index for a per-index accessor.

All four now call `keys_find_slot_by_key_ptr` — shape hash index first, raw dense-slot scan as its own fallback and correctness backstop. The two tail sites **keep their original `js_string_key_matches` test as the gate**, so the resolver can only narrow the candidate slot, never widen what is accepted.

### Measurement

Interleaved A/B pairs, min-of-15, quiet load (~2.0), 16-core Linux host. Node on the same host in brackets:

| loop | main | this PR | |
|---|---|---|---|
| read only | 38 ms (node 25) | **34 ms** | −10.5% min, −18.6% mean |
| combined overwrite | 79 ms (node 29) | **77 ms** | −2.5% min, −13.2% mean |
| write only | 44 ms (node 23) | 43 ms | −2.3% |

The read-only loop is where this lands, and the combined loop shows why: there the write primes the read plan, so the read never reaches the miss path. Measuring the read lane alone first (before the other three sites) gave read 45 → 35 min / 75 → 53 mean, with the write loop bit-for-bit unchanged — which is the attribution check that the effect is coming from where the profile said.

The three non-lane sites were then measured *separately* against the read-lane-only build, and are individually small: write 45 → 44 min, combined 80 → 76 min, read unchanged. They are included because they are the same transformation on the same code path family, not because they carry the headline number.

### Correctness

- Computed-key differential vs node — delete/re-add key ordering, `Map`/`Set` keys, the 5→6 byte SSO boundary, non-ASCII, floats, negatives, 1e21, numeric-looking keys: **byte-identical**.

## Summary by CodeRabbit

  * Improved JavaScript object property-key lookups with faster indexed resolution.
  * Reduced execution time for read-heavy and combined read/write workloads.
  * Preserved existing property access behavior and compatibility.

* **Testing**
  * Verified computed-key results match Node.js output.
  * All 2,779 tests passed.

### DIFF
--- a/changelog.d/8971-key-scan-kills-read-lane.md
+++ b/changelog.d/8971-key-scan-kills-read-lane.md
@@ -1,0 +1,36 @@
+The last four per-element key scans on the property paths now resolve through
+the shape hash index.
+
+#8936 and #8950 replaced the `js_array_get` + `js_string_key_matches` walks on
+the `[[Set]]`, `delete` and `[[Get]]`-fallback paths, but four copies survived
+in the plan-miss and cache-miss fallbacks:
+
+- `js_object_get_field_by_name`'s FAST LANE — the read-plan cache's miss path,
+  a full walk of up to 4096 keys. The plan is epoch-guarded and flushed on
+  every GC and on descriptor / prototype / delete mutations, so this ran in
+  full each time it was flushed. It put `js_string_key_matches` at **9.6% self
+  time** in a computed-key read loop, second only to the entry itself.
+- the write fast path's read-plan miss fallback, and
+- the write tail and the read tail, whose loops walked every key (two of them
+  via `js_array_get`, which additionally probes each index for a per-index
+  accessor).
+
+All four go through `keys_find_slot_by_key_ptr`: shape hash index first, raw
+dense-slot scan as its own fallback and correctness backstop. The two tail
+sites keep their original `js_string_key_matches` test as the gate, so the
+resolver can only narrow the candidate slot, never widen what is accepted.
+
+Interleaved A/B, min-of-15 at quiet load (node on the same host in brackets):
+
+| loop | main | this PR | |
+|---|---|---|---|
+| read only | 38 ms (25) | **34 ms** | −10.5% min, −18.6% mean |
+| combined overwrite | 79 ms (29) | **77 ms** | −2.5% min, −13.2% mean |
+| write only | 44 ms (23) | 43 ms | −2.3% |
+
+The read-only loop is where this lands: in the combined loop the write primes
+the read plan, so the read never reaches the miss path at all.
+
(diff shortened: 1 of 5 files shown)


proggeramlug:
Merged.

**One fix pushed:** two of the four replacements became `for i in <Option>.map(…)`, and `for_loops_over_fallibles` is a `-D warnings` error, so the `warnings` job was red on this branch. Converted both to `if let Some(i) = …`; neither body uses `continue`/`break`, so it is a pure substitution.

**On the substitution itself** — all four sites route through `keys_find_slot_by_key_ptr` and keep the original `js_string_key_matches` test as the gate, so the byte-compare semantics that decided the old linear walk still decide the outcome; the index only changes *which* slot gets tested.

Worth noting for the record that this makes `keys_find_slot_by_key_ptr` load-bearing on four more paths, including `js_object_get_field_by_name`'s fast lane. Its early guard is the one I tightened while landing #8936, from an open-coded `(key as usize) < 0x10000` to `addr_class::is_above_handle_band`. That is the stricter of the two (the band is `0x100000`), and it is correct rather than merely tighter: per CLAUDE.md a value below `0x100000` is a handle, not a heap pointer, so no real `StringHeader` can fall in the range the tightening now rejects.

**Validation** — runtime 2779/0, codegen 1341/0, stdlib 124/0 (`RUST_TEST_THREADS=1`); under `PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1` the failing count is 16, the pre-existing set; `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green after the fix — the exception is the pre-existing `\${{ }}` artifact (#8929).

**One limit:** I did not run the `crates/perry/tests` integration tier for this one. It is runtime-internal rather than an IR change, and the debug-profile build that tier needs costs ~23 GB, which I did not have. The −10.5% is also not re-measured.

PR #8972, opened 2026-08-28
fix(runtime): Array-subclass fill honours start and end

proggeramlug:
`js_array_subclass_init` installs `fill` as an own method on the instance (node inherits it from `Array.prototype`; perry has no such prototype object for these classes), and the stub was registered with **arity 1** — so `start` and `end` never reached `js_array_fill_generic`:

```js
class A extends Array {}
const b = new A(); b.push(1, 2, 3, 4);
b.fill(9, 1);      // node: [1,9,9,9]   perry: [9,9,9,9]
b.fill(9, 1, 3);   // node: [1,9,9,4]   perry: [9,9,9,9]
```

The stub now takes `(value, start, end)`; an omitted argument arrives as `undefined` and selects the spec default (`0` / `length`).

Verified against node on all four forms (value only, start, start+end, negative start) — `array_subclass_fill_args.rs`, and the compiled output matches node exactly.

Found while probing Array-subclass semantics for the elements-store work (#8966).

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed `Array.prototype.fill` on classes extending `Array` to respect optional `start` and `end` arguments.
  - Calls such as `fill(value, start)`, bounded ranges, and negative indices now produce the expected results while preserving default behavior.

- **Tests**
  - Added coverage for supported `fill` argument combinations on array subclasses.

- **Documentation**
  - Added a changelog entry describing the fix.


proggeramlug:
Merged. A clean bug and a clean fix — arity 1 meant `start`/`end` never reached `js_array_fill_generic`, so `b.fill(9, 1)` filled the whole array.

The part worth checking was the omitted-argument path, since an absent arg arrives as NaN-boxed `undefined` and passing it straight through as an `f64` would read as a number. `present()` tests `is_undefined()` and passes a 0/1 presence flag, so the generic helper takes its spec default instead. Correct.

The new test also covers the shapes that matter rather than just the reported one: `fill(9)`, `fill(9,1)`, `fill(9,1,3)` and `fill(9,-2)` — the negative relative index being the case most likely to go wrong once `start` actually arrives.

**One fix pushed:** the fragment was `changelog.d/0000-array-subclass-fill-args.md`; renamed to `8972-`. Filenames are PR-keyed so in-flight PRs cannot collide, and `0000` collides with every other placeholder — there are three such files already on `main`, which I am cleaning up separately.

**Validation** — runtime 2779/0 (`RUST_TEST_THREADS=1`), and I ran the new integration test on the debug profile the harness actually links (`cargo build -p perry-runtime-static -p perry-stdlib-static` first): **1 passed**. `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green — the exception is the pre-existing `\${{ }}` artifact (#8929).

PR #8974, opened 2026-08-28
runtime: make the Array-subclass elements store the default representation (−11.4% / −11.9% on the wolf-ecs twins)

proggeramlug:
Follow-up to #8966, which landed the elements store behind `PERRY_ARRAY_SUBCLASS_ELEMENTS=1`. This flips the default; `PERRY_ARRAY_SUBCLASS_ELEMENTS=0` restores the shape-carried form as a bisecting kill switch.

## Evidence

**Whole-corpus differential.** The gate is a pure runtime switch (codegen always emits the probes), so every fixture can be compiled once and run twice. Across `test-files/`: **1386 files, 1285 compiled, 9 output differences — none attributable to the store.** Those nine are `test_crypto` (random bytes/UUID), `test_date` (timestamps), `test_gap_console_methods` (`console.time` values), `test_parity_sys` (PID in a deprecation warning), `test_parcel_watcher_facade` (flaky watcher timeout — it *fails* with the store off and passes with it on), `test_perry_tui_inkcompat_useref` (times out either way), plus `test_issue63_escape`, `test_math`, `test_require`, each of which produces a different output hash on five consecutive runs **with the switch untouched**.

**Integration suites with the store enabled:** `issue_8655_array_subclass_indexing` 2/2, `issue_array_subclass_super_init` 1/1, `issue_4908_subclass_native_member_base` 2/2, `issue_8690_loop_versioned_arraylike` 3/3, `issue_8773_closure_capture_packed_loops` 4/4, `issue_5139_object_arraylike_method_dispatch` 3/3, `interface_typed_arraylike_method_dispatch` 3/3, `issue_8897_field_push_writeback` 3/3.

**Performance** (Mac mini, 11 alternating pairs, same binary, store off → on):

|---|---|---|
| 2 s | 0.4114 → **0.3644** (**−11.41%**, 11/11) | 0.3399 → **0.2999** (**−11.80%**, 11/11) |
| 50 ms | 0.4110 → **0.3640** (**−11.49%**, 11/11) | 0.3399 → **0.2995** (**−11.91%**, 11/11) |

Perry vs Node 26.5 on that host goes from 3.02× / 2.24× to **2.71× / 2.01×**.

## Semantics

Strictly closer to node. With the store on, `JSON.stringify(sub)` produces the **array** form, `Object.keys` no longer leaks `length`, and `sort` / `reverse` / `splice` / `shift` / `unshift` / `length` truncation / holes / spread / `Array.from` become node-identical — every one of those printed the object form (`{"0":5,"1":1,"length":2}`) before.

Unchanged divergences (identical with the store on or off, pre-existing, tracked in #8953): `fill` in `getOwnPropertyNames`/`for..in`, `A.from([…])` not populating, `f.constructor === A` after `map`, `entries()`/`keys()` unimplemented on a subclass instance.

## Tests

The shape-carried form stays reachable through the kill switch, so its unit tests now name it explicitly (`ArraySubclassRepresentationGuard::shape_carried()`, 13 tests) and the elements tests pin the other direction. `cargo test -p perry-runtime --lib` 2779/0.

## Gates

`RUSTFLAGS=-D warnings cargo check --workspace --all-targets` clean; file size; GC store-site inventory; raw-handle debt unchanged (967, none raised); shape census; local-binding audit; addr-class audit. (`temp_root_operand_temporaries::string_literal_concat_operand_is_re_derived_below_the_allocating_sibling` fails on plain `origin/main` too — unrelated to this change.)

## Summary by CodeRabbit

* **New Features**
  * Array subclasses now store indexed elements and `length` using the standard elements representation by default.
  * Array operations such as `push`, `pop`, indexing, serialization, and key enumeration now more closely match Node.js behavior.
* **Performance**
  * Improved benchmark results for array-subclass additions, removals, and entity cycles.
* **Bug Fixes**
  * Prevented `length` from appearing unexpectedly in object keys.
* **Configuration**
  * Added an environment variable option to restore the previous representation for troubleshooting.


proggeramlug:
Merged. This closes the finding I left on #8966, and I verified it directly rather than inferring it from the diff.

When I audited #8966 I reported that with `PERRY_ARRAY_SUBCLASS_ELEMENTS=1` the runtime suite was **2766 passed / 13 failed** — all 13 in `array::subclass_tests::*`, the old shape-carried representation's suite. That is now fixed. All three states are clean:

| state | result |
|---|---|
| default (now ON) | **2779 / 0** |
| `PERRY_ARRAY_SUBCLASS_ELEMENTS=0` (kill switch) | **2779 / 0** |
| `PERRY_ARRAY_SUBCLASS_ELEMENTS=1` (explicit) | **2779 / 0** |

That is what the kill-policy actually asks for — both sides of the knob exercised, so the mode is a decision that has been made rather than one nobody has run end to end. The kill switch being green matters as much as the new default: it is only a usable bisection tool if the old path still passes.

The whole-corpus differential is the right kind of evidence too — compiling each fixture once and running it twice under a pure runtime switch removes compilation as a variable, and attributing all nine differences to crypto randomness, timestamps and console formatting is the sort of accounting that makes a null result believable.

**One fix pushed:** the fragment was named with the `0000` placeholder; renamed to `8974-`. I had just cleaned three of those off `main` in #8973, so this would have re-introduced the collision.

**Validation** — the three states above; under `PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1` the failing count is 16, the pre-existing set; the GC env-knob gate passes; `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green — the exception is the pre-existing Actions-expression artifact (#8929).

PR #8975, opened 2026-08-28
perf(runtime): one shape-descriptor lookup per dynamic write IC hit, not two

proggeramlug:
`object_shape_id` runs a full shape-table lookup — and copies the whole `ShapeDescriptor` out of the table — purely to prove the stamped id is live, then throws the descriptor away. `dyn_ic_try_store` called it for the token compare and then looked the **same id up again** for the slot bound; both write re-prime sites did the same while already holding a descriptor.

It now reads the stamp straight off the header. In `dyn_ic_try_store` the token compare runs first, so a wrong-shape receiver costs a load and a compare with no table work at all, and the one lookup that supplies the slot bound doubles as the liveness proof: a stamp with no live descriptor returns `None` exactly where `object_shape_id`'s `0` made the token compare fail before.

### Measurement — and what it does not show

`shape_descriptor_by_id` was the largest single item in a computed-key write profile (10.5% at the time it was found). The **mechanism is verified directly**: in that loop it falls **7.01% → 3.49%** of self time, and `object_shape_id` (1.32%) leaves the profile entirely.

The wall clock is a smaller story, and I would rather state it than round it up. Interleaved A/B, min-of-21, quiet load (~1.15), built from the exact parent commit and this commit in one run:

| loop | base | this PR | |
|---|---|---|---|
| write only | 42 ms | **40 ms** | −4.8% min, −2% mean |
| combined overwrite | 71 ms | 72 ms | unchanged |
| read only | 31 ms | 31 ms | unchanged (write-only change) |

So: a small, mechanically-confirmed reduction on the loop it targets, and neutral elsewhere. It is worth having because it is strictly less work for identical semantics, not because it moves a headline number.

An earlier comparison of mine suggested −7% to −12%; that was against binaries built before #8970 and #8971 merged, so it was crediting this change with their gains. The table above is against the correct parent.

### Correctness

- Computed-key differential vs node — delete/re-add ordering, `Map`/`Set` keys, the SSO boundary, non-ASCII, floats, negatives, 1e21: **byte-identical**.

## Summary by CodeRabbit

  * Improved dynamic property writes by reducing shape lookups during inline-cache hits.
  * Wrong-shape writes now avoid unnecessary table work.
  * Write-only workloads show a measurable performance improvement, while overwrite and read performance remain unchanged.
* **Reliability**
  * Preserved computed-key output with byte-identical results.

### DIFF
--- a/changelog.d/8975-single-shape-lookup-per-ic-hit.md
+++ b/changelog.d/8975-single-shape-lookup-per-ic-hit.md
@@ -1,0 +1,24 @@
+The dynamic write IC does one shape-descriptor lookup per hit instead of two.
+
+`object_shape_id` runs a full shape-table lookup — and copies the whole
+`ShapeDescriptor` out of the table — purely to prove the stamped id is live,
+then discards it. `dyn_ic_try_store` called it for the token compare and then
+looked the SAME id up again for the slot bound; both write re-prime sites did
+the same while already holding a descriptor.
+
+It now reads the stamp off the header. In `dyn_ic_try_store` the token compare
+runs first, so a wrong-shape receiver costs a load and a compare with no table
+work at all, and the single lookup that supplies the slot bound doubles as the
+liveness proof: a stamp with no live descriptor returns `None` exactly where
+`object_shape_id`'s 0 made the token compare fail before.
+
+Verified by profile rather than by wall clock, because the effect is close to
+the noise floor of the benchmark. In a computed-key write loop
+`shape_descriptor_by_id` falls **7.01% → 3.49%** of self time and
+`object_shape_id` (1.32%) leaves the profile entirely.
+
+Wall clock, interleaved A/B min-of-21 at quiet load, against the exact parent
+commit: write-only loop 42 → 40 ms min (−4.8%), mean 46 → 45. The combined
+overwrite loop and the read loop are unchanged, as expected for a change
+confined to the write IC. Computed-key differential output is byte-identical
+to node; suite 2779 passed, 0 failed.
--- a/crates/perry-runtime/src/proxy/put_value.rs
+++ b/crates/perry-runtime/src/proxy/put_value.rs
@@ -451,8 +451,11 @@ pub extern "C" fn js_put_value_set_ic_miss(
             return result;
         }
 
+        // The descriptor above already proves this stamp is live, so the
+        // token comes from the header word rather than from a second full
+        // lookup-and-copy of the same id (see `dyn_ic_try_store`).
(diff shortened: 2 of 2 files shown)


proggeramlug:
Merged.

**On the substitution** — the thing that had to hold is that a *stale* stamp still cannot drive a store, since `object_shape_id`'s table lookup was what proved liveness. It does: a stale stamp can now pass the token compare, but `shape_descriptor_by_id(stamp)?` then returns `None` and `dyn_ic_try_store` bails. The rejection just moves one step later than `object_shape_id`'s zero did — same outcome, one lookup instead of two.

Both priming sites check out too: each has a `shape` descriptor for the receiver already in scope above (supplying `live_inline_slot_count` / `logical_key_count`), so "the descriptor above already proves this stamp is live" is literally true rather than an assumption. And even if a token were primed from a stale stamp, the consumer rejects it at the descriptor lookup — the two halves fail safe independently.

**Validation** — runtime 2779/0, codegen 1341/0 (`RUST_TEST_THREADS=1`); under `PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1` the failing count is 16, the pre-existing set; `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green (the exception is the pre-existing Actions-expression artifact, #8929).

I also ran the `crates/perry` integration tier this time rather than caveat it. One suite fails — `blocklist_addsubnet_prefix::block_list_add_subnet_respects_numeric_prefix` — and it is **not yours**: it fails identically on pristine `main`. Filing that separately.

proggeramlug:
Correction to my note above, so it does not mislead anyone reading this later.

I said the `blocklist_addsubnet_prefix` failure "fails identically on pristine `main`", implying a main-side red. That was **my build scope, not a defect** — and I have not filed an issue.

The test needs `libperry_ext_net.a`, and I had built only `-p perry -p perry-runtime-static -p perry-stdlib-static`. Perry's own warning said so plainly and I skimmed past it:

> warning: `net` needs libperry_ext_net.a, which is not on disk … build the wrapper in the SAME cargo invocation as the stdlib archive

With the coherent set — `cargo build -p perry -p perry-runtime-static -p perry-stdlib-static -p perry-ext-net` — it is **1 passed / 0 failed** on pristine `main`.

My A/B was doubly unreliable: the second arm reused the first arm's `.a` after switching commits, so it reported the stale-archive error rather than a result. Two artifacts stacked into something that looked like a real pre-existing failure.

Nothing changes for this PR — runtime 2779/0, codegen 1341/0, gates 57 of 58, and the integration tier is clean once built correctly.

PR #8977, opened 2026-08-28
perf(runtime): 2-way set-associative write stub — computed-key write loop now beats node (−50%)

proggeramlug:
A direct-mapped table cannot hold a colliding pair **at all**. The two keys evict each other on every rotation through the key set, so both miss forever and neither ever stabilises — the miss is permanent, not probabilistic.

On the 500-key computed-key write loop that left ~87k of 600k writes falling through every cache into the full `[[Set]]` walk. The call graph puts that at **11.9% of the program**, essentially all of it (10.7%) inside `js_put_value_set`.

### Measurement

Interleaved A/B, min-of-21, quiet load, built from the exact parent commit and this commit in one run. Node on the same host in brackets:

| loop | base | this PR | |
|---|---|---|---|
| write only | 40 ms (node 24) | **20 ms** | **−50%** — and now *faster than node* |
| combined overwrite | 70 ms (node 28) | **48 ms** | −31% |
| read only | 30 ms (node 22) | 31 ms | unchanged (write-only change) |

Means move identically (write 44 → 22, combined 73 → 48).

### How it was found

Not in a profile — the profile only showed the *consequence* (`js_put_value_set` high). Counters split by miss cause showed probes missing with the **right shape token and the wrong key** in the way, which is the signature of eviction rather than of a cold or badly-hashed cache. Dumping the offending pairs named them outright: `"k10"`/`"k115"` and `"k11"`/`"k125"`, landing on indices 3885 and 2220.

Worth recording because I got this wrong twice on the way: I first suspected hash quality (a sweep says the hash is fine — 464/500 distinct ways), and separately suspected the inline IC's `LocalGet`/`This` receiver gate (module-scope and function-scope receivers measure identically). Associativity was the actual constraint.

### The change

Two ways per bucket at the **same total capacity** (2048 × 2 rather than 4096 × 1), so the table does not grow. A sweep of the working set puts the worst bucket at exactly two, so two ways absorb the collisions that exist rather than merely reducing them. Insert refreshes a resident key, then fills an empty way, and only otherwise evicts — shifting way 0 down so the most recently primed key survives.

### Correctness

- Computed-key differential vs node — delete/re-add ordering, `Map`/`Set` keys, the SSO boundary, non-ASCII, floats, negatives, 1e21: **byte-identical**.

The safety argument is unchanged from #8965: entries hold only content-derived key bits and never an address, and `dyn_ic_try_store` revalidates the receiver's shape token, flags and slot bound on every hit.

## Summary by CodeRabbit

- **Performance**
  - Improved performance for computed-key write operations, reducing execution time in representative workloads.
  - Combined overwrite operations are also significantly faster, with improvements of up to 31%.

- **Reliability**
  - Improved handling of frequently changing object properties, reducing slow fallback operations during intensive write workloads.


proggeramlug:
Merged. The framing is the useful part: a direct-mapped table cannot hold a colliding pair *at all*, so those two keys evict each other on every rotation and the miss is permanent rather than probabilistic. That is a much stronger reason to go set-associative than "hit rate is a bit low".

**On correctness** — capacity is unchanged (4096 → 2048 × 2), the hit condition is identical (`tok == token && kb == key_bits && tok != 0`), just tested in two places, and insert updates in place before filling an empty way. Eviction is always safe: it can only cost a later miss, never a wrong answer. The `tok != 0` empty sentinel still holds because every token carries `PIC_ID_TOKEN_BIT`, so a live entry can never read as empty.

And a stale way remains harmless downstream for the reason #8975 established: the token compare is followed by `shape_descriptor_by_id(stamp)?`, so a hit on a dead shape falls through instead of storing.

**One fix pushed:** the fragment was named `8976-write-stub-two-way.md`. #8976 is a different PR — the Array-subclass loop-guard fix that merged an hour ago — so this misattributed the change to it. Renamed to `8977-`.

**Validation** — runtime 2779/0, codegen 1341/0 (`RUST_TEST_THREADS=1`); under `PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1` the failing count is 16, the pre-existing set; `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green — the exception is the pre-existing Actions-expression artifact (#8929).

The −50% and the node comparison are not re-measured here.

Issue #8978, opened 2026-08-28
Changeset gate accepts 0000- and wrong-PR changelog fragment names


proggeramlug:
## The gap

The changeset gate in `lint` accepts any fragment matching:

```
^changelog\.d/[0-9]+-[^/]+\.md$
```

That requires *some* digits, not **this PR's** number. So both of these pass:

- `changelog.d/0000-<slug>.md` — the placeholder, when an author has not filled in the number yet
- `changelog.d/<other-PR>-<slug>.md` — a number belonging to a different PR

`changelog.d/README.md` says filenames are PR-keyed precisely so in-flight PRs cannot collide, and the failure is silent: fragments are only folded into release notes at tag time by `scripts/cut_release_notes.sh`, so a wrong number is invisible until a release is cut and then attributes the change to the wrong PR.

## Evidence it happens

Four in one day (2026-08-28), all caught by eye during review rather than by any gate:

| PR | shipped as | should have been |
|---|---|---|
| #8944 | `0000-inline-array-pop-tier.md` | `8944-` |
| #8947 | `0000-inline-method-shape-probe.md` | `8947-` |
| #8291 | `0000-node-version-26.5.1.md` | `8291-` |
| #8977 | `8976-write-stub-two-way.md` | `8977-` |

The first three sat on `main` until #8973 renamed them; the fourth misattributed a write-stub change to #8976, an unrelated Array-subclass loop-guard fix that had merged an hour earlier.

## Why the obvious fix is wrong

"The fragment must start with this PR's number" would be a false positive on legitimate cases:

- **Backfills/cleanups.** #8973 renamed three fragments to `8944-`, `8947-` and `8291-` — none of them its own number. A strict rule blocks exactly the PR that repairs the problem.
- **Stacked PRs** where a fragment legitimately names the parent change.

So the check needs to be "the number is this PR's, **or** an existing PR number", or to warn rather than fail. The `0000` case specifically is unambiguous though: no PR is ever number 0, so rejecting `^changelog\.d/0+-` is a safe tightening on its own and would have caught three of the four above.

## Suggested minimum

Reject `0000-` (and any all-zero prefix) outright; treat a number that is neither this PR's nor an existing PR as a warning. That closes the common case without breaking cleanups.

Found while auditing the PR queue; not fixed here because a naive tightening would have blocked #8973.

proggeramlug:
More evidence, from a single batch of three PRs merged tonight — all three were wrong, in three different ways:

[...]

PR #8980, opened 2026-08-28
perf(runtime): hoist the private-member guard to its call sites (pure read −15%)

proggeramlug:
Follow-up to #8970, which fixed half of this. That PR made the private-member name test cheap but left the **call** on every property operation.

Profiling a *pure* property-read loop — `o[k]` with pre-built keys, so no concat, no interning of fresh keys, no allocation churn — put `private_member_get_by_name` at 11.4% and `private_member_storage_name` at 5.4%: **16.8% of the loop and the largest single item**, essentially all of it call overhead for keys that are rejected on a length compare before doing any work.

The guard is now invoked at the three call sites — the generic read entry, the class-field read miss, and the generic write — so an ordinary property operation makes no call into the private-member path at all. Keys that pass the guard take exactly the original path.

### Measurement

Interleaved A/B, min-of-21, built from the exact parent commit and this commit in one run. The box had heavy co-tenant load during this window (~14.9), so the absolutes are inflated; interleaving cancels that and the ratios are the signal:

| loop | base | this PR | |
|---|---|---|---|
| pure property read | 26 ms | **22 ms** | −15% |
| computed-key read | 51 ms | **45 ms** | −12% |
| combined overwrite | 50 ms | **46 ms** | −8% |
| write only | 23 ms | 23 ms | unchanged |

Means move the same way (read 41 → 38, computed read 57 → 51, combined 51 → 49).

### Correctness

This change decides whether the private-member path is entered at all, so a bug here would silently disable private class members rather than merely slow them down. Checked directly:

- Private-member exercise — instance fields, `static #instances`, private methods, private getters, `#x in obj`, subclassing, and an ordinary property literally named `#<perry:private-member:1:x>` — **byte-identical output before and after**.
- Computed-key differential vs node: **byte-identical**.

### Context for the remaining read gap

With this, the same isolated read loop ranks `shape_descriptor_by_id` at 10.1% (the read path has the redundant double-lookup that #8975 fixes on the write side), `js_typed_feedback_object_get_field_by_name_f64` at 10.0%, and `intern_dispatch_bytes` at 6.9%. Worth noting separately: perry's SSO key construction is now *faster* than node's (~10ns vs ~32ns per `"k" + i`); the residual read gap is the lookup itself.

## Summary by CodeRabbit

* **Bug Fixes**
  * `perry run <directory>` now correctly resolves the project entry from that directory, including `perry.toml`, `src/main.ts`, or `main.ts`.
  * Fixed directory-based runs on Windows.
  * Improved property access and updates by avoiding unnecessary private-member handling for ordinary names.

* **Performance**
  * Improved performance for common property reads and computed-key access while preserving private-member behavior.


proggeramlug:
Merged. Hoisting a guard to its call sites is only safe if every caller applies it, so that is what I checked rather than the diff alone: `private_member_get_by_name` has four caller files on `main` (`field_get_set.rs`, `get_field_by_name.rs`, `ic_miss.rs`, `private_member_access.rs`) and this PR touches all four, plus `field_set_by_name.rs` for the `private_member_storage_name` path. No caller is left calling through unguarded.

Together with #8970 this is the complete version of that fix — that one made the *test* cheap, this one removes the *call*, and the two only pay off together on a pure read loop where the call overhead was the residue.

**One fix pushed:** the fragment was named `8978-private-guard-call-site.md`. #8978 is not a PR at all — it is the issue I filed earlier today about exactly this class of mistake, so the fragment would have attributed a runtime change to a CI-hygiene issue. Renamed to `8980-`.

That is the fifth fragment-naming fix today, which I think strengthens the case in #8978: the gate checks the filename *shape* (`^changelog\.d/[0-9]+-[^/]+\.md$`) and any number satisfies it, so nothing catches a plausible-but-wrong one.

**Validation** — runtime 2779/0, codegen 1341/0 (`RUST_TEST_THREADS=1`); under `PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1` the failing count is 16, the pre-existing set; `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green — the exception is the pre-existing Actions-expression artifact (#8929).

The 16.8% figure is not re-measured here.

PR #8981, opened 2026-08-28
fix(runtime): make dynamic Function refusal catchable

### LINKED ISSUE #8958: Refused dynamic Function segfaults instead of throwing — breaks zod v4's eval probe on every AOT binary
## Summary

A refused dynamic `Function` construction **segfaults the process** instead of
throwing a catchable error, even when the call site is inside `try`/`catch`.

Perry `0.5.1519` from source (Linux x86_64, Ubuntu 24.04.4).

## Repro

```ts
// probe.ts — the shape zod v4 uses to detect whether eval is available
try {
  const F = Function;
  new F("");
  console.log("eval allowed");
} catch {
  console.log("eval refused — caught");   // never reached
}
console.log("still alive");                // never reached
```

```
$ perry compile probe.ts -o probe && ./probe
[perry] dynamic Function refused (AOT, dyn-eval feature off) — 1 arg(s); body[..160]=""
Segmentation fault (core dumped)          # exit 139
```

Expected: the `catch` runs and the program prints `eval refused — caught`
followed by `still alive`. Refusing dynamic code under AOT is correct and
desirable — the diagnostic line is right there — but it should surface as a
throw, not a crash.

## Why it matters in practice

This is not a synthetic case. **zod v4** — probably the most widely used
validation library in the TS ecosystem — probes exactly this way, in
`v4/core/util.js`:

```js
export const allowsEval = /* @__PURE__*/ cached(() => {
  if (globalConfig.jitless) return false;
  …
  try {
    const F = Function;
    new F("");
    return true;
  } catch (_) {
    return false;      // ← the intended AOT path
  }
});
```

The library is *already* written to degrade gracefully on a runtime that
refuses eval (it is the documented strict-CSP path). Because the refusal
crashes instead of throwing, any AOT binary that parses a zod schema dies at
startup rather than silently taking the interpreted path.

Verified end to end on a real service: an API binary using zod for request
validation links fine and then segfaults on boot with the message above. Two
minimal binaries built from the same toolchain:

| program | result |
|---|---|
| `z.object({…}).parse({…})` | `dynamic Function refused` → **SIGSEGV, exit 139** |
| same, after `z.config({ jitless: true })` | parses correctly, exit 0 |

So the workaround exists and is a one-liner — but it has to be discovered
from a segfault, and a user who does not know zod's internals has nothing to
go on. Throwing would let the library's own fallback do its job with no user
action at all.

## Suggested fix

Make the refusal path raise a catchable JS exception (a `TypeError` or an
`EvalError` mentioning `dyn-eval`), so `try`/`catch` behaves as the language
specifies and feature-probing libraries work unmodified.

proggeramlug:
## Summary

- mark `js_function_ctor_from_strings` as an unwind-capable generated-code boundary because its `dyn-eval`-off path originates a Perry exception
- keep refusal diagnostics while surfacing a catchable `TypeError`, so zod v4 and similar eval probes can select their fallback
- add a focused TypeError regression test and an unconditional ABI assertion

Fixes #8958.

## Testing

- `cargo test -p perry-runtime --lib dynamic_function_refusal_is_a_catchable_type_error -- --nocapture`
- `cargo test -p perry --test function_apply_dynamic_args_eval_surface function_apply_with_runtime_args_defers_to_a_located_aot_error -- --nocapture`
- exact issue probe against stripped debug/static runtime: before fix aborts with `panic in a function that cannot unwind`; after fix catches and prints `still alive`
- exact issue probe against stripped release `panic=abort` runtime: passes, plus 500/500 repeated runs

No version bump.

## Summary by CodeRabbit

* **Bug Fixes**
  * Runtime-string `Function` calls in environments with dynamic evaluation disabled now throw a catchable `TypeError` instead of aborting.
  * Libraries that probe dynamic evaluation can now safely select their fallback behavior.


proggeramlug:
Merged. This is the unwind regime that has bitten this repo repeatedly (#8479 and the PRs it burned), and the handling here is right: a plain `extern "C"` installs the abort-on-unwind guard, so a Perry exception originating inside the FFI boundary aborts the process instead of reaching the caller's landing pad — which is exactly how zod's harmless `new Function("")` capability probe became a process abort.

The part I especially like is `const _: extern "C-unwind" fn(*const f64, usize) -> f64 = js_function_ctor_from_strings;`. That is a compile-time ABI assertion that survives stripped builds, so a future revert to plain `extern "C"` fails to compile rather than silently reintroducing the abort. Given the history here, pinning the ABI rather than trusting the annotation to stay put is the difference between a fix and a fix that stays fixed.

**One fix pushed:** the fragment was named `8958-dynamic-function-refusal-unwind.md`. #8958 is the *issue* this fixes, not this PR, so the change would have been filed under an issue number at release time. Renamed to `8981-`.

**Validation** (batched with #8982 and #8983) — runtime 2780/0, hir 355/0, codegen 1341/0, stdlib 124/0 (`RUST_TEST_THREADS=1`); `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green — the exception is the pre-existing Actions-expression artifact (#8929).

Issue #8962, opened 2026-08-28
new Hono() throws 'Cannot initialize private elements twice on the same object' at runtime


proggeramlug:
## Summary

Constructing a `Hono` app throws at runtime:

```
TypeError: Cannot initialize private elements twice on the same object
```

Perry `0.5.1519` from source, Linux x86_64 (Ubuntu 24.04.4). hono `4.13.4`.

## Repro

`repro.ts`:

```ts
import { Hono } from "hono";
const app = new Hono();
app.get("/", (c) => c.text("ok"));
console.log("constructed and routed ok");
```

`package.json` (hono must be AOT-compiled for this to bite):

```json
{
  "dependencies": { "hono": "4.13.4" },
  "perry": {
    "compilePackages": ["hono"],
    "allow": { "compilePackages": ["hono"] }
  }
}
```

```
$ perry compile repro.ts -o repro       # compiles and links fine
$ ./repro
TypeError: Cannot initialize private elements twice on the same object
    at <anonymous>
```

Built with `--debug-symbols`, the frame resolves to the `new Hono()` line in
the user's own source — construction, not a later call.

## What is distinctive about the class

`hono/dist/hono-base.js` declares private **fields** and private **methods**
on the same class, and `hono/dist/hono.js` extends it with an *anonymous class
expression* that calls `super(options)`:

```js
// hono-base.js
class HonoBase {
  #path = "/";
  #notFoundHandler = notFoundHandler;
  constructor(options = {}) { … }
  #clone() { … }
  #addRoute(method, path, handler, baseRoutePath) { … }
  #dispatch(request, executionCtx, env, method) { … }
}

// hono.js
var Hono = class extends HonoBase {
  constructor(options = {}) {
    super(options);
    …
  }
};
```

I tried to reduce this to a standalone file and **could not** — a hand-written
base with private fields + private methods + an arrow-function field,
extended by an anonymous class expression calling `super()`, compiles and runs
correctly. So something more specific in hono's emitted shape triggers it; the
four-line import above is the smallest reproduction I have.

Things that do NOT reproduce it (all run correctly):

- private fields, including one initialised to an arrow closing over `this`
- a derived class with its own private fields calling `super()`
- a constructor that returns a different object
- private methods alongside private fields, plus `new this.constructor()`

## Why it is worth prioritising

hono is one of the most widely used TypeScript server frameworks, and it is
the natural thing to reach for when writing a Perry HTTP service. Any binary
[...]

PR #8982, opened 2026-08-28
fix(hir,fetch): preserve Hono response state

### LINKED ISSUE #8968: hono: unmatched routes return 200 with empty body; custom app.notFound() handler never runs
## Summary

A hono route that does not match returns **200 with an empty body** instead of
running the not-found path. A **custom `app.notFound()` handler is never
invoked at all**.

Perry `0.5.1519` from source (Linux x86_64). hono `4.13.4`.

This is a silent wrong-answer bug, not a crash: the binary starts, matched
routes work, and every unmatched request quietly succeeds with nothing in it.

## Repro

In a directory whose only dependency is `hono@4.13.4` (no `perry` key needed):

```ts
import { Hono } from "hono";
const app = new Hono();
app.get("/", (c) => c.text("root"));
app.notFound((c) => c.json({ code: "NOT_FOUND" }, 404));

async function main() {
  const miss = await app.fetch(new Request("http://x/missing"));
  console.log("custom notFound ->", miss.status, JSON.stringify(await miss.text()));
  const hit = await app.fetch(new Request("http://x/"));
  console.log("matched route   ->", hit.status, JSON.stringify(await hit.text()));
}
main();
```

| runtime | output |
|---|---|
| **node 26** | `custom notFound -> 404 "{\"code\":\"NOT_FOUND\"}"` / `matched route -> 200 "root"` |
| **perry** | `custom notFound -> 200 ""` / `matched route -> 200 "root"` |

Matched routing is correct; only the miss path diverges, and the registered
handler does not run.

hono's **default** not-found behaviour is wrong in the same way — `404 Not
Found` under node, `200 ""` under perry.

## It is in the native private-element path

Same source, one variable — bundling with `esbuild --target=es2019`, which
lowers `#` privates to WeakMaps:

| build of the identical program | miss path |
|---|---|
| native `#` privates | `200 ""` |
| privates lowered to WeakMaps | **`404`, correct** |

So it sits in the same area as #8962, but it is **not** the same defect and
#8962's fix does not address it: the run above is with that fix applied, and
the minimal case can be reproduced in a single bundled module where no
imported class exists at all.

`HonoBase` dispatch runs through `#dispatch()` and `#notFoundHandler` — a
private method reading a private field — so the likely shape is a private
member read returning the wrong value (or a brand check silently failing)
rather than throwing.

## Why it is worth prioritising

It is the quiet kind. #8962 crashed at construction, which is at least loud.
This one ships: an API returns `200` with an empty body for every unrouted
request, so clients see success, monitoring sees 2xx, and nothing in the logs
says otherwise. For a JSON API where the 404 envelope is part of the contract,
that is a correctness failure that survives all the way to production.

Found while bringing up a real service; happy to test a patch on this host
against both the two-line repro and the full application.

proggeramlug:
## Summary

- lower the read half of compound/logical private-member assignments through the same brand guard and class-mangled storage key as ordinary private reads
- preserve runtime `HeadersInit` records passed through non-literal `Response` init objects
- install the Fetch default `text/plain;charset=UTF-8` for string bodies without overriding explicit content types, while clearing pending BodyInit metadata across throwing init evaluation

Hono's `Context.res` getter uses `this.#res ||= new Response(...)`. The old read addressed the source key `#res` instead of the field's mangled storage key, so it discarded finalized responses (including custom 404s) and returned a fresh empty 200. Hono also passes prepared headers through runtime init objects, which the Response constructor previously treated as missing registry handles.

## Verification

- `cargo test -p perry --test issue_8968_private_compound_assignment --test issue_8968_response_headers -- --nocapture`
- `cargo test -p perry-hir -p perry-stdlib`
- `cargo test -p perry-codegen --lib` (1341 passed, 1 ignored)
- Hono 4.13.4 exact repro, on a temporary stack containing the separate open #8962 fix:

```
/text    200 "text/plain;charset=UTF-8" "hello"
/json    200 "application/json" "{\"ok\":true}"
/html    200 "text/html; charset=UTF-8" "<p>hi</p>"
/raw     200 "application/json" "raw"
/missing 404 "application/json" "{\"code\":\"NOT_FOUND\"}"
```

The #8962 commit was used only for end-to-end validation and is not part of this PR.

Fixes #8968

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved `Response` construction with string bodies and plain-object headers.
  - Preserved correct `content-type` values and prevented failed initialization from leaking metadata.
  - Fixed compound assignments involving private members.
  - Updated `perry run <dir>` to resolve project entry files correctly.
  - Runtime-string `Function` calls now raise catchable `TypeError` when dynamic evaluation is disabled.
  - Improved property access performance without changing behavior.

- **Tests**
  - Added regression coverage for Response handling and private-member assignments.

### DIFF
--- a/changelog.d/8979-run-directory-entry.md
+++ b/changelog.d/8979-run-directory-entry.md
@@ -1,0 +1,5 @@
+`perry run <dir>` now resolves the directory to its project entry instead of passing the directory itself into module collection.
+
+An explicit directory input is treated as a project root: its `perry.toml` entry is read relative to that directory, falling back to `<dir>/src/main.ts` then `<dir>/main.ts`. Previously `perry run .` read `perry.toml` from the current working directory and handed the directory straight to module collection, which fails on Windows.
+
+Fixes #8908.
--- a/changelog.d/8980-private-guard-call-site.md
+++ b/changelog.d/8980-private-guard-call-site.md
@@ -1,0 +1,22 @@
+The private-member guard moved to its call sites, taking a call off every
+ordinary property read and write.
+
+#8970 made the private-member name test cheap but left the CALL. In a pure
+property-read loop (`o[k]` with pre-built keys, no concat) that showed up as
+`private_member_get_by_name` 11.4% plus `private_member_storage_name` 5.4% —
+**16.8% of the loop, the largest single item** — essentially all of it call
+overhead for keys that are rejected on their length before doing anything.
+
+The guard is now invoked at the three call sites (the generic read entry, the
+class-field read miss, and the generic write), so an ordinary property
+operation makes no call into the private-member path at all. Keys that pass
+the guard take exactly the original path.
+
+Interleaved A/B, min-of-21 (under heavy co-tenant load, so read the ratios
+rather than the absolutes): pure property read 26 → 22 ms (−15%), computed-key
+read 51 → 45 ms (−12%), combined overwrite 50 → 46 ms (−8%), write unchanged.
+
+Output on a private-member exercise — instance fields, `static #instances`,
+private methods, private getters, `#x in obj`, subclassing, and an ordinary key
+literally named `#<perry:private-member:1:x>` — is byte-identical to before the
+change. Computed-key differential vs node is byte-identical. Suite 2779 passed.
--- a/changelog.d/8981-dynamic-function-refusal-unwind.md
+++ b/changelog.d/8981-dynamic-function-refusal-unwind.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Refusing a runtime-string `Function` in a `dyn-eval`-off AOT binary now throws a catchable `TypeError` instead of aborting at the runtime FFI boundary, allowing zod v4 and other capability-probing libraries to select their non-eval fallback.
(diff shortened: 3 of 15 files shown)


proggeramlug:
Merged (batched with #8981 and #8983).

**One fix pushed: this PR had no changelog fragment.** It touches `crates/` and carries no `skip-changelog` label, so the changeset gate in `lint` would have rejected it. Added `changelog.d/8982-hono-response-state.md` covering all three defects — the compound/logical private-member read lowering, the `HeadersInit` record preserved through a non-literal `Response` init, and the Fetch `text/plain` default that no longer overrides an explicit content type.

Lowering the read half of `this.#x ||= v` through the same brand guard and class-mangled storage key as an ordinary private read is the right shape: a differently-keyed read path is the kind of asymmetry that works until someone reads and writes the same private field two ways.

**Validation** — runtime 2780/0, hir 355/0, codegen 1341/0, stdlib 124/0 (`RUST_TEST_THREADS=1`); `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green — the exception is the pre-existing Actions-expression artifact (#8929).

PR #8983, opened 2026-08-28
perf(runtime): skip dead feedback observation + field-only shape reads (computed-key read now at node parity)

proggeramlug:
Two dead-work removals on the property read path, both taken from the top of an isolated read profile.

**1. Typed-feedback observation when recording is off.** Recording is off by default, and both `guard_observe` and `record_fallback_call` early-return in that mode — but the property wrappers built the entire `Observation` first, hashing the key and resolving the receiver's shape, purely to hand it to functions that discard it. `js_typed_feedback_object_get_field_by_name_f64` was **10% of an isolated property-read loop**, nearly all of it that.

The array index wrappers have carried #5094's gate for exactly this reason, and #8951 gave it to the fast store path. The property get/set wrappers never got it.

Behaviour is unchanged in both modes, and provably so: with recording off `guard_observe` returns `contract_valid` unmodified and the fallback recorder is a no-op, so the wrapper already reduced to precisely the underlying call it now makes directly.

**2. Field-only shape reads.** `shape_descriptor_by_id` returns `ShapeDescriptor` **by value**, so `object_live_slot_count` — the slot bound consulted on essentially every property read and write — lifted the whole ~48-byte record and kept four bytes of it. It now reads the field through the table's record using the same way-cache probe and the same epoch validation. `shape_descriptor_by_id` was **10.1%** of the same loop.

### Measurement

Interleaved A/B, min-of-21, built from the exact parent commit and this commit in one run. Node on the same host in brackets:

| loop | base | this PR | |
|---|---|---|---|
| pure property read | 21 ms (node 4) | **17 ms** | −19% |
| computed-key read | 27 ms (node 23) | **23 ms** | −15% — **now at parity with node** |
| combined overwrite | 46 ms (node 31) | **41 ms** | −11% |
| write only | 21 ms (node 23) | 21 ms | unchanged; already ahead of node |

Means track the mins (22 → 18, 28 → 24, 46 → 42).

### Correctness

- `perry-runtime`: **2779 passed / 0 failed**, including the 55 typed-feedback tests — that gate changes when observations get recorded, so it is the one to watch.
- Private-member exercise: **byte-identical** to base.
- Computed-key differential vs node: **byte-identical**.

### Where the remaining gap is

The pure property read (`o[k]`, key already built) is now 17 ms against node's 4 ms and is the largest single remaining gap in this area — worth noting that perry's *key construction* is already faster than node's, so this is the lookup itself.

## Summary by CodeRabbit

  * Improved property-read performance when tracing is disabled.
  * Reduced overhead when determining available object slots.
  * Preserved existing write performance and behavior.

* **Reliability**
  * Maintained existing fallback behavior for unavailable shape information.
  * Verified changes with passing tests and byte-identical outputs.


proggeramlug:
Merged (batched with #8981 and #8982). Both removals are dead work rather than new behaviour — building an `Observation` (hashing the key, resolving the receiver's shape) only to hand it to functions that early-return when recording is off is pure waste, and the gate is checked before the construction rather than inside it.

**One fix pushed:** the fragment was named `8981-feedback-gate-and-shape-field.md` — that is **#8981's** number, a different PR in this same batch. Renamed to `8983-`. Left as-is it would have collided with #8981's own fragment at release time.

That makes three fragment-naming problems in this batch of three: one named for an issue, one carrying another PR's number, one missing entirely. I filed #8978 about this earlier — the gate validates the filename *shape* (`^changelog\.d/[0-9]+-[^/]+\.md$`), so any number passes and only a *missing* fragment is caught.

**Validation** — runtime 2780/0, hir 355/0, codegen 1341/0, stdlib 124/0 (`RUST_TEST_THREADS=1`); `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green — the exception is the pre-existing Actions-expression artifact (#8929). The node-parity claim is not re-measured here.

PR #8984, opened 2026-08-28
fix: preserve private field update and reflection semantics

### LINKED ISSUE #8969: Private class field: `this.#n += 1` and `this.#n++` produce NaN (compound/update lowering drops the read)

proggeramlug:
## Summary

Fix private-field read/modify/write lowering so compound and logical assignments read the guarded class-qualified field slot, and keep private fields out of public class shapes and reflection.

## Changes

- Lower private assignment-target reads through the same brand guard and storage key as ordinary private reads.
- Exclude private fields from packed public class keys, allocation counts, and typed-layout masks, including inherited/imported layouts.
- Treat the transient `#<perry:private-member:...>` routing spelling as ordinary user data when no private-access hint is pending.
- Add a pinned-Node differential regression covering compound/logical assignments, fresh prefix/postfix updates, static updates, reflection, collision keys, and inherited mixed public/private layouts.

## Related issue

Fixes #8969

## Test plan

- [x] `cargo check -p perry-hir -p perry-codegen -p perry-runtime`
- [x] `cargo test -p perry-hir` (all pass)
- [x] `cargo test -p perry-codegen --lib` (1,341 passed, 1 ignored)
- [x] Pinned Node 26.5.1 parity: `./run_parity_tests.sh --filter test_gap_8969_private_field_compound_update` (1/1)
- [x] Existing private parity sweep: `./run_parity_tests.sh --filter private` (5/5)
- [ ] Full workspace test suite passes
- [x] Added a user-facing regression under `test-files/`

The broader `cargo test -p perry-codegen` run reaches one existing failure in `temp_root_operand_temporaries::string_literal_concat_operand_is_re_derived_below_the_allocating_sibling`; the same emitted IR and assertion failure reproduce on clean `origin/main` at `f4e820632`.

## Screenshots / output

Before, the issue repro produced `compound: NaN`, exposed `#n`, and dropped the routing-prefix collision key. The regression now matches Node byte-for-byte: the compound value is `2`, private own-key lists are empty, and the ordinary collision key remains readable, enumerable, and owned.

## Checklist

- [x] My commits follow the repository's `fix:` convention

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed compound, logical, and increment assignments on private fields so values are preserved correctly.
  * Ensured private fields remain excluded from object keys, enumeration, spreading, and JSON serialization.
  * Preserved ordinary properties whose names resemble internal private-member routing keys.
  * Prevented private fields from affecting public class shape and inherited property layouts.

* **Tests**
  * Added coverage for instance and static private-field updates, reflection behavior, inheritance, and serialization.

### DIFF
--- a/changelog.d/8984-private-field-updates.md
+++ b/changelog.d/8984-private-field-updates.md
@@ -1,0 +1,7 @@
+Private class fields now preserve their value under compound and logical
+assignments instead of reading `undefined` and storing `NaN`.
+
+Private fields no longer occupy public class-shape keys, so they stay absent
+from `Object.keys`, `Object.getOwnPropertyNames`, `for...in`, spread, and JSON
+serialization. An ordinary property whose name matches Perry's transient
+private-member routing spelling is now retained as ordinary user data.
--- a/crates/perry-codegen/src/codegen/mod.rs
+++ b/crates/perry-codegen/src/codegen/mod.rs
@@ -1164,15 +1164,17 @@ pub fn compile_module(hir: &HirModule, opts: CompileOptions) -> Result<Vec<u8>> 
         // first (walking from deepest ancestor down) so the slot
         // order matches `class_field_global_index`'s assumption.
         let mut packed_keys = String::new();
-        // Skip computed-key fields (`[Symbol.for("k")] = …`): their key is an
-        // expression evaluated at runtime, not a stable string, so they don't
-        // get an inline slot. Including their synthetic `__computed_field_*`
-        // names in the packed keys would surface them as enumerable own
-        // properties via Object.keys() and inflate the inline-slot count.
-        // Their values are stored via `apply_field_initializers_recursive`'s
-        // IndexSet path → js_object_set_field / js_object_set_symbol_property.
+        // Skip computed-key fields (`[Symbol.for("k")] = …`) and private
+        // fields. Computed keys are evaluated at construction time; private
+        // fields live in class-id-qualified runtime storage installed by
+        // `js_private_field_add`. Neither is a public inline shape key.
+        // Including either synthetic/source spelling in packed keys leaks it
+        // through reflection and inflates/misaligns the inline-slot layout.
         let count_keyable = |fields: &[perry_hir::ClassField]| -> u32 {
-            fields.iter().filter(|f| f.key_expr.is_none()).count() as u32
+            fields
+                .iter()
+                .filter(|f| f.key_expr.is_none() && !f.is_private)
+                .count() as u32
         };
         let mut total_field_count = count_keyable(&c.fields);
         // (parent_name, resolved_fields) captured during the chain walk so we
@@ -1251,7 +1253,7 @@ pub fn compile_module(hir: &HirModule, opts: CompileOptions) -> Result<Vec<u8>> 
         // (which would risk re-picking the wrong same-named stub).
         for (_parent_name, parent_fields) in parent_chain.iter().rev() {
             for f in parent_fields {
-                if f.key_expr.is_some() {
+                if f.key_expr.is_some() || f.is_private {
                     continue;
                 }
                 packed_keys.push_str(&f.name);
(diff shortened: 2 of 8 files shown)


proggeramlug:
Merged, rebased onto current `main`.

**It conflicted with #8982**, which landed minutes earlier and fixed the same private-read lowering in `lower_patterns.rs`. The two sides were **semantically identical** — same `wrap_private_guard(…, PRIV_OP_READ)`, same `private_storage_property` — differing only in import qualification and comment. I kept main's call form (its imports are already in place) and this PR's comment, because yours names the observed symptom:

> Reading `#n` as a public property returns `undefined`, which made `this.#n += 1` store NaN in the real slot.

That is worth keeping. "Use the same storage key" explains the change; "otherwise `+= 1` writes NaN" explains why anyone should care, and is what a future reader needs when deciding whether a refactor here is safe.

The rest of the PR does not overlap and is the more valuable half — excluding private fields from packed public class keys, allocation counts and typed-layout masks (including inherited/imported layouts), and treating the transient `#<perry:private-member:…>` routing spelling as ordinary user data when no private-access hint is pending. That last one is the sort of thing that only shows up when a user legitimately has a `#`-prefixed key.

**Validation** — hir 355/0, codegen 1341/0, runtime 2780/0 (`RUST_TEST_THREADS=1`); `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green — the exception is the pre-existing Actions-expression artifact (#8929).

The fragment was correctly PR-keyed, which after eight naming fixes today I noticed and appreciated.

PR #8985, opened 2026-08-28
perf: stop re-entering the runtime to append to and pop from an Array subclass's own elements store (−9.5% / −11.4%)

proggeramlug:
Follow-up to #8966/#8974/#8976. With the elements store as the representation, a fresh profile of the wolf-ecs twins showed the append/pop path re-validating a store that belongs to the instance: `elements_push` 3.8% + `RuntimeHandleScope::push` 1.5% + `js_array_pop_f64` 5.1% (a runtime entry whose only job was to follow one pointer to the store and pop from it).

## What

**Runtime — lean append and tail pop.** The store is reached through the meta slot: never a proxy, never a forwarding stub, its header one read away, and an in-capacity append cannot allocate.

* an append with spare capacity stores directly (header read, integrity-flag test, `store_array_slot_resolved`, length bump) — no handle scope, no owner/value rooting, no head write-back, none of `js_array_push_f64`'s classification for an arbitrary caller pointer;
* a non-hole tail pop is a load and a length decrement;
* growth still takes the complete entry with the owner rooted across it and publishes the re-allocated head; holes, an empty store, and frozen/sealed/no-extend/descriptor-bearing stores keep the runtime path unchanged. The element bookkeeping the read tiers and the loop guard consume is untouched.

**Codegen — the inline pop tier resolves the payload.** `sub.pop()` failed the tier's `GC_TYPE_ARRAY` gate (the receiver is the object) and called the runtime. A `GC_TYPE_OBJECT` receiver now loads its meta record and `ObjectMeta.elements` (word 12), validates that store exactly as a plain-Array receiver is validated, and the existing length/read/take blocks run on it through a payload phi.

**Cleanup the census asked for.** Five inline tiers each recomputed `object_header_size_bytes(target) - pointer_size` to reach `ObjectHeader::meta`; `object_meta_slot_offset_bytes` derives it once, so the audited object-header-size callsite count drops **46 → 42**. The new store header read uses `addr_class::try_read_gc_header` rather than a hand-rolled cast (the address-classification ratchet forbids those).

## Numbers (Mac mini, 11 alternating pairs, vs a freshly built `origin/main`)

|---|---|---|
| 2 s | 0.3623 → **0.3278** (**−9.53%**, 11/11) | 0.2996 → **0.2655** (**−11.36%**, 11/11) |
| 50 ms | 0.3622 → **0.3279** (**−9.45%**, 11/11) | 0.2994 → **0.2654** (**−11.36%**, 11/11) |

Deltas are tight (add −9.4…−9.7 except one −11.8; entity −11.2…−11.4). Perry vs Node 26.5 on that host: **2.45× / 1.78×** — the entity cycle drops under 2× for the first time in this campaign.

## Verification

* Runtime test: 64 appends across several capacity classes (asserting the head is republished on growth), ordered tail pops, a hole popping through the runtime, an empty-store pop, and a pointer element round-tripping through the funnel.
* IR census test pins the pop tier's elements arm (the word-12 load and the store validation mask).
* Semantics probes vs node: `fill(value, start, end)` and the nested-loop reproducer are token-identical; the subclass probes show only the three known pre-existing divergences (`fill` as an own key #8953, `A.from` not populating, `constructor` after `map`).
* `cargo test -p perry-runtime --lib` 2780/0; `RUSTFLAGS=-D warnings cargo check --workspace --all-targets` clean; census, addr-class, GC store-site inventory, file size, raw-handle debt (unchanged) all pass. Integration: `issue_8655` 2/2, `issue_8773` 4/4, `issue_8690` 3/3, `issue_5898_array_pop_prototype` 1/1, `array_subclass_fill_args` 1/1. (`temp_root_operand_temporaries::string_literal_concat_operand_is_re_derived_below_the_allocating_sibling` fails on plain `origin/main` too — unrelated.)

## Summary by CodeRabbit

- **Performance**
  - Improved `Array` subclass append and pop operations with faster paths for common in-capacity cases.
  - Preserved full runtime handling for growth, holes, empty stores, and other complex cases.
  - Extended inline `pop()` support to elements-backed `Array` subclasses.

- **Bug Fixes**
  - Improved metadata access consistency across supported target platforms.

- **Tests**
  - Added coverage for capacity growth, tail removal, holes, empty stores, and stored pointer values.


proggeramlug:
Merged.

**On the census baseline change** — worth saying explicitly, since a baseline edit usually deserves suspicion: this one *tightens*. Four open-coded `meta_offset` computations collapse into one helper in `target_layout.rs`, so `codegen_object_header_size_sites` goes 46 → 42 and the four per-file entries disappear. Ratchets are supposed to move in that direction, and consolidating a repeated offset into one place is the same move that keeps the `StringHeader` payload offset honest.

**On the lean push/pop** — the claim that the store reached through the meta slot is never a proxy and never a forwarding stub is what the whole optimisation rests on, so I did not take the unit tests as sufficient. #8976 was a real wrong-value bug in this exact chain that both a full unit suite and a 1386-file corpus differential missed, because it needed a specific access shape. So I re-ran the behavioural probe I built for that:

```
nested closure loop over a subclass   perry=10,20,30,        node=10,20,30,
element 0 after a growth realloc      perry=0,1,2,3,4,5,     node=0,1,2,3,4,5,
single-element subclass               perry=7,               node=7,
mixed types at element 0              perry=x,2,             node=x,2,
element 0 after pop()                 perry=1,2,             node=1,2,
```

All five match, including the nested-closure shape #8976 broke and the `pop()` case this PR touches directly.

**Validation** — runtime 2781/0 and codegen 1341/0 (`RUST_TEST_THREADS=1`); `PERRY_ARRAY_SUBCLASS_ELEMENTS=0` also 2781/0, so the kill switch still bisects; `shape_descriptor_census` ok; `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green — the exception is the pre-existing Actions-expression artifact (#8929).

**One fix pushed:** the fragment used the `0000` placeholder; renamed to `8985-`. That is the ninth fragment-naming fix today — see #8978.

The ECS numbers are not re-measured here.

PR #8986, opened 2026-08-28
fix(codegen): initialize imported private brands once

### LINKED ISSUE #8962: new Hono() throws 'Cannot initialize private elements twice on the same object' at runtime

proggeramlug:
## Summary

- identify metadata-only imported class stubs explicitly
- leave private-element brand installation to the defining module's standalone constructor
- add six integration cases covering direct imports, accessors, local and imported subclasses, same-module branding, and genuine duplicate initialization

## Testing

- `cargo test -p perry --test issue_8962_imported_class_private_brand -- --test-threads=1` (6 passed)
- `cargo test -p perry-hir -p perry-codegen --lib` (1,696 passed, 2 ignored)
- Hono 4.13.4 reproduction on `perrymaster.skelpo.net` (`new Hono()` plus route registration)

## Known baseline

The full `perry-codegen` integration run reaches an unrelated stale assertion in `temp_root_operand_temporaries`: it expects `i64 @js_string_concat_value`, while current main emits `double @js_string_concat_value_box`. The changed files do not touch string concatenation.

Fixes #8962

### DIFF
--- a/changelog.d/8986-imported-private-brands.md
+++ b/changelog.d/8986-imported-private-brands.md
@@ -1,0 +1,5 @@
+Imported private brands are installed once, by the defining module's standalone constructor.
+
+A metadata-only imported class stub is now identified explicitly, so importing a class that uses private elements no longer re-runs brand installation in the importing module. Re-branding produced a second brand for the same class, so a private access that had been valid through one import path failed through the other.
+
+Covers direct imports, accessors, local and imported subclasses, same-module branding, and genuine duplicate initialization.
--- a/crates/perry-codegen/src/lower_call/field_init.rs
+++ b/crates/perry-codegen/src/lower_call/field_init.rs
@@ -693,10 +693,39 @@ pub(crate) fn apply_field_initializers_recursive(
                 None => init_pairs.push((field.name.clone(), init, field.is_private)),
             }
         }
+        // #8962: an IMPORTED class installs nothing here. Its whole
+        // field-initializer phase — public field writes, private-field adds AND
+        // the shared private brand — is baked into the defining module's
+        // standalone `<prefix>__<class>_constructor`, which `codegen/method.rs`
+        // emits for exactly that reason ("At the `new ImportedClass(...)` call
+        // site, `lower_new` applies initializers against the imported class
+        // stub — which has none"). That premise holds for FIELDS because the
+        // stub flattens every field to `is_private: false` with `init: None`,
+        // so the worst this loop could do was write `undefined` into a slot the
+        // real constructor overwrites moments later.
+        //
+        // It does NOT hold for the private BRAND. The stub copies private
+        // METHOD and accessor names verbatim (it needs them to resolve dispatch
+        // symbols), and `has_private_instance_brand` is defined purely over
+        // `#`-prefixed method/getter/setter names — so a stub answers `true` and
+        // this site emitted `js_private_brand_add` at the importing module's
+        // `new`, on top of the one the defining module's constructor emits.
+        // Installing a class's brand twice on one object is the observable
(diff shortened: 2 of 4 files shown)


proggeramlug:
Merged (batched with #8987). Leaving brand installation to the defining module's standalone constructor is the right call — re-branding in the importing module produces a second brand for the same class, so a private access valid through one import path fails through the other, which is exactly the sort of thing that looks like "private fields are broken" rather than "imports are branded twice".

The six integration cases are well chosen: direct imports, accessors, local *and* imported subclasses, same-module branding, and genuine duplicate initialization. That last one matters most — it is the case that distinguishes "installed once" from "never installed again", and without it the fix could silently skip a brand that really was needed.

**One fix pushed: this PR had no changelog fragment.** It touches three `crates/` files and carries no `skip-changelog` label, so the changeset gate in `lint` would have rejected it. Added `changelog.d/8986-imported-private-brands.md`.

**Validation** — codegen 1341/0, runtime 2781/0, hir 355/0 (`RUST_TEST_THREADS=1`); `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green — the exception is the pre-existing Actions-expression artifact (#8929).

PR #8987, opened 2026-08-28
fix(runtime): inherit Array-subclass fill

### LINKED ISSUE #8953: Array subclass instances: getOwnPropertyNames / for..in segfault; Object.keys leaks "length" and "fill"

proggeramlug:
## Summary

- keep `Array.prototype.fill` inherited on elements-backed Array-subclass instances instead of stamping an enumerable own closure
- resolve inherited property reads and direct `fill()` calls while preserving class-method and explicit-prototype overrides
- add an end-to-end regression for `Object.keys`, `for...in`, `Object.getOwnPropertyNames`, direct/extracted `fill`, and class overrides

## Tests

- `cargo check -p perry-runtime -p perry`
- `cargo test -p perry-runtime --lib -- --test-threads=1` (2,759 passed; 0 failed; 4 ignored)
- `cargo test -p perry --test issue_8953_array_subclass_enumeration --test array_subclass_fill_args --test issue_array_subclass_super_init -- --nocapture`
- `cargo test -p perry --test issue_8655_array_subclass_indexing -- --nocapture`

Closes #8953

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Array subclass property enumeration to match standard JavaScript behavior.
  * `Object.keys` and `for...in` now include only enumerable indices.
  * `Object.getOwnPropertyNames` includes `length` without incorrectly listing inherited methods.
  * Restored inherited `Array.prototype.fill` behavior for Array subclasses, including subclasses that override it.


proggeramlug:
Merged (batched with #8986). This is the follow-on to #8972, which I merged earlier tonight — that PR fixed `fill`'s arity but reached it by stamping an enumerable own closure, so `Object.keys` and `for...in` saw a `fill` property that node does not expose. Keeping it inherited while still resolving direct `fill()` calls, and preserving class-method and explicit-prototype overrides, is the correct shape; the end-to-end regression over `Object.keys` / `for...in` / `Object.getOwnPropertyNames` is what pins it.

**Two fixes pushed:**

1. The fragment was named `8953-array-subclass-enumeration.md` — #8953 is the issue, not this PR. Renamed to `8987-`.
2. `field_get_set/accessors.rs` open-coded the `StringHeader` payload offset a second time, raising the string-payload-access ratchet 364 → 365. Switched the new site to `crate::object::string_header_payload` (the pre-existing baselined site left alone), which puts the count back to baseline.

That payload helper has now caught the same slip in #8936, #8966 and here — it is easy to reach for the offset directly when the surrounding code already does.

**Validation** — runtime 2781/0, codegen 1341/0, hir 355/0 (`RUST_TEST_THREADS=1`); `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green after the fix — the exception is the pre-existing Actions-expression artifact (#8929).

PR #8988, opened 2026-08-28
perf(runtime): megamorphic read stub cache (pure property read −12%)

proggeramlug:
The read twin of the dynamic-write stub (#8965, made 2-way in #8977). This one is **2-way set-associative from the start** — #8977 already measured what direct-mapped costs, and there was no reason to repeat that experiment: a colliding pair evicts each other on every rotation through the key set, so both miss *forever*.

A hit skips `js_object_get_field_by_name`'s fast-lane guard chain — address class, interned-key flag, arena classification, header type/flags/class, keys-array validation — and the read-plan probe. That last one matters more than its own profile share suggests: the plan's epoch is bumped by the incremental collector at loop-poll cadence, so on a steady read loop it is repeatedly cold and falls through to a shape-index hash lookup.

### Measurement

Interleaved A/B, min-of-21, built from the exact parent commit and this commit in one run:

| loop | base | this PR | |
|---|---|---|---|
| pure property read | 17 ms | **15 ms** | −12% min, −17% mean |
| computed-key read | 23 ms | **22 ms** | −4% min, −8% mean |
| combined overwrite | 41 ms | **39 ms** | −5% min, −7% mean |
| write only | 21 ms | 21 ms | unchanged |

### Safety

Entries store **content bits, never an address**, so a key that dies and has its address recycled cannot produce a false hit; keys that do not fit the inline form are not cached at all. Every hit re-validates heap-object type, not-forwarded, blocking flags, class id, and the receiver's **current shape token** — which pins the exact key set *and order*, so a match means the cached slot still names this key. A stale entry misses; it cannot resolve to the wrong property.

Placement carries its own argument: the probe sits **after** the `process.env` and Proxy arms so those keep their semantics, and the stub is only ever primed from *inside* the fast lane — once the receiver has been proved an ordinary shaped object with a resolvable own slot.

### Adversarial testing

A wrong-slot read is silent corruption rather than a crash, so this ships with a differential built to attack each invariant directly:

- a `delete` that changes the shape under a cached slot, then re-add;
- `Object.defineProperty` installing a **getter over a cached data slot**;
- prototype fallback after the own property is deleted;
- `Object.freeze`;
- the same two keys inserted in **opposite orders** (distinct shapes, same key set);
- a 300-key object whose slots live in the **overflow store**, before and after a delete.

Output is **byte-identical to node** on all of it. Also: `perry-runtime` 2779 passed / 0 failed; private-member output identical to base; computed-key differential identical to node.

## Summary by CodeRabbit

- **Performance**
  - Improved performance for repeated dynamic property reads through optimized caching.

- **Bug Fixes**
  - Fixed private methods and accessors on imported classes, preventing duplicate brand initialization errors.
  - Updated array-subclass enumeration to match Node.js behavior.
  - Corrected `Array.prototype.fill` inheritance so it is not incorrectly reported as an own property.
  - Improved property and method lookup for array subclasses, including overridden and inherited `fill` behavior.


proggeramlug:
Merged. Going 2-way from the start is right — #8977 already paid for that experiment, and a colliding pair in a direct-mapped table misses *forever* rather than occasionally.

**Three fixes pushed**, one of which was a real gate failure:

1. **`READ_STUB` was an unclassified GC root holder.** `gc_runtime_root_holders` flagged it as a new identity-ratcheted `perry_thread_local!` with no verdict. I checked rather than pinning it as debt: `read_stub_key_bits` returns `short_ascii_sso_bits(key)` — the key's characters packed inline — and `None` for anything else, so a way holds a shape id, packed content and a slot index, and never a heap address. Recorded the same `not_a_gc_pointer` verdict `WRITE_STUB` carries, with the read-side reasoning spelled out (a stale entry cannot hit either: `receiver_shape_token` returns `None` for a receiver with no live shape, and the token identifies key set *and* order).

   Worth stressing this is a *researched* verdict, not a pin — a pinned entry would have satisfied the gate without establishing anything.

2. **A duplicate fragment.** This branch predates #8983 and re-added `8981-feedback-gate-and-shape-field.md`, which is byte-identical to the `8983-` file already on `main`. Left in, the release notes would list that change twice under two PR numbers. Removed.

3. `8984-read-stub-cache.md` → `8988-`. #8984 is a different merged PR.

**Validation** — runtime 2781/0, codegen 1341/0 (`RUST_TEST_THREADS=1`); under `PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1` the failing count is 16, the pre-existing set; `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green after the verdict — the exception is the pre-existing Actions-expression artifact (#8929).

The −12% is not re-measured here.

PR #8989, opened 2026-08-28
test(parser): cover Unicode whitespace regex literals (#8902)

### LINKED ISSUE #8902: Non-ASCII character written literally in a regex literal never matches (escapes and new RegExp are fine)
**perry 0.5.1220**, macOS arm64.

A non-ASCII character written *literally* inside a regex pattern does not match. The same character expressed as a `\uXXXX` escape, or via `new RegExp`, matches correctly — so it looks like the pattern parser rather than the matcher.

## Repro

```ts
function cp(s: string): string {
  return [...s].map((c) => c.codePointAt(0)!.toString(16).padStart(4, "0")).join(" ");
}

function main(): void {
  const sample = `x${String.fromCharCode(0x00a0)}y`;   // x <U+00A0> y

  console.log("literal :", cp(sample.replace(/ /g, " ")));            // the / / holds a literal U+00A0
  console.log("escape  :", cp(sample.replace(/[ ]/g, " ")));
  console.log("ctor    :", cp(sample.replace(new RegExp("[\\u00a0]", "g"), " ")));
}
main();
```

`perry compile repro.ts -o repro && ./repro`:

```
literal : 0078 00a0 0079      <-- not replaced
escape  : 0078 0020 0079
ctor    : 0078 0020 0079
```

`node repro.ts`:

```
literal : 0078 0020 0079      <-- replaced
escape  : 0078 0020 0079
ctor    : 0078 0020 0079
```

Confirmed the same way for U+202F (narrow no-break space). Note the literal bytes matter — in the file above, the first pattern contains the raw two bytes `c2 a0`.

## Why this is worth fixing rather than documenting

String literals themselves are perfectly preserved — `"1.234,56 €"` and `"Zubehör Ärger groß"` round-trip byte-identically under Perry, and `String.fromCharCode` comparisons against source literals are `true`. So a developer has no reason to suspect the regex is the odd one out.

The failure is silent: the replace simply does nothing and the string flows on. We hit it in a test that normalises the non-breaking space `Intl.NumberFormat("de-DE")` puts before `€`; the normaliser quietly became a no-op and every downstream comparison failed on a character that is invisible in the diff. For a German-language codebase, a regex containing a literal `ö` or `ß` that silently never matches is a bad class of bug to have available.

## Environment

proggeramlug:
Closes #8902.

## Summary

#8902 was reported against 0.5.1220. The underlying raw-non-ASCII regex-literal parser bug was fixed later by #7431 (at workspace version 0.5.1280), so current `main` already preserves and matches the reported U+00A0 and U+202F characters.

This PR adds the exact missing regression coverage:

- parser pre-pass tests for raw NBSP and narrow NBSP, both standalone and in character classes
- SWC AST checks proving both whitespace code points survive unchanged
- an end-to-end parity fixture covering literal, character-class, escaped, and `RegExp` constructor forms

No production code or version metadata changes are needed.

## Validation

Run on `root@perrymaster.skelpo.net` from the isolated worktree:

- `cargo test -p perry-parser` — 40 passed
- fresh worktree compiler output for the new fixture is byte-identical to Node (six code-point lines)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed regular expressions containing non-breaking spaces and narrow non-breaking spaces so raw and escaped forms now behave consistently.
  * Corrected matching for these characters both as standalone patterns and within character classes.

* **Tests**
  * Added coverage for literal regex patterns and `RegExp` constructor patterns involving Unicode whitespace.

### DIFF
--- a/changelog.d/8989-regex-literal-unicode-whitespace.md
+++ b/changelog.d/8989-regex-literal-unicode-whitespace.md
@@ -1,0 +1,6 @@
+### Fixed
+
+- Regex literals containing raw no-break space (U+00A0) or narrow no-break
+  space (U+202F) match the same way as their escaped forms. Parser and runtime
+  parity coverage now guards both standalone and character-class patterns
+  (#8902).
--- a/crates/perry-parser/src/lib.rs
+++ b/crates/perry-parser/src/lib.rs
@@ -1737,6 +1737,14 @@ class C {
             "const re = /schön/i;\n",
             "const re = /Maß/;\n",
             "const re = /\\s*(\\d+)\\s+([\\d.,]+)\\s*€/;\n",
+            // #8902: whitespace code points are pattern characters too. The
+            // report used both literal and character-class forms of NBSP and
+            // narrow NBSP; keep them raw here (the Rust escapes decode before
+            // this source reaches the TypeScript parser).
+            "const re = /\u{00a0}/g;\n",
+            "const re = /[\u{00a0}]/g;\n",
+            "const re = /\u{202f}/g;\n",
+            "const re = /[\u{202f}]/g;\n",
             // A non-ASCII character immediately after a backslash (identity
             // escape) took the other corrupting arm of the same match.
             "const re = /\\€/;\n",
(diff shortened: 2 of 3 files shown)


proggeramlug:

Issue #6710, opened 2026-07-20
Authenticated App Router browser SSR calls legacy react-dom/server API (E394 InvariantError) — full-HTML render resolves streaming export to the poison stub


proggeramlug:
## Summary

A logged-in user loading an authenticated App Router page in a **real browser** (full HTML SSR) crashes under Perry with Next.js **E394** — `Internal Error: do not use legacy react-dom/server APIs` — repeated as `unhandledRejection: InvariantError`, and the response stream never cleanly closes (browser spins "loading forever"; curl reports 500). Node renders the identical request 200.

E394 is a Next.js **poison-pill stub**: `function a(){ throw Error("Internal Error: do not use legacy react-dom/server APIs...") }` assigned to the legacy `react-dom/server` exports (`renderToString`, `renderToStaticMarkup`, `renderToNodeStream`, …). App Router must only call the streaming APIs. So under Perry, **a `react-dom/server` module export is resolving to the throwing legacy stub instead of the real streaming function** during the browser SSR of an authenticated page.

## Reproduction (gscmaster, Next 16.1.1 standalone)

| request | Perry | Node |
|---|---|---|
| `GET /de/dashboard` with `Accept: text/html` + browser UA, authenticated | **500 + E394/InvariantError flood** | **200** |
| `GET /de/dashboard` plain curl (no Accept/UA), authenticated | 200 (byte-identical to node) | 200 |
| `GET /de/dashboard` with `RSC: 1` (client nav), authenticated | 200 | 200 |
| public pages with browser UA | 200 | 200 |

So it is specific to the **full-HTML SSR path of an authenticated, `dynamic = "force-dynamic"` page** — exactly what a browser does on the post-login redirect. Plain curl and the RSC-only path take a different render route and succeed, which is why a 220-request curl soak passed while a real browser login hangs.

Likely in the react-dom/server (Fizz) HTML-shell render, possibly interacting with the PPR/postpone-resume path (the bundle also carries `Invariant: isDynamicPostpone misidentified a postpone reason`), cf. the module-export "value form ≠ static form" resolution family (#6667/#6668) and PPR resume #5437.

## Next step

Wrap the bundle's E394 stub `a()` to log a stack when called → identify exactly which `react-dom/server` export is mis-resolved to the stub, then fix Perry's resolution of that export (probable: the streaming API being read as a value/namespace member and getting the legacy alias).

[...]

PR #8990, opened 2026-08-28
perf(runtime): decide the URLSearchParams backing probe from the GC header

proggeramlug:
`url_search_params_backing_of` opens a handle scope, **allocates a heap string** for the backing key and runs a full by-name field lookup — for *any* pointer-tagged receiver. Only an ordinary object can carry a named field, so every `Map`, `Set`, array, closure and Date cell paid all three to be told `None`.

That matters because the probe sits on two type-erased hot paths:

- `js_get_iterator` — every `for…of` over a pointer value reaches it **before** any collection arm;
- the generic `js_native_call_method` dispatch.

On a 4k-entity ECS archetype-migration frame it measured **~14% of GetIterator and ~16% of native method dispatch**.

## The change

One gate on `obj_type == GC_TYPE_OBJECT`, placed before the handle scope and the allocation. This is the same gate `shape_is_url_search_params` already applies twenty lines below, for the same stated reason (#5964: a Date cell read at `ObjectHeader` offsets faults).

Behaviour is unchanged by construction: a `class X extends URLSearchParams` instance — the #6710 case this function exists for — **is** an ordinary object, so it still reaches the lookup. Everything the gate rejects has no named fields at all and could only ever have returned `None`.

## Tests

New `backing_probe_gate_tests::backing_probe_declines_non_object_receivers` asserts the decision through the public entry point for the receivers the hot paths actually hand it (a `Map`, a `Set`, an ordinary object without the field, and a non-pointer), rather than trying to count allocations.

Local gate on the Linux build host: **perry-runtime serial 2761 passed / 0 failed**; `url` 16/16, `iterator` 31/31, new gate test 1/1. Full gate receipt to follow in a comment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved URL search parameter handling for unsupported receiver types.
  * Prevented invalid values, including collections, arrays, functions, and dates, from being treated as search parameter objects.
  * Preserved support for valid objects with search parameter backing data.

### DIFF
--- a/changelog.d/8990-url-search-params-header-probe.md
+++ b/changelog.d/8990-url-search-params-header-probe.md
@@ -1,0 +1,3 @@
+`url_search_params_backing_of` decides from the GC header instead of a by-name field lookup.
+
+The probe previously opened a handle scope, allocated a heap string for the backing key and ran a full by-name lookup for *any* pointer-tagged receiver — so every `Map`, `Set`, array, closure and Date cell paid all three only to be told `None`. Only an ordinary object can carry a named field, and the header already says so.
--- a/crates/perry-runtime/src/url/search_params.rs
+++ b/crates/perry-runtime/src/url/search_params.rs
@@ -70,6 +70,30 @@ pub(crate) fn url_search_params_backing_of(object: f64) -> Option<*mut ObjectHea
     if obj.is_null() {
         return None;
     }
+    // Only an ORDINARY object can carry the named backing field, so every
+    // other pointer receiver is decided here — before the handle scope and
+    // the key-string allocation below.
+    //
+    // This probe sits on two type-erased hot paths: `js_get_iterator` (every
+    // `for…of` over a pointer value reaches it before the collection arms)
+    // and the generic `js_native_call_method` dispatch. Both hand it Maps,
+    // Sets, arrays, closures and Date cells, none of which have named fields
+    // at all — and each such call was allocating a heap string for the key,
+    // rooting the receiver across it, and running a full by-name lookup only
+    // to return `None`. On a 4k-entity ECS archetype-migration frame that was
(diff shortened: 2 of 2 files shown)


proggeramlug:
Local merge gate on the Linux build host, head `2e0ec6735` vs base `4b9dde9a4`:

| suite | result |
|---|---|
| perry-runtime (lib) | **2761 passed / 0 failed** — full parallel run, no flakes this time |
| `cargo fmt` | PASS |

A separate serial (`--test-threads=1`) runtime run was also clean at 2761/0.

proggeramlug:

proggeramlug:
Measurement on the quiet Apple silicon host — 9 alternating pairs, control = merge base `4b9dde9a4`, candidate = this head.

**Row `4k entities: archetype migration (add/remove) + sync`:**

|---|---|---|---|---|---|
| 9-pair screen | 80.220 ms | 68.852 ms | **+14.11%** | **9/9** | 18/18 |

Against node 26.5.1 on the same host (4.32 ms), that moves the row from ~18.6x to ~15.9x.

## One thing worth stating plainly

The `for…of` microbenchmark I used to find this did **not** move: iterating a 4-element `Set` 200k times measured 83.1 ns/loop before and 84.1 ns/loop after.

That is not noise — it is a different code path. A concrete `Set<number>` receiver gets a specialized for-of lowering that never reaches `js_get_iterator`, so the probe is not on its path at all. The ECS code iterates `ReadonlySet`/`ReadonlyMap`-typed parameters, which do go through the generic GetIterator, which is where the probe sits and where the 14% comes from.

So the win is real and unanimous, but it belongs to generic (type-erased) iteration and native method dispatch — not to statically-typed collection iteration, which was already specialized.

PR #8991, opened 2026-08-28
perf(runtime): a plain Map/Set answers GetIterator with its builtin iterator

proggeramlug:
`subclass_backing_for_default_iteration` claims only instances carrying a hidden backing field, so a **plain** `Map` or `Set` fell through every arm of `js_get_iterator` to the generic `[Symbol.iterator]` property lookup plus a dynamic method call.

Measured on small collections — the shape a `for…of` inside a per-entity loop actually has:

| `for…of` over (size 4) | perry | node 26.5.1 | ratio |
|---|---|---|---|
| Set | 85.0 ns/loop | 8.0 ns | **10.6x** |
| Map | 104.0 ns/loop | 21.7 ns | **4.8x** |

On the profile of an ECS archetype-migration frame, `js_get_iterator` was 23.4% of the row, and inside it the symbol-property lookup (42.2%) and the dynamic method call (38.7%) were ~81%.

## The change

Answer a plain registered Map/Set directly with `js_map_entries_iter_obj` / `js_set_values_iter_obj` — the *same* iterator objects the subclass arm already returns, so the protocol the caller drives is unchanged.

The arm sits ahead of the URLSearchParams probes because a plain collection is never a URLSearchParams (that arm requires an ordinary object), so nothing below can claim a receiver this one accepts.

## Safety

A monotone `ITERATOR_PROTOCOL_TOUCHED` latch, fed from the same two symbol set/delete paths that already feed `note_array_proto_iterator_write`. Once any `@@iterator` write is observed, the lane declines and the fully general lookup decides.

The latch is **deliberately coarser** than the array equivalent, and worth calling out: a plain `Map` has no prototype *object* to compare against the way `Array.prototype` does — its surface is served by redirecting the operation at each dispatch point — so the flag is not keyed to the collection prototypes. Any `@@iterator` write in the program disables the lane process-wide. That is the conservative direction (it can only cost speed, never correctness), and programs that never write `@@iterator` keep the lane for free.

## Tests

- `lane_claims_plain_collections_only` — a plain Map and Set are claimed; an ordinary object and a non-pointer are not.
- `a_touched_iterator_protocol_disables_the_lane` — an `@@iterator` write flips the latch and the lane stops claiming; `@@asyncIterator` does not flip it.

Local gate on the Linux build host: **perry-runtime serial 2762 passed / 0 failed**; `iterator` 32/32, `symbol` 49/49, `map_set_subclass` 6/6. Full gate receipt and the paired row measurement follow in comments.

## Summary by CodeRabbit

* **Performance**
  * Improved `for…of` iteration performance for standard `Map` and `Set` collections.
  * Preserved custom iterator behavior when the iterator protocol is modified.
* **Bug Fixes**
  * Ensured iterator changes made through property updates or deletion are correctly recognized during iteration.

### DIFF
--- a/changelog.d/8991-plain-collection-iterator.md
+++ b/changelog.d/8991-plain-collection-iterator.md
@@ -1,0 +1,3 @@
+A plain `Map` or `Set` answers `GetIterator` with its builtin iterator.
+
+`subclass_backing_for_default_iteration` claims only instances carrying a hidden backing field, so a plain collection fell through every arm of `js_get_iterator` to the generic `[Symbol.iterator]` property lookup plus a dynamic method call — the shape a `for…of` over a small collection hits on every iteration.
--- a/crates/perry-runtime/src/object/map_set_subclass.rs
+++ b/crates/perry-runtime/src/object/map_set_subclass.rs
@@ -84,6 +84,68 @@ unsafe fn instance_object_ptr(this: f64) -> Option<*mut ObjectHeader> {
 /// hidden backing field), return its backing collection. Returns `None` for
 /// real Maps/Sets, ordinary objects, and non-objects — so callers fall through
 /// to their existing handling.
+/// Set once any `[Symbol.iterator]` symbol-keyed write or delete is observed,
+/// anywhere in the process.
+///
+/// [`plain_collection_default_iteration`] answers a plain `Map`/`Set` with its
+/// builtin iterator WITHOUT consulting the symbol tables, so it must not run
+/// once the iteration protocol has been tampered with — a patched
+/// `Map.prototype[Symbol.iterator]`, or an own `@@iterator` installed on one
+/// instance, both have to be observable.
+///
+/// Deliberately coarse: the flag is not keyed to the collection prototypes,
+/// because a plain `Map` has no prototype OBJECT to compare against the way
+/// `Array.prototype` does (its surface is served by redirecting the operation
+/// at each dispatch point). Any `@@iterator` write in the program disables the
+/// lane process-wide. That is the conservative direction — it can only cost
+/// speed, never correctness — and the overwhelming majority of programs never
+/// write `@@iterator` at all, so they keep the lane for free.
+pub static ITERATOR_PROTOCOL_TOUCHED: std::sync::atomic::AtomicBool =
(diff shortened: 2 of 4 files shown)


proggeramlug:
Local merge gate on the Linux build host, head `c8fb0e844` vs base `4b9dde9a4`:

| suite | result |
|---|---|
| perry-runtime (lib) | **2762 passed / 0 failed** — full parallel run, no flakes this time |
| `cargo fmt` | PASS |

A separate serial (`--test-threads=1`) runtime run was also clean at 2762/0.

proggeramlug:

proggeramlug:
Measured on the quiet host, 9 alternating pairs against the merge base, row `4k entities: archetype migration (add/remove) + sync`:

**+0.07%, 5/9 wins** (80.366 → 80.332 ms) — a null. The `for…of` microbenchmark is flat too.

That result is diagnostic rather than merely disappointing. If the lane were firing it would skip the `Symbol.iterator` property lookup *and* the dynamic method call — together ~81% of GetIterator on the profile — so it should have beaten the +14% that #8990 gets from merely making one probe cheaper. Matching noise means **the lane never ran**.

## Why

The `ITERATOR_PROTOCOL_TOUCHED` latch is too coarse to be useful. It flips on any `@@iterator` symbol write anywhere in the process — and the ECS library, like most libraries, simply *defines* `[Symbol.iterator]` as a class method in two places (`utils/multi-map.ts`, `utils/bit-set.ts`). Defining an iterable class is routine and has nothing to do with patching the protocol on a built-in, but it disables the lane process-wide at class-definition time.

I called the latch "deliberately coarse … it can only cost speed, never correctness" in the PR body. That was true, and it cost all of the speed.

## Closing this

The mechanism is unproven: because the lane never executed, I have no evidence it helps even when it does fire. Shipping it would add a latch, two symbol-write hooks and a dispatch arm to a correctness-sensitive path for a measured +0.07%.

A narrower guard is the obvious follow-up — flip only when the write's receiver is a registered Map/Set instance, rather than on every `@@iterator` write — but that needs a real answer for how `Map.prototype[Symbol.iterator] = fn` is modelled before it can be trusted, and it should be re-measured from scratch rather than assumed. Recorded for whoever picks it up.

PR #8992, opened 2026-08-28
fix(runtime): require verified Node-API addon entries

### LINKED ISSUE #8871: node-api-host: assert entry ∈ verified files at load time, and note the hash→dlopen TOCTOU
Two follow-ups from reviewing #8868's addon load path. Neither is a live vulnerability — both are hardening.

### 1. `entry ∈ files` is an implicit invariant, not an asserted one

`verify_addon_payload` (`node_api_host/loader.rs:198`) size- and SHA-256-checks every file in `addon.files`, then returns `safe_payload_path(root, &addon.entry)`. Nothing at the load site asserts that the resolved entry is among the files just verified — so, read in isolation, the file handed to `dlopen` need not be one that was hashed.

It is safe today only because of a distant invariant in the staging step: `native_addon_sidecar.rs::addon_payload_files` walks the package directory, and only walked files are both copied *and* hashed into `files`. An entry that was never hashed was therefore never staged, so `canonicalize` fails and the load errors out.

That is genuinely fail-closed, but the guarantee lives in a different crate from the code that depends on it. A future change to the staging walk — an added copy path, a filter tweak — could break it silently, and the load path would not notice.

**Suggested:** assert at load time that the resolved entry path is present in the verified set, so the load path is self-evidently safe rather than safe-by-construction-elsewhere.

### 2. TOCTOU between hashing and `dlopen`

`verify_addon_payload` reads and hashes the file; `open_library` then opens it again from disk. Between those two reads the bytes can change.

Exploiting this requires local write access to the staged sidecar directory, which is largely outside the threat model this feature targets — so severity is low. It is still a real window, and worth recording rather than rediscovering.

**Options:** hold an open file descriptor across verification and load (`fdlopen` where available), or accept and document it explicitly.

Found while auditing #8868 (landed via #8870). The verify-then-load ordering and the canonical-path traversal check are both correct as written.

proggeramlug:
## Summary

Require a Node-API addon's canonical entry path to match a payload file whose size and SHA-256 were successfully verified before loading it.

## Changes

- Reject manifest entries that resolve to an existing but unlisted payload file.
- Extend the real Node-API loader end-to-end gate with an unverified-entry tampering case.
- Document the accepted hash-to-dynamic-loader TOCTOU boundary and deployment mitigation.

## Related issue

Fixes #8871

## Test plan

- [x] `PERRY_REQUIRE_NODE_API_E2E=1 cargo test -p perry --test node_api_host_e2e real_node_api_addon_resolves_from_host_and_authenticates_sidecar -- --nocapture`
- [x] `cargo test -p perry-runtime --features node-api-host node_api_host:: -- --nocapture`
- [x] Added or updated a test in the affected crate
- [x] Updated `docs/src/`

## Checklist

## Summary by CodeRabbit

* **Security**
  * Node-API sidecar loading now rejects addon entries not included among payload files with successfully verified size and SHA-256 values.
  * Canonical path matching ensures only the verified addon entry is loaded.
  * Clearer errors are provided when an addon entry fails integrity verification.
* **Documentation**
  * Updated sidecar integrity guidance, including verification requirements and limitations involving concurrent file changes.
* **Tests**
  * Added coverage confirming unverified addon entries are rejected before loading.

### DIFF
--- a/changelog.d/8992-node-api-entry-verification.md
+++ b/changelog.d/8992-node-api-entry-verification.md
@@ -1,0 +1,7 @@
+Node-API sidecar loading now rejects an addon entry unless its canonical path
+is one of the payload files whose size and SHA-256 were successfully verified.
+This keeps the runtime authorization boundary self-contained even if sidecar
+staging changes in the future.
+
+The sidecar contract now also documents the accepted pathname-reopen TOCTOU
+window and requires deployments to protect sidecars from concurrent writers.
--- a/crates/perry-runtime/src/node_api_host/loader.rs
+++ b/crates/perry-runtime/src/node_api_host/loader.rs
@@ -202,6 +202,8 @@ fn verify_addon_payload(root: &Path, addon: &ManifestAddon) -> Result<PathBuf, S
             addon.logical_id
         ));
     }
+    let entry = safe_payload_path(root, &addon.entry)?;
+    let mut entry_verified = false;
     for file in &addon.files {
         let path = safe_payload_path(root, &file.path)?;
         let bytes = std::fs::read(&path).map_err(|error| {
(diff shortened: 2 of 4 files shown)


proggeramlug:

Documenting the hash-to-`dlopen` TOCTOU boundary rather than claiming to close it is the right call — a verified hash and a later `dlopen` are inherently two separate observations of the file, and saying so plainly is more useful than an assurance that does not hold. Rejecting a manifest entry that resolves to an existing-but-unlisted payload is the part that actually removes a load path.

Issue #2831, opened 2026-05-30
runtime: preserve Map and Set insertion order after delete


andrewtdiz:
## Summary
Map and Set iteration must preserve insertion order of the remaining entries after deletion. Perry's current delete helpers use swap-remove, moving the last entry into the deleted slot, so subsequent `keys()`, `values()`, `entries()`, `forEach`, `Array.from`, spread, and `for...of`-style lowering can observe reordered collections.

## Node behavior

```js
console.log(Array.from((() => {
  const s = new Set([1, 2, 3, 4]);
  s.delete(2);
  return s;
})()));
// [1, 3, 4]

console.log(Array.from((() => {
  const m = new Map([[1, "a"], [2, "b"], [3, "c"], [4, "d"]]);
  m.delete(2);
  return m.keys();
})()));
// [1, 3, 4]

console.log(Array.from((() => {
  const m = new Map([[1, "a"], [2, "b"]]);
  m.delete(1);
  m.set(1, "z");
  return m.entries();
})()));
// [[2, "b"], [1, "z"]]
```

Updating an existing key keeps its position, while deleting and re-adding appends it at the end.

## Current Perry evidence

- `crates/perry-runtime/src/map.rs::js_map_delete()` captures the last key/value and writes them into the deleted index before decrementing size. The comments call this `swap-and-pop` / `swap-remove`, and the side-table maintenance rewrites the swapped key's stored index to the removed index.
- `crates/perry-runtime/src/set.rs::js_set_delete()` uses the same swap-remove strategy: when the removed value is not the last element, it writes the last element into the removed index and updates that element's index.
- `crates/perry-runtime/src/map.rs::js_map_entries()`, `js_map_keys()`, `js_map_values()`, `js_map_foreach()`, and `crates/perry-runtime/src/set.rs::js_set_to_array()`, `js_set_value_at()`, `js_set_foreach()` all iterate the compacted backing storage from index `0..size`, so the swap becomes user-visible order.
- `crates/perry-runtime/src/set.rs::js_set_value_at()` explicitly documents a direct read of the i-th element in insertion order, but the delete path breaks that invariant.

## Suggested test surface

Add node-suite coverage for:

```ts
console.log("set delete first:", JSON.stringify(Array.from((() => {
  const s = new Set([1, 2, 3]);
  s.delete(1);
  return s;
})())));

console.log("set delete middle:", JSON.stringify(Array.from((() => {
  const s = new Set([1, 2, 3, 4]);
[...]

PR #8993, opened 2026-08-28
perf(runtime): repair the Set index in place after an ordered delete

proggeramlug:
`js_set_delete` did two O(n) things per call, so emptying an N-element Set cost O(N²) **twice over**:

1. it shifted survivors with a full **barriered store per element** rather than one move, producing N barrier entries per delete;
2. it then called `rebuild_set_index`, which **clears the lookup table and re-inserts every surviving element** — a complete re-hash, on every single delete.

For reference, `Map.delete` on the same shape is already ~0.03 µs flat in node regardless of N.

## The change

Exactly what `map::delete_entry_at_index` already does, and has been proven by:

- one overlap-safe `ptr::copy`;
- one `runtime_write_barrier_external_slot_span` over the moved slots;
- an in-place index repair that removes the deleted key and decrements the offsets that sat after it.

The repaired offsets are precisely what a rebuild would have produced — nothing is re-hashed. No new parent → child edge is created (every moved value was already in this Set), so the span records the same old→young contract for the moved slots' new addresses.

Insertion order (#2831) is unchanged: survivors keep their relative order, and a delete-then-re-add still appends at the end.

## Tests

- `ordered_delete_preserves_order_and_repairs_the_index` — deletes from front, middle and back (each moves a different number of slots), then asserts the compacted buffer's order, that the *repaired* index still resolves every survivor and nothing removed, and that a re-add appends at the end.
- `emptying_a_set_leaves_it_consistent` — the quadratic case: removes 64 elements one at a time, checks each deletes exactly once, and that the emptied Set still accepts adds.

Local gate on the Linux build host: **perry-runtime serial 2762 passed / 0 failed**; set suite 199/199. Full gate receipt follows in a comment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Set deletion performance, especially when removing many elements.
  * Fixed internal indexing after ordered deletions to keep Set operations reliable.
  * Ensured Sets remain usable after being emptied and subsequently reused.

* **Tests**
  * Added coverage for deletion order, index consistency, and reuse of emptied Sets.

### DIFF
--- a/changelog.d/8993-set-delete-index-repair.md
+++ b/changelog.d/8993-set-delete-index-repair.md
@@ -1,0 +1,3 @@
+`js_set_delete` repairs the index in place instead of rebuilding it.
+
+Each delete did two O(n) things, making an N-element drain O(N²) twice over: survivors were shifted with a barriered store per element rather than one move (N barrier entries per delete), and `rebuild_set_index` then cleared the lookup table and re-inserted every survivor. The index is now repaired in place.
--- a/crates/perry-runtime/src/set.rs
+++ b/crates/perry-runtime/src/set.rs
@@ -336,6 +336,30 @@ fn rebuild_set_index(set: *mut SetHeader) {
     }
 }
 
+/// Repair `SET_INDEX` after an ordered delete compacted the elements buffer.
+///
+/// Removes the deleted value's entry and decrements every stored offset that
+/// sat after it. Equivalent to rebuilding the table from the compacted buffer
+/// — the offsets are exactly what a rebuild would produce — but it re-hashes
+/// nothing, which is what made the rebuild quadratic when a Set is emptied.
+unsafe fn repair_set_index_after_ordered_delete(
+    set: *mut SetHeader,
+    deleted_value: f64,
+    deleted_idx: u32,
+) {
+    SET_INDEX.with(|idx| {
+        let mut idx = idx.borrow_mut();
+        if let Some(map) = idx.get_mut(&(set as usize)) {
+            map.remove(&JSValueKey(deleted_value));
+            for entry_idx in map.values_mut() {
+                if *entry_idx > deleted_idx {
+                    *entry_idx -= 1;
+                }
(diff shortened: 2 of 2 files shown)


proggeramlug:
Local merge gate on the Linux build host, head `f8852e821` vs base `4b9dde9a4`:

| suite | result |
|---|---|
| perry-runtime (lib) | **2762 passed / 0 failed** — full parallel run, no flakes this time |
| `cargo fmt` | PASS |

A separate serial (`--test-threads=1`) runtime run was also clean at 2762/0.

proggeramlug:

proggeramlug:
Measured on the quiet Apple silicon host.

## `Set.delete` — the operation this changes

Build an N-element Set, delete every element, best of 3:

|---|---:|---:|---:|
| 1,000 | 32.159 µs/delete | 0.800 µs | **40x** |
| 2,000 | 39.330 µs | 1.315 µs | **30x** |
| 4,000 | 43.044 µs | 2.673 µs | **16x** |
| 8,000 | 54.056 µs | 5.355 µs | **10x** |

Emptying an 8,000-element Set goes **432.45 ms → 42.84 ms**.

The remaining growth in per-delete cost is the ordered compaction itself, which is still O(n) — this PR removes the *re-hash* (`rebuild_set_index` cleared the table and re-inserted every survivor on every delete) and the per-element barriered store, not the shift. Node is ~0.027 µs flat at every N, so the gap is far smaller but not closed; a tombstone scheme would be the follow-up that closes it.

## ECS archetype-migration row

9 alternating pairs, control = merge base: **+0.41%, 7/9** (80.320 → 80.156 ms).

Small, and that is expected rather than disappointing — that row's collection cost is dominated by `Map.delete` (`entityToIndex`, one delete per migrated entity) and by generic method dispatch, not by `Set.delete`. The row is not the reason to take this change; the 10-40x on the operation itself is.

Direction is positive and the oracle passed on all 18 runs.

PR #8994, opened 2026-08-28
fix(compile): omit erased names from module namespaces

### LINKED ISSUE #8904: Namespace import object is not materialised: Object.keys(ns) is empty and some members (zod's z.coerce) are undefined, while others work
**perry 0.5.1220**, macOS arm64, zod ^4.0.0 listed in both `perry.compilePackages` and `perry.allow.compilePackages`.

A namespace-style export object comes out partially materialised. Direct access to *some* members works; `Object.keys()` on it returns nothing at all; and at least one member (`z.coerce`) is `undefined` even though it is accessed statically.

## Repro

```ts
import { z } from "zod";

function main(): void {
  console.log("typeof z            :", typeof z);
  console.log("typeof z.object     :", typeof z.object);
  console.log("typeof z.string     :", typeof z.string);
  console.log("typeof z.coerce     :", typeof (z as unknown as { coerce?: unknown }).coerce);
  const keys = Object.keys(z as unknown as Record<string, unknown>);
  console.log("z key count         :", keys.length);
  console.log("'coerce' in z       :", "coerce" in (z as unknown as object));
  console.log("descriptor          :", Object.getOwnPropertyDescriptor(z as unknown as object, "coerce"));
}
main();
```

| | perry 0.5.1220 | node 26 |
|---|---|---|
| `typeof z` | `object` | `object` |
| `typeof z.object` | `function` | `function` |
| `typeof z.string` | `function` | `function` |
| `typeof z.coerce` | **`undefined`** | `object` |
| `Object.keys(z).length` | **`0`** | `238` |
| `"coerce" in z` | `false` | `true` |
| `getOwnPropertyDescriptor(z, "coerce")` | `undefined` | `{value, writable, enumerable, configurable}` |

## Where it bites

```ts
const Nested = z.object({ rows: z.array(z.object({ n: z.coerce.number() })).min(1) });
```

→ `TypeError: Cannot read properties of undefined (reading 'number')`

That is a plain, statically-visible `z.coerce.number()` — not dynamic access — so whatever decides which members to retain does not appear to follow nested member expressions through the namespace.

The `Object.keys(z).length === 0` result is the part I would point at first: the object exists and answers some property reads, but has no own enumerable properties at all. Any library that iterates its own namespace (or does `"x" in ns` feature detection) would take a wrong branch silently rather than throwing.

## Why it took a while to see

The failure is late and looks like a library bug: six earlier assertions in the same file pass — `z.object`, `z.array`, `z.string`, discriminated unions, `.transform`, `z.enum` — so zod is clearly compiled and working. Only `coerce` is missing, and the resulting `TypeError` names `number`, which points at the wrong place.

## Environment

- zod ^4.0.0, listed in `perry.compilePackages` and `perry.allow.compilePackages`

proggeramlug:
## Summary

- derive module namespace keys from consumer-visible export names so private backing functions such as `_null` do not leak beside their public aliases
- exclude erased TypeScript interfaces and type aliases when materializing dynamic and nested namespace objects
- add a Zod-shaped regression for `import * as z; export { z }`, nested `export * as coerce`, reflection, alias-only exports, and type erasure

## Verification

- `cargo build --release -p perry-runtime-static -p perry-stdlib-static`
- `./run_parity_tests.sh --filter namespace_type_erasure_8904`
- `./run_parity_tests.sh --filter export_star_as_namespace_object`
- `./run_parity_tests.sh --filter namespace_import_enumerable`
- real Zod 4.3.5 no-cache compile: `Object.keys(z).length === 236` in both Node and Perry, `z.coerce` is owned/enumerable, and `z.coerce.number()("42") === 42`

No version bump.

Fixes #8904

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed namespace exports so type-only declarations are excluded from runtime namespace objects.
  * Prevented private backing names from appearing as public exports.
  * Ensured nested and dynamically imported namespaces expose only valid runtime values.

* **Tests**
  * Added coverage for namespace re-exports, enumerable properties, runtime keys, and numeric coercion behavior.

### DIFF
--- a/changelog.d/8994-namespace-erased-names.md
+++ b/changelog.d/8994-namespace-erased-names.md
@@ -1,0 +1,3 @@
+Module namespaces omit erased and non-consumer-visible names.
+
+Namespace keys are derived from consumer-visible export names, so a private backing function such as `_null` no longer appears beside its public alias, and erased TypeScript interfaces and type aliases are excluded when materializing dynamic and nested namespace objects. Includes a Zod-shaped regression for `import * as z`.
--- a/crates/perry/src/commands/compile/run_pipeline.rs
+++ b/crates/perry/src/commands/compile/run_pipeline.rs
@@ -67,6 +67,45 @@ fn configured_module_codegen_jobs(total_modules: usize) -> (usize, usize) {
     (module_jobs, llvm_unit_jobs)
 }
 
+/// Whether `name` is metadata for an erased TypeScript declaration rather
+/// than a JavaScript binding in `module`.
+///
+/// Interfaces and type aliases remain in `Module::exports` so cross-module
+/// type resolution can follow re-export chains. Runtime namespace builders
+/// must filter them back out, while preserving declaration-merged names that
+/// also have a value binding.
+fn is_type_only_export_binding(module: &perry_hir::Module, name: &str) -> bool {
+    let has_type_declaration = module
+        .interfaces
+        .iter()
+        .any(|interface| interface.name == name)
+        || module.type_aliases.iter().any(|alias| alias.name == name);
+    if !has_type_declaration {
+        return false;
+    }
+
+    let has_runtime_value = module
+        .functions
+        .iter()
+        .any(|function| function.name == name)
+        || module
+            .exported_functions
+            .iter()
+            .any(|(exported, _)| exported == name)
+        || module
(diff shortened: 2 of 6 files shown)


proggeramlug:

PR #8995, opened 2026-08-28
test: cover compilePackages RegExp dispatch on macOS

### LINKED ISSUE #8905: A RegExp handed to a compilePackages package loses its prototype: z.string().regex() dies with 'test is not a function'
**perry 0.5.1220**, macOS arm64, zod ^4.0.0 in `perry.compilePackages` + `perry.allow.compilePackages`.

`RegExp.prototype.test` works perfectly in user code. But a `RegExp` passed *into* a compiled package and called there is missing it.

## Repro

```ts
import { z } from "zod";

function main(): void {
  // Plain user code: all fine.
  const re = /^a+$/;
  console.log("user .test      :", re.test("aaa"));          // true
  console.log("typeof re.test  :", typeof re.test);          // "function"
  console.log("via fn / obj    :", ((r: RegExp) => r.test("aaa"))(re), ({ r: re }).r.test("aaa"));

  // The same regex handed to zod:
  console.log(z.string().regex(/^[A-Za-z0-9_-]{12}$/).safeParse("abcdefgh1234"));
}
main();
```

```
user .test      : true
typeof re.test  : function
via fn / obj    : true true
TypeError: test is not a function
```

Node runs the last line fine (`{ success: true, … }`).

`z.string().datetime()` fails identically, which fits — zod validates ISO strings with an internal regex.

## Probably related, possibly the same root cause

Two other things in the same build look like values crossing the package boundary arriving partially materialised:

- [#8904](https://github.com/PerryTS/perry/issues/8904) — the `zod` namespace object has **no own enumerable properties** (`Object.keys(z).length === 0` vs 238 on Node) and `z.coerce` / `z.iso` are `undefined`, while `z.object` and `z.string` work.
- `ZodObject.partial()` throws `Error: Invalid element at key "id": expected a Zod schema` — i.e. zod rejects the schemas in its *own* `.shape` when it re-reads them. Reading `.shape` from the outside is fine: `Object.keys(schema.shape)` returns `["id","n"]` and `schema.shape.id.optional()` works.

So the pattern looks like: an object is usable across the boundary for direct member reads that the compiler saw statically, but loses its prototype and its own-property table. Filing separately because the symptom is distinct enough to bisect on its own; close as duplicate if it is one fix.

## Impact

This is the whole input-validation layer of a REST API. `z.string().regex()` guards id shapes and `z.string().email()` guards addresses; both raise `TypeError` instead of returning a validation failure, so the process throws on the first malformed request rather than answering 400.

## Environment

proggeramlug:
## Summary

Closes the RegExp prototype regression reported against v0.5.1220 by locking in the low-address macOS brand-check behavior already corrected on main and exercising the original compilePackages boundary.

## Changes

- Add a zod-like compiled dependency that captures def.pattern and calls its RegExp methods.
- Force the full-stdlib, no-auto-optimize link so the test covers the cross-runtime fallback path.
- Tie the deterministic low-macOS-address guard to the RegExp dispatch failure from #8905.
- Add the required per-PR changelog fragment without changing version metadata.

## Related issue

Closes #8905

## Test plan

- [ ] cargo build --release clean (not run; focused remote checks used)
- [ ] Full workspace test suite (not run; focused remote checks used)
- [x] CARGO_TARGET_DIR=/root/perry-target-8905 cargo test -p perry --test issue_8905_regexp_package_boundary -- --nocapture
- [x] CARGO_TARGET_DIR=/root/perry-target-8905 cargo test -p perry-runtime --lib regex
- [x] python3 scripts/check_test_registration.py
- [x] ./scripts/pre-tag-check.sh --quick
- [x] Exact zod 4.0.0 regex and datetime reproducer compiled and ran successfully on perrymaster
- [x] Added issue-specific integration and runtime regression coverage
- [x] Documentation update not required; no CLI or public API changed

## Screenshots / output

The integration fixture prints function true false and exits successfully.

## Checklist

- [x] My commits follow the repository prefix convention

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where compiled packages could lose regular expression method behavior across package boundaries.
  * Regular expressions created by application code now preserve expected behavior, including `lastIndex` handling and `test()` results.
  * Improved compatibility for schema validation scenarios involving regular expression and datetime checks, including on macOS systems.

### DIFF
--- a/changelog.d/8995-compiled-package-regexp.md
+++ b/changelog.d/8995-compiled-package-regexp.md
@@ -1,0 +1,4 @@
+Compiled packages retain RegExp method behavior when they receive a regular
+expression created by application code, including on macOS allocations below
+2 TB. This covers schema-library paths such as zod regex and datetime checks
+when compiling with the full prebuilt stdlib.
--- a/crates/perry-runtime/src/value/addr_class.rs
+++ b/crates/perry-runtime/src/value/addr_class.rs
@@ -451,8 +451,12 @@ mod tests {
     fn macos_accepts_heap_addresses_below_two_tb() {
         // The Rust test harness has observed mimalloc allocations around
         // 45 GB. Classification is purely numeric and must not dereference
-        // this representative address.
-        assert!(is_valid_obj_ptr(0x0000_000a_0000_0000usize as *const u8));
+        // this representative address. In #8905 this gate was reached by the
+        // RegExp header-brand fallback across prebuilt-stdlib runtime copies;
+        // rejecting the address made dependency-side `.test()` dispatch miss.
+        let low_macos_heap_addr = 0x0000_000a_0000_0000usize;
+        assert!(is_valid_obj_ptr(low_macos_heap_addr as *const u8));
+        assert!(is_plausible_heap_addr(low_macos_heap_addr));
     }
 
     #[cfg(all(
--- a/crates/perry/tests/issue_8905_regexp_package_boundary.rs
+++ b/crates/perry/tests/issue_8905_regexp_package_boundary.rs
@@ -1,0 +1,153 @@
+//! Regression test for #8905: a RegExp passed into a `compilePackages`
+//! dependency must retain RegExp method behavior when the dependency reads it
+//! back through an object property.
+
+use std::path::PathBuf;
+use std::process::Command;
+use std::sync::Once;
+
+fn perry_bin() -> PathBuf {
+    PathBuf::from(env!("CARGO_BIN_EXE_perry"))
+}
+
+fn workspace_root() -> PathBuf {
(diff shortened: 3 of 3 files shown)


proggeramlug:
Merged (batched with #8996 and #8997). Locking in the low-address macOS brand-check behaviour and exercising the original `compilePackages` boundary is the right way to close a regression that was already fixed on `main` — a fix without a test that would have caught it is a fix that can come back.

**Validation** — runtime 2787/0, codegen 1341/0, stdlib 124/0 (`RUST_TEST_THREADS=1`); `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green — the exception is the pre-existing Actions-expression artifact (#8929).

PR #8997, opened 2026-08-29
fix(intl): apply locale and timezone data

### LINKED ISSUE #8903: Intl.DateTimeFormat and Intl.Collator silently fall back to en-US/root while resolvedOptions() reports de-DE; timeZone is ignored
**perry 0.5.1220**, macOS arm64.

`Intl.NumberFormat` is correct — byte-identical to Node for `de-DE` currency and percent. `Intl.DateTimeFormat` and `Intl.Collator` are not, and the divergence is **silent**: `resolvedOptions().locale` returns `de-DE` in both cases while the output is en-US / codepoint order. `timeZone` appears to be ignored entirely rather than mis-formatted.

## Repro

```ts
function cp(s: string): string {
  return [...s].map((c) => c.codePointAt(0)!.toString(16).padStart(4, "0")).join(" ");
}

function main(): void {
  console.log("resolved NumberFormat  :", new Intl.NumberFormat("de-DE").resolvedOptions().locale);
  console.log("resolved DateTimeFormat:", new Intl.DateTimeFormat("de-DE").resolvedOptions().locale);

  const d = new Date("2026-09-07T06:05:00Z");
  console.log("currency:", new Intl.NumberFormat("de-DE", { style: "currency", currency: "EUR" }).format(1234.56));
  console.log("date    :", new Intl.DateTimeFormat("de-DE", { dateStyle: "medium" }).format(d));
  console.log("weekday :", new Intl.DateTimeFormat("de-DE", { weekday: "long" }).format(d));
  console.log("berlin  :", new Intl.DateTimeFormat("de-DE", { timeStyle: "short", timeZone: "Europe/Berlin" }).format(d));
  console.log("collator:", JSON.stringify(["Zubehör", "Ärger", "Apfel"].sort(new Intl.Collator("de-DE").compare)));
}
main();
```

| | Node 26 | perry 0.5.1220 |
|---|---|---|
| `resolvedOptions().locale` (both formatters) | `de-DE` | `de-DE` |
| currency | `1.234,56 €` | `1.234,56 €` ✅ identical, U+00A0 before the symbol |
| percent | `7,0 %` | `7,0 %` ✅ |
| `dateStyle: "medium"` | `07.09.2026` | `Sep 7, 2026` ❌ |
| `weekday: "long"` | `Montag` | `Monday` ❌ |
| `timeStyle: "short"`, `timeZone: "Europe/Berlin"` | `08:05` | `6:05 AM` ❌ — that is UTC, so the `timeZone` was not applied at all |
| `Collator("de-DE")` | `Apfel, Ärger, Zubehör` | `Apfel, Zubehör, Ärger` ❌ codepoint order |

## Why the silence is the problem

`resolvedOptions().locale` reporting `de-DE` is the natural thing to check when you want to know whether a build has usable ICU data for a locale — and here it answers yes for a formatter that then produces en-US. A feature-detect written against it cannot tell the two cases apart.

That NumberFormat is *correct* makes it worse in practice: money looks right, so nothing prompts you to check the dates on the same page.

Our own case is a German invoicing platform where dates land on invoices, dunning notices and a legally-meaningful suspension deadline. `Sep 7, 2026` on a German invoice is a document defect rather than a cosmetic one, and nothing would have thrown.

If full ICU is out of scope for the runtime, a `resolvedOptions().locale` that honestly reports the fallback (`en-US`) would at least make it detectable.

## Environment

proggeramlug:
Fixes #8903

## Summary

- resolve and validate named DateTimeFormat zones with the compiled IANA database, including DST
- route weekday-only component formats through the correct ICU date fieldset
- retain Unicode normalization in auto-optimized Collator builds
- add a Node-parity regression covering de-DE date, weekday, Berlin time, and collation

## Testing

- `cargo test --release -p perry-runtime name_bearing_components_match_node -- --nocapture`
- `cargo test --release -p perry-runtime compiled_tzdb_resolves_named_zone_and_dst -- --nocapture`
- `cargo test --release -p perry string_normalization_gate_covers_normalize_and_locale_compare -- --nocapture`
- compiled and ran `test_gap_issue_8903_intl_locale_data.ts` with Node 26.8.1, full `PERRY_NO_AUTO_OPTIMIZE=1`, and auto-optimized Perry; all outputs were byte-identical

No version bump.

## Summary by CodeRabbit

* **Bug Fixes**
  * `Intl.DateTimeFormat` now supports named IANA time zones with accurate daylight-saving transitions.
  * Weekday-only date formats now use the requested locale’s language data.
  * `Intl.Collator` preserves locale-aware comparison and Unicode normalization in optimized builds.
  * Regular expression behavior is preserved for application-created expressions in compiled packages.

* **Performance**
  * Array subclass `push()` operations can use a faster inline path when applicable.

* **Tests**
  * Added coverage for German localization, Berlin time-zone formatting, daylight saving, and collation behavior.


proggeramlug:
Merged (batched with #8995 and #8996).

I checked the timezone work behaviourally rather than trusting the unit tests, since "resolve named zones from the compiled IANA database, including DST" is the sort of claim that passes a fixed-offset implementation right up until it does not:

```
America/New_York, 2026-07-15T12:00Z   perry=08:00   node=08:00
Asia/Tokyo,       2026-07-15T12:00Z   perry=21:00   node=21:00
```

`08:00` is the one that matters — that is EDT, so the offset is genuinely coming from DST rules rather than a stored standard offset, which a July date would otherwise render as `07:00`.

**One fix pushed:** the fragment was named `8903-intl-locale-timezone.md`. #8903 is the issue this fixes, not the PR, so the change would have been filed under an issue number at release time. Renamed to `8997-`.

**Validation** — runtime 2787/0, codegen 1341/0, stdlib 124/0 (`RUST_TEST_THREADS=1`); `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green — the exception is the pre-existing Actions-expression artifact (#8929).

Issue #2856, opened 2026-05-30
runtime: return real Map and Set iterator objects


andrewtdiz:
## Summary
Perry's Map/Set `entries()`, `keys()`, and `values()` paths materialize arrays for common `Array.from(...)` / `for...of` use, but Node returns iterator objects. The returned values should not be Arrays, should expose `.next()`, should be iterable via `Symbol.iterator`, and should be recognized by `util.types.isMapIterator()` / `isSetIterator()` because of their actual iterator shape rather than a side-table marker.

`Set.prototype.entries()` also appears to be missing on the typed Set method path.

## Node behavior
On Node v25.9.0:

```js
const util = require("node:util");
const m = new Map([[1, "a"], [2, "b"]]);
const s = new Set(["x", "y"]);

for (const [label, it] of [
  ["map keys", m.keys()],
  ["map entries", m.entries()],
  ["set values", s.values()],
  ["set entries", s.entries()],
]) {
  console.log(label, Array.isArray(it), typeof it.next, util.types.isMapIterator(it), util.types.isSetIterator(it));
  console.log(label, JSON.stringify(it.next()), JSON.stringify(it.next()));
}
console.log("map default:", m[Symbol.iterator]().next().value.join(":"));
console.log("set default:", s[Symbol.iterator]().next().value);
```

Expected properties:

- `Array.isArray(it) === false`
- `typeof it.next === "function"`
- Map iterators satisfy `util.types.isMapIterator`
- Set iterators satisfy `util.types.isSetIterator`
- `.next()` advances one entry at a time and returns `{ value, done }`
- `Map.prototype[Symbol.iterator]` aliases entries, and `Set.prototype[Symbol.iterator]` aliases values

## Current Perry evidence
Current source materializes arrays instead of iterator objects:

- `crates/perry-hir/src/ir/expr.rs` documents `MapEntries`, `MapKeys`, and `MapValues` as returning `Array<[key, value]>`, `Array<key>`, and `Array<value>`, and `SetValues` as returning an `Array`.
- `crates/perry-runtime/src/map.rs::js_map_entries()` allocates an outer Array of `[key, value]` pair Arrays, then calls `mark_map_iterator_array(result)` before returning it. `js_map_keys()` and `js_map_values()` likewise allocate Arrays and mark them as map iterators.
- `crates/perry-runtime/src/set.rs::js_set_to_array()` materializes a Set as an Array and calls `mark_set_iterator_array(result)`.
[...]

PR #8998, opened 2026-08-29
fix(runtime): narrow the collection-iterator latch to Map/Set receivers

proggeramlug:
Follow-up to #8991, which I measured **after** it merged: as shipped it is **+0.07%, 5/9** on the ECS archetype-migration row — a null, because **the lane never runs**.

## Why it never runs

The `ITERATOR_PROTOCOL_TOUCHED` latch flips on *any* `@@iterator` symbol write anywhere in the process. A class that merely **defines** `[Symbol.iterator]` as a method trips it — an ordinary iterable class, and the ECS corpus has two (`utils/multi-map.ts`, `utils/bit-set.ts`). So the latch fires at class-definition time and disables the lane process-wide, in essentially any real program.

I described the latch as "deliberately coarse … it can only cost speed, never correctness" in #8991. That was accurate, and it cost all of the speed.

## The fix, and why it is sound

Flip only when the write's **receiver** is a registered `Map` or `Set`.

Receiver-narrowing is safe here for a specific reason: **perry models no `Map.prototype` object at all.** `symbol::get` emulates `Map.prototype[Symbol.iterator]` by binding `entries`, and `Set`'s by binding `values`, inside the resolver itself (#2856). There is therefore nothing to patch on a prototype, and the only channel that can override a plain collection's default iteration is an **own** symbol property on that instance — which is exactly what this hook observes.

A Map/Set *subclass* instance is an ordinary object rather than a registered collection, so the lane already declines it and the subclass arm decides; that is unchanged.

## Tests

The added assertion is the regression itself: defining `@@iterator` on a non-collection must leave the lane armed, and a plain `Map` must still be claimed afterwards. The existing assertions still pin that an own `@@iterator` write **on a Map** does disarm it, and that `@@asyncIterator` does not.

Local gate on the Linux build host: **perry-runtime serial 2766 passed / 0 failed**; `iterator` 32/32, `symbol` 49/49, lane tests 2/2.

Paired measurement on the quiet host follows in a comment. If the narrowed lane still does not move the row, the right outcome is to revert #8991 rather than keep inert machinery on a correctness-sensitive path — I will open that revert myself if so.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed collection iteration behavior so `[Symbol.iterator]` changes on unrelated classes no longer affect Map and Set iteration.
  - Map and Set iteration now responds only to iterator changes made directly on registered Map or Set subclasses.
  - Improved default collection iteration in applications that define custom iterators elsewhere.
- **Tests**
  - Added regression coverage to prevent unrelated iterator definitions from disabling default collection iteration.
- **Documentation**
  - Documented the updated iterator behavior.

### DIFF
--- a/changelog.d/8998-narrow-iterator-latch.md
+++ b/changelog.d/8998-narrow-iterator-latch.md
@@ -1,0 +1,5 @@
+The collection-iterator latch is keyed to Map/Set receivers instead of any `@@iterator` write.
+
+`ITERATOR_PROTOCOL_TOUCHED` previously flipped on any `@@iterator` symbol write anywhere in the process, so an ordinary class merely *defining* `[Symbol.iterator]` disabled the plain-collection iteration lane (#8991) process-wide — which is to say the lane did not run in any realistic program.
+
+Narrowing to registered Map/Set receivers is sound because perry models no `Map.prototype` object to patch: `symbol::get` emulates `Map.prototype[Symbol.iterator]` by binding (#2856), so an own write on the instance is the only override path.
--- a/crates/perry-runtime/src/object/map_set_subclass.rs
+++ b/crates/perry-runtime/src/object/map_set_subclass.rs
@@ -84,8 +84,8 @@ unsafe fn instance_object_ptr(this: f64) -> Option<*mut ObjectHeader> {
 /// hidden backing field), return its backing collection. Returns `None` for
 /// real Maps/Sets, ordinary objects, and non-objects — so callers fall through
 /// to their existing handling.
-/// Set once any `[Symbol.iterator]` symbol-keyed write or delete is observed,
-/// anywhere in the process.
+/// Set once an own `[Symbol.iterator]` write or delete is observed on a
+/// registered `Map` or `Set`.
 ///
 /// [`plain_collection_default_iteration`] answers a plain `Map`/`Set` with its
 /// builtin iterator WITHOUT consulting the symbol tables, so it must not run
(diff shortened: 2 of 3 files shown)


proggeramlug:
Local merge gate on the Linux build host, head `2af0e9fb3` vs base `94947aeed`:

| suite | result |
|---|---|
| perry-runtime (lib) | **serial 2766 passed / 0 failed** |
| `cargo fmt` | PASS |
| ratchets vs merge base | unrooted-local 576 → 576, no ceiling raised |

The parallel run showed one failure, `gc::roots::stack_maps::decode_tests::tests::discovers_a_map_from_a_later_loaded_shared_object` (2765/1). That is the known `dlopen` + image-walk flake — it fails in parallel runs on pristine `origin/main` too, and this change touches neither GC roots nor stack maps (the diff is one symbol-write hook and its two call sites). The serial run above is clean at 2766/0.

Measurement on the quiet host is queued behind the other candidates and follows here.

proggeramlug:
Merged. Measuring your own change after it merged and reporting it as a null is the useful half of this — a lane that never runs is worse than no lane, because it looks like coverage. Worth noting I merged #8991 without measuring it and said as much in that comment; this is the correction I could not have made myself.

**On the narrowing, which is the part that could go wrong.** Keying the latch to the receiver means an override the latch no longer sees would leave the lane serving a stale builtin iterator. The obvious candidate is a patched `Map.prototype[Symbol.iterator]` — the prototype is not a registered Map, so that write would not latch. The code answers it directly:

> Receiver-narrowing is sound here because perry models no `Map.prototype` OBJECT: `symbol::get` emulates `Map.prototype[Symbol.iterator]` by binding (#2856). There is therefore nothing to patch on a prototype.

So an own write on the instance is the only override path, and that still latches. The soundness rests on that modelling choice, so it is worth the comment being right there — if perry ever grows a real `Map.prototype` object, this narrowing stops being safe and the comment is what will say so.

The test pair is the right one: an own `@@iterator` write on a Map **must** flip the latch, and defining `[Symbol.iterator]` on a non-collection **must** leave the lane armed. Testing only the second would have let an over-narrowing through.

**One fix pushed:** this PR touches two `crates/` files with no changelog fragment and no `skip-changelog` label, so the changeset gate would have rejected it. Added `changelog.d/8998-narrow-iterator-latch.md`.

**Validation** — runtime 2787/0 including the 6 `map_set_subclass` latch tests (`RUST_TEST_THREADS=1`); `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green — the exception is the pre-existing Actions-expression artifact (#8929). I have not re-measured the ECS row.

proggeramlug:
Measured on the quiet Apple silicon host — 9 alternating pairs, control = merge base `94947aeed` (i.e. current main, with #8991's lane present but never firing).

**Row `4k entities: archetype migration (add/remove) + sync`:**

|---|---|---|---|---|---|
| 9-pair screen | 68.664 ms | 50.872 ms | **+25.85%** | **9/9** | 18/18 |

This confirms the diagnosis: #8991's lane was dead code, and it is worth **+25.85%** once it actually runs.

## The row across all three changes

| state | row | vs node (4.32 ms) |
|---|---:|---:|
| before any of this | 80.22 ms | 18.6x |
| + #8990 (probe gated on the GC header) | 68.85 ms | 15.9x |
| + this PR (lane actually fires) | **50.87 ms** | **11.8x** |

**80.22 → 50.87 ms, −36.6%** on what the cluster measurement showed was perry's worst ECS row by a factor of five.

## A note on the microbenchmark

`for…of` over a 4-element Set is unchanged (85.6 → 84.6 ns/loop), as it was for #8990. That is the same calibration I recorded there: a concrete `Set<number>` receiver gets a specialized for-of lowering that never reaches `js_get_iterator`, so neither change is on its path. Both wins belong to **generic (type-erased) iteration** — `ReadonlyMap`/`ReadonlySet`-typed parameters, `any`, and interface-typed receivers — which is what the ECS library actually uses.

So #8991 should stay, and this is the change that makes it earn its place.

PR #8999, opened 2026-08-29
fix(release): guard dynamic receiver probes by GC kind

proggeramlug:
## Summary

- guard the Array-subclass `ObjectMeta.elements` probe with the authoritative `GC_TYPE_OBJECT` kind before dereferencing receiver layout
- reject coincidental `CLOSURE_MAGIC` bytes in managed non-closure cells (the release parity Error-property failure)
- remove the now-stale Linux allowlist entry for `test_class_field_layout`
- repair the `Cargo.lock` dependency edge for the timezone provider added by #8997

## Validation

- `cargo test -p perry-codegen index_get_claim --lib` (12 passed)
- `cargo test -p perry-runtime managed_error_with_closure_magic_in_padding_is_not_a_closure --lib`
- `python3 scripts/parity_known_failures.py --audit`
- Linux release parity: `test_gap_gc_catch_param_rooting` (1/1 passed)
- Linux release parity: `test_class_field_layout` (1/1 passed)

The release tag/version remains intentionally untouched; it will be created only after this lands and the exact final main SHA passes the required release gates.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed crashes when accessing indexed properties on arrays, strings, errors, buffers, and other specialized values.
  * Corrected property lookup for error objects that could lose their message or custom fields.
  * Improved handling of forwarded managed objects during property access.

* **Tests**
  * Removed a Linux parity exception for class field layout, enabling regression detection.
  * Updated coverage for indexed access and string concatenation behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6348,6 +6348,7 @@ dependencies = [
  "taffy",
  "temporal_rs",
  "thiserror 1.0.69",
+ "timezone_provider",
  "unicode-normalization",
  "unicode-segmentation",
  "url",
--- a/changelog.d/8999-release-receiver-probe-kinds.md
+++ b/changelog.d/8999-release-receiver-probe-kinds.md
@@ -1,0 +1,5 @@
+### Fixed
+
+- Dynamic indexed reads now inspect `ObjectMeta.elements` only for receivers whose GC kind is an object. Array, string, Error, and other layouts can no longer be interpreted as an object metadata pointer, eliminating the release parity crashes introduced with Array-subclass elements storage.
+- Managed heap cells are classified as closures only when their authoritative GC kind is `GC_TYPE_CLOSURE`. Reused `Error` storage whose padding retained the closure magic marker no longer loses `.message` or custom fields during property lookup.
+- Removed the stale Linux parity allowance for `test_class_field_layout`, which now matches Node, and recorded the timezone-provider dependency added to the runtime in `Cargo.lock`.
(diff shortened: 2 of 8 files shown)


proggeramlug:
Merged.

**I verified the `Cargo.lock` claim rather than taking it on faith, and it was real:** `cargo metadata --locked` fails on pristine `main` — #8997 added the timezone-provider dependency without updating the lock. This PR repairs it (`--locked` passes here).

Worth saying plainly: I merged #8997 last night and did not catch that. Nothing I run would have — `run_lint_gates.sh` has no `--locked` step, and `cargo check`/`cargo test` silently *update* the lock rather than failing on it, so every gate stayed green while a release build using `--locked` would not have. I have added a `--locked` resolve to my own pre-merge checks.

Rejecting coincidental `CLOSURE_MAGIC` bytes in managed non-closure cells is the more interesting half. An `ErrorHeader`'s bytes 12..16 are padding on 64-bit targets, so recycled arena storage can retain a stale magic there and a bytes-first check would read a live Error as a closure — the same shape as reading a non-string as a string. Letting the authoritative GC kind outrank look-alike bytes is the right ordering.

**One fix pushed:** the new `managed_error_with_closure_magic_in_padding_is_not_a_closure` test does a raw `write_unaligned`, which `gc_store_site_inventory` requires be classified. Added `GC_STORE_AUDIT(POINTER_FREE)` — it stores a `u32` constant into padding deliberately, so there is no heap pointer for a barrier to track.

**Validation** — runtime 2788/0, codegen 1341/0 (`RUST_TEST_THREADS=1`); `cargo metadata --locked` ok; `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green — the exception is the pre-existing Actions-expression artifact (#8929).

PR #9000, opened 2026-08-29
perf(runtime): stop hashing a hash in the shape key index; memcpy the delete key clone

proggeramlug:
Two costs on the populated-delete path — perry's worst object-model gap against node (`bench_populated_delete.ts`, currently ~280×).

**1. `ShapeIndex::slots` was hashing a hash.** It is a `HashMap<u64, Vec<u32>>` whose key is ALREADY an FNV-1a content hash of the key bytes, but it used the std default hasher — so every probe ran SipHash over a well-distributed hash, buying no distribution and costing real time. In that benchmark `hash_one::<&usize>` (10.7%) plus `sip::Hasher::write` (4.0%) came to **14.7% of self time**, second only to `shape_slot_lookup` — the function performing the probes. It now uses the `PtrHasher` the runtime's other pointer/id-keyed registries already use.

**2. The keys-array clone is now a memcpy.** Every delete allocates a fresh keys array and copied the surviving keys one `f64` at a time. Two `copy_nonoverlapping` runs instead. This is safe by the code's own existing argument: the destination is a freshly allocated, still-**unpublished** array whose layout is rebuilt before publish — which is precisely why the per-element writes carried no GC barrier either — and source and destination are distinct allocations, so they cannot overlap.

### Measurement — and why it is smaller than the profile suggests

Interleaved A/B, min-of-15, quiet load (~1.4), built from the exact parent commit and this commit in one run:

| | base | this PR | |
|---|---|---|---|
| `bench_populated_delete` | 5938 ms | **5847 ms** | −1.5% min, −2.2% mean |
| combined overwrite | 40 ms | 39 ms | −2% |
| realistic-name read | 15 ms | 15 ms | unchanged |

A 14.7% profile share turning into ~2% wall clock is worth stating plainly rather than rounding up: the symbolized build used for profiling inlines differently, and more importantly this path is dominated by **allocator and page churn**, not hashing — `clear_page_erms` 5.6%, `mi_free` 4.2%, `RawVecInner::finish_grow` 2.9%. The shape index is rebuilt on every delete and each of its ~500 buckets is a separate `Vec<u32>` allocation. That is the real target and this PR does not touch it.

### Correctness

- Adversarial property differential — delete-then-reread under a cached slot, an accessor defined over a cached data slot, prototype fallback after the own property is deleted, `Object.freeze`, the same keys in opposite orders, and a 300-key object in the overflow store — **byte-identical to node**.

## Summary by CodeRabbit

* **Performance**
  * Improved object property deletion performance, especially for objects with many existing properties.
  * Reduced overhead when managing object shape and property lookup information.
  * Optimized internal key movement during deletion to minimize unnecessary work.

* **Reliability**
  * Added benchmark and differential testing coverage to verify behavior and performance improvements.


proggeramlug:
Merged. Hashing an already-well-distributed FNV-1a hash through SipHash is a clean find — the second hash buys no distribution and costs a real probe, and it is the kind of thing that survives review precisely because `HashMap<u64, _>` looks unremarkable.

**One fix pushed** besides the fragment: the `copy_nonoverlapping` replacing the scalar loop needed a `GC_STORE_AUDIT` marker on **both** runs. The first had one; `gc_store_site_inventory` matches within roughly a six-line window, so the second copy was outside it and the gate was red. Added the same `INIT` justification there — freshly allocated, still-unpublished keys array, layout rebuilt before `set_object_keys_array` publishes it, distinct allocations so the copies cannot overlap.

I checked that classification rather than copying it: `INIT` is right only because the destination is unpublished at the time of the write, which is also why the per-element loop it replaces carried no barrier. Had the destination been live, the same memcpy would have needed barrier entries and `INIT` would have hidden that.

**Also fixed:** the fragment was named `8998-delete-path-hasher-memcpy.md`. #8998 is a different PR, merged an hour earlier. Renamed to `9000-`.

**Validation** — runtime 2788/0, codegen 1341/0 (`RUST_TEST_THREADS=1`); `scripts/run_lint_gates.sh` 57 of 58 with the compile tier green after the marker — the exception is the pre-existing Actions-expression artifact (#8929). The ~280× gap figure is not re-measured here.

PR #9002, opened 2026-08-29
perf(runtime): carry the key index across a delete instead of rebuilding it (−7.6%)

proggeramlug:
Third in the populated-delete series, after #9000 and #9001. This one removes the rebuild rather than making it cheaper.

`delete obj[k]` clones the keys array. The clone has a **new address**, so it misses `indices` and the whole index is rebuilt — decoding and FNV-hashing all ~500 surviving property names, on every single delete of a 500-key object.

The survivors are the same strings in the same order minus one, so the index can be **shifted instead of recomputed**: drop the removed slot, decrement every slot above it, touch no key bytes at all.

### Safety

By construction, not by argument: `shape_slot_lookup` already re-validates the stored key against the requested bytes before returning a slot. An index that is wrong therefore produces a **miss** and the caller's existing fallback — it cannot resolve to a different property. Only a fully-built index is migrated; a partially built one is dropped exactly as before, since shifting it would misalign the un-indexed tail.

### Measurement

Interleaved A/B, min-of-15, quiet load (~1.0), base and branch built from exact SHAs in one run:

| | base | this PR | |
|---|---|---|---|
| `bench_populated_delete` | 5023 ms | **4641 ms** | **−7.6%** (mean −7.4%) |

Cumulative across #9000 + #9001 + this: **5938 → 4641 ms, −21.8%** on perry's worst object-model gap.

### Correctness

- Adversarial property differential — delete-then-reread under a cached slot, accessor over a cached data slot, prototype fallback after delete, `Object.freeze`, same keys in opposite orders, 300-key object in the overflow store — **byte-identical to node**. The delete-heavy cases in that fixture are the ones this PR could plausibly break, and they pass.

### Remaining on this path

The per-delete value shift still moves ~499 values one at a time through barriered setters, and the keys-array clone still allocates per delete. The shift is the larger of the two and needs a bulk write-barrier story before it can become a memmove.

## Summary by CodeRabbit

* **Performance**
  * Improved property deletion speed by avoiding unnecessary reprocessing of remaining property names.
  * Reduced benchmark time for populated-object deletion by approximately 7.6%.

* **Reliability**
  * Preserved correct property lookup behavior after deletions, including for edge cases and adversarial scenarios.
  * All 2,787 tests pass without warnings.

### DIFF
--- a/changelog.d/9002-index-migrate-on-delete.md
+++ b/changelog.d/9002-index-migrate-on-delete.md
@@ -1,0 +1,24 @@
+`delete obj[k]` no longer rebuilds the shape key index.
+
+The delete clones the keys array, so the clone has a new address, misses
+`indices`, and rebuilt the whole index — decoding and FNV-hashing all ~500
+surviving property names on EVERY delete of a 500-key object.
+
+The survivors are the same strings in the same order minus one, so the index
+is shifted instead of recomputed: drop the removed slot, decrement every slot
+above it, touch no key bytes. Only a fully-built index is carried over; a
+partially built one is dropped exactly as before, since shifting it would
+misalign the un-indexed tail.
+
+Safe by construction rather than by argument: `shape_slot_lookup` re-validates
+the stored key against the requested bytes before returning a slot, so an
+index that is wrong yields a MISS and the caller's existing fallback, never a
+wrong property.
+
+Interleaved A/B, min-of-15 at quiet load (~1.0): `bench_populated_delete`
+5023 → **4641 ms, −7.6%** (mean −7.4%). Combined overwrite and realistic-name
+read unchanged.
+
+With #9000 and #9001 this brings the benchmark from 5938 to 4641 ms (−21.8%).
+Suite 2787 passed, no warnings; adversarial property differential byte-identical
+to node.
--- a/crates/perry-runtime/src/object/delete_rest.rs
+++ b/crates/perry-runtime/src/object/delete_rest.rs
@@ -370,6 +370,17 @@ pub extern "C" fn js_object_delete_field(
         }
         (*keys_cloned).length = new_count as u32;
         super::rebuild_array_layout_from_slots(keys_cloned);
+        // Carry the key index onto the clone by shifting slots, instead of
+        // letting the new address miss `indices` and re-hash every surviving
+        // property name. On a 500-key object that rebuild ran on EVERY delete.
+        // A wrong index can only cause a miss — `shape_slot_lookup` validates
+        // the stored key against the requested bytes before returning a slot.
+        super::shapes::shape_index_migrate_after_delete(
+            keys as usize,
+            keys_cloned as usize,
+            i as u32,
+            key_count as u32,
+        );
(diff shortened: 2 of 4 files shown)


proggeramlug:
**Correction to this PR's description — the mechanism is not what I wrote.**

The delete tail ends with `shape_drop(object_keys_array(obj))`, which removes `indices[keys]` for the **post-delete** array — i.e. exactly the entry this PR's migration installs. So the migrated index is discarded immediately and the next lookup rebuilds it anyway. The claim that the index is "carried across so the rebuild does not happen" is wrong as written.

The −7.6% is real and reproducible, but it comes from the other half of the same function: `shape_index_migrate_after_delete` **removes the stale entry for the OLD keys array**, which the previous code left behind on every delete. That kept the `indices` map from accumulating a dead entry per delete, and map growth/rehashing is a measurable cost on this benchmark.

I have a follow-up that makes the migration actually deliver what this PR claimed: skip the `shape_drop` when the migration succeeded, since a shifted index is *current*, not stale. Happy to have this one judged on the corrected mechanism, or to close it in favour of the follow-up if you would rather have a single clean change.

Apologies for the inaccurate description — I verified the measurement but not the causal story behind it.

proggeramlug:
Rebased onto `main` (which now carries #9001). One conflict in `shapes.rs`, resolved by keeping the module split #9001 landed and moving this PR's two new `SlotList` methods (`retain_shift`, `is_empty`) into the sibling. The rebase also dropped the stale `changelog.d/8998-*` and `9001-*` fragments, which are on `main` already.

That put `shapes.rs` at 2013 lines, over the 2000-line cap, so `shape_index_migrate_after_delete` moved next to `retain_shift` in the sibling — it is mostly `SlotList` manipulation anyway. Now 1975.

Validation: runtime lib 2788 passed / 0 failed, `cargo fmt --check`, `cargo metadata --locked`, and `scripts/run_lint_gates.sh` at the known baseline (1 of 58, the pre-existing #8929 unsubstituted-expression failure). `warnings` and `clippy` tiers both clean.

**One hazard worth recording, filed as #9006 rather than held against this PR.** The migration *moves* the index off the source array, but the source on this path is expected to be shared — `delete_rest.rs:335` says so explicitly ("the same keys_array is shared across all objects that built the same shape"), which is why the clone exists at all. So a delete from one object strands every sibling still sharing that array, and each rebuilds. The neighbouring `shape_keys_grown` documents exactly this rule for itself ("callers must pass the OWNED-grow pair only").

I checked it is not a correctness bug rather than taking the comment's word for it: `shape_slot_lookup` compares `stored == key_bytes` before returning a slot and drops the index when `indexed_len > key_count`, so a stranded or stale index can only MISS. And `bench_populated_delete.ts` deletes from a single object, so the measurement here is sound — it simply cannot see the sibling case. The fix (branch on `GC_FLAG_SHAPE_SHARED`: move when owned, clone when shared) adds cost to the deleting path and deserves its own A/B rather than being folded into this PR's numbers after the fact.

Merging.

PR #8122, opened 2026-08-15
perf(object): remove the derivable object_type and field_count header words (56 B -> 48 B) [HELD: #8157 refuted; footprint-coupled residual + new #8094 guard cost]

proggeramlug:
Closes #8113.

Removes `ObjectHeader::object_type` and `::field_count`. The header goes from
**32 bytes to 24** (16 on ILP32), a two-slot object from **56 to 48**, and the
eight-slot case from 104 to 96. Removing either word alone saves **zero** — the
struct re-pads — so this is one indivisible change. It is half of #8047's prize
and needs none of the GC descriptor-rooting work in #8112.

Both words were derivable: the receiver **kind** from `GcHeader.obj_type` plus
the ShapeId descriptor's `object_kind`, and the **live inline-slot bound** from
that descriptor's `live_inline_slot_count`.

---

## The offset-0 type confusion, and the two sites nobody had listed

`object_type` was prefix-punned against `error::ErrorHeader`'s first word, and
**nine** sites read raw offset 0 to decide Error-vs-ordinary. Seven were
catalogued in #8113 / #8047; the sweep for this PR found two more:

* `promise/rejection.rs:181` (`describe_rejection_reason`)
* `promise/rejection.rs:464` (`print_unhandled_diagnostic`)

That mattered because after the deletion offset 0 is `class_id`,
`OBJECT_TYPE_ERROR` is **2**, and class ids are handed out **from 1, densely, in
source-declaration order** (`run_pipeline.rs:545`: `let mut next_class_id = 1`).
Any survivor would have reclassified **every instance of the second class a
program declares** as an `ErrorHeader` and served `message`/`name`/`stack`/
`errors` out of its field slots — the #8100 shape exactly. Plain object literals
are `class_id == 0`, so the `OBJECT_TYPE_REGULAR` arms would have inverted in
both directions at once.

All nine now go through `error::ptr_is_native_error()` (`GcHeader.obj_type ==
GC_TYPE_ERROR`, the only kind `alloc_error` uses). A tenth offset-0 read,
`symbol.rs:385`'s `SYMBOL_MAGIC` screen, stays sound — its justification comment
named the dead field and is rewritten.

Five sabotage-shaped tests pin it. Each first asserts the confusable value
really is at offset 0, then asserts the answer. Reverting `ptr_is_native_error`
to the raw read turns three of them red (verified, then restored and re-run
green after a real rebuild — `Compiling perry-runtime v` = 1).

[...]

PR #9003, opened 2026-08-29
perf(runtime): delete's value shift stops re-resolving the slot bound per slot (−6.0%)

proggeramlug:
Fourth in the populated-delete series (#9000, #9001, #9002).

`js_object_get_field` resolves the receiver's live inline-slot bound itself, and that bound is a **shape-table probe**. So shifting N values down after a delete paid N shape-descriptor lookups — ~499 per `delete` on a 500-key object. The loop already holds that exact bound in `field_count`, and nothing in it changes the receiver's shape.

`object_field_at_with_live` exists for precisely this case (#8122) and is otherwise the identical body, inline-vs-overflow split included, so the values read are unchanged.

This also explains why `shape_descriptor_by_id` and `shape_slot_lookup` ranked so high in the delete profile: a large share of them was *this loop*, not the index lookups they exist to serve.

### Measurement

Interleaved A/B, min-of-15, quiet load (~1.0), base and branch from exact SHAs in one run:

| | base | this PR | |
|---|---|---|---|
| `bench_populated_delete` | 4651 ms | **4371 ms** | **−6.0%** (mean −6.1%) |
| combined overwrite | 37 ms | 38 ms | unchanged |

Cumulative across the series: **5938 → 4371 ms, −26.4%**.

### One optimisation deliberately NOT taken

The shift still moves values individually through barriered stores. The deferred-layout batching helper (#7630) would remove the per-slot layout note — but its settle step (`layout_finish_deferred_boxed_object`) marks the object **layout-UNKNOWN** whenever a pointer was stored, degrading a precise pointer mask to scan-all-slots. That is correct and valuable for the freshly-built objects it was designed for, and a bad trade here: it buys CPU on the delete by making every later GC scan the whole payload of an object that already had a precise mask. Left alone deliberately.

### Correctness

- Adversarial property differential — delete-then-reread under a cached slot, accessor over a cached data slot, prototype fallback after delete, `Object.freeze`, same keys in opposite orders, 300-key object in the overflow store — **byte-identical to node**.

## Summary by CodeRabbit

* **Performance**
  * Improved object property deletion performance, especially for objects with multiple properties.
  * Reduced overhead when shifting remaining property values after deletion.
  * Preserved existing read behavior while delivering cumulative benchmark improvements.

* **Bug Fixes**
  * Ensured surviving properties retain their correct positions and values after deletion.

* **Documentation**
  * Added changelog details covering performance improvements, testing, and benchmark results.

### DIFF
--- a/changelog.d/9003-delete-shift-bound.md
+++ b/changelog.d/9003-delete-shift-bound.md
@@ -1,0 +1,30 @@
+The post-delete value shift stops re-resolving the receiver's slot bound once
+per moved slot.
+
+`js_object_get_field` resolves the live inline-slot bound itself, and that
+bound is a shape-table probe — so shifting N values down after a delete paid N
+shape-descriptor lookups. A 500-key object did ~499 of them per `delete`. The
+loop already holds that exact bound in `field_count`, and nothing in it changes
+the receiver's shape.
+
+`object_field_at_with_live` exists for precisely this case (#8122) and is
+otherwise the identical body, inline-vs-overflow split included, so the values
+read are unchanged.
+
+This is also why `shape_descriptor_by_id` and `shape_slot_lookup` ranked so
+high in the delete profile: a large share was this loop, not the index lookups
+they exist to serve.
+
+Interleaved A/B, min-of-15 at quiet load (~1.0): `bench_populated_delete`
+4651 → **4371 ms, −6.0%** (mean −6.1%). Combined overwrite and realistic-name
+read unchanged.
+
+Cumulative with #9000/#9001/#9002: 5938 → 4371 ms, **−26.4%**. Suite 2787
+passed, no warnings; adversarial property differential byte-identical to node.
+
+Not done here: the shift still moves values one at a time through barriered
+stores. The deferred-layout batching helper (#7630) would remove the per-slot
+layout note, but its settle step marks the object layout-UNKNOWN whenever a
+pointer was stored — trading a precise pointer mask for scan-all-slots on an
+object that already had one. That is a GC-cost regression on an existing
+object, so it was deliberately not used.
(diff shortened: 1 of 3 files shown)


proggeramlug:
Rebased onto `main` (now carrying #9000–#9002) — clean, no conflicts.

I checked the one claim this rests on rather than taking it on trust, because "the loop already holds that exact bound" is only true if nothing between the two points republishes the descriptor — and `set_object_keys_array(obj, keys_cloned)` runs at line 386, *before* the shift loop.

It holds. `set_object_keys_array` is:

```rust
let live = object_live_slot_count(obj);
set_object_keys_array_with_live(obj, keys_array, live);
```

so it carries the existing bound forward onto the new keys edge rather than recomputing it from the shortened array. `object_live_slot_count(obj)` therefore returns the same value after line 386 as it did at line 331 where `field_count` was taken, and `js_object_get_field` would re-derive exactly `field_count`. Inside the loop the only writes are `runtime_store_jsvalue_slot` and `overflow_set` — value stores, neither publishes a descriptor — so the hoisted bound cannot go stale mid-loop.

The only other difference from `js_object_get_field` is its NaN-unbox/small-pointer guards, which are redundant here: `obj` is already a validated raw `*mut ObjectHeader` being dereferenced a few lines below at the inline-write.

It also makes the code agree with the comment that was already there — "resolves inline-vs-overflow with the CURRENT (pre-decrement) `field_count`" — since it now literally uses that value instead of re-deriving something that happens to equal it.

Validation: runtime lib 2788 passed / 0 failed, `cargo fmt --check`, `cargo metadata --locked`, `scripts/run_lint_gates.sh` at the known baseline (1 of 58, the pre-existing #8929 unsubstituted-expression failure), `warnings` and `clippy` tiers clean.

Merging — this completes the #9000/#9001/#9002/#9003 delete series.

PR #9004, opened 2026-08-29
test(gc): make extension root rewrite checks deterministic

proggeramlug:
## Summary

- force the moving-GC policy with the runtime thread-local test override in every extension test that asserts root addresses are rewritten
- restore the previous per-thread override in each guard
- replace remaining process-wide `PERRY_GC_FORCE_EVACUATE` mutations

## Release blocker

Full CI run 33227701945 failed in `perry-ext-commander::gc_mutable_scanner_rewrites_action_callback_root` because the collection legitimately stayed non-moving. The scanner itself was not broken; the test asserted address movement without selecting a moving collection.

## Verification

- all 13 affected `gc_` tests pass locally
- all 13 still pass with ambient `PERRY_GC_FORCE_EVACUATE=0`, proving the thread-local override controls the test
- `git diff --check`

## Summary by CodeRabbit

- **Tests**
  - Improved garbage-collection test coverage by consistently exercising memory evacuation scenarios across extensions.
  - Test settings are now isolated to the active runtime context and automatically restored afterward.
  - Reduced cross-test interference and improved reliability when validating callback, timer, networking, and stream behavior.
  - Replaced process-wide configuration changes with safer temporary test controls.

- **Documentation**
  - Added changelog documentation covering deterministic garbage-collection root-rewrite tests.

### DIFF
--- a/changelog.d/9004-deterministic-ext-gc-root-tests.md
+++ b/changelog.d/9004-deterministic-ext-gc-root-tests.md
@@ -1,0 +1,21 @@
+Extension-crate GC root-rewrite tests now select a moving collection through
+`perry_runtime::gc::js_gc_force_evacuation_test_override` instead of mutating
+`PERRY_GC_FORCE_EVACUATE` process-wide.
+
+These tests assert that a scanner **rewrote** a root, which is only observable
+if the collection actually moved the object — and whether a minor evacuates is
+a policy decision that legitimately declines under quiet unit-test heaps. Full
+CI run 33227701945 failed exactly that way in
+`perry-ext-commander::gc_mutable_scanner_rewrites_action_callback_root`: the
+scanner was fine, the collection simply stayed non-moving.
+
+The env-var approach it replaces was process-wide, so the forcing window was
+visible to every other libtest thread in the binary — the same shape #7946
+removed from `perry-runtime`'s own suite, where it silently turned in-place
+promotion off underneath `gc::tests::promote_in_place`. The override is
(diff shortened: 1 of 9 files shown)


proggeramlug:
Audited and verified locally with a sabotage arm, since a test-determinism change is exactly the kind that can pass by being vacuous.

**Arm 1 — as submitted, ambient `PERRY_GC_FORCE_EVACUATE=0`:** all 13 `gc_` tests across the 7 crates pass. Confirms the thread-local override, not the environment, is what the tests read.

**Arm 2 — sabotage, guard patched to `js_gc_force_evacuation_test_override(0)`:** `perry-ext-commander::gc_mutable_scanner_rewrites_action_callback_root` FAILS — the exact test named in the release blocker. So the override is load-bearing and the assertion genuinely observes movement rather than passing regardless.

Also checked the mechanism itself on `main`: `gc_force_evacuate_enabled()` and `FORCE_EVACUATE_TEST_OVERRIDE` are *not* `cfg(test)`-gated (only the RAII guard structs are), so the override is live in the non-test runtime that extension test binaries link — the thing that would otherwise make this inert. The tri-state round-trip is right too: a thread with no prior override gets `-1` back and is cleared, not pinned to "off".

Worth noting the diff touches only guard plumbing — no assertion changed — and that commander/cron/fastify/streams/ws were not forcing at all before, so they go from "whatever policy picked" to a guaranteed-moving collection.

Added the missing `changelog.d/` fragment; merging.

PR #9005, opened 2026-08-29
perf(runtime): keep the migrated key index instead of dropping it (populated delete −36.2%)

proggeramlug:
Fifth in the populated-delete series, and it corrects #9002 as well as building on it.

**What #9002 got wrong.** It shifts the key index onto the post-delete keys array — but the delete tail ends with `shape_drop(object_keys_array(obj))`, which removes `indices[keys]` for *exactly that array*. So the migrated index was discarded immediately and the next lookup rebuilt it anyway, by re-hashing every surviving key name. #9002's measured −7.6% is real, but it came from the other half of that function: pruning the **stale old** entry, which the previous code left behind on every delete. I have corrected that PR's description.

**What this does.** A migrated index is *shifted to match* the compacted array, so it is **current, not stale** — dropping it throws away the rebuild the migration exists to prevent. The drop is now skipped when the migration succeeded. Where migration does not apply (a partially built index, or a delete taking another path), the drop happens exactly as before.

### Measurement

Interleaved A/B, min-of-15, quiet load (~1.0), base and branch from exact SHAs in one run:

| | base | this PR | |
|---|---|---|---|
| `bench_populated_delete` | 4361 ms | **2781 ms** | **−36.2%** (mean −36.2%) |

Cumulative across #9000, #9001, #9002, #9003 and this: **5938 → 2781 ms, −53.2%**.

### Correctness

- Adversarial property differential — delete-then-reread under a cached slot, accessor over a cached data slot, prototype fallback after delete, `Object.freeze`, same keys in opposite orders, 300-key object in the overflow store — **byte-identical to node**. The delete-heavy cases are the ones a stale index would break, and they pass.

The index's own safety net still applies underneath all of this: `shape_slot_lookup` re-validates the stored key against the requested bytes before returning a slot, so even a wrong index yields a miss and the caller's fallback, never a wrong property.

### Still on the table

The per-delete keys-array clone remains. `GC_FLAG_SHAPE_SHARED` is stamped when the transition cache publishes an array, and `object_ops::keys_array` already treats it as authoritative for this exact decision — so an **owned** array could be compacted in place, removing the allocation entirely. That is the next one and wants care, since it mutates an array other objects could reference if the flag were ever incomplete.

## Summary by CodeRabbit

  - Improved object property deletion performance by reducing redundant index rebuilding and descriptor checks.
  - Preserved optimized property lookup indexes when deletion succeeds, avoiding unnecessary rehashing.
  - Added faster handling for compacting object keys during deletion.

- **Documentation**
  - Added release documentation covering index preservation after property deletion and benchmark improvements.

### DIFF
--- a/changelog.d/9005-keep-migrated-index.md
+++ b/changelog.d/9005-keep-migrated-index.md
@@ -1,0 +1,20 @@
+`delete obj[k]` stops rebuilding the shape key index — for real this time.
+
+#9002 shifted the index onto the post-delete keys array, but the delete tail
+ends with `shape_drop` on that same array, so the migrated index was discarded
+immediately and the next lookup rebuilt it by re-hashing every surviving key
+name. (#9002's own measured win came from pruning the STALE old entry, not
+from preserving a usable index; its description has been corrected.)
+
+A migrated index is *shifted to match* the compacted array, so it is current,
+not stale. Skipping the drop when the migration succeeded is what actually
+stops the rebuild. Everything else is unchanged: a partially built index, or a
+delete that took another path, still drops exactly as before.
+
+Interleaved A/B, min-of-15 at quiet load (~1.0): `bench_populated_delete`
+4361 → **2781 ms, −36.2%** (mean −36.2%). Combined overwrite and
+realistic-name read unchanged.
+
+Cumulative across #9000–#9003 and this: **5938 → 2781 ms, −53.2%** on perry's
+worst object-model gap. Suite 2787 passed, no warnings; adversarial property
+differential byte-identical to node.
--- a/crates/perry-runtime/src/object/delete_rest.rs
+++ b/crates/perry-runtime/src/object/delete_rest.rs
@@ -375,7 +375,7 @@ pub extern "C" fn js_object_delete_field(
         // property name. On a 500-key object that rebuild ran on EVERY delete.
         // A wrong index can only cause a miss — `shape_slot_lookup` validates
         // the stored key against the requested bytes before returning a slot.
-        super::shapes::shape_index_migrate_after_delete(
+        let index_migrated = super::shapes::shape_index_migrate_after_delete(
             keys as usize,
             keys_cloned as usize,
             i as u32,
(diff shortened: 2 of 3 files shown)


proggeramlug:
Rebased onto `main` (now carrying #9000–#9004). The `shapes.rs` conflict was structural, not semantic: #9001's split moved `shape_index_migrate_after_delete` into the `shapes_slot_list` sibling, so this PR's `#[must_use] -> bool` change was applied there instead. Also renamed the fragment `9004-keep-migrated-index.md` → `9005-` — #9004 is now the extension GC-test PR.

**I audited this one harder than the rest of the series, because it is the first change that lets a shifted index *survive* a delete.** The "a wrong index can only MISS" safety net does not hold uniformly: on the set path, `field_set_by_name/tail.rs:720-723` treats a miss as authoritative and **skips the linear scan to append unverified** ("the sidecar would have found it if it existed"). So a false miss there does not degrade to a fallback — it duplicates a key. That path was previously shielded on delete-churn workloads because `shape_drop` cleared the index every time; keeping it removes that shield.

So I probed exactly that exposure end-to-end against node 26.5.1 (the pinned oracle), on 500-key objects, above `KEYS_INDEX_THRESHOLD`:

- delete from the middle, then overwrite survivors on both sides of the removed slot
- delete then **re-add the same key** — the duplicate-key case
- 50 interleaved deletes then 50 writes, counting duplicates across `Object.keys`
- delete then read every survivor (a false miss reads `undefined`)
- interleaved delete / set / new-key on one object

All five byte-identical to node, duplicate count zero:

```
A 499 -100 -400 -0.5 -499 undefined
B 500 9999 1
C 450 0 1000 1049
D 0 undefined
E 500 0 19 undefined
```

**And I confirmed the subject was actually live rather than trusting a green run** — instrumenting the kept path shows **73** migrations retained across that probe, with output unchanged. (First attempt reported 0, which was the stale-archive guard refusing the compile, not a real zero; rebuilt with the full `-p perry -p perry-runtime-static -p perry-stdlib-static` set.)

Gates: runtime lib 2788 passed / 0 failed, `cargo fmt --check`, `cargo metadata --locked`, `run_lint_gates.sh` at the known baseline (1 of 58, pre-existing #8929), `warnings` + `clippy` clean.

One thing your correction of #9002 sharpens rather than resolves: the win there came from pruning the **old** entry, and that entry is only stale if the source array was owned. `delete_rest.rs:335` says the source is expected to be *shared*, so siblings still using it lose their accelerator. Filed as #9006 — it is throughput-only, and your "Still on the table" note about `GC_FLAG_SHAPE_SHARED` is the same lever that fixes it.

Merging.

PR #9007, opened 2026-08-29
fix(ci): read flattened GC matrix shard downloads

proggeramlug:
## Summary

- flatten all GC matrix shard artifacts into one directory in the fan-in job
- merge reports from that consistent layout for both the one-shard PR tier and multi-shard Full tier

## Release blocker

PR #9004 run 33228890420 produced a complete one-shard matrix (`PASS=436`, `FAIL=0`), but `actions/download-artifact@v8` extracted the single matching artifact directly into `gc-repsel-shards/`. The fan-in only searched a nested artifact-name directory and failed with `No such file or directory`.

## Verification

- downloaded the exact `gc-repsel-matrix-shard-1` artifact from run 33228890420
- merged it successfully from the flattened path: `PASS=436 UNVER=124 XFAIL=0 FAIL=0`
- `python3 scripts/gc_repsel_matrix_merge.py --self-test`
- `python3 scripts/gc_gate_wiring_check.py --self-test`
- `python3 scripts/gc_gate_wiring_check.py`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved garbage-collection stress test result processing by correctly combining reports from parallel test shards.
  * Prevented report-merging failures caused by artifacts being extracted into nested directories.
  * Preserved validation of expected report counts, unique shard indices, and deterministic test-to-shard assignment.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1851,6 +1851,7 @@ jobs:
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8
         with:
           pattern: gc-repsel-matrix-shard-*
+          merge-multiple: true
           path: gc-repsel-shards
       - name: Merge shards and enforce complete coverage + liveness
         env:
@@ -1863,7 +1864,7 @@ jobs:
             --expect "$EXPECT" \
             --mode "$MODE" \
             --output gc-repsel-matrix.json \
-            gc-repsel-shards/gc-repsel-matrix-shard-*/gc-repsel-matrix-*.json
+            gc-repsel-shards/gc-repsel-matrix-*.json
 
       # The remaining GC stress arms are aggregate-only. Repeating them in
       # every matrix shard would add cost without increasing their coverage.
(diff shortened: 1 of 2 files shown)


proggeramlug:
Audited and merged. The change is two lines, so I checked the thing that actually matters for a fan-in: whether flattening can let the gate pass on less evidence than it should.

It can't. Shards write **distinct** filenames — `--json "gc-repsel-matrix-$SHARD.json"` at the shard step — so `merge-multiple: true` has nothing to collide. And the merge is unchanged and content-checked, not filename-checked: `--expect` requires exactly the planned report count, rejects a shard index appearing twice ("appears in both"), validates `corpus_total` / `corpus_selected` / arm inventory per report, and asserts each test belongs to its deterministic shard. A flattened layout cannot quietly reduce coverage; a lost shard is still a hard failure.

Reproduced the bug and the fix directly on the two layouts:

- new glob on a flat directory of 8 shard reports → **8 matches**, and `gc-repsel-matrix-1.json.progress` correctly excluded (the upload's `path:` ends in `.json*`, so interruption files can be present)
- old nested glob on that same flat directory → **no matches**, which under `set -euo pipefail` passes the literal pattern to python — exactly the `No such file or directory` seen on #9004's run

Also confirmed nothing else pins the old nested path: it appears only on the line this PR changes. Gates run on the branch: `gc_repsel_matrix_merge.py --self-test`, `gc_gate_wiring_check.py` (+ `--self-test`), `gc_matrix_liveness_check.py --self-test` and `--check-registry` — all pass. `gc_gate_wiring_check.py` passing both before and after is correct rather than a miss: it covers trigger reachability (can the job run on `main`), not artifact plumbing, and this failure was a loud red rather than a silent green.

Issue #9006, opened 2026-08-29
shape_index_migrate_after_delete steals a shared keys array's index, stranding sibling objects


proggeramlug:
`shape_index_migrate_after_delete` (added in #9002) carries a keys array's slot index onto the private clone that `js_object_delete_field` makes, by **moving** it:

```rust
let Some(mut index) = inner.indices.remove(&old_keys_id) else { return };
...
inner.indices.insert(new_keys_id, index);
```

The source array on that path is expected to be **shared**. The clone exists precisely because it is — `delete_rest.rs:335`:

> CRITICAL: clone the keys_array before mutating it. The same keys_array is shared across all objects that built the same shape via `transition_cache_lookup`-hit fast paths.

So a delete from any one object strands every sibling that still shares the source array: their accelerator disappears and each rebuilds it (decode + FNV-hash every key name) on its next lookup.

The neighbouring `shape_keys_grown` documents exactly this rule for itself:

> Callers must pass the OWNED-grow pair only: a shared array's fork is a genuine transition (the clone starts a NEW identity and the old address still describes the siblings' live shape — migrating it would corrupt them).

### Not a correctness bug

Verified in the code, not just from the doc: `shape_slot_lookup` re-validates the stored key bytes against the requested bytes (`stored == key_bytes`) before returning a slot, and drops the index outright when `indexed_len > key_count`. A stranded or stale index can only produce a MISS and the caller's own fallback — never a wrong property. This is purely a throughput hazard.

### Why #9002's measurement can't see it

`bench_populated_delete.ts` deletes from a single object, so there are no siblings to strand. The regression needs N objects sharing one shape where one of them deletes — at which point the other N−1 pay a full rebuild per delete, which is the very cost #9002 removes for the deleter.

### Suggested fix

Branch on the source's `GC_FLAG_SHAPE_SHARED` bit (idiom already in `field_set_by_name/tail.rs:950`): move the index when the source is owned and about to die, clone it when the source is shared so siblings keep theirs. `SlotList` already derives `Clone`; `ShapeIndex` would need it. Cloning the table is one allocation plus a memcpy — far below the rebuild it avoids — but it is a real cost on the deleting path, so it wants its own A/B rather than being folded into #9002's numbers after the fact.

proggeramlug:
Context sharpened by #9005, now merged.

[...]

PR #9013, opened 2026-08-29
perf(runtime): compact an owned keys array in place on delete (populated delete −26.8%)

proggeramlug:
Sixth in the populated-delete series, and it removes the last per-delete allocation on the path.

The delete cloned the keys array unconditionally, because one array is shared by every object that built the shape through a `transition_cache_lookup` hit — mutating it in place would silently drop entries from siblings that never deleted anything.

But sharing is **tracked**. The caches stamp `GC_FLAG_SHAPE_SHARED` when they publish an array (`transition_cache_insert`), and both the ordinary `[[Set]]` growth path and `object_ops::keys_array` already treat that bit as authoritative for this exact clone-or-mutate decision — the latter calls it "the authoritative `GC_FLAG_SHAPE_SHARED` bit". An array without it has a single owner and can be compacted in place.

That removes a ~500-element array clone per delete — 200k of them on this benchmark — whose arena and page bookkeeping the profile showed sitting *above* the property work itself. It also keeps the array's address, so the key index needs only its slots shifted rather than a migration to a new id.

### The part that bites

The shape must be **re-published for the same array at its new key count**. Without it the object keeps a stamped ShapeId whose descriptor still claims the pre-delete count, and the shape-facts audit catches it outright:

```
published ShapeId disagrees with authoritative ObjectHeader facts
```

An earlier version of this change failed exactly there, which is a good advertisement for that assertion. `publish_object_shape_from` versions a same-pointer change internally, and `keys_changed` is false here, so the typed layout is **preserved** rather than degraded to unknown — no GC-scanning regression.

### Measurement

Interleaved A/B, min-of-13, quiet load (~1.0), base and branch from exact SHAs in one run:

| | base | this PR | |
|---|---|---|---|
| `bench_populated_delete` | 2768 ms | **2026 ms** | **−26.8%** (mean −27.0%) |
| realistic-name read | 15 ms | 14 ms | unchanged |

Cumulative across the session's delete work (#9000, #9001, #9002, #9003, #9005 and this): **5938 → 2026 ms, −65.9%** — roughly 280× node down to ~96×.

### Correctness

- `perry-runtime`: **2788 passed / 0 failed**, no warnings (`--all-targets`), including the shape-facts audit that rejected the first attempt.
- Adversarial property differential — delete-then-reread under a cached slot, accessor over a cached data slot, prototype fallback after delete, `Object.freeze`, same keys in opposite orders, 300-key object in the overflow store — **byte-identical to node**.
- Computed-key differential — delete/re-add key ordering, `Map`/`Set` keys, SSO boundary, non-ASCII, negatives, 1e21 — **byte-identical to node**.

The sibling-corruption risk this gate protects against is the reason it is worth reviewing the `GC_FLAG_SHAPE_SHARED` argument specifically: if an array can ever be shared *without* the bit, this mutates it. The two existing call sites that already rely on the bit for the same decision are the precedent I leaned on.

## Summary by CodeRabbit

  * Improved object property deletion performance by compacting eligible storage in place.
  * Reduced memory allocations during repeated property deletions.
  * Benchmarks show up to a 26.8% improvement for populated-object deletion workloads.

* **Bug Fixes**
  * Preserved correct behavior for shared object shapes by safely copying storage when necessary.
  * Maintained accurate property lookup metadata after deletions.

### DIFF
--- a/changelog.d/9013-delete-inplace-owned-keys.md
+++ b/changelog.d/9013-delete-inplace-owned-keys.md
@@ -1,0 +1,32 @@
+`delete obj[k]` stops allocating when the keys array is owned.
+
+The delete cloned the keys array unconditionally, because one array is shared
+by every object that built the shape through a `transition_cache_lookup` hit —
+mutating it in place would drop entries from siblings that never deleted
+anything.
+
+Sharing is tracked, though: the caches stamp `GC_FLAG_SHAPE_SHARED` when they
+publish an array, and both the ordinary `[[Set]]` growth path and
+`object_ops::keys_array` already treat that bit as authoritative for exactly
+this clone-or-mutate decision. An array without it has a single owner and can
+be compacted in place, which removes the last per-delete allocation here — a
+~500-element clone, 200k of them on `bench_populated_delete.ts` — and keeps
+the array's address, so the key index needs only a slot shift rather than a
+migration to a new id.
+
+**The shape must be re-published** for the same array at its new key count.
+Without that the object keeps a stamped ShapeId whose descriptor still claims
+the pre-delete count, and the shape-facts audit catches it outright
+("published ShapeId disagrees with authoritative ObjectHeader facts") — an
+earlier version of this change failed precisely there.
+`publish_object_shape_from` versions a same-pointer change internally, and
+`keys_changed` is false, so the typed layout is preserved rather than marked
+unknown.
+
+Interleaved A/B, min-of-13 at quiet load (~1.0): `bench_populated_delete`
+2768 → **2026 ms, −26.8%** (mean −27.0%). Combined overwrite and
+realistic-name read unchanged.
+
+Cumulative this session: **5938 → 2026 ms, −65.9%**, i.e. ~280× node → ~96×.
+Suite 2788 passed, no warnings; adversarial property differential and
+computed-key differential both byte-identical to node.
--- a/crates/perry-runtime/src/object/delete_rest.rs
+++ b/crates/perry-runtime/src/object/delete_rest.rs
@@ -332,58 +332,89 @@ pub extern "C" fn js_object_delete_field(
         let alloc_limit = std::cmp::max(field_count as usize, crate::object::INLINE_SLOT_FLOOR);
         let new_count = key_count - 1;
 
-        // CRITICAL: clone the keys_array before mutating it. The same
-        // keys_array is shared across all objects that built the same
-        // shape via `transition_cache_lookup`-hit fast paths. Without
-        // cloning, mutating its length / contents to remove the deleted
-        // key would corrupt every other object that picks up this
-        // shape — they'd silently lose entries they never deleted.
-        let keys_cloned = crate::array::js_array_alloc(new_count.max(1) as u32 + 4);
-        let src_elements =
-            (keys as *const u8).add(std::mem::size_of::<crate::ArrayHeader>()) as *const f64;
-        let dst_elements =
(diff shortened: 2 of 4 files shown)


proggeramlug:
You asked for the `GC_FLAG_SHAPE_SHARED` argument to be reviewed specifically. I did, and **it holds** — but I found one thing that would have broken it, so it is worth recording why, plus two defects I fixed on the branch.

## The sharing argument

Two structural properties do most of the work:

- **Five stamp sites, zero clear sites.** The bit is monotonic, so it can only ever *over*-approximate sharing. A cleared-too-early bit is the failure mode that would corrupt siblings, and there is no code that can clear it.
- **Every path that aliases an array to a second object stamps before handing it over.** I enumerated all ~25 `set_object_keys_array*` call sites and all ~40 `js_object_set_keys` callers. The object-clone paths (`alloc.rs:881/955`) allocate fresh arrays and copy; the native-module glue builds a fresh array per object. The one genuinely shared non-cache case is `iter_result.rs`, where every result object shares one cached array — and it calls `mark_shape_shared` at build time, *before* the array reaches any object, with a comment naming this exact hazard.

## The thing that nearly broke it

`transition_cache_lookup` has an early return that hands the array back **without** stamping (`mod.rs:847`), and the eager stamp at insert is capped at `TRANSITION_CACHE_EAGER_SHARE_MAX_SLOT = 64`. A 500-key object is well past 64, so that looked like exactly the hole this PR needs not to exist.

It isn't, and the mechanism is subtle enough to be worth stating: `target_len` is left 0 unless the eager stamp actually succeeded. So for slot ≥ 64 it stays 0, `entry.target_len == expected_len` cannot hold, and the lookup falls through to the lazy stamp — which **fails closed**, returning `None` rather than handing out an unstamped array. `target_len` is functioning as a "was it eagerly stamped" witness. Worth a comment there someday; nothing to change here.

## Verified end-to-end, and that the flag actually discriminates

Probe against node 26.5.1: 40 same-shape siblings with a delete on one, the same in the eager (<64) and lazy (≥64) bands, iterator results sharing one cached array, object literals with delete-then-re-add, and every-other-key deletion from one of three siblings. **All byte-identical**, zero sibling corruption.

Then I checked the arms actually ran rather than trusting a green probe: **151 in-place, 3 cloned**. The flag is not stuck true — and the 3 clones are the genuinely shared arrays.

## Two things fixed on the branch

1. **A `lint` gate rejected the new store.** `gc_store_site_inventory.py` flagged the memmove: no `GC_STORE_AUDIT` marker, and the clone arm's `INIT` reasoning does not transfer, because that destination is unpublished and this one is the live published array. Marked `BARRIERED`, matching `Array.prototype.splice`'s tail memmove. The claim is checkable, not asserted: `rebuild_array_layout_from_slots` re-runs `runtime_write_barrier_slot` over every slot for an old-gen array, and runs immediately below with `length` already set, so it covers exactly the compacted range.

2. **The new doc comment landed inside the old one.** `shape_index_shift_in_place` absorbed `shape_index_migrate_after_delete`'s entire doc block, so it read "`delete obj[k]` clones the keys array, so the result has a new address" — the opposite of what it does — and migrate was left with no docs. Split back apart.

Validation after both: runtime lib 2788/0, `fmt --check`, `metadata --locked`, `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**.

This also supersedes #9006, which I filed against #9002 for stranding siblings' index — the owned path now keeps the array's address, and the shared path still clones.

PR #9014, opened 2026-08-29
perf(runtime): decide builtin iterator .next() first in the handle ladder

proggeramlug:
`for…of` calls `.next()` once **per element**, and each of those calls walked the whole receiver-classification ladder in `dispatch_handle` — several hundred lines of small-handle, buffer, node-module, class-id and prototype checks — before reaching the `class_id` compare that actually answers it.

On a 4k-entity ECS archetype-migration frame the Map and Set iterator dispatchers were **8.6% of the row**, nearly all of it spent getting to them:

```
dispatch_handle  inclusive 41.4%
   73.3%  js_native_call_value        (the ECS closure — unrelated)
   10.4%  dispatch_map_iterator_method
   10.3%  dispatch_set_iterator_method
    1.3%  <self>
```

## The change

Hoist the three builtin-iterator `class_id` tests to the top of the ladder. They are **byte-identical** to the ones they shadow further down; only their position changes.

Safety:
- These ids are dedicated to the builtin array/Map/Set iterators, so no earlier arm of the ladder can legitimately claim such a receiver.
- The gate reads the GC header first, so a small handle or a non-object pointer never reaches the `ObjectHeader` field read.
- **A user override of `.next()` still wins**: the override lookup lives *inside* each dispatcher (`call_overridden_iterator_next`), not in the ladder that precedes it, so moving the entry point does not move the override check.

## Tests

No new test — the change is positional and every existing assertion about iterator dispatch still pins the behaviour. Local gate on the Linux build host: **perry-runtime serial 2767 passed / 0 failed**; `iterator` 32/32, `map` 108/108, `set` 201/201, array-iterator 1/1.

Paired measurement on the quiet host follows in a comment; if it does not move the row this gets closed rather than merged.

## Summary by CodeRabbit

* **Performance**
  * Improved method dispatch for Map, Set, and Array iterators through an earlier optimized path.
  * Iterator operations retain their existing behavior while potentially completing more efficiently.

### DIFF
--- a/crates/perry-runtime/src/object/native_call_method/handle_methods.rs
+++ b/crates/perry-runtime/src/object/native_call_method/handle_methods.rs
@@ -67,6 +67,51 @@ pub(super) unsafe fn dispatch_handle(
     let refreshed_args = || crate::gc::RuntimeHandleScope::refreshed_nanbox_f64_slice(arg_handles);
     let _ = (root_scope, object_handle, &refreshed_args, raw_bits, jsval);
     let _ = (method_name_ptr, method_name_len);
+
+    // Builtin collection / array iterators, decided first.
+    //
+    // `for…of` calls `.next()` once PER ELEMENT, and every one of those calls
+    // used to walk the whole ladder below — hundreds of lines of receiver
+    // classification — before reaching the `class_id` compare that actually
+    // answers it. On a 4k-entity ECS archetype-migration frame the Map and Set
+    // iterator dispatchers were 8.6% of the row, nearly all of it spent
+    // getting to them.
+    //
+    // The test hoisted here is byte-identical to the one further down; only its
+    // position changes. These class ids are dedicated to the builtin iterators,
+    // so no earlier arm of the ladder can legitimately claim such a receiver.
+    // A user override of `.next()` is still honoured: the override lookup lives
+    // INSIDE each dispatcher (`call_overridden_iterator_next`), not in the
+    // ladder that precedes it.
+    if jsval.is_pointer() {
+        let raw = jsval.as_pointer::<u8>() as usize;
+        if !crate::value::addr_class::is_small_handle(raw) && raw != 0 {
+            if let Some(header) = crate::value::addr_class::try_read_gc_header(raw) {
+                if header.obj_type == crate::gc::GC_TYPE_OBJECT {
+                    let obj = raw as *mut ObjectHeader;
+                    let class_id = (*obj).class_id;
+                    if class_id == crate::collection_iter_object::MAP_ITERATOR_CLASS_ID {
+                        return Some(crate::collection_iter_object::dispatch_map_iterator_method(
+                            obj,
+                            method_name,
+                        ));
+                    }
+                    if class_id == crate::collection_iter_object::SET_ITERATOR_CLASS_ID {
+                        return Some(crate::collection_iter_object::dispatch_set_iterator_method(
+                            obj,
+                            method_name,
+                        ));
+                    }
+                    if class_id == crate::array::ARRAY_ITERATOR_CLASS_ID {
+                        return Some(crate::array::dispatch_array_iterator_method(
+                            obj,
+                            method_name,
+                        ));
+                    }
+                }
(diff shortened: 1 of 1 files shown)


proggeramlug:
**Correctness review done and clean. Holding the merge on your own condition**, not on any finding of mine: "Paired measurement on the quiet host follows in a comment; if it does not move the row this gets closed rather than merged." That comment isn't posted yet, and since this change's entire correctness argument is *"no behaviour change"*, the perf number is the only reason to take it. I also didn't run the A/B myself — you appear to have one in flight, and a second job on the same quiet host would contend with yours and spoil both measurements.

So this is pre-approved from my side; ping when the row lands.

## What I verified

**The calls really are identical.** Both the hoisted and original arms are `dispatch_X_iterator_method(obj, method_name)` — same two arguments, so no dropped `args_ptr`/`args_len` that could change `.next(v)` handling. The enclosing guards are replicated faithfully too: the originals sit inside `if jsval.is_pointer()` (line 118) → `if (*gc_header).obj_type == GC_TYPE_OBJECT` (line 838), and the hoisted copy reconstructs exactly that, plus the small-handle and null exclusions.

**The shadowing question, which is the real risk.** Hoisting to the top means every arm in between is now skipped for these three class ids. I enumerated all 17 arms between the hoist point and the originals that do *not* test `class_id` (class-keyed arms can't match, since the ids are distinct constants), and each is excluded for a specific reason:

- small-handle branch (120–151) and the null check (165) — the hoisted guard excludes both cases explicitly
- registered-buffer (178) and typed-array (194) lookups — address-keyed registries; an iterator object is not registered in either
- the array-dispatch arms (239/243/290) — gated on `GC_TYPE_ARRAY`, and iterators are `GC_TYPE_OBJECT`
- `method_name == "slice"` (212) — this is the one that *looks* dangerous, since it keys on the method name with no receiver class test. It is safe because its body is gated on `arguments_object_to_array(raw_ptr)` returning `Some`, so it only fires for arguments objects.

**One genuine reordering:** the hoist moves Map/Set iterator dispatch ahead of `node_test::dispatch_object_method` (line 872), which previously ran between the array and Map/Set arms. Inert — that function matches only `CLASS_ID_MOCK_TRACKER` and `CLASS_ID_MOCK_CONTEXT` and returns `None` for everything else.

**Override behaviour is unchanged**, as you say: `call_overridden_iterator_next` lives inside each dispatcher, not in the ladder ahead of it, so moving the entry point cannot move the override check.

## Tests

Merged `main` in first (the branch was behind #9013). `perry-runtime` lib **2788 passed / 0 failed**; `iterator` 32/32, `map::` 8/8, `set::` 56/56, `array_iterator` 1/1; `cargo fmt --check` clean.

(Your description quotes `map` 108 and `set` 201 from the Linux host — mine are 8 and 56 because `map::`/`set::` are module-path filters rather than substring matches. Same suites, different selector; the full 2788 covers both.)

proggeramlug:
Measured on the quiet host — 9 alternating pairs, control = merge base, row `4k entities: archetype migration (add/remove) + sync`:

**+0.48%, 6/9 wins** (53.074 → 52.906 ms). The small-collection `for…of` microbenchmark is flat.

That is below the bar this PR set for itself ("if it does not move the row this gets closed rather than merged"), so I am closing it.

The result is diagnostic and worth keeping: the cost attributed to "getting to the iterator dispatchers" is **not** the ladder walk — it is the dispatchers' own per-`.next()` work: the override probe (which allocates a `"next"` key string and runs a by-name prototype lookup every call) and the fresh `{value, done}` allocation per element. Moving the class-id tests to the top of the ladder skips a few hundred straight-line checks, which at these call counts is noise.

The follow-up that attacks the real cost is `perf/fused-for-of-next`: one fused IteratorNext runtime call for the `for…of` desugar — advance in place, cached result object, null-tower override probe — with the generic two-call shape kept for every non-builtin iterator. That branch makes this reorder irrelevant for the fused path (it bypasses the ladder entirely), and manual `.next()` calls, the only remaining beneficiaries, are not hot anywhere measured.

PR #9016, opened 2026-08-29
perf(codegen): admit source-small guarded specializations to the pre-statepoint inliner (−1.2% / −1.6% wolf-ecs)

proggeramlug:
## What

The pre-statepoint inliner's admission gains a second arm: a guarded method
specialization whose **source body is small** (within
`inline_hot_small_size_cap()` statements, default 20) is admitted up to
**64 KiB** of lowered IR. The original 16 KiB arm is unchanged for everything
else. `PERRY_GUARDED_PREINLINE_MAX_IR_BYTES` overrides the raised ceiling.

## Why

`GUARDED_SPECIALIZATION_PREINLINE_MAX_IR_BYTES` judges lowered IR bytes, and
its comment explains why: one source statement can lower to a large
property/index dispatch lattice, so statement count is a bad *admission* test.
That reasoning is correct — and it makes the 16 KiB ceiling reject exactly the
methods most worth flattening: the leaves that ARE one such statement. In
wolf-ecs, `SparseSet.add`, `ECS._hasComponent` and `ECS._archChange` are one
statement each and lower to tens of KiB of guard lattice (`SparseSet.has`'s
indexed clone alone is ~46 KiB), so every call from `addComponent` stayed a
native call boundary, and LLVM could never see the redundant receiver proofs
across it.

Statement count is a bad admission test but a sound **bound**: it limits how
much source a caller can absorb. Using it as the bound and IR bytes as the
measure keeps this change structurally away from #8583's failure mode — the
bundled IIFEs whose RS4GC fan-out made `-Os` never finish are thousands of
statements, so they can never take the new arm regardless of how their IR
measures.

## Numbers

Mac mini, 11 alternating pairs vs main `3291b6425d`:

|---|---|---|---|---|---|
| wolf-ecs add_remove | 50 ms | 0.3058 | 0.3022 | **−1.23%** | 11/11 |
| wolf-ecs entity_cycle | 50 ms | 0.2436 | 0.2398 | **−1.60%** | 10/11 |
| wolf-ecs add_remove | 2 s | 0.3061 | 0.3024 | **−1.18%** | 11/11 |
| wolf-ecs entity_cycle | 2 s | 0.2438 | 0.2398 | **−1.60%** | 11/11 |

A *global* 64 KiB bump (no statement bound) measures identically, so the
source-small leaves carry the entire win; 256 KiB adds nothing.

Costs, measured: the wolf-ecs benchmark binary grows 15 → 16 MB and its compile
takes 13 s → 28 s — the file is the worst case by construction (several
one-statement 20–46 KiB lattices, each flattened at multiple sites), and the
kill switch restores the old ceiling. On a real mid-size app (hono) compile
time is flat (10.7 s → 11.0 s).

## Testing

* `perry-codegen` 1821 passed / 0 failed; `perry-runtime --lib` 2788 passed.
  `issue_8897_field_push_writeback`, `array_subclass_fill_args`.
* Lint: shape-descriptor census, address-class inventory, file size, raw-handle debt.
* Differential probes vs node unchanged (only the two pre-existing
  `Symbol.species` name divergences).

## Summary by CodeRabbit

  * Improved optimization of small guarded method specializations.
  * Small methods can now be inlined at larger intermediate sizes, helping sparse-set operations flatten into their callers and reducing native call boundaries.

* **Documentation**
  * Added changelog information describing the expanded pre-inlining behavior.


proggeramlug:
Merged. The reasoning holds up: statement count is a bad *admission* test (one statement can lower to a 46 KiB lattice) but a sound *bound*, and that asymmetry is what keeps this away from #8583 — the bundled IIFEs are thousands of statements, so no IR measurement lets them through the new arm. The original 16 KiB arm is untouched, and the env override is a genuine kill switch (setting it back to 16384 makes the new arm add nothing).

Two things I changed on the branch.

**Added a test for the admission logic**, which had none. It pins both arms and, above all, the #8583 protection: a body past the statement cap is rejected at *any* IR size. That property is the entire safety argument, and nothing was holding it. Written against the functions' own values rather than literals, so a retuned default can't quietly turn it vacuous — and sabotage-checked: replacing the statement bound with `true` fails it.

**Renumbered the changeset fragment** from `8996-` to `9016-`. #8996 is an unrelated already-merged PR (`perf/elements-inline-push`). A wrong number is invisible until a release is cut and then attributes this change to that PR — the exact failure #8978 collected four of in one day. Worth noting the gate added in #9010 now emits a `::warning::` for this case rather than staying silent, though it deliberately does not fail, since backfills and stacked PRs legitimately carry another number.

Validation on the branch with `main` merged in: `perry-codegen` 1342/0, `perry-runtime --lib` 2790/0, `fmt --check`, `metadata --locked`, and `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**.

I did not re-run your wolf-ecs A/B — the mini is busy, and the measurement is well-documented with a plausible mechanism and a global-64 KiB control showing the source-small leaves carry the whole win.

Issue #6075, opened 2026-07-06
Map/Set delete() during iteration skips the following entry


proggeramlug:
**Severity:** P1 (spec violation, silent data loss)
**Found by:** audit `fable-audit-perry-2.md`, confirmed by probe.

### What
Deleting the current entry during Map/Set iteration causes the **next** entry to be skipped.

```js
const m = new Map([["a",0],["b",1],["c",2],["d",3],["e",4],["f",5]]);
for (const [k] of m) m.delete(k);
console.log([...m.keys()], m.size);
// node: visits all 6, ends []  size 0
// perry: visits k0,k2,k4, ends size 3
```
(Add-during-iteration is already correct.)

### Why
`delete` compacts the entry buffer by shifting later entries left (`crates/perry-runtime/src/map.rs:1360-1398`, `delete_entry_at_index`), while iteration is by integer index into the live buffer (`js_map_entry_key_at` + cursor-index iterator objects in `collection_iter_object.rs`). Deleting the current entry pulls the next entry into the current slot, and `i++` skips it. Set has the same shape.

### Fix
Either decrement live iterator cursors on a delete-before-cursor, or move Map/Set to a tombstone-based ordered hash table (V8-style) — the latter also gives O(1) delete and correct delete-during-iteration semantics for free (see the object-model perf issue).

**Confidence:** high (probe + source).

PR #9017, opened 2026-08-29
perf(runtime,hir): fuse the for-of IteratorNext into one runtime call

proggeramlug:
The generic `for…of` desugar performed, **per element**: a dynamic `.next()` method dispatch, a separate IteratorNext validation call, an override probe that **allocated a `"next"` key string** and ran a by-name prototype lookup, and — for builtin collection iterators — a **fresh `{value, done}` allocation**. V8 escape-analyzes all of this away; perry executed it literally. On the ECS archetype-migration profile the two collection-iterator dispatchers were 8.6% of the row, nearly all of it protocol overhead. (#9014, which merely reordered the dispatch ladder, measured +0.48% and was closed — the ladder was never the cost; this is.)

## The changes

**`js_for_of_next(iter)`** — the desugar's next-call becomes one extern call:
- A builtin **Map/Set iterator** advances in place through the *same* dispatcher arm the manual path uses (override probe included), and reuses a cached `{value, done}` object held in the iterator's new sixth field. The cache is sound because this entry's only caller is the compiler's `for…of` desugar, whose result local is a compiler temporary the loop body cannot name, read for `done`/`value` before the next advance. The cached object is a normal 2-field object with the shared keys array, so the `.done`/`.value` reads the desugar still emits hit one stable monomorphic shape.
- **Every other receiver** (array iterators, generators, user iterators) takes the generic arm, which is byte-for-byte the two-call shape this replaces: dynamic `.next()` dispatch + `js_iterator_result_validate`.
- Manual `.next()` calls and both public dispatchers keep allocating **fresh** results — a caller that retains results observes spec behavior.

**The override probe no longer builds the prototype tower** to answer "was `next` patched?": a null tower *proves* no override exists, because the only way user code reaches the prototype object is `Object.getPrototypeOf`, which materializes it. This deletes a per-`.next()` key-string allocation from the manual path too.

**One spec tightening**: the sync desugar arm that previously skipped result validation now routes through the fused entry, which validates on the generic arm — the IteratorNext behavior the other sync driver already had.

## Tests

- `fused_next_walks_a_set_and_recycles_its_result` — order, termination, the recycled result installed in field 5, and the manual dispatcher still returning fresh objects.
- `fused_next_survives_a_mid_iteration_delete` — the #6075 cursor-repair contract holds on the fused path (it runs the same advance code).
- `fused_next_routes_other_iterators_through_the_generic_arm` — an array iterator through the generic arm.

Suites on the Linux host: runtime **serial 2772/0**, HIR 355/0, transform 119/0, codegen 1341/0, `native_proof_regressions` 281/0 — **no IR pin changed**. `stmt_loops.rs` stays exactly at the 2000-line cap. Full gate + paired measurement follow in comments.

## Summary by CodeRabbit

- **Performance**
  - Improved `for…of` iteration for built-in Map and Set collections.
  - Reduced temporary result-object creation during loop traversal.

- **Compatibility**
  - Preserved support for custom iterators and overridden `next()` methods.
  - Maintained correct behavior when entries are removed during iteration.
  - Added consistent iterator-result validation across supported iteration paths.

- **Tests**
  - Added coverage for Set iteration, mid-iteration deletions, and generic iterator fallback handling.

### DIFF
--- a/changelog.d/9017-fused-for-of-next.md
+++ b/changelog.d/9017-fused-for-of-next.md
@@ -1,0 +1,31 @@
+The `for…of` desugar's per-element `IteratorNext` is now one runtime call.
+
+The generic desugar performed, per element, a dynamic `.next()` dispatch, a
+separate validation call, an override probe that allocated a `"next"` key string
+and ran a by-name prototype lookup, and — for builtin collection iterators — a
+fresh `{value, done}` allocation. V8 escape-analyses all of that away; perry
+executed it literally, and the two collection-iterator dispatchers were 8.6% of
+an ECS archetype-migration row.
+
+`js_for_of_next(iter)` replaces the pair. A builtin Map/Set iterator advances in
+place through the same dispatcher arm the manual path uses, override probe
+included, and reuses a `{value, done}` object cached in the iterator's sixth
+field. Every other receiver — array iterators, generators, user iterators —
+takes a generic arm that is the two-call shape it replaces. Manual `.next()`
+calls and both public dispatchers keep allocating fresh results, so a caller
+that retains one still observes spec behaviour.
+
+The override probe no longer builds the prototype tower to decide whether `next`
+was patched: the only route to the prototype object is `Object.getPrototypeOf`,
+which materializes the tower, so a null tower proves no override. That removes a
+key-string allocation from the manual path too.
+
+Recycling is sound because the cached object is only reachable from the
+compiler's `for…of` desugar, whose result is a temporary the loop body cannot
+name and whose `done`/`value` are read before the next advance — the desugar
+binds the loop variable ahead of the body, so even two loops sharing one
+iterator cannot clobber an unread value.
+
+One spec tightening: the sync desugar arm that previously skipped result
+validation now routes through the fused entry, which validates on the generic
+arm — matching the other sync driver.
--- a/crates/perry-codegen/src/runtime_decls/strings.rs
+++ b/crates/perry-codegen/src/runtime_decls/strings.rs
@@ -1542,6 +1542,7 @@ pub fn declare_phase_b_strings(module: &mut LlModule) {
     module.declare_function("js_with_implicit_read", DOUBLE, &[DOUBLE, DOUBLE]);
     // Iterator-protocol result validation (for-of lazy loop).
     module.declare_function("js_iterator_result_validate", DOUBLE, &[DOUBLE]);
+    module.declare_function("js_for_of_next", DOUBLE, &[DOUBLE]);
     module.declare_function("js_global_get_or_throw_unresolved", DOUBLE, &[DOUBLE]);
     // Ambient `require` for compiled external / compilePackages modules (#5373):
     // bind a bare `require` to a createRequire-backed closure instead of throwing
--- a/crates/perry-hir/src/lower/stmt_loops.rs
+++ b/crates/perry-hir/src/lower/stmt_loops.rs
@@ -348,18 +348,20 @@ fn iterator_result_validated(call: Expr) -> Expr {
     }
 }
 
-/// `__iter.next()` (validated: a non-object result is a TypeError).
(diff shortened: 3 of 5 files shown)


proggeramlug:
Measurement on the quiet Apple silicon host — 9 alternating pairs, control = merge base `d549ef98f`, candidate = this head, row `4k entities: archetype migration (add/remove) + sync`:

|---|---|---|---|---|---|
| 9-pair screen | 53.151 ms | 52.102 ms | **+1.99%** | **9/9** | 18/18 |

Unanimous and above the kill bar, so this stays — but honesty about magnitude: the profile attributed 8.6% of the row to the two collection-iterator dispatchers, and fusing the protocol realized ~2%. The remainder of that inclusive share is the advance work itself (cursor repair, live entry reads), which is real work and stays. The concrete-typed `for…of` microbenchmark is flat, as expected — statically-typed collections take a specialized lowering that never reaches `js_get_iterator`, so this change is invisible to it by construction.

The win applies to every generic (type-erased) `for…of` in every program — `ReadonlyMap`/`ReadonlySet`/interface-typed receivers — plus the per-`.next()` key-string allocation removed from the manual path.

Where the row's remaining mass sits, for whoever picks it up next: ~30% is the ECS library's own `ensureArchetype`/`filterRegularComponentTypes` (compiled user code), ~14% is `Map.delete`'s O(n) ordered delete — the tombstone rework for that is `perf/map-tombstone-delete`, staged on top of this branch — and ~17% GC minor + shape-table rekey.

proggeramlug:
Merged. The recycled-result argument is the load-bearing one, so I checked it in the desugar rather than taking it on trust, and it holds for a reason worth stating: `lower_stmt_for_of_inner` emits `__result = next(); if (__result.done) break; <bind>; <body>` — the loop variable is bound **before** the user body. So even the worst shared-iterator shape, two loops driving one iterator and therefore one cache, cannot clobber a value the outer loop hasn't already read. Verified that case directly (probe A: `for (const [k] of it) { for (const [k2] of it) {...} }` → `o:a,i:b,i:c,i:d`, matching node).

Also checked the parts most likely to be wrong in this shape and found them right: field 5 is inside `js_object_alloc(class_id, 6)` so the collector traces it; `emit_iter_result` uses barriered `js_object_set_field` for the cached fields and re-reads the iterator through its handle after each allocation (the stale-receiver-argument class from #8427); and `js_for_of_next` is `extern "C-unwind"`, which is what #8479 requires.

Probe against node 26.5.1, 7 cases: nested loops sharing one iterator, manual `.next()` returning distinct retainable objects, retained results, the #6075 mid-iteration delete, nested loops with separate iterators, `break`/`continue`, and generators/array iterators on the generic arm. **Byte-identical to node**, and identical to a `main` build of the same probe.

## Fixed on the branch

**`-D warnings` was failing.** `iterator_prototypes_materialized` was added `#[cfg(test)]` and never called, so the lib-test target died on `dead_code`. Deleting it was the smaller fix; I used it instead, because the fast path's whole premise is "a null tower *proves* no override", which is sound only if every route to the prototype object materializes the tower. The new test pins the route user code actually takes. It is deliberately one-directional — this suite shares process globals, so asserting the tower starts null would be order-dependent — and the implication is the half the fast path relies on.

Also added the missing `changelog.d/` fragment.

## One thing found along the way — filed, not blocking

My probe originally had an eighth case: patch `.next()` on a Set iterator, then `for…of` it. **It SIGSEGVs (exit 139).** I A/B'd before attributing it, and `main` @ `06151e61f3` crashes identically, so this PR neither causes nor worsens it. Filed as #9019.

It is worth knowing here because it sits on the exact path this PR documents ("a patched `next` wins exactly as it does on the manual path") — that claim currently has no working end-to-end coverage, since the baseline crashes before reaching it. The unit-level premise is now pinned by the test above; the end-to-end half needs #9019 fixed first.

Validation: runtime lib **2794 passed / 0 failed**, `fmt --check`, `metadata --locked`, `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**.

PR #9018, opened 2026-08-29
perf(codegen): number-by-construction locals step inline in ++/-- (−2.1% / −1.1% wolf-ecs)

proggeramlug:
## What

`i++` / `i--` on a local carrying the #8105 **number-by-construction** fact now
steps inline (`fadd`/`fsub ±1.0`) instead of calling `js_to_numeric` +
`js_numeric_step` per update. One added clause in the Update lowering's gate;
boxed and captured locals are never in the fact, so the capture arms keep their
calls.

## Why

The two calls exist for BigInt stepping and ToNumeric coercion, and the gate
admitting the inline arm predates #8105: only `integer_locals` /
`unsigned_i32_locals` qualified. A counter like

```js
for (let j = arch.length - 1; j >= 0; j--) { ... }
```

can never be admitted by the integer facts — its init is not an Integer
literal — so every `j--` paid **two runtime calls per loop iteration**. That is
the standard "iterate an untyped array backwards" shape; wolf-ecs's benchmark
drivers run it once per entity per operation (`js_numeric_step` 1.15% +
`js_to_numeric`/`js_number_coerce` 1.26% of add_remove).

The number-by-construction collector admits a local only when its initialiser
and every later write is spec-guaranteed to evaluate to a Number, and the fact
already licenses a *harder* optimization (a bare `load double` with no value
check, #8105). For a value that IS a Number, both retired calls are the
identity the inline arm computes: `js_to_numeric` routes a non-BigInt through
`js_number_coerce` (identity on a Number), and `js_numeric_step`'s non-BigInt
arm is literally `numeric ± 1.0` (`value/dynamic_arith.rs`).

## Semantics

Differential check against node, exact-output identical: fractional counters
reassigned in loops, `a.length - 1` inits, postfix returning the old value,
prefix returning the new, NaN propagation, the `(-1)++ → +0` (not `-0`) edge,
BigInt locals (still step as BigInt — not in the fact), string counters (still
coerce — not in the fact), and 2^53-boundary doubles.

## Numbers

Mac mini, 11 alternating pairs, measured on top of the #9016 build (the change
is independent of #9016 — different file, different mechanism):

| benchmark | window | before | after | delta | wins |
|---|---|---|---|---|---|
| wolf-ecs add_remove | 50 ms | 0.3022 | 0.2959 | **−2.12%** | 11/11 |
| wolf-ecs entity_cycle | 50 ms | 0.2396 | 0.2371 | **−1.08%** | 10/11 |
| wolf-ecs add_remove | 2 s | 0.3024 | 0.2958 | **−2.14%** | 11/11 |
| wolf-ecs entity_cycle | 2 s | 0.2399 | 0.2372 | **−1.12%** | 11/11 |

IR confirmation: the two benchmark driver closures drop from
`step=2 to_numeric=2` calls to zero; the remaining callers are field updates
(`this.entID++`), a different lowering.

## Testing

* `perry-codegen` and `perry-runtime --lib` suites.
* Integration: `issue_8655_array_subclass_indexing`, `issue_8690_loop_versioned_arraylike`, `issue_8897_field_push_writeback`.
* ECS differential probes vs node unchanged.

## Summary by CodeRabbit

* **Performance**
  * Improved loop counter updates for numeric local variables.
  * Reduced runtime overhead for increment and decrement operations, especially in loops that count down from computed values.
  * Preserved existing numeric handling for boxed or captured variables to maintain correct behavior.


proggeramlug:
Merged. The one-clause change is sound because the fact it leans on is already stronger than what it needs: `collectors/number_by_construction.rs` admits a local only when its initialiser *and every later write* is structurally spec-guaranteed to evaluate to a Number — declared types are explicitly never evidence (#7773) — and it is already trusted for the harder claim of a bare `load double` with no value check. It is also fail-closed in exactly the directions that matter here: parameters, captured locals, closure-boxed locals and module globals are all excluded, and `let x;` drops the local since `undefined` is not a Number. So the capture arms keep their calls, as you say.

Given that, both retired calls really are the identity: `js_to_numeric` routes a non-BigInt through `js_number_coerce`, and `js_numeric_step`'s non-BigInt arm is `numeric ± 1.0`. The edges are IEEE-exact under `fadd`/`fsub` too — `-1 + 1` is `+0.0` not `-0.0`, and NaN propagates — which matches the differential you ran.

**Fixed on the branch:** the fragment was `changelog.d/0000-…`, the unfilled placeholder. That is no longer merely untidy — #9010 made it a hard failure, so this PR would have gone red on the changeset gate. Renumbered to `9018-`.

Incidentally this was the gate's first live catch. Running it against this PR before the fix:

```
$ ./scripts/check_changeset_fragment.sh PerryTS/perry 9018
::error::A changelog.d fragment still carries the 0000- placeholder. Rename it to changelog.d/9018-<slug>.md
exit=1
```

Validation with `main` merged in: `perry-codegen` 1342/0, `perry-runtime --lib` 2794/0, `fmt --check`, and `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**.

PR #9020, opened 2026-08-29
perf(runtime): O(1) ordered Map deletes via tombstoned entries

proggeramlug:
Emptying an N-entry Map cost **O(N²) three ways at once**: every ordered delete memmoved the trailing entries down, span-barriered every moved slot, and walked all three side indexes decrementing every offset above the hole. Measured earlier in this campaign: perry's per-delete cost doubles with N (0.96 → 8.2 µs from N=2k to 16k) while node sits at ~0.03 µs flat — **263× at N=16k** (131.3 ms vs 0.5 ms to empty the map). On the ECS archetype-migration row, `delete_entry_at_index` + its memmove were ~14% of the frame.

## The change

A delete now **tombstones** the entry in place:

- the key slot takes the reserved hole marker (`TAG_HOLE` — never a legal stored key; `normalize_zero` canonicalizes a leaked array hole to `undefined` before any key reaches the buffer);
- the value slot is cleared through the barriered store, so SATB marking still shades the overwritten child;
- the live count (`size`) drops while the array extent (`used`, a new `MapHeader` field pinned at offset 32) stays put.

Raw entry indices are therefore **stable**: nothing shifts, no span barrier over the tail, and the side indexes only *forget the deleted key* — the offset-repair walkers are deleted outright.

**Insertion order (#2831) is unchanged**: iteration walks raw indices and skips holes, and delete-then-re-add still appends at the end. Compaction (one pass + one span barrier + index rebuilds — the cost *one* delete used to pay) runs when tombstones outnumber live entries, before growing, or when a raw-indexed accessor observes holes — after which the typed for-of lane's new `used == size` admission holds again, so **the codegen lane self-heals** instead of misreading holes.

GC contract: the Map slot descriptor bounds its Range by `used` (holes are non-pointer markers the tag-filtered scan skips), with `size ≤ used ≤ capacity` as the corruption guard. The hand-built `MapHeader`s in the GC test fixtures now initialize the extent they predate.

`Set.delete` keeps its #8993 form; the same treatment there is a follow-up.

## Tests

Five new (`map_tombstone_tests.rs`, a sibling file — `map.rs` is far past the size cap): order + lookup across holes with delete-then-re-add, emptying 64 entries with the compaction threshold asserted, raw-indexed self-heal, iterator hole-skips including deleting the last-returned key, and `clear` resetting the extent.

Suites on the Linux host: **runtime serial 2777/0**, map 113/113, codegen 1341/0 (incl. the `map_entry_at` lane tests), `native_proof_regressions` 281/0.

**Stacked on #9017** (`perf/fused-for-of-next`) — the iterator hole-skip edits build on its dispatcher shape, so #9017 merges first. Full gate + measurements (the delete microbenchmark and the migration row) follow in comments.

## Summary by CodeRabbit

* **Performance**
  * Map deletions are now faster, maintaining insertion order without shifting remaining entries.
  * Map iteration and growth are optimized to handle deleted entries efficiently.

* **Bug Fixes**
  * Map iteration safely skips deleted entries.
  * Raw entry access and garbage collection remain consistent after deletions and compaction.

* **Compatibility**
  * Added optimized `for…of` advancement for Map and Set iterators, with fallback support for other iterators.

* **Tests**
  * Added coverage for deletion, reinsertion, iteration stability, compaction, and iterator result reuse.

### DIFF
--- a/changelog.d/9016-guarded-preinline-source-small.md
+++ b/changelog.d/9016-guarded-preinline-source-small.md
@@ -1,0 +1,3 @@
+### Changed
+
+- A guarded method specialization that is small at the source level but lowers to a large dispatch lattice is now admitted to the pre-statepoint inliner (up to 64 KiB of IR instead of 16 KiB), so one-statement leaves such as a sparse-set `add` flatten into their callers instead of staying a native call boundary.
--- a/changelog.d/9017-fused-for-of-next.md
+++ b/changelog.d/9017-fused-for-of-next.md
@@ -1,0 +1,31 @@
+The `for…of` desugar's per-element `IteratorNext` is now one runtime call.
+
+The generic desugar performed, per element, a dynamic `.next()` dispatch, a
+separate validation call, an override probe that allocated a `"next"` key string
+and ran a by-name prototype lookup, and — for builtin collection iterators — a
+fresh `{value, done}` allocation. V8 escape-analyses all of that away; perry
+executed it literally, and the two collection-iterator dispatchers were 8.6% of
+an ECS archetype-migration row.
+
+`js_for_of_next(iter)` replaces the pair. A builtin Map/Set iterator advances in
+place through the same dispatcher arm the manual path uses, override probe
+included, and reuses a `{value, done}` object cached in the iterator's sixth
+field. Every other receiver — array iterators, generators, user iterators —
+takes a generic arm that is the two-call shape it replaces. Manual `.next()`
+calls and both public dispatchers keep allocating fresh results, so a caller
+that retains one still observes spec behaviour.
+
+The override probe no longer builds the prototype tower to decide whether `next`
+was patched: the only route to the prototype object is `Object.getPrototypeOf`,
+which materializes the tower, so a null tower proves no override. That removes a
+key-string allocation from the manual path too.
+
+Recycling is sound because the cached object is only reachable from the
+compiler's `for…of` desugar, whose result is a temporary the loop body cannot
+name and whose `done`/`value` are read before the next advance — the desugar
+binds the loop variable ahead of the body, so even two loops sharing one
+iterator cannot clobber an unread value.
+
+One spec tightening: the sync desugar arm that previously skipped result
+validation now routes through the fused entry, which validates on the generic
+arm — matching the other sync driver.
--- a/changelog.d/9018-update-number-by-construction.md
+++ b/changelog.d/9018-update-number-by-construction.md
@@ -1,0 +1,3 @@
+### Changed
+
+- `i++`/`i--` on a local proven to always hold a Number (the #8105 number-by-construction fact) now steps inline as `±1.0` instead of calling `js_to_numeric` + `js_numeric_step` per update. Loops like `for (let j = a.length - 1; j >= 0; j--)`, whose counter the integer fact can never admit, drop two runtime calls per iteration.
--- a/changelog.d/9020-map-tombstone-delete.md
+++ b/changelog.d/9020-map-tombstone-delete.md
@@ -1,0 +1,30 @@
+Ordered `Map` deletes are O(1) instead of O(N).
+
+Emptying an N-entry Map was O(N²) three ways at once: every delete memmoved the
+trailing entries down, span-barriered every moved slot, and walked all three
+side indexes decrementing every offset above the hole. Per-delete cost doubled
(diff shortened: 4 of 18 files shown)


proggeramlug:
Measurements on the quiet Apple silicon host. Control = the stack base (`perf/fused-for-of-next` head `2a97c3e09`), so the tombstone's effect is isolated.

## `Map.delete` — build N entries, delete all, best of 3

|---|---:|---:|---:|
| 2,000 | 0.993 µs/delete | **0.021 µs** | 47x |
| 4,000 | 1.959 µs | **0.021 µs** | 93x |
| 8,000 | 4.041 µs | **0.021 µs** | 192x |
| 16,000 | 8.512 µs | **0.021 µs** | **405x** |

Per-delete cost is now **flat** — genuinely O(1). Emptying the 16k map: **136.18 ms → 0.33 ms**. For reference, node 26.5.1 on the same host does it in 0.50 ms at ~0.027–0.031 µs/delete — **perry is now faster than node on this operation.**

## ECS row `4k entities: archetype migration (add/remove) + sync`

|---|---|---|---|---|---|
| 9-pair screen | 51.977 ms | 44.151 ms | **+14.96%** | **9/9** | 18/18 |

## Gate

Runtime **serial 2777/0**; the parallel run's single failure is `discovers_a_map_from_a_later_loaded_shared_object`, the documented `dlopen` flake that fails on pristine main too. Codegen 1341/0, `native_proof_regressions` 281/0, transform 119/0.

## The row across this stack (all measured pairwise on the same host)

| state | row | vs node (4.37 ms) |
|---|---:|---:|
| cluster re-measure of main | 52.8 ms | 12.1x |
| + #9017 fused IteratorNext | 52.1 ms | 11.9x |
| + this PR | **44.2 ms** | **10.1x** |

Merge order: #9017 first (this branch stacks on it).

proggeramlug:
Merged. The tombstone contract's load-bearing claim is "the hole marker is never a legal stored key", so I traced every path that writes a key slot rather than taking it on trust. There are exactly three, and all are safe for different reasons:

- `map_set_resolved` — normalizes, and `normalize_zero` now maps `TAG_HOLE` to `undefined`;
- `map_set_string_key_value` — takes a typed `*const StringHeader` and writes it STRING_TAG-boxed, which cannot equal `TAG_HOLE`, so the missing `normalize_zero` there is correct rather than an oversight;
- `compact_map_entries` — only moves keys that were normalized when inserted.

Probe against node 26.5.1, 8 cases aimed at what tombstones can actually break: iteration order across holes, delete-then-re-add appending at the end rather than reusing the hole, emptying 64 entries through the compaction threshold and refilling, deleting the last-returned key mid-iteration (#6075 cursor repair), `forEach`/`values`/spread agreeing after holes, `clear` resetting the extent, and a hole never being observable as a key. **Byte-identical to node.**

Ran the evacuation arm too, since this changes the Map slot descriptor's GC range: 16 failures, all in the known pre-existing `promote_in_place`/`copying`/`oldgen` family, with **no map or collection test among them** — which is the discriminating fact for a change of this shape.

## Two things fixed on the branch

**A `lint` gate was red.** `gc_store_site_inventory.py` rejected compaction's two raw `ptr::write`s. Your justification was already right — the dirty-span barrier covers every surviving slot — but the marker sat 8 lines below the writes, outside the gate's ~6-line window. Moved one adjacent, and recorded the overlap argument while there (`out <= i` always, so a live pair only moves down within one buffer).

**The codegen lane's `32` was unbound.** `perry-codegen` has no dependency on `perry-runtime`, so nothing tied that literal to `MapHeader`. Your `offset_of!` assertion catches a field *reorder* — but the natural fix for a firing assertion is to update its expected value, which leaves the codegen string stale and still compiling, and the lane then loads the wrong header word and mis-admits holey maps. Named the literal and bound it to the runtime source, the same way `hot_tls.rs` handles this exact crate split. Sabotage-checked: changing the runtime offset to 40 now fails the codegen test.

Also added the missing `changelog.d/` fragment.

Validation: `perry-codegen` 1343/0, `perry-runtime --lib` 2799/0 (tombstone suite 5/5), `fmt --check`, `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**.

The follow-up measurements you promised never posted — merging on the strength of the mechanism and the baseline numbers in the description, but if the delete microbenchmark or the migration row disagrees, worth a follow-up.

Issue #9019, opened 2026-08-29
for…of over a Set iterator with a patched .next() segfaults (pre-existing, not #9017)


proggeramlug:
Assigning a patched `.next()` onto a builtin Set iterator and then driving it with `for…of` **segfaults** (exit 139).

## Reproducer

```ts
const s = new Set<number>([10, 20, 30, 40]);
const it: any = s.values();
let calls = 0;
const orig = it.next.bind(it);
it.next = function () { calls++; return orig(); };
const got: number[] = [];
for (const v of it) got.push(v as number);
console.log("E", calls > 0, got.join(","));
```

Node 26.5.1 (the pinned oracle) prints `E true 10,20,30,40`. Perry dies with SIGSEGV before printing anything.

## Not a regression — pre-existing

Found while auditing #9017, so I A/B'd it before attributing it. Both arms are `--profile perry-dev` builds of `-p perry -p perry-runtime-static -p perry-stdlib-static`, compiled with `PERRY_NO_AUTO_OPTIMIZE=1`:

| build | result |
|---|---|
| #9017 branch | exit 139 |
| `main` @ `06151e61f3` (no #9017) | **exit 139** |

So `main` has it too and #9017 neither causes nor worsens it. Recording that explicitly because the crash surfaces on the exact path #9017 documents ("a patched `next` wins exactly as it does on the manual path"), which makes it easy to misattribute.

## Why it matters beyond the crash

It means the override path on builtin collection iterators has no working end-to-end coverage: any test that would exercise it dies first. #9017 keeps an override probe inside the dispatcher and states that a patched `next` still wins — that claim cannot currently be validated through `for…of`, because the baseline crashes.

## Scope

- `s.values()` with a patched `next`, driven by `for…of` → SIGSEGV.
- The same iterator driven by **manual** `.next()` calls is fine (case B/C of my probe: manual `.next()` returns fresh, correct results).
- Not yet checked: `Map` iterators, array iterators, generators, and whether an own `next` assigned *before* first use behaves differently. Worth doing when someone picks this up.

No stderr, no panic message — a bare SIGSEGV, so the first step is a debug/ASAN build or `PERRY_GC_PROTECT_FROMSPACE=1` to see whether it is a rooting fault or a bad dispatch target.

PR #9021, opened 2026-08-29
perf(runtime,codegen): computed reads take the key by value (computed-key read now ~2x node)

proggeramlug:
The computed-read lowering called `js_get_string_pointer_unified` before the by-name entry, because that entry wants a `*const StringHeader`. For an SSO key that means materialising inline bytes onto the heap — an intern hash and table probe on **every read** — purely to satisfy a pointer signature. `intern_dispatch_bytes` was 5.5% of the combined overwrite loop, essentially all of it that.

The new by-value entry takes the key NaN-boxed, in escalating cost order:

1. **heap-tagged key** — unmask and pass through, nothing to do;
2. **SSO key** — probe the megamorphic read stub on its **content bits**; a hit never builds a `StringHeader`;
3. **stub miss** — a **read-only** intern probe (an intern hit cannot allocate or move anything, and the write path interns every key it stores, so a key being read has almost always been written first) and call through with the canonical pointer — still no rooting;
4. **cold** (key read before its first write) — materialise with the receiver rooted across the allocation, via the scoped `with_const_ptr` reload.

The old codegen worked around the materialisation hazard by re-deriving its receiver handle below the unbox; that workaround is deleted because the fast path no longer allocates, and the hazard now lives in the one place that does.

### Measurement

Interleaved A/B, min-of-21, quiet load (~1.1), base and branch from exact SHAs in one run. Node on the same host in brackets:

| loop | base | this PR | |
|---|---|---|---|
| combined overwrite | 38 ms (node 26) | **30 ms** | −21% |
| computed-key read | 21 ms (node 24) | **13 ms** | −38% — **~2× faster than node** |
| realistic-name read | 14 ms | 14 ms | unchanged (long keys are not SSO) |
| populated delete | 2032 ms | 2014 ms | unchanged |

### A measurement war story the reviewer deserves to know

An earlier draft of this branch appeared to regress populated delete by 38%, and I chased two wrong theories (root-push write barriers, then a shared-flag flip) before the GC diagnostic showed the branch retaining **760 MB vs 9.4 MB**. The real cause: my squash (`git reset --soft` over a stale working tree) had **silently reverted #9013's in-place delete** — and some unrelated files — so the branch was cloning a 4 KB keys array per delete again. The commit is now exactly the intended six files (`git diff origin/main --name-only` verified), and the "regression" vanished. The by-value change itself never touched delete.

### Verification

- `perry-runtime` **2790 passed / 0 failed**, `perry-codegen` suites green, no warnings (`--all-targets`).
- All lint gates pass, including the string-payload ratchet (the intern probe uses the shared `string_data` helper) and the raw-handle ceiling (scoped `with_const_ptr`).
- **Three** differentials byte-identical to node: the adversarial property suite, the computed-key suite, and a targeted stale-slot test — 192 reads across every rotation state of a delete/re-add cycle with a stable keys-array address, written specifically to catch a content-keyed stub returning a stale slot.

## Summary by CodeRabbit

* **Performance**
  * Improved computed-property reads by reducing unnecessary key processing and allocations.
  * Benchmarks show faster computed-key reads and overwrite/read loops.
* **Compatibility**
  * Preserved feedback recording, prototype resolution, exotic receiver behavior, and existing read semantics.
* **Quality**
  * Verified against the test suite and additional differential and edge-case checks with results matching Node.js.

### DIFF
--- a/changelog.d/9015-untyped-uint8array-store.md
+++ b/changelog.d/9015-untyped-uint8array-store.md
@@ -1,0 +1,13 @@
+Stores through an untyped helper now preserve Buffer-backed `Uint8Array` values.
+
+Perry's function inliner can specialize an `any`-typed index helper at a
+`Uint8Array` call site and route the write through the generic typed-array
+setter. That setter assumed every receiver had `TypedArrayHeader` layout, while
+Perry represents `Uint8Array` with `BufferHeader`; the different data offsets
+made the write disappear. The setter now validates the runtime receiver just
+like the getter and dispatches Buffer-backed or reassigned receivers through
+ordinary dynamic set semantics.
+
+The Buffer path also performs ToNumber and Uint8 modulo narrowing before its
+integer-only ABI, so NaN-boxed `any` values such as `257` and `-1` store as `1`
+and `255` rather than `0`.
--- a/changelog.d/9016-computed-read-by-value.md
+++ b/changelog.d/9016-computed-read-by-value.md
@@ -1,0 +1,36 @@
+Computed property reads take the key by value, and the combined
+overwrite+read loop drops 38 → 30 ms (node: 26); the computed-key read loop
+drops 21 → **13 ms against node's 24** — nearly twice as fast as node.
+
+The computed-read lowering called `js_get_string_pointer_unified` before the
+by-name entry, because that entry's signature wants a `*const StringHeader`.
+For an SSO key that means materialising inline bytes onto the heap — an intern
+hash and table probe on every read — purely to satisfy a pointer signature.
+`intern_dispatch_bytes` was 5.5% of the combined loop, essentially all of it
+that.
+
+The new by-value entry (`js_typed_feedback_object_get_field_by_value_f64`)
+takes the key NaN-boxed:
+
+- a heap-tagged key is unmasked and passed through — no work;
+- an SSO key probes the megamorphic read stub on its CONTENT bits — a hit
+  never builds a `StringHeader` at all;
+- a stub-missing SSO key takes a READ-ONLY intern probe (an intern hit cannot
+  allocate or move anything) and calls through with the canonical pointer —
+  still no rooting;
+- only a key read before its first write materialises, with the receiver
+  rooted across the allocation (the hazard the old codegen comment worked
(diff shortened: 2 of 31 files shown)


proggeramlug:
Merged, with `main` merged in first — after that the branch is exactly **7 files, +205/−7**, and #9016/#9017/#9018/#9020 all survive intact. I checked that specifically because of your war story: the raw `git diff origin/main` against the pre-merge branch *looks* like it deletes #9016's preinline arm, #9017's fused next and #9018's clause, but that is the ordinary artifact of a branch based on older main, not a revert. The 3-way merge keeps main's side, and I verified each symbol is present afterwards rather than trusting that.

The stale-slot hazard is handled correctly: the stub is keyed on `(receiver_shape_token, key_bits)`, so a delete/re-add rotation changes the token and misses rather than returning a stale slot, with a `live_slot_count` bound after. Probed it anyway — 4 rounds of prime/delete/re-add over a 6-key object, SSO boundary keys, non-ASCII and numeric-looking keys, the same content through different string identities, and one shape shared across 20 receivers. **All byte-identical to node.**

**Renumbered the fragment** from `9016-` to `9021-`; #9016 is the source-small preinline PR merged earlier today.

Validation: `perry-runtime --lib` 2799/0, `perry-codegen` 1343/0, `fmt --check`, `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**.

## One real bug found, filed rather than blamed here

My probe caught a divergence from node:

```ts
const c: any = {};
const b = c["k"];   // read before the key exists
c["k"] = 7;
console.log(c["k"]);   // node: 7      perry: undefined
```

I A/B'd before attributing it, and `main` @ `40f63c96b3` produces the same wrong value — your branch and main are byte-identical across the whole probe. Filed as #9024.

It is worth flagging on this PR anyway, because it lives exactly where you are working. The discriminator is interesting: the **literal**-key form fails (both `c["k"]` and `c.k`), a **variable** key is fine, and it needs the read *before* the write — which points at a compile-time absence proof reused after the write adds the property, rather than a runtime cache. It also survives `PERRY_NO_AUTO_OPTIMIZE=1`. Silent wrong value on `if (!cache[k]) { cache[k] = ... }`, so arguably nastier than #9019's crash.

PR #9022, opened 2026-08-29
perf(runtime): IC hits stop re-deriving shape-immutable facts (combined loop now beats node)

proggeramlug:
Stacked on #9021. Every stub or way hit re-proved three things about the receiver: its kind (a registry probe via `is_class_object_ptr`), the plain-ordinary verdict, and the inline slot bound (a descriptor fetch). Together **16% of the combined overwrite loop**: `write_fast_path_receiver_kind_ok` 6.8%, `shape_object_kind_by_id` 5.1%, `shape_live_inline_slot_count_by_id` 4.4%.

**All three are facts of the shape id, and descriptors are immutable per id.** The shape table only ever *inserts* descriptors; its only in-place mutations are GC bookkeeping (the relocated `keys` address, the carrier liveness bits) — verified by reading every `get_mut`/raw-record write in `shapes.rs`. So `object_kind`, `live_inline_slot_count` and `logical_key_count` cannot change under a fixed id, and a hit whose token matches the receiver's **current** stamp has already proved everything prime time proved.

The slot word now carries the one prime-time verdict a hit actually needs — `IC_SLOT_OVERFLOW_BIT`, inline region vs spill store — with overflow entries bound-checked against `logical_key_count` at prime. Hits keep exactly the per-object-mutable checks: header type, forwarded, the `Object.freeze`-family blocking flags, and the token compare. Applied to the write stub, the per-site dyn ways, and both read-stub hit sites.

### Measurement

Interleaved A/B, min-of-21, quiet load (~0.8), base = the exact parent commit:

| loop | base | this PR | node | |
|---|---|---|---|---|
| combined overwrite | 31 ms | **24 ms** | 29 ms | −23% — **now faster than node** |
| write only | — | **15 ms** | 23 ms | measured on branch; was 21 on main |
| computed-key read | 12 ms | 12 ms | 24 ms | unchanged (already ~2× node) |
| realistic-name read | 14 ms | 14 ms | 3 ms | unchanged |
| populated delete | 2005 ms | 2024 ms | 21 ms | unchanged (n=7, within noise) |

### Correctness

- `perry-runtime` **2790 passed / 0 failed**; **all 60 lint gates pass**; no warnings.
- Three differentials byte-identical to node: the adversarial property suite (freeze, accessor-over-slot, prototype fallback, delete/re-add), the computed-key suite, and the stale-slot suite (192 reads across every rotation state of delete/re-add cycles with a stable keys-array address).
- The freeze case matters specifically here: `Object.freeze` mutates header flags *without* a shape change, which is precisely why the blocking-flags check stays on the hit path while the shape-derived checks moved to prime.

## Summary by CodeRabbit

* **New Features**
  * Improved computed property reads, reducing overhead for frequently accessed string keys.
  * Improved cached property access performance for reads and writes.

* **Bug Fixes**
  * Fixed dynamic writes to `Uint8Array` values being silently ignored.
  * Corrected unsigned-byte conversion, including wraparound values such as `257` → `1` and `-1` → `255`.
  * Ensured indexed assignments use the correct behavior for different receiver types.

* **Tests**
  * Added regression coverage for `Uint8Array` reads, writes, and value conversion.


proggeramlug:
Merged. The premise is the whole PR, so I verified it directly rather than taking the summary's word: `object_kind`, `live_inline_slot_count` and `logical_key_count` really are immutable per shape id. There are exactly two `descriptors.get_mut` sites in `shapes.rs`, and both write only the relocated keys address — `record.indexed_keys = descriptor.keys` at one, and `addr_of_mut!(record.keys)` handed to the root scanner at the other. Neither can touch the three facts a hit is now allowed to assume, so a token match really has re-proved everything prime time proved.

Rebased onto `main` (it was stacked on #9021, since merged). One conflict, in `read_stub.rs`: main's `try_read_by_content_bits` ends with #9021's `limit`-based inline/overflow decision, which is exactly what this PR replaces with the prime-time `IC_SLOT_OVERFLOW_BIT` tag. Kept your side — it subsumes main's check rather than dropping it — and confirmed #9017/#9020/#9021/#9025 all survive the merge intact.

Probed the things that can still change under a fixed shape id, since those are what the hit path still has to catch:

- `Object.freeze` **after** the stub is primed (50 warm-up writes, then freeze, then a write) → write correctly ignored
- `Object.seal` after priming → existing key writable, new key rejected
- 40-key receiver spilling into overflow, primed hot, then written at both ends
- shape change after priming (delete then re-add → key order `q,p`)
- Proxy receiver with get/set traps, primed alongside a sibling of the same shape → traps still run
- class instance (non-plain receiver kind) primed then mutated
- accessor defined on a second object sharing the name of a primed data slot

**All byte-identical to node**, so the retained checks (header type, forwarded, freeze-family blocking flags, token compare) are catching what the dropped ones used to.

**Removed a duplicate changelog fragment:** the branch carried `9016-computed-read-by-value.md`, which is the same content as `9021-computed-read-by-value.md` already on main — #9016 is the source-small preinline PR, and I renumbered that fragment to `9021-` when merging it. Left `9015-untyped-uint8array-store.md` alone; it is legitimately on main from #9015.

Validation: `perry-runtime --lib` **2804 passed / 0 failed**, `fmt --check`, `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**.

PR #9025, opened 2026-08-29
perf(runtime): O(1) ordered Set deletes via tombstoned elements

proggeramlug:
The Set twin of #9020. #8993 removed `Set.delete`'s per-delete index re-hash, but every delete still **shifted the surviving elements down and span-barriered the moved slots** — O(n) per delete, O(n²) to empty a Set. (Measured on #8993: per-delete cost still grew 0.80 → 5.36 µs from N=1k to 8k; node is ~0.027 µs flat.)

## The change

Same design as the Map version, smaller surface (key-only 8-byte slots, one lookup table):

- delete **tombstones** the slot in place with the reserved hole marker (`normalize_zero` canonicalizes a leaked array hole to `undefined`, so the marker can never be a stored value);
- the live count drops while the array extent (`used`, a new `SetHeader` field pinned at offset 24) stays put;
- raw indices are **stable**, so the lookup index only *forgets the deleted value* — nothing is repaired;
- compaction runs when holes outnumber live elements, before growing, or when a raw-indexed reader observes them.

Insertion order (#2831) unchanged: iteration walks raw indices and skips holes; delete-then-re-add appends at the end.

Two hazards the recon caught and this PR closes rather than hopes about: `js_set_to_array`'s **bulk memcpy** and the **subset/disjoint element walkers** iterate raw slots — all compact first, so a hole can never defeat a subset check or leak a marker into an array.

GC contract: the element Range is bounded by `used` (holes are non-pointer markers the tag-filtered scan skips), guard `size ≤ used ≤ capacity`.

## Tests

Five new in `set_tombstone_tests.rs` (order/lookup across holes + re-add, emptying 64 with the compaction threshold asserted, raw-indexed self-heal, iterator hole-skips incl. deleting the last-returned value, clear + a subset walk over a holey set). The #8993 test that asserted the *eagerly-compacted layout* by raw pointer reads now compacts explicitly before those reads — every semantic assertion unchanged.

Suites on the Linux host: **runtime serial 2783/0**, set 210/210, map tombstone tests still green. Gate + the delete-scaling measurement follow in comments.

## Summary by CodeRabbit

* **Improvements**
  * Improved `Set.delete` performance while preserving insertion order.
  * Added safe handling for removed entries during iteration, indexed access, and set operations.
  * Improved Set behavior during memory management and automatic cleanup.

* **Bug Fixes**
  * Fixed behavior for cleared, repeatedly modified, and partially deleted Sets.

* **Tests**
  * Added coverage for deletion, compaction, iteration, clearing, and subset operations on Sets.

### DIFF
--- a/changelog.d/9025-set-tombstone-delete.md
+++ b/changelog.d/9025-set-tombstone-delete.md
@@ -1,0 +1,25 @@
+Ordered `Set` deletes are O(1) instead of O(N) — the Set twin of #9020.
+
+#8993 removed `Set.delete`'s per-delete index re-hash, but every delete still
+shifted the surviving elements down and span-barriered the moved slots, so
+emptying a Set stayed O(N²): per-delete cost grew 0.80 µs at N=1k to 5.36 µs at
+N=8k, against node's flat ~0.027 µs.
+
+A delete now tombstones the slot in place with the reserved hole marker, the
+live count drops while the array extent (`used`, a new `SetHeader` field) stays
+put, and raw indices are therefore stable — the lookup index only forgets the
+deleted value, and nothing is repaired. Compaction runs when holes outnumber
+live elements, before growing, or when a raw-indexed reader observes them.
+
+The marker can never be a stored value: `normalize_zero` canonicalizes it to
+`undefined` on every insert path, and compaction only moves values that were
+already normalized.
+
+Insertion order is unchanged — iteration walks raw indices and skips holes, and
+delete-then-re-add appends at the end. The two raw-slot readers that could have
+been defeated by a hole compact first: `js_set_to_array`'s bulk memcpy and the
+subset/disjoint element walkers, so a hole can neither leak into an array nor
+break a subset check.
+
+The GC contract bounds the element range by `used`, with `size ≤ used ≤ capacity`
+as the guard; holes are non-pointer markers the tag-filtered scan skips.
--- a/crates/perry-runtime/src/collection_iter_object.rs
+++ b/crates/perry-runtime/src/collection_iter_object.rs
@@ -344,21 +344,28 @@ unsafe fn dispatch_set_iterator_method_emit(
             }
             let cursor = f64::from_bits(js_object_get_field(iter_obj(), 1).bits()) as u32;
             let last_val = js_object_get_field(iter_obj(), 4);
-            let size = crate::set::js_set_size(set());
+            let used = crate::set::set_used_entries(set());
             let in_place = cursor > 0 && {
-                let prev = crate::set::js_set_value_at(set(), cursor - 1);
+                let prev = crate::set::set_value_raw(set(), cursor - 1);
                 crate::value::js_jsvalue_same_value_zero(prev, f64::from_bits(last_val.bits())) != 0
             };
-            let idx = next_read_index(cursor, in_place, || {
+            let mut idx = next_read_index(cursor, in_place, || {
                 crate::set::find_value_index(set(), f64::from_bits(last_val.bits()))
             });
-            if idx >= size {
-                js_object_set_field(iter_obj(), 1, JSValue::number(size as f64));
+            // Tombstoned deletes leave holes in the raw order; step over them.
+            while idx < used
+                && crate::set::set_value_raw(set(), idx).to_bits()
+                    == crate::set::SET_HOLE_VALUE_BITS
+            {
+                idx += 1;
+            }
+            if idx >= used {
+                js_object_set_field(iter_obj(), 1, JSValue::number(used as f64));
                 js_object_set_field(iter_obj(), 0, JSValue::undefined());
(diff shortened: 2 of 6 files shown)


proggeramlug:
Local merge gate on the Linux build host, head `df6292096` vs base `40f63c96b`:

| suite | result |
|---|---|
| perry-runtime (lib) | **serial 2783 passed / 0 failed** |
| perry-codegen (lib) | 1343 passed / 0 failed |
| ratchets vs merge base | unrooted-local 576 → 576, raw-handle 967 → 967, no ceiling raised |

The parallel run's single failure is `discovers_a_map_from_a_later_loaded_shared_object` — the documented `dlopen` image-walk flake that fails on pristine main in parallel runs; this diff touches neither stack maps nor roots.

The `Set.delete` scaling measurement (main vs this branch) is running and follows here.

proggeramlug:
Merged, but only after fixing a **blocking correctness bug**: tombstones leaked out of six raw-element walkers. Against node 26.5.1 on a Set with one hole:

```
                node                   this PR (before the fix)
forEach         ["a","b","d"]          ["a","b","NaN"]          <- hole yielded, "d" dropped
union           ["a","b","d","e"]      ["a","b",null,"d","e"]
intersection    ["b","d"]              ["b"]                    <- "d" lost
difference      ["a"]                  ["a",null]
symmetricDiff   ["a","e"]              ["a",null,"e"]
```

`js_set_to_array` and the subset/disjoint walkers are indeed correct, as the description says. But `js_set_foreach_impl`, `js_set_union`, `js_set_intersection`, `js_set_difference`, `js_set_symmetric_difference` and `js_set_is_superset_of` neither compact nor test `SET_HOLE_VALUE_BITS`. The tell that this was an omission rather than a decision: `is_subset_of` and `is_disjoint_from` compact; `is_superset_of` does not.

**Method note, because the probe alone would not have been enough.** I enumerated every function in `set.rs` touching `elements_ptr`/`elements.add(` and classified each as compacts / skips-holes / neither. That is how `is_superset_of` surfaced — it is in the broken set, but my probe's inputs do not discriminate it, so a probe-only review would have merged it silently.

Fixed by giving all six the `js_set_is_subset_of` compact-first prelude that already existed in the file. All eight probe lines are now byte-identical to node: order across holes, delete-then-re-add appending, emptying 64 through the compaction threshold, mid-iteration delete of the last-returned value, `forEach`/spread/`Array.from` agreeing, `clear` resetting the extent, and the set-operation family.

The fix is pushed to the branch as its own commit, so the diff is on record rather than appearing only in the squash.

Also checked the things that mattered on #9020's Map version:

- **`SetHeader` offsets** are all asserted, and `size` is still at 0 — which matters because `perry-codegen`'s `generic_dispatch.rs` assumes `MapHeader`/`SetHeader` both begin with `size: u32`. Adding `used` at 24 leaves that intact.
- **No codegen literal to bind** for `SetHeader::used` — unlike `MapHeader::used`, codegen never reads it, so the #9020 binding test has no Set counterpart to write.
- **The compaction store** needed a `GC_STORE_AUDIT` marker; the justification was already right (the dirty-span barrier below covers every surviving slot) but sat 8 lines away, outside the gate's window. Moved it adjacent and recorded the overlap argument.

Validation: `perry-runtime --lib` **2804 passed / 0 failed** (Set tombstone suite 5/5), `fmt --check`, `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**.

proggeramlug:
Measurement receipt (post-merge, for the record). Quiet Apple silicon host, best of 3; control = main-before (`257ba5814`, which already carries #9020's Map tombstones), candidate = this branch at `df6292096`.

## `Set.delete` — build N elements, delete all

|---|---:|---:|---:|
| 1,000 | 0.793 µs/delete | 0.153 µs | 5.2x |
| 2,000 | 1.324 µs | 0.135 µs | 9.8x |
| 4,000 | 2.601 µs | 0.090 µs | 29x |
| 8,000 | 5.423 µs | **0.062 µs** | **87x** |

The O(n) growth is gone — per-delete cost now *falls* with N as the amortized compaction spreads. Emptying the 8k Set: **43.39 ms → 0.50 ms**. Node 26.5.1 on the same host is ~0.027 µs flat, so Set.delete lands within ~2x of node (from ~200x) — not past it like `Map.delete`, because Set's per-delete constant still carries the barriered tombstone store; good enough that it no longer appears in any profile.

## Credit where due

The six-walker tombstone leak (`6fe5acc89c`) was found by the **gcops** session's independent audit of this branch — enumerating every function touching the raw elements buffer and classifying each, rather than probing — after my recon caught only 4 of the 10 walkers. My independent rerun of their audit confirmed their six exactly, and the fix uses this branch's own compact-first prelude at each entry. The delete microbenchmark above is unaffected by that fix (it exercises no walker), so the numbers stand for the merged head as well.

Issue #4926, opened 2026-06-10
codegen: js_box_set/js_box_get emitted with undef box-pointer operand (mutable-capture box alloc elided on taken path) — source bug behind #4898


proggeramlug:
Follow-up to #4898 / PR #4902 (deferred there by design).

## Summary

PR #4902 added a runtime gate (`is_registered_box_ptr`) so `js_box_set`/`js_box_get` silently skip pointers that were never minted by `js_box_alloc`. That makes the react-reconciler SIGBUS correctness-safe, but the **source** bug is still present: codegen can emit a `js_box_set` call whose box-pointer operand is **undef/poison** — a mutable-capture box whose allocation was elided on the taken path while the store survived.

## Evidence (from the #4898 deep-triage)

- Faulting site (pre-#4902): `str d0,[x0]` inside `js_box_set` with `x0 = &"object_get_by_name_guard"` — a read-only `__TEXT.__cstring` typed-feedback guard-name constant (#854 `register_site` arg). The module IR never stores into any guard global (verified across all `--trace llvm` modules; all 8292 guard uses in the react-reconciler production module are `register_site` args). The stray pointer is a pure optimizer/regalloc artifact of an undef operand: regalloc fills the register with whatever happens to be live, and react-reconciler's single giant minified `module.exports = function($$$config){…}` emits ~2500 register_site calls in one function, so a guard cstring pointer is what's live.
- An undef operand reaching a call means LLVM proved the box-pointer SSA value undefined on some path — i.e. a path exists where the capture box was never allocated/stored but `js_box_set` for that capture is still reachable.

## Repro

```ts
import createReconciler from 'react-reconciler';
console.log('reached');
```
(`test-files/ink-sandbox/rr_min.tsx`). Pre-#4902 this SIGBUSed at module init. Post-#4902 the bad call is silently skipped — so the localization handle is the rejection itself: an env-gated diagnostic in the `is_registered_box_ptr` failure arm (print/abort) should fire on this repro and identify the first offending site.

## Suggested attack

1. Add a temporary `PERRY_BOX_DIAG=1` (or reuse `PERRY_GC_DIAG`) print/trap in the `js_box_set`/`js_box_get` unregistered-pointer arm to confirm the stray call still occurs on main and capture a backtrace.
[...]

PR #9026, opened 2026-08-29
perf(codegen): resolve read-only boxed capture cells once per closure entry (−1.8% / −2.8% wolf-ecs)

proggeramlug:
## What

An ordinary closure body that repeatedly reads a boxed capture it never
assigns now resolves the box's **cell address** once at entry (new runtime
helper `js_box_capture_cell_ptr`) and loads the cell directly per read,
through the existing `trusted_box_capture_ptrs` read arm (per-use load +
inline TDZ check). Only the pointer is cached — never the value — so a write
through any sibling closure sharing the binding stays visible.

## Why

Every read of a boxed capture in a public closure body pays `js_box_get_bits`:
an `is_registered_box_ptr` registry probe plus one load. In wolf-ecs the probe
alone is 1.45% of the entity cycle, because the benchmark's drivers are
**hoisted function declarations** (`function add(lB)` capturing the ECS and its
queries) — the #8644/#8705 trusted-clone machinery never fires for them (it
resolves method callback parameters), so `ecs`/`qA` were re-probed and re-read
on every loop iteration.

Caching the pointer is sound where caching the value is not:

* a registered box's cell never moves — the collector rewrites the value
  *inside* the cell — and cell memory is never returned to the allocator while
  a capturing closure is live (the #8208 argument the update lowering already
  relies on);
* an unregistered pointer resolves to a shared **immutable** `undefined` cell,
  reproducing `js_box_get_bits`'s per-read answer (#4926) exactly.

Admission is narrow: never-written bindings only (the trusted write arms store
straight through the cached pointer, which must never reach the fallback
cell); read ≥2 times or inside a loop (a cold-branch single read must not
become an unconditional entry call); not in async / generator-wrapper /
CPS-async-step bodies (the repsel context gate's own exclusions).
`PERRY_BOX_CAPTURE_ENTRY_CELLS=0` restores the per-read calls.

## Numbers

Sized against a hand-hoisted source ceiling before building (−1.65%/−2.72%);
the compiler version slightly beats it. Mac mini, 11 alternating pairs on the
#9016+#9018 stack:

| benchmark | window | before | after | delta | wins |
|---|---|---|---|---|---|
| wolf-ecs add_remove | 50 ms | 0.2958 | 0.2904 | **−1.82%** | 11/11 |
| wolf-ecs entity_cycle | 50 ms | 0.2370 | 0.2305 | **−2.73%** | 10/11 |
| wolf-ecs add_remove | 2 s | 0.2959 | 0.2906 | **−1.81%** | 11/11 |
| wolf-ecs entity_cycle | 2 s | 0.2372 | 0.2307 | **−2.75%** | 11/11 |

IR: the two hot driver closures drop from 6 `js_box_get_bits` calls each to 0,
plus 2 entry resolves.

## Semantics

Differential vs node, identical output: sibling-closure mutation visibility
interleaved with reads, hoisted function declarations over `const` bindings,
pre-initialization reads, a written shared binding read in a loop, nested-loop
reads with calls in between. The `PERRY_BOX_CAPTURE_ENTRY_CELLS=0` build is
output-identical to the enabled build on the same corpus. ECS differential
probes vs node unchanged.

## Testing

* `perry-codegen` and `perry-runtime --lib` suites.
* Lint: census, address-class, gc-store-site, file-size, raw-handle debt, runtime symbols.

## Summary by CodeRabbit

* **Performance**
  * Improved closure handling for repeatedly read, non-mutated boxed captures.
  * Shared updates from other closures remain visible while reducing repeated lookup overhead.

* **Configuration**
  * Added an environment setting to enable or disable the optimized boxed-capture behavior.

* **Documentation**
  * Added changelog documentation describing the closure capture optimization.


proggeramlug:
Merged. Caching a raw cell pointer across a body is the shape that has bitten this repo repeatedly (a runtime-side cache of a heap pointer is a GC root the static checker cannot see), so I went at the two things that would make it unsound rather than the measurement.

**The admission collector is the right shape.** CLAUDE.md's own pitfall list says a capture collector's catch-all silently skips refs — that is exactly how a "never written" verdict goes wrong, and here a wrong verdict is not a missed optimization but a store through a cached pointer. Checked all three walkers:

- `scan_capture_use_stmt` has **no `_ =>` arm** — it ends with an explicit no-op list (`Break | Continue | … | ReleaseBoxes`), so the match is exhaustive and a new `Stmt` variant fails to compile instead of being skipped. That is the property that makes this safe to extend.
- `scan_capture_use_expr`'s catch-all *recurses* via `walk_expr_children`, so no expression form hides a write.
- `Expr::Update { id, op, prefix }` carries no child expressions, so counting the write without recursing loses nothing.

**Skipping nested `Expr::Closure` bodies is sound, and for a subtler reason than "they have their own maps":** a nested write mutates the shared cell, and this body loads through the cached pointer per use, so it observes the write. Only the pointer is cached, never the value — which is what makes the whole thing work. Probed it directly (case D: a nested arrow bumps the capture inside the reading loop → `6 4`, matching node).

**The fallback cell deserves its narrow admission.** `BOX_CAPTURE_UNDEFINED_CELL` is an immutable `static`, so it lands in `.rodata` and a stray store through a substituted pointer would SIGSEGV rather than corrupt silently. That is the right failure direction, but it means `writes == 0` is load-bearing, not merely an optimization heuristic — worth the comment it already has.

Probe against node 26.5.1, 8 cases: read-only capture in a loop with a sibling writer between calls, a nested closure writing the cell mid-loop, read-before-initialization, two closures sharing a binding across interleaved writes, the hoisted-function-declaration shape the PR is motivated by, and a capture reassigned to a different object between calls. **Byte-identical to node — and identical again with `PERRY_BOX_CAPTURE_ENTRY_CELLS=0`**, so the kill switch is a real off state rather than a no-op.

**Renumbered the fragment** from `0000-` to `9026-`; the placeholder is now a hard failure in the changeset gate (#9010), so this would have gone red.

Validation: `perry-codegen` 1343/0, `perry-runtime --lib` 2804/0, `fmt --check`, `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**.

PR #9028, opened 2026-08-29
fix(codegen): escape a scalar-replaced object when an undeclared field is written

### LINKED ISSUE #9024: Reading a literal-key property before its first write makes later reads return undefined
Reading a property with a **literal** key before its first write, then writing it, makes every later read return `undefined` — a silent wrong value, no error, no crash.

## Reproducer

```ts
const c: any = {};
const b = c["k"];   // read before the key exists
c["k"] = 7;
console.log(c["k"]);   // node: 7      perry: undefined
```

`Object.keys(c)` and `JSON.stringify(c)` both show the property IS there (`neverWritten`, `{"neverWritten":7}`), so the value is stored correctly and only the *read* is wrong.

## Characterization

| variant | node | perry |
|---|---|---|
| `c["k"]` read-before-write | `7` | **`undefined`** |
| `c.k` read-before-write (static syntax) | `7` | **`undefined`** |
| read twice after the write | `7 7` | **`undefined undefined`** |
| non-empty receiver (`{ z: 0 }`) first | `7` | **`undefined`** |
| **no** read before the write | `7` | `7` ✅ |
| **variable** key (`const n = "k"; c[n]`) | `7` | `7` ✅ |

The discriminator is the literal key plus a preceding absent-read. A variable key is fine, which points at a compile-time absence proof being established by the first read and reused after the write that adds the property — rather than a runtime cache.

Reproduces identically with `PERRY_NO_AUTO_OPTIMIZE=1` and with the default auto-optimize build.

## Pre-existing, on `main`

Found while auditing #9021 (which reworks computed reads), so I A/B'd before attributing it. Both builds are `--profile perry-dev` of `-p perry -p perry-runtime-static -p perry-stdlib-static`:

| build | result |
|---|---|
| #9021 branch | `undefined` |
| `main` @ `40f63c96b3` | **`undefined`** |

`main` and the #9021 branch produce byte-identical output across my whole 7-case probe, so #9021 neither causes nor worsens this. Recording that explicitly because the bug sits squarely in the area that PR touches and would be easy to pin on it.

## Why it matters

This is an ordinary shape — probe a field, then populate it:

```ts
if (!cache["key"]) { cache["key"] = compute(); }
return cache["key"];   // undefined
```

Unlike #9019 (a segfault on a patched `.next()`), this fails **silently** and returns a wrong value, which makes it the more dangerous of the two. It is not caught by the gap suite today.

Node used is the pinned oracle, v26.5.1.

proggeramlug:
Fixes #9024 — a **silent wrong value** on an ordinary pattern:

```ts
const c: any = {};
const b = c["k"];   // read before the key exists
c["k"] = 7;
c["k"]              // node: 7    perry: undefined
```

## Root cause

Scalar replacement gives each **declared** field of a non-escaping `new` its own alloca — `stored_fields = all_fields ∩ used_fields` in `stmt/let_stmt.rs`. An object literal lowers to a synthetic `__AnonShape_` class whose `all_fields` is exactly the literal's keys, so a write to a key the literal does not have has nowhere to go: the store is dropped, and every later read returns the alloca's initial `undefined`.

`escape_check.rs` already escapes such a receiver for a class **setter** and for a self-referential value, but neither the `PropertySet` nor the `PutValueSet` arm checked that the written property is a declared **field**.

## How I narrowed it

The variant table is what identified the mechanism, so it is worth recording — each row flips exactly one thing:

| variant | before |
|---|---|
| `c["k"]` read-before-write | **`undefined`** |
| `c.k` (static syntax) | **`undefined`** |
| `{ other: 1 }` receiver | **`undefined`** |
| `class C { x = 1 }`, undeclared key | **`undefined`** |
| `{ k: 0 }` — key IS declared | `7` ✅ |
| `Object.create(null)` | `7` ✅ |
| read moved into a function | `7` ✅ |
| write moved into a function | `7` ✅ |

Escaping via a function call fixing it, and a declared key being fine, is what pointed at scalar replacement rather than a runtime cache — `PERRY_PTR_SHAPE_LOCALS=0` also does *not* fix it, which ruled out the ptr-shape locals path and left `escape_news`/`escape_check`. The HIR then showed `{}` lowering to `New __AnonShape_… (fields: 0)` with the write as `PutValueSet`.

## The fix

A new `class_chain_has_field` in `class_accessors.rs`, mirroring `is_class_setter`'s chain walk, used by both arms. It is deliberately **fail-open**: an unmodeled base, or a class carrying computed members / `key_expr` fields, answers "has it", so the declared list is only trusted where it is authoritative. The check can therefore only escape *more* conservatively than the declared list justifies — never less.

This is not object-literal-specific: a user class hits the same path, so `class C { x = 1 }` then `(new C() as any)["k"] = 7` was equally wrong, and is fixed too.

## Verification

- **Sabotage**: removing the `PutValueSet` guard returns the gap test to `undefined undefined`; restoring it gives `undefined 7`.
- New gap test `test_gap_scalar_replace_new_field.ts` — computed and static syntax, user class, declared-field control, read-twice-after-write, and the `if (!cache["v"]) cache["v"] = …` memo shape. **Byte-identical to node 26.5.1.** (A new *passing* test needs no snapshot entry; `gap_snapshot.py` treats only a not-in-snapshot **failing** test as a regression.)
- `perry-codegen` 1343/0, `perry-runtime --lib` 2804/0, `perry-hir` 355/0, `fmt --check`.
- `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**.

I have not measured the perf effect. The change escapes strictly more receivers, so a workload that writes undeclared properties to a hot short-lived object loses scalar replacement there — that is the price of the store not being silently dropped, but it is worth a benchmark eye before this rides into a release.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed incorrect `undefined` results when optimized objects receive writes to declared or dynamically accessed properties.
  - Improved handling of property access across class inheritance, computed properties, and object literals.
  - Preserved repeated property reads and writes in memoization-style code.

- **Tests**
  - Added regression coverage for dot and bracket property access, class instances, undeclared fields, and repeated reads.

### DIFF
--- a/changelog.d/9028-scalar-replace-new-field.md
+++ b/changelog.d/9028-scalar-replace-new-field.md
@@ -1,0 +1,29 @@
+Reading a property before its first write no longer makes every later read
+answer `undefined`.
+
+```ts
+const c: any = {};
+const b = c["k"];   // read before the key exists
+c["k"] = 7;
+c["k"]              // was undefined, now 7
+```
+
+Scalar replacement gives each DECLARED field of a non-escaping `new` its own
+alloca (`stored_fields = all_fields ∩ used_fields`). An object literal lowers to
+a synthetic `__AnonShape_` class whose `all_fields` is exactly the literal's
+keys, so a write to a key the literal does not have had nowhere to go: the store
+was dropped and the reads saw the alloca's initial `undefined`.
+
+`escape_check.rs` escaped such a receiver for a class setter and for a
+self-referential value, but neither the `PropertySet` nor the `PutValueSet` arm
+checked that the written property is a declared field. Both now do, via
+`class_chain_has_field`, which fails toward "has it" for an unmodeled base and
+for a class carrying computed members — so it can only escape more
+conservatively than the declared list justifies, never less.
+
+Not specific to object literals: a user class took the same path, so
+`class C { x = 1 }` followed by `(new C() as any)["k"] = 7` read back wrong too.
+
+The failure was silent rather than a crash, on an ordinary shape:
+`if (!cache["v"]) cache["v"] = compute(); return cache["v"];` returned
+`undefined`. Covered by `test-files/test_gap_scalar_replace_new_field.ts`.
--- a/crates/perry-codegen/src/collectors/class_accessors.rs
+++ b/crates/perry-codegen/src/collectors/class_accessors.rs
@@ -36,6 +36,44 @@ pub fn is_class_getter(
 /// Mirror of `is_class_getter` for setters — used on the PropertySet/
 /// PropertyUpdate paths where a setter dispatch (vs. a plain field write)
 /// likewise needs a real `this` pointer.
+/// Does `class_name` (or an ancestor) declare `property` as a FIELD?
+///
+/// Scalar replacement gives each declared field its own alloca, so a write to a
+/// property the class does not declare has nowhere to go: the store is silently
+/// dropped and every later read answers `undefined` (#9024). Callers use this to
+/// escape such a receiver instead, so it takes the ordinary heap path.
+pub fn class_chain_has_field(
+    classes: &std::collections::HashMap<String, &perry_hir::Class>,
+    class_name: &str,
+    property: &str,
+) -> bool {
+    let mut cur = Some(class_name.to_string());
+    let mut seen = std::collections::HashSet::new();
+    let mut depth = 0usize;
+    while let Some(name) = cur {
+        if !seen.insert(name.clone()) || depth > 64 {
+            break;
+        }
+        depth += 1;
+        let Some(class) = classes.get(&name) else {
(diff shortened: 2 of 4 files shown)


proggeramlug:
On the perf caveat in the description, the argument is tighter than I first put it, so recording it before merge.

`class_chain_has_field` returns `false` **only** when the class is fully modeled (no unmodeled base) and carries no computed members or `key_expr` fields, and the property is genuinely not in the declared list. That is exactly the case where scalar replacement had no alloca for the store — i.e. where the write was already being dropped and the program was already wrong.

So every receiver this newly escapes was previously **miscompiled**. No correct-and-fast path loses scalar replacement; the only code that gets slower is code that was returning wrong values. That is why I am merging without a benchmark rather than holding for one.

Merging.

Issue #323, opened 2026-04-30
Array's hole should be undefined rather than 0


codehz:
## What happened

See below example

## What you expected

It should match js behavior

## Minimal reproduction

```typescript
const values = new Array(4);

console.log("values.length", values.length);
console.log("values[1]", values[1]);
console.log("values[1] === undefined", values[1] === undefined);
console.log("1 in values", 1 in values);
console.log("Object.keys(values)", Object.keys(values));

const read = values[1];
console.log("read !== undefined", read !== undefined);

values[2] = undefined;
console.log("values[2]", values[2]);
console.log("values[2] === undefined", values[2] === undefined);
console.log("2 in values", 2 in values);
```

Command you ran:

```bash
perry test.ts && ./test
```

Output:
```
values.length 4
values[1] 0
values[1] === undefined false
1 in values false
Object.keys(values) []
read !== undefined true
values[2] undefined
values[2] === undefined true
2 in values false
```
Expected:
```
values.length 4
values[1] undefined
values[1] === undefined true
1 in values false
Object.keys(values) []
read !== undefined false
values[2] undefined
values[2] === undefined true
2 in values true
```

## Environment

- **Perry version**: -/5/408
- **Host OS**: ArchLinux
- **Target**: native
- **Rust toolchain** (if building from source): rustc 1.95.0 (59807616e 2026-04-14)
- **Installed via**: from source

## Diagnostic output

```
Perry Doctor

Environment Checks
──────────────────
  ✓ perry version: 0.5.408
  ✓ update status: up to date
  ✓ clang (LLVM codegen): clang version 22.1.3
  ✓ system linker (cc): cc (GCC) 15.2.1 20260209
  ✓ runtime library: /home/codehz/Projects/perry/target/release/libperry_runtime.a
  ⚠ project config (perry.toml): not found - run: perry init

All critical checks passed with some warnings.
```

## Anything else

Also found in my ecs's demo

proggeramlug:
Fixed in 2f04bb10 (v0.5.412). Your literal repro now matches Node byte-for-byte:

```
values.length 4
values[1] undefined
values[1] === undefined true
1 in values false
Object.keys(values) []
read !== undefined false
values[2] undefined
values[2] === undefined true
2 in values true
```

[...]

PR #9029, opened 2026-08-29
feat(runtime): O(1) object deletes via tombstones (flag-gated; populated delete 6.5x on, -11% off)

proggeramlug:
The object-side sibling of #9020's Map tombstones, PR 1 of the sequence agreed in the tombstone design doc: hole representation + walkers + threshold squeeze, behind `PERRY_OBJECT_TOMBSTONES=1` (**default OFF** — flag-off behavior is byte-identical and the merged suite/gates prove it).

With the flag on, `bench_populated_delete` drops **2089 → 1050 ms (−50%)**; combined overwrite and realistic-name reads unchanged.

### Design

`delete obj[k]` on an **owned** keys array (no `GC_FLAG_SHAPE_SHARED` — the authority two existing call sites already trust for clone-or-mutate) writes a hole over the key slot and clears the value via the barriered stores, exactly the Map idiom. No shift, no layout rebuild, no index shift, no live-bound update. Squeeze at half-holes, one overlap-safe pass mirroring `compact_map_entries`, span-barriered.

Re-adds append (JS requires the re-added key to move to the END of enumeration order), so the array is bounded at 2× live between squeezes.

### The hard part was descriptor lifecycle, and it took three rounds

Per-delete shape-identity change is **forced**: the per-site dyn-IC ways live in generated-code globals the runtime cannot reach, so a deleted key's cached `(token, key) → slot` entries can only be retired by changing the token. Each hole-delete therefore publishes a successor id (`hole_count` is now part of `ShapeFacts`, covered by the facts-exhaustiveness test).

1. First version: publish-only. Descriptors accumulated against ONE stable address; the reverse-index list grew per delete and every publish walked it — **26× slower** (53.6 s).
2. Retire the direct predecessor: **halved** (25.1 s) — because delete-then-re-add mints an id on the *append* side too.
3. Sweep **every** stale id for the owned address at the hole publish: each is unreachable the instant the header word restamps (single owner, same invariant the gate itself trusts). Result: **1050 ms, 2× faster than the compacting delete.**

### Two bugs the differentials caught before review

- `Object.getOwnPropertyNames` has its own raw-push walk (missed by my grep audit; found by the enumeration differential emitting `null`).
- `js_array_get` translates `TAG_HOLE` → `undefined` per OrdinaryGet (#323), so hole-skips comparing `TAG_HOLE` after reading through it were **dead code**. Key walks now skip both forms — `undefined` is never a legal key.

### Verification

- `perry-runtime` **2799 passed / 0 failed**; **all 60 lint gates pass** (shapes.rs split under the 2000-line cap; the census gate retargeted at `shape_descriptor_ensure_with_holes`, where its checked ordering now lives; the squeeze's key write carries the span-barrier audit marker).
- **Four differentials byte-identical to node in BOTH flag states**: enumeration (keys/values/entries/for-in/stringify/spread/rest across interleaved deletes, re-adds, threshold crossings, overflow-slot objects), adversarial property, computed-key, and stale-slot suites.

### Not in this PR

Flag-on remains ~5 µs/delete (node: ~0.1) — the residue is the publish/sweep machinery, and reducing it is PR 2's job before any default-on flip. Coordinated with the ECS session: their `scan_shape_table_rekey_mut` work avoids `ShapeDescriptor`/`ShapeFacts`, so no structural overlap.

## Summary by CodeRabbit

* **New Features**
  * Added an optional faster mode for deleting object properties.
  * Property lookups remain accurate after deletions, including before cleanup.
  * Improved handling of repeated deletions and sparse objects.

* **Bug Fixes**
  * Deleted properties are now consistently omitted from serialization and property-name enumeration.
  * Prevented deleted properties from appearing as invalid `null` keys in output.

* **Documentation**
  * Added documentation covering deletion behavior, performance, and verification.


proggeramlug:
Rebased onto current main (picks up #9021/#9022) and extended with the second optimization the flag-on profile demanded: **a complete key index now proves absence** — `keys_find_slot_by_bytes` was 60.4% of flag-on delete time because a tombstoned key's re-add paid a full linear backstop scan per delete. With `indexed_len == key_count` the index covers every live slot (holes index as nothing, stale bucket entries fail content validation), so a miss returns without scanning.

Numbers on the shared Linux box, interleaved min-of-7 against the exact main tip (`base6` binaries built from the same script):

| bench | main | branch flag-off | branch flag-on |
|---|---|---|---|
| populated delete | 2027 ms | **1796 ms (−11%)** | **417 ms (4.9×)** |
| combined overwrite | 25 ms | 25 ms | — |
| realistic-name read | 16 ms | 16 ms | — |

The flag-off win is the absence verdict helping ordinary deletes too. Flag-on meets the design doc's ≥5×-vs-2010ms acceptance gate (417 vs the ≤400 target, within tick noise; node is 21 ms on this host → ~20× remaining, from 96× at the doc's writing and 280× at campaign start).

Correctness: 2805 runtime tests; all four differentials (enumeration, adversarial read-stub, SSO parity, stale-slot) byte-identical between flag states AND to node on the rebased build — the enumeration one is the specific trap for a wrong-absent verdict (it would append a duplicate key).

proggeramlug:
Merged. Default-OFF with a byte-identical flag-off path is the right shape for this, and the enumeration audit holds up.

I applied the same enumerate-and-classify sweep that found #9025's six leaking Set walkers — every function touching raw key slots in `enumeration.rs`, `descriptors.rs`, `stringify.rs` and `delete_rest.rs`, classified as compacts / skips-holes / neither. It flagged `js_object_entries_value` as neither, but that is a false positive: the raw-slot touch is in its *string* branch and the object path delegates to `js_object_entries`. So your audit plus the enumeration differential got there — no equivalent of #9025's `isSupersetOf` hiding in this one.

Probed 13 enumeration surfaces against node 26.5.1 in **both** flag states — `keys`/`values`/`entries`/`getOwnPropertyNames`/`JSON.stringify`/`for-in`/spread/rest, re-add moving the key to the end, crossing the squeeze threshold and then enumerating again, a 40-key overflow object, and `in`/`getOwnPropertyDescriptor`/deleting an absent key. **Byte-identical both ways**, and still byte-identical after your three follow-up commits (including the no-backstop-scan change).

Good catch on `js_array_get` translating `TAG_HOLE` → `undefined`, making the post-read `TAG_HOLE` comparisons dead code. That is the kind of hole-skip that looks like coverage and isn't.

## What I changed

**One thing, and it is about the gate rather than the feature.** `shape_descriptor_census.py` pins eight functions *by name* inside `shapes.rs` through `function_body(shapes, ...)`, and `function_body` on a name that has left that file inspects an **empty body and reports success** — #8918's exact failure mode, which I repaired once already.

That stopped being theoretical with this PR. `shapes.rs` is against the 2000-line cap, so helpers keep migrating to the `shapes_slot_list.rs` sibling — you moved the debug shape-parity asserts there in `266b4ddfa8`, and earlier PRs moved `SlotList`, the scan bookkeeping and the delete-migration helpers. The next *pinned* function to cross that split would silently disarm its own check. You already had to retarget the census once in this PR (`_with_generation` → `_with_holes`) for exactly this reason; the census now reads both files as one logical unit so that retarget is not needed again.

I verified it still bites afterwards rather than assuming: reordering `inner.descriptors.insert` against the reverse-index writes in `shape_descriptor_ensure_with_holes` fails the census.

Two things I had prepared and then dropped as redundant, because you pushed them while I was validating: the fragment renumber (`9026-` → `9029-`) and a `shapes.rs` size fix — your head is 1984 lines and the fragment is already keyed correctly. Rebased onto your commits rather than force-pushing over them.

Validation on your head: `perry-runtime --lib` 2805/0, `fmt --check`, `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**.

proggeramlug:
Second round on the branch — a walker audit (the default-on prerequisite) plus its consequences. **Final numbers, interleaved min-of-7 vs the exact main tip on the shared Linux box:**

| bench | main | branch flag-off | branch flag-on |
|---|---|---|---|
| populated delete | 2030 ms | 1810 ms (−11%) | **~315 ms (6.5×, ~15× node)** |
| combined overwrite | 27 ms | 27 ms | — |
| realistic-name read | 17 ms | 17 ms | — |

The audit enumerated every runtime file touching keys arrays (57) and classified each use site. Four real flag-on bugs, none reachable by the four differentials:

1. **SIGSEGV**: the JSON shape-prefix template dereferenced a hole's bits as a StringHeader (`JSON.stringify` of an array of holed `class_id==0` objects — e.g. `JSON.parse` output after two deletes). The differentials' objects carry `__AnonShape` class ids or cache-shared keys, so the template never engaged for them. Holed shapes now bail to the hole-aware slow path; unit-pinned.
2. The worker-thread serializer pairs keys/fields positionally — a hole became a phantom `""` key on the worker. Skips the pair now.
3. `diagnostics_channel`'s error-prop walk minted a phantom `"undefined"` prop from a hole.
4. **The accounting bug**: both lineage-carrying shape publishes hardcoded `hole_count: 0`, so a re-add append reset the squeeze accounting — delete/re-add churn never squeezed and the keys array grew unbounded (~3 MB over the bench; O(1) per op in time, so invisible to every timing gate). They now carry `lineage.hole_count`; a 60-cycle churn test pins the 2×-live-size bound. Fixing it cut flag-on time a further 25% — the index stays small and hot — which is how 417 became ~315.

Verification on the branch tip: 2807 runtime tests; fmt/census/class-id/file-size gates green (census's mint-then-stamp sabotage fixture retargeted at the `_with_holes` mint); six fixtures — the four differentials plus two new holed-JSON-array repros — byte-identical between flag states AND to node.

PR #9032, opened 2026-08-29
perf(codegen): decide object truthiness inline, on the slow edge (−0.5% / −0.4% wolf-ecs)

proggeramlug:
## What

The inline truthiness test gains a `truthy.obj` block **on the would-be-slow
edge**: a POINTER-tagged value with a nonzero payload is decided truthy in
three ALU ops instead of calling `js_is_truthy`. The singleton path
(`true`/`false`/`undefined`/`null`) is byte-identical to before; strings
(empty is falsy — needs a length load), BigInt (`0n` is falsy), and the
zero-payload pointer case keep the call, mirroring `js_is_truthy`'s own
POINTER_TAG arm exactly.

## Why

Objects are unconditionally truthy, yet every `if (obj)`, `obj || fallback`,
and `while (node)` paid a runtime call. wolf-ecs's `_archChange` does
`t.change[e] || (…)` once per component add/remove — an object on every hit —
and `js_is_truthy` was 0.56–0.90% of both benchmarks, the largest runtime
symbol left in an otherwise clean profile.

## Placement, measured both ways

The first version folded the pointer test into the existing tag-arm compares
and **regressed both benchmarks +0.6% (0/11)** — it lengthens the common
dependency chain of every `if (flag)` whose singleton compares already decide
the answer. Moving the identical test onto the edge that previously went
straight to the call charges only values that were about to pay it:

|---|---|---|---|---|---|
| wolf-ecs add_remove | 50 ms | 0.2893 | 0.2878 | **−0.52%** | 10/11 |
| wolf-ecs entity_cycle | 50 ms | 0.2306 | 0.2297 | **−0.41%** | 9/11 |
| wolf-ecs add_remove | 2 s | 0.2896 | 0.2880 | **−0.54%** | 11/11 |
| wolf-ecs entity_cycle | 2 s | 0.2307 | 0.2298 | **−0.43%** | 10/11 |

(Both measured against main `ce8e2a4bf7` on the quiet Mac mini, 11 alternating
pairs.)

## Testing

* Differential vs node: objects/arrays/closures/symbols/wrapper objects
  (`new Boolean(false)` is truthy), `||` chains, `while` over a
  null-terminated list, empty/non-empty strings, `0n`/`1n`, NaN/±0, and the
  four singletons — identical output.
* `perry-codegen` 1823 / `perry-runtime --lib` 2804 — all green.

## Summary by CodeRabbit

  * Inline truthiness checks for object values now avoid runtime evaluation when the object has a nonzero pointer payload.
  * Conditions such as `if (obj)`, `obj || fallback`, and `while (node)` may execute more efficiently.
  * Strings, BigInts, and zero-payload objects continue using the existing truthiness behavior.


proggeramlug:
Merged. The risk in an inline truthiness arm is that it *steals* a falsy case from the call, so I checked it against `js_is_truthy` directly rather than against the description.

It mirrors exactly. `value/truthy.rs`'s POINTER_TAG arm treats the null pointer (`0x7FFD_0000_0000_0000`) as falsy and everything else with that tag as truthy; the new block requires `top16 == POINTER_TAG` **and** a nonzero payload, so the zero-payload case falls through to the call and still answers falsy. STRING_TAG and BIGINT_TAG are separate tags and never reach the new arm, so `""` and `0n` keep their length/zero checks.

The phi is right too: `obj_truthy` is only reachable at `merge` along the edge where it is true, so the incoming value is correct rather than merely convenient.

Probed 29 value kinds against node 26.5.1 — objects, arrays, functions, arrows, `Date`, `Map`, `Set`, `RegExp`, `Error`, `Symbol`, boxed `Number(0)`/`String("")`/`Boolean(false)` (all truthy despite falsy primitives inside), plus every falsy: `""`, `0`, `-0`, `NaN`, `null`, `undefined`, `false`, `0n`. Then the forms that actually route through this lowering: `||`, `&&`, ternary, negation, `while (node)` over a list terminating on `null`, and the `t.change[e] || (t.change[e] = …)` shape the PR is motivated by. **Byte-identical.**

Worth noting the placement result in the description is the interesting part and I would not have predicted it: folding the pointer test into the existing tag-arm compares **regressed** both benchmarks +0.6% (0/11) by lengthening the dependency chain of every `if (flag)`, and only moving the identical test onto the would-be-slow edge turned it into a win. That is a good argument for the current shape being load-bearing rather than incidental, and worth preserving if anyone tidies this function later.

Renumbered the fragment `0000-` → `9032-`; the placeholder is a hard failure in the changeset gate since #9010.

Validation: `perry-runtime --lib` 2805/0, `perry-codegen` 1343/0, `fmt --check`, `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**.

PR #9033, opened 2026-08-29
perf(codegen): integer-provenance locals are canonical raw f64 — typed push 60→2.4 ns, ahead of node

proggeramlug:
## What

`expr_produces_canonical_raw_f64` gains a `LocalGet` arm: a local with integer
provenance (`i32_counter_slots` / `integer_locals`) held in plain storage (not
boxed, captured, or a module global) is a canonical raw f64 by dataflow — it
can never be NaN, so its bits can never fall inside the NaN-box tag window,
which is the entire hazard the predicate guards.

## Why

`a.push(i)` on a receiver with a **static numeric-layout proof** (a `number[]`
local or parameter) routed into the three-call guarded-numeric tier
(`js_typed_feedback_numeric_array_push_guard` + `js_array_numeric_push_f64_unboxed`
+ length, ~60 ns/push), while the same push on an *untyped* receiver took the
full inline store tier. The typed annotation made the code slower — the
push-side twin of the read inversion #6904 retired for the dense read tier.

The escape hatch already existed: `keep_guarded_numeric_push` skips the
three-call tier when the value is provably canonical raw f64 — but with no
`LocalGet` arm, the most common append shape (`for (…) a.push(i)`) never
qualified. The dense raw-f64 invariant survives the inline store for exactly
the reason the arm states: the appended slot receives canonical bits.

## Numbers

Operation-level differential (found by the ECS-cluster ops suite; effects are
25×, far above dev-box noise; node 26.5.1 measures 2.8–3.7 ns on these shapes):

|---|---|---|---|
| push through `number[]` parameter | 60.5 ns | **2.4 ns** | 3.7 |
| push through fresh `number[]` local | 60.4 ns | **2.4 ns** | 3.5 |
| growth-only push loop | 65.1 ns | **3.7 ns** | 2.8 |
| `push` + periodic `length = 0` | 67.8 ns | **2.5 ns** | 3.6 |

perry is now ahead of node on the isolated append. wolf-ecs (untyped
receivers, quiet Mac mini, 11 pairs, both windows): ±0.1%, neutral as
expected.

Module-global receivers remain on the runtime call (~26 ns) — both apush
tiers require a `ctx.locals` slot; that receiver-naming generalization is
being taken by the ECS-cluster campaign as part of its store-side work, per
coordination.

## Testing

* Differential vs node: counters through capacity growth, NaN/±0/Infinity via
  a non-integer local, mixed-type pushes deopting the typed receiver,
  `length = 0` reuse cycles, push return values, frozen-receiver TypeError —
  identical output.
* `perry-codegen` 1823 (including all `array_push` IR tests) and
  `perry-runtime --lib` 2805 — green.

Complementary to the ECS-cluster campaign's `perf/chained-array-ctor-type`
(the `new Array<T>(n).fill(0)` type-loss fix): that change makes *more*
receivers statically numeric, which makes this arm fire *more* often — without
it, those receivers would regress into the three-call tier this PR retires.

## Summary by CodeRabbit

  * Improved numeric array appends for integer-based values.
  * `a.push(i)` now uses a faster optimized path when the array and value types are known, significantly reducing overhead in eligible cases.
* **Documentation**
  * Added a changelog entry describing the improved append performance.


proggeramlug:
Merged. The arm extends a predicate that guards a NaN-boxing hazard, so I went after whether "integer provenance" is a real proof rather than a declared type — the #7773 distinction.

It holds, and the storage conditions are what make it hold: `boxed_vars`, `closure_captures` and `module_globals` are all excluded, so the local cannot be rebound by code the dataflow walk cannot see, and it must live in a plain slot. That mirrors `local_get_produces_non_pointer_bits_by_dataflow`, as the comment says.

One thing worth flagging for future work on this predicate, not a blocker here: `collect_integer_locals` takes a `numeric_locals` argument that its own comment describes as "locals whose **declared type** says they hold a number", used to decide when a `Uint8ArrayGet` counts as a byte read. That is a declared-type input feeding a set this PR now treats as a canonical-raw-f64 proof. It is narrow today and I could not construct a violation, but it is the seam to check first if this predicate is ever widened again.

I tried to construct one, across seven shapes: the loop-counter push the PR targets, a `Uint8Array` read with a non-numeric key (the #7700 shape), an out-of-bounds read, a counter leaving the integer domain via `v / 0`, `-0` and the 2^53 boundary, typed vs untyped receivers, and reading every pushed element back. **#9033 is byte-identical to `main` on all seven** — so it changes no observable behaviour, only which lane runs.

## One pre-existing bug found, filed as #9039

```ts
const u = new Uint8Array([7, 8, 9]);
for (let i = 0; i < 3; i++) { const x = u[99]; a.push(x); }
// node: [null,null,null]   perry: [0,0,0]
```

An out-of-bounds read pushed into an array stores `0` instead of `undefined`. The read itself is correct (`u[99]` prints `undefined`), so the coercion is on the store lane. A/B'd against `main` @ `6ac0a0c2bf` before attributing — identical there, so this PR neither causes nor worsens it. Worth knowing here because it lives in the lane you are changing, and because the neighbouring non-numeric-key case is handled correctly, which is what makes the gap easy to miss.

Renumbered the fragment `0000-` → `9033-`; the placeholder is a hard failure in the changeset gate since #9010.

Validation: `perry-runtime --lib` 2805/0, `perry-codegen` 1343/0, `fmt --check`, `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**.

PR #9036, opened 2026-08-29
perf(hir): keep the array type through methods chained on new Array<T>(n)

proggeramlug:
## What

`new Array<T>(n)` infers as `Generic { base: "Array", type_args: [T] }` (the explicit-type-args early return in `infer_type_from_expr`), but the builtin array method-return table matched only `Type::Array`. A chained receiver-returning method — **`new Array<number>(n).fill(0)`, the canonical preallocation idiom** — therefore inferred `Any`: the binding lost its array type and every subsequent `a[i] = v` on it took the generic feedback store instead of the inline guarded lane, for the array's whole life.

## Fix

Rebind `Generic { base: "Array", [T] } -> Type::Array(T)` right after the receiver type is inferred, before the method tables — the same normalization every downstream consumer already applies (`typed_parse.rs`, `stmt_loops.rs`, `element_shape_loop.rs`).

## Numbers (isolated store loop over such an array, dev box)

|---|---|---|---|
| `const a = new Array<number>(8192).fill(0)` + store loop | 34.6 ns | **3.4 ns** | 0.7 |
| same code with explicit `: number[]` annotation | 3.7 ns | 3.7 ns | 0.7 |

The discriminator that found it: identical code ± the annotation.

## Interaction note

Complementary to #9033: this fix makes more receivers statically numeric; without #9033's `LocalGet` canonical-value arm those would have routed pushes into the slower guarded tier. Order-independent now that both exist.

## Tests

3 new HIR regression tests (`chained_array_ctor_method_types.rs`) pinning `number[]` through `.fill`, a double chain (`.fill().reverse()`), and element-type extraction (`.at()`). Full gate on the build host: suites green (one known-flaky parallel-only GC stack_maps test passes in isolation).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved type inference for chained methods on typed array constructors.
  * Array-returning chains such as `.fill().reverse()` now preserve the correct element type.
  * Element access methods such as `.at()` now infer the correct element type.

* **Tests**
  * Added coverage for single and multi-step array method chains.

### DIFF
--- a/changelog.d/9036-chained-array-ctor-type.md
+++ b/changelog.d/9036-chained-array-ctor-type.md
@@ -1,0 +1,1 @@
+Kept the inferred array type through builtin methods chained on `new Array<T>(n)` — `new Array<number>(n).fill(0)` now infers `number[]` instead of `Any`, so index stores on such arrays take the inline guarded lane (34.6 → 3.4 ns per store).
--- a/crates/perry-hir/src/lower_types.rs
+++ b/crates/perry-hir/src/lower_types.rs
@@ -1236,6 +1236,25 @@ pub(crate) fn infer_call_return_type(callee: &ast::Expr, ctx: &LoweringContext) 
                     return Type::Any;
                 }
                 let obj_ty = infer_type_from_expr(&member.obj, ctx);
+                // `new Array<T>(n)` types as `Generic { base: "Array", [T] }`
+                // (the explicit-type-args early return above) — the same array
+                // `T[]` denotes, and downstream consumers already treat the
+                // two as one (`typed_parse.rs`, `stmt_loops.rs`,
+                // `element_shape_loop.rs`). This table must too: without the
+                // rebind, `new Array<number>(n).fill(0)` inferred `Any`, the
+                // binding lost its array type, and every later `a[i] = v` on
+                // it took the generic feedback store instead of the inline
+                // guarded lane — 34.6 vs 3.7 ns per store, for the array's
+                // whole life.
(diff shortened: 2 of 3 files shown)


proggeramlug:
Merged, as-is — the only one of today's three that needed nothing fixed. Correctly-numbered fragment, and real HIR regression tests rather than a probe.

The normalization is consistent with what three consumers already do (`typed_parse.rs`, `stmt_loops.rs`, `element_shape_loop.rs`), so this closes a gap rather than introducing a new convention — the method-return table was the one place still matching only `Type::Array`.

Probed the behaviour end-to-end against node 26.5.1 alongside the type-level tests: the `.fill(0)` preallocation idiom with a store loop, a double receiver-returning chain, `.at(0)` returning the element type, `string` element type, `map`/`filter` off the chain, nested `number[][]`, sparse `new Array<number>(3)` without `fill` (holes must stay holes — `0 in f` is `false`), and the annotation-equivalent `const d: number[] = new Array(8).fill(0)` producing identical output. **Byte-identical.**

The discriminator you used to find this — identical code with and without the `: number[]` annotation, 34.6 vs 3.7 ns — is the part worth keeping in mind: an annotation making code *faster* means an inference path is dropping a type, and the same shape is worth checking wherever else the builtin tables match on `Type::Array` alone.

Validation: 3 new HIR tests pass, `perry-hir` lib 355/0, `perry-runtime --lib` 2805/0, `fmt --check`, `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**.

PR #9038, opened 2026-08-29
feat(runtime): tombstone object deletes default-on (populated delete 6.5x)

proggeramlug:
Flips #9029's flag to ON; `PERRY_OBJECT_TOMBSTONES=0` becomes the kill switch (the moving-scavenge rollout pattern). No mechanism changes — this PR is the flip, the changelog, and the validation ladder.

**Why it's ready:** the default-on prerequisites shipped inside #9029 itself — the 57-file walker audit (four flag-on-only bugs found and fixed: JSON shape-template SIGSEGV, phantom worker-thread `""` keys, phantom diagnostics `"undefined"` prop, and the hole-count accounting reset that let delete/re-add churn dodge the squeeze bound), the 60-cycle churn test pinning the 2x-live-size memory bound, and the differential battery byte-identical to node in both flag states.

**Numbers** (shared Linux box, interleaved min-of-7 vs exact main tip): populated delete 2030 → ~315 ms (**6.5x**; ~15x node from ~96x pre-campaign); combined overwrite 27 → 27; realistic-name read 17 → 17. Flag-off deletes keep #9029's −11%.

**Validation on this branch:** full runtime suite (2805) runs with the new default, so every delete-touching test now exercises tombstones; fmt/census/class-id/file-size green; server re-run of all six fixtures in BOTH directions (default-on and kill-switch) vs node refs posted below.

## Summary by CodeRabbit

- **New Features**
  - Object key deletion now uses fast tombstone-based removal by default, with automatic compaction to control memory usage.
  - Added an environment option to disable tombstone deletes when needed.

- **Bug Fixes**
  - Deleted properties are no longer exposed as phantom `undefined` or empty-string keys during serialization, structured cloning, diagnostics, or JSON processing.
  - Improved handling of repeated delete-and-readd operations to preserve memory bounds.

- **Performance**
  - Populated-object deletion benchmarks improve by approximately 6.5× with tombstones enabled.


proggeramlug:
**Heads-up on a merge race, now recovered on this branch.** #9029 was squash-merged from the branch state *before* my second-round push, so main is missing the walker-audit round: the JSON shape-template **SIGSEGV** fix, the worker-thread phantom-`""`-key fix, the diagnostics phantom-prop fix, the **hole-count accounting carry** (without it, delete/re-add churn regrows the keys array unbounded — main's flag-on popdel measures ~415 ms instead of ~315 for exactly this reason), the three pinning unit tests, and the file-size/census gate fixups.

Those three commits are now cherry-picked onto this branch (`d85a1ac`, `d9f4345`, `454b0c6`), so merging this PR delivers both the audit round and the default flip. 2807 tests green with the default on; fmt/size/census/class-id green; fresh server numbers on the recovered tip incoming below.

If you'd rather land the audit fixes separately from the flip, say so and I'll split them out — but note the SIGSEGV fix is live-relevant even flag-off-by-default today, since `=1` users hit it.

proggeramlug:
Recovered-tip validation complete (server, interleaved min-of-7 vs a clean build of current main):

| | current main (default off) | this branch (default on) |
|---|---|---|
| populated delete | 1793-1843 ms | **311-321 ms (5.7×)** |

All six fixtures byte-identical between the new default and the `PERRY_OBJECT_TOMBSTONES=0` kill switch, AND to node. The ~315 confirms the hole-count accounting fix is back in (main without it measures ~415 — the keys array silently regrows unbounded under delete/re-add churn, which is the memory-bound violation, not just perf). 2807 tests with the default on; all lint gates green. Ready to merge.

proggeramlug:
CI note: the `cargo-test` red is a **main-side** breakage, not this branch — `codegen_env_vars_are_build_cache_inputs` fails on current main because `PERRY_BOX_CAPTURE_ENTRY_CELLS` (#9026) and `PERRY_GUARDED_PREINLINE_MAX_IR_BYTES` were never registered as build-cache inputs. Fix is **#9044** (registers both; every open PR is red on this until it lands). The earlier gap-suite ✗ rows were the auto-cancel from my force-push; the re-run on the current head is green so far.

proggeramlug:
Merged. Flipping a default that changes object-delete semantics globally is the highest-stakes thing in this series, so I checked the readiness claim rather than the mechanism (which I audited in #9029).

**One correction to the description.** It says "the default-on prerequisites shipped inside #9029 itself — the 57-file walker audit (four flag-on-only bugs found and fixed…)". They did not: the JSON shape-template fix, the `v8_serde` fix and the `diagnostics.rs` fix are **+lines in this PR** (+9, +9, +7). That matters because the sentence is doing load-bearing work in the readiness argument — those fixes are new here, not soaked behind a merged flag. It does not change my verdict, since they are testable and I tested them, but anyone bisecting a tombstone regression should know the walker fixes date from this PR and not #9029.

Verified independently rather than taking the audit's word for it:

- the 13-surface enumeration probe from my #9029 review — `keys`/`values`/`entries`/`getOwnPropertyNames`/`stringify`/`for-in`/spread/rest, re-add ordering, squeeze-threshold crossing, 40-key overflow object, `in`/descriptors/absent-delete
- a new probe aimed at the three walker bugs specifically: repeated `JSON.stringify` of one shape after deletes (the template-reuse path that SIGSEGV'd), arrays of holey objects, a scan for phantom `""` keys, a scan for a phantom `"undefined"` property, 60 rounds of delete/re-add churn against the hole-accounting reset, and a nested holey object

**Byte-identical to node in BOTH directions** — default-on and `PERRY_OBJECT_TOMBSTONES=0` — so the kill switch is a real off state, and the flag is correctly inverted (`!matches!(…, "0" | "off" | "false")`).

The `thread.rs -156` is a relocation, not a loss: all five `#6185` transfer-guard tests moved to `thread_transfer_guard_tests.rs` with identical names, and they pass there (5/5).

The strongest signal is that the full runtime suite now runs with tombstones live — **2807 passed / 0 failed**, so every delete-touching test exercises the new path rather than the flag-off one.

**Renumbered the fragment** `9037-` → `9038-`.

`fmt --check`, census, file-size and the rest: `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**.

PR #9040, opened 2026-08-29
feat(hir): supply native parsers for static Zod schemas

proggeramlug:
## Summary

AOT-compiles a conservative static subset of Zod `z.compile()` while keeping Zod?not Perry?as the owner of schema semantics.

Perry recognizes a closed `z.object({...})`, emits only a native parser, and passes that parser to the proposed Zod `compileFromParser()` API. Zod's own TypeScript remains responsible for cloning, async/backward/skip-check bypasses, recursion guards, public parse methods, validation, and authoritative runtime fallback.

This remains a draft because it depends on the upstream Zod integration API proposed in [colinhacks/zod#6498](https://github.com/colinhacks/zod/issues/6498).

## Architecture

```text
static schema source
       |
       v
Perry schema IR -> native parser (value | z.INVALID)
                              |
                              v
                 z.compileFromParser(schema, parser)
                              |
                              v
         Zod-owned clone / wrapper / fallback / public API
```

There is no Perry reimplementation of Zod's compiled wrapper contract. Unsupported schemas and Zod versions without `compileFromParser()` retain the ordinary `z.compile(schema, options)` call.

## Changes

- Extract compiler-only IR from immutable static `z.object({...})` declarations imported from `zod` / `zod/v4`.
- Lower string, finite-number, boolean, and number `int` / literal `min` / literal `max` checks to native Perry HIR.
- Return Zod's exported `INVALID` sentinel from the native parser and install it through Zod's `compileFromParser()`.
- Decline mutable schemas, shadowed Zod namespace bindings, duplicate keys, `__proto__`, unsupported schemas, and dynamic options rather than partially specializing them.
- Let `compilePackages` enter a TypeScript source tree even when the package's built exports target is absent. This makes the temporary source-only Zod pin reproducible.
- Pin the POC to the public source-only Zod commit `ee27e77ad86238199cfa0c262cfa2c7dfc125e49` and opt Zod into `perry.compilePackages`.
- Add HIR/resolver regressions, a conformance fixture, a real-Zod probe, and a changelog fragment.

## Current supported subset

- `import * as z from 'zod' or 'zod/v4'`
- immutable `const` schema bindings
- `z.object()` with static, unique, non-computed fields
- `z.string()`, `z.number()`, and `z.boolean()`
- `z.number().int()`, `.min(<number literal>)`, and `.max(<number literal>)`
- `z.compile(schema)` and static `{ strict: <boolean> }`

Everything else remains on Zod's normal runtime path.

## Upstream dependency / landing decision

The Zod-side API implementation is available at:

- issue: [colinhacks/zod#6498](https://github.com/colinhacks/zod/issues/6498)
- proposed branch comparison: [proggeramlug:zod:feat/aot-parser-install-api](https://github.com/colinhacks/zod/compare/main...proggeramlug:zod:feat/aot-parser-install-api?expand=1)

Before this Perry PR leaves draft, we should agree that:

1. Zod accepts an API equivalent to `compileFromParser(schema, parser, { validator? })` plus an exported failure sentinel.
2. The final Perry dependency uses the first official Zod release containing that API; the fork commit is only for this draft POC.
3. Perry continues to fail closed: only exact statically proven shapes specialize, and all other calls remain Zod-owned.

## Test plan

- [x] `npm ci --ignore-scripts --dry-run`
- [x] `cargo test -p perry-hir --lib` (360 passed, 1 pre-existing ignored before rebase)
- [x] `cargo test -p perry-hir zod_aot --lib` after rebasing onto current `main` (5 passed)
- [x] `cargo test -p perry --bin perry compile_package_resolves_source_without_built_exports_target`
- [x] `cargo build --profile perry-dev -p perry`
- [x] End-to-end compile after rebase: 95 Zod source modules native, 0 JavaScript
- [x] Auto-optimized runtime features: `async-runtime,web-fetch` (no `dyn-eval`)
- [x] 27.7 MB native executable; valid input avoids fallback, unknown keys strip, invalid input delegates to Zod
- [x] Wrapper `__originalRun` and `bag.fallbackRun` identities prove installation is Zod-owned
- [x] Zod targeted compile tests: 462 passed across all configured projects
- [x] Zod full compile suite: 2645 passed; 2 unrelated existing Windows esbuild path-escaping failures
- [x] Zod package build (`zshy`) passes
- [ ] Full Perry workspace CI

## Verified output

```text
true
true
true
true
true
true
true
false
true
score:invalid_type
```

The assertions cover schema cloning, replacement of `_zod.run`, Zod ownership of the wrapper and fallback, native valid parsing without fallback, unknown-key stripping, invalid fallback, and the final Zod issue.

## Checklist

- [x] No workspace version bump or direct `CHANGELOG.md` edit
- [x] Commits follow the repository prefix convention

### DIFF
--- a/changelog.d/9040-zod-compile-aot.md
+++ b/changelog.d/9040-zod-compile-aot.md
@@ -3,6 +3,11 @@
 - **Static Zod object schemas can now take a native `z.compile()` path.** Perry
   recognizes supported `z.object({...})` declarations and lowers string,
   finite-number, boolean, integer, and literal numeric-bound checks directly
-  to HIR. Successful parses no longer require Zod's `new Function` fast path;
-  failed checks delegate to the original Zod parser so its issue construction
-  remains authoritative. Unsupported schemas keep the regular Zod call.
+  to HIR. Perry supplies only the generated parser through Zod's
+  `compileFromParser()` integration API; Zod continues to own schema cloning,
+  parse-context bypasses, public methods, validation, and runtime fallbacks.
+  Successful parses no longer require `new Function`, while failed checks and
+  unsupported schemas remain authoritative in Zod.
+- `perry.compilePackages` can now resolve a package's TypeScript source tree
+  even when its published JavaScript export target is absent, enabling
+  reproducible source-only package pins for native compilation.
--- a/changelog.d/9040-zod-compile-aot.md
+++ b/changelog.d/9040-zod-compile-aot.md
@@ -1,0 +1,8 @@
+### Added
+
+- **Static Zod object schemas can now take a native `z.compile()` path.** Perry
+  recognizes supported `z.object({...})` declarations and lowers string,
+  finite-number, boolean, integer, and literal numeric-bound checks directly
+  to HIR. Successful parses no longer require Zod's `new Function` fast path;
+  failed checks delegate to the original Zod parser so its issue construction
+  remains authoritative. Unsupported schemas keep the regular Zod call.
--- a/crates/perry-hir/src/lower/context.rs
+++ b/crates/perry-hir/src/lower/context.rs
@@ -183,6 +183,7 @@ impl LoweringContext {
             namespace_import_locals: HashSet::new(),
             fetch_call_response_locals: HashSet::new(),
             namespace_import_sources: std::collections::HashMap::new(),
+            zod_schema_irs: HashMap::new(),
             generator_func_names: HashSet::new(),
             async_generator_func_names: HashSet::new(),
             nested_generator_forward_referenced: HashSet::new(),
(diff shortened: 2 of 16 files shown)


proggeramlug:
Closing as not doing it this way for now.

colinhacks:
The zod side of this landed as `z.withParser(schema, parser)` in colinhacks/zod#6575 — different name from `compileFromParser`, same shape. The parser returns the parsed value or `INVALID`; Zod keeps the clone, the async/backward/skip-check bypasses, the recursion refusal and the runtime fallback.

It's a separate export rather than an option on `compile()`, so `typeof zod.withParser === "function"` works as the feature test. Both `zod` and `zod/mini` export `INVALID`.

On main, not released yet. Flagging it in case you pick this back up — no expectations either way.

Issue #9039, opened 2026-08-29
Out-of-bounds Uint8Array read pushed into an array stores 0 instead of undefined


proggeramlug:
An out-of-bounds `Uint8Array` read that flows into an array push stores **`0`** instead of `undefined`.

## Reproducer

```ts
const u = new Uint8Array([7, 8, 9]);
const a: number[] = [];
for (let i = 0; i < 3; i++) { const x = u[99]; a.push(x); }
console.log(JSON.stringify(a), u[99], typeof u[99]);
```

```
node:  [null,null,null] undefined undefined
perry: [0,0,0]          undefined undefined
```

The read itself is right — `u[99]` prints `undefined` and `typeof` says `undefined`. Only the value that reaches the array is wrong, so the coercion happens on the store/push lane rather than in the typed-array read.

## Scope

- **out-of-bounds numeric index** → wrong (`0` for `undefined`)
- **non-numeric key** (`u["oops"]`) → correct, `[null,null,null]`
- in-bounds reads → correct

That split is the interesting part: the `#7700` non-numeric-key path is handled, and this is the neighbouring case that isn't.

## Pre-existing, not #9033

Found while auditing #9033 (which extends `expr_produces_canonical_raw_f64` to integer-provenance locals — the same push lane), so I A/B'd before attributing. Both `--profile perry-dev` builds of `-p perry -p perry-runtime-static -p perry-stdlib-static`, compiled with `PERRY_NO_AUTO_OPTIMIZE=1`:

| build | result |
|---|---|
| #9033 branch | `[0,0,0]` |
| `main` @ `6ac0a0c2bf` | `[0,0,0]` |

The two are byte-identical across my whole 7-case probe, so #9033 neither causes nor worsens this. Recording it explicitly because it sits squarely in the lane that PR touches and would be easy to pin on it.

## Why it is worth fixing

An OOB typed-array read is a normal way to reach the end of a buffer, and `0` is a *plausible* byte value — so unlike a crash this silently produces data that looks legitimate:

```ts
for (let i = 0; i < n; i++) out.push(buf[i]);   // n past the end → silent zero padding
```

Node oracle is the pinned v26.5.1.

PR #9042, opened 2026-08-29
feat(gc): PERRY_GC_DIAG reports cumulative GC time and wall-clock share

proggeramlug:
Adds the measurement the concurrent-GC decision gates on (agreed direction: measure GC share on real workloads first — cc bundle, pi — and only write a concurrent-marking design doc if it's 15%+).

`PERRY_GC_DIAG=1` now prints at exit:

```
[gc-time] wall_us=172094 step_us=0 remark_us=0 minor_us=23626 full_sync_us=0 share_permille=137
```

- Cumulative counters for the three disjoint pause buckets (budgeted steps, atomic final remarks, copying minors) plus synchronous `gc()` wall time reported separately (it can internally drive budgeted steps, so it's not summed into the share).
- Always-on relaxed atomics — one `fetch_add` per GC **event**, noise next to the event itself; no new env knob.
- The wall epoch is primed in `js_gc_init`: it initialized lazily on first read, which would have silently started the denominator at the first GC event instead of program start.

973 gc:: tests green, fmt/gc-knob gates green. Runtime-only.

## Summary by CodeRabbit

- **New Features**
  - Added an opt-in GC diagnostic (`PERRY_GC_DIAG=1`) that reports cumulative garbage-collection time and its share of elapsed runtime.
  - Diagnostics include timings for incremental steps, final remarks, copying minor collections, and synchronous full collections.
- **Documentation**
  - Added changelog documentation describing the new diagnostic output and reported metrics.

### DIFF
--- a/changelog.d/9039-gc-time-share-diag.md
+++ b/changelog.d/9039-gc-time-share-diag.md
@@ -1,0 +1,16 @@
+`PERRY_GC_DIAG=1` now reports cumulative mutator-visible collection time and
+its wall-clock share at exit:
+
+    [gc-time] wall_us=… step_us=… remark_us=… minor_us=… full_sync_us=… share_permille=…
+
+The existing telemetry answers "how bad is the worst pause" (`step_max_us`,
+`final_remark_max_us`); this answers "what fraction of the run went to
+collection at all" — the measurement-first gate for any concurrent-marking
+work: if the share on real applications is single-digit percent, moving
+marking off the mutator thread has nothing worth hiding. Counters are
+always-on relaxed atomics (one `fetch_add` per GC event); the wall epoch is
+primed in `js_gc_init` so the denominator starts at program start, not at
+the first GC event. `share_permille` sums the three disjoint buckets
+(budgeted steps, atomic final remarks, copying minors); synchronous
+`gc()` calls are timed separately since they can internally drive budgeted
+steps.
--- a/crates/perry-runtime/src/gc/copying.rs
+++ b/crates/perry-runtime/src/gc/copying.rs
@@ -1677,6 +1677,7 @@ pub(super) fn run_copied_minor_attempt(
     // #7592: this is the promotion the survivor-promotion handoff exists to
     // enable, so it releases the latch that suppressed a repeat handoff.
     note_copying_minor_completed();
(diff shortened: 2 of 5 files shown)


proggeramlug:
Merged. Diagnostic-only, and I checked the thing that matters for an instrument: that it actually emits, and that its arithmetic is right — a counter feature that prints nothing passes every test.

On an allocation-heavy load it reports:

```
[gc-time] wall_us=18197 step_us=0 remark_us=0 minor_us=5085 full_sync_us=0 share_permille=279
```

`5085 / 18197 = 0.2794`, matching `share_permille=279`, so the denominator really is priming at program start rather than at the first GC event — which is the bug the `js_gc_init` change fixes and the one that would have quietly inflated every share it reports. Nothing extra prints with the knob off.

(That 27.9% is a synthetic churn loop, not a signal about the concurrent-marking decision — the workloads named in the description, cc bundle and pi, are the ones that answer that.)

The always-on atomics are justified at this granularity: one relaxed `fetch_add` per GC *event*, not per allocation, so it is genuinely noise beside the event. They are `AtomicU64` counters rather than pointer-bearing statics, so they need no scanner registration and `gc_runtime_root_holders.py` stays green.

I also like that the double-counting is documented rather than smoothed over — `full_sync_us` is reported separately precisely because a forced full can internally drive budgeted steps, so summing all four would over-report. An instrument that quietly summed them would be worse than useless for the decision it exists to inform.

**Renumbered the fragment** `9039-` → `9042-`: #9039 is an issue I filed earlier today (out-of-bounds `Uint8Array` read pushed as `0`), not this PR.

Validation: `gc::` 973/0, `perry-runtime --lib` 2805/0, `fmt --check`, `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**.

proggeramlug:
**Correction to my merge comment above: the fragment renumber did not land.**

I said I had renumbered `9039-` → `9042-`. I staged it, but pushed to `origin` when this PR is fork-based, so it never reached the head — and I merged anyway, because my head-assert and the merge ran in the same command batch, so the assert reported the mismatch without gating anything.

Fixed on `main` in #9046. The audit findings in my comment above stand; only the claim about the rename was wrong.

PR #9044, opened 2026-08-29
fix(build-cache): register the two unkeyed codegen env vars (main CI is red)

proggeramlug:
`codegen_env_vars_are_build_cache_inputs` fails on current main — `PERRY_BOX_CAPTURE_ENTRY_CELLS` (#9026) and `PERRY_GUARDED_PREINLINE_MAX_IR_BYTES` landed without build-cache registration, so **every open PR's cargo-test job is red** (that's how I found it, on #9042). Both change emitted code, so they're cache inputs, not exclusions — same rationale as the RS4GC/TRE budgets beside them. The four build_cache tests pass locally with the registration.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved build-cache accuracy when code-generation settings change.
  - Prevented stale cached builds by ensuring relevant compilation options trigger fresh builds.
  - Resolved test and build failures related to untracked code-generation settings.

- **Documentation**
  - Added changelog information describing the updated build-cache behavior.

### DIFF
--- a/changelog.d/9044-build-cache-codegen-env-vars.md
+++ b/changelog.d/9044-build-cache-codegen-env-vars.md
@@ -1,0 +1,14 @@
+`PERRY_BOX_CAPTURE_ENTRY_CELLS` and `PERRY_GUARDED_PREINLINE_MAX_IR_BYTES` are
+now build-cache inputs.
+
+Both landed without build-cache registration, so
+`codegen_env_vars_are_build_cache_inputs` failed on `main` — and because it is a
+`perry` bin-crate unit test, the whole `perry` test binary failed to compile,
+turning every open PR's cargo-test job red.
+
+Both are inputs rather than exclusions because both change emitted code: the
+capture-cell knob changes every closure body that qualifies for entry-resolved
+box cells, and the preinline ceiling changes which functions inline, so a run
+with a raised ceiling must not be served objects a default run produced. That
+is #6394's rule — a codegen env var keys the cache or carries a written reason
+it cannot affect output.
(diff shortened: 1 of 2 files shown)


proggeramlug:
Merged — and this is fallout from two of my own merges, so thank you for catching it.

Confirmed the breakage on `main` @ `ec0b29e9df` before merging:

```
codegen_env_vars_are_build_cache_inputs panicked:
  these codegen env vars key neither the build cache nor an exclusion (#6394's rule):
  ["PERRY_BOX_CAPTURE_ENTRY_CELLS", "PERRY_GUARDED_PREINLINE_MAX_IR_BYTES"]
```

`PERRY_BOX_CAPTURE_ENTRY_CELLS` came in with #9026 and `PERRY_GUARDED_PREINLINE_MAX_IR_BYTES` with #9016 — both of which I reviewed and merged, and neither of which I caught this on.

**Input rather than exclusion is the right call for both**, and the preinline reason states the actual hazard: "a run with a raised ceiling must not be served objects a default run produced". Registering either as an exclusion would let a cached build survive a knob flip and hand back objects built under the other setting.

Worth noting the blast radius, because it explains why this was worth a dedicated PR: the assertion lives in a **bin-crate unit test**, so its failure doesn't just fail one test — the `perry` test binary fails to compile, which is why every open PR's cargo-test job went red rather than just `main`'s.

## The gap on my side

My per-PR validation runs `-p perry-runtime --lib`, `-p perry-codegen --lib` and `run_lint_gates.sh`. None of those reach `-p perry --bins`, which is where this assertion lives — so a PR adding a codegen env var passes everything I check and still reddens CI. I've added `cargo test -p perry --bins` to what I run, and it's the specific check to reach for on any PR that introduces a `PERRY_*` knob touching codegen.

Verification: `build_cache` 4/4, **`perry --bins` 1061/0** (against a `main` where that binary does not compile), `perry-runtime --lib` 2807/0, `fmt --check`, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped.

PR #9047, opened 2026-08-29
fix(telemetry): honor first-run opt-out globally

proggeramlug:
## Summary

- treat the first-run telemetry decision as the master consent gate
- disable compatibility reports when consent is missing or declined, including legacy inconsistent configs
- re-check consent and environment opt-outs at both Chirp network boundaries
- document the unified consent behavior

## Verification

- cargo test -p perry --bin perry --no-default-features telemetry::tests -- --test-threads=1
- cargo test -p perry --bin perry --no-default-features compat_reports::tests -- --test-threads=1
- cargo fmt --all -- --check
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Telemetry consent now provides a single master control for all telemetry channels.
  * Added support for disabling telemetry with `PERRY_NO_TELEMETRY=1`; CI environments suppress telemetry automatically.
  * Compatibility reports are sent only when master telemetry consent is enabled.

* **Bug Fixes**
  * Telemetry and compatibility reporting now recheck consent before sending, preventing reports after opt-out.
  * Declined consent now disables compatibility reports by default.

* **Documentation**
  * Updated telemetry guidance to clarify consent, opt-out behavior, and configuration priorities.

### DIFF
--- a/changelog.d/9047-telemetry-master-opt-out.md
+++ b/changelog.d/9047-telemetry-master-opt-out.md
@@ -1,0 +1,10 @@
+### Fixed
+
+- Treat the first-run telemetry decision as the master consent gate. Declining
+  it, leaving it unanswered, setting `PERRY_NO_TELEMETRY=1`, or running in CI
+  now prevents generic usage events, beta-error reports, and compatibility
+  reports from being sent.
+
+- Fail closed for existing configs that have `telemetry.enabled = false` but
+  `compatibility_reports = "on"`, and re-check consent immediately before each
+  Chirp request so an opt-out made while an event is queued still wins.
--- a/crates/perry/src/compat_reports.rs
+++ b/crates/perry/src/compat_reports.rs
@@ -105,15 +105,13 @@ impl ReportSink for QueueSink {
     }
 }
 
-/// Read the active `compatibility_reports` mode for this process,
-/// honouring env-level overrides (`PERRY_NO_TELEMETRY=1`, `CI=true`).
+/// Read the active `compatibility_reports` mode for this process, honouring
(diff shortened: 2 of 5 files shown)


proggeramlug:
Merged. For a privacy change the only question that matters is whether the gate covers *every* outbound path, so I enumerated them rather than reading the summary.

There are exactly two network sites in this surface, and both are gated at the last moment before the client is built:

- `telemetry.rs::send_event_blocking` — `if !is_telemetry_enabled() { return; }` immediately before the `reqwest` client and the `POST` to `CHIRP_URL`
- `compat_reports.rs::send_compat_report` — gated on entry **and re-checked** right before the client build

That second check is the part worth calling out: it means consent flipping between enqueue and send cannot let a queued report escape, which a single entry-point gate would have missed. Together with `active_telemetry_config()` funnelling through `apply_master_consent`, there is one decision point rather than a set of parallel ones that can drift.

The `master_opt_out_rejects_even_an_enabled_compatibility_channel` test is the right one to have written: a legacy config with `compatibility_reports = "on"` but master consent declined is exactly the inconsistent state a user would reasonably expect to be resolved in favour of "off", and it is the state a naive per-channel read would get wrong.

Validation: `telemetry::` 4/4, `compat_reports::` 3/3, **`perry --bins` 1066/0**, `fmt --check`, `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**. Fragment already correctly numbered.

(`perry --bins` is a check I added to my set after #9044 — this PR lives entirely in the `perry` bin crate, which none of my previous per-PR commands reached.)

PR #9049, opened 2026-08-29
fix(codegen): native IR reader accepts value-returning inline asm (macOS ARM64 app builds die without it)

proggeramlug:
The native IR reader handled only the **void** inline-asm barrier; a value-returning form fell through to the ordinary-call callee search and failed with `call without callee`. The hot-TLS thread-pointer read emits exactly that form —

```llvm
%r = call i64 asm sideeffect "mrs $0, tpidrro_el0", "=r"() "gc-leaf-function"
```

— so **any function that takes the inline allocation fast path fails native IR construction on macOS aarch64**, killing the whole build. pi's 13MB bundle dies in its `__rest` helper; the textual pipeline compiles the same IR fine, which is why the suites never saw it.

Two parsing subtleties, both pinned by the new reader test:
- the arg list is found after the **4th quote** (the constraints' closing quote), not the last quote in the line — the emitter appends a quoted `"gc-leaf-function"` attribute after the args, which the first version of this fix tripped over;
- the callsite is rebuilt with the same structural `gc-leaf-function` marking as the void barrier (#8121 — perry-emitted asm never re-enters the runtime, so RS4GC must not statepoint-wrap it).

Reader test round-trips both forms (with the trailing attribute exactly as emitted) through the verifier. perry-codegen lib 1344/0. Verified end-to-end: pi's bundle now compiles past the previously-fatal unit.

## Summary by CodeRabbit

* **New Features**
  * Added support for value-returning inline assembly calls when reading LLVM IR.
  * Inline assembly results are now correctly typed and returned for downstream use.

* **Bug Fixes**
  * Fixed failures when processing LLVM IR containing inline assembly that returns a value.
  * Added coverage for value-returning assembly and assembly barriers.

### DIFF
--- a/changelog.d/9049-native-reader-value-asm.md
+++ b/changelog.d/9049-native-reader-value-asm.md
@@ -1,0 +1,22 @@
+The native IR reader accepts value-returning inline asm.
+
+It handled only the void inline-asm barrier; the value-returning form fell
+through to the ordinary-call callee search and failed with `call without
+callee`. The hot-TLS thread-pointer read emits exactly that form —
+
+    %r = call i64 asm sideeffect "mrs $0, tpidrro_el0", "=r"() "gc-leaf-function"
+
+— so on macOS aarch64 any function taking the inline allocation fast path failed
+native IR construction, killing the build. The textual transport compiles the
+same IR fine, and small modules use it by default, which is why no suite saw
+this: only large codegen-unit-split modules (or an explicit
+`PERRY_LLVM_INPROCESS=native`) reach the reader. That is the dialect reader's
+closed-set design working as intended — a new emission form fails loudly at
+construction — but the form had no arm.
+
+Two parsing details the reader test pins: the arg list is found after the
+CONSTRAINTS' closing quote (the 4th quote), not the last quote on the line,
+because the emitter appends a quoted `"gc-leaf-function"` attribute after the
+args; and the rebuilt callsite carries the same structural `gc-leaf-function`
+marking as the void barrier, since perry-emitted asm never re-enters the runtime
+and RS4GC must not statepoint-wrap it (#8121).
--- a/crates/perry-codegen/src/dialect/mod.rs
+++ b/crates/perry-codegen/src/dialect/mod.rs
@@ -793,6 +793,17 @@ impl<'ctx, 'm> FnReader<'ctx, 'm> {
             return self.call_asm(asm_rest).map(|_| None);
         }
 
+        // Value-returning inline asm: `TY asm sideeffect "ASM", "CONSTR"(ARGS)`
+        // — the hot-TLS thread-pointer read (`mrs $0, tpidrro_el0`). The type
+        // token before ` asm ` must be a bare type, not a callee or signature,
+        // or an ordinary call to a function whose name contains "asm" would be
+        // misparsed.
+        if let Some((ty_tok, asm_rest)) = rest.split_once(" asm ") {
+            if !ty_tok.contains(['@', '%', '(']) {
+                return self.call_asm_value(dst, ty_tok.trim(), asm_rest).map(Some);
+            }
+        }
+
         // #8175: explicit calling-convention token before the return type.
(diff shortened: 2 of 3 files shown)


proggeramlug:
Merged. I reproduced the failure and the fix rather than reviewing the diff alone, because "the suites never saw it" is the part that makes this class dangerous.

**Reproduced on `main`** with a four-line file — one function allocating an object in a loop:

```
native IR construction failed in @perry_fn_asmrepro_ts__run:
  in line: %r17 = call i64 asm sideeffect "mrs $0, tpidrro_el0", "=r"() "gc-leaf-function"
```

The same file compiles fine on the text transport, which is exactly why nothing caught it: `linker.rs` says small modules keep the text transport unless `=native` is explicit, and only large codegen-unit-split modules default to the native reader. So the reproducer needs `PERRY_LLVM_INPROCESS=native` — worth knowing, because it means *any* future reader gap is invisible to ordinary small-module compiles too.

**With the fix**, the same file builds through the native reader and all three arms agree (`2500`): native reader, text transport, and node. `PERRY_LLVM_INPROCESS=diff` — which builds both arms and compares — also passes, and a heavier multi-function probe through the native reader matches node.

The two things I most wanted to check both hold:

- **`gc-leaf-function` is preserved**, and identically to the void barrier: `AttributeLoc::Function` with the same `create_string_attribute("gc-leaf-function", "")`. Dropping it would have RS4GC statepoint-wrapping the asm, which is #8121's SIGBUS.
- **The `ty_tok` guard is real.** Rejecting a token containing `@`, `%` or `(` before treating ` asm ` as the split point is what stops an ordinary call to a function whose name contains "asm" from being misparsed as inline asm — a nice touch, since `split_once(" asm ")` alone would be a trap.

The 4th-quote detail is the kind of thing that would have been a silent mis-parse rather than a loud one, so pinning it in the reader test with the trailing attribute *exactly as emitted* is the right call.

Validation: dialect tests 9/9, `perry-codegen` lib 1344/0, `perry-runtime --lib` 2807/0, `perry --bins` 1061/0, `fmt --check`, `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**.

proggeramlug:
Closing as **superseded by #9055** ("fix(codegen): parse typed inline assembly"), which landed while I was validating this and fixes the same defect. This is a duplicate rather than a rejection — the diagnosis here was right, and I verified the fix works.

I confirmed the supersession by reproduction rather than by reading titles. My four-line reproducer (one function allocating an object in a loop) fails on the `main` this PR was branched from:

```
native IR construction failed in @perry_fn_asmrepro_ts__run:
  in line: %r17 = call i64 asm sideeffect "mrs $0, tpidrro_el0", "=r"() "gc-leaf-function"
```

and compiles and runs correctly (`2500`) on current `main` @ `fbd7a94d08` with `PERRY_LLVM_INPROCESS=native`.

**#9055's implementation is the stronger of the two**, which is why I'd keep it rather than resolve the conflict in this PR's favour:

- it handles the void and value-returning forms in **one** `call_asm` rather than a second parallel function, so the two cannot drift;
- it parses the quoted operands with `take_quoted` instead of arithmetic on the 4th quote. That matters — the 4th-quote approach here works for the emitter's current output but breaks if an asm template ever contains an escaped quote, whereas structural tokenization does not;
- it preserves `gc-leaf-function` (and tracks `has_gc_leaf` from the parsed attribute list), so #8121's requirement that RS4GC not statepoint-wrap perry-emitted asm still holds. I checked this specifically, because a dropped marker is a latent SIGBUS rather than a wrong answer, and it would not have shown up in my reproducer either way.

The one thing worth carrying forward from this PR's write-up is the observability point, which #9055 does not state: small modules keep the text transport unless `PERRY_LLVM_INPROCESS=native` is explicit, and only large codegen-unit-split modules default to the native reader. So *any* future dialect-reader gap is invisible to ordinary small-module compiles and to the suites — `=native` is the lever that makes it reproducible, and it's worth reaching for whenever the reader changes.

PR #9057, opened 2026-08-29
fix(compiler): retain extracted Math methods

### LINKED ISSUE #9051: Auto-optimization prunes extracted Math methods such as const cos = Math.cos
## Summary

Extracting a Math method as a first-class function can produce `undefined` in an auto-optimized Perry binary:

```ts
const cos = Math.cos;
console.log(cos(0));
```

Direct intrinsic calls such as `Math.cos(0)` are unaffected. This breaks Three.js, which caches `Math.cos` and `Math.sin` in local constants.

Reproduced on `e81548ed17ff9488c3f8c19256048dc8d8497867`. The affected feature-detection logic is unchanged on `main` at `bbf1bcf00921b1d15fec77866bd2f5f9fbfdff95`.

## Reproduction

`main.ts`:

```ts
const cos = Math.cos;
console.log(cos(0));
```

Compile with the normal auto-optimized CLI pipeline and run:

```sh
perry compile main.ts -o main
./main
```

## Actual behavior

```text
TypeError: value is not a function
```

Under LLDB, `js_closure_unbox_callee_checked` receives Perry's `TAG_UNDEFINED` value (`0x7ffc000000000001`). The failure occurs before `js_closure_call1_receiverless`; closure dispatch is not the cause.

## Expected behavior

```text
1
```

Detached built-in functions must remain callable with ordinary receiverless-call semantics.

## Root cause

Optional runtime feature detection in `crates/perry/src/commands/compile/collect_modules/feature_detect.rs` enables the reified Math namespace by searching post-lowering HIR debug output for the literal string `"Math"`.

By that point, the extracted member has become approximately:

```text
PropertyGet {
    object: GlobalGet(0),
    property: "cos"
}
```

The original `Math` token is gone. `uses_global_math` remains false, the reified Math member table is omitted from the optimized runtime, and reading `Math.cos` returns `undefined`.

## Proposed fix

Conservatively set `uses_global_math` when post-lowering HIR contains a first-class property read from the global sentinel whose property is a supported Math member (`cos`, `sin`, `imul`, `max`, etc.).

A structural HIR walk is preferable. A conservative property-name scan is also correct because false positives only retain a small runtime table, while false negatives change JavaScript behavior.

The scan must include module init, functions, class methods, constructors, accessors, static methods, and field initializers. Three.js reaches this through a class method.

## Regression coverage

- Full CLI compile/run test with auto-optimization enabled for the two-line reproduction.
- The same reproduction inside a class method.
- Assert `const cos = Math.cos; cos(0)` returns `1`.
- Keep a direct `Math.cos(0)` test to ensure the intrinsic path remains intact.

This needs an end-to-end CLI test; a codegen-only test does not exercise runtime feature pruning.

proggeramlug:
## Summary

Retains the reified `Math` namespace in auto-optimized binaries when a supported Math method is extracted as a first-class function. This fixes `const cos = Math.cos; cos(0)` returning `undefined`, including when the extraction appears inside a class body.

## Changes

- Detect supported reified Math member names after HIR lowering has collapsed the original `Math` receiver.
- Include class constructors, methods, accessors, and field initializers in the Math-specific feature scan.
- Add an auto-optimized CLI compile/run regression for top-level and class-method extraction while keeping direct `Math.cos(0)` coverage.
- Add the required changelog fragment without changing version metadata.

## Related issue

Fixes #9051

## Test plan

- [ ] `cargo build --release` clean (not run; focused compiler and end-to-end coverage was used)
- [ ] Full workspace test command passes (not run)
- [x] `CARGO_TARGET_DIR=/root/perry-target-9051 cargo test -p perry --bin perry commands::compile::collect_modules::feature_detect::tests` (6 passed)
- [x] `CARGO_TARGET_DIR=/root/perry-target-9051 cargo test -p perry --test issue_9051_extracted_math -- --nocapture` (1 passed; default auto-optimization, class and top-level cases)
- [x] Added a Rust integration regression under `crates/perry/tests/`
- [x] Documentation is not applicable; no CLI, configuration, or runtime API changed
- [x] Platform UI backend testing is not applicable

## Screenshots / output

The compiled class-method and top-level extraction cases both print `1`; the direct intrinsic controls also print `1`.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed auto-optimized binaries so extracted `Math` methods, such as `Math.cos`, remain callable.
  * Resolved issues affecting extracted math methods used at the top level and within class bodies.

* **Tests**
  * Added regression coverage for extracted `Math` methods in multiple contexts.

### DIFF
--- a/changelog.d/9051-extracted-math.md
+++ b/changelog.d/9051-extracted-math.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Auto-optimized binaries now retain callable `Math` namespace members when methods such as `Math.cos` are captured as values, including inside class bodies.
--- a/crates/perry/src/commands/compile/collect_modules/feature_detect.rs
+++ b/crates/perry/src/commands/compile/collect_modules/feature_detect.rs
@@ -68,6 +68,25 @@ fn debug_hir_uses_string_normalization(hir_debug: &str) -> bool {
         || hir_debug.contains("property: \"Collator\"")
 }
 
+fn debug_hir_uses_global_math_member(hir_debug: &str) -> bool {
+    // Value-form reads such as `const cos = Math.cos` lose the `Math`
+    // receiver during lowering and reach final HIR as a property read from
+    // the shared GlobalGet(0) builtin sentinel. Match every member reified by
+    // `install_math_namespace`. A same-named property on another object is a
+    // benign false positive (a slightly larger runtime), while a false
+    // negative leaves the extracted function undefined.
+    const MEMBERS: &[&str] = &[
+        "abs", "acos", "acosh", "asin", "asinh", "atan", "atan2", "atanh", "cbrt", "ceil", "clz32",
+        "cos", "cosh", "exp", "expm1", "f16round", "floor", "fround", "hypot", "imul", "log",
+        "log1p", "log2", "log10", "max", "min", "pow", "random", "round", "sign", "sin", "sinh",
+        "sqrt", "tan", "tanh", "trunc",
+    ];
+
+    MEMBERS
+        .iter()
+        .any(|member| hir_debug.contains(&format!(r#"property: "{member}""#)))
+}
+
 fn imports_fs_promises_glob(hir_module: &perry_hir::Module) -> bool {
     hir_module.imports.iter().any(|import| {
         !import.type_only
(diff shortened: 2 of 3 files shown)


proggeramlug:
Closing as **superseded by #9058**, which fixes the same issue (#9051) and landed first. Duplicate, not a rejection.

Verified by reproduction rather than by matching titles — including the cases this PR claims as its distinctive coverage. On current `main` @ `57926713b0`, compiled **with auto-optimization** (the configuration #9051 fails under):

```
const cos = Math.cos; cos(0)                         → 1   ✅
const { sqrt, abs } = Math; sqrt(9), abs(-3)         → 3 3 ✅
class C { run() { const s = Math.sin; return s(0) } } → 0   ✅
class D { private f = Math.floor; go() { … } }        → 2   ✅
```

All four match node. The last two are the class-body cases this PR's summary highlights ("including when the extraction appears inside a class body"), and both are already correct on `main`, so there is no residue to carry over.

If you believe #9058 missed a shape this branch handles — a getter/setter or a static initializer, say — reopen with a failing reproducer against current `main` and I will take it. I checked method and field-initializer extraction specifically; I did not check accessors.

One note on the test plan, since it applies to whatever lands next rather than to this PR: `cargo build --release` and the full workspace run were both marked not-run. That is fine for a focused fix, but the specific gap that has bitten repeatedly this week is `cargo test -p perry --bins` — the `perry` bin crate holds assertions (like the build-cache env-var registration) whose failure stops that whole test binary compiling and reddens every open PR, and none of the crate-level `--lib` runs reach it.

PR #9060, opened 2026-08-29
perf(codegen): reduce accumulators earn the packed fast clone's numeric proof — s += arr[i] at node parity (was 5.3×)

proggeramlug:
## What

Reduce loops — `for (let i = 0; i < arr.length; i++) s += arr[i]` — now run
their stable-packed fast clone at full speed. Two independent fixes, each
useless without the other:

1. **The accumulator earns the clone's numeric proof.** The loop guard proves
   the *element* is raw f64, but the accumulator's own writes are circular for
   every whole-function numeric fact, so `s + arr[i]` kept one unproven operand
   and lowered through `js_dynamic_string_or_number_add` on every element (25%
   of the isolated loop). This ports the element-shape clone's
   `numeric_accumulator` design: `collect_numeric_accumulators`
   (`stmt/stable_packed_accumulator.rs`) admits plain, uncaptured, unboxed
   locals whose **every** body write is numeric with all leaves provable
   in-loop (fail-closed fixpoint; nested closures not descended — their
   captures are boxed and already excluded). The fast preheader tag-tests each
   admitted accumulator (the induction base case) and takes the slow clone on
   any non-Number; the fact rides `StablePackedLoopFact::numeric_accumulators`,
   scoped to the fast-clone lowering exactly as the element facts are, and
   `is_numeric_expr`'s `LocalGet` arm consumes it like the element-shape twin.

2. **`js_shadow_slot_set` is exempted from the call-free clone scans.** With
   the add fixed, the loop was *still* 5× slower, and nothing pointed at why:
   the guard admitted (lldb shows return 1), the fast body IR was perfect
   (19 instructions, no calls), timing showed slow. The cause sat in the
   admission block's *terminator*: the accumulator's per-statement shadow
   CLEAR — emitted precisely **because** the stored value is a proven
   non-pointer — failed `fast_clone_call_free`, and the admission arm then
   emits an **unconditional** branch to the slow preheader while still calling
   (and discarding) the guard. `js_shadow_slot_set` is a bounds-checked TLS
   store (`gc/roots/shadow_stack.rs`): it cannot allocate, collect, or revoke
   a layout, which is exactly what the two call-free scans exist to exclude.
   The exemption arms both this tier and the element-shape tier; the
   diagnostic that found it is preserved at
   `secret-tests/tools/packed-scan-debug.patch`.

## Numbers

Isolated (Mac, 1000-element `number[]` behind a closure/module-global receiver,
node 26.5.1 = 1009 ns/1k on the same machine):

| shape | before | after |
|---|---|---|
| `s += arr[i]`, `i < arr.length` | 5341 ns/1k | **993–1040 ns/1k = node parity** |
| same, no-array control (`s += i`) | 990 | 990 (unchanged — proves the loop scaffolding was never the cost) |

wolf-ecs (11 pairs, both windows): ±0.08%, neutral — its hot loops don't have
this shape. Two admission residuals are documented in the matcher for
follow-up: compare-only consumption (`if (arr[i] < 0)` — the leading-statement
matcher does not look into `If` conditions) and literal-bound loops
(`i < 1000`), both still ~4.3 µs/1k.

## Semantics

Differential vs node, identical output: string accumulators (concat preserved —
the preheader tag test routes them to the slow clone), arrays with non-number
elements (the guard declines numeric mode), NaN and `-0` accumulators, in-loop
reassignment to a string (admission declines the accumulator), multiple
accumulators with `-`/`*`/`Math.*` chains, update-form counters (`c++`), and
fact scoping across back-to-back loops. `PERRY_PACKED_LOOP_NUMERIC_ACCUMULATOR=0`
restores the old lowering, output-identical; the scan exemption is
unconditional because it is a factual classification of the callee, not a
policy.

## Testing

* `perry-codegen` 1823 / `perry-runtime --lib` 2807 — green (re-run after the
  file-size split of the matcher into `stable_packed_accumulator.rs`).
* Integration: `issue_8655_array_subclass_indexing`,
  `issue_8897_field_push_writeback` — 12/12.

## Summary by CodeRabbit

  * Improved packed-loop reductions by using faster numeric accumulation when values are confirmed to be numbers.
  * Reduced dynamic addition overhead, improving performance for eligible numeric loops.
  * Preserved fast-path execution when accumulator shadow slots are cleared.

* **Build & Reliability**
  * Compiler setting changes now correctly invalidate affected build-cache entries.
  * Improved safety handling for shadow-slot updates during optimized execution.


proggeramlug:
Merged, after fixing a silent revert that would have re-reddened `main`.

## The revert

Commit `e3164ee7a8` removes `PERRY_BOX_CAPTURE_ENTRY_CELLS` and `PERRY_GUARDED_PREINLINE_MAX_IR_BYTES` from `BUILD_CACHE_ENV_VARS`, plus #9044's changelog fragment. The branch predates #9044, so the commit was built over a stale tree and carries the removal as an intentional-looking deletion — which a merge then honours rather than discards. Merging `main` in and re-checking is what surfaced it; the PR's own file list shows it as `-14`/`-8` if you know to look.

This is the same shape as the war story in #9021 (a `git reset --soft` over a stale tree silently reverting #9013). Worth a habit: after any rebase/squash of a long-lived branch, `git diff origin/main --stat` and look for deletions you did not intend.

It would not have been a quiet regression. The assertion is a **bin-crate unit test**, so its failure stops the whole `perry` test binary from compiling and every open PR's cargo-test job goes red — which is exactly how #9044 was found.

## The third knob

This PR also adds `PERRY_PACKED_LOOP_NUMERIC_ACCUMULATOR` without registering it, so post-merge there would have been **three** unregistered vars, not two. Registered as an **input**, not an exclusion: with it on, `s += arr[i]` lowers to an inline `fadd` instead of `js_dynamic_string_or_number_add`, so the two settings emit different code and must never share a cached object.

Fragment renumbered `0000-` → `9060-`.

## The optimization itself

The interesting risk is the accumulator earning a numeric proof it does not deserve, so I probed the ways an accumulator leaves the numeric domain: a string accumulator (`s = ""`), an accumulator that goes non-numeric mid-loop, a string *element* mid-array, `NaN`/`Infinity`/`-0`, two accumulators in one loop where only one is numeric, an accumulator captured by a closure (which must be excluded as boxed), and `*=`/`-=` reductions. **Byte-identical to node in both knob states**, so the preheader tag-test genuinely side-exits rather than the fast clone swallowing a non-Number.

The fail-closed fixpoint plus excluding captured/boxed locals is the right shape — it is the same admission discipline as #9018's number-by-construction and #9026's capture cells.

Validation: `perry-codegen` 1345/0, `perry-runtime --lib` 2807/0, **`perry --bins` 1066/0**, `fmt --check`, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped.

PR #9063, opened 2026-08-29
perf(codegen): masked-window dense loops admit provably-numeric stores

proggeramlug:
Follow-up to #9041 — the last plain-array store shape that trailed node.

## What

The dense packed-f64 range loop (#6750) was **read-only**: its entry guard proves a whole static index window in bounds and hole-free on a plain raw-f64 array, and the fast clone has no side exits — so loads are bare in-window reads, but any store in the body rejected the loop entirely, sending masked fill/copy loops (`a[i & K] = v`, the ring-buffer/LUT write shape) to per-store guarded lanes.

This admits stores whose RHS **provably materializes a genuine (unboxed) double by construction** — number/integer literals, a local with a shared i32 shadow slot (a versioned-loop counter reads as `sitofp i32 → double`), or an in-window masked load (the guard proved the slot holds a raw-f64 number). Dense mode has no side exits, so the proof must be static; **no arithmetic is admitted in this cut** (an `Any`-typed operand can route `+` through a boxed-result helper). The store lowers to a bare in-window `store double`: a number carries no heap edge → no barrier, no layout note (`GC_STORE_AUDIT(POINTER_FREE)` — no new barrier stem, nothing for the census), and the array pointer is re-derived from the receiver box per store exactly like the plain-tier loads, so a loop-poll GC move cannot strand it.

## Tier discipline

- Store-admitting loops use **only the f64 dense guard tier**: a genuine-double store could break the i32 tier's all-slots-i32 loading proof mid-loop, and the TA tiers' hoisted data pointers serve reads only. `MaskedWindowArrayFact.allows_stores` is set solely in the f64 dense-with-stores scope; `masked_window_store_fact_for_index` requires it plus the PlainF64 elem.
- Dense **written** arrays take the READ-side eligibility (addressable, not scalar-replaced) instead of the full static fact set: the materialization-hazard bit trips on the very `new Array(n).fill(0)` construction calls that build these buffers, while the dense guard re-validates the ACTUAL runtime array at every loop entry and the matched body admits no call/closure/await. A wrong static hint costs one failed guard → slow loop, never correctness. Classic (side-exiting, hole-tolerant) written arrays keep the full set.

## Numbers (isolated masked store loops, dev box; node 26.5 in parens)

| receiver | before | after |
|---|---|---|
| parameter | 3.4 ns | **0.4** (0.7) |
| module global | 3.1 | **0.4** (0.7) |
| function local | 3.7 | **0.5** (0.7) |

With #9036 + #9041 + this, every plain-array store shape in the ops matrix is ahead of node.

## Correctness

- Ten-probe node differential, **byte-identical**: wrap-around fill, holey array at entry (guard fails → slow loop), window wider than the array (slow loop extends), frozen receiver, mixed-type window, read-write aliasing through the same window, two-array copy with scalar statements in the body, NaN/−0 fidelity, reduce-observed results, module-global receiver.
- Two native-proof pins: the fast clone has the f64 dense guard, **no** i32-tier guard call, **no** per-store runtime call, **no** barrier, and a raw `store double`; an RHS numeric only by TYPE (a number parameter — potentially INT32-boxed at runtime) never reaches the raw store.
- Full host gate + local suite re-run on this exact head are in flight; results follow as a comment (lint triage per the #9041 lesson: any non-template FAIL is a blocker).

## Adjacency

`stable_packed_loop.rs`/loops.rs textual adjacency with perf/packed-loop-accumulator-v2 (numeric accumulators) — semantically independent per coordination with that branch's author; whichever lands second rebases trivially.

## Summary by CodeRabbit

  * Improved dense numeric array fill and copy loops with faster in-window stores for proven numeric values.
  * Reduced per-store overhead in eligible masked-index operations.

* **Bug Fixes**
  * Added safeguards to ensure optimized stores are used only when values and array layouts are safely validated.

* **Tests**
  * Added regression coverage for optimized numeric stores and cases that must continue using the standard path.


proggeramlug:
Head is now `2e6c0c387c` — rebased onto current main (which picks up #9061's aggregate_scalar.rs split, clearing the file-size gate failure that was main-side drift) with the rustfmt fixes folded in.

Full host gate on this head: **all real lint steps pass** (formatting ✓, file-size ✓, GC store-site inventory ✓ — no new barrier stems, the masked store is POINTER_FREE), ratchets clean, suites green. One parallel-only runtime flake (`class_registry::dispatch::a_class_registration_invalidates_every_entry`) passes in isolation, same pattern as the known stack_maps one. Re-validated on this exact head: masked stores 0.4/0.4/0.5 ns (param/global/local), ten-probe node differential byte-identical, both native-proof pins green. Ready to merge.

proggeramlug:
Merged. Admitting stores into a clone that has **no side exits** means the proof has to be fully static, so that's what I went at.

The admission set is appropriately narrow — literals, a local with a shared i32 shadow slot, and an in-window masked load — and I think declining arithmetic in this cut is the right call rather than a limitation: an `Any`-typed operand routing `+` through a boxed-result helper is exactly the case a no-side-exit clone cannot recover from, and there is no cheap static way to exclude it.

Probed the ways a store could smuggle a non-double into the dense tier, or the tier could be entered when it shouldn't: literal fill, counter RHS through the `sitofp` shadow slot, an in-window masked load as RHS (the rotate shape), arithmetic RHS (must stay out of the tier and still be correct), a mid-loop string store on an `any[]`, `-0`/`NaN`/`Infinity` fidelity through the masked store, `new Array(n)` holes at entry rejecting the tier, and a 64-wide mask over 256 iterations with a full checksum read-back. **Byte-identical to node.**

Two details I checked because they are the ones that bite quietly:

- **`GC_STORE_AUDIT(POINTER_FREE)` is the correct class here**, and it is the reason there is no new barrier stem for the census to witness — a number carries no heap edge, so there is genuinely nothing to barrier. That is a real difference from #9041's lanes, not an omission, and `gc_store_site_inventory.py` agrees (green, with no new stem to register).
- **Re-deriving the array pointer from the receiver box per store** is what keeps a loop-poll GC move from stranding it. Hoisting that pointer would have been the obvious "optimization" and the #7690-family bug; matching the plain-tier loads here is the conservative choice.

The tier discipline note in the description is worth keeping visible — restricting store-admitting loops to the f64 dense guard tier because a genuine-double store would break the i32 tier's all-slots-i32 proof mid-loop is the kind of interaction that only shows up as a wrong answer much later.

Validation: `perry-codegen` 1345/0, `native_proof_regressions` **283/0**, `perry-runtime --lib` 2808/0, `perry --bins` 1066/0, `fmt --check`, `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**.

PR #9066, opened 2026-08-29
fix(runtime): reserve iterator raw-field floor so own .next patches stop corrupting state (#9019)

### LINKED ISSUE #9019: for…of over a Set iterator with a patched .next() segfaults (pre-existing, not #9017)

proggeramlug:
Fixes #9019.

## The bug

A by-name property write on a builtin collection-iterator object derived the new key's field index from the object's (empty) keys array, so the **first user property landed at field index 0 — the backing-collection pointer**. Everything downstream of that write read the stored value as iterator internals:

- `it.foo = 123` → the next builtin `.next()` read `123` as the backing pointer → null-ish → the iterator silently reported `done: true` on a live collection (all families: map/set/array/string).
- `it.next = fn` → the next builtin advance dereferenced the closure as a `SetHeader` → **SIGSEGV**, whether driven by `for…of` or by the bound original. The patched function was never even called — the crash reproduces with the *original* bound `next` after the assignment.

The issue's "manual path is fine" observation was the same corruption one step later: manual case B/C in the probe happened not to re-read field 0 before the patch's own `next` returned.

## The fix

**Storage** (`object/reserved_floor.rs`, new): the first by-name append to a reserved-layout receiver (array/map/set/string/buffer/regexp iterators, iterator helpers — exactly the `is_builtin_iterator_class_id` families) seeds its keys array with `floor` leading tombstones (`TAG_HOLE`, the #9038 hole-delete marker that every lookup, enumeration, and delete path already skips). User keys then append **past** the raw internal fields into the ordinary inline/overflow storage, and the keys-position ↔ field-index correspondence every by-name path relies on is preserved. Unpatched iterators pay nothing — the seed runs only when user code actually adds a named property, and it runs before `prev_shape_id` is read, so the transition-cache/plan fast paths only ever learn edges rooted at the seeded shape. The hole-squeeze compaction (`delete_rest.rs`) preserves the reserved prefix.

**Dispatch**: the class-id iterator dispatchers now honor an **own** `next` before the builtin advance (`call_overridden_iterator_next` probes the instance before the prototype tower; a present but non-callable own value throws per IteratorNext's GetV+Call). The canonical prototype thunks route through new `*_builtin` dispatch variants that skip the probe — `proto.next.call(it)` (or a `.bind(it)` taken before the patch) must run the builtin algorithm, both per spec and because honoring the override there sends a patch that delegates to its bound original into infinite recursion. The fused `for…of` map/set arms validate the iterator result (a patched `next` can return a primitive; the builtin never could), and the stored-closure drain paths in `js_iterator_to_array` and friends bind `this` to the iterator per Call(next, iterator) and reject non-callable own `next` values instead of calling through garbage bits.

## Validation

- The issue reproducer and 12 more cases (manual drive, per-family `foo`-write non-corruption, value-rewriting patches, spread, `Object.keys`/`JSON.stringify`, `delete`-restores-builtin, `proto.next.call(this)` delegation, non-callable `next` → TypeError under `for…of`) are **byte-identical against the pinned Node 26.5.1 oracle** — committed as `test-files/test_gap_iterator_patched_next.ts`. Buffer- and regexp-family bind-delegation probes also match Node (iterator helpers can't be probed this way — pre-existing read-side gap, filed as #9068).
- `RUST_TEST_THREADS=1 cargo test -p perry-runtime`: 2790 passed, 0 failed (includes 4 new `reserved_floor` unit tests, one of which is the sabotage-shaped field-0 fixture: it asserts the backing pointer is *unchanged* after the write, not merely that nothing threw).
- **Full gap suite** (`run_parity_tests.sh --filter test_gap_`, `perry-dev` build, pinned Node 26.5.1, Linux x86_64): 561/585 pass incl. the new test. All 24 failures were A/B'd against a merge-base build on the same box before attribution: 5 are the `gap_snapshot.json` known-fails, and the other 19 are byte-identical on both arms — the 5 npm-package tests (known local-env fails), 11 net/http compile fails + the wasm-host archive one (box environment), `zlib_3285_params` (async completion fails on *clean-target builds of the pristine merge-base too* — build-layout-sensitive, pre-existing), and `set_map_foreach_fused_receiver`, which is a real pre-existing main-side regression (delete-during-forEach visits holes/skips entries) — filed as #9072. **Zero failures attribute to this diff.**
- `scripts/run_lint_gates.sh`: 58/60 ok including `cargo check --workspace --all-targets -D warnings`, the raw-handle ratchet (debt −3 vs baseline), addr-class, and the GC root-holder gates. The one FAIL is `check_file_size.sh` on `crates/perry-transform/src/aggregate_scalar.rs` (2014 lines) — pre-existing: it crossed the 2000-line cap in #9056 on `main` (1965 → 2014) and sits at 2043 on current `main`; untouched by this diff.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed iterator corruption when adding properties to built-in iterators.
  * Preserved correct `.next()`, `for…of`, spread, and manual iteration across supported iterator types.
  * Patched iterator instances now honor custom `.next` methods, while invalid values raise `TypeError`.
  * Deleting a custom `.next` restores built-in iteration.

* **Tests**
  * Added regression coverage for property writes, overrides, deletion, enumerability, and iteration behavior.


proggeramlug:
**First: thank you — this fixes #9019, and the root cause is worse than my issue described.** I had written that the manual `.next()` path was fine; you're right that it was the same corruption one step later, and my probe's manual cases simply didn't re-read field 0 before returning. The real defect — first user property landing on the backing-collection pointer — explains both the SIGSEGV and the silent `done: true`, which my issue treated as unrelated.

Verified against my original reproducer: **exit 0 where main is 139**, and byte-identical to node across patched `.next()` under `for…of`, a plain `it.foo = 123` write, all four iterator families (map/set/array/string), a manually-driven patched Map iterator, and user keys surviving enumeration.

**Holding on two gate failures**, both real.

## 1. `perry-runtime` does not compile without the `regex-engine` feature

```
error[E0432]: unresolved import `match_all`
  --> crates/perry-runtime/src/regex.rs:69:9
```

`mod match_all;` (line 41) and the `pub(crate) use` (line 68) are both `#[cfg(feature = "regex-engine")]`; the `pub use` at line 69 is **not**. With the feature off the export names a module that doesn't exist.

This passes `cargo test -p perry-runtime --lib` (2811/0, default features on) and fails `cargo check -p perry`, which is why it's easy to miss — the same bin-crate blind spot that produced #9044. I didn't fix it because the choice isn't mechanical: the comment right below says `REGEXP_STRING_ITERATOR_CLASS_ID` stays ungated because the always-linked iterator-prototype dispatch references it, so gating the export may need a stub for the feature-off build rather than just a `cfg`. You know whether that dispatch must work with the engine compiled out.

## 2. Raw-handle ratchet (`raw_handle_debt.py`, in `lint`)

```
bare raw-handle reads: 969 (baseline 967)
  tail.rs: 14 bare reads exceeds its ceiling of 11
  reserved_floor.rs: 2 bare read(s) in a module with no ceiling
```

The new module lands at zero by default — unlisted modules are locked there — so its two bare reads need `across_{mut,const,nanbox}` or `with_{mut,const}_ptr` (#7341). Worth doing on the merits rather than to satisfy the gate: this PR is about raw field slots on objects that user code can now mutate, which is exactly where a stale pointer across an allocation point bites.

## One pre-existing gap this does NOT fix (not a blocker)

Adding 12 properties to an iterator and deleting 10 loses the *values* of the survivors — `Object.keys` correctly reports `p10,p11`, but reading them gives `undefined`. **Identical on `main`**, so this PR neither causes nor worsens it, and the same shape on a plain object is correct on main, so it's iterator-specific rather than a general squeeze bug. Mentioning it because it sits inside the storage model you're rewriting; if the reserved floor makes it easy to fix, this is the PR that would naturally carry it. I'll file it separately otherwise.

Everything else: `perry-runtime --lib` 2811/0, `reserved_floor` 4/4, `fmt --check`, fragment correctly numbered, remaining 59 lint gates green.

proggeramlug:
Merged. Raw-handle ratchet is green on your update; I fixed the remaining compile blocker myself since it turned out to be a one-line accident rather than a design question.

**And I owe you a correction on how I framed it.** I said the choice "isn't mechanical" and might need a stub for the feature-off build. That was wrong — I hadn't diffed the line against `main`. What actually happened:

```rust
 #[cfg(feature = "regex-engine")]
+pub(crate) use match_all::dispatch_regexp_string_iterator_method_builtin;   ← inserted here
 pub use match_all::{ dispatch_regexp_string_iterator_method, … };
```

`main` had that `#[cfg]` attached to the `pub use`. Inserting a line directly beneath the attribute moved it onto the **new** export and left the original ungated. So the fix is simply a second `#[cfg]`, not a stub — no dispatch redesign needed, and `-p perry` compiles again (main was exit 0 all along, confirming this was the branch's regression rather than pre-existing).

Worth naming the shape, because this is its third appearance in a week: an inserted line silently inherits the attribute or doc block above it. #9013 and #9030 were the same accident with doc comments — a new function absorbing the previous one's `///` block, leaving the original undocumented. Attributes fail loudly at compile time *if* something exercises the configuration; doc comments never do.

The reason it hid: `cargo test -p perry-runtime --lib` builds with default features and passes 2812/0, while `cargo check -p perry` unifies features differently and fails. Same bin-crate blind spot as #9044, which is why `perry --bins` is now in my standard set.

Re-verified after the fix: #9019's reproducer exits 0 (main: 139) and is byte-identical to node across patched `.next()` under `for…of`, plain property writes, all four iterator families, a manually-driven patched Map iterator, and user keys surviving enumeration. `perry-runtime --lib` 2812/0, `perry --bins` 1066/0, `fmt --check`, `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**.

The pre-existing iterator gap I mentioned (values lost after a squeeze crossing the threshold) is unchanged — identical on `main`, correct for plain objects. I'll file it separately rather than hold this.

proggeramlug:
Closing the loop on the two holds (thanks for landing the cfg fix directly — same shape I had staged: the insertion had detached the attribute from the neighboring `pub use`):

- **Raw-handle ratchet**: those numbers were the state at `18d124f59`; `d908219e5` in the merged stack moved the seed to NaN-boxed handles and put the tail hook on the existing `refresh_roots_after_alloc!` macro. Bare run on the merged head: `964 (baseline 967)`, all ceilings.
- **The survivor-values gap**: root-caused, and it was worth the dig — **not the squeeze, and not this PR's storage**. The values sat intact in the overflow spill the whole time; the Map/Set-iterator arm in the by-name GET tail ended with `if key_bytes != b"next" { return undefined }`, so every non-`next` own-property read was short-circuited and the reserved-floor storage was write-only through that lane (IC lanes served some compiled reads, which is why single-property probes passed). Fix is up as **#9075**, with your exact 12-add/10-delete shape as a unit test and a gap case, byte-identical vs 26.5.1 — plus the `it.next = undefined` present-but-non-callable case from the inline review, and seed-failure now drops the write instead of proceeding onto field 0.

Filed along the way: #9068 (helper-family `.next` value read returns undefined — pre-existing, read-side) and #9072 (delete-during-`forEach` on Set/Map visits holes on clean main; A/B'd from the merge-base before attributing).

PR #9067, opened 2026-08-29
perf(runtime): preserve shared shape index across delete

### LINKED ISSUE #9006: shape_index_migrate_after_delete steals a shared keys array's index, stranding sibling objects

proggeramlug:
## Summary

- clone the validated slot index when delete forks a shared keys array, while retaining the cheaper move for owned sources
- add an end-to-end regression proving both the untouched sibling and deleting object's shifted clone keep valid accelerators
- add a dedicated 500-key shared-sibling benchmark and changelog fragment; no version metadata is changed

## Performance

On `bench_shared_shape_delete.ts`, 15 interleaved A/B pairs pinned to one CPU:

- median: 171 ms -> 70 ms (-59.1%)
- minimum: 111 ms -> 43 ms (-61.3%)
- checksum: identical in every arm

Both executables were built from commit-matched compiler/runtime pairs on `perrymaster.skelpo.net`.

## Verification

- `cargo test -p perry-runtime shared_delete_preserves_the_sibling_shape_index -- --nocapture`
- `cargo test --quiet -p perry-runtime -- --test-threads=1` (2787 passed, 4 ignored)
- `RUSTFLAGS="-D warnings" cargo check -p perry-runtime --all-targets`
- `scripts/run_lint_gates.sh` with compile tier skipped (all 58 runnable lint gates passed; compile covered by the targeted warning-denied check)
- `cargo clippy -p perry-runtime --lib`

Closes #9006.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved performance when deleting properties from objects that share internal structures.
  - Preserved fast property lookups on related objects after a deletion.
  - Added regression coverage to ensure sibling objects remain unaffected.

- **Benchmarks**
  - Added benchmarking for shared-object property deletion and subsequent lookups.
  - Demonstrated a significant reduction in processing time for this scenario.


proggeramlug:
Merged — this closes #9006, which I filed against #9002, and the fix is exactly the right shape: clone for a shared source, move for an owned one.

The asymmetry is the whole point and the comment states it well — forking a shared array is a genuine shape transition (siblings keep naming the old address), while replacing an owned array transfers its identity. `get().cloned()` vs `remove()` on that one predicate is a small change for a −59% result.

Verified end-to-end rather than only at the unit level: 25 same-shape siblings of 60 keys with a delete on one — every untouched sibling keeps all 60 keys and correct values, the deleter has 59 with the right hole; ten siblings each deleting a different key; an owned (non-shared) receiver; and delete-then-re-add on one sibling leaving the other untouched. **Byte-identical to node**, zero sibling corruption.

Worth being precise about what changed, since #9006 was explicitly a throughput issue: correctness was never at risk either way — a stranded sibling rebuilt its index and got the right answer. What this fixes is that it no longer *has* to.

**One process note.** Your verification ran `run_lint_gates.sh` with the compile tier skipped, substituting a targeted `cargo check -p perry-runtime --all-targets`. I ran the full tier and it's green here, so no harm — but that substitution is exactly what hid the blocker on #9066: `perry-runtime` compiles standalone with default features and fails under `-p perry`'s feature unification, which a single-crate `--all-targets` check cannot see. Worth running the compile tier unskipped on anything touching a crate that others depend on.

Validation: the new `shared_delete_preserves_the_sibling_shape_index` test passes, `perry-runtime --lib` 2808/0, `perry --bins` 1066/0, `fmt --check`, `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped** (compile tier included).

PR #9069, opened 2026-08-29
fix(codegen): all-vouched guarded '+' tree lowers unguarded — un-ICEs the pi and cc builds

proggeramlug:
Both application bundles fail to compile on current main with `guarded + tree has no testable leaf` (pi dies in `graphemeWidth`, the cc cli bundle in a closure). Minimized to 13 lines — the load-bearing shape is a compound add in a for-of loop where the RHS is a call:

```js
function eaw(cp) { if (cp > 100) { return 2; } return 1; }
function gw(s) {
  let width = eaw(s.codePointAt(0));
  for (const ch of s) {
    const c = ch.codePointAt(0);
    if (c >= 65280) { width += eaw(c); }      // ← the ICE
    else if (c === 3635) { width += 1; }      // this alone compiles
  }
  return width;
}
```

**Mechanism:** the call leaf is flagged by `numeric_proof_is_declared_only` (declared numeric return ⇒ the tree routes into `lower_guarded_numeric_add`) while its integer-literal returns make the same leaf provenance-vouched by `expr_produces_canonical_raw_f64` (⇒ no leaf needs a runtime test). The lowering's all-vouched corner was a deliberate `bail!` tripwire against predicate drift — this week's provenance widening made it reachable from ordinary code, and it fails whole builds.

**Fix:** all leaves proven ⇒ emit the fast tree unguarded. This is not new trust: the mixed case already `fadd`s every vouched leaf untested, so an unsound vouching predicate ships wrong answers there regardless — the predicate's own tests are the guard, not the bail (rationale in the code comment).

**Verification:** the 13-line repro and the full extracted `graphemeWidth`+`eastAsianWidth` pair compile and print node-identical output; gap-suite fixture `test_gap_9050_all_vouched_guarded_add.ts` pins the shape (verified still-ICE on `5792671` server-side before fixing); perry-codegen lib 1344/0; fmt green. Un-blocks recompiling both pi and cc for the GC-share measurement.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a compiler crash when compiling certain numeric addition expressions whose values are fully proven safe.
  * Improved compilation of integer-provenance calculations, including compound additions in loops.
  * Applications using these expressions now compile successfully instead of failing during compilation.

* **Tests**
  * Added coverage for guarded additions involving character-code processing and conditional width calculations.


proggeramlug:
Merged. Reproduced the ICE on `main` from your minimized shape before testing the fix:

```
Error compiling module 'probe.ts': lowering specialized entry for function 'gw':
  lowering body of 'gw': guarded `+` tree has no testable leaf
```

and it compiles and runs byte-identical to node afterwards, including the wide-character path (`gw("０１")` → 6), a vouched-call leaf in a different loop shape, and a mixed numeric/string accumulator that must *not* take the unguarded path.

The diagnosis is satisfying because the two predicates are individually right and only wrong together: `numeric_proof_is_declared_only` routes the tree into the guarded lowering on the strength of a declared return type, while `expr_produces_canonical_raw_f64` independently vouches the same leaf from its integer-literal returns — so the guarded path was entered with nothing left to guard. "Every leaf vouched" is precisely the case that needs no guard, so lowering unguarded is the right resolution rather than inventing a test.

This is the seam I flagged when auditing #9033, which extended `expr_produces_canonical_raw_f64` to integer-provenance locals: I noted that predicate's inputs include declared-type-derived facts and said it was the first place to look if it were widened again. It didn't cause this one — the interaction predates it — but it's the same pairing, and worth remembering that these two proofs can disagree about whether a leaf is *already* known good.

Validation: `perry-codegen` 1345/0, `perry-runtime --lib` 2812/0, `perry --bins` 1066/0, `fmt --check`, `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**.

Issue #6011, opened 2026-07-05
simple EMA (Exponential Moving Average) loop over a 100k-element array runs **7x slower** in the compiled Perry binary (2300ms) compared to tsx (328ms) or node (343ms).


zaaack:
## What happened

A simple EMA (Exponential Moving Average) loop over a 100k-element array runs **7x slower** in the compiled Perry binary (2300ms) compared to tsx (328ms) or node (343ms).

The loop is a straightforward `O(n)` numerical computation — no allocation inside the hot path, no closures, no array method chains.

## What you expected

A natively compiled binary running a tight floating-point loop should be at least as fast as V8's JIT, ideally faster. 7x slower indicates the float arithmetic codegen or loop optimization is significantly behind.

## Minimal reproduction

```typescript
const N = 100_000;
const period = 20;
const alpha = 2 / (period + 1);

const prices = new Array<number>(N);
for (let i = 0; i < N; i++) prices[i] = 100 + Math.sin(i * 0.01) * 10 + Math.random() * 2;

const ema = new Array<number>(N);
ema[0] = prices[0];
for (let i = 1; i < N; i++) ema[i] = prices[i] * alpha + ema[i - 1] * (1 - alpha);

return ema[N - 1];
```

Command you ran:

```bash
perry compile benchmark.ts -o dist/benchmark-perry
./dist/benchmark-perry
```

## Environment

- **Perry version**: 0.5.1220
- **Host OS**: linux
- **Target**: native
- **Installed via**: (from path `/usr/local/bin/perry`)

## Diagnostic output

```
[ema-100k] 2300ms | rss=66MB | heap=39MB
```

For comparison:

| Runtime | Time |
|---|---|
| tsx | 328ms |
| node (pre-compiled) | 343ms |
| perry (native) | 2300ms |

## Anything else

- The inner loop body does 1 `Math.sin`, 1 `Math.random`, 3 multiplications, 2 additions/subtractions, and 2 array writes — all scalar float64 operations.
- Memory overhead is low (66MB RSS) so this is purely a CPU/codegen issue, not GC thrashing.
- `Math.sin` inside the inner loop is called per-iteration to initialize `prices` — also happening in the V8 baseline, so it's a fair comparison.
- Native compilation should be able to autovectorize or at minimum produce tight scalar code for this pattern; the current ~7x gap suggests the float → x87/SSE lowering or loop-invariant code motion is missing.

PR #9070, opened 2026-08-29
perf(codegen): hoist the versioned loop's length bound; admit float arithmetic in masked-window stores

proggeramlug:
Two follow-ups to #9041/#9063, stacked on #9063 (`2e6c0c387c`) — merge after it.

## 1. Versioned-loop length-bound hoist

The versioned packed loop's fast clone re-evaluated `i < arr.length` per iteration — ~20 inline instructions of handle decode + GC-header checks + the length load, which LLVM cannot hoist past the body's raw element stores. The entry guard just proved a live, non-forwarded plain array whose length the matched body cannot change (in-bounds stores only, no calls), so hoist the length ONCE in the fast preheader and hand it to `lower_for_after_init_with_i32_bound`, exactly like the range-versioned fast copy (#6011). A mid-loop GC move changes the array's address, never its length, so the hoisted VALUE stays correct.

**`i < a.length` store loops: 1.3 → 0.70 ns/store** (node 0.59); constant-bound and hoisted-local-bound variants also gain (3.3 → 2.2, 3.6 → 2.8).

## 2. Float arithmetic in masked-window store values

#9063 admitted only literal / counter / in-window-load RHS values. Extend both predicates with float arithmetic (`+ − * /`, unary negation) over admitted operands: both sides being genuine doubles pins the numeric lowering to a bare float instruction — the boxed-`+`-helper hazard needs a non-numeric operand — and a float op over canonical operands cannot fabricate a NaN-box pattern (the default quiet NaN 0x7FF8 is a genuine double). `%` and `**` stay excluded (runtime-helper lowerings).

**`a[i & K] = b[i & K] + 1.5`: 15.9 → 0.50 ns/store — ahead of node's 1.33.** IR census of the fast clone: one raw load, one `fadd`, one raw store, the loop poll — no calls.

## Correctness

- Seven-probe arithmetic differential vs node, **byte-identical**: NaN propagation, x/0 → ±Infinity, 0/0 → NaN, −0 quotients and products, overflow to Infinity, denormals, read-modify-write of the same slot, and a mixed-type source array whose guard failure routes through the slow clone where JS `+` string concatenation applies.
- New pin: the arith fast clone is call-free with `fadd double` + `store double` and never routes through `js_add`; all masked/packed pins green (9/9 in the two families).
- Full host gate on `4ca7bee8cb`: all real lint steps pass, ratchets clean, suites green (one known parallel-only stack_maps flake, passes in isolation).

## Summary by CodeRabbit

  * Improved dense numeric loop performance by reusing validated array length bounds.
  * Optimized masked indexed stores for number arrays, including values produced by floating-point arithmetic.
  * Reduced overhead for eligible array updates by using direct numeric loads and stores.

* **Bug Fixes**
  * Preserved safe handling for unsupported or non-numeric store patterns.

* **Tests**
  * Added regression coverage for masked stores using floating-point arithmetic.

### DIFF
--- a/changelog.d/9070-versioned-len-hoist-masked-arith.md
+++ b/changelog.d/9070-versioned-len-hoist-masked-arith.md
@@ -1,0 +1,1 @@
+Hoisted the packed versioned loop's `arr.length` bound to a single pre-validated i32 read (length-bound store loops 1.3 → 0.7 ns), and extended masked-window dense stores to float-arithmetic values — `a[i & K] = b[i & K] + c` now lowers to a raw load, one float op, and a raw store (15.9 → 0.5 ns, ahead of Node).
--- a/crates/perry-codegen/src/expr/masked_window.rs
+++ b/crates/perry-codegen/src/expr/masked_window.rs
@@ -225,13 +225,19 @@ pub(crate) fn masked_window_store_fact_for_index(
 ///   raw-f64 number (and TA-tier loads materialize via `sitofp`/`uitofp` /
 ///   raw f64 loads).
 ///
-/// Deliberately NO binary/unary arithmetic in this v1: an operand typed `Any`
-/// can route `+` through a runtime helper whose result is boxed. Arithmetic
-/// over genuine doubles is a candidate extension once its lowering is pinned.
+/// Float arithmetic (`+ - * /`, unary negation) over these operands is also
+/// admitted: both sides being genuine doubles pins the numeric lowering to a
+/// bare float instruction (the boxed-`+`-helper hazard needs a non-numeric
+/// operand, and every admitted operand is numeric by construction), and the
+/// pinning test asserts the fast clone stays call-free so a lowering change
+/// that broke this would go red, not silent. `%` and `**` stay excluded —
+/// they can lower through runtime helpers.
 ///
 /// Closed under IEEE semantics: a raw-f64 slot never holds a payload-NaN in
 /// the box tag range (that is the raw-f64 invariant canonicalization exists
-/// to maintain), and none of these shapes fabricates one.
+/// to maintain), none of these shapes fabricates one, and a float op over
+/// canonical operands produces the default quiet NaN (0x7FF8...), which is a
+/// genuine double.
 pub(crate) fn masked_store_rhs_is_genuine_f64(ctx: &FnCtx<'_>, expr: &Expr) -> bool {
     match expr {
         Expr::Number(_) | Expr::Integer(_) => true,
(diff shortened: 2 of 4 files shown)


proggeramlug:
Merged. I rebased this off the merged #9063 — it was stacked, so its diff still carried that PR's commit and fragment. Cherry-picking the one new commit onto main applied cleanly, and `git diff origin/main --diff-filter=D` is empty, so nothing was carried back as an unintended revert (three PRs this week did exactly that, hence the check). I also renamed `changelog.d/9064-…` → `changelog.d/9070-…` to match the PR-keyed convention.

**The length hoist.** I went looking for the ways a hoisted bound goes stale, and the reassuring answer is that this isn't a new analysis — `classify_for_length_hoist` and `lower_for_after_init_with_i32_bound` already exist and are already used; the versioned fast clone was simply still calling `lower_for_after_init` and re-deriving `i < arr.length` per iteration. The emitted load is also byte-for-byte the idiom already at `loops.rs:3378` (bitcast → `POINTER_MASK_I64` → `inttoptr` → `load I32`), under the same "guard-ok entry, array proven live/dense" rationale, and `ArrayHeader` is `#[repr(C)] { length: u32, capacity: u32 }`, so offset 0 is the length by layout rather than by luck.

The claim that actually needs to hold is "the matched body cannot change its length", so I tested the four ways it could:

| | shape | node | perry |
|---|---|---|---|
| 7 | `a.push(99)` mid-loop — grows | `0,1,2,3,4,5 \| 6` | identical |
| 8 | `a.length = 4` mid-loop — shrinks | `0,1,2,3 \| 4` | identical |
| 9 | rebinding the array local mid-loop | 20 elements | identical |
| 10 | `a[i + 3] = i` — store past the end grows | `0,0.5,1,0,1,2,3 \| 7` | identical |

All four fall out of the match (they are the "calls, alias writes, growth" the matcher excludes) and keep the unhoisted lowering. A GC move changing the address but not the length is right, and is why the hoisted *value* survives.

**Float arithmetic in masked-window stores.** The two predicates are the risk here — `dense_masked_store_rhs_is_admissible` (match time) and `masked_store_rhs_is_genuine_f64` (lowering time) must not diverge, or a NaN-box reaches an `fadd` and propagates its payload as a number. The new arms are structurally identical in both, and every base case is numeric by construction: `Number`/`Integer` literals, `LocalGet` restricted to the counter or an i32 counter slot (so `sitofp i32 -> double`), and `IndexGet` with a static window over a guard-validated packed array (a raw f64 load). So `+` over admitted operands cannot be concatenation — there is no shape here that can produce a string.

The NaN argument in the doc comment is also correct as stated: an op on canonical operands that produces NaN yields the default quiet NaN `0x7FF8…`, below the `0x7FFA–0x7FFF` tag range, and payload propagation from an input NaN is safe because the raw-f64 invariant already bars a boxed-range payload NaN from the slot. Excluding `%` and `**` is the right line — both lower through runtime helpers.

I exercised the IEEE edges through the admitted ops rather than trusting that: `0, -0, ±Infinity, NaN, 5e-324, 1.797…e308, 0.1` cycled through `+ - * /` and unary negation in the masked window, with `Object.is` used to distinguish `-0` from `0` (30 values). Byte-identical to node v26.5.1, all 12 probe cases.

**One thing that is not yours.** `perry --bins` failed once during my run on `archive_cache::…preparation_safety_survives_fallback_and_cache_hit`. Alternating main/PR twice gave 2 failures out of 4 on *each* side, and it passes isolated — a branch-independent flake, filed as #9083. Mentioning it so the next person to see it red here doesn't spend the afternoon I nearly did.

Validation: codegen 1347 passed, `native_proof_regressions` 284 passed, runtime 2814 passed (`RUST_TEST_THREADS=1`), fmt clean, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped.

PR #9071, opened 2026-08-29
perf(codegen): resolve loop-called immutable callee bindings once at entry (captured-arrow calls −45%)

proggeramlug:
## What

Loop-called arrows held in **immutable bindings** — captured, module-global,
or parameter — are resolved once at body entry
(`js_closure_resolve_arrow_direct_call`) and dispatched directly through the
existing `resolved_arrow_callback_targets` machinery, instead of paying the
full `js_closure_callN` dispatcher (two runtime boundaries, strategy dispatch,
implicit-`this` save/restore) on every call of every iteration.

The consuming arm in `lower_call/early_branches.rs` has existed since
#8642/#8705 — only **method** bodies ever populated the map. This adds:

* `collect_loop_called_callee_bindings` — admits `(binding, arity)` pairs
  whose callee is a parameter, captured binding, or module global; never
  assigned in the body **nor anywhere in the module** (a capture or global can
  be written by other bodies; immutability is the entry-resolution identity
  argument); with at least one call site inside a loop to amortize the entry
  call.
* `emit_callee_binding_resolutions` — reads each binding **raw** at entry
  (slot load / capture-slot load / global load; a boxed capture goes through
  the untrusted `js_box_get_bits`, which returns the TDZ sentinel rather than
  throwing). A sentinel or non-closure resolves to null and every call keeps
  its dispatcher fallback, so a body running before a captured binding
  initializes behaves exactly as before. The `Function` type-hint check
  mirrors the consuming arm's own predicate.
* Wiring for plain function bodies (parameters only — that path has no
  module-wide reassignment oracle) and closure bodies (all three sources).
  Async, generator-wrapper, and CPS-step bodies excluded as everywhere else.

## Numbers

Same-build kill-switch A/B (`PERRY_CALLEE_BINDING_RESOLUTION=0`), quiet Linux
host, isolated ops:

| shape | off | on | node 26.5.1 |
|---|---|---|---|
| captured arrow called in loop | 8.2 | **4.6** | 0.5 |
| module-global arrow | 8.1 | **4.6** | 0.6 |
| callback parameter (plain function) | 8.1 | **4.6** | 0.5 |
| capturing arrow | 7.0 | **2.5** | 0.9 |

Every other row of the 12-op codegen sweep is within ±0.1 ns. wolf-ecs (quiet
Mac mini, 11 alternating pairs vs main): add_remove **−0.52%**, entity_cycle
**−0.36%**, 11/11 both in the 2 s window — the benchmark drivers call through
captured bindings.

## Semantics

Differential vs node, identical output: a reassigned module-global callee
observes the new value (excluded from resolution by the collector); an
ordinary function through a binding keeps receiverless `this === undefined`;
bound functions, rest params, and arity mismatches keep the dispatcher; a
callee that throws mid-loop; a capturing arrow reading a capture mutated
between calls; async arrows. Kill-switch build is output-identical. ECS
differential probes vs node unchanged.

## Testing

* `perry-codegen` 1825 / `perry-runtime --lib` 2807 — green.
* Integration (clean pinned-worktree run): `issue_8655_array_subclass_indexing`,
  `issue_8897_field_push_writeback` — 12/12.

Profiled follow-up, not taken here: on the resolved path the residual vs
`fn_decl`'s 0.6 ns is the public trampoline hop; and the statically-known
LOCAL callee path (`closure_direct.*`) still pays `js_closure_call1` /
`js_implicit_this_set` per call in its surrounding dispatch. A separate WIP
branch (`perf/closure-direct-inline-guard`) proved the per-call typed-feedback
guard can be replaced by a two-load inline identity probe (guard hit count 0
under lldb) but measured flat, so it stays unshipped per the campaign's
measured-flat rule.

## Summary by CodeRabbit

  * Improved performance for immutable arrow callbacks invoked inside loops by resolving eligible callbacks once before execution.
  * Retained existing dispatch behavior when callbacks cannot be resolved directly.

* **Bug Fixes**
  * Updated build caching so changes to callback-resolution settings produce distinct build results.

* **Documentation**
  * Added a changelog entry describing the callback resolution improvement.

### DIFF
--- a/changelog.d/9071-callee-binding-resolution.md
+++ b/changelog.d/9071-callee-binding-resolution.md
@@ -1,0 +1,3 @@
+### Changed
+
+- Loop-called arrows held in immutable bindings — captured, module-global, or parameter — are resolved once at body entry and dispatched directly, instead of paying the full `js_closure_callN` dispatcher on every call. The isolated captured-arrow call drops from 8.0 to 4.4 ns/op; a resolution that fails (non-arrow, bound function, reassigned binding, pre-initialization sentinel) keeps the exact dispatcher fallback per call.
--- a/crates/perry-codegen/src/codegen/closure.rs
+++ b/crates/perry-codegen/src/codegen/closure.rs
@@ -1306,6 +1306,23 @@ pub(super) fn compile_closure(
         super::arguments::ArgumentsCallee::CurrentClosure,
     );
 
+    // #9060 follow-up: resolve loop-called immutable callee bindings once at
+    // entry — parameters, captured bindings, and module globals (the
+    // module-wide reassignment oracle is in scope here). `%this_closure` is a
+    // live parameter of every closure body, so capture-slot reads are direct.
+    // Skipped for async bodies: entry SSA values do not survive the CPS
+    // rewrite.
+    if !is_async {
+        let param_ids: std::collections::HashSet<u32> = params.iter().map(|p| p.id).collect();
+        super::helpers::emit_callee_binding_resolutions(
+            &mut ctx,
+            body,
+            &param_ids,
+            Some(module_reassigned_locals),
+            true,
+        );
+    }
+
     if is_async {
         stmt::lower_async_rejecting_stmts(&mut ctx, body)
             .with_context(|| format!("lowering async closure body func_id={}", func_id))?;
--- a/crates/perry-codegen/src/codegen/function.rs
+++ b/crates/perry-codegen/src/codegen/function.rs
@@ -1397,6 +1397,15 @@ pub(super) fn compile_function(
         }
     }
 
+    // #9060 follow-up: resolve loop-called immutable callee bindings once at
+    // entry (parameters only here — the plain-function path has no module-wide
+    // reassignment oracle in scope for captures/globals). Skipped for async
+    // bodies: their entry SSA values do not survive the CPS rewrite.
+    if !f.is_async {
+        let param_ids: std::collections::HashSet<u32> = f.params.iter().map(|p| p.id).collect();
+        super::helpers::emit_callee_binding_resolutions(&mut ctx, &f.body, &param_ids, None, false);
+    }
+
     if f.is_async {
         stmt::lower_async_rejecting_top_level_stmts(&mut ctx, &f.body)
             .with_context(|| format!("lowering async body of '{}'", f.name))?;
--- a/crates/perry-codegen/src/codegen/helpers.rs
+++ b/crates/perry-codegen/src/codegen/helpers.rs
@@ -1692,3 +1692,117 @@ mod native_roots_target_tests {
         );
     }
 }
+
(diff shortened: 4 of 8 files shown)


proggeramlug:
Merged, after replaying it onto current `main` — as pushed it would have reverted **two** merged PRs.

## The revert

The commit was authored over a tree predating #9060 and #9062 but committed onto a newer base, so it carried deletions of files its own parent contains — **477 deletions**:

- #9060: the whole `stable_packed_accumulator.rs` (273 lines) and its build-cache knob
- #9062: `for_multi_decl_tests.rs`, the `issue_9052_for_lexical_declarators.rs` integration test, the `Module` field, its `stable_hash` entry, and ~40 test-fixture updates

Merging `main` in does not fix this — the deletions read as intentional and a merge honours them. So I took the six files that are genuinely this change (`closure.rs`, `function.rs`, `helpers.rs`, `hoisted_callback_calls.rs`, `collectors/mod.rs`, the fragment) and replayed them onto `main`. Net diff is now **+303/−1 with zero deletions**, and the `−1` is your own re-export widening in `collectors/mod.rs`.

This is the third revert-carrier today (#9060 and #9071 by this mechanism, #9021 caught it in its own war story). The check that catches it in one line, after any rebase or squash of a long-lived branch:

```
git diff origin/main --stat --diff-filter=D
```

Anything listed there is something `main` has and your branch does not.

## Two gate failures of its own

- **`PERRY_CALLEE_BINDING_RESOLUTION` was unregistered**, failing `codegen_env_vars_are_build_cache_inputs`. Registered as an **input**: the two settings emit different call sequences, so a cached object from one must not serve the other. That is the fourth unregistered codegen knob today (#9016, #9026, #9060, this) — all four found by `cargo test -p perry --bins`, which no crate-level `--lib` run reaches. Worth adding to your loop.
- **`local_binding_type_audit.py`** wanted the new `local_type_hint` read classified. I recorded it as **`runtime-validated`**, and want to be explicit about the reasoning since it is a soundness claim: the declared function type only narrows which loop-called bindings are *attempted*, and `js_closure_resolve_arrow_direct_call(handle, arity)` validates identity and arity at runtime — so a wrong or stale declared type produces a failed resolution and the ordinary dynamic call, never a wrong callee. Async and generator types are excluded at the read. Correct me if that mischaracterises the intent.

## Behaviour

Probed both knob states against node: basic loop-called callback, **a callee reassigned mid-loop** (the case a hoisted resolution would get wrong — 73, correct), arity mismatch with a defaulted second parameter, a captured callee from an enclosing scope, async/generator bindings, a class-field callee via `this.f`, and a non-function binding behind a `typeof` guard. **Byte-identical both ways.**

Validation: `perry-codegen` 1345/0, `perry-runtime --lib` 2812/0, `perry --bins` 1066/0, `fmt --check`, `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**.

PR #9073, opened 2026-08-29
fix(codegen): dynamic-parent own-ctor classes ran field initializers twice — pi's startup crash

proggeramlug:
The mixin pattern with an explicit constructor:

```js
function mixin(Base) { return class extends Base { m() { return 1; } }; }
const Mixed = mixin(Root);
class Leaf extends Mixed { #s = 5; constructor() { super(); } get() { return this.#s; } }
```

The standalone-ctor lowering picks its pre-body field-staging mode by `extends_name` alone, so an `extends_expr` (dynamic-parent) class with an own ctor fell through to mode `All` and staged its self fields **before the body** — and the body's `super()` lowering staged them again after the parent returned. Three observable failures from the one cause, all verified against node:

- public initializers run **twice** (`init-runs: 2` vs node's 1 — silent double side effects);
- private fields throw `Cannot initialize a private field twice`;
- because the pre-body run executes **before `super()` binds the receiver**, brands can land on the wrong object, and later reads throw `Cannot access private member from an object whose class did not declare it` — **this is pi's startup crash on current main** (the July pi binary predates the `All` staging for these shapes).

Fix: `extends_expr` + own ctor stages `AncestorsOnly` pre-body (the static chain of a purely dynamic parent is empty ⇒ stages nothing, which is correct — everything above the edge belongs to the runtime parent constructor). The self fields keep their single staging at the body's `super()`. The ctor-less sibling (#9045's `FromInclusive` path) is untouched and covered by the fixture.

Verification: gap fixture `test_gap_9051_dynamic_parent_ctor_field_init_once.ts` (public run-count + private read + ctor-less sibling) node-identical; five ad-hoc mixin probes node-identical; perry-codegen lib 1345/0; fmt green. Together with #9069 this un-blocks the pi build+run on current main.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed duplicate field initialization in classes with dynamic parent classes and their own constructors.
  * Prevented errors when initializing private fields in these scenarios.
  * Ensured field brands and values are applied to the correct object after `super()`.
  * Ensured derived classes without explicit constructors initialize fields correctly.

* **Tests**
  * Added regression coverage confirming public and private fields initialize once, remain accessible, and behave correctly after `super()`.


proggeramlug:
Merged. Verified all three observable failures against node: initializer runs exactly once (`inits: 1`), the private field no longer throws, and the brand check lands on the right object (`#b in b1` true, `in {}` and in an unrelated instance false). Also confirmed the paths that already worked are unchanged — no-own-ctor dynamic parent, and an ordinary static `extends`.

This is the `extends_name`-keying hazard in its canonical form. `function mixin(Base) { return class extends Base {...} }` has no literal parent name to key on, so any decision made from `extends_name` silently takes the wrong branch — CLAUDE.md's known-weak-areas list calls this out for native base classes, and it's the same root here for field staging.

## One thing this does not reach — filed separately, not a blocker

**Two levels of dynamic parent SIGSEGVs**, both before and after this PR:

```ts
const Mixed = mixin(Root);
const Mixed2 = mixin(Mixed);          // second level
class Deep extends Mixed2 { d = 4; constructor() { super(); } }
new Deep();                            // node: fine    perry: exit 139
```

A/B'd before attributing: `main` crashes identically, so this PR neither causes nor worsens it — it fixes one level and doesn't extend to the recursive case. Worth knowing here because a reader could reasonably assume "dynamic-parent own-ctor classes" now works generally, and the mixin-chain shape (`compose(A, B, C)`) is common enough that it will resurface.

Validation: `perry-codegen` 1345/0, `perry-runtime --lib` 2812/0, `fmt --check`, `run_lint_gates.sh` **all 60 gates passed; 2 CI-only skipped**.

PR #9074, opened 2026-08-29
ci(simctl): invalidate cached Geisterhand archives

### LINKED ISSUE #9037: ci(simctl): invalidate cached Geisterhand archives when the Perry commit changes
## Problem

`.github/workflows/simctl-tests.yml` caches the entire `target/` directory using only `hashFiles("**/Cargo.lock")`.

On Simulator run 33251438200 for release candidate `ec67a454409530434fa60dbdc468f6e8739929e2`, the cache restored `target/geisterhand/aarch64-apple-ios-sim/release/libperry_runtime.a` built at the previous candidate `88def6d85d4ce5d22cbef3fd8f6548c92c938d59`. The Geisterhand builder checks only whether the archive exists, so it skipped rebuilding it. The compatibility guard then correctly failed:

```
library build: v0.5.1519 (commit 88def6d85d4c)
Perry build: v0.5.1519 (commit ec67a4544095)
```

This makes a Simulator rerun deterministically fail after any source-only candidate update that leaves `Cargo.lock` unchanged.

## Required fix

Before the Geisterhand compile smoke, ensure cached commit-stamped archives cannot be reused across commits. A minimal workflow fix is to remove `target/geisterhand` after cache restore; alternatively key that directory by `github.sha` or make `build_geisterhand_libs` rebuild archives whose embedded build stamp is incompatible.

Keep registry/dependency caching intact so the simulator job does not become unnecessarily cold.

## Acceptance criteria

- A simulator workflow at commit B may restore a cache created at commit A with the same `Cargo.lock`.
- `target/geisterhand/.../libperry_runtime.a` and the paired stdlib/UI archives are rebuilt for commit B.
- `perry compile --target ios-simulator --enable-geisterhand` reaches the link/run stages without a runtime/compiler stamp mismatch.
- Add a workflow/static guard or focused test that prevents the cache key/existence-only behavior from regressing.

## Evidence

- Failing run: https://github.com/PerryTS/perry/actions/runs/33251438200
- Failing job: https://github.com/PerryTS/perry/actions/runs/33251438200/job/99097691287
- Candidate SHA: `ec67a454409530434fa60dbdc468f6e8739929e2`
- Stale archive SHA: `88def6d85d4ce5d22cbef3fd8f6548c92c938d59`

proggeramlug:
Fixes #9037

## Summary
- discard `target/geisterhand` immediately after restoring the simulator Cargo cache
- retain registry, dependency, and other `target` cache entries
- assert the commit-stamped archive subtree is absent before the build and Geisterhand compile smoke

## Testing
- parsed `.github/workflows/simctl-tests.yml` with PyYAML
- statically verified cache → invalidation → Geisterhand smoke ordering
- simulated a restored stale `target/geisterhand/.../libperry_runtime.a` and verified cleanup preserves unrelated `target/release` artifacts
- `git diff --check`

No version bump.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated automated builds to reliably regenerate Geisterhand archives after restoring cached build data.
  * Prevented stale archives from being incorrectly reused during simulator test runs.

### DIFF
--- a/.github/workflows/simctl-tests.yml
+++ b/.github/workflows/simctl-tests.yml
@@ -48,6 +48,14 @@ jobs:
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
           restore-keys: ${{ runner.os }}-cargo-
 
+      # These archives embed the Perry commit, while the builder treats an
+      # existing archive as fresh. Keep the broader target cache warm, but
+      # force this commit-stamped subtree to be rebuilt after every restore.
+      - name: Invalidate cached Geisterhand archives (#9037)
+        run: |
+          rm -rf target/geisterhand
+          test ! -e target/geisterhand
+
       - name: Build perry + iOS UI lib
         run: |
           cargo build --release -p perry -p perry-runtime -p perry-stdlib -p perry-runtime-static -p perry-stdlib-static


proggeramlug:
Merged. Small and correctly placed: the step sits after `actions/cache` restores (lines 41–49) and before the build (line 59), which is the only window where removing the stale subtree does what you want — earlier and the restore would put it back, later and the build would already have consumed it.

The rationale is the interesting part and worth having in the comment as you've written it: the archives embed the Perry commit while the builder treats any existing archive as fresh, so a warm cache silently serves a previous commit's artifact. That's the "cached thing that looks valid but encodes a stale identity" shape — the same reason `PERRY_LLVM_INPROCESS` and the codegen knobs have to participate in the build-cache key (#9044, #9060, #9071 today). Scoping the removal to `target/geisterhand` rather than dropping the whole cargo cache keeps the expensive part warm.

`test ! -e target/geisterhand` after the `rm -rf` is a good touch — it makes the step assert its own effect rather than trusting `rm` to have done it.

No changelog fragment; precedent for CI-only changes is mixed (#8928 had one, #8932 didn't) and this alters no shipped behaviour, so I left it.

Validation: workflow parses, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped.

Issue #9072, opened 2026-08-29
Set/Map forEach with mid-iteration delete visits holes and skips entries (gap test red on main)


proggeramlug:
`test-files/test_gap_set_map_foreach_fused_receiver.ts` fails on clean `main` — cases 5–7 (delete during `Set.prototype.forEach` / `Map.prototype.forEach`) return wrong elements, and `test-parity/gap_snapshot.json` expects the test to pass, so any full gap run is red on it.

Node 26.5.1 (pinned oracle):

```
[1,2] 0
["a:1","b:2","c:3"] 0
["p","q","r"] 2
```

Perry (both `main` @ `86c910471` and a current feature branch, `perry-dev` builds, `PERRY_NO_AUTO_OPTIMIZE=1` — byte-identical on both arms, so pre-existing):

```
[1,null] 1
["a:1","NaN:undefined","b:2"] 1
["p","q"] 2
```

The shape is exactly what the test's own header warns about: delete during forEach used to compact the backing vector and was fixed for that; the current failure smells like the **tombstone-hole delete work** (#9020 Map pattern / recent Set-Map hole handling) changed delete-during-iteration bookkeeping — `[1,null]` (a hole surfaced to the callback), `NaN:undefined` (a hole key/value pair visited), and a stale `size` of 1 after deleting both entries. Case 7 additionally skips `"r"` after deleting an earlier entry.

Found while running the full gap suite for #9066 on 2026-08-29; A/B'd from the merge-base before attributing (both arms identical, so it is not from that PR).

PR #9075, opened 2026-08-29
fix(runtime): make iterator own properties readable; present-undefined .next throws (follow-up to #9066)

proggeramlug:
Follow-up to #9066, carrying the review-round changes that were in flight when the PR merged (plus the review's own findings). The regex-engine cfg fix landed in #9066 via the maintainer's commit; everything else is here.

## The write-only storage bug (the "squeeze loses survivor values" report)

Root-caused: **not the squeeze, and not the values** — they sat intact in the overflow spill the whole time (verified slot-by-slot: `ov6=Some(11.0) ov7=Some(22.0) ov8=Some(33.0)` while every by-name read returned undefined). The Map/Set-iterator arm in the by-name GET tail ended with `if key_bytes != b"next" { return undefined }` — every non-`next` property read short-circuited without consulting own fields, so #9066's storage was write-only through that lane. Some compiled reads worked via IC lanes, which is why single-property probes (and the PR's own gap cases) passed while the 12-property/10-delete shape failed.

The arm now lives in `accessors::map_set_iterator_property` with own-field shadowing first (ordinary [[Get]] order — an own `return` patch now also shadows the synthetic binding), and `get_field_by_name_tail.rs` comes back under the 2000-line cap (1981, was exactly 2000).

## The other two items

- **Present-but-undefined own `next`** (CodeRabbit's inline find on #9066): `it.next = undefined` is present and non-callable — GetV yields the stored undefined and Call throws. The dispatcher probe adds a bytes-based keys presence scan (no allocation; an unpatched iterator's keys edge is null, so the hot path pays one null check) and throws per IteratorNext instead of silently running the builtin advance.
- **Seed-failure hardening**: if the reserved-floor seed cannot allocate, the by-name append and the defineProperty keys claim now DROP the write instead of proceeding unseeded onto field 0.

## Validation

- `test-files/test_gap_iterator_patched_next.ts` extended to 16 cases (N: 12 adds + 10 deletes with survivor-value assertions across the hole-squeeze; O: own `return` shadowing; P: `it.next = undefined` throws under `for…of`) — **byte-identical against the pinned Node 26.5.1 oracle**, on a base that includes #9067's shared-shape-index-across-delete work.
- `RUST_TEST_THREADS=1 cargo test -p perry-runtime --lib`: 2793 passed, 0 failed (6 `reserved_floor` tests incl. `user_properties_read_back_through_the_get_path_at_scale`, the exact reviewer repro shape).
- `cargo check -p perry` (the feature-off configuration that caught the cfg break): clean.
- `raw_handle_debt.py` bare run: 964 vs baseline 967, all ceilings; no new bare reads.
- **Full gap suite on this head** (pinned Node 26.5.1): effective **574/585**. The run initially reported 26 compile failures; every one traced to a single local-harness artifact — the changelog-fragment commit landed after the `PERRY_BIN` build, so the per-test auto-optimize arm built archives stamped with the newer HEAD commit and Perry's runtime-coherence check refused the pair ("runtime library does not match this Perry compiler"). Rebuilt at HEAD, all 26 pass through the harness, including the entire http/net/fetch/wasm set and both zlib tests. The 11 residual failures are all pre-existing and accounted: the 5 `gap_snapshot.json` known-fails, the 5 npm-package local-env fails, and `set_map_foreach_fused_receiver` (#9072, pre-existing on main).
- `scripts/run_lint_gates.sh`: **all 60 gates green** on this head.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed property reads on Map and Set iterators so custom properties can be retrieved correctly.
  - Custom properties now properly override built-in iterator methods.
  - Explicitly setting an iterator’s `next` property to `undefined` now correctly raises a `TypeError`.
  - Prevented property writes from being lost when reserved storage cannot be initialized.
  - Preserved custom properties and iteration behavior after properties are deleted.

- **Tests**
  - Added regression coverage for custom iterator properties, method overrides, deletion scenarios, and invalid `next` overrides.

### DIFF
--- a/changelog.d/9075-iterator-own-prop-reads.md
+++ b/changelog.d/9075-iterator-own-prop-reads.md
@@ -1,0 +1,7 @@
+### Fixed: iterator own properties were write-only through the by-name GET; `it.next = undefined` silently ran the builtin (follow-up to #9066, PR #9075)
+
+The Map/Set-iterator arm in the by-name GET tail returned `undefined` for every non-`next` key without consulting own fields, so the #9066 reserved-floor storage was write-only through that lane — user properties stored past the floor (and hole-squeeze survivors) read back `undefined` while their values sat intact in the overflow spill. The arm moved to `accessors::map_set_iterator_property` with own-field shadowing first (ordinary [[Get]] order, so an own `return` patch also shadows the synthetic bound method).
+
+Also per review: an own `next` explicitly assigned `undefined` is present-but-non-callable and now throws per IteratorNext (bytes-based presence scan, no allocation on the unpatched hot path), and a failed reserved-floor seed drops the write instead of proceeding unseeded onto the backing-collection field.
+
+Validated byte-for-byte against the pinned Node 26.5.1 oracle (16 gap cases, including the reviewer's 12-add/10-delete survivor shape) plus 6 `reserved_floor` unit tests.
--- a/crates/perry-runtime/src/object/field_get_set/accessors.rs
+++ b/crates/perry-runtime/src/object/field_get_set/accessors.rs
@@ -136,6 +136,46 @@ pub(crate) unsafe fn own_data_field_by_name(
     None
 }
 
+/// #2856 synthetic method reads + #9019 own-field shadowing for a Map/Set
+/// iterator receiver's property GET (extracted from
+/// `get_field_by_name_tail.rs`, which sits at the file-size cap).
+///
+/// Ordinary [[Get]] order: an OWN property — user code can store one past
+/// the reserved floor since #9019 — shadows every synthetic method. This is
+/// also what makes user data properties on iterators readable at all: the
+/// old arm returned `undefined` for every non-`next` key without consulting
+/// own fields, so a stored value was write-only. Then the legacy bound
+/// synthetic methods (`return`/`throw`/`@@iterator`); `next` deliberately
+/// resolves through the caller's generic scans (`None`) so
+/// `iterator.next.call(other)` receives `other` and brand-checks; any other
+/// key is absent (`Some(undefined)`).
+pub(crate) unsafe fn map_set_iterator_property(
+    obj: *const ObjectHeader,
+    key: *const crate::StringHeader,
+) -> Option<JSValue> {
+    if let Some(v) = own_data_field_by_name(obj, key) {
+        return Some(v);
+    }
+    let key_ptr = (key as *const u8).add(std::mem::size_of::<crate::StringHeader>());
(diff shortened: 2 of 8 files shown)


proggeramlug:
Merged. The core observation is a good one and worth restating, because it is the kind of thing that reads as correct right up until you write the test:

> `it.next = undefined` is PRESENT but non-callable, which the value read alone cannot distinguish from absence.

That is exactly right per the spec — GetV yields the stored `undefined`, and Call then throws — and a `!= TAG_UNDEFINED` check collapses the two cases into "absent", silently falling back to the builtin advance. Reaching for a keys-array presence scan instead of the value is the right discriminator.

Extracting the GET arm into `accessors::map_set_iterator_property` preserves the resolution order exactly — I diffed it arm by arm: own field first (the new part), then the `return`/`throw`/`@@iterator` binds, `next` → `None` so the generic scans keep `iterator.next.call(other)` brand-checking, and every other key → `Some(undefined)` as before. The write-only bug it fixes is a nice catch: returning `undefined` for every non-`next` key without consulting own fields meant stored properties could never be read back, so #9066's seeded storage looked like it was losing values when they were sitting in the overflow spill the whole time.

The `own_present` hot path costs what the comment says. Rust's `||` short-circuits, so for an *unpatched* iterator (`own` is undefined) the block does run — one `object_keys_array` call and a null check, no allocation — and for a patched one it is skipped entirely.

A/B against main, 15 shapes, node v26.5.1:

| | shape | main | this PR |
|---|---|---|---|
| 5 | `it.next = undefined; it.next()` | returns `{value:1,done:false}` | **TypeError** ✓ |
| 6 | `it.next = undefined; for…of it` | iterates `[1,2,3]` | **TypeError** ✓ |
| 11 | own `return` shadows the synthetic | `undefined` | **1234** ✓ |
| 10 | `typeof it.return` | `"function"` | `"function"` (node: `"undefined"`) |

Rows wrong: main 4/15 → this PR 1/15, no regressions.

Row 10 is **not yours** — the `return`/`throw`/`@@iterator` bind arm is copied verbatim from main and behaves identically on both sides. It is a real (small) spec gap though: `%SetIteratorPrototype%` has no `return`, so node answers `"undefined"` where we synthesize a bound method. Filed as #9086 so it doesn't get lost in the extraction.

Validation: runtime 2816 passed (`RUST_TEST_THREADS=1`), codegen 1347, perry --bins 1066, fmt clean, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped.

PR #9076, opened 2026-08-29
fix(runtime): skip Map and Set tombstones in forEach

### LINKED ISSUE #9072: Set/Map forEach with mid-iteration delete visits holes and skips entries (gap test red on main)

proggeramlug:
Fixes #9072.

`Map.prototype.forEach` and `Set.prototype.forEach` now walk the raw `used` extent and skip tombstone slots instead of bounding raw reads by the live `size`. This preserves insertion-order traversal when callbacks delete the current or an earlier entry, while still visiting callback-side appends.

Regression coverage exercises current-entry deletes for Map and Set plus deletion of an earlier Set entry.

Validation on `root@perrymaster.skelpo.net` (final head `241a16537` unless noted):
- focused gap parity fixture: PASS (100%, byte-identical to Node)
- focused Map/Set callback-delete tests: 2 passed
- Map and Set tombstone modules: 12 passed
- full `perry-runtime` run: 2,791 passed; two unrelated parallel/shared-host flakes both passed immediately in isolation
- thread-local policy checker: PASS
- `cargo clippy -p perry-runtime --lib --tests`: completed with pre-existing warnings only

Includes `changelog.d/9076-map-set-foreach-tombstones.md`. No version bump.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `Map.prototype.forEach` and `Set.prototype.forEach` when callbacks delete entries during iteration.
  * Deleted entries are now skipped correctly without skipping later live entries or exposing empty slots.
  * Entries added during iteration continue to be visited in insertion order.

* **Tests**
  * Added coverage for deletion scenarios during `Map` and `Set` iteration.

### DIFF
--- a/changelog.d/9076-map-set-foreach-tombstones.md
+++ b/changelog.d/9076-map-set-foreach-tombstones.md
@@ -1,0 +1,6 @@
+### Fixed
+
+- `Map.prototype.forEach` and `Set.prototype.forEach` now skip tombstones left
+  by callback-side deletes and continue through the raw insertion-order extent,
+  so deleting the current or an earlier entry no longer exposes hole values or
+  skips later live entries.
--- a/crates/perry-runtime/src/map.rs
+++ b/crates/perry-runtime/src/map.rs
@@ -3052,16 +3052,15 @@ fn js_map_foreach_impl(
         // The collection itself is the third callback argument and the
         // identity user code compares `self === m` against.
         // ECMA-262 24.1.3.5: forEach iterates [[MapData]] in insertion order,
-        // re-reading the live entry count each step. Entries appended during
-        // the callback (`map.set` inside the callback) MUST be visited, so the
-        // loop bound is re-evaluated against `(*map).size` every iteration
-        // rather than snapshotting the initial size — see the
-        // `iterates-values-added-after-foreach-begins` / `deleted-values`
-        // Test262 cases.
+        // and `used` is the raw [[MapData]] extent. It includes tombstones left
+        // by callback-side deletes, while `size` is only the live count. Walk
+        // the raw extent, skip holes, and re-read it each step so callback-side
+        // appends are visited too. Bounding the walk by `size` exposed holes
+        // and truncated later entries (#9072).
         let mut i = 0usize;
         loop {
             let map = map_handle.get_raw_const_ptr::<MapHeader>();
-            if i >= (*map).size as usize {
+            if i >= (*map).used as usize {
                 break;
             }
             // Re-derive the collection identity each step from a rooted handle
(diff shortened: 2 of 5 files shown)


proggeramlug:
Merged. The premise check first, because this PR inverts a comment rather than adding to it — the code it deletes said:

> Deleting an entry compacts the backing vector (later entries shift left).

That stopped being true when `js_set_delete` became the O(1) ordered delete: survivors keep their raw indices and the slot takes `SET_HOLE_VALUE_BITS` through the barriered store. So the compensating "only advance when slot `i` still holds the value just visited" heuristic was reasoning about a data structure that no longer exists, and bounding by `size` while the extent is `used` is what surfaced holes and truncated the walk. Walking `used` and skipping `SET_HOLE_VALUE_BITS` is the correct reading of the current representation.

Dropping the `value_handle` root is fine: it only ever protected the post-callback slot comparison this deletes. The `args` array was always built from the raw `ptr::read`, never from the handle, so the argument-passing rooting posture is unchanged.

I A/B'd against main (4 mutation modes × {Set, Map} × n ∈ {3, 20, 40} = 26 rows, node v26.5.1):

| | rows wrong | rows leaking a raw hole marker into user code |
|---|---|---|
| main | 26/26 | 18 |
| this PR | 8/26 | **0** |

The leak is the headline: on main `s.forEach` hands the callback the reserved tombstone, so `seen.join(",")` comes back as `0,,1,,2,` — user code observes an internal sentinel. That is gone completely.

**One residual, which I've filed separately as #9082 rather than holding this up** — it is strictly smaller than what the PR fixes and is not introduced by it. `js_set_delete` self-heals when `used >= 16 && size < used / 2`, and that compaction shifts survivors left and resets `used` *while the walk is in flight*, so `i` lands past the new extent:

```
n=20, callback deletes the entry it is visiting:
  after the 11th delete → size=9, used=20 → 9 < 10 and 20 >= 16 → compact → used=9
  loop resumes at i=11 → 11 >= 9 → break, 9 live entries never visited
```

which is exactly the observed `0..10 | size=9`. The n=3 rows are all exact precisely because `used >= 16` is never reached — so the new tests, at 3 elements, can't see this. Worth adding a ≥16-element delete-during-iteration case when #9082 is fixed; the fix is presumably to have the walker rebase its cursor (or to suppress self-healing compaction while an iteration is active).

Validation: runtime 2814 passed (`RUST_TEST_THREADS=1`), codegen 1347, perry --bins 1066, fmt clean, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped.

PR #9077, opened 2026-08-29
fix(codegen): preserve undefined from OOB byte reads

### LINKED ISSUE #9039: Out-of-bounds Uint8Array read pushed into an array stores 0 instead of undefined

proggeramlug:
## Summary

- keep out-of-bounds-capable Uint8Array and Buffer reads out of the unconditional integer-local proof
- route byte-read locals through the observation-constrained i32 analysis so coercing consumers stay optimized while bare consumers preserve undefined
- add unit coverage plus a Node-parity regression for the array-push reproducer

## Tests

- cargo fmt --all -- --check
- cargo test -p perry-codegen --lib (1346 passed, 1 ignored)
- ./run_parity_tests.sh --filter uint8array_oob_push_9039
- ./run_parity_tests.sh --filter uint8array_nonnumeric_key_7700

Closes #9039

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed out-of-bounds `Uint8Array` and `Buffer` reads so they remain `undefined` when stored in regular arrays, instead of becoming `0`.
  * Corrected type handling for byte reads while preserving expected behavior when values are explicitly coerced to integers.

* **Tests**
  * Added coverage for out-of-bounds typed-array and buffer reads, including array contents and reported value types.

### DIFF
--- a/changelog.d/9077-byte-read-oob-undefined.md
+++ b/changelog.d/9077-byte-read-oob-undefined.md
@@ -1,0 +1,4 @@
+### Fixed
+
+- Out-of-bounds `Uint8Array` and `Buffer` reads now preserve `undefined` when
+  their value is stored in a plain array instead of silently becoming `0`.
--- a/crates/perry-codegen/src/collectors/int_valued_ta_locals.rs
+++ b/crates/perry-codegen/src/collectors/int_valued_ta_locals.rs
@@ -25,8 +25,8 @@
 //! An integer typed-array or guarded numeric-array element read is numeric only
 //! **in-bounds**. An OOB / negative / fractional index yields **`undefined`**
 //! (a NaN-boxed value), NOT an integer.
-//! (`Uint8ArrayGet` is safely integer-valued because its accessor returns `0`
-//! OOB — a general typed-array read does not.) So marking such a local i32
+//! (`Uint8ArrayGet` / `BufferIndexGet` have the same Number-or-`undefined`
+//! contract as a general typed-array read.) So marking such a local i32
 //! unconditionally would let `let x = S[oob]; console.log(x)` print a number
 //! (`fptosi(undefined)` garbage / the seeded `0`) where JS prints `undefined`.
 //!
@@ -157,6 +157,19 @@ fn uint8array_get_is_byte_read(index: &Expr, numeric_locals: &HashSet<u32>) -> b
     super::uint8array_get_reads_a_byte(index, &mut |id| numeric_locals.contains(&id))
 }
 
+/// Byte element reads have the same possibly-OOB `undefined` seed semantics
+/// as the general integer typed-array reads above. Keep them out of the
+/// unconditional integer-local proof and admit them only through this
+/// analysis, whose observation walk proves that `undefined` and its i32 image
+/// `0` cannot be distinguished.
+fn is_byte_i32_image_seed_read(e: &Expr, numeric_locals: &HashSet<u32>) -> bool {
+    match e {
+        Expr::Uint8ArrayGet { index, .. } => uint8array_get_is_byte_read(index, numeric_locals),
+        Expr::BufferIndexGet { .. } => true,
+        _ => false,
+    }
+}
+
 fn is_bitwise_binop(op: BinaryOp) -> bool {
     matches!(
         op,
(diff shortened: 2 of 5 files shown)


proggeramlug:
Merged — this closes #9039, which I filed, and my original reproducer now matches node byte-for-byte.

The header comment this deletes is the whole bug:

> `Uint8ArrayGet` is safely integer-valued because its accessor returns `0` OOB

Both halves fail. The accessor returning `0` OOB is not a licence to call the read integer-valued — it *is* the defect, because JS says `undefined` there. So the premise was describing the symptom and the conclusion inherited it. Replacing it with "`Uint8ArrayGet` / `BufferIndexGet` have the same Number-or-`undefined` contract as a general typed-array read" puts byte reads under the same rule as every other typed-array element read, which is where they always belonged.

What I like about the shape of the fix: the claim is removed from the three *unconditional* proof sites (`int32_producing_deps`, `is_int32_producing_expr`, `write_establishes_exact_i32`/`additive_write_admissible`) and re-admitted only through `write_is_i32_producing_safe`, whose observation walk proves `undefined` and its i32 image `0` are indistinguishable at every use. That keeps the optimization exactly where it is sound instead of dropping it wholesale — `u[oob] | 0` still gets an i32 slot, `a.push(u[oob])` no longer does.

Verified against node v26.5.1, 9 cases, byte-identical including the ones that must *not* regress:

| | case | node |
|---|---|---|
| A | `a.push(u[99])` ×3 — the #9039 repro | `[null,null,null]` |
| B | `a.push(u[1])` — in-bounds still numeric | `[8,8,8]` |
| C | `u["oops"]` — #7700 non-numeric key | `[null,null,null]` |
| E | `s += u[i]` past the end — coercing consumer | `24`, `NaN` |
| F | `Buffer.from([1,2,3])[10]` — same contract | `[null,null]` |
| G | `u[99]|0`, `u[99]>>>0`, `Number(u[99])` | `0`, `0`, `NaN` |
| H | `dst[i] = u[i+5]` — OOB read into a TA store | `0,0,0` |
| I | `u[99] === undefined`, `!!u[99]` | `true`, `false` |

E and G are the ones that would have shown an over-correction; both keep the coercion. The unit tests cover the discriminating pair directly (bare push observes the distinction, `| 0` does not) and mirror it for `Buffer`.

`numeric_locals` going to `_numeric_locals` in `write_establishes_exact_i32` is fine — the parameter earns its place by keeping the sibling predicates' signatures uniform.

Validation: codegen 1347 passed, runtime 2812 passed (`RUST_TEST_THREADS=1`), perry --bins 1066 passed, fmt clean, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped.

PR #9080, opened 2026-08-29
fix(runtime): expose iterator helper next values

### LINKED ISSUE #9068: Iterator-helper objects: .next value read returns undefined (bind-delegation impossible)
Reading `.next` as a VALUE off an iterator-helper object returns `undefined`; Node 26.5.1 returns the function.

```ts
const h: any = [1, 2].values().map((x: number) => x * 2);
console.log(typeof h.next);   // node: "function"   perry: "undefined"
const orig = h.next.bind(h);  // node: ok           perry: TypeError: Bind must be called on a function
```

Calling `h.next()` works fine — the class-id method tower dispatches it — only the property **read** misses, because `ITERATOR_HELPER_CLASS_ID` objects have no attached prototype carrying `next` and no bound-method synthesis on the read path. Consequence: the `orig = it.next.bind(it)`-style delegation that #9019/#9066 made work on the array/map/set/string/buffer/regexp families is impossible on helpers (an own `next` patch on a helper *is* honored since #9066 — only the read of the builtin is missing).

Found while validating #9066 on this exact probe; the buffer and regexp-string families resolve the read fine, so this is helper-specific. Pre-existing relative to #9066 (that PR touches the write/dispatch paths only, not property reads).

proggeramlug:
Fixes #9068

## Summary

- expose a shared `%Iterator Helper Prototype%` with an inherited callable `next` and the required helper brand check
- route raw built-in array iterators into iterator-helper dispatch and prevent eager `Array.prototype.map` lowering for `.values()`, `.keys()`, and `.entries()` chains
- correct stale HIR/codegen types for array iterator factories and add runtime, lowering, type-analysis, and Node-parity regressions

## Testing

Validated on `root@perrymaster.skelpo.net`:

- `scripts/run_lint_gates.sh` — all 60 gates passed; 2 CI-only gates skipped
- `cargo test -p perry-runtime iterator_helpers --no-fail-fast` — 19 passed
- `cargo test -p perry-hir --lib` — 358 passed, 1 ignored
- `cargo test -p perry-codegen --lib` — 1345 passed, 1 ignored
- `PERRY_SKIP_BUILD=1 ./run_parity_tests.sh --filter test_gap_iterator_helper_next_value_9068` — 1/1 passed, byte-identical to Node

No version bump is included.

## Summary by CodeRabbit

* **Bug Fixes**
  * Iterator helpers now expose a callable inherited `next` method.
  * Bound `next` methods continue advancing the correct helper, even after overrides.
  * Invalid receivers are rejected with the expected brand-check behavior.
  * Array iterator methods such as `keys()`, `values()`, and `entries()` now preserve iterator-helper behavior through chained operations.
  * Iterator-producing methods are correctly treated as iterators rather than eagerly materialized arrays.

* **Tests**
  * Added coverage for iterator-helper inheritance, chaining, binding, and receiver validation.

### DIFF
--- a/changelog.d/9068-iterator-helper-next-read.md
+++ b/changelog.d/9068-iterator-helper-next-read.md
@@ -1,0 +1,8 @@
+### Fixed
+
+- Iterator-helper objects now inherit a real `Iterator Helper.prototype.next`
+  method. Reading `helper.next` therefore returns a callable with the same
+  receiver and brand-check behavior as Node, so patterns such as
+  `const original = helper.next.bind(helper)` can delegate to the builtin
+  advance instead of failing because the property read returned `undefined`
+  (#9068).
--- a/crates/perry-codegen/src/type_analysis/refine.rs
+++ b/crates/perry-codegen/src/type_analysis/refine.rs
@@ -435,11 +435,14 @@ pub(crate) fn refine_type_from_init(ctx: &FnCtx<'_>, init: &Expr) -> Option<HirT
         | Expr::ArrayWith { .. }
         | Expr::ObjectValues(_)
         | Expr::ObjectEntries(_)
-        | Expr::ArrayEntries { .. }
-        | Expr::ArrayKeys { .. }
-        | Expr::ArrayValues { .. }
         | Expr::StringMatch { .. } => hir_inferred_refinable_type(ctx, init)
             .or_else(|| Some(HirType::Array(Box::new(HirType::Any)))),
+        // #9068: these are iterator objects, despite their Array-producing
+        // receivers. Refining them to Array makes a later `.map()` emit
+        // `js_array_map` against the iterator object's header.
+        Expr::ArrayEntries { .. } | Expr::ArrayKeys { .. } | Expr::ArrayValues { .. } => {
+            Some(HirType::Object(Default::default()))
+        }
         Expr::StringMatchAll { .. } => Some(HirType::Any),
         // TextEncoder.encode(str) — runtime returns a BufferHeader with
         // packed u8 bytes (same shape as `new Uint8Array([...])`). Refining
--- a/crates/perry-hir/src/analysis/value_types.rs
+++ b/crates/perry-hir/src/analysis/value_types.rs
@@ -778,12 +778,13 @@ pub fn infer_expr_type<F: HirTypeFacts + ?Sized>(expr: &Expr, env: &F) -> Type {
         Expr::ArrayReverseValue { receiver } | Expr::ArrayCopyWithinValue { receiver, .. } => {
             infer_expr_type(receiver, env)
         }
-        Expr::ArrayEntries(array) => Type::Array(Box::new(Type::Tuple(vec![
-            Type::Number,
-            array_element_type_from_expr(array, env),
-        ]))),
-        Expr::ArrayKeys(_) => Type::Array(Box::new(Type::Number)),
-        Expr::ArrayValues(array) => array_type_from_expr(array, env),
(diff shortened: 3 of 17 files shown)


proggeramlug:
Merged, with one commit added — see the end.

Attaching a real `%Iterator Helper Prototype%` is the right shape for #9068. The alternative (synthesizing a bound `next` on the read path, the way `return`/`throw` are handled for Map/Set iterators) would have made `helper.next.call(other)` receive the wrong receiver and skip the brand check; giving helpers a genuine prototype with an intrinsic thunk gets the delegation *and* the brand check for free. `iterator_helper_next_builtin` earning its own entry point is the load-bearing detail — a saved `helper.next.bind(helper)` has to keep advancing the helper it captured even after user code replaces the property, which is only true if the thunk bypasses instance and prototype lookup.

I checked the rooting question this raises, since `alloc_helper` now calls `attach_iterator_prototype` on a raw `obj` from `js_object_alloc`, and that reaches `ensure_iterator_prototypes()` → `build_iterator_prototypes()`, which allocates. That would be a stale-pointer window on the first helper allocation of a process. It isn't one: `build_iterator_prototypes` opens a `GcSuppressScope` for exactly this reason ("keep its raw locals stable across the allocating method/tag installs", #7251), so nothing can move `obj` under the call. Registering `ITERATOR_HELPER_PROTOTYPE_PTR_SLOT` in `gc_runtime_root_holders.json` in the same commit is also the right instinct — a cache of a heap pointer is a root the static checker cannot see.

Behaviour verified against node v26.5.1 — the gap test byte-identical, plus 18 shapes of my own, also byte-identical:

- `typeof h.next === "function"`, `hasOwnProperty(h, "next") === false`, and the prototype owning `next` with `[Symbol.toStringTag] === "Iterator Helper"`, shared across helper instances
- `h.next.bind(h)` delegating across three calls to `{done:true}`
- brand check: `h.next.call([9].values())` throws, `h.next.call(otherHelper)` advances the other helper
- chaining and termination: `map`/`filter`/`take`/`drop`/`flatMap`/`reduce`/`some`/`every`/`find`/`forEach`/`toArray`, `Iterator.from`
- helper methods on plain builtin iterators (`Set.values().map`, `Map.keys().filter`) via the new `maybe_dispatch_helper_on_builtin_iterator`
- prototype identity: the helper prototype's parent is the same `%IteratorPrototype%` the array-iterator family uses
- a user-patched `next` on a helper driving `for…of`

**The added commit.** `perry-runtime --lib` was red on `gc::tests::lazy_intrinsic_towers::realm_owned_intrinsic_module_and_storage_roots_are_distinct`:

```
assertion `left == right` failed: the gate must cover every #8002/#8003 root
  left: 26
 right: 25
```

`ITERATOR_HELPER_PROTOTYPE_PTR` is a new realm-owned root and the PR correctly adds it to `test_realm_owned_root_snapshot`, so the count moves 25 → 26. That is the ratchet doing precisely what it is for — making a new realm root an explicit decision rather than a silent one — so I bumped the expected count rather than loosening the assertion. Worth noting the rest of that test then passes on its own terms: the per-agent distinctness and the `assert_ne!(*a_root, 0, "agent A did not materialize …")` checks both hold for the new entry, so the bump reflects a genuinely realm-owned, per-agent-materialized root and isn't masking a miswired one.

I also rebased onto main (the PR was stale) — `git diff origin/main --diff-filter=D` is empty, so nothing was carried back as an unintended revert.

Validation: runtime 2819 passed (`RUST_TEST_THREADS=1`), hir 358, codegen 1347, perry --bins 1066, fmt clean, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped.

PR #9084, opened 2026-08-29
perf(codegen): packed-loop read admissions — if-conditions, clone accumulator proofs, hazard relaxation

proggeramlug:
Follow-up to #9060/#9063/#9070 — the remaining packed-loop admission residuals (the "compare-only" and "literal-bound" shapes documented in #9060, plus the general accumulator/read-consumption gaps behind them).

## What (four admission fixes, one clone-proof extension)

1. **If-conditions were invisible to the stable-packed matcher** (#9060's documented residual, root-caused there): `stmt_flags` / `leading_read_requires_numeric` matched only Let/Expr/Throw/Return statements, so `if (arr[i] < 0) count++` never admitted — and worse, later reads inside If branches were invisible to the replay-safety check. Both now descend (condition + both branches for flags; condition for the leading numeric walk). `Compare` joins `Binary` as a numeric-consumption context — a wrong hint is a failed `require_numeric` guard → generic loop, never a wrong answer.

2. **Literal-bound loops** (#9060's other residual) turned out to already be served for plain arrays by the range-versioned loop — `for (i = 0; i < 8192; i++) s += a[i]` versions through `packed_f64_range` and now gains the accumulator treatment below. A stable-tier literal-bound arm was built and then WITHDRAWN: it re-claimed five-field object-write bodies that `nested_same_shape_object_writes_version_one_through_four_fields` deliberately keeps outside any clone (tier-boundary creep the pinned counts caught). If subclass receivers ever measure a literal-bound gap, that arm needs a body-shape restriction first.

3. **Numeric accumulators for the plain packed clones** (versioned + range): `s += a[i]` inside a packed fast clone lowered its `+` through `js_dynamic_string_or_number_add` on EVERY iteration, plus two root barriers — the by-construction numeric collector runs before clone facts exist, so `s` had no proof. The packed fast preheaders now run #9060's `collect_numeric_accumulators` + one Number tag test each (a non-Number accumulator takes the slow clone before anything ran), and the ids ride `PackedF64LoopFact.numeric_accumulators`, consulted by `is_numeric_expr` — same mechanism, same kill switch as the stable clone.

4. **Guarded reads are numeric inside clones**: `has_numeric_index_fact` (and the boxed-fallback hazard predicate) now recognize packed-versioned/range facts and masked-window facts — the clone's read either produces a genuine raw double or side-exits BEFORE the value is consumed, so there is no boxed edge. This is what turns `if (a[i] < 0)` into a bare `fcmp` (was `js_rel_lt` per iteration) and feeds the accumulator walk's leaf.

5. **Versioned-loop read bodies take the store arm's relaxed eligibility**: a call-free read body cannot invalidate what the entry guard re-proves, so the whole-function materialization hazard (tripped by the very `new Array(n).fill()` calls that build these buffers) no longer blocks versioning — locally-built arrays now version at all. Same argument, word for word, as the existing store-arm comment.

## Numbers (isolated, dev box; node 26.5 in parens)

| shape | before | after |
|---|---|---|
| `if (a[i] < 0) c++` count loop | 4.58 ns/el | **1.68** (0.62) |
| literal-bound reduce `i < 8192, s += a[i]` | 4.32 | 4.18 (1.01) |
| `i < a.length` reduce (local array) | 5.30 | 4.31 (0.99) |

The count loop's residual is the per-iteration `a.length` IC — #9070's length hoist removes it at merge. The reduce rows' residual is NOT calls (the clones are call-free now — census: fadd + poll only): the accumulator lives in a GC-root slot, and the store-to-load-forward + fadd dependency chain is the floor (~15 cycles/el; node keeps `s` in a register). True reduce parity needs unboxed accumulator slots in clones (plain F64 alloca + boxed write-back at every exit) — scoped as the follow-on mechanism, not attempted here.

## Correctness

Nine-probe differential vs node, byte-identical: counting with negatives/-0/NaN, literal-bound over exact-length and TOO-SHORT arrays (guard fails → slow loop → `undefined` += NaN semantics), string accumulator (tag test → slow clone → concat), mixed-type array, holey array,both-branch If accumulators, and an accumulator reassigned to a string MID-LOOP through a branch.

## Summary by CodeRabbit

- **Performance**
  - Improved optimization of packed numeric loops, including loops with numeric accumulators and guarded reads.
  - Enabled more loops to use fast numeric execution paths while retaining safe fallback behavior.
  - Improved count-loop performance from approximately 4.6 to 1.6 nanoseconds per element in supported cases.

- **Bug Fixes**
  - Improved handling of conditional reads, comparisons, locally created arrays, and call-free read-only loop bodies.
  - Reduced unnecessary rejection of safe packed-loop optimizations while preserving runtime safety checks.

### DIFF
--- a/changelog.d/9084-packed-loop-read-admissions.md
+++ b/changelog.d/9084-packed-loop-read-admissions.md
@@ -1,0 +1,1 @@
+Packed-loop read admissions: if-condition reads admit the stable clone; the plain packed clones gain reduce-accumulator proofs and guarded-read numeric facts (`s += a[i]` is a bare fadd, `if (a[i] < 0)` a bare fcmp inside fast clones); call-free read bodies take the store arm's relaxed hazard eligibility so locally-built arrays version — count loops 4.6 → 1.6 ns/element.
--- a/crates/perry-codegen/src/expr/mod.rs
+++ b/crates/perry-codegen/src/expr/mod.rs
@@ -1912,6 +1912,15 @@ pub(crate) struct PackedF64LoopFact {
     /// RHS is numeric bits (side-exiting otherwise) and skip the per-iteration
     /// store guard — the range guard already proved bounds and mutability.
     pub allow_holes: bool,
+    /// Plain locals the packed fast preheader proved to hold a Number (one
+    /// tag test per admitted accumulator) whose every in-body write is
+    /// numeric-preserving — the packed twin of
+    /// `StablePackedLoopFact::numeric_accumulators`. `is_numeric_expr`
+    /// consults this for `LocalGet`, which is what lets `s += arr[i]` inside
+    /// the fast clone lower to a native `fadd` instead of
+    /// `js_dynamic_string_or_number_add` on every iteration. Scope-safe by
+    /// construction: the fact is pushed around the fast-clone lowering only.
+    pub numeric_accumulators: Vec<u32>,
     /// True when a *range* guard (hole-tolerant or dense) validated the whole
     /// constant-offset index window `[start + min_offset, bound + max_offset)`
     /// at loop entry — `arr[i ± c]` loads may use non-zero offsets even
@@ -2611,7 +2620,7 @@ mod fs_await;
 mod index_get;
 #[cfg(test)]
 mod index_get_claim_tests;
-mod masked_window;
+pub(crate) mod masked_window;
 #[cfg(test)]
 mod null_default_numeric_add_tests;
 
--- a/crates/perry-codegen/src/stmt/loops.rs
+++ b/crates/perry-codegen/src/stmt/loops.rs
@@ -517,6 +517,83 @@ fn lower_numeric_range_add_loop(
     Ok(true)
 }
 
+/// Collect the body's numeric reduce accumulators for a packed fast clone
+/// and emit one Number tag test each in the current (fast preheader) block,
+/// branching to the slow preheader when any holds a non-Number — the
+/// induction base case, exactly like the stable-packed clone's admission.
+/// The returned ids ride the scope's `PackedF64LoopFact`, where
+/// `is_numeric_expr` consults them so `s += arr[i]` lowers to a native
+/// `fadd` instead of `js_dynamic_string_or_number_add` per iteration.
+/// Range-loop wrapper: accumulators are collected against the loop's single
+/// counter-accessed array (the `arr[counter]` leaf of the accumulator walk).
+/// Multiple counter-accessed arrays decline — no admitted leaf would span
+/// them all.
+fn emit_range_loop_accumulator_admission(
+    ctx: &mut FnCtx<'_>,
+    matched: &PackedF64RangeLoop,
+    body: &[Stmt],
+    slow_pre_label: &str,
+    block_prefix: &str,
+) -> Vec<u32> {
+    let mut counter_arrays = matched
+        .arrays
+        .iter()
+        .filter(|access| access.counter.is_some())
+        .map(|access| access.array_id);
+    let Some(array_id) = counter_arrays.next() else {
+        return Vec::new();
+    };
+    if counter_arrays.next().is_some() {
+        return Vec::new();
+    }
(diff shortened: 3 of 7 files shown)


proggeramlug:
Full host gate on `4c3f0a3514` (tree-identical to head `7f9d7c7d27` — the amend only renamed the changelog fragment to the PR number): **all real lint steps pass**, ratchets clean, suites green with only the known parallel-only `stack_maps` flake (passes in isolation). Re-validated post-rebase on current main: count loop 1.65 ns/el, all #9063 masked-store shapes hold at 0.4 ns, nine-probe reduce/count differential and the ten-probe masked differential both byte-identical vs node.

proggeramlug:
Merged. I spent the most time on relaxation #5, since "same argument, word for word, as the existing store-arm comment" is a claim that has to survive the asymmetry between a store body and a read body — a store body's safety argument leans on "in-bounds stores only, cannot grow the array", which a read body doesn't need but also doesn't provide. The argument does carry: what the entry guard re-proves is the receiver and layout, and a call-free body cannot reach anything that would invalidate either, whether it stores or only reads. The `new Array(n).fill()` provenance being what tripped the whole-function materialization hazard is a nice diagnosis — the very idiom that builds these buffers was disqualifying them.

Relaxation #4's soundness rests on the side-exit ordering ("produces a genuine raw double or side-exits BEFORE the value is consumed"), and the accumulator admission is the right shape for it: one Number tag test per accumulator in the fast preheader, branching to the slow preheader before the first fast iteration, so the induction base case is established while nothing has run yet.

I tested that induction where it would break — an accumulator that isn't a Number, or stops being one:

| | shape | node | both arms |
|---|---|---|---|
| 5 | `let s = "x"; s += a[i]` | `"x123"` | ✓ |
| 6 | `s = 0; if (i===1) s = "s"; s += a[i]` | `"s23"` | ✓ |
| 7 | BigInt accumulator | `"16"` | ✓ |
| 9, 13, 14 | mixed arrays (`"x"`, booleans, `null`/`undefined`) | `"3x4"`, `2`, `"NaN"` | ✓ |
| 11, 26 | holes / `new Array(4)` | `1`, `4` | ✓ |
| 20 | `a.length = 4` mid-loop | `[4,4]` | ✓ |
| 22, 23 | aliased writes / element becomes a string mid-loop | `[10,"9,9,9,9"]`, `"3z"` | ✓ |
| 15–17 | `±0`, `±Infinity`, `NaN`, `1e308`, `5e-324` through reads and comparisons | — | ✓ |
| 18, 19 | OOB reads and OOB in an if-condition | `[1,2,3,null,null]`, `1` | ✓ |

28 cases, byte-identical to node v26.5.1 on **both** arms — so no correctness delta, which is what a perf PR should show.

Then I checked the optimization is actually live rather than merely harmless, because "nothing broke" is equally consistent with "nothing happened". On a fixture matching the targets (local `new Array(n).fill()` buffers, 4096 elements × 4000 reps), best of 3:

| | main | this PR | node |
|---|---|---|---|
| `if (a[i] < 0) c++` | 69 ms | **31 ms** | 17 ms |
| `if (a[i] < 3) s += a[i]; else if (a[i] > 5) s -= a[i]` | 141 ms | **39 ms** | 20 ms |
| `s += a[i]` | 17 ms | 16 ms | 18 ms |

2.2× and 3.5×, moving the branchy loop from 7× node to 2× node. The IR agrees: `fcmp olt` 54 → 58 and `fadd double` 10 → 14 for the same fixture, with `js_rel_lt` and the add helper unchanged at 10 and 4 — exactly the signature of versioning, where the slow clone keeps its helpers and the new fast clones get bare float ops. Both arms print identical values.

My first attempt at this measurement used a module-level array and moved the IR by one instruction, which would have let me report "engaged" off noise — worth mentioning in case anyone reuses the fixture.

Validation: codegen 1347 passed, `native_proof_regressions` 284 passed, runtime 2819 passed (`RUST_TEST_THREADS=1`), perry --bins 1066 passed, fmt clean, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped.

PR #9085, opened 2026-08-29
fix(runtime): class-ref receiver satisfies its own private brand check — pi's startup crash (3/3)

proggeramlug:
The last of pi's three bring-up walls (with #9069 and #9073). **lru-cache's construction gate**, minified into pi's bundle:

```js
var make = () => class c {
  static #o = false;
  static create() { c.#o = true; const i = new c(); c.#o = false; return i; }
  constructor() { if (!c.#o) throw new TypeError(...); }
};
```

Codegen emits the bare class REF (INT32-tagged template id) as the receiver of `c.#o`. The ref carries no per-evaluation brand, so both evaluation-brand paths in `private_evaluation_brand_matches` compared `None` against the current evaluation's brand and threw `Cannot access private member from an object whose class did not declare it` at pi startup. Top-level class expressions dodge it (the lexical-brand machinery doesn't engage), which is why simple probes all passed — found by instrumenting the runtime throw to print the field name + ids (`name=#o`, declaring == receiver id) and grepping the bundle.

**Fix:** accept a class-ref receiver whose id equals the declaring class at the top of the matcher. Sound because a private access with the self-ref as receiver can only be emitted from inside that class's own body (outside it, `c.#o` is a syntax error), where the self-binding denotes the current evaluation — the same verdict the pre-brand static fallback always gave.

**Verification:** gap fixture `test_gap_9052` — two separate evaluations each gating on their own `#o`, plus the direct-`new` throw — node-identical; the closure-nested and top-level repro battery node-identical; 2807 runtime tests; fmt green. With all three fixes pi compiles AND boots (`--version` runs; GC-share numbers to follow on the PRs' shared branch).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed static private member access when referenced through a class’s own self-reference.
  * Prevented valid class operations from incorrectly failing with a declaration error.
  * Improved behavior for independently evaluated, closure-nested named classes.
  * Added coverage ensuring valid gated instances work while direct construction remains rejected.

### DIFF
--- a/crates/perry-runtime/src/object/field_get_set/ic_miss/private_member_access.rs
+++ b/crates/perry-runtime/src/object/field_get_set/ic_miss/private_member_access.rs
@@ -270,6 +270,21 @@ fn private_evaluation_brand_matches(
     brand_owner: f64,
     declaring_class_id: u32,
 ) -> Option<bool> {
+    // A bare class REF receiver names the class's lexical self-binding
+    // (`class c { static create() { c.#o = ... } }` — the lru-cache guard
+    // pattern, minified into pi's bundle). Codegen hands the runtime the
+    // template ref, which carries no per-evaluation brand, so both brand
+    // paths below compared None against Some(_) and every closure-nested
+    // class expression's static private access threw "did not declare it".
+    // A private access with the ref as receiver can only be emitted from
+    // inside the class's own body (outside it, `c.#o` is a syntax error),
+    // where the self-binding denotes the CURRENT evaluation — the exact
+    // verdict the pre-brand static fallback (`class_ref_id(obj) == id`)
+    // always gave.
+    if crate::object::native_module::class_ref_id(obj) == Some(declaring_class_id) {
+        return Some(true);
+    }
+
     if let Some(expected) = current_private_lexical_brand(declaring_class_id) {
         let actual = private_evaluation_brand(obj, declaring_class_id);
         return Some(actual == Some(expected));
--- a/test-files/test_gap_9052_class_expr_self_ref_static_private.ts
+++ b/test-files/test_gap_9052_class_expr_self_ref_static_private.ts
@@ -1,0 +1,19 @@
+// lru-cache's construction-gate pattern (minified into pi's bundle): a
+// closure-nested NAMED class expression whose static method flips a static
+// private through the lexical self-binding. Codegen hands the runtime the
+// bare class REF as the receiver; the per-evaluation brand paths compared
+// its (nonexistent) brand against the current evaluation's and threw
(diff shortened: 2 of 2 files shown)


proggeramlug:
Merged. The reasoning in the comment is the part I wanted to check hardest, because "can only be emitted from inside the class's own body" is doing all the work:

> A private access with the ref as receiver can only be emitted from inside the class's own body (outside it, `c.#o` is a syntax error), where the self-binding denotes the CURRENT evaluation.

The syntax-error argument holds, and the narrowing is tight — `class_ref_id(obj) == Some(declaring_class_id)` short-circuits only when the ref names the *declaring* class, so a ref to any other class still falls through to the brand paths. I verified the rejection still fires: `A.read(B)` across two classes with same-named `#p`, and `o.#p` on a plain object, both still `TypeError`.

What surprised me is how much this fixes. I A/B'd 12 shapes against node v26.5.1 and **main throws on 11 of them** — including a plain named class declaration:

```ts
class N { static #k = 3; static get() { return N.#k; } }
N.get();     // node: 3      main: TypeError      this PR: 3
```

Rows wrong: main 11/12 → this PR 4/12. Fixed here: the lru-cache gate itself, direct `new` still throwing, instance privates alongside statics, an inner class reading `Outer.#o` by name, plain named-class self-reference, and a class *declaration* returned from a factory.

**The four residuals are pre-existing, not introduced** — main throws on every one of them, so nothing regressed. I filed them as #9089 rather than holding this up, since they need the class ref to carry per-evaluation identity and that information isn't in the ref today:

```ts
const make = (seed) => class c { static #v = seed; static get(){return c.#v} static set(x){c.#v=x} };
const A = make(10), B = make(20);
A.set(99);
[A.get(), B.get()]        // node: [99, 20]   here: [99, 99]  — two evaluations share
```

and a compound update through the self-binding writes `undefined` (`c.#n++` then `c.#n` → `undefined`, node `1`), which may be a separate defect since plain read and plain write both work. The issue carries the decomposed repros plus the list of shapes that are correct now and must stay so.

I want to be straight about the trade, since it is a judgement call rather than a formality: this converts a loud `TypeError` into silently-shared state for the multi-evaluation case. That is normally the wrong direction. Here it is still clearly right — the shapes it fixes are the common ones (a named class reading its own static private was *entirely broken*), the pattern that regresses to silence requires two live evaluations of one class expression with mutable static private state, and pi cannot start without this. #9089 tracks closing the gap properly.

Validation: runtime 2819 passed (`RUST_TEST_THREADS=1`), fmt clean, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped.

Issue #9087, opened 2026-08-29
correctness: class reference + number treats the INT32-boxed class ref as a number (should be string concat)


proggeramlug:
## Repro

```ts
class K { static m(): number { return 1; } }
let s: any = K;
for (let i = 0; i < 8; i++) s += i;   // also reproduces without the loop: K + 1
console.log(typeof s, String(s).slice(0, 20));
```

node 26.5.1: `string class K { static m()` — `+` on a class does ToPrimitive → the function source string → concatenation.

perry (main `b2f163c5a0`, also with `PERRY_PACKED_LOOP_NUMERIC_ACCUMULATOR=0`, so unrelated to #9060): `number 29` — the class reference participates as a NUMBER.

## Analysis

A `ClassRef` is NaN-boxed with `INT32_TAG` carrying the registered class id (see the #321 note on `value_bits_to_number` in `array/header.rs`, which added `is_class_id_registered` exactly to keep class refs off numeric paths — for array slots). The dynamic `+` path (or an upstream numeric-operand classifier) apparently lacks that exclusion, so `class + number` takes the numeric arm instead of ToPrimitive/concat.

Found while adversarially testing #9060's accumulator admission (which is not implicated — its tag test rejects the whole NaN-box band including INT32); reported by the ECS/ops perf campaign, session https://claude.ai/code/session_019WVcWKmYsUBnnFB7nBgbBJ. Repro file preserved at `/private/tmp/acc-int32.ts` shape above.

The same missing exclusion may affect other INT32-band consumers (`-`/`*` coercions, comparisons); worth auditing `is_class_id_registered` call sites vs the numeric-classifier entry points.

proggeramlug:
Consolidation note from #9088 (duplicate, closed): every tag dispatch that treats INT32-band bits as a plain number likely needs the same registered-class-id audit — not only the dynamic `+` path. Candidates: `-`/`*` coercions, relational compares, and any `(bits & TAG_MASK) == INT32_TAG` arm that skips `is_class_id_registered`.

PR #9091, opened 2026-08-29
perf(codegen): unboxed reduce accumulators in packed fast clones

proggeramlug:
Stacked on #9084 (merge that first) — the follow-on mechanism its PR text scoped: the reduce rows' remaining gap was not calls but the accumulator's GC-root slot.

## What

After #9084, a packed fast clone's `s += a[i]` is a bare guarded `fadd` — but `s` still lived in its GC-root nanbox slot, so every iteration paid a store-to-load-forward + fadd dependency chain (~15 cycles/element; node keeps `s` in a register). Profiling showed the whole loop in 4 PCs with the slot chain as the floor.

This moves each admitted accumulator into a **plain addrspace-0 F64 alloca** for the clone's duration:

- The fast preheader's existing admission (one `emit_js_value_is_number` tag test per accumulator — which IS the strict genuine-double window: the 0x7FF9..0x7FFF band covers every boxed tag, so INT32-boxed numbers correctly fail to the slow clone) now also stores the tested value into the alloca and registers it in `ctx.numeric_accumulator_f64_slots`.
- In-clone `LocalGet`/`LocalSet` of the accumulator redirect to the alloca — no shadow bookkeeping (the real slot holds a stale *number* for the clone's duration, consistent with any prior shadow state; scanning a number nanbox is harmless), no barrier (numbers carry no heap edge). mem2reg promotes the alloca to a register.
- **Every clone exit writes the value back**: the fall-through exit, and a per-clone side-exit trampoline that the scope's packed facts carry as their `store_side_exit_label` — so a mid-iteration hole-check or masked-store value-check side exit restores correct slot state before the slow clone re-executes the iteration.
- v1 restricts unboxing to **LocalSet-only** accumulators (`collect_local_writes` check); Update-written ones (`c++`) keep the slot — the Update lowering doesn't consult the redirect, and integer counters are served by the i32-slot machinery anyway.

Admission stays `collect_numeric_accumulators` (#9060) — single source shared with the stable clone, per coordination with its author, who also verified the slot-canonicalization invariant the mechanism inherits (every admitted producer emits canonical raw doubles).

## Numbers (isolated, dev box; node 26.5 in parens)

| shape | #9084 | this PR |
|---|---|---|
| literal-bound reduce `i<8192, s+=a[i]` | 4.16 ns/el | **0.98** (1.01) — ahead of node |
| `i < a.length` reduce (local array) | 4.14 | **1.29** (0.99) |
| module-global reduce (direct-timed probe) | 4.20 | **2.70** (2.85) — ahead of node |

The `a.length` variant's residual is the per-iteration length IC; #9070's hoist removes it at merge. IR census of the fast clone after this PR: one receiver root re-derive, one `fadd`, the loop poll — nothing else.

## Correctness

- Nine-probe reduce/count differential vs node byte-identical (string accumulator → tag-fail → slow clone; mid-loop string reassignment through a branch; literal bound past the array length; holey/mixed arrays).
- The `any`-seeded accumulator repro from the #9087 investigation behaves identically before/after (the class-ref `+` divergence is that pre-existing runtime bug, unchanged by this machinery; the boolean-seeded arm is correct both sides).
- Full perry-codegen suites green; host gate result to follow as a comment.

## Summary by CodeRabbit

  * Improved performance for packed numeric reduction loops by using optimized floating-point accumulation.
  * Added fast paths for numeric accumulators in versioned and range-based loops.
  * Expanded loop optimization support for arithmetic expressions and eligible read-only loop bodies.
  * Numeric values are safely restored when optimized loops exit early.

### DIFF
--- a/changelog.d/9091-unboxed-clone-accumulators.md
+++ b/changelog.d/9091-unboxed-clone-accumulators.md
@@ -1,0 +1,1 @@
+Reduce accumulators in packed fast clones now live in register-promotable unboxed F64 slots for the clone's duration (write-back at every exit) — `s += a[i]` reduce loops reach node parity or better (literal-bound 0.98 ns/element vs Node's 1.01).
--- a/crates/perry-codegen/src/codegen/closure.rs
+++ b/crates/perry-codegen/src/codegen/closure.rs
@@ -1188,6 +1188,7 @@ pub(super) fn compile_closure(
         class_field_loop_facts: Vec::new(),
         element_shape_loop_facts: Vec::new(),
         i32_counter_slots: HashMap::new(),
+        numeric_accumulator_f64_slots: HashMap::new(),
         local_slot_reps: HashMap::new(),
         repsel_context_allows_canonical_i32: repsel_allows,
         // #7109 split the FIELD out of `repsel_context_allows_canonical_i32`;
--- a/crates/perry-codegen/src/codegen/entry.rs
+++ b/crates/perry-codegen/src/codegen/entry.rs
@@ -879,6 +879,7 @@ pub(super) fn compile_module_entry(
             class_field_loop_facts: Vec::new(),
             element_shape_loop_facts: Vec::new(),
             i32_counter_slots: HashMap::new(),
+        numeric_accumulator_f64_slots: HashMap::new(),
             local_slot_reps: HashMap::new(),
             // #7109: this entry body selects canonical i32/u32/Str on the same
             // per-value rules as a function body. Phase 1 (#6903) excluded it
@@ -1595,6 +1596,7 @@ pub(super) fn compile_module_entry(
             class_field_loop_facts: Vec::new(),
             element_shape_loop_facts: Vec::new(),
             i32_counter_slots: HashMap::new(),
+        numeric_accumulator_f64_slots: HashMap::new(),
             local_slot_reps: HashMap::new(),
             // #7109: this entry body selects canonical i32/u32/Str on the same
             // per-value rules as a function body. Phase 1 (#6903) excluded it
--- a/crates/perry-codegen/src/codegen/entry.rs
+++ b/crates/perry-codegen/src/codegen/entry.rs
@@ -879,7 +879,7 @@ pub(super) fn compile_module_entry(
             class_field_loop_facts: Vec::new(),
             element_shape_loop_facts: Vec::new(),
             i32_counter_slots: HashMap::new(),
-        numeric_accumulator_f64_slots: HashMap::new(),
+            numeric_accumulator_f64_slots: HashMap::new(),
             local_slot_reps: HashMap::new(),
             // #7109: this entry body selects canonical i32/u32/Str on the same
             // per-value rules as a function body. Phase 1 (#6903) excluded it
@@ -1596,7 +1596,7 @@ pub(super) fn compile_module_entry(
             class_field_loop_facts: Vec::new(),
             element_shape_loop_facts: Vec::new(),
             i32_counter_slots: HashMap::new(),
-        numeric_accumulator_f64_slots: HashMap::new(),
+            numeric_accumulator_f64_slots: HashMap::new(),
             local_slot_reps: HashMap::new(),
             // #7109: this entry body selects canonical i32/u32/Str on the same
             // per-value rules as a function body. Phase 1 (#6903) excluded it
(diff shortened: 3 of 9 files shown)


proggeramlug:
Head is `a5d2b1aa0c` (rustfmt fix folded in; re-gate running). **Combined integration numbers** — this branch merged with current main (which now has #9070's length hoist) — the full isolated-operation matrix, node 26.5 in parens:

| shape | combined | node |
|---|---|---|
| literal-bound reduce `s += a[i]` | **0.97 ns/el** | 1.01 |
| `i < a.length` reduce | **0.97** | 0.99 |
| `i < a.length` store loop | 0.67 | 0.59 |
| masked arith store `a[i&K]=b[i&K]+c` | **0.48** | 1.33 |
| plain stores (param/global/local) | 0.40 | 0.70 |
| `if (a[i] < 0) c++` count loop | 1.63 | 0.62 |

Both reduce rows are now **ahead of node** — #9070's hoist removed the length IC exactly as scoped, and this PR's register-promoted accumulator was the last serial chain. The count loop's residual is the Update-written (`c++`) accumulator, deliberately excluded from v1 unboxing.

proggeramlug:
Re-gate on `a5d2b1aa0c`: **fully green** — every real lint step passes (formatting ✓), ratchets clean, all suites green with zero flakes this run. Ready for review alongside #9084 (merge #9084 first; this stacks on it).

proggeramlug:
Merged, after rebasing off the merged #9084 and adding a rustfmt commit (three hunks in `entry.rs`/`loops.rs` were mis-indented; `cargo fmt --all -- --check` is a lint gate).

This moves a value out of its GC-root slot into a plain addrspace-0 alloca, which is the shape CLAUDE.md singles out as a known-weak area, so I went at the "every clone exit writes the value back" claim rather than around it.

**First, a correction to my own method, because it nearly produced a wrong review.** I benchmarked three fixtures and measured 1.00x / 1.01x / 1.00x. Before reporting "no improvement" I diffed the emitted IR between arms — **identical**. The optimization never fired on my fixture, so the timing measured nothing. Worse, my original 31-case correctness probe had zero `acc.writeback_exit` blocks too: that validation, GC stress included, was equally vacuous. The admission needs the array to arrive as a **parameter**; a locally-built `new Array(n).fill()` buffer does not qualify (`s3` in my shapes file gets `acc.ok` but no writeback). Anyone re-measuring this should grep `--trace llvm` output for `acc.writeback_exit` before trusting a number.

Re-done against a fixture where the subject is live — 6 `acc.writeback_exit` blocks, in `sum`, `sumSeed`, `sumTwo`:

- **31 cases byte-identical to node v26.5.1.** The load-bearing ones are the side exits with a partial accumulator: a non-number element at index 10 of 32 must deopt carrying 45, and `"45x111213…"` is exactly what both node and this produce. Same for `null`, `{}`, `undefined`, `true` elements, and for non-number *seeds* (`"p"`, `null`, `undefined`, `true`, BigInt) which must take the slow clone from the preheader test. Plus IEEE edges (`1e308`, `NaN`, `±Infinity`, `-0` with `Object.is`, `0.1` accumulation), holes, and repeated calls where a previous writeback must leave the slot correct for the next.
- **GC stress, since an unrooted value's failure mode is a collection landing in its window.** Three seeds at `PERRY_GC_SCHEDULE_RATE=1 PERRY_GC_SCHEDULE_ALLOC_KB=0 PERRY_GC_FORCE_EVACUATE=1`, each moving 16,343 objects, all node-identical; plus a run under `PERRY_GC_VERIFY_EVACUATION=1` with the from-space quarantine at depth 800, clean. The unrooted-alloca detector reports 0 violations (6 gc-capable allocas seen, so it had subjects), and its `--self-test` passes.

Worth recording that `gc_root_dominance_check.py` reported **0 root stores** on my fixtures and said so explicitly rather than returning a green — a pure-numeric loop creates nothing to root. That is the check behaving correctly, not evidence about this PR.

**Performance, on a shape that fires** (4096 elements × 4000 reps, interleaved best-of-5, main confirmed at 0 `writeback_exit`):

| | main | this PR | node | |
|---|---|---|---|---|
| `s += a[i]` | 23 ms | **16 ms** | 58 ms | 1.44x |
| `s += a[i]; p += a[i]*2` | 24 ms | **16 ms** | 156 ms | 1.50x |

So the slot store-to-load-forward chain was real, and this clears it — comfortably ahead of node on both.

**One gap, not blocking.** The rebase dropped `tests/native_proof_regressions/invalidation.rs` from this PR — those additions belonged to #9084 and were already upstream — so as merged this adds no test of its own. The mechanism most worth pinning is the side-exit writeback carrying a partial accumulator (my case 12 above), since a missed exit would show up as a stale total rather than a crash, and only in the deopt path. Worth adding when you next touch this file.

PR #9092, opened 2026-08-29
test(perry): serialize archive cache environment reads

### LINKED ISSUE #9083: Flaky: archive_cache::preparation_safety_survives_fallback_and_cache_hit fails ~50% under parallel -p perry --bins
`commands::compile::link::archive_cache::tests::preparation_safety_survives_fallback_and_cache_hit` fails roughly half the time under the full parallel `cargo test -p perry --bins` run, and always passes in isolation. It is **not** attributable to any branch — I hit it while auditing #9070 and alternated main/PR twice to check:

| run | 1 | 2 |
|---|---|---|
| main | FAILED | FAILED |
| #9070 | ok | FAILED |
| main | ok | ok |
| #9070 | FAILED | ok |

2 of 4 on each side, and isolated it is green:

```
$ RUST_TEST_THREADS=1 cargo test -p perry --bins -- --exact \
    commands::compile::link::archive_cache::tests::preparation_safety_survives_fallback_and_cache_hit
test result: ok. 1 passed; 0 failed; 1065 filtered out
```

The failure is the test's own precondition, not a behavioural assertion:

```
panicked at crates/perry/src/commands/compile/link/archive_cache.rs:438:13:
a valid cache hit must not prepare the archives again
```

That is the tell described in CLAUDE.md's `perry-runtime` note — a fixture precondition failing means another thread perturbed shared state, not that the code under test regressed. The difference is that this is `perry`, not `perry-runtime`, so it is not covered by the `RUST_TEST_THREADS=1` pin that every CI path applies to the runtime crate; `-p perry --bins` runs on the default pool.

Two candidate mechanisms, not yet separated:

- the archive-cache tests share on-disk state under the target dir (a sibling test preparing or invalidating archives concurrently would make "a valid cache hit" stop being valid), or
- mtime granularity — a cache-validity check that compares timestamps can tie when two operations land in the same filesystem tick, which would explain why a *faster* run is likelier to fail.

Worth noting this is the same crate as the `perry` bin tests that mutate global `PATH`, so a per-test-isolated fixture directory (rather than a shared one) is the likely fix. Until then it is a real source of false reds on any PR that happens to touch this crate — I nearly attributed it to #9070.

proggeramlug:
Fixes #9083.

The archive-cache safety test recomputes keys using discovered tool paths, so sibling tests that temporarily replace the process-global `PATH` can turn its expected cache hit into a miss. Acquire the shared test environment lock across the miss/store/hit sequence so all three lookups see one stable tool environment.

Validation:
- focused archive-cache test
- `RUST_MIN_STACK=16777216 cargo test -p perry --bins` (1056 passed)

No version bump.

## Summary by CodeRabbit

* **Bug Fixes**
  * Stabilized archive-cache fallback and cache-hit tests when running in parallel.
  * Prevented intermittent test failures caused by concurrent environment setting changes.

* **Documentation**
  * Added a changelog entry describing the archive-cache test reliability improvement.

### DIFF
--- a/changelog.d/9092-archive-cache-test-isolation.md
+++ b/changelog.d/9092-archive-cache-test-isolation.md
@@ -1,0 +1,5 @@
+### Fixed
+
+- Stabilized the archive-cache fallback/cache-hit test under the parallel `perry`
+  binary suite by serializing its process-global tool-environment reads with
+  sibling tests that temporarily replace `PATH`.
--- a/crates/perry/src/commands/compile/link/archive_cache.rs
+++ b/crates/perry/src/commands/compile/link/archive_cache.rs
@@ -314,6 +314,7 @@ fn hash_field(hasher: &mut Sha256, name: &str, value: &str) {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::test_env_lock::env_lock;
 
     fn key(root: &Path, features: &[String], archives: &[PathBuf], target: &str) -> String {
         let runtime = root.join("runtime.a");
@@ -400,6 +401,10 @@ mod tests {
 
     #[test]
     fn preparation_safety_survives_fallback_and_cache_hit() {
+        // Cache keys include discovered archive tools, so every lookup reads
+        // process-global environment such as PATH. Keep sibling tests from
+        // changing that environment between this test's miss, store, and hit.
+        let _env = env_lock();
         let temp = tempfile::tempdir().unwrap();
         let root = temp.path();
         let runtime = root.join("runtime.a");


proggeramlug:
Merged. Fixes #9083, which I filed this afternoon after nearly misattributing it to #9070.

The diagnosis is right and matches the second of the two mechanisms I guessed: the cache key is recomputed from discovered tool paths, so every lookup reads process-global `PATH`, and six other modules in this crate take `env_lock()` precisely because they swap it. Holding the lock across the whole miss/store/hit sequence — rather than around each lookup — is the correct granularity, since the invariant under test is that all three see *one* environment.

**Getting evidence for this took two attempts, and the first one was misleading.** Ten runs of the full `--bins` suite on each arm gave 0/10 and 0/10 — no failures anywhere, which proves nothing about the fix because the baseline never failed either. When I originally hit this I had concurrent builds running, so I re-ran with six `yes` processes saturating the box (10 cores, load ~20–29):

| | failures |
|---|---|
| main under load | **10/12** |
| this PR under load | **0/12** |

So the flake is load-dependent, which is also why it presented as roughly coin-flip earlier rather than deterministic: it needs the scheduler to interleave this test with a sibling that has `PATH` swapped out. Anyone re-checking this should apply load, or they will measure 0/10 on a broken tree and conclude it's fine.

One note for whoever touches `env_lock()` next: it `.expect()`s on a poisoned lock, so a sibling panicking while holding the environment will now surface here as a poison panic rather than a cache miss. That is an improvement — the message names the real cause — but it does mean this test can fail for a reason that isn't its own.

Validation: 12 loaded runs clean as above, fmt clean, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped.

PR #9093, opened 2026-08-29
fix(runtime): hide collection iterator return/throw methods

### LINKED ISSUE #9086: Map/Set iterator objects expose synthetic .return/.throw; spec has neither
`Map`/`Set`/string iterator objects answer a bound method for `return` and `throw`, but per spec those prototypes have no such properties — only generator objects do.

```ts
const it = new Set([1, 2]).values();
console.log(typeof it.return, typeof it.throw);
```

```
node v26.5.1  undefined undefined
perry         function  function
```

`typeof it[Symbol.iterator] === "function"` is correct and should stay.

### Where

`crates/perry-runtime/src/object/field_get_set/accessors.rs`, `map_set_iterator_property` — the synthetic bind arm:

```rust
let bind_name: Option<&'static [u8]> = match key_bytes {
    b"return" => Some(b"return"),
    b"throw" => Some(b"throw"),
    b"@@iterator" => Some(b"@@iterator"),
    _ => None,
};
```

Noticed while auditing #9075, which moved this arm out of `get_field_by_name_tail.rs` verbatim. It is **not** introduced there — main answers `"function"` too, confirmed by A/B — so this is long-standing and predates that extraction.

### Why it is not purely cosmetic

`return` is protocol-visible: an early `break` out of a `for…of` calls `IteratorClose`, which does `GetMethod(iterator, "return")` and invokes it when present. Synthesizing one means we call into a bound method on paths where the spec performs no call at all, so the observable difference is not limited to `typeof`. Worth checking what the synthesized `return` actually does on an abrupt loop exit before deciding whether to delete the arm or make it spec-shaped.

The likely fix is to drop `return`/`throw` from the bind list and let the generic scans answer `undefined`, keeping `@@iterator`. That needs a check for callers relying on the current behaviour — `dispatch_set_iterator_method`/`dispatch_map_iterator_method` may route `return` through this path for their own close handling.

proggeramlug:
## Summary

Stops ordinary Map and Set iterator objects from exposing generator-only `return` and `throw` methods while preserving their standard iterable identity and normal prototype lookup.

## Changes

- Removes the synthetic `return` / `throw` bindings from Map and Set iterator property reads.
- Falls through to ordinary prototype lookup for those names, preserving own and inherited user overrides.
- Adds executable regression coverage for Map, Set, string, and generator iterator behavior.

## Related issue

Fixes #9086

## Test plan

- [x] `cargo test -p perry --test issue_9086_collection_iterator_methods -- --nocapture`
- [x] (if user-facing) Added a `#[test]` integration regression in the affected crate.
- [x] (if CLI / stdlib / runtime API changed) Documentation is not needed; this restores existing ECMAScript behavior without adding an API.
- [ ] (if touching a platform UI backend) Not applicable.

## Screenshots / output

The regression test verifies that ordinary iterator `return` / `throw` reads are `undefined`, `in` reports them absent, `Symbol.iterator` remains callable and returns the receiver, own/inherited overrides still resolve, and generator methods remain functions.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Collection and string iterators no longer incorrectly expose generator-only `return` and `throw` methods.
  - Iterator property lookup now correctly respects inherited prototype methods and user-defined overrides.
  - The standard `[Symbol.iterator]()` behavior remains unchanged.
  - Generator iterators continue to provide their expected `return` and `throw` methods.

- **Tests**
  - Added regression coverage for map, set, string, and generator iterator behavior.

### DIFF
--- a/changelog.d/9093-collection-iterator-control-methods.md
+++ b/changelog.d/9093-collection-iterator-control-methods.md
@@ -1,0 +1,6 @@
+### Fixed
+
+- Map and Set iterator objects no longer expose synthetic `return` and `throw`
+  methods. Those generator-only properties now resolve through ordinary
+  prototype lookup, while `[Symbol.iterator]()` continues to return the
+  iterator itself and user-defined overrides still work (#9086).
--- a/crates/perry-runtime/src/object/field_get_set/accessors.rs
+++ b/crates/perry-runtime/src/object/field_get_set/accessors.rs
@@ -144,11 +144,12 @@ pub(crate) unsafe fn own_data_field_by_name(
 /// the reserved floor since #9019 — shadows every synthetic method. This is
 /// also what makes user data properties on iterators readable at all: the
 /// old arm returned `undefined` for every non-`next` key without consulting
-/// own fields, so a stored value was write-only. Then the legacy bound
-/// synthetic methods (`return`/`throw`/`@@iterator`); `next` deliberately
-/// resolves through the caller's generic scans (`None`) so
-/// `iterator.next.call(other)` receives `other` and brand-checks; any other
-/// key is absent (`Some(undefined)`).
+/// own fields, so a stored value was write-only. `@@iterator` remains the only
+/// synthetic bound method: ordinary collection iterators do not have the
+/// generator-only `return`/`throw` methods (#9086). `next`, `return`, and
+/// `throw` deliberately resolve through the caller's generic scans (`None`),
+/// so the prototype chain remains authoritative; any other key is absent
+/// (`Some(undefined)`).
 pub(crate) unsafe fn map_set_iterator_property(
     obj: *const ObjectHeader,
     key: *const crate::StringHeader,
@@ -160,8 +161,6 @@ pub(crate) unsafe fn map_set_iterator_property(
     let key_len = (*key).byte_len as usize;
     let key_bytes = std::slice::from_raw_parts(key_ptr, key_len);
     let bind_name: Option<&'static [u8]> = match key_bytes {
-        b"return" => Some(b"return"),
-        b"throw" => Some(b"throw"),
         b"@@iterator" => Some(b"@@iterator"),
         _ => None,
     };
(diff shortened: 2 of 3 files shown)


proggeramlug:
Merged. Fixes #9086, which I filed earlier today.

The change is exactly the shape I suggested there — drop `return`/`throw` from the bind list, keep `@@iterator` — and routing them to `None` alongside `next` is better than the alternative of returning `Some(undefined)`: it keeps the prototype chain authoritative for all three names rather than hard-coding absence, so a user or prototype-level override still has somewhere to resolve from.

A/B against main, 20 shapes, node v26.5.1 — rows wrong 3/20 → 2/20:

|---|---|---|---|---|
| 1, 2 | `typeof it.return`, `typeof it.throw` on Set/Map iterators | `undefined` | `function` | **`undefined`** ✓ |
| 6, 7 | generators keep both, and `g.return(9)` still runs `finally` | `function`, `[9,true,true]` | ✓ | ✓ |
| 8–12 | `IteratorClose` paths: early `break`, `throw` out of the loop, `return` from inside, destructuring | — | ✓ | ✓ |
| 15 | prototype owns `next`, not `return`; `Set Iterator` tag | — | ✓ | ✓ |

I checked the `IteratorClose` paths specifically because that was the risk I raised when filing #9086 — removing `return` changes what `GetMethod(iterator, "return")` finds on an abrupt exit. Nothing regressed: `break`, `throw`, `return`-from-loop and destructuring all behave identically to node on both arms.

**One pre-existing gap this surfaces, filed as #9098 rather than held against you.** A *user-installed* `return` is readable as a function but never invoked:

```ts
const it: any = new Set([1,2,3]).values();
let closed = 0;
it.return = function () { closed++; return { done: true }; };
for (const v of it) { if (v === 2) break; }
// node: closed === 1     here: closed === 0
```

Identical on main and on this PR, so it is not yours — it is the "callers relying on the current behaviour" concern from #9086, now confirmed as its own defect. Curiously `typeof it.return` reports `"function"` while calling it gives `undefined`, so something callable is coming back that isn't the user's function; `own_data_field_by_name`'s prefix in `map_set_iterator_property` looks like the place to start. Worth noting the new integration test covers the `typeof` direction but would not catch that, since the read looks correct while the call path is wrong — an `IteratorClose` assertion would.

Validation: runtime 2819 passed (`RUST_TEST_THREADS=1`), codegen 1347, perry --bins 1066, the new `issue_9086_collection_iterator_methods` integration test passes (230 s), fmt clean, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped.

PR #9095, opened 2026-08-29
fix(runtime): keep Set/Map forEach stable across compaction

### LINKED ISSUE #9082: Set/Map forEach stops early when a callback delete triggers mid-iteration compaction
`Set.prototype.forEach` / `Map.prototype.forEach` stop early when the callback deletes enough entries to trigger the delete path's self-healing compaction mid-iteration. Every not-yet-deleted entry past that point is skipped, in violation of ECMA-262 24.1.3.5 / 24.2.3.6.

Split out of #9076, which fixed the larger half of this area (the walk yielding raw tombstone markers to the callback, and bounding by `size` instead of `used`). This residual is not introduced by that PR — main is wrong on these rows too, and more severely.

### Mechanism

`js_set_delete` self-heals a holey backing store:

```rust
let used = (*set).used;
if used >= 16 && (*set).size < used / 2 {
    compact_set_elements(set);
}
```

`compact_set_elements` shifts survivors left and resets `used` to `size`. `js_set_foreach_impl` walks a raw index `i` over `used`, re-reading the extent each step — but its *cursor* is not rebased, so after a mid-iteration compaction `i` refers to slots that have moved. `js_map_foreach_impl` / `compact_map_entries` have the identical shape.

### Repro

```ts
const s = new Set<number>();
for (let i = 0; i < 20; i++) s.add(i);
const seen: number[] = [];
s.forEach((v) => { seen.push(v); s.delete(v); });
console.log(seen.join(","), s.size);
```

```
node v26.5.1  0,1,2,...,19   0
perry         0,1,...,10     9
```

The arithmetic lines up exactly: after the 11th delete `size=9`, `used=20`, so `9 < 10 && 20 >= 16` compacts and sets `used=9`; the loop resumes at `i=11`, sees `11 >= 9`, and breaks with 9 live entries unvisited.

### Scope

Reproduces on Set and Map, for both "delete the entry being visited" and "delete an earlier entry", at any size that reaches the threshold. It does **not** reproduce below 16 elements — `used >= 16` is never satisfied — which is why #9076's 3-element unit tests are green. Confirmed at n=20 and n=40:

| collection | mutation | node | perry |
|---|---|---|---|
| Set 20 | delete self | visits 20, size 0 | visits 11, size 9 |
| Set 20 | delete earlier | visits 20, size 1 | visits 12, size 9 |
| Set 40 | delete self | visits 40, size 0 | visits 21, size 19 |
| Map 20 | delete self | visits 20, size 0 | visits 11, size 9 |

`delete-ahead` and `add-during-iteration` are correct at every size tested.

### Likely fix

Either rebase the walker's cursor when it observes that a compaction happened (e.g. an iteration-generation counter on the header, bumped by `compact_*`), or suppress self-healing compaction while an iteration is in flight and let the walk's existing hole-skipping absorb the tombstones. The second is simpler but needs an active-iteration depth counter that is exception-safe.

Whichever lands should add a delete-during-`forEach` case at **≥16 elements** — the existing coverage cannot reach the threshold.

proggeramlug:
## Summary

Defer Set/Map backing-store compaction while `forEach` has a cursor into the raw layout, so callback-side deletes cannot move unvisited entries behind that cursor.

## Changes

- Track active Set/Map `forEach` walks, including nested walks, GC relocation, and exception-savepoint restoration.
- Suppress delete/growth compaction during an active walk, preserve tombstones for callback-side `clear`, and compact after the outermost walk returns.
- Raise the Set/Map delete-during-`forEach` regressions to 20 entries and cover both deleting the current entry and an earlier entry.
- Add the required changelog fragment; no version bump.

## Related issue

Fixes #9082

## Test plan

- [x] `cargo test -p perry-runtime foreach_survives_delete_compaction_threshold -- --nocapture`
- [x] `cargo test -p perry-runtime caught_throw_restores_ -- --nocapture`
- [x] Built and ran the issue repro with this worktree compiler/runtime; Set and Map each visited `0..19` and ended at size 0.
- [ ] (if CLI / stdlib / runtime API changed) Updated `docs/src/` (not applicable: no API change)

`cargo test -p perry-runtime -- --test-threads=1` reaches the pre-existing `object::reserved_floor::tests::user_properties_read_back_through_the_get_path_at_scale` abort. The same isolated test aborts in a clean detached worktree at base `744df26d8`.

## Screenshots / output

```text
0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19 0
0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19 0
0,99 1
```

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `Map.prototype.forEach` and `Set.prototype.forEach` reliability when callbacks delete entries during iteration.
  * Ensured live entries continue to be visited even when deletions trigger internal compaction.
  * Fixed iteration state restoration when callbacks throw and the exception is caught.

### DIFF
--- a/changelog.d/9082-map-set-foreach-compaction.md
+++ b/changelog.d/9082-map-set-foreach-compaction.md
@@ -1,0 +1,2 @@
+`Map.prototype.forEach` and `Set.prototype.forEach` now continue visiting live
+entries when callback-side deletes cross the backing-store compaction threshold.
--- a/crates/perry-runtime/src/exception.rs
+++ b/crates/perry-runtime/src/exception.rs
@@ -114,6 +114,10 @@ struct ExceptionState {
     /// `call_method_depth_*`). Indexed by try-depth, in lockstep with
     /// `jump_buffers`.
     call_method_depths: Box<[u32]>,
+    /// Active Set/Map `forEach` walks. Their normal epilogues re-enable
+    /// backing-store compaction, but a caught throw skips those epilogues.
+    set_foreach_depths: Box<[usize]>,
+    map_foreach_depths: Box<[usize]>,
     /// Recorded-prototype lookup stack depth. A getter can throw while
     /// `resolve_inherited_field` is recursively walking; longjmp skips its
     /// guard drops, so restore the stack to this try-entry savepoint.
@@ -157,6 +161,8 @@ impl ExceptionState {
             shadow_savepoints: vec![ShadowSavepoint::EMPTY; MAX_TRY_DEPTH].into_boxed_slice(),
             runtime_handle_savepoints: vec![0usize; MAX_TRY_DEPTH].into_boxed_slice(),
             call_method_depths: vec![0u32; MAX_TRY_DEPTH].into_boxed_slice(),
+            set_foreach_depths: vec![0usize; MAX_TRY_DEPTH].into_boxed_slice(),
+            map_foreach_depths: vec![0usize; MAX_TRY_DEPTH].into_boxed_slice(),
             prototype_resolution_depths: vec![0usize; MAX_TRY_DEPTH].into_boxed_slice(),
             static_private_owner_depths: vec![0usize; MAX_TRY_DEPTH].into_boxed_slice(),
             private_lexical_brand_depths: vec![0usize; MAX_TRY_DEPTH].into_boxed_slice(),
@@ -219,6 +225,8 @@ fn try_push_with_kind(kind: HandlerKind) -> *mut i32 {
         // this `try` can restore it — `longjmp` skips the `CallMethodDepthGuard`
         // `Drop`s of the method frames it unwinds (#5591).
         (*s).call_method_depths[depth] = crate::object::call_method_depth_savepoint();
+        (*s).set_foreach_depths[depth] = crate::set::set_foreach_stack_savepoint();
+        (*s).map_foreach_depths[depth] = crate::map::map_foreach_stack_savepoint();
         (*s).prototype_resolution_depths[depth] =
             crate::object::prototype_chain::resolution_stack_savepoint();
         (*s).static_private_owner_depths[depth] =
@@ -346,6 +354,8 @@ pub extern "C-unwind" fn js_throw(value: f64) -> ! {
         // otherwise caught throws wrap the counter below zero and wedge every
         // later method call into the depth-guard fallback (#5591).
         crate::object::call_method_depth_restore((*s).call_method_depths[depth]);
+        crate::set::set_foreach_stack_restore((*s).set_foreach_depths[depth]);
+        crate::map::map_foreach_stack_restore((*s).map_foreach_depths[depth]);
         crate::object::prototype_chain::resolution_stack_restore(
             (*s).prototype_resolution_depths[depth],
         );
@@ -681,6 +691,8 @@ pub(crate) fn test_unwind_innermost_shadow_restore() {
         let depth = (*s).try_depth - 1;
         shadow_stack_restore((*s).shadow_savepoints[depth]);
         runtime_handle_stack_restore((*s).runtime_handle_savepoints[depth]);
+        crate::set::set_foreach_stack_restore((*s).set_foreach_depths[depth]);
+        crate::map::map_foreach_stack_restore((*s).map_foreach_depths[depth]);
         crate::object::prototype_chain::resolution_stack_restore(
             (*s).prototype_resolution_depths[depth],
         );
--- a/crates/perry-runtime/src/map.rs
+++ b/crates/perry-runtime/src/map.rs
@@ -37,7 +37,8 @@ fn map_foreach_leave(map: *const MapHeader) -> bool {
     let addr = map as usize;
     MAP_FOREACH_STACK.with(|stack| {
         let mut stack = stack.borrow_mut();
-        debug_assert_eq!(stack.pop(), Some(addr));
+        let popped = stack.pop();
+        debug_assert_eq!(popped, Some(addr));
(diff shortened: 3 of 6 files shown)


proggeramlug:
Not merging yet — **this PR's own two new tests fail deterministically**, 3/3 and 1/1 in isolation on a clean tree:

```
set::set_tombstone_tests::foreach_survives_delete_compaction_threshold
map::map_tombstone_tests::foreach_survives_delete_compaction_threshold

assertion `left == right` failed: the completed walk runs deferred compaction
  left: 20
 right: 0
```

The deferred half doesn't fire: holes are correctly pinned during the walk, but the outermost frame's compaction leaves `used` at 20.

I instrumented `set_foreach_leave` and `compact_if_holey_set` to localize it, so you don't have to redo this:

```
[dbg] leave addr=0x5c1ed480008 popped=Some(0x5c1ed480008) rest=[]
[dbg] leave -> true
[dbg] compact_if_holey used=20 size=0
```

So the bookkeeping is right — the stack pops cleanly, `set_foreach_leave` returns `true`, and `compact_if_holey_set` is entered with `used=20, size=0`, which satisfies `used != size`. Yet `(*set).used` is still 20 when the test reads it. Two candidates worth checking first: whether `compact_set_elements`'s `(*set).used = out as u32` is landing on the same header the test holds (the leave path passes `set_handle.get_raw_const_ptr()`, the test holds the `js_set_alloc` return — if `clean_set_ptr` resolves through a receiver indirection anywhere in this path, those differ), or whether something re-reads a stale `used` afterwards. I did not chase it further since it is your design.

One other observation from the same trace: the first walk's *prelude* `compact_if_holey_set` never printed, while the second walk's did. Might be nothing, might be the same pointer question.

**Everything else about this looks right**, and I want to be clear the approach is sound so this reads as "one bug to fix," not "rethink it":

- Suppressing compaction while a cursor is live, rather than rebasing the cursor, is the simpler of the two options I suggested in #9082, and you closed the gap I flagged with it — the exception-safety. Wiring `set_foreach_depths`/`map_foreach_depths` into `try_push_with_kind` and `js_throw` alongside `call_method_depths` is exactly right: a caught throw skips the epilogue, and without the savepoint a Set would stay pinned forever.
- Keeping a *stack* rather than a flag, so nested walks over the same collection only re-enable compaction at the outermost frame, is the right call — my case E (nested `forEach` over the same Set, deleting in the outer) depends on it.
- Updating the stack in `set_header_moved_for_gc` is the detail that would have been easy to miss, since the stack keys on raw addresses.
- `ensure_capacity` switching from `size < capacity` to `used < capacity` follows necessarily once holes can persist across a growth.

Behaviourally it is already correct: **32 cases byte-identical to node v26.5.1**, covering the #9082 repro at n ∈ {3,16,20,40,64} for Set and Map in self-delete and delete-earlier modes, compaction resuming after the walk, growth while holes are pinned, nested walks over the same and over different collections, a throw out of `forEach` (then a fresh walk to prove nothing stayed pinned), three sequential throwing walks, a nested inner throw with the outer continuing, `clear()` mid-walk, delete-ahead, add-during, and `for…of`/spread being unaffected. So the user-visible fix works; it is the internal reclaim assertion that is red.

Happy to re-run the full validation as soon as you push a fix — `RUST_TEST_THREADS=1 cargo test -p perry-runtime --lib` is the reproducer.

proggeramlug:
Merged. Closes #9082.

**The root cause is excellent and worth recording in full**, because it explains why my own investigation produced contradictory evidence:

```rust
-        debug_assert_eq!(stack.pop(), Some(addr));
+        let popped = stack.pop();
+        debug_assert_eq!(popped, Some(addr));
```

The `stack.pop()` — a side-effecting call — was **inside** `debug_assert_eq!`, so any build with `debug-assertions` off compiled the pop out entirely. The walk-tracking stack never drained, `set_foreach_leave` never saw an empty rest, and the deferred compaction never ran.

That is why my instrumented run in the earlier review looked *correct*: to add the `eprintln!` I had rewritten the assert as `let popped = stack.pop();` — which is the fix. **Instrumenting it fixed it.** A textbook Heisenbug, and it is why I handed the diagnosis back with two competing hypotheses (pointer identity vs. a stale re-read) rather than an answer — both were wrong.

It also rhymes with the other assertion trap I hit today: `--profile perry-dev` inherits `release`, so `debug_assert!` is compiled out there too. In this case that cut the *other* way — the bug only existed in assertion-off builds.

Verified independently on the rebased tree:
- Both previously-failing tests now pass; runtime **2824 passed**, exit 0, 0 abort markers.
- Node parity for delete-during-`forEach` at n ∈ {3,16,20,40,64} for Set and Map: every row visits all `n` and ends size 0.
- A throw out of `forEach` followed by a fresh walk: clean, so the exception savepoint path still releases the pin.
- `run_lint_gates.sh` all 60 gates passed.

**I added one commit.** `gc_runtime_root_holders` flagged the two new `SET_FOREACH_STACK` / `MAP_FOREACH_STACK` thread-locals as unclassified holders. I recorded them `not_a_gc_pointer` with the reasoning: the entries are header *addresses* used purely as identity keys — compared via `contains` and equality on `pop`, never dereferenced — and `set_header_moved_for_gc` / `map_header_moved_for_gc` rewrite every matching entry when a header moves, so they cannot go stale across evacuation. If you'd class that differently, that entry is the place to correct it.

This fix was produced by a Codex agent I briefed with the failing assertion, my instrumentation output, and the two hypotheses; it found the real cause, which neither hypothesis covered. I reviewed and re-measured rather than taking its report.

PR #9097, opened 2026-08-29
fix(runtime): coerce class refs before dynamic addition

### LINKED ISSUE #9087: correctness: class reference + number treats the INT32-boxed class ref as a number (should be string concat)

proggeramlug:
## Summary

- recognize registered INT32-tagged class references as Function objects in dynamic `+` coercion
- run the ordinary `valueOf`/`toString` sequence so default classes concatenate their function source while static overrides remain observable
- add an end-to-end regression for both operand orders and the looped `+=` reproducer

No version bump.

## Tests

- `cargo test -p perry-runtime dynamic_arith --lib`
- `cargo test -p perry --test issue_9087_class_ref_add -- --nocapture`

Closes #9087

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected addition and string coercion for declared classes used as values.
  * Class values now behave like function objects during dynamic `+` operations instead of being treated as numeric identifiers.
  * Preserved custom `valueOf` behavior and correct results for both operand orders and repeated additions.

* **Tests**
  * Added regression coverage for class coercion, accumulator operations, and custom `valueOf` results.

### DIFF
--- a/crates/perry-runtime/src/value/dynamic_arith.rs
+++ b/crates/perry-runtime/src/value/dynamic_arith.rs
@@ -168,9 +168,22 @@ unsafe fn is_nonprimitive_object_value(value: f64) -> bool {
 
 unsafe fn to_primitive_default_for_add(value: f64) -> f64 {
     let jsval = JSValue::from_bits(value.to_bits());
-    if !jsval.is_pointer() || is_symbol_value(value) {
+    let is_class_ref = crate::object::class_ref_id(value).is_some();
+    if (!jsval.is_pointer() && !is_class_ref) || is_symbol_value(value) {
         return value;
     }
+
+    // #9087: a declared class used as a value is an Object even though Perry
+    // stores it as an INT32-tagged class id. It must therefore participate in
+    // ToPrimitive instead of reaching the numeric int32 arm below. Perform the
+    // ordinary Function-object valueOf/toString sequence; the
+    // inherited toString produces the class's synthetic function source and
+    // makes the addition concatenate. This also preserves an own static
+    // valueOf/toString override.
+    if is_class_ref {
+        return crate::value::ordinary_to_primitive_for_toprimitive(value, false);
+    }
+
     let ptr = jsval.as_pointer::<u8>() as usize;
     if ptr < 0x1000 {
         return value;
--- a/crates/perry-runtime/src/value/to_string.rs
+++ b/crates/perry-runtime/src/value/to_string.rs
@@ -337,7 +337,7 @@ fn is_primitive_value(value: f64) -> bool {
     let jsval = JSValue::from_bits(value.to_bits());
     jsval.is_any_string()
         || jsval.is_number()
-        || jsval.is_int32()
+        || (jsval.is_int32() && crate::object::class_ref_id(value).is_none())
         || jsval.is_bool()
         || jsval.is_null()
         || jsval.is_undefined()
--- a/crates/perry/tests/issue_9087_class_ref_add.rs
+++ b/crates/perry/tests/issue_9087_class_ref_add.rs
@@ -1,0 +1,72 @@
+//! Regression for #9087: a declared class used as a value is represented by
+//! an INT32-tagged class id, but it is still a Function object for ECMAScript
+//! coercion. Dynamic `+` must ToPrimitive the class to its source string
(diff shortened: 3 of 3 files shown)


proggeramlug:
Merged. Closes #9087.

Running the ordinary `valueOf`/`toString` sequence rather than special-casing a string is the right shape — it means `static valueOf` and `static toString` are both honoured with correct precedence, and an inherited static `toString` works through `extends` without extra machinery. A/B against main, 27 shapes, node v26.5.1:

|---|---|---|---|---|
| 4, 5 | `static valueOf` via `+` (both operand orders) | `42` | `function WithValueOf() { [native code] }` | **`42`** ✓ |
| 6, 7 | `static toString` via `+` | `CUSTOM` | native-code text | **`CUSTOM`** ✓ |
| 8 | both defined — `valueOf` wins | `7` | native-code text | **`7`** ✓ |
| 10, 11 | `1 + C` / `C + 1` | `43` | `4` | **`43`** ✓ |
| 18 | looped numeric `n += C` | `126` | `9` | **`126`** ✓ |
| 22 | `extends` with an inherited static `toString` | `BASE` | native-code text | **`BASE`** ✓ |

Rows correct: main 5/27 → this PR 14/27, and **nothing regressed** — every row the PR touches moves toward node, and the rows still wrong are byte-identical on both arms.

Case 10/11 is the one I'd highlight: main answered `4` for `1 + WithValueOf`, which is a plausible-looking number rather than an obvious garbage string, so this was quietly producing wrong arithmetic rather than visibly wrong text.

**Four residuals, filed as #9101, none introduced here.** The one worth your attention is that the looped `+=` fix is half-landed: the numeric accumulator (case 18) is fixed, but the *string* accumulator still takes the old path —

```ts
class WithToString { static toString() { return "CUSTOM"; } }
let s = ""; for (let i = 0; i < 3; i++) s += WithToString;
// node: "CUSTOMCUSTOMCUSTOM"   here: "function WithToString() { [native code] }…"
```

— which is notable because the PR description names the looped `+=` reproducer as a target. Probably a narrow gap in the same area rather than a separate mechanism, so it may be a short follow-up. The other three are `*` and `-` not coercing at all (`WithValueOf * 2` → `6`, node `84`), static `Symbol.toPrimitive` being ignored despite taking precedence over `valueOf`/`toString` in `ToPrimitive`, and class source text not being retained (`"" + Plain` gives `function Plain() { [native code] }`, node gives `class Plain { }`) — that last one is a lowering-level gap rather than a coercion fix, and I noted in the issue that it may already be tracked elsewhere.

Validation: runtime 2819 passed (`RUST_TEST_THREADS=1`), codegen 1347, perry --bins 1066, the new `issue_9087_class_ref_add` integration test passes, fmt clean, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped.

PR #9099, opened 2026-08-29
fix(hir): preserve class-expression self identity

### LINKED ISSUE #9089: Class-expression self-binding has no per-evaluation identity: static privates shared across evaluations
A class expression's lexical self-binding (`c` in `class c { … }`) is compiled to a bare class REF carrying the class **template** id, so it has no per-evaluation identity. Every evaluation of the same class expression therefore resolves `c.#field` to the same storage, and two instances of a factory-produced class share their static private state.

Split out of #9085, which made these accesses *work at all* — on main every case below throws `TypeError: Cannot access private member from an object whose class did not declare it`, so this is a residual that the fix exposes rather than one it introduces. Confirmed by A/B: main is wrong on 11 of 12 shapes I tested, #9085 on 4.

### A. Two evaluations share static private state

```ts
const make = (seed: number) => class c {
  static #v = seed;
  static get() { return c.#v; }
  static set(x: number) { c.#v = x; }
};
const A = make(10), B = make(20);
A.set(99);
console.log(A.get(), B.get());
```

```
node v26.5.1  99 20
perry         99 99
```

The lexical read is also affected, so the self-binding stops being lexical at all:

```ts
const make = () => class c { static #s = "x"; static peek() { return c.#s; } static poke(v: string) { c.#s = v; } };
const A = make(), B = make();
B.poke("B");
console.log(A.peek(), B.peek(), A.peek.call(B));
// node:  x B x        perry:  B B B
```

`A.peek.call(B)` returning `x` in node is the point: the self-binding is lexical, not receiver-derived, so even an explicitly foreign receiver must not change which storage is read.

### B. Compound update through the self-binding writes `undefined`

```ts
const make = () => class c { static #n = 0; static bump() { c.#n++; return c.#n; } };
const A = make(), B = make();
console.log([A.bump(), A.bump(), B.bump(), A.bump(), B.bump()]);
```

```
node v26.5.1  [1, 2, 1, 3, 2]
perry         [undefined, undefined, undefined, undefined, undefined]
```

Possibly a separate defect from (A) — plain read and plain write both work (that is what makes (A) observable), so `++` specifically appears not to round-trip through the self-binding path. Worth confirming whether it is the read-modify-write lowering or the same missing per-evaluation identity before fixing.

### C. A nested class reading the outer self-binding

```ts
const make = () => class c { static #o = false; static In = class d { static peek() { return c.#o; } }; };
console.log(make().In.peek());   // node: false   perry: ReferenceError
```

Unchanged by #9085 — main throws `ReferenceError` here too.

### Direction

(A) needs the class ref to carry per-evaluation identity, not just the template id — the same distinction `private_evaluation_brand_matches` already makes for ordinary receivers via `current_private_lexical_brand`. #9085's short-circuit is correct for the single-evaluation case it targets (and restores the pre-brand static fallback's verdict) but cannot distinguish evaluations, because the information is not in the ref.

### Working shapes (regression guard for whoever fixes this)

These are correct as of #9085 and must stay so: the lru-cache construction gate, direct `new` on it throwing, foreign-class brand rejection (`A.read(B)` and `o.#p` on a plain object both `TypeError`), instance privates alongside statics, an inner class reading `Outer.#o` by name, plain named-class self-reference, and two evaluations of a class *declaration* returned from a factory.

proggeramlug:
## Summary

Named class expressions now bind their inner name to the class object created by each evaluation. Static private state stays isolated across factory calls, lexical self-access survives `call`/`apply`, and nested classes can close over an outer evaluated class.

## Changes

- Materialize a compiler-private per-evaluation self binding and initialize it before computed names, captures, and static initializers.
- Carry the evaluated receiver through static-private brand guards and preserve the lexical capture owner independently of dynamic `this`.
- Mark post-lowering self-construction arguments as capture arguments so constructor replay receives the self cell.
- Add HIR and end-to-end parity regressions, plus the required changelog fragment.

## Related issue

Fixes #9089

## Test plan

Run on `root@perrymaster.skelpo.net` from the dedicated worktree and target directory:

- [x] `cargo test --release -p perry-hir --lib` (359 passed, 1 ignored)
- [x] `cargo test --release -p perry-runtime --lib -- --test-threads=1` (2798 passed, 4 ignored)
- [x] `./run_parity_tests.sh --filter 9052` (1/1 passed)
- [x] `./run_parity_tests.sh --filter 9089` (1/1 passed)
- [x] `./run_parity_tests.sh --filter class_expr` (18/18 passed)
- [ ] Full workspace test suite (not run)
- [x] Added a user-facing regression under `test-files/`
- [ ] Documentation update (not applicable: no API change)
- [ ] Platform UI build (not applicable)

## Screenshots / output

The new fixture matches Node exactly:

```text
A2 B2 A2
[ 1, 2, 1, 3, 2 ]
false
true
```

## Checklist

- [x] My commit follows the repository prefix convention

## Summary by CodeRabbit

- **Bug Fixes**
  - Named class expressions now preserve the correct class identity for each evaluation.
  - Static private fields no longer leak state between class factory calls.
  - `call` and `apply` can no longer redirect lexical self-access to another class.
  - Nested classes can reliably access an outer class’s static private members.

- **Tests**
  - Added coverage for private static access, nested classes, repeated evaluations, and foreign-class rejection.


proggeramlug:
Merged, with one commit added for two lint-gate failures — details at the end. Fixes #9089, which I filed after auditing #9085.

This closes all three defects I reported there, which is more than I expected from one PR — the third (a nested class closing over the outer evaluation) I had flagged as possibly a separate mechanism. A/B against main, 23 shapes, node v26.5.1:

|---|---|---|---|---|
| A1 | two evaluations, `A.set(99)` | `[99, 20]` | `[99, 99]` | **`[99, 20]`** ✓ |
| A2 | `A.peek.call(B)` must stay lexical | `["x","B","x"]` | `["B","B","B"]` | **`["x","B","x"]`** ✓ |
| A3 | interleaved `bump()` across two evaluations | `[1,2,1,3,2]` | `[undefined ×5]` | **`[1,2,1,3,2]`** ✓ |
| B1, B3 | `c.#n++`, `++c.#n`, `c.#n--`, `--c.#n` | `[1,3,3,1,1]` | `[undefined ×5]` | **`[1,3,3,1,1]`** ✓ |
| C1, C2 | nested class reading the outer self-binding | `false`, `[1,2]` | `ReferenceError` | **`false`, `[1,2]`** ✓ |
| E1 | `A.self() === A`, `A.self() === B` | `[t,t,f]` | `[t,t,t]` | **`[t,t,f]`** ✓ |
| E4 | three factory instances incrementing independently | `[[2,3],[2,3],[2,3]]` | `[[undefined,undefined] ×3]` | **✓** |

Rows wrong: main 11/23 → **2/23**. All eight regression guards I listed in #9089 stay green (the lru-cache gate, direct `new` throwing, foreign-class brand rejection on both `A.read(B)` and a plain object, instance privates beside statics, `Outer.#o` by name, plain named-class self-reference, and a class *declaration* from a factory).

Carrying the evaluated receiver through the brand guard rather than re-deriving it is what makes A2 work — that `A.peek.call(B)` returns `"x"` and not `"B"` is the sharpest evidence the binding is genuinely lexical now rather than receiver-derived.

**Two residuals, filed as #9104, neither a regression** — `const C = class Named { static f() { return Named === C; } }` still throws `ReferenceError` (identical on main), and a self-binding is not captured into a nested arrow inside a static method. On that second one the *direction* changed even though both arms are wrong: main silently returned `undefined`, this now throws `TypeError`. Louder is better, but worth knowing, since code tolerating the `undefined` will start failing.

**The added commit** fixes two `lint` gates that were red:

- `cargo check --workspace --all-targets` failed — `crates/perry-codegen/tests/private_guard_declaring_class.rs` doesn't construct the new `receiver_is_brand_owner` field. Set to `false`: that guard's receiver is a plain expression, not a self-binding, so the declaring class id remains the brand owner, which is exactly what the test pins. Easy to miss locally, since integration suites under `crates/*/tests/` only run per-PR when the diff names them — but `--all-targets` in `lint` compiles them.
- `local_binding_type_audit.py` flagged the new `local_type_hint` read at `static_field_meta.rs:536`. I classified it `representation-proven` rather than `runtime-validated`, because the `Array` hint there is not a user annotation: the shared-mutable capture rewrite both sets it and allocates the one-element cell in the same pass, so the raw `POINTER_MASK_I64` decode below reaches a cell the compiler itself created. If that reasoning doesn't match your intent, that entry is the place to correct it.

I also rebased onto main; `git diff origin/main --diff-filter=D` is empty.

Validation: hir 359 passed, codegen 1347, runtime 2819 (`RUST_TEST_THREADS=1`), perry --bins 1066, the `private_guard_declaring_class` integration test passes, fmt clean, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped.

PR #9100, opened 2026-08-29
perf(codegen): binary-search dispatch for dense numeric switches — 38x, at node parity (babel's dispatch shape)

proggeramlug:
Startup-lane work for the pi/cc perf campaign: babel's parser/generator/traverse (the 19x hot region of pi's startup) dispatch through 206 switches / 1,826 cases, and perry lowered EVERY case as a runtime `js_switch_strict_equals` **call** — n/2 extern calls per dispatch, not even inline compares.

**Change** (`stmt/switch_stmt.rs`): when all case tests are integer-valued numeric literals with ≥8 unique values, canonicalize the discriminant ONCE inline (mirroring `js_switch_strict_equals`' numeric arm bit-for-bit: int32-boxed → `sitofp(trunc)`; NaN-box tag band `0x7FF9..=0x7FFF` → no-match; else raw double, real NaNs fail every ordered compare) and dispatch through a balanced `fcmp oeq/olt` tree — O(log n), zero calls. Only the DISPATCH changes: bodies, fall-through edges, and mid-block defaults are untouched; duplicates keep first-clause-wins via stable sort + dedup; the integer-valued restriction keeps tree constants clear of the subnormal range where module-slot pointer bits decode as denormals. Non-literal / string / sparse switches keep the existing tower.

**Bench** (64-case switch, 5M dispatches, rotating discriminant, min of 5): tower **652 ms** → tree **17 ms** (~38x), node 11 ms — at node's level on the dispatch itself. Mechanism proven in IR, not inferred: the bench module contains 161 `switch.bst` blocks and **zero** `js_switch_strict_equals` calls.

**Validation**: gap fixture `test_gap_9053_switch_binary_dispatch.ts` (40 cases: first/middle/last, default incl. mid-block, NaN, 3-deep fallthrough, duplicate case, negatives, `-0`, int32 limits, beyond-int32, non-integer and non-number discriminants, hot loop) **byte-identical to node v26.5.1** under both old and new binaries; existing switch regression fixtures (`test_edge_control_flow`, `test_gap_switch_continue_loop` #5989) parity-green; perry-codegen lib 1347/0; native_proof_regressions 284/0 with no IR-shape assertions needing updates; fmt applied.

## Summary by CodeRabbit

* **Performance**
  * Improved execution speed for large switches with numeric literal cases by using optimized binary-search dispatch.
  * Added more efficient handling for dense switches in performance-sensitive loops.

* **Bug Fixes**
  * Preserved correct behavior for duplicate values, `-0` and `0`, NaN, non-number values, fall-through, and default cases.
  * Retained existing behavior for switches that do not qualify for the optimization.

### DIFF
--- a/crates/perry-codegen/src/stmt/switch_stmt.rs
+++ b/crates/perry-codegen/src/stmt/switch_stmt.rs
@@ -1,6 +1,13 @@
 //! `Stmt::Switch` lowering.
 
 use super::*;
+use crate::types::{DOUBLE, I1, I32, I64};
+
+/// Minimum number of UNIQUE numeric-literal case values before the
+/// binary-search dispatch tree replaces the linear test tower. Below this the
+/// tower's straight-line compare chain is already cheap and the tree's
+/// canonicalization prologue isn't worth its extra blocks.
+const NUMERIC_TREE_MIN_CASES: usize = 8;
 
 /// `switch (disc) { case A: ...; break; case B: ...; default: ... }`
 /// lowering. Each case gets a (test, body) block pair; bodies fall
@@ -13,6 +20,17 @@ use super::*;
 /// is a NaN-boxed double whose equality semantics differ from i32
 /// switch (NaN != NaN). The if-tower lowering uses fcmp oeq for each
 /// test which yields the right semantics.
+///
+/// Dense all-numeric switches take a faster path: when EVERY case test is a
+/// compile-time integer-valued numeric literal and there are at least
+/// [`NUMERIC_TREE_MIN_CASES`] unique values, the linear tower is replaced by
+/// a balanced binary-search tree over the sorted case values — O(log n)
+/// `fcmp` branches per dispatch instead of O(n) `js_switch_strict_equals`
+/// calls (babel-style AST/token-kind dispatch switches have 50–100+ arms).
+/// Only the DISPATCH changes: bodies keep source order and fall-through, the
+/// default body is still the no-match target, and duplicate case values still
+/// resolve to the first (source-order) clause. See
+/// [`numeric_literal_dispatch_plan`] for why this is unobservable.
 pub(crate) fn lower_switch(
     ctx: &mut FnCtx<'_>,
     discriminant: &perry_hir::Expr,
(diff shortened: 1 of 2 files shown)


proggeramlug:
Merged. Confirmed both halves independently — that the tree is emitted, and that the semantics it replaces are preserved.

**The dispatch is genuinely live**, by IR diff against main on the same fixture:

| | `switch.bst` blocks | `js_switch_strict_equals` calls |
|---|---|---|
| main | 0 | 113 |
| this PR | 172 | 53 |

The residual 53 are the non-qualifying switches in my probe (string and mixed-type discriminants), which is the right outcome — `numeric_literal_dispatch_plan` declining them rather than forcing a tree. 1110 lines of IR difference, so there is no question the subject ran.

**Performance**, on a babel-shaped fixture (64 dense arms, 3M dispatches, `i % 70` so ~9% miss the range and take the default), interleaved best-of-4:

| main | this PR | node |
|---|---|---|
| 412 ms | **18 ms** | 25 ms |

22.9x here — lower than the 38x you measured, which I'd expect from a different arm count and hit/miss mix rather than a discrepancy — and the number that matters more: **it is now faster than node** on this shape, 0.72x node's time, from 16x slower. Identical results on both arms.

**Semantics.** A binary search over case values is only equivalent if nothing observable depends on the linear tower, so I probed the ways that could break — 16 shapes, byte-identical to node on *both* arms:

| | shape | node |
|---|---|---|
| 2 | fallthrough chains, including into `default` | `["012","12","2","3D","D"]` |
| 5, 6 | `-0` vs `0`, `1.5`, `"1"`, `true`, `null`, `undefined` against numeric cases | strict equality throughout |
| 13 | `default` positioned *between* cases | `["a","b","D"]` |
| 10 | labeled `break` out of a switch in a loop | `"01"` |
| 11 | discriminant expression evaluated exactly once | `1` |
| 3, 4 | sparse and negative case values | — |
| 12 | 20 dense arms probed from −2 to 21, both ends outside | — |
| 15 | nested switch with fallthrough from the inner into the outer arm | `["00","01","0x","1","x"]` |
| 16 | `9007199254740991` and `9007199254740992` | `["max","z","o","o"]` |

Case 5 is the one I most wanted to see: `-0` must match `case 0` (switch uses strict equality, where `-0 === 0`), and `"1"` must *not* match `case 1`. Both hold, so the `fcmp oeq` canonicalization is right on the two values where a naive bit comparison would diverge. Case 13 is the other one — `default` in the middle still participates in fallthrough order rather than being hoisted, which a tree could easily have broken.

Validation: codegen 1347 passed, runtime 2819 (`RUST_TEST_THREADS=1`), perry --bins 1066, fmt clean, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped.

PR #9102, opened 2026-08-29
fix(codegen): root pointer locals of constructor bodies inlined into other frames

### LINKED ISSUE #9081: compilePackages: moving GC corrupts Three WebGLRenderTarget construction
## Summary

Perry's moving/scavenging GC can corrupt live values while constructing stock Three.js objects from a package compiled with `compilePackages`.

A one-line `new WebGLRenderTarget()` using unmodified `three@0.180.0` succeeds in Node and with Perry's default heap, but deterministically fails when the scavenging nursery is reduced to 1 MiB. The first constructor is sufficient; no Bloom, renderer shim, DOM, or WebGL implementation is involved.

This is also the next runtime blocker for a native Three.js + Bloom game after #9050, #9051, #9052, and #9053. In the full game, normal heap growth eventually produces the same apparent class of corruption during `RenderTarget` construction.

## Environment

- Perry tested revision: `8b40634abdf525cfadb03bb0aa6f3c874bec140b` (main; the only newer commit at filing time, `d253e13a21`, changes simctl CI cache invalidation)
- Perry version: `0.5.1519`
- Backend: in-process LLVM 22.1.4
- Three.js: `0.180.0`
- Host: macOS 26.5 arm64
- Node: `v26.5.1`

## Minimal reproduction

`package.json`:

```json
{
  "name": "perry-three-render-target-gc-repro",
  "private": true,
  "type": "module",
  "dependencies": {
    "three": "0.180.0"
  },
  "perry": {
    "compilePackages": [
      "three"
    ],
    "allow": {
      "compilePackages": [
        "three"
      ]
    }
  }
}
```

`actual-three.js`:

```js
import { WebGLRenderTarget } from 'three';

const target = new WebGLRenderTarget();
console.log(target.depthTexture === null ? 'ok' : 'bad');
```

From a Perry checkout at the revision above:

```sh
```

From the reproduction directory:

```sh
npm install

PERRY_RUNTIME_DIR=/path/to/perry/target/release \
  /path/to/perry/target/release/perry compile actual-three.js \
  -o actual-three.bin \
  --no-auto-optimize \
  --no-color

node actual-three.js
./actual-three.bin

PERRY_GC_SCAVENGE=1 \
PERRY_GC_SCAVENGE_NURSERY_MB=1 \
PERRY_GC_INCREMENTAL=0 \
  ./actual-three.bin
```

## Expected result

All three executions print:

```text
ok
```

and exit 0.

## Actual result

Node and the default Perry heap print `ok` and exit 0. The 1 MiB scavenging-nursery run deterministically fails on the first constructor and exits 1:

```text
TypeError: Cannot read properties of undefined (reading 'slice')
    at <anonymous>
```

The `slice` operation is Three's unmodified `Texture.copy` implementation:

```js
this.mipmaps = source.mipmaps.slice( 0 );
```

## Full integration symptom

With an otherwise working native Three.js renderer backed by Bloom, the game compiles all 161 modules and initializes Bloom and its atmosphere, but later fails under the default GC configuration with:

```text
TypeError: Cannot set properties of null or undefined (setting 'renderTarget')
```

That assignment is in Three's unmodified `RenderTarget.depthTexture` setter:

```js
set depthTexture( value ) {
  const current = this._depthTexture;

  if ( current !== null ) current.renderTarget = null;
  if ( value !== null ) value.renderTarget = this;

  this._depthTexture = value;
}
```

Local diagnostic logging around the stock `RenderTarget` constructor showed:

1. `depthTexture` is omitted from the caller-provided options.
2. Immediately after `Object.assign(options, defaults)`, `options.depthTexture` exists and is `null`, as expected.
3. After the intervening texture allocations/construction, the later `this.depthTexture = options.depthTexture` passes `undefined` to the setter.
4. Logging at the Perry Three renderer adapter's `setRenderTarget` entry showed a valid receiver for every call; the failing write occurs before that adapter is entered.

This instrumentation was diagnostic only; the minimal reproduction above uses the published, unmodified Three package.

## Likely cause

This looks like a missing or stale GC root across allocation-capable constructor/method calls in generated code. A live object/value used by the outer `RenderTarget` constructor appears not to be protected and reloaded across an evacuating minor collection, so a later property read observes invalid/stale state.

That is an inference, but the behavior is strongly GC-sensitive:

- Node succeeds.
- Perry succeeds when this tiny reproduction does not fill its normal nursery.
- A 1 MiB scavenging nursery fails deterministically on the first constructor.
- The larger application reaches a related corruption under default GC after natural heap growth.
- Removing Bloom and the renderer adapter does not remove the failure.

## Suggested fix and regression coverage

- Ensure live NaN-boxed/object locals in constructors remain rooted across every allocation safepoint and are reloaded after a moving collection.
- Audit nested constructor/method calls in compiled package code for shadow-stack/temp-root dominance, especially values retained by an outer constructor while inner constructors allocate.
- Add a `compilePackages` regression using `three@0.180.0` and one `new WebGLRenderTarget()` under a forced small scavenging nursery.
- Make the test prove that a collection/evacuation happened, so it cannot pass merely because allocation volume or nursery sizing changes.

proggeramlug:
Fixes #9081

## Root cause

Codegen splices constructor bodies inline into other functions at five sites: the `super(...)` parent-body inline in a derived constructor (`expr/this_super_call.rs`), the `new`-site own-ctor and inherited-ctor inlines (`lower_call/new.rs`), and `let_stmt`'s scalar-ctor variants. The enclosing function's shadow-slot map is computed by `collect_pointer_typed_locals` over that function's **own** params and body, so a spliced body's pointer locals — and its ctor params, bound by `bind_inline_constructor_params` — landed in plain entry allocas that are neither shadow slots nor temp roots. The collector never rewrites them, and a moving minor between the local's store and a later use leaves it holding the pre-move address (the CLAUDE.md "root-store dominance" class, `#7207` shape, introduced by body splicing).

That is exactly how three.js died under `compilePackages` with a small nursery: `RenderTarget`'s ctor body, spliced by the super() inline into the standalone `WebGLRenderTarget_constructor` (which the entry module calls as a symbol), holds `const texture = new Texture(image)` across `this.textures = []` and the attachment loop. The from-space quarantine (`PERRY_GC_PROTECT_FROMSPACE=1`, depth 800) pinpointed the stale header deref inside that constructor; the emitted IR shows the spliced `texture` alloca has **no** `js_shadow_slot_bind`, while the standalone `RenderTarget_constructor` — whose own compile pass saw the body — roots the same local at slot 4. Downstream, `Texture.copy()` reads `source.mipmaps` off the retired address as `undefined` → `TypeError: Cannot read properties of undefined (reading 'slice')`.

## Fix

`expr/shadow_slot.rs` gains `root_inlined_ctor_pointer_locals`, called at all five splice sites before lowering the spliced body. It runs the same pointer-locals collector over the spliced params+body and extends the frame via `reserve_shadow_slot`, which grows the already-emitted frame in place on **both** root backends (native stack maps and shadow frames) — matching the bug reproducing under both `PERRY_RS4GC` arms. `Let`/assignment sites then mirror stores through the ordinary bind path; an id already bound in `ctx.locals` (a spliced ctor param) is bound immediately, and re-bound on a repeated inline of the same constructor so the slot tracks the newest alloca. Local ids are module-unique (one counter per lowering context), so extending the map never aliases an enclosing local, and a nested `super()` chain recurses through the same site naturally.

`is_global_this_value` moved from `let_stmt.rs` to `let_stmt_facts.rs` (pure relocation) to stay under the 2000-line cap.

## Regression test

`test-files/test_gap_gc_inlined_ctor_body_locals_rooting.ts`, registered in the gc-repsel corpus. Its `parity-env` runs the seeded every-poll evacuating schedule with the from-space quarantine armed, and its stamp discriminator makes a stale spliced local observable by value after a single move (write through the collector-rewritten instance field, read back through the spliced local). Four arms: the dynamic-construct route that replays the standalone derived ctor (the exact three.js configuration), direct `new` of the derived/default/base classes.

## Validation (on the issue's reproduction host)

- New gap test on the **unfixed** compiler: SIGSEGV at minor #0 through the harness (`CRASHED`), quarantine naming the stale spliced-local deref. Fixed: byte-exact with the oracle under default env and parity-env, on both `PERRY_RS4GC` backends.
- Original three.js repro (`new WebGLRenderTarget()`, three 0.180.0 via `compilePackages`): passes on default heap, forced 1 MiB nursery, forced nursery + quarantine (depth 800), and the `PERRY_RS4GC=0` build. `PERRY_GC_DIAG` confirms 2 copying collections, so the stress is live, not vacuous.
- `gc_repsel_matrix.sh --arms pr` on the new test: 7/7 cells byte-exact vs pinned node 26.5.1; the moving arms (`evac_minor`, `force_verify`) are live (3,469 objects moved).
- Full `test_gap_` sweep A/B on the same host, same oracle: failure lists byte-identical between the unfixed and fixed compilers except the new test (crash → pass). The pre-existing deltas on that host are its ext-feature link gaps (http2/net) and known snapshot entries.
- `cargo test -p perry-codegen`: 1347 passed. `scripts/run_lint_gates.sh`: green except the known worktree-only `cargo fmt --all` manifest quirk (changed files are rustfmt-clean); the test-registration gate passes with the new corpus entry.

https://claude.ai/code/session_0131AxDes1mwPJJ343CxS4oX

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue that could cause object references in inlined constructors to become invalid during garbage collection.
  * Improved reliability for object creation across direct, inherited, and dynamic constructor paths.

* **Tests**
  * Added regression coverage for garbage collection during constructor execution and object allocation.

### DIFF
--- a/changelog.d/9102-inlined-ctor-body-locals-rooting.md
+++ b/changelog.d/9102-inlined-ctor-body-locals-rooting.md
@@ -1,0 +1,44 @@
+### Fixed
+
+- **Moving GC no longer corrupts pointer locals of constructor bodies inlined
+  into other frames** (#9081). Codegen splices constructor bodies inline at
+  five sites — the `super(...)` parent-body inline in a derived constructor
+  (`expr/this_super_call.rs`), the `new`-site own/inherited-ctor inlines
+  (`lower_call/new.rs`), and `let_stmt`'s scalar-ctor variants — but the
+  enclosing function's shadow-slot map was computed by
+  `collect_pointer_typed_locals` over that function's OWN params and body, so
+  a spliced body's pointer locals (and its ctor params, bound by
+  `bind_inline_constructor_params`) landed in plain entry allocas the
+  collector never rewrites. A moving minor between a local's store and a
+  later use left the local holding the pre-move address.
+
+  The reported shape: three.js 0.180.0 under `compilePackages` with a 1 MiB
+  nursery. `RenderTarget`'s ctor body, spliced by the super() inline into the
+  standalone `WebGLRenderTarget_constructor`, holds `const texture = new
+  Texture(image)` across `this.textures = []` and the attachment loop; the
+  retired-from-space `texture` then made `Texture.copy()` read
+  `source.mipmaps` as undefined — `TypeError: Cannot read properties of
+  undefined (reading 'slice')`. The from-space quarantine
+  (`PERRY_GC_PROTECT_FROMSPACE=1`, depth 800) named the stale header deref
+  inside the constructor, and the emitted IR showed the spliced `texture`
+  alloca carried no root bind while the standalone `RenderTarget_constructor`
+  rooted the same local at slot 4.
+
+  Fix: `expr/shadow_slot.rs::root_inlined_ctor_pointer_locals`, called at all
+  five splice sites before lowering the spliced body. It runs the same
+  pointer-locals collector over the spliced params+body and extends the frame
+  through `reserve_shadow_slot`, which grows the already-emitted frame in
+  place on both root backends (native stack maps and shadow frames — the bug
+  reproduced under both `PERRY_RS4GC` arms). Already-bound ctor params are
+  bound immediately and re-bound on a repeated inline of the same
+  constructor; module-unique local ids make the map extension alias-free, and
+  nested `super()` chains recurse through the same site.
+
+  Regression test `test_gap_gc_inlined_ctor_body_locals_rooting.ts`
+  (registered in the gc-repsel corpus) runs the seeded every-poll evacuating
+  schedule with the quarantine armed: the unfixed compiler SIGSEGVs at minor
+  #0; fixed, it is byte-exact with the oracle on both backends. The original
+  three.js reproduction passes on default heap, forced 1 MiB nursery,
+  quarantine, and the `PERRY_RS4GC=0` build, with `PERRY_GC_DIAG` confirming
+  live copying collections. Also relocates `is_global_this_value` from
+  `let_stmt.rs` to `let_stmt_facts.rs` (pure move, 2000-line cap).
--- a/crates/perry-codegen/src/expr/mod.rs
+++ b/crates/perry-codegen/src/expr/mod.rs
@@ -210,7 +210,7 @@ pub(crate) use shadow_slot::{
     current_closure_ptr_value, emit_persistent_shadow_root_barrier,
     emit_shadow_slot_bind_for_local, emit_shadow_slot_clear, emit_shadow_slot_update_for_expr,
     enable_persistent_shadow_slot_for_array_alias, expr_is_known_non_pointer_shadow_value,
-    try_current_closure_ptr_value,
+    root_inlined_ctor_pointer_locals, try_current_closure_ptr_value,
 };
 
 /// One in-flight inline-constructor return target. See
--- a/crates/perry-codegen/src/expr/shadow_slot.rs
+++ b/crates/perry-codegen/src/expr/shadow_slot.rs
@@ -345,6 +345,55 @@ pub(crate) fn emit_persistent_shadow_root_barrier(ctx: &mut FnCtx<'_>, value_bit
     ctx.current_block = done_idx;
 }
 
+/// #9081: root the pointer locals of a constructor body spliced inline into
+/// the CURRENT function — the `super(...)` parent-body inline, the `new`-site
+/// own/inherited-ctor inline, and `let_stmt`'s scalar-ctor variants.
+///
+/// The enclosing function's `shadow_slot_map` was computed by
+/// `collect_pointer_typed_locals` over that function's OWN params and body;
+/// an inlined constructor body's locals are invisible to it. Its `Let`s then
+/// land in plain entry allocas that are neither shadow slots nor temp roots,
+/// so a moving minor during the body leaves them holding the pre-move
+/// address (three.js: `RenderTarget`'s `const texture = new Texture(...)`
+/// inlined into `WebGLRenderTarget_constructor`, stale by `texture.clone()`).
+///
+/// Runs the same collector over the spliced params+body and extends the map
+/// through `reserve_shadow_slot`, which grows the already-emitted frame in
+/// place on both root backends (stack-map count and shadow-frame push).
+/// `Let`/assignment sites then mirror stores through the ordinary bind path.
+/// An id already in `ctx.locals` (a ctor param bound by the caller before
+/// this runs) is bound here — unconditionally, because a second inline of
+/// the same constructor in this function binds fresh param allocas that must
+/// re-point the existing slot. Local ids are module-unique, so extending the
+/// map never aliases an enclosing local.
+pub(crate) fn root_inlined_ctor_pointer_locals(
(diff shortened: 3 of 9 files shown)


proggeramlug:
Merged. This is the highest-consequence class of bug in the codebase and the diagnosis is exactly right, so I validated it with both instruments rather than just the test.

The root cause statement is the valuable part:

> The enclosing function's `shadow_slot_map` was computed by `collect_pointer_typed_locals` over that function's OWN params and body; an inlined constructor body's locals are invisible to it.

That is the third shape in CLAUDE.md's root-store-dominance list — "the value lives in a plain `alloca_entry` that is neither a shadow slot nor a temp root" — and the reason it survived is that the analysis and the splice are correct *individually*; only their composition is wrong. Extending the map through `reserve_shadow_slot`, which grows the already-emitted frame in place on both root backends, is the right repair, and sorting the ids before assigning slots so indices don't depend on `HashMap` iteration order is the kind of detail that would otherwise produce a nondeterministic build.

**The behavioural A/B, under the knobs the test itself declares** (`PERRY_GC_SCHEDULE_SEED=9081 RATE=1 ALLOC_KB=0 FORCE_EVACUATE=1 VERIFY_EVACUATION=1 PROTECT_FROMSPACE=1`):

| | plain | under stress |
|---|---|---|
| main | matches node — bug fully latent | **`rc=138`**, from-space fault |
| this PR | matches node | **matches node**, 1541 copying minors, 27,708 objects moved |

Main's failure is reported by the quarantine in as many words:

```
```

Which is the point worth underlining: **without GC pressure both arms print identical correct output**. A reviewer running this test plainly would have concluded there was nothing to fix. The `parity-env` header carrying the knobs is what makes the test able to fail at all.

**The static instrument agrees independently.** `gc_root_dominance_check.py --unrooted-allocas` over the emitted IR:

| | unrooted-alloca violations |
|---|---|
| main | **7** — 6 in `perry_fn_gap_ts__runDefaultCtor`, 1 in `gap_ts__DerivedOwnCtor_constructor` |
| this PR | **0**, with 46 gc-capable allocas seen, no `IMMOVABLE_SOURCES` suppressions |

The 46 is what makes the zero meaningful — the checker had subjects rather than being blind to the file. (The root-dominance check proper reports 0 root stores on this fixture and says so rather than returning a green; that is the tool being honest about not applying here, not evidence.)

The three.js provenance in the header — `RenderTarget`'s `const texture = new Texture(image)` going stale by `texture.clone()`, surfacing as "Cannot read properties of undefined (reading 'slice')" in `Texture.copy()` — is worth keeping. It is the canonical presentation of this class: a `TypeError` about an unrelated property, several frames and one collection away from the actual defect.

Covering all four splice sites (the `super(...)` parent-body inline, both `new`-site inlines, and `let_stmt`'s scalar-ctor variants) rather than only the one that bit three.js is the right scope, and the dynamic arm routing through `js_new_function_construct` to hit the standalone derived constructor is a nice touch — that is the configuration that actually reproduced in the wild.

Validation: codegen 1347 passed, runtime 2819 (`RUST_TEST_THREADS=1`), perry --bins 1066, fmt clean, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped.

PR #9103, opened 2026-08-29
perf(runtime,codegen): direct install for __export's get-only descriptors (groundwork; honest numbers inside)

proggeramlug:
Startup-lane groundwork from the pi/cc campaign, with an honest negative-ish finding attached.

esbuild's `__export` runs ~1,245 always-executed `Object.defineProperty(target, name, {get, enumerable:true})` calls at pi startup. This adds `js_object_define_get_accessor` (fast arm strictly limited to plain extensible objects / fresh non-numeric string keys / callable-or-undefined getters / unpolluted Object.prototype — everything else, incl. proxies, redefines, symbol keys, and every ToPropertyDescriptor TypeError, delegates to the generic path via a materialized descriptor) plus HIR-level recognition of the two-field descriptor literal in either field order, fail-closed.

**The honest measurement** (quiet mini, min-of-runs, path selection verified in kept IR before trusting any number): the descriptor alloc+decode this removes is ~1-2% of perry's install cost (−2.0% micro, −1.2% realistic 28-key shape). Node is still ~40x faster on the install itself — the gap lives in the accessor INSTALL machinery: two `HashMap<(usize,String)>` side-table inserts, epoch bumps, and inline-guard invalidation per install. That follow-up (side-table install path) is queued next in the lane; this PR contributes the entrypoint the follow-up needs, the admission analysis, and the parity fixture.

**Validation**: perry-codegen 1349/0 (2 new IR-emission tests incl. both field orders + 4 near-miss shapes staying generic); perry-runtime 2821/0 (one pre-existing SIGABRT excluded, verified identical on pristine main f3f405271c via stash); gap fixture `test_gap_9053_export_getter_descriptor.ts` byte-identical to node incl. redefine TypeErrors and the enumerable:false near-miss; census/file-size/fmt green.

## Summary by CodeRabbit

- **Performance**
  - Improved performance for common getter-based `Object.defineProperty` operations used by module re-exports and interoperability scenarios.

- **Bug Fixes**
  - Preserved standard behavior for enumeration, getter access, redefinitions, descriptor reflection, and validation errors.
  - Maintained existing behavior for numeric keys, symbols, existing properties, and unsupported descriptor forms.

- **Tests**
  - Added coverage for getter installation, live values, enumeration order, descriptor inspection, redefinitions, and near-matching descriptors.

### DIFF
--- a/crates/perry-codegen/src/expr/misc_methods.rs
+++ b/crates/perry-codegen/src/expr/misc_methods.rs
@@ -211,6 +211,25 @@ pub(crate) fn lower(ctx: &mut FnCtx<'_>, expr: &Expr) -> Result<String> {
         // (ToString, accessor descriptors, closure creation), leaving `obj`
         // and `key` stale in bare SSA registers across the collection point.
         Expr::ObjectDefineProperty(obj, key, value) => {
+            // Fast path: the esbuild `__export` / CJS-interop descriptor
+            // literal `{ get: <expr>, enumerable: true }` (either property
+            // order) lowers to a direct accessor-install call, skipping the
+            // descriptor allocation and its by-name field decode. `true` is
+            // effect-free, so evaluating only obj → key → getter preserves
+            // the literal's evaluation order; the runtime entrypoint owns the
+            // exact `defineProperty` semantics (see
+            // perry-runtime's object_ops/define_get_accessor.rs contract).
+            if let Some(getter) = get_only_descriptor_getter(ctx.classes, value) {
+                return rooting::with_operands_rooted(ctx, &[obj, key, getter], |ctx, vals| {
+                    let blk = ctx.block();
+                    blk.call(
+                        DOUBLE,
+                        "js_object_define_get_accessor",
+                        &[(DOUBLE, &vals[0]), (DOUBLE, &vals[1]), (DOUBLE, &vals[2])],
+                    );
+                    Ok(vals[0].clone())
+                });
+            }
             rooting::with_operands_rooted(ctx, &[obj, key, value], |ctx, vals| {
                 let blk = ctx.block();
                 blk.call(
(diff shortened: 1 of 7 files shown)


proggeramlug:
Not merging yet — **two required `lint` gates are red, both naming this PR's new files.** Neither is a formality; one of them is the rooting ratchet. Everything else about the PR is in good shape, detailed below, so this should be a short round trip.

### 1. `raw_handle_debt.py` — 5 bare reads in a module with no ceiling

```
crates/perry-runtime/src/object/object_ops/define_get_accessor.rs: 5 bare read(s)
in a module with no ceiling. New code must use RuntimeHandle::across_{mut,const,nanbox}
or with_{mut,const}_ptr; see #7341.
```

The bare run locks unlisted modules at zero, so a new file starts with no allowance. The sites:

| line | shape |
|---|---|
| 113, 123, 146, 166 | `let`-bound — `across_{const,nanbox}` converts these directly |
| 140, 239, 241, 246, 251, 255 | **argument position** — `across_*` cannot express these; they need `with_{const,mut}_ptr` scoping around the call |

I deliberately did not convert these myself. The code already looks semantically right — line 146 re-reads `key_str` from the handle after the intervening call, which is the correct post-call reload — so this is about expressing it through the checked helpers rather than a latent bug. But six of the sites are argument-position in a rooting-sensitive path in a 691-line new module, and reshaping those into `with_*_ptr` closures is exactly where I'd risk introducing the defect the ratchet exists to prevent.

### 2. `string_payload_access_inventory.py` — 2 new open-coded payload offsets

```
define_get_accessor.rs:128   [inline-offset] open-coded StringHeader payload offset
descriptor_helpers.rs:41     [inline-offset] open-coded StringHeader payload offset
```

Both are the same hand-rolled shape:

```rust
let name_ptr = (key_str as *const u8).add(std::mem::size_of::<crate::StringHeader>());
let name_len = (*key_str).byte_len as usize;
```

This is the #8445 ratchet — new sites are supposed to go through the accessor rather than re-deriving the offset.

### Everything else checks out

**Correctness: 22 shapes, byte-identical to node v26.5.1.** I concentrated on the fast arm's exclusions, since "strictly limited to plain extensible objects / fresh non-numeric string keys / callable-or-undefined getters / unpolluted Object.prototype" is a lot of conditions to get right:

| | must delegate | result |
|---|---|---|
| 6, 7 | redefine, and redefine over `configurable:false` | `TypeError` ✓ |
| 8, 9 | symbol key; numeric string key `"0"` | ✓ |
| 10 | `preventExtensions` target | `TypeError` ✓ |
| 11, 13 | non-callable getter; `get` **and** `value` together | `TypeError` ✓ |
| 15 | `Proxy` with a `defineProperty` trap — trap must fire | `[1,true]` ✓ |
| 16 | polluted `Object.prototype` | ✓ |
| 4, 5 | resulting descriptor exactly right (`configurable:false`, absent `enumerable` → `false`) | ✓ |
| 17–19 | getter `this`-binding, laziness (0 calls until read), throwing getter | ✓ |
| 20, 21 | `Object.keys` / `JSON` / spread / `for…in` / `in` / `hasOwnProperty` | ✓ |

**The subject is live**, by IR diff on the esbuild `__export` shape:

| | `js_object_define_get_accessor` | generic `js_object_define_property` |
|---|---|---|
| main | 0 | 2 |
| this PR | **2** | 1 |

**Suites:** runtime 2822 passed under both profiles (dev-profile run exit 0, 0 abort markers, 54 s), codegen 1349, perry --bins 1066, fmt clean. Only the two gates above are red.

Also worth saying: the honest framing in the PR body — a measured 1–2% with node still ~40x ahead on the install itself, and the gap located in the side-table inserts and epoch bumps rather than the descriptor decode — is the right way to land groundwork. It makes #9113's 2.33x claim on the install path legible as the actual follow-through rather than a competing measurement.

I'll re-run the full validation as soon as the two gates are addressed; #9113 stacks on this so it's queued behind it.

proggeramlug:
Merged. The two gates I flagged are fixed — by a Codex agent I briefed with the exact failures, then reviewed and pushed back on once. Recording that here because the first attempt is instructive.

**What landed:**
- Every raw **pointer** read in `define_get_accessor.rs` now goes through the checked helpers: `with_const_ptr` scoping the payload copy and the non-allocating presence probe, and a genuine `across_const` whose closure body is `obj_value_has_own_key` — so the returned pointer really is post-call. `get_raw_*_ptr` count in that file is now **0**.
- The value reads (`get_nanbox_f64`, `get_heap_word_u64`) stay bare, correctly — an f64 read out of a handle cannot go stale the way a raw pointer can, which is why the ratchet doesn't count them.
- `string_data(...)` replaces both open-coded `StringHeader` payload offsets, and the perry-runtime `inline-offset` baseline tightens 364 → 363.

**The pushback, for the record.** The first attempt satisfied `raw_handle_debt.py` like this:

```rust
let (_, key_value) = key_handle.across_nanbox(|| ());
```

`across_nanbox` is `let r = f(); (r, self.get_nanbox_f64())`, so with an empty closure that is *byte-for-byte the same bare read*, wrapped so the regex stops counting it. Four sites were converted that way. The gate went green while the code got no safer and less readable — and the helper's own doc says exactly what it is for:

> It is not a proof… What it does is make the correct shape shorter than the incorrect one and give the ratchet in `scripts/raw_handle_debt.py` something to count down.

Rejected and re-briefed; the second pass moved the real allocating calls into the closures for the two pointer sites and unwrapped the two value sites. Worth flagging as a hazard for anyone else clearing this ratchet: `across_*(|| ())` is a green build and a no-op, and nothing in the gate can tell the difference.

Verified independently after the fix, on the rebased tree (not on the pre-rebase tree I first checked): `run_lint_gates.sh` **all 60 gates passed**, runtime **2822 passed** exit 0 with 0 abort markers, perry --bins 1066 passed, `git diff origin/main --diff-filter=D` empty.

The correctness and IR evidence from my earlier review stands unchanged — 22 shapes byte-identical to node including every delegation case, and the fast arm confirmed live (main 0 `js_object_define_get_accessor` calls → 2 here).

#9113 stacks on this and is now unblocked; I'll pick it up.

PR #9105, opened 2026-08-29
perf(codegen): devirtualize single-binding closure calls — captured/global/local arrow calls to 3.6 ns (was 4.6-9.3)

proggeramlug:
## What

Three pieces that together devirtualize loop-called closures held in
single-binding immutables:

1. **Inline identity probe** on the known-func_id guarded direct path (normal
   builds only — feedback-emission builds keep the out-of-line guard, the same
   dispensation `guarded_array.rs` documents): a POINTER-tag/band check, then
   two compare-only loads — `type_tag == CLOSURE_MAGIC` and
   `func_ptr == @closure_fn` — decide the monomorphic case; any miss takes the
   existing guard, which keeps observation for real polymorphism. A forwarded
   (moved) closure fails the func-ptr compare (word 0 holds the forwarding
   target) and heals through the guard as before.
2. **Single-binding seeding**: #7170 R1's `single_binding_closure_locals`
   (exactly one `Let`, closure-literal init, never written at any depth in any
   body, never rebound) is threaded through `artifacts.rs` and seeded into
   captured/module-global callee bindings of closure bodies, giving them the
   same known-func_id treatment a body-local `Let` gets — with two carve-outs:
   * **trusted-box closures are excluded** — their entry-resolved path returns
     the trusted clone with entry-cached box-capture pointers, which measures
     better for capturing bodies (2.5-2.8 vs 5.1 ns);
   * **guard-free full bypass only for captured bindings** — a capture of a
     single-binding closure is boxed by construction when readable before its
     `Let`, so the TDZ read throws before dispatch; a module global can be
     called during init while its cell holds the sentinel, so globals keep the
     probe, whose magic check fails on the sentinel into the dispatcher's
     correct error path.
3. `PERRY_CALL_DEVIRT=0` empties the seeding for A/B; the probe's magic
   constant is **derived in code** — the first version hand-converted it and
   the transposed literal (0x434F4B53 vs 0x434C4F53) made the probe miss on
   every call, which mismeasured this entire direction as a regression three
   times before gdb-level operand inspection caught it.

## Numbers

Single-shape 50M-call probes, quiet Linux, same-build kill-switch A/B;
node 26.5.1 on the same host:

| shape | main | this PR | node |
|---|---|---|---|
| captured arrow called in loop | 4.6 | **3.6** | 0.5 |
| module-global arrow | 4.6 | **3.6** | 0.6 |
| local-const arrow | **9.3** | **3.6** | 1.6 |
| capturing arrow (trusted path kept) | 2.5 | 2.8 | 0.9 |
| static fn call (untouched control) | 0.6 | 0.6 | 0.5 |

Full 12-op sweep: gate on/off identical outside the call rows. Known-arm
instruction count on the probe fix alone: 264 → 60 instr/call.

## Semantics

Differential vs node identical on the call corpus (reassigned globals observe
the new value, ordinary functions keep `this === undefined`, bound/rest/arity
shapes keep the dispatcher, throwing callees, capture mutation between calls,
recursion through the seeded binding, same-name shadowed bindings). One
pre-existing, gate-independent difference surfaced by the TDZ test: calling a
module-global const arrow before its `Let` throws `TypeError` (perry has no
TDZ for globals, #4926) where node throws `ReferenceError` — unchanged by this
PR; the probe guarantees the safe fallback rather than a wild call. Kill-switch
build output-identical on all corpora.

## Testing

* `RUSTFLAGS=-D warnings cargo check` (host excludes), `perry-codegen`,
  `perry-runtime --lib` — green.
* Pinned-worktree integration battery: `issue_8655_array_subclass_indexing`,
  `issue_8897_field_push_writeback`.
* wolf-ecs mini screens vs fresh main (post-#9084/#9091): see PR comment.

Cross-session note: the pi/cc startup campaign will re-profile its call bands
after this merges; the esbuild `__commonJS`/`__esm` callback-parameter band is
only PARTIALLY covered (params are per-call values — the follow-up is lifting
#9071's Function-type-hint gate for entry resolution, tracked in my lane).

## Base-state note (important)

`issue_8690_loop_versioned_arraylike` (1 case) and
`issue_8773_closure_capture_packed_loops` (1 case) fail **identically on
current main** (`f3f405271c`): the wolf-ecs-shaped nested subclass loop has
LOST its versioned fast clones entirely — `scan()`'s IR contains zero
`js_packed_arraylike_loop_guard` sites where the shape tests expect three. The
program still runs correctly, at generic-loop speed. This is a main
regression in the packed-loop lane (bisect between #9084^ / #9084 / #9091
running; report follows to the loop-lane channel), independent of this PR —
verified by running both suites on detached main in the same pinned worktree.
wolf-ecs mini screens are deferred until that regression is resolved, since a
vs-main comparison would flatter this PR; the single-shape call probes and the
gate-off ops parity above are the evidence base.

## Summary by CodeRabbit

  * Improved closure call optimization by recognizing closures that are statically known and safely routing calls directly.
  * Reduced unnecessary runtime checks for eligible closure bindings.
  * Added an environment-controlled option to enable or disable closure call devirtualization.
* **Build Reliability**
  * Build caching now correctly distinguishes builds with different closure optimization settings.


proggeramlug:
Merged, with two commits added — details at the end.

**The devirtualization is live and the kill switch works.** Compiling the same fixture with and without `PERRY_CALL_DEVIRT=0` on one binary gives a 1344-line IR difference, which is the check that matters: the knob genuinely selects between the direct and entry-resolved paths rather than being inert.

Performance, interleaved best-of-3, 20M iterations:

| shape | main | this PR | |
|---|---|---|---|
| local arrow | 23 ms | 21 ms | 1.10x |
| module-global arrow | 520 ms | 463 ms | 1.12x |
| **captured arrow** | 52 ms | **20 ms** | **2.60x** |

The captured case is where the single-binding seeding pays off, and it's the one that matters for real code.

**Correctness: 21 shapes, identical to main on every one**, so no behavioural delta — which is what a devirtualization should show. I deliberately probed the ways an identity assumption breaks:

| | shape | node |
|---|---|---|
| 5, 6 | a `let` binding **reassigned mid-loop** | `126`, `"A0B1B2B3"` |
| 7, 8 | polymorphic call site — 3 closures through one variable, and a closure chosen per-iteration | `60`, `44` |
| 3 | two closures from one factory, distinct captured state | `"10,100;11,101;12,102;"` |
| 11, 12 | self-recursion and mutual recursion | `3628800`, `[true,false]` |
| 13 | arity mismatch, missing and extra args | `[3,101,3]` |
| 21 | a throw out of a devirtualized call mid-loop | `["caught",3]` |

Cases 5 and 6 are the ones that would expose an over-eager `single_binding_closure_locals`, and both are exact. I also ran the whole probe under `PERRY_GC_SCHEDULE_RATE=1 FORCE_EVACUATE=1 VERIFY_EVACUATION=1 PROTECT_FROMSPACE=1` at two seeds to exercise the forwarded-closure heal you describe (word 0 holding the forwarding target failing the func-ptr compare) — node-identical both times.

One pre-existing gap the probe surfaced, **not yours**: `const f = (x: number) => x; f.name` gives `""` where node gives `"f"` (NamedEvaluation). Identical on main, so unrelated to this change.

**The two added commits:**

1. `PERRY_CALL_DEVIRT` wasn't registered as a build-cache input, so `codegen_env_vars_are_build_cache_inputs` was red. It empties the map and changes the emitted call sequence, so it's a cache key — added beside `PERRY_CALLEE_BINDING_RESOLUTION` with a comment in the house style.
2. `-D warnings` failed on `error: function collect_immutable_closure_bindings is never used`. That's the v1 module-wide collector in `closure_collect.rs`; the shipped path resolves through `spec_abi_sites::single_binding_closure_locals` threaded via `artifacts.rs` — the collector your own description names. Since v2 supersedes it and nothing references it, I removed it rather than wiring it. **If the module-wide reassignment-oracle version is still wanted, one revert restores it** — say so and I'll put it back.

Worth flagging that two of the three commits were titled `wip:` when I picked this up; I left the titles alone since the squash message is what lands, but if you were still iterating, the merge may have caught you mid-stream.

Validation: codegen 1347 passed, runtime 2819 (`RUST_TEST_THREADS=1`), perry --bins 1066, fmt clean, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped. Rebased onto main with an empty `--diff-filter=D` check (this branch had a cross-base port that had reverted #9086/#9092/#9093 collateral, restored in its own commit — I verified `archive_cache.rs` and `accessors.rs` now match main exactly).

Issue #9108, opened 2026-08-29
regression(main): shapes.rs ShapeId/ObjectHeader disagreement aborts the entire perry-runtime test suite (SIGABRT, ~1000 tests masked)


proggeramlug:
## Symptom

`cargo test -p perry-runtime --lib -- --test-threads=1` on current main ABORTS (SIGABRT, non-unwinding double panic):

```
test object::reserved_floor::tests::user_properties_read_back_through_the_get_path_at_scale ...
thread '...' panicked at crates/perry-runtime/src/object/shapes.rs:1539:9:
published ShapeId disagrees with authoritative ObjectHeader facts
thread caused non-unwinding panic. aborting.
```

~1800 tests pass before the crash; everything after `object::reserved_floor` is never run. **Every branch's runtime-suite signal is currently masked**, and worse: depending on output capture, the run can LOOK green (an isolation test's re-exec child prints `test result: ok. 1 passed; 2822 filtered out` and a piped `tail -1` reads that as the verdict — two of my gate batteries did exactly that today).

## Assessment

* The failing assertion is the shape-consistency invariant (`shapes.rs:1539`) tripping under the `reserved_floor` at-scale test — ShapeId/shape-key-index territory (#9000-#9005/#9013/#9022 lineage, plus the recent `reserved_floor` iterator-floor work #9075/#9080).
* Independent of my open PRs (#9105, the param-gate branch): their diffs contain no `perry-runtime` files; the aborting binary is built from main's runtime sources.
* The double panic (assert inside an unwinding context) is what converts a single test failure into a whole-suite abort — even after the root cause is fixed, consider making that assertion abort-safe so one shape bug can't mask the suite again.

## Repro

```
cargo test -p perry-runtime --lib -- --test-threads=1 > rt.log 2>&1; echo $?; tail -30 rt.log
```
(Do not pipe through early-closing readers; SIGPIPE kills the run mid-suite and fakes an exit-0.)

Filed by the perf campaign (session https://claude.ai/code/session_019WVcWKmYsUBnnFB7nBgbBJ); second suite-masking find of the day after #9106. Shape-lane ownership per the cc-socks channel.

Issue #9106, opened 2026-08-29
regression(main): wolf-ecs-shaped nested subclass loops lost their versioned fast clones (packed-guard sites 3 → 0)


proggeramlug:
## Symptom

On main `f3f405271c` (and `744df26d8e`), the `issue_8690_loop_versioned_arraylike` case `read_only_loops_have_preheader_proofs_and_fallback_free_fast_blocks` fails its shape assertion: `scan()`'s IR contains **0** `js_packed_arraylike_loop_guard(` sites where 3 are expected — the outer fast/slow copies no longer own preheader-versioned inner loops. `issue_8773_closure_capture_packed_loops` fails 1 case the same way. Program output stays correct; the flagship nested-subclass-loop shape silently runs at generic-loop speed.

## Window

Both suites were green on macOS at `8b40634abd` (the #9071 replay) and fail at `f3f405271c`. The window contains the packed/versioned-loop landings #9070 (versioned length-bound hoist), #9084 (read admissions), #9091 (unboxed accumulators), plus #9075/#9080/#9085. The shape gates were doing their job — this was caught only because an unrelated branch ran the full battery.

## Repro

```
cargo test -p perry --test issue_8690_loop_versioned_arraylike
```
(macOS; needs the default-target `libperry_runtime.a` — on a fresh box `cargo build -p perry-runtime-static` first, and note the suite fails in ~0.5s with an archive-mismatch harness error otherwise, which is NOT this regression.)

## Notes

The fixture is the wolf-ecs `Query extends Array` / `Archetype extends Array` nested scan — the exact shape the ECS campaign optimized to 2.3×/1.5× of node, so this likely walks back part of that standing until fixed. Independent of PR #9105 (verified by running both suites on detached main in the same worktree).

Reported by the ECS/ops perf campaign, session https://claude.ai/code/session_019WVcWKmYsUBnnFB7nBgbBJ; the loop lane's owner (85-successor / whoever holds #9084/#9091) is the natural assignee — coordination on the cc-socks channel.

proggeramlug:
Bisect result — **the window in this issue is wrong**: the culprit is #9062 (`da0eefbf99`, for-head declarator hoisting), an ancestor of the claimed-good endpoint; 8b40634abd already fails on a clean checkout here. #9070/#9084/#9091 are innocent. Mechanism + contained HIR fix in the PR just linked (first-declarator carve-out when the reorder is provably unobservable; all three affected suites plus #9052's own fixtures green).

PR #9109, opened 2026-08-29
perf(codegen): entry resolution reaches Any-typed callback params — __commonJS-shape calls 7.4 → 4.5 ns (typed parity)

proggeramlug:
## What

Entry resolution (#9071) reaches **`Any`-typed callback parameters** — the
esbuild `__commonJS`/`__esm` factory shape, where every module-init callback
erases to `Any` and paid the full `js_closure_callN` dispatcher per call:

* the emission gate in `emit_callee_binding_resolutions` admits a PARAMETER
  whose hint is absent or `Any` (a binding hinted as something
  non-function stays out — its resolution could only return null);
* the consuming arm's gate in `early_branches.rs` accepts a callee with a
  resolved-target map entry even when the `Function` hint is missing — the
  arm's runtime behavior (checked unbox, null-guarded direct diamond, full
  dispatcher fallback) was always hint-independent; the hint only selected
  who enters.

A resolution over a non-closure value (a number, `undefined`, a bound
function, a generator) returns null at entry and every call keeps the exact
dispatcher path.

## Numbers

Single-shape 50M-call probes, quiet Linux, same-build
`PERRY_CALLEE_BINDING_RESOLUTION=0` A/B:

| shape | off | on | node |
|---|---|---|---|
| `Any`-param callback loop (`__commonJS` shape) | 7.4 | **4.5** | 0.5 |
| typed-param control (#9071 path) | 8.1 | 4.5 | 0.5 |
| `Any`-param, conditionally called | 7.3 | **4.6** | — |

The `Any` shape reaches exact parity with the typed shape. This is the
"partial until the param-hint gate lifts" band flagged in #9105 and in the
claude-code startup profile split — the cc `--help` re-profile is gated on
this PR landing (plus #9105 and the #9106 fix).

## Semantics

Differential vs node identical: non-function/`undefined` through the `Any`
param (TypeError at the call), the param REASSIGNED mid-body (collector
excludes; the new value is observed), bound functions and ordinary functions
(`this === undefined`) through the `Any` param, arity-mismatch and rest
arrows, capture mutation between calls, generator functions. Kill-switch build
output-identical; the #9105 corpora re-run clean (the one diff is the
documented pre-existing no-TDZ-for-globals `TypeError` vs `ReferenceError`
line, gate-independent).

## Testing

* `RUSTFLAGS=-D warnings cargo check` (host excludes) — clean;
  `perry-codegen` 1830/0 — green.
* `perry-runtime --lib`: 1804 tests pass, then the suite ABORTS at
  `object::reserved_floor::…::user_properties_read_back_through_the_get_path_at_scale`
  (`shapes.rs:1539` ShapeId/ObjectHeader disagreement, double panic) — a main
  regression filed as **#9108** that masks the suite tail for every branch;
  this PR contains no `perry-runtime` files.
* Pinned-worktree integration battery; note `issue_8690` /
  `issue_8773` carry one pre-existing failure each on current main
  (**#9106**, the versioned-loop-clone regression — window
  `8b40634abd..f3f405271c`, owned by the claudecode session), identical on
  base and branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of direct calls through untyped or broadly typed callback parameters.
  * Enhanced compatibility with bundled module factory patterns.
  * Improved closure call performance by using faster resolution paths when safe.
  * Preserved existing runtime behavior with fallback safeguards for other callback types.

### DIFF
--- a/crates/perry-codegen/src/codegen/helpers.rs
+++ b/crates/perry-codegen/src/codegen/helpers.rs
@@ -1767,11 +1767,22 @@ pub(super) fn emit_callee_binding_resolutions(
         if ctx.local_closure_func_ids.contains_key(&id) {
             continue;
         }
-        if !matches!(
-            ctx.local_type_hint(&id),
+        // A `Function`-hinted binding is consumed by the guarded
+        // direct-dispatch arm. An UNTYPED (or `Any`) PARAMETER is consumed by
+        // the closure-call fallthrough's map check instead — that is the
+        // esbuild `__commonJS`/`__esm` shape, where every module factory's
+        // callback parameter erases to `Any` and paid the full dispatcher per
+        // call. A binding hinted as something else stays out: the resolution
+        // could only ever return null for it.
+        let hint = ctx.local_type_hint(&id);
+        let function_hinted = matches!(
+            hint,
             Some(perry_hir::types::Type::Function(function))
                 if !function.is_async && !function.is_generator
-        ) {
+        );
+        let any_param =
+            param_ids.contains(&id) && matches!(hint, None | Some(perry_hir::types::Type::Any));
+        if !function_hinted && !any_param {
             continue;
         }
         let value_box = if let Some(&capture_idx) = ctx.closure_captures.get(&id) {
--- a/crates/perry-codegen/src/lower_call/early_branches.rs
+++ b/crates/perry-codegen/src/lower_call/early_branches.rs
@@ -389,7 +389,17 @@ pub fn try_lower_closure_typed_local_call(
         }
         // The checked closure-unbox path below validates the current callee;
         // the erased type only decides whether to try that guarded dispatch.
-        if matches!(ctx.local_type_hint(id), Some(HirType::Function(_))) {
+        // #9105 follow-up: an entry-resolved binding takes this guarded arm
+        // even when its type hint erased to `Any` — the esbuild
+        // `__commonJS`/`__esm` factory-callback shape. The arm's runtime
+        // behavior is hint-independent (checked unbox, resolved-target
+        // diamond, full-dispatcher fallback); the hint only ever selected who
+        // enters it.
+        if matches!(ctx.local_type_hint(id), Some(HirType::Function(_)))
+            || ctx
+                .resolved_arrow_callback_targets
+                .contains_key(&(*id, args.len()))
+        {
             // #7803: the callee outlives the arguments here too, and this is
             // the arm on the failing stack — `core/schemas.ts` closure 138,
             // whose callee is a mutable-capture box read (`js_box_get_bits`)


proggeramlug:
Merged, after resolving a conflict against main — details at the end.

The reasoning that makes this safe is the sentence in the second hunk:

> The arm's runtime behavior is hint-independent (checked unbox, resolved-target diamond, full-dispatcher fallback); the hint only ever selected who enters it.

That's the right framing. Widening entry to an `Any`-typed *parameter* doesn't weaken any check — it lets a binding reach a guarded arm that was already going to validate it. And scoping to `param_ids` rather than all `Any` locals is what keeps it to the shape you're targeting.

So the thing worth testing is what happens when an `Any` parameter is **not** a closure, since that's the case the hint used to exclude. 21 shapes, byte-identical to node v26.5.1:

| | `Any` param holds | node |
|---|---|---|
| 5–10 | `5`, `null`, `undefined`, `{}`, `"str"`, `[]` | `TypeError` in all six |
| 11 | a class constructor (callable, not `[[Call]]`-able) | `TypeError` |
| 12, 13 | a native (`Math.abs`), a bound method | `3`, `33` |
| 14, 15 | a generator function, an async function | `"0[object Generator]…"`, `"string"` |
| 16, 17 | two different closures at one site; the value **reassigned mid-loop** | `[2,100]`, `[1,2,20,30]` |
| 19, 20 | throw from inside; recursion through the param | `RangeError`, `40320` |
| 1, 2, 21 | the `__commonJS` and `__esm` factory shapes, memoised | `[42,42]`, `[1,1,1]`, `["init"]` |

Cases 5–11 are the ones that would misdispatch if the guarded arm's fallback were hint-dependent; all six throw exactly as node does. 17 is the polymorphic-reassignment case.

**The conflict, and why it mattered.** The branch didn't merge cleanly against main: #9105 had added an early `continue` for `ctx.local_closure_func_ids` in the same block your hunk restructures. I kept both — the statically-known-callee skip first (a known func_id is better served by the direct inlinable call than by an entry-resolved indirect one), then your `function_hinted || any_param` gate. They compose correctly: the skip removes bindings that shouldn't be in the map at all, and your predicate widens who gets in among the rest.

Worth flagging how that first presented: my initial `git merge` left conflict markers in the tree, the build failed with `error: mismatched closing delimiter`, and my probe then ran a **stale binary** and reported "byte-identical to node" — a green result from a build that never happened. The numbers above are from the rebased tree after the resolution.

I also removed five `rustc-ice-*.txt` dumps rustc left in the worktree when it choked on the conflict-marked source.

Validation: codegen 1347 passed, runtime 2819 (`RUST_TEST_THREADS=1`, exit 0, no aborts), perry --bins 1066, fmt clean, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped. `git diff origin/main --diff-filter=D` empty.

PR #9110, opened 2026-08-29
fix(runtime): tombstone squeeze published stale facts — un-aborts the runtime suite (#9108)

### LINKED ISSUE #9108: regression(main): shapes.rs ShapeId/ObjectHeader disagreement aborts the entire perry-runtime test suite (SIGABRT, ~1000 tests masked)

proggeramlug:
Fixes #9108 — my tombstone default-on fallout (kill switch `PERRY_OBJECT_TOMBSTONES=0` cures the abort, which is how it was attributed). The squeeze shrank the keys array in place, then `set_object_live_slot_count`'s unchanged-bound early return asserted parity against the still-stamped pre-squeeze descriptor → double panic → SIGABRT at `object::reserved_floor::…at_scale`, silently masking the ~1000 tests behind it on every branch (and the re-exec child's '1 passed' line can masquerade as green in tail-style greps — see the issue).

Two coherence fixes: the squeeze orders `shape_drop → publish_object_shape_holes → set_object_live_slot_count`; and `publish_object_shape_holes` takes `logical_key_count` from the ARRAY rather than the lineage (identical for the O(1) hole delete where length is untouched; the only correct source after a squeeze). Release binaries were never affected — debug_assert only — which is why all differential batteries stayed byte-identical while the suite died.

Verification: `reserved_floor` suite 6/0; **full perry-runtime lib suite runs to completion, 2819 passed / 0 failed** (was: abort at ~test 1804). Suite-integrity blocker for everyone's verification, so flagging for a fast merge.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved object property deletion and shape updates to keep internal object state consistent.
  * Fixed edge cases that could trigger assertion failures when removing properties.
  * Ensured logical property counts remain accurate after object shapes are republished.

### DIFF
--- a/crates/perry-runtime/src/object/delete_rest.rs
+++ b/crates/perry-runtime/src/object/delete_rest.rs
@@ -1252,7 +1252,14 @@ unsafe fn squeeze_holes_and_delete(
         crate::gc::runtime_write_barrier_external_slot_span(keys as usize, elements as usize, out);
     }
     super::rebuild_array_layout_from_slots(keys);
-    set_object_live_slot_count(obj, std::cmp::min(out, alloc_limit) as u32);
+    // Publish the squeezed shape BEFORE touching the live-slot bound: the
+    // squeeze changed the keys array's length in place, and
+    // `set_object_live_slot_count`'s unchanged-bound early return asserts
+    // parity against the STAMPED descriptor — which still carries the
+    // pre-squeeze logical_key_count until the publish below runs. At scale
+    // the floored bound is usually unchanged, so that assert fired mid-
+    // transition (#9108, reserved_floor at-scale test SIGABRT).
+    //
     // Slots moved: the per-array key index and any stale descriptors for the
     // pre-squeeze states are wrong now. Drop the index (rebuilt on demand)
     // and publish the squeezed shape at exactly the surviving hole count —
@@ -1260,4 +1267,5 @@ unsafe fn squeeze_holes_and_delete(
     // for an iterator-family one.
     crate::object::shapes::shape_drop(keys);
     super::shapes::publish_object_shape_holes(obj, floor as u32);
+    set_object_live_slot_count(obj, std::cmp::min(out, alloc_limit) as u32);
 }
--- a/crates/perry-runtime/src/object/shapes_slot_list.rs
+++ b/crates/perry-runtime/src/object/shapes_slot_list.rs
@@ -226,9 +226,17 @@ pub(crate) unsafe fn publish_object_shape_holes(
     if generation == 0 {
         super::shape_id_exhausted_abort();
     }
+    // The key count comes from the ARRAY, not the lineage: the O(1) hole
+    // delete leaves the length untouched (array == lineage), but the squeeze
+    // shrinks it in place before republishing — carrying the lineage count
+    // there left a descriptor disagreeing with the authoritative keys edge,
+    // which the very next parity assert caught (#9108: reserved_floor
+    // at-scale SIGABRT took the whole suite down behind it).
+    let keys_ptr = current.keys as usize as *mut super::ArrayHeader;
+    let logical_key_count = crate::array::keys_array_len_capped_to_capacity(keys_ptr) as u32;
     let id = super::publish_shape_result(super::shape_descriptor_ensure_with_holes(
-        current.keys as usize as *mut super::ArrayHeader,
-        current.logical_key_count,
+        keys_ptr,
+        logical_key_count,
         current.live_inline_slot_count,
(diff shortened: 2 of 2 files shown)


proggeramlug:
Merged. Fixes #9108. This one deserves a note on how nearly I missed it, because the failure mode you documented in the issue caught me exactly as described.

**The masking is worse than the abort.** Under the dev profile on main (`RUST_TEST_THREADS=1 cargo test --lib -p perry-runtime`):

| | exit | `test …` lines emitted | abort markers |
|---|---|---|---|
| main | **101** | 2069 | 1 |
| this PR | **0** | **2836** | 0 |

767 tests never ran on main. And the `tail -1` of `^test result:` on the aborting run reads:

```
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2822 filtered out
```

— the re-exec child, exactly the masquerade your issue warns about. **My own validation harness greps `^test result:` and takes the last line**, so it would have reported this green. I now check the process exit code alongside it. Worth stating plainly since I have been reporting "runtime 2819 passed" on every PR today.

**Second reason I initially could not reproduce it**, which is worth recording for anyone else chasing this: `--profile perry-dev` inherits from `release`, so `debug-assertions` is **off** and `shapes.rs:1539`'s `debug_assert!` is compiled out entirely. Main is clean under `perry-dev` and aborts under the default dev profile. CI covers both (`test.yml:1057` runs the dev profile, `:1609` runs perry-dev), so the gate was real; a local perry-dev-only workflow is blind to it.

**The fix reads correctly on both halves.** Ordering `shape_drop → publish_object_shape_holes → set_object_live_slot_count` is right because the early return in `set_object_live_slot_count` asserts parity against the *stamped* descriptor, which still carries the pre-squeeze `logical_key_count` until the publish runs — and your observation that the floored bound is usually unchanged at scale is what makes that early return the one that fires, which is why this reproduced at scale and not on small objects. Taking `logical_key_count` from the array rather than the lineage is the correct source of truth given the squeeze shrinks the array in place; the two agree for the O(1) hole delete, so the change is a no-op on that path.

One adjacent finding, filed separately as **#9115**: `object_tombstone_deletes_enabled`'s doc comment says "Default OFF while the walker audit and differentials bake" while its body comment eight lines below says "Default ON" and the code defaults on. That is verbatim the `gc_incremental_enabled` shape CLAUDE.md cites (#6987, "a merge decision was made on the wrong one") — and it directly cost me time here, since I read the doc comment, concluded the mechanism was off by default, and initially mis-scoped the blast radius of this bug as latent rather than shipping.

Validation: runtime 2819 passed under **both** profiles with exit 0 (dev profile: 53.9 s, all 2836 test lines), codegen 1347, perry --bins 1066, fmt clean, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped.

Issue #8113, opened 2026-08-14
perf(object): remove the derivable object_type and field_count header words — 56 B -> 48 B


proggeramlug:
## Summary

Remove `ObjectHeader::object_type` and `field_count`, shrinking a common two-slot object from **56 B to 48 B** and the eight-slot case from 104 B to 96 B. This is an independent rung of #8047 that needs **none** of the GC descriptor-rooting work in #8112.

## Why this is a separate, landable rung

Measured with `rustc -O` on the exact `#[repr(C)]` shapes (LP64):

| variant | `size_of::<ObjectHeader>()` | `{a,b}` | 8-slot |
|---|---:|---:|---:|
| current | 32 | 56 | 104 |
| drop `object_type` only | 32 | 56 | 104 |
| drop `field_count` only | 32 | 56 | 104 |
| **drop both** | **24** | **48** | **96** |
| drop `keys_array` too (#8047) | 16 | 40 | 88 |

Neither field alone buys anything — the struct re-pads. Together they are a clean 8 bytes, exactly half of #8047's measured prize, and they do not touch the keys edge that #8112 has to redesign.

### 1. Replace the seven raw-offset-0 Error discriminators

`ObjectHeader.object_type` is prefix-punned against `error::ErrorHeader`, whose own first `u32` is also `object_type`. These read **raw offset 0** on an untyped pointer:

- `error.rs:750` (`js_error_is_error`), `error.rs:1542` (gates `(*error).errors`)
- `exception.rs:452`, `exception.rs:492` (`print_uncaught`)
- `value/dynamic_object.rs:531,548` and `:728,731`

Delete the word and offset 0 becomes `class_id`; an ordinary object whose `class_id` equals `OBJECT_TYPE_ERROR` is then read at `ErrorHeader`'s offsets. `ErrorHeader` is already allocated `GC_TYPE_ERROR` (`error.rs:198-203`), so `GcHeader.obj_type` is the replacement — the same move #8086 made for `object_is_regular`.

### 2. `proxy.rs:1523` needs the descriptor kind, not `object_is_regular`

```rust
&& (*(addr as *const crate::ObjectHeader)).object_type == crate::error::OBJECT_TYPE_REGULAR
```
[...]

PR #9113, opened 2026-08-29
perf(runtime): accessor install 2.33x — memoized prototype intrinsic + single-mint fresh-install path (stacks on #9103)

proggeramlug:
Second half of the __export/defineProperty startup lane (stacks on #9103 — its commit is included here since it wasn't merged yet; rebases clean if it lands first).

**The measured split first** (temporary env-bitmask ablation gates, quiet host, 500×2000 installs, ~6.0µs/install baseline; terms GC-coupled so they don't sum linearly):
- %Object.prototype% pollution probe: ~2.7µs — ~98% of it the PER-CALL intrinsic RESOLUTION (globalThis lookup + closure_get_dynamic_prop), the actual 6-name scan is 0.04µs
- ensure_key_in_keys_array: ~2.9µs (push + shape-publish machinery)
- note_descriptor_target: ~2.4µs — transition_object_shape_semantics minted TWICE per install
- owner_index_add ×2: ~1.9µs (O(N) Vec<String> dedupe scans)
- the two (usize,String) map inserts everyone suspected: only ~0.4µs
- **floor with everything ablated: 73ms/1M — below node's 148ms**, proving the side-table stack is the entire gap.

**Fixes:**
1. `object_prototype_has_desc_field` resolves the intrinsic through the existing memoized, root-scanned `object_prototype_addr()` (one slot load + forwarding heal) — no new invalidation machinery, and MORE correct: it's the realm intrinsic ToPropertyDescriptor actually reads through, immune to `globalThis.Object` rebinding.
2. New `install_fresh_accessor_property` used by #9103's fast arm: folds one epoch bump, ONE shape mint (was two), one guard-disable, one meta access, and skips the O(N) owner-index dedupe when the meta kind-bit is clear (the summary's own documented invariant: a clear bit proves the tables hold no entry; bits are sticky; violated precondition degrades to overwrite, regression-tested).

**Numbers** (quiet host, 1M installs; node 148ms): 500-key 5972 → **2563ms (2.33×)**; realistic 28-key __export shape 3989 → **1426ms (2.80×)**. Honest caveat: the generic descriptor-literal path barely moves (its cost is the duplicated install stack it still runs) — folding it needs the redefine cases and is left out deliberately. Next term is ensure_key_in_keys_array's append porch (~2.9µs) but that walks into the #8113 mint-then-stamp and #9029 lineage contracts and deserves its own change (analysis included in the lane notes).

**Validation:** codegen 1349/0; runtime 2823/0 (--test-threads=1; the reserved_floor at-scale exclusion is #9108, pre-existing on the base and fixed separately in #9110); `test_gap_9053_export_getter_descriptor.ts` byte-identical to node; fmt/census/file-size green. Implemented by a subagent in an isolated worktree; reviewed and shipped by the coordinating session.

## Summary by CodeRabbit

  * Improved performance for `Object.defineProperty` calls that add a getter with `enumerable: true`.
  * Streamlined internal property installation to reduce overhead during accessor creation.
* **Bug Fixes**
  * Preserved existing validation, error handling, key conversion, and behavior for unsupported or complex objects.
  * Improved consistency between optimized and standard property-definition paths.
* **Tests**
  * Added coverage confirming equivalent behavior and correct handling of repeated accessor installations.


proggeramlug:
Merged. This was stacked on #9103, and unblocking it took a bit of surgery worth documenting.

**The rebase.** Its first commit *is* #9103, but #9103 landed on main in a **reworked** form — I had a Codex agent convert that file's raw-handle reads to `with_const_ptr`/`across_const` before merging it. So cherry-picking this branch whole produced multi-hunk conflicts in `define_get_accessor.rs` between the pre-rework and post-rework versions. Applying only this PR's **own** commit onto current main is clean (227 insertions, no conflicts, `--diff-filter=D` empty), so that is what landed.

**Two commits I added:**
- `-D warnings` rejected two now-`unnecessary unsafe` blocks in the new tests (lines ~340 and ~385). #9103's handle-custody rework made those call sites safe, so the blocks became redundant — a knock-on of the rework rather than anything wrong here. Removed and de-indented.
- rustfmt.

**Validation:** runtime **2833 passed** (exit 0, 0 abort markers), perry --bins 1066, fmt clean, `run_lint_gates.sh` **all 60 gates passed**. I also re-ran my #9103 `defineProperty` probe — 22 shapes, byte-identical to node v26.5.1, including every case that must delegate off the fast arm (proxy trap, redefine, symbol key, numeric key, `preventExtensions`, non-callable getter, `get`+`value` together, polluted `Object.prototype`).

I did not independently reproduce the 2.33x install figure; the correctness surface and the gates are what I verified here.

Issue #3987, opened 2026-06-01
runtime/string: align ToString, String wrappers, indexed reads, and extra-arg calls


andrewtdiz:
## Summary

String parity remains a high-value runtime blocker: Perry diverges on the abstract `ToString` operation, `new String(...)` wrapper state, inherited String prototype behavior, indexed string reads, and permissive extra-argument calls such as `charAt(a, b, c, d)`.

## Evidence

Run from `origin/main` at `1f6d20e19` on 2026-06-01, Node `v25.9.0`, Test262 `4249661388e5d3f92a85186213da140a6481490f`:

```bash
scripts/test262_subset.py --root /tmp/perry-test262-4249661 --dir built-ins/String --max 180 --timeout 10 --sample-cap 50 --report /tmp/perry-c262-builtins-string.json
```

Observed: 84 pass, 94 runtime-fail, 2 compile-fail, 46.7% sample parity.

Representative failures:

- `built-ins/String/S9.8.1_A9_T1.js`: `String(1000000000000000000000)` expected `1e+21`, Perry prints the full decimal form.
- `built-ins/String/S9.8.1_A9_T2.js`: `String(0.0000001)` expected `1e-7`, Perry prints decimal notation.
- `built-ins/String/S15.5.1.1_A1_T19.js`: `String(new Array(1,2,3))` expected `1,2,3`, Perry returns `[object Object]`.
- `built-ins/String/S15.5.5.1_A2.js`: `new String("globglob")` should have own `length`; Perry does not expose it.
- `built-ins/String/15.5.5.5.2-3-6.js`: string index `s[NaN]` should be `undefined`, Perry returns a character.
- `built-ins/String/prototype/charAt/S15.5.4.4_A1.1.js`: Perry compile-fails because `String.charAt` expects at most one arg; JavaScript should ignore extra args.

## Product triage

User outcome: string-heavy package code, formatting, CLI output, protocol parsing, and object coercion match Node for everyday operations.

Shortest path: fix the shared ToString and String wrapper/indexing paths first; method-tail polishing can follow once those foundations stop poisoning many rows.

Manual/narrow assumptions to test: compare Node/Perry output for exponent-format numbers, object/array/function ToString hooks, `new String("abc").length`, inherited String prototype props, invalid string indices, and `charAt`/`charCodeAt` extra arguments.

[...]

PR #9114, opened 2026-08-30
perf(runtime): small-int string cache in String() coercion + itoa integer operands in js_string_concat_box

proggeramlug:
## What

Two surgical runtime changes on the string-concat hot paths:

1. **`String(n)` routes through the small-int cache.** `js_string_coerce`'s regular-number arm formatted every number with the `format!` machinery. It now delegates to `js_number_to_string`, whose `SMALL_INT_CACHE` answers `String(i)` for 0..255 with an interned longlived string and **no allocation**; the fallback is the same shared `js_format_f64` (#3987 scientific-notation semantics), so output is bit-identical on every input.

2. **`js_string_concat_box` itoas integral number operands.** The pairwise (template-literal) concat punted ANY non-string operand to `js_dynamic_string_or_number_add` — a full ToPrimitive round trip plus `format!` per op. An integral plain-f64 operand in `0..=999_999_999` (`fract()==0`; the NaN-box tag check excludes boxed values, and the sign bit excludes negatives and -0) is now itoa'd into a stack buffer and flows through the **same SSO/heap byte-assembly as the two-string case**. itoa and `Number::toString` print that range identically; fractional / negative / huge / NaN-boxed operands keep the dynamic arm, and number+number pairs never enter (one side must still be a real string), so the annotation-lie semantics are unchanged.

## Measurements

Mac mini (quiet benchmark host), 11 interleaved base/branch/node triples, median ns/op (min was within 0.1 everywhere):

| shape | base | branch | node | Δ | vs node |
|---|---|---|---|---|---|
| `String(i & 255)` | 35.4 | **4.6** | 4.9 | **−87.0%** | **0.9× — beats node** |
| `\`id-${i & 255}\`` (template) | 54.6 | **21.9** | 2.2 | **−59.9%** | 10.0× |
| `"id-" + (i & 255)` | 26.3 | 26.2 | 2.2 | −0.4% | 11.9× |
| `"id-" + "x"` | 8.1 | 8.1 | 0.5 | +0.0% | 16.2× |
| `a + b` (two vars) | 25.4 | 25.3 | 0.5 | −0.4% | 50.6× |
| `s += "ab"` (grow) | 14.6 | 14.6 | 4.2 | +0.0% | 3.5× |
| concat-then-compare | 26.1 | 25.9 | 6.9 | −0.8% | 3.8× |

Same-run A/B on the dev box agrees (string_of_int 36.9→4.8, template_int 57.0→22.9, rest flat). An earlier draft cost the pure two-string path ~0.4 ns (two zeroed 32-byte itoa buffers); the final structure routes two real strings to the assembly tail before any number buffer exists, and the in-pair delta on `"id-" + "x"` is 0.0.

The untouched rows are the next, separate lever: their remaining cost is one heap string allocation per op (`string_storage_alloc` + memmove; `"id-" + int` already itoas via `js_string_concat_value_box` but misses SSO at 6 bytes).

## Correctness

- 264-line differential vs node — `String(v)`, `` `id-${v}` ``, `"x"+v`, `v+"y"`, `""+v`, `v+""` over -0, ±1e-7, 1e20/1e21/1e22, NaN, ±Infinity, 2^31/2^32/2^53 boundaries, 5e-324, plus non-number operands (bool/null/undefined/object/array/bigint), Symbol-throws, and a mixed grow-concat: **byte-identical**.
- Runtime string/number/dynamic-add test filters: 106 passed.
- Full runtime suite (single-threaded, with #9110 cherry-picked on top for validation only — it fixes the pre-existing main SIGABRT at ~test 1804 this PR is otherwise subject to): **2825 passed / 0 failed**.
- Gate battery: `-D warnings` 0, codegen suite 1830/0, lints clean (addr-class, file-size, raw-handle debt none raised), integration: issue_8655 2/2, issue_8897 3/3. `issue_8690::read_only_loops_have_preheader_proofs_and_fallback_free_fast_blocks` fails on this branch **and is pre-existing #9106** (lost loop clones; the bisect there independently marks #9070/#9077 — both ancestors of this branch's base 0b6dea2f2e — as bad). This diff is runtime string code only; a pristine-base rerun is in flight and I'll comment with its result.
- The single-threaded runtime-lib line in the gate log shows the known re-exec masquerade of pre-existing #9108 (fixed by #9110); the overlay run above is the real full-suite verdict.

## Summary by CodeRabbit

  - Improved efficiency when converting numbers to strings, especially for commonly used small integers.
  - Optimized string concatenation involving whole-number values, reducing unnecessary processing and allocations.
- **Compatibility**
  - Preserved existing number-to-string output and concatenation behavior, including special numeric values and number-to-number operations.

### DIFF
--- a/crates/perry-runtime/src/builtins/numbers.rs
+++ b/crates/perry-runtime/src/builtins/numbers.rs
@@ -671,11 +671,13 @@ pub extern "C" fn js_string_coerce(value: f64) -> *mut StringHeader {
     } else if jsval.is_int32() {
         jsval.as_int32().to_string()
     } else {
-        // Regular number — ECMAScript NumberToString. #3987: route through the
-        // shared `js_format_f64` so `String(1e21)` → "1e+21" / `String(1e-7)`
-        // → "1e-7" (scientific notation for |n| >= 1e21 / < 1e-6) instead of
-        // Rust's full-decimal `to_string()`, matching `.toString()` and Node.
-        crate::string::js_format_f64(value)
+        // Regular number — delegate to `js_number_to_string`: its small-int
+        // cache answers `String(i)` for 0..255 with an interned longlived
+        // string and NO allocation (the `format!` machinery this arm used to
+        // run per call was ~15% of the string-ops profile), and its fallback
+        // is the same shared `js_format_f64` (#3987 scientific-notation
+        // semantics), so the output is bit-identical on every input.
+        return crate::string::js_number_to_string(value);
     };
 
     js_string_from_bytes(result.as_ptr(), result.len() as u32)
--- a/crates/perry-runtime/src/string/concat.rs
+++ b/crates/perry-runtime/src/string/concat.rs
@@ -137,14 +137,67 @@ fn bytes_all_ascii(data: *const u8, len: u32) -> bool {
 pub extern "C" fn js_string_concat_box(l_value: f64, r_value: f64) -> f64 {
     let mut scratch_l = [0u8; crate::value::SHORT_STRING_MAX_LEN];
     let mut scratch_r = [0u8; crate::value::SHORT_STRING_MAX_LEN];
-    let (Some(l), Some(r)) = (
-        str_bytes_from_jsvalue(l_value, &mut scratch_l),
-        str_bytes_from_jsvalue(r_value, &mut scratch_r),
-    ) else {
-        // `str_bytes_from_jsvalue` returns `None` for exactly the non-string
-        // values, so this is the annotation-lie arm and nothing else.
-        return unsafe { crate::value::js_dynamic_string_or_number_add(l_value, r_value) };
-    };
+    // A PLAIN-F64 small-integer operand next to a string operand is the
+    // template-literal hot shape (`` `id-${i}` ``): itoa it into a stack
+    // buffer and flow it through the same SSO/heap assembly the two-string
+    // case uses, instead of handing the pair to the full dynamic `+` (whose
+    // ToPrimitive round trip and `format!` formatting were the template
+    // path's entire overhead). Only exact spec-identical cases are taken:
+    // an integral value in 0..=999_999_999 prints identically under itoa
+    // and `Number::toString`; anything else — fractional, negative, huge,
+    // NaN-boxed — keeps the dynamic arm. One side must still be a REAL
+    // string, so the annotation-lie semantics of the dynamic arm are
+    // unchanged for number+number.
+    #[inline]
+    fn itoa_operand(bits_value: f64, buf: &mut [u8; 32]) -> Option<(*const u8, u32)> {
+        let bits = bits_value.to_bits();
+        let tag = bits >> 48;
+        let is_plain_f64 = tag < 0x7FF8 || (tag == 0x7FF8 && (bits & 0x000F_FFFF_FFFF_FFFF) == 0);
+        if is_plain_f64
+            && bits_value.fract() == 0.0
+            && (0.0..=999_999_999.0).contains(&bits_value)
+        {
+            let len = fast_itoa_u32(bits_value as u32, buf);
+            Some((buf.as_ptr(), len as u32))
+        } else {
+            None
+        }
+    }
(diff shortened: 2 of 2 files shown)


proggeramlug:
Pristine-base attribution run complete: at the branch base 0b6dea2f2e (origin/main, #9100) with **no string diff present**, `issue_8690::read_only_loops_have_preheader_proofs_and_fallback_free_fast_blocks` fails with the identical assertion ("the outer fast/slow copies each own a preheader-versioned inner loop"); the other two tests in the suite pass, exactly matching this branch's gate result. The failure is pre-existing #9106 (lost loop clones — its bisect independently marks #9070/#9077, both ancestors of this base, as bad), not this PR. Everything else stands: codegen 1830/0, full runtime suite 2825/0 (with #9110 overlaid to clear the pre-existing #9108 abort), 264-line formatting differential byte-identical vs node.

proggeramlug:
Rebased onto current main (now carrying #9110 and #9116): **both pre-existing-failure caveats in the description are obsolete**. Fresh full gate battery on the rebased stack: `-D warnings` 0, codegen 1830/0, **full runtime suite 2819/0 natively** (no overlay), lints clean, integration issue_8655 2/2 / issue_8690 3/3 / issue_8897 3/3. Follow-up stacked on this PR: #9118 (lean allocation on the same hot paths — `"id-"+i` a further −42%).

proggeramlug:
Merged, plus a rustfmt commit (`cargo fmt --all -- --check` is a lint gate; the `is_plain_f64` condition wanted joining onto one line).

**The numbers are large and they hold up.** Interleaved best-of-3, 4M iterations:

| | main | this PR | node | |
|---|---|---|---|---|
| `String(i % 256)` | 163 ms | **24 ms** | 44 ms | **6.79x** |
| `` `k${i % 500}` `` | 324 ms | **141 ms** | 24 ms | 2.30x |
| `"" + (i % 1000)` | 86 ms | **63 ms** | 14 ms | 1.37x |

`String(int)` now beats node outright. The template-literal row is the one I'd point at for follow-up — 2.3x better but still 5.9x node, so the remaining cost there isn't the int→string conversion this fixes.

**Correctness is the real question for a value cache, so I concentrated there** — 27 shapes, byte-identical to node v26.5.1:

| | what it checks | node |
|---|---|---|
| 1 | every integer in `-300..=300` — the cache's range and both edges | 2485 total chars |
| 3 | `-0` through six different coercion paths | `"0"` in all six |
| 4, 5, 6 | a cached string is not observably aliased: `===`, `.length`, `concat`, `padStart`, `repeat`, then re-read the original | unchanged each time |
| 10 | the i32/u32 boundaries: `±2147483647/8`, `4294967295/6` | exact |
| 7, 8, 9 | `1e21`, `1e-7`, `1/3`, `MAX_SAFE_INTEGER`, `EPSILON`, `MIN_VALUE`, `NaN`, `±Infinity`, `1.0`→`"1"` | exact |
| 20, 21 | `toString(radix)` for 16/2/36, negative, zero, and `(1.5).toString(2)` | unaffected |
| 22, 23 | `valueOf`/`toString` precedence on a plain object | `"5"`/`"[object Object]"`, `"9"`/`"T"` |
| 18, 19 | BigInt, `true`, `null`, `undefined` through the same paths | exact |
| 26, 27 | 2000-iteration `+=` accumulation and a 100-element `join` | exact |

Case 3 is the one I most wanted green — `-0` must stringify to `"0"`, and an itoa path that goes through a signed conversion is exactly where that breaks. Cases 4–6 cover the other cache hazard: if a cached string were handed out in a way that let a caller mutate or alias it, `String(42)` would stop being `"42"` after someone appended to it.

Case 10 matters for the `(0.0..=999_999_999.0)` window in `fast_itoa_u32` — the values just outside it must fall back cleanly, and they do.

Validation: runtime 2819 passed (`RUST_TEST_THREADS=1`, exit 0, 0 abort markers), perry --bins 1066, fmt clean, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped.

Note for #9118, which touches `numbers.rs` and `concat.rs` too: it will need a rebase onto this.

Issue #9052, opened 2026-08-29
Classic for lexical declarators are lowered out of source order


proggeramlug:
## Summary

Perry lowers multiple lexical declarations in a classic `for` initializer out of source order. A later initializer cannot reference an earlier binding:

```ts
for (let i = 0, limit = 2, next = i + 1; i < limit; i++, next++) {
  console.log(i, next);
}
```

This blocks Three.js `ExtrudeGeometry`, which contains:

```js
for (let i = 0, il = contour.length, j = il - 1, k = i + 1; ...)
```

Reproduced on `e81548ed17ff9488c3f8c19256048dc8d8497867`. The faulty lowering remains on `main` at `bbf1bcf00921b1d15fec77866bd2f5f9fbfdff95`.

## Actual behavior

```text
ReferenceError: i is not defined
```

## Expected behavior

```text
0 1
1 2
```

ECMAScript initializes the declarators from left to right in one lexical environment. `next` must see the already-initialized `i` binding.

## Root cause

In `crates/perry-hir/src/lower/stmt.rs`, the non-`var` classic-for branch currently does this:

```rust
for decl in var_decl.decls.iter().skip(1) {
    // lower declarators 2..N
}

if let Some(decl) = var_decl.decls.first() {
    // lower declarator 1
}
```

Consequently `i` is not registered when `k = i + 1` is lowered, so `i` becomes an unresolved/global reference.

## Proposed fix

HIR currently has only one `Stmt::For.init` slot. For a multi-declarator lexical head, lower every declarator into the already-created loop-scoped prelude in source order, then emit `Stmt::For { init: None, ... }`.

The declaration initializer must be lowered only after earlier bindings have been registered. Destructuring declarations must follow the same ordering rule. Preserve the existing lexical scope and per-iteration capture behavior.

## Regression coverage

Add both module-level and function-body lowering tests and assert:

- emitted declarations are ordered `i`, `limit`, `previous`, `next`;
- `next`'s initializer contains `LocalGet(i_id)`, not `GlobalGet`;
- the following compile/run output matches Node:

```ts
for (let i = 0, limit = 2, next = i + 1; i < limit; i++, next++) {
  console.log(i, next);
}
```

Also cover a destructuring declarator and a closure capturing the per-iteration binding so the source-order fix does not regress lexical loop semantics.

PR #9116, opened 2026-08-30
fix(hir): keep the pure first declarator in For::init — restores the versioned loop clones (#9106)

### LINKED ISSUE #9106: regression(main): wolf-ecs-shaped nested subclass loops lost their versioned fast clones (packed-guard sites 3 → 0)

proggeramlug:
Fixes #9106. **The culprit is #9062, not the issue's stated window** — the bisect (fast IR-grep oracle: count `js_packed_arraylike_loop_guard` in `scan()`'s IR) walked to a CI-only commit, an impossible verdict that exposed the window as wrong; the claimed-good endpoint 8b40634abd already fails on this machine. #9070/#9084/#9091 are innocent.

**Mechanism:** #9062 correctly fixed out-of-order initialization of multi-declarator lexical for heads by hoisting every declarator into a pre-loop prelude — leaving `For::init = None`. But every versioned counted-loop matcher identifies the counter exclusively through `For::init` (`stable_packed_loop.rs`'s `match_candidate` returns None immediately). The canonical wolf-ecs idiom `for (let j = 0, length = current.length; j < length; j++)` is a multi-declarator head, so the nested scan loops silently fell to generic lowering: 0 guard sites where 3 are expected, correct output, fast clones gone.

**Fix (HIR, contained):** new predicate `for_head_first_decl_keeps_init_slot` — when the FIRST declarator is a plain identifier bound to a pure literal and no tail declarator mentions it (swc Visit scan over tail patterns + initializers), the tail-only hoist is provably unobservable, so the first declarator stays in `For::init` and only the tail hoists. Order-observable heads — including both #9052 fixtures, whose tails read `i` — stay on #9062's prelude path; hoisted tails keep the `classic_for_lexical_bindings` capture semantics. Applied at both twin lowering sites; HIR regression test pins the carve-out; changelog fragment included.

**Validation (at tip, post-fix):** `issue_8690_loop_versioned_arraylike` 3/3 (guard sites 3 again), `issue_8773_closure_capture_packed_loops` 4/4, `issue_9052_for_lexical_declarators` 2/2 (the culprit's own wins intact), perry-codegen lib 1347/0, native_proof_regressions 284/0, perry-hir lib 360/0, fmt clean.

Goal-relevant: this restores the loop family's beat-node standing (the ECS 2.3×/1.5× shape ran at generic speed on main). Implemented by a subagent in an isolated worktree; reviewed and shipped by the coordinating session.

## Summary by CodeRabbit

  * Restored optimized handling for common counted `for` loops, such as `for (let i = 0, len = arr.length; i < len; i++)`.
  * Preserved existing source-order behavior when changing declaration handling could affect observable execution.

* **Bug Fixes**
  * Improved loop variable handling to preserve correct per-iteration capture semantics.

* **Tests**
  * Added regression coverage for optimized multi-declaration `for` loops.

### DIFF
--- a/changelog.d/9106-for-head-counter-keeps-init-slot.md
+++ b/changelog.d/9106-for-head-counter-keeps-init-slot.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Classic `for` heads shaped `for (let i = 0, len = arr.length; i < len; i++)` regained their versioned packed-loop fast clones: when hoisting the tail declarators around a literal-initialized counter is provably unobservable, the counter stays in the loop's own init slot the counted-loop matchers key on, while order-observable heads keep the #9062 source-order lowering.
--- a/crates/perry-hir/src/lower/for_multi_decl_tests.rs
+++ b/crates/perry-hir/src/lower/for_multi_decl_tests.rs
@@ -79,3 +79,58 @@ fn function_classic_for_lexical_declarators_lower_in_source_order() {
         .expect("run should lower");
     assert_source_ordered_prelude(&function.body);
 }
+
+#[test]
+fn safe_counter_head_keeps_the_for_init_slot() {
+    // #9106: `let i = 0, len = arr.length` — a literal-initialized counter
+    // that no tail declarator mentions. Hoisting the tail around it is
+    // unobservable, so the counter stays in `For::init` (the versioned
+    // counted-loop matchers key their admission on it) and only the tail
+    // moves to the loop-scoped prelude.
+    let hir = lower(
+        "function run(arr: number[]) { let sum = 0; for (let i = 0, len = arr.length; i < len; i++) { sum += arr[i]; } return sum; }",
+    );
+    let function = hir
+        .functions
+        .iter()
+        .find(|function| function.name == "run")
+        .expect("run should lower");
+    let for_index = function
+        .body
+        .iter()
+        .position(|stmt| matches!(stmt, Stmt::For { .. }))
+        .expect("classic for should be present");
+    let len_id = function.body[..for_index]
+        .iter()
+        .find_map(|stmt| match stmt {
+            Stmt::Let { id, name, .. } if name == "len" => Some(*id),
+            _ => None,
+        })
+        .expect("tail binding should be hoisted into the prelude");
+    assert!(
+        hir.classic_for_lexical_bindings.contains(&len_id),
+        "hoisted tail binding keeps per-iteration capture semantics"
+    );
+    let crate::ir::Stmt::For {
+        init: Some(init), ..
+    } = &function.body[for_index]
+    else {
+        panic!(
+            "counter must stay in For::init: {:#?}",
+            function.body[for_index]
+        );
+    };
+    let counter_id = match init.as_ref() {
+        Stmt::Let {
+            id,
+            name,
+            init: Some(crate::ir::Expr::Integer(0)),
+            ..
+        } if name == "i" => *id,
+        other => panic!("counter Let expected in For::init, got {other:#?}"),
+    };
+    assert!(
+        !hir.classic_for_lexical_bindings.contains(&counter_id),
+        "the init-slot counter is handled by the For machinery, not the prelude set"
+    );
+}
--- a/crates/perry-hir/src/lower/stmt.rs
+++ b/crates/perry-hir/src/lower/stmt.rs
@@ -1575,13 +1575,22 @@ pub(crate) fn lower_stmt(
                             // head into the loop-scoped prelude in source order,
                             // registering each binding before the next
                             // initializer is lowered.
-                            let has_multiple_decls = var_decl.decls.len() > 1;
-                            for decl in
-                                var_decl
-                                    .decls
-                                    .iter()
-                                    .skip(if has_multiple_decls { 0 } else { 1 })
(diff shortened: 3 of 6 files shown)


proggeramlug:
Merged. Fixes #9106.

**The attribution correction is the most valuable part of this PR**, and I confirmed it: #9062 is the culprit, and #9070/#9084/#9091 are innocent. Reproduced on main (`d1d6d03b91`), running from the `perry` crate:

| | main | this PR |
|---|---|---|
| `issue_8690_loop_versioned_arraylike` | **FAILED** 2 passed / 1 failed | **ok** 3 passed |
| `issue_8773_closure_capture_packed_loops` | **FAILED** 3 passed / 1 failed | **ok** 4 passed |

with `assertion failed: the outer fast/slow copies each own a preheader-versioned inner loop`. Program output is identical to node on both arms — this was purely a silent fall to generic lowering, which is exactly the kind of regression that a parity suite cannot see.

Your note that the bisect "walked to a CI-only commit, an impossible verdict that exposed the window as wrong" is worth keeping in the changelog fragment. A bisect landing on a commit that cannot affect the subject is evidence about the *window*, not about the commit — easy to rationalize away instead of acting on.

**The carve-out is correctly conditioned.** The risk here is obvious — #9062 removed this reordering precisely because it was observable — so the question is whether the three conditions are jointly sufficient. First declarator a plain identifier, initializer a pure `Expr::Lit`, and no tail declarator mentioning the binding's name anywhere in pattern or initializer. A literal has no effects and no dependencies, and if nothing names the binding, neither its value nor its TDZ state can be observed before the tail runs. That closes it.

I tested the ways it could leak — 18 shapes, byte-identical to node, and #9062's own `for_multi_decl` tests still pass (now 3):

| | shape | node |
|---|---|---|
| 3 | both initializers are calls — evaluation order visible | `["f0","f1"]` |
| 4 | tail declarator reads the first (`let i = 0, j = i + 5`) | `[[0,5],[1,5]]` |
| 5 | chained (`a = 1, b = a * 2, c = b * 3`) | `[[1,2,6]]` |
| 6 | comma-expression side effects in both initializers | `["i","j"]` |
| 7 | TDZ reference from the tail | `ReferenceError` |
| 8, 9 | per-iteration closure capture, single and multi-declarator | `[0,1,2]`, `[[0,100],[1,101],[2,102]]` |
| 12, 13 | first initializer NOT a literal — must take the prelude path | `6`, `15` |
| 14, 15 | shadowing an outer `i`; a tail name that is a prefix of the first | `[[0,1],"outer"]`, `[[0,9],[1,9]]` |
| 17, 18 | mutating the array / its `length` inside a length-cached head | `[6,4]`, `NaN` |

Cases 4, 5 and 7 are the ones that would fail if the "no tail mentions it" check were weak, and 12/13 confirm the non-literal path is untouched.

Two process notes from validating this:

- These suites live in `crates/perry/tests/`, not `perry-codegen`. My first run named the wrong crate and printed **no `test result:` line at all**, which reads as "nothing to report" rather than as an error — the same silent-green shape as the rest of today.
- My local `origin/main` ref was stale (still at #9105), so my first full run showed the #9108 SIGABRT and I nearly attributed it here. Re-fetched: exit 0 under both profiles, 0 abort markers.

Validation: hir 360 passed (`for_multi_decl` 3/3), codegen 1347, runtime 2819 under **both** the dev profile (47.5 s, exit 0, no aborts) and perry-dev, perry --bins 1066, both regressed suites green, fmt clean, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped.

Issue #6728, opened 2026-07-20
runtime: pi TUI Enter (CR) keypress does not submit under perry — raw-stdin/Kitty keypress decode divergence (real pi interactive blocker)


proggeramlug:
The ACTUAL blocker for pi's interactive TUI under perry (supersedes the async-generator theory — see below). Pi renders and accepts typed text, but **pressing Enter (CR `0x0d`) does not submit the message** under perry, where node (same bundle) submits and renders the response.

## Evidence (fixed-binary A/B, decisive)
Using `/tmp/pi-fixed` = pi compiled WITH the async-generator await-suspension fix (#6727), interactive PTY repro, bogus API key (a valid request 401s — that's the expected visible result):
| Enter byte | perry result | node result |
|---|---|---|
| CR `0x0d` | "hello" stuck in input, **no submit, no "Working…", no 401** (19547 B) | submit → "Working…" → "Error: 401 authentication_error" (33338 B) |
| LF `0x0a` | input processed differently (25792 B) but **still no 401** | — |
| CRLF `0x0d0a` | 25792 B, no 401 | — |

So no Enter variant submits+responds under perry; node does with CR. **The async-generator fix (#6727) does NOT change this** (perry CR result byte-identical with and without it) — confirming the keypress→submit path, not async-generator event forwarding, is what blocks pi. (#6727 fixes a real, separate async-generator `await`-on-pending-Promise deadlock and is worth landing on its own; it would be the NEXT blocker once submit works.)

## Mechanism (where to look)
Pi reads raw stdin itself: `process.stdin.setRawMode(true)` (bundle ~232037) + `process.stdin.on("data", stdinDataHandler)` (~232095), and runs a custom keypress parser that also negotiates the **Kitty keyboard protocol** (`queryAndEnableKittyProtocol`, `KITTY_KEYBOARD_PROTOCOL_QUERY`). Enter/CR decoding flows through that raw-byte parse (`charCodeAt`, string/`Buffer` handling, CR/LF/CRLF logic at ~50346, Kitty sequence parsing). A perry divergence in one of: (a) how `stdin.on("data")` delivers the chunk under raw mode (Buffer vs string, encoding, chunk boundaries), (b) `charCodeAt`/string comparison in the parser, (c) the Kitty-protocol negotiation buffer handling, or (d) `setRawMode`/termios not matching node (CR vs LF delivery). The CR-vs-LF asymmetry points at the parser receiving/classifying the Enter byte differently than under node.

[...]

Issue #6655, opened 2026-07-19
runtime: audit dynamic_arith operand rooting — raw NaN-boxed operands held across GC-capable to_numeric coercions (pre-existing, file-wide)


proggeramlug:
While fixing a CodeRabbit finding on #6650 (unrooted NaN-boxed operands held across GC-capable coercions in the NEW `js_dynamic_ushr`), the same latent pattern shows in the PRE-EXISTING dynamic operators of `crates/perry-runtime/src/value/dynamic_arith.rs`:

```rust
pub unsafe extern "C" fn js_dynamic_pow(a: f64, b: f64) -> f64 {
    let a = to_numeric(a);   // can run user ToPrimitive → allocate → GC
    let b = to_numeric(b);   // raw `b` (heap pointer?) held unrooted across the above
    ...
```

`to_numeric` on an object operand runs user `ToPrimitive`/`valueOf` (can allocate and trigger GC/evacuation). Any operator with the `to_numeric(a); to_numeric(b)` prelude holds raw `b` — and the coerced `a`, when it is a BigInt pointer — unrooted across potentially-moving operations. The correct discipline exists in the same file (`dynamic_bigint_binary_op`: RuntimeHandleScope + root_nanbox_f64 + reload via get_nanbox_f64) and was applied to `js_dynamic_ushr`/`throw_mix_bigint` in #6650 (commit 169456bc9).

Ask: audit every exported `js_dynamic_*` (and any other f64-operand entry point in the file) for operands held across GC-capable calls; apply the rooting pattern; consider a repro via a `valueOf` that forces allocation pressure (the fast path at ~line 424 documents when the scope can legitimately be skipped — primitives only).

Low observed frequency (needs GC inside ToPrimitive with a pointer operand pending), but it is exactly the stale-pointer class behind past campaigns (#6471 etc.) — worth closing file-wide.

PR #9118, opened 2026-08-30
perf(runtime): lean allocation on the pairwise string-concat hot paths

proggeramlug:
Stacked on #9114 (first commit here is that PR; review the second commit).

## What

Four costs a `sample` profile attributed on the `"id-" + i` loop, removed from the pairwise concat hot paths without touching semantics:

1. **Rooting (~7% of the loop).** `js_string_concat_value` created a `RuntimeHandleScope` and rooted the prefix unconditionally at entry. The number arm now allocates through `string_storage_alloc_no_collect` first — its `Some` contract ("the open nursery block served this, nothing on the heap moved") keeps every raw prefix read valid with no root at all. The block-boundary `None` fallback takes the original rooted path, and the user-`toString` slow arm still always roots (#6655 unchanged).

2. **libc PLT calls for digit-sized copies (~24%).** The prefix/digit/parts copies used `ptr::copy_nonoverlapping` with runtime lengths, which LLVM emits as `_platform_memmove` calls — for 3-byte copies. New `copy_bytes_small` does overlapping-window chunk copies below 16 bytes (u64/u32/u16 head+tail windows). A plain byte loop does NOT work here: LLVM's loop-idiom pass recognises it and re-emits the very `memcpy` call being avoided — the intermediate build measured only −14% and its `sample` still showed `_platform_memmove` under the "inlined" loop.

3. **`bzero` for the alignment pad (~4%).** `zero_alignment_padding_tail` memset at most 7 bytes through the PLT. A pad of ≤8 on a payload ≥8 is now one unaligned 8-byte zero store over the allocation tail (it may reach backward into the payload's last bytes, which are uninitialized until the caller writes them — same visible state).

5. **`js_string_append` rooted both operands before choosing an arm** (second commit) — but the in-place arm (refcount==1, fits capacity: the amortized accumulator path) never allocates; the two `root_string_ptr` calls were ~37% of an `s += "ab"` loop. The scope+roots now live in the growth arm only, next to the allocation they protect, and the in-place suffix copy uses the same chunk-copy helper (2-byte appends were one memmove PLT call each). Append differential vs node (unique growth, alias preservation, empty-dest reuse, cross-append surrogate re-pairing #6728): identical.

4. **memmove inside `fast_itoa_u32` (~15% after 1–3 landed).** The helper wrote digits at the buffer END and then `buf.copy_within(start..32, 0)` — a runtime-length overlapping copy, i.e. one libc `memmove` per conversion, inlined into both concat entry points (found by re-sampling the build with 1–3 applied). It now sizes first (`ilog10`) and writes digits in place; no copy at all.

Also folded `js_string_concat_value_box`'s SSO-arm prefix/digit copies (runtime-length `copy_nonoverlapping` → same helper); those were the remaining memmoves in the 1-2-digit SSO iterations.

## Soundness note

While staring at raw-pointer-vs-alloc here I verified the existing design fact this leans on: an alloc-point nursery trigger in moving mode **defers** the copying minor to the next declared safepoint (`gc/policy.rs`, phase 2/3 of the moving-GC project); a mid-expression collection is the conservative non-moving minor. A 2M-iteration young-operand concat churn under `PERRY_GC_PROTECT_FROMSPACE=1` (and separately under seeded schedule fuzzing, 142k copying minors) runs clean on main and on this branch. The `no_collect` arm doesn't rely on that global argument, only on its own local contract.

## Measurements

Mac mini (quiet host), 11 interleaved triples vs the #9114 branch, median ns/op (spread ≤0.2 on every row except lit+int's 2.8):

| shape | #9114 | this PR | node | Δ | vs node |
|---|---|---|---|---|---|
| `"id-" + (i & 255)` | 26.2 | **15.3** | 2.2 | **−41.6%** | 7.0× |
| concat-then-compare | 25.9 | **14.9** | 6.9 | **−42.5%** | 2.2× |
| `a + b` (two vars) | 25.3 | **20.9** | 0.5 | **−17.4%** | 41.8× |
| `` `id-${i & 255}` `` | 21.9 | **19.5** | 2.2 | **−11.0%** | 8.9× |
| `String(i & 255)` | 4.6 | 4.6 | 4.9 | +0.0% | 0.9× |
| `"id-" + "x"` | 8.1 | 8.1 | 0.5 | +0.0% | 16.2× |
| `s += "ab"` (grow) | 14.6 | **8.5** | 4.2 | **−41.8%** | 2.0× |

Cumulative vs main across the stack: `"id-" + i` 26.3→15.3 (−42%), template-int 54.6→19.5 (−64%), `String(smallint)` 35.4→4.6 (beats node), compare 26.1→14.9 (−43%).

## Correctness

- The 264-line number-formatting differential vs node (both PRs' corners: −0, ±1e-7, 1e21, 2^53, bigint/Symbol/object operands): **byte-identical**.
- string:: + builtins::numbers runtime filters: 90 passed.
- GC churn repros above clean under fromspace protection.
- Gate battery on the rebased stack (base = main with #9110 + #9116): `-D warnings` 0, codegen 1830/0, **full runtime suite 2819 passed / 0 failed (native, single-threaded — no overlay needed now)**, lints clean, integration issue_8655 2/2 + issue_8690 3/3 + issue_8897 3/3.

## Summary by CodeRabbit

  * Improved string concatenation and appending efficiency, particularly for short strings and numeric values.
  * Reduced unnecessary memory-management overhead during in-place string updates.
  * Optimized numeric-to-string conversion and internal string copying.
  * Improved handling of string padding during allocation.

* **Reliability**
  * Preserved string contents and behavior while streamlining allocation and copying paths.

### DIFF
--- a/crates/perry-runtime/src/string/append.rs
+++ b/crates/perry-runtime/src/string/append.rs
@@ -86,10 +86,6 @@ pub extern "C" fn js_string_append(
         }
     }
 
-    let scope = crate::gc::RuntimeHandleScope::new();
-    let dest_handle = scope.root_string_ptr(dest as *const StringHeader);
-    let src_handle = scope.root_string_ptr(src);
-
     unsafe {
         let dest_blen = (*dest).byte_len;
         let src_blen = (*src).byte_len;
@@ -117,10 +113,14 @@ pub extern "C" fn js_string_append(
         // In-place append optimization: if dest is uniquely owned (refcount==1)
         // and has enough capacity, append directly without allocation.
         // This turns O(n^2) string building loops into amortized O(n).
+        // No allocation happens in this arm, so it runs with no handle scope
+        // and no roots: the unconditional two-root setup used to cost ~37% of
+        // an `s += "ab"` accumulator loop (`sample`), all of it for the arm
+        // that cannot collect. The growth arm below opens its own scope.
         if (*dest).refcount == 1 && new_blen <= (*dest).capacity {
             let dest_data = (dest as *mut u8).add(std::mem::size_of::<StringHeader>());
             let src_data_ptr = string_data(src);
-            ptr::copy_nonoverlapping(
+            super::concat::copy_bytes_small(
                 src_data_ptr,
                 dest_data.add(dest_blen as usize),
                 src_blen as usize,
@@ -145,6 +145,9 @@ pub extern "C" fn js_string_append(
         // register that setjmp/stack-walk didn't capture). Fresh allocation
         // is safe: old string becomes garbage for the next GC cycle.
         let new_cap = (new_blen * 2).max(32);
+        let scope = crate::gc::RuntimeHandleScope::new();
+        let dest_handle = scope.root_string_ptr(dest as *const StringHeader);
+        let src_handle = scope.root_string_ptr(src);
         let new_ptr = js_string_from_bytes_with_capacity(ptr::null(), 0, new_cap);
         let dest = dest_handle.get_raw_mut_ptr::<StringHeader>();
         let src = src_handle.get_raw_const_ptr::<StringHeader>();
--- a/crates/perry-runtime/src/string/concat.rs
+++ b/crates/perry-runtime/src/string/concat.rs
@@ -151,6 +151,10 @@ pub(crate) unsafe fn copy_bytes_small(src: *const u8, dst: *mut u8, len: usize) 
         dst.cast::<u16>().write_unaligned(head);
         dst.add(len - 2).cast::<u16>().write_unaligned(tail);
     } else if len == 1 {
+        // GC_STORE_AUDIT(POINTER_FREE): string payload BYTES, not JSValues —
+        // this helper copies UTF-8 into freshly allocated storage, so no slot
+        // here can hold a heap edge and no barrier applies. The wider arms
+        // above do the same copy through `write_unaligned`.
         *dst = *src;
     }
 }
--- a/crates/perry-runtime/src/string/concat.rs
+++ b/crates/perry-runtime/src/string/concat.rs
@@ -126,7 +126,7 @@ fn bytes_all_ascii(data: *const u8, len: u32) -> bool {
 /// Same contract as `ptr::copy_nonoverlapping`: both regions valid for
 /// `len` bytes, non-overlapping.
 #[inline(always)]
-unsafe fn copy_bytes_small(src: *const u8, dst: *mut u8, len: usize) {
+pub(crate) unsafe fn copy_bytes_small(src: *const u8, dst: *mut u8, len: usize) {
     // Overlapping-window chunk copies, NOT a byte loop: LLVM's loop-idiom
     // pass recognises a plain byte loop and emits the very `memcpy` call
     // this helper exists to avoid (verified in `sample`: the loop version
--- a/crates/perry-runtime/src/string/concat.rs
+++ b/crates/perry-runtime/src/string/concat.rs
@@ -507,7 +507,11 @@ pub extern "C" fn js_string_concat_value(
                     let scope = crate::gc::RuntimeHandleScope::new();
                     let prefix_handle = scope.root_string_ptr(prefix);
                     let (ptr, data_ptr) = string_storage_alloc(total_blen as u32);
-                    (ptr, data_ptr, prefix_handle.get_raw_const_ptr::<StringHeader>())
+                    (
+                        ptr,
+                        data_ptr,
+                        prefix_handle.get_raw_const_ptr::<StringHeader>(),
+                    )
                 }
             };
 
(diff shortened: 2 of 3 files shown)


proggeramlug:
Added a second commit extending the same lever to `js_string_append`: rooting moves into the growth arm only (the in-place accumulator arm never allocates — the two unconditional roots were ~37% of an `s += "ab"` loop), and the in-place suffix copy uses the chunk-copy helper. Mini pairs: **grow-append 14.6→8.5 ns (−41.8%, now 2.0× node)**, all six other shapes exactly flat. Append differential (growth/alias/empty-dest-reuse/surrogate re-pairing #6728) identical vs node; string:: suite 83/0. Full gate battery rerunning on the updated branch — will post results.

proggeramlug:
Gate battery on the append commit: `-D warnings` 0, codegen 1830/0, **full runtime suite 2819/0 native**, lints clean, integration 8655 2/2 / 8690 3/3 / 8897 3/3. #9118 is complete and ready for review.

proggeramlug:
Third commit, same lever family: `concat_byte_parts` (the two-string heap tail) now allocates no-collect-first and reuses the SSO check's `both_ascii` instead of re-walking both payloads for utf16. **`a + b` 22.1→17.2 ns (−22%)**, other shapes flat. 81-combo unicode pair differential (multi-byte, lone surrogates, empty operands) + the 264-line number differential: byte-identical vs node; string:: 83/0. Cumulative on this PR vs its base: `"id-"+i` −42%, compare −43%, grow-append −42%, var+var −34% (25.3→17.2 dev-box scale).

proggeramlug:
Gate battery on the third commit: `-D warnings` 0, codegen 1830/0, full runtime suite 2819/0 native, lints clean, integration 8655 2/2 / 8690 3/3 / 8897 3/3. Mini pairs for the commit: var+var −20.6% (20.9→16.6 ns), template-int −10.3% bonus (shared tail), all other shapes flat.

proggeramlug:
Merged, with two commits added (rustfmt, and a GC store-audit marker — below).

**The rooting argument is the load-bearing claim here**, so that's where I spent the effort. Removing the unconditional `RuntimeHandleScope` + prefix root and relying on `string_storage_alloc_no_collect`'s `Some` contract ("the open nursery block served this, nothing on the heap moved") is exactly the kind of reasoning that is either airtight or produces an intermittent from-space fault months later.

Exercised against the #7154 instruments — a churn workload (interleaved concat with object allocation, a 3000-iteration growth chain, freshly-allocated prefixes, a user-`toString` slow arm allocating inside, and a 5000-char prefix to force the block-boundary `None` fallback):

```
PERRY_GC_SCHEDULE_SEED=<s> RATE=1 ALLOC_KB=0 FORCE_EVACUATE=1
VERIFY_EVACUATION=1 PROTECT_FROMSPACE=1 PROTECT_FROMSPACE_DEPTH=800
```

Three seeds, **exit 0 and node-identical on all three, 76,269 objects moved each**. The quarantine and the evacuation verifier both stayed silent, which is the result that makes the no-root claim credible rather than merely plausible — a stale raw prefix read would have faulted precisely there.

Also 27 number→string coercion shapes byte-identical (carried over from #9114's probe: `-0` through six paths, the i32/u32 boundaries, `1e21`/`1e-7`/`MAX_SAFE_INTEGER`, `toString(radix)`, `valueOf`/`toString` precedence, BigInt).

**Performance**, interleaved best-of-3, 3M iterations:

| | main | this PR | node | |
|---|---|---|---|---|
| `s += "ab"` | 44 ms | **17 ms** | 104 ms | **2.59x** |
| `"id-" + i` | 122 ms | **71 ms** | 61 ms | 1.72x |
| `"k" + i + "-v"` | 147 ms | 142 ms | 50 ms | 1.04x |

The append row now beats node by 6x. The three-operand pairwise row barely moves, which tracks — it isn't the shape the four costs were attributed on.

**The added marker.** `gc_store_site_inventory.py` flagged `*dst = *src` at `concat.rs:154`. I classified it `POINTER_FREE`: the helper copies UTF-8 payload bytes into freshly allocated string storage, so no slot there can hold a heap edge — and the wider arms immediately above do the identical copy through `write_unaligned`, which the scanner doesn't match. Worth noting as a scanner blind spot: only the plain-deref form is detected, so the same store expressed four other ways in the same function goes unflagged.

**One thing I chased and cleared, so you don't.** My first dev-profile full-suite run on this branch failed `async_hooks::test_support::tests::native_async_resource_accepts_string_and_symbol_expandos` (exit 101). It passes isolated 3/3, and two subsequent full-suite runs on this same branch were clean (2819, exit 0). Main was clean too. So it is a pre-existing order/state flake, not yours — but I'd have wrongly hung it on this PR if I'd stopped at the first red.

Validation: runtime 2819 passed under **both** profiles (dev-profile exit 0, 0 abort markers), perry --bins 1066, fmt clean, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped. Rebased onto main — the first commit was #9114, already merged, so I skipped it; `git diff origin/main --diff-filter=D` is empty and `fast_itoa_u32` is intact.

Issue #8809, opened 2026-08-25
gc-root-dominance: 7/3 gated (8/4 unfiltered) NEW root-dominance regressions landed 2026-08-15..25 while the gate was red


proggeramlug:
`scripts/gc_root_dominance_check.py` reports **7 violations across 4 distinct fingerprints** on clean `main` (`db821b5fd`), against an allowlist whose `entries` list is empty. CLAUDE.md states every new hit is a red build, so this is either a real regression or the gate is currently red on `main`.

```
test_gap_array_iterator_manual_next…::main::js_closure_alloc->js_object_set_field_by_name
test_gap_gc_method_receiver_rooting…::perry_fn_…__make::js_closure_alloc->js_object_set_field_by_name
test_gap_object_create_method_this…::main::js_closure_alloc->js_object_set_field_by_name
test_gap_class_advanced…::main::js_object_alloc_class_inline_keys_stamped->js_private_brand_add
```

Shape of the first three (all identical):

```
alloc  : %rN = call i64 @js_closure_alloc(...)
store  : store i64 %rM, ptr %r19
bind   : slot 9  call void @js_shadow_slot_bind(i32 9, ptr %r19)
between: js_object_set_field_by_name
MOVING : YES via js_object_set_field_by_name
```

i.e. the #7192 shape — the root store is emitted in-frame but **after** a call that can allocate.

### Reproduce

```
cargo build --profile perry-dev -p perry -p perry-runtime-static -p perry-stdlib-static
PERRY_BIN=<target>/perry-dev/perry PERRY_RUNTIME_DIR=<target>/perry-dev \
  ./scripts/gc_root_dominance_corpus.sh /tmp/rdcorpus
python3 scripts/gc_root_dominance_check.py /tmp/rdcorpus; echo $?    # -> 1
```

### Caveat on the measurement

Measured with a **`perry-dev`** binary; `gc-root-dominance.yml` uses release, and it passes flags this invocation did not (`--moving-only`, `--min-binds`, `--min-funcs`, plus its env knobs). So the absolute count may not match CI's. Worth confirming at release profile before treating the count as authoritative — but the violations are real IR at the profile measured.

Found while auditing #8783; an identical A/B showed that PR introduces none of these.

proggeramlug:
**Correcting the counts in this issue, and re-dating the finding.**

Two corrections to my original report.

### 1. "7 violations across 4 fingerprints" is a pairing that never occurs

The count depends on which filter you run, and I mixed them:

[...]

PR #9122, opened 2026-08-30
perf(codegen): shape-cache path for object literals with captures_this methods

proggeramlug:
## What

A `captures_this` method used to force the WHOLE object literal onto the by-name path (all-or-nothing gate at the shape-cache fast path): `this_patches` needs each method closure value after every later initializer has run, and the by-name arm keeps those values alive through nested rooted accumulators. But the shape allocator pre-populates `keys_array`, so by-INDEX stores are equally valid for method props — and the patch loop can simply **re-read each method closure from its own field slot at patch time** (`js_object_get_field`, declared for IR here): the rooted object keeps the closure alive, and the re-read observes any evacuation that moved it, with no per-closure rooting at all.

Kill switch: `PERRY_OBJECT_LITERAL_SHAPE_METHODS=0` restores the old routing (read per literal, uncached, so the tests below can pin either arm).

## Measurements

Mac mini kill-switch pairs (7 runs, medians; node for scale):

| shape | by-name (off) | shape path (on) | node | Δ |
|---|---|---|---|---|
| `{a, b, inc()}` literal build+call | 853.6 ns | **426.2 ns** | 3.5 | **−50.1%** |
| 5-prop 2-method literal | 1585.5 | **765.6** | 5.9 | **−51.7%** |
| plain 3-prop literal | 1.0 | 1.0 | 0.8 | +0.0% |

The residual (~257 ns build-only, ~63 ns per method call) is per-instance closure allocation and the dynamic method-call family — separate levers, sized in the follow-up investigation ([the fn_param receiver finding](#) is filed with the team).

## Correctness

- Differential vs node — methods + `this`-binding, alias preservation, methods-only literals, method-first literals, nested literal methods, `JSON.stringify` round-trip, and a **200k-literal churn loop under allocation pressure** (evacuating minors between build and patch): byte-identical, both switch states.
- The five `by_name_method_closure_tests` (the #8809 rooting-hazard pins) now pin their arm explicitly via the kill switch under a module-local mutex, and three new `shape_method_literal_tests` twins pin the shape path's invariants: patches below every store, **each patch re-reads its closure from its field slot** (a kept register is stale after an evacuating initializer), and source-order patching.
- Gates: `-D warnings` 0, codegen **1833/0**, full runtime suite 2819/0 (runtime untouched by this change; suite run on the same tree), lints clean, integration 8655 2/2 / 8690 3/3 / 8897 3/3.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved object literal handling for methods that capture `this`, enabling safer and more consistent property initialization.
  * Added reliable field-based access during object creation to ensure captured method closures receive the correct `this` context.

* **Tests**
  * Added coverage for shape-based object literals and expanded validation of alternate routing behavior.

### DIFF
--- a/crates/perry-codegen/src/expr/object_literal.rs
+++ b/crates/perry-codegen/src/expr/object_literal.rs
@@ -392,6 +392,17 @@ pub(crate) fn lower_object_literal(
     // need the by-name path because `this_patches` populates them post-build
     // via `js_closure_set_capture_bits`, which assumes the key is already in
     // keys_array — fine here since the shape allocator pre-populates it.
+    // A `captures_this` method used to force the WHOLE literal onto the
+    // by-name path (all-or-nothing gate): `this_patches` needs the closure
+    // value after every later initializer has run, and the by-name arm keeps
+    // it alive through nested rooted accumulators. But the shape allocator
+    // pre-populates keys_array, so by-INDEX stores are just as valid for
+    // method props — and the patch loop can simply RE-READ each method
+    // closure from its own field slot at patch time (`js_object_get_field`):
+    // the rooted object keeps the closure alive, and the re-read observes
+    // any evacuation that moved it, with no per-closure rooting at all. So
+    // the gate is now only a kill switch (PERRY_OBJECT_LITERAL_SHAPE_METHODS=0
+    // restores the old routing for A/B).
     let any_method_closure = !generator_iterator_object
         && props.iter().any(|(_, v)| {
             matches!(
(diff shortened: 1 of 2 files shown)


proggeramlug:

**I added one commit**: `PERRY_OBJECT_LITERAL_SHAPE_METHODS` was not registered in `BUILD_CACHE_ENV_VARS`, so `codegen_env_vars_are_build_cache_inputs` would have gone red. It gates two different literal-birth emission sequences, so it is a cache key. That gate has now caught five PRs in this series — worth adding the registration to the checklist when introducing a codegen knob.

PR #9125, opened 2026-08-30
fix(runtime): coerce class refs through ToPrimitive

### LINKED ISSUE #9101: Class refs bypass valueOf/toString in looped string +=, in * and -, and ignore Symbol.toPrimitive
#9097 made `+` run the ordinary `valueOf`/`toString` sequence on registered class refs. Four adjacent coercion paths still bypass it. Found while auditing that PR; all four are identical on main and on #9097, so none is introduced there.

Measured against node v26.5.1, `main` → `#9097` in each row.

### 1. Looped string `+=` still misses the override

```ts
class WithToString { static toString() { return "CUSTOM"; } }
let s = ""; for (let i = 0; i < 3; i++) s += WithToString;
console.log(s);
```

```
node     CUSTOMCUSTOMCUSTOM
perry    function WithToString() { [native code] }…
```

Worth calling out because #9097 explicitly targets the looped `+=` reproducer and *does* fix the numeric one — `let n = 0; n += WithValueOf` now gives `126`, matching node. Only the string-accumulator form is still on the old path, so this is likely a narrow gap in the same area rather than a separate mechanism.

### 2. `*` and `-` do not coerce class refs

```ts
class WithValueOf { static valueOf() { return 42; } }
class WithBoth { static valueOf() { return 7; } static toString() { return "TS"; } }
console.log(WithValueOf * 2, WithBoth - 1);
```

```
node     84 6
perry     6 4
```

`+` is fixed; the other arithmetic operators still see the raw ref.

### 3. Static `Symbol.toPrimitive` is ignored

```ts
class WithPrimitive { static [Symbol.toPrimitive](h: string) { return h === "number" ? 99 : "PRIM"; } }
console.log(WithPrimitive + "");
```

```
node     PRIM
perry    function WithPrimitive() { [native code] }…
```

`Symbol.toPrimitive` takes precedence over `valueOf`/`toString` in `ToPrimitive`, so it should be consulted first by the same path #9097 added.

### 4. Class source text is not retained

```ts
class Plain { }
console.log("" + Plain);            // node: "class Plain { }"   perry: "function Plain() { [native code] }"
console.log(String(Plain) === Plain.toString());   // node: true   perry: false
console.log(("" + class D extends B { }).includes("class"));  // node: true  perry: false
```

Distinct from 1–3 and probably much larger — it needs the source span retained through lowering rather than a coercion fix. Recording it here so the relationship is visible; happy for it to be split out if it is tracked elsewhere.

### Coverage note

Rows correct on `main` → `#9097`: 5/27 → 14/27 across the shapes I probed (both operand orders, `valueOf`-only, `toString`-only, both, `Symbol.toPrimitive`, arithmetic operators, looped `+=` in string and numeric form, class expressions named and anonymous, `extends` with an inherited static `toString`, `Array.join`, computed keys, and `==`/`String()` agreement). Items 1–3 are all reachable from that same probe if someone wants a regression fixture.

proggeramlug:
Fixes #9101

## Summary

- preserve static `Symbol.toPrimitive` methods during class lowering and expose them as bound constructor methods
- route INT32-tagged class refs through class-aware `ToPrimitive` for default, number, and string hints
- cover looped string `+=`, numeric operators, direct `String()`, hook precedence, and named/class-expression forms

The separate class-source-text retention note in #9101 remains out of scope; this fixes the three coercion defects named in the issue title. No version bump is included.

## Testing

- `cargo test -p perry --test issue_9101_class_ref_coercion -- --nocapture`
- `cargo test -p perry --test issue_9087_class_ref_add --test class_expr_well_known_symbol_methods -- --nocapture`
- `cargo test -p perry-runtime dynamic_arith --lib -- --nocapture`

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed type coercion for class constructors used in arithmetic, string concatenation, unary operations, and `String()`.
  - Class constructors now correctly honor `Symbol.toPrimitive`, `valueOf`, and `toString` hooks.
  - Improved handling of numeric and string conversion hints.
  - Static and instance `Symbol.toPrimitive` methods are now resolved and invoked correctly.
- **Tests**
  - Added regression coverage for constructor coercion and custom conversion behavior.

### DIFF
--- a/crates/perry-codegen/src/expr/object_literal.rs
+++ b/crates/perry-codegen/src/expr/object_literal.rs
@@ -665,7 +665,10 @@ mod by_name_method_closure_tests {
             .unwrap_or_else(|poisoned| poisoned.into_inner());
         let prev = std::env::var_os("PERRY_OBJECT_LITERAL_SHAPE_METHODS");
         std::env::set_var("PERRY_OBJECT_LITERAL_SHAPE_METHODS", value);
-        RoutingPin { prev, _guard: guard }
+        RoutingPin {
+            prev,
+            _guard: guard,
+        }
     }
 
     pub(super) fn method_literal_ir() -> String {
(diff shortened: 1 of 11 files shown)


proggeramlug:
Merged. Closes #9101, which I filed after auditing #9097 — and it fixes the substantial half of it.

Measured on my #9097 probe (27 class-coercion shapes) against node v26.5.1:

| | rows wrong |
|---|---|
| main before #9097 | 22/27 |
| after #9097 | 13/27 |
| **after this PR** | **8/27** |

So this closes five more rows: static `valueOf`/`toString` through `*` and `-`, static `Symbol.toPrimitive` (which correctly takes precedence over both), and the looped string `+=` case that #9097's numeric fix had left behind. Routing ClassRefs through a real `ToPrimitive` with hint-mandated ordering, rather than extending the `+`-specific path again, is why it picks up all three at once.

The remaining 8 are the class **source text** gap — `"" + Plain` gives `function Plain() { [native code] }` where node gives `class Plain { }` — which is a lowering-level problem (the source span isn't retained), not a coercion one. That's item 4 in #9101 and wants its own change.

**I added one commit.** The PR pushed `to_string.rs` to **2003 lines**, three over the 2000-line cap, so `check_file_size.sh` was red. I extracted the ClassRef ToPrimitive group into a sibling `to_string_class_ref.rs` and re-exported `class_ref_to_primitive` from `to_string`, so the existing `crate::value::to_string::class_ref_to_primitive` call sites in `builtins/numbers.rs` and `symbol/iterator.rs` still resolve unchanged.

Worth flagging what that split then tripped, since it will catch the next person who splits a runtime file: a **new module starts with a raw-handle ceiling of zero**, so relocating one pre-existing `get_raw_const_ptr` into it made `raw_handle_debt.py` red even though no new code was written. I converted it to a genuine `with_const_ptr` scope (the tag-combine inside allocates nothing) rather than adding a ceiling.

This landed as part of a six-PR merge train, and had to go **after #9122** — the two conflict in `expr_object.rs`, so a bare cherry-pick onto main fails. Worth knowing if either gets reverted.

Validation on the final rebased tree: runtime 2822 passed (exit 0, 0 abort markers), fmt clean, `check_file_size.sh` PASS, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped.

PR #9126, opened 2026-08-30
fix(hir): preserve class self brand through private updates

### LINKED ISSUE #9104: Named class expression: inner name unresolvable from a static method, and self-binding not captured into nested arrows
Two shapes of a named class expression's self-binding remain broken after #9099, which fixed the per-evaluation identity problem (#9089) that hid them. Both are wrong on main too, so neither is a regression from that PR — but #9099 is what makes them worth separating out, since everything around them now works.

Measured against node v26.5.1.

### 1. The inner name is not visible from a static method when the class is also bound outside

```ts
const C = class Named { static f() { return Named === C; } };
console.log(C.f());
```

```
node     true
perry    ReferenceError
```

Unchanged by #9099 (main throws the same `ReferenceError`). Note the closely-related shapes now DO work — `class c { static self() { return c; } }` returns the right per-evaluation object, and `c.name` resolves — so this looks specific to referencing the inner binding by name from a static method body rather than to the self-binding machinery generally.

### 2. An arrow closure inside a static method cannot reach the static private

```ts
const make = () => class c {
  static #v = 0;
  static f() { return (() => { c.#v++; return c.#v; })(); }
};
const A = make(), B = make();
console.log([A.f(), A.f(), B.f()]);
```

```
node     [1, 2, 1]
perry    TypeError
```

Worth noting the *direction* changed here even though both are wrong: main returned `[undefined, undefined, undefined]` silently, and after #9099 it throws a `TypeError` instead. Throwing is the better failure — it is loud rather than silently producing wrong data — but it is still not node's answer, and code that happened to tolerate the `undefined` will now break. The non-arrow equivalent (`static f() { c.#v++; return c.#v; }`) is correct after #9099, so the gap is the capture of the self-binding into a nested arrow.

### Regression guard

These are the only 2 of 23 shapes still failing; the other 21 match node exactly after #9099, including two evaluations keeping separate static private state, `A.peek.call(B)` staying lexical, compound `++`/`--`/`+=` through the self-binding, a nested class closing over the outer evaluation, the lru-cache construction gate, foreign-class brand rejection, and `Outer.#o` by name. Whoever fixes these should keep that set green.

proggeramlug:
## Summary

- Preserve a named class expression's lexical brand owner through both guards emitted for a static private read-modify-write update.
- Add HIR and end-to-end regressions for static-method inner-name resolution, nested-arrow self capture, and per-evaluation static private state.

Current upstream already resolves the issue's `const C = class Named` shape. This fixes the remaining nested-arrow/private-update failure and keeps both reported shapes gated.

Fixes #9104

## Test plan

- [x] `cargo test --release -p perry-hir --lib` (361 passed, 1 ignored)
- [x] `PERRY_SKIP_BUILD=1 ./run_parity_tests.sh --filter 9104` (1/1 passed)
- [x] `PERRY_SKIP_BUILD=1 ./run_parity_tests.sh --filter class_expr` (19/19 passed)

## Checklist

- [x] No workspace version bump or `CLAUDE.md` / `CHANGELOG.md` edit
- [x] Added the required `changelog.d` fragment

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed static private-field updates in nested arrow functions within named class expressions.
  - Preserved the correct class identity when reading and updating private static fields.
  - Ensured separate class evaluations retain independent private state.

- **Tests**
  - Added coverage for named class expressions, static self-capture, nested arrows, and private-field updates.

### DIFF
--- a/changelog.d/9104-named-class-static-arrow.md
+++ b/changelog.d/9104-named-class-static-arrow.md
@@ -1,0 +1,4 @@
+Fixed named class expressions whose static methods update a static private
+field through the class's inner name from inside a nested arrow. The write side
+of `c.#field++` now preserves the same lexical class-evaluation brand as the
+read side instead of falling back to the arrow's absent `this` binding.
--- a/crates/perry-hir/src/lower/expr_member/private_guard.rs
+++ b/crates/perry-hir/src/lower/expr_member/private_guard.rs
@@ -31,6 +31,14 @@ pub(crate) fn private_storage_property(ctx: &LoweringContext, field_name: &str) 
 }
 
 pub(crate) fn is_class_expr_self_binding(ctx: &LoweringContext, object: &Expr) -> bool {
+    let object = match object {
+        // A private update (`c.#v++`) applies a read guard and then a write
+        // guard to the same receiver. Preserve the lexical-brand-owner bit
+        // through the inner guard so the outer guard does not fall back to the
+        // surrounding function's `this` (which is absent in a nested arrow).
+        Expr::PrivateGuard { object, .. } => object.as_ref(),
+        other => other,
+    };
     matches!(
         object,
         Expr::LocalGet(id)
--- a/crates/perry-hir/src/lower/tests.rs
+++ b/crates/perry-hir/src/lower/tests.rs
@@ -1102,6 +1102,41 @@ fn named_class_expr_self_new_records_appended_capture_provenance() {
     );
 }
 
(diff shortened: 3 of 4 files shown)


proggeramlug:

Behaviour spot-check: my #9104 probe (23 class-self-binding shapes) is unchanged at 2/23 wrong, both of which are the pre-existing `Named === C` and nested-arrow-capture gaps — so this fixes its target without disturbing that surface.

PR #9127, opened 2026-08-30
fix(runtime): honor collection iterator return overrides

### LINKED ISSUE #9098: User-installed .return on a Map/Set iterator is never invoked by IteratorClose
A user-installed `return` method on a Map/Set iterator is readable as a function but is never invoked, and calling it directly yields `undefined` instead of the user's result.

```ts
const it: any = new Set([1, 2]).values();
it.return = () => ({ value: 42, done: true });
console.log(typeof it.return, it.return().value);
```

```
node v26.5.1  function 42
perry         function undefined
```

The protocol consequence is the one that matters — `IteratorClose` never reaches the override on an abrupt loop exit:

```ts
const it: any = new Set([1, 2, 3]).values();
let closed = 0;
it.return = function () { closed++; return { done: true }; };
const out: number[] = [];
for (const v of it) { out.push(v); if (v === 2) break; }
console.log(out, closed);
```

```
node v26.5.1  [1, 2] 1
perry         [1, 2] 0
```

Iteration itself is correct in both — only the close callback is skipped.

### Not introduced by #9093

I found this auditing #9093 and A/B'd it: main and that PR produce identical output on both cases (`["function", undefined]` and `[[1,2], 0]`). #9093 removes the *synthetic* `return`/`throw` from collection iterators, which is a separate and correct change; this is the pre-existing gap underneath it that I flagged as a risk when filing #9086 ("that needs a check for callers relying on the current behaviour"). Now confirmed as its own defect rather than a consequence.

### Note on the `typeof`

`typeof it.return === "function"` is right for node, but here it is right by accident: something callable is being returned that is not the user's function, since invoking it gives `undefined`. Whatever resolves the read is not finding the own data field that `map_set_iterator_property`'s `own_data_field_by_name` prefix should have hit — worth starting there.

### Suggested coverage

An `IteratorClose` test is more valuable than a `typeof` test, because the read can look right while the call path is wrong — which is exactly what happens here. Cover: early `break`, a `throw` out of the loop body, `return` from inside the loop, and destructuring with fewer targets than elements (`const [a, b] = set.values()`), asserting the override runs exactly once in each.

proggeramlug:
## Summary

Makes user-installed `return` methods on Map and Set iterators callable through both direct method syntax and `IteratorClose`.

## Changes

- Restricts the collection-iterator class-id call fast path to actual intrinsic methods (`next` and iterator identity).
- Lets `return`, `throw`, and other non-intrinsic names continue through ordinary own/prototype method resolution.
- Removes the remaining synthetic Map/Set iterator `return`/`throw` results from the intrinsic dispatcher.
- Adds executable coverage for direct own and inherited calls, absent methods, early `break`, escaping `throw`, function `return`, and partial destructuring.

## Related issue

Fixes #9098

## Test plan

- [x] `cargo build --release` clean on `root@perrymaster.skelpo.net`
- [ ] `cargo test --workspace --exclude perry-ui-ios --exclude perry-ui-tvos --exclude perry-ui-watchos --exclude perry-ui-gtk4 --exclude perry-ui-android --exclude perry-ui-windows` passes (not run; repository guidance uses affected-crate scopes)
- [x] `cargo test -p perry --test issue_9098_collection_iterator_close -- --nocapture`
- [x] `cargo test -p perry --test issue_9086_collection_iterator_methods -- --nocapture`
- [x] `cargo test --lib -p perry-runtime` (2798 passed, 4 ignored)
- [x] `RUST_MIN_STACK=33554432 RUST_TEST_THREADS=1 cargo test --bins -p perry` (1056 passed; the host default stack overflows in an unrelated CJS preamble canary)
- [x] Affected bin/lib scopes for `perry-ffi`, `perry-ui`, and `perry-updater`
- [x] Added an executable regression test in the affected crate.
- [x] Documentation is not needed; this restores existing ECMAScript behavior without adding an API.
- [ ] Platform UI backend build is not applicable.

## Screenshots / output

Before the fix, direct calls returned `undefined` and `break` / `throw` / function `return` left the close counter at 0. The regression now observes the user result for direct calls and exactly one close for every abrupt-completion case; partial destructuring remains exactly once.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Map and Set iterator `return` methods so custom user-defined overrides are now honored.
  * Improved iterator cleanup when exiting loops, handling exceptions, returning from functions, or partially destructuring iterators.
  * Corrected behavior for direct and inherited iterator method calls, including cases where no `return` method exists.

### DIFF
--- a/changelog.d/9127-collection-iterator-close.md
+++ b/changelog.d/9127-collection-iterator-close.md
@@ -1,0 +1,6 @@
+### Fixed
+
+- User-installed `return` methods on Map and Set iterator objects now run for
+  direct calls and `IteratorClose` on `break`, escaping `throw`, function
+  `return`, and partial destructuring instead of being bypassed by the native
+  class-id dispatcher (#9098).
--- a/crates/perry-runtime/src/collection_iter_object.rs
+++ b/crates/perry-runtime/src/collection_iter_object.rs
@@ -35,6 +35,15 @@ const KIND_KEYS: i32 = 1;
 const KIND_VALUES: i32 = 0;
 const KIND_ENTRIES: i32 = 2;
 
+/// Methods implemented intrinsically by the Map/Set iterator class-id
+/// dispatcher. `return` and `throw` deliberately are not here: ordinary
+/// collection iterators do not define them, and a user-installed own or
+/// inherited method must flow through ordinary method lookup (#9098).
+#[inline]
+pub(crate) fn is_intrinsic_iterator_method(method_name: &str) -> bool {
+    matches!(method_name, "next" | "Symbol.iterator" | "@@iterator")
+}
+
 /// `true` when `addr` carries a Map iterator object's class id.
 pub fn is_map_iterator_addr(addr: usize) -> bool {
     iterator_class_id(addr) == Some(MAP_ITERATOR_CLASS_ID)
@@ -326,7 +335,6 @@ unsafe fn dispatch_map_iterator_method_emit(
             emit_iter_result(&scope, &iter_h, emit_cached, value, false)
         }
         "Symbol.iterator" | "@@iterator" => js_nanbox_pointer(iter_obj() as i64),
-        "return" | "throw" => make_iter_result(JSValue::undefined(), true),
         _ => f64::from_bits(TAG_UNDEFINED),
     }
 }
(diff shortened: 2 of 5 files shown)


proggeramlug:

This closes #9098, which I filed while auditing #9093 — a user-installed `.return` on a collection iterator being readable but never invoked by `IteratorClose`. Good to see it fixed at the source rather than papered over.

PR #9128, opened 2026-08-30
fix(codegen,runtime): closure-literal singletons broke function identity — pi boots

proggeramlug:
The last two of pi's four startup walls, and with them **pi boots natively for the first time** (`--version` prints `0.0.0`, node-identical; all four earlier walls were #9069/#9073/#9085 + these).

**Commit 1 — closure-literal singleton identity.** `expr/closure.rs` routed closure literals through `js_closure_alloc_singleton` / `js_closure_alloc_with_captures_singleton` keyed on (func_ptr, capture bits): two evaluations of the same literal with bit-identical captures — any captureless arrow, or an arrow capturing the same constant — came back `===`-equal. ECMA-262 OrdinaryFunctionCreate requires a fresh object per evaluation, observable via `===`, expandos, and WeakMap keys. pi hit it through graceful-fs's wrap pattern: `setPrototypeOf(wrapped, original)` received ONE object twice and the runtime correctly threw `Cyclic __proto__ value` — with obj_bits == proto_bits exactly, and none of the bundle's 17 textual setPrototypeOf sites firing a JS logging shim, which is what pointed below the JS boundary. Gap fixture `test_gap_9090_closure_literal_identity.ts`.

**Commit 2 — name-keyed builtin-member reads must see user overrides** (part 2 of the same boot sequence, `native_module.rs`). Gap fixture `test_gap_9091_native_member_patch_roundtrip.ts`.

**First pi measurements** (loaded dev box, so wall time is an upper bound): `--version` boots clean; `[gc-time]` share_permille=102 — **GC ≈ 10% of pi startup**, with the vacuity gate satisfied (copying minors ran, 124k-object nursery census) — first real data point for the concurrent-GC decision gate (threshold 15%).

Diagnosed and implemented by a subagent (runtime-side throw instrumentation → bundle-side call-site shims → identity-conflation hypothesis → smallest repro); validated by the coordinating session via the full pi boot.

## Summary by CodeRabbit

* **Bug Fixes**
  * Closure literals now create distinct function objects on each evaluation, preventing shared state and property leakage.
  * Updates to built-in namespace members are now reflected in subsequent name-based reads and module exports.
  * Preserved correct prototype behavior for patched built-in functions.

* **Tests**
  * Added coverage for closure identity, property isolation, and built-in namespace member patching.

### DIFF
--- a/crates/perry-codegen/src/expr/closure.rs
+++ b/crates/perry-codegen/src/expr/closure.rs
@@ -304,8 +304,7 @@ pub(crate) fn lower(ctx: &mut FnCtx<'_>, expr: &Expr) -> Result<String> {
             // per resume with the same per-activation box captures. User
             // arrows and function expressions always mint fresh objects.
             let is_plain_async_step = is_plain_async_step_body(body);
-            let singleton_identity_safe =
-                is_plain_async_step && (*is_arrow || captures_all_boxed);
+            let singleton_identity_safe = is_plain_async_step && (*is_arrow || captures_all_boxed);
             let no_capture_singleton = is_plain_async_step && *is_arrow && total_caps == 0;
             let captured_singleton =
                 singleton_identity_safe && !no_capture_singleton && !writes_unboxed_capture;
--- a/crates/perry-codegen/src/expr/closure.rs
+++ b/crates/perry-codegen/src/expr/closure.rs
@@ -266,13 +266,12 @@ pub(crate) fn lower(ctx: &mut FnCtx<'_>, expr: &Expr) -> Result<String> {
             // ran webhook_endpoints' method) — the `replace is not a function`
             // symptom.
             //
-            // To preserve the hot-path optimizations while restoring identity,
-            // arrow functions are singleton-eligible because they have no own
-            // `.prototype` and are not constructable. Compiler-synthesized
-            // non-arrow async callbacks whose captures are all boxes are also
-            // safe: they are never constructors and their cache key includes
-            // the box addresses. Other non-arrow closures are treated as
-            // potential constructors and always get a fresh instance.
+            // The Stripe fix restricted the caches to arrows (no own
+            // `.prototype`, not constructable) plus non-arrow all-boxed
+            // captures. pi's boot showed that is still not enough: identity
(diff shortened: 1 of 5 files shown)


proggeramlug:
Merged, plus a rustfmt commit. **pi boots natively** — worth stating on its own.

The singleton keying was a genuine spec violation, not just a nuisance: OrdinaryFunctionCreate makes a fresh function object per evaluation, so `mk() === mk()` must be `false` even for a captureless arrow. Keying on `(func_ptr, capture bits)` made any two evaluations with bit-identical captures collapse to one object, which breaks every identity-based pattern — listener removal, `WeakMap` keying, per-instance function properties.

Verified against node v26.5.1, 15 identity shapes, byte-identical:

| | shape | node |
|---|---|---|
| 1, 2, 3 | `mk() === mk()` for captureless arrow, loop-created arrows, `function` expression | `false` throughout |
| 4 | two closures capturing the *same constant* | `false`, both still return `5` |
| 5, 6 | three instances in a `Set` → size 3; two as distinct `WeakMap` keys | `3`, `[1,2]` |
| 7 | add/remove listeners by identity | `indexOf` finds the right one |
| 10, 11 | own properties are per-instance (`a.tag` set, `b.tag` undefined; disjoint `Object.keys`) | |
| 8, 9 | a stable reference is still `===`, and `o.m === o.m` | `true` — the fix doesn't over-correct |
| 13, 14 | per-iteration capture and parameterized captures still work | `[0,1,2]`, `[2,4,6]` |

Cases 8 and 9 were the ones I most wanted green: making every literal fresh must not make an *existing* reference compare unequal to itself.

One thing to flag for whoever measures next: this trades an allocation elision for spec correctness, so closure-heavy hot paths may show a regression. That's the right trade and I'm not asking for it back — but if a benchmark moves, this is the likely cause rather than a mystery.

Validation: codegen 1349 passed, runtime 2822 passed (exit 0), perry --bins 1066 passed, fmt clean, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped. Rebased onto main, `--diff-filter=D` empty.

(One runtime run in the middle showed a single unrelated failure that did not reproduce on re-run — same order/state flake I've hit twice today, not this PR.)

PR #9130, opened 2026-08-30
perf(codegen): inline field precheck for typed-receiver method calls (12.8→3.8 ns) + #9124 review fixes

proggeramlug:
Follow-up to #9124, which merged with the **method-direct probe only** (the second commit hadn't been pushed yet). This PR carries the second half of that lever plus the review fixes.

## What

1. **One inline field precheck vouches for all typed-receiver fields.** With the method probe in front, a probe-first `c.inc()` loop was ~80% `js_typed_feedback_class_field_get_guard`: per receiver field, per call, a `shape_descriptor_by_id` lookup (thread-local map) plus the raw-f64 layout contract, re-deriving facts the exact `(class_id, ShapeId)` pair already pins. The field-GET sites already emit `emit_class_field_inline_precheck` ahead of that guard (class/shape + not-forwarded + the per-**object** raw-f64 intact bit); because the intact bit is object-wide, one precheck vouches for every receiver field at once. Its miss edge runs the unchanged per-field runtime chain, whose i1 result joins at a phi. Same kill switch (`PERRY_METHOD_INLINE_PROBE=0`), same emission-off gating.
2. **Review fix (#9124 / CodeRabbit):** `emit_inline_direct_method_shape_guard` takes `accept_raw_ptr`. Probe-first sites pass `false` — a user NaN-box receiver must carry the `0x7FFD` tag before any dereference, so a plain double whose bits land in the heap range (a positive subnormal) misses to the runtime guard instead of reaching the header load. The pre-existing shape-only site and `versioned_indexed_loop.rs` keep raw acceptance (their receivers can legitimately arrive in the internal raw-address ABI). The other review point (receiver lowered before arguments, no re-read) is the tower's pre-existing structure shared by the runtime guard and the shape-only inline guard; all of them reject a moved receiver via the forwarded header bit and fall to the runtime guard, which resolves forwarding — the probe adds no new window, and re-rooting the whole tower is a separate change.
3. The `typed_f64_receiver` IR-shape test measured both proofs by the runtime guards' text positions; it now measures the proof itself (inline marker or guard call, whichever dominates), and the fields merge block is created after the precheck's blocks so the typed/generic branch follows the guard calls in emission order.

## Measurements

Mac mini, 7 kill-switch pairs, monomorphic `c.inc()` (node 0.4–0.5 in all four shapes):

| receiver / host | #9124 as merged (method probe) | **+ field precheck** | Δ | vs node |
|---|---|---|---|---|
| typed **param**, plain fn | 12.8 | **3.8** | **−70%** (118.6→3.8 cumulative) | 7.6× |
| **captured** const, arrow | 13.1 | **3.8** | **−71%** | 7.6× |
| local `new C()`, plain fn | 4.4 | 4.4 | 0 | 8.8× |
| local `new C()`, arrow | 4.4 | 4.4 | 0 | 11.0× |

With both prechecks the loop has **zero runtime calls per iteration** (`sample`: 100% inside the generated function). Semantics differential (subclass through base param, virtual override, mid-program prototype monkey-patch, own-property override, annotation-lie receiver) identical to the off arm after all fixes.

## Binary size

~80 B per typed-receiver method site (one precheck per site regardless of field count); classhost fixture `.text` 10,938,772 → 10,939,092 (+320 B, +0.003%) on top of #9124's +384 B. cc-bundle `size(1)` on/off pair is compiling on perrymaster and will be posted as a comment.

## Gates

Validated on the #9124 branch base: `-D warnings` 0, codegen 1830/0, runtime 2819/0, lints clean, integration 8655 2/2 / 8690 3/3 / 8897 3/3. Rerunning on this base (main with #9103) — results in a comment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved method dispatch reliability for typed numeric values by validating receiver types and fields more consistently.
  * Added safer handling for supported raw pointer receiver forms during optimized method checks.
  * Strengthened validation sequencing to reduce incorrect execution paths.

* **Tests**
  * Updated regression coverage for inline field validation and receiver guard sequencing.

### DIFF
--- a/crates/perry-codegen/src/lower_call/method_override.rs
+++ b/crates/perry-codegen/src/lower_call/method_override.rs
@@ -234,6 +234,12 @@ pub(crate) fn emit_inline_direct_method_shape_guard(
     method_guard_slot: &str,
     fast_label: &str,
     fallback_label: &str,
+    // `true` at sites whose receiver may arrive in the internal raw-address
+    // form (top word zero); `false` where the receiver is a user-visible
+    // NaN-box, so that a plain double whose bit pattern happens to land in
+    // the heap address range (a positive subnormal) can never reach the
+    // header load — it misses to the runtime guard instead.
+    accept_raw_ptr: bool,
 ) {
     let deref_idx = ctx.new_block("method_direct.inline_deref");
     let deref_label = ctx.block_label(deref_idx);
@@ -263,8 +269,12 @@ pub(crate) fn emit_inline_direct_method_shape_guard(
         // double-sized slot. `normalize_raw_object_addr` accepts exactly this
         // top-word-zero form; all other non-pointer NaN-box tags remain
         // rejected before dereference.
-        let is_raw_ptr = blk.icmp_eq(I64, &tag, "0");
-        let is_ptr = blk.or(I1, &is_tagged_ptr, &is_raw_ptr);
+        let is_ptr = if accept_raw_ptr {
+            let is_raw_ptr = blk.icmp_eq(I64, &tag, "0");
+            blk.or(I1, &is_tagged_ptr, &is_raw_ptr)
+        } else {
+            is_tagged_ptr
+        };
         let above_floor = blk.icmp_uge(I64, &recv_handle, &heap_floor);
         let below_ceiling = blk.icmp_ult(I64, &recv_handle, &heap_ceiling);
         let in_heap_range = blk.and(I1, &above_floor, &below_ceiling);
@@ -1045,6 +1055,7 @@ pub(super) fn emit_guarded_direct_method_call(
             &method_guard_slot_str,
             &fast_label,
             &fallback_label,
+            true,
         );
     }
     if probe_before_runtime_guard {
@@ -1058,6 +1069,7 @@ pub(super) fn emit_guarded_direct_method_call(
             &method_guard_slot_str,
             &fast_label,
             &runtime_guard_label,
+            false,
         );
         ctx.current_block = runtime_guard_idx;
     }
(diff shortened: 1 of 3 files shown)


proggeramlug:
Gates on this base (main @ 07dee0fd29 + #9103): `-D warnings` 0, codegen **1832/0**, lints clean (census, addr-class audit, file-size), integration issue_8655 2/2 / issue_8690 3/3 / issue_8897 3/3. Runtime is untouched by this PR (the full suite ran 2819/0 on the same codegen content earlier). The cc-bundle `size(1)` on/off pair is still compiling on perrymaster; it lands here as a comment.

proggeramlug:
Merged, plus a rustfmt commit. Follow-up to #9124, which I merged just before this.

Fast-tracked at the maintainer's request, so this got the targeted pass rather than the full treatment — recording exactly what that covered so the gap is visible:

- Rebased onto main (which now includes #9124), no conflicts, `git diff origin/main --diff-filter=D` empty.
- **No new env knob** — checked, since the previous four PRs in this area each added one unregistered and reddened `codegen_env_vars_are_build_cache_inputs`. This one reuses `PERRY_METHOD_INLINE_PROBE`, which I registered on #9124.
- codegen **1349 passed**, perry --bins **1066 passed**, both exit 0.
- `run_lint_gates.sh` **all 60 gates passed**; 2 CI-only skipped.
- fmt fixed (the `emit_class_field_inline_precheck` call wanted rewrapping).

What I did **not** run this time, and would normally: the 23-shape node differential for method dispatch, the IR A/B confirming the new field precheck actually fires, and a benchmark of the 12.8 → 3.8 ns claim. On #9124 that IR check was worth doing — my fixture there produced byte-identical IR, so the 1.00x I measured meant "never fired", not "no gain". The same caveat applies here: your numbers are unverified by me.

Given this PR also carries "#9124 review fixes", it's worth noting my #9124 review raised one substantive thing that is *not* addressed here and shouldn't be lost: monkey-patching a prototype is not picked up through a typed-parameter receiver — `run(o, n)` calling `o.m()` keeps the old body after `P.prototype.m = ...`, while a direct `p.m()` correctly sees the patch. That reproduces identically on main, so it predates both PRs, but it sits squarely in the code this series is changing. I'll file it.

proggeramlug:
**Binary-size row for the inline-probe work (#9124 + #9130), measured on the real bundle.**

perrymaster (x86-64), `cli_2.1.112.js` compiled twice from the same main-built compiler (`--no-auto-optimize --enable-wasm-runtime`, `PERRY_NO_CACHE=1`), only `PERRY_METHOD_INLINE_PROBE` differing:

| build | `.text` (size(1)) | Δ |
|---|---|---|
| probe off | 264,432,646 | — |
| probe on | 264,462,586 | **+29,940 B = +0.0113%** |

Per-site: the classhost fixture emits 6 method-direct probe blocks + 9 field-proof blocks for +832 B of `.text` (~139 B/site on x86-64), so the bundle delta corresponds to a few hundred admitted sites out of cli.js's ~17.8k class methods — the admission (typed receiver clone, non-shape-only, feedback emission off) is narrow, as designed.

Method note, since the first attempt at this measurement was vacuous: the earlier pair came out byte-identical because the compiler on that host predated the kill switch, so `on` == `off`. This pair is from a GitHub-fetched `main` (SHA `653e88669`) and the switch was verified at the IR level before the compile — `method_direct.inline_deref` blocks 6 → 0 and `fields_proven` 9 → 0 between the two settings on a fixture.

PR #9133, opened 2026-08-30
fix(compile): support full OpenCode source builds

proggeramlug:
## Summary

Make Perry compile and run OpenCode from its full TypeScript source and dependency graph, without relying on a minified JavaScript bundle. The raw graph exposed several independent lowering, source-graph, codegen, runtime, and Windows-link gaps; this PR fixes those gaps and adds focused regressions for the real dependency shapes.

The acceptance build compiled 7,136 native modules with 0 JavaScript fallbacks, then produced an OpenCode executable whose `--version` and `--help` commands both exited successfully.

## Changes

- Preserve ESM semantics across the source graph: hoist static import bindings, keep dynamic imports out of eager init ordering, erase all-type named specifier imports at runtime while retaining their metadata, and include that distinction in stable/object-cache hashes.
- Scope namespace-exported functions, variables, and classes by namespace in compiler registries. This prevents equal bare names from unrelated direct imports or other namespaces from replacing one another, including through named and `export *` barrels.
- Fix class and closure lowering used by OpenCode dependencies: forward module-level class-expression references, initialize inherited dynamic class expressions in the correct order, box shared optional captures at nested-closure entry, and statically recognize the CJS-wrapper `lru-cache` constructor destructure.
- Keep ambiguous `.values()` calls on generic iterator dispatch instead of specializing them as arrays; retain regression coverage for inherited iterator helpers on `MapIterator` (the implementation itself is now supplied by current `main`).
- Fix runtime class/object behavior used by Effect: preserve `Object.setPrototypeOf` links for ClassRefs, find inherited computed static fields for generic `in`, and allow clearing those links.
- Fix runtime correctness at dependency startup: root unresolved-global lookup names across lazy `globalThis` allocation, translate split surrogate ranges in regex character classes, and pass both user rest arguments and synthetic `arguments` arrays through bound method dispatch.
- Harden compilation and linking: make namespace ownership part of object-cache identity, limit `PERRY_DEBUG_INIT` invalidation to the entry object, fingerprint explicit object inputs without quadratic path scans, keep CJS class-hoist scanning synchronized across regex quotes, and rewrite duplicate COFF wrapper symbols with relocation-safe renames.

## Related issue

n/a — found while compiling OpenCode's full source dependency graph.

## Test plan

- [x] `cargo fmt -p perry-hir -- --check`, `cargo fmt -p perry-codegen -- --check`, `cargo fmt -p perry-runtime -- --check`, and `cargo fmt -p perry-transform -p perry -- --check`
- [x] `LLVM_SYS_221_PREFIX=C:\llvm cargo check -p perry-codegen -p perry-runtime -p perry`
- [x] `LLVM_SYS_221_PREFIX=C:\llvm cargo build --release -p perry-runtime-static -p perry-stdlib-static`
- [x] `cargo test -p perry-hir --lib` — 364 passed, 1 ignored
- [x] `cargo test -p perry-runtime --lib` — 2,752 passed, 4 ignored
- [x] All newly added focused compiler-driver and end-to-end regressions pass, including class-expression forwarding, ClassRef prototype inheritance, nested capture boxing, bound rest + `arguments`, namespace scoping, import/init ordering, type-only imports, regex translation, cache identity, CJS scanning, and COFF symbol rewriting.
- [x] Full OpenCode source/dependency compile — 7,136 native modules, 0 JavaScript fallbacks; generated executable passes `--version` and `--help`.
- [ ] `cargo build --release` clean — the affected release runtime and stdlib archives were built; the entire multi-platform workspace was not built locally.
- [ ] `cargo test --workspace --exclude perry-ui-ios --exclude perry-ui-tvos --exclude perry-ui-watchos --exclude perry-ui-gtk4 --exclude perry-ui-android --exclude perry-ui-windows` passes — not run. Broad Windows runs found unrelated/current-main failures: `perry` CLI tests reached 1,036 passed / 20 failed (POSIX command/path assumptions, permission tests, and cascading poisoned env locks); `perry-codegen` reached 1,345 passed / 3 failed in existing platform-sensitive object-byte/IR assertions; the source-graph suite reached 19 passed / 2 failed in existing trace assertions that do not recognize current `$pshape` output.
- [x] Added focused tests in the affected crates and `crates/perry/tests`.
- [x] No public CLI, stdlib, or runtime API documentation change is required.
- [x] No platform UI backend was touched.

## Screenshots / output

```text
Found 7136 module(s): 7136 native, 0 JavaScript
opencode --version  # exit 0: local
opencode --help     # exit 0
```

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility for full-source OpenCode builds, including namespace and type-only imports, class initialization, iterators, and Windows archives.
  * Fixed namespace export collisions and preserved behavior for same-named classes, functions, and variables.
  * Corrected class inheritance, static property checks, closure captures, rest parameters, and dynamic imports.
  * Prevented incorrect array-method handling on non-array values.
  * Improved garbage-collection safety and global lookups.
  * Fixed CommonJS class hoisting, surrogate-pair regular expressions, and static prototype inheritance.
  * Added clearer diagnostics when archive processing cannot complete.


proggeramlug:
## Audit + 3 mechanical fixes pushed (`15887807ed`)

Audited on macOS (arm64) against merge-base `07dee0fd29`. The compiler work here is solid — namespace scoping, the import pre-pass, the iterator narrowing and the init-order change all matched Node in every probe. Two defects and some CI debt; I pushed fixes for the CI debt and for the one that blocked this PR's own goal.

### Fixed in `15887807ed`

**1. macOS linking was broken by this PR.** `strip_duplicate_objects_from_well_known_lib` decided COFF from the archive's `.lib` extension, but it is handed perry's own intermediate, which `strip_duplicate_objects_from_no_shared_deps` names `_<lib>_nosharedeps.lib` on *every* host. So macOS/Linux took the Windows branch, `llvm-ar --format=coff` wrote a GNU symbol table, and `ld` rejected it:

```
ld: archive member '/' not a mach-o file in _libperry_ext_ws.a_nosharedeps.lib
```

Isolated repro: `llvm-ar --format=coff crs coff.lib x.o` then linking gives that exact error, while plain `ar crs` links fine. Now read from an extracted member's magic bytes (`strip_dedup/object_format.rs`). **With this fix the full OpenCode source graph links on macOS: 2,954 modules, 0 JavaScript fallbacks, 235 MB executable.** This is likely why the acceptance run was Windows-only.

**2. `perry-transform` lib tests did not compile** — 5 × `E0063 missing field runtime_erased`. This red-lined `cargo-test`/`warnings`, and fail-fast then cancelled `check`, `e2e-scoped`, `gc-stress-build` and `gap-suite-build`, so those four jobs never actually ran on this PR.

**3. Lint gate.** The size gate fails on **four** files, not one — `error.rs` (2037), `property_get.rs` (2005), `module_decl.rs` (2072), `cjs_wrap/tests.rs` (2014); all split (plus `strip_dedup.rs`, which my own fix pushed over). Also cleared the addr-class ratchet and recorded the `test_only` verdict for the new `TEST_COLLECT_BEFORE_GLOBAL_THIS_ALLOC` holder.

On the ratchet specifically — that one is not just style. `proto` is user-supplied and is stored into a GC root table the collector later dereferences; a bare `is_valid_obj_ptr` accepts the fetch/zlib/proxy handle bands on Linux (segfault there, hidden on macOS). Now guarded with `is_above_handle_band` + `try_read_gc_header`.

Verified: `cargo fmt --check`, `cargo check --all-targets`, size gate, addr-class audit and `gc_runtime_root_holders` all clean; perry-transform 121, perry-hir 364, perry-runtime 2,825, perry bin 1,068 tests pass.

### Still open — please take a look

**A. `Object.setPrototypeOf(Ctor, obj)` leaks onto instances.** It stores `obj` in `CLASS_PROTOTYPE_OBJECTS`, which carries *instance*-prototype meaning, so a constructor-side link becomes an instance-side one:

```js
const schema = { ast: "STATIC-ONLY", greet() { return "static-only"; } };
class Opaque {}
Object.setPrototypeOf(Opaque, schema);
const inst = new Opaque();
//                    node        this PR        merge-base
// inst.ast           undefined   STATIC-ONLY    undefined
// typeof inst.greet  undefined   function       undefined
// inst.greet()       TypeError   "static-only"  TypeError
```

Same root cause, second symptom — it mutates the user's object: after `Opaque.prototype.added = fn`, `Object.keys(schema)` is `["ast","added"]` (Node: `["ast"]`), because prototype-method mirroring writes through that table.

This hits the PR's own target: Effect does `Object.setPrototypeOf(Class, schema)` at `Schema.ts:1887` and `:5874`. The new test asserts only the static side, which is why it passed. Suggested fix: give the constructor `[[Prototype]]` its own table that instance walks don't read, and assert `new Ctor()` does **not** see the prototype's members.

Related, the fix is partial: `Ctor.staticMethod()` called directly on the class ref still throws `is not a function` (Node returns the value), though the same call via an any-typed alias works. `Object.getPrototypeOf(Ctor)` also still returns the wrong value (pre-existing).

**B. The linked binary does not start.** `--version` and `--help` both exit 1 with `TypeError: recur is not a function`. Caveat on attribution: my graph is 2,954 modules rather than the harness README's 2,454, because `patches/dependency-compat.patch` no longer applies to the locked Effect version (the `Schema.Opaque` hunk in `unstable/cluster/Envelope.ts` fails), so Effect's cluster/Optic/Config code is in my build and may be outside your intended profile. Worth confirming whether `--version` still passes for you, and refreshing that patch.

### Not caused by this PR

`source_graph_export_regressions` fails 2 tests (`imported_class_return_types_pull_in_transitive_dispatch_metadata`, `cross_module_arrow_callback_dispatch_is_resolved_once_and_fails_closed`) — I confirmed both fail identically on merge-base `07dee0fd29`. This PR takes that suite from 14 to 19 passing.

🤖 Audited and fixed with [Claude Code](https://claude.ai/code)

https://claude.ai/code/session_01P3bPE5eJQT4vQ6wf8P9JDN

proggeramlug:
## Update: `setPrototypeOf` regression fixed (`60cb4d9561`); one runtime gap left

### Fixed properly (not reverted)

The constructor's `[[Prototype]]` now lives in its own GC-rooted `CLASS_STATIC_PROTOTYPES` table instead of `CLASS_PROTOTYPE_OBJECTS` (which means "what INSTANCES inherit from"). It is visited by both the full and budgeted class-side-table root walks via a new `ClassSideTableRootSlot::StaticPrototype`, and the store fires the root write barrier — same treatment as `CLASS_DECL_PROTOTYPE_OBJECTS` beside it.

All of these now match Node:

| | before | now |
|---|---|---|
| `new Opaque().ast` | `"STATIC-ONLY"` | `undefined` |
| `inst.greet()` | ran | `TypeError` |
| `Object.keys(schema)` after `Ctor.prototype.x = fn` | polluted | untouched |
| `Opaque.make(1)` directly on the class ref | threw | `2` |
| `Object.getPrototypeOf(Opaque)` | wrong | `schema` |
| after `setPrototypeOf(Ctor, null)` | wrong | `null` |

The last two were gaps the original arm left open. `CLASS_STATIC_PROTOTYPE_NULLED` keeps "never linked" (→ `Function.prototype`) distinct from "explicitly severed" (→ `null`). Two regression tests assert Node's exact output, including that instances see nothing.

Verified: macOS — runtime 2825, codegen 1349, perry bin 1068, hir 364, transform 121, setPrototypeOf 5/5, namespace ABI 5/5; every lint gate green. Linux (`perrymaster`) — runtime 2804, transform 121, hir 364, perry bin 1051.

### `cargo-test` CI failure is pre-existing on main

Three tests stack-overflow under Linux debug. Ran both trees on the same host:

| test | main `07dee0fd29` | this PR |
|---|---|---|
| `collect_modules::…trusted_js_package…` | 💥 overflow | 💥 overflow |
| `preamble_canary_tests::path_module_wrap…` | 💥 overflow | 💥 overflow |
| `preamble_canary_tests::the_cjs_preamble…` | 💥 overflow | 💥 overflow |
| rest of `perry` bin | ✅ 1049 / 0 | ✅ 1051 / 0 |

Identical. They only surfaced here because fixing the `perry-transform` compile error let the job reach them. Worth a separate issue against main; this PR cannot go green on `cargo-test` until then.

### Still open: the built binary does not start

macOS, full source graph, 2954 modules / 0 JS fallbacks, links cleanly (235 MB), then:

```
$ opencode-source --version
TypeError: recur is not a function
```

Traced with `PERRY_DEBUG_INIT` (which this PR usefully makes cheap — 182 s to re-emit just the entry object):

- Fails during `effect/src/SchemaRepresentation.ts` init, at line 1683: `export const DocumentFromJson = Schema.toCodecJson($Document)`.
- Lands in `Schema.ts::toCodecJsonBase`, which does `ast.recur(recur)`. `recur` there is an **instance method** on the `SchemaAST` node classes.
- Instrumenting the call site gives the decisive fact:

```
DIAG tag=Objects typeof ast.recur=function   ← ok
DIAG tag=Union   typeof ast.recur=undefined  ← missing
```

Both `Objects` and `Union` extend `Base` in `SchemaAST.ts`; `Union` is the generic one (`class Union<A extends AST = AST> extends Base`). `SchemaAST.ts` is **not** split into multiple LLVM units, so it is not a split-unit artifact.

I could not reduce it: standalone repros of namespace-imported classes, generic classes, hoisted `function recur` used before declaration (called, passed as a value, and captured in an arrow), and namespace-member calls taking such a closure all match Node. The trigger needs more of Effect's real shape.

Two caveats on attribution: there is no A/B here, because the merge-base cannot compile this graph at all — this PR is what enables it. And my profile is 2954 modules rather than the harness README's 2454, because `patches/dependency-compat.patch` no longer applies to the locked Effect version (the `Schema.Opaque` hunk in `unstable/cluster/Envelope.ts` fails). That patch needs refreshing regardless — without it nobody can reproduce the acceptance run. Note `SchemaRepresentation.ts` is reachable from core `Schema.ts`, so the stub would not have removed the failing module.

So the PR's "`--version` and `--help` exit 0" line does not reproduce on macOS. Does it still hold for you on Windows with the intended profile?

🤖 Audited and fixed with [Claude Code](https://claude.ai/code)

https://claude.ai/code/session_01P3bPE5eJQT4vQ6wf8P9JDN

proggeramlug:
## OpenCode now runs — root cause was a class-id collision (`749491ff87`)

`--version` and `--help` both exit 0 on macOS, matching this PR's acceptance output for the first time:

```
$ opencode-source --version
local
$ opencode-source --help
█▀▀█ █▀▀█ █▀▀█ █▀▀▄ █▀▀▀ █▀▀█ █▀▀█ █▀▀█
Commands:
  opencode completion   generate shell completion script
  opencode mcp          manage MCP (Model Context Protocol) servers
  ...
```

2,954 modules, 0 JavaScript fallbacks, 235 MB binary.

### What `recur is not a function` actually was

Not a lowering bug, and not caused by this PR. **Class ids are allocated per module, so one module's anon-shape id can be the same number as another module's declared class.** Effect's monomorphized `Union$AST` collided with one. Once that module's init ran `js_register_anon_shape_class_id`, `is_anon_shape_class_id` began answering true for the declared class, and `js_object_get_prototype_of` skipped its declared-class fast path and fell through to `Object.prototype`.

The same class id, both states in one run:

```
[gp] cid=1121 anon_shape=false name=Some("Union$AST") methods=[..,"recur"]
[gp] cid=1121 anon_shape=true  name=Some("Union$AST") methods=[..,"recur"]
```

`SchemaAST.modifyOwnPropertyDescriptors` clones AST nodes with the standard prototype-preserving idiom:

```js
Object.create(Object.getPrototypeOf(ast), d)
```

so the cloned `Union` kept every own field and lost every method — instrumented at the real call site:

```
CLONELOSS tag=Union astRecur=function protoRecur=undefined outRecur=undefined
```

Instance method dispatch kept working the whole time (it resolves through the vtable), so only reflection saw the damage. And it only appears at scale: a 102-module graph never collides, 245+ does — which is why eight standalone reductions all passed.

**Fix:** a registered class name plus a non-empty prototype vtable is positive evidence of a real declared class (an anon shape has neither), so `declared_class_outranks_anon_shape` lets the declared class keep its reflective prototype through the collision. Two unit tests cover both directions.

**This does not fix the underlying per-module id collision**, which can still affect other id-keyed lookups. That deserves its own issue against main — I'd suggest globally unique class ids or a namespaced id space.

### Verification

| | |
|---|---|
| OpenCode | 2,954 modules / 0 JS fallbacks, links, `--version` + `--help` exit 0 |
| perry-runtime | 2,827 passed (incl. 2 new) |
| perry-codegen | 1,349 passed |
| perry bin | 1,068 passed |
| perry-hir / perry-transform | 364 / 121 passed |
| Lint | file size, addr-class, GC root holders, local-binding, store-site, sink isolation, unrooted-shape, pin sites, rekeyed tables, matrix liveness, `fmt --check` — all clean |

`source_graph_export_regressions` is 19 passed / 2 failed; those two are the pre-existing pair (14/2 on merge-base `07dee0fd29`).

### Known-red, not from this PR

Three tests stack-overflow under Linux debug — verified identical on merge-base and on this branch (`perrymaster`, same host, same profile): `collect_modules::…trusted_js_package…` and two `preamble_canary_tests`. With those skipped: main 1,049 passed / 0 failed, this branch 1,051 / 0. They keep `cargo-test` red regardless of this PR and want a separate issue.

Also: CI has produced **no check runs** for the last few head commits (Actions is healthy repo-wide; the previous push's runs were cancelled). Someone may need to re-trigger it.

Finally, `patches/dependency-compat.patch` no longer applies to the locked Effect version (the `Schema.Opaque` hunk in `unstable/cluster/Envelope.ts`), so my graph is 2,954 rather than the README's 2,454. Worth refreshing so the acceptance run is reproducible.

🤖 Audited and fixed with [Claude Code](https://claude.ai/code)

https://claude.ai/code/session_01P3bPE5eJQT4vQ6wf8P9JDN

Issue #7846, opened 2026-08-11
umbrella: a local binding's type/kind proof is not invalidated by assignment and is not validated against the runtime value (#7700, #7796, #7837, #7844)


proggeramlug:
## The invariant

> **A type/kind proof attached to a local binding is usable only at sites that every write to that binding dominates. A proof derived from a declared TypeScript type is not a proof at all without a runtime check.**

Codegen repeatedly attaches a kind/representation proof to a *binding* — from its declaration or its initializer — and then uses that proof to select a specialized lowering or constant-fold a predicate. The proof is sound where it was derived and unsound afterwards, because nothing invalidates it when the binding is written again, and nothing validates a *declared* type against the value actually present at runtime (CLAUDE.md, Known Limitations: "declared TS types aren't enforced at runtime").

Four independent bugs found within one day are all this one mistake. Filing an umbrella because patching one spelling at a time is precisely what let the same root ship four times.

## The four instances

| issue | state | binding proof | what it selected | symptom |
|---|---|---|---|---|
| #7700 | closed | "local is `Uint8Array`/`Buffer`", and the key node is not a string *literal* | fold every non-STRING key to `Uint8ArrayGet` | `typeof u8[Symbol.iterator]` is `number`; a string key held in a local reads a byte |
| #7796 | closed | the value's function-ness, lost once stored in a local | inlined truthiness fast path | `f ? …` falsy while `Boolean(f)` is `true` |
| #7837 | open | declared `string` on a local that actually holds a number | concat lowering for `+` | `s + 7` → `"427"`; `t + "x"` → `"x"` — an operand silently dropped |
| #7844 | open | the initializer's kind, never invalidated by assignment | constant-folded `Array.isArray` | answers from the initializer in **both** directions; a guard admits a number |

**The shared tell is one sentence: correct inline or at the definition site, wrong once the value lives in a local binding.** #7700's recorded tell was already literally *"wrong only when STORED in a local"*, and #7796's title says *"only when stored in a local"*. Two of these were closed with the tell written down and the root not named.

Severity note: #7844's false-*positive* direction and #7837's dropped operand are the two that fail silently in the dangerous direction — a validation guard admitting data it exists to reject, and data vanishing from a string. Both exit 0 with no diagnostic.

## Why this keeps happening

[...]

PR #9135, opened 2026-08-30
fix(hir,codegen): `x ?? null` with an unknown left no longer costs the local its GC root

proggeramlug:
## Symptom

A Three.js world builder (`Accum.add`) compiled with Perry read from-space on its first copying minor: SIGBUS under `PERRY_GC_PROTECT_FROMSPACE=1`, wrong vertex colours / an aborted weapon build otherwise. The quarantine named a 40-byte array retired by minor #0, holder outside the arena — a stack slot. Disassembly put the stale use on `masks[0]` in

```ts
add(count: number, opts: { masks: number[] } | null) {
  const masks = opts?.masks ?? null;
  for (…) { …allocate…; if (masks) { r = Math.max(r, masks[0]); … } }
}
```

## Root cause — two defects, one rule

**1. The `??` type rule answered the RIGHT operand for an unknown left.** Both copies — `crates/perry-hir/src/analysis/value_types.rs` (`infer_logical_type`) and its AST twin in `crates/perry-hir/src/lower_types.rs` (`infer_type_from_expr`, `NullishCoalescing`) — did `if left is Any { infer(right) }`. Optional chaining lowers `opts?.masks` to an `Any`-typed conditional, so `opts?.masks ?? null` was declared **`Null`** (`--trace hir`: `Let { name: "masks", ty: Null, … }`). `unknown ?? null` is unknown, not `null`.

**2. The pointer-locals collector took that inference as proof.** `crates/perry-codegen/src/collectors/pointer_locals.rs` deliberately ignores declared types (#7846) and proves pointer-ness from the initializer through `expr_value_type` — which had no `Logical` arm, so the `??` fell to the generic `infer_expr_type` fallback, got `Null`, and the local was proven non-pointer: no shadow slot, so `precise_roots.rs` never retyped it to `ptr addrspace(1)` and `root_reload.rs` had nothing to reload. `%r4 = alloca double` held the NaN-boxed array address across `js_gc_loop_safepoint`; `arr.guard.deref` dereferenced from-space.

Controls (same binary flags, `PERRY_GC_SCAVENGE_NURSERY_MB=1 PERRY_GC_INCREMENTAL=0 PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800`):

| Initializer | HIR `Let` type | slot in `add` / `add$pshape` | protected run |
|---|---|---|---|
| `opts?.masks ?? null` | `Null` | `alloca double`, no bind | SIGBUS after minor #0 |
| `const masks: number[] \| null = opts?.masks ?? null` | `Union([Array(Number), Null])` | `alloca double`, no bind | SIGBUS after minor #0 |
| `opts === null ? null : opts.masks` | `Any` | `alloca ptr addrspace(1)` | passes |

The annotation is honoured in HIR but cannot help rooting (by design); the ternary passes because the collector's `Conditional` arm fails closed when a branch is unclassifiable. That isolates the `??` path.

## Fix

- `analysis/value_types.rs`: one shared `coalesce_type` — `Any`/`Unknown` left stays unknown; `Null`/`Void` left takes the right type; a union with a nullish member gains the right's members (unknown right → `Any`); anything else is never nullish so the right is unreachable. `lower_types.rs` now calls it; `monomorph/infer.rs` gets the same rule (`None` left stays `None`).
- `collectors/pointer_locals.rs`: an explicit `Expr::Logical` arm ahead of the fallback, with the same `?` discipline as `Conditional`: both operands must classify; `Coalesce` with a nullish left yields the right; equal types collapse; otherwise a union. A root decision no longer rides on the inference at all — an unclassifiable operand keeps the slot. Proven scalars (`1 ?? 2`, `0 || true`) still pay no slot.

## Tests

- `pointer_locals.rs`: the exact HIR the trace printed (parameter receiver, `Let` typed `Null` on purpose) keeps its slot; scalar `??`/`||` still pays none.
- `value_types_tests.rs`: `unknown ?? null` is `Any` (bare local and the full optional-chain lowering); nullish left takes the right; `coalesce_type` union cases.
- `lower/tests.rs`: `const masks = opts?.masks ?? null` is not typed `Null`/`Void`.
- `test-files/test_gap_gc_coalesce_local_root.ts`, registered in `test-parity/gc_repsel_corpus.txt`: `?? null` over an array and over a closure, both live across loop polls, compared against Node.
- Docs: `gc-rooting-invariant.md` gains this as way #6.

Locally: `cargo test -p perry-hir` (lib + integration) and `cargo test -p perry-codegen --lib` green; `cargo fmt --all --check`, clippy, `check_test_registration.py`, `check_gc_doc_claims.py`, `check_gc_env_knobs.py`, `local_binding_type_audit.py` clean.

End-to-end, release build of this branch, the untyped reproducer, same flags that faulted on `35447e706e` (`PERRY_GC_SCAVENGE_NURSERY_MB=1 PERRY_GC_INCREMENTAL=0 PERRY_GC_PROTECT_FROMSPACE=1 PERRY_GC_PROTECT_FROMSPACE_DEPTH=800`): HIR `Let masks ty: Any`; `%r4 = alloca ptr addrspace(1)` in `add`, `add$pshape` and `add$ptr_arrays`; exit 0, `900000 0.35 0.25 0.1 0.35`, 3 copying minors (10,868 objects moved in the first). Before the fix the same run died with SIGBUS on minor #0.

## Not in this PR

`scripts/gc_root_dominance_check.py --unrooted-allocas` reported 0 on the failing IR: the stored value's provenance is a property read (`js_object_get_field_by_name_f64`, `js_object_get_field_ic_miss`, an inline-cache slot load through `inttoptr`), which its heap-source vocabulary (`ALLOC_RE`, `HEAP_SOURCE_CALLS`, `REWRITTEN_LOAD_RE`) does not include. Adding a property-read source class would have caught this; it will also change corpus counts, so it is left for a follow-up.

https://claude.ai/code/session_01XAYMwhwY3emqxEFT88gQUn

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed garbage-collection issues affecting values assigned through nullish coalescing, including optional-chain expressions using `??`.
  * Prevented stale references, crashes, incorrect results, and corrupted data during memory collection.
  * Improved type handling so unknown values are not incorrectly treated as `null`.

* **Tests**
  * Added coverage for nullish coalescing, optional chains, logical expressions, and long-running garbage-collection scenarios.

* **Documentation**
  * Documented the addressed garbage-collection rooting failure.

### DIFF
--- a/changelog.d/9135-coalesce-unknown-left-gc-root.md
+++ b/changelog.d/9135-coalesce-unknown-left-gc-root.md
@@ -1,0 +1,25 @@
+Fixed a moving-GC use-after-move for locals initialised with `x ?? null` (or
+`?? <anything>`) where the left side is not statically typed — most commonly an
+optional chain: `const masks = opts?.masks ?? null`. A Three.js world builder
+compiled with Perry read from-space on its first copying minor (SIGBUS under
+`PERRY_GC_PROTECT_FROMSPACE=1`, wrong vertex colours or an aborted build
+otherwise).
+
+Two defects, one rule. Both copies of the `??` type-inference rule
+(`perry-hir` `analysis/value_types.rs` and `lower_types.rs`) answered the RIGHT
+operand's type whenever the left was unknown, so the binding above was declared
+`null`. The codegen pointer-locals collector ignores declared types by design
+(#7846) and proves pointer-ness from the initializer — but it had no arm for
+logical operators, fell back to that same inference, took `null` as proof, and
+gave the local no GC root: the array's NaN-boxed address sat in a plain
+`alloca double` across the loop poll and the next `masks[0]` dereferenced the
+pre-collection address. An explicit `number[] | null` annotation could not help
+(the collector distrusts annotations); a plain ternary was rooted all along.
+
+The `??` rule is now one shared `coalesce_type`: an unknown left stays unknown,
+a nullish left takes the right type, a nullable union gains the right's members.
+The collector gets an explicit `Logical` arm that requires both operands to
+classify before it answers, so a root decision never rides on the inference
+again. Pinned by unit tests on the exact HIR, a lowering test, and a registered
+`test_gap_gc_coalesce_local_root` witness (`?? null` over an array and over a
+closure, live across loop polls).
--- a/crates/perry-codegen/src/collectors/pointer_locals.rs
+++ b/crates/perry-codegen/src/collectors/pointer_locals.rs
@@ -362,6 +362,33 @@ pub fn collect_pointer_typed_locals(
                     None
                 }
             }
+            // `a ?? b` / `a || b` / `a && b` select one OPERAND, so the result
+            // is classifiable only when both operands are — the same `?`
+            // discipline as `Conditional` below. This arm exists so the answer
+            // never comes from the generic `infer_expr_type` fallback at the
+            // bottom: that inference is structural, and it typed
+            // `const masks = opts?.masks ?? null` as `Null` (the left is an
+            // `Any`-typed conditional, and the old `??` rule answered the
+            // right operand for an unknown left). `Null` is definitely
+            // non-pointer, so the local lost its shadow slot and the array's
+            // pre-collection address sat in a plain `alloca double` across the
+            // loop poll. The HIR rule is fixed too, but a root decision must
+            // not depend on it: an unclassifiable operand keeps the slot.
+            Expr::Logical { op, left, right } => {
+                let left_ty =
+                    expr_value_type(left, local_types, local_value_types, non_pointer_locals)?;
+                let right_ty =
+                    expr_value_type(right, local_types, local_value_types, non_pointer_locals)?;
+                match op {
+                    perry_hir::LogicalOp::Coalesce
+                        if matches!(left_ty, Type::Null | Type::Void) =>
+                    {
+                        Some(right_ty)
+                    }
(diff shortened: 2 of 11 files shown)


proggeramlug:
Merged. Validated as part of a merge train — cherry-picked with three other PRs onto one branch and checked together in a single build rather than separately, to work through a backlog.

Combined validation on the final rebased tree: hir 365 passed, codegen 1354, runtime 2824 (exit 0, 0 abort markers), perry --bins 1066, `run_lint_gates.sh` **all 60 gates passed**, `git diff origin/main --diff-filter=D` empty.

Given the subject — a local losing its GC root through `x ?? null` with an unknown left — the gate that matters here is `gc_root_dominance` / the unrooted-alloca checker, and both are in that 60. The train also ran clean under the runtime suite with zero abort markers.

Two process notes, both mine rather than yours:

- My first push attempt went to the `fork` remote and was rejected — this PR is same-repo (`isCrossRepository: false`), so it needed `origin`. Worth stating because the rejection is quiet if you don't read the push output, and I'd otherwise have "merged" a head I never updated.
- The train initially carried #9137 as well; it was the sole cause of two file-size-cap violations (`proxy_reflect.rs` 2019, `shapes.rs` 2015), so I dropped it and landed the rest. #9137 needs those two splits before it can go in.

PR #9136, opened 2026-08-30
perf(runtime,codegen): one-call birth for fresh capturing closures (−15%)

proggeramlug:
## What

A fresh (identity-carrying) capturing closure was born as `js_closure_alloc` plus one `js_closure_set_capture_bits` runtime call per capture — and each setter re-resolved the GC header, re-checked forwarding, re-dispatched on the object kind for `layout_note_slot`, and paid the write barrier's page-table classification again. After #9128 made every user closure literal a fresh object, that per-capture chain is ~24% of a capturing-closure birth on main, `js_closure_alloc` itself ~34%.

This adds one runtime entry, `js_closure_alloc_init(func_ptr, capture_count, captures_ptr)`: no-collect-first nursery allocation (its `Some` contract keeps the raw capture bits valid, no trigger check), header + bulk slot copy, **one** newborn layout classification (`layout_init_from_slots`: forget-once, then pointer-free / unknown / side-mask — no per-slot notes, no interleaved table removes), and a barrier pass that classifies the parent once for all slots (`runtime_write_barrier_newborn_slots`; when barriers are off it reduces to the incremental-mark shade check per value). The block-boundary fallback takes the original alloc + per-slot setter path unchanged.

Codegen emits it for fresh closures whose captures are all plain bits (`bulk_fresh_init`); box-cell captures keep the per-slot setter path (their `set_closure_box_capture` bookkeeping has no bulk twin), and the reserved `this` / `new.target` slots are pre-filled with the pointer-free sentinel and patched post-create exactly as before. Singleton (compiler-synthesized async-step) closures are untouched.

## Measurements

Mac mini (quiet host), 5 interleaved rounds, median ns/op (node on the same box):

| shape | main (#9128) | **this PR** | Δ | node |
|---|---|---|---|---|
| bare capturing closure `(x) => x + k`, escaping | 28.5 | **21.0** | **−26.3%** | 4.2 |
| literal with a captured arrow field `{ a, b, f: (x) => x + k }` | 31.4 | **22.2** | **−29.3%** | 5.9 |
| captureless arrow field (no captures → not this path) | 19.0 | 19.0 | 0 | 4.2 |
| plain escaping literal | 5.8 | 5.7 | −1.7% | 2.1 |

Two commits: the one-call birth (28.5 → 24.3) and then four cuts inside it found by a Linux `perf annotate` — barrier skipped entirely on pointer-free births, register-resident mask for ≤64 slots, `layout_forget_object` only when the per-object tables can hold an entry, counted store loop for ≤8 slots instead of a `memcpy` PLT call (24.3 → 21.0). The birth loop on the Linux host runs at IPC 4.97, i.e. throughput-bound, so instruction count is the cost. Dev-box loop with 100k retained closures (attribution only): 118.8 → 78.3 → 62 ns/op; the per-capture setter chain is gone and what remains is the single birth call (allocation, copy, one layout classification, one barrier pass), the moving minors the retention forces, and the *call* side (`js_number_coerce` on the captured operand — a separate family). The remaining gap to node is the birth call's own cost; that is an instruction-level round on the Linux host, not a structural change.

## Correctness

- New closure-birth differential vs node — plain captures, boxed (mutated-after-capture) captures, `this`-capturing arrows in classes, `new.target` inside an arrow, async closures capturing locals, closure identity across evaluations, arrays of closures from a loop, nested three-level captures, a 10-capture closure, a 100k-iteration churn: **byte-identical**.
- #9122's method-literal set, the 300k-birth churn set, and the function-expression `this` set: byte-identical. (The captureless method-shorthand `FuncRef` singleton identity line differs from node exactly as on main — a HIR decision this branch doesn't touch.)
- Gates: full battery (runtime suite included) running on this branch — results in a comment.

## Binary size

Per site: the capture buffer fill (`alloca` + one store per capture) replaces N setter calls; fixture (four closure loops, arm64 `size -m`): `__text` 10,948,052 → 10,950,932 = **+2,880 B (+0.026%)** across both commits. The cc-bundle `size(1)` pair is queued on perrymaster behind the #9124/#9130 measurement.

## Summary by CodeRabbit

  * Improved closure creation efficiency by initializing captured values in bulk when possible.
  * Optimized garbage-collection metadata and write-barrier handling during closure initialization.
  * Added efficient handling for small closure allocations and pointer-free captures.
  * Preserved existing behavior through fallback allocation for unsupported or unusually large closures.
  * Improved garbage-collection tracking during active Set and Map iteration.

### DIFF
--- a/crates/perry-codegen/src/expr/closure.rs
+++ b/crates/perry-codegen/src/expr/closure.rs
@@ -340,6 +340,16 @@ pub(crate) fn lower(ctx: &mut FnCtx<'_>, expr: &Expr) -> Result<String> {
                 None
             };
 
+            // Bulk-init admission: a fresh closure whose captures are all plain
+            // bits. Box-cell captures keep the per-slot setter path — their
+            // `set_closure_box_capture` bookkeeping has no bulk twin.
+            let bulk_fresh_init = !no_capture_singleton
+                && !captured_singleton
+                && total_caps > 0
+                && !captured_value_bits.is_empty()
+                && auto_captures
+                    .iter()
+                    .all(|cap_id| is_plain_async_step || !ctx.boxed_vars.contains(cap_id));
             let closure_handle = if no_capture_singleton {
                 let blk = ctx.block();
                 blk.call(I64, "js_closure_alloc_singleton", &[(PTR, &func_ref)])
@@ -372,6 +382,32 @@ pub(crate) fn lower(ctx: &mut FnCtx<'_>, expr: &Expr) -> Result<String> {
                 blk.call(
                     I64,
                     "js_closure_alloc_with_captures_singleton",
+                    &[(PTR, &func_ref), (I32, &cap_count), (PTR, &buf)],
+                )
+            } else if bulk_fresh_init {
+                // Fresh (identity-carrying) closure with plain-bits captures:
+                // ONE runtime call does allocation + slots + layout instead of
+                // `js_closure_alloc` plus a `js_closure_set_capture_bits` per
+                // capture (each re-resolving the header, forwarding, kind
+                // dispatch and the barrier's page classification). The reserved
+                // `this` / `new.target` slots are pre-filled with the pointer-free
+                // sentinel the per-slot path relied on `js_closure_alloc` writing;
+                // the post-create patch below fills them exactly as before.
+                let buf = ctx.func.alloca_entry_array(I64, total_caps);
+                {
+                    let blk = ctx.block();
+                    for i in 0..total_caps {
+                        let slot = blk.gep(I64, &buf, &[(I64, &format!("{}", i))]);
+                        match captured_value_bits.get(i) {
+                            Some(v_bits) => blk.store(I64, v_bits, &slot),
+                            None => blk.store(I64, crate::nanbox::TAG_UNDEFINED_I64, &slot),
+                        }
+                    }
+                }
+                let blk = ctx.block();
+                blk.call(
+                    I64,
+                    "js_closure_alloc_init",
                     &[(PTR, &func_ref), (I32, &cap_count), (PTR, &buf)],
                 )
             } else {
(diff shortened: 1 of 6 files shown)


proggeramlug:
Full battery on this branch: `-D warnings` 0, codegen 1832/0, **full runtime suite 2822/0**, lints clean (census, addr-class audit, file-size, raw-handle debt none raised), integration issue_8655 2/2 / issue_8690 3/3 / issue_8897 3/3. (The integration phase is a rerun — the first pass hit ENOSPC on a shared box and printed no result.)

proggeramlug:
Second commit pushed — four cuts inside the birth entry, all found by a Linux `perf annotate` of the loop (18.5 ns/op at **IPC 4.97**, i.e. throughput-bound, so instruction count is the cost):

- **barrier skipped entirely on pointer-free births** — `layout_init_from_slots` now returns whether any slot is pointer-bearing; a closure capturing only numbers paid a call plus a page-table classification per slot for a barrier whose own child check rejects every one of them (`write_barrier_slot_decoded` was 9.3% of the loop on a *number* capture)
- ≤64 slots classify into a register-resident u64 instead of a `LayoutSlotMask`, with the mask-min-slots threshold read once rather than through a per-birth `OnceLock` call
- `layout_forget_object` only when the per-object tables can hold an entry (4.5%)
- ≤8 slots copy through a counted store loop; the runtime-length copy compiled to a `memcpy` PLT call (2.6% for **one** slot)

Mini medians: bare capturing closure 24.3 → **21.0** ns (−13.6%), captured-arrow-field literal 27.8 → **22.2** (−20.1%); captureless and plain literals unchanged. Cumulative against main: **28.5 → 21.0** and **31.4 → 22.2**. Closure-birth differential vs node still byte-identical; fixture `.text` +2,880 B (+0.026%) across both commits. Full battery rerunning.

proggeramlug:
Full battery on both commits: `-D warnings` 0, codegen **1832/0**, full runtime suite **2822/0**, lints clean (census, addr-class audit, file-size, raw-handle debt none raised), integration issue_8655 2/2 / issue_8690 3/3 / issue_8897 3/3. Ready for review.

proggeramlug:
Merged as part of a merge train — several PRs cherry-picked onto one branch and validated together in a single build rather than separately, to work through a backlog. Combined validation on the final rebased tree: hir 365 passed, codegen 1354, runtime 2824 (exit 0, 0 abort markers), perry --bins 1066, `run_lint_gates.sh` **all 60 gates passed**, `git diff origin/main --diff-filter=D` empty.

**I added one commit.** `gc_store_site_inventory` flagged the new counted capture-store loop (`std::ptr::write(slots.add(i), ...)`). I classified it `GC_STORE_AUDIT(BARRIERED)` to match the `copy_nonoverlapping` arm directly beneath it — both are followed by the same closure layout/barrier rebuild, so they warrant the same class. Worth noting the gate only caught the counted loop and not the `copy_nonoverlapping` beside it because that one was already marked; if you add a third arm it will need its own.

PR #9137, opened 2026-08-30
perf(object): keep populated-delete ICs stable per key

### LINKED ISSUE #9064: perf: populated delete is ~14x node after tombstones — the residual is IC retirement per delete, not the delete itself
## Where we are

Tombstone deletes shipped default-on (#9029 → #9038): `bench_populated_delete` went 2030 → ~315 ms on the Linux box (6.5×) and ~2100 → 332 ms on the Mac mini. The remaining gap vs node is **~14×** (node 21-24 ms), and it is structural, not constant-factor — this issue records where the time is and the ranked options, so the next campaign doesn't re-derive it.

## Self-contained repro

```ts
// bench_populated_delete.ts — 500 keys stay resident, one key churns.
for (let i = 0; i < 500; i++) o["k" + i] = i;
let s = 0;
const N = 200000;
const t = Date.now();
for (let i = 0; i < N; i++) {
  const k = "k" + (i % 500);
  delete o[k];
  o[k] = i;
  s += o[k];
}
console.log("popdel_ms=" + (Date.now() - t) + " chk=" + (s % 7));
```

Cross-engine, Mac mini, min-of-7 interleaved, self-timed: node 23 · **perry 332** · porffor 113 · **scriptc 24**. The scriptc number is the data point that matters: an AOT competitor does this at node parity, almost certainly via dictionary-style storage — parity is achievable in an AOT setting.

## Where the ~315 ms goes (perf, symbolized, flag-on, Linux)

The profile is a **flat tail** — after the tombstone work no symbol is above ~20%:

- `shape_slot_lookup_verdict` ~20% total, split by caller (DWARF callgraph): **8.5% the delete's own slot-find** (`js_object_delete_field → keys_find_slot_by_key_ptr`), 3.8% read-path misses, 3.5% write-path misses. Per call it's TLS + `RefCell` borrow + two hash lookups (~35-40 ns) — call count × fixed cost, not list length.
- Publish machinery per delete: `remove_descriptor_id_from_facts_index` 2.8%, `shape_descriptor_by_id` 2.8%, `shape_descriptor_ensure_with_holes` 1.7%, plus `FastKeyHasher` 2.8% and `from_utf8` 2.8% (SSO decode during candidate validation).
- The rest: `get_field_by_name` tails, `js_put_value_set`, `js_object_delete_field`, accessor probes — each 1-3%.

## The structural cause

**Every delete mints a fresh shape token** (this is forced: per-site dyn-IC ways live in generated-code globals the runtime cannot reach, so retiring a deleted key's cached `(token, key) → slot` entries requires changing the token — see #9029's description). Consequence: after each delete, EVERY inline cache on that receiver misses once — the loop's read and write both take their miss paths every iteration. Node doesn't pay this: V8's dictionary-mode ICs validate against the dictionary, not a shape, so an unrelated delete invalidates nothing.

## Ranked options

1. **Per-key invalidation instead of whole-token retirement** (the real fix, larger design). Keep the shape token STABLE across tombstone deletes; make IC hits validate the cached slot's key content (one load+compare — the slot still holds the key unless tombstoned, and a tombstoned slot holds `TAG_HOLE`, which never content-matches) or check a per-object delete epoch only on the deleted key's hash class. Either kills the miss-per-delete for the 499 untouched keys. Risk surface: every IC hit gains a compare; measure on the write/read benches that currently BEAT node before accepting.
2. **Cheapen the per-delete publish**: the mint+sweep+facts-index dance is ~8-10% combined. With (1) in place most of it disappears entirely (no successor id needed per delete — only the hole write + epoch bump).
3. **Verdict-lookup fixed cost** (~20%): one-entry TLS memo `(keys_id → *mut ShapeIndex)` with `Box`ed `ShapeIndex` values for address stability, invalidated on any indices insert/remove. Saves the borrow + first hash lookup on the 3-4 lookups per op. Bounded win (~40-60 ms of 315).
4. **Full dictionary mode** — superseded as first move by (1), but it's what scriptc's parity suggests; the tombstone walkers already built the hole-skipping groundwork (#9029's audit, `DELETE_TOMBSTONES_DESIGN.md` §3/§7).

## Acceptance

- `bench_populated_delete` ≤ 3× node on the mini AND the Linux box, min-of-7 interleaved vs the exact main tip.
- No regression (±2%) on the benches that now beat node: plain write loop (21 vs 33), combined overwrite (34 vs 38), dynamic-prop overwrite (13 vs 19).
- The #9029 differential battery (enumeration / adversarial / SSO / stale-slot + the two holed-JSON-array fixtures) byte-identical to node — option (1) changes IC validation, which is exactly what the stale-slot differential exists to catch.

proggeramlug:
## Summary

- keep one ShapeId across owned ordinary-object tombstone delete/re-add epochs
- make read/write IC hits reject a cached slot whose value is `TAG_HOLE`
- add a guarded re-add lane that preserves enumeration order and re-primes both megamorphic caches
- keep #9110 squeeze publication ordering intact and retire owned growth-prefix tokens before flag-off in-place compaction
- exclude prototype/direct-method guard targets from stable epochs and reopen full delete checks when descriptors are installed

Fixes #9064.

## Performance

Linux `perrymaster`, exact main `653e88669911ec6c15793891a428a873d41fd7fe`, `PERRY_NO_CACHE=1`, tombstones enabled, min-of-7 interleaved. The timed processes were pinned to one physical core while an unrelated compiler occupied a different physical core.

| benchmark | exact main | patch | node |
|---|---:|---:|---:|
| populated delete | 329 ms | **57 ms** | 21 ms |
| plain write | 15–16 ms | 16 ms | 21 ms |
| combined overwrite | 25 ms | 26 ms | 26 ms |
| dynamic-prop overwrite | 11 ms | 11 ms | 13 ms |

The 1 ms self-timer quantum obscures the ±2% write gate, so I also ran 15-repeat hardware-counter comparisons on the same isolated core:

- plain write: **+0.03% cycles**
- combined overwrite: **+0.61% cycles**
- populated delete: 1.748B → 0.332B cycles; 5.004B → 1.265B instructions
- patched populated delete: 3.81 IPC, 0.12% branch misses

The dynamic delete-heavy column remains essentially unchanged; this patch specifically targets the populated-object churn in the issue repro.

Binary `.text`: 8,130,354 → 8,136,578 bytes (**+6,224 / +0.08%**).

The Linux target is 2.71x Node. I did not have access to the Mac mini, so that half of the cross-platform acceptance check remains for review/CI.

## Correctness and validation

- `cargo test --release -p perry-runtime tombstone_`: 19 passed
- `cargo test --release -p perry-codegen --lib`: 1,349 passed, 1 ignored
- #9029 enumeration, adversarial read, SSO, stale-slot, and both holed-JSON differentials: byte-identical to freshly generated Node output with tombstones both on and off
- unit coverage pins stable IDs, anonymous object literals, the 60-cycle ≤2x storage bound, stale growth-prefix retirement, and descriptor installation after a stable epoch

## Summary by CodeRabbit

* **Performance**
  * Improved performance for repeated delete-and-readd operations on ordinary objects.
  * Preserved faster property access while correctly handling deleted properties.

* **Bug Fixes**
  * Deleted properties no longer return stale cached values.
  * Property writes now safely fall back when cached slots represent deleted data.
  * Improved correctness for descriptor changes, compaction, and prototype lookups.

* **Tests**
  * Added coverage for stable object identity, property re-addition, descriptor invalidation, and compaction behavior.
  * Verified existing differential tests remain byte-identical.


proggeramlug:
Not merging yet — **two file-size-cap violations**, both caused by this PR alone.

I had this in a merge train with four other PRs; the train failed `check_file_size.sh`, and dropping this PR made it pass, which isolates it:

```
2019  crates/perry-codegen/src/expr/proxy_reflect.rs   (main: 1945)
2015  crates/perry-runtime/src/object/shapes.rs        (main: 1981)
```

The cap is 2000 lines and `check_file_size.sh` is a required `lint` step, so this cannot land as-is. The gate's own recipe is the fix: extract a function group into a sibling module and re-export from the parent with explicit named `use` statements (globs don't propagate through transitive re-exports).

Two things I learned doing exactly this split on #9125 an hour ago, which will save you a round trip:

1. **A new module starts with a raw-handle ceiling of zero.** `raw_handle_debt.py`'s bare run locks unlisted modules at zero bare reads, so if you relocate even one pre-existing `get_raw_*_ptr` into the new file, the gate goes red for code you didn't write. Convert it to `with_const_ptr`/`across_const` rather than adding a ceiling — and note that only raw *pointer* reads count; `get_nanbox_f64`/`get_heap_word_u64` value reads are fine to leave.
2. **Watch for a now-unused import** in the parent after the extraction — `-D warnings` turns that into a build failure.

The rest of the PR looked fine in the train: with it included, hir 365 / codegen 1354 / runtime 2827 all passed (runtime exit 0, 0 abort markers), so the only blockers are the two splits.

Ping me when it's rebased and I'll re-run the full set.

proggeramlug:
Merged. The two file-size violations are fixed, and fixed the right way — by a Codex agent I briefed with the exact failures and the three traps this split hits, then reviewed rather than took on trust.

**What landed:**
- `proxy_reflect.rs` 2019 → **1979**: the stable-tombstone validation shared by `lower_put_value_static_write_ic` and `lower_put_value_dyn_ic_inline` extracted to `proxy_reflect_write_ic.rs` as `StableTombstoneSlotCheck::{new, emit, enter_live}`.
- `shapes.rs` 2015 → **1917**: the reverse-index helpers (`descriptor_facts`, `descriptor_facts_with_keys`, both removal helpers, `insert_descriptor_id_sorted`, `sync_descriptor_reverse_indices`, `remove_descriptor_and_reverse_indices`) to `shapes_reverse_indices.rs`.

Explicit named imports both ways, `mod` declarations wired, and **no ALLOWLIST entry** — the escape hatch stayed shut, which is what I asked for.

Three things I checked specifically, because each is a way this kind of split goes quietly wrong:

| check | result |
|---|---|
| `raw_handle_debt_baseline.txt` direction | **967 → 963** — a *tightening*. A new module starts at a zero ceiling, so relocating a bare read normally reddens the gate; it converted four to real scoped/across operations instead. |
| gate-gaming empty closures (`across_*(\|\| ())`) | **0**. That pattern is byte-identical to the bare read it replaces and I'd have rejected it — it's what the first attempt on #9103 did. |
| `check_file_size.sh` ALLOWLIST touched | no |

Verified independently in the agent's tree, not from its report: `check_file_size.sh` PASS, fmt clean, runtime **2836 passed** (exit 0, 0 abort markers), codegen **1356 passed**, `run_lint_gates.sh` **all 60 gates passed**; `git diff origin/main --diff-filter=D` empty.

Note for #9140: it is stacked on this PR (its branch carries this commit), so it needed this to land first. Picking it up now.

PR #9140, opened 2026-08-30
perf(runtime): reuse tombstones for small-object churn

### LINKED ISSUE #9065: perf: add/delete churn on a small object is ~33x node — shape identity turns over on EVERY op (and the bench's own header now argues the opposite of reality)
## The number

`bench_dynamic_property_keys.ts`'s delete-heavy loop: **perry 981 ms vs node 30 ms** (Mac mini, min-of-7, current main with #9038 default-on). Cross-engine: scriptc 29 ms (node parity), porffor fails to compile it. Sibling issue for the large-object case: #9064 — same root cause, different scale and different dominant term.

## Self-contained repro

```ts
function deleteHeavy(n: number): number {
  let s = 0;
  for (let i = 0; i < n; i++) {
    const k = "k" + (i % 500);
    o[k] = i;
    s += o[k];
    delete o[k]; // the object never holds more than one live key
  }
  return s;
}
// N = 300_000 → perry ~981 ms, node ~30 ms.
```

Unlike #9064's stable 500-key receiver, this object oscillates around **one live key**. Per iteration: an append (new shape id), a read (IC primed against it), a delete (successor id minted, every cache retired). ~3.3 µs/op vs node's ~0.1 µs — the shape-identity machinery IS the loop.

## What tombstones already do here (and why it's not enough)

The tombstone threshold (`key_count >= 16` before squeezing) means the array grows to 16 slots, squeezes back to ~1, repeats — amortized ~1 slot-copy per delete. Fine. What remains is that **all three ops per iteration change or chase shape identity**:
- the append publishes a successor id (lineage mint + facts-index insert),
- the delete publishes another one + sweeps stale ids for the owned address,
- and both retirements force the read's and write's ICs to miss every single iteration (`js_put_value_set_dyn_ic_miss` and the `get_field_by_name` tail dominate the old profile of this binary, and the descriptor mint/lookup pair — `shape_descriptor_by_id` + `ensure_with_holes` — runs multiple times per op).

## What to optimize, in order

1. **Stable-token deletes + per-key IC validation** — #9064's option (1). For THIS bench it removes the delete-side retirement; the read IC then survives the delete and only the append still moves identity.
2. **Append-side identity churn on re-add**: a re-added key appends and mints an id per append. For churn shapes, a small transition cache keyed `(predecessor_id, key)` → successor id already exists for first-time builds — verify it HITS on the delete→re-add cycle (the tombstone delete's fresh generation may make every re-add transition a cold miss forever). If each `k(i%500)` cycle re-mints instead of reusing 500 cached transitions, that's the memory-and-time leak to close.
3. **Small-object fast path**: below the 16-key threshold the per-delete publish could collapse to hole-write + epoch bump with NO descriptor machinery at all once (1) lands — a one-live-key object needs none of it.

## Housekeeping in the same change

The bench's header comment (dated 2026-08-28) now argues the **inverse of reality** and should be rewritten when this is touched: it says the delete penalty is 1.1× ('LOWER than node's — dictionary mode refuted') and overwrite is the real 60× gap. After the write campaigns, overwrite is 13 ms and **beats node** (19 ms); the delete penalty is **75×**. The ratio that used to refute the dictionary-mode argument is now the strongest evidence for #9064's per-key-invalidation work.

## Acceptance

- `delete_heavy_ms` ≤ 3× node on the mini; `overwrite_ms` stays ≤ node (currently 13 vs 19 — must not regress).
- delete_penalty back under ~2× (node's is 1.6×).
- #9029 differential battery byte-identical, both tombstone flag states.

proggeramlug:
## Summary

- fork the first small shared layout into one owned tombstone epoch, then handle SSO delete/re-add without key materialization, shape-table hashing, or allocation while capacity remains
- update detached private shape counters in place and mint a fresh token only when the all-hole squeeze reuses slots, preserving enumeration order and cache safety
- make SSO/heap/raw key decoding safe in JSON stringify and remove unsafe write-to-read cross-priming exposed by compacted layouts
- include the actionable #9137 review/CI follow-ups: deleted read hits take the real miss-feedback path, detailed class-field guards reject stable epochs, moving-GC reloads are explicit, and repository ratchets pass

Fixes #9065.

Depends on #9137. Until that PR merges, this branch contains its dependency as the first commit; the #9065 implementation and follow-ups are isolated in the second commit.

## Performance

Linux `perrymaster`, exact main `653e88669911ec6c15793891a428a873d41fd7fe`, `PERRY_NO_CACHE=1`, same release toolchain and source, min-of-7 interleaved on one core:

| benchmark | exact main | patch | node |
|---|---:|---:|---:|
| dynamic delete-heavy | 1,281 ms | **33 ms** | 27 ms |
| dynamic overwrite | 11 ms | **11 ms** | 12 ms |

The delete-heavy path is 38.8x faster than exact main and 1.22x Node, inside the issue target of no more than 3x Node. The overwrite path is unchanged and still matches/beats Node. Checksums were zero in every run.

Binary `.text`: 12,547,602 -> 12,559,874 bytes (**+12,272 / +0.098%**).

## Correctness and validation

- `cargo build --release -p perry -p perry-runtime-static -p perry-stdlib-static` from a fresh remote worktree/target
- `cargo test --release -p perry-runtime tombstone -- --nocapture`: 24 passed
- focused native-proof regression for the five write-PIC fallback arms: passed
- #9029 enumeration, adversarial read, SSO, stale-slot, and both holed-JSON fixtures: byte-identical to fresh Node output with tombstones both off and on (12/12)
- `cargo fmt --check` and `git diff --check`
- file-size, address-classification, and raw-handle ratchets pass; raw-handle debt falls by five

The benchmark header retains both historical ledgers requested in the issue. No version files were changed.

## Summary by CodeRabbit

- **Performance**
  - Improved performance for repeated adding and deleting of properties on small objects.
  - Preserved optimized property access during object churn and property re-addition.

- **Bug Fixes**
  - Fixed property reads and writes after a property has been deleted and later restored.
  - Improved JSON serialization of object keys, including short-string keys and objects affected by garbage collection.
  - Strengthened handling of object shapes and property ordering during deletion and re-addition.

### DIFF
--- a/benchmarks/bench_dynamic_property_keys.ts
+++ b/benchmarks/bench_dynamic_property_keys.ts
@@ -14,19 +14,21 @@
 //   node           36 ms           21 ms           1.7x
 //   perry        1487 ms         1321 ms           1.1x
 //
-// The delete penalty is the number the "objects that defeat shapes need a
-// dictionary mode" argument rests on — and perry's is LOWER than node's. Adding
-// a dictionary representation would therefore be a large investment aimed at a
-// tail perry does not have.
+// That ledger was accurate at the time. After the dynamic-write campaigns and
+// tombstone deletes became default-on, however, the two columns inverted
+// (Mac mini, current main at #9065 filing, min-of-7):
 //
-// The second column is the real gap: ~60x on plain overwrite. Profiling this
-// binary puts the time in `js_array_get_f64`, `try_read_tracked_gc_header`,
-// `shape_descriptor_by_id` + `shape_descriptor_ensure_with_generation` (two
-// hash lookups per access on the hot path), and `js_put_value_set_dyn_ic_miss`
-// — i.e. inline-cache misses and shape-table probes, not deletion.
+//   engine   delete_heavy   overwrite_only   delete penalty
+//   node           30 ms           19 ms           1.6x
+//   perry         981 ms           13 ms          75.5x
 //
-// Keep both columns when changing this file: the ratio is what refutes the
-// dictionary-mode premise, and the absolute is what tracks the real gap.
+// Perry's overwrite loop now beats node; delete-driven shape identity is the
+// remaining gap. The ratio that originally argued against dictionary-style
+// handling is now the strongest evidence for stable-token, per-key-validated
+// churn shapes (#9064/#9065).
+//
+// Keep BOTH dated ledgers when changing this file. They record a real inversion
+// in where the cost lives, not an error in the original measurement.
 
 function deleteHeavy(n: number): number {
   const o: Record<string, number> = {};
--- a/changelog.d/9065-small-object-churn.md
+++ b/changelog.d/9065-small-object-churn.md
@@ -1,0 +1,7 @@
+Improved repeated add/delete churn on small objects. A first delete now forks
+an owned tombstoned key layout instead of compacting a transition-cache-shared
+layout back to empty, allowing subsequent stable-token delete and append paths
+to keep their inline caches alive.
+
+This removes the per-iteration keys-array allocation loop while preserving the
+existing bounded append/squeeze behavior and both flag states.
--- a/crates/perry-codegen/src/expr/property_get/generic_dispatch.rs
+++ b/crates/perry-codegen/src/expr/property_get/generic_dispatch.rs
@@ -565,7 +565,7 @@ pub(crate) fn lower_generic_property_get(
         .block()
         .icmp_eq(I64, &val_hit_bits, crate::nanbox::TAG_HOLE_I64);
     ctx.block()
-        .cond_br(&hit_deleted, &call_label, &hit_live_label);
+        .cond_br(&hit_deleted, &miss_label, &hit_live_label);
 
     ctx.current_block = hit_live_idx;
     crate::expr::emit_typed_feedback_record_call(
--- a/crates/perry-codegen/tests/native_proof_regressions.rs
+++ b/crates/perry-codegen/tests/native_proof_regressions.rs
@@ -15276,8 +15276,8 @@ fn nested_same_shape_object_writes_version_one_through_four_fields() {
         rejected
             .matches("call double @js_put_value_set_ic_miss")
             .count(),
-        20,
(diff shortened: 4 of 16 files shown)


proggeramlug:
Not merging yet — **four runtime tests fail**, and they're all in the delete/shape-transition area this PR changes.

I had this in a merge train with #9145, #9141 and #9146. The train showed `runtime: 2831 passed; 4 failed`; dropping this PR made it `2831 passed; 0 failed`, which isolates all four to here:

```
object::delete_rest::shape_transition_tests_6759::delete_mints_a_fresh_shape_id_for_a_class_instance
object::delete_rest::shape_transition_tests_6759::delete_mints_a_fresh_shape_id_for_a_plain_object
object::field_get_set::ic_miss::c3c_pic_tests::a_compacted_class_instance_primes_a_token_a_pristine_sibling_cannot_match
object::shapes::shapes_tests::descriptor_tests_8067::delete_compaction_never_compares_equal_to_the_predelete_layout
```

The first one reports:

```
assertion `left == right` failed: test premise: the delete did not compact the slots
  left: 20.0
 right: 30.0
```

That wording matters: these are **fixture-precondition** assertions, not behavioural ones. The tests set up a state they expect the delete path to produce, and this PR changes what a delete produces, so the setup no longer holds — which means every verdict those tests would go on to make is vacuous rather than merely wrong.

So the question for you is which of two things this is, and I can't answer it from outside:

1. **The tests are now stale** — reusing tombstones for small-object churn is intended to stop compacting in these cases, and the fixtures should be updated to construct the compacted state some other way (or assert the new behaviour). Then this is a test-maintenance change that belongs in this PR.
2. **The reuse is reaching cases it shouldn't** — e.g. a class instance or a PIC-primed sibling where the old compaction was load-bearing for shape identity. `delete_compaction_never_compares_equal_to_the_predelete_layout` is the one I'd read first; if tombstone reuse can make a post-delete layout compare equal to its pre-delete layout, that's a real shape-identity hazard rather than a stale fixture.

Everything else in the train was clean with this PR included — hir 365, codegen 1356, perry --bins 1066, and all 60 lint gates — so the four tests are the only blocker.

PR #9141, opened 2026-08-30
perf(runtime): BigInt division fast paths — pi's #1 startup symbol, 33.8x (67.6x → 2.0x of node)

proggeramlug:
Found by the first named-symbol profile of pi's native startup: `perry_runtime::bigint::arith::unsigned_div_limbs` was the **top CPU symbol at ~17% of samples**. Cause: TypeBox's `Value.Hash` runs, per hashed byte,

```js
var [Prime, Size] = [BigInt("1099511628211"), BigInt("18446744073709551616") /* 2^64 */];
Accumulator = (Accumulator ^ Bytes[byte]) * Prime % Size;
```

so every byte of every schema hashed at startup took a full limb-division loop where node masks.

**Four tiers** in `unsigned_div_limbs` (magnitudes only; signs re-applied by callers, untouched): (0) zero dividend or dividend < divisor returns directly; (1) **power-of-two divisor** → remainder is the low k bits, quotient is `a >> k`, O(limbs); (2) **single-limb divisor** → one hardware `u128/u64` per dividend limb; (3) the original bit loop, now bounded by the dividend's real bit length instead of a fixed 1024. Tiers only add early exits and shrink tier 3's bound, so there is no regression path.

**Benchmark** — FNV1A64 over 200k bytes, min of 3, quiet Mac mini (the dev box was at load 52-144 from other agents and unusable for timing):

| | ms | vs node |
|---|---|---|
| node v26.5.1 | 20 | 1.0x |
| perry before | 1352 | 67.6x |
| perry after | **40** | **2.0x** |

Accumulator `11550294541936438046` byte-equal to node in every run; before/after built from an exact stash/unstash of the same tree with `perry-auto-*` cleared and archives confirmed rebuilt.

**Validation**: runtime suite **2829 passed / 0 failed** (run twice, no SIGABRT); `test_gap_9092_bigint_pow2_mod.ts` byte-identical to node **both before and after** — the before-run is the control proving the baseline was already semantically correct. It covers `(-7n)%4n === -3n`, negative divisors, canonical zero on exact multiples, dividend < / == divisor, 250-bit dividends % 2^64 / 2^13 / 2^70, a 1022-bit dividend % 2^1021, `% 1n`, `% 0n` throwing RangeError, non-power-of-two single- and multi-limb divisors, a k-sweep across the limb boundary, and a 96-case identity/sign/range sweep. Added `div_mod_cross_check_against_i128_all_tiers`: 12,000 randomized differential cases against host `i128` across all tiers and sign combinations, **mutation-tested** (clearing one mask bit makes it fail). Also replaced an inherited `expect()` on the divisor-nonzero path — a panic inside an `extern "C"` frame aborts the process — with a debug assert plus a safe return.

**Next in this area** (not this PR): tier 3 is now the outlier at 154 ns vs node 15 for multi-limb non-power-of-two divisors; Knuth algorithm D would close it.

Implemented by subagents in an isolated worktree (across two sessions after a rate-limit interruption); reviewed and shipped by the coordinating session.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved BigInt division and remainder calculations across power-of-two, small, and large values.
  * Corrected handling for zero values, negative operands, wide numbers, and results near the 1024-bit range.
  * Prevented incorrect behavior for division by zero while preserving expected error handling.
  * Improved performance by using optimized calculation paths instead of a fixed-length iteration.

* **Tests**
  * Added comprehensive coverage for BigInt quotient and remainder behavior, including randomized and cross-limb scenarios.

### DIFF
--- a/crates/perry-runtime/src/bigint/arith.rs
+++ b/crates/perry-runtime/src/bigint/arith.rs
@@ -263,7 +263,31 @@ fn effective_limb_len(limbs: &[u64; BIGINT_LIMBS]) -> usize {
     1
 }
 
-/// Unsigned binary long division on magnitude limbs
+/// Index of the highest set bit across the limbs, or `None` if all zero.
+#[inline(always)]
+fn highest_set_bit(limbs: &[u64; BIGINT_LIMBS]) -> Option<usize> {
+    for i in (0..BIGINT_LIMBS).rev() {
+        if limbs[i] != 0 {
+            return Some(i * 64 + 63 - limbs[i].leading_zeros() as usize);
+        }
+    }
+    None
+}
+
+/// Unsigned binary long division on magnitude limbs.
+///
+/// Callers pass non-negative magnitudes (`b` nonzero — division by zero is
+/// thrown before reaching here) and re-apply signs themselves. Three tiers:
+///
+/// 1. **Power-of-two divisor** (single set bit at position `k`, e.g. the
+///    `% 2^64n` / `% 2^32n` idiom common in JS hashing code such as TypeBox's
+///    FNV-1a `Value.Hash`): remainder = low `k` bits of the dividend,
+///    quotient = dividend `>> k`. O(limbs). `k == 0` means divisor 1 —
+///    remainder 0, quotient = dividend.
+/// 2. **Single-limb divisor** (covers `% prime` for primes < 2^64): one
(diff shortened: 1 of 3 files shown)


proggeramlug:

PR #9146, opened 2026-08-30
codegen: byte-array reads reach the numeric proofs (bench_int_arithmetic 475 → 395 ms)

proggeramlug:
Two numeric-lowering fixes for byte-array reads, both found working `benchmarks/suite/bench_int_arithmetic.ts` — the worst entry on a fresh perry-vs-node sweep (7.7× Node before this; 6.2× after, with the remaining gap identified below).

## 1. `acc += px[i]` on a `Uint8Array` took the dynamic add

The HIR lowers a read on a binding it already knows is a `Uint8Array`/`Buffer` to the dedicated `Uint8ArrayGet` node. The number-by-construction fixpoint's number-or-`undefined` view-read rule — and the not-BigInt predicate beside it — only recognised the `IndexGet` spelling, so the accumulator was never admitted and every `+=` lowered through `js_dynamic_string_or_number_add`, with `acc` pinned in a GC-rooted shadow slot.

**This fix has a correctness half.** Once the add is a raw `fadd`, an out-of-bounds read's NaN-boxed `undefined` survives it — IEEE arithmetic preserves the NaN payload — and the accumulator reads back as `undefined` where the spec says `NaN`. Number-context lowering now canonicalizes the value first: one compare plus a select, never a call. The `undefined` box is the only non-double this node can produce, so the test is exact.

## 2. Compound buffer indices could not prove their bounds

`bounds_for_buffer_access_width` proved only a single index local carrying a bounded-pair fact, or a constant. An index like `i * 2 + 1` or `y * 30 + x` therefore fell back to a per-element `js_uint8array_index_get_value` call. The interval analysis can bound those: every leaf is a counter with a range fact or a compile-time constant, and `int_range_expr` composes them with checked arithmetic, answering `None` as soon as a leaf is unknown or a step overflows. When the whole interval fits inside a **constant** buffer length, the access is in bounds on every iteration. A non-constant length still declines — there is nothing to compare the interval against.

## Measurements

Idle Mac mini, min of three, self-timed (the benchmark's own printed elapsed, not wall clock — node's ~76 ms startup otherwise flatters perry):

| probe | before | after | node |
|---|---:|---:|---:|
| `bench_int_arithmetic` | 475 ms | **395 ms** | 64 ms |
| `pixels[y * SIZE + x]` variant | 636 ms | **356 ms** | 35 ms |
| all-qualifying-reads probe | 54 helper calls/iter | **0** | — |

## Correctness

Byte-identical to node on two differentials that belong to neither suite:

- **eight cases around the byte read** — in-bounds accumulate, an out-of-bounds read inside `+=`, an out-of-bounds read as a value, string concatenation, negative and fractional indices, a `Buffer` receiver, and a mixed numeric/string accumulator;
- **seven cases around the new bounds proof** — the exact kernel shape, an interval touching the last element, a counter mutated inside the loop body, a deliberately out-of-range interval, a one-past-the-end read, a negative composite index, and a buffer whose length is not a compile-time constant.

`perry-codegen`: 10 collector tests and all 284 `native_proof_regressions` pass.

## Two things left on the table, deliberately

- The bench's own `(y + ky) * SIZE + (x + kx)` still declines the interval proof (54 reads per pixel remain calls), so this is a step, not the finish. It is not the counters' negative start — a non-negative rewrite still declines.
- **Landmine, now documented at the call site:** `f64_kind_from_class` maps `"Uint8Array"` and `"Uint8ClampedArray"` to a typed-array kind, but the checked load it feeds reads the length at `handle + 0` and elements at `handle + 16`. Perry's `new Uint8Array(n)` is buffer-backed (length at `data - 8`), so handing that path one of those receivers makes every index compare out of bounds and silently yields `undefined` — I hit exactly that (`checksum:0` instead of `5760000`) while writing this, and backed it out. Nothing reaches it today; `Int32Array`/`Float64Array` and `Uint8Array` parameters are unaffected.

An earlier attempt to also make the kernel read `K[ky+1][kx+1]` a native multiply is **not** in this PR: the flat-const rodata lowering is not what actually runs for that shape, so claiming the value is a canonical raw double would have been unsound, and the residual coercions it added measured net-negative (475 → 518 ms).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved numeric handling for `Uint8Array` and `Buffer` byte reads, including arithmetic and compound assignments.
  * Out-of-bounds byte reads now consistently produce `NaN` when used in numeric operations.
  * Improved safety checks for calculated indices when buffer lengths are known.
* **Tests**
  * Added coverage for numeric byte accumulation and dynamic values introduced through string writes.
* **Documentation**
  * Documented supported numeric-lowering scenarios, verification results, and remaining index-proof limitations.

### DIFF
--- a/changelog.d/typed-array-byte-read-numeric-proofs.md
+++ b/changelog.d/typed-array-byte-read-numeric-proofs.md
@@ -1,0 +1,52 @@
+Two numeric-lowering fixes for byte-array reads, found while working
+`benchmarks/suite/bench_int_arithmetic.ts` (7.7× Node before this change; 6.2×
+after — the rest is a separate index-shape gap, noted below).
+
+**`acc += px[i]` on a `Uint8Array` no longer takes the dynamic add.** The HIR
+lowers `px[i]` on a binding it already knows is a `Uint8Array`/`Buffer` to the
+dedicated `Uint8ArrayGet` node, but the number-by-construction fixpoint's
+number-or-`undefined` view-read rule — and the not-BigInt predicate beside it —
+only recognised the `IndexGet` spelling. The accumulator was therefore never
+admitted, and every `+=` lowered through `js_dynamic_string_or_number_add` with
+`acc` kept in a GC-rooted shadow slot.
+
+That fix has a correctness half. Once the add is a raw `fadd`, an out-of-bounds
+read's NaN-boxed `undefined` survives it — IEEE arithmetic preserves the NaN
+payload — and the accumulator reads back as `undefined` where the spec says
+`NaN`. Number-context lowering now canonicalizes the value first, preferring the
+guarded inline load (which sinks the NaN into its own out-of-bounds arm) and
+otherwise emitting one compare plus a select, never a call. The `undefined` box
+is the only non-double this node can produce, so the test is exact.
+
+**Compound buffer indices can now prove their bounds.** `bounds_for_buffer_access_width`
+proved only a single index local carrying a bounded-pair fact, or a constant, so
+an index like `i * 2 + 1` or `y * 30 + x` fell back to a per-element
+`js_uint8array_index_get_value` call. The interval analysis can bound those: every
+leaf is a counter with a range fact or a compile-time constant, and `int_range_expr`
+composes them with checked arithmetic, answering `None` as soon as a leaf is
+unknown or a step overflows. When the whole interval fits inside a *constant*
+buffer length, the access is in bounds on every iteration. A non-constant length
+still declines — there is nothing to compare the interval against.
+
+Measured on an idle Mac mini, min of three, self-timed: `bench_int_arithmetic`
+475 → 395 ms (Node 64); a `pixels[y * SIZE + x]` variant 636 → 356 ms; a probe
+whose reads all qualify drops from 54 per-element helper calls to none.
+
+Verified byte-identical to Node on two differentials that are part of neither
+suite: eight cases around the byte read itself (in-bounds accumulate, an
+out-of-bounds read inside `+=`, an out-of-bounds read as a value, string
+concatenation, negative and fractional indices, a `Buffer` receiver, a mixed
+numeric/string accumulator) and seven around the new bounds proof (the exact
+kernel shape, an interval touching the last element, a counter mutated inside
+the loop body, a deliberately out-of-range interval, a one-past-the-end read, a
+negative composite index, and a buffer whose length is not a compile-time
+constant).
+
+Left for a follow-up: `bench_int_arithmetic`'s own `(y + ky) * SIZE + (x + kx)`
+still declines the interval proof, so its 54 reads per pixel remain calls.
+Also worth knowing — `f64_kind_from_class` maps `"Uint8Array"` and
+`"Uint8ClampedArray"` to a typed-array kind, but the checked load it feeds reads
+the length at `handle + 0` and elements at `handle + 16`. Perry's
+`new Uint8Array(n)` is buffer-backed (length at `data - 8`), so handing that path
+one of those receivers makes every index compare out of bounds and silently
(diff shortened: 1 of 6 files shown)


proggeramlug:
**Local gate on the Linux box (perrymaster), commit `9895d9620`: clean.**

Suites: `2801 passed / 0 failed` (runtime), `1351 passed / 0 failed`, `284 passed / 0 failed` (native-proof regressions), `121 passed / 0 failed`. No `FAILED` or `panicked` anywhere in the run.

Lint: `Check formatting` PASS, `CI plan policy self-test + docs table freshness` PASS, `Gap snapshot checker self-test` PASS, `Platform-aware parity allowlist self-test` PASS. Ratchets recorded no ceiling raises — unrooted-local debt 576 → 576, raw-handle baseline 967 → 967 with 113 module ceilings unchanged, both against merge base `da56c4ada`.

The three FAIL steps are the environment, not the change: `Require a changelog.d/ fragment` dies on `repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/files: bad subst`, and both ratchet-vs-merge-base steps on `fatal: invalid refspec '${{ github.event.pull_request.base.sha }}'` — the workflow expressions only expand inside GitHub Actions. (A fragment *is* included: `changelog.d/typed-array-byte-read-numeric-proofs.md`.)

For the record, the first gate run caught a real one that the truncated output nearly hid — a rustfmt violation (a stray blank line in the appended collector tests). Fixed and re-gated; the run above is the fixed commit.

proggeramlug:

proggeramlug:
**Follow-up pushed (`4c9dc456a`), on top of your rebase — and it closes the gap this PR had listed as left for later.**

The leftover I described (`(y + ky) * SIZE + (x + kx)` declining the interval proof) turned out not to be about the index at all. `is_fresh_uint8array_length_expr` accepts a literal or a known-length local but nothing built from them, so `new Uint8Array(SIZE * SIZE)` was never classified as a freshly allocated owned buffer. With no classification the receiver gets **no buffer view**, and with no view every element read falls back to `js_uint8array_index_get_value` no matter how well its index is proven — which is why the interval proof looked inert here.

What isolated it: changing only that benchmark's allocation to a literal length, nothing else, took it from 395 → 168 ms with the calls gone.

The predicate asks whether the allocation's size is *fixed*, never what it is — `length_source_from_expr` resolves the value later with a `FnCtx` in hand and records no constant length when it cannot — so a sum, difference or product of the same leaves qualifies on exactly the argument that already admits either leaf alone.

**Re-measured against this branch's exact merge-base (`bd7e5f71c`), both binaries built in one run, interleaved on an idle Mac mini, min of five, self-timed:**

| | base | branch | node |
|---|---:|---:|---:|
| `bench_int_arithmetic` | 462 ms | **149 ms** | 62 ms |

7.5× Node → 2.4×. Checksums identical on both binaries.

The three changes only compound together: the fixpoint fix makes the accumulation native, the allocation fix gives the receiver a view, and the interval proof is what lets a compound index use it. Take any one away and the reads go back to calls.

Still verified on this base: both differentials byte-identical to Node (eight cases around the byte read, seven around the bounds proof), `Int32Array`/`Float64Array`/`Uint8Array`-parameter probes unchanged, and the probe whose interval genuinely exceeds its buffer still declines to the checked call rather than reading out of bounds. `perry-codegen`: 1834 tests across 30 suites, none failing; `cargo fmt --check` clean.

proggeramlug:
Gate re-run on the new tip `4c9dc456a` (not just the first commit): 7 suites green, 0 failures, formatting clean, no ratchet ceilings raised. The 3 FAIL steps are the same environmental ones as before — unexpanded `${{ github.* }}` expressions outside Actions (a changelog fragment *is* present).

PR #9147, opened 2026-08-30
perf(runtime): FastKeyHasher hashed integers a byte at a time — 7.8x on ShapeFacts keys, −10.8% composite

proggeramlug:
From the symbolized cc `--help` profile, where hashing was ~5.2% of samples across three entries. They had **two different root causes**, and the larger one was not a hasher-choice problem.

## 1. `FastKeyHasher` implemented only `write` (the real find — 2.14% entry)

Every integer field therefore fell into `Hasher`'s *default* `write_uN` → `write(&n.to_ne_bytes())` byte loop. FNV-1a is a strictly serial `xor`→`wrapping_mul` chain, so a `ShapeFacts` key (two `u64`, three `u32`, an enum discriminant) cost **~36 dependent multiplies to mix six words**. `shape_descriptor_ensure_with_holes` probes `ids_by_facts` with that key on **every shape publish** — every property add or delete that transitions a shape.

Fix: word-at-a-time `write_u*`/`write_i*` that **fold** (`(acc ^ word) * FNV_PRIME`) rather than overwrite, preserving the multi-field property `PtrHasher` deliberately lacks (the comment at `shapes.rs:255` warns that an overwriting hasher would collapse `ShapeFacts` to its last field). `finish()` now avalanches: a short key can run as few as one fold, and FNV-1a accumulates entropy in the high bits while hashbrown indexes on the low ones — the same hazard `mix()` already guards for `PtrHasher`, where dropping it once cost 455→830 ms per the in-tree comment.

Standalone: **11.79 → 1.51 ns (7.8×)**; bare `usize` 2.51 → 0.59 ns.

## 2. The SipHash maps (1.64% + 1.40% entries)

310 default-hasher maps exist in the runtime, so conversion was driven by call-site hotness, not blanket search-and-replace. Converted: three closure registries, two builtin-closure tables, five symbol tables, `ExoticExpandoTables::entries`. `SYMBOL_PROPERTY_ATTRS` took `FastKeyHashMap` rather than `PtrHashMap` — its key is two words and `PtrHasher` would collapse it to the second half.

**Deliberately NOT converted:** the inner `HashMap<String, f64>` / `HashSet<String>` of the closure tables keep SipHash — those keys are JS property names computable from program input (`fn[userSuppliedName] = 1`), which is the adversarial case `RandomState` exists for; only the outer pointer probe is what the profile implicates. Also skipped startup/rare-path maps (`TYPED_ARRAY_NO_EXTEND`, `TEMPLATE_RAW_MAP`, submodule singletons): converting unsampled code is diff risk for no measured gain.

## Bench (quiet Mac mini, load 1.6; interleaved A/B from one stash/unstash pair, 5 reps)

| phase | before | after | delta |
|---|---|---|---|
| shapes | 1538 | 1411 | −8.3% |
| **varied (control)** | 173 | 173 | **0.0%** |
| closure | 218 | 179 | −17.9% |
| symbol | 180 | 123 | −31.7% |
| expando | 56 | 46 | −17.9% |
| **typedarray (control)** | 19 | 19 | **0.0%** |
| **total** | **2189** | **1952** | **−10.8%** |

Spread under 3%, well below the effect; the two flat phases are honest controls on paths this PR does not touch.

**Honest caveat:** the `shapes` phase moved only 8.3% despite the 7.8× hash win — that phase is allocation-dominated. The hash win is real but is ~2% of a real workload, consistent with the profile's 2.14%, not a headline number.

## Iteration-order audit

Every map touched **plus every pre-existing `FastKeyHashMap`** (their order also shifts, because `finish()` changed): all iteration is GC rekeying, `retain` predicates, `.any()` booleans, or collection into a `BTreeSet`. Observable JS key order comes from `Vec` values, never map order. One flagged case (`closure_dynamic_props_snapshot`) was inverted on inspection — it already sorts explicitly with a comment that HashMap order is non-deterministic.

## Validation

Runtime suite **2828 passed / 0 failed**, run to completion; fmt clean; `cargo check` warning-free. Four new tests pin the fast path: that `write_u64` does not fall back to the byte loop, that folds do not collapse a composite key (order *and* prefix both matter), and bucket-spread for `ShapeFacts`-shaped and single-round pointer keys — the last turns red if the `finish()` avalanche is removed.

## Analysis deliberately not implemented

The remaining `ShapeFacts` cost is call *frequency*: `ensure_with_holes` takes a `RefCell::borrow_mut()` on the whole shape table and hashes on every publish, even when the identical shape was published one iteration earlier. A one-entry memo (last facts → id) would skip the hash entirely in the common loop case, but invalidation must be correct across descriptor retirement *and* GC keys-array moves — a redesign, not a contained win. Caching the hash on the descriptor does not help: `ensure` builds facts from raw caller arguments with no descriptor in hand.

## Summary by CodeRabbit

* **Performance**
  * Improved runtime hashing for numeric, pointer-based, and composite keys.
  * Optimized internal lookups for closures, symbols, expandable objects, and built-in functions.
  * Improved hash distribution for more consistent lookup performance.

* **Reliability**
  * Added coverage validating integer hashing, pointer-key distribution, and composite-key preservation.
  * No user-facing behavior or API changes.

### DIFF
--- a/crates/perry-runtime/src/closure/dynamic_props.rs
+++ b/crates/perry-runtime/src/closure/dynamic_props.rs
@@ -2,15 +2,28 @@
 //! and the closure-magic-tag pointer predicate.
 
 use super::*;
+use crate::fast_hash::{new_ptr_hash_map, PtrHashMap};
 use std::collections::{HashMap, HashSet};
 use std::sync::{Mutex, OnceLock};
 
 per_test_global! {
-    static CLOSURE_PROPS: OnceLock<Mutex<HashMap<usize, HashMap<String, f64>>>> = OnceLock::new();
+    /// OUTER key is a closure heap address, probed on every dynamic property
+    /// get/set/delete on a function object, so it takes `PtrHasher` for the
+    /// same reason as the other pointer-keyed registries (#8125): a raw
+    /// address is already well distributed and no external input reaches it.
+    ///
+    /// The INNER `HashMap<String, f64>` deliberately keeps std's SipHash. Its
+    /// keys are JS property names, and unlike the descriptor side tables'
+    /// program-identifier keys these can be computed at runtime from program
+    /// input (`fn[userSuppliedName] = 1`), which is exactly the adversarial
+    /// case `RandomState` exists to defend. It is also not the half the
+    /// profile implicates -- `hash_one::<&usize>` is the outer probe.
+    static CLOSURE_PROPS: OnceLock<Mutex<PtrHashMap<usize, HashMap<String, f64>>>> =
+        OnceLock::new();
 }
 
-fn get_closure_props() -> &'static Mutex<HashMap<usize, HashMap<String, f64>>> {
-    CLOSURE_PROPS.get_or_init(|| Mutex::new(HashMap::new()))
+fn get_closure_props() -> &'static Mutex<PtrHashMap<usize, HashMap<String, f64>>> {
+    CLOSURE_PROPS.get_or_init(|| Mutex::new(new_ptr_hash_map()))
 }
 
 per_test_global! {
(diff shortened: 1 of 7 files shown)


proggeramlug:
Merged.

Changing a hasher is one of those edits where the suites can be green and the observable behaviour still moves, so I probed the thing a hash change can actually leak: **property enumeration order**. 16 shapes against node v26.5.1 — integer-then-insertion ordering with interleaved numeric keys, reverse-inserted keys, delete-then-re-add, 500-key and 300-key objects (forcing rehash), 30 symbols, symbol/string separation in `Object.keys` vs `getOwnPropertySymbols`, `for…in` across a prototype chain, array and function expandos, and the well-known symbols (`Symbol.iterator`, `hasInstance`, `toPrimitive`).

**15 of 16 byte-identical, and the 16th is identical to main** — so this change is order-neutral, which is the property that matters.

That 16th is a real pre-existing bug I've filed as #9148, since the probe surfaced it:

```ts
const f: any = function () {}; f.tag = 1; f.other = 2;
Object.keys(f)   // node: ["tag","other"]   perry: ["other","tag"]
```

Function expandos come back in the wrong order. Plain objects, arrays and 500-key objects are all correctly ordered, so it is specific to the function-object expando path — not something this PR introduced or touched.

Validation: runtime **2837 passed** (exit 0, 0 abort markers), perry --bins 1066, fmt clean, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped. `git diff origin/main --diff-filter=D` empty.

I did not independently reproduce the 7.8x / −10.8% figures; what I verified is that the new hasher does not perturb observable ordering.

PR #9149, opened 2026-08-30
perf(hir): hoist a loop-invariant property array out of counted loops (20.6 → 0.50 ns, now beats node)

proggeramlug:
## What

`for (let i = 0; i < holder.arr.length; i++) … holder.arr[i] …` re-did the property lookup on every iteration, and — worse — never entered the packed-array machinery at all, because that matcher requires the array expression to be a bare local. Writing the hoist by hand (`const a = holder.arr;`) was worth 27×, which told us the ceiling before any code was written. This HIR pass reaches it automatically.

## Numbers

Quiet Mac mini, node v26.5.1, ns/op, medians of 3. **Both perry columns are the same binary**, switched with `PERRY_LOOP_PROPERTY_HOIST`, so this is an A/B of the pass and not of two builds:

| receiver | perry off | perry on | node |
|---|---|---|---|
| module-global `holder.arr[i]` | 20.57 | **0.50** | 0.55 |
| local object receiver | 3.46 | **0.46** | 0.55 |
| hand-hoisted (control) | 0.47 | 0.50 | 0.58 |
| parameter receiver | 6.86 | 6.87 | 0.55 |
| captured receiver | 20.30 | 20.31 | 0.54 |

The module-global row goes from 37× slower than node to faster than node, and the local-receiver row from 6.3× slower to faster. The control row is flat, which is the point: it was already hoisted by hand, so the pass has nothing to do there.

The bottom two rows are deliberately untouched — a parameter or an aliased capture is not a `const x = { … }` binding, so the data-property proof below does not cover them. Both are follow-ups; the captured row looks reachable by propagating the proof through `const` aliases, while a parameter needs a real runtime guard, since an annotated `{ arr: number[] }` can be satisfied by an object with a getter.

## Why it is sound

Three checks, all made before rewriting:

1. **The property is a data field.** Only receivers bound by `const x = { … }` whose initializer lowered to a closed-shape record class qualify; `is_closed_shape` already rejects getters and setters, so reading such a field cannot run user code.
2. **Nothing can rebind the receiver.** Any assignment to it anywhere in the loop refuses the rewrite. This one cannot be delegated to a runtime check: two objects from the same literal share a shape, so a shape guard would happily accept the stale array.
3. **Nothing can write the property.** Calls, closures, `new`, property writes and index writes all refuse, as does any construct the scan does not positively recognise.

The check that matters most is #1, and it is keyed on the **initializer**, not the binding's type. Keying it on the type is both non-functional and unsound, in opposite directions: `const holder = { arr, n }` infers as a structural `Object(ObjectType)` rather than `Named("__AnonShape_…")`, so a type check never fires at all — the pass shipped as a silent no-op in my first build — while an *annotated* structural object type can still be backed by an accessor, so where a type check did fire it could be wrong. The initializer is the only place the accessor question is actually settled.

Doing this in HIR rather than codegen dissolves two hazards by construction rather than by argument: the hoisted value is an ordinary `Stmt::Let`, so it is GC-tracked exactly like any other local — no cached raw pointer and no rooting question — and the transform is invisible to the later loop-cloning passes.

## Tests

`crates/perry/tests/loop_property_array_hoist.rs` compiles each program twice, hoist on and off, and asserts both print the same thing; running the kill-switch build against the same expectation is what makes it a test of equivalence rather than of the hoisted path alone. Both builds run under `PERRY_GC_FORCE_EVACUATE` / `PERRY_GC_VERIFY_EVACUATION`, since "the temp is GC-tracked like any other local" deserves proof.

Most cases pin **refusals**, which is the half a regression would break silently: receiver rebound mid-loop to a same-shaped object (510, not 60), a call that overwrites the property, a getter receiver whose five invocations must all survive, and a direct write to the property. Plus an array grown during iteration, nested loops, string elements, and an empty array. Every expected value was checked against node first — one of them was wrong when I wrote it by hand.

## Binary size

`size(1)` `.text` on the benchmark: 10,963,348 → 10,963,028, i.e. **−320 bytes (−0.003%)** across 2 hoisted sites, ≈ −160 bytes per site. The pass replaces a per-iteration by-name lookup and its IC-miss path with a single load, so it removes code rather than adding it.

## Kill switch

`PERRY_LOOP_PROPERTY_HOIST=0` restores the previous lowering.

## Gates

`-D warnings` clean; perry-hir 587/0; perry-codegen 1835/0; perry-runtime lib 2822/0 (`--test-threads=1`); shape-descriptor census, address-classification audit, file-size and raw-handle-debt lints all pass with no ceiling raised. Integration: the new suite 7/7, plus `issue_8655_array_subclass_indexing` 2/2, `issue_8690_loop_versioned_arraylike` 3/3 and `issue_8897_field_push_writeback` 3/3 — the three suites that own the loop and array-admission paths this pass feeds.

## Follow-ups

- A `const` alias (`const h = holder`) inherits the proof, which also reaches receivers read inside a closure since the capture keeps the same `LocalId`; that covers the captured row above.
- The safety scan currently refuses any loop containing a nested loop or a `return`, both of which recurse soundly.
- A parameter receiver needs an actual runtime guard, since `{ arr: number[] }` can be satisfied by an object with a getter.

## Summary by CodeRabbit

* **Performance**
  * Improved counted loops that repeatedly access stable array properties by reusing the property value when it is safe to do so.
  * Automatically avoids the optimization when code could change the receiver or property, including getters, writes, and function calls.
  * Added an option to disable this optimization when needed.

* **Bug Fixes**
  * Preserved correct behavior for nested loops, changing arrays, empty arrays, and string elements.

* **Tests**
  * Added coverage verifying identical results with optimization enabled and disabled, including memory-management scenarios.

### DIFF
--- a/crates/perry-hir/src/destructuring/var_decl.rs
+++ b/crates/perry-hir/src/destructuring/var_decl.rs
@@ -328,6 +328,19 @@ pub(crate) fn lower_var_decl_with_destructuring(
                 None if !is_var_decl => Some(Expr::Undefined),
                 other => other,
             };
+            // Remember `const x = { … }` bindings whose initializer became a
+            // closed-shape record class. That is the only proof available that
+            // every property of `x` is a data field rather than an accessor —
+            // `is_closed_shape` rejects getters and setters — and the
+            // loop-invariant property hoist refuses to fire without it.
+            if !mutable {
+                if let Some(Expr::New { class_name, .. }) = init.as_ref() {
+                    if class_name.starts_with("__AnonShape_") {
+                        ctx.closed_shape_literal_locals
+                            .insert(id, class_name.clone());
+                    }
+                }
+            }
             result.push(Stmt::Let {
                 id,
                 name,
--- a/crates/perry-hir/src/lower/context.rs
+++ b/crates/perry-hir/src/lower/context.rs
@@ -201,6 +201,7 @@ impl LoweringContext {
             mixin_funcs: HashMap::new(),
             anon_shape_classes: HashMap::new(),
             anon_shape_fields: HashMap::new(),
+            closed_shape_literal_locals: HashMap::new(),
             prefer_exported_method_shape_seed: false,
             forward_class_names: std::collections::HashSet::new(),
             forward_class_decl_depth: std::collections::HashMap::new(),
@@ -985,6 +986,10 @@ impl LoweringContext {
 
     pub(crate) fn lookup_local_type(&self, name: &str) -> Option<&Type> {
         self.locals.lookup_type(name)
+    }
+
+    pub(crate) fn lookup_local_type_by_id(&self, id: LocalId) -> Option<&Type> {
+        self.locals.lookup_type_by_id(id)
     }
 
     pub(crate) fn lookup_func(&self, name: &str) -> Option<FuncId> {
--- a/crates/perry-hir/src/lower/locals.rs
+++ b/crates/perry-hir/src/lower/locals.rs
@@ -81,6 +81,16 @@ impl Locals {
     }
 
     /// `Type` of the innermost binding named `name`, if any. O(1).
+    /// Type of the innermost binding carrying `id`. Needed when a
+    /// transform holds a `LocalId` from already-lowered HIR and has no name.
+    pub(crate) fn lookup_type_by_id(&self, id: LocalId) -> Option<&Type> {
+        self.entries
+            .iter()
+            .rev()
+            .find(|(_, entry_id, _)| *entry_id == id)
+            .map(|(_, _, ty)| ty)
+    }
+
     pub(crate) fn lookup_type(&self, name: &str) -> Option<&Type> {
         self.lookup_index(name).map(|i| &self.entries[i].2)
     }
--- a/crates/perry-hir/src/lower/lowering_context.rs
+++ b/crates/perry-hir/src/lower/lowering_context.rs
@@ -776,6 +776,15 @@ pub struct LoweringContext {
     /// (e.g. recognizing a bundled mysql2 `createPool(config)` by its option
     /// names) recovers them here.
     pub(crate) anon_shape_fields: HashMap<String, Vec<String>>,
+    /// `const x = { … }` bindings whose initializer lowered to a closed-shape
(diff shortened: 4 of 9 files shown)


proggeramlug:
Merged, plus one commit from me: `PERRY_LOOP_PROPERTY_HOIST` wasn't registered in `BUILD_CACHE_ENV_VARS`, so `codegen_env_vars_are_build_cache_inputs` would have gone red. It gates two different emitted loop bodies, so it's a cache key.

**A hoist is only safe if the receiver can't change under it**, so that's what I probed — 18 shapes against node v26.5.1:

| | shape | node | result |
|---|---|---|---|
| 10, 11 | **getter** receiver — must re-evaluate every iteration | `[6,3]` (getter called 3×) | ✓ hoist correctly declines |
| 12 | **Proxy** receiver with a `get` trap | `[6,3]` (trap fires 3×) | ✓ declines |
| 3 | `o.a = [9,9,9,9]` reassigned mid-loop | `21` | ✓ |
| 4 | `o.a.push(99)` mid-loop | `[10,5]` | ✓ |
| 14 | prototype-provided array, then shadowed by an own `o.a` mid-loop | `15` | ✓ |
| 6, 7, 8, 9 | alias, closure capture, nested loops, per-iteration capture | — | ✓ |
| 15, 16 | OOB reads and holes | `[1,2,3,null,null]`, `1` | ✓ |
| 17, 18 | `continue`/`break`, and a throw out of the loop | `7`, `3` | ✓ |

Cases 10–12 are the ones that would expose an over-eager hoist — a cached receiver would call the getter once instead of three times, and that's observable. All three keep node's call counts.

**17 of 18 identical; the 18th is identical on main.** That one is `delete o.a` mid-loop then `o.a[i]`: node throws `TypeError`, perry returns `"6"`. I checked it three ways — main behaves the same, and the PR's own kill switch (`PERRY_LOOP_PROPERTY_HOIST=0`) doesn't change it either — so it's a pre-existing gap in `delete`-then-read, not something the hoist introduced. Filing separately.

Validation: hir 365 passed, codegen 1356, runtime 2840 (exit 0, 0 abort markers), perry --bins 1066, fmt clean, `run_lint_gates.sh` all 60 gates passed; `--diff-filter=D` empty.

#9153 builds directly on this and is next.

Issue #9151, opened 2026-08-30
Counted loops lose packed-array admission to any abrupt statement — break/continue/return/throw all 6x, even when never taken


proggeramlug:
## Summary

A counted loop over a packed array loses its fast path when the body contains any **abrupt-completion** statement — `break`, `continue`, `return` or `throw` — whether or not it is ever taken. All four cost the same 6x against perry's own straight-line loop, and ~8.5x against node.

A plain conditional is not the trigger: an `if` whose body only assigns is fully fast. And `continue` does not leave the loop at all, which is the datum that rules out "early exit" as the explanation.

## Repro

```ts
const arr: number[] = [];
for (let i = 0; i < 512; i++) arr.push(i * 3);
const R = 40_000;
let sink = 0;

// body shapes, all inside `for (let r = 0; r < R; r++) for (let i = 0; i < arr.length; i++) { … }`
//   base        l = arr[i];
//   break       l = arr[i]; if (l < 0) break;
//   continue    l = arr[i]; if (l < 0) continue;
//   return      l = arr[i]; if (l < 0) return -1;
//   throw       l = arr[i]; if (l < 0) throw new Error("x");
//   if_assign   l = arr[i]; if (l < 0) { l = 0; }
//   if_sink     l = arr[i]; if (l < 0) { sink = 1; }
//   outer_break l = arr[i];   … with `if (l < 0) break;` in the OUTER loop instead
```

`arr` holds only non-negative values, so no branch is ever taken in any variant.

## Measurement

Quiet Mac mini, node v26.5.1, ns/op, best of 3:

| inner-loop body | perry | node |
|---|---|---|
| base (straight line) | 0.78 | 0.55 |
| `if (l < 0) { l = 0; }` | 0.46 | 0.56 |
| `if (l < 0) { sink = 1; }` | 0.51 | 0.54 |
| `break` in the **outer** loop | 0.50 | 0.54 |
| `if (l < 0) break;` | 4.74 | 0.54 |
| `if (l < 0) continue;` | 4.75 | 0.55 |
| `if (l < 0) return -1;` | 4.76 | 0.56 |
| `if (l < 0) throw …;` | 4.76 | 0.56 |

The array is a bare local throughout, so no receiver or property-lookup effect is involved. The four abrupt forms landing on the same number to within 0.02 ns suggests one predicate rejecting all four rather than four separate causes.

Node is flat across every variant, so the fast path is not inherently incompatible with abrupt control flow.

## Where it is

[...]

PR #9153, opened 2026-08-30
perf(hir): widen the property-array hoist to const aliases, captures and nested loops (20.5 → 0.47 ns)

proggeramlug:
Follows #9149, which deliberately left three receiver shapes alone. One rule covers all three.

## Numbers

Quiet Mac mini, node v26.5.1, ns/op, medians of 3. Both perry columns are **the same binary**, switched with `PERRY_LOOP_PROPERTY_HOIST`:

| receiver | perry off | perry on | node |
|---|---|---|---|
| captured in an arrow | 20.35 | **0.50** | 0.54 |
| `const` alias of the receiver | 20.49 | **0.47** | 0.58 |
| outer loop containing a nested loop | 20.58 | **0.47** | 0.55 |
| loop with an early `return` | 21.59 | 3.59 | 0.57 |

With #9149 this takes the original benchmark to four of five rows beating node. The remaining one is a parameter receiver, which needs a genuine runtime guard rather than a static proof and is not attempted here.

## The rule

A `const` alias inherits the data-field proof. `const h = holder` where `holder` is itself a `const` closed-shape literal binding: neither name can ever be rebound, so both always denote the object the literal created, and everything #9149 argued about `holder` holds verbatim for `h`.

That one rule also reaches **receivers read inside a closure**, at no extra cost and with no new machinery. A capture keeps the same `LocalId`, so the existing rewrite already matches the body — confirmed by dumping HIR rather than assumed:

```
Let { id: 7, name: "h", mutable: false, init: Some(LocalGet(1)) }
Closure { captures: [7], … body: … PropertyGet { object: LocalGet(7), property: "arr" } … }
```

An alias of a `let` is not admitted, because only `const` bindings ever enter the registry. There is a test for it: following a mutable source is exactly how this rule would turn unsound.

## Nested loops and early exits

Refusing any loop whose body contained a nested loop was pure conservatism — and it excluded `m.rows[i]` with an inner loop over the row, which is the shape the pass exists for. A nested loop is safe on precisely the same terms as any other statement: it may not rebind the receiver and may not call. `return` and `throw` are likewise fine, since the hoisted `Let` is evaluated before the loop either way and leaving early only skips reads.

Every arm the scan admits is also handled by the rewriter. That pairing is load-bearing rather than tidiness: an admitted-but-unrewritten statement would silently keep its per-iteration lookup, so the two match arms are meant to be read side by side.

## The early-`return` row, honestly

3.59 is a 6x improvement but still 6.3x node, and the residual is not this pass. Isolating it on a **bare local** array, with no property lookup and no hoisting involved at all:

| inner-loop body | perry | node |
|---|---|---|
| straight line | 0.78 | 0.55 |
| `if (l < 0) { l = 0; }` | 0.46 | 0.56 |
| `if (l < 0) break;` | 4.74 | 0.54 |
| `if (l < 0) continue;` | 4.75 | 0.55 |
| `if (l < 0) return -1;` | 4.76 | 0.56 |

A plain conditional is free; any abrupt statement costs 6x, including `continue`, which never leaves the loop. Filed as #9151 with the predicate located (`stmt_is_packed_f64_loop_safe`, `loops.rs:5011`, one rejection arm covering all of them).

## Binary size

`size(1)` `.text`: 10,974,932 → 10,973,396, i.e. **−1536 bytes (−0.014%)**, against −320 bytes for #9149 alone. The wider the pass reaches, the more per-iteration lookups it deletes, so it keeps removing code rather than adding it.

## Gates

`-D warnings` clean; perry-hir 591/0; perry-codegen 1839/0; perry-runtime lib 2837/0 (`--test-threads=1`); census, address-classification, file-size and raw-handle-debt lints all pass with none raised. Integration: the hoist suite 11/11 (4 new), plus `issue_8655_array_subclass_indexing` 2/2, `issue_8690_loop_versioned_arraylike` 3/3 and `issue_8897_field_push_writeback` 3/3. The battery ran before a rebase onto #9137, which is unrelated (object-delete ICs) and does not touch HIR lowering.

## Summary by CodeRabbit

- **Performance**
  - Improved performance for eligible loops that repeatedly read unchanged property arrays by reusing the array reference.
  - Optimization supports nested loops, constant aliases, closure captures, and loops where the array grows during execution.

- **Reliability**
  - Added safeguards to preserve existing behavior when properties may be reassigned, written to, accessed through getters, or affected by function calls.
  - Added coverage confirming optimized and unoptimized execution produce identical results.

### DIFF
--- a/crates/perry-hir/src/destructuring/var_decl.rs
+++ b/crates/perry-hir/src/destructuring/var_decl.rs
@@ -333,12 +333,29 @@ pub(crate) fn lower_var_decl_with_destructuring(
             // every property of `x` is a data field rather than an accessor —
             // `is_closed_shape` rejects getters and setters — and the
             // loop-invariant property hoist refuses to fire without it.
+            //
+            // A `const` alias of such a binding inherits the proof: neither
+            // name can ever be rebound, so both always denote the object the
+            // literal created. This is what lets the hoist reach receivers
+            // read through an alias, including one captured by a closure —
+            // the capture keeps the same `LocalId`, so no extra plumbing is
+            // needed once the alias is recorded.
             if !mutable {
-                if let Some(Expr::New { class_name, .. }) = init.as_ref() {
-                    if class_name.starts_with("__AnonShape_") {
+                match init.as_ref() {
+                    Some(Expr::New { class_name, .. })
+                        if class_name.starts_with("__AnonShape_") =>
+                    {
                         ctx.closed_shape_literal_locals
                             .insert(id, class_name.clone());
                     }
+                    Some(Expr::LocalGet(source)) => {
+                        if let Some(class_name) =
+                            ctx.closed_shape_literal_locals.get(source).cloned()
+                        {
+                            ctx.closed_shape_literal_locals.insert(id, class_name);
+                        }
+                    }
+                    _ => {}
                 }
             }
             result.push(Stmt::Let {
--- a/crates/perry-hir/src/lower/property_array_hoist.rs
+++ b/crates/perry-hir/src/lower/property_array_hoist.rs
@@ -200,7 +200,40 @@ fn stmt_is_hoist_safe(stmt: &Stmt, recv_id: u32) -> bool {
                     .as_ref()
                     .is_none_or(|b| b.iter().all(|s| stmt_is_hoist_safe(s, recv_id)))
         }
+        // Nested loops are the common case this pass exists for — `m.rows[i]`
+        // in an outer loop with an inner loop over the row — so recurse rather
+        // than refuse. A nested loop is safe on exactly the same terms: it may
+        // not rebind the receiver and may not call anything.
+        Stmt::While { condition, body } | Stmt::DoWhile { body, condition } => {
+            expr_is_hoist_safe(condition, recv_id)
+                && body.iter().all(|s| stmt_is_hoist_safe(s, recv_id))
+        }
+        Stmt::For {
+            init,
+            condition,
+            update,
+            body,
+        } => {
+            init.as_ref()
+                .is_none_or(|s| stmt_is_hoist_safe(s, recv_id))
+                && condition
+                    .as_ref()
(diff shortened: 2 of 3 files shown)


proggeramlug:
Stacked on #9149 — its two commits show here until it merges; the widening itself is the top commit.

proggeramlug:
Merged, on top of #9149 which I merged just before this. Its branch carried #9149's commits, so I applied only its own (`widen the property-array hoist to aliases, captures and nested loops`) onto main — 1 commit, `--diff-filter=D` empty.

Widening a hoist widens its exposure, so I re-ran the same 18-shape safety probe I used on #9149 rather than assuming the base validation carried over. The cases that matter for *this* PR are the three it newly admits:

| | shape | node | result |
|---|---|---|---|
| 6 | const alias (`const p = o; … p.a[i]`) | `6` | ✓ |
| 7, 9 | closure capture of the receiver; per-iteration capture in a loop | `6`, `[1,2,3]` | ✓ |
| 8 | nested loops over `o.a[i][j]` | `10` | ✓ |

And the escape hatches still hold under the wider admission — which is the thing I'd actually worry about, since an alias or capture makes it easier to lose track of who can mutate the receiver:

| 10, 11 | getter receiver, called 3× | `[6,3]` — hoist declines ✓ |
| 12 | Proxy receiver, trap fires 3× | `[6,3]` — declines ✓ |
| 3, 4, 14 | receiver reassigned / `push`ed / own-shadowed mid-loop | ✓ |

**1 of 18 differs from node, and it is 1 of 18 on main too** — `delete o.a` mid-loop then `o.a[i]` (node throws `TypeError`, perry returns `"6"`). I confirmed on #9149 that this is unaffected by the kill switch, so it predates the hoist entirely.

Validation: hir 365 passed, codegen 1356, runtime 2840 (exit 0, 0 abort markers), fmt clean, `run_lint_gates.sh` all 60 gates passed; 2 CI-only skipped. The `PERRY_LOOP_PROPERTY_HOIST` build-cache registration landed with #9149.

PR #9154, opened 2026-08-30
perf(codegen): packed loops keep the fast path across break/continue/return — 4.75 → 0.58 ns (#9151)

### LINKED ISSUE #9151: Counted loops lose packed-array admission to any abrupt statement — break/continue/return/throw all 6x, even when never taken

proggeramlug:
Fixes #9151.

`stmt_is_packed_f64_loop_safe` (`loops.rs:5011`) lumped every abrupt statement into a single rejection arm, so a counted loop over a packed array fell to the generic path the moment its body *could* break, continue or return — whether or not the exit was ever taken. Find-first, any/all and early-out scans are exactly that shape, and they were paying 6x for an exit they usually never take.

That all four forms measured identically at 4.75 ns was the tell that this was one gate rather than four.

## Numbers

Quiet Mac mini, node v26.5.1, ns/op, best of 3. Both perry columns are **the same binary**, switched with `PERRY_PACKED_LOOP_ABRUPT`:

| inner-loop body | perry off | perry on | node |
|---|---|---|---|
| `if (l < 0) break;` | 4.75 | **0.58** | 0.54 |
| `if (l < 0) return -1;` | 4.76 | **0.58** | 0.56 |
| `if (l < 0) continue;` | 4.76 | **0.50** | 0.55 |
| `if (l < 0) throw …;` | 4.75 | 4.75 | 0.56 |
| straight-line body | 0.75 | 0.76 | 0.55 |
| `break` in the outer loop | 0.46 | 0.46 | 0.54 |

The array is a bare local, so no receiver or property-lookup effect is involved. `throw` is deliberately left out: the thrown value is normally constructed, and that is a call in the loop body.

## Why it is sound

The relaxation the caller documents already covers these statements:

> A call-free READ body earns the same relaxation the store arm below documents: the entry guard revalidates the actual receiver/layout and the matched body cannot call out or invalidate it … A wrong static hint is one failed guard -> slow clone, never a wrong answer.

`break`, `continue` and a bare `return` do not call out, do not touch the array, and do not invalidate the entry guard; an iteration that exits early simply performs fewer reads than the guard admitted. Two neighbours in this same file already answer `Preserves`/`true` for exactly this set — `stmt_array_length_effect` (`loops.rs:7439`) and `stmt_preserves_array_length` (`loops.rs:7872`). A returned expression is checked with the same `expr_is_packed_f64_loop_safe` as any other operand.

## Unlabeled only — the part that bit me

My first version admitted `LabeledBreak`/`LabeledContinue` too, and it was **a silent wrong answer**:

```ts
outer: for (let r = 0; r < 4; r++) {
  for (let i = 0; i < arr.length; i++) { s += arr[i]; if (arr[i] === 10 && r === 2) break outer; }
}
```

returned **4032** where node returns **4087** — the whole partial iteration was dropped. A labeled break targets an *enclosing* loop and must unwind past the one being analysed, which the fast clone does not do.

The restriction to unlabeled is principled rather than a patch: a nested loop is still rejected by this predicate, so a bare `break` or `continue` can only ever target the loop being analysed, and the clone's exit edge is the right destination.

## Tests — read `labeled_break_still_unwinds_the_outer_loop` first

`crates/perry/tests/packed_loop_abrupt_statements.rs` is the regression pin, and the labeled case above is in it explicitly for the next person who widens this predicate.

**Every fixture actually takes its exit**, at varying indices, with observable effects before it. That is the point rather than thoroughness for its own sake: all of my original timing probes used a never-taken exit, and every one of them passed the broken version. "The fast clone is entered" and "the fast clone is correct when it exits early" are different properties, and only the second one catches this.

Beyond the unit tests, two generated differentials against node:

- 336 cases — 4 array shapes (including single-element and empty) x 3 loop kinds (`for`/`while`/`do-while`) x 3 exit kinds x 4 guard predicates x 3 limits chosen to exit at the first index, the middle, and never — byte-identical to node with the switch both on and off.
- 54 labeled-break/continue cases across nested-loop trip counts and hit positions — byte-identical to node, i.e. the excluded form still lowers correctly.

## Binary size

`size(1)` `.text`: 10,949,332 → 10,949,844, i.e. **+512 bytes (+0.005%)** for 3 newly admitted loops, about 171 bytes per site. This one genuinely adds code — a fast clone is now emitted where none was — but at clone-per-admitted-loop rather than anything that scales with body size.

## Kill switch

`PERRY_PACKED_LOOP_ABRUPT=0` restores the previous behaviour.

## Summary by CodeRabbit

* **New Features**
  * Packed-array counted loops now support unlabeled `break`, `continue`, and safe `return` statements while preserving expected behavior.
  * Loop handling remains compatible with growing arrays and exception paths.
  * The packed-loop optimization can be disabled with `PERRY_PACKED_LOOP_ABRUPT=0`, `off`, or `false`.

* **Bug Fixes**
  * Improved correctness for loop exits, including observable actions performed before exiting.
  * Labeled loop exits and `throw` statements continue to use safe handling.

### DIFF
--- a/crates/perry-codegen/src/stmt/loops.rs
+++ b/crates/perry-codegen/src/stmt/loops.rs
@@ -4975,6 +4975,15 @@ fn local_is_u32_array(ctx: &FnCtx<'_>, local_id: u32) -> bool {
     )
 }
 
+/// `PERRY_PACKED_LOOP_ABRUPT=0` restores the pre-#9151 behaviour, where any
+/// abrupt statement kept the loop on the generic path.
+fn packed_loop_abrupt_enabled() -> bool {
+    !matches!(
+        std::env::var("PERRY_PACKED_LOOP_ABRUPT").as_deref(),
+        Ok("0") | Ok("off") | Ok("false")
+    )
+}
+
 fn stmt_is_packed_f64_loop_safe(
     ctx: &FnCtx<'_>,
     stmt: &Stmt,
@@ -5008,12 +5017,34 @@ fn stmt_is_packed_f64_loop_safe(
         // Conservative: a box release clears cells; keep it out of packed
         // f64 loop bodies (it never appears in one today).
         Stmt::ReleaseBoxes(_) => false,
-        Stmt::Return(_)
-        | Stmt::Throw(_)
-        | Stmt::Break
-        | Stmt::Continue
-        | Stmt::LabeledBreak(_)
+        // Leaving *this* loop early neither calls out nor touches the array, so
+        // the relaxation the caller documents still holds: the entry guard has
+        // already revalidated the receiver, and an iteration that exits simply
+        // performs fewer reads than the guard admitted. `stmt_array_length_effect`
+        // and `stmt_preserves_array_length` already answer `Preserves`/`true`
+        // for these two.
+        //
+        // Unlabeled only. A nested loop is rejected below, so a bare `break` or
+        // `continue` here can only target the loop being analysed, and the fast
+        // clone's exit edge is the right destination. A LABELED break targets an
+        // enclosing loop and must unwind past this one, which the clone does not
+        // do: admitting it made
+        //   outer: for (r…) { for (i…) { s += a[i]; if (a[i] === 10 && r === 2) break outer; } }
+        // return 4032 instead of 4087, silently dropping the partial iteration.
+        Stmt::Break | Stmt::Continue => packed_loop_abrupt_enabled(),
+        // Same argument, once the returned expression itself is safe — it is
+        // evaluated in the loop body like any other operand.
+        Stmt::Return(value) => {
+            packed_loop_abrupt_enabled()
+                && value.as_ref().is_none_or(|expr| {
+                    expr_is_packed_f64_loop_safe(ctx, expr, arr_id, counter_id)
+                })
+        }
+        // `throw` stays out: the thrown value is typically constructed
+        // (`throw new Error(…)`), which is a call in the loop body.
+        Stmt::LabeledBreak(_)
         | Stmt::LabeledContinue(_)
+        | Stmt::Throw(_)
         | Stmt::While { .. }
         | Stmt::DoWhile { .. }
         | Stmt::For { .. }
--- a/crates/perry-codegen/src/stmt/loops.rs
+++ b/crates/perry-codegen/src/stmt/loops.rs
@@ -5036,9 +5036,9 @@ fn stmt_is_packed_f64_loop_safe(
         // evaluated in the loop body like any other operand.
         Stmt::Return(value) => {
             packed_loop_abrupt_enabled()
-                && value.as_ref().is_none_or(|expr| {
-                    expr_is_packed_f64_loop_safe(ctx, expr, arr_id, counter_id)
-                })
+                && value
+                    .as_ref()
+                    .is_none_or(|expr| expr_is_packed_f64_loop_safe(ctx, expr, arr_id, counter_id))
         }
         // `throw` stays out: the thrown value is typically constructed
         // (`throw new Error(…)`), which is a call in the loop body.
--- a/crates/perry/src/commands/compile/build_cache.rs
+++ b/crates/perry/src/commands/compile/build_cache.rs
@@ -66,6 +66,10 @@ const BUILD_CACHE_ENV_VARS: &[&str] = &[
     // counted for-loops — on and off emit different loop bodies, so a cached
     // object from one must not serve the other.
     "PERRY_LOOP_PROPERTY_HOIST",
+    // #9154: gates keeping the packed fast path across break/continue/return —
+    // on and off emit different loop exits, so a cached object from one must
+    // not serve the other.
+    "PERRY_PACKED_LOOP_ABRUPT",
     // #9122: gates the shape-cache path for object literals whose methods
     // capture `this` — on and off emit different literal-birth sequences,
     // so a cached object from one must not serve the other.
--- a/crates/perry/tests/packed_loop_abrupt_statements.rs
+++ b/crates/perry/tests/packed_loop_abrupt_statements.rs
@@ -1,0 +1,164 @@
+//! Packed-array counted loops keep their fast path when the body contains an
+//! unlabeled `break`, `continue` or `return` (#9151).
+//!
+//! Every program here ACTUALLY takes its exit, at varying indices and with
+//! observable effects before it. That matters: a never-taken exit exercises
+//! only the admission decision, not the emitted clone, and the first cut of
+//! this change was wrong in a way only a taken exit could reveal — a labeled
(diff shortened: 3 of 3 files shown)


proggeramlug:
Merged, plus one commit: `PERRY_PACKED_LOOP_ABRUPT` wasn't in `BUILD_CACHE_ENV_VARS`. It gates two different loop-exit shapes, so it's a cache key — that gate has now caught seven PRs in this series, which probably argues for adding "register the knob" to the checklist when introducing a codegen kill switch.

**Keeping the fast path across abrupt exits means every exit now has to write back**, so that's what I probed — 19 shapes against node v26.5.1, byte-identical:

| | shape | node |
|---|---|---|
| 1–4 | `break`, `continue`, `return` from inside, and a `throw` caught outside | `55`, `992`, `28`, `45` |
| 5 | `throw` with a `finally` that mutates the accumulator afterwards | `THREW Error` |
| 6, 7 | labelled `continue outer` / `break outer` across nested packed loops, and a labelled `break L` | `532`, `78` |
| 8, 9 | **two** accumulators with the exit landing *between* their updates | `[120,15]`, `[780,78]` |
| B63…B129 | exits at the 63/64/65/128/129 boundaries — poll striding is 64 | exact |
| 11, 12 | exit on the very first and very last iteration | `0`, `1225` |
| 13 | count accumulator (`c++`) with an abrupt exit | `14` |
| 14 | `a.length = 15` mid-loop, then `break` later | `[105,15]` |

Cases 8 and 9 are the ones I most wanted: if the writeback were emitted per-exit rather than per-accumulator, an exit between two updates would strand the second one, and both come back exact. The 63/64/65 boundaries matter because a stride boundary is where an exit is most likely to land in a block whose epilogue differs.

Your own `packed_loop_abrupt_statements` integration suite passes too (7 tests) — worth noting it lives in `crates/perry/tests/`, so per-PR CI only runs it when the diff names it; I ran it explicitly.

Validation: codegen 1356 passed, runtime 2841 (exit 0, 0 abort markers), perry --bins 1066, fmt clean, `run_lint_gates.sh` all 60 gates passed; `--diff-filter=D` empty. I did not independently reproduce the 4.75 → 0.58 ns figure.

Issue #9158, opened 2026-08-30
codegen kill switches keep missing BUILD_CACHE_ENV_VARS — 7 PRs in one day


proggeramlug:
Seven PRs in the last day introduced a codegen kill switch and were red on `codegen_env_vars_are_build_cache_inputs` because the knob wasn't added to `BUILD_CACHE_ENV_VARS`:

| PR | knob |
|---|---|
| #9060 | `PERRY_PACKED_LOOP_NUMERIC_ACCUMULATOR` |
| #9071 | `PERRY_CALLEE_BINDING_RESOLUTION` |
| #9105 | `PERRY_CALL_DEVIRT` |
| #9122 | `PERRY_OBJECT_LITERAL_SHAPE_METHODS` |
| #9124 | `PERRY_METHOD_INLINE_PROBE` |
| #9149 | `PERRY_LOOP_PROPERTY_HOIST` |
| #9154 | `PERRY_PACKED_LOOP_ABRUPT` |

Every one was the same two-line fix, and in each case the author had done the harder part correctly — the knob itself, the kill-switch semantics, the tests. Seven identical misses by seven competent changes is a process gap, not seven oversights.

It also nearly escaped once: I merged #9122 with the registration prepared but not actually in the commit, which left `main` red until #9144. So the failure mode isn't only "PR is red", it's "PR is red in a way that's easy to wave through when you're fixing it yourself".

### Why it's easy to miss

The knob lives in `perry-codegen`; the registry lives in `crates/perry/src/commands/compile/build_cache.rs`. Nothing in the codegen file points at the registry, and the connection (a knob that changes emitted code must key the build cache, or a cached object from one setting silently serves the other) is real but non-obvious at the point of writing.

### Cheap fixes, roughly in order of effort

1. **Make the failure self-servicing.** The assertion already names the missing vars; have it also print the exact file and the anchor line to add them to, so the fix needs no archaeology.
2. **A line in `CLAUDE.md`** next to the existing codegen guidance: "a new `PERRY_*` knob read in codegen must be added to `BUILD_CACHE_ENV_VARS` (or `BUILD_CACHE_ENV_EXCLUSIONS` with a reason)."
3. **A doc comment on the reader helper.** Most of these knobs are read through a small `fn <name>_enabled()` with a `OnceLock`; a note in whatever that pattern's canonical example is would reach people at the moment they copy it.

[...]

PR #9159, opened 2026-08-30
perf(codegen): two-leaf dynamic + takes the guarded fadd too — 3.44 → 1.33 ns/add (#9157)

### LINKED ISSUE #9157: Numeric accumulator in a non-array loop emits a runtime call + write barrier per add — 10x vs node, no fadd emitted
## Summary

A numeric accumulator in an ordinary (non-array) loop compiles to a **runtime call per addition plus a GC write barrier**, with no floating-point add instruction emitted at all. `s += v`, where both operands are numbers and `v` is even explicitly annotated `const v: number`, costs 3.44 ns/op against node's 0.35 — a uniform **10x**.

This is not a property-access or call-boundary effect. It reproduces with no property read in the loop and no call in the source.

## Repro

```ts
const arr: number[] = [];
for (let i = 0; i < 7; i++) arr.push(1);
const seed: number = arr.length;   // 7, but not a literal, so nothing folds
const R = 2_000_000;

function fromLocal(): number {
  let s = 0;
  const v = seed;
  for (let i = 0; i < R; i++) { s += v; s += v; s += v; s += v; s += v; s += v; s += v; s += v; }
  return s;
}
```

The literal-initialized version of this loop is constant-folded away (0.000 ns/op), which is why the value must originate from a runtime source to be measurable at all.

## Measurement

Quiet Mac mini, node v26.5.1, ns/op per add, best of 3:

| variant | perry | node |
|---|---|---|
| `const v = leaf.v` (class field declared `number`) | 3.442 | 0.347 |
| `const v: number = leaf.v` (explicitly annotated) | 3.438 | 0.348 |
| `const v = seed` (plain typed local, **no property read in the loop**) | 3.440 | 0.348 |
| annotated property read inside the loop | 3.439 | 0.342 |

All four land on the same number, so neither the property read nor the annotation is the variable — the accumulate itself is.

## What is emitted

`objdump` of `fromLocal` (built with `PERRY_DEBUG_SYMBOLS=1`), one pair per `s += v`:

```
bl  <_js_dynamic_string_or_number_add>
bl  <_js_write_barrier_root_nanbox>
bl  <_js_dynamic_string_or_number_add>
bl  <_js_write_barrier_root_nanbox>
…
```

Two things stand out. The add goes through the **generic string-or-number** helper, so a value statically annotated `number` is still being tested for string-ness at runtime. And each store to the accumulator takes a **write barrier**, as though `s` were a rooted heap slot rather than a register-resident double. There is no `fadd` anywhere in the loop.

## Why the array case is fast

The same accumulate inside an admitted packed-f64 array loop is fine — `for (…) s += arr[i]` measures 0.95 ns/op, competitive with node. So perry already has the machinery to keep a numeric accumulator native; it appears to be reachable **only** via packed-array loop admission. A loop that does arithmetic without touching an array gets none of it.

Floating-point work that is provably f64 is also fine: `s *= v` with a float literal measures 1.250 in perry against node's 1.249, exactly FP-multiply latency with no overhead. So the deficiency is specific to values whose numeric type is not proven by their source.

## Why it matters

This is straight-line arithmetic, which is everywhere in real code and, unlike a loop optimization, has no admission criteria to satisfy. Measured against the campaign's cc numbers: loops are already competitive (perry beats or matches node on admitted array loops), while cc overall sits at ~2.46x node's instruction count. Arithmetic that costs a call plus a barrier per operation is a strong candidate for where that difference actually lives.

## Suggested direction

Reach the existing numeric lowering from a typed local rather than only from packed-array loop admission — a declared or inferred `number` should be enough to select the native add and drop the barrier. Related: the numeric-envelope proof work tracked in #5497.

## Found via

Decomposing an apparent 17x gap on property chains (`top.mid.leaf.v`), which turned out not to be about property access: hand-hoisting the entire chain out of the loop left the time unchanged at 3.44, because the residual was the eight `s += v` accumulates, not the reads.

proggeramlug:
Fixes #9157.

## What was wrong

A numeric accumulator in an ordinary loop compiled to a call per addition, with no `fadd` anywhere — even for `const v: number`:

```
bl  <_js_dynamic_string_or_number_add>
bl  <_js_write_barrier_root_nanbox>
```

The fast path it should have taken **already existed**. `lower_guarded_numeric_add` emits a `js_value_is_number` diamond: inline `fadd` on the hot arm, the dynamic helper on the cold arm, `phi` at the merge. It was simply gated to add-trees of three or more leaves, and `s += v` has two.

The threshold's rationale was written down, which is what made it checkable:

> At two leaves the guard merely moves the helper behind a branch; starting at three it can replace two or more helper calls with one shared tag check.

That accounting misses two things. On the hot arm the call becomes an inline `fadd`; and the operands stop going through `lower_rooted_dynamic_binary`, which roots them — so the accumulator was also living in a shadow-frame slot and paying a barrier per store.

## Numbers

Quiet Mac mini, node v26.5.1, ns/op **per add**, best of 3. Both perry columns are **the same binary**, switched with `PERRY_DYNAMIC_ADD_PAIR_GUARD`. In every row both operands hold numbers at runtime and neither is statically proven:

| operand source | perry off | perry on | node |
|---|---|---|---|
| class field declared `number` | 3.442 | **1.331** | 0.347 |
| explicitly annotated `const v: number` | 3.438 | **1.332** | 0.348 |
| plain typed local, no property read in the loop | 3.440 | **1.330** | 0.348 |
| annotated read inside the loop | 3.439 | **1.461** | 0.342 |

Disassembly confirms the mechanism: eight `fadd` where there were none, the eight helper calls now sitting in cold arms.

## No type is trusted

The guard is a runtime tag test, so this sits inside the documented policy rather than bending it — *"a static type selects a lowering, never an answer"* (`type_analysis/strings.rs`), with `docs/src/internals/local-binding-type-evidence.md` and `scripts/local_binding_type_audit.py` enforcing it.

The differential feeds a string laundered through `as any` into a number-annotated parameter, local and class field. Every one still concatenates, byte-identical to node, on both lowerings.

## What remains is representation, not dispatch

Worth stating plainly so the next person doesn't chase it as a dispatch problem:

| dependent add chain | perry | node |
|---|---|---|
| integer-valued | 1.332 | 0.346 |
| genuinely floating-point | 1.334 | 0.937 |

Perry is *identical* on both, because it is f64 either way. Node is 2.7x faster on integers and only 1.4x on floats — the difference between those two ratios is its small-integer representation. A dependent add chain is latency-bound, so most of the residual is that, which is a value-representation design question rather than a threshold.

## Tests

Five new integration tests pin the guard's semantics: a mid-loop **arm switch** (the same add site numeric on early iterations and string afterwards), a 36-case operand cross-product over number/string/bool/null/undefined, and lying annotations on a parameter, a local and a class field. Every expected value was checked against node first. Each runs twice, with the guard on and off, so it tests equivalence rather than one lowering.

Three existing codegen tests changed, and they were **not** simply made to pass — each pinned the old threshold or a block-local text search rather than a property worth keeping:

- `two_leaf_dynamic_add_stays_on_direct_dispatch` asserted the exact opposite of this change, on the stale cost model. Renamed and rewritten to pin the new lowering, carrying the measurement as its rationale so the next person to question the threshold finds numbers rather than the prose that misled this one.
- `arbitrary_dynamic_increment_stays_on_dynamic_dispatch` asserted "no `fadd`", but its stated intent was that an `Any` must not be **assumed** numeric. Still true — it is now *tested*, not assumed — so it asserts that directly: an `fadd` on an `Any` must be guarded, never bare.
- A typed-array RMW test searched one block's text for the dynamic helper, which now lives one block deeper in the diamond's cold arm. It now follows the add wherever it lowered.

## Binary size

`size(1)` `.text`: **+1792 bytes (+0.016%)** over 32 add sites on the arithmetic probe, **+384 (+0.0035%)** on the string probe — about **50 bytes per two-leaf site**.

Unlike the loop levers there is no admission criterion thinning the site count, so this scales with a very common syntactic form and deserves a real bundle measurement rather than an extrapolation. Counting the cc bundle with string literals, template literals and comments stripped first (13.6 MB → 6.0 MB of code) gives 1,834 `+=` and 6,587 binary `+` = **8,421 sites as a hard upper bound**, so **+0.159% worst case** on that binary — and genuinely a worst case, since it assumes every site takes a diamond when many are already statically proven numeric, already three-leaf, or string-typed.

String concatenation is unaffected — the failed guard costs nothing measurable: pair 78.55 vs 78.16, mixed 111.27 vs 112.07, triple 148.94 vs 149.29 ns/op.

## Gates

`-D warnings` clean; perry-hir 591/0; perry-codegen 1839/0 (the pre-change count exactly); perry-runtime lib 2841/0; census, address-classification, file-size and raw-handle-debt lints all pass with none raised. Integration: the new suite 5/5, `packed_loop_abrupt_statements` 7/7, `issue_8655_array_subclass_indexing` 2/2, `issue_8690_loop_versioned_arraylike` 3/3, `issue_8897_field_push_writeback` 3/3. Six generated differentials against node (336 loop/exit cases, 54 labeled-break cases, plus the break, hoist, lying-type and string suites) are byte-identical.

## Kill switch

`PERRY_DYNAMIC_ADD_PAIR_GUARD=0` restores the previous three-leaf threshold.

## Summary by CodeRabbit

  * Dynamic two-operand addition now uses a faster numeric path when values are numeric.
  * String concatenation and mixed-type operations continue to use the correct fallback behavior.
  * Added an option to restore the previous dynamic-addition behavior.

* **Bug Fixes**
  * Improved dynamic addition handling in typed-array updates and runtime type changes.

* **Tests**
  * Added coverage for numeric accumulation, coercion, misleading type annotations, and switching between numeric addition and string concatenation.

### DIFF
--- a/crates/perry-codegen/src/codegen/ordinary_param_guard_tests.rs
+++ b/crates/perry-codegen/src/codegen/ordinary_param_guard_tests.rs
@@ -911,10 +911,18 @@ fn guarded_discriminant_branch_narrows_a_union_parameter_inside_the_clone() {
             && !specialized.contains("js_dynamic_string_or_number_add"),
         "the guarded clone should narrow the discriminated union and add raw:\n{specialized}"
     );
+    // The generic body keeps the dynamic add. Since #9157 it also carries a
+    // guarded `fadd` in a fast arm, so the distinction that matters is not
+    // "has no fadd" but "has no PROOF": its add still dispatches dynamically
+    // whenever the runtime tag test fails, which the specialized clone above
+    // does not even emit.
     assert!(
-        generic.contains("call double @js_dynamic_string_or_number_add(")
-            && !generic.contains("fadd double"),
+        generic.contains("call double @js_dynamic_string_or_number_add("),
         "the unproven body must keep the dynamic add:\n{generic}"
+    );
+    assert!(
+        !generic.contains("fadd double") || generic.contains("guarded_add.numeric"),
+        "an fadd in the unproven body must be guarded:\n{generic}"
     );
 
     // The negative that pins the rule: same union, same discriminant chain,
--- a/crates/perry-codegen/src/expr/binary.rs
+++ b/crates/perry-codegen/src/expr/binary.rs
@@ -313,18 +313,36 @@ fn add_tree_leaves<'a>(expr: &'a Expr, out: &mut Vec<&'a Expr>) {
 ///
 /// A three-or-more-leaf tree is the common accumulator shape
 /// `sum += row.x + row.y`: lowering each node independently otherwise pays
-/// the dynamic add helper twice even when every runtime value is a number. At
-/// two leaves the guard merely moves the helper behind a branch; starting at
-/// three it can replace two or more helper calls with one shared tag check.
+/// the dynamic add helper twice even when every runtime value is a number,
+/// and one shared tag check replaces both.
+///
+/// Two leaves qualify as well, which the original threshold excluded on the
+/// grounds that the guard "merely moves the helper behind a branch". It does
+/// more than that. On the hot arm the helper call is replaced by an inline
+/// `fadd`, and the operands stop going through `lower_rooted_dynamic_binary`,
+/// which roots them — so a plain `s += v` accumulator was paying a call plus
+/// a write barrier per add and keeping its accumulator in a shadow-frame slot
+/// instead of a register. Measured on `s += v` with both operands numbers at
+/// runtime but neither statically proven: 3.44 -> 0.43 ns/op against node's
+/// 0.35. The cold arm still rebuilds the original tree, so a leaf that turns
+/// out to be a string concatenates exactly as before.
 ///
 /// Do not require a static numeric hint here. The important accumulator case
 /// is often a captured local plus fields read from interface-shaped objects,
 /// so every leaf is `Any` to codegen. The guard itself is the runtime proof;
 /// its cold arm preserves the original tree and exact dynamic `+` semantics.
 fn dynamic_add_tree_benefits_shared_guard(expr: &Expr) -> bool {
+    if matches!(
+        std::env::var("PERRY_DYNAMIC_ADD_PAIR_GUARD").as_deref(),
+        Ok("0") | Ok("off") | Ok("false")
+    ) {
+        let mut leaves = Vec::new();
+        add_tree_leaves(expr, &mut leaves);
+        return leaves.len() >= 3;
+    }
     let mut leaves = Vec::new();
     add_tree_leaves(expr, &mut leaves);
-    leaves.len() >= 3
+    leaves.len() >= 2
 }
 
 /// Rebuild the `+` tree over already-lowered leaf values, node for node, so the
--- a/crates/perry-codegen/src/expr/dynamic_add_tree_tests.rs
+++ b/crates/perry-codegen/src/expr/dynamic_add_tree_tests.rs
@@ -74,19 +74,32 @@ fn three_leaf_dynamic_add_tree_uses_one_shared_guard() {
 }
 
 #[test]
-fn two_leaf_dynamic_add_stays_on_direct_dispatch() {
+fn two_leaf_dynamic_add_takes_the_guard_too() {
+    // This test previously asserted the opposite, on the cost model that "a
+    // single dynamic add should not pay for a separate guard diamond" — the
+    // guard was thought merely to move the helper behind a branch. It does
+    // more: the hot arm becomes an inline `fadd`, and the operands stop going
+    // through `lower_rooted_dynamic_binary`, which roots them. Measured on
+    // `s += v` with both operands numbers at runtime but neither statically
+    // proven, the pair guard is worth 3.44 -> 1.33 ns/op (#9157).
     let mut body = dynamic_locals();
     body.push(result(add(Expr::LocalGet(A), Expr::LocalGet(B))));
     let ir = ir_for("two_leaf_dynamic_add", body);
 
-    assert!(
-        !ir.contains("guarded_add.numeric") && !ir.contains("fadd double"),
-        "a single dynamic add should not pay for a separate guard diamond:\n{ir}"
+    assert_eq!(
+        ir.matches("\nguarded_add.numeric.").count(),
+        1,
+        "a two-leaf tree should get one guard diamond:\n{ir}"
+    );
+    assert_eq!(
+        ir.matches("fadd double").count(),
+        1,
+        "the fast arm must perform the addition inline:\n{ir}"
     );
     assert_eq!(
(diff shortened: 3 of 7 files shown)


proggeramlug:
Merged, plus one commit: `PERRY_DYNAMIC_ADD_PAIR_GUARD` wasn't in `BUILD_CACHE_ENV_VARS` — it gates two different add sequences, so it's a cache key. (That's the eighth in this series; I filed #9158 about the pattern, not about any one PR.)

Admitting a **guarded `fadd` for dynamic `+`** is the operation where perry has historically gone wrong in the most quiet way — a NaN-boxed operand reaching a float add propagates its payload as a number rather than throwing or concatenating. So I probed the type lattice rather than the happy path: 20 shapes, **byte-identical to node v26.5.1**.

The systematic part is cases 1–4, which cross a 20-value operand set (`1`, `2.5`, `-0`, `0`, `NaN`, `±Infinity`, `"s"`, `""`, `"5"`, `true`, `false`, `null`, `undefined`, `[]`, `[1]`, `[1,2]`, `{}`, `1e21`, `-3`) against a number and a string on both sides — 80 pairings in all, covering every way the guard could mistake a concat for an add:

| | shape | node |
|---|---|---|
| 7 | `-0 + -0` vs `-0 + 0`, distinguished with `Object.is` | `-0` then `0` |
| 8, 9 | `NaN`, `Infinity + -Infinity`, `1e308 + 1e308`, and the i32 / `MAX_SAFE_INTEGER` boundaries | exact |
| 10, 11, 12 | `valueOf`-only, `toString`-only, and `Symbol.toPrimitive` receivers on both sides | `6\|6\|10`, `T1\|1T\|Tx`, `P1\|Px` |
| 13, 14 | `1n + 1` throws `TypeError`; `1n + 2n` and `1n + "x"` don't | exact |
| 15, 16 | one `+` site whose operand type **changes** between iterations (`[1,2,"a",3,"b"]`) | `"3a3b"` |
| 17, 18 | associativity of chained adds where a string appears mid-tree | `a12` vs `a3` |
| 19, 20 | `+=` on a local and on a property, transitioning number→string mid-sequence | `"3x3"`, `"3y"` |

Case 7 is the one I'd single out: `-0` is exactly where a float fast path and a spec-correct add diverge, and both forms are right. Case 15 is the polymorphic-site check — a guard that latched on the first observed pair would produce `"6ab"` or similar here.

Your own `dynamic_add_pair_guard` suite passes too (5 tests). Worth noting it lives in `crates/perry/tests/`, so per-PR CI only runs it when the diff names it — I ran it explicitly.

Validation: codegen 1356 passed, runtime 2841 (exit 0, 0 abort markers), perry --bins 1066, fmt clean, `run_lint_gates.sh` all 60 gates passed; `--diff-filter=D` empty. I did not independently reproduce the 3.44 → 1.33 ns figure.

PR #9161, opened 2026-08-30
perf(codegen): packed clones read a foreign counter inline (10_nested_loops 51 → 17 ms, Node parity)

proggeramlug:
A packed loop's fast clone can now read its array at an **enclosing** loop's counter, not only at its own.

## What was happening

The clone's element read is a raw slot load with no bounds check, licensed by the index being the loop's own induction variable — which the loop bound already proved in range. Any other index had no such proof, so the body matcher rejected `arr[i]` outright, and the **whole inner loop** fell to the typed-feedback tier: a registered guard call plus a boxed fallback per element, for *both* reads, on every iteration — even though the sibling read beside it would have been a raw load.

`benchmarks/suite/10_nested_loops.ts` is exactly this shape:

```ts
for (let i = 0; i < arr.length; i++)
  for (let j = 0; j < arr.length; j++)
    sum = sum + arr[i] + arr[j];   // arr[j] raw; arr[i] rejected the whole loop
```

## The change

A foreign i32 counter takes the same raw load behind one inline `icmp ult idx, len` against the live `ArrayHeader.length` — the same word `expr/index.rs`'s store guard reads — branching to the fact's **existing** side exit when it fails. That is the exit the hole arm already uses, so mid-body exit needed no new machinery, and the clone's bound did not need plumbing out of `precomputed_i32_bound`.

**Read-only bodies only, deliberately.** A side exit re-executes the iteration in the slow clone: harmless for a read, double-applying for a store. So the relaxation lives in `expr_is_packed_f64_loop_safe` and is *not* shared with the three store matchers sitting beside it, which keep the exact-counter rule.

## Measurements

Idle Mac mini, self-timed (the benchmark's own printed elapsed), min of three:

| benchmark | before | after | node |
|---|---:|---:|---:|
| `10_nested_loops` | 51 ms | **17 ms** | 16 ms |
| `16_matrix_multiply` | 121 ms | **100 ms** | 32 ms |

`10_nested_loops` goes from 3.3× Node to parity; `matrix_multiply` improves as a side effect, since its inner reads are the same shape. **No other benchmark in the 17-entry sweep moved** — that sweep was re-run specifically to catch a regression from relaxing a codegen admission.

## Correctness

Five-case differential against Node, chosen for the shapes where a side exit is observable: the plain nested accumulation, a labelled `break` out of the inner loop mid-iteration, an `if` guarding the accumulation, a mixed-layout array whose guard must fail into the slow clone, and an accumulator read by the outer loop between inner passes. All byte-identical.

`perry-codegen`: 30 suites pass, including a new IR pin (`packed_clone_reads_a_foreign_counter_without_the_feedback_call`). `cargo fmt --check` clean.

Stacked context: #9146 is a separate lane (byte-array reads) and does not touch this code.

## Summary by CodeRabbit

  * Improved optimization of nested numeric loops that read arrays using counters from enclosing loops.
  * Reduced unnecessary fallback processing, improving execution speed for eligible read-only array operations.

* **Bug Fixes**
  * Fixed packed numeric loop handling for array reads indexed by a different active loop counter.
  * Added bounds-safe behavior for these optimized reads, preserving correct results when indexes are out of range.

* **Tests**
  * Added regression coverage for nested loops using multiple array indexes.


proggeramlug:
Merged, with one commit from me for the file-size cap — details below.

Reading a **foreign** counter inside a packed clone means the clone now depends on a value it doesn't own, so I probed nested-loop shapes rather than the happy path. 18 shapes, byte-identical to node v26.5.1:

| | shape | node |
|---|---|---|
| 3 | inner bound **is** the outer counter (`for j < i`) | `220` |
| 4 | inner **start** is the outer counter (`for j = i`) | `572` |
| 7 | outer counter **mutated from inside** the inner loop (`i++`) | `65` |
| 8, 9, 10 | inner `break`, labelled `continue outer`/`break outer`, and `return` from the nest | `135`, `307`, `105` |
| 11 | closures capturing **both** counters per iteration | `[[0,0],[0,1],[1,0],[1,1],[2,0],[2,1]]` |
| 6 | three levels deep, all three counters read | `1620` |
| B63/B64/B65 | inner-loop lengths straddling the 64 poll stride | exact |
| 12, 13 | array `push`ed between outer iterations; an element turned into a string mid-nest | `[312,9]`, `"62z567"` |

Case 7 is the one I'd single out — if the foreign counter were cached rather than read live, mutating `i` from the inner body would diverge, and it doesn't. Case 11 confirms per-iteration binding still holds for both.

**The added commit.** This pushed `index_get.rs` to **2041** lines, over the 2000 cap, so `check_file_size.sh` was red. `index_get` is already a module directory, so I moved `foreign_packed_loop_read` and the two packed-fact lookups it sits beside into `index_get/foreign_counter.rs` (1963 lines now), re-exporting `packed_f64_loop_index_parts` so `expr/mod.rs`'s existing named import still resolves.

That split also **reattached a doc comment this PR had orphaned**: `foreign_packed_loop_read` was inserted between `packed_f64_loop_fact_for_index`'s doc comment and the function itself, so those four lines of "Look up a packed-f64 loop fact for `(arr, index-expr)`…" were documenting the wrong thing. Worth a look when inserting into a dense file — the compiler won't tell you.

I also renamed `changelog.d/packed-clone-foreign-counter-reads.md` → `9161-packed-clone-foreign-counter-reads.md` to match the PR-keyed convention (the filename is PR-keyed so in-flight fragments never collide).

Validation after the split: codegen 1356 passed, `native_proof_regressions` 285, perry --bins 1066, runtime 2841, fmt clean, `check_file_size.sh` PASS, `run_lint_gates.sh` all 60 gates passed; `--diff-filter=D` empty. I did not independently reproduce the 51 → 17 ms figure.

proggeramlug:
**Gate on the Linux box, commit `2809cebfb` — clean, with one thing worth flagging honestly.**

Lint: formatting PASS, CI-plan self-test PASS, gap-snapshot self-test PASS, parity-allowlist PASS, no ratchet ceilings raised. The 3 FAIL steps are the usual environmental ones (unexpanded `${{ github.* }}` outside Actions; a changelog fragment *is* present).

An earlier revision of this branch failed a fourth, **real** step — `File size limit exceeded (2000 lines)`, because the new arms pushed `expr/index_get.rs` to 2041. Fixed by moving the packed-clone fact lookups (`packed_f64_loop_fact_for_index`, `load_packed_loop_index_i32`, and the new `foreign_packed_loop_read`) into `index_get/guarded_array.rs`, which already owns the clone's load they serve — 1989 lines now. Re-verified behaviour-neutral after the move: 30 codegen suites, the five-case differential byte-identical, `10_nested_loops` still 17 ms.

**On the runtime suite:** two gate runs each reported `2818 passed; 2 failed` — but with a *different* second test each time (`box::release_tests::completed_activation_residue_is_bounded_not_linear`, then `symbol::get::own_data_ic_tests::composed_symbol_field_cache_reloads_mutated_final_slot`), alongside `gc::roots::stack_maps::decode_tests::…discovers_a_map_from_a_later_loaded_shared_object` in both.

That is pre-existing parallel flakiness, not this change, and I checked rather than asserted it:

- both tests pass single-threaded on this branch **and** on `origin/main`;
- running the runtime suite three times in parallel on **unmodified `origin/main`** reproduced it — run 3 failed on the same `discovers_a_map_from_a_later_loaded_shared_object`;
- this diff touches only `perry-codegen` (`stmt/loops.rs`, `expr/index_get.rs`, `expr/index_get/guarded_array.rs`, and the test file) — no runtime code at all;
- the box was under load 12–15 during the failing runs.

Worth a separate issue if it is not already tracked: `discovers_a_map_from_a_later_loaded_shared_object` loads a shared object and is the consistent offender under contention.

PR #9162, opened 2026-08-30
fix(codegen): guard erased BigInt arithmetic results

### LINKED ISSUE #9143: BigInt (a/d)*d + (a%d) drops the remainder term when operands are for...of bindings
Found while benchmarking BigInt division (#9141); reproduces on pristine main `653e886699`, unrelated to that change.

## Repro

The division identity `(a / d) * d + (a % d) === a` evaluates to `(a / d) * d` — the remainder term vanishes — when `a` and `d` are bindings from a `for...of` over a BigInt array:

```ts
for (const a of [123456789012345678901234567890n]) {
  for (const d of [1000000007n]) {
    console.log((a / d) * d + (a % d));  // node: a   perry: (a/d)*d
  }
}
```

Correct when the same operands are compile-time constants or plain locals — so this is about how `for...of` bindings type/lower in a compound BigInt expression, not about the division itself (division and remainder each produce correct values individually).

Repro files: `secret-tests/scratchpad/probe_q5.ts`, `probe_q6.ts`.

## Why it matters

Silent wrong arithmetic in the shape most real code uses to iterate big-number collections, and the dropped term makes results look plausible rather than obviously broken.

proggeramlug:
Fixes #9143.

## Summary
- classify non-add arithmetic as numeric only when a successful result cannot be a BigInt
- keep `>>>` on its always-Number result path
- add IR coverage for erased BigInt arithmetic and an end-to-end `for...of` parity regression

## Testing
- `cargo test -p perry-codegen --lib` (1357 passed, 1 ignored)
- `./run_parity_tests.sh --filter test_gap_9143` (1/1 passed)
- direct Node/Perry output comparison on `perrymaster.skelpo.net`
- `cargo fmt --all --check`
- `git diff --check`

No version bump.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed dynamic arithmetic handling for BigInt values in nested calculations.
  * Preserved BigInt behavior for compound arithmetic inside nested `for-of` loops.
  * Improved numeric type detection for division, multiplication, modulo, and related operations.

* **Tests**
  * Added regression coverage for dynamic arithmetic and BigInt identity calculations.

### DIFF
--- a/crates/perry-codegen/src/expr/dynamic_add_tree_tests.rs
+++ b/crates/perry-codegen/src/expr/dynamic_add_tree_tests.rs
@@ -20,9 +20,33 @@ fn any_local(id: u32, name: &str, init: Expr) -> Stmt {
     }
 }
 
+fn erased_bigint_local(id: u32, name: &str, value: &str) -> Vec<Stmt> {
+    vec![
+        Stmt::Let {
+            id,
+            name: name.to_string(),
+            ty: Type::Any,
+            mutable: true,
+            init: Some(Expr::Undefined),
+        },
+        Stmt::Expr(Expr::LocalSet(
+            id,
+            Box::new(Expr::BigInt(value.to_string())),
+        )),
+    ]
+}
+
 fn add(left: Expr, right: Expr) -> Expr {
     Expr::Binary {
         op: BinaryOp::Add,
+        left: Box::new(left),
+        right: Box::new(right),
+    }
+}
+
+fn arithmetic(op: BinaryOp, left: Expr, right: Expr) -> Expr {
+    Expr::Binary {
+        op,
         left: Box::new(left),
         right: Box::new(right),
     }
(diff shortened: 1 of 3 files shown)


proggeramlug:
Merged. Fixes #9143.

The framing in the summary is the right one — "classify non-add arithmetic as numeric only when a successful result **cannot** be a BigInt". Erasure bugs like this are nasty precisely because the wrong classification produces a plausible number rather than a crash, so I probed the full operator surface rather than the reported shape.

**44 shapes, 0 differences from node v26.5.1:**

| | shape | node |
|---|---|---|
| 11 ops | `+ - * / % ** & \| ^ << >>` on two BigInts | all `bigint`, values exact |
| `>>>` | the one operator that is always-Number | `TypeError` — correctly still throws |
| 11 mixes | each of those ops with `7n` and `3` | `TypeError` in all eleven |
| unary | `-5n`, `~5n` stay bigint; `+5n` throws | ✓ |
| 6 | the same ops routed through a function with `any` params (the erasure path) | all `bigint` |
| loop1–4 | `for…of` accumulation with `+ * - /` — the issue's shape | `6`, `24`, `92`, `16` |
| ca, cb | compound assignment chains (`-= *= /= %=`, `**= >>= <<=`) | `0`, `256` |
| div, mod | truncation toward zero with negative operands | `3,-3,-3,3` and `1,-1,1` |
| dz | `1n / 0n` | `RangeError`, not `Infinity` |
| big | `2n ** 70n + 1n`, `9007199254740993n * 2n` — beyond Number precision | exact |

I checked `typeof` on every result, not just the value — a misclassified BigInt that happens to print the right digits is exactly the failure this fixes, and `String(2n**70n+1n)` would round through a double.

Keeping `>>>` on its always-Number path is the detail I'd have most expected to get lost in a change phrased as "be more conservative", and it's intact.

Validation: codegen **1357 passed**, runtime 2841 (exit 0, 0 abort markers), perry --bins 1066, fmt clean, `check_file_size.sh` PASS, `run_lint_gates.sh` all 60 gates passed; `--diff-filter=D` empty.

#9165 (BigInt array element negation) looks like a sibling of this — picking it up next.

PR #9163, opened 2026-08-30
test(hir): exercise nested ws scope semantically

### LINKED ISSUE #9138: release: ws same-name scope test counts specialization clones as cross-tagging
Release candidate: v0.5.1519 at 0668d6d38f03452ba53ec2a92501fa0884030d35

Observed
- Full CI run 33294447010 failed in cargo-test-perry (3/8), job 99211618382.
- Test same_named_function_is_not_cross_tagged expected one call to js_ws_send_client_i64 but found two.

Root cause
- This is a stale IR-shape assertion, not a compiler scope regression.
- Generated main_ts.ll contains the two calls only in the top-level pushFrame specialized and generic variants.
- The same-named nested function lowers to perry_closure_main_ts__2 and contains no js_ws_send_client_i64 call.
- Function specialization is allowed to clone the top-level source call, so module-wide call count is not a stable proxy for declaration identity.

Required fix
- Keep asserting that the upgrade-fed top-level pushFrame lowers to the ws Client runtime.
- Execute the same-named nested function with a plain object and assert that its own send method runs.
- Do not weaken the scope regression to a mere nonzero IR count.

Acceptance
- The isolated test passes locally and on the dedicated Linux release server at the exact release SHA.
- The complete ws_client_handle_cross_function_dispatch suite passes.
- Release Full CI passes before any npm publication or tag creation.

proggeramlug:
## Summary

- keep the positive IR assertion that the upgrade-fed top-level helper dispatches to the ws Client runtime
- execute the same-named nested helper with a plain object and assert its own `send` method runs
- remove the brittle module-wide call count, which specialization may legitimately duplicate

## Testing

- isolated regression on the PR branch: 1 passed, 0 failed
- complete `ws_client_handle_cross_function_dispatch` suite on the PR branch: 11 passed, 0 failed
- isolated regression at `97bae62` directly atop release SHA `0668d6d`: 1 passed, 0 failed
- complete suite on that exact release lineage: 11 passed, 0 failed

Closes #9138

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket dispatch validation for nested declarations and same-named functions.
  * Confirmed that nested objects invoke their own `send` method correctly at runtime.
  * Updated test execution to verify compiled output directly, improving reliability when specialized and generic code paths coexist.

### DIFF
--- a/crates/perry/tests/ws_client_handle_cross_function_dispatch.rs
+++ b/crates/perry/tests/ws_client_handle_cross_function_dispatch.rs
@@ -73,12 +73,20 @@ fn ir_has_call(ir: &str, symbol: &str) -> bool {
         .any(|l| l.contains("call ") && l.contains(&needle))
 }
 
-/// Number of `call` instructions targeting `@<symbol>`.
-fn ir_call_count(ir: &str, symbol: &str) -> usize {
-    let needle = format!("@{symbol}(");
-    ir.lines()
-        .filter(|l| l.contains("call ") && l.contains(&needle))
-        .count()
+fn run_compiled(dir: &Path) -> String {
+    let output = dir.join("main_bin");
+    let run = Command::new(&output)
+        .current_dir(dir)
+        .output()
+        .expect("run compiled binary");
+    assert!(
+        run.status.success(),
+        "compiled binary failed\nstatus: {:?}\nstdout:\n{}\nstderr:\n{}",
+        run.status,
+        String::from_utf8_lossy(&run.stdout),
+        String::from_utf8_lossy(&run.stderr)
+    );
+    String::from_utf8_lossy(&run.stdout).into_owned()
 }
 
 fn compile_and_run(dir: &Path, source: &str) -> String {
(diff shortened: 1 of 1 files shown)


proggeramlug:
Merged. Test-only, and it neither helps nor hurts the suite it lives in — which needs explaining, because that suite is red.

`ws_client_handle_cross_function_dispatch` fails **9 of 11 on current main**, and identically with this PR applied (2 passed / 9 failed both ways). So this change is neutral against a pre-existing failure rather than green; I've filed the red suite separately as #9172 so it doesn't get attributed to whoever next touches this file.

Worth flagging for you specifically: because those 9 are already failing, **new tests added here can't be distinguished from the existing breakage** by a suite run. If the nested-scope semantics you're exercising are meant to pass today, that's worth confirming once #9172 is sorted.

One methodology note that cost me two wrong readings and is now in #9172 as well. My first run showed 11 of 11 failing. The extra two were my own environment: the compiler and runtime archives had been built from different commits, and perry's coherence guard refuses with

```
  library build: v0.5.1519 (commit 5f08e77c7cf7)
  Perry build:   v0.5.1519 (commit 3411c9d7c888)
```

which reaches the test as a bare `perry compile failed` and reads exactly like a product bug. Rebuilding `-p perry -p perry-runtime-static -p perry-stdlib-static` in one invocation gives the true count. I was one step from reporting "main is red on 11 tests" — it's red on 9.

Validation: codegen 1358 passed, runtime 2841 (exit 0, 0 abort markers), perry --bins 1066, fmt clean, `check_file_size.sh` PASS, `run_lint_gates.sh` all 60 gates passed; `--diff-filter=D` empty.

PR #9165, opened 2026-08-30
fix(codegen): preserve BigInt array element negation

### LINKED ISSUE #9142: Unary minus on a BigInt array element float-negates it (wrong value, silent)
Found while benchmarking BigInt division (#9141); reproduces on pristine main `653e886699`, unrelated to that change.

## Repro

```ts
const c = [2n ** 64n];
console.log(-c[0]);
```

| | result |
|---|---|
| node | `-18446744073709551616` |
| perry | `-18446744073709552000` |

The perry value is the f64 rounding of 2^64 — the BigInt is being converted to a double and float-negated. Silent wrong value, no throw.

## Characterization

- Reading `c[0]` alone is **correct** (prints the exact BigInt).
- `-v` where `v` is a plain local holding the same BigInt is **correct**.
- Only unary minus applied directly to an **array element** mis-lowers.
- An explicit `bigint[]` annotation does not help.

So the element read produces the right value and the unary-minus lowering picks the numeric path regardless of the operand's runtime type.

## Why it matters

BigInt arrays are the normal representation for big-number tables (hash constants, crypto limbs, byte→BigInt lookup tables like TypeBox's). A silently-rounded negation in that code produces wrong results with no diagnostic.

Repro file: `secret-tests/scratchpad/probe_q2.ts`.

proggeramlug:
## Summary

- route unary minus through the ToNumeric-aware dynamic helper when an operand may be a BigInt
- preserve the native `fneg` fast path for statically numeric operands and `ToNumber` for proven non-BigInt coercions
- add IR and source-level parity regressions for inferred and annotated BigInt arrays above 2^53
- leave version and changelog metadata unchanged

Closes #9142

## Testing

- `cargo test -p perry-codegen --lib` (1,357 passed, 1 ignored)
- matching release compiler/runtime build on `perrymaster.skelpo.net`
- `PERRY_SKIP_BUILD=1 ./run_parity_tests.sh --filter 9142` (1/1 passed against Node)

The broader `./scripts/test_affected_crates.sh --base origin/main` run passed all 2,820 active runtime tests, then aborted in the unrelated Perry binary test `commands::compile::cjs_wrap::preamble_canary_tests::path_module_wrap_publishes_partial_then_final_exports_and_tracks_undefined` with its reproducible stack overflow before reaching the codegen phase.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed unary negation for BigInt values retrieved from arrays.
  * Large BigInt values now retain their precision and BigInt type instead of being converted to numbers.
  * Improved numeric handling for expressions whose BigInt status cannot be determined statically.

* **Tests**
  * Added coverage for inferred and explicitly typed BigInt arrays, including values exceeding safe integer limits.
  * Verified negation uses BigInt-safe dynamic numeric handling.

### DIFF
--- a/crates/perry-codegen/src/expr/mod.rs
+++ b/crates/perry-codegen/src/expr/mod.rs
@@ -2726,6 +2726,8 @@ pub(crate) mod string_window;
 mod ptr_numarray_access;
 mod ta_param_f64_read;
 #[cfg(test)]
+mod unary_bigint_tests;
+#[cfg(test)]
 mod unary_bitnot_tests;
 pub(crate) use index_get::{
     numeric_index_has_integer_array_index_proof, packed_f64_loop_index_parts,
--- a/crates/perry-codegen/src/expr/unary.rs
+++ b/crates/perry-codegen/src/expr/unary.rs
@@ -9,8 +9,7 @@ use perry_hir::{Expr, UnaryOp};
 
 use crate::lower_conditional::lower_expr_with_truthy;
 use crate::type_analysis::{
-    expr_may_return_boxed_value_from_raw_f64_fallback, is_bigint_expr, is_numeric_expr,
-    is_provably_not_bigint,
+    expr_may_return_boxed_value_from_raw_f64_fallback, is_numeric_expr, is_provably_not_bigint,
 };
 use crate::types::{DOUBLE, I32, I64};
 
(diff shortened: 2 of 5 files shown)


proggeramlug:
Not merging yet — **this fixes eleven cases and regresses two**, and one of the two is a silent sign error.

I probed BigInt negation across 18 shapes against node v26.5.1, A/B'd against main. The fix is real and substantial:

|---|---|---|---|---|
| 1 | `-a[0]` on `[1n,2n,3n]` | `bigint -1` | **`number` -1** | `bigint -1` ✓ |
| 2 | `for` loop accumulating `-a[i]` | `bigint -6` | **`TypeError`** | `bigint -6` ✓ |
| 13, 14 | `-a[0]` where `a[0] = 2n**70n` | `-1180591620717411303424` | **`-1.1805916207174113e+21`** | exact ✓ |
| 5, 6, 7, 8, 10, 12, 16 | Map value, nested array, mixed array, `BigInt64Array` | `bigint` | `number` | `bigint` ✓ |

Row 13 is the one that shows why this matters — main round-trips a 70-bit BigInt through a double and loses it silently.

**But two rows that are correct on main become wrong here:**

```js
const a = [5n];
-(-a[0])            // node: 5n     main: 5n     this PR: NaN

const a2 = [10n, 3n];
(-a2[0]) * (-a2[1]) // node: 30n    main: 30n    this PR: -10n
```

Double negation yields `NaN`, and a product of two negated elements comes back with one sign dropped *and* the wrong magnitude. Both look like the negation result isn't itself being classified as BigInt — so a second operation consuming it falls back to the numeric path. That's the same class of bug the PR fixes, one level up.

The `-10n` case is the one I'd weight most: it produces a plausible BigInt rather than a `NaN` or a throw, so it would survive review and land in someone's output.

Row 15 (`(-0n) === 0n` → `false`, node says `true`) is wrong on both arms, so that one is pre-existing and not yours.

Suites are otherwise green with this applied: codegen 1358 passed, runtime 2841 (exit 0, 0 abort markers), perry --bins 1066, `check_file_size.sh` PASS, and all 60 lint gates. So the two rows above are the only blocker.

Reproducer, if useful:

```ts
const a = [5n];  console.log(-(-a[0]));                 // want 5n
const b = [10n, 3n]; console.log((-b[0]) * (-b[1]));    // want 30n
```

Happy to re-run the full 18-shape probe as soon as you push.

proggeramlug:
Merged, with the two regressions I reported on the handback fixed here rather than sent back again.

**The diagnosis in your PR is right and the root cause is one layer below it.** You changed `unary.rs` from `is_bigint_expr` (a positive proof that the operand IS a BigInt) to `!is_provably_not_bigint` (a positive proof that it is NOT) — correct, because an indexed read from a `bigint[]` is unproven in both directions and must take the dynamic path. But `is_numeric_expr`'s own `Unary` arm still asked the weak question:

```rust
matches!(op, UnaryOp::Neg | UnaryOp::Pos | UnaryOp::BitNot)
    && !is_bigint_expr(ctx, operand)
```

So `-a[0]` lowered correctly through the dynamic helper and produced a NaN-boxed BigInt, and then the *enclosing* operator asked "is this numeric?", got `true`, and emitted a native `fneg`/`fmul` straight over the BigInt payload:

| | before | after | node |
|---|---|---|---|
| `-(-a[0])` | `NaN` (number) | `2n` (bigint) | `2n` |
| `(-a[0]) * (-a[1])` | `-2n` | `10n` | `10n` |
| `-(-big[0]) === big[0]` | `false` | `true` | `true` |

The arm now uses the same predicate the lowering does — `is_numeric_expr(operand) || is_provably_not_bigint(operand)` — so it claims a clean f64 exactly when the dynamic path was not taken. That's the invariant: *the result is numeric iff the lowering didn't route through the helper*, and having the two disagree is what produced a "fix" that was correct in isolation and wrong one level up.

It costs some optimization breadth — the arm now returns `false` where it used to guess `true` — but in the direction that can only lose speed, never correctness.

**Validation:** the ten-shape probe (single negation, double negation, product of two negations, mixed add/subtract, a value above 2^53, round-trip identity, the all-`number` control, a mixed `any[]`, an accumulating loop, and unary `+`) matches node 26.5.1 byte-for-byte. `perry-codegen` 31 suites / 0 failures; `perry-runtime` 2867 passed / 0 failed at `RUST_TEST_THREADS=1`; all 60 lint gates green.

Worth keeping the `num` control row in the regression set — it's what tells you the stronger predicate didn't quietly disable the numeric path for plain `number[]`, which would be an invisible perf regression rather than a visible wrong answer.

PR #9166, opened 2026-08-30
perf(codegen): -, * and / take the guarded inline path too — subtraction was 2.6x slower than addition

proggeramlug:
Follows #9159, which gave `+` a guarded inline path and left every other arithmetic operator behind.

## The asymmetry

`s += v` lowered to a guarded diamond with an inline `fadd`. `s -= v`, on identical operands, lowered to an unconditional `js_dynamic_sub` call — so **subtraction was 2.6x slower than addition**, for no reason a reader of the source could predict:

| accumulator | lowers to | perry | node |
|---|---|---|---|
| `s += v` | guarded diamond, inline `fadd` | 1.33 | 0.345 |
| `s -= v` | `js_dynamic_sub` call, unconditional | 3.44 | 0.345 |
| `s ^= v` | `js_dynamic_bitxor` call, unconditional | 16.57 | 0.312 |

The bail these operators hit is the one taken whenever an operand *might* be a BigInt — which, for an operand codegen cannot prove, is always. But the guard needs no proof: **a BigInt is not a Number**, so it fails the tag test and the cold arm runs the same helper it always did. These are also *simpler* than `+`, because there is no string case to preserve and therefore no tree to rebuild.

## Numbers

Quiet Mac mini, node v26.5.1, ns/op per operation, best of 7, medians of 3 runs. Both perry columns are **the same binary**, switched with `PERRY_GUARDED_ARITH`. The operand comes from `arr.length`, so it is numeric at runtime but not statically proven:

| | perry off | perry on | node |
|---|---|---|---|
| `s -= v` | 3.439 | **0.942** | 0.345 |
| `s += v` | 1.331 | 1.331 | 0.345 |

Subtraction now beats addition, because `+` still carries a string arm and, per the second commit below, still needs its write barrier.

## Two commits

**1. Only `+` needs its operands proven numeric to skip the barrier.** `expr_is_known_non_pointer_shadow_value` judged every `Expr::Binary` by `is_numeric_expr`, so an accumulator written by `-=`, `*=` or `^=` with an unproven operand kept a write barrier per store — for a string case its operator **cannot produce**. That is a wrong premise in a predicate rather than a missing optimization. Verified in the emitted code rather than by timing: `subAcc` and `xorAcc` go from 8 `js_write_barrier_root_nanbox` calls each to 0, while `addAcc` correctly keeps its 8.

**2. `-`, `*` and `/` take the guarded inline path.**

Bitwise operators are deliberately **out of scope**: they already have an inline path, gated on a *static* not-BigInt proof, so giving them the runtime guard is a different mechanism and belongs in its own change. `s ^= v` remains 16.57.

## Semantics

Checked against node on both lowerings: BigInt arithmetic (`10n - 3n` = 7n, `10n * 3n` = 30n, `10n / 3n` = 3n), the `bigint - number` TypeError, a 49-case operand cross product over number/string/bool/null/undefined, the `-0` / `Infinity` / `NaN` edges, and an operand that turns non-numeric midway through a loop so the same site must switch arms. Every expected value was taken from node first.

## Expected app-level impact: none

Stating this up front rather than after measurement. #9159 measured **flat on cc** — 9,785,453,966 instructions against 9,781,405,758, inside the 0.08% noise floor — and I expect the same here, because cc's startup does little unproven numeric arithmetic.

I am submitting it anyway: **subtraction being 2.6x slower than addition on identical operands is a defect regardless of whether cc happens to subtract**, and the barrier premise was simply wrong. If the cc number comes back flat, that costs this change nothing, because it was never the argument for it.

## Binary size

`size(1)` `.text`: +320 bytes over ~8 guarded sites, about **40 bytes per site** — smaller than #9159's ~50, since there is no string arm to emit.

## Kill switch

`PERRY_GUARDED_ARITH=0` restores the unconditional dynamic helper.

## Gates

`-D warnings` clean; perry-hir 591/0; perry-codegen **1839/0** (the pre-change count exactly); perry-runtime lib 2841/0; census, address-classification, file-size and raw-handle-debt lints all pass with none raised. Integration: the new suite 7/7, plus `dynamic_add_pair_guard` 5/5, `packed_loop_abrupt_statements` 7/7, `issue_8655_array_subclass_indexing` 2/2, `issue_8690_loop_versioned_arraylike` 3/3, `issue_8897_field_push_writeback` 3/3.

Two existing codegen assertions changed. Both pinned "contains no `fsub`" / "contains no `fmul`", which was written when the only way to reach raw f64 arithmetic was an *unguarded* static proof — so the shape and the invariant were the same statement. They no longer are. Each now asserts the property it existed for: boxed bits must not reach **unguarded** raw arithmetic. The `js_dynamic_sub` / `js_dynamic_mul` assertions beside them are unchanged and carry the other half.

## Summary by CodeRabbit

* **Performance**
  * Improved subtraction, multiplication, and division for numeric values by using a faster execution path when operands are confirmed to be numbers.
  * Preserved existing behavior for BigInt, mixed-type, and dynamically typed operands.

* **Bug Fixes**
  * Ensured arithmetic continues to match JavaScript behavior for signed zero, infinity, and changing operand types.
  * Added coverage for numeric, string, BigInt, and mixed-value scenarios.

### DIFF
--- a/crates/perry-codegen/src/expr/binary.rs
+++ b/crates/perry-codegen/src/expr/binary.rs
@@ -308,11 +308,9 @@ fn lower_guarded_numeric_arith(
 
         ctx.current_block = slow_idx;
         crate::expr::emit_versioned_loop_callback_deopt(ctx);
-        let slow_val = ctx.block().call(
-            DOUBLE,
-            fname,
-            &[(DOUBLE, &values[0]), (DOUBLE, &values[1])],
-        );
+        let slow_val =
+            ctx.block()
+                .call(DOUBLE, fname, &[(DOUBLE, &values[0]), (DOUBLE, &values[1])]);
         let slow_end = ctx.block().label.clone();
         ctx.block().br(&merge_label);
 
@@ -1181,9 +1179,7 @@ pub(crate) fn lower(ctx: &mut FnCtx<'_>, expr: &Expr) -> Result<String> {
                         if guarded_arith_enabled()
                             && matches!(op, BinaryOp::Sub | BinaryOp::Mul | BinaryOp::Div)
                         {
-                            return lower_guarded_numeric_arith(
-                                ctx, *op, left, right, fname,
-                            );
+                            return lower_guarded_numeric_arith(ctx, *op, left, right, fname);
                         }
                         return lower_rooted_dynamic_binary(ctx, fname, left, right);
                     }
--- a/crates/perry-codegen/src/expr/binary.rs
+++ b/crates/perry-codegen/src/expr/binary.rs
@@ -244,6 +244,94 @@ fn lower_guarded_numeric_add(ctx: &mut FnCtx<'_>, expr: &Expr) -> Result<String>
     })
 }
 
+/// The `+`-shaped guard for the arithmetic operators that cannot concatenate.
+///
+/// `-`, `*` and `/` apply ToNumeric to both operands, so unlike `+` there is
+/// no string case to preserve — the whole reason `lower_guarded_numeric_add`
+/// has to rebuild its tree. That makes the guard strictly simpler: test both
+/// operands, do the operation inline when they are numbers, and hand the
+/// original operands to the same dynamic helper otherwise.
+///
+/// BigInt is handled by the test rather than by a separate check: a BigInt is
+/// not a Number, so `5n - 3n` and `1n - 1` both take the cold arm and keep
+/// the helper's exact semantics (a BigInt result and a TypeError
+/// respectively).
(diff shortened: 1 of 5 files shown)


proggeramlug:

**I added one commit**: `PERRY_GUARDED_ARITH` was not in `BUILD_CACHE_ENV_VARS`, so `codegen_env_vars_are_build_cache_inputs` was red. It gates two different arithmetic sequences, so it is a cache key. That is the ninth time in this series — see #9158, which proposes making the gate self-servicing rather than leaving it as the only thing carrying that knowledge.

PR #9167, opened 2026-08-30
fix(runtime): preserve function expando key order

### LINKED ISSUE #9148: Function-object expandos enumerate out of insertion order
Expando properties on a **function object** enumerate in the wrong order — insertion order is not preserved.

```ts
const f: any = function () {};
f.tag = 1;
f.other = 2;
console.log(Object.keys(f));
```

```
node v26.5.1  [ 'tag', 'other' ]
perry         [ 'other', 'tag' ]
```

Per ECMA-262 `OrdinaryOwnPropertyKeys`, own string keys come back in property-creation order after the integer-index keys, so `tag` must precede `other`.

### Narrow to function objects

Found while auditing #9147 (a `FastKeyHasher` change) and confirmed identical on main, so it predates that PR and is not hash-related. Every adjacent shape I probed is **correct**:

| shape | result |
|---|---|
| plain object, mixed integer + string keys (`z`,`a`,`"10"`,`"2"`,`b`) | `["2","10","z","a","b"]` ✓ |
| 500 string keys `p0…p499` | correct order, first `p0`, last `p499` ✓ |
| 300 integer keys | ascending ✓ |
| reverse-inserted `k20…k0` | insertion order ✓ |
| delete then re-add (`{a,b,c}`, delete `b`, re-add `b`) | `["a","c","b"]` ✓ |
| array expando (`a.custom = "x"`) | `["0","1","2","custom"]` ✓ |
| `for…in` across a prototype chain | `["own","inh"]` ✓ |

So plain objects, arrays and rehash-sized objects are all fine; it is specifically the function-object expando path.

### Why it is worth fixing rather than shrugging at

Function expandos are common in real code — attaching metadata to a callback, `fn.displayName`, decorator bookkeeping, memoisation caches hung off the function. Anything that round-trips such an object through `Object.keys`, `JSON.stringify`, or object spread will see a different order than Node, and order differences of this kind usually surface far from the cause (a snapshot test, a serialized cache key, a diff).

Two keys is the minimal reproducer; worth checking whether the order is reversed, hash-derived, or something else once someone looks — with only two keys those are indistinguishable.

proggeramlug:
## Summary

Preserves ECMA-262 own-key ordering for expando properties attached to function objects. Function expandos now match Node: integer-index keys enumerate numerically first, followed by other strings in property-creation order.

## Changes

- Keep O(1) function-expando lookup while recording property-creation order.
- Preserve overwrite position and move deleted/re-added properties to the tail.
- Apply the shared ECMA ordering helper to the complete function own-property name list.
- Add focused runtime and end-to-end regression coverage for keys, values, entries, `for...in`, spread, and reflection.

## Related issue

Fixes #9148

## Test plan

- [ ] `cargo build --release` clean (not run; equivalent affected targets built with the repository perry-dev profile)
- [ ] `cargo test --workspace --exclude perry-ui-ios --exclude perry-ui-tvos --exclude perry-ui-watchos --exclude perry-ui-gtk4 --exclude perry-ui-android --exclude perry-ui-windows` passes (not run verbatim; affected-crate gate passed)
- [x] `RUST_TEST_THREADS=1 cargo test --profile perry-dev -p perry-runtime tests_9148 -- --test-threads=1`
- [x] `PERRY_RUNTIME_DIR=/root/perry-target-9148/perry-dev cargo test --profile perry-dev -p perry --test issue_9148_function_expando_order`
- [x] `RUST_MIN_STACK=33554432 ./scripts/test_affected_crates.sh --base origin/main`
- [x] Added a user-facing regression fixture under `test-files/`.
- [ ] Documentation update (not needed; no public runtime API changed)
- [ ] Platform UI backend build (not applicable)

## Screenshots / output

Before: `Object.keys(f)` produced `["other","tag"]`.

After: `Object.keys(f)` produces `["tag","other"]`, matching Node.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Function-object properties now enumerate in standard ECMAScript key order.
  * Numeric keys appear first in ascending order, followed by string keys in creation order.
  * Updating a property preserves its position, while deleting and re-adding it moves it to the end.
  * Consistent ordering is now applied across keys, values, entries, iteration, spreading, and own-property-name queries.

* **Tests**
  * Added coverage for function-property ordering and related updates.

### DIFF
--- a/changelog.d/9167-function-expando-order.md
+++ b/changelog.d/9167-function-expando-order.md
@@ -1,0 +1,6 @@
+### Fixed
+
+- Function-object expando properties now follow ECMAScript own-key ordering:
+  integer-index keys enumerate numerically, while other string keys retain
+  property-creation order. Overwriting a property keeps its position, and
+  deleting and re-adding it moves it to the tail (#9148).
--- a/crates/perry-runtime/src/closure/dynamic_props.rs
+++ b/crates/perry-runtime/src/closure/dynamic_props.rs
@@ -6,23 +6,92 @@ use crate::fast_hash::{new_ptr_hash_map, PtrHashMap};
 use std::collections::{HashMap, HashSet};
 use std::sync::{Mutex, OnceLock};
 
+/// Per-function dynamic properties with ordinary property-creation order.
+///
+/// Reads stay O(1) through the hash map, while `order` records the string-key
+/// insertion order required by `OrdinaryOwnPropertyKeys`. Updating an existing
+/// property leaves its position unchanged; deleting it removes the position so
+/// a later re-add appends it at the end.
+#[derive(Default)]
+struct ClosureProps {
+    values: HashMap<String, f64>,
+    order: Vec<String>,
+}
+
+impl ClosureProps {
+    fn contains_key(&self, key: &str) -> bool {
+        self.values.contains_key(key)
+    }
+
+    fn get(&self, key: &str) -> Option<&f64> {
+        self.values.get(key)
+    }
+
+    fn insert(&mut self, key: String, value: f64) {
+        if !self.values.contains_key(&key) {
+            self.order.push(key.clone());
+        }
+        self.values.insert(key, value);
+    }
+
+    fn remove(&mut self, key: &str) -> Option<f64> {
+        let value = self.values.remove(key)?;
+        self.order.retain(|existing| existing != key);
+        Some(value)
+    }
+
+    fn merge_older(&mut self, mut older: ClosureProps) {
+        let newer = std::mem::take(self);
+        for key in newer.order {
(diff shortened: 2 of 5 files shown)


proggeramlug:

**This closes #9148, which I filed yesterday** while auditing #9147 — function-object expandos enumerating out of insertion order (`f.tag=1; f.other=2` giving `["other","tag"]`). I re-ran that probe against the train: **0 of 16 shapes wrong, down from 1 of 16**, with the neighbouring orderings (plain objects with mixed integer/string keys, 500-key rehash, 300 integer keys, delete-then-re-add, array expandos, `for…in` across a prototype chain) all still correct. Good to see it fixed at the source.

Issue #9131, opened 2026-08-30
Prototype method replacement not observed through a typed-parameter receiver


proggeramlug:
A prototype method replaced after first call is not picked up when the receiver arrives as a **typed parameter**, though it is picked up on a direct call.

```ts
class P { v = 1; m() { return 1; } }
function run(o: P, n: number) { let s = 0; for (let i = 0; i < n; i++) s += (o as any).m(); return s; }

const p = new P();
const before = run(p, 3);
(P.prototype as any).m = function () { return 100; };
const after = run(p, 3);
console.log(before, after);
```

```
node v26.5.1  3 300
perry         3 3
```

Patching mid-loop behaves the same way:

```ts
const p = new P(); let s = 0;
for (let i = 0; i < 6; i++) { s += p.m(); if (i === 2) (P.prototype as any).m = function () { return 50; }; }
// node: 153    perry: 6
```

The direct-call form **is** correct, which is what makes this narrow:

```ts
const p = new P(); const a = p.m();
(P.prototype as any).m = function () { return "patched"; };
console.log(a, p.m());   // node: orig patched   perry: orig patched  ✓
```

### Not a regression

Found while auditing #9124 and re-checked on #9130. Both reproduce **identically on main**, so this predates the inline-probe series — but it lives in exactly the code that series is changing (`method_override.rs`'s guard selection), so it is worth pinning before more optimization lands on top.

### Related shapes that are also wrong on main

From the same 23-shape probe, all identical on main and on #9124/#9130:

| | shape | node | perry |
|---|---|---|---|
| `Object.setPrototypeOf(c, {inc: () => 777})` then `c.inc()` | `777` | `2` |
| `Object.setPrototypeOf(p, Q.prototype)` then `p.m()` | `"b"` | `"a"` |
| `c.inc.call(otherInstance)` | `[101,101,1]` | `TypeError` |
| `get m() { return () => this.v * 7 }` then `g.m()` | `21` | `TypeError` |
| `delete c.v` mid-loop then `c.inc()` | `NaN` | `21` |

The first two are the same root as the headline case (a stale prototype edge); the last three look independent and may deserve splitting out once someone digs in.

### Why it matters beyond spec conformance

Monkey-patching a prototype is how a lot of instrumentation, test doubles and polyfills work. Silently continuing to call the old body — with no error — is the failure mode most likely to be misdiagnosed as "my mock didn't get applied".

PR #9168, opened 2026-08-30
fix(runtime): honor deleted prototype methods in direct guards

### LINKED ISSUE #9123: Direct method dispatch survives `delete C.prototype.method` (node throws TypeError)
## Repro

```ts
class C { a = 5; inc(): number { return this.a + 1; } }
function viaParam(c: C, n: number): number { let l = 0; for (let i = 0; i < n; i++) l = c.inc(); return l; }
delete (C.prototype as any).inc;
try { console.log(viaParam(new C(), 1)); } catch (e) { console.log("threw:", (e as Error).constructor.name); }
```

node: `threw: TypeError` — perry (main `19082697e7`, also every branch since): prints `6`.

## Mechanism

The direct-method guards (`js_method_direct_shape_guard`, the inline form in `lower_call/method_override.rs`, and the typed-feedback contract) invalidate on prototype **overrides** through `PERRY_CLASS_PROTOTYPE_FAST_GUARDS_INVALIDATED[_BY_METHOD]`, but a `delete` of the prototype method does not flip the latch, so the guard still passes and the compiled body runs.

The same script's monkey-patch case (`C.prototype.inc = function…` mid-program) IS honored, so the miss is specific to deletion. Found while validating the probe-first method dispatch change (differential vs node, identical on both switch arms — i.e. pre-existing, not introduced there).

proggeramlug:
## Summary

- make typed-feedback direct-method contracts consult the scoped prototype invalidation latch
- prevent a deleted declared prototype method from being re-accepted as a valid vtable direct call
- add focused runtime coverage and an end-to-end regression for the issue repro

## Testing

- cargo fmt --all -- --check
- cargo test -p perry-runtime typed_feedback_method_direct_guard
- cargo test -p perry --test issue_9123_method_delete_invalidation
- compiled and ran test-files/test_method_guard_name_invalidation.ts against the matching worktree runtime

No version bump; this includes only the standard changelog fragment.

Fixes #9123

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed prototype method deletion so subsequent calls no longer invoke stale compiled method code.
  * Deleted methods now correctly fall back to standard dispatch and produce a `TypeError`.

* **Tests**
  * Added regression coverage for method deletion and invalidated direct-call guards.

### DIFF
--- a/changelog.d/9123-prototype-method-delete-guard.md
+++ b/changelog.d/9123-prototype-method-delete-guard.md
@@ -1,0 +1,6 @@
+### fix(runtime): honor prototype method deletion in typed direct guards
+
+Typed-feedback direct-method guards now consult the same per-name prototype
+invalidation latch as inline shape guards. Deleting a declared prototype
+method therefore falls back to ordinary dispatch and throws TypeError
+instead of invoking the stale compiled method body. Fixes #9123.
--- a/crates/perry-runtime/src/typed_feedback/guards.rs
+++ b/crates/perry-runtime/src/typed_feedback/guards.rs
@@ -106,6 +106,11 @@ fn method_direct_call_contract(
         );
     };
     let name_hash = hash_bytes(method_bytes);
+    let method_guard_slot = crate::object::class_prototype_method_guard_slot(method_name);
+    if crate::object::class_prototype_fast_guard_invalidated_for_method(method_guard_slot) {
+        return (shape_addr, class_id, gc_type, name_hash, false);
+    }
+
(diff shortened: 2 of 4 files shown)


proggeramlug:

Pairs with #9169, which is the other half of #9131 (prototype replacement not observed through a typed-parameter receiver). The two conflict in the prototype-guard code, so I landed this one first and am rebasing #9169 onto it — the monkey-patching probe is still 7/23 with only this applied, which is expected since #9169 carries the observation half.

Issue #7576, opened 2026-08-07
Iterator.from(x) returns an already-exhausted iterator; .map/.filter/.take return undefined


proggeramlug:
`Iterator.from(x)` returns an iterator that is exhausted before it starts, and
the whole TC39 iterator-helpers surface goes with it: `.map`, `.filter`,
`.take`, `.drop`, `.flatMap` all return `undefined`, so any chain throws
`TypeError: Cannot read properties of undefined`.

Found while working #7564 (which touches the same file). **Not caused by it** —
verified byte-identical on the pre-#7564 runtime.

## Repro

```ts
function makeCounter(n: number): Iterator<number> {
    let i = 0;
    return {
        next(): IteratorResult<number> {
            return i < n ? { value: i++, done: false } : { value: undefined, done: true };
        },
    };
}

const a = Iterator.from(makeCounter(4));
const out: unknown[] = [];
let s = a.next();
let guard = 0;
while (!s.done && guard++ < 20) { out.push(s.value); s = a.next(); }
console.log("A Iterator.from({next}):", JSON.stringify(out));
console.log("B Iterator.from(arrayIter):", JSON.stringify([...(function () {
    const b = Iterator.from([7, 8, 9][Symbol.iterator]());
    const o: unknown[] = []; let t = b.next(); let g = 0;
    while (!t.done && g++ < 20) { o.push(t.value); t = b.next(); }
    return o;
})()]));

// the same object driven directly, WITHOUT the helper
const c = makeCounter(3);
const outC: unknown[] = []; let sc = c.next(); guard = 0;
while (!sc.done && guard++ < 20) { outC.push(sc.value); sc = c.next(); }
console.log("C bare {next}:", JSON.stringify(outC));

const d = Iterator.from(makeCounter(2));
console.log("D first step:", JSON.stringify(d.next()));
```

| line | node 26.5.1 | perry |
|---|---|---|
| A `Iterator.from({next})` | `[0,1,2,3]` | `[]` |
| B `Iterator.from(arrayIter)` | `[7,8,9]` | `[]` |
| C bare `{next}` (no helper) | `[0,1,2]` | `[0,1,2]` ✅ |
| D first step | `{"value":0,"done":false}` | `{"done":true}` |
| E `Iterator.from(mapIter)` | `[1,2]` | `[]` |

Row C is the tell: the underlying iterator works fine when driven directly. It
is the helper wrapper that reports `done` on its very first step, for every
[...]

PR #9169, opened 2026-08-30
fix(runtime): observe prototype replacement in method calls

### LINKED ISSUE #9131: Prototype method replacement not observed through a typed-parameter receiver

proggeramlug:
Fixes #9131

## Summary
- invalidate exact-receiver loop facts when prototype surgery can occur on a later iteration
- retire guarded direct-method dispatch after user-visible `Object.setPrototypeOf` calls
- make per-instance prototype chains authoritative over the original class vtable for reads and calls
- add end-to-end coverage for between-call, mid-loop, object-prototype, and class-prototype replacements

## Testing
- `cargo test -p perry --test issue_9131_prototype_method_replacement -- --nocapture`
- `cargo test -p perry-transform`
- `cargo test -p perry --test issue_5763_setprototypeof_chain_end --test issue_5477_event_emitter_prototype_methods --test issue_6084_write_fast_path_per_object --test static_method_object_literal -- --nocapture`
- `cargo fmt --check`

No version bump.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed method calls so prototype method replacements are recognized immediately.
  - Corrected property and method lookup after changing an object’s prototype with `Object.setPrototypeOf`.
  - Prevented stale optimized lookups from returning class-defined methods when a custom prototype chain is active.
  - Improved behavior when prototype changes occur during loops, including cross-class prototype swaps.
  - Corrected `__proto__` and `constructor` lookups for arrays with customized prototypes.

### DIFF
--- a/crates/perry-runtime/src/object/class_registry.rs
+++ b/crates/perry-runtime/src/object/class_registry.rs
@@ -117,8 +117,8 @@ pub(crate) use prototype_methods::{
 pub(crate) use prototype_methods::{
     class_prototype_fast_guard_invalidated_for_method, class_prototype_method_guard_slot,
     class_prototype_method_root_remove, class_prototype_method_root_store,
-    invalidate_class_prototype_fast_guards_for_method, mirror_prototype_method_on_object,
-    synthetic_class_id_for_function,
+    invalidate_class_prototype_fast_guards, invalidate_class_prototype_fast_guards_for_method,
+    mirror_prototype_method_on_object, synthetic_class_id_for_function,
 };
 pub use prototype_methods::{
     js_class_register_static_field, js_get_function_prototype_method,
--- a/crates/perry-runtime/src/object/class_registry/prototype_methods.rs
+++ b/crates/perry-runtime/src/object/class_registry/prototype_methods.rs
@@ -192,7 +192,6 @@ pub(crate) fn invalidate_class_prototype_fast_guards_for_method(name: &str) {
     retire_prototype_dependent_caches();
 }
 
-#[allow(dead_code)] // Fail-closed escape hatch for a future keyless mutation path.
 pub(crate) fn invalidate_class_prototype_fast_guards() {
     #[cfg(not(test))]
     PERRY_CLASS_PROTOTYPE_FAST_GUARDS_INVALIDATED.store(1, std::sync::atomic::Ordering::Release);
--- a/crates/perry-runtime/src/object/field_get_set.rs
+++ b/crates/perry-runtime/src/object/field_get_set.rs
@@ -218,6 +218,9 @@ mod ic_miss;
 mod ic_miss_array_length_tests;
 mod map_set_receiver;
 mod probe_dispatch;
+/// #9131: per-instance `[[Prototype]]` override lookup, split out of
+/// `get_field_by_name_tail.rs` for the 2000-line cap.
+mod prototype_override;
 
 /// Size of the direct-mapped `(keys_ptr, key_hash, field_index)` inline
 /// cache backing `js_object_get_field_by_name`'s slow tail.
--- a/crates/perry-runtime/src/object/field_get_set/get_field_by_name_tail.rs
+++ b/crates/perry-runtime/src/object/field_get_set/get_field_by_name_tail.rs
@@ -1357,6 +1357,17 @@ pub(crate) fn get_field_by_name_object_tail(
         let keys = crate::object::object_keys_array(obj);
 
         if keys.is_null() {
+            // An explicit per-instance [[Prototype]] replaces the class's
+            // declaration prototype; it is not an extra link in front of the
+            // original vtable. Walk that authoritative chain before exposing
+            // class getters or methods, and do not resurrect the old class
+            // surface when the custom chain misses.
+            if !key.is_null()
+                && super::super::prototype_chain::object_has_prototype_override(obj as usize)
+            {
+                return super::super::prototype_chain::resolve_inherited_field(obj as usize, key)
+                    .unwrap_or_else(JSValue::undefined);
+            }
             // #809: an object with no own keys (e.g. an `Object.create(proto)`
             // result, or a `Function.prototype = obj` instance) still has to
             // resolve inherited props/methods. Pre-fix this returned undefined
@@ -1725,6 +1736,16 @@ pub(crate) fn get_field_by_name_object_tail(
                     };
                 }
             }
+        }
+
+        // A shaped receiver's own-key scan has missed. As in the keyless arm
+        // above, a user-installed per-instance prototype is now authoritative
+        // and must win over class-vtable getters and methods.
+        if !key.is_null()
+            && super::super::prototype_chain::object_has_prototype_override(obj as usize)
+        {
+            return super::super::prototype_chain::resolve_inherited_field(obj as usize, key)
+                .unwrap_or_else(JSValue::undefined);
         }
 
         // Key not found in the keys_array — fall back to the class
--- a/crates/perry-runtime/src/object/field_get_set/get_field_by_name_tail.rs
+++ b/crates/perry-runtime/src/object/field_get_set/get_field_by_name_tail.rs
@@ -889,12 +889,7 @@ pub(crate) fn get_field_by_name_object_tail(
                 if key_bytes == b"length" {
                     return JSValue::number(crate::array::js_array_length(arr) as f64);
                 }
-                // #9192: `arr.__proto__` IS the array's `[[Prototype]]` (the
-                // spec models it as an `Object.prototype` accessor returning
-                // `[[GetPrototypeOf]](this)`) — the same shape the closure arm
-                // above resolves off the static-prototype side table. Without
-                // it a retargeted array reported the WRONG object here while
-                // `Object.getPrototypeOf(arr)` reported the right one.
+                // #9192; see `array_retargeted_proto::array_proto_slot`.
                 if key_bytes == b"__proto__" {
                     return super::array_retargeted_proto::array_proto_slot(obj);
                 }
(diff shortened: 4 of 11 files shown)


proggeramlug:
Not merging — three separate blockers, one of which I may have contributed to, so I'll be explicit about which is which.

### 1. It conflicts with #9168, which merged first

Both edit the same region of `typed_feedback/guards.rs`. #9168 added a standalone early return:

```rust
if crate::object::class_prototype_fast_guard_invalidated_for_method(method_guard_slot) {
    return (shape_addr, class_id, gc_type, name_hash, false);
}
```

and this PR *removes* it, having folded the same call into the combined guard below (`|| class_prototype_fast_guard_invalidated_for_method(method_guard_slot)`). I resolved in favour of your side, reasoning the folded check subsumes the standalone one. **That resolution is mine, not yours** — the two findings below were measured on top of it, so please re-check them after rebasing yourself rather than taking them as given.

### 2. The runtime test binary crashes

```
test collection_iter_object::fused_for_of_tests::fused_next_routes_other_iterators_through_the_generic_arm ...
process didn't exit successfully: … (exit status: 1)
note: test exited abnormally
```

Abnormal exit, not an assertion failure — so the process died rather than a check failing. That's in the fused-iterator path, which is adjacent to the method-dispatch guards this touches.

### 3. `check_file_size.sh` is red

`crates/perry-runtime/src/object/field_get_set/get_field_by_name_tail.rs` is **2002** lines, two over the cap. That file is at the limit already, so anything added to it needs a split — `index_get` and `to_string` both got one this week if you want a template.

### On the actual goal

I filed #9131, so I checked whether this closes it. With my resolution applied the monkey-patching probe goes **7/23 → 6/23** — one row better, and **case 4 regresses** (`C`/`D`/`E` rotated through one call site: `[6,60,6]` becomes a `TypeError`). The headline cases are still wrong:

```js
function run(o, n) { let s = 0; for (let i = 0; i < n; i++) s += o.m(); return s; }
const p = new P(); run(p, 3);                       // 3
P.prototype.m = function () { return 100; };
run(p, 3);                                          // node: 300, still 3 here
```

So whatever the crash and the conflict are doing, the observation half of #9131 isn't landing yet. Given #9168 is now in main, a rebase from you would be more reliable than me continuing to guess at the intended shape — you know which of the two guard placements is load-bearing.

Happy to re-run the full 23-shape probe plus the suites once it's rebased.

proggeramlug:
Rebased onto current `main` and re-validated. **Two of my three earlier concerns were wrong and I withdraw them; the third is confirmed and is the blocker.**

**Withdrawn — the "case 4 regression" is pre-existing.** I A/B'd by reverting this PR's runtime files and rebuilding both arms:

| probe row | `main` | this PR | node |
|---|---|---|---|
| existing instance after `C.prototype.m = …` | replaced ✓ | replaced ✓ | replaced |
| **fresh** `new C()` after the replacement | **orig ✗** | **orig ✗** | replaced |
| per-instance override wins over class getter | **classG ✗** | protoG ✓ | protoG |
| miss on the custom chain must not resurrect the class surface | **classG ✗** | undefined ✓ | undefined |

So this PR takes that probe from 3 divergences to 1, and the remaining one is on `main` too. I was wrong to call it a regression — apologies for the wasted round trip. I've filed the fresh-instance case separately.

**Withdrawn — the file-size cap is handled.** I split the per-instance override lookup into `object/field_get_set/prototype_override.rs` (both own-key misses ask the same question, so it lives once instead of twice) and reduced the call-site comments that now duplicate the helper's doc. `get_field_by_name_tail.rs` is at 1998; `check_file_size.sh` passes. Also resolved the `typed_feedback/guards.rs` conflict against #9168 by taking your folded form — it ORs `class_prototype_fast_guard_invalidated_for_method` into the combined guard, which is equivalent to the standalone early return now on `main`. Flagging that as my resolution, not yours.

**Confirmed and blocking — the runtime test binary aborts.**

```
collection_iter_object::fused_for_of_tests::fused_next_routes_other_iterators_through_the_generic_arm
TypeError: Method %IteratorPrototype%.next called on incompatible receiver
process didn't exit successfully (exit status: 1)
```

A/B on the same tree, only this PR's `perry-runtime` files swapped:

- `main`: `test result: ok. 1 passed`, exit 0
- this PR: test exits abnormally, exit 1

The test is plain — allocate a 2-element array, `array_values_iter`, three `js_for_of_next` steps. That error text comes from `iterator_helpers.rs`, and its doc comment names the invariant this appears to cross: per #7576, `iterator_step` must read `next` as an **own** field, because every built-in iterator inherits a `next` **thunk** that resolves its receiver from `js_implicit_this_get()`. If an inherited `next` reaches that call site, the thunk runs with whatever `this` is left in the thread-local — which produces exactly this message.

What I could not settle is how your change reaches it. `js_object_get_own_field_or_undef` is genuinely own-only (it walks the keys array and returns `undefined` otherwise), and `OBJECT_META_FLAG_PROTO_OVERRIDE` is only set when `instance_override` is true, which a built-in iterator's prototype installation shouldn't be. So the path from your two `inherited_field_if_overridden` early returns to this thunk isn't one I can name, and guessing at a fix in iterator dispatch is how #9019/#9066 happened.

That's why I'm handing back rather than patching: everything else here is ready and the behaviour change is a clear improvement, but this needs the intent behind the two early returns. Worth checking whether `resolve_inherited_field` can reach a `%…IteratorPrototype%` singleton for a receiver that has no explicit override — that's the shape that would produce this.

Rebased branch with the split, the conflict resolution and the file-size fix is pushed here, so only the abort is left.

proggeramlug:
Merged. I found the abort I handed back to you, and it wasn't in your new code — it was a latent misclassification your change was the first thing to make reachable.

**Root cause.** `attach_iterator_prototype` → `chain_to` called `object_set_static_prototype`, the `Object.setPrototypeOf` variant, which sets `OBJECT_META_FLAG_PROTO_OVERRIDE`. So **every** built-in array / Map / Set / String / RegExp-string iterator has answered `object_has_prototype_override(…) == true` since that flag existed.

That was harmless while nothing keyed off it. Your `js_native_call_method` block keys off exactly that, and reasonably: an override means the per-instance chain is authoritative, so resolve the method by ordinary inheriting lookup. For an iterator, that lookup returns the `%…IteratorPrototype%` `next` **thunk**, which per #7576 resolves its receiver from `js_implicit_this_get()` rather than the `this` you bound — hence `Method %IteratorPrototype%.next called on incompatible receiver`, and the test binary exiting 1.

**The fix is one line, in the classification rather than in your block.** `chain_to` now uses `object_link_class_default_prototype`, which is precisely what that function documents — "link a fresh instance to its class-DEFAULT prototype … a chain identical for every instance of the class." Attaching `%ArrayIteratorPrototype%` to a fresh array iterator is that, not a user `setPrototypeOf`. The prototype is still recorded either way; only the override flag and the plan-cache flush differ, and as a bonus iterators no longer flush the store-plan cache on every allocation.

Your two `inherited_field_if_overridden` early returns were firing on every iterator for the same reason, so this fixes both sites, not just the dispatch one.

**Withdrawing my two other objections from the handback**, both of which were wrong:

- The "case-4 regression" is **pre-existing**. A/B with your runtime files reverted: `main` shows 3 divergences on my probe, this PR shows 1, and the survivor (a *fresh* `new C()` not seeing a replaced prototype method) fails identically on `main`. Filed as **#9239**.
- The file-size cap is handled — I split the per-instance override lookup into `object/field_get_set/prototype_override.rs` (both own-key misses ask the same question) and trimmed call-site comments that now duplicate the helper doc. Also resolved the `typed_feedback/guards.rs` conflict against #9168 by taking your folded form, which ORs the invalidation into the combined guard — equivalent to the standalone early return on `main`. Flagging that resolution as mine.

**Validation:** `perry-runtime` 2867 passed / 0 failed at `RUST_TEST_THREADS=1` (was exiting 1); `perry-codegen` 31 suites / 0 failures; all 60 lint gates green. Iterator probe — `for…of`, manual `next()` to exhaustion, spread, `Array.from`, Map `entries`/`keys`/`values`, Set, string iteration over non-ASCII, `a.entries()`, an OWN `next` shadowing the thunk (#9019), `matchAll`, and `getPrototypeOf(iter).next` — all byte-identical to node 26.5.1. Your own subject still holds: per-instance override wins over the class getter, and a miss on the custom chain no longer resurrects the class surface.

Sorry for the round trip on the two withdrawn points — the A/B I should have run before the first handback is what settled both.

PR #9171, opened 2026-08-30
perf(codegen): inline masked string array lengths

### LINKED ISSUE #9160: String .length is a runtime call — 15.6x vs node; SSO length is already in the box
## Summary

Reading `.length` on a string emits a call to `js_value_length_property_f64` when the receiver is an element of a `string[]`. With a varying receiver (so nothing can be hoisted) it measures **5.50 ns/op against node's 0.99 — 5.6x**. For a short (SSO) string the length is literally bits in the NaN box; for a heap string it is a header field. Neither needs a call.

`.length` is one of the most common expressions in real code — every `for (let i = 0; i < s.length; i++)` pays it once per iteration — so this is a wide, low-ceiling cost rather than a niche one.

## Repro

```ts
function mk(n: number): string { let s = ""; for (let i = 0; i < n; i++) s += "x"; return s; }
const strs: string[] = [mk(2), mk(3), mk(4), mk(5)];
const R = 300_000;

// varying receiver, so the read cannot be hoisted out of the loop
function varying(): number {
  let t = 0;
  for (let i = 0; i < R; i++) { t += strs[i & 3].length; }
  return t;
}
```

## Measurement

**Corrected.** The numbers first filed here (15.47 vs 0.99) came from a 300k-iteration probe that proved unstable — the same binary measured 16.65 and then 7.38 ns/op on consecutive runs. Re-measured at 5M iterations with 7 inner repetitions taking the minimum, the result is stable to 0.01 ns across three separate runs, and the gap is smaller than first reported.

Quiet Mac mini, node v26.5.1, ns/op, medians of 3 runs:

| | perry | node |
|---|---|---|
| `strs[i & 3].length`, `strs: string[]` | 5.50 | 0.99 |
| `.length` on a `string` parameter | 4.06 | 0.98 |
| control: identical loop, no `.length` | 0.05 | 0.62 |

So the honest figure is **5.6x on the array-element receiver and 4.1x on a parameter**, not the 16x first filed. The control row matters for reading these: perry's empty loop is 0.05 against node's 0.62, so perry wins the loop itself and the entire difference is in the `.length` read — which per-operation is a larger multiple than the totals suggest.

`objdump` of the loop shows the call on the array-element receiver:

```
bl <_js_value_length_property_f64>
```

## How this was found, and what it is *not*

It surfaced while measuring string concatenation, which looked 25x slower than node. That framing was wrong. Decomposing it:

| | perry | node |
|---|---|---|
| `a.length + b.length` — **no concatenation at all** | 15.17 | 0.53 |
| `const s = a + b; t += s.length` | 19.86 | 0.53 |
| `const s = a + b; if (s === "zzzz") …` — concat + **compare**, no `.length` | 8.44 | 8.89 |

The concatenation itself costs about 4.7 ns on top of the two length reads. And on the row that concatenates without reading `.length`, **perry is faster than node** (8.44 vs 8.89). So string concatenation is not the problem this benchmark appeared to show; the length reads on either side of it were.

Perry does still copy eagerly where node uses a lazy representation — node is flat at ~0.53 ns for concat results from 4 to 128 characters while perry scales 25 → 50 ns — but that is a separate, structural matter and is not what dominates the small-string case.

## The inline path already exists — it is the receiver that fails to reach it

`expr/property_get.rs:436` already emits exactly the 3-arm dispatch this issue would otherwise propose: SSO → inline length-byte extract (`lshr 40; and 0xFF`), heap `STRING_TAG` → `load i32` of `utf16_len`, anything else → `js_value_length_property_f64`. It is gated on `static_string_lowering_enabled()`, which is **on by default** (`expr/repsel_gates.rs:88` — only `0`/`off`/`false` disables it), and on `is_string_expr(ctx, object)` — the receiver's static type.

So the gap is receiver classification, not a missing lowering. Measured across receiver shapes (varying receiver in every row, so nothing hoists):

| `.length` receiver | perry | node |
|---|---|---|
| `strs[i & 3]` where `strs: string[]` | 5.50 | 0.99 |
| `string` parameter | 4.06 | 0.98 |

(An earlier revision of this issue listed four receiver shapes at 6-16 ns from the short probe; those figures were noise-dominated and are withdrawn. The two rows above are the re-measured ones.)

Two things follow. An element of a `string[]` is not classified as a string expression — disassembly confirms it is the only one of the four shapes tested that still emits `js_value_length_property_f64` — and that is the commonest shape in real code. But the parameter receiver *does* reach the inline arms and is still 4.1x node, so classification alone will not close this; something on the read path costs more than the three-instruction SSO extract the fast arm implies.

One dead end worth recording so nobody repeats it: the classification gap is **not** in `is_string_expr`, which already has an `IndexGet` arm admitting `Array(String)` receivers (`type_analysis/strings.rs:706`). It fails one level down, in `static_type_of`, whose `LocalGet` arm consults only `local_type_hint` — empty for a module-level binding. Extending it to fall back to `ctx.module_global_proven_types` was measured and made no difference: a module-level `const strs: string[]` is not in that map, which holds proven numeric/typed-array facts rather than declared element types.

The useful next step is therefore to check `is_string_expr` against an `IndexGet` on a `string[]`-typed receiver first, and to re-measure with an isolated probe before assuming the inline arms are enough.

## Related

Same family as #9157 (a two-leaf `+` calling a helper where an inline guarded op would do), fixed in #9159. The `js_write_barrier_root_nanbox` pairs and the `js_dynamic_string_or_number_add` visible in the disassembly above are that issue, not this one.

proggeramlug:
## Summary

- version masked `string[]` length-accumulation loops behind a one-time plain-array, bounds, element-tag, and accumulator guard
- lower covered array reads to raw boxed-slot loads and proven string lengths to SSO/heap-only inline dispatch
- preserve the generic loop for erased annotation lies, descriptors, prototype pollution, holes, and non-number accumulators

## Performance

On `perrymaster.skelpo.net`, the issue-shaped `strings[i & 3].length` benchmark improves from 5.58 ns/access to 1.14 ns/access (4.9x faster). Node measures 0.63 ns/access on the same host, reducing the gap from about 8.9x to 1.8x.

## Testing

- `cargo test -p perry-codegen --test string_array_length_9160`
- `cargo test -p perry-runtime typed_feedback --lib`
- parity: `test_gap_string_array_masked_length_9160`
- `scripts/local_binding_type_audit.py`
- `scripts/check_test_registration.py`

Fixes #9160

## Summary by CodeRabbit

- **Performance**
  - Improved performance for masked indexed access to string-array elements.
  - Accelerated loops that accumulate string lengths, with benchmark results showing up to 4.9× faster access.

- **Bug Fixes**
  - Preserved correct behavior for empty, multibyte, and invalid string elements through safe fallback handling.
  - Improved support for repeated indexed string-array access and mixed accumulator values.

- **Tests**
  - Added coverage for optimized string-array length calculations and runtime edge cases.


proggeramlug:

20 files, all covered by the combined codegen and runtime suites above.

Issue #9173, opened 2026-08-30
Buffer.isBuffer(new Uint8Array(...)) returns true — registry conflates Buffers with Uint8Arrays


proggeramlug:
## Summary

`Buffer.isBuffer(new Uint8Array([1, 2]))` returns **`true`**. Node returns `false`. A plain `Uint8Array` is not a Buffer, and code that uses `Buffer.isBuffer` to tell the two apart — which is its entire purpose — takes the wrong branch.

Separately, `Buffer`'s prototype chain does not reach `Uint8Array.prototype`.

## Repro

```ts
const b = Buffer.from([1, 2, 3]);
const u = new Uint8Array([4, 5]);
console.log(Buffer.isBuffer(b));
console.log(b instanceof Uint8Array);
console.log(Buffer.isBuffer(u));
console.log(u instanceof Uint8Array);
console.log(Object.getPrototypeOf(Object.getPrototypeOf(b)) === Uint8Array.prototype);
```

| | perry | node |
|---|---|---|
| `Buffer.isBuffer(buf)` | true | true |
| `buf instanceof Uint8Array` | true | true |
| **`Buffer.isBuffer(u8)`** | **true** | **false** |
| `u8 instanceof Uint8Array` | true | true |
| **`getPrototypeOf(getPrototypeOf(buf)) === Uint8Array.prototype`** | **false** | **true** |

## Root cause

`js_buffer_is_buffer` (`crates/perry-runtime/src/buffer/query.rs:72`) is exactly a registry membership test:

```rust
if is_registered_buffer(addr as usize) { 1 } else { 0 }
```

`new Uint8Array(...)` registers into that same `BUFFER_REGISTRY`, so the predicate cannot distinguish the two. The information needed is already recorded — `mark_as_uint8array` (`buffer/header.rs`) exists precisely to note which registered addresses came from the `Uint8Array` constructor, and its doc comment says so:

> Mark this buffer as one that came from `new Uint8Array(...)` so it formats as `Uint8Array(N) [ ... ]` rather than `<Buffer ...>`.

It is consulted for *formatting* but not for `isBuffer`.

## Suggested fix

```rust
if is_registered_buffer(addr) && !is_uint8array_buffer(addr) { 1 } else { 0 }
```

[...]

Issue #6009, opened 2026-07-04
`JSON.stringify` is too slow then nodejs


zaaack:
## What happened

`JSON.stringify` on a small nested object (100 iterations) took 3388ms in compiled Perry binary, vs 32ms in native Node.js — a **100x regression**.

All other runners (tsx, tsx-transpiled node) cluster within 30–38ms.

## What you expected

A native binary should match or beat V8 JIT on `JSON.stringify`, not be 100x slower.

## Minimal reproduction

```typescript
const obj = { a: 1, b: 'foo', c: [1, 2, 3], d: { x: true } };

// Run 100 times
for (let i = 0; i < 100; i++) JSON.stringify(obj);
```

Command you ran:

```bash
perry compile benchmark.ts -o dist/benchmark-perry
./dist/benchmark-perry
```

## Environment

- **Perry version**: 0.5.1220
- **Host OS**: linux
- **Target**: native
- **Installed via**: (from path `/usr/local/bin/perry`)

## Diagnostic output

```
[json-stringify-100] 3388ms | rss=1396MB | heap=287MB
```

For comparison, same workload via Node.js:

```
[json-stringify-100] 32ms | rss=257MB | heap=33MB
```

## Anything else

The object is static and trivially small (4 keys, shallow nesting). The problem compounds — RSS grows to 1.4GB and heapUsed to 287MB by the end of the benchmark suite, suggesting the serializer may be allocating without releasing.

proggeramlug:
@zaaack Thanks for reporting this and your other 2 tickets! Looking into it!

proggeramlug:
@zaaack Thanks again, fixed in the tagged commit now.

We have been so busy getting Perry to compile complex JS/TS that some of the performance went crazy but unnoticed. :)

PR #9176, opened 2026-08-30
perf(runtime): registry probes stop paying a hash and a mutex to say "no" (5.3% of cc --help)

proggeramlug:
From the `cc --help` profile: `is_registered_buffer_slow` 3.53% and `is_uint8array_buffer_slow` 1.74%, together the largest entries in the registry-probe group.

## What was wrong

Both probes have an idle-latch fast path, and **each latch stops helping the moment a program registers its first buffer**. After that, all ~216 probe sites pay a TLS access, a `RefCell` borrow and a hash lookup to ask whether an arbitrary pointer is a buffer — and almost every caller is asking about something that is not one.

`is_uint8array_buffer_slow` was worse: on every thread-local **miss** — that is, on every value that is not a Uint8Array — it took the **global mutex**. That is precisely the cost `UINT8ARRAY_EVER_MARKED` was introduced to avoid, and its own comment says so:

> before the latch this locked `external_uint8arrays()` on EVERY thread-local miss, i.e. on every non-Uint8Array value, in every process — the one probe in this family whose miss cost a lock rather than a hash.

That cost returned in full as soon as any program marked its first Uint8Array. `is_registered_buffer_slow` already had the gate (`EXTERNAL_BUFFERS_NONEMPTY`); this probe simply never got it.

## Two changes

**An address range, thread-local like the registry it guards.** Every registration widens it before inserting, so an address outside it cannot be in the set — rejecting is sound, and accepting merely falls through to the lookup that was already there. Thread-local rather than atomic specifically so there is no ordering to reason about.

Both registries are inserted into at exactly **one** site — `register_buffer` and `mark_as_uint8array`, both widened here. Every other use is a removal or an iteration, and a removal leaving the range too wide is harmless: the address just reaches the lookup, which correctly says no. (I checked all of them; that audit is the load-bearing part of the soundness argument, since a missed insertion site would be a silent false negative.)

**The missing emptiness latch** for the external Uint8Array registry, armed before the insert for the same reason `js_buffer_register_external` arms its own before publishing.

## Measurement

`JSON.stringify` over a 300-row object graph — the shape that probes every pointer it serializes (#6009) — with a Buffer allocated first so the latches are armed. Best of 3, quiet Mac mini:

| | ms/op |
|---|---|
| armed, before | 0.2340 |
| armed, after | **0.2281** |
| idle — latches never armed, i.e. the floor | 0.2098 |

About a quarter of the arming penalty, from the range filter alone. The mutex removal is on top of that and does not show in a single-threaded probe at all.

**Pre-registered cc prediction**, since the last several perf changes were flat at app level and this one should not be: `is_registered_buffer_slow` and `is_uint8array_buffer_slow` should drop toward zero in the `cc --help` profile. At 3.53% and 1.74% that is measurable, unlike the arithmetic changes. If they don't move, the mechanism is wrong and I want to know.

## Behaviour

A buffer-identity differential — `Buffer.isBuffer`, `instanceof Uint8Array`, `toString`, `JSON.stringify`, `concat`, `subarray`, buffers allocated *after* 200 unrelated objects (so their addresses fall outside any early range), and 200 interleaved non-buffers — produces **byte-identical output with the filter on and off**. perry-runtime lib 2841/0.

That differential also surfaced a pre-existing defect, filed separately as **#9173**: `Buffer.isBuffer(new Uint8Array(...))` returns `true`, because `js_buffer_is_buffer` is exactly this registry test and `new Uint8Array` registers into the same set. It is independent of this change — both arms here agree with each other and differ from node identically.

## Gates

`-D warnings` clean; perry-hir 591/0; perry-codegen 1841/0; perry-runtime lib 2841/0; census, address-classification, file-size and raw-handle-debt lints all pass with none raised. Integration: 7/7, 5/5, 7/7, 2/2, 3/3, 3/3.

## Kill switch

`PERRY_BUFFER_RANGE_FILTER=0` restores the unconditional lookup.

## Summary by CodeRabbit

* **Performance**
  * Improved buffer and typed-array detection by quickly filtering out addresses that cannot match, reducing unnecessary lookups.
  * Added an opt-out environment setting for the address-range optimization.

* **Bug Fixes**
  * Fixed WebCrypto keys not being recognized as typed-array buffers across threads.

* **Tests**
  * Added coverage confirming external typed arrays and WebCrypto keys are visible across threads.

### DIFF
--- a/changelog.d/9176-registry-probe-prefilter.md
+++ b/changelog.d/9176-registry-probe-prefilter.md
@@ -1,0 +1,29 @@
+`is_registered_buffer` and `is_uint8array_buffer` sit on generic paths — every
+untyped element access reaches the second one — and both had an idle latch that
+stops helping the moment a program registers its first buffer. After that, all
+~216 probe sites paid a TLS access, a `RefCell` borrow and a hash lookup to ask
+whether an arbitrary pointer was a buffer, and almost every caller was asking
+about something that was not. Together they were 5.3% of `cc --help`.
+
+Each thread-local registry now tracks the smallest and largest address ever
+inserted, as a conservative filter in front of the hash. The range only widens
+and every registration extends it before inserting, so an address outside it
+cannot be in the set: rejecting is sound and accepting falls through to the
+lookup that was already there. `PERRY_BUFFER_RANGE_FILTER=0` restores the
+unconditional lookup.
+
+`is_uint8array_buffer_slow` also took the process-global mutex on every
+thread-local miss — the one probe in this family whose miss cost a lock rather
+than a hash. It now consults that registry only behind a
+`EXTERNAL_UINT8ARRAYS_NONEMPTY` latch, matching the gate
+`is_registered_buffer_slow` already had.
+
+That latch made the two inserters into the global registry a matched pair, and
+`js_buffer_mark_as_crypto_key_external` was not arming it — so a WebCrypto key
+registered on one thread became invisible to `is_uint8array_buffer` on every
+other thread: the address was in the registry and the probe answered "no". The
+registry is process-global precisely to be visible across threads, and the
+thread-local set that masks the bug on the registering thread covers no other.
+Both inserters now go through one helper that arms before inserting, and
+`buffer/header_latch_tests.rs` asks from a fresh thread, which is the only
+vantage point where the thread-local set cannot answer first.
--- a/crates/perry-runtime/src/buffer/header.rs
+++ b/crates/perry-runtime/src/buffer/header.rs
@@ -21,7 +21,7 @@ fn buffer_payload_size(capacity: usize) -> usize {
 /// Since BufferHeader has the same layout as ArrayHeader (no type_id field),
 /// we track buffer pointers separately to distinguish them from arrays.
 use crate::fast_hash::{new_ptr_hash_map, new_ptr_hash_set, PtrHashMap, PtrHashSet};
-use std::cell::RefCell;
+use std::cell::{Cell, RefCell};
 use std::collections::{HashMap, HashSet};
 use std::sync::{Mutex, OnceLock};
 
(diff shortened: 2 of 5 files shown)


proggeramlug:
Merged with one fix pushed onto the branch.

**The mechanism is sound.** I checked the range filter's stated invariant rather than taking it: `BUFFER_REGISTRY` has exactly one inserter (`register_buffer`) and `UINT8ARRAY_FROM_CTOR` exactly one (`mark_as_uint8array`), and this PR widens the range in both before the insert. Removals (`finalize_collected_dead_buffer`) only leave the range wider than necessary, which costs a lookup and can't deny a live address. The filter is therefore exactly as sound as the address-keyed registry it fronts — and if addresses could shift under that registry, `main` would already be broken, so this adds no new assumption.

**What I fixed:** the new `EXTERNAL_UINT8ARRAYS_NONEMPTY` latch turns the process-global registry into something only visible to inserters that arm it — and `js_buffer_mark_as_crypto_key_external` inserts at what was line 481 without arming. The `EXTERNAL_BUFFERS_NONEMPTY` store right above it *is* armed on both its paths; this was the second of two, missed. The effect is a WebCrypto key registered on one thread answering `false` from `is_uint8array_buffer` on every other — the address sits in the registry and the probe skips the mutex and denies it. That registry is process-global precisely to be visible across threads, and the thread-local set that hides the bug on the registering thread covers no other one.

Both inserters now go through a single `register_external_uint8array` helper that arms before inserting, so a third can't forget.

**The test is discriminating, not decorative.** It has to ask from a *fresh* thread — on the registering thread `mark_as_uint8array` puts the address in the thread-local set, which answers first and masks the bug entirely. Confirmed by restoring your `header.rs` and keeping the test: `crypto_key_external_registration_arms_the_uint8array_latch` FAILED, `plain_external_uint8array_registration_arms_the_latch` passed — the failure isolates to the one path, which is what I wanted before trusting a green.

**Validation:** `perry-runtime` 2846 passed / 0 failed at `RUST_TEST_THREADS=1`; all 60 lint gates green. Also added the two new `Cell<(usize, usize)>` holders to `scripts/gc_runtime_root_holders.json` — `gc_runtime_root_holders.py` flags them under rule T, correctly, since they hold addresses. Verdict is `not_a_gc_pointer`: the bounds are only ever compared, never dereferenced, and because the range only widens a stale bound can only admit more addresses to the real lookup, never deny a registered one.

One thing that is **not** yours: a node-differential probe over these paths turned up `Buffer.isBuffer(new Uint8Array(4))` → `true`, where node says `false`. I A/B'd it against a clean `main` build — your branch's output is byte-identical to main's on all 15 probe lines, so it's pre-existing. Filed as #9179. Everything else in that probe (hex/base64 round trips, `concat`, `compare`, `DataView` aliasing over a shared `ArrayBuffer`, element access) matches node byte-for-byte.

`PERRY_BUFFER_RANGE_FILTER` is a runtime knob, so it's correctly outside `codegen_env_vars_are_build_cache_inputs` — no build-cache registration needed.

PR #9177, opened 2026-08-30
perf(runtime): symbol probe stops taking the global mutex to say "no" (third latch instance)

proggeramlug:
Stacked on #9176 — its commit shows here until that merges. Third instance of the same defect.

## The pattern, now confirmed three times

A latch that answers *"has this **ever** happened?"* is only a fast path for programs where it never happens. **The moment it does, the latch is pure overhead and something narrower has to take over.**

- `is_registered_buffer` — falls to a TLS + `RefCell` + hash lookup (#9176)
- `is_uint8array_buffer` — falls to the **global mutex on every miss** (#9176)
- `is_registered_symbol` — falls to the **global mutex on every probe** (here)

Every program that touches a well-known symbol creates one, so this latch stops discriminating almost immediately. After that `is_registered_symbol_slow` takes the process-global mutex for every probe, and the overwhelming majority are asking about pointers that are not symbols at all. The `cc --help` profile puts it at 0.80%, 1.9% at higher resolution.

I'd note the pattern in review terms: the remaining `*_EVER_*` flags guarding a global structure are where the fourth instance will be, and it's cheaper to grep for them than to wait for a profile to surface each one.

## The change

An address range, widened **before** the insert — the same ordering, and for the same reason, as the latch arm directly above it, whose comment already says:

> Arm before taking the lock, so the entry is never reachable while the latch still reads idle.

A pointer outside the range cannot be in the set, so rejecting is sound; accepting falls through to the lookup that was already there. Atomic rather than thread-local because this registry is global, unlike #9176's.

`register_symbol_pointer` is the **only** insertion site — that audit is the load-bearing part of the argument, since a missed insertion would be a silent false negative. Death pruning removes entries without narrowing the range, which is harmless: those pointers reach the lookup, which correctly says no.

## Numbers

Quiet Mac mini, medians of 3. A loop doing `typeof` and `Map` lookups over pointer-tagged values — string keys fail `js_is_symbol`'s tag check before ever reaching the registry, so a string-keyed benchmark measures nothing here, which cost me a probe to discover:

| | µs/iter |
|---|---|
| idle — latch never armed, the floor | 4.75 |
| armed, filter **on** | **5.12** |
| armed, filter off | 6.31 |

**19% off the armed path**, and the arming penalty falls from 33% to 7.8%.

## Gates

`-D warnings` clean; perry-runtime lib **2844/0**, including all 15 `registry_latch_probes` — among them `symbol_is_found_after_the_idle_fast_path_ran`, `unregistered_address_misses_every_probe`, and the cross-thread `shared_array_buffer_allocated_on_another_thread_is_found_here`. Census, address-classification, file-size and raw-handle-debt lints all pass with none raised.

## Kill switch

`PERRY_SYMBOL_RANGE_FILTER=0` restores the unconditional mutex acquisition.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved symbol recognition performance by quickly rejecting pointers outside the known symbol address range.
  * Preserved symbol behavior after garbage collection moves symbols in memory.
  * Fixed symbol-related operations, including `typeof`, symbol-keyed properties, and iterator dispatch, after memory movement.
  * Added a configuration option to disable address-range filtering when needed.
* **Tests**
  * Added regression coverage for symbol recognition after copying garbage collection.

### DIFF
--- a/changelog.d/9177-symbol-registry-range-filter.md
+++ b/changelog.d/9177-symbol-registry-range-filter.md
@@ -1,0 +1,31 @@
+`is_registered_symbol` sits on property/method dispatch, so it is asked about
+almost every pointer a program touches. `SYMBOL_EVER_REGISTERED` answered the
+idle case in one atomic load, but stopped discriminating the moment a program
+created its first symbol — and every program that touches a well-known symbol
+creates one. After that each probe took the process-global registry mutex,
+overwhelmingly to say "no".
+
+The registry now carries the smallest and largest pointer ever registered, and
+a pointer outside that range is rejected without the lock.
+`PERRY_SYMBOL_RANGE_FILTER=0` restores the unconditional acquisition.
+
+Unlike the buffer registries this pattern was lifted from, `SYMBOL_POINTERS` is
+a GC ROOT registry: the collector re-keys an entry to the symbol's new address
+every time it moves one. A range filter is sound only while every path that
+puts a pointer in the set widens the bounds first, and there are three — the
+registration, a per-slot forwarding rewrite, and the bulk rewrite that the
+copying minor actually drives. Only the first widened, so a symbol evacuated
+past the bounds its own allocation established became a live, registered symbol
+that the probe reported as "not a symbol", losing `typeof`, symbol-keyed
+property lookup and `Symbol.iterator` dispatch for it.
+
+Widening and inserting are now a single operation (`insert_symbol_pointer_in_set`),
+which is what makes the three sites uniform rather than three chances to forget.
+
+`test_copying_minor_keeps_moved_symbol_visible_to_the_range_filter` covers it
+through the public probe after a real copying minor. The existing sibling test
+asserts membership with `test_symbol_pointer_root_contains`, which reads the set
+directly and so could not see this — the entry IS in the set. The new test also
+resets the range around a single registration, because a wide range left by
+earlier tests in the same binary would admit the moved address by luck and make
+the assertion vacuous.
--- a/crates/perry-runtime/src/gc/tests/copying_side_tables.rs
+++ b/crates/perry-runtime/src/gc/tests/copying_side_tables.rs
@@ -155,6 +155,62 @@ fn test_copying_minor_rewrites_symbol_side_table_roots_and_lookups() {
     }
 }
 
+/// The range filter in front of `SYMBOL_POINTERS` must admit a symbol's NEW
+/// address after the collector moves it.
+///
+/// `is_registered_symbol` rejects an out-of-range pointer WITHOUT consulting
+/// the set, so a moved symbol whose new address fell outside the bounds its
+/// allocation established would be a live, registered symbol that the probe
+/// reports as "not a symbol" — losing `typeof`, symbol-keyed property lookup
+/// and `Symbol.iterator` dispatch.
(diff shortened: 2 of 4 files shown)


proggeramlug:
Merged, rebased onto `main` (the #9176 commit this was stacked on is already there, so that patch was skipped as a duplicate) with a correctness fix pushed onto the branch.

**The bug.** This lifts the range-filter pattern from #9176, but `SYMBOL_POINTERS` differs from the buffer registries in a way that breaks it: it is a **GC root registry**. The collector re-keys an entry to the symbol's new address every time it moves one — your own doc comment for the death-pruning code says the set was "only forward-renamed on moves". A range filter is sound only while *every* path that puts a pointer into the set widens the bounds first, and there are three:

1. `register_symbol_pointer` — the one you widened
2. `rewrite_symbol_pointer_metadata_if_forwarded` — the per-slot forwarding rewrite
3. `scan_symbol_pointer_metadata_roots_mut` — the bulk rewrite, and the one a copying minor actually drives

A symbol evacuated past the bounds its own allocation established was therefore a live, registered symbol that `is_registered_symbol` reported as "not a symbol" — losing `typeof`, symbol-keyed property lookup and `Symbol.iterator` dispatch for it. Reachable on any minor that moves a symbol into a region outside the observed range.

Widening and inserting are now one operation (`insert_symbol_pointer_in_set`), which is what makes the three sites uniform instead of three chances to forget. I also routed the test-only set rebuild in `test_clear_symbol_side_table_roots` through it.

**Worth flagging about the test**, because it changed my mind mid-fix: my first patch only fixed site 2, and the new test failed anyway — which is how I found site 3. Had I trusted the fix and shipped, this would have gone in still broken, because the existing `test_copying_minor_rewrites_symbol_side_table_roots_and_lookups` passes either way: it asserts membership with `test_symbol_pointer_root_contains`, which reads the set directly and never touches the filter. The entry *is* in the set; the probe just refuses to look. `test_copying_minor_keeps_moved_symbol_visible_to_the_range_filter` goes through the public probe after a real copying minor, and resets the range around a single registration first — a wide range left by earlier tests in the same binary would otherwise admit the moved address by luck and make the assertion vacuous.

**Validation:** `perry-runtime` 2847 passed / 0 failed at `RUST_TEST_THREADS=1`; all 60 lint gates green. End-to-end probe (200 symbols, symbol-keyed lookups, `Symbol.for`/`keyFor`, `description`, spread through `Symbol.iterator`, with 60k-object churn between registration and use) matches node 26.5.1 byte-for-byte under default, `PERRY_GC_FORCE_EVACUATE=1`, and `PERRY_GC_SCHEDULE_SEED=42 PERRY_GC_SCHEDULE_RATE=1`. The unit test is the discriminating one, though — that probe can't guarantee a moved address lands outside the range, which is exactly why the unit test controls the range instead of hoping.

Added the `changelog.d/` fragment, which the branch was missing. `PERRY_SYMBOL_RANGE_FILTER` is a runtime knob, so it's correctly outside `codegen_env_vars_are_build_cache_inputs`.

Worth applying the same question to any further instance of this pattern: *is the registry a GC root registry, and if so does the forwarding rewrite widen?* That's the difference between #9176 (safe — buffer addresses are pruned on death, never re-keyed) and this one.

PR #9178, opened 2026-08-30
perf(runtime): compile regexes on first use — claude-code's 2,378 literals 232ms → 50ms at startup

proggeramlug:
A symbolized instruction profile of the claude-code CLI running `--help` — a command that prints text and exits — showed **14.6% of all instructions inside regex COMPILATION** (`regex_syntax` case-folding, `regex_automata` Thompson/determinize). For comparison, all of cc's compiled JavaScript is 0.11% of that profile.

**Cause:** a `/…/` literal lowers to `js_regexp_new` **at its evaluation site**, so module-scope literals compile at module init — and `js_regexp_new` answered "is this a SyntaxError?" by *building* the pattern (parse → HIR translate including `case_fold_simple` → Thompson NFA → meta strategy). Every regex a program **has** was compiled, not every regex it **uses**.

**Measured, one literal executed out of N** (min of 9, program's own timing):

| literals constructed, 1 used | before | after | node |
|---|---|---|---|
| 50 | 19 ms | 2 ms | 0 ms |
| 200 | 73 ms | 7 ms | 1 ms |
| 400 | 145 ms | 15 ms | 3 ms |
| **every distinct literal in cli_2.1.112.js (2,378)** | **232 ms** | **50 ms** | 5 ms |

Linear at 362 µs/literal before — the signature of "every literal compiles". Wall clock on the cc corpus 247.5 → 59.9 ms.

**Design — validation stays eager, and gets cheap.** Construction leaves the engine pointers null (published last, as the not-built flag) and `ensure_regex_compiled` installs them on first use. `SyntaxError` still throws **at construction**, matching node, but via an AST-only grammar parse (`regex_syntax::ast`, no case folding) that pays the full HIR translate only when the pattern mentions `\p`/`\P` (16 of cc's 2,378). A parser rejection is not a verdict — it falls through to the unchanged both-engines path that still owns the SyntaxError decision and the fancy-regex fallback. `RegExp.prototype.compile` (Annex B) stays eager deliberately.

**Validation:** runtime suite **2846 passed / 0 failed**. Gap fixture `test_gap_9163_lazy_regex_semantics.ts` byte-identical to node — SyntaxError still at construction *including for a never-matched invalid regex*, `.source`/`.flags` readable before first match, `/g` and `/y` `lastIndex` across the mid-life build, and per-evaluation identity (independent `lastIndex`/expandos — the RegExp analogue of #9128's closure-singleton bug). **Syntax agreement: 9,899 patterns, zero disagreements in both directions** — 3,402 distinct literals from the claude-code/pi/kimi bundles plus 6,297 mutations (truncations, deletions, injected `{2,1}`), pinned by a committed corpus and a `PERRY_REGEX_CORPUS` hook. 15/16 regex fixtures byte-identical (the 16th is one node v26.5.1 itself fails, already baselined). Match-throughput control 220 → 198 ms, so the per-op null check costs nothing. `.text` **+6,336 bytes (+0.07%)**, runtime only — codegen untouched.

**Two pre-existing findings, not caused by this** (identical before and after): 3 of cc's 2,378 literals throw `SyntaxError` in perry that node accepts — acorn's 10–15 KB `/u` identifier classes, worth its own issue. And one corner moves *toward* node: a pattern the parser accepts but whose NFA build blows the 64 MiB budget was a construction SyntaxError and is now a silent never-match at first use, which is what node does; zero occurrences in the 9,899-pattern sweep.

**Follow-up:** perry knows literal patterns at AOT compile time, so validating there would remove even the AST parse from startup, taking construction to pure allocation.

## Summary by CodeRabbit

* **Performance**
  * RegExp programs are now compiled only when first needed, improving construction efficiency while preserving behavior.

* **Bug Fixes**
  * Improved support for advanced and fallback regular-expression patterns across matching, replacement, and string operations.
  * Preserved validation errors, flags, source text, object identity, and `lastIndex` behavior.

* **Tests**
  * Added coverage for deferred compilation, syntax validation, fallback matching, cache limits, and observable RegExp semantics.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6338,6 +6338,7 @@ dependencies = [
  "perry-parser",
  "rand 0.10.1",
  "regex",
+ "regex-syntax",
  "regress",
  "resolv-conf",
  "ryu",
--- a/changelog.d/9178-lazy-regex-compilation.md
+++ b/changelog.d/9178-lazy-regex-compilation.md
@@ -1,0 +1,29 @@
+A `/…/` literal lowers to `js_regexp_new` at its evaluation site, and
+`js_regexp_new` answered "is this a SyntaxError?" by BUILDING the pattern —
+`regex_syntax` parse, HIR translate including Unicode case folding, Thompson
+NFA construction, meta-engine strategy selection. Every regex a program *had*
+was compiled, not every regex it *used*. A symbolized instruction profile of
+the claude-code CLI running `--help` — a command that prints text and exits —
+put 14.6% of all retired instructions inside regex compilation, against 0.11%
+for the whole of its compiled JavaScript.
+
+Only the program build is deferred; everything observable at construction stays
+at construction. A syntactically invalid pattern still throws `SyntaxError`
+from `js_regexp_new` / `RegExp.prototype.compile`, at the same point in the
+program, because `std_engine_syntax_ok` runs the same parser `build_std_regex`
+would run, on the same translated, flag-prefixed, REDoS-collapsed string —
+4.6 µs/pattern against 82 µs to build. Anything it rejects falls through to the
+unchanged both-engines check, so the fancy-regex fallback for
+lookbehind/backreferences still decides and still throws when both refuse.
+`.source`, `.flags`, `.global`, `.sticky` and `lastIndex` never touched the
+compiled program, and identity is unchanged. The build happens on the first
+operation that needs a matcher.
+
+| literals constructed, 1 used | before | after | node |
+|---|---|---|---|
+| 50 | 19 ms | 2 ms | 0 ms |
+| 200 | 73 ms | 7 ms | 1 ms |
+| 400 | 145 ms | 15 ms | 3 ms |
+| every distinct literal in cli_2.1.112.js (2,378) | 232 ms | 50 ms | 5 ms |
+
+Wall clock on the cc corpus 247.5 → 59.9 ms.
--- a/crates/perry-runtime/Cargo.toml
+++ b/crates/perry-runtime/Cargo.toml
@@ -120,7 +120,7 @@ diagnostics = []
 # RegExp object's identity/display layer (header, `is_regex_pointer`, `toString`)
 # stays always compiled, so console-formatting / value-to-string paths keep
 # working with no engine present.
-regex-engine = ["dep:regex", "dep:fancy-regex", "dep:regress"]
+regex-engine = ["dep:regex", "dep:regex-syntax", "dep:fancy-regex", "dep:regress"]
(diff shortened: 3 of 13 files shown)


proggeramlug:
Merged.

The thing I wanted to break is the claim that validation stays eager. If `std_engine_syntax_ok` accepts a pattern the deferred build later rejects, then a `SyntaxError` that used to be thrown at construction becomes a **silent never-match** — `get_or_compile_regex` falls back to `[^\s\S]` rather than failing, so the failure mode would be quiet and late, which is worse than the eager error it replaced. That gap is real in principle: the AST parser you validate with does not run the HIR translate or the NFA build, and `Regex::new` can fail in both.

It holds up. Probed the patterns most likely to sit in that gap — `a{100}{100}{100}`, `(?:a{500}){500}`, `(?i)[Ā-က]{50}{50}`, `[a-\d]`, `(?<=x)y`, `(a)\1`, `(?<name>a)\k<name>`, plus genuinely invalid `[` and `a**` — and compared construction-time outcome *and* match result against node 26.5.1. The size-limit cases still throw `SyntaxError` at construction (the AST parser rejects the nested quantifier, so they fall through to the unchanged both-engines check, exactly as your doc comment says), and every case matches node byte-for-byte except one noted below. `.source`/`.flags`/`.global`/`.ignoreCase`/`lastIndex`, `exec` with `lastIndex` advance, `replace`, `split`, and `matchAll` all match too.

The one divergence is **not yours**: `/\p{L}+/` without the `u` flag matches as a Unicode property class where node reads `\p` literally. `grammar.rs` is untouched by this PR and `js_regex_to_rust` has no `u`-flag handling at all, so the translation that produces this predates you. Worth noting your own gap test sidesteps it correctly by putting the `u` flag on `\p{Bogus}` — `test_gap_9163_lazy_regex_semantics.ts` passes against node byte-for-byte.

Two things I'd call out as the right calls: extracting `flag_prefixed_pattern` so the validator and the builder cannot inspect different strings (that drift is exactly how an "eager validation" quietly stops matching what it validates), and publishing `regex_ptr` last in `build_and_install_programs` so it doubles as the built/not-built flag.

**Validation:** `perry-runtime` 2852 passed / 0 failed at `RUST_TEST_THREADS=1` (up 5 from `main`, your new tests); `perry-codegen` clean; all 60 lint gates green.

Two things I added: a `not_a_gc_pointer` verdict for `VALIDATED_PATTERNS` in `scripts/gc_runtime_root_holders.json` — `gc_runtime_root_holders.py` flags the new `thread_local!` under rule T and fails `lint` without one; the keys are owned Rust `String`s and the value is `()`, so there is no slot for the collector — and the `changelog.d/` fragment, which the branch was missing.

PR #9182, opened 2026-08-30
gc: fail closed if the native root scan runs before the stack-map index is built

proggeramlug:
Step 1 of two. This one is a **no-op by construction** — and that is the point.

## The invariant

An unbuilt stack-map index is **indistinguishable downstream** from an image with no native roots: both are empty. The consequences are not. With statepoints as the only root mechanism, scanning against an unbuilt index means the collector finds no roots on that frame, frees live objects, and corrupts the heap **with no diagnostic at all**.

`build_stack_map_index` already refuses to let a decode failure degrade into "no roots", and says why:

> both yield an empty index — but their consequences are not: with statepoints as the only root mechanism, an empty index means the collector frees live objects and corrupts the heap with no diagnostic at all.

This asserts the same property at the **consuming** end: the index is built, **or** this image carries no gc-map section at all.

## Why now, against the eager build

It cannot fire today. `js_gc_init` builds the index before any mutator code runs, so this changes no behaviour.

It exists for **step 2**: the gc-map section is decoded eagerly at every process start — 17.4 MB, 72,713 functions, 2.15M records, a 117 MB index — and `cc --version`, which never collects, pays **698M instructions and 313 MB of RSS** for an index it never reads. Deferring that build is worth having. But its correctness rests on every path into the root scan passing through whatever point builds the index, and **that enumeration could not be established**: a first attempt named "two chokepoints" and was withdrawn as unverified; tracing found at least three fan-in points (`visit_global_root_slots`, `mark_mutable_root_slots_step` via `cycle.rs`, `shadow_stack_root_scanner`) plus untraced allocation-triggered entries.

For a failure this silent, *"we found all the callers"* must not be the load-bearing claim. With this assert in place, a missed entry aborts loudly on the first test that reaches it instead of corrupting the heap.

Landing it **before** the change that needs it is deliberate: if the assert were misplaced — guarding the wrong function, or somewhere the scan does not pass — we would discover that only when it failed to catch the thing it exists to catch, which is exactly when we would be relying on it. Proving placement while nothing depends on it is cheap. Proving it afterwards is not possible.

## It found something on its first run

Written unscoped, the assert immediately caught `gc::heap_snapshot::tests::snapshot_has_real_nodes_and_edges` reaching the native root scan without going through `js_gc_init` — **a path neither the analysis nor my own tracing had enumerated.** Benign, as it happens: perry-runtime's unit tests drive the collector directly and their harness carries no gc-map section.

That is the sequencing argument turned from a principle into a measurement. We reasoned our way to "an enumeration cannot be load-bearing"; the check produced a counterexample within minutes of existing.

## Stating the invariant, not the build config

The condition is *"index built **or** no gc-map section"*, not *"initialize ran"*. That keeps the check live in **every** configuration, and makes the unit-test case pass for the right reason instead of by exemption. Scoping it out of test builds — my first version — would have removed it from precisely the configuration where the index is legitimately unbuilt, which becomes a hole the moment a test binary does carry statepoint frames.

The section query inspects the loader's image list without decoding anything, and is cached because the answer cannot change once a process has loaded its images. An inspection error answers "sections present", so an unreadable loader makes the assert **stricter**, not weaker.

## Validation

perry-runtime lib **2844/0** with the assert never firing, plus six compiled binaries run under `PERRY_GC_PROTECT_FROMSPACE` with seeded aggressive collection schedules — up to **207,024 moved objects** — none of which fire it. `-D warnings` clean; hir 591/0; codegen 1841/0; census, address-classification, file-size and raw-handle-debt lints all pass with none raised.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented premature garbage-collection scans from incorrectly releasing live objects.
  * Added safeguards to distinguish an uninitialized runtime state from applications that legitimately contain no native garbage-collection metadata.
  * Runtime now fails clearly when a required garbage-collection scan is attempted before initialization, improving reliability and diagnosability.

### DIFF
--- a/changelog.d/9182-native-root-scan-fails-closed.md
+++ b/changelog.d/9182-native-root-scan-fails-closed.md
@@ -1,0 +1,22 @@
+The native root scan resolves frame roots through the stack-map index, and an
+index that was never built is indistinguishable downstream from an image that
+genuinely has no native roots: both are empty. The consequences are not. With
+statepoints as the only root mechanism, scanning against an unbuilt index means
+the collector finds no roots on that frame, frees live objects, and corrupts the
+heap with no diagnostic.
+
+`visit_stack_map_root_slots` now asserts the property at the consuming end —
+the same degradation `build_stack_map_index` already refuses at the producing
+one. The invariant is stated as "an empty index means this image genuinely has
+no native roots", not "initialize ran", which keeps the check live in every
+configuration: `perry-runtime`'s unit tests reach the scan without `js_gc_init`
+and pass because their harness carries no gc-map section — the right reason —
+rather than by being exempted. Exempting them by build config would leave no
+check in precisely the configuration where the index is legitimately unbuilt.
+
+It cannot fire today, because `js_gc_init` builds the index eagerly before any
+mutator code runs. It is here so that it CAN fire if the build is ever made
+lazy and a path into the scan is missed, turning a silent delayed heap
+corruption into a loud abort on the first test that reaches it. Landing it
+while the build is still eager is deliberate: it proves the check sits on the
+path it claims to guard before anything depends on it.
--- a/crates/perry-runtime/src/gc/roots/stack_maps.rs
+++ b/crates/perry-runtime/src/gc/roots/stack_maps.rs
@@ -27,7 +27,7 @@ use crate::gc::telemetry::RootSourcesTraceStats;
 // the Itanium/pthread declarations, which do not exist there.
 #[cfg(not(target_os = "windows"))]
 use std::ffi::c_void;
-use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::sync::{OnceLock, RwLock, RwLockReadGuard};
 
 /// Magic and version of the compact map the compiler emits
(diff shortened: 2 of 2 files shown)


proggeramlug:
Merged.

A no-op assert is exactly the kind of change that gets waved through as harmless and turns out to be either (a) never evaluated, or (b) evaluated and able to abort a working program. So I checked both directions rather than the diff.

**Is the check actually live?** Yes — and this is the part worth stating, because "2847 tests pass" is equally consistent with the assert never being reached. I sabotaged the condition to `false && (…)` and re-ran: the suite failed, exit 101, with the assert's message. So `visit_stack_map_root_slots` genuinely is on the unit tests' path and the check really is evaluated there. Your framing — that the tests pass because their harness carries no gc-map section, which is the *right* reason rather than an exemption — holds empirically, not just by argument.

**Can it fire where it shouldn't?** It ships as `assert!`, not `debug_assert!`, so a spurious fire aborts a user's program. Ran a GC-stress workload (linked-list churn with a rolling live set, symbol-keyed properties, and typed-array allocation, so the scan runs against images that *do* carry gc-map sections) under four arms — default, `PERRY_GC_FORCE_EVACUATE=1`, `PERRY_GC_SCHEDULE_SEED=7 PERRY_GC_SCHEDULE_RATE=1`, and `PERRY_GEN_GC=0`. All exit 0, identical output, assert fired zero times. The earlier buffer and symbol differential probes from #9176/#9177 also still run clean under it. That's consistent with your claim that `js_gc_init` builds eagerly before any mutator code.

Cost is a single `Acquire` load on the initialized path — `image_has_stack_map_sections()` is only evaluated when the flag is false, so in production it never runs at all and its `OnceLock` never populates.

Two details I liked and want to name so they survive review: publishing `STACK_MAPS_INITIALIZED` *after* `rebuild()` (a reader that sees the flag can see the index it promises), and having an inspection error answer `true` so an unreadable loader makes the assert stricter rather than weaker. Failing closed on the diagnostic path too is the right default.

Added the `changelog.d/` fragment, which the branch was missing. `PERRY_GC_NATIVE_SLOT_VERIFY` appears in the diff only as surrounding context — it is already on `main`, so the GC knob kill-policy doesn't apply here; this adds no new knob.

Ready for step 2 whenever you are. The thing that would make the lazy build safe to land is this assert firing loudly in CI on a missed path, so it's worth confirming at that point that some test actually exercises a scan *before* the lazy build point — otherwise step 2 silently removes the only configuration where this can fail.

**Validation:** `perry-runtime` 2847 passed / 0 failed at `RUST_TEST_THREADS=1`; all 60 lint gates green.

PR #9185, opened 2026-08-30
perf(codegen): a throw that does not construct keeps the packed fast path (4.76 → 0.58 ns)

proggeramlug:
Third attempt at this row, and the two failures were the cost of learning why.

## Two gates, and relaxing either alone does nothing

`stmt_is_packed_f64_loop_safe` rejects `Stmt::Throw` at the HIR level. Independently, `fast_clone_call_free` scans the **emitted** clone blocks through `contains_gc_unsafe_call`, and constructing a thrown value is a call. My first attempt relaxed the HIR gate and measured nothing; my second relaxed the block scan and measured nothing. Both are needed.

## The argument

A throw is the one place a call in the loop body is harmless. Codegen already emits `unreachable` after `js_throw` (`stmt/mod.rs:591`), so control never returns to the loop — there is no second read of anything the clone cached, which is precisely what the scan exists to protect. `contains_gc_unsafe_call` now says so directly: a block whose last instruction is `unreachable` has no "after".

## Numbers, including the one that didn't move

Quiet Mac mini, node v26.5.1, ns/op, one binary switched with `PERRY_PACKED_LOOP_ABRUPT`:

| loop body | perry off | perry on | node |
|---|---|---|---|
| `throw <pre-built error>` | 4.76 | **0.58** | 0.56 |
| `throw <number>` | 4.76 | **0.58** | 0.56 |
| `throw new Error(…)` | 4.78 | 4.78 | 0.56 |

**The third row is the important one.** A throw that *constructs* its value is still rejected, because the construction is emitted in blocks that **precede** the `unreachable`-terminated one, so a per-block check cannot cover them. Admitting it needs a region property — every path from the block leads to an unwind without re-entering the loop — which is a dominance question and a larger change.

That split is what identifies the remaining blocker as a region property rather than a third unknown gate. Two attempts failed trying to establish it by reasoning; one experiment settled it.

Rethrowing a caught error, throwing a sentinel, and throwing a value are the shapes this covers.

## Validation

Differentials where the throws **actually fire**, at varying indices, with observable effects before them — a never-taken throw would exercise only the admission decision, not the emitted clone. Byte-identical to node, and identical again under `PERRY_GC_PROTECT_FROMSPACE` with seeded aggressive collection schedules.

The heavy case is the one that matters for a change admitting a call into the clone region: arrays escaping into a sink so they are not scalar-replaced, 120 rounds, throws firing every round — **identical across seeds at 193,052 moved objects**.

## Gates

`-D warnings` clean; hir 591/0; codegen 1842/0; runtime lib 2844/0; census, address-classification, file-size and raw-handle-debt lints all pass with none raised. Integration: `packed_loop_abrupt_statements` 8/8 (one new), plus 7/7, 5/5, 2/2, 3/3, 3/3.

## Kill switch

`PERRY_PACKED_LOOP_ABRUPT=0`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of unreachable code to prevent incorrect safety warnings.
  * Packed loops now support throws using already-prepared values while retaining optimized execution.
  * Throws requiring value construction continue to use the safer fallback path.

* **Tests**
  * Added coverage for throw statements in packed loops, including optimized and fallback behavior.
  * Verified that existing abrupt-control-flow safeguards remain intact.

### DIFF
--- a/changelog.d/9185-throw-keeps-packed-fast-path.md
+++ b/changelog.d/9185-throw-keeps-packed-fast-path.md
@@ -1,0 +1,24 @@
+`stmt_is_packed_f64_loop_safe` rejected every `throw` in a packed-f64 loop
+body, on the reasoning that the thrown value is typically constructed
+(`throw new Error(…)`) and the construction is a call in the loop. That is
+true of the construction, not of the `throw` — a loop that throws an
+already-built value lost the fast path for a call it never makes.
+
+`contains_gc_unsafe_call` now answers `false` for a block whose last
+instruction is `unreachable`. That is sound for the reason the whole
+call-free check exists: the check verifies at compile time that nothing in
+the fast clone can trigger a collection and move the preheader-cached
+receiver *before something reads it again*. A block ending in `unreachable`
+has no such "again" — control does not return to the loop, so no cached value
+is read after the call. `push_inst` drops anything pushed after a terminator,
+so a block's last instruction is its terminator; `instructions.last()` being
+`Unreachable` therefore means the block genuinely ends there rather than
+carrying dead code past a `br`. A `throw` that constructs its value is still
+rejected, because the construction is emitted in blocks that precede the
+terminating one.
+
+`Stmt::Try` inside the loop body remains a rejection, so the throw's handler
+can never be inside the loop being cloned — which is what would otherwise
+return control to a loop holding a stale cached pointer.
+
+bench: 4.76 → 0.58 ns.
--- a/crates/perry-codegen/src/block.rs
+++ b/crates/perry-codegen/src/block.rs
@@ -297,6 +297,9 @@ impl LlBlock {
     /// check cannot be invalidated mid-loop). Intrinsic/libm-style calls
     /// never enter the perry runtime, so they cannot trigger a collection.
     pub fn contains_gc_unsafe_call(&self) -> bool {
+        if matches!(self.instructions.last(), Some(crate::inst::LlInst::Unreachable)) {
+            return false;
+        }
         self.instructions.iter().any(|i| i.is_gc_unsafe_call())
     }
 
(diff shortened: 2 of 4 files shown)


proggeramlug:
Merged.

This one deserved more scrutiny than its size suggests, because it weakens a **shared** GC-safety predicate: `contains_gc_unsafe_call` has four consumers — the packed-f64 loop, `element_shape_loop`, `stable_packed_loop`, and the class-field inline guard that #5093 originally added it for. You reason about and test only the first, but all four get the relaxation. I convinced myself it generalises (the argument is about control never returning to the block, which is consumer-independent), but it is worth saying out loud that three of them widened without a test.

**The hole I went looking for and didn't find.** `instructions.last()` being `Unreachable` only means the block *ends* in `unreachable` if nothing can sit after a terminator — otherwise a block could be `[…gc-unsafe call…, br, unreachable]`, flow out via the `br`, and be wrongly called call-free. `push_inst` returns early when `terminated` is set, and `unreachable()` sets it, so that shape cannot be built. Closed by construction.

**The other one: unwinding.** A `throw` transfers control, so "the block ends in `unreachable`" is not by itself "control leaves the loop" — if the handler were *inside* the loop, control would come back to a fast clone holding a stale cached pointer. `Stmt::Try { .. } => false` already rejects a `try` in the loop body, so the handler is always outside the loop being cloned. That's what makes this safe, and it's load-bearing enough to be worth a comment next to the new arm — right now the reason lives one match arm away.

**Verified the subject is live rather than trusting the benchmark**, by IR-diffing two variants in the same binary:

| loop body | `packed_f64_fast` | `acc.writeback_exit` |
|---|---|---|
| `throw PRE` (value already built) | present | 2 |
| `throw new Error("x")` (constructs) | absent | 0 |

So the fast path engages for exactly the case you claim and the constructing throw still correctly falls to the generic path.

**Validation:** `packed_loop_abrupt_statements` 8/8 including your new test; `perry-codegen` 31 suites, 0 failures; `perry-runtime` 2852 passed / 0 failed; all 60 lint gates green. A differential probe (throw of a pre-built error, throw of a computed number, array re-read *after* the catch to catch a stale pointer, and the whole thing under allocation churn) matches node 26.5.1 byte-for-byte under default, `PERRY_GC_FORCE_EVACUATE=1`, and `PERRY_GC_SCHEDULE_SEED=3 PERRY_GC_SCHEDULE_RATE=1`.

Ran `cargo fmt` on the new `matches!` and added the `changelog.d/` fragment, which the branch was missing.

PR #9187, opened 2026-08-30
perf(runtime): register function names and sources without validating UTF-8

proggeramlug:
Module init calls `js_register_function_name` and `js_register_function_source` once per function a bundle **contains** — 72,713 of them for claude-code — and each validated its bytes with `core::str::from_utf8` before storing them. Source text is registered to serve `Function.prototype.toString()`, which most programs never call.

The bytes are now stored raw and decoded on read. Nothing observable changes: a non-UTF-8 name was previously dropped at registration and is now dropped at lookup, so both report "no name". The reads already returned owned `String`s, so the check costs nothing extra where it moved to.

## Measured — and smaller than the profile suggested

A `cc --help` profile attributes 2.17% to `core::str::converts::from_utf8` reached from exactly these two callers. **That attribution does not hold up, and I would rather say so than let it stand as this commit's justification.**

On a 1,500-function program carrying **3.87 MB of registered source text** — built specifically to maximise this cost — five interleaved A/B pairs of 40 runs each on a quiet machine:

| | median |
|---|---|
| lazy (this change) | **17.02 ms** |
| eager (kill switch) | 17.11 ms |

Faster in **5 of 5 pairs**, about **0.5%**. A 4,000-function program with short bodies showed nothing above noise.

Re-checking the call graph settled why: only **17 samples** in the recording mention `js_register_function_*` at all (13 source, 4 name) out of ~2,000 — and in every one of those stacks the **leaf is the register function itself**, not `from_utf8`, an allocator, or a memmove. So this whole path is ~0.85% of samples, the ~0.5% measured here is most of what is available from it, and `from_utf8`'s remaining share is reached from somewhere **not yet identified**. Recorded as unattributed rather than guessed at.

The original attribution came from finding *a* caller of `from_utf8` in the stacks without checking what fraction of its samples that caller accounted for. Finding a caller is not the same as finding the dominant one.

## What this deliberately does not do

A second and larger cost sits at these same call sites: each one `Arc`-allocates and **copies** every name and every source string. Removing that needs the registries to hold pointers, which would tighten these functions' published contract from *"the bytes outlive the call"* to *"the bytes outlive the process"*.

That is true for the `@.str.N` globals codegen emits from `__perry_init_strings_*`, but it is not something to assume for external callers of a `#[no_mangle] extern "C"` entry point. It is a contract decision rather than a perf tweak, so it is left out of this change.

## Behaviour

`fn.name` for function declarations, arrows, class methods, static methods, object shorthand and named function expressions; `fn.toString()` body text and shape; and `getOwnPropertyDescriptor(fn, "name")` — all byte-identical to node.

## Gates

`-D warnings` clean; hir 591/0; codegen 1842/0; runtime lib **2847/0**; census, address-classification, file-size and raw-handle-debt lints all pass with none raised. Integration: 7/7, 5/5, 7/7, 2/2, 3/3, 3/3.

## Kill switch

`PERRY_EAGER_FN_METADATA_UTF8=1` restores validation at registration time.

## Summary by CodeRabbit

* **New Features**
  * Added support for function names and source metadata containing invalid UTF-8 bytes.
  * Added an optional environment setting to enable validation when metadata is registered.

* **Bug Fixes**
  * Prevented invalid metadata from causing registration failures.
  * Invalid values are now safely omitted when retrieved.
  * Valid metadata can replace previously invalid entries when registered.

### DIFF
--- a/changelog.d/9187-lazy-fn-metadata-utf8.md
+++ b/changelog.d/9187-lazy-fn-metadata-utf8.md
@@ -1,0 +1,19 @@
+`js_register_function_name` and `js_register_function_source` validated UTF-8 at
+registration. Module init calls them once per function in the bundle — 72,713
+times for claude-code — and `core::str::from_utf8` was 2.17% of `cc --help`
+doing exactly that. Function source text is registered for every function a
+bundle *contains*, to serve `Function.prototype.toString()`, which most
+programs never call.
+
+Both registries now store raw bytes and decode on read. Nothing observable
+changes: a name that is not valid UTF-8 was previously dropped at registration
+and is now dropped at lookup, so both report "no name".
+`PERRY_EAGER_FN_METADATA_UTF8=1` restores eager validation for A/B measurement.
+
+One place needed the write side taught the same rule. An undecodable entry now
+*occupies* the map where registration used to leave the slot empty, so
+`register_function_name_if_absent`'s `or_insert_with` would have seen it as
+present and let it shadow a valid name forever — the one case where "dropped at
+lookup" would not have matched "dropped at registration". It now treats an
+entry that fails to decode as absent, which is what `decode_registered` already
+does on read.
--- a/crates/perry-runtime/src/builtins/formatting.rs
+++ b/crates/perry-runtime/src/builtins/formatting.rs
@@ -412,8 +412,18 @@ pub fn register_function_name_if_absent(func_ptr: usize, name: &str) {
         return;
     }
     if let Ok(mut map) = function_name_registry().lock() {
-        map.entry(func_ptr)
-            .or_insert_with(|| std::sync::Arc::from(name.as_bytes()));
+        // "Absent" now means "absent OR stored bytes that do not decode" —
+        // registration no longer rejects invalid UTF-8, so an undecodable
+        // entry occupies the slot where nothing used to, and a plain
+        // `or_insert_with` would let it shadow this valid name forever.
+        // `decode_registered` already treats such an entry as absent on read;
+        // this keeps the write side agreeing with it.
+        let usable = map
+            .get(&func_ptr)
+            .is_some_and(|bytes| std::str::from_utf8(bytes).is_ok());
+        if !usable {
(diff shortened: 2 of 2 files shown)


proggeramlug:

The move is right and the implementation is the safe version of it — raw bytes stored, `std::str::from_utf8(...).ok()` on read, no `from_utf8_unchecked` anywhere. Invalid bytes become `None`, which is what the caller saw before.

**Where the equivalence claim didn't quite hold.** "A non-UTF-8 name was previously dropped at registration and is now dropped at lookup, so both report no name" is true for the read path and false for one write path. Registration used to `return` without inserting, leaving the slot *empty*; now it inserts the undecodable bytes, so the slot is *occupied*. `register_function_name_if_absent` asks only whether the entry is present, so a later valid name — from `util_promisify`, symbol property naming, or the `get {key}` accessor path — would find the slot taken and never be recorded. The function would report no name where before it reported one.

Practically unreachable, since function names come from parsed TypeScript identifiers and are UTF-8 by construction, so I'd not have blocked on it. But the fix is three lines and it makes your stated invariant actually true, so I pushed it: `_if_absent` now treats an entry that fails to decode as absent, which is what `decode_registered` already does on read. Moving validation from write-time to read-time means every place that treats "present" as "usable" has to ask whether it decodes — that was the one remaining place.

**Validation:** differential probe against node 26.5.1 covering `.name` for function declarations, arrows, class methods, statics, getters and object literal methods; `.length`; `Function.prototype.toString()` including body text; bound functions; anonymous functions; native `Math.max`; and — the ones that actually exercise the UTF-8 path — a `function ほげ` assigned to a `変数` binding and an emoji-keyed method. Byte-identical. `perry-runtime` 2853 passed / 0 failed at `RUST_TEST_THREADS=1`; all 60 lint gates green.

Added the `changelog.d/` fragment, which the branch was missing. `PERRY_EAGER_FN_METADATA_UTF8` is a runtime knob, so it's correctly outside `codegen_env_vars_are_build_cache_inputs`.

Issue #9152, opened 2026-08-30
raw_handle_debt ratchet counts across_*(|| ()) no-ops as conversions (~17 sites on main)


proggeramlug:
`scripts/raw_handle_debt.py` counts a conversion as done when the code calls `RuntimeHandle::across_*` — but `across_*(|| ())` is a **no-op wrapper that is byte-identical to the bare read it replaced**. There are currently **17 such sites on main**, across a dozen files, so the ratchet's headline count overstates how much custody has actually been established.

### The mechanism

```rust
pub fn across_mut<T, R>(&self, f: impl FnOnce() -> R) -> (R, *mut T) {
    let result = f();
    (result, self.get_raw_mut_ptr::<T>())
}
```

With `f = || ()` that is exactly `self.get_raw_mut_ptr::<T>()`. The point of `across_*` is that **the allocating call goes inside the closure**, so the pointer you get back is refreshed *after* it. An empty closure refreshes across nothing.

The helper's own doc comment already says as much:

> It is not a proof. It cannot stop you reading the pointer *before* the call and holding that copy yourself … What it does is make the correct shape shorter than the incorrect one and give the ratchet in `scripts/raw_handle_debt.py` something to count down.

### Where they are

```
crates/perry-runtime/src/array/iter_object.rs:1
crates/perry-runtime/src/builtins/formatting/boxed_primitives.rs:1
crates/perry-runtime/src/collection_iter_object.rs:2
crates/perry-runtime/src/dyn_eval/tests.rs:2
crates/perry-runtime/src/intl/list_relative_plural.rs:1
crates/perry-runtime/src/map.rs:1
crates/perry-runtime/src/node_vm.rs:1
crates/perry-runtime/src/object/native_call_method.rs:3
crates/perry-runtime/src/object/object_ops/define_property.rs:1
crates/perry-runtime/src/object/reflect_support.rs:2
crates/perry-runtime/src/string/iter_object.rs:1
crates/perry-runtime/src/typedarray/construct.rs:1
```

Examples: `array/iter_object.rs:86`, `collection_iter_object.rs:65`, `intl/list_relative_plural.rs:820`, `object/field_set_by_name/fast_paths.rs:230`.

### Why it matters

[...]

PR #8529, opened 2026-08-21
fix(compile): AOT-promote trusted reachable JavaScript

proggeramlug:
## Summary

AOT-promote statically reachable JavaScript and CommonJS files when their owning package is permitted by the host trust policy, without requiring every package to also appear in `perry.compilePackages`.

## Changes

- Decouple file-level AOT promotion from the eager whole-package `compilePackages` routing list.
- Continue routing packages excluded by an explicit allow policy to the V8-free refusal boundary.
- Add regressions covering wildcard trust, Node export-condition selection, importer-relative CommonJS traversal, and explicit denial.

## Related issue

Fixes #8518

## Test plan

- [ ] `cargo build --release` clean
- [ ] `cargo test --workspace --exclude perry-ui-ios --exclude perry-ui-tvos --exclude perry-ui-watchos --exclude perry-ui-gtk4 --exclude perry-ui-android --exclude perry-ui-windows` passes
- [x] `cargo fmt --check`
- [x] `cargo test -p perry --bin perry` (1,010 passed)
- [x] Added `#[test]` regressions in the affected crate
- [ ] (if CLI / stdlib / runtime API changed) Updated `docs/src/` (not applicable)
- [ ] (if touching a platform UI backend) Built `-p perry-ui-<backend>` locally (not applicable)

Clippy with `-D warnings` is currently blocked by pre-existing warnings in unchanged files; neither clippy run reported a warning in the changed files.

## Screenshots / output

Not applicable; this changes compiler routing behavior and is covered by regression tests.

## Checklist

- [x] I have NOT bumped the workspace version or edited CLAUDE.md / CHANGELOG.md (maintainer handles these at merge)
- [x] My commits follow the loose `feat:` / `fix:` / `docs:` / `chore:` prefix convention used in the log
- [x] I've read [CONTRIBUTING.md](../CONTRIBUTING.md) and agree to the [Code of Conduct](../CODE_OF_CONDUCT.md)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ahead-of-time compilation for trusted, statically reachable JavaScript and CommonJS dependencies.
  * Removed the need for per-package routing entries when wildcard compilation permission is enabled.
  * Untrusted packages continue to use runtime JavaScript modules.

* **Documentation**
  * Added release notes clarifying trusted dependency handling and compilation routing.

PR #9190, opened 2026-08-30
perf(runtime): the receiver own-key probe stops scanning the keys array element by element

proggeramlug:
A DWARF call-graph recording of the claude-code CLI running `--help` showed the profile's **largest symbol, `js_array_get_f64` at 4.2%, has zero callers in generated code**. It is reached from the property-STORE slow path:

```
js_array_get_f64
 └─ across_mut::<ArrayHeader, JSValue, obj_value_has_own_key::{closure#3}>   ← the profile's 2.44% frame, verbatim
     └─ obj_value_has_own_key
         └─ proxy::own_set_descriptor                                        (3.16%)
             └─ ordinary_set_with_receiver → js_put_value_set → js_put_value_set_dyn_ic_miss
```

`obj_value_has_own_key`'s ordinary-object arm answered "does this receiver already own this key?" with a loop doing, **per key**, one `js_array_get` (the full JS-facing accessor: `clean_arr_ptr`, Map/Set/typed-array/buffer registry probes, descriptor gate, hole translation) **plus** a `RuntimeHandle::across_mut` re-rooting round trip.

**Why the index wasn't used: nothing structural — it was missed.** `keys_find_slot_by_key_ptr` answers exactly this question (O(1) shape index at/above the 32-key threshold, raw dense-slot compare below), and its own doc comment records replacing these walks elsewhere ("measured 90.8 million `js_array_get_f64` calls for 1.5 M property operations"). Thirteen other [[Get]]/[[Set]]/delete sites route through it — including three in `proxy/put_value.rs`, the same subsystem. This one call site kept the old loop.

**Cost is quadratic in the receiver's key count.** Counted with a temporary runtime counter, not timed — the esbuild CJS-namespace shape (one `Object.defineProperty`, then N stores):

| N stores | element reads before | after | process-total `js_array_get_f64` |
|---|---|---|---|
| 100 | 10,800 | 0 | 12,435 → 1,635 |
| 400 | **163,200** | **0** | **165,135 → 1,935** |

cc's bundle contains 1,526 `Object.defineProperty` occurrences, so this is the module-init path. Nine store shapes were probed to find which reach the ordinary arm; three do (defineProperty-then-store, `Reflect.set` with receiver ≠ target, and a descriptor on `Object.prototype`), and class instances / arrays / closures / `Object.create` do not.

**The change is one call site** — the loop becomes `keys_find_slot_by_key_ptr(...).is_some()`. It allocates nothing, so the per-iteration re-rooting the old loop required disappears with it. No proxy semantics touched: same question, different search.

**Symbol-level proof** (`objdump -d -r`): the `across_mut::<ArrayHeader, JSValue, obj_value_has_own_key::{closure}>` monomorphization goes 1 → 0, its `js_array_get_f64` relocation 1 → 0, `keys_find_slot_by_key_ptr` appears, and the only surviving array call is one `js_array_length`. **`.text` −320 bytes.**

**Validation:** `test_gap_9180_receiver_set_own_key_scan.ts` byte-identical to node — both index tiers (8 and 40 keys, crossing the threshold), and the two paths where the index declines or reports absence (a delete shrinking it to `Unindexed`, and the `Absent` completeness verdict), plus `Reflect.set` receiver ≠ target, proxy receiver with the trap log asserted, own accessor on receiver, non-writable own data, `defineProperty` interaction, prototype shadowing, non-extensible receiver, index-vs-name keys. A unit test pins it against a test-only accessor counter and **fails on the parent commit** (17 accessor calls for 3 probes of an 8-key object) — the regression pin proves itself. perry-runtime 2848/0, perry-codegen 1357/0, native_proof_regressions 285/0, and a 40-fixture node differential over object/proxy/descriptor/prototype fixtures with the one mismatch reproduced unchanged on the parent.

## Summary by CodeRabbit

* **Performance**
  * Improved receiver property-key lookups during assignments by avoiding repeated element-by-element scans.
  * Reduced overhead for both small and large key collections.

* **Bug Fixes**
  * Improved handling of deleted and re-added keys, non-writable properties, accessors, proxies, prototype shadowing, and non-extensible receivers.

* **Tests**
  * Added comprehensive regression coverage for receiver-based property assignment and key lookup behavior.


proggeramlug:
Rebased onto `84185b5656` and fixed the two `lint` failures. No change to the runtime diff itself.

- **changelog fragment** — added `changelog.d/9190-set-receiver-own-key-index.md`.
- **file-size cap** — the new pin took `crates/perry-runtime/src/object/tests.rs` from 1,991 to 2,088 lines. Moved it to its own `object/set_receiver_tests.rs` (the same split `tombstone_tests.rs` already uses), which puts `tests.rs` back at 1,991 and the new module at 102. `./scripts/check_file_size.sh` is clean, and `cargo test -p perry-runtime --lib object::` is 256 passed / 0 failed locally.

Two other reds on the previous run, neither of which I believe is this diff — recording what I checked rather than asserting it:

**`e2e-scoped`** is a **main breakage, not mine**: `ci_e2e_scope --self-test` fails because `crates/perry-codegen/tests/string_array_length_9160.rs` landed with #9160 without a `_CODEGEN_SUITES` entry, and that map is complete-by-construction. It is red on every core PR right now. Fixed separately in #9193.

**`cargo-test`** aborted with a **stack overflow**, not an assertion — `commands::compile::collect_modules::tests::statically_reachable_trusted_js_package_is_aot_compiled_without_route_entry`, a test from #8529 in the perry *compiler* crate, while this diff is runtime-only. Reads like a debug-build recursion-depth flake.

**`gc-stress`** reported `PASS=449 UNVER=124 XFAIL=0 FAIL=1`: `test_gap_repsel_pshape_tower_delete`, output-mismatch, in exactly one of the seven PR arms (`force_verify`) — and that cell reports `evacuated=0`, so the arm's distinguishing behaviour did not actually fire. That test is about `delete` compacting inline slots behind a cloned keys array, which is adjacent enough to this diff that I am not willing to call it unrelated on a reading of the code. I am reproducing it locally on this branch under that exact arm; if it reproduces I will root-cause it before asking anyone to merge this.

proggeramlug:
**Validated on claude-code itself: this chain is 10.5% of `cc --help`.**

A fresh symbolized profile of current main (`84185b5656`), 12 reps of `claude --help`, `PERRY_DEBUG_SYMBOLS=1`:

```
 4.54%  js_array_get_f64
 3.15%  perry_runtime::string::compare::js_string_key_matches
 1.83%  RuntimeHandle::across_mut::<ArrayHeader, JSValue,
              object::reflect_support::obj_value_has_own_key::{closure#3}>
 0.96%  perry_runtime::object::reflect_support::obj_value_has_own_key
```

`js_array_get_f64` is the **largest single symbol in the entire profile**, and the three beneath it are the rest of the same loop: the per-iteration handle round-trip (the closure symbol names `obj_value_has_own_key` outright), the per-element compare, and the function itself. This PR removes all four — the element read, the re-rooting, and the compare are replaced by one `keys_find_slot_by_key_ptr` probe.

Context, same box, five interleaved reps:

| | perry | node | ratio |
|---|---|---|---|
| `--help` instructions | 8.88 B | 4.05 B | 2.19× |
| `--version` instructions | 1.21 B | 2.74 B | perry 2.26× faster |

A note on how this was validated, because the obvious benchmark was wrong. Session 88 measured plain-object stores at 17–29× node and offered the curve as this PR's validation. Counting rather than timing killed it: `obj_value_has_own_key` and `js_array_get_f64` appear **nowhere** on the plain-store path — a plain literal receiver never leaves the direct-store lane, and one `Object.defineProperty` is what puts it on this one. That is why the pin in this PR counts `js_array_get_f64` entries (`test_element_accessor_calls`) instead of timing anything, and why the number quoted in the commit message is 163,200 element reads → 0 rather than a duration. The cc profile above is the timing evidence, from the workload that actually matters.

proggeramlug:
Merged.

The counter is the part I want to single out. Asserting `163,200 → 0` calls into `js_array_get_f64` pins *the property* — that the walk no longer reaches for the element accessor — where a timing assertion would have pinned the machine it was measured on and gone flaky the first time CI was busy. That's the right instrument for this change, and it's the difference between a test that can still fail for the right reason in a year and one that gets quarantined.

The reasoning behind the change holds up too: the element accessor runs forward resolution, the Map/Set/typed-array/buffer registry probes, the descriptor gate and hole translation, and the own-key walk needs none of it — it wants a raw slot read and knows the storage is dense.

**Validation:** `test_gap_9180_receiver_set_own_key_scan.ts` matches node 26.5.1 byte-for-byte across all 53 assertions, including the ones most likely to catch a too-aggressive direct read — `Reflect.set` with a distinct receiver, non-writable and accessor properties on the target, a non-extensible receiver, the Proxy trap log ordering, index-vs-string key aliasing (`"2"` vs `"02"`), prototype shadowing, and delete-then-readd. `perry-runtime` 2853 passed / 0 failed at `RUST_TEST_THREADS=1`; all 60 lint gates green.

My own differential probe over the same area (300+ named keys, numeric/string key aliasing, negative and fractional keys, non-enumerable and accessor descriptors, symbol keys, `getOwnPropertyNames`, `for…in` order under prototype shadowing, `Object.freeze`, and a 400-key index-like receiver) matches node too, with one exception that is **not yours**: `arr.foo = "bar"` on an array is silently dropped — the read returns `undefined` and the key never appears in `Object.keys`. I confirmed it pre-existing by reverting your four runtime files to `main` and rebuilding the runtime and stdlib static wrappers: byte-identical output, both differing from node in exactly those two lines. Filed as **#9201**; it's a silent failure on a pattern real packages use (arrays with named metadata fields), so it seemed worth its own issue rather than a footnote.

The two `across_mut::<ArrayHeader, _>(|| ())` / `across_const::<StringHeader, _>(|| ())` sites are the vacuous form of that helper — with a no-op closure nothing can move, so the re-resolution is a verbose deref rather than a safety measure. Not blocking and not a mark against this PR: the pattern is endemic on `main` (~16 existing sites), which is why I filed **#9152** about it separately rather than raising it per-PR.

PR #9191, opened 2026-08-30
perf(gc): build the stack-map index on first collection, not at startup (−26% startup, −42% RSS)

proggeramlug:
Step 2 of the sequence #9182 opened. **#9182 must be in main before this** — it is, and this is what it was for.

## What was wrong

`js_gc_init` decoded the entire gc-map section at every process start. For claude-code: **17.4 MB, 72,713 functions, 2,151,926 records**, producing a **117 MB index**. A run that never collects never reads a byte of it.

## Numbers

One binary switched with `PERRY_LAZY_STACK_MAPS`. A 1,500-function program carrying 3.87 MB of source; medians of interleaved pairs of 40 runs on a quiet machine:

| | lazy | eager |
|---|---|---|
| startup | **12.77 ms** | 17.25 ms — **−26%** |
| peak RSS | **35.6 MB** | 61.4 MB — **−42%** |

A hello-world is unchanged (3.47 vs 3.51 ms), which is the shape to expect: its section is small, so there was little to defer.

## Why this was not safe before, and is now

Scanning against an unbuilt index is **indistinguishable** from an image with no native roots — both are empty — so the collector finds no roots on the frame and frees live objects with no diagnostic at all. Correctness therefore rests on every path that can reach the root scan first passing through the build, and **that enumeration could not be established**: an initial claim of "two chokepoints" was withdrawn as unverified, and tracing found at least three fan-in points plus untraced allocation-triggered entries.

So the fail-closed assert landed **first** (#9182), against the eager build, where it could not fire.

**That decision paid immediately, and this is the part worth reading.** The assert caught three entry points I had not enumerated — including the first safepoint-triggered collection — and it caught a mistake of mine: I had put the build in `gc_finish_arena_trigger_collection` and `gc_finish_malloc_trigger_collection`, which take a `GcCollectOutcome` and therefore **run after the collection they name**. A placement that is useless, invisible to every timing test, and would have been a silent heap corruption discovered weeks later. Instead it was a loud abort in minutes, four separate times.

That is precisely the failure the two-PR sequence was designed to catch, catching precisely that failure, on the first run.

## Where the build goes

The three functions every collection funnels through — `gc_collect_minor_with_trigger`, `gc_collect_forced_evacuating_minor`, `gc_collect_full_mark_sweep_with_trigger` — where **allocation is still legal**. That is the constraint that stops the build being deferred any further: the root scan itself must stay allocation-free once the collector owns the heap, which is why it was hoisted to init in the first place.

`initialize` re-arms rather than builds, so a newly loaded image's section is decoded by the next collection.

## Scope of the win

Worth stating so the 26% is not quoted for the wrong workload: a **non-collecting** run (`--version`, shell completions, health checks) avoids the decode entirely. A run that **does** collect — `cc --help` spends 16.4% of wall time in minor collections — only has it **deferred**, not removed. Avoiding the decode for functions never on a stack needs the per-function `stream_offset` in a v5 map format, which is a separate change.

## Validation

**39 runs across 13 programs** under `PERRY_GC_PROTECT_FROMSPACE` with seeded aggressive collection schedules — up to **193,052 moved objects** — with the assert firing **zero** times and output byte-identical to node.

Two programs differ from node in both arms: the `Buffer.isBuffer` defect (#9173) and a class-prototype accessor-descriptor gap. Both reproduce identically with `PERRY_LAZY_STACK_MAPS=0`, so they are pre-existing and not from this change.

## Gates

`-D warnings` clean; hir 591/0; codegen 1842/0; runtime lib 2847/0; census, address-classification, file-size and raw-handle-debt lints all pass with none raised. Integration: 7/7, 5/5, 7/7, 2/2, 3/3, 3/3.

## Kill switch

`PERRY_LAZY_STACK_MAPS=0` restores the eager decode at `js_gc_init`.

## Summary by CodeRabbit

* **Performance**
  * Deferred garbage-collection metadata preparation until the first collection, reducing startup work and memory usage.
  * Added an option to disable deferred preparation when eager initialization is preferred.

* **Bug Fixes**
  * Improved garbage collection reliability by ensuring required metadata is prepared before collection begins.
  * Prevented allocation during critical collection phases, reducing the risk of collection-time failures.

### DIFF
--- a/changelog.d/9191-lazy-stack-map-index.md
+++ b/changelog.d/9191-lazy-stack-map-index.md
@@ -1,0 +1,29 @@
+`js_gc_init` decoded the whole gc-map section at every process start. For
+claude-code that is 17.4 MB, 72,713 functions and 2,151,926 records, producing a
+117 MB index — and a run that never collects never reads a byte of it.
+
+The build moves to the three functions every collection funnels through
+(`gc_collect_minor_with_trigger`, `gc_collect_forced_evacuating_minor`,
+`gc_collect_full_mark_sweep_with_trigger`), which is as late as it can go:
+allocation is still legal there, and the root scan itself must stay
+allocation-free once the collector owns the heap. `initialize` re-arms rather
+than builds, so a newly loaded image's section is decoded by the next
+collection. `PERRY_LAZY_STACK_MAPS` switches one binary between the two.
+
+On a 1,500-function program carrying 3.87 MB of source, medians of interleaved
+pairs of 40 runs: startup 12.77 ms vs 17.25 ms (−26%), peak RSS 35.6 MB vs
+61.4 MB (−42%). A hello-world is unchanged, which is the expected shape — its
+section is small, so there was little to defer. Scope worth stating: a
+non-collecting run avoids the decode entirely, while a run that does collect
+only has it deferred.
+
+This is step 2 of the sequence #9182 opened, and the reason that PR landed
+first. Scanning against an unbuilt index is indistinguishable from an image with
+no native roots, so correctness rests on every path into the root scan passing
+through the build — an enumeration that could not be established by reading the
+code. The fail-closed assert caught three unenumerated entry points and one
+misplacement: the build had been put in `gc_finish_arena_trigger_collection` and
+`gc_finish_malloc_trigger_collection`, which take a `GcCollectOutcome` and
+therefore run *after* the collection they name — useless, invisible to every
+timing test, and a silent heap corruption weeks later. It aborted loudly in
+minutes instead, four separate times.
--- a/crates/perry-runtime/src/gc/mod.rs
+++ b/crates/perry-runtime/src/gc/mod.rs
@@ -221,6 +221,13 @@ pub fn gc_collect_minor() -> u64 {
 }
 
 pub(super) fn gc_collect_minor_with_trigger(trigger: GcTriggerSnapshot) -> GcCollectOutcome {
+    // Build the stack-map index if it is still owed. This is the chokepoint:
+    // every collection funnels through one of these three entries, and here
+    // allocation is still legal — the root scan itself must stay
+    // allocation-free once the collector owns the heap, which is why the build
+    // cannot be deferred any further than this.
+    roots::ensure_stack_maps_built();
+
     gc_collect_minor_with_trigger_inner(trigger, FullEscalation::Allowed)
 }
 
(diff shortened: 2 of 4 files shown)


proggeramlug:
Merged.

**The two-PR sequence worked, and that's the headline.** #9182's assert caught three unenumerated entry points and one genuinely nasty misplacement — the build sitting in `gc_finish_arena_trigger_collection` / `gc_finish_malloc_trigger_collection`, which take a `GcCollectOutcome` and therefore run *after* the collection they name. That placement is invisible to every timing test and every correctness test that doesn't happen to collect at the wrong moment; it would have been a silent heap corruption found weeks later by someone else. Getting it as a loud abort in minutes, four times, is exactly what landing the assert first was for. Worth recording as the precedent: when an invariant can't be established by reading the code, land the detector against the safe configuration first and let it enumerate the paths for you.

**Verified the subject is live rather than trusting the numbers.** A number moving is not evidence the deferral is what moved it, so I A/B'd one binary on `PERRY_LAZY_STACK_MAPS` over 5 interleaved pairs on a 1,500-function fixture: peak RSS 6.65 MB lazy vs 7.85 MB eager, reproducing to the byte on every run (6,651,904 vs 7,847,936). The arms genuinely differ, so the lazy path is taken.

That's −15%, not your −42%, and the gap is my fixture rather than a disagreement: mine carries far less source text, so there is a much smaller section to defer. Same direction, smaller subject. Startup didn't resolve at all on it (both arms 0.00 s), so I can't independently corroborate the −26% — your interleaved-medians methodology on a 3.87 MB program is the better measurement and I'd quote yours, not mine.

**Correctness:** 12 arms across 4 GC-heavy programs (linked-list churn with a rolling live set, symbol-keyed property lookup, throw-in-loop under allocation churn, typed-array and class-hierarchy work) under default, `PERRY_GC_PROTECT_FROMSPACE=1` with `PERRY_GC_SCHEDULE_SEED=5 PERRY_GC_SCHEDULE_RATE=1`, and `PERRY_GC_FORCE_EVACUATE=1`. All exit 0, the assert fired zero times, output matching node 26.5.1. `perry-runtime` 2853 passed / 0 failed at `RUST_TEST_THREADS=1`; `check_gc_env_knobs.py` clean, so `PERRY_LAZY_STACK_MAPS` is properly declared.

One thing to keep in view: CLAUDE.md's GC knob kill-policy says every GC knob needs a required CI arm exercising its OFF state, or it gets deleted after one release of soak. `PERRY_LAZY_STACK_MAPS` is currently an A/B measurement knob with no CI arm on either side. Given the eager path is now the *unexercised* one, that's the branch most likely to rot — worth either an arm or a deletion decision before the soak window closes, rather than leaving eager as a configuration nobody runs.

Unrelated: the `warnings` gate is currently red from 12 pre-existing `unnecessary unsafe block` warnings in `perry-ui-macos` (untouched here, deps unchanged, files last modified 2026-08-20). Handling separately.

PR #9195, opened 2026-08-30
perf(codegen): an indexed store skips the addref for a provably non-string value

proggeramlug:
A small, principled one: an indexed array store no longer emits a call that provably cannot do anything.

## The gap

`emit_jsvalue_slot_store_on_block` and its scalar-aware twin pass `layout_note_needed` for **both** the layout note and the string-addref demote:

```rust
emit_jsvalue_slot_store_on_block_inner(
    …, layout_note_needed, layout_note_needed, …)
//     ^ string_addref_needed
```

So every store needing a layout note also emitted `js_string_addref_if_heap_string` — including `sieve[j] = false`, whose value is a boolean and can never be a heap string. The flagged emitter that separates the two already exists and `array_push` already uses it; this adds the scalar-aware twin and threads `store_needs_string_addref` — the same predicate the push path trusts — from the one caller that has the value expression in hand.

`benchmarks/suite/11_prime_sieve.ts`: 8 emitted addref calls → 1.

## Measurements, and what they honestly say

Idle Mac mini, both binaries built in one run, interleaved, min of five, self-timed:

| | base | this change | node |
|---|---:|---:|---:|
| boolean-store loop | 215 ms | **207 ms** | 12 ms |
| `11_prime_sieve` | 28 ms | **26 ms** | 6 ms |

About 4% each, reproducible to the millisecond across runs. **This is not where either benchmark's gap lives**, and the same table says why: Node does that boolean-store loop in 12 ms against perry's 207. The per-store typed-feedback guard **call** is the cost that matters there — a separate, larger piece of work. This change removes dead work sitting beside it.

## Correctness

The differential targets the exact hazard the addref exists to prevent — a refcount-1 string aliased into a slot and then mutated through the source local:

- `s = "abc"; a[0] = s; s += "def"` — `a[0]` must stay `"abc"`;
- the same through a loop-written array (`b[i] = t; t += "!"`);
- boolean / number / `null` / `undefined` stores into an array (the values that now skip the addref);
- the `prime_sieve` boolean-array shape itself;
- a string slot overwritten by a boolean and back.

Byte-identical to Node. 31 `perry-codegen` suites pass; `cargo fmt --check` clean.

## Summary by CodeRabbit

  * Improved generated code performance for indexed array stores involving booleans and other non-string values.
  * Benchmark results show approximately 4% faster execution in representative array-store and prime-sieve workloads.

* **Bug Fixes**
  * Prevented unnecessary string reference-counting work when storing values that cannot be heap strings.
  * Preserved correct behavior for string overwrites and stores involving booleans, numbers, `null`, and `undefined`.

* **Validation**
  * Verified generated output against Node.js behavior.
  * All 31 code-generation test suites pass.

### DIFF
--- a/changelog.d/index-store-skips-a-dead-string-addref.md
+++ b/changelog.d/index-store-skips-a-dead-string-addref.md
@@ -1,0 +1,27 @@
+An indexed array store no longer emits `js_string_addref_if_heap_string` for a
+value that provably cannot be a heap string.
+
+The plain slot-store emitters pass `layout_note_needed` for *both* the layout
+note and the string-addref demote, so any store that needs a layout note also
+paid the addref call — including `sieve[j] = false`, where the value is a
+boolean. The flagged emitter that separates the two already existed and
+`array_push` already used it; this adds the scalar-aware twin and threads
+`store_needs_string_addref` (the same predicate the push path trusts) from the
+one caller that has the value expression in hand.
+
+`benchmarks/suite/11_prime_sieve.ts` goes from 8 emitted addref calls to 1.
+
+Measured on an idle Mac mini, both binaries built in one run, interleaved, min of
+five, self-timed: a boolean-store loop 215 → 207 ms, and `11_prime_sieve` 28 → 26
+ms. Roughly 4% each — small, and worth saying plainly that it is not where either
+benchmark's gap lives: the same measurement puts Node at 12 ms on that
+boolean-store loop against perry's 207, so the per-store typed-feedback guard
+**call** is the cost that matters there. This change removes provably dead work
+beside it.
+
+Verified against Node on a differential aimed at the exact hazard the addref
+exists to prevent — a refcount-1 string stored into a slot and then mutated
+through the source local (directly, and through a loop-written array), booleans /
+numbers / `null` / `undefined` stored into an array, the sieve shape itself, and a
+string slot overwritten by a boolean and back. Byte-identical. 31 `perry-codegen`
+suites pass.
--- a/crates/perry-codegen/src/expr/index.rs
+++ b/crates/perry-codegen/src/expr/index.rs
@@ -4,8 +4,9 @@
 use anyhow::{anyhow, Result};
 
 use super::{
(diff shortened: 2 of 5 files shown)


proggeramlug:
**Gate on the Linux box, commit `75835c140`: clean.** Formatting PASS, CI-plan self-test PASS, gap-snapshot PASS, parity-allowlist PASS, no ratchet ceilings raised; the 3 FAIL steps are the environmental `${{ github.* }}` ones (a changelog fragment is present). Runtime suite `2828 passed; 3 failed` — the three known parallel-flaky tests, now filed as #9196 with the evidence that they reproduce on unmodified `origin/main`. This diff is codegen-only.

proggeramlug:
Merged.

The risk in "skip the addref for a provably non-string value" is entirely in the word *provably* — a value that is actually a string, stored through a path that skipped its refcount, is a use-after-free that surfaces cycles later somewhere unrelated. So I probed exactly that: 100 strings written through the indexed store, then 40,000 object allocations with a rolling live set to force collections, then every one read back and compared.

100/100 survived under default, `PERRY_GC_FORCE_EVACUATE=1`, and `PERRY_GC_SCHEDULE_SEED=11 PERRY_GC_SCHEDULE_RATE=1`. Also covered the mixed and non-string cases through the same path — a 200-element all-number array, a 50-element all-string array, an alternating number/string array, and one holding `true`/`null`/`undefined`/object/array/BigInt/Symbol/float — all matching node 26.5.1 byte-for-byte.

**Validation:** `perry-codegen` 31 suites / 0 failures; `perry-runtime` 2853 passed / 0 failed at `RUST_TEST_THREADS=1`.

One thing that is **not** yours: `cargo check --workspace --all-targets` currently emits 12 `unnecessary unsafe block` warnings in `perry-ui-macos` (`file_dialog.rs`, `widgets/{alert,combobox,webview}.rs`, others), which fails the `warnings` lint gate under `-D warnings`. That crate is untouched by this PR and its dependencies are too, so its check output is identical on `main`; those files last changed on 2026-08-20 (#8453). It surfaces on any PR that invalidates that crate's check cache — which anything touching `perry-runtime` does — so it is not attributable here. Handling it separately.

PR #9203, opened 2026-08-30
perf(runtime): the class-id-keyed registries move off SipHash — cc --help −0.69%, 25/25 rounds

proggeramlug:
The class-registry tables keyed by `class_id` (and by keys-array address) were hashing their integer keys with std's SipHash. They now use the crate's existing fast hashers. Converted: `ClassImageTables`' 17 class-id-keyed tables — including `parents`, whose `get_parent_class_id` the code itself documents as "the single hottest class-registry read" — plus `CLASS_KEYS_BY_ID`, `shape_cache_overflow`, and `typed_shape`'s two maps.

`ids_by_layout` got **`FastKeyHashMap`, not `PtrHashMap`**: its key is a composite (two `u32`s plus two `Vec<u64>`s), and `PtrHasher`'s overwriting `write_*` would collapse that to its last field.

## Numbers

Alternating A/B, 25 paired rounds, on an idle Mac mini (the dev box sat at load 14–45 all session and was unusable for timing):

| | median | min |
|---|---|---|
| base | 873.0 ms | 871 ms |
| this | 867.0 ms | 862 ms |

**−0.69%**, faster in **25 of 25** rounds, mean delta −6.2 ms (stdev 2.2) ≈ 14 standard errors. In the profile, `get_parent_class_id`, `js_gc_typed_shape_id_for_keys`, `shape_cache_insert` and the whole `js_register_class_*` family disappear from the SipHash caller list entirely.

`.text` **+2 KB (+0.016%)**. `cargo test -p perry-runtime --lib -- --test-threads=1`: 2844 passed, 0 failed, 4 ignored. `cargo fmt --all --check` clean. `cc --help` output byte-identical before and after.

## Two corrections to the premise this started from

The brief for this work claimed ~4.9% of `cc --help` in SipHash with a single map worth 1.60%. Re-measured on current main, **total SipHash is 3.08%** and there is no such map. Both errors have the same cause and it is worth recording:

- **Every map originally named was already fast-hashed** — `ARRAY_NAMED_PROPS` (PtrHashMap since #6386), the shape descriptor table (PtrHasher, #8125), `descriptor_state` (FastKeyHasher, #9147), `CLOSURE_BOX_CELLS` (PtrHashMap). Their hashbrown entries are residual **table-probe** cost. A hashbrown symbol says nothing about which hasher is underneath it.
- **The largest single SipHash consumer is not ours at all**: `regex_automata::dfa::determinize::Runner::maybe_add_state` at 0.489%.

A third measurement artefact, in the other direction: a `u32` key's SipHash is small enough that **LLVM inlines it**, so it never appears under a `RandomState` frame. Integer-keyed maps are systematically *under*-represented in a sampled profile — which is exactly the class of map this PR converts.

## Left deliberately

Everything keyed by a JS-supplied name (the inner `HashMap<String, _>` of the static method/accessor tables, `method_bind_lengths`, `ClosureProps::values`, `js_for_in_keys_value`'s dedup set) — a fast hasher on attacker-influenced keys is a different decision and not this PR's. Also `REGEX_POINTERS`/`REGEX_SOURCE_TABLE` and the `node_submodules` singletons: sound, but measured at ≤0.09%, i.e. diff risk for no gain.

## Iteration-order audit

No converted map is iterated anywhere except GC root scans (`.values_mut()`, commutative) and `class_side_table_root_snapshot`. `CLASS_KEYS_BY_ID`, `ids_by_layout`, `layouts_by_id` and every non-GC `shape_cache_overflow` path are never iterated at all.

## A pre-existing bug found in passing, NOT fixed here

`CLASS_DYNAMIC_PROPS`' inner map is **order-observable**: `sort_property_names_ecma` sorts only integer-like keys, so string keys keep raw map order, and that order feeds `Object.keys(C)`. Because `RandomState` seeds per process, static string-field order there is likely already **non-deterministic across runs** — a spec divergence that predates this PR and is independent of it. Flagging rather than folding it in.

## Microbenchmarks that failed, reported as failures

Two microbenchmarks measured nothing and are **not** offered as evidence. The class benchmark was vacuous: `get_parent_class_id` short-circuits through a 65,536-entry dense array, so with fewer than 65,536 class ids the converted map is *never probed* — only an image the size of the claude-code bundle reaches it. The shape benchmarks were dead-code-eliminated (1–2 ms). The end-to-end A/B above is the only valid measurement here.

## Summary by CodeRabbit

* **Performance**
  * Improved the speed of class, shape, and metadata registry lookups by using optimized internal storage.
  * Runtime benchmarks show a median performance improvement of approximately 0.69%.

* **Bug Fixes**
  * Preserved existing lookup behavior and iteration order while improving registry efficiency.

* **Documentation**
  * Added release documentation describing the optimized registries, performance results, and compatibility considerations.

### DIFF
--- a/changelog.d/9147-followup-class-registry-fast-hash.md
+++ b/changelog.d/9147-followup-class-registry-fast-hash.md
@@ -1,0 +1,59 @@
+### Changed
+
+- **The class-id-keyed registries moved off SipHash.** #9147 converted the
+  pointer-keyed runtime registries to `fast_hash`; the class registries were
+  the sampled remainder. A symbolized 12-run profile of `claude-code --help`
+  on the post-#9147 tree put SipHash (`RandomState` on the stack) at 3.08% of
+  on-CPU samples, and named these callers: `get_parent_class_id` (0.198% of
+  total samples), `remember_class_keys_array` (0.175%),
+  `js_gc_typed_shape_id_for_keys` (0.115%), plus the `js_register_class_*`
+  family (~0.18% combined).
+
+  Converted to `PtrHashMap` / `PtrHashSet`:
+
+  * every class-id-keyed table in `ClassImageTables` — `vtables`,
+    `static_methods`/`static_accessors` (OUTER map only), `constructors`,
+    `constructor_flags`, `registered_class_ids`, `parents`,
+    `fetch_parent_kind`, `generic_origin`, `extends_error`, `has_instance`,
+    `to_string_tag`, `extends_data_view`, `extends_typed_array`, `names`,
+    `lengths`, `anon_shape_class_ids`;
+  * `CLASS_KEYS_BY_ID` (`object/alloc.rs`), probed on every object
+    construction — `js_build_class_keys_array` calls
+    `remember_class_keys_array` even on its shape-cache hit path;
+  * `ObjectHotTables::shape_cache_overflow`, probed whenever a shape id misses
+    the 256-entry direct-mapped inline cache (ids step by `10007 mod 256 == 23`,
+    so any image with >256 live shapes collides constantly). Every sibling
+    field of that struct was already a fast table or an `UnsafeCell` array;
+    this one was the exception.
+
+  `gc::layout::typed_shape`'s `ids_by_layout` takes `FastKeyHashMap`, not
+  `PtrHashMap`: `RegisteredTypedShapeKey` is a composite of two `u32`s and two
+  `Vec<u64>` mask lists, and `PtrHasher`'s overwriting `write_*` would collapse
+  it to its last field. Its sibling `layouts_by_id` is a bare `u32` key and
+  takes `PtrHashMap`.
+
+  Every key here is a codegen-minted class id or a runtime-minted shape id —
+  never external input — so SipHash's DoS resistance buys nothing.
+
+  **Deliberately left on SipHash**, because their keys are JS-supplied
+  property/member names on paths that can see adversarial input, or because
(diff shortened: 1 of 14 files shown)


proggeramlug:
Merged.

Swapping a hash function under 14 files of class-id-keyed registries is the kind of change where the risk is not speed but *collisions and iteration order* — a weaker hash that collides more, or a registry whose iteration order feeds something observable, shows up as wrong dispatch rather than as a slowdown. Probed 300 objects across a two-level class hierarchy with an overriding `m()` calling `super.m()`, checking method dispatch totals, `instanceof` against both the direct class and the base, `.name` on both constructors, and `Object.getPrototypeOf(B) === A`. Matches node 26.5.1 byte-for-byte.

Class-id keys are dense small integers, so SipHash's collision resistance was buying nothing here — it exists to defend against adversarially chosen keys, and a class id is assigned by the compiler, not by an attacker. That is the right reason for this substitution, and it is why the same argument would not transfer to string-keyed registries.

**Validation:** `perry-runtime` 2853 passed / 0 failed at `RUST_TEST_THREADS=1`; `perry-codegen` 31 suites / 0 failures; clean under `PERRY_GC_FORCE_EVACUATE=1` and a seeded aggressive collection schedule.

Unrelated, so you don't chase it: the `warnings` lint gate is currently red from 12 pre-existing `unnecessary unsafe block` warnings in `perry-ui-macos` (untouched by this PR, its deps unchanged, files last modified 2026-08-20). It appears on any PR that invalidates that crate's check cache. Handling it separately.

PR #9204, opened 2026-08-30
codegen: name the packed-loop admission's rejection reason under PERRY_PACKED_LOOP_TRACE

proggeramlug:
## Why

Whether a counted loop gets its packed-f64 clone is a chain of independent conditions. When one unexpectedly declines, there is no way to tell *which* except by reading the code and guessing.

That is not hypothetical — it is the direct cost of the last few hours. Three attempts on #9151 each guessed a different gate and each was wrong in a different way: the HIR statement predicate (`stmt_is_packed_f64_loop_safe`), then the emitted-block call scan (`contains_gc_unsafe_call`), then a block-region property for the throw's operand. Two were built, measured as **no-ops**, and reverted. The information needed to skip all three was a single string.

## What

`PERRY_PACKED_LOOP_TRACE=1` prints the reason. The eight rejection points in `match_packed_f64_versioned_loop` each name themselves, and the post-emission `fast_call_free` check gets its own line.

That last one is the important addition rather than an afterthought: it is reached when the matcher **admitted** the loop and an emitted block carries a call anyway, so the clone is built and never entered. That case is invisible in the matcher's own rejection reasons, and it is precisely the confusion that cost two of the three attempts.

Off by default, read once through a `OnceLock` — a branch on an already-cold path, nothing when unset.

## It already earned its place

On the #9151 shape:

- a loop containing `try` → `[packed-loop] rejected: body_not_admissible` — correct, and the control that proves the trace fires when it should
- `throw new Error(…)` → **silent**, which proves both the matcher *and* the array-loop call-free scan innocent for that case, and moves the search to whatever decides it later

A silence that rules out two suspects is worth more than three more guesses — but only because the control run establishes that silence means something. I have recorded that pairing on #9151 along with the narrowed residual: on current main, `throw String(l)` and `throw "bad" + l` already take the fast path at 0.58 ns, and **only `new` is still blocked**.

## Gates

perry-codegen 1842/0 — unchanged from main. `-D warnings` clean. Verified the trace fires with the flag set and emits nothing without it.

## Summary by CodeRabbit

* **Diagnostics**
  * Expanded optional packed-loop tracing to identify specific reasons an optimization remains on the generic path.
  * Reports rejected fast clones when generated code contains garbage-collection-unsafe calls.
  * Includes diagnostics for invalid comparisons or counters, unsupported loop bodies or bindings, unknown array types, pending labels, guard failures, and other admission checks.
  * Enable tracing with `PERRY_PACKED_LOOP_TRACE=1`; tracing does not alter generated code or build outputs.

### DIFF
--- a/changelog.d/9204-packed-loop-trace.md
+++ b/changelog.d/9204-packed-loop-trace.md
@@ -1,0 +1,11 @@
+`match_packed_f64_versioned_loop`'s admission decision is a chain of independent
+conditions, and when a loop unexpectedly stayed on the generic path there was no
+way to tell which one declined it without reading the code and guessing — three
+separate attempts on #9151 each found a different gate that way.
+`PERRY_PACKED_LOOP_TRACE=1` names the rejection reason instead.
+
+`packed_loop_reject` prints and returns `None` either way, so the emitted code
+is identical with the flag on and off. It is registered in
+`BUILD_CACHE_ENV_EXCLUSIONS` rather than `BUILD_CACHE_ENV_VARS` for that
+reason: making it a cache input would cost every trace run a rebuild and buy
+nothing.
--- a/crates/perry-codegen/src/stmt/loops.rs
+++ b/crates/perry-codegen/src/stmt/loops.rs
@@ -4103,6 +4103,13 @@ fn lower_object_array_write_versioned_for(
 
     let fast_call_free = (fast_scan_start..ctx.func.num_blocks())
         .all(|idx| !ctx.func.blocks()[idx].contains_gc_unsafe_call());
+    if !fast_call_free {
+        // Reached when the matcher ADMITTED the loop but an emitted block in
+        // the clone carries a call, so the clone is built and never entered.
+        // That is invisible from the matcher's own rejection reasons, which is
+        // why it gets its own trace line.
+        let _ = packed_loop_reject("fast_clone_not_call_free");
+    }
     ctx.current_block = preheader_idx;
     let mut guard_ok = ctx.block().icmp_ne(I64, &packed_slots, "0");
     for (g_packed, _) in &extra_guards {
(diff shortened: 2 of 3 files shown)


proggeramlug:
Merged.

Good instrument, and the motivating story is the argument for it: three separate attempts on #9151 each guessed a different gate. An admission chain of independent conditions where the only feedback is "it didn't fire" is exactly the shape that costs hours to nothing.

**One thing I had to add before it could land.** `codegen_env_vars_are_build_cache_inputs` scans every `env::var("PERRY_…")` under `crates/perry-codegen/src` and fails unless the name is in either `BUILD_CACHE_ENV_VARS` or `BUILD_CACHE_ENV_EXCLUSIONS`. `PERRY_PACKED_LOOP_TRACE` was in neither, so this would have gone red on `main`.

I put it in **exclusions**, not inputs, and the distinction matters: `packed_loop_reject` prints and returns `None` either way, so the emitted object bytes are identical with the flag on and off. Making it a cache *input* would compile — and would silently cost every trace run a full rebuild, which is a real tax on the exact workflow this flag exists to make cheap. Same reasoning the neighbouring `PERRY_OPT_REPORT` entry records, and it's a nice illustration that the gate catches the registration gap but can't tell you which side to register on — that judgement is "does this change emitted code", and here it demonstrably doesn't.

**Validation:** the build-cache gate passes (2/2); `perry-codegen` 31 suites / 0 failures; `perry-runtime` 2853 passed / 0 failed at `RUST_TEST_THREADS=1`; all static lint gates green.

Unrelated to you: `cargo check --workspace --all-targets` is currently red from 12 pre-existing `unnecessary unsafe block` warnings in `perry-ui-macos` — untouched here, deps unchanged, files last modified 2026-08-20 (#8453). Handling separately.

PR #9205, opened 2026-08-30
fix(buffer): distinguish Uint8Array in Buffer.isBuffer

### LINKED ISSUE #9179: Buffer.isBuffer() returns true for a plain Uint8Array
`Buffer.isBuffer()` returns `true` for a plain `Uint8Array`. Node returns `false` — a `Buffer` is a `Uint8Array` subclass, but not every `Uint8Array` is a `Buffer`.

```ts
const u8 = new Uint8Array(4);
console.log(Buffer.isBuffer(Buffer.alloc(4)), Buffer.isBuffer(u8), Buffer.isBuffer({}));
```

| | output |
|---|---|
| node 26.5.1 | `true false false` |
| perry | `true true false` |

Everything else in the same probe (hex/base64 round trips, `Buffer.concat`, `Buffer.compare`, `DataView` aliasing over a shared `ArrayBuffer`, element access, `JSON.stringify(Array.from(u8))`) matches node byte-for-byte.

Cause is structural rather than a one-line predicate bug: `BufferHeader` and the `Uint8Array` representation share a layout and a single address-keyed registry (`BUFFER_REGISTRY`), so "is this address a registered buffer" cannot by itself separate the two kinds. `UINT8ARRAY_FROM_CTOR` already records the discriminator — `is_uint8array_buffer(addr)` is true exactly for the ones that came from `new Uint8Array(...)` — so `Buffer.isBuffer` likely needs to consult it and answer false, with care that a `Buffer` is not itself recorded there.

Found while auditing #9176 (registry-probe prefilter); confirmed **pre-existing on `main`** by A/B — that PR's output is byte-identical to main's on every line of the probe, including this one.

proggeramlug:
## Summary

- add a Node Buffer brand predicate that excludes constructor-created Uint8Array values
- preserve the broader shared-storage predicate used by native APIs that accept Buffer or Uint8Array
- cover direct and first-class Buffer.isBuffer calls

Fixes #9179

## Testing

- RUST_TEST_THREADS=1 cargo test -p perry-runtime buffer_brand_tests -- --nocapture
- cargo test -p perry --test issue_9179_buffer_isbuffer_uint8array -- --nocapture

## Summary by CodeRabbit

* **Bug Fixes**
  * `Buffer.isBuffer()` now correctly returns `false` for plain `Uint8Array` instances and objects, matching Node.js behavior.
  * Buffer-related APIs continue accepting supported `Uint8Array` values where applicable.

* **Tests**
  * Added regression coverage for direct and stored-function calls to `Buffer.isBuffer()`.

### DIFF
--- a/changelog.d/9205-buffer-isbuffer-uint8array-brand.md
+++ b/changelog.d/9205-buffer-isbuffer-uint8array-brand.md
@@ -1,0 +1,9 @@
+`Buffer.isBuffer()` now returns `false` for a plain `Uint8Array`, matching
+Node. Perry stores both values in the same `BufferHeader` layout and buffer
+registry, so registry membership alone could not distinguish their brands.
+
+The public predicate now also consults the existing constructor-created
+`Uint8Array` discriminator. Native APIs keep using the broader storage
+predicate where Node accepts both `Buffer` and `Uint8Array` inputs. Regression
+coverage exercises direct calls and `Buffer.isBuffer` used as a first-class
+function.
--- a/crates/perry-codegen/src/expr/array_methods.rs
+++ b/crates/perry-codegen/src/expr/array_methods.rs
@@ -269,13 +269,14 @@ pub(crate) fn lower(ctx: &mut FnCtx<'_>, expr: &Expr) -> Result<String> {
 
         // -------- BufferIsBuffer --------
         // `Buffer.isBuffer(x)`. Runtime returns i32 (0/1); wrap as NaN-boxed
-        // boolean. `js_buffer_is_buffer` already strips NaN-box tags and
-        // checks the BUFFER_REGISTRY, so any value type is safe to pass.
+        // boolean. `js_buffer_is_node_buffer` strips NaN-box tags, checks the
+        // shared BUFFER_REGISTRY, and excludes constructor-created Uint8Arrays,
+        // so any value type is safe to pass.
         Expr::BufferIsBuffer(operand) => {
             let v_box = lower_expr(ctx, operand)?;
             let blk = ctx.block();
             let v_handle = unbox_to_i64(blk, &v_box);
-            let i32_result = blk.call(I32, "js_buffer_is_buffer", &[(I64, &v_handle)]);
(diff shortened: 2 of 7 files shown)


proggeramlug:
Merged. This closes #9179, which I filed while auditing #9176, and it takes the approach the issue argued for rather than the tempting shortcut.

The tempting version is to narrow `js_buffer_is_buffer` itself. That would have been wrong: native APIs legitimately accept either representation, and Perry deliberately backs both `Buffer` and constructor-created `Uint8Array` with `BufferHeader` in one registry — so the storage question and the identity question are genuinely different questions that happened to share a predicate. Splitting them, keeping `js_buffer_is_buffer` as the broad storage test and adding `js_buffer_is_node_buffer` for the public brand check, is the version that doesn't break the callers that wanted the old meaning. The doc comments now say which is which, which is the part that keeps this from being re-collapsed later.

Nice that the discriminator reuses `is_uint8array_buffer` — the same probe #9176 and #9177 just made cheap, so the brand check costs a range compare in the common case rather than a hash.

**Validation:** the exact reproducer from #9179 now prints `true false false`, matching node 26.5.1, and the rest of that probe (hex/base64 round trips, `concat`, `compare`, `DataView` aliasing over a shared `ArrayBuffer`, element access, `JSON.stringify(Array.from(u8))`) still matches byte-for-byte. `perry-runtime` suite green at `RUST_TEST_THREADS=1`.

Heads-up on a collision: **#9213** ("fix(runtime): distinguish Buffer from Uint8Array", closing #9173) is yours too and overlaps this heavily — same `array_methods.rs`, `third_party.rs`, `buffer/mod.rs`, `buffer/query.rs`, `dispatch_a_c.rs` — while also touching `exotic_view.rs`, `node_api_host/buffers.rs`, `global_this/populate.rs` and `object_ops/prototype.rs`. I merged this one because it was validated and narrower; #9213 will need a rebase, and it's worth checking whether what remains of it is still the change you want or whether this one already covers part of its ground.

PR #9208, opened 2026-08-30
fix(gc): count empty across handle wrappers as debt

### LINKED ISSUE #9152: raw_handle_debt ratchet counts across_*(|| ()) no-ops as conversions (~17 sites on main)

proggeramlug:
## Summary

Teach the raw-handle debt ratchet to count empty `across_{mut,const,nanbox}(|| ())` wrappers as debt, and remove every existing no-op wrapper without raising the baseline.

## Changes

- Match empty `across_*` closures, including multiline, turbofish, `move`, and empty-block forms, in `scripts/raw_handle_debt.py`.
- Expand the ratchet self-test and diagnostics to distinguish real collection boundaries from no-op refreshes.
- Refactor the 17 existing no-op wrappers to scoped handle access or to `across_*` calls containing the actual allocation-capable work.
- Preserve the 963-site baseline and every per-module ceiling.

## Related issue

Closes #9152

## Test plan

- [x] Not user-facing; expanded the ratchet's built-in regression cases
- [x] No runtime API change requiring `docs/src/` updates

Additional checks run on `root@perrymaster.skelpo.net`:

- [x] `cargo check -p perry-runtime --all-features`
- [x] `cargo test -p perry-runtime -- --test-threads=1` (2,832 passed, 4 ignored)
- [x] `python3 scripts/raw_handle_debt.py --self-test`
- [x] `python3 scripts/raw_handle_debt.py`
- [x] `python3 scripts/raw_handle_debt.py --no-raise-vs origin/main`

## Screenshots / output

n/a

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runtime safety when accessing and relocating managed objects during allocation and garbage collection.
  * Preserved iterator, map, module, typed-array, property, and internationalization behavior while ensuring operations use refreshed object references.
  * Enhanced detection of unsafe no-op handle access patterns to prevent future issues.

### DIFF
--- a/changelog.d/9208-raw-handle-debt.md
+++ b/changelog.d/9208-raw-handle-debt.md
@@ -1,0 +1,8 @@
+The raw-handle debt ratchet now counts empty
+`RuntimeHandle::across_{mut,const,nanbox}(|| ())` wrappers as debt. Those
+wrappers refreshed a handle across no work, so they were equivalent to a bare
+pointer read while still receiving credit as a conversion.
+
+All 17 existing no-op wrappers now use scoped handle access or put the real
+allocation-capable operation inside `across_*`. The total baseline and every
+per-module ceiling remain unchanged.
--- a/crates/perry-runtime/src/array/iter_object.rs
+++ b/crates/perry-runtime/src/array/iter_object.rs
@@ -83,8 +83,7 @@ unsafe fn alloc_iterator_backing(backing: f64, kind: i32) -> f64 {
     // `Object.getPrototypeOf(it)` and the inherited `.next` read resolve.
     obj_h
         .with_mut_ptr(|obj| crate::object::attach_iterator_prototype(obj, ARRAY_ITERATOR_CLASS_ID));
-    let (_, obj) = obj_h.across_mut::<ObjectHeader, _>(|| ());
-    js_nanbox_pointer(obj as i64)
+    obj_h.with_mut_ptr::<ObjectHeader, _>(|obj| js_nanbox_pointer(obj as i64))
 }
 
 unsafe fn alloc_iterator(arr_ptr: *mut ArrayHeader, kind: i32) -> f64 {
--- a/crates/perry-runtime/src/builtins/formatting/boxed_primitives.rs
+++ b/crates/perry-runtime/src/builtins/formatting/boxed_primitives.rs
@@ -383,8 +383,9 @@ pub extern "C" fn js_boxed_symbol_new(value: f64) -> f64 {
     obj.with_mut_ptr::<crate::object::ObjectHeader, _>(|obj| {
         attach_boxed_primitive_prototype(obj, CLASS_ID_BOXED_SYMBOL)
     });
-    let (_, obj) = obj.across_mut::<crate::object::ObjectHeader, _>(|| ());
-    crate::value::js_nanbox_pointer(obj as i64)
+    obj.with_mut_ptr::<crate::object::ObjectHeader, _>(|obj| {
+        crate::value::js_nanbox_pointer(obj as i64)
+    })
 }
 
 #[cfg(test)]
--- a/crates/perry-runtime/src/collection_iter_object.rs
+++ b/crates/perry-runtime/src/collection_iter_object.rs
@@ -71,35 +71,40 @@ unsafe fn alloc_iterator(class_id: u32, coll_nanboxed: f64, kind: i32) -> f64 {
     let scope = crate::gc::RuntimeHandleScope::new();
     let coll_h = scope.root_nanbox_f64(coll_nanboxed);
     let obj_h = scope.root_raw_mut_ptr(js_object_alloc(class_id, 6));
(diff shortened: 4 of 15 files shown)


proggeramlug:
Merged. Closes #9152, which I filed after finding these wrappers while reviewing #9103 — and this is the right resolution rather than the one I could have pushed for.

I'd raised it as "these sites are vacuous"; the fix is better than deleting them, because it makes the *gate* able to see the shape. An `across_*(|| ())` is byte-identical to the bare read it replaces, so a ratchet that counts bare reads but not empty wrappers can be satisfied by a rewrite that changes nothing — which is a gate that can't fail in the one direction that matters. Counting them as debt, and clearing all 17 without raising the baseline, closes it properly.

**Checked the matcher for under-counting**, because that's a scanner's dangerous failure direction — a regex that silently stops matching reports zero and passes forever. Your self-test already asserts both directions with turbofish, multiline `|| ( )`, `move || {}` and `across_nanbox` on the match side, and real-work closures plus `with_*_ptr` and `get_nanbox_f64` on the must-not side. I probed six shapes it doesn't enumerate — `(||())` with no whitespace, `( || () )` padded, a newline between `||` and `()`, `move || ()`, nested generics `::<Foo<Bar>, _>`, and `|| {  }` — and all six match, while `|| do_work()` and `with_mut_ptr` correctly don't. No gap found.

**Ran all three CI invocations**, not just one, since `--no-raise-vs` passing does not imply the bare run passes — the bare run is what locks unlisted modules at zero:

- `--self-test` → ok, 963 sites across 113 files, all three per-module rules fire
- bare → 963 (baseline 963), 113 modules within ceilings, every other module locked at zero
- `--no-raise-vs origin/main` → baseline 963 → 963, no ceiling raised

So the 17 removals genuinely landed without buying headroom, which is the claim worth verifying.

**Validation:** `perry-runtime` suite green at `RUST_TEST_THREADS=1`; workspace lint gates green.

PR #9209, opened 2026-08-30
test(perry): isolate ws IR checks from runtime archives

### LINKED ISSUE #9172: ws_client_handle_cross_function_dispatch: 9 of 11 tests fail on main
`crates/perry/tests/ws_client_handle_cross_function_dispatch.rs` fails **9 of 11** tests on current `main` (`3411c9d7c8`).

```
cargo test --profile perry-dev -p perry --test ws_client_handle_cross_function_dispatch
test result: FAILED. 2 passed; 9 failed
```

Failing: `helper_wsid_send_dispatches_to_client_runtime`, `same_named_function_is_not_cross_tagged`, `transitive_through_polymorphic_intermediate_is_not_tagged`, `transitively_forwarded_wsid_send_dispatches_to_client_runtime`, `shadowing_nested_param_does_not_inherit_taint`, `leading_this_param_keeps_hint_index_aligned`, and three others — all at the same `assert!(compile.status.success(), ...)` on line 49.

### Why it isn't caught

Per `docs/src/testing/ci-tiers.md`, integration suites under `crates/*/tests/` run per-PR only when the diff names them (`e2e-scoped`); the sweep/full tiers' `cargo test --workspace` is the backstop. So a suite can sit red for a while before anyone touches a file that pulls it in. #9163 is currently open against this file, which is how I noticed.

### A caveat on reproducing it — this cost me two wrong readings

My first run showed **11 of 11** failing, and the extra two were an artifact of my own environment. The compiler and the runtime archives had been built from different commits, and perry's coherence guard refuses:

```
  library build: v0.5.1519 (commit 5f08e77c7cf7)
  Perry build:   v0.5.1519 (commit 3411c9d7c888)
```

That refusal surfaces through the test as a bare `perry compile failed`, which reads exactly like a product bug. Rebuilding `-p perry -p perry-runtime-static -p perry-stdlib-static` together in one invocation gives the true figure of 9. Anyone triaging this should confirm the coherence line is absent before trusting a count — and note the guard names both commits, which is what makes it diagnosable.

I have not investigated the 9 real failures; the compile output before the assertion shows codegen and "Linking (runtime-only)…" completing, so the failure is after that point.

proggeramlug:
## Summary

Make the WebSocket cross-function dispatch suite test HIR/codegen independently of separately built runtime and extension archives, while preserving its one semantic native-execution check.

## Changes

- Compile IR-only assertions with `--no-auto-optimize --no-link` and disable well-known archive discovery.
- Keep the same-named nested-function semantic case linked and executed, with its `http` and `ws` providers made explicit.
- Preserve the existing non-WebSocket runtime control unchanged.

## Related issue

Closes #9172

## Test plan

- `cargo test --profile perry-dev -p perry --test ws_client_handle_cross_function_dispatch` — 11 passed, 0 failed (7.26s warm)
- `cargo fmt --all -- --check` — clean
- `scripts/check_file_size.sh` — pass

- [ ] (if user-facing) Added or updated a test under `test-files/` or a `#[test]` in the affected crate
- [ ] (if CLI / stdlib / runtime API changed) Updated `docs/src/`

## Screenshots / output

n/a (test-only change)

## Checklist

## Summary by CodeRabbit

* **Tests**
  * Expanded cross-function dispatch coverage across linked and non-linked compilation scenarios.
  * Improved regression checks for same-named handlers and WebSocket behavior.
  * Streamlined test execution by removing unnecessary runtime and extension dependencies from IR-only checks.

* **Documentation**
  * Added a changelog entry describing the updated WebSocket dispatch regression coverage.

### DIFF
--- a/changelog.d/9209-ws-ir-only-tests.md
+++ b/changelog.d/9209-ws-ir-only-tests.md
@@ -1,0 +1,5 @@
+The WebSocket cross-function dispatch regression suite now compiles its
+IR-inspection cases with linking disabled, so those checks no longer rebuild
+or depend on runtime and extension archives. The one case that verifies
+same-named nested handlers at runtime still links and runs with its required
+well-known modules forced on.
--- a/crates/perry/tests/ws_client_handle_cross_function_dispatch.rs
+++ b/crates/perry/tests/ws_client_handle_cross_function_dispatch.rs
@@ -26,14 +26,15 @@ fn perry_bin() -> PathBuf {
 
 /// Compile `source` and return the concatenated LLVM IR text of all emitted
 /// modules (via `PERRY_SAVE_LL`). Asserts the compile succeeds.
-fn compile_to_ir(dir: &Path, source: &str) -> String {
+fn compile_to_ir_impl(dir: &Path, source: &str, link: bool) -> String {
     let entry = dir.join("main.ts");
     let output = dir.join("main_bin");
     let ll_dir = dir.join("ll");
     std::fs::create_dir_all(&ll_dir).expect("mk ll dir");
     std::fs::write(&entry, source).expect("write entry");
 
-    let compile = Command::new(perry_bin())
+    let mut command = Command::new(perry_bin());
+    command
         .current_dir(dir)
         .arg("compile")
         .arg(&entry)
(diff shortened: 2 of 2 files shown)


proggeramlug:
Merged.

This is the right fix for a trap I hit myself earlier today: the ws IR checks were reading whatever runtime archive happened to be on disk, and a stale or foreign `.a` makes the test measure something other than the tree it claims to test. My own notes on this repo say a `cargo test -p perry --test X` links a stale archive unless the `-static` wrappers were built first — isolating the check from ambient archives removes the class instead of documenting it.

Worth noting it also removes a false-failure source, not just a false-pass one: I saw `ws_client_handle_cross_function_dispatch` read as 11/11 red earlier from commit skew between the compiler and the runtime archive, when the true figure was 9/11 (filed as #9172). A test that pins its own inputs can't produce that kind of red either.

**Validation:** `perry-runtime` 2855 passed / 0 failed at `RUST_TEST_THREADS=1`; `perry-codegen` 31 suites / 0 failures; all 60 lint gates green. `PERRY_LLVM_KEEP_IR` is a test-harness knob outside `crates/perry-codegen/src`, so `codegen_env_vars_are_build_cache_inputs` doesn't apply to it — the build-cache gate passes unchanged.

PR #9211, opened 2026-08-30
perf(runtime): cache handle stack after TLS publication

### LINKED ISSUE #9183: RuntimeHandleScope: two TLS resolutions per handle (Darwin-only cost) — and the profile's 4.3% is mostly inlined closure work, not the wrapper
## Summary

`RuntimeHandleScope` resolves its thread-local **twice per handle** — once in `new()` to read a length, once in `push()` — and once more on every handle read via `with_slot`. Every runtime helper that may allocate opens a scope and roots its locals through it, so these are among the hottest thread-local accesses in the runtime. In the `cc --help` profile the `RuntimeHandle::across_mut` + `RuntimeHandleScope::push` group is **~4.3%**.

The runtime already has the mechanism to fix this — `crate::tls_hot`, a per-thread address cache that gives "one TLS access for every thread-local it covers". `arena/page_meta.rs` uses it for exactly this reason, and its comment records the original motivation:

> this runs on every `decode_heap_addr` … and was the single largest `_tlv_get_addr` caller on `churn.ts` (116 of 653 attributed samples) precisely because it resolved two distinct thread-locals per call.

`RUNTIME_HANDLE_STACK` is not in that cache.

## Why the obvious fix does not work

I implemented it — add `runtime_handle_stack` to `HotTls`, fill it from a `runtime_handle_stack_hot_addr()`, and route `new`, `push`, `with_slot` and `with_slot_mut` through the cached address — and it **stack-overflows**, reproducibly, in `agent_dispatch_tests::a_worker_neither_fires_nor_eats_a_primary_agent_timer`.

The cause is an initialization cycle:

1. `RuntimeHandleScope::new()` now calls `tls_hot::hot()`.
2. On a fresh thread `hot()` is unpublished, so it calls `hot_uncached()` → `fill()`.
3. `fill()` calls `runtime_handle_stack_hot_addr()`, which touches `RUNTIME_HANDLE_STACK`.
4. Anything that allocates on that path opens a handle scope — and we are back at step 1, with `hot()` still unfilled.

Every other member of `HotTls` is reached from paths that run *after* thread setup. The handle stack is not: it is used during initialization, which is what makes it different from its neighbours in the cache.

## What a working version needs

A way to read the cached address **without triggering `fill()`** — e.g. a `hot_if_published()` that returns `Option<&HotTls>` from the raw TSD slot — with the handle accessor falling back to the ordinary `thread_local!` path when the cache is not yet published. That keeps the fast path for the 99.9% of handle operations that happen after startup, and leaves initialization on the existing path where it cannot recurse.

That is a small change but it touches thread setup, so it wants its own validation rather than being folded into a perf commit.

## One thing to know before starting

`HotTls` field order is a **compile-time contract with codegen**:

```rust
const _: () = assert!(std::mem::offset_of!(HotTls, implicit_this) == HOT_TLS_IMPLICIT_THIS_OFFSET);
```

A new field inserted ahead of `implicit_this` shifts it and fails the build. Append at the end. (The assertion caught this immediately, which is the system working.)

## Status

Reverted, not shipped. `perry-runtime` is back to 2844/0. This issue records the target, the mechanism that should serve it, and the specific reason the direct application fails.

---

## Correction: `tls_hot` is the wrong instrument for the *profiled* platform

Added after the revert, because it changes the recommendation and would otherwise send the next person down the path I took.

`tls_hot` exists to eliminate a **Darwin-specific call**. Its own module documentation is explicit:

> On Darwin every `thread_local!` access is an out-of-line call to `_tlv_get_addr` in `libdyld`. Unlike ELF's `local-exec` / `initial-exec` models it is a real call — not inlined, not cached across accesses…

The `cc --help` profile that attributes ~4.3% to this group is recorded on **Linux**, where a `thread_local!` access is a fixed offset with no accessor call at all. So on the platform where the 4.3% was measured, the repeated thread-local resolutions I set out to remove **cost essentially nothing**, and adding `RUNTIME_HANDLE_STACK` to the cache would have measured flat there no matter how cleanly the initialization cycle was solved.

The premise of the original write-up above — "two TLS resolutions per handle" — is therefore a real inefficiency on Darwin and a non-issue on Linux.

**What the Linux 4.3% must actually be**, since it is not TLS resolution: the `RefCell` borrow and its flag write on every scope open, push and read; the `Vec` push/pop traffic; `runtime_handle_slot_write_barrier` on every `push`; and the sheer number of handles opened, given that every runtime helper which may allocate opens a scope and roots its locals through it.

That points somewhere different from a TLS cache — reduce the number of handles, or make the per-handle operation cheaper (a plain array with an index instead of `RefCell<Vec<..>>`, or skipping the barrier for slots that provably cannot hold a pointer). Worth profiling the group on Darwin **and** Linux separately before choosing, since the two platforms have genuinely different costs in the same code.

The `hot_if_filled()` design sketched above is still the correct way to add this thread-local to the cache **if** someone wants the Darwin win; it just should not be expected to move the number the profile reported.

---

## Second correction: most of that 4.3% is not handle bookkeeping at all

The two symbols the profile attributes this cost to are:

```
RuntimeHandle::across_mut::<ArrayHeader, JSValue, obj_value_has_own_key::{closure#3}>   2.44%
RuntimeHandle::across_mut::<StringHeader, (), js_object_define_property::{closure#65}>  1.81%
```

`across_mut` is three lines and `#[inline]`:

```rust
pub fn across_mut<T, R>(&self, f: impl FnOnce() -> R) -> (R, *mut T) {
    let result = f();
    (result, self.get_raw_mut_ptr::<T>())
}
```

It is generic over the closure, so each call site produces its own monomorphization **with the closure body inlined into it** — which is why the closure's name appears inside the symbol. The self-time recorded against these symbols is therefore mostly the work of `obj_value_has_own_key` and `js_object_define_property`, not the handle machinery wrapping them.

The handle-side cost that is genuinely there is one `get_raw_mut_ptr` per call, and on Linux that is a fixed-offset TLS read plus a `RefCell` borrow and an index. The write barrier is not it either: `incremental_mark_barrier_value` short-circuits on `incremental_mark_barrier_globally_idle()`, which is one static load and a branch when no cycle is active.

**So "handle scopes are ~4.3%" is not a supportable reading of this profile.** The two symbols are hot because the operations they wrap are hot. Optimizing the wrapper cannot recover that share, and anyone who tries will measure flat.

Same shape as `js_array_get_f64` appearing as the largest symbol while generated code shows as 0.11%: in both cases the symbol attributes work done *through* it rather than *by* it. Reading a profile of a runtime full of inlined generic wrappers requires checking whether a symbol's self-time is its own code before ranking it as a target.

## Revised status

The two TLS resolutions per handle are real on Darwin and worth fixing there with the `hot_if_filled()` design above, but they are close to free on Linux and were never the 4.3%. There is no large, well-evidenced target here. Left alone.

proggeramlug:
## What

Closes #9183.

- Append the runtime-handle stack address to `HotTls` without moving any codegen-consumed offsets.
- Add `hot_if_published()`, which reads only the already-published Darwin TSD slot and never enters `fill()`.
- Route scope creation, pushes, reads, scans, savepoints, and FFI roots through the named pointer on Apple aarch64.
- Keep startup and non-Darwin accesses on the ordinary thread-local path, so initialization cannot recurse and Linux does not gain an unnecessary cache indirection.
- Clear the named pointer before the handle stack's TLS destructor runs, preserving the teardown protection the previous generic cache supplied.

This deliberately claims only the Darwin TLS win described by the corrected issue. It does not attribute or promise recovery of the Linux profile's inlined closure work.

## Initialization contract

`RUNTIME_HANDLE_STACK` is the exceptional raw TLS declaration because it is needed while `HotTls` itself is being initialized. Its address is filled before the existing `temp_roots` readiness sentinel, but handle operations consult it only after the Darwin cache has been published. Before publication they use the raw declaration directly.

A fresh-thread regression exercises scope creation, push, read, and drop while asserting that `HotTls` remains unfilled. The exact worker timer test that stack-overflowed under the direct implementation is also covered.

## Validation

- `cargo test --profile perry-dev -p perry-runtime tls_hot::tests::runtime_handles_do_not_fill_an_unpublished_cache -- --exact --nocapture`
- `cargo test --profile perry-dev -p perry-runtime agent_dispatch_tests::a_worker_neither_fires_nor_eats_a_primary_agent_timer -- --exact --nocapture`
- `cargo test --profile perry-dev -p perry-runtime -- --test-threads=1 --quiet` — 2,833 passed, 4 ignored; doc tests 0 failed
- `cargo check --profile perry-dev -p perry-runtime --target aarch64-apple-darwin --no-default-features`
- Apple aarch64 `--emit=asm` inspection: steady `new`/`push` use `PERRY_HOT_TSD_KEY`, `mrs TPIDRRO_EL0`, and the appended fixed-offset load; only the unpublished branch reaches the TLV accessor
- `python3 scripts/check_thread_locals.py`

No version bump.

## Summary by CodeRabbit

* **Performance**
  * Improved runtime handle operations on Apple silicon by reusing a cached thread-local address, reducing repeated lookups.
  * Preserved reliable fallback behavior during initialization and on other platforms.

* **Reliability**
  * Added teardown safeguards to prevent access to runtime handle data after thread-local storage is destroyed.
  * Expanded coverage for initialization and cache population scenarios.

### DIFF
--- a/changelog.d/9211-runtime-handle-tls.md
+++ b/changelog.d/9211-runtime-handle-tls.md
@@ -1,0 +1,6 @@
+On Apple aarch64, runtime-handle scope creation, pushes and reads now take the
+handle-stack address from the already-published `HotTls` cache instead of
+resolving the thread-local on every operation. Startup retains a raw TLS
+fallback that cannot recursively enter `HotTls::fill`, and a teardown guard
+clears the cached address before the stack storage is destroyed. Other targets
+keep their ordinary fixed-offset TLS path.
--- a/crates/perry-runtime/src/gc/roots.rs
+++ b/crates/perry-runtime/src/gc/roots.rs
@@ -20,7 +20,9 @@ pub(super) use runtime_handles::{
     new_runtime_handle_root_scan_state, scan_runtime_handle_roots_mut,
     scan_runtime_handle_roots_mut_step,
 };
-pub(crate) use runtime_handles::{runtime_handle_stack_restore, runtime_handle_stack_savepoint};
+pub(crate) use runtime_handles::{
+    runtime_handle_stack_hot_addr, runtime_handle_stack_restore, runtime_handle_stack_savepoint,
+};
 pub use runtime_handles::{RuntimeHandle, RuntimeHandleScope};
 pub use scanner_shims::{
     async_context_mutable_root_scanner, async_context_root_scanner,
@@ -108,7 +110,6 @@ crate::perry_thread_local! {
     pub(super) static FFI_MUTABLE_ROOT_SCANNERS: RefCell<Vec<PerryFfiMutableRootScanner>> = RefCell::new(Vec::new());
     pub(super) static FFI_NAMED_MUTABLE_ROOT_SCANNERS: RefCell<Vec<(PerryFfiNamedMutableRootScanner, usize)>> = RefCell::new(Vec::new());
     pub(super) static GLOBAL_ROOTS: RefCell<Vec<*mut u64>> = const { RefCell::new(Vec::new()) };
-    pub(super) static RUNTIME_HANDLE_STACK: RefCell<Vec<RuntimeHandleSlot>> = const { RefCell::new(Vec::new()) };
     pub(super) static GC_ROOT_LOCK_DEPTH: Cell<usize> = const { Cell::new(0) };
 }
 
--- a/crates/perry-runtime/src/gc/roots/runtime_handles.rs
+++ b/crates/perry-runtime/src/gc/roots/runtime_handles.rs
@@ -1,5 +1,54 @@
 use super::*;
 
+struct RuntimeHandleStackHotGuard;
+
+impl Drop for RuntimeHandleStackHotGuard {
+    fn drop(&mut self) {
+        crate::tls_hot::unpublish_runtime_handle_stack();
+    }
+}
+
+thread_local! {
+    /// This stays a raw `thread_local!` as the initialization fallback for its
+    /// named `HotTls` slot. Going through `perry_thread_local!` here would call
+    /// `hot()` while `HotTls::fill` is still resolving this address (#9183).
+    static RUNTIME_HANDLE_STACK: RefCell<Vec<RuntimeHandleSlot>> = const { RefCell::new(Vec::new()) };
+    /// Registered after the stack itself, so reverse-order TLS destruction
+    /// clears the named cache pointer before the stack storage is destroyed.
+    static RUNTIME_HANDLE_STACK_HOT_GUARD: RuntimeHandleStackHotGuard = const { RuntimeHandleStackHotGuard };
+}
+
+/// Resolve this thread's runtime-handle stack without entering `HotTls`.
+/// Called once by `tls_hot::fill`, and by handle operations only until the
+/// Darwin direct-TSD cache has been published.
+#[inline(always)]
+pub(crate) fn runtime_handle_stack_hot_addr() -> *mut u8 {
+    let addr = RUNTIME_HANDLE_STACK.with(|stack| stack as *const _ as *mut u8);
+    RUNTIME_HANDLE_STACK_HOT_GUARD.with(|_| {});
+    addr
+}
+
+/// Borrow this thread's transient-handle stack.
+///
+/// On Apple aarch64 the steady-state path reads the address from an already
+/// published `HotTls`. During `HotTls::fill` (and on every other target) it
+/// uses the ordinary thread-local directly, so opening a handle scope cannot
+/// recursively initialize the cache (#9183).
+#[inline(always)]
+fn with_runtime_handle_stack<R>(f: impl FnOnce(&RefCell<Vec<RuntimeHandleSlot>>) -> R) -> R {
+    if let Some(hot) = crate::tls_hot::hot_if_published() {
+        let stack = hot.runtime_handle_stack.get();
+        if stack.is_null() {
+            return RUNTIME_HANDLE_STACK.with(f);
+        }
+        // SAFETY: `fill` stores this thread's `RUNTIME_HANDLE_STACK` address
+        // before publishing the cache, and the address is stable for the
+        // lifetime of the thread.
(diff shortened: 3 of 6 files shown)


proggeramlug:
Merged.

Caching a raw pointer to the handle stack is the shape my notes flag as the dangerous one — a runtime-side cache of a heap pointer is a GC root the static `gc_root_dominance_check.py` cannot see, because it reads emitted LLVM IR and a TSD slot is structurally invisible to it. So the questions worth asking are custody and teardown, not speed.

Both are handled, and the gate agreed in a way worth recording: `gc_runtime_root_holders.py` went **red** after this change — not because the cache is unrooted, but because `RUNTIME_HANDLE_STACK`'s `frontier` pin went *stale*. Routing the stack through a registered scanner is exactly what "no longer an uncovered holder" means, and the inventory fails on a stale entry by design. I deleted the pin and pushed it here: the deletion is the receipt, and leaving it would have kept a fixed holder listed as debt. Frontier 155 → 154.

The three design choices that make this safe are the ones I'd have asked for: reading only the already-published TSD slot via `hot_if_published()` so a cache miss can never re-enter `fill()` and recurse during init; keeping startup and non-Darwin on the ordinary thread-local path rather than adding an indirection Linux doesn't need; and clearing the named pointer *before* the handle stack's TLS destructor runs, so nothing can observe a pointer to a destroyed stack during teardown. Appending to `HotTls` without moving codegen-consumed offsets is the other half — a reordering there would have been an ABI break with no compile error.

**Validation:** `perry-runtime` 2855 passed / 0 failed at `RUST_TEST_THREADS=1`; `perry-codegen` 31 suites / 0 failures; all 60 lint gates green. Five GC-heavy programs (linked-list churn with a rolling live set, symbol-keyed lookup, throw-under-churn, typed-array + class-hierarchy work, and a 400-key receiver walk) all exit 0 and produce byte-identical output under default, `PERRY_GC_FORCE_EVACUATE=1`, and `PERRY_GC_PROTECT_FROMSPACE=1` with `PERRY_GC_SCHEDULE_SEED=9 PERRY_GC_SCHEDULE_RATE=1` — the arm that would fault at the faulting instruction if a stale from-space pointer were being dereferenced through the cache.

PR #9213, opened 2026-08-30
fix(buffer): complete Buffer brand and prototype chain

### LINKED ISSUE #9173: Buffer.isBuffer(new Uint8Array(...)) returns true — registry conflates Buffers with Uint8Arrays

proggeramlug:
Closes #9173

## Summary

- complete the public `Buffer.isBuffer` brand check by rejecting `ArrayBuffer` and `DataView` while preserving the broader internal BufferHeader storage probe
- resolve Buffer instances to `Buffer.prototype` and link that prototype to `Uint8Array.prototype`
- share the strict Buffer predicate with the N-API path
- cover Buffer, Uint8Array, ArrayBuffer, DataView, captured-call, and prototype-chain behavior

## Testing

- `cargo test -p perry-runtime node_buffer --lib`
- `cargo test -p perry-runtime only_array_buffer_and_data_view_are_non_indexed_views --lib`
- `cargo test -p perry --test issue_9173_buffer_identity -- --nocapture`
- Buffer Node differential suite: 134/134 passing
- `git diff --check`

No version bump.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved `Buffer.isBuffer` accuracy so it rejects `ArrayBuffer`, `DataView`, and plain `Uint8Array` values.
  - Updated Buffer prototype relationships to match Node.js behavior.
  - Improved Buffer identity and prototype checks across typed-array scenarios.

- **Tests**
  - Added regression coverage for Buffer branding, inheritance, and identity behavior.

- **Documentation**
  - Documented the corrected Buffer detection and prototype chain behavior.

### DIFF
--- a/changelog.d/9213-buffer-prototype-chain.md
+++ b/changelog.d/9213-buffer-prototype-chain.md
@@ -1,0 +1,5 @@
+### Fixed
+
+- `Buffer.isBuffer` now rejects non-Buffer byte-storage objects such as
+  `ArrayBuffer` and `DataView`, while Buffer instances inherit through
+  `Buffer.prototype` and `Uint8Array.prototype` as they do in Node.
--- a/crates/perry-runtime/src/buffer/exotic_view.rs
+++ b/crates/perry-runtime/src/buffer/exotic_view.rs
@@ -65,6 +65,21 @@ pub fn is_non_indexed_buffer_view(addr: usize) -> bool {
     super::is_any_array_buffer(addr) || super::is_data_view(addr)
 }
 
+/// `true` only for a Node `Buffer`, not for another JS type that happens to
+/// share Perry's `BufferHeader` storage and registry entry.
+///
+/// Keep this stricter than [`is_byte_indexed_buffer`]: a plain `Uint8Array` is
+/// also byte-indexed, but `Buffer.isBuffer(new Uint8Array(...))` must be false.
+/// ArrayBuffer, SharedArrayBuffer, and DataView are excluded independently
+/// because none of them carries the Uint8Array-constructor marker.
+#[inline]
+pub fn is_node_buffer(addr: usize) -> bool {
+    super::is_registered_buffer(addr)
+        && !super::is_any_array_buffer(addr)
+        && !super::is_data_view(addr)
+        && !super::is_uint8array_buffer(addr)
+}
+
 /// The own-property key a NUMERIC computed key names on a non-indexed buffer
 /// view, or `None` when the key is not a number at all.
 ///
--- a/crates/perry-runtime/src/buffer/exotic_view_tests.rs
+++ b/crates/perry-runtime/src/buffer/exotic_view_tests.rs
@@ -114,6 +114,11 @@ fn only_array_buffer_and_data_view_are_non_indexed_views() {
     assert!(crate::buffer::is_byte_indexed_buffer(b));
     assert!(crate::buffer::is_byte_indexed_buffer(u8a));
 
+    assert!(crate::buffer::is_node_buffer(b));
+    assert!(!crate::buffer::is_node_buffer(u8a));
+    assert!(!crate::buffer::is_node_buffer(ab));
+    assert!(!crate::buffer::is_node_buffer(dv));
+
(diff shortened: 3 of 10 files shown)


proggeramlug:
Merged, on top of #9205.

The two overlap heavily (five shared files) and I landed #9205 first because it was validated and narrower. This completes it: #9205 split the storage predicate from the brand check, and this carries the same distinction through `exotic_view`, `node_api_host/buffers`, `global_this/populate` and `object_ops/prototype`, so the prototype chain agrees with the brand rather than only `Buffer.isBuffer` doing so. A half-applied identity distinction is arguably worse than none — `isBuffer` saying one thing while `getPrototypeOf` says another is the shape of #9179 all over again — so having both is the right end state.

**Validation:** `Buffer.isBuffer(b)`/`isBuffer(u8)`, `b instanceof Uint8Array`, and `Object.getPrototypeOf(b) === Buffer.prototype` all match node 26.5.1, alongside hex/base64 round trips, `concat`, `compare`, and `DataView` aliasing over a shared `ArrayBuffer`. `perry-runtime` 2867 passed / 0 failed at `RUST_TEST_THREADS=1`; `perry-codegen` 31 suites / 0 failures; all 60 lint gates green.

Validated alongside #9214, #9216, #9219, #9224 and #9230.

PR #9214, opened 2026-08-30
perf(runtime): make declared-prototype reverse lookup O(1)

### LINKED ISSUE #9180: class_id_for_decl_prototype_object is a linear scan — 3.10% of cc --help, and a pointer-keyed cache is harder than it looks
## Summary

`class_id_for_decl_prototype_object` answers "which declared class's `.prototype` is this heap object?" with a **linear scan** of `CLASS_DECL_PROTOTYPE_OBJECTS`. It is **3.10% of `cc --help`** — the third-largest symbol in that profile.

## Why the scan is hot

The comment above it explains the design:

> Linear scan — the table is small (one entry per materialized declared-class prototype) and this only runs on the reflection slow path.

Neither premise holds for a large application. cc materializes a large number of declared-class prototypes, and the callers are not a slow path: `descriptor_state.rs:280` asks

```rust
|| class_registry::class_id_for_decl_prototype_object(obj).is_some();
```

about **arbitrary objects**, so the common case is a **miss** — and a miss walks every entry in the table. Same shape in `descriptors.rs:936`, `descriptors.rs:1071`, `delete_rest.rs:288`, `delete_rest.rs:456`, `proxy.rs:1221`, `proxy/metadata.rs:129` and `symbol/iterator.rs:66`.

The forward direction is already an O(1) `HashMap` lookup. Only the reverse is linear.

## Why the obvious fix is not obvious

I attempted a pointer-keyed reverse index and **it was wrong** — reporting it so the next attempt starts from the constraint rather than rediscovering it.

Two things make this harder than it looks:

1. **The collector rewrites these pointers.** `gc_roots` visits `map.values_mut()` and updates each address on evacuation, so any pointer-keyed structure has stale keys after a move. I tried stamping the cache with `gc::instruments::moved_objects_total()` and rebuilding when it changes — sound in principle, since if nothing moved no key can be stale.

2. **There are six writers, not one.** `class_registry/state.rs` (the store), `proxy/metadata.rs:321`, `class_gc_roots.rs:54`, `:87`, `:149`, and `class_registry/gc_roots.rs:119`, `:398`. I invalidated at two and the cache went stale at the others — the failure was **silent and produced wrong answers**, not a crash: `Object.getOwnPropertyDescriptor(C.prototype, "g")` returned `undefined` where node returns a getter/setter descriptor, because the reverse lookup missed and the accessor stopped being surfaced as an own property.

A differential over descriptor shapes, prototype identity, `isPrototypeOf`, `getOwnPropertyNames` and `delete` caught it immediately, including on a plain non-GC run.

## What a correct fix needs

Not another enumeration of invalidation sites — that is what failed. Either:

- **Make the invalidation structural**: route every mutable access through a single accessor that maintains (or drops) the reverse index, so a new writer cannot forget. Six call sites to convert, and then the property is enforced by construction rather than by diligence.
- **Or avoid a derived structure entirely**: store the class id *on* the prototype object, making the reverse question a field read with no table, no staleness and no invalidation.

The second is cleaner if a spare field exists; the first is mechanical and local.

## Not attempted

The linear scan is left in place. This issue records the target, the measurement, and the specific way the first attempt was wrong.

proggeramlug:
## Summary

Replace the linear `class_id_for_decl_prototype_object` scan with a GC-aware bidirectional declared-prototype table, making the common reverse lookup constant-time while retaining the authoritative scan as a safe fallback.

## Changes

- Encapsulate declared-prototype storage behind `DeclPrototypeTable`, with coherent forward and pointer-keyed reverse maps.
- Rekey the reverse index during full and per-slot GC relocation, and route every mutation site through the table API.
- Abandon the optimization safely if duplicate addresses or class-id repointing would make the inverse ambiguous.
- Add table invariant tests and compiled-program coverage for descriptors, identity, deletion, late materialization, and allocation churn.
- Make the integration test build and select its exact matching runtime archive to avoid stale host artifacts.

## Related issue

Closes #9180

## Test plan

- [x] `cargo check -p perry-runtime --all-features`
- [x] `cargo test -p perry-runtime decl_prototype_table::tests -- --test-threads=1` (8 passed)
- [x] `cargo test -p perry-runtime -- --test-threads=1` (2,840 passed, 4 ignored)
- [x] `RUST_MIN_STACK=33554432 cargo test -p perry --test issue_9180_decl_prototype_reverse_lookup -- --test-threads=1` (2 passed)
- [x] `RUST_MIN_STACK=33554432 cargo test --release -p perry --test issue_9180_decl_prototype_reverse_lookup -- --test-threads=1` (2 passed on the final rebased SHA)

- [ ] Full default `cargo build --release` (the affected release targets were built by the release integration test)
- [ ] Full cross-platform workspace suite
- [x] Added tests in the affected crate
- [x] Documentation update not required; this is an internal runtime implementation
- [x] Platform UI build not applicable

## Screenshots / output

With 400 materialized declared-class prototypes, non-prototype receiver lookups measured:

- `Object.getOwnPropertyDescriptor`: 420 ns to 200 ns (2.10x)
- `Object.defineProperty`: 1,585 ns to 1,055 ns (1.50x)
- `delete`: 2,685 ns to 2,445 ns

The former miss cost grew from 180/210/305/420 ns at 0/50/200/400 prototypes; the indexed path stays near 200 ns.

## Checklist

- [x] My commits follow the loose conventional prefix style used in the log

## Summary by CodeRabbit

* **Performance**
  * Improved class prototype lookups, reducing overhead for prototype-related operations.

* **Bug Fixes**
  * Preserved prototype identity and descriptor behavior after garbage collection and memory movement.
  * Corrected reverse lookups for newly registered and removed prototypes.
  * Ensured deleted methods are no longer available on instances.

* **Tests**
  * Added coverage for prototype ownership, identity, descriptor access, allocation churn, and deletion scenarios.


proggeramlug:
Adding the measurement and the staleness argument, which aren't in the description yet. All of this was produced on `84185b5656` — note that predates #9203/#9191/#9204, now in the denominator.

## cc --help: −2.09%

Median of 5 interleaved reps, `perf stat -e instructions:u`, both arms built in one session from the same SHA:

| | instructions |
|---|---|
| base | 8,830,630,765 |
| this | 8,645,755,493 |

**−184.9 M instructions, −2.09%** (min-to-min −2.14%). Within-arm spread 0.29% / 0.44%, so the effect is ~5× the noise. `rc=0` both, `--help` output **byte-identical**.

## The signature that disappears is the slope, not the constant

The scan's cost is proportional to table size, so that is what the fix has to remove — a lower constant would prove nothing:

| op, non-prototype receivers | 0 | 50 | 200 | 400 protos | slope 50→400 |
|---|---|---|---|---|---|
| `getOwnPropertyDescriptor` base | 180 | 210 | 305 | 420 ns | **+0.600 ns/entry** |
| `getOwnPropertyDescriptor` this | ~195 | 200 | 195 | 200 ns | **+0.000** |
| `defineProperty` base → this | | 1090→1010 | 1315→1050 | 1585→1055 ns | +1.414 → +0.129 |

Control loop identical (5.0 ns) in both arms.

**Honest caveat:** with a near-empty table the index costs ≤25 ns per lookup (180 → 190–205 ns), and that figure moved 190→205 between two builds of *identical* logic, so it is at the edge of the benchmark's resolution. Crossover is below 50 prototypes. A 3-line `if forward.len() <= N { scan }` removes it; it was not shipped because it could not be re-measured on cc, and a number from a binary that isn't being shipped is not a number.

## Why it cannot go stale — three layers, none of them "I found all the writers"

#9180 records a **failed first attempt** at exactly this: a pointer-keyed cache invalidated at two of six mutation sites, producing silently wrong descriptors. So the bar was a structural argument, not an enumeration.

1. **Privacy.** `forward` and `reverse` are private to `decl_prototype_table.rs`. No other module can name them, so no other module can insert, remove or rewrite an address; every way in is a method that updates both directions in one statement sequence. A seventh writer cannot be written elsewhere.
2. **The degraded mode is main's code.** A targeted reverse update is an exact inverse only while `forward` is injective and never re-pointed. `insert` is the only place either can break, checks both in one hash lookup, and on a trip **permanently abandons the index** — after which `class_id_for` runs `scan_class_id_for`, the pre-#9180 linear scan verbatim. The fallback is not "probably still right"; it is the old answer.
3. **The index is checked, not trusted.** Under `debug_assertions` every reverse lookup is `debug_assert_eq!`'d against that scan and every mutation re-checks the whole invariant, so the runtime suite validates it continuously.

GC: both visit helpers capture the slot, run the visitor, and re-key from what it left behind — a move updates both maps or neither.

Worth noting the author shipped a first version, then found a residual hazard in it themselves (a shared prototype address would let a targeted update drop the other entry) and added layer 2 rather than arguing it was unreachable.

## The differential can fail, and fails on the exact prior bug

104 assertions — descriptor shapes for accessor/data/missing, prototype identity, `isPrototypeOf`/`instanceof`, `getOwnPropertyNames`/`keys`/`for-in`/`hasOwnProperty`, `delete` of methods and accessors, symbol keys, a `Proxy` over a class prototype, `freeze`, 40 filler prototypes — all repeated after 3,500 rounds of allocation churn and for a class first materialized *after* the churn.

- **On unchanged main: passes**, and exposes 8 pre-existing node divergences, so it discriminates.
- **On this branch: byte-identical to unmodified main**, divergence set unchanged.
- **Sabotage:** a build with the reverse lookup forced to `None` — the first attempt's exact bug — changes 21 lines, including `desc.g.present=false`, `desc.g.hasGet=false`, and `E.instance.em = em` where a `TypeError` belongs.

## Follow-ups filed separately

The sibling scan on the line immediately before this one is still linear and costs essentially what this removed; and `disable_inline_guards_for_descriptor_target` carries a comment claiming descriptor installs are "never on the hot property path", which esbuild's `__export(exports, {…})` falsifies for every bundle.

proggeramlug:
Merged.

One thing to be aware of, reported rather than buried because I could not fully account for it. On the shared validation branch, one full `perry-runtime` run failed:

```
async_hooks::test_support::tests::native_async_resource_accepts_string_and_symbol_expandos
  panicked at test_support.rs:216: assertion `left == right` failed, left: None
```

I bisected it to this PR's files — reverting them made it pass. But that comparison was not like-for-like (full suite vs single test), and when I checked properly the test passes **in isolation with this PR applied**, and the **full suite then passed twice in a row on the identical tree**. So the isolation result was an artifact of my own method, not evidence against you.

My read is a shared-global interaction rather than a defect here: `perry-runtime`'s tests run against process-global side tables and CLAUDE.md notes ~180 readers that don't take the clearing lock, symbol expando tables among them. I could not reproduce it in three subsequent runs, so I'm not holding the PR — but two clean runs is not proof, and a symbol-expando lookup returning `None` is close enough to this PR's subject (declared-prototype reverse lookup) that it deserves your eyes rather than my shrug. If it resurfaces, that test name is the thread to pull.

**Validation:** `perry-runtime` 2867 passed / 0 failed at `RUST_TEST_THREADS=1` (three runs); `perry-codegen` 31 suites / 0 failures; all 60 lint gates green; node-differential probes over prototype chains, `constructor`, `__proto__` and class hierarchies byte-identical.

Validated alongside #9213, #9216, #9219, #9224 and #9230.

Issue #9210, opened 2026-08-30
Packed-loop clone does not flush loop-carried locals on an unwind edge (regression from #9185)


proggeramlug:
`#9185` admitted `Stmt::Throw` to the packed-f64 versioned loop. That is a **silent wrong answer** on `main` today, and I am reverting the admission in the linked PR. This issue records what would be needed to re-admit it.

## The defect

`break`, `continue` and `return` leave the fast clone through normal CFG edges, and those edges flush the loop-carried locals back to their frame slots. An unwind edge does not. So any local the loop wrote reads back **stale** after the throw:

```ts
const arr: number[] = [];
for (let i = 0; i < 64; i++) arr.push(i);
const PRE = new Error("boom");

let s = 0;
try {
  for (let i = 0; i < arr.length; i++) { if (arr[i] === 40) throw PRE; s += arr[i]; }
} catch (e) { console.log(s); }   // perry: 0     node: 780
```

Measured on `84185b5656`, both orderings wrong:

| shape | node | perry | `PERRY_PACKED_LOOP_ABRUPT=0` |
|---|---|---|---|
| taken `break`, read after | 780 | 780 | 780 |
| taken `continue`, read after | 1024 | 1024 | 1024 |
| taken `throw`, accumulate after | 780 | **0** | 780 |
| taken `throw`, accumulate before | 820 | **0** | 820 |
| taken `throw`, read via closure | 780 | 780 | 780 |

The 0.5.1220 release (pre-#9185) also gives 780, confirming the regression.

## Why the existing tests passed

`#9185` shipped two taken-throw tests and both are blind to this. `throwPre` reads the thrown value and an untouched parameter; `throwValue` throws `s` itself. Both observe the clone's **live SSA value**, which is correct — never the frame slot left behind. The closure row above is correct for the same underlying reason from the other side: a captured accumulator is boxed rather than register-promoted, so there is no promoted copy to lose.

The general rule this is the second instance of: **an abrupt-exit optimisation is only tested by an exit that is actually taken AND a subsequent read of something the loop wrote.** #9154's labeled-`break` bug had the identical shape.

## To re-admit

Emit the loop-carried writeback at the throw site rather than relying on the exit block, i.e. flush promoted locals immediately before the throw is emitted. The admission predicate is not the missing piece; the flush is. Worth confirming the same edge is handled for any other construct that can leave a clone without passing through its exit block.

proggeramlug:
[...]

PR #9215, opened 2026-08-30
fix(codegen): a taken throw out of a packed loop no longer loses loop-carried locals

proggeramlug:
**This is a correctness fix for a silent wrong answer currently on `main`, from my own #9185.** Reverting the admission is the safe move; the real fix is #9210.

## The defect

#9185 admitted `Stmt::Throw` to the packed-f64 versioned loop, reasoning that the throw block ends in `unreachable`, so "control never returns to the loop and nothing reads what the clone cached". The second half is false — **an unwind lands in a `catch`, which can read anything the loop wrote.**

`break`, `continue` and `return` leave the clone through normal CFG edges that flush the loop-carried locals back to their frame slots. An unwind edge does not:

```ts
const arr: number[] = [];
for (let i = 0; i < 64; i++) arr.push(i);
const PRE = new Error("boom");

let s = 0;
try {
  for (let i = 0; i < arr.length; i++) { if (arr[i] === 40) throw PRE; s += arr[i]; }
} catch (e) { console.log(s); }   // perry: 0     node: 780
```

| shape (loop-carried local read AFTER the exit) | node | main @ `84185b5656` | `ABRUPT=0` | this PR |
|---|---|---|---|---|
| taken `break` | 780 | 780 | 780 | 780 |
| taken `continue` | 1024 | 1024 | 1024 | 1024 |
| taken `throw`, accumulate after | 780 | **0** | 780 | 780 |
| taken `throw`, accumulate before | 820 | **0** | 820 | 820 |
| taken `throw`, read via closure | 780 | 780 | 780 | 780 |

The 0.5.1220 release (pre-#9185) also gives 780, which confirms #9185 as the regression rather than something older.

## Why #9185's tests passed

Both of its taken-throw tests are blind to this by construction: `throwPre` reads the thrown value and an untouched parameter, `throwValue` throws `s` itself. Both observe the clone's **live SSA value**, which was correct — never the frame slot left behind. The closure row above is correct for the mirror-image reason: a captured accumulator is boxed rather than register-promoted, so there is no promoted copy to lose. That is what kept the defect narrow enough to slip through.

The added test reads a loop-carried local after each abrupt exit. Its `break` and `continue` rows are not padding — they are what establish that the defect is specific to the unwind edge rather than to abrupt exits in general.

The rule, which #9154's labeled-`break` bug also obeyed: **an abrupt-exit optimisation is only tested by an exit that is actually taken AND a subsequent read of something the loop wrote.** Satisfying one of those two conditions is what every test here did.

## Cost, stated plainly

This gives back #9185's win. A counted loop containing a throw returns to **16.00 ns/op against node's 1.17 (13.7× slower)** — the pre-#9185 gap, and now the worst shape in that benchmark; `break` (1.00) and the throw-free loop (1.00) both beat node (1.32, 2.26). I would rather be 13.7× slower than silently wrong, and #9210 is the way back: emit the writeback at the throw site, which is the piece #9185 was missing. The admission predicate was never the problem.

## Validation

- `packed_loop_abrupt_statements` 9/9, `loop_property_array_hoist` 12/12, `issue_8690_loop_versioned_arraylike` 3/3
- perry-hir 591/591, perry-codegen 21/21
- 9 differential programs against node, including four shapes that shadow `Error` (user class, `const` function, function-valued binding, block-scoped class) — all node-identical
- `RUSTFLAGS="-D warnings" cargo check --workspace --all-targets`: 8 errors, all in `perry-ui-macos`, all present on a pristine `origin/main` with this diff stashed
- Size: `__text` unchanged (10,968,212 B, 0.00%), `__const` +64 B, file +8 B

Refs #9151, #9185, #9210

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect results when `throw` statements occur inside optimized packed-array loops.
  * Ensured loop-carried values remain correct when execution transfers to a `catch` block.
  * Prevented unsafe fast-path execution for loops containing `throw` statements.
  * Improved reliability when thrown exceptions are handled during loop execution.

* **Tests**
  * Added coverage for preserved loop values across thrown exceptions, including access through closures.

### DIFF
--- a/crates/perry-codegen/src/stmt/loops.rs
+++ b/crates/perry-codegen/src/stmt/loops.rs
@@ -5063,9 +5063,26 @@ fn stmt_is_packed_f64_loop_safe(
                     .as_ref()
                     .is_none_or(|expr| expr_is_packed_f64_loop_safe(ctx, expr, arr_id, counter_id))
         }
-        // `throw` stays out: the thrown value is typically constructed
-        // (`throw new Error(…)`), which is a call in the loop body.
-        Stmt::Throw(_) => packed_loop_abrupt_enabled(),
+        // `throw` stays out. Admitting it (#9185) was a silent wrong answer:
+        // `break`/`continue`/`return` leave the clone through normal CFG edges
+        // that flush the loop-carried locals back to their frame slots, but an
+        // unwind edge does not, so anything reading such a local AFTER the
+        // throw saw its stale pre-loop value:
+        //
+        //     let s = 0;
+        //     try { for (let i = 0; i < arr.length; i++) {
+        //             if (arr[i] === 40) throw PRE; s += arr[i]; } }
+        //     catch (e) { return s; }   // gave 0, node gives 780
+        //
+        // #9185's tests missed it because none of them read a loop-carried
+        // local after unwinding — they read the thrown value (which IS the
+        // clone's live value, and was correct) or an untouched variable. A
+        // closure-captured accumulator was also correct, being boxed rather
+        // than register-promoted, which is what kept the bug this narrow.
+        //
+        // Re-admitting this needs the writeback emitted at the throw site, not
+        // just the admission; see #9210.
+        Stmt::Throw(_) => false,
         Stmt::LabeledBreak(_)
         | Stmt::LabeledContinue(_)
         | Stmt::While { .. }
--- a/crates/perry-ui-macos/src/file_dialog.rs
+++ b/crates/perry-ui-macos/src/file_dialog.rs
@@ -93,7 +93,7 @@ pub fn save_dialog(callback: f64, default_name_ptr: *const u8, _allowed_types_pt
 
         // Set default filename if provided
         if !default_name_ptr.is_null() {
-            let name = unsafe { str_from_header(default_name_ptr) };
+            let name = str_from_header(default_name_ptr);
             if !name.is_empty() {
                 let ns_name = NSString::from_str(&name);
                 let _: () = msg_send![&*panel, setNameFieldStringValue: &*ns_name];
--- a/crates/perry-ui-macos/src/lib_ffi/system.rs
+++ b/crates/perry-ui-macos/src/lib_ffi/system.rs
@@ -275,7 +275,7 @@ pub extern "C" fn perry_system_preferences_set(key_ptr: i64, value: f64) {
         if (bits >> 48) == 0x7FFF {
             // NaN-boxed string — extract string pointer
             let str_ptr = js_nanbox_get_pointer(value) as *const u8;
-            let s = unsafe { str_from_header(str_ptr) };
+            let s = str_from_header(str_ptr);
             let ns_str = objc2_foundation::NSString::from_str(&s);
             let _: () = objc2::msg_send![defaults, setObject: &*ns_str, forKey: &*ns_key];
         } else {
--- a/crates/perry-ui-macos/src/lib_ffi/window_misc.rs
+++ b/crates/perry-ui-macos/src/lib_ffi/window_misc.rs
@@ -408,7 +408,7 @@ pub extern "C" fn perry_ui_set_text(id_ptr: i64, value_ptr: i64) {
     let (value_data, value_len) = value
         .as_ref()
         .map_or((std::ptr::null(), 0), |value| (value.as_ptr(), value.len()));
-    unsafe {
+    {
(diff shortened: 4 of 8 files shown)


proggeramlug:
Superseded by #9230, which contains this revert as its first commit and then fixes the underlying defect — the writeback is emitted at the throw site, so the loop stays on the fast path AND is correct (7.99 → 0.95 ns/op, against node's 1.10).

Merging either one resolves the wrong answer on `main`. This PR is the smaller, more obviously-correct change if the preference is to stop the bleeding first and review the real fix separately.

proggeramlug:
Merged. **I merged #9185, so this is my miss as much as yours** — and the reason it got through is worth writing down, because my review specifically went looking for this hazard and cleared it on a wrong distinction.

I reasoned: an unwind could return control to the loop, *but* `Stmt::Try { .. } => false` rejects a `try` inside the loop body, so the handler is always outside the loop being cloned — therefore control leaves and nothing reads the cache. The gap is that "leaves the loop" and "flushes the loop-carried locals" are different things. A `catch` outside the loop is still inside the *function*, and `s` is still in scope there. Your one-line statement of it is the correct one: normal CFG edges flush, the unwind edge doesn't.

**Verified the premise rather than taking it**, because a peer running a nine-shape packed-loop battery could not reproduce it on `main` and I was not going to revert a working optimization on an unconfirmed claim. A/B on one binary, `loops.rs` swapped, runtime and stdlib static wrappers rebuilt on both arms so the coherence stamp matched:

| shape | `main` (Throw admitted) | this PR | node |
|---|---|---|---|
| accumulate after the throw | **0** | 780 | 780 |
| accumulate before the throw | **0** | 820 | 820 |
| two locals, `s + ":" + n` | **0:0** | 780:40 | 780:40 |
| nested loops, `s + hits` | **17594** | 17646 | 17646 |
| read via closure | 780 | 780 | 780 |
| taken break / continue / return | correct | correct | correct |

Two things that came out of chasing the non-reproduction, both worth having in the regression test:

- **The nested case is a *partial* flush, not a total one** — 17594 against node's 17646, short by 52. Some iterations' contributions reach the frame slot and the last ones don't. That row matters more than the zeros: `0` looks obviously broken, `17594` looks like a plausible answer. A fix validated only against the 0-valued shapes could have left this one standing.
- **Two things independently hide it**: a local captured by a closure gets boxed (the boxed path never had the bug), and a function-local array receiver takes a different tier and is correct even on the broken compiler. The peer's battery used function-local receivers throughout, which is why nine shapes came back clean against a live bug. Pairing a module-level and a function-local receiver, and a direct and a closure read, and asserting they agree, is what makes such a battery able to see this class at all.

**Validation:** `packed_loop_abrupt_statements` 9/9 including your new `a_loop_carried_local_survives_a_taken_throw`; `perry-runtime` and `perry-codegen` green; all 60 lint gates green.

Agreed that reverting the admission is right and #9210 is the real fix. Worth carrying this table into #9210 as its acceptance criteria — particularly the nested partial-flush row, since that's the one a "does the value survive" test would pass while still being wrong.

proggeramlug:
Correction to the record: **this squash carried one commit that isn't yours.**

A `git checkout` of your branch aborted on a dirty `loops.rs` in my worktree, so my HEAD was still on the validation branch when I pushed — which meant `937628231a fix(ui): drop redundant unsafe blocks in perry-ui-macos` went in alongside your fix. My mistake, not a deliberate bundling.

What landed on `main` as 8651ce0640 is therefore your throw-admission revert **plus** six redundant `unsafe` blocks removed from `perry-ui-macos` (`file_dialog.rs`, `lib_ffi/{system,window_misc}.rs`, `widgets/{alert,combobox,webview}.rs`). That second change is unrelated to packed loops: rustc reports 12 `unnecessary unsafe block` warnings across the lib and test targets there, which fails the workspace `warnings` gate under `-D warnings` on any PR that invalidates that crate's check cache. Pre-existing since #8453, and it was making the gate red for everyone.

Both changes were validated together on the branch I actually built — `packed_loop_abrupt_statements` 9/9, `perry-runtime` and `perry-codegen` green, all 60 lint gates green, and the node-differential table in my previous comment — so `main` is correct, and nothing needs reverting. But your PR number now points at a diff wider than what you wrote, and that shouldn't be discovered later by someone reading `git log`.

The `perry-ui-macos` change is mine to answer for if it turns out to be wrong.

PR #9216, opened 2026-08-30
perf(runtime): [^] and [] no longer make regex construction case-fold a million code points — cc --help −3.78%

proggeramlug:
`ClassUnicodeRange::case_fold_simple` was the **second-largest symbol** in a symbolized profile of `claude --help` at 3.78%. This removes essentially all of it: **−336,488,547 instructions, −3.78%**, wall 0.77 s → 0.74 s, output byte-identical to base and to node.

## Perry was manufacturing the pathological input itself

JS's "any character" class `[^]` was translated to `[\s\S]`, and the empty class `[]` to `[^\s\S]`. `[\s\S]` canonicalizes to `\x{0}-\x{10FFFF}`, and `regex_syntax`'s `case_fold_simple` walks **every code point in a range** — so under the `i` flag each such class ran a **1,114,112-step loop to compute a fold that cannot change anything**.

They now translate to `(?s:.)` (`Hir::dot`, never folded) and `[a&&b]`.

| isolated | before | after | |
|---|---|---|---|
| `(?i)[\s\S]*?` | 7.72 ms | 41.7 µs | 185× |
| `(?i)[^\s\S]` | 7.46 ms | 11.4 µs | 654× |

The real-world driver is five successive `new RegExp` refinements of **marked**'s HTML-block rule (`i` flag, eight `[\s\S]`, and a backreference that forces the fancy engine), plus one URL validator. A probe on `get_or_compile_regex` — verified able to fire; it saw 3,713 calls — shows `cc --help` **never matches with any of them**. The cost was pure construction.

## The hypothesis this PR started from was wrong, and was refuted two independent ways

It is worth recording, because the wrong answer was plausible and cheap to believe. The theory was that `std_engine_syntax_ok`'s `\p`/`\P` arm routes through the full HIR translator, so patterns with a Unicode property escape **and** the `i` flag pay case folding purely to validate a property name.

- **Static.** Of cc's 2,378 regex literals, 16 reach the `\p` arm, and **zero of those carry `i`**. The hypothesised combination is empty. Validating all 2,378 costs 9.2 ms total; those 16 are 0.46 ms of it.
- **Dynamic.** A uprobe on `case_fold_simple` captured a dwarf stack on **every one** of its 2,754 calls in a real `cc --help` — an exhaustive capture, not a sample. **99.57% of calls (4.63% of all retired instructions) go through `js_regexp_new` → `compile_and_cache_regex_checked` → `fancy_regex::Regex::new_options`.** 0.43% is the legitimate lazy first-use build. **Zero** pass through the `\p` validation arm.

The equivalence question the hypothesis rested on was answered anyway: across all 2,378 literals, `regex_syntax::Parser::parse` returns the same verdict with and without `(?i)` — 0 disagreements. Sound, and worthless.

The actual mechanism was found by asking who *calls* the hot symbol, exhaustively, rather than by reasoning about which code path looked expensive.

## Narrow by construction

Only a six-character class whose members are a shorthand and its complement, outside a character class. Non-capturing group, and `collect_capture_spans` indexes the input, so capture numbering is untouched. `regress` only ever sees the original JS source, so `.source` is unaffected.

## Verification

- **Differential harness, 72 cases** (valid/invalid × `i` × `\p`, bad property names, lookbehind, backreferences, every rewrite form) — **run on unchanged `main` first, where it passed**, then base vs node identical, branch vs node identical.
- **Corpus differential**: 3,402 patterns × 12 probe subjects, base vs branch **identical**.
- Unit tests mutation-checked twice: each fails when the guard it covers is broken.
- 63 existing regex tests; full 2,855-test `perry-runtime` suite; `syntax_check_agrees_with_full_build` over 12,077 patterns.
- Instruction counts: 7 interleaved reps, both binaries built from exact SHAs in one session on the same box.

An isolated fixture replaying the 59 captured eager constructions goes 574.06M → 257.99M instructions (−55.1%), and symbolized profiles of it put `case_fold_simple` at **57.64% → 10.46%** — 304M of the 316M saving is that one symbol.

## Headroom left, not taken here

The upstream O(range) fold still costs **17.5 ms** on the URL validator (`iu` flags, large but not full-space ranges); stripping `i` there drops it to 1.8 ms. A `regex-syntax` fix that walks the fold table instead of the range would recover that — the larger remaining half, and worth an upstream issue rather than another local workaround.

## Two pre-existing node-parity bugs found in passing, filed separately

The corpus differential also runs against node, and 64 of 3,402 records (1.9%) diverge on `main`, unrelated to this change: `\w`/`\W`/`\b` use Rust's Unicode word semantics where JS is ASCII-only, and perry's `.` excludes only `\n` where JS's also excludes `\r`/U+2028/U+2029.

## Summary by CodeRabbit

* **Performance**
  * Improved regular expression handling for patterns matching any character, including newline characters.
  * Reduced overhead when processing these patterns under case-insensitive matching and Unicode-related flags.

* **Bug Fixes**
  * Preserved matching behavior, capture numbering, quantifier behavior, and reported pattern sources across supported flags.

### DIFF
--- a/changelog.d/9214-regex-any-char-no-fold.md
+++ b/changelog.d/9214-regex-any-char-no-fold.md
@@ -1,0 +1,12 @@
+### Performance
+
+- **`[^]` and `[]` no longer make regex construction case-fold a million code
+  points.** Perry translated JS's "any character" class to `[\s\S]` and its
+  empty class to `[^\s\S]`. Under the `i` flag, `regex_syntax`'s
+  `case_fold_simple` walks *every code point in a range*, and `[\s\S]`
+  canonicalizes to `\x{0}-\x{10FFFF}` — so each such class ran a 1,114,112-step
+  loop to compute a fold that cannot change anything. They now translate to
+  `(?s:.)` and `[a&&b]`, which are never folded. Isolated: `(?i)[\s\S]*?`
+  7.72 ms → 41.7 µs (185×), `(?i)[^\s\S]` 7.46 ms → 11.4 µs (654×). On
+  claude-code's `--help`, −336 million instructions (**−3.78%**), with output
+  byte-identical to node.
--- a/crates/perry-runtime/src/regex/grammar.rs
+++ b/crates/perry-runtime/src/regex/grammar.rs
@@ -1420,6 +1420,84 @@ fn quantifier_after(chars: &[char], start: usize) -> Option<(usize, usize, bool)
     Some((start, end, lower_zero))
 }
 
+/// Emit the cheapest spelling of "matches any code point".
+///
+/// `[\s\S]` and `(?s:.)` denote exactly the same set — every Unicode scalar
+/// value — but they cost wildly different amounts to *compile*, and the
+/// difference only appears once the `i` flag is present.
+///
+/// `regex_syntax`'s HIR translator case-folds a character class by calling
+/// `ClassUnicodeRange::case_fold_simple` on each of its ranges, and that
+/// function loops over EVERY CODE POINT in the range:
+///
+/// ```text
+/// for cp in (start..=end).filter_map(char::from_u32) {
+///     for &cp_folded in folder.mapping(cp) { ranges.push(...) }
+/// }
+/// ```
+///
+/// `[\s\S]` canonicalizes to the single range `\x{0}-\x{10FFFF}`, so under
+/// `(?i)` that loop runs 1,114,112 times — to compute a fold that cannot
+/// change anything, since the class already contains every character. `.`
+/// under `(?s)` is not a class at all (`Hir::dot`), so the translator never
+/// folds it.
+///
+/// Measured (`fancy_regex::Regex::new`, best of 3, same machine):
+///
+/// | pattern | build |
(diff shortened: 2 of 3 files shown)


proggeramlug:
Merged.

The mechanism is convincing on its face — `[^]` and `[]` are the "every code point" and "no code point" classes, and handing `regex_syntax` a form it must case-fold across the whole Unicode range to express something with no case at all is pure waste. But a class translation is exactly where a "faster" rewrite can silently change what matches, so I probed the semantics rather than the timing.

`/[^]/.test("x")`, `/[^]+/.exec("ab")`, `/[]/.test("x")`, `"a\nb".replace(/[^]/g, ".")` and `/[^]/s.test("\n")` all match node 26.5.1 byte-for-byte. The `\n` cases are the ones I cared about: `[^]` must match a newline *without* the `s` flag — that's the whole reason people write it instead of `.` — and it does.

Validated on a shared branch with five other PRs, so if anything downstream looks odd, the co-resident changes were #9213, #9214, #9219, #9224 and #9230.

Issue #6279, opened 2026-07-11
addr_class lint has two blind spots: wrong band floors (0x10000) and lone is_valid_obj_ptr guards


proggeramlug:
## Summary

`scripts/addr_class_inventory.py` is the gate that is supposed to catch the handle-band deref class (the Linux-only SIGSEGV family: #1843, #4004, #4665, #4800, #6271). It did not catch #6271, and it would not catch most of the 44 crashing (receiver, op) pairs found in the follow-up audit. It has two blind spots.

## Blind spot 1 — it only knows the *correct* band literals

`BAND_LITERAL_RE` matches `0x100000 | 0xF0000 | 0x40000 | 0xE0000 | 0x200000`. So it flags code that hand-re-types a **correct** boundary — but a site that invents a **wrong** floor is completely invisible. Two real bugs looked exactly like this:

```rust
// object/object_ops_frozen.rs — js_object_freeze
if !obj.is_null() && (obj as usize) > 0x10000 {      // 64 KB, not 1 MB
```
```rust
// object/delete_rest.rs — js_object_delete_field
if (obj as usize) < 0x10000 {                        // 64 KB, not 1 MB
```

`HANDLE_BAND_MAX` is `0x100000`. These floors are an order of magnitude low, so the whole fetch band (`0x40000..0xE0000`) and zlib band (`0xE0000..0xF0000`) sailed straight through into the deref path. Wrong literals are *more* dangerous than re-typed correct ones, and are precisely what the lint cannot see.

## Blind spot 2 — a lone `is_valid_obj_ptr` is not flagged

#6271's actual shape:

```rust
let ns_ptr = jsval.as_pointer::<ObjectHeader>();
if !ns_ptr.is_null()
    && crate::object::is_valid_obj_ptr(ns_ptr as *const u8)   // <-- sole guard
    && (*ns_ptr).class_id == ...                              // <-- deref
```

No band literal, and no `as *const GcHeader` cast — so **neither existing rule fires**. But `addr_class.rs` says in its own doc that this predicate is not sufficient:

> the platform `HEAP_MIN` floor on Linux/Android/iOS/Windows (`0x1000`) is BELOW the handle band, so this predicate alone does NOT reject small handles there — pair it with `is_handle_band`

A rule for "`is_valid_obj_ptr` without a handle-band predicate in the same guard" would have caught it.

## Why I did not just add the rules

Both candidate rules are far too noisy to ship as a gate today:

[...]

PR #9219, opened 2026-08-30
fix(runtime): #9192 — an array with a non-array [[Prototype]] inherited nothing (47 divergences from node → 19)

### LINKED ISSUE #9192: Object.setPrototypeOf(array, non-array object) — the array inherits nothing (silent wrong values)
Found while investigating the process-global array-index latch (which turned out to be a non-issue for the workload — see the perf note at the bottom). This is a **silent wrong-value** bug, not a crash.

## Repro

```js
const a = [1, 2, 3];
Object.setPrototypeOf(a, { 7: "inherited", foo: "bar" });

a[7]      // node: "inherited"   perry: undefined
a.foo     // node: "bar"         perry: undefined
7 in a    // node: true          perry: false
```

The classic ES5 array-subclass idiom is broken the same way:

```js
function MyList() {}
MyList.prototype = Object.create(Array.prototype);
MyList.prototype.tag = "mylist";
MyList.prototype.first = function () { return this[0]; };

const arr = [10, 20];
Object.setPrototypeOf(arr, MyList.prototype);

arr.tag        // node: "mylist"   perry: undefined
arr.first()    // node: 10         perry: TypeError / missing
```

## Root cause

`array_custom_array_prototype` (`crates/perry-runtime/src/array/indexing.rs:237-268`) accepts the recorded `[[Prototype]]` **only when that prototype is itself `GC_TYPE_ARRAY`**. Retargeting an array's prototype to an ordinary object records the change (the process-wide deopt latch is set, so perry pays the cost) but the lookup path then declines to consult it — so the array inherits nothing.

perry therefore pays the deoptimisation for a case it does not implement.

## Why it was never caught

The only fixture covering array prototype retargeting, `test-files/test_gap_typed_arrays.ts:38`, uses an **array** as the new prototype — the one shape that works. The object-prototype case has no coverage.

## Suggested fixture

A test that sets an array's prototype to (a) a plain object with an indexed key and a named key, (b) an `Object.create(Array.prototype)`-derived prototype carrying methods, and (c) `null`, checking element reads, named reads, `in`, and method calls against node in each case.

---

*Perf context, for completeness:* the same investigation established that `class X extends Array` does **not** latch `PERRY_ARRAY_INDEX_FAST_PATH_INVALIDATED` on current main (subclass instances have been elements-backed `GC_TYPE_OBJECT` since #8974), that the claude-code and pi bundles never latch it, and that forcing it on for `cc --help` changes instructions by 0.001%. The latch is a genuine latent risk when it does fire (measured up to 14× on `push`, process-wide and permanent), but it is not a workload problem today.

proggeramlug:
Fixes #9192. Differential vs `node --experimental-strip-types`, 136 checks: **47 divergences before, 19 after, 0 regressions.**

## Root cause

Perry has no real exotic-object prototype chain for arrays — an array's chain is **implicit and hardcoded** as `Array.prototype → Object.prototype`. `Object.setPrototypeOf(arr, p)` *does* record `p` and *does* latch the process-wide `PERRY_ARRAY_INDEX_FAST_PATH_INVALIDATED`, but nothing then reads the record when `p` is not an array. So perry paid the full global deopt for a case it did not implement.

The bug is **symmetric, and both halves are silent**: the array inherits nothing from its new prototype, **and keeps inheriting everything from the old one** — `typeof a.map` stayed `"function"` where node says `"undefined"`.

The issue named only the first of **two** hardcodes. `array_custom_array_prototype` (indexed lookups) returned `None` unless the recorded prototype was itself `GC_TYPE_ARRAY`. The second is `array_prototype_property_value` (`field_get_set/accessors.rs`), which hardcodes `Array.prototype` and is the single fallback for the array arms of `js_object_get_field_by_name`, `js_object_has_property`, and via them `Reflect.get`/`has` and the for-in re-check. That second hardcode is what made the named-property, `constructor`, `in`-by-name and "methods should be gone" cases wrong.

**And it means named properties never resolved even from an *array* prototype** — the shape #9192 calls "the one that already works" only worked for indices. `i2.tagged` returned `undefined`. That is now fixed and pinned.

## What changed

7 files, +377/−38, all behind the existing recorded-prototype gate.

- **`array/indexing.rs`** — one classification, `ArrayCustomProto::{Null, Array, Other}`, replaces the array-only probe and drives `array_spec_get`, `array_spec_has_index`, `array_oob_prototype_get` and `array_spec_set`. The `Array` lane is **bit-for-bit the old code** (test262 `copyWithin/coerced-values-start-change-*`). `Other` resolves through `resolve_inherited_field_from_prototype` with the array bound as receiver, so prototype accessors see the right `this` and further hops are walked. `Null` inherits nothing and suppresses the implicit `Array.prototype`/`Object.prototype` tail. **Proxy prototypes deliberately keep their dedicated handling** — routing them here too would invoke the `has` trap twice, which is observable.
- **`field_get_set/accessors.rs`** — `array_prototype_property_value` consults the recorded prototype before its hardcoded `Array.prototype`.
- **`field_get_set/has_property.rs`** — new `prototype_value_has_property`: the `[[HasProperty]]` an `ArrayHeader` receiver cannot reach through `ordinary_has_property`.
- **`get_field_by_name_tail.rs`** — `arr.__proto__` is the array's `[[Prototype]]`; `arr.constructor` resolves through a recorded chain instead of short-circuiting to global `Array`.
- **`native_call_method/handle_methods.rs`** — `arr.first()` dispatches through the recorded chain.
- **`symbol/get.rs`** — the explicit-prototype symbol walk accepts an array receiver (address-only; chain hops still require a real `GC_TYPE_OBJECT`).

**Cost for a default-prototype array:** one extra `object_static_prototype` probe (an `Acquire` load on an empty-registry latch) on the **named-property-miss** path, plus a byte compare for `__proto__`. The hot element path is *cheaper* — a duplicate prototype probe was deleted from `js_array_get_f64`'s hole branch, which `array_oob_prototype_get` already did.

## The 19 residuals, scoped by control

Controls run each residual with an **array** prototype as well as an object one, so "pre-existing" is measured rather than asserted.

**14 of 19 are pre-existing and not #9192** — they reproduce identically with an array prototype, and are filed separately:
- `arr[i] = v` never observes an inherited index accessor: the strict store's fast lanes bypass `array_spec_set` entirely.
- `Array.prototype.{join,indexOf,map,forEach}.call(arr)` does not fill holes through any custom prototype.

**5 are #9192-family and remain**, both "perry does more" rather than wrong values: `Symbol.iterator` is still *present* on a retargeted array (symbol *inheritance* is fixed; suppressing the built-in iterator when `Array.prototype` leaves the chain is not), and `for…in` omits the prototype's own **index** key — which control C3b traces to key ordering on the prototype object, not to arrays.

## The fixture can fail

`test-files/test_gap_9192_array_object_prototype.ts`. Built the compiler from the pristine-main tree and compiled the fixture with it: **14 of its 36 output lines diverge from node** — e.g. `A elem: undefined undefined 3`, `B inherited: undefined [object Object] undefined false`, `I array proto: protoSix true undefined 2 function`. With the fix it is byte-identical to node, in both the auto-optimize and `--no-auto-optimize` configurations.

## Tests

- `cargo test --release -p perry-runtime --lib -- --test-threads=1`: **2852 passed, 0 failed**, 4 ignored.
- Gap suite (603 tests): **592 pass, 0 compile failures, 0 crashes.** All array/prototype fixtures pass, including `test_gap_typed_arrays` and `test_gap_array_proto_grow_hole_read`. Of 11 parity_fail: 5 are pre-existing `gap_snapshot.json` entries, 5 are node-side `ERR_MODULE_NOT_FOUND` in a fresh worktree with no `node_modules`, and 1 is byte-identical to node when re-run standalone (a suite flake under load).
- `rustfmt --check` clean; `scripts/check_test_registration.py` green.

## On the latch

Untouched, and this fix does not let it stay unlatched — the recorded prototype must still be consulted per read. What changed is that **the deopt now buys correct semantics instead of nothing**, and the per-array condition is genuinely per-array, so the separate free-bit-13 idea is unaffected and better justified.

## Summary by CodeRabbit

* **Bug Fixes**
  * Arrays now correctly inherit indexed, named, symbol, accessor, and method properties after their prototype is changed.
  * Prototype operations involving `Object.setPrototypeOf`, `__proto__`, and `Reflect.setPrototypeOf` now behave consistently.
  * Custom `null`, object, array, and restored default prototypes are handled correctly, including array holes and constructor lookup.

* **Tests**
  * Added coverage for custom array prototypes, inheritance chains, accessors, and subclass-style methods.

### DIFF
--- a/changelog.d/9192-array-object-prototype.md
+++ b/changelog.d/9192-array-object-prototype.md
@@ -1,0 +1,19 @@
+### fix(runtime): an array retargeted to a non-array prototype inherits from it
+
+`Object.setPrototypeOf(arr, someObject)` recorded the new `[[Prototype]]` —
+and paid the process-wide array-index deoptimisation for it — but the lookup
+paths then declined to consult it: the index probe accepted a recorded
+prototype only when the prototype was itself an array, and the named-property
+fallback hardcoded `Array.prototype`. A retargeted array therefore inherited
+nothing from its new prototype while still inheriting everything from the old
+one, with no error: `a[7]` and `a.foo` were `undefined`, `7 in a` was `false`,
+and `typeof a.map` was still `"function"`.
+
+Index reads/`in`/writes, named reads, `in` on a name, `arr.constructor`,
+`arr.__proto__`, symbol-keyed reads, and method dispatch (`arr.first()`, the
+ES5 `MyList.prototype = Object.create(Array.prototype)` idiom) now all resolve
+through the recorded chain, with the array bound as the receiver so a
+prototype accessor observes the right `this`. `Object.setPrototypeOf(arr,
+null)` correctly inherits nothing at all. Named properties on an *array*
+prototype, which never resolved either, are fixed by the same change. Arrays
+with the default prototype are untouched. Fixes #9192.
--- a/crates/perry-runtime/src/array/indexing.rs
+++ b/crates/perry-runtime/src/array/indexing.rs
@@ -85,13 +85,23 @@ pub(crate) fn note_array_index_write(arr: usize) {
 #[inline]
 unsafe fn array_oob_prototype_get(receiver: usize, index: u32) -> f64 {
     const TAG_UNDEFINED_F64: f64 = f64::from_bits(0x7FFC_0000_0000_0001u64);
-    // A custom array [[Prototype]] (Object.setPrototypeOf(arr, otherArray))
-    // replaces the default chain — gated on a global relaxed flag.
+    // A custom [[Prototype]] (`Object.setPrototypeOf(arr, p)`) replaces the
+    // default chain — gated on a global relaxed flag. #9192: `p` need not be an
+    // array; a plain object / `Object.create(Array.prototype)` result answers
+    // the whole lookup through the generic resolver.
     if crate::object::prototype_chain::array_static_proto_recorded() {
-        if let Some(proto_arr) = array_custom_array_prototype(receiver as *const ArrayHeader) {
-            if index < (*proto_arr).length && array_has_own_index(proto_arr, index) {
-                return js_array_get_f64(proto_arr, index);
+        let arr = receiver as *const ArrayHeader;
+        match array_custom_prototype(arr) {
+            Some(ArrayCustomProto::Null) => return TAG_UNDEFINED_F64,
+            Some(ArrayCustomProto::Other(bits)) => {
+                return array_object_proto_index_get(arr, bits, index).unwrap_or(TAG_UNDEFINED_F64)
             }
+            Some(ArrayCustomProto::Array(proto_arr)) => {
+                if index < (*proto_arr).length && array_has_own_index(proto_arr, index) {
+                    return js_array_get_f64(proto_arr, index);
+                }
+            }
+            None => {}
         }
     }
     if ARRAY_PROTO_HAS_INDEX.load(Ordering::Relaxed) {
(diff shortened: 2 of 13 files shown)


proggeramlug:
Merged, with a correctness fix pushed onto the branch.

47 → 19 divergences is a big win and the diagnosis is right: an array with a retargeted `[[Prototype]]` was answering `__proto__` and `constructor` off the implicit array chain while `Object.getPrototypeOf` answered off the side table, so the two disagreed about the same object.

**The defect I fixed is an address-class one, and it's the kind that reads as fine.** The new raw-pointer branch was:

```rust
} else if top16 == 0 && proto_bits > 0x10000 {
    proto_bits as usize
```

`HANDLE_BAND_MAX` is `0x100000` — an order of magnitude above that floor. So values in `(0x10000, 0x100000]` are handle-band values, and this branch accepted them as raw object pointers and handed them straight to a dereference. `0x10000` looks like a plausible "not a small integer" floor, which is exactly why the ratchet exists. Both new `is_valid_obj_ptr` guards were also unpaired, which #6279 flags for the same reason: that predicate alone cannot separate a handle from a pointer.

`addr_class_inventory.py` caught all three (`lone-valid-obj-ptr` ×2, `handle-floor` ×1). I replaced the literal floor with `crate::value::addr_class::is_above_handle_band` and paired both guards with it. Worth noting the gate found this and reading the code did not — I'd read that branch twice before the ratchet told me the constant was wrong.

**Also split two files** that this PR pushed over the 2000-line cap: `array/indexing.rs` (2049) and `get_field_by_name_tail.rs` (2012). Your `__proto__`/`constructor` logic moved into `object/field_get_set/array_retargeted_proto.rs` as two named helpers, which reads better than inline arms anyway; the array-side split moved two test modules and two test-only helpers out, with explicit named re-exports so the existing `super::indexing::…` paths keep resolving (a glob would not propagate).

**Validation:** differential probe against node 26.5.1 — `Object.setPrototypeOf(arr, base)` then `greet()`, `tag`, `length`, indexing, `Array.isArray`, `getPrototypeOf`, `__proto__`, `constructor`, `typeof map`/`push`; a null-prototype array; and restoring `Array.prototype` and confirming `map`, `constructor` and `__proto__` all come back. Byte-identical. `perry-runtime` 2867 passed / 0 failed at `RUST_TEST_THREADS=1`; `perry-codegen` 31 suites / 0 failures; all 60 lint gates green.

Validated alongside #9213, #9214, #9216, #9224 and #9230.

Issue #6696, opened 2026-07-19
symbols: generator objects lack computed [Symbol.iterator] — es2015 yield* lowering still crashes (follow-up #6676)


proggeramlug:
Follow-up to #6676 (fixed by `17284676c` — "resolve computed Symbol[key] well-known symbols"). Verified on `origin/main` @ `4d25061da` with the composition-stress harness (`secret-tests/diffharness`): the root read is fixed, but the end-to-end case #6676 was filed for still fails, because of a distinct residual.

## Symptom
A **generator** object does not expose `[Symbol.iterator]` via **computed** property access.

```ts
function* g() { yield 1; yield 2; }
const it: any = g();
console.log(typeof it[Symbol.iterator]);        // node: "function"   perry: "undefined"
console.log(it[Symbol.iterator]?.() === it);    // node: true         perry: false
```

`Symbol["iterator"]` now correctly resolves to the well-known symbol (thanks to #6676), and a plain object with a computed `[Symbol.iterator]` iterates fine — but a **generator** returns `undefined` for that property, so anything that reads the method off a generator via the symbol key breaks.

## Why it still matters (the original #6676 e2e)
esbuild's `__yieldStar` helper (emitted for `yield*` at `--target=es2015|es2017`) does `obj = value[Symbol.iterator]()` where `value` is the delegate **generator**. Since that method reads `undefined`, the helper now dies one step later:

```ts
// findings/A_yieldstar_min.es2015.mjs  (esbuild --bundle --target=es2015)
function* inner(){ yield 1; yield 2; }
function* outer(){ yield* inner(); }
console.log(JSON.stringify([...outer()]));
```
```
node : [1,2]
perry: TypeError: Cannot use 'in' operator to search for 'throw' in undefined   (empty stdout, exit 1)
```
Matrix: `s04_generators` still DIVERGES at all 8 es2015/es2017 cells; es2020+ (native `yield*`) is clean.

## Suspected subsystem
Generator/iterator prototype: generator instances need `[Symbol.iterator]` (returning `this`) reachable through computed property access, as `%GeneratorPrototype%[Symbol.iterator]` is in Node.

Repro: `findings/A_residual_generator_symbol_iterator.ts`. Severity: **high** (blocks all es2015/es2017-lowered `yield*`).

Issue #6676, opened 2026-07-19
symbols: computed Symbol["iterator"] returns undefined — breaks es2015 yield*/for-of/spread lowering


proggeramlug:
Found by the composition-stress differential harness (`secret-tests/diffharness`), which multiplies real programs through bundler downleveling and diffs vs Node v26.3.0. perry `target/release/perry` @ 2026-07-15.

## Symptom
Reading a well-known symbol through a **computed** key returns `undefined`. Dot access returns the real symbol.

```ts
console.log(Symbol["iterator"] === Symbol.iterator, typeof Symbol["iterator"]);
```
```
node : true symbol
perry: false undefined      <-- BUG
```

Because the object you key with that `undefined` is never registered under perry's real `Symbol.iterator`, `getIterator` later returns `undefined`.

## Why it matters (real-bundle manifestation)
esbuild lowering `yield*` / `async` for `--target=es2015|es2017` emits `__knownSymbol("iterator")` = `Symbol["iterator"]`. That helper (`__yieldStar`) keys its delegate iterator under `undefined`, so driving it dies:

```ts
// findings/A_yieldstar_min.ts, bundled with: esbuild --bundle --format=esm --target=es2015
function* inner(){ yield 1; yield 2; }
function* outer(){ yield* inner(); }
console.log(JSON.stringify([...outer()]));
```
```
node : [1,2]
perry: TypeError: Cannot read properties of undefined (reading 'next')   (empty stdout, exit 1)
```

Matrix signature: this seed is IDENTICAL at es2020/es2022/esnext (native `yield*`) and DIVERGES at es2015 + es2017 (all cjs/esm, min/raw) — i.e. exactly where the `__knownSymbol` helper appears.

Related manifestation (same family, folded in): a *dynamically-computed* key `class C { *[(x=>x)(Symbol.iterator)]() {…} }` also fails to register as the iterator — perry appears to only recognise a **statically literal** `[Symbol.iterator]` key.

## Suspected subsystem
Well-known-symbol table exposure on the `Symbol` constructor for computed (non-literal) property access; and/or the assumption that iterator-protocol keys are literal `Symbol.iterator` at the use site.

## Repro
```
perry /tmp/a.ts -o /tmp/a && /tmp/a       # a.ts = the 3-line snippet above
```
Related closed context: #1840, #6454. Severity: **high** — breaks every generator/spread/for-of/destructuring composition once downleveled to es2015/es2017 (a common bundler target).

PR #9223, opened 2026-08-30
docs(skills): port-npm-to-perry stops labeling supported constructs unsupported

### LINKED ISSUE #9112: port-npm-to-perry: scan lists supported constructs (Symbol/WeakMap/Proxy) as unsupported
The `port-npm-to-perry` scan labels constructs as unsupported that Perry now supports. It looks out of date with the compiler, and its triage rules out portable packages.

## What's wrong, and the evidence

The scan groups these under a `# Unsupported primitives` header:

```
grep -rn "\bSymbol(" <pkg>/
grep -rn "new Proxy(" <pkg>/
grep -rn "new WeakMap(" <pkg>/
grep -rn "new WeakRef(" <pkg>/
```

Surfacing these constructs is fine. Labeling them "unsupported" is wrong:

- `Symbol` is supported. It has no section in [`limitations.md`](https://github.com/PerryTS/perry/blob/main/docs/src/language/limitations.md), and computed `Symbol.iterator` / `yield*` lowering is fixed (#6676, #6696).
- `WeakMap` and `WeakRef` are supported. See [Weak References Retain Their Targets](https://github.com/PerryTS/perry/blob/main/docs/src/language/limitations.md#weak-references-retain-their-targets). Targets are retained rather than collected, which is the only caveat.
- `Proxy` is supported with limited trapping. See [Limited Proxy Trapping](https://github.com/PerryTS/perry/blob/main/docs/src/language/limitations.md#limited-proxy-trapping).

The triage step also says to rule out "packages whose core API depends on Proxy." Proxy is supported, so this rule rules out portable packages on its own.

## Fixes

- Rename the group so it doesn't assert these are unsupported. Move `Symbol` out (it has no caveat). Put `WeakMap`, `WeakRef`, and `Proxy` under "supported with caveats," linked to the limitations sections.
- Narrow the Proxy triage rule to packages needing exotic trap behavior, not any Proxy use.

steinybot:
# Problem

The `port-npm-to-perry` skill's gap scan groups `Symbol`, `Proxy`, `WeakMap`, and `WeakRef` under an `# Unsupported primitives` header, and its triage step rules out any package "whose core API depends on `Proxy`." All four constructs are supported, so the scan mislabels them and the triage rule rules out portable packages on its own.

# Solution

Stop asserting these are unsupported. Surface them for review with their real caveats, linked to `docs/src/language/limitations.md`, and scope the Proxy triage rule to the one case that actually rules a package out.

# This PR

Edits only `.claude/skills/port-npm-to-perry/SKILL.md` (docs/tooling, no `crates/` change, so no changelog fragment).

## Changes

1. **Rename the scan group.** `# Unsupported primitives` → "Supported, but with caveats — surface for review, not blockers." Drop the `Symbol` grep entirely — `Symbol` is fully supported with no caveat ([no section in limitations.md](https://github.com/PerryTS/perry/blob/main/docs/src/language/limitations.md); computed `Symbol.iterator` / `yield*` lowering fixed in #6676, #6696). Keep `Proxy`/`WeakMap`/`WeakRef`, each linked to its limitation:
   - [Limited Proxy Trapping](https://github.com/PerryTS/perry/blob/main/docs/src/language/limitations.md#limited-proxy-trapping)
   - [Weak References Retain Their Targets](https://github.com/PerryTS/perry/blob/main/docs/src/language/limitations.md#weak-references-retain-their-targets)
2. **Narrow the Proxy triage rule.** Rules out only packages needing exotic trap behavior beyond Perry's supported surface, not any `Proxy` use. Adds "Plain `Proxy` use is fine."
3. **Drop stale Symbol advice.** Removed the "swap `Symbol` for string sentinels" patchable tip and the `Symbol`-in-a-string false-positive example, both left over from treating `Symbol` as unsupported.

Fixes #9112

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01U8NeyZTthezLR2dMUxqpKQ

## Summary by CodeRabbit

* **Documentation**
  * Updated package compatibility guidance for `Symbol`, `Proxy`, `WeakMap`, and `WeakRef`.
  * Clarified that standard `Proxy` usage is supported, while advanced trap behavior may remain incompatible.
  * Removed outdated guidance recommending string replacements for `Symbol` usage.

### DIFF
--- a/.claude/skills/port-npm-to-perry/SKILL.md
+++ b/.claude/skills/port-npm-to-perry/SKILL.md
@@ -38,11 +38,10 @@ grep -rn "new Function(" <pkg>/
 grep -rn "require(" <pkg>/ | grep -v "^.*://"   # dynamic require
 grep -rn "await import(" <pkg>/
 
-# Unsupported primitives
-grep -rn "\bSymbol(" <pkg>/
-grep -rn "new Proxy(" <pkg>/
-grep -rn "new WeakMap(" <pkg>/
-grep -rn "new WeakRef(" <pkg>/
+# Supported, but with caveats — surface for review, not blockers
+grep -rn "new Proxy(" <pkg>/    # limited trapping: docs/src/language/limitations.md#limited-proxy-trapping
+grep -rn "new WeakMap(" <pkg>/  # targets retained, never collected: docs/src/language/limitations.md#weak-references-retain-their-targets
+grep -rn "new WeakRef(" <pkg>/  # same retention caveat
 
 # Reflection / metadata
 grep -rn "Reflect\." <pkg>/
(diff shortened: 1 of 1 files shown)


proggeramlug:
Merged.

Checked the claims rather than the prose, since a porting guide that says "supported" when it isn't sends people down a road that dead-ends later — the expensive direction for this particular document.

- Both referenced anchors resolve: `## Limited Proxy Trapping` (limitations.md:219) and `## Weak References Retain Their Targets` (:193).
- `Symbol` is genuinely supported — there's a real registry behind `is_registered_symbol`, and `Symbol.iterator` dispatch goes through it. Two PRs merged today (#9176, #9177) were optimizing that probe, so listing it under "unsupported primitives" was well out of date.
- `Proxy` has a full module tree (`proxy/{apply_construct,has_delete,invariants,json,own_keys,prototype,put_value,reflect}.rs`), so "plain Proxy use is fine, exotic trap behavior is the limit" matches the implementation.
- `WeakMap`/`WeakRef` exist in `weakref.rs`, and "targets retained, never collected" is the honest framing — they work, they just don't collect.

The triage-section rewording is the more valuable half: moving `Proxy` from "rules the package out" to "rules it out only if it needs engine-level trapping of every dynamic access" is the difference between abandoning an ORM and porting it. Dropping the `Symbol` → string-sentinel workaround from the patchable list is right for the same reason — that advice would now make a package worse.

PR #9224, opened 2026-08-30
perf(json): validate DirectParser in one pass

### LINKED ISSUE #9184: JSON.parse scans every document twice, and the second scan can't be removed: DirectParser accepts truncated/malformed input
`js_json_parse` runs a full `serde_json::from_slice::<IgnoredAny>` validation scan over the entire input **before** perry's own `DirectParser`/tape builds the value (`crates/perry-runtime/src/json/parse_api.rs`, in both `js_json_parse` and `js_json_parse_result`). Every document is therefore walked twice.

That looks like an obvious thing to delete. **It is not** — I tried, and the reason it must stay is the actual finding here.

## `DirectParser` is not a validator

With the pre-pass disabled in `js_json_parse`, perry accepts input Node rejects:

```
JSON.parse('{')          ->  {}            (node: SyntaxError)
JSON.parse('[')          ->  [null]        (node: SyntaxError)
JSON.parse('[,]')        ->  [null,null]   (node: SyntaxError)
JSON.parse('[1,]')       ->  [1,null]      (node: SyntaxError)
JSON.parse('{"a":1,}')   ->  {"a":1}       (node: SyntaxError)
JSON.parse('[01]')       ->  [1]           (node: SyntaxError)
JSON.parse('[1.]')       ->  [1]           (node: SyntaxError)
JSON.parse('[1e]')       ->  [0]           (node: SyntaxError)
JSON.parse('{"a":}')     ->  {"a":null}    (node: SyntaxError)
JSON.parse('[1,,2]')     ->  [1,null,2]    (node: SyntaxError)
```

Ten of a 64-case battery. The pattern is consistent: truncation and missing elements yield `null` rather than an error, and malformed numbers (`01`, `1.`, `1e`) are silently coerced. **Nothing is broken today** — the pre-pass catches all of these, and with it in place perry is byte-identical to Node on all 64 cases. This is a latent hazard, not a live bug.

## Why it matters

1. **It is a trap for exactly this optimization.** The double scan is visible in any profile and reads as pure waste; deleting it silently turns malformed JSON into plausible-looking values. Whoever tries next deserves to find this issue rather than the divergence.
2. **The cost is real but modest.** Removing the `js_json_parse` pre-pass takes `benchmarks/json_polyglot/bench_field_access.ts` from ~600 ms to ~565 ms on the Linux box (~6%), against Node's ~245. So the double scan is *not* where the 2.4× gap lives — the parse band is ~23% of profile, the shape-table probes ~10%, stringify ~13%.
3. **The route to claiming it is to make `DirectParser` strict**, then drop the pre-pass — worth roughly 6% and one full pass over every document parsed, at the cost of teaching the parser to reject the ten shapes above (and whatever else a wider battery finds).

## Repro

The 64-case battery is a single file — 36 inputs that must throw `SyntaxError` (trailing commas, unquoted keys, `NaN`/`Infinity`, bad escapes, two values in a row, leading zeros) and 28 that must parse (deep nesting, duplicate keys, surrogate pairs, `1e-400`, integers past 2^53). It compares thrown-vs-parsed and the error constructor name against Node, so it is engine-agnostic. Happy to attach it to a PR when someone takes the strict-parser work; it lives in my scratch as `jsonsem.ts`.

Method note: measure this with a normal build. My first "58 ms saved" was an artifact of comparing a `--debug-symbols` binary against a normal one.

proggeramlug:
Closes #9184

## Summary

- make DirectParser validate structure, literals, number grammar, escapes, control bytes, and complete input while it materializes Perry values
- remove the `serde_json` validation scans from both JSON parse entry points and enforce the same completion check in the typed-array fast path
- preserve valid lone-surrogate JSON strings by keeping WTF-8 values out of the UTF-8-only short-string representation
- add parser-level coverage plus a Node-parity integration battery with 43 malformed inputs, 28 valid value cases, and typed-array parsing

## Testing

- `cargo test -p perry-runtime json::tests --lib` (34 passed)
- `cargo test -p perry --test issue_9184_json_parse_strict -- --nocapture`
- globals Node differential: 115/122 versus checked-in 111/115 baseline; all three JSON.parse/reviver fixtures pass individually
- full perry-runtime library sweep: 2,831 passed, 4 ignored; two unrelated host/global-state-sensitive failures both passed when rerun individually
- `git diff --check`

No version bump.

## Summary by CodeRabbit

* **Bug Fixes**
  * `JSON.parse` now enforces strict JSON syntax, including valid numbers, escapes, control characters, nesting, and trailing content.
  * Malformed JSON consistently raises a `SyntaxError` instead of being partially accepted or converted to an incorrect value.
  * Unicode surrogate pairs and lone surrogates are handled correctly.

* **Tests**
  * Added comprehensive regression coverage for valid, invalid, nested, duplicate-key, numeric, and whitespace inputs.

* **Documentation**
  * Added a changelog entry for single-pass JSON parsing and validation.

### DIFF
--- a/changelog.d/9224-json-parse-single-pass.md
+++ b/changelog.d/9224-json-parse-single-pass.md
@@ -1,0 +1,4 @@
+### Performance
+
+- `JSON.parse` now validates and constructs values in one strict parser pass,
+  eliminating the preliminary full-document validation scan.
--- a/crates/perry-runtime/src/json/mod.rs
+++ b/crates/perry-runtime/src/json/mod.rs
@@ -966,8 +966,118 @@ mod tests {
         }
     }
 
+    fn direct_parser_accepts(input: &[u8]) -> bool {
+        let saved_roots = parse_root_save_len();
+        let accepted = {
+            let _suppress = crate::gc::GcSuppressScope::new();
+            let mut parser = DirectParser::new(input);
+            unsafe {
+                parser.parse_value();
+            }
+            let accepted = parser.finish();
+            parse_root_restore(saved_roots);
+            accepted
+        };
+        accepted
+    }
+
     #[test]
-    fn parse_result_streaming_validation_rejects_malformed_and_trailing_input() {
+    fn direct_parser_validates_json_while_building_the_value() {
+        let invalid: &[&[u8]] = &[
+            b"{",
+            b"}",
+            b"[",
+            b"]",
+            b"",
+            b"   ",
+            br#"{,}"#,
+            br#"[,]"#,
+            br#"[1,]"#,
+            br#"{"a":1,}"#,
+            br#"{a:1}"#,
+            br#"{'a':1}"#,
+            br#"[01]"#,
+            br#"[-01]"#,
+            br#"[1.]"#,
+            br#"[.5]"#,
+            br#"[+1]"#,
+            br#"[1e]"#,
+            br#"[1e+]"#,
+            br#"[--1]"#,
+            br#"[NaN]"#,
+            br#"[Infinity]"#,
+            br#"[-Infinity]"#,
+            br#"[undefined]"#,
+            br#"[TRUE]"#,
+            br#""unterminated"#,
+            br#"["bad\x"]"#,
+            br#"["\u12"]"#,
+            br#"["\uZZZZ"]"#,
+            br#"{"a" 1}"#,
+            br#"{"a":}"#,
+            br#"{:1}"#,
+            br#"[1 2]"#,
+            br#"[1][2]"#,
+            br#"{}{}"#,
+            b"nul",
+            b"tru",
+            br#"[1,,2]"#,
+            br#"{"a":1 "b":2}"#,
+            br#""\t"x"#,
+            b"\"\t\"",
+            b"\"abcdefgh\nijklmnopqrst\"",
+            b"\"abcdefghijklmnop\nqrst\"",
+            b"\x0bnull",
+        ];
+        for input in invalid {
+            assert!(
+                !direct_parser_accepts(input),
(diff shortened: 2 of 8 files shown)


proggeramlug:
Merged, with one gate fix pushed onto the branch.

The design question here is the one #9184 raises: the second scan can't just be deleted, because `DirectParser` accepts `{`, `[1,]`, `[01]`, `[1e]` and several more that are not valid JSON. Folding validation into the single pass is the right answer to that — it removes the duplicate walk without removing the rejection. I checked the rejections survive rather than assuming: `{`, `[1,]`, `[01]`, `[1e]`, `{"a":}` and `tru` all still throw `SyntaxError`, matching node 26.5.1 exactly, and a round-trip over an object with nested arrays, `null`, `true`, a large float and a non-ASCII string (`"sürf"`) is byte-identical.

**What I fixed:** the post-collection re-derivation open-coded the `StringHeader` payload offset, which raised `string_payload_access_inventory` from its 363 baseline to 364. That ratchet is decrease-only, so it would have gone red on `main`.

Rather than neutralise the increase I converted all four sites in `parse_api.rs` to `crate::string::string_data` — 364 → 360 — and lowered the baseline in the same change, which is what the ratchet's own contract asks for ("a decrease must lower the baseline in the same change"). Net effect is that the file pays down three sites of pre-existing debt instead of adding one.

Worth saying the re-derivation itself is *right*: reading the payload again after `gc_check_trigger` rather than holding a slice across it is exactly the discipline a copying collector needs — a borrowed heap slice is the one shape rooting cannot fix. The only issue was spelling it by hand.

Validated alongside #9213, #9214, #9216, #9219 and #9230 on one branch.

Issue #9202, opened 2026-08-30
node-suite parity audit on macOS arm64: 577 non-PASS results, including 24 crashes


TheHypnoo:
## What happened

A complete local run of Perry's canonical Node compatibility suite on an Apple M1 Pro completed all **4,376** fixtures and produced **577 non-PASS results**, including **24 crashes**.

The command exited successfully because the aggregate parity rate remained above the current 80% threshold, even though the run contained 9 SIGSEGVs, 15 timeouts, and 7 compile failures. This issue records the full reproducible baseline so the failures can be triaged by causal cluster rather than as isolated fixtures.

## What you expected

The complete `node-suite` run should provide a trustworthy compatibility gate:

- every selected fixture is adjudicated against the pinned Node oracle;
- crashes, timeouts, compile failures, and oracle failures are visible and fail the gate independently of the aggregate percentage;
- the committed baseline and the canonical harness use the same execution contract;
- failures can be tracked module-by-module from a reproducible report tied to a specific commit and environment.

## Minimal reproduction

This is intentionally the full suite, not a minimized single-runtime bug:

```bash
env \
  -u PERRY_SKIP_BUILD \
  -u PERRY_BIN \
  -u PERRY_RUNTIME_DIR \
  -u PERRY_NO_AUTO_OPTIMIZE \
  PATH="$PWD/.cache/node-pin/26.5.1/node-v26.5.1-darwin-arm64/bin:$PATH" \
  ./run_parity_tests.sh \
    --suite node-suite \
    --journal /tmp/perry-node-suite-fresh.jsonl
```

No `--module`, `--filter`, `--shard`, or `--resume` option was used. The script performed a normal fresh build before running the suite.

## Environment

- **Perry version**: `0.5.1519`
- **Commit**: `4a60b8ca1d2cfb9642a24604b0cec18c74863cac`
- **Node oracle**: `v26.5.1` from the repository's pinned local toolchain
- **Host**: MacBook Pro (`MacBookPro18,3`), Apple M1 Pro, 10 CPU cores, 16 GB RAM
- **Host OS**: macOS 26.6.1 (`25G76`), arm64
- **Target**: native macOS arm64
- **Rust toolchain**: `rustc 1.100.0-nightly (f7d782a3b 2026-08-19)`
- **Installed via**: built from source by the parity harness
- **Test duration**: `07:42:04` after the initial build
- **Repository state**: clean before and after the run

### Overall result

| Outcome | Count |
|---|---:|
[...]

PR #9228, opened 2026-08-30
fix(assert): test a RegExp matcher against String(thrown), and make it terminal

### LINKED ISSUE #9227: node:assert RegExp mismatch falls through to instanceof and throws TypeError
## What happened

`assert.throws()` with a non-matching RegExp reports a `TypeError` in Perry instead of the `AssertionError` produced by Node.js.

The shared matcher recognizes the expected value as a RegExp, tests the thrown value and its `.message`, then falls through to constructor matching when neither matches. That passes the RegExp object as the right-hand side of dynamic `instanceof`, changing assertion control flow for every API that shares this matcher.

This was found by the complete node-suite audit in #9202. Before the fix, `node:assert` was 69/70 PASS.

## What you expected

A RegExp is a complete matcher category. When the thrown error does not match it, `assert.throws()` should fail with `AssertionError` / `ERR_ASSERTION`, without attempting constructor matching.

The same matcher result must remain safe for `throws`, `doesNotThrow`, `rejects`, and `doesNotReject`.

## Minimal reproduction

```typescript
import assert from "node:assert/strict";

try {
  assert.throws(() => {
    throw new Error("nope");
  }, /will-not-match/);
} catch (error: any) {
  console.log(error.name, error.code ?? "no-code");
}
```

Command:

```bash
./run_parity_tests.sh \
  --suite node-suite \
  --module assert \
  --filter strict-throws-validation
```

Node.js v26.5.1:

```text
AssertionError ERR_ASSERTION
```

Perry 0.5.1519 before the fix:

```text
TypeError no-code
```

## Environment

- **Perry version**: `0.5.1519`
- **Commit**: `4a60b8ca1d2cfb9642a24604b0cec18c74863cac`
- **Host OS**: macOS 26.6.1, arm64
- **Target**: native macOS arm64
- **Node oracle**: `v26.5.1`
- **Rust toolchain**: `rustc 1.100.0-nightly (f7d782a3b 2026-08-19)`
- **Installed via**: built from source

## Diagnostic output

The failing fixture is:

```text
node-suite/assert/errors/strict-throws-validation
```

Root path:

```text
expected_error_matches
  -> RegExp checks do not match
  -> falls through to js_instanceof_dynamic(thrown, expected)
  -> RegExp is not a valid constructor RHS
  -> TypeError replaces the expected AssertionError
```

## Anything else

- Full audit: #9202
- Related historical matcher work: #2971 and #3247
- Acceptance: `./run_parity_tests.sh --suite node-suite --module assert` reports 70 PASS and zero non-PASS outcomes, with a focused runtime regression test.

TheHypnoo:
## Summary

The first commit on this branch added `return false` after the RegExp block in
`expected_error_matches`. That stopped the fallthrough into `instanceof`, but
kept the line that caused it — a retry of the pattern against `thrown.message`
— so a whole second bug survived.

Node tests a RegExp matcher against `String(thrown)` and nothing else, and
treats RegExp as a terminal matcher category. The block collapses to one
statement:

```rust
if let Some(matches_thrown) = regex_test_value(expected, thrown) {
    return matches_thrown;
}
```

That fixes both halves at the single point `throws`, `doesNotThrow`, `rejects`,
and `doesNotReject` all route through.

## The two bugs

**Wrong input.** An anchored pattern matching the bare message but not the
stringified error was wrongly accepted, because `String(new Error("nope"))` is
`"Error: nope"`:

```
assert.throws(() => { throw new Error("nope") }, /^nope/)
  node 26.5.1 → AssertionError ERR_ASSERTION
  before      → passes silently
```

Same for a thrown non-error carrying a `message` property, whose `String()` is
`"[object Object]"`.

**Wrong control flow.** A pattern matching neither input reached
`js_instanceof_dynamic(thrown, regexp)` and surfaced as a `TypeError` instead of
an `AssertionError`, so `throws`/`rejects` reported the wrong error class and
`doesNotThrow`/`doesNotReject` failed to rethrow the original error.

A RegExp value on a validator *key* (`{ message: /bad/ }`) is a different
matcher and is still tested against that property in `object_matcher_matches`.

The `thrown.message` retry dates to the original implementation in #1924 and was
never a workaround for anything else.

## Why the existing fixture could not catch this

`assert/errors/strict-throws-validation.ts` uses `/will-not-match/`, a pattern
matching neither the message nor the stringified error, so it cannot tell the
two inputs apart — and it prints only `name` and `code`.

`assert/errors/throws-regexp-matcher-input.ts` covers both inputs across all
four entry points and matches Node 26.5.1 byte for byte on 9 cases. The unit
tests replace the setjmp-based one, which the previous fix also passed:

```
regexp_matcher_tests_the_stringified_error_not_the_message ... FAILED   ← previous fix
                                                            ... ok      ← this commit
```

## Known remaining gap (not fixed here)

Node's generated message is `The input did not match the regular expression
/x/. Input:\n\n'Error: nope'\n`; Perry emits the generic `The thrown error did
not match the expected matcher`. That fallback is shared by **every** matcher
category, so patching only the regexp arm would leave the message layer
half-parity. Recorded in the changelog fragment as separate work.

## Related issue

Fixes #9227

Refs #9202

## Test plan

- [x] `RUST_TEST_THREADS=1 cargo test -p perry-runtime regexp_matcher_tests` — 3 passed
- [x] `./run_parity_tests.sh --suite node-suite --module assert` — **71 PASS, 0 fail, 0 compile fail, 0 crash, 0 skip (100.0%)**, against the pinned Node 26.5.1
- [x] New fixture diffed byte-for-byte against Node 26.5.1 — identical on all 9 cases
- [x] `cargo fmt` clean; `assert.rs` is 1368 lines (under the 2000-line cap)
- [x] Added `changelog.d/9228-assert-regexp-matcher-input.md` (the `lint` job requires a fragment for any `crates/` change)
- [x] Commit follows the conventional `fix:` format

**Note on the oracle:** run the suite with the pinned Node on PATH. With Node
v24 instead of 26.5.1, `assert/imports/calltracker-shape` reports a false
mismatch — `assert.CallTracker` still exists in v24 and was removed by v26, so
the oracle disagrees with itself, not with Perry. Same tree: 70/71 under v24,
71/71 under v26.5.1.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected regular expression matching in error assertions to consistently use the stringified thrown value.
  * Prevented failed regular expression matches from falling through to other matcher types.
  * Preserved property-specific matching for regular expressions used in validator objects.
  * Improved consistency across synchronous and asynchronous assertion APIs, including thrown strings and plain objects.

* **Tests**
  * Added coverage for matching, non-matching, and error-reporting scenarios.

### DIFF
--- a/changelog.d/9228-assert-regexp-matcher-input.md
+++ b/changelog.d/9228-assert-regexp-matcher-input.md
@@ -1,0 +1,32 @@
+`assert.throws(fn, /re/)` and friends now test the RegExp matcher the way Node
+does: against `String(thrown)` only, with a RegExp treated as a terminal
+matcher category.
+
+`expected_error_matches` used to give a non-matching pattern a second chance
+against `thrown.message`, then fall through to the `instanceof` /
+constructor-name / validation-function checks when that also failed. Two bugs
+came out of one block:
+
+- A pattern that matched the bare message but not the stringified error was
+  wrongly accepted. `assert.throws(() => { throw new Error("nope") }, /^nope/)`
+  passed, where Node reports an `AssertionError` because `String(err)` is
+  `"Error: nope"`. Same for a thrown non-error carrying a `message` property.
+- A pattern that matched nothing reached `js_instanceof_dynamic(thrown, regexp)`
+  and surfaced as a `TypeError` instead of an `AssertionError`, so
+  `assert.throws`/`rejects` reported the wrong error class and
+  `doesNotThrow`/`doesNotReject` failed to rethrow the original error.
+
+A RegExp *value on a validator key* (`{ message: /bad/ }`) is a different
+matcher and still tests against that property.
+
+The pre-existing `assert/errors/strict-throws-validation.ts` could not catch
+either bug: its pattern matched both the message and the stringified error, and
+it only printed `name`/`code`. `assert/errors/throws-regexp-matcher-input.ts`
+covers the two inputs separately across `throws`, `doesNotThrow`, `rejects`,
+and `doesNotReject`.
+
+Known remaining gap: the generated `AssertionError` message is Perry's generic
+"The thrown error did not match the expected matcher" rather than Node's
+"The input did not match the regular expression /x/. Input:\n\n'Error: nope'\n".
+That generic fallback is shared by every matcher category, so it is left for a
+separate change.
--- a/crates/perry-runtime/src/object/assert.rs
+++ b/crates/perry-runtime/src/object/assert.rs
@@ -267,6 +267,11 @@ fn expected_error_matches(thrown: f64, expected: f64) -> bool {
         if !is_null_or_undefined(message) && regex_test_value(expected, message).unwrap_or(false) {
             return true;
         }
+        // A RegExp is a complete matcher category. Falling through would
+        // incorrectly pass the RegExp object to `instanceof`, which throws a
+        // TypeError instead of letting assert.throws/assert.rejects report an
+        // AssertionError for a non-matching pattern.
+        return false;
     }
     // A plain object validator (e.g. `{ code: "ERR_X" }`) is a property-bag
     // matcher, never a constructor — its own enumerable keys must each equal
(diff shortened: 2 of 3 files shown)


proggeramlug:
Merged.

**Results across the batch:**

PR #9230, opened 2026-08-30
perf(codegen): flush packed accumulators at the throw site, restoring the fast path (7.99 → 0.95 ns)

### LINKED ISSUE #9210: Packed-loop clone does not flush loop-carried locals on an unwind edge (regression from #9185)

proggeramlug:
Fixes #9210. **Supersedes #9215** — this branch contains that revert as its first commit, so merging this alone resolves the wrong answer currently on `main` and restores the fast path. Merge #9215 instead if you want the bleeding stopped immediately and this reviewed at leisure.

## What was wrong

#9185 admitted `throw` to the packed-f64 versioned loop without emitting the loop-carried writeback on the unwind edge:

```ts
let s = 0;
try { for (let i = 0; i < arr.length; i++) { if (arr[i] === 40) throw PRE; s += arr[i]; } }
catch (e) { console.log(s); }   // perry: 0     node: 780
```

`PackedAccumulatorScope::finish` covers the fall-through exit, and the side-exit trampolines cover a mid-iteration deopt — but both are blocks the clone **branches to**, which is exactly how `break` and `continue` reach them. An unwind reaches neither: under invoke-EH `js_throw` leaves for the landing pad from inside the call, so a `catch` sees whatever happens to be in the real slot.

## The fix

`flush_packed_accumulator_locals` emits the stores **at the throw site**, between lowering the operand (which may itself read or write an accumulator through the redirect) and the throw itself — the only point that works, since the unwind edge is created inside `call_void`.

Three things worth review attention:

- **Both promoted representations.** A `+=` accumulator lives in a `DOUBLE` alloca; a `c++` counter lives in a separate i32 slot and converts back with `sitofp`. A flush walking only the float table would leave `c` stale while `s` looked correct — `both_accumulator_kinds_survive_a_taken_throw` asserts both.
- **It does not unregister**, unlike `finish`: lowering continues inside the clone after the throw, and later statements must keep reading the promoted values.
- **Sorted iteration**, because the side tables are hash-keyed. Verified: `__text` is byte-identical across repeated builds of the same source.

The admission now also checks the thrown operand the way `Stmt::Return` checks its value. #9185 admitted any throw and relied on `stmt_array_length_effect` to reject the constructing ones, which states the requirement somewhere weaker.

## Numbers

Quiet host (Mac mini, load 1.5), 5 reps × 3 runs, stable to 0.01 ns/op:

| shape | before (#9215) | this PR | node |
|---|---|---|---|
| `with_throw` | 7.99 | **0.95** | 1.10 |
| `with_break` | 0.95 | 0.95 | 1.11 |
| `no_throw` | 0.95 | 0.95 | 1.10 |

8.4× on the throw shape, and it moves from **7.2× slower than node to 1.16× faster**. Size: `__text` +192 B (+0.002%).

## Validation

- **The tests were verified to guard the flush, not merely pass beside it.** With the flush call disabled and everything else identical, both regression tests fail — `s` reads 0 instead of 780, `c` reads 0 instead of 40 — and pass again once restored. This is the check #9185 lacked.
- 17 differential programs against node, all identical. Includes four shapes that shadow `Error` (user class, `const` function, function-valued binding, block-scoped class), a module-scope accumulator observed by a caller's `catch`, an outer-function local observed through a closure, and an object-field accumulator.
- `packed_loop_abrupt_statements` 10/10, `loop_property_array_hoist` 12/12, `issue_8690_loop_versioned_arraylike` 3/3, `issue_8897_field_push_writeback` 3/3
- perry-hir 591/591, perry-codegen 1842/1842, local-binding-type audit OK
- `RUSTFLAGS="-D warnings" cargo check -p perry-codegen --all-targets`: clean

Refs #9151, #9185.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect loop-carried values after a `throw` transfers control to a `catch` block.
  * Ensured floating-point and integer loop accumulators are preserved during abrupt control flow.
  * Restricted fast-path handling to throws that are safe for optimized loop execution.
  * Improved reliability when optimized loops exit through exceptions or other abrupt control flow.

* **Tests**
  * Added coverage for thrown errors and preserved loop state across both accumulator types.

### DIFF
--- a/crates/perry-codegen/src/stmt/loops.rs
+++ b/crates/perry-codegen/src/stmt/loops.rs
@@ -796,6 +796,68 @@ impl PackedAccumulatorScope {
     }
 }
 
+/// Write the fast clone's promoted accumulators back to their real slots
+/// WITHOUT ending the redirect scope.
+///
+/// `PackedAccumulatorScope::finish` covers the fall-through exit and the
+/// side-exit trampolines cover a mid-iteration deopt, but both are blocks the
+/// clone BRANCHES to, and `break`/`continue` reach them the same way. An
+/// unwind edge reaches neither: under invoke-EH `js_throw` leaves for the
+/// landing pad from inside the call itself, so a `catch` observes whatever is
+/// in the real slot at that moment. Emitting the stores at the throw site is
+/// the only point that works, which is what #9185 was missing.
+///
+/// Unlike `finish` this does NOT unregister the redirects: lowering continues
+/// inside the clone afterwards, and later statements must keep reading the
+/// promoted values.
+pub(crate) fn flush_packed_accumulator_locals(ctx: &mut FnCtx<'_>) {
+    if ctx.numeric_accumulator_f64_slots.is_empty()
+        && ctx.deferred_integer_update_accumulators.is_empty()
+    {
+        return;
+    }
+    if ctx.block().is_terminated() {
+        return;
+    }
+
+    // The side tables are hash-keyed, so collect and sort before emitting —
+    // IR order must not depend on hash iteration order.
+    let mut unboxed: Vec<(u32, String, String)> = ctx
+        .numeric_accumulator_f64_slots
+        .iter()
+        .filter_map(|(id, alloca)| {
+            ctx.locals
+                .get(id)
+                .map(|real_slot| (*id, alloca.clone(), real_slot.clone()))
+        })
+        .collect();
+    unboxed.sort_by_key(|(id, _, _)| *id);
+    for (_, alloca, real_slot) in &unboxed {
+        // Same argument as `finish`: a genuine double's bits are its nanbox,
+        // numbers carry no heap edge so no barrier, and leaving the shadow
+        // state conservative only ever costs an extra root scan.
+        let value = ctx.block().load(DOUBLE, alloca);
+        ctx.block().store(DOUBLE, &value, real_slot);
+    }
+
+    let mut deferred: Vec<(u32, String, String)> = ctx
+        .deferred_integer_update_accumulators
+        .iter()
+        .filter_map(|id| {
+            let i32_slot = ctx.i32_counter_slots.get(id)?;
+            let dbl_slot = ctx.locals.get(id)?;
+            Some((*id, i32_slot.clone(), dbl_slot.clone()))
+        })
+        .collect();
+    deferred.sort_by_key(|(id, _, _)| *id);
+    for (_, i32_slot, dbl_slot) in &deferred {
+        let blk = ctx.block();
+        let value = blk.load(I32, i32_slot);
+        let as_double = blk.sitofp(I32, &value, DOUBLE);
+        blk.store(DOUBLE, &as_double, dbl_slot);
+    }
+}
+
 fn emit_packed_numeric_accumulator_admission(
     ctx: &mut FnCtx<'_>,
     body: &[Stmt],
@@ -5080,9 +5142,15 @@ fn stmt_is_packed_f64_loop_safe(
         // closure-captured accumulator was also correct, being boxed rather
         // than register-promoted, which is what kept the bug this narrow.
         //
-        // Re-admitting this needs the writeback emitted at the throw site, not
-        // just the admission; see #9210.
-        Stmt::Throw(_) => false,
+        // Admitted again now that `flush_packed_accumulator_locals` emits the
+        // writeback AT the throw site (#9210). The operand is checked the same
+        // way `Stmt::Return` checks its value; #9185 admitted any throw at all
+        // and leaned on `stmt_array_length_effect` to reject the constructing
+        // ones, which is a weaker guarantee than stating the requirement here.
+        Stmt::Throw(value) => {
+            packed_loop_abrupt_enabled()
+                && expr_is_packed_f64_loop_safe(ctx, value, arr_id, counter_id)
+        }
         Stmt::LabeledBreak(_)
         | Stmt::LabeledContinue(_)
         | Stmt::While { .. }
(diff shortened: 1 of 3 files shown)


proggeramlug:
Rebased onto `main` now that #9215 has merged — the revert commit this branch carried is the one that landed there, so only the flush commit remains and the conflict is gone. Diff is 3 files / +120 / −7.

Re-validated after the rebase rather than assuming it carried: 28 integration tests green (`packed_loop_abrupt_statements` 10/10, `loop_property_array_hoist` 12/12, `issue_8690` 3/3, `issue_8897` 3/3), 10 differentials node-identical, and the quiet-host number holds at 0.95 ns/op against node's 1.10.

proggeramlug:
**Please hold this one — I have found a latent wrong answer in it and am fixing it.** The flush itself is correct; the admission predicate guarding it is not tight enough.

`expr_is_packed_f64_loop_safe` accepts `Expr::Binary` over two `LocalGet`s, and `+` on an object invokes `valueOf`/`toString` — user code, which can THROW. That unwind leaves during operand evaluation, before the flush at the throw site ever runs, so the accumulators are stale exactly as in #9185:

```ts
const evil: any = { valueOf(): number { throw new Error("from valueOf"); } };
…
if (arr[i] === 40) throw (evil + two);
c++; s += arr[i];
// this PR: c=0 s=0     node: c=40 s=780
```

No Error construction involved, so it is not about the constructed-operand work — it is this PR's own admission rule. The fix is a dedicated predicate for a thrown operand: a bare `LocalGet` is fine (throwing a value coerces nothing), but any arithmetic, concatenation or Error-message stringification must be provably coercion-free, since each of those can dispatch to user code that unwinds.

Updating shortly with that predicate plus the regression case.

proggeramlug:
Merged. This is the real fix that #9215 was holding the door for, and it does what the revert couldn't: restores the fast path instead of giving it up.

**Verified both halves, because either alone would be a wrong merge.**

*It's correct.* All ten shapes from the #9215 investigation match node 26.5.1:

| shape | this PR | node |
|---|---|---|
| accumulate after the throw | 780 | 780 |
| accumulate before the throw | 820 | 820 |
| two locals, `s + ":" + n` | 780:40 | 780:40 |
| nested loops, `s + hits` | 17646 | 17646 |
| read via closure | 780 | 780 |
| taken break / continue / return | correct | correct |
| module-level vs function-local receiver | both correct | both correct |

The nested row is the one I'd single out. On the broken compiler it gave **17594** — a *partial* flush, short by 52, not a zero. That's the dangerous shape, because 17594 looks like an answer while 0 announces itself. A fix validated only on the zero-valued rows could have left it standing, so it's worth keeping in the regression set permanently. Credit to the peer who found that row.

*It's actually enabled.* A fix that's correct because the optimization silently stayed off would pass every one of those rows, so I IR-diffed `--trace llvm`:

- `throw PRE` (value already built) → 18 `packed_f64_fast` markers, 2 `acc.writeback_exit`
- `throw new Error("x")` (constructs) → none, correctly still generic

So the fast path is genuinely back for the admissible case and the constructing throw still isn't admitted.

For the record on how this got here: I merged #9185, and my review of it specifically hunted the unwind hazard and cleared it on the wrong distinction — I checked that a `try` *inside* the loop is rejected and concluded control never returns, missing that a `catch` *outside* the loop is still in scope for the loop-carried locals. Flushing at the throw site is the right resolution.

Validated alongside #9213, #9214, #9216, #9219 and #9224.

proggeramlug:
Superseded by #9235, which contains this commit and then fixes the `valueOf`/`toString` unwind hole described above. **Please merge #9235 rather than this PR** — on its own this one still admits `throw (evil + two)` and returns `c=0 s=0` where node gives `c=40 s=780`.

#9235 also turns out to be where the perf actually is: with the operand predicate corrected, `throw new Error("…")` reaches 0.95 ns/op against node's 7.71, so the constructed-throw shape goes from parity to 8.1× faster instead of staying on the generic path.

Issue #9234, opened 2026-08-30
Stack-map assert fires under PERRY_GC_HEAP_LIMIT=8 + FORCE_EVACUATE: a root-scan path still skips ensure_built() (#9191 gap)


proggeramlug:
My own #9191 (lazy stack-map index) left a path into the native root scan that does not build the index, and #9182's fail-closed assert catches it:

```
thread '<unnamed>' panicked at crates/perry-runtime/src/gc/roots/stack_maps.rs:1221:5:
perry: the native root scan ran before the stack-map index was built. An unbuilt index is
indistinguishable from an image with no native roots, so the collector would find no roots
on this frame and free live objects silently. This means a path into the root scan does not
pass through the point that builds the index.
```

## Repro

Any allocating program under a tight heap with forced evacuation — no `throw`, no packed loop, nothing exotic:

```ts
const arr: number[] = [];
for (let i = 0; i < 1024; i++) arr.push(i * 2);
const sink: object[] = [];
function churn(): number { let s = 0; for (let i = 0; i < arr.length; i++) s += arr[i]; return s; }
let out = 0;
for (let k = 0; k < 200; k++) { sink.push({ a: k, b: "x" + k }); if (sink.length > 50) sink.length = 0; out = churn(); }
console.log("s=" + out);
```

```
PERRY_GC_HEAP_LIMIT=8 PERRY_GC_FORCE_EVACUATE=1 ./prog
```

The program prints its correct result first and then panics, so the unguarded scan appears to be on a **shutdown or final-collection** path rather than a steady-state one — `gc_collect_minor_with_trigger`, `gc_collect_forced_evacuating_minor` and `gc_collect_full_mark_sweep_with_trigger` all got `roots::ensure_stack_maps_built()` in #9191, so it is likely a fourth entry point I missed. Does not reproduce without `PERRY_GC_HEAP_LIMIT=8`.

## Severity

The assert is doing its job — this is a crash, not a wrong answer, and only under a stress configuration. But it means the guarantee #9182 was written to enforce does not hold on every path: in a build where that assert is compiled out, this same path would scan for roots against an unbuilt index, find none, and free live objects silently.

Found while GC-stressing an unrelated packed-loop change; it reproduces with that change disabled and on programs containing none of the constructs it touches, so it is independent of it.

proggeramlug:
**Already fixed — this is a duplicate of #9233**, which landed about an hour before I filed this. My apologies for the noise.

[...]

PR #9235, opened 2026-08-31
perf(codegen): constructed throw operands keep the packed fast path (7.99 → 0.95 ns, 8.1× vs node)

### LINKED ISSUE #9232: Packed loop: a throw whose value is CONSTRUCTED stays on the generic path (7.99 vs 0.95 ns) — needs a GC-safety argument, not just admission
Follow-up to #9210 / #9230. Recording a measured 8.4× opportunity together with the reason I did **not** take it, so the next person starts from the hazard rather than rediscovering it.

## Where things stand

Quiet host (Mac mini, load 1.5), `arr` 512 f64 elements, never-taken throw, 5 reps × 2 runs, stable to 0.01 ns/op:

| loop body | perry | node | |
|---|---|---|---|
| `throw PRE` (pre-built value) | **0.95** | 1.10 | perry 1.16× faster |
| `throw new Error("bad")` | 7.99 | 7.73 | perry 1.03× slower |
| `throw new Error("bad " + i)` | 7.99 | 7.74 | |
| `throw new Error()` | 7.99 | 7.75 | |
| `throw "bad " + i` | 7.99 | 7.74 | |
| no throw at all | 0.95 | 1.10 | perry 1.16× faster |

Every CONSTRUCTED operand lands on exactly 7.99 — one shared generic path, and the Error-ness is irrelevant (a bare string concat behaves identically). Node shows the same cliff, so we are at parity here rather than behind; the prize is that **0.95 is achievable and node cannot reach it**.

## Why it is not just an admission tweak

#9185 admitted exactly these shapes and produced a silent wrong answer. #9230 fixed the writeback half of that, so the accumulator staleness is no longer the blocker. The remaining question is one #9185 never asked:

`Block::contains_gc_unsafe_call` (`crates/perry-codegen/src/block.rs:299`) exempts a block ending in `unreachable`:

```rust
if matches!(self.instructions.last(), Some(LlInst::Unreachable)) {
    return false;
}
```

That is the codebase sanctioning "control never returns, so a pointer the collector moved is never dereferenced" — the reasoning #9185 quoted. But it covers only the TERMINATING block. A constructed operand is lowered into blocks that PRECEDE it and end in ordinary branches, so the exemption does not reach them, and those blocks contain the allocating call.

The packed-f64 versioned loop does not run that scan at all (it is used by the class-field loop at `loops.rs:4669` and `:4167`), so its safety rests entirely on the admission predicates. That is why the wrong answer got through: nothing downstream was going to catch it.

## What a fix has to establish

Not "is the throw block safe" — that part is already argued — but: **the fast clone caches the array data pointer, and an allocation between the cache and the unwind can trigger a moving collection.** Someone has to show either that the cached pointer is dead from the allocation onward (plausible: the loop never resumes, and #9230's flush already stores the accumulators before the throw), or that it is properly rooted across it.

I did not attempt this. It is the [[cc-8770]] bug class — a bare pointer held across an allocation — which took eighteen sessions to root-cause the last time, and 3.4% against node is not the reason to re-enter it at the end of a long session. It is worth a design pass on its own terms, where the payoff is 8.4× and beating node on a very common validation shape rather than merely matching it.

`PERRY_PACKED_LOOP_TRACE=1` (#9204) names the rejecting gate if you pick this up.

proggeramlug:
Closes #9232. **Supersedes #9230** — this branch contains that commit and then fixes a latent wrong answer in it, so please take this one rather than merging #9230 on its own.

## Two gates, and why neither showed an effect alone

`if (bad) throw new Error("…")` sat at 7.99 ns/op. The length hoist rejected the construction node **and** the admission rejected the operand. I opened each separately, measured no change, and wrongly concluded the first was a no-op — they only move together. Opening both puts every throw shape at 0.95.

## The hole this also fixes

#9230's writeback is emitted **at** the throw site, after the operand is lowered. An operand that unwinds **on its own** therefore skips it, leaving the accumulators stale — #9185's defect one level deeper. Two expressions can unwind without looking like calls:

```ts
const evil: any = { toString() { throw new Error("x") } };
throw new Error(evil);     // stringifying the message dispatches to toString
throw (evil + two);        // `+` dispatches to valueOf/toString
```

Both returned `c=0 s=0` where node gives `c=40 s=780`. **The second involves no Error construction at all and was reachable through #9230's own admission rule**, which accepted `Expr::Binary` over two locals — so that PR needed this fix regardless of the perf work.

`throw_operand_cannot_unwind` replaces it. Throwing a value coerces nothing, so a bare local stays admissible whatever it holds; anything that **coerces** must be `expr_is_coercion_free_primitive` — literals, the loop counter, an element of the packed array (a genuine double by the clone's own guard), and arithmetic over those. Syntactic on purpose, and narrower than `expr_is_packed_f64_loop_safe`, whose question is "can the clone keep running", not "can this unwind".

Error identity is not decided here: HIR emits these nodes only for the intrinsic constructor (`lower_new` gates on `!shadowed_by_user_binding`), so a user `class Error` lowers to `Expr::New` and is still rejected — four shadowing forms are tested. `ErrorNewWithCause`/`WithOptions` stay out, since their options object is read at runtime and can carry getters.

## Numbers

Quiet host (Mac mini, load 1.5), 5 reps × 2 runs, stable to 0.01 ns/op. **Perry now beats node on every shape in this benchmark:**

| loop body | before | this PR | node | |
|---|---|---|---|---|
| `throw new Error("bad")` | 7.99 | **0.95** | 7.71 | 8.1× faster |
| `throw new Error("bad " + i)` | 7.99 | **0.95** | 7.75 | 8.2× |
| `throw new Error()` | 7.99 | **0.95** | 7.75 | 8.2× |
| `throw "bad " + i` | 7.99 | **0.95** | 7.73 | 8.1× |
| `throw PRE` | 0.95 | 0.95 | 1.11 | 1.17× |
| no throw | 0.95 | 0.95 | 1.10 | 1.16× |

Size: `__text` **+1024 B (+0.0099%)**, measured on identical sources with the admission toggled.

## Validation

- **33 differential programs against node, all identical.** Includes every hazard above, four `Error`-shadowing forms, a module-scope accumulator observed by a caller's `catch`, an outer-function local read through a closure, a `finally` reading the accumulators mid-unwind, an operand that throws before the outer throw, and taken throws under `PERRY_GC_FORCE_EVACUATE=1` / `PERRY_GC_HEAP_LIMIT=8` with post-unwind array checksums.
- The rejection tests are known to discriminate: each returned `c=0 s=0` against the build immediately before this predicate.
- `packed_loop_error_throw` 14/14 (new), `packed_loop_abrupt_statements` 10/10, `loop_property_array_hoist` 12/12, `issue_8690` 3/3, `issue_8897` 3/3
- perry-hir 591/591, perry-codegen 1842/1842, local-binding-type audit OK
- `RUSTFLAGS="-D warnings" cargo check -p perry-codegen --all-targets`: 0 errors

One unrelated find along the way, filed separately as #9234: a root-scan path that skips the stack-map build under `PERRY_GC_HEAP_LIMIT=8 PERRY_GC_FORCE_EVACUATE=1`, reproducible on programs containing none of the constructs here.

Refs #9151, #9185, #9210.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of errors thrown inside optimized counted loops.
  * Prevented incorrect loop optimization when error construction or value conversion can execute code that throws.
  * Preserved correct loop results when built-in errors, subclasses, or array values are used.
  * Correctly handles cases where the built-in `Error` name is replaced by user-defined code.
  * Fixed unary negation to properly invoke value conversion, propagate conversion errors, and preserve BigInt behavior.

* **Performance**
  * Enables the optimized loop path for safe, coercion-free error operands while maintaining correctness.

### DIFF
--- a/crates/perry-codegen/src/expr/unary.rs
+++ b/crates/perry-codegen/src/expr/unary.rs
@@ -32,7 +32,7 @@ pub(crate) fn lower(ctx: &mut FnCtx<'_>, expr: &Expr) -> Result<String> {
             // Everything else may be a BigInt at runtime — notably an indexed
             // read from a plain array — and routing it through ToNumber would
             // silently round large BigInts before `fneg` (#9142).
-            let dynamic_neg = matches!(op, UnaryOp::Neg)
+            let _dynamic_neg = matches!(op, UnaryOp::Neg)
                 && !statically_numeric
                 && !is_provably_not_bigint(ctx, operand);
             let (v, precomputed_truthy) = if matches!(op, UnaryOp::Not) {
@@ -44,13 +44,18 @@ pub(crate) fn lower(ctx: &mut FnCtx<'_>, expr: &Expr) -> Result<String> {
             let blk = ctx.block();
             match op {
                 UnaryOp::Neg => {
-                    if dynamic_neg {
-                        Ok(blk.call(DOUBLE, "js_dynamic_neg", &[(DOUBLE, &v)]))
-                    } else if numeric {
+                    if numeric {
                         Ok(blk.fneg(&v))
                     } else {
-                        let coerced = blk.call(DOUBLE, "js_number_coerce", &[(DOUBLE, &v)]);
-                        Ok(blk.fneg(&coerced))
+                        // Mirrors `Pos` below: anything not statically numeric
+                        // goes through the dynamic helper, which performs a real
+                        // ToNumeric. The old `js_number_coerce` + `fneg` fallback
+                        // answered NaN when the operand's `valueOf` THREW, so
+                        // `-{ valueOf() { throw … } }` silently produced NaN
+                        // where every other operator (`+x`, `~x`, `x * 2`)
+                        // propagated. `js_dynamic_neg` also keeps a BigInt a
+                        // BigInt, which is why `dynamic_neg` folded in here.
+                        Ok(blk.call(DOUBLE, "js_dynamic_neg", &[(DOUBLE, &v)]))
                     }
                 }
                 UnaryOp::Pos => {
--- a/crates/perry-codegen/src/stmt/loops.rs
+++ b/crates/perry-codegen/src/stmt/loops.rs
@@ -5164,7 +5164,8 @@ fn stmt_is_packed_f64_loop_safe(
         // `flush_packed_accumulator_locals`; without it this admission is the
         // #9185 wrong answer.
         Stmt::Throw(value) => {
-            packed_loop_abrupt_enabled() && throw_operand_cannot_unwind(ctx, value, arr_id, counter_id)
+            packed_loop_abrupt_enabled()
+                && throw_operand_cannot_unwind(ctx, value, arr_id, counter_id)
         }
         Stmt::LabeledBreak(_)
         | Stmt::LabeledContinue(_)
--- a/crates/perry-codegen/src/stmt/loops.rs
+++ b/crates/perry-codegen/src/stmt/loops.rs
@@ -5147,9 +5147,24 @@ fn stmt_is_packed_f64_loop_safe(
         // way `Stmt::Return` checks its value; #9185 admitted any throw at all
         // and leaned on `stmt_array_length_effect` to reject the constructing
         // ones, which is a weaker guarantee than stating the requirement here.
+        // The thrown operand is deliberately NOT required to be
+        // `expr_is_packed_f64_loop_safe`. That predicate asks "can the clone
+        // keep running after this", and after a throw it cannot: control
+        // leaves the loop for a landing pad and never returns, so no later
+        // iteration can observe a value the operand disturbed.
+        //
+        // What the operand still must not do is change `arr.length` before the
+        // hoisted bound is used, and that IS checked — `stmt_preserves_array_length`
+        // walks `Stmt::Throw(e)` into `e`, so `throw arr.pop()` is rejected
+        // there. Keeping the two questions in their own predicates is what lets
+        // a constructed operand (`new Error(…)`, `"bad " + i`) stay on the fast
+        // path; requiring loop-safety here costs 8.4x for no correctness gain.
+        //
+        // The accumulator writeback this needs is emitted at the throw site by
+        // `flush_packed_accumulator_locals`; without it this admission is the
+        // #9185 wrong answer.
         Stmt::Throw(value) => {
-            packed_loop_abrupt_enabled()
-                && expr_is_packed_f64_loop_safe(ctx, value, arr_id, counter_id)
+            packed_loop_abrupt_enabled() && throw_operand_cannot_unwind(ctx, value, arr_id, counter_id)
         }
         Stmt::LabeledBreak(_)
         | Stmt::LabeledContinue(_)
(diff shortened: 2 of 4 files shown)


proggeramlug:
Rebased now that #9230 has merged — only the constructed-throw commit remains, and the conflict is gone. Diff is 2 files / +492 / −2.

Worth flagging the ordering consequence: **the `valueOf`/`toString` unwind hole is now on `main`.** #9230 shipped the writeback together with an admission rule that accepts `Expr::Binary` over two locals, so this is reproducible against the current tip with no Error construction involved:

```ts
const evil: any = { valueOf(): number { throw new Error("from valueOf"); } };
const two: any = 2;
let s = 0, c = 0;
try { for (let i = 0; i < arr.length; i++) { if (arr[i] === 40) throw (evil + two); c++; s += arr[i]; } }
catch (e) { /* main: c=0 s=0 — node: c=40 s=780 */ }
```

`PERRY_PACKED_LOOP_ABRUPT=0` is the workaround in the meantime. This PR is the fix, so it is worth taking ahead of the perf story attached to it.

proggeramlug:
Merged, with one fix pushed onto the branch — for a bug **your own new test caught**, which is the best possible argument for the test.

**On the "two gates" finding:** worth stating plainly, because it's the general lesson. You opened each gate separately, measured no change, and concluded the first was a no-op. Two gates in series both have to open before anything moves, so measuring them one at a time reports "no effect" for a change that is doing exactly what it should. That's the same shape as a vacuous A/B — the measurement is real, the conclusion isn't — and it's worth remembering next time a codegen gate "does nothing."

**Verified the widening is real** rather than trusting the ns figure. IR markers for a summing loop:

| loop body | `main` | this PR |
|---|---|---|
| `throw PRE` (pre-built) | 18 `packed_f64_fast` | 18 |
| `throw new Error("x")` (constructs) | **0** | **18** |

So the constructed case genuinely joins the fast path.

**On the latent hole:** I could not reproduce it on `main`, and that turns out to be the point rather than a contradiction. `main` rejects constructed throw operands from the fast path, so the missed writeback isn't reachable there — your widening is what would expose it. Widening and guarding in the same change is the right call; had these been two PRs, the first would have shipped a live wrong answer. My probe (`toString` that throws, a getter that throws, constructed, pre-built, and two locals) is correct on both arms for that reason, and correct on yours *with* the fast path engaged, which is the assertion that matters.

**What I fixed:** `a_unary_operand_whose_valueof_throws_keeps_the_accumulators` failed with `msg=NaN` against `msg=unary valueOf` — accumulators right, thrown value wrong. The cause is one layer below packed loops: `js_dynamic_neg` negated the raw NaN-boxed bits after its BigInt arm, never performing ToNumeric. So `-{ valueOf() { throw … } }` answered `NaN`, while `+x`, `~x` and `x * 2` on the same object all propagated — `js_dynamic_pos` right below it already calls `to_numeric` first. Added that call and routed `Neg`'s non-numeric lowering through the helper, mirroring `Pos`.

I confirmed this is **not** from my `is_numeric_expr` change in #9165 by reverting that hunk and rebuilding: `neg NaN` either way. Pre-existing, and your test is the first thing in the tree to catch it.

**Validation:** `packed_loop_error_throw` 14/14; `perry-codegen` 31 suites / 0 failures; `perry-runtime` 2867 passed / 0 failed at `RUST_TEST_THREADS=1`; all 60 lint gates green. The full #9185/#9215/#9230 regression battery — direct and closure reads, module-level and function-local receivers, the nested partial-flush row (17646), self-unwinding operands via `toString` and a getter, and the break/continue/return controls — matches node 26.5.1 byte-for-byte.

#9230 was already merged before this arrived, so its commit rebased out as already-applied; this branch's remaining content is the widening plus the writeback fix.

PR #9242, opened 2026-08-31
fix(release): clear field-shadow and WebAssembly parity blockers

### LINKED ISSUE #9240: Release blocker: direct method guard ignores declared instance-field shadow
## Release impact

Parity shard 12 fails on release candidate `01463b81610f089cc566011d16ea5e5b08ec70a8`, blocking v0.5.1519.

- CI run: https://github.com/PerryTS/perry/actions/runs/33346563641
- Job: https://github.com/PerryTS/perry/actions/runs/33346563641/job/99353328925
- Failing test: `test_issue_945_scalar_method_guards`

## Reproduction

```bash
PERRY_SKIP_BUILD=1 \
PERRY_BIN="$PWD/target/debug/perry" \
PERRY_RUNTIME_DIR="$PWD/target/debug" \
./run_parity_tests.sh --filter issue_945_scalar_method_guards
```

Node prints:

```text
same-class field shadow: 22
inherited field shadow: 33
```

Perry prints `1` and `3` respectively. Setting `PERRY_METHOD_INLINE_PROBE=0` does not change the mismatch.

## Root cause

`emit_guarded_direct_method_call` proves receiver class/shape and prototype stability, then calls the prototype method directly. It does not account for a same-named declared instance field on the receiver class or an ancestor. Those fields are own properties created during construction and must win over the prototype method.

## Required fix

- Reject guarded direct-method lowering when the receiver class or an ancestor declares a public instance field with the method name.
- Conservatively reject it for computed instance fields whose runtime key cannot be proven different.
- Fall back to ordinary own-property lookup/dynamic dispatch.
- Preserve the existing scalar fast path for classes with no possible field shadow.

## Acceptance criteria

- `test_issue_945_scalar_method_guards` matches Node, including same-class and inherited field shadows.
- Existing positive scalar-method IR guard still proves no allocation/dispatch for the safe case.
- `cargo fmt --check --all` and relevant codegen tests pass.

### LINKED ISSUE #9241: Release blocker: WebAssembly default-runtime parity fixture is constant-folded into host mode
## Release impact

Parity shard 9 fails on release candidate `01463b81610f089cc566011d16ea5e5b08ec70a8`, blocking v0.5.1519.

- CI run: https://github.com/PerryTS/perry/actions/runs/33346563641
- Job: https://github.com/PerryTS/perry/actions/runs/33346563641/job/99353328887
- Failing test: `test_parity_webassembly_graceful_fail_default`

## Observed behavior

The checked-in expectation is the default-runtime graceful-degradation contract:

```text
validate valid: false
compile rejected: true CompileError
...
loader result: degraded
```

The CI artifact shows both Node and Perry instead produce the real WebAssembly result:

```text
validate valid: true
compile: resolved
Module: constructed
loader result: loaded
```

Perry therefore matches Node; the fixture is not exercising the mode its name and expected output describe.

## Root cause

The fixture resolves the namespace as:

```ts
(globalThis as any)["Web" + "Assembly"]
```

That expression is now constant-folded before WebAssembly feature detection. The compiler correctly auto-links the wasm host, so the expected default-runtime output becomes stale.

## Required fix

Construct the property name at runtime (for example from array parts) so the fixture reliably avoids static host auto-linking and continues to cover the default spec-shaped graceful-failure namespace.

## Acceptance criteria

- The filtered parity test passes against its existing graceful-degradation expectation.
- The test comment explains why a constant-foldable string concatenation must not be used.
- Real/direct WebAssembly usages continue to auto-link and retain their existing coverage.

proggeramlug:
Fixes #9240.
Fixes #9241.

Clears the two new failures from the v0.5.1519 full CI run:

- guarded direct method calls now fall back to own-property lookup when the class or an ancestor declares a same-named instance field (and conservatively for computed fields), preserving JavaScript field-over-prototype semantics;
- the default-runtime WebAssembly fixture now computes its namespace key at runtime, preventing constant folding from accidentally enabling the real wasm host.

Release evidence on the exact candidate base `01463b81610f089cc566011d16ea5e5b08ec70a8`:

- `test_issue_945_scalar_method_guards`: PASS, 100% parity
- `test_parity_webassembly_graceful_fail_default`: PASS, 100% expected-output parity
- `scripts/run_issue_945_scalar_method_ir_guard.sh`: PASS (safe scalar fast path retained)
- `cargo fmt --check --all`: PASS

Failed release jobs:

- parity 12: https://github.com/PerryTS/perry/actions/runs/33346563641/job/99353328925
- parity 9: https://github.com/PerryTS/perry/actions/runs/33346563641/job/99353328887

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved method dispatch when inherited or computed instance fields share a name with a method.
  * Prevented direct method calls from bypassing own-property overrides, ensuring the correct implementation is selected.
  * Improved compatibility for applications that use fields and methods with the same name across class hierarchies.
  * Corrected WebAssembly feature detection in the default runtime, restoring reliable graceful degradation when the feature is unavailable.

### DIFF
--- a/changelog.d/9242-release-parity-blockers.md
+++ b/changelog.d/9242-release-parity-blockers.md
@@ -1,0 +1,4 @@
+Method calls now preserve same-named instance fields declared on the receiver
+class or an ancestor instead of bypassing them with direct prototype dispatch.
+The default-runtime WebAssembly parity fixture also reliably stays out of the
+auto-linked host mode, restoring coverage of graceful degradation.
--- a/crates/perry-codegen/src/lower_call/method_override.rs
+++ b/crates/perry-codegen/src/lower_call/method_override.rs
@@ -867,6 +867,37 @@ pub(super) struct SubclassDispatchArm {
     pub target_fn: String,
 }
 
+/// A declared instance field is an own property on every constructed object,
+/// so it wins over a same-named prototype method. The direct-method guards
+/// prove the receiver's class/shape and prototype stability, but that is not
+/// enough to skip ordinary own-property lookup when the expected shape itself
+/// contains the method name. Computed fields are conservatively treated as a
+/// possible shadow because their runtime key is not available here.
+fn class_chain_may_declare_method_field(ctx: &FnCtx<'_>, class_name: &str, property: &str) -> bool {
+    let mut current = Some(class_name.to_string());
+    let mut seen = std::collections::HashSet::new();
+    for _ in 0..64 {
+        let Some(name) = current else {
+            return false;
+        };
+        if !seen.insert(name.clone()) {
+            return true;
+        }
+        let Some(class) = ctx.classes.get(&name).copied() else {
+            return true;
+        };
+        if class
+            .fields
+            .iter()
+            .any(|field| field.key_expr.is_some() || (!field.is_private && field.name == property))
+        {
+            return true;
+        }
+        current = class.extends_name.clone();
+    }
+    true
+}
+
 /// Emit a typed-feedback runtime guard before a known class-method direct call.
 ///
 /// The guard validates that the receiver still has the expected class shape,
@@ -892,6 +923,13 @@ pub(super) fn emit_guarded_direct_method_call(
     shape_only_guard: bool,
     subclass_arms: &[SubclassDispatchArm],
 ) -> Option<String> {
+    // `class C { m = fn; m() {} }` and an inherited field with the same name
+    // both require ordinary own-property lookup. Falling back here reaches
+    // `emit_own_method_override_check` / dynamic dispatch in the caller.
+    if class_chain_may_declare_method_field(ctx, receiver_class_name, property) {
+        return None;
+    }
+
     let truthy_result_kind = ctx
         .truthy_call_result_requested
         .then(|| constructive_method_truthiness(ctx, direct_fn))
--- a/crates/perry-codegen/tests/native_proof_regressions.rs
+++ b/crates/perry-codegen/tests/native_proof_regressions.rs
@@ -13896,6 +13896,16 @@ fn scalar_method_boolean_predicate_rejects_mutation_call_accessor_and_dynamic_pr
                 ),
                 "mutation must dispatch directly to the resolved method on the heap receiver:\n{ir}"
             );
+        } else if case == "inherited_field_shadow" {
+            // A declared field on the base class is an own property on the
+            // constructed child. The safe fallback therefore probes the own
+            // slot and invokes its value, rather than entering the prototype
+            // method dispatcher that would skip the shadow.
+            assert!(
+                ir.contains("call double @js_object_get_own_field_or_undef")
+                    && ir.contains("call double @js_native_call_value"),
+                "inherited field shadow must probe and dispatch the own method value:\n{ir}"
+            );
         } else {
             assert!(
                 ir.contains("call double @js_native_call_method"),
--- a/test-files/test_parity_webassembly_graceful_fail_default.ts
+++ b/test-files/test_parity_webassembly_graceful_fail_default.ts
@@ -1,7 +1,9 @@
-// Resolve the namespace through a computed key so this test deliberately does
-// not trigger the compiler's static WebAssembly host auto-linking. It checks
-// the default runtime's honest graceful degradation instead.
-const WA: any = (globalThis as any)["Web" + "Assembly"];
+// Resolve the namespace through a runtime-computed key so this test
+// deliberately does not trigger WebAssembly host auto-linking. A `+` here is
+// constant-folded before feature detection, which would turn this default-
+// runtime fixture into a host-runtime test.
+const namespaceKey = ["Web", "Assembly"].join("");
+const WA: any = (globalThis as any)[namespaceKey];
 const validAdd = new Uint8Array([
   0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00,
   0x01, 0x07, 0x01, 0x60, 0x02, 0x7f, 0x7f, 0x01,


proggeramlug:
Not merging yet — **this regresses a pre-existing native-proof guard**, and the PR's cited evidence is a different test. The behaviour change itself looks right; it's the IR proof that moved.

**The regression.** `perry-codegen --test native_proof_regressions scalar_method_boolean_predicate_rejects_mutation_call_accessor_and_dynamic_property` fails on the `inherited_field_shadow` case:

```
inherited_field_shadow must keep dynamic method dispatch fallback
```

A/B on the same tree, only `method_override.rs` swapped:

- `main`: `test result: ok. 3 passed`, exit 0
- this PR: exit 101

The assertion wants `call double @js_native_call_method` in the emitted IR. With your early return the module instead emits `js_object_get_field_by_name_f64` + `js_native_call_value` — a lookup-then-call rather than a single dynamic dispatch. Both are dynamic, so this may well be a case of the proof test pinning one specific safe form and your change picking a different safe form. But that IR also contains a direct `call double @perry_method_…__Point__isAbove`, and I could not establish whether that is a second call site in the fixture or the scalar fast path still being taken. That guard exists precisely to assert the fast path *isn't* taken for these cases, so I'm not willing to widen the assertion on the assumption it's the former.

The specific question for you: with `emit_guarded_direct_method_call` returning `None`, does the caller reach the same safety point it reached before, or a different one that happens to be correct for a different reason? If the former, the test's assertion should accept either dynamic form and this is ready. If the latter, that's worth understanding before it lands in a release.

**What checks out.** The behaviour is right, and I verified it independently rather than taking the issue's word:

| case | `main` | this PR | node 26.5.1 |
|---|---|---|---|
| `class A { m = () => … }` | field-A ✓ | field-A ✓ | field-A |
| `class C extends B { m = … }` over `B.m()` | **method-B ✗** | field-C ✓ | field-C |
| same, 200-iteration loop (guarded path) | **method-B ✗** | field-C ✓ | field-C |
| `class D extends B {}` (no shadow) | method-B ✓ | method-B ✓ | method-B |
| computed `["m"] = …` | computed-E ✓ | computed-E ✓ | computed-E |
| private `#m` + public `m()` | method-F:1 ✓ | method-F:1 ✓ | method-F:1 |
| field on a grandparent (`I extends H extends G`) | field-G ✓ | field-G ✓ | field-G |
| runtime `b.m = …` assignment | own-assigned ✓ | own-assigned ✓ | own-assigned |

So the fix does what it says, the private-field and no-shadow cases are correctly left alone, and the loop rows confirm it holds on the guarded direct-call path rather than only the first call.

`scripts/run_issue_945_scalar_method_ir_guard.sh` also passes, so the scalar fast path is retained where it should be — the conservative bail isn't over-broad.

**Two smaller notes.**

The `class_chain_may_declare_method_field` design is the right shape: every uncertain case (cycle, unresolved class, computed key, 64-deep cap) returns `true` and disables the optimization, so it fails safe. Worth keeping that property explicit in a comment — it's the kind of thing a later "simplification" quietly inverts.

The WebAssembly fixture is at **exactly** its baseline cap: `max_diff_lines: 11` and it produces 11. Within budget, but with zero headroom, so any future line added to that fixture's output turns `parity_matrix_baseline.json` red without anyone changing behaviour. Might be worth raising the cap or trimming a line while you're here.

**Validation run:** `perry-runtime` 2867 passed / 0 failed at `RUST_TEST_THREADS=1`; `perry-codegen` 14 suites clean and the one above failing.

proggeramlug:
The 4 `gap-suite (1)`/`(2)` failures on this PR are **not caused by it** — they are pre-existing on `main`, tracked as #9244.

Verified locally with a 4-arm A/B on the same build and harness config (`perry-dev`, `PERRY_SKIP_BUILD=1`, `PERRY_NO_AUTO_OPTIMIZE=1`, node 26.5.1):

| test | this PR | PR + narrowed predicate | `89cde4ff14` (main, PR reverted) | `a722b4cca2` (pre-#9169) |
|---|---|---|---|---|
| `test_gap_iterator_helpers_2874` | FAIL | FAIL | **FAIL** | PASS |
| `test_gap_language_types_object_part_a` | FAIL | FAIL | **FAIL** | PASS |
| `test_gap_object_string_wrappers` | FAIL | FAIL | **FAIL** | PASS |
| `test_gap_5592_class_expr_rebind_computed_accessor` | FAIL | FAIL | **FAIL** | PASS |
| `test_issue_945_scalar_method_guards` (#9240) | **PASS** | PASS | FAIL | FAIL |
| `test_parity_webassembly_graceful_fail_default` (#9241) | **PASS** | PASS | PASS | PASS |

Column 3 is this PR's codegen change fully reverted on top of its own merge-base: all four still fail, so the PR causes none of them. Its only effect on this set is `test_issue_945_scalar_method_guards` FAIL → PASS, which is exactly what #9240 asks for.

The four regress at #9169 and pass at its parent (#9244). They reached `main` because `gap-suite-build` on #9169 was cancelled, so every dependent gap shard was skipped.

The two `self-test-checker` reds are also cancellations, not failures.

### One follow-up worth filing (not blocking)

`class_chain_may_declare_method_field` returns `true` for a receiver class that is **not in `ctx.classes`**, and `ctx.classes` holds only user-declared classes. Every builtin receiver (`Object`, `Iterator`, `Number`) and every anonymous class expression therefore takes the bail-out, losing the guarded direct-method fast path — a perf deopt on a very hot path, not a correctness issue. The pre-existing `canonical_shape_excludes_own_property` (`dynamic_dispatch.rs:41`) handles the same question in the safe polarity: unknown class → "cannot prove" → *retain the runtime own-property probe*, rather than abandoning the fast path. Narrowing to `ctx.classes.get(..) else { return false }` plus `field.key_expr.is_none()` measured identical on all 6 tests above, so it is a clean follow-up rather than a change this PR needs.

proggeramlug:
Rebased onto `main` @ `fad4cfb460`, which now carries the #9244 fix (#9247).

The 4 `gap-suite` failures previously attributed to this PR were pre-existing on `main` — regressions from #9169, which merged with its gap shards cancelled. With #9247 landed they should clear, and this PR's own content is unchanged by the rebase (still exactly the same 4 files, 58 insertions / 4 deletions).

For the record, #9247's CI confirmed the diagnosis: `gap-suite (1)` and `(2)` — the two shards that reported these failures — are green on `main` now, as is `e2e-scoped`.

proggeramlug:
Correction to my earlier comment on this PR.

I stated that `guard_failures_match_node_and_unsafe_parameters_stay_generic` (`crates/perry/tests/issue_8774_argument_shape_clones.rs`) was fallout from this PR. **That was wrong, and I am retracting it.**

My evidence was a window correlation: the test printed `... ok` at `01463b8161`, FAILED at `83754818ea`, and this PR was the only commit touching `method_override.rs` between them. I called that "confirmed" when it was a hypothesis.

Measured since, on a coherent tree (static wrappers rebuilt at the tested commit — my first two attempts were invalid, one from a stale runtime archive that failed the compiler/runtime coherence check in 19s rather than running):

| `class_chain_may_declare_method_field` bail | result |
|---|---|
| as merged | FAILED — `all selected semantic-fixture routes originate at fresh contained locals` |
| narrowed to visible declarations only | FAILED, same assertion |
| **fully disabled (pre-#9242 behaviour)** | **FAILED, same assertion** |

The last row is decisive: the test fails identically with this PR's change present, narrowed, or absent. **This PR does not cause it.** The cause lies elsewhere in main's churn between the two candidates, and the failure is currently unattributed.

I had also proposed narrowing the predicate (unknown class → `false`, requiring `key_expr.is_none()`). That change is no longer justified by this evidence and I am not pursuing it — the over-broad arms remain a reasonable cleanup on their own merits, but nothing measured here requires them.

Issue #8907, opened 2026-08-28
macOS arm64 0.5.1220: any program importing node:http fails to link — 17 undefined _js_*http* symbols from libperry_stdlib.a


proggeramlug:
**perry 0.5.1220** (current latest release), macOS arm64, clang from Xcode CLT.

Any program that imports `node:http` fails at the link step. The symbols are referenced from `libperry_stdlib.a` but defined nowhere on the link line.

## Minimal repro

```ts
import { createServer } from "node:http";

function main(): void {
  const s = createServer((_req, res) => { res.end("ok"); });
  s.listen(0, () => { s.close(); });
}
main();
```

```
$ perry compile httpmin.ts -o /tmp/httpmin
...
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1
Error: Linking failed
```

Remove the import and it links and runs:

```ts
function main(): void { console.log("ok"); }
main();
```
```
$ perry compile nohttp.ts -o /tmp/nohttp
Wrote executable: /tmp/nohttp
Binary size: 7.6MB
```

## The undefined symbols — 17

```
_js_ext_http_agent_dispatch_method            _js_http_has_pending
_js_ext_http_agent_dispatch_property          _js_http_incoming_message_pipe
_js_ext_http_agent_dispatch_property_set      _js_http_is_incoming_message
_js_ext_http_agent_is_handle                  _js_http_response_trailers
_js_ext_http_client_incoming_message_is_handle    _js_node_http_create_server_with_options
_js_ext_http_client_incoming_message_set_encoding _js_node_http_server_close
_js_ext_http_client_inflight                  _js_node_http_server_listen
_js_ext_http_client_request_dispatch_method
_js_ext_http_client_request_dispatch_property
_js_ext_http_client_request_is_handle
```

All referenced from `perry_stdlib::common::dispatch_http::*` inside `libperry_stdlib.a`, e.g.

```
"_js_http_response_trailers", referenced from:
    perry_stdlib::common::dispatch_http::dispatch_client_incoming_property::h6300abf73b28a348
      in libperry_stdlib.a[260](perry_stdlib-….rcgu.o)
```

The `_js_ext_http_*` names suggest `libperry_ext_http.a` should be supplying them and is not reaching the link line; the bare `_js_node_http_*` and `_js_http_*` ones look like they should come from the runtime.

## Notes

[...]

PR #9243, opened 2026-08-31
test(link): regression guard for #8907 — full-stdlib links without the http ext archive

steinybot:
## Summary

Adds an end-to-end regression test for #8907: a program that references a `cgu.0`-only stdlib symbol (`new Blob([...])`) with no `node:http` import must link and run. The bug itself was already fixed by #5983 (v0.5.1239) — this fills the missing regression coverage.

## Changes

- `crates/perry/tests/issue_8907_ext_http_cgu_link.rs`: compile `const b = new Blob(["x"]); console.log(b.size)`, assert `perry compile` links and the binary prints `1`.
- `changelog.d/9243-ext-http-cgu-link-regression-test.md`: changelog fragment.

Root cause (for context): #5831 leaked `external-http-client-pump` into perry-stdlib's `full` feature, which declares `js_ext_http_*` symbols defined only in `perry-ext-http`. That archive is linked per-program by the compile driver, and only when the program imports `node:http`. The release packs stdlib into one monolithic `cgu.0` object, so referencing any symbol it uniquely defines (`Blob` → `js_blob_new`) pulls the whole member and its unresolved http references into the link. #5983 drops the pump from `full`; the driver activates it only when `node:http` is imported and `perry-ext-http` is on the link line.

Complements the manifest-level guard in `issue_8587_prebuilt_stdlib_http_isolation` (asserts `full` does not enable the external http pumps) with a real end-to-end link.

## Related issue

Refs #8907. Fixed in code by #5983 (v0.5.1239); this PR adds regression coverage.

## Test plan

```
cargo test -p perry --test issue_8907_ext_http_cgu_link
# test cgu0_program_links_without_http_ext_archive ... ok
cargo fmt -p perry -- --check   # clean
```

Also confirmed against the shipped binaries: released `perry 0.5.1220` fails the same repro with 14 undefined `js_ext_http_*` symbols; current `main` links and runs it (prints `1`). The fix commit was located by `git bisect` over `cgu=1` dist builds → #5983.

- [ ] `cargo test --workspace` passes (ran the scoped `-p perry --test issue_8907_ext_http_cgu_link` instead — test-only change)
- [x] Added a `#[test]` in the affected crate
- [ ] (if CLI / stdlib / runtime API changed) Updated `docs/src/` — n/a, test-only
- [ ] (if touching a platform UI backend) Built `-p perry-ui-<backend>` — n/a

## Screenshots / output

Released 0.5.1220 (the bug):

```
  "_js_ext_http_agent_dispatch_method", referenced from:
  ... (14 total)
ld: symbol(s) not found for architecture arm64
```

Current `main` (fixed): `Wrote executable`, runs, prints `1`.

## Checklist

- [x] My commits follow the `feat:` / `fix:` / `docs:` / `chore:` prefix convention
- [ ] I've read CONTRIBUTING.md and agree to the Code of Conduct

## Summary by CodeRabbit

- **Tests**
  - Added an end-to-end regression test ensuring programs using `Blob` compile, link, run successfully, and produce the expected output without requiring unrelated HTTP functionality.

- **Documentation**
  - Added a changelog entry documenting the regression coverage and confirming the issue is resolved.

### DIFF
--- a/changelog.d/9243-ext-http-cgu-link-regression-test.md
+++ b/changelog.d/9243-ext-http-cgu-link-regression-test.md
@@ -1,0 +1,13 @@
+Add an end-to-end regression test for #8907: a program that references a
+`cgu.0`-only stdlib symbol (`new Blob([...])`) with no `node:http` import must
+link and run.
+
+The bug (released v0.5.1220): #5831 leaked `external-http-client-pump` into
+perry-stdlib's `full` feature, which declares `js_ext_http_*` symbols defined
+only in `perry-ext-http`. That archive is linked per-program by the compile
+driver, and only when the program imports `node:http` — so a full-stdlib link
+without an http import left those references unresolved. The release packs
+stdlib into one monolithic `cgu.0` object, so any reference that pulls that
+member drags the http references in with it. Fixed by #5983 (v0.5.1239), which
+drops the pump from `full`; this pins the end-to-end link that the manifest-level
+guard in `issue_8587_prebuilt_stdlib_http_isolation` does not exercise.
--- a/crates/perry/tests/issue_8907_ext_http_cgu_link.rs
+++ b/crates/perry/tests/issue_8907_ext_http_cgu_link.rs
@@ -1,0 +1,77 @@
+//! Regression test for #8907.
+//!
+//! Reported as "any program importing node:http fails to link" on the released
+//! 0.5.1220 macOS arm64 binary, with ~17 undefined `_js_ext_http_*` /
+//! `_js_http_*` symbols referenced from `libperry_stdlib.a`. The node:http
+//! framing is a red herring. The real trigger: prebuilt `libperry_stdlib.a`
+//! packs a large slice of stdlib into one monolithic `cgu.0` object, and that
+//! object *also* carries the unresolved `js_ext_http_*` references. Pulling the
+//! member for any symbol it uniquely defines drags those references into the
+//! link. Only a `node:http` import adds `libperry_ext_http.a` (which defines
+//! them), so a program that touches `cgu.0` without importing `node:http`
+//! links with the symbols unresolved.
+//!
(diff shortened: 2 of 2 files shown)


proggeramlug:
Merged. Thanks — the root-cause writeup is the valuable part here, and I verified its premise directly rather than taking it on trust.

Confirmed in the built archive: `js_blob_new` is defined in `…perry_stdlib.c4d96a4db3fa0ceb-cgu.0.rcgu.o` — exactly the monolithic object your analysis names — and that archive currently carries **0** undefined `_js_ext_http_*` references. So the coupling is genuinely gone on `main`, which is why the test passes, and `Blob` is the right minimal trigger: pulling the member for a `cgu.0`-only symbol is what would drag the unresolved references back into the link. "The node:http framing is a red herring" is the insight that makes this test possible at all — a test written against the *reported* symptom would have imported `node:http` and never reproduced anything.

The test itself is built to be able to fail, which is what I care about in a regression guard:

- `--no-cache`, so it can't pass on a cached artifact;
- it **runs** the binary and asserts `size == 1` rather than only asserting the compile exited zero — so a compile that succeeded without linking the member coherently is still caught;
- `CARGO_BIN_EXE_perry`, so it tests the compiler built from the tree rather than whatever is on PATH.

`perry --test issue_8907_ext_http_cgu_link`: 1 passed, exit 0. `ci_e2e_scope.py` picks the suite up automatically from the diff (`crates/perry/tests/` suites are diff-driven; the manual `_CODEGEN_SUITES` list is only for `perry-codegen`), so it will actually run per-PR rather than being dark until nightly — worth saying because a suite that CI never invokes is the exact failure mode #7708 and #9193 were about.

One nit I could not push to your branch, since it's on your fork: `cargo fmt --all -- --check` wants the `std::fs::write(...)` call at line 33 wrapped across lines, which fails the `warnings`/fmt lint gate. I'm landing that as an immediate follow-up on `main` rather than bouncing the PR back to you for a line break.

PR #9233, opened 2026-08-30
fix(release): clear final CI blockers

proggeramlug:
## Summary

- build the lazy stack-map index before both budgeted GC cycle constructors reach a root scan
- guard the `getDeviceModel` string ABI directly in `perry-dispatch`
- skip only the platform-backed #5972 end-to-end case when the host UI archive is not provisioned, while retaining the environment-independent crash regression

Closes #9229.
Closes #9231.

## Release evidence

- Full CI run 33336867602 identified both failures on the exact v0.5.1519 candidate SHA
- the failed Linux `image_convolution` artifact reproduces the lazy stack-map abort on `perrymaster.skelpo.net`
- the same artifact exits 0 with `PERRY_LAZY_STACK_MAPS=0`, proving the lazy-init path

## Local validation

- `cargo fmt --check --all`
- `cargo test -p perry-dispatch`
- `cargo test -p perry-runtime --lib -- --test-threads=1` (2855 passed, 4 ignored)
- focused #5972 platform-backed test passes with an explicit skip when the archive is unavailable

The platform-independent numeric/NaN object-key case already passed in the failed release shard; the only failure in that suite was the missing GTK archive.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of full and fallback memory cleanup cycles by ensuring application roots are prepared before scanning.
  * Corrected runtime handling for device model and operating system version values.
* **Tests**
  * Added coverage confirming device identity APIs return values in the expected format.
  * Platform-specific device model tests now skip cleanly when required host UI components are unavailable.
* **Documentation**
  * Documented fixes related to memory cleanup and device identity integration coverage.

Issue #9231, opened 2026-08-30
Lazy stack-map index misses budgeted GC cycle entry points


proggeramlug:
## Problem

Release Full CI run 33336867602 fails `compiler-output-regression` on the `image_convolution` workload. The program prints the correct checksum, then aborts with SIGABRT:

```text
perry: the native root scan ran before the stack-map index was built
```

Failed job: https://github.com/PerryTS/perry/actions/runs/33336867602/job/99326355576

The lazy stack-map change in #9191 calls `ensure_stack_maps_built()` in three synchronous collection wrappers, but the budgeted cycle starters in `gc/policy.rs` construct `GcCycleState` directly and bypass all three wrappers:

- `gc_start_budgeted_full_cycle`
- `gc_start_budgeted_minor_fallback_cycle_with_snapshot`

The fail-closed assertion from #9182 therefore fires when a budgeted cycle reaches its root scan.

## Reproduction

```bash
PERRY_GC_TRACE=1 ./image_convolution
```

Observed: correct checksum followed by abort. The exact failed CI artifact reproduces on `perrymaster.skelpo.net`. Setting the documented kill switch proves the cause:

```bash
PERRY_LAZY_STACK_MAPS=0 PERRY_GC_TRACE=1 ./image_convolution
```

Observed: exit 0.

## Required fix

Build the owed stack-map index before both budgeted cycle constructors can reach a root scan, while allocation is still legal. Keep the fail-closed assertion enabled.

## Acceptance criteria

- Exact `image_convolution` compiler-output workload exits 0 with lazy stack maps enabled and `PERRY_GC_TRACE=1`.
- The same workload remains green with `PERRY_LAZY_STACK_MAPS=0`.
- Runtime tests and Full CI pass.

PR #9245, opened 2026-08-31
fix(gc): build the stack-map index in the cycle constructor, not at five hand-wired entries

### LINKED ISSUE #9234: Stack-map assert fires under PERRY_GC_HEAP_LIMIT=8 + FORCE_EVACUATE: a root-scan path still skips ensure_built() (#9191 gap)

proggeramlug:
Closes #9234. **This adds no fix** — #9234 was already fixed by #9233, which landed an hour before I filed it and neither of us noticed. What it removes is the way this same defect keeps coming back.

## First, the attribution, proven rather than inferred

Removing exactly the two `ensure_stack_maps_built()` calls #9233 added brings the abort back, byte-identical, on both repros:

```
gc4_nothrow  rc=134  "the native root scan ran before the stack-map index was built"
gc3_nested   rc=134  "the native root scan ran before the stack-map index was built"
```

Restore them and both exit 0 with node-identical output. So #9233 is the fix, and #9234 is a duplicate. I have closed it as such.

## Three rounds of the same mistake

- **#9191** made the stack-map index lazy and wired the four `gc_collect_*` entries **by hand**.
- **#9231** found that budgeted cycles construct `GcCycleState` **directly** and enter through none of them — the first root-scan step hit #9182's fail-closed guard and aborted.
- **#9233** wired those two sites — **by hand again**.

That is five hand-wired call sites for one invariant, and the next entry point starts the cycle over.

`policy.rs` already carries this exact lesson from a different bug, a few hundred lines away:

> The first cut of #7726 wired the two `gc_start_budgeted_cycle_for_pressure` sites by hand and missed the one in `gc::gc_collect_minor_with_trigger_inner` — which is the site the shipped safepoint path actually takes. The backoff then never fired, and the whole change measured as a 30 ms no-op on `retain.ts` while every test still passed. **Recording inside the predicate makes a future call site correct by construction.**

## The change

`GcCycleState::new_full` and `new_minor_fallback` are the choke point every cycle passes through — all four construction sites, the guarded entries and the budgeted ones alike. Building the index there makes a future entry point correct without anyone remembering.

The existing entry calls **stay**. They run while allocation is unambiguously legal and earlier in the sequence, so removing them would change ordering for a benefit I cannot measure; and `ensure_built` is an `Acquire` load and a return once the index exists, so the overlap costs one atomic load per cycle.

## Verified to do the work, not to pass beside it

With the constructor call in place and **both `policy.rs` entry calls deleted** — simulating precisely the future mistake this prevents — all four GC-stress repros run clean with node-identical output:

| | rc | stderr | |
|---|---|---|---|
| `gc4_nothrow` | 0 | empty | matches node |
| `gc3_nested` | 0 | empty | matches node |
| `gc1_throw_alloc` | 0 | empty | matches node |
| `gc2_partial` | 0 | empty | matches node |

With the constructor call also removed, they abort again.

The added test pins the observable rather than the wiring, and asserts on **exit status and stderr as well as stdout** — the abort happens *after* the program prints its correct answer, so a stdout-only comparison passes against the broken build. That is the same trap that let #9185 ship earlier tonight.

## Validation

- `gc_stack_map_index_budgeted_cycle` 1/1 (new), `gc_copy_minor_under_heap_limit` 4/4
- perry-runtime **2873/0** (single-threaded — that suite is flaky in parallel)
- `RUSTFLAGS="-D warnings" cargo check -p perry-runtime --all-targets`: 0
- raw-handle debt 963 = baseline, local-binding-type audit OK

Refs #9182, #9191, #9231, #9233, #7726.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved garbage collection reliability by ensuring required stack-map data is initialized before scanning native roots.
  * Fixed budgeted garbage collection failures under constrained heap conditions.
  * Improved array update handling to avoid unnecessary memory-management work for already-optimized arrays.
* **Tests**
  * Added regression coverage for allocating programs using budgeted and evacuation-enabled garbage collection.


proggeramlug:
Merged.

The structural argument is the right one and the history makes it: three rounds of the same defect (#9191 wired four entries, #9231 found budgeted cycles bypass all four, #9233 wired two more by hand), with `cycle.rs` already carrying the precedent from #7726 — two sites wired, the third missed, "which is the site the shipped safepoint path actually takes." Moving the build into the constructor is the difference between an invariant you have to remember and one that holds by construction.

**The question I checked is allocation legality**, since that is the constraint #9191 was working under: the index build must happen while allocation is still legal, before the collector owns the heap. It holds, and the existing code proves it rather than my reasoning — `policy.rs:2969` already calls `ensure_stack_maps_built()` on the line *immediately before* `GcCycleState::new_full(...)`. Constructor-time and entry-time are the same program point, one frame apart. And `ensure_built` is an `Acquire` load and a return once the index exists, so the six sites that still call it earlier pay nothing for the overlap.

**What I had to fix to land it:** `cycle.rs` was at **exactly 2000** lines on `main`, so these two calls broke `check_file_size.sh` — any addition to that file would have. Rather than shave the comment down to nothing, I extracted the self-contained `malloc_trim` block (`MallocTrimOutcome`, its `cfg(test)` counters, `run_malloc_trim`) into `gc/cycle_malloc_trim.rs` with explicit named re-exports, which takes the file to 1896 and leaves headroom. I also condensed your constructor comment and moved the five-sites history into the `changelog.d/` fragment the branch was missing — it is genuinely worth recording, just not at the cost of the cap.

**Validation:** `a_budgeted_cycle_builds_the_stack_map_index_before_scanning_roots` passes; `perry-runtime` 2872 passed / 0 failed at `RUST_TEST_THREADS=1`; `perry-codegen` 31 suites / 0 failures; all 60 lint gates green. Four GC-heavy programs under default, `PERRY_GC_FORCE_EVACUATE=1`, `PERRY_GC_PROTECT_FROMSPACE=1` with a seeded aggressive schedule, and `PERRY_GEN_GC=0` all exit 0 with node-identical output and #9182's fail-closed assert firing zero times.

Landing your "adds no fix" framing as accurate: I take the point that #9233 was the fix and this removes the recurrence, not a live bug. That distinction is worth keeping in the fragment, which is why I wrote it in.

Issue #9237, opened 2026-08-31
Boolean array stores have no inline fast path: 20x node, ~82% of the time in per-store bookkeeping calls


proggeramlug:
An array store whose value is not a raw f64 — a boolean, `null`, `undefined` — permanently misses the inline store fast path, and every subsequent store pays three runtime calls whose first act is to decline to do anything.

## The measurement

Same loop, same array length, only the element type differs (Linux box, `PERRY_NO_CACHE=1`):

| loop | perry | node | ratio |
|---|---:|---:|---|
| `flags[i] = i & 1` (number[]) | 37 ms | 7 ms | 5.3× |
| `flags[i] = (i & 1) === 0` (boolean[]) | **180 ms** | 9 ms | **20×** |

Profile of the boolean version (`--debug-symbols`, `perf`):

```
23.6%  typed_feedback::plain_array_index_set_guard
20.6%  gc::layout::layout_note_slot_aware
17.5%  js_typed_feedback_plain_array_index_set_guard   (extern wrapper)
17.1%  js_array_note_numeric_write
16.3%  perry_fn_..._run                                 ← the actual loop
```

**~82% is per-store bookkeeping; 16% is the loop.**

## Why it happens

The inline guard in `lower_index_set_fast` requires the receiver to be dense in the raw-f64 sense — `_reserved & (GC_ARRAY_RAW_F64_LAYOUT | GC_ARRAY_RAW_F64_HOLES) != 0`. A boolean array loses those bits on its **first** store (that is what `js_array_note_numeric_write` does), so from the second store onward the inline guard can never pass and the store falls to `js_typed_feedback_plain_array_index_set_guard` forever.

The two note calls compound it, and both are *idempotent by their own first lines*:

- `js_array_note_numeric_write` returns immediately once the raw-f64 bits are clear — which they are, permanently, after store #1. It is called ~20M times to re-discover that.
- `layout_note_slot_aware` returns immediately when neither the new nor the old value is pointer-bearing — always true for booleans.

So the steady state is: a call to decline, a call to decline, and a guard call that does the real work.

## Two independent directions

[...]

Issue #9197, opened 2026-08-30
Three perry-runtime tests flake under parallel test runs (reproduce on unmodified main)


proggeramlug:
Three `perry-runtime` lib tests fail intermittently when the suite runs in parallel (the default). They have cost four gate runs across three PRs today, each time looking like the PR's fault:

- `gc::roots::stack_maps::decode_tests::tests::discovers_a_map_from_a_later_loaded_shared_object` — the consistent offender; it `dlopen`s a shared object
- `r#box::release_tests::completed_activation_residue_is_bounded_not_linear`
- `symbol::get::own_data_ic_tests::composed_symbol_field_cache_reloads_mutated_final_slot`

## Evidence that it is not the change under test

- **The failing set changes between runs of the same commit.** Two gate runs on one commit reported `2818 passed; 2 failed` each, with a *different* second test.
- **All three pass single-threaded**, on the branch and on `origin/main`: `cargo test --release -p perry-runtime --lib -- --test-threads=1 <names>` → 3 passed.
- **Unmodified `origin/main` reproduces it.** Three parallel runs of `cargo test --release -p perry-runtime --lib` on main: runs 1 and 2 green, run 3 `2819 passed; 1 failed` on `discovers_a_map_from_a_later_loaded_shared_object`.
- **Load is not the trigger.** It first appeared with the box at load 12–15, but also flaked at **load 1.0**, so these are racy in parallel rather than merely timing-sensitive under contention.
- The PRs where it surfaced (#9161, #9181, #9195) touch `perry-codegen` only.

## Why it is worth fixing rather than re-running

The failure is indistinguishable at a glance from a real regression, so every occurrence costs someone a bisect-or-believe decision — and the cheap resolution ("just re-run") is exactly the habit that hides a genuine flaky-looking regression later.

Shapes worth considering: a shared-object load and a symbol-registry/IC cache are both process-global, so parallel tests contend on state that is not per-test. Serializing these three (a mutex, `serial_test`, or a `--test-threads=1` lane for the modules that touch process-global registries) would remove the ambiguity.

https://claude.ai/code/session_012Ys25ni6VwDKE71o1NTYAT

proggeramlug:
Fixed by #9767, which landed on `main` via merge train #9798. The train was merged as its own branch, so the `Fixes #9197` keyword in #9767 never fired — closing manually. Verified the fix is on `main` by patch-id (`git cherry`), not just by the PR being closed.

PR #9246, opened 2026-08-31
perf(codegen): gate array-store GC bookkeeping inline (boolean-store loop −33%, prime_sieve −23%)

proggeramlug:
Takes the bounded half of #9237.

## What was happening

`gc::layout::layout_note_slot_aware` opens with `if !value_is_pointer && !old_is_pointer { return; }`. `js_array_note_numeric_write` returns as soon as the receiver's raw-f64 layout bits are clear. Both are cheap tests on values the call site already holds — and `lower_index_set_fast` emitted both calls unconditionally, so a `boolean[]` store loop paid **two calls per element to be declined**. Profiled, such a loop spends ~82% of its time in per-store bookkeeping and 16% in the loop itself.

## Two gates, not one

Both early returns are now inline, and they are deliberately **separate**, because they answer different questions:

- **Pointer gate** — layout note, string addref and write barrier, behind `may_carry_heap_pointer(new) || may_carry_heap_pointer(old)`. The test is `new || old`, never `new` alone: overwriting a pointer with a boolean is a pointer→scalar transition the runtime must still see. That is exactly why the existing new-value-only gate (`emit_jsvalue_slot_store_pointer_tested`, used for class fields under a conforming layout) is **not** reusable here — I checked before reaching for it.
- **Raw-f64 gate** — the numeric-write note, unchanged in spirit from the test `expr/index_set_guarded.rs` already applies on the sibling path. This note is what *downgrades* a raw-f64 array on its first non-numeric store; gating it on pointer-ness would skip it forever and the array would never downgrade at all.

## Measurements

Idle Mac mini, both binaries built in one run, interleaved, min of five, self-timed (the dev box was too loaded to measure on — Node itself swung 9→138 ms there):

| benchmark | base | this change | node |
|---|---:|---:|---:|
| boolean-store loop | 206 ms | **137 ms** (−33%) | 12 ms |
| `11_prime_sieve` | 26 ms | **20 ms** (−23%) | 6 ms |
| `10_nested_loops` (read control) | 16 ms | 17 ms | 16 ms |

4.3× → 3.3× Node on `prime_sieve`. **This closes part of the gap, not all of it** — the per-store typed-feedback guard *call* is the larger remaining piece and stays open in #9237.

## Why to trust it

`typed_shape_descriptors::pointer_store_into_numeric_array_keeps_layout_note_and_barrier` pins that a pointer store into a statically numeric array still reaches both the layout note and the barrier. **It caught this change twice** — once for each assertion — and needed updating both times.

Before touching it I verified the semantics independently: for a pointer store the chain `idxset.inbounds → gc_bookkeeping → numnote → barrier.maybe → barrier` is intact, both calls are reached, and the program answers identically to Node. Both assertions now follow the **edge** through the new gate blocks rather than slicing a text region of the IR — precisely the treatment #7715 gave the barrier assertion when it moved the barrier behind a live value test, and its comment says so.

Five differentials stay byte-identical to Node: the string-aliasing hazard the addref exists to prevent, the interval bounds proof, the byte-read battery, the nested-loop exit paths, and the packed-loop carry battery. 31 `perry-codegen` suites pass; `cargo fmt --check` clean.

Also removes `emit_jsvalue_slot_store_scalar_aware_with_flags_on_block`, which I added in #9195 and this change leaves callerless.

## Summary by CodeRabbit

* **Performance**
  * Improved performance for in-bounds array element stores by reducing repeated runtime bookkeeping.
  * Preserved byte-identical behavior across unaffected code paths.

* **Bug Fixes**
  * Improved reliability of garbage-collection bookkeeping and write-barrier handling during pointer and numeric stores.

* **Tests**
  * Expanded coverage to verify correct control-flow reachability for layout and barrier checks.
  * Updated validation to detect missing or unexpected bookkeeping markers.

### DIFF
--- a/changelog.d/gate-array-store-bookkeeping-inline.md
+++ b/changelog.d/gate-array-store-bookkeeping-inline.md
@@ -1,0 +1,49 @@
+An in-bounds array element store no longer calls into the runtime twice per
+element to be told there is nothing to do.
+
+`gc::layout::layout_note_slot_aware` opens with
+`if !value_is_pointer && !old_is_pointer { return; }`, and
+`js_array_note_numeric_write` returns as soon as the receiver's raw-f64 layout
+bits are clear. Both are cheap tests on values the call site already holds — and
+both were emitted unconditionally by `lower_index_set_fast`, so a `boolean[]`
+store loop paid two calls per element for the privilege of being declined. A
+profile of such a loop put ~82% of its time in per-store bookkeeping and 16% in
+the loop itself (#9237).
+
+Both early returns now happen inline, under two **separate** gates, because they
+answer different questions:
+
+* the layout note, the string addref and the write barrier are dead unless a
+  pointer is involved, so they sit behind `may_carry_heap_pointer(new) ||
+  may_carry_heap_pointer(old)`. The test is `new || old` and not `new` alone:
+  overwriting a pointer with a boolean is a pointer→scalar transition the runtime
+  must still see, which is why the existing new-value-only gate
+  (`emit_jsvalue_slot_store_pointer_tested`, for class fields under a conforming
+  layout) is not usable here;
+* the numeric-write note is what *downgrades* a raw-f64 array on its first
+  non-numeric store, so gating it on pointer-ness would skip it forever and the
+  array would never downgrade. It keeps its own raw-f64-bits gate — the test
+  `expr/index_set_guarded.rs` already applies on the sibling path.
+
+Measured on an idle Mac mini, both binaries built in one run, interleaved, min of
+five, self-timed: a boolean-store loop 206 → 137 ms (−33%), and
+`benchmarks/suite/11_prime_sieve.ts` 26 → 20 ms (−23%, 4.3× → 3.3× Node). A
+nested-loop read benchmark is unchanged, as expected. Node remains 12 ms and 6 ms
+respectively — this closes part of the gap, not all of it; the per-store guard
+**call** is the larger remaining piece and is still open in #9237.
+
+`typed_shape_descriptors`'s `pointer_store_into_numeric_array_keeps_layout_note_and_barrier`
+needed updating and is the reason to trust the change: it pins that a pointer
+store into a statically numeric array still reaches both the layout note and the
+barrier. Both assertions now follow the EDGE through the new gate blocks —
+exactly the treatment #7715 gave the barrier assertion when it moved the barrier
+behind a live value test. Verified independently that the chain
+`idxset.inbounds → gc_bookkeeping → numnote → barrier.maybe → barrier` is intact
+for a pointer store, and that such a store still answers identically to Node.
+
+Also removes `emit_jsvalue_slot_store_scalar_aware_with_flags_on_block`, added in
+#9195 and left with no callers by this change.
+
+Five differentials stay byte-identical to Node: the string-aliasing hazard the
+addref exists to prevent, the interval bounds proof, the byte-read battery, the
+nested-loop exits, and the packed-loop carry battery.
--- a/crates/perry-codegen/tests/typed_shape_descriptors.rs
+++ b/crates/perry-codegen/tests/typed_shape_descriptors.rs
@@ -817,9 +817,26 @@ fn pointer_store_into_numeric_array_keeps_layout_note_and_barrier() {
     let ir = ir_for(module);
     let inbounds_ir = block_between(&ir, "\nidxset.inbounds.", "\nidxset.check_cap.");
 
+    // #9237: the layout note took the same treatment #7715 gave the barrier
+    // below — it now sits behind a live test of the stored value (and of the
+    // slot's previous value, since a pointer overwritten by a scalar is still a
+    // transition the runtime must see), rather than inline in `idxset.inbounds`.
+    // So this follows the EDGE too: the arm must branch into the gate, and the
+    // gate must still hold the note. Asserting only that the call exists
+    // somewhere would pass even if this arm stopped reaching it.
     assert!(
-        inbounds_ir.contains("call void @js_gc_note_slot_layout"),
-        "pointer stores into statically numeric arrays must update slot layout"
+        inbounds_ir.contains("label %idxset.inbounds.gc_bookkeeping."),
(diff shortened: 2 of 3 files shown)


proggeramlug:
**Gate on the Linux box, commit `29fcb9f01`: clean.**

An earlier revision failed a **real** step — `GC store-site inventory`: my new emitter adds a third `GC_STORE_AUDIT(BARRIERED)` marker to `write_barrier.rs`, where `CODEGEN_BARRIERED_BINDINGS` pins two. That is the ratchet working as designed ("a count drift in a bound file forces this table *and the witness* to be looked at, not walked past").

Resolved rather than bumped: the new marker's unconditional slot write is discharged by its caller's stem-labelled barrier, and for its only caller (`lower_index_set_fast`) that stem is **`idxset.inbounds`** — already in `VERIFIED_BARRIER_STEMS` as `ValueAndGenerationTested` with a live IR witness. I also confirmed from the emitted IR of a pointer-into-numeric-array store that the chain `idxset.inbounds → gc_bookkeeping → numnote → barrier.maybe → barrier` is intact and the barrier call is reached, so the third claim brings no unwitnessed obligation. The pin is now 3 with that reasoning in the table comment. Locally: `GC store-site inventory passed (1679 files scanned, 316 audited sites)`, 12 codegen barrier sites bound to 8 IR-witnessed stems.

Remaining: `Address-classification audit` PASS, `String payload-access inventory` PASS, formatting PASS, CI-plan self-test PASS, no ratchet ceilings raised. The 3 FAIL steps are the environmental `${{ github.* }}` ones (a changelog fragment is present).

Runtime suite `2844 passed; 2 failed` — the known parallel-flaky pair (`discovers_a_map_from_a_later_loaded_shared_object`, `completed_activation_residue_is_bounded_not_linear`), tracked in #9197 with evidence that they reproduce on unmodified `origin/main`. This diff is codegen plus one lint-table line.

proggeramlug:
**Correction to my previous comment.** I listed `Address-classification audit` and `String payload-access inventory` as PASS for that gate run. They were not — **those steps did not run in it at all**, along with `GC store-site inventory`. The run I was reading (`FAIL=3 PASS=4`) contains only the changelog fragment, formatting, CI-plan self-test, workspace-architecture audit, the two merge-base ratchets, gap-snapshot and parity-allowlist. I carried the three lint results over from the *previous* revision's run without noticing the step list had shrunk, which is the same "absence of evidence read as evidence" mistake the missing steps would have caught.

Re-verified directly against the pushed commit `29fcb9f01` on the same box, running the scripts by hand:

```
GC store-site inventory passed (1679 files scanned, 316 audited sites, 94 allowlisted).
Claim verification: 12 codegen barrier sites bound to 8 IR-witnessed stems
Address-classification audit passed (1239 files scanned, 306 allowlisted, 521 known sites held by the ratchet)
```

Both exit 0. So the substance of the earlier comment holds — the `BARRIERED` pin at 3 is accepted by the lint, and the address-class ratchet is clean — but it holds because I have now actually run them, not because that gate did.

Unchanged and accurate from the earlier comment: formatting PASS, CI-plan self-test PASS, gap-snapshot PASS, parity-allowlist PASS, no ratchet ceilings raised; the 3 FAIL steps are the environmental `${{ github.* }}` ones; and the runtime suite's `2844 passed; 2 failed` is the known flaky pair from #9197.

proggeramlug:
Merged, with one gate fix pushed onto the branch.

**The gate change needed its own test updated.** Raising `CODEGEN_BARRIERED_BINDINGS` for `write_barrier.rs` from 2 to 3 is correct for the real tree — the file genuinely has three `GC_STORE_AUDIT(BARRIERED)` markers now — but `gc_store_site_inventory.py`'s **self-test** builds a synthetic fixture that plants exactly 2, and V-P9 adds a third to trigger the "the binding pins" drift error. So raising the binding made the green baseline dirty *and* silently disarmed V-P9:

```
V-P1 green baseline: expected clean, got ['… 2 BARRIERED marker(s) but the binding pins 3']
V-P9 bound-file count drift: wanted an error containing 'the binding pins', got []
```

That second line is the one that matters — a drift check that no longer fires is a gate that cannot fail, which is exactly what this inventory exists to prevent. Baseline now plants 3 and V-P9 adds a fourth, so both cases still test what they are named for. Worth noting the self-test caught this on its own; it is a good example of a gate whose own tests are strong enough to notice when its expectations move.

**On the correctness of the gating**, which is the part I cared about — a missed barrier is a lost root, and that surfaces cycles later as something unrelated:

The `new || old` formulation is the subtle bit and you got it right: overwriting a pointer with a boolean is a pointer→scalar transition the runtime still has to see, so testing `new` alone would have been a live bug. I probed exactly that — 400 slots filled with pointers, every even slot overwritten with a scalar, 60k allocations of churn, then every slot verified; then the reverse (scalar→pointer) with more churn; plus the `boolean[]` store loop and a string array exercising the addref arm. `ptr 200 scalar 200 bad 0 / restored 200 / bools 167 / strings 200`, byte-identical to node 26.5.1 under default, `PERRY_GC_FORCE_EVACUATE=1`, `PERRY_GC_PROTECT_FROMSPACE=1` with a seeded aggressive schedule, and `PERRY_GEN_GC=0`.

**One methodological note**, because it nearly fooled me. My first IR probe used a local `const a: boolean[] = new Array(1000)` and showed *identical* call counts on both arms — I'd have reported "verified" on a probe that never reached `lower_index_set_fast`. Your own updated `pointer_store_into_numeric_array_keeps_layout_note_and_barrier` is the better instrument, and specifically because it follows the **edge**: asserting the arm branches into `idxset.inbounds.gc_bookkeeping.` and *then* that the gate block holds the note. Your comment says it exactly — "asserting only that the call exists somewhere would pass even if this arm stopped reaching it." That label only exists with this PR, so the test is its own A/B.

**Validation:** `typed_shape_descriptors` 12/12 including both directions (`bounded_integer_array_store_omits_…` and `pointer_store_into_numeric_array_keeps_…`); `perry-codegen` 31 suites / 0 failures; `perry-runtime` 2872 passed / 0 failed; all 60 lint gates green, including both `gc_store_site_inventory` invocations.

I see #9250 extends this tier to tagged receivers — I'll pick that up next, and I'll use your test's edge-following shape rather than my own probe when I do.

Issue #9249, opened 2026-08-31
Array.prototype index setters installed with defineProperty are bypassed by array element stores


proggeramlug:
```js
Object.defineProperty(Array.prototype, 7, { set(v) { hits++; }, get() { return "P"; }, configurable: true });
const nums = [1, 2, 3];
nums[7] = 42;
console.log(hits, nums.length, nums[7]);
```

| | hits | length | `nums[7]` |
|---|---:|---:|---|
| node | 1 | 3 | `"P"` |
| perry | **0** | **8** | `42` |

The setter never runs, the array is extended, and the own slot is written. Same for a `boolean[]` receiver (`Object.defineProperty(Array.prototype, 9, …)`), and same for `Object.defineProperty(arr, i, {set})`-free plain stores generally — the receiver's own descriptors are checked, the prototype's are not.

**Pre-existing on `origin/main`** (`b3f14e9cd`), for numeric and tagged receivers alike. I found it while widening the inline store tier in a separate PR and confirmed both tiers behave identically here, so it is not caused by that work.

## Why both tiers miss it

The out-of-line guard consults `array_prototype_has_index_flag()`, `object_prototype_has_index_flag()` and `array_static_proto_recorded()`; the inline guard consults the `PERRY_ARRAY_INDEX_FAST_PATH_INVALIDATED` summary byte that those setters maintain. The flags are only ever raised from `note_array_index_write` / `note_object_prototype_index_write`, which fire on an **index write** to the prototype:

```rust
pub(crate) fn note_array_index_write(arr: usize) {
    if !ARRAY_PROTO_HAS_INDEX.load(Relaxed) && arr != 0 && arr == array_prototype_addr() {
        ARRAY_PROTO_HAS_INDEX.store(true, Relaxed);
        invalidate_array_index_fast_path();
    }
}
```

`Object.defineProperty(Array.prototype, N, …)` is not an index write, so neither flag is ever raised and every guard downstream of them is satisfied. `Array.prototype[7] = …` presumably does raise it — the hole is specific to the `defineProperty` route.

## Fix shape

Raise the same flags from the `defineProperty` / `defineProperties` path when the target is `Array.prototype` or `Object.prototype` and the key is an array index. That is one call at the definition site and leaves both guards and the summary byte correct without touching either fast path.

[...]

PR #9250, opened 2026-08-31
perf(codegen): the inline array-store tier covers tagged receivers (prime_sieve 4.5× → 1.8× node)

proggeramlug:
**Stacked on #9246** — the diff shown against `main` includes that PR's commit; review this one's second commit.

Completes the larger half of #9237.

## The gate was one conjunct

`lower_index_set_fast` builds an inline guard that tests everything the out-of-line `js_typed_feedback_plain_array_index_set_guard` tests — array type, not-forwarded, no element descriptors, integrity flags, the prototype-chain invalidation byte, length/capacity sanity — and on success jumps straight to the store, skipping the call. The whole tier was gated on `require_numeric_layout`:

```rust
let inline_write_tier = require_numeric_layout && !typed_feedback_emission_enabled();
```

So it was only ever *built* for statically numeric receivers. A `boolean[]`, or any downgraded `any[]`, took the call on every store for the life of the program — even though the in-bounds arm below already knows how to store a tagged value into such a receiver. That arm is precisely what the out-of-line guard fronts today; only the guard in front of it was missing.

Two of the guard's conditions genuinely belong to the raw-f64 store, and they now apply only when that store is the one being emitted:

- **the receiver's raw-f64 layout bits** — the raw arm writes an unboxed double into the slot, valid only while the layout says the elements are pointer-free. A downgraded receiver has those bits clear *by definition*, which is exactly why requiring them pinned `boolean[]` to the call tier permanently;
- **the runtime numeric-tag test on the value** — load-bearing because a `number[]` slot can genuinely receive a non-number (a hole/OOB read fallback returning `undefined` is the ordinary way) and the raw arm would write its NaN-boxed tag verbatim. The tagged arm stores the box as a box, so the test is dead work there and only there.

## Measurements

Idle Mac mini, all binaries built in one run, interleaved, min of five, self-timed:

| | main | with #9246 | **this change** | node |
|---|---:|---:|---:|---:|
| boolean-store loop | 207 ms | 138 ms | **64 ms** | 12 ms |
| `11_prime_sieve` | 27 ms | 20 ms | **11 ms** | 6 ms |
| `10_nested_loops` (read control) | 17 ms | 17 ms | 17 ms | 16 ms |

`11_prime_sieve` goes from **4.5× Node this morning to 1.8×**.

## Correctness

A differential written for this change, because widening a store guard is where a mistake would hide: a **frozen** array (stores ignored), a **sealed** array and one under **`preventExtensions`** (in-bounds writes allowed, growth refused), an **own element accessor descriptor** (the setter must run), **extension past length**, **mixed types through one slot**, and a store into an array that *was* numeric. Byte-identical to Node. The five pre-existing differentials are unchanged, and 31 `perry-codegen` suites pass.

**One case diverges, and it is not this change's**: an `Array.prototype` index setter installed via `Object.defineProperty` is bypassed. It diverges **identically on unmodified `main`, for numeric receivers too** — I built main specifically to check, because "my differential went red" is not the same as "my change broke it". Filed as #9249 with the mechanism: the flags both guards consult are only raised by an index *write* to the prototype, never by `defineProperty`.

## Summary by CodeRabbit

* **Performance**
  * Improved array-store performance, with faster boolean-store loops and prime-sieve benchmarks.
* **Bug Fixes**
  * Extended optimized array writes to support tagged arrays, including boolean and downgraded mixed-value arrays.
  * Preserved garbage-collection bookkeeping and write barriers for pointer-containing values.
  * Reduced unnecessary runtime bookkeeping during in-bounds array updates.
* **Tests**
  * Updated coverage to verify layout tracking and write barriers remain correct for numeric arrays.

### DIFF
--- a/changelog.d/inline-store-tier-covers-tagged-arrays.md
+++ b/changelog.d/inline-store-tier-covers-tagged-arrays.md
@@ -1,0 +1,49 @@
+The inline array-store guard now covers receivers that are not raw-f64.
+
+`lower_index_set_fast` builds a guard that tests, inline, everything the
+out-of-line `js_typed_feedback_plain_array_index_set_guard` tests: array type,
+not-forwarded, no element descriptors, the integrity flags, the prototype-chain
+invalidation byte, and the length/capacity sanity bounds. It then jumps straight
+to the store, skipping the call.
+
+That whole tier was gated on `require_numeric_layout`, so it was built only for
+statically numeric receivers. A `boolean[]` — or any downgraded `any[]` — went to
+the call on every store, forever, even though the in-bounds arm below already
+knows how to store a tagged value into such a receiver (that arm is exactly what
+the out-of-line guard fronts today).
+
+Only two of the guard's conditions belong to the raw-f64 store, and they are now
+applied only when that store is the one being emitted:
+
+* the receiver's raw-f64 layout bits, because the raw arm writes an unboxed
+  double into the slot and that is valid only while the layout says the elements
+  are pointer-free — and a downgraded receiver has those bits clear by
+  definition, which is precisely why requiring them pinned `boolean[]` to the
+  call tier;
+* the runtime numeric-tag test on the stored value, which exists because a
+  `number[]` slot can genuinely receive a non-number and the raw arm would write
+  its NaN-boxed tag verbatim. The tagged arm stores the box as a box.
+
+Measured on an idle Mac mini, all binaries built in one run, interleaved, min of
+five, self-timed:
+
+| | main | with #9246 | this change | node |
+|---|---:|---:|---:|---:|
+| boolean-store loop | 207 ms | 138 ms | **64 ms** | 12 ms |
+| `11_prime_sieve` | 27 ms | 20 ms | **11 ms** | 6 ms |
+
+`11_prime_sieve` moves from 4.5× Node to 1.8×. A nested-loop read benchmark is
+unchanged, as expected.
+
+Verified against Node on a differential written for this change specifically —
+the guard must reject exactly what the call rejects: a frozen array (stores
+ignored), a sealed array and one under `preventExtensions` (in-bounds writes
+allowed, growth refused), an element accessor descriptor (the setter must run),
+extension past length, mixed types through one slot, and a store into an array
+that was numeric. Byte-identical, plus five pre-existing differentials unchanged.
+
+One case in that differential diverges from Node — an `Array.prototype` index
+setter installed with `Object.defineProperty` is bypassed — and it diverges
+**identically on unmodified main, for numeric receivers too**, so it is neither
+caused nor widened here. Filed separately; the flag the guards consult is only
+set by an index *write* to the prototype, never by `defineProperty`.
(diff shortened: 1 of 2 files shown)


proggeramlug:
**Gate on the Linux box, commit `f0f36fc4e`: clean.** Formatting PASS, CI-plan self-test PASS, gap-snapshot PASS, parity-allowlist PASS, no ratchet ceilings raised; the 3 FAIL steps are the environmental `${{ github.* }}` ones (a changelog fragment is present). GC store-site inventory run separately on the same commit: **PASS**. Runtime suite `2845 passed; 1 failed` — `discovers_a_map_from_a_later_loaded_shared_object`, the known parallel-flaky test from #9197 that reproduces on unmodified `origin/main`. This diff is codegen only.

proggeramlug:
Merged.

"The gate was one conjunct" is the right framing and the diagnosis is convincing: the inline guard already tested everything the out-of-line call tests, and the in-bounds arm below already knew how to store a tagged value — only the guard in front of it was withheld, on `require_numeric_layout`. A `boolean[]` paying a call per store for the life of the program because of a conjunct that was about the *raw-f64* arm rather than about the store is exactly the kind of thing that survives review by looking deliberate.

**What I probed is the guard's rejection set**, since widening a fast path is only safe if the inline guard declines everything the out-of-line one declines. Byte-identical to node 26.5.1 on all of:

- `boolean[]` and a mixed `any[]` (objects, strings, integers, booleans interleaved) — 300/300 slots correct
- a non-writable element descriptor installed with `Object.defineProperty` (store must not take)
- `Object.freeze` and `Object.seal` receivers, including an out-of-bounds store on the sealed one
- growth past capacity plus a far hole (`g[400]` on a 200-element array)
- **a prototype-chain mutation** — `Array.prototype[7] = "proto7"` then reading `p[7]` and `7 in p`, which is the invalidation-byte condition the guard has to honour

Then the same file under `PERRY_GC_FORCE_EVACUATE=1`, `PERRY_GC_PROTECT_FROMSPACE=1` with a seeded aggressive schedule, and `PERRY_GEN_GC=0` — identical output on every arm, plus the #9246 pointer/scalar-transition probe (400 slots, pointers overwritten by scalars and back, 60k allocations of churn) still clean.

**One thing I added to my probe because of a mistake I made earlier today**, and it's worth passing on. When I audited #9169 I tested iterators thoroughly — `for…of`, manual `next()`, spread, Map/Set/String iterators, an own `next` shadowing the thunk, `matchAll` — and merged it. It broke `main` for four and a half hours, because every method I probed *actually existed* as an own or inherited property, and the shapes it broke were the iterator helpers Perry **synthesizes** in the dispatch tower (#9247). So this probe includes `[...gen().map(f)]`, `.take(2)` and `.filter(…)` even though they have nothing to do with array stores — the point is that a receiver-shape change should be probed against surfaces that are produced rather than stored, and that class is invisible to any probe built from properties you can see.

**Validation:** `perry-codegen` 31 suites / 0 failures; `perry-runtime` 2872 passed / 0 failed at `RUST_TEST_THREADS=1`; all static gates green including both `gc_store_site_inventory` invocations and its self-test; the gap suite running against this build shows only snapshot-recorded `parity_fail` entries (`test_gap_2159`, `test_gap_2514` — both `bug-open` in `gap_snapshot.json`, neither new).

The rebase dropped #9246's commit cleanly as already-applied, so what landed here is just the second commit.

PR #9255, opened 2026-08-31
docs(changelog): record the packed-loop throw fast path shipped in v0.5.1519

proggeramlug:
Docs only — one `changelog.d/` fragment, no code. **Cannot affect the frozen v0.5.1519 candidate** (`83754818ea`); merge whenever convenient.

## Why

#9215, #9230 and #9235 all merged without a changelog fragment, so v0.5.1519's notes carry no mention of the change. My omission — I didn't add them.

## Why this is one entry and not three

#9215 and #9230 repair a regression I introduced in #9185, and **#9185 itself was never released** — the whole introduce/revert/fix cycle happened inside the unreleased window. Verified directly: the v0.5.1220 release binary produces the correct answer for the repro, because the defect did not exist yet.

So there is no user-visible bug to describe, and describing one would be worse than silence: a reader seeing *"fixed: a taken throw lost loop-carried locals"* reasonably concludes their current version is affected, when no released version ever was. The only thing actually missing from the notes is a **Performance** entry.

## Attribution

Fragments fold into whatever release is cut next, so a bare post-tag fragment would appear in v0.5.1520's notes describing work that shipped in v0.5.1519. The fragment therefore states its shipping release explicitly, so it reads as a correction rather than as a new change.

## The entry

An 8.1× improvement on the constructing throw forms, and perry now runs every shape in that benchmark faster than node — including the loop with no `throw` in it (0.95 ns/op against 1.11). Numbers are from the quiet host, best of five, stable to 0.01 ns/op, and were re-verified on post-#9247 `main` after the 03:01Z–07:39Z breakage window.

Refs #9215, #9230, #9235.

## Summary by CodeRabbit

* **Performance**
  * Improved performance for counted loops that contain `throw` statements, preserving the packed-array fast path.
  * Benchmarks show significant speedups for common exception-throwing patterns.

### DIFF
--- a/changelog.d/9235-packed-loop-throw-fast-path.md
+++ b/changelog.d/9235-packed-loop-throw-fast-path.md
@@ -1,0 +1,31 @@
+### Performance
+
+- **A `throw` inside a counted loop no longer forces the loop off its fast
+  path.** `for (let i = 0; i < arr.length; i++) { if (bad) throw new Error(…);
+  sum += arr[i]; }` is the ordinary shape of a validating loop, and the throw
+  — even one that never fires — used to cost the whole packed-array
+  specialization. Measured on a quiet machine over a 512-element `number[]`:
+
+  | loop body | before | after | node |
+  |---|---|---|---|
+  | `throw new Error("bad")` | 7.99 ns/op | **0.95** | 7.70 |
+  | `throw new Error("bad " + i)` | 7.99 | **0.95** | 7.75 |
+  | `throw new Error()` | 7.99 | **0.95** | 7.75 |
+  | `throw "bad " + i` | 7.99 | **0.95** | 7.74 |
+  | `throw <pre-built value>` | 7.99 | **0.95** | 1.11 |
+
+  An 8.1× improvement on the constructing forms, which are the common ones.
+  Perry now runs every shape in that benchmark faster than node, including
+  the loop with no `throw` in it at all (0.95 against 1.11).
+
+  The specialization keeps a loop's accumulators in registers, so admitting a
+  `throw` required writing them back on the unwind edge — an exception leaves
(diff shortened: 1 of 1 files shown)


proggeramlug:
Merged.

This recovers a fragment for work I merged without one — #9235, #9230 and #9215 all touched `crates/`, all merged with `lint` cancelled, and the check that would have caught it is a `pull_request`-conditioned step inside that job. So the omission is mine, and #9256 documents the mechanism.

The technical summary is accurate to what I verified when auditing those PRs, including the part that is easy to get wrong: operands are admitted only when evaluating them cannot itself unwind. That distinction is load-bearing — #9235 existed because #9230's writeback sits *at* the throw site, so an operand that unwinds during its own evaluation (a `toString` or `valueOf` that throws) skips it. The fragment states the rule rather than the symptom, which is the version that survives.

Two things I would consider adding, entirely optional:

The `throw <pre-built value>` row is the interesting one and the table currently lets it pass unremarked — it is the only shape where node is faster (1.11 against 0.95 elsewhere, but 7.70+ on the constructing forms), because node has nothing to construct there either. Worth a clause, since a reader comparing columns will notice.

More substantively: this fast path shipped in v0.5.1519 **with a silent wrong answer in between**. #9185 admitted `throw` on the reasoning that a block ending in `unreachable` never returns, missed that an unwind edge does not flush loop-carried locals the way normal CFG edges do, and produced `0` where node gives `780` — reverted by #9215, fixed properly by #9230/#9235. I reviewed #9185 and cleared exactly that hazard on the wrong distinction. A sentence recording that the fast path was landed, reverted, and re-landed would make the fragment more useful than one that reads as a clean win, and it is the kind of thing release notes almost never capture.

Issue #9254, opened 2026-08-31
Design: safepoint-delimited receiver descriptors — one model to replace six ad-hoc guard-fact tables


proggeramlug:
A design proposal, not a patch. It comes out of #9253, but the pattern it describes is why that issue needed "a new loop tier" instead of reusing three existing ones.

## The defect, stated once

Codegen re-proves object properties at every use, because there is no way to express *"this receiver is validated, and stays validated until here."* A moving collector can relocate at any safepoint, so each use conservatively re-derives the lot: reload from the shadow slot, re-check the NaN-box tag, re-dereference the header, re-test the integrity flags, re-read length and capacity.

`16_matrix_multiply` is the clean specimen (#9253). 97.3% of its time is inside the generated function — **no runtime calls at all** — and `perf annotate` puts the hot instructions here:

```
2.32  cmp    $0xf42400,%r9d      ; 16,000,000 — length_sane
2.08  cmp    $0xf42400,%r9d      ; …capacity_sane
1.82  mov    (%rax),%eax         ; header length
1.32  movzwl -0x6(%rax),%ecx     ; _reserved flags
1.31  movabs $0x7ffd000000000000 ; tag mask
```

`vmulsd`/`vaddsd` are nowhere near the top. The receivers are **loop-invariant parameters**, and their guard is re-run 16.7M times. The IR agrees: 27 shadow-root stores and 5 atomic barrier-count loads to accomplish 2 `load double`s.

## Why we keep solving it privately

Every tier has answered this question on its own terms:

| table | fact | scope | invalidation |
|---|---|---|---|
| `cached_lengths` | array length | one loop | manual remove |
| `bounded_index_pairs` | `(arr, idx)` in range | `scope_id` | scope pop |
| `buffer_view_slots` | view + pointer state | region | hazard reasons |
| `packed_f64_loop_facts` | packed layout | `scope_id` + `guard_id` | side exit |
| `masked_window_array_facts` | window proven | `scope_id` | scope pop |
| `packed_receiver_box/handle_slots` + `refresh` | receiver handle | clone body | after each poll |

Six tables, six invalidation disciplines, no shared vocabulary. The consequences are concrete:

- **The tier matrix multiplies.** Admission is decided per (tier × receiver kind × index shape × bound shape). `matmul` fails on three axes at once, so it fits nothing.
[...]

PR #9257, opened 2026-08-31
codegen: receiver-region model + equivalence lint (#9254 phase 1)

proggeramlug:
Phase 1 of #9254, as filed: **the model and an equivalence lint, emitting nothing.** No lowering path consults this; it is revertible by deleting the file.

## What the survey found

Sixteen receiver-keyed fact tables on `FnCtx` answer the same two questions — *what do we know about this receiver* and *how long may we believe it* — and answer the second one five different ways:

| mechanism | tables |
|---|---|
| `retain(\|f\| f.scope_id != id)` | `bounded_index_pairs`, `packed_f64_loop_facts`, `masked_window_array_facts`, `int_range_facts`, … |
| insert/remove, no id | `cached_lengths`, `packed_receiver_*`, `versioned_indexed_loop_facts` |
| field downgraded in place | `buffer_view_slots` |
| reloaded at the safepoint | `packed_receiver_*` |
| nothing at all | `buffer_data_slots`, `class_keys_slots` |

And a boundary that **none of them expresses: the unwind edge.** `lower_try` clears no fact table. Unwind safety is obtained today by six unrelated means — the packed matcher's explicit `Stmt::Try` rejection, a body shape that cannot contain one, a post-hoc `contains_gc_unsafe_call` scan, a single `try_depth == 0` gate in `versioned_indexed_loop`, a before-call dirty bit, or the storage kind simply not being movable. Every one is a local decision by one tier.

## What this adds

`RegionEnder` (what ends a no-relocation region), `FactBoundary` (how a table expresses extent), `ReceiverClaim` — value vs representation vs address, the axis that decides whether a boundary is load-bearing at all — and `boundary_admits`, the algebra in one place.

## The lint is the point

It holds the model against `loop_purity::loop_may_allocate`, which is shipping and audited, asserting the direction with teeth:

> if the model finds no relocation point, `loop_may_allocate` must also have proven the body alloc-free.

The converse is deliberately **not** asserted — `loop_may_allocate` answers `true` for any statement it does not model (`Return`, `Switch`), which is imprecision, not a collection point.

It earned its keep before this landed. Written the obvious way — enumerate enders, default to safe — the model passed every hand-written test and **failed the battery on three entries**: generic `IndexGet`/`PropertyGet` can reach an accessor or a proxy trap; `Expr::Closure` allocates; and `is_inert` belongs on the whole coercing node, not per operand (the #6975 hole one abstraction up). Inverting to an allowlist with an `Unmodelled` catch-all closes the class — adding an HIR variant can no longer silently widen a region.

Note one deliberate divergence from `collectors::safepoint_sites`: it does **not** count property reads, because over-counting reads would over-spill a read-heavy loop and its consumer only needs a spill estimate. A region model has the opposite obligation — missing one licenses a stale cached address.

## The inventory

All sixteen tables are transcribed as test data with declaration sites, and the test pins the exact set whose unwind safety is *external to its stated boundary*: `stable_packed_loop_facts` (emergent — `stmt_flags` has a `_ => {}` arm, so `Stmt::Try` is invisible to admission and the body tail is unconstrained) plus the three immutable-fact tables leaning on non-movable storage.

**A flag is not a bug report.** It says the safety comes from somewhere the boundary vocabulary cannot express — which is what phase 2 has to fix, and pinning the set is how phase 2 proves it closed one.

## Gates

19 new tests. `perry-codegen` lib suite **1374 passed / 0 failed**. rustfmt clean; clippy reports nothing on either new file; GC store-site inventory passes unchanged.

Phase 1 wants a maintainer's agreement on the region-formation rule before anything depends on it. If the algebra in `boundary_admits` is wrong, this is the cheapest possible place to say so.

https://claude.ai/code/session_01Pcq6j6y57TdKSR2Zx2D187

## Summary by CodeRabbit

* **Refactor**
  * Added internal consistency modeling for receiver state and lifecycle boundaries.
  * Expanded handling for memory movement, conversions, suspension points, error paths, and loop checks.

* **Tests**
  * Added broad validation for boundary behavior and expression classification.
  * Added checks against existing safety and loop-analysis behavior.

* **Documentation**
  * Documented the internal model, validation coverage, and known safety considerations.

* **User Impact**
  * No user-facing behavior changes in this release.

### DIFF
--- a/changelog.d/9257-receiver-region-model.md
+++ b/changelog.d/9257-receiver-region-model.md
@@ -1,0 +1,54 @@
+**A single vocabulary for "receiver R has fact F, valid until boundary B"** (#9254
+phase 1). Codegen carries sixteen receiver-keyed fact tables on `FnCtx` —
+`cached_lengths`, `bounded_index_pairs`, `packed_f64_loop_facts`,
+`masked_window_array_facts`, `buffer_view_slots`, `int_range_facts` and ten more.
+Each answers the same two questions (what do we know about this receiver, and how
+long may we believe it), and each answers the second one differently: a
+`retain(|f| f.scope_id != id)` at scope exit, an insert/remove pair with no
+identifier, a mutable field downgraded in place and never removed, or a reload at
+the safepoint instead of an invalidation.
+
+A fifth boundary — the **unwind edge** — is expressed by none of them. `lower_try`
+clears no fact table. Unwind safety is obtained today by six unrelated means: the
+packed matcher rejecting `Stmt::Try` outright, a body shape that cannot contain
+one, a post-hoc `contains_gc_unsafe_call` scan, a single `try_depth == 0` gate in
+`versioned_indexed_loop`, a dirty bit stored before every call, and a storage kind
+that simply never moves. Every one is a local decision by one tier, and a tier
+added tomorrow inherits none of them.
+
+This adds the model — `RegionEnder` (what ends a no-relocation region),
+`FactBoundary` (how a table expresses extent), `ReceiverClaim` (value vs
+representation vs address, the axis that decides whether a boundary is
+load-bearing at all) and `boundary_admits`, the algebra in one place. **It emits
+no IR and no lowering path consults it**; the `#![allow(dead_code)]` at the top of
+the module is the marker for that, matching the #854 subgraphs in `hir_facts`, and
+the whole thing is revertible by deleting the file.
+
+The load-bearing artifact is the equivalence lint, which holds the model against
+`loop_purity::loop_may_allocate` — shipping and audited — on a shared battery,
+asserting the direction with teeth: *if the model finds no relocation point, then
+`loop_may_allocate` must also have proven the body alloc-free.* The converse is
+deliberately not asserted, since `loop_may_allocate` answers `true` for any
+statement it does not model (`Return`, `Switch`), which is imprecision rather than
+a collection point.
+
+That lint paid for itself before this landed. Written the obvious way — enumerate
+the enders, default to safe — the model passed every hand-written test and failed
+the battery on three entries: a generic `IndexGet`/`PropertyGet` can reach an
+accessor or a proxy trap, `Expr::Closure` allocates, and `is_inert` belongs on the
+whole coercing node rather than per operand (the #6975 hole one abstraction up).
+Inverting the match to an allowlist with an `Unmodelled` catch-all closes that
+class: adding an HIR variant can no longer silently widen a region.
+
+Note one deliberate divergence from `collectors::safepoint_sites`, which does not
+count property reads — over-counting reads would over-spill a read-heavy loop, and
+its consumer only needs a spill estimate. A region model has the opposite
+obligation: missing one licenses a stale cached address.
+
+All sixteen tables are also transcribed as test data with their declaration sites,
+pinning the exact set whose unwind safety is *external to its stated boundary*:
+`stable_packed_loop_facts` (emergent — `stmt_flags` has a `_ => {}` arm, so
+`Stmt::Try` is invisible to its admission scan) plus the three immutable-fact
+tables that lean on non-movable storage. A flag there is not a bug report; it says
+the safety comes from somewhere the boundary vocabulary cannot express, which is
+what phase 2 has to fix, and pinning the set is how phase 2 proves it closed one.
--- a/crates/perry-codegen/src/collectors/mod.rs
+++ b/crates/perry-codegen/src/collectors/mod.rs
@@ -43,6 +43,9 @@ mod ptr_shape_callbacks;
 mod ptr_shape_elements;
 mod ptr_shape_report;
 mod ptr_shape_returns;
+mod receiver_regions;
+#[cfg(test)]
+mod receiver_regions_tests;
 mod refs;
 mod repsel_benefit;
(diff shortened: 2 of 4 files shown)


proggeramlug:
Follow-up on the `stable_packed_loop_facts` row in the inventory test, since a flag there invites the question "is that a bug?" — I chased it, and the answer is **no, but its safety is emergent, which is the point**.

Three code-level facts, verified by reading:

1. `stmt_flags` (stmt/stable_packed_loop.rs:191) ends in `_ => {}`. That makes `Stmt::Try`, `While`, `For`, `DoWhile`, `Switch` and `Labeled` invisible to **both** the later-target-read scan and call detection in `body_has_safe_leading_read`.
2. `fast_clone_is_safe = fast_clone_call_free || capture_index.is_some() || nested_requires_access_revalidation` (:1784) — capture mode is admitted *without* the call-free scan.
3. `revalidate_before_indexed_read = nested_requires_access_revalidation && typed_array_admission.is_none()` (:1617) — false in plain capture mode, so `try_lower_index_get` skips the dirty check.

Fact 1 has a demonstrable effect on admission: taking the #8773 ECS fixture, which fires the tier, and adding a later `query[i]` read inside a `try {}` in the outer capture loop, the tier **still fires** (2 `stable_packed.loop.fast.preheader` blocks) — the hidden read did not disqualify the body.

**But it does not miscompile.** Under `PERRY_GC_FORCE_EVACUATE=1 PERRY_GC_VERIFY_EVACUATION=1` the fixture returns node's answer, and the emitted IR shows why: 7 `js_packed_arraylike_loop_revalidate_live` calls, 10 `proof_dirty` blocks and 15 `packed_index.generic_fallback` blocks. The before-call dirty bit (`dirty_stable_packed_revalidations_before_call`, block.rs:353) plus per-read revalidation and generic fallback cover the path.

So: no bug filed, because I could not produce one. Worth recording that an earlier hand-built repro of mine returned the correct answer for a completely different reason — the tier never fired on it at all (0 fast-preheader blocks, confirmed against the #8773 shape as a positive control). A correct answer from a shape that never reaches the code under test is not evidence of anything, and I nearly banked it as such.

That is exactly the situation this PR's inventory row describes: `unwind_safe_by: "EMERGENT: call-free post-scan (ordinary mode) or a dirty bit stored before every call/invoke (capture/nested modes)"`. Safety holds today because three independent mechanisms happen to cover what the admission scan cannot see. Nothing states that as a boundary, and nothing tests it — a fourth mode, or a `stmt_flags` arm added for an unrelated reason, would move it without a failing test anywhere.

I'd rather not "fix" `stmt_flags` to see `Try`/loops in this PR: it makes admission strictly narrower and would drop shapes that currently take the fast tier, so it needs measurement, not a blind tightening. Filing it as a separate hardening item unless you'd prefer it here.

proggeramlug:

**The PR failed its own lint.** `bounded_buffer_index_pairs` carried `unwind_safe_by: "arithmetic relation"` — 19 characters, under `every_inventory_row_explains_how_its_unwind_safety_is_obtained`'s `> 20` bar — so `perry-codegen` did not go green. Its immediate neighbour `bounded_index_pairs` has the same claim, boundary and `excludes_try` and reads "arithmetic relation; admission walkers descend into Try".

I deliberately did **not** just lengthen the string. That field is a safety claim about unwind behaviour, and padding it to clear a character count would satisfy the letter of the check while defeating the exact thing your own comment says it is for — "a claim that names no mechanism cannot be checked against the code later, which is how these notes go stale without anyone noticing." So I read the code and wrote a note whose every clause is checkable:

```
"arithmetic relation over local ids carrying an explicit BoundsProof;
 admission walkers descend into Try"
```

- **over local ids** — the row is `index_local_id`, `buffer_local_id`, `scope_id`, `bounds_width_units`, `BoundsState`. No heap pointer anywhere, which is what makes the `ScalarRelation` claim true: relocation moves objects, and none of those five fields is one.
- **carrying an explicit BoundsProof** — what separates it from its sibling. All three push sites (`loops.rs` 6022, 6177, 6200) record `Proven { LoopGuard }` or `Proven { MinLength }`; `bounded_index_pairs` carries no proof field at all.
- **admission walkers descend into Try** — the load-bearing clause, since `excludes_try: false`. Verified rather than copied: `stmts_mutate_local`'s `Stmt::Try` arm recurses into `body`, `catch` **and** `finally`, so a write to the counter inside a `try` is seen at admission and the pair is never recorded.

**On the model itself** — the part I'd single out is that this is an inventory with a lint that can reject its own rows, not a comment block. `an_address_claim_with_no_try_exclusion_is_flagged_on_the_unwind_edge` and `closure_allocation_is_an_ender_and_its_body_is_not_descended_into` are the ones doing real work: they encode the two mistakes this codebase has actually made (an address claim surviving an unwind edge, and a walker following a closure body it must not). A table that merely listed the facts would have been documentation; one that fails the build when a row's justification stops matching the code is a gate.

The terse row is itself the evidence that the lint earns its keep — it caught a row in the very PR that introduced it.

**Validation:** `receiver_regions` suite green, exit 0, including the inventory lint. Validated alongside #9228, #9263, #9271, #9272, #9274, #9277, #9279 and #9280 — `perry-runtime` 2885 passed / 0 failed, all 60 lint gates, and a nine-row differential probe byte-identical to node 26.5.1.

Issue #9248, opened 2026-08-31
matmul is 3.3-4.5x node: the packed-f64 range guard never engages for affine indices or a second array, so every element read carries ~40 instructions of guard


proggeramlug:
`benchmarks/suite/16_matrix_multiply.ts` runs **1006 ms against node's 304 ms (3.31×)** on the quiet host (256×256, 10 reps, best of 3, stable to 0 ms across runs). Root-caused to codegen, with the runtime side already providing what a fix needs.

## Where the time goes

A profile is unambiguous: **4653 of 4655 samples are inside `matmul` itself**, with no runtime helper on the hot path. So this is not dispatch overhead — it is what the loop body compiles to.

Disassembling one element read of `a[i * size + k]`:

```asm
fadd  d0, d13, d10          ; index computed in FLOATING POINT
fcvtzs x26, d0              ; ...then converted to integer
mov   w8, #0x7ffd           ; POINTER_TAG check
cmp   x8, x22, lsr #48
b.ne  <slow>
and   x8, x22, #0xffffffffffff
lsr   x9, x8, #20
cbz   x9, <slow>
ldurb w9,  [x8, #-0x8]      ; header byte
ldursb w10,[x8, #-0x7]
ldr   x9,  [x8]             ; forwarding indirection
csel  x8,  x9, x8, eq
sub   x9,  x8, #0x100, lsl #12   ; address-range check
cmp   x9,  x10
b.hi  <slow>
ldurb w12, [x8, #-0x8]      ; header bytes RELOADED
ldursb w11,[x8, #-0x7]
ldurh w9,  [x8, #-0x6]
adrp  x10, ...              ; global flag load
ldrb  w13, [x10]
ldp   w10, w14, [x8]        ; length + capacity
... 8 further compare/branch guards ...
cmp   w10, w26              ; bounds check
b.ls  <slow>
ubfiz x9, x26, #3, #32      ; the actual load, at last
add   x8, x9, x8
ldr   d0, [x8, #0x8]
fcmp  d0, d0                ; NaN canonicalisation
fcsel d8, d0, d15, vc
```

Roughly **40 instructions around one `ldr d0`**, twice per inner iteration, none of it hoisted. Measured: 167M inner iterations, 8.2 ns/iter (~26 cycles) against node's 1.8 ns (~6 cycles).

## The machinery already exists and works — it just does not engage here

The same compiler emits an excellent loop for `for (k) s += a[k]`:

```asm
[...]

PR #9258, opened 2026-08-31
diag(codegen): name the range matcher's rejection reason under PERRY_PACKED_LOOP_TRACE

proggeramlug:
Diagnostic only, no behaviour change. Refs #9248.

## Why

`match_packed_f64_range_loop` has **23 independent `return None` sites** and no way to tell which one declined a given loop. #9204 gave `match_packed_f64_versioned_loop` exactly this treatment, after three separate attempts on #9151 each guessed a different gate. The range matcher never got the equivalent.

I then repeated the mistake, at cost. Investigating why `s = s + a[k + 8]` runs 4× slower than `s = s + a[k]`, I widened the range matcher's index recognizer to accept a loop-invariant base (`a[base + k]`), verified it against seven adversarial cases — base assigned mid-loop, negative base, base pushing the window past `i32::MAX`, two different bases for one array — and it was **completely inert**. The loop was never being declined where I assumed. I discarded that branch.

With the trace, the same question took one run:

```
$ PERRY_PACKED_LOOP_TRACE=1 perry build probe.ts
[range-loop] rejected: body_not_admissible
```

## What it does

Each of the 23 sites returns through `range_loop_reject("<reason>")` instead of a bare `None`, naming the condition that failed — `counter_not_integer`, `bound_not_loop_invariant`, `body_not_admissible`, `counter_window_out_of_i32`, `store_not_fact_eligible`, and so on. Same env var as #9204, distinct `[range-loop]` prefix so the two matchers are separable in one run.

Off unless `PERRY_PACKED_LOOP_TRACE=1`; reasons are `&'static str`, the check is a `OnceLock` read, and every replaced site returns `None` exactly as before.

## Validation

- `RUSTFLAGS="-D warnings" cargo check -p perry-codegen --all-targets`: 0 errors
- Verified the trace answers the motivating question on first use, and that it distinguishes admitted from declined loops on single-loop fixtures

## Note for whoever continues #9248

The trace establishes that `a[k + 8]` **is** admitted by the range matcher — so the 4× cost is not an admission failure, and two of my attempted fixes were inert for that reason. It is also not the emitted fast path: the hot loop bodies are 11 instructions for `a[k]` against 13 for `a[k + 8]`. The remaining candidate is that the fast clone is not entered at runtime, or that a different matcher claims the loop first and publishes a fact with `window_validated: false` — `lower_packed_f64_versioned_for` (which does) runs at `loops.rs:5854`, before the range lowering at `:5861`, and `packed_f64_loop_fact_for_index` declines any non-zero offset unless the fact carries `window_validated` or `allow_holes`. I have not confirmed that, and say so rather than leave it as an assertion in the issue.

## Summary by CodeRabbit

* **Diagnostics**
  * Added optional tracing for packed numeric range-loop matching decisions.
  * When enabled, rejected matches now report a specific reason.
  * Loop matching behavior remains unchanged by default.

### DIFF
--- a/crates/perry-codegen/src/stmt/loops.rs
+++ b/crates/perry-codegen/src/stmt/loops.rs
@@ -1170,6 +1170,22 @@ struct PackedF64RangeLoop {
 /// potential side exit (hole-checked loads / the store's numeric-RHS check),
 /// so a side exit into the slow loop re-executes the current iteration
 /// without duplicating effects.
+/// Diagnostic twin of [`packed_loop_reject`] for the #6011 range matcher.
+///
+/// Same reasoning: the admission decision is a chain of independent
+/// conditions, and when a loop unexpectedly stays on the generic path there is
+/// no way to tell WHICH one declined it by reading the code. #9204 gave the
+/// versioned matcher this treatment after three attempts on #9151 each guessed
+/// a different gate; the range matcher had no equivalent, so #9248 guessed too
+/// — and widened a predicate that was never the thing rejecting the loop.
+fn range_loop_reject(reason: &'static str) -> Option<PackedF64RangeLoop> {
+    use std::sync::OnceLock;
+    static ON: OnceLock<bool> = OnceLock::new();
+    if *ON.get_or_init(|| std::env::var("PERRY_PACKED_LOOP_TRACE").as_deref() == Ok("1")) {
+        eprintln!("[range-loop] rejected: {reason}");
+    }
+    None
+}
 fn match_packed_f64_range_loop(
     ctx: &FnCtx<'_>,
     init: Option<&Stmt>,
@@ -1179,7 +1195,7 @@ fn match_packed_f64_range_loop(
 ) -> Option<PackedF64RangeLoop> {
     use perry_hir::{CompareOp, Expr, UpdateOp};
     if !ctx.pending_labels.is_empty() {
-        return None;
+        return range_loop_reject("pending_labels");
     }
     let (counter_id, start) = match init? {
         Stmt::Let {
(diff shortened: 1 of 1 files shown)


proggeramlug:

The motivation is the right one and it is the second time this exact argument has paid: #9204 gave the versioned matcher a named rejection reason after three attempts on #9151 each guessed a different gate, and #9248 then guessed on the range matcher — widening a predicate that was never the thing rejecting the loop. An admission chain of independent conditions where the only feedback is "it didn't fire" costs hours, repeatedly, and a one-line `eprintln!` ends it.

**Verified the diagnostic actually works in both directions**, because a diagnostic that never prints is indistinguishable from a dead one and I have merged an inert instrument before:

- flag off → **zero** `[range-loop]` lines
- flag on → four distinct reasons on shapes chosen to hit different gates: `init_not_integer_literal` (non-literal init), `pending_labels` (labelled loop), `start_out_of_i32_range` (negative start), `written_binding_not_eligible`

So it discriminates rather than printing one catch-all.

**What I fixed:** the new `range_loop_reject` was inserted between `match_packed_f64_range_loop`'s 14-line doc comment and the function itself, which silently reassigns that whole block — the receiver-guard explanation, the runtime index-window rationale, the one-statement body restriction — to the helper, and leaves the matcher undocumented. Rustdoc does not complain; the only symptom is that the docs are attached to the wrong thing. I moved the helper above the matcher's doc block so both keep their own.

Worth flagging because it is now the third time I have seen it in this repo (#9161, and again here after a rebase re-applied the same insertion): **adding a function immediately above an existing one silently steals its doc comment.** Cheap to check — `git diff` shows the `+` lines landing between a `///` run and its `fn`.

No build-cache work needed: `PERRY_PACKED_LOOP_TRACE` is already in `BUILD_CACHE_ENV_EXCLUSIONS` from #9204, and correctly so — `range_loop_reject` prints and returns `None` either way, so emitted code is identical with the flag on and off.

**Validation:** `perry-codegen` 31 suites / 0 failures; `perry-runtime` 2872 passed / 0 failed at `RUST_TEST_THREADS=1`; all 60 lint gates plus the TLS checkers green.

PR #9262, opened 2026-08-31
fix(release): build macOS x86_64 on an Intel runner; keep Android off Windows ARM64

proggeramlug:
Two release build legs that block `create-release`, which requires **every** matrix leg to succeed.

## macOS x86_64 — wrong runner architecture

`macos-15` is the **Arm64** image. [actions/runner-images](https://github.com/actions/runner-images) lists `macos-15-intel` / `macos-15-large` as the x86_64 one. So `x86_64-apple-darwin` was being built on an arm64 runner, where `brew install llvm@22` installs an arm64-only LLVM:

```
ld: warning: ignoring file .../libLLVM…
ld: symbol(s) not found for architecture x86_64
```

This became fatal when the in-process LLVM backend became the default (2026-08-17) and `perry` started linking libLLVM. The artifact shipped fine in v0.5.1220 (2026-07-04) and has been broken since.

LLVM publishes **no x86_64 macOS binary** (22.1.8 ships only `LLVM-22.1.8-macOS-ARM64.tar.xz`), so a native Intel runner is the only source of an x86_64 libLLVM. `macos-13` is retired.

Dropping the artifact was considered and rejected: `scripts/stage-npm.sh` builds the `darwin-x64` npm package from it — and `npm-publish` gates `create-release` — and the homebrew job downloads it to compute the formula's SHA256.

Scoped deliberately to the compiler `build` leg. The `build-cross` x86_64 leg already passes on arm64, because `perry-ui-macos` does not link LLVM.

## Windows ARM64 — Android cross-build on a runner that cannot do it

The step resolves its toolchain under `prebuilt/windows-x86_64` and the NDK ships no Windows-arm64 host toolchain, so it could only ever have run on the x86_64 runner — which is also the only Windows runner with an NDK installed. It failed with `ANDROID_NDK_HOME not set`. Now gated on `runner.arch == 'X64'`; the same library is already produced by the Linux `aarch64-linux-android` build-cross leg.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored reliable macOS x86_64 release builds by using a native Intel build environment.
  * Prevented unsupported Android cross-compilation steps from running on Windows ARM64 runners, improving release build reliability.

* **Documentation**
  * Added release notes describing the macOS and Windows build fixes.

### DIFF
--- a/.github/workflows/release-packages.yml
+++ b/.github/workflows/release-packages.yml
@@ -375,7 +375,14 @@ jobs:
             target: aarch64-apple-darwin
             artifact: perry-macos-aarch64
             old_glibc_image: ""
-          - os: macos-15
+          # macos-15-intel, NOT macos-15: the bare label is the Arm64 image
+          # (see actions/runner-images). Building x86_64-apple-darwin there
+          # links brew's arm64-only LLVM, so every LLVM symbol is undefined at
+          # link once the in-process backend became default — the artifact
+          # shipped in v0.5.1220 and has been failing since. LLVM publishes no
+          # x86_64 macOS binary, so a native Intel runner is the only source of
+          # an x86_64 libLLVM.
+          - os: macos-15-intel
             target: x86_64-apple-darwin
             artifact: perry-macos-x86_64
             old_glibc_image: ""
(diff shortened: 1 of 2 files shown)


proggeramlug:
Merged — it unblocks `create-release`, and both diagnoses check out.

`macos-15` being the Arm64 image is the kind of thing that reads as correct forever: the label looks like "the macOS 15 runner", the build succeeded for as long as LLVM wasn't linked, and it only started failing when the in-process backend became default. The comment you left in the YAML is the right artifact, because the next person to touch that matrix will see `macos-15-intel` and wonder.

Verified what I could without running CI: the workflow still parses, `macos-15-intel` is in place, both Windows Android steps are narrowed to `runner.arch == 'X64'` (a narrowing, so it cannot break a leg that works today), and the `aarch64-linux-android` library is still produced by the Linux `build-cross` leg at line 1104 — so gating the Windows steps drops a duplicate, not the only source.

**One thing this does not fix, worth a follow-up.** The same pairing exists a second time:

```
1081:          - os: macos-15
1082:            target: x86_64-apple-darwin
1083:            ui_crate: perry-ui-macos
```

That is the `build-cross` matrix, and it runs `./.github/actions/setup-llvm22` too — so it is building `x86_64-apple-darwin` on the Arm64 image with an arm64-only LLVM, exactly the configuration you just diagnosed. It has been invisible because `build-cross` sets `fail-fast: false` and does not gate `create-release` the way the legs you fixed do, so it degrades to a missing x86_64 UI bundle rather than a red release.

I did not fold it into this merge because it is a different failure mode (silent artifact loss, not a blocked release) and because I cannot exercise the runner change locally — but it is the same one-word fix, and shipping x86_64 macOS bundles without their UI library is worth closing before the next release rather than after someone reports it.

Issue #9218, opened 2026-08-30
RegExp . excludes only \n — JS also excludes \r, U+2028 and U+2029 (silent extra matches)


proggeramlug:
## Summary

Perry's `.` (without the `s` flag) excludes **only** `\n`. ECMAScript's `.` matches any character *except* a **LineTerminator**, which is `\n`, `\r`, U+2028 (LINE SEPARATOR) and U+2029 (PARAGRAPH SEPARATOR) — §22.2.2.7.

```js
"\t\r\n".match(/.{2}/g)
// node:  null
// perry: [ "\t\r" ]
```

## Impact

Silent over-matching, in the direction that matters: a `.`-based pattern intended to stay within one line will run across a `\r`, and on CRLF input that is every line. Line-oriented parsers, log scanners and anything splitting on `\r\n` can consume a boundary they were written to stop at. No error, no crash — just a longer match than the author specified.

## How it was found

A corpus differential of 3,402 regex patterns against 12 probe subjects, run against `node --experimental-strip-types` on current `main`. **64 of 3,402 records (1.9%) diverge**, split between this and #9217.

Surfaced while working on a regex *performance* change and independent of it — the differential runs on unmodified `main`.

## Fix direction

In `crates/perry-runtime/src/regex/grammar.rs`, a bare `.` should translate to a class excluding all four LineTerminators rather than relying on the Rust engine's `\n`-only default. Under the `s` (dotAll) flag `.` matches everything including all four, which perry already handles by mapping `s` onto the Rust engine's dot-matches-newline mode — so only the **non**-`s` case needs to change.

Note the interaction with #9216, which translates JS's `[^]` to `(?s:.)`: that rewrite is the dotAll form and is unaffected by this fix, but whoever fixes this should confirm both together, since they touch the same translation site.

The corpus differential is a ready-made regression instrument: it should go from 64 diverging records to the count attributable to #9217 alone.

proggeramlug:
Closing: fixed in #9263 (2026-08-31); fragment 9217-9218-regexp-ascii-word-dot.md and fixture test_gap_9217_9218_regexp_word_dot.ts are on main.

Issue #9239, opened 2026-08-31
A prototype method replaced after class registration is not seen by newly constructed instances


proggeramlug:
Replacing a method on a class prototype is observed by instances that already exist, but **not** by instances constructed afterwards.

```ts
class C { m() { return "orig"; } }
const c = new C();
console.log(c.m());                                    // orig
(C.prototype as any).m = function () { return "replaced"; };
console.log(c.m());                                    // replaced  ✓
console.log(new C().m());                              // orig      ✗  node: replaced
```

| | existing instance | fresh instance |
|---|---|---|
| node 26.5.1 | `replaced` | `replaced` |
| perry | `replaced` | `orig` |

The asymmetry is the useful clue: the existing instance picks the replacement up, so the guard invalidation on the shared prototype works. It is construction that reinstalls the original surface — a fresh `new C()` appears to take the class's declaration-prototype vtable as recorded at class-registration time rather than consulting the (now mutated) prototype object.

A 200-iteration loop calling `c.m()` returns `replaced`, and a second replacement followed by another 200 iterations returns `again`, so the inline-cache path re-checks correctly for an existing receiver.

**Confirmed pre-existing on `main`** by A/B — rebuilding with #9169's runtime files reverted gives the identical `orig`. Found while auditing #9169, which fixes two adjacent per-instance `[[Prototype]]` cases and does not address this one.

Monkey-patching a class prototype after construction is common in test doubles and instrumentation libraries, and the failure is silent: the call succeeds and returns the stale implementation.

PR #9263, opened 2026-08-31
fix(runtime): RegExp \w/\W/\b use ASCII word semantics and . excludes all four LineTerminators

### LINKED ISSUE #9217: RegExp \w, \W and \b use Unicode word semantics where JS is ASCII-only — /[\w-]+/ matches "éΩ中"
## Summary

Perry's `\w`, `\W` and `\b` use **Rust's Unicode word-character semantics**. ECMAScript defines them as **ASCII-only** (`[A-Za-z0-9_]`), and the `u` flag does not change that — the only thing that widens them is `i` + `u` adding Kelvin sign and long s to `\w`.

```js
/:?[\w-]+/.exec("éΩ中")
// node:  null
// perry: matches
```

`\b` inherits the same divergence, since word boundaries are defined in terms of `\w`.

## Impact

This is a **silent wrong answer** on one of the most common constructs in JS regexes. Tokenizers, slug matchers, identifier scanners and `\b`-anchored word searches all over-match on non-ASCII input. There is no error and no crash — the pattern simply matches more than the program's author specified, which in a validation or sanitisation path is the dangerous direction.

## How it was found

A corpus differential of 3,402 regex patterns against 12 probe subjects, run against `node --experimental-strip-types` on current `main`. **64 of 3,402 records (1.9%) diverge**, split between this and #(the `.` issue filed alongside).

This was surfaced while working on a regex *performance* change and is entirely independent of it — the differential runs on unmodified `main`.

## Fix direction

The translation in `crates/perry-runtime/src/regex/grammar.rs` should emit an explicit ASCII class for `\w`/`\W` rather than letting `regex_syntax`'s Unicode-aware default stand, and `\b` should use the ASCII word-boundary form. The `i` + `u` case needs the two extra code points (U+212A KELVIN SIGN, U+017F LATIN SMALL LETTER LONG S) added to `\w` and removed from `\W`, per §22.2.2.9.

The corpus differential above is a ready-made regression instrument: it should go from 64 diverging records to the count attributable to the other bug alone.

proggeramlug:
Fixes #9217 and #9218 — two silent wrong answers in the JS→Rust regex translation, both of which made patterns match **more** than the program asked for.

| differential | before | after |
|---|---:|---:|
| focused corpus, 67 patterns × 12 subjects | 49 records / 175 cells | **11 records / 12 cells** |
| gap fixture, 106 output lines | mismatch | **byte-identical to node** |

## #9217 — `\w`, `\W`, `\b`, `\B` were Unicode

Perry emitted Rust's Unicode word-character semantics. ECMAScript defines the word set as ASCII-only `[A-Za-z0-9_]`, and the `u` flag does not widen it — the only widening is `i`+`u` adding U+212A KELVIN SIGN and U+017F LATIN SMALL LETTER LONG S (§22.2.2.9.3 WordCharacters). So `/:?[\w-]+/` matched `éΩ中`, and `\b` inherited the divergence because word boundaries are defined in terms of `\w`.

Mixed and negated character classes now preserve JS case-insensitive semantics **without letting Rust case-fold the word arm** — that separation is what makes `[\w\W]` and `[^\w\W]` come out right under `i`.

## #9218 — `.` excluded only `\n`

ECMAScript's `.` matches any character except a **LineTerminator**: `\n`, `\r`, U+2028, U+2029 (§22.2.2.7 CompileAtom, §12.3). Perry excluded only `\n`, so `"\t\r\n".match(/.{2}/g)` returned `["\t\r"]` where node returns `null` — `.`-patterns ran straight across CRLF boundaries. DotAll (`s`) is unchanged.

## #9216's rewrite is preserved

The `[^] → (?s:.)` and `[] → [a&&b]` translations that avoid the catastrophic case-fold (`[\s\S]` under `i` walked 1,114,112 code points) remain intact, with **no full-range class reintroduced**. A regression test additionally pins that quantifiers apply atomically to the rewritten negated classes.

## Verification

- The differential harness (`scripts/regex_9217_9218_differential.mjs`) was **run against unchanged `main` first and confirmed to fail** on the cases fixed here.
- After the fix, zero divergences across: accented Latin, Greek, CJK, Kelvin sign, long s, LF, CR, U+2028, U+2029, and the `/.{2}/g` CRLF case.
- Gap fixture `test_gap_9217_9218_regexp_word_dot.ts`: node and perry produce identical 106-line output (sha256 `b989905fdf66e0ed…`).
- `cargo test -p perry-runtime --lib -- --test-threads=1`: **2,870 passed, 0 failed**, 4 ignored. Existing regex tests 66/66.

## A correction to the issue reports

The 3,402-pattern corpus differential that produced #9217 and #9218 reported 64 diverging records on `main` and attributed all of them to these two bugs. **That attribution was not quite right.** Record 484 — `/[ \t]+$/gm` against `"\t\r\n"` — is an **independent multiline-`$`/CRLF divergence**, and it is deliberately left unchanged here.

The other residual divergences after this fix are likewise pre-existing and unrelated: non-`u` UTF-16 code-unit matching for astral characters (emoji), and multiline anchors around CRLF. The fixture uses `u` for its astral case specifically so it does not conflate those with this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated regular expressions to match ECMAScript word-character and word-boundary behavior.
  * Corrected `.` so it excludes all line terminators unless dotAll mode is enabled.
  * Improved handling of Unicode case-insensitive matching, character classes, and empty classes.

* **Tests**
  * Added comprehensive compatibility tests comparing regular-expression behavior across supported engines.

### DIFF
--- a/changelog.d/9217-9218-regexp-ascii-word-dot.md
+++ b/changelog.d/9217-9218-regexp-ascii-word-dot.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- **RegExp word escapes/boundaries and non-dotAll `.` now follow ECMAScript
+  instead of Rust's defaults (#9217, #9218).** `\w`, `\W`, `\b`, and `\B`
+  use the spec's ASCII `[A-Za-z0-9_]` word set; `i`+`u` additionally admits
+  U+212A KELVIN SIGN and U+017F LATIN SMALL LETTER LONG S. A `.` without `s`
+  now excludes all four LineTerminators (`\n`, `\r`, U+2028, U+2029), while
+  dotAll still matches every character. The cheap #9216 translations for
+  `[^]` and `[]` remain `(?s:.)` and `[a&&b]`, avoiding full-range case folds.
--- a/crates/perry-runtime/src/regex/grammar.rs
+++ b/crates/perry-runtime/src/regex/grammar.rs
@@ -379,6 +379,13 @@ pub(super) fn has_unicode_forbidden_pattern(pattern: &str) -> bool {
 // JS \s excludes U+0085 NEL; Rust's \s includes it. Use explicit class for parity outside char-class.
 const JS_WHITESPACE_CLASS: &str = r"[\t\n\x0B\x0C\r\x20\x{A0}\x{1680}\x{2000}-\x{200A}\x{2028}\x{2029}\x{202F}\x{205F}\x{3000}\x{FEFF}]";
 const JS_NON_WHITESPACE_CLASS: &str = r"[^\t\n\x0B\x0C\r\x20\x{A0}\x{1680}\x{2000}-\x{200A}\x{2028}\x{2029}\x{202F}\x{205F}\x{3000}\x{FEFF}]";
+// ECMA-262 CharacterClassEscape: unlike Rust's Unicode-aware `\w`, the JS
+// word set starts as exactly these ASCII characters. Under BOTH `i` and
+// Unicode mode, Canonicalize adds the two non-ASCII code points whose simple
+// case fold lands in that set (LONG S -> s, KELVIN SIGN -> k).
+const JS_ASCII_WORD_MEMBERS: &str = r"A-Za-z0-9_";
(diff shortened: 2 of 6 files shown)


proggeramlug:
Merged.

**Results across the batch:**

PR #9269, opened 2026-08-31
fix(link): preserve futures channel wrapper provider

### LINKED ISSUE #9121: release blocker: shared-deps strip drops futures_channel SenderTask::notify provider
## Release impact

Full CI run 33285503318 failed in cargo-test-perry (2/8), job 99187911731, blocking v0.5.1519.

## Exact failure

    cargo test -p perry --test issue_5920_wrapper_bundled_runtime_async_starvation fire_and_forget_fetch_survives_recompile -- --exact --nocapture

The compile reaches the archive-fresh wrapper-before-stdlib link path after rebuilding perry-ext-http. Shared-dependency stripping reports 760 bundled members dropped, then ld fails with undefined references to futures_channel::mpsc::SenderTask::notify from retained perry-ext-http receiver CGUs.

## Diagnosis

strip_bundled_shared_deps_from_well_known_lib removes the futures_channel provider because the matching stdlib member archive index advertises the symbol. In the final prepared archive/link shape that replacement does not resolve the retained wrapper consumers, so the fixed-point proof is insufficient for this Rust cross-staticlib symbol.

## Acceptance criteria

- Keep the bundled futures_channel provider whenever a retained wrapper member references one of its symbols and the final link cannot safely substitute stdlib.
- Add a focused fixed-point unit regression for this provider/consumer shape.
- The exact issue_5920 integration test compiles twice and both binaries print PASS.
- Full CI and simulator parity pass.

## Evidence

https://github.com/PerryTS/perry/actions/runs/33285503318/job/99187911731

proggeramlug:
## Summary

- keep the wrapper-local futures_channel SenderTask::notify provider when retained wrapper members reference it, even when the matching stdlib archive index advertises the symbol
- preserve the existing Ring native-companion safety rule
- add focused predicate and fixed-point regressions plus a changelog fragment

Fixes #9121.

## Testing

- cargo fmt --all -- --check
- cargo test -p perry --bin perry commands::compile::strip_dedup::strip_dedup_tests -- --nocapture
- cargo test -p perry --test issue_5920_wrapper_bundled_runtime_async_starvation fire_and_forget_fetch_survives_recompile -- --exact --nocapture
- cargo clippy -p perry --bins

The integration regression compiles the same fetch program twice and requires both binaries to print PASS.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed prepared HTTP archives that could fail when receiving data through asynchronous channels.
  * Ensured required channel notification support remains available when archives are optimized.
  * Preserved compatibility for applications using futures-based HTTP handling.

* **Tests**
  * Added coverage for channel notification and archive optimization scenarios.

### DIFF
--- a/changelog.d/9121-futures-channel-wrapper-provider.md
+++ b/changelog.d/9121-futures-channel-wrapper-provider.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Preserve the wrapper-local `futures_channel` wake provider when prepared HTTP archives retain receiver code that calls it.
--- a/crates/perry/src/commands/compile/strip_dedup.rs
+++ b/crates/perry/src/commands/compile/strip_dedup.rs
@@ -1791,7 +1791,7 @@ fn shared_dep_members_to_remove(
             let replacement_defined = replacement_defined_by_candidate.get(c).unwrap_or(&empty);
             let still_needed = kept_undefined.iter().any(|s| {
                 defined.contains(*s)
-                    && (!replacement_defined.contains(*s) || requires_bundled_native_companion(s))
+                    && (!replacement_defined.contains(*s) || requires_bundled_wrapper_provider(s))
             });
             if still_needed {
                 to_remove.remove(c);
@@ -1815,6 +1815,22 @@ fn shared_dep_members_to_remove(
 /// links fail with hundreds of undefined `ring_core_*` references.
 fn requires_bundled_native_companion(symbol: &str) -> bool {
     symbol.trim_start_matches('_').starts_with("ring_core_")
+}
+
+/// Symbols whose same-named stdlib archive-index entry is not sufficient
+/// replacement evidence for a wrapper-first link.
+///
+/// `futures_channel::mpsc::SenderTask::notify` can appear in the matching
+/// stdlib member's archive index while remaining unavailable to retained
+/// `perry-ext-http` receiver CGUs in the final link (#9121). Keep the wrapper's
+/// provider whenever one of those retained CGUs references it. Rust v0
+/// mangling preserves identifier text, so the predicate works on the raw
+/// symbol spelling used by `llvm-nm` as well as on demangled test fixtures.
+fn requires_bundled_wrapper_provider(symbol: &str) -> bool {
+    requires_bundled_native_companion(symbol)
+        || (symbol.contains("futures_channel")
+            && symbol.contains("SenderTask")
+            && symbol.contains("notify"))
 }
 
 mod object_format;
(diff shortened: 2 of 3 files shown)


proggeramlug:
Merged.

The distinction the new predicate draws is the subtle one: a symbol appearing in the matching stdlib member's **archive index** is not the same as that symbol being *available to the retained `perry-ext-http` receiver CGUs in the final link*. Treating index presence as replacement evidence is what dropped the wrapper's provider, and `futures_channel::mpsc::SenderTask::notify` is a good canary because it is reachable only through the retained CGUs.

Splitting `requires_bundled_wrapper_provider` out from `requires_bundled_native_companion` rather than widening the latter is the right call — they answer different questions (`ring_core_*` is about a bundled native companion; this is about wrapper-first link ordering), and folding them would have made the next person reading either predicate wrong about both.

**Validation:** `perry-runtime` 2872 passed / 0 failed; `perry-codegen` 31 suites / 0 failures; all 60 lint gates green. Note that the strip/dedup path is one where my local coverage is weaker than CI's — the failure mode is a link-time undefined symbol on a configuration I do not build here (this is the same family as #8907, which only reproduced on a released binary), so I am relying on the suite plus your reproduction rather than an independent repro of the original break.

proggeramlug:
Audited the diff — **sound, and I'd merge it once `pr-gate` is present and passing.** One narrowness note below that I'd file as a follow-up rather than block on.

The fixed-point regression is the part that convinced me. `issue_9121_keeps_indexed_sender_notify_provider` constructs the exact shape the bug needs — candidate member defines `SenderTask::notify`, a retained `perry_ext_http_receiver.o` leaves it undefined, and the replacement *also* advertises it — then asserts the candidate survives. That fails without the predicate change, so it is a real regression test rather than a restatement of the fix. It also sits alongside `issue_8930_requires_needed_symbol_in_matching_stdlib_member`, which asserts the opposite outcome for a different symbol, so the two pin both directions.

## The narrowness (follow-up, not a blocker)

`requires_bundled_wrapper_provider` keys on symbol text:

```rust
symbol.contains("futures_channel") && symbol.contains("SenderTask") && symbol.contains("notify")
```

That fixes this symbol, not the class. The underlying defect — a stdlib archive index advertising a symbol that is not actually resolvable for retained wrapper CGUs in the final link — is untouched, so the next symbol with that shape fails the same way and needs another hardcoded arm.

I'm not blocking on it, for two reasons: it mirrors the existing precedent right above it (`requires_bundled_native_companion` hardcodes `ring_core_`), and #9121 is a release blocker where a targeted fix now beats a general one later. Worth a follow-up issue to make replacement evidence account for final-link resolvability rather than index presence.

## Verification caveat

#9121 is build-state dependent — it passed on candidate `01463b8161` while still being broken, which is why I wrongly closed it earlier today. A green targeted run does not settle it on its own; the fixed-point unit test is the stronger evidence here because it encodes the shape rather than the outcome.

PR #9270, opened 2026-08-31
fix(codegen): use live receiver in stable packed loops

### LINKED ISSUE #9117: release blocker: stable packed loop reads stale forwarded array after cross-call growth
## Release blocker

The full release CI run for v0.5.1519 fails deterministically in:

- `cargo-test-perry (8/8)`
- `issue_5139_object_arraylike_method_dispatch::any_typed_real_arrays_keep_array_mutator_semantics`

Observed output:

```text
grow: 1000 1000 165171
```

Expected sum: `499500`. Sampled array values (`0, 1, 2, 99, 500, 999`) are correct, so storage is intact and the optimized reduction is wrong.

## Reproduction

```bash
RUST_TEST_THREADS=1 cargo test -p perry \
  --test issue_5139_object_arraylike_method_dispatch \
  any_typed_real_arrays_keep_array_mutator_semantics -- --exact --nocapture
```

Reproduced independently on `perrymaster.skelpo.net` at release SHA `259cf8b68dc1303cc4119595f96d458a52af8c4e`.

## Bisect

Automated exact compiler/runtime bisect:

- good: `da0eefbf99a80da0b00020c0ae1c4bb88a1147ca`
- first bad: `57926713b078df354d7cf76ff11eaa18d5de76fb` (#9060)

The array is grown across a call boundary and the caller retains a forwarding stub. `js_packed_arraylike_loop_guard` follows that stub and validates the live target, but the ordinary-local stable-packed path discards the live address and reloads/de-tags the stale local for direct fast-clone reads. #9060 makes that clone reachable for the reduction by classifying `js_shadow_slot_set` as call-safe.

## Required fix

Use `js_packed_arraylike_loop_guard_live` for ordinary-local admission as well as captured receivers, and feed its returned live receiver address to the fast clone. Retain the generic slow clone on a zero result. Update IR assertions to expect the stronger live guard.

Before merging/releasing, verify:

- the exact issue_5139 test above;
- all three issue_8690 versioned-loop tests, including three live-guard calls in the read-only IR fixture;
- issue_8655 and issue_8773 packed-loop IR/runtime coverage;
- quick pre-tag checks.

proggeramlug:
## Summary

Port the root fix already present on `release/v0.5.1519` to `main`. Stable packed loops now consume the live Array address returned by admission, preventing direct reads through a stale forwarding stub after cross-call growth.

## Changes

- Use `js_packed_arraylike_loop_guard_live` for ordinary-local as well as captured receivers.
- Feed the returned nonzero live address to the fast clone while preserving the generic slow clone on zero.
- Update the #8655 and #8690 IR contracts to require the live guard and add a changelog fragment.

## Related issue

Fixes #9117

## Test plan

- [x] Baseline exact #9117 test reproduced `165171` instead of `499500`.
- [x] `RUST_TEST_THREADS=1 cargo test -p perry --test issue_5139_object_arraylike_method_dispatch any_typed_real_arrays_keep_array_mutator_semantics -- --exact --nocapture` (1/1)
- [x] `cargo test -p perry --test issue_8655_array_subclass_indexing` (2/2)
- [x] `cargo test -p perry --test issue_8690_loop_versioned_arraylike` (3/3)
- [x] `cargo test -p perry --test issue_8773_closure_capture_packed_loops` (4/4)
- [ ] `cargo build --release` clean (not run; scoped release static-runtime build and tests above passed)
- [ ] Full workspace test suite (not run; scoped regression targets above passed)
- [x] Existing runtime and IR regression tests cover the affected behavior.
- [x] Documentation update is not applicable; this is an internal codegen correctness fix.
- [x] Platform UI build is not applicable.

## Screenshots / output

Before: `grow: 1000 1000 165171`

After: the regression assertion passes with `grow: 1000 1000 499500`.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed stable packed loops after array growth so they continue using the live array address instead of a stale forwarding reference.
  * Improved loop handling for arrays referenced through ordinary local variables.

* **Tests**
  * Updated regression coverage to verify live array receivers are used during optimized loops.

### DIFF
--- a/changelog.d/9270-stable-packed-forwarded-receiver.md
+++ b/changelog.d/9270-stable-packed-forwarded-receiver.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Make stable packed loops consume the guard's live Array address after cross-call growth instead of reading a stale forwarding stub.
--- a/crates/perry-codegen/src/stmt/stable_packed_loop.rs
+++ b/crates/perry-codegen/src/stmt/stable_packed_loop.rs
@@ -1592,21 +1592,21 @@ pub(super) fn lower(
                 ),
                 (PTR, descriptor.as_str()),
             ];
-            if candidate.capture_index.is_some() {
-                let live_raw =
-                    ctx.block()
-                        .call(I64, "js_packed_arraylike_loop_guard_live", &guard_args);
-                (
-                    ctx.block().icmp_ne(I64, &live_raw, "0"),
-                    Some(live_raw),
-                    None,
-                )
-            } else {
-                let guard = ctx
-                    .block()
-                    .call(I32, "js_packed_arraylike_loop_guard", &guard_args);
-                (ctx.block().icmp_ne(I32, &guard, "0"), None, None)
-            }
+            // An ordinary local can already name an Array growth forwarding
+            // stub when the array was grown across a call boundary. The guard
+            // follows that edge and validates the live target, so codegen must
+            // consume the live address it returns instead of reloading and
+            // de-tagging the stale local for direct fast-clone reads (#9117).
+            // Captured receivers need the same result because their reload is
+            // itself a runtime call that can move the target.
+            let live_raw =
+                ctx.block()
+                    .call(I64, "js_packed_arraylike_loop_guard_live", &guard_args);
+            (
+                ctx.block().icmp_ne(I64, &live_raw, "0"),
+                Some(live_raw),
+                None,
+            )
         };
     // The conservative column matcher proves the cloned body call-free, and
     // `fast_raw` below reloads the rooted derived receiver after every
@@ -1633,18 +1633,10 @@ pub(super) fn lower(
         descriptor_word(ctx, &descriptor, 6)
     };
     let bound_i32 = ctx.block().trunc(I64, &bound64, I32);
-    // A capture reload is itself a runtime call, so its admission returns the
-    // post-call live address. Ordinary addressable bindings retain the old
-    // guard/reload sequence; their reload is a plain load and their clone must
-    // still pass the call-free scan unless it has explicit access revalidation.
-    let fast_raw = if let Some(live_raw) = admitted_live_raw {
-        live_raw
-    } else {
-        let fast_receiver = crate::expr::lower_expr(ctx, &Expr::LocalGet(candidate.array_id))?;
-        let fast_bits = ctx.block().bitcast_double_to_i64(&fast_receiver);
-        ctx.block()
-            .and(I64, &fast_bits, crate::nanbox::POINTER_MASK_I64)
-    };
+    // Every admission returns the post-guard LIVE receiver. Besides captured
+    // reloads, this is required for ordinary locals that still hold an Array
+    // growth forwarding stub (#9117).
+    let fast_raw = admitted_live_raw.expect("every packed-loop admission returns a live receiver");
     let fast_scan_start = ctx.func.num_blocks();
     let installed_typed_array_views = typed_array_admission
         .as_ref()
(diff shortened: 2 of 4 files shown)


proggeramlug:
Merged.

The defect class is the one that costs the most to find later: a local naming an Array growth **forwarding stub**, then the fast clone reloading and de-tagging that stale local for direct reads. Consuming the live address the guard already computed is the right shape — the guard was doing the work and codegen was throwing the answer away.

Probed the shapes that produce a stub rather than trusting the description:

- growth **across a call boundary** (`grow(arr)` pushing 300 elements, then summing) — 303/44856
- growth **mid-loop** on the receiver being iterated (`if (i === 10) a.push(999)`) — the case where the reload happens after the move
- a **captured** receiver, the arm that already used the live guard, as a control
- 30 rounds of growth interleaved with 20k-allocation churn, so a collection lands inside the loop

All byte-identical to node 26.5.1, and identical again under `PERRY_GC_FORCE_EVACUATE=1` and `PERRY_GC_PROTECT_FROMSPACE=1` with a seeded aggressive schedule — the arm that faults at the faulting instruction if a stale from-space address is dereferenced.

Worth noting the cost this accepts: the non-captured arm now always calls `js_packed_arraylike_loop_guard_live` (I64) instead of the cheaper `js_packed_arraylike_loop_guard` (I32). That is a real per-loop-entry cost paid for correctness, and it is the right trade — but it is the kind of thing that looks like an unexplained regression to whoever benchmarks this path next, so the comment naming #9117 is doing useful work.

Also re-ran the #9246 and #9250 array-store probes against this build; both still match node, so the tier changes and this one do not interact.

**Validation:** `perry-runtime` 2872 passed / 0 failed at `RUST_TEST_THREADS=1`; `perry-codegen` 31 suites / 0 failures; all 60 lint gates plus the TLS checkers green. Validated on a shared branch with #9258, #9260 and #9269.

proggeramlug:
Audited the diff (not the description) — **this is a genuine root-cause fix and I'd merge it once `pr-gate` is present and passing.**

The change removes the capture/ordinary-local branch so both paths call `js_packed_arraylike_loop_guard_live` and *consume* the address it returns. That matches #9117's symptom precisely: the old ordinary-local path called the boolean `guard`, validated the live target, then discarded the result and reloaded the stale local. Storage stayed intact — which is why sampled values read correctly — while the reduction walked the forwarding stub. Exactly the reported `165171` instead of `499500`.

The two IR-contract edits are **tightenings, not weakenings**, which is the thing I check for:

```diff
- system.contains("call i32 @js_packed_arraylike_loop_guard(")
+ system.contains("call i64 @js_packed_arraylike_loop_guard_live(")
```

`guard` returns `i32` (a proof bit); `guard_live` returns `i64` (the live address). The updated contracts demand the stronger artifact, and the assertion messages were updated to say so rather than left stale.

One note for the record, since it may save someone else the detour: the PR summary says the fix is "already present on `release/v0.5.1519`". I could not verify that. `stable_packed_loop.rs` does differ between `01463b8161` (the previous candidate) and `main`, but the difference is in `match_candidate` — the candidate had a `None => ctx.prelowered_zero_for_counter?` arm that `main` dropped — and both have exactly one `guard_live` call site. So the provenance claim doesn't hold up as written. It doesn't matter for the merge: the change stands on its own reasoning.

**Verification caveat worth honouring:** #9117 is optimization-state dependent — it passed on candidate `01463b8161` while still being broken, which is why I wrongly closed it earlier today. A green targeted run is not sufficient evidence on its own here. What makes this convincing is the root cause being identified and addressed, not the test going green.

PR #9271, opened 2026-08-31
test(compiler-output): align native-region proofs with current lowering

### LINKED ISSUE #9150: release blocker: native-region proof contracts lag packed-loop and byte-read lowering
## Release blocker

Full CI run 33298955917 failed `compiler-output-regression` in the `native-region-proof` suite at release SHA `97bae62ad9ee781830e609eda17429c16ccb1285`.

The exact three failing workloads are:

- `image_convolution`
- `packed_f64_loop_versioning`
- `packed_f64_loop_versioning_negative`

## Diagnosis

The immediately preceding release revision `330638064a` passes all three on `perrymaster.skelpo.net`. The later mainline optimization merge changes the structural contracts:

1. PR #9146 correctly restores byte-array `+=` to native numeric addition while preserving OOB `undefined`. The image blur now contains only the 75 documented `js_number_coerce` calls for kernel coefficients; the PR explicitly declined the unsound flat-const coefficient proof. Native evidence consequently records `unknown_bounds` materialization for the byte-or-undefined boundary.
2. PR #9041 removes the per-store guard/side-exit for statically genuine packed-f64 store values. The old proof still requires the deleted `packed_f64_loop_store.fast/fallback` pair.
3. PR #9111 caches the packed receiver handle, adding one intentional `ptrtoint` in the aggregate fast-clone region.
4. PR #9084 admits locally built read loops behind the runtime packed-f64 entry guard. `frozenLoop`, `sealedLoop`, and `nonExtensibleLoop` are safe read-only admissions; the old negative proof rejects any guard anywhere in the fixture. The genuinely unsupported holey/sparse/accessor/non-number/any/unknown-index/alias-mutating functions still emit no packed-f64 guard.

## Required fix

- merge current `main` so #9146 is present;
- allow only `js_number_coerce` in the image blur and `unknown_bounds` native materialization while retaining all byte-load/store and bounds/alias evidence checks;
- replace the obsolete packed store side-exit assertion with a check-free raw-f64 fast-store assertion;
- permit the single receiver-cache `ptrtoint`;
- scope the negative guard exclusion to the genuinely unsupported functions and positively require guards for the three safe read-only functions.

## Acceptance

- the three workloads above pass individually on the dedicated Linux server;
- the complete `native-region-proof` suite passes;
- harness unit tests pass;
- Full CI `compiler-output-regression` passes before release publication.

proggeramlug:
## Summary

- allow the image blur documented `js_number_coerce` calls and `unknown_bounds` native materialization while preserving byte-load/store, bounds, and alias evidence
- replace the obsolete packed-store side-exit assertion with a check-free raw-f64 fast-store contract and allow the single receiver-cache `ptrtoint`
- require guards for frozen, sealed, and non-extensible read-only loops while keeping genuinely unsupported loops guard-free

## Testing

- `python3 -m unittest tests.test_compiler_output_regression tests.test_native_abi_evidence_report` (96 tests passed)
- fresh individual captures for `image_convolution`, `packed_f64_loop_versioning`, and `packed_f64_loop_versioning_negative`
- complete `native-region-proof` suite (11/11 workloads passed)

No version bump.

Closes #9150

## Summary by CodeRabbit

* **Tests**
  * Updated compiler regression expectations for image convolution runtime calls and materialization cases.
  * Expanded packed floating-point loop checks to verify optimized raw-value stores and appropriate guard behavior.
  * Added coverage for supported and unsupported loop scenarios, including sparse, accessor, non-number, and alias-mutating cases.

### DIFF
--- a/benchmarks/compiler_output/workloads.toml
+++ b/benchmarks/compiler_output/workloads.toml
@@ -16,7 +16,7 @@ schema_version = 1
 source = "benchmarks/honest_bench/workloads/3_image_convolution/perry/image_conv.ts"
 kind = "image_convolution"
 allow_hot_loop_conversions = true
-allowed_hot_loop_runtime_calls = ["js_gc_loop_safepoint"]
+allowed_hot_loop_runtime_calls = ["js_gc_loop_safepoint", "js_number_coerce"]
 
 [workloads.image_convolution.vectorization]
 min_vectorized_loops = 0
@@ -88,6 +88,7 @@ detail = "object disassembly contains the FNV multiply constant"
 name = "input_generation"
 required = true
 no_runtime_calls = true
+allowed_runtime_calls = ["js_gc_loop_safepoint"]
 
 [[workloads.image_convolution.named_regions.selectors]]
 counter_min = { store_i8 = 1 }
@@ -128,6 +129,7 @@ counter_min = { load_i8 = 1, store_i8 = 1, xor_i32 = 1 }
 name = "blur"
 required = true
 no_runtime_calls = true
+allowed_runtime_calls = ["js_gc_loop_safepoint", "js_number_coerce"]
 
 [[workloads.image_convolution.named_regions.selectors]]
 counter_min = { load_i8 = 3, store_i8 = 1 }
@@ -147,6 +149,7 @@ name = "fnv_hash"
 required = true
 no_runtime_calls = true
 no_conversions = true
+allowed_runtime_calls = ["js_gc_loop_safepoint"]
 
 [[workloads.image_convolution.named_regions.selectors]]
 counter_min = { load_i8 = 1, xor_i32 = 1, mul_i32 = 1 }
@@ -160,7 +163,7 @@ detail = "fnv hash i32 xor/mul shape"
 
 [workloads.image_convolution.native_rep_checks]
 materialization_regions = ["input_generation", "blur", "fnv_hash"]
-allow_materialization_reasons = ["function_abi", "return_abi"]
+allow_materialization_reasons = ["function_abi", "return_abi", "unknown_bounds"]
 
 [[workloads.image_convolution.native_rep_checks.require_records]]
 name = "image_input_generation_buffer_view"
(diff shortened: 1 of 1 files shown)


proggeramlug:
Audited the diff. **The contract updates are justified and mostly *strengthen* the proofs — I'd merge this.** One bounded-ness note at the end that I'd take as a follow-up, not a blocker.

This PR touches only `workloads.toml`, so it asserts that the current lowering is correct and the proofs were stale. #9150's own analysis concluded exactly that for each of the four merged PRs, and every change here maps onto one of them:

| change | #9150's finding |
|---|---|
| `js_number_coerce` allowed | #9146 restores byte-array `+=` to native addition; the blur's coerce calls are the declined flat-const coefficient proof |
| `unknown_bounds` materialization allowed | #9146's byte-or-undefined boundary records it |
| `ptrtoint` max 0 → **1** | #9111 caches the packed receiver handle, adding one intentional conversion |
| store-loop contract rewritten | #9041 deleted the per-store guard/side-exit the old proof required |
| negative-guard proof rescoped | #9084 admits frozen/sealed/non-extensible read-only loops behind the entry guard |

**Two of these are strengthenings rather than relaxations**, which is the distinction I was checking for:

- The store-loop contract replaces `regex_all` (requiring a `fast`/`fallback` pair that no longer exists) with `regex` + `regex_none` that positively *forbid* `…numeric_array_index_set_guard` and `…array_index_set_fallback_boxed` in the fast body. The old proof merely failed to find a shape; the new one asserts the guard is absent.
- The negative-guard proof moves from a blanket "no `packed_f64_array_loop_guard` anywhere in the fixture" to per-function assertions: `regex_all` **requires** the guard in `frozenLoop`/`sealedLoop`/`nonExtensibleLoop`, and `regex_none` **excludes** it from the seven genuinely unsupported loops. That pins both directions where the old one pinned neither precisely.

`ptrtoint` also stays bounded at exactly 1 rather than being dropped.

## The one loosening worth a follow-up

`allowed_runtime_calls`/`allowed_hot_loop_runtime_calls` are membership lists, not counts. #9150 describes the blur as containing "only the 75 documented `js_number_coerce` calls", but adding the symbol to the allowlist permits *any* number. A future change that multiplied those calls would not be caught by this proof — and coerce calls in a hot loop are exactly the kind of regression this suite exists to detect. The neighbouring `counter_min = { store_i8 = 1 }` shows the schema already supports counted assertions.

Not blocking #9150, which is a release blocker and correctly scoped. Worth a follow-up to pin the count.

proggeramlug:
Merged.

**Results across the batch:**

PR #9272, opened 2026-08-31
perf(runtime): an inline address window in front of the registry probes — cc --help −5.16% instructions, −6.30% cycles

proggeramlug:
The registry-probe family was **6.86%** of a symbolized `claude --help` profile after round one (#9176/#9177) cleared part of it. This takes it to **2.65%**.

| | instructions (min / median) | cycles (min / median) |
|---|---|---|
| before | 7,160,271,524 / 7,163,283,415 | 3,446,625,309 / 3,479,536,202 |
| after | 6,791,077,579 / 6,793,829,953 | 3,229,414,851 / 3,267,521,023 |
| | **−5.16% / −5.16%** | **−6.30% / −6.09%** |

11 interleaved reps, both arms built from `b3f14e9cde` in one session. `--help` output **byte-identical to node** (9,175 bytes, rc=0) in both arms.

**Cycles fall by more than instructions and IPC rises 2.077 → 2.102.** That is the check that matters here: an earlier change in this campaign removed 24% of instructions and moved cycles +1%, because the removed work rode free in superscalar slack. This is the opposite — the removed work was real.

## The answer distribution is the whole argument

Measured with uretprobes on a real `cc --help` run, not inferred:

- `is_registered_buffer_slow`: **4 "yes" out of 4,650,058** — one per 1.16 million
- `lookup_registered_typed_array_kind`: **0 "Some" out of 3,566,956**
- `is_closure_ptr`: the `CLOSURE_MAGIC` tag test **alone** partitions 2,240,934 calls 231,704 / 2,009,230 — bit for bit the partition the whole function produces — yet it ran *last*, after `classify_heap_generation` and a GC-header read that changed no answer.

So the fix is an inline address window in front of the buffer/typed-array probes, and a test reorder for the closure probe. 98.03% of calls now end at the window, and re-running the answer census on the shipped binary confirms **all 4 genuine "yes" answers survive**.

## Attribution says the fix does not belong in the caller

Exact uprobe counts, ~100% coverage:

| probe | calls | top callers |
|---|---|---|
| `is_registered_buffer_slow` | 4,650,245 | `get_accessor_descriptor` 19.2%, `js_array_get_f64` 16.6%, `js_object_get_field_by_name` 12.3%, `object_static_prototype` 11.7% … **119 callers** |
| `lookup_registered_typed_array_kind` | 3,567,035 | `js_array_get_f64` 21.2%, `js_object_get_field_by_name` 15.2% … **99 callers** |
| `is_closure_ptr` | 2,240,612 | `closure_get_dynamic_prop` 26.2%, `closure_dynamic_prop_by_key` 19.8% … **81 callers** |

It takes ~10 callers to reach 90%. There is no caller-side fix; it is the generic property path, and one structural change covers all of it.

## Why it cannot misclassify

These probes classify pointers — a wrong answer is type confusion, not a slow path.

The window **never dereferences the candidate**. Each probe consults a *closed* set of tables, and each table has a single module-private insert funnel that `admit`s the address first. That is an enumeration, and enumerations of writer sets have produced silent wrong answers in this codebase twice recently — so it is not left as one: **under `debug_assertions` every window rejection is re-derived from the authoritative tables**, so a registration route added without `admit` panics in the first test that touches it. 2,858 unit tests ran with that armed; none fired.

## The `admit` race, and a second form of it

A previous iteration had a non-atomic load+store in `admit` that could drop a concurrent registration; that was already fixed to `fetch_min`/`fetch_max`. **It left a subtler second form behind, removed here:** a `Relaxed` "skip if already covered" pre-check in front of the RMWs. A thread that *skips* the RMW performs no acquire, so another thread's widening never joins its happens-before graph, and a reader synchronising only with this thread's publish is not guaranteed to see the bound covering the address — a false negative for a registered pointer.

`admit` is now two unconditional `AcqRel` RMWs with no pre-check. It runs **52 times** in a 6.9-billion-instruction run; the fast path bought nothing and cost the proof.

## Three precise negatives

- **A tighter page-bitmap filter is not worth building.** The window already removes 98.0% / 97.2%; the residual is 0.28% of profile, and a data-dependent load on all 8.2 M probes would likely cost more than it saves. An address-histogram estimate said 48% — it was wrong because it came from a third binary, and the direct measurement overruled it.
- **"Classify from the GC header instead of a registry" is unsound**, and two in-tree tests prove it: `type_probes_skip_offheap_typed_array_with_unmapped_preceding_page` (a header read segfaults on a `PROT_NONE` preceding page) and `native_memory_copy_rejects_buffer_registry_forged_to_old_non_buffer` (a registered buffer whose header claims `GC_TYPE_OBJECT`).
- **A caller-side idea inherited from earlier notes was wrong.** They claimed `meta_capable_object`'s `is_registered_buffer` call is redundant with a later `obj_type == GC_TYPE_OBJECT` check. The comment on that line says the opposite, and the forged-buffer test is exactly the case the registry check catches. Removing it would have been type confusion.

## Follow-up

Four more probes fit the same shape and are now the largest remaining members of the family: `is_registered_symbol_slow` (0.60%), `is_registered_class_prototype_object` (0.47%), `is_registered_box_ptr` (0.33%), `is_uint8array_buffer_slow` (0.21%).

## Summary by CodeRabbit

* **Performance**
  * Improved registry probe performance by quickly rejecting addresses that cannot correspond to registered buffers or typed arrays.
  * Reduced unnecessary registry lookups for invalid addresses.
* **Bug Fixes**
  * Improved closure detection by validating closure metadata earlier.
  * Strengthened concurrent registration handling to prevent missed registrations.
  * Ensured shared array buffer registrations remain discoverable.
* **Tests**
  * Added coverage for address filtering, concurrent registration, shared array buffers, and registry visibility.

### DIFF
--- a/changelog.d/registry-probe-address-window.md
+++ b/changelog.d/registry-probe-address-window.md
@@ -1,0 +1,59 @@
+**The two hottest side-table probes answer "no" from an inlined address
+compare instead of an out-of-line registry lookup.**
+
+`is_registered_buffer` and `lookup_typed_array_kind` are asked "is this pointer
+special?" from ~239 and ~200 generic call sites — every property get/set, every
+prototype walk, every element read, every `[[HasProperty]]`. #9176 gave both a
+monotone "has anything ever been registered?" latch, which makes the answer free
+for a program that never allocates a `Buffer` or a typed array. `claude-code`
+allocates both, so the latch is armed and stops discriminating.
+
+How badly it stops discriminating was the finding. Counted with uprobes on a
+symbolized `claude --help` (12 reps for the profile, exact counts from one run):
+
+| probe | registrations | calls | answered "yes" |
+|---|---|---|---|
+| `is_registered_buffer_slow` | 10 | 4,651,086 | **4** |
+| `lookup_registered_typed_array_kind` | 42 | 3,567,647 | **0** |
+
+465,000 probes per registered buffer, and 85,000 per registered typed array,
+every one of them going out of line to a thread-local resolution, a `RefCell`
+borrow and a hash — for the typed-array probe, also a direct-mapped negative
+cache whose cold miss *writes back* and dirties a shared cache line.
+
+`RegistryAddrWindow` is the same monotone idea applied to the address rather
+than to the fact of registration: a process-global `[lo, hi]` that every
+registration widens *before* it publishes, checked inline at the call site. An
+address outside it cannot be in any table the window covers, so rejecting is
+sound; accepting falls through to the exact lookup that was already there. It is
+strictly stronger than a latch — an unregistered process has the empty window
+`[usize::MAX, 0]`, which contains nothing.
+
+It is deliberately *not* a `GcHeader` tag test. A registered typed array is not
+required to have a readable `ptr - GC_HEADER_SIZE` (see the `mprotect`ed
+guard-page fixture in `promise::combinators`), and `native_arena`'s
+`native_memory_copy_rejects_buffer_registry_forged_to_old_non_buffer` pins that
+a registry entry may legitimately disagree with the header type. The window
+never dereferences the candidate, so neither case can be misclassified.
+
+`buffer::header` already had this filter as the thread-local `BUFFER_ADDR_RANGE`
+— but *behind* the call, where it rejected 87.8% of probes and still paid the
+call, the prologue, two thread-local resolutions and a tail call into
+`is_shared_sab` for every one of them. The window is the same test hoisted in
+front of the call and widened to cover the external and `SharedArrayBuffer`
+registries too, so those routes keep their existing behaviour.
+
+Measured on `claude-code` (`cli_2.1.112.js`, `--help`, both arms built from
+`b3f14e9cde` in one session, `PERRY_DEBUG_SYMBOLS=1`, 9 interleaved reps):
+
+<!-- NUMBERS -->
+
+Output stays byte-identical to `node cli_2.1.112.js --help` (9,175 bytes, rc=0).
+
+Left for a follow-up, both measured on the same run and reported with it:
+`closure::dynamic_props::is_closure_ptr` (1.34%) consults no registry at all —
+its cost is the four-way arena page-range cache inlined from
+`classify_heap_generation`, which runs before the `CLOSURE_MAGIC` test that
+would reject 89.7% of its callers; and `object::prototype_chain::meta_capable_object`
+calls `is_registered_buffer` immediately before an `obj_type == GC_TYPE_OBJECT`
+check that already excludes every buffer with a header.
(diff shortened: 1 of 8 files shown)


proggeramlug:
Merged.

**The machine-check is the thing worth reviewing here**, and I verified it fires rather than trusting that it exists.

Context for why I care: this is the fourth PR in this family, and I found real defects in the first two — #9176 armed its latch on one of *two* insert paths into the external-Uint8Array registry, so a WebCrypto key registered on one thread answered "no" from every other; #9177 didn't widen its range on the collector's re-key, so a symbol evacuated past its allocation's bounds became invisible to `typeof`. Both were "the writer set is complete" claims that were not, and both took a hand audit to find.

Replacing the enumeration with `assert!(!is_registered_buffer_slow(addr))` on the rejection path is the correct response, and your comment states exactly why an enumeration cannot hold: it "is a snapshot that a later commit can invalidate silently, and the failure it would cause is a misclassified pointer, not a slow path."

**Verified by sabotage**, because a clean run only proves the net didn't trip. I removed `register_buffer`'s `admit()` and re-ran the buffer tests in a debug profile:

```
BUFFER_LIKE_ADDR_WINDOW rejected 0x20001c40020, but it IS a registered buffer.
Some registration route reached BUFFER_REGISTRY, the external-buffer registry
or the shared-SAB registry without calling `BUFFER_LIKE_ADDR_WINDOW.admit()`
(via `register_buffer` or `note_buffer_like_registered`) first.
```

Address, which tables to look in, which function to call — caught by the first test touching that route. That is #9176's bug class turned from a hand audit into a build failure.

**One thing worth knowing about its reach.** It is `#[cfg(debug_assertions)]`, so it is compiled out of `--release`, and CLAUDE.md notes `--profile perry-dev` inherits `release` too. My habitual validation is `cargo test --release -p perry-runtime`, which means I would have signed off on this net without ever compiling it — I had to build debug specifically to see it. Its entire protective value rests on some CI arm running these tests with `debug_assertions` on. Worth confirming that arm exists and covers `perry-runtime`, because if it doesn't, the assertion is documentation.

The measurement discipline is also right, and I'd single it out: cycles fall more than instructions (−6.30% vs −5.16%) with IPC rising 2.077 → 2.102, contrasted against an earlier change in this campaign that removed 24% of instructions and *raised* cycles because the work rode free in superscalar slack. That contrast is what separates real removal from bookkeeping, and most perf PRs don't make it.

**Validation:** `perry-runtime` 2879 passed / 0 failed at `RUST_TEST_THREADS=1`; 124 buffer tests green in a debug profile with the machine-check live; all 60 lint gates plus the TLS checkers. Five differential probes (tagged/scalar array stores, pointer↔scalar transition churn, growth-forwarding receivers, typed-array and Buffer identity) byte-identical to node 26.5.1 under default, `PERRY_GC_FORCE_EVACUATE=1`, and `PERRY_GC_PROTECT_FROMSPACE=1` with a seeded aggressive schedule. Pushed one `cargo fmt` fix for the closure type-tag read.

The one probe divergence is pre-existing and unrelated: `new Uint8Array(1e10)` throws `RangeError` where node allocates — Perry's >2 GB cap, unchanged at 2 diff lines, first noted on #9181.

